PR #466, opened 2018-10-30
Parse flags on startup.

mindprince:
The generated code had imported flag but was not using it.

Not having flag.Parse() in the beginning was resulting in errors like the following when we hit a glog logged line.
ERROR: logging before flag.Parse: ...


droot:
@mengqiy tests are failing because golden files need to be regenerated. Can you help @mindprince with the steps to generate golden files. And, we should also add that to our instructions for contributing

mengqiy:
> we should also add that to our instructions for contributing.

We can probably add the instructions in the error message in the scaffolding tests.

@mindprince Please run `./generated_golden.sh`.

mindprince:
Done.

PR #480, opened 2018-11-14
add integration testing guide

crimsonfaith91:
Added a `.md` guide for integration testing.


crimsonfaith91:
@droot Will you have time to review this guide?

integration.md
```diff
@@ -0,0 +1,104 @@
+**Writing and Running Integration Tests**
+
+This article explores steps to write and run integration tests for Kubebuilder.
+
+Based on an example controller that watches *Parent* objects that in turn create *Child* objects, Kubebuilder provides framework for writing integration tests under `pkg/controller/parent/parent_controller_test.go`. Below is a detailed example of the file:
+```
```
droot: We can improve the clarity if we use some *real* resources for parent and child. Consider using `ReplicaSet` as Parent and `Pod` as child.

integration.md
```diff
@@ -0,0 +1,104 @@
...
+
+	// Verify Child is created.
+	child := &childv1alpha1.Child{}
+	g.Eventually(func() error { return c.Get(context.TODO(), childKey, child) }, timeout).
+		Should(gomega.Succeed())
+
+	// Manually delete Child since GC isn't enabled in the test control plane.
+	g.Expect(c.Delete(context.TODO(), child)).To(gomega.Succeed())
+}
+```
```
droot: Consider removing boilerplate stuff such as import statements. Highlight the stuff which user is tied to reconcile logic and how to test that.

integration.md
```diff
@@ -0,0 +1,104 @@
+**Writing and Running Integration Tests**
+
+This article explores steps to write and run integration tests for Kubebuilder.
+
+Based on an example controller that watches *Parent* objects that in turn create *Child* objects, Kubebuilder provides framework for writing integration tests under `pkg/controller/parent/parent_controller_test.go`. Below is a detailed example of the file:
```
droot: Also, one paragraph explaining what the test does. For. ex. how setup function brings up an APIServer and Etcd. Manager is started as part of the test itself. I have seen users struggling with it.

integration.md
crimsonfaith91: I included them to ensure what does `parentapis`, `parentv1alpha1`, `childapis`, and `childv1alpha1` means. I will remove all other import statements except for these 4.

integration.md
droot: I understand all of that you said above. Using some concrete resource examples are bit easier to understand. Also, I realized that when it comes to testing, testing core resources should not be any different from testing CRD resources :)

Don't have any strong opinion, so lets go ahead with the merge.

droot:
Looks all good (except that we should move it under `docs` dir), pl. go ahead and squash the commits. Thanks.

integration.md
crimsonfaith91: > testing core resources should not be any different from testing CRD resources
Agreed, but we should also think what is the focus of kubebuilder, which (in my opinion) is about reconciliating custom resources, not core resources. Just my 2 cents. :D

droot:
Thanks @crimsonfaith91 for the PR.

PR #485, opened 2018-11-14
Update GitBook Release Process

pwittrock:
- Use `npm-shrinkwrap.json` for specifying module deps
- Pin `lodash` module to `4.17.5` transitively
- Use `npm ci` for installing node_modules
- Release from `_book`, so no more need to copy files


docs/book/RELEASE
```diff
@@ -1,6 +1,32 @@
...
+
+## Run locally
+
+Check that everything looks good
+
+- `$ gitbook serve`
+
+## Build the book
+
+Book the index.html & friends
```
droot: s/Book/Build

pwittrock: Done

PR #493, opened 2018-11-21
Add bash installation script for macos or linux users

### LINKED ISSUE #482: [Feature] Bash install script
## Feature

I propose that we release a bash script in order to simplify the installation process of the kuberbuilder binary

Commands for macos
```
export version=1.0.5
export arch=amd64
curl -L -O https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${version}/kubebuilder_${version}_darwin_${arch}.tar.gz
tar -zxvf kubebuilder_${version}_darwin_${arch}.tar.gz
sudo mv kubebuilder_${version}_darwin_${arch} /usr/local/kubebuilder
```

to be converted to something like this 

```
curl -L https://git.io/getLatestKubebuilder | sh -
#!/bin/sh
#
# This file will be fetched as: curl -L https://git.io/getLatestKubebuilder | sh -
# so it should be pure bourne shell, not bash (and not reference other scripts)
#
# The script fetches the latest kubebuilder release candidate and untars it.
# It lets users to do curl -L https://git.io/getLatestIstio | KUBEBUILDER_VERSION=0.1.5 sh -
# for instance to change the version fetched.

OS="$(uname)"
if [ "x${OS}" = "xDarwin" ] ; then
  OSEXT="darwin"
else
  OSEXT="linux"
fi
ARCH=amd64

if [ "x${KUBEBUILDER_VERSION}" = "x" ] ; then
  KUBEBUILDER_VERSION=$(curl -L -s https://api.github.com/repos/kubernetes-sigs/kubebuilder/releases/latest | \
                  grep tag_name | sed "s/ *\"tag_name\": *\"\\(.*\\)\",*/\\1/")
fi

# https://github.com/kubernetes-sigs/kubebuilder/releases/download/v1.0.5/kubebuilder_1.0.5_darwin_amd64.tar.gz

KUBEBUILDER_VERSION=${KUBEBUILDER_VERSION#"v"}
NAME="kubebuilder_${KUBEBUILDER_VERSION}"
URL="https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${KUBEBUILDER_VERSION}/${NAME}_${OSEXT}_${ARCH}.tar.gz"
echo "Downloading $NAME from $URL ..."
curl -L "$URL" | tar xz
# TODO: change this so the version is in the tgz/directory name (users trying multiple versions)
echo "Downloaded into $NAME: "
ls "${NAME}_${OSEXT}_${ARCH}/bin"
sudo mv ${NAME}_${OSEXT}_${ARCH} /usr/local/kubebuilder
echo "Add kubebuilder to your path; e.g copy paste in your shell and/or ~/.profile:"
echo "export PATH=$PATH:/usr/local/kubebuilder/bin"
```

cmoulliard:
- Add bash installation script for macos or linux users.
- Download latest version of kuberbuilder or specify the version, untar it and move project under /usr/local/bin/kubebuilder if it is empty. Otherwise exit
- Fix #482

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+#
+# This file will be  fetched as: curl -L https://git.io/getLatestKubebuilder | sh -
+# so it should be pure bourne shell, not bash (and not reference other scripts)
+#
+# The script fetches the latest kubebuilder release candidate and untars it.
+# It lets users to do curl -L https://git.io//getLatestKubebuilder | KUBEBUILDER_VERSION=1.0.5 sh -
+# for instance to change the version fetched.
+
+OS="$(uname)"
+if [ "x${OS}" = "xDarwin" ] ; then
+  OSEXT="darwin"
+else
+  OSEXT="linux"
+fi
+ARCH=amd64
+
+if [ "x${KUBEBUILDER_VERSION}" = "x" ] ; then
+  KUBEBUILDER_VERSION=$(curl -L -s https://api.github.com/repos/kubernetes-sigs/kubebuilder/releases/latest | \
+                  grep tag_name | sed "s/ *\"tag_name\": *\"\\(.*\\)\",*/\\1/")
+fi
+
+KUBEBUILDER_VERSION=${KUBEBUILDER_VERSION#"v"}
+KUBEBUILDER_VERSION_NAME="kubebuilder_${KUBEBUILDER_VERSION}"
+URL="https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${KUBEBUILDER_VERSION}/${KUBEBUILDER_VERSION_NAME}_${OSEXT}_${ARCH}.tar.gz"
+echo "Downloading ${KUBEBUILDER_VERSION_NAME} from $URL ..."
+curl -L "$URL" | tar xz
+
+echo "Downloaded these executable files into $NAME: "
+ls "${KUBEBUILDER_VERSION_NAME}_${OSEXT}_${ARCH}/bin"
+mv ${KUBEBUILDER_VERSION_NAME}_${OSEXT}_${ARCH} kubebuilder && sudo mv -f kubebuilder /usr/local/
+RETVAL=$?
+
+if [ $RETVAL -eq 0 ]; then
+  echo "Add kubebuilder to your path; e.g copy paste in your shell and/or ~/.profile:"
+  echo "export PATH=\$PATH:/usr/local/kubebuilder/bin"
+else
+  echo "\n/usr/local/kubebuilder folder is not empty. Please delete or backup it before to install ${KUBEBUILDER_VERSION_NAME}"
+fi
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 #
 # This file will be  fetched as: curl -L https://git.io/getLatestKubebuilder | sh -
@@ -8,13 +8,13 @@
 # It lets users to do curl -L https://git.io//getLatestKubebuilder | KUBEBUILDER_VERSION=1.0.5 sh -
 # for instance to change the version fetched.
 
-OS="$(uname)"
-if [ "x${OS}" = "xDarwin" ] ; then
-  OSEXT="darwin"
-else
-  OSEXT="linux"
-fi
-ARCH=amd64
+# Check if the program is installed, otherwise exit
+function command_exists () {
+  if ! [ -x "$(command -v $1)" ]; then
+    echo "Error: $1 program is not installed." >&2
+    exit 1
+  fi
+}
 
 if [ "x${KUBEBUILDER_VERSION}" = "x" ] ; then
   KUBEBUILDER_VERSION=$(curl -L -s https://api.github.com/repos/kubernetes-sigs/kubebuilder/releases/latest | \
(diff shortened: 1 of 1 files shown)


cmoulliard:
Why do we still have this icla fails ?

cmoulliard:
Can you please review/approve it @droot @pwittrock ?

scripts/install.sh
```diff
@@ -0,0 +1,40 @@
...
+
+if [ "x${KUBEBUILDER_VERSION}" = "x" ] ; then
+  KUBEBUILDER_VERSION=$(curl -L -s https://api.github.com/repos/kubernetes-sigs/kubebuilder/releases/latest | \
+                  grep tag_name | sed "s/ *\"tag_name\": *\"\\(.*\\)\",*/\\1/")
+fi
+
+KUBEBUILDER_VERSION=${KUBEBUILDER_VERSION#"v"}
+KUBEBUILDER_VERSION_NAME="kubebuilder_${KUBEBUILDER_VERSION}"
+URL="https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${KUBEBUILDER_VERSION}/${KUBEBUILDER_VERSION_NAME}_${OSEXT}_${ARCH}.tar.gz"
+echo "Downloading ${KUBEBUILDER_VERSION_NAME} from $URL ..."
```
droot: We should probably use `mktemp` to create a temporary directory and do the download and extraction etc. in the temporary directory and delete it after we have installed kubebuilder. That will make the entire workflow idempotent as well which is important from testability perspective as well ?

scripts/install.sh
```diff
@@ -0,0 +1,40 @@
...
+echo "Downloaded these executable files into $NAME: "
+ls "${KUBEBUILDER_VERSION_NAME}_${OSEXT}_${ARCH}/bin"
+mv ${KUBEBUILDER_VERSION_NAME}_${OSEXT}_${ARCH} kubebuilder && sudo mv -f kubebuilder /usr/local/
+RETVAL=$?
+
+if [ $RETVAL -eq 0 ]; then
+  echo "Add kubebuilder to your path; e.g copy paste in your shell and/or ~/.profile:"
+  echo "export PATH=\$PATH:/usr/local/kubebuilder/bin"
+else
+  echo "\n/usr/local/kubebuilder folder is not empty. Please delete or backup it before to install ${KUBEBUILDER_VERSION_NAME}"
```
droot: we should move this up before we start the downloading etc.

scripts/install.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+#
+# This file will be  fetched as: curl -L https://git.io/getLatestKubebuilder | sh -
+# so it should be pure bourne shell, not bash (and not reference other scripts)
```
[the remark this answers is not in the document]
cmoulliard: Done.

scripts/install.sh
cmoulliard: I agree with you. Will do that

scripts/install.sh
```diff
@@ -0,0 +1,40 @@
...
+# for instance to change the version fetched.
+
+OS="$(uname)"
+if [ "x${OS}" = "xDarwin" ] ; then
+  OSEXT="darwin"
+else
+  OSEXT="linux"
+fi
+ARCH=amd64
+
```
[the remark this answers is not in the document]
cmoulliard: I will include a function to check if the program is installed.

Example 
```
check_program curl
check_program tar

# Check one of the prereq program needed by the script
function check_program () {
  if ! [ -x "$(command -v $1)" ]; then
    echo 'Error: $1 is not installed.' >&2
    exit 1
  fi
}
```

scripts/install.sh
```diff
@@ -0,0 +1,40 @@
...
+# The script fetches the latest kubebuilder release candidate and untars it.
+# It lets users to do curl -L https://git.io//getLatestKubebuilder | KUBEBUILDER_VERSION=1.0.5 sh -
+# for instance to change the version fetched.
+
+OS="$(uname)"
+if [ "x${OS}" = "xDarwin" ] ; then
+  OSEXT="darwin"
+else
+  OSEXT="linux"
+fi
```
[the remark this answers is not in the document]
cmoulliard: I will convert the `if then else` to a 
```
case $OS in
esac
```

scripts/install.sh
cmoulliard: Correct. That makes no sense to download the file if we can't install it ;-)

cmoulliard: See new commit pushed including your remarks @droot

scripts/install.sh
```diff
@@ -0,0 +1,78 @@
...
+# Check if the program is installed, otherwise exit
+function command_exists () {
+  if ! [ -x "$(command -v $1)" ]; then
+    echo "Error: $1 program is not installed." >&2
+    exit 1
+  fi
+}
+
+if [ "x${KUBEBUILDER_VERSION}" = "x" ] ; then
+  KUBEBUILDER_VERSION=$(curl -L -s https://api.github.com/repos/kubernetes-sigs/kubebuilder/releases/latest | \
```
droot: we should move the `curl` check (and other command checks) before this.

scripts/install.sh
```diff
@@ -0,0 +1,78 @@
...
+    OSEXT="darwin"
+    ;;
+  Linux)
+    OSEXT="linux"
+    ;;
+  *)
+    echo "Only OSX and Linux OS are supported !"
+    exit 1
+    ;;
+esac
```
droot: move this at the top ?

scripts/install.sh
```diff
@@ -0,0 +1,78 @@
...
+
+HW=$(uname -m)
+case $HW in
+    x86_64)
+      ARCH=amd64 ;;
+    *)
+      echo "Only x86_64 machines are supported !"
+      exit 1
+      ;;
+esac
```
droot: move this at the top ?

scripts/install.sh
```diff
@@ -0,0 +1,78 @@
...
+echo "Downloading ${KUBEBUILDER_VERSION_NAME}\nfrom $URL\n"
+curl -L "$URL"| tar xz -C $TMP_DIR
+
+echo "Downloaded executable files"
+ls "$TMP_DIR/${KUBEBUILDER_VERSION_NAME}_${OSEXT}_${ARCH}/bin"
+
+echo "Moving files to $KUBEBUILDER_DIR folder\n"
+mv $TMP_DIR/${KUBEBUILDER_VERSION_NAME}_${OSEXT}_${ARCH} $TMP_DIR/kubebuilder && sudo mv -f $TMP_DIR/kubebuilder /usr/local/
+
+rm -rf $TMP_DIR
```
droot: With the above approach, the release archive is not deleted. 

An alternative approach could be: 
```sh
create the temp dir
pushd && cd temp dir
download the archieve
untar archive && rename unarchive-dir kubebuilder && sudo mv -f kubebuilder /usr/local
popd
delete the tmp-dir
```

?

scripts/install.sh
cmoulliard: Sorry for my mistake.

scripts/install.sh
cmoulliard: I agree that OS + HW checks should be the first too before to verify if curl,  tar programs are there or not

scripts/install.sh
cmoulliard: I like also to use pushd/popd. will add it

droot:
Thank you for your patience and addressing all those comments. The script looks good.

PR #514, opened 2018-12-09
.gitignore: Ignore bazel symlinks

justinsb:


.gitignore
```diff
@@ -2,3 +2,10 @@
 
 # Not check in node_modules to a specific arch and os.
 docs/book/node_modules/
+
+# Skip bazel dirs
+bazel-bin
```
mengqiy: Can we simply use `bazel-*`?

justinsb: Good call - I did `/bazel-*` so that we won't accidentally match any files from subdirectories.

PR #515, opened 2018-12-09
Update bazel rules-go/gazelle

justinsb:
Working towards a working bazel build!


mengqiy:
> Working towards a working bazel build!

I assume bazel build doesn't work before, but you make it work now!
I guess it will be useful to add a test in the presubmit to ensure it always work.

Issue #516, opened 2018-12-10
Error when run "kubebuilder create api --group apps --version v1"


answer1991:
**What happened**:
Error when run "kubebuilder create api --group apps --version v1 --kind Foo"

**What you expected to happen**:
No error occur when run "kubebuilder create api --group apps --version v1 --kind Foo"

**How to reproduce it (as minimally and precisely as possible)**:
```bash
$ kubebuilder init --domain test.io --license apache2 --owner "answer1991"
Run `dep ensure` to fetch dependencies (Recommended) [y/n]?
y
dep ensure
Running make...
make
go generate ./pkg/... ./cmd/...
go fmt ./pkg/... ./cmd/...
go vet ./pkg/... ./cmd/...
go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go all
CRD manifests generated under '/Users/Joe/go/src/github.com/answer1991/kubebuilder-test/config/crds'
RBAC manifests generated under '/Users/Joe/go/src/github.com/answer1991/kubebuilder-test/config/rbac'
go test ./pkg/... ./cmd/... -coverprofile cover.out
?   	github.com/answer1991/kubebuilder-test/pkg/apis	[no test files]
?   	github.com/answer1991/kubebuilder-test/pkg/controller	[no test files]
?   	github.com/answer1991/kubebuilder-test/pkg/webhook	[no test files]
?   	github.com/answer1991/kubebuilder-test/cmd/manager	[no test files]
go build -o bin/manager github.com/answer1991/kubebuilder-test/cmd/manager
Next: Define a resource with:
$ kubebuilder create api

$ kubebuilder create api --group apps --version v1 --kind Foo
Create Resource under pkg/apis [y/n]?
y
Create Controller under pkg/controller [y/n]?
y
Writing scaffold for you to edit...
pkg/apis/apps/v1/foo_types.go
pkg/apis/apps/v1/foo_types_test.go
pkg/controller/foo/foo_controller.go
pkg/controller/foo/foo_controller_test.go
Running make...
go generate ./pkg/... ./cmd/...
go fmt ./pkg/... ./cmd/...
go vet ./pkg/... ./cmd/...
# github.com/answer1991/kubebuilder-test/pkg/controller/foo
pkg/controller/foo/foo_controller.go:25:2: appsv1 redeclared as imported package name
	previous declaration at pkg/controller/foo/foo_controller.go:24:2
[...]

PR #517, opened 2018-12-10
fix bug when run "kubebuilder create api --group apps --version v1"

answer1991:
this PR will fix the bug #516

### DIFF
--- a/pkg/scaffold/controller/controller.go
+++ b/pkg/scaffold/controller/controller.go
@@ -107,8 +107,8 @@ import (
 	"log"
 	"reflect"
 
-	appsv1 "k8s.io/api/apps/v1"
-	corev1 "k8s.io/api/core/v1"
+	kubeappsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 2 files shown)


answer1991:
@pwittrock

droot:
@answer1991 We need a way better (and scalable) way to resolve the conflicts in import paths when the user resources conflicts with the core resources. This PR addresses just for the Deployments resource. For your use-case, a possible workaround would be to use `--example false` in the create command.

answer1991:
@droot hi,

Yes, you are right. I was creating apis with “--group apps --version v1” and fix this issue in the most directly way. 

If you have any other plan to fix this issue, please feel free to close my PR.

Otherwise, I would like to continue to work on this issue and enhance my PR.

droot:
@answer1991 

>If you have any other plan to fix this issue, please feel free to close my PR.

Not currently. So for now, you can improve this PR. Pl. note tests are not passing on this PR currently. Check the travis failures.

Issue #518, opened 2018-12-10
The kube-apiserver bundled with kubebuilder should print out a valid version string


roberthbailey:
I was trying to troubleshoot an issue with the version of the kube-apiserver bundled with kubebuilder and wasn't able to see what version is being used. 

```
$ cd kubebuilder_1.0.5_darwin_amd64/bin/
$ ./kube-apiserver --version
Kubernetes v0.0.0-master+$Format:%h$
```

This issue was present when I looked at 1.0.3 and 1.0.5, but may be present in other releases.

droot:
We are fixing this in next release. our current process builds all the binaries from the release-tag. Instead of building the binaries ourselves, we are going to just packages the binaries available on dl.k8s.io. 

/cc @mengqiy who is working on it.

droot:
fixes #518

PR #523, opened 2018-12-11
update how to build tools for kubebuilder

mengqiy:
KB now only need 3 tools for integration testing:
- kube-apiserver
- etcd
- kubectl

For linux, these 3 are fetched from official releases.
For darwin (MacOS), kube-apiserver is built from source, because k8s doesn't release a darwin kube-apiserver; while etcd and kubectl are fetched from official releases.


droot:
fixes #518

Issue #359, opened 2018-08-17
No GOPATH environment variable breaks code generators


mheese:
**Problem**

If one uses the default gopath at `$HOME/go`, and therefore does not set a GOPATH environment variable, the deepcopy-gen will save its output files into the wrong destination path.

This is the output if one just tries to follow the quickstart without a defined GOPATH environment variable:
```
go generate ./pkg/... ./cmd/...
go fmt ./pkg/... ./cmd/...
go vet ./pkg/... ./cmd/...
# ships/pkg/apis/ships/pkg/apis/ships/v1beta1
pkg/apis/ships/pkg/apis/ships/v1beta1/zz_generated.deepcopy.go:27:11: undefined: Sloop
make: *** [Makefile:38: vet] Error 2
```

As one can see the output gets saved to `pkg/apis/ships/pkg/apis/ships`, so the relative output base is wrong.

The deepcopygen command line is defined in the `pkg/apis/apis.go` like this:
```
// Generate deepcopy for apis
//go:generate go run ../../vendor/k8s.io/code-generator/cmd/deepcopy-gen/main.go -O zz_generated.deepcopy -i ./... -h ../../hack/boilerplate.go.txt
```

It is not setting the output base on the command line, which works like this according to the help:
```
-o, --output-base string               Output base; defaults to $GOPATH/src/ or ./ if $GOPATH is not set. (default "./")
```

As there is no GOPATH set, the default is the current working directory, and this is why there is the duplicate `pkg/apis/ships` path in the destination.

**Workaround:**

Simply set your GOPATH environment variable like:
```
export GOPATH="$HOME/go"
```

**Possible Solutions:**
1. simply document the workaround that the GOPATH environment variable must be set
2. maybe be specific in the `go:generate` command line and add a `-o`
3. change the default behaviour for `deepcopy-gen` ? I don't think this is a good idea though, it's pretty clear how it should work

droot:
I like the option 2 that you suggested. Do you mind sharing the value for parameter for `-o` that works for you ?

If you are interested, this is where additional argument can be supplied to fix it. https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/scaffold/manager/apis.go#L41

fraenkel:
So the problem is that there are other go generate stanzas that are outside the control of kubebuiler.
When I tried without a GOPATH, I get the constant failure as shown by the CI on my closed PR.
[...]

PR #542, opened 2018-12-20
:bug: report error if GOPATH env is missing

### LINKED ISSUE #531: `make generate`  creates the deep-copy.go in two different folders
Version: version.Version{KubeBuilderVersion:"1.0.5", KubernetesVendor:"1.11", GitCommit:"f15ca6865b192f9222882365f525d897564e6bc0", BuildDate:"2018-10-16T21:40:14Z", GoOs:"unknown", GoArch:"unknown"}

Running `make generate` from the main folder of the project:
 - `GOPATH/src/testexample`

 Two identical zz_generated.deepcopy.go are generated in the following folders:
 - pkg/apis/testexample/pkg/apis/group/v1beta1/
 - pkg/apis/group/v1beta1/

Then the one placed in pkg/apis/testexample/pkg/apis/group/v1beta1/ create misbehaviour

droot:
The root-cause of the issues we are seeing is deepcopy-gen doesn't support building/generating outside GOPATH. So for now, failing fast when we detect missing GOPATH in `generate` target can save users in this case.

fixes #531 #359

### DIFF
--- a/pkg/scaffold/project/makefile.go
+++ b/pkg/scaffold/project/makefile.go
@@ -89,6 +89,9 @@ vet:
 
 # Generate code
 generate:
+ifndef GOPATH
+	$(error GOPATH not defined, please define GOPATH. Run "go help gopath" to learn more about GOPATH)
+endif
 	go generate ./pkg/... ./cmd/...
 
 # Build the docker image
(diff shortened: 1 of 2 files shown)


DirectXMan12:
long-term we should fix deepcopy-gen to support this case (detecting the default gopath, working for go modules, etc).  Short-term this is fine.

/lgtm

PR #545, opened 2018-12-26
Skip update op when instance is created

xychu:
Also changed one test of reconcile on delete

### DIFF
--- a/pkg/scaffold/controller/controller.go
+++ b/pkg/scaffold/controller/controller.go
@@ -248,23 +248,18 @@ func (r *Reconcile{{ .Resource.Kind }}) Reconcile(request reconcile.Request) (re
 	// TODO(user): Change this for the object type created by your controller
 	// Check if the Deployment already exists
 	found := &appsv1.Deployment{}
-	created := false
 	err = r.Get(context.TODO(), types.NamespacedName{Name: deploy.Name, Namespace: deploy.Namespace}, found)
 	if err != nil && errors.IsNotFound(err) {
 		log.Info("Creating Deployment", "namespace", deploy.Namespace, "name", deploy.Name)
 		err = r.Create(context.TODO(), deploy)
-		if err != nil {
-			return reconcile.Result{}, err
-		} else {
-			created = true
-		}
+		return reconcile.Result{}, err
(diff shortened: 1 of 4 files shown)


pkg/scaffold/controller/controller.go
```diff
@@ -248,20 +248,23 @@ func (r *Reconcile{{ .Resource.Kind }}) Reconcile(request reconcile.Request) (re
...
 	found := &appsv1.Deployment{}
+	created := false
 	err = r.Get(context.TODO(), types.NamespacedName{Name: deploy.Name, Namespace: deploy.Namespace}, found)
 	if err != nil && errors.IsNotFound(err) {
 		log.Info("Creating Deployment", "namespace", deploy.Namespace, "name", deploy.Name)
 		err = r.Create(context.TODO(), deploy)
 		if err != nil {
 			return reconcile.Result{}, err
+		} else {
+			created = true
```
droot: we can probably `return reconcile.Result{}, nil` here and not change anything below ?

xychu: @droot nice catch 👍 , I've updated.

pkg/scaffold/controller/controller.go
```diff
@@ -251,10 +251,7 @@ func (r *Reconcile{{ .Resource.Kind }}) Reconcile(request reconcile.Request) (re
 	err = r.Get(context.TODO(), types.NamespacedName{Name: deploy.Name, Namespace: deploy.Namespace}, found)
 	if err != nil && errors.IsNotFound(err) {
 		log.Info("Creating Deployment", "namespace", deploy.Namespace, "name", deploy.Name)
-		err = r.Create(context.TODO(), deploy)
-		if err != nil {
-			return reconcile.Result{}, err
-		}
+		return reconcile.Result{}, r.Create(context.TODO(), deploy)
```
droot: Though this is more concise, but has poor readability. I had to search where did `Create` go :)

I think the previous version is more readable.
```Go
err = r.Create(....)
return reconcile.Result{}, err
```

xychu: Agreed and thanks.

PR #561, opened 2019-01-09
add all resources validate

runzexia:
Now only some of the resource's validate have been added, which may result in some files being generated and eventually failing.

Signed-off-by: runzexia <runzexia@yunify.com>


mengqiy:
> Now only some of the resource's validate have been added, which may result in some files being generated and eventually failing.

@runzexia Can you please explain a bit? When and how they may eventually fail?

mengqiy:
Thanks for your explanation.
/lgtm
/approve

PR #563, opened 2019-01-11
add redirect for netlify

mengqiy:


DirectXMan12:
ref https://www.netlify.com/docs/redirects/

Issue #571, opened 2019-01-18
Remove race condition in templated test code "StartTestManager" function.


shaneutt:
Hello! When working with `kubebuilder` it was noticed that the templated testing code provided for a new project technically has a race condition in it, specifically it appears to be the `StartTestManager` function in `pkg/scaffold/controller/controllersuitetest.go`.

Here's a diff of my suggestion for a fix for this:

```
diff --git a/pkg/scaffold/controller/controllersuitetest.go b/pkg/scaffold/controller/controllersuitetest.go
index bbdfff7c..37001ca2 100644
--- a/pkg/scaffold/controller/controllersuitetest.go
+++ b/pkg/scaffold/controller/controllersuitetest.go
@@ -96,10 +96,10 @@ func SetupTestReconcile(inner reconcile.Reconciler) (reconcile.Reconciler, chan
 func StartTestManager(mgr manager.Manager, g *gomega.GomegaWithT) (chan struct{}, *sync.WaitGroup) {
        stop := make(chan struct{})
        wg := &sync.WaitGroup{}
+       wg.Add(1)
        go func() {
-               wg.Add(1)
+               defer wg.Done()
                g.Expect(mgr.Start(stop)).NotTo(gomega.HaveOccurred())
-               wg.Done()
        }()
        return stop, wg
 }
```
Here is a PR for this: https://github.com/kubernetes-sigs/kubebuilder/pull/572

Thank you! :+1:

PR #572, opened 2019-01-18
Remove race condition from templated StartTestManager

shaneutt:
This resolves https://github.com/kubernetes-sigs/kubebuilder/issues/571 by removed the race condition from the templated controller test code and putting the `wg.Done()` in a `defer`.


droot:
Thanks for submitting the PR. Change looks good, but following needs to be done:

 - Sign CLA
 - Run `generate_golden.sh` for scaffolding tests to pass

PR #582, opened 2019-01-29
GOPATH is a multi-value environment variable

### LINKED ISSUE #285: GOPATH is a multi-value environment variable
I see you catch an error by this message:

```
2018/06/22 23:28:20 kubebuilder must be run from the project root under $GOPATH/src/package.
```

When `GOPATH=/Users/abd/go`, it works. and when `GOPATH=/Users/abd/v1/go:/Users/abd/v2/go`, it does not work.

I assume you didn't consider the multi-value aspect of `GOPATH`.

Indeed, you can append multiple values to `GOPATH` by separating them through `:`.

luisdavim:
This should fix #285

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -20,11 +20,11 @@ import (
 	gobuild "go/build"
 	"log"
 	"os"
-	"path/filepath"
-	"strings"
+	"regexp"
 
 	"github.com/spf13/cobra"
 
+	toolsutil "sigs.k8s.io/controller-tools/pkg/crd/util"
 	"sigs.k8s.io/kubebuilder/cmd/util"
 	"sigs.k8s.io/kubebuilder/cmd/version"
 )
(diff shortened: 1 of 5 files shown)


cmd/main.go
```diff
@@ -21,10 +21,10 @@ import (
 	"log"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/spf13/cobra"
 
+	toolsUtil "sigs.k8s.io/controller-tools/pkg/crd/util"
```
DirectXMan12: we don't generally use camelCase package names (the general practice is for all-lower-case package names, which makes them a bit easier to distinguish when reading code).

luisdavim: Thanks for the review, I've changed it to `toolsUtil `

DirectXMan12:
looks like you've got an unused variable (`goSrc`).

DirectXMan12:
the OSX job gave output (looks like one of your tests doesn't pass).  Looks like the issue on the linux test came from dep solving for too long.

luisdavim:
I don't know what's wrong with the tests now, they do pas on my laptop. It seems to be some issue with `dep`...

droot:
Travis tests have been super flaky past couple of days. I retriggered and this time it passed.

PR #579, opened 2019-01-25
:running: updated k8s version to 1.13.1 in build scripts

droot:

PR #584, opened 2019-01-31
:running: update k8s dep for test project

mengqiy:
Found this step is something we also need to do besides #578 #579 
Otherwise, our test project is still using the older CR and CT.


mengqiy:
`glog` and `klog` together give me some trouble, but it should be resolved now.

droot:
Good catch. Thanks for fixing it @mengqiy

PR #587, opened 2019-02-01
:warning: simplify webhook scaffolding

mengqiy:
Simplified webhook scaffolding


droot:
New scaffolding definitely looks simpler. I don't have much comments other than are the changes covered under any test ? 

Also travis doesn't seem to be happy.

mengqiy:
> are the changes covered under any test ?

We have scaffolding test to cover them

PR #608, opened 2019-02-21
[WIP]change group validation regex to DNS subdomain

### LINKED ISSUE #298: `kubebuilder create resource --group concourse-ci.org` fails
When I use `concourse-ci.org` as my `--group` param, I see:

`2018/07/02 12:06:21 --group must match regex ^[a-z]+$ but was (concourse-ci.org)`

I would expect this to succeed. The problem appears to be that [the validating regex](https://github.com/kubernetes-sigs/kubebuilder/blob/019fa3c8c8550e817a065cb82058fe790903ed1e/cmd/kubebuilder/create/util/util.go#L59) does not cover all legal domain names, which [may contain numbers and hyphens](https://www.ietf.org/archive/id/draft-liman-tld-names-06.txt).

---
```
$ kubebuilder version
Version: version.Version{KubeBuilderVersion:"0.1.12", KubernetesVendor:"1.10", GitCommit:"a523fb6b174c2785084eb0948c4cfdaf2f818bc4", BuildDate:"2018-06-21T20:07:48Z", GoOs:"unknown", GoArch:"unknown"}
```

M00nF1sh:
Fix #298 
Changes done:
1. change the group validation regex to the one used by kubernetes: [DNS subdomain](https://github.com/kubernetes/kubernetes/blob/6902f3112d98eb6bd0894886ff9cd3fbd03a7f79/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L137)

### DIFF
--- a/pkg/scaffold/resource/resource.go
+++ b/pkg/scaffold/resource/resource.go
@@ -65,9 +65,9 @@ func (r *Resource) Validate() error {
 		r.Resource = rs.Pluralize(strings.ToLower(r.Kind))
 	}
 
-	groupMatch := regexp.MustCompile("^[a-z]+$")
+	groupMatch := regexp.MustCompile("^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$")
 	if !groupMatch.MatchString(r.Group) {
-		return fmt.Errorf("group must match ^[a-z]+$ (was %s)", r.Group)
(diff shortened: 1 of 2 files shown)


M00nF1sh:
Hi guys,
Need some guideline to move this out of WIP.

The current problem is there is an file named as 
`/pkg/apis/$group/group.go`, with golang `package $group`. However, golang package name only allows for [a-z0-9].

1. To support a group of `sub-0.network`, what  should be the package structure? `/pkg/apis/sub0/network` or `/pkg/apis/network/sub0` or `/pkg/apis/sub0`?
2. Should i just extend the regex to include [0-9] without '-' and '.'? (fits my use case by just do this)

DirectXMan12:
if we want to support this, we should think about divorcing package name from group name in the scaffolding (controller-tools supports this now, though).  Closing for now, re-open if you're still interested.

PR #613, opened 2019-02-28
:running: add a design folder and template

DirectXMan12:
This adds a design folder and template that we can use for documenting
major design decisions.


DirectXMan12:
cc @droot

PR #614, opened 2019-02-28
update kustomize scaffolding to work with kustomize 2.0

### LINKED ISSUE #595: make deploy is broken for kustomize v2.0.0 release
kustomize v2.0.0 requires that all paths other than bases have to terminate in same or subdirectory of the kustomization directory. So `kustomize build config/default` in `make deploy` fails because `kustomization.yaml` contains resource paths like `../rbac/rbac_role.yaml`. According to kustomize v2.0.0, `rbac` and `manager` folder should be under `config/default` directory.

Liujingfang1:
Fix #595

### DIFF
--- a/pkg/scaffold/project/kustomize.go
+++ b/pkg/scaffold/project/kustomize.go
@@ -65,21 +65,9 @@ namePrefix: {{.Prefix}}-
 #commonLabels:
 #  someName: someValue
 
-# Each entry in this list must resolve to an existing
-# resource definition in YAML.  These are the resource
-# files that kustomize reads, modifies and emits as a
-# YAML string, with resources separated by document
-# markers ("---").
-resources:
-- ../rbac/rbac_role.yaml
-- ../rbac/rbac_role_binding.yaml
-- ../manager/manager.yaml
-  # Comment the following 3 lines if you want to disable
-  # the auth proxy (https://github.com/brancz/kube-rbac-proxy)
-  # which protects your /metrics endpoint.
-- ../rbac/auth_proxy_service.yaml
-- ../rbac/auth_proxy_role.yaml
-- ../rbac/auth_proxy_role_binding.yaml
+bases:
+- ../rbac
+- ../manager
 
 patches:
 - manager_image_patch.yaml
--- a/cmd/init_project.go
+++ b/cmd/init_project.go
@@ -142,6 +142,8 @@ func (o *projectOptions) runInit() {
 		&manager.Config{Image: imgName},
 		&project.GitIgnore{},
 		&project.Kustomize{},
+		&project.KustomizeRBAC{},
+		&project.KustomizeManager{},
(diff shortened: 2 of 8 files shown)


droot:
@Liujingfang1 Will this new structure work with kustomize 1.0.0 as well ? 

I am thinking of cases where users have projects with current/new kustomization structure, but they have older kustomize (1.0.0) will it work for both ? because upgrading to kustomize 2.0.0 will force them to migrate to newer structure.

Also, was wondering if we can update the issue with set of kustomize commands to run to migrate the old structure to work with new structure.

Liujingfang1:
@droot Yes, this structure works with kustomize 1.0 as well.

Liujingfang1:
> kustomize commands to run to migrate the old structure to work with new structure

There is no commands in Kustomize to migrate the structures. It requires manual work :(

Liujingfang1:
@droot  I added some instructions for migrating from the old structure to the new structure in the issue.

pkg/scaffold/project/kustomize.go
```diff
@@ -65,21 +65,9 @@ namePrefix: {{.Prefix}}-
...
-- ../rbac/rbac_role_binding.yaml
-- ../manager/manager.yaml
-  # Comment the following 3 lines if you want to disable
-  # the auth proxy (https://github.com/brancz/kube-rbac-proxy)
-  # which protects your /metrics endpoint.
-- ../rbac/auth_proxy_service.yaml
-- ../rbac/auth_proxy_role.yaml
-- ../rbac/auth_proxy_role_binding.yaml
+bases:
+- ../rbac
```
droot: I have to double check, but rbac manifests get generated everytime make runs, so not sure if delete all yamls in `rbac` directory.

mengqiy: IIRC it doesn't delete everything, it overwrites the content in `../rbac/rbac_role.yaml`. We'd better confirm it.

droot: I confirmed, it doesn't delete, so the change is good.

droot:
@Liujingfang1 The change is good, but we do not have an e2e test in travis which can verify this. we generally test `kustomize` related changes manually. Will you be able to perform a manual test before the merge ?

Liujingfang1:
@droot Sure, I'll test it manually :)

Liujingfang1:
I ran following commands and all of them succeeded.
```
kubebuilder init --domain jingfang
kubebuilder create api --group my --version v1 --kind MyKind --resource=true --controller=true
make install
make run
kubectl apply -f config/samples/.

# deploy the manager to cluster
make docker-build IMG=jiliu/kb
make docker-push IMG=jiliu/kb
make deploy
```

droot:
Awesome. Thanks for the e2e test.

PR #603, opened 2019-02-14
:sparkles: use builder in controller scaffolding

mengqiy:
Webhook scaffolding can be further simplified after we have #587 and https://github.com/kubernetes-sigs/controller-runtime/pull/328.
They probably should be done in separate PR.

PR #620, opened 2019-03-04
:running: Design for Simplified Scaffolding

DirectXMan12:
This introduces a design for a simplified scaffolding that addresses
both persistent pedagogical issues and dependency injection issues with
our current scaffolding.

### DIFF
--- a/designs/simplified-scaffolding.md
+++ b/designs/simplified-scaffolding.md
@@ -1,0 +1,514 @@
+Simplified Builder-Based Scaffolding
+====================================
+
+## Background
+
+The current scaffolding in kubebuilder produces a directory structure that
+looks something like this (compiled artificacts like config omitted for
+brevity):
+
+<details>
+
+<summary>`tree -d ./test/project`</summary>
+
+```shell
+$ tree -d ./test/project
+./test/project
+├── cmd
+│   └── manager
+├── pkg
+│   ├── apis
+│   │   ├── creatures
+│   │   │   └── v2alpha1
+│   │   ├── crew
+│   │   │   └── v1
+│   │   ├── policy
+│   │   │   └── v1beta1
+│   │   └── ship
+│   │       └── v1beta1
+│   ├── controller
+│   │   ├── firstmate
+│   │   ├── frigate
+│   │   ├── healthcheckpolicy
+│   │   ├── kraken
+│   │   └── namespace
+│   └── webhook
+│       └── default_server
+│           ├── firstmate
+│           │   └── mutating
+│           ├── frigate
+│           │   └── validating
+│           ├── kraken
+│           │   └── validating
+│           └── namespace
+│               └── mutating
+└── vendor
+```
+
+</details>
+
+API packages have a separate file for each API group that creates a SchemeBuilder,
+a separate file to aggregate those scheme builders together, plus files for types,
+and the per-group-version scheme builders as well:
+
+<details>
+
+<summary>`tree ./test/project/pkg/apis`</summary>
+
+```shell
+$ ./test/project/pkg/apis
+├── addtoscheme_creatures_v2alpha1.go
+├── apis.go
+├── creatures
+│   ├── group.go
+│   └── v2alpha1
+│       ├── doc.go
+│       ├── kraken_types.go
+│       ├── kraken_types_test.go
+│       ├── register.go
+│       ├── v2alpha1_suite_test.go
+│       └── zz_generated.deepcopy.go
+...
+```
+
+</details>
+
+Controller packages have a separate file that registers each controller with a global list
+of controllers, a file that provides functionality to register that list with a manager,
+as well as a file that constructs the individual controller itself:
+
+<details>
+
+<summary>`tree ./test/project/pkg/controller`</summary>
+
+```shell
+$ tree ./test/project/pkg/controller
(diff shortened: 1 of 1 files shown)


DirectXMan12:
see also #603.  Holding off on a full-on prototype for a bit, but this looks quite a bit similar to how I structure the controllers when I'm developing or prototyping with CR.

DirectXMan12:
cc @droot @mengqiy

designs/simplified-scaffolding.md
```diff
@@ -0,0 +1,512 @@
...
+```shell
+$ tree ./test/project
+./test/project
+├── main.go
+├── controller
+│   ├── mykind_controller.go
+│   ├── mykind_controller_test.go
+│   └── controllers_suite_test.go
+├── api
+│   └── v1
```
mengqiy: It sounds like this is the proposed structure if we have more than one versions.
```
├── api
│   ├── v1
│   │   └── foo_types.go
│   │   └── groupversion_info.go
│   └── v2
│       └── foo_types.go
│       └── groupversion_info.go
```
It sounds like the info about group will be duplicated across versions.
Given we assume only one group in this structure, does it make sense to have structure like
```
├── api
│   ├── group_info.go
│   ├── v1
│   │   └── foo_types.go
│   │   └── version_info.go
│   └── v2
│       └── foo_types.go
│       └── version_info.go
```

designs/simplified-scaffolding.md
```diff
@@ -0,0 +1,512 @@
...
+
+2. Try to move things and rename references.  This takes a lot more effort
+   on the KubeBuilder maintainers' part if we try to rename references
+   across the codebase.  Not so much if we force the user to, but that's
+   a poorer experience.
+
+3. Tell users to move things, and scaffold out with the new structure.
+   This is fairly messy for the user.
+
+Since growing to multiple API groups seems to be fairly uncommon, it's
```
mengqiy: > fairly uncommon

IIUC you mentioned that you have some data to back this when we chat offline. If you put the number here it's more convincing to make the decision :)

designs/simplified-scaffolding.md
DirectXMan12: good call

designs/simplified-scaffolding.md
DirectXMan12: swear I had this somewhere, but can't seem to find the data now.  I'll see what I can put together

DirectXMan12: hmm... reproducing what I had before is proving difficult (the query I was using before is now returning wildly truncated results), but with a random sampling of smaller projects on github using kubebuilder (presence of manager.yaml) and larger projects that I know of that make use of controller runtime or just CRDs in general (e.g. kubeflow, knative, cluster-api, cert-manager, app-crd, prometheus operator, etcd operator) most seem to follow the pattern of 1 API group per repo.  The main exception was a couple of the knative repositories (but not all of them) and openshift, which puts all of its types in a single repo anyway, so isn't relevant to us.

This makes some amount of sense -- most operators or controllers logically operate on a single set of resources (the CRD defining the operator, a collection of related resources that work together) which would generally be grouped into a single API group.

designs/simplified-scaffolding.md
```diff
@@ -0,0 +1,512 @@
...
+		Owns(&corev1.Pod{}).
+		Complete(r)
+}
+```
+
+This makes it abundantly clear where to start looking at the code
+(`main.go` is the defacto standard entrypoint for many go programs), and
+simplifies the levels of hierarchy.  Furthermore, since `main.go` actually
+instantiates an instance of the reconciler, users are able to add custom
+logic having to do with flags.
```
droot: +1

designs/simplified-scaffolding.md
```diff
@@ -0,0 +1,512 @@
...
+   a nicer UX.
+
+2. Simply adding a new controller, and reminding the user to add the
+   builder themselves.  This is easier for the maintainers, but perhaps
+   a slighly poorer UX for the users.  However, writing out a builder by
+   hand is significantly less complex than adding a controller by hand in
+   the current structure.
+
+Option 1 should be fairly simple, since the logic is already needed for
+registering types to the scheme.
```
droot: In case where user has removed that particular code comment (or we are unable to wire second controller for some reason), we can always fallback to option 2 where we just emit the snippet which user can copy-paste in their main.go.

designs/simplified-scaffolding.md
droot: hmm...another crazy idea :)

What if we removed the `api` and introduced the *group* hierarchy.

```shell
./test/project
  - groupa/v1/....
  - groupb/v2/...
```

DirectXMan12:
> What if we removed the api and introduced the group hierarchy.

hmm... in this case, would controllers go under this hierarchy as well?  I'm not certain this would be clear, depending on the group name -- e.g. with a group name of `apps` or `dependencies`, it's not clear that it's an API group and not something else.

droot:
>hmm... in this case, would controllers go under this hierarchy as well? I'm not certain this would be clear, depending on the group name -- e.g. with a group name of apps or dependencies, it's not clear that it's an API group and not something else.

Yeah, removing `api` and introducing `groups` directly can get confusing.

PR #621, opened 2019-03-05
Improve Dockerfile template such that copying vendor dir is cached.

kaitoy:
The vendor directory is rarely changed during development and is big.
So, moving its copy to lower layer in docker build can make docker builds faster.


droot:
Not much familiar with docker image optimization, but this change looks safe, so approving it. Thanks for submitting the PR.

PR #631, opened 2019-03-12
Fix function comments based on best practices from Effective Go

CodeLingoTeam:
Hi, we updated some exported function comments based on best practices from Effective Go. It’s admittedly a relatively minor fix up. Does this help you?


DirectXMan12:
Yep!  Thanks.  I think that's actually an artifact of one of our linters being disabled -- we normally follow this pattern anyway :-).

DirectXMan12:
/lgtm
/approve

Neat-o

PR #638, opened 2019-03-18
Update documentation for client-go informers

mvladev:
- `manager.RunnableFunc` should block or it'll trigger shutdown.
- `source.Informer`'s `InformerProvider` field was renamed to `Informer`


mvladev:
/kind documetantation 

It seems that the travis build failed some other unrelated reason.

PR #661, opened 2019-04-08
🐛 fix windows releases

### LINKED ISSUE #300: Windows release?
There are releases for both Mac and Linux (appears to be POSIX based). Would it be possible to cross-compile to Windows and have a Windows release as well?

alexeldeib:
Fixes #300. ~~Add windows binaries for kubebuilder~~. Depends on https://github.com/kubernetes-sigs/controller-tools/pull/179.

### DIFF
--- a/pkg/scaffold/project/project.go
+++ b/pkg/scaffold/project/project.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"go/build"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -96,7 +97,7 @@ func (Project) RepoFromGopathAndWd(gopath string, getwd func() (string, error)) 
 			// '$GOPATH/src/kubernetes-sigs/controller-tools' becomes 'kubernetes-sigs/controller-tools'
 			repo := ""
 			for wd != goSrc {
-				repo = filepath.Join(filepath.Base(wd), repo)
+				repo = path.Join(filepath.Base(wd), repo)
 				wd = filepath.Dir(wd)
 			}
 			return repo, nil
--- a/pkg/scaffold/v1/manager/apis.go
+++ b/pkg/scaffold/v1/manager/apis.go
@@ -52,7 +52,7 @@ func (a *APIs) GetInput() (input.Input, error) {
 	}
 	if len(a.Comments) == 0 {
 		a.Comments = append(a.Comments,
-			"// Generate deepcopy for apis", fmt.Sprintf("%s -h %s", deepCopy, b))
+			"// Generate deepcopy for apis", fmt.Sprintf("%s -h %s", deepCopy, filepath.ToSlash(b)))
 	}
 	a.TemplateBody = apisTemplate
 	return a.Input, nil


build/.goreleaser.yml
```diff
@@ -7,16 +7,16 @@ builds:
...
    - amd64
   env:
    - CGO_ENABLED=0
 archive:
   files:
     - "/workspace/_output/{{ .Os }}_{{ .Arch }}/kubebuilder/bin/*"
-  wrap_in_directory: true
 # wrap all the binaries under 'bin/' dir
-  wrap_in_directory_path: "bin"
+  wrap_in_directory: "bin"
```
alexeldeib: When I installed the latest version of goreleaser, using `wrap_in_directory_path` threw an error for me. Can revert if this is a versioning issue, but I had to use this locally.

pkg/scaffold/manager/apis.go
```diff
@@ -52,7 +52,7 @@ func (a *APIs) GetInput() (input.Input, error) {
 	}
 	if len(a.Comments) == 0 {
 		a.Comments = append(a.Comments,
-			"// Generate deepcopy for apis", fmt.Sprintf("%s -h %s", deepCopy, b))
+			"// Generate deepcopy for apis", fmt.Sprintf("%s -h %s", deepCopy, filepath.ToSlash(b)))
```
alexeldeib: Windows will understand forward slashes in generate annotations.

pkg/scaffold/project/project.go
```diff
@@ -87,7 +87,7 @@ func (Project) repoFromGopathAndWd(gopath string, getwd func() (string, error))
 			// '$GOPATH/src/kubernetes-sigs/controller-tools' becomes 'kubernetes-sigs/controller-tools'
 			repo := ""
 			for wd != goSrc {
-				repo = filepath.Join(filepath.Base(wd), repo)
+				repo = filepath.ToSlash(filepath.Join(filepath.Base(wd), repo))
```
DirectXMan12: shouldn't the filepath.ToSlash be on `repo`?

alexeldeib: The final output should be normalized to forward slash to match with Go import statements. On Windows and filepath operation will flip it back.

DirectXMan12: then this should just be `path.Join`, and dispense with the `filepath` stuff entirely (or we should be calling `filepath.Rel` and then `ToSlash` instead of this for loop, but that's a different story).

alexeldeib: Fixed. In the other PR, the initial 3 filepath changes were also the only necessary ones (they compared with OS filepaths). The rest were okay functionally but I assumed the preference was for them to be the same. LMK what you want?

build/.goreleaser.yml
droot: I don't remember the exact change, but we had to modify goreleaser to add this additional field to support packaging multiple binaries under a "bin" directory. So this change will break the build steps in cloud-build, so lets revert this.

alexeldeib: @droot reverted

build/.goreleaser.yml
```diff
@@ -7,6 +7,7 @@ builds:
   goos:
    - darwin
    - linux
+   - windows
```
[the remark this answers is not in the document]
alexeldeib: @droot Makes sense, especially as this still depends on https://github.com/kubernetes-sigs/controller-tools/pull/179. I've removed this for now

Re: dependent binaries, kube-apiserver is tricky. Would you be open to spinning up tests inside a container for Windows (or in general)?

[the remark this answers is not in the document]
droot: I think testing in general has not been explored much, so would be great it if you can elaborate more on this approach either in the issue or the kubebuilder mailing list. Running it inside a container might also enable testing for different versions of kubernetes (which is also one of the concerns I have been hearing lately).

droot:
PR has conflict with `pkg/scaffold/manager/apis.go` and also can you pl. squash all the commits, then we are all set. Thanks.

alexeldeib:
Squashed and rebased. I noticed when I merged there are some 20MB+ .tar.gz artifacts checked in, these will wreaked havoc on git on windows 😄 Will tackle that later, I suppose.

droot:
>Squashed and rebased. I noticed when I merged there are some 20MB+ .tar.gz artifacts checked in, these wreaked havoc for me with git on windows 😄 Will tackle that later, I suppose.

Ouch. Yes, we checked in vendor archive and we are using Git LFS (large file storage) for that. You will have to install the git lfs plugin https://git-lfs.github.com/.

PR #662, opened 2019-04-08
Add issue templates

DirectXMan12:
These templates auto-apply kind labels, and add a comment talking about where to route issues (controller-tools, controller-runtime, or here)


DirectXMan12:
cc @mengqiy @droot

PR #668, opened 2019-04-11
[WIP] New tutorial

DirectXMan12:
Still some spots to clean up, need to actually use the new scaffolding instead of doing it by hand, but wanted stuff up for review.

Need to get gitbook working properly on my machine, too


docs/book/tutorial/page-3.md
```diff
@@ -0,0 +1,736 @@
...
+		// or the active/completed info was somehow lost (contract for status
+		// in kubernetes says it may need to be recreated), or that we have
+		// started a job, but have not noticed it yet (distributed systems can
+		// have arbitrary delays).  In any case, use the creation time of the
+		// CronJob as last known start time.
+		earliestTime = cronJob.ObjectMeta.CreationTimestamp.Time
+	}
+	if cronJob.Spec.StartingDeadlineSeconds != nil {
+		// Controller is not going to schedule anything below this point
+		schedulingDeadline := now.Add(-time.Second * time.Duration(*cronJob.Spec.StartingDeadlineSeconds))
```
mengqiy: > now.Add(-time.Second * time.Duration(*cronJob.Spec.StartingDeadlineSeconds))

It seems it's wrong here. `now.Add(time.Duration(*cronJob.Spec.StartingDeadlineSeconds))` should do the job.

docs/book/tutorial/page-3.md
```diff
@@ -0,0 +1,736 @@
...
+	}
+
+```
+{%% endmethod %%}
+
+{% method %}
+
+If we actually have to run a job, we'll need to either wait till existing ones finish,
+replace the existing ones, or just add new ones.  If our information is out of date due
+to cache delay, we'll get a requeue when we get up-to-date information.
```
mengqiy: > If our information is out of date due to cache delay, we'll get a requeue when we get up-to-date information.

It's not very clear which piece of code is actually dealing with this case. I'd like to discuss about this more in person.

mengqiy: Discussed with Solly offline. He is going to rephrase this piece.

PR #672, opened 2019-04-17
simplify GOPATH

BenTheElder:
`go/build.Default.GOPATH` already has identical defaulting behavior (using `$GOPATH` if set)

https://golang.org/pkg/go/build/#Context

> Default is the default Context for builds. It uses the GOARCH, GOOS, GOROOT, and GOPATH environment variables if set, or else the compiled code's GOARCH, GOOS, and GOROOT.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -30,10 +30,7 @@ import (
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


BenTheElder:
cc @mengqiy

PR #677, opened 2019-04-23
Fix kustomze relative path issue.

yuwenma:
User encounters the following error message when use kubebuilder to
"make deploy" to their cluster:
Error: rawResources failed to read Resources: Load from path <PROJECT>/config/rbac/rbac_role.yaml failed: security; file '<PROJECT>/config/rbac/rbac_role.yaml' is not in or below '<PROJECT>/config/default'

Root cause is filed in kustomize repo kubernetes-sigs/kustomize#766

This is a temp fix aimed at unblocking kubebuilder users.


yuwenma:
/assign @droot 
/assign @DirectXMan12 

/cc @Liujingfang1  
/cc @monopole


Just a quick fix to unblock kubebuilder users. Let me know if there's any easier/better approaches. :)

monopole:
Resources should be shared by using bases - they can be located anywhere.

Patches, however, cannot be shared in v2.0.3.

One can share patches in HEAD:

https://github.com/kubernetes-sigs/kustomize/blob/master/docs/FAQ.md#security-file-foo-is-not-in-or-below-bar

The code is in, and will be in release 2.1, coming in a week or two. Need to write docs and a add some more tests for other stuff.

PR #681, opened 2019-04-30
Add leader election usage section

cizixs:
In practice, controllers are created with multiple replicas,
and it is important to use leader election to avoid unexpected
issues.

This commit adds a section on how to enable leader election.

### DIFF
--- a/docs/book/basics/simple_controller_manager.md
+++ b/docs/book/basics/simple_controller_manager.md
@@ -46,3 +46,34 @@ func main() {
   log.Fatal(mgr.Start(signals.SetupSignalHandler()))}
 ```
 {% endmethod %}
+
+{% method %}
+## Leader Election
+
+Controllers are usually designed to only have one active instance at any time, otherwise unexpected issues might occur. For example:
+
+* kubernetes scheduler will have duplicated resource allocation with multiple instance active
+* deployment controller will have unexpected number of pods running with multiple instances active
+
+In practice, it is suggested to run controller in active-backup mode to achieve high availability. Leader Election is how we do this.
+
+If three replicas of a certain controller pod is running, only the one wins leader election will run, and the other two pods is idle.
+When leader pod is unavailable for some reason(network disconnectivity, server failure etc), the other two pods will start another
+leader election round, and whoever becomes leader can continue running.
+
+This example enables leader election, with two other parameters:
+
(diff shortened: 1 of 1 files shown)


cizixs:
@DirectXMan12 This pr adds doc on how to enable leader election. Please take a look.

docs/book/basics/simple_controller_manager.md
```diff
@@ -46,3 +46,34 @@ func main() {
   log.Fatal(mgr.Start(signals.SetupSignalHandler()))}
 ```
 {% endmethod %}
+
+{% method %}
+## Leader Election
```
droot: I would suggest moving this under `beyond the basics/controllers/` right after `Using finalizers`

docs/book/basics/simple_controller_manager.md
```diff
@@ -46,3 +46,34 @@ func main() {
...
+
+If three replicas of a certain controller pod is running, only the one wins leader election will run, and the other two pods is idle.
+When leader pod is unavailable for some reason(network disconnectivity, server failure etc), the other two pods will start another
+leader election round, and whoever becomes leader can continue running.
+
+This example enables leader election, with two other parameters:
+
+* LeaderElectionNamespace: leaderelection need to create configmap underhood, this field indicates which namespace should be used to create the configmap. If not provided, namespace the controller pod is running in will be used
+* LeaderElectionID: a unique id for leaderelection. The default value is a constant, so it's better provided with a unique value.
+
```
droot: This will also require granting RBAC permissions for CRUD ops on configmaps on this namespace, so we should add a *NOTE* for that.

cizixs: Will kubebuidler create RBAC if leader election is enabled? Or users have to explicitly specify it above `Reconcile` methods as usual?

cizixs:
It seems kubebuilder is moving to v2, including docs/book. Is v1 still maintained? Or should I just abandon this PR?

@droot @DirectXMan12

DirectXMan12:
v1 will receive bugfixes, but the focus will be on the v2 book.

DirectXMan12:
We'll be building up the v2 book a bit more as time rolls on

droot:
@cizixs This doc can be added as a user guide or in a reference section in the new layout in v2 book. Since PR needs to be completely reworked, I will close this for now.

PR #689, opened 2019-05-09
🏃 task: enable running tests in prow

alexeldeib:
Update common.sh to fetch missing dependencies when SKIP_FETCH_TOOLS is false.

We could also build base images with the artifacts (dep, kubebuilder, etcd, api-server, v1 and v2 project tarballs) inside them. I'm not familiar with the (any?) process for publishing official images in Kubernetes projects, and we can always iterate, so I'm offering this up as an initial cut

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -116,7 +116,11 @@ function fetch_tools {
   if [ -n "$SKIP_FETCH_TOOLS" ]; then
     return 0
   fi
+  fetch_go_tools
+  fetch_kb_tools
+}
 
+function fetch_kb_tools {
   header_text "fetching tools"
   kb_tools_archive_name="kubebuilder-tools-$k8s_version-$goos-$goarch.tar.gz"
   kb_tools_download_url="https://storage.googleapis.com/kubebuilder-tools/$kb_tools_archive_name"
@@ -138,7 +142,32 @@ function build_kb {
     opts=-ldflags "-X sigs.k8s.io/kubebuilder/cmd/version.kubeBuilderVersion=$INJECT_KB_VERSION"
   fi
 
-  go build $opts -o $tmp_root/kubebuilder/bin/kubebuilder ./cmd
+  GO111MODULE=on go build $opts -o $tmp_root/kubebuilder/bin/kubebuilder ./cmd
+}
+
+function fetch_go_tools {
+  header_text "Checking for dep"
+  export PATH=$(go env GOPATH)/src/github.com/golang/dep/bin:$PATH
+  if ! is_installed dep ; then
+    header_text "Installing dep"
+    DEP_DIR=$(go env GOPATH)/src/github.com/golang/dep
+    mkdir -p $DEP_DIR
+    pushd $DEP_DIR
+    git clone https://github.com/golang/dep.git .
+    DEP_LATEST=$(git describe --abbrev=0 --tags)
+    git checkout $DEP_LATEST
+    mkdir bin
+    go build -ldflags="-X main.version=$DEP_LATEST" -o bin/dep ./cmd/dep
+    popd
+  fi
+
+}
+
+function is_installed {
+  if command -v $1 &>/dev/null; then
+    return 0
+  fi
+  return 1
 }
 
 function prepare_testdir_under_gopath {
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,6 @@ services:
 before_install:
 # NOTE: The latest version of dep requires go 1.13
 - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then mkdir -p /Users/travis/gopath/bin; fi
-- GO111MODULE=on curl https://raw.githubusercontent.com/golang/dep/v0.5.1/install.sh | sh
 - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then GO111MODULE=on scripts/install_and_setup.sh; fi
 
 before_script:
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -77,6 +77,12 @@ curl -L "$URL"| tar xz -C $TMP_DIR
 echo "Downloaded executable files"
 ls "${KUBEBUILDER_VERSION_NAME}_${OSEXT}_${ARCH}/bin"
 
+if [[ -z "${PROW}" ]]; then 
+  MOVE="sudo mv"
+else
+  MOVE="mv"
+fi
+
 echo "Moving files to $KUBEBUILDER_DIR folder\n"
 mv ${KUBEBUILDER_VERSION_NAME}_${OSEXT}_${ARCH} kubebuilder && sudo mv -f kubebuilder /usr/local/
 
--- a/test_ci.sh
+++ b/test_ci.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+# Copyright 2018 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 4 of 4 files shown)


common.sh
```diff
@@ -141,6 +147,28 @@ function build_kb {
...
+function is_installed {
+  if [ command -v $1 &>/dev/null ]; then
+    return 0
+  fi
+  return 1
+}
+
+function fetch_git_tools {
+    header_text "Installing git lfs"
+    curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash
```
alexeldeib: `curl | bash` ☹️ don't enjoy this

DirectXMan12: let's bake it in to an image then

alexeldeib: yeah, I honestly didn't know where to put it but that was the plan. I'll make the image and we can figure out where to put it.

alexeldeib: Going to leave the git lfs artifact download here if that's okay. removing the install though.

alexeldeib: I could also make a separate script in `scripts/` which is basically:
```
CI=true # don't use sudo in scripts"
git lfs pull <repo>
,/install-test.sh
```

alexeldeib: (To invoke in prow)

DirectXMan12:
FYI, didn't mean to drop this -- just got busy with kubecon stuff.  Will take a look at it shortly.

alexeldeib:
No worries -- your Kubecon workshop materials are awesome btw 😄 Wasn't at the conference but have been using them a bunch at home.

I think I synced up the stuff on the Prow side. I dropped the Dockerfile in a gist here (per a now fairly old slack comment): https://gist.github.com/alexeldeib/33f63d719fd163b78d811857f34bc90f

IMO, main question is where to store the image artifact and where to handle the git lfs artifact pulls.

DirectXMan12:
> your Kubecon workshop materials are awesome btw
:-D glad you liked it.

For the image stuff, we've got a kubebuilder gcr account that we can host it on.  We can set up cloud-builds to build it automatically if we check it into the repo.

It's what we do for a few other things.

DirectXMan12:
Assume that it needs to be at `gcr.io/kubebuilder/<something>`

alexeldeib:
So I started scaffolding it out, I'm not super familiar with cloudbuild but seems easy enough. The app for testing it locally seems to 404, I did find https://github.com/GoogleCloudPlatform/cloud-build-local but having issues installing it (gcloud components install isnt't finding the component for me) so haven't been able to test it myself yet.

build/cloudbuild_prow.yaml
```diff
@@ -0,0 +1,18 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+steps:
+- name: gcr.io/cloud-builders/docker
+  args: ['build', '-t', 'gcr.io/kubebuilder/prow-base', '.']
+images: ['gcr.io/kubebuilder/prow-base']
```
DirectXMan12: oh, heh.  If this is all we're doing, we can tell cloudbuild to do this directly.

DirectXMan12: this is fine, though

DirectXMan12: tweak the hardcoded `kubebuilder` to be `$PROJECT_ID` and this should be set.

DirectXMan12: (should've mentioned that earlier -- my fault)

DirectXMan12: I've pushed that version for you to test with -- we'll set up an autobuild later.

alexeldeib:
/remove-lifecycle stale

I've been a bit strapped on capacity, but i'll be able to circle back to this shortly if @alvaroaleman doesn't beat me to it

alexeldeib:
err actually, I think this one might be really stale, and the test-infra one is the one we want (?) will check

alexeldeib:
ref https://github.com/kubernetes/test-infra/pull/12557

alexeldeib:
@DirectXMan12 finally tackled this. quick summary:

- used golang and kubekins instead of custom image. 
  - We can iterate on this but in the interest of expediency I just fixed the installation (and it's quite fast, so I'm not that worried about dependency caching)
  - installed git-lfs from source to avoid `curl | bash`
- in test-infra PR, changed the job config to match what @alvaroaleman did for controller-runtime (thanks for the comments on test-infra PR -- I was matching what other repos had done, and didn't realize some of those didn't match up with the desired defaults in KB repos). 
 
I validated the `test.sh` job with phaino locally, seems okay. 

I think I also got the configuration for the `test_e2e.sh` but I' haven't been able to run this locally easily using phaino because it does some fun stuff like set up kind and dind to run tests. I can either try to get that working locally, or add it as optional in the test-infra PR and iterate? thoughts?

alexeldeib:
Updated test-infra PR: https://github.com/kubernetes/test-infra/pull/14692

common.sh
```diff
@@ -138,7 +144,43 @@ function build_kb {
...
+  if ! is_installed git-lfs; then
+    header_text "Installing git-lfs"
+    GO111MODULE=on go get -u github.com/git-lfs/git-lfs
+    git lfs install
+  fi
+}
+
+function fetch_git_artifacts {
+  header_text "Fetching git artifactss"
+  git lfs pull https://github.com/kubernetes-sigs/kubebuilder.git # git lfs inside docker with ssh seems to strongly dislike github.com host keys..
```
DirectXMan12: lfs should be disabled at this point, I think, so I *think* we can drop this.

alexeldeib:
added test_ci.sh and re-resquashed

alexeldeib:
made travis unhappy -_-

build/cloudbuild_prow.yaml
alexeldeib: Guess we didn't end up needing this 🙂 We can set up a nice image with our deps + auto-bumping if we want, but seems like we're okay for now?

alexeldeib:
grr...struggling to figure out why travis keeps double installing dep. once that's done, hopefully this can finally merge

alexeldeib:
@DirectXMan12 ptal, think that should fix the bash fire. Proactively squashed again.

DirectXMan12:
this last one looks good

/lgtm
/approve
/hold

(hold since this is being used for testing prow)

alexeldeib:
same issue as travis, doesn't like the dep installation and my bash conditionals. tweaking..

common.sh
```diff
@@ -138,7 +142,32 @@ function build_kb {
...
+    header_text "Installing dep"
+    DEP_DIR=$(go env GOPATH)/src/github.com/golang/dep
+    mkdir -p $DEP_DIR
+    pushd $DEP_DIR
+    git clone https://github.com/golang/dep.git .
+    DEP_LATEST=$(git describe --abbrev=0 --tags)
+    git checkout $DEP_LATEST
+    mkdir bin
+    go build -ldflags="-X main.version=$DEP_LATEST" -o bin/dep ./cmd/dep
+    export PATH=$(go env GOPATH)/src/github.com/golang/dep/bin:$PATH
```
alexeldeib: this is what I changed, sorry I squashed before letting you review. `go install` seemed to be a bit finnicky between prow/travis, and with the changes to GOPATH later I figured I should force it to a static location.

DirectXMan12: yeah, that's fine

DirectXMan12:
/lgtm

restarted travis, prob a flake

alexeldeib:
think we're all set here 😃 

/hold cancel

PR #694, opened 2019-05-13
Fix `kubebuilder` spelling

yashbhutwala:


DirectXMan12:
This is in the legacy book version, so needs to be submitted against that branch now

PR #698, opened 2019-05-15
:running: Fix netlify redirects for v1 book

DirectXMan12:
This fixes the redirects for the book-v1 branch so that they properly
evaluate the alias v1.book --> book-v1.book.


DirectXMan12:
manual merge as per previous LGTM to get the book's v1 link working properly.

PR #702, opened 2019-05-15
:sparkles: fix webhook related scaffolding

mengqiy:
- remove dead code in `generated_golden.sh`
- fix broken webhook related scaffolding due to name or path change.
- split `NAMESPACE` into 2 different kustomize config, so `webhook` base can be used w/o `certmanager`.

- use `distroless/static`

### DIFF
--- a/pkg/scaffold/v2/certmanager/kustomize.go
+++ b/pkg/scaffold/v2/certmanager/kustomize.go
@@ -39,6 +39,7 @@ func (p *Kustomization) GetInput() (input.Input, error) {
 var kustomizationTemplate = `resources:
 - certificate.yaml
 
+# the following config is for teaching kustomize how to do var substitution
 vars:
 - name: CERTIFICATENAME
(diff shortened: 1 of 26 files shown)


pkg/scaffold/v2/certmanager/kustomize.go
```diff
@@ -39,13 +39,22 @@ func (p *Kustomization) GetInput() (input.Input, error) {
...
 
+# the following config is for teaching kustomize how to do var substitution
 vars:
 - name: CERTIFICATENAME
   objref:
     kind: Certificate
     group: certmanager.k8s.io
     version: v1alpha1
     name: serving-cert # this name should match the one in certificate.yaml
+- name: CERTIFICATENAMESPACE
```
droot: CERTIFICATE_NAMESPACE ? to improve readability
It will also be good to mention why do we need this to be var ?

PR #726, opened 2019-05-20
Update cronjob_controller.go

ilansh:
Fix ignoreNotFound function.  
Correct me if I'm wrong, but I Believe the request will be requeued if `err != nil`, so nil should be returned on NotFound error


mengqiy:
@ilansh Please sign the CLA if you haven't do it.

ilansh:
@mengqiy I have now. hope it went through

PR #682, opened 2019-05-03
:warning: controller-tools v0.2.0 scaffolding update

DirectXMan12:
This updates the v2 scaffolding to support the upcoming v0.2.0 controller-tools branch.  It disables webhook-related things, since v2 scaffolding doesn't scaffold webhook Go code yet and it was breaking stuff.

It also assumes controller-gen is packaged as a binary.  Since we've got Go modules, we can just `GO111MODULE=on go get sigs.k8s.io/controller-tools/cmd/controller-gen@<temporary-branch-for-now>`, which is what this does.

Still waiting on:
- [x] #644 to merge
- [ ] controller-tools v0.2.0
- [ ] controller-runtime with modules to merge
- [x] add makefile rule to fetch controller-tools ?

PR #727, opened 2019-05-21
:sparkles: use Deployment for v2 scaffolding

mengqiy:
- 1st commit:
v2 scaffolding: deployment w/ leader election (add necessary marks for rbac and change Makefile)
v1 scaffolding (no changes): statefulSet w/o leader election
- 2nd commit:
fix a singular|plural noun bug introduced in #682 that causes it to not be able to locate a patch.
- 3rd commit:
improve webhook scaffolding by re-org kustomize config and variable and better comments
 
close https://github.com/kubernetes-sigs/kubebuilder/issues/535


mengqiy:
This is ready for review.

pkg/scaffold/v2/main.go
```diff
@@ -79,7 +79,15 @@ func main() {
...
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{ Scheme: scheme, MetricsBindAddress: metricsAddr })
+	// Persmissions to do leader election.
+	// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
+	// +kubebuilder:rbac:groups="",resources=configmaps/status,verbs=get;update;patch
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme:                  scheme,
+		MetricsBindAddress:      metricsAddr,
+		LeaderElection:          true,
+		LeaderElectionNamespace: "default",
```
droot: "default" namespace. Can we parameterize this and supply the namespace where deployment itself is deployed. I think we can supply using downward API ?

pkg/scaffold/v2/crd/enablewebhook_patch.go
```diff
@@ -68,7 +66,7 @@ spec:
       # but we're going to set it later using the cert-manager (or potentially a patch if not using cert-manager)
       caBundle: Cg==
       service:
-        namespace: $(NAMESPACE)
+        namespace: system
         name: webhook-service
         path: /convert-{{ lower .Resource.Kind }}
```
droot: I think we enable conversion webhook at the default path `/convert`, so we need to take the kind out.

pkg/scaffold/v2/makefile.go
```diff
@@ -73,7 +73,7 @@ deploy: manifests
 
 # Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
-	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./api/...;./controllers/..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```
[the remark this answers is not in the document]
mengqiy: I didn't find a good way to make it work with `main` package.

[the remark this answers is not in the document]
mengqiy: What is your concern using `./...`
I can tweak this piece a bit more.

pkg/scaffold/v2/main.go
```diff
@@ -79,7 +79,15 @@ func main() {
 
 	ctrl.SetLogger(zap.Logger(true))
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{ Scheme: scheme, MetricsBindAddress: metricsAddr })
+	// Persmissions to do leader election.
+	// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
+	// +kubebuilder:rbac:groups="",resources=configmaps/status,verbs=get;update;patch
```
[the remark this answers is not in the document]
mengqiy: > I remember some conversation about using a resource for leader election instead of configmaps

It is available on newer cluster (I will need to check from what version). We are still using configmap in our CR library. Ref: https://github.com/kubernetes-sigs/controller-runtime/blob/6fc7baa65f8966f45353903c0e886fe23a21af73/pkg/leaderelection/leader_election.go#L82

pkg/scaffold/v2/main.go
mengqiy: I believe `DownwardAPI` will work here.

pkg/scaffold/v2/crd/enablewebhook_patch.go
mengqiy: Good catch.

pkg/scaffold/v2/makefile.go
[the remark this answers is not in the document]
droot: Not a major concern, just wanted to limit the scanning to the relevant code.

pkg/scaffold/v2/main.go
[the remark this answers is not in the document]
droot: We can probably switch to resource based leader-election in the next release of controller-runtime.

pkg/scaffold/v2/main.go
```diff
@@ -72,14 +72,28 @@ func init() {
 	// %s
 }
 
+// Persmissions to do leader election.
```
DirectXMan12: nit: spelling

pkg/scaffold/v2/main.go
```diff
@@ -72,14 +72,28 @@ func init() {
...
 func main() {
 	var metricsAddr string
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.Parse()
 
 	ctrl.SetLogger(zap.Logger(true))
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{ Scheme: scheme, MetricsBindAddress: metricsAddr })
+	leaderElectionNamespace := "default"
+	if len(os.Getenv("POD_NAMESPACE")) != 0 {
```
DirectXMan12: IIRC we do this automatically when leader election is enabled -- no need to re-do it here.

pkg/scaffold/v2/manager/config.go
```diff
@@ -88,6 +86,11 @@ spec:
       containers:
       - command:
         - /manager
+        env:
```
DirectXMan12: shouldn't be necessary -- the leader election code determines this automatically, IIRC

pkg/scaffold/v2/main.go
```diff
@@ -72,14 +72,28 @@ func init() {
 	// %s
 }
 
+// Persmissions to do leader election.
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=configmaps/status,verbs=get;update;patch
```
DirectXMan12: we should scaffold this as a separate static role -- it's always the same, and shouldn't be something the user is concerned about.

pkg/scaffold/v2/main.go
mengqiy: LE in CR current default ns to `default` and configmap name to `controller-leader-election-helper`.
Ref: https://github.com/kubernetes-sigs/controller-runtime/issues/445

DirectXMan12:
FWIW, I'm not certain we actually need to switch (what's the actual benefit here?).  Turning on leader election is still useful though.

mengqiy:
> FWIW, I'm not certain we actually need to switch (what's the actual benefit here?). Turning on leader election is still useful though.

For the most common case (stateless controller), people don't get more benefits by using StatefulSet instead of Deployment when leader election is enabled. And using StatefulSet actually confuses the users, I saw multiple users confused by this before.
For the uncommon case (stateful controller), we document it in the book how to use StatefulSet.

pkg/scaffold/v2/main.go
mengqiy: After thinking it more, we realize we should set different [`LeaderElectionID`](https://github.com/kubernetes-sigs/controller-runtime/blob/687d7fc8ef4fa76256188c9046afe8bbf30acbf8/pkg/manager/manager.go#L116:2) here. Probably something like `<kind>-controller-leader-election-helper`.
So users can deploy 2 controllers in the same namespace without leader election object conflicts.

DirectXMan12: it uses the current namespace: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/leaderelection/leader_election.go#L63

pkg/scaffold/v2/main.go
mengqiy: Done

mengqiy:
Addressed comments and rebased to resolve conflicts.
PTAL

pkg/scaffold/v2/staticrole.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	r.TemplateBody = staticRoleTemplate
+	return r.Input, nil
+}
+
+var staticRoleTemplate = `# permissions to do leader election.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: static-role
```
droot: Lets rename it `leader-election-role`

pkg/scaffold/v2/manager/config.go
```diff
@@ -76,9 +76,7 @@ spec:
     matchLabels:
       control-plane: controller-manager
       controller-tools.k8s.io: "1.0"
-  serviceName: controller-manager-service
-  replicas: 1
-  podManagementPolicy: Parallel
+  replicas: 2
```
droot: Can we keep it 1 to limit the resource usage.

pkg/scaffold/v2/staticrole.go
```diff
@@ -0,0 +1,67 @@
...
+
+// StaticRole scaffolds the config/rbac/auth_proxy_role.yaml file
+type StaticRole struct {
+	input.Input
+}
+
+// GetInput implements input.File
+func (r *StaticRole) GetInput() (input.Input, error) {
+	if r.Path == "" {
+		r.Path = filepath.Join("config", "rbac", "static_role.yaml")
```
droot: we should rename the file to `leader_election_role.yaml` ?

pkg/scaffold/v2/manager/config.go
```diff
@@ -76,9 +76,7 @@ spec:
     matchLabels:
       control-plane: controller-manager
       controller-tools.k8s.io: "1.0"
```
droot: Can we remove this annotation from all manifests ?

mengqiy: > controller-tools.k8s.io: "1.0"

We can definitely remove these labels. But this may end up causing we duplicate almost all of the shared files, because most manifests files have that label. So I'd like to do it in a followup PR, since it's a cleanup.

mengqiy:
I will squash commits before merging. So putting a hold label for now.
/hold

pkg/scaffold/v2/manager/config.go
mengqiy: Filed #764 to track

mengqiy:
Squashed commits
/hold cancel
No code changes. Re-applying label per https://github.com/kubernetes-sigs/kubebuilder/pull/727#issuecomment-499292440

PR #728, opened 2019-05-21
Fix path prefixes in cmd help

### LINKED ISSUE #717: Create api command references removed pkg directory in v2
The `create api` command asks users if they would like to create resources and controllers under the `pkg` directory, which seems to have been removed in the v2 structure.  Not that big of a deal, but it was confusing.

```
$ kubebuilder create api --version v1alpha1 --group foo --kind Bar
Create Resource under pkg/apis [y/n]?
y
Create Controller under pkg/controller [y/n]?
y
api/v1alpha1/bar_types.go
controllers/bar_controller.go
```

/kind bug

nirnanaaa:
This should fix #717 There were also some more spelling mistakes for controller vs. controllers

### DIFF
--- a/cmd/api.go
+++ b/cmd/api.go
@@ -69,12 +69,12 @@ func (o *apiOptions) runAddAPI() {
 
 	reader := bufio.NewReader(os.Stdin)
 	if !o.resourceFlag.Changed {
-		fmt.Println("Create Resource under pkg/apis [y/n]?")
+		fmt.Println("Create Resource under api [y/n]?")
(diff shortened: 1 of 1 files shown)


abursavich:
IIRC, v1 and v2 are both supported by the cli and they write to different directories.

nirnanaaa:
@abursavich you mean project-version on init? Let me see if i can fix that. Thanks.

nirnanaaa:
before doing: Is that something we wanted to fix, or is it that v1 get's decomissioned?

PR #732, opened 2019-05-22
Add version dependent questions for generating APIs

nirnanaaa:
As a follow up to #728 here is a simple approach to version specific questions. Although I am not sure if this is the right way to approach this, it really seems more complex to add a helper for this. @abursavich what do you think about the `Examples`? I haven't found a proper way to make those version specific.

### DIFF
--- a/cmd/api.go
+++ b/cmd/api.go
@@ -28,7 +28,6 @@ import (
 
 	"sigs.k8s.io/kubebuilder/cmd/util"
 	"sigs.k8s.io/kubebuilder/pkg/scaffold"
-	"sigs.k8s.io/kubebuilder/pkg/scaffold/project"
 	"sigs.k8s.io/kubebuilder/pkg/scaffold/v1/resource"
 )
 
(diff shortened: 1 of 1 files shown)


abursavich:
Consider having a default in case there's a v3?
```
dir := "api"
if projectInfo.Version == project.Version1 {
	dir = "pkg/apis"
}
fmt.Printf("Create Resource under %s [y/n]\n", dir)
```

cmd/api.go
```diff
@@ -66,15 +67,28 @@ func resourceForFlags(f *flag.FlagSet) *resource.Resource {
...
 	}
 
 	if !o.controllerFlag.Changed {
-		fmt.Println("Create Controller under controllers [y/n]?")
+		if projectInfo.Version == project.Version1 {
+			fmt.Println("Create Controller under pkg/controller [y/n]")
+		}
+		if projectInfo.Version == project.Version2 {
+			fmt.Println("Create Controller under controllers [y/n]")
+		}
```
droot: I think we can take out file-path (pkg/apis or api, controllers or pkg/controller) from these prompts because we do print-out where resource/controllers source code is being scaffolded.

nirnanaaa: how would we go ahead with the examples section?

droot: yes, we should update those to keep it in sync.

PR #734, opened 2019-05-22
:bugfix: fixed path-to-crd-dir in controller test suite

### LINKED ISSUE #733: Generated controllers/suite_test.go uses wrong CRD path
`controllers/suite_test.go` bootstraps the test environment as follows:

```go
testEnv = &envtest.Environment{
	CRDDirectoryPaths: []string{filepath.Join("..", "..", "config", "crd", "bases")},
}
```

This is incorrect - there should be only a single ".." since `controllers` is only one level deep (unlike `api/v1alpha1`, which is the other place where `suite_test.go` is generated). This bug means that the generated client fails to recognize any custom resources.

droot:
fixes #733


droot:
fixes https://github.com/kubernetes-sigs/kubebuilder/issues/733

PR #738, opened 2019-05-24
Removing unrequired package from Dep

### LINKED ISSUE #737: controller-tools seems unrequired
/kind bug

I know that `kubebuilder` moved (or is moving) to `gomod`, however the `dep` installation is claiming a missing requirement for `controller-tools`.

```
Warning: the following project(s) have [[constraint]] stanzas in Gopkg.toml:

  ✗  sigs.k8s.io/controller-tools

However, these projects are not direct dependencies of the current project:
they are not imported in any .go files, nor are they in the 'required' list in
Gopkg.toml. Dep only applies [[constraint]] rules to direct dependencies, so
these rules will have no effect.

Either import/require packages from these projects so that they become direct
dependencies, or convert each [[constraint]] to an [[override]] to enforce rules
on these projects, if they happen to be transitive dependencies.
```

In fact, I noticed that is not imported in any source code, except for the template scaffolding of `Makefile`, so I guess that it can be totally removed.

prometherion:
Closes #737.


DirectXMan12:
We can just drop dep entirely here, since we build KB itself with modules now, and it's not designed to be vendored.  Can you just convert this to drop `Gopkg.*`

prometherion:
@DirectXMan12 sure, just pushed f00bb87

prometherion:
Already merged with [749](https://github.com/kubernetes-sigs/kubebuilder/pull/749)

PR #744, opened 2019-05-29
:bug: fix missing add-to-scheme calls on new API version

### LINKED ISSUE #740: main.go should be updated when creating another CRD with existing Group Kind
Steps to reproduce:
- kubebuilder init --project-version 2 --domain testproject.org --license apache2 --owner "The Kubernetes authors"
- kubebuilder create api --group crew --version v1 --kind Captain --controller=true --resource=true --make=false
- kubebuilder create api --group crew --version v2 --kind Captain --controller=false --resource=true --make=false

We get
```
import (
       ...
	crewv1 "sigs.k8s.io/play-kubebuilder/api/v1"
	// +kubebuilder:scaffold:imports
)
...
func init() {

	crewv1.AddToScheme(scheme)
	// +kubebuilder:scaffold:scheme
}
```

`crewv2` should be there as well.

/kind bug

droot:
The change improves the logic for updating the main.go when
new API/Controller is added to the project.

fixes #740

### DIFF
--- a/pkg/scaffold/v2/internal/string_utils.go
+++ b/pkg/scaffold/v2/internal/string_utils.go
@@ -18,7 +18,6 @@ package internal
 import (
 	"bufio"
 	"bytes"
-	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
@@ -29,40 +28,37 @@ import (
 )
 
 // insertStrings reads content from given reader and insert string below the
-// line line containing marker string. So for ex. in insertStrings(r, m1, v1, m2, v2)
+// line containing marker string. So for ex. in insertStrings(r, {'m1':
+// [v1], 'm2': [v2]})
 // v1 will be inserted below the lines containing m1 string and v2 will be inserted
 // below line containing m2 string.
-func insertStrings(r io.Reader, markerAndValues ...string) (io.Reader, error) {
-	if len(markerAndValues)%2 != 0 {
-		return nil, fmt.Errorf("invalid marker and value pairs")
-	}
-
-	mvPairs := map[string]string{}
-	for i, s := range markerAndValues {
-		if i%2 == 0 {
-			mvPairs[s] = markerAndValues[i+1]
-		}
-	}
-
+func insertStrings(r io.Reader, markerAndValues map[string][]string) (io.Reader, error) {
+	// reader clone is needed since we will be reading twice from the given reader
 	buf := new(bytes.Buffer)
+	rClone := io.TeeReader(r, buf)
 
-	scanner := bufio.NewScanner(r)
+	err := filterExistingValues(rClone, markerAndValues)
+	if err != nil {
+		return nil, err
+	}
+
+	out := new(bytes.Buffer)
+
+	scanner := bufio.NewScanner(buf)
 	for scanner.Scan() {
 		line := scanner.Text()
 
-		for m, v := range mvPairs {
-			if strings.TrimSpace(line) == strings.TrimSpace(v) {
-				// since value already exist, so avoid duplication
-				delete(mvPairs, m)
-			}
-			if strings.Contains(line, m) {
-				_, err := buf.WriteString(v)
-				if err != nil {
-					return nil, err
+		for marker, vals := range markerAndValues {
+			if strings.TrimSpace(line) == strings.TrimSpace(marker) {
+				for _, val := range vals {
+					_, err := out.WriteString(val)
+					if err != nil {
+						return nil, err
+					}
 				}
 			}
 		}
-		_, err := buf.WriteString(line + "\n")
+		_, err := out.WriteString(line + "\n")
 		if err != nil {
 			return nil, err
 		}
@@ -70,10 +66,10 @@ func insertStrings(r io.Reader, markerAndValues ...string) (io.Reader, error) {
 	if err := scanner.Err(); err != nil {
 		return nil, err
 	}
-	return buf, nil
+	return out, nil
 }
 
-func InsertStringsInFile(path string, markerAndValues ...string) error {
+func InsertStringsInFile(path string, markerAndValues map[string][]string) error {
 	isGoFile := false
 	if ext := filepath.Ext(path); ext == ".go" {
 		isGoFile = true
@@ -84,7 +80,7 @@ func InsertStringsInFile(path string, markerAndValues ...string) error {
 		return err
 	}
 
-	r, err := insertStrings(f, markerAndValues...)
+	r, err := insertStrings(f, markerAndValues)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


pkg/scaffold/v2/internal/string_utils.go
```diff
@@ -115,3 +111,36 @@ func InsertStringsInFile(path string, markerAndValues ...string) error {
 
 	return err
 }
+
+// filterExistingValues removes the marker,value pair for values that already
+// exists in the given reader.
+func filterExistingValues(r io.Reader, markerAndValues ...string) (map[string][]string, error) {
```
mengqiy: IIRC each item in `markerAndValues` must NOT be a multi-line string.
Please add a comment for it.

pkg/scaffold/v2/main.go
```diff
@@ -45,6 +47,67 @@ func (m *Main) GetInput() (input.Input, error) {
...
+}
+
+// MainUpdateOptions contains info required for wiring an API/Controller in
+// main.go.
+type MainUpdateOptions struct {
+	// Project contains info about the project
+	Project *input.ProjectFile
+
+	// Resource is the resource being added
+	Resource *resource.Resource
```
mengqiy: `resource.Resource` comes from `"sigs.k8s.io/kubebuilder/pkg/scaffold/v1/resource"`.
IMO a `v2` package should not depend on a `v1` package. IOW, `resource` should be moved to a shared package.

pkg/scaffold/v2/internal/string_utils.go
```diff
@@ -33,44 +33,40 @@ import (
...
-			if strings.TrimSpace(line) == strings.TrimSpace(v) {
-				// since value already exist, so avoid duplication
-				delete(mvPairs, m)
-			}
-			if strings.Contains(line, m) {
-				_, err := buf.WriteString(v)
-				if err != nil {
-					return nil, err
+		for marker, vals := range markerVals {
+			if strings.Contains(line, marker) {
```
DirectXMan12: I'm not sure `strings.Contains` is correct here -- that means that `// +foo:marker:bar` would match `foo:marker`, which seems incorrect.

pkg/scaffold/v2/internal/string_utils.go
DirectXMan12: this should just take in a map

pkg/scaffold/v2/internal/string_utils.go
droot: Good catch. Added strict comparison.

pkg/scaffold/v2/internal/string_utils.go
droot: Done.

droot: Done.

pkg/scaffold/v2/main.go
droot: > IMO a v2 package should not depend on a v1 package.

Thats debatable :). If you don't feel strongly about it, I would keep this refactor out of scope for this PR.

pkg/scaffold/v2/internal/string_utils.go
```diff
@@ -33,44 +33,40 @@ import (
 // v1 will be inserted below the lines containing m1 string and v2 will be inserted
 // below line containing m2 string.
 func insertStrings(r io.Reader, markerAndValues ...string) (io.Reader, error) {
```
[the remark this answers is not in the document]
droot: I have enhanced one of the tests to include multiple strings. Let me know if there is any particular test you think we should add.

droot:
> we shouldn't be re-calling inflect all over the place, re-implementing package naming etc. We're going to have to burn some of this down soon

I agree.

pkg/scaffold/v2/main.go
mengqiy: IMHO when I change the scaffolding under v1 package, I wouldn't expect v2 scaffolding being affected. Otherwise, it makes it hard to reasoning.

I did a quick check, it seems we have quite some places in v2 that are importing v1 packages. It makes me sad. 
The change here is not the first one doing it. I'm OK to leave it in a separate PR. :)

droot: I took a look at the use of v1.Resource in v2 scaffolding and it turns out data to represent a resource is entirely same when it comes to v1/v2 scaffolding, so if we try to completely isolate v1/v2, we will end up duplicating the structure. Now for sharing the same data structure, one could move v1.Resource to a common pkg, but that has the same affect of importing v1 pkg in v2 because one has to reason about the usage of that shared structure in across v1/v2 scaffolding.
WDYT ?

pkg/scaffold/v2/internal/string_utils.go
```diff
@@ -115,3 +111,32 @@ func InsertStringsInFile(path string, markerAndValues ...string) error {
...
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// removeElementAt removes the element at index i from a.
+func removeElementAt(a []string, i int) []string {
```
mengqiy: This function can be one single line according to [this](https://github.com/golang/go/wiki/SliceTricks#delete).

pkg/scaffold/v2/main.go
mengqiy: Reorganize the code into `common`, `v1` and `v2` can make it easier to maintain the scaffolding the code in long term. But if adding change something in `common`, contributors still need to reason the usage of share structure across v1 and v2 scaffolding.

We can keep the scope of this PR as is, and have these thoughts in mind when we reorganizing the code in the followup PR.

pkg/scaffold/v2/internal/string_utils.go
droot: Thanks for the tip. Done :)

pkg/scaffold/v2/internal/string_utils.go
```diff
@@ -29,51 +28,48 @@ import (
...
+// line line containing marker string. So for ex. in insertStrings(r, {'m1':
+// [v1], 'm2': [v2]})
 // v1 will be inserted below the lines containing m1 string and v2 will be inserted
 // below line containing m2 string.
-func insertStrings(r io.Reader, markerAndValues ...string) (io.Reader, error) {
-	if len(markerAndValues)%2 != 0 {
-		return nil, fmt.Errorf("invalid marker and value pairs")
-	}
+func insertStrings(r io.Reader, markerAndValues map[string][]string) (io.Reader, error) {
+	// reader clone is needed since we will be reading twice from the given reader
```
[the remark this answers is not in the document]
droot: Yes, we are still reading twice, so this is needed.

pkg/scaffold/v2/internal/string_utils.go
```diff
@@ -29,51 +28,48 @@ import (
 )
 
 // insertStrings reads content from given reader and insert string below the
-// line line containing marker string. So for ex. in insertStrings(r, m1, v1, m2, v2)
+// line line containing marker string. So for ex. in insertStrings(r, {'m1':
```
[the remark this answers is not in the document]
droot: Done :)

pkg/scaffold/v2/main.go
droot: Ack.

PR #747, opened 2019-05-29
:book: Fix controller chapter indentation

DirectXMan12:
The comments in the cronjob controller were indented too much, causing
them to be interpretted as code blocks and not as normal text blocks.

This should fix that.


DirectXMan12:
cc @droot

DirectXMan12:
cc @pwittrock

PR #753, opened 2019-05-31
:bug: added multigroup check while creating resource

### LINKED ISSUE #739: Groups are no longer separated by package in v2.0.0-alpha.1 or v2.0.0-alpha2
KubeBuilder v1.0.8 would separate API groups into different packages (i.e. .../pkg/apis/group1/v1 and .../pkg/apis/group2/v1). With the new directory structure, it is not possible to make two different API groups that have the same version, as Kinds from both groups are generated in the same version folder (i.e. v1) and each version folder has only one groupversion_info.go file.

droot:
This change does the following:
  - introduces book-keeping of scaffolded resources in PROJECT file
  - Validates if new resource belong to existing scaffolded groups

fixes #739

### DIFF
--- a/pkg/scaffold/api.go
+++ b/pkg/scaffold/api.go
@@ -143,6 +143,10 @@ func (api *API) scaffoldV2() error {
 	r := api.Resource
 
 	if api.DoResource {
+		if err := api.validateResourceGroup(r); err != nil {
+			return err
+		}
+
 		fmt.Println(filepath.Join("api", r.Version,
 			fmt.Sprintf("%s_types.go", strings.ToLower(r.Kind))))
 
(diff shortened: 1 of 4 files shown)


DirectXMan12:
title edit: `s/:bugfix:/:bug:/`

DirectXMan12:
seems reasonable to me

droot:
> title edit: `s/:bugfix:/:bug:/`

Done.

DirectXMan12:
although it's not clear that we need the individual resources atm -- maybe just group-version?

also, you undid the fix :-P  -- it should be `:bug:`

pkg/scaffold/api.go
```diff
@@ -177,6 +181,15 @@ func (api *API) scaffoldV2() error {
 		if err != nil {
 			return fmt.Errorf("error updating kustomization.yaml: %v", err)
 		}
+
+		// update scaffolded resource in project file
+		api.project.Resources = append(api.project.Resources,
+			input.Resource{Group: r.Group, Version: r.Version, Kind: r.Kind})
+		err = saveProjectFile("PROJECT", api.project)
+		if err != nil {
+			fmt.Printf("error updating project file with resource information : %v \n", err)
```
mengqiy: Should it give up in this case?

droot: No, because this isn't *really* critical at the moment. This info just helps us to check multi-group check next time user create the resource. If this changes in future, we can make it fatal.

PR #758, opened 2019-06-04
:running: support GOBIN for installed binaries.

### LINKED ISSUE #746: Generated makefiles should run `go env GOBIN` instead of assuming `$GOPATH/bin`
The generated Makefile assumes compiled binaries from `go get` end up in `$GOPATH/bin`, but if you've overridden `$GOBIN`, they could end up elsewhere.  We should run `go env GOBIN` instead.

/good-first-issue
/kind bug

nirnanaaa:
This fixes #746 

I've removed the GOPATH entirely, since it was a redundant call anyways.

### DIFF
--- a/pkg/scaffold/v2/makefile.go
+++ b/pkg/scaffold/v2/makefile.go
@@ -48,6 +48,13 @@ IMG ?= {{ .Image }}
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
+# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
+ifeq (,$(shell go env GOBIN))
+GOBIN=$(shell go env GOPATH)/bin
+else
+GOBIN=$(shell go env GOBIN)
+endif
+
 all: manager
 
 # Run tests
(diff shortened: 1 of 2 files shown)


DirectXMan12:
Oh!  I remember why I didn't just use which in th second case (should add a comment there): gopath/bin or gobin isn't always on the user's path, so which might not find it

nirnanaaa:
Good point, but how was that solved with the old approach? If it was not in the `PATH` it would run go get every time.

Let me see if i can come up with something more suitable.

DirectXMan12:
yeah, it would run `go get` every time, but go would realize that it was already downloaded, so it wasn't a huge problem

nirnanaaa:
hmm @DirectXMan12 it works this way, but I am not quite happy about the approach I had to take. I am not sure if it's a good idea to name the variable `GOBIN`, but since it actually represents the golang install folder it seemed as if it would make sense. https://golang.org/cmd/go/#hdr-GOPATH_environment_variable states that GOBIN is not a mandatory setting.

DirectXMan12:
seems reasonable to me, although you'll need to mark it as phony.  You could also just not put it in a target at all, and have it always run, which is probably fine.

DirectXMan12:
/lgtm
/approve

Thanks!

PR #760, opened 2019-06-04
Use pointers for replicas field in the gitbook

### LINKED ISSUE #434: We should use pointer in the gitbhook example
We should use a pointer. i.e. *int32 instead of int32.
~~https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/basics/simple_resource.md#containersetspec~~
~~Maybe same for the string field.~~

Using a pointer can avoid a lot of tricky problems, we should use that in our gitbhook example.

EDIT: It's in https://github.com/kubernetes-sigs/kubebuilder/blob/book-v1/docs/book/basics/simple_resource.md

nzoueidi:
Update the replica field to use a pointer in the type definition and the controller reconciliation code.

Fixes #434 

/cc @mengqiy

### DIFF
--- a/docs/book/basics/simple_resource.md
+++ b/docs/book/basics/simple_resource.md
@@ -69,7 +69,7 @@ field, which may be set by autoscalers.
 // ContainerSetSpec defines the desired state of ContainerSet
 type ContainerSetSpec struct {
   // replicas is the number of replicas to maintain
-  Replicas int32 `json:"replicas,omitempty"`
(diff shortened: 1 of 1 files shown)


DirectXMan12:
Can you add a bit of explantation in the PR body beyond the `Fixes` clause

mengqiy:
@nzoueidi your editing to the PR description didn't generate a notification. Sorry for the delay.
/lgtm
/approve

PR #763, opened 2019-06-05
Add support for the resource scope marker

ironcladlou:
This enables generation of cluster-scoped types.

Depends on https://github.com/kubernetes-sigs/controller-tools/pull/220.

TODO:
- [x] Update controller-tools version to include https://github.com/kubernetes-sigs/controller-tools/pull/220
- [ ] Test(s)

### DIFF
--- a/pkg/scaffold/v2/types.go
+++ b/pkg/scaffold/v2/types.go
@@ -75,6 +75,9 @@ type {{.Resource.Kind}}Status struct {
 }
 
 // +kubebuilder:object:root=true
+{{ if not .Resource.Namespaced }}
+// +kubebuilder:resource:scope=Cluster
+{{ end }}
 
 // {{.Resource.Kind}} is the Schema for the {{ .Resource.Resource }} API
 type {{.Resource.Kind}} struct {
--- a/generated_golden.sh
+++ b/generated_golden.sh
@@ -64,6 +64,7 @@ scaffold_test_project() {
 		$kb init --project-version $version --domain testproject.org --license apache2 --owner "The Kubernetes authors"
 		$kb create api --group crew --version v1 --kind Captain --controller=true --resource=true --make=false
 		$kb create api --group crew --version v1 --kind FirstMate --controller=true --resource=true --make=false
+		$kb create api --group crew --version v1 --kind Admiral --controller=true --resource=true --namespaced=false --make=false
 		# TODO(droot): Adding a second group is a valid test case and kubebuilder is expected to report an error in this case. It
 		# doesn't do that currently so leaving it commented so that we can enable it later.
 		# $kb create api --group ship --version v1beta1 --kind Frigate --example=false --controller=true --resource=true --make=false


ironcladlou:
@DirectXMan12 I'm having a lot of trouble figuring out exactly how to test this change. If someone could outline the steps involved I can do the work.

DirectXMan12:
For tests, just add a command that generates a cluster-scoped resource to `generated_golden.sh`, regenerate the golden project, then make sure you get the right output.  We've got tests that are supposed to ensure that the generated golden projects stay in line with the source.

ironcladlou:
The generation/test toolchain is too platform-dependent for me to easily make use of (which is a longer discussion) — so, it might take me a while to get back to this. I'll chip away at it. I would benefit from a containerized process here.

DirectXMan12:
> The generation/test toolchain is too platform-dependent for me to easily make use of 

What do you mean?

ironcladlou:
>>The generation/test toolchain is too platform-dependent for me to easily make use of
>What do you mean?

Just trying to run `make generate`: modules issues (e.g. requires GOPATH), at least one Bash-ism incompatible with MacOS's ancient bash, requires at least an `etcd` binary... I haven't tried to go further yet.

DirectXMan12:
> requires GOPATH

We don't strictly need this any more -- we just need to purge the last bits from our infra scripts.

>  at least one Bash-ism incompatible with MacOS's ancient bash

Is is actually common for people to not have a newer version installed?

DirectXMan12:
> requires an etcd binary

we package the tooling required for running testenv for OSX -- see https://storage.googleapis.com/kubebuilder-tools/ (it's also in the KubeBuilder releases).  We should add a `go.kubebuilder` link for those

@droot does KB dev on his macbook -- if you need suggestions for how to set up OSX, I'm sure he can help

ironcladlou:
Thanks for the context. All the problems are solvable, I’m just saying I don’t have time to solve them at the moment, and thought you might be interested in contributor speed bumps.

DirectXMan12:
> thought you might be interested in contributor speed bumps

Yes, definitely.  FWIW, I've got #777 up for fixing the "running outside gopath" issues.

ironcladlou:
Had a moment to hack up the tree to get `make generate` running — next issue I'll need to run down is that `Admiral.DeepCopyObject()` isn't getting generated in testdata/project-v2/api/zz_generated.deepcopy.go (other copy methods are generated, and AdmiralList does have DeepCopyObject).

DirectXMan12:
Try building with master for the deepcopy bug -- we had a recent fix involving a case where not all struct members would get deepcopy generated.  Need to port the original deepcopy tests over.

pkg/scaffold/v2/types.go
```diff
@@ -75,6 +75,9 @@ type {{.Resource.Kind}}Status struct {
 }
 
 // +kubebuilder:object:root=true
+{{ if not .Resource.Namespaced }}
+// +kubebuilder:resource:scope=Cluster
```
mengqiy: Per https://github.com/kubernetes-sigs/controller-tools/blob/4d432d9dbc86bbf594c03f853878af64df8b95ec/pkg/crd/markers/crd.go#L181-L187, IIUC `path` should present as well. Otherwise, `make manifests` will complain.

DirectXMan12: I think we'll just want to make that optional

mengqiy: SGTM

droot:
closing this in favor of https://github.com/kubernetes-sigs/kubebuilder/pull/864

PR #771, opened 2019-06-07
:sparkles: add a flag for the leader election feature

mengqiy:
The leader election feature is disabled when running locally using `make
run`, while it is enabled by default when running in a k8s cluster.

### DIFF
--- a/pkg/scaffold/v2/main.go
+++ b/pkg/scaffold/v2/main.go
@@ -140,15 +140,18 @@ func init() {
 
 func main() {
 	var metricsAddr string
+	var enableLeaderElection bool
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
+		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
 	flag.Parse()
 
 	ctrl.SetLogger(zap.Logger(true))
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
-		Scheme:                  scheme,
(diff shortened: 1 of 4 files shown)


pkg/scaffold/v2/main.go
```diff
@@ -145,11 +145,15 @@ func main() {
...
-		Scheme:                  scheme,
-		MetricsBindAddress:      metricsAddr,
-		LeaderElection:          true,
-	})
+	op := ctrl.Options{
+		Scheme:             scheme,
+		MetricsBindAddress: metricsAddr,
+		LeaderElection:     true,
+	}
+	if ns := os.Getenv("KUBEBUILDER_LEADER_ELECTION"); len(ns) > 0 {
```
droot: Can we take this input as commandline flag ?

DirectXMan12: yeah, I think a flag works fine here.

mengqiy: Done

mengqiy:
PTAL

pkg/scaffold/v2/main.go
```diff
@@ -140,15 +140,17 @@ func init() {
 
 func main() {
 	var metricsAddr string
+	var enableElectionElection bool
```
droot: s/enableElectionElection/enableLeaderElection

pkg/scaffold/v2/main.go
```diff
@@ -140,15 +140,17 @@ func init() {
...
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.BoolVar(&enableElectionElection, "enable-leader-election", false, "Enable leader election feature.")
```
droot: `Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.`

PR #775, opened 2019-06-11
gitignore zz_generated.* files except for vendored stuff

ljdelight:
The vendor exception existed but there was a missing ignore for `**/zz_generated.*` :-)

### DIFF
--- a/pkg/scaffold/project/gitignore.go
+++ b/pkg/scaffold/project/gitignore.go
@@ -52,7 +52,7 @@ bin
 *.out
 
 # Kubernetes Generated files - skip generated files, except for vendored files
-
+**/zz_generated.*
 !vendor/**/zz_generated.*
 
 # editor and IDE paraphernalia


ljdelight:
Alright bot, I signed it.

ljdelight:
Hey, thanks a bunch for your time on this project and I'm happy to see that the CI is working. I tried to run the tests (`./test.sh`) locally but I've encountered issues. I think it's (mostly) from being outside of the GOPATH with `GO111MODUE=on`, and I'm not sure to proceed.

My problems are likely from being trunk and the super newness of gomodules (please don't interpret this issue as a complaint). 

These are the issues when working outside of the gopath and using go modules:
- Script `common.sh:51` (executed by test.sh) exits with error "Current directory is not in $GOPATH" if `$go_workspace` is empty. 
   - I expect this to work outside of gopath since gomodules are in use.
- Error in `pkg/scaffold/project.go` "dep is not installed"
   - Is this expectation is out of date since, according to the latest install instructions, `dep` isn't required?
- `common.sh: line 159: kind: command not found`
   - Thoughts? Probably user error here?
- `preparing test directory /src/sigs.k8s.io/kubebuilder-test` ... `mkdir: cannot create directory ‘/src’: Permission denied`

I'm unable to run the tests locally so I'm having trouble figuring out what else needs patched to fix the broken unit test. I'm very new to go and not experienced enough with go tests; unfortunately I'm not seeing how the `result.Actual` and `result.Golden` are set with the .gitignore scaffolding.

DirectXMan12:
> please don't interpret this issue as a complaint

Quite the contrary!  I'm always happy to receive detailed feedback on what's borked.  I run with gomodules, but happen to still organize stuff in a gopath, so it's quite likely that there are a couple of infra scripts that don't quite deal with being outside a gopath well.

I believe you're correct, most of those are gomodules-related.  Let me see if I can push a fix.

DirectXMan12:
> Error in `pkg/scaffold/project.go` "dep is not installed"

That's actually from the v1 scaffold tests, which still use dep.  The new v2 scaffold doesn't use dep, and nor does kubebuilder itself, but legacy generated projects still do, and we have tests to make sure we don't accidentally break those.

ljdelight:
Ah okay, that makes sense. I'll get `dep` on my system and in the meantime I can work in the GOPATH.

I now have a basic understading of go tests (ish), I need to update `testdata/gopath/src/project/.gitignore`. If you like the patch anyway!

DirectXMan12:
#777 should help.  Try cherry-picking it over and lmk if there's anything broken (ran through some basic tests, but was having some problems due to questionable internet)

pkg/scaffold/project/gitignore.go
```diff
@@ -52,7 +52,7 @@ bin
 *.out
 
 # Kubernetes Generated files - skip generated files, except for vendored files
-
+**/zz_generated.*
```
droot: I think we had this earlier and had to specifically take it out because without generated files checked-in, project will not be in compilable state which is a problem for projects that needs to be imported using `go get`.

ljdelight: Hmmmm yeah a compilable state would be important for `go get`, that's a great point. I'm currently working through the kubebuilder book and piecing things together, so my question is mostly naive, but isn't some of the code generation from `// +` declarations done at `go get` build time? Meaning two types of generations (commit vs not) where scaffolding has files to check in but build time stuff shouldn't be checked in?

droot: >but isn't some of the code generation from // + declarations done at go get build time

No. Makefile has a `generate` target that does the code generation.  `generate` is invoked as pre-requisite for other targets.

We used to use `go: generate` directive in project version "1" and that would trigger code generation at `go build`, but we realized it was inflexible and was also hiding too much from the users, so we took it away.

IMO, this change is not needed (feel free to close this PR).

ljdelight: I appreciate the comparison to v1, thank you. I'll close this :+1:

ljdelight:
The intention is to commit generated files, closing this

PR #776, opened 2019-06-11
:book: add doc for admission webhooks

### LINKED ISSUE #400: Come up with webhook developer workflow to test it locally
There are at least 2 possible solutions:
- Have a minikube cluster running locally and also run the webhook server locally.
- Have a cluster running in the cloud and webhook server running locally. Put a proxy service between them. The proxy service can either in the cluster or provided by a 3rd party.

Document the dev workflow.

mengqiy:
There is a section about suggested development workflow with `kind`, so this fixes #400


docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -0,0 +1,182 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	apimachineryutilvalidation "k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+/*
+Next, we'll setup logger for the webhooks.
```
DirectXMan12: Use the log handle in the CronJob struct.  If we're not scaffolding that, that's a mistake.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -0,0 +1,182 @@
...
+webhook manifest respectively.
+
+The meaning of each marker can be found [here](../TODO.md).
+*/
+
+// +kubebuilder:webhook:path=/mutate-batch-tutorial-kubebuilder-io-v1-cronjob,mutating=true,failurePolicy=fail,groups=batch.tutorial.kubebuilder.io,resources=cronjobs,verbs=create;update,versions=v1,name=mcronjob.kb.io
+// +kubebuilder:webhook:path=/validate-batch-tutorial-kubebuilder-io-v1-cronjob,mutating=false,failurePolicy=fail,groups=batch.tutorial.kubebuilder.io,resources=cronjobs,verbs=create;update,versions=v1,name=vcronjob.kb.io
+
+/*
+We implement the `webhook.Defaulter` interface for our CronJob CRD.
```
DirectXMan12: this sentence need to say why.  For example:

> we use the `webhook.Defaulter interface to add a defaulting to our CRD.  A webhook will automatically be served that calls this defaulting.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -0,0 +1,182 @@
...
+`webhook.Defaulter` only has one method: `Default()`.
+If `webhook.Defaulter` is implemented, a webhook server will be autowired to
+serve as the mutating webhook for CronJobs.
+*/
+
+var _ webhook.Defaulter = &CronJob{}
+
+// Default implements webhookutil.defaulter so a webhook will be registered for the type
```
DirectXMan12: uppercase on the interface?

DirectXMan12: You should explain a little bit about what this does in the block comment.  For instance

> The `Default` method is expected to mutate the receiver, setting the defaults.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -0,0 +1,182 @@
...
+		c.Spec.FailedJobsHistoryLimit = new(int32)
+		*c.Spec.FailedJobsHistoryLimit = 1
+	}
+}
+
+/*
+We implement the `webhook.Validator` interface for our CronJob CRD.
+`webhook.Validator` interface has two methods: `ValidateCreate()` and `ValidateUpdate()`.
+If `webhook.Validator` is implemented, a webhook server will be autowired to
+serve as the validating webhook for CronJobs.
```
DirectXMan12: Talk a bit about what both do.  For example:

> The ValidateCreate method is expected to validate that its receiver is a valid object upon creation.
> If not, it can return an error that's used as a validation message.
> We separate out ValidateCreate from ValidateUpdate to allow behavior like making certain fields immutable, so that they can only be set on creation.
> Here, however, we just use the same shared validation.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// ValidateUpdate implements webhookutil.validator so a webhook will be registered for the type
+func (c *CronJob) ValidateUpdate(old runtime.Object) error {
+	log.Info("validate update", "namespace", c.Namespace, "name", c.Name)
+	return c.validateCronJob()
+}
+
+/*
+We uses the same the validation logic for creation and update.
```
DirectXMan12: grammar: `We use`

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -0,0 +1,182 @@
...
+*/
+
+func (c *CronJob) validateCronJob() error {
+	allErrs := c.validateCronJobSpec()
+	if len(c.ObjectMeta.Name) > apimachineryutilvalidation.DNS1035LabelMaxLength-11 {
+		// The cronjob controller appends a 11-character suffix to the cronjob (`-$TIMESTAMP`) when
+		// creating a job. The job name length limit is 63 characters.
```
DirectXMan12: say why -- `the job name length is 63 character like all Kubernetes objects (which must fit in a DNS subdomain)`

DirectXMan12: Add a note that we could validate this declaratively, but we keep it here as an example

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -0,0 +1,182 @@
...
+		// then job creation will fail later.
+		allErrs = append(allErrs, field.Invalid(field.NewPath("metadata").Child("name"), c.Name, "must be no more than 52 characters"))
+	}
+
+	return apierrors.NewInvalid(
+		schema.GroupKind{Group: "batch.tutorial.kubebuilder.io", Kind: "CronJob"},
+		c.Name, allErrs)
+}
+
+func (c *CronJob) validateCronJobSpec() field.ErrorList {
```
DirectXMan12: add some comments in this.  Also, most of this isn't needed because we validate it declaratively.  We don't want to confuse users into thinking that both are needed.  We should make a point that you only really need this if you have complex logic that can't be validated declaratively (e.g. perhaps the format of the cron schedule?)

docs/book/src/cronjob-tutorial/webhook-implementation.md
```diff
@@ -0,0 +1,17 @@
...
+Kubebuilder take care of the rest for you, such as
+
+1. Creating the webhook server.
+1. Ensuring the server has been added in the manager.
+1. Creating handlers for your webhooks.
+1. Registering each handler with a path in your server.
+
+{{#literatego ./testdata/project/api/v1/cronjob_webhook.go}}
+
+That was a doozy, but now we've got a working controller.  Let's test
```
DirectXMan12: ??

docs/book/src/cronjob-tutorial/webhook-overview.md
```diff
@@ -0,0 +1,51 @@
+# Webhook
```
DirectXMan12: maybe move most of this to a reference section, and just give a brief summary here?

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
mengqiy: > Use the log handle in the CronJob struct.

Do you mean something like this?
```
type CronJob struct {
	metav1.TypeMeta   `json:",inline"`

	Spec   CronJobSpec   `json:"spec,omitempty"`
	Status CronJobStatus `json:"status,omitempty"`

	log logr.Logger `json:"-"`
}
```
It's not clear to me how to wire the log handle when creating a new CronJob instance.
I'm not convinced putting a log handle in the struct is better. So please convince me :)

mengqiy: An alternative is we always generate scaffold a package-level log handle in `groupversion_info.go` and just use that.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
mengqiy: > Add a note that we could validate this declaratively

This is `ObjectMeta.Name` which is not owned by the user. Not sure if it's doable through validation schema. Thoughts?

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
DirectXMan12: oh, shoot, I misread.  nvm

mengqiy:
PTAL

docs/book/src/reference/admission-webhook.md
```diff
@@ -0,0 +1,38 @@
+# Admission Webhook
+
+Admission webhooks are HTTP callbacks that receive admission requests, process them and return admission responses.
+There are two types of admission webhooks: mutating admission webhook and validating admission webhook.
```
droot: May be use list/bullets to enumerate the two webhooks ? 

Kubernetes provides following types of admission webhooks:
  - Mutating Admission Webhook: ....desc.....
  - Validating Admission Webhook: .....desc.....

docs/book/src/reference/admission-webhook.md
```diff
@@ -0,0 +1,38 @@
...
+With mutating admission webhooks, you may change the request object before it is stored (e.g. for implementing defaulting of fields)
+With validating admission webhooks, you may not change the request, but you can reject it (e.g. for implementing validation of the request).
+
+### Why Admission Webhooks are Important
+
+Admission webhooks are the mechanism to enable kubernetes extensibility through CRD.
+- Mutating admission webhook is the only way to do defaulting for CRDs.
```
droot: Can we re-word this to say: it can be used to default fields in a resource requests, for example setting <some-field> in cronjob if not specified by the user ? 

Also, users may not be familiar with the term 'defaulting'

docs/book/src/reference/admission-webhook.md
```diff
@@ -0,0 +1,38 @@
...
+- Validating admission webhook allows for more complex validation than pure schema-based validation.
+e.g. cross-field validation or cross-object validation.
```
droot: we discourage cross-object validation. But we can say, validating with an external source etc.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
DirectXMan12: it's explicitly doable, in fact.  `GenerateName` and `Name` are the only things that are allowed to have validation, going forward.  Unfortunately, we don't have a good way to accomplish it in KB, but it's doable in in Kubernetes.

docs/book/src/reference/admission-webhook.md
DirectXMan12: yeah, you should almost never do cross-object validation.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -0,0 +1,187 @@
...
+declaring validation in [here](../TODO.md).
+*/
+
+func (c *CronJob) validateCronJobSpec() field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	spec := c.Spec
+	fldPath := field.NewPath("spec")
+
+	if len(spec.Schedule) == 0 {
```
DirectXMan12: perhaps put an example of actually parsing the schedule and seeing if it has an error?

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
mengqiy: Yes, that's what I meant.
Upstream k/k can add a marker for `ObjectMeta.Name` to have maximum 63 length. But as a downstream user, if I can't change the marker in upstream to accomplish my need which is 52 max length.

mengqiy:
PTAL

mengqiy:
Rebased to resolve conflict.
PTAL

docs/book/src/SUMMARY.md
```diff
@@ -19,16 +19,25 @@
...
 
-      - [You said something about main?](./cronjob-tutorial/main-revisited.md)
+    - [You said something about main?](./cronjob-tutorial/main-revisited.md)
 
+  - [Implementing webhook](./cronjob-tutorial/webhook-implementation.md)
   - [Running and deploying the controller](./cronjob-tutorial/running.md)
+  
+    - [Deploying the cert manager](./cronjob-tutorial/cert-manager.md)
+    - [Deploying webhooks](./cronjob-tutorial/running-webhook.md)
+    - [Kind cluster](./cronjob-tutorial/kind.md)
```
DirectXMan12: This needs to either go first, or go in the reference documentation, or something -- it feels awkward here.

docs/book/src/SUMMARY.md
```diff
@@ -19,16 +19,25 @@
   - [What's in a controller?](./cronjob-tutorial/controller-overview.md)
   - [Implementing a controller](./cronjob-tutorial/controller-implementation.md)
 
-      - [You said something about main?](./cronjob-tutorial/main-revisited.md)
+    - [You said something about main?](./cronjob-tutorial/main-revisited.md)
 
+  - [Implementing webhook](./cronjob-tutorial/webhook-implementation.md)
```
DirectXMan12: Call this something like `Implementing validating/defaulting webhooks` or something slightly more descriptive (also grammar nit: webhooks should be plural (note the `s`))

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -0,0 +1,180 @@
...
+*/
+package v1
+
+import (
+	"github.com/robfig/cron"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	apimachineryutilvalidation "k8s.io/apimachinery/pkg/util/validation"
```
DirectXMan12: nit: this name is ridiculously long.  we should find something shorter -- even `validationutils` or something is fine.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -0,0 +1,180 @@
...
+
+// +kubebuilder:webhook:path=/mutate-batch-tutorial-kubebuilder-io-v1-cronjob,mutating=true,failurePolicy=fail,groups=batch.tutorial.kubebuilder.io,resources=cronjobs,verbs=create;update,versions=v1,name=mcronjob.kb.io
+// +kubebuilder:webhook:path=/validate-batch-tutorial-kubebuilder-io-v1-cronjob,mutating=false,failurePolicy=fail,groups=batch.tutorial.kubebuilder.io,resources=cronjobs,verbs=create;update,versions=v1,name=vcronjob.kb.io
+
+/*
+We use the `webhook.Defaulter` interface to set defaults to our CRD.
+A webhook will automatically be served that calls this defaulting.
+*/
+
+var _ webhook.Defaulter = &CronJob{}
```
DirectXMan12: just merge these two doc blocks together.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -0,0 +1,180 @@
...
+		*c.Spec.SuccessfulJobsHistoryLimit = 3
+	}
+	if c.Spec.FailedJobsHistoryLimit == nil {
+		c.Spec.FailedJobsHistoryLimit = new(int32)
+		*c.Spec.FailedJobsHistoryLimit = 1
+	}
+}
+
+/*
+We use the `webhook.Validator` interface to validate our CRD.
```
DirectXMan12: `to validate our CRD beyond what's possible with declarative validation.  Generally, declarative validation should be sufficient, but sometimes more advanced usecases call for complex validation` or something

DirectXMan12: For instance, we'll see below that we use this to validate a well-formed cron schedule without making up a long regular expression.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -0,0 +1,180 @@
...
+func validateScheduleFormat(schedule string, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if _, err := cron.ParseStandard(schedule); err != nil {
+		allErrs = append(allErrs, field.Invalid(fldPath, schedule, err.Error()))
+	}
+
+	return allErrs
+}
+
+// +kubebuilder:docs-gen:collapse=Validate schedule format
```
DirectXMan12: I wouldn't collapse this -- it's kinda the interesting part

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -0,0 +1,180 @@
...
+Some fields are declaratively validated by openapi schema.
+You can find kubebuilder validation markers (prefixed
+with `// +kubebuilder:validation`) in the [API](api-design.md)
+You can find all of the kubebuilder supported markers for
+declaring validation in [here](../TODO.md).
+*/
+
+func (c *CronJob) validateCronJobSpec() field.ErrorList {
+	allErrs := field.ErrorList{}
+	allErrs = append(allErrs, validateScheduleFormat(
```
DirectXMan12: add a comment about the field stuff and why we're using it (e.g. `the field helpers from the kubernetes API machinery help us return nicely structured validation errors`)

docs/book/src/reference/webhook-overview.md
```diff
@@ -0,0 +1,15 @@
...
+HTTP request (typically POST) to other application when certain event happens.
+In the kubernetes world, there are 3 kinds of webhooks:
+[admission webhook](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#admission-webhooks),
+[authorization webhook](https://kubernetes.io/docs/reference/access-authn-authz/webhook/) and
+[CRD conversion webhook](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#webhook-conversion).
+
+In [controller-runtime](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/webhook)
+libraries, we support admission webhooks and CRD conversion webhooks.
+
+Admission webhook feature is beta since kubernetes 1.9. CRD conversion webhook
```
DirectXMan12: grammar nit: this sentence is a bit awkward -- maybe turn it around and say something like `Kubernetes supports these dynamic admission webhooks as of version 1.9 (when the feature entered beta).` etc

docs/book/src/reference/webhook-overview.md
```diff
@@ -0,0 +1,15 @@
+# Webhook
+
+Webhooks are HTTP callbacks, providing a way for notifications to be delivered
+to an external web server. A web application implementing webhooks will send an
```
DirectXMan12: This explanation confuses things.  They're not really "notifications" (which imply fire-and-forget on the part of the caller) -- they're requests for information sent in a blocking fashion.

docs/book/src/reference/webhook-overview.md
```diff
@@ -0,0 +1,15 @@
...
+HTTP request (typically POST) to other application when certain event happens.
```
DirectXMan12: The exact HTTP verb is a level of detail that we don't need here.

docs/book/src/reference/admission-webhook.md
```diff
@@ -0,0 +1,21 @@
+# Admission Webhook
```
DirectXMan12: `Admission Webhooks` (note the `s`)

docs/book/src/reference/admission-webhook.md
```diff
@@ -0,0 +1,21 @@
...
+
+Admission webhooks are HTTP callbacks that receive admission requests, process
+them and return admission responses.
+
+Kubernetes provides the following types of admission webhooks:
+
+- **Mutating Admission Webhook**: It can mutate the object before it is stored.
```
DirectXMan12:
```markdown
* ***Mutating Webhooks***: These can mutate the object while it's being created or updated, before it gets stored.
(continue with the rest of your explanation
```

Similar with validating

docs/book/src/reference/admission-webhook.md
```diff
@@ -0,0 +1,21 @@
...
+It can be used to default fields in a resource requests, e.g. fields in
+Deployment that are not specified by the user. It can be used to inject sidecar
+containers.
+
+- **Validating Admission Webhook**: It can validate the object and reject it if
+invalid. It allows more complex validation than pure schema-based validation.
+e.g. cross-field validation and pod image whitelisting.
+
+The apiserver by default doesn't authenticate itself to the webhooks. But if you
```
DirectXMan12: `However, if you` (don't start a sentence with `but` like that)

docs/book/src/reference/admission-webhook.md
```diff
@@ -0,0 +1,21 @@
...
+want to authenticate the clients, you need to configure the apiserver to use
```
DirectXMan12: s/need to/can/

mengqiy:
PTAL
I will squash commits before merging.

cassianoleal:
Shouldn't automated tests (ginkgo/gomega) be a part of this?

mengqiy:
@cassianoleal kubebuilder v2 ATM doesn't scaffold much testing files, since we are trying to figure out what is the best way to scaffold tests. We do have plan to support it soon after we have resolve our v2 release blockers :)

mengqiy:
This should be ready to go.

docs/book/src/reference/kind.md
```diff
@@ -0,0 +1,52 @@
...
+
+```bash
+kind create cluster
+```
+
+To customize your cluster, you can provide additional configuration.
+For example, the following is a sample `kind` configuration.
+
+```yaml
+{{#include ./testdata/project/hack/kind-config.yaml}}
```
droot: This is not being rendered properly. Not sure what is missing.

docs/book/src/reference/kind.md
```diff
@@ -0,0 +1,52 @@
...
+You can use `--image` flag to specify the cluster version you want, e.g.
+`--image=kindest/node:v1.13.6`, the supported version are listed
+[here](https://hub.docker.com/r/kindest/node/tags)
+
+## Cheetsheet
+
+- [Load a local image into a kind cluster](https://kind.sigs.k8s.io/docs/user/quick-start/#loading-an-image-into-your-cluster).
+
+```bash
+kind load docker-image your-image-namge:your-tag
```
droot: typo: your-image-name

docs/book/src/reference/kind.md
mengqiy: Good catch.
This issue is caused by moving this doc from `tutorial` to `reference`.

mengqiy:
Ready to ship.

mengqiy:
Thanks for raising this.
We have noticed this yesterday. We plan to temporarily drop it from the example https://github.com/kubernetes-sigs/kubebuilder/pull/853.
We will add it back when we fix it in controller-tools.

mengqiy:
filed https://github.com/kubernetes-sigs/controller-tools/issues/256 to track

mengqiy:
@gautamkmr https://github.com/kubernetes-sigs/kubebuilder/pull/857 fixed it.

PR #777, opened 2019-06-11
:running: Support running tests outside of the gopath

DirectXMan12:
This reduces reliance on the gopath for running tests, instead
provisioning temporary gopaths for that purpose, and using special
gobin locations to download binaries.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -41,17 +41,10 @@ cd "$base_dir" || {
   exit 1
 }
 
-go_workspace=''
-for p in ${GOPATH//:/ }; do
-  if [[ $PWD/ = $p/* ]]; then
-    go_workspace=$p
-  fi
-done
-
-if [ -z $go_workspace ]; then
-  echo 'Current directory is not in $GOPATH' >&2
-  exit 1
-fi
+# create a temporary gopath, and clean it up (with some sanity checks to ensure
+# we don't do anything terrible)
+tmp_gopath=$(mktemp --tmpdir -d kubebuilder-test-gopath-XXXXXX)
+trap "[[ ${tmp_gopath} == *kubebuilder-test-gopath* ]] && rm -rf ${tmp_gopath}" SIGHUP SIGINT SIGTERM
 
 # k8s_version=1.11.0
 k8s_version=1.14.1
@@ -111,6 +104,10 @@ function prepare_staging_dir {
   fi
 }
 
+# make sure to use our test gobin too
+test_gobin="/tmp/kubebuilder-test-bin"
+export PATH=${PATH}:${test_gobin}
+
 # fetch k8s API gen tools and make it available under kb_root_dir/bin.
 function fetch_tools {
   if [ -n "$SKIP_FETCH_TOOLS" ]; then
@@ -126,6 +123,19 @@ function fetch_tools {
     curl -sL ${kb_tools_download_url} -o "$kb_tools_archive_path"
   fi
   tar -zvxf "$kb_tools_archive_path" -C "$tmp_root/"
+
+  header_text "ensuring dep is present for legacy v1 scaffold tests"
+  if ! which dep; then
+    mkdir -p test_gobin
+    # can't install dep with modules due to dependency versioning (lol)
+    curl -sL -o ${test_gobin}/dep https://github.com/golang/dep/releases/download/0.5.2/dep-$(go env GOOS)-$(go env GOARCH)
+  fi
+
+  header_text "ensuring kind is present for e2e tests"
+  if ! which kind; then
+    mkdir -p test_gobin
+    GOBIN=${test_gobin} GO111MODULE="on" go get sigs.k8s.io/kind@v0.3.0
+  fi
 }
 
 function build_kb {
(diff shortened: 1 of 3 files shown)


DirectXMan12:
cc #775

droot:
Change looks reasonable to me. e2e test is failing for v2 scaffolding /cc @mengqiy

common.sh
```diff
@@ -126,6 +123,19 @@ function fetch_tools {
     curl -sL ${kb_tools_download_url} -o "$kb_tools_archive_path"
   fi
   tar -zvxf "$kb_tools_archive_path" -C "$tmp_root/"
+
+  header_text "ensuring dep is present for legacy v1 scaffold tests"
+  if ! which dep; then
+    mkdir -p test_gobin
+    # can't install dep with modules due to dependency versioning (lol)
+    curl -sL -o ${test_gobin}/dep https://github.com/golang/dep/releases/download/0.5.2/dep-$(go env GOOS)-$(go env GOARCH)
```
ljdelight: I get the file but it's not executable, `chmod +x` please?
PS in github I'm not sure if I should use "Add a single comment" or "Start a review", I'm picking at random

ljdelight:
Ohk I didn't have git-lfs installed so I've made it a bit further. From that issue the CONTRIBUTING might have a line that needs updated since it fails out of the gopath `go build -o /output/path/kubebuilder ./cmd` https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md

common.sh
ljdelight: I'm testing this stuff so I can see what happens using `bash -exu ./test.sh` and I found that the `mkdir -p test_gobin` isn't using the variable `${test_gobin}`.  I know that I had dep in the /tmp/.. directory so it's probably created elsewhere and not failing here

common.sh
```diff
@@ -126,6 +123,19 @@ function fetch_tools {
...
+  fi
+
+  header_text "ensuring kind is present for e2e tests"
+  if ! which kind; then
+    mkdir -p test_gobin
+    GOBIN=${test_gobin} GO111MODULE="on" go get sigs.k8s.io/kind@v0.3.0
```
ljdelight: This causes a diff on the projects root-level go.mod, which is not expected? Does go literally need us to `go get` outside of _any_ project? I'm too new to golang (and modules) to have an opinion, and I'm looking forward to knowing what the go tooling expects from users here...

```
diff --git a/go.mod b/go.mod
index bbb594bf..0163e4b6 100644
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,6 @@ go 1.12
 
 require (
        github.com/gobuffalo/envy v1.6.15 // indirect
-       github.com/inconshreveable/mousetrap v1.0.0 // indirect
        github.com/markbates/inflect v1.0.4
        github.com/onsi/ginkgo v1.8.0
        github.com/onsi/gomega v1.5.0
@@ -12,9 +11,9 @@ require (
        github.com/spf13/afero v1.2.2
        github.com/spf13/cobra v0.0.3
        github.com/spf13/pflag v1.0.3
-       golang.org/x/net v0.0.0-20190514140710-3ec191127204 // indirect
        golang.org/x/sys v0.0.0-20190506115046-ca7f33d4116e // indirect
        golang.org/x/text v0.3.2 // indirect
        golang.org/x/tools v0.0.0-20190506145303-2d16b83fe98c
        gopkg.in/yaml.v2 v2.2.2
+       sigs.k8s.io/kind v0.3.0 // indirect
 )
```

common.sh
ironcladlou: What about incorporating the `dep` into the dependency graph as a [build-scoped tool](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module)?

mengqiy:
@droot It's failing on macOS only now.

common.sh
DirectXMan12: dep, unfortunately, doesn't install properly using using modules without quite a bit of fiddling (mismatched versions, somewhat ironically).

common.sh
DirectXMan12: go tooling likes modifying everything.  I think there may be a flag we can pass to tell it to not do this.

DirectXMan12: hmm... looked around a bit, maybe not.  can run `go mod tidy` afterwards to fix it.

common.sh
```diff
@@ -41,17 +41,10 @@ cd "$base_dir" || {
...
-  fi
-done
-
-if [ -z $go_workspace ]; then
-  echo 'Current directory is not in $GOPATH' >&2
-  exit 1
-fi
+# create a temporary gopath, and clean it up (with some sanity checks to ensure
+# we don't do anything terrible)
+tmp_gopath=$(mktemp --tmpdir -d kubebuilder-test-gopath-XXXXXX)
```
mcristina422: This [fails on OSX](https://travis-ci.org/kubernetes-sigs/kubebuilder/jobs/544391172) with `mktemp: illegal option -- -`

I suggest that this be changed to `mktemp -d "${TMPDIR:-/tmp}/kubebuilder-test-gopath-XXXXXX"` as it should work regardless of platform.

camilamacedo86:
Hi @DirectXMan12, 

It was approved and not merged. Now it shows outdated with the master branch. Should we still move forward with? Also, see that the GOBIN issue will be just faced in the prow. Note that in the #1135 to upgrade for 1.13 we start to use the curl to install the dep as well.

DirectXMan12:
closing, very old and already dealt with

Issue #379, opened 2018-09-05
Document how to use finalizers with examples


droot:
Add a section under `Beyond basics` in kubebuilder book describing details of `finalizers` and how to use them with example code. 

Also investigate if we can do something during `scaffolding` steps and includes comments to guide users about it.

droot:
We have a section on [`using finalizers`](https://book.kubebuilder.io/beyond_basics/using_finalizers.html) , so closing this.

gordonbondon:
This link now goes to 404 and I was not able to find this info in repo. Any idea where I find this info?

gordonbondon:
For anyone like me still looking for this doc, looks like it was deleted during migration to new gitbook for v2. Here is the last version of finalizer doc https://github.com/kubernetes-sigs/kubebuilder/blob/e46e06eb12dab4c83ad54f27a02344a1e70857fe/docs/book/beyond_basics/using_finalizers.md

DirectXMan12:
That should still exist in the v1 book: https://book-v1.book.kubebuilder.io/beyond_basics/using_finalizers.html.  We're working on adding it to the next version of the book -- in the mean time it's a help-wanted.

/help-wanted

PR #778, opened 2019-06-11
:book: docs for using finalizers

droot:
This PR adds reference documentation for using finalizers.

fixes https://github.com/kubernetes-sigs/kubebuilder/issues/379

### DIFF
--- a/docs/book/src/reference/reference.md
+++ b/docs/book/src/reference/reference.md
@@ -1,0 +1,5 @@
+# Reference
+  
+  - [Using Finalizers](./using-finalizers.md): Finalizers are a mechanism to
+	execute any custom logic related to a resource before it gets deleted from
+	Kubernetes cluster.
--- a/docs/book/src/reference/using-finalizers.md
+++ b/docs/book/src/reference/using-finalizers.md
@@ -1,0 +1,25 @@
+# Using Finalizers
+
+`Finalizers` allow controllers to implement asynchronous pre-delete hooks. Let's
+say you create an external resource (such as a storage bucket) for each object of
+your API type, and you want to delete the associated external resource
+on object's deletion from Kubernetes, you can use a finalizer to do that.
+
+You can read more about the finalizers in the [Kubernetes reference docs](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#finalizers). The section below demonstrates how to register and trigger pre-delete hooks
+in the `Reconcile` method of a controller.
+
+The key point to note is that a finalizer causes "delete" on the object to become 
+an "update" to set deletion timestamp. Presence of deletion timestamp on the object
+indicates that it is being deleted. Otherwise, without finalizers, a delete
+shows up as a reconcile where the object is missing from the cache.
+
+Highlights:
+- If the object is not being deleted and does not have the finalizer registered,
+  then add the finalizer and update the object in Kubernetes.
+- If object is being deleted and the finalizer is still present in finalizers list,
+  then execute the pre-delete logic and remove the finalizer and update the
+  object.
+- Ensure that the pre-delete logic is idempotent.
+
+{{#literatego ../cronjob-tutorial/testdata/finalizer_example.go}}
+
--- a/docs/book/src/SUMMARY.md
+++ b/docs/book/src/SUMMARY.md
@@ -26,4 +26,10 @@
 
 ---
 
+- [Reference](./reference/reference.md)
+
+  - [Using Finalizers](./reference/using-finalizers.md)
+
+---
+
 [Appendix: The TODO Landing Page](./TODO.md)
--- a/docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
+++ b/docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
@@ -1,0 +1,116 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+// +kubebuilder:docs-gen:collapse=Apache License
+
+/*
+First, we start out with some standard imports.
+As before, we need the core controller-runtime library, as well as
+the client package, and the package for our API types.
+*/
+package controllers
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	batchv1 "tutorial.kubebuilder.io/project/api/v1"
+)
+
+// +kubebuilder:docs-gen:collapse=Imports
+
+/*
+The code snippet below shows skeleton code for implementing a finalizer.
+*/
+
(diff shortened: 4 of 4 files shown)


docs/book/src/reference/using-finalizers.md
```diff
@@ -0,0 +1,99 @@
+# Using Finalizers
+
+`Finalizers` allow controllers to implement asynchronous pre-delete hooks. Let
+say, you create an external resource such as a storage bucket for each object of
+an API type you are implementing and you need to delete the associated external resource
+when the corresponding object is deleted from Kubernetes, you can use a
+finalizer to delete the external resource.
+
+You can read more about the finalizers in the [kubernetes reference docs](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#finalizers). Section below
```
DirectXMan12: sp: `kubernetes` --> `Kubernetes` (capitalization)

docs/book/src/reference/using-finalizers.md
```diff
@@ -0,0 +1,99 @@
+# Using Finalizers
+
+`Finalizers` allow controllers to implement asynchronous pre-delete hooks. Let
```
DirectXMan12: grammar: `let's say`, and no comma following

docs/book/src/reference/using-finalizers.md
```diff
@@ -0,0 +1,99 @@
...
+say, you create an external resource such as a storage bucket for each object of
```
DirectXMan12: `(such as a storage bucket)`

docs/book/src/reference/using-finalizers.md
```diff
@@ -0,0 +1,99 @@
...
+an API type you are implementing and you need to delete the associated external resource
```
DirectXMan12: `, and you need`

docs/book/src/reference/using-finalizers.md
```diff
@@ -0,0 +1,99 @@
...
+when the corresponding object is deleted from Kubernetes, you can use a
+finalizer to delete the external resource.
```
DirectXMan12: this sentence runs on quite a bit -- please clean it up a bit.

docs/book/src/reference/using-finalizers.md
DirectXMan12: grammar: `The section below`

docs/book/src/reference/using-finalizers.md
```diff
@@ -0,0 +1,99 @@
...
+
+You can read more about the finalizers in the [kubernetes reference docs](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#finalizers). Section below
+demonstrates how to register and trigger pre-delete hooks in the `Reconcile`
+method of a controller.
+
+Highlights:
+- If object is not being deleted and does not have the finalizer registered,
```
DirectXMan12: grammar: `the object`

docs/book/src/reference/using-finalizers.md
```diff
@@ -0,0 +1,99 @@
...
+  then add the finalizer and update the object in kubernetes.
```
DirectXMan12: should we encourage a mutating webhook for this instead?  It's probably safer.

docs/book/src/reference/using-finalizers.md
```diff
@@ -0,0 +1,99 @@
...
+- If object is being deleted and the finalizer is still present in finalizers list,
```
DirectXMan12: we should probably note somewhere that a finalizer causes a "delete" to become an update to set deletion timestamps, so that's how you detect a delete.  Otherwise, without finalizers, a delete shows up as a reconcile where the object is missing from the cache.

docs/book/src/reference/using-finalizers.md
```diff
@@ -0,0 +1,99 @@
...
+  then execute the pre-delete logic and remove the finalizer and update the
+  object.
+- You should implement the pre-delete logic in such a way that it is safe to 
+ invoke it multiple times for the same object.
+
+```go
```
DirectXMan12: use `literatego` for this -- you can put the full code in, but mark sections that you don't want to emphasize as collapsed (see `main_revisited.md`)

docs/book/src/reference/using-finalizers.md
```diff
@@ -0,0 +1,99 @@
...
+	// delete any external resources associated with the instance
+	//
+	// Ensure that delete implementation is idempotent and safe to invoke
+	// multiple types for same object.
+}
+
+//
+// Helper functions to check and remove string from a slice of strings.
+//
+func containsString(slice []string, s string) bool {
```
DirectXMan12: We should have a helper for this in CR -- something like `NeedsFinalizing(string, runtime.Object)` and `RemoveFinalizer(string, runtime.Object)`.  Perhaps even wrap those up into: `FinalizeIfNeeded(string, runtime.Object, func())`

droot: Agreed. Have created an issue in CR --> https://github.com/kubernetes-sigs/controller-runtime/issues/485

docs/book/src/reference/using-finalizers.md
droot: Thanks for the tip, exploring it.

droot:
@DirectXMan12 we need Github Action/Suggestion for grammarly :)

docs/book/src/reference/using-finalizers.md
droot: I took a look. Earlier I was thinking of reusing Cronjob controller file to demonstrate that, but looks like sharing the same `.go` file for two different pages will be challenging because the two pages will have different requirement for collapsing portions of the `.go` file.

I took a look at `empty_controller.go` also, so I can follow that pattern also where I write this in another `.go` file, but then I can't use `cronjob` reconciler because it won't compile (not sure if we compile the Go files that we have under docs) because I will end-up overriding Reconciler. 

I am trying to understand now which approach is better ? (code embedded in markdowns or markdown (HTML) embedded in Go) :)? 

Want to understand how are you approaching this ?

droot: For now, I have created a file `finalizer_example.go` and moved the go code there so that we have something consistent to begin with. We don't sanity check (compile) go files under book dir but I see a few challenges around that.

droot:
@DirectXMan12 addressed the comments, please take a look.

docs/book/src/reference/using-finalizers.md
```diff
@@ -0,0 +1,26 @@
...
+indicates that it is being deleted. Otherwise, without finalizers, a delete
+shows up as a reconcile where the object is missing from the cache.
+
+Highlights:
+- If the object is not being deleted and does not have the finalizer registered,
+  then add the finalizer and update the object in Kubernetes.
+- If object is being deleted and the finalizer is still present in finalizers list,
+  then execute the pre-delete logic and remove the finalizer and update the
+  object.
+- You should implement the pre-delete logic in such a way that it is safe to
```
mengqiy: nit: It'd be more concise to say something like `the pre-delete logic should be idempotent for the same logic`.

docs/book/src/reference/reference.md
```diff
@@ -0,0 +1,5 @@
+# Reference
+  
+  - [Using Finalizers](./using-finalizers.md): Finalizers are a mechanism to
+	execute any custom logic related to a resource before it gets deleted from
+	Kubernetes.
```
mengqiy: should be `Kubernetes cluster`

docs/book/src/reference/using-finalizers.md
droot: Done.

docs/book/src/reference/reference.md
droot: Done.

droot:
@mengqiy Addressed the comments. I think this should be good to go now.

PR #780, opened 2019-06-12
fix: additonal->additional

sethp-nr:
Fixes a minor typo in the templates.


sethp-nr:
Hmm, I have found a few more minor issues (by running golangci-lint against a new project). Specifically:

An unused variable in the generated tests: https://github.com/kubernetes-sigs/kubebuilder/commit/92bbf218ea6088d080e3c45648c7122aac6db408
and an ignored return value when adding to scheme: https://github.com/kubernetes-sigs/kubebuilder/commit/60602c494370f82452631c1d13d139e0be2c4735

Would you prefer I submit each as an individual PR, or wrap them all up into one?

PR #781, opened 2019-06-12
fix golangci-lint reported issues

sethp-nr:
Follow up from #780, this fixes a few more issues.

I haven't managed to get the tests working yet, so I expect these will fail while I figure that out.

### DIFF
--- a/generated_golden.sh
+++ b/generated_golden.sh
@@ -33,7 +33,7 @@ scaffold_test_project() {
 	testdata_dir=$(pwd)/testdata
 	mkdir -p ./testdata/$project
 	rm -rf ./testdata/$project/*
-	pushd . 
+	pushd .
 	cd testdata/$project
 
 	kb=$testdata_dir/../bin/kubebuilder
@@ -79,6 +79,13 @@ scaffold_test_project() {
 }
 
 set -e
+
+if ! git lfs > /dev/null 2>&1 ; then
+    echo "this project requires git-lfs, see: https://git-lfs.github.com/"
+    echo "after installing, you'll need to git checkout ./testdata/*tgz"
+    exit 1
+fi
+
 build_kb
 scaffold_test_project gopath/src/project 1
 scaffold_test_project project-v2 2
--- a/testdata/project-v2/main.go
+++ b/testdata/project-v2/main.go
@@ -37,8 +37,8 @@ var (
 
 func init() {
 
-	crewv1.AddToScheme(scheme)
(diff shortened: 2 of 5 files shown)


droot:
You need to run `make generate` to update the test project, otherwise it looks good.

generated_golden.sh
```diff
@@ -79,6 +79,13 @@ scaffold_test_project() {
...
+
+if ! git lfs > /dev/null 2>&1 ; then
+    echo "this project requires git-lfs, see: https://git-lfs.github.com/"
+    echo "after installing, you'll need to git checkout ./testdata/*tgz"
+    exit 1
+fi
+
 build_kb
-scaffold_test_project gopath/src/project 1
+#scaffold_test_project gopath/src/project 1
```
droot: uncomment :)

droot:
Can you also squash all the commits. Thanks

DirectXMan12:
we should see if we can get the lint running as part of our tests, too.

generated_golden.sh
sethp-nr: Oops, thought I caught that one. Will fix.

sethp-nr:
@DirectXMan12 that would be a good idea. I think golangci-lint is a superset of `go vet`, so it might be as easy as swapping out the `go vet` for `golangci-lint run` with a `which golangci-lint`-guarded go get.

testdata/project-v2/main.go
```diff
@@ -37,8 +37,8 @@ var (
 
 func init() {
 
-	crewv1.AddToScheme(scheme)
-	corev1.AddToScheme(scheme)
+	_ = crewv1.AddToScheme(scheme)
```
mcristina422: Shouldn't these errors be checked? I've seen `Must(crewv1.AddToScheme(scheme))` in k8s codebases

where 
```go
func Must(err error) {
  if err != nil {
    panic(err)
  }
}
```

droot: Ideally yes and that is a correct resolution for the error reported by `AddToScheme`. Implementing  suggested fix will require a bit of work/refactoring on how we generate the code fragments for `AddToScheme` calls and don't want to hold this PR for that to happen. We can do that in a separate PR (may be with a PR that runs `golangci-lint run` instead of `go vet` for generated project as suggested above).

droot: created an issue https://github.com/kubernetes-sigs/kubebuilder/issues/791 to track it.

PR #794, opened 2019-06-18
:running: use vendor tarball from GCS bucket

droot:
Due to git-lfs quota issues, we are hosting vendor tarball on GCS.
This change enables using the GCS hosted vendor tarball instead of using the one from Github LFS.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -85,14 +85,27 @@ y
 EOF
 }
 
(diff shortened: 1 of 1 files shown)


DirectXMan12:
for posterity -- we'll have to have a manual vendor update step now, but such is life.  GitHub's bandwidth quota is a bit low :-(.

Issue #797, opened 2019-06-19
Default scheme should registered builtin types


mengqiy:
Default scheme scaffolded by the kubebuilder should have builtin types all registered, which is what we did in v1. This reduce users burden of understanding and manage the scheme.

/kind feature

DirectXMan12:
We should also see if we can make this error clearer, since it's not obvious unless you already know the problem.

functicons:
I was trying to create a Deployment from my reconciler, then I got this error:

```no kind is registered for the type v1.Deployment in scheme pkg/runtime/scheme.go:101```

Is this issue related? Where can I find an example to fix it?

mengqiy:
Yes
What you need to add is this line.
https://github.com/kubernetes-sigs/kubebuilder/blob/b18a4e3b8f83d7d0ca461c28794aa7ac92c278ef/testdata/project-v2/main.go#L40

functicons:
@mengqiy Does `clientgoscheme.AddToScheme(scheme)` register all builtin types? I used `appsv1.AddToScheme(scheme)` and it worked.

mengqiy:
> Does clientgoscheme.AddToScheme(scheme) register all builtin types?

Yes. 

> If so, why does it also calls corev1.AddToScheme(scheme)?

This is scaffolded by kubeubilder. KB currently is not smart enough to skip it for builtin types. AFAIK registering it twice should not have any side effects.

PR #802, opened 2019-06-19
:sparkles: scheme should have all builtin types registered by default

mengqiy:
This new scheme should make users life easier.

fixes https://github.com/kubernetes-sigs/kubebuilder/issues/797

xref: https://github.com/kubernetes-sigs/controller-runtime/issues/492

### DIFF
--- a/pkg/scaffold/v2/main.go
+++ b/pkg/scaffold/v2/main.go
@@ -119,6 +119,7 @@ import (
 	"flag"
     "os"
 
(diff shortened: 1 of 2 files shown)


DirectXMan12:
We should encourage proper behavior -- adding the existing types to a new scheme is good behavior, but adding new types to someone else's scheme is not.

DirectXMan12:
I'll post a PR to fix

PR #809, opened 2019-06-20
Add existing k8s types to new scheme

### LINKED ISSUE #797: Default scheme should registered builtin types

DirectXMan12:
It's generally not good practice to modify someone else's scheme, so we
add k8s types to our new scheme instead.

(re) fixes #797

### DIFF
--- a/pkg/scaffold/v2/main.go
+++ b/pkg/scaffold/v2/main.go
@@ -130,11 +130,12 @@ import (
 )
 
 var (
-	scheme   = clientgoscheme.Scheme
-    setupLog = ctrl.Log.WithName("setup")
+	scheme = runtime.NewScheme()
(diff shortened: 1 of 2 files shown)


mengqiy:
> modify someone else's scheme

Honestly, I don't think it make much differences in kubebuilder, since in most cases, there is only one controller manager and one scheme. Thus, there is no difference between using `clientgoscheme.Scheme` and `NewScheme()`.
There is a difference when there are 2 controller managers and they want to use different schemes. But I haven't seen any use case so far.

But I agree it's a good practice to always use your own scheme.

PR #812, opened 2019-06-24
:running: use a released version of kustomize

mcristina422:
Now that this https://github.com/kubernetes-sigs/kustomize/commit/f9c631e9eec7a2d6e46eb9e1bf5122f68b97d12d is part of a release switch to the release build

### DIFF
--- a/scripts/install_and_setup.sh
+++ b/scripts/install_and_setup.sh
@@ -15,8 +15,7 @@
 # limitations under the License.
 
 go get sigs.k8s.io/kind@v0.3.0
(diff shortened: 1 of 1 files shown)


mengqiy:
@mcristina422 It will be great if you can also update https://book.kubebuilder.io/quick-start.html#installation to point to kustomize's installation document and saying it needs v2.1.0+

mcristina422:
Good point! But as I'm seeing v3-pre being released I was wondering if we should wait until the final release and make everything point to v3

PR #816, opened 2019-06-25
:sparkles: scaffold the webhook builder

mengqiy:
Add a new command `kubebuilder create webhook` for user to scaffold webhook.
Kubebuilder now dynamically register commands based on metadata in PROJECT.
If PROJECT doesn't exist or PROJECT exists with version=2, register
`create webhook`, but no `alpha webhook`.
If PROJECT exists and version=1, register `alpha webhook` but no `create
webhook`


cmd/api.go
```diff
@@ -78,6 +81,11 @@ func (o *apiOptions) runAddAPI() {
 		o.apiScaffolder.DoController = util.Yesno(reader)
 	}
 
+	if !o.controllerFlag.Changed {
+		fmt.Println("Create Webhook [y/n]")
```
DirectXMan12: Which webhooks does this create?  defaulting?  validating?  conversion?  we should spell this out in the message (especially since more "manual" webhooks might be different)

cmd/api.go
```diff
@@ -48,6 +48,9 @@ func (o *apiOptions) bindCmdFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&o.apiScaffolder.DoController, "controller", true,
 		"if set, generate the controller without prompting the user")
 	o.controllerFlag = cmd.Flag("controller")
+	cmd.Flags().BoolVar(&o.apiScaffolder.DoWebhook, "webhook", false,
```
DirectXMan12: do we want to differentiate from "manual" webhooks (i.e. not our special defaulting/validating)?

DirectXMan12:
discussed this a bit more -- we probably want a separate subcommand for enabling this

cmd/webhook_v2.go
```diff
@@ -0,0 +1,119 @@
...
+	resourcev2 "sigs.k8s.io/kubebuilder/pkg/scaffold/v2"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/v2/webhook"
+)
+
+func newWebhookV2Cmd() *cobra.Command {
+	o := webhookV2Options{}
+
+	cmd := &cobra.Command{
+		Use:   "webhook",
+		Short: "Scaffold a webhook using v2 scaffolding.",
```
droot: We don't have to reveal `v1/v2` to the end-users. So we may be we can just word it:
`Scaffold a webhook for an API or resource ?`

cmd/webhook_v2.go
```diff
@@ -0,0 +1,119 @@
...
+		Long:  `Scaffold a webhook using v2 scaffolding. You can choose to scaffold defaulting and (or) validating webhooks.`,
```
droot: Same comment as above.

cmd/webhook_v2.go
```diff
@@ -0,0 +1,119 @@
...
+		Example: `	# Create defaulting and validating webhooks for CRD of group crew, version v1 and kind FirstMate.
+	kubebuilder webhook --group crew --version v1 --kind FirstMate --defaulting --programmatic-validation
+`,
```
droot: hmm. Currently we plan to enable conversion using the same builder scaffolding. Do we scaffold differently on basis of defaulting/validation ?

cmd/webhook_v2.go
```diff
@@ -0,0 +1,119 @@
...
+			fmt.Println("Writing scaffold for you to edit...")
+
+			if len(o.res.Resource) == 0 {
+				o.res.Resource = flect.Pluralize(strings.ToLower(o.res.Kind))
+			}
+		},
+	}
+	o.res = gvkForFlags(cmd.Flags())
+	cmd.Flags().BoolVar(&o.defaulting, "defaulting", false,
+		"if scaffold the defaulting webhook")
```
droot: It should say `if set` ?

cmd/webhook_v2.go
```diff
@@ -0,0 +1,119 @@
...
+	cmd.Flags().BoolVar(&o.validation, "programmatic-validation", false,
+		"if scaffold the validating webhook")
```
droot: it should say `if set` ?

pkg/scaffold/v2/main.go
```diff
@@ -64,21 +65,25 @@ func (m *Main) Update(opts *MainUpdateOptions) error {
...
-	 }).SetupWithManager(mgr)
-	 if err != nil {
+	}).SetupWithManager(mgr); err != nil {
 	 	setupLog.Error(err, "unable to create controller", "controller", "%s")
 	 	os.Exit(1)
-	 }
+    }
 `, opts.Resource.Kind, opts.Resource.Kind, opts.Resource.Kind)
+	webhookSetupCodeFragment := fmt.Sprintf(`if err = (&%s%s.%s{}).SetupWebhookWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create webhook", "webhook", "FirstMate")
```
droot: Typo: FirstMate

testdata/project-v2/main.go
```diff
@@ -64,27 +64,32 @@ func main() {
...
+	}
+	if err = (&controllers.NamespaceReconciler{
 		Client: mgr.GetClient(),
 		Log:    ctrl.Log.WithName("controllers").WithName("Namespace"),
-	}).SetupWithManager(mgr)
-	if err != nil {
+	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "Namespace")
 		os.Exit(1)
 	}
```
droot: I think we should update our e2e tests for webhook to use new command ? and we should also add commands in generated_golden.sh to generate webhook files ?

cmd/webhook_v1.go
```diff
@@ -58,7 +58,7 @@ This command is only available for v1 scaffolding project.
 			}
 
 			if projectInfo.Version != project.Version1 {
-				fmt.Printf("webhook scaffolding is not supported for this project version: %s \n", projectInfo.Version)
+				fmt.Printf("kubebuilder alpha webhook is for project version: 1, the version of this project is: %s \n", projectInfo.Version)
```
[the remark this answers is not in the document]
DirectXMan12: :+1: from me to that.  I think we should figure out how to hide this command from `--help` if the project is v2 -- it shouldn't be that hard.

cmd/main.go
```diff
@@ -116,6 +116,7 @@ func main() {
 	rootCmd.AddCommand(
```
DirectXMan12: maybe conditionally add based on the project version, and then override help here to say "some v1-specific commands not shown"

pkg/scaffold/v2/gomod.go
```diff
@@ -43,6 +43,6 @@ module {{ .Repo }}
 go 1.12
 
 require (
-	sigs.k8s.io/controller-runtime v0.2.0-beta.2
+	sigs.k8s.io/controller-runtime da24e0fa23ff83f615764d71e7ec3eef4b893c0f
```
mengqiy: Update this when next release comes out.

mengqiy:
PTAL

mengqiy:
PTAL

cmd/create.go
```diff
@@ -18,16 +18,32 @@ package main
...
 			fmt.Println("Coming soon.")
 		},
 	}
+	cmd.AddCommand(
+		newAPICommand(),
+	)
+
+	foundProject, version := getProjectVersion()
+	// It add webhook v2 command in the following 2 cases:
+	// - There are no PROJECT file found.
```
droot: What is the reasoning for this case ? IMO, absence of project file indicates an error and we can't assume anything about the project structure at that point.

cmd/webhook_v2.go
droot: Need to update example to add `--conversion` flag.

cmd/create.go
mengqiy: Here is more about what is the default commends to show.
The reason of this choice is that if a user is starting from scratch and trying to explore what commands and flags does kubebuilder support, the user should be able to see the new one.

This piece of code is about registering the command instead of running it. When running the command, we do required PROJECT to present.

mengqiy:
> Need to update example to add --conversion flag.

I didn't add it in the existing example intentionally, since from our earlier offline discussion. `--conversion` is for advanced use case and flag description should be good enough.
I can still add a separate example for conversion flag if you believe we need it :)

I don't know why GH doesn't allow me to reply under it. :/

cmd/create.go
droot: Got it. I miss-understood earlier. Thanks for clarification.

mengqiy:
@droot @DirectXMan12  PTAL

cmd/api.go
mengqiy: Discussed with Solly offline. We can have something like `kubebuilder create webhook --low-level-defaulting --low-level-programmatic-validation` for "manual" webhooks if we want to do it in the future.

mengqiy:
Travis still has a lot of backlog now: https://www.traviscistatus.com/incidents/g9xm3l02dknv

cmd/api.go
mengqiy: Filed tracking issue https://github.com/kubernetes-sigs/kubebuilder/issues/833

mengqiy:
Should be ready to ship now.

PR #821, opened 2019-06-26
fix enable-leader-election overriden by patch.

youyongsong:
when `manager_auth_proxy_patch` is enabled, the
`--enable-leader-election` arg inside `manager/manager.yaml` will be
overriden.


mengqiy:
@youyongsong good catch!
You need to run `./generated_golden.sh` to fix the tests.

youyongsong:
> @youyongsong good catch!
> You need to run `./generated_golden.sh` to fix the tests.

Thanks for help, the tests are fixed.

PR #835, opened 2019-07-04
minor typo fix in kind string format validation

Azadehkhojandi:
kinds expected to be Pascal case


Azadehkhojandi:
I've signed CLA

mengqiy:
Please run `generated_golden.sh` to re-gen test projects.

DirectXMan12:
This prob needs to be fixed in v2 as well, right?

camilamacedo86:
Hi @DirectXMan12,
 
It is in the resource.go so shows that it used for both currently. In this way, it is quite small too and for me, it shows /lgtm 👍

PR #847, opened 2019-07-09
:book: update kubebuilder version

mengqiy:


docs/book/src/quick-start.md
```diff
@@ -16,32 +16,18 @@ os=$(go env GOOS)
...
-
-```bash
-os=$(go env GOOS)
-arch=$(go env GOARCH)
-
-# download kustomize to the kubebuilder assets folder
-curl -o /usr/local/kubebuilder/bin/kustomize -sL https://go.kubebuilder.io/kustomize/${os}/${arch}
-```
-
-Ensure that kustomize is executable:
```
mengqiy: Drop these lines, since https://sigs.k8s.io/kustomize/docs/INSTALL.md already covers that.

PR #854, opened 2019-07-12
fix typo

leodotcloud:


droot:
@leodotcloud its blocked on CLA. Thanks for the PR.

PR #783, opened 2019-06-13
:sparkles: Switched to "kustomize edit set image"

hpandeycodeit:
Fixes #754 

Switched to "images" tag as per discussion in #754.

PR #862, opened 2019-07-16
:sparkles: use kustomize edit set image

### LINKED ISSUE #754: switch to kustomize `images` field from patch
kustomize supports `images` as a first class transformer, so we should use that instead of creating a patch.

/kind feature

mengqiy:
This supersedes #783. This PR picks up the commit from #783 and squashed them. And fix the failing e2e test.

Use `kustomize edit set image` to `sed` in kubebuilder.

Fixes #754


mengqiy:
Closing in favor of #783

PR #868, opened 2019-07-19
:sparkles: Makefile: Create manifests with 'run' target

ljdelight:
The `manifests` makefile target isn't called as expected during the build process. This section from [the kubebuilder book](https://kubebuilder.io/reference/generating-crd.html) (specifically "If you examine the Makefile ... manifests target is also listed as prerequisite for other targets like run ...) implies a `make run` should create manifests. 

This issue was found because the yaml files in the cluster didn't match what the binary expected (with a make run).

### DIFF
--- a/pkg/scaffold/v2/makefile.go
+++ b/pkg/scaffold/v2/makefile.go
@@ -66,7 +66,7 @@ manager: generate fmt vet
 	go build -o bin/manager main.go
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
-run: generate fmt vet
+run: generate fmt vet manifests
 	go run ./main.go
 
 # Install CRDs into a cluster
--- a/testdata/project-v2/Makefile
+++ b/testdata/project-v2/Makefile
@@ -22,7 +22,7 @@ manager: generate fmt vet
 	go build -o bin/manager main.go
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
-run: generate fmt vet
+run: generate fmt vet manifests
 	go run ./main.go
 
 # Install CRDs into a cluster


ljdelight:
@DirectXMan12 if there's a way to fix the description removals (I mean to continue to generate the descriptions so they're not deleted with this path), please guide me.

droot:
@ljdelight 

>if there's a way to fix the description removals (I mean to continue to generate the descriptions so they're not deleted with this path), please guide me.

The generated YAML depends on the `controller-gen` in your path that is being invoked to generate the YAML files. You may be want to delete `rm -f $GOPATH/bin/controller-gen` and `go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.0-beta.3` (or whatever last stable release). 
I know this is annoying and we need to fix this.

ljdelight:
That did it @droot! The patch is much smaller

pkg/scaffold/v2/makefile.go
```diff
@@ -70,11 +70,11 @@ run: generate fmt vet
 	go run ./main.go
 
 # Install CRDs into a cluster
-install: manifests
+install: generate
```
mengqiy: Generating go code here is not useful, since we don't compile go code here.

pkg/scaffold/v2/makefile.go
```diff
@@ -70,11 +70,11 @@ run: generate fmt vet
...
 	kubectl apply -f config/crd/bases
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-deploy: manifests
+deploy: generate
```
mengqiy: Generating go code here is not useful, since we don't compile go code here.

pkg/scaffold/v2/makefile.go
ljdelight: Fixed

pkg/scaffold/v2/makefile.go
ljdelight: Fixed

mengqiy:
> When users call make the all make target is called, what's your opinion on the manifests being created during a build? 

Build doesn't require the update-to-date manifests. IMO it's not necessary, it may also slow us down for ~1 second.

ljdelight:
I was working in a freshly generated project and noticed this comment in go files: `Important: Run "make" to regenerate code after modifying this file`. If a new field is added and "make" is called, is it expected to generate manifests? Should these comments be saying to "make manifests" and "make" or what's the intention here?

mengqiy:
> Important: Run "make" to regenerate code after modifying this file

These comments are all in `types.go` files. The intend is to ensure `deepcopy` files are always up-to-date. Otherwise, go compiler may complain. Manifests are not required for building go program, and they matter only when deploying the application.

ljdelight:
Thanks a bunch, that clears things up for me. Patch updated 👍

PR #872, opened 2019-07-20
:sparkles: replace patches with patchesStrategicMerge

mengqiy:
kustomize has deprecated `patches` fields for months.
It should be replaced by `patchesStrategicMerge`.
`patches` if in old format will be auto convert to the new format in v3.0.4+. So this PR is not considering a breaking change.

Ref: https://github.com/kubernetes-sigs/kustomize/pull/1355 and kubernetes-sigs/kustomize#1363


Liujingfang1:
/hold
I’d like to make kustomize auto convert the legacy patches to new patches so it won't break kubebuilder users.

droot:
@Liujingfang1 
> I’d like to make kustomize auto convert the legacy patches to new patches so it won't break kubebuilder users.

Does that mean we are not deprecating `patches` ? because if we are deprecating, then moving to `patchesStrategicMerge` ASAP is a good idea.

Liujingfang1:
@droot We already deprecated `patches` since version 2.0 : ), but it is still parsed and works with kustomize 3.0.
So I think it's better not to break that.

mengqiy:
After merging https://github.com/kubernetes-sigs/kustomize/pull/1363, we probably still should do this PR. I will change it to `:sparkles:` since it's not breaking.

Liujingfang1:
@mengqiy sounds good to me.

mengqiy:
PTAL

PR #876, opened 2019-07-22
Change obsolete patches: section to patchesStrategicMerge:

jbrette:
`patches:` was rendered obsolete in kustomize 1.0.9. The user is supposed to use `patchesStrategicMerge:` instead ([see](https://github.com/kubernetes-sigs/kustomize/blob/v1.0.9/pkg/types/kustomization.go#L124)

The problem is that the kustomize 3.0.3 is reusing `patches:` section for more extend patching. The kustomize figured that patches had been obsoleted for long enough [see](https://github.com/kubernetes-sigs/kustomize/pull/1355#pullrequestreview-262678277)

As a side effect kubebuilder would not be able to produce kustomize compliant files.

This PR is only getting kubebuilder to use the right field.


jbrette:
This PR is identical to [PR](https://github.com/kubernetes-sigs/kubebuilder/pull/872). Closing it.

PR #912, opened 2019-08-05
:running: remove install script test from travis

droot:
We have a kubebuilder install script which is being tested by travis. Given that our current installation steps are simple, the install script is not being used, so this change drops the travis test for install-script test, which I suspect was never working.


DirectXMan12:
remove the script itself, and then this looks good

droot:
@DirectXMan12 done. PTAL.

Issue #921, opened 2019-08-07
Typo in book: int64 should be marked as allowed


daneharrigan:
The kubebuilder 2.0 book says ints should be `int32`, but `StartingDeadlineSeconds` is an `int64` in `CronJobSpec`.

/kind bug

mengqiy:
> Fields may use most of the primitive types. Numbers are the exception: for API compatibility purposes, we accept two forms of numbers: int32 for integers, and resource.Quantity for decimals.

I guess you mean this.
`int32 for integers` is a little misleading. `int64` is allowed. @DirectXMan12 We should fix it.

daneharrigan:
@mengqiy yes, that's exactly what I was referring to. If int64 is allowed, my related PR doesn't make sense anymore. I'm happy to open a new one with an update to the doc.

DirectXMan12:
I can't seem to find the spot that said "you should use int32" in the [api conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md), so yeah, we can fix the docs and the comment in the source.

daneharrigan:
@DirectXMan12 I'm having a hard time following your suggestion to comment in the source. Which source, the cronjob-tutorial or kubebuilder?

DirectXMan12:
"comment in the source" --> there's a comment in the schema generator that talks about disallowed types (right above the code that works with them), so we'll want to fix that up.

dlespiau:
Can this be closed now that both PRs are merged?

PR #922, opened 2019-08-07
Typo: CronJobSpec StartingDeadlineSeconds should be int32

daneharrigan:
Offering a fix to the issue I opened #921 921


daneharrigan:
The `project/pkg/controller/namespace` test may be flaky. I'll run the tests a few more times locally and create an issue if that's the case.

daneharrigan:
The `project/pkg/controller/namespace` package passes consistently on my local machine. I think the failure I experienced before was due to a bad default kubectl context. Please ignore my original concern about the test being flaky

DirectXMan12:
I've commented on the related issue -- namely, I can't find the "must be int32" line anymore in the k8s api conventions, so I think we can just fix this message in CT and the book instead.

PR #929, opened 2019-08-09
:book: Fix cronjob tutorial to delete and get next schedule correctly

khchan123:
- Support default parameter field
- Fix wrong deletion on SuccessfulJobsHistoryLimit & FailedJobsHistoryLimit
- Fix wrong next schedule as returned lastMissed is a pointer
- Delete dependent Job in background when deleting CronJob


khchan123:
I signed it

DirectXMan12:
Can we just move the defaulting to the defaulting webhook, please?

DirectXMan12:
otherwise seems reasonable

khchan123:
I overlooked and figured out the defaulting code already on webhook. Removed that part in the new version.

Issue #933, opened 2019-08-12
Set preserveUnknownFields to false in the CRD conversion webhook patch


mengqiy:
kube-apiserver requires `spec.preserveUnknownFields` to be `false` when `conversion.strategy` is `Webhook`.
Ref: https://github.com/kubernetes/kubernetes/blob/1d812991b63f42485d5f2aa8c883f142445fd09a/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go#L265

> CustomResourcesDefinition conversion via Web Hooks is promoted to beta. Note that you must set spec.preserveUnknownFields to false.

See release note in https://github.com/kubernetes/kubernetes/pull/78426

/kind bug

mengqiy:
cc: @droot

pires:
This seems to also affect clusters that enable `CustomResourceDefaulting` feature-gate when using `+kubebuilder:default` marker.

DirectXMan12:
yes, defaulting requires the v1 CRD and a structural schema.  We could document this better in the defaulting marker docs and the book

camilamacedo86:
/assing @camilamacedo86 

c/c @estroz

ttlv:
`
# The following patch enables conversion webhook for CRD
# CRD conversion requires k8s 1.13 or later.
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: nodedeployments.edge.harmonycloud.cn
spec:
  preserveUnknownFields: false
  conversion:
    strategy: Webhook
    webhookClientConfig:
      # this is "\n" used as a placeholder, otherwise it will be rejected by the apiserver for being blank,
      # but we're going to set it later using the cert-manager (or potentially a patch if not using cert-manager)
      caBundle: Cg==
      service:
        namespace: system
        name: webhook-service
        path: /convert
`
I think it's better to add  preserveUnknownFields: false when use kubebuilder to generate.

PR #935, opened 2019-08-13
:sparkles: set preserveUnknownFields in crd conversion patch

droot:
With Kubernetes 1.15, CRD conversion support went Beta and it requires
setting preserveUnknownFields to false. Refer to this PR
https://github.com/kubernetes/kubernetes/pull/7842 for details.

### DIFF
--- a/pkg/scaffold/v2/crd/enablewebhook_patch.go
+++ b/pkg/scaffold/v2/crd/enablewebhook_patch.go
@@ -68,4 +68,7 @@ spec:
         namespace: system
         name: webhook-service
         path: /convert
+  # With conversion webhook promoted to Beta in Kubernetes 1.15, it requires setting preserveUnknownFields to false.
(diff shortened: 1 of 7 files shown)


droot:
Putting it on hold because this is Kubernetes 1.15 related change and will be better to merge when we are ready for moving to Kubernetes 1.15.

droot:
fixes https://github.com/kubernetes-sigs/kubebuilder/issues/933

dionysius:
Any plans to pursue this further? _To easier find this issue, on `make install` with enabled webhook:_
```
The CustomResourceDefinition "my.custom.resource" is invalid: spec.conversion.strategy: Invalid value: "Webhook": must be None if spec.preserveUnknownFields is true
```

DirectXMan12:
We came up for an alternate solution for this -- see https://book.kubebuilder.io/reference/controller-gen.html

PR #943, opened 2019-08-15
Initial addons via plugin model

justinsb:
Plugin model; --pattern=addon flag that generates an addon operator
    
    We implement the initial plugin model, starting with an in-process
    implementation that can support out-of-process later.  As plugins do
    not yet form part of the stable interface for kubebuilder, we only
    enable them when the KUBEBUILDER_ENABLE_PLUGINS env var is set.
    
    We implement an initial plugin for addons: `--pattern=addon` creates a
    controller & resource that follow the style of
    https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern
    
    This style of declarative addon operators is being investigated in the
    cluster-addons subject, with code at
    https://github.com/kubernetes-sigs/addon-operators

### DIFF
--- a/pkg/model/types.go
+++ b/pkg/model/types.go
@@ -1,0 +1,55 @@
+package model
+
+import (
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/input"
+)
+
+// Universe describes the entire state of file generation
+type Universe struct {
+	Boilerplate string `json:"boilerplate,omitempty"`
+
+	Resource *Resource `json:"resource,omitempty"`
+
+	Files []*File `json:"files,omitempty"`
+}
+
+// Resource describes the resource currently being generated
+// TODO: Just use the resource type?
+type Resource struct {
+	// Namespaced is true if the resource is namespaced
+	Namespaced bool `json:"namespaces,omitempty"`
+
+	// Group is the API Group.  Does not contain the domain.
+	Group string `json:"group,omitempty"`
+
+	// Version is the API version - e.g. v1beta1
+	Version string `json:"version,omitempty"`
+
+	// Kind is the API Kind.
+	Kind string `json:"kind,omitempty"`
+
+	// Plural is the plural lowercase of Kind.
+	Plural string `json:"plural,omitempty"`
+
+	// Resource is the API Resource.
+	Resource string `json:"resource,omitempty"`
+
+	// ResourcePackage is the go package of the Resource
+	GoPackage string `json:"goPackage,omitempty"`
+
+	// GroupDomain is the Group + "." + Domain for the Resource
+	GroupDomain string `json:"groupDomain,omitempty"`
+}
+
+// File describes a file that will be written
+type File struct {
+	// Path is the file to write
+	Path string `json:"path,omitempty"`
+
+	// Contents is the generated output
+	Contents string `json:"contents,omitempty"`
+
+	// TODO: Move input.IfExistsAction into model
+	// IfExistsAction determines what to do if the file exists
+	IfExistsAction input.IfExistsAction `json:"ifExistsAction,omitempty"`
+}
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -1,0 +1,46 @@
+# Kubebuilder plugins
+
+**Status: Experimental**
+
+We are developing a plugin system to kubebuilder, so that we can generate
+operators that follow other patterns.
+
+While plugins remain experimental, you must pass the `KUBEBUILDER_ENABLE_PLUGINS=1`
+environment variable to enable plugin functionality.  (Any non-empty
+value will work!)
+
+When you specify `KUBEBUILDER_ENABLE_PLUGINS=1`, a flag `--pattern` will become
+available for resource generation.  Specifying `--pattern=addon` will change
+resource code generation to generate code that follows the addon pattern, as
+being developed in the
+[addon-operators](https://github.com/kubernetes-sigs/addon-operators)
+subproject.
+
+The `pattern=addon` plugin is intended to serve both as an example of a plugin,
+and as a real-world use case for driving development of the plugin system.  We
+don't intend for the plugin system to become an emacs competitor, but it must be
+sufficiently flexible to support the various patterns of operators that
+kubebuilder will generate.
+
+## Plugin model
+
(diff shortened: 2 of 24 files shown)


justinsb:
/assign @DirectXMan12 

I think this is what you had in mind for the plugin model, but wanted to check before it got too far along!

pkg/scaffold/v2/gomod.go
```diff
@@ -42,6 +42,8 @@ module {{ .Repo }}
 
 go 1.12
 
+replace sigs.k8s.io/kubebuilder-declarative-pattern => ../../kubebuilder-declarative-pattern
```
DirectXMan12: this shouldn't be here any more, right?

pkg/model/types.go
```diff
@@ -0,0 +1,55 @@
...
+	// ResourcePackage is the go package of the Resource
+	GoPackage string `json:"goPackage,omitempty"`
+
+	// GroupDomain is the Group + "." + Domain for the Resource
+	GroupDomain string `json:"groupDomain,omitempty"`
+}
+
+type File struct {
+	// Type is an identifier for the class of the generator
+	Type string `json:"type"`
```
DirectXMan12: what's the purpose of this?

justinsb:
Thanks for the review @DirectXMan12 .  I made the "whoopsie" changes - thanks for spotting them.  I also put `--pattern=addon` behind a (currently undocumented) feature flag `KUBEBUILDER_ENABLE_PLUGINS` - does that work?

Still lots to do here, but I really like the more strongly-typed model - I know you said you wanted to rework a lot of the existing generation logic anyway - do you think this approach could here there?

In any case, would be great to get this in so we can start e.g. adding tests and getting going on addons!

shawn-hurley:
I thought that we had talked about adding this as an experimental branch to collaborate on and then submit to master? is that not the case and I am misremembering? 

@DirectXMan12

plugins/addon/manifest.go
```diff
@@ -0,0 +1,47 @@
...
+const exampleManifestVersion = "0.0.1"
+
+const exampleManifestContents = `# Placeholder manifest - replace with the manifest for your addon
+`
+
+func ExampleManifest(u *model.Universe) error {
+	packageName := getPackageName(u)
+
+	m := &model.File{
+		Path:           filepath.Join("channels", "packages", packageName, exampleManifestVersion, "manifest.yaml"),
```
jwforres: Is there a design for manifests, channels, packages, etc?

jwforres:
1) I agree with @DirectXMan12 that this should be going to the experimental branch for kubebuilder.

2) I think this PR is overloading two sets of changes, one set of changes to introduce plugins generally, and one set of changes to propose an experimental concept for an addons plugin

justinsb:
I think this is throwing off some good changes @jwforres , so I was thinking we could keep it feature-flagged off (so it's not easily availably) but still in a place that people can try out.  For example, I think we should look at using the strongly-typed model instead of injecting objects, and whether the existing templating code then becomes simpler.  What do you think?

And I'm doing an implementation of a plugin at the same time as the plugin model - otherwise the addon interface feels a little abstract...  We can also see the problems with it - for example should we combine the controller & api phases so we can actually check that the expected file exists in a plugin.  Is there actually any value in a single kubebuilder, vs just having a separate generator CLI .. particularly as we're not really reusing much of the kubebuilder template?

I could split it into 2 commits or 2 PRs if that helps, but I am developing them together!

plugins/addon/manifest.go
justinsb: Just what we're discussing in the cluster-addons WG and https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern  We're still in development mode, I believe!

DirectXMan12:
I think not having a motivating usecase makes things a lot more abstract, so we'll want *something* to demonstrate.  We could do this cross repo, but that makes things super hard in the initial stages.

I *think* the plan is: 

- Do this, make sure it works fine, and that we're good with the interface
- Add exec, make sure that works with OSDK
- Split the addon stuff back into exec in kubebuilder-declarative-pattern again

Cross-repo iteration is *really* frustrating to work with, and a demo plugin is a bit too abstract for initial development.

That being said, I have a few more thoughts here on the generator/plugin parts.

As for feature flag vs experimental branch: feature flag is probably fine, and we've been saying "do this in the plugin framework" on the addon PR for a while now (despite being under the KB umbrella), so I think feature flag is prob fine for the moment.

jwforres:
I think what was tripping me up is that I didn't see anything obvious that indicated the addon plugin is still under development. Since @DirectXMan12 agrees with doing the feature flag I'm fine with that. Since this isn't going into experimental, can something else indicate the addons plugin is a prototype under development?

I think it brings up the general question of how do you guys want to document plugins in this repo. Would something like a README in the plugin/addons folder help describe the goal of the plugin and its current development / support state? Or maybe something under a docs/plugins/addons structure?

DirectXMan12:
sure, something indicating that it's more of an experiment is fine with me

:+1: on more docs too

justinsb:
Added a readme to the plugins folder!  The env-var to enable is currently `KUBEBUILDER_ENABLE_PLUGINS` but we could rename that to `KUBEBUILDER_ENABLE_EXPERIMENTAL_PLUGINS` if anyone is worried.

plugins/README.md
```diff
@@ -0,0 +1,46 @@
...
+resource code generation to generate code that follows the addon pattern, as
+being developed in the
+[addon-operators](https://github.com/kubernetes-sigs/addon-operators)
+subproject.
+
+The `pattern=addon` plugin is intended to serve both as an example of a plugin,
+and as a real-world use case for driving development of the plugin system.  We
+don't intend for the plugin system to become an emacs competitor, but it must be
+sufficiently flexible to support the various patterns of operators that
+kubebuilder will generate.
```
droot: emacs competitor 😆

DirectXMan12:
@briantopping yep, we'd like to have pattern information stored in the project file, it's one of the next things we want to add.  Feel free to submit a PR once this merges :-)

DirectXMan12:
/lgtm
/approve

as per discussion in the monthly meeting and the discussions on this thread

Issue #932, opened 2019-08-12
Package Root


moserke:
Add the ability to set the Go package root.  Currently `api` and `controller` is created in the root of the project.  It would be nice to be able to specify putting these packages in `pkg` to maintain good Go hygiene.

/kind feature

DirectXMan12:
I suspect this'll have to wait until after we clean up the scaffolding code -- it'd add to the mess to implement this right now.

That being said, while I'm not against supporting this, I'd like to note that "pkg" isn't necessarily the "one true way" to do Go projects.  Moving away from "pkg" was somewhat intentional -- while it was clear to Kubernetes contributors, it wasn't always clear to external users, who looked for a main.go file.

moserke:
Agreed that it's not the only way, but personally I feel like the `pkg` & `cmd` separation is becoming the idiomatic way of laying out a project.  Coming from more than just k8s repos it did feel a little out of the normal not to have that structure.  

That said though, I completely understand how serving everyone is difficult and there is never going to be the perfect solution.  Appreciate the willingness to look into supporting this.

Once the code is ready for more changes like this, I'm also happy to help implement this!  Would enjoy getting more plugged in to the community.

michal-hudy:
Can two locations be considered? Root for `api` and separate root for `controllers` and `webhooks`. In our projects, we like to have core implementation that shouldn't be imported by any other project in the `internal` package and things like `api` in `pkg`. So for us, it would be convenient to have the ability to configure it.

camilamacedo86:
HI @DirectXMan12, 

Could/Should we do it now as well or add to v3 project?

DirectXMan12:
prob v3 (or in a plugin) if we decide to do it at all.

Adirio:
@DirectXMan12 @mengqiy @droot can any of you add this to "scaffolding-v3" milestone, it seems I can't do it

DirectXMan12:
done

DirectXMan12:
you need to be a maintainer to use `/milestone` at the moment, which is a weird designation that we carried over from Kubernetes but didn't actually want to use.  Someone needs to fix the permissions so that reviewers & approvers can `/milestone` too, or something.  The current permission set is *weird*.

DirectXMan12:
This seems like a decent candidate to use as a test for plugins phase 2 -- it's mostly a matter of preference, and *should* be something that's easily doable under our general deisgn currently.

camilamacedo86:
[...]

PR #949, opened 2019-08-19
:warning: Copy all files in Dockerfile

vincepri:
This PR fixes an issue where the user Dockerfile might not work if they create packages outside `api/` or `controllers/` folders. 

With the preference for Kubebuilder v2 controller to abandon the `pkg/` directory, in [Cluster API](https://github.com/kubernetes-sigs/cluster-api) we've moved all our main packages outside of the `pkg/` folder and made it available for other repositories to use them.

### DIFF
--- a/pkg/scaffold/v2/dockerfile.go
+++ b/pkg/scaffold/v2/dockerfile.go
@@ -47,10 +47,8 @@ COPY go.sum go.sum
 # and so that source changes don't invalidate our downloaded layer
 RUN go mod download
 
-# Copy the go source
-COPY main.go main.go
-COPY api/ api/
-COPY controllers/ controllers/
+# Copy the sources
+COPY ./ ./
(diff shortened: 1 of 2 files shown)


DirectXMan12:
This'll end up cache busting on things like changed kustomize configuration, temp files, etc.  Probably not the best :-/

vincepri:
I don't disagree, but it's probably better than having a broken build after adding a new package. 

Given the comments in https://github.com/kubernetes-sigs/kubebuilder/issues/932 seems we're going away from having `pkg/`

DirectXMan12:
I'm tempted to just say here "add the rest of your files in COPY statements to your dockerfile" -- it's not too much of a burden, and it means that people still avoid cache busting.  Closing for now, but if that's a huge problem feel free to poke this issue

PR #951, opened 2019-08-20
Use gofmt, not goimports

justinsb:
goimports is slower, should not be necessary (we don't want to add
other imports), and the order of import declarations is no longer
consistent, because goimport groups them alphabetically.

goimports also means that the behaviour of kubebuilder depends on what
is on the GOPATH.

However, goimports will also respect existing import blocks, so we
remain compatible with goimports.

### DIFF
--- a/pkg/scaffold/v1/controller/controller.go
+++ b/pkg/scaffold/v1/controller/controller.go
@@ -119,6 +119,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/source"
+
 	{{ .Resource.Group}}{{ .Resource.Version }} "{{ .ResourcePackage }}/{{ .Resource.Group}}/{{ .Resource.Version }}"
 )
 
@@ -134,6 +135,7 @@ var log = logf.Log.WithName("{{ lower .Resource.Kind }}-controller")
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
+
 	{{ .Resource.Group}}{{ .Resource.Version }} "{{ .ResourcePackage }}/{{ .Resource.Group}}/{{ .Resource.Version }}"
 )
 {{ end -}}
--- a/pkg/scaffold/v1/controller/controller.go
+++ b/pkg/scaffold/v1/controller/controller.go
@@ -103,14 +103,12 @@ package {{ lower .Resource.Kind }}
 
 import (
 {{ if .Resource.CreateExampleReconcileBody }}	"context"
-	"log"
 	"reflect"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -129,6 +127,7 @@ var log = logf.Log.WithName("{{ lower .Resource.Kind }}-controller")
 
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
--- a/pkg/scaffold/v1/manager/cmd.go
+++ b/pkg/scaffold/v1/manager/cmd.go
@@ -46,14 +46,15 @@ import (
 	"flag"
 	"os"
 
+	"{{ .Repo }}/pkg/apis"
+	"{{ .Repo }}/pkg/controller"
+	"{{ .Repo }}/pkg/webhook"
+
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
-	"{{ .Repo }}/pkg/apis"
-	"{{ .Repo }}/pkg/controller"
-	"{{ .Repo }}/pkg/webhook"
 )
 
 func main() {
--- a/pkg/scaffold/v1/manager/cmd.go
+++ b/pkg/scaffold/v1/manager/cmd.go
@@ -46,15 +46,15 @@ import (
 	"flag"
 	"os"
 
-	"{{ .Repo }}/pkg/apis"
-	"{{ .Repo }}/pkg/controller"
-	"{{ .Repo }}/pkg/webhook"
-
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
+
+	"{{ .Repo }}/pkg/apis"
+	"{{ .Repo }}/pkg/controller"
+	"{{ .Repo }}/pkg/webhook"
 )
 
 func main() {
--- a/pkg/scaffold/v1/webhook/admissionbuilder.go
+++ b/pkg/scaffold/v1/webhook/admissionbuilder.go
@@ -81,6 +81,7 @@ package {{ .Type }}
 
 import (
 	{{ .Resource.Group}}{{ .Resource.Version }} "{{ .ResourcePackage }}/{{ .Resource.Group}}/{{ .Resource.Version }}"
+
 	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission/builder"
 )
--- a/pkg/scaffold/v1/webhook/admissionbuilder.go
+++ b/pkg/scaffold/v1/webhook/admissionbuilder.go
@@ -80,10 +80,10 @@ const admissionWebhookBuilderTemplate = `{{ .Boilerplate }}
 package {{ .Type }}
 
 import (
-	{{ .Resource.Group}}{{ .Resource.Version }} "{{ .ResourcePackage }}/{{ .Resource.Group}}/{{ .Resource.Version }}"
-
 	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission/builder"
+
+	{{ .Resource.Group}}{{ .Resource.Version }} "{{ .ResourcePackage }}/{{ .Resource.Group}}/{{ .Resource.Version }}"
 )
 
 func init() {
--- a/pkg/scaffold/v1/webhook/admissionhandler.go
+++ b/pkg/scaffold/v1/webhook/admissionhandler.go
@@ -83,7 +83,7 @@ import (
 	"net/http"
 
 	{{ .Resource.Group}}{{ .Resource.Version }} "{{ .ResourcePackage }}/{{ .Resource.Group}}/{{ .Resource.Version }}"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission/types"
--- a/pkg/scaffold/v1/webhook/admissionhandler.go
+++ b/pkg/scaffold/v1/webhook/admissionhandler.go
@@ -82,11 +82,11 @@ import (
 	"context"
 	"net/http"
 
-	{{ .Resource.Group}}{{ .Resource.Version }} "{{ .ResourcePackage }}/{{ .Resource.Group}}/{{ .Resource.Version }}"
-
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission/types"
+
+	{{ .Resource.Group}}{{ .Resource.Version }} "{{ .ResourcePackage }}/{{ .Resource.Group}}/{{ .Resource.Version }}"
 )
 
 func init() {
--- a/pkg/scaffold/v2/webhook/webhook.go
+++ b/pkg/scaffold/v2/webhook/webhook.go
@@ -92,10 +92,14 @@ const (
 package {{ .Resource.Version }}
 
 import (
+{{ if or .Validating .Defaulting }}
 	"k8s.io/apimachinery/pkg/runtime"
+{{ end -}}
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+{{- if or .Validating .Defaulting }}
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
+{{ end }}
 )
 
 // log is for logging in this package.
--- a/testdata/gopath/src/project/pkg/controller/firstmate/firstmate_controller.go
+++ b/testdata/gopath/src/project/pkg/controller/firstmate/firstmate_controller.go
@@ -26,7 +26,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	crewv1 "project/pkg/apis/crew/v1"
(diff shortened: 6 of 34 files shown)


justinsb:
Although the timing isn't the real reason to merge this (real reason is that goimports walks gopath)...

Before:
```
> time go test -v ./cmd/... ./pkg/... --test.count=1
real    1m57.619s
user    20m54.870s
sys     2m0.850s
```

After:
```
real    0m1.621s
user    0m9.431s
sys     0m2.471s
```

DirectXMan12:
so, there's a problem with this -- we scaffold out imports that may or may not be needed in certain cases.  We'll have to overhaul the scaffolding system, since we currently rely on goimports to do this work for us.

droot:
I am surprised to that timing difference. As Solly pointed out, we sort of exploit `goimport` to sanitize the imports in scaffolded files. Travis failures are related to that.

justinsb:
I guess my view is that if we're using goimports to remove imports, we should prefer to just not add them (some debate there about that increasing complexity of generation, we'll see...).  If we're adding imports, we _really_ shouldn't be doing that, because who knows what `goimports` is going to find on my gopath - we should just add them to the imports block.

But ... this does depend on my being able to get rid of the test failures, which are currently a mixture of the two above.

justinsb:
Travis is happy now!

DirectXMan12:
(FWIW, I'm happy to get rid of goimports if we get it working -- it's been a pain for a while)

camilamacedo86:
Hi @justinsb, 

For you be able to move forward with shows that is missing:
- rebase the branch with the master branch
- then update the testdata by running `make generate-setup` and then `make generate-testdata` to do it by using the Makefile targets.

Adirio:
This code seems to be using go1.12 yet. Please, rebase your changes to master so that changes related to the upgrade to 1.13 are applied.

testdata/project-v2/go.mod
```diff
@@ -1,11 +1,12 @@
 module sigs.k8s.io/kubebuilder/testdata/project-v2
 
-go 1.13
+go 1.12
```
camilamacedo86: Note that the master branch is using go 1.13. 
In order to collab with the project, you need to use the latest master branch and have go 1.13 locally

pkg/scaffold/v2/webhook/webhook.go
```diff
@@ -91,10 +91,14 @@ var (
 package {{ .Resource.Version }}
 
 import (
+{{ if or .Validating .Defaulting }}
```
camilamacedo86: why the ifs are required to be added here?

pkg/scaffold/v1/controller/controller.go
```diff
@@ -103,14 +103,12 @@ package {{ lower .Resource.Kind }}
...
 import (
 {{ if .Resource.CreateExampleReconcileBody }}	"context"
-	"log"
 	"reflect"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime/schema"
```
camilamacedo86: why the schema was removed?

pkg/scaffold/v2/webhook/webhook.go
Adirio: Those packages are probably only being used in case you use those flags. As before this change `goimport` tool was being used, this tool removed them, but as this PR swaps `goimport` for `go fmt` imports are not being checked any more and it would scaffold a file with two import compile errors.

Adirio:
IMHO, this PR should not be merged.

You mention some `goimport` drawbacks:
- Being slower: as you said the slowdown is related to `GOPATH`. However, `GOPATH` is only being used for `v1` which is already being deprecated. After that the whole project, scaffolded test and doc packages will be using gomod, which will remove the `GOPATH` dependency and therefore should remove the slowdown.
- Depends on `GOPATH`: not sure if you were still talking about the slowdown issue, but as explained in the previous one, `GOPATH` dependency should not be a thing in a couple months.
- Import order not inconsistency: as you already said yourself, it understands import groups, so once the groups are properly set up though the scaffold (which is a WIP) this disadvantage dissapears.

You also mention some advantages:
- Sanitizes the scaffolding: instead of having to make the scaffolder aware of which imports were already added, it handles duplicates itself.

As all the drawbacks mentioned will be outdated in 2 months time, and the advantage will still hold, I think that swaping `goimport` for `go fmt` would be a step backwards.

justinsb:
I think there's a really important one: correctness.  When we don't specify our imports and rely on goimports, we have no guarantee of what package is going to be imported.

pkg/scaffold/v2/webhook/webhook.go
justinsb: Exactly - using gofmt instead of goimports requires us to be precise about what we want to import (good) and not import (less good)

pkg/scaffold/v1/controller/controller.go
justinsb: It doesn't appear to be used - am I missing a spot where it is used?

testdata/project-v2/go.mod
justinsb: Thanks - updated!

pkg/scaffold/v1/manager/cmd.go
```diff
@@ -46,14 +46,15 @@ import (
 	"flag"
 	"os"
 
+	"{{ .Repo }}/pkg/apis"
+	"{{ .Repo }}/pkg/controller"
+	"{{ .Repo }}/pkg/webhook"
+
```
Adirio: I would leave this as a separate group but at the end. Reason: these imports are not those of kubebuilder, but of the package we are scaffolding, and thus `{{ .Repo }} ...` imports are the local package ones while the rest are external imports.

Adirio: I would leave this as a separate group but at the end. Reason: these imports are not those of kubebuilder, but of the package we are scaffolding, and thus `{{ .Repo }} ...` imports are the local package ones while the rest are external imports.

pkg/scaffold/v1/webhook/admissionbuilder.go
```diff
@@ -81,6 +81,7 @@ package {{ .Type }}
 
 import (
 	{{ .Resource.Group}}{{ .Resource.Version }} "{{ .ResourcePackage }}/{{ .Resource.Group}}/{{ .Resource.Version }}"
+
```
Adirio: Same

pkg/scaffold/v1/webhook/admissionhandler.go
```diff
@@ -83,7 +83,7 @@ import (
 	"net/http"
 
 	{{ .Resource.Group}}{{ .Resource.Version }} "{{ .ResourcePackage }}/{{ .Resource.Group}}/{{ .Resource.Version }}"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+
```
Adirio: Same

testdata/gopath/src/project/pkg/controller/firstmate/firstmate_controller.go
```diff
@@ -20,14 +20,13 @@ import (
...
 
-	crewv1 "project/pkg/apis/crew/v1"
-
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	crewv1 "project/pkg/apis/crew/v1"
```
Adirio: Should be in a new group at the end of import block, same reason as above

testdata/gopath/src/project/pkg/controller/firstmate/firstmate_controller_suite_test.go
```diff
@@ -23,11 +23,10 @@ import (
 	"sync"
 	"testing"
 
-	"project/pkg/apis"
-
 	"github.com/onsi/gomega"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
+	"project/pkg/apis"
```
Adirio: Same

testdata/gopath/src/project/pkg/controller/firstmate/firstmate_controller_test.go
```diff
@@ -20,14 +20,13 @@ import (
...
 
-	crewv1 "project/pkg/apis/crew/v1"
-
 	"github.com/onsi/gomega"
 	"golang.org/x/net/context"
 	appsv1 "k8s.io/api/apps/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	crewv1 "project/pkg/apis/crew/v1"
```
Adirio: Same

testdata/gopath/src/project/pkg/controller/frigate/frigate_controller.go
```diff
@@ -19,11 +19,10 @@ package frigate
 import (
 	"context"
 
-	shipv1beta1 "project/pkg/apis/ship/v1beta1"
-
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	shipv1beta1 "project/pkg/apis/ship/v1beta1"
```
Adirio: Same

testdata/gopath/src/project/pkg/controller/frigate/frigate_controller_suite_test.go
```diff
@@ -23,11 +23,10 @@ import (
 	"sync"
 	"testing"
 
-	"project/pkg/apis"
-
 	"github.com/onsi/gomega"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
+	"project/pkg/apis"
```
Adirio: Same

testdata/gopath/src/project/pkg/controller/frigate/frigate_controller_test.go
```diff
@@ -20,13 +20,12 @@ import (
...
 	"time"
 
-	shipv1beta1 "project/pkg/apis/ship/v1beta1"
-
 	"github.com/onsi/gomega"
 	"golang.org/x/net/context"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	shipv1beta1 "project/pkg/apis/ship/v1beta1"
```
Adirio: Same

testdata/gopath/src/project/pkg/controller/healthcheckpolicy/healthcheckpolicy_controller.go
```diff
@@ -19,11 +19,10 @@ package healthcheckpolicy
 import (
 	"context"
 
-	policyv1beta1 "project/pkg/apis/policy/v1beta1"
-
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	policyv1beta1 "project/pkg/apis/policy/v1beta1"
```
Adirio: Same

testdata/gopath/src/project/pkg/controller/healthcheckpolicy/healthcheckpolicy_controller_suite_test.go
```diff
@@ -23,11 +23,10 @@ import (
 	"sync"
 	"testing"
 
-	"project/pkg/apis"
-
 	"github.com/onsi/gomega"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
+	"project/pkg/apis"
```
Adirio: Same

testdata/gopath/src/project/pkg/controller/healthcheckpolicy/healthcheckpolicy_controller_test.go
```diff
@@ -20,13 +20,12 @@ import (
...
 	"time"
 
-	policyv1beta1 "project/pkg/apis/policy/v1beta1"
-
 	"github.com/onsi/gomega"
 	"golang.org/x/net/context"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	policyv1beta1 "project/pkg/apis/policy/v1beta1"
```
Adirio: Same

testdata/gopath/src/project/pkg/controller/kraken/kraken_controller.go
```diff
@@ -19,11 +19,10 @@ package kraken
 import (
 	"context"
 
-	creaturesv2alpha1 "project/pkg/apis/creatures/v2alpha1"
-
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	creaturesv2alpha1 "project/pkg/apis/creatures/v2alpha1"
```
Adirio: Same

testdata/gopath/src/project/pkg/controller/kraken/kraken_controller_suite_test.go
```diff
@@ -23,11 +23,10 @@ import (
 	"sync"
 	"testing"
 
-	"project/pkg/apis"
-
 	"github.com/onsi/gomega"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
+	"project/pkg/apis"
```
Adirio: Same

testdata/gopath/src/project/pkg/controller/kraken/kraken_controller_test.go
```diff
@@ -20,13 +20,12 @@ import (
...
 	"time"
 
-	creaturesv2alpha1 "project/pkg/apis/creatures/v2alpha1"
-
 	"github.com/onsi/gomega"
 	"golang.org/x/net/context"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	creaturesv2alpha1 "project/pkg/apis/creatures/v2alpha1"
```
Adirio: Same

testdata/gopath/src/project/pkg/controller/namespace/namespace_controller_suite_test.go
```diff
@@ -23,11 +23,10 @@ import (
 	"sync"
 	"testing"
 
-	"project/pkg/apis"
-
 	"github.com/onsi/gomega"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
+	"project/pkg/apis"
```
Adirio: Same

testdata/gopath/src/project/pkg/webhook/default_server/namespace/mutating/namespace_update_handler.go
```diff
@@ -21,6 +21,7 @@ import (
 	"net/http"
 
 	corev1 "k8s.io/api/core/v1"
+
```
Adirio: At the end

testdata/gopath/src/project/pkg/webhook/default_server/namespace/mutating/update_webhook.go
```diff
@@ -17,8 +17,9 @@ limitations under the License.
 package mutating
 
 import (
-	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+
+	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
```
Adirio: Why do you split these into groups?

Adirio:
> I think there's a really important one: correctness. When we don't specify our imports and rely on goimports, we have no guarantee of what package is going to be imported.

We are not relying on goimport to magically choose the package, we are just relying on it to correctly dedup imports and order them inside each group.

justinsb:
I haven't found a way to tell goimports not to add imports, and because of that we have had omissions sneak into the code e.g. https://github.com/kubernetes-sigs/kubebuilder/pull/951/files#diff-c3eb8f2983d72d8620c4a319e158ef1fR130

If we could figure out how to tell goimports to just remove & reformat - that would be great - I'd probably use that mode in my editor as well!

testdata/gopath/src/project/pkg/webhook/default_server/namespace/mutating/update_webhook.go
justinsb: I _think_ it's because corev1 is our target api in this case - interesting corner case!

justinsb:
Looks like tests are flaking, but looks like we're still discussing so not going to worry about it right now

Adirio:
> If we could figure out how to tell goimports to just remove & reformat - that would be great - I'd probably use that mode in my editor as well!

I found a way.

Passing a `nil` options pointer is equivalent to:
```
&imports.Options{
	Comments:  true,
	TabIndent: true,
	TabWidth:  8,
}
```

So if we change it for:
```
&imports.Options{
	AllErrors:  true,
	Comments:   true,
	TabIndent:  true,
	TabWidth:   8,
	FormatOnly: true,
}
```

We would neither add missing packages nor remove unused ones. We would just order and dedup each group separately.

See https://play.golang.org/p/p-mibpxju1d
The `time` package is not added and `github.com/another` is not removed. You can quickly verify that both changes would be made if you swap `FormatOnly` to `false`.

It still fails to dedup' packages found in two import groups, but that also happens with that option set to false so I guess it is an improvement.

Adirio:
@justinsb Are you ok if I include my proposed alternative in #1187? It will require to add some of the changes you have made (removing non-needed and adding missing imports)

Adirio:
@mengqiy This PR is outdated due to #1187. IMHO it can be closed.

DirectXMan12:
closing as per https://github.com/kubernetes-sigs/kubebuilder/pull/951#issuecomment-572911081

PR #952, opened 2019-08-20
Start to introduce go testing instead of ginkgo

justinsb:
* Simpler and more "go".
* Test output and failure messages are clearer.
* Able to replace some panics with better handling

### DIFF
--- a/pkg/scaffold/scaffoldtest/scaffoldtest.go
+++ b/pkg/scaffold/scaffoldtest/scaffoldtest.go
@@ -18,12 +18,13 @@ package scaffoldtest
 
 import (
 	"bytes"
+	"go/build"
 	"io"
 	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
-	"go/build"
+	"testing"
 
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
@@ -92,4 +93,47 @@ func NewTestScaffold(writeToPath, goldenPath string) (*scaffold.Scaffold, *TestR
 		r.Golden = string(b)
 	}
 	return s, r
+}
+
+// NewGoTestScaffold returns a new Scaffold and TestResult instance for testing
+func NewGoTestScaffold(t *testing.T, writeToPath, goldenPath string) (*scaffold.Scaffold, *TestResult) {
+	projRoot := getProjectRoot()
+	r := &TestResult{}
+	// Setup scaffold
+	s := &scaffold.Scaffold{
+		GetWriter: func(path string) (io.Writer, error) {
+			if path != writeToPath {
+				t.Errorf("call to GetWriter on unexpected path.  expected=%s, actual=%s", writeToPath, path)
+			}
+			return &r.Actual, nil
+		},
+		FileExists: func(path string) bool {
+			return path != writeToPath
+		},
+		ProjectPath: filepath.Join(projRoot, "testdata", "gopath", "src", "project"),
+	}
+	oldGoPath := build.Default.GOPATH
+	build.Default.GOPATH = filepath.Join(projRoot, "testdata", "gopath")
+	defer func() { build.Default.GOPATH = oldGoPath }()
+	if _, err := os.Stat(build.Default.GOPATH); err != nil {
+		t.Fatalf("error from stat on %s: %v", build.Default.GOPATH, err)
+	}
+
+	if len(goldenPath) > 0 {
+		p := filepath.Join(projRoot, "testdata", "gopath", "src", "project", goldenPath)
+		b, err := ioutil.ReadFile(p)
+		if err != nil {
+			t.Fatalf("unexpected error reading golden file %q: %v", p, err)
+		}
+		r.Golden = string(b)
+	}
+	return s, r
+}
+
+func (r *TestResult) CheckGoldenOutput(t *testing.T, actual string) {
+	if r.Golden == actual {
+		return
+	}
+
+	t.Errorf("output did not match golden output\nactual=%s\nexpected=%s", actual, r.Golden)
 }
--- a/pkg/scaffold/v1/resource/resource_test.go
+++ b/pkg/scaffold/v1/resource/resource_test.go
@@ -2,6 +2,7 @@ package resource_test
 
 import (
 	"path/filepath"
+	"testing"
 
 	"fmt"
 
@@ -124,7 +125,9 @@ var _ = Describe("Resource", func() {
 			Expect(instance.Resource).To(Equal("myresource"))
 		})
 	})
+})
 
+func TestScaffoldResources(t *testing.T) {
 	resources := []*resource.Resource{
 		{Group: "crew", Version: "v1", Kind: "FirstMate", Namespaced: true, CreateExampleReconcileBody: true},
 		{Group: "ship", Version: "v1beta1", Kind: "Frigate", Namespaced: true, CreateExampleReconcileBody: false},
@@ -133,7 +136,7 @@ var _ = Describe("Resource", func() {
 
 	for i := range resources {
 		r := resources[i]
-		Describe(fmt.Sprintf("scaffolding API %s", r.Kind), func() {
+		t.Run(fmt.Sprintf("scaffolding API %s", r.Kind), func(t *testing.T) {
 			files := []struct {
 				instance input.File
 				file     string
@@ -178,14 +181,14 @@ var _ = Describe("Resource", func() {
 
 			for j := range files {
 				f := files[j]
-				Context(f.file, func() {
-					It("should write a file matching the golden file", func() {
-						s, result := scaffoldtest.NewTestScaffold(f.file, f.file)
-						Expect(s.Execute(scaffoldtest.Options(), f.instance)).To(Succeed())
-						Expect(result.Actual.String()).To(Equal(result.Golden), result.Actual.String())
-					})
+				t.Run(fmt.Sprintf("file %s", f.file), func(t *testing.T) {
+					s, result := scaffoldtest.NewGoTestScaffold(t, f.file, f.file)
+					if err := s.Execute(scaffoldtest.Options(), f.instance); err != nil {
+						t.Fatalf("error from Execute: %v", err)
+					}
+					result.CheckGoldenOutput(t, result.Actual.String())
 				})
 			}
 		})
 	}
-})
+}


DirectXMan12:
err... I'd like to have a discussion on this.  I really find Go's built in testing framework infuriating to work with, so I'd like to discuss this a bit.

DirectXMan12:
What don't you find clear with the ginkgo output?

negz:
Is there an issue tracking this proposal? I'd strongly support kubebuilder being less opinionated about how tests should be written - we've spent a good amount of time in Crossplane unwinding gomega and replacing them with table driven tests. Given that any choice of framework will bother some people the stdlib framework seems like a reasonable lowest common denominator.

https://golang.org/doc/faq#testing_framework

DirectXMan12:
There's not currently a proposal.  I'll warn in advanced that I'm fairly against recommending standard Go unit tests -- I'll try to outline my thoughts a bit below:

DirectXMan12:
First, to start out: the [design of KubeBuilder](https://github.com/kubernetes-sigs/kubebuilder/blob/master/DESIGN.md#kubebuilder) is to have opinions.  We try to make those general enough to be mostly acceptable to everyone, but sometimes we take approaches that aren't universal.  That's ok, although we'll try to fix things if something is actively detrimental.

One of the other goals of KubeBuilder is to try and gently guide people in the direction of good practices.  This is a hard task, and the definition of "good practices" is often inherently fuzzy, especially in the implementation.  For instance, while almost everyone agrees that you should have tests, what those tests look like is up for debate ;-)

Here's where we encounter the first issue with Go unit tests, IMO: it's *really* easy to write poor tests, and you have to be fairly deliberate to write ones that function well.  For instance, this code looks suspiciously like a good test:

```go
if actual != expected {
  t.Errorf("expected actual (%#v) to be expected (%#v)", actual, expected)
}
```

It's probably fine, assuming actual and expected can be compared with `=` properly (otherwise we'll need to break out `reflect.DeepEqual` or somesuch), but when it fails, the error output is almost certainly bad, since `%#v` doesn't expand pointers.  Plus, you had to think about writing that error message.  Just as often, I've seen

```go
if actual != expected {
  t.Errorf("actual foo wasn't expected foo")
}
```

which doesn't even tell you what broke (I've seen both of those quite a bit in table-driven tests too). 
 Plus, testing a condition is a quite a bit of typing that's probably fairly repetitive. 

Contrast this with

```go
Expect(actual).To(Equal(expected))
```

We get the deepequal for free w/o an extra import, don't really have to think about writing at least an acceptable error message, and have a lot less typing.  ***It reduces the minimum cognitive load for writing the tests, letting you focus on actually testing the thing in question.***

Let's not even get started on attempting to test things with timeouts or asynchronicity. 

So, at this point, you probably want an assertion library (which generally means either gomega or testify).

So why Ginkgo at this point, and not just standard `TestXYZ`?

- BDD-style tests encourage writing down desired behaviors, as english (`It` and `By` really help here), which helps you think about what and how to test, and helps define the expected behavior more clearly

- Grouped Ginkgo tests encourage a more logical organization of your tests into coherent chunks, while still communicating that information in the output.  For instance, you can (and totally should) group your native-Go tests, but in order to communicate that, you really need to use comments (which people often don't).  That comment information doesn't show up in the test results, so there's often context lost when reading the output.  Same with steps within a test, which are often grouped via comments, but could be grouped using something like `By` to communicate those groupings in the output.

negz:
> the design of KubeBuilder is to have opinions.

I can see your point, but I would expect a project that helps me scaffold and build Kubernetes controllers to have opinions about how I write Kubernetes controllers, and that those might extend to how I test Kubernetes controllers (i.e. libraries and utilities for testing Kubernetes controllers), but having opinions about how I should be testing my software in general feels like a bit of scope creep.

If this is something kubebuilder is bullish on, that's fine. Projects (including Crossplane) that don't agree with kubebuilder about how tests should be written can remove any scaffolded tests and DIY.

DirectXMan12:
For instance, consider the following real set of tests for part of horizontal pod autoscaler: https://github.com/kubernetes/kubernetes/blob/9243fcdd656bc84d0a48e3575f5206f982fe8538/pkg/controller/podautoscaler/horizontal_test.go

Let's take a look at the names test names and comments only (no helpers):

<details>
<summary>77 test names or so</summary>

```go
func TestScaleUp(t *testing.T) {
func TestScaleUpUnreadyLessScale(t *testing.T) {
func TestScaleUpHotCpuLessScale(t *testing.T) {
func TestScaleUpUnreadyNoScale(t *testing.T) {
func TestScaleUpHotCpuNoScale(t *testing.T) {
func TestScaleUpIgnoresFailedPods(t *testing.T) {
func TestScaleUpIgnoresDeletionPods(t *testing.T) {
func TestScaleUpDeployment(t *testing.T) {
func TestScaleUpReplicaSet(t *testing.T) {
func TestScaleUpCM(t *testing.T) {
func TestScaleUpCMUnreadyAndHotCpuNoLessScale(t *testing.T) {
func TestScaleUpCMUnreadyandCpuHot(t *testing.T) {
func TestScaleUpHotCpuNoScaleWouldScaleDown(t *testing.T) {
func TestScaleUpCMObject(t *testing.T) {
func TestScaleUpFromZeroCMObject(t *testing.T) {
func TestScaleUpFromZeroIgnoresToleranceCMObject(t *testing.T) {
func TestScaleUpPerPodCMObject(t *testing.T) {
func TestScaleUpCMExternal(t *testing.T) {
func TestScaleUpPerPodCMExternal(t *testing.T) {
func TestScaleDown(t *testing.T) {
func TestScaleUpOneMetricInvalid(t *testing.T) {
func TestScaleUpFromZeroOneMetricInvalid(t *testing.T) {
func TestScaleUpBothMetricsEmpty(t *testing.T) { // Switch to missing
func TestScaleDownStabilizeInitialSize(t *testing.T) {
func TestScaleDownCM(t *testing.T) {
func TestScaleDownCMObject(t *testing.T) {
func TestScaleDownToZeroCMObject(t *testing.T) {
func TestScaleDownPerPodCMObject(t *testing.T) {
func TestScaleDownCMExternal(t *testing.T) {
func TestScaleDownToZeroCMExternal(t *testing.T) {
func TestScaleDownPerPodCMExternal(t *testing.T) {
func TestScaleDownIncludeUnreadyPods(t *testing.T) {
func TestScaleDownOneMetricInvalid(t *testing.T) {
func TestScaleDownOneMetricEmpty(t *testing.T) {
func TestScaleDownIgnoreHotCpuPods(t *testing.T) {
func TestScaleDownIgnoresFailedPods(t *testing.T) {
func TestScaleDownIgnoresDeletionPods(t *testing.T) {
func TestTolerance(t *testing.T) {
func TestToleranceCM(t *testing.T) {
func TestToleranceCMObject(t *testing.T) {
func TestToleranceCMExternal(t *testing.T) {
func TestTolerancePerPodCMObject(t *testing.T) {
func TestTolerancePerPodCMExternal(t *testing.T) {
func TestMinReplicas(t *testing.T) {
func TestZeroMinReplicasDesiredZero(t *testing.T) {
func TestMinReplicasDesiredZero(t *testing.T) {
func TestZeroReplicas(t *testing.T) {
func TestTooFewReplicas(t *testing.T) {
func TestTooManyReplicas(t *testing.T) {
func TestMaxReplicas(t *testing.T) {
func TestSuperfluousMetrics(t *testing.T) {
func TestMissingMetrics(t *testing.T) {
func TestEmptyMetrics(t *testing.T) {
func TestEmptyCPURequest(t *testing.T) {
func TestEventCreated(t *testing.T) {
func TestEventNotCreated(t *testing.T) {
func TestMissingReports(t *testing.T) {
func TestUpscaleCap(t *testing.T) {
func TestUpscaleCapGreaterThanMaxReplicas(t *testing.T) {
func TestConditionInvalidSelectorMissing(t *testing.T) {
func TestConditionInvalidSelectorUnparsable(t *testing.T) {
func TestConditionFailedGetMetrics(t *testing.T) {
func TestConditionInvalidSourceType(t *testing.T) {
func TestConditionFailedGetScale(t *testing.T) {
func TestConditionFailedUpdateScale(t *testing.T) {
func TestNoBackoffUpscaleCM(t *testing.T) {
func TestNoBackoffUpscaleCMNoBackoffCpu(t *testing.T) {
func TestStabilizeDownscale(t *testing.T) {
func TestComputedToleranceAlgImplementation(t *testing.T) {
func TestScaleUpRCImmediately(t *testing.T) {
func TestScaleDownRCImmediately(t *testing.T) {
func TestAvoidUncessaryUpdates(t *testing.T) {
func TestConvertDesiredReplicasWithRules(t *testing.T) {
func TestNormalizeDesiredReplicas(t *testing.T) {
func TestScaleUpOneMetricEmpty(t *testing.T) {
func TestNoScaleDownOneMetricInvalid(t *testing.T) {
func TestNoScaleDownOneMetricEmpty(t *testing.T) {
 ```

</details>

I've got a new test for scaling up.  Where do I put it?  Does it go with the `TestScaleUpYXZ` at the beginning?  With the `TestUpscaleXYZ`?  With the `TestScaleUpRCImmediately`?  What about if we want to check to see if a test for our behavior already exists?  Where do we look?  This could be fixed, but it requires a number of deliberate nit-style code reviews when new tests are merged.

Let's say we fixed it by sorting them, and adding comments above each group.  That's fine, except that now we've got a bunch of duplicate things we had to type over-and-over, and a grouping (via comments) that only shows up in the source.  This grouping gets lost in the test output, meaning we need an extra step (finding the comment above a group) to start tracking down issues.

As a final exercise, using these test names, find the desired behavior of the horizontal pod autoscaler.  It's not easy.  Even with the test code, it's not easy.  These tests make sure things don't break when we refactor, but they certainly don't tell us very well as humans what the intention is.

Granted, the HPA tests are far from perfect (and I'm picking on them because I wrote or rewrote quite a few).  The practical reality, however, is that most people write tests like this by default.  Sure, you can have perfect Go-style tests that have nice titles and good grouping (although I still maintain that camel-case tests names are a pain to ready for real sentences), but enforcing that requires a lot more effort on the part of the test writer and reviewer.  Similarly, it's certainly possible to write tests with bad titles or that are poorly grouped with Ginkgo.  The difference, however, is that Ginkgo nudges you in the right direction, and makes it just a bit more likely that you'll "accidentally" start writing better-organized and better-titled tests.

DirectXMan12:
Whoops, post race :-P

> but having opinions about how I should be testing my software in general feels like a bit of scope creep.

I'd argue that we kinda *have* to have opinions here if we want to scaffold out tests at all.  Either we're of the opinion that the Go tests are good, or we're of the opinion that they're not, but both are an opinion, because scaffolding out Go tests isn't useful to people who want to use Ginkgo, and vice versa.

Furthermore, how you test your controller is inherently tied to how you write tests for your controller-based software -- if the recommendation is "Set up input conditions, run controller, await desired final conditions", then sooner or later someone's going to come along and ask for a helper for "await output conditions".  At that point, we either have to say "just use `Eventually`" or provide a helper that's equivalent to `Eventually` w/o Ginkgo.

>  Projects (including Crossplane) that don't agree with kubebuilder about how tests should be written can remove any scaffolded tests and DIY.

That's kinda where I'm at currently -- if you know what you're doing, feel free to blaze your own trail on certain bits.  If you don't, we provide some defaults that you can follow along with.

DirectXMan12:
delivered rant.  no further discussion seen.  closing.

PR #953, opened 2019-08-20
Replace other ginkgo in pkg/scaffold/v1/resource

justinsb:
We use the grid pattern; it is more go.

As we replace ginkgo, we'll likely move more tests out of the grid into
specialized tests like TestValidateVersion, which will make the intent of
the tests clearer still.


DirectXMan12:
same with #952 .  Let's have a discussion in one place

/hold

DirectXMan12:
closing per the above and #952.  Comment if you disagree

PR #967, opened 2019-08-25
fix range value

toshi0607:
# What
* fix bug

# Why
* remove`_` because `range` that has 2 return values returns index first.
* `range` that has 1 return value returns value only.


toshi0607:
I signed it

PR #968, opened 2019-08-25
scaffold/v2/group: fix duplicate import package name and sort order

zchee:
Fix duplicate import package name and sort package order.

The `k8s.io/apimachinery/pkg/runtime/schema` and `sigs.k8s.io/controller-runtime/pkg/scheme` are the same package name. Fix it use the `runtimeschema` custom import package name to `k8s.io/apimachinery/pkg/runtime/schema` package.

Also, When running goimports to generated `groupversion_info.go`, will change the import order of packages that `k8s.io/apimachinery/pkg/runtime/schema` is the first.

### DIFF
--- a/pkg/scaffold/v2/group.go
+++ b/pkg/scaffold/v2/group.go
@@ -55,14 +55,14 @@ var groupTemplate = `{{ .Boilerplate }}
 package {{ .Resource.Version }}
 
 import (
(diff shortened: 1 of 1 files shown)


zchee:
ah, not dups package name. ignore it :P
close.

PR #970, opened 2019-08-29
Fix the API conventions link

alenkacz:
The contributor repo got re-arranged to be per-sig, so we need to adjust accordingly.


DirectXMan12:
good catch, thanks!

PR #975, opened 2019-09-01
Add int64 to numbers section in api-design doc

daneharrigan:
This PR is to address the typo in #921


DirectXMan12:
Can you follow up with a fix to https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/schema.go#L393 as well?

PR #976, opened 2019-09-01
clarify usage text on init --repo flag

rosenhouse:
As best I can tell, `--repo` is used to set go module and package names. It doesn't need to be hosted on github.


DirectXMan12:
yep.  That documentation was misleading.

/lgtm
/approve

Issue #954, opened 2019-08-21
Allow group names with `-`


camilamacedo86:
## Currently:

```
 $ kubebuilder create api --group webapp-hyphen --version v1 --kind Guestbook
Create Resource [y/n]
y
Create Controller [y/n]
y
Writing scaffold for you to edit...
api/v1/guestbook_types.go
2019/08/15 23:10:22 error scaffolding APIs: group must match ^[a-z]+$ (was webapp-hyphen)
```

## RFE
However, the controller allows it. The change is just in the regex ^[a-z]+$.

camilamacedo86:
PR: https://github.com/kubernetes-sigs/kubebuilder/pull/977

dlespiau:
It seems this issue should be closed now that #977 is merged.

dlespiau:
Thank you!

PR #977, opened 2019-09-02
Allow group names with dash

camilamacedo86:
## Motivation:
- https://github.com/kubernetes-sigs/kubebuilder/issues/954

## What 
allow the usage of dash in the group names

## Why 

Allow users to create API's with a group name like `mypapp-operator.example.com`. Note that the controller allows it.

### DIFF
--- a/pkg/scaffold/v1/resource/resource.go
+++ b/pkg/scaffold/v1/resource/resource.go
@@ -24,6 +24,8 @@ import (
 	"github.com/gobuffalo/flect"
 )
 
+const GroupMatchRegex = "^[a-z-]+$"
+
 // Resource contains the information required to scaffold files for a resource.
 type Resource struct {
 	// Namespaced is true if the resource is namespaced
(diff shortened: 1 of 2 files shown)


pkg/scaffold/v1/resource/resource.go
```diff
@@ -24,6 +24,8 @@ import (
 	"github.com/gobuffalo/flect"
 )
 
+const GroupMatchRegex = "^[a-z -]+$"
```
DirectXMan12: shouldn't this be `^[a-z-]+$` or `^[a-z\-]+$`?  Otherwise we'll include whitespace in the allowed set.

camilamacedo86: Really tks. Good point. Change requested performed.

DirectXMan12:
/lgtm
/approve

we'll want to tighten this up eventually to match kube's validation (DNS subdomain label), but this is fine for now.

Issue #883, opened 2019-07-24
OSDK->KB: Non Root Images


DirectXMan12:
Set a non-root user on the scaffolded images

/kind feature

camilamacedo86:
PR for it: https://github.com/kubernetes-sigs/kubebuilder/pull/983

camilamacedo86:
Hi @mengqiy and @DirectXMan12,

It is done. Should we not close this one?

PR #983, opened 2019-09-10
Setup user on docker image to run it as no root

camilamacedo86:
## What:
- Setup user on docker image to run it as rootless

## Motivation:
- https://github.com/kubernetes-sigs/kubebuilder/issues/883
- https://jira.coreos.com/browse/OSDK-473

## Steps performed to check it locally

```
$ cd upimage/
$ kubebuilder init --domain up.image
$ kubebuilder create api --group kubeimage --version v1 --kind TestImg
$ make
$ oc login -u system:admin
$ make install
$ make docker-build docker-push IMG=cmacedo/upimage:test
$ make deploy IMG=cmacedo/upimage:test
```
Checked that it worked in Minishift/OCP as well. 

Screenshot 2019-09-17 at 17 03 57
Screenshot 2019-09-17 at 17 03 40
Screenshot 2019-09-17 at 17 03 34
Screenshot 2019-09-17 at 17 03 27

### DIFF
--- a/pkg/scaffold/v2/dockerfile.go
+++ b/pkg/scaffold/v2/dockerfile.go
@@ -57,8 +57,10 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager 
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:latest
+FROM gcr.io/distroless/static:nonroot 
 WORKDIR /
 COPY --from=builder /workspace/manager .
+USER nonroot:nonroot
+
 ENTRYPOINT ["/manager"]
 `
--- a/pkg/scaffold/v2/main.go
+++ b/pkg/scaffold/v2/main.go
@@ -166,6 +166,7 @@ func main() {
 		Scheme:             scheme,
 		MetricsBindAddress: metricsAddr,
 		LeaderElection:     enableLeaderElection,
+		Port:               9443, 
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
--- a/pkg/scaffold/v2/webhook_manager_patch.go
+++ b/pkg/scaffold/v2/webhook_manager_patch.go
@@ -47,7 +47,7 @@ spec:
       containers:
       - name: manager
         ports:
-        - containerPort: 443
+        - containerPort: 9443
           name: webhook-server
           protocol: TCP
         volumeMounts:
--- a/testdata/project-v2/config/webhook/service.yaml
+++ b/testdata/project-v2/config/webhook/service.yaml
@@ -7,6 +7,6 @@ metadata:
 spec:
   ports:
     - port: 443
-      targetPort: 443
+      targetPort: 9443
   selector:
     control-plane: controller-manager
--- a/pkg/scaffold/v2/webhook/service.go
+++ b/pkg/scaffold/v2/webhook/service.go
@@ -48,7 +48,7 @@ metadata:
 spec:
   ports:
     - port: 443
-      targetPort: 443
+      targetPort: 9443
   selector:
     control-plane: controller-manager
 `
--- a/testdata/project-v2/config/default/manager_webhook_patch.yaml
+++ b/testdata/project-v2/config/default/manager_webhook_patch.yaml
@@ -9,7 +9,7 @@ spec:
       containers:
       - name: manager
         ports:
-        - containerPort: 443
+        - containerPort: 9443
           name: webhook-server
           protocol: TCP
         volumeMounts:
--- a/testdata/project-v2/main.go
+++ b/testdata/project-v2/main.go
@@ -58,6 +58,7 @@ func main() {
 		Scheme:             scheme,
 		MetricsBindAddress: metricsAddr,
 		LeaderElection:     enableLeaderElection,
+		Port:               9843,
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")


pkg/scaffold/v2/dockerfile.go
```diff
@@ -60,5 +68,12 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager
...
-ENTRYPOINT ["/manager"]
-`
+COPY build/bin /usr/local/bin
+
+ENV OPERATOR=/workspace/manager \
+    USER_UID=1001 \
+    USER_NAME=manager
+
+RUN  /usr/local/bin/user_setup
+ENTRYPOINT ["/usr/local/bin/entrypoint"]
```
DirectXMan12: What's up with the rest of these changes?

camilamacedo86: Hi @DirectXMan12, 

Really tks for your time. But it still in WIP  so I did not think that you would review it. Few changes were made, please feel free to check them. However, shows that I am still with some problem to exec the /manager in the Entrypoint.

camilamacedo86:
Hi @DirectXMan12,

It was tested properly with the branch `test-image` in this project "https://github.com/camilamacedo86/memcached-kubebuilder/tree/test-image". Note that it is working in Minishift/OCP too. 

Screenshot 2019-09-11 at 12 56 34

Screenshot 2019-09-11 at 12 58 18

Screenshot 2019-09-11 at 12 58 40

pkg/scaffold/v2/dockerfile.go
joelanford: Just some background on this:

By default in OpenShift, containers are run as uid=\<random UID\>, gid=0

The user setup script sets up unix perms so that gid=0 processes can access the user's home directory and update `/etc/passwd`.

When the container starts, the custom entrypoint adds an entry to `/etc/passwd` if the user is not already present, which will typically be the case since the user setup script doesn't actually create the user.

The outcomes of these are:
1. In non-OpenShift clusters, the operator is executed as `manager:root` (`1001:0`)
2. In OpenShift clusters the operator is executed as `manager:root` (`<random uid>:0`)

In both cases, the operator files and executables are owned by `1001:0` and have permissions that allow gid=0 to read/write/execute as necessary, thus allowing random uids to run correctly as long as they're in the `root` group.

joelanford: This is how it currently works in SDK, but we may be able to simplify this significantly.

DirectXMan12: ah, yeah.

So, distroless/static has a "nonroot" user that already is set up with a passwd entry, etc.  We can make use of that.  Since we're in static, we don't have (or want) things like a shell.

mengqiy:
> I do not know why it is not passing in 1 of the tests. See here.

The change probably breaks the [v2 e2e test](https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/e2e_v2.go).

DirectXMan12:
yep, that's my suspicion as well.  this needs to be overhauled re: my comment before we move forward

pkg/scaffold/v2/dockerfile.go
camilamacedo86: Really tks. The changes suggested are done and all shows working fine in the OCO without the shell as well. 

c/c @joelanford @DirectXMan12

camilamacedo86:

Could you give me a hand to understand why the change here which is very small now still breaking the 1 of the tests? 

The error faced is because it tries to delete the `certificates.certmanager.k8s.io` but it is not found. 

However, has not any error in the step where it is applied and I cannot see what in this change could affect it. See [here](https://travis-ci.org/kubernetes-sigs/kubebuilder/jobs/586131387#L655) that the error in the  `uninstalling cert manager bundle` in [this](https://github.com/kubernetes-sigs/kubebuilder/blob/7be1de3909425ca2cdbe39f4ab461958d9dc86da/test/e2e/e2e_v2.go#L49) line of code.

mengqiy:
> kubectl -n e2e-xdvj-system apply -f config/samples/barxdvj_v1alpha1_fooxdvj.yaml failed with error: Error from server (InternalError): error when creating "config/samples/barxdvj_v1alpha1_fooxdvj.yaml": Internal error occurred: failed calling webhook "mfooxdvj.kb.io": Post https://e2e-xdvj-webhook-service.e2e-xdvj-system.svc:443/mutate-barxdvj-example-comxdvj-v1alpha1-fooxdvj?timeout=30s: dial tcp 10.96.220.66:443: connect: connection refused

It seems apiserver can't talk to the webhook. Let me take a closer look.

mengqiy:
> 2019-09-17T18:45:06.408Z	ERROR	setup	problem running manager	{"error": "listen tcp :443: bind: permission denied"}

After looking at the container log, it seems it no longer have permissions to bind to 443, which is the default port for webhook.

mengqiy:
To fix this issue, you will need to change the scaffolding of
- main.go: set `Port` in `manager.Options`.
- config/default/manager_webhook_patch.yaml: change `containerPort`.
- config/webhook/service.yaml: change `targetPort`.

DirectXMan12:
ah, yeah.  non-root and privileged ports and all that.

docs/book/src/cronjob-tutorial/testdata/project/config/webhook/service.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   namespace: system
 spec:
   ports:
-    - port: 443
```
mengqiy: Only change `targetPort` in the service, but don't change `port`.
Since `port` is required to be 443 in admissionregistion/v1beta1. IIRC it is relaxed in admissionregistion/v1 (~1.15), but we still use v1beta1.

testdata/project-v2/config/webhook/service.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   namespace: system
 spec:
   ports:
-    - port: 443
```
mengqiy: Same comment here.

docs/book/src/cronjob-tutorial/testdata/project/config/webhook/service.yaml
camilamacedo86: Thank you for your help with.

pkg/scaffold/v2/dockerfile.go
camilamacedo86: closing it since shows all sorted.

camilamacedo86:
Hi @mengqiy, 

Really tks for your help with. It is done as you suggested but the same issue still been faced. See [here](https://travis-ci.org/kubernetes-sigs/kubebuilder/jobs/586604103#L700).

pkg/scaffold/v2/webhook_manager_patch.go
```diff
@@ -47,7 +47,7 @@ spec:
       containers:
       - name: manager
         ports:
-        - containerPort: 443
+        - containerPort: 8443
```
mengqiy: Looking at the manager container log, I realized 8443 is already taken by auth proxy container.
See: https://github.com/kubernetes-sigs/kubebuilder/blob/76963b330e697c92399b1f0dace5f24c736b0768/testdata/project-v2/config/default/manager_auth_proxy_patch.yaml#L20

mengqiy: Changing it to a different port should fix the failing test.

camilamacedo86: It is true .. I could check it now .. tks .. it is one of my first contribution here .. so I am still learning how the things are working ..

pkg/scaffold/v2/dockerfile.go
```diff
@@ -57,8 +57,10 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager
...
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:latest
+FROM gcr.io/distroless/static:nonroot 
 WORKDIR /
 COPY --from=builder /workspace/manager .
-ENTRYPOINT ["/manager"]
-`
+EXPOSE 443
+EXPOSE 8443
```
mengqiy: ?

pkg/scaffold/v2/dockerfile.go
```diff
@@ -57,8 +57,10 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager
...
+USER nonroot:nonroot
+ENTRYPOINT ["/manager"]`
```
mengqiy: I'd keep a newline here.

pkg/scaffold/v2/main.go
```diff
@@ -166,6 +166,7 @@ func main() {
 		Scheme:             scheme,
 		MetricsBindAddress: metricsAddr,
 		LeaderElection:     enableLeaderElection,
+        Port:               9443,
```
mengqiy: Please fix the indentation.

pkg/scaffold/v2/dockerfile.go
camilamacedo86: Trying to check if by exposing the port we could solve this issue.

camilamacedo86: To expose the port and solve the issue.

camilamacedo86: Just testing

pkg/scaffold/v2/dockerfile.go
camilamacedo86: 👍

pkg/scaffold/v2/main.go
camilamacedo86: 👍

pkg/scaffold/v2/main.go
```diff
@@ -166,6 +166,7 @@ func main() {
 		Scheme:             scheme,
 		MetricsBindAddress: metricsAddr,
 		LeaderElection:     enableLeaderElection,
+		Port:				9443,
```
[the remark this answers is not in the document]
camilamacedo86: Done 👍

camilamacedo86:
Hi @mengqiy, 

Shows that all requests are done. Really tks for the help, to tell how to solve the test issue and time spend in the review.

PR #998, opened 2019-09-17
[WIP] :sparkles: use a non-default service account

### LINKED ISSUE #454: Separate kubernetes service account for controller manager
Currently generated yaml for controller manager rely on default kubernetes service account that is created in namespace.
Maybe it is better to use separate Service Account for role binding.

mengqiy:
fixes #454


DirectXMan12:
for posterity: it's unclear to me if we actually need this -- we scaffold out as if each manager gets its own namespace, so the the default SA should be fine.  If something else is starting pods in your NS, this won't help, since they can just run with whatever SA we choose.

PR #1016, opened 2019-09-24
scaffold: Don't hardcode the boilerplate path in the Makefile

dlespiau:
init has a --path argument that specifies where the boilerplate file should be
written to. Unfortunately, the Makefile generation wasn't taking it into
account.

### DIFF
--- a/pkg/scaffold/v2/makefile.go
+++ b/pkg/scaffold/v2/makefile.go
@@ -94,7 +94,7 @@ vet:
 
 # Generate code
 generate: controller-gen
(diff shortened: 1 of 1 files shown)


dlespiau:
Unfortunately `kubebuilder create api` then fails with:

```
2019/09/24 17:56:13 error scaffolding APIs: open hack/boilerplate.go.txt: no such file or directory
```

I guess we should store some state in `PROJECT` to be shared between sub-command invocation.

DirectXMan12:
yeah, preserving those in the project file seems reasonable

Issue #1040, opened 2019-09-26
Use the controller util validation for group names instead of a regex.


camilamacedo86:
/kind feature

WHAT: Use the https://godoc.org/k8s.io/apimachinery/pkg/util/validation#IsDNS1123Subdomain to check if the group informed is valid or not
WHERE: https://github.com/kubernetes-sigs/kubebuilder/blob/178d59db212f0d09bf03d6e124976d58f4e02595/pkg/scaffold/v1/resource/resource.go#L69

```go
if err := validation.IsDNS1123Subdomain(r.Group); err != nil {
		return fmt.Errorf("group name is invalid: %v", err)
	}
```

camilamacedo86:
Closing it with the merge of the PR.

Issue #1037, opened 2019-09-26
Support numbers in group names


nrvnrvn:
## Currently:
```
$ kubebuilder create api --group k8s --version v1alpha1 ...
Create Resource [y/n]
y
Create Controller [y/n]
y
Writing scaffold for you to edit...
api/v1alpha1/redis_types.go
2019/09/25 18:22:12 error scaffolding APIs: group must match ^[a-z-]+$ (was k8s)
```

## Why
Allow users create group names like `k8s.example.com`

/kind feature

DirectXMan12:
we'll need to match whatever kubernetes allows.  I think the answer for that is DNS subdomain, so we should switch to that.

/help

just verify that constraint in the kubernetes docs, and then plumb it through KB

nrvnrvn:
Here are the findings:
https://github.com/kubernetes/apiserver/blob/af8f81df54d7d7a4fbd7634bf54becb5802554af/pkg/apis/audit/validation/validation.go#L102-L108
https://github.com/kubernetes/apimachinery/blob/master/pkg/api/validation/generic.go#L36-L41
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L141-L150

I could open up a PR if it looks ok at first glance and discuss it further there?

camilamacedo86:
The PR https://github.com/kubernetes-sigs/kubebuilder/pull/1048 would solve it. 
/assign @camilamacedo86

camilamacedo86:
Hi @DirectXMan12 and @nrvnrvn,

I understand that we can close it with the merge of PR  #1048. WDYT?

nrvnrvn:
I would say - yes.
But I would also check if there are any more places where the api machinery’s validation could be leveraged in the same manner.

camilamacedo86:
Hi @nrvnrvn, 

So far, just the group is the place which we identify this need. Also, note that each issue/task would deal with a problem/subject/question. So, in this issue, you did an RFE in order to be allowed to `kubebuilder create api --group k8s --version v1alpha1 ...`  Are you able to do it with the master branch? If yes, then IMO we should close this one. 

However, see that we are free to raise new issues/tasks for each need/requirement/problem that we may face. 

Really thank you for your attention and support this matter.

DirectXMan12:
we should file separate follow-ups for separate instances -- it's easier to track that way

camilamacedo86:
[...]

PR #1048, opened 2019-09-30
Replace group name regex validation for k8s framework validation

camilamacedo86:
**Description**
Replace group name regex validation for k8s framework validation

**Motivation**
https://github.com/kubernetes-sigs/kubebuilder/issues/1040
https://github.com/kubernetes-sigs/kubebuilder/issues/1037

### DIFF
--- a/pkg/scaffold/v1/resource/resource.go
+++ b/pkg/scaffold/v1/resource/resource.go
@@ -24,8 +24,6 @@ import (
 	"github.com/gobuffalo/flect"
 )
 
-const GroupMatchRegex = "^[a-z-]+$"
-
 // Resource contains the information required to scaffold files for a resource.
 type Resource struct {
 	// Namespaced is true if the resource is namespaced
@@ -70,9 +68,8 @@ func (r *Resource) Validate() error {
 		r.Resource = flect.Pluralize(strings.ToLower(r.Kind))
 	}
 
-	groupMatch := regexp.MustCompile(GroupMatchRegex)
-	if !groupMatch.MatchString(r.Group) {
-		return fmt.Errorf("group must match %s (was %s)", GroupMatchRegex, r.Group)
+	if err := IsDNS1123Subdomain(r.Group); err != nil {
+		return fmt.Errorf("group name is invalid: (%v)", err)
 	}
 
 	r.GroupImportSafe = strings.Replace(r.Group, "-", "", -1)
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -4,18 +4,14 @@ go 1.12
...
 	github.com/spf13/afero v1.2.2
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.3
-	golang.org/x/net v0.0.0-20190620200207-3b0461eec859 // indirect
 	golang.org/x/sys v0.0.0-20190621203818-d432491b9138 // indirect
-	golang.org/x/text v0.3.2 // indirect
 	golang.org/x/tools v0.0.0-20190614205625-5aca471b1d59
-	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
 	gopkg.in/yaml.v2 v2.2.2
+	k8s.io/apimachinery v0.0.0-20190927035529-0104e33c351d
```
mengqiy: We were trying to avoid any dependencies on k8s, since what kubebuilder does is just scaffolding.
Since you only depend on one simple validation function in apimachinery, copy it over may not be a bad idea.

@droot WDYT?

pkg/scaffold/v1/resource/resource.go
```diff
@@ -66,9 +65,8 @@ func (r *Resource) Validate() error {
 		r.Resource = flect.Pluralize(strings.ToLower(r.Kind))
 	}
 
-	groupMatch := regexp.MustCompile(GroupMatchRegex)
```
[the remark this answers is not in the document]
camilamacedo86: Why would we keep the regex and the validation which actually checks a valid group for the framework?

go.mod
camilamacedo86: HI @mengqiy, 

IMHO the problem with is that who is responsible to do the validation is the k8s. So, if its change their and we are using it we will know before doing a new release. In POV is not a good idea decentralised the impl but let me know your final decision. 

c/c @droot @DirectXMan12

pkg/scaffold/v1/resource/resource.go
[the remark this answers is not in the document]
mengqiy: I misread it.
Sorry, please ignore this comment.

go.mod
DirectXMan12: The definition of DNS subdomains are unlikely to change unless it's a bugfix, and we'd basically just be pulling this in for  a regular expression.  That being said, the regular expression is non-trivial.

I think overall though it's probably worth just copying to avoid all the extra dependencies.

DirectXMan12: (add a comment to just replace with library if we ever pull more stuff in)

camilamacedo86:

Really thank you for your review.  I think it is done now to merge. WDYT?

DirectXMan12:
awesome

/lgtm
/approve

PR #1049, opened 2019-09-30
fix: Error in the scaffold files when the group name has "-"

camilamacedo86:
**Description**
fix: Error in the scaffold files when the group name has "-"

**Motivation**
https://github.com/kubernetes-sigs/kubebuilder/issues/1007

**Steps to Test**
Follow the Quick Start with this PR and use the group name as webaap-test. Then, all should work well.

### DIFF
--- a/pkg/scaffold/v1/resource/resource.go
+++ b/pkg/scaffold/v1/resource/resource.go
@@ -34,6 +34,10 @@ type Resource struct {
 	// Group is the API Group.  Does not contain the domain.
 	Group string
 
+	// GroupImportSafe is the API Group.  Does not contain the domain and it the "-"
+	// It is used to do safe imports.
+	GroupImportSafe string
+
 	// Version is the API version - e.g. v1beta1
 	Version string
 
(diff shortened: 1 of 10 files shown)


pkg/scaffold/v1/resource/resource.go
```diff
@@ -34,6 +34,10 @@ type Resource struct {
 	// Group is the API Group.  Does not contain the domain.
 	Group string
```
mengqiy: Does it make sense if we enforce no `-` in `Group` field?
Then we don't need to introduce a new `GroupImportSafe` field.

camilamacedo86: Hi @mengqiy, 

Note that the correct validation for the group is the func, so users should be able to create groups as `myapp-operator` for example.  See more over it is https://godoc.org/k8s.io/apimachinery/pkg/util/validation#IsDNS1123Subdomain. (done a PR for it as well see https://github.com/kubernetes-sigs/kubebuilder/issues/1040)
 Also, see more about this here: https://github.com/kubernetes-sigs/kubebuilder/issues/1007#issuecomment-534455225

Issue #1000, opened 2019-09-17
Document testenv environment variables


DirectXMan12:
We should have reference for how to make use of the various options to testenv, either in the book, in the godoc, or both.

Especially the environment variable to attach control plane output, that one's super useful.

Ref #920 

/kind documentation

mariantalla:
I should be able to add that one, is it ok if I pick it up? Or is there work on it somewhere? 👀

DirectXMan12:
yeah, go ahead.  You'll want to edit the reference section on testing in the KB book (check docs/book/src/testing.md), and possibly double-check the godoc for pkg/envtest in controller-runtime.

PR #1055, opened 2019-09-30
Document envtest options

mariantalla:
ref: https://github.com/kubernetes-sigs/kubebuilder/issues/1000

This commit adds documentation on available options (environment
variables and flags) to configure the control plane set up using
envtest.

For now, it also adds the pre-existing integration.md file to
/references and links it to the kubebuilder book; this _might_ turn out
being out of scope (more on this below).

Feedback on the following would be very useful (in addition to anything missing or not quite right):
* Level of detail on what each option affects. Is it too vague/too detailed/ about right at the moment?
* (as mentioned above) I've added this content to integration.md, because some documentation related to integration testing was there already, so it seemed like a fitting spot. I'm not sure what intentions the author of integration.md had for their content though, so please let me know if that doesn't work and I'll split it out.
* On using API server flags, I've documented how to set all of them. I've also seen and used the pattern of `use the defaults but change that specific one that I want to override by appending the string to the list of default flags`. Is this a recommended pattern/worth adding?
* I haven't used `KUBEBUILDER_ASSETS ` and `USE_EXISTING_CLUSTER ` so that part is perhaps a bit thin. What else (if anything) is worth adding there?

### DIFF
--- a/docs/book/src/reference/testing/envtest.md
+++ b/docs/book/src/reference/testing/envtest.md
@@ -1,0 +1,56 @@
+## Using envtest in integration tests
+[`controller-runtime`](http://sigs.k8s.io/controller-runtime) offers `envtest` ([godoc](https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/envtest)), a package that helps write integration tests for your controllers by setting up and starting an instance of etcd and the Kubernetes API server, without kubelet, controller-manager or other components.
+
+Using `envtest` in integration tests follows the general flow of:
+
+```
+import sigs.k8s.io/controller-runtime/pkg/envtest
+
+//specify testEnv configuration
+testEnv = &envtest.Environment{
+	CRDDirectoryPaths: []string{filepath.Join("..", "config", "crd", "bases")},
+}
+
+//start testEnv
+cfg, err = testEnv.Start()
+
+//write test logic
+
+//stop testEnv
+err = testEnv.Stop()
+```
+
+`kubebuilder` does the boilerplate setup and teardown of testEnv for you, in the ginkgo test suite that it generates under the `/controllers` directory.
+
+Logs from the test runs are prefixed with `test-env`.
+
+### Configuring your test control plane
+You can use environment variables and/or flags to specify the `api-server` and `etcd` setup within your integration tests.
+
+#### Environment Variables
+
+| Variable name | Type | When to use |
+| --- | :--- | :--- |
+| `USE_EXISTING_CLUSTER` | boolean | Instead of setting up a local control plane, point to the control plane of an existing cluster. |
+| `KUBEBUILDER_ASSETS` | path to directory | Point integration tests to a directory containing all binaries (api-server, etcd and kubectl). |
+| `TEST_ASSET_KUBE_APISERVER`, `TEST_ASSET_ETCD`, `TEST_ASSET_KUBECTL` | paths to, respectively, api-server, etcd and kubectl binaries | Similar to `KUBEBUILDER_ASSETS`, but more granular. Point integration tests to use binaries other than the default ones. These environment variables can also be used to ensure specific tests run with expected versions of these binaries. |
+| `KUBEBUILDER_CONTROLPLANE_START_TIMEOUT` and `KUBEBUILDER_CONTROLPLANE_STOP_TIMEOUT` | durations in format supported by [`time.ParseDuration`](https://golang.org/pkg/time/#ParseDuration) | Specify timeouts different from the default for the test control plane to (respectively) start and stop; any test run that exceeds them will fail. |
+| `KUBEBUILDER_ATTACH_CONTROL_PLANE_OUTPUT` | boolean | Set to `true` to attach the control plane's stdout and stderr to os.Stdout and os.Stderr. This can be useful when debugging test failures, as output will include output from the control plane. |
+
+
+#### Flags
+Here's an example of modifying the flags with which to start the API server in your integration tests, compared to the default values in `envtest.DefaultKubeAPIServerFlags`:
+
+```
+customApiServerFlags := []string{
+	"--secure-port=6884",
+	"--admission-control=MutatingAdmissionWebhook",
+}
+
+apiServerFlags := append(envtest.DefaultKubeAPIServerFlags, customApiServerFlags)
+
+testEnv = &envtest.Environment{
+	CRDDirectoryPaths: []string{filepath.Join("..", "config", "crd", "bases")},
+	KubeAPIServerFlags: apiServerFlags,
+}
+```
--- a/docs/book/src/SUMMARY.md
+++ b/docs/book/src/SUMMARY.md
@@ -70,6 +70,8 @@
   - [Artifacts](./reference/artifacts.md)
   - [Writing controller tests](./reference/writing-tests.md)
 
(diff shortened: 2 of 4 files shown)


docs/book/src/reference/testing/integration.md
```diff
@@ -77,4 +77,62 @@ func TestReconcile(t *testing.T) {
 
 The manager is started as part of the test itself (`StartTestManager` function).
 
-Both functions are located in `pkg/controller/parent/parent_controller_suite_test.go` file. The file also contains a `TestMain` function that allows you to specify CRD directory paths for the testing environment.
\ No newline at end of file
+Both functions are located in `pkg/controller/parent/parent_controller_suite_test.go` file. The file also contains a `TestMain` function that allows you to specify CRD directory paths for the testing environment.
+
+## Using envtest in your integration tests
+[`controller-runtime`]() offers `envtest` ([godoc](https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/envtest)), a package that helps write integration tests for your controllers by setting up and starting an api-server binary and an etcd binary.
```
DirectXMan12: nit: "an instance of etcd and the Kubernetes API server, without kubelet, controller-manager etc" reads better IMO.

docs/book/src/reference/testing/integration.md
```diff
@@ -77,4 +77,62 @@ func TestReconcile(t *testing.T) {
...
+Logs from the test runs are prefixed `test-env`.
+
+### Configuring your test control plane
+You can use environment variables and/or flags to specify the `api-server` and `etcd` setup within your integration tests.
+
+#### Environment Variables
+
+| Variable name | Type | When to use |
+| --- | :--- | :--- |
+| `USE_EXISTING_CLUSTER` | boolean | Instead of setting up a local control plane, point to the control plane of an existing cluster. |
```
DirectXMan12:
```suggestion
| `USE_EXISTING_CLUSTER` | boolean | Instead of setting up a local control plane, point to the control plane of an existing cluster specified via the normal Kubernetes methods (in-cluster configuration, kubeconfig, etc). |
```

docs/book/src/reference/testing/integration.md
```diff
@@ -77,4 +77,62 @@ func TestReconcile(t *testing.T) {
...
+| `TEST_ASSET_KUBE_APISERVER`, `TEST_ASSET_ETCD`, `TEST_ASSET_KUBECTL` | paths to, respectively, api-server, etcd and kubectl binaries | Point integration tests to use binaries other than the default ones. They can also be used to ensure specific tests run with expected versions of these binaries.|
```
DirectXMan12: put `KUBEBUILDER_ASSETS` before this (see below).

docs/book/src/reference/testing/integration.md
```diff
@@ -77,4 +77,62 @@ func TestReconcile(t *testing.T) {
...
+| `KUBEBUILDER_CONTROLPLANE_START_TIMEOUT` and `KUBEBUILDER_CONTROLPLANE_STOP_TIMEOUT` | durations in seconds | Specify timeouts different from the default for the test control plane to (respectively) start and stop; any test run that exceeds them will fail. |
+| `KUBEBUILDER_ATTACH_CONTROL_PLANE_OUTPUT` | boolean | Set to `true` to attach the control plane's stdout and stderr to os.Stdout and os.Stderr. This can be useful when debugging test failures, as output will include output from the control plane. |
+| `KUBEBUILDER_ASSETS` | path to the `kubebuilder` binary | Use a different `kubebuilder` than the one in the machine's `PATH`. |
```
DirectXMan12: This is wrong.  KUBEBUILDER_ASSETS is a shortcut for `TEST_ASSET_XYZ` that assumes they're all in the same directory.  So if you would have

```bash
TEST_ASSET_KUBE_APISERVER=/opt/kube/bin/kube-apiserver
TEST_ASSET_ETCD=/opt/kube/bin/etcd
# etc
```

You can just say `KUBEBUILDER_ASSETS=/opt/kube/bin` and avoid repeating yourself.

docs/book/src/reference/testing/integration.md
```diff
@@ -77,4 +77,62 @@ func TestReconcile(t *testing.T) {
...
+
+```
+exampleApiServerFlags := []string{
+	"--etcd-servers=<ip:port>",
+	"--cert-dir=/path/to/cert",
+	"--insecure-port=8081",
+	"--insecure-bind-address=localhost",
+	"--secure-port=6443",
+	"--admission-control=MutatingAdmissionWebhook",
+}
```
DirectXMan12: Prob just append copy and then append to DefaultAPIServerFlags (the variable from envtest) here.

docs/book/src/reference/testing/integration.md
mariantalla: Thanks! I rephrased it.

docs/book/src/reference/testing/integration.md
mariantalla: Oh dear. Corrected it, thank you. Let me know if it's clear enough.

docs/book/src/reference/testing/integration.md
mariantalla: 👍 Changed [here](https://github.com/kubernetes-sigs/kubebuilder/pull/1055/commits/92c2875db101b684babb57b9559c6784fab83270).

docs/book/src/reference/testing/integration.md
mariantalla: 👍

mariantalla:
@DirectXMan12 - Thanks for the comments! I think I've addressed everything except GoDoc - I'll start a separate followup on that if that's ok. How does it look now?

docs/book/src/reference/testing/envtest.md
```diff
@@ -0,0 +1,58 @@
...
+| `KUBEBUILDER_CONTROLPLANE_START_TIMEOUT` and `KUBEBUILDER_CONTROLPLANE_STOP_TIMEOUT` | durations in format supported by [`time.ParseDuration`](https://golang.org/pkg/time/#ParseDuration) | Specify timeouts different from the default for the test control plane to (respectively) start and stop; any test run that exceeds them will fail. |
+| `KUBEBUILDER_ATTACH_CONTROL_PLANE_OUTPUT` | boolean | Set to `true` to attach the control plane's stdout and stderr to os.Stdout and os.Stderr. This can be useful when debugging test failures, as output will include output from the control plane. |
+
+
+#### Flags
+Here's an example of modifying the flags with which to start the API server in your integration tests, compared to the default values in `envtest.DefaultKubeAPIServerFlags`:
+
+```
+defaultApiServerFlags := envtest.DefaultKubeAPIServerFlags[0 : len(envtest.DefaultKubeAPIServerFlags)-2]
```
DirectXMan12: shouldn't just appending the new flags override the existing ones?  I don't think you have to slice like this.

Either way, at least add a comment like "don't reply on the order of these in a real setup" -- don't want people depending on `secure-port` always being the last one ;-)

DirectXMan12:
> think I've addressed everything except GoDoc - I'll start a separate followup on that if that's ok

SGTM :+1:

mengqiy:
@mariantalla Any update?

mariantalla:
Hey @mengqiy - sorry catching up after holiday got the better of me. I think I've addressed everything now (and combined in a single commit), let me know if anything is missing.

PR #1059, opened 2019-10-01
plugins/addon: perform replacements for type and controller

johnsonj:
This change adds the plumbing for plugins to act on the controller and fixes the addon pattern to output a buildable controller.

changes:
- controller runs in a second phase and needs the Plugins configured on
  the Scaffold
- fix path to the api/<version>/<resource>_type.go file for file
  replacement
- make the addon plugin work with the 2 phase run (api then controller)
  by not failing if the manifest/channel file has already been generated
- fixup import paths

### DIFF
--- a/pkg/scaffold/api.go
+++ b/pkg/scaffold/api.go
@@ -228,9 +228,13 @@ func (api *API) scaffoldV2() error {
 	if api.DoController {
 		fmt.Println(filepath.Join("controllers", fmt.Sprintf("%s_controller.go", strings.ToLower(r.Kind))))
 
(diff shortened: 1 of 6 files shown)


justinsb:
Thanks @johnsonj - not sure how I missed these, but you're quite right!

/lgtm

droot:
Noticed that it has one of the checks pending `The Travis CI build is in progress`, but travis is reporting the build to be successful. Couldn't find a way to retrigger the travis build. @mengqiy any idea how to unblock this merge ?

Issue #1070, opened 2019-10-04
Unable to run ` ./generated_golden.sh` and `make test-project-generation`


camilamacedo86:
```
$ ./generated_golden.sh
~/go/src/sigs.k8s.io/kubebuilder ~/go/src/sigs.k8s.io/kubebuilder
vendor/golang.org/x/crypto/sha3/testdata/keccakKats.json.deflate: (Empty error message)
tar: Error exit delayed from previous errors.
```

* go version go1.12.10 darwin/amd64

* kubebuilder version : Master

`Version: version.Version{KubeBuilderVersion:"2.0.0-beta.0", KubernetesVendor:"1.14.1", GitCommit:"b12371ccabd1bc8de7ec7ec1a1ca7bfabe567701", BuildDate:"2019-07-19T19:02:15Z", GoOs:"unknown", GoArch:"unknown"}
`

```
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/camilamacedo/Library/Caches/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/camilamacedo/go"
GOPROXY="https://proxy.golang.org"
GORACE=""
GOROOT="/usr/local/Cellar/go@1.12/1.12.10/libexec"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go@1.12/1.12.10/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/camilamacedo/go/src/sigs.k8s.io/kubebuilder/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/r_/ym6yn3cd0q3fm5jcgcwvj4z00000gn/T/go-build872400189=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

I am facing the same issue with : go version go1.13.1 darwin/amd64 too. 

/kind bug

[...]

PR #1072, opened 2019-10-04
fix contribution steps

camilamacedo86:
**Description**
Fix the steps as adding the prerequisites. 

**Motivation**
Issues faced trying to contribute with. 
https://github.com/kubernetes-sigs/kubebuilder/issues/1070


gerred:
I can confirm these steps and pre-reqs work for me. After working with @camilamacedo86 on this for some time today, it'd be nice if we could remove the internal dep dependency and make this all work with Go Modules, but I think that'll take some additional investigation.

gerred:
I meant without the leading /, muscle memory got the better of me. 🙃

Issue #1069, opened 2019-10-04
testdata has duplicated import after run golden script


camilamacedo86:
/kind bug

Place: `testdata/project-v2/main.go`

```go
	crewv1 "sigs.k8s.io/kubebuilder/testdata/project-v2/api/v1"
	v1 "sigs.k8s.io/kubebuilder/testdata/project-v2/api/v1"
```

Shows that are caused or triggered by https://github.com/kubernetes-sigs/kubebuilder/pull/1049/files#diff-c5b7fb8b471ae519bdfd5005565af33bR60-R77

It  is possible that GroupImportSafe is empty somehow: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/scaffold/v2/main.go#L64-L77

camilamacedo86:
solved.

PR #1054, opened 2019-09-30
add schema for CRD controllers

camilamacedo86:
**Description**
Pass schema to CRD cntrolller. 

**Motivation**
https://github.com/kubernetes-sigs/kubebuilder/issues/986

**Local Test**
Screenshot 2019-09-30 at 16 29 56

PR #1073, opened 2019-10-04
Fix double import

camilamacedo86:
**Description**
- Add a new unit test
- Fix issue where webhooks were to checking/validating the input data. 

**Motivation**
https://github.com/kubernetes-sigs/kubebuilder/issues/1069

### DIFF
--- a/pkg/scaffold/v1/resource/resource_test.go
+++ b/pkg/scaffold/v1/resource/resource_test.go
@@ -119,6 +119,12 @@ var _ = Describe("Resource", func() {
 			Expect(instance.Resource).To(Equal("cats"))
 		})
 
+		It("should allow hyphens in group names", func() {
+			instance := &resource.Resource{Group: "my-project", Kind: "Cat", Version: "v1"}
+			Expect(instance.Validate()).To(Succeed())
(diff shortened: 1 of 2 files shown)


camilamacedo86:
HI @mengqiy. 

Could you help us by doing the review of the issue faced with the golden script (double import after the introduction of GroupImportSafe)?

pkg/scaffold/v1/resource/resource_test.go
```diff
@@ -119,6 +119,12 @@ var _ = Describe("Resource", func() {
 			Expect(instance.Resource).To(Equal("cats"))
 		})
 
+		It("should allow webapp-project as a Group", func() {
```
DirectXMan12: test names should describe the intent of the test, so this should say something like "should allow hyphens in group names".

camilamacedo86: Done 👍 Tks for your input.

camilamacedo86:
Hi @DirectXMan12,

Could we merge this one and the https://github.com/kubernetes-sigs/kubebuilder/pull/1054 which is blocked by it now?

PR #1080, opened 2019-10-06
On kubebuilder init checking go.mod file before getting path

knabben:
When running the ```kubebuilder init``` the findCurrentRepo function returns "github.com", since go edit mod tries to "guess" the directory without being in.

The solution is to check if a go.mod file exists before running go edit mod to fetch the path.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -69,9 +69,11 @@ func findCurrentRepo() (string, error) {
 	}
 
 	// next easy case: existing go module
-	path, err := findGoModulePath(false)
-	if err == nil {
-		return path, nil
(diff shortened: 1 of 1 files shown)


DirectXMan12:
`go mod edit` shouldn't try to guess the current module name.  Do you have a `go.mod` under your github.com folder?

For instance, if I `go mod edit -json` under a non-module, I get

```
go: cannot find main module, but found .git/config in /path/to/some/gopath/repo
        to create a module there, run:
        go mod init
```

or

```
go: cannot find main module; see 'go help modules'
```

knabben:
my bad, I had the file there for mistake, that's the cause of the confusion, closing this.

PR #1084, opened 2019-10-09
fix(ci): By using fix version of dep which is compatible with go 1.12

camilamacedo86:
**Description**
- Replace get the latest version of dep by using the version 0.5.1 which we know that should work with go 1.12+. 
- Add GOPROXY=https://proxy.golang.org/ env in the Travis
- Fix indentation as well 
 
**Motivation**
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/1085

**Root Cause**
It was getting the latest one which will require 1.13 now and the project is not supporting it yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ os:
 go:
 - "1.12"
 
+env:
+  - GOPROXY=https://proxy.golang.org/
+
 git:
   depth: 3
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -10,19 +10,24 @@ os:
   - osx
 
 go:
-- "1.12"
+  - "1.12"
```
DirectXMan12: The previous indentation should've been fine -- YAML has an explicit provision that allows lists to avoid the extra indent.
Please revert that part of the change, since it makes it easier to examine history.

camilamacedo86: Done.

DirectXMan12:
/lgtm
/approve

once travis succeeds

PR #1090, opened 2019-10-09
Fix CI which is broken 

camilamacedo86:
**Description**
- Restructuring e2e tests by version to solve the timeout issue. See #1097. Also, it will be helpful to identify issues and increase the manutence ability of it. 

**Motivation**
Closes: #1097

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,7 @@ install:
 
 script:
 - if [ "$TRAVIS_OS_NAME" = "linux" ]; then GO111MODULE=on TRACE=1 PATH=$PATH:$(pwd) ./test_e2e.sh ; fi
+- if [ "$TRAVIS_OS_NAME" = "linux" ]; then GO111MODULE=on TRACE=1 PATH=$PATH:$(pwd) ./test_e2e_v2.sh ; fi
 - GO111MODULE=on TRACE=1 ./test.sh
 
 # TBD. Suppressing for now.
--- a/test_e2e_v2.sh
+++ b/test_e2e_v2.sh
@@ -27,4 +27,4 @@ setup_envs
 docker pull gcr.io/kubebuilder/kube-rbac-proxy:v0.4.1
 kind load docker-image gcr.io/kubebuilder/kube-rbac-proxy:v0.4.1
 
-go test ./test/e2e/v1
+go test ./test/e2e/v2
--- a/Makefile
+++ b/Makefile
@@ -41,4 +41,5 @@ test-project-generation:
 
 .PHONY: test-e2e
 test-e2e:
-	./test_e2e.sh
+	./test_e2e_v1.sh
+	./test_e2e_v2.sh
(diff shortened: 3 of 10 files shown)


test_e2e_v2.sh
```diff
@@ -0,0 +1,30 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+source common.sh
+
+fetch_tools
```
mengqiy: We can restructure the 2 e2e tests scripts to avoid duplication (e.g. fetch_tools, build_kb, docker pull and kind load docker-image).

test/e2e/v1/e2e_v1_test.go
```diff
@@ -14,12 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
```
mengqiy: > test/e2e/v1/e2e_v1_test.go

nit: `v1` appears twice.
We probably can change it to `test/e2e/v1/e2e_test.go` to reduce duplicated information in the path :)

test/e2e/v2/e2e_v2_test.go
```diff
@@ -14,12 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
```
mengqiy: > test/e2e/v2/e2e_v2_test.go

Same comment regarding the path, drop the latter `v2` in the path.

.travis.yml
```diff
@@ -26,8 +26,9 @@ services:
 before_install:
 # NOTE: The latest version of dep requires go 1.13
 - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then mkdir -p /Users/travis/gopath/bin; fi
+- GO111MODULE=on go get sigs.k8s.io/kind
```
mengqiy: Why we install kind one more time here?
We already do that in install_and_setup.sh.
BTW we should use a deterministic version e.g. `GO111MODULE=on go get sigs.k8s.io/kind@v0.5.1`

camilamacedo86: 👍 it just is required in the Make file because of another scenario/issue. Remove from here. Tks.

test_e2e_v2.sh
camilamacedo86: 👍

test/e2e/v2/e2e_v2_test.go
camilamacedo86: 👍

test/e2e/v1/e2e_v1_test.go
camilamacedo86: 👍

test_e2e_v2.sh
camilamacedo86: 👍

camilamacedo86: Hi @mengqiy, 

I tried more than one approach to attend it and all fail. See that we are with the CI broken for a while which is blocking we move forward with another priority. Also, IMHO the script still requiring a good review/refactor. Let's do it piece by piece? I mean what do you think about we track an issue to refactory them, centralize them in the scripts dir an etc?

PR #1096, opened 2019-10-11
Updated docs

surajnarwade:
Updated bash command in doc so that it can be easily copy and paste
from website/book


DirectXMan12:
/lgtm
/approve

thanks!

PR #1101, opened 2019-10-15
Use zap.New instead of zap.Logger in ctrl.SetLogger

micnncim:
`zap.Logger` has been deprecated so it's necessary to use `zap.New`.

ref) https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/log/zap/zap.go#L44

### DIFF
--- a/designs/simplified-scaffolding.md
+++ b/designs/simplified-scaffolding.md
@@ -417,7 +417,9 @@ func init() {
 }
 
 func main() {
-	ctrl.SetLogger(zap.Logger(true))
+	ctrl.SetLogger(zap.New(func(o *zap.Options) {
+		o.Development = true
+	}))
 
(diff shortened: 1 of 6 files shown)


designs/simplified-scaffolding.md
```diff
@@ -417,7 +417,9 @@ func init() {
 }
 
 func main() {
-	ctrl.SetLogger(zap.Logger(true))
+	ctrl.SetLogger(zap.New(func(o *zap.Options) {
```
DirectXMan12: This is really pretty hacky -- let's avoid making users do this in the future -- we should have a pre-made option for development mode, and we shouldn't be scaffolding things out with something hacky like this.  Remember, scaffolding doesn't get to change, so if we scaffold out something clunky, there'll be a whole set of users that get this code.

DirectXMan12: Filed PR for CR: https://github.com/kubernetes-sigs/controller-runtime/pull/646

PR #1106, opened 2019-10-17
🏃 install kind if not present in e2e

alexeldeib:
for prow integration with e2e

needed for https://github.com/kubernetes/test-infra/pull/14829

this contains the commits in #689 as well, but wanted to work on these separately, so will clean up history as necessary once that merges. I just need the PR to test the prowjob locally 😄

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -152,7 +152,10 @@ function fetch_kind {
   header_text "Checking for kind"
   if ! is_installed kind ; then
     header_text "Installing kind"
+    KIND_DIR=$(mktemp -d)
+    pushd $KIND_DIR
     GO111MODULE=on go get sigs.k8s.io/kind@v0.5.1
+    popd
   fi
 }
 
--- a/common.sh
+++ b/common.sh
@@ -117,6 +117,10 @@ function fetch_tools {
   fi
   fetch_go_tools
   fetch_kb_tools
+  IS_E2E=${E2E:-""}
+  if [ -n "$IS_E2E" ]; then
+    fetch_kind
+  fi
 }
 
 function fetch_kb_tools {
@@ -144,6 +148,14 @@ function build_kb {
   GO111MODULE=on go build $opts -o $tmp_root/kubebuilder/bin/kubebuilder ./cmd
 }
 
+function fetch_kind {
+  header_text "Checking for kind"
+  if ! is_installed kind ; then
+    header_text "Installing kind"
+    GO111MODULE=on go get sigs.k8s.io/kind@v0.5.1
+  fi
+}
+
 function fetch_go_tools {
   header_text "Checking for dep"
   export PATH=$(go env GOPATH)/src/github.com/golang/dep/bin:$PATH
@@ -159,7 +171,6 @@ function fetch_go_tools {
     go build -ldflags="-X main.version=$DEP_LATEST" -o bin/dep ./cmd/dep
     popd
   fi
-
 }
 
 function is_installed {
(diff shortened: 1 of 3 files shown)


mengqiy:
Code LGTM but this PR needs rebase.

alexeldeib:
rebased

alexeldeib:
hmm..

alexeldeib:
that doesn't bode well...it failed to find the etcd binary on disk? have you or @DirectXMan12 seen this kind of flake before? don't think it's related to this PR, but I want to make sure the prow stuff works

common.sh
```diff
@@ -145,6 +149,14 @@ function build_kb {
   GO111MODULE=on go build $opts -o $tmp_root/kubebuilder/bin/kubebuilder ./cmd
 }
 
+function fetch_kind {
+  header_text "Checking for kind"
+  if ! is_installed kind ; then
+    header_text "Installing kind"
+    GO111MODULE=on go get sigs.k8s.io/kind@v0.5.1
```
droot: I was recently bit by this behavior. `go get` will modify `go.mod` file if `go get` fetches new deps and also introduces indirect dependency on the project. So the work around is to run `go get` from a temporary directory instead of kb directory.

droot: We have install_and_setup.sh under `scripts` that is used in travis setup for setting up kind cluster. Would be good if we can have common code for installing kind so that we don't end up with two different ways of installing kind.

scripts/install_and_setup.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
```
alexeldeib: any reason not to do this?

alexeldeib: nvm, will try with `.`

alexeldeib: wow, sourcing files is hard in bash.

common.sh
alexeldeib: how's that?

common.sh
```diff
@@ -144,6 +148,17 @@ function build_kb {
   GO111MODULE=on go build $opts -o $tmp_root/kubebuilder/bin/kubebuilder ./cmd
 }
 
+function fetch_kind {
+  header_text "Checking for kind"
```
camilamacedo86: I think it is not required since we can call `./sccripts/install_and_setup.sh` to this setup. WDYT?

alexeldeib: I think, maybe I need to do some more work than I thought. I was duplicating it because in prow you only get one command, so unlike travis there is no "pre-run" of installing and setting up kind. 

I think I should probably make the whole invocation idempotent, handle the setup in a root script which invokes v1/v2 e2e test. Do you think that makes sense?

camilamacedo86: I am working to try refac the scripts. I just started. See https://github.com/kubernetes-sigs/kubebuilder/pull/1151. ( note that it has the changes performed in the https://github.com/kubernetes-sigs/kubebuilder/pull/1135 as well for I am able to move forward and work with using go 1.13. ). Let's see if in this PR I will be able to reduce the complexity and remove the duplications which may help us better understand what/where is required.

alexeldeib:
Explicitly adding 

/hold

mengqiy:
> Explicitly adding /hold

@alexeldeib Are you blocked by anything?

common.sh
camilamacedo86: @mengqiy ^

alexeldeib:
see @camilamacedo86's remark, just waiting to settle that refactor first.

alexeldeib:
think this is no longer necessary, thanks @camilamacedo86 !

alexeldeib:
actually, it is 😄 sorry to keep starting fresh PRs, but i'll port it over there

PR #1107, opened 2019-10-18
:running: more robust golden test

### LINKED ISSUE #1003: Golden files test is broken
kubebuilder has a test to ensure the files under `/testdata` always reflects what is generated. Test will fail if `/testdata` is out of date.
But it seems to be broken now. See: https://github.com/kubernetes-sigs/kubebuilder/pull/983#pullrequestreview-290262845

/kind bug

mengqiy:
Added a more robust golden test:
1) build a new kubebuilder
2) generate the new golden files
3) check there are any difference

fixes #1003


droot:
Can we add makefile target or add it to existing makefile targets ?
Also, travis is not happy with the changes.

Otherwise looking good to me.

mengqiy:
Tests are passing. PTAL

.travis.yml
```diff
@@ -27,6 +27,8 @@ before_install:
 # NOTE: The latest version of dep requires go 1.13
 - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then mkdir -p /Users/travis/gopath/bin; fi
 - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then GO111MODULE=on scripts/install_and_setup.sh; fi
+- GO111MODULE=on go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.1
```
droot: we released 0.2.2 yesterday, may be use that. May be we should make these (kustomize below) version as some variables, so that it is easier to change them.

Issue #1119, opened 2019-10-24
Missing import in scaffolded code


Adirio:
# Bug description
Missing `"k8s.io/apimachinery/pkg/runtime"` import in controller scaffolding.
I think my IDE added the import automatically as I can't find it in the scaffold file.

https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/scaffold/v2/controller.go#L71-L86

/kind bug

Adirio:
@camilamacedo86

mengqiy:
Yes, this is a bug.
Thanks for bringing this up.
/assign @camilamacedo86

mengqiy:
It seems @Adirio has already started to work on this in #1120.

droot:
@mengqiy I am wondering why isn't this caught by our tests ?

camilamacedo86:
Hi @mengqiy, @droot and @Adirio, 

I tested it locally and I could not find any issue regards to it. The import will be added in the controller as it was done by the golden script [here](https://github.com/kubernetes-sigs/kubebuilder/blob/3103ad75279f83805495ace03095df88ee42af81/testdata/project-v2/controllers/captain_controller.go#L23) in the testdata.

Following the steps. 

```
$ mkdir example
$ cd example/
$ kubebuilder init --domain my.domain
$ kubebuilder create api --group webapp --version v1 --kind Guestbook
```

- Select y and y. 
- Then, let's check the controller by the vim in order to ensure that any IDE will not o the import automatically `vim controllers/guestbook_controller.go`

package controllers
    
```go
import (
        "context"
    
        "github.com/go-logr/logr"
        "k8s.io/apimachinery/pkg/runtime"
        ctrl "sigs.k8s.io/controller-runtime"
        "sigs.k8s.io/controller-runtime/pkg/client"

        webappv1 "sigs.k8s.io/example/api/v1"
)

// GuestbookReconciler reconciles a Guestbook object
type GuestbookReconciler struct {
        client.Client
        Log    logr.Logger
        Scheme *runtime.Scheme
}

```

See that `"k8s.io/apimachinery/pkg/runtime"`  is in the imports and it is what is used in ` Scheme *runtime.Scheme`.

mengqiy:
The reason why missing import haven't cause us problem in test is that the generated go code go through goimport. After processing, goimports seems to quite smart to add the proper imports.
We should not rely on goimports to add that for us.

droot:
[...]

PR #1120, opened 2019-10-24
:sparkles: Cleanup scaffolding 

Adirio:
Fixes #1119

PR #1130, opened 2019-10-25
Fix import

camilamacedo86:
**Description** 
Add missing import.

**Motivation**
Closes: #1119

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ generate-vendor: ## Update/generate the vendor by using the path $GOPATH/src/sig
 .PHONY: generate-setup
 generate-setup: ## Current workarround to generate the testdata with the correct controller-gen version
 	- rm -rf $(CONTROLLER_GEN_BIN_PATH)
-	- GO111MODULE=on go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.1
+	- GO111MODULE=on go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.2
 
 ##@ Tests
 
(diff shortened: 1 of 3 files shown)


camilamacedo86:
c/c @mengqiy, @droot and @Adirio.

droot:
Looks good to me. Pl. combine the last two commits, then its good to go.

camilamacedo86:
Done @droot :-)

Adirio:
This was already being tackled in #1120

camilamacedo86:
Adirio, this PR in oder to solve it and help you since you are facing issues to use the golden script.

Makefile
```diff
@@ -60,7 +60,7 @@ generate-vendor: ## Update/generate the vendor by using the path $GOPATH/src/sig
 .PHONY: generate-setup
 generate-setup: ## Current workarround to generate the testdata with the correct controller-gen version
 	- rm -rf $(CONTROLLER_GEN_BIN_PATH)
-	- GO111MODULE=on go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.1
+	- GO111MODULE=on go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.2
```
Adirio: Should this be part of a different PR? Bumping controller-gen to v0.2.2 doesn't seem related to the import fix.

camilamacedo86: Hi Adirio, the command `make generate-setup` is just a helper for development purpose. We can use it before exec the golden script to ensure that the correct version of the `controller-gen` will be used. So, it is a very small update and since it is too small I believe that would be ok be in this PR.

Issue #1091, opened 2019-10-10
mixed up between camelCase and PascalCase in kind validation


Azadehkhojandi:
There is typo/bug in the validation message => mixed up between camelCase and PascalCase

`kubebuilder create api --group databricks --version v2beta1 --kind testKind`

The error says camelcase but what is expected is PascalCase
> kind must be camelcase (expected TestKind was testKind)

reference:
https://blogs.msdn.microsoft.com/brada/2004/02/03/history-around-pascal-casing-and-camel-casing/
https://stackoverflow.com/questions/41768733/camel-case-and-pascal-case-mistake

DirectXMan12:
you are technically correct (and well sourced), which is the best kind of correct :-D

/kind bug
/good-first-issue
/priority backlog

vishakhanihore:
@DirectXMan12 Can you point me to the file location ?

KaenDagger:
Could I take up this issue to get started ?

Adirio:
@mariantalla should this be closed by #1133 which just got merged?

mariantalla:
Yep, with #1133 now merged, this should be fixed. `Kind`s are still expected to start with an uppercase character, but the error message should be clearer now.

/close

rayudu-gangadhara-rao:
jhsfjhsdv

Issue #1019, opened 2019-09-24
Kind validation is a bit too strict


dlespiau:
There are some circumstances where the the CamelCase check is a bit too strict. For instance, the [`AWSCluster` kind](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/PROJECT#L10) doesn't work today:

```
$ kubebuilder create api --group infrastructure --version xxxx --kind AWSCluster
...
2019/09/24 18:21:13 error scaffolding APIs: kind must be camelcase (expected Awscluster was AWSCluster)
```

Acronyms should be handled by the checker, this looks like a valid kind to me.

/kind bug

DirectXMan12:
yeah, we should be matching whatever kube specifies here, I think

/help

Azadehkhojandi:
on another note -> related bug #1091

mariantalla:
Re: matching what kube specifies, one option is to replicate validation logic from core Kubernetes to where validation happens in Kubebuilder ([this file](https://github.com/kubernetes-sigs/kubebuilder/blob/c9f4434235379e1d0ffca9b9cf36d720d7c39014/pkg/scaffold/v1/resource/resource.go), I think). 

Another option is to use the validation logic from "core" Kubernetes directly. I think that logic lives [here](https://github.com/kubernetes/kubernetes/blob/41a413331835d10d68f480e58bee862c6ef4084a/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go#L545). In practice it means importing `apiextensions-apiserver` and calling out to `ValidateCustomResourceDefinitionNames` or similar. Kubebuilder seems to vendor `apiextensions-apiserver` already so it shouldn't be too bad. Does that sound like a reasonable path forward? 🤔

DirectXMan12:
Yeah, that's probably fine.  Just validating the names directly is probably ok since we already vendor that.

camilamacedo86:
HI, @mariantalla  could you please assign the issue to you in order to avoid other dev get it since you gave a PR already one for this one? Also, devs will usually filter for the unassigned ones to know what they can do to contribute with. 

PS.: To do it just add a new comment with `/assign @your user`

mariantalla:
With https://github.com/kubernetes-sigs/kubebuilder/pull/1133 now merged, this should be fixed. Will close, please reopen if it's still a problem.

[...]

PR #1133, opened 2019-10-28
Support mixed case Kind

mariantalla:
This PR should address issues https://github.com/kubernetes-sigs/kubebuilder/issues/1019 and https://github.com/kubernetes-sigs/kubebuilder/issues/1091.

The implementation uses the validation logic that runs against `Kind` in
apimachinery; i.e. allows any `Kind` value that, when converted to
lowercase, is a valid DNS-1035 label.

Reusing that logic instead of reimplementing it helps with ensuring that
the two (i.e. `Kind` validation in core Kubernetes and `Kind` validation
in kubebuilder) are as similar as possible.

Would love feedback on:
* The choice to use [`k8s.io/apimachinery` logic](https://github.com/kubernetes/kubernetes/blob/b3b15a9e260ac65edfa14dd80add9046f3fa9b1f/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go#L549-L582) vs continuing to extend the validation logic [in kubebuilder](https://github.com/kubernetes-sigs/kubebuilder/blob/702b272f9c73c723d0aec3569489ffcf79d6d1b9/pkg/scaffold/v1/resource/resource.go#L82).
* If the above does sound like a good idea, is it worth refactoring [`Validate()`](https://github.com/kubernetes-sigs/kubebuilder/blob/702b272f9c73c723d0aec3569489ffcf79d6d1b9/pkg/scaffold/v1/resource/resource.go#L56) more broadly, to use available logic in `k8s.io/apimachinery`?
* There were more changes than I expected in go.mod 😅 I'll take an extra look at that.

Remaining work in this PR:
(once the approach seems reasonable)
* Refactor tests to describe successful & failed validation more succinctly (at the moment there's duplication and missing test cases while the PR is in progress)
* Ensure any logging and error messages are descriptive of the validation that takes place.

### DIFF
--- a/pkg/model/resource/options_test.go
+++ b/pkg/model/resource/options_test.go
@@ -1,7 +1,10 @@
 package resource_test
 
 import (
+	"strings"
+
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 
 	. "sigs.k8s.io/kubebuilder/pkg/model/resource"
@@ -73,24 +76,38 @@ var _ = Describe("Resource Options", func() {
 			Expect(options.Validate().Error()).To(ContainSubstring("kind cannot be empty"))
 		})
 
-		It("should fail if the Kind is not pascal cased", func() {
-			// Base case
-			options := &Options{Group: "crew", Version: "v1", Kind: "FirstMate"}
-			Expect(options.Validate()).To(Succeed())
+		DescribeTable("valid Kind values-according to core Kubernetes",
+			func(kind string) {
+				options := &Options{Group: "crew", Kind: kind, Version: "v1"}
+				Expect(options.Validate()).To(Succeed())
+			},
+			Entry("should pass validation if Kind is camelcase", "FirstMate"),
+			Entry("should pass validation if Kind has more than one caps at the start", "FIRSTMate"),
+		)
 
-			// Can't detect this case :(
-			options = &Options{Group: "crew", Version: "v1", Kind: "Firstmate"}
-			Expect(options.Validate()).To(Succeed())
+		It("should fail if Kind is too long", func() {
+			kind := strings.Repeat("a", 64)
 
-			options = &Options{Group: "crew", Version: "v1", Kind: "firstMate"}
-			Expect(options.Validate()).NotTo(Succeed())
-			Expect(options.Validate().Error()).To(ContainSubstring(
-				`kind must be PascalCase (expected FirstMate was firstMate)`))
+			options := &Options{Group: "crew", Kind: kind, Version: "v1"}
+			err := options.Validate()
+			Expect(err).To(MatchError(ContainSubstring("must be no more than 63 characters")))
+		})
 
-			options = &Options{Group: "crew", Version: "v1", Kind: "firstmate"}
-			Expect(options.Validate()).NotTo(Succeed())
-			Expect(options.Validate().Error()).To(ContainSubstring(
-				`kind must be PascalCase (expected Firstmate was firstmate)`))
+		DescribeTable("invalid Kind values-according to core Kubernetes",
+			func(kind string) {
+				options := &Options{Group: "crew", Kind: kind, Version: "v1"}
+				Expect(options.Validate()).To(MatchError(
+					ContainSubstring("a DNS-1035 label must consist of lower case alphanumeric characters")))
+			},
+			Entry("should fail validation if Kind contains whitespaces", "Something withSpaces"),
+			Entry("should fail validation if Kind ends in -", "KindEndingIn-"),
+			Entry("should fail validation if Kind starts with number", "0ValidityKind"),
+		)
+
+		It("should fail if Kind starts with a lowercase character", func() {
+			options := &Options{Group: "crew", Kind: "lOWERCASESTART", Version: "v1"}
+			err := options.Validate()
+			Expect(err).To(MatchError(ContainSubstring("Kind must start with an uppercase character")))
 		})
 	})
 })
--- a/pkg/model/resource/options.go
+++ b/pkg/model/resource/options.go
@@ -126,9 +126,17 @@ func (opts *Options) Validate() error {
 			"version must match %s (was %s)", versionPattern, opts.Version)
 	}
 
-	// Check if the Kind is PascalCase
-	if opts.Kind != flect.Pascalize(opts.Kind) {
-		return fmt.Errorf("kind must be PascalCase (expected %s was %s)", flect.Pascalize(opts.Kind), opts.Kind)
+	validationErrors := []string{}
+
+	// require Kind to start with an uppercase character
+	if string(opts.Kind[0]) == strings.ToLower(string(opts.Kind[0])) {
+		validationErrors = append(validationErrors, "Kind must start with an uppercase character")
+	}
+
+	validationErrors = append(validationErrors, isDNS1035Label(strings.ToLower(opts.Kind))...)
+
+	if len(validationErrors) != 0 {
+		return fmt.Errorf("Invalid Kind: %#v", validationErrors)
 	}
 
 	// TODO: validate plural strings if provided
--- a/pkg/model/resource/validation.go
+++ b/pkg/model/resource/validation.go
@@ -33,9 +33,17 @@ const (
 
 	// dns1123SubdomainMaxLength is a subdomain's max length in DNS (RFC 1123)
 	dns1123SubdomainMaxLength int = 253
+
+	dns1035LabelFmt    string = "[a-z]([-a-z0-9]*[a-z0-9])?"
+	dns1035LabelErrMsg string = "a DNS-1035 label must consist of lower case alphanumeric characters or '-', " +
+		"start with an alphabetic character, and end with an alphanumeric character"
+
+	// DNS1035LabelMaxLength is a label's max length in DNS (RFC 1035)
+	dns1035LabelMaxLength int = 63
 )
 
 var dns1123SubdomainRegexp = regexp.MustCompile("^" + dns1123SubdomainFmt + "$")
+var dns1035LabelRegexp = regexp.MustCompile("^" + dns1035LabelFmt + "$")
 
 // IsDNS1123Subdomain tests for a string that conforms to the definition of a
 // subdomain in DNS (RFC 1123).
@@ -46,6 +54,17 @@ func IsDNS1123Subdomain(value string) []string {
 	}
 	if !dns1123SubdomainRegexp.MatchString(value) {
 		errs = append(errs, regexError(dns1123SubdomainErrorMsg, dns1123SubdomainFmt, "example.com"))
+	}
+	return errs
+}
+
+func isDNS1035Label(value string) []string {
+	var errs []string
+	if len(value) > dns1035LabelMaxLength {
+		errs = append(errs, maxLenError(dns1035LabelMaxLength))
+	}
+	if !dns1035LabelRegexp.MatchString(value) {
+		errs = append(errs, regexError(dns1035LabelErrMsg, dns1035LabelFmt, "my-name", "abc-123"))
 	}
 	return errs
 }


go.mod
```diff
@@ -4,18 +4,20 @@ go 1.12
...
-	golang.org/x/text v0.3.2 // indirect
+	golang.org/x/net v0.0.0-20190812203447-cdfb69ac37fc
+	golang.org/x/oauth2 v0.0.0-20190604053449-0f29369cfe45 // indirect
 	golang.org/x/tools v0.0.0-20190614205625-5aca471b1d59
-	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
-	gopkg.in/yaml.v2 v2.2.2
+	google.golang.org/appengine v1.5.0 // indirect
+	gopkg.in/yaml.v2 v2.2.4
+	k8s.io/api v0.0.0-20190918195907-bd6ac527cfd2
+	k8s.io/apimachinery v0.0.0-20191004115801-a2eda9f80ab8
```
mariantalla: This does map to [`k8s.io/apimachinery`@`kubernetes-1.16.2`](https://github.com/kubernetes/apimachinery/releases/tag/kubernetes-1.16.2) (the latest stable at the moment), but I can't get go modules to save it like that. I could be missing something 🤔

mariantalla:
/hold
(while there's open TODO items)

go.mod
mariantalla: OK, from learning a bit more about go modules, that's because the tag names are not semver-style (i.e. `vx.y.z`). The commits correspond to the right tags though, so that should be fine.

mariantalla:
Any thoughts @mengqiy @DirectXMan12 @Liujingfang1 ?

/assign @mengqiy @DirectXMan12 @Liujingfang1

pkg/scaffold/v1/resource/resource.go
```diff
@@ -79,7 +80,10 @@ func (r *Resource) Validate() error {
 		return fmt.Errorf(
 			"version must match ^v\\d+(alpha\\d+|beta\\d+)?$ (was %s)", r.Version)
 	}
-	if r.Kind != flect.Pascalize(r.Kind) {
+
+	validationErrors := validation.IsDNS1035Label(strings.ToLower(r.Kind))
+
+	if len(validationErrors) != 0 {
 		return fmt.Errorf("kind must be camelcase (expected %s was %s)", flect.Pascalize(r.Kind), r.Kind)
```
jdef: there's still a call to `Pascalize` here; AFAICT this type of normalization should be eliminated.

pkg/scaffold/v1/resource/resource_test.go
```diff
@@ -79,24 +79,19 @@ var _ = Describe("Resource", func() {
 			Expect(instance.Validate().Error()).To(ContainSubstring("kind cannot be empty"))
 		})
 
-		It("should fail if the Kind is not camel cased", func() {
+		It("should succeed if the Kind is camel cased", func() {
```
jdef: maybe include a negative test here too? something that fails dns1035 label validation, just so that we can see that validation error checks are working as expected.

go.mod
```diff
@@ -3,19 +3,17 @@ module sigs.k8s.io/kubebuilder
 go 1.12
 
 require (
```
camilamacedo86: Update the go.mo and go.sum shows not be part of the scope of this PR. 
I'd recommend you revert it and ensure that your branch is rebased with the master.

mariantalla:
Thanks everyone for the comments. There's indeed a couple of follow-ups before this is on its way to merge (I've updated the PR description so it's clearer what work is left).

At the moment I was curious to hear how folks left towards using `k8s.io/apimachinery` code for validation (as opposed to continuing to extend the validation that kubebuilder is doing). Once there's consensus in that, error messages and test cases are pretty straightforward to change.

pkg/scaffold/v1/resource/resource.go
```diff
@@ -79,7 +80,10 @@ func (r *Resource) Validate() error {
 		return fmt.Errorf(
 			"version must match ^v\\d+(alpha\\d+|beta\\d+)?$ (was %s)", r.Version)
 	}
-	if r.Kind != flect.Pascalize(r.Kind) {
+
+	validationErrors := validation.IsDNS1035Label(strings.ToLower(r.Kind))
```
camilamacedo86: See here that the import is not required. In the PR https://github.com/kubernetes-sigs/kubebuilder/pull/1048/files#diff-ccdb7c3b4336337dc2e653f1fbd88c7dR71 a similar validation from apimachinery was added in the project. There was decide do not do the import. So, I'd recommend you do the same.

mariantalla:
Following the advice in https://github.com/kubernetes-sigs/kubebuilder/pull/1048#discussion_r329712728 (thanks @camilamacedo86  for the pointer), we've removed dependencies on apimachinery.

Adirio:
The Travis error seems not to be related to this PR but to #1169

mariantalla:
oops, forgot to
/hold cancel

pkg/scaffold/resource/resource.go
```diff
@@ -96,9 +99,16 @@ const (
 
 	// dns1123SubdomainMaxLength is a subdomain's max length in DNS (RFC 1123)
 	dns1123SubdomainMaxLength int = 253
+
+	dns1035LabelFmt    string = "[a-z]([-a-z0-9]*[a-z0-9])?"
```
camilamacedo86: Shows that it is not the same validation for Kinds. Could we not have a kind == AppServer for example?

pkg/scaffold/resource/resource_test.go
```diff
@@ -71,24 +71,27 @@ var _ = Describe("Resource", func() {
 			Expect(instance.Validate().Error()).To(ContainSubstring("kind cannot be empty"))
 		})
 
-		It("should fail if the Kind is not pascal cased", func() {
-			// Base case
-			instance := &Resource{Group: "crew", Kind: "FirstMate", Version: "v1"}
+		It("should succeed if the Kind is valid, according to core Kubernetes", func() {
```
camilamacedo86: IHMO the existent KInds check should pass. 
Here for example : https://github.com/kubernetes-sigs/kubebuilder/issues/1019 is for we allow AWSCluster

pkg/scaffold/resource/resource.go
```diff
@@ -79,8 +79,11 @@ func (r *Resource) Validate() error {
 		return fmt.Errorf(
 			"version must match ^v\\d+(alpha\\d+|beta\\d+)?$ (was %s)", r.Version)
 	}
-	if r.Kind != flect.Pascalize(r.Kind) {
-		return fmt.Errorf("kind must be PascalCase (expected %s was %s)", flect.Pascalize(r.Kind), r.Kind)
```
camilamacedo86: My understand to solve the https://github.com/kubernetes-sigs/kubebuilder/issues/1091 is just replace the text in the msg `CamelCase` to `PascalCase` which shows done already.

pkg/scaffold/resource/resource.go
```diff
@@ -79,8 +79,11 @@ func (r *Resource) Validate() error {
...
+
+	validationErrors := isDNS1035Label(strings.ToLower(r.Kind))
```
camilamacedo86: my recommendation here to solve it is:
- Remove the Pascalize
- Check if the Kind matches with: 
`go
	ResourceKindRegexp = regexp.MustCompile("^[A-Z]{1}[a-zA-Z0-9]+$")
`

pkg/scaffold/resource/resource.go
mariantalla: Yep; the dns1035-like validation happens after the `Kind` [has been lowercased](https://github.com/kubernetes-sigs/kubebuilder/blob/5742ddd7a096570cd7fcc0881176d9dcb189d27b/pkg/scaffold/resource/resource.go#L83), so all of these combinations would be supported. apimachinery follows the same pattern, as far as I can tell (I'll find it and link here in a moment).

pkg/scaffold/resource/resource_test.go
mariantalla: Agreed; this PR relaxes `Kind` validation (to bring it in line with the way it happens in core k8s). Previous passing tests continue to pass, and some previously failing should no longer fail (e.g. [this one](https://github.com/kubernetes-sigs/kubebuilder/blob/971cf03ea49aefc2380fdb6c5a91c55316f4617e/pkg/scaffold/resource/resource_test.go#L83-L91)). Or am I missing something? 🤔

pkg/scaffold/resource/resource.go
mariantalla: This PR will remove `Pascalize` completely though (in favor of the kube-like validation), and the new error messages describe what went wrong (hopefully! Open to feedback on that 😄 )

mariantalla:
Thanks for the comments @camilamacedo86 ! I've responded inline; but just to make sure I got your last suggestion right:

> Remove the Pascalize

I think this is happening [here](https://github.com/kubernetes-sigs/kubebuilder/pull/1133/files#diff-1264bafa536db5239815912ade8bc615L82-L83); where there other spots it needs to change?

> Use the func ... from apimachinery

Kind validation now uses IsDns1350Label, same as in core kubernetes

> Then, add a unit test

I think [this test](https://github.com/kubernetes-sigs/kubebuilder/blob/5742ddd7a096570cd7fcc0881176d9dcb189d27b/pkg/scaffold/resource/resource_test.go#L74-L77) covers that case, is there something missing?

jdef:
Maybe also include a test that rejects a Kind that's all lowercase? non-exported types don't seem very useful and the old Pascal validation used to guard against them.

pkg/scaffold/resource/resource.go
camilamacedo86: Hi @mariantalla, 

See [here](https://github.com/kubernetes/apimachinery/blob/6e68a40eebf94cb3e778d807bce5637d6aa079ea/pkg/apis/meta/v1/types_swagger_doc_generated.go#L57-L58) that  kubernetes API andd apimachinery will allow Kinds with upper case.

The validation that you got from there is not used to check Kinds.  You can use this one [here](https://github.com/kubernetes/apimachinery/blob/ced427e1ea5fdd2d307fa083f960961ceff71b65/pkg/util/validation/validation.go#L123 ) to verify the kinds.

pkg/scaffold/resource/resource_test.go
camilamacedo86: The validation obtained is not the correct one for we check the kinds. If you use the correct one the only change here will be adding more one test to ensure that `AWSCluster` will pass. Also, note that is required to remove the pascalze as well since it is not something check-in k8s API too.

camilamacedo86:
HI @mariantalla, 

> validation now uses IsDns1350Label, same as in core Kubernetes  

I am able to apply KInds with upper cases in Kubernetes so it shows not be the validation used by it to check Kinds.

pkg/scaffold/resource/resource.go
DirectXMan12: I think @mariantalla means that apimachinery validates that the lower-case form of any kind is a valid dns-like label

DirectXMan12: (although we should have links in comments referencing that)

DirectXMan12: the relevant section: 

https://github.com/kubernetes/kubernetes/blob/b3b15a9e260ac65edfa14dd80add9046f3fa9b1f/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go#L549-L582

@mariantalla is correct

DirectXMan12:
> Maybe also include a test that rejects a Kind that's all lowercase? non-exported types don't seem very useful and the old Pascal validation used to guard against them.

Yeah.  Technically this is allowed, practically we should prevent it.

DirectXMan12:
otherwise looks good to me

pkg/scaffold/resource/resource_test.go
```diff
@@ -71,24 +71,27 @@ var _ = Describe("Resource", func() {
...
+		})
 
-			// Can't detect this case :(
-			instance = &Resource{Group: "crew", Kind: "Firstmate", Version: "v1"}
-			Expect(instance.Validate()).To(Succeed())
+		It("should fail if Kind is too long", func() {
+			var kind string
 
-			instance = &Resource{Group: "crew", Kind: "firstMate", Version: "v1"}
-			Expect(instance.Validate()).NotTo(Succeed())
```
DirectXMan12: should bring these tests back though, so we have a full range

pkg/scaffold/resource/resource.go
camilamacedo86: HI @mariantalla and @DirectXMan12,
 
Really tks for the explanations. I got that the lower case is just for the check and also, I tested locally the PR and all is working fine. It all shows fine, just the tests that were removed that IHMO should be back. We should still be ensuring that Crew and Guestbook and etc are accepted.

pkg/scaffold/resource/resource.go
camilamacedo86: Solved can close this one. great work 👍

pkg/scaffold/resource/resource.go
```diff
@@ -79,8 +79,11 @@ func (r *Resource) Validate() error {
...
+
+	if len(validationErrors) != 0 {
+		return fmt.Errorf("Kind failed validation: %#v", validationErrors)
```
camilamacedo86:
```suggestion
		return fmt.Errorf("Invalid Kind: %#v", validationErrors)
```

pkg/scaffold/resource/resource.go
mariantalla: Added the link in a few places in the conversation so that future-us can find it too :)

mariantalla:
Pushed a couple of commits that I think cover any remaining points:
* added back previous test cases
* added a test and implementation to reject Kinds that have lowercase first characters

pkg/scaffold/resource/resource.go
mariantalla: Thanks, I added it.

pkg/scaffold/resource/resource_test.go
mariantalla: Done ✅

pkg/scaffold/v1/resource/resource_test.go
mariantalla: [This test](https://github.com/kubernetes-sigs/kubebuilder/pull/1133) on length and [this one](https://github.com/kubernetes-sigs/kubebuilder/blob/cc8d4ffed8f5d02be97c528998e581a7b8776e13/pkg/scaffold/resource/resource_test.go#L91) on allowed characters cover failed validations. Is there something missing? 🤔

Adirio:
Travis is failing because the golden test files were not regenerated (and we changed year)

mariantalla:
> Travis is failing because the golden test files were not regenerated (and we changed year)

Oh yes... Thanks for the pointer!

Adirio:
#1282 should fix it once it gets merged

pkg/scaffold/resource/resource.go
```diff
@@ -73,9 +73,17 @@ func (r *Resource) Validate() error {
 		return fmt.Errorf(
 			"version must match ^v\\d+(alpha\\d+|beta\\d+)?$ (was %s)", r.Version)
 	}
-	// Check if the Kind is a valid value
-	if r.Kind != flect.Pascalize(r.Kind) {
-		return fmt.Errorf("kind must be PascalCase (expected %s was %s)", flect.Pascalize(r.Kind), r.Kind)
+
+	validationErrors := []string{}
```
Adirio: Just a minor comment? Could this be a `[]error{}`?

pkg/scaffold/resource/resource_test.go
DirectXMan12: table is fine :-)

pkg/scaffold/resource/resource.go
DirectXMan12: agreed with @Adirio  here -- make the the corresponding function return `[]error`

mariantalla: Sure 👍 However, a lot of other functions in that file use the same pattern (collect error messages in a  `[]string`, return an error in the end with all the error messages)

e.g. (from current master): https://github.com/kubernetes-sigs/kubebuilder/blob/cf6cec42bdf594eac67cfb8f13eee6f57ba5c47b/pkg/scaffold/resource/resource.go#L126-L135

It traces back to `func maxLenError(length int) string {` returning a string, so it would be a broader change. It's pretty straightforward, just checking that it would be fine to include in this PR.

pkg/scaffold/resource/resource.go
```diff
@@ -74,7 +74,13 @@ func (r *Resource) Validate() error {
 			"version must match ^v\\d+(alpha\\d+|beta\\d+)?$ (was %s)", r.Version)
 	}
 
-	validationErrors := isDNS1035Label(strings.ToLower(r.Kind))
+	validationErrors := []string{}
+
+	if string(r.Kind[0]) == strings.ToLower(string(r.Kind[0])) {
```
mariantalla: Is this line saying clearly enough "if Kind's first letter is lowercase"?

Kinda wanted to extract in its own function, but didn't think the benefit was worth the indirection 🤔 

@camilamacedo86 @Adirio @DirectXMan12

Adirio: I would say it is pretty clear but a comment before it saying that we check that the first letter is lower case would not hurt anyone.

pkg/scaffold/resource/resource.go
mariantalla: (treats me right for saying `It's pretty straightforward` :) ) This change would also require refactoring test expectations slightly. 

e.g.  `Expect(err).To(MatchError(ContainSubstring("a DNS-1035 label must consist of lower case alphanumeric characters")))` 
would fail if err now holds a slice of `error`s rather than `string`s. Again, happy to go ahead with it, just making sure it would make sense.

DirectXMan12: oh, if the rest of the file does something similar, it's fine to leave for now :-).  Up to you.  Let me know what you decide.

mariantalla: Left it as it was for this PR in the end. I will spend a bit of time right after to experiment, and will open a small cleanup PR if I figure out a way to work with `error`s in `resource.go` while keeping the test cases readable enough.

Adirio:
I'm expecting Travis to fail (do to an unrelated bug) but should be fixed after #1288 gets merged. LGTM otherwise.

mariantalla:
👋 Is there anything else missing here?

Adirio:
I think you may want to rebase to solve the Travis issue and benefit from all the added lint checks that were added in the meantime.

pkg/scaffold/resource/resource.go
mariantalla: Fixed, thanks! Resolving this one.

DirectXMan12:
whoops, lost track of this.  Can merge after rebase.

/assign @mengqiy 
please do the rebase and we'll get this merge.

mengqiy:
@mariantalla Sorry for the delay on this PR.
Can you please rebase, so we can get this merged.
Some files have been refactored recently by @Adirio, it may require a little more work to rebase.

Adirio:
With the recent scaffolding it may be easier to branch from master and copy-paste your checks. `pkg/model/resource/options.go` should be where the check for Kind is placed, in the `Validate` method. `pkg/model/resource/options_test.go` should be where the additional tests go. Want some help?

mariantalla:
Thanks! I'll make a start at rebasing, will poke you if it all goes wrong @Adirio :)

mariantalla:
Did a `go mod tidy` too, which brought in the changes to `go.mod` and `go.sum`; let me know if we don't want these in.

Adirio:
> Did a `go mod tidy` too, which brought in the changes to `go.mod` and `go.sum`; let me know if we don't want these in.

We should probably add a test that checks that go mod tidy was done to Travis so that we do not forget about it.

go.mod
```diff
@@ -4,17 +4,19 @@ go 1.13
...
-	golang.org/x/sys v0.0.0-20190621203818-d432491b9138 // indirect
+	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.4.0 // indirect
+	golang.org/x/net v0.0.0-20191004110552-13f9640d40b9 // indirect
+	golang.org/x/sys v0.0.0-20190826190057-c7b8b68b1456 // indirect
 	golang.org/x/text v0.3.2 // indirect
 	golang.org/x/tools v0.0.0-20190621195816-6e04913cbbac
-	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
+	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
+	gopkg.in/yaml.v2 v2.2.4 // indirect
```
Adirio: I would say that updating versions should not be performed in this PR unless you need one of them. Always updating to the last version makes tagging a version useless as next PR may change the version and it may break the code.

pkg/model/resource/options_test.go
```diff
@@ -73,24 +74,37 @@ var _ = Describe("Resource Options", func() {
...
-			options = &Options{Group: "crew", Version: "v1", Kind: "Firstmate"}
-			Expect(options.Validate()).To(Succeed())
+		DescribeTable("valid Kind values - according to core Kubernetes",
+			func(kind string) {
+				options := &Options{Group: "crew", Kind: kind, Version: "v1"}
+				Expect(options.Validate()).To(Succeed())
+			},
+			Entry("camelcase", "FirstMate"),
+			Entry("more than one caps at the start", "FIRSTMate"),
+		)
```
Adirio: IMHO, I would not use `DescribeTable` here as all the rest of the tests are using raw `Describe`, `Context` and `It`. The added benefit of using `DescribeTable` also seems pretty low, as it is a 2LoC function and 2 entries. So I would just stick to what it is being used already. But thats just an opinion.

pkg/model/resource/options_test.go
```diff
@@ -73,24 +74,37 @@ var _ = Describe("Resource Options", func() {
...
+
+		It("should fail if Kind is too long", func() {
+			var kind string
+
+			for i := 0; i < 65; i++ {
+				kind += "a"
+			}
```
Adirio: `kind := strings.Repeat("a", 64)`

pkg/model/resource/options_test.go
```diff
@@ -73,24 +74,37 @@ var _ = Describe("Resource Options", func() {
...
+			options := &Options{Group: "crew", Kind: kind, Version: "v1"}
+			err := options.Validate()
+			Expect(err).To(MatchError(ContainSubstring("must be no more than 63 characters")))
+		})
 
-			options = &Options{Group: "crew", Version: "v1", Kind: "firstMate"}
-			Expect(options.Validate()).NotTo(Succeed())
-			Expect(options.Validate().Error()).To(ContainSubstring(
-				`kind must be PascalCase (expected FirstMate was firstMate)`))
+		It("should fail if Kind is invalid, according to core Kubernetes", func() {
```
Adirio:
```suggestion
		It("should fail if Kind has whitespaces", func() {
```

pkg/model/resource/options_test.go
```diff
@@ -73,24 +74,37 @@ var _ = Describe("Resource Options", func() {
...
+		It("should fail if Kind is too long", func() {
+			var kind string
+
+			for i := 0; i < 65; i++ {
+				kind += "a"
+			}
+
+			options := &Options{Group: "crew", Kind: kind, Version: "v1"}
+			err := options.Validate()
+			Expect(err).To(MatchError(ContainSubstring("must be no more than 63 characters")))
```
Adirio: I think you should add `Expect(err).To(HaveOccurred())` above this line to avoid panics in case the error is nil.

pkg/model/resource/options_test.go
```diff
@@ -73,24 +74,37 @@ var _ = Describe("Resource Options", func() {
...
+		})
 
-			options = &Options{Group: "crew", Version: "v1", Kind: "firstMate"}
-			Expect(options.Validate()).NotTo(Succeed())
-			Expect(options.Validate().Error()).To(ContainSubstring(
-				`kind must be PascalCase (expected FirstMate was firstMate)`))
+		It("should fail if Kind is invalid, according to core Kubernetes", func() {
+			options := &Options{Group: "crew", Kind: "Something withSpaces", Version: "v1"}
+			err := options.Validate()
+			Expect(err).To(MatchError(ContainSubstring("a DNS-1035 label must consist of lower case alphanumeric characters")))
```
Adirio: I think you should add `Expect(err).To(HaveOccurred())` above this line to avoid panics in case the error is nil.

pkg/model/resource/options_test.go
```diff
@@ -73,24 +74,37 @@ var _ = Describe("Resource Options", func() {
...
+		})
 
-			options = &Options{Group: "crew", Version: "v1", Kind: "firstmate"}
-			Expect(options.Validate()).NotTo(Succeed())
-			Expect(options.Validate().Error()).To(ContainSubstring(
-				`kind must be PascalCase (expected Firstmate was firstmate)`))
+		It("should fail if Kind starts with a lowercase character", func() {
+			options := &Options{Group: "crew", Kind: "lOWERCASESTART", Version: "v1"}
+			err := options.Validate()
+			Expect(err).To(MatchError(ContainSubstring("Kind must start with an uppercase character")))
```
Adirio: I think you should add `Expect(err).To(HaveOccurred())` above this line to avoid panics in case the error is nil.

go.mod
camilamacedo86: Hi @mariantalla,  

I still not getting why the modules need to be updated with these changes. So, I'd suggest you:

- Ensure that the branch is rebased with master
- Revert the changes of go.mod and go.sum
- Run `make generate`

pkg/model/resource/options_test.go
mariantalla: `MatchError(...)` can handle `nil` errors gracefully from what I remember, so `HaveOccured()` is likely redundant 🤔 I'll double check

mariantalla: Yep indeed (for cross reference: [gomega implementation](https://github.com/onsi/gomega/blob/6be6c439588487cd908a3a700795660c2a16dfec/matchers/match_error_matcher.go#L16-L18))

No super strong opinions though, happy to add it if there's significant value in adding the error expectation.

go.mod
mariantalla: Sounds good-checked that out from `master`.

pkg/model/resource/options_test.go
mariantalla: Thanks, changed it 👌

pkg/model/resource/options_test.go
mariantalla: Good point; I've added a small collection of `Kind`s that would be considered invalid instead-of which whitespace is one. My reasoning here is that this test also serves as documentation for future us :) (or other humans that work on this code) on what logic we decided more broadly.

pkg/model/resource/options_test.go
Adirio: Didn't know about this helper until today so nice to learn something new (I should probably apply it in the `pkg/scaffold/internal/machienry` package too). Probably the error messages would be a bit different but they should both clear enough.

pkg/model/resource/options_test.go
```diff
@@ -73,24 +76,37 @@ var _ = Describe("Resource Options", func() {
...
-			Expect(options.Validate().Error()).To(ContainSubstring(
-				`kind must be PascalCase (expected Firstmate was firstmate)`))
+		DescribeTable("invalid Kind values-according to core Kubernetes",
+			func(kind string) {
+				options := &Options{Group: "crew", Kind: kind, Version: "v1"}
+				Expect(options.Validate()).To(MatchError(
+					ContainSubstring("a DNS-1035 label must consist of lower case alphanumeric characters")))
+			},
+			Entry("should fail validation if Kind contains whitespaces", "Something withSpaces"),
+			Entry("should fail validation if Kind ends in -", "Hello-"),
```
Adirio: What about an extra `Entry("should fail validation if Kind starts with a number", "0Hello")`

Adirio:
Otherwise, LGTM.

pkg/model/resource/options_test.go
mariantalla: Good point, added one. Thanks!

PR #1137, opened 2019-10-28
Deprecate v1 projects

### LINKED ISSUE #1136: Deprecate v1 projects
v2 project structure has been out for a while now. We should deprecate v1 projects and eventually (3 months later) drop the support. It will reduce the maintenance burden of v1 projects.
So as a first step, we should start displaying warning indicating `v1 projects are deprecated` and encourage users to migrate to v2 projects whenever they run KB commands on a v1 project.

hypnoglow:
This PR deprecates v1 projects. Fixes #1136

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -28,6 +28,11 @@ import (
 
 	"sigs.k8s.io/kubebuilder/cmd/version"
 	"sigs.k8s.io/kubebuilder/pkg/scaffold"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/project"
+)
+
+const (
+	NoticeColor = "\033[1;36m%s\033[0m"
 )
 
 // module and goMod arg just enough of the output of `go mod edit -json` for our purposes
@@ -112,7 +117,9 @@ func main() {
 	)
 
 	foundProject, version := getProjectVersion()
-	if foundProject && version == "1" {
+	if foundProject && version == project.Version1 {
+		printV1DeprecationWarning()
+
 		rootCmd.AddCommand(
 			newAlphaCommand(),
 			newVendorUpdateCmd(),
@@ -187,4 +194,8 @@ func getProjectVersion() (bool, string) {
 		log.Fatalf("failed to read the PROJECT file: %v", err)
 	}
 	return true, projectInfo.Version
+}
+
+func printV1DeprecationWarning() {
+	fmt.Printf(NoticeColor, "[Deprecation Notice] The v1 projects are deprecated and will not be supported beyond Feb 1, 2020.\nSee how to upgrade your project to v2: https://book.kubebuilder.io/migration/guide.html\n")
 }
--- a/cmd/init_project.go
+++ b/cmd/init_project.go
@@ -115,6 +115,10 @@ func (o *projectOptions) initializeProject() {
 		log.Fatal(err)
 	}
 
+	if o.project.Version == project.Version1 {
+		printV1DeprecationWarning()
+	}
+
 	if err := o.scaffolder.Scaffold(); err != nil {
 		log.Fatalf("error scaffolding project: %v", err)
 	}


cmd/main.go
```diff
@@ -188,3 +191,20 @@ func getProjectVersion() (bool, string) {
 	}
 	return true, projectInfo.Version
 }
+
+func printV1DeprecationWarning() {
+	fmt.Printf(`=============== DEPRECATION WARNING ===============
```
hypnoglow: If this looks too aggressive, let me know how to rephrase/reformat it.

cmd/main.go
```diff
@@ -188,3 +191,20 @@ func getProjectVersion() (bool, string) {
...
+v1 projects are deprecated. The support for v1 projects will be dropped at 1 February 2020.
```
hypnoglow: I dropped an approximate date, please correct me if it is wrong.

Adirio: What about a broad date such as "first 2020 quarter"? @droot

cmd/main.go
droot: We can try making it more concise. How about:
```
warning: [deprecation] v1 projects are deprecated and will not be supported beyond Feb 1, 2020. Refer to https://book.kubebuilder.io/migration/guide.html for more details.
```

cmd/main.go
droot: Giving a specific date makes the message more clear and that's the intent behind this.

cmd/main.go
```diff
@@ -112,7 +113,9 @@ func main() {
 	)
 
 	foundProject, version := getProjectVersion()
-	if foundProject && version == "1" {
+	if foundProject && version == project.Version1 {
```
droot: Cobra command library does provide mechanism to mark commands [as deprecated](https://godoc.org/github.com/spf13/cobra#Command) but I think the implemented approach will work better for our case.

cmd/main.go
hypnoglow: Thanks for the suggestion! Done

cmd/main.go
```diff
@@ -188,3 +191,7 @@ func getProjectVersion() (bool, string) {
 	}
 	return true, projectInfo.Version
 }
+
+func printV1DeprecationWarning() {
+	fmt.Printf("warning: [deprecation] v1 projects are deprecated and will not be supported beyond Feb 1, 2020. Refer to https://book.kubebuilder.io/migration/guide.html for more details.\n")
```
camilamacedo86: WDYT about: 
```suggestion
	fmt.Printf("warning: [deprecation] v1 projects are deprecated and will not be supported beyond Feb 1, 2020. See the https://book.kubebuilder.io/migration/guide.html for know how to migrate the project to from v1 to v2.\n")
```

hypnoglow: @camilamacedo86 Is the "See ... **for know** how to ..." correct English? Sorry for this question, I'm not a native speaker, thus I never heard this idiom :)

cmd/main.go
camilamacedo86:
```suggestion
	if hasProjectFile && projectVersion == project.Version1 {
```
WDYT?

Adirio: Why do you suggest to do this change from another PR (#1143) into this?

camilamacedo86: Hi @Adirio, 

Because this PR is doing/replacing `if foundProject && version == "1" {` for `if foundProject && version == project.Version1 {`. 

So, as part of its review, I am suggesting `if foundProject && version == "1" {` for `if hasProjectFile && projectVersion == project.Version1 {` instead of.

hypnoglow: I agree that @camilamacedo86 suggested naming is more suitable. But I also agree that we can lay it on #1143, as the narrow scope of my change was only to replace the string `"1"` with the constant `project.Version1`, and not to change other things.

If you still think we should introduce this change here, let me know.

cmd/main.go
hypnoglow: @camilamacedo86 ping, I would be thankful if you clarify this moment for me.

cmd/main.go
Adirio: @hypnoglow The change she is suggesting is not just a more suitable name, it has to do with the package with the same alias (`version`). The context is missing as this is a change I suggested in #1143 and she suggested a name change. Suggesting to change this here makes no sense as this PR is about a completely different topic.

cmd/main.go
Adirio: `... for know how to ...` and `project to from` are definetely typos. I would go for `... Refer to https://.. for instructions on how to migrate to v2.\n`. Saying from `v1` doesn't give that much info as you already said that `v1` is the one being deprecated.

Just my 2 cents.

hypnoglow: @Adirio agreed, thanks for help.

cmd/main.go
camilamacedo86: It is not a big deal so I am OK with.

cmd/main.go
```diff
@@ -188,3 +191,7 @@ func getProjectVersion() (bool, string) {
 	}
 	return true, projectInfo.Version
 }
+
+func printV1DeprecationWarning() {
+	fmt.Printf("warning: [deprecation] v1 projects are deprecated and will not be supported beyond Feb 1, 2020. Refer to https://book.kubebuilder.io/migration/guide.html for instructions on how to migrate to v2.\n")
```
camilamacedo86:
```suggestion
	log.Printf(NoticeColor,"[Deprecation Notice] - The v1 projects are deprecated and will not be supported beyond Feb 1, 2020.\n See how to upgrade your project to v2: https://book.kubebuilder.io/migration/guide.html\n")
```
For it noticed ;-) WDYT?

Screenshot 2019-11-07 at 00 58 51

cmd/main.go
camilamacedo86: I was not clear in my suggestion. Sorry. It is updated following.

cmd/main.go
hypnoglow: Done

PR #1143, opened 2019-10-31
Solve name collision and document why version strings can not be const

Adirio:
Change list:
- Avoid variable and package name collision
- Stop exporting get version function
- Document why version strings cannot be constants

/kind cleanup

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -116,8 +116,8 @@ func main() {
 		version.NewVersionCmd(),
 	)
 
-	foundProject, version := getProjectVersion()
-	if foundProject && version == project.Version1 {
+	foundProject, projectVersion := getProjectVersion()
+	if foundProject && projectVersion == project.Version1 {
 		printV1DeprecationWarning()
 
 		rootCmd.AddCommand(
--- a/cmd/version/version.go
+++ b/cmd/version/version.go
@@ -22,6 +22,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// var needs to be used instead of const as ldflags is used to fill this
+// information in the release process
 var (
 	kubeBuilderVersion      = "unknown"
 	kubernetesVendorVersion = "unknown"
@@ -41,7 +43,7 @@ type Version struct {
 	GoArch             string `json:"goArch"`
 }
 
-func GetVersion() Version {
+func getVersion() Version {
 	return Version{
 		kubeBuilderVersion,
 		kubernetesVendorVersion,
@@ -67,5 +69,5 @@ func NewVersionCmd() *cobra.Command {
 }
 
 func runVersion(_ *cobra.Command, _ []string) {
-	GetVersion().Print()
+	getVersion().Print()
 }


cmd/init_project.go
```diff
@@ -95,8 +95,8 @@ func (o *projectOptions) bindCmdlineFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&o.dep, "dep", true, "if specified, determines whether dep will be used.")
 	o.depFlag = cmd.Flag("dep")
 	cmd.Flags().StringArrayVar(&o.depArgs, "depArgs", nil, "Additional arguments for dep")
-	cmd.Flags().MarkDeprecated("dep", "use the fetch-deps flag instead")
-	cmd.Flags().MarkDeprecated("depArgs", "will be removed with version 1 scaffolding")
+	_ = cmd.Flags().MarkDeprecated("dep", "use the fetch-deps flag instead")
```
camilamacedo86: Why it `_` would be required here?

cmd/init_project.go
```diff
@@ -186,16 +186,16 @@ func fetchAndCheckGoVersion() error {
 	cmd := exec.Command("go", "version")
 	out, err := cmd.Output()
 	if err != nil {
-		return fmt.Errorf("Failed to retrieve 'go version': %v", string(out))
+		return fmt.Errorf("failed to retrieve 'go version': %v", string(out))
```
camilamacedo86: Note that all project use capital letters to do the outputs. So, why would be required do this change?

cmd/main.go
```diff
@@ -98,7 +98,9 @@ func findCurrentRepo() (string, error) {
 		// give up, let the user figure it out
 		return "", fmt.Errorf("could not determine repository path from module data, package data, or by initializing a module: %v", err)
 	}
-	defer os.Remove("go.mod") // clean up after ourselves
+	defer func() {
+		_ = os.Remove("go.mod")
+	}() // clean up after ourselves
```
camilamacedo86: What is the motivation of this change?

cmd/main.go
```diff
@@ -111,8 +113,8 @@ func main() {
 		version.NewVersionCmd(),
 	)
 
-	foundProject, version := getProjectVersion()
-	if foundProject && version == "1" {
+	foundProject, versionString := getProjectVersion()
+	if foundProject && versionString == "1" {
```
camilamacedo86: I do not think that is a good practice to add the type of var in the nomenclature of it.

cmd/version/version.go
```diff
@@ -22,7 +22,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var (
+const (
```
camilamacedo86: 👍

cmd/version/version.go
```diff
@@ -41,15 +41,13 @@ type Version struct {
...
-		goarch,
-	}
+var version = Version{
+	kubeBuilderVersion,
+	kubernetesVendorVersion,
+	gitCommit,
+	buildDate,
+	goos,
+	goarch,
 }
```
camilamacedo86: Keep as it was before is clear and do more performative.

cmd/version/version.go
```diff
@@ -66,6 +64,6 @@ func NewVersionCmd() *cobra.Command {
 	}
 }
 
-func runVersion(cmd *cobra.Command, args []string) {
-	GetVersion().Print()
+func runVersion(_ *cobra.Command, _ []string) {
```
camilamacedo86: What is the motivation of this change?

cmd/init_project.go
Adirio: This explicitly "handles" the error by not doing anything. It is recommended by some IDEs and I find it a good way to show that there is something being returned that we are not using.

cmd/init_project.go
Adirio: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

cmd/main.go
Adirio: Same as with the previous change, the error is explicitly handled by using `_`. The way to do so is using a wrapper funtion.

cmd/main.go
Adirio: `version` is a package, what do you suggest?

cmd/version/version.go
Adirio: Building a Version each time is not more performant. About the clearness, that could be discussed.

cmd/version/version.go
Adirio: `cmd` and `args` are not used inside the function, so no need to refer them.
https://golang.org/ref/spec#Blank_identifier

Adirio:
> In order to help you with the review, I added so comments. Please feel free to check it.

Done. I added some answers justifying some of the changes you asked about.

cmd/init_project.go
camilamacedo86: So, it is a specific motivation and then IHMO should have a PR just for that where you would do this change across the repo. Note that in order to make clear what is done in the PR we are able to track the changes each PR should have a specific and clear motivation. So, wdyt about split it?

cmd/main.go
camilamacedo86: The same as above. Note that `Command clean up` is not leading a good understanding of your motivation and what and why is done here. Also, I'd recommend you look for the golang good practices and check if it is actually recommended for it in order to have a good argumentation/justification for this change. It is important to highlight that in POV the changes could not be done in just a single place in this scenario and the same standard should be applied across all repo.

cmd/main.go
camilamacedo86: My suggestion would be:

```suggestion
	if isProjectFound && projectVersion == "1" {
```
WDYT? Note that I'd use here the prefix is because is a boolean. (the nomenclature to make clear the understanding for boolean is to use `is` or `has` as a prefix)

Also, note that I'd apply the same above recommendation over it as well. I understand that your motivation here is to improve the understanding of the code in this func. So, it shows enough for we do a PR just with it.

cmd/version/version.go
camilamacedo86: I'd recommend you do like this [example](https://github.com/dev4devs-com/postgresql-operator/blob/master/pkg/config/backup.go). Note that has no need to use the memory space to store the version var and we can return it directly.

Adirio:
@camilamacedo86 I answered some of your comments. The three motivations you mentioned are listed in the commit text and in the PR message, with a couple more. `version.go` modifications are not only aimed towards understandability: `version` is a package name so using it as a variable is a name collision, the other two changes, while they may increase the performance in a negligible manner, could be considered aimed towards understandability.

As the title states, it is just a cleanup, tidying up some random stuff. You may be right that some of these changes should be done for the whole project and not only the `cmd` package. I would like to hear other opinions about this (@droot @pwittrock @mengqiy @DirectXMan12). Expecifically, about using explicit `_` for unhandled errors and using lowercase and no final dot for error messages (https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

cmd/main.go
camilamacedo86: Hi @Adirio, 

After your comment, my recommendation/suggestion would be hasProjectFile since it is a boolean. 

Just to share. But, it is a really tiny thing so. I am ok with how it is today. 

Note that: `Booleans can hold only 2 values, true or false. Given this, using prefixes like “is”, “has”, and “can” will help the reader infer the type of the variable.` From [ref](https://hackernoon.com/the-art-of-naming-variables-52f44de00aad). However, if you looking for it you will see that in order to follow the nomenclature good practices for booleans usually, we should use `is` or `has` as a prefix.  See more few refs [1](https://www.samanthaming.com/tidbits/34-better-boolean-variable-names) and [2](http://bouncetadiss.blogspot.com/2012/04/good-practice-when-naming-boolean.html)

camilamacedo86:
HI @Adirio, 

It is fine. Just to clarifies the key points of my review here: 

As a good practice no matter the project that you collab with you should keep a small context per PR. PR's should not address many motivations/purposes in order to keep simple and easier others understand what is going on, do the reviews, track the changes and  etc.. 

Also, note that IMHO make no sense apply motivations like `Error strings should not be capitalized in order to follow the good practices` just in one file. If it should not be done then it should be changed in all repo.

Hi @droot @pwittrock @mengqiy @DirectXMan12, could you please give a hand here and let us know wdyt about these changes?

cmd/version/version.go
```diff
@@ -41,15 +41,13 @@ type Version struct {
 	GoArch             string `json:"goArch"`
 }
 
-func GetVersion() Version {
```
mengqiy: I wonder if we are doing this for release purpose. @droot can you confirm if this change is OK?

droot: This is used to print the version when user runs `kubebuilder version` command. This change doesn't buy much. Making it pkg private by renaming `getVersion` makes sense.

cmd/version/version.go
droot: Performance is not a concern here. Given that we don't have a good test coverage for this code and also `version` specific changes requires testing using build/release workflow (that is always a pain), I would recommend keeping changes to minimal here and may be making this method private `GetVersion` --> `getVersion` will be more reasonable change here.

Adirio:
Ok, I will split this PR into the different parts following your comments on monday

Adirio:
Simplified the PR to follow the comments and letting the formatting and error handling parts to their specific PR in a project-wide scope.

Adirio:
> If we have cleanup all of the linter errors, it'd be great to enable the linters like controller-runtime does.
> Ref: https://github.com/kubernetes-sigs/controller-runtime/blob/81842d0e78f7111f0566156189806e2801e3adf1/hack/verify.sh#L27-L45

I started #1176 to track all changed that may be needed for this purpose.

mengqiy:
@Adirio  Hmm, it seems it conflicts with your other PR. Can you please rebase?

mengqiy:
Please squash the commits.

Adirio:
I resolved the conflict in GitHub tool.

Tomorrow I will be able to do it with my normal development environment and squash the commits if you think the commit history would be cleaner.

mengqiy:
No rush. We can wait til tomorrow.
BTW can you update the PR description? It's a little out of date.

cmd/main.go
```diff
@@ -116,8 +116,8 @@ func main() {
 		version.NewVersionCmd(),
 	)
 
-	foundProject, version := getProjectVersion()
-	if foundProject && version == project.Version1 {
+	foundProject, projectVersion := getProjectVersion()
+	if foundProject && projectVersion == project.Version1 {
```
camilamacedo86:
```suggestion
	if isProjectFound && projectVersion == project.Version1 {
```

Adirio:
@mengqiy 
> No rush. We can wait til tomorrow.

👍 
> BTW can you update the PR description? It's a little out of date.

Done.

@camilamacedo86
> * small nit replace `foundProject ` for `isProjectFound `

I think we already discussed this. `isProjectFound` sounds really strange as it is a verb followed by a noun and then by an adjetive. If any alternative variable name was to be used `hasProjectFile` sounds better. The thing is that this PR is not changing `version` because the new name suits better, but because there is already a package called like that. So it is solving a package and variable name collision, which is a project-wide change. Yes, it is changing a variable name in the same line, but the motivation is different and I would like to enphasize that. If you want to change all boolean variable names to `isXxxxx` or `hasXxxxx` I think it makes more sense to have a separate PR do it and make it be project-wide.

camilamacedo86:
Hi @Adirio hasProjectFile sounds better 🥇 
I think it is too small to be another one and could be done here .. however.. it is fine if you didn't

Adirio:
@mengqiy Done

Issue #1144, opened 2019-10-31
Extract Resource from the v1 Scaffolding package


Adirio:
The `Resource` type is being used by both scaffolding versions, but it is placed inside v1, which is missleading. Furthermore, v1 scaffolding is being deprecated (#1136, #1137).

Shouldn't this package be extracted to the root scaffolding package?

/kind feature

DirectXMan12:
yeah, it should be.  cc #1133, in which I think there was confusion about this

PR #1146, opened 2019-10-31
Remove duplicated scaffold API validation code

### LINKED ISSUE #1145: Remove duplicated scaffold API validation code
[`API.Validate()`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/scaffold/api.go#L60-L68) is reimplementing part of the [`Resource.Validate()`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/scaffold/v1/resource/resource.go#L57-L65) functionality. Shouldn't it just call the `Resource.Validate()` method?

/kind bug

Adirio:
Delegate resource validation to itself

Fixes #1145

### DIFF
--- a/pkg/scaffold/api.go
+++ b/pkg/scaffold/api.go
@@ -57,14 +57,8 @@ func (api *API) Validate() error {
 	if err := api.setDefaults(); err != nil {
 		return err
 	}
-	if api.Resource.Group == "" {
-		return fmt.Errorf("missing group information for resource")
-	}
-	if api.Resource.Version == "" {
-		return fmt.Errorf("missing version information for resource")
-	}
-	if api.Resource.Kind == "" {
-		return fmt.Errorf("missing kind information for resource")
+	if err := api.Resource.Validate(); err != nil {
+		return err
 	}
 
 	return nil


Adirio:
This error has nothing to do with my code, you can check it with the base kubebuilder and you will get the same.

camilamacedo86:
HI @Adirio, 

IHMO what we can do regards this subject to collab with is to add a test for each scenario and ensure that the correct message will be returned.

Adirio:
As I mentioned in #1144, the `Resource` inside the `v1` package is being used for both `v1` and `v2`. That's the reason I suggest to move it outside of the `v1` package.

The link you provided is exactly the function I'm calling in this PR (`Resource.Validate`). As you can see that function does pretty much the same thing as the code I deleted in this PR. Instead of checking in a string is `""` it checks that its length is `0`, but that are two ways of doing the same. The messages are also a bit different, the ones I deleted may be a bit more descriptive so we could keep these instead of the ones in the file you linked. Additionally it does some extra checks which you are surely aware of as you recently modified some of them.

<details><summary><code>Resource.Validate</code> function usage:</summary>
It can be called either directly or thorugh `input.Validate` interface. The interface is only being called inside `Scaffold.Execute` which basically calls `Validate` on any file that has a method `Validate`. The direct method is being called in 33 tests (`pkg/scaffold/v1/resource/resource_test.go`), 8 v1 file types (`pkg/scaffold/v1/resource/`) and 7 v2 file types (`pkg/scaffold/v2/`, `pkg/scaffold/v2/crd/` and `pkg/scaffold/v2/webhook/`). The 15 file types for both versions just implement the `input.Validate` interface by calling the wrapped `Resource`, so the interface usage mention above applies to them. The 33 tests check most, if not all, of the corner cases directly on `Resource`s.
</details>

As seen in the `Resource.Validate` function usage found above, this function is only being called at `Scaffold.Execute`. As seen in this PR, part of the `Resource.Validate` code was being duplicated to check for empty strings when issuing a `kubebuilder create api ...` command. This PR just delegates the validation to the already tested and more comprehensive `Resource.Validate` function instead of having duplicated code.

About the bug you mentioned, it is not being catched by the 33 tests as the resources are being created directly instead of though the command line. I will create a bug report for it to track it separately.

camilamacedo86:
Hi Adario, 

This change I need to do here https://github.com/kubernetes-sigs/kubebuilder/pull/1163 to solve the issue raised by you. Do you mind in close this one and we keep all there?

PR #1147, opened 2019-11-01
Add webhook scaffolding to docs

anyasabo:
Adds the command to scaffold out webhooks to the docs. It was in the migration guide, but the main docs were incomplete. If you didn't run the scaffolding and just satisfy the validator/defaulter interfaces as mentioned, you would not actually be calling `SetupWebhookWithManager()` in your main and you might be confused why the thing that it says automatically happens does not in fact automatically happen.

### DIFF
--- a/docs/book/src/reference/markers/webhook.md
+++ b/docs/book/src/reference/markers/webhook.md
@@ -1,6 +1,6 @@
 # Webhook
 
-These markers describe how [webhook configuration](/TODO.md) is generated.
+These markers describe how [webhook configuration](../webhook-overview.md) is generated.
 Use these to keep the description of your webhooks close to the code that
 implements them.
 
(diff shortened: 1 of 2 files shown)


docs/book/src/reference/markers/webhook.md
```diff
@@ -1,6 +1,6 @@
 # Webhook
 
-These markers describe how [webhook configuration](/TODO.md) is generated.
+These markers describe how [webhook configuration](../webhook-overview.md) is generated.
```
mengqiy: The document should be: https://book.kubebuilder.io/reference/markers/webhook.html

anyasabo: This doc is the marker reference (the URL you shared), so it seemed like this was supposed to go back to the overall webhook overview, correct?

mengqiy: I wrote the webhook doc initially. The intent was to point to the webhook section of the marker doc, which was not available when I wrote it. We forgot to fix `TODO.md` when we got our marker document.

anyasabo: I think I am definitely misunderstanding something. This file is:
docs/book/src/reference/markers/webhook.md
which is what ends up in:
https://book.kubebuilder.io/reference/markers/webhook.html

So if I'm understanding your suggestion correctly it would be a link to itself?

Issue #1042, opened 2019-09-28
Use gomod go get v2 error


644262163:
I have a project that I want to import this project

## go version
```
go version go1.13 linux/amd64
```
### 1
```
$ go get sigs.k8s.io/kubebuilder@v2.0.0

go get sigs.k8s.io/kubebuilder@v2.0.0: sigs.k8s.io/kubebuilder@v2.0.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

### 2
```
$ go get sigs.k8s.io/kubebuilder/v2@v2.0.0

go get sigs.k8s.io/kubebuilder/v2@v2.0.0: sigs.k8s.io/kubebuilder@v2.0.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

### 3
```
$ cat go.mod

module test
go 1.13
require sigs.k8s.io/kubebuilder v2.0.0+incompatible

$ go get

verifying sigs.k8s.io/kubebuilder@v2.0.0+incompatible/go.mod: sigs.k8s.io/kubebuilder@v2.0.0+incompatible/go.mod: reading https://sum.golang.org/lookup/sigs.k8s.io/kubebuilder@v2.0.0+incompatible: 410 Gone
```

### 4
```
$ cat go.mod

module test
go 1.13
require sigs.k8s.io/kubebuilder v2.0.0

$ go get

/test/go.mod:3: require sigs.k8s.io/kubebuilder: version "v2.0.0" invalid: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

### 5
```
$ cat go.mod

module test
go 1.13
require sigs.k8s.io/kubebuilder/v2 v2.0.0

$ go get

go: sigs.k8s.io/kubebuilder/v2@v2.0.0: go.mod has non-.../v2 module path "sigs.k8s.io/kubebuilder" (and .../v2/go.mod does not exist) at revision v2.0.0
```
/kind bug

camilamacedo86:
Could you try to use go 1.12? I believe that the version 1.13 is not supported yet.

DirectXMan12:
Go 1.13 is supported.  This is a bug.

DirectXMan12:
/priority backlog

backlog since we do binary releases and don't expose a public API.

We should fix this eventually, though.

camilamacedo86:
Great to know that it is supporting 1.13 already :-)

[...]

PR #1148, opened 2019-11-04
:running: structure and expose cobra cmds

hasbro17:
**Description:**
This restructures the cmd directory into pkgs that follows the
hierarchy of CLI subcommands. This helps to expose the `New***Command()` functions
that return the cobra subcommands so they can be consumed by downstream
projects that need to reuse Kubebuilder CLI for scaffolding projects.

The new pkg layout for cmd:
```console
cmd
├── alpha
│   ├── alpha.go
│   └── webhookv1
│       └── webhook_v1.go
├── create
│   ├── api
│   │   └── api.go
│   ├── create.go
│   └── webhookv2
│       └── webhook_v2.go
├── init
│   ├── go_version_test.go
│   └── init_project.go
├── main.go
├── update
│   └── vendor_update.go
├── util
│   ├── stdin.go
│   └── util.go
└── version
    └── version.go
```

**Motivation:**
This follows up on the [proposal](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/integrating-kubebuilder-and-osdk.md) to integrate the Operator SDK with Kubebuilder.
The goal here is to expose the cobra commands for `kubebuilder init` and `kubebuilder create` so that the SDK can consume them downstream in its own CLI.
This allow the SDK to easily reuse Kubebuilder's scaffolding and CLI workflow.

While the plugins-based-extensions approach would still be a long term goal, consuming the Kubebuilder CLI in the SDK in the short run allows us to prototype and make changes in the SDK to make it compatible with Kubebuilder's project layout and expected workflow.


hasbro17:
While this does expose all the cobra subcommands functions, the ones strictly needed for our purposes are the `NewCreateCmd()` and `NewInitProjectCmd()`, but since the other subcommands are no longer in pkg `main` they need to be exposed so they can be imported by their parent commands. 
We can move the other subcommand pkgs to `cmd/internal/...` if needed however.

Also since this effectively exposes a public API in Kubebuilder, #1042 would eventually need a fix to allow importing this as `sigs.k8s.io/kubebuilder/v2` or `sigs.k8s.io/kubebuilder/v3`. Although this can still be worked around by pinning to the commit so probably not that pressing.

DirectXMan12:
o_O I was under the impression that OperatorSDK wouldn't being doing a separate CLI like this (hence the plugin stuff)?

/hold

DirectXMan12:
@joelanford

DirectXMan12:
> While the plugins-based-extensions approach would still be a long term goal, consuming the Kubebuilder CLI in the SDK in the short run allows us to prototype and make changes in the SDK to make it compatible with Kubebuilder's project layout and expected workflow.

This makes me inherently nervous, since it means that things are depending on the internals, more or less.  Can we chat about this next week at the meeting?  I was imagining prototyping would be done via "fork & custom plugin" or something similar, and then any changes to the plugin subsystem would be resubmitted.

joelanford:
@DirectXMan12 Sorry, we didn't mean to catch you off guard with this. Let's chat more at the community meeting next week. We can talk through our thought process and discuss if there are any other alternatives.

hasbro17:
Closing this per our offline discussion to explore a different strategy of reusing the plugin interface without exposing the CLI directly.

PR #1158, opened 2019-11-06
Update scheme handling in the book

### LINKED ISSUE #1118: Update scheme handling in the book
# What do you want to happen?
[Section 1.7.1](https://book.kubebuilder.io/cronjob-tutorial/main-revisited.html) of the book claims that we need to add `k8s.io/api/batch/v1` to the scheme as we are using `Job`s. `k8s.io/client-go/kubernetes/scheme` already does that for us. Update that part to explain that we do not need to add builtin types but we do need to add new types.

/kind documentation

Adirio:
# Description:
Update the book to remove outdated add of builtin scheme.

Fixes #1118

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/main.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/main.go
@@ -20,7 +20,6 @@ import (
 	"flag"
 	"os"
 
-	kbatchv1 "k8s.io/api/batch/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
@@ -38,8 +37,8 @@ The first difference to notice is that kubebuilder has added the new API
 group's package (`batchv1`) to our scheme.  This means that we can use those
 objects in our controller.
 
-We'll also need to add the kubernetes batch v1 (`kbatchv1`) scheme, since we're creating
-and listing Jobs.
+If we would be using any other CRD we would have to add their scheme the same way.
+Builtin types such as Job have their scheme added by `clientgoscheme`.
 */
 var (
 	scheme   = runtime.NewScheme()
@@ -49,7 +48,6 @@ var (
 func init() {
 	_ = clientgoscheme.AddToScheme(scheme)
 
-	_ = kbatchv1.AddToScheme(scheme) // we've added this ourselves
 	_ = batchv1.AddToScheme(scheme)
 	// +kubebuilder:scaffold:scheme
 }


Adirio:
I also modified the testdata accordingly, don't know if that will pass the golden test. As I'm working on Windows currently, I will try to solve any issues with that test by reading Travis log.

testdata/project-v2/main.go
```diff
@@ -20,7 +20,6 @@ import (
 	"flag"
 	"os"
 
-	corev1 "k8s.io/api/core/v1"
```
mengqiy: You are changing the generated file. You also need to change the scaffolding.
Otherwise the golden test will fail.

mengqiy: https://github.com/kubernetes-sigs/kubebuilder/blob/3202ae8d778fc10680abf29366604034db0d98d9/pkg/scaffold/v2/main.go#L127

Adirio: But there is no `corev1 "k8s.io/api/core/v1"` in the scaffolding. I thought it was introduced manually but it may be introduced by some other file. `generate_golden.sh` maybe?

Adirio: https://github.com/kubernetes-sigs/kubebuilder/blob/3202ae8d778fc10680abf29366604034db0d98d9/generate_golden.sh#L74

Is this the line to blame?

mengqiy: This is generated by https://github.com/kubernetes-sigs/kubebuilder/blob/eb3d7967cae29ba8bcf6893f359621551523bdeb/pkg/scaffold/v2/main.go#L51-L110
We currently just blindly add the import and the scheme, though corev1 has already been added.
It doesn't cause any problem by adding a scheme twice.

mengqiy: One way to fix this is using [`KnownTypes`](https://github.com/kubernetes/apimachinery/blob/6eb29fdf75dcc3a9e33ec30ef32b25c923789dd1/pkg/runtime/scheme.go#L216-L227) to check if a <group, version> already exists before update it in `(m *Main) Update(opts *MainUpdateOptions)`.

Adirio: The problem is that we do not have an scheme as it is scaffolding the scheme. I guess we could make a copy but that would add as dependancies `k8s.io/apimachinery` for `NewScheme` and `KnownTypes`, and `k8s.io/client-go` for `AddToScheme`. What do you think about this dependencies?

Adirio:
/hold
Not closing yet because there is an ongoing discussion, but the added value of this PR is pretty small for the dependencies that it will add.

testdata/project-v2/main.go
mengqiy: Thanks for bringing this up.
IMO it doesn't worth to introduce dependencies on client-go for this reason.
We are always trying to avoid having dependencies on k8s repos, since kubebuilder CLI is just a scaffolding tool and we don't want to confuse users that one kubebuilder release only works with a specific k8s version(s).

Adirio:
/hold cancel

I left the testdata as is, as removing already registered schemas by client-go would require to import `k8s.io/apimachinery` and `k8s.io/client-go`. The book related change is still doable.

Adirio:
@mengqiy I removed the part that would require adding dependencies, but the one that was added manually in the docs has been removed and the explanation improved. Should we merge this small change?

PR #1162, opened 2019-11-08
Add Schema Reconciler attr in the tutorial and plugins scaffold impl

camilamacedo86:
**Description**
Add Schema Reconciler attr in the tutorial, design doc and plugins scaffold impl

**Motivation**
Checked that this places still missing the value/updated.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/emptycontroller.go
+++ b/docs/book/src/cronjob-tutorial/testdata/emptycontroller.go
@@ -23,6 +23,7 @@ package controllers
 
 import (
 	"context"
+	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/go-logr/logr"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -41,6 +42,7 @@ objects, so these are added out of the box.
 type CronJobReconciler struct {
 	client.Client
 	Log logr.Logger
+	Scheme *runtime.Scheme
 }
 
 /*
--- a/plugins/addon/controller.go
+++ b/plugins/addon/controller.go
@@ -42,6 +42,7 @@ import (
 	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative"
 
 	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -54,6 +55,7 @@ var _ reconcile.Reconciler = &{{ .Resource.Kind }}Reconciler{}
 type {{ .Resource.Kind }}Reconciler struct {
 	client.Client
 	Log logr.Logger
+	Scheme *runtime.Scheme
 
 	declarative.Reconciler
 }


mengqiy:
I don't understand why we want to add scheme field to the reconciler. It is not used in the tutorial.
IMO we should only keep necessary stuff in the tutorial, otherwise it may add more complexity for users to understand.

camilamacedo86:
HI @mengqiy,

The schema is required for we add the ownership in the resources managed by it like the following example. I mean, be able to use the `ctrl.SetControllerReference(m, ser, r.Scheme)` to add the owner of the secondary objects built in the project.

```go
// serviceForMemcached function takes in a Memcached object and returns a Service for that object.
func (r *MemcachedReconciler) serviceForMemcached(m *memcachedv1.Memcached) *corev1.Service {
	ls := labelsForMemcached(m.Name)
	ser := &corev1.Service{
		ObjectMeta: v1.ObjectMeta{
			Name:      m.Name,
			Namespace: m.Namespace,
		},
		Spec: corev1.ServiceSpec{
			Selector: ls,
			Ports: []corev1.ServicePort{
				{
					Port: 11211,
					Name: m.Name,
				},
			},
		},
	}
	// Set Memcached instance as the owner of the Service.
	ctrl.SetControllerReference(m, ser, r.Scheme) //todo check how to get the schema
	return ser
}
```


Then, it was introduced here: https://github.com/kubernetes-sigs/kubebuilder/pull/1054. However, I notice that we did not update the section in the tutorial which describes the gen controller impl and the design doc. Also, the plugins were after introduced without this.

designs/simplified-scaffolding.md
```diff
@@ -468,6 +469,7 @@ import (
 type MyKindReconciler struct {
 	client.Client
 	log logr.Logger
+	Scheme *runtime.Scheme
```
Adirio: I don't think you should modify this document. Two reasons:
- As said by @mengqiy, as this example doesn't require the `Scheme` adding it just adds compelxity.
- It is a design document, it does not have to reflect future changes.

docs/book/src/cronjob-tutorial/testdata/emptycontroller.go
```diff
@@ -41,6 +42,7 @@ objects, so these are added out of the box.
 type CronJobReconciler struct {
 	client.Client
 	Log logr.Logger
+	Scheme *runtime.Scheme
```
Adirio: This file is used at the start of the book as an **empty** controller. The scheme is introduced later in [section 1.7](https://book.kubebuilder.io/cronjob-tutorial/controller-implementation.html). The corresponding file is `docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller.go`. In that page it is explained why do we add the `Scheme`, so I again think that you should not add the `Scheme` in the **empty** controller.

plugins/addon/controller.go
```diff
@@ -54,6 +55,7 @@ var _ reconcile.Reconciler = &{{ .Resource.Kind }}Reconciler{}
 type {{ .Resource.Kind }}Reconciler struct {
 	client.Client
 	Log logr.Logger
+	Scheme *runtime.Scheme
```
Adirio: Plugins may not need an Scheme, why should you enforce them to have that field?

docs/book/src/cronjob-tutorial/testdata/emptycontroller.go
camilamacedo86: The empty controller now, after this section is written, will have the scheme attribute by . default so IMO makes sense the doc also be updated.

plugins/addon/controller.go
camilamacedo86: Please, see the above explanation which described why it is required.

docs/book/src/cronjob-tutorial/testdata/emptycontroller.go
Adirio: I missed #1054. This change does make perfect sense.

plugins/addon/controller.go
Adirio: Same as above, mark as resolved please.

designs/simplified-scaffolding.md
Adirio: I still think you should not modify the design document.

Adirio:
As an additional comment, you should edit [section 1.7](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller.go#L43-L48).
 I would suggest:
```
/*
Next, we'll need a Clock, which will allow us to fake timing in our tests.
*/
```

designs/simplified-scaffolding.md
mengqiy: +1

camilamacedo86:
HI @Adirio, 

I do not see reason to change section 1.7, since it is already with the schema. Note that has reasons for in the tests we highlight the importance of its usage.

Adirio:
@camilamacedo86

The point is that the comment in 1.7 says that we are adding the Scheme and the Clock, when that's not true, as the Scheme is part of the "empty" controller. I mean, the Scheme is not something special from the CronJob controller but part of every controller (as a result of this and the previous related PR) so it should not be presented as added manually in section 1.7.

docs/book/src/cronjob-tutorial/testdata/project/go.mod
```diff
@@ -5,7 +5,6 @@ go 1.12
 require (
 	github.com/go-logr/logr v0.1.0
 	github.com/robfig/cron v1.2.0
-	golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2 // indirect
```
Adirio: Non-blocking: Can I ask why are you removing this? I already asked it but I didn't understand your point. I just ask this because my IDE is automatically updating two go-mod files already adding this line because it does think it is needed. I would like to understand what is the reason behind this removal.

camilamacedo86:
hi @Adirio, 

The text is: 

> /*
> Next, we'll need a few more bits in our Reconciler:
> We'll need the Scheme, in order to call some helpers that set owner references,
> and we'll need a Clock, which will allow us to fake timing in our tests.
> */ 

Then, IMO the explanation does not go against as it is and is very important to highlight here in the tests **We'll need the Scheme, in order to call some helpers that set owner references,**. For a further understanding see: https://github.com/kubernetes-sigs/kubebuilder/pull/1175/files

Adirio:
@camilamacedo86 I think my suggestion should have been to move instead of remove, let me improve my suggestion and explain it clearly.

The documentation of the controller struct is documented in two places: sections 1.6 ([`.../emptycontroller.go`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/emptycontroller.go#L34-L38)) and 1.7 ([`.../project/controllers/cronjob_controller.go`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller.go#L43-L48)). In the first, the fields that are generated by the scaffold are described, while in the second, the ones that are added manually. With PR1054 `Scheme` was made one of the first kind, and this PR updates the book to reflect that change.

My sugeestion should have been moving the description of `Scheme` from section 1.7 to 1.6, so that it is still explained, but in the relevant section. I add some suggestions on how I would write these two parts:

`.../emptycontroller.go#L34-L38`
```
/*
Next, kubebuilder has scaffold out a basic reconciler struct for us.
Pretty much every reconciler needs to log, and needs to be able to fetch
objects, so a logger and a client are added out of the box.
Additionally, some client helpers require to provide the scheme of the
objects you are working with, so it is also injected.
*/
```
`.../project/controllers/cronjob_controller.go#L43-L48`
```
/*
Next, we'll need a few more bits in our Reconciler: a Clock, which will
allow us to fake timing in our tests.
*/
```

camilamacedo86:
Hi @Adirio, 

Sorry, I still not agree because in POV our suggested text is not accurate at all and I think it is better with the current one. Let's keep as it is then if you would like please feel free to do a new PR after that where we can check better and discuss this specific point.

Adirio:
> Sorry, I still not agree because in POV our suggested text is not accurate at all and I think it is better with the current one(reason described above. It is important to highlight the schema and ownership for the tests).

Why do you say it "is not accurate at all"? My text includes nearlly all the information that is currently in the two comments, just reorder to reflect the change done in these 2 PR (making `Scheme` part of the default scaffold of controllers). The only thing that is different is not mentioning "set owner references". Is this missing part the one you think that makes this "not accurate at all"? It can be easily included:
`.../emptycontroller.go#L34-L38`
```
/*
Next, kubebuilder has scaffold out a basic reconciler struct for us.
Pretty much every reconciler needs to log, and needs to be able to fetch
objects, so a logger and a client are added out of the box.
Additionally, some client helpers (such as setting owner references) require
the scheme of the objects you are working with, so it is also injected.
*/
```

> So, let's keep as it is and then, if you would like please feel free to do a new PR after that where we can check better and discuss this specific point. I hope that it is fine for you.

If you really don't want to do this changes go ahead with this PR, but I don't think that the documentation of a change you are doing should be left for a separate PR as it will make it harder to find when looking at the commit history.

camilamacedo86:
HI @Adirio, 

Reafac the tutorial and or/improve its text is not the goal here. It is just a really tiny PR to add the current empty controller in the tutorial and the schema in the plugins since it is missing. 

It is great that you would like to improve the docs, however, let's do it in another PR.

camilamacedo86:
Hi @mengqiy, 

I addressed your comment. Let's merge this one? It is a really tiny fix/updated.

mengqiy:
Thanks!
/lgtm
/approve

DirectXMan12:
@Adirio  is right here -- we should've updated the documentation with this change.  When possible, all changes should be followed up with an equivalent docs change.  In this case, since we no longer need to manually add `Scheme` in, we should tweak the wording in the appropriate places.

DirectXMan12:
(better to do it in one PR when possible so as to avoid forgetting it)

Can someone follow up with the changes @Adirio suggested?

Issue #1153, opened 2019-11-05
Empty argument for create api command prints wrong error


Adirio:
As reported by @camilamacedo86 in https://github.com/kubernetes-sigs/kubebuilder/pull/1146#pullrequestreview-311425585, when issuing a `dockerbuilder create api ...` command with an empty value for a command line argument that is not at the end, the error reported does not describe the error properly.

<details><summary><code>$ kubebuilder create api --group webapp --version  --kind Guestbook</code></summary><pre>Create Resource [y/n]
y
Create Controller [y/n]
y
2019/11/05 08:31:42 missing kind information for resource</pre></details>

<details><summary><code>$ kubebuilder create api --group webapp --kind  --version Guestbook</code></summary><pre>Create Resource [y/n]
y
Create Controller [y/n]
y
2019/11/05 08:44:55 missing version information for resource</pre></details>

/kind bug

DirectXMan12:
interesting

/help

Adirio:
I think this may have to do with https://github.com/spf13/cobra

debojitkakoti:
@Adirio So, just need to change the command description for users to use '=' in command or pre-checking arguments ? What is your suggestion? Because it seems like intended behavior in cobra.

camilamacedo86:
/assing @camilamacedo86 

The issue occurs because when we pass --group --version it will set the "--version" as value to the group. I am doing a PR to fix it.

PR #1163, opened 2019-11-08
add required flags validation for create api command

camilamacedo86:
**Description**
-  Fix the scenarios when the group, version, and or kind can be empty. 

**Motivation**
Closes: #1153 

**Solution**
Screenshot 2019-11-13 at 12 06 31

### DIFF
--- a/cmd/api.go
+++ b/cmd/api.go
@@ -89,6 +89,10 @@ func (o *apiOptions) runAddAPI() {
 		log.Fatalf("unknown pattern %q", o.pattern)
 	}
 
+	if err := o.apiScaffolder.Validate(); err != nil {
+		log.Fatalln(err)
+	}
+
 	reader := bufio.NewReader(os.Stdin)
 	if !o.resourceFlag.Changed {
 		fmt.Println("Create Resource [y/n]")
@@ -98,10 +102,6 @@ func (o *apiOptions) runAddAPI() {
 	if !o.controllerFlag.Changed {
 		fmt.Println("Create Controller [y/n]")
 		o.apiScaffolder.DoController = util.Yesno(reader)
-	}
-
-	if err := o.apiScaffolder.Validate(); err != nil {
-		log.Fatalln(err)
 	}
 
 	fmt.Println("Writing scaffold for you to edit...")
--- a/pkg/scaffold/resource/resource.go
+++ b/pkg/scaffold/resource/resource.go
@@ -54,37 +54,57 @@ type Resource struct {
 
 // Validate checks the Resource values to make sure they are valid.
 func (r *Resource) Validate() error {
-	if len(r.Group) == 0 {
+	if r.isGroupEmpty() {
 		return fmt.Errorf("group cannot be empty")
 	}
-	if len(r.Version) == 0 {
+	if r.isVersionEmpty() {
 		return fmt.Errorf("version cannot be empty")
 	}
-	if len(r.Kind) == 0 {
+	if r.isKindEmpty() {
 		return fmt.Errorf("kind cannot be empty")
 	}
-
-	if len(r.Resource) == 0 {
-		r.Resource = flect.Pluralize(strings.ToLower(r.Kind))
-	}
-
+	// Check if the Group has a valid value for for it
 	if err := IsDNS1123Subdomain(r.Group); err != nil {
 		return fmt.Errorf("group name is invalid: (%v)", err)
 	}
-
-	r.GroupImportSafe = strings.Replace(r.Group, "-", "", -1)
-	r.GroupImportSafe = strings.Replace(r.GroupImportSafe, ".", "", -1)
-
+	// Check if the version is a valid value
 	versionMatch := regexp.MustCompile("^v\\d+(alpha\\d+|beta\\d+)?$")
 	if !versionMatch.MatchString(r.Version) {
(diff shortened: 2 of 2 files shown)


pkg/scaffold/v1/resource/resource.go
```diff
@@ -54,13 +54,13 @@ type Resource struct {
 
 // Validate checks the Resource values to make sure they are valid.
 func (r *Resource) Validate() error {
-	if len(r.Group) == 0 {
+	if len(r.Group) == 0 || r.Group == "--version" || r.Group == "--kind" {
```
mengqiy: This makes the code ugly.
Can we leverage https://godoc.org/github.com/spf13/cobra#Command.MarkFlagRequired?

camilamacedo86: Sure .. very better, tks for point it.

Adirio:
Are you sure that marking them as required solves it? I read in the Cobra issue (linked in the issue that is solved by this PR) that it was not solved by required. Didn't investigate ir further though.

camilamacedo86:
HI @Adirio, 

Yes. I tested it and it is solved. You can pull this PR in order to test it locally.

Adirio:
I do not have a computer right now available, I just remembered having read something about that.

cmd/api.go
```diff
@@ -161,6 +161,9 @@ After the scaffold is written, api will run make on the project.
 	make run
 `,
 		Run: func(cmd *cobra.Command, args []string) {
+			cmd.MarkFlagRequired("group")
```
mengqiy: I have tried locally, the way you wrote it doesn't seem to work.
I tweak it a bit. I found the following way works. Please give it a try.
It can give an error like `Error: required flag(s) "kind" not set`.
```
func newAPICommand() *cobra.Command {
	options := apiOptions{
		apiScaffolder: scaffold.API{},
	}

	apiCmd := &cobra.Command{
		Use:   "api",
		Short: "Scaffold a Kubernetes API",
		Long: `blah`,
		Example: `blah`,
		Run: func(cmd *cobra.Command, args []string) {
			options.runAddAPI()
		},
	}

	options.bindCmdFlags(apiCmd)
	apiCmd.MarkFlagRequired("group")
	apiCmd.MarkFlagRequired("version")
	apiCmd.MarkFlagRequired("kind")

	return apiCmd
}
```

camilamacedo86: Hi @mengqiy,
 
It will just work if we pass the 3 args as follows. 

```go
 $ kubebuilder create api --group kubebuilder --version v1alpha1 --kind Foobar
Create Resource [y/n]
y
Create Controller [y/n]
y
api/v1alpha1/foobar_types.go
controllers/foobar_controller.go
Running make...
/Users/camilamacedo/go/bin/controller-gen object:headerFile=./hack/boilerplate.go.txt paths="./..."
go fmt ./...
go vet ./...
camilamacedo@Camilas-MacBook-Pro ~/go/src/example $ 
```

```go
See that the error is just faced if we do not inform value for the Kind. 

 ~/go/src/example $ kubebuilder create api --group kubebuilder --version v1alpha1 --kind 
Error: flag needs an argument: --kind
Usage:
```

camilamacedo86: How you face the error? Are you passing the Kind?

mengqiy: To fix https://github.com/kubernetes-sigs/kubebuilder/issues/1153, I believe you need to mark flag as required outside of `Run` method.
 The code probably should look like what I proposed in https://github.com/kubernetes-sigs/kubebuilder/pull/1163/files#r344915843

camilamacedo86: Hi @mengqiy, 

Your propose is not solving the issue and keeping the previous behaviour. 
See:
Screenshot 2019-11-13 at 01 34 28

If we do as you suggested then, the tag `--version` will be set as the value of `--group` which is exactly the issue hat this PR solves. 

NOTE: Also, I could reproduce the scenario where it still not solved here `kubebuilder create api --group --version x --kind D` too.

So, I solved it properly in another way. Let's me know what do you think?

pkg/scaffold/resource/resource.go
```diff
@@ -54,38 +54,58 @@ type Resource struct {
...
 	}
 
+	// todo: move it for the proper place since they are not validations and then, should not be here
+	// Add in r.Resource the Kind plural
+	if len(r.Resource) == 0 {
+		r.Resource = flect.Pluralize(strings.ToLower(r.Kind))
+	}
+	// Replace the caracter "-" for "" to allow scaffold the go imports
+	r.GroupImportSafe = strings.Replace(r.Group, "-", "", -1)
 	return nil
```
camilamacedo86: @mengqiy just note that I move the impl to a better order as ; check if is empty, then check if are valid values.
Also, add a TODO for we refactory it after in another PR since here should have just validations and not sets for the auxiliar values.

Adirio: I completely agree with you in this point, the `Validate` method is doing too many things. I was thinking on using constructor-oriented design (use a function called `New` or `NewXxxx` to create the instances of each type instead of literal struct constructs) and this not only applies here, but in a few more places.

camilamacedo86:
Hi @mengqiy, 

The /lgtm was lost with the master rebase. Could you please add again?

PR #1164, opened 2019-11-08
Handle creation of duplicate API resources

### LINKED ISSUE #1110: Detect duplicate api creation commands
Executing two times the same `kubebuilder create api ...` with the same group, version and kind did not recognize that it was already in the project. I'm not sure if the changes done to some of the generated files will stay after the second execution of the command, as it was a fresh start. But what I do know is that 3 files got duplicated code:

- `PROJECT`: `resources` lists the GVK twice
- `main.go`: `SetupWithManager` called twice
- `controllers/suite_test.go`: `AddToScheme` checked twice

I expected either an error message telling me that that GVK already exists or a warning that asks me if I want to reset that GVK to the default.

To reproduce:
```
kubebuilder init --my.domain
kubebuilder create api --group mygroup --version v1alpha1 --kind Example
kubebuilder create api --group mygroup --version v1alpha1 --kind Example
```

Versions:
- Golang: `go version go1.12.3`
- kubebuilder: `Version: version.Version{KubeBuilderVersion:"unknown", KubernetesVendor:"unknown", GitCommit:"$Format:%H$", BuildDate:"1970-01-01T00:00:00Z", GoOs:"unknown", GoArch:"unknown"}`
- Scaffolding: `version: "2"`
- controller-runtime: `v0.2.2`
- controller-tools: `v0.2.1`
- Kubernetes: `v1.14.6`
- kubectl: `v1.14.6`

/kind bug

hypnoglow:
This changeset updates `create api` command logic, rejecting duplicate APIs by default, but allowing to continue with `--force` flag.

Note that it only affects v2 projects, since in v1 projects resources are not tracked by the PROJECT file.

Fixes #1110

### DIFF
--- a/cmd/api.go
+++ b/cmd/api.go
@@ -57,6 +57,8 @@ func (o *apiOptions) bindCmdFlags(cmd *cobra.Command) {
 		cmd.Flags().StringVar(&o.pattern, "pattern", "",
 			"generates an API following an extension pattern (addon)")
 	}
+	cmd.Flags().BoolVar(&o.apiScaffolder.Force, "force", false,
+		"attempt to create resource even if it already exists")
 	o.apiScaffolder.Resource = resourceForFlags(cmd.Flags())
 }
 
--- a/pkg/scaffold/api.go
+++ b/pkg/scaffold/api.go
@@ -49,6 +49,9 @@ type API struct {
 
 	// DoController indicates whether to scaffold controller files or not
 	DoController bool
+
+	// Force indicates that the resource should be created even if it already exists.
+	Force bool
 }
 
 // Validate validates whether API scaffold has correct bits to generate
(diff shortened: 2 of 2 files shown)


Adirio:
I think Travis fail is not related to this, but to #1169. Try to /retest

DirectXMan12:
looks good.
Do you still consider it WIP?  It looks done ish.

hypnoglow:
@DirectXMan12 I guess this is not WIP anymore. 

I thought about also handling code duplicates in `main.go` and `controllers/suite_test.go` (as described in issue #1110) so that running with `--force` would not result in that duplication. But I think this will complicate things, so maybe it is better to leave this off to the end-user so they can see a diff and remove unnecessary code themselves. 

Anyway, as I understood, creating the same API twice, is rarely useful, and most users probably attempting to create already existing APIs by mistake and not intentionally. We want to prevent these mistakes and that's what this PR already does.

DirectXMan12:
sounds good.

I'll merge this then :-)

Issue #886, opened 2019-07-24
Generate suggested Roles for CR users


erictune:
/kind feature

Today, KB use some static roles and it generates a dynamic Role from the `//+kubebuilder:rbac:...` comments.  These cover the permissions the controller needs.

These are _not_ the ones this feature request is talking about.

KB should also generate ClusterRoles to be used by end users who want to CRUD CRs defined by the CRDs in the KB project.

Say my KB project defines something called `admirals`.
KB should defined a ClusterRole called `admiral-editor` and `admiral-viewer`.
These should be something the consumer of the operator is recommended to install.  

The consumer of the operator (usually different from the KB user) would be responsible for making any bindings between their users and these Roles.

mengqiy:
Reopen this since we may want to revise it per https://github.com/kubernetes-sigs/kubebuilder/pull/1165#discussion_r347689204

PR #1165, opened 2019-11-08
feat(rbac) : add editor and viewer role for crds

camilamacedo86:
**Description**
- Add editor and viewer roles for each CRD

**Motivation**
Closes: #886

### DIFF
--- a/pkg/scaffold/v2/crd_editor_rbac.go
+++ b/pkg/scaffold/v2/crd_editor_rbac.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v2
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/input"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/resource"
+)
+
+var _ input.File = &CRDEditorRole{}
+
+// CRD Editor role scaffolds the config/rbca/<kind>_editor_role.yaml
+type CRDEditorRole struct {
+	input.Input
+
+	// Resource is a resource in the API group
+	Resource *resource.Resource
+}
+
+// GetInput implements input.File
+func (g *CRDEditorRole) GetInput() (input.Input, error) {
+	if g.Path == "" {
+		g.Path = filepath.Join("config", "rbac", fmt.Sprintf("%s_editor_role.yaml", strings.ToLower(g.Resource.Kind)))
+	}
+
+	g.TemplateBody = crdRoleEditorTemplate
+	return g.Input, nil
+}
+
+// Validate validates the values
+func (g *CRDEditorRole) Validate() error {
+	return g.Resource.Validate()
+}
+
+const crdRoleEditorTemplate = `# permissions to do edit {{ .Resource.Resource }}.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ lower .Resource.Kind }}-editor-role
+rules:
+- apiGroups:
+  - {{ .Resource.Group }}.{{ .Domain }}
+  resources:
+  - {{ .Resource.Resource }}
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - {{ .Resource.Group }}.{{ .Domain }}
+  resources:
+  - {{ .Resource.Resource }}/status
+  verbs:
+  - get
+  - patch
+  - update
+`
--- a/pkg/scaffold/v2/crd_viewer_rbac.go
+++ b/pkg/scaffold/v2/crd_viewer_rbac.go
@@ -1,0 +1,73 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v2
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/input"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/resource"
+)
+
+var _ input.File = &CRDViewerRole{}
+
+// CRD Viewer role scaffolds the config/rbca/<kind>_viewer_role.yaml
+type CRDViewerRole struct {
+	input.Input
+
+	// Resource is a resource in the API group
+	Resource *resource.Resource
+}
+
+// GetInput implements input.File
+func (g *CRDViewerRole) GetInput() (input.Input, error) {
+	if g.Path == "" {
+		g.Path = filepath.Join("config", "rbac", fmt.Sprintf("%s_viewer_role.yaml", strings.ToLower(g.Resource.Kind)))
+	}
+
+	g.TemplateBody = crdRoleViewerTemplate
+	return g.Input, nil
+}
+
+// Validate validates the values
+func (g *CRDViewerRole) Validate() error {
+	return g.Resource.Validate()
+}
+
+const crdRoleViewerTemplate = `# permissions to do viewer {{ .Resource.Resource }}.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ lower .Resource.Kind }}-viewer-role
+rules:
+- apiGroups:
+  - {{ .Resource.Group }}.{{ .Domain }}
+  resources:
+  - {{ .Resource.Resource }}
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - {{ .Resource.Group }}.{{ .Domain }}
+  resources:
+  - {{ .Resource.Resource }}/status
+  verbs:
+  - get
+`
--- a/testdata/project-v2/config/rbac/admiral_viewer_role.yaml
+++ b/testdata/project-v2/config/rbac/admiral_viewer_role.yaml
@@ -1,0 +1,20 @@
+# permissions to do viewer admirals.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: admiral-viewer-role
+rules:
+- apiGroups:
+  - crew.testproject.org
+  resources:
+  - admirals
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - crew.testproject.org
+  resources:
+  - admirals/status
+  verbs:
+  - get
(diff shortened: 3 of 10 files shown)


pkg/scaffold/v2/crd_editor_rbac.go
```diff
@@ -0,0 +1,71 @@
...
+kind: Role
+metadata:
+  name: {{ lower .Resource.Kind }}-editor-role
+rules:
+- apiGroups:
+  - apps
+  resourceNames:
+  - {{ .Resource.Kind }}
+  resources:
+  - deployments/finalizers
```
mengqiy: We don't need `resourceNames`
It should be something like
```
  resources:
  - foos
  - foos/status
```
You can use `github.com/gobuffalo/flect` to get the plural. You can find some examples in the existing codebase.

pkg/scaffold/v2/crd_viewer_rbac.go
```diff
@@ -0,0 +1,68 @@
...
+
+var crdRoleViewerTemplate = `# permissions to do viewer {{ .Resource.Kind }}.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ lower .Resource.Kind }}-viewer-role
+rules:
+- apiGroups:
+  - apps
+  resourceNames:
```
mengqiy: same comment here

pkg/scaffold/v2/crd_viewer_rbac.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+var crdRoleViewerTemplate = `# permissions to do viewer {{ .Resource.Kind }}.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ lower .Resource.Kind }}-viewer-role
+rules:
+- apiGroups:
+  - apps
```
mengqiy: same

pkg/scaffold/v2/crd_editor_rbac.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+var crdRoleEditorTemplate = `# permissions to do edit {{ .Resource.Kind }}.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ lower .Resource.Kind }}-editor-role
+rules:
+- apiGroups:
+  - apps
```
mengqiy: group name is wrong.

camilamacedo86:
Hi @mengqiy, 

Really thank you. I was looking for exactly this kind of review. Working on to address it.

testdata/project-v2/config/rbac/admiral_viewer_role.yaml
```diff
@@ -0,0 +1,27 @@
+# permissions to do edit Admiral.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: admiral-editor-role
```
mengqiy: Content in all of the viewer role files seem to be duplicates of editor roles.
Please ensure generate_golden.sh has been run.

pkg/scaffold/v2/crd_editor_rbac.go
```diff
@@ -0,0 +1,84 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v2
+
+import (
+	"path/filepath"
```
mengqiy: nit: we usually keep imports in different groups:
```
import (
  std libs

  non-k8s libs

  k8s libs
)

File: pkg/scaffold/v2/crd_editor_rbac.go [hunk 7]
```diff
@@ -0,0 +1,84 @@
...
+// Validate validates the values
+func (g *CRDEditorRole) Validate() error {
+	return g.Resource.Validate()
+}
+
+var crdRoleEditorTemplate = `# permissions to do edit {{ .Resource.Kind }}.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
```
mengqiy: We don't need this.

pkg/scaffold/v2/crd_editor_rbac.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/gobuffalo/flect"
+)
+
+var _ input.File = &Group{}
+
+// CRD Editor role scaffolds the config/rbca/<kind>_editor_role.yaml
+type CRDEditorRole struct {
+	input.Input
+	// Plural is the plural lowercase of kind
+	Plural string
```
mengqiy: It seems you don't need to do it yourself. Using https://github.com/kubernetes-sigs/kubebuilder/blob/eb3d7967cae29ba8bcf6893f359621551523bdeb/pkg/scaffold/v1/resource/resource.go#L46 should be fine. And we already populate it in https://github.com/kubernetes-sigs/kubebuilder/blob/eb3d7967cae29ba8bcf6893f359621551523bdeb/pkg/scaffold/v1/resource/resource.go#L67-L69

pkg/scaffold/v2/crd_viewer_rbac.go
```diff
@@ -0,0 +1,78 @@
...
+	"strings"
+)
+
+var _ input.File = &CRDViewerRole{}
+
+// CRD Viewer role scaffolds the config/rbca/<kind>_viewer_role.yaml
+type CRDViewerRole struct {
+	input.Input
+	// Plural is the plural lowercase of kind
+	Plural string
```
mengqiy: Same here

pkg/scaffold/v2/crd_editor_rbac.go
camilamacedo86: Are you suggesting not declare the in the struct use to create the template? 

However, note that the plural and resource are required. See that it will be undefined if we remove it since the struct use has no Plural attribute. 

```
2019/11/11 22:45:48 error scaffolding APIs: template: *v2.CRDEditorRole:10:7: executing "*v2.CRDEditorRole" at <.Plural>: can't evaluate field Plural in type *v2.CRDEditorRole
make[1]: *** [generate-testdata] Error 1
make: *** [generate] Error 2

```

Also, see that the other scaffolds are also declaring them in the same way

https://github.com/kubernetes-sigs/kubebuilder/blob/100b673ad608e54cab6e5d2aaca6d98301b4ca22/pkg/scaffold/v2/crd/kustomization.go#L63-L64

https://github.com/kubernetes-sigs/kubebuilder/blob/cd0b10ff1a7966ce3d449ae9fbb8d7029259bc31/pkg/scaffold/v2/crd/enablewebhook_patch.go#L35

pkg/scaffold/v2/crd_viewer_rbac.go
camilamacedo86: ˆ

pkg/scaffold/v2/crd_editor_rbac.go
camilamacedo86: Noted for the future PR's as well 👍

pkg/scaffold/v2/crd_editor_rbac.go
mengqiy: {{ .Resource.Resource }} should be the plural of the resources. e.g. https://github.com/kubernetes-sigs/kubebuilder/blob/cd0b10ff1a7966ce3d449ae9fbb8d7029259bc31/pkg/scaffold/v2/crd/enablewebhook_patch.go#L59

Viewer role should look like:
```
type CRDViewerRole struct {
	input.Input

	// Resource is a resource in the API group
	Resource *resource.Resource
}

func (g *CRDViewerRole) GetInput() (input.Input, error) {
	if g.Path == "" {
		g.Path = filepath.Join("config", "rbac", fmt.Sprintf("%s_viewer_role.yaml", strings.ToLower(g.Resource.Kind)))
	}
	g.TemplateBody = crdRoleViewerTemplate
	return g.Input, nil
}

// Validate validates the values
func (g *CRDViewerRole) Validate() error {
	return g.Resource.Validate()
}

var crdRoleViewerTemplate = `
kind: ClusterRole
metadata:
  name: {{ lower .Resource.Kind }}-viewer-role
rules:
- apiGroups:
  - {{ .Resource.Group }}.{{ .Domain }}
  resources:
  - {{ .Resource.Resource }}  # <==== this should the plural of your resources
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - {{ .Resource.Group }}.{{ .Domain }}
  resources:
  - {{ .Resource.Resource }}/status
  verbs:
  - get
`
```

pkg/scaffold/v2/crd_editor_rbac.go
```diff
@@ -0,0 +1,85 @@
...
+	g.TemplateBody = crdRoleEditorTemplate
+	return g.Input, nil
+}
+
+// Validate validates the values
+func (g *CRDEditorRole) Validate() error {
+	return g.Resource.Validate()
+}
+
+var crdRoleEditorTemplate = `# permissions to do edit {{ .Resource.Kind }}.
```
mengqiy: > permissions to do edit {{ .Resource.Kind }}.

permissions to edit {{ .Resource.Resource }}.

pkg/scaffold/v2/crd_viewer_rbac.go
```diff
@@ -0,0 +1,79 @@
...
+	g.TemplateBody = crdRoleViewerTemplate
+	return g.Input, nil
+}
+
+// Validate validates the values
+func (g *CRDViewerRole) Validate() error {
+	return g.Resource.Validate()
+}
+
+var crdRoleViewerTemplate = `# permissions to do viewer {{ .Resource.Kind }}.
```
mengqiy: > permissions to do viewer {{ .Resource.Kind }}.

permissions to view {{ .Resource.Resource }}.

pkg/scaffold/v2/crd_editor_rbac.go
camilamacedo86: Catcher 👍 Thank you. doing it now.

pkg/scaffold/v2/crd_viewer_rbac.go
camilamacedo86: Done 👍

pkg/scaffold/v2/crd_editor_rbac.go
camilamacedo86: Done 👍

pkg/scaffold/v2/crd_viewer_rbac.go
camilamacedo86: Done 👍

pkg/scaffold/v2/crd_editor_rbac.go
camilamacedo86: Done 👍

mengqiy:
Travis is failing. Please fix.

camilamacedo86:
Hi @mengqiy, 

Now, the test is fixed and all is passing. Please, feel free to check this one.

testdata/project-v2/config/rbac/firstmate_editor_role.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - crew.testproject.org
+  resources:
+  - firstmates/status
+  verbs:
```
erictune: I was comparing this with the build-in cluster role `edit` which is  you can see with `kubectl get clusterroles edit -o yaml`.  The edit role gives read and write to most of the built-in kubernetes resources.  I noted that it does not give patch or update permissions on any `*/status` resources.  Is it by design that these generated roles are different?

mengqiy: If `*/status` should be included depends on who will be granted these permission.
For controllers, permissions to edit `*/status` is necessary.
For human, they should never edit `*/status`.

Reading #886 again, it seems that means human users.
I guess we can still revise it and make it clear that it is for end users.

camilamacedo86: > I was comparing this with the build-in cluster role edit which is you can see with kubectl get clusterroles edit -o yaml.

The roles done here are to give permissions to the specific resource. I am not sure if I could understand properly the point.

> I noted that it does not give patch or update permissions on any */status resources.

I understand that the idea is to allow admin using this role to create a group of permissions. So, the status is required for the operator to update the CR's and users would also would like to check the status. Am I right? 

I am not sure if I understood what should be done in another way here. May the question could be what behaviour are you face or should not be faced when these roles are applied?

PR #1167, opened 2019-11-09
Use sigs.k8s.io/yaml for parsing.

### LINKED ISSUE #1166: Use sigs.k8s.io/yaml over gopkg.in/yaml.v2
Kubebuilder currently uses the `gopkg.in/yaml.v2` libraries for parsing `ProjectFiles`, since there is already a `sigs.k8s.io/yaml` which is a fork of `yaml.v2` potentially we can reduce the use of outside packages by using that instead.

/kind feature

christopherhein:
This implements the `yaml` parsing using the `sigs.k8s.io/yaml` parser instead of the `gopkg.in/yaml.v2` libraries.

This was motivated by trying to keep all dependencies within sig/k8s owned projects.

closes #1166

### DIFF
--- a/testdata/project-v2/PROJECT
+++ b/testdata/project-v2/PROJECT
@@ -1,13 +1,13 @@
-version: "2"
 domain: testproject.org
 repo: sigs.k8s.io/kubebuilder/testdata/project-v2
 resources:
 - group: crew
-  version: v1
   kind: Captain
-- group: crew
   version: v1
+- group: crew
   kind: FirstMate
-- group: crew
   version: v1
+- group: crew
   kind: Admiral
+  version: v1
+version: "2"
--- a/go.mod
+++ b/go.mod
@@ -16,5 +16,5 @@ require (
 	golang.org/x/text v0.3.2 // indirect
 	golang.org/x/tools v0.0.0-20190621195816-6e04913cbbac
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
-	gopkg.in/yaml.v2 v2.2.2
+	sigs.k8s.io/yaml v1.1.0
 )
--- a/go.sum
+++ b/go.sum
@@ -89,3 +89,5 @@ gopkg.in/yaml.v2 v2.2.1 h1:mUhvW9EsL+naU5Q3cakzfE91YhliOondGd6ZrsDBHQE=
 gopkg.in/yaml.v2 v2.2.1/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 gopkg.in/yaml.v2 v2.2.2 h1:ZCJp+EgiOT7lHqUV2J862kp8Qj64Jo6az82+3Td9dZw=
 gopkg.in/yaml.v2 v2.2.2/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
+sigs.k8s.io/yaml v1.1.0 h1:4A07+ZFc2wgJwo8YNlQpr1rVlgUDlxXHhPJciaPY5gs=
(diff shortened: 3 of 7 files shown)


camilamacedo86:
Hi @christopherhein, 

You need to update the testdata. Following the commands, could you please run them and do a new commit? 

```sh
	make generate-setup
	make generate-testdata
	go mod tidy
```

christopherhein:
Thanks for the review @camilamacedo86 ! 

I've rebased, and ran `generate-setup`, `generate-testdata` and `tidy`.

christopherhein:
@camilamacedo86 is something going up with the test suite right now? Looks like all builds are failing.

camilamacedo86:
Hi @christopherhein,

Shows that yes. I did the https://github.com/kubernetes-sigs/kubebuilder/pull/1168 to check it. Then, we need to fix the CI :-( for we are able to/ before to do any merge. If you wish, please feel free to help I will try to check it tomorrow if has no solution so far.

christopherhein:
@camilamacedo86 tests now pass :+1:

testdata/project-v2/PROJECT
```diff
@@ -1,13 +1,13 @@
...
-- group: crew
   version: v1
-  kind: FirstMate
 - group: crew
+  kind: FirstMate
   version: v1
+- group: crew
   kind: Admiral
+  version: v1
+version: "2"
```
Adirio: Can we control the order somehow? It is using alphabetical ordering, and while this may not be improtant for resources, having `version: "X"` before the resources list seems pretty convenient for both users and scaffolding extra resources as we will just be appending to the file instead of inserting in the middle.

camilamacedo86: It is not inserted in the middle and it is generated automatically. In this way, any change regards to it will not be part of the scope of this PR.

Adirio: I know this file is generated automatically and that the scaffold is able to insert in the middle. I just asked if the current behavior can be achieved as this may lead to future optimizations (appending to a file requires less time from the underlying os than creating it from scratch most of the times). If it can not be achieved, I think it will be ok, but if it can be achieved it may be interesting to keep the order as is.

camilamacedo86: Hi @Adirio

Thank you for your input.  However, IHMO too many comments or too much verbose can make harder the pr get reviewed and merged. So, any subject which is not revalent to the pr scope should not be addressed here. 

But note that you are able always to address all your ideas/suggestions by raising new issues which I think would be more appropriate in this kind of scenario. WDYT?

Adirio: I do not agree that this comment is out of scope of this PR. This PR is changing the YAML parser, which results in a different field order. As this PR is changing current behavior (version is at the top and resources at the end) asking if the order can be kept the same its IMHO in the scope of this project.
About the length of comments, thats what the "mark conversation as resolved" is for. Once the author solves the comment (in this case that means either keeping the current order or saying it cannot be done but the owners of the repo still find it ok) this comment can be marked as solved and it won't take any space.
Offtopi: what it is adding noise is all this conversation about the ammount of comments

christopherhein: Thanks for the comment, I don't believe this is possible at least not to reflect it's original output, albeit it seems like it might be ordering in alphabetical order which is at least interesting. Resolving.

christopherhein:
Just rebased and waiting on Travis, thanks @camilamacedo86 for getting #1168 merged.

christopherhein:
Looks like everything's passed now @camilamacedo86

camilamacedo86:
I show /lgtm /approved for me. 
WDYT? Could we merge this one?
/assign @mengqiy 
/assign @droot
/assign @Liujingfang1

testdata/project-v2/PROJECT
mengqiy: I'm OK as long as the order is deterministic and stable.

PR #1135, opened 2019-10-28
Upgrade project to use go 1.13 

camilamacedo86:
**Description**
Upgrade the project to work/used 1.13 go and its tests

**Motivation**
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/1134

PR #1168, opened 2019-11-09
Fix CI

camilamacedo86:
**Description**
Fix CI - By updating go version in the images

**Motivation**
Closes: #1169

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,8 +7,8 @@ Kubernetes projects require that you sign a Contributor License Agreement (CLA) 
 Please see https://git.k8s.io/community/CLA.md for more info.
 
 ## Prerequisites
-- [go](https://golang.org/dl/) version v1.12+.
-- [dep](https://github.com/golang/dep) dep v0.5.1+ < v.0.5.4
+- [go](https://golang.org/dl/) version v1.13+.
+- [dep](https://github.com/golang/dep) dep v0.5+
 - [docker](https://docs.docker.com/install/) version 17.03+.
 - [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) version v1.11.3+.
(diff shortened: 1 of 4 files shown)


Adirio:
Small nit: I think you wanted to say `Closes: #1169` not `Closes: #1168`.

ncdc:
I would not recommending upgrading to go 1.13 until kubebuilder's dependencies have all upgraded to the versions of apimachinery, client-go, etc. that are built with go 1.13, which will be those coming from Kubernetes v1.17.

/hold

Adirio:
The point is that 1.13 is already being used by an already merged PR, if I'm not mistaken.

ncdc:
I see that #1135 upgraded kubebuilder to use go1.13. @DirectXMan12 don't you want to wait until the staging repos have been updated to use go1.13 as well (which means waiting for kube v1.17)?

ncdc:
I'll remove my hold since I'm not an active kubebuilder contributor, but I would have thought this repo would have waited to upgrade to go1.13 until after a release of Kubernetes came out that is built with go1.13.

/hold cancel

DirectXMan12:
> but I would have thought this repo would have waited to upgrade to go1.13 until after a release of Kubernetes came out that is built with go1.13

We were mainly holding till we could be sure that we wouldn't cause people too much of a headache with their dev environments -- code wise, things seem to work fine.  Kubernetes tends to move pretty slowly with these things because of its complicated build scripts & the way gengo works, and I'm not sure we also want to couple ourselves to that cadence.

If there are any active issues, please do mention them before the release.

@ncdc and @camilamacedo86 have already made me aware of https://github.com/kubernetes/kubernetes/issues/85103  already (the module solver fails w/o goproxy when people name branches after commit shas, which seems like a Bad Idea™ anyway)

Issue #1138, opened 2019-10-29
Error in the scaffold files when the group name has '.'


ctiml:
Hello, I'm trying to create apis using kubebuilder v2.1.0 and I got this error message:

```
$ kubebuilder create api --group example.com --version v1alpha1 --kind Foobar

package controllers

import (
        "context"

        ctrl "sigs.k8s.io/controller-runtime"
        "sigs.k8s.io/controller-runtime/pkg/client"
        "github.com/go-logr/logr"
        "k8s.io/apimachinery/pkg/runtime"

        example.comv1alpha1 "test/api/v1alpha1"
)

// FoobarReconciler reconciles a Foobar object
type FoobarReconciler struct {
        client.Client
        Log logr.Logger
        Scheme *runtime.Scheme
}

// +kubebuilder:rbac:groups=example.com.,resources=foobars,verbs=get;list;watch;create;update;patch;delete
// +kubebuilder:rbac:groups=example.com.,resources=foobars/status,verbs=get;update;patch

func (r *FoobarReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
        _ = context.Background()
        _ = r.Log.WithValues("foobar", req.NamespacedName)

        // your logic here

        return ctrl.Result{}, nil
}

func (r *FoobarReconciler) SetupWithManager(mgr ctrl.Manager) error {
        return ctrl.NewControllerManagedBy(mgr).
                For(&example.comv1alpha1.Foobar{}).
                Complete(r)
}

2019/10/29 14:52:52 error scaffolding controller: controllers/foobar_controller.go:26:9: expected 'STRING', found '.' (and 4 more errors)
```

`example.comv1alpha1 "test/api/v1alpha1"` is not valid import alias.

I think this is a follow-up issue of https://github.com/kubernetes-sigs/kubebuilder/pull/1049. It should be replaced when group name has '.' (dot).

go version: go1.13.3 darwin/amd64
kubebuilder version: v2.1.0

/kind bug

droot:
@ctiml This has not been an issue for most Kubebuilder users. Can you share what do you want the API Group for the CRD to look like ? 

For ex.  if the API group we want is `apps.example.com` with Foobar kind in it, we will begin the Kubebuilder project with domain `example.com` by running `kubebuilder init --domain example.com`. And then we add new API by running `kubebuilder create api --group apps --kind Foobar --version v1alpha1`.

Hope this helps.

[...]

PR #1172, opened 2019-11-10
solve issue with groups with '.' as example.com

camilamacedo86:
**Description**
- solve the scaffolding issue with groups with `.` as example.com by doing the same fix that is done with `-`

**Motivation**
Closes: #1138


Adirio:
LGTM

PR #1173, opened 2019-11-10
Add init name check

camilamacedo86:
**Description**
- add validation for the name of the project in the init

**Motivation**
Closes: #891

### DIFF
--- a/cmd/init_project.go
+++ b/cmd/init_project.go
@@ -21,6 +21,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"path/filepath"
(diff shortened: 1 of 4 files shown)


cmd/init_project.go
```diff
@@ -137,6 +138,20 @@ func (o *projectOptions) validate() error {
 			return err
 		}
 	}
+
+	// use directory name as prefix
+	dir, err := os.Getwd()
```
DirectXMan12: we'll want to enventually divorce this from directory name, IMO.

PR #1190, opened 2019-11-11
Avoid rewriting template values, mark them as const

justinsb:
Fix a place where we rewrote a shared template value and potentially
made it unsafe for concurrent / repeated evaluation.  Also mark the
templates as const to have the compiler help us stick to it!


DirectXMan12:
> Just one small nit: the files where you have removed the extra blank spaces at the end, is this to normalize all of them? So that none of them have an extra blank line after it.

I think gofmt may do this automatically in some cases

Issue #923, opened 2019-08-08
Multi group support for v2 scaffolding


DirectXMan12:
We should have proper multi-group support in the v2 scaffolding.  We won't scaffold it by default, and migrating to it will always be a manual change, but once you've enabled it, we should be able to scaffold API versions in a per-group directory.

Presumably, this would look like adding a `multigroup: true` stanza to the `PROJECT` file, and the KubeBuilder would generate new groups at

`apis/<group>/<version>` instead of `api/<version>`.

/kind feature

paroxp:
I assume this won't be as part of the initial stable kubebuilder release? :(

DirectXMan12:
automated scaffolding for multi-group repos will most likely be part of v2.1.0.  That being said, we've gone to a lot of effort to ensure that the underlying tooling and libraries work with pretty much any directory layout, so you should still be able to use the corresponding controller-runtime and controller-tools v0.2.0 releases to support a multi-group setup.

DirectXMan12:
rewriting import paths like that is "scary" since it can contain unintended changes or miss things but still works.  That's why we wouldn't have automatic migration.  There's also no guarantee that things haven't been reworked, files haven't been added, etc since the user did the scaffolding.  We generally follow the principle that scaffolding is one-off, so we don't try to change it (with a couple of small exceptions in main.go).

If you're generating from scratch on a new project, sure, but you'd have to "migrate" your existing project by hand so that we don't mess it up, then we could easily start scaffolding in the new setup.

Issue #1196, opened 2019-11-14
Make install is not working with multiversions


camilamacedo86:
## Bug Report

**What did you do?**
Follow the quick started and create a v2 for the same Kind. Following the steps to reproduce the issue. 

```shell
$ mkdir $GOPATH/src/example
$ cd $GOPATH/src/example
$ kubebuilder init --domain my.domain
$ kubebuilder create api --group webapp --version v1 --kind Guestbook
$ kubebuilder create api --group webapp --version v2 --kind Guestbook
$ make install
```
**What did you expect to see?**
All worked with success. 

**What did you see instead? Under which circumstances?**

```shell
$ make install
/Users/camilamacedo/go/bin/controller-gen "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
test/api/v1:-: CRD for Guestbook.webapp.my.domain has no storage version
Error: not all generators ran successfully
run `controller-gen crd:trivialVersions=true rbac:roleName=manager-role webhook paths=./... output:crd:artifacts:config=config/crd/bases -w` to see all available markers, or `controller-gen crd:trivialVersions=true rbac:roleName=manager-role webhook paths=./... output:crd:artifacts:config=config/crd/bases -h` for usage
make: *** [manifests] Error 1
```

**Environment**
* kubebuilder version: (master branch)
* go version: 1.13.4
* Kubernetes version information:
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"clean", BuildDate:"2019-09-18T14:36:53Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:09:08Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```
* Kubernetes cluster kind: Minikube

/kind bug

camilamacedo86:
c/c @DirectXMan12 @mengqiy

DirectXMan12:
[...]

Issue #1197, opened 2019-11-14
Artefacts are not updated with the new version gen


camilamacedo86:
## Feature Request

**Is your feature request related to a problem? Please describe.**

Artefacts are not updated with the new version gen. Following the steps to reproduce the scenario

```shell
$ mkdir $GOPATH/src/example
$ cd $GOPATH/src/example
$ kubebuilder init --domain my.domain
$ kubebuilder create api --group webapp --version v1 --kind Guestbook
$ kubebuilder create api --group webapp --version v2 --kind Guestbook
```
<img width="1146" alt="Screenshot 2019-11-14 at 11 03 58" src="https://user-images.githubusercontent.com/7708031/68853815-b94dbd80-06d2-11ea-90c0-1b566045e19e.png">

**Describe the solution you'd like**
I'd like to see the Artefacts updated to use the v2 when we press `y` to create Controller [y/n]

/kind feature

DirectXMan12:
We generally don't update scaffolded files because we have no way to know if the user has significantly changed them.

PR #1198, opened 2019-11-14
:sparkles: Add multigroup support for v2

camilamacedo86:
**Description**
- Add support to allow users to create multigroup in v2
- Add new command to edit the PROJECT file and the multigroup flag
- Add new testdata project with multigroup layout
- Update the multigroup doc

**Motivation**
Closes: #923
Closes: #1202


pkg/scaffold/api.go
```diff
@@ -260,13 +271,30 @@ func (api *API) scaffoldV2() error {
...
+func (api *API) isNewGroup(r *resource.Resource) bool {
+	for _, existingGroup := range api.project.ResourceGroups() {
+		if strings.ToLower(r.Group) != strings.ToLower(existingGroup) {
+			return true
+		}
+	}
+	return false
+}
+
 // Since we support single group only in v2 scaffolding, validate if resource
```
christopherhein: Assume we want to remove this comment before this is merged?

pkg/scaffold/util/util.go
```diff
@@ -53,7 +53,7 @@ func GetResourceInfo(r *resource.Resource, repo, domain string) (resourcePackage
 		"setting":               "k8s.io",
 		"storage":               "k8s.io",
 	}
-	resourcePath := filepath.Join("api", r.Version, fmt.Sprintf("%s_types.go", strings.ToLower(r.Kind)))
+	resourcePath := filepath.Join("api", r.Group, r.Version, fmt.Sprintf("%s_types.go", strings.ToLower(r.Kind)))
```
christopherhein: Should it be `api` or `apis` when it's multigroup?

christopherhein:
Correct me if I'm wrong but it looks like this will be a breaking change for anyone using KB v2.0, right? since it moves the `<resource>_types.go` into a `group` named directory for all generators, instead of supporting the current path for non-multigroup projects.

christopherhein:
A separate question, should `controllers/<group>/<controller>.go` exist or should it still be a single place for all groups?

pkg/scaffold/util/util.go
camilamacedo86: It is generated in the apis dir. We can check the updates in the testdata dir to ensure that. 
However, you make a good point here. I will check why here is api and not apis.

camilamacedo86: By following the common practices/name conventions the nomenclature of the dirs should be always single if I am not wrong. However, any way I do not think we should change the name of the current dir in this PR.

pkg/scaffold/api.go
camilamacedo86: I will remove it.

pkg/scaffold/util/util.go
christopherhein: Make sense. Thanks.

camilamacedo86:
HI @christopherhein, 


> Correct me if I'm wrong but it looks like this will be a breaking change for anyone using KB v2.0, right? since it moves the <resource>_types.go into a group named directory for all generators, instead of supporting the current path for non-multigroup projects.

The projects created before the change will not stop to work and i cannot see any issue that could be faced on them when run the makefile targets. However, any new API created will be done in`/api/<grooup>/version` instead of `/api/version`. So, I agree that it is a breakage and we should recommend the users manually move the files for the new structure. 

> A separate question, should controllers/<group>/<controller>.go exist or should it still be a single place for all groups?
I think that could/should be done as well in order to start to support the same kind for diff groups. Currently and with this change as well the Kinds per project still need to be unique. However, it shows that is required to change the gen tools as well and not only here. We definitely should track this need too. I will do it, but, this one, would be a pre-requirement anyway. 

Also, IHMO before supporting the same kind for diff groups in the project we need to do https://github.com/kubernetes-sigs/kubebuilder/issues/1197 and this fix https://github.com/kubernetes-sigs/kubebuilder/issues/1196 as well.

mengqiy:
This is our original design doc
https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/simplified-scaffolding.md#revised-structure
The intent is to simplify the project structure.

I'd suggest you to have this well discussed before spending a lot of time on the code.

camilamacedo86:
HI @mengqiy, 

Really thank you for your input and for the doc provided. However, this one shows done exactly was define here:  #923 by @DirectXMan12. Also, I do not see another way of we support multi groups without adding the group in the path `/api/<grooup>/version ` as described there. 

I will try contact you folks to check further this one. really tks :-)

cmd/webhook_v2.go
```diff
@@ -71,7 +71,7 @@ func newWebhookV2Cmd() *cobra.Command {
 			}
 
 			fmt.Println("Writing scaffold for you to edit...")
-			fmt.Println(filepath.Join("api", o.res.Version,
+			fmt.Println(filepath.Join(o.res.Group, "api", o.res.Version,
```
christopherhein: Should this be `fmt.Println(filepath.Join("api", o.res.Group, o.res.Version, ...)`

camilamacedo86: good catch. 🥇 done.

christopherhein:
Lgtm

camilamacedo86:
HI @mengqiy,

It was done exactly as described in #923 which is an issue created by @DirectXMan12. Also, received an lgtm already. WDYT, could we move forward with this one? Have you any objection @DirectXMan12?

mengqiy:
The implementation is different from what is described in #923.
My understanding is that we still **by default** generated simplified structure. But users can opt in the multi-group structure by add `multigroup: true` stanza into the `PROJECT` file. With `multigroup: true` set, kubebuilder will generated `apis/<group>/<version>`.
IOW, we always need to load `PROJECT` file and then decide which project structure to generate.

camilamacedo86:
Hi @mengqiy, 

Thank you for your input. I understanding what you mean. I have just some questions, in the following scenario:

**Scenario:**
- A preexistent project with many CRD's already scaffolded
- Then, the user runs `kubebuilder create api` without adding the  `multigroup: true` in the PROJECT file. 

**What should happen?**

1. Should we advise the user that it is a new group and then, that is required to enable the project to use multigroup (enable can be a flag that will add the `multigroup: true` in the PROJECT file)?
2. Should we delete the current scaffold files and then add the new structure for all? 

IHMO: What happens if they added CRD's without using the tool or did manual changes on them which are not reflecting the _type.go since they decided customized the makefile? (it for me not would make sense, however, I know a few cases that for many reasons they decided that some specific CRD's should be kept but not so long gen and they have this kind of flexibility with the tool)

So, in POV the current impl in this PR do not really increase the complexity go against the [doc](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/simplified-scaffolding.md#revised-structure). I think just keep the group in the path for all scenarios may keep the things easier to be maintained and used by the users. 

However, please just let me know if you would like to change it and how. ( the answers for the questions 1, and 2 probably would be enough in this case)

mengqiy:
> For the scenario:
A preexistent project with many CRD's already scaffolded
Then, the user runs kubebuilder create api without adding the multigroup: true in the PROJECT file.
What should happen?

If it's for scaffolding an existing group, scaffold it in `api/version` as we used to do.
If it's for a new group, just fail with a message that instruct the user to either add `multigroup: true` in the `PROJECT` file or ask him/her to specify a flag (if we decide to provide a multigroup flag).
We should NOT modify or delete user's files.

Presumably, the sedo code may look like this:
```
proj :=  load("PROJECT")
newGroup := proj.oldGroup != getFlag("--group")
switch {
case newGroup && !proj.MultiGroup:
    // Fail and print instructions for users.
case proj.MultiGroup:
    // scaffold under `apis/<group>/<version>`.
case !proj.MultiGroup:
     // scaffold under `api/<version>`.
}
```

camilamacedo86:
Really tks @mengqiy, 
I will change its implementation as you described. Also, I found the doc https://deploy-preview-1229--kubebuilder.netlify.com/migration/multi-group.html which also be updated with.

mengqiy:
`migration/multi-group.html` covers the migration steps for move existing code to the new multi-group structure.
To have a good UX, we probably should print a message pointing to this doc if we detect that the user want to scaffold another group and already has code in singe-group structure.

camilamacedo86:
Hi @mengqiy and @christopherhein, 

It was changed conformed requested. Also, note that I hidden the previous comment to keep it less verbose and be easier to make the review.

- The [doc](https://book.kubebuilder.io/migration/multi-group.html?highlight=multi#single-group-to-multi-group) is updated
- The changes performed has been tested by the golden script. 
- Now we have a new command to add the multigroup support ( kb multigroup )
- The scaffold just occurs under `apis/<group>/<version>` when the multigroup is set
- The previous files are not changed or moved. 

Please, let me know wdyt and if you see something that needs to be changed here for we are able to merge it.

cmd/main.go
```diff
@@ -112,6 +112,7 @@ func main() {
 
 	rootCmd.AddCommand(
 		newInitProjectCmd(),
+		newMultiGroupProjectCmd(),
```
camilamacedo86: The new cmd to enable the multigroup layout.

docs/book/src/migration/multi-group.md
```diff
@@ -2,16 +2,15 @@
 
 <aside class="note warning">
 
-<h1>Work In Progress: Manual Work Incoming</h1>
+<h1>Note</h1>
```
camilamacedo86: The doc changes since when it be merged this option will be available.

generate_golden.sh
```diff
@@ -40,6 +40,7 @@ scaffold_test_project() {
 
 	oldgopath=$GOPATH
```
camilamacedo86: new tests were added in the generate_golden.sh script to ensure the changes.

pkg/scaffold/api.go
```diff
@@ -271,12 +284,33 @@ func (api *API) scaffoldV2() error {
...
 		if strings.ToLower(r.Group) != strings.ToLower(existingGroup) {
-			return fmt.Errorf("group '%s' is not same as existing group '%s'. Multiple groups are not supported yet.", r.Group, existingGroup)
+			return true
+		}
+	}
+	return false
+}
+
+// validateResourceGroup will return an error if the group cannot be created
+func (api *API) validateResourceGroup(r *resource.Resource) error {
```
camilamacedo86: Here we do the validations to check if the resources can or not be created.

pkg/scaffold/api.go
```diff
@@ -271,12 +284,33 @@ func (api *API) scaffoldV2() error {
...
 }
 
-// Since we support single group only in v2 scaffolding, validate if resource
-// being created belongs to existing group.
-func (api *API) validateResourceGroup(r *resource.Resource) error {
+// isNewGroup will return true if the group informed is diff of the pre-existents
+func (api *API) isNewGroup(r *resource.Resource) bool {
 	for _, existingGroup := range api.project.ResourceGroups() {
 		if strings.ToLower(r.Group) != strings.ToLower(existingGroup) {
-			return fmt.Errorf("group '%s' is not same as existing group '%s'. Multiple groups are not supported yet.", r.Group, existingGroup)
```
camilamacedo86: It was centralized in the validateResourceGroup

testdata/project-v2/PROJECT
```diff
@@ -1,4 +1,5 @@
 domain: testproject.org
+multigroup: true
 repo: sigs.k8s.io/kubebuilder/testdata/project-v2
```
camilamacedo86: mockdata updates

cmd/multigroup.go
```diff
@@ -0,0 +1,82 @@
...
+
+	multiGroupCmd := &cobra.Command{
+		Use:   "multigroup",
+		Short: "Switch to a multi-group layout",
+		Long: `This command will switch the project to a multi-group layout. Note that the tag
+		multigroup: true will be added in the PROJECT file. Also, additional manual work will be needed to
+		move the pre-existent API's and groups for the multi-group layout. For further information
+		see: https://book.kubebuilder.io/migration/multi-group.html?highlight=multi#single-group-to-multi-group`,
+		Example: `# Scaffold a project using the apache2 license with "The Kubernetes authors" as owners
+kubebuilder init --domain example.org --license apache2 --owner "The Kubernetes authors"
```
mengqiy: wrong example

cmd/main.go
mengqiy: I'm wondering what'd be the best layout for a new command.
I'm think about `kubebuilder edit --multigroup=<boolean>`
`kubebuilder edit` may be useful in the future e.g. enabling webhook kustomize patches
@DirectXMan12 @droot WDYT?

cmd/multigroup.go
```diff
@@ -0,0 +1,82 @@
...
+
+package main
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+	"io/ioutil"
+	"log"
+	"os"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/input"
```
mengqiy: nit: imports grouping

pkg/scaffold/api.go
```diff
@@ -174,13 +181,21 @@ func (api *API) scaffoldV2() error {
 			return err
 		}
 
-		fmt.Println(filepath.Join("api", r.Version,
-			fmt.Sprintf("%s_types.go", strings.ToLower(r.Kind))))
+		var path string
+		if api.project.MultiGroup {
+			fmt.Println(filepath.Join("apis", r.Group, r.Version,
```
mengqiy: Probably this is less verbose:
```
path = filepath.Join("apis", r.Group, r.Version, fmt.Sprintf("%s_types.go", strings.ToLower(r.Kind)))
fmt.Println(path)
```

pkg/scaffold/api.go
```diff
@@ -174,13 +181,21 @@ func (api *API) scaffoldV2() error {
...
+				fmt.Sprintf("%s_types.go", strings.ToLower(r.Kind))))
+			path = filepath.Join("apis", r.Group, r.Version, fmt.Sprintf("%s_types.go", strings.ToLower(r.Kind)))
+		} else {
+			fmt.Println(filepath.Join("api", r.Version,
```
mengqiy: Same here

pkg/scaffold/api.go
```diff
@@ -271,12 +284,33 @@ func (api *API) scaffoldV2() error {
...
+	return false
+}
+
+// validateResourceGroup will return an error if the group cannot be created
+func (api *API) validateResourceGroup(r *resource.Resource) error {
+	if api.resourceExists() {
+		return fmt.Errorf("group '%s', version '%s' and kind '%s' already exists.", r.Group, r.Version, r.Kind)
+	}
+	if api.isNewGroup(r) {
+		if api.project.Version == "1" {
```
mengqiy: Hmm... It seems `validateResourceGroup` is used by v2 only, so this won't happen.
And V1 actually only support multi-group structure.

pkg/scaffold/api.go
```diff
@@ -271,12 +284,33 @@ func (api *API) scaffoldV2() error {
...
+			printV1DeprecationWarning()
+			return fmt.Errorf("group '%s' is not same as existing group. Multiple groups are not supported in the v1. ", r.Group)
+		}
+		if !api.project.MultiGroup {
+			return fmt.Errorf("group '%s' is not same as existing group. Multiple groups are not enabled in this project. To enable, use the multigroup command.", r.Group)
+		}
+	}
+	for _, existingKind := range api.project.ResourceVersionKinds(r.Version) {
+		if strings.ToLower(r.Kind) == strings.ToLower(existingKind) {
+			return fmt.Errorf("kind '%s' already exists in the project for the version '%s'. Multiple groups are supported but not for the same Kinds.", r.Kind, r.Version)
```
mengqiy: > Multiple groups are supported but not for the same Kinds.

Why?

generate_golden.sh
```diff
@@ -68,11 +68,11 @@ scaffold_test_project() {
 		$kb create api --group crew --version v1 --kind FirstMate --controller=true --resource=true --make=false
 		$kb create webhook --group crew --version v1 --kind FirstMate --conversion
 		$kb create api --group crew --version v1 --kind Admiral --controller=true --resource=true --namespaced=false --make=false
-		# TODO(droot): Adding a second group is a valid test case and kubebuilder is expected to report an error in this case. It
-		# doesn't do that currently so leaving it commented so that we can enable it later.
-		# $kb create api --group ship --version v1beta1 --kind Frigate --example=false --controller=true --resource=true --make=false
-		$kb create api --group core --version v1 --kind Namespace --example=false --controller=true --resource=false --namespaced=false --make=false
-		# $kb create api --group policy --version v1beta1 --kind HealthCheckPolicy --example=false --controller=true --resource=true --namespaced=false --make=false
+
+		header_text 'Switching to multigroup layout ...'
```
mengqiy: It seems the generated single-group and multi-group code are mixed in the same project `testdata/project-v2`.
I'd suggest keep them separate:
- single group one still in `testdata/project-v2`
- multi-group one probably lives in `testdata/project-v2-multi-group`

The reason behind this is that we sometimes point our users to the testdata projects as examples, so we'd better keep them clean.

pkg/scaffold/api.go
camilamacedo86: For we support it more changes are required. 
Note that the controller layout it controller/kind_controller.go. So, it currently is not supported at all.

generate_golden.sh
camilamacedo86: I agree. Tks

pkg/scaffold/api.go
camilamacedo86: Good catch makes sense.

cmd/main.go
camilamacedo86: I will leave this decision for you folks. Just let me know. 
However:
- Thinking in usability I'd prefer `kb enable --<option>` or `kb --<option>=<bool>` ( true by default ) as it was done. 
- Thinking in maintenance ability I would keep `kb --<option>=<bool>` .

camilamacedo86:
Hi @mengqiy, 

All requests are addressed. Really tks for your review. 
I am looking for the decision of the command and the next steps.

pkg/scaffold/api.go
mengqiy: Implementation details should not stop us from supporting it.
We can change it. Probably change the controller file path to `controller/group_kind_controller.go` if multi-group.

camilamacedo86: See that just adding the group will not solve the problem.
Also, we will have `KindReconciler` than it should be changed as well for GroupKindReconciler?

mengqiy: Good point!

> So, IHMO it shows something which requires more details and is out of its scope. I tracked it in the past as RFE. See: #1202.

Probably it's not out-of-scope.
We have to decide what approach to take before going forward.

To be more consistent with the api directory in multi-group setup, I personally leaning towards using `controllers/<group>/<kind>_controller.go`.

@DirectXMan12 @droot Thoughts?

mengqiy:
Please rebase

cmd/main.go
Adirio: I really dislike the `kubebuilder --multigroup[=<bool>]` because you are adding functionality to the base kubebuilder command when everything else is under subcommands. Subcommands act like verbs in sentences and make the instruction easy to remember to user (UX is quite important in CLI design).

I would opt for one of the following options:
1- `kubebuilder edit --multigroup=<bool>`: in this case I consider the bool mandatory as the subcommand is not descriptive enough any other way.
2- `kubebuilder enable multigroup`: in this case, I would make `multigroup` a parameter and not a flag. Also, equivalent `kubebuilder disable multigroup` and even a  `kuebebuilder toggle multigrup` would also need to be added.

docs/book/src/migration/multi-group.md
```diff
@@ -29,36 +28,29 @@ can check `api/v1/groupversion_info.go` to find that out:
...
+Then, we'll need to fix all the references to the old package name. For
+CronJob, that'll be `main.go` and `controllers/cronjob_controller.go`.  If
+you've added additional files to your project, you'll need to track down
+imports there as well.
 
-Now that we've migrated this to a multi-group project, normal KubeBuilder
-scaffolding won't work (as of KubeBuilder v2.0.0 -- see the warning at the
-top of this page).  You'll need to copy the setup that KubeBuilder
+You'll need to copy the setup that KubeBuilder
 normally does to add new groups for the moment:
```
Adirio: This sentence does not apply anymore. It used to say that, after enabling multigroup, kubebuilder will still deploy in  `api/VERSION/...` and you had to move them manually to `apis/GROUP/VERSION/...` but this PR already does that. For previous files, it has already been explained how they should be moved, this is talking about future files.

docs/book/src/migration/multi-group.md
```diff
@@ -29,36 +28,29 @@ can check `api/v1/groupversion_info.go` to find that out:
...
-```bash
-mv api apis
-mkdir apis/batch
-mv apis/v* batch/
-```
-
-Next, we'll need to update all the references to the old package name. For
-CronJob, that'll be `main.go` and `controllers/cronjob_controller.go`.  If
-you've added additional files to your project, you'll need to track down
-imports there as well.
```
Adirio: Do not remove this.

docs/book/src/migration/multi-group.md
```diff
@@ -29,36 +28,29 @@ can check `api/v1/groupversion_info.go` to find that out:
...
+
+In this way, if the project is not new and has previous APIs already implemented will be
+in the previous structure. See that with the `multigroup`the Kind API's files are
+created under `apis/<group>/<version>` instead of `api/<version>`. So, will be required to manually
+move all for the new structure and then, updated its references.
+
+Then, we'll need to fix all the references to the old package name. For
+CronJob, that'll be `main.go` and `controllers/cronjob_controller.go`.  If
+you've added additional files to your project, you'll need to track down
+imports there as well.
```
Adirio: Remove this part. The previous step-by-step guide was more descriptive and easier to follow.

docs/book/src/migration/multi-group.md
```diff
@@ -29,36 +28,29 @@ can check `api/v1/groupversion_info.go` to find that out:
...
-Next, we'll need to update all the references to the old package name. For
-CronJob, that'll be `main.go` and `controllers/cronjob_controller.go`.  If
-you've added additional files to your project, you'll need to track down
-imports there as well.
-
-Finally, we'll add a new line to `PROJECT` that marks this as
-a multi-group project:
-
+When the command `kubebuilder multigroup` be executed it will add a new line 
+to `PROJECT` that marks this a multi-group project:
```
Adirio: This is a guide:

> Then we will notify kubebuilder that this is a multigroup project by executing:
> ```bash
> kubebuilder multigroup
> ```
> 
> This will modify the `Project` file to mark it as a multi-group project:

pkg/scaffold/api.go
```diff
@@ -31,6 +31,7 @@ import (
 	"sigs.k8s.io/kubebuilder/pkg/scaffold/v1/controller"
 	crdv1 "sigs.k8s.io/kubebuilder/pkg/scaffold/v1/crd"
 	scaffoldv2 "sigs.k8s.io/kubebuilder/pkg/scaffold/v2"
+	ctlv2 "sigs.k8s.io/kubebuilder/pkg/scaffold/v2/controller"
```
Adirio: `controllerv2` using `ctlv2` makes is way less clear just not to type 7 characters. See how `scaffoldv2` is not shortened to `sfdv2` or something similar. `crdv2` is the full path too even if it only 3 letters long.

pkg/scaffold/api.go
```diff
@@ -212,14 +220,12 @@ func (api *API) scaffoldV2() error {
 			return fmt.Errorf("error updating kustomization.yaml: %v", err)
 		}
 
-		if !api.resourceExists() {
```
Adirio: Why do you remove the check if it already exists? It is needed to prevent having duplicates in `PROJECT` if you call the same `kubebuilder create api` twice.

pkg/scaffold/api.go
```diff
@@ -269,12 +275,27 @@ func (api *API) scaffoldV2() error {
...
+		if strings.ToLower(r.Group) != strings.ToLower(existingGroup) {
+			return true
+		}
+	}
+	return false
+}
+
+// validateResourceGroup will return an error if the group cannot be created
+func (api *API) validateResourceGroup(r *resource.Resource) error {
+	if api.resourceExists() {
```
Adirio: You can not move the resource already existing check here, because if you use for example do a `kubebuilder create api --controller=false`, edit the type and then decide you want to create a controller, you would have to do `kubebuilder create api --resource=false` so that it only generates the controller this time, and if you do this check here you would not scaffold the controller.

pkg/scaffold/api.go
camilamacedo86: Hi @mengqiy,

 I also agree that the path `controllers/<group>/<kind>_controller.go.` is more consistent and I could do the changes for we support it. Please, feel free to check.  Problem sorted out -)

pkg/scaffold/api.go
camilamacedo86: It is done. The check was just centralized in the Validation func where validates all.
Closing this one a sort out.

pkg/scaffold/api.go
camilamacedo86: HI @Adirio, 

Note that the change here was just centralized the validations in the validate func. Where and when it is called was NOT changed at all. So, I mean I kept it as it is today since it is not part of the scope. 

See:

- If you run `kubebuilder create api --group webapp --version v1 --kind Guestbook`

> resource=false
> controller=true
**NOTE:** It will not work since the controller requires in the resource in the scaffold.  (it is a bug in the master which should not allow this option)

- if you run `kubebuilder create api --group webapp --version v1 --kind Guestbook`
> resource=true
> controller=false
OR both true it will works

- if run `kubebuilder create api --group webapp --version v1 --kind Guestbook`
> resource=true
> controller=false
And then, try again
With just controller true. 
Then, you will see the msg: `API resource already exists`
Which cames from:

```go
	if api.resourceExists() && !api.Force {
		return fmt.Errorf("API resource already exists")
	}
```

**And that was NOT changed at all.**

The current, behaviour is that is required to use the flag --force. 

Closing it as sort out.

camilamacedo86:
Hi @mengqiy, 

All your suggestions without exception are addressed now. I hope that you like.
Really thank you for all your support and help with. 🥇

docs/book/src/migration/multi-group.md
```diff
@@ -39,36 +38,33 @@ mv apis/v* batch/
 ```
 
 Next, we'll need to update all the references to the old package name. For
-CronJob, that'll be `main.go` and `controllers/cronjob_controller.go`.  If
+CronJob, that'll be `main.go` and `controllers/<group>/cronjob_controller.go`.  If
```
Adirio: If you are finally moving controllers under a GROUP directory, I think you need to add also the equivalent command for them as the bash for the api

> [triple backquote]bash
> mkdir controllers/batch
> mv controller/* controllers/batch/
> [triple backquote]

Didn't write the backquotes because it was formating it as code.

Also, shouldn't the last previous command be `mv apis/v* apis/batch/`?

docs/book/src/migration/multi-group.md
```diff
@@ -39,36 +38,33 @@ mv apis/v* batch/
...
 
-Finally, we'll add a new line to `PROJECT` that marks this as
-a multi-group project:
+Finally, we'll run the command which enable the multi-group layout in the project:
+
+```
+kubebuilder edit --multigroup
+```
 
+When the command `kubebuilder edit --multigroup` be executed it will add a new line
```
Adirio: > ... `kubebuilder edit --multigroup` is executed ...
s/be/is/

docs/book/src/migration/multi-group.md
```diff
@@ -39,36 +38,33 @@ mv apis/v* batch/
...
-- Register the new group-version with the Scheme in `main.go`
-
-- Add a new `<kind>_controller.go` file
+Note that this option indicates to KubeBuilder that this is a multi-group project. 
 
-- Add the controller to the manager in `main.go`
+In this way, if the project is not new and has previous APIs already implemented will be
+in the previous structure. See that with the `multigroup`the Kind API's files are
+created under `apis/<group>/<version>` instead of `api/<version>`. 
+Also, note that the controllers will be created under `pkg/controller/<group>` instead of `pkg/controller`.
```
Adirio: `controllers`? (2 times)

pkg/scaffold/api.go
Adirio: This is not resolved, unmark it please.

```bash
kubebuilder create api --group wepapp --version v1 --kind Guestbook --resource --controller=false
// Change things in kind
kubebuilder create api --group wepapp --version v1 --kind Guestbook --resource=false --controller --force
```

That is perfectly valid. The user thought he didnt need a controller and changed mind so added it later. The second time he doesn't want to remake the resource as he already added fields between the two command calls. Your PR is addding a bug as the validation is only done if `DoResource` is `true`, and as I show you above, you may want to create a controller and not a resource.

pkg/scaffold/scaffold.go
```diff
@@ -58,6 +58,9 @@ type Scaffold struct {
 
 	// Plugins is the list of plugins we should allow to transform our generated scaffolding
 	Plugins []Plugin
+
+	// Indicates that the Scaffold should respect the multigrup structure
+	Multigrup bool
```
Adirio: Typo that shows that you are not using this field as it would have given an error.

pkg/scaffold/util/util.go
```diff
@@ -53,20 +53,35 @@ func GetResourceInfo(r *resource.Resource, repo, domain string) (resourcePackage
...
+	}
+
 	if _, err := os.Stat(resourcePath); os.IsNotExist(err) {
 		if domain, found := coreGroups[r.Group]; found {
 			// TODO: support apiextensions.k8s.io and metrics.k8s.io.
 			// apiextensions.k8s.io is in k8s.io/apiextensions-apiserver/pkg/apis/apiextensions
 			// metrics.k8s.io is in k8s.io/metrics/pkg/apis/metrics
 			resourcePackage := path.Join("k8s.io", "api", r.Group)
-			groupDomain := r.Group
+			if isMultiGroup {
```
Adirio: This is not true, you do not need to change `"api"` to `"apis"` for core resources.

pkg/scaffold/v2/controller/controller_suitetest.go
```diff
@@ -14,15 +14,15 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package v2
+package controller
 
 import (
 	"fmt"
 	"path/filepath"
-
```
Adirio: Leave the blank line here to split standard library to k8s imports

pkg/scaffold/v2/controller/controller_suitetest.go
```diff
@@ -32,16 +32,30 @@ var _ input.File = &ControllerSuiteTest{}
 type ControllerSuiteTest struct {
 	input.Input
 
-	// Resource is the resource to scaffold the controller_kind_test.go file for
+	// Resource is the Resource to make the Controller for
 	Resource *resource.Resource
+
+	// ResourcePackage is the package of the Resource
+	ResourcePackage string
```
Adirio: We just removed this field because it is not needed, leave the call to `GetResourceInfo` in the `Update` method as it was, you have access to `a.Multigroup`over there.

pkg/scaffold/api.go
```diff
@@ -52,6 +53,9 @@ type API struct {
 
 	// Force indicates that the resource should be created even if it already exists.
 	Force bool
+
+	// Indicates that the Resource structure allows support to multigroups
+	Multigroup bool
```
Adirio: I think you are not using this field for anything, as the `Multigroup` is stored in `project.Multigroup` and then it is injecting it in each file.

pkg/scaffold/v2/main.go
```diff
@@ -27,9 +27,9 @@ import (
 )
 
 const (
-	apiPkgImportScaffoldMarker    = "// +kubebuilder:scaffold:imports"
-	apiSchemeScaffoldMarker       = "// +kubebuilder:scaffold:scheme"
-	reconcilerSetupScaffoldMarker = "// +kubebuilder:scaffold:builder"
+	ApiPkgImportScaffoldMarker    = "// +kubebuilder:scaffold:imports"
+	ApiSchemeScaffoldMarker       = "// +kubebuilder:scaffold:scheme"
+	ReconcilerSetupScaffoldMarker = "// +kubebuilder:scaffold:builder"
```
Adirio: Why are you exporting these constants?

plugins/addon/type.go
```diff
@@ -18,8 +20,20 @@ func ReplaceTypes(u *model.Universe) error {
 		return err
 	}
 
+	projectInfo, err := scaffold.LoadProjectFile("PROJECT")
```
Adirio: Plugins should not need to read the `PROJECT` file, they should receive all the required info when called.

pkg/scaffold/v2/controller/controller_suitetest.go
camilamacedo86: See the line https://github.com/kubernetes-sigs/kubebuilder/pull/1198/files/360198f2079528274d762ad755290454f38047bd#diff-ec3d96c196a007869f26993a5a8a7492L122 which was removed and the line https://github.com/kubernetes-sigs/kubebuilder/pull/1198/files/360198f2079528274d762ad755290454f38047bd#diff-ec3d96c196a007869f26993a5a8a7492L122 which was changed. 

Closing as sorted out.

pkg/scaffold/v2/main.go
camilamacedo86: Because they are used out of here. See that in order to keep the project standard the controller and its tests suit were moved for the /pkg/scaffold/v2/controller dir.

Closing as sorted out.

pkg/scaffold/v2/controller/controller_suitetest.go
Adirio: Again, no, dont close.
This field is not required and was removed in a previous PR.
Keeo it as it was and use the multigroup from uodate instead of getinput

pkg/scaffold/v2/main.go
Adirio: Where are they used?

Adirio:
Please stop marking my coments as resolved when they are not resolved. It is really annoying, makes really difficult to find the conversations, and make this PR to look to be ready to merge when it is definetely not.

pkg/scaffold/api.go
camilamacedo86: Hi @Adirio, 

Please, check the explanation above. Note that the validation or when they are called was NOT changed at all. However, if you still checking any issue with, please could you provide the steps to reproduce the scenario which works in the master and not with the changes made in this PR?

pkg/scaffold/util/util.go
camilamacedo86: Sorry, your comment is not clear for me.

camilamacedo86:
HI @mengqiy, 

I think it is ok to be merged. All recommendations made by you are addressed. Please, let me know if has anything else that should be done for we are able to merge it.

pkg/scaffold/api.go
Adirio: You are changing the order:

Consider `api.resourceExists()` returns `true`. In master the following steps are happening to scaffold a v2 resource.

1- `validateResourceGroup()` is called. It returns no error so we keep going. (lines 171-173)
2- The resource is scaffolded. (lines 175-213)
3- `api.resourceExists` is called to check if the resource should be added to the PROJECT file, and as it is `true`we do not add anything to the PROJECT file. (This was done to remove duplicates) (lines 215-223)
4- The controller is scaffolded if needed. (lines 233-256)
5- Main is updated. (lines 258-267)

But in your case:

1- `validateResourceGroup()` is called. It returns an error so we stop here. (lines 172-174)
2- The resource is **NOT** scaffolded. (lines 176-219)
3- The resource is not added to the PROJECT file. (lines 221-227)
4- The controller is **NOT** scaffolded if needed. (lines 237-264)
5- Main is **NOT** updated. (lines 266-275)

As you can see, in master the resource and the controller were scaffolded and main updated, while in your case it yields an error.

Now, the question may be, how do we reach line 171 (or 172 in your case) if `api.resourceExists()` returns `true`. And that has an easy answer, by using the `--force` flag.

1- Create an API: `kubebuilder create api --group webapp --version v1 --kind Guestbook --resource --controller=false`
2- Delete the generated files:
  - `api/v1/guestbook_types.go`
  - `api/v1/groupversion_info.go`
  - `config/samples/webapp_v1_guestbook.yaml`
  - `config/rbac/guestbook_editor_role.yaml`
  - `config/rbac/guestbook_viewer_role.yaml`
  - `config/crd/patches/webhook_in_guestbooks.yaml`
  - `config/crd/patches/cainjection_in_guestbooks.yaml`
  - `config/crd/kustomization.yaml`
  - `config/crd/kustomizeconfig.yaml`
3- Recreate the API: `kubebuilder create api --group webapp --version v1 --kind Guestbook --resource --controller=false --force`

To recreate the api we need to pass `--force` as the API is already in the PROJECT file. Master will recreate the API successfully while your case will not.

pkg/scaffold/v2/controller/controller_suitetest.go
```diff
@@ -32,16 +33,30 @@ var _ input.File = &ControllerSuiteTest{}
...
 }
 
 // GetInput implements input.File
 func (v *ControllerSuiteTest) GetInput() (input.Input, error) {
+
+	v.ResourcePackage, _ = util.GetResourceInfo(v.Resource, v.Repo, v.Domain, v.MultiGroup)
```
Adirio: This was removed in #1219. See below.

pkg/scaffold/v2/controller/controller_suitetest.go
```diff
@@ -32,16 +33,30 @@ var _ input.File = &ControllerSuiteTest{}
 type ControllerSuiteTest struct {
 	input.Input
 
-	// Resource is the resource to scaffold the controller_kind_test.go file for
+	// Resource is the Resource to make the Controller for
 	Resource *resource.Resource
+
+	// ResourcePackage is the package of the Resource
+	ResourcePackage string
```
Adirio: This was removed in #1219. See below.

pkg/scaffold/v2/controller/controller_suitetest.go
```diff
@@ -119,12 +134,10 @@ var _ = AfterSuite(func() {
 // adding import paths and code setup for new types.
 func (a *ControllerSuiteTest) Update() error {
 
-	resourcePackage, _ := util.GetResourceInfo(a.Resource, a.Repo, a.Domain)
-
 	ctrlImportCodeFragment := fmt.Sprintf(`"%s/controllers"
 `, a.Repo)
 	apiImportCodeFragment := fmt.Sprintf(`%s%s "%s/%s"
-`, a.Resource.GroupImportSafe, a.Resource.Version, resourcePackage, a.Resource.Version)
```
Adirio: This was introduced in #1219. `ResourcePackage` is only needed at `Update`, that is why it is computed here instead of in `GetInput` and why the field is not needed, just a local variable.

plugins/addon/type.go
Adirio: This is still unsolved.

pkg/scaffold/util/util.go
Adirio: Doesn't matter if you are in a multigroup or a single-group project. The path `"k8s.io/..."` is for builtin groups and the path will always be `"k8s.io/api/..."`. Just remove this conditional.

pkg/scaffold/api.go
```diff
@@ -269,14 +277,24 @@ func (api *API) scaffoldV2() error {
...
+// isNewGroup will return true if the group informed is diff of the pre-existents
+func (api *API) isNewGroup(r *resource.Resource) bool {
 	for _, existingGroup := range api.project.ResourceGroups() {
-		if !strings.EqualFold(r.Group, existingGroup) {
-			return fmt.Errorf("group '%s' is not same as existing group '%s'. Multiple groups are not supported yet.", r.Group, existingGroup)
+		if strings.ToLower(r.Group) != strings.ToLower(existingGroup) {
+			return true
 		}
 	}
+	return false
```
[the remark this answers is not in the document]
Adirio: Just to make it clear, it will work while project is single group as `api.project.ResourceGroups()` will just have one element but when a multigroup project is present we need to check every element not just the first.

pkg/scaffold/v2/controller/controller_suitetest.go
camilamacedo86: Please, see: https://github.com/kubernetes-sigs/kubebuilder/pull/1219/files#r363710055

pkg/scaffold/v2/controller/controller_suitetest.go
camilamacedo86: Please, see: https://github.com/kubernetes-sigs/kubebuilder/pull/1219/files#r363710055

pkg/scaffold/v2/controller/controller_suitetest.go
camilamacedo86: Please, see: https://github.com/kubernetes-sigs/kubebuilder/pull/1219/files#r363710055

docs/book/src/migration/multi-group.md
```diff
@@ -33,42 +32,47 @@ Then, we'll rename `api` to `apis` to be more clear, and we'll move our
...
-
-- Add a new `<kind>_controller.go` file
+Note that this option indicates to KubeBuilder that this is a multi-group project. 
 
-- Add the controller to the manager in `main.go`
+In this way, if the project is not new and has previous APIs already implemented will be
+in the previous structure. See that with the `multigroup`the Kind API's files are
+created under `apis/<group>/<version>` instead of `api/<version>`. 
+Also, note that the controllers will be created under `pkg/controllers/<group>` instead of `pkg/controllers`. 
+So, will be required to manually move just the old structures and then, updated its references.
```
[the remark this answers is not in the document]
camilamacedo86: Hi @Adirio, 

Thank you for your suggestion. However, I prefer to keep the same structure/text info as it was before with the changes instead of your suggestion. I'd like to leave it for the project maintainers check.  

IMO, this conclusion makes more clear the changes in the layout for the users than just leave the tutorial steps only.

pkg/scaffold/api.go
```diff
@@ -172,13 +173,18 @@ func (api *API) scaffoldV2() error {
...
 		}
 
-		fmt.Println(filepath.Join("api", r.Version,
-			fmt.Sprintf("%s_types.go", strings.ToLower(r.Kind))))
+		var path string
+		if api.project.MultiGroup {
+			path = filepath.Join("apis", r.Group, r.Version, fmt.Sprintf("%s_types.go", strings.ToLower(r.Kind)))
+		} else {
+			path = filepath.Join("api", r.Version, fmt.Sprintf("%s_types.go", strings.ToLower(r.Kind)))
+			fmt.Println(path)
```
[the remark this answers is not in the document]
camilamacedo86: Good spot.

camilamacedo86:
Hi @Adirio, 

Thank you for your collab and suggestions. I think I addressed 99% of them :-) 
Let's see now what the project `approvers/maintainers` think about.

pkg/scaffold/api.go
camilamacedo86: Summary "the scenario which was affected is when the resource files are deleted manually and not from the PROJECT file and the force is used to re-created them"

It is checking now the force in order to allow re-create the resource which should solve it and allow us to keep the validations encapsulated and centralized as it is now. Could you please check?

pkg/scaffold/api.go
```diff
@@ -269,14 +277,24 @@ func (api *API) scaffoldV2() error {
...
 
-// Since we support single group only in v2 scaffolding, validate if resource
-// being created belongs to existing group.
-func (api *API) validateResourceGroup(r *resource.Resource) error {
+// isNewGroup will return true if the group informed is diff of the pre-existents
+func (api *API) isNewGroup(r *resource.Resource) bool {
 	for _, existingGroup := range api.project.ResourceGroups() {
-		if !strings.EqualFold(r.Group, existingGroup) {
-			return fmt.Errorf("group '%s' is not same as existing group '%s'. Multiple groups are not supported yet.", r.Group, existingGroup)
+		if strings.ToLower(r.Group) != strings.ToLower(existingGroup) {
```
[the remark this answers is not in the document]
camilamacedo86: We can keep the `EqualFold`. Probably your change was made after this PR.

pkg/scaffold/api.go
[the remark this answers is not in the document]
camilamacedo86: The func/check is just used/valid when multigroup is not enabled.
Added a comment in order to make it clear. So, has no issue with 

>  it will work while project is single group as api.project.ResourceGroups() will just have one element but when a multigroup project is present we need to check every element not just the first.

Because it is not used for the multigroup scenario at all.

plugins/addon/type.go
camilamacedo86: I think will be not easy to pass the info for the plugin in this scenario. In order to simplify, I think we can use/check the PROJET file. @mengqiy do you see any problem with? Can we move in this way or do you think that it is a problem/blocker for we merge this PR?

docs/book/src/migration/multi-group.md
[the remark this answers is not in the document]
camilamacedo86: Hi @mengqiy could you check the final text info and see if you think that more improvements or changes still required or if we could move forward with?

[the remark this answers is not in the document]
Adirio: Hi @camilamacedo86 
In that case, the english needs to be corrected.

> In this way, if the project is not new and has previous APIs already implemented will be in the previous structure.

I can't even understand what you are trying to say with the first sentence.

> ~See~ **Notice** that, **(comma inserted)** ~with the `multigroup`~ **in a multigroup-enabled project, (comma inserted)** the Kind API's files are created under `apis/<group>/<version>` instead of `api/<version>`. Also, note that the controllers will be created under `pkg/controllers/<group>` instead of `pkg/controllers`. ~So, will be required to manually move just the old structures and then, updated its references.~ That is the reason why we moved the previously generated APIs with the provided scripts in the previous steps. Remember to update the references afterwards.

I added parethenthesis with "comma inserted" where I added a comma as they are hard to notice otherwise even if they are in bold font. The rest of the changes are self-explanatory.

pkg/scaffold/api.go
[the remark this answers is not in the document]
Adirio: It was you who made the change to EqualFold when you added Golang-CI to comply with some linting rules.

pkg/scaffold/api.go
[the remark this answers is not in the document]
Adirio: Just negate the check and swap `true` and `false` and it will work for both:

```golang
// isNewGroup will return true if the group informed is diff of the pre-existents
func (api *API) isNewGroup(r *resource.Resource) bool {
	for _, existingGroup := range api.project.ResourceGroups() {
		if strings.EqualFold(r.Group, existingGroup) {
			return false
		}
	}
	return true
}
```

docs/book/src/migration/multi-group.md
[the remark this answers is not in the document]
camilamacedo86: HI @Adirio,  

Your new suggestion was accepted and addressed. 
Could you please close as solved all your comments addressed in order to make easier the review for others?

pkg/scaffold/v2/controller/controller_suitetest.go
camilamacedo86: Closing the same comment/obs was made 3 times. Just one to address your POV I think is enough :-)

pkg/scaffold/v2/controller/controller_suitetest.go
camilamacedo86: Closing the same comment/obs was made 3 times. Just one to address your POV I think is enough :-)

pkg/scaffold/api.go
```diff
@@ -212,14 +218,12 @@ func (api *API) scaffoldV2() error {
...
 		}
 
-		if !api.resourceExists() {
-			// update scaffolded resource in project file
-			api.project.Resources = append(api.project.Resources,
-				input.Resource{Group: r.Group, Version: r.Version, Kind: r.Kind})
-			err = saveProjectFile("PROJECT", api.project)
-			if err != nil {
-				fmt.Printf("error updating project file with resource information : %v \n", err)
-			}
```
Adirio: The check is needed here. It avoids having duplicates in the PROJECT file.

pkg/scaffold/api.go
Adirio: I added a small comment where a check you removed previously is needed back. As we can run that part by using `--force` we have to prevent it writting to the PROJECT file as it already has it.

pkg/scaffold/v2/controller/controller_suitetest.go
camilamacedo86: Hi @Adirio, 

It is not a big deal at all. However, the `a.ResourcePackage` is used in this case. You just replace the attribute of the struct for the usage of local var instead of. IMO, the previous code has a better approach in this case in order to keep it done as the conttroller.go as well. However, it is clear that you do not agree with and it is fine.

So, if @mengqiy also prefers as you keep this part as done in #1219 + changes I will do that. Otherwise, if @mengqiy be ok with and see that is no required perform changes here I leave it as it is done, OK?

pkg/scaffold/api.go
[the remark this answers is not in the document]
camilamacedo86: your suggestion will not work. Fixed for all scenarios any way.

pkg/scaffold/api.go
camilamacedo86: Solved. Good point.

pkg/scaffold/api.go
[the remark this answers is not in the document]
Adirio: Why do you say it doens't work?

Imagine that we already have "group1", "group2" and "group3":
```
api.project.ResourceGroups() = []string{
    "group1",
    "group2",
    "group`3",
}
```

A new resource for `"group2"` is being created (`r.group = "group2"`):
1- `strings.EqualFold("group2", "group1")` is false, do not return anything
2- `strings.EqualFold("group2", "group2")` is true, return false
`isNewGroup() = false` as "group2" was already in the list of groups

A new resource for `"unknown"` is being created (`r.group = "unknown"`):
1- `strings.EqualFold("unknown", "group1")` is false, do not return anything
2- `strings.EqualFold("unknown", "group2")` is false, do not return anything
3- `strings.EqualFold("unknown", "group3")` is false, do not return anything
4- End of the loop, return true
`isNewGroup() = true` as "unknown" was not in the list of groups.

pkg/scaffold/v2/controller/controller_suitetest.go
Adirio: The thing is that `controller.go` does require it:

https://github.com/kubernetes-sigs/kubebuilder/blob/cf6cec42bdf594eac67cfb8f13eee6f57ba5c47b/pkg/scaffold/v2/controller.go#L79

You can see how `main.go` doesn't need it (same as in this case) and thus just uses a local variable:

https://github.com/kubernetes-sigs/kubebuilder/blob/cf6cec42bdf594eac67cfb8f13eee6f57ba5c47b/pkg/scaffold/v2/main.go#L56

pkg/scaffold/api.go
camilamacedo86: Solved it sees the other comment. 
Closing this one since is fixed and it is very verbose which can make harder others review the PR.

docs/book/src/migration/multi-group.md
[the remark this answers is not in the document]
camilamacedo86: Closing since you marked as solved.

pkg/scaffold/api.go
```diff
@@ -269,14 +279,25 @@ func (api *API) scaffoldV2() error {
...
+func (api *API) isGroupAllowed(r *resource.Resource) bool {
 	for _, existingGroup := range api.project.ResourceGroups() {
-		if !strings.EqualFold(r.Group, existingGroup) {
-			return fmt.Errorf("group '%s' is not same as existing group '%s'. Multiple groups are not supported yet.", r.Group, existingGroup)
+		if !strings.EqualFold(r.Group, existingGroup) && !api.project.MultiGroup {
+			return false
 		}
 	}
+	return true
+}
```
Adirio: I do like this approach, but you should only check once for multigroup at the start (short-circuit) to avoid checking in every loop iteration.

```
// isGroupAllowed will check if the group is == the group used before
// and not allow new groups if the project is not enabled to use multigroup layout
func (api *API) isGroupAllowed(r *resource.Resource) bool {
	// Groups are always allowed in multi-group projects
	if api.project.Multigroup {
		return true
	}
	for _, existingGroup := range api.project.ResourceGroups() {
		if !strings.EqualFold(r.Group, existingGroup) {
			return false
		}
	}
	return true
}
```

pkg/scaffold/api.go
```diff
@@ -207,18 +213,18 @@ func (api *API) scaffoldV2() error {
...
 			api.project.Resources = append(api.project.Resources,
 				input.Resource{Group: r.Group, Version: r.Version, Kind: r.Kind})
-			err = saveProjectFile("PROJECT", api.project)
-			if err != nil {
-				fmt.Printf("error updating project file with resource information : %v \n", err)
+
+			// If the --force was used to re-crete a resource that was created before then,
+			// the PROJECT file will not be updated.
+			if err = saveProjectFile("PROJECT", api.project); err != nil {
+				return fmt.Errorf("error updating project file with resource information : %v \n", err)
```
Adirio: Why are you changing this behavior? I mean it was printing the error but it was not returning an error so now it will not continue to create the controller and update `main.go` if it is unable to save the PROJECT file while before it was just printing the error but scaffolding the controller.

plugins/addon/type.go
Adirio: Would you mind raising an Issue about this when this get merged? So that we consider it for future PR. Plugins will suffer changes so I don't think this should block it.

pkg/scaffold/v2/controller/controller_suitetest.go
camilamacedo86: Done as you requested. 👍 
Marked as solved.

pkg/scaffold/api.go
camilamacedo86: Would make sense continue in this case? 
I think it was just printing the error by mistake. However, your point is valid. 
So, @mengqiy wdyt that would be the correct behaviour in this scenario? Should we return an error when is not possible update the PROJECT file in order to keep the consistency or just print the error in the logs?

pkg/scaffold/api.go
camilamacedo86: good suggestion

pkg/scaffold/v2/controller/controller_suitetest.go
```diff
@@ -32,16 +33,25 @@ var _ input.File = &ControllerSuiteTest{}
...
+			v.Path = filepath.Join("controllers",
+				v.Resource.Group,
+				"suite_test.go")
+		} else {
+			v.Path = filepath.Join("controllers", "suite_test.go")
+		}
 	}
+
 	v.TemplateBody = controllerSuiteTestTemplate
+	v.Input.IfExistsAction = input.Error
```
Adirio: I'm a bit worried about this, it seems to work but my logic sais it shouldn't.

Adirio:
Except for the two topics were we mentioned @mengqiy and the last comment I just added the rest of the PR looks pretty solid to me. LGTM aside from those 3 pending topics.

pkg/scaffold/v2/controller/controller_suitetest.go
camilamacedo86: To add a note:

Just to ensure it we can see the testdata where all is gen with success. 

Screenshot 2020-01-07 at 15 19 30

The same behaviour done for a single group is now done per group.

DirectXMan12:
Yay!  Lmk when those are addressed and we can get this merged :tada:

pkg/scaffold/api.go
DirectXMan12: is the comment above relevant?  My gut says that if we can't save the state of what we're doing to the project file, we should consider that a failure, but I'm not familiar enough with that code.  cc @mengqiy

plugins/addon/type.go
DirectXMan12: I don't think plugins should be manually reading the project file -- they should be passed the desierialized form as input.

pkg/scaffold/v2/controller/controller_suitetest.go
DirectXMan12: @Adirio what specifically are you concerned about here?

Adirio: Lets say Admiral, Captain and Firstmate are created in that order. If I understand that flag correctly, all the three will try to scaffold the suite_test.go file. Admiral will not find the file so no problem. But then Captain does see the file when he tries to create it, as it was already created by Admiral. And if the flag is set to Error when the file exists, my logic sais that it should fail. But as we can see in the testdata is not failing. So it may be me missunderstanding that flag or something. I expected it to be set to input.Skip in order not to do anythin if the suite test file was already created.

pkg/scaffold/api.go
Adirio: That part is executed after scaffolding the resource related files but before scaffolding the controller related ones, or before updating main.go. Anyway we will end in a incomplete state, so I think that it may be better to let kubebuilder scaffold the controller and update main and log instructions to the user to include the resource into the PROJECT file manually.

plugins/addon/type.go
Adirio: I definetely agree.

However, being the plugins in such an early state, I feel that they will change a lot before being available so we may be able to leave that for a plugin related PR and leave it here in this unortodox way. Maybe a vomment saying that plugins shouldnt need to access that file, that the info should be serialized through parameters could be added here in order not to forget.

camilamacedo86: Done the change requested.

pkg/scaffold/v2/controller/controller_suitetest.go
camilamacedo86: I still not getting the concern with since this PR does not change the logic to create/update the suit_tests. The same behaviour done when it has just 1 group is what we do for each new one.

pkg/scaffold/api.go
camilamacedo86: IHMO the PROJECT file should ensure the consistency of the project because of this I do not think that would make sense continue scaffold if was not possible to add the new resource info into it. However, please @mengqiy and @DirectXMan12 just let me know your final decision to change it if is required for we move forward with this PR.

pkg/scaffold/v2/controller/controller_suitetest.go
mengqiy: What @Adirio said seems to be reasonable. 
If that's the case, I wonder why the golden test doesn't catch this :thinking: 

This line probably should be moved into `if v.MultiGroup {}` statement.

pkg/scaffold/v2/webhook/webhook.go
```diff
@@ -62,9 +62,15 @@ func (a *Webhook) GetInput() (input.Input, error) {
 	}
 
 	if a.Path == "" {
-		a.Path = filepath.Join("api", a.Resource.Version,
-			fmt.Sprintf("%s_webhook.go", strings.ToLower(a.Resource.Kind)))
+		if a.MultiGroup {
+			a.Path = filepath.Join("apis", a.Resource.Group, a.Resource.Version,
+				fmt.Sprintf("%s_webhook.go", strings.ToLower(a.Resource.Kind)))
```
mengqiy: Indentation

pkg/scaffold/v2/webhook/webhook.go
```diff
@@ -62,9 +62,15 @@ func (a *Webhook) GetInput() (input.Input, error) {
...
+		} else {
+			a.Path = filepath.Join("api", a.Resource.Version,
+				fmt.Sprintf("%s_webhook.go", strings.ToLower(a.Resource.Kind)))
```
mengqiy: Same here please.

pkg/scaffold/v2/controller/controller_suitetest.go
camilamacedo86: I am still unable to check why with multigroup would an issue be faced with that cannot be faced with the single group layout. However, moved the line to the if as suggested.
Thank you for the review and collab.

pkg/scaffold/api.go
Adirio: Basically right now we do:

1- Scaffold resource related files
2- Update PROJECT file
3- Scaffold controller related files
4- Update main.go

What about swapping the first two elements?

1- Update PROJECT file
2- Scaffold resource related files
3- Scaffold controller related files
4- Update main.go

This would mean moving all the `if !api.resourceExists() { ... }` block just after checking `if err := api.validateResourceGroup(r); err != nil { return err }`.

pkg/scaffold/v2/controller/controller_suitetest.go
Adirio: > What @Adirio said seems to be reasonable.
> If that's the case, I wonder why the golden test doesn't catch this 🤔

I'm really trying to discover why it is not behaving the way I expected. There is a switch that returns an error if this flag is set to `input.Error` as expected (`pkg/scaffold/scaffold.go:259-269`) and this function is called in a loop but any error will make it stop (`pkg/scaffold/scaffold.go:210-214`). Can it be related to the way we are detecting if a file is already present (`pkg/scaffold/scaffold.go:178-183`)? I suspect I have already found some problems with this in an unrelated PR when dealing with the `testdata` directory.

> This line probably should be moved into `if v.MultiGroup {}` statement.

Shouldn't it just be removed? I mean there are still several controllers potentially sharing the suite test file, why should it give an error?

pkg/scaffold/api.go
mengqiy: I don't have a strong opinion about these 2 orders.
It may fail at any step.
@DirectXMan12 @droot thoughts?

pkg/scaffold/v2/controller/controller_suitetest.go
camilamacedo86: Hi @mengqiy,
 
IHMO we should not change the logic here and we should keep it as it has been working in the master. See that in the master we also are able to add many kinds of controllers for the same suit_test. I mean the same scenario occurs with the single group layout. Make no sense for me we change the code base and/or try to find a solution for a problem that we are unable to face/see and or reproduce.

plugins/addon/type.go
camilamacedo86: Closing since it was done.

pkg/scaffold/api.go
camilamacedo86: I like the @Adirio suggestion since if we face some issue to save the PROJECT file than, no scaffold files will be gen.

generate_golden.sh
```diff
@@ -28,63 +28,85 @@ build_kb() {
...
+            header_text 'initializing  project-v2-multigroup ...'
+            $kb init --project-version $version --domain testproject.org --license apache2 --owner "The Kubernetes authors"
+
+            header_text 'Switching to multigroup layout ...'
+            $kb edit --multigroup
+
+            header_text 'Creating APIs ...'
+            $kb create api --group crew --version v1 --kind Captain --controller=true --resource=true --make=false
+            $kb create webhook --group crew --version v1 --kind Captain --defaulting --programmatic-validation
+            $kb create api --group ship --version v1 --kind FirstMate --controller=true --resource=true --make=false
```
mengqiy: I'd suggest exercise besides using `v1` as version. And also I'd suggest tweaking the naming, though it's not critical. e.g. <group: ship, kind: FirstMate> seems to be strange. [These](https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/gopath/src/project/pkg/apis) are the groups and kinds we are using in v1 testdata.
Probably you can do something like this:
```
            $kb create api --group crew --version v1 --kind Captain --controller=true --resource=true --make=false
            $kb create webhook --group crew --version v1 --kind Captain --defaulting --programmatic-validation
            $kb create api --group crew --version v1beta1 --kind FirstMate --controller=true --resource=true --make=false
            $kb create api --group ship --version v1 --kind Frigate --controller=true --resource=true --make=false
            $kb create webhook --group ship --version v1 --kind Frigate --conversion
            $kb create api --group ship --version v2alpha1 --kind Cruiser --controller=true --resource=true --namespaced=false --make=false
            $kb create api --group example-app --version v1 --kind Foo --controller=true --resource=true --make=false
```

cmd/edit.go
```diff
@@ -0,0 +1,89 @@
...
+			projectInfo.MultiGroup = opts.multigroup
+
+			err = saveProjectFile("PROJECT", &projectInfo)
+			if err != nil {
+				log.Fatalf("error updating project file with resource information : %v \n", err)
+			}
+		},
+	}
+
+	editProjectCmd.Flags().BoolVar(&opts.multigroup, "multigroup", true,
```
mengqiy: We really should default it to false.
Because `edit` may have more functionality in the future. It'd be a bad UX if the user always has to do `kubebuilder edit --foo=true --multigroup=false`.

generate_golden.sh
```diff
@@ -28,63 +28,85 @@ build_kb() {
...
+            $kb create webhook --group ship --version v1 --kind FirstMate --conversion
+            $kb create api --group ship --version v1 --kind Admiral --controller=true --resource=true --namespaced=false --make=false
+            $kb create api --group ship --version v1 --kind Captain --controller=true --resource=true --namespaced=false --make=false
+            $kb create api --group example-app --version v1 --kind App --controller=true --resource=true --make=false
+            $kb create api --group example.com --version v1 --kind Example --controller=true --resource=true --make=false
```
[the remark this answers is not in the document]
camilamacedo86: I changed the test. I just would like to ensure/test it with groups names as `something-something` and `something.something` too which are valid in order to cover all scenarios.

cmd/edit.go
camilamacedo86: Make sense .. tks for the good feedback.

pkg/scaffold/v2/controller/controller_suitetest.go
Adirio: @camilamacedo86 Exactly, master was not adding this flag. This PR added it. Thats why I think that you should not add it even if it seems to working by some strange magic we are unable to understand.

pkg/scaffold/api.go
camilamacedo86: Done as suggested  by @Adirio

generate_golden.sh
camilamacedo86: Done. Please, let me know if you liked it now.

Adirio:
NIT: You included a coverage output file

generate_golden.sh
mengqiy: > <group: ship, kind: FirstMate> seems to be strange.

I mean that `FirstMate` is not a kind of ship. This part is strange.
You new changes <ship, Sailor> and <ship, Admiral> IMO still have this problem.

Sorry about bikeshedding this :P

generate_golden.sh
```diff
@@ -28,63 +28,86 @@ build_kb() {
...
+            header_text 'Creating APIs ...'
+            $kb create api --group crew --version v1 --kind Captain --controller=true --resource=true --make=false
+            $kb create webhook --group crew --version v1 --kind Captain --defaulting --programmatic-validation
+            $kb create api --group ship --version v1beta1 --kind Sailor --controller=true --resource=true --make=false
+            $kb create webhook --group ship --version v1beta1 --kind Sailor --conversion
+            $kb create api --group ship --version v1 --kind Admiral --controller=true --resource=true --namespaced=false --make=false
+            $kb create api --group ship --version v2alpha1 --kind Cruiser --controller=true --resource=true --namespaced=false --make=false
+            $kb create api --group sea-creatures --version v1beta1 --kind Kraken --controller=true --resource=true --make=false
+            $kb create api --group sea-creatures --version v1beta2 --kind Leviathan --controller=true --resource=true --make=false
+            $kb create api --group core.policy --version v1 --kind HealthCheckPolicy --controller=true --resource=true --make=false
```
mengqiy: > core.policy

Using `core` here can be potential misleading.
Just call `policy`, or just drop this line is also fine.

generate_golden.sh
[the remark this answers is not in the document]
mengqiy: That's a good point.

generate_golden.sh
mengqiy: If you want to exercise format like `foo.bar`here.
You can probably use `thirdparty.policy` or `foo.policy`.

generate_golden.sh
camilamacedo86: Oh .. catcher that . This are kind of ships and not roles on the ship :-) Fixing it now.

mengqiy:
@camilamacedo86 Thanks for working on this!
/lgtm
/approve

PR #1201, opened 2019-11-14
:running: 🛠 fix prow presubmits

alexeldeib:
Fixes prow e2e optional presubmit (maybe?)

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -19,10 +19,9 @@ K8S_VERSION=$1
 export GO111MODULE=on
 
 # setup go module to create the cluster
-go get sigs.k8s.io/kind@v0.5.1
 
 # You can use --image flag to specify the cluster version you want, e.g --image=kindest/node:v1.13.6, the supported version are listed at https://hub.docker.com/r/kindest/node/tags
-kind create cluster --config test/kind-config.yaml --image=kindest/node:$K8S_VERSION
+kind create cluster -v 4 --retain --wait=1m --config test/kind-config.yaml --image=kindest/node:$K8S_VERSION
 
 # setup the go modules required
 go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,6 @@ services:
 
 before_install:
 - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then mkdir -p /Users/travis/gopath/bin; fi
-- if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then scripts/setup.sh $KIND_K8S_VERSION; fi
 
 before_script:
 
--- a/common.sh
+++ b/common.sh
@@ -156,7 +156,7 @@ function install_dep_by_git {
     DEP_LATEST=$(git describe --abbrev=0 --tags)
     git checkout $DEP_LATEST
     mkdir bin
-    go build -ldflags="-X main.version=$DEP_LATEST" -o bin/dep ./cmd/dep
+    GO111MODULE=off go build -ldflags="-X main.version=$DEP_LATEST" -o bin/dep ./cmd/dep
     popd
   fi
 }
@@ -167,6 +167,17 @@ function install_go_dep {
   if ! is_installed dep ; then
     header_text "Installing dep"
     curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+  fi
+}
+
+function install_kind {
+  header_text "Checking for kind"
+  if ! is_installed kind ; then
+    header_text "Installing kind"
+    KIND_DIR=$(mktemp -d)
+    pushd $KIND_DIR
+    GO111MODULE=on go get sigs.k8s.io/kind@v0.6.0
+    popd
   fi
 }
 
--- a/test_e2e.sh
+++ b/test_e2e.sh
@@ -23,11 +23,14 @@ export TRACE=1
 export GO111MODULE=on
 
 fetch_tools
-install_go_dep
+# This test is used by prow and if the dep not be installed by git then it will face the GOBIN issue.
+install_dep_by_git
+install_kind
 build_kb
 
 setup_envs
 
+source "$(pwd)/scripts/setup.sh" ${KIND_K8S_VERSION}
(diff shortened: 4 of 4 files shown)


alexeldeib:
needs https://github.com/kubernetes/test-infra/pull/15288

alexeldeib:
Hmm, seems like they moved the runner in test-infra? Looking at Kind presumbits, they use `runner.sh` rather than `/usr/local/bin/runner`. Might need to switch this one more time :/

alexeldeib:
oh, the actual runner is `/usr/local/bin/runner.sh` -_-

alexeldeib:
fixing in https://github.com/kubernetes/test-infra/pull/15291

alexeldeib:
cc @camilamacedo86

mengqiy:
@alexeldeib Any update?
If you need more time to fix `pull-kubebuilder-e2e`, I'd suggest merging the fix for `pull-kubebuilder-test` first (maybe in a separate PR)

alexeldeib:
@mengqiy sorry, I've been at Kubecon since the test-infra PR merged and it seems the test runner requires more adjustments on the invocation side. Happy to merge this as is to fix the vanilla test and finish e2e separately. 

If that sounds ok to you, feel free to cancel the hold.

alexeldeib:
Oh, actually I think it's because the files changed to `test_e2e_v1.sh` and `test_e2e_v2.sh`. PR incoming...

alexeldeib:
https://github.com/kubernetes/test-infra/pull/15342

If this PR still doesn't pass the e2e after that, I would go ahead and remove hold + merge anyway to get the other fix done.

alexeldeib:
@mengqiy sorry, I see this was from a dated copy. It seems like the runner is either in wrong workdir or else the job is still not configured correctly -- you can compare against known good jobs like this one: https://github.com/kubernetes/test-infra/blob/8d07c8204a20a0ee062b802655c4f60bcdbf7979/config/jobs/kubernetes-sigs/kind/kind-presubmits.yaml#L13-L15 but I don't see much difference, in test-infra or kind setup. 

While I continue trying to figure that out, https://github.com/kubernetes-sigs/kubebuilder/pull/1201#issuecomment-556451322 still seems reasonable to me.

alexeldeib:
thanks to test-infra improvements since I started this chain of PRs, I'm now able to run the e2e with DinD locally, and I can repro this. Should make it **much** easier to test.

alexeldeib:
@mengqiy tested `pull-kubebuilder-e2e-1-16-2` prowjob locally with this PR + https://github.com/kubernetes/test-infra/pull/15346 and things seem to be working now.

Apologies again for all the trial and error -- the new test-infra tooling should help avoid this kind of mess in the future.

mengqiy:
@alexeldeib It seems we have some problem with kubeadm

alexeldeib:
few additional changes for kind using DinD:

https://github.com/kubernetes-sigs/kind/issues/303
https://github.com/kubernetes-sigs/kind/issues/759 --> think prow runner already handles this
https://github.com/kubernetes/test-infra/pull/11510 --> preset for volumes

alexeldeib:
I can't repro this locally, but it seems even after fixing the immediate issue the clusters still fails to come up. Hoping https://github.com/kubernetes/test-infra/pull/15348 will get the last bits fixed or else need to get some additional debug help from Ben w.r.t kind inside a pod using dind.

I don't want to make the CI more and more red with each attempt to fix it...I keep thinking I have it, but it's been a few rounds of the same. Open to direction on how to proceed without making a bigger mess :slightly_frowning_face:

mengqiy:
It is great that this is finally passing!
Thanks for your hard work!
To merge it, can you please drop the change related to debugging and squashed the commits to a reasonable number?

alexeldeib:
Will do, I’m also relieved this finally works :) 

I’m traveling for personal reasons over the weekend post-Kubecon, but will clean this up for merge Monday if that is okay?

mengqiy:
Sure. Let's merge this on Monday.

alexeldeib:
@camilamacedo86 I think the bots didn't like the two commands in one line.

scripts/setup.sh
```diff
@@ -19,10 +19,9 @@ K8S_VERSION=$1
 export GO111MODULE=on
 
 # setup go module to create the cluster
-go get sigs.k8s.io/kind@v0.5.1
 
 # You can use --image flag to specify the cluster version you want, e.g --image=kindest/node:v1.13.6, the supported version are listed at https://hub.docker.com/r/kindest/node/tags
-kind create cluster --config test/kind-config.yaml --image=kindest/node:$K8S_VERSION
+kind create cluster -v 4 --retain --wait=1m --config test/kind-config.yaml --image=kindest/node:$K8S_VERSION
```
mengqiy: > -v 4 --retain --wait=1m

I wonder if these are related to debugging?
Re.`--wait`, do we have to wait for 1 minute? Is it causing any flakes? It seems the default wait time is 0s.

mengqiy: `kind` cluster is known to be fast to bring up.
Adding 1 minute wait time for bringing up the cluster slow it down ~10%. Look at https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kubebuilder/1201/pull-kubebuilder-e2e-k8s-1-16-2/1198845676235001858, the e2e test only takes <10 minutes.

alexeldeib: I can remove this if desired, but I think it should stay. I matched it to what kind itself does in e2e: https://github.com/kubernetes-sigs/kind/blob/36eae75e067681658d59a31ed679e67e39d9cfc5/hack/ci/e2e-k8s.sh#L88-L93

Without this, kind won't wait at all for the nodes to come up: https://github.com/kubernetes-sigs/kind/blob/6a0de6124de068cd55b24025cf05e8e4ea172641/pkg/internal/cluster/create/actions/waitforready/waitforready.go#L45-L48

If anything, I can reduce `-v` from 4 to 3 I think.

alexeldeib: It's not adding a static 1 minute wait. It's only checking that the nodes actually come up.

alexeldeib: note that it's also a fairly tight loop so there is not really additional wait time

PR #1205, opened 2019-11-16
Add lint check and fix min required issues

camilamacedo86:
**Description**
- add golangci-lint as linter (same used in controller-runtime see [here](https://github.com/kubernetes-sigs/controller-runtime/blob/15b5243f9fde6794ee298dfa9d21d7072bec3a3f/hack/check-everything.sh#L69-L75)). 
- fix the basic tests which are enabled when we run only ` golangci-lint run` in order to add it

Screenshot 2019-11-16 at 18 02 17

Following the fixes made.

```
$ golangci-lint run
pkg/scaffold/api.go:278:6: SA6005: should use !strings.EqualFold(a, b) instead of strings.ToLower(a) != strings.ToLower(b) (staticcheck)
		if strings.ToLower(r.Group) != strings.ToLower(existingGroup) {
		   ^
cmd/util/util.go:25:2: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (gosimple)
	if err != nil {
	^
test/e2e/v2/e2e_suite.go:335:5: S1003: should use !strings.Contains(input, match) instead (gosimple)
	if strings.Index(input, match) == -1 {
	   ^
pkg/scaffold/resource/resource.go:71:18: S1007: should use raw string (`...`) with regexp.MustCompile to avoid having to escape twice (gosimple)
	versionMatch := regexp.MustCompile("^v\\d+(alpha\\d+|beta\\d+)?$")
	                ^
pkg/scaffold/v2/internal/string_utils_test.go:28:2: `got` is unused (structcheck)
	got           string
	^
pkg/scaffold/api.go:40:2: `scaffold` is unused (structcheck)
	scaffold *Scaffold
	^
cmd/init_project.go:99:28: Error return value of `(*github.com/spf13/pflag.FlagSet).MarkDeprecated` is not checked (errcheck)
	cmd.Flags().MarkDeprecated("dep", "use the fetch-deps flag instead")
	                          ^
cmd/init_project.go:100:28: Error return value of `(*github.com/spf13/pflag.FlagSet).MarkDeprecated` is not checked (errcheck)
	cmd.Flags().MarkDeprecated("depArgs", "will be removed with version 1 scaffolding")
	                          ^
cmd/main.go:181:12: Error return value of `cmd.Help` is not checked (errcheck)
			cmd.Help()
			        ^
test/e2e/utils/kubectl.go:47:17: Error return value of `io.WriteString` is not checked (errcheck)
		io.WriteString(stdin, stdinInput)
		              ^

```

**Motivation**
- Issues raised by @Adirio

### DIFF
--- a/cmd/init_project.go
+++ b/cmd/init_project.go
@@ -96,8 +96,13 @@ func (o *projectOptions) bindCmdlineFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&o.dep, "dep", true, "if specified, determines whether dep will be used.")
 	o.depFlag = cmd.Flag("dep")
 	cmd.Flags().StringArrayVar(&o.depArgs, "depArgs", nil, "Additional arguments for dep")
-	cmd.Flags().MarkDeprecated("dep", "use the fetch-deps flag instead")
-	cmd.Flags().MarkDeprecated("depArgs", "will be removed with version 1 scaffolding")
+
+	if err := cmd.Flags().MarkDeprecated("dep", "use the fetch-deps flag instead"); err != nil {
+		log.Printf("error to mark dep flag as deprecated: %v", err)
+	}
+	if err := cmd.Flags().MarkDeprecated("depArgs", "will be removed with version 1 scaffolding"); err != nil {
+		log.Printf("error to mark dep flag as deprecated: %v", err)
+	}
 
 	// boilerplate args
 	cmd.Flags().StringVar(&o.boilerplate.Path, "path", "", "path for boilerplate")
--- a/common.sh
+++ b/common.sh
@@ -17,6 +17,17 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+# check if modules are enabled
+(go mod edit -json &>/dev/null)
+MODULES_ENABLED=$?
+
+MOD_OPT=""
+MODULES_OPT=${MODULES_OPT:-""}
+if [[ -n "${MODULES_OPT}" && $MODULES_ENABLED ]]; then
+    MOD_OPT="-mod=${MODULES_OPT}"
+fi
+
+
 # Enable tracing in this script off by setting the TRACE variable in your
 # environment to any value:
 #
@@ -112,14 +123,13 @@ function prepare_staging_dir {
 
 # fetch k8s API gen tools and make it available under kb_root_dir/bin.
 function fetch_tools {
-  if [ -n "$SKIP_FETCH_TOOLS" ]; then
-    return 0
+  if [ -z "$SKIP_FETCH_TOOLS" ]; then
+    fetch_kb_tools
   fi
-  fetch_kb_tools
 }
 
 function fetch_kb_tools {
-  header_text "fetching tools"
+  header_text "fetching kb tools"
   kb_tools_archive_name="kubebuilder-tools-$k8s_version-$goos-$goarch.tar.gz"
   kb_tools_download_url="https://storage.googleapis.com/kubebuilder-tools/$kb_tools_archive_name"
 
--- a/pkg/scaffold/resource/resource.go
+++ b/pkg/scaffold/resource/resource.go
@@ -68,7 +68,7 @@ func (r *Resource) Validate() error {
 		return fmt.Errorf("group name is invalid: (%v)", err)
 	}
 	// Check if the version is a valid value
-	versionMatch := regexp.MustCompile("^v\\d+(alpha\\d+|beta\\d+)?$")
+	versionMatch := regexp.MustCompile(`^v\d+(alpha\d+|beta\d+)?$`)
 	if !versionMatch.MatchString(r.Version) {
 		return fmt.Errorf(
 			"version must match ^v\\d+(alpha\\d+|beta\\d+)?$ (was %s)", r.Version)
@@ -85,7 +85,7 @@ func (r *Resource) Validate() error {
 	}
 	// Replace the caracter "-" for "" to allow scaffold the go imports
 	r.GroupImportSafe = strings.Replace(r.Group, "-", "", -1)
-    r.GroupImportSafe = strings.Replace(r.GroupImportSafe, ".", "", -1)
+	r.GroupImportSafe = strings.Replace(r.GroupImportSafe, ".", "", -1)
 	return nil
 }
 
--- a/scripts/verify.sh
+++ b/scripts/verify.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+
+#  Copyright 2018 The Kubernetes Authors.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+set -e
+
+source ./common.sh
+
+header_text "running go vet"
+go vet ${MOD_OPT} ../...
+
+header_text "running go fmt"
+go fmt ${MOD_OPT} ../...
+
+header_text "running golangci-lint"
+cd .. # To go to the root of the project
+golangci-lint run --disable-all \
+    --deadline 5m \
+    --enable=misspell \
+    --enable=structcheck \
+    --enable=deadcode \
+    --enable=errcheck \
+    --enable=varcheck \
+    --enable=unparam \
+    --enable=ineffassign \
+    --enable=nakedret \
+    --enable=interfacer \
+    --enable=misspell \
+    --enable=dupl
+
+##todo(camilamacedo86): The following checks requires fixes in the code
+# --enable=golint
+# --enable=gocyclo
+# --enable=goimports
+# --enable=lll
+# --enable=goconst
+# --enable=gosec
+# --enable=maligned
+
+GO111MODULES=on go list -mod=readonly ./...
--- a/test/e2e/utils/kubectl.go
+++ b/test/e2e/utils/kubectl.go
@@ -44,7 +44,7 @@ func (k *Kubectl) CommandWithInput(stdinInput string, cmdOptions ...string) (str
 	}
 	go func() {
 		defer stdin.Close()
-		io.WriteString(stdin, stdinInput)
+		_, _ = io.WriteString(stdin, stdinInput)
 	}()
 	output, err := k.Run(cmd)
 	return string(output), err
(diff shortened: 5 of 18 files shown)


scripts/verify.sh
```diff
@@ -0,0 +1,61 @@
...
+
+##todo(camilamacedo86): The following checks requires fixes in the code
+# --enable=golint
+# --enable=gocyclo
+# --enable=goimports
+# --enable=lll
+# --enable=goconst
+# --enable=gosec
+# --enable=maligned
+
```
camilamacedo86: @Adirio, 

After merging this one, then we can do 1 PR to do the required fixes to enable each kind off test.

camilamacedo86:
c/c @alexeldeib

cmd/util/util.go
```diff
@@ -34,10 +34,12 @@ import (
 // writeIfNotFound returns true if the file was created and false if it already exists
 func WriteIfNotFound(path, templateName, templateValue string, data interface{}) bool {
 	// Make sure the directory exists
-	os.MkdirAll(filepath.Dir(path), 0700)
-
+	err := os.MkdirAll(filepath.Dir(path), 0700)
+	if err != nil {
+		return false
```
Adirio: The description of the function says that it returns `false` if it already exists and here it is returning false if we are unable to make the directories. I would suggest using an `error` as a second return argument and return `false, err`.

cmd/init_project.go
```diff
@@ -96,8 +96,13 @@ func (o *projectOptions) bindCmdlineFlags(cmd *cobra.Command) {
...
 	o.depFlag = cmd.Flag("dep")
 	cmd.Flags().StringArrayVar(&o.depArgs, "depArgs", nil, "Additional arguments for dep")
-	cmd.Flags().MarkDeprecated("dep", "use the fetch-deps flag instead")
-	cmd.Flags().MarkDeprecated("depArgs", "will be removed with version 1 scaffolding")
+
+	if err := cmd.Flags().MarkDeprecated("dep", "use the fetch-deps flag instead"); err != nil {
+		log.Fatalf("error to mark dep flag as deprecated: %v", err)
+	}
+	if err := cmd.Flags().MarkDeprecated("depArgs", "will be removed with version 1 scaffolding"); err != nil {
+		log.Fatalf("error to mark dep flag as deprecated: %v", err)
```
Adirio: Isn't fatal too high for being unable to mark it as deprecated?

pkg/scaffold/resource/resource.go
```diff
@@ -75,7 +79,7 @@ func (r *Resource) Validate() error {
 	r.GroupImportSafe = strings.Replace(r.Group, "-", "", -1)
 	r.GroupImportSafe = strings.Replace(r.GroupImportSafe, ".", "", -1)
 
-	versionMatch := regexp.MustCompile("^v\\d+(alpha\\d+|beta\\d+)?$")
+	versionMatch := regexp.MustCompile(versionRegex)
```
Adirio: As you are already moving the pattern to a constant, why don't you move the regex to a `var` so that we do not compile it every time `Validate` gets called?

Adding after the `const` declaration
```
var (
	versionMatch := regexp.MustCompile(versionRegex)
)
```

test/e2e/utils/kubectl.go
```diff
@@ -44,7 +44,7 @@ func (k *Kubectl) CommandWithInput(stdinInput string, cmdOptions ...string) (str
 	}
 	go func() {
 		defer stdin.Close()
-		io.WriteString(stdin, stdinInput)
+		_, err = io.WriteString(stdin, stdinInput)
```
Adirio: You are not using `err` so you should use `_` instead.

camilamacedo86: The error is used in the return.

Adirio: Wait, you are using the `err` variable from outside the goroutine. That's even more danagerous (race conditions) and you won't know which `err` (the one returned by `io.WriteString(stdin, stdinInput)` or the one returned by `k.Run(cmd)`) you are returning. Maybe the first fails and the second doesn't and a real error is overriden by a `nil`.

cmd/util/util.go
camilamacedo86: Good spot it should fail.

cmd/init_project.go
camilamacedo86: Make sense 👍

camilamacedo86: Solved

pkg/scaffold/resource/resource.go
camilamacedo86: The var is used only this function, the has no need and should not be global.

camilamacedo86: Indeed I moved the const for the above line.

test/e2e/utils/kubectl.go
Adirio: This doesn't solve the issue as it is running in a goroutine.

pkg/scaffold/resource/resource.go
Adirio: Using globals is not a problem as they are not exported. Compiling a regex is resource hungry and shouldn't be done multiple times unless the pattern is dynamic.

camilamacedo86:
it is rebased with the master now to pass in the new prow tests @mengqiy

camilamacedo86:
Hi @mengqiy, 

This one is a simple one. We are just adding the lint as it is used in the controller-runtime and we spoke in other PR's/issues. Could we merge this one or has anything here that you think that is required improvements/changes in order to?

cmd/util/util.go
```diff
@@ -34,10 +34,12 @@ import (
 // writeIfNotFound returns true if the file was created and false if it already exists
 func WriteIfNotFound(path, templateName, templateValue string, data interface{}) bool {
 	// Make sure the directory exists
-	os.MkdirAll(filepath.Dir(path), 0700)
-
+	err := os.MkdirAll(filepath.Dir(path), 0700)
+	if err != nil {
+		log.Fatalf("error to create the file %s : %v", filepath.Dir(path), err)
```
DirectXMan12: nit: should be "failed to ensure that path %q exists: %v"

common.sh
```diff
@@ -112,14 +123,22 @@ function prepare_staging_dir {
...
+    fetch_kb_tools
+  fi
+}
+
+
+function fetch_go_lint {
+  header_text "Checking for gometalinter.v2"
+  if ! is_installed golangci-lint; then
+    header_text "Installing golangci-lint"
+    curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.21.0
```
DirectXMan12: I'm not a huge fan of `curl | bash`, and this script just downloads a release for the current platform (which we can easily get from `go env GOOS` and `go env GOARCH`).  Can we just do that directly instead?

pkg/scaffold/resource/resource.go
```diff
@@ -67,8 +67,14 @@ func (r *Resource) Validate() error {
 	if err := IsDNS1123Subdomain(r.Group); err != nil {
 		return fmt.Errorf("group name is invalid: (%v)", err)
 	}
-	// Check if the version is a valid value
-	versionMatch := regexp.MustCompile("^v\\d+(alpha\\d+|beta\\d+)?$")
+
+	// Remove the . and - from the group to do GO imports in the scaffold project
+	r.GroupImportSafe = strings.Replace(r.Group, "-", "", -1)
```
DirectXMan12: what's up with this change here?

pkg/scaffold/resource/resource.go
```diff
@@ -67,8 +67,14 @@ func (r *Resource) Validate() error {
...
+	r.GroupImportSafe = strings.Replace(r.GroupImportSafe, ".", "", -1)
+
+	const versionRegex = "^v\\d+(alpha\\d+|beta\\d+)?$"
+	versionMatch := regexp.MustCompile(versionRegex)
```
DirectXMan12: why is this being split into two lines

common.sh
camilamacedo86: We could get by  using go, however, it is NOT recommended. 
See here: 

Screenshot 2019-12-05 at 22 02 40

And that indeed we already changed in the controller-runtime as well. https://github.com/kubernetes-sigs/controller-runtime/pull/700

pkg/scaffold/resource/resource.go
camilamacedo86: I added in the first comment all fixes made. 
Note that was just fixed the basic ones to add the lint.

The following is the issue fixed here. 

> pkg/scaffold/resource/resource.go:71:18: S1007: should use raw string (`...`) with regexp.MustCompile to avoid having to escape twice (gosimple)
> 	versionMatch := regexp.MustCompile("^v\\d+(alpha\\d+|beta\\d+)?$")

common.sh
DirectXMan12: No, I mean that script just does a whole bunch of nonsense that resolves to 

```
OS=$(go env GOOS)
ARCH=$(go env GOARCH)

curl https://github.com/golangci/golangci-lint/releases/vx.y.z./golang-ci-lint-$OS-ARCH.tar.gz # or something like this
```

We should just do that directly (and fix CR to do that as well).   I've got less of a problem if it's just in CI, but I'd prefer not to be running arbitrary bash scripts from the internet.

pkg/scaffold/resource/resource.go
DirectXMan12: just use a raw string:

```go
regexp.MustCompile(`^v\d+(alpha\d+|beta\d+)?$`)
```

(raw strings are the kind surrounded by backticks instead of double quotes)

camilamacedo86: True :-) I was very silly now :-P

camilamacedo86: Done.

pkg/scaffold/resource/resource.go
camilamacedo86: Fixed

cmd/util/util.go
camilamacedo86: It was gone with the rebase 👍

common.sh
camilamacedo86: I understand your concerns. However, since it is the way suggested by golangci-lint, I just moved it for the CI and added a note in the CONTRIBUTION docs

camilamacedo86: Please, let me know if you are ok with.

camilamacedo86:
Hi @DirectXMan12, 

It is ok now. After the merge with the master was required to solve as well:

```

pkg/scaffold/v1/webhook/util.go:52:99: getResourceInfo - result 1 (string) is never used (unparam)
func getResourceInfo(coreGroups map[string]string, r *resource.Resource, in input.Input) (string, string) {
                                                                                                  ^
pkg/scaffold/v1/webhook/util.go:52:108: getResourceInfo - result groupDomain is never used (unparam)
func getResourceInfo(coreGroups map[string]string, r *resource.Resource, in input.Input) (resourcePackage, groupDomain string) {
```

Please, let me know if has anything else that we should address to be able to merge this one.

DirectXMan12:
minor tweak, squash

/lgtm
/approve

PR #1209, opened 2019-11-19
Resource enhancement

Adirio:
NOTE: I suggest reviewing the PR commit by commit as it will be easier. If needed, I will squash the commits afterwards.

# Description
Resource validation was being called a couple dozen times when one is enough. Plural form of the Resources was also being computed and stored by multiple scaffold files. This PR avoids all the duplication in code, time and memory.

### DIFF
--- a/pkg/scaffold/resource/resource.go
+++ b/pkg/scaffold/resource/resource.go
@@ -49,6 +49,10 @@ type Resource struct {
 	// CreateExampleReconcileBody will create a Deployment in the Reconcile example.
 	// Deprecated: v1
 	CreateExampleReconcileBody bool
+
+	// Plural is the plural form of the API kind.
+	// Does not need to be provided, it gets filled by Init.
+	Plural string
 }
 
 // Validate checks the Resource values to make sure they are valid.
@@ -90,8 +94,9 @@ func (_ *Resource) isEmpty(value string) bool {
 // Init fills Resource fields that can be obtained from other fields.
 func (r *Resource) Init() error {
 	// Obtain the Kind plural
+	r.Plural = flect.Pluralize(strings.ToLower(r.Kind))
 	if len(r.Resource) == 0 {
-		r.Resource = flect.Pluralize(strings.ToLower(r.Kind))
+		r.Resource = r.Plural
 	}
 
 	// Remove the characters "-" & "." from the group to use it as import alias
--- a/pkg/scaffold/resource/resource.go
+++ b/pkg/scaffold/resource/resource.go
@@ -26,17 +26,18 @@ import (
 
 // Resource contains the information required to scaffold files for a resource.
 type Resource struct {
-	// Namespaced is true if the resource is namespaced
+	// Namespaced is true if the resource is namespaced.
 	Namespaced bool
 
-	// Group is the API Group.  Does not contain the domain.
+	// Group is the API Group. Does not contain the domain.
 	Group string
 
-	// GroupImportSafe is the API Group.  Does not contain the domain and it the "-"
-	// It is used to do safe imports.
+	// GroupImportSafe is the API Group cleaned to be used for imports.
+	// Does not contain the domain.
+	// Does not need to be provided, it gets filled by Init.
 	GroupImportSafe string
 
-	// Version is the API version - e.g. v1beta1
+	// Version is the API version - e.g. v1beta1.
 	Version string
 
 	// Kind is the API Kind.
@@ -45,22 +46,20 @@ type Resource struct {
 	// Resource is the API Resource.
 	Resource string
 
-	// ShortNames is the list of resource shortnames.
-	ShortNames []string
-
-	// CreateExampleReconcileBody will create a Deployment in the Reconcile example
+	// CreateExampleReconcileBody will create a Deployment in the Reconcile example.
+	// Deprecated: v1
 	CreateExampleReconcileBody bool
 }
 
 // Validate checks the Resource values to make sure they are valid.
 func (r *Resource) Validate() error {
-	if r.isGroupEmpty() {
+	if r.isEmpty(r.Group) {
 		return fmt.Errorf("group cannot be empty")
 	}
-	if r.isVersionEmpty() {
+	if r.isEmpty(r.Version) {
 		return fmt.Errorf("version cannot be empty")
 	}
-	if r.isKindEmpty() {
+	if r.isEmpty(r.Kind) {
 		return fmt.Errorf("kind cannot be empty")
 	}
 	// Check if the Group has a valid value for for it
@@ -78,33 +77,28 @@ func (r *Resource) Validate() error {
 		return fmt.Errorf("kind must be PascalCase (expected %s was %s)", flect.Pascalize(r.Kind), r.Kind)
 	}
 
-	// todo: move it for the proper place since they are not validations and then, should not be here
-	// Add in r.Resource the Kind plural
-	if len(r.Resource) == 0 {
-		r.Resource = flect.Pluralize(strings.ToLower(r.Kind))
-	}
-	// Replace the caracter "-" for "" to allow scaffold the go imports
-	r.GroupImportSafe = strings.Replace(r.Group, "-", "", -1)
-    r.GroupImportSafe = strings.Replace(r.GroupImportSafe, ".", "", -1)
 	return nil
 }
 
-// isKindEmpty will return true if the --kind flag do not be informed
-// NOTE: required check if the flags are assuming the other flags as value
-func (r *Resource) isKindEmpty() bool {
-	return len(r.Kind) == 0 || r.Kind == "--group" || r.Kind == "--version"
+// isEmpty will return true if the provided flag is not set or the value is missing.
+// If provided without value it may take another flag as value, so we need to
+// check if it starts by a dash.
+func (_ *Resource) isEmpty(value string) bool {
+	return len(value) == 0 || strings.HasPrefix(value, "-")
 }
 
-// isVersionEmpty will return true if the --version flag do not be informed
-// NOTE: required check if the flags are assuming the other flags as value
-func (r *Resource) isVersionEmpty() bool {
-	return len(r.Version) == 0 || r.Version == "--group" || r.Version == "--kind"
-}
+// Init fills Resource fields that can be obtained from other fields.
+func (r *Resource) Init() error {
+	// Obtain the Kind plural
+	if len(r.Resource) == 0 {
+		r.Resource = flect.Pluralize(strings.ToLower(r.Kind))
+	}
 
-// isVersionEmpty will return true if the --group flag do not be informed
-// NOTE: required check if the flags are assuming the other flags as value
-func (r *Resource) isGroupEmpty() bool {
-	return len(r.Group) == 0 || r.Group == "--version" || r.Group == "--kind"
+	// Remove the characters "-" & "." from the group to use it as import alias
+	r.GroupImportSafe = strings.Replace(r.Group, "-", "", -1)
+	r.GroupImportSafe = strings.Replace(r.GroupImportSafe, ".", "", -1)
+
+	return nil
 }
 
 // The following code came from "k8s.io/apimachinery/pkg/util/validation"
--- a/pkg/scaffold/resource/resource_test.go
+++ b/pkg/scaffold/resource/resource_test.go
@@ -95,16 +95,19 @@ var _ = Describe("Resource", func() {
 			instance := &Resource{Group: "crew", Kind: "FirstMate", Version: "v1"}
 			Expect(instance.Validate()).To(Succeed())
 			Expect(instance.Init()).To(Succeed())
+			Expect(instance.Plural).To(Equal("firstmates"))
 			Expect(instance.Resource).To(Equal("firstmates"))
 
 			instance = &Resource{Group: "crew", Kind: "Fish", Version: "v1"}
 			Expect(instance.Validate()).To(Succeed())
 			Expect(instance.Init()).To(Succeed())
+			Expect(instance.Plural).To(Equal("fish"))
 			Expect(instance.Resource).To(Equal("fish"))
 
 			instance = &Resource{Group: "crew", Kind: "Helmswoman", Version: "v1"}
 			Expect(instance.Validate()).To(Succeed())
 			Expect(instance.Init()).To(Succeed())
+			Expect(instance.Plural).To(Equal("helmswomen"))
 			Expect(instance.Resource).To(Equal("helmswomen"))
 		})
 
@@ -112,6 +115,7 @@ var _ = Describe("Resource", func() {
 			instance := &Resource{Group: "crew", Kind: "Cat", Version: "v1"}
 			Expect(instance.Validate()).To(Succeed())
 			Expect(instance.Init()).To(Succeed())
+			Expect(instance.Plural).To(Equal("cats"))
 			Expect(instance.Resource).To(Equal("cats"))
 		})
 
--- a/pkg/scaffold/resource/resource_test.go
+++ b/pkg/scaffold/resource/resource_test.go
@@ -94,38 +94,45 @@ var _ = Describe("Resource", func() {
 		It("should default the Resource by pluralizing the Kind", func() {
 			instance := &Resource{Group: "crew", Kind: "FirstMate", Version: "v1"}
 			Expect(instance.Validate()).To(Succeed())
+			Expect(instance.Init()).To(Succeed())
 			Expect(instance.Resource).To(Equal("firstmates"))
 
 			instance = &Resource{Group: "crew", Kind: "Fish", Version: "v1"}
 			Expect(instance.Validate()).To(Succeed())
+			Expect(instance.Init()).To(Succeed())
 			Expect(instance.Resource).To(Equal("fish"))
 
 			instance = &Resource{Group: "crew", Kind: "Helmswoman", Version: "v1"}
 			Expect(instance.Validate()).To(Succeed())
+			Expect(instance.Init()).To(Succeed())
 			Expect(instance.Resource).To(Equal("helmswomen"))
 		})
 
 		It("should allow Cat as a Kind", func() {
 			instance := &Resource{Group: "crew", Kind: "Cat", Version: "v1"}
 			Expect(instance.Validate()).To(Succeed())
+			Expect(instance.Init()).To(Succeed())
 			Expect(instance.Resource).To(Equal("cats"))
 		})
 
 		It("should allow hyphens in group names", func() {
 			instance := &Resource{Group: "my-project", Kind: "Cat", Version: "v1"}
 			Expect(instance.Validate()).To(Succeed())
+			Expect(instance.Init()).To(Succeed())
 			Expect(instance.GroupImportSafe).To(Equal("myproject"))
 		})
 
 		It("should allow dots in group names", func() {
 			instance := &Resource{Group: "example.com", Kind: "Cat", Version: "v1"}
 			Expect(instance.Validate()).To(Succeed())
+			Expect(instance.Init()).To(Succeed())
 			Expect(instance.GroupImportSafe).To(Equal("examplecom"))
 		})
 
 		It("should keep the Resource if specified", func() {
 			instance := &Resource{Group: "crew", Kind: "FirstMate", Version: "v1", Resource: "myresource"}
 			Expect(instance.Validate()).To(Succeed())
+			Expect(instance.Init()).To(Succeed())
 			Expect(instance.Resource).To(Equal("myresource"))
 		})
 	})
--- a/pkg/scaffold/v1/controller/controller.go
+++ b/pkg/scaffold/v1/controller/controller.go
@@ -23,7 +23,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/gobuffalo/flect"
 	"sigs.k8s.io/kubebuilder/pkg/scaffold/input"
 	"sigs.k8s.io/kubebuilder/pkg/scaffold/resource"
 )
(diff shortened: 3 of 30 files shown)


pkg/scaffold/resource/resource.go
```diff
@@ -78,33 +81,29 @@ func (r *Resource) Validate() error {
...
 
-// isKindEmpty will return true if the --kind flag do not be informed
-// NOTE: required check if the flags are assuming the other flags as value
-func (r *Resource) isKindEmpty() bool {
-	return len(r.Kind) == 0 || r.Kind == "--group" || r.Kind == "--version"
+// isEmpty will return true if the provided flag is not set or the value is missing.
+// If provided without value it may take another flag as value, so we need to
+// check if it starts by a dash.
+func (_ *Resource) isEmpty(value string) bool {
+	return len(value) == 0 || strings.HasPrefix(value, "-")
```
camilamacedo86: Note that it was not made generic in order to verify the specific values. We should not consider a text with the prefix `-` as empty since it can be a valid value for the k8s API. However, may you could verify here all possibilities `--group`, `--version`, `--kind`. However, the name of the func should be meaningful for it.

pkg/scaffold/resource/resource.go
```diff
@@ -78,33 +81,29 @@ func (r *Resource) Validate() error {
...
 }
 
-// isKindEmpty will return true if the --kind flag do not be informed
-// NOTE: required check if the flags are assuming the other flags as value
-func (r *Resource) isKindEmpty() bool {
-	return len(r.Kind) == 0 || r.Kind == "--group" || r.Kind == "--version"
+// isEmpty will return true if the provided flag is not set or the value is missing.
+// If provided without value it may take another flag as value, so we need to
+// check if it starts by a dash.
+func (_ *Resource) isEmpty(value string) bool {
```
camilamacedo86: Note that you can remove (_ *Resource) and can the func drectly. Has no need to use `_` here.

pkg/scaffold/resource/resource_test.go
```diff
@@ -94,38 +94,49 @@ var _ = Describe("Resource", func() {
...
+			Expect(instance.Init()).To(Succeed())
+			Expect(instance.Plural).To(Equal("helmswomen"))
 			Expect(instance.Resource).To(Equal("helmswomen"))
 		})
 
 		It("should allow Cat as a Kind", func() {
 			instance := &Resource{Group: "crew", Kind: "Cat", Version: "v1"}
 			Expect(instance.Validate()).To(Succeed())
+			Expect(instance.Init()).To(Succeed())
+			Expect(instance.Plural).To(Equal("cats"))
```
camilamacedo86: Has no need to add the plural since the {{ .Resource.Resource  }} has been using to keep it.

pkg/scaffold/resource/resource_test.go
```diff
@@ -94,38 +94,49 @@ var _ = Describe("Resource", func() {
 		It("should default the Resource by pluralizing the Kind", func() {
 			instance := &Resource{Group: "crew", Kind: "FirstMate", Version: "v1"}
 			Expect(instance.Validate()).To(Succeed())
+			Expect(instance.Init()).To(Succeed())
+			Expect(instance.Plural).To(Equal("firstmates"))
 			Expect(instance.Resource).To(Equal("firstmates"))
 
 			instance = &Resource{Group: "crew", Kind: "Fish", Version: "v1"}
 			Expect(instance.Validate()).To(Succeed())
+			Expect(instance.Init()).To(Succeed())
```
camilamacedo86: What would mean the init?

pkg/scaffold/v1/controller/controller.go
```diff
@@ -199,8 +191,8 @@ type Reconcile{{ .Resource.Kind }} struct {
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=apps,resources=deployments/status,verbs=get;update;patch
 {{ end -}}
-// +kubebuilder:rbac:groups={{.GroupDomain}},resources={{ .Plural }},verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups={{.GroupDomain}},resources={{ .Plural }}/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups={{ .GroupDomain }},resources={{ .Resource.Plural }},verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups={{ .GroupDomain }},resources={{ .Resource.Plural }}/status,verbs=get;update;patch
```
camilamacedo86:
```suggestion
// +kubebuilder:rbac:groups={{ .GroupDomain }},resources={{ .Resource.Resource }}/status,verbs=get;update;patch
```

pkg/scaffold/v1/controller/controller.go
```diff
@@ -199,8 +191,8 @@ type Reconcile{{ .Resource.Kind }} struct {
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=apps,resources=deployments/status,verbs=get;update;patch
 {{ end -}}
-// +kubebuilder:rbac:groups={{.GroupDomain}},resources={{ .Plural }},verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups={{.GroupDomain}},resources={{ .Plural }}/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups={{ .GroupDomain }},resources={{ .Resource.Plural }},verbs=get;list;watch;create;update;patch;delete
```
camilamacedo86:
```suggestion
// +kubebuilder:rbac:groups={{ .GroupDomain }},resources={{ .Resource.Resource }},verbs=get;list;watch;create;update;patch;delete
```

pkg/scaffold/v2/controller.go
```diff
@@ -86,8 +77,8 @@ type {{ .Resource.Kind }}Reconciler struct {
 	Scheme *runtime.Scheme
 }
 
-// +kubebuilder:rbac:groups={{.GroupDomain}},resources={{ .Plural }},verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups={{.GroupDomain}},resources={{ .Plural }}/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups={{ .GroupDomain }},resources={{ .Resource.Plural }},verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups={{ .GroupDomain }},resources={{ .Resource.Plural }}/status,verbs=get;update;patch
```
camilamacedo86: Same ^

pkg/scaffold/resource/resource.go
Adirio: I will check this but I think none of these 3 values can start by a dash. For example versions need to start with `v`. The benefit of checking it like this is that it will be future proof in case we add a new flag.
About the naming, it is a non exported method and its documentation clearly specifies that it checks flags, but if some alternative name such as `isGVKEmpty` feels more natural I can replace it.

pkg/scaffold/resource/resource.go
Adirio: This is how you implement static method in Golang. A package level function would also work. Compilers are intelligent enough to handle both cases the same way and by using a static method we explicitly show that this method belongs to the `Resource` struct.

pkg/scaffold/resource/resource_test.go
Adirio: `kubebuilder create webhook` (and its v1 equivalent `kubebuilder alpha webhook`) offer a `--resource` flag that replaces the default behavior of storing the plural in the `Resource.Resource` field. Different scaffolding files were using a `.Plural` field to ensure that it doesn't use a user-specified resource. Basically I moved the file `.Plural` field into the `Resource.Plural` to avoid calling `flect.Pascalize` and `strings.ToLower` in multiple places.

pkg/scaffold/resource/resource_test.go
Adirio: I wasn't very sure about this method's name either. I was also considering calling it `Fill`. I'm open to name suggestions for this method. It basically computes the `Resource` fields that are derived from others.

pkg/scaffold/v1/controller/controller.go
Adirio: As I said, `Resource.Resource` was being used for two different things, as its value could be changed with a flag for certain commands. Here we do not want the user-specific resource in case he provides it but the plural.

pkg/scaffold/v1/controller/controller.go
Adirio: See above reason.

pkg/scaffold/v2/controller.go
Adirio: See above reason.

Adirio:
@camilamacedo86

> We cannot assume that all Resources added in the project was gen by its tool and that no changes can occur after creating them and exec other commands to do further scaffolds.

I'm not sure if you mean CRDs or `Resource` structs.
In case you mean they can modify CRDs manually or with other commands, I'm not assuming that in any way. The commands that require the struct partially build it from the flags provided.
In case you are talking about the `Resource` structs, they are created from flags for three possible commands: `kubebuilder create api`, `kubebuilder alpha webhook` (v1) and `kubebuilder create webhook` (v2). The first was already calling `Validate`, I just added the other method call. The other two weren't fully validating, but I added it. 

> Note that when some scaffold will be done we need to validate the data required for it and indeed one kind of feature/scaffold can require specific validations that are not required in another one.

I do agree that they may require additional validation, but right now they were just re-validating the resources in case they had one (there were a couple exceptions that the `Validate` method was missing). As the resources only need to be validated once, its just a waste of time. I will add back again the File validation for those files that have a `Validate` method even if none of them will have it for now.

> In my POV, the code has not been duplicated, it has been centralized and reused since in the biggest part of the cases the common validations will be the same to create the resource by using the tool.

The duplication is not in code lines, but in method calls. `kubebuilder create api ...` calls 13 times `Validate` on the same resource.

> See that the plural could be replaced by resources as it was done here

`Resource.Resource` may not have the plural if the user specified manually a `--resource` flag. Resource and Plural are two different things, despite having the same value by default. Remember that Plural was just introduced inside the struct but it was already present outside of Resource.

NOTES:

I'd recommend you try to keep one subject per PR.
The changes here broke the project and it is no passing in the CI.

pkg/scaffold/resource/resource.go
camilamacedo86: So, my point is: we cannot block more scenarios. We can just consider as empty the specific values that group, version and kind can assume when they are not properly filled.

pkg/scaffold/resource/resource.go
camilamacedo86: The Resouce is used when we want to Embedded. In this case, after you change it is a func which has no longer this need.  See: https://golang.org/doc/effective_go.html#embedding

pkg/scaffold/resource/resource_test.go
camilamacedo86: I  understand, however, IMO the only refactory over it that we can do is use the ` {{ .Resource.Resource }} ` and has no need to add the check in the tests.

pkg/scaffold/v1/controller/controller.go
camilamacedo86: Please, see that it has been used in the scaffolds implementations already instead of the plural as point out in the other comments.

camilamacedo86:
HI @Adirio, 

We can create the CRD's manually like any other artefact in the project or using any other tool. Also, we can change the code implementation in the projects as well and because of this, it has its design and we need to ensure that the resources will be validated before doing the scaffolds.

pkg/scaffold/v1/controller/controller.go
camilamacedo86: Screenshot 2019-11-20 at 11 17 02

pkg/scaffold/resource/resource.go
Adirio: Checked it. `Group`s are validated to be a DNS1123 subdomain (that need to start with either a lowercase letter or a digit), `Version`s are validated against a regex that enforces first character to be `v`, and `Kind`s are checked against their PascalCase version which removes dashes. So none of then can start by dashes.

Adirio: And what happens with `kubebuilder create api --kind --controller --group example --version v1`?  Now we are giving as an error `Kind must be PascalCase (expected Controller was --controller)` which is the same issue you were solving but for other flags. There are other flags to be considered too, so the list will grow bigger. And in the future, we may add additional flags, like for example options for plugins, or `--builtin` to use the builtin apis. They are just example (and probably not that useful), but the point is that we need to handle more flags right now and we will probably have to handle even more later. Having to add each new flag to these 3 `if`s is error prone (will be hard to remember to do this and hard to debug to find that it needs to be done). Thats why I went for a flag agnostic method that just checks if they are flags (== start by dash).

pkg/scaffold/resource/resource.go
Adirio: You did not understand me. `func (_ *Resource) isEmpty(value string) bool` is equivalent to `func isEmpty(value string) bool` but making it clearer for future readers that this method is used for the `Resource` type. It doesn't have to do with embedding.

@mengqiy @droot Is there any coding policy against static methods in kubebuilder?

pkg/scaffold/resource/resource_test.go
Adirio: Tests have to be unaware of implementation. Both are expected to be the same in case we do not provide the `--resource` flag, so we should test both. Imagine someone in a later commit replaces plural with a PascalCase version without noticing. We would have a bug that tests are not catching. More tests = better tests always.

pkg/scaffold/v1/controller/controller.go
Adirio: I insist, `resource.Resource` is not guaranteed to be the plural form, tahts only the default behavior. This file requires the plural form, not the one provided by the user. It was using `.Plural` for that. I just moved `.Plural` inside `Resource`, so that it is used as `.Resource.Plural`. In most of the cases `.Resource.Plural` will be the same as `.Resource.Resource` but not in all of them and we need the plural version here.

pkg/scaffold/resource/resource.go
camilamacedo86: The [embedding](https://golang.org/doc/effective_go.html#embedding) is for we use to spef that that func is related to the structure. 

For example person.getNames() or person.addAdress(street). In this case, you change the code to a func to check the values and you are passing the values as arg they no longer came from Resources so has no need for it.

pkg/scaffold/resource/resource.go
camilamacedo86: Please, I am just trying that we cannot block anything that starts with `-` (`|| strings.HasPrefix(value, "-")`). Each attribute has in its own validation and in POV it should be kept in this way.  

The idea of isEmptyGroup,  isEmptyVersion and isEmptyKind is just because if we  ` create api --kind --group example --version v1` for example the value for Kind will be `--group`.  

> `happens with kubebuilder create api --kind --controller --group example --version v1?`

In this case, the --kind is not empty and its value is ` --controller ` then, the specific validation will check if it is or not valid for it.

pkg/scaffold/resource/resource_test.go
camilamacedo86: The pascalized should be removed and we should use the k8s validation to check the Kind values. See : https://github.com/kubernetes-sigs/kubebuilder/pull/1133

pkg/scaffold/resource/resource.go
Adirio: Thats not embedding. Read the link you provided and you will see that. Embedding is using a field of a struct without a name so that the methods of the embedded type can be used from the wrapper instance without needing to access the field.

What you are trying to say is that receivers are used to avoid passing that variable as a parameter to a function. That is true. Static methods are equivalent to functions in the sense that they don't have a receiver (well, they actually have an empty receiver). It is an OOP concept.

So the short answer is, the blank receiver are not needed. But they are a good way to show the relation between the function and the type. The compiler will handle them the same way, but when reading the code, you inmediately know that it relates to that type.

I already mentioned a couple of the main mantainers of this project. If they do not want to have static methods I will remove the receiver.

pkg/scaffold/resource/resource_test.go
Adirio: #1133 still enforces a letter to be in first position, so my argument still aplies.

Adirio:
> We can create the CRD's manually like any other artefact in the project or using any other tool.  Also, we can change the code implementation in the projects as well and because of this, it has its design where a validate func is called for all scaffolds implementations.
>
> See that is required to ensure that the resources will be validated before doing the scaffolds which indeed can be done in any moment and not just after creating them using the feature `kubebuilder create api` as you point out. Note too that many of them will just be made if and when the user does customizations, as for example, for webhooks and metrics.

This PR doesn't interfere with any of that. You can still create the CRDs however you want. Want to do it manually, go ahead. Want to use another tool, feel free. File scaffolds are used to create files, they are not used to validate a file you created manually.

The only thing changed is that the resource will not be validated 13 times as it was right now when you issued a `kubebuilder create api ...` command. The responsability of doing `Resource.Validate` nasn't even moved to a different package. It is just done once instead of 13 times. I added back the ability to provide a Validate function to scaffold files, but they shouldn't validate their resources. The resources are created, validated, filled and then passed to the files. The files should not validate the resources again. You can't design a software making each step of the software re-validate every step that was made previously.

pkg/scaffold/resource/resource_test.go
Adirio: The PascalCase was just an example. Including both fields in tests let us be sure that the behavior won't change without noticing. If we don't include Plural in the tests this may change in a later comit resulting in bugs in the scaffolded files that tests will not be catching.

pkg/scaffold/resource/resource.go
camilamacedo86: Let's see if the following comment can clarify:

Here you have 2 options; keep the embedding and not pass the values as arg since they are embedded already or pass as arg as you did but then make no sense keep the embedding`(_ *Resource)`.

pkg/scaffold/resource/resource.go
Adirio: `--controller` is a valid flag. It avoids being asked if you want to scaffold a controller. So the error given should not be `"--controller" is not a valid kind` but `kind is empty`. And the same as `--controller` there are other flags that could be used and we would have a very long `||` concatenation for each `if`. And if new flags were added in the future, we would still have to come here and update the `if`s. You need to catch any flag, you can't only provide a list of flags that are being catched here.

pkg/scaffold/resource/resource.go
Adirio: Again, that is not called embedding, that is called a receiver. And again, using blank receivers is a OOP approach. I will try to explain you what a static method is in Slack, this is not the place to explain what a static method is.

pkg/scaffold/resource/resource.go
camilamacedo86: The best approach would check if the cobra lib has a way to do it. I could not find a solution using cobra.  The second approach ass you wish. 

My point in the review is:
- We cannot block any value which starts with `-` since it can be a valid value for the attribute in k8s API
- isEmpty has not the responsibly to check if the value is valid or not for each attribute.

Adirio:
I resolved some comments that were too long, the main points made by @camilamacedo86 (please correct me if my summary is not correct) were:

1. Avoid using blank pointer receivers as static methods (`func (_ *Resource) isEmpty(value string) bool`), use a package level function or separate methods as they were before.
2. Do not check flags in a general manner (`strings.HasPrefix(..., "-")`) as a value starting with a `-` may be a valid value for one of the 3 flags.
3. The function `Init` doesn't seem the right approach.

My point of view on those topics:

1. I like the static method version more than the other options.
2. Checking for a `-` will capture not only these 3 flags but others such as `--make`, `--resource`, `--controller`, `--pattern`, `--force`, `--namespaced` or even others that are added in the future.
3. I don't like the name of the fucntion either.

Can others express their opinions?
@droot @mengqiy @DirectXMan12

camilamacedo86:
Hi @Adirio, 

The summary made in https://github.com/kubernetes-sigs/kubebuilder/pull/1209#issuecomment-556088529 is not accurate at all with my review. My points are:

Subject 1 - Improve of Empty GKV validation. 

Here: https://github.com/kubernetes-sigs/kubebuilder/pull/1209/files#diff-1264bafa536db5239815912ade8bc615R90
```
func (_ *Resource) isEmpty(value string) bool {
	return len(value) == 0 || strings.HasPrefix(value, "-")
}
```
- Make no sense defined as empty any value that starts with `-` 
- The ideal solution would be to use cobra to avoid this scenario
- In this scenario make no sense for me the ` (_ *Resource)`. ( my recommendation here will be our keep the (r *Resource) and get the value from `r` as it was done before or if you do an auxiliar generic func as it was done then remove it since it is no longer specific for Resource ).

Subject 2 - Plural.

- Use {{ resource.Resource }} instead of. See:

Screenshot 2019-11-20 at 11 17 02

Subject 3 - Remove the Validation func of all scaffolds impl. 

- I do not think that we should move forward with. See: https://github.com/kubernetes-sigs/kubebuilder/pull/1209#issuecomment-555957682

Adirio:
- Cobra will not change the behavior as it was already noticed to them and they said it was expected behavior: https://github.com/spf13/cobra/issues/909. They say that you need to use either `--flag=` notation or `--flag ""` (this last doesn0t apply here) to solve it. And in https://github.com/spf13/cobra/issues/909#issuecomment-512563379 they clearly say that the flag will not check for emptyness of this value.
- `Resource.Resource` and `Resource.Plural` are two different things. `Resource.Resource` defaults to the plural if it wasn't provided by the user but may be different. Both may be used in different context in scaffolded files. Right now they are different fields, so I am not changing that behavior. The only thing I'm doing is introducing `Plural` inside of `Resource` as previously was being a field of the `File` implementations.

Adirio:
I will split this PR in easier to handle parts

PR #1211, opened 2019-11-20
:book: use v2.2.0 in quick start

mengqiy:

### DIFF
--- a/docs/book/src/quick-start.md
+++ b/docs/book/src/quick-start.md
@@ -9,7 +9,7 @@ This Quick Start guide will cover:
 
 ## Prerequisites
 
-- [go](https://golang.org/dl/) version v1.12+.
+- [go](https://golang.org/dl/) version v1.13+.
 - [docker](https://docs.docker.com/install/) version 17.03+.
 - [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) version v1.11.3+.
 - [kustomize](https://sigs.k8s.io/kustomize/docs/INSTALL.md) v3.1.0+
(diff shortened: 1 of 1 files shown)


docs/book/src/quick-start.md
```diff
@@ -24,11 +24,11 @@ os=$(go env GOOS)
...
 # download kubebuilder and extract it to tmp
-curl -sL https://go.kubebuilder.io/dl/2.1.0/${os}/${arch} | tar -xz -C /tmp/
+curl -sL https://go.kubebuilder.io/dl/2.2.0/${os}/${arch} | tar -xz -C /tmp/
 
 # move to a long-term location and put it on your path
 # (you'll need to set the KUBEBUILDER_ASSETS env var if you put it somewhere else)
-sudo mv /tmp/kubebuilder_2.1.0_${os}_${arch} /usr/local/kubebuilder
+sudo mv /tmp/kubebuilder_2.2.0_${os}_${arch} /usr/local/kubebuilder
 export PATH=$PATH:/usr/local/kubebuilder/bin
 ```
```
camilamacedo86: @mengqiy if the release will be made from the master we need to update [here](https://github.com/kubernetes-sigs/kubebuilder/blob/46a504b19f61d55349e6983a4efbd5c6ac73c4af/docs/book/src/quick-start.md#prerequisites) as well. I mean we need to update the prerequisites. `Go version 1.13+` and `kubectl ` which I think that should be already 1.14.1+

mengqiy: I have updated required golang version to 1.13+.
kubectl version is not a hard requirement for kubebuilder. 1.11 kubectl should still work. So I'd leave it as is.

camilamacedo86: Sorry, I wrote and so kubectl when I was thinking about k8s API cluster. 
I note that we are not adding the info of the min version required, however, I think we should add that and in this case the mn version tested is 1.14.1+.

Issue #1218, opened 2019-11-23
[Umbrella issue] Scaffolding enhancements


Adirio:
# Description

The following issue will group together all the different enhancements aimed to obtain a smoother file scaffolding system that improves redability, removes duplicate sections, avoids calling several methods multiple times, and centralizes functionality.

Some of the enhancements will be straight forward and will be directly implemented by a PR while others will have their own issue in order to contain the discussion. Please, keep comments in this issue just related to general topics related this issue or suggestions on which other enhancements to apply and use the respective issues or PR to talk about each of the enhancements.

# Motivation

The main scheme followed by kubebuilder is the following:

1. Create a Cobra command and subcommands
2. This command creates a scaffolder for the specific command that:
   1. Sets default values
   2. Validates the provided input
   3. Creates the specific files

### Issues found

- Some commands do not have an associated scaffolder and their scaffolding logic is done in the command itself (see `cmd/webhook_1.go` or  `cmd/webhook_2.go` as examples). 
- Validations are spread all accross the project. Some of them are done by the command and some by the associated scaffolder.
- Files are scaffolded with multiple non-needed fields (for example, `ProjectPath` is not required for any file and is provided to all, in `pkg/scaffold/v1/crd/doc.go` a field called `Comments` is not used, ...), information about the resource is kept outside the resource (`GroupDomain`, `ResourcePackage`), or stored in two places (`Resource.Resource` vs `Plural`).
- Resource is being validated by the scaffolders and by each file. This implies that, for example, for a single `kubebuilder create api ...` command the resource is validated 13 times.

### Proposed solutions:
- Scaffolders for commands that do not have one should be created to keep the scaffolding logic in `pkg/scaffold` and not in `cmd`.
- Validation, unless directly related to the command, should be delegated to the scaffolder.
- Design interfaces for both scaffolders and files.
- Re-design the `Resource` struct to contain the fields that are outside: `GroupDomain`, `Plural` and `ResourcePackage`.
    - Determine if both `Resource.Resource` and `(Resource.)Plural` are needed or keep only one of them.

### Related Issues and Pull Requests

Issues:
- #1226
- #1264
- #1358

Pull requests:
- #1219
- #1221
- #1248
- #1252
- #1255
- #1265
- #1266
- #1313
- #1314
- #1316
- #1325
- #1330
- #1339
- #1340
- #1348
- #1349
- #1359
- #1375

More coming soon

/kind feature

[...]

PR #1219, opened 2019-11-23
Remove scaffolding fields that were not used

Adirio:
Removes some fields that `File` structs had but are not being used.

/kind cleanup

### DIFF
--- a/cmd/webhook_v1.go
+++ b/cmd/webhook_v1.go
@@ -77,8 +77,8 @@ This command is only available for v1 scaffolding project.
 				&webhook.AdmissionWebhookBuilder{Resource: o.res, Config: webhook.Config{Server: o.server, Type: o.webhookType, Operations: o.operations}},
 				&webhook.AdmissionWebhooks{Resource: o.res, Config: webhook.Config{Server: o.server, Type: o.webhookType, Operations: o.operations}},
 				&webhook.AddAdmissionWebhookBuilderHandler{Resource: o.res, Config: webhook.Config{Server: o.server, Type: o.webhookType, Operations: o.operations}},
-				&webhook.Server{Resource: o.res, Config: webhook.Config{Server: o.server, Type: o.webhookType, Operations: o.operations}},
-				&webhook.AddServer{Resource: o.res, Config: webhook.Config{Server: o.server, Type: o.webhookType, Operations: o.operations}},
+				&webhook.Server{Config: webhook.Config{Server: o.server, Type: o.webhookType, Operations: o.operations}},
+				&webhook.AddServer{Config: webhook.Config{Server: o.server, Type: o.webhookType, Operations: o.operations}},
 			)
 			if err != nil {
 				log.Fatal(err)
--- a/pkg/scaffold/v1/webhook/webhook_test.go
+++ b/pkg/scaffold/v1/webhook/webhook_test.go
@@ -20,7 +20,6 @@ var _ = Describe("Webhook", func() {
 	}
 
 	serverName := "default"
-	domainName := "testproject.org"
 	inputs := []*webhookTestcase{
 		{
 			Resource: resource.Resource{Group: "crew", Version: "v1", Kind: "FirstMate", Namespaced: true, CreateExampleReconcileBody: true},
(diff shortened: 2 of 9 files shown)


pkg/scaffold/v1/webhook/webhook_test.go
```diff
@@ -20,7 +20,6 @@ var _ = Describe("Webhook", func() {
 	}
 
 	serverName := "default"
-	domainName := "testproject.org"
```
camilamacedo86: Why you remove the tests?

Adirio: That is checked in resource, this hard-coded string was used to fill `GroupDomain` which is one of the removed fields.

camilamacedo86: Did you find it has other place testing and ensuring that the webhooks after this change will still work on with domains as `"testproject.org"`? If yes, could you please provide the info in the PR? I mean add the link where it still been checked. If not, then is required to add the test. 

Also, did you check it manually?

Adirio: The test just filled that value with an string that was never used. You can try changing that string for a invalid domain and it will still work, as it was not being used for anything.

You can see the removal of that field in [here](https://github.com/kubernetes-sigs/kubebuilder/pull/1219/files/8c42e7ef0628965330882d9f02ec377ec1f732a8#diff-4b3b663bcd342da03284c5e58972d498) and [here](https://github.com/kubernetes-sigs/kubebuilder/pull/1219/files/8c42e7ef0628965330882d9f02ec377ec1f732a8#diff-ae6f681b954c192fb1fb62fdb59e7091). Doing a search on those files (not in the changes section of the PR but in the real file) shows that `GroupDomain` is only used to declare it and assign a value, but its value is never used in the scaffolded file.

cmd/webhook_v1.go
```diff
@@ -77,8 +77,8 @@ This command is only available for v1 scaffolding project.
 				&webhook.AdmissionWebhookBuilder{Resource: o.res, Config: webhook.Config{Server: o.server, Type: o.webhookType, Operations: o.operations}},
 				&webhook.AdmissionWebhooks{Resource: o.res, Config: webhook.Config{Server: o.server, Type: o.webhookType, Operations: o.operations}},
 				&webhook.AddAdmissionWebhookBuilderHandler{Resource: o.res, Config: webhook.Config{Server: o.server, Type: o.webhookType, Operations: o.operations}},
-				&webhook.Server{Resource: o.res, Config: webhook.Config{Server: o.server, Type: o.webhookType, Operations: o.operations}},
-				&webhook.AddServer{Resource: o.res, Config: webhook.Config{Server: o.server, Type: o.webhookType, Operations: o.operations}},
+				&webhook.Server{Config: webhook.Config{Server: o.server, Type: o.webhookType, Operations: o.operations}},
+				&webhook.AddServer{Config: webhook.Config{Server: o.server, Type: o.webhookType, Operations: o.operations}},
 			)
```
camilamacedo86: The v1 is deprecated then I am not sure if we should do refractories on it.

Adirio: I did the changes project-wide. Removing all fields that were not used. As I am not changing any functionality, I don't think v1 being deprecated should be a stopper.

Adirio:
Rebased (no changes)

DirectXMan12:
/lgtm
/approve

code cleanups to v1 are fine-ish.  I think this isn't a big deal.

pkg/scaffold/v2/controller_suitetest.go
```diff
@@ -130,15 +119,12 @@ var _ = AfterSuite(func() {
...
-	if a.Plural == "" {
-		a.Plural = flect.Pluralize(strings.ToLower(a.Resource.Kind))
-	}
+	resourcePackage, _ := util.GetResourceInfo(a.Resource, a.Repo, a.Domain)
 
 	ctrlImportCodeFragment := fmt.Sprintf(`"%s/controllers"
 `, a.Repo)
 	apiImportCodeFragment := fmt.Sprintf(`%s%s "%s/%s"
-`, a.Resource.GroupImportSafe, a.Resource.Version, a.ResourcePackage, a.Resource.Version)
+`, a.Resource.GroupImportSafe, a.Resource.Version, resourcePackage, a.Resource.Version)
```
camilamacedo86: The `a.ResourcePackage` is used in this case. 
You are replacing the attribute of the struct for an env var. 
IMO, the previous code is a better approach and follows the same impl of conyttroller.go as well.

Adirio: This has already been merged. We do not need that field in the struct. The struct represents a file object. The file does not require the `ResourcePackage` field for anything. When updating this file we do need it, so we compute it locally there. This was applied to all the files. Every field that is not needed for the file has been removed. `controller.go` does need it for the file (you can make a search for `.ResourcePackage` in the template to see where it is used.

PR #1221, opened 2019-11-25
Move files inside a new directory

Adirio:
This PR creates a directory where all the files have been moved to: `pkg/scaffold/files`.

- `pkg/scaffold/v1` -> `pkg/scaffold/files/v1`
- `pkg/scaffold/v2` -> `pkg/scaffold/files/v2`
- `pkg/scaffold/project` -> `pkg/scaffold/files/v1/project` and `pkg/scaffold/files/v2/project` with two exceptions: `project` and `boilerplate`. These two files are scaffolded in a different manner as they are version agnostic and thus have been placed at `pkg/scaffold/files`
- `Version1` and `Version2` strings have been moved to the base `pck/scaffold` package.

Additionally some import alias have been added for clarity.


Adirio:
/kind cleanup
This cleanup PR should also help with the deprecation of v1 as some files that were v1-only were not placed in the right place.

pkg/scaffold/api.go
```diff
@@ -153,10 +152,10 @@ func (api *API) scaffoldV1() error {
 			fmt.Sprintf("%s_controller_test.go", strings.ToLower(r.Kind))))
 
 		err := (&Scaffold{}).Execute(api.buildUniverse(), input.Options{},
-			&controller.Controller{Resource: r},
-			&controller.AddController{Resource: r},
-			&controller.Test{Resource: r},
-			&controller.SuiteTest{Resource: r},
+			&controllerv1.Controller{Resource: r},
```
camilamacedo86: it could be ctrlv1 and ctrlv2 instead of.

cmd/init_project.go
```diff
@@ -108,15 +108,15 @@ func (o *projectOptions) bindCmdlineFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.project.Repo, "repo", "", "name to use for go module, e.g. github.com/user/repo.  "+
 		"defaults to the go package of the current working directory.")
 	cmd.Flags().StringVar(&o.project.Domain, "domain", "my.domain", "domain for groups")
-	cmd.Flags().StringVar(&o.project.Version, "project-version", project.Version2, "project version")
+	cmd.Flags().StringVar(&o.project.Version, "project-version", scaffold.Version2, "project version")
```
camilamacedo86: I do not think that the we should change project for scaffold here. 
IHMO project is a better name since the commands are executed for/in a project and scaffold means the implementations/files scaffolded into the project by default which can be improved/changed by the users.

pkg/scaffold/files/project.go
```diff
@@ -14,21 +14,16 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package project
+package files
```
camilamacedo86: > pkg/scaffold/v1 -> pkg/scaffold/files/v1
> pkg/scaffold/v2 -> pkg/scaffold/files/v2

Why we should add more one dir here? What is the advantage which justifies this change?

testdata/project-v2/.gitignore
```diff
@@ -13,10 +13,6 @@ bin
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 
-# Kubernetes Generated files - skip generated files, except for vendored files
-
-!vendor/**/zz_generated.*
-
```
camilamacedo86: Why it was removed from the .gitignore ?

pkg/scaffold/api.go
Adirio: It could, but IMHO it is not used that many times to justify loosing the clarity the full name gives to type 6 letters less.

cmd/init_project.go
Adirio: They previously were inside `pkg/scaffold/project` (only the last part is seen in this comment). That package contained mostly `v1` files and some `v2` files. Thus they have been moved inside `pkg/scaffold/files/v1/project` and `pkg/scaffold/files/v2/project` directories. It doesn't make sense to keep `Version1` and `Version2` strings inside `pkg/scaffold/files/v1/project` as that is only for `v1` (same for `pkg/scaffold/files/v2/project`).

pkg/scaffold/files/project.go
Adirio: Kubebuilder has three kind of objects. In the first level we have the Cobra commands. The second layer is the scaffolders. The last one is the files. Using `files` as a directory allows the directory structure to represent the conceptual design of kubebuilder.

testdata/project-v2/.gitignore
Adirio: The `.gitignore` file was generated exactly the same for `v1` and `v2`. Now they are scaffolded separately. As `v1` uses GOPATH and vendor, that line is needed for `v1`. But `v2` used GOMODULES which doesn't have a `vendor` folder.

camilamacedo86: This change make sense, however, IHMO the extraction of any scaffold feature which has been shared between the versions should be done in a specific PR for each in order to make easier the review, track the changes and revert if be required.

pkg/scaffold/files/project.go
camilamacedo86: I'd suggest you do a specific PR with only the layout changes and make clear the motivations for your suggestion of changes. In fact, IHMO this kind of suggestion/change would be better addressed by raising an issue with and then do a PR only/if the maintainers accept and agree with.

Adirio:
> I'd suggest you first raise an issue with the suggestion and your justifications/argumentations for the changes in place since and then, when and if it is accepted by the maintainers a PR could be done.

This is part of the umbrella Issue I am working on. Will add it to the description at the top, I missed that part.

> Also, I'd recommend you do one PR with your suggestions to improve the clarity of the imports (change the import alias) and another with the changes in the layout (add new dirs, move files and etc ...).

Most of the import aliases were needed with the changes in the layout. Just a couple of them were missing, so I added them to homogenize them.

> PS.: note that this PR is not passing in the CI. If you are still working on it, please use the text `WIP` in the title.

There was a circular import I just removed. It was setting a default version for project files `GetInput` method. That method will never be executed without a version, but proving this right now would be a bit hard so instead of removing the whole default version I just removed the circular import. I plan to modify how files set their default values as the File interface is not very good.

Adirio:
Rebased (no changes)

Adirio:
Closed in favor of #1264

PR #1222, opened 2019-11-25
docs/book: small grammar/spelling tweaks

mattnworb:
First of all, thanks so much for kubebuilder, and especially for the book. I learned a lot of new things about building controllers from reading it, even after having built a few controllers (not using kubebuilder) in production. The book does a phenomenal job of explaining why certain things need to be done, and it is a really great idea to explain the code-level concepts in the actual code itself (via the literate Go snippets).

While reading through the book, I noticed a few small spelling mistakes and tweaks that I thought I'd send a PR to fix.

### DIFF
--- a/docs/book/src/cronjob-tutorial/api-design.md
+++ b/docs/book/src/cronjob-tutorial/api-design.md
@@ -17,7 +17,7 @@ machines.  You've probably noticed them when specifying resources requests
 and limits on pods in Kubernetes.
 
 They conceptually work similar to floating point numbers: they have
-a significand, base, and exponent.  Their serialize, human readable for
+a significand, base, and exponent.  Their serialized human readable format
 uses whole numbers and suffixes to specify values much the way we describe
 computer storage.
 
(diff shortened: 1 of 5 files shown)


docs/book/src/cronjob-tutorial/api-design.md
```diff
@@ -17,7 +17,7 @@ machines.  You've probably noticed them when specifying resources requests
 and limits on pods in Kubernetes.
 
 They conceptually work similar to floating point numbers: they have
-a significand, base, and exponent.  Their serialize, human readable for
+a significand, base, and exponent.  Their serialized human readable format
```
mengqiy: Maybe we should use `Their serializable and human readable format`

mattnworb: sure, I wasn't sure what the original intent for the wording here was.

mattnworb: @mengqiy updated

PR #1224, opened 2019-11-25
:bug: end users should not have permissions to edit */status

### LINKED ISSUE #886: Generate suggested Roles for CR users

mengqiy:
As @erictune pointed out in https://github.com/kubernetes-sigs/kubebuilder/pull/1165#discussion_r347689204, editor role is intended for end users which should not have permissions to edit `*/status`.

fixes #886

### DIFF
--- a/pkg/scaffold/v2/crd_editor_rbac.go
+++ b/pkg/scaffold/v2/crd_editor_rbac.go
@@ -50,7 +50,7 @@ func (g *CRDEditorRole) Validate() error {
 	return g.Resource.Validate()
 }
 
-const crdRoleEditorTemplate = `# permissions to do edit {{ .Resource.Resource }}.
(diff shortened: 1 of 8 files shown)


Adirio:
GitHub was a bit laggy and I think that caused the Travis fail, you should probably re-run Travis

PR #1247, opened 2019-12-09
community: fix discussion group links

pires:
There really isn't more to say about this.


camilamacedo86:
/ok-to-test

Really tks for the contribution 🥇

PR #1186, opened 2019-11-11
Unification of receiver names

Adirio:
Linter is reporting different receiver names (i.e., methods of same struct with different receiver names) as if they were generic. Additionally most of the files use generic receivers such as `a`. Tiis PR changes all receiver names for implementations fo the `File` interface to `f` as a mnemonic of `file`.

Closes: #1182

/kind cleanup

Issue #1150, opened 2019-11-04
Non-default value for kubebuilder init path argument


Adirio:
`kubebuilder init --path "hack2/boilerplate.go.txt"`

Error message may not be exact, as I translated it:
```
The system couldn't find the specified path.
make: *** [generate] Error 1
2019/11/04 17:11:05 exit status 2
```

Path argument can be provided to specify where to generate the boilerplate file. While the file is created, the Makefile and following `kubebuilder` commands expect it to be in the default path.

This issue could be solve by #1149.

/kind bug

Adirio:
This bug was solved in #1330

PR #1252, opened 2019-12-10
Named arguments for path formatting

Adirio:
### Description

This PR allows the usage of named-arguments while formatting files' paths.

Resources have a `Replacer` method that returns an `strings.Replacer` that is used to format paths. The `strings.Replacer` dictionary is as follows:
  - `%[group]s`: `Resource.Group`
  - `%[group-package-name]s`: `Resource.GroupPackageName`
  - `%[version]s`: `Resource.Version`
  - `%[kind]s`: `strings.ToLower(Resource.Kind)`
  - `%[plural]s`: `strings.ToLower(Resource.Plural)`

### Motivation

Despite the fact that the path can be provided to these file structs, access to some of the fields is restricted to the default path. This formatting ("%[kind]s" getting replaced by strings.ToLower(resource.Kind)) does not only apply to default paths. Paths provided outside the GetInputfunction do also benefit from it.

For example, if you want the types file to be inside a directory of its version, this can be done in the default package because you have access to the resource, but it can't be done before that. By providing this path temnplating options paths can be configured before having access to the exact resource. This allows to provide paths **before** having access to the resource like in a flag (for implementing a flag that is in kubebuilder` but breaks the scaffolding right now, the boilerplate path.

### DIFF
--- a/pkg/model/resource/options.go
+++ b/pkg/model/resource/options.go
@@ -168,9 +168,11 @@ func (opts *Options) safeImport(unsafe string) string {
 func (opts *Options) NewV1Resource(c *config.Config, doResource bool) *Resource {
 	res := opts.newResource()
 
+	replacer := res.Replacer()
+
 	// NOTE: while directories can have "-" and ".", v1 needs that directory to be a Go package
 	//       and that is the reason why we remove them for the directory
-	pkg := path.Join(c.Repo, "pkg", "apis", res.GroupPackageName, opts.Version)
+	pkg := replacer.Replace(path.Join(c.Repo, "pkg", "apis", "%[group-package-name]", "%[version]"))
 	domain := c.Domain
 
 	// pkg and domain may need to be changed in case we are referring to a builtin core resource:
@@ -180,11 +182,11 @@ func (opts *Options) NewV1Resource(c *config.Config, doResource bool) *Resource 
 	//  - In any other case, default to                          => project resource
 	// TODO: need to support '--resource-pkg-path' flag for specifying resourcePath
 	if !doResource {
-		file := filepath.Join("pkg", "apis", res.GroupPackageName, opts.Version,
-			fmt.Sprintf("%s_types.go", strings.ToLower(opts.Kind)))
+		file := replacer.Replace(filepath.Join(
+			"pkg", "apis", "%[group-package-name]", "%[version]", "%[kind]_types.go"))
 		if _, err := os.Stat(file); os.IsNotExist(err) {
 			if coreDomain, found := coreGroups[opts.Group]; found {
-				pkg = path.Join("k8s.io", "api", opts.Group, opts.Version)
+				pkg = replacer.Replace(path.Join("k8s.io", "api", "%[group]", "%[version]"))
 				domain = coreDomain
 			}
 		}
@@ -205,9 +207,11 @@ func (opts *Options) NewV1Resource(c *config.Config, doResource bool) *Resource 
 func (opts *Options) NewResource(c *config.Config, doResource bool) *Resource {
 	res := opts.newResource()
 
-	pkg := path.Join(c.Repo, "api", opts.Version)
+	replacer := res.Replacer()
+
+	pkg := replacer.Replace(path.Join(c.Repo, "api", "%[version]"))
 	if c.MultiGroup {
-		pkg = path.Join(c.Repo, "apis", res.Group, opts.Version)
+		pkg = replacer.Replace(path.Join(c.Repo, "apis", "%[group]", "%[version]"))
 	}
 	domain := c.Domain
 
@@ -220,7 +224,7 @@ func (opts *Options) NewResource(c *config.Config, doResource bool) *Resource {
 	if !doResource {
 		if !c.HasResource(opts.GVK()) {
 			if coreDomain, found := coreGroups[opts.Group]; found {
-				pkg = path.Join("k8s.io", "api", opts.Group, opts.Version)
+				pkg = replacer.Replace(path.Join("k8s.io", "api", "%[group]", "%[version]"))
 				domain = coreDomain
 			}
 		}
--- a/pkg/model/resource/resource.go
+++ b/pkg/model/resource/resource.go
@@ -17,6 +17,9 @@ limitations under the License.
 package resource
 
 import (
+	"fmt"
+	"strings"
+
 	"sigs.k8s.io/kubebuilder/pkg/model/config"
 )
 
@@ -61,4 +64,20 @@ func (r *Resource) GVK() config.GVK {
 		Version: r.Version,
 		Kind:    r.Kind,
 	}
+}
+
+func wrapKey(key string) string {
+	return fmt.Sprintf("%%[%s]", key)
+}
+
+func (r Resource) Replacer() *strings.Replacer {
+	var replacements []string
+
+	replacements = append(replacements, wrapKey("group"), r.Group)
+	replacements = append(replacements, wrapKey("group-package-name"), r.GroupPackageName)
+	replacements = append(replacements, wrapKey("version"), r.Version)
+	replacements = append(replacements, wrapKey("kind"), strings.ToLower(r.Kind))
+	replacements = append(replacements, wrapKey("plural"), strings.ToLower(r.Plural))
+
+	return strings.NewReplacer(replacements...)
 }
--- a/pkg/scaffold/api.go
+++ b/pkg/scaffold/api.go
@@ -18,8 +18,6 @@ package scaffold
 
 import (
 	"fmt"
-	"path/filepath"
-	"strings"
 
 	"sigs.k8s.io/kubebuilder/internal/config"
 	"sigs.k8s.io/kubebuilder/pkg/model"
@@ -90,11 +88,6 @@ func (s *apiScaffolder) newUniverse() *model.Universe {
 
 func (s *apiScaffolder) scaffoldV1() error {
 	if s.doResource {
-		fmt.Println(filepath.Join("pkg", "apis", s.resource.GroupPackageName, s.resource.Version,
-			fmt.Sprintf("%s_types.go", strings.ToLower(s.resource.Kind))))
-		fmt.Println(filepath.Join("pkg", "apis", s.resource.GroupPackageName, s.resource.Version,
-			fmt.Sprintf("%s_types_test.go", strings.ToLower(s.resource.Kind))))
-
 		if err := machinery.NewScaffold().Execute(
 			s.newUniverse(),
 			&crdv1.Register{},
@@ -117,11 +110,6 @@ func (s *apiScaffolder) scaffoldV1() error {
 	}
 
 	if s.doController {
-		fmt.Println(filepath.Join("pkg", "controller", strings.ToLower(s.resource.Kind),
-			fmt.Sprintf("%s_controller.go", strings.ToLower(s.resource.Kind))))
-		fmt.Println(filepath.Join("pkg", "controller", strings.ToLower(s.resource.Kind),
-			fmt.Sprintf("%s_controller_test.go", strings.ToLower(s.resource.Kind))))
-
 		if err := machinery.NewScaffold().Execute(
 			s.newUniverse(),
 			&controllerv1.Controller{},
@@ -143,14 +131,6 @@ func (s *apiScaffolder) scaffoldV2() error {
 			if err := s.config.Save(); err != nil {
 				return fmt.Errorf("error updating project file with resource information : %v", err)
 			}
-		}
-
-		if s.config.MultiGroup {
-			fmt.Println(filepath.Join("apis", s.resource.Group, s.resource.Version,
-				fmt.Sprintf("%s_types.go", strings.ToLower(s.resource.Kind))))
-		} else {
-			fmt.Println(filepath.Join("api", s.resource.Version,
-				fmt.Sprintf("%s_types.go", strings.ToLower(s.resource.Kind))))
 		}
 
 		if err := machinery.NewScaffold(s.plugins...).Execute(
@@ -183,14 +163,6 @@ func (s *apiScaffolder) scaffoldV2() error {
 	}
 
 	if s.doController {
-		if s.config.MultiGroup {
-			fmt.Println(filepath.Join("controllers", s.resource.Group,
-				fmt.Sprintf("%s_controller.go", strings.ToLower(s.resource.Kind))))
-		} else {
-			fmt.Println(filepath.Join("controllers",
-				fmt.Sprintf("%s_controller.go", strings.ToLower(s.resource.Kind))))
-		}
-
 		if err := machinery.NewScaffold(s.plugins...).Execute(
 			s.newUniverse(),
 			&controllerv2.SuiteTest{},
--- a/pkg/scaffold/internal/templates/v1/controller/add.go
+++ b/pkg/scaffold/internal/templates/v1/controller/add.go
@@ -17,9 +17,7 @@ limitations under the License.
 package controller
 
 import (
-	"fmt"
 	"path/filepath"
-	"strings"
 
 	"sigs.k8s.io/kubebuilder/pkg/model/file"
 )
@@ -37,9 +35,9 @@ type AddController struct {
 // SetTemplateDefaults implements input.Template
 func (f *AddController) SetTemplateDefaults() error {
 	if f.Path == "" {
-		f.Path = filepath.Join("pkg", "controller", fmt.Sprintf(
-			"add_%s.go", strings.ToLower(f.Resource.Kind)))
+		f.Path = filepath.Join("pkg", "controller", "add_%[kind].go")
 	}
+	f.Path = f.Resource.Replacer().Replace(f.Path)
 
 	f.TemplateBody = addControllerTemplate
 
--- a/pkg/scaffold/internal/templates/v1/controller/controller.go
+++ b/pkg/scaffold/internal/templates/v1/controller/controller.go
@@ -17,8 +17,8 @@ limitations under the License.
 package controller
 
 import (
+	"fmt"
 	"path/filepath"
-	"strings"
 
 	"sigs.k8s.io/kubebuilder/pkg/model/file"
 )
@@ -35,10 +35,10 @@ type Controller struct {
 // SetTemplateDefaults implements input.Template
 func (f *Controller) SetTemplateDefaults() error {
 	if f.Path == "" {
-		f.Path = filepath.Join("pkg", "controller",
-			strings.ToLower(f.Resource.Kind),
-			strings.ToLower(f.Resource.Kind)+"_controller.go")
+		f.Path = filepath.Join("pkg", "controller", "%[kind]", "%[kind]_controller.go")
 	}
+	f.Path = f.Resource.Replacer().Replace(f.Path)
+	fmt.Println(f.Path)
 
 	f.TemplateBody = controllerTemplate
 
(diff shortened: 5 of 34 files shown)


Adirio:
/assign @mengqiy

This is ready to review, tests passed before squashing so they should also pass now.

pkg/fmt/alias.go
```diff
@@ -0,0 +1,50 @@
...
+
+type Stringer interface {
+	fmt.Stringer
+}
+
+type GoStringer interface {
+	fmt.GoStringer
+}
+
+var (
```
mengqiy: Not sure how much value does these changes create. Just save typing `fmt.`?

pkg/scaffold/v1/controller/add.go
```diff
@@ -38,9 +37,9 @@ type AddController struct {
 // GetInput implements input.File
 func (a *AddController) GetInput() (input.Input, error) {
 	if a.Path == "" {
-		a.Path = filepath.Join("pkg", "controller", fmt.Sprintf(
-			"add_%s.go", strings.ToLower(a.Resource.Kind)))
+		a.Path = filepath.Join("pkg", "controller", "add_%[kind]s.go")
```
camilamacedo86: IHMO: -1 for this change. In POV it makes the manutence ability be lower and do not improve the performance and/or design either.

pkg/scaffold/internal/util/path.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"strings"
+
+	"github.com/gobuffalo/flect"
+	"sigs.k8s.io/kubebuilder/pkg/fmt"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/resource"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/webhook"
+)
+
+func Path(path string, parameters ...interface{}) string {
```
camilamacedo86: IHMO: -1 for this change. In POV it makes the manutence ability be lower and do not improve the performance and/or design either.

pkg/fmt/alias.go
Adirio: By this change you mean the whole PR? Exporting `fmt` funtions? Will answer both:

In case you refer to exporting `fmt`, if you need the new functions provided by this package, you import `"kubebuilder/pkg/fmt"` and it replaces your previous `"fmt"` import. This way you don't have to be providing import alias to packages. You would still need to write it `fmt.Xxxxxx()` but now you will also have `fmt.Printm` family.

In case you refer to the whole PR, this does add functionality. This formatting (`"%[kind]s" getting replaced by `strings.ToLower(resource.Kind)`) does not only apply to default paths. Paths provided outside the `GetInput` function do also benefit from it. This allows to provide paths **before** having access to the resource like in a flag (for implementing a flag that is in `kubebuilder` but breaks the scaffolding right now, the boilerplate path.

pkg/scaffold/internal/util/path.go
Adirio: @camilamacedo86 Please consider reading the previous comment, especifically the last paragraph, where I describe the functionality it adds. Should write it in the description too though.

pkg/fmt/alias.go
camilamacedo86: IHMO: 

We should not move forward with this change. My vote for this is -1  since it makes the maintenance ability be lower and do not improve the performance and/or design either. 

Following some reasons.

- The fmt is part of the golang lib and everybody knows how to use and I do see any problem with write `fmt.Xxxxxx()`  or read/understand `fmt.Xxxxxx()`
- Provide `"kubebuilder/pkg/fmt"` would allow people to import it and create a need to support a new fmt lib as well
- Replace the get/set done directly in the attribute (`r.Resource.Attribute`) for a map["entry"] not implicit yet as `"%[kind]s"` make harder understanding from where the information came and how it works.  
- It will increase unnecessarily the usage of resources. 
- Regards:
   
> In case you refer to the whole PR, this does add functionality. This formatting ("%[kind]s" getting replaced by strings.ToLower(resource.Kind))

It can be solved in other ways as for example: Add a new attribute to Resources struct such as e.g `LowerKind` as it is done for other similar scenarios.

Adirio: > * The fmt is part of the golang and everybody knows how to use and I do see any problem with write `fmt.Xxxxxx()`  or read/understand `fmt.Xxxxxx()`

As I said before you still use `fmt` the same way. You are not removing the `fmt.Xxxxx` syntax.
<details><summary>Example usage</summary>

Standard `fmt` package:
```
package main

import (
    "fmt"
)

func main() {
    fmt.Println("Hello world!")
}
```

Extended `kubeuilder/pkg/fmt` package
```
package main

import (
    "fmt"
    fmt2 "kubebuilder/pkg/fmt" // We added this import
)

func main() {
    fmt.Println("Hello world!")
    // Now we can also use fmt2.Printm(..., map[string]string{...})
}
```

Extended `kubeuilder/pkg/fmt` package that provides alias to standard `fmt`
```
package main

import (
    "kubebuilder/pkg/fmt" // Instead of adding an import we changed it
)

func main() {
    fmt.Println("Hello world!") // This line is exactly the same
    // Now we can also use fmt.Printm(..., map[string]string{...})
}
```

</details>

As you can see, providing the alias to the standard `fmt` package just makes usage exactly the same as if it was the standard package.

> * Provide `"kubebuilder/pkg/fmt"` would allow people to import it and create a need to support a new fmt lib as well

This can be solved by moving it to `kubebuilder/pkg/internal/fmt`, which will prevent external usage.

> * Replace the get/set done directly in the attribute (`r.Resource.Attribute`) for a `map["entry"]` not implicit yet as `"%[kind]s"` make harder understanding from where the information came and how it works.

But it allows us to define paths in flags, or other places where we can just write a string but we don't have the resource available. Basically it offers templating capabilities to paths the same way that files can template parts with the `{{ .Xxxxx }}` syntax. One example of a path that is suposed to be provided as a string is with the `kubebuilder init --flag ...` which actually doesn't work due to bug #1150.

> * It will increase unnecessarily the usage of resources.

You are saying that it will be less performant? This is not a performance PR, it is adding a functionality and the performance difference will be negligible. We are templating whole files, simple templating for path will not have a visible effect.

> > In case you refer to the whole PR, this does add functionality. This formatting ("%[kind]s" getting replaced by strings.ToLower(resource.Kind))
> 
> It can be solved in other ways as for example: Add a new attribute to Resources struct such as e.g `LowerKind` as it is done for other similar scenarios.

The problem is not the strings.ToLower part. The example still aplies to other replacements. Please re-read that sentence after I edit the comment to a example that doesn't include `strings.ToLower`. The point is not that it includes the lowercasing part, the point is that the templating is done to non-default paths too, where we may not have the resource variable yet, and then cannot use it to get the path.

Adirio:
Rebased and removed the extended package `fmt`.

Adirio:
Rebased, removed the extended package `fmt`, and used the added funtionality in `pkg/scaffold/api`.

Adirio:

mengqiy:
/cc @DirectXMan12 
Need your input. Can you please help review this PR?

Adirio:
Rebase to include multi-group implementation changes. This PR will conflict with #1186 but will rebase after it gets merged

Adirio:
Travis error seems unrelated, please take a look at #1303.

Adirio:
Rebased and fixed style

DirectXMan12:
@Adirio I think this was rendered out of date by one of your other PRs

I think this is reasonable at a glance -- we don't particularly need it now, but it'll make flexibility in the future easier if we want it.

I think @droot has some comments -- assigning to him

/assign @droot

Adirio:
Will rebase and update this after #1340 and related PRs

pkg/scaffold/internal/util/path.go
```diff
@@ -0,0 +1,75 @@
...
+func sprintm(format string, parameters map[string]string) string {
+	return namedParameterRegexp.ReplaceAllStringFunc(format, func(key string) string {
+		if value, found := parameters[key[2:len(key)-2]]; found {
+			return value
+		}
+		return key
+	})
+}
+
+func Path(path string, parameters ...interface{}) string {
```
droot: Looking at the invocation of `util.Path`, I was wondering if you considered this API:
```Go
func PathJoin(paths ...string, params interface{}) string
```

Where Path can internally call `filepath.Join(paths...)` followed by named param substitution. I think this can help with the code readability where `util.Path` is being called.

cmd/webhook_v1.go
```diff
@@ -31,7 +31,8 @@ import (
 	"sigs.k8s.io/kubebuilder/pkg/scaffold/project"
 	"sigs.k8s.io/kubebuilder/pkg/scaffold/resource"
 	"sigs.k8s.io/kubebuilder/pkg/scaffold/v1/manager"
-	"sigs.k8s.io/kubebuilder/pkg/scaffold/v1/webhook"
+	webhookv1 "sigs.k8s.io/kubebuilder/pkg/scaffold/v1/webhook"
```
droot: I think we have already deprecated project `v1` structure, now will be a good time to get rid of `v1`. That can simplify lot of code.

Adirio: I think that the code right now is pretty clean and doesn't "need" v1 to be removed, I would just keep the support as it is right now, without further fixes/features. But that is just my opinion.

pkg/scaffold/internal/util/path.go
Adirio: Thats sounds interesting. Can you provide arguments after a `...string`?

If that is not allowed we could do something like:

```go
func JoinPath(paths ...string) Pather {
    return Pather{filepath.Join(paths...)}
}

type Pather struct {
    path string
}

func (p Pather) Replace(params map[string]string) string {...}
```

It would be called like `PathJoin(paths...).Replace(params)`. We could also go for the other way round, creating a type that holds the `params` and provides a `Join` method: `Formatter(params).Join(paths...)` or the simplest option: `PathJoin(params, paths...)`.

cmd/webhook_v1.go
droot: I wasn't referring to code in this PR but overall moving away from v1 will reduce the maintenance burden.

pkg/scaffold/internal/util/path.go
droot: >Can you provide arguments after a ...string?

No, we can't :). 

I like all the suggestions you listed. Don't have strong opinion about any particular one. I looked at https://golang.org/pkg/strings/#Replacer for some inspiration and your suggestion for `Formatter(params).Join(paths...)` can be tweaked a bit where `resource` can implement method called `Replacer()` which returns a replacer implementation that can be invoked with `Replace()` or `Join` (in our case). So it will look like:

```Go
# assuming r points to API resource object and APIResource implements Replacer() method
r := APIResource{}

# join path version
r.Replacer().Join(paths....)

# normal string replace if we need in future.
r.Replacer().Replace(s)

```

Adirio: Okey, I do like the design. Adding a `Replacer` method to `pkg/model/resource.Resource` that provides those two methods, `Join(...string) string` and `Replace(string) string`.

The `%[KEY]s` syntax is okey for the replacement? We would also have to revisit which fields should be replaced, for example, should `kind` key refer to lowercase Kind or should we use `lower(kind)`, `lower-kind` or `kind-lower`? I will try to get this design implemented when I get some time and that way I will also build a list of which items we want to be replaced so that we can find which keys we think fit best.

cmd/webhook_v1.go
camilamacedo86: Hi @droot i am working on that https://github.com/kubernetes-sigs/kubebuilder/pull/1355

pkg/scaffold/internal/util/path.go
droot: %[KEY]s seems fine to me. I think we should start with lower-case by default (don't expose lower/upper directives yet) and only expose those if needed in the future because I doubt we need those.

cmd/webhook_v1.go
droot: > Hi @droot i am working on that #1355

Thanks for taking this up. I think we added some date for removing v1 in the deprecation warning. So let's ensure we respect that date :)

Adirio:
Retriggering Travis

pkg/scaffold/internal/util/path.go
Adirio: @droot I guessed that, if we were gonna follow `strings.Replacer` format, we may also use it directly. Let me know what you think about this implementation.

Adirio:
Rebased.

@droot Shall we merge this PR?

pkg/model/resource/resource.go
```diff
@@ -67,3 +70,20 @@ func (r *Resource) GVK() config.GVK {
 		Kind:    r.Kind,
 	}
 }
+
+func wrapKey(key string) string {
+	return fmt.Sprintf("%%[%s]s", key)
+}
+
+func (r Resource) Replacer() *strings.Replacer {
+	replacements := make([]string, 0, 10)
```
droot: Using a bounded slice is good for optimization when performance is an issue. Since performance isn't an issue this case, preferring maintainable code over performant code in this case, we should switch to unbounded slice (nil) in this case.

pkg/model/resource/resource.go
```diff
@@ -67,3 +70,20 @@ func (r *Resource) GVK() config.GVK {
...
+
+	replacements = append(replacements, wrapKey("group"), r.Group)
+	replacements = append(replacements, wrapKey("group-package-name"), r.GroupPackageName)
+	replacements = append(replacements, wrapKey("version"), r.Version)
+	replacements = append(replacements, wrapKey("kind"), strings.ToLower(r.Kind))
+	replacements = append(replacements, wrapKey("plural"), strings.ToLower(r.Plural))
+	// To add new replacements just append both the original and the new values and increase the capacity by 2
+
```
droot: This comment should not be needed if we change it to unbounded slice (refer to comment above).

pkg/model/resource/options.go
```diff
@@ -178,11 +180,11 @@ func (opts *Options) NewV1Resource(c *config.Config, doResource bool) *Resource
...
 	// TODO: need to support '--resource-pkg-path' flag for specifying resourcePath
 	if !doResource {
-		file := filepath.Join("pkg", "apis", res.GroupPackageName, opts.Version,
-			fmt.Sprintf("%s_types.go", strings.ToLower(opts.Kind)))
+		file := replacer.Replace(filepath.Join(
+			"pkg", "apis", "%[group-package-name]s", "%[version]s", "%[kind]s_types.go"))
 		if _, err := os.Stat(file); os.IsNotExist(err) {
 			if coreDomain, found := coreGroups[opts.Group]; found {
-				pkg = path.Join("k8s.io", "api", opts.Group, opts.Version)
+				pkg = replacer.Replace(path.Join("k8s.io", "api", "%[group]s", "%[version]s"))
```
droot: Re-reading the code with fresh eyes, I realized, we introduced many new concepts while working through the iterations:

- replacer (strings.Replacer)
- Keys that can be replaced (kind, package-name, group etc...) which now contributors will have to dig through resource.go's Replace function to understand what all is supported.

I am not sure if the benefits are worth introducing this complexity. I know I should have spotted these in my earlier reviews, but didn't, so sorry about that.

One thing I realized, one thing is missing though: quick abstraction to `get-the-types-path-dir` to store `..._types.go` and `get-the-controller-path-dir` to store `..._controller.go`.

For example, in this function, we need `get-the-types-path-dir` in three places, it will help to have a helper just to do that.

pkg/scaffold/api.go
```diff
@@ -85,11 +84,13 @@ func (s *apiScaffolder) newUniverse() *model.Universe {
...
+
 	if s.doResource {
-		fmt.Println(filepath.Join("pkg", "apis", s.resource.GroupPackageName, s.resource.Version,
-			fmt.Sprintf("%s_types.go", strings.ToLower(s.resource.Kind))))
-		fmt.Println(filepath.Join("pkg", "apis", s.resource.GroupPackageName, s.resource.Version,
-			fmt.Sprintf("%s_types_test.go", strings.ToLower(s.resource.Kind))))
+		fmt.Println(replacer.Replace(filepath.Join(
+			"pkg", "apis", "%[group-package-name]s", "%[version]s", "%[kind]s_types.go")))
+		fmt.Println(replacer.Replace(filepath.Join(
+			"pkg", "apis", "%[group-package-name]s", "%[version]s", "%[kind]s_types_test.go")))
```
droot: Here also, we need `get-path-to-types-dir`

pkg/scaffold/api.go
```diff
@@ -113,10 +114,8 @@ func (s *apiScaffolder) scaffoldV1() error {
 	}
 
 	if s.doController {
-		fmt.Println(filepath.Join("pkg", "controller", strings.ToLower(s.resource.Kind),
-			fmt.Sprintf("%s_controller.go", strings.ToLower(s.resource.Kind))))
-		fmt.Println(filepath.Join("pkg", "controller", strings.ToLower(s.resource.Kind),
-			fmt.Sprintf("%s_controller_test.go", strings.ToLower(s.resource.Kind))))
+		fmt.Println(replacer.Replace(filepath.Join("pkg", "controller", "%[kind]s", "%[kind]s_controller.go")))
+		fmt.Println(replacer.Replace(filepath.Join("pkg", "controller", "%[kind]s", "%[kind]s_controller_test.go")))
```
droot: Same here, we also need `get-path-to-controllers-dir`

pkg/scaffold/api.go
```diff
@@ -142,11 +143,9 @@ func (s *apiScaffolder) scaffoldV2() error {
 		}
 
 		if s.config.MultiGroup {
-			fmt.Println(filepath.Join("apis", s.resource.Group, s.resource.Version,
-				fmt.Sprintf("%s_types.go", strings.ToLower(s.resource.Kind))))
+			fmt.Println(replacer.Replace(filepath.Join("apis", "%[group]s", "%[version]s", "%[kind]s_types.go")))
 		} else {
-			fmt.Println(filepath.Join("api", s.resource.Version,
-				fmt.Sprintf("%s_types.go", strings.ToLower(s.resource.Kind))))
+			fmt.Println(replacer.Replace(filepath.Join("api", "%[version]s", "%[kind]s_types.go")))
```
droot: here we need `get-path-to-types-dir`

pkg/model/resource/options.go
Adirio: The helper issue has been solved, the complexity concern is being discussed.

pkg/model/resource/resource.go
```diff
@@ -62,3 +65,19 @@ func (r *Resource) GVK() config.GVK {
 		Kind:    r.Kind,
 	}
 }
+
+func wrapKey(key string) string {
+	return fmt.Sprintf("%%[%s]s", key)
+}
+
+func (r Resource) Replacer() *strings.Replacer {
+	replacements := make([]string, 0)
```
droot: `var replacements []string` ?

pkg/model/resource/resource.go
```diff
@@ -62,3 +65,19 @@ func (r *Resource) GVK() config.GVK {
 		Kind:    r.Kind,
 	}
 }
+
+func wrapKey(key string) string {
+	return fmt.Sprintf("%%[%s]s", key)
```
droot: we should get rid of the extra `s` from the syntax, it causing confusion.

pkg/model/resource/resource.go
Adirio: Linters prefer the `make` version, I don't know the exact reasoning

PR #1254, opened 2019-12-11
Resolve package and variable name collision

Adirio:
`resource` is a package name, so using the same for the variable is a collision:

https://github.com/kubernetes-sigs/kubebuilder/blob/3ec07ea772fe1f392f37347b50c354d3b578f4bb/pkg/scaffold/api.go#L29

/kind bug


Adirio:

PR #1255, opened 2019-12-11
Two-step resource creation

Adirio:
### Description

Create resource Options to split the two step resource creation into two different types: `resource.Options` and `resource.Resource`. The first is the one that commands will be aware of while the second is created from the first once it has been validated. This allows to split the validation and new fields computation logic into a more natural flow.

Cobra commands will only create a `resource.Options` object where the flags value will be inserted. Then, that values will be validated and a resource with additional fields will be created for the scaffolding process.

Some fields of the `Resource` object have been modified:

- `GroupImportSafe` was being used mainly with the version sufixed as an import alias for the file containing the CRD types. An `ImportAlias` field has been created which already includes the version. The only exception to this is in v1 where a package is created with the group name. A `GroupPackageName` has been kept for this purpose but can be removed after v1 deprecation period.
- `Resource` field was being used to store the `Plural` and an external string was also used in some fields for holding the same value, which meant that the plural computation was being made in several points in the code. They have been centralized in a `Plural` field for the `Resource`.
- `ShortNames` was not used at all so it has been removed.
- `Package` field has been added to replace the external `ResourcePackage` string that was being used in several file scaffolds. The computation of this field varies from v1 to v2 as the paths are different. Additionally, mostly all its uses were preprended by `/GROUP/VERSION` in the v1 case and `/VERSION` in the v2 case, so those fields have been added to the `Package` field to simplify file scaffolds.
- `Domain` field has been added to replace the external `GroupDomain`. Despite not being a path, the existence of a file needs to be checked, so its also computed different for v1 and v2. Additional use cases of these field have been found in other file scaffolds replacing `{{ .Resource.Group }}.{{ .Domain }}`.

All the values have been provided `json` tags so that it can be used in replacement of `model.Resouce` as it was already suggested in a `TODO`.

### Motivation

This PR is scoped under #1218.


Adirio:
### 1) Keeping both `Plural` and `Resource` fields or just keep one of them

I am not sure if two fields are required, it should be evaluated if any time we may need different values for those fields. In case we don't, only one should be kept. I would prefer `Plural` as it is more descriptive than a field called `Resource` of a object of type `Resource`.

### 2) Removing `model.Resource` in favor of `resource.Resource`

This was already suggested in a `TODO` comment, so now that it includes all the fields that it had, I added the json block comments. I also suggested a couple of json key renames. Would like to hear from the plugin system developers.

### 3) Checking for dash-prefixed values vs exact strings at resource validation.

This change was already suggested in a previous PR. @camilamacedo86 commented that she didn't like this change. I included it here with some modifications because I think it is needed to completely solve the bug that was (partially) fixed by #1163.

Checking for flags need to be done first, because if you check for 0-length strings at the same time you may get non-exact errors. For example, if you do a `kubectl create api --group --version v1 --kind Example` will error with `group cannot be empty` while `kubectl create api --version --group example --kind Example` will error with `group cannot be empty too` when `version cannot be empty is expected`. This is because `group` is checked first and as `--group` was inserted as the value to version, group is actually the zero-length string, and then will print that error before the version value is checked to find `--group` as a value. So it is clear that we first have to check all the fields for flags and then check all them for zero-length strings.

About using a generic approach `strings.HasPrefix(..., "-")` vs listing the exact flags, my approach will have more descriptive messages for cases like: `kubebuilder create api --group --namespaced --version v1 --kind Example` where we would expect `group cannot be empty` and instead, as `--namespaced` was inserted as a the group, it will fail in its validation printing `kind must be PascalCase (expected Namespaced was --namespace)`.

Adirio:
@DirectXMan12 @droot @pwittrock @Liujingfang1 @mengqiy @pmorie @camilamacedo86 @joelanford I would like to hear your opinion on the topics under discussion. Would be great if you could review the PR but focusing on the discussion topics will be less time demanding. Thank you all in advance.

pkg/scaffold/resource/options.go
```diff
@@ -0,0 +1,231 @@
...
+		"admissionregistration": "k8s.io",
+		"apiextensions":         "k8s.io",
+		"authentication":        "k8s.io",
+		"autoscaling":           "",
+		"batch":                 "",
+		"certificates":          "k8s.io",
+		"core":                  "",
+		"extensions":            "",
+		"metrics":               "k8s.io",
+		//"policy":                "",
```
Adirio: This commented line is conflictive. If uncommented, tests will fail. I will post my understanding about this, some help would be apreciated, feel free to contact me through kubebuilder slack channel.

The behavior (both before and after this PR) is to check if the provided resource had the expected type file. In case it doesn't, check if it is one of the builtin apis, and change the domain and package import path accordingly. This allows to have custom api groups with the same name than builting api groups. In case neither the file nor the builtin api group is found, it defaults to consider it local.

When the golden test is building the testdata folder, it seems to be unable to find this file, so it tries to look for a builtin group. The resources in the `crew` group do not find a builtin api group so they default back to local files. The resources in `core` group do not find the file and default to the builtin "core" api group (as expected). But the resources in the `policy` group do not find the file and thus look for a bultin api group, which they would find without this line commented, and thus consider the resources in this group part of the builtin "policy" api group when it shouldn't.

The thing is that checking this file was already being done, the behavior hasn't changed, just moved to this file.

When creating the testdata to check the expected files, the `os.Stat` call in line 149 is not finding the file. This check was already being done previously. In case you cant find the types definition file, you look for the builtin groups, but you prioritize the custom types in case you do find the mentioned file.

Adirio:
/hold

https://github.com/kubernetes-sigs/kubebuilder/pull/1255#discussion_r357567451 needs to be resolved before merging.

pkg/scaffold/resource/options.go
DirectXMan12: It's unclear to me why this was commented out in the first place, though.  Policy is definitely a core API group (FWIW, this is why we have the whole subdomain thing in core kubernetes now).

DirectXMan12: First, let me be supremely unhelpful: I don't like this logic in the first place.  We should have a syntax for explicitly specifying external API groups, and use it to disambiguate.

Ok, now for what to do about this: if the issue is actually that we're in testdata, we might need to `Chdir` into the testdata directory in the file first, or something.  testdata gets ignored by go when you're outside of it.  Let me try to reproduce.

Adirio: I strongly agree with that. Like a --builtin flag or something similar.

DirectXMan12: Aaaah.  Our scaffolding is just a mess (but you already knew that).  Before your refactor, for v1, we'd do the check and build the universe with `policy` treated as core.

Then, we'd *completely* ignore that, generate the API anyway, and then *re-run that check in a different method* when generating the controller.  The second time we run the check, API exists, so the file exists, so we treat policy as local.

With your change, AFAICT, we just run the check once, so we treat policy as core.

DirectXMan12: :woman_facepalming:

DirectXMan12: (your version is sane, FWIW)

As a follow-up: all of this "check if the file exists" logic should probably go away eventually in favor of:

- if the API is NON-fully qualified, treat as local (e.g. `policy`, `metrics`, etc)
- if the API is fully qualified, check the PROJECT file (e.g. `policy.` (note the trailing period), `metrics.k8s.io`):
  - if it exists in the PROJECT file, treat as local
  - if it doesn't exist in the PROJECT file, treat as external

DirectXMan12: For now, I think we just need to special-case v1 to *always* consider the file as existing when generating *both* API and *controller*.

Adirio: > As a follow-up: all of this "check if the file exists" logic should probably go away eventually in favor of:
> 
> * if the API is NON-fully qualified, treat as local (e.g. `policy`, `metrics`, etc)
> * if the API is fully qualified, check the PROJECT file (e.g. `policy.` (note the trailing period), `metrics.k8s.io`):
>   
>   * if it exists in the PROJECT file, treat as local
>   * if it doesn't exist in the PROJECT file, treat as external

What would you say that are the requirements to consider an API fully qualified? GVK is enforced, and domain is obtained from the PROJECT file. Should there be a optional domain flag for `kubebuilder create api ...` that defaults to the PROJECT domain? In this case, by using "k8s.io" as the API domain the scaffolder will know that is one of the core types.

Adirio: > Aaaah. Our scaffolding is just a mess (but you already knew that). Before your refactor, for v1, we'd do the check and build the universe with `policy` treated as core.
> 
> Then, we'd _completely_ ignore that, generate the API anyway, and then _re-run that check in a different method_ when generating the controller. The second time we run the check, API exists, so the file exists, so we treat policy as local.
> 
> With your change, AFAICT, we just run the check once, so we treat policy as core.

I recreated the resources after scaffolding the resource type files as a temporary solution. I did it for both v1 and v2 (policy was not giving any issues for v2 because it is commented in generate_golden.sh due to v2 being single-group only but v2 multi-group is coming soon).

pkg/scaffold/v1/crd/register.go
```diff
@@ -36,7 +36,7 @@ type Register struct {
 // GetInput implements input.File
 func (r *Register) GetInput() (input.Input, error) {
 	if r.Path == "" {
-		r.Path = filepath.Join("pkg", "apis", r.Resource.Group, r.Resource.Version, "register.go")
+		r.Path = filepath.Join("pkg", "apis", r.Resource.GroupPackageName, r.Resource.Version, "register.go")
```
camilamacedo86: I liked 👍 the rename

cmd/webhook_v1.go
```diff
@@ -61,20 +61,21 @@ This command is only available for v1 scaffolding project.
 				os.Exit(1)
 			}
 
-			if err := o.res.Validate(); err != nil {
-				log.Fatal(err)
+			if err := o.resOpts.Validate(); err != nil {
+				log.Fatalf("non-valid resource provided: %v", err)
```
camilamacedo86:
```suggestion
				log.Fatalf("invalid resource information provided: %v", err)
```
WDYT?

cmd/webhook_v2.go
```diff
@@ -60,25 +60,27 @@ func newWebhookV2Cmd() *cobra.Command {
 				os.Exit(1)
 			}
 
+			if err := o.resOpts.Validate(); err != nil {
```
camilamacedo86: 👍 I liked the approach to move the Validate() to be done before all.

cmd/webhook_v2.go
```diff
@@ -60,25 +60,27 @@ func newWebhookV2Cmd() *cobra.Command {
...
+				log.Fatalf("non-valid resource provided: %v", err)
+			}
+			res := o.resOpts.NewResource(projectInfo.Repo, projectInfo.Domain)
```
camilamacedo86: So, is note the resourceOptionals creating the model? `o.resOpts.NewResource`
Then, I think that would make more sense to keep the resource model as attribute of resource optional instead of. 
Also, why pass the `projectInfo.Repo, projectInfo.Domain`?. IMO: It is info that should be in the resOptionals.

cmd/webhook_v2.go
```diff
@@ -60,25 +60,27 @@ func newWebhookV2Cmd() *cobra.Command {
...
+			fmt.Println(filepath.Join("api", res.Version,
+				fmt.Sprintf("%s_webhook.go", strings.ToLower(res.Kind))))
+
 			if o.conversion {
 				fmt.Println(`Webhook server has been set up for you.
 You need to implement the conversion.Hub and conversion.Convertible interfaces for your CRD types.`)
 			}
 			webhookScaffolder := &webhook.Webhook{
-				Resource:   o.res,
+				Resource:   res,
```
camilamacedo86: I would keep `o.resource` here

cmd/webhook_v2.go
```diff
@@ -120,7 +122,7 @@ You need to implement the conversion.Hub and conversion.Convertible interfaces f
 
 // webhookOptions represents commandline options for scaffolding a webhook.
 type webhookV2Options struct {
-	res        *resource.Resource
+	resOpts    *resource.Options
 	defaulting bool
```
camilamacedo86: Would not make more sense it has the model instead of Options?

Adirio: What do you mean?
1. Renaming `resource.Options` to `resource.Model`?
2. Using `pkg/model.Resource`?

Adirio:
This PR needs #1288 to get merged. The last force-push is just squashing all the commits and removing the fix merge, but the previous [Travis build](https://travis-ci.org/kubernetes-sigs/kubebuilder/builds/634629234?utm_source=github_status&utm_medium=notification), which had the fix merged, did pass.

Currently we are looking for the types file to determine if a resource is local or external. This behavior is not-reliable, and adding a flag or modifying the resource of some existing ones would be a better option. However, this could be backwards incompatible. Therefore, the current behavior has been kept in this PR. The two resource fields that depend on this information have been hardcoded to local versions **during the resource scaffolding** (only in the case we do scaffold the resource). If we are scaffolding the resource, it is obviously local, so hardcoding it in this section has no side-effects. Afterwards, the resource is re-generated to remove hardcoded values and ensure that it detects the just-generated file.

Adirio:
This PR needs #1288 to get merged. The last force-push is just squashing all the commits and removing the fix merge, but the previous [Travis build](https://travis-ci.org/kubernetes-sigs/kubebuilder/builds/634629234?utm_source=github_status&utm_medium=notification), which had the fix merged, did pass.

Currently we are looking for the types file to determine if a resource is local or external. This behavior is not-reliable, and adding a flag or modifying the resource of some existing ones would be a better option. However, this could be backwards incompatible. Therefore, the current behavior has been kept in this PR. The two resource fields that depend on this information have been hardcoded to local versions **during the resource scaffolding** (only in the case we do scaffold the resource). If we are scaffolding the resource, it is obviously local, so hardcoding it in this section has no side-effects. Afterwards, the resource is re-generated to remove hardcoded values and ensure that it detects the just-generated file.

/hold cancel

@DirectXMan12 What do you think of this solution?

cmd/webhook_v2.go
camilamacedo86: But would make more sense the webhookV2Options received the flags or the model?

> // webhookOptions represents commandline options for scaffolding a webhook.

Yep, I agree the Options shows more appropriated to be here.

pkg/scaffold/resource/options.go
```diff
@@ -0,0 +1,234 @@
...
+	versionRequired = "version cannot be empty"
+	kindRequired = "kind cannot be empty"
+)
+
+var versionRegex = regexp.MustCompile(versionPattern)
+
+// Options contains the information required to build a new Resource
+type Options struct {
+	// Namespaced is true if the resource is namespaced.
+	Namespaced bool
```
camilamacedo86: add the bool at the end in order to use better the memory space. 
If you run the `maligned` you will check the error.

pkg/scaffold/resource/options.go
```diff
@@ -0,0 +1,234 @@
...
+		return fmt.Errorf(
+			"version must match %s (was %s)", versionPattern, opts.Version)
+	}
+
+	// Check if the Kind is PascalCase
+	if opts.Kind != flect.Pascalize(opts.Kind) {
+		return fmt.Errorf("kind must be PascalCase (expected %s was %s)", flect.Pascalize(opts.Kind), opts.Kind)
+	}
+
+	// TODO: validate plural and resource strings if provided
```
camilamacedo86: I think it can be removed. 
Why would like to check it? What is your motivation?

pkg/scaffold/resource/options.go
```diff
@@ -0,0 +1,234 @@
...
+
+	return nil
+}
+
+// safeImport returns a cleaned version of the provided string that can be used for imports
+func (opts *Options) safeImport(unsafe string) string {
```
camilamacedo86: Great 👍

pkg/scaffold/resource/options.go
```diff
@@ -0,0 +1,234 @@
...
+		"metrics":               "k8s.io",
+		"policy":                "",
+		"rbac.authorization":    "k8s.io",
+		"storage":               "k8s.io",
+	}
+
+	pkg := path.Join(repo, "pkg", "apis", res.GroupPackageName, opts.Version)
+	// TODO: remove the dependency on the created file to differentiate local and external resources
+	// as this requires the Resource to be recreated after scaffolding the file
+	file := filepath.Join("pkg", "apis", res.GroupPackageName, opts.Version, fmt.Sprintf("%s_types.go", strings.ToLower(opts.Kind)))
```
camilamacedo86: I need to use the scroll to check that could you please split the line?

pkg/scaffold/resource/options.go
```diff
@@ -0,0 +1,234 @@
...
+	res := opts.newResource()
+
+	// Use the k8s.io/api package for core resources
+	coreGroups := map[string]string{
+		"admission":             "k8s.io",
+		"admissionregistration": "k8s.io",
+		"apps":                  "",
+		"auditregistration":     "k8s.io",
+		"apiextensions":         "k8s.io", // NOTE: where is this?
+		"authentication":        "k8s.io",
```
camilamacedo86: Remove the comments? 
Why should we add them in the code?

pkg/scaffold/resource/resource.go
```diff
@@ -16,144 +16,42 @@ limitations under the License.
...
-
-	"github.com/gobuffalo/flect"
-)
-
 // Resource contains the information required to scaffold files for a resource.
 type Resource struct {
-	// Namespaced is true if the resource is namespaced
-	Namespaced bool
+	// Namespaced is true if the resource is namespaced.
+	Namespaced bool `json:"namespaces,omitempty"`
```
camilamacedo86: Could it be empty? If not, then : 

```suggestion
	Namespaced bool `json:"namespaces"`
```

pkg/scaffold/resource/options.go
Adirio: It is a bit outdated, yes, but plural may still require some validation. Remember that there is a `--plural` flag that you can set to use a custom plural so we may need to do some validations (like for example only allowing lowercased letters).

pkg/scaffold/resource/resource.go
```diff
@@ -16,144 +16,42 @@ limitations under the License.
...
-	// Group is the API Group.  Does not contain the domain.
-	Group string
+	// Group is the API Group. Does not contain the domain.
+	Group string `json:"group,omitempty"`
 
-	// GroupImportSafe is the API Group.  Does not contain the domain and it the "-"
-	// It is used to do safe imports.
-	GroupImportSafe string
+	// GroupPackageName is the API Group cleaned to be used as the package name.
+	GroupPackageName string `json:"-"`
```
camilamacedo86: `json:"-"`? Why ?

pkg/scaffold/resource/resource.go
```diff
@@ -16,144 +16,42 @@ limitations under the License.
...
 
-	// Version is the API version - e.g. v1beta1
-	Version string
+	// Version is the API version.
```
camilamacedo86:
```suggestion
	// Version is the API version. - e.g. v1beta1
```

pkg/scaffold/resource/resource_test.go
```diff
@@ -26,123 +28,158 @@ import (
...
+			Expect(resource.Group).To(Equal(options.Group))
+			Expect(resource.GroupPackageName).To(Equal("crew"))
+			Expect(resource.Version).To(Equal(options.Version))
+			Expect(resource.Kind).To(Equal(options.Kind))
+			Expect(resource.Plural).To(Equal("firstmates"))
+			Expect(resource.ImportAlias).To(Equal("crewv1"))
+			// Note: The following two tests will work because, despite not
+			// finding the file, "crew" is not of of the core groups.
+			Expect(resource.Package).To(Equal(path.Join("test", "pkg", "apis", "crew", "v1")))
+			Expect(resource.Domain).To(Equal("crew.test.io"))
```
camilamacedo86: IHMO GroupDomain make more intuitive its understanding but it is a nit.

pkg/scaffold/v1/controller/controller.go
```diff
@@ -133,7 +79,7 @@ var log = logf.Log.WithName("{{ lower .Resource.Kind }}-controller")
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
-	{{ .Resource.Group}}{{ .Resource.Version }} "{{ .ResourcePackage }}/{{ .Resource.Group}}/{{ .Resource.Version }}"
+	{{ .Resource.ImportAlias }} "{{ .Resource.Package }}"
```
camilamacedo86: I liked a lot 👍

pkg/scaffold/v2/types.go
```diff
@@ -38,7 +38,7 @@ type Types struct {
 // GetInput implements input.File
 func (t *Types) GetInput() (input.Input, error) {
 	if t.Path == "" {
-		t.Path = filepath.Join("pkg", "apis", t.Resource.Group, t.Resource.Version,
+		t.Path = filepath.Join("api", t.Resource.Version,
```
camilamacedo86: it does not show correctly.

pkg/scaffold/resource/resource.go
Adirio: False values will be considered empty and thus not passed to the JSON or YAML files, which makes if a little bit more memory efficient. If default values are true this doesn't work

pkg/scaffold/resource/options.go
Adirio: Oh, I didn't mean to get that committed. I just wanted to ask about those. I guess this is as good place as any other to ask about it. @mengqiy @DirectXMan12 

I was looking if these map required and updated. I checked in https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/api and I found a different set of apis. Is that the right place to look at?

The found differences are:
- `apiextensions` and `metric` are not subpackages of `k8s.io/api`
- `rbac.authorization` is not a subpackage either, but `rbac` is
- `setting` is not a subpackage either, but `settings` is
- There were two additional subpackages, `discovery` and `flowcontrol`
Should these reflect into changes to the map somehow?

pkg/scaffold/resource/resource.go
Adirio: `GroupPackageName` and `ImportAlias` are used for scaffolding purposes but can be easily obtained from the rest of the fields. Serializing them to json/yaml doesn't make sense, they would be duplicated information.

pkg/scaffold/resource/resource.go
Adirio: I dont think this provides any value. As a kubebuilder user you will never see a code comment and as a developer you have some lines before the `versionPattern` which provides the possible value. Do you really think it helps?

pkg/scaffold/resource/resource_test.go
Adirio: It is both the Resource and the Group domain, and as we are already in the `Resource` struct i left it as `Domain`. Let's see if others agree with you and I'll change it if so.

pkg/scaffold/v2/types.go
Adirio: This is `v2` where the path was being changed manually in `pkg/scaffold/api` to this value. `v1` keeps his original value.

Adirio:
@camilamacedo86 Thanks for the review. I answered or applied all the comments you did.

About the size of this refactor, you are right, it is pretty big. I will not introduce more changes in this but the ones applied here were pretty bound to eachother and was pretty hard to split. Take into account that splitting into `resource.Options` and `resource.Resource` required to change the `cmd` package to start using `resource.Options`, the `scaffold` package to convert those `resource.Options` into `resource.Resource`, the whole related tests, etc.

Adirio:
Major rebase to include multi-group for v2 changes. Including extending the `resource.Resource` tests that I had before (`v1` and `v2`) adding the `v2-multigroup` case.

Adirio:
Okey now it passes the checks.

One difference between `v1` and `v2-multigroup` is that in the first case the `pkg/apis/GROUP` directory has files belonging to the `GROUP` package so the clean version (wihout `-` and `.`) is used for the directory name in order to keep package and directory name equal. However, in `v2-multigroup`,  no file belongs directly to the `apis/GROUP` directory and `controllers` is used as the package name for controllers, so no nead to clean the directories.

camilamacedo86:
@Adirio it needs a rebase.

Adirio:
@camilamacedo86

Rebase done

Adirio:
@DirectXMan12 @mengqiy @camilamacedo86 I would love to have some feedback on this PR as it is a pivotal part of the scaffold enhancement I'm doing.

mengqiy:
> Create resource Options to split the two step resource creation into two different types: resource.Options and resource.Resource. The first is the one that commands will be aware of while the second is created from the first once it has been validated. This allows to split the validation and new fields computation logic into a more natural flow.

I feel this change can help to make our code cleaner.
We can drop and cleanup our v1 code after Feb 1. We can start to prepare the PR now.
Then you can rebase this PR on top of that. This PR will become easier to review.

Adirio:
> I feel this change can help to make our code cleaner.
> We can drop and cleanup our v1 code after Feb 1. We can start to prepare the PR now.
> Then you can rebase this PR on top of that. This PR will become easier to review.

Rebased to the current state. I tried to keep v1 related stuff clearly differentiated so that we can either remove it easily or just leave it there but knowing clearly that it is `v1` and not being `v1` and `v2` code all mangled up.

mengqiy:
Can you rebase on top of #1330? I will do one pass after that

Adirio:
> Can you rebase on top of #1330? I will do one pass after that

@mengqiy Done, hope it passes the rest of the tests. The temporary fix mentioned is discussed in this PR in one of the outdated or resolved conversations, if you want some more context.

pkg/model/resource/options.go
```diff
@@ -0,0 +1,259 @@
...
+	if !doResource {
+		file := filepath.Join("pkg", "apis", res.GroupPackageName, opts.Version,
+			fmt.Sprintf("%s_types.go", strings.ToLower(opts.Kind)))
+		if _, err := os.Stat(file); os.IsNotExist(err) {
+			if coreDomain, found := v1CoreGroups[opts.Group]; found {
+				pkg = path.Join("k8s.io", "api", opts.Group, opts.Version)
+				domain = coreDomain
+			}
+		}
+	}
```
Adirio: > Before your refactor, for v1, we'd do the check and build the universe with `policy` treated as core.
> 
> Then, we'd _completely_ ignore that, generate the API anyway, and then _re-run that check in a different method_ when generating the controller. The second time we run the check, API exists, so the file exists, so we treat policy as local.

@DirectXMan12 what do you think about this logic for v1? Basically it enforces it to be a project-local resource if `doResource` flag is set (after asking the user in case he didnt specify it) and defaults to looking for the file instead.

pkg/model/resource/options.go
```diff
@@ -0,0 +1,259 @@
...
+	domain := c.Domain
+
+	// pkg and domain may need to be changed in case we are referring to a builtin core resource:
+	//  - Check if we are scaffolding the resource now           => project resource
+	//  - Check if we already scaffolded the resource            => project resource
+	//  - Check if the resource group is a well-known core group => builtin core resource
+	//  - In any other case, default to                          => project resource
+	// TODO: need to support '--resource-pkg-path' flag for specifying resourcePath
+	if !doResource {
+		if !c.HasResource(opts.GVK()) {
```
Adirio: V2 has an optimization, using the tracked resources instead of looking for the file if `doController` wasn't set. Unfortunately V1 doesn't track resources in the config file.

pkg/model/resource/options.go
```diff
@@ -0,0 +1,259 @@
...
+
+	groupRequired   = "group cannot be empty"
+	versionRequired = "version cannot be empty"
+	kindRequired    = "kind cannot be empty"
+)
+
+var (
+	versionRegex = regexp.MustCompile(versionPattern)
+
+	v1CoreGroups = map[string]string{
```
mengqiy: I believe they can share the same map.

pkg/model/resource/options.go
```diff
@@ -0,0 +1,259 @@
...
+	// Namespaced is true if the resource is namespaced.
+	Namespaced bool
+
+	// CreateExampleReconcileBody will create a Deployment in the Reconcile example.
+	// v1 only
+	CreateExampleReconcileBody bool
+}
+
+func (opts *Options) Validate() error {
+	// Check that the required flags did not get a flag as their value
```
mengqiy: I remember that we tried to work around this issue by checking if it starts with `-` last year.
I wonder if the issue has been fix by upstream Cobra. If so, we can just bump the Cobra version.

pkg/model/resource/options_test.go
```diff
@@ -0,0 +1,96 @@
+package resource_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	. "sigs.k8s.io/kubebuilder/pkg/model/resource"
```
mengqiy: We normally don't use dot import for packages besides ginkgo and gomega.

pkg/model/resource/resource_test.go
```diff
@@ -0,0 +1,349 @@
...
+package resource_test
+
+import (
+	"path"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"sigs.k8s.io/kubebuilder/pkg/model/config"
+	. "sigs.k8s.io/kubebuilder/pkg/model/resource"
```
mengqiy: Same here

pkg/model/resource/options.go
Adirio: Cobra stated that this is normal unix-like flag parsing and is not a bug, so not luck in that sense.

pkg/model/resource/options_test.go
Adirio: Self imports are also used with dot imports. It just needs to import its own package because it belongs to the secial resource_test package to avoid circular imports.

pkg/model/resource/resource_test.go
Adirio: Exactly the same as above.

pkg/model/resource/options_test.go
mengqiy: This seems to be fine according to https://www.agiratech.com/12-best-golang-agile-practices-we-must-follow/#Import_Dot

PR #1265, opened 2019-12-16
Split injecting fields and validating files

Adirio:
### Description

The method `setFieldsAndValidate` was doing two things, as the name suggests. The logic has been split into two separate methods: `setFields` and `validate`. The `Validate` interface has been modified so that it is only implemented by files, as other structs as resources or some scaffolders also use the `Validate() error` signature.

### Motivation

/kind cleanup

### DIFF
--- a/pkg/scaffold/input/input.go
+++ b/pkg/scaffold/input/input.go
@@ -158,8 +158,9 @@ type File interface {
 	GetInput() (Input, error)
 }
 
-// Validate validates input
-type Validate interface {
+// RequiresValidation is a file that requires validation
+type RequiresValidation interface {
+	File
 	// Validate returns true if the template has valid values
 	Validate() error
 }
--- a/pkg/scaffold/scaffold.go
+++ b/pkg/scaffold/scaffold.go
@@ -75,7 +75,7 @@ type Plugin interface {
 	Pipe(u *model.Universe) error
 }
 
-func (s *Scaffold) setFieldsAndValidate(t input.File) error {
+func (s *Scaffold) setFields(t input.File) {
 	// Set boilerplate on templates
 	if b, ok := t.(input.BoilerplatePath); ok {
 		b.SetBoilerplatePath(s.BoilerplatePath)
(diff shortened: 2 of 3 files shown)


Adirio:
Retriggering Travis

Adirio:
/assign @DirectXMan12 

This is ready to be merged, should be pretty easy to review.

Adirio:

Adirio:
@DirectXMan12 can we merge this, it is a pretty small change that splits a function that was handling two different concerns.

pkg/scaffold/input/input.go
```diff
@@ -144,8 +144,9 @@ type File interface {
 	GetInput() (Input, error)
 }
 
-// Validate validates input
-type Validate interface {
+// RequiresValidation is a file that requires validation
+type RequiresValidation interface {
```
camilamacedo86: I agreed with you. It is hurting concepts as a single responsibility and cohesion. 
However, why change the name of the interface? Could we keep it as it was?

```suggestion
type Validation interface {
```

Adirio: The interface as you can see is used to check if a `File` requires to be validated. Look how the test looks like with this name:

```golang
func (_ *Scaffold) validate(file input.File) error {
	if reqValFile, ok := file.(input.RequiresValidation); ok {
		return reqValFile.Validate()
	}

	return nil
}
```

I think that is really readable, without knowing what the interface does, you can understand that function pretty easy. That was the reason behind the new name.

camilamacedo86: IHMO if I call a `Validation()` it is implicit that the code `RequiresValidation`.
So, I'd prefer the previous name in order to avoid to be prolix, I mean unnecessary verbose and/or redundant.  However, I leave it to the maintainers and you decided.

mengqiy: IMO using `RequiresValidation` is slightly better than the old name `Validate`.

mengqiy:
Please rebase

Adirio:
@mengqiy Done

mengqiy:
It seems a linter is not happy in Travis. Please fix if related. Otherwise, maybe retry it.

Adirio:
> It seems a linter is not happy in Travis. Please fix if related. Otherwise, maybe retry it.

I dont think it has anything to do, will retry trigger Travis or debug it tomorrow

Adirio:
Travis error seems unrelated, please take a look at #1303.

camilamacedo86:
@Adirio, 

The error here in the CI is:

Screenshot 2020-01-12 at 12 11 41

**See that this error can be fixed automatically. To do that run the lint locally with the flag `--fix`.**

c/c @mengqiy

Adirio:
This PR has been approved and every test passes now. Can someone add again the `lgtm` label as it was removed in a rebase?

PR #1266, opened 2019-12-16
Validate resources for webhooks and tests

Adirio:
### Description

Webhooks and some tests were using resources without validation. This may end up in some unexpected bugs in the future

### Motivation

/kind cleanup

### DIFF
--- a/cmd/webhook_v1.go
+++ b/cmd/webhook_v1.go
@@ -21,9 +21,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
-	"strings"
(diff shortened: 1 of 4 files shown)


Adirio:
/assign @DirectXMan12

This is ready to be merged, should be pretty easy to review.

mengqiy:
@Adirio It seems a test is failing, please fix.

Adirio:
@mengqiy I think that the issue with that test is due to the year change. #1282 should fix it.

PR #1273, opened 2019-12-19
🏃 Bump k8s tools to 1.16.4

vincepri:
Changes:

Use Kubernetes 1.16.4 binaries in tests
Use Kubernetes 1.16.4 in release bundle

PR #1284, opened 2020-01-03
🏃 Bump k8s tools to 1.16.4

vincepri:
Changes:

Use Kubernetes 1.16.4 binaries in tests
Use Kubernetes 1.16.4 in release bundle

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
   - os: osx
     env: KIND_K8S_VERSION="v1.15.3"
   - os: osx
-    env: KIND_K8S_VERSION="v1.14.1" 
+    env: KIND_K8S_VERSION="v1.14.1"
   - os: linux
     env: GOPROXY=https://proxy.golang.org/
 
--- a/build/cloudbuild.yaml
+++ b/build/cloudbuild.yaml
@@ -25,18 +25,18 @@ steps:
   args: [fetch, --tags, --depth=100]
 - name: "ubuntu"
   args: ["mkdir", "-p", "/workspace/_output/linux_amd64"]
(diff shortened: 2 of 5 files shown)


mengqiy:
@vincepri I'd expect `1.16.4` which is the same as #1273

vincepri:
@mengqiy Should be good to go!

.travis.yml
```diff
@@ -14,19 +14,19 @@ go:
 
 # A build matix defines the K8s versions to use for e2e tests. Travis runs these in parrellel
 env:
-  - KIND_K8S_VERSION="v1.16.2"
+  - KIND_K8S_VERSION="v1.16.4"
```
mengqiy: [`kind`](https://kind.sigs.k8s.io/) doesn't all of the k8s versions.
Looking at https://hub.docker.com/r/kindest/node/tags, the latest supported 1.16 is 1.16.3.
You can either change it to 1.16.3 or leave it as is.

.travis.yml
```diff
@@ -14,19 +14,19 @@ go:
...
   - KIND_K8S_VERSION="v1.15.3"
   - KIND_K8S_VERSION="v1.14.1"
   - GOPROXY=https://proxy.golang.org/
 # As OSX isn't running KIND e2e tests exlude it from the matrix
 matrix:
   exclude:
   - os: osx
-    env: KIND_K8S_VERSION="v1.16.2"
+    env: KIND_K8S_VERSION="v1.16.4"
```
mengqiy: Same here.

.travis.yml
vincepri: Fixed!

.travis.yml
vincepri: Fixed!

PR #1285, opened 2020-01-05
remove ignoreNotFound code and rely on controller-runtime

gregoryguillou:
`sigs.k8s.io/controller-runtime/pkg/client` includes the code for ignoreNotFound for a few months, see
[IgnoreNotFound](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/interfaces.go#L130)

This PR proposes to remove the code from kubebuilder. That is because the client package has to be part of the dependencies, it simplifies the work and it remove another dependency.


DirectXMan12:
/lgtm
/approve

thanks, been meaning to tackle this for a while now

PR #1288, opened 2020-01-08
Hotfix for TRAVIS error introduced by #1016

Adirio:
Fixes the Travis error introduced by #1016

The result of `{{ printf "%q" .BoilerplatePath }}` is `"hack/boilerplate.txt"` while Travis is expecting `./hack/boilerplate.txt`. Either the `testdata` should be change to reflect #1016 or `./{{ .Boilerplate }}` can be used instead. Both work as expected.

PR #1287, opened 2020-01-07
helpfully show progress of kubebuilder download

drnic:
It isn't obvious that the download is 60+Mb and may take some time on slow internet. Progress bars FTW.

```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    69  100    69    0     0    315      0 --:--:-- --:--:-- --:--:--   315
100   125  100   125    0     0    480      0 --:--:-- --:--:-- --:--:--   480
100   628    0   628    0     0    853      0 --:--:-- --:--:-- --:--:--  613k
100 60.6M  100 60.6M    0     0   563k      0  0:01:50  0:01:50 --:--:--  955k
```

Issue #909, opened 2019-08-02
make generate cannot pin the  version of the controller-gen


jiachengxu:
In the current Makefile generated by kubebuilder:
```
controller-gen:
ifeq (, $(shell which controller-gen))
    go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.0-beta.4
CONTROLLER_GEN=$(GOBIN)/controller-gen
else
CONTROLLER_GEN=$(shell which controller-gen)
endif
```
This can not override the version of the controller-gen to v0.2.0-beta.4 if an outdated version of `controller-gen` exists in the  `$GOPATH/bin`, this may lead to different behaviors when generating the code.

Also, checking the version of `controller-gen` is impossible, since there is no `version` flag in `controller-gen`.
Ref: https://github.com/kubernetes-sigs/controller-tools/issues/259

/kind bug

flameshimmer:
I am running into the exactly the same issue. Anyone please help?

jiachengxu:
Hi @flameshimmer , you can just remove the controller-gen binary in your path then do make generate to use the correct version

DirectXMan12:
this is fairly intentional -- it's intended that you use whatever's on your path, to allow you to specify custom versions of controller-gen, or to download pre-compiled versions.

DirectXMan12:
I'm not sure what the right solution is here.  If we had a `--version` flag we could warn at least.

mcristina422:
I've run into issues where different collaborators end up with different versions. Having a `--version` flag would at least help us detect this. Additionally the `go get` in the makefile is modifying the go.mod and that isn't desirable. 

But in the meantime I think my team will just build a docker container for controller-gen and run it with volume mounts to generate the manifests 


```Makefile
controller-gen:
	docker build -t controller_gen -f Dockerfile.controller_gen .
```

```Dockerfile
FROM golang:1.12.8 as builder

WORKDIR /workspace

RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.0-beta.5

ENTRYPOINT ["/go/bin/controller-gen"]
```

DirectXMan12:
we've got an open issue for the version flag, and a way to implement it, so it should come eventually (it's kubernetes-sigs/controller-tools#259)

adinunzio84:
[...]

PR #1289, opened 2020-01-08
don't depend on shared controller-gen which could be too old, use expected version directly

dionysius:
If, because of any reason, the `$GOPATH/bin/controller-gen` binary is not the expected version, it will generate faulty resources. The `which` check prevents any update to controller-gen.

Why not relying on go modules and reference controller-gen in there and using it with go run? (If controller-gen is `go get`ed, it will not download as long as the requested version is still in your go modules cache).

This enforces developers using kubebuilder to use go modules as dependency management for their project, but this is already enforced by providing `testdata/project-v2/go.mod`

**This PR is the minimal version to get the desired effect without workflow change**! But we can do better (additionally to this PR):

We could go a step ahead, instead of the target `controller-gen` in the Makefile, we could offer a `tools.go` file with the following content. [This is currently the best working way to define dev dependencies](https://github.com/golang/go/issues/25922)
```go
// +build tools

package main

import (
	_ "sigs.k8s.io/controller-tools/cmd/controller-gen"
)
```

```diff
--- a/testdata/project-v2/go.mod
+++ b/testdata/project-v2/go.mod
@@ -10,4 +10,5 @@ require (
        k8s.io/apimachinery v0.0.0-20190913080033-27d36303b655
        k8s.io/client-go v0.0.0-20190918160344-1fbdaa4c8d90
        sigs.k8s.io/controller-runtime v0.4.0
-       sigs.k8s.io/controller-tools v0.2.4 // indirect
+       sigs.k8s.io/controller-tools v0.2.4
 )
```

This will ensure, during creation, the default version of controller-tools is installed (currently v0.2.4), but the developer is free to up that version using `go get sigs.k8s.io/controller-tools@vX.X.X` and the reference is kept in go.mod

This makes `go get`ing the controller-gen in the Makefile abundant, remove that
```diff
--- a/testdata/project-v2/Makefile
+++ b/testdata/project-v2/Makefile
@@ -39,8 +39,8 @@ deploy: manifests
        kustomize build config/default | kubectl apply -f -
 
 # Generate manifests e.g. CRD, RBAC etc.
-manifests: controller-gen
+manifests:
       go run sigs.k8s.io/controller-tools/cmd/controller-gen $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
 # Run go fmt against code
 fmt:
@@ -51,7 +51,7 @@ vet:
        go vet ./...
 
 # Generate code
-generate: controller-gen
+generate:
        go run sigs.k8s.io/controller-tools/cmd/controller-gen object:headerFile=./hack/boilerplate.go.txt paths="./..."
 
 # Build the docker image
@@ -61,8 +61,3 @@ docker-build: test
 # Push the docker image
 docker-push:
        docker push ${IMG}
-
-# find or download controller-gen
-# download controller-gen if necessary
-controller-gen:
-       go get sigs.k8s.io/controller-tools@v0.2.4
```

### DIFF
--- a/pkg/scaffold/v2/makefile.go
+++ b/pkg/scaffold/v2/makefile.go
@@ -51,7 +51,7 @@ IMG ?= {{ .Image }}
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
 # Define how controller-gen is executed
-CONTROLLER_GEN ?= "go run sigs.k8s.io/controller-tools/cmd/controller-gen"
+CONTROLLER_GEN ?= go run sigs.k8s.io/controller-tools/cmd/controller-gen
 
 all: manager
 
--- a/pkg/scaffold/v2/makefile.go
+++ b/pkg/scaffold/v2/makefile.go
@@ -50,12 +50,8 @@ IMG ?= {{ .Image }}
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
-# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
-ifeq (,$(shell go env GOBIN))
-GOBIN=$(shell go env GOPATH)/bin
-else
-GOBIN=$(shell go env GOBIN)
-endif
+# Define how controller-gen is executed
+CONTROLLER_GEN ?= "go run sigs.k8s.io/controller-tools/cmd/controller-gen"
 
 all: manager
 
@@ -111,17 +107,5 @@ docker-push:
 # find or download controller-gen
 # download controller-gen if necessary
 controller-gen:
-ifeq (, $(shell which controller-gen))
-	@{ \
-	set -e ;\
-	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
-	cd $$CONTROLLER_GEN_TMP_DIR ;\
-	go mod init tmp ;\
-	go get sigs.k8s.io/controller-tools/cmd/controller-gen@{{.ControllerToolsVersion}} ;\
-	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
-	}
-CONTROLLER_GEN=$(GOBIN)/controller-gen
-else
-CONTROLLER_GEN=$(shell which controller-gen)
-endif
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@{{.ControllerToolsVersion}}
 `
--- a/testdata/project-v2/Makefile
+++ b/testdata/project-v2/Makefile
@@ -4,12 +4,8 @@ IMG ?= controller:latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
-# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
-ifeq (,$(shell go env GOBIN))
-GOBIN=$(shell go env GOPATH)/bin
-else
-GOBIN=$(shell go env GOBIN)
-endif
+# Define how controller-gen is executed
+CONTROLLER_GEN ?= "go run sigs.k8s.io/controller-tools/cmd/controller-gen"
 
 all: manager
 
@@ -65,16 +61,4 @@ docker-push:
 # find or download controller-gen
 # download controller-gen if necessary
 controller-gen:
-ifeq (, $(shell which controller-gen))
-	@{ \
-	set -e ;\
-	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
-	cd $$CONTROLLER_GEN_TMP_DIR ;\
-	go mod init tmp ;\
-	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.4 ;\
-	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
-	}
-CONTROLLER_GEN=$(GOBIN)/controller-gen
-else
-CONTROLLER_GEN=$(shell which controller-gen)
-endif
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.4
--- a/testdata/project-v2/Makefile
+++ b/testdata/project-v2/Makefile
@@ -5,7 +5,7 @@ IMG ?= controller:latest
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
 # Define how controller-gen is executed
-CONTROLLER_GEN ?= "go run sigs.k8s.io/controller-tools/cmd/controller-gen"
+CONTROLLER_GEN ?= go run sigs.k8s.io/controller-tools/cmd/controller-gen
 
 all: manager
 
--- a/docs/book/src/reference/generating-crd.md
+++ b/docs/book/src/reference/generating-crd.md
@@ -179,7 +179,7 @@ doing.
 Each controller-gen "generator" is controlled by an option to
 controller-gen, using the same syntax as markers.  For instance, to
 generate CRDs with "trivial versions" (no version conversion webhooks), we
-call `controller-gen crd:trivialVersions=true paths=./api/...`.
+call `go run sigs.k8s.io/controller-tools/cmd/controller-gen crd:trivialVersions=true paths=./api/...`.
 
 controller-gen also supports different output "rules" to control how
 and where output goes.  Notice the `manifests` make rule (condensed
@@ -198,13 +198,13 @@ CRD-related config (non-code) artifacts should end up in
 To see all the options for `controller-gen`, run
 
 ```shell
-$ controller-gen -h
+$ go run sigs.k8s.io/controller-tools/cmd/controller-gen -h
 ```
 
 or, for more details:
 
 ```shell
-$ controller-gen -hhh
+$ go run sigs.k8s.io/controller-tools/cmd/controller-gen -hhh
 ```
 
 [marker-ref]: ./markers.md "Markers for Config/Code Generation"
--- a/testdata/project-v2/go.mod
+++ b/testdata/project-v2/go.mod
@@ -10,4 +10,5 @@ require (
 	k8s.io/apimachinery v0.0.0-20190913080033-27d36303b655
 	k8s.io/client-go v0.0.0-20190918160344-1fbdaa4c8d90
 	sigs.k8s.io/controller-runtime v0.4.0
+	sigs.k8s.io/controller-tools v0.2.4 // indirect
 )


dionysius:
another cla provider... CLA should be signed by now

dionysius:
Fixed golden script and reflected also in the docs.

I assume this is not relevant for this PR: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/utils/markerdocs/main.go#L222

testdata/project-v2/Makefile
```diff
@@ -52,7 +52,7 @@ vet:
 
 # Generate code
 generate: controller-gen
-	$(CONTROLLER_GEN) object:headerFile=./hack/boilerplate.go.txt paths="./..."
+	go run sigs.k8s.io/controller-tools/cmd/controller-gen object:headerFile=./hack/boilerplate.go.txt paths="./..."
```
Adirio: Did you do this change manually or did you run the generate golden script?

dionysius: that was manually, here https://github.com/kubernetes-sigs/kubebuilder/commit/e75141929f0f6e626b5087fc4b675ddcc6115148, i can revert that if you want

pkg/scaffold/v2/makefile.go
```diff
@@ -111,17 +111,5 @@ docker-push:
...
-	cd $$CONTROLLER_GEN_TMP_DIR ;\
-	go mod init tmp ;\
-	go get sigs.k8s.io/controller-tools/cmd/controller-gen@{{.ControllerToolsVersion}} ;\
-	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
-	}
-CONTROLLER_GEN=$(GOBIN)/controller-gen
-else
-CONTROLLER_GEN=$(shell which controller-gen)
-endif
+	go get sigs.k8s.io/controller-tools@{{.ControllerToolsVersion}}
```
estroz: I would keep `CONTROLLER_GEN=$(GOBIN)/controller-gen`.

dionysius: Thats absolutely possible, it would look like
`CONTROLLER_GEN="go run sigs.k8s.io/controller-tools/cmd/controller-gen"`

Now i kinda want to change it that way, the more we need to use it, the better it is as a variable

But we have to keep the go get part: From experience, if we don't strictly define it as dev-dependency (see lower part of this PR), a go mod tidy **will remove** it from go.mod

Adirio: @estroz The `$(GOBIN)` approach is less OS-friendly.

dionysius: i just pushed https://github.com/kubernetes-sigs/kubebuilder/pull/1289/commits/3bc20527b84716661b4aae02393f52908cc0c414, GOBIN is gone

testdata/project-v2/Makefile
Adirio: No, what I mean is that `object:headerFile=./hack/boilerplate.go.txt` should be `object:headerFile="hack/boilerplate.go.txt"` (#1288)

dionysius: Can someone explain to me why in the tests this line leads to this diff (and thus the [test fails](https://travis-ci.org/kubernetes-sigs/kubebuilder/builds/634870273?utm_source=github_status&utm_medium=notification))

```diff
diff --git a/testdata/project-v2/Makefile b/testdata/project-v2/Makefile
index 9992458..5e20236 100644
--- a/testdata/project-v2/Makefile
+++ b/testdata/project-v2/Makefile
@@ -52,7 +52,7 @@ vet:
 
 # Generate code
 generate: controller-gen
-	$(CONTROLLER_GEN) object:headerFile=./hack/boilerplate.go.txt paths="./..."
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 # Build the docker image
 docker-build: test
```

Adirio: It is a bug introduced by other commit that got merged. #1288 should fix it, but you can also change that if you want.

dionysius: @Adirio perfect, I merged your pull request https://github.com/kubernetes-sigs/kubebuilder/pull/1289/commits/af9bb4ea4cf67a9fcdd09d7bd1b9f55d859cfd87

This branch now got pretty big... maybe I should squash everything?

Adirio: Yeah go for it, squash and force-push

testdata/project-v2/Makefile
```diff
@@ -52,7 +48,7 @@ vet:
 
 # Generate code
 generate: controller-gen
-	$(CONTROLLER_GEN) object:headerFile=./hack/boilerplate.go.txt paths="./..."
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
```
Adirio: I think this may need a rebase as #1287 was merged including this change

dionysius: rebased with master

dionysius:
`go run` has one (IMO small) drawback: The last compilation step will always be done on every call of `go run`, while the binary is ready forever. So you trade a bit of cpu/time for flexibility.

@camilamacedo86 Thank you. Should we move the discussion somewhere else so we don't clutter this issue? See #1301

mengqiy:
It seems this effort can help with #909.

mengqiy:
> go run has one (IMO small) drawback: The last compilation step will always be done on every call of go run, while the binary is ready forever. So you trade a bit of cpu/time for flexibility.

First invocation may be slow, but subsequent invocation should be fast thanks to compiler cache.
Compiling on every call is not great, but probably acceptable.

Adirio:
> > go run has one (IMO small) drawback: The last compilation step will always be done on every call of go run, while the binary is ready forever. So you trade a bit of cpu/time for flexibility.
> 
> First invocation may be slow, but subsequent invocation should be fast thanks to compiler cache.
> Compiling on every call is not great, but probably acceptable.

Let's not forget that Golang was desing to be super fast at compilation time. IMHO, the advantages (version handling, OS-friendly) have a greater weight than the small compilation time that will be required. It has my LGTM.

Adirio:
Only is missing approval. @mengqiy did you make a final decission?

mengqiy:
@DirectXMan12 @droot Do you guys have any concerns before this getting in?

DirectXMan12:
This was actually semi-intentional originally -- technically, CT can have different versions of the k8s dependencies that CR and friends, and we don't want to cause issues there.  Until we get a version of k8s deps with proper import path support, we're gonna need to either stick with what we have now, or do something wacky like create a "tools" module with different deps.

DirectXMan12:
/hold

(just to play it safe here -- feel free to cancel whenever this get resolved w/o me)

estroz:
How about something like this Makefile:

```make
EXTERNAL_BIN_TMP=/tmp/kubebuilder-bin
CONTROLLER_GEN=$(EXTERNAL_BIN_TMP)/bin/controller-gen

...
controller-gen:
ifeq (,$(wildcard $(CONTROLLER_GEN)))
	@{ \
	set -e ;\
	mkdir -p $(EXTERNAL_BIN_TMP) ;\
	pushd $(EXTERNAL_BIN_TMP) ;\
	go mod init tmp ;\
	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.4 ;\
	go build -o bin/controller-gen sigs.k8s.io/controller-tools/cmd/controller-gen ;\
	rm go.mod go.sum ;\
	}
endif
```

And making a note in the docs to check your global `controller-gen` version before running it directly.

pkg/scaffold/v2/makefile.go
```diff
@@ -50,12 +50,8 @@ IMG ?= {{ .Image }}
...
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
-# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
-ifeq (,$(shell go env GOBIN))
-GOBIN=$(shell go env GOPATH)/bin
-else
-GOBIN=$(shell go env GOBIN)
-endif
+# Define how controller-gen is executed
+CONTROLLER_GEN ?= go run sigs.k8s.io/controller-tools/cmd/controller-gen
```
Adirio: Can we tag a version here? Would this work?

```suggestion
CONTROLLER_GEN ?= go run sigs.k8s.io/controller-tools/cmd/controller-gen@{{.ControllerToolsVersion}}
```

dionysius: Unfortunately not, I played around a bit with https://github.com/dionysius/go-run-version

dionysius:
By the way kustomized is used like 3 times in the makefile:

```Makefile
install: manifests #with or without target kustomize analog controller-gen depending if with or without tools.go file
	$(KUSTOMIZE) build config/crd | kubectl apply -f -

KUSTOMIZE ?= go run sigs.k8s.io/kustomize/kustomize/v3

#with or without target kustomize analog controller-gen depending if with or without tools.go file
kustomize:
	go get  sigs.k8s.io/kustomize/kustomize/v3@...
```

pkg/scaffold/v2/makefile.go
Adirio: My question here was, can we tag a version for `go run`? Went to your repo but I see no results there. Is is possible?

I still don't see why the kubebuilder binary will have this dependency outside of the `testdata` packages which are actually other modules so theorically it shouldn't matter. Kubebuilder will just have a string containing the version, but why does that add a dependency? The makefile from the generated code does install and call controller-gen, so the dependency there is clear.

Adirio:
I still don't see why the kubebuilder binary will have this dependency outside of the `testdata` packages which are actually other modules so theorically it shouldn't matter. Kubebuilder will just have a string containing the version, but why does that add a dependency? The makefile from the generated code does install and call controller-gen, so the dependency there is clear.

(Lets keep this discussion outside of my `go run` comment to keep things sorted.)

dionysius:
I have never looked from that angle, so everything I say is now subject to be wrong. Since I was focused on only the resulting project built by kubebuilder, I was only in that context. 

But I assume, since testdata is a full-fleshed own module with go.mod, shouldn't it stay it's own module and doesn't poison upwards. We can only find this out with testing properly (also with different go versions).

pkg/scaffold/v2/makefile.go
dionysius: to keep it short: correct, that's what I wanted to say, not possible to use a tag/version with `go run` with my tests as well

Adirio: You can mark this conversation as resolved to keep the PR clean.

Adirio:
Exactly, separate modules so there **SHOULD** not interfere, that's why I said I think I'm not understanding @DirectXMan12 right. You can be sure that this PR doesn't poison upwards just by calling `go mod tidy` and watching that there were no added dependencies.

dionysius:
But so far I'm experienced with go modules, `go get controller-gen...` will "poison" your project, sharing their dependencies

Adirio:
> But so far I'm experienced with go modules, `go get controller-gen...` will "poison" your controller project, sharing their dependencies (if there are any at all)

Ohhhhh, now I understand. Calling `go get ...` from the scaffolded project injects it as a dependency of the project. There was some discussion about a `-g` flag for `go get` so that it doesn't do that but seems abandoned. What about creating a temporal directory, using `go mod init`, using the `go get ...` and finally removing the temporal directory?

dionysius:
Well, this part is actually me removing it here in this PR ':D But more becase `which ...` is a global check, and my suggestion was to tie the version to your scaffold project (with all its pros and cons)

https://github.com/kubernetes-sigs/kubebuilder/pull/1289/files#diff-5e36985ed225e6a2699eb38abb06ce54L114-L126

Adirio:
> Well, this part is actually me removing it here in this PR ':D But more becase `which ...` is a global check, and my suggestion was to tie the version to your scaffold project (with all its pros and cons)

I definetely needed to step back and review the PR from the start. Will hide useless comments while I walk backwards slowly as if I was never here.

mengqiy:
> ifeq (,$(wildcard $(CONTROLLER_GEN)))

@estroz what does this do in your suggestion?

estroz:
>@estroz what does this do in your suggestion?
Test if the path `$(CONTROLLER_GEN)` exists.

dionysius:
wildcard sound like just as an alternative to `which` - if I understood you correctly. How about using the bin/ folder for that? then it is local to the scaffolded project

```
bin/controller-gen:
  # steps for creating controller-gen in separate tmp folder
  mv where/tmp/is/controller-gen bin/controller-gen
```

Just beeing curious: Is the dependency sharing really that of a problem? What are the risks I may not see?

droot:
Sorry for coming late to the party :)
Appreciate everyone taking time to review this PR. In previous project versions `v1`, we had `go run` based approach and struggled with incompatible dependencies between CR and CT and decided to decouple the deps in project version `v2` and we have been very happy with it so far. So, we don't want to go back to that world where KB project shares deps between CR/CT.
Having said that, the original problem (that started this PR) still remains, where 

>If, because of any reason, the $GOPATH/bin/controller-gen binary is not the expected version, it will generate faulty resources. The which check prevents any update to controller-gen.

One way to solve this would be to keep a `MinControllerGenVer: vX.Y.Z (filled during scaffolding)` in the `PROJECT` file and `make controller-gen` can query `controller-gen --version` and compare if minControllerGenVer invariant holds otherwise give a warning to user or auto-upgrade `controller-gen` binary to `MinControllerGenVer` or (`latest stable` version by querying controller-tools-releases page). [I think @mengqiy suggested this in an offline conversation in the past]

I this this PR has done its job and would suggest closing this PR and continue the discussion in a separate issue. 

/cc @dionysius @Adirio @mengqiy @estroz

dionysius:
I'll close this PR now. I just like to remind you to try to couple the controller-gen version to the scaffolded project - not globally. There are many ideas in the comments - whoever coming here to read back, go through these comments.
There is also [gobin](https://github.com/myitcv/gobin), I personally don't like the additional cache location, but it would do the job as well

droot:
> I just like to remind you to try to couple the controller-gen version to the scaffolded project - not globally. 

Yes, that's important. Thanks for pointing it out.

PR #1290, opened 2020-01-08
feature: implement plugin proposal in #1250

estroz:
Implements #1250 except for version format enforcement and plugin name resolution (will be done in follow-ups).

TODO:
- ~finalize plugin interface set~
- ~discuss pro's/con's of this implementation, smooth over rough edges~

/cc @DirectXMan12 @droot @Adirio @joelanford @shawn-hurley @camilamacedo86


pkg/cli/alpha.go
```diff
@@ -14,27 +14,27 @@ See the License for the specific language governing permissions and
...
+package cli
 
 import (
+	"fmt"
+
 	"github.com/spf13/cobra"
 )
 
-// newAlphaCommand returns alpha subcommand which will be mounted
+// newAlphaCnd returns alpha subcommand which will be mounted
```
joelanford: Nit: typo
```suggestion
// newAlphaCmd returns alpha subcommand which will be mounted
```

pkg/cli/cli.go
```diff
@@ -0,0 +1,271 @@
...
+	cmd            *cobra.Command
+	extraCommands  []*cobra.Command
+	plugins        map[string]plugin.BasePlugin
+}
+
+// New creates a new CLI instance.
+func New(opts ...Option) (*CLI, error) {
+	c := &CLI{
+		commandName:    "kubebuilder",
+		defProjVersion: project.Version2,
```
joelanford: Given the seeming consensus around name and version, we may need to re-think the `--project-version` flag and this `defProjVersion` field.

If we care about backward compatibility, we could keep and deprecate the `--project-version` flag, and make it automatically choose the "go" plugins. Maybe we don't care since `--project-version` is used during project initialization only (I think?). 

We probably need a new `init` flag, maybe `--plugin`, that takes something like `<name>/<version>`. So for the existing kubebuilder plugins, it would understand `go/v1` and `go/v2`.

Lastly, do we want to have a convention to make it easier to distinguish similar plugins from different authors? Consider two different plugins named `foo`. Should we include a domain component (e.g. `foo.kubebuilder.io` and `foo.operator-framework.io`)? If so, this field starts to look a lot like a Kubernetes group version string (maybe more familiar for users) and it would make it more apparent which binary a given KB project would require. One thing I _don't_ like about this is that it requires more typing from users.

cmd/main.go
```diff
@@ -17,187 +17,28 @@ limitations under the License.
...
-		newInitProjectCmd(),
-		newCreateCmd(),
-		version.NewVersionCmd(),
+	c, err := cli.New(
+		cli.WithPlugins(
+			&pluginv1.Plugin{},
+			&pluginv2.Plugin{},
+		),
+		cli.WithExtraCommands(
+			version.NewVersionCmd(),
```
joelanford: Good idea leaving the `version` command out of the base CLI.

pkg/cli/create.go
```diff
@@ -0,0 +1,92 @@
...
+func (c CLI) newCreateCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Scaffold a Kubernetes API or webhook",
+		Long:  `Scaffold a Kubernetes API or webhook.`,
+	}
+	cmd.AddCommand(c.newCreateAPICmd())
+
+	foundProject, projectVersion := getProjectVersion()
+	if !foundProject || projectVersion == project.Version2 {
```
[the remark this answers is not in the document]
estroz: The latter, will update.

pkg/cli/cli.go
```diff
@@ -0,0 +1,271 @@
...
+		return nil
+	}
+}
+
+// WithPlugins is an Option that sets the CLI's plugins. Duplicate plugin
+// versions will result in an error.
+func WithPlugins(plugins ...plugin.BasePlugin) Option {
+	return func(c *CLI) error {
+		for _, p := range plugins {
+			if _, ok := c.plugins[p.Version()]; ok {
```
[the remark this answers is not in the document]
estroz: Since `c.plugins` is accessed by project version below, the key type needs to be `p.Version()`. However `c.plugins` should contain a slice of `Plugin` per key as to not overwrite preceding plugins.

pkg/cli/cli.go
estroz: @DirectXMan12 @droot thoughts on ^?

DirectXMan12: I kinda imagined it'd go like this:

- `kubebuilder init --project-version v{1,2}` are mostly an alias for `kubebuilder init --plugin go:v{1,2}` (see below)
- `kubebuilder init --plugin NAME:VERSION` (or `NAME/VERSION`) initializes a project with the given plugin, writing that to the project file with the special project version of `3x` or something.  We might want special handling if you actually used `--project-version` to write `1` or `2`
- After `init`, you never specify project version or base plugin again

As for suffixes, might not be a bad idea to encourage it, but I don't expect it'll be a big issue (famous last words) and it is a lot of extra typing.  Maybe allow it as a disambiguation form, but don't require it?

pkg/cli/cli.go
```diff
@@ -0,0 +1,276 @@
...
+			if in, ok := p.(plugin.InjectsCommandName); ok {
+				in.InjectCommandName(c.commandName)
+			}
+		}
+	}
+
+	// Check the current working directory for a PROJECT file.
+	// Return whether it was found, and if so, what project version
+	// it is for.
+	foundProject, projectVersion := getProjectVersion()
```
DirectXMan12: seems like we should be loading this once and keeping it around (and potentially passing to plugins, because we don't want plugins to have to implement this individually, which'd be a mess for updating)

pkg/cli/cli.go
```diff
@@ -0,0 +1,276 @@
...
+
+	// If a PROJECT file wasn't found, see if we can find the
+	// project version in a command line flag.
+	if !foundProject {
+		projectVersion, _ = c.getBaseFlags()
```
DirectXMan12: As per my other comment, I think we can actually just move project-version to `init`.  That still leaves plugin though.

pkg/cli/cli.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	return true, projectInfo.Version
+}
+
+// getBaseFlags parses the command line arguments, looking for --project-version
+// and help. If an error occurs or only --help is set, getBaseFlags returns an
+// empty string and true. Otherwise, getBaseFlags returns the project version
+// and false.
+func (c CLI) getBaseFlags() (string, bool) {
+	fs := pflag.NewFlagSet("base", pflag.ExitOnError)
```
DirectXMan12: FYI, this logic yields "unknown flag: project-version" when we get to the main flag parsing logic.  We prob want this defined on the main command too.

pkg/cli/create.go
```diff
@@ -0,0 +1,108 @@
...
+		Use:   "api",
+		Short: "Scaffold a Kubernetes API",
+		Long: `Scaffold a Kubernetes API.
+`,
+		RunE: errCmdFunc(
+			fmt.Errorf("api subcommand requires an existing project"),
+		),
+	}
+
+	found, projectVersion := getProjectVersion()
```
DirectXMan12: as above -- let's not re-parse this a bunch of times

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,94 @@
...
+	// Version returns the project version that this plugin implements.
+	// For example, Kubebuilder's Go v2 plugin implementation would return 2.
+	Version() string
+}
+
+func Key(p Base) string {
+	return path.Join(p.Name(), "v"+strings.TrimLeft(p.Version(), "v"))
+}
+
+type InjectsCommandName interface {
```
DirectXMan12: can we just switch this to the context-based approach?

pkg/cli/init_project.go
```diff
@@ -0,0 +1,127 @@
...
+	// so that it shows up in help and does not cause a parse error.
+	cmd.Flags().String("project-version", c.defProjVersion, fmt.Sprintf("project version, (%s)", strings.Join(c.getAvailableProjectVersions(), ", ")))
+
+	// Pre-parse the project version and help flags so that we can
+	// dynamically bind to a plugin's init implementation (or not).
+	projectVersion, isHelpOnly := c.getBaseFlags()
+
+	// If only the help flag was set, return the command as is.
+	if isHelpOnly {
+		return cmd
```
DirectXMan12: doesn't this mean that a plugin doesn't get to modify help?

pkg/cli/webhook.go
```diff
@@ -0,0 +1,91 @@
...
+		cwg, hasGetter = p.(plugin.CreateWebhookPluginGetter)
+	}
+	if !hasGetter {
+		err := fmt.Errorf("project version %q does not support a webhook creation plugin",
+			projectVersion)
+		cmdErr(cmd, err)
+		return
+	}
+
+	cwp := cwg.GetCreateWebhookPlugin()
```
DirectXMan12: This logic seems like we can add a helper since it's pretty identical across commands.

pkg/cli/cli.go
joelanford: I assume you mean passing a `ProjectSpec` (or similarly named) struct that is read in from the `PROJECT` file?

pkg/cli/init_project.go
joelanford: The intent with this was to handle `kubebuilder init --help`, in which case we could print out a general help. In this case, we don't know which plugin to use for help output.

pkg/cli/cli.go
Adirio: I'm working on centrallizing the configuration file (#1325). I was using `internal/config` as the package to prevent third party to modify the file, but as I see you do need this config, I will probably expose the struct. IMHO, `pkg/model/config` seems to be a good point to have a `Config` struct with the information from that file.

pkg/cli/alpha.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 The Kubernetes Authors.
+Copyright 2019 The Kubernetes Authors.
```
shawn-hurley: I don't think that we should be updating the copyright

test/e2e/v2/e2e_suite.go
```diff
@@ -66,7 +66,7 @@ var _ = Describe("kubebuilder", func() {
 			err := kbc.Init(
 				"--project-version", "2",
 				"--domain", kbc.Domain,
-				"--dep=false")
+				"--fetch-deps=false")
```
Adirio: Does this imply that there was a backwards incompatible change?

pkg/cli/webhook.go
```diff
@@ -0,0 +1,91 @@
...
+			return
+		}
+		cwg, hasGetter = p.(plugin.CreateWebhookPluginGetter)
+	}
+	if !hasGetter {
+		err := fmt.Errorf("project version %q does not support a webhook creation plugin",
+			projectVersion)
+		cmdErr(cmd, err)
+		return
+	}
```
Adirio: See equivalent comment for `bindCreateAPI`

pkg/cli/init_project.go
```diff
@@ -0,0 +1,127 @@
...
+		if hasGetter {
+			log.Fatalf("duplicate initialization plugins for project version %q: %s, %s",
+				projectVersion, ig.Name(), p.Name())
+		}
+		ig, hasGetter = p.(plugin.InitPluginGetter)
+	}
+	if !hasGetter {
+		log.Fatalf("project version %q does not support a project initialization plugin",
+			projectVersion)
+	}
```
Adirio: See comment for `bindCreateAPI`

pkg/cli/init_project.go
Adirio: Does this have something to do with the use of `log.Fatal` where you were using `cdmErr` and `return` in create api and webhook commands?

pkg/plugin/v1/api.go
```diff
@@ -0,0 +1,149 @@
...
+	if err := p.apiScaffolder.Scaffold(); err != nil {
+		log.Fatal(err)
+	}
+
+	if err := p.postScaffold(); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func (p *createAPIPlugin) postScaffold() error {
```
Adirio: You changed this method name for the init plugins, were you trying to avoid the "scaffold" word? If so you still have one here. `postAPI`?

pkg/cli/cli.go
```diff
@@ -0,0 +1,276 @@
...
+	"fmt"
+	"log"
+	"os"
+
+	"sigs.k8s.io/kubebuilder/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/project"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
```
Adirio: The order of these two groups should be reverted.

pkg/cli/create.go
```diff
@@ -0,0 +1,108 @@
...
+
+package cli
+
+import (
+	"fmt"
+
+	"sigs.k8s.io/kubebuilder/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/project"
+
+	"github.com/spf13/cobra"
```
Adirio: Import order.

pkg/cli/init_project.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"fmt"
+	"log"
+	"strconv"
+	"strings"
+
+	"sigs.k8s.io/kubebuilder/cmd/util"
+	"sigs.k8s.io/kubebuilder/pkg/plugin"
+
+	"github.com/spf13/cobra"
```
Adirio: Import order.

pkg/cli/webhook.go
```diff
@@ -0,0 +1,91 @@
...
+*/
+
+package cli
+
+import (
+	"fmt"
+
+	"sigs.k8s.io/kubebuilder/pkg/plugin"
+
+	"github.com/spf13/cobra"
```
Adirio: Import order.

pkg/plugin/util/util.go
```diff
@@ -0,0 +1,175 @@
...
+	"os/exec"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"sigs.k8s.io/kubebuilder/pkg/scaffold"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/resource"
+
+	"github.com/spf13/pflag"
+	"golang.org/x/tools/go/packages"
```
Adirio: Import order.

pkg/plugin/v1/api.go
```diff
@@ -0,0 +1,149 @@
...
+	"os/exec"
+	"strings"
+
+	"sigs.k8s.io/kubebuilder/cmd/util"
+	"sigs.k8s.io/kubebuilder/pkg/plugin"
+	pluginutil "sigs.k8s.io/kubebuilder/pkg/plugin/util"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold"
+	"sigs.k8s.io/kubebuilder/plugins/addon"
+
+	"github.com/spf13/pflag"
```
Adirio: Import order.

pkg/plugin/v1/init.go
```diff
@@ -0,0 +1,190 @@
...
+	"path/filepath"
+	"strings"
+
+	"sigs.k8s.io/kubebuilder/cmd/util"
+	"sigs.k8s.io/kubebuilder/pkg/plugin"
+	pluginutil "sigs.k8s.io/kubebuilder/pkg/plugin/util"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/project"
+
+	"github.com/spf13/pflag"
```
Adirio: Import order.

pkg/plugin/v2/init.go
```diff
@@ -0,0 +1,165 @@
...
+	"path/filepath"
+	"strings"
+
+	"sigs.k8s.io/kubebuilder/cmd/util"
+	"sigs.k8s.io/kubebuilder/pkg/plugin"
+	pluginutil "sigs.k8s.io/kubebuilder/pkg/plugin/util"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/project"
+
+	"github.com/spf13/pflag"
```
Adirio: Import order.

shawn-hurley:
Overall I think that this good

some points:

I think we need to discuss the version more.
We need an interface for CLI instead of a struct with hidden fields.

Want to see what folks think about the default for help will show the plugins that can be used as well as the default flags and such? Is that possible or is that already working and I just can't see it?

pkg/cli/cli.go
```diff
@@ -0,0 +1,276 @@
...
+const (
+	noticeColor = "\033[1;36m%s\033[0m"
+)
+
+// Option is a function that can configure the CLI
+type Option func(*CLI) error
+
+// CLI defines the command line structure and interfaces that are used to
+// scaffold kubebuilder project files.
+type CLI struct {
```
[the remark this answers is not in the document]
Adirio: Why? It does have methods.

pkg/plugin/v1/plugin.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+var _ plugin.Base = Plugin{}
+
+func (_ Plugin) Name() string {
+	return "go"
+}
+
+func (_ Plugin) Version() string {
+	return "1"
```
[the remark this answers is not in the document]
Adirio: I think that may rise a lot of issues as we are expecting the default `go` plugin to exist for every version and using `Bar` as version would break that.

pkg/cli/cli.go
```diff
@@ -0,0 +1,276 @@
...
+	if err := c.initialize(); err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+// Run runs the CLI.
+func (c CLI) Run() error {
+	return c.cmd.Execute()
+}
```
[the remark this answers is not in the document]
estroz: If not `Run`, `Execute` would make sense. The CLI isn't really deploying anything, ex. resources, it is a program that is running/executing.

pkg/cli/cli.go
estroz: I will merge #1325 here one it gets some reviews.

pkg/cli/create.go
```diff
@@ -0,0 +1,108 @@
...
+			return
+		}
+		cag, hasGetter = p.(plugin.CreateAPIPluginGetter)
+	}
+	if !hasGetter {
+		err := fmt.Errorf("project version %q does not support an API creation plugin",
+			projectVersion)
+		cmdErr(cmd, err)
+		return
+	}
```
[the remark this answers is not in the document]
estroz: I coded this out too hastily. I only care about duplicates not order, so something like your solution is correct.

pkg/plugin/v1/api.go
estroz: Its an unnecessary change so I'm switching back to `postScaffold`.

pkg/plugin/v1/plugin.go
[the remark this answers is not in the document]
estroz: Version formatting and plugin composition go hand-in-hand, so we should follow up this PR with a discussion on both.

test/e2e/v2/e2e_suite.go
estroz: Yes, these [particular flags](https://github.com/kubernetes-sigs/kubebuilder/blob/c52ba01/cmd/init_project.go#L94-L98) are not replicated in v2 from v1 plugin `BindFlags`. To be backwards-compatible I will add deprecated flags to v2.

Adirio:
@estroz merging Master wasn't the best idea IMHO. I think you need to rebase your branch to Master, 257 files changed, and 10k lines modified seem to be a bit too much.

test/e2e/v2/e2e_suite.go
Adirio: I actually looked into the flags and this change is ok, you can mark this conversation as resolved to reduce the noise.

estroz:
@Adirio its because the target feature branch isn't up to date with master. It needs to be updated.

pkg/cli/alpha.go
estroz: Reverted.

pkg/cli/cli.go
estroz: Handling this information will be easier if we merge #1325 first.

pkg/plugin/plugin.go
estroz: Done.

Adirio:
> its because the target feature branch isn't up to date with master. It needs to be updated.

Oh, right, you are not PR-ing against master

pkg/plugin/v1/init.go
```diff
@@ -81,12 +79,17 @@ func (p *initPlugin) BindFlags(fs *pflag.FlagSet) {
...
+	}
+	if err := fs.MarkDeprecated("depArgs", "will be removed with version 1 scaffolding"); err != nil {
+		panic(err)
+	}
 
 	// boilerplate args
 	fs.StringVar(&p.boilerplate.Path, "path", "", "path for boilerplate")
-	fs.StringVar(&p.boilerplate.License, "license", "apache2", "license to use to boilerplate.  May be one of apache2,none")
+	fs.StringVar(&p.boilerplate.License, "license", "apache2", "license to use to boilerplate. "+
+		"May be one of apache2,none")
```
Adirio: NIT: I would break between params:
```go
	fs.StringVar(&p.boilerplate.License, "license", "apache2",
		"license to use to boilerplate.  May be one of apache2,none")
```

pkg/plugin/v2/init.go
```diff
@@ -73,9 +75,21 @@ func (p *initPlugin) BindFlags(fs *pflag.FlagSet) {
...
+	}
+	if err := fs.MarkDeprecated("depArgs", "will be removed with version 1 scaffolding"); err != nil {
+		panic(err)
+	}
+
 	// boilerplate args
 	fs.StringVar(&p.boilerplate.Path, "path", "", "path for boilerplate")
-	fs.StringVar(&p.boilerplate.License, "license", "apache2", "license to use to boilerplate.  May be one of apache2,none")
+	fs.StringVar(&p.boilerplate.License, "license", "apache2", "license to use to boilerplate. "+
+		"May be one of apache2,none")
```
Adirio: Same as above

pkg/plugin/v2/plugin.go
```diff
@@ -18,7 +18,7 @@ package v2
 
 import (
 	"sigs.k8s.io/kubebuilder/pkg/plugin"
-	"sigs.k8s.io/kubebuilder/pkg/plugin/v1"
+	v1 "sigs.k8s.io/kubebuilder/pkg/plugin/v1"
```
Adirio: I think this is because it expects packages `v1` and `v2` to be different versions of the `plugin` package. It made me think that we may want to include the plugin name in the package name.

estroz:
@Adirio if we can update the target feature branch with master, we should be ok since we don't need to merge CI fixes going forward. The only following merge will be your changes in #1325.

@DirectXMan12 @droot @mengqiy I don't have access to the feature branch. Can you merge it with master?

Adirio:
> The only following merge will be your changes in #1325.

I think that #1330 may also be a good idea to merge before, as it targets `cmd` package. Basically it is making the separation between `cmd` and `pkg/scaffold` clearer. This way I can keep working on enhancing the `pkg/scaffold` while the plugin system gets implemented without requiring too many rebases and conflict resolution from any of the two sides.

DirectXMan12:
Hey, sorry about that.  I'll fast-forward as needed.  Feel free to ping me on Slack if I'm not responding here.

DirectXMan12:
should be up-to-date now

DirectXMan12:
Gonna fix this branch too

DirectXMan12:
please do pull out the merge commit though

pkg/cli/cli.go
DirectXMan12: The intention was at least that 3x would indicate to aware things that plugins are involved -- if you try to work with a project scaffolded with a particular set of plugins and aren't aware, things could go awry.

We might want to punt on that for later, though.

pkg/plugin/v1/plugin.go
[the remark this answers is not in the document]
DirectXMan12: yeah.  I think for the moment: add a `TODO` for this, and we'll track them all down before final merge.

We can do the "move fast and break things" strategy on this branch.

DirectXMan12:
approved or almost approved a couple of those PRs from @Adirio.  LMK if there's another one blocking, but I think if we rebase on that we can probably merge this as a "here's a working step" and keep iterating.

pkg/cli/cli.go
camilamacedo86: The  #1325 is merged now @estroz

pkg/cli/cli.go
```diff
@@ -0,0 +1,273 @@
...
+
+// errCmdFunc returns a cobra RunE function that returns the provided error
+func errCmdFunc(err error) func(*cobra.Command, []string) error {
+	return func(*cobra.Command, []string) error {
+		return err
+	}
+}
+
+// getProjectVersion tries to load PROJECT file and returns if the file exist
```
camilamacedo86:
```suggestion
// getProjectVersion load project and returns the version or an error.
```

pkg/cli/cli.go
```diff
@@ -0,0 +1,273 @@
...
+// and the version string
```
camilamacedo86:
```suggestion
```

pkg/cli/cli.go
```diff
@@ -0,0 +1,273 @@
...
+func getProjectVersion() (string, bool) {
```
camilamacedo86: I do not think that it should return a bool. Just the version or an error.

pkg/cli/internal/config.go
```diff
@@ -0,0 +1,47 @@
...
+	"os"
+
+	"sigs.k8s.io/kubebuilder/internal/config"
+)
+
+const (
+	noticeColor = "\033[1;36m%s\033[0m"
+)
+
+// ConfiguredAndV1 returns true if the project is already configured.
```
camilamacedo86:
```suggestion
// Configured returns true if the project is already configured.
```

pkg/cli/internal/config.go
```diff
@@ -0,0 +1,47 @@
...
+func Configured() bool {
+	_, err := config.Read()
+	return err == nil || os.IsExist(err)
+}
+
+// ConfiguredAndV1 returns true if the project is already configured and is v1.
+func ConfiguredAndV1() bool {
```
camilamacedo86: Why we should care about the V1? 
V1 is deprecated and I think that indeed we could remove it from the project already.

camilamacedo86: Also, `ConfiguredAndV1` shows hurting the cohesion

pkg/cli/internal/config.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"sigs.k8s.io/kubebuilder/internal/config"
+)
+
+const (
+	noticeColor = "\033[1;36m%s\033[0m"
+)
+
+// ConfiguredAndV1 returns true if the project is already configured.
+func Configured() bool {
```
camilamacedo86: WDYT about : 
```suggestion
func IsConfigured() bool {
```
In order to follow the namenclatue standards for booleans?

pkg/cli/create.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (c cli) newCreateCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Scaffold a Kubernetes API or webhook",
+		Long:  `Scaffold a Kubernetes API or webhook.`,
+	}
+	cmd.AddCommand(c.newCreateAPICmd())
+
+	if !internal.ConfiguredAndV1() {
```
camilamacedo86:
```suggestion
	if !internal.IsConfigured() && !internal.IsV1()  {
```

camilamacedo86: IHMO it is more clear and does not hurt the cohesion.

camilamacedo86: Is the logic here correct? 
Should not we check if isConfigured and if is not V1 to do both scaffolds?

pkg/cli/create.go
```diff
@@ -0,0 +1,113 @@
...
+		Use:   "api",
+		Short: "Scaffold a Kubernetes API",
+		Long: `Scaffold a Kubernetes API.
+`,
+		RunE: errCmdFunc(
+			fmt.Errorf("api subcommand requires an existing project"),
+		),
+	}
+
+	projectVersion, found := getProjectVersion()
```
camilamacedo86:
```suggestion
	projectVersion, err := getProjectVersion()
```

pkg/cli/create.go
```diff
@@ -0,0 +1,113 @@
...
+	if !found {
```
camilamacedo86:
```suggestion
	if err != nill {
```

pkg/cli/create.go
```diff
@@ -0,0 +1,113 @@
...
+		msg := `For project-specific information, run this command in the root directory of a
+project.
+`
+		cmd.Long = fmt.Sprintf("%s\n%s", cmd.Long, msg)
```
camilamacedo86: I think we should return an error.

pkg/cli/webhook.go
```diff
@@ -0,0 +1,97 @@
...
+		),
+	}
+
+	projectVersion, found := getProjectVersion()
+	if !found {
+		msg := `For project-specific information, run this command in the root directory of a
+project.
+`
+		cmd.Long = fmt.Sprintf("%s\n%s", cmd.Long, msg)
+		return cmd
```
camilamacedo86: same here. 
Have we not already a func to getProjectVersion that can be used here instead of re-impl a local func?
Also, I think that it should return error and not bool as the command.

pkg/cli/webhook.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	// Lookup the plugin for projectVersion and bind it to the command.
+	c.bindCreateWebhook(cmd, projectVersion)
+	return cmd
+}
+
+func (c cli) bindCreateWebhook(cmd *cobra.Command, projectVersion string) {
+	ps, ok := c.plugins[projectVersion]
```
camilamacedo86: Have we not a global func to check the Project Version?
Also, I think it should return an error if the GetProjectVersion cannot be found.

pkg/cli/webhook.go
```diff
@@ -0,0 +1,97 @@
...
+			hasGetter = true
+			getter = tmpGetter
+		}
+	}
+	if !hasGetter {
+		err := fmt.Errorf("project version %q does not support a webhook creation plugin",
+			projectVersion)
+		cmdErr(cmd, err)
+		return
+	}
```
camilamacedo86: IHMO: From line 61 to 79 the impl is a little confused. 
Could we not check if already exist first and then, return the error?
and wdyt about if no errors were found then check if the plugin is supported?

pkg/plugin/internal/go_version.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"fmt"
+	"os/exec"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+func ValidateGoVersion() error {
+	err := fetchAndCheckGoVersion()
```
camilamacedo86: WDYT about one func for each action?
```go
func ValidateGoVersion() error {
     version, err := fetchGoVersion()
     if err != nil {
            return err
     }

     err := validteGoVersion(version)
     if err != nil {
            return err
     }

``

File: pkg/cli/cli.go [hunk 47]
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	rootCmd := c.defaultCommand()
+
+	rootCmd.AddCommand(
+		c.newInitProjectCmd(),
+		c.newCreateCmd(),
+	)
+
+	if internal.ConfiguredAndV1() {
```
camilamacedo86:
```suggestion
	if internal.IsConfigured && internal.IsV1() {
```
IHMO: Combine 2 or more actions in the same func may hurts concepts as cohesion and single responsibility. Also, I think is easier to read and understanding as suggested.

pkg/cli/cli.go
```diff
@@ -0,0 +1,273 @@
...
+	if internal.Configured() {
+		projectVersion, _ = getProjectVersion()
+	} else {
+		// If the project isn't configured, see if we can find the
+		// project version in a command line flag.
+		projectVersion, _ = c.getBaseFlags()
+	}
+
+	if projectVersion != "" {
+		ps, ok := c.plugins[projectVersion]
```
camilamacedo86: Could we centralize the GetProjectVersion and then, make it global to use in all places?

pkg/cli/create.go
```diff
@@ -0,0 +1,113 @@
...
+			getter = tmpGetter
+		}
+	}
+	if !hasGetter {
+		err := fmt.Errorf("project version %q does not support an API creation plugin",
+			projectVersion)
+		cmdErr(cmd, err)
+		return
+	}
+
```
camilamacedo86: The code in the line 77 to 95 is duplicated in many part of the impl. Could we not centralized it?
WDYT about something as validatePluginGetter ... such as following pseudo code that could be re-used?

```go 
func ValidatePluginGetter() error {
   err: = isDuplicated()
   if err != nil {
          return err
    }

   err: = isVersionSupported()
   if err != nil {
          return err
    }
}

```

pkg/cli/create.go
Adirio: `kubebuilder create webhook` needs to be added for any version except for v1.  `NOT (A & B) != (NOT A) & (NOT B)` so the logic was correct but your suggestion isn't. Your suggestion will only add the subcommand for not configured projects, and you can't use `kubebuilder create webhook` in a not configured projects, as it requires a `kubebuilder init` first.

pkg/cli/create.go
```diff
@@ -0,0 +1,113 @@
...
+		return cmd
+	}
+
+	// Lookup the plugin for projectVersion and bind it to the command.
+	c.bindCreateAPI(cmd, projectVersion)
+	return cmd
+}
+
+func (c cli) bindCreateAPI(cmd *cobra.Command, projectVersion string) { // nolint:dupl
+	ps, ok := c.plugins[projectVersion]
```
Adirio:
```suggestion
	plugins, ok := c.plugins[projectVersion]
```

Could we use a more descriptive name such as `plugins` or `pluginSet`? (I don't know where the `s` in `ps` comes from, but I would try to avoid including type information in variable names, so if it comes from `slice` I would just use `plugins`)

pkg/cli/cli.go
Adirio: That would require to read the config file twice.

pkg/cli/create.go
Adirio: They use different types (`plugin.CreateAPIPluginGetter`, `plugin.InitPluginGetter`, ...) and we can't pass types to functions so no, not possible.

pkg/cli/init_project.go
```diff
@@ -0,0 +1,135 @@
...
+		return cmd
+	}
+
+	// Lookup the plugin for projectVersion and bind it to the command.
+	c.bindInit(cmd, projectVersion)
+	return cmd
+}
+
+func (c cli) getInitHelpExamples() string {
+	vs := c.getAvailableProjectVersions()
```
Adirio:
```suggestion
	versions := c.getAvailableProjectVersions()
```

pkg/cli/init_project.go
```diff
@@ -0,0 +1,135 @@
...
+	var sb strings.Builder
+
+	for _, v := range vs {
```
Adirio:
```suggestion
	for _, version := range versions {
```

pkg/cli/init_project.go
```diff
@@ -0,0 +1,135 @@
...
+				continue
+			}
+			if _, ok := p.(plugin.Init); !ok {
+				continue
+			}
+			projectVersions = append(projectVersions, strconv.Quote(p.Version()))
+		}
+	}
+	return projectVersions
+}
```
Adirio: I think you should break from the internal loop once you add a version to the slice.
Why use `p.Version()` if the outter loop already has the version as key?
Additionally I suggest some new variable names to make them more descriptive:

```go
func (c cli) getAvailableProjectVersions() (versions []string) {
	for version, plugins := range c.plugins {
		for _, p := range plugins {
			// Only return project versions from non-deprecated plugins
			if _, deprecated := p.(plugin.Deprecated); deprecated {
				continue
			}
			if _, isInitPlugin := p.(plugin.Init); !isInitPlugin {
				continue
			}
			versions = append(versions, strconv.Quote(version))
			break
		}
	}
	return
}
```

Also, is `strconv.Quote` needed?

pkg/cli/init_project.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (c cli) bindInit(cmd *cobra.Command, projectVersion string) {
+	ps, ok := c.plugins[projectVersion]
```
Adirio:
```suggestion
	plugins, versionExists := c.plugins[projectVersion]
```

pkg/cli/init_project.go
```diff
@@ -0,0 +1,135 @@
...
+			hasGetter = true
+			getter = tmpGetter
+		}
+	}
+	if !hasGetter {
+		log.Fatalf("project version %q does not support a project initialization plugin",
+			projectVersion)
+	}
+
+	ip := getter.GetInitPlugin()
```
Adirio:
```suggestion
	initPlugin := getter.GetInitPlugin()
```

pkg/cli/create.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+	}
+	if !hasGetter {
+		err := fmt.Errorf("project version %q does not support an API creation plugin",
+			projectVersion)
+		cmdErr(cmd, err)
+		return
+	}
+
+	cap := getter.GetCreateAPIPlugin()
```
Adirio:
```suggestion
	createAPIPlugin := getter.GetCreateAPIPlugin()
```

pkg/cli/init_project.go
```diff
@@ -0,0 +1,135 @@
...
+		CommandName: c.commandName,
+	}
+	ip.UpdateContext(&ctx)
+	ip.SetVersion(projectVersion)
+	cmd.Long = ctx.Description
+	cmd.Example = ctx.Examples
+	cmd.RunE = func(cmd *cobra.Command, args []string) error {
+		if internal.Configured() {
+			return fmt.Errorf("failed to initialize project because project is already initialized")
+		}
```
Adirio: Shouldn't this check be left to each init command? Maybe in a future we want to allow `kubebuilder init` on previously configured projects to change some configuration setting or something.

Without this part `RunE` will be very similar for the init and the create api commands, and thus you could make a helper function.

pkg/cli/vendor_update.go
```diff
@@ -0,0 +1,47 @@
...
+			if err != nil {
+				log.Fatal(err)
+			}
+			if err = scaffold.NewUpdateScaffolder(cfg).Scaffold(); err != nil {
+				log.Fatal(err)
+			}
+			return nil
+		},
+	}
+}
```
Adirio: Why did you remove all the homogeneization recently done to this command.
- Errors were wrapped with `updateError`
- If `cfg, err := config.Read()` was returning a file-doesn't-exist error (`os.IsNotExist(err)`), a better error message was being yielded.
- You are not validating that it may only be run in `v1`.

I know that some of the steps of the generic plugins are no-ops for this case, but I still think that we should keep all of them homogenized.

pkg/cli/webhook.go
Adirio: The code is short-circuiting when it finds a duplicate. Iterate over the plugins, store the first into `getter` and set the `hasGetter` flag so that next time we try to do so we discover it is a duplciate. We keep iterating, and if we find another the flag is set so we don't need to keep iterating because we already know there is a duplicate. At the end we check that we found one. IMO this code is very clear.

pkg/cli/webhook.go
```diff
@@ -0,0 +1,97 @@
...
+
+	cwp := getter.GetCreateWebhookPlugin()
```
Adirio:
```suggestion
	createWebhookPlugin := getter.GetCreateWebhookPlugin()
```

pkg/cli/cli.go
```diff
@@ -0,0 +1,273 @@
...
+
+import (
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	internalconfig "sigs.k8s.io/kubebuilder/internal/config"
```
Adirio:
```suggestion
	"sigs.k8s.io/kubebuilder/internal/config"
```
I would only used the `internalconfig` alias when both `internal/config` and `pkg/model/config` are being imported in the same file.

pkg/cli/create.go
Adirio: I think you pointed to the wrong line or something. That replacement doesn't make sense.

pkg/cli/webhook.go
Adirio: `pkg/cli/internal/config.go` should be a nice place for this function.

Off-topic: @camilamacedo86 would be easier to follow the review if this suggestion was done where `getProjectVersion` is defined instead of in this lane that doesn't even call it.

pkg/cli/cli.go
Adirio: Duplicate of https://github.com/kubernetes-sigs/kubebuilder/pull/1290#discussion_r374980126, lets keep the discussion over there.

pkg/cli/cli.go
Adirio: `pkg/cli/internal/config.go` should be a nice place for this function as @camilamacedo86 suggested in other places.

I'm not very sure about the `bool` vs `error` topic, maybe leaning a bit towards `error` right now. Why did you choose `bool` @estroz?

pkg/cli/cli.go
Adirio:
```suggestion
		plugins, ok := c.plugins[projectVersion]
```

pkg/cli/cli.go
```diff
@@ -0,0 +1,273 @@
...
+		projectVersion, _ = c.getBaseFlags()
+	}
+
+	if projectVersion != "" {
+		ps, ok := c.plugins[projectVersion]
+		if !ok {
+			return fmt.Errorf("unknown project version %q", projectVersion)
+		}
+		for _, p := range ps {
+			if dp, ok := p.(plugin.Deprecated); ok {
```
Adirio:
```suggestion
			if deprecatedPlugin, isDeprecated := p.(plugin.Deprecated); isDeprecated {
```

pkg/cli/cli.go
```diff
@@ -0,0 +1,273 @@
...
+
+	if internal.ConfiguredAndV1() {
+		rootCmd.AddCommand(
+			c.newAlphaCmd(),
+			c.newVendorUpdateCmd(),
+		)
+	}
+
+	for _, cmd := range c.extraCommands {
+		for _, subc := range rootCmd.Commands() {
```
Adirio:
```suggestion
		for _, subCmd := range rootCmd.Commands() {
```

pkg/cli/cli.go
Adirio:
```suggestion
// getProjectVersion tries to load configuration file and returns ...
```
The configuration file being called `PROJECT` should be an implementation detail, let's try to avoid calling it PROJECT file in documentation.
The rest of the suggestion depends on the function signature that is being discussed elsewhere.

pkg/plugin/v2/init.go
```diff
@@ -0,0 +1,185 @@
...
+
+	return nil
+}
+
+func (p *initPlugin) GetScaffolder(c *config.Config) (scaffold.Scaffolder, error) { // nolint:unparam
+	return scaffold.NewInitScaffolder(c, p.license, p.owner), nil
+}
+
+func (p *initPlugin) PostScaffold(c *config.Config) error {
+	if (!p.depFlag.Changed && p.fetchDeps) || (p.depFlag.Changed && p.dep) {
```
Adirio: Nice catch, I missed that when refactoring, probably this should be fixed outside of the plugin implementation too.

pkg/plugin/v2/init.go
```diff
@@ -0,0 +1,185 @@
...
+		if err != nil {
+			return err
+		}
+
+		err = internal.RunCmd("Update go.mod", "go", "mod", "tidy")
+		if err != nil {
+			return err
+		}
+	} else {
+		fmt.Println("Skipping fetching dependencies.")
```
Adirio: `make` shouldn't be run if fetching dependencies is skipped, so you are missing a `return nil` here or move the make part inside the `if` block. The second option will still print the message when skipping dependencies.

pkg/plugin/v1/init.go
```diff
@@ -0,0 +1,190 @@
...
+func (p *initPlugin) Run() error {
+	return cmdutil.Run(p)
+}
+
+func (p *initPlugin) SetVersion(v string) {
+	if p.config == nil {
+		p.config = config.New(config.DefaultPath)
+	}
+	p.config.Version = v
+}
```
Adirio: Can `SetVersion` be called before`BindFlags`?
- If the answer is yes, `BindFlags` will make that call be useless as it will create a new config and won't keep the version, you should add a `p.config == nil` there too.
- If the answer is no, this check is useless as `BindFlags` already initializes this field.

Aplies to v2

estroz:
@Adirio @camilamacedo86 will get around to your comments in a bit. I'm still rebasing on top of the target branch, and want to wait until #1250 is merged to make any major updates.

pkg/plugin/v1/init.go
estroz: Order doesn't matter as long as `p.config` isn't nil, so I'll remove the check in `SetVersion`.

Adirio: It does matter as BindFlags is creating a new instance of Config, so the stored version is lost in case it is run before, and if it is run after BindFlags no need to check for nil as you already know it won't be nil.

pkg/cli/cli.go
```diff
@@ -0,0 +1,273 @@
...
+		c.newInitProjectCmd(),
+		c.newCreateCmd(),
+	)
+
+	if internal.ConfiguredAndV1() {
+		rootCmd.AddCommand(
+			c.newAlphaCmd(),
+			c.newVendorUpdateCmd(),
+		)
+	}
```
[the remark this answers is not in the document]
estroz: I can centralize in a follow-up commit once I address easier changes.

go.mod
```diff
@@ -3,18 +3,16 @@ module sigs.k8s.io/kubebuilder
 go 1.13
 
 require (
+	github.com/go-delve/delve v1.3.2
```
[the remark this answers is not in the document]
estroz: Leftover after running e2e tests, removing.

pkg/cli/vendor_update.go
estroz: Changes were lost in the rebase. Will update.

pkg/cli/init_project.go
estroz: I made an update to this method that got lost in the rebase. Will update.

pkg/cli/cli.go
```diff
@@ -0,0 +1,263 @@
...
+// Option is a function that can configure the cli
+type Option func(*cli) error
+
+// cli defines the command line structure and interfaces that are used to
+// scaffold kubebuilder project files.
+type cli struct {
+	// Base command name. Can be injected downstream.
+	commandName string
+	// Default project version. Used in CLI flag setup.
+	defaultProjectVersion string
```
Adirio: Where is this field being used? I think that `DefaultVersion` from `internal/config` may be enough.

pkg/plugin/v2/api.go
```diff
@@ -0,0 +1,203 @@
...
+
+	fs.BoolVar(&p.force, "force", false,
+		"attempt to create resource even if it already exists")
+	p.resource = &resource.Options{}
+	fs.StringVar(&p.resource.Kind, "kind", "", "resource Kind")
+	fs.StringVar(&p.resource.Group, "group", "", "resource Group")
+	fs.StringVar(&p.resource.Version, "version", "", "resource Version")
+	fs.BoolVar(&p.resource.Namespaced, "namespaced", true, "resource is namespaced")
+	fs.BoolVar(&p.resource.CreateExampleReconcileBody, "example", true,
+		"if true an example reconcile body should be written while scaffolding a resource.")
```
Adirio: `CreateExampleReconcileBody` is v1-specific

pkg/cli/init_project.go
Adirio: NOTE: this applies to `pkg/cli/api.go`, `pkg/cli/init.go` and `pkg/cli/webhook.go`.

Checking if the project is configured is the only difference between these three, which means that if we remove it we can make a helper that creates the RunE function for us similar to `errCmdFunc`. This check is already being done by the `LoadConfig` step of `RunOptions` (`internal/cmdutil/cmdutil.go:28`), so it is redundant.

pkg/cli/cli.go
estroz:
```go
// WithDefaultProjectVersion is an Option that sets the cli's default project
// version. Setting an unknown version will result in an error.
func WithDefaultProjectVersion(version string) Option {
	return func(c *cli) error {
		c.defaultProjectVersion = version
		return nil
	}
}
```
A downstream project may want to default to another project version.

Adirio: That is setting it, but is it used anywhere? The field is unexported

estroz: It allows projects using custom `init` logic to inject a default project version so they don't have to set `init --project-version $MY_DEFAULT` every time.

Adirio: My question is, where is that field used. You can set it with the exported function you mention, but then it is not used anywhere, is it? And being an unexported field it can't be used from outside the cli library.

cmd/main.go
```diff
@@ -19,115 +19,27 @@ package main
...
-	// Step 3: create scaffolder
-	scaffolder, err := options.scaffolder(projectConfig)
+func main() {
+	c, err := cli.New(
+		cli.WithPlugins(
+			&pluginv1.Plugin{},
+			&pluginv2.Plugin{},
+		),
+		cli.WithExtraCommands(
+			newEditCmd(),
```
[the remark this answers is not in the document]
estroz: There are two options here to avoid naming conflicts:
1. Create an `EditPlugin(Getter)` interface set.
1. Have `edit` implemented as a standalone subcommand added by `WithExtraCommands` in `cmd/`.

If we add `edit` directly in `buildRootCmd` we prevent downstream users from implementing their own `edit` command:

```go
func (c *cli) initialize() error {
	...
	// rootCmd will already contain any plugin and directly added
	// command names, so any conflicting names in c.extraCommands
	// will result in an error.
	rootCmd := c.buildRootCmd()
	for _, cmd := range c.extraCommands {
		for _, subCmd := range rootCmd.Commands() {
			if cmd.Name() == subCmd.Name() {
				return fmt.Errorf("command %q already exists", cmd.Name())
			}
		}
		rootCmd.AddCommand(cmd)
	}
	...
}
```

Generally any command that we don't want to reserve for kubebuilder alone (so pretty much every command) should not be added directly _without a plugin interface_ to `buildRootCmd`. Currently `update` does this, and should be refactored using one of the 2 methods above.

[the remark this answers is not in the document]
Adirio: I think update is something too specific (and deprecated) so let it as is. Edit, however, should be open for other plugins to use. Thats my opinion.

pkg/cli/cli.go
estroz: It is used to define the `project-version` default in `newInitCmd`:
```go
	cmd.Flags().String("project-version", c.defaultProjectVersion,
		fmt.Sprintf("project version, possible values: (%s)", strings.Join(c.getAvailableProjectVersions(), ", ")))
```

cmd/main.go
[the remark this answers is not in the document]
estroz: @DirectXMan12 @mengqiy thoughts on `update`?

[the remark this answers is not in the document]
Adirio: I saw your last commit and don't know if I was clear enough with my last comment. I would leave update as a extra command but I would provide an interface to edit as it whould allow plugins have their own edit subcommands. However, I would leave all commands in either `cli` os `plugin` packages, not in `cmd`.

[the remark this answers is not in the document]
estroz: I moved update to `cmd` so we can merge things as-is without inciting unnecessary conflict while we decide how to handle these commands. It would be best to do so in a follow-up IMO.

pkg/cli/internal/config.go
```diff
@@ -17,33 +17,13 @@ limitations under the License.
...
-	}
-
-	if err != nil {
-		log.Fatalf("failed to read the configuration file: %v", err)
-	}
-
-	return projectConfig.IsV1()
-}
-
-func PrintV1DeprecationWarning() {
```
camilamacedo86: Are we removing the notice? I think we should keep.

estroz:
Bits of the #1250 proposal to implement in follow-ups:
- Version format enforcement
- Fully-qualified plugin name logic

Other work
- `edit`/`update` commands: should these be plugin types or extra commands?

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+type Base interface {
+	// Name returns a name defining the plugin type.
+	// For example, Kubebuilder's plugins would return "go".
+	Name() string
+	// Version returns the project version that this plugin implements.
+	// For example, Kubebuilder's Go v2 plugin implementation would return 2.
+	Version() string
+	SupportedProjectVersions() []string
```
camilamacedo86: Nit: missing comment with eg on top.

pkg/cli/webhook.go
estroz: Will refactor in a follow-up if possible.

pkg/cli/cli.go
Adirio: That was the part I was missing, mark this as resolved

pkg/plugin/v1/webhook.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (p *createWebhookPlugin) Validate(_ *config.Config) error {
+	if err := p.resource.Validate(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (p *createWebhookPlugin) GetScaffolder(c *config.Config) (scaffold.Scaffolder, error) { // nolint:unparam
```
camilamacedo86: // The uparam reports that it is unused. 
Would be great add a comment as fixme: remove the // nolint:unparam when the layout implementation be made and it is start to be used.

pkg/plugin/v2/init.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/spf13/pflag"
+
+	"sigs.k8s.io/kubebuilder/internal/cmdutil"
+	"sigs.k8s.io/kubebuilder/internal/config"
+	"sigs.k8s.io/kubebuilder/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/pkg/plugin/internal"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold"
+)
+
+type initPlugin struct { // nolint:maligned
```
camilamacedo86:
```suggestion
type initPlugin struct { // nolint:maligned
```
 it is OK for me. When we remove the deprecated flags it will be solved.

cmd/edit.go
```diff
@@ -89,10 +83,10 @@ func (o *editOptions) validate(c *config.Config) error {
 	return nil
 }
 
-func (o *editOptions) scaffolder(c *config.Config) (scaffold.Scaffolder, error) { // nolint:unparam
+func (o *editOptions) GetScaffolder(c *config.Config) (scaffold.Scaffolder, error) { // nolint:unparam
```
camilamacedo86:
```suggestion
// todo: remove the nolint:unparam when it started to be used. 
func (o *editOptions) GetScaffolder(c *config.Config) (scaffold.Scaffolder, error) { // nolint:unparam
```

cmd/update.go
```diff
@@ -39,52 +38,40 @@ func (e updateError) Error() string {
...
 	if !c.IsV1() {
-		return fmt.Errorf("vendor was only used for v1, this project is %s", c.Version)
+		return fmt.Errorf("vendor was only used in project version 1, this project is %s", c.Version)
 	}
 
 	return nil
 }
 
-func (o *updateOptions) scaffolder(c *config.Config) (scaffold.Scaffolder, error) { // nolint:unparam
+func (o *updateOptions) GetScaffolder(c *config.Config) (scaffold.Scaffolder, error) { // nolint:unparam
```
camilamacedo86:
```suggestion



// todo: remove the nolint:unparam when it started to be used.
func (o *updateOptions) GetScaffolder(c *config.Config) (scaffold.Scaffolder, error) { // nolint:unparam
```

pkg/plugin/v1/webhook.go
Adirio: `uparam` in this case is reporting that the error is always `nil`, and it suggest removing the second output param. But it can't be removed as it is a method of an interface and other implementations of that interface do return a non-nil error.

cmd/edit.go
Adirio: See https://github.com/kubernetes-sigs/kubebuilder/pull/1290#discussion_r376382422

cmd/update.go
Adirio: See https://github.com/kubernetes-sigs/kubebuilder/pull/1290#discussion_r376382422

pkg/cli/api.go
```diff
@@ -0,0 +1,89 @@
...
+	createAPI.BindFlags(cmd.Flags())
+	// TODO: inject defaults.
+	ctx := plugin.Context{
+		CommandName: c.commandName,
+	}
+	createAPI.UpdateContext(&ctx)
+	cmd.Long = ctx.Description
+	cmd.Example = ctx.Examples
+	cmd.RunE = runECmdFunc(createAPI, fmt.Sprintf("failed to create api for project with version %q", c.projectVersion))
+}
```
camilamacedo86: Suggestion to centralize duplicated code:

```go
func (c cli) bindCreateAPI(cmd *cobra.Command) {

	if err:= c.checkPlugin(); err != nil {
		cmdErr(cmd, err)
	}

	var getter plugin.CreateAPIPluginGetter
	createAPI := getter.GetCreateAPIPlugin()
	createAPI.BindFlags(cmd.Flags())

	// TODO: inject defaults.
	ctx := plugin.Context{
		CommandName: c.commandName,
	}

	createAPI.UpdateContext(&ctx)
	cmd.Long = ctx.Description
	cmd.Example = ctx.Examples
	cmd.RunE = runECmdFunc(createAPI, fmt.Sprintf("failed to create api for project with version %q", c.projectVersion))
}

```

pkg/cli/cli.go
```diff
@@ -0,0 +1,302 @@
...
+	if c.projectVersion == "" {
+		return nil, errors.New("project version not set")
+	}
+	versionedPlugins, versionFound := c.plugins[c.projectVersion]
+	if !versionFound {
+		return nil, fmt.Errorf("unknown project version %q", c.projectVersion)
+	}
+	return versionedPlugins, nil
+}
+
```
camilamacedo86: Suggestion to centralize the duplicate code: 

```go
// checkPlugin will ensure that the plugin can be executed
func (c cli) checkPlugin() error {
	var getter plugin.CreateWebhookPluginGetter
	versionedPlugins, err := c.getVersionedPlugins()
	if err != nil {
		return err
	}
	var hasGetter bool
	for _, p := range versionedPlugins {
		tmpGetter, isGetter := p.(plugin.CreateWebhookPluginGetter)
		if isGetter {
			if hasGetter {
				err := fmt.Errorf("The command %s was alredy executed. This plugin is duplicated for the project version %q: %s, %s",
					c.commandName, c.projectVersion, getter.Name(), p.Name())
				return err
			}
			hasGetter = true
			getter = tmpGetter
		}
	}
	if !hasGetter {
		err := fmt.Errorf("project version %q do not support the command %s",
			c.projectVersion, c.commandName)
		return err
	}
	return nil
}
```

pkg/cli/webhook.go
```diff
@@ -0,0 +1,90 @@
...
+	// TODO: inject defaults.
+	ctx := plugin.Context{
+		CommandName: c.commandName,
+	}
+	createWebhook.UpdateContext(&ctx)
+	cmd.Long = ctx.Description
+	cmd.Example = ctx.Examples
+	cmd.RunE = runECmdFunc(createWebhook,
+		fmt.Sprintf("failed to create webhook for project with version %q", c.projectVersion))
+}
```
camilamacedo86: Suggestion to centralize the duplicated code:

```go
func (c cli) bindCreateWebhook(cmd *cobra.Command) {
	if err:= c.checkPlugin(); err != nil {
		cmdErr(cmd, err)
	}

	var getter plugin.CreateWebhookPluginGetter
	createWebhook := getter.GetCreateWebhookPlugin()
	createWebhook.BindFlags(cmd.Flags())
	// TODO: inject defaults.
	ctx := plugin.Context{
		CommandName: c.commandName,
	}
	createWebhook.UpdateContext(&ctx)
	cmd.Long = ctx.Description
	cmd.Example = ctx.Examples
	cmd.RunE = runECmdFunc(createWebhook,
		fmt.Sprintf("failed to create webhook for project with version %q", c.projectVersion))
}
```

pkg/cli/init.go
```diff
@@ -0,0 +1,150 @@
...
+	// TODO: inject defaults.
+	ctx := plugin.Context{
+		CommandName: c.commandName,
+	}
+	init.UpdateContext(&ctx)
+	init.SetVersion(c.projectVersion)
+	cmd.Long = ctx.Description
+	cmd.Example = ctx.Examples
+	cmd.RunE = runECmdFunc(init, fmt.Sprintf("failed to initialize project with version %q", c.projectVersion))
+}
```
camilamacedo86: Suggestion to centralize the duplicate code:


```go
func (c cli) bindInit(cmd *cobra.Command) {
	if err:= c.checkPlugin(); err != nil {
		cmdErr(cmd, err)
	}
	
	var getter plugin.InitPluginGetter
	init := getter.GetInitPlugin()
	init.BindFlags(cmd.Flags())
	
	// TODO: inject defaults.
	ctx := plugin.Context{
		CommandName: c.commandName,
	}
	init.UpdateContext(&ctx)
	init.SetVersion(c.projectVersion)
	cmd.Long = ctx.Description
	cmd.Example = ctx.Examples
	cmd.RunE = runECmdFunc(init, fmt.Sprintf("failed to initialize project with version %q", c.projectVersion))
}
```

pkg/cli/cli.go
```diff
@@ -0,0 +1,302 @@
...
+// errCmdFunc returns a cobra RunE function that returns the provided error
+func errCmdFunc(err error) func(*cobra.Command, []string) error {
+	return func(*cobra.Command, []string) error {
+		return err
+	}
+}
+
+// runECmdFunc returns a cobra RunE function that runs gsub and returns its value.
```
camilamacedo86: To fix the interface. 

```suggestion
// runECmdFunc returns a cobra RunE function that runs gsub and returns its value.
func (c cli) runECmdFunc(msg string) func(*cobra.Command, []string) error {
	return func(*cobra.Command, []string) error {
		if err := c.Run(); err != nil {
			return fmt.Errorf("%s: %v", msg, err)
		}
		return nil
	}
}
```

pkg/cli/cli.go
```diff
@@ -0,0 +1,305 @@
...
+		c.configured = false
+		c.projectVersion = c.defaultProjectVersion
+	} else if err == nil {
+		c.configured = true
+		c.projectVersion = projectConfig.Version
+	} else {
+		return fmt.Errorf("failed to read config: %v", err)
+	}
+
+	rootCmd := c.buildRootCmd()
```
Adirio:
```suggestion
	c.cmd := c.buildRootCmd()
```

pkg/cli/cli.go
```diff
@@ -0,0 +1,305 @@
...
+
+	// Add extra commands injected by options.
+	for _, cmd := range c.extraCommands {
+		for _, subCmd := range rootCmd.Commands() {
+			if cmd.Name() == subCmd.Name() {
+				return fmt.Errorf("command %q already exists", cmd.Name())
+			}
+		}
+		rootCmd.AddCommand(cmd)
```
Adirio:
```suggestion
		c.cmd.AddCommand(cmd)
```

pkg/cli/cli.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+		for _, p := range versionedPlugins {
+			if d, isDeprecated := p.(plugin.Deprecated); isDeprecated {
+				fmt.Printf(noticeColor, fmt.Sprintf("[Deprecation Notice] %s\n\n",
+					d.DeprecationWarning()))
+			}
+		}
+	}
+
+	c.cmd = rootCmd
```
Adirio:
```suggestion
```

pkg/cli/create.go
```diff
@@ -14,16 +14,17 @@ See the License for the specific language governing permissions and
...
+package cli
 
 import (
 	"github.com/spf13/cobra"
 )
 
-func newCreateCmd() *cobra.Command {
-	return &cobra.Command{
+func (c *cli) newCreateCmd() *cobra.Command {
+	cmd := &cobra.Command{
```
Adirio:
```suggestion
	return &cobra.Command{
```

pkg/cli/create.go
```diff
@@ -14,16 +14,17 @@ See the License for the specific language governing permissions and
...
 		Use:   "create",
-		Short: "Scaffold a Kubernetes API or webhook.",
+		Short: "Scaffold a Kubernetes API or webhook",
 		Long:  `Scaffold a Kubernetes API or webhook.`,
 	}
+	return cmd
```
Adirio:
```suggestion
```

cmd/main.go
```diff
@@ -17,117 +17,41 @@ limitations under the License.
...
-	// kubebuilder version
-	rootCmd.AddCommand(version.NewVersionCmd())
-
-	return rootCmd
 }
 
-func main() {
-	if err := buildCmdTree().Execute(); err != nil {
-		log.Fatal(err)
+func loadExistingConfig() (*config.Config, error) {
```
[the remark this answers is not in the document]
estroz: `Must` implies that the function exits if the operation fails. Is this the behavior you intend `MustLoad` to have? Otherwise `LoadExisting` makes more sense.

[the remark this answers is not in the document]
Adirio: Must usually means that errors will be handled with a panic, I guess another name could be better. But moving it there seems like a good thing to do.

camilamacedo86:
@estroz @DirectXMan12. 

Should not it be an issue in https://github.com/kubernetes-sigs/kubebuilder/projects/7? as well?
WDYT?
c/c @hasbro17

cmd/main.go
```diff
@@ -19,115 +19,28 @@ package main
...
 	"log"
 
-	"github.com/spf13/cobra"
-
-	"sigs.k8s.io/kubebuilder/cmd/internal"
 	"sigs.k8s.io/kubebuilder/cmd/version"
-	"sigs.k8s.io/kubebuilder/internal/config"
-	"sigs.k8s.io/kubebuilder/pkg/scaffold"
+	"sigs.k8s.io/kubebuilder/pkg/cli"
+	pluginv1 "sigs.k8s.io/kubebuilder/pkg/plugin/v1"
```
camilamacedo86: We should not care about V1 since it is deprecated and should no longer be used.
Also, IHMO we need to remove the V1 code in order to kept the thing simple. 
However, since it is already done in the PR we can merge with.

pkg/cli/cli.go
```diff
@@ -0,0 +1,304 @@
...
+	projectConfig, err := internalconfig.Read()
+	if os.IsNotExist(err) {
+		c.configured = false
+		c.projectVersion = c.defaultProjectVersion
+	} else if err == nil {
+		c.configured = true
+		c.projectVersion = projectConfig.Version
+	} else {
+		return fmt.Errorf("failed to read config: %v", err)
+	}
```
camilamacedo86: Just a nit. Suggestion ( I think make more clear when has an error and not what should be done )

```go
if err != nil {
      if os.IsNotExist(err) {
            c.configured = false
	    c.projectVersion = c.defaultProjectVersion
      } else {
           return fmt.Errorf("failed to read config: %v", err)
      }
} else {
     c.configured = true
     c.projectVersion = projectConfig.Version
}
```

pkg/cli/api.go
```diff
@@ -0,0 +1,90 @@
...
+		cmdErr(cmd, err)
+		return
+	}
+	var getter plugin.CreateAPIPluginGetter
+	var hasGetter bool
+	for _, p := range versionedPlugins {
+		tmpGetter, isGetter := p.(plugin.CreateAPIPluginGetter)
+		if isGetter {
+			if hasGetter {
+				err := fmt.Errorf("duplicate API creation plugins for project version %q: %s, %s",
```
camilamacedo86: Could not the API be called/used more than once to create different resouces?

pkg/cli/webhook.go
```diff
@@ -0,0 +1,91 @@
...
+		cmdErr(cmd, err)
+		return
+	}
+	var getter plugin.CreateWebhookPluginGetter
+	var hasGetter bool
+	for _, p := range versionedPlugins {
+		tmpGetter, isGetter := p.(plugin.CreateWebhookPluginGetter)
+		if isGetter {
+			if hasGetter {
+				err := fmt.Errorf("duplicate webhook creation plugins for project version %q: %s, %s",
```
camilamacedo86: Can we just use/call the webhook just once as well?

Adirio:
> However, I have not seen the prow jobs called here. So, /lgtm cancel
> 
> Could you please check?

This is being merged in a branch that is not `Master` and thus the tests are not being run.

pkg/cli/api.go
estroz: Each `create` command is associated with exactly one versioned plugin, otherwise plugin behavior can conflict. If you want to create more than one resource, call `create api` for each resource.

pkg/cli/cli.go
estroz: This is less clear IMO.

pkg/cli/webhook.go
estroz: See my `create api` comment.

cmd/main.go
estroz: Removing v1 may be easier after merging the feature branch with master. @DirectXMan12 @mengqiy thoughts on this? Regardless, out of scope for this PR.

pkg/cli/alpha.go
```diff
@@ -14,16 +14,22 @@ See the License for the specific language governing permissions and
...
 
-func newAlphaCmd() *cobra.Command {
+func (c *cli) newAlphaCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:   "alpha",
 		Short: "Expose commands which are in experimental or early stages of development",
 		Long:  `Command group for commands which are either experimental or in early stages of development`,
+		Example: fmt.Sprintf(`
+# scaffolds webhook server
+%s alpha webhook <params>`,
```
DirectXMan12: opposite side of same question here

pkg/cli/api.go
```diff
@@ -0,0 +1,90 @@
...
+		Use:     "api",
+		Short:   "Scaffold a Kubernetes API",
+		Long:    ctx.Description,
+		Example: ctx.Examples,
+		RunE: errCmdFunc(
+			fmt.Errorf("api subcommand requires an existing project"),
+		),
+	}
+
+	// Lookup the plugin for projectVersion and bind it to the command.
```
DirectXMan12: is this comment correct any more?

cmd/main.go
camilamacedo86: Just to track here: We have a PR for that already:  https://github.com/kubernetes-sigs/kubebuilder/pull/1355

PR #1187, opened 2019-11-11
Imports ordering

Adirio:
Imports have been grouped and sorted in the whole project. Two different schemas have been followed.

For kubebuilder source files, the imports have been split into three groups according to the current policy: standard library are placed first, followed by packages non related to k8s and at the end k8s packages.
```
import (
    std

    non-k8s

    k8s
)
```

Scaffolded files use the schema that go formatting tools (`golang.org/x/tools/imports`) implement: standard library first, followed by external packages and at the bottom imports from the current repository.
```
import (
    std

    external

    repository
)
```

Template strings for scaffolded files need to keep imports in a single group and `imports.Process` wil handle the divission.

### References:
- [Imports](https://github.com/golang/go/wiki/CodeReviewComments#imports)

Note: despite being the de facto standard, couldn't find a CodeReviewComments or Effective Go section that talks about the order of imports in each group, but you can see in the example how this rule is being followed.

Closes: #1183

/kind cleanup

Issue #1183, opened 2019-11-11
Imports ordering


Adirio:
### Linter message:
> Unsorted imports inspection

### Linter description:
> Reports unsorted imports.

### References:
- [Imports](https://github.com/golang/go/wiki/CodeReviewComments#imports)

Note: despite being the de facto standard, couldn't find a CodeReviewComments or Effective Go section that talks about the order of imports in each group, but you can see in the example how this rule is being followed.

Umbrella issue: #1176

/kind cleanup

camilamacedo86:
HI, could you please assign the issue to you in order to avoid other dev get it since you gave a PR already one for this one? Also, devs will usually filter for the unassigned ones to know what they can do to contribute with. 

PS.: To do it just add a new comment with `/assign @your user`

PR #1292, opened 2020-01-09
:running: Enable the lint check goimports 

camilamacedo86:
**Description**
Enable the lint check goimports
Note that nothing needed to be fixed in this case.

### DIFF
--- a/scripts/verify.sh
+++ b/scripts/verify.sh
@@ -38,12 +38,13 @@ golangci-lint run --disable-all \
     --enable=nakedret \
     --enable=interfacer \
     --enable=misspell \
-    --enable=dupl
+    --enable=dupl \
+    --enable=goimports \
+
 
 ##todo(camilamacedo86): The following checks requires fixes in the code
 # --enable=golint
 # --enable=gocyclo
-# --enable=goimports
 # --enable=lll
 # --enable=goconst
 # --enable=gosec


scripts/verify.sh
```diff
@@ -38,15 +38,15 @@ golangci-lint run --disable-all \
...
 ##todo(camilamacedo86): The following checks requires fixes in the code
 # --enable=golint
 # --enable=gocyclo
-# --enable=goimports
 # --enable=lll
 # --enable=goconst
 # --enable=gosec
 # --enable=maligned
 
-GO111MODULES=on go list -mod=readonly ./...
```
estroz: Does `golangci-lint run --enable=goimports` supersede this check?

camilamacedo86: You check to fast before my fix pr :-) 
Tks for the help .

testdata/project-v2/Makefile
```diff
@@ -52,7 +52,7 @@ vet:
 
 # Generate code
 generate: controller-gen
-	$(CONTROLLER_GEN) object:headerFile=./hack/boilerplate.go.txt paths="./..."
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
```
estroz: Is this change related to enabling the goimports linter?

camilamacedo86: To do the PR's we need to run `make generate` which will update the testdata.
Some PR was merged in the master without this update. (may the PR passed in the CI and was merged after the change which affects this artefact)

So, until it is merged (ps.: has a PR already for that  + many ones :-P ) you will see that in all diffs since whiteout it is not possible to check the CI. Mainly for we allow do the reviews.

Adirio: #1288 fixes it in a separate PR and is tracking all the affected PRs.

Adirio:
LGTM after #1288 gets merged. Directly related to #1183 and #1187.

PR #1293, opened 2020-01-09
enable maligned lint check

camilamacedo86:
**Description**
- enable maligned lint check
- perform code fixes

### DIFF
--- a/cmd/init_project.go
+++ b/cmd/init_project.go
@@ -68,21 +68,22 @@ kubebuilder init --domain example.org --license apache2 --owner "The Kubernetes 
 }
 
 type projectOptions struct {
+	boilerplate project.Boilerplate
+	project     project.Project
+
+	// final result
+	scaffolder scaffold.ProjectScaffolder
+
+	// deprecated flags
+	depFlag *flag.Flag
+	depArgs []string
 
 	// flags
 	fetchDeps          bool
 	skipGoVersionCheck bool
 
-	boilerplate project.Boilerplate
-	project     project.Project
-
 	// deprecated flags
-	dep     bool
-	depFlag *flag.Flag
-	depArgs []string
-
-	// final result
-	scaffolder scaffold.ProjectScaffolder
+	dep bool
 }
 
 func (o *projectOptions) bindCmdlineFlags(cmd *cobra.Command) {
--- a/pkg/scaffold/resource/resource.go
+++ b/pkg/scaffold/resource/resource.go
@@ -26,9 +26,6 @@ import (
 
 // Resource contains the information required to scaffold files for a resource.
 type Resource struct {
-	// Namespaced is true if the resource is namespaced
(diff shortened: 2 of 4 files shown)


testdata/project-v2/Makefile
```diff
@@ -52,7 +52,7 @@ vet:
 
 # Generate code
 generate: controller-gen
-	$(CONTROLLER_GEN) object:headerFile=./hack/boilerplate.go.txt paths="./..."
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
```
Adirio: Fixed and tracked by #1288

cmd/init_project.go
```diff
@@ -68,21 +68,23 @@ kubebuilder init --domain example.org --license apache2 --owner "The Kubernetes
...
 	depArgs []string
 
-	// final result
-	scaffolder scaffold.ProjectScaffolder
+	// flags
+	fetchDeps          bool
+	skipGoVersionCheck bool
+
+	// @Deprecated
+	dep bool
```
Adirio: Why split this from the other deprecated flags?

camilamacedo86: In order to solve the `maligned` issue check, I need to to add the bools at the end. 
This is about memory space usage.

Adirio:
LGTM from a code perspective.

The benefit from adding maligned check to kubebuilder isn't that big as most of the types, if not all of them, will only have 1 or 2 instances, so the memory reduction will be in the order of bytes. Even if small, it is still a benefit and if it doens't introduce any harm it is an enhancement.

cmd/init_project.go
```diff
@@ -68,21 +68,23 @@ kubebuilder init --domain example.org --license apache2 --owner "The Kubernetes
...
+	// @Deprecated
 	depArgs []string
 
-	// final result
-	scaffolder scaffold.ProjectScaffolder
+	// flags
+	fetchDeps          bool
+	skipGoVersionCheck bool
+
+	// @Deprecated
```
mengqiy: > @Deprecated

Is it required to use this format?
Can you please point me to the document if any?

camilamacedo86: No. The check has no relation with the comment.
I will replace for the previous comment since shows that you prefer.

camilamacedo86: Done.

camilamacedo86:
@mengqiy could you please re-apply the /lgtm label here?

mengqiy:
@camilamacedo86 It needs rebase

camilamacedo86:
Hi @mengqiy tks for the review. 
Rebased just missing 
/lgtm

Adirio:
@camilamacedo86 Travis should just require to be re-triggered.

mengqiy:
Need rebase again...

PR #1296, opened 2020-01-10
Update OWNERS_ALIASES

gerred:
Active on controller runtime, want to make sure I'm active here too. Going for the bold move of maintainer - happy to change this to reviewer.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -9,8 +9,8 @@ aliases:
     - Liujingfang1
     - mengqiy
     - pmorie
-    - gerred
   kubebuilder-reviewers:
     - camilamacedo86
     - joelanford
     - estroz
+    - gerred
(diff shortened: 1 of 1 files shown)


Adirio:
Travis fail is due to a bug introduced by #1016 that is fixed in #1288.

gerred:
Thanks @Adirio - stuck between a rock and a hard place. I'll review #1288 in the meantime.

gerred:
Thanks @camilamacedo86 :+1: - I will adjust.

DirectXMan12:
needs rebase to merge properly otherwise fine I think

DirectXMan12:
@gerred I'd also prefer to see a few more reviews here first, if you can
/hold

gerred:
@DirectXMan12 Absolutely :+1:. What wasn't clear when I opened this was if we were grouping these all across the entire subproject rather than on a per-repo basis, and based on what we talked about, it makes much more sense about having a consistent set of reviewers/approvers per project. I think there's enough to do over in CR that I'll focus there for now, and if my help can be useful over on this side and in CT, I can always open this PR again.

I'll close for now. :)

PR #1304, opened 2020-01-12
enable gosec lint check

camilamacedo86:
**Description**
enable gosec lint check

### DIFF
--- a/test/e2e/utils/test_context.go
+++ b/test/e2e/utils/test_context.go
@@ -161,6 +161,7 @@ func (kc *KBTestContext) Make(makeOptions ...string) error {
 
 // CleanupImage is for cleaning up the docker images for testing
 func (kc *KBTestContext) Destroy() {
+	// nolint:gosec
 	cmd := exec.Command("docker", "rmi", "-f", kc.ImageName)
 	if _, err := kc.Run(cmd); err != nil {
 		fmt.Fprintf(GinkgoWriter, "warning: error when removing the local image: %v\n", err)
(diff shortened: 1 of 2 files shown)


test/e2e/utils/test_context.go
```diff
@@ -160,6 +160,8 @@ func (kc *KBTestContext) Make(makeOptions ...string) error {
 
 // CleanupImage is for cleaning up the docker images for testing
 func (kc *KBTestContext) Destroy() {
+	// nolint:gosec
+	// False positive: G204: Subprocess launched with function call as argument or cmd arguments (gosec)
```
mengqiy: I'd drop this line, since it's not accurate. It is certainly risky to launch a subprocess. 
What if the docker binary has been replaced by an attacker somehow?

camilamacedo86: Yes it true. It is not a false positive at all. However, I am not sure how we could fix it.
Have you any idea?

mengqiy: I mean we only keep `// nolint:gosec` which is necessary.

test/e2e/utils/test_context.go
```diff
@@ -161,6 +161,8 @@ func (kc *KBTestContext) Make(makeOptions ...string) error {
 
 // CleanupImage is for cleaning up the docker images for testing
 func (kc *KBTestContext) Destroy() {
+	// nolint:gosec
+	// G204: Subprocess launched with function call as argument or cmd arguments (gosec)
```
mengqiy: We should keep `// nolint:gosec`
And drop `// G204: Subprocess launched with function call as argument or cmd arguments (gosec)`

PR #1310, opened 2020-01-14
Delegate go vet and go fmt to golangci-lint

Adirio:
`go vet` and `go fmt` were being called before `golangci-lint`. The later is able to run those two checks and that will benefit from the speed enhancements golangci-lint does (concurrency, evaluating files only once, building shared caches, etc.)

### DIFF
--- a/scripts/verify.sh
+++ b/scripts/verify.sh
@@ -14,20 +14,14 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-set -e
-
 source ./common.sh
-
-header_text "running go vet"
-go vet ${MOD_OPT} ../...
-
-header_text "running go fmt"
-go fmt ${MOD_OPT} ../...
 
 header_text "running golangci-lint"
 cd .. # To go to the root of the project
 golangci-lint run --disable-all \
     --deadline 5m \
+    --enable=govet \
+    --enable=gofmt \
     --enable=misspell \
     --enable=structcheck \
     --enable=deadcode \
@@ -37,7 +31,6 @@ golangci-lint run --disable-all \
     --enable=ineffassign \
     --enable=nakedret \
     --enable=interfacer \
-    --enable=misspell \
     --enable=dupl \
     --enable=goimports \
     --enable=gocyclo \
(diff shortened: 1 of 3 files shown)


Adirio:
@camilamacedo86 Thoughts on this?

scripts/verify.sh
```diff
@@ -14,20 +14,14 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-set -e
```
estroz: Any reason for removing this?

scripts/verify.sh
```diff
@@ -37,18 +31,14 @@ golangci-lint run --disable-all \
...
 
 ##todo(camilamacedo86): The following checks requires fixes in the code
 # --enable=golint
-# --enable=gocyclo
 # --enable=lll
 # --enable=goconst
 # --enable=gosec
 # --enable=maligned
-
-GO111MODULES=on go list -mod=readonly ./...
```
estroz: Any reason for removing this?

scripts/verify.sh
```diff
@@ -37,18 +31,14 @@ golangci-lint run --disable-all \
     --enable=ineffassign \
     --enable=nakedret \
     --enable=interfacer \
-    --enable=misspell \
```
estroz: Any reason for removing this?

camilamacedo86: duplicated :-)

scripts/verify.sh
camilamacedo86: has no need .

scripts/verify.sh
Adirio: Set e makes it return when first fails but there is only 1 test from bash perspective, right?

scripts/verify.sh
Adirio: It is just lisring the modules in Travis, not needed, just consumes resources (I know, its negligible but...)

scripts/verify.sh
Adirio: Exactly, it's a duplicate

mengqiy: I'd suggest always sort this list, so dup will be very unlikely.

Adirio: Should we wait until all the Golint related PR are merged?

mengqiy: That is fine as long as we don't forget to do it.
We can probably track it in the umbrella issue.

PR #1313, opened 2020-01-16
Include PROJECT file information into model.Universe

Adirio:
### Description

Include the PROJECT file into `model.Universe`.

It also provides a constructor for `model.Universe` called `model.NewUniverse` that uses the functional options paradigm to provide optional configuration.

### Motivation

The main motivation is providing plugins access to the configuration file. Additionally, `scaffold.Execute` needs some information aside from the files it scaffolds. Right now this informnation is provided in three different ways. As fields to the `Scaffold` instance, the first parameter to `Scaffold.Execute` (`model.Universe`), and the second parameter to `Scaffold.Execute` (`input.Options`). This change is a first step into providing a single way to provide that information.

/kind feature

### DIFF
--- a/cmd/vendor_update.go
+++ b/cmd/vendor_update.go
@@ -39,13 +39,21 @@ kubebuilder update vendor
 		Run: func(cmd *cobra.Command, args []string) {
 			internal.DieIfNotConfigured()
 
-			err := (&scaffold.Scaffold{}).Execute(
-				&model.Universe{},
+			universe, err := model.NewUniverse(
+				model.WithConfigFrom("PROJECT"),
+				model.WithoutBoilerplate,
+			)
+			if err != nil {
+				log.Fatalf("error updating vendor dependencies: %v", err)
+			}
+
+			err = (&scaffold.Scaffold{}).Execute(
+				universe,
 				input.Options{},
 				&project.GopkgToml{},
 			)
 			if err != nil {
-				log.Fatalf("error updating vendor dependecies %v", err)
+				log.Fatalf("error updating vendor dependencies: %v", err)
 			}
 		},
 	}
--- a/pkg/model/universe.go
+++ b/pkg/model/universe.go
@@ -1,0 +1,133 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package model
+
+import (
+	"io/ioutil"
+	"strings"
+
+	"github.com/gobuffalo/flect"
+
+	internalconfig "sigs.k8s.io/kubebuilder/internal/config"
+	"sigs.k8s.io/kubebuilder/pkg/model/config"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/resource"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/util"
+)
+
+// Universe describes the entire state of file generation
+type Universe struct {
+	// Config stores the project configuration
+	Config *config.Config `json:"config,omitempty"`
+
+	// Boilerplate is the copyright comment added at the top of scaffolded files
+	Boilerplate string `json:"boilerplate,omitempty"`
+
+	// Resource contains the information of the API that is being scaffolded
+	Resource *Resource `json:"resource,omitempty"`
+
+	// Files contains the model of the files that are being scaffolded
+	Files []*File `json:"files,omitempty"`
+}
+
+// NewUniverse creates a new Universe
+func NewUniverse(options ...UniverseOption) (*Universe, error) {
+	universe := &Universe{}
+
+	// Apply options
+	for _, option := range options {
+		if err := option(universe); err != nil {
+			return nil, err
+		}
+	}
+
+	return universe, nil
+}
+
+// UniverseOption configure Universe
+type UniverseOption func(*Universe) error
+
+// WithConfigFrom loads the project configuration from the provided path
+func WithConfigFrom(path string) UniverseOption {
+	return func(universe *Universe) error {
+		projectConfig, err := internalconfig.ReadFrom(path)
+		if err != nil {
+			return err
+		}
+
+		universe.Config = projectConfig
+		return nil
+	}
+}
+
+// WithConfig stores the already loaded project configuration
+func WithConfig(projectConfig *config.Config) UniverseOption {
+	return func(universe *Universe) error {
+		universe.Config = projectConfig
+		return nil
+	}
+}
+
+// WithBoilerplateFrom loads the boilerplate from the provided path
+func WithBoilerplateFrom(path string) UniverseOption {
+	return func(universe *Universe) error {
+		boilerplate, err := ioutil.ReadFile(path)
+		if err != nil {
+			return err
+		}
+
+		universe.Boilerplate = string(boilerplate)
+		return nil
+	}
+}
+
+// WithBoilerplate stores the already loaded project configuration
+func WithBoilerplate(boilerplate string) UniverseOption {
+	return func(universe *Universe) error {
+		universe.Boilerplate = string(boilerplate)
+		return nil
+	}
+}
+
+// WithoutBoilerplate is used for files that do not require a boilerplate
+func WithoutBoilerplate(universe *Universe) error {
+	universe.Boilerplate = "-"
+	return nil
+}
+
+// WithResource stores the provided resource
+func WithResource(resource *resource.Resource, project *config.Config) UniverseOption {
+	return func(universe *Universe) error {
+		resourceModel := &Resource{
+			Namespaced: resource.Namespaced,
+			Group:      resource.Group,
+			Version:    resource.Version,
+			Kind:       resource.Kind,
+			Resource:   resource.Resource,
+			Plural:     flect.Pluralize(strings.ToLower(resource.Kind)),
+		}
+
+		resourceModel.GoPackage, resourceModel.GroupDomain = util.GetResourceInfo(
+			resource,
+			project.Repo,
+			project.Domain,
+			project.MultiGroup,
+		)
+
+		universe.Resource = resourceModel
+		return nil
+	}
+}
--- a/pkg/scaffold/api.go
+++ b/pkg/scaffold/api.go
@@ -21,13 +21,10 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/gobuffalo/flect"
-
 	"sigs.k8s.io/kubebuilder/internal/config"
 	"sigs.k8s.io/kubebuilder/pkg/model"
 	"sigs.k8s.io/kubebuilder/pkg/scaffold/input"
 	"sigs.k8s.io/kubebuilder/pkg/scaffold/resource"
-	"sigs.k8s.io/kubebuilder/pkg/scaffold/util"
 	"sigs.k8s.io/kubebuilder/pkg/scaffold/v1/controller"
 	crdv1 "sigs.k8s.io/kubebuilder/pkg/scaffold/v1/crd"
 	scaffoldv2 "sigs.k8s.io/kubebuilder/pkg/scaffold/v2"
@@ -98,27 +95,12 @@ func (api *API) Scaffold() error {
 	}
 }
 
-func (api *API) buildUniverse() *model.Universe {
-	resourceModel := &model.Resource{
-		Namespaced: api.Resource.Namespaced,
-		Group:      api.Resource.Group,
-		Version:    api.Resource.Version,
-		Kind:       api.Resource.Kind,
-		Resource:   api.Resource.Resource,
-		Plural:     flect.Pluralize(strings.ToLower(api.Resource.Kind)),
-	}
-
-	resourceModel.GoPackage, resourceModel.GroupDomain = util.GetResourceInfo(
-		api.Resource,
-		api.config.Repo,
-		api.config.Domain,
-		api.config.MultiGroup,
+func (api *API) buildUniverse(resource *resource.Resource) (*model.Universe, error) {
+	return model.NewUniverse(
+		model.WithConfig(&api.config.Config),
+		// TODO: missing model.WithBoilerplate[From], needs boilerplate or path
+		model.WithResource(resource, &api.config.Config),
 	)
-
-	return &model.Universe{
-		Resource:   resourceModel,
-		MultiGroup: api.config.MultiGroup,
-	}
 }
 
 func (api *API) scaffoldV1() error {
@@ -130,7 +112,14 @@ func (api *API) scaffoldV1() error {
 		fmt.Println(filepath.Join("pkg", "apis", r.Group, r.Version,
 			fmt.Sprintf("%s_types_test.go", strings.ToLower(r.Kind))))
 
-		err := (&Scaffold{}).Execute(api.buildUniverse(), input.Options{},
+		universe, err := api.buildUniverse(r)
+		if err != nil {
+			return fmt.Errorf("error building API scaffold: %v", err)
+		}
+
+		err = (&Scaffold{}).Execute(
+			universe,
+			input.Options{},
 			&crdv1.Register{Resource: r},
 			&crdv1.Types{Resource: r},
 			&crdv1.VersionSuiteTest{Resource: r},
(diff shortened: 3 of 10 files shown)


Adirio:
Fixed [circular import](https://travis-ci.org/kubernetes-sigs/kubebuilder/jobs/637857227#L254-L258)

cmd/vendor_update.go
```diff
@@ -37,12 +37,22 @@ kubebuilder update vendor
 `,
 		Run: func(cmd *cobra.Command, args []string) {
 			dieIfNoProject()
-			err := (&scaffold.Scaffold{}).Execute(
-				&model.Universe{},
+
+			universe, err := model.NewUniverse(
+				model.WithProjectFrom("PROJECT"),
+				model.WithoutBoilerplate,
+				// No resource at vendor update
```
camilamacedo86: I am not sure if I got the comment `// No resource at vendor update`.
Could you please clarify?

cmd/vendor_update.go
```diff
@@ -37,12 +37,22 @@ kubebuilder update vendor
...
+			)
+			if err != nil {
+				log.Fatalf("error creating scaffolding universe: %v", err)
```
camilamacedo86: I am not sure if make sense print for the user it. 
What does it mean in the user perspective?

pkg/scaffold/api.go
```diff
@@ -152,7 +147,14 @@ func (api *API) scaffoldV1() error {
 		fmt.Println(filepath.Join("pkg", "controller", strings.ToLower(r.Kind),
 			fmt.Sprintf("%s_controller_test.go", strings.ToLower(r.Kind))))
 
-		err := (&Scaffold{}).Execute(api.buildUniverse(), input.Options{},
+		universe, err := api.buildUniverse(r)
+		if err != nil {
+			return fmt.Errorf("error creating scaffolding universe: %v", err)
```
camilamacedo86: same ^

cmd/vendor_update.go
Adirio: There are three set of options to pass, for `Project`, `Boilerplate` & `Resource` fields. I'll clarify the comment (or maybe even delete it if it is confussing) with the next push.

cmd/vendor_update.go
Adirio: What would you print then? Just `err`

camilamacedo86:
```suggestion
				log.Fatalf("error creating scaffolding: %v", err)
```
IHMO: the 'scaffolding universe'' will make sense for the end-user.

pkg/model/universe.go
```diff
@@ -0,0 +1,126 @@
...
+	Project *input.ProjectFile
+
+	Boilerplate string `json:"boilerplate,omitempty"`
+
+	Resource *Resource `json:"resource,omitempty"`
+
+	Files []*File `json:"files,omitempty"`
+
+	// Multigroup tracks if the project has more than one group
+	MultiGroup bool `json:"multigroup,omitempty"`
```
camilamacedo86: WDYT about add a comment on top of it attribute in order to make clear their usage?

pkg/model/universe.go
```diff
@@ -0,0 +1,126 @@
...
+			return err
+		}
+
+		universe.Project = project
+		return nil
+	}
+}
+
+// WithProject stores the already loaded project configuration
+func WithProject(project *input.ProjectFile) UniverseOption {
```
camilamacedo86: May : 
```suggestion
func WithProjectFile(project *input.ProjectFile) UniverseOption {
```

pkg/model/universe.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	return universe, nil
+}
+
+// UniverseOption configure Universe
+type UniverseOption func(*Universe) error
+
+// WithProjectFrom loads the project configuration from the provided path
+func WithProjectFrom(path string) UniverseOption {
```
camilamacedo86: May
```suggestion
func WithProjectFromPath(path string) UniverseOption {
```

pkg/model/universe.go
```diff
@@ -0,0 +1,126 @@
...
+// WithProject stores the already loaded project configuration
+func WithProject(project *input.ProjectFile) UniverseOption {
+	return func(universe *Universe) error {
+		universe.Project = project
+		return nil
+	}
+}
+
+// WithBoilerplateFrom loads the boilerplate from the provided path
+func WithBoilerplateFrom(path string) UniverseOption {
```
camilamacedo86: May
```suggestion
func WithBoilerplateFromPath(path string) UniverseOption {
```

pkg/model/universe.go
Adirio: I like this one

pkg/model/universe.go
```diff
@@ -0,0 +1,126 @@
...
+		universe.Boilerplate = string(boilerplate)
+		return nil
+	}
+}
+
+// WithoutBoilerplate is used for files that do not require a boilerplate
+func WithoutBoilerplate(universe *Universe) error {
+	universe.Boilerplate = "-"
+	return nil
+}
```
camilamacedo86: could not be WithBoilerplate or WithoutBoilerplate automatically/by default?
I mean, could not be `universe.Boilerplate = "-"` by default?  Which means that the above func would not be required at all.

pkg/model/universe.go
Adirio: But I don't like this one. Path is already the argument name when reading the function and when using it you have the path itself

pkg/model/universe.go
Adirio: Sam as above

pkg/model/universe.go
```diff
@@ -0,0 +1,126 @@
...
+	Files []*File `json:"files,omitempty"`
+
+	// Multigroup tracks if the project has more than one group
+	MultiGroup bool `json:"multigroup,omitempty"`
+}
+
+// NewUniverse creates a new Universe
+func NewUniverse(options ...UniverseOption) (*Universe, error) {
+	universe := &Universe{}
+
```
camilamacedo86:
```suggestion
universe.Boilerplate = "-"
```
For by default it be just with  `"-"` and then, we be able to simplify by removing the func WithoutBoilerplate

WDYT?

pkg/model/universe.go
camilamacedo86: I think that would be possible keep WithoutBoilerplate as the default behaviour before the options be applied.
Then, it could be removed. WDYT?

pkg/scaffold/api.go
```diff
@@ -74,11 +72,11 @@ func (api *API) Validate() error {
 
 func (api *API) setDefaults() error {
 	if api.project == nil {
-		p, err := LoadProjectFile("PROJECT")
+		p, err := util.LoadProjectFile("PROJECT")
```
camilamacedo86: 👍 I liked a lot. It really needs to be centralized in util. 
WDYT about rename `util` for `utils`

pkg/scaffold/scaffold.go
```diff
@@ -53,7 +53,7 @@ type Scaffold struct {
 	BoilerplateOptional bool
```
camilamacedo86: when is used the BoilerplateOptional?

pkg/scaffold/scaffold.go
```diff
@@ -53,7 +53,7 @@ type Scaffold struct {
...
 
 	// Project is the project
-	Project input.ProjectFile
+	Project *input.ProjectFile
 
 	ProjectOptional bool
```
camilamacedo86: When is used the ProjectOptional?

pkg/scaffold/scaffold.go
```diff
@@ -185,14 +166,19 @@ func (s *Scaffold) Execute(u *model.Universe, options input.Options, files ...in
...
 	}
 
+	// Set default values for the universe
+	if u.Project == nil {
+		u.Project = &s.Project
+	}
+	if u.Boilerplate == "" {
+		u.Boilerplate = s.Boilerplate
+	}
+	u.Files = make([]*model.File, 0, len(files))
```
camilamacedo86: Since you add a comment to described wdyt about extract it a func? E,g

```go
setDefaultValuesUniverse()
...
func setDefaultValuesUniverse() {
	if u.Project == nil {
		u.Project = &s.Project
	}
	if u.Boilerplate == "" {
		u.Boilerplate = s.Boilerplate
	}
	u.Files = make([]*model.File, 0, len(files))
}
```

pkg/scaffold/project.go
```diff
@@ -122,8 +127,18 @@ func (p *V1Project) Scaffold() error {
...
 
 	s = &Scaffold{}
+
+	universe, err = model.NewUniverse(
+		model.WithProject(&p.Project.ProjectFile),
+		model.WithBoilerplateFrom(p.Boilerplate.Path),
+		// No resource at project initialization
+	)
+	if err != nil {
+		return fmt.Errorf("error creating scaffolding universe: %v", err)
```
camilamacedo86:
```suggestion
		return fmt.Errorf("error creating int project: %v", err)
```

pkg/model/universe.go
Adirio: There is a difference between using the functional option `WithoutBoilerplate` and not using it. I'm not sure if it is applied in this PR or will be in a following one (Im on my phone now and can't check it) but if you leave `Boilerplate` as `""`, `Scaffold.Execute` will try to find the Boilerplate in the default path but if you set it to anything else it will not overwrite it. That's why `WithoutBoilerplate` sets it to `"-"` (any value different to the empty string would work).

pkg/scaffold/api.go
camilamacedo86:
```suggestion
			return fmt.Errorf("error creating int project: %v", err)
```

pkg/scaffold/api.go
```diff
@@ -125,7 +113,14 @@ func (api *API) scaffoldV1() error {
 		fmt.Println(filepath.Join("pkg", "apis", r.Group, r.Version,
 			fmt.Sprintf("%s_types_test.go", strings.ToLower(r.Kind))))
 
-		err := (&Scaffold{}).Execute(api.buildUniverse(), input.Options{},
+		universe, err := api.buildUniverse(r)
+		if err != nil {
+			return fmt.Errorf("error creating scaffolding universe: %v", err)
```
camilamacedo86:
```suggestion
			return fmt.Errorf("error creating int project: %v", err)
```

pkg/model/universe.go
Adirio: Lets keep the discussion in the above comment.

pkg/scaffold/api.go
Adirio: I used `util` because it already existed. I have plans to move everything related to the Project config file, including loading, saving, etc to a separate package.

pkg/scaffold/scaffold.go
Adirio: `Scaffold.Execute` or `Scaffold.defaultValues` IIRC.

pkg/scaffold/scaffold.go
```diff
@@ -185,14 +166,19 @@ func (s *Scaffold) Execute(u *model.Universe, options input.Options, files ...in
...
-		u.Boilerplate = s.Boilerplate
-	}
-
 	if err := s.defaultOptions(&options); err != nil {
 		return err
 	}
 
+	// Set default values for the universe
+	if u.Project == nil {
+		u.Project = &s.Project
```
camilamacedo86: ` cannot use &s.Project (type **input.ProjectFile) as type *input.ProjectFile in assignment`

pkg/scaffold/scaffold.go
Adirio: Same places as above. The usage is to skip errors when trying to load the file if it doesnt exist. IIRC it is used for some tests as the Project and Boilerplate files do not really exist.

pkg/scaffold/project.go
Adirio: Will probably go for `"error initializing project: %v"`

pkg/scaffold/scaffold.go
Adirio: Yes, the `&` needs to be removed. Saw that in Travis too

pkg/model/universe.go
camilamacedo86: Yep, make sense. I think is better just from 👍

pkg/scaffold/scaffold.go
```diff
@@ -31,7 +31,7 @@ import (
 
 	"sigs.k8s.io/kubebuilder/pkg/model"
 	"sigs.k8s.io/kubebuilder/pkg/scaffold/input"
-	"sigs.k8s.io/kubebuilder/pkg/scaffold/project"
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/util"
```
camilamacedo86: IHMO: te should be utils in the plural.

cmd/edit.go
```diff
@@ -46,7 +46,7 @@ func newEditProjectCmd() *cobra.Command {
 		Run: func(cmd *cobra.Command, args []string) {
 			dieIfNoProject()
 
-			projectInfo, err := scaffold.LoadProjectFile("PROJECT")
+			projectInfo, err := scaffoldutil.LoadProjectFile("PROJECT")
```
camilamacedo86: IHMO: we could keep the alias as utils only.
```suggestion
			projectInfo, err := utils.LoadProjectFile("PROJECT")
```
Has any other utils in the project?

pkg/scaffold/scaffold.go
Adirio: Lets keep the discussion in https://github.com/kubernetes-sigs/kubebuilder/pull/1313#discussion_r367866962

cmd/edit.go
Adirio: ### About the alias
There are at least two `util` packages:
1. `cmd/util`
2. `pkg/scaffold/util`

That's the reason why I used `scaffoldutil` as the alias in the files in the `cmd` package: `cmd/edit.go`, `cmd/main.go`, `cmd/webhook_v1.go` and `cmd/webhook_v2.go`.

### About the package name
I used `util` because it already existed. It contains the following functions:
1. `GetResourceInfo`: the `v2` version of this function (there is a different one for `v1` as paths are different. #1255 moves this function to the `resource` package.
2. `LoadProjectFile`: introduced in this PR as a temporary place. I will make a PR moving this function and all the project configuration related things to `internal/config`.

When both functions are moved, the `pkg/scaffold/util` package will be empty, so the name will not matter.

Adirio: Update: #1325 moves the remaining function out of `pkg/scaffold/util`, so this should not be a stopper for this PR.

pkg/scaffold/api.go
Adirio: Just for reference, this has been done in #1325

Adirio:
All the suggestions were tackled. Could we move forward?

DirectXMan12:
@Adirio sure, I'm on board.  Might want to rebase on #1325 since it should be merging soonish.  Ping me on slack if things don't get merged

Adirio:
Rebased, there were a lot of overlaping sections so I may have missed something, so I'm force-pushing now while I do a final check to make Travis also check it meanwhile.

pkg/scaffold/api.go
```diff
@@ -194,6 +190,15 @@ func (api *API) scaffoldV2() error {
 		}
 		fmt.Println(path)
 
+		scaffold := &Scaffold{
+			Plugins: api.Plugins,
+		}
+
+		universe, err := api.buildUniverse(r)
+		if err != nil {
+			return fmt.Errorf("error scaffolding APIs: %v", err)
```
estroz: A different error message differentiates this return from that under `scaffold.Execute`:
```suggestion
			return fmt.Errorf("error building API scaffold: %v", err)
```

pkg/scaffold/api.go
```diff
@@ -157,7 +146,14 @@ func (api *API) scaffoldV1() error {
 		fmt.Println(filepath.Join("pkg", "controller", strings.ToLower(r.Kind),
 			fmt.Sprintf("%s_controller_test.go", strings.ToLower(r.Kind))))
 
-		err := (&Scaffold{}).Execute(api.buildUniverse(), input.Options{},
+		universe, err := api.buildUniverse(r)
+		if err != nil {
+			return fmt.Errorf("error scaffolding controller: %v", err)
```
estroz:
```suggestion
			return fmt.Errorf("error building controller scaffold: %v", err)
```

pkg/scaffold/api.go
```diff
@@ -208,16 +213,18 @@ func (api *API) scaffoldV2() error {
...
-			Plugins: api.Plugins,
+		if err = scaffold.Execute(universe, input.Options{}, files...); err != nil {
+			return fmt.Errorf("error scaffolding APIs: %v", err)
 		}
 
-		if err := scaffold.Execute(api.buildUniverse(), input.Options{}, files...); err != nil {
-			return fmt.Errorf("error scaffolding APIs: %v", err)
+		universe, err = api.buildUniverse(r)
+		if err != nil {
+			return fmt.Errorf("error scaffolding kustomization: %v", err)
```
estroz:
```suggestion
			return fmt.Errorf("error building kustomization scaffold: %v", err)
```

pkg/scaffold/api.go
```diff
@@ -249,13 +256,17 @@ func (api *API) scaffoldV2() error {
 			Plugins: api.Plugins,
 		}
 
-		ctrlScaffolder := &controllerv2.Controller{Resource: r}
+		universe, err := api.buildUniverse(r)
+		if err != nil {
+			return fmt.Errorf("error scaffolding controller: %v", err)
```
estroz:
```suggestion
			return fmt.Errorf("error building controller scaffold: %v", err)
```

pkg/scaffold/api.go
```diff
@@ -130,7 +112,14 @@ func (api *API) scaffoldV1() error {
 		fmt.Println(filepath.Join("pkg", "apis", r.Group, r.Version,
 			fmt.Sprintf("%s_types_test.go", strings.ToLower(r.Kind))))
 
-		err := (&Scaffold{}).Execute(api.buildUniverse(), input.Options{},
+		universe, err := api.buildUniverse(r)
+		if err != nil {
+			return fmt.Errorf("error scaffolding APIs: %v", err)
```
estroz:
```suggestion
			return fmt.Errorf("error building API scaffold: %v", err)
```

Adirio:
@DirectXMan12 The rebase should be ready

PR #1314, opened 2020-01-16
Offer a flag to change the image name

Adirio:
### Description

Add `--image` flag to `kubebuilder init ...` command that allows to specify a image name instead of having it hardcoded.

### Motivation

/kind feature

### DIFF
--- a/cmd/init_project.go
+++ b/cmd/init_project.go
@@ -26,6 +26,7 @@ import (
 	"strconv"
 	"strings"
 
+	docker "github.com/docker/distribution/reference"
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 
@@ -70,6 +71,7 @@ kubebuilder init --domain example.org --license apache2 --owner "The Kubernetes 
 type projectOptions struct {
 	boilerplate project.Boilerplate
 	project     project.Project
+	imageName   string
 
 	// final result
 	scaffolder scaffold.ProjectScaffolder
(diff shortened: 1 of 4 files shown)


cmd/init_project.go
```diff
@@ -114,6 +116,8 @@ func (o *projectOptions) bindCmdlineFlags(cmd *cobra.Command) {
 		"defaults to the go package of the current working directory.")
 	cmd.Flags().StringVar(&o.project.Domain, "domain", "my.domain", "domain for groups")
 	cmd.Flags().StringVar(&o.project.Version, "project-version", project.Version2, "project version")
+
```
camilamacedo86:
```suggestion
```

Adirio: I left the blank line to separate from the projects args block. Is there any special reason to suggest this?

Adirio:
> I am not sure if it would be really useful.
> What is the use case that justifies this addition?
> I mean, what is the scenario where it would be required?

Basically I saw how it was hardcoded inside the functions and I was debating if I should leave it as a const at the start of the file or if I should give it a flag and opted for the second.

> See that the user is able to define the image name when will do the building.
> `make docker-build docker-push IMG=<some-registry>/<project-name>:tag`

Yeah, the only advantage of doing it at project initialization is that you only need to do it once, and at building you may need to do it multiple times. I agree that the added benefit is small though.

mengqiy:
This can improve the UX, but not a lot.
We can get this in unless @DirectXMan12 @droot you have concerns.

Adirio:
I do agree that it is a very small UX enhancement. It was just a few mins worth of work to so I guessed I could do a PR.

Adirio:
Closing in favor of #1330 that uses a const to define the image as providing a flag doesn't have a significant impact in the UX.

PR #1316, opened 2020-01-16
Fix unofficial way of handling already existing files

Adirio:
### Description

Two files were being scaffolded with the `IfExistsAction` set to `Error` which will make the scaffolder return an error, and these kinds of errors were being skipped afterwards. `IfExistsAction` set to `Skip` (default value) is the official way of achieving this.

After doing so, the `isAlreadyExistsError` was unused, and therefore there was no reason to keep `alreadyExistsError` as a struct, `fmt.Errorf` was used instead to simplify the code.

### Motivation

/kind bug

### DIFF
--- a/pkg/scaffold/api.go
+++ b/pkg/scaffold/api.go
@@ -221,7 +221,7 @@ func (api *API) scaffoldV2() error {
 			crdKustomization,
 			&crdv2.KustomizeConfig{},
 		)
-		if err != nil && !isAlreadyExistsError(err) {
+		if err != nil {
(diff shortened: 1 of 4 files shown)


camilamacedo86:
HI @Adirio, 

It shows 100% fine for me. 
/lgtm
/ok-to-test

@mengqiy wdyt?

mengqiy:
Thanks for cleanup this.
/lgtm
/approve

camilamacedo86:
/lgtm
Great work @Adirio

Issue #1319, opened 2020-01-17
Generated makefile controller-gen install errors on 1st run if you don't have either GOBIN or GOPATH set


opetrushka-zendesk:
Generated makefile controller-gen install errors on 1st run if you don't have either GOBIN or GOPATH set
~

Generated the makefile and tried to run in 1st time without having controller-gen installed
```
 >> make generate
go: creating new go.mod: module tmp
go: finding sigs.k8s.io/controller-tools/cmd v0.2.4
go: finding sigs.k8s.io v0.2.4
go: finding sigs.k8s.io/controller-tools/cmd/controller-gen v0.2.4
/Users/opetrushka/go/1.13.5/bin/controller-gen object:headerFile=./hack/boilerplate.go.txt paths="./..."
/bin/sh: /Users/opetrushka/go/1.13.5/bin/controller-gen: No such file or directory
make: *** [generate] Error 127
```
I think this line should be amended
`CONTROLLER_GEN=$(GOBIN)/controller-gen`
https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/scaffold/v2/makefile.go#L123

Identification for both the need to install and using and installed binary uses `$(shell which controller-gen)` which should always be true
GOBIN and GOPATH are optional settings
https://golang.org/doc/install

My suggestion is to use the same line as used elsewhere `CONTROLLER_GEN=$(shell which controller-gen)`
which would give a consistent result and would match the rest of the code.

/kind bug

mengqiy:
> $(GOBIN) might not always be the right place, for example with using goenv, running the makefile points to the global one, while go points to a versioned one.

I'm not following. Can you please explain a bit?

opetrushka-zendesk:
Sorry, I've made a mistake, this isn't really a bug.
It does has an assumption that is common but not 100% must be true, which is either GOBIN or GOPATH are set - which doesn't have to be true.
And even with these two unset, the `which` path can still work.
So I was trying to get it to just work out of the box without these assumptions.
Call it a 0.1% improvement.

As the main go install doc just tells you to add the go/bin to your main PATH, but not to set any of the others
https://golang.org/doc/install

I'm happy to close it as won't do if you don't like this approach.

PR #1321, opened 2020-01-17
Improve controller-gen installer task

hatkyinc2:
This fixes if for the first run for goenv
https://github.com/kubernetes-sigs/kubebuilder/issues/1319

Removing the variation of controler-gen discovery `CONTROLLER_GEN=$(GOBIN)/controller-gen` in favour of the common `CONTROLLER_GEN=$(shell which controller-gen)`
It also works post install, works first and follow up runs..


hatkyinc2:
https://github.com/kubernetes-sigs/kubebuilder/pull/1322

PR #1322, opened 2020-01-17
Improve controller-gen installer task

hatkyinc2:
This fixes if for the first run for minimally configured go setup (not setting the optional gobin and gopath)
kubernetes-sigs#1319

### DIFF
--- a/pkg/scaffold/v2/makefile.go
+++ b/pkg/scaffold/v2/makefile.go
@@ -121,8 +121,6 @@ ifeq (, $(shell which controller-gen))
 	go get sigs.k8s.io/controller-tools/cmd/controller-gen@{{.ControllerToolsVersion}} ;\
 	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
 	}
-CONTROLLER_GEN=$(GOBIN)/controller-gen
-else
-CONTROLLER_GEN=$(shell which controller-gen)
 endif
+CONTROLLER_GEN=$(shell which controller-gen)
 `
--- a/pkg/scaffold/v2/makefile.go
+++ b/pkg/scaffold/v2/makefile.go
@@ -51,13 +51,6 @@ IMG ?= {{ .Image }}
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
-# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
-ifeq (,$(shell go env GOBIN))
-GOBIN=$(shell go env GOPATH)/bin
-else
-GOBIN=$(shell go env GOBIN)
-endif
-
 all: manager
 
 # Run tests
@@ -112,7 +105,9 @@ docker-push:
 # find or download controller-gen
 # download controller-gen if necessary
 controller-gen:
-ifeq (, $(shell which controller-gen))
+  # Get CONTROLLER_GEN from one of the possibly locations, ordered by priority
+  @CONTROLLER_GEN=$(shell PATH=$$PATH:$$GOPATH/bin:$$GOBIN:$$GOPATH/bin:$$HOME/go/bin which controller-gen)
+ifeq (, $(CONTROLLER_GEN))
 	@{ \
 	set -e ;\
 	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
@@ -121,6 +116,6 @@ ifeq (, $(shell which controller-gen))
 	go get sigs.k8s.io/controller-tools/cmd/controller-gen@{{.ControllerToolsVersion}} ;\
 	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
 	}
+  @CONTROLLER_GEN=$(shell PATH=$$PATH:$$GOPATH/bin:$$GOBIN:$$GOPATH/bin:$$HOME/go/bin which controller-gen)
 endif
-CONTROLLER_GEN=$(shell which controller-gen)
 `
--- a/testdata/project-v2-multigroup/Makefile
+++ b/testdata/project-v2-multigroup/Makefile
@@ -74,7 +74,5 @@ ifeq (, $(shell which controller-gen))
 	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.4 ;\
 	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
 	}
-CONTROLLER_GEN=$(GOBIN)/controller-gen
-else
-CONTROLLER_GEN=$(shell which controller-gen)
 endif
+CONTROLLER_GEN=$(shell which controller-gen)
--- a/testdata/project-v2-multigroup/Makefile
+++ b/testdata/project-v2-multigroup/Makefile
@@ -4,13 +4,6 @@ IMG ?= controller:latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
-# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
-ifeq (,$(shell go env GOBIN))
-GOBIN=$(shell go env GOPATH)/bin
-else
-GOBIN=$(shell go env GOBIN)
-endif
-
 all: manager
 
 # Run tests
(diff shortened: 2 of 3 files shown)


hatkyinc2:
Signed CLA agreement

hatkyinc2:
>k8s-ci-robot added size/S and removed size/XS labels 3 minutes ago

😞

mengqiy:
@Adirio Can you help review this?

Adirio:
> @Adirio Can you help review this?

I gave it a fast review: I do think it needs some enhancement but this is just removing one of the options. If `go get ...` doesn't place the controller-gen executable somewhere in the path, it won't work. I think there was other PR aimed towards this which was using go run to delegate dowload and execution and I would opt for that option.

estroz:
@Adirio #1289

Adirio:
> @Adirio #1289

I won you this race (looked for it an edited the answer seconds before you posted it) :)

estroz:
It would be best to close this in favor of #1289 since we can knock out two code-birds (versioning and install path) with one PR.

hatkyinc2:
I'm not against https://github.com/kubernetes-sigs/kubebuilder/pull/1289, but it seems @DirectXMan12 is blocking the approach taken for a reason. (I don't fully understand how long that's relevant for, short/long plan of something else?..)

There doesn't seem to be a contradiction between both of them.
This could be seen as step 1, while the other is step 2.
And if step 2 would take a long while, we can still take step 1...

This one shouldn't have the same blocker as it doesn't use the `go run` which @DirectXMan12 has blocked the other on.

I understand there is more context I probably don't know like when is the other expected to be unblocked, how frequent are "releases" if relevant... 

Is closing this in favour of the 2nd step going to result in a faster turn around after all the details are accounted for?..

hatkyinc2:
Does anyone know what this testing error is?
```
object:headerFile="hack/boilerplate.go.txt" paths="./..."
/bin/sh: 1: object:headerFile=hack/boilerplate.go.txt: not found
make: *** [Makefile:48: generate] Error 127
```

(Can `make test` fine locally and such)

Adirio:
The logic when you do `make controller-gen` (or basically any other than `make fmt`, `make vet` or `make docker-push`)  was using `shell which controller-gen` to guess if `controller-gen` is installed:
- If `shell which controller-gen` finds `controller-gen` then use its output.
- But if it doesn't, use `go get ...` to install it and define the path were it is installed (`$GOPATH/bin/controller-gen` or `$GOBIN/controller-gen`, will use them indistinctly).

Which are the issues with this logic:
1. `shell which controller-gen` will only find `controller-gen` if it is in our `$PATH`. It may already be installed in `$GOPATH/bin` but this folder may not be part of `$PATH`. We should probably check this path before installing.
2. If neither `$GOPATH` nor `$GOBIN` global variables are set, and `controller-gen` gets installed this way, `$(GOBIN)` will be an empty string so `CONTROLLER_GEN=/controller-gen` and that will fail as it is an absolute path.
3. If we do find controller-gen, it may be a completely different version. This is the main issue, but it is out of the scope of this PR so lets leave it aside for now.

So we have 4 possible scenarios depending on `$GOPATH` or `$GOBIN` env variables or `controller-gen` being in `$PATH`:
1. `$GOPATH` or `$GOBIN` set & `controller-gen` in `$PATH`: no issues other than the third one.
2. `$GOPATH` or `$GOBIN` **NOT** set & `controller-gen` in `$PATH`: fail first time you run it, but work the rest of the times.
3. `$GOPATH` or `$GOBIN` set & `controller-gen` **NOT** in `$PATH`: all the times you run `make ...`, it will try to `go get controller-gen@...` when it may not be needed. `go get ...` should be intelligent enough to be a no-op.
4. `$GOPATH` or `$GOBIN` **NOT** set & `controller-gen` **NOT** in `$PATH`: the universe explodes and nothing works.

This PR targets the second scenario, making it work ~as smoothly~ as the first, but breaks the third scenario which was working (maybe a bit less intelligent that it should, but it was working). So, IMHO, solving an issue with a one-time error for a scenario doesn't justify breaking another scenario. That is the beuty of the other PR, using `go run ...` delegates both finding the executable and downloading it to go and solves the versioning issue.

hatkyinc2:
@Adirio Do you value the fact that the golang installation guild as part of the installation say you need to setup the PATH to include the bin path?
https://golang.org/doc/install
(Also goenv products set that path)

(GOPATH has a side mention as optional under testing)
(GOBIN has no mention)

Adirio:
> @Adirio Do you value the fact that the golang installation guild as part of the installation say you need to setup the PATH to include the go bin path? (for me at the point, it should be taken as a given it's setup... while it's not a perfect absolute, it should be very close to that..)
> https://golang.org/doc/install
> (Also goenv products set that path)
> 
> (GOPATH has a side mention as optional under testing.
> GOBIN has no mention.)

I do think that the second scenario is more common, sure, but you should not break the third one. You could for example do the following steps.

1. If `shell which controller-gen` finds it, use that.
2. If `$GOPATH` and `$GOPATH/bin/controller-gen` exists, use that.
3. `go get controller-gen@...`
4. If `shell which controller-gen` finds it, use that.
5. If `$GOPATH` and `$GOPATH/bin/controller-gen` exists, use that.
6. Log an error that says that `controller-gen` wasn't found even after an installation attempt.

That would solve the second scenario but would still work for the third scenario.

P.S.: as steps 1 and 2 are the same as 4 and 5, they should probably be extracted into a function that returns the value that needs to be set to `CONTROLLER_GEN`. So basically call this function, check if it blank, install, call the function again, and if it is still blank print an error.

hatkyinc2:
Would something like be an acceptable proposal?
```
CONTROLLER_GEN=$(export PATH=$$PATH:$$GOBIN;$$GOPATH/bin; which controller-gen)
if empty:
 - install 
 - CONTROLLER_GEN=$(export PATH=$$PATH:$$GOBIN;$$GOPATH/bin; which controller-gen)
```
Possible to add another if empty - output error 

(if at it we can also add the GOPATH/bin ..)

Adirio:
> Would something like be an acceptable proposal?
> 
> ```
> CONTROLLER_GEN=$(export PATH=$$PATH:$$GOBIN;$$GOPATH/bin; which controller-gen)
> if empty:
>  - install 
>  - CONTROLLER_GEN=$(export PATH=$$PATH:$$GOBIN;$$GOPATH/bin; which controller-gen)
> ```
> 
> Possible to add another if empty - output error

I would not modify `$PATH`

hatkyinc2:
>I would not modify $PATH

In a makefile it doesn't touch you'r actual env or drags into any other command, it's just valid for that one liner,
Is that still a concern?

Adirio:
Not really a concern, go for it

pkg/scaffold/v2/makefile.go
```diff
@@ -51,12 +51,8 @@ IMG ?= {{ .Image }}
...
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
-# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
-ifeq (,$(shell go env GOBIN))
-GOBIN=$(shell go env GOPATH)/bin
-else
-GOBIN=$(shell go env GOBIN)
-endif
+# Get CONTROLLER_GEN from one of the possibly locations, ordered by priority
+CONTROLLER_GEN := $(shell PATH=$$PATH:$$GOPATH/bin:$$GOBIN:$$GOPATH/bin:$$HOME/go/bin which controller-gen)
```
Adirio: Why is `$$PATH/bin` twice

Also applies to line 120

pkg/scaffold/v2/makefile.go
```diff
@@ -121,8 +117,6 @@ ifeq (, $(shell which controller-gen))
 	go get sigs.k8s.io/controller-tools/cmd/controller-gen@{{.ControllerToolsVersion}} ;\
 	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
 	}
-CONTROLLER_GEN=$(GOBIN)/controller-gen
-else
-CONTROLLER_GEN=$(shell which controller-gen)
+CONTROLLER_GEN := $(shell PATH=$$PATH:$$GOPATH/bin:$$GOBIN:$$GOPATH/bin:$$HOME/go/bin which controller-gen)
 endif
```
Adirio: Could you add the error message we were talking about?

DirectXMan12:
/assign @droot 

who has volunteered to handle the controller-gen dependency kerfuffle

hatkyinc2:
The more I worked on it the more I understood how we got to the current state..
I'm supporting https://github.com/kubernetes-sigs/kubebuilder/pull/1289 and do not wish to put any more effort into this.

PR #1324, opened 2020-01-17
Order travis stages and clean non-needed tests

Adirio:
Order Travis script execution in order to avoid using resources if easier checks fail. If any stage fails, the rest are not executed, and thus easy to run stages should be placed first.

The order chosen has been:
1. Linting: fastest to run (`verify.sh`).
2. Golden test: only executed once (`golden_test.sh`).
3. Local test: only executed once (`test.sh`).
4. Test: executed for several OS, Go and K8s configurations (`test_e2e.sh`).
5. Coverage: trivial (`make test-coverage`).

Additional changes:
- `test.sh` was being run with `test_e2e.sh` so it was being executed 4 times when only 2 are needed (`linux` and `osx`).
- `golden_test.sh` was being run with `verify.sh` so if `verify.sh` failed, you needed to wait 5-10 minutes to finish that test.
- `test.sh` was not testing the multi-group enabled test project.

NOTE: the total number of processes has been increased by two, but it is because several tests were grouped in the same process. If we count the number of tests, we have actually reduced it by 2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,72 +5,72 @@ cache:
   - $HOME/.cache/go-build
   - $GOPATH/pkg/mod
 
-os:
-  - linux
-  - osx
+os: linux
 
 go:
-- "1.13"
+  - "1.13"
 
-# A build matix defines the K8s versions to use for e2e tests. Travis runs these in parrellel
+# A build matrix defines the K8s versions to use for e2e tests. Travis runs these in parallel
 env:
   - KIND_K8S_VERSION="v1.16.2"
   - KIND_K8S_VERSION="v1.15.3"
   - KIND_K8S_VERSION="v1.14.1"
-  - GOPROXY=https://proxy.golang.org/
-# As OSX isn't running KIND e2e tests exlude it from the matrix
-matrix:
-  exclude:
-  - os: osx
-    env: KIND_K8S_VERSION="v1.16.2"
-  - os: osx
-    env: KIND_K8S_VERSION="v1.15.3"
-  - os: osx
-    env: KIND_K8S_VERSION="v1.14.1"
-  - os: linux
-    env: GOPROXY=https://proxy.golang.org/
 
 git:
   depth: 3
 
 go_import_path: sigs.k8s.io/kubebuilder
 
-services:
-- docker
-
-before_install:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -55,23 +56,31 @@ script:
 
 jobs:
   include:
-    - stage: Validations
+    - stage: linting
       before_script:
         - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.21.0
       script:
         - ./scripts/verify.sh
+    - stage: golden
```
camilamacedo86:
```suggestion
    - stage: generate
```

IHMO:  Golden has no meaning and we have the target that `make generate` which gen the testdata and do all required tricks as, for example, ensure the correct module versions which should be used. (Described  in the CONTRIBUTION.md)

.travis.yml
```diff
@@ -18,6 +18,7 @@ env:
   - KIND_K8S_VERSION="v1.15.3"
   - KIND_K8S_VERSION="v1.14.1"
   - GOPROXY=https://proxy.golang.org/
```
mengqiy: I guess this doesn't matter any more. We can probably drop this line.
One benefit of dropping it is that it reduces the number of concurrent jobs for one PR. IIRC we can only run limited number of concurrent jobs for one project.

mengqiy: You can either do it in this PR or in another PR. Up to you.

Adirio: @mengqiy Do you mean `GOPROXY=https://proxy.golang.org/`? That would remove every `osx` test, wouldn't it? The `osx` line under `os` and the `exceptions` for `osx` could also be removed.

camilamacedo86: Would be nice add a comment in the file to make clear that the os run the test.sh and the Linux run the e2e in the vars wdyt?

Adirio: > You can either do it in this PR or in another PR. Up to you.

I think we should do it in another PR, discussing it here is ok

Adirio: @mengqiy Was this what you were thinking about? Each stage is gated behind the previous one, so we are saving resources as those PR that do not pass the linting stage will not take any more resource.

mengqiy:
Looks great! Thanks for working on it!
/lgtm
/approve

Adirio:
Travis has passed, I don't know why GitHub hasn't updated it. Should I close and reopen?

PR #1330, opened 2020-01-23
:running: Homogenize commands through Scaffolder interface

Adirio:
# Description

This PR defines a Scaffolder interface with a single method:

```go
type Scaffolder interface {
	Scaffold()
}
```

This interface is used as the API between the `cmd` and the `pkg/scaffold` packages. This allows to provide a clearer limit between the two packages, as previously the limit was not clear and logic that should be placed in `pkg/scaffold` could be found in `cmd` (e.g., webhook scaffolding was being done directly in `cmd`).

Additionally the option structs for each command has also been homogenized. An unexported interface has been used for this purpose:

```go
type commandOptions interface {
	// bindFlags binds the command flags to the fields in the options struct
	bindFlags(command *cobra.Command)

	// The following steps define a generic logic to follow when developing new commands. Some steps may be no-ops.
	// - Step 1: load the config failing if expected but not found or if not expected but found
	loadConfig() (*config.Config, error)
	// - Step 2: verify that the command can be run (e.g., go version, project version, arguments, ...)
	validate(*config.Config) error
	// - Step 3: create the Scaffolder instance
	scaffolder(*config.Config) (scaffold.Scaffolder, error)
	// - Step 4: call the Scaffold method of the Scaffolder instance
	// Doesn't need any method
	// - Step 5: finish the command execution
	postScaffold(*config.Config) error
}
}
```

The logic flow described for the `run` method has been used for every command. In some cases, some steps are simply no-ops, like the 5th step for the edit command.

# Motivation

This PR is one of the main contributors to #1218.


Adirio:
> Is there any way that we can put the steps into an interface / helper so we don't have to repeat them in every run method?

Something like the following?

```go

type commandOptions interface {
	bindFlags(cmd *cobra.Command)
	loadConfig() error
	validate(*config.Config) error
	scaffolder(*config.Config) error
	postScaffold(*config.Config) error
}

func run(options *commandOptions) err {
	// Step 1: load config
	config, err := options.loadConfig()
	if err != nil {
		return err
	}

	// Step 2: validate
	if err := options.validate(config); err != nil {
		return err
	}

	// Step 3: create scaffolder
	scaffolder, err := options.scaffolder(config)
	if err != nil {
		return err
	}

	// Step 4: scaffold
	if err := scaffolder.Scaffold(); err != nil {
		return err
	}

	// Step 5: finish
	if err := options.postScaffold(); err != nil {
		return err
	}
}
```

Adirio:
> Is there any way that we can put the steps into an interface / helper so we don't have to repeat them in every run method?

@DirectXMan12 Something like the following?

```go

type commandOptions interface {
	bindFlags(cmd *cobra.Command)
	loadConfig() error
	validate(*config.Config) error
	scaffolder(*config.Config) error
	postScaffold(*config.Config) error
}

func run(options *commandOptions) err {
	// Step 1: load config
	config, err := options.loadConfig()
	if err != nil {
		return err
	}

	// Step 2: validate
	if err := options.validate(config); err != nil {
		return err
	}

	// Step 3: create scaffolder
	scaffolder, err := options.scaffolder(config)
	if err != nil {
		return err
	}

	// Step 4: scaffold
	if err := scaffolder.Scaffold(); err != nil {
		return err
	}

	// Step 5: finish
	if err := options.postScaffold(); err != nil {
		return err
	}
}
```

Adirio:
Didn't expect the golden test to check file permission, but [it does](https://travis-ci.org/kubernetes-sigs/kubebuilder/jobs/642921640#L828-L842).

Adirio:
@DirectXMan12 I made a separate commit with the helper as I described it [here](https://github.com/kubernetes-sigs/kubebuilder/pull/1330#issuecomment-578866647).

DirectXMan12:
looks good once it's rebased

cmd/api.go
```diff
@@ -18,157 +18,205 @@ package main
...
 	}
 
-	if err := o.postScaffold(); err != nil {
-		log.Fatal(err)
-	}
-}
+	// In case we want to scaffold a resource API we need to do some checks
+	if o.doResource {
+		// Skip the following check for v1 as resources aren't tracked
+		if !c.IsV1() {
```
mengqiy: nit: maybe this is better in terms of readability.
```
		if !c.IsV1() {
                	return nil
                }
		if !o.force {
			resourceExists := false
			for _, r := range c.Resources {
				if r.Group == o.resource.Group &&
					r.Version == o.resource.Version &&
					r.Kind == o.resource.Kind {
					resourceExists = true
					break
				}
			}
			if resourceExists {
			}
		}
```

mengqiy:
This is a great refactor! Thanks for working on this.

cmd/api.go
Adirio: I wanted to make the difference from things that were `v2`-specific, like `multigroup` from things that are just not in `v1`. In case of a potential `v3`, `multigroup` may be supported straight away, but I'm quite sure that resources will net stop being tracked. Does that make sense?

mengqiy: Fair enough

PR #1332, opened 2020-01-26
Fix wrong paths in multi-group migratig doc

### LINKED ISSUE #1331: Wrong description of multi-group in book
As described in https://book.kubebuilder.io/migration/multi-group.html:
image
However, controllers are not created under pkg/controllers/\<group\>. They're created under controllers/\<group\>
/kind documentation

Adirio:
Fixes #1331


Adirio:
@camilamacedo86 @DirectXMan12 @mengqiy This should be a straight up PR. It just fixes the migrationg guide paths of the `controllers` package that is in the root directory, not under `pkg`.

DirectXMan12:
/ok-to-test

@Adirio you should file an issue in kubernetes/org to get kubernetes and/or k-sigs membership -- then we won't have to manually `ok-to-test` on your PRs ;-).  You'll need two existing sponsors if you're not already a kubernetes org member.  I'm happy to be one, and I'm sure one of the other reviewers (from another company) across the KB project would be happy to sponsor you too :-)

PR #1338, opened 2020-01-29
Clean old users from the OWNERS file

DirectXMan12:
This cleans up old users from the OWNERS and OWNERS_ALIASES files.
This should mean that folks who get assigned for reviews are now active.


Adirio:
Travis needs to be re-trigered

PR #1349, opened 2020-02-04
Implement Wrapper interface for errors

Adirio:
# Description

Most of the errors in `kubebuilder` are wrapping other errors. By implementing the `Wrapper` interface defined [here](https://go.googlesource.com/proposal/+/master/design/29934-error-values.md) we can make `gomega.MatchError` be able to check wrapped errors.

This behavior has been implemented for several wrapper errors. Gomega needs to be upgraded to v1.8.0+ to benefit from this behavior. Latest version as of today is v1.9.0.

# Motivation

This PR is scoped under #1218


Adirio:
@DirectXMan12 this is the PR I mentioned in https://github.com/kubernetes-sigs/kubebuilder/pull/1375#discussion_r382282884

DirectXMan12:
This is the same interface for Go 1.13 errors, so it'll get us `Is` and `As` support in the main codebase as well.

Seems conceptually reasonable to me.

/ok-to-test

pkg/scaffold/internal/filesystem/errors.go
```diff
@@ -78,13 +96,19 @@ type createFileError struct {
...
+// Unwrap implements Wrapper interface
+func (e createFileError) Unwrap() error {
+	return e.err
+}
+
 // IsCreateFileError checks if the returned error is because the file could not be created
-func IsCreateFileError(e error) bool {
-	_, ok := e.(createFileError)
+func IsCreateFileError(err error) bool {
+	_, ok := err.(createFileError)
```
DirectXMan12: prob should move to `return errors.As(err, createFileError{})`

Adirio: I feel like that is less descriptive. IMO it seems to try to use a new feature when it doesn't really fit, I may just not be used to it.

pkg/scaffold/internal/machinery/errors.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"fmt"
+
+	"sigs.k8s.io/kubebuilder/pkg/model/file"
+)
+
+// This file contains the errors returned by the scaffolding machinery
+// They are not exported as they should not be created outside of this package
+// Exported functions are provided to check which kind of error was returned
+
```
djzager: Should these errors implement the `Unwrap` method?

pkg/scaffold/internal/filesystem/errors.go
Adirio: The thing is that they do not do exactly the same. Right now it only detects if the error returned is of that type, and using `errors.As(err, ERROR_TYPE)` will check if any of the chained errors is of that type. Do we want to move towards that?

pkg/scaffold/internal/machinery/errors.go
Adirio: They do not wrap any error.

pkg/scaffold/internal/filesystem/errors.go
estroz: I say we should move towards using `errors.As` since the underlying error may have been wrapped by an anonymous `fmt.Errorf` for a more informative error message by a caller.

Adirio: Done

djzager:
/assign @DirectXMan12 

I believe this is ready.

pkg/scaffold/internal/filesystem/filesystem_test.go
```diff
@@ -153,144 +151,6 @@ var _ = Describe("FileSystem", func() {
 		})
 	})
 
-	Describe("Errors", func() {
```
camilamacedo86, line 156: HI @Adirio, 
 
My only concern here is: why these tests were removed? 
Could we not keep them? Otherwise, it shows ok

go.mod
```diff
@@ -4,19 +4,21 @@ go 1.13
 
 require (
 	github.com/gobuffalo/flect v0.2.0
-	github.com/golang/protobuf v1.3.2 // indirect
-	github.com/kr/pretty v0.1.0 // indirect
-	github.com/onsi/ginkgo v1.10.1
-	github.com/onsi/gomega v1.7.0
+	github.com/golang/protobuf v1.3.4 // indirect
```
camilamacedo86, line 7: Could we not remove the //indirect imports?

pkg/scaffold/internal/filesystem/filesystem_test.go
Adirio: They were moved to the `errors_test.go` file in the same directory.

go.mod
Adirio: Out of scope of this PR.

camilamacedo86: The above comment is to remove the new // indirect modules added here. 
Also, why was added new direct ones as well?

pkg/scaffold/internal/machinery/errors_test.go
```diff
@@ -0,0 +1,71 @@
...
+var _ = Describe("Errors", func() {
+	var (
+		path                     = filepath.Join("path", "to", "file")
+		err                      = errors.New("test error")
+		fileAlreadyExistsErr     = fileAlreadyExistsError{path}
+		modelAlreadyExistsErr    = modelAlreadyExistsError{path}
+		unknownIfExistsActionErr = unknownIfExistsActionError{path, -1}
+	)
+
+	DescribeTable("IsXxxxError should return true for themselves and false for the rest",
```
camilamacedo86, line 43: IsXxxxError?

go.mod
Adirio: If the other PR gets merged I will.

pkg/scaffold/internal/machinery/errors_test.go
Adirio: There are several functions with this pattern and all check with this table. In the entry comment which error is being check for that specific entry is defined.

go.mod
camilamacedo86: Hi @Adirio you added here new entries. These ones are that one that I a suggestion you remove for we are able to move forward. I mean, can you just revert your changes in these files?

camilamacedo86: After your chat I will approve your PR here and then, working on to clean it in the #1454

Adirio: As talked offline, we will handle this in #1454

PR #1355, opened 2020-02-05
Clean Code - Remove V1 implementation

camilamacedo86:
**Description**
- Remove the V1 code which should no longer more be used
- Remove the V1 tests and testdata
- Remove deprecated flags valid just for V1

**Motivation**
- "The v1 projects are deprecated for too long and not supported since Feb 1, 2020.

> **Note that by removing the V1 we can:**
> 
> - Reduce the complexity of the code and tests
> - Reduce the time to check the result of tests
> - Remove the vendor's/deps and needs for its combability.


Adirio:
I said this in several places but will say it here as its where it makes sense.

**I think we should not remove v1**

The code has received a huge refactor lately and v1-v2 separation is quite clear right now. Other commits such as #1348 or #1290 will help in this sense even more. I would just leave v1 as it is right now, change the deprecation warning to something like: "v1 has been completely deprecated since 1st of February, 2020, and will no longer receive any fix or update. Upgrading your projects to v2 is highly recommended. In the following link you can find a step-by-step guide on how to do it: URL". I would probably also prevent the initialization of new v1 projects.

camilamacedo86:
Hi @Adirio, 

Thank you for your input. Following my comments inline. 

> I said this in several places but will say it here as its where it makes sense.
> The code has received a huge refactor lately and v1-v2 separation is quite clear right now. 

Your work was not envious. It was great and indeed allow us to continue the lifecycle doing more cleanups.  And then, the work made in the V2 will still, however, the V1 was already flagged for no longer be supported and as I said many times we should not change/update and/or care so much because we knew that at some point it would be removed. 

Why should we keep a code, tests, mocks, scripts, vendor and etc if they should no longer be used/updated/changed? 

Note that by removing the V1 we can:

- Reduce the complexity of the code and tests
- Reduce the time to check the result of tests
- Remove the vendor's/deps and needs for its combability.

So, IMO by removing all unnecessary V1 implementation it can simplify a lot. However, I'd like to still keep for now just the updated command to allow users are able to upgrade it to V2 with the latest versions. 

@droot @mengqiy @DirectXMan12  WDYT? 
PS.: PR is in WIP yet. ( I will probably waiting for the plugin merge to ask review on it )

Adirio:
> > I said this in several places but will say it here as its where it makes sense.
> > The code has received a huge refactor lately and v1-v2 separation is quite clear right now.
> 
> Your work was not envious. It was great and indeed allow us to continue the lifecycle doing more cleanups. And then, the work made in the V2 will still as the code that will allow us to create the V3 in the future, however, the V1 was already flagged for no longer be supported and as I said many times we should not change/update and/or care so much because we knew that at some point it would be removed.

I'm not worried about my code dissapearing. 😄 The point I was trying to make is that the situation when it was decided that v1 should be deprecated (3 months ago) and now is quite different, so despite being deprecated we may not need to remove the code completely.

> Why should we keep a code, tests, mocks, scripts, vendor and etc if they should no longer be used/updated/changed?
> 
> Note that by removing the V1 we can:
> 
> * Reduce the complexity of the code and tests
> * Reduce the time to check the result of tests
> * Remove the vendor's/deps and needs for its combability.

I do agree that we should remove the v1 tests from `golden_test.sh`, `test.sh`, and `test_e2e.sh` and thus the `testdata/gopath` dir. But I think that that is enough. I would keep the code for v1. Right now I would not say that v1 makes the code more complex outside of the mentioned tests.

.travis.yml
```diff
@@ -24,8 +24,6 @@ go_import_path: sigs.k8s.io/kubebuilder
 services: docker
 
 # Install must be set to prevent default `go get` to run.
-# The dependencies have already been vendored by `dep` so
-# we don't need to fetch them.
 install: skip
```
Adirio: Do we still need `install: skip`? I think we should ask the one that wrote the comment you deleted why it was used and think if that applies for v2 tests.

CONTRIBUTING.md
```diff
@@ -8,11 +8,10 @@ Please see https://git.k8s.io/community/CLA.md for more info.
 
 ## Prerequisites
 - [go](https://golang.org/dl/) version v1.13+.
-- [dep](https://github.com/golang/dep) dep v0.5+
 - [docker](https://docs.docker.com/install/) version 17.03+.
-- [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) version v1.11.3+.
+- [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) version v1.14.1+.
 - [kustomize](https://sigs.k8s.io/kustomize/docs/INSTALL.md) v3.1.0+
-- Access to a Kubernetes v1.11.3+ cluster.
+- Access to a Kubernetes v1.14.1+ cluster.
```
Adirio: Do we need 1.14.1 for v2? Which feature is required for v2 that was added by 1.14.1?

cmd/init.go
```diff
@@ -17,18 +17,14 @@ limitations under the License.
...
 	"fmt"
 	"log"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"strings"
 
 	"github.com/spf13/cobra"
-	flag "github.com/spf13/pflag"
-
```
Adirio: Keep this blank line to split k8s-related and external libraries in two import groups

cmd/init.go
```diff
@@ -48,7 +44,7 @@ func newInitCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "init",
 		Short: "Initialize a new project",
-		Long: `Initialize a new project including vendor/ directory and Go package directories.
+		Long: `Initialize a new project including Go package directories.
```
Adirio: The rest of the sentence gives no information at all.
```suggestion
		Long: `Initialize a new project.
```

cmd/internal/config.go
```diff
@@ -40,10 +35,20 @@ func ConfiguredAndV1() bool {
...
 
-	return projectConfig.IsV1()
+	return true
 }
 
-func PrintV1DeprecationWarning() {
-	fmt.Printf(noticeColor, "[Deprecation Notice] The v1 projects are deprecated and will not be supported beyond "+
-		"Feb 1, 2020.\nSee how to upgrade your project to v2: https://book.kubebuilder.io/migration/guide.html\n")
+// IsSupportable returns true if the project is already configured and it is v2
+func IsSupportable() bool {
```
Adirio:
```suggestion
func IsVersionSupported() bool {
```

cmd/main.go
```diff
@@ -78,19 +84,22 @@ func run(options commandOptions) error {
...
 	}
 
 	// kubebuilder
 	rootCmd := newRootCmd()
 
+	// kubebuilder update (v1 only)
+	if !internal.IsSupportable() {
+		rootCmd.AddCommand(newUpdateCmd())
+	}
+
```
Adirio: Could you leave this after init and before version? They are ordered alphabetically.

cmd/webhook.go
```diff
@@ -40,110 +39,6 @@ func (e webhookError) Error() string {
...
-		err := internal.RunCmd("Running make", "make")
-		if err != nil {
-			return err
-		}
-	}
-
-	return nil
-}
-
 func newWebhookV2Cmd() *cobra.Command {
```
Adirio: Probably you should remove v2 from names for this function and the rest of the functions of this file.

generate_golden.sh
```diff
@@ -37,69 +37,45 @@ scaffold_test_project() {
...
+        $kb create webhook --group crew --version v1 --kind Captain --defaulting --programmatic-validation
+        $kb create api --group ship --version v1beta1 --kind Frigate --controller=true --resource=true --make=false
+        $kb create webhook --group ship --version v1beta1 --kind Frigate --conversion
+        $kb create api --group ship --version v1 --kind Destroyer --controller=true --resource=true --namespaced=false --make=false
+        $kb create api --group ship --version v2alpha1 --kind Cruiser --controller=true --resource=true --namespaced=false --make=false
+        $kb create api --group sea-creatures --version v1beta1 --kind Kraken --controller=true --resource=true --make=false
+        $kb create api --group sea-creatures --version v1beta2 --kind Leviathan --controller=true --resource=true --make=false
+        $kb create api --group foo.policy --version v1 --kind HealthCheckPolicy --controller=true --resource=true --make=false
     fi
-    make all test # v2 doesn't test by default
```
Adirio: You should not remove this. The comment means that we need to run `make all test` instead of `make` (or `make all`) because in v1 when you do `make` you run the tests but in v2 `make` doesn't run the tests. So it is not saying that v2 doesn't need this line, it says that v2 requires to specify test target.

You can remove the comment though.

test.sh
```diff
@@ -96,17 +93,8 @@ function test_project {
...
-  vendor_tarball=$tmp_root/vendor.v$version.tgz
   old_gopath=$GOPATH
-  if [[ $version == "1" ]]; then
-      export GOPATH=$(pwd)/testdata/gopath
-      download_vendor_archive
-  fi
   cd testdata/$project_dir
-  # v2 uses modules, and thus doesn't have a vendor directory
-  [[ -e ${vendor_tarball} ]] && tar -zxf $vendor_tarball
-  make all test # v2 doesn't test on all by default
```
Adirio: Same as above comment, `make all test` needs to stay, the comment can go.

.travis.yml
Adirio: @droot That comment was added by you a couple ages ago. Should it stay for v2 now that it uses gomodules?

test.sh
```diff
@@ -96,17 +93,8 @@ function test_project {
   project_dir=$1
   version=$2
   header_text "performing tests in dir $project_dir for project version v$version"
-  vendor_tarball=$tmp_root/vendor.v$version.tgz
   old_gopath=$GOPATH
```
Adirio: This should also be removed as we are not using the GOPATH any more

test.sh
```diff
@@ -96,17 +93,8 @@ function test_project {
...
-      export GOPATH=$(pwd)/testdata/gopath
-      download_vendor_archive
-  fi
   cd testdata/$project_dir
-  # v2 uses modules, and thus doesn't have a vendor directory
-  [[ -e ${vendor_tarball} ]] && tar -zxf $vendor_tarball
-  make all test # v2 doesn't test on all by default
-  [[ -e ${vendor_tarball} ]] && rm -rf ./vendor && rm -f Gopkg.lock
   cd -
   export GOPATH=$old_gopath
```
Adirio: Same as above

CONTRIBUTING.md
camilamacedo86: We have no tests to ensure the usage of the project with old versions < 1.14.1+.

Also, I know that from v1.11.+ to the latest has significant changes, however, it is very hard to know what feature could or not been impacted by it.  And then, IMO make sense we keep the lower version which we are using in the CI to verify the supportability of it.

cmd/webhook.go
camilamacedo86: IHMO not now at least. Note that we still checking if is V1 to say that is deprecated and we do not know if we will still follow the same V1, V2 approach to creating the V3 in the future. However, I agree that a second phase we could do that.

generate_golden.sh
camilamacedo86: Good catch tks.

CONTRIBUTING.md
Adirio: But the hard requirement set in code is 1.11, what about saying 1.11.3+ is needed, 1.14.1+ is recommended as that is the lowest one with automatic tests?

test.sh
camilamacedo86: Good spot 👍 tks

cmd/internal/config.go
```diff
@@ -40,10 +35,20 @@ func ConfiguredAndV1() bool {
...
+
+	if os.IsNotExist(err) {
+		return false
+	}
+
+	if err != nil {
+		log.Fatalf("failed to read the configuration file: %v", err)
+	}
+
+	return projectConfig.IsV2()
```
Adirio:
```suggestion
	return !projectConfig.IsV1()
```

This way we will be ready for v3

cmd/main.go
```diff
@@ -78,19 +84,22 @@ func run(options commandOptions) error {
...
 	alphaCmd := newAlphaCmd()
-	// kubebuilder alpha webhook (v1 only)
-	if internal.ConfiguredAndV1() {
-		alphaCmd.AddCommand(newWebhookCmd())
+
+	// kubebuilder update (v1 only)
+	if !internal.IsVersionSupported() {
+		rootCmd.AddCommand(newUpdateCmd())
 	}
+
```
Adirio: Commads are added in alphabetical order, `Update` should be last.

cmd/main.go
```diff
@@ -100,10 +109,8 @@ func buildCmdTree() *cobra.Command {
 	createCmd := newCreateCmd()
 	// kubebuilder create api
 	createCmd.AddCommand(newAPICmd())
-	// kubebuilder create webhook (v2 only)
```
Adirio: Do not remove the whole comment, if you want to remove the v2 only part I'm ok with that

generate_golden.sh
```diff
@@ -37,69 +37,46 @@ scaffold_test_project() {
     cd testdata/$project
     kb=$testdata_dir/../bin/kubebuilder
     oldgopath=$GOPATH
-    if [ $version == "1" ]; then
```
Adirio: What do you think about letting this check but just logging a message saying that v1 tests are not supported?

camilamacedo86: What would be the goal of it since it is just used by contributors?

cmd/internal/config.go
camilamacedo86: Solved.

cmd/main.go
camilamacedo86: Solved.

generate_golden.sh
camilamacedo86: Solved

test.sh
camilamacedo86: Solved as well.

CONTRIBUTING.md
camilamacedo86: It is true. I will revert since it is not part of the scope of this PR as well.

cmd/main.go
camilamacedo86: I did not change the current order and I am not quite sure about the role `Commands are added in alphabetical order, Update should be last`. See that the order was not changed ate all:

Screenshot 2020-02-15 at 14 04 02

cmd/main.go
camilamacedo86: has no need for the comment any more since it has not v1 and v12 webhooks.

camilamacedo86:
/assign @directxman12
/assign @diroot 
/assign @mengqiy 

WDYT?

cmd/main.go
Adirio: You did change the order, newUpdateCmd was added the last and you are adding it first.

Your IDE is comparing the alpha subcommand with the update subcommand.

cmd/main.go
Adirio: Remove the `(v2 only)` part, but leave the `// kubebuilder create webhook` as there are comments documenting all commands, if you see two lineas above you have `// kubebuilder create api`

generate_golden.sh
Adirio: If someone adds a v1 test it will be notified instead of passed silently.

Makefile
```diff
@@ -74,11 +70,11 @@ test: ## Run the go tests ($ go test -v ./cmd/... ./pkg/...)
 
 .PHONY: test-ci
 test-ci: ## Run the unit tests (used in the CI)
-	./setup.sh
 	./test.sh
 
-.PHONY: test-e2e
+.PHONY: test-ci-e2e
```
Adirio: Why are you changing the name? Where are this tests run should not be part of their name.

cmd/main.go
camilamacedo86: Oh. I see what you mean now. Good spot. 
Change has done and closing this one as solved.

generate_golden.sh
camilamacedo86: What you are suggestion is: 

> if [ $version == "1" ]; then
> // do nothing

Why, how someone would add a V1 test since it is deprecated and should no longer have implementations for? And then, if a need as this one started to be required then the if should be added, however, for me make no sense keep an if without need at the moment.

Makefile
camilamacedo86: We have `test-ci` for the unit tests then, make sense keep the same approach to make clear its usage.

cmd/main.go
camilamacedo86: Done

.travis.yml
camilamacedo86: It was removed since as no longe need.

CONTRIBUTING.md
camilamacedo86: Done.

Makefile
Adirio: The target is `test-e2e, having:
```
.PHONY: test-ci-e2e
test-e2e: ...
```
makes no sense.

generate_golden.sh
Adirio: What I suggest is to write a message error so that in the future we don't write `
scaffold_test_project project-v2 1` by mistake.

Makefile
camilamacedo86: Oh. Now I got what do you mean. Solved 👍

generate_golden.sh
camilamacedo86: Sorry, but I still not thinking that it is required and I do not agree with your suggestion. So, I will let it as it is and if others think that it is required as you then I can follow up the suggestion.

Makefile
Adirio: You changed both to `test-ci-e2e` and that wasn't exactly what I meant.

You should probably also have to look in some `.md` files for instructions that used to call `make test-e2e`.

cmd/main.go
```diff
@@ -115,14 +118,14 @@ func buildCmdTree() *cobra.Command {
...
-	}
-
 	// kubebuilder version
 	rootCmd.AddCommand(version.NewVersionCmd())
 
+	// kubebuilder update (v1 only) to allow update the project
+	if !internal.IsVersionSupported() {
+		rootCmd.AddCommand(newUpdateCmd())
+	}
+
```
Adirio: Okey, this was partly my fault. I know I said last, but I forgot about `NewVersionCmd`. It should be just before `NewVersionCmd`.

camilamacedo86: Tks. Make sense done.

cmd/webhook.go
camilamacedo86: Can we close this one as solved?

cmd/main.go
```diff
@@ -115,14 +118,14 @@ func buildCmdTree() *cobra.Command {
...
-	if internal.ConfiguredAndV1() {
+	// kubebuilder update (v1 only) to allow update the project
+	if !internal.IsVersionSupported() {
 		rootCmd.AddCommand(newUpdateCmd())
 	}
 
 	// kubebuilder version
 	rootCmd.AddCommand(version.NewVersionCmd())
-
+
```
Adirio: The spaces or tab here is breaking linting test

droot:
@camilamacedo86 I have assigned `v3.0.0` milestone for this PR because removing support for v1 will be a breaking change. So we should get the comments addressed and merge it only when we are nearing `v3.0.0` release. We are in process of collecting breaking (or big) changes for v3 release, plugin feature is another candidate for v3.

camilamacedo86:
/hold 

Because of:

Screenshot 2020-02-27 at 10 58 57

cmd/api.go
```diff
@@ -163,23 +164,21 @@ func (o *apiOptions) validate(c *config.Config) error {
...
+			resourceExists := false
+			for _, r := range c.Resources {
+				if r.Group == o.resource.Group &&
+					r.Version == o.resource.Version &&
+					r.Kind == o.resource.Kind {
+					resourceExists = true
+					break
 				}
 			}
+			if resourceExists {
```
Adirio:
```suggestion
			if c.HasResource(o.resource.GVK()) {
```

Config already has a function to check if the resource is tracked (== exists): https://github.com/kubernetes-sigs/kubebuilder/blob/f0c25d083cc896c3b3328b8ecb9108a6dd7ed9d8/pkg/model/config/config.go#L71-L86

cmd/main.go
```diff
@@ -115,11 +114,6 @@ func buildCmdTree() *cobra.Command {
 	// kubebuilder init
 	rootCmd.AddCommand(newInitCmd())
 
-	// kubebuilder update (v1 only)
-	if internal.ConfiguredAndV1() {
-		rootCmd.AddCommand(newUpdateCmd())
-	}
```
Adirio: Did you completely remove the `update` command at the end?

camilamacedo86: See that it is not part of https://book.kubebuilder.io/migration/guide.html.
And it has not the purpose of upgrade V1 to V2

cmd/api.go
camilamacedo86: It should be changed in the master. Other PR. 
(not part of the scope)

cmd/main.go
Adirio: Okey perfect, was just verifying that it was intended. You can mark as resolved.

cmd/api.go
Adirio: I'll submit a PR for this.

cmd/webhook.go
```diff
@@ -220,9 +115,9 @@ func (o *webhookV2Options) validate(c *config.Config) error {
 	return nil
 }
 
-func (o *webhookV2Options) scaffolder(c *config.Config) (scaffold.Scaffolder, error) { //nolint:unparam
+func (o *webhookV2Options) scaffolder(c *config.Config) (scaffold.Scaffolder, error) { // nolint:unparam
```
Adirio:
```suggestion
func (o *webhookV2Options) scaffolder(c *config.Config) (scaffold.Scaffolder, error) { //nolint:unparam
```

These spaces were recently removed: read https://github.com/kubernetes-sigs/kubebuilder/pull/1368#discussion_r379111906 for the justification

cmd/webhook.go
```diff
@@ -220,9 +115,9 @@ func (o *webhookV2Options) validate(c *config.Config) error {
...
 	// Load the boilerplate
-	bp, err := ioutil.ReadFile(filepath.Join("hack", "boilerplate.go.txt")) //nolint:gosec
+	bp, err := ioutil.ReadFile(filepath.Join("hack", "boilerplate.go.txt")) // nolint:gosec
```
Adirio:
```suggestion
	bp, err := ioutil.ReadFile(filepath.Join("hack", "boilerplate.go.txt")) //nolint:gosec
```

Same as above

pkg/model/config/config.go
```diff
@@ -113,3 +115,9 @@ func (r GVK) isEqualTo(other GVK) bool {
 		r.Version == other.Version &&
 		r.Kind == other.Kind
 }
+
+// GetV1UnsupportedError returns the error to inform that V1 is no longer supported
+func (config Config) GetV1UnsupportedError() error {
+	return fmt.Errorf(noticeColor, "The v1 projects are deprecated since Feb 1, 2020.\n"+
+		"See how to upgrade your project to v2: https://book.kubebuilder.io/migration/guide.html\n")
+}
```
Adirio: I don't like this being part of this struct:

1. The instance is not used for anything.
2. We could technically call this method from a config that is v2.

```suggestion
// V1UnsupportedError informs that V1 is no longer supported
var V1UnsupportedError= fmt.Errorf(noticeColor, "The v1 projects are deprecated since Feb 1, 2020.\n"+
		"See how to upgrade your project to v2: https://book.kubebuilder.io/migration/guide.html\n")
```

Additionally I don't know if this should be part of `pkg/model/config` or `internal/config` package. I feel like this is not part of the public and thus should be moved to `internal/config`.

pkg/model/resource/options.go
```diff
@@ -164,45 +162,6 @@ func (opts *Options) safeImport(unsafe string) string {
...
-	}
-
-	res.CreateExampleReconcileBody = opts.CreateExampleReconcileBody
-
-	return res
-}
-
 // NewResource creates a new resource from the options
 func (opts *Options) NewResource(c *config.Config, doResource bool) *Resource {
 	res := opts.newResource()
```
Adirio: `newResource` was a non-exported method that was used by both `NewResource` and `NewV1Resource`. As the second is being remvoed, I feel like we should just move what is being done in `newResource` here.

pkg/model/resource/resource_test.go
```diff
@@ -111,33 +89,6 @@ var _ = Describe("Resource", func() {
...
-			Expect(options.Validate()).To(Succeed())
-
-			resource = options.NewResource(singleGroupConfig, true)
-			Expect(resource.Plural).To(Equal("helmswomen"))
-
-			resource = options.NewResource(multiGroupConfig, true)
-			Expect(resource.Plural).To(Equal("helmswomen"))
-
-			resource = options.NewV1Resource(v1Config, true)
-			Expect(resource.Plural).To(Equal("helmswomen"))
```
Adirio: Some of these tests are not v1. If you watch carefully, all the tests have 4 pair of lines, the first pair creates the options and validates it, the second creates a resource for a single-group v2 project, the third creates it for a multi-group v2 project and the forth creates it for a v1 project. So basically only the pairs where `NewV1Resource(v1Config, ...)` is being used should be removed, the rest need to stay.

pkg/model/config/config.go
camilamacedo86: The internal cannot be used in pkg so it cannot be part of the internal/config.
Unfortunately, your suggestions here will force we import the pkg. I would like to not move in this way if you do not mind.

cmd/webhook.go
camilamacedo86: One you suggested add the space and in the other remove. What is the rule?

camilamacedo86: Anyway, I will keep it as the master since it not part of the scope of this PR.

pkg/model/resource/options.go
camilamacedo86: For now, I would like to keep in this PR just the changes to remove the V1 code.  Then, we can do more refac(clean codes) as follow up. 
The reason for it is: keep easier the rebases and the review.

pkg/model/resource/resource_test.go
camilamacedo86: It is true good catch.

cmd/webhook.go
Adirio: Both sugestions are removing, the master has them removed.

pkg/model/config/config.go
Adirio: The `internal/config` can be imported from inside `pkg` subpackages, thats not an issue. There are definetely some examples: https://github.com/kubernetes-sigs/kubebuilder/blob/9552d129f7d12cc6c7072a1d323ed9f4b461d550/pkg/scaffold/api.go#L22

pkg/model/resource/options.go
Adirio: Mark this as resolved, let's do this in a follow up.

pkg/model/resource/resource_test.go
```diff
@@ -136,8 +108,6 @@ var _ = Describe("Resource", func() {
 			resource = options.NewResource(multiGroupConfig, true)
 			Expect(resource.Plural).To(Equal("helmswomen"))
```
Adirio: Minot NIT: there is an extra blank line here.

pkg/model/resource/resource_test.go
```diff
@@ -311,9 +238,6 @@ var _ = Describe("Resource", func() {
 			Expect(resource.Package).To(Equal(path.Join("k8s.io", "api", options.Group, options.Version)))
 			Expect(resource.Domain).To(Equal("authentication.k8s.io"))
```
Adirio: Same minor NIT here: there is an extra blank line.

pkg/model/resource/resource_test.go
Adirio: You can mark this as solved.

cmd/webhook.go
Adirio: Can be marked as solved.

pkg/model/resource/resource_test.go
camilamacedo86: solved

pkg/model/resource/resource_test.go
camilamacedo86: solved

pkg/model/config/config.go
camilamacedo86: done as suggested.

camilamacedo86:
Hi @droot, @DirectXMan12 and @mengqiy

Could we move forward with this one? IHMO it will make easier we review/merge the plugin one since with it will be o lot of fewer files changed for we achieve the goal of the plugin implementation.  

c/c @estroz

mengqiy:
@camilamacedo86  Are we very close to v3?

> IHMO it will make easier we review/merge the plugin one since with it the quantity of files changes will be lower.

@estroz Thoughts?

estroz:
@mengqiy v3 is ready to go, just needs some reviews: #1469. Merging this first would result in less code changes in #1469.

cmd/internal/config.go
```diff
@@ -17,20 +17,15 @@ limitations under the License.
...
 
-const (
-	noticeColor = "\033[1;36m%s\033[0m"
-)
-
-// ConfiguredAndV1 returns true if the project is already configured and it is v1
-func ConfiguredAndV1() bool {
-	projectConfig, err := config.Read()
+// IsConfigured returns true if the project is already configured
+func IsConfigured() bool {
```
mengqiy, line 27: I guess it's easier to follow if using `ProjectFileExists`.

cmd/main.go
```diff
@@ -100,10 +100,9 @@ func buildCmdTree() *cobra.Command {
 	createCmd := newCreateCmd()
 	// kubebuilder create api
 	createCmd.AddCommand(newAPICmd())
-	// kubebuilder create webhook (v2 only)
-	if !internal.ConfiguredAndV1() {
-		createCmd.AddCommand(newWebhookV2Cmd())
-	}
+	// kubebuilder create webhook
+	createCmd.AddCommand(newWebhookV2Cmd())
```
mengqiy, line 104: Probably rename `newWebhookV2Cmd` to `newWebhookCmd`

mengqiy:
/approve
@camilamacedo86 @estroz

cmd/init.go
```diff
@@ -186,21 +170,7 @@ func (o *initOptions) scaffolder(c *config.Config) (scaffold.Scaffolder, error)
 func (o *initOptions) postScaffold(c *config.Config) error {
 	switch {
 	case c.IsV1():
```
estroz, line 172: In some places we keep checks for v1 projects, and in others we remove them. It would be better to centralize the check for unsupported project versions somewhere like `main.go` so we don't have distributed (error-prone) checks.

mengqiy:
Trying to re-trigger Travis.

cmd/internal/config.go
camilamacedo86: Renamed to `HasProjectFile` since return a bool.

cmd/main.go
camilamacedo86: done

cmd/init.go
camilamacedo86: You are right. We have the check the main go already.

camilamacedo86:
Hi @mengqiy and @estroz, 

All nits sugestions in the review addressed. Should we not do a release as v2.4.0 before merging this one? Should not it be part of v3.0.0? Could you help us with @mengqiy ?

camilamacedo86:
/hold 
Until check, if we should not do a release before merge this one.

cmd/init.go
```diff
@@ -185,22 +169,6 @@ func (o *initOptions) scaffolder(c *config.Config) (scaffold.Scaffolder, error)
 
 func (o *initOptions) postScaffold(c *config.Config) error {
 	switch {
-	case c.IsV1():
-		if !o.depFlag.Changed {
```
estroz, line 189: You can remove deprecated flags/fields as well.

cmd/main.go
```diff
@@ -47,6 +48,13 @@ type commandOptions interface {
 
 // run executes a command
 func run(options commandOptions) error {
+	// // Step 0: Before execute any command ensure that the project version is supported
```
estroz, line 51: No need to duplicate this check here since it's already done in `main`.

cmd/main.go
```diff
@@ -130,6 +125,13 @@ func buildCmdTree() *cobra.Command {
 }
 
 func main() {
+	// before create the cmd tree ensure that the project version is supported
+	if internal.HasProjectFile() {
+		if _, err := internal.IsVersionSupported(); err != nil {
+			fmt.Print(err)
```
estroz, line 131:
```suggestion
			log.Fatal(err)
```

internal/config/config.go
```diff
@@ -164,3 +167,9 @@ type saveError struct {
 func (e saveError) Error() string {
 	return fmt.Sprintf("unable to save the configuration: %v", e.err)
 }
+
+// GetV1UnsupportedError returns the error to inform that V1 is no longer supported
+func GetV1UnsupportedError() error {
+	return fmt.Errorf(noticeColor, "The v1 projects are deprecated since Feb 1, 2020 and are no longer supported.\n"+
```
estroz, line 173: This statement is incorrect since v1 projects have been deprecated for a long time, not just since feb. Instead of stating a date, just say they're no longer supported.

camilamacedo86:
/hold cancel

The plugins were removed and it is rebased with the plugin. So, can be merged now.  
@estroz could you help by doing the last check to see if all is ok?

internal/config/config.go
```diff
@@ -179,3 +182,23 @@ type saveError struct {
 func (e saveError) Error() string {
 	return fmt.Sprintf("unable to save the configuration: %v", e.err)
 }
+
+// IsProjectVersionSupported returns true if the project is already configured and it is v2
+func IsProjectVersionSupported() (bool, error) {
```
estroz, line 187: This function is a little weird as-is. The `bool` return value is never used, and returning no error for a non-existent config doesn't make sense. The following is more succinct, and should be used only once in [`pkg/cli/cli.initialize()`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/cli.go#L195):
```suggestion
func (c Config) CheckProjectVersionSupported() error {
	if c.IsV2() || c.IsV3() {
		return nil
	}
	return fmt.Errorf(NoticeColor, "Project version "+c.Version+" is not supported. "+
		"See how to upgrade your project to the latest version: https://book.kubebuilder.io/migration/guide.html")
}
```

pkg/scaffold/api.go
```diff
@@ -86,90 +83,46 @@ func (s *apiScaffolder) newUniverse() *model.Universe {
...
-			&controllerv1.SuiteTest{},
-		); err != nil {
-			return fmt.Errorf("error scaffolding controller: %v", err)
-		}
-	}
-
-	return nil
-}
-
 func (s *apiScaffolder) scaffoldV2() error {
```
estroz, line 86: If we're removing `v2` from scaffold import paths, this should be
```suggestion
func (s *apiScaffolder) scaffold() error {
```

Ditto for other method/function names.

pkg/scaffold/init.go
```diff
@@ -21,18 +21,16 @@ import (
 	"io/ioutil"
 	"path/filepath"
 
+	"sigs.k8s.io/kubebuilder/pkg/scaffold/internal/templates"
+
```
estroz, line 25: Remove newline.

pkg/scaffold/update.go
```diff
@@ -20,7 +20,7 @@ import (
 	"sigs.k8s.io/kubebuilder/pkg/model"
```
estroz, line 20: I think we can fully remove the `update` command right @mengqiy?

test/e2e/v2/e2e_suite.go
```diff
@@ -66,8 +66,7 @@ var _ = Describe("kubebuilder", func() {
 			By("init v2 project")
 			err := kbc.Init(
 				"--project-version", "2",
-				"--domain", kbc.Domain,
-				"--dep=false")
```
estroz, line 70: Needs `--fetch-deps=false`.

test/e2e/v3/e2e_suite.go
```diff
@@ -66,8 +66,7 @@ var _ = Describe("kubebuilder", func() {
 			By("init v3 project")
 			err := kbc.Init(
 				"--project-version", "3-alpha",
-				"--domain", kbc.Domain,
-				"--dep=false")
```
estroz, line 70: Needs `--fetch-deps=false`.

pkg/scaffold/internal/templates/gopkg.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
```
estroz, line 14: This file should be removed.

camilamacedo86: It is used in the update.

estroz: See https://github.com/kubernetes-sigs/kubebuilder/pull/1355#discussion_r417417819.

pkg/scaffold/update.go
estroz: Or at least remove it's logic and check `updateCmd.HasSubCommands()` before adding it, like kubebuilder does with `alpha`.

pkg/scaffold/api.go
camilamacedo86: good catcher

internal/config/config.go
camilamacedo86: > This function is a little weird as-is. The bool return value is never used, and returning no error for a non-existent config doesn't make sense.

Agreed. 

> The following is more succinct, and should be used only once in pkg/cli/cli.initialize():

Follow up your suggestion but kept the Read because it shows not part of the model. For your fully suggestion work we would need to move it for there.

estroz: Why wouldn't this work in `internal/config`?

pkg/scaffold/internal/templates/gopkg.go
camilamacedo86: Removed . I agree that it shows no longer required at all.

camilamacedo86: I agree that we could/should remove. After thinking about and checking it I just think that it should be better done in a follow-up PR. note that we need also change the docs and check all places that are speaking about it. So, shows out of the scope.

internal/config/config.go
camilamacedo86: See that we would need to do the Read outside as well. Also, after check it, I realize that has parts that are using the helpers in the internal and others not. So, IMO it requires a specific cleanup. Could we agree in do it in a follow-up?

camilamacedo86: reverted the helper as we chat about.

pkg/scaffold/update.go
mengqiy: It seems this file is no longer needed.

test/e2e/v3/e2e_suite.go
```diff
@@ -67,7 +67,7 @@ var _ = Describe("kubebuilder", func() {
 			err := kbc.Init(
 				"--project-version", "3-alpha",
 				"--domain", kbc.Domain,
-				"--dep=false")
+				"--fetch-deps=false")
```
mengqiy, line 70: I may miss something, but why do we switch to use the new flag? What's the motiviation?

camilamacedo86: In some point, we received a suggestion to remove the deprecated flags. However, I revert this suggestion. It is not part of the scope of this PR. Here we are using removing the V1 code. So, good catcher. It is using dep again.

pkg/scaffold/update.go
camilamacedo86: I agree but IHMO we should remove the update command in a follow-up PR. 
Then, if we need for any reason to do a revert will be easier. 
The scope here is NOT to remove commands or deprecated args. The goal here is just clean up and remove the v1 code source.

test/e2e/v3/e2e_suite.go
estroz: Is `—dep` not deprecated? It says so in code comments. Was this depreciation unrelated to the v2 change?

camilamacedo86: it is deprecated but still valid for v2. 
We can remove the flag for V3 since is deprecated and I think we should do that. 
However, it shows not part of the scope of this one and then, I think that would be better we have a specific or for that.

pkg/cli/cli.go
```diff
@@ -198,6 +198,11 @@ func (c *cli) initialize() error {
 		return fmt.Errorf("failed to read config: %v", err)
 	}
 
+	if projectConfig.IsV1() {
```
estroz, line 201: This will panic if `err != nil`. Move into the `err == nil` case above, or:
```suggestion
	if c.projectVersion == config.Version1 {
```

camilamacedo86: Done . Good catcher.

test/e2e/v2/e2e_suite.go
```diff
@@ -65,7 +65,6 @@ var _ = Describe("kubebuilder", func() {
 			var controllerPodName string
 			By("init v2 project")
 			err := kbc.Init(
-				"--project-version", "2",
```
estroz, line 68: This is still necessary now that the default project version is `3-alpha`.

pkg/model/config/project_type.go
```diff
@@ -0,0 +1,12 @@
+package config
```
estroz, line 1: Should be removed.

Issue #1226, opened 2019-11-26
[Scaffold enhancement] File interface


Adirio:
# Context

## Current situation

The current `File` interface is as follows:

https://github.com/kubernetes-sigs/kubebuilder/blob/03f70fb1a0f7521f383a08e41562124021161baf/pkg/scaffold/input/input.go#L141-L145

The single method is not descriptive enough and does different tasks:
1. Sets default values
2. Returns a non-descriptive `Input` struct which contains:
    - Two fields needed to prepare the template: `Path` and `TemplateBody`
    - Several other fileds that are not used

Note: the `Input` struct also works as a parent class for mostly all `File`-implementing structs so that they can be injected some data from the scaffold. By mostly I mean all except the `PROJECT` file, which is an special-cased yaml file which doesn't have an embedded `Input`.

## `GetInput` usage

<details><summary>I have tracked <code>GetInput</code> usages.</summary>

### `pkg/scaffold/project.go`
Lines 101 and 106 for v1, 192 and 197 for v2:
https://github.com/kubernetes-sigs/kubebuilder/blob/03f70fb1a0f7521f383a08e41562124021161baf/pkg/scaffold/project.go#L101
https://github.com/kubernetes-sigs/kubebuilder/blob/03f70fb1a0f7521f383a08e41562124021161baf/pkg/scaffold/project.go#L106
https://github.com/kubernetes-sigs/kubebuilder/blob/03f70fb1a0f7521f383a08e41562124021161baf/pkg/scaffold/project.go#L192
https://github.com/kubernetes-sigs/kubebuilder/blob/03f70fb1a0f7521f383a08e41562124021161baf/pkg/scaffold/project.go#L197
The created variables are used in lines 113 and 127 for v1, and 204 and 218 for v2:
https://github.com/kubernetes-sigs/kubebuilder/blob/03f70fb1a0f7521f383a08e41562124021161baf/pkg/scaffold/project.go#L113
https://github.com/kubernetes-sigs/kubebuilder/blob/03f70fb1a0f7521f383a08e41562124021161baf/pkg/scaffold/project.go#L127
https://github.com/kubernetes-sigs/kubebuilder/blob/03f70fb1a0f7521f383a08e41562124021161baf/pkg/scaffold/project.go#L204
https://github.com/kubernetes-sigs/kubebuilder/blob/03f70fb1a0f7521f383a08e41562124021161baf/pkg/scaffold/project.go#L218

[...]

PR #1359, opened 2020-02-07
Template mixins

### LINKED ISSUE #1358: Template mixins
# Current situation

`template.Input` struct is used to inject fields from the scaffolder to the templates such as the domain, the repository or the boilerplate. It is basically a mixin that, when embedded to a template, provides inject functionality.

Having a single mixin for all the fields is neither memory efficient, as most templates only need one or two of these fields, nor performant, as we are injecting fields we are not using.

# Proposal

Instead of having a single mixin for all the injectable fields, several specialized mixins would be used instead.

Each mixin would be implemented by an interface and a struct with a single field and a single method to set this field. For the following example, a made up field `MadeUp` of type `string` is going to be used.

```go
// HasMadeUp allows a made-up to be used on a template
type HasmadeUp interface {
	// InjectMadeUp sets the template made-up
	InjectMadeUp(string)
}

// MadeUpMixin provides templates with a injectable made-up field
type MadeUpMixin struct {
	MadeUp string
}

// InjectMadeUp implements HasMadeUp
func (m *MadeUpMixin) InjectMadeUp(madeUp string) {
	if m.MadeUp == "" {
		m.MadeUp = madeUp
	}
}
```

The interface would be used to inject this field:

```go
func (u Universe) InjectInto(t file.Template) {
	/* ... */
	// Inject made-up
	if u.MadeUp != "" {
		if templateWithMadeUp, hasMadeUp := t.(file.HasMadeUp); hasMadeUp {
			templateWithMadeUp.InjectMadeUp(u.MadeUp)
		}
	}
}
```

And the mixin would be used for those templates that needs that field:

```go
var _ file.Template = &MadeUpTemplate{}

// MadeUpTemplate scaffolds a made up file
type MadeUp struct {
	file.Input
	file.MadeUpMixin
}
```

/kind feature
This belongs to the scaffold enhancement tracked by the umbrella issue #1218

Adirio:
# Description

- 76 file templates did not require the domain but it was being injected.
- 66 file templates did not require the repository but it was being injected.
- 72 file templates did not require the multi-group flag but it was being injected.
- 16 file templates did not require the boilerplate but it was being injected.
- 27 file templates required the resource and they were being provided manually.

This PR creates mixins for domain, repository, multi-group, boilerplate and resource as proposed in #1358, that inject those fields from the universe only on those templates that need it.

# Motivation

This PR is scoped under #1218
Closes #1358

### DIFF
--- a/pkg/model/file/template.go
+++ b/pkg/model/file/template.go
@@ -16,6 +16,10 @@ limitations under the License.
 
 package file
 
+import (
+	"sigs.k8s.io/kubebuilder/pkg/model/resource"
+)
+
 // Input is the input for scaffolding a file
 type Input struct {
 	// Path is the file to write
@@ -88,6 +92,24 @@ type MultiGroup interface {
 // SetVersion sets the MultiGroup value
 func (i *Input) SetMultiGroup(v bool) {
 	i.MultiGroup = v
+}
+
+// HasResource allows a resource to be used on a template
+type HasResource interface {
+	// InjectResource sets the template resource
+	InjectResource(*resource.Resource)
+}
+
+// ResourceMixin provides templates with a injectable resource field
+type ResourceMixin struct {
+	Resource *resource.Resource
+}
+
+// InjectResource implements HasResource
+func (m *ResourceMixin) InjectResource(res *resource.Resource) {
+	if m.Resource == nil {
+		m.Resource = res
+	}
 }
 
 // Template is a scaffoldable file template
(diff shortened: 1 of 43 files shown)


Adirio:
@DirectXMan12 This PR will help with the new file interface that was proposed by #1226 (which is quite outdated). Could you take a look?

Despite saying that it has 43 modified files, most of the logic is in `pkg/model/file/template.go`.
`pkg/model/universe.go` updates the interface names in its inject method, `pkg/scaffold/api.go` and `pkg/scaffold/webhook.go` stop manually providing the resource, and the other 39 files are just using the mixin in the file templates.

Issue #1263, opened 2019-12-13
Option  provided in the create API cannot work


camilamacedo86:
If we run `kubebuilder create api --group webapp --version v1 --kind Guestbook`

> resource=false
> controller=true

NOTE: It will not work since the controller requires the resource in the scaffold. (it is a bug in the master which should not allow this option)

/kind bug

Adirio:
As i said in the comment in 1198, this is not a bug if the resource already exists, either manual creation or previous command with same GVK and `--resource=true --controller=false`

camilamacedo86:
It is a bug because the option appears for the use and when we choose then an error is outputted. So, it is a problem regards usability and user POV. A possible solution is to handle the error and inform the user what and why and how instead of output its stack trace.

Adirio:
You may want to get a controller without a resource for several reasons.

1- Builtin apis controllers. For example the `Namespace` controller that is being checked in the golden test. You do not need a resource as it is already defined in the k8s api. 
2- You already scaffolded the resource with `--controller=false` but you then realize you do need a controller. The resource is already there so that command is valid.
3- You used other tools to scaffold the resource or did it manually. Same as 2, the resource is already there so the command is valid.

For those cases, and probably more that I may be missing, controllers need to be possible to scaffold even with `--resource=false`.

camilamacedo86:
Hi @Adirio, 

I think has some misunderstand here. 

Please, see the above description and try to reproduce the steps before contribute with. Note that if the user follow these steps an **ERROR** will be outputted. As described above, a valid solution in POV would handle the error and provide a msg to allow the user to knows how they should proceed.

Adirio:
If you run `kubebuilder create api ... --resource false --controller true` in an empty new project, it may yield an error. Haven't tried it. If it does, you should include the error message in the first message of these Issue.

But we can't prevent the use of `--resource false --controller true` together as it may be useful if the resource was somehow created previously. For example by a previous `kubebuilder create api ... --resource true --controller false`.

That combination (`--resource false --controller true`) can also be used for builtin resources such as `Namespace`.

So we can't remove `kubebuilder create api ... --resource false --controller true`. The most we can do is, when we are trying to create a controller and no resource file or builtin resource is found, ask the user if he really wants to create a controller for a non-existent resource.

camilamacedo86:
Hi @Adirio, 

[...]

PR #1362, opened 2020-02-09
fix create api command for it be idempotent

camilamacedo86:
**Description**

- Check if has the resource if not it always will scaffold it
- Users are allowed to chosen between [y/n] to scaffold the controllers only.

**Motivation**
If we run `kubebuilder create api --group webapp --version v1 --kind Guestbook`

> resource=false
> controller=true

NOTE: It will break the project since the controller requires the resource in the scaffold. (it is a bug in the master which should not allow this option)

Closes: #1263

**Solution**
Screenshot 2020-02-26 at 09 32 46

### DIFF
--- a/cmd/api.go
+++ b/cmd/api.go
@@ -161,29 +161,27 @@ func (o *apiOptions) validate(c *config.Config) error {
 		o.doController = internal.YesNo(reader)
 	}
 
+	resourceExists := c.HasResource(o.resource.GVK())
+
+	if !c.IsV1() {
+		if !resourceExists && !o.doResource {
+			return errors.New("API resource do not exists. " +
+				"A Controller cannot be create for a resource that do not exist.")
+		}
+	}
+
 	// In case we want to scaffold a resource API we need to do some checks
 	if o.doResource {
 		// Skip the following check for v1 as resources aren't tracked
 		if !c.IsV1() {
 			// Check that resource doesn't exist or flag force was set
 			if !o.force {
-				resourceExists := false
-				for _, r := range c.Resources {
-					if r.Group == o.resource.Group &&
-						r.Version == o.resource.Version &&
-						r.Kind == o.resource.Kind {
-						resourceExists = true
-						break
-					}
-				}
 				if resourceExists {
-					return errors.New("API resource already exists")
+					fmt.Printf("API resource already exists. Creating just the Controller ...")
+					o.doResource = false
 				}
 			}
-		}
 
-		// The following check is v2 specific as multi-group isn't enabled by default
-		if c.IsV2() {
 			// Check the group is the same for single-group projects
 			if !c.MultiGroup {
 				validGroup := true


cmd/api.go
```diff
@@ -165,21 +172,19 @@ func (o *apiOptions) validate(c *config.Config) error {
...
+			resourceExists := false
+			for _, r := range c.Resources {
+				if r.Group == o.resource.Group &&
+					r.Version == o.resource.Version &&
+					r.Kind == o.resource.Kind {
+					resourceExists = true
+					break
 				}
 			}
+			if resourceExists {
```
Adirio: Use `c.HasResource(o.resource.GVK())` instead of the whole loop.

cmd/api.go
```diff
@@ -125,9 +124,6 @@ func (o *apiOptions) bindFlags(cmd *cobra.Command) {
 			"generates an API following an extension pattern (addon)")
 	}
 
-	cmd.Flags().BoolVar(&o.force, "force", false,
```
mengqiy: Dropping a flag is definitely a breaking change.
If we really want to do this, we need to start to deprecate it in v2 and drop it in v3.

Adirio: And we would need a better way to track both external resources and already created CRDs.

camilamacedo86: HI @mengqiy, 

Thank you for your input. Regards your comment, could you please answer my following questions to help me understand how it works? 

> Dropping a flag is definitely a breaking change.
> If we really want to do this, we need to start to deprecate it in v2 and drop it in v3.

Was not thee V1 / V2 done just because of the break changes in the layout of the scaffolded project and in order to still supporting the projects made in the old and new layout? Do we need to create a new structure/layout to scaffold files besides it has no changes at all?   Can we not just release a new tag version 3.0.0 since the breaking change is just in the command?

Regards the change/solution to fix. I will see if we can achieve the same goal without the need to remove the flag.

camilamacedo86: Hi @Adirio and @mengqiy. 

Could you please check the changes made. We can solve the same scenario whiteout a need to introduce a break change in the command tool.

Adirio: **Disclaimer:** the following comment only focuses on whether a flag change should require a amjor version bump or not, as they issue solved by this PR has been considered NOT to be a bug: https://github.com/kubernetes-sigs/kubebuilder/issues/1263#issuecomment-591113294

The kubebuilder version/release (currently `v2.2.0`) vs scaffold version (currently `"2"`) issue has been appearing in several places now. As you have said, removing a flag does not affect the scaffolded files, so the scaffold version could stay as is, but it does affect the CLI, which would require a major version bump for the kubebuilder version/release to `v3.0.0`.

Implementing the plugin system will make the difference clearer, but a third one is going to appear, plugin version/release. Once plugins are implemented, flags belong to plugins as it can be seen in the code fragment below, so the plugin version/release is the one that should be updated.
https://github.com/kubernetes-sigs/kubebuilder/blob/ebd0d8c99aa648778ce6598146667f13b717306e/pkg/plugin/v2/api.go#L94-L116

camilamacedo86:
/cancel approve

It has been discussed in #1263

camilamacedo86:
/approve cancel

It has been discussed in #1263

camilamacedo86:
Closing this one as deferred. 
@mengqiy clarifies the scenarios and why it is not a bug.

PR #1370, opened 2020-02-14
Remove no-op validations

Adirio:
After resouce validation was moved to the `Options` struct (#1255), `Resource.Validate()` is a no-op.

This PR requests removes these `Validate` functions.


Adirio:
@mengqiy Can we merge this? It's just removing empty functions.

`Resource.Validate()` is an empty function, and the rest are empty after removing their call to `Resource.Validate`.

PR #1374, opened 2020-02-15
Remove outdated TODO

Adirio:
The TODO was refering to computing the plural form in this file when a custom plural form was provided. Since #1255 was merged, the plural form is obtained from the resource itself so the issue is no longer valid.

/kind cleanup


Adirio:
Resolved by #1375

PR #1227, opened 2019-11-26
File interface

Adirio:
Closes #1226

Issue #1342, opened 2020-01-31
Cleanup go unit tests


mengqiy:
There are some go unit tests in kubebuilder to test the generated files match expected.
Given that we have the [golden files test](https://github.com/kubernetes-sigs/kubebuilder/blob/master/golden_test.sh), we can cleanup our go unit tests.
We can probably only keep the tests for testing the scaffolding machinery, but not for each generated file.
Maintaining go unit tests requires more efforts than the golden test, since whenever a new file is added or a file is deleted, we need to change the go test to reflect it.

2 actionable items:
- [x] remove go unit tests for scaffolding files
- [ ] ensure enough coverage for the scaffolding machinery

camilamacedo86:
HI @mengqiy, 

The golden tests are e2e tests and have not the same purpose as the unit tests. See that the golden will not cover all scenarios cases as the unit does. IHMO they have different purposes and both should be kept. Also, unit tests are faster and allow us to check, debug and ensure specific points locally which is very helpful for the development purpose. And then, in POV we should add more unit tests instead of just keep the golden one.   

Regards `remove go unit tests for scaffolding files` I understand that it has a completed different purpose. The scaffold tests are to help users test their projects then they are very valuable to them. See for example https://github.com/operator-framework/operator-sdk/issues/2467 a user asking for we scaffold tests for all gen code. However, IHMO it is also a point that should be improved since I do not think that is a good strategy we force users to do their tests using BDD and third-party APIs. See that BDD is may not the best approach for all scenarios and we may have users that would like or not working with them. Also, third-party APIs can be discontinued at any moment which would cause a not good experience for its users since that would mean re-duo all tests implementations in their projects too. So, over this point, I think we should scaffold tests using the go test framework provide in the language/stack instead of.  

Wdyt? Is it make sense?

Adirio:
I think the name of unit tests brings a bit of confussion because they were really like e2e tests disguised as unit tests. They were basically creating a file in memory and checking it against the testdata folder. They were probably left-over code from before golden test was being used for this purpose. The discussion me and @mengqiy were having that brought up his issue was that these disguised tests were overlapping with the golden test and were way harder to maintain as new files required to write additional tests, where the golden test will already understand new files without requiring changes to the tests.

camilamacedo86:
Hi @Adirio, 

[...]

PR #1375, opened 2020-02-17
Update files as part of the Scaffold.Execute

### LINKED ISSUE #1226: [Scaffold enhancement] File interface

### LINKED ISSUE #1358: Template mixins

Adirio:
# Description
Update already scaffolded files inside `Scaffold.Execute` following the same flow used for new files (injecting fields, validating, creating a model, piping to plugins and persisting it to disk).

Coverage reports 72.1% in `pkg/scaffold/internal/filesystem` (the remaining 27.9% are calls to the filesystem that cannot be further tested) and 93.8% in `pkg/scaffold/internal/machinery`, aiming to fulfill #1342.

# Motivation
File updates (i.e., inserting code fragments to already scaffolded files) was not needed for v1. When v2 came, the feature was needed and it was added outside of the main `Scaffold.Execute`. Providing `Scaffold.Execute` the capability of inserting those code fragments achieves a better work flow.

# Implementation details
This PR is made of 3 commits. The first two were submitted as separate PR previously:
1. #1359
2. #1227
I suggest reviewing them commit by commit (GitHub offers the possibility of showing only changes in one commit) as it affects more than 3500 LoC.

Regarding to the last commit, the main changes are found in `pkg/scaffold/internal/machinery/scaffold.go` and the interfaces in `pkg/model/file/interfaces.go`.

This PR is scoped under #1218
Closes #1226
Closes #1227
Closes #1358
Closes #1359


pkg/model/file/template.go
```diff
@@ -20,16 +20,40 @@ import (
...
-type Input struct {
+// Template is a scaffoldable file template
+type Template interface {
+	// GetPath returns the path to the file location
+	GetPath() string
+	// GetBody returns the template body
+	GetBody() string
+	// GetIfExistsAction returns the behavior when creating a file that already exists
+	GetIfExistsAction() IfExistsAction
+	// SetTemplateDefaults returns the TemplateMixin for creating a scaffold file
```
DirectXMan12: this godoc doesn't make sense -- it returns error

pkg/model/file/marker.go
```diff
@@ -0,0 +1,56 @@
...
+
+	panic(fmt.Errorf("unknown file extension: '%s', expected '.go' or '.yaml'", ext))
+}
+
+// String implements Stringer
+func (m Marker) String() string {
+	return m.comment + prefix + m.value
+}
+
+type CodeFragments []string
```
DirectXMan12: needs godoc

pkg/scaffold/internal/filesystem/errors.go
```diff
@@ -24,6 +24,38 @@ import (
...
+	path string
+	err  error
+}
+
+func (e fileExistsError) Error() string {
+	return fmt.Sprintf("failed to check if %s exists: %v", e.path, e.err)
+}
+
+// IsFileExistsError checks if the returned error is because the file could not be checked for existence
+func IsFileExistsError(e error) bool {
```
DirectXMan12: we should eventually switch these all to use `errors.As`

pkg/scaffold/internal/templates/v2/controller/controller_suitetest.go
```diff
@@ -56,6 +58,51 @@ func (f *SuiteTest) Validate() error {
...
+`
+)
+
+// GetCodeFragments implements file.Inserter
+func (f *SuiteTest) GetCodeFragments() file.CodeFragmentsMap {
+	fragments := make(file.CodeFragmentsMap, 2)
+
+	// Generate import code fragments
+	imports := make([]string, 0)
+	imports = append(imports, fmt.Sprintf(apiImportCodeFragment, f.Resource.ImportAlias, f.Resource.Package))
```
DirectXMan12: `imports := []string{fmt.Sprintf(...)}`?

pkg/scaffold/internal/templates/v2/controller/controller_suitetest.go
```diff
@@ -56,6 +58,51 @@ func (f *SuiteTest) Validate() error {
...
+
+	// Generate add scheme code fragments
+	addScheme := make([]string, 0)
+	addScheme = append(addScheme, fmt.Sprintf(addschemeCodeFragment, f.Resource.ImportAlias))
+
+	// Only store code fragments in the map if the slices are non-empty
+	if len(imports) != 0 {
```
DirectXMan12: don't we know that these are non-empty due the the append?

pkg/scaffold/internal/filesystem/errors.go
Adirio: I agree, I submited a PR for that and didnt get any answer, it can be done after this gets merged.

mengqiy:
> Closes #1226
Closes #1227
Closes #1358
Closes #1359

2 of them are PRs. Why?

pkg/scaffold/internal/templates/v2/controller/controller_suitetest.go
Adirio: I did it this way on purpose, to allow easy addition and deletion of code fragments. Do you have a strong opinion on this?

pkg/scaffold/internal/templates/v2/controller/controller_suitetest.go
Adirio: Same as https://github.com/kubernetes-sigs/kubebuilder/pull/1375#discussion_r382309187

Adirio:
> 2 of them are PRs. Why?

The two first commits in this PR are those PRs. Merging this would be like merging the three.

pkg/model/file/marker.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"fmt"
+	"path/filepath"
+)
+
+const prefix = "+kubebuilder:scaffold:"
+
+var commentsByExt = map[string]string{
+	// TODO(v3): machine-readable comments should not have spaces by Go convention. However,
```
mengqiy: Can you create an issue to track everything related to `TODO(v3)`?

Adirio: #1382, #1383, and #1384 raised. Could you add them to the corresponding milestone?

PR #1380, opened 2020-02-21
plugins: validate project and plugin versions

estroz:
This PR implements plugin version format enforcement. See [the plugins design doc](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-1.md#required) and #1290 for more details.

Changes:
* pkg/cli: validate cli struct's fields
* pkg/plugin: validation for plugin version strings
* pkg/internal/validation: project-wide validation logic; move `pkg/model/resource/validation.go` functions here

/cc @DirectXMan12 @mengqiy @hasbro17 @Adirio


pkg/cli/cli.go
```diff
@@ -175,6 +182,37 @@ func (c *cli) initialize() error {
...
+	}
+
+	// Validate plugin versions and name.
+	for _, versionedPlugins := range c.plugins {
+		for _, versionedPlugin := range versionedPlugins {
+			pluginName := versionedPlugin.Name()
+			pluginVersion := versionedPlugin.Version()
+			if err := plugin.ValidateVersion(pluginVersion); err != nil {
+				return fmt.Errorf("failed to validate plugin %q version %q: %v",
+					pluginName, pluginVersion, err)
```
camilamacedo86:
```suggestion
					versionedPlugin.Name(), pluginVersion, err)
```

pkg/cli/cli.go
```diff
@@ -175,6 +182,37 @@ func (c *cli) initialize() error {
...
+		return fmt.Errorf("failed to validate default project version %q: %v", c.defaultProjectVersion, err)
+	}
+	if err := validation.ValidateProjectVersion(c.projectVersion); err != nil {
+		return fmt.Errorf("failed to validate project version %q: %v", c.projectVersion, err)
+	}
+
+	// Validate plugin versions and name.
+	for _, versionedPlugins := range c.plugins {
+		for _, versionedPlugin := range versionedPlugins {
+			pluginName := versionedPlugin.Name()
```
camilamacedo86:
```suggestion
```

camilamacedo86: I think we do not need the var.

pkg/cli/cli.go
camilamacedo86: We can use directly here.

pkg/cli/cli.go
```diff
@@ -175,6 +182,37 @@ func (c *cli) initialize() error {
...
+			pluginVersion := versionedPlugin.Version()
+			if err := plugin.ValidateVersion(pluginVersion); err != nil {
+				return fmt.Errorf("failed to validate plugin %q version %q: %v",
+					pluginName, pluginVersion, err)
+			}
+			for _, projectVersion := range versionedPlugin.SupportedProjectVersions() {
+				if err := validation.ValidateProjectVersion(projectVersion); err != nil {
+					return fmt.Errorf("failed to validate plugin %q supported project version %q: %v",
+						pluginName, projectVersion, err)
```
camilamacedo86:
```suggestion
						versionedPlugin.Name(), projectVersion, err)
```

pkg/cli/cli.go
Adirio: Both `versionedPlugin.Name()` and `versionedPlugin.Version()` are used twice. Three of those four usages are in error messages (where having to access it again will not impact performance as we already found an error). I don't really have a strong opinion about this, but I would lean slightly towards not declaring variables.

pkg/model/validation/validation.go
```diff
@@ -14,28 +14,34 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package resource
+package validation
```
Adirio: From a design point of view I don't like this package. `pkg/model/validation`? Is the validation a model?

I see what you did removing the duplicated code in `pkg/model/resource/validation.go` and `pkg/plugin/internal/validations.go` but maybe we can think of a better path. A `validation` package makes sense, but being it a subpackage of `cli`, `model``, `plugin` or `scaffold` doesn't seem right. My suggestion would be to move it to `pkg/internal/validation/dns.go` as they are all DNS related validations for now. That way we would not have to mantain backwards compatibility of those functions (that are exported) but will still be accessible from almost everywhere. (NOTE: moving `internal/*` to `pkg/internal/*` looks like a nice thing to do if we end up removing the dependencies from `cmd`.

camilamacedo86: I agree with @Adirio here as well. 👍

pkg/cli/cli.go
estroz: Going to leave this as-is.

pkg/cli/cli.go
estroz: Going to leave this as-is.

pkg/cli/cli.go
estroz: Going to leave this as-is.

pkg/cli/cli.go
```diff
@@ -26,6 +26,7 @@ import (
 
 	internalconfig "sigs.k8s.io/kubebuilder/internal/config"
 	"sigs.k8s.io/kubebuilder/pkg/model/config"
+	"sigs.k8s.io/kubebuilder/pkg/internal/validation"
```
Adirio: You need to swap the order

pkg/model/resource/options.go
```diff
@@ -27,6 +27,7 @@ import (
 	"github.com/gobuffalo/flect"
 
 	"sigs.k8s.io/kubebuilder/pkg/model/config"
+	"sigs.k8s.io/kubebuilder/pkg/internal/validation"
```
Adirio: You need to swap the order

pkg/internal/validation/dns.go
```diff
@@ -0,0 +1,114 @@
...
+// to avoid package dependencies. In case of additional functionality from
+// "k8s.io/apimachinery" is needed, re-consider whether to add the dependency.
+
+const (
+	dns1123LabelFmt     string = "[a-z0-9]([-a-z0-9]*[a-z0-9])?"
+	dns1123SubdomainFmt string = dns1123LabelFmt + "(\\." + dns1123LabelFmt + ")*"
+	dns1035LabelFmt     string = "[a-z]([-a-z0-9]*[a-z0-9])?"
+)
+
+type dnsValidationConfig struct {
```
estroz: @Adirio @camilamacedo86 refactored these for convenience after rebasing onto the feature branch, PTAL.

Adirio: I do like the refactor you did, avoiding code duplications through the use of the `dnsValidationConfig` struct and its `check` method. A couple of considerations:

- When I read `IsDNS1XXXLabel` I instantly imagine a `func (string) bool` signature. As the different errors need to be differentiate, we can not change the function signature to that one, so what about changing the name? Maybe around `MustBe...` or `ConformsTo...`. I guess we could check how the standard library 
- Returning a `[]string` containing the errors doesn't seem very idiomatic in Go. Maybe we should just return an `error` and have the user solve the problem in two steps in case he encounters two errors, as the second one will not be reported until the first one is solved.
- Would it make sense to group functions for the same specification? Something in the line of `validation.DNS1XXX.ConformsToLabel(myString)`. You would have either to export the struct instance or an interface to it:

```Golang
type DNSValidator interface {
	ConformsToLabel(string) error
	ConformsToSubdomain(string) error
}

var (
	DNS1023 DNSValidator = dns1023Config
	DNS1135 DNSValidator = dns1135Config
)
```

pkg/internal/validation/dns.go
```diff
@@ -0,0 +1,114 @@
...
+	format   string
+	maxLen   int
+	re       *regexp.Regexp
+	errMsg   string
+	examples []string
+}
+
+var dns1123LabelConfig = dnsValidationConfig{
+	format:   dns1123LabelFmt,
+	maxLen:   56, // = 63 - len("-system")
```
Adirio: I think the `validation` package should not know about the `"-system"` prefix, the method should be called with that already prefixed and the `maxLen` field should be left to 63.

camilamacedo86: Just to clarifies: The project name will be used as part of the namespace added in the files.
So, it will be "<projectname-system>". In this way, when we scaffold the project we cannot allow a projectName with more than  56 characters. However, I agree that how it will be implemented as a hall for improvements like you described. 

However, it may not part of the context of this PR as well.

pkg/internal/validation/dns.go
estroz: The naming and return values could use some work. Doing this in a follow-up sounds good to me.

Adirio:
An `ok-to-test` should be applied for prow tests, right?

PR #1386, opened 2020-02-21

Adirio:
I don't know if this is the approach you want to follow to update that branch when conflicts arose.

/cc @estroz @mengqiy @DirectXMan12


Adirio:
Fail to select target branch

PR #1387, opened 2020-02-21

Adirio:
I don't know if this is the approach you want to follow to update that branch when conflicts arose.

/cc @estroz @mengqiy @DirectXMan12


DirectXMan12:
This is prob fine, since it allows folks w/o direct repo access to update the branch.  I think we'll want to rebase on master and get rid of the merge commits before merging with master, but in the interim we can do this.

/lgtm
/approve

PR #1388, opened 2020-02-21
NITs raised in 1375

Adirio:
As requested in https://github.com/kubernetes-sigs/kubebuilder/pull/1375#pullrequestreview-362318348

NITs:
- https://github.com/kubernetes-sigs/kubebuilder/pull/1375#discussion_r382265396
- https://github.com/kubernetes-sigs/kubebuilder/pull/1375#discussion_r382274049

### DIFF
--- a/pkg/model/file/interfaces.go
+++ b/pkg/model/file/interfaces.go
@@ -40,7 +40,7 @@ type Template interface {
 	Builder
 	// GetBody returns the template body
 	GetBody() string
-	// SetTemplateDefaults returns the TemplateMixin for creating a scaffold file
+	// SetTemplateDefaults sets the default values for templates
 	SetTemplateDefaults() error
 }
 
--- a/pkg/model/file/marker.go
+++ b/pkg/model/file/marker.go
@@ -28,15 +28,17 @@ var commentsByExt = map[string]string{
 	//  this is a backwards incompatible change, and thus should be done for next project version.
 	".go":   "// ",
 	".yaml": "# ",
+	// When adding additional file extensions, update also the NewMarkerFor documentation and error
 }
 
-// Marker represents a comment LoC that will be used to insert code fragments by update operations
+// Marker represents a machine-readable comment that will be used for scaffolding purposes
 type Marker struct {
 	comment string
 	value   string
 }
 
 // NewMarkerFor creates a new marker customized for the specific file
+// Supported file extensions: .go, .ext
 func NewMarkerFor(path string, value string) Marker {
 	ext := filepath.Ext(path)
 	if comment, found := commentsByExt[ext]; found {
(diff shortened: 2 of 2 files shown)


Adirio:
@DirectXMan12 as requested in #1375, the NITs have been tackled here as the other PR was merged.

pkg/model/file/interfaces.go
```diff
@@ -40,7 +40,7 @@ type Template interface {
 	Builder
 	// GetBody returns the template body
 	GetBody() string
-	// SetTemplateDefaults returns the TemplateMixin for creating a scaffold file
+	// SetTemplateDefaults sets the default values for templates
```
camilamacedo86: Nit which we are tryng to address here: 

Screenshot 2020-02-25 at 12 57 11

However, I could not find any implementation of `SetTemplateDefaults` which is returning an error actually. So, hould not bee better we change it  as follows:

```suggestion
	// SetTemplateDefaults sets the default values for templates
	SetTemplateDefaults() 
```

pkg/model/file/marker.go
```diff
@@ -51,6 +51,8 @@ func (m Marker) String() string {
 	return m.comment + prefix + m.value
 }
 
+// CodeFragments represents a set of code fragments
```
camilamacedo86: Could you please add more info? When and why the CodeFragments is used?

pkg/model/file/marker.go
```diff
@@ -51,6 +51,8 @@ func (m Marker) String() string {
...
 type CodeFragments []string
 
+// CodeFragmentsMap binds CodeFragments to their Marker
```
camilamacedo86: Could you please add more info? When and why the CodeFragments is used?
Would be great we doc why we have the  `CodeFragments []string` and `CodeFragmentsMap map[Marker]CodeFragments` as well.

pkg/model/file/interfaces.go
Adirio: I could find at least one case when it returns an error: `pkg/scaffold/internal/templates/v1/gopkg.go`. Skipping the rest of v1 files, will check v2 files now.

Adirio: Found one file in v2 too: `pkg/scaffold/internal/templates/v2/kustomize.go`

Adirio:
Hi @camilamacedo86

I rebased updating the documentation a bit. They are pretty short descriptions because they are part of the model package, they are not restricted to the current usage. That's the reason why I don't include information on how they are used in the model description. Further information on how they are used is provided in the `pkg/scaffold/model.Inserter` interface functions.

pkg/model/file/interfaces.go
camilamacedo86:
```suggestion
	// SetTemplateDefaults will return an error if not be possible to set the default values for templates
```
WDYT about something as ^?

Adirio: IMHO that doesn't add any information as returning an `error` is part of the Go language.

camilamacedo86: The comment is to describe what the func does. So, I understand that `it returns an error if not be possible to set the default values for templates`. Am I right? In this way, I did not get your comment.

mengqiy:
It seems GH didn't get the signal that Travis has succeeded. :-/

Adirio:
> It seems GH didn't get the signal that Travis has succeeded. :-/

I guess the easiest way to solve this is retriggering Travis

PR #1389, opened 2020-02-22
Fix code errors in finalizer example

phillebaba:
Fixes some pointer and return errors in the finalizers example.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
+++ b/docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
@@ -41,8 +41,8 @@ func (r *CronJobReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	ctx := context.Background()
 	log := r.Log.WithValues("cronjob", req.NamespacedName)
 
-	var cronJob batch.CronJob
-	if err := r.Get(ctx, req.NamespacedName, &cronJob); err != nil {
+	var cronJob *batchv1.CronJob
+	if err := r.Get(ctx, req.NamespacedName, cronJob); err != nil {
 		log.Error(err, "unable to fetch CronJob")
 		// we'll ignore not-found errors, since they can't be fixed by an immediate
 		// requeue (we'll need to wait for a new notification), and we can get them
@@ -81,10 +81,13 @@ func (r *CronJobReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			}
 		}
 
-		return ctrl.Result{}, err
+		// Stop reconciliation as the item is being deleted
+		return ctrl.Result{}, nil
 	}
 
-	// rest of the reconciler code
+	// Your reconcile logic
+
+	return ctrl.Result{}, nil
 }
 
 func (r *Reconciler) deleteExternalResources(cronJob *batch.CronJob) error {


phillebaba:
I signed the CLA

docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
```diff
@@ -68,23 +68,23 @@ func (r *CronJobReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 				return ctrl.Result{}, err
 			}
 		}
-
-		return ctrl.Result{}, err
 	}
 
 	// rest of the reconciler code
+
+	return ctrl.Result{}, nil
```
Adirio: The comment `// rest of the reconciler code` is not expected to be empty like in this example, and it shouldn't run if the object is being deleted, so the `return` inside the `else` is needed. If you want to add the `return` also at the end is ok, but it was being included in the `// rest of the reconciler code`.

docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
```diff
@@ -68,23 +68,23 @@ func (r *CronJobReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-			if err := r.Update(context.Background(), cronJob); err != nil {
+			if err := r.Update(context.Background(), &cronJob); err != nil {
 				return ctrl.Result{}, err
 			}
 		}
-
-		return ctrl.Result{}, err
 	}
 
 	// rest of the reconciler code
```
camilamacedo86: In order to make clear, we could change it for:
```suggestion
	// here will be the rest of the reconciler code
```

docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
```diff
@@ -68,23 +68,23 @@ func (r *CronJobReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 
 			// remove our finalizer from the list and update it.
 			cronJob.ObjectMeta.Finalizers = removeString(cronJob.ObjectMeta.Finalizers, myFinalizerName)
-			if err := r.Update(context.Background(), cronJob); err != nil {
+			if err := r.Update(context.Background(), &cronJob); err != nil {
 				return ctrl.Result{}, err
 			}
 		}
-
-		return ctrl.Result{}, err
```
camilamacedo86: Revert and may add a comment.
```suggestion
// It will finish the reconciliation after the actions of the finalizer be done successfully. 
return ctrl.Result{}, nil
```
IHMO return nil here maker easier to understand it, since if `err != nil` then, it will be returned above.

docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
```diff
@@ -68,23 +68,23 @@ func (r *CronJobReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		// The object is being deleted
```
camilamacedo86: Also, the example is missing the import. 

```
	batch "k8s.io/api/batch/v1beta1"
```

docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
```diff
@@ -68,23 +68,23 @@ func (r *CronJobReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 		if containsString(cronJob.ObjectMeta.Finalizers, myFinalizerName) {
 			// our finalizer is present, so lets handle any external dependency
-			if err := r.deleteExternalResources(cronJob); err != nil {
+			if err := r.deleteExternalResources(&cronJob); err != nil {
```
camilamacedo86: Instead of use & here. I would recommend change the var to:

`var cronJob *batch.CronJob`

docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
Adirio: I would go for:
```suggestion
	// rest of the reconciler code
	// Your reconcile logic
```

docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
Adirio: I think the issue here is not that it is missing that, it is that the `batchv1` alias is wrong, it should be `batch`.

camilamacedo86: See [here](https://github.com/kubernetes-sigs/kubebuilder/blob/cb877e9f82a168c7ac80cee1ee5e194e63d8ed18/docs/book/src/cronjob-tutorial/testdata/finalizer_example.go#L24) that the import is not in the k8s API for CronJobs is not in the example @Adirio

docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
Adirio: I agree with returning `nil`. Indent your suggestiong please Camila.

```suggestion
		// Stop reconciliation as the item is being deleted.
		return ctrl.Result{}, nil
```

docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
Adirio: But this is the reconciler for `tutorial.kubebuilder.io/project/api/v1.CronJob` not for `k8s.io/api/batch/v1beta1.CronJob`. Remember this tutorial is creating a clone of the builtin CronJob to provide an example of how a controller can be developed.

camilamacedo86: Make sense. So we need to fix the alias and replace the `batch` for `batchv1`.

Adirio: Exactly

phillebaba:
Thanks for the comments! I think I fixed everything you asked for.

phillebaba:
Done! Everybody seems to have their own rules about whether or not to squash commits, so its hard to keep track of

camilamacedo86:
Hi @mengqiy @DirectXMan12.

It is a small fix in the docs which shows ok to be merged. WDYT?

PR #1391, opened 2020-02-26
:sparkles: bump CR to v0.5.0 and CT to v0.2.5

mengqiy:
This PR needs #1394.


mengqiy:
This is ready to go. And then we can do a release.

PR #1394, opened 2020-02-27
:sparkles: default a LeaderElectionID

mengqiy:
https://github.com/kubernetes-sigs/controller-runtime/pull/446 removed the defaulting for LeaderElectionID.
It makes sense for controller-runtime. But kubebuilder should still be able to scaffold a project that work out-of-box.
So we will give users some instructions to update it in code comment.

### DIFF
--- a/pkg/scaffold/internal/templates/v2/main.go
+++ b/pkg/scaffold/internal/templates/v2/main.go
@@ -18,7 +18,9 @@ package v2
 
 import (
 	"fmt"
+	"hash/fnv"
 	"path/filepath"
+	"text/template"
 
 	"sigs.k8s.io/kubebuilder/pkg/model/file"
 )
@@ -26,11 +28,14 @@ import (
 const defaultMainPath = "main.go"
 
 var _ file.Template = &Main{}
+var _ file.UseCustomFuncMap = &Main{}
 
 // Main scaffolds the controller manager entry point
 type Main struct {
 	file.TemplateMixin
 	file.BoilerplateMixin
+	file.DomainMixin
+	file.RepositoryMixin
 }
 
 // SetTemplateDefaults implements file.Template
@@ -46,6 +51,19 @@ func (f *Main) SetTemplateDefaults() error {
 	)
 
 	return nil
+}
+
+func hash(s string) (string, error) {
+	hasher := fnv.New32a()
+	hasher.Write([]byte(s)) // nolint:errcheck
+	return fmt.Sprintf("%x", hasher.Sum(nil)), nil
+}
+
+// GetFuncMap implements file.UseCustomFuncMap
+func (f *Main) GetFuncMap() template.FuncMap {
+	fm := file.DefaultFuncMap()
+	fm["hash"] = hash
+	return fm
 }
 
 var _ file.Inserter = &MainUpdater{}
@@ -220,8 +238,9 @@ func main() {
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:             scheme,
 		MetricsBindAddress: metricsAddr,
-		LeaderElection:     enableLeaderElection,
 		Port:               9443, 
+		LeaderElection:     enableLeaderElection, 
+		LeaderElectionID:   "{{ hash .Repo }}.{{ .Domain }}",
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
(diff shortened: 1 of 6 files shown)


mengqiy:
This PR is a prerequisite for #1391.

pkg/scaffold/internal/templates/v2/main.go
```diff
@@ -220,8 +220,11 @@ func main() {
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:             scheme,
 		MetricsBindAddress: metricsAddr,
-		LeaderElection:     enableLeaderElection,
 		Port:               9443, 
+		LeaderElection:     enableLeaderElection,
+		// TODO(user): You MUST ensure LeaderElectionID is unique for your controller. Otherwise, 2 controllers using
+		// the same default LeaderElectionID in the same namespace will race and one of them will stop working silently. 
+		LeaderElectionID: "replace-this-with-a-unique-id",
```
Adirio: I would create a unique string from `domain` and `repo`, adding some hashing too. Something like `"{{ .Domain }}/{{ hash .Repo }}"`.

WDYT?

mengqiy:
@Adirio v1 doesn't support leader election.

pkg/scaffold/internal/templates/v2/main.go
mengqiy: I guess I need to change `file.Template` a bit to make it more flexible to pass the additional `FuncMap` around.
```
type Template interface {
	Builder
	// GetBody returns the template body
	GetBody() string
	// SetTemplateDefaults sets the default values for templates
	SetTemplateDefaults() error
	// GetFuncMap allow each template to provide additional functions which will
	// be merged with the default functions.
	GetFuncMap() template.FuncMap
}
```

Adirio: In that case, I would go for:
```Go
type CustomTemplate interface {
	Template
	GetFuncMap() template.FuncMap
}
```
This way only those that need it have to implement it.

But we could also implement it by having a `HashedRepo` field, set it in the `SetTemplateDefaults`, and then use it like `"{{ .Domain }}/{{ .HashedRepo }}"`.

pkg/scaffold/internal/templates/v2/main.go
```diff
@@ -48,6 +53,23 @@ func (f *Main) SetTemplateDefaults() error {
 	return nil
 }
 
+var hasher = fnv.New32a()
```
estroz: Can this not be a local variable in `hash`?

pkg/scaffold/internal/templates/v2/main.go
```diff
@@ -48,6 +53,23 @@ func (f *Main) SetTemplateDefaults() error {
...
+
+func hash(s string) (string, error) {
+	if _, err := hasher.Write([]byte(s)); err != nil {
```
estroz: This `Write` method will never return an error.

pkg/scaffold/internal/templates/v2/main.go
Adirio: That would mean that we are creating a `Hash` every time.

pkg/scaffold/internal/templates/v2/main.go
mengqiy: Yay, but this is to make the linter happy.
I probably can use `// nolint` to disable the checking for this line.

pkg/scaffold/internal/templates/v2/main.go
mengqiy: Is creating a `Hash` an expensive operation?

estroz: [No](https://golang.org/src/hash/fnv/fnv.go?s=1113:1137#L33).

Adirio: No need to reset it then

pkg/scaffold/internal/templates/v2/main.go
mengqiy: @Adirio Any thoughts on current implementation?

mengqiy:
PTAL

pkg/scaffold/internal/templates/v2/main.go
Adirio: The only NIT would be adding the `Template` to the `UseCustomFuncMap`

PR #1396, opened 2020-02-27
Delegate config checks to the config model

Adirio:
As discussed in https://github.com/kubernetes-sigs/kubebuilder/pull/1355#discussion_r385074333

### DIFF
--- a/cmd/api.go
+++ b/cmd/api.go
@@ -163,41 +163,15 @@ func (o *apiOptions) validate(c *config.Config) error {
 
 	// In case we want to scaffold a resource API we need to do some checks
 	if o.doResource {
-		// Skip the following check for v1 as resources aren't tracked
-		if !c.IsV1() {
-			// Check that resource doesn't exist or flag force was set
-			if !o.force {
-				resourceExists := false
-				for _, r := range c.Resources {
-					if r.Group == o.resource.Group &&
-						r.Version == o.resource.Version &&
-						r.Kind == o.resource.Kind {
-						resourceExists = true
-						break
-					}
-				}
-				if resourceExists {
-					return errors.New("API resource already exists")
-				}
-			}
+		// Check that resource doesn't exist or flag force was set
+		if !o.force && c.HasResource(o.resource.GVK()) {
+			return errors.New("API resource already exists")
 		}
 
-		// The following check is v2 specific as multi-group isn't enabled by default
-		if c.IsV2() {
-			// Check the group is the same for single-group projects
-			if !c.MultiGroup {
-				validGroup := true
-				for _, existingGroup := range c.ResourceGroups() {
-					if !strings.EqualFold(o.resource.Group, existingGroup) {
-						validGroup = false
-						break
-					}
-				}
-				if !validGroup {
-					return fmt.Errorf("multiple groups are not allowed by default, to enable multi-group visit %s",
-						"kubebuilder.io/migration/multi-group.html")
-				}
-			}
+		// Check that the provided group can be added to the project
+		if c.IsV2() && !c.MultiGroup && len(c.Resources) != 0 && !c.HasGroup(o.resource.Group) {
+			return fmt.Errorf("multiple groups are not allowed by default, to enable multi-group visit %s",
+				"kubebuilder.io/migration/multi-group.html")
 		}
 	}
 
--- a/pkg/model/config/config.go
+++ b/pkg/model/config/config.go
@@ -16,6 +16,10 @@ limitations under the License.
 
 package config
 
+import (
+	"strings"
+)
+
 // Scaffolding versions
 const (
 	Version1 = "1"
@@ -51,29 +55,8 @@ func (config Config) IsV2() bool {
 	return config.Version == Version2
 }
 
-// ResourceGroups returns unique groups of scaffolded resources in the project
-func (config Config) ResourceGroups() []string {
-	groupSet := map[string]struct{}{}
-	for _, r := range config.Resources {
-		groupSet[r.Group] = struct{}{}
-	}
-
-	groups := make([]string, 0, len(groupSet))
-	for g := range groupSet {
-		groups = append(groups, g)
-	}
-
-	return groups
-}
-
 // HasResource returns true if API resource is already tracked
-// NOTE: this works only for v2, since in v1 resources are not tracked
 func (config Config) HasResource(target GVK) bool {
-	// Short-circuit v1
-	if config.Version == Version1 {
-		return false
-	}
-
 	// Return true if the target resource is found in the tracked resources
 	for _, r := range config.Resources {
 		if r.isEqualTo(target) {
@@ -87,10 +70,10 @@ func (config Config) HasResource(target GVK) bool {
 
 // AddResource appends the provided resource to the tracked ones
 // It returns if the configuration was modified
-// NOTE: this works only for v2, since in v1 resources are not tracked
+// NOTE: in v1 resources are not tracked, so we return false
 func (config *Config) AddResource(gvk GVK) bool {
 	// Short-circuit v1
-	if config.Version == Version1 {
+	if config.IsV1() {
 		return false
 	}
 
@@ -102,6 +85,19 @@ func (config *Config) AddResource(gvk GVK) bool {
 	// Append the resource to the tracked ones, return true
 	config.Resources = append(config.Resources, gvk)
 	return true
+}
+
+// HasGroup returns true if group is already tracked
+func (config Config) HasGroup(group string) bool {
+	// Return true if the target group is found in the tracked resources
+	for _, r := range config.Resources {
+		if strings.EqualFold(group, r.Group) {
+			return true
+		}
+	}
+
+	// Return false otherwise
+	return false
 }
 
 // GVK contains information about scaffolded resources


cmd/api.go
```diff
@@ -163,41 +163,15 @@ func (o *apiOptions) validate(c *config.Config) error {
 
 	// In case we want to scaffold a resource API we need to do some checks
 	if o.doResource {
-		// Skip the following check for v1 as resources aren't tracked
-		if !c.IsV1() {
```
camilamacedo86: We still needing keeping the V1/V2 checks besides the depreciation.

estroz: The `v1` check is still done in [Config.HasResources()](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/model/config/config.go#L73).

Adirio: Both checks are done in the method themselves (this way it will only need to be changed in one place when we remove v1).
- [`Config.HasResource()`](https://github.com/Adirio/kubebuilder/blob/1ac891650d2b641b4eb0b290ac07b93c998eb2a0/pkg/model/config/config.go#L61-L64)
- [`Config.ValidGroup()`](https://github.com/Adirio/kubebuilder/blob/1ac891650d2b641b4eb0b290ac07b93c998eb2a0/pkg/model/config/config.go#L98-L101)

camilamacedo86: It is true. Its perfect!

pkg/model/config/config.go
```diff
@@ -104,6 +93,28 @@ func (config *Config) AddResource(gvk GVK) bool {
 	return true
 }
 
+// ValidGroup checks that the provided group can be added to the project
+func (config Config) ValidGroup(group string) bool {
+	// Only v2 is single-group by default and requires a flag to allow multi-group, short-circuit the rest
+	if config.Version != Version2 {
```
camilamacedo86: I would prefer we use the centralized implementation instead of duplicate it here. ( IsV1/IsV2)

pkg/model/config/config.go
```diff
@@ -104,6 +93,28 @@ func (config *Config) AddResource(gvk GVK) bool {
 	return true
 }
 
+// ValidGroup checks that the provided group can be added to the project
+func (config Config) ValidGroup(group string) bool {
```
camilamacedo86: The nomenclature of methods that return bool should be isX/hasX. So, IMO it should be. isGroupValid().

pkg/model/config/config.go
Adirio: I did it for any `config.Version == VersionX` check.

cmd/api.go
```diff
@@ -163,41 +163,15 @@ func (o *apiOptions) validate(c *config.Config) error {
...
-					}
-				}
-				if !validGroup {
-					return fmt.Errorf("multiple groups are not allowed by default, to enable multi-group visit %s",
-						"kubebuilder.io/migration/multi-group.html")
-				}
-			}
+		// Check that the provided group can be added to the project
+		if !c.IsGroupAllowed(o.resource.Group) {
+			return fmt.Errorf("multiple groups are not allowed by default, to enable multi-group visit %s",
```
camilamacedo86: Since it is a func now, thee specific error should came from the `IsGroupAllowed` because it could have more than one check/error.

Adirio: `IsGroupAllowed` doesn't have any error. You will be able to call that function anytime and it won't break. The error is in validating the api that doesn't fulfill one of the requirements (the group being allowed).

camilamacedo86: Since`IsGroupAllowed`  can do more than one check, so why it would have here a 1  specific error?

```
if !c.IsGroupAllowed(o.resource.Group) {
			return fmt.Errorf("multiple groups are not allowed by default, to enable multi-group visit %s",
 ...
```

Adirio: There is only one reason why the group may not be allowed, being v2, single-group, and already existing a group with a different name. The other checks return `true`, they are just shot-circuits that we need to have to support different versions and multi-group v2.

camilamacedo86: Currently, yes. However, the change allows it to be introduced. Another option is to make clear what will be checked by replacing the name for `isMultiGroupAllowed`. I would prefer this one.

Adirio: Summing up the offline discussion: having a generic method `IsGroupAllowed` being the only check for that error will mean that if in the future another check is added to that function, an unexpected error may be returned. However moving the api-specific error to the `model` package doesn't seem right either. Instead we moved to a `HasGroup` approach with extra checks in the API side to check for v2 and single-group.

pkg/model/config/config.go
```diff
@@ -104,6 +93,25 @@ func (config *Config) AddResource(gvk GVK) bool {
 	return true
 }
 
+// HasGroup returns true if group is already tracked
+// NOTE: in v1 resources are not tracked, so we return false
+func (config Config) HasGroup(group string) bool {
+	// Short-circuit v1
+	if config.IsV1() {
+		return false
+	}
```
camilamacedo86: just remove it .. if HasGroup should not check if isV1 or V2. 
if isV1 the config.Resources will be empty and it will return false.

estroz: From the method comment:
> NOTE: in v1 resources are not tracked, so we return false
Checking version maintains legacy behavior.

camilamacedo86: hi @estroz,

The ` legacy behaviour` was check in the command the version in order to verify if is MultiGroup enabled or not since it is a feature required only for V2(V1 allows multigroup by default). Then, in POV a new method `HasGroup` in the config which can be used in any place not requires it.

camilamacedo86: Hi @estroz,

Really tks for the input.

However, note that the legacy behaviour was; when the api command is performed then check if is multigroup supported or not (v2 do not support multigroup by default)

I do not think that HasGroup is reasonable for check versions, and also it is not required at all.

Adirio: Both `HasGroup` and `HasResource` had shortcircuits for v1, but as v1 doesn't track resources the behavior without the shortcircuits was the same, so I guess we can remove them.

cmd/api.go
```diff
@@ -163,41 +163,15 @@ func (o *apiOptions) validate(c *config.Config) error {
...
-						break
-					}
-				}
-				if !validGroup {
-					return fmt.Errorf("multiple groups are not allowed by default, to enable multi-group visit %s",
-						"kubebuilder.io/migration/multi-group.html")
-				}
-			}
+		// Check that the provided group can be added to the project
+		if c.IsV2() && !c.MultiGroup && len(c.Resources) != 0 && !c.HasGroup(o.resource.Group) {
```
camilamacedo86:
```suggestion
		if c.IsV2() && !c.MultiGroup && !c.HasGroup(o.resource.Group) {
```

pkg/model/config/config.go
```diff
@@ -104,6 +87,19 @@ func (config *Config) AddResource(gvk GVK) bool {
 	return true
 }
 
+// HasGroup returns true if group is already tracked
+func (config Config) HasGroup(group string) bool {
+	// Return true if the target group is found in the tracked resources
```
camilamacedo86:
```suggestion
	// Return true if the target group is found in the tracked resources
	
	if  len(config.Resources) < 1 {
	      return false
	}
```

pkg/model/config/config.go
```diff
@@ -87,10 +70,10 @@ func (config Config) HasResource(target GVK) bool {
...
 // AddResource appends the provided resource to the tracked ones
 // It returns if the configuration was modified
-// NOTE: this works only for v2, since in v1 resources are not tracked
+// NOTE: in v1 resources are not tracked, so we return false
 func (config *Config) AddResource(gvk GVK) bool {
 	// Short-circuit v1
-	if config.Version == Version1 {
+	if config.IsV1() {
 		return false
 	}
```
camilamacedo86:
```suggestion
```
I do not think that it should be added here as well. The V1/V2 checks are the responsibility of the command impl. If for the command A addResources just can be done when isV1 then this validation should be made in the command as itt was done before.

pkg/model/config/config.go
```diff
@@ -87,10 +70,10 @@ func (config Config) HasResource(target GVK) bool {
 
 // AddResource appends the provided resource to the tracked ones
 // It returns if the configuration was modified
-// NOTE: this works only for v2, since in v1 resources are not tracked
+// NOTE: in v1 resources are not tracked, so we return false
```
camilamacedo86:
```suggestion
```

cmd/api.go
```diff
@@ -163,41 +163,15 @@ func (o *apiOptions) validate(c *config.Config) error {
 
 	// In case we want to scaffold a resource API we need to do some checks
 	if o.doResource {
```
camilamacedo86:
```suggestion
	if o.doResource {
	     if !c.isV1() {
	
```

cmd/api.go
```diff
@@ -163,41 +163,15 @@ func (o *apiOptions) validate(c *config.Config) error {
...
-					return fmt.Errorf("multiple groups are not allowed by default, to enable multi-group visit %s",
-						"kubebuilder.io/migration/multi-group.html")
-				}
-			}
+		// Check that the provided group can be added to the project
+		if c.IsV2() && !c.MultiGroup && len(c.Resources) != 0 && !c.HasGroup(o.resource.Group) {
+			return fmt.Errorf("multiple groups are not allowed by default, to enable multi-group visit %s",
+				"kubebuilder.io/migration/multi-group.html")
 		}
 	}
```
camilamacedo86:
```suggestion
}
```

camilamacedo86: Just to clarifies, the suggestion on this is:


```go
	if o.doResource {
                 if !c.IsV1 {
		        if !o.force && c.HasResource(o.resource.GVK()) {
	 	        }

		        // Check that the provided group can be added to the project
		        if !c.MultiGroup !c.HasGroup(o.resource.Group) {
			    return fmt.Errorf("multiple groups are not allowed by default, to enable multi-group visit %s",
				"kubebuilder.io/migration/multi-group.html")
		        }
                 }
		
	}
```

cmd/api.go
Adirio: `c.IsV2()` is needed. Imagine you do a `kubebuilder init --version=1 ...` and then `kubebuilder create api ...`. `!c.MultiGroup && !c.HasGroup(o.resource.Group)` will be true and print the error when it shouldn't.

`len(c.Resources) != 0` is needed for the first resource. Imagine you do a `kubebuilder init ...` then  a `kubebuilder create api ...`. `!c.MultiGroup && !c.HasGroup(o.resource.Group)` will be true printing the error when it is the first resource so it should be admited.

pkg/model/config/config.go
Adirio: That check is useless, if `len(config.Resources) == 0` then loop will finish inmediately, we do not need to check that it has any item first.

pkg/model/config/config.go
Adirio: This is needed, if we do not add it here v1 projects will add tracked resources and they didn't. I removed every check that was not needed but this is needed. Not having it completely breaks v1.

pkg/model/config/config.go
Adirio: As I said above, this check is needed so the note has to stay.

cmd/api.go
Adirio: So basically your suggestion is to move the version check to the scaffolder insted of having it in the config model? Okey, I'm fine with that.

Adirio: First of all, the check should be `c.IsV2()` instead of `!c.IsV1()`. While currently they are the same, as there is only v1 and v2, they have different semantics.

Saying `c.IsV2()` means thata this is specific logic of v2 while `!c.IsV1()` means that this is general logic of every version except v1. If we use `!c.IsV1()` we are expecting every future version to behave the same v2 does in this sense (having a force flag, a multi-group flag, etc.).   This is setting expectation on future versions, which is a bad coding pattern.

Additionally, `c.IsV2()` is not needed for the first of these two checks, as `c.Resource(...)` will always return `false` for v1 as it doesn't track any.

The `len(c.Resources) != 0` that you removed is needed. Imagine you do a `kubebuilder init ...` then  a `kubebuilder create api ...`. `!c.MultiGroup && !c.HasGroup(o.resource.Group)` will be true printing the error when it is the first resource so it should be admited.

Does it make sense?

pkg/model/config/config.go
Adirio: If we do this, v1 configuration file will track resources. It will break backwards compatibility.

Issue #1378, opened 2020-02-20
Feature Request: Plugins Phase 2


estroz:
## Overview

Phase 1 of the proposed [scaffold plugin system](https://github.com/kubernetes-sigs/kubebuilder/blob/5187a15/designs/extensible-cli-and-scaffolding-plugins-phase-1.md) is being implemented (#1290) at the time of writing. Progress on phase 2, which adds support for chaining plugins together (more below), will begin shortly afterwards, starting with a proposal doc. Before writing that doc, which will contain implementation details for chaining, I'd like to start a general discussion of how chaining will occur.

<details>
<summary>Phase 2 requirements, summarized from previous discussions</summary>

1. [mdbook](https://github.com/kubernetes-sigs/kubebuilder/tree/5187a15/docs/book)-like structure
	1. A Plugin receives a JSON blob "universe" from upstream plugins containing file "nodes", each of which contain file metadata.
	1. The plugin can add or delete nodes (nodes are not mutable), then returns the modified universe so downstream plugins can do the same.
1. A debug function dumps the universe with configurable verbosity and some intelligence as to which node failed (latter is a nice-to-have).
1. Pre-generate a project file (for complex plugin config use cases).
1. Enhanced ["layout"](https://github.com/kubernetes-sigs/kubebuilder/blob/5187a15/designs/extensible-cli-and-scaffolding-plugins-phase-1.md) PROJECT file key.
	1. Either and ordered list of plugin `{name}/{version}`, or ordered map of `{name}/{version}` to plugin-specific config.
	1. Order matters for certain plugins, ex. a bazel plugin would need to run after all Go files are generated.

</details>

## Initial thoughts

Hypothetical CLI setup: a global `--plugins` flag that takes an ordered list of plugin names. The subcommand invoked with `--plugins` executes "downstream" plugins from that list that match those the kubebuilder binary knows about.

Given that we want to pass some initial universe between plugins, that universe should be initialized by the plugin invoked via CLI (the "base" plugin). For example, `kubebuilder init --plugins addon` will invoke an `Init` plugin, which passes its generated state to the [`addon`](https://github.com/kubernetes-sigs/kubebuilder/tree/master/plugins/addon) plugin, modifying scaffolded files then writing them to disk.

## Open questions

Assuming this flow is what we want, a few questions pop up:
1. For the initial state: at what point do we pass a base plugin's state to downstream plugins?
[...]

PR #1403, opened 2020-03-03
feature/plugins: config `layout`, `--plugins` flag, and plugin name validation

estroz:
**Description of Change:**
* pkg/cli: `--plugins` allows a binary invoker to select which plugins known to a `CLI` to run; this flag can be set with a shortened (non-fully-qualified) name. Either a plugin name passed to `--plugins` or the name set in an initialized project's config layout, in order of precedence, is resolved to a known plugin.
* pkg/model/config: add project file `layout` key, containing the `Init` plugin key that scaffolded a project
* pkg/{cli,plugin}: add plugin name validation
* pkg/plugin/v2: set `layout` key in config

**Motivation for Change:** finishes up phase 1 of plugin implementation. The `layout` key determines which plugin (name and version) initialized a project, useful for a plugin implementation to deduce which post-init plugins should be called. `--plugins`/`layout`, in combination with a default plugin, are necessary to select which plugin to use if there are multiple viable plugins per project version.

TODO:
- [x] instead of getting first `*Getter` type, infer by config layout or default (in the case of `Init`)
- [ ] unit tests in `pkg/cli`

/cc @Adirio @hasbro17 @camilamacedo86 
/kind feature

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -31,6 +31,10 @@ func main() {
 			&pluginv1.Plugin{},
 			&pluginv2.Plugin{},
 		),
+		cli.WithDefaultPlugins(
+			&pluginv1.Plugin{},
+			&pluginv2.Plugin{},
+		),
 		cli.WithExtraCommands(
 			newEditCmd(),
 			newUpdateCmd(),
--- a/designs/extensible-cli-and-scaffolding-plugins-phase-1.md
+++ b/designs/extensible-cli-and-scaffolding-plugins-phase-1.md
@@ -62,7 +62,7 @@ project versions (via `SupportedProjectVersions()`).
 Example `PROJECT` file:
 
 ```yaml
-version: "3-alpha"
+version: "2"
 layout: go/v1.0.0
 domain: testproject.org
 repo: github.com/test-inc/testproject
--- a/pkg/cli/api.go
+++ b/pkg/cli/api.go
@@ -55,14 +55,9 @@ func (c cli) newAPIContext() plugin.Context {
 }
 
 func (c cli) bindCreateAPI(ctx plugin.Context, cmd *cobra.Command) {
-	versionedPlugins, err := c.getVersionedPlugins()
-	if err != nil {
-		cmdErr(cmd, err)
-		return
-	}
 	var getter plugin.CreateAPIPluginGetter
 	var hasGetter bool
-	for _, p := range versionedPlugins {
+	for _, p := range c.resolvedPlugins {
 		tmpGetter, isGetter := p.(plugin.CreateAPIPluginGetter)
 		if isGetter {
 			if hasGetter {
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -17,12 +17,13 @@ limitations under the License.
 package cli
 
 import (
-	"errors"
 	"fmt"
 	"log"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 
 	internalconfig "sigs.k8s.io/kubebuilder/internal/config"
 	"sigs.k8s.io/kubebuilder/pkg/internal/validation"
@@ -35,6 +36,10 @@ const (
 	runInProjectRootMsg = `For project-specific information, run this command in the root directory of a
 project.
 `
+
+	projectVersionFlag = "project-version"
+	helpFlag           = "help"
+	pluginNamesFlag    = "plugins"
 )
 
 // CLI interacts with a command line interface.
@@ -58,21 +63,32 @@ type cli struct {
 	projectVersion string
 	// True if the project has config file.
 	configured bool
+	// Whether the command is requesting help.
+	doGenericHelp bool
+
+	// Plugins injected by options.
+	pluginsFromOptions map[string][]plugin.Base
+	// Default plugins injected by options.
+	defaultPluginsFromOptions map[string][]plugin.Base
+	// A mapping of plugin name to version passed by to --plugins.
+	cliPluginKeys map[string]string
+	// A filtered set of plugins that should be used by command constructors.
+	resolvedPlugins []plugin.Base
 
 	// Base command.
 	cmd *cobra.Command
 	// Commands injected by options.
 	extraCommands []*cobra.Command
-	// Plugins injected by options.
-	plugins map[string][]plugin.Base
 }
 
 // New creates a new cli instance.
 func New(opts ...Option) (CLI, error) {
 	c := &cli{
-		commandName:           "kubebuilder",
-		defaultProjectVersion: internalconfig.DefaultVersion,
-		plugins:               map[string][]plugin.Base{},
+		commandName:               "kubebuilder",
+		defaultProjectVersion:     internalconfig.DefaultVersion,
+		pluginsFromOptions:        make(map[string][]plugin.Base),
+		defaultPluginsFromOptions: make(map[string][]plugin.Base),
+		cliPluginKeys:             make(map[string]string),
 	}
 	for _, opt := range opts {
 		if err := opt(c); err != nil {
@@ -113,10 +129,25 @@ func WithPlugins(plugins ...plugin.Base) Option {
 	return func(c *cli) error {
 		for _, p := range plugins {
 			for _, version := range p.SupportedProjectVersions() {
-				if _, ok := c.plugins[version]; !ok {
-					c.plugins[version] = []plugin.Base{}
+				if _, ok := c.pluginsFromOptions[version]; !ok {
+					c.pluginsFromOptions[version] = []plugin.Base{}
 				}
-				c.plugins[version] = append(c.plugins[version], p)
+				c.pluginsFromOptions[version] = append(c.pluginsFromOptions[version], p)
+			}
+		}
+		return nil
+	}
+}
+
+// WithDefaultPlugins is an Option that sets the cli's default plugins.
+func WithDefaultPlugins(plugins ...plugin.Base) Option {
+	return func(c *cli) error {
+		for _, p := range plugins {
+			for _, version := range p.SupportedProjectVersions() {
+				if _, ok := c.defaultPluginsFromOptions[version]; !ok {
+					c.defaultPluginsFromOptions[version] = []plugin.Base{}
+				}
+				c.defaultPluginsFromOptions[version] = append(c.defaultPluginsFromOptions[version], p)
 			}
 		}
 		return nil
@@ -134,12 +165,20 @@ func WithExtraCommands(cmds ...*cobra.Command) Option {
 
 // initialize initializes the cli.
 func (c *cli) initialize() error {
+	// Initialize cli with globally-relevant flags or flags that determine
+	// certain plugin type's configuration.
+	if err := c.parseBaseFlags(); err != nil {
+		return err
+	}
+
 	// Configure the project version first for plugin retrieval in command
 	// constructors.
 	projectConfig, err := internalconfig.Read()
 	if os.IsNotExist(err) {
 		c.configured = false
-		c.projectVersion = c.defaultProjectVersion
+		if c.projectVersion == "" {
+			c.projectVersion = c.defaultProjectVersion
+		}
 	} else if err == nil {
 		c.configured = true
 		c.projectVersion = projectConfig.Version
@@ -151,6 +190,40 @@ func (c *cli) initialize() error {
 	// out before an error resulting from an incorrect cli is returned downstream.
 	if err = c.validate(); err != nil {
 		return err
+	}
+
+	// With or without a config, a user can:
+	// 1. Not set --plugins
+	// 2. Set --plugins to a plugin, ex. --plugins=go
+	// In case 1, either default plugins or the config layout can be used to
+	// determine which plugin to use. Otherwise, the value passed to --plugins
+	// is used. Since both layout and a --plugins value are user input, they
+	// may need to be resolved to known plugin keys.
+	var filterKeys map[string]string
+	switch {
+	case len(c.cliPluginKeys) != 0:
+		// Filter plugins by keys passed in CLI.
+		filterKeys = c.cliPluginKeys
+	case c.configured && !projectConfig.IsV1():
+		// All non-v1 configs must have a layout key. This check will help with
+		// migration.
+		if projectConfig.Layout == "" {
+			return fmt.Errorf("config must have a layout value")
+		}
+		// Use defaults if no config is found, otherwise infer plugins from the
+		// config's layout.
+		name, version := plugin.SplitKey(projectConfig.Layout)
+		filterKeys = map[string]string{name: version}
+	default:
+		// Use the default plugins for this project version.
+		c.resolvedPlugins = c.defaultPluginsFromOptions[c.projectVersion]
+	}
+	if len(filterKeys) != 0 {
+		plugins := c.pluginsFromOptions[c.projectVersion]
+		c.resolvedPlugins, err = resolvePluginsByKeys(plugins, filterKeys)
+		if err != nil {
+			return err
+		}
 	}
 
 	c.cmd = c.buildRootCmd()
@@ -166,17 +239,49 @@ func (c *cli) initialize() error {
 	}
 
 	// Write deprecation notices after all commands have been constructed.
-	if c.projectVersion != "" {
-		versionedPlugins, err := c.getVersionedPlugins()
-		if err != nil {
-			return err
+	for _, p := range c.resolvedPlugins {
+		if d, isDeprecated := p.(plugin.Deprecated); isDeprecated {
+			fmt.Printf(noticeColor, fmt.Sprintf("[Deprecation Notice] %s\n\n",
+				d.DeprecationWarning()))
 		}
-		for _, p := range versionedPlugins {
-			if d, isDeprecated := p.(plugin.Deprecated); isDeprecated {
-				fmt.Printf(noticeColor, fmt.Sprintf("[Deprecation Notice] %s\n\n",
-					d.DeprecationWarning()))
-			}
+	}
+
+	return nil
+}
+
+// parseBaseFlags parses the command line arguments, looking for flags that
+// affect initialization of a cli. An error is returned only if an error
+// unrelated to flag parsing occurs.
+func (c *cli) parseBaseFlags() error {
+	// Create a dummy "base" flagset to populate from CLI args.
+	fs := pflag.NewFlagSet("base", pflag.ExitOnError)
+	fs.ParseErrorsWhitelist = pflag.ParseErrorsWhitelist{UnknownFlags: true}
+
+	// Set base flags that require pre-parsing to initialize c.
+	fs.StringVar(&c.projectVersion, projectVersionFlag, c.defaultProjectVersion, "project version")
+	help := false
+	fs.BoolVarP(&help, helpFlag, "h", false, "print help")
+	pluginKeys := []string{}
+	fs.StringSliceVar(&pluginKeys, pluginNamesFlag, nil, "plugins to run")
+
+	// Parse current CLI args outside of cobra.
+	err := fs.Parse(os.Args[1:])
+	// User needs *generic* help if args are incorrect or --help is set and
+	// --project-version is not set. Plugin-specific help is given if a
+	// plugin.Context is updated, which does not require this field.
+	c.doGenericHelp = err != nil || help && !fs.Lookup(projectVersionFlag).Changed
+
+	// Parse plugin keys into a more manageable data structure (map) and check
+	// for duplicates.
+	for _, key := range pluginKeys {
+		pluginName, pluginVersion := plugin.SplitKey(key)
+		if pluginName == "" {
+			return fmt.Errorf("plugin key %q must at least have a name", key)
 		}
+		if _, exists := c.cliPluginKeys[pluginName]; exists {
+			return fmt.Errorf("duplicate plugin name %q", pluginName)
+		}
+		c.cliPluginKeys[pluginName] = pluginVersion
 	}
 
 	return nil
@@ -192,24 +297,73 @@ func (c cli) validate() error {
 		return fmt.Errorf("failed to validate project version %q: %v", c.projectVersion, err)
 	}
 
+	if _, versionFound := c.pluginsFromOptions[c.projectVersion]; !versionFound {
+		return fmt.Errorf("no plugins for project version %q", c.projectVersion)
+	}
+	// If --plugins is not set, no layout exists, and no defaults exist, we
+	// cannot know which Getter to use for a plugin type.
+	if !c.configured && len(c.cliPluginKeys) == 0 {
+		_, versionExists := c.defaultPluginsFromOptions[c.projectVersion]
+		if !versionExists {
+			return fmt.Errorf("no default plugins for project version %s", c.projectVersion)
+		}
+	}
+
 	// Validate plugin versions and name.
-	for _, versionedPlugins := range c.plugins {
-		for _, versionedPlugin := range versionedPlugins {
-			pluginName := versionedPlugin.Name()
-			pluginVersion := versionedPlugin.Version()
+	for _, versionedPlugins := range c.pluginsFromOptions {
+		if err := validatePlugins(versionedPlugins...); err != nil {
+			return err
+		}
+	}
+	for _, defaultPlugins := range c.defaultPluginsFromOptions {
+		if err := validatePlugins(defaultPlugins...); err != nil {
+			return err
+		}
+	}
+
+	// Validate plugin keys set in CLI.
+	for pluginName, pluginVersion := range c.cliPluginKeys {
+		if err := plugin.ValidateName(pluginName); err != nil {
+			return fmt.Errorf("failed to validate plugin name %q: %v", pluginName, err)
+		}
+		// CLI-set plugins do not have to contain a version.
+		if pluginVersion != "" {
 			if err := plugin.ValidateVersion(pluginVersion); err != nil {
 				return fmt.Errorf("failed to validate plugin %q version %q: %v",
 					pluginName, pluginVersion, err)
 			}
-			for _, projectVersion := range versionedPlugin.SupportedProjectVersions() {
-				if err := validation.ValidateProjectVersion(projectVersion); err != nil {
-					return fmt.Errorf("failed to validate plugin %q supported project version %q: %v",
-						pluginName, projectVersion, err)
-				}
-			}
 		}
 	}
 
+	return nil
+}
+
+// validatePlugins validates the name and versions of a list of plugins.
+func validatePlugins(plugins ...plugin.Base) error {
+	pluginNameSet := make(map[string]struct{}, len(plugins))
+	for _, p := range plugins {
+		pluginName := p.Name()
+		if err := plugin.ValidateName(pluginName); err != nil {
+			return fmt.Errorf("failed to validate plugin name %q: %v", pluginName, err)
+		}
+		pluginVersion := p.Version()
+		if err := plugin.ValidateVersion(pluginVersion); err != nil {
+			return fmt.Errorf("failed to validate plugin %q version %q: %v",
+				pluginName, pluginVersion, err)
+		}
+		for _, projectVersion := range p.SupportedProjectVersions() {
+			if err := validation.ValidateProjectVersion(projectVersion); err != nil {
+				return fmt.Errorf("failed to validate plugin %q supported project version %q: %v",
+					pluginName, projectVersion, err)
+			}
+		}
+		// Check for duplicate plugin names. Names outside of a version can
+		// conflict because multiple project versions of a plugin may exist.
+		if _, seen := pluginNameSet[pluginName]; seen {
+			return fmt.Errorf("two plugins have the same name: %q", pluginName)
+		}
+		pluginNameSet[pluginName] = struct{}{}
+	}
 	return nil
 }
 
@@ -219,6 +373,7 @@ func (c cli) buildRootCmd() *cobra.Command {
 	configuredAndV1 := c.configured && c.projectVersion == config.Version1
 
 	rootCmd := c.defaultCommand()
+	rootCmd.PersistentFlags().StringSlice(pluginNamesFlag, nil, "plugins to run")
 
 	// kubebuilder alpha
 	alphaCmd := c.newAlphaCmd()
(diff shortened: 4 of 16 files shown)


pkg/model/config/config.go
```diff
@@ -20,6 +20,7 @@ const (
 	// Scaffolding versions
 	Version1 = "1"
 	Version2 = "2"
+	Version3 = "3-alpha"
```
camilamacedo86: I am not sure if we should use alpha here.

pkg/plugin/plugin.go
```diff
@@ -38,9 +38,32 @@ type Base interface {
...
+		return key, ""
+	}
+	keyParts := strings.Split(key, "/")
+	return keyParts[0], keyParts[1]
+}
+
+// GetShortName returns plugin's short name (name before domain) if name
+// is fully qualified (has a domain suffix), otherwise GetShortName returns name.
+func GetShortName(name string) string {
+	return strings.SplitN(name, ".", 2)[0]
```
Adirio: No need to make `SplitN` keep splitting after the first dot if we only want the first part, even if every domain needs to have at least 2 parts.

```suggestion
	return strings.SplitN(name, ".", 1)[0]
```

pkg/plugin/v2/plugin.go
```diff
@@ -20,6 +20,13 @@ import (
 	"sigs.k8s.io/kubebuilder/pkg/plugin"
 )
 
+const (
+	pluginName    = "go" + plugin.DefaultNameQualifier
+	pluginVersion = "v2.0.0"
+)
+
+var supportedProjectVersions = []string{"2", "3-alpha"}
```
Adirio: (Remember I consider we should not use v3-alpha)

pkg/cli/cli.go
```diff
@@ -248,20 +335,90 @@ func (c cli) buildRootCmd() *cobra.Command {
...
+
+			switch {
+			// Exact match.
+			case pluginName == longName && pluginVersion == versionedPlugin.Version():
+				definitelyMatch = append(definitelyMatch, versionedPlugin)
+				hasDefinitely = true
+			// Is at least a name match, and the CLI version wasn't set or was set
+			// and matches.
+			case pluginName == longName || pluginName == shortName:
+				if !hasDefinitely && (pluginVersion == "" || pluginVersion == versionedPlugin.Version()) {
```
Adirio: As you are `break`ing `if hasDefinetely` after this `switch`, you can avoid checking for a salfe `hasDefinetely` here.

```suggestion
				if pluginVersion == "" || pluginVersion == versionedPlugin.Version() {
```

pkg/cli/cli.go
```diff
@@ -49,7 +55,7 @@ type Option func(*cli) error
 
 // cli defines the command line structure and interfaces that are used to
 // scaffold kubebuilder project files.
-type cli struct {
+type cli struct { //nolint:maligned
```
camilamacedo86: Has no need for ` //nolint:maligned`.
To solve it but the bool definition at the end.

Adirio: I have no strong opinion about this, `maligned` linter is useful for structs which will spawn multiple instances, as those lost 7 bits (not sure if struct fingerprint is rounded to bytes, words, ...) will be added together and end up being a considerable amount of memory.
However structs that will span a single instance do not really have an impact if they are using 7 additional bits, and by being forced to put booleans at the end you may loose readability (you are not allowed to group them as you want).

pkg/cli/cli.go
```diff
@@ -193,9 +255,13 @@ func (c cli) validate() error {
 	}
 
 	// Validate plugin versions and name.
-	for _, versionedPlugins := range c.plugins {
+	for _, versionedPlugins := range c.pluginsFromOptions {
+		pluginNameSet := map[string]struct{}{}
 		for _, versionedPlugin := range versionedPlugins {
 			pluginName := versionedPlugin.Name()
+			if err := plugin.ValidateName(pluginName); err != nil {
+				return fmt.Errorf("failed to validate plugin name %q: %v", pluginName, err)
```
camilamacedo86: This code shows duplicated. 

Could we not:
- Get the plugin that will be built stored and save (new one)
- Then, get all project plugins and check if has the same name and version

pkg/cli/cli.go
```diff
@@ -207,6 +273,26 @@ func (c cli) validate() error {
...
+			// conflict because multiple project versions of a plugin may exist.
+			if _, seen := pluginNameSet[pluginName]; seen {
+				return fmt.Errorf("two plugins have the same name: %q", pluginName)
+			}
+			pluginNameSet[pluginName] = struct{}{}
+		}
+	}
+
+	// Validate plugin keys set in CLI.
+	for pluginName, pluginVersion := range c.cliPluginKeys {
```
camilamacedo86: IHMO the impl could be:
- parseFlags()
- validate()
- save()

I mean the plugin validations should be encapsulated in a Validate method and not be in the parseFlags.

Adirio: This is actually in `func (c cli) validate() error`. GitHub hides some lines and I think you didn't opened them.

pkg/model/config/config.go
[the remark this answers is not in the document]
camilamacedo86: Will the plugins be a breaking change for the layout for the scaffold files?
 I do not think so, in this way IHMO the V3 should not be added. The breaking change will be clear in the new bump tag version as 3.0.0. Note that V2 was created just because the structure of files scaffolded changed.

pkg/cli/cli.go
```diff
@@ -248,20 +335,90 @@ func (c cli) buildRootCmd() *cobra.Command {
...
+		return versionedPlugins, nil
+	}
+
+	// Find all potential mateches for this plugin's name.
+	definitelyMatch, maybeMatch := []plugin.Base{}, map[string][]plugin.Base{}
+	for pluginName, pluginVersion := range cliPluginKeys {
+		cliPluginKey := plugin.Key(pluginName, pluginVersion)
+		// Prevents duplicate versions with the same name from being added to
+		// maybeMatch.
+		hasDefinitely := false
```
camilamacedo86: The API plugin, for example, can be used N times. So, will not it throw an error when we try to use the API more than once? Should we not store in the Project file the plugin and flags used for it scaffold files?
 
So, I do not think that we can check for duplicates. What is the scenario that we are trying to avoid with this validation?

pkg/cli/cli.go
camilamacedo86: I think has no need and reason for to use the `//nolint:maligned.` when it can be easily fixed.

pkg/cli/cli.go
Adirio: I think that the check verifies that no plugin is specified twice, for example using `go` plugin twice.

pkg/plugin/plugin.go
estroz: [`SplitN`](https://golang.org/pkg/strings/#SplitN) will return a slice with at most `n` substrings, so `n = 1` will give you the whole name no matter how many `"."` exist in `name`.

pkg/cli/cli.go
estroz: I think this code can be refactored a bit more too, will update.

pkg/cli/cli.go
estroz: Duplicated? How do you mean?

pkg/cli/cli.go
estroz: @camilamacedo86 the API plugin is a child of `v{1,2}.Plugin{}`, which have the same name but different versions. If they supported an intersecting set of project versions, we would have ambiguous names (`go/v1` and `go/v2` would both be viable for those common project versions).

pkg/plugin/plugin.go
Adirio: You are right, you can mark this as resolved, will update the other.

pkg/cli/cli.go
estroz: I don't mind doing some field reorg, but for manager-type structs like `cli` (low instance numbers at runtime) I'd vote for readability over alignment.

pkg/plugin/v1/init.go
```diff
@@ -29,15 +29,16 @@ import (
 	"github.com/spf13/pflag"
```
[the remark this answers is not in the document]
estroz: We can apply these changes here and in master, then resolve the conflict along with all others when merging the plugins feature branch with master. This should be the last PR before that happens, unless there's more cleanup to do.

estroz:
I'm fine with keeping config version `2`, although the `layout` key represents a breaking change to a config's structure. This should reasonably be handled by `kubebuilder` version though.

@DirectXMan12 @mengqiy @joelanford thoughts on changing config version to `3-alpha` vs going ahead with `2`?

cmd/edit.go
```diff
@@ -74,7 +74,7 @@ func (o *editOptions) LoadConfig() (*config.Config, error) {
 }
 
 func (o *editOptions) Validate(c *config.Config) error {
-	if !c.IsV2() {
+	if c.IsV1() {
```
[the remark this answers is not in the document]
estroz: This change is more for convenience, but I'll explicitly check v2 and v3 if we decide to stick with v3.

pkg/cli/cli.go
```diff
@@ -153,6 +173,17 @@ func (c *cli) initialize() error {
...
+	// Filter plugins by keys passed in CLI, if any.
+	versionedPlugins, err := c.getVersionedPlugins()
+	if err != nil {
+		return err
+	}
+	versionedPlugins, err = filterPluginsByKeys(versionedPlugins, c.cliPluginKeys)
+	if err != nil {
+		return err
+	}
+	c.pluginsFromOptions[c.projectVersion] = versionedPlugins
```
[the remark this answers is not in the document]
estroz: The [`WithPlugins`](https://github.com/estroz/kubebuilder/blob/feature/plugins-config-layout/pkg/cli/cli.go#L131) option needs this to be a map; options areg called before flags are parsed or the config is loaded.

pkg/cli/cli.go
```diff
@@ -248,20 +335,90 @@ func (c cli) buildRootCmd() *cobra.Command {
...
+// returning unambiguously matched plugins.
+// A match occurs if:
+// - Name and version are the same.
+// - Long or short name is the same, and only one version for that name exists.
+// If long or short name is the same, but multiple versions for that name exist,
+// don't guess which version to use and instead return an error.
+func filterPluginsByKeys(versionedPlugins []plugin.Base, cliPluginKeys map[string]string) ([]plugin.Base, error) {
+	if len(cliPluginKeys) == 0 {
+		return versionedPlugins, nil
+	}
```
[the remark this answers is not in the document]
estroz: I think this depends on the semantics of your filter. If a filter is empty then all plugins pass through it correct? Anyway this function is used to determine if the values passed to `--plugins` contain ambiguous names, so if there are no names to filter by then we can rely on the binding functions for `Init`, `CreateAPI`, and `CreateWebhook` to choose the right plugin to run. Perhaps the naming isn't great; `findRelevantPlugins` instead?

pkg/plugin/v1/init.go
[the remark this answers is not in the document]
Adirio: I'm fine with that approach too.

pkg/cli/cli.go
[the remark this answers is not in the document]
Adirio: If I understood this right, `cliPluginVersion` is a whitelist not a blacklist, so I expect an empty whitelist to yield an empty result, while if it was an empty blacklist I would expect a full result. Maybe I'm not interpreting it the same way you do.

internal/config/config.go
```diff
@@ -31,7 +31,7 @@ const (
 	DefaultPath = "PROJECT"
 
 	// Default version if flag not provided
-	DefaultVersion = config.Version2
+	DefaultVersion = config.Version3
```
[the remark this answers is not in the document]
camilamacedo86: Ditto https://github.com/kubernetes-sigs/kubebuilder/pull/1403/files#r387589683
I do not think that we should add V3 as well.

cmd/edit.go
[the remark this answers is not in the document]
camilamacedo86: Ditto https://github.com/kubernetes-sigs/kubebuilder/pull/1403/files#r387589683
I do not think that we should add V3 as well.

pkg/scaffold/webhook.go
```diff
@@ -85,7 +85,7 @@ func (s *webhookScaffolder) Scaffold() error {
 	switch {
 	case s.config.IsV1():
 		return s.scaffoldV1()
-	case s.config.IsV2():
+	case s.config.IsV2() || s.config.IsV3():
```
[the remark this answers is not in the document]
camilamacedo86: Ditto https://github.com/kubernetes-sigs/kubebuilder/pull/1403/files#r387589683
I do not think that we should add V3 as well.

pkg/scaffold/init.go
```diff
@@ -77,7 +77,7 @@ func (s *initScaffolder) Scaffold() error {
 	switch {
 	case s.config.IsV1():
 		return s.scaffoldV1()
-	case s.config.IsV2():
+	case s.config.IsV2() || s.config.IsV3():
```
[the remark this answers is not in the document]
camilamacedo86: Ditto https://github.com/kubernetes-sigs/kubebuilder/pull/1403/files#r387589683
I do not think that we should add V3 as well.

pkg/scaffold/api.go
```diff
@@ -69,7 +69,7 @@ func (s *apiScaffolder) Scaffold() error {
 	switch {
 	case s.config.IsV1():
 		return s.scaffoldV1()
-	case s.config.IsV2():
+	case s.config.IsV2() || s.config.IsV3():
```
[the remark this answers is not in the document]
camilamacedo86: Ditto https://github.com/kubernetes-sigs/kubebuilder/pull/1403/files#r387589683
I do not think that we should add V3 as well.

pkg/plugin/v2/init.go
```diff
@@ -118,16 +116,23 @@ func (p *initPlugin) SetVersion(v string) {
...
 	if err == nil || os.IsExist(err) {
 		return nil, errors.New("config already initialized")
 	}
-
-	return p.config, nil
+	// Only V3 projects get a layout.
+	if p.config.IsV3() {
+		// Init plugins set the config layout.
+		p.config.Layout = plugin.Key(pluginName, pluginVersion)
+	}
```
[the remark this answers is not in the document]
camilamacedo86: Ditto https://github.com/kubernetes-sigs/kubebuilder/pull/1403/files#r387589683
I do not think that we should add V3 as well.

pkg/model/config/config.go
```diff
@@ -51,6 +55,11 @@ func (config Config) IsV2() bool {
 	return config.Version == Version2
 }
 
+// IsV3 returns true if it is a v3 project
+func (config Config) IsV3() bool {
+	return config.Version == Version3
+}
+
```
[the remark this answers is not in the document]
camilamacedo86: Ditto https://github.com/kubernetes-sigs/kubebuilder/pull/1403/files#r387589683
I do not think that we should add V3 as well.

pkg/cli/cli.go
camilamacedo86: The same checks in the inside of both the loops are identical. See that the lines https://github.com/kubernetes-sigs/kubebuilder/blob/2003c53102f972a8b3b65ca3cb38014dd3dc6d36/pkg/cli/cli.go#L261-L269 are the same as https://github.com/kubernetes-sigs/kubebuilder/blob/2003c53102f972a8b3b65ca3cb38014dd3dc6d36/pkg/cli/cli.go#L287-L296 which lead us that could be an implementation centralized and encapsulated. 

Why do we need to re-check the name, version and etc of all plugins that are stored in the Project file?
IHMO, we need to check the current input data only. I mean just the command/plugin which was called and if it is or not supported for the project version.

pkg/cli/cli.go
camilamacedo86: Hi @estroz, 

I understand that:
- A plugin allows us to call/exec the command to scaffold the files
- Many kinds of plugins/command can be called/used more than once 
- If a scaffolded be done by a plugin we will track the plugin/version used for it in the project file. ( It is my understanding from the f2f discussion over it and the main reason is to allow us troubleshooting the scenarios)

Then we will have: 
- A plugin/command X which can be used/called (supportable) if the project version is 1 or 2, for example, but was deprecated and removed for v3.
- So, if the project version is == 1 or 2 this plugin can be used otherwise it will throw the unsupportable error when called.

**Let's think about an example:**

- project init with the go/layout plugin
- and then, with 2 GKV (api) scaffolded
 
Then we will have:
```yaml
domain: testproject.org
layout: go/v2.0.0
multigroup: false
repo: sigs.k8s.io/kubebuilder/testdata/project-v2-multigroup
resources:
- group: crew
  kind: Captain
  version: v1
  plugin: api/v2.0.0
- group: crew
  kind: Frigate
  version: v1beta1
  plugin: api/v2.0.0
version: "2"
```

**NOTE** Just the above example, without the interception scenario, shows not be achievable with the current implementation. 

**Regards the interception** 

Let's migrate the above example project to V3. Then, it means that the ` project.version` will be bumped to 3.
So, after we migrate the project let's scaffold a new API using the plugin. 

Then, the project file will be: 

```yaml
domain: testproject.org
layout: go/v2.0.0 -> Indicates the plugin version/used to init the project
multigroup: false
repo: sigs.k8s.io/kubebuilder/testdata/project-v2-multigroup
resources:
- group: crew
  kind: Captain
  version: v1
  plugin: api/v2.0.0 -> Indicates the plugin version/used to scaffold API
- group: crew
  kind: Frigate
  version: v1beta1
  plugin: api/v2.0.0 -> Indicates the plugin version/used to scaffold API
- group: crew
  kind: NewKind
  version: v2
  plugin: api/v3.0.0 -> Indicates that API was scaffolded with an upper plugin version
version: 3"
```

**Who has the responsibility to know if the command/plugin cannot be executed twice?**

IMO it is part of the command context and not of the plugin, which means that if the init (go/layout)  plugin be called more than once the error will come from the command (It is how it is checked/done already)

**See an example:**

- Project init ( init plugin call for example )
```shell
$ mkdir $GOPATH/src/example
$ cd $GOPATH/src/example
$ kubebuilder init --domain my.domain
Get controller runtime:
$ go get sigs.k8s.io/controller-runtime@v0.4.0
Update go.mod:
$ go mod tidy
Running make:
$ make
/Users/camilamacedo/go/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
Next: define a resource with:
```

Then, let's try to init twice (call the init plugin for the second time)

```shell
$ kubebuilder create api
$ kubebuilder init --domain my.domain
2020/03/05 09:54:36 failed to initialize project: already initialized
```

**Conclision**

- I do not think that duplicates checks for plugins should be implemented 
- The check to allow or not a command/plugin be executed more than once is part of the command context

Is it make sense? 
If not, could you please help me understand what I misunderstand here.

pkg/cli/cli.go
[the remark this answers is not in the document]
Adirio: To understand the filtering semantics I went back to the `package main` entrypoint:

<details><summary>Step by step walk to know which are the input params</summary>

Entrypoint: [`func main() {...}`](https://github.com/estroz/kubebuilder/blob/2003c53102f972a8b3b65ca3cb38014dd3dc6d36/cmd/main.go#L28-L46)
It does two things: `cli.New(...)` and `c.Run()` with their error handlers. [`c.Run()`](https://github.com/estroz/kubebuilder/blob/2003c53102f972a8b3b65ca3cb38014dd3dc6d36/pkg/cli/cli.go#L101-L104) is only calling `Execute` on the `*cobra.Command` so let's focus on [`cli.New(...)`](https://github.com/estroz/kubebuilder/blob/2003c53102f972a8b3b65ca3cb38014dd3dc6d36/pkg/cli/cli.go#L81-L99).
First it sets some values for the struct and allocates some maps. Then it parses the provided options which set additional fields (you can find the options in [`pkg/cli/cli.go:106-145`](https://github.com/estroz/kubebuilder/blob/2003c53102f972a8b3b65ca3cb38014dd3dc6d36/pkg/cli/cli.go#L106-L145)). And when the fields are set, it calls [`c.initialize()`](https://github.com/estroz/kubebuilder/blob/2003c53102f972a8b3b65ca3cb38014dd3dc6d36/pkg/cli/cli.go#L147-L208).
The first thing `initialize` does is to parse some common flags in [`c.parseBaseFlags()`](https://github.com/estroz/kubebuilder/blob/2003c53102f972a8b3b65ca3cb38014dd3dc6d36/pkg/cli/cli.go#L210-L245). This is important as [one of them is the list of plugins](https://github.com/estroz/kubebuilder/blob/2003c53102f972a8b3b65ca3cb38014dd3dc6d36/pkg/cli/cli.go#L223) which are later [parsed and stored](https://github.com/estroz/kubebuilder/blob/2003c53102f972a8b3b65ca3cb38014dd3dc6d36/pkg/cli/cli.go#L231-L242) in `c.cliPluginKeys`. This parsing does a first duplicate check, it doesn't accept multiple plugins with the same name, for example `go/v1,go/v2` or `go/v2,go/v2` will yield errors.
Then `initialize` reads the configuration file and validates the already set fields, but none of these modify them. So [by this point](https://github.com/estroz/kubebuilder/blob/2003c53102f972a8b3b65ca3cb38014dd3dc6d36/pkg/cli/cli.go#L175) we have in `c.cliPluginKeys` a map of plugin names to versions as passed in the command flag without duplicates and with valid names and versions. This is one of the parameters to the function we are studying. The other one is obtained with [`c.getVersionedPlugins()`](https://github.com/estroz/kubebuilder/blob/2003c53102f972a8b3b65ca3cb38014dd3dc6d36/pkg/cli/cli.go#L408-L419) which basically returns the slice of plugins configured for the current scaffold version with [`WithPlugins`](https://github.com/estroz/kubebuilder/blob/2003c53102f972a8b3b65ca3cb38014dd3dc6d36/pkg/cli/cli.go#L123-L136) option to `cli.New`.
So the input to our function are a list of the plugins that kubebuilder has and the list of plugins requested in the command line.
</details>

First of all, in case a configuration file is read successfully, I think we need to add the plugins found under `layout` to the requested ones in the command line, right? It is only reading the project version: https://github.com/estroz/kubebuilder/blob/2003c53102f972a8b3b65ca3cb38014dd3dc6d36/pkg/cli/cli.go#L163-L166.

And after that, if we don't get any plugins requested (`len(cliPluginKeys) == 0`), I think we should use the latest `go` plugin for that scaffold version (we may need a global map to track this).

pkg/cli/cli.go
Adirio: I don't think that resources should track layouts because they are not expected to be mixed. That would be like having some scaffold v1 types and some scaffold v2 in the same project. I don't think that is expected to work.

P.S.: I don't see any relation with `hasDefinetely`.

camilamacedo86: Hi @Adirio, 

I understand that:
- The same project can have scaffolded files made from diff origins. For example, by plugins ( via SDK for example) and indeed via Kubuilder too. 
- If we migrate the project from V2 to V2 and the files do not be re-scaffolded with new plugin versions then it means that the version tracked should be kept for we are able to check it. 
- Then, also as described above, my understanding is that we should track all plugin/versions used to do the scaffolds in the order we are able to troubleshoot the scenarios which is something alleged with kubebuilder admins and discussed in our last f2f. 

PS.: It is not about the `hasDefinetely` as described above it is over the checks and validations made to verify if the plugin is duplicated or not (was used or not before) which IMO should not be done.

Adirio: When I talked about duplications in https://github.com/kubernetes-sigs/kubebuilder/pull/1403#discussion_r387598337 I was refering to plugin duplications, having `go/v2` plugin twice in the list of plugins, not if it was used or not before.

But regarding your points, a plugin is not owned by a plugin. Not even a file is owned by a plugin. For example, if you watch the addon plugin that is implemented right now with the old plugin system, it creates some new files apart from the ones created by kubebuilder when you are scaffolding a resource but it also modifies others. So if `go/v2` created `a.go` but `addon/v1` edited it, who owns the file? Who owns the resource? and if `go/v2` created `api/v1/foo_types.go` and `addon/v1` created `api/v1/foo_additional_types.go`, both belonging to the foo v1 resource, who owns the resource? I think that with this demostration it is clear that we cannot assign ownership of neither a resource nor a file to a single plugin.

So the question that remains is, if we can't track ownership, how do we auto-upgrade from scaffolding versions (for example, scaffolding v2 to v3). The answer is easy. There is no auto-upgrade tool. There has never been one. If you take the v1->v2 example, there is a guide in the book that tells the user how to manually upgrade, but once he has follow the guide, the resulting files are the same as if they were scaffolded with v2. So basically to convert a project from one scaffolding version to another you would need to do it manually.

pkg/cli/cli.go
```diff
@@ -207,6 +273,26 @@ func (c cli) validate() error {
 						pluginName, projectVersion, err)
 				}
 			}
+			// Check for duplicate plugin names. Names outside of a version can
```
Adirio: A couple lines above this, we are validating that the returned supported project versions are valid, but that function is being called after `WithPlugins` has already been evaluated. Should we do the check inside `WithPlugins` instead of here? Additionally, as a plugin can be stored in different scaffolding versions, this check may be evaluated multiple times if we leave it here.

pkg/cli/cli.go
```diff
@@ -248,20 +335,90 @@ func (c cli) buildRootCmd() *cobra.Command {
...
+		hasDefinitely := false
+		for _, versionedPlugin := range versionedPlugins {
+			longName := versionedPlugin.Name()
+			shortName := plugin.GetShortName(longName)
+
+			switch {
+			// Exact match.
+			case pluginName == longName && pluginVersion == versionedPlugin.Version():
+				definitelyMatch = append(definitelyMatch, versionedPlugin)
+				hasDefinitely = true
```
Adirio: I think you should clean the corresponding entry in `maybeMatch` if you find an exact match.

pkg/cli/cli.go
```diff
@@ -248,20 +335,90 @@ func (c cli) buildRootCmd() *cobra.Command {
...
+			// Is at least a name match, and the CLI version wasn't set or was set
+			// and matches.
+			case pluginName == longName || pluginName == shortName:
+				if !hasDefinitely && (pluginVersion == "" || pluginVersion == versionedPlugin.Version()) {
+					maybeMatch[cliPluginKey] = append(maybeMatch[cliPluginKey], versionedPlugin)
```
Adirio: In case of no match being found, neither a full match nor a partial match, there will be no plugin added to the list and no `maybeMatch` to try to educately add one. Shouldn't this be considered an error? For example, imagine I misspell `gu/v2`. I feel like that should definetely return an error.

It should be as easy as checking if `hasDefinetely` is false and there is no entry in `maybeMatch` when the inner loop ends.

pkg/cli/cli.go
camilamacedo86: Tks for the explanation @Adirio. I got it now.  The duplication checks are just about to check if we call in the implementation of the command which will use the plugin it duplicated.

pkg/cli/cli.go
camilamacedo86: It is nit so I do not car too. much. Just not understand why to change the position of the bool defined to the bottom would make it harder to read.

pkg/cli/api.go
```diff
@@ -60,24 +60,15 @@ func (c cli) bindCreateAPI(ctx plugin.Context, cmd *cobra.Command) {
...
-				err := fmt.Errorf("duplicate API creation plugins for project version %q: %s, %s",
-					c.projectVersion, getter.Name(), p.Name())
-				cmdErr(cmd, err)
-				return
-			}
-			hasGetter = true
-			getter = tmpGetter
+		getter, foundGetter = p.(plugin.CreateAPIPluginGetter)
+		if foundGetter {
+			break
```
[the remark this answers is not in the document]
estroz: There should still be a duplicate check. I'm reworking this logic to include information from the config file or `--plugins` values. Applies to `api`, `webhook`, `init`.

pkg/cli/cli.go
estroz: These are checking two different `cli` fields: `cliPluginKeys` and `pluginsFromOptions`. They have different types (`string` vs `plugin.Base`); combining keys from `pluginsFromOptions` would still require a loop, and we'd lose error information. I'd like to keep field validation separate as much as possible, if not for readability then for proper error reporting.

pkg/cli/cli.go
estroz: Either place works for me, however putting as much validation as possible in `cli.validate()` improves readability by semantic encapsulation for a small performance loss from re-evaluating plugin validity.

pkg/cli/cli.go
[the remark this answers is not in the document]
estroz: The goal of `filterPluginsByKeys` is to return a basic set of plugins if `--plugins` is not set, which it currently accomplishes but fails to consider a few cases as you mention. It also breaks semantics of a whitelist, which is what `cliPluginKeys` should be.

A new option `WithDefaultPlugins` that sets a field `defaultPluginsFromOptions` containing default plugins to run if `--plugins` is unset should be added since returning a default list of plugins if the whitelist is empty makes sense.

Unknown plugins cause an error to be returned:
```console
$ kubebuilder init --plugins foo
2020/03/05 10:44:00 no plugins found for CLI plugin keys ["foo"], likely a naming discrepancy
```

pkg/cli/cli.go
estroz: An error is returned if a plugin key doesn't match anything.

internal/config/config.go
```diff
@@ -98,7 +98,7 @@ type Config struct {
 func New(path string) *Config {
 	return &Config{
 		Config: config.Config{
-			Version: config.Version2,
+			Version: DefaultVersion,
```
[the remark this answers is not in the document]
estroz: PR: #1440

pkg/plugin/v1/init.go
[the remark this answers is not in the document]
estroz: PR: #1439

testdata/project-v2/PROJECT
```diff
@@ -1,4 +1,5 @@
 domain: testproject.org
+layout: go.kubebuilder.io/v2.0.0
 repo: sigs.k8s.io/kubebuilder/testdata/project-v2
```
camilamacedo86: I think would great rebase the branch with the master. 
Note that it has a mock now for the add-on as well.

cmd/main.go
```diff
@@ -31,6 +31,10 @@ func main() {
 			&pluginv1.Plugin{},
 			&pluginv2.Plugin{},
 		),
+		cli.WithDefaultPlugins(
+			&pluginv1.Plugin{},
+			&pluginv2.Plugin{},
+		),
```
camilamacedo86: Why we need to: 

```
		cli.WithPlugins(
			&pluginv1.Plugin{},
			&pluginv2.Plugin{},
		),
		cli.WithDefaultPlugins(
			&pluginv1.Plugin{},
			&pluginv2.Plugin{},
		),
```

What it the diff between plugin and default plugin? Is not it redundant?

estroz: Imagine the case where there are multiple plugins per project version available to a user, ex. two different types of Go scaffolds that support project version 2. One of those two should be defaulted to in case the user doesn’t set `—plugins` on `init` or they don’t have a `layout` key in their config (project version 1). However the `CLI` won’t know which it should default to unless one is explicitly set. See https://github.com/kubernetes-sigs/kubebuilder/pull/1403#discussion_r388496659 for more details on why I chose to set defaults explicitly.

camilamacedo86:
Hi @estroz, 

It is not passing now after the rebase in the testdata check. You need run `make generate` locally to update the testdata dir with your changes.

camilamacedo86:
Hi @estroz,

The GH did not get the Travis result. If you close and re-open this one the tests will be re-trigged.

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
```diff
@@ -62,7 +62,7 @@ project versions (via `SupportedProjectVersions()`).
 Example `PROJECT` file:
 
 ```yaml
-version: "3-alpha"
+version: "2"
```
DirectXMan12: ??

pkg/cli/cli.go
```diff
@@ -134,12 +165,20 @@ func WithExtraCommands(cmds ...*cobra.Command) Option {
 
 // initialize initializes the cli.
 func (c *cli) initialize() error {
+	// Initialize cli with globally-relevant flags or flags that determine
+	// certain plugin type's configuration.
+	if err := c.parseBaseFlags(); err != nil {
```
DirectXMan12: why do we need this globally?  isn't this just needed on the `init` invocation?

pkg/cli/cli.go
```diff
@@ -153,6 +192,40 @@ func (c *cli) initialize() error {
...
+	// In case 1, either default plugins or the config layout can be used to
+	// determine which plugin to use. Otherwise, the value passed to --plugins
+	// is used. Since both layout and a --plugins value are user input, they
+	// may need to be resolved to known plugin keys.
+	var filterKeys map[string]string
+	switch {
+	case len(c.cliPluginKeys) != 0:
+		// Filter plugins by keys passed in CLI.
+		filterKeys = c.cliPluginKeys
+	case c.configured && !projectConfig.IsV1():
```
DirectXMan12: doesn't this break current `v2` projects?

pkg/cli/cli.go
```diff
@@ -153,6 +192,40 @@ func (c *cli) initialize() error {
...
+		// Use the default plugins for this project version.
+		c.resolvedPlugins = c.defaultPluginsFromOptions[c.projectVersion]
+	}
+	if len(filterKeys) != 0 {
+		plugins := c.pluginsFromOptions[c.projectVersion]
+		c.resolvedPlugins, err = resolvePluginsByKeys(plugins, filterKeys)
+		if err != nil {
+			return err
+		}
+	}
```
DirectXMan12: what happens if filterKeys is not set here? (i.e. if there's no plugins set for options)?

DirectXMan12: (if we check elsewhere, add a comment here)

DirectXMan12: (it's also maybe a good indication that we're failing at the principle of "parse instead of validate")

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
camilamacedo86: @DirectXMan12 
I understand that the version here will still be 2 since we have not broken changes in the layout. However, kb tag will need to be bumped to 3.0.0. Am I right?

estroz: There was pushback against upgrading to "3-alpha". Regardless of whether we bump I figured it was better to make the change in a follow-up.

pkg/cli/cli.go
estroz: It will, and probably a good reason to upgrade to project version "3-alpha".

pkg/cli/cli.go
estroz: The CLI needs to pre-parse the `--plugins` global flag for each command, so I figured wrapping all flag pre-parsing in one method made sense. The `cli.doGenericHelp` flag is set by `cli.parseBaseFlags` so `init` can still print generic help if `--project-version` is not set.

pkg/cli/cli.go
estroz: `cli.resolvedPlugins` defaults to the [default plugins](https://github.com/kubernetes-sigs/kubebuilder/pull/1403/files#diff-c11db1404a63b574b13465c6a00e4ee1R219) for a project version if `filterKeys` is empty. Default plugins are ensured to exist by validating them [here](https://github.com/kubernetes-sigs/kubebuilder/pull/1403/files#diff-c11db1404a63b574b13465c6a00e4ee1R305). Does this still fail parse instead of validate")? I'll add a comment describing this scenario if not.

Looks like the case when `c.configured && projectConfig.IsV1() == true` is not validated however. Making a follow-up now.

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
DirectXMan12: that's ok, but I think if we're gonna make a field mandatory that's a change in version.

pkg/cli/cli.go
DirectXMan12: > The CLI needs to pre-parse the --plugins global flag for each command,

wait, does it?  I thought we figured out that it didn't...

estroz: Summary of discussion: having a global `--plugins` flag is not in scope for phase 1 since the type of plugin used in `init` solely determines what plugin is used for subsequent commands (the `layout` config key). A global `--plugins` flag will be discussed in phase 2: #1378.

PR #1413, opened 2020-03-05
Fix broken plugin addon template

rajansandeep:
The plugin addon template is out-of-date and breaks functionality when creating api using the `--pattern=addon` flag.
This PR fixes the plugin addon template.

### DIFF
--- a/plugins/addon/controller.go
+++ b/plugins/addon/controller.go
@@ -48,7 +48,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	api "{{ .Resource.GoPackage }}/{{ .Resource.Version }}"
+	api "{{ .Resource.Package }}"
 )
 
 var _ reconcile.Reconciler = &{{ .Resource.Kind }}Reconciler{}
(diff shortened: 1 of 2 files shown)


Adirio:
> Could we add a mock testdata to check the plugin in order to enure this change?

I definetely agree with your suggestion, some thoughts an open questions:

1. Probably tests should be implemented in another PR. Do you want to raise an issue to track addon plugin tests @camilamacedo86 or should I do it myself?
1. I would use "project-v2-addon" and save "plugins" for the new plugin system.
1. Should we wait until tests are implemented to merge this or should we merge this and then implement the tests? IMHO we could merge this first as it is behind a non-stable environmental flag and will not affect the bast mayority of the users.
1. The addon plugin will probably have to be migrated to the new plugin system. Should we hold the test implementation until then? IMHO we should implement the tests meanwhile and then use them to verify that the addon migration to the new plugin system was done properly.

camilamacedo86:
Hi @Adirio, 

Thank you for check that as well.  Following my comments inline.

> Probably tests should be implemented in another PR. Do you want to raise an issue to track addon plugin tests @camilamacedo86 or should I do it myself?

IMO when we do a bug fix usually means that the project is missing a test and because of this the bug was introduced. So,  as part of the fix and to allow us to verify, check the changes/pr and indeed ensure that the same will not be re-introduced the tests should be done with. I mean in the same PR and then I do not think that we should have a specific issue to add the test missing which allow the bug fixed. 

However, in this case I think we can move forward with your suggestion. The pr for the mock data is: https://github.com/kubernetes-sigs/kubebuilder/pull/1420

> I would use "project-v2-addon" and save "plugins" for the new plugin system.

Agreed 👍 

> Should we wait until tests are implemented to merge this or should we merge this and then implement the tests? IMHO we could merge this first as it is behind a non-stable environmental flag and will not affect the bast mayority of the users.

I think we can move forward with and after and then, merge: https://github.com/kubernetes-sigs/kubebuilder/pull/1420

> The addon plugin will probably have to be migrated to the new plugin system. Should we hold the test implementation until then? IMHO we should implement the tests meanwhile and then use them to verify that the addon migration to the new plugin system was done properly.

I agree with 👍  you and I do not think that we should hold since the addon is already implemented and this is not the first issue/bug fix done on it in the latest days. See: https://github.com/kubernetes-sigs/kubebuilder/pull/1395/files

Adirio:
Perfect, I will review #1420, this PR just needs a `/approve` label

PR #1428, opened 2020-03-12
[WIP] Dependencies checks

### LINKED ISSUE #1371: Missing Kustomize Dependency for `make install`
Fresh linux system, trying to go through the getting started
```
[nova@nova kapad]$ make install
/home/nova/go/bin/controller-gen "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
kustomize build config/crd | kubectl apply -f -
/bin/sh: kustomize: command not found
error: no objects passed to apply
make: *** [Makefile:30: install] Error 1
```

Either docs that mention this, a check in `make install`, a better error message with a path forward, or including `kustomize` in the `build/bin/` directory of a release

/kind bug

Adirio:
# Description

Cleans dependency version checks and adds one for `kustomize`.

# Motivation

Fixes #1371


Adirio:
Errors in Travis and `pull-kubebuilder-test` are expected as those tests are being run in images that do not have the `kustomize` dependency. Still have to debug the rest.

`pull-kubebuilder-test` image configuration: [link](https://github.com/kubernetes/test-infra/blob/efcfec65aeb8ff9592104f07ef53c3dfaffacfd4/config/jobs/kubernetes-sigs/kubebuilder/kubebuilder-presubmits.yaml#L3-L13)

cmd/internal/dependencies/semantic_version.go
```diff
@@ -0,0 +1,113 @@
...
+*/
+
+package dependencies
+
+import (
+	"fmt"
+	"strconv"
+)
+
+const semanticVersionPattern = `v?(?P<major>\d+)(?:\.(?P<minor>\d+)(?:\.(?P<patch>\d+))?)?(?P<extra>[\w\.-]+)?`
```
estroz: Instead of re-implementing semver parsing, can a library like [`blang/semver`](https://github.com/blang/semver) be used? Plugin version enforcement is already using this library.

Adirio: Does it support `1.14rc`? I think that may not technichally be a semantic version.

estroz: No, some pre-processing would need doing for atypical semver strings. Most of the semver processing logic implemented in this PR exists in the linked library. Check it out, see if it can be dropped in easily.

camilamacedo86:
HI @Adirio, 

The issue was closed. Has any reason for we keep this PR or can we close it as well? WDYT?

camilamacedo86:
Hi @Adirio, 

Thee issue was solved with the makefile solution in https://github.com/kubernetes-sigs/kubebuilder/pull/1430 so I understand that has no reason for we keep this PR.  In this way, I am casing this one. However, please feel free to re-open if you see anything that should be addressed with. Really tks for your collab. 🥇 

/close

Issue #1429, opened 2020-03-12
scaffolded controller-gen dependency is not checksummed


toonsevrin:
I'm not sure what the security requirements for this project are but the scaffolded controller-gen dependency is not checksummed. This makes the project vulnerable to for example sigs.k8s.io being compromised or inconsistencies when someone forces a new commit on the `{{ .ControllerToolsVersion }}` tag.

[pkg/scaffold/internal/templates/v2/makefile.go#L128](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/scaffold/internal/templates/v2/makefile.go#L128)
```
go get sigs.k8s.io/controller-tools/cmd/controller-gen@{{ .ControllerToolsVersion }} ;
```

A go.sum should be added to the Makefile template (near the line I quoted):
```bash
echo sigs.k8s.io/controller-tools/cmd/controller-gen {{ .ControllerToolsVersion }}  {{ .ControllerToolsChecksum }}  > go.sum
```

**I'm not sure if a go.sum works without a go.mod file though, please confirm!**

camilamacedo86:
HI @toonsevrin, 

Go 1.13+ setup by default its proxies and the modules are download from there usually. IHMO add a checksum as you suggested is a little overwhelming. However, users are able to customize their Makefile as they wish and then. So, if they think that would be beneficial and required for the work and develop their project check the checksum as you suggested they are able to do it. WDYT?  Would be not it enough?

toonsevrin:
Yep :) and checksummed against sum.golang.org or something (so an attacker would have to take over sum.golang.org and proxy.golang.org).

And I guess so, I believe checksumming dependencies should be the standard, especially in open-source projects. But as this is an ideological belief (see the note above about this being quite secure), I won't argue you further :)

camilamacedo86:
IHMO is better we try to keep the things as simple as possible but we need to allow users to customize it as they wish. Also, if someone has too much concern could indeed not allow doing the download and instead, use what is available in the machine already. 

We have also the RFE:https://github.com/kubernetes-sigs/kubebuilder/issues/1351

So, I think we could close this one. WDYT? Feel free to keep open if besides the above argumentation you think that it is something that should be addressed anyway.

PR #1430, opened 2020-03-12
Install kustomize if not present

### LINKED ISSUE #1426: Remove external kustomize dependency
As kustomize is integrated in kubectl nowadays (without an external dependency), I recommend kubebuilder does the same.

I propose that kustomize becomes an internal go dependency and will provide some documentation here for one of the developers to implement this.

[Here](https://github.com/kubernetes-sigs/kustomize/blob/master/functions/examples/template-go-nginx/) is a simple example of using kustomize as a go dependency. [Here](https://github.com/kubernetes-sigs/kustomize/blob/master/functions/examples/template-go-nginx/image/main.go#L26) is the specific line where a resource is generated with the kustomize go library.

As external behavior shouldn't change for this, it can be implemented in v2. Good luck!

toonsevrin:
Resolves #1426

As long as the kustomize binary is already present in the PATH, the makefile behavior will be identical to before this patch (See the [which statement](https://github.com/kubernetes-sigs/kubebuilder/compare/master...toonsevrin:master#diff-c29bfcc8d5d17695c61078bd7fd55ec4R139)). In my opinion this is perfect backwards compatibility and this patch should be included in v2.

Note that the test `continuous-integration/travis-ci/pr` will fail due to the Makefile being changed.

Notes:
* I have not yet tested this (or written automated tests for this)
* Neither controller-gen nor kustomize are currently checksummed (this is a small vulnerability, but was already present): See #1429. This should be patched in a separate PR.

### DIFF
--- a/pkg/scaffold/internal/templates/v2/makefile.go
+++ b/pkg/scaffold/internal/templates/v2/makefile.go
@@ -32,6 +32,8 @@ type Makefile struct {
 	BoilerplatePath string
 	// Controller tools version to use in the project
 	ControllerToolsVersion string
+	//Kustomize version to use in the project
+	KustomizeVersion string
 }
 
 // SetTemplateDefaults implements input.Template
@@ -80,17 +82,17 @@ run: generate fmt vet manifests
 	go run ./main.go
 
 # Install CRDs into a cluster
-install: manifests
-	kustomize build config/crd | kubectl apply -f -
+install: manifests kustomize
+	$(KUSTOMIZE) build config/crd | kubectl apply -f -
 
 # Uninstall CRDs from a cluster
-uninstall: manifests
-	kustomize build config/crd | kubectl delete -f -
+uninstall: manifests kustomize
+	$(KUSTOMIZE) build config/crd | kubectl delete -f -
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-deploy: manifests
+deploy: manifests kustomize
 	cd config/manager && kustomize edit set image controller=${IMG}
-	kustomize build config/default | kubectl apply -f -
+	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
 # Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
@@ -131,5 +133,20 @@ ifeq (, $(shell which controller-gen))
 CONTROLLER_GEN=$(GOBIN)/controller-gen
 else
 CONTROLLER_GEN=$(shell which controller-gen)
+endif
+
+kustomize:
+ifeq (, $(shell which kustomize))
+	@{ \
+	set -e ;\
+	KUSTOMIZE_GEN_TMP_DIR=$$(mktemp -d) ;\
+	cd $$KUSTOMIZE_GEN_TMP_DIR ;\
+	go mod init tmp ;\
+	go get sigs.k8s.io/kustomize/kustomize/v3@{{ .KustomizeVersion }} ;\
+	rm -rf $$KUSTOMIZE_GEN_TMP_DIR ;\
+	}
+KUSTOMIZE=$(GOBIN)/kustomize
+else
+KUSTOMIZE=$(shell which kustomize)
 endif
 `
--- a/pkg/scaffold/init.go
+++ b/pkg/scaffold/init.go
@@ -40,6 +40,8 @@ const (
 	ControllerRuntimeVersion = "v0.5.0"
 	// ControllerToolsVersion is the kubernetes-sigs/controller-tools version to be used in the project
 	ControllerToolsVersion = "v0.2.5"
+	// KustomizeVersion is the kubernetes-sigs/kustomize version to be used in the project
+	KustomizeVersion = "v3.5.4"
 
 	imageName = "controller:latest"
 )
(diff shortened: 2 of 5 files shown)


toonsevrin:
I am now testing this in a virtual machine. I will update this comment with the results.

/hold

toonsevrin:
:heavy_check_mark: I've tested this on fresh ubuntu vms. It seems to work as expected.

I tested creating and installing an API with and without kustomize in the PATH. Without kustomize in the path, "go get kustomize" is ran, with kustomize nothing is ran.

Please confirm (and perhaps test some more paths) and merge.

Adirio:
You need to run the `golden_script.sh` and that will uodate the `Makefile`s under the `testdata` directory.

toonsevrin:
All right I'll run goldenscript and squash.

toonsevrin:
@Adirio can you or someone else generate the golden_script for me? I receive the following error during the test:
```bash
go test ./pkg/... ./cmd/... -coverprofile cover.out
...
2020/03/12 17:33:22 failed to start the controlplane. retried 5 times: fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory
...
```
Not sure where it gets the idea that I have a /usr/local/kubebuilder folder :P

Adirio:
Sorry it is `generate_golden.sh`.

I can't run it because I'm on Windows.

toonsevrin:
@Adirio no worries, I've started a new vm with 4 times the RAM. Should complete this time.

As I'm on 1Mbps expect this to take a while :laughing:

toonsevrin:
@Adirio looks like the test froze, are you able to restart it?
/test all

Adirio:
> @Adirio looks like an unexpected exception with the test occured, are you able to restart it?

Close the PR and reopen it.

Adirio:
/ok-to-test
This will trigger the rest of the tests, but you still have to reopen for Travis

camilamacedo86:
Hi @toonsevrin, 

Instead of run the golden script manually I'd recommend use the `make generate`.  Note that it will ensure that the script will be executed with the correct versions required and is easier as well. 

c/c @Adirio

camilamacedo86:
Hi @toonsevrin, 

Great contribution.  Your solution for me make totally sense and you are following the current design/approach that is already applied for another dependence. 

/lgtm

HI @Adirio,   

It does not change the design and etc. Do you see any objection/reason for we not move forward here?

camilamacedo86:
/lgtm cancel 

Reason: Missing squash the commits.

Adirio:
#1371 suggested to check for the dependency instead of installing it. One of the approaches should be followed by I'm not sure which of them feels better. Will keep the discussion over there.

toonsevrin:
@camilamacedo86 squashed :tada: 

@Adirio Imo the kustomize dependency should automatically be installed just like any other internal dependency (kustomize is no longer used by developers as they can `apply -k`). In an ideal world, it is installed in a path that is not accessible externally to the makefile (that way the external environment is not influenced). But as that is a bit perfectionistic (and is not done with controller-gen), I think my PR is superior to a simple check (and once we integrate checksums it will be safer as well, as users can trust us to peer-review the dependency)

Adirio:
Let's see what do the owners think and we can remove the `hold` label.

pkg/scaffold/internal/templates/v2/makefile.go
```diff
@@ -132,4 +134,19 @@ CONTROLLER_GEN=$(GOBIN)/controller-gen
...
+ifeq (, $(shell which kustomize))
+	@{ \
+	set -e ;\
+	KUSTOMIZE_GEN_TMP_DIR=$$(mktemp -d) ;\
+	cd $$KUSTOMIZE_GEN_TMP_DIR ;\
+	go mod init tmp ;\
+	go get sigs.k8s.io/kustomize/kustomize/v3@{{ .KustomizeVersion }} ;\
+	rm -rf $$KUSTOMIZE_GEN_TMP_DIR ;\
+	}
+KUSTOMIZE=$(GOBIN)/kustomize
```
mengqiy: What if the `GOBIN` is not set?
`go get sigs.k8s.io/kustomize/kustomize/v3@v3.5.4` works for me with go 1.14 with `GOPATH` and `GO111MODULE=on` set, but `GOBIN` unset.

toonsevrin:
https://github.com/toonsevrin/kubebuilder/blob/7c6dc784a0c212a4f45e6a7f630ffced3c51ab57/pkg/scaffold/internal/templates/v2/makefile.go#L65

pkg/scaffold/internal/templates/v2/makefile.go
camilamacedo86: @mengqiy it is exactly the same impl for cotroller-gen. So, because of this I think would make sense as well. Also, see: https://github.com/kubernetes-sigs/kubebuilder/blob/7c6dc784a0c212a4f45e6a7f630ffced3c51ab57/pkg/scaffold/internal/templates/v2/makefile.go#L63-L68

mengqiy:
/lgtm
/approve
/hold cancel
IMO this improvement is fine, since 1) it doesn't break any backward compatibility. 2) it respects it if users have already installed the kustomize binary. 3) it doesn't mess up go.mod, since it's done in a tmp dir 4) It deterministic what version of kustomize is installed if the installation is triggered. 

Sorry for the delay.

PR #1431, opened 2020-03-13
:book: note multiversion kubectl behavior

DirectXMan12:
This describes how kubectl handles preferred versions, with a link to
the official docs.  This should make it clearer how folks can expect
users to interact with multiversion APIs.  It also adds a reminder to
always use qualified GVRs in scripts.

### DIFF
--- a/docs/book/src/multiversion-tutorial/deployment.md
+++ b/docs/book/src/multiversion-tutorial/deployment.md
@@ -84,8 +84,40 @@ respectively.  Notice that each has a different API version.
 Finally, if we wait a bit, we should notice that our CronJob continues to
 reconcile, even though our controller is written against our v1 API version.
 
+<aside class="note">
+
+<h1>kubectl and Preferred Versions</h1>
+
+When we access our API types from Go code, we ask for a specific version
+by using that version's Go type (e.g. `batchv2.CronJob`).
+
+You might've noticed that the above invocations of kubectl looked
+a little different from what we usually do -- namely, they specify
+a *group-version-resource*, instead of just a resource.
+
+When we write `kubectl get cronjob`, kubectl needs to figure out which
+group-version-resource that maps to.  To do this, it uses the *discovery
+API* to figure out the preferred version of the `cronjob` resource.  For
+CRDs, this is more-or-less the latest stable version (see the [CRD
+docs][CRD-version-pref] for specific details).
+
+With our updates to CronJob, this means that `kubectl get cronjob` fetches
+the `batch/v2` group-version.
+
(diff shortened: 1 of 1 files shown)


docs/book/src/multiversion-tutorial/deployment.md
```diff
@@ -84,8 +84,40 @@ respectively.  Notice that each has a different API version.
...
 reconcile, even though our controller is written against our v1 API version.
 
+<aside class="note">
+
+<h1>kubectl and Preferred Versions</h1>
+
+When we access our API types from Go code, we ask for a specific version
+by using that version's Go type (e.g. `batchv2.CronJob`).
+
+You might've noticed that the above invocations of kubectl looked
```
djzager: I wonder if this would catch non-native English readers off guard. Just not the typical contraction I am used to seeing.

djzager: Not sure if I was trying to make this sound less nitpicky than it is or not... :man_shrugging:

djzager: Not sure if I was trying to make this sound less nitpicky than it is or not... :man_shrugging:

camilamacedo86: I am non-native English and IMO it is very OK. 

Just to share my IHMO: What is hard to understand for me (non-native English) is when the text does not pass the info in a straightway or more informal way to write. Usually when native speakers try to make easier is harder. :-P since when you folks are kids you learn more informal English and use too much "phrasal verbs" and when we are learning English as an outsider we see more the formal English and the phrasal verbs are crazy because they are 2/3 words/verbs that together have a completed different mean.

docs/book/src/multiversion-tutorial/deployment.md
```diff
@@ -84,8 +84,40 @@ respectively.  Notice that each has a different API version.
...
+a little different from what we usually do -- namely, they specify
+a *group-version-resource*, instead of just a resource.
+
+When we write `kubectl get cronjob`, kubectl needs to figure out which
+group-version-resource that maps to.  To do this, it uses the *discovery
+API* to figure out the preferred version of the `cronjob` resource.  For
```
camilamacedo86: WDYT about add the link/ref for *discovery
API* here?

camilamacedo86: The link is just a nit and not a blocker for we move forward so: 

/lgtm

Adirio:
@DirectXMan12 As this is a documentation change that should be applied rightaway, I think this PR needs to be cherry-picked at `book-v2` branch.

PR #1433, opened 2020-03-14
Ignore

camilamacedo86:
Ignore

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller.go
@@ -124,6 +124,19 @@ func (r *CronJobReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	}
 
 	/*
+
+		<aside class="note"> 
+
+		<h1>What is this index about?</h1>
+
+		<p>The reconciler fetches all jobs owned by the cronjob for the status. As our number of cronjobs increases, 
+		looking these up can become quite slow as we have to filter through all of them. For a more efficient lookup, 
+		these jobs will be indexed locally on the controller's name. A jobOwnerKey field is added to the 
+		cached job objects. This key references the owning controller and functions as the index. Later in this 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -32,11 +32,17 @@ jobs:
...
+    - stage: generate
       # The golden_test.sh check if the the testdata is updated according to the current changes
       # To update the testdata use the Makefile targets `make generate-setup` then `make generate-testdata`
       script: ./golden_test.sh
 
+    - stage: OSX Supportability
+      os: osx
+      # it will runs the golden.sh script which allows we check its supportability for the SO
+      script: 
+      - make generate
```
Adirio:
```suggestion
      script: make generate
```

.travis.yml
```diff
@@ -32,11 +32,17 @@ jobs:
       before_script: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.23.3
       script: ./scripts/verify.sh
 
-    - stage: golden
+    - stage: generate
       # The golden_test.sh check if the the testdata is updated according to the current changes
       # To update the testdata use the Makefile targets `make generate-setup` then `make generate-testdata`
       script: ./golden_test.sh
```
Adirio:
```suggestion
      script: make generate
```

camilamacedo86:
HI @Adirio, 

> From a conceptual viewpoint, I'm not sure if we really need to run this test in every OS. This test is chekcing that testdata was updated properly, doing that once should be enough. Right?

See that what we are trying to address here is:

> How can we ensure that users will be able to use the Kubebuilder in Linux and Osx?

Then, not that. the golden.sh will run the mainly used commands with a good combination to scaffold the projects. So, if we exec all these commands successfully in an osx and Linux we can ensure that the project still working for both.

.travis.yml
camilamacedo86: The `make generate` just update and not check if it was updated or not which is the goal of this job.

PR #1439, opened 2020-03-20
cmd/init.go: initialize `pkg/model/config.Config` when binding flags

estroz:
Separate r/w config from config initialized by flags in `init`.

/cc @DirectXMan12 @Adirio


estroz:
Doing a larger rewrite of how scaffolders handle config objects.

PR #1440, opened 2020-03-20
internal/config/config.go: use `DefaultVersion` instead of `config.Version2`

estroz:
So `Config.Version` does not also need to be updated when `DefaultVersion` is changed in the future.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -98,7 +98,7 @@ type Config struct {
 func New(path string) *Config {
 	return &Config{
 		Config: config.Config{
-			Version: config.Version2,
+			Version: DefaultVersion,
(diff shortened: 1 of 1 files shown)


camilamacedo86:
HI @estroz 
It is just a replacement for the constant. (too small) Should not break the CI. 
I will be looking at it.

camilamacedo86:
Confirmed that it is a Trevi's incident. See: https://travis-ci.org/github/kubernetes-sigs/kubebuilder/builds/665054120?utm_source=github_status&utm_medium=notification

Also, check it locally and all worked fine as expected since it is just a small change to use the var instead of. 

/lgtm
/approve

PR #1442, opened 2020-03-21
cli: delegate management of a config to a `CLI`

estroz:
A config object is initialized/loaded/saved by the CLI. Scaffolds or plugins shouldn't have to load/save a config object, nor should anything exported have internal parameters that cannot be populated if called outside of kubebuilder.

/cc @DirectXMan12 @mengqiy @Adirio @camilamacedo86

### DIFF
--- a/cmd/edit.go
+++ b/cmd/edit.go
@@ -48,6 +48,10 @@ func newEditCmd() *cobra.Command {
 	# Disable the multigroup layout
 	kubebuilder edit --multigroup=false`,
 		Run: func(_ *cobra.Command, _ []string) {
+			var err error
+			if options.config, err = config.LoadInitialized(); err != nil {
+				log.Fatal(err)
+			}
 			if err := cmdutil.Run(options); err != nil {
 				log.Fatal(editError{err})
 			}
@@ -62,6 +66,8 @@ func newEditCmd() *cobra.Command {
 var _ cmdutil.RunOptions = &editOptions{}
 
 type editOptions struct {
+	config *config.Config
+
 	multigroup bool
 }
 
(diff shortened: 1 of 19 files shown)


estroz:
If necessary we can port this to master too.

pkg/plugin/v2/api.go
```diff
@@ -196,10 +195,12 @@ func (p *createAPIPlugin) GetScaffolder(c *config.Config) (scaffold.Scaffolder,
 		return nil, fmt.Errorf("unknown pattern %q", p.pattern)
 	}
 
-	return scaffold.NewAPIScaffolder(c, string(bp), res, p.doResource, p.doController, plugins), nil
+	// Create the actual resource from the resource options
+	res := p.resource.NewResource(p.config, p.doResource)
```
camilamacedo86: could it not be inside of NewAPIScaffolder as well?

pkg/plugin/v2/webhook.go
```diff
@@ -25,13 +25,15 @@ import (
 	"github.com/spf13/pflag"
 
 	"sigs.k8s.io/kubebuilder/internal/cmdutil"
-	"sigs.k8s.io/kubebuilder/internal/config"
+	"sigs.k8s.io/kubebuilder/pkg/model/config"
```
camilamacedo86: 👍 great

pkg/plugin/v2/webhook.go
```diff
@@ -93,18 +95,18 @@ func (p *createWebhookPlugin) Validate(_ *config.Config) error {
...
 	bp, err := ioutil.ReadFile(filepath.Join("hack", "boilerplate.go.txt")) // nolint:gosec
 	if err != nil {
 		return nil, fmt.Errorf("unable to load boilerplate: %v", err)
 	}
 
 	// Create the actual resource from the resource options
-	res := p.resource.NewResource(&c.Config, false)
-	return scaffold.NewV2WebhookScaffolder(&c.Config, string(bp), res, p.defaulting, p.validation, p.conversion), nil
+	res := p.resource.NewResource(p.config, false)
+	return scaffold.NewV2WebhookScaffolder(p.config, string(bp), res, p.defaulting, p.validation, p.conversion), nil
```
camilamacedo86: Same here. The `NewV2WebhookScaffolder` has the required data, so could not the `p.resource.NewResource(p.config, false)` be inside of?

pkg/plugin/v2/api.go
estroz: There's a distinction controlled by plugin version between resources created by `NewV1Resource` and `NewResource`, which is why `*Resource` is passed instead of `*Options`. Ditto for webhook plugins.

estroz: To be fair, scaffolds know project version internally and make decisions based on that information so we could move this into the scaffolder. However I think we should wait until we remove project version 1 and re-assess where we create resources then.

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
```diff
@@ -102,9 +102,13 @@ type GenericSubcommand interface {
...
     // BindFlags binds the plugin's flags to the CLI. This allows each plugin to define its own
     // command line flags for the kubebuilder subcommand.
-    BindFlags(fs *pflag.FlagSet)
+    BindFlags(*pflag.FlagSet)
     // Run runs the subcommand.
     Run() error
+    // InjectConfig passes a config to a plugin. The plugin may modify the
+    // config. Initializing, loading, and saving the config is managed by the
+    // cli package.
+    InjectConfig(*config.Config)
```
camilamacedo86: Could not it be BindConfig? 
Could we find a better name then Inject?

cmd/edit.go
```diff
@@ -48,6 +48,10 @@ func newEditCmd() *cobra.Command {
 	# Disable the multigroup layout
 	kubebuilder edit --multigroup=false`,
 		Run: func(_ *cobra.Command, _ []string) {
+			var err error
```
camilamacedo86: Could you please rebase the PR with the master to be easier to verify this one?

camilamacedo86:
Hi @estroz 

/lgtm

Since it is for `kubernetes-sigs:feature/plugins-part-2-electric-boogaloo`. However, could we address the nits raised here in a follow-up PR?

PR #1443, opened 2020-03-22
config: allow addition of arbitrary extra fields to `Config`

estroz:
Other plugins will need to configure their scaffolded projects, requiring the config format to support arbitrary fields in addition to well-defined keys kubebuilder uses. These keys will be housed under the `plugins` key/`Config.ExtraFields`; `EncodeExtraFields()` and `DecodeExtraFields()` handle adding and populating arbitrary objects to/from `ExtraFields`. This feature does not interfere with existing configs or how kubebuilder interacts with them.

Extra fields are only relevant to plugins so this should not go into master first.

/cc @DirectXMan12 @mengqiy @Adirio @camilamacedo86

### DIFF
--- a/pkg/model/config/config.go
+++ b/pkg/model/config/config.go
@@ -17,7 +17,10 @@ limitations under the License.
 package config
 
 import (
+	"fmt"
 	"strings"
+
+	"sigs.k8s.io/yaml"
 )
 
 // Scaffolding versions
@@ -46,22 +49,26 @@ type Config struct {
 
 	// Layout contains a key specifying which plugin created a project.
 	Layout string `json:"layout,omitempty"`
+
+	// ExtraFields is an arbitrary YAML blob that can be used by non-kubebuilder
+	// plugins for plugin-specific configure.
+	ExtraFields map[string]interface{} `json:"plugins,omitempty"`
 }
 
 // IsV1 returns true if it is a v1 project
-func (config Config) IsV1() bool {
-	return config.Version == Version1
+func (c Config) IsV1() bool {
+	return c.Version == Version1
 }
 
 // IsV2 returns true if it is a v2 project
-func (config Config) IsV2() bool {
-	return config.Version == Version2
+func (c Config) IsV2() bool {
+	return c.Version == Version2
 }
 
 // HasResource returns true if API resource is already tracked
-func (config Config) HasResource(target GVK) bool {
+func (c Config) HasResource(target GVK) bool {
 	// Return true if the target resource is found in the tracked resources
-	for _, r := range config.Resources {
+	for _, r := range c.Resources {
 		if r.isEqualTo(target) {
 			return true
 		}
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/gobuffalo/flect v0.1.5
 	github.com/golang/protobuf v1.3.1 // indirect
 	github.com/kr/pretty v0.1.0 // indirect
+	github.com/mitchellh/mapstructure v1.2.2
```
camilamacedo86: I understand that Kubebuilder has a straight politic to add new dependencies for some reasons as for example keep the Maintenance and supportability ability. 

Looking at the dep https://github.com/mitchellh/mapstructure I am not sure if it is something really adopted by the community and that could/should be added here(IMO we should NOT use the mitchellh's module). So, if it has no other way to do be solved whiteout this module then, I believe that this PR will require the admin approvals. @mengqiy wdyt?

pkg/model/config/config.go
```diff
@@ -88,27 +100,27 @@ func (config Config) HasResource(target GVK) bool {
...
+	c.Resources = append(c.Resources, gvk)
 	return true
 }
 
 // GVK contains information about scaffolded resources
 type GVK struct {
-	Group   string `json:"group,omitempty"`
-	Version string `json:"version,omitempty"`
-	Kind    string `json:"kind,omitempty"`
+	Group   string `json:"group,omitempty" mapstructure:"group"`
```
camilamacedo86: Why we need the mapstructure here?

estroz:
`mapstructure` is a well maintained, stable library for parsing arbitrary keyed data, ex. a config. While it is possible to write a config system with both arbitrary and well-defined keys without it, mapstructure is built to efficiently managed such a system.

Adirio:
I like the use of `afero` to mock the filesystem for tests, and we probably need a `extraFields` field somewhere in the YAML but I would probably organize it a bit more so that is not that random. For example I would go for:

```
version: "2"
...
plugins:
    - pluginName: {}
    - pluginName2: {}
```

Aside from that, what is the new dependency adding? Can we do the same without the import?

estroz:
@Adirio if we go for a system organized like you suggest, we should be able to avoid using `mapstructure`. The original problem was dealing with arbitrary, flattened keys next to well-defined keys.

Adirio:
> @Adirio if we go for a system organized like you suggest, we should be able to avoid using `mapstructure`. The original problem was dealing with arbitrary, flattened keys next to well-defined keys.

If `mapstructure` was used for that I definetely prefer my approach. IMO it's better organized and we remove a dependency.

estroz:
@Adirio @camilamacedo86 done.

pkg/model/config/config.go
```diff
@@ -113,3 +120,87 @@ func (r GVK) isEqualTo(other GVK) bool {
...
+// be used for custom configuration objects.
+func (c *Config) EncodeExtraFields(key string, extraFieldsObj interface{}) error {
+	// Short-circuit v1
+	if c.IsV1() {
+		return fmt.Errorf("v1 project configs do not have extra fields")
+	}
+	// Ignore zero'd objects so we don't add empty map keys.
+	if reflect.ValueOf(extraFieldsObj).IsZero() {
+		return nil
+	}
```
Adirio: Do we really need this check? If the field was added to the ExtraFields map, I think we should encode it, the default doesn't need to be the zero value (for example versioning defaulting to 0 doesn't make much sense in case some version wants to be stored).

It has the added benefit of removing `reflect` which I think is nice.

pkg/model/config/config.go
```diff
@@ -116,3 +123,83 @@ func (r GVK) isEqualTo(other GVK) bool {
...
+}
+
+// EncodeExtraFields encodes extraFieldsObj in c. This method is intended to
+// be used for custom configuration objects.
+func (c *Config) EncodeExtraFields(key string, extraFieldsObj interface{}) error {
+	// Short-circuit v1
+	if c.IsV1() {
+		return fmt.Errorf("v1 project configs do not have extra fields")
+	}
+	// Shouldn't expect random objects to have mapstructure tags.
```
Adirio: I think this comment should be rephrased as `mapstructure` is not being used anymore.

go.mod
DirectXMan12: We copied validations because it was trivial.  Non-trivial dependencies are fine.  That being said: this code already more-or-less exists in kubernetes in the unstructured codec, if we want to use it (unclear that it makes much of a difference).

pkg/model/config/config.go
```diff
@@ -46,22 +49,26 @@ type Config struct {
 
 	// Layout contains a key specifying which plugin created a project.
 	Layout string `json:"layout,omitempty"`
+
+	// ExtraFields is an arbitrary YAML blob that can be used by non-kubebuilder
+	// plugins for plugin-specific configure.
+	ExtraFields map[string]interface{} `json:"plugins,omitempty"`
```
DirectXMan12: non-blocking: can we wrap this in a type so that people don't depend on the exact deserialization internals?  If we want people to use types and the DecodeExtraFields here, we might want to wrap `interface{}` in something.

estroz: Sounds reasonable. I'll make a follow-up PR so we can discuss specifics there.

PR #1445, opened 2020-03-23

estroz:
/cc @mengqiy @DirectXMan12 @Adirio @camilamacedo86


Adirio:
For future reference, I think that showing the resolved conflicts here could help check that all the added features were merged successfully by reviewers.

PR #1454, opened 2020-04-01
 Clean the GO mod files by removing the unnecessary indirect imports

camilamacedo86:
**Description**

- Clean the GO mod files by removing the unnecessary indirect imports 
- fix make generate target to do cause the scenario again.


Adirio:
My understanding of `go mod` is pretty basic, I just know what it is needed to use them, but not the internals, so I'm not sure if those indirect imports are needed or not.

P.S.: Checking that tests pass in this case doesn't seem enough because probably none of those of those libraries made a backwards incompatible change so removing the indirect requirement may just get the last version and still work. But in case they make a backwards incompatible change it may break the build.

camilamacedo86:
@assign @mengqiy

mengqiy:
I'd trust the go modules to do the job of tracking necessary entries in the go.mod file.
@camilamacedo86 What you can try is that first remove indirect entries and then run `go mod tidy`. If go modules add some entries back, then we probably should keep them.

camilamacedo86:
Hi @mengqiy, 

Unfortunately, `go mod tidy` has been not removed the unnecessary imports from the `go.mod`.  It has been just updating the `go.sum`. note that it has some issues opened as well. see https://github.com/golang/go/issues/33008,  
 
See for example that "github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e" is not required in the main module. 

```
$ go mod why github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e
# github.com/niemeyer/pretty
sigs.k8s.io/kubebuilder/internal/config
sigs.k8s.io/yaml
gopkg.in/yaml.v2
gopkg.in/yaml.v2.test
gopkg.in/check.v1
github.com/niemeyer/pretty

# v0.0.0-20200227124842-a10e7caefd8e
(main module does not need package v0.0.0-20200227124842-a10e7caefd8e)
camilamacedo@Camilas-MacBook-Pro ~/go/src/sigs.k8s.io/kubebuilder (master) $  
```

**Also, note that these indirect imports are just added when we run  `make generated` because of the test data. And then, see that all indirect imports are in the same situation.** 

> # v0.0.0-20200301022130-244492dfa37a
> (main module does not need package v0.0.0-20200301022130-244492dfa37a)

```
# v0.0.0-20200302083256-062a44052db1
(main module does not need package v0.0.0-20200302083256-062a44052db1)
```

```
# v0.0.0-20191204190536-9bdfabe68543
(main module does not need package v0.0.0-20191204190536-9bdfabe68543)
```

```
# v1.0.0-20200227125254-8fa46927fb4f
(main module does not need package v1.0.0-20200227125254-8fa46927fb4f)
```

```
$ go mod why gopkg.in/yaml.v2 v2.2.8 // indirect
# gopkg.in/yaml.v2
sigs.k8s.io/kubebuilder/internal/config
sigs.k8s.io/yaml
gopkg.in/yaml.v2

# v2.2.8
(main module does not need package v2.2.8)

# /
(main module does not need package /)

# indirect
(main module does not need package indirect)

```

mengqiy:
> Also, note that these indirect imports are just added when we run make generated.

Though we do `go mod tidy` in https://github.com/kubernetes-sigs/kubebuilder/blob/29d1dc1589b5e61b3d4211e2b8dd0691adf5c153/Makefile#L50,  if `go get` in  https://github.com/kubernetes-sigs/kubebuilder/blob/29d1dc1589b5e61b3d4211e2b8dd0691adf5c153/Makefile#L63 still cause the problem, we should do `go get` in a temp directory. This can avoid it messing up the go.mod file.

camilamacedo86:
HI @mengqiy, 

Now the go mod is clean the make generate target is fixed. I mean it will no longer install the indirect modules.

Makefile
```diff
@@ -57,11 +66,6 @@ generate-testdata: ## Update/generate the testdata in $GOPATH/src/sigs.k8s.io/ku
 generate-vendor: ## (Deprecated) Update/generate the vendor by using the path $GOPATH/src/sigs.k8s.io/kubebuilder-test
 	GO111MODULE=off ./generate_vendor.sh
 
-.PHONY: generate-setup
-generate-setup: ## Current workarround to generate the testdata with the correct controller-gen version
-	- rm -rf $(CONTROLLER_GEN_BIN_PATH)
-	- GO111MODULE=on go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.5
```
mengqiy, line 63: You don't need to remove this section. I imagine that what you can do here is
1) create a temp directory (e.g. using https://golang.org/pkg/io/ioutil/#TempDir)
2) cd to that directory
3) run go get with go modules on
4) cd to the previous working directory
5) remove the temp dir.

camilamacedo86: The idea of this makefile target was just to ensure that the correct version will be used.
Note that we we just need to remove any pre-existent installation to ensure that the version defined in the makefile will be used. So, the point is that it in POV it is not required at all and because of this I keep just the rm -rf line. See: https://github.com/kubernetes-sigs/kubebuilder/pull/1454/files#diff-b67911656ef5d18c4ae36cb6741b7965R57

Makefile
```diff
@@ -45,7 +52,9 @@ install: ## Build and install the binary with the current source code. Use it to
 
 .PHONY: generate
 generate: ## Update/generate all mock data. You should run this commands to update the mock data after your changes.
-	make generate-setup
+	# if an controller-gen be locally installed already it will be removed to ensure that the generate will
+	# get and use the correct version defined in the testdata makefiles instead of an previous installation.
```
mengqiy, line 56: This statement doesn't match what is done.
controller-gen is not installed after removing.

camilamacedo86: HI @mengqiy.

See: 

Screenshot 2020-04-13 at 17 55 26

And then: 

Screenshot 2020-04-13 at 17 55 53

mengqiy: Ha... You rely on the Makefile in the generated project to install the controller-gen for you.
I didn't realized it. It's not intuitive, but seems to be OK.

mengqiy: Please update the comment here to reflect how it work.
Then it should be OK to go.

camilamacedo86: Let me know if you are ok with the updates in the comments.

PR #1447, opened 2020-03-27
cli: check if project is v1 when validating default plugins

estroz:
See https://github.com/kubernetes-sigs/kubebuilder/pull/1403/files#r399502355.

/kind bug

PR #1455, opened 2020-04-01
cli: check if project is v1 when validating default plugins

estroz:
Clone of #1447 since travis isn't updating its status.


estroz:
Still not updating status.

PR #1460, opened 2020-04-02
update the go check for the minimal required version

camilamacedo86:
**Description**
Upgrade the go check to ensure that the version used is according to the minimal requirements

### DIFF
--- a/cmd/internal/go_version.go
+++ b/cmd/internal/go_version.go
@@ -68,8 +68,8 @@ func checkGoVersion(verStr string) error {
 		return fmt.Errorf("error parsing minor version '%s': %s", m[2], err)
 	}
 
-	if major < 1 || minor < 11 {
-		return fmt.Errorf("requires version >= 1.11")
+	if major < 1 || minor < 13 {
+		return fmt.Errorf("requires version >= 1.13")
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


Adirio:
IIRC I already tried this and there were more changes needed. Some involved modifying some of the CI tools (not sure if it was Travis, Prow, KinD or the api server images). Also, you would need to require 13+ in https://github.com/kubernetes-sigs/kubebuilder/blob/29d1dc1589b5e61b3d4211e2b8dd0691adf5c153/cmd/internal/go_version.go#L71-L73

camilamacedo86:
@Adirio, 

> IIRC I already tried this and there were more changes needed. Some involved modifying some of the CI tools (not sure if it was Travis, Prow, KinD or the api server images). Also, you would need to require 13+ in

Did you find any place that is using a go version < 1.13 ?  

See that Travis form example is using 1.13 https://github.com/kubernetes-sigs/kubebuilder/blob/master/.travis.yml#L13 and it is the info the Kuebebuilder docs. Also, since all CI tests passed with the check I would assume that has not a place using a lower version as well. 

However, please let me know if you see any reason for we do not move forward here.

Adirio:
I didn't realize that a `lgtm` was going to merge it.

@DirectXMan12 @mengqiy @droot if you consider this shouldn't be merged we can clease the mess, sorry.

mengqiy:
LGTM

PR #1462, opened 2020-04-07
Using utilruntime.Must() for adding schemes

christopherhein:
Add the use of `utilruntime.Must()` from apimachinery to catch errors with scheme configurations. 

Signed-off-by: Chris Hein <me@chrishein.com>

### DIFF
--- a/pkg/scaffold/internal/templates/v2/main.go
+++ b/pkg/scaffold/internal/templates/v2/main.go
@@ -113,7 +113,7 @@ const (
 	//  change, and thus should be done for next project version.
 	multiGroupControllerImportCodeFragment = `%scontroller "%s/controllers/%s"
 `
-	addschemeCodeFragment = `_ = %s.AddToScheme(scheme)
+	addschemeCodeFragment = `utilruntime.Must(%s.AddToScheme(scheme))
 `
 	reconcilerSetupCodeFragment = `if err = (&controllers.%sReconciler{
 		Client: mgr.GetClient(),
@@ -206,6 +206,7 @@ import (
 	"flag"
 	"os"
 	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -219,7 +220,7 @@ var (
 )
 
 func init() {
-	_ = clientgoscheme.AddToScheme(scheme)
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 
 	%s
 }
--- a/testdata/project-v2-addon/main.go
+++ b/testdata/project-v2-addon/main.go
@@ -21,6 +21,7 @@ import (
 	"os"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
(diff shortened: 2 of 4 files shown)


pkg/scaffold/internal/templates/v2/main.go
```diff
@@ -113,7 +113,7 @@ const (
 	//  change, and thus should be done for next project version.
 	multiGroupControllerImportCodeFragment = `%scontroller "%s/controllers/%s"
 `
-	addschemeCodeFragment = `_ = %s.AddToScheme(scheme)
+	addschemeCodeFragment = `utilruntime.Must(%s.AddToScheme(scheme))
```
camilamacedo86, line 116: Sorry, but could you add in the first comment one example of scenario that is solved/attended with?

christopherhein: I'm not directly experiencing issues from this but from a best practices side in most other implementations of `AddingToScheme()` they implement `utilruntime.Must()` to ensure panics causing the controller to crash. https://github.com/search?q=org%3Akubernetes+utilruntime.Must&type=Code and https://github.com/search?q=org%3Akubernetes-sigs+utilruntime.Must&type=Code

Potentially it doesn't matter?

camilamacedo86: HI @christopherhein, 

I checked the Must impl [here]( https://github.com/kubernetes/apimachinery/blob/master/pkg/util/runtime/runtime.go#L169-L173), and it just causes a panic error.

> Regards that would be a `best practices` 

See [here](https://github.com/search?q=org%3Akubernetes-sigs+AddToScheme%28scheme%29&type=Code) that it is not used in the biggest part of the places where the addSchema is used in the example provided. Also, if it is a good practice then: 
- Should we not add this info in the [AddToScheme doc](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/scheme#Builder.AddToScheme)? 
- Why is the controller-runtime not using it all examples/scenarios?
 
> Regards Kubebuilder project scenarios

Note that the errors returned has been explicitly ignored in the Kubebuilder projects, and then I was curious if you face any issue that justifies it. 

Also, start to cause an panic error in all these scenarios may not be the expected behaviour. For example, will the panic occurs if the CRD required not be found in the cluster? What are the possible errors/scenarios where an error can be returned from it? Would be required to kill the project in all these circumstances?

Did you already checked that?

mengqiy: IMHO it's reasonable to just let the controller fail if it can't even add scheme.
@droot @DirectXMan12 Thoughts?

> will the panic occurs if the CRD required not be found in the cluster?

From my understanding no.

christopherhein: Hey @camilamacedo86 sorry for the delayed response here. For now I've put in a PR to also add documentation to the scheme - https://github.com/kubernetes-sigs/controller-runtime/pull/894

I had looked into a handful of the other implementation and most will cause Panics already for example if you register all the schemes from `client-go this https://github.com/kubernetes/client-go/blob/master/kubernetes/scheme/register.go#L133 will panic if there are issues.

I discussed this yesterday with @DirectXMan12  and they'd mentioned that this "either a) it was an oversight, or b) that pattern turned up after we had started writing our examples" & "we really should be checking the return value"

christopherhein: It seems like the errors usually come from errors in the conversion or defaulting registration, since I've been able to repro panics by doing things like registering schemes of CRDs with the same GVK as standard k8s objects.

mengqiy:
Per https://github.com/kubernetes-sigs/kubebuilder/pull/1462#discussion_r406980310, I guess we are on the same page. We can get this in.
/lgtm
/approve

christopherhein:
thanks @mengqiy!

PR #1465, opened 2020-04-08
plugins: move `--plugins` from global to `init`-only

estroz:
Per discussion in https://github.com/kubernetes-sigs/kubebuilder/pull/1403#discussion_r405014466:
>[H]aving a global --plugins flag is not in scope for phase 1 since the type of plugin used in init solely determines what plugin is used for subsequent commands (the layout config key). A global --plugins flag will be discussed in phase 2: #1378.

~There still may be a use case for phase 1 global `--plugins`: I scaffold a project with `--plugins=go-x` then want to create an API using plugin `go-y`, which "understands" a project initialized by `go-y`; without a global `--plugins` flag I would have to change my config's `layout` value.~

~Thoughts?~

/cc @DirectXMan12 @mengqiy @joelanford

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -39,7 +39,7 @@ project.
 
 	projectVersionFlag = "project-version"
 	helpFlag           = "help"
-	pluginNamesFlag    = "plugins"
+	pluginsFlag        = "plugins"
 )
 
 // CLI interacts with a command line interface.
@@ -70,8 +70,8 @@ type cli struct {
 	pluginsFromOptions map[string][]plugin.Base
 	// Default plugins injected by options.
 	defaultPluginsFromOptions map[string][]plugin.Base
-	// A mapping of plugin name to version passed by to --plugins.
-	cliPluginKeys map[string]string
+	// A plugin key passed to --plugins on invoking 'init'.
+	cliPluginKey string
 	// A filtered set of plugins that should be used by command constructors.
 	resolvedPlugins []plugin.Base
 
@@ -88,7 +88,6 @@ func New(opts ...Option) (CLI, error) {
 		defaultProjectVersion:     internalconfig.DefaultVersion,
 		pluginsFromOptions:        make(map[string][]plugin.Base),
 		defaultPluginsFromOptions: make(map[string][]plugin.Base),
-		cliPluginKeys:             make(map[string]string),
 	}
 	for _, opt := range opts {
 		if err := opt(c); err != nil {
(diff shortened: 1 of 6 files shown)


estroz:
/hold

while we discuss

pkg/cli/cli.go
```diff
@@ -192,38 +190,34 @@ func (c *cli) initialize() error {
...
+	// For all other commands, a config's 'layout' key is used. Since both
+	// layout and --plugins values can be short (ex. "go/v2.0.0") or unversioned
+	// (ex. "go.kubebuilder.io") keys or both, their values may need to be
+	// resolved to known plugins by key.
+	plugins := c.pluginsFromOptions[c.projectVersion]
 	switch {
-	case len(c.cliPluginKeys) != 0:
-		// Filter plugins by keys passed in CLI.
-		filterKeys = c.cliPluginKeys
+	case c.cliPluginKey != "":
```
camilamacedo86, line 204: I'd prefer len instead of != ""

pkg/cli/cli.go
```diff
@@ -302,7 +282,7 @@ func (c cli) validate() error {
 	}
 	// If --plugins is not set, no layout exists (no config or project is v1 or v2),
 	// and no defaults exist, we cannot know which plugins to use.
-	if (!c.configured || c.projectVersion != config.Version3) && len(c.cliPluginKeys) == 0 {
+	if (!c.configured || c.projectVersion != config.Version3) && c.cliPluginKey == "" {
```
camilamacedo86, line 285: same here

pkg/cli/cli.go
```diff
@@ -322,7 +302,8 @@ func (c cli) validate() error {
 	}
 
 	// Validate plugin keys set in CLI.
-	for pluginName, pluginVersion := range c.cliPluginKeys {
+	if c.cliPluginKey != "" {
```
camilamacedo86, line 305: same

pkg/cli/cli.go
estroz: Both are idiomatic. It should be clear that `c.cliPluginKey` is a string, hence `!= ""`. Ditto below.

camilamacedo86: the problem with != "" is that will not work for !=" " when should for example.

pkg/cli/cli.go
```diff
@@ -257,32 +252,19 @@ func (c *cli) parseBaseFlags() error {
...
-		pluginName, pluginVersion := plugin.SplitKey(key)
-		if pluginName == "" {
-			return fmt.Errorf("plugin key %q must at least have a name", key)
-		}
-		if _, exists := c.cliPluginKeys[pluginName]; exists {
-			return fmt.Errorf("duplicate plugin name %q", pluginName)
-		}
-		c.cliPluginKeys[pluginName] = pluginVersion
-	}
+	c.cliPluginKey = strings.TrimSpace(c.cliPluginKey)
```
camilamacedo86, line 267: 👍

camilamacedo86:
Hi @estroz missing run `make generate`.

testdata/project-v3-addon/go.mod
```diff
@@ -9,5 +9,5 @@ require (
 	k8s.io/apimachinery v0.17.2
 	k8s.io/client-go v0.17.2
 	sigs.k8s.io/controller-runtime v0.5.0
-	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200408160153-97d9e01751e0
+	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200414151614-92702a3972f3
```
camilamacedo86, line 12: it needs to be rebased with the master

pkg/cli/cli.go
```diff
@@ -70,8 +70,8 @@ type cli struct {
 	pluginsFromOptions map[string][]plugin.Base
 	// Default plugins injected by options.
 	defaultPluginsFromOptions map[string][]plugin.Base
-	// A mapping of plugin name to version passed by to --plugins.
-	cliPluginKeys map[string]string
+	// A plugin key passed to --plugins on invoking 'init'.
+	cliPluginKey string
```
mengqiy, line 74: I feel a little confused, maybe because I forgot some context.
Only one plugin can be passed to `--plugins`?

estroz: For now yeah. Having multiple plugin arguments to `--plugins` only makes sense if you can chain them together (phase 2).

mengqiy:
/approve
/lgtm
Feel free to go ahead and merge this when you are ready.

Issue #1475, opened 2020-04-16
Let's create a cookbook with useful patterns


hendrikmaus:
Hey everyone, I would like to get some feedback from the community on an idea I had, which is to start a cookbook with useful patterns for operator development.

What I immediately appreciated about the kubebuilder project was the git book - fantastic effort. Makes the entry into the operator realm very smooth and creates a good understanding of the basics.
Now, correct me if I am wrong, often enough the real life use cases stretch beyond the very basics and this is where I want to hook in and start to collect more knowledge as I could not find anything out there.

For example: I specifically searched for the involvement of external resources. May it be reconciling a load balancer of a cloud provider, listening to a message queue, watching a GitHub PR, you name it.
This kind of problem is never touched in the tutorials out there, but I think it is pretty common - at least that is what my gut feeling says.

I will cut it here and would be very happy to discuss your ideas.

alexeldeib:
Dynamically adding new controllers at runtime would be a good cookbook example, as well as showcasing alternative structures for projects as they grow.

I'm happy to help write some samples if we develop a list of desired recipes 🙂

otakumike:
I think this is a fantastic idea. I've been writing operators for a little while now and even though the git book is very helpful, it leaves me wondering if I'm following best practices and designing my controllers well. Having a resource I can look to for examples of real-world scenarios would be a great help I think

alexeldeib:
As far as implementation, I think this can _probably_  be as simple as adding a top level section (maybe near reference) to https://book.kubebuilder.io/ and adding full source code back in github

hendrikmaus:
Ok, initial PR is opened. Should we continue to collect recipes in this issue for now?
I'd also be happy to write some samples.

drpaneas:
> I'd also be happy to write some samples.

+1 -- I think at least one example would be great for the beginning, so to establish what kind of contributions are expected in some sort of way.

PR #1477, opened 2020-04-18
start a cookbook section in the book per #1475

hendrikmaus:
**Description:**
The changeset adds a new top level section to the book below "Reference", called "Cookbook".

**Motivation:**
The existing book does a fantastic job introducing new users to the operator realm. I want to enable the community to create a place in the book to share and validate each others most common design patterns and problem solutions from daily operator development as it often goes well beyond the onboarding tutorials.

Issue: https://github.com/kubernetes-sigs/kubebuilder/issues/1475

### DIFF
--- a/docs/book/src/cookbook/cookbook.md
+++ b/docs/book/src/cookbook/cookbook.md
@@ -1,0 +1,32 @@
+# Cookbook
+
+A curated collection of operator development patterns, use-cases and generic\
+recommendations.
+
+## Rationale
+
+While the tutorial part of this book provides onboarding into the operator
+realm, real world use cases often surpass the tutorial example(s). Hence we
+want to provide a place to share common implementation patterns in order
+to enable other users, may they be operator experts or novices, a palce to both
+share and validate common approaches they encounter while developing operators.
+
+## Recipies
+
+<aside class="note">
+
+<h1>Following Along</h1>
+
+Note that code examples of cookbook entries live as literate Go files in the
+book source directory:
+[docs/book/src/cookbook/testdata][cookbook-sources]. You may find both snippets
+and fully runnable projects among the examples.
+
+[cookbook-sources]: https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/cookbook/testdata
+
+</aside>
+
+*If you are reading this and you are wondering where the entries are: there are
+none yet. It is the perfect time for you to contribute something from your
+experience in developing operators right now.*
+
--- a/docs/book/src/cookbook/cookbook.md
+++ b/docs/book/src/cookbook/cookbook.md
@@ -8,7 +8,7 @@ recommendations.
 While the tutorial part of this book provides onboarding into the operator
 realm, real world use cases often surpass the tutorial example(s). Hence we
 want to provide a place to share common implementation patterns in order
-to enable other users, may they be operator experts or novices, a palce to both
+to enable other users, may they be operator experts or novices, a place to both
 share and validate common approaches they encounter while developing operators.
 
 ## Recipies
--- a/docs/book/src/cookbook/cookbook.md
+++ b/docs/book/src/cookbook/cookbook.md
@@ -1,6 +1,6 @@
 # Cookbook
 
-A curated collection of operator development patterns, use-cases and generic\
+A curated collection of operator development patterns, use-cases and generic
 recommendations.
 
 ## Rationale
@@ -11,7 +11,7 @@ want to provide a place to share common implementation patterns in order
 to enable other users, may they be operator experts or novices, a place to both
 share and validate common approaches they encounter while developing operators.
 
-## Recipies
+## Recipes
 
 <aside class="note">
 
--- a/docs/book/src/SUMMARY.md
+++ b/docs/book/src/SUMMARY.md
@@ -79,4 +79,6 @@
 
 ---
 
+- [Cookbook](./cookbook/cookbook.md)
+
 [Appendix: The TODO Landing Page](./TODO.md)


docs/book/src/cookbook/cookbook.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Rationale
+
+While the tutorial part of this book provides onboarding into the operator
+realm, real world use cases often surpass the tutorial example(s). Hence we
+want to provide a place to share common implementation patterns in order
+to enable other users, may they be operator experts or novices, a place to both
+share and validate common approaches they encounter while developing operators.
+
+## Recipies
```
alexeldeib, line 14:
```suggestion
## Recipes
```

alexeldeib:
If you haven't already seen it - https://deploy-preview-1477--kubebuilder.netlify.app/cookbook/cookbook.html

Would you mind checking the Travis failure? It doesn't seem related, but should be a quick fix. If not I can pick it up.

/ok-to-test

docs/book/src/cookbook/cookbook.md
```diff
@@ -0,0 +1,32 @@
+# Cookbook
+
+A curated collection of operator development patterns, use-cases and generic\
```
alexeldeib, line 3: typo? why force the line break here and no where else?

alexeldeib:
Would you also remind rebasing and removing the issue # from your commit message? It generates a lot of noise over time in the associated issue. When you want to close the issue via PR, you can just put "fixes #<whatever>" in the body of the Github PR, not in the commit message.

image

docs/book/src/cookbook/cookbook.md
```diff
@@ -0,0 +1,32 @@
...
+<aside class="note">
+
+<h1>Following Along</h1>
+
+Note that code examples of cookbook entries live as literate Go files in the
+book source directory:
+[docs/book/src/cookbook/testdata][cookbook-sources]. You may find both snippets
+and fully runnable projects among the examples.
+
+[cookbook-sources]: https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/cookbook/testdata
```
camilamacedo86, line 25: It is an invalid link.

docs/book/src/cookbook/cookbook.md
```diff
@@ -0,0 +1,32 @@
...
+
+</aside>
+
+*If you are reading this and you are wondering where the entries are: there are
```
drpaneas, line 29: I like the idea of tips, tricks and advices one should know when writes an operator! I am not sure though I understand what a valid/expected contribution might be considered in this though `testdata` -- which by the way the word `test` is confusing me in this context. If you could come up with an example of a such a contribution already, that would be great.

docs/book/src/cookbook/cookbook.md
mengqiy: To be clear, our top level `testdata` directory holds several generated projects.
We also have several `testdata` directory in `docs`. 
`testdata` is a magic name in golang, go build ignores it.
IMO it's fine to have `testdata` directory `docs` directory.

camilamacedo86: HI @mengqiy,

So, if I understood, the idea proposed here with this PR is to create a directory where many examples and projects will start to be added by the contributors. Am I right?  And then, it means that any content added on it would be "blessed" by the project as a good practice and indeed started to be supported and maintained by it. 

Then, my concern is:  
- How much effort will be required to understand and maintain manually these examples and projects with an example provided by the community? Will it be something achievable in a medium long term? ( IHMO it is not )
- Is it the best way to improve the docs content to provide further explanations?  
- Should not better we improve the doc providing more specific examples about the subjects with explanations over it then add many projects that we and users has no context about in a dir in the repo? 
- Is it something that we really would like to persuade? 

WDYT @DirectXMan12 and @joelanford

hendrikmaus: I am also concerned with keeping every fully functional example which could be added up to date and working. I reckon code snippets, to emphasize certain practices, embedded in the docs would be easier to maintain as, for _example_, an IDE will find references in arbitrary strings when something is refactored (rename, change of signature, move etc.)

docs/book/src/cookbook/cookbook.md
hendrikmaus: I borrowed this section from another part of the book as I thought it has been established as the idiomatic way to provide examples throughout the docs.

I agree that an example would be helpful. Right now, especially after diving into https://github.com/kubernetes-sigs/kubebuilder/pull/1477#discussion_r415032376, I am tending to start with a pure-test example which would only use embedded code snippets if it has to at all.

docs/book/src/cookbook/cookbook.md
DirectXMan12: Before I finishing fully parsing this comment: `testdata` is just a special directory that is ignored by `go` tooling when calculating dependencies, etc.  That's all.  We put full probjects/examples in `testdata` (as well as book source code) so that it doesn't affect the dep list of the main project, and so that we can easily get the main project's tooling to ignore it most of the time.

DirectXMan12: FWIW, I do like this idea (and we've discussed similar before).  I think we just want to keep it clear that this is a bit more tightly controlled than, say, a wiki -- we want focused, well-maintained examples.

That being said, I think literate code is the way to go, since we can check that it builds, have ide tooling as @hendrikmaus suggested, etc.

docs/book/src/cookbook/cookbook.md
DirectXMan12: FYI on the testdata directory: https://github.com/kubernetes-sigs/kubebuilder/pull/1477#discussion_r416165222

docs/book/src/cookbook/cookbook.md
camilamacedo86: Hi @DirectXMan12, 

what is your idea for we ensure and keep well-maintained these examples in a long-term?

Issue #1351, opened 2020-02-04
Controller-gen lifecycle (install/update) enhancement proposal 


droot:
This issue proposes a new workflow for managing life-cycle of `controller-gen` for KubeBuilder projects. The ideas in the proposal are inspired from the discussion in [this PR](https://github.com/kubernetes-sigs/kubebuilder/pull/1289).

**Problem:**
In the current workflow of Kubebuilder projects, Makefile checks if `controller-gen` executable is available in the PATH, if not, it installs `controller-gen` using `go get /path/to/specific/version/of/controller-tools/controller-gen` globally. Current workflow leads to following problems:

- If multiple developers are collaborating on a KB project and they all are using different versions of `controller-gen`, will end up with different artifacts (CRDs, deepcopy, RBAC manifests etc.). This is the problem we (KubeBuilder collaborators) are running in to wrt to `test/project-v2` project.

- Makes generated artifacts (CRDs, deepcopy, RBAC etc) environment dependent hurting reproducibility.

**Proposal:**
Introduce `ControllerGenVersion` field in KubeBuilder `PROJECT` file:
```sh

# controller-gen version to use
ControllerGenVersion: v02.4

```

Create a new directory `bin/` in the project to store `controller-gen` binary (and potentially other binaries in the future). Tweak the logic for the Makefile target `controller-gen` to ensure `bin/controller-gen` points to controller-gen version specified in `PROJECT` file. `./bin/controller-gen --version` command can be used to determine the currently installed version.

Ensure that Makefile uses `bin/controller-gen` instead of global `controller-gen` binary.

If a developer wants to bump the version of `controller-gen` in their project, they can simply change the version in `PROJECT` file and run `make controller-gen` to switch to correct version.

**BONUS:**
we can add another target in Makefile to update to latest version of `controller-gen` by introducing new target `update-controller-gen` which can:

- Query github releases for `controller-gen` and show latest available version and prompt to update.
- If user says 'yes`, then updates the `ControllerGenVersion:<new-version>` in `PROJECT` file and runs `make controller-gen`.

/kind feature

I am happy to help with reviews/brainstorm/get-this-merged if someone wants to take this up once we agree if this is a good idea.

/cc  @dionysius @mengqiy @estroz @Adirio @camilamacedo86 @DirectXMan12

Adirio:
How will we keep the Makefile and the PROJECT configuration file synced? Will we parse the PROJECT file from the Makefile?

[...]

PR #1478, opened 2020-04-22
Makefile: Support multiple GOPATH directories

### LINKED ISSUE #1402: Failed to parse the correct GOPATH within the Makefile
`# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
ifeq (,$(shell go env GOBIN))
GOBIN=$(shell go env GOPATH)/bin
else
GOBIN=$(shell go env GOBIN)
endif`

It takes $GOPATH as the golang install path directly but fails to run controller-gen command if $GOPATH is made of multiple directories like "/usr/local/golang:/usr/local/gopath" which must be separated

drpaneas:
If $GOPATH is consisted of multiple directories separated by ':'
then pick the first one. This is what 'go get' does as well.

Fixes #1402

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,9 +20,17 @@
 export GOPROXY?=https://proxy.golang.org/
 CONTROLLER_GEN_BIN_PATH := $(shell which controller-gen)
 
+# Default to the first directory in the list, if GOPATH has multiple paths
+CHECK_GOPATH=$(shell go env GOPATH)
+ifneq (,$(findstring :,$(CHECK_GOPATH)))
+GOPATH=$(firstword $(subst :, ,$(CHECK_GOPATH)))
+else
+GOPATH=$(shell go env GOPATH)
+endif
+
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
-GOBIN=$(shell go env GOPATH)/bin
+GOBIN=$(GOPATH)/bin
 else
 GOBIN=$(shell go env GOBIN)
 endif
@@ -46,7 +54,7 @@ build: ## Build the project locally
 .PHONY: install
 install: ## Build and install the binary with the current source code. Use it to test your changes locally.
 	make build
-	cp ./bin/kubebuilder $(shell go env GOPATH)/bin/kubebuilder
+	cp ./bin/kubebuilder $(GOBIN)/kubebuilder
 
 ##@ Development
 
--- a/common.sh
+++ b/common.sh
@@ -29,6 +29,13 @@ if [[ -n "${MODULES_OPT}" && -n "${MODULES_ENABLED}" ]]; then
     MOD_OPT="-mod=${MODULES_OPT}"
 fi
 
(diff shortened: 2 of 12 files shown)


Adirio:
Linux-path-only solution, would have to be adapted to WIndows when we add compatibility.

Travis CI seems to have failed but is under maintainance right now.

drpaneas:
Yes, this works only for Linux/Mac.
The CI seems borked as it fails for another reason, see:

```diff
git diff testdata
diff --git a/testdata/project-v2-addon/go.mod b/testdata/project-v2-addon/go.mod
index ec17ee5..adaa9b0 100644
--- a/testdata/project-v2-addon/go.mod
+++ b/testdata/project-v2-addon/go.mod
@@ -9,5 +9,5 @@ require (
 	k8s.io/apimachinery v0.17.2
 	k8s.io/client-go v0.17.2
 	sigs.k8s.io/controller-runtime v0.5.0
-	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200413154948-0284e6ed8028
+	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200415210853-85eb326a6add
 )
Generate Testdata failed!
Please, if you have changed the scaffolding make sure you have run: make generate
```

Since I didn't touch `go.mod` I guess the CI is _diff_-ing against an older version instead of the HEAD. Is there anything I can do to fix that or just wait?

Makefile
```diff
@@ -46,7 +54,7 @@ build: ## Build the project locally
 .PHONY: install
 install: ## Build and install the binary with the current source code. Use it to test your changes locally.
 	make build
-	cp ./bin/kubebuilder $(shell go env GOPATH)/bin/kubebuilder
+	cp ./bin/kubebuilder $(GOBIN)/kubebuilder
```
camilamacedo86, line 58: This is the Makefile with the tool target and not the Makefile that is scaffolded for the projects. 
I understand that you would like to change the Makefile which is scaffolded for the operator's projects. Am I right? 
if yes, you need to apply the changes in: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/scaffold/internal/templates/makefile.go and then, run `make generate` to update the testdata dir with mock projects.

drpaneas: ah crap, there are multiple makefiles? I need to check this. Thanks @camilamacedo86 for spotting this! I will go through the Makefile throughly and try to figure out the connection and what's happening.

In the meantime, I will add a `WIP` on this one.

drpaneas:
Ok, I have updated the PR with (hopefully) all the points in code where GOPATH is used. Truthfully I believe it's ugly way of fixing this issue, yet again most of the projects do not support multiple GOPATH directories so they just don't care.

WDYT?

camilamacedo86:
IHMO the best way we solved it is by doing as suggested here: https://github.com/kubernetes-sigs/kubebuilder/issues/1351

drpaneas:
Then I guess it makes more sense to work on it after the 1351 gets implemented to see what needs to be changed then.

PR #1482, opened 2020-04-27
upgrade controller-tools version from v0.2.5 to v0.3.0

camilamacedo86:
**Description**
upgrade controller-tools version from v0.2.5 to v0.3.0

**Motivation**
- avoid technical debts issues
- get bug fix and improvements to support CRD and Webhook api V1. 

PS.:  Has no breaking changes. See: https://github.com/kubernetes-sigs/controller-tools/releases

### DIFF
--- a/docs/book/install-and-build.sh
+++ b/docs/book/install-and-build.sh
@@ -16,6 +16,17 @@
 
 set -e
 
+# The following code is required to allow the preview works with an upper go version
+# More info : https://community.netlify.com/t/go-version-1-13/5680
+# Get the directory that this script file is in
+THIS_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+
+cd "$THIS_DIR"
+
+[[ -n "$(command -v gimme)" ]] && eval "$(gimme stable)"
+echo go version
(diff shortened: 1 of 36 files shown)


mengqiy:
After reading the log, it seems the netlify failed to build due to a too old golang was used.
It seems there is a way to use a newer golang version: https://community.netlify.com/t/go-version-1-13/5680
@camilamacedo86 can you try to fix this issue?

docs/book/install-and-build.sh
```diff
@@ -56,9 +67,10 @@ ${cmd} /tmp/mdbook.${ext}
 chmod +x /tmp/mdbook
 
 echo "grabbing the latest released controller-gen"
-go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.5
+go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0
 
 # make sure we add the go bin directory to our path
+echo "adding the go bin directory to our path"
```
estroz, line 73: Unrelated change.

camilamacedo86: required to fix the CI preview after the change and allow the merge
but you are right. I will split into another PR.

PR #1489, opened 2020-05-03
feat: remove update command which was useful just for v1

camilamacedo86:
**Description**

We no longer need to keep this command to help to migrate the projects to use go mod.  
It was useful just to v1 projects which were deprecated to long ago and is no longer supported. 

**Motivation**
Suggestions/comments raised when we remove the v1 code source.


camilamacedo86:
c/c @estroz

estroz:
@mengqiy @droot do you foresee this command being used in the future? Perhaps it would be better to remove the Gopkg scaffold, leave the command constructor, and only add this command to the CLI after checking `updateCmd.HasSubCommands()` so we can use `update` later.

camilamacedo86:
HI @estroz, 

Really tks for your input and help with.

> @mengqiy @droot do you foresee this command being used in the future? Perhaps it would be better to remove the Gopkg scaffold, leave the command constructor, and only add this command to the CLI after checking updateCmd.HasSubCommands() so we can use update later.

As we spoke, I do not think that we should keep the command at all. See that it is just useful for V1 projects and then, I am afraid that by following your suggestion we will keep a command that has no utility with the justification that we might need or not an updated cmd that works in another way in the future. So, IHMO it shows quite safe be removed. 

However, let's see if @mengqiy and/or @droot see any reason for we keep it.

estroz:
It would be trivial to add back, ex. `git checkout v2.3.0 -- cmd/update.go`, so we can get rid of it here.

PR #1490, opened 2020-05-03
remove the deprecated flags: depArgs and dep. Use fetch-deps instead of them.

camilamacedo86:
**Description**
remove the deprecated flags: depArgs and dep. Use fetch-deps instead of them. See:

```go
if err := fs.MarkDeprecated("dep", "use the fetch-deps flag instead"); err != nil {
	log.Printf("error to mark dep flag as deprecated: %v", err)
}
if err := fs.MarkDeprecated("depArgs", "will be removed with version 1 scaffolding"); err != nil {
	log.Printf("error to mark dep flag as deprecated: %v", err)
}
```

**Motivation**
comments/suggestions made when we remove the V1 source code

### DIFF
--- a/pkg/plugin/v2/init.go
+++ b/pkg/plugin/v2/init.go
@@ -18,7 +18,6 @@ package v2
 
 import (
 	"fmt"
-	"log"
 	"os"
 	"path/filepath"
 	"strings"
@@ -41,11 +40,6 @@ type initPlugin struct {
 	// boilerplate options
 	license string
 	owner   string
-
-	// deprecated flags
-	depFlag *pflag.Flag
-	depArgs []string
-	dep     bool
 
 	// flags
 	fetchDeps          bool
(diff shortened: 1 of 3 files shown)


camilamacedo86:
c/c @estroz

estroz:
@mengqiy can these flags be fully removed for both v2 and v3 projects or just v3?

camilamacedo86:
Hi @estroz, 

V1,V2,V3 are versions related to the project scaffolded layouts and not cli-tool. Since we will bump an upper version for the next release 3.0.0 then, it allows we remove from the tool the args. 

@mengqiy please let me know if I misunderstand something here.

estroz:
@camilamacedo86 fair point, although we can control what different project versions see in their CLI. I just wasn’t sure whether dep flags were deprecated alongside or independent of v1 deprecation.

Issue #1279, opened 2020-01-03
Docs cleanup


Adirio:
The book needs to be cleaned up.

- [ ] The files are not completely up-to-date and the tutorial should be followed step by step and checking the differences in the code as tracked in #1251 
- [ ] Is the folder `/docs/book/src/migration/testdata` being used for anything? It should be removed if it isn't. Why should we have a `testdata` folder that is the one used by the golden script inside one of the book chapters?
- [ ] Several files related to the old documentation like migration from v0, gifs, or some files that look like the utils used for previous versions of the documentation before moving to the new book format are still found in this directory

/kind cleanup

vaibhavk:
Hi, I would like to work on it.
/assign

Adirio:
@vaibhavk the first point should be pretty straightforward. Just follow the tutorial in a new project and doccument the differences in the file.

Second point requires to verify that the testdata directory is not used for anything, and remove it.

Third would require deep undestanding of the whole book creation process.

camilamacedo86:
Hi @Adirio, 

Following a few comments inline. 

> The files are not completely up-to-date and the tutorial should be followed step by step and checking the differences in the code as tracked in #1251

It shows a duplicate of #1251 so I think that can be removed from the scope of this issue. I mean the repo has another issue already to address the same. Also, the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/1492/files probably address it. 

> Is the folder /docs/book/src/migration/testdata being used for anything? It should be removed if it isn't. Why should we have a testdata folder that is the one used by the golden script inside one of the book chapters?

The `/docs/book/src/migration/testdata` is used just in the migration guide to help users move forward from V1 to V2. Since V1 was deprecated to long and has no longer supported I think we could remove the migration guide and the testdata dirs on it. WDYT? 

>  Several files related to the old documentation like migration from v0, gifs, or some files that look like the utils used for previous versions of the documentation before moving to the new book format are still found in this directory

I think the above would solve this one as well. WDYT?

camilamacedo86:
HI @mengqiy @droot, 

Do you think that we could remove the V1 to V2 migration guide as a solution to close this one?

Adirio:
I listed them to remember we need to do it but I don't have enough knowledge of the inner workins of how the documentation is generated to really know if those files are being used somewhere or they aren't.

camilamacedo86:
HI @Adirio, 

[...]

PR #1499, opened 2020-05-08
cleanup-book: remove docs/book/src/migration/testdata

camilamacedo86:
Closes: #1279

The mainly complain of #1279 were the outdated mock testdata in the book which was just used as an example over the final results of the migration from v1 o v2. 

In this way, since the V1 is deprecated and no longer for too long then, shows quite safe to remove the old mock data that was just added as an illustration.


camilamacedo86:
Hi @Adirio @mengqiy, 

Could you please help by checking this one? Wdyt? Are you ok with?

testdata/project-v3-addon/go.mod
```diff
@@ -6,8 +6,10 @@ require (
...
-	k8s.io/apimachinery v0.17.2
-	k8s.io/client-go v0.17.2
-	sigs.k8s.io/controller-runtime v0.5.0
-	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200415210853-85eb326a6add
+	gonum.org/v1/netlib v0.0.0-20190331212654-76723241ea4e // indirect
+	k8s.io/apimachinery v0.18.2
+	k8s.io/client-go v0.18.2
+	sigs.k8s.io/controller-runtime v0.6.0
+	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200507203943-d901af431e3a
+	sigs.k8s.io/structured-merge-diff v1.0.1-0.20191108220359-b1b620dd3f06 // indirect
```
camilamacedo86, line 14: Indirect deps which were updated with sigs.k8s.io/kubebuilder-declarative-pattern

Adirio:
/hold

See https://github.com/kubernetes-sigs/kubebuilder/issues/1279#issuecomment-627831376

TLDR: Now that the purpose of the testdata folder is clear, IMHO we should not remove it.

camilamacedo86:
HI @mengqiy @droot, 

IHMO it is quite safe we remove the V1 example from the books since:
- V1 code is deprecated, no longer support and removed
- Users are still able to follow up on the steps. The example do not add too much value at all.
- Also, since it uses dep and it is no longer a required at all the following error will be faced `

> Cannot run program "/Users/camilamacedo/go/bin/dep" (in directory "/Users/camilamacedo/go/src/sigs.k8s.io/kubebuilder/docs/book/src/migration/testdata/gopath/project-v1"): error=2, No such file or directory

WDYT?

camilamacedo86:
hi @pwittrock,

I saw that you have been helping with the reviews. really tks btw. 🥇 So, this one is a case that is ok to be merged but requires more opinions to reach a consensus. if you have time would be great get your input.

camilamacedo86:
Hi, @gabbifish @pwittrock @DirectXMan12 it is open for a while. WDYT? Are you ok with we removing the mock data of v1 since it is no longer supported a while?

camilamacedo86:
I am removing the 
/hold cancel

Otherwise will be hard to get reviews.

Issue #1384, opened 2020-02-21
[V3] Marker pattern


Adirio:
In v2, markers follow `// +kubebuilder:*` pattern.

In v3, we may want to remove the space (`//+kubebuilder:*`) as machine-readable comments should not have spaces by Go conventions.

Code to modify:

https://github.com/kubernetes-sigs/kubebuilder/blob/5dc6055cae706f0b80b05484ffe2632d2de56dd3/pkg/model/file/marker.go#L27-L29

/kind feature

camilamacedo86:
Hi @Adirio, 

I have thought more about this issue and I am not sure if we should do that. See that by default the common practice is to add comments as `// mycomment` with space. Also, note that it has the marker `# ` as well, so this change would not be consistent. 

also, could you add please add the link for we are able to check this Go convection?

Adirio:
/remove-lifecycle stale

@camilamacedo86 You can see in https://github.com/golang/go/issues/37974 how they are even including support in go doc to not document directives (which are machine-readable comments that don't start by a whitespace).

kensipe:
I initially had concerns that this would cause more work due to linters... the gocritic linter does check for spaces after the `//` however it allows for build tags to not have spaces.   
After some research... this makes sense and aligns with recent changes in go conventions
nice recommendation @Adirio !

kensipe:
@camilamacedo86 we can remove the `triage` label (I don't have rights and I don't think we need any more information)

Adirio:
I can take this one
/assign

PR #1505, opened 2020-05-08
WIP: fix: remove space from marker annotations

camilamacedo86:
Closes: #1384

TODO:
- Allow adding fragments when has the marker no matter if has or not an space between the comment // and marker.

### DIFF
--- a/pkg/model/file/marker.go
+++ b/pkg/model/file/marker.go
@@ -24,11 +24,8 @@ import (
 const prefix = "+kubebuilder:scaffold:"
 
 var commentsByExt = map[string]string{
-	// TODO(v3): machine-readable comments should not have spaces by Go convention. However,
-	//  this is a backwards incompatible change, and thus should be done for next project version.
-	".go":   "// ",
-	".yaml": "# ",
-	// When adding additional file extensions, update also the NewMarkerFor documentation and error
+	".go": "//", // Marker annotations for go files should not have spaces by go convention: //+kubebuilder:*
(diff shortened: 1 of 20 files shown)


camilamacedo86:
c/c @Adirio

pkg/model/file/marker.go
```diff
@@ -24,9 +24,9 @@ import (
 const prefix = "+kubebuilder:scaffold:"
 
 var commentsByExt = map[string]string{
-	// TODO(v3): machine-readable comments should not have spaces by Go convention. However,
-	//  this is a backwards incompatible change, and thus should be done for next project version.
-	".go":   "// ",
+	// it is used just to gen the marker annotations (//+)
+	".go": "//",
+	// It is used just to gen the marker commanets in the yaml files (# +)
 	".yaml": "# ",
```
Adirio, line 30: I would re-write the comments inline to keep 1 line per file type and including a longer example:
```suggestion
	".go":   "//", // Marker annotations for go files should not have spaces by go convention: //+kubebuilder:*
	".yaml": "# ", // Marker annotations for yaml files are more readable with spaces: # +kubebuilder:*
```

camilamacedo86:
HI @Adirio, 

Remove an space of the annotations which are scaffolded do not change the layout of the projects and is not a breaking change. Projects scaffolded before and after this change still working with the tool and then, it is not a breaking change at all.

camilamacedo86:
/hold
The problem here is that annotation without space will not found to add new fragments.
Users also could change it so we need to allow it works with space or not.

camilamacedo86:
I am closing this one. I think better about this change and I do not think that we should move forward with.

Issue #1500, opened 2020-05-08
Build is not working for multi-group


camilamacedo86:
The Dockerfile is scaffolded with 

<img width="438" alt="Screenshot 2020-05-08 at 02 25 07" src="https://user-images.githubusercontent.com/7708031/81360305-6712b580-90d3-11ea-8a3d-b97303572411.png">

And then, for the multigroup structure, we have apis/ instead of api/.

camilamacedo86:
c/c @estroz

PR #1507, opened 2020-05-08
fix: unable to build multigroup projects

camilamacedo86:
closes: #1500

PS.: I do not consider this bug fix as a braking change because:

After having project scaffold and with the project layout update to multigroup support then, users are still able to use the command to re-run the edit --multigroup=true which would apply the fix. 

ps.: shows that we have not too many users using the multigroup since the issue was not reported so far.

### DIFF
--- a/pkg/scaffold/edit.go
+++ b/pkg/scaffold/edit.go
@@ -17,6 +17,10 @@ limitations under the License.
 package scaffold
 
 import (
+	"fmt"
+	"io/ioutil"
+	"strings"
+
 	"sigs.k8s.io/kubebuilder/pkg/model/config"
 )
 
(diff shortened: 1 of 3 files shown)


camilamacedo86:
PS.: I will block the merge until we have 2 reviewers in this one. 
/hold

camilamacedo86:
Hi @estroz @mengqiy @droot  could you help by checking this one?

estroz:
Instead of breaking scraffolded directory structure, something kubebuilder should do as infrequently as possible, why not add a sentence in the multi group migration docs to update the Dockerfile?

camilamacedo86:
HI @estroz, 

Thank you for your input. However, I do no thinking that we should follow-up with your suggestion. 

1. Why change the api to apis when is multi-group? It has no relation with so we should not.
2. Add `add a sentence in the multi-group migration docs to update the Dockerfile?` it would be possible, however, may have other places that were affected and are facing issues too. Note that we did not get the issue before which shows that users are not really using it yet. And then, I do not think that request these unnecessary manually changes are a good user experience at all.

estroz:
My 2 cents:

1. A project fundamentally changes when it transitions to support multiple groups, which is probably why the directory structure changes and should change.
1. I have not seen any related issues aside from #1500. 

This change sounds more like a matter of preference than one that should happen for a good technical reason, since we can easily document the Dockerfile change like all other multigroup migration changes. Perhaps a better approach would be to write a phase 2 plugin to support this preference, like the one suggested in #932.

camilamacedo86:
HI @estroz, 

I am not sure if it is a matter of the preference since it is broken the project and do not let the users run the testadata gen with multigroup as suggested in the docs. When we added the multi-group option we decided add `if is multi-group the use apis/ instead of api, however IHMO it shows that was wrong decision.

camilamacedo86:
Hi @estroz, 

I will try to solve only the bug instead of keeping api/ for multigroup since you do not agree with.
I am closing this now .. and I will re-open when I have a new fix.

camilamacedo86:
Hi @estroz, 
 
After the discussion, the solution was changed to just provide the fix and add a test to ensure that all layouts are working building.

Issue #1473, opened 2020-04-15
leader-election-role role refers to non-existent configmaps/status resource


wallrj:
In https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/scaffold/internal/templates/v2/leaderelectionrole.go#L63 we noticed that the generated leader-election-role refers to a resource `configmaps/status`, but configmaps do not have a status.

See https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#configmap-v1-core

PR #1498, opened 2020-05-08
fix: controller alias for multigroups (v3)

camilamacedo86:
**Description**

 - Fix controller alias 

**Motivation**

Closes: #1382

PR #1511, opened 2020-05-09
fix: remove role for configmaps/status which do not exists from leader_election_role.yaml

camilamacedo86:
Closes: #1473

Theoretically, any deletion would be a breaking change. However, in this case, we are scaffolding something that does not exist and has no value at all. So, because of this was done for V2,V3. 

blocked;

TODO: Apply to V3 plugin 2.X only.
Waiting for #1498


camilamacedo86:
TODO: Apply to V3 plugin 2.X only. 
Waiting for https://github.com/kubernetes-sigs/kubebuilder/pull/1498

camilamacedo86:
/hold cancel
it is applied just for 3-alpha now.

Issue #1518, opened 2020-05-15
Add envtest example for CronJob controller


hasbro17:
As I was reading the docs for envtest I see the following:

> Until more documentation has been written, your best bet to get started is to look at some existing examples, such as:
> - Azure Databricks Operator: see their fully fleshed-out [suite_test.go](https://github.com/microsoft/azure-databricks-operator/blob/0f722a710fea06b86ecdccd9455336ca712bf775/controllers/suite_test.go) as well as any *_test.go file in that directory [like this one](https://github.com/microsoft/azure-databricks-operator/blob/0f722a710fea06b86ecdccd9455336ca712bf775/controllers/secretscope_controller_test.go).

I assume `more documentation` probably meant the following sub-section that was added later on which covers the setup and configuration of envtest but references no example test for users to look at.
https://book.kubebuilder.io/reference/testing/envtest.html

While the Azure Databricks Operator example seems sufficient I was wondering if it would be beneficial to have a envtest example for the [CronJob controller](https://github.com/kubernetes-sigs/kubebuilder/blob/f07a0146a40b1dc4048b1caaf075efcf1312aa00/docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller.go) itself so that we have a canonical example to reference within the repo.

Low priority but I wanted to see if there's interest in this before trying to write any tests for the CronJob controller.

/kind documentation

gabbifish:
I’ve been working on docs for a generalized approach to envtest in the context of writing tests for the CronJob example :) Planning on opening a PR for it tomorrow morning, and I’ll reference it here. (It will be my first kubebuilder project PR so I will especially appreciate feedback/guidance on it).

gabbifish:
Entrypoint docs for envtest, in the context of the cronjob tutorial, were put into the production kubebuilder book with PR https://github.com/kubernetes-sigs/kubebuilder/pull/1579. I think we can consider this issue closed :)

PR #1492, opened 2020-05-04
WIP: docs: update book mocks

camilamacedo86:
**Description**
- Update the mock examples for the docs based in the latest version scaffold files
- Perform a few fixes as well. 

**Motivation**
- https://github.com/kubernetes-sigs/kubebuilder/issues/1251
- V3 
Closes: #1544

Issue #1251, opened 2019-12-10
Review tutorial to ensure that it has not outdated code


camilamacedo86:
This task is tracked in order to follow up all tutorial steps and check if is all updated.
If something was changed then, we could update its steps.

awgreene:
I'll be trying out controller-gen later this week and will be using the tutorials as a starting point, happy to update the docs as I go.

/assign

awgreene:
Working on this now!

PR #1521, opened 2020-05-18
Add envtest testing docs to extend cronjob example

gabbifish:
This PR adds testing docs to extend the CronJob example and teach people general patterns for writing tests with envtest. It addresses #1518.


gabbifish:
I'm trying to sign the CLA but when I link my github account to my LF account, I keep seeing `There was an error processing your request.` :( I'm going to email the LF helpdesk to see what they can do.

camilamacedo86:
/ok-to-test

@gabbifish really thank you for your great contribution. Please, see that you need to assign the CLA as well for this contribution be able to be accepted.

gabbifish:
All right! My CLA should be signed now.

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,243 @@
...
+				(b) the number of attempts * timeout period exceed the provided interval value.
+
+				In the examples below, let `FAILURE bool = false` and `SUCCESS bool = true`.
+				timeout and interval are Go Duration values of your choosing.
+			*/
+			By("Cronjob status should initially have no active jobs")
+			cronjobLookupKey := types.NamespacedName{Name: CronjobName, Namespace: CronjobNamespace}
+			createdCronjob := &cronjobv1.CronJob{}
+
+			// We'll need to retry getting this newly creatd CronJob, given that creation may not immediately happen.
```
hasbro17, line 186: Nit
```suggestion
			// We'll need to retry getting this newly created CronJob, given that creation may not immediately happen.
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
```diff
@@ -0,0 +1,154 @@
...
+	Expect(err).NotTo(HaveOccurred())
+
+	// Step 4: Get that k8s client and use it for tests!
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
+	Expect(err).ToNot(HaveOccurred())
+	Expect(k8sClient).ToNot(BeNil())
+
+	/*
+		One thing that this autogenerated file is missing, however, is a way to actually start your controller.
+		The code above will set up a client that for interacting with your custom Kind,
```
hasbro17, line 93: Or `that is for`. 
```suggestion
		The code above will set up a client for interacting with your custom Kind,
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+	/*
+		Finally, we close the Gingko test suite's done channel when setup is completed.
+	*/
+
+	close(done)
+}, 60)
+
+/*
+Kubebuilder also generates also boilerplate functions for cleaning up envtest and actually running your test files in your controllers/ directory.
```
hasbro17, line 134:
```suggestion
Kubebuilder also generates boilerplate functions for cleaning up envtest and actually running your test files in your controllers/ directory.
```

docs/book/src/reference/writing-tests.md
```diff
@@ -1,19 +1,33 @@
...
+
+{{#literatego ../cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go}}
+
+This Status update example above demonstrates a general testing strategy for a custom Kind with downstream objects. By this point, you hopefully have learned the following methods for testing your controller behavior:
+
+* Setting up your controller to run on an envtest cluster
+* Writing stubs for creating test objects
+* Isolating changes to an object to test specific controller behavior.
+
+If you want to tinker with how your envtest cluster is configured, see section [Writing and Running Integration Tests](/reference/testing/envtest.md) as well as the [`envtest docs`](https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/envtest).
```
hasbro17, line 24: This is probably a bit subjective but I sort of liked how the current doc points me to the [Writing and Running Integration Tests](https://book.kubebuilder.io/reference/testing/envtest.html) so I can get a bit of an overview of envtest, and the configurations and caveats to consider before jumping into an example.

So I'm wondering if we want to move these lines to the top of the doc as part of the overview to better highlight this. Currently I have to scroll quite a bit through the example to learn more about envtest itself.

hasbro17:
Still going through the example itself but this is pretty good and a lot more comprehensive than what I originally had in mind.

docs/book/src/reference/writing-tests.md
```diff
@@ -1,19 +1,33 @@
...
+
+## Testing your Controller's Behavior
+
+{{#literatego ../cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go}}
+
+This Status update example above demonstrates a general testing strategy for a custom Kind with downstream objects. By this point, you hopefully have learned the following methods for testing your controller behavior:
+
+* Setting up your controller to run on an envtest cluster
+* Writing stubs for creating test objects
+* Isolating changes to an object to test specific controller behavior.
```
hasbro17, line 22: Nit: No periods for list items if they're not a complete sentence, like the two before.
```suggestion
* Isolating changes to an object to test specific controller behavior
```

docs/book/src/reference/writing-tests.md
gabbifish: I'm happy to move these lines to the top of the doc. I agree, a lot of users might want to have this envtest information up front.

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
```diff
@@ -0,0 +1,154 @@
...
+	batchv1 "tutorial.kubebuilder.io/project/api/v1"
+
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	// +kubebuilder:scaffold:imports
```
camilamacedo86, line 45: we need to fix the import orders: 


```go
	"path/filepath"
	"testing"

	. "github.com/onsi/ginkgo"
	. "github.com/onsi/gomega"
	"k8s.io/client-go/rest"
	"sigs.k8s.io/controller-runtime/pkg/envtest"
	logf "sigs.k8s.io/controller-runtime/pkg/log"
	"sigs.k8s.io/controller-runtime/pkg/log/zap"

	batchv1 "tutorial.kubebuilder.io/project/api/v1"
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
```diff
@@ -0,0 +1,154 @@
...
+)
+
```
camilamacedo86, line 47: The cron-job example is outdated. 
However, the suit_test.go is gen automatically. IMO should use sem file that is generated and just add the explanations. See: https://github.com/kubernetes-sigs/kubebuilder/blob/d6e2b995831ce26228f2303e9d1b3a55acd3d225/testdata/project-v3/controllers/suite_test.go

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
```diff
@@ -0,0 +1,154 @@
...
+	}).SetupWithManager(mgr)
+
+	Expect(err).ToNot(HaveOccurred())
+	go func() {
+		err = mgr.Start(ctrl.SetupSignalHandler())
+		Expect(err).ToNot(HaveOccurred())
+	}()
+
+	k8sClient = mgr.GetClient()
+	Expect(k8sClient).ToNot(BeNil())
```
camilamacedo86, line 124: Here is the before implementation for each test that should be executed. 
So, we need to add a test for each controller and then, the tests/checks in the controller.

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
```diff
@@ -0,0 +1,154 @@
...
+	// Step 5: Run your test controller
+	// +kubebuilder:scaffold:scheme
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+		Scheme: scheme.Scheme,
+	})
+	Expect(err).ToNot(HaveOccurred())
+	err = (&CronJobReconciler{
+		Client: mgr.GetClient(),
+		Log:    ctrl.Log.WithName("controllers").WithName("CronJob"),
+	}).SetupWithManager(mgr)
```
camilamacedo86, line 115: How it will work if we have more than 1 controller?

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,243 @@
...
+	cronjobv1 "tutorial.kubebuilder.io/project/api/v1"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	batchv1 "k8s.io/api/batch/v1"
+	batchv1beta1 "k8s.io/api/batch/v1beta1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
```
camilamacedo86, line 41: ^ requires to fix the import order

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,243 @@
...
+/*
+We still have a little more setup before we can write our tests.
+The first step to writing a simple integration test is to actually create an instance of CronJob you can run tests against.
+Note that to create a CronJob, you’ll need to create a stub CronJob struct that contains your CronJob’s specifications.
+We define a `testHelper_GetCronJobStub()` function, as well as stub creation functions for downstream objects like `testHelper_GetJobTemplateSpecStub()`,
+that can create a CronJob definition for us.
+
+Some k8s Kinds require values for certain spec fields, so you’ll need to make sure your stubs populate those required fields.
+*/
+func testHelper_GetCronJobStub(name string, namespace string) *cronjobv1.CronJob {
```
camilamacedo86, line 78: The funcs done here are not used in more than 1 place in the test. So, IMO would fit better indeed to make easier understand/explain if the code were inside of each check. Eg: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/envtest/webhook_test.go#L23

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+	Expect(err).ToNot(HaveOccurred())
+	go func() {
+		err = mgr.Start(ctrl.SetupSignalHandler())
+		Expect(err).ToNot(HaveOccurred())
+	}()
+
```
camilamacedo86, line 122: I think would be better we do here as the example: https://github.com/microsoft/azure-databricks-operator/blob/0f722a710fea06b86ecdccd9455336ca712bf775/controllers/suite_test.go#L103

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
gabbifish: Gotcha. Does this mean I would also have to re-write the example unit test to now support this new example instead of the cronjob one? Are there plans to roll over to using this new example in the docs any time soon?

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
gabbifish: I redid the formatting to match what I saw in the azure databricks example--please let me know if anything else is missing!

gabbifish:
Thank you for the great feedback, @camilamacedo86 and @hasbro17! I've incorporated most of it, and left a question above I would love some more clarification on. :)

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
camilamacedo86: Hi @gabbifish I created the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/1492 to try updated the project with the latest versions. Also, we have an issue to follow up all docs and try to update some outdated point found. See; https://github.com/kubernetes-sigs/kubebuilder/issues/1251

However, IMO it is fine to move forward here, we just need to ensure that this file is as generated by the master branch changes and with your changes on top. Is it make sense? Could I explain properly my suggestion?

docs/book/src/cronjob-tutorial/epilogue.md
```diff
@@ -4,16 +4,17 @@ By this point, we've got a pretty full-featured implementation of the
 CronJob controller, and have made use of most of the features of
 KubeBuilder.
 
+If you want to learn how to write integration tests for the CronJob kind
```
camilamacedo86, line 7: The commits need to be squashed. We can have just one for be able to merge.

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
gabbifish: ^ This makes sense! The `suite_test.go` code in this example is consistent with the autogenerated code provided by `kubebuilder create api`. I did move 
```go
func TestAPIs(t *testing.T) {
	RegisterFailHandler(Fail)

	RunSpecsWithDefaultAndCustomReporters(t,
		"Controller Suite",
		[]Reporter{envtest.NewlineReporter{}})
}
``` 
underneath the `Before` and `After` function calls because I thought this ordering might be easier for new users to read through, but am happy to put this function in its original position if you think that would be better!

docs/book/src/cronjob-tutorial/epilogue.md
gabbifish: Got it! Squashed commits :)

camilamacedo86:
The falling in the CI here has no relation with your changes. 
It is because a new module which is just used in the scaffolded projects had a new release. See; https://github.com/kubernetes-sigs/kubebuilder/pull/1527

gabbifish:
Gotcha! @camilamacedo86, I'll wait for #1527 I to get merged and then I'll rebase on master then. :) Is there any other feedback you had for my PR before we consider merging it?

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
```diff
@@ -0,0 +1,154 @@
...
+		err = k8sManager.Start(ctrl.SetupSignalHandler())
+		Expect(err).ToNot(HaveOccurred())
+	}()
+
+	k8sClient = k8sManager.GetClient()
+	Expect(k8sClient).ToNot(BeNil())
+
+	/*
+		Finally, we close the Gingko test suite's done channel when setup is completed.
+	*/
```
camilamacedo86, line 128: I think i is not required. 
```suggestion
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,217 @@
...
+We then set up consts we'll use in our tests, including test names for our objects as well as helper definitions
+like SUCCESS and FAILURE to make our tests more readable. The usage of timeout and interval durations will become evident soon.
+*/
+const (
+	CronjobName      string = "test-cronjob"
+	CronjobNamespace string = "test-cronjob-namespace"
+	JobName          string = "test-job"
+
+	SUCCESS bool = true
+	FAILURE bool = false
```
camilamacedo86, line 54: Actually it is not required at all.
The  best approach is to use `Expect(result.Passed).To(BeTrue())`

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,217 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+
+	cronjobv1 "tutorial.kubebuilder.io/project/api/v1"
+)
+
+/*
+We then set up consts we'll use in our tests, including test names for our objects as well as helper definitions
+like SUCCESS and FAILURE to make our tests more readable. The usage of timeout and interval durations will become evident soon.
+*/
```
camilamacedo86, line 47:
```suggestion
```
See that we should use the framework methods instead of it.

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+The first step to writing a simple integration test is to actually create an instance of CronJob you can run tests against.
+Note that to create a CronJob, you’ll need to create a stub CronJob struct that contains your CronJob’s specifications.
+
+To create a stub CronJob, we'll also need to create stubs of its required downstream objects--namely, its Job template spec and
+its Pod template spec.
+*/
+var _ = Describe("Run CronJob Controller", func() {
+	Context("Can create CronJob", func() {
+		It("Should create all sub resources successfully", func() {
```
camilamacedo86, line 80:
```suggestion
		It("Should create the CronJob resource with default values successfully", func() {
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,217 @@
...
+					APIVersion: "batch.tutorial.kubebuilder.io/v1",
+					Kind:       "CronJob",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      CronjobName,
+					Namespace: CronjobNamespace,
+				},
+				Spec: cronjobv1.CronJobSpec{
+					// For simplicity, we only fill out the required fields and fall back to the CronJob CRD's defaults wherever possible.
+					// Runs every minute!
```
camilamacedo86, line 95:
```suggestion
```
I do not think that it is required

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,217 @@
...
+The first step to writing a simple integration test is to actually create an instance of CronJob you can run tests against.
+Note that to create a CronJob, you’ll need to create a stub CronJob struct that contains your CronJob’s specifications.
+
+To create a stub CronJob, we'll also need to create stubs of its required downstream objects--namely, its Job template spec and
+its Pod template spec.
+*/
+var _ = Describe("Run CronJob Controller", func() {
+	Context("Can create CronJob", func() {
+		It("Should create all sub resources successfully", func() {
+			By("Create CronJob")
```
camilamacedo86, line 81:
```suggestion
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,217 @@
...
+			ctx := context.Background()
```
camilamacedo86, line 82: It shows not accurate for me. I understand that we should here create a CronJob CR and then, check that all resources that will manage by this CR will be created and in the expected state. See: https://github.com/microsoft/azure-databricks-operator/blob/0f722a710fea06b86ecdccd9455336ca712bf775/controllers/dbfsblock_controller_test.go#L76-L95

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+		`Eventually()` will repeatedly run the function provided as an argument every interval seconds until
+		(a) the function’s output matches what’s expected in the subsequent `Should()` call, or
+		(b) the number of attempts * timeout period exceed the provided interval value.
+
+		In the examples below, let `FAILURE bool = false` and `SUCCESS bool = true`.
+		timeout and interval are Go Duration values of your choosing.
+	*/
+	Context("Can perform status updates", func() {
+		It("CronJob Status.Active count should increase as Jobs are added", func() {
```
camilamacedo86, line 135: Should ... when ... occurs

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,217 @@
+/*
```
camilamacedo86, line 1: It might not be part of this PR, however, IMO would be nice we add a new block of tests in the Travis to verify the book testdata and then, in this call this test implemented here to be sure that all is fine.

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
```diff
@@ -0,0 +1,154 @@
...
+	cfg, err = testEnv.Start()
+	Expect(err).ToNot(HaveOccurred())
+	Expect(cfg).ToNot(BeNil())
+
+	// Step 3: If using new CRD, add its corresponding Kind to the client-go
+	// k8s scheme. After this, it will be available in your k8s clientset.
+	err = batchv1.AddToScheme(scheme.Scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	// Step 4: Get that k8s client and use it for tests!
```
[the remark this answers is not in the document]
gabbifish:
```suggestion
	/*
	A client will be created to be use in the tests.
	*/
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
camilamacedo86: Fell free to close this one .. all solved

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	// Step 2: Start envtest!
+	var err error
+	cfg, err = testEnv.Start()
+	Expect(err).ToNot(HaveOccurred())
+	Expect(cfg).ToNot(BeNil())
+
+	// Step 3: If using new CRD, add its corresponding Kind to the client-go
+	// k8s scheme. After this, it will be available in your k8s clientset.
```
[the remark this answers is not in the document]
gabbifish:
```suggestion
	/*
	The autogenerated test code will already add the CronJob Kind schema to the default client-go k8s scheme.
	This ensures that the CronJob API/Kind that will be used in our test controller.
	*/
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
```diff
@@ -0,0 +1,154 @@
...
+	logf.SetLogger(zap.LoggerTo(GinkgoWriter, true))
+
+	// Step 1: Configure envtest.
+	// Include CRDDirectoryPaths arg if you're testing a new CRD.
+	By("bootstrapping test environment")
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths: []string{filepath.Join("..", "config", "crd", "bases")},
+	}
+
+	// Step 2: Start envtest!
```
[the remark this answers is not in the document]
gabbifish:
```suggestion
	/*
	Then, we start the envtest cluster.
	*/
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+var cfg *rest.Config
+var k8sClient client.Client // You'll be using this client in your tests.
+var testEnv *envtest.Environment
+
+var _ = BeforeSuite(func(done Done) {
+	logf.SetLogger(zap.LoggerTo(GinkgoWriter, true))
+
+	// Step 1: Configure envtest.
+	// Include CRDDirectoryPaths arg if you're testing a new CRD.
```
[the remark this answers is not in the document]
gabbifish:
```suggestion
	/*
	First, the envtest cluster is configured to read CRDs from the CRD directory Kubebuilder scaffolds for you.
	*/
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
```diff
@@ -0,0 +1,154 @@
...
+	var err error
+	cfg, err = testEnv.Start()
+	Expect(err).ToNot(HaveOccurred())
+	Expect(cfg).ToNot(BeNil())
+
+	// Step 3: If using new CRD, add its corresponding Kind to the client-go
+	// k8s scheme. After this, it will be available in your k8s clientset.
+	err = batchv1.AddToScheme(scheme.Scheme)
+	Expect(err).NotTo(HaveOccurred())
+
```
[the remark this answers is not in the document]
gabbifish:
```suggestion
	/*
	After your schema is set up, you will notice the following marker.
	This annotation is what allows new schemas to be added here automatically when a new API is added to the project. The marker should be after the schemas and before the client.
	*/
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
```diff
@@ -71,18 +62,28 @@ var _ = BeforeSuite(func(done Done) {
...
+	/*
+	The autogenerated test code will already add the CronJob Kind schema to the default client-go k8s scheme.
+	This ensures that the CronJob API/Kind that will be used in our test controller.
+	*/
 	err = batchv1.AddToScheme(scheme.Scheme)
 	Expect(err).NotTo(HaveOccurred())
-
-	// Step 4: Get that k8s client and use it for tests!
+	/*
+	After the schemas, you will the following marker.
```
camilamacedo86, line 81:
```suggestion
	After the schemas, you will check the following marker.
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
```diff
@@ -71,18 +62,28 @@ var _ = BeforeSuite(func(done Done) {
...
+	This annotation is what allows new schemas to be added here automatically when a new API is added to the project.
```
camilamacedo86, line 82:
```suggestion
	This maker is what allows new schemas to be added here automatically when a new API is added to the project.
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
gabbifish: I'd be happy to take a look into this :)

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
gabbifish: I reformatted this test section to mirror https://github.com/microsoft/azure-databricks-operator/blob/0f722a710fea06b86ecdccd9455336ca712bf775/controllers/dbfsblock_controller_test.go#L76-L95 better--now, we only use one `Context()` and one `It()`, and the check for no active downstream objects immediately follows the creation of the test CronJob.

gabbifish:
@camilamacedo86 thank you for the thorough review! :D I agree, I'm keen to whittle this PR down so it's less verbose (I figured it would be easy to write too much at first rather than write too little).

I responded to about all of your feedback, minus one question above. I'll also take a look at what the travis logic later today.

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+	timeout  = time.Second * 10
+	interval = time.Millisecond * 250
+)
+
+/*
+We also expose the GVK for our CronJob version, which we will use to manage ownership of Jobs that belong to our test CronJob.
+One way you can set this GVK is by defining it as a global var like so:
+*/
+var (
+	Kind string                  = reflect.TypeOf(cronjobv1.CronJob{}).Name()
```
camilamacedo86, line 58: The only place that is using it is the next line .. so I am not sure if it would be required at all.

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+	GVK  schema.GroupVersionKind = cronjobv1.GroupVersion.WithKind(Kind)
```
camilamacedo86, line 59: Then, the GKV is just used 1 place too. So, IMO we should to keep the above implementation inside of the test.

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+)
+
```
camilamacedo86, line 61: ^
```suggestion
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,200 @@
+/*
+We can create the tests in a separate file, cronjob_controller_test.go.
+Keeping consistent with the Gingko usage in our autogenerated suite_test.go code, we’ll use Gingko to write our tests.
```
camilamacedo86, line 3: Is it really crucial info over how to do it?  I think we can remove it. To try to have less verbose. 
```suggestion
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,200 @@
+/*
+We can create the tests in a separate file, cronjob_controller_test.go.
```
camilamacedo86, line 2: Ideally, we should have 1 file to the test each controller/reconciler. So, I'd suggest we shape it with something as:
 
```suggestion
Ideally, we should have one `<kind>_conroller_test.go` for each controller scaffolded and called in the `test_suite.go`. So, let's write our example test for the CronJob controller (`cronjob_controller_test.go.`)    
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+// +kubebuilder:docs-gen:collapse=Apache License
+
+/*
+As usual, we start with necessary imports. We also define some utility variables.
```
camilamacedo86, line 23:
```suggestion
As usual, we start with the necessary imports. 
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+	batchv1 "k8s.io/api/batch/v1"
+	batchv1beta1 "k8s.io/api/batch/v1beta1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+
+	cronjobv1 "tutorial.kubebuilder.io/project/api/v1"
+)
+
```
camilamacedo86, line 43:
```suggestion
/*
Here, we will define some constants that we will use in this example. 
*/
```

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+We also expose the GVK for our CronJob version, which we will use to manage ownership of Jobs that belong to our test CronJob.
+One way you can set this GVK is by defining it as a global var like so:
+*/
+var (
+	Kind string                  = reflect.TypeOf(cronjobv1.CronJob{}).Name()
+	GVK  schema.GroupVersionKind = cronjobv1.GroupVersion.WithKind(Kind)
+)
+
+/*
+Now we set up the Gingko testing framework to run CronJob controller tests.
```
camilamacedo86, line 63:
```suggestion
```
IMO we already said many times that it is Gingko so in order to try to be less verbosity I'd suggest remove it.

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+The first step to writing a simple integration test is to actually create an instance of CronJob you can run tests against.
+Note that to create a CronJob, you’ll need to create a stub CronJob struct that contains your CronJob’s specifications.
+
+To create a stub CronJob, we'll also need to create stubs of its required downstream objects--namely, its Job template spec and
+its Pod template spec.
```
camilamacedo86, line 69: The text here is not very clear for me. Also, is it something that  is vital for the users get how it works or is something that we could just remove?

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
camilamacedo86: it could be after the Describe since will be specifically used in the description as well. Do not need to be global in the test/file. 

```go
var _ = Describe("....", func() {

       // const
       // var used  

	BeforeEach(func() {
		// Add here which is required to init before each func/test
	})
```

```

File: docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go (hunk 20 above)
Comment (camilamacedo86) [reply docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go:74]: It still a little confusing for me. IMO it should be something as: 


```
	BeforeEach(func() {
		// - Apply a new CronJobCR 
	})
```

**Should work successfully when a new CronJob CR is applied** 

- Check that the CronJob object was created successfully
- Check that the spec a or b is of the CronJob resource is == the value expected.  

**Should increase the spec X when Jobs are created** 
- create the Job
- check that the spec X was increased. 

I hope that  it helps you with

File: docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go (hunk 30 above)
Comment (gabbifish) [reply docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go:58]: `Kind` is only used in one place but I do like breaking `Kind` and `GVK` apart here--it makes the code more readable (rather than cramming this all into one line) and also maps nicely to the Kind and GVK concepts discussed in earlier documentation chapters like https://book.kubebuilder.io/cronjob-tutorial/gvks.html. What do you think?

File: docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go (hunk 31 above)
Comment (gabbifish) [reply docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go:59]: We can do this in the testing docs! Once testers start writing more tests that require setting owner references, `kind` and `gvk` will be used more places and users might like having these variables available as a global. But they can do that little refactor themselves :)

File: docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go (hunk 33 above)
Comment (gabbifish) [reply docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go:3]: I can remove this! It's not the most necessary, it's just additional context for the testing framework also used in this file.

File: docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go (hunk 38 above)
Comment (gabbifish) [reply docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go:69]: What I try to convey here is that to create a stubbed CronJob, you'll also have to create stubbed downstream objects like the `CronJobSpec` required by the CronJob's `Spec` field, the `JobSpec` required by the `CronJobSpec`, and the `PodSpecTempate` and `PodSpec` required by the `JobSpec`. I wanted to emphasize that a tester could not just create a `CronJob` object but would also have to be responsible for creating some downstream objects to initialize a CronJob object correctly.

I tried to clear up the phrasing a bit--let me know what you think!

File: docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go (hunk 20 above)
Comment (gabbifish) [reply docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go:65]: I honestly do prefer the preexisting format of the tests, because it saves readers from having to learn about `BeforeEach()` and the scopes it applies to, and also mirrors what we see in https://github.com/microsoft/azure-databricks-operator/blob/master/api/v1alpha1/dbfsblock_types_test.go.

I can add the extra check for `Check that the spec a or b is of the CronJob resource is == the value expected.` but am honestly a bigger fan of how the tests are formatted now and how they match the azure databricks example. What do you think?

**Comment by gabbifish** [comment]
Alright, responded to latest round of feedback. I am curious on your thoughts re: not using `BeforeEach()` given the additional overhead/confusion it may introduce to users.

**Comment by gabbifish** [comment]
Just rebased on master now that #1527 has been merged. Hopefully that'll take care of the travis CI issue.

**Comment by jdumars** [comment]
/lgtm 
thank you for taking this on Gabbi!

**Comment by gabbifish** [comment]
@pwittrock has also generously offered to give this a gander!

**Comment by pwittrock** [comment]
cc @DirectXMan12

File: docs/book/src/cronjob-tutorial/testdata/project/go.sum [hunk 39]
```diff
@@ -128,11 +128,13 @@ github.com/gogo/protobuf v1.1.1/go.mod h1:r8qH/GZQm5c6nD/R0oafs1akxWv10x8SbQlK7a
 github.com/gogo/protobuf v1.2.1/go.mod h1:hp+jE20tsWTFYpLwKvXlhS1hjn+gTNwPg2I6zVXpSg4=
 github.com/gogo/protobuf v1.2.2-0.20190723190241-65acae22fc9d h1:3PaI8p3seN09VjbTYC/QWlUZdZ1qS1zGjy7LH2Wt07I=
 github.com/gogo/protobuf v1.2.2-0.20190723190241-65acae22fc9d/go.mod h1:SlYgWuQ5SjCEi6WLHjHCa1yvBfUnHcTbrrZtXPKa29o=
+github.com/gogo/protobuf v1.3.1 h1:DqDEcV5aeaTmdFBePNpYsp3FlcVH/2ISVVM9Qf8PSls=
```
pwittrock, line 131: nit / non-blocking: consider running `go mod tidy` in this directory

pwittrock:
@gabbifish would you squash the commits?  (6->1)

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+				After creating this CronJob, let's check that the CronJob's Spec fields match what we passed in.
+				Note that, because the k8s apiserver may not have finished creating a CronJob after our `Create()` call from earlier, we will use Gomega’s Eventually() testing function instead of Expect() to give the apiserver an opportunity to finish creating our CronJob.
+
+				`Eventually()` will repeatedly run the function provided as an argument every interval seconds until
+				(a) the function’s output matches what’s expected in the subsequent `Should()` call, or
+				(b) the number of attempts * interval period exceed the provided timeout value.
+
+				In the examples below, timeout and interval are Go Duration values of our choosing.
+			*/
+			By("CronJob spec should match arguments passed in during creation")
```
DirectXMan12, line 109: these by messages are a bit strange -- they seem like they belong at the end of the assertions instead.

`By` messages should read like english specifications of a given step of the test, and should outline the test when all code is removed.  DItto with Context.  Think more like RSpec or something, and less like XUnit.  For instance:

- CronJob controller
  - when updating status
    - should increase active count when new jobs are created
      - by creating a new cronjob
      - by checking that it has zero active jobs
      - by creating a new job
      - by checking that the crojob has 1 active job

docs/book/src/reference/writing-tests.md
```diff
@@ -1,19 +1,33 @@
...
-existing examples, such as:
+The basic approach is that, in your generated `suite_test.go` file, you will use envtest to create a local Kubernetes API server, instantiate and run your controllers, and then write additional `*_test.go` files to test it using [Ginko](http://onsi.github.io/ginkgo).
+
+If you want to tinker with how your envtest cluster is configured, see section [Writing and Running Integration Tests](/reference/testing/envtest.md) as well as the [`envtest docs`](https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/envtest).
+
+## Test Environment Setup
+
+{{#literatego ../cronjob-tutorial/testdata/project/controllers/suite_test.go}}
+
+## Testing your Controller's Behavior
```
DirectXMan12, line 16: instead of putting this under reference, why don't we just make it a page under the cronjob tutorial?

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+			// We'll need to retry getting this newly created CronJob, given that creation may not immediately happen.
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, cronjobLookupKey, createdCronjob)
+				if err != nil {
+					return false
+				}
+				return true
+			}, timeout, interval).Should(BeTrue())
+			Expect(createdCronjob.Spec.Schedule).Should(Equal("1 * * * *"))
```
DirectXMan12, line 121: can you add a comment here or something explaining this?  I'm assuming it's a sanity test for when you have pruning turned on on the API server (CRD v1 or CRD v1beta1 w/ conversion), but it's worth pointing out.

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+			/*
+				Now that we've created a CronJob in our test cluster, the next step is to write a test that actually tests our CronJob controller’s behavior.
+				Let’s test the CronJob controller’s logic responsible for updating CronJob.Status.Active with actively running jobs.
+				We’ll verify that when a CronJob has a single active downstream Job, its CronJob.Status.Active field contains a reference to this Job.
+
+				First, we should get the test CronJob we created earlier, and verify that it currently does not have any active jobs.
+			*/
+			By("CronJob status should initially have no active Jobs")
+			Expect(len(createdCronjob.Status.Active)).Should(Equal(0))
```
DirectXMan12, line 130: This should be a `Consistently` -- we want to check that we're stable, not just that we're fine at some given moment.

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -0,0 +1,200 @@
...
+			controllerRef := metav1.NewControllerRef(createdCronjob, gvk)
+			testJob.SetOwnerReferences([]metav1.OwnerReference{*controllerRef})
+			Expect(k8sClient.Create(ctx, testJob)).Should(Succeed())
+			/*
+				Adding this Job to our test CronJob should trigger our controller’s reconciler logic.
+				After that, we can write a test that evaluates whether our controller eventually updates our CronJob’s Status field as expected!
+			*/
+			By("Active Jobs should be incremented to one")
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, cronjobLookupKey, createdCronjob)
```
DirectXMan12, line 177: IIRC, you can use assertions inside eventually blocks.  Plus, an assertion function may return a second argument as an error.  Returning a non-nil argument will fail the assertion.  Ergo, I'd do something like:

Eventually(func() ([]string, error) {
  if err := k8sClient.Get(ctx, cronJobLookupKey, createdCronJob); err != nil {
    return nil, err
  }
  var names []string
  for _, job := range createdCronJob.Status.Active {
    names = append(names, job.Name)
  }
  return names, nil
}, timeout, interval).Should(ConsistOf(JobName), "should list our active job %s in the active jobs list in status", JobName)

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
camilamacedo86: You can close this one now. IMO, if you address the @DirectXMan12 suggestions it would be enough.

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
gabbifish: This isn't as much a sanity test for pruning as it is a check to ensure that our CronJob spec `Schedule` field matches the parameters we passed in. (I don't think `Schedule` is supposed to ever get pruned?) Please let me know if I misunderstood something!

gabbifish:
Thank you for the helpful feedback, @DirectXMan12! I incorporated all of it (except for one follow-up question I asked above). I also put the `Writing Tests` page under the CronJob tutorial and hoisted the envtest docs up to section 4.9 in the Reference section. Let me know what you think :)

camilamacedo86:
PS.: For we are able to merge the commits need to be squashed

gabbifish:
rebased! now playing a game of whack-a-mole with the e2e tests 😂

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
DirectXMan12: I think i'm just not understanding the failure condition you're testing for.  The two failure modes I can imaging are conversion screwing up something during round-tripping (which I forgot about when leaving the comment here) or pruning removing the field.  Otherwise, "field passed in" == "field read" should be ensured by the apiserver.

gabbifish: This failure condition was to check for "field passed in" == "field read" (which is ensured by the APIserver). The idea for adding this came from some conversations @camilamacedo86 and I had earlier (see https://github.com/kubernetes-sigs/kubebuilder/pull/1521#discussion_r431373610). Given that the consistency of a created object and its input spec is ensured by/tested in the apiserver, I think it honestly makes sense to just delete this check--what do you think?

gabbifish: I went with the second option (add extra comment to clarify that this tests the conversion of the Schedule string).

camilamacedo86:
@pwittrock  we need fix this one. See that it is not showing the in the book and then the previous info also shows no longer appears.

PR #1524, opened 2020-05-24
:book: don't requeue when constructJobForCronJob fails

bhavin192:
- Change the return statement to match the comment above it

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller.go
@@ -496,7 +496,7 @@ func (r *CronJobReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	if err != nil {
 		log.Error(err, "unable to construct job from template")
 		// don't bother requeuing until we get a change to the spec
-		return scheduledResult, nil
+		return ctrl.Result{}, nil
 	}
 
 	// ...and create it on the cluster
--- a/testdata/project-v2-addon/go.mod
+++ b/testdata/project-v2-addon/go.mod
@@ -9,5 +9,5 @@ require (
 	k8s.io/apimachinery v0.18.2
 	k8s.io/client-go v0.18.2
 	sigs.k8s.io/controller-runtime v0.6.0
-	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200512162422-ce639cbf6d4c
+	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200522144838-848d48e5b073
 )
--- a/testdata/project-v3-addon/go.mod
+++ b/testdata/project-v3-addon/go.mod
@@ -9,5 +9,5 @@ require (
 	k8s.io/apimachinery v0.18.2
 	k8s.io/client-go v0.18.2
 	sigs.k8s.io/controller-runtime v0.6.0
-	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200512162422-ce639cbf6d4c
+	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200522144838-848d48e5b073
 )


docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller.go
```diff
@@ -496,7 +496,7 @@ func (r *CronJobReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	if err != nil {
 		log.Error(err, "unable to construct job from template")
 		// don't bother requeuing until we get a change to the spec
-		return scheduledResult, nil
+		return ctrl.Result{}, nil
```
camilamacedo86, line 499: See [in this example](https://github.com/kubernetes-sigs/kubebuilder/blob/cda1ddbc192c2712a0f9e6b3b1610ec2bff3726d/docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller.go#L403-L438) that is using the `scheduledResult` in all successful cases to configure that reconcile should be re-executed after X time. (like a cron)

(line 519 - process finished successfully)
```go
// we'll requeue once we see the running job, and update our status
return scheduledResult, nil
```

So, the `return scheduledResult, nil` matches with `don't bother requeuing until we get a chance to the spec` since besides it faced an error the reconcile will NOT be retrigged immediately in this case as can be checked, for example, in the next operation made:

```go
	// ...and create it on the cluster
	if err := r.Create(ctx, job); err != nil {
		log.Error(err, "unable to create Job for CronJob", "job", job)
		return ctrl.Result{}, err // -> because it is NOT nil the reconcile will be called imeditally  
	}
```


Please, let us know if it clarifies the example implementation and if you agree in close this one since the proposed change would break the expected behaviour here.

bhavin192: Right, since we are not returning an error, it won't trigger the reconcile immediately. 

I thought the change in spec refers to changes in`CronJob.Spec.JobTemplate`. My consideration was,  `constructJobForCronJob()` might fail due to issues with template from `CronJob.Spec.JobTemplate` or the `CronJob` object itself (setting the owner reference in `ctrl.SetControllerReference()`. 

If we are considering spec created by `constructJobForCronJob()` then returning `scheduledResult, nil` might create a different spec (change in the scheduledTimeAnnotation value only).  It will make sure that we retry at the next schedule time. Am I understanding it correctly?

camilamacedo86: I am not sure if I could understand properly your comment here. However, see:

**Reconcile()**
The reconcile function is responsible for synchronizing the resources and their specifications according to the business logic implemented on them. In this way, it works like a loop, and it does not stop until all conditionals match its implementation. The following is pseudo-code with an example that clarifies it.

```go
reconcile App {

   // Check if a Deployment for the app exists, if not create one
   // If has an error, then go to the beginning of the reconcile
   if err != nil {
       return reconcile.Result{}, err 
   } 
   
   // Check if a Service for the app exists, if not create one 
   // If has an error, then go to the beginning of the reconcile
   if err != nil {
       return reconcile.Result{}, err 
   }  

   // Looking for Database CR/CRD 
   // Check the Database Deployments Replicas size
   // If deployment.replicas size != cr.size, then update it
   // Then, go to the beginning of the reconcile
   if err != nil {
       return reconcile.Result{Requeue: true}, nil
   }  
   ...
   
   // If it is at the end of the loop, then:
   // All was done successfully and the reconcile can stop  
   return reconcile.Result{}, nil

}
```

The following are possible return options to restart the Reconcile:

- With the error: `return reconcile.Result{}, err`
- Without an error: `return reconcile.Result{Requeue: true}, nil`
- Therefore, to stop the Reconcile, use: `return reconcile.Result{}, nil`

In this scenario, the example is using `scheduledResult := ctrl.Result{RequeueAfter: nextRun.Sub(r.Now())} // save this so we can re-use it elsewhere` which means that it will re-started after X time.

So, in the example, if the error/scenario be faced it is showing that it should not continue and wait for the next reconciliation after X time or when the object is created/updated/edited if it has a watch implemented which also will re-trigger the reconcile as described in the comment `don't bother requeuing until we get a change to the spec`.  I hope that it clarifies and because of this, I do not think that the change for `return ctrl.Result{}, nil` will fit.

bhavin192: Sorry for the trouble but let me try to rephrase what I was trying to say,

I'm not sure which spec we are considering when we say `don't bother requeuing until we get a change to the spec.
My consideration was we are talking about the CronJob spec, if that is the case, then the reconcile should trigger when there in change in that spec (as we are watching for changes in the CronJob it will get triggered). 

I was considering a scenario in which the `constructJobFromCronJob()` fails because of some issue in the `CornJob` spec. I was looking at the function `ctrl.SetControllerReference()`, by seeing what it does, I came to the conclusion that this operation will fail only when the `CronJob` spec has some issues. And the CronJob spec will only change when user edits it.

I think that's the mistake here? Basically I should not go into to implementation of `SetControllerReference()` and treat it as an abstraction. That way the issue can be either with CronJob spec or the Job spec which we are constructing. And just retry for next time (the cron's schedule time).

camilamacedo86: The above line will set the OwnerRef in the resource.  The func is creating the resource and as part of it is setting the OwnerRef. Its explanation can be checked [here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controllerutil/controllerutil.go#L53-L58). So, I still not seen anything wrong with. 

Do you agree in close this PR? Is it ok for you? Also, please feel free to raise issues with questions as well in the repo. It is a nice way to get help with too. Btw, really tks for wants to contribute with and make it's better.

bhavin192: Yes, let me close this PR. I will ask in Slack channel or create issues if I have any questions. I should have just created an issue first and discuss all this there instead of directly creating a pull request.

camilamacedo86: It is fine .. please feel free to collab with and raise PR's with fixes, improvements and etc ..

Issue #1525, opened 2020-05-26
controller-gen is unable to work with addons projects. 


camilamacedo86:
## Bug

When addon projects, such as follows are scaffolded, the go mod is updated with the latest version/commit of module [kubernetes-sigs
/kubebuilder-declarative-pattern](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern). The module has not been working with releases/tags and kb is not scaffolding a specific version and then, after the commit/version `v0.0.0-20200512162422-ce639cbf6d4c` the make targets which are using the controller-gen has been kept stuck.

```
            export KUBEBUILDER_ENABLE_PLUGINS=1
            $kb create api --group crew --version v1 --kind Captain --controller=true --resource=true --pattern=addon
            $kb create api --group crew --version v1 --kind FirstMate --controller=true --resource=true --make=false --pattern=addon
            $kb create api --group crew --version v1 --kind Admiral --controller=true --resource=true --namespaced=false --make=false --pattern=addon
```

## Notes 
- We should fix the module version in the code to avoid breaking changes
- It is not the only issue scenario that we have been facing because of the version is not fixed in the go.mod. See that the CI can also break because of it. https://github.com/kubernetes-sigs/kubebuilder/issues/1503

## Suggested Solution
- For now, we can use replace to fix the problem and be unlocked. Note that it is booking all PR's since will break the CI. However, in the long term run shows that we could think if has any reason for we still keeping the `--pattern=addon` flag and using it.

camilamacedo86:
/bug

c/c @DirectXMan12 @mengqiy @estroz @hasbro17 @joelanford

camilamacedo86:
I am pushing the PR with the fix.

camilamacedo86:
I checked that stuck behaviour: 

- It was related to my local Go cache env. After cleaning it worked. 

However, it will break the CI. Any change in the repo will change the go.mod scaffolded by default and also can result in a breaking change.

Issue #1503, opened 2020-05-08
Add `make check-testdata` to the prow CI


camilamacedo86:
This task is for we add in the prow the test which checks if the testdata is updated. The reason for that is avoid broke the CI for the master branch. Following a scenario as an example. 

The mod `kubebuilder-declarative-pattern` is added when the testdata add-on project is gen, if it changes after the PR be pushed and the CI finished then, the boot will merge the PR in the master because it will just re-run the prow tests. However, the `make check-testdata` for the master branch will be executed after the merge and will fail. 

**Notes**
- Check the contribution.md to have further information and be able to add this test in the prow.

camilamacedo86:
c/c @mengqiy @alexeldeib

PR #1527, opened 2020-05-26
fix: fixed version of sigs.k8s.io/kubebuilder-declarative-pattern which should be used

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/1525 
Closes: #1503 (no longer required after it as well)

### DIFF
--- a/pkg/plugin/v2/api.go
+++ b/pkg/plugin/v2/api.go
@@ -182,6 +182,21 @@ func (p *createAPIPlugin) GetScaffolder() (scaffold.Scaffolder, error) {
 }
 
 func (p *createAPIPlugin) PostScaffold() error {
+	// Load the requested plugins
+	switch strings.ToLower(p.pattern) {
+	case "":
+		// Default pattern
+	case "addon":
+		// Ensure that we are pinning sigs.k8s.io/kubebuilder-declarative-pattern version
+		err := internal.RunCmd("Get controller runtime", "go", "get",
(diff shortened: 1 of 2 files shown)


mengqiy:
I don't quite understand why we need this?

camilamacedo86:
The addon requires the `kubernetes-sigs/kubebuilder-declarative-pattern`, however, its version is not defined/specified for the projects built with KB. In this way, KB will start to scaffold any new change introduced in the project which can be indeed a breaking change. Another common scenario, which is low important since do not affect the users, is that it makes the CI break. See the master now [here](https://travis-ci.org/github/kubernetes-sigs/kubebuilder/jobs/686290289#L893-L920). 

However, it has not changed so much and because of this, I think that it did not bring attention.  
So, fixed the version shows reasonable besides I do not like the fixed replace. Is it make sense?

camilamacedo86:
Hi @DirectXMan12, 

Yep. I agree. I will come up with a solution that will ensure a fixed version just for when we use the addon.

camilamacedo86:
Hi @DirectXMan12, 
Now it is solved in the correct way. 👍

camilamacedo86:
/hold 
Until @DirectXMan12 be able to check.

DirectXMan12:
I *think* this is fine for now.  We probably want to follow up with the addon folks to see what's going on there, and eventually turn this into a proper plugin with plugins stage two

/lgtm

camilamacedo86:
Tracked an issue for that: https://github.com/kubernetes-sigs/kubebuilder/issues/1543

PR #1536, opened 2020-05-30
[WIP] cli: change plugin name/version semantics

estroz:
Multiple plugins with the same name are now allowed by resolvePluginsByKey to permit a CLI to understand multiple plugin versions. For example, a config with "layout: go/v2.0" and another config with "layout: go/v2.1" should both be understood by a particular CLI.

* pkg/cli: remove validations requiring plugin names be unique within a CLI,
and resolve all matching plugins such that they can contain multiple
plugins with the same name. As long as there is only one plugin type,
ex. Init, per set of resolved plugins, no errors will occur.


pkg/cli/cli.go
```diff
@@ -351,12 +344,6 @@ func validatePlugins(plugins ...plugin.Base) error {
 					pluginName, projectVersion, err)
 			}
 		}
-		// Check for duplicate plugin names. Names outside of a version can
```
camilamacedo86, line 354: We still need to check for duplicates. We should not allow 2 plugins with the same name and version.

pkg/cli/api.go
```diff
@@ -61,7 +61,7 @@ func (c cli) bindCreateAPI(ctx plugin.Context, cmd *cobra.Command) {
 		if isGetter {
 			if getter != nil {
 				err := fmt.Errorf("duplicate API creation plugins for project version %q: %s, %s",
-					c.projectVersion, getter.Name(), p.Name())
+					c.projectVersion, plugin.KeyFor(getter), plugin.KeyFor(p))
```
camilamacedo86, line 64: requires to rebase with the master to get the fix

pkg/cli/cli_test.go
```diff
@@ -17,120 +17,142 @@ limitations under the License.
...
+				By("setting two plugins of the same name and version")
+				_, err = New(WithDefaultPlugins(pluginAV1), WithPlugins(pluginAV1, pluginAV1))
+				Expect(err).To(MatchError(`broken pre-set plugins: two plugins have the same key: "go.example.com/v1.0"`))
+			})
 		})
-	})
 
-	Describe("Check if has duplicate plugins", func() {
-		It("should work successfully when a plugin has many versions", func() {
-			plugins := makePluginsForKeys(
```
camilamacedo86, line 91: we should not allow duplicating plugins ( same name/version )

pkg/cli/plugins_test.go
```diff
@@ -0,0 +1,111 @@
...
+		Expect(makePluginKeySlice(resolvedPlugins...)).To(Equal([]string{"baz.example.com/v1.0"}))
+
+		By("Resolving foo/v2.0")
+		resolvedPlugins, err = resolvePluginsByKey(plugins, "foo/v2.0")
+		Expect(err).NotTo(HaveOccurred())
+		Expect(makePluginKeySlice(resolvedPlugins...)).To(Equal([]string{"foo.kubebuilder.io/v2.0"}))
+	})
+
+	It("should resolve keys with errors", func() {
+
```
camilamacedo86, line 59: IMO the tests should be :  

`should works successfully when <scenario>`
`should fail when <scenario>`

Since it makes clear what is expected and when.  could we shape these testes?

pkg/cli/cli_test.go
estroz: This is tested [here](https://github.com/kubernetes-sigs/kubebuilder/pull/1536/files#diff-4c87b5f848c90e8582b2bad56402342fR91).

pkg/cli/cli.go
```diff
@@ -218,22 +216,32 @@ func (c *cli) initialize() error {
 	// layout and --plugins values can be short (ex. "go/v2.0.0") or unversioned
```
camilamacedo86, line 216:
```suggestion
	// layout and --plugins values can be short (ex. "go/v2.0") or unversioned
```

pkg/cli/plugins_test.go
```diff
@@ -0,0 +1,114 @@
...
+package cli
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"sigs.k8s.io/kubebuilder/pkg/plugin"
+)
+
```
camilamacedo86, line 25: Coud we add here : 

```suggestion
func TestPlugin(t *testing.T) {
	RegisterFailHandler(Fail)
	RunSpecs(t, "Plugin test")
}
```

The reason is that it makes easier we work with since we can run just this file.

pkg/cli/plugins_test.go
```diff
@@ -0,0 +1,114 @@
...
+		By("resolving baz")
+		resolvedPlugins, err = resolvePluginsByKey(plugins, "baz")
+		Expect(err).NotTo(HaveOccurred())
+		Expect(makePluginKeySlice(resolvedPlugins...)).To(Equal([]string{"baz.example.com/v1.0"}))
+
+		By("resolving foo/v2.0")
+		resolvedPlugins, err = resolvePluginsByKey(plugins, "foo/v2.0")
+		Expect(err).NotTo(HaveOccurred())
+		Expect(makePluginKeySlice(resolvedPlugins...)).To(Equal([]string{"foo.kubebuilder.io/v2.0"}))
```
camilamacedo86, line 61: The following scenario is not working: 

```
	It("should return the correct version when just the name of the plugin is informed", func() {

		var (
			plugins = makePluginsForKeys(
				"foo.example.com/v1.0",
				"foo.example.com/v2.0",
			)
			resolvedPlugins []plugin.Base
			err             error
		)


		By("resolving foo.example.com")
		resolvedPlugins, err = resolvePluginsByKey(plugins, "foo.example.com")
		Expect(makePluginKeySlice(resolvedPlugins...)).To(Equal([]string{"foo.example.com/v2.0"}))

	})
```

should we not return the upper version in this case since we are allowing pass just the name?

pkg/cli/plugins_test.go
```diff
@@ -0,0 +1,114 @@
...
+			"foo.kubebuilder.io/v1.0",
+			"foo.kubebuilder.io/v2.0",
+			"bar.kubebuilder.io/v1.0",
+			"bar.kubebuilder.io/v2.0",
+		)
+		resolvedPlugins []plugin.Base
+		err             error
+	)
+
+	It("should resolve keys correctly", func() {
```
camilamacedo86, line 42: I think it would be easier to understand if we have 1 IT for scenario such as: 

```
It("should work when the plugin informed exist", func() {
		By("resolving foo.example.com/v1.0")
		resolvedPlugins, err = resolvePluginsByKey(plugins, "foo.example.com/v1.0")
		Expect(makePluginKeySlice(resolvedPlugins...)).To(Equal([]string{"foo.example.com/v1.0"}))
	})
```

pkg/cli/plugins_test.go
```diff
@@ -0,0 +1,114 @@
...
+		Expect(err).To(MatchError(errAmbiguousPlugin{
+			key: "foo/v1.0",
+			msg: `possible names: ["foo.example.com" "foo.kubebuilder.io"]`,
+		}))
+
+		By("resolving foo")
+		_, err = resolvePluginsByKey(plugins, "foo")
+		Expect(err).To(MatchError(errAmbiguousPlugin{
+			key: "foo",
+			msg: `possible keys: ["foo.example.com/v1.0" "foo.kubebuilder.io/v1.0" "foo.kubebuilder.io/v2.0"]`,
```
camilamacedo86, line 83: IMO this scenario make is better covered with the currently duplicated issue for plugins. 
I mean the plugin/v1.0.0 is duplicated in my POV is a more intuitive error for it.

pkg/cli/plugins_test.go
estroz: Using multiple `By()`'s within an `It()` is idiomatic, check out [the docs](https://onsi.github.io/ginkgo/).

pkg/cli/plugins_test.go
estroz: The plugins known to the `CLI` are not duplicated, i.e. there is no duplicate key in `plugins` here. The user-defined key `foo` is ambiguous because it can match multiple known plugins.

pkg/cli/plugins_test.go
estroz: This key would be ambiguous because there are multiple major versions for key `foo.example.com`. See [this case](https://github.com/kubernetes-sigs/kubebuilder/pull/1536/files#diff-228d4ad7b6960d220e6d20f73ee048c3R65).

pkg/cli/cli.go
estroz: We haven't yet removed patch versions yet, PR to come.

pkg/cli/plugins_test.go
camilamacedo86: Agreed that it would be a RFE

pkg/cli/plugins_test.go
camilamacedo86: I am not sure if here, it is so clear as the examples in the doc but ok.

PR #1537, opened 2020-05-30
plugins: abstract base Go plugin for easier version upgrades

estroz:
Because major and minor versions will be bumped often, plugin implementations should be abstracted so they're easier to update.

* pkg/plugin/internal/v0: `Init`, `CreateAPI`, and `CreateWebhook` plugins in the state at which scaffolds/CLI was when plugins were implemented. These "base" plugins can be selectively used by versioned plugins.
* pkg/plugin/v0: Go plugin implementation v0.0. Used only for project version 2.
* pkg/plugin/v2: Go plugin implementation v2.x. Used for project versions 3-alpha and beyond.


pkg/plugin/base/api.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package v2
+package base
```
camilamacedo86, line 17: IHMO we should keep V2. 
a) at some point we will be able to remove the V2 and all will be V3
b) `In the case of the `v1` and `v2` PROJECT version, a corresponding Plugin version is implied and constant -- `v1` implies the `go.kubebuilder.io/v1` Plugin, and similarly for `v2`.  This is for legacy purposes -- no such implication is made with the `v3` PROJECT version.`

pkg/plugin/validation.go
```diff
@@ -17,32 +17,62 @@ limitations under the License.
...
 import (
-	"errors"
 	"fmt"
 
 	"github.com/blang/semver"
 
 	"sigs.k8s.io/kubebuilder/pkg/internal/validation"
 )
 
+type errInvalidPluginVersion struct {
```
camilamacedo86, line 27: Add new plugin validations shows not be part of the scope `plugins: abstract base Go plugin for easier version upgrades `

estroz: See #1541.

pkg/plugin/base/api.go
estroz: Updated base -> v0.

cmd/main.go
```diff
@@ -21,15 +21,18 @@ import (
 
 	"sigs.k8s.io/kubebuilder/cmd/version"
 	"sigs.k8s.io/kubebuilder/pkg/cli"
+	pluginv0 "sigs.k8s.io/kubebuilder/pkg/plugin/v0"
 	pluginv2 "sigs.k8s.io/kubebuilder/pkg/plugin/v2"
```
camilamacedo86, line 25: IMO we should not have pluginv0 for the same reasons described in https://github.com/kubernetes-sigs/kubebuilder/pull/1537#discussion_r432801716

See :
- v1 -> Project layout v1
- v2 -> Project layout v2
- v3 -> Project layout v3 (where the plugins started to be in place )

Keep clear the concept adopted: In the case of the v1 and v2 PROJECT version, a corresponding Plugin version is implied and constant --v1 implies thego.kubebuilder.io/v1 Plugin, and similarly for v2. This is for legacy purposes -- no such implication is made with the v3 PROJECT version.

So, for me replace it for Vo will not make easier or more intuitive the understatement. 
Also, what is the problem that will be solved with?

estroz:
@camilamacedo86 package names reflect plugin version, not project version. `v0` is the state of scaffolding/CLI when plugins were created and is the "base" plugin that should be built on top of _and_ the plugin to be used for project version 2 which does not "know" about plugins (no `layout` key).

Is this package structure not easier to create new plugin versions with than the current structure?

DirectXMan12:
@estroz can you clarify versioning of the go plugin then a bit?  I had originally envisioned that `project: v1 ~ project: v3, layout: go/v1` and `project: v2 ~ project: v3, layout: go/v2` (hand-waving a bit here, but take `~` to mean "equivalent if that were possible")

estroz:
@DirectXMan12 that's how I see the pre-v3 relation as well. My goal here is to enable easy re-use of the frozen project version 2 plugin (`v2.0`) so that we don't have to duplicate the whole thing for something like #1498. The way I have packages set up might not be the right way to do so.

I also think we should consider not packaging plugins by project version, but given the above relation it might make some sense to continue doing so. We'd eventually have plugins `v(3+).Y` packaged in `pkg/plugin/v3`.

camilamacedo86:
Hi @estroz,

I understand what you are trying to achieve. See that in #1498 the `plugin/v2` are == `plugin/v3` however, could it not become an impedivie for customizations into the plugins? For example, with the idea proposed here how can we add a new plugin (SpecialPluginGetter) valid just for V3?   

Also, see:

- plugin/v2 - contains the plugin source code based in the v2 project scaffold
- plugin/v3 - contains the plugin source code valid for V3 projects (where we will do the bumps Plugin V3+N)

And then one day we will be able to remove v2, however, until there we need to make clear what is or not the code used by v2 projects since we cannot change it.

Also, see that we have other places in the project, for example, the tests which follow the current structure and concepts adopted already:

- tests/v2/ -> test V2 projects with v2 plugins
- tests/v3/ -> test V3 projects with v3+N plugins
 
Screenshot 2020-06-02 at 00 01 33

Regards ` My goal here is to enable easy re-use of the frozen project version 2 plugin (v2.0) so that we don't have to duplicate` 

As you said in another comment/pr the V2 plugins is a very special scenario. So, I am OK to "duplicate" it for V3 now since it will not be repeated after that and then, **the implementation will likely be different in the future.**  

So, I still think that we should not move forward with.

estroz:
We can punt on this for now, and re-address when we start changing things after #1498.

Issue #1526, opened 2020-05-26
Plugins are not supporting multi-project version


camilamacedo86:
**What did you do?**
Implement a MAJOR plugin version with breaking checkings supported ONLY for V3 Project layouts. See: https://github.com/kubernetes-sigs/kubebuilder/pull/1498

**What did you expect to see?**

- When I run the KB commands for V2 PROJECT version the upper plugin versions which is supported by V2 projects be used.  
- When I run the KB commands for V3 PROJECT version the upper plugin version which is supported by V3 projects be used. 

**What did you see instead? Under which circumstances?**
The error `duplicate initialization plugins for project version "3-alpha": go.kubebuilder.io, go.kubebuilder.io`.

**Possible Solution**

The error is in:

https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/api.go#L59-L69

Note that `c.resolvedPlugins` will return all supported plugins for the PROJECT version.
So, in the scenario where we have : 

```
// When has more than one supportable plugin. E.g:
// - go.kubebuilder.io/v2.0.0 which is supported by V2 and V3
// - go.kubebuilder.io/v3.0.0 which is supported by V3
```

And we are scaffolding a project for V3 it will return  `go.kubebuilder.io/v2.0.0` and `go.kubebuilder.io/v3.0.0 `. In this scenario, we should use the upper one which is `go.kubebuilder.io/v3.0.0` and the piece to get the upper version is not currently implemented.

camilamacedo86:
c/c @estroz

camilamacedo86:
/kind bug

HI, @estroz I updated the issue description because I notice that it was not clear enough. I hope that the scenario is better clarified.

camilamacedo86:
According to the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/1542/files @estroz decided to move in the opposite direction and do not allow multi-project support. It means that this scenario: 

// - go.kubebuilder.io/v2.0.0 which is supported by V2 and V3
// - go.kubebuilder.io/v3.0.0 which is supported by V3

Would raise an error. But the plugin design would still allow : 

// - go.kubebuilder.io/v2.0.0 which is supported by V2 and V3
// - go.kubebuilder.io/v3.0.0 which is supported by V4  

Closing it so. 

/close

PR #1538, opened 2020-05-30
fix: plugin resolution for multi-projects support

camilamacedo86:
Closes: #1526

Note that the problem was faced/found in the https://github.com/kubernetes-sigs/kubebuilder/pull/1498. **(the error is faced when we gen the testdata which is checked in the CI)** In this way, it is easy verified by checking that the same PR is working successfully with this fix applied.

### DIFF
--- a/pkg/cli/api.go
+++ b/pkg/cli/api.go
@@ -19,6 +19,7 @@ package cli // nolint:dupl
 import (
 	"fmt"
 
+	"github.com/blang/semver"
 	"github.com/spf13/cobra"
 
 	"sigs.k8s.io/kubebuilder/internal/config"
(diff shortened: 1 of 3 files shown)


camilamacedo86:
HI @estroz, 

> If I understand what you're doing here, I think these changes are conflating "use the latest plugin version that supports a given project version" with "allow multiple non-default plugins per project version". 

It will return always the latest plugin version which is supported by the PROJECT version. 

**Example**
We have:
- go.kubebuilder.io/v2.0.0 which is supported by V2 and V3
- go.kubebuilder.io/v3.0.0 which is supported by V3

- When I run the KB commands for V2 PROJECT version the upper plugin versions which are supported by V2 projects be used. In this case is: `go.kubebuilder.io/v2.0.0`
- When I run the KB commands for V3 PROJECT version the upper plugin version which is supported by V3 projects be used. In this case is: `go.kubebuilder.io/v3.0.0`

Then, if we implement a new version such as `go.kubebuilder.io/v3.1.0` then, it will be the default version returned/used by V3. 

I hope that it clarifies.

camilamacedo86:
Closing this one based in the comment: https://github.com/kubernetes-sigs/kubebuilder/issues/1526#issuecomment-637166156

Issue #1519, opened 2020-05-15
Add in the contribution guide purpose of versions 


camilamacedo86:
Currently, we have:
- the PROJECT version (v1,v2,v3)
- the release CLI version (1.0.0, 2.0.0)
- and now we will have the plugin's version (dir V2 and 2.0.0)

This task is for we clarify its purpose and how to work with them. This section should address the answer for questions such as:  When should I as a contributor bump a new plugin version or not?  When should I as contributor create a v3 plugin dir?  When should I as a contributor create a new project version/? Can an introduce a breaking change for a stable VX project version?

camilamacedo86:
Hi @DirectXMan12 @mengqiy @estroz, 

wdyt? Have you inputs that you think would be valuable to be added on it.

estroz:
Summary from kubebuilder/controller-runtime/controller-tools biweekly meeting today:
* Configuration (`PROJECT` file) version was used to inform on scaffolding version up to and including `version: "2"`. Now configuration version only informs on the configuration spec.
	* Only changes to this spec will merit a configuration version bump.
* Plugin version, ex. the semver in key `go.kubebuilder.io/v2.0.0`, informs on the scaffolding version was used to run `init`, `create api`, or `create webhook`. This version, in the form of a full plugin key, is recorded in the `layout` configuration field.
	* Changes to the set of scaffolded files will result in either a major or minor version bump in accordance with a plugin versioning spec (TBD).
	* Patch version will be dropped from versions since it doesn't conform to the concept a scaffolding change: either a scaffold change breaks previous versions or does not.
* `kubebuilder` binary version is unrelated to both configuration and plugin version.
	* When either a configuration or plugin version bumps major versions, a new binary release must be made.
	* More than one version for a given plugin will be allowed so users do not have to upgrade on each plugin version change. For example: `kubebuilder` binary version v3.0.0 will support projects with a `PROJECT` file containing one of the following `layout` values:
		* `go.kubebuilder.io/v2.0`
		* `go.kubebuilder.io/v2.1`
		* `go.kubebuilder.io/v3.0`
	* Migration guides will be written for changes warranting version bumps.

Some open questions:
* How do we determine if a plugin change warrants a major, a minor, or no version bump?
	* This needs to be codified in a versioning spec.
* The binary will be bumped to a new major version if there is a configuration version bump. Will a major binary version bump also occur if a plugin's major version is bumped?
	* This can get tricky if we have more than one plugin.
[...]

PR #1539, opened 2020-05-31
docs: add/update CLI plugin docs

estroz:
Plugin versions do not need to contain a patch version number, since patch-like changes should not require the user to change their config's layout value. This commit rewrites and adds some documentation on general plugin development and versioning.

See https://github.com/kubernetes-sigs/kubebuilder/issues/1519#issuecomment-632334210 for details.

Will follow up with patch version removal once #1536 is merged, which will allow both patched and non-patched version (`v2.0.0` vs `v2.0`) resolution without a breaking change.

/cc @camilamacedo86 @DirectXMan12 @joelanford

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,13 +35,13 @@ $ git clone git@github.com:<user>/kubebuilder.git $GOPATH/src/sigs.k8s.io/kubebu
 ```
 
 1. Ensure you activate module support before continue (`$ export GO111MODULE=on`)
-1. Run the command `make install` to create a bin with the source code 
+1. Run the command `make install` to create a bin with the source code
 
 **NOTE** In order to check the local environment run `make go-test`.
 
-## What to do before submitting a pull request 
+## What to do before submitting a pull request
 
-1. Run the script `make generate` to update/generate the mock data used in the e2e test in `$GOPATH/src/sigs.k8s.io/kubebuilder/testdata/` 
+1. Run the script `make generate` to update/generate the mock data used in the e2e test in `$GOPATH/src/sigs.k8s.io/kubebuilder/testdata/`
 
 **IMPORTANT:** The `make generate` is very helpful. By using it, you can check if good part of the commands still working successfully after the changes. Also, note that its usage is a pre-requirement to submit a PR.
 
@@ -60,8 +60,8 @@ Following the targets that can be used to test your changes locally.
 
 ## Where the CI Tests are configured
 
-1. See the [Travis](.travis.yml) file to check its tests and the scripts used on it. 
-1. Note that the prow tests used in the CI are configured in [kubernetes-sigs/kubebuilder/kubebuilder-presubmits.yaml](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/kubebuilder/kubebuilder-presubmits.yaml). 
+1. See the [Travis](.travis.yml) file to check its tests and the scripts used on it.
+1. Note that the prow tests used in the CI are configured in [kubernetes-sigs/kubebuilder/kubebuilder-presubmits.yaml](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/kubebuilder/kubebuilder-presubmits.yaml).
 1. Check that all scripts used by the CI are defined in the project.  
 
 ## How to contribute to docs
@@ -82,39 +82,49 @@ cherry-picked into `book-v2` branch.
 
 Check the CI job after to do the Pull Request and then, click on in the `Details` of `netlify/kubebuilder/deploy-preview`
 
-## Understanding the versions
+## Versioning
 
 |   Name	|   Example	|  Description |
 |---	|---	|---	|
-|  PROJECT version |  `v1`,`v2`,`v3` | As of the introduction of [Extensible CLI and Scaffolding Plugins](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-1.md), PROJECT version represents the layout of PROJECT file itself.  For `v1` and `v2` projects, there's extra meaning -- see below.  |
-|  Release version | `v2.2.0`, `v2.3.0`, `v2.3.1` | Tagged versions of the KubeBuilder project, representing changes to the source code in this repository. See the [releases](https://github.com/kubernetes-sigs/kubebuilder/releases) page. |
-|  Plugin Versions | `go.kubebuilder.io/v2.0.0` | Represents the version of an individual plugin, as well as the corresponding scaffolding that it generates. |
+|  KubeBuilder version | `v2.2.0`, `v2.3.0`, `v2.3.1` | Tagged versions of the KubeBuilder project, representing changes to the source code in this repository. See the [releases][kb-releases] page for binary releases. |
+|  Project version |  `"1"`, `"2"`, `"3-alpha"` | Project version defines the scheme of a `PROJECT` configuration file. This version is defined in a `PROJECT` file's `version`. |
+|  Plugin version | `v2`, `v3-alpha` | Represents the version of an individual plugin, as well as the corresponding scaffolding that it generates. This version is defined in a plugin key, ex. `go.kubebuilder.io/v2`. See the [design doc][cli-plugins-versioning] for more details. |
 
-Note that PROJECT version should only be bumped if a breaking change is introduced in the PROJECT file format itself.  Changes to the Go scaffolding or the KubeBuilder CLI *do not* affect the PROJECT version.
+### Incrementing versions
 
-Similarly, the introduction of a new major version (`(x+1).0.0`) of the Go plugin might only lead to a new minor (`x.(y+1).0`) release of KubeBuilder, since no breaking change is being made to the CLI itself.  It'd only be a breaking change to KubeBuilder if we remove support for an older version of the plugin.
+For more information on how KubeBuilder release versions work, see the [semver](https://semver.org/) documentation.
 
-For more information on how the release and plugin versions work, see the the [semver](https://semver.org/) documentation.
+Project versions should only be increased if a breaking change is introduced in the PROJECT file scheme itself. Changes to the Go scaffolding or the KubeBuilder CLI *do not* affect project version.
 
-**NOTE:** In the case of the `v1` and `v2` PROJECT version, a corresponding Plugin version is implied and constant -- `v1` implies the `go.kubebuilder.io/v1` Plugin, and similarly for `v2`.  This is for legacy purposes -- no such implication is made with the `v3` PROJECT version.
+Similarly, the introduction of a new plugin version might only lead to a new minor version release of KubeBuilder, since no breaking change is being made to the CLI itself. It'd only be a breaking change to KubeBuilder if we remove support for an older plugin version. See the plugins design doc [versioning section][cli-plugins-versioning]
+for more details on plugin versioning.
 
-## Introducing changes in the scaffold files
+**NOTE:** the scheme for project version `"2"` was defined before the concept of plugins was introduced, so plugin `go.kubebuilder.io/v2` is implicitly used for those project types. Schema for project versions `"3-alpha"` and beyond define a `layout` key that informs the plugin system of which plugin to use.
 
-Changes in the scaffolded files require a new Plugin version. If we delete or update a file that is scaffolded by default, it's a breaking change and requires a `MAJOR` Plugin version.  If we add a new file, it may not be a breaking change.
+## Introducing changes to plugins
 
-**More simply:** any change that will break the expected behaviour of a project built with the previous `MINOR` Plugin versions is a breaking change to that plugin. 
+Changes made to plugins only require a plugin version increase if and only if a change is made to a plugin
+that breaks projects scaffolded with the previous plugin version. Once a plugin version `vX` is stabilized (it doesn't
+have an "alpha" or "beta" suffix), a new plugin package should be created containing a new plugin with version
+`v(X+1)-alpha`. Typically this is done by (semantically) `cp -r pkg/plugin/vX pkg/plugin/v(X+1)` then updating
+version numbers and paths. All further breaking changes to the plugin should be made in this package; the `vX`
+plugin would then be frozen to breaking changes.
 
-**EXAMPLE:**
+### Example
 
-KubeBuilder Release version (`5.3.1`) scaffolds projects with the plugin version `3.2.1` by default.
+KubeBuilder scaffolds projects with plugin `go.kubebuilder.io/v2` by default. A `v3-alpha` version
+was created after `v2` stabilized.
 
-The changes introduced in our PR will not work well with the projects which were built with the plugin versions `3.0.0...3.5.1` without users taking manual steps to update their projects. Thus, our changes introduce a breaking change to the Go plugin, and require a `MAJOR` Plugin version bump.
+You create a feature that adds a new marker to the file `main.go` scaffolded by `init`
+that `create api` will use to update that file. The changes introduced in your feature
+would cause errors if used with projects built with plugins `go.kubebuilder.io/v2`
+without users manually updating their projects. Thus, your changes introduce a breaking change
+to plugin `go.kubebuilder.io`, and can only be merged into plugin version `v3-alpha`.
+This plugin's package should exist already, so a PR must be made against the
 
-In the PR, we should add a migration guide to the [Migrations](https://book.kubebuilder.io/migrations.html) section of the KubeBuilder book. It should detail the required steps that users should take to upgrade their projects from `go.kubebuilder.io/3.X.X` to the new `MAJOR` Plugin version `go.kubebuilder.io/4.0.0`.
-
-This also means we should introduce a new KubeBuilder minor version `5.4.0` when the project is released: since we've only added a new plugin version without removing the old one, this is considered a new feature to KubeBuilder, and not a breaking change.
-
-**IMPORTANT** Breaking changes cannot be made to PROJECT versions v1 and v2, and consequently plugin versions `go.kubebuilder.io/1` and `go.kubebuilder.io/2`.
+You must also add a migration guide to the [migrations](https://book.kubebuilder.io/migrations.html)
+section of the KubeBuilder book in your PR. It should detail the steps required
+for users to upgrade their projects from `v2` to `v3-alpha`.
 
 ## Community, discussion and support
 
@@ -127,10 +137,12 @@ You can reach the maintainers of this project at:
 ## Becoming a reviewer or approver
 
 Contributors may eventually become official reviewers or approvers in
-KubeBuilder and the related repositories.  See
+KubeBuilder and the related repositories. See
 [CONTRIBUTING-ROLES.md](docs/CONTRIBUTING-ROLES.md) for more information.
 
 ## Code of conduct
 
 Participation in the Kubernetes community is governed by the [Kubernetes Code of Conduct](code-of-conduct.md).
 
+[kb-releases]:https://github.com/kubernetes-sigs/kubebuilder/releases
+[cli-plugins-versioning]:docs/book/src/reference/cli-plugins.md
--- a/docs/book/src/reference/cli-plugins.md
+++ b/docs/book/src/reference/cli-plugins.md
@@ -1,0 +1,250 @@
+# CLI Plugins
+
+Kubebuilder CLI plugins wrap scaffolding and CLI features in conveniently packaged Go types that are executed by the
+`kubebuilder` binary, or any binary that imports them. More specifically, a plugin configures the execution of one
+of the following CLI commands:
+* `init`: project initialization.
+* `create api`: scaffold Kubernetes API definitions.
+* `create webhook`: scaffold Kubernetes webhooks.
+
+Plugins are identified by a key of the form `<name>/<version>`. There are two ways to specify a plugin to run:
+* Setting `kubebuilder init --plugins=<plugin key>`, which will initialize a project configured for plugin with key
+ `<plugin key>`.
+* A `layout: <plugin key>` in the scaffolded `PROJECT` configuration file. Commands (except for `init`, which scaffolds
+  this file) will look at this value before running to choose which plugin to run.
+
+By default, `<plugin key>` will be `go.kubebuilder.io/vX`, where `X` is some integer.
+
+## Plugin interfaces
+
+Each plugin is required to implement the [`Base`][plugin-base] interface.
+
+```go
+type Base interface {
+  // Version returns the plugin's version, which contains a positive integer
+  // and an optional "stage" string. The string representation of this version
+  // has format:
+  // (v)?[1-9][0-9]*(-(alpha|beta))?
+  Version() Version
+  // Name returns a DNS1123 label string defining the plugin type.
+  // For example, Kubebuilder's main plugin would return "go".
+  //
+  // Plugin names can be fully-qualified, and non-fully-qualified names are
+  // prepended to ".kubebuilder.io" to prevent conflicts.
+  Name() string
+  // SupportedProjectVersions lists all project configuration versions this
+  // plugin supports, ex. []string{"2", "3"}. The returned slice cannot be empty.
+  SupportedProjectVersions() []string
+}
+```
+
+### Plugin types
+
+On top of being a `Base`, a plugin should also implement the [`GenericSubcommand`][plugin-subc] interface so it can be
+run with a CLI:
+
+```go
+type GenericSubcommand interface {
+  // UpdateContext updates a PluginContext with command-specific help text,
+  // like description and examples. Can be a no-op if default help text is desired.
+  UpdateContext(*PluginContext)
+  // BindFlags binds the plugin's flags to the CLI. This allows each plugin to
+  // define its own command line flags for the kubebuilder subcommand.
+  BindFlags(*pflag.FlagSet)
+  // InjectConfig passes a config to a plugin. The plugin may modify the
+  // config. Initializing, loading, and saving the config is managed by the
+  // cli package.
+  InjectConfig(*config.Config)
+  // Run runs the subcommand.
+  Run() error
+}
+```
+
+The [plugin context][plugin-context] is optionally updated by `UpdateContext(ctx)` to set custom help text for the target
+command; this method can be a no-op, which will preserve the default help text set by the [cobra][cobra] command constructors.
+
+A plugin also implements one of the following interface pairs to declare its support for specific subcommands:
+
+```go
+// To implement the 'init' subcommand.
+type InitPluginGetter interface {
+  Base
+  GetInitPlugin() Init
+}
+
+type Init interface {
+  GenericSubcommand
+}
+
+// To implement the 'create api' subcommand.
+type CreateAPIPluginGetter interface {
+  Base
+  GetCreateAPIPlugin() CreateAPI
+}
+
+type CreateAPI interface {
+  GenericSubcommand
+}
+
+// To implement the 'create webhook' subcommand.
+type CreateWebhookPluginGetter interface {
+  Base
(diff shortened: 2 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -81,39 +81,58 @@ cherry-picked into `book-v2` branch.
...
-|  Release version | `v2.2.0`, `v2.3.0`, `v2.3.1` | Tagged versions of the KubeBuilder project, representing changes to the source code in this repository. See the [releases](https://github.com/kubernetes-sigs/kubebuilder/releases) page. |
-|  Plugin Versions | `go.kubebuilder.io/v2.0.0` | Represents the version of an individual plugin, as well as the corresponding scaffolding that it generates. |
+|  KubeBuilder version | `v2.2.0`, `v2.3.0`, `v2.3.1` | Tagged versions of the KubeBuilder project, representing changes to the source code in this repository. See the [releases] page for binary releases. |
+|  Project version |  `"1"`, `"2"`,`"3-alpha"` | Project version defines the scheme of a `PROJECT` configuration file. This version is defined in a `PROJECT` file's `version`. |
+|  Plugin version | `v2.0`, `v2.1` | Represents the version of an individual plugin, as well as the corresponding scaffolding that it generates. This version is defined in a plugin key, ex. `go.kubebuilder.io/v2.0`. See the [design doc][plugins-design-versioning] for more details. |
```
camilamacedo86, line 90: Currently, we have implemented v2.0.0. So, why change for v2.0?
Otherwise, we are unable to bum v2.0.1, for example, which represents a change that is NOT a breaking change and is just a bugfix. Was not it that you indeed suggested in the latest KB meeting?

CONTRIBUTING.md
```diff
@@ -81,39 +81,58 @@ cherry-picked into `book-v2` branch.
...
 
-Note that PROJECT version should only be bumped if a breaking change is introduced in the PROJECT file format itself.  Changes to the Go scaffolding or the KubeBuilder CLI *do not* affect the PROJECT version.
```
camilamacedo86, line 92: Why remove this information?

CONTRIBUTING.md
```diff
@@ -35,13 +35,13 @@ $ git clone git@github.com:<user>/kubebuilder.git $GOPATH/src/sigs.k8s.io/kubebu
 ```
 
 1. Ensure you activate module support before continue (`$ export GO111MODULE=on`)
-1. Run the command `make install` to create a bin with the source code 
+1. Run the command `make install` to create a bin with the source code
```
camilamacedo86, line 39: spaces not part of the scope but 👍

CONTRIBUTING.md
```diff
@@ -133,3 +152,5 @@ KubeBuilder and the related repositories.  See
 
 Participation in the Kubernetes community is governed by the [Kubernetes Code of Conduct](code-of-conduct.md).
 
+[releases]:ttps://github.com/kubernetes-sigs/kubebuilder/releases
+[plugins-design-versioning]:design/extensible-cli-and-scaffolding-plugins-phase-1.md
```
camilamacedo86, line 156: The links are broken.

CONTRIBUTING.md
```diff
@@ -81,39 +81,58 @@ cherry-picked into `book-v2` branch.
...
+Similarly, the introduction of a new major version (`v(X+1).Y`) of the Go plugin might only lead to a new minor (`vX.(Y+1).Z`) release of KubeBuilder, since no breaking change is being made to the CLI itself. It'd only be a breaking change to KubeBuilder if we remove support for an older version of the plugin.
 
-**NOTE:** In the case of the `v1` and `v2` PROJECT version, a corresponding Plugin version is implied and constant -- `v1` implies the `go.kubebuilder.io/v1` Plugin, and similarly for `v2`.  This is for legacy purposes -- no such implication is made with the `v3` PROJECT version.
+For more information on how the release and plugin versions work, see the [semver](https://semver.org/) documentation.
 
-## Introducing changes in the scaffold files
+**NOTE:** schema for project versions `"1"` and `"2"` were defined before the concept of plugins was introduced, so plugin `go.kubebuilder.io/v2.0` is implicitly used for those project types. Schema for project versions `"3-alpha"` and beyond define a `layout` key that informs the plugin system of which plugin to use.
```
camilamacedo86, line 100: `plugin `go.kubebuilder.io/v2.0` is implicitly used for those project types` -> It is not true since it ONLY represents the layout of the files scaffolded by default in V2 projects. V1 projects have a different layout and following the same analogy would be `go.kubebuilder.io/v1.0` besides its code source no longer exist in the project.

CONTRIBUTING.md
```diff
@@ -81,39 +81,58 @@ cherry-picked into `book-v2` branch.
...
 
-## Understanding the versions
+## Versioning
 
 |   Name	|   Example	|  Description |
 |---	|---	|---	|
-|  PROJECT version |  `v1`,`v2`,`v3` | As of the introduction of [Extensible CLI and Scaffolding Plugins](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-1.md), PROJECT version represents the layout of PROJECT file itself.  For `v1` and `v2` projects, there's extra meaning -- see below.  |
-|  Release version | `v2.2.0`, `v2.3.0`, `v2.3.1` | Tagged versions of the KubeBuilder project, representing changes to the source code in this repository. See the [releases](https://github.com/kubernetes-sigs/kubebuilder/releases) page. |
-|  Plugin Versions | `go.kubebuilder.io/v2.0.0` | Represents the version of an individual plugin, as well as the corresponding scaffolding that it generates. |
+|  KubeBuilder version | `v2.2.0`, `v2.3.0`, `v2.3.1` | Tagged versions of the KubeBuilder project, representing changes to the source code in this repository. See the [releases] page for binary releases. |
```
camilamacedo86, line 88: Missing the link

CONTRIBUTING.md
```diff
@@ -81,39 +81,58 @@ cherry-picked into `book-v2` branch.
...
 
-## Introducing changes in the scaffold files
+**NOTE:** schema for project versions `"1"` and `"2"` were defined before the concept of plugins was introduced, so plugin `go.kubebuilder.io/v2.0` is implicitly used for those project types. Schema for project versions `"3-alpha"` and beyond define a `layout` key that informs the plugin system of which plugin to use.
 
-Changes in the scaffolded files require a new Plugin version. If we delete or update a file that is scaffolded by default, it's a breaking change and requires a `MAJOR` Plugin version.  If we add a new file, it may not be a breaking change.
+## Introducing changes to plugins
 
-**More simply:** any change that will break the expected behaviour of a project built with the previous `MINOR` Plugin versions is a breaking change to that plugin. 
+Changes made to plugins may require a plugin version increase depending on whether
+they break the current version or not. See the plugins design doc [versioning section][plugins-design-versioning]
```
camilamacedo86, line 105: The above line brings unnecessary complexity  see: 

You replace a concise definition rule ` any change that will break the expected behaviour of a project built with the previous `MINOR` Plugin versions is a breaking change to that plugin. `

BY : 

`Changes made to plugins may require a plugin version increase depending on whether they break the current version or not` 

Which makes this guidance very subjective. Note that the doc/link do not define and/or clarifies what means a breaking change for the plugins too. I am ok with add the link but I prefer the previous definition which makes the rule very clear and objective.

CONTRIBUTING.md
```diff
@@ -81,39 +81,58 @@ cherry-picked into `book-v2` branch.
...
+for more details.
 
-**EXAMPLE:**
+Breaking changes require either a new Kubebuilder _minor_ or _major_ release.
+A new _major_ release is only necessary for changes that fundamentally alter
+all `kubebuilder` projects.
```
camilamacedo86, line 110: The above is very confusing for me. See: 

`Breaking changes require either a new Kubebuilder _minor_ or _major_ release`

If we are following the semantic version convention it shows make no sense. How a breaking change can require either? What does it means? 

`A new _major_ release is only necessary for changes that fundamentally alter
all `kubebuilder` projects.`

What do you mean with  `kubebuilder` projects.` here? What is a change that fundamentally alter all `kubebuilder` projects.?

CONTRIBUTING.md
```diff
@@ -81,39 +81,58 @@ cherry-picked into `book-v2` branch.
...
 
-KubeBuilder Release version (`5.3.1`) scaffolds projects with the plugin version `3.2.1` by default.
+Non-breaking changes do not require a release.
```
camilamacedo86, line 112:
```suggestion
Non-breaking changes do not require a release.
```
It makes no sense for me since the releases are not ONLY done when has breaking changes in the plugins Actually, has or not breaking changes in the plugins has no relation with the releases.

CONTRIBUTING.md
```diff
@@ -81,39 +81,58 @@ cherry-picked into `book-v2` branch.
...
 
-The changes introduced in our PR will not work well with the projects which were built with the plugin versions `3.0.0...3.5.1` without users taking manual steps to update their projects. Thus, our changes introduce a breaking change to the Go plugin, and require a `MAJOR` Plugin version bump.
+### Example
 
-In the PR, we should add a migration guide to the [Migrations](https://book.kubebuilder.io/migrations.html) section of the KubeBuilder book. It should detail the required steps that users should take to upgrade their projects from `go.kubebuilder.io/3.X.X` to the new `MAJOR` Plugin version `go.kubebuilder.io/4.0.0`.
+KubeBuilder release version `v2.3.1` scaffolds projects with plugin `go.kubebuilder.io/v2.1`
+by default. Previously, plugin `go.kubebuilder.io/v2.0` was used.
```
camilamacedo86, line 117: I am afraid that use the `v2.3.1` which is a valid release brings confusing since because of this it the example was using a version that was not released yet to illustrate how it works.

CONTRIBUTING.md
camilamacedo86, line 100: IMO we should keep X.X.X always  ( have the possibility of us X.X or X.X.X increase the complexity and can lead from wrong understanding such as that we should not bump the minor for bugfixes )

CONTRIBUTING.md
estroz: See [this conversation](https://github.com/kubernetes-sigs/kubebuilder/issues/1519#issuecomment-632334210), which discusses dropping minor version from plugin versions.

CONTRIBUTING.md
estroz: That is incorrect, they both use(d) `go.kubebuilder.io/v2.0.0`.

CONTRIBUTING.md
estroz: I didn't replace this sentence, I moved it to the plugins design doc.

CONTRIBUTING.md
estroz: >What means a change that fundamentally alter all kubebuilder projects.?

I agree this is a little nebulous and could use some refinement. Thoughts @DirectXMan12 @joelanford?

CONTRIBUTING.md
estroz: I think this should be "if you remove a plugin major version, you must release a kubebuilder major version".

CONTRIBUTING.md
estroz: Personally I don't think the versions we use in an example matter.

CONTRIBUTING.md
estroz: https://github.com/kubernetes-sigs/kubebuilder/pull/1539#discussion_r433429664

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
```diff
@@ -50,6 +51,38 @@ existing qualifier defined by the plugin. FQ names prevent conflicts between
...
+
+There are two types of non-breaking changes: those requiring a _minor_ version
+increase, and those that do not require a version be changed.
+
+Situations that require a _minor_ version bump:
+* Adding a new file that other commands do not assume exists.
+* Modifying a scaffold file in a way that does not affect other commands.
+
+Situations that do not require a version be changed:
+* Typo fixes.
```
camilamacedo86, line 83: it is not true. Typo fixes can be a breaking change. E.g 1498

CONTRIBUTING.md
camilamacedo86: The doc defines the proposal solution, but I am not sure if should be used as a reference to define what is not breeding cange and how to contribute with.  I would prefer to keep a simple and no verbose explanation here as before.

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
estroz: Is #1498 really just fixing a typo though?

CONTRIBUTING.md
camilamacedo86: The doc defines the proposal solution, but I am not sure if should be used as a reference to define what is not breaking cange and how to contribute with.  I would prefer to keep a simple and no verbose explanation here as before.

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
```diff
@@ -50,6 +51,38 @@ existing qualifier defined by the plugin. FQ names prevent conflicts between
...
+* Cosmetic changes that do not affect program behavior.
```
camilamacedo86, line 84: it is not always true. Typo fixes can be a breaking change. E.g 1498
So, I'd suggest just remove it since you detailed enough on top.
```suggestion
```

CONTRIBUTING.md
camilamacedo86: OK. I am ok with we move in this way.

CONTRIBUTING.md
estroz: Ah yes there was a `v1.0.0` plugin. We should just remove discussion of project version 1 to avoid confusion, since we don't talk about it anywhere else.

camilamacedo86: Agreed .. Remove V1 from here as well.

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
estroz: #1498 is not a typo fix, it changes scaffolding.

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
```diff
@@ -25,7 +25,8 @@ Each plugin would minimally be required to implement the `Plugin` interface.
 
 ```go
 type Plugin interface {
-    // Version returns the plugin's semantic version, ex. "v1.2.3".
+    // Version returns the plugin's version. This version, while being a
```
estroz, line 28: While it is nice to have an up-to-date design doc to reference, it is perhaps prudent to have a separate doc that details the current plugin spec. This would be beneficial for when phase 2 is implemented.

CONTRIBUTING.md
```diff
@@ -133,3 +149,5 @@ KubeBuilder and the related repositories.  See
 
 Participation in the Kubernetes community is governed by the [Kubernetes Code of Conduct](code-of-conduct.md).
 
+[kb-releases]:ttps://github.com/kubernetes-sigs/kubebuilder/releases
```
camilamacedo86, line 152:
```suggestion
[kb-releases]:https://github.com/kubernetes-sigs/kubebuilder/releases
```

CONTRIBUTING.md
```diff
@@ -81,39 +81,55 @@ cherry-picked into `book-v2` branch.
...
 
-## Introducing changes in the scaffold files
+**NOTE:** the scheme for project version `"2"` was defined before the concept of plugins was introduced, so plugin `go.kubebuilder.io/v2.0` is implicitly used for those project types. Schema for project versions `"3-alpha"` and beyond define a `layout` key that informs the plugin system of which plugin to use.
 
-Changes in the scaffolded files require a new Plugin version. If we delete or update a file that is scaffolded by default, it's a breaking change and requires a `MAJOR` Plugin version.  If we add a new file, it may not be a breaking change.
+## Introducing changes to plugins
 
-**More simply:** any change that will break the expected behaviour of a project built with the previous `MINOR` Plugin versions is a breaking change to that plugin. 
+Changes made to plugins may require a plugin version increase depending on whether
+they break the current version or not. See the plugins design doc [versioning section][plugins-design-versioning]
```
camilamacedo86, line 105:
```suggestion
Changes that will break the expected behaviour of a project built with the previous _minor_ Plugin versions is a breaking change to that plugin.  See the plugins design doc [versioning section][plugins-design-versioning]

```

CONTRIBUTING.md
camilamacedo86: The text `KubeBuilder release version `v2.3.1` scaffolds projects with plugin `go.kubebuilder.io/v2.1` shows a fact because the release and version exists and lead the impression that the v2.X.X would just release plugins v2.X when both are not true. So, I'd suggest:

CONTRIBUTING.md
```diff
@@ -81,39 +81,55 @@ cherry-picked into `book-v2` branch.
...
 
-KubeBuilder Release version (`5.3.1`) scaffolds projects with the plugin version `3.2.1` by default.
+### Example
 
-The changes introduced in our PR will not work well with the projects which were built with the plugin versions `3.0.0...3.5.1` without users taking manual steps to update their projects. Thus, our changes introduce a breaking change to the Go plugin, and require a `MAJOR` Plugin version bump.
+KubeBuilder release version `v2.3.1` scaffolds projects with plugin `go.kubebuilder.io/v2.1`
+by default. Previously, plugin `go.kubebuilder.io/v2.0` was used.
 
-In the PR, we should add a migration guide to the [Migrations](https://book.kubebuilder.io/migrations.html) section of the KubeBuilder book. It should detail the required steps that users should take to upgrade their projects from `go.kubebuilder.io/3.X.X` to the new `MAJOR` Plugin version `go.kubebuilder.io/4.0.0`.
+You make a PR that adds a new marker to the file `main.go` scaffolded by `init`
```
camilamacedo86, line 116:
```suggestion
Let's imagine that our PR adds a new marker to the file `main.go` scaffolded by `init`
```

CONTRIBUTING.md
```diff
@@ -81,39 +81,55 @@ cherry-picked into `book-v2` branch.
...
+that `create api` will use to update that file. The changes introduced in your PR
+will cause errors when used with projects built with plugins `go.kubebuilder.io/v2.0`
+and `go.kubebuilder.io/v2.1` without users manually updating their projects.
+Thus, your changes introduce a breaking change to plugin `go.kubebuilder.io`,
```
camilamacedo86, line 120:
```suggestion
Thus, our changes introduce a breaking change to plugin `go.kubebuilder.io`,
```

CONTRIBUTING.md
```diff
@@ -81,39 +81,55 @@ cherry-picked into `book-v2` branch.
...
+### Example
 
-The changes introduced in our PR will not work well with the projects which were built with the plugin versions `3.0.0...3.5.1` without users taking manual steps to update their projects. Thus, our changes introduce a breaking change to the Go plugin, and require a `MAJOR` Plugin version bump.
+KubeBuilder release version `v2.3.1` scaffolds projects with plugin `go.kubebuilder.io/v2.1`
+by default. Previously, plugin `go.kubebuilder.io/v2.0` was used.
 
-In the PR, we should add a migration guide to the [Migrations](https://book.kubebuilder.io/migrations.html) section of the KubeBuilder book. It should detail the required steps that users should take to upgrade their projects from `go.kubebuilder.io/3.X.X` to the new `MAJOR` Plugin version `go.kubebuilder.io/4.0.0`.
+You make a PR that adds a new marker to the file `main.go` scaffolded by `init`
+that `create api` will use to update that file. The changes introduced in your PR
+will cause errors when used with projects built with plugins `go.kubebuilder.io/v2.0`
```
camilamacedo86, line 118:
```suggestion
will cause errors when used with projects built with plugins `go.kubebuilder.io/v3.0`
```

CONTRIBUTING.md
```diff
@@ -81,39 +81,55 @@ cherry-picked into `book-v2` branch.
...
+and `go.kubebuilder.io/v2.1` without users manually updating their projects.
+Thus, your changes introduce a breaking change to plugin `go.kubebuilder.io`,
+and require a _major_ version increase to `v3.0` in the PR.
```
camilamacedo86, line 121:
```suggestion
Let's imagine that our PR adds a new marker to the file `main.go` scaffolded by `init`
that `create api` will use to update that file. The changes introduced in our PR
will cause errors when used with projects which were built with plugins `go.kubebuilder.io/v3.0`
and `go.kubebuilder.io/v3.1` without users manually updating their projects.
Thus, our changes introduce a breaking change to plugin `go.kubebuilder.io` and then,
the Plugin requires a _major_ version increase to `v4.0` in the PR.
```

CONTRIBUTING.md
```diff
@@ -81,39 +81,55 @@ cherry-picked into `book-v2` branch.
...
 
-This also means we should introduce a new KubeBuilder minor version `5.4.0` when the project is released: since we've only added a new plugin version without removing the old one, this is considered a new feature to KubeBuilder, and not a breaking change.
+You must also add a migration guide to the [migrations](https://book.kubebuilder.io/migrations.html)
+section of the KubeBuilder book in your PR. It should detail the steps required
+for users to upgrade their projects from `go.kubebuilder.io/v2.Y` to the new
+plugin version `v3.0`.
```
camilamacedo86, line 126:
```suggestion
We must also add a migration guide to the [migrations](https://book.kubebuilder.io/migrations.html)
section of the KubeBuilder book in your PR. It should detail the steps required
for users to upgrade their projects from `go.kubebuilder.io/v3.Y` to the new
plugin version `v4.0`.
```

CONTRIBUTING.md
```diff
@@ -81,39 +81,55 @@ cherry-picked into `book-v2` branch.
...
 
-**IMPORTANT** Breaking changes cannot be made to PROJECT versions v1 and v2, and consequently plugin versions `go.kubebuilder.io/1` and `go.kubebuilder.io/2`.
+You've only added a new plugin version without removing old versions, so this
```
camilamacedo86, line 128:
```suggestion
Note that, we've only added a new plugin version without removing old versions, so this
```

CONTRIBUTING.md
```diff
@@ -81,39 +81,55 @@ cherry-picked into `book-v2` branch.
...
+considered a new feature to KubeBuilder and not a breaking change. Therefore
+a new _minor_ KubeBuilder version `v2.4.0` should be release.
+
```
camilamacedo86, line 131: The explanation here lead us to believe that after each change we need to do a release, which shows not true at all.
```suggestion
considered a new feature to KubeBuilder and not a breaking change. Therefore
if we decide to release a new version of KubeBuilder then, a new _minor_ KubeBuilder version with the tag `v5.4.0` should be done.
```

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
camilamacedo86: I agree. It could be in the Reference section in the book/docs. But it shows out of the scope of this PR. We can track a new issue to do just it.

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
camilamacedo86: If the typo fix is not in the scaffold files then we will not change the plugin at all.

docs/book/src/reference/cli-plugins.md
```diff
@@ -1,16 +1,262 @@
...
+There are two types of non-breaking changes: those requiring a _minor_ version increase, and those that do not
+require a version be changed.
+
+Situations that require a _minor_ version bump:
+* Adding a new file that other commands do not assume exists.
+* Modifying a scaffold file in a way that does not affect other commands.
+
+Situations that do not require a version be changed:
+* Typo fixes.
+* Cosmetic changes that do not affect program behavior.
```
camilamacedo86, line 253: I am not just ok with this ^. Both examples may be or not breaking changes. So, I would suggest just remove it.

docs/book/src/reference/cli-plugins.md
```diff
@@ -0,0 +1,262 @@
+# CLI Plugins
```
camilamacedo86, line 1: missing add the menu. I am unable to find it here: https://deploy-preview-1539--kubebuilder.netlify.app/

docs/book/src/reference/cli-plugins.md
estroz: We need to define when a version does not need to be changed. Do you have alternative suggestions?

camilamacedo86: So, I understand that; 

- Any change which will result in a scaffold change made by the plugins requires a new mInor/major plugin bump
- Any change that affects the behaviour of the plugin also shows requires a bump right?

However, if we do a change that has no relation with the code scaffolded by the plugin or its behaviours do not require a plugin bump. E.g 
- in the edit/autocomplete commands that are not using a plugin
- or in the books/ci/ or indeed in the tests.

estroz: Those things do not relate to plugins though. How does a typo fix cause any issue when scaffolding? Once again, #1498 is not a good example because that isn't a typo _and_ changes program behavior.

camilamacedo86: I think I got what you mean here .You can close this as solved

CONTRIBUTING.md
estroz: Should be fixed.

CONTRIBUTING.md
estroz: We have a concrete definition of breaking plugin changes now, and no discussion of releasing kubebuilder in relation to plugins.

CONTRIBUTING.md
estroz: Removed kubebuilder version from this section.

CONTRIBUTING.md
```diff
@@ -81,39 +81,49 @@ cherry-picked into `book-v2` branch.
...
 
-## Understanding the versions
+## Versioning
 
 |   Name	|   Example	|  Description |
 |---	|---	|---	|
-|  PROJECT version |  `v1`,`v2`,`v3` | As of the introduction of [Extensible CLI and Scaffolding Plugins](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-1.md), PROJECT version represents the layout of PROJECT file itself.  For `v1` and `v2` projects, there's extra meaning -- see below.  |
-|  Release version | `v2.2.0`, `v2.3.0`, `v2.3.1` | Tagged versions of the KubeBuilder project, representing changes to the source code in this repository. See the [releases](https://github.com/kubernetes-sigs/kubebuilder/releases) page. |
-|  Plugin Versions | `go.kubebuilder.io/v2.0.0` | Represents the version of an individual plugin, as well as the corresponding scaffolding that it generates. |
+|  KubeBuilder version | `v2.2.0`, `v2.3.0`, `v2.3.1` | Tagged versions of the KubeBuilder project, representing changes to the source code in this repository. See the [releases][kb-releases] page for binary releases. |
```
camilamacedo86, line 88: /hold

Until we are able to update it related to the latest changes.

estroz:
@camilamacedo86 @DirectXMan12 these docs changes reflect the most recent plugin spec changes and should be ok to merge.

camilamacedo86:
HI @estroz this content was moved to VERSIONING.md. Could you please update your PR for we are able to get it merged?

camilamacedo86:
Hi @estroz, 

Since it is open for too long without interaction, in order to us be able to close this issue  #1519 I pushed your changes in the VERSIONING.md doc which is the place where it is now via the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/1769. Would be very nice to get your help to review on that.  

In this way, I hope that is fine we close this one. However, feel free to re-open if you see that is required.

PR #1540, opened 2020-05-31
pkg/cli: use cmd error idiom for 'init' and plugin keys in cmd errors

estroz:
Instead of only using plugin names in command errors, keys should be used for more informative error messages. An action should be suggested in the error message as well.

Additionally, the `init` command setup should use the `cmdErr` idiom used by other command setups for uniform error reporting.

/cc @camilamacedo86 @joelanford @mengqiy

### DIFF
--- a/pkg/cli/api.go
+++ b/pkg/cli/api.go
@@ -60,8 +60,8 @@ func (c cli) bindCreateAPI(ctx plugin.Context, cmd *cobra.Command) {
 		tmpGetter, isGetter := p.(plugin.CreateAPIPluginGetter)
 		if isGetter {
 			if getter != nil {
-				err := fmt.Errorf("duplicate API creation plugins for project version %q: %s, %s",
-					c.projectVersion, getter.Name(), p.Name())
+				err := fmt.Errorf("duplicate API creation plugins for project version %q (%s, %s), "+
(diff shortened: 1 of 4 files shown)


pkg/cli/api.go
```diff
@@ -60,8 +60,8 @@ func (c cli) bindCreateAPI(ctx plugin.Context, cmd *cobra.Command) {
 		tmpGetter, isGetter := p.(plugin.CreateAPIPluginGetter)
 		if isGetter {
 			if getter != nil {
-				err := fmt.Errorf("duplicate API creation plugins for project version %q: %s, %s",
```
camilamacedo86, line 63: IMO this error should not occur at all. Note that `c.resolvedPlugins` will return all supported plugins for the PROJECT version which can be, for example: `go.kubebuilder.io/v3.0.0, go.kubebuilder.io/v3.1.0, go.kubebuilder.io/v3.2.0`

Then, we should use the upper version in this scenario as described in https://github.com/kubernetes-sigs/kubebuilder/issues/1526.

Note that this problem has no longer been faced because KB has just 1 plugin version at the moment which is implemented.

estroz:
@camilamacedo86 which error are you talking about, the duplicate error or not-found error? The duplicate error will happen if a user specifies an ambiguous key that doesn't correspond to a default plugin. The not-found error will happen if a user specifies a key that doesn't correspond to an existing plugin.

PR #1541, opened 2020-05-31
pkg/plugin/validation.go: wrap errors in types

estroz:
Plugin validation errors can be abstracted into types.

/cc @camilamacedo86

### DIFF
--- a/pkg/plugin/validation.go
+++ b/pkg/plugin/validation.go
@@ -17,7 +17,6 @@ limitations under the License.
 package plugin
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/blang/semver"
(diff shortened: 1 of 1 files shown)


pkg/plugin/validation.go
```diff
@@ -17,32 +17,64 @@ limitations under the License.
...
 	// ex. "3" or "v3.0".
-	if _, err := semver.ParseTolerant(version); err != nil {
-		return fmt.Errorf("failed to validate plugin version %q: %v", version, err)
+	v, err := semver.ParseTolerant(version)
+	if err != nil {
+		return errInvalidVersion{version, err.Error()}
+	}
+	cv := semver.Version{Major: v.Major, Minor: v.Minor}
+	if !v.Equals(cv) {
+		return errInvalidVersion{version, "must contain major and minor version only"}
```
camilamacedo86, line 53: The plugin version in the master is `v2.0.0` so should it not be failing in the CI to run the script which generates the testdata?

pkg/plugin/validation.go
```diff
@@ -17,32 +17,64 @@ limitations under the License.
...
-	"errors"
 	"fmt"
 
 	"github.com/blang/semver"
 
 	"sigs.k8s.io/kubebuilder/pkg/internal/validation"
 )
 
+// errInvalidVersion should be returned when a plugin's version is invalid.
+type errInvalidVersion struct {
```
camilamacedo86, line 28: Since we are implementing specific types of error would be good to export them to be used. Also, wdyt about simplified it by: 
```suggestion
type ErrInvalidVersion struct {
```

pkg/plugin/validation.go
```diff
@@ -17,32 +17,64 @@ limitations under the License.
...
+	cv := semver.Version{Major: v.Major, Minor: v.Minor}
+	if !v.Equals(cv) {
+		return errInvalidVersion{version, "must contain major and minor version only"}
 	}
+
 	return nil
 }
 
+// errInvalidName should be returned when a plugin's name is invalid.
+type errInvalidName struct {
```
camilamacedo86, line 60:
```suggestion
type ErrInvalidName struct {
```

pkg/plugin/validation.go
```diff
@@ -17,32 +17,64 @@ limitations under the License.
...
 	"github.com/blang/semver"
 
 	"sigs.k8s.io/kubebuilder/pkg/internal/validation"
 )
 
+// errInvalidVersion should be returned when a plugin's version is invalid.
+type errInvalidVersion struct {
+	version, msg string
+}
+
```
camilamacedo86, line 31: also, I'd suggest keep the struct definition on top and the methods on the bottom.

pkg/plugin/validation.go
```diff
@@ -17,32 +17,64 @@ limitations under the License.
...
 func ValidateName(name string) error {
+	if name == "" {
+		return errInvalidName{}
+	}
 	if errs := validation.IsDNS1123Subdomain(name); len(errs) != 0 {
-		return fmt.Errorf("plugin name %q is invalid: %v", name, errs)
+		return errInvalidName{name, fmt.Sprintf("%s", errs)}
 	}
 	return nil
 }
```
camilamacedo86, line 80: I think we need to cover it with unit test as well. (validation_test.go)

pkg/plugin/validation.go
```diff
@@ -17,32 +17,64 @@ limitations under the License.
 package plugin
```
camilamacedo86, line 18: the name of the file is validation, so should we not centralize here all plugin validations? 
Or your idea is just to have the plugin version validations? If yes, could we rename the file to make it clear or centralize all plugin validations?

pkg/plugin/validation.go
estroz: Good catch, this should be done using string comp.

pkg/plugin/validation.go
estroz: These aren't used anywhere else, yet.

pkg/plugin/validation.go
estroz: These are really simple wrappers for errors, I don't thing that is necessary. Also the functions themselves are fairly simple, although I could see unit testing the version validator.

pkg/plugin/validation.go
estroz: This file contains functions that let you build a plugin validator by defining what a valid plugin name/version is. I don't see a reason to change anything but error types.

estroz:
Closing in favor of #1547

PR #1542, opened 2020-06-01
pkg/cli: only allow one default plugin per project version

estroz:
A `CLI`'s default plugins now contain only one plugin per project version, and the `Option` returned by `WithDefaultPlugins` will return an error if multiple are found. The one-to-one mapping identifies which plugin that should be run when no plugin key is passed via `layout` or `--plugins`.

* pkg/cli: change the type of cli.defaultPluginsFromOptions such that
it maps one project version to one plugin.

Relates to #1536.

/cc @joelanford @camilamacedo86 @DirectXMan12

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -68,8 +68,9 @@ type cli struct {
 
 	// Plugins injected by options.
 	pluginsFromOptions map[string][]plugin.Base
-	// Default plugins injected by options.
-	defaultPluginsFromOptions map[string][]plugin.Base
+	// Default plugins injected by options. Only one plugin per project version
+	// is allowed.
+	defaultPluginsFromOptions map[string]plugin.Base
 	// A plugin key passed to --plugins on invoking 'init'.
 	cliPluginKey string
 	// A filtered set of plugins that should be used by command constructors.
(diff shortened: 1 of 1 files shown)


estroz:
@camilamacedo86 #1498 should fail with this PR because both v2 and v3 plugins from that PR support an overlapping set of project versions, which this PR aims to avoid. The goal of this PR is to say "a user will use this plugin if they have not specified one for their project's version". Right now, having more than one is ambiguous given how a `CLI` is constructed.

camilamacedo86:
Hi @estroz, 

So, you are saying that the following scenario is invalid? 

// When has more than one supportable plugin. E.g:
// - go.kubebuilder.io/v2.0.0 which is supported by V2 and V3
// - go.kubebuilder.io/v3.0.0 which is supported by V3

If yes, should we not change the `var supportedProjectVersions = []string{config.Version2, config.Version3Alpha}`. I mean, should it really be an array?

estroz:
Project versions 2 and 3 should have different default plugins. Realistically the plugin for version 2 should be “special” since there is no layout key for that project scheme. See #1537 for an example.

camilamacedo86:
I changed the #1498 by removing the fix and keep v2.0.0 to V2 default and v3.0.0 to V3 default as you suggested above. WDYT about we merge this one for we are able to test in the CI all these plugins PR changes with +2 plugins versions at least to ensure?

camilamacedo86:
Hi @estroz, 

I thought that we would allow the multi-project version. However, thinking about keeping it simple and reduce the complexity I agree with you. It shows the best approach to move forward.

PR #1547, opened 2020-06-03
*: remove plugin minor/patch versions

estroz:
Plugin versions are now considered single numbers with optional "v" prefix and stability "alpha" or "beta" suffix, ex. "v2-alpha". Minor and patch version changes are poorly defined and may constitute breaking changes; having a single version number makes versioning simpler and permits breaking changes to be grouped into one "major" release for a plugin. See #1519 for more details.

* *: remove references to minor and patch versions for plugins
* pkg/plugin: implement a `Version` type returned by `Base.Version()` that is easily parsed and validated
* pkg/cli: simplify plugin resolver given the minor/patch number drop

**Note:** This change will affect project configs with `layout: go.kubebuilder.io/v2.0.0`, but because this value was not scaffolded by a released version of kubebuilder we shouldn't have to migrate anyone.

/cc @DirectXMan12 @joelanford @camilamacedo86


estroz:
While this gets reviewed/discussed

/hold

estroz:
Ready for review. It looks like there are a bunch of changes here, but most of them are removal of minor+patch version from comments/strings and centralizing plugin-related code where appropriate.

/hold cancel

pkg/cli/cli_suite_test.go
```diff
@@ -36,12 +36,13 @@ func TestCLI(t *testing.T) {
 
 // Test plugin types and constructors.
 type mockPlugin struct {
-	name, version   string
+	name            string
+	version         plugin.Version
 	projectVersions []string
```
camilamacedo86, line 41: I think it is now beyond what we need. but I am ok with 👍

pkg/plugin/plugin_test.go
```diff
@@ -0,0 +1,107 @@
...
+		err error
+	)
+
+	g.It("should pass validation", func() {
+		g.By("validating 1")
+		v, err = ParseVersion("1")
+		Expect(err).NotTo(HaveOccurred())
+		Expect(v.Number).To(BeNumerically("==", int64(1)))
+		Expect(v.Stability).To(Equal(""))
+
```
camilamacedo86, line 44: I still not happy with the tests because still not possible read them t not what/when/why as its docs examples.
Let's try to clarify. See that they are : 
- should pass validation by validating 1 - Why? 
- should pass validation by validating 2 - Why? 
- should pass validation by validating 22 - Why? 

When I think that they should be: 

```go
Describe("Checking plugin version validation", func() {

It("should work successfully when a plugin version is a positive number")
By(passing version number as 1)
By(passing version number as 2)

It("should fail when a plugin version is a negative number")
By(passing version number as -2)

It("should work successfully when a plugin version has a stage nomenclature)
By(passing version stage as alpha)
By(passing version stage as beta)

It(should fail if the plugin version has an invalid stage nomenclature)
```

pkg/cli/plugins_test.go
```diff
@@ -27,88 +29,89 @@ var _ = Describe("resolvePluginsByKey", func() {
```
camilamacedo86, line 29: the same scenario of https://github.com/kubernetes-sigs/kubebuilder/pull/1547/files#r435137335 (test descriptions are unreadable and do not allow we know easily what/when/why). See:  

Currenty, it is:
```go
Describe("resolvePluginsByKey", func() {
It("should resolve keys correctly", func() {
By("resolving foo.example.com/v1") // When? Why? 
By("resolving foo.example.com") //  When? Why?
```
 
IMO it should be such as:

```go
Describe("Checking plugin key resolution", func() {

It("should work successfully when is found a plugin that matches")
By(passing the name and version of the plugin)
By(passing just a valid plugin name)
By(passing just a valid plugin short name)

It("should fail when is not found a plugin that matches")
By(passing the name and version of the plugin)
By(passing just a valid plugin name)
By(passing just a valid plugin short name)

It("should fail when is found more than one plugin that matches")
By(passing the name and version of the plugin)
By(passing just a valid plugin name)
By(passing just a valid plugin short name)
```

pkg/plugin/plugin.go
```diff
@@ -17,27 +17,88 @@ limitations under the License.
...
-	// plugin supports, ex. []string{"2", "3"}. The returned slice cannot be empty.
-	SupportedProjectVersions() []string
+// Version is a plugin version containing a non-zero integer and an option stability value
+// that if present identifies a version as not stable to some degree.
+type Version struct {
+	Number    int64
+	Stability string
+}
+
+func (v Version) String() string {
```
camilamacedo86, line 39:
```suggestion
// String() returns the Version object in the string format such as v<number>-<stage>
func (v Version) String() string {
```

pkg/plugin/plugin.go
```diff
@@ -17,27 +17,88 @@ limitations under the License.
...
-	// Note: this version is different from config version.
-	Version() string
-	// SupportedProjectVersions lists all project configuration versions this
-	// plugin supports, ex. []string{"2", "3"}. The returned slice cannot be empty.
-	SupportedProjectVersions() []string
+// Version is a plugin version containing a non-zero integer and an option stability value
+// that if present identifies a version as not stable to some degree.
+type Version struct {
+	Number    int64
+	Stability string
```
camilamacedo86, line 36: IMO we could use the same k8s nomenclature : `stage`

pkg/plugin/plugin.go
```diff
@@ -17,27 +17,88 @@ limitations under the License.
...
+}
+
+func (v Version) String() string {
+	s := "v" + strconv.FormatInt(v.Number, 10)
+	if v.Stability != "" {
+		s += "-" + v.Stability
+	}
+	return s
+}
```
camilamacedo86, line 45:
```suggestion
// String returns the Plugin Versin in the string format (v<number> || v<number>-<stage>) 
func (v Version) String() string {
	if len(v.Stability) < 1 {
		return fmt.Sprintf("v%v",v.Number)
	}
	return fmt.Sprintf("v%v-%v",v.Number, v.Stability)
}
```

pkg/plugin/plugin.go
```diff
@@ -17,27 +17,88 @@ limitations under the License.
...
+// Validate ensures v contains a number and if it has a stability suffix that it is either "alpha" or "beta".
+func (v Version) Validate() error {
+	if v.Number == 0 {
+		return errors.New("integer value cannot be 0")
+	}
+	if v.Stability != "" && v.Stability != "alpha" && v.Stability != "beta" {
+		return errors.New(`suffix must be "alpha" or "beta"`)
+	}
+	return nil
+}
```
camilamacedo86, line 56:
```suggestion
const( // todo; add on top of the program
	AlphaStage = "alpha"
	BetaStage = "beta"
)

func (e ErrInvalidPluginVersion) Error() string {
	return fmt.Sprintf("invalid plugin version (%v:%v) : %v", e.number, e.stage, e.msg)
}

// Validate ensures v contains a number and if it has a stability suffix that it is either "alpha" or "beta".
func (v Version) Validate() error {
	if v.Number < 1 {
		return ErrInvalidPluginVersion{
			v.Number, 
			v.Stage, 
			"cannot be 0 or a negative number" }
	}
	if v.Stage != "" && v.Stage != AlphaStage && v.Stage != BetaStage { 
		return ErrInvalidPluginVersion{
			v.Number,
			v.Stage,
			"suffix stage must be alpha or beta" }
	}
	return nil
}
```

pkg/plugin/plugin.go
```diff
@@ -17,27 +17,88 @@ limitations under the License.
...
+
+// verRe defines the string format of a version.
+var verRe = regexp.MustCompile("^(v)?([1-9][0-9]*)(-alpha|-beta)?$")
```
camilamacedo86, line 59: add on top of the program

pkg/plugin/plugin.go
```diff
@@ -17,27 +17,88 @@ limitations under the License.
...
+	}
+	if v.Number, err = strconv.ParseInt(submatches[2], 10, 64); err != nil {
+		return v, err
+	}
+
+	// Parse stability suffix, if any.
+	v.Stability = strings.TrimPrefix(submatches[3], "-")
+
+	return v, v.Validate()
+}
```
camilamacedo86, line 86: All errors here also can be of the type `ErrInvalidPluginVersion`

pkg/plugin/plugin.go
```diff
@@ -17,27 +17,88 @@ limitations under the License.
...
+		return v, fmt.Errorf("version format must match %s", verRe.String())
+	}
+
+	// Parse version number.
+	if submatches[2] == "" {
+		return v, errors.New("version must contain an integer")
+	}
+	if v.Number, err = strconv.ParseInt(submatches[2], 10, 64); err != nil {
+		return v, err
+	}
```
camilamacedo86, line 80:
```suggestion
	versionNumber = submatches[2]
	if versionNumber == "" {
		return v, errors.New("version must contain an integer")
	}
	if v.Number, err = strconv.ParseInt(versionNumber, 10, 64); err != nil {
		return v, err
	}
```

pkg/plugin/plugin.go
```diff
@@ -17,27 +17,88 @@ limitations under the License.
...
+		if s == sp {
+			return 0
+		}
+		if s == "beta" || (s == "alpha" && sp == "") {
+			return 1
+		}
+	} else if v.Number > vp.Number {
+		return 1
+	}
+	return -1
```
camilamacedo86, line 101: I think that it would be the same of :  

 ```suggestion
	return strings.Compare( v.String(), vp.String())
```
PS.: passed in the tests as well.

estroz: You can't use string comparison on integers unfortunately because their bytes do not equate to their numerical value, ex. `strings.Compare("10", "2") == -1`.

pkg/plugin/plugin_test.go
estroz: Done.

pkg/cli/plugins_test.go
camilamacedo86: fair enough 👍

pkg/plugin/plugin.go
camilamacedo86: good point

PR #1550, opened 2020-06-04
pkg: refactor scaffolds and machinery under plugins

estroz:
All scaffold templates and machinery are housed under `pkg/plugin` since that machinery directly and only relates to plugins. Scaffold templates are also versioned with plugin major version, ex. `pkg/plugin/v2`.

Changes:
* `pkg/scaffold/internal/{filesystem,machinery} -> pkg/plugin/internal`
* `pkg/scaffold/interface.go -> pkg/plugin/scaffold`
* `pkg/scaffold -> pkg/plugin/v2/scaffolds`
* `pkg/plugin/internal/*.go -> pkg/plugin/util`

/cc @camilamacedo86 @joelanford @DirectXMan12


cmd/edit.go
```diff
@@ -24,7 +24,8 @@ import (
 
 	"sigs.k8s.io/kubebuilder/internal/cmdutil"
 	"sigs.k8s.io/kubebuilder/internal/config"
-	"sigs.k8s.io/kubebuilder/pkg/scaffold"
+	"sigs.k8s.io/kubebuilder/pkg/plugin/scaffold"
+	scaffolds "sigs.k8s.io/kubebuilder/pkg/plugin/v2/scaffolds"
```
camilamacedo86, line 28: may v2scaffolds

camilamacedo86: can be in the v3 follow up 👍

Issue #1543, opened 2020-06-01
make addon be an plugin 


camilamacedo86:
We probably want to follow up with the addon folks to see what's going on there, and eventually turn this into a proper plugin with plugins stage two
From: https://github.com/kubernetes-sigs/kubebuilder/pull/1527#issuecomment-637103722

camilamacedo86:
c/c @DirectXMan12 
/kind feature

camilamacedo86:
It might be better addressed after phase 2( https://github.com/kubernetes-sigs/kubebuilder/issues/1378) be in place. So, let's add this one to the next.

PR #1551, opened 2020-06-04
pkg/plugin/v3: add plugin `v3-alpha`

estroz:
A new plugin version `v3-alpha` will contain all new plugin/scaffold changes/features. Since plugin version `v2` is tested with project version 2, we probably should just test the new plugin with project version `3-alpha`. This won't necessarily be the case going forward, ex. once `4-alpha` is created we should generate separate testdata.

/cc @DirectXMan12 @camilamacedo86 @joelanford @mengqiy

### DIFF
--- a/pkg/plugin/v3/api.go
+++ b/pkg/plugin/v3/api.go
@@ -1,0 +1,213 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v3
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/pflag"
+
+	"sigs.k8s.io/kubebuilder/internal/cmdutil"
+	"sigs.k8s.io/kubebuilder/pkg/model"
+	"sigs.k8s.io/kubebuilder/pkg/model/config"
+	"sigs.k8s.io/kubebuilder/pkg/model/resource"
+	"sigs.k8s.io/kubebuilder/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/pkg/plugin/internal/util"
+	"sigs.k8s.io/kubebuilder/pkg/plugin/scaffold"
+	"sigs.k8s.io/kubebuilder/pkg/plugin/v3/scaffolds"
+	"sigs.k8s.io/kubebuilder/plugins/addon"
+)
+
+// (used only to gen api with --pattern=addon)
+// KbDeclarativePatternVersion is the sigs.k8s.io/kubebuilder-declarative-pattern version
+// TODO: remove this when a better solution for using addons is implemented.
+const KbDeclarativePatternVersion = "v0.0.0-20200522144838-848d48e5b073"
+
+type createAPIPlugin struct {
+	config *config.Config
+
+	// pattern indicates that we should use a plugin to build according to a pattern
+	pattern string
+
+	resource *resource.Options
+
+	// Check if we have to scaffold resource and/or controller
+	resourceFlag   *pflag.Flag
+	controllerFlag *pflag.Flag
+	doResource     bool
+	doController   bool
+
+	// force indicates that the resource should be created even if it already exists
+	force bool
+
+	// runMake indicates whether to run make or not after scaffolding APIs
+	runMake bool
+}
+
+var (
+	_ plugin.CreateAPI   = &createAPIPlugin{}
+	_ cmdutil.RunOptions = &createAPIPlugin{}
+)
+
+func (p createAPIPlugin) UpdateContext(ctx *plugin.Context) {
+	ctx.Description = `Scaffold a Kubernetes API by creating a Resource definition and / or a Controller.
+
+create resource will prompt the user for if it should scaffold the Resource and / or Controller.  To only
+scaffold a Controller for an existing Resource, select "n" for Resource.  To only define
+the schema for a Resource without writing a Controller, select "n" for Controller.
+
+After the scaffold is written, api will run make on the project.
+`
+	ctx.Examples = fmt.Sprintf(`  # Create a frigates API with Group: ship, Version: v1beta1 and Kind: Frigate
+  %s create api --group ship --version v1beta1 --kind Frigate
+
+  # Edit the API Scheme
+  nano api/v1beta1/frigate_types.go
+
+  # Edit the Controller
+  nano controllers/frigate/frigate_controller.go
+
+  # Edit the Controller Test
+  nano controllers/frigate/frigate_controller_test.go
+
+  # Install CRDs into the Kubernetes cluster using kubectl apply
+  make install
+
+  # Regenerate code and run against the Kubernetes cluster configured by ~/.kube/config
+  make run
+	`,
+		ctx.CommandName)
+}
+
+func (p *createAPIPlugin) BindFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&p.runMake, "make", true, "if true, run make after generating files")
+
+	fs.BoolVar(&p.doResource, "resource", true,
+		"if set, generate the resource without prompting the user")
+	p.resourceFlag = fs.Lookup("resource")
+	fs.BoolVar(&p.doController, "controller", true,
+		"if set, generate the controller without prompting the user")
+	p.controllerFlag = fs.Lookup("controller")
+
+	// TODO: remove this when a better solution for using addons is implemented.
+	if os.Getenv("KUBEBUILDER_ENABLE_PLUGINS") != "" {
+		fs.StringVar(&p.pattern, "pattern", "",
+			"generates an API following an extension pattern (addon)")
+	}
+
+	fs.BoolVar(&p.force, "force", false,
+		"attempt to create resource even if it already exists")
+	p.resource = &resource.Options{}
+	fs.StringVar(&p.resource.Kind, "kind", "", "resource Kind")
+	fs.StringVar(&p.resource.Group, "group", "", "resource Group")
+	fs.StringVar(&p.resource.Version, "version", "", "resource Version")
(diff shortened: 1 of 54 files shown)


estroz:
A thought on bumping plugin version stages, ex. `alpha` to `beta` and `beta` to "" (stable): we should probably:
1. Just change the plugin's stage string instead of creating a new package.
1. Build a mechanism into `pkg/cli` that auto-upgrades an `alpha` or `beta` to the next stage if available. This should never be an issue because the next stage should be no different from the previous one, just more stable.

Nothing needs to be done in this PR regarding ^.

estroz:
We may want to consider adding e2e test for plugin version `v3-alpha` too. Will likely follow up with these.

pkg/plugin/v3/api.go
```diff
@@ -0,0 +1,208 @@
...
+	"sigs.k8s.io/kubebuilder/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/pkg/plugin/internal/util"
+	"sigs.k8s.io/kubebuilder/pkg/plugin/scaffold"
+	"sigs.k8s.io/kubebuilder/pkg/plugin/v3/scaffolds"
+	"sigs.k8s.io/kubebuilder/plugins/addon"
+)
+
+// (used only to gen api with --pattern=addon)
+// KbDeclarativePatternVersion is the sigs.k8s.io/kubebuilder-declarative-pattern version
+const KbDeclarativePatternVersion = "v0.0.0-20200522144838-848d48e5b073"
```
DirectXMan12, line 43: do we need this?  I'm assuming going forward we just want `--plugin declarative` or something

pkg/plugin/v3/api.go
```diff
@@ -0,0 +1,208 @@
...
+		"if set, generate the resource without prompting the user")
+	p.resourceFlag = fs.Lookup("resource")
+	fs.BoolVar(&p.doController, "controller", true,
+		"if set, generate the controller without prompting the user")
+	p.controllerFlag = fs.Lookup("controller")
+
+	if os.Getenv("KUBEBUILDER_ENABLE_PLUGINS") != "" {
+		fs.StringVar(&p.pattern, "pattern", "",
+			"generates an API following an extension pattern (addon)")
+	}
```
DirectXMan12, line 114: ditto here

pkg/plugin/v3/api.go
```diff
@@ -0,0 +1,208 @@
...
+
+func (p *createAPIPlugin) Validate() error {
+	if err := p.resource.Validate(); err != nil {
+		return err
+	}
+
+	reader := bufio.NewReader(os.Stdin)
+	if !p.resourceFlag.Changed {
+		fmt.Println("Create Resource [y/n]")
+		p.doResource = util.YesNo(reader)
```
DirectXMan12, line 141: at some point we should re-evaluate the yes/no stuff.  IIRC we had wanted to do it for v2, but didn't want to break compat

pkg/plugin/v3/api.go
```diff
@@ -0,0 +1,208 @@
...
+	if p.doResource {
+		// Check that resource doesn't exist or flag force was set
+		if !p.force && p.config.HasResource(p.resource.GVK()) {
+			return errors.New("API resource already exists")
+		}
+
+		// Check that the provided group can be added to the project
+		if !p.config.MultiGroup && len(p.config.Resources) != 0 && !p.config.HasGroup(p.resource.Group) {
+			return fmt.Errorf("multiple groups are not allowed by default, to enable multi-group visit %s",
+				"kubebuilder.io/migration/multi-group.html")
```
DirectXMan12, line 158: why is this a `%s`?

pkg/plugin/v3/api.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+func (p *createAPIPlugin) PostScaffold() error {
+	// Load the requested plugins
+	switch strings.ToLower(p.pattern) {
+	case "":
+		// Default pattern
+	case "addon":
+		// Ensure that we are pinning sigs.k8s.io/kubebuilder-declarative-pattern version
+		err := util.RunCmd("Get controller runtime", "go", "get",
```
DirectXMan12, line 195: ??
this isn't getting CR, and shouldn't it be using the constant from above anyway?

pkg/plugin/v3/init.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = util.RunCmd("Update go.mod", "go", "mod", "tidy")
+	if err != nil {
+		return err
+	}
+
+	err = util.RunCmd("Running make", "make")
```
DirectXMan12, line 158: shouldn't this be conditional on the make flag defined in the other file?

pkg/plugin/v3/scaffolds/api.go
```diff
@@ -0,0 +1,125 @@
...
+		model.WithResource(s.resource),
+	)
+}
+
+func (s *apiScaffolder) scaffold() error {
+	if s.doResource {
+		s.config.AddResource(s.resource.GVK())
+
+		if err := machinery.NewScaffold(s.plugins...).Execute(
+			s.newUniverse(),
```
DirectXMan12, line 85: shouldn't we be passing the universe between scaffold calls here?

pkg/plugin/v3/scaffolds/init.go
```diff
@@ -0,0 +1,131 @@
...
+	return s.scaffold()
+}
+
+func (s *initScaffolder) scaffold() error {
+	bpFile := &templates.Boilerplate{}
+	bpFile.Path = s.boilerplatePath
+	bpFile.License = s.license
+	bpFile.Owner = s.owner
+	if err := machinery.NewScaffold().Execute(
+		s.newUniverse(""),
```
DirectXMan12, line 85: ditto here and elsewhere re: re-using the universe

DirectXMan12:
> If the answer to 1 is no (alpha implies the plugin is unstable), then the previous stable plugin version should be the default (v2 in this case). If yes, then v3-alpha is fine.

Per the meeting, IIRC we decided that the answer was "no", and that the default scaffolding for the v3 project version was going to be v2 till v3 stable was ready

pkg/plugin/v3/api.go
estroz: This is a typo, it should be "Get kubebuilder-declarative-pattern dependency". I'll change this text for now, but also add a comment to handle  the issue this tries to fix.

pkg/plugin/v3/api.go
estroz: I think this attempts to fix a CI issue where the testdata scaffolded with addons enabled will pull the latest kubebuilder-declarative-pattern commit, which messes up `git diff`.

We should probably scaffold a `tools.go` containing a pinned version of this. I could also see kustomize and controller-tools versions being pinned this way, which are currently dealt with in the Makefile.

pkg/plugin/v3/init.go
estroz: Currently there's no `init --make` flag unlike `create api`. Perhaps we should add one.

estroz:
@DirectXMan12 fixed what was easy/reasonable and added TODO's for the rest.

pkg/plugin/v3/api.go
```diff
@@ -0,0 +1,213 @@
...
+		// Default pattern
+	case "addon":
+		// Ensure that we are pinning sigs.k8s.io/kubebuilder-declarative-pattern version
+		// TODO: either find a better way to inject this version (ex. tools.go).
+		err := util.RunCmd("Get kubebuilder-declarative-pattern dependency", "go", "get",
+			"sigs.k8s.io/kubebuilder-declarative-pattern@"+KbDeclarativePatternVersion)
+		if err != nil {
+			return err
+		}
+	default:
```
camilamacedo86, line 205: Why change the impl that was done in https://github.com/kubernetes-sigs/kubebuilder/pull/1527/files?

pkg/plugin/v3/api.go
camilamacedo86: Hi @estroz why change the impl that was done in https://github.com/kubernetes-sigs/kubebuilder/pull/1527/files?

pkg/plugin/v3/init.go
camilamacedo86: Should we not here just replicate so the code to v3? IMO we should not change any impl in this pr just add the v3-alpha plugin and then we can check each case/scenario in new pr's.

estroz: This won't be changed here, just added a TODO.

pkg/plugin/v3/api.go
estroz: This won't be changed here, just added a TODO.

camilamacedo86: OK. Yep, we have an issue tracked for we create a plugin for that. See: https://github.com/kubernetes-sigs/kubebuilder/issues/1543

pkg/plugin/v3/scaffolds/api.go
camilamacedo86: I think that it also can be a TODO @estroz ?

pkg/plugin/v3/scaffolds/init.go
camilamacedo86: I think that it also can be a TODO @estroz ?

pkg/plugin/v3/init.go
```diff
@@ -0,0 +1,166 @@
...
+- a go.mod with project dependencies
+- a Kustomization.yaml for customizating manifests
+- a Patch file for customizing image for manager manifests
+- a Patch file for enabling prometheus metrics
+- a cmd/manager/main.go to run
+
+project will prompt the user to run 'dep ensure' after writing the project files.
+`
+	ctx.Examples = fmt.Sprintf(`  # Scaffold a project using the apache2 license with "The Kubernetes authors" as owners
+  %s init --project-version=2 --domain example.org --license apache2 --owner "The Kubernetes authors"
```
camilamacedo86, line 71:
```suggestion
  %s init --project-version=3-alpha --domain example.org --license apache2 --owner "The Kubernetes authors"
```

pkg/plugin/v3/init.go
```diff
@@ -0,0 +1,166 @@
...
+	skipGoVersionCheck bool
+}
+
+var (
+	_ plugin.Init        = &initPlugin{}
+	_ cmdutil.RunOptions = &initPlugin{}
+)
+
+func (p *initPlugin) UpdateContext(ctx *plugin.Context) {
+	ctx.Description = `Initialize a new project including vendor/ directory and Go package directories.
```
camilamacedo86, line 56:
```suggestion
	ctx.Description = `Initialize a new project including Go package directories.
```
I should be removed already.

pkg/plugin/v3/init.go
estroz: Follow-up

pkg/plugin/v3/init.go
estroz: Follow-up

PR #1546, opened 2020-06-03
🌱 kubebuilder init generates a deployable project

sethp-nr:
This is an early step toward better external webhook support (for types
not local to the project). By ensuring the project is working before
running `kubebuilder create api` we open the possibility of having a
project that's made up entirely of controllers or webhooks for types
defined elsewhere. For more context, see: #1476 

~Additionally, this change includes a `test_e2e_local.sh` script for
running e2e tests locally. (maybe resolving #1435?)~ split out into #1556

Issue #1435, opened 2020-03-14
allow contributors to run the e2e tests (test-e2e.sh) locally


camilamacedo86:
**Story**

I am as Kubebuilder contributor, would like to be able to check the e2e tests (test-e2e.sh) used to verify the PR in the CI via Prow locally. 

**Acceptance Criteria**
- The target needs work for OSX and Linux

**Eng Notes**
- before running the script is required to export the env var with the k8s version. See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test_e2e.sh#L33 (E.g export KIND_K8S_VERSION=v1.17.0)
- Also, have the kind installed. See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/scripts/setup.sh#L24.  
- See that is possible to perform an install which will work for both SO: https://kind.sigs.k8s.io/docs/user/quick-start/
- The arg `-v` [here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/scripts/setup.sh#L24)  shows invalid for the latest version of kind and needs to be removed

And then, solve the following problem to run it locally:

```

$ export KIND_K8S_VERSION=v1.17.0
camilamacedo@Camilas-MacBook-Pro ~/go/src/sigs.k8s.io/kubebuilder (contribute-guide) $ make test-ci-e2e
export KIND_K8S_VERSION=v1.17.0
brew install kind
Warning: kind 0.7.0 is already installed and up-to-date
To reinstall 0.7.0, run `brew reinstall kind`
./test_e2e.sh
installing prerequirements ...
fetching kb tools
x kubebuilder/
x kubebuilder/bin/
x kubebuilder/bin/etcd
x kubebuilder/bin/kubectl
x kubebuilder/bin/kube-apiserver
Checking for dep
Checking for kind
building kubebuilder
setting up env vars
done
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.17.0) 🖼 
 ✓ Preparing nodes 📦 📦 📦 📦  
 ✓ Writing configuration 📜 
 ✓ Starting control-plane 🕹️ 
 ✓ Installing CNI 🔌 
 ✓ Installing StorageClass 💾 
 ✓ Joining worker nodes 🚜 
 ✓ Waiting ≤ 1m0s for control-plane = Ready ⏳ 
 • Ready after 27s 💚
Set kubectl context to "kind-kind"
You can now use your cluster with:

[...]

PR #1556, opened 2020-06-10
🌱 local e2e script

sethp-nr:
This change introduces a `test_e2e_local.sh` script for running e2e
tests locally. Addresses #1435.

Split out from #1546

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,6 +54,7 @@ Following the targets that can be used to test your changes locally.
 | make lint |  Check the code implementation | yes   |
 | make test-coverage |  Run coveralls to check the % of code covered by tests | yes   |
 | make check-testdata |  Checks if the testdata dir is updated with the latest changes | yes   |
+| make test-e2e-local |  Runs the CI e2e tests locally | no   |
 
 **NOTE** To use the `make lint` is required to install `golangci-lint` locally. More info: https://github.com/golangci/golangci-lint#install
 
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,11 @@ test-coverage:  ## Run coveralls
 	# run the go tests and gen the file coverage-all used to do the integration with coverrals.io
 	go test -failfast -tags=integration -coverprofile=coverage-all.out -covermode=count ./pkg/... ./cmd/...
 
+.PHONY: test-e2e-local
+test-e2e-local: ## It will run the script to install kind and run e2e tests
+	## To keep the same kind cluster between test runs, use `SKIP_KIND_CLEANUP=1 make test-e2e-local`
+	./test_e2e_local.sh
+
 .PHONY: check-testdata
 check-testdata: ## Run the script to ensure that the testdata is updated
 	./check_testdata.sh
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -15,13 +15,17 @@
 # limitations under the License.
 
 K8S_VERSION=$1
+KIND_NAME="kind"
+if [ $# -gt 1 ]; then
+    KIND_NAME=$2
+fi
 
 export GO111MODULE=on
 
 # setup go module to create the cluster
 
 # You can use --image flag to specify the cluster version you want, e.g --image=kindest/node:v1.13.6, the supported version are listed at https://hub.docker.com/r/kindest/node/tags
-kind create cluster -v 4 --retain --wait=1m --config test/kind-config.yaml --image=kindest/node:$K8S_VERSION
+kind create cluster -v 4 --name $KIND_NAME --retain --wait=1m --config test/kind-config.yaml --image=kindest/node:$K8S_VERSION
 
 # setup the go modules required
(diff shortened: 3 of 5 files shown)


sethp-nr:
I was going off the eng notes from that #1435, which include:

> Also, have the kind installed. See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/scripts/setup.sh#L24.

and 

> before running the script is required to export the env var with the k8s version. See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test_e2e.sh#L33 (E.g export KIND_K8S_VERSION=v1.17.0)

Following those instructions, the local script here provided does work for me on macOS. 

> So, we cannot install and set up the kind in this scenario.

 > However, for local dev env people has minikube or any other k8s cluster to work with already.

Would you explain a little more what you mean here? Removing the `kind` dependency for the e2e tests and letting devs bring their own clusters would be a bit more work, I think, and make the environment less similar to the real e2e test.

In my experience it's been acceptable to folks to ask them to install `kind` to do kubernetes related development, and is the growing "default" path for doing these kinds of things. What do you think of keeping the `kind` dependency for now, and if someone wants to use minikube in the future for the local tests they can file a PR for that?

sethp-nr:
I had another thought about the `kind` dependency: would it be better to install a project-local version of the kind binary by way of a `tools` module?

camilamacedo86:
Hi @sethp-nr, 

Sorry, I catcher now. If it works for mac and Linux shows fine. I am testing it now in a mac.

scripts/setup.sh
```diff
@@ -15,13 +15,17 @@
 # limitations under the License.
```
camilamacedo86, line 15: Could we add a new Makefile target for that? 

.PHONY: test-e2e-local
test-e2e-local: ## It will run the script to install kind and run e2e tests         
	./test_e2e_local.sh

test_e2e_local.sh
```diff
@@ -0,0 +1,43 @@
...
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+source common.sh
+
+export TRACE=1
+export GO111MODULE=on
+
```
camilamacedo86, line 24: Could we export here by default `export KIND_K8S_VERSION=v1.17.0`  and/or in the Makefile?

test_e2e_local.sh
```diff
@@ -0,0 +1,43 @@
...
+export KIND_CLUSTER=local-kubebuilder-e2e
+if ! kind get clusters | grep -q $KIND_CLUSTER ; then
+    source "$(pwd)/scripts/setup.sh" ${KIND_K8S_VERSION} $KIND_CLUSTER
+    docker pull gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0
+    kind load --name $KIND_CLUSTER docker-image gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0
+fi
+
+kind export kubeconfig --kubeconfig $tmp_root/kubeconfig --name $KIND_CLUSTER
+export KUBECONFIG=$tmp_root/kubeconfig
+
```
camilamacedo86, line 40: We need the trap to ensure that all will be cleaned if an issue is faced.

```shell
# remove running containers on exit
function cleanup() {
    kind delete cluster
}

trap cleanup EXIT
``

File: scripts/setup.sh (hunk 1 above)
Comment (camilamacedo86) [reply scripts/setup.sh:15]: Also could you add new makefile target in : https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#what-to-do-before-submitting-a-pull-request

File: test_e2e_local.sh (hunk 3 above)
Comment (sethp-nr) [reply test_e2e_local.sh:41]: I found that on my machine, standing up the kind cluster a significant chunk of the total test run time. So for the local script I added in an override that leaves the kind cluster in place between runs: the tests seem very good at either cleaning up after themselves or not conflicting with each other, or both.

File: CONTRIBUTING.md [hunk 4]
```diff
@@ -54,14 +54,15 @@ Following the targets that can be used to test your changes locally.
 | make lint |  Check the code implementation | yes   |
 | make test-coverage |  Run coveralls to check the % of code covered by tests | yes   |
 | make check-testdata |  Checks if the testdata dir is updated with the latest changes | yes   |
+| make test-e2e-local |  Runs the CI e2e tests locally | no   |
```
camilamacedo86, line 57: Could you please revert the changes that has no relation with the scope of this pr? (E.g the spaces an just keep what is related to the new target)

Makefile
```diff
@@ -77,11 +77,16 @@ test: ## Run the unit tests (used in the CI)
 
 .PHONY: test-coverage
 test-coverage:  ## Run coveralls
-	# remove all coverage files if exists
+	# remove all coverage files if existss
```
camilamacedo86, line 80: revert ^

sethp-nr: oof, good catch!

Can you tell I have one of the awful scissor apple keyboards? :)

sethp-nr:
Thanks @camilamacedo86 for your feedback!

PR #1564, opened 2020-06-17
pkg/cli/init.go: make flag help consistent

estroz:
Any flag help printed should be consistent between invocations so generated CLI docs (cobra has a nice API for this) don't break randomly. Also remove duplicate plugin keys from flag help.

/cc @camilamacedo86 @DirectXMan12

### DIFF
--- a/pkg/cli/init.go
+++ b/pkg/cli/init.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


pkg/cli/init.go
```diff
@@ -100,18 +101,24 @@ func (c cli) getAvailableProjectVersions() (projectVersions []string) {
...
 }
 
 func (c cli) getAvailablePlugins() (pluginKeys []string) {
+	keySet := make(map[string]struct{})
 	for _, versionedPlugins := range c.pluginsFromOptions {
 		for _, p := range versionedPlugins {
 			// Only return non-deprecated plugins.
 			if _, isDeprecated := p.(plugin.Deprecated); !isDeprecated {
-				pluginKeys = append(pluginKeys, strconv.Quote(plugin.KeyFor(p)))
+				keySet[plugin.KeyFor(p)] = struct{}{}
```
hasbro17, line 114: I can see that we're doing `sort.Strings(pluginKeys)` which orders the plugin keys but any reason for building the `keySet` here and then appending the keys to `pluginKeys`?
`pluginKeys = append(pluginKeys, strconv.Quote(plugin.KeyFor(p)))` isn't too unreadable imo.

estroz: `c.pluginsFromOptions` are indexed by project version, so two `versionedPlugins` slices can be intersecting sets if one plugin supports two project versions (`go/v2` supports project versions `2` and `3-alpha`).

hasbro17: Ah, double counting. Gotcha.

PR #1569, opened 2020-06-20
feat: add go test with -race conditional

camilamacedo86:
Add -race conditional check in the go tests. 
See: https://hermanschaaf.com/running-the-go-race-detector-with-cover/


camilamacedo86:
c/c @joelanford

Makefile
```diff
@@ -80,7 +80,7 @@ test-coverage:  ## Run coveralls
 	# remove all coverage files if exists
 	- rm -rf *.out
 	# run the go tests and gen the file coverage-all used to do the integration with coverrals.io
-	go test -failfast -tags=integration -coverprofile=coverage-all.out -covermode=count ./pkg/... ./cmd/...
+	go test -race -failfast -tags=integration -coverprofile=coverage-all.out -covermode=atomic ./pkg/... ./cmd/...
```
gabbifish, line 83: For the same reason included in my review of https://github.com/kubernetes-sigs/kubebuilder/pull/1570, it looks like `cover` is in atomic mode by default
```suggestion
	go test -race -failfast -tags=integration -coverprofile=coverage-all.out ./pkg/... ./cmd/...
```

PR #1570, opened 2020-06-20
feat: improve go test for the scaffolded projects due checking race conditionals (v3 only)

camilamacedo86:
More info:  https://hermanschaaf.com/running-the-go-race-detector-with-cover/

### DIFF
--- a/pkg/plugin/v3/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugin/v3/scaffolds/internal/templates/makefile.go
@@ -71,7 +71,7 @@ all: manager
 
 # Run tests
 test: generate fmt vet manifests
-	go test ./... -coverprofile cover.out
+	go test -race -coverprofile cover.out ./...
 
 # Build manager binary
 manager: generate fmt vet
--- a/testdata/project-v3-addon/Makefile
+++ b/testdata/project-v3-addon/Makefile
@@ -15,7 +15,7 @@ all: manager
 
 # Run tests
 test: generate fmt vet manifests
-	go test ./... -coverprofile cover.out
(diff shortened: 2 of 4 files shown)


camilamacedo86:
c/c @joelanford

pkg/plugin/v3/scaffolds/internal/templates/makefile.go
```diff
@@ -71,7 +71,7 @@ all: manager
 
 # Run tests
 test: generate fmt vet manifests
-	go test ./... -coverprofile cover.out
+	go test -race -covermode atomic -coverprofile cover.out ./...
```
gabbifish, line 74: It looks like, as of Go 1.3, the atomic mode will be enabled by default when `-cover` is run with `-race`, so we don't have to provide the `-covermode atomic` flag anymore (as mentioned in the update section of https://hermanschaaf.com/running-the-go-race-detector-with-cover/). 
```suggestion
	go test -race -coverprofile cover.out ./...
```

pwittrock: Cool.  @gabbifish -- would you provide a link to where that is documented so that @camilamacedo86 can include it as a comment on the makefile.go and future folks will know.

gabbifish: The Go `cover` docs [are quite light](https://golang.org/cmd/cover/) and don't seem to have documentation for this behavior. https://github.com/golang/go/issues/7013 is the issue where this default atomic behavior is formally accepted and implemented.

camilamacedo86: Done

PR #1571, opened 2020-06-20
feat: add undeploy IMG option (V3 only)

camilamacedo86:
Add option to undeploy controller.

### DIFF
--- a/pkg/plugin/v3/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugin/v3/scaffolds/internal/templates/makefile.go
@@ -94,6 +94,10 @@ deploy: manifests kustomize
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
+# UnDeploy controller from the configured Kubernetes cluster in ~/.kube/config
+undeploy:
+	$(KUSTOMIZE) build config/default | kubectl delete -f -
+
 # Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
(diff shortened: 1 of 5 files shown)


camilamacedo86:
c/c @joelanford

pkg/plugin/v3/scaffolds/internal/templates/makefile.go
```diff
@@ -94,6 +94,11 @@ deploy: manifests kustomize
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
+# UnDeploy controller from the configured Kubernetes cluster in ~/.kube/config
+undeploy:
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
```
pwittrock, line 99: Is setting the image necessary for deletion?

gabbifish: I don't think setting the image is necessary for deletion (I just tried out `make deploy` followed by `make undeploy` on a kind cluster). I think this logic might just be copied over from `make deploy` but can be deleted?
```suggestion
	cd config/manager
```

camilamacedo86: The `$(KUSTOMIZE) edit set image controller=${IMG}` will update the kustomization files with img for we are able to delete that img informed. But, I agree that we updated the files in the deploy so we could just assume that it is right and not updated again. I will change that. Good catcher, btw 👍

camilamacedo86: Done

PR #1580, opened 2020-06-30
fix crd-version-pref url

wyyxd2017:
i fix crd-version-pref url！thanks!


gabbifish:
These link updates look correct to me (tested them on the deploy-preview instance for docs). We'll need a kubebuilder repo owner to actually /lgtm this. :)

PR #1587, opened 2020-07-05
Add missing kubebuilder annotation to the getting started document

### LINKED ISSUE #1586: The Getting started guide missing a prepocessor annotation
# In the documentation there is an example that probably misses an annotation

I expected the snippet to work, but the DeepCopyObject is not created so the `make run` exits with this error
```

api/v1/guestbook_types.go:76:25: cannot use &Guestbook literal (type *Guestbook) as type runtime.Object in argument to SchemeBuilder.Register:
	*Guestbook does not implement runtime.Object (missing DeepCopyObject method)

```

It can be replicated by following the gettting started guide including the optional part

- go version go1.14 linux/amd64
- kubebuilder version: 
  - KubeBuilderVersion:"2.3.1", 
  - KubernetesVendor:"1.16.4", 
  - GitCommit:"8b53abeb4280186e494b726edf8f54ca7aa64a49"
- controller-tools version: ?
- kubernetes:
  - kubectl: v1.17.0
  - server: v1.16.8-eks-e16311

/kind documentation

natalefalco:
I changed the documentation, so it will contain a working example

Previously, the example not generated the deep copy function required by the SchemeBuilder.Register

It fixes #1586 issue, that complains about the not out of box working example

### DIFF
--- a/docs/book/src/quick-start.md
+++ b/docs/book/src/quick-start.md
@@ -131,6 +131,11 @@ type GuestbookStatus struct {
 	Standby []string `json:"standby"`
 }
 
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
+
+// Guestbook is the Schema for the guestbooks API
 type Guestbook struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`


docs/book/src/quick-start.md
```diff
@@ -131,6 +131,7 @@ type GuestbookStatus struct {
 	Standby []string `json:"standby"`
 }
 
+// +kubebuilder:object:root=true
 type Guestbook struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
```
camilamacedo86, line 137: So, we not to add all markers added by default. See: https://github.com/kubernetes-sigs/kubebuilder/blob/1eccd8fbe9f53ffa117b6367976230eb1fe13b27/testdata/project-v3/api/v1/admiral_types.go could you please update this one?

natalefalco: I added the status marker if the scoping required too, ping me and i add that too

docs/book/src/quick-start.md
```diff
@@ -131,6 +131,8 @@ type GuestbookStatus struct {
 	Standby []string `json:"standby"`
 }
 
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
 type Guestbook struct {
```
camilamacedo86, line 136:
```suggestion
// +kubebuilder:resource:scope=Cluster

// Guestbook is the Schema for the guestbooks API
type Guestbook struct {
```

camilamacedo86:
Hi @natalefalco, 

Really tks for your contribution 👍 . Just a few nits for we are able to merge it:

- See that still missing one marker
- Also, all commits need to be squashed.

Issue #1006, opened 2019-09-21
Webhook Patch Template K8s Version Incorrect


cliveseldon:
The patch for a webhook adds a comment:

https://github.com/kubernetes-sigs/kubebuilder/blob/cd0b10ff1a7966ce3d449ae9fbb8d7029259bc31/pkg/scaffold/v2/crd/enablewebhook_patch.go#L54-L56

But I see `preserveUnknownFields` only in k8s version >=1.15

Running this patch on earlier clusters gives an error.

```
kustomize build config/crd | kubectl apply -f -
error: error validating "STDIN": error validating data: ValidationError(CustomResourceDefinition.spec): unknown field "preserveUnknownFields" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceDefinitionSpec; if you choose to ignore these errors, turn validation off with --validate=false

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"clean", BuildDate:"2019-09-18T14:36:53Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.7-gke.8", GitCommit:"7d3d6f113e933ed1b44b78dff4baf649258415e5", GitTreeState:"clean", BuildDate:"2019-06-19T16:37:16Z", GoVersion:"go1.11.5b4", Compiler:"gc", Platform:"linux/amd64"}
```

/kind bug

mengqiy:
Thanks for bring this up!
The patch itself doesn't have the `preserveUnknownFields` field.
It seems this bug is caused by an unexpected defaulting kicks in somehow. My guess there is an undesired roundtrip.
Ref: https://github.com/kubernetes/apiextensions-apiserver/blob/release-1.15/pkg/apis/apiextensions/v1beta1/defaults.go#L72-L74

I need to dig it further.

mengqiy:
@cliveseldon What is your kustomize version?

cliveseldon:
Yes, sorry I think the issue is https://github.com/kubernetes-sigs/controller-tools/issues/279

cliveseldon:
[...]

Issue #1065, opened 2019-10-02
Migrate existing KB project to v1 CRDs and Webhooks with minimal user effort


erictune:
Facts:
- 1.16 introduced CRD and Webhooks v1 APIs.
- CRD and Webhooks v1beta1 are deprecated and scheduled for removal in v1.19.
- Kubernetes users use wide range of Kubernetes minor versions.   

Consequences:
- In about 12 months, expect clusters running versions 1.15  through 1.19.  
- If KB only generates v1beta1 or only generates v1 CRDs/WHs, then my controller can't be installed on that range of clusters.

Request for KubeBuilder:
- Develop CRD and Webhook generation that can emit both v1beta1 and v1 CRDs/WHs.
- Document how a developer can have a single project and a single release artifact and have it work on v1beta1-only, both, and v1-only clusters.
- Don't decide that KubeBuilder v2.x.0 (or likewise for controller-tools, etc) is going to do a hard switch to generating v1 CRDs/WHs.    I want to get new features and fixes for KB while still releasing controllers work with 1.15 and earlier Kubernetes versions.

Related issues:
- #903 is about using the new features of v1 CRDs and webhooks.  This one is about expressing existing KB projects using v1 types.
- #944 is about APIs being deprecated in 1.16.  This one is about APIs being deprecated in 1.1

Additional Data about Versions:
   - From this slide of the [Kubernetes Apps Survey](https://docs.google.com/presentation/d/1utT0K-u1nl2apXRo29GaBvRV1x7mFLeQSgpw8mI_nGM/edit#slide=id.g3882d9116d_2_34), it appears that most (~95%) of users are running a cluster from one of the 4 most recent minor versions.
  - From the [detailed data of the same survey in column AN](https://docs.google.com/spreadsheets/d/12ilRCly2eHKPuicv1P_BD6z__PXAqpiaR-tDYe2eudE/edit#gid=840936477),  many users have 3 or even 4 minor versions in use concurrently.
- I'd expect this distribution to broaden over time, so having users spread across 5 versions in 2020 seems likely.

/kind feature

DirectXMan12:
> Don't decide that KubeBuilder v2.x.0 (or likewise for controller-tools, etc) is going to do a hard switch to generating v1 CRDs/WHs. I want to get new features and fixes for KB while still releasing controllers work with 1.15 and earlier Kubernetes versions.

[...]

PR #1589, opened 2020-07-06
set spec.preserveUnknownFields to false for Webhook conversion

liusy182:
This is to fix error:  `"spec.conversion.strategy: Invalid value: "Webhook": must be None if spec.preserveUnknownFields is true"`. The error occurs while working on the webhook example in [kubebook](https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation.html).

The same error was mentioned in #1006. Although this is a closed ticket, some people have encountered similar failure very recently [in a different issue](https://github.com/kubernetes-sigs/controller-tools/issues/279#issuecomment-640081748).

The root cause of the error is the 2 conflicting specs that are not properly specified, according to documentation:

- [`spec.preserveUnknownFields`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#customresourcedefinitionspec-v1beta1-apiextensions-k8s-io) defaults to `true` for `v1beta`.

- [`CustomResourceConversion`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#customresourceconversion-v1beta1-apiextensions-k8s-io) requires the value of `spec.preserveUnknownFields` to be `false` when the `strategy` is `Webhook `.

This PR provides a fix by setting `spec.preserveUnknownFields` to `false`. 

Tested with examples provided by kubebook.

### DIFF
--- a/pkg/plugin/v2/scaffolds/internal/templates/crd/enablewebhook_patch.go
+++ b/pkg/plugin/v2/scaffolds/internal/templates/crd/enablewebhook_patch.go
@@ -49,6 +49,7 @@ kind: CustomResourceDefinition
 metadata:
   name: {{ .Resource.Plural }}.{{ .Resource.Domain }}
 spec:
+  preserveUnknownFields: false
   conversion:
     strategy: Webhook
     webhookClientConfig:
--- a/pkg/plugin/v2/scaffolds/internal/templates/crd/enablewebhook_patch.go
+++ b/pkg/plugin/v2/scaffolds/internal/templates/crd/enablewebhook_patch.go
@@ -49,7 +49,6 @@ kind: CustomResourceDefinition
 metadata:
   name: {{ .Resource.Plural }}.{{ .Resource.Domain }}
 spec:
-  preserveUnknownFields: false
   conversion:
     strategy: Webhook
     webhookClientConfig:
--- a/pkg/plugin/v3/scaffolds/internal/templates/crd/enablewebhook_patch.go
+++ b/pkg/plugin/v3/scaffolds/internal/templates/crd/enablewebhook_patch.go
@@ -49,6 +49,7 @@ kind: CustomResourceDefinition
 metadata:
   name: {{ .Resource.Plural }}.{{ .Resource.Domain }}
 spec:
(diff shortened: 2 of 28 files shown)


liusy182:
Hi! I realize that the [samples in the docs folder](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/project/config/crd/patches/webhook_in_cronjobs.yaml#L9) might need update as well. Since this is somewhat unrelated to the code generation, is the preference to have them in the same PR or as a separate issue?

camilamacedo86:
/hold 

Really nice see your contribution. However, we can just apply changes to v3 as @estroz described above.
Also, I am not sure if we should fix this by setting it. We might just need to migrate to v1 api as described in: https://github.com/kubernetes-sigs/kubebuilder/issues/1065

liusy182:
Thanks! please let me know if we want to apply this to V3 and I can update that. Otherwise feel free to close the PR if migration is the way to go.

camilamacedo86:
Hi @liusy182, 

IMO: What we need to do is: update the v3 plugin only to all scaffold by default CRD and Webhooks using the API v1 instead of betav1. Then this field is not required and the project starts to works by default just in clusters >= 0.16. Is it make sense? Would like to work on it?

liusy182:
sure i can give a try on that.

estroz:
This PR either needs a changes to make all CRD patches v1, or a follow-up. I’m fine with the latter.

@camilamacedo86 https://github.com/kubernetes-sigs/controller-tools/pull/468 has been merged so we can avoid explicitly setting default version to v1 in the Makefile.

camilamacedo86:
@estroz Great news  kubernetes-sigs/controller-tools#468 👍

estroz:
On second thought, do we still want to support v1beta1 CRDs for those targeting older clusters? If so we should parameterize CRD version like `init --crd-version` and `create api --crd-version`, as @camilamacedo86 has suggested previously.

/cc @DirectXMan12 @pwittrock 

/hold

estroz:
This PR is superseded by #1644

PR #1590, opened 2020-07-09
Update quick-start.md

Devying:
go proxy set

### DIFF
--- a/docs/book/src/quick-start.md
+++ b/docs/book/src/quick-start.md
@@ -57,6 +57,29 @@ cd $GOPATH/src/example
 kubebuilder init --domain my.domain
 ```
 
+If there is an error similar to `dial tcp x.x.x.x:443: i/o timeout`, you need to set go proxy.
+Go version >= 1.13
+
+```bash
+go env -w GO111MODULE=on
+go env -w GOPROXY="https://goproxy.io,direct"
+
+# Set environment variable allow bypassing the proxy for selected modules (optional)
+go env -w GOPRIVATE="*.corp.example.com"
+
+# Set environment variable allow bypassing the proxy for specified organizations (optional)
+go env -w GOPRIVATE="example.com/org_name"
+```
+
+Go version <= 1.12
+
+```bash
+# Enable the go modules feature
+export GO111MODULE="on"
+# Set the GOPROXY environment variable
+export GOPROXY="https://goproxy.io"
+```
+
 <aside class="note">
 <h1>Not in $GOPATH</h1>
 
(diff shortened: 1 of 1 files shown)


Devying:
I signed it

Devying:
ok-to-test

docs/book/src/quick-start.md
```diff
@@ -57,6 +57,29 @@ cd $GOPATH/src/example
...
+```
+
+Go version <= 1.12
+
+```bash
+# Enable the go modules feature
+export GO111MODULE="on"
+# Set the GOPROXY environment variable
+export GOPROXY="https://goproxy.io"
+```
```
camilamacedo86, line 81: The project do not support go 1.12

docs/book/src/quick-start.md
```diff
@@ -57,6 +57,29 @@ cd $GOPATH/src/example
 kubebuilder init --domain my.domain
 ```
 
+If there is an error similar to `dial tcp x.x.x.x:443: i/o timeout`, you need to set go proxy.
```
camilamacedo86, line 60: It shows specific to your environment. I do not think that the quick start would be an appropriate place for that.

docs/book/src/quick-start.md
```diff
@@ -57,6 +57,29 @@ cd $GOPATH/src/example
...
+Go version >= 1.13
```
camilamacedo86, line 61: It is in the pre-requirements section already.

docs/book/src/quick-start.md
```diff
@@ -57,6 +57,29 @@ cd $GOPATH/src/example
...
+
+```bash
+go env -w GO111MODULE=on
+go env -w GOPROXY="https://goproxy.io,direct"
```
camilamacedo86, line 65: Go 1.13 already set these values by default. However, if you might would like to help others I think it the place would be tracking and issue with the solution and/or creating a FAQ section in the docs instead of it. WDYT?

docs/book/src/quick-start.md
```diff
@@ -57,6 +57,29 @@ cd $GOPATH/src/example
...
+
+# Set environment variable allow bypassing the proxy for selected modules (optional)
+go env -w GOPRIVATE="*.corp.example.com"
+
+# Set environment variable allow bypassing the proxy for specified organizations (optional)
+go env -w GOPRIVATE="example.com/org_name"
```
camilamacedo86, line 71: It is to solve a specific need and a specific env setup. IMO also, we could address it better in another doc section (may FAQ) where we would add when and what are the circumstances that would make it required as well.

docs/book/src/quick-start.md
```diff
@@ -190,13 +213,40 @@ Build and push your image to the location specified by `IMG`:
...
+#Go version >=1.13
+RUN go env -w GO111MODULE=on
+RUN go env -w GOPROXY=https://goproxy.io,direct
+RUN go mod download
+
+#Go version <=1.12
+RUN export GO111MODULE="on"
+RUN export GOPROXY="https://goproxy.io"
+RUN go mod download
+
```
camilamacedo86, line 236: Ditto above.

docs/book/src/quick-start.md
```diff
@@ -190,13 +213,40 @@ Build and push your image to the location specified by `IMG`:
...
+You may get another error similar to `Get https://gcr.io/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)`, You need to edit the `Dockerfile` and replace the image `gcr.io/distroless/static:nonroot`
 
 Deploy the controller to the cluster with image specified by `IMG`:
 
 ```bash
 make deploy IMG=<some-registry>/<project-name>:tag
 ```
 
+If the pod's status is `ErrImagePull`,You need to edit the file `config/default/manager_auth_proxy_patch.yaml` and replace the image.
```
camilamacedo86, line 247: what was the problem faced and what was the edition that you did in this case? Could you please raise an issue for we are able to check? It might be something that could be solved by the tool.

docs/book/src/quick-start.md
```diff
@@ -190,13 +213,40 @@ Build and push your image to the location specified by `IMG`:
...
+RUN go mod download
+
+#Go version <=1.12
+RUN export GO111MODULE="on"
+RUN export GOPROXY="https://goproxy.io"
+RUN go mod download
+
+```
+
+You may get another error similar to `Get https://gcr.io/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)`, You need to edit the `Dockerfile` and replace the image `gcr.io/distroless/static:nonroot`
```
camilamacedo86, line 239: ditto above.

docs/book/src/quick-start.md
Devying: All right,It is more appropriate to the issue or FAQ

docs/book/src/quick-start.md
Devying: All right,It is more appropriate to the issue or FAQ

docs/book/src/quick-start.md
Devying: All right,It is more appropriate to the issue or FAQ

docs/book/src/quick-start.md
Devying: All right,It is more appropriate to the issue or FAQ

Devying:
All right,It is more appropriate to the issue or FAQ

PR #1595, opened 2020-07-14
fix: use kube-system instead of system

pigletfly:
I think it's better to use kube-system instead of system, otherwise, we should provide a way to choose namespace and define custom ServiceAccount name.

### DIFF
--- a/pkg/plugin/v2/scaffolds/internal/templates/mgrrolebinding.go
+++ b/pkg/plugin/v2/scaffolds/internal/templates/mgrrolebinding.go
@@ -51,5 +51,5 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: default
-  namespace: system
+  namespace: kube-system
 `
--- a/pkg/plugin/v3/scaffolds/internal/templates/mgrrolebinding.go
+++ b/pkg/plugin/v3/scaffolds/internal/templates/mgrrolebinding.go
@@ -51,5 +51,5 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: default
-  namespace: system
+  namespace: kube-system
 `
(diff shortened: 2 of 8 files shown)


pwittrock:
The namespace should probably be unique to each project.  Overall the whole config generation approach should be reconsidered.

pigletfly:
Agreed.If the namespace is not unique, I'd rather put all the controllers in the kube-system namespace,cause it's a protected namespace.

estroz:
These namespaces are not actually used though right? They're `kustomize`d away on `make deploy`, so they're effectively just placeholders. You could apply them directly, but they're not intended to be used that way afaik.

camilamacedo86:
/hold

I'd suggest raising an issue instead of for we are able to discuss if you think that it would be valid. Also, I'd recommend you add the user's cases where it would be helpful as well.

pigletfly:
> Hi @pigletfly.
> 
> Note that, it is a breaking change and we cannot apply to breaking changes in the v2/plugin just for v3-alpha.
> 
> However, IMO we cannot move with this pr indeed for v3-alpha. See that kubebuilder creates a default namespace to deploy the manager which will be `projectName + "-system"`. Your change here will cause problems when users try to deploy more than one project built with the tool in the same cluster.
> 
> Also, shows not a good idea deploy by default any operator in `kube-system` too.
> 
> /hold

Agreed.It's better to use `projectName + "-system"`. The reason is described at https://github.com/kubernetes-sigs/kubebuilder/issues/1613

camilamacedo86:
Hi @pigletfly, 


It is great that you agreed. So, I am closing this one. however, feel free to re-open if you wish. 
And then, we can be working on in the #1613. See my comments there. 

/close

PR #1603, opened 2020-07-22
config: add `project-name` config key and optional `init --project-name`

estroz:
This PR adds a `project-name` config key and an optional `init --project-name` to the v3-alpha plugin. Plugin authors for non-Go plugins will find such a key useful, as `repo` doesn't make much sense for non-Go plugins. This reverts #1596 behavior.

Changes:
* pkg/model/config: add 'project-name' which configures a project's name
* pkg/plugin/v3/init.go: add --project-name which sets a config's 'project-name' key

/cc @pwittrock @camilamacedo86

### DIFF
--- a/pkg/model/config/config.go
+++ b/pkg/model/config/config.go
@@ -41,6 +41,9 @@ type Config struct {
 	// Repo is the go package name of the project root
 	Repo string `json:"repo,omitempty"`
 
+	// ProjectName is the name of this controller project set on initialization.
+	ProjectName string `json:"project-name,omitempty"`
+
 	// Resources tracks scaffolded resources in the project
 	// This info is tracked only in project with version 2
 	Resources []GVK `json:"resources,omitempty"`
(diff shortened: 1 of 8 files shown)


estroz:
Holding for discussion

/hold

pwittrock:
Thanks!

pwittrock:
LGTM

pkg/model/config/config.go
```diff
@@ -41,6 +41,9 @@ type Config struct {
 	// Repo is the go package name of the project root
 	Repo string `json:"repo,omitempty"`
 
+	// ProjectName is the name of this controller project set on initialization.
+	ProjectName string `json:"project-name,omitempty"`
```
joelanford, line 45: Just a nit here:

Since Kubernetes YAMLs usually use `camelCase`, should we make this `projectName`?

pwittrock: probably :) @estroz want to follow up with that.

estroz: #1609

PR #1609, opened 2020-07-23
plugins: change project name config tag, add `pkg/model/file.ProjectNameMixin`

estroz:
This PR adds `pkg/model/file.ProjectNameMixin`, which will inject the `Config.ProjectName` into any structs that embed this mixin.

/cc @joelanford @pwittrock

### DIFF
--- a/pkg/model/file/mixins.go
+++ b/pkg/model/file/mixins.go
@@ -128,3 +128,15 @@ func (m *ResourceMixin) InjectResource(res *resource.Resource) {
 		m.Resource = res
 	}
 }
+
+// ProjectNameMixin provides templates with an injectable project name field.
+type ProjectNameMixin struct {
+	ProjectName string
+}
+
+// InjectProjectName implements HasProjectName.
+func (m *ProjectNameMixin) InjectProjectName(projectName string) {
+	if m.ProjectName == "" {
+		m.ProjectName = projectName
+	}
+}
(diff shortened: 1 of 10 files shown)


pkg/model/file/mixins.go
```diff
@@ -128,3 +130,15 @@ func (m *ResourceMixin) InjectResource(res *resource.Resource) {
...
+
+// ProjectNameMixin provides templates with an injectable project name field.
+type ProjectNameMixin struct {
+	ProjectName string
+}
+
+// InjectProjectName implements HasProjectName.
+func (m *ProjectNameMixin) InjectProjectName(projectName string) {
+	if m.ProjectName == "" {
+		m.ProjectName = strings.ToLower(projectName)
```
joelanford, line 142: Do we really want to be calling `strings.ToLower` at this point? This might be confusing to someone seeing `projectName: MyProject` in their `PROJECT` file and then seeing `myproject` in their scaffolded files.

joelanford: Do we have any concept of `PROJECT` file validation anywhere?

estroz: I'm fine with always lowercasing `projectName`.

estroz: >Do we have any concept of PROJECT file validation anywhere

No. This file is assumed to be more or less "valid" after `init`.

PR #1608, opened 2020-07-23
feat: add script in the repo to setup envtest

camilamacedo86:
This pr is just for we add the script that will be used in the solution to set up the env test locally for we are able to close #1599. See:

```
camilamacedo@Camilas-MacBook-Pro ~/go/src/sigs.k8s.io/kubebuilder (only-script) $ rm -rf testbin/
camilamacedo@Camilas-MacBook-Pro ~/go/src/sigs.k8s.io/kubebuilder (only-script) $ ./scripts/setup_envtest_bins.sh v1.18.2 v3.4.3
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 26.6M  100 26.6M    0     0  6759k      0  0:00:04  0:00:04 --:--:-- 6760k
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 47.8M  100 47.8M    0     0  7932k      0  0:00:06  0:00:06 --:--:-- 8369k
./testbin/kube-apiserver: replacing existing signature
./testbin/kube-apiserver: signed Mach-O thin (x86_64) [kube-apiserver-b9be6305b853bc12b0bf0ab6cf1b219fc4927cad]

```

PR #1600, opened 2020-07-19
WIP: ignore

camilamacedo86:
WIP

PR #1610, opened 2020-07-24
fix: script call for linux

camilamacedo86:
**decription**
- small fix regards issue faced running the script in the Linux
- add a test  to check its execution in both platforms

Related to the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/1608

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ os:
   - osx
 
 go:
-  - "1.13"
+  - 1.13.x
 
 git:
   depth: 3
--- a/test.sh
+++ b/test.sh
@@ -152,4 +152,7 @@ GO111MODULE=on test_project project-v3 3-alpha
 GO111MODULE=on test_project project-v3-multigroup 3-alpha
 GO111MODULE=on test_project project-v3-addon 3-alpha
 
+# test script that setup envtest
+./scripts/setup_envtest_bins.sh v1.18.2 v3.4.3
+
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -10,7 +10,7 @@ os:
   - osx
 
 go:
-  - "1.13"
+  - 1.13.x
```
camilamacedo86, line 13: To test we need 1.13.4 or greater

camilamacedo86:
/test pull-kubebuilder-e2e-k8s-1-15-3

/test pull-kubebuilder-e2e-k8s-1-17-0
--

test.sh
```diff
@@ -152,4 +152,7 @@ GO111MODULE=on test_project project-v3 3-alpha
 GO111MODULE=on test_project project-v3-multigroup 3-alpha
 GO111MODULE=on test_project project-v3-addon 3-alpha
 
+# test script that setup envtest
+./scripts/setup_envtest_bins.sh v1.18.2 v3.4.3
```
djzager, line 156: I'm struggling to understand why we are installing the envtest binaries right before exiting the test.

camilamacedo86: It just for we ensure that the scripts are working in both platform. 
The idea is to remove it in the follow-up https://github.com/kubernetes-sigs/kubebuilder/pull/1600 where they will be called for each testadata sample via the makefile target instead.

Issue #1599, opened 2020-07-19
unable to run the tests scaffold for the projects without the kb bin in /usr/local/kubebuilder/bin


camilamacedo86:
The following error is faced when the KB is installed from master or its bin be added in another place when the tests are executed via the makefile targets:

```
  Unexpected error:
      <*fmt.wrapError | 0xc000109ec0>: {
          msg: "failed to start the controlplane. retried 5 times: fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory",
          err: {
              Op: "fork/exec",
              Path: "/usr/local/kubebuilder/bin/etcd",
              Err: 0x2,
          },
      }
      failed to start the controlplane. retried 5 times: fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory
  occurred
```

Root cause explained in: https://github.com/kubernetes-sigs/kubebuilder/issues/686#issuecomment-647376924

camilamacedo86:
An possible solution is to setup the bins as: 

```
OS=$(uname -s | tr '[:upper:]' '[:lower:]')
ARCH=$(uname -m | sed 's/x86_64/amd64/')
curl -fsL "https://storage.googleapis.com/kubebuilder-tools/kubebuilder-tools-1.16.4-${OS}-${ARCH}.tar.gz" -o kubebuilder-tools
tar -zvxf kubebuilder-tools
sudo mv kubebuilder/ /usr/local/kubebuilder
```

georgettica:
if these files are required, integrating into `brew` is also helpful

camilamacedo86:
Hi @georgettica, 

If you run brew install kubebuilder it will install the binaries because they are shipped with Kubebuilder currently. The goal here is in the long term we do not need to ship them.

It was solved for v3+. See: https://github.com/kubernetes-sigs/kubebuilder/pull/1626 

So, when we are able to removed the support for v2 and removed it from the projects as was done for v1 it will be solved.

georgettica:
@camilamacedo86 I installed kubebuilder and they were not installed

Issue #686, opened 2019-05-07
Release kubebuilder and k8s bundles separately


mengqiy:
kubebuilder code doesn't have dependencies on k8s, since it's mostly scaffolding.
What have dependencies on k8s are controller-runtime and controller-tools. And both are not managed by kubebuilder.
Thus, IMO it's more reasonable to release kubebuilder and k8s bundle (i.e. kubeapiserver, kubectl and etcd) separately. We can have some mechanic to let the user to choose what k8s bundle to use with kubebuilder.

- [x] merge https://github.com/kubernetes-sigs/kubebuilder/pull/685
- [x] move `test` to `testData` (part of #682), so go doesn't evaluate it to calculate deps.
- [ ] develop mechanic to let the user easily choose a k8s bundle version and install it
- [ ] change our release tools to support release kubebuilder and k8s bundle separately.

mengqiy:
cc: @droot @DirectXMan12

timothysc:
/assign @DirectXMan12  
/cc @pwittrock @detiber @ncdc 

So we (Cluster API) have a transitive dependency problem due to kube-builder deps.  Kube builder forces us to take a specific version of kubernetes dependencies that we ideally would like to have abstracted.  This causes a bunch of issues when trying to vendor in other Types whose versions may be newer. 

This is a pretty big problem for us and would love for it to be fixed.

DirectXMan12:
@timothysc this issue is mostly about testenv bundles, not about dependencies.

camilamacedo86:
It was requested here a well: https://github.com/kubernetes-sigs/kubebuilder/issues/1446

DirectXMan12:
The remaining things to do here are to rev the release process and maybe add a little helper to the kb binary to download different deps.  We already release a standalone copy of the deps, so we just need to to tweak the main process.

DirectXMan12:
For what I'd look for in the download deps: list from GCS bucket's XML response, convert to version when possible, falling back to straight names in case we change the format, download to location specified in the `KUBEBUILDER_ASSETS` env variable, potentially with a separate folder per version.

DirectXMan12:
don't do this automatically -- have a separate command.  Do warn on `kb init` if we can't find them.  Also, I'd probably say we should still ship a combined bundle too, just to make things easy, but I could be convinced otherwise.

camilamacedo86:
Hi @DirectXMan12, 

[...]

PR #1626, opened 2020-08-05
feat: envtest setup via target with docs (v3 only)

camilamacedo86:
**Description**
- Add envtest setup in the Makefile
- Add more info over envtest configuration to the docs
- Update te quick start since the download the binaries manually is no longer required for v3+ projects with. 
- remove the script `scripts/setup_envtest_bins.sh` added in https://github.com/kubernetes-sigs/kubebuilder/pull/1608 in favor to use the https://github.com/kubernetes-sigs/controller-runtime/blob/master/hack/setup-envtest.sh #1092

**Motivation**
- We are able to Release kubebuilder and k8s bundles separately #686 (*)
- Make the process easier since no longer will be required to ask to:

```bash
# move to a long-term location and put it on your path
# (you'll need to set the KUBEBUILDER_ASSETS env var if you put it somewhere else)
sudo mv /tmp/kubebuilder_2.3.1_${os}_${arch} /usr/local/kubebuilder
```
- Keep the bins inside of the project instead of `/usr/local/` since it is only used by it. (IMO not good requires by default root permission too to follow the quickstart )
- Solve the issue faced when the installation was done via master and the above steps were not performed.  https://github.com/kubernetes-sigs/kubebuilder/issues/1599 (*)

(*) It will just be 100% solved when the v2 is no longer supported since it shows that would be a breaking change for v2.


camilamacedo86:
c/c @vincepri

docs/book/src/quick-start.md
```diff
@@ -25,19 +25,25 @@ arch=$(go env GOARCH)
 
 # download kubebuilder and extract it to tmp
 curl -L https://go.kubebuilder.io/dl/2.3.1/${os}/${arch} | tar -xz -C /tmp/
+```
+
+<aside class="note">
+<h1>(Only V2+) Configuring envtest</h1>
```
gabbifish, line 31: Are these steps for the configuration of envtest? It looks like the text that immediately follows is focused on setting up kubebuilder, not installing/running envtest. Perhaps you meant to link to the .envtest.md page here?

If we're going to put a link to envtest docs here, we also might want to move it to the bottom of the the document (where I think it would fit better, given that this early in the quickstart no-one has things to test yet!).

docs/book/src/quick-start.md
```diff
@@ -25,19 +25,25 @@ arch=$(go env GOARCH)
...
 ```
+</aside>
 
 <aside class="note">
 <h1>Using master branch</h1>
 
 Also, you can install a master snapshot from `https://go.kubebuilder.io/dl/latest/${os}/${arch}`.
-
-
+s
```
estroz, line 46:
```suggestion
```

docs/book/src/reference/envtest.md
```diff
@@ -25,7 +25,12 @@ err = testEnv.Stop()
 Logs from the test runs are prefixed with `test-env`.
 
 ### Configuring your test control plane
-You can use environment variables and/or flags to specify the `api-server` and `etcd` setup within your integration tests.
+
+The controller-runtime’s [envtest](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/envtest) framework requires `kubectl`, `kube-apiserver`, and `etcd` binaries to simulate the API portions of a real cluster. 
+
+For projects built with `v3+`, these binaries and configuration is done via the `test` makefile target. It will create the `testbin/` directory where the binaries will be downloaded.
```
estroz, line 31:
```suggestion
For projects built with plugin v3+ (see your PROJECT file's `layout` key), the `make test` command will install these binaries to the `testbin/` directory and use them when running tests that use `envtest`.
```

docs/book/src/reference/envtest.md
```diff
@@ -25,7 +25,12 @@ err = testEnv.Stop()
 Logs from the test runs are prefixed with `test-env`.
 
 ### Configuring your test control plane
-You can use environment variables and/or flags to specify the `api-server` and `etcd` setup within your integration tests.
+
+The controller-runtime’s [envtest](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/envtest) framework requires `kubectl`, `kube-apiserver`, and `etcd` binaries to simulate the API portions of a real cluster.
```
estroz, line 29:
```suggestion
controller-runtime’s [envtest](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/envtest) framework requires `kubectl`, `kube-apiserver`, and `etcd` binaries be present locally to simulate the API portions of a real cluster. 
```

docs/book/src/reference/envtest.md
```diff
@@ -25,7 +25,12 @@ err = testEnv.Stop()
 Logs from the test runs are prefixed with `test-env`.
 
 ### Configuring your test control plane
-You can use environment variables and/or flags to specify the `api-server` and `etcd` setup within your integration tests.
+
+The controller-runtime’s [envtest](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/envtest) framework requires `kubectl`, `kube-apiserver`, and `etcd` binaries to simulate the API portions of a real cluster. 
+
+For projects built with `v3+`, these binaries and configuration is done via the `test` makefile target. It will create the `testbin/` directory where the binaries will be downloaded.  
+
+You can use environment variables and/or flags to specify the `kubectl`,`api-server` and `etcd` setup within your integration tests.
```
estroz, line 33: Link needed.

pkg/plugin/v3/scaffolds/internal/templates/makefile.go
```diff
@@ -70,8 +70,11 @@ endif
 all: manager
 
 # Run tests
+ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
 test: generate fmt vet manifests
-	go test -race -coverprofile cover.out ./...
+	mkdir -p ${ENVTEST_ASSETS_DIR}
+	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/master/hack/setup-envtest.sh
```
estroz, line 76: This link should point to the controller-runtime version used by the scaffolded project, so can you add a TODO to version this script on the next controller-runtime release?

docs/book/src/quick-start.md
estroz: `make docker-build` will invoke `make test`, so perhaps move this to the bottom and add a link in the `Run It On the Cluster` section.

estroz:
```suggestion
<h1>(Plugin versions > v2) Configuring envtest</h1>
```

docs/book/src/quick-start.md
```diff
@@ -25,19 +25,25 @@ arch=$(go env GOARCH)
 
 # download kubebuilder and extract it to tmp
 curl -L https://go.kubebuilder.io/dl/2.3.1/${os}/${arch} | tar -xz -C /tmp/
+```
+
+<aside class="note">
+<h1>(Only V2+) Configuring envtest</h1>
 
+If you are using v3+ it is no longer required. Otherwise, run:
```
estroz, line 33: This is implied by the title.
```suggestion
```

docs/book/src/quick-start.md
camilamacedo86: Hi @gabbifish, 

This steps in the installation are to set up the required binaries in the /user/local/kubebuilder path and will no longer be required with this PR for plugin versions v3+. I added the link as suggested and tried to make it better.  

HI @estroz, 

We cannot move it because this steps are part of the installation see that we are moving the binaries downloaded with the kubebuilder.

docs/book/src/reference/envtest.md
camilamacedo86: Hi @estroz, 

I did not get what link should be added here? This text was not changed at all and it is added because bellow it has the envvar/flags details.

pkg/plugin/v3/scaffolds/init.go
```diff
@@ -40,6 +40,9 @@ const (
 	ControllerToolsVersion = "v0.3.0"
 	// KustomizeVersion is the kubernetes-sigs/kustomize version to be used in the project
 	KustomizeVersion = "v3.5.4"
+	// todo: update the tag release when the next version of the project be released with this script
```
camilamacedo86, line 43: @estroz your suggestion is applied here ^

docs/book/src/quick-start.md
```diff
@@ -25,19 +25,29 @@ arch=$(go env GOARCH)
...
 # move to a long-term location and put it on your path
 # (you'll need to set the KUBEBUILDER_ASSETS env var if you put it somewhere else)
 sudo mv /tmp/kubebuilder_2.3.1_${os}_${arch} /usr/local/kubebuilder
 export PATH=$PATH:/usr/local/kubebuilder/bin
 ```
 
+<aside class="note">
+<h1>Kubernetes binaries</h1>
+
+The above setup is no longer required if you are using plugin version `v3+.
```
gabbifish, line 42: With the comment above, we could probably delete this new section.

docs/book/src/reference/envtest.md
```diff
@@ -37,6 +42,33 @@ You can use environment variables and/or flags to specify the `api-server` and `
...
+var _ = BeforeSuite(func(done Done) {
+	Expect(os.Setenv("TEST_ASSET_KUBE_APISERVER", "../testbin/bin/kube-apiserver")).To(Succeed())
+	Expect(os.Setenv("TEST_ASSET_ETCD", "../testbin/bin/etcd")).To(Succeed())
+	Expect(os.Setenv("TEST_ASSET_KUBECTL", "../testbin/bin/kubectl")).To(Succeed())
+
+	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
+	testenv = &envtest.Environment{}
+
+	var err error
+	cfg, err = testenv.Start()
```
gabbifish, line 57: nit: how do you feel about using the following instead, it's slightly less verbose
```suggestion
	cfg, err := testenv.Start()
```

docs/book/src/quick-start.md
```diff
@@ -25,19 +25,29 @@ arch=$(go env GOARCH)
 
 # download kubebuilder and extract it to tmp
 curl -L https://go.kubebuilder.io/dl/2.3.1/${os}/${arch} | tar -xz -C /tmp/
+```
+
+You must configure the Kubernetes binaries required for the [envtest][envtest], run:
```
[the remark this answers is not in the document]
camilamacedo86: The problem is : It is not the KB version. I mean, we have the CLI version an we have the Plugin and Project versions.
 
Also, it is not better to keep the info as it is by default today and have the following note to clarifies? See:

Screen Shot 2020-08-28 at 12 13 18

docs/book/src/quick-start.md
camilamacedo86: I am not sure. I think is better we keep until the v3 alpha plugin version is stable and default at least. Then, we will be able to deprecate the v2 and remove the note as the text above you must run. See that currently, users must run.

docs/book/src/reference/envtest.md
camilamacedo86: good cather.

PR #1629, opened 2020-08-09
using built in finalizer function and fix some syntax

kanolato:
Using GetFinalizers and SetFinalizer methods as they seem more proper.


kanolato:
yes! it is done @camilamacedo86

Issue #1544, opened 2020-06-02
The CustomResourceDefinition "cronjobs.batch.tutorial.kubebuilder.io" is invalid


zehuaiWANG:
I followed the tutorial but the cronjob install failed. I clone the master code, but it seems to not work. Could anyone tell me what to do next?  Thanks a lot.
```
[root@test ~/kubebuilder/docs/book/src/cronjob-tutorial/testdata/project]# kubectl version --short
Client Version: v1.18.0
Server Version: v1.19.0-beta.0.318+b618411f1edb98

[root@test ~/kubebuilder/docs/book/src/cronjob-tutorial/testdata/project]# make install
/usr/local/bin/controller-gen "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths="./api/...;./controllers/..." output:crd:artifacts:config=config/crd/bases
kubectl apply -f config/crd/bases
The CustomResourceDefinition "cronjobs.batch.tutorial.kubebuilder.io" is invalid: 
* spec.validation.openAPIV3Schema.properties[spec].properties[jobTemplate].properties[spec].properties[template].properties[spec].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[jobTemplate].properties[spec].properties[template].properties[spec].properties[containers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
make: *** [install] Error 1
```

/kind bug

zehuaiWANG:
If I use kubernentes 1.16, it won't happen. I guess it's a problem after kubernetes 1.18.

camilamacedo86:
Hi @zehuaiWANG, 

The `/kubebuilder/docs/book/src/cronjob-tutorial/testdata/project` is outdated. I have been working on a PR to update that.  See: https://github.com/kubernetes-sigs/kubebuilder/pull/1492

estroz:
This is likely related to https://github.com/kubernetes-sigs/controller-tools/issues/444.

camilamacedo86:
Hi @estroz, 

See that the user is trying to check the example code in the repo: 

[root@test ~/kubebuilder/docs/book/src/cronjob-tutorial/testdata/project]# kubectl version --short

[...]

PR #1632, opened 2020-08-10
 doc: update testadata projects used in the tutorial

camilamacedo86:
**Description**
- Upgrade the tesdata used in the tutorial for the latest version since it is very outdated
- Add script which generates the base of the project at least for be easier we manually perform this process. 

**Motivation**
- The example is not working and has issues, however, the first step is to update the project for them we are able to address them. The project has many issues raised regards the tutorial testdata. Also, it will show the code in the docs more accurate. 

E.g - https://github.com/kubernetes-sigs/kubebuilder/issues/1251 #1544


fbalicchia:
Hi @camilamacedo86 ,
in general seems that config files are aligned with new versions `kubebuilder 2.3.1` 

Have tried example test and seems that it fails with `namespaces "test-cronjob-namespace" not found.`
Please consider that I'm relatively novice and probably I need to familiarize better with env.
In `Main.go` I've see that to use `utilruntime` to load external schema interesting but seems that last version of kubebuider doesn't use it and in book seems not mentioned

camilamacedo86:
Hi @fbalicchia, 

The samples are updated with the latest version on master for the docs be updated for the next release. See the script : https://github.com/kubernetes-sigs/kubebuilder/pull/1632/files#diff-95f6fb0104f19012b07b292e718bed49R53. 

regards  `namespaces "test-cronjob-namespace"` it shows that you are not using the make commands to deploy it. 

However, these projects still with issues and we have many issues tracked over that. But the first step is to keep them updated with the master for them we are able to check each issue and scenario.

docs/book/src/multiversion-tutorial/testdata/project/main.go
```diff
@@ -20,8 +21,8 @@ import (
 	"flag"
 	"os"
 
-	kbatchv1 "k8s.io/api/batch/v1"
```
gabbifish, line 23: Why is kbatch1 removed here?

docs/book/src/cronjob-tutorial/testdata/generate_cronjob.sh
```diff
@@ -0,0 +1,58 @@
...
+}
+
+function gen_cronjob_tutorial {
+  header_text "removing project ..."
+  rm -rf project
+  header_text "starting to generate the cronjob ..."
+  mkdir project
+  cd project
+  header_text "creating tutorial.kubebuilder.io base  ..."
+  kubebuilder init --plugins=go/v3-alpha --domain=tutorial.kubebuilder.io --project-version=3-alpha --repo=tutorial.kubebuilder.io/project --license apache2 --owner "The Kubernetes authors"
```
gabbifish, line 53: I'm using the latest version of kubebuilder (top of master) and it looks like the `--plugins` flag still doesn't exist here--is that expected / will this come later?

gabbifish (commented):
I'm trying to run `./generate_cronjob.sh` so I can confirm it works as expected, but it looks like the `--plugins` flag for `kubebuilder init` still isn't implemented yet (I set `KUBEBUILDER_ENABLE_PLUGINS=1`, too!).

I'm using `kubebuilder` built from top of the master branch so I'm surprised I'm not finding this `--plugins` flag; maybe there's more I need to configure so I can use this flag and run the testdata generation script?

docs/book/src/cronjob-tutorial/testdata/generate_cronjob.sh
camilamacedo86: It worked when the PR was made. However, now shows that we need to pass the full name of the plugin. I will fix it and also add. the scripts to the Travis. Then we can ensure that it works always. 👍 Tks

camilamacedo86: Hi @gabbifish, 

These scripts are just to allow us re-gen the project inside off this dir and diff with the master to know what requires updates. the plugins only exist on the master branch and by merging this PR we are updating the tutorial with the latest version available which is v3-alpha plugin on the master. It will just be available to production after the next release.

camilamacedo86: IMO what we need to do here is to ensure that we are updating the projects and not losing any text/info. I will mark it with: 

WIP

Because I would like to recheck it too.

docs/book/src/multiversion-tutorial/testdata/project/main.go
camilamacedo86: should not good catcher

camilamacedo86:
Hi @gabbifish,

Really thank you for your help here.  I re-duo everything and re-check this one shows that all is ok. 
Note that the multigroup version is passing in the `make tests` but the cronjob-tutorial not. See: https://github.com/kubernetes-sigs/kubebuilder/issues/1680

I think we will sting having issues to address after this update. However, IMO we need first update the examples since they are very outdated for them be able to try to fix them. What IMO we need to ensure here in this PR? That the examples used in the tutorial were not changed at all and that we are not losing any text/explanation.

docs/book/src/cronjob-tutorial/testdata/generate_cronjob.sh
camilamacedo86: did you run make install before the script? I will add it on the script for we ensure that it will get the latest bin.

gabbifish:
@camilamacedo86 I reviewed the rest of the PR (except for testing out the `--plugins` flag, knowing this will be added/expected later). Everything else looks consistent to me :)

gabbifish (approved):
/lgtm

I still cannot test the generate_multiversion scripts given the lack of support for the `--plugin` flag but it sounds like that is coming shortly? I couldn't use `--plugin` even when I was using the tip of the master branch. I would love to get a chance to test this feature later, but as you mentioned above, this feature is still coming!

gabbifish:
NOO I just found a tiny nit--note the file `docs/book/src/multiversion-tutorial/testdata/generate_mutiversion.sh` should be called `docs/book/src/multiversion-tutorial/testdata/generate_multiversion.sh` 😅 I'll make a PR for this tomorrow morning, it's an easy fix

camilamacedo86:
hi @gabbifish. 

really tks for your help with. done: https://github.com/kubernetes-sigs/kubebuilder/pull/1683

PR #1635, opened 2020-08-11
:sparkles: specify nonroot uid for manager (v3+ only)

harpratap:
PR #983 changes the manager's user to nonroot but it doesn't work out of the box. 

This change modifies the default behaviour of manager pod to use non-root uid for both kube-rbac-proxy container and the manager.

There are two issues when deploying the manager in non-root cluster if we try to use the config/default as-is  

1. kube-rbac-proxy uses root by default instead of a nonroot image - https://github.com/brancz/kube-rbac-proxy/blob/master/Dockerfile#L1 which causes pod to request for root and fail. 

2. Kubernetes cannot verify the `nonroot` user in manager because it is a string (https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/v2/scaffolds/internal/templates/dockerfile.go#L65) resulting in `container has runAsNonRoot and image has non-numeric user (nonroot), cannot verify user is non-root`

### DIFF
--- a/pkg/plugin/v3/scaffolds/internal/templates/dockerfile.go
+++ b/pkg/plugin/v3/scaffolds/internal/templates/dockerfile.go
@@ -62,7 +62,7 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager 
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/manager .
-USER nonroot:nonroot
+USER 65532:65532
 
 ENTRYPOINT ["/manager"]
 `
--- a/testdata/project-v3-addon/config/manager/manager.yaml
+++ b/testdata/project-v3-addon/config/manager/manager.yaml
@@ -22,6 +22,8 @@ spec:
       labels:
         control-plane: controller-manager
     spec:
+      securityContext:
+        runAsUser: 65532
       containers:
       - command:
         - /manager
@@ -29,6 +31,8 @@ spec:
         - --enable-leader-election
         image: controller:latest
         name: manager
+        securityContext:
+          allowPrivilegeEscalation: false
         resources:
           limits:
             cpu: 100m
--- a/pkg/plugin/v3/scaffolds/internal/templates/config/manager/config.go
+++ b/pkg/plugin/v3/scaffolds/internal/templates/config/manager/config.go
@@ -67,6 +67,8 @@ spec:
       labels:
         control-plane: controller-manager
     spec:
+      securityContext:
+        runAsUser: 65532
       containers:
       - command:
         - /manager
@@ -74,6 +76,8 @@ spec:
         - --enable-leader-election
         image: {{ .Image }}
         name: manager
+        securityContext:
+          allowPrivilegeEscalation: false
         resources:
           limits:
             cpu: 100m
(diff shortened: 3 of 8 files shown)


estroz:
@harpratap can you make this change to `pkg/plugin/v3`, and run `./generate_testdata.sh` from the project root

camilamacedo86:
/hold

HI @harpratap, 

+1 for @estroz commands and then: 

Could you please raise an issue with the scenarios faced and ALL step performed to check them? 
I am unable so far to check these issues, also if I understood your scenario, the first one is indeed tested in the e2e so I am not sure if it would be the case. 

Also, after we are able to reproduce the problems faced we need to discuss if it would be the best approach as well.

harpratap:
Sorry about that, I have created an issue - #1637 
Once this is confirmed as an issue and we have a consensus on the correct approach to solve it I will update this PR

pkg/plugin/v2/scaffolds/internal/templates/manager/config.go
```diff
@@ -67,6 +67,8 @@ spec:
       labels:
         control-plane: controller-manager
     spec:
+      securityContext:
+        runAsUser: 1337
```
camilamacedo86, line 71: Could you please do it for v3 only and as it was done for the proxy in https://github.com/brancz/kube-rbac-proxy/pull/86? We need to update the image as well. So the change will be:
- in /pkg/plugin/v3/scaffolds/internal/templates/manager/config.go
- in v3/scaffolds/internal/templates/dockerfile.go
And then, see that you need to run `make generate` to update the samples.

camilamacedo86 (changes requested):
It is great. could you please check the comment

pkg/plugin/v2/scaffolds/internal/templates/manager/config.go
harpratap: Okay, updated the PR

pkg/plugin/v3/scaffolds/internal/templates/dockerfile.go
```diff
@@ -62,7 +62,7 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/manager .
-USER nonroot:nonroot
+USER 65532:65532
```
camilamacedo86, line 65:
```suggestion
USER 1001:1001
```
Could we use this ID?

testdata/project-v3-addon/config/manager/manager.yaml
```diff
@@ -22,6 +22,8 @@ spec:
       labels:
         control-plane: controller-manager
     spec:
+      securityContext:
+        runAsUser: 65532
       containers:
       - command:
```
camilamacedo86, line 28: Could we add in the manifests `allowPrivilegeEscalation: false`?

harpratap: Done

camilamacedo86:
hi @harpratap, 

It shows great. 🥇 You just need to run `make generate` for all samples in the testdata/ get properly update. Also, it is good ensure that it is rebased with the master.   It needs pass in the CI for we are able to merge. 

Also, in the title of the issue could you please add the mojo? See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#pr-process

camilamacedo86 (approved):
/approve

Great work 👍

estroz (approved):
/lgtm

PR #1641, opened 2020-08-17
:book: fix dark theme, update mdbook, 404 page

DirectXMan12:
This updates mdbook to the latest binary, fixes handling of devices that perfer dark theme (which were previously broken due to a mismatch between the version our index.hbs file was based on and the binary we were using), and adds a custom 404 page.

/kind documentation


DirectXMan12:
Manually confirmed custom 404 works on netlify

PR #1642, opened 2020-08-19
Match the port 9443 to match in main.go

jameszh:
For the cronjob tutorial, match targetPort to what specified in main.go


jameszh:
just signed a CLA

PR #1643, opened 2020-08-19
Fix path to generated main.go 

hudymi:
Fixed path to the generated `main.go` file in `init` command description. I'm not sure about `v3`, because as I can see it will be possible to define the project's layout, so I can revert it.


camilamacedo86 (approved):
/approve

PR #1649, opened 2020-08-22
dev(dx): lint improvements (use github actions to call it)

camilamacedo86:
**Description**
- Add makefile targets to run lint and to install the version used locally
- Add makefile targets to run the lint --fix
- Add GitHub Action to call it and do the CI integration. 
- Update the Contribution doc with its information

**Motivation**
- A better way to check to linting the project in the CI which allows automatically comments be done (pre-code review)

screen-shot-2020-08-21-at-17 19 57


camilamacedo86:

Any idea over why GitHub actions are not called in this repo? could it be any permission/role configuration in the repo which is blocking it?

gabbifish:
@camilamacedo86 unfortunately, I think your action definition might need to be merged into master before it can be run/tested--see https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#creating-a-workflow-file (step 4 implies we can't actually evaluate how the action works until it has been merged...)

camilamacedo86:
Hi @gabbifish, 

We can try to merge and check if it will work since we are not removing the check from the lint here. For the other projects it shows not required at all. However, they had GA already too.  

/test pull-kubebuilder-e2e-k8s-1-17-0

PR #1656, opened 2020-08-28
Direct link to the CronJob tutorial source code

vladimirvivien:
This is to add a link to the CronJob tutorial discussed in the book in a 
more obvious location. Currently, it is not obvious where the code resides.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Check out the Kubebuilder [book](https://book.kubebuilder.io).
 
 - Kubebuilder Book: [book.kubebuilder.io](https://book.kubebuilder.io)
 - GitHub Repo: [kubernetes-sigs/kubebuilder](https://github.com/kubernetes-sigs/kubebuilder)
+- Tutorial: [CronJob tutorial source code](https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/cronjob-tutorial/testdata/project)
 - Slack channel: [#kubebuilder](https://slack.k8s.io/#kubebuilder)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,6 +28,7 @@ Check out the Kubebuilder [book](https://book.kubebuilder.io).
 
 - Kubebuilder Book: [book.kubebuilder.io](https://book.kubebuilder.io)
 - GitHub Repo: [kubernetes-sigs/kubebuilder](https://github.com/kubernetes-sigs/kubebuilder)
+- Tutorial: [CronJob tutorial source code](https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/cronjob-tutorial/testdata/project)
```
camilamacedo86, line 31: I am not sure if we should add it because the master branch not necessary will have the same version of the doc. 
See that the docs in production are from another branch.

gabbifish: Perhaps it would be more useful to have the docs point to the example code itself, if they don't already? And modify the label above to read `Kubebuilder Book and Tutorial` so folks know they can find the tutorial within the book?

camilamacedo86: Hi @gabbifish, 

Currently the testdata is outdated and it has not been updating to match with the version/doc. So, it might bring confusion. Also, we cannot point to the master branch. We would need to work with the tags releases instead of it.

vladimirvivien:
It sounds like this is more involved than anticipated.  Reason I open this because when reading the doc it was not obvious (at a glance) where to link to the example (from the doc). Maybe it's a matter of updating the doc when testdata is updated.  So closing this for now.

Issue #1663, opened 2020-09-07
Conversion webhook scaffolding (v3-alpha plugin)


varshaprasad96:
Currently, Kubebuilder supports scaffolding for validating/defaulting Webhooks. While using `--kubebuilder create webhook` when defaulting or validating option is specified, the project scaffolds  `$kind_webhook.go`, which contains the necessary method signatures directing the user to implement them. Whereas when `--conversion` option is specified, the cli only prints out a message stating necessary interfaces which are to be implemented.

**What do you want to happen?**

Kubebuilder should scaffold conversion webhook implementation containing templates directing the users to complete `Convertible` interface.

For example:
When `kubebuilder create webhook  --group crew --version v1 --kind FirstKind --conversion` is used, I should be able to specify the Hub and Spoke version, after which 
 1. `$kind_conversion.go` scaffolds inside right `api/version` folders
2. `conversion.Hub` is implemented
3. User is directed to fill up the required methods for `conversion.Convertible` interface.

/kind feature

PR #1664, opened 2020-09-07
(:sparkles:)Scaffold conversion webhook template (v3-alpha plugin)

varshaprasad96:
This PR adds scaffolding when conversion webhook is specified in the
`kubebuilder create webhook` command of kubebuilder.

Example usage:
`kubebuilder create webhook  --group crew --version v1 --kind FirstKind --conversion --hub v1 --spoke v2`

1. FirstKind_conversion.go scaffolded inside `api/v1/firstmate_webhook.go` having [`conversion.Hub`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/conversion/conversion.go#L37) interface.
2. FirstKind_conversion.go scaffolded inside `api/v2/firstmate_webhook.go` having method signatures to implement [`conversion.convertible`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/conversion/conversion.go#L27) interface.
3. Webhook registered in `main.go`

Related to issue: #1663 

c/c: @estroz


pkg/plugin/v3/webhook.go
```diff
@@ -96,6 +100,15 @@ func (p *createWebhookPlugin) Validate() error {
...
 
+	if p.conversion {
+		if p.spoke == "" || p.hub == "" {
+			return fmt.Errorf("Hub and Spoke version need to be specified when using conversion webhooks")
+		}
+		if p.spoke == p.hub {
+			return fmt.Errorf("Cannot have the same hub and spoke version")
+		}
+	}
+
```
camilamacedo86, line 111: Why should we make mandatory inform them?

go.mod
```diff
@@ -4,8 +4,8 @@ go 1.13
 
 require (
 	github.com/gobuffalo/flect v0.2.1
-	github.com/onsi/ginkgo v1.12.0
-	github.com/onsi/gomega v1.9.0
+	github.com/onsi/ginkgo v1.12.1
+	github.com/onsi/gomega v1.10.1
 	github.com/spf13/afero v1.2.2
```
camilamacedo86, line 9: It shows not be part of the scope of this pr. Could you please revert?

camilamacedo86 (changes requested):
Hi @varshaprasad96, 

Really nice contribution 🥇  just a few points. Please, check the comments:
- It is not passing in the check testdata because you add the args as mandatory. See that you will need change the script that generates the samples in the testdata: https://github.com/kubernetes-sigs/kubebuilder/blob/master/generate_testdata.sh 
- It is missing be covered by tests. 
- Also, we need to update its docs

pkg/plugin/v3/webhook.go
varshaprasad96: Because it depends on the user on what version they want to have as the hub. Without it, I think it wouldn't be fair to randomly chose a hub version and a non-storage version.

camilamacedo86: Can we keep the current behaviour when these values are not informed?

go.mod
```diff
@@ -4,11 +4,14 @@ go 1.13
...
-	github.com/spf13/cobra v0.0.7
+	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/tools v0.0.0-20200403190813-44a64ad78b9b
+	golang.org/x/tools v0.0.0-20200714190737-9048b464a08d
+	k8s.io/apimachinery v0.19.2 // indirect
+	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200816135617-dbfe418e405f // indirect
+	sigs.k8s.io/kubebuilder/testdata/project-v2-addon v0.0.0-20200918110645-1b5953e6cc43 // indirect
```
camilamacedo86, line 15: could you please remove the indirect ones?

pkg/plugin/v3/webhook.go
```diff
@@ -96,6 +100,19 @@ func (p *createWebhookPlugin) Validate() error {
 			" --programmatic-validation and --conversion to be true", p.commandName)
 	}
 
+	if p.conversion {
+		if (p.spoke == "" && p.hub != "") || (p.hub == "" && p.spoke != "") {
+			return fmt.Errorf("Hub and Spoke version need to be specified when using conversion webhooks")
+		}
+		if p.spoke == "" || p.hub == "" {
+			fmt.Println("Hub and Spoke version can be specified in the command to scaffold Convertible and Hub interfaces")
```
camilamacedo86, line 108: Is it a warn/info? I do not think that it is required,

pkg/plugin/v3/webhook.go
```diff
@@ -96,6 +100,19 @@ func (p *createWebhookPlugin) Validate() error {
...
+			return nil
+		}
+		if p.spoke == p.hub {
+			return fmt.Errorf("Cannot have the same hub and spoke version")
+		}
+	}
+
```
camilamacedo86, line 115: I think we need to start to cover the code with tests. WDYT about add a webhook_test here to check these scenarios?

pkg/plugin/v3/scaffolds/internal/templates/config/api/spoke_conversion.go
```diff
@@ -0,0 +1,73 @@
...
+package {{ .Resource.Version }}
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/conversion"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+
+func (src  *{{ .Resource.Kind }}) ConvertTo(dstRaw conversion.Hub) error {
+	// Implement logic to convert from hub to spoke version.
```
camilamacedo86, line 63:
```suggestion
	// Implement your logic here to convert from hub to spoke version.
```

camilamacedo86: To be similar to: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/controllers/admiral_controller.go#L44

pkg/plugin/v3/scaffolds/internal/templates/config/api/spoke_conversion.go
```diff
@@ -0,0 +1,73 @@
...
+	
+	return nil
+}
+
+func (dst *{{ .Resource.Kind }}) ConvertFrom(srcRaw conversion.Hub) error {
+	// Implement logic to convert from spoke to hub version.
```
camilamacedo86, line 69:
```suggestion
	// Implement your logic here to convert from spoke to hub version.
```

pkg/plugin/v3/scaffolds/internal/templates/config/api/spoke_conversion.go
```diff
@@ -0,0 +1,73 @@
...
+
+const spokeConversionTemplate = `{{ .Boilerplate }}
+
+package {{ .Resource.Version }}
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/conversion"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
camilamacedo86, line 60: hav you check this in any file scaffolded? is this an standard in the project?

camilamacedo86 (changes requested):
Really great contribution. Following a few things that needs to be done here:
- [ ] the title of the pr requires the mojo. See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#pr-process
- [ ] revert the changes in the go mod that shows not be part of its scope. 
- [ ] the file generate_testdata.sh needs to be a supplement to gen in the tests data this scenario as well for the v3-projects. 
- [ ] we need to update the docs to describe the new options and how it works. 
- [ ] could we add this scenario to the e2e tests? 
- [ ] some nits in the code
- [ ] rebase with master

pkg/plugin/v3/scaffolds/internal/templates/config/api/hub_conversion.go
```diff
@@ -0,0 +1,68 @@
...
+	c.Resource.Version = c.HubVersion
+	if c.Path == "" {
+		if c.MultiGroup {
+			c.Path = filepath.Join("apis", "%[group]", "%[version]", "%[kind]_conversion.go")
+		} else {
+			c.Path = filepath.Join("api", "%[version]", "%[kind]_conversion.go")
+		}
+	}
+	c.Path = c.Resource.Replacer().Replace(c.Path)
+	fmt.Printf("Scaffolding conversion logic for hub version %s in path %s", c.Resource.Version, c.Path)
```
camilamacedo86, line 47: IHMO we should not add prints here. Have we any other boilerplate printing. the actions?
If we would like to log out the action then, I think it needs to be in the plugin which is reasonable for the execution. 
The boilerplate just define the template which will be executed. so, I do not think that it is part of its domain of responsibility.

pkg/plugin/v3/scaffolds/internal/templates/config/api/hub_conversion.go
```diff
@@ -0,0 +1,68 @@
...
+const conversionTemplate = `{{ .Boilerplate }}
+
+package {{ .Resource.Version }}
+
+/*
+Implementing the hub method is pretty easy -- we just have to add an empty
+method called 'Hub()' to serve as a
+[marker](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/conversion?tab=doc#Hub).
+We could also just put this inline in our 'cronjob_types.go' file.
+*/
```
camilamacedo86, line 63: Should we scaffold the project with or it should be in the docs?

pkg/plugin/v3/webhook.go
varshaprasad96: This is an additional info informing users that the feature is available for them to set.

prafull01:
Once this PR gets merged and  conversion tests are supported in envtest ([PR here](https://github.com/kubernetes-sigs/kubebuilder/pull/1710/files#r500513284)), the test suite scaffolding is also needs to be done. For More details: [https://github.com/kubernetes-sigs/kubebuilder/pull/1710/files#r500513284](url)

pkg/plugin/v3/scaffolds/internal/templates/config/api/spoke_conversion.go
varshaprasad96: This appears we scaffold admission/mutating WebHooks too. So I had added it here.

pkg/plugin/v3/scaffolds/internal/templates/config/api/hub_conversion.go
varshaprasad96: Had added this, because webhook conversion scaffolding prints the path: [here](https://github.com/kubernetes-sigs/kubebuilder/blob/49f7a5eb70f85034473ff18c85827337f3c7d9a9/pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook.go#L59). Shall I remove it or still have it ? Im fine with removing it too.

pkg/plugin/v3/scaffolds/internal/templates/config/api/hub_conversion.go
varshaprasad96: This is as present in the example conversion WH operator - https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/multiversion-tutorial/testdata/project/api/v1/cronjob_conversion.go#L19-L24. Do we have it or shall I remove it?

pkg/plugin/v3/webhook.go
camilamacedo86: IMO we ough to not use the `fmt.Println` to print WARN/INFO logs for the tool. We can use either logr with zapr, or klog v2

pkg/plugin/v3/scaffolds/internal/templates/config/api/hub_conversion.go
```diff
@@ -0,0 +1,68 @@
...
+
+// Hub marks this type as a conversion hub.
+func (*{{ .Resource.Kind }}) Hub() {}
```
estroz, line 66: This doesn't necessarily need to be a pointer since this method does nothing.
```suggestion
func ({{ .Resource.Kind }}) Hub() {}
```

pkg/plugin/v3/webhook.go
```diff
@@ -96,6 +100,19 @@ func (p *createWebhookPlugin) Validate() error {
 			" --programmatic-validation and --conversion to be true", p.commandName)
 	}
 
+	if p.conversion {
+		if (p.spoke == "" && p.hub != "") || (p.hub == "" && p.spoke != "") {
+			return fmt.Errorf("Hub and Spoke version need to be specified when using conversion webhooks")
+		}
+		if p.spoke == "" || p.hub == "" {
```
estroz, line 107: This condition encompasses the above one logically, can you combine them?

pkg/plugin/v3/webhook.go
```diff
@@ -76,6 +78,8 @@ func (p *createWebhookPlugin) BindFlags(fs *pflag.FlagSet) {
 		"if set, scaffold the validating webhook")
 	fs.BoolVar(&p.conversion, "conversion", false,
 		"if set, scaffold the conversion webhook")
+	fs.StringVar(&p.spoke, "spoke", "", "provide a hub version for the conversion webhook")
+	fs.StringVar(&p.hub, "hub", "", "provide the non-hub (spoke) verision for conversion webhook")
```
estroz, line 82: Can we simplify these flags by removing `--conversion`, and treating setting `conversion: true` as `--hub --spoke`, seeing as:
- `--hub --spoke` always requires `--conversion` always be set
- there's no point in setting `--conversion` without `--hub --spoke`

pkg/plugin/v3/scaffolds/webhook.go
```diff
@@ -47,14 +49,18 @@ func NewWebhookScaffolder(
 	defaulting bool,
 	validation bool,
 	conversion bool,
+	spokeVersion string,
```
estroz, line 52:
```suggestion
	webhookSpokeVersion string,
```
and ditto below.

estroz (changes requested):
Mostly nits, one major suggestion regarding flag consolidation.

varshaprasad96:
@camilamacedo86  To include test for conversion WH in e2e, we may still need https://github.com/kubernetes-sigs/kubebuilder/pull/1644, because with v1beta1 crds, we still get an error: Invalid value: "Webhook": must be None if spec.preserveUnknownFields is true
The workaround would be to modify and set this separately, because including this in CRD_OPTIONS does not work either. Shall I implement the workaround in this test, or wait for the other PR to get merged and add test for this feature in the follow up (since in this PR we would only be adding the scaffolding) ?

camilamacedo86:
I think is better to wait for #1644. Then, we can ensure that it works. Otherwise, will not the users face the same?

test/e2e/utils/test_context.go
```diff
@@ -66,14 +67,15 @@ func NewTestContext(binaryName string, env ...string) (*TestContext, error) {
...
-		TestSuffix: testSuffix,
-		Domain:     "example.com" + testSuffix,
-		Group:      testGroup,
-		Version:    "v1alpha1",
-		Kind:       "Foo" + testSuffix,
-		Resources:  "foo" + testSuffix + "s",
-		ImageName:  "e2e-test/controller-manager:" + testSuffix,
-		CmdContext: cc,
+		TestSuffix:    testSuffix,
+		Domain:        "example.com" + testSuffix,
```
camilamacedo86, line 71:
```suggestion
		Domain:        fmt.Sprintf("example.com%s" + testSuffix),
```

test/e2e/utils/test_context.go
```diff
@@ -66,14 +67,15 @@ func NewTestContext(binaryName string, env ...string) (*TestContext, error) {
...
+		Group:         testGroup,
+		Version:       "v1alpha1",
+		SecondVersion: "v1",
+		Kind:          "Foo" + testSuffix,
```
camilamacedo86, line 75:
```suggestion
		Kind:           fmt.Sprintf("Foo%s" + testSuffix), 
```

test/e2e/utils/test_context.go
```diff
@@ -66,14 +67,15 @@ func NewTestContext(binaryName string, env ...string) (*TestContext, error) {
...
+		Resources:     "foo" + testSuffix + "s",
+		ImageName:     "e2e-test/controller-manager:" + testSuffix,
+		CmdContext:    cc,
```
camilamacedo86, line 78: Could we use the fmt.Sprintf to concat the values above?

testdata/project-v2/PROJECT
```diff
@@ -5,9 +5,9 @@ resources:
   kind: Captain
   version: v1
 - group: crew
-  kind: FirstMate
+  kind: Admiral
   version: v1
 - group: crew
-  kind: Admiral
+  kind: FirstMate
```
camilamacedo86, line 11: Why we are changing v2 samples if the change is only for v3+?

test/e2e/utils/test_context.go
```diff
@@ -35,15 +35,16 @@ const prometheusOperatorURL = "https://raw.githubusercontent.com/coreos/promethe
...
-	Kind       string
-	Resources  string
-	ImageName  string
-	Kubectl    *Kubectl
-	BinaryName string
+	TestSuffix    string
+	Domain        string
+	Group         string
+	Version       string
+	SecondVersion string
```
camilamacedo86, line 42: IMO we sould not add Secondversion here. What does it means for the TestContext?

generate_testdata.sh
```diff
@@ -61,9 +61,15 @@ scaffold_test_project() {
...
             $kb create api --group crew --version v1 --kind Captain --controller=true --resource=true --make=false
             $kb create webhook --group crew --version v1 --kind Captain --defaulting --programmatic-validation
-            $kb create api --group crew --version v1 --kind FirstMate --controller=true --resource=true --make=false
-            $kb create webhook --group crew --version v1 --kind FirstMate --conversion
             $kb create api --group crew --version v1 --kind Admiral --controller=true --resource=true --namespaced=false --make=false
+            $kb create api --group crew --version v1 --kind FirstMate --controller=true --resource=true --make=false
+            if [ $plugin_flag == "go/v3-alpha" ]; then
+                header_text 'Creating API for conversion webhook ...'
+                $kb create api --group crew --version v2 --kind FirstMate --controller=false --resource=true --make=false
+                $kb create webhook --group crew --version v2 --kind FirstMate --hub=v1 --spoke=v2
```
camilamacedo86, line 69: I did not find the scaffold in the project-v3 with the --hub/spoke content. Should not it change the scaffolds?

pkg/plugin/v3/webhook.go
varshaprasad96: Since it was suggested that we preserve the current implementation, have modified this such that we can either specify `hub` and `spoke`(need not add `--conversion`) or just add `--conversion`. If former format is followed, we would scaffold implementations for hub & spoke, if not then do the same as it was followed previously.

test/e2e/utils/test_context.go
varshaprasad96: Done.

test/e2e/utils/test_context.go
varshaprasad96: Sorry, this was an intermediate change while testing. Have reverted it back.

generate_testdata.sh
varshaprasad96: Have modified the script. The scaffold appears now.

generate_testdata.sh
```diff
@@ -62,9 +62,16 @@ scaffold_test_project() {
             $kb create api --group crew --version v1 --kind Captain --controller=true --resource=true --make=false
```
camilamacedo86, line 62: The commits needs to be squashed

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook.go
```diff
@@ -36,10 +36,12 @@ type Webhook struct { // nolint:maligned
...
 	GroupDomainWithDash string
 
-	// If scaffold the defaulting webhook
+	// If defaulting webhook is to be scaffolded
 	Defaulting bool
-	// If scaffold the validating webhook
+	// If validating webhook is to be scaffolded
 	Validating bool
+	// If conversion webhook is to be scaffolded
+	Conversion bool
```
Adirio, line 44: This is not used, right? Maybe some leftovers from previous commits.
```suggestion
```

pkg/plugin/v3/scaffolds/webhook.go
```diff
@@ -74,18 +80,30 @@ func (s *webhookScaffolder) newUniverse() *model.Universe {
...
+		}
+		if err := machinery.NewScaffold().Execute(
+			s.newUniverse(),
+			&api.Webhook{Defaulting: s.defaulting, Validating: s.validation, Conversion: s.conversion},
+			&templates.MainUpdater{WireWebhook: true},
+			&kdefault.InjectCAPatch{},
+		); err != nil {
+			return err
+		}
 	}
```
Adirio, line 107:
```suggestion
	if err := machinery.NewScaffold().Execute(
		s.newUniverse(),
		&api.Webhook{Defaulting: s.defaulting, Validating: s.validation},
		&templates.MainUpdater{WireWebhook: true},
		&kdefault.InjectCAPatch{},
	); err != nil {
		return err
	}
	if s.conversion {
		if s.webhookSpokeVersion != "" && s.webhookHubVersion != "" {
			if err := machinery.NewScaffold().Execute(
				&api.HubConversion{HubVersion: s.webhookHubVersion},
				&api.SpokeConversion{SpokeVersion: s.webhookSpokeVersion},
			); err != nil {
				return err
			}
		} else {
			fmt.Println(`Webhook server has been set up for you.		
 You need to implement the conversion.Hub and conversion.Convertible interfaces for your CRD types.`)
			fmt.Println("Alternatively, Hub and Spoke version can be specified in the command to scaffold Convertible and Hub interfaces")
		}
	}
```

pkg/plugin/v3/webhook.go
```diff
@@ -102,6 +110,27 @@ func (p *createWebhookPlugin) Validate() error {
 			" kind and version provided", p.commandName)
 	}
 
+	// if conversion flag is set but hub and spoke version is not provided, suggest it to user and return
+	if p.conversion && (p.spoke == "" && p.hub == "") {
+		klog.V(2).Info("Hub and Spoke version can be specified in the command to scaffold Convertible and Hub interfaces")
```
Adirio, line 115: Not worth a new dependency for this. Also, better if we move the message out of `Validate` (already did in the previous suggestion).
```suggestion
```

pkg/plugin/v3/webhook.go
```diff
@@ -102,6 +110,27 @@ func (p *createWebhookPlugin) Validate() error {
...
+
+	// set conversion to true if hub and spoke version are specified
+	if p.hub != "" && p.spoke != "" {
+		p.conversion = true
+	}
+
+	// validate the values provided for hub and spoke
+	if p.conversion {
+		if p.spoke == "" || p.hub == "" {
+			return errors.New("Both Hub and Spoke version need to be specified when using conversion webhooks")
```
Adirio, line 127: Lets use `fmt.Errorf` not to import another library.
```suggestion
			return fmt.Errorf("Both Hub and Spoke version need to be specified when using conversion webhooks")
```

test/e2e/utils/test_context.go
```diff
@@ -70,9 +70,9 @@ func NewTestContext(binaryName string, env ...string) (*TestContext, error) {
 		Domain:     "example.com" + testSuffix,
 		Group:      testGroup,
 		Version:    "v1alpha1",
-		Kind:       "Foo" + testSuffix,
-		Resources:  "foo" + testSuffix + "s",
-		ImageName:  "e2e-test/controller-manager:" + testSuffix,
+		Kind:       fmt.Sprintf("Foo%s", testSuffix),
+		Resources:  fmt.Sprintf("foo%s", testSuffix) + "s",
+		ImageName:  fmt.Sprintf("e2e-test/controller-manager:%s", testSuffix),
```
Adirio, line 75: Can you revert this change? It has nothing to do with this PR and is only done partially.

go.mod
```diff
@@ -10,5 +10,6 @@ require (
 	github.com/spf13/cobra v0.0.7
 	github.com/spf13/pflag v1.0.5
 	golang.org/x/tools v0.0.0-20200403190813-44a64ad78b9b
+	k8s.io/klog v1.0.0
```
Adirio, line 13:
```suggestion
```

pkg/plugin/v3/webhook.go
```diff
@@ -17,11 +17,13 @@ limitations under the License.
 package v3
 
 import (
+	"errors"
```
Adirio, line 20:
```suggestion
```

Adirio (changes requested):
This PR needs as some of the packages have changed location. `pkg/plugin/v3` can now be found at `pkg/plugins/golang/v3`.

Also, I added several comments inline.

camilamacedo86:
Hi @varshaprasad96, 

Is possible to release it with the master and see if we can get merged or is no longer it required?  WDYT?

varshaprasad96:
@camilamacedo86 Ill resolve the conflicts and update with the most recent changes

varshaprasad96:
Closing this PR in favor of #2084. This PR required some significant amount of changes based on v3-alpha plugin. Hence, started fresh and created a new PR for this feature on v3 plugin.

PR #1675, opened 2020-09-11
:book: Set default-theme to light explicitly if theme is set to empty string

### LINKED ISSUE #1674: Book theme not loading correctly for newly accessed users
When there is no browser localStorage holding the theme information, the theme gets set incorrectly, and makes theme updates later impossible.

You can reproduce the error by removing the localStorage from the browser, or use incognito mode.

```
(index):71 Uncaught DOMException: Failed to execute 'add' on 'DOMTokenList': The token provided must not be empty.
    at https://book.kubebuilder.io/:71:28
(anonymous) @ (index):71
book.js:362 Uncaught DOMException: Failed to execute 'remove' on 'DOMTokenList': The token provided must not be empty.
    at set_theme (https://book.kubebuilder.io/book.js:362:24)
    at themes (https://book.kubebuilder.io/book.js:371:5)
    at https://book.kubebuilder.io/book.js:433:3
```

After this error, the theme selection would not work, and the sidebar control also stops functioning (though responsive view is handled correctly at reload of the page).

/kind documentation
/kind bug

rytswd:
As per [mdbook documentation](https://github.com/rust-lang/mdBook/blob/master/book-example/src/format/config.md#configuring-renderers), the `default-theme` should be inferred to be `light`, and that should fix the problem.
As I could see `default-theme` was rendered empty in browser, this attempts to update the configuration to explicitly provide the `default-theme`.

Also, `theme` holding empty string can cause the same result, and thus we need to ensure it is corrected for early before `set_theme()` function from `book.js` being triggered.

This fixes #1674

### DIFF
--- a/docs/book/theme/index.hbs
+++ b/docs/book/theme/index.hbs
@@ -53,6 +53,10 @@
         <script type="text/javascript">
             var path_to_root = "{{ path_to_root }}";
             var default_theme = window.matchMedia("(prefers-color-scheme: dark)").matches ? "{{ preferred_dark_theme }}" : "{{ default_theme }}";
+            
+            if (default_theme === "") {
+                default_theme = "light";
+            }
         </script>
 
         <!-- Work around some values being stored in localStorage wrapped in quotes -->


rytswd:
I wasn't sure how to test this mdbook setup locally, so just trying my hypothesis on a draft PR for now. If anyone could guide me how to test this, that would be appreciated.

I could see there were some book related branches - if this is a wrong fix, please go ahead with closing.

camilamacedo86:
Hi @rytswd, 

You can click on in the preview here in the PR. See: 

Screen Shot 2020-09-14 at 12 41 50

Also, could you please update the title with the mojo: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#pr-process?

rytswd:
Hi @camilamacedo86,

Thanks for the info! After checking this further, I found this was due to the following difference in theme setup at page generation:

Netlify deploy-preview:
`var default_theme = window.matchMedia("(prefers-color-scheme: dark)").matches ? "navy" : "light";`

image

Live book.kubebuilder.io:
`var default_theme = window.matchMedia("(prefers-color-scheme: dark)").matches ? "" : "light";`

image

Because I couldn't spot how the deploy-preview pulls in `navy` theme for dark scheme, I simply added a fallback to `light` in case the theme setup fails. The fix is simpler this way, and hopefully not too intrusive even when more setup change is required.

--

It seems Netlify build setup between preview and live are slightly different, as I don't see any error with deploy-preview. Is the current live book on an older build than preview, i.e. master branch is ahead of the current book? If that's the case, this PR may not be necessary at all.

camilamacedo86:
Hi @rytswd, 

Really thank you for your contribution. I am not sure if I got your comment. However, see the changes made in your PR can be checked via: https://deploy-preview-1675--kubebuilder.netlify.app/.

The live book in https://book.kubebuilder.io/quick-start.html is not the master branch. See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#how-to-contribute-to-docs

rytswd:
Hi @camilamacedo86,

Apologies for the unclear note, and I should have gone over the contribution guideline once more as it answered most of the points I raised above 🙇

I think the change in this PR is valid, but may be redundant. I can see the deploy-preview for other PR such as https://deploy-preview-1670--kubebuilder.netlify.app/ is loading up correctly without any error. I'll have a closer look at the diff between `book-v2` and `master` branch to see if we really need a change like this.

camilamacedo86:
Hi @rytswd,

It is all fine. I think what you need to ensure is that the previews show what you would like to. Then, in the next release, the live doc will be updated and will look like the preview. It is possible that you find things solved for the docs on master already as well.

rytswd:
After looking at this closely on the master branch and deploy-preview, this looks to be already resolved. Closing this and the issue ticket associated.

PR #1676, opened 2020-09-11
doc: Fix Kind name for CronJob book and correct wording

rytswd:
### Change Description
 
1. `main.go` referenced in book

- Where: Kind name
- What: Book was using `Captain` as the Kind name whereas it was suggesting to run `kubebuilder create api` with `CronJob`

2. `conversion-concept.md`

- Where: Doc
- What: Some minor wording cleanup

Also, this file has an incomplete sentence at [the top of the file](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/multiversion-tutorial/conversion-concepts.md), where the sentence ends with "Namely:" as per below:
> Before we do that, though, we'll need to understand how controller-runtime thinks about versions. Namely:

I suppose it's about to mention something along the line of "conversion", but kept it as is as I wasn't sure of the actual intention.

### Change Motivation
Doc consistency

---

As this is a minor change, there is no associated issue.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/main.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/main.go
@@ -85,10 +85,10 @@ func main() {
 
 	if err = (&controllers.CronJobReconciler{
 		Client: mgr.GetClient(),
-		Log:    ctrl.Log.WithName("controllers").WithName("Captain"),
+		Log:    ctrl.Log.WithName("controllers").WithName("CronJob"),
 		Scheme: mgr.GetScheme(),
 	}).SetupWithManager(mgr); err != nil {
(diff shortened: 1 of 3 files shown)


gabbifish:
Hi @rytswd, thank you for your PR! I think the changes in this PR are already included in https://github.com/kubernetes-sigs/kubebuilder/pull/1632, which should be merged soon. @camilamacedo86 let me know if you think this PR should be considered/merged given that https://github.com/kubernetes-sigs/kubebuilder/pull/1632 will fix this (and other cases where `Captain` is incorrectly used).

rytswd:
Ah sorry missed that PR - I'll close this to avoid merge conflict. Thanks for pointing out!

PR #1679, opened 2020-09-13
:sparkles: Added support for empty group in kubebuilder api creation with v3+ #1404

### LINKED ISSUE #1404: Feature Request: support "empty" group for API resources, i.e. with group == domain
I would like to see kubebuilder support creating APIs with empty groups. By that I mean that the actual `apiVersion` of CRD to be `domain/version` without additional group prefix.

I think this can be very helpful for controllers/operators with a very narrow scope and a small number of `Kind`-s fitting into single "root" group.

Consider popular examples, such as [cert-manager](https://github.com/jetstack/cert-manager) or [jaeger-operator](https://github.com/jaegertracing/jaeger-operator):

```yaml
apiVersion: cert-manager.io/v1alpha2
kind: Certificate
metadata:
  name: example-com
```

```yaml
apiVersion: jaegertracing.io/v1
kind: Jaeger
metadata:
  name: simplest
```

In this examples there is no "groups" in terms of kubebuilder, and one group in terms of Kubernetes resource (for kubebuilder it is`group == domain`). And for many such cases it would be desirable to design APIs in a similar simple way.

/kind feature

prafull01:
## Description of the change: 
1. Added support for the kubebuilder api creation without providing any group. If no group is provided then instead of "group"."domain", apiVersion will have only "domain"
2. Added validation that you can't provide group and domain both as empty
3. Added multi group support for the empty group by not creating a directory for the empty group in the apis and controller directory.

## Motivation for the change:
It will improve user experience in api creation using kubebuilder.
It will solve the workarounds used by user to just have their domain name as the apiVersion.

Fixes #1404


gabbifish (commented):
Thank you for seeing https://github.com/kubernetes-sigs/kubebuilder/issues/1404 through! I had some minor nits, but this looks good to me :)

pkg/model/resource/options.go
```diff
@@ -188,8 +183,12 @@ func (opts *Options) NewResource(c *config.Config, doResource bool) *Resource {
 
 	res.Package = pkg
 	res.Domain = opts.Group
-	if domain != "" {
+	if domain != "" && res.Domain != "" {
```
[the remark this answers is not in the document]
prafull01: Done.

pkg/model/resource/options.go
```diff
@@ -188,8 +183,12 @@ func (opts *Options) NewResource(c *config.Config, doResource bool) *Resource {
...
 		res.Domain += "." + domain
+	} else if opts.Group == "" {
```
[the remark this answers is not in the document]
prafull01: Done

pkg/model/resource/resource.go
```diff
@@ -69,7 +69,13 @@ func wrapKey(key string) string {
 func (r Resource) Replacer() *strings.Replacer {
 	var replacements []string
 
-	replacements = append(replacements, wrapKey("group"), r.Group)
+	// If group is empty replace the path with core.
+	group := r.Group
+	if group == "" {
+		group = "core"
+	}
```
camilamacedo86, line 76: I do not think that we should replace the empty "" group by the core. I think we need be able to make the tool works with empty groups.

pkg/plugin/v2/scaffolds/internal/templates/main.go
```diff
@@ -152,6 +152,11 @@ func (f *MainUpdater) GetCodeFragments() file.CodeFragmentsMap {
 		return fragments
 	}
 
+	// If Group is empty add the core in import path
+	if f.Resource.Group == "" {
+		f.Resource.Group = "core"
+	}
+
```
camilamacedo86, line 159: same here.

pkg/plugin/v3/scaffolds/internal/templates/main.go
```diff
@@ -146,6 +146,11 @@ func (f *MainUpdater) GetCodeFragments() file.CodeFragmentsMap {
 		return fragments
 	}
 
+	// If Group is empty add the core in import path
+	if f.Resource.Group == "" {
+		f.Resource.Group = "core"
+	}
```
camilamacedo86, line 152: same

pkg/plugin/v3/scaffolds/internal/templates/main.go
```diff
@@ -146,6 +146,11 @@ func (f *MainUpdater) GetCodeFragments() file.CodeFragmentsMap {
...
+
 	// Generate import code fragments
 	imports := make([]string, 0)
 	imports = append(imports, fmt.Sprintf(apiImportCodeFragment, f.Resource.ImportAlias, f.Resource.Package))
```
camilamacedo86, line 156: We need to add this scenario at https://github.com/kubernetes-sigs/kubebuilder/blob/master/generate_testdata.sh as well in order to generate for the v3 testdata projects an API with a group empty too.

camilamacedo86 (changes requested):
HI @prafull01., 

Really thank you for the contribution 👍 It is great. IMO we need to apply this only to v3+, we need make the tool be able to work with empty groups instead of replacing the empty string by core and it is missing this scenario to the generate_testadata.sh script. 

However, it is great that you are looking for to address this need.

pkg/model/resource/options.go
```diff
@@ -91,19 +90,13 @@ func (opts *Options) Validate() error {
 	// We can safely look for a '-' as the first char as none of the fields accepts it
 	// NOTE: We must do this for all the required flags first or we may output the wrong
 	// error as flags may seem to be missing because Cobra assigned them to another flag.
-	if strings.HasPrefix(opts.Group, "-") {
-		return fmt.Errorf(groupRequired)
-	}
```
camilamacedo86, line 96: We need to keep these checks for v2.

pkg/model/resource/options_test.go
```diff
@@ -17,10 +17,9 @@ var _ = Describe("Resource Options", func() {
 			Expect(options.Validate()).To(Succeed())
 		})
 
-		It("should fail if the Group is not specified", func() {
+		It("should succeed if the Group is not specified", func() {
 			options := &Options{Version: "v1", Kind: "FirstMate"}
-			Expect(options.Validate()).NotTo(Succeed())
-			Expect(options.Validate().Error()).To(ContainSubstring("group cannot be empty"))
+			Expect(options.Validate()).To(Succeed())
```
camilamacedo86, line 22: if is v2 the "group cannot be empty" otherwise yes. we need to improve the check

pkg/model/resource/resource.go
prafull01: This has been done to support the multi group with the empty group specified in this [comment](https://github.com/kubernetes-sigs/kubebuilder/issues/1404#issuecomment-594504374). The empty group will only have path in core directory, the PROJECT file or any configuration files will not have any group.

camilamacedo86:
Also, could you please update the title with the mojo: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#pr-process and add (v3+ only)? Also, please ensure that your commits are squashed.

pkg/model/resource/resource.go
prafull01: @camilamacedo86 To support empty group name in multi groups without replacing the directory paths as "core". The path for empty group types.go will be "api/version/types.go" instead of "api/group/version/types.go" and same goes for the controller. Kindly suggest Is this an appropriate way to handle this?

camilamacedo86: IMO the problem here is that we are defining the resource group as core. So, it goes to PROJECT file and etc. It will no only about the paths and imports generated. 

The tool has 2 layouts:

**default** (not support multigroup)  

> project/api/v1/<kind>_types.go 
> project/controllers/<kind>_controller.go 

Which in this case shows that would keep the same. 

However, we have the multigroup layout: 

> project/api/<version>/<group>/<kind>_types.go 
> project/controllers/<group>/<kind>_controller.go 

Then, if it is an empty group we could check if would be possible have 

> project/api/v1/<group>/<kind>_types.go 
> project/controllers/<group>/<kind>_controller.go 

AND

> project/api/<version>/<kind>_types.go 
> project/controllers/<kind>_controller.go 

If the above not works then we could use as a workaround the core but in this way, IMO we need to find a way to fix the paths without set group="core" in the PROJECT file.

pkg/plugin/v2/scaffolds/internal/templates/main.go
prafull01: Done

pkg/plugin/v3/scaffolds/internal/templates/main.go
prafull01: Done

pkg/model/resource/options.go
prafull01: Added check for not allowing empty group in the api create validation for v2.

testdata/project-v3-multigroup/PROJECT
```diff
@@ -25,4 +25,6 @@ resources:
 - group: foo.policy
   kind: HealthCheckPolicy
   version: v1
+- kind: Lakers
+  version: v1
```
camilamacedo86, line 29: 👍

pkg/model/resource/options_test.go
prafull01: Can't check the project version here, as to check that func IsV2() or IsV3() is part of project config which is not available here at unit test level.

pkg/plugin/v2/api.go
```diff
@@ -131,6 +131,14 @@ func (p *createAPIPlugin) Validate() error {
 		return err
 	}
 
+	if p.resource.Group == "" && p.config.Domain == "" {
+		return fmt.Errorf("can not have group and domain both empty")
+	}
+
+	if p.config.IsV2() && p.config.Domain == "" {
+		fmt.Errorf("group cannot be empty")
```
camilamacedo86, line 139: we cannot do it for v2.

pkg/model/resource/resource_test.go
```diff
@@ -234,5 +235,27 @@ var _ = Describe("Resource", func() {
 			Expect(resource.Package).To(Equal(path.Join("k8s.io", "api", options.Group, options.Version)))
 			Expect(resource.Domain).To(Equal("authentication.k8s.io"))
 		})
+		It("should use domain if the group is empty", func() {
```
camilamacedo86, line 238: we need do the check for v2 and v3

pkg/plugin/v2/api.go
prafull01: We can not validate the group length based on the version at options level. The type Options do not have the information about the project config and we need project config to determine the project is V2 or V3

pkg/model/resource/resource_test.go
prafull01: Done

pkg/plugin/v2/api.go
prafull01: Fixed it

generate_testdata.sh
```diff
@@ -79,6 +79,10 @@ scaffold_test_project() {
             $kb create api --group sea-creatures --version v1beta1 --kind Kraken --controller=true --resource=true --make=false
             $kb create api --group sea-creatures --version v1beta2 --kind Leviathan --controller=true --resource=true --make=false
             $kb create api --group foo.policy --version v1 --kind HealthCheckPolicy --controller=true --resource=true --make=false
+            if [ $project == "project-v3-multigroup" ]; then
+              $kb create api --version v1 --kind Lakers --controller=true --resource=true --make=false
+              $kb create webhook --version v1 --kind Lakers --defaulting --programmatic-validation
+            fi
           elif [ $project == "project-v2-addon" ] || [ $project == "project-v3-addon" ]; then
             header_text 'enabling --pattern flag ...'
             export KUBEBUILDER_ENABLE_PLUGINS=1
```
camilamacedo86, line 88: We need to the same for project-v3 and project-v3-addon too.

prafull01: @camilamacedo86 I need to create a separate project for those "project-v3-emptygroup" and "project-v3-addon-emptygroup" because multiple groups are not enabled by default. Let me know if we should that?

prafull01: Done

camilamacedo86: Sorry, it was my mistake. It is true we cannot add this scenario in the other both projects because they do not support multigroup. we should not add new scenarios because will be hard to keep to maintain many of them.

pkg/model/resource/options.go
camilamacedo86: See that the above check will not be done. 
IMO we need to create a ValidateV2 which will be the original code and have these changes in the Validate. 
The plugin v2 will call the ValiddateV2 and any upper version will use the Validate. Is it make sense?

pkg/model/resource/options_test.go
camilamacedo86: we need to have the ValidateV2 test and the Validate.
When we no longer support V2 this test and func will be removed. However, so far we need to ensure that has no breaking changes.

pkg/plugin/v3/scaffolds/internal/templates/main.go
camilamacedo86: You can close this one it is solved.

pkg/model/resource/resource.go
camilamacedo86: You can close this one it is solved. I cloned your PR locally and re-check it. I like your solution here. It was a good way to solve it. 👍

camilamacedo86 (commented):
It is a great contribution. 🥇 
Just a few nits:
- we cannot add more testdata projects will be hard to keep maintained
- IMO it is better we have teh ValidatonV2 and Validation, then we can ensure that we did not break anything and will be easier to remove it with the v2 code in the future.

Otherwise, it shows terrific. 👍

pkg/model/resource/options.go
prafull01: Done.

pkg/model/resource/options_test.go
prafull01: Done

generate_testdata.sh
prafull01: Reverted.

camilamacedo86: You can close this one.

prafull01:
replacements = append(replacements, wrapKey("group"), r.Group)

pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller_suitetest.go
```diff
@@ -38,7 +38,7 @@ type SuiteTest struct {
 // SetTemplateDefaults implements file.Template
 func (f *SuiteTest) SetTemplateDefaults() error {
 	if f.Path == "" {
-		if f.MultiGroup {
+		if f.MultiGroup && f.Resource.Group != "" {
 			f.Path = filepath.Join("controllers", "%[group]", "suite_test.go")
 		} else {
 			f.Path = filepath.Join("controllers", "suite_test.go")
```
camilamacedo86, line 44: We need to merge the fix #1682 and the will be required add more a conditional here. https://github.com/kubernetes-sigs/kubebuilder/pull/1682/files#diff-71066f95caa43267d8f2443a3c222fbeR57-R63  

If is multigroup and the resource is not empty then the path of the CRDs is not ../../ otherwise it is ../

pkg/model/resource/options.go
camilamacedo86: You can close this one it is solved 👍

pkg/model/resource/options_test.go
camilamacedo86: You can close this one it is solved 👍

pkg/plugin/v2/api.go
camilamacedo86: You can close this one it is solved 👍

pkg/model/resource/resource_test.go
camilamacedo86: You can close this one it is solved 👍

generate_testdata.sh
```diff
@@ -103,4 +107,4 @@ scaffold_test_project project-v2-multigroup 2
 scaffold_test_project project-v2-addon 2
 scaffold_test_project project-v3 3-alpha go/v3-alpha
 scaffold_test_project project-v3-multigroup 3-alpha go/v3-alpha
-scaffold_test_project project-v3-addon 3-alpha go/v3-alpha
+scaffold_test_project project-v3-addon 3-alpha go/v3-alpha
```
camilamacedo86, line 110: We need to give an space at the end of the file here. See the sign ^

pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller_suitetest.go
prafull01: Got it, I will wait for this PR to be merged.

generate_testdata.sh
prafull01: will do

pkg/model/resource/options.go
```diff
@@ -165,9 +214,15 @@ func (opts *Options) NewResource(c *config.Config, doResource bool) *Resource {
...
+	apiPath := "api"
 	if c.MultiGroup {
-		pkg = replacer.Replace(path.Join(c.Repo, "apis", "%[group]", "%[version]"))
+		apiPath = "apis"
+	}
+
+	pkg := replacer.Replace(path.Join(c.Repo, apiPath, "%[version]"))
+	if c.MultiGroup && opts.Group != "" {
+		pkg = replacer.Replace(path.Join(c.Repo, apiPath, "%[group]", "%[version]"))
 	}
```
camilamacedo86, line 226: Nit we can simplify the change here by just adding to the original code the additional condition. See; 

```
	pkg := replacer.Replace(path.Join(c.Repo, "api", "%[version]"))
	if c.MultiGroup && opts.Group != ""  {
		pkg = replacer.Replace(path.Join(c.Repo, "apis", "%[group]", "%[version]"))
	}
	domain := c.Domain
```

Screen Shot 2020-09-16 at 10 50 28

pkg/plugin/v3/scaffolds/internal/templates/config/api/group.go
```diff
@@ -35,10 +40,14 @@ type Group struct {
...
 		if f.MultiGroup {
-			f.Path = filepath.Join("apis", "%[group]", "%[version]", "groupversion_info.go")
+			apiPath = multiGroupAPI
+		}
+		if f.MultiGroup && f.Resource.Group != "" {
+			f.Path = filepath.Join(apiPath, "%[group]", "%[version]", "groupversion_info.go")
 		} else {
-			f.Path = filepath.Join("api", "%[version]", "groupversion_info.go")
+			f.Path = filepath.Join(apiPath, "%[version]", "groupversion_info.go")
 		}
```
camilamacedo86, line 51: Same: https://github.com/kubernetes-sigs/kubebuilder/pull/1679/files#r489311174

pkg/plugin/v3/scaffolds/internal/templates/config/api/types.go
```diff
@@ -36,10 +36,14 @@ type Types struct {
...
 		if f.MultiGroup {
-			f.Path = filepath.Join("apis", "%[group]", "%[version]", "%[kind]_types.go")
+			apiPath = multiGroupAPI
+		}
+		if f.MultiGroup && f.Resource.Group != "" {
+			f.Path = filepath.Join(apiPath, "%[group]", "%[version]", "%[kind]_types.go")
 		} else {
-			f.Path = filepath.Join("api", "%[version]", "%[kind]_types.go")
+			f.Path = filepath.Join(apiPath, "%[version]", "%[kind]_types.go")
 		}
```
camilamacedo86, line 47: Same : https://github.com/kubernetes-sigs/kubebuilder/pull/1679/files#r489311174

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook.go
```diff
@@ -45,10 +45,14 @@ type Webhook struct { // nolint:maligned
...
 		if f.MultiGroup {
-			f.Path = filepath.Join("apis", "%[group]", "%[version]", "%[kind]_webhook.go")
+			apiPath = multiGroupAPI
+		}
+		if f.MultiGroup && f.Resource.Group != "" {
+			f.Path = filepath.Join(apiPath, "%[group]", "%[version]", "%[kind]_webhook.go")
 		} else {
-			f.Path = filepath.Join("api", "%[version]", "%[kind]_webhook.go")
+			f.Path = filepath.Join(apiPath, "%[version]", "%[kind]_webhook.go")
 		}
```
camilamacedo86, line 56: Same: https://github.com/kubernetes-sigs/kubebuilder/pull/1679/files#r489311174

camilamacedo86 (commented):
All good just a few nits see my latest comments.

pkg/model/resource/options.go
prafull01: There are two path right now, for single group "api" and for multigroup project "apis". If we go by above logic, the multigroup project with empty group will take the path as "api" instead of "apis". That is why I choose to have apiPath as "api" and "apis" based on the single or multi group first.

generate_testdata.sh
prafull01: done

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook.go
prafull01: Done

pkg/model/resource/options.go
camilamacedo86: It is perfect tks you can close this one.

camilamacedo86 (commented):
Hi @prafull01, 

It is very good. We just need to get merged the fix https://github.com/kubernetes-sigs/kubebuilder/pull/1682 to address the change here is well. Otherwise, IMO it is ok to be :   
/approve

Feel free to close all comments they are addressed 👍 
really thank you for your contribution.

camilamacedo86:
/approve cancel
Let's wait for it has the missing nit :-)

pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller_suitetest.go
prafull01: Completed the last nit :)

pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller_suitetest.go
```diff
@@ -57,7 +57,7 @@ func (f *SuiteTest) SetTemplateDefaults() error {
 	// If is multigroup the path needs to be ../../ since it has
 	// the group dir.
 	f.CRDDirectoryRelativePath = `".."`
-	if f.MultiGroup {
+	if f.MultiGroup && f.Resource.Group != "" {
 		f.CRDDirectoryRelativePath = `"..", ".."`
```
camilamacedo86, line 61: 👍

camilamacedo86 (approved):
/approve

Great work 🥇

pkg/model/resource/options.go
```diff
@@ -188,8 +241,14 @@ func (opts *Options) NewResource(c *config.Config, doResource bool) *Resource {
 
 	res.Package = pkg
 	res.Domain = opts.Group
-	if domain != "" {
+	if domain != "" && opts.Group != "" {
 		res.Domain += "." + domain
+	} else if opts.Group == "" && c.IsV3() {
```
camilamacedo86, line 246: nit 
```suggestion
	} else if opts.Group == "" && !c.IsV2() {
```

camilamacedo86: supported for any upper version that we might add as well.

prafull01: Done

testdata/project-v3-multigroup/controllers/suite_test.go
```diff
@@ -0,0 +1,81 @@
...
+		"Controller Suite",
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = BeforeSuite(func(done Done) {
+	logf.SetLogger(zap.LoggerTo(GinkgoWriter, true))
+
+	By("bootstrapping test environment")
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths: []string{filepath.Join("..", "config", "crd", "bases")},
```
camilamacedo86, line 57: 👍

camilamacedo86:
Hi @prafull01, 

The coveralls check is not passing because it is missing cover the ValidateV2 with all same checks that is done for Validate + the group one. See: https://coveralls.io/builds/33545877/source?filename=pkg/model/resource/options.go#L88

prafull01:
> Hi @prafull01,
> 
> The coveralls check is not passing because it is missing cover the ValidateV2 with all same checks that is done for Validate + the group one. See: https://coveralls.io/builds/33545877/source?filename=pkg/model/resource/options.go#L88

There are two ways I can handle it:
1. Use the same test data to validate against Validate() and ValidateV2() and in the same test cases.
2. Write the similar test case to test the ValidateV2(), however it will duplicate the lots of test cases.

How should we handle this?

camilamacedo86:
> Write the similar test case to test the ValidateV2(), however it will duplicate the lots of test cases.

I think would be fine just duplicate as well. We can remove the test when the V2 will be removed. 
If you face lint issues because of the duplication use //nolint:dupl to ignore the check and please just add a comment explaining that we are duplicating the cases for V2 but that would also be removed when v2 will no longer be supported.

DirectXMan12:
minor nit: the commit message mentions the "core" directory, but that bit seems gone.  Can you fix up the commit message.  AIUI, with multigroup we'll end up with

```
pkgs/apis/v1/xyz_types.go # test.kubebuilder.io/v1
pkgs/apis/foo/v1/xyz_types.go # foo.test.kubebuilder.io/v1
```

Is that correct?

DirectXMan12:
(otherwise, I delegate to @camilamacedo86 for the lgtm ;-) )

prafull01:
> minor nit: the commit message mentions the "core" directory, but that bit seems gone. Can you fix up the commit message. AIUI, with multigroup we'll end up with
> 
> ```
> pkgs/apis/v1/xyz_types.go # test.kubebuilder.io/v1
> pkgs/apis/foo/v1/xyz_types.go # foo.test.kubebuilder.io/v1
> ```
> 
> Is that correct?

Fixed the commit message. Yes, the path you mentioned is correct for empty group and non empty group.

pkg/plugin/v3/api.go
```diff
@@ -137,6 +137,10 @@ func (p *createAPIPlugin) Validate() error {
 		return err
 	}
 
+	if p.resource.Group == "" && p.config.Domain == "" {
```
gabbifish, line 140: I think this check is redundant, given that `p.resource.Validate()` never allows p.Config.Domain to be an empty string in the first place?

gabbifish (approved):
I had one nit about what looked like an unnecessary check, but otherwise this looks good to me!

pkg/plugin/v3/api.go
prafull01: @gabbifish IMO this is a necessary check, while initializing a project (kb init) you can provide a domain which is empty. 

Validate() function doesn't check the value of domain, it only checks the group, version, kind and related values which are provided in create api command. (kb create api). Domain is provided at the init level. Hence this check is necessary to know a user do not provide domain and group both as empty.

prafull01:
@camilamacedo86 Rebase done. `make generate` done, to have the finalizers markers in the multi-group with empty group api(lakers).

camilamacedo86:
It shows great for me. I could not find any issue with that. 
Also, received the @DirectXMan12 oks. 

So, it shows fine for we move forward 
/lgtm

PR #1683, opened 2020-09-16
:seeding: rename generate_mutiversion.sh to generate_multiversion.sh

camilamacedo86:
Just a nit to rename the script which is a helper for us. Also, added the bin projects to the .gitigonore to ensure that the k8s binaries used to setup env_test will NOT be committed


gabbifish:
/lgtm

Thank you!! :D

PR #1688, opened 2020-09-21
:sparkles: Added the changes for include finalizer as a default controller markers

### LINKED ISSUE #1654: Include <resource>/finalizers permission in default scaffold
When deploying operators to clusters that have the [`OwnerReferencesPermissionEnforcement` admission controller](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement) enabled, I want my operator to work without having to figure out why I get an error like this:

`"deployments.apps \"memcached-sample\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"`

I think this could be solved by including the following in the default controller scaffold:

```go
// +kubebuilder:rbac:groups=<group>.<domain>,resources=<resource>/finalizers,verbs=get;update;patch
```

/kind feature

prafull01:
**Description of the change**
Added a <resource>/finalizer kubebuilder marker as a default marker for the controller. This is to allow the update;patch;get operation on finalizer of the resource if the [ownerReferencesPermissionEnforcement](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement) enabled on the cluster.

**Motivation of the change**
The kubebuilder operator doesn't work seamlessly on the vanilla kubernetes like openshift unless the finalizers permission marker is added on the Reconcile function. 

Fixes #1654

### DIFF
--- a/pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller.go
+++ b/pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller.go
@@ -75,6 +75,7 @@ type {{ .Resource.Kind }}Reconciler struct {
 
 // +kubebuilder:rbac:groups={{ .Resource.Domain }},resources={{ .Resource.Plural }},verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups={{ .Resource.Domain }},resources={{ .Resource.Plural }}/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups={{ .Resource.Domain }},resources={{ .Resource.Plural }}/finalizers,verbs=update
 
 func (r *{{ .Resource.Kind }}Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	_ = context.Background()
--- a/testdata/project-v3-multigroup/config/rbac/role.yaml
+++ b/testdata/project-v3-multigroup/config/rbac/role.yaml
@@ -21,6 +21,12 @@ rules:
 - apiGroups:
   - crew.testproject.org
   resources:
+  - captains/finalizers
+  verbs:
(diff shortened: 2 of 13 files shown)


camilamacedo86 (approved):
it shows great 🥇 

/approve

gabbifish (approved):
Thank you for this, LGTM!

joelanford:
@prafull01 I just realized that the issue I filed may have extraneous/unnecessary permissions. Reading the docs for [OwnerReferencesPermissionEnforcement](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement) again, I see this:

> This admission controller also protects the access to `metadata.ownerReferences[x].blockOwnerDeletion` of an object, so that only users with "update" permission to the finalizers subresource of the referenced owner can change it.

I think we can scope this down to just `update` permission on `<resource>/finalizers`.

prafull01:
> @prafull01 I just realized that the issue I filed may have extraneous/unnecessary permissions. Reading the docs for [OwnerReferencesPermissionEnforcement](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement) again, I see this:
> 
> > This admission controller also protects the access to `metadata.ownerReferences[x].blockOwnerDeletion` of an object, so that only users with "update" permission to the finalizers subresource of the referenced owner can change it.
> 
> I think we can scope this down to just `update` permission on `<resource>/finalizers`.

Done. Updated with exact permission :)

camilamacedo86:
All suggestions addressed. 
Also, received more than 1 review and OKs. So

/lgtm

PR #1696, opened 2020-09-25
:sparkles: store all config info (go/v3-alpha) 

camilamacedo86:
**Description**

Store in the PROJECT file:
- webhooks data
- if has or not an api scaffolded
- if has or not a controller scaffolded
- API domain  and endpoint when it is not the default form
- Plural form when it is not the default form

Refractories/Cleanups suggested: 
- Rename Domain in the internal code for QualifiedGroup
- Rename Package for Endpoint
- Rename GVK to ResourceData since it does not return only the GKV
- Replace `doResource` with `doAPI` only for v3  internal code. 

PS.: Note the plugin v2 internal code was changed/updated where only it would be mandatory in order to move forward with the changes. The idea is we ensure and not risk introduces breaking changes or bugs into it since we are closer to deprecated it.  

Regards the tests:
- Add new tests to cover the config
- Add new api scaffolded in the testdata for multigroup projects to cover better the k8s core types option.  

**Example**

Following an example with the edge cases and that cover nearly all scenarios. 

```yaml
domain: testproject.org
layout: go.kubebuilder.io/v3-alpha
multigroup: true
projectName: project-v3-multigroup
repo: sigs.k8s.io/kubebuilder/testdata/project-v3-multigroup
resources:
- controller: true
  domain: k8s.io
  endpoint: k8s.io/api/authentication/v1
  group: authentication
  kind: TokenReview
  version: v1
- api:
    crdVersion: v1
    namespaced: true
  controller: true
  domain: testproject.org
  kind: Lakers
  version: v1
  webhooks:
    defaulting: true
    validation: true
    webhookVersion: v1
version: 3-alpha
```

**Motivation**
- Allow we have all project config into the config which can, for example, grant we re-create the project based on it. 
- keep the consistency across the project
- Allow we easily manage diff versions of apiVersions for webhooks and CRDs. See that the apiVersion fits very well as an attribute of its models. So, we could store the apiVersion which would allow how the tool should dealing with each case scenario. See: https://github.com/kubernetes-sigs/kubebuilder/pull/1644#discussion_r491227667

Close: https://github.com/kubernetes-sigs/kubebuilder/issues/1826

/hold

If we do not able o ve forward with https://github.com/kubernetes-sigs/kubebuilder/pull/1869 instead of this one. Then, here would be required to do a few changes : 
- Use the nomenclature Path instead of endpoint 
- Do not store the domain when it is a core value.

### DIFF
--- a/pkg/model/config/config.go
+++ b/pkg/model/config/config.go
@@ -18,8 +18,10 @@ package config
 
 import (
 	"fmt"
+	"path"
 	"strings"
 
+	"github.com/gobuffalo/flect"
 	"sigs.k8s.io/yaml"
 )
 
@@ -34,7 +36,7 @@ type Config struct {
 	// Version is the project version, defaults to "1" (backwards compatibility)
 	Version string `json:"version,omitempty"`
 
-	// Domain is the domain associated with the project and used for API groups
+	// QualifiedGroup is the domain associated with the project and used for API groups
 	Domain string `json:"domain,omitempty"`
 
 	// Repo is the go package name of the project root
@@ -45,7 +47,7 @@ type Config struct {
 
 	// Resources tracks scaffolded resources in the project
 	// This info is tracked only in project with version 2
-	Resources []GVK `json:"resources,omitempty"`
+	Resources []ResourceData `json:"resources,omitempty"`
 
 	// Multigroup tracks if the project has more than one group
 	MultiGroup bool `json:"multigroup,omitempty"`
@@ -78,32 +80,30 @@ func (c Config) IsV3() bool {
 	return c.Version == Version3Alpha
 }
 
-// HasResource returns true if API resource is already tracked
-func (c Config) HasResource(target GVK) bool {
+// GetResource returns the GKV if the resource is found
+func (c Config) GetResource(target ResourceData) *ResourceData {
 	// Return true if the target resource is found in the tracked resources
 	for _, r := range c.Resources {
-		if r.isEqualTo(target) {
-			return true
+		if r.isGVKEqualTo(target) {
+			return &r
 		}
 	}
-
-	// Return false otherwise
-	return false
+	return nil
 }
 
 // UpdateResources either adds gvk to the tracked set or, if the resource already exists,
 // updates the the equivalent resource in the set.
-func (c *Config) UpdateResources(gvk GVK) {
+func (c *Config) UpdateResources(resource ResourceData) {
 	// If the resource already exists, update it.
 	for i, r := range c.Resources {
-		if r.isEqualTo(gvk) {
-			c.Resources[i].merge(gvk)
+		if r.isGVKEqualTo(resource) {
+			c.Resources[i].merge(resource)
 			return
 		}
 	}
 
 	// The resource does not exist, append the resource to the tracked ones.
-	c.Resources = append(c.Resources, gvk)
+	c.Resources = append(c.Resources, resource)
 }
 
 // HasGroup returns true if group is already tracked
@@ -136,9 +136,13 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
 		var currVersion string
 		switch verType {
 		case "crd":
-			currVersion = res.CRDVersion
+			if res.API != nil {
+				currVersion = res.API.CRDVersion
+			}
 		case "webhook":
-			currVersion = res.WebhookVersion
+			if res.Webhooks != nil {
+				currVersion = res.Webhooks.WebhookVersion
+			}
 		}
 		if currVersion != "" && version != currVersion {
 			return false
@@ -147,20 +151,58 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
 	return true
 }
 
-// GVK contains information about scaffolded resources
-type GVK struct {
+// ResourceData contains information about scaffolded resources
+type ResourceData struct {
 	Group   string `json:"group,omitempty"`
 	Version string `json:"version,omitempty"`
 	Kind    string `json:"kind,omitempty"`
 
-	// CRDVersion holds the CustomResourceDefinition API version used for the GVK.
+	Domain   string `json:"domain,omitempty"`
+	Plural   string `json:"plural,omitempty"`
+	Endpoint string `json:"endpoint,omitempty"`
+
+	// API holds true when the api is scaffolded
+	API *API `json:"api,omitempty"`
+
+	// Controller holds true when the controller is scaffolded
+	Controller bool `json:"controller,omitempty"`
+
+	// Webhooks holds the Webhooks data
+	Webhooks *Webhooks `json:"webhooks,omitempty"`
+}
+
+// API contains information about scaffolded APIs
+type API struct {
+	// Namespaced is true if the resource is namespaced
+	Namespaced bool `json:"namespaced,omitempty"`
+
+	// CRDVersion holds the CustomResourceDefinition API version used for the ResourceData.
 	CRDVersion string `json:"crdVersion,omitempty"`
-	// WebhookVersion holds the {Validating,Mutating}WebhookConfiguration API version used for the GVK.
+}
+
+// IsEmpty returns if the API's fields all contain zero-values.
+func (a API) IsEmpty() bool {
+	return a.CRDVersion == "" && !a.Namespaced
+}
+
+// Webhooks contains information about scaffolded webhooks
+type Webhooks struct {
+	// Types which are scaffolded
+	Defaulting bool `json:"defaulting,omitempty"`
+	Validation bool `json:"validation,omitempty"`
+	Conversion bool `json:"conversion,omitempty"`
+
+	// WebhookVersion holds the {Validating,Mutating}WebhookConfiguration API version used for the Options.
 	WebhookVersion string `json:"webhookVersion,omitempty"`
 }
 
-// isEqualTo compares it with another resource
-func (r GVK) isEqualTo(other GVK) bool {
+// IsEmpty returns if the Webhooks' fields all contain zero-values.
+func (w Webhooks) IsEmpty() bool {
+	return w.WebhookVersion == "" && !w.Defaulting && !w.Validation && !w.Conversion
+}
+
+// isGVKEqualTo compares it with another resource
+func (r ResourceData) isGVKEqualTo(other ResourceData) bool {
 	return r.Group == other.Group &&
 		r.Version == other.Version &&
 		r.Kind == other.Kind
@@ -168,13 +210,44 @@ func (r GVK) isEqualTo(other GVK) bool {
 
 // merge combines fields of two GVKs that have matching group, version, and kind,
 // favoring the receiver's values.
-func (r *GVK) merge(other GVK) {
-	if r.CRDVersion == "" && other.CRDVersion != "" {
-		r.CRDVersion = other.CRDVersion
+func (r *ResourceData) merge(other ResourceData) {
+	r.Controller = r.Controller || other.Controller
+
+	if other.Webhooks != nil {
+		if r.Webhooks == nil {
+			r.Webhooks = other.Webhooks
+		} else {
+			r.Webhooks.merge(other.Webhooks)
+		}
 	}
-	if r.WebhookVersion == "" && other.WebhookVersion != "" {
-		r.WebhookVersion = other.WebhookVersion
+
+	if other.API != nil {
+		if r.API == nil {
+			r.API = other.API
+		} else {
+			r.API.merge(other.API)
+		}
 	}
+}
+
+// merge compares it with another webhook by setting each webhook type individually so existing values are
+// not overwritten.
+func (w *Webhooks) merge(other *Webhooks) {
+	if w.WebhookVersion == "" && other.WebhookVersion != "" {
+		w.WebhookVersion = other.WebhookVersion
+	}
+	w.Defaulting = w.Defaulting || other.Defaulting
+	w.Validation = w.Validation || other.Validation
+	w.Conversion = w.Conversion || other.Conversion
+}
+
+// merge compares it with another webhook by setting each webhook type individually so existing values are
+// not overwritten.
+func (a *API) merge(other *API) {
+	if a.CRDVersion == "" && other.CRDVersion != "" {
+		a.CRDVersion = other.CRDVersion
+	}
+	a.Namespaced = a.Namespaced || other.Namespaced
 }
 
 // Marshal returns the bytes of c.
@@ -182,6 +255,45 @@ func (c Config) Marshal() ([]byte, error) {
 	// Ignore extra fields at first.
 	cfg := c
 	cfg.Plugins = nil
+
+	// Ignore some fields if v2.
+	if cfg.IsV2() {
+		for i := range cfg.Resources {
+			cfg.Resources[i].API = nil
+			cfg.Resources[i].Controller = false
+			cfg.Resources[i].Webhooks = nil
+			cfg.Resources[i].Plural = ""
+			cfg.Resources[i].Domain = ""
+			cfg.Resources[i].Endpoint = ""
+		}
+	}
+
+	for i, r := range cfg.Resources {
+		//If the plural is regular, omit it.
+		if r.Plural != "" && r.Plural == flect.Pluralize(strings.ToLower(r.Kind)) {
+			c.Resources[i].Plural = ""
+		}
+
+		// If the Endpoint is the default location, omit it.
+		if r.Endpoint != "" && r.Endpoint == c.DefaultAPIEndpointFor(r.Group, r.Version) {
+			c.Resources[i].Endpoint = ""
+		}
+
+		// If the domain is the default one, omit it.
+		if r.Domain != "" && r.Domain == fmt.Sprintf("%s.%s", r.Group, c.Domain) {
+			c.Resources[i].Domain = ""
+		}
+
+		// If API is empty, omit it (prevents `api: {}`).
+		if r.API != nil && r.API.IsEmpty() {
+			cfg.Resources[i].API = nil
+		}
+		// If Webhooks is empty, omit it (prevents `webhooks: {}`).
+		if r.Webhooks != nil && r.Webhooks.IsEmpty() {
+			cfg.Resources[i].Webhooks = nil
+		}
+	}
+
 	content, err := yaml.Marshal(cfg)
 	if err != nil {
 		return nil, fmt.Errorf("error marshalling project configuration: %v", err)
@@ -271,4 +383,15 @@ func (c Config) DecodePluginConfig(key string, configObj interface{}) error {
 		}
 	}
 	return nil
+}
+
+// DefaultAPIEndpointFor returns the default path for the group version informed in the project
+func (c Config) DefaultAPIEndpointFor(group, version string) string {
+	if c.MultiGroup {
+		if group != "" {
+			return path.Join(c.Repo, "apis", group, version)
+		}
+		return path.Join(c.Repo, "apis", version)
+	}
+	return path.Join(c.Repo, "api", version)
 }
--- a/pkg/model/config/config_test.go
+++ b/pkg/model/config/config_test.go
@@ -30,6 +30,88 @@ var _ = Describe("PluginConfig", func() {
 		Data1 string `json:"data-1"`
 		Data2 string `json:"data-2"`
 	}
+	const defaultWebhookVersion = "v1"
+
+	resource := ResourceData{Group: "Foo", Kind: "Baz", Version: "v1"}
+	resource.Webhooks = &Webhooks{false, false, true, defaultWebhookVersion}
+
+	It("should merge webhook when webhook is NOT equals", func() {
+		otherResource := ResourceData{Group: "Foo", Kind: "Baz", Version: "v1"}
+		otherResource.Webhooks = &Webhooks{false, true, false, defaultWebhookVersion}
+		resource.merge(otherResource)
+		Expect(resource.Webhooks.Defaulting).To(BeFalse())
+		Expect(resource.Webhooks.Validation).To(BeTrue())
+		Expect(resource.Webhooks.Conversion).To(BeTrue())
+	})
+
+	It("should merge not update the conversion and validation to false when it is true already", func() {
+		otherResource := ResourceData{Group: "Foo", Kind: "Baz", Version: "v1"}
+		otherResource.Webhooks = &Webhooks{false, false, false, defaultWebhookVersion}
+		resource.merge(otherResource)
+		Expect(resource.Webhooks.Defaulting).To(BeFalse())
+		Expect(resource.Webhooks.Validation).To(BeTrue())
+		Expect(resource.Webhooks.Conversion).To(BeTrue())
+	})
+
+	It("should merge API when API is NOT equals", func() {
+		otherResource := ResourceData{Group: "Foo", Kind: "Baz", Version: "v1"}
+		otherResource.API = &API{Namespaced: true, CRDVersion: "v1"}
+		resource.merge(otherResource)
+		Expect(resource.API.Namespaced).To(BeTrue())
+		Expect(resource.API.CRDVersion).To(BeIdenticalTo("v1"))
+	})
+
+	It("should merge not update the API when its Namespaced is true and has CRDVersion already", func() {
+		otherResource := ResourceData{Group: "Foo", Kind: "Baz", Version: "v1"}
+		otherResource.API = &API{Namespaced: false, CRDVersion: "betav1"}
+		resource.merge(otherResource)
+		Expect(resource.API.Namespaced).To(BeTrue())
+		Expect(resource.API.CRDVersion).To(BeIdenticalTo("v1"))
+	})
+
+	It("should return true when has the ResourceData is equals", func() {
+		Expect(resource.isGVKEqualTo(ResourceData{Group: "Foo", Kind: "Baz", Version: "v1"})).To(BeTrue())
+	})
+
+	It("should return false when ResourceData is NOT equals", func() {
+		Expect(resource.isGVKEqualTo(ResourceData{Group: "Foo", Kind: "Baz", Version: "v2"})).To(BeFalse())
+	})
+
+	It("should return true when the API is empty", func() {
+		resource.API = &API{
+			Namespaced: false,
+			CRDVersion: "",
+		}
+		Expect(resource.API.IsEmpty()).To(BeTrue())
+	})
+
+	It("should return true when the API is NOT empty", func() {
+		resource.API = &API{
+			Namespaced: true,
+			CRDVersion: "",
+		}
+		Expect(resource.API.IsEmpty()).To(BeFalse())
+	})
+
+	It("should return true when the WebHooks is empty", func() {
+		resource.Webhooks = &Webhooks{WebhookVersion: ""}
+		Expect(resource.Webhooks.IsEmpty()).To(BeTrue())
+	})
+
+	It("should return true when the WebHooks is NOT empty", func() {
+		resource.Webhooks = &Webhooks{WebhookVersion: "v1"}
+		Expect(resource.Webhooks.IsEmpty()).To(BeFalse())
+	})
+
+	It("IsV2 should return true when the config is V2", func() {
+		cfg := Config{Version: Version2}
+		Expect(cfg.IsV2()).To(BeTrue())
+	})
+
+	It("IsV3 should return true when the config is V3", func() {
+		cfg := Config{Version: Version3Alpha}
+		Expect(cfg.IsV3()).To(BeTrue())
+	})
 
 	It("should encode correctly", func() {
 		var (
@@ -165,56 +247,61 @@ var _ = Describe("PluginConfig", func() {
 	})
 })
 
-var _ = Describe("Resource Version Compatibility", func() {
+var _ = Describe("ResourceData Version Compatibility", func() {
 
 	var (
-		c          *Config
-		gvk1, gvk2 GVK
+		c                    *Config
+		resource1, resource2 ResourceData
 
 		defaultVersion = "v1"
 	)
 
 	BeforeEach(func() {
 		c = &Config{}
-		gvk1 = GVK{Group: "example", Version: "v1", Kind: "TestKind"}
-		gvk2 = GVK{Group: "example", Version: "v1", Kind: "TestKind2"}
+		resource1 = ResourceData{Group: "example", Version: "v1", Kind: "TestKind"}
+		resource2 = ResourceData{Group: "example", Version: "v1", Kind: "TestKind2"}
 	})
 
 	Context("resourceAPIVersionCompatible", func() {
 		It("returns true for a list of empty resources", func() {
 			Expect(c.resourceAPIVersionCompatible("crd", defaultVersion)).To(BeTrue())
+			Expect(c.resourceAPIVersionCompatible("webhook", defaultVersion)).To(BeTrue())
 		})
 		It("returns true for one resource with an empty version", func() {
-			c.Resources = []GVK{gvk1}
+			c.Resources = []ResourceData{resource1}
 			Expect(c.resourceAPIVersionCompatible("crd", defaultVersion)).To(BeTrue())
+			Expect(c.resourceAPIVersionCompatible("webhook", defaultVersion)).To(BeTrue())
 		})
 		It("returns true for one resource with matching version", func() {
-			gvk1.CRDVersion = defaultVersion
-			c.Resources = []GVK{gvk1}
+			resource1.API = &API{CRDVersion: defaultVersion}
+			resource1.Webhooks = &Webhooks{WebhookVersion: defaultVersion}
+			c.Resources = []ResourceData{resource1}
 			Expect(c.resourceAPIVersionCompatible("crd", defaultVersion)).To(BeTrue())
+			Expect(c.resourceAPIVersionCompatible("webhook", defaultVersion)).To(BeTrue())
 		})
 		It("returns true for two resources with matching versions", func() {
-			gvk1.CRDVersion = defaultVersion
-			gvk2.CRDVersion = defaultVersion
-			c.Resources = []GVK{gvk1, gvk2}
+			resource1.API = &API{CRDVersion: defaultVersion}
+			resource1.Webhooks = &Webhooks{WebhookVersion: defaultVersion}
+			resource2.API = &API{CRDVersion: defaultVersion}
+			resource2.Webhooks = &Webhooks{WebhookVersion: defaultVersion}
+			c.Resources = []ResourceData{resource1, resource2}
 			Expect(c.resourceAPIVersionCompatible("crd", defaultVersion)).To(BeTrue())
+			Expect(c.resourceAPIVersionCompatible("webhook", defaultVersion)).To(BeTrue())
 		})
 		It("returns false for one resource with a non-matching version", func() {
-			gvk1.CRDVersion = v1beta1
-			c.Resources = []GVK{gvk1}
+			resource1.API = &API{CRDVersion: v1beta1}
+			resource1.Webhooks = &Webhooks{WebhookVersion: v1beta1}
+			c.Resources = []ResourceData{resource1}
 			Expect(c.resourceAPIVersionCompatible("crd", defaultVersion)).To(BeFalse())
+			Expect(c.resourceAPIVersionCompatible("webhook", defaultVersion)).To(BeFalse())
 		})
 		It("returns false for two resources containing a non-matching version", func() {
-			gvk1.CRDVersion = v1beta1
-			gvk2.CRDVersion = defaultVersion
-			c.Resources = []GVK{gvk1, gvk2}
+			resource1.API = &API{CRDVersion: v1beta1}
+			resource1.Webhooks = &Webhooks{WebhookVersion: v1beta1}
+			resource2.API = &API{CRDVersion: defaultVersion}
+			resource2.Webhooks = &Webhooks{WebhookVersion: defaultVersion}
+			c.Resources = []ResourceData{resource1, resource2}
 			Expect(c.resourceAPIVersionCompatible("crd", defaultVersion)).To(BeFalse())
-		})
-
-		It("returns false for two resources containing a non-matching version (webhooks)", func() {
-			gvk1.WebhookVersion = v1beta1
-			gvk2.WebhookVersion = defaultVersion
-			c.Resources = []GVK{gvk1, gvk2}
 			Expect(c.resourceAPIVersionCompatible("webhook", defaultVersion)).To(BeFalse())
 		})
 	})
@@ -223,35 +310,49 @@ var _ = Describe("Resource Version Compatibility", func() {
 var _ = Describe("Config", func() {
 	var (
 		c          *Config
-		gvk1, gvk2 GVK
+		gvk1, gvk2 ResourceData
 	)
 
 	BeforeEach(func() {
 		c = &Config{}
-		gvk1 = GVK{Group: "example", Version: "v1", Kind: "TestKind"}
-		gvk2 = GVK{Group: "example", Version: "v1", Kind: "TestKind2"}
+		gvk1 = ResourceData{Group: "example", Version: "v1", Kind: "TestKind"}
+		gvk2 = ResourceData{Group: "example", Version: "v1", Kind: "TestKind2"}
 	})
 
 	Context("UpdateResource", func() {
 		It("Adds a non-existing resource", func() {
 			c.UpdateResources(gvk1)
-			Expect(c.Resources).To(Equal([]GVK{gvk1}))
+			Expect(c.Resources).To(Equal([]ResourceData{gvk1}))
 			// Update again to ensure idempotency.
 			c.UpdateResources(gvk1)
-			Expect(c.Resources).To(Equal([]GVK{gvk1}))
+			Expect(c.Resources).To(Equal([]ResourceData{gvk1}))
 		})
 		It("Updates an existing resource", func() {
 			c.UpdateResources(gvk1)
-			gvk := GVK{Group: gvk1.Group, Version: gvk1.Version, Kind: gvk1.Kind, CRDVersion: "v1"}
-			c.UpdateResources(gvk)
-			Expect(c.Resources).To(Equal([]GVK{gvk}))
+			resource := ResourceData{Group: gvk1.Group, Version: gvk1.Version, Kind: gvk1.Kind}
+			c.UpdateResources(resource)
+			Expect(c.Resources).To(Equal([]ResourceData{resource}))
 		})
 		It("Updates an existing resource with more than one resource present", func() {
 			c.UpdateResources(gvk1)
 			c.UpdateResources(gvk2)
-			gvk := GVK{Group: gvk1.Group, Version: gvk1.Version, Kind: gvk1.Kind, CRDVersion: "v1"}
+			gvk := ResourceData{Group: gvk1.Group, Version: gvk1.Version, Kind: gvk1.Kind}
 			c.UpdateResources(gvk)
-			Expect(c.Resources).To(Equal([]GVK{gvk, gvk2}))
+			Expect(c.Resources).To(Equal([]ResourceData{gvk, gvk2}))
 		})
+	})
+
+	Context("HasGroup", func() {
+		It("should return true when config has a resource with the group", func() {
+			c.UpdateResources(gvk1)
+			Expect(c.Resources).To(Equal([]ResourceData{gvk1}))
+			Expect(c.HasGroup(gvk1.Group)).To(BeTrue())
+		})
+		It("should return false when config has a resource with not the same group", func() {
+			c.UpdateResources(gvk1)
+			Expect(c.Resources).To(Equal([]ResourceData{gvk1}))
+			Expect(c.HasGroup("hasNot")).To(BeFalse())
+		})
+
 	})
 })
--- a/pkg/model/resource/options.go
+++ b/pkg/model/resource/options.go
@@ -33,6 +33,7 @@ const (
 	groupRequired   = "group cannot be empty"
 	versionRequired = "version cannot be empty"
 	kindRequired    = "kind cannot be empty"
+	k8sDomain       = "k8s.io"
 )
 
 var (
@@ -81,12 +82,15 @@ type Options struct {
 	Plural string
 
 	// Namespaced is true if the resource is namespaced.
+	// todo: remove when v2 is no longer be supported
+	// this attr was only kept because it still in usage in the v2 go plugin templates
 	Namespaced bool
 
-	// CRDVersion holds the CustomResourceDefinition API version used for the Options.
-	CRDVersion string
-	// WebhookVersion holds the {Validating,Mutating}WebhookConfiguration API version used for the Options.
-	WebhookVersion string
+	// API holds the api data
+	API config.API
+
+	// Webhooks holds the webhooks data
+	Webhooks config.Webhooks
 }
 
 // ValidateV2 verifies that V2 project has all the fields have valid values
@@ -190,8 +194,8 @@ func (opts *Options) Validate() error {
 
 	// Ensure apiVersions for k8s types are empty or valid.
 	for typ, apiVersion := range map[string]string{
-		"CRD":     opts.CRDVersion,
-		"Webhook": opts.WebhookVersion,
+		"CRD":     opts.API.CRDVersion,
+		"Webhook": opts.Webhooks.WebhookVersion,
 	} {
 		switch apiVersion {
 		case "", "v1", "v1beta1":
@@ -205,14 +209,14 @@ func (opts *Options) Validate() error {
 	return nil
 }
 
-// GVK returns the group-version-kind information to check against tracked resources in the configuration file
-func (opts *Options) GVK() config.GVK {
-	return config.GVK{
-		Group:          opts.Group,
-		Version:        opts.Version,
-		Kind:           opts.Kind,
-		CRDVersion:     opts.CRDVersion,
-		WebhookVersion: opts.WebhookVersion,
+// Data returns the ResourceData information to check against tracked resources in the configuration file
+func (opts *Options) Data() config.ResourceData {
+	return config.ResourceData{
+		Group:    opts.Group,
+		Version:  opts.Version,
+		Kind:     opts.Kind,
+		API:      &opts.API,
+		Webhooks: &opts.Webhooks,
 	}
 }
 
@@ -228,9 +232,8 @@ func (opts *Options) safeImport(unsafe string) string {
 }
 
 // NewResource creates a new resource from the options
-func (opts *Options) NewResource(c *config.Config, doResource bool) *Resource {
-	res := opts.newResource()
-
+func (opts *Options) NewResource(c *config.Config, doAPI, doController bool) *Resource {
+	res := opts.newResource(doAPI, doController)
 	replacer := res.Replacer()
 
 	pkg := replacer.Replace(path.Join(c.Repo, "api", "%[version]"))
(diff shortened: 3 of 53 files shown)


pkg/plugin/v3/webhook.go
```diff
@@ -96,6 +107,18 @@ func (p *createWebhookPlugin) Validate() error {
...
+		return errors.New("An webhook for this resource already exists")
+	}
+
+	// Check that the provided group can be added to the project
+	if !p.config.MultiGroup && (len(p.config.Resources) != 0 ||
+		len(p.config.Webhooks) != 0) && !p.config.HasGroup(p.resource.Group) {
+		return fmt.Errorf("multiple groups are not allowed by default, " +
+			"to enable multi-group visit kubebuilder.io/migration/multi-group.html")
+	}
+
```
gabbifish, line 121: Is it worth adding an extra check to ensure that if multigroup mode is on, `len(p.config.Webhooks) <= len(p.config.Resources)`?

gabbifish (commented):
I have some minor nits and a suggestion for an extra validity check, otherwise this looks great!! Your tests are so comprehensive 😍

pkg/model/config/config_test.go
```diff
@@ -29,6 +29,83 @@ var _ = Describe("Config", func() {
 		Data2 string `json:"data-2"`
 	}
 
+	web := NewWebhook(GVK{Group: "Foo", Kind: "Baz", Version: "v1"}, false, false, true)
+	gkv := GVK{Group: "Foo", Kind: "Baz", Version: "v1"}
+
+	It("should return a new Webhook", func() {
+		web := NewWebhook(GVK{Group: "Foo", Kind: "Baz", Version: "v1"}, false, false, true)
```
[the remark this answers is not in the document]
camilamacedo86: good catcher

pkg/plugin/v3/webhook.go
camilamacedo86: The problem is that currently we are able to; 

```
$ kubebuilder create api
$ kubebuilder create api --group ship --version v1beta1 --kind Frigate
$ kubebuilder create webhook --group other --version v1 --kind FirstMate --defaulting --programmatic-validation
```

So, should we allow create a webhook for another group when the multigroup layout/support is not enable?

pkg/plugin/v3/webhook.go
```diff
@@ -96,6 +107,18 @@ func (p *createWebhookPlugin) Validate() error {
...
 	return nil
```
camilamacedo86, line 122: /hold

I found an issue to fix regards the webhooks paths.

camilamacedo86:
/hold cancel

@gabbifish fell free to re-checl

pkg/plugin/v3/webhook.go
gabbifish: Ahh, I understand now! I think #1708 will fix this. Thank you for taking that up in another PR! (I'll try to review it in the next day or so!)

pkg/model/config/config.go
```diff
@@ -93,6 +97,19 @@ func (c Config) HasResource(target GVK) bool {
 	return false
 }
 
+// HasWebhookWithGKV returns true if webkook is already tracked
+func (c Config) HasWebhookWithGKV(target GVK) bool {
```
gabbifish, line 101: minor typo: 
```suggestion
func (c Config) HasWebhookWithGVK(target GVK) bool {
```

pkg/model/config/config.go
```diff
@@ -132,13 +170,38 @@ type GVK struct {
...
+	Version    string `json:"version,omitempty"`
+	Kind       string `json:"kind,omitempty"`
+	Defaulting bool   `json:"defaulting,omitempty"`
+	Validation bool   `json:"validation,omitempty"`
+	Conversion bool   `json:"conversion,omitempty"`
+}
+
+// NewWebhook instantiate an Webhook
+func NewWebhook(gkv GVK, defaulting, validation, conversion bool) Webhook {
+	return Webhook{Group: gkv.Group, Kind: gkv.Kind, Version: gkv.Version,
```
gabbifish, line 185: minor typos: 
```suggestion
	return Webhook{Group: gvk.Group, Kind: gvk.Kind, Version: gvk.Version,
```

gabbifish (commented):
Some tiny nits for `gkv->gvk`--once those are done, I'll approve! :D

gabbifish (approved):
/approve

:D

camilamacedo86:
/hold 

Until getting approval from +2 reviewers.

DirectXMan12:
let's finish the discussion on https://github.com/kubernetes-sigs/kubebuilder/pull/1644#discussion_r491006675 before we move forward here.

camilamacedo86:
Hi @estroz, 

> let's finish the discussion on #1644 (comment) before we move forward here.

I think that we all agree over store the webhook data in the PROJECT file in order to: 

- Allow we have all project config into the config which can, for example, grant we re-create the project based on it.
keep the consistency across the project
- Allow we easily manage diff versions of apiVersions for webhooks and CRDs. See that the apiVersion fits very well as an attribute of its models. So, we could store the apiVersion which would allow how the tool should dealing with each case scenario. 

WDYT? Could we cancel the hold on this one?

estroz:
We still need to discuss how webhook info is stored in the config. I don’t necessarily think we need/want a completely separate field, and instead store webhook booleans in the `resources` field.

We should probably wait until #1644 is merged to finalize ^ because the way webhook info is stored could be impacted by when it is added to the config. In the meantime I’d like to suggest the following:

```yaml
resources:
- group: g
   version: v1
   kind: Kind
   validatingWebhook: true
   mutatingWebhook: true
   conversionWebhook: true
```

camilamacedo86:
Hi @estroz. 

Regards your suggestion on https://github.com/kubernetes-sigs/kubebuilder/pull/1696#issuecomment-710290530 `. I don’t necessarily think we need/want a completely separate field, and instead store webhook` I am afraid that I do NOT agree with your suggestion because they are NOT the same object and structure. It is the equivalent to store in a database the client and its addresses in the same table instead of having one to address the client and another to address and the relation of 1..N between then. 

Would not possible to have different kinds of webhooks for the same resource?

estroz:
>I mean a webhook is NOT a resource

Of course, but they are inextricably linked, i.e. you cannot have a webhook without a resource. This relationship was recently enforced by #1708.

camilamacedo86:
Hi @estroz,

> Of course, but they are inextricably linked, i.e. you cannot have a webhook without a resource. This relationship was recently enforced by #1708. 

The relation is 1..N ( 1 resource for 1 or Many webhooks) by adopting your suggestion done in https://github.com/kubernetes-sigs/kubebuilder/pull/1696#issuecomment-712143182 it would not be possible as the other use cases described in https://github.com/kubernetes-sigs/kubebuilder/pull/1696#issuecomment-712143182.

camilamacedo86:
I cacher your idea now @estroz . It can works. 👍 
/hold 

until we merge #1644 to address this one.

camilamacedo86:
/hold cancel

@estroz it can get merged now.

pkg/model/config/config.go
```diff
@@ -143,34 +158,62 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
...
+	if !r.Webhooks.isEqualTo(other.Webhooks) {
+		r.Webhooks = other.Webhooks
+	}
+}
+
+// isEqualTo compares it with another webhook
+func (w *Webhook) isEqualTo(other Webhook) bool {
+	return w.Conversion == other.Conversion &&
+		w.Defaulting == other.Defaulting &&
+		w.Validation == other.Validation
```
estroz, line 216:
```suggestion
// merge combines fields of two Webhooks, favoring the receiver's values.
func (w *Webhook) merge(other Webhook) {
	// Set each webhook type individually so existing values are not overwritten.
	w.Defaulting = w.Defaulting || other.Defaulting
	w.Validating = w.Validating || other.Validating
	w.Conversion = w.Conversion || other.Conversion
```

pkg/model/config/config.go
```diff
@@ -143,34 +158,62 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
...
+func (r *Resource) merge(other Resource) {
 	if r.CRDVersion == "" && other.CRDVersion != "" {
 		r.CRDVersion = other.CRDVersion
 	}
 	if r.WebhookVersion == "" && other.WebhookVersion != "" {
 		r.WebhookVersion = other.WebhookVersion
 	}
+	if !r.Webhooks.isEqualTo(other.Webhooks) {
+		r.Webhooks = other.Webhooks
+	}
```
estroz, line 209: If you run `create webhook --defaulting` then `create webhook --conversion`, the webhook booleans won't be merged correctly. Instead:
```suggestion

	r.Webhooks.merge(other.Webhooks)
```

pkg/model/config/config_test.go
```diff
@@ -147,58 +192,90 @@ var _ = Describe("PluginConfig", func() {
 		Expect(config.DecodePluginConfig(key, &pluginConfig)).To(Succeed())
 		Expect(pluginConfig).To(Equal(expectedPluginConfig))
 	})
+
+	It("should Marshal and umarchal a plugin", func() {
```
estroz, line 196: Are these cases testing anything added by this PR?

pkg/plugin/v3/scaffolds/webhook.go
```diff
@@ -75,12 +75,14 @@ func (s *webhookScaffolder) newUniverse() *model.Universe {
...
 
 func (s *webhookScaffolder) scaffold() error {
+
 	if s.conversion {
 		fmt.Println(`Webhook server has been set up for you.
 You need to implement the conversion.Hub and conversion.Convertible interfaces for your CRD types.`)
 	}
 
-	s.config.UpdateResources(s.resource.GVK())
+	s.config.AddWebhook(s.resource.Resource(),
```
estroz, line 84: So you don't need a new config method:

You can set webhook booleans on `webhookScaffolder` like:
```go
type webhookScaffolder struct {
	config      *config.Config
	boilerplate string
	resource    *resource.Resource

	// Webhook type options.
	webhooks config.Webhooks
}

func NewWebhookScaffolder(
	config *config.Config,
	boilerplate string,
	resource *resource.Resource,
	webhooks config.Webhooks,
) cmdutil.Scaffolder {
	return &webhookScaffolder{
		config:      config,
		boilerplate: boilerplate,
		resource:    resource,
		webhooks:    webhooks,
	}
}
```

Then do:
```suggestion
	res := s.resource.Resource()
	res.Webhooks = s.webhooks
	s.config.UpdateResource(res)
```

pkg/plugin/v3/scaffolds/webhook.go
```diff
@@ -75,12 +75,14 @@ func (s *webhookScaffolder) newUniverse() *model.Universe {
...
+		config.Webhook{Defaulting: s.defaulting, Validation: s.validation, Conversion: s.conversion})
```
estroz, line 85:
```suggestion
```

pkg/plugin/v3/webhook.go
```diff
@@ -74,6 +78,7 @@ func (p *createWebhookSubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&p.resource.WebhookVersion, "webhook-version", defaultWebhookVersion,
 		"version of {Mutating,Validating}WebhookConfigurations to scaffold. Options: [v1, v1beta1]")
 
+	fs.BoolVar(&p.runMake, "make", true, "if true, run make after generating files")
```
estroz, line 81: Is this required for webhook changes?

pkg/model/config/config.go
```diff
@@ -143,34 +158,62 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
...
 	WebhookVersion string `json:"webhookVersion,omitempty"`
+
+	// Webhooks holds the Webhooks data
+	Webhooks Webhook `json:"webhooks,omitempty"`
 }
 
-// isEqualTo compares it with another resource
-func (r GVK) isEqualTo(other GVK) bool {
+// Webhook contains information about scaffolded webhooks
+type Webhook struct {
```
estroz, line 177: This should be plural since it contains information about multiple webhook types
```suggestion
type Webhooks struct {
```

pkg/plugin/v3/webhook.go
```diff
@@ -42,6 +43,9 @@ type createWebhookSubcommand struct {
 	defaulting bool
 	validation bool
 	conversion bool
```
estroz, line 45:
```suggestion
	webhooks config.Webhooks
```

pkg/model/config/config.go
```diff
@@ -143,34 +158,62 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
...
+	Defaulting bool `json:"defaulting,omitempty"`
+	Validation bool `json:"validation,omitempty"`
+	Conversion bool `json:"conversion,omitempty"`
```
estroz, line 180: These will be unmarshaled weirdly if not set, ex `webhooks: {}`. I think this should work
```suggestion
	Defaulting bool `json:"defaulting"`
	Validation bool `json:"validation"`
	Conversion bool `json:"conversion"`
```

estroz (changes requested):

pkg/model/config/config.go
```diff
@@ -143,34 +158,62 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
...
+// AddWebhook add the webhook values to the resource with the gkv informed
+func AddWebhook(resource Resource, defaulting, validation, conversion bool) Resource {
+	resource.Webhooks.Defaulting = defaulting
+	resource.Webhooks.Validation = validation
+	resource.Webhooks.Conversion = conversion
+	return resource
+}
+
+// hasSameGKV compares it with another resource
+func (r *Resource) hasSameGKV(other Resource) bool {
```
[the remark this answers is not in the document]
camilamacedo86: Great suggestion 🥇

pkg/plugin/v3/webhook.go
camilamacedo86: The initial idea here is to keep it == resources. See `store webhooks config info and add make run (go/v3-alpha)`
However, I agree that it might better be done in another PR.

pkg/model/config/config.go
camilamacedo86: `omitempty` means that the attributes can be empty then `webhooks: {}` means that it is empty which matches with how its rules. I mean, I resource can have 0 to Many webhooks. 

However, if we follow up your suggestion it is ok but we will store all attributes with the false value always when none webhooks were scaffold which is unnecessary . E.g 

```yaml
webhooks:
    conversion: false
    defaulting: false
    validation: false

```

So, I think that keep `omitempty` (allows be empty) still better.

pkg/model/config/config_test.go
camilamacedo86: Moved to another PR: https://github.com/kubernetes-sigs/kubebuilder/pull/1844/files

pkg/model/config/config.go
```diff
@@ -143,34 +158,62 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
...
+	Validation bool `json:"validation,omitempty"`
+	Conversion bool `json:"conversion,omitempty"`
+}
+
+// AddWebhook add the webhook values to the resource with the gkv informed
+func AddWebhook(resource Resource, defaulting, validation, conversion bool) Resource {
+	resource.Webhooks.Defaulting = defaulting
+	resource.Webhooks.Validation = validation
+	resource.Webhooks.Conversion = conversion
+	return resource
```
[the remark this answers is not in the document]
camilamacedo86: It was removed.

pkg/model/config/config.go
camilamacedo86: good catcher.

camilamacedo86:
Hi @estroz, 

All your suggestions are addressed.  Only this one https://github.com/kubernetes-sigs/kubebuilder/pull/1696#discussion_r525437783 is not because I an afraid that I do not agree with it. I hope that it can get merged now. Really tks for your review.

pkg/model/config/config.go
estroz: I think I'm ok with this.

pkg/model/resource/resource.go
```diff
@@ -54,13 +55,17 @@ type Resource struct {
 
 	// CRDVersion holds the CustomResourceDefinition API version used for the Resource.
 	CRDVersion string `json:"crdVersion,omitempty"`
+
 	// WebhookVersion holds the {Validating,Mutating}WebhookConfiguration API version used for the Resource.
 	WebhookVersion string `json:"webhookVersion,omitempty"`
+
+	// Webhooks holds the webhook data used to do th scaffolds
+	Webhooks config.Webhooks `json:"webhooks,omitempty"`
```
estroz, line 63: Is this actually used anywhere?

pkg/plugins/golang/v3/scaffolds/webhook.go
```diff
@@ -37,26 +37,22 @@ type webhookScaffolder struct {
 	boilerplate string
 	resource    *resource.Resource
 
-	// Webhook type options.
-	defaulting, validation, conversion bool
+	// Webhooks type options.
+	webhook *config.Webhooks
```
estroz, line 41: This doesn't need to be a pointer
```suggestion
	webhook config.Webhooks
```

pkg/plugins/golang/v3/scaffolds/webhook.go
```diff
@@ -37,26 +37,22 @@ type webhookScaffolder struct {
...
 
 // NewWebhookScaffolder returns a new Scaffolder for v2 webhook creation operations
 func NewWebhookScaffolder(
 	config *config.Config,
 	boilerplate string,
 	resource *resource.Resource,
-	defaulting bool,
-	validation bool,
-	conversion bool,
+	webhook *config.Webhooks,
```
estroz, line 49:
```suggestion
	webhook config.Webhooks,
```

pkg/plugins/golang/v3/webhook.go
```diff
@@ -123,7 +123,8 @@ func (p *createWebhookSubcommand) GetScaffolder() (cmdutil.Scaffolder, error) {
 
 	// Create the actual resource from the resource options
 	res := p.resource.NewResource(p.config, false)
-	return scaffolds.NewWebhookScaffolder(p.config, string(bp), res, p.defaulting, p.validation, p.conversion), nil
+	webhook := config.Webhooks{Defaulting: p.defaulting, Validation: p.validation, Conversion: p.conversion}
```
estroz, line 126: `createWebhookSubcommand` should have a [`webhooks config.Webhooks`](https://github.com/kubernetes-sigs/kubebuilder/pull/1696/#discussion_r525435638) field.

pkg/model/config/config_test.go
```diff
@@ -168,53 +205,53 @@ var _ = Describe("PluginConfig", func() {
 var _ = Describe("Resource Version Compatibility", func() {
 
 	var (
-		c          *Config
-		gvk1, gvk2 GVK
+		c                    *Config
+		Resource1, Resource2 Resource
```
Adirio, line 209:
```suggestion
		resource1, resource2 Resource
```

pkg/model/config/config_test.go
```diff
@@ -168,53 +205,53 @@ var _ = Describe("PluginConfig", func() {
...
 
 		defaultVersion = "v1"
 	)
 
 	BeforeEach(func() {
 		c = &Config{}
-		gvk1 = GVK{Group: "example", Version: "v1", Kind: "TestKind"}
-		gvk2 = GVK{Group: "example", Version: "v1", Kind: "TestKind2"}
+		Resource1 = Resource{Group: "example", Version: "v1", Kind: "TestKind"}
+		Resource2 = Resource{Group: "example", Version: "v1", Kind: "TestKind2"}
```
Adirio, line 217:
```suggestion
		resource1 = Resource{Group: "example", Version: "v1", Kind: "TestKind"}
		resource2 = Resource{Group: "example", Version: "v1", Kind: "TestKind2"}
```

pkg/model/config/config_test.go
```diff
@@ -168,53 +205,53 @@ var _ = Describe("PluginConfig", func() {
...
 	})
 
 	Context("resourceAPIVersionCompatible", func() {
 		It("returns true for a list of empty resources", func() {
 			Expect(c.resourceAPIVersionCompatible("crd", defaultVersion)).To(BeTrue())
 		})
 		It("returns true for one resource with an empty version", func() {
-			c.Resources = []GVK{gvk1}
+			c.Resources = []Resource{Resource1}
```
Adirio, line 225:
```suggestion
			c.Resources = []Resource{resource1}
```

pkg/model/config/config_test.go
```diff
@@ -168,53 +205,53 @@ var _ = Describe("PluginConfig", func() {
...
 			Expect(c.resourceAPIVersionCompatible("crd", defaultVersion)).To(BeTrue())
 		})
 		It("returns true for one resource with matching version", func() {
-			gvk1.CRDVersion = defaultVersion
-			c.Resources = []GVK{gvk1}
+			Resource1.CRDVersion = defaultVersion
+			c.Resources = []Resource{Resource1}
```
Adirio, line 230:
```suggestion
			resource1.CRDVersion = defaultVersion
			c.Resources = []Resource{resource1}
```

pkg/model/config/config_test.go
```diff
@@ -168,53 +205,53 @@ var _ = Describe("PluginConfig", func() {
...
 			Expect(c.resourceAPIVersionCompatible("crd", defaultVersion)).To(BeTrue())
 		})
 		It("returns true for two resources with matching versions", func() {
-			gvk1.CRDVersion = defaultVersion
-			gvk2.CRDVersion = defaultVersion
-			c.Resources = []GVK{gvk1, gvk2}
+			Resource1.CRDVersion = defaultVersion
+			Resource2.CRDVersion = defaultVersion
+			c.Resources = []Resource{Resource1, Resource2}
```
Adirio, line 236:
```suggestion
			resource1.CRDVersion = defaultVersion
			resource2.CRDVersion = defaultVersion
			c.Resources = []Resource{resource1, resource2}
```

pkg/model/config/config_test.go
```diff
@@ -168,53 +205,53 @@ var _ = Describe("PluginConfig", func() {
...
 			Expect(c.resourceAPIVersionCompatible("crd", defaultVersion)).To(BeTrue())
 		})
 		It("returns false for one resource with a non-matching version", func() {
-			gvk1.CRDVersion = v1beta1
-			c.Resources = []GVK{gvk1}
+			Resource1.CRDVersion = v1beta1
+			c.Resources = []Resource{Resource1}
```
Adirio, line 241:
```suggestion
			resource1.CRDVersion = v1beta1
			c.Resources = []Resource{resource1}
```

pkg/model/config/config_test.go
```diff
@@ -168,53 +205,53 @@ var _ = Describe("PluginConfig", func() {
...
 			Expect(c.resourceAPIVersionCompatible("crd", defaultVersion)).To(BeFalse())
 		})
 		It("returns false for two resources containing a non-matching version", func() {
-			gvk1.CRDVersion = v1beta1
-			gvk2.CRDVersion = defaultVersion
-			c.Resources = []GVK{gvk1, gvk2}
+			Resource1.CRDVersion = v1beta1
+			Resource2.CRDVersion = defaultVersion
+			c.Resources = []Resource{Resource1, Resource2}
```
Adirio, line 247:
```suggestion
			resource1.CRDVersion = v1beta1
			resource2.CRDVersion = defaultVersion
			c.Resources = []Resource{resource1, resource2}
```

pkg/model/config/config_test.go
```diff
@@ -168,53 +205,53 @@ var _ = Describe("PluginConfig", func() {
...
 			Expect(c.resourceAPIVersionCompatible("crd", defaultVersion)).To(BeFalse())
 		})
 
 		It("returns false for two resources containing a non-matching version (webhooks)", func() {
-			gvk1.WebhookVersion = v1beta1
-			gvk2.WebhookVersion = defaultVersion
-			c.Resources = []GVK{gvk1, gvk2}
+			Resource1.WebhookVersion = v1beta1
+			Resource2.WebhookVersion = defaultVersion
+			c.Resources = []Resource{Resource1, Resource2}
```
Adirio, line 254:
```suggestion
			resource1.WebhookVersion = v1beta1
			resource2.WebhookVersion = defaultVersion
			c.Resources = []Resource{resource1, resource2}
```

Adirio (changes requested):
You changed `GVK` to `Resource`. I do agree that `GVK` is no longer a good name, but `Resource` is not a good name either because we already have one type called like that. Maybe `ResourceData`?

Additionally I don't like the `webhooks: {}` added to every resource. I think we can avoid this by making webhooks being `[]Webhook` where:
```
type Webhook uint8

const (
	Defaulting Webhook = iota
	Validation
	Conversion
)

func (w Webhook) String() string {
	switch w {
	case Defaulting:
		return "defaulting"
	case Validation:
		return "validation"
	case Conversion:
		return "conversion"
	default:
		return "unknown webhook type"
	}
}
```
It would require parsing from the struct to a slice and the other way round but the output would be none if no webhook is implemented.

camilamacedo86:
Hi @Adirio, 

Thank you for your review. All suggestions are addressed less over the webhook types done in https://github.com/kubernetes-sigs/kubebuilder/pull/1696#pullrequestreview-536457035. 

If I did no misunderstood your suggestions, IMO it will increase the complexity to maintain/dealing with so I am not sure if it would be a good approach only for we do not have the `webhooks : {}`when it is empty in the PROJECT file which is used only by tool and not really by users. 

So, WDYT about we move forward with it as it is now since it allows us to achieve its goal and then you push a PR with your suggestion as a follow-up?

pkg/model/resource/resource.go
```diff
@@ -54,13 +55,14 @@ type Resource struct {
...
+
+	// WebhookVersion holds the {Validating,Mutating}WebhookConfiguration API version used for the ResourceData.
 	WebhookVersion string `json:"webhookVersion,omitempty"`
 }
 
-// GVK returns the group-version-kind information to check against tracked resources in the configuration file
-func (r *Resource) GVK() config.GVK {
-	return config.GVK{
+// ResourceData returns the group-version-kind information to check against tracked resources in the configuration file
+func (r *Resource) ResourceData() config.ResourceData {
```
Adirio, line 64: `Resource.ResourceData` seems unnecesarily verbose:
```suggestion
// Data returns the group-version-kind information to check against tracked resources in the configuration file
func (r *Resource) Data() config.ResourceData {
```

pkg/plugins/golang/v2/api.go
```diff
@@ -143,7 +143,7 @@ func (p *createAPISubcommand) Validate() error {
 	// In case we want to scaffold a resource API we need to do some checks
 	if p.doResource {
 		// Check that resource doesn't exist or flag force was set
-		if !p.force && p.config.HasResource(p.resource.GVK()) {
+		if !p.force && p.config.HasResource(p.resource.ResourceData()) {
```
Adirio, line 146: If you apply the above change, this would be:
```suggestion
		if !p.force && p.config.HasResource(p.resource.Data()) {
```
which IMO is easier to read.

(There are other places too with similar statements, won't mark them all to avoid noise.)

pkg/plugins/golang/v3/webhook.go
```diff
@@ -74,17 +72,17 @@ func (p *createWebhookSubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&p.resource.Group, "group", "", "resource Group")
 	fs.StringVar(&p.resource.Version, "version", "", "resource Version")
 	fs.StringVar(&p.resource.Kind, "kind", "", "resource Kind")
-	fs.StringVar(&p.resource.Plural, "resource", "", "resource Resource")
+	fs.StringVar(&p.resource.Plural, "resource", "", "resource ResourceData")
```
Adirio, line 75: This is not refering to the same Resource so it shouldnt be changed.

Adirio (changes requested):

estroz:
Visually speaking I do prefer not writing a `resources.webhooks` field to the config when there are no webhooks, although I did say I think I would be earlier. Adding `webhooks: {}` to the config file is confusing, even though users don't typically look at that file; I'd rather the config file not mention webhooks at all until a user adds them, so using a type that is not marshaled when empty like a slice makes more sense to me.

estroz:
Suggested alternative on where to add these values: https://github.com/kubernetes-sigs/kubebuilder/issues/1826#issuecomment-732460408.

camilamacedo86:
Hi @estroz and @Adirio, 

All suggestion made was addressed less the https://github.com/kubernetes-sigs/kubebuilder/pull/1696#pullrequestreview-536457035 because I do not agree with it.

In POV it reduces the maintainability and increases the complexity a lot since it will not allow us to add new attr for the webhooks and make us parse/unparse strings. Because of this, as recommended in the comment https://github.com/kubernetes-sigs/kubebuilder/pull/1696#issuecomment-732220971 we move forward here as it is now shaped with all other suggestions made so far and if the suggestion author (@Adirio ) or any other person would like to change it make a PR with his/her ideas. In this way, I believe that the reasons for I do not agree with it will be clarified and also if I am wrong here or missing something I will be very happy to review and approve the suggestion as well.

pkg/model/config/config.go
```diff
@@ -45,7 +45,7 @@ type Config struct {
 
 	// Resources tracks scaffolded resources in the project
 	// This info is tracked only in project with version 2
-	Resources []GVK `json:"resources,omitempty"`
+	Resources []Data `json:"resources,omitempty"`
```
Adirio, line 48: Can we call it `ResourceData`? `Resource` collided with the other type but `Data` seems pretty generic.
```suggestion
	Resources []ResourceData `json:"resources,omitempty"`
```

pkg/model/config/config.go
```diff
@@ -147,34 +147,53 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
...
+type Webhooks struct {
+	Defaulting bool `json:"defaulting,omitempty"`
+	Validation bool `json:"validation,omitempty"`
+	Conversion bool `json:"conversion,omitempty"`
 }
 
-// isEqualTo compares it with another resource
-func (r GVK) isEqualTo(other GVK) bool {
+// isGVKEqualTo compares it with another resource
+func (r *Data) isGVKEqualTo(other Data) bool {
```
Adirio, line 173: The name should be updated to reflect the type rename from `GVK` to `ResourceData`. Also, this method does not modify `r` and thus a non-pointer receiver should be preferred.
```suggestion
// isSameResource compares it with another resource
func (r ResourceData) isSameResource(other ResourceData) bool {
```

pkg/model/config/config.go
```diff
@@ -147,34 +147,53 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
...
 
-	// CRDVersion holds the CustomResourceDefinition API version used for the GVK.
+	// CRDVersion holds the CustomResourceDefinition API version used for the Data.
 	CRDVersion string `json:"crdVersion,omitempty"`
-	// WebhookVersion holds the {Validating,Mutating}WebhookConfiguration API version used for the GVK.
+	// WebhookVersion holds the {Validating,Mutating}WebhookConfiguration API version used for the Data.
 	WebhookVersion string `json:"webhookVersion,omitempty"`
+
+	// Webhooks holds the Webhooks data
+	Webhooks Webhooks `json:"webhooks,omitempty"`
```
Adirio, line 162: I think that making this field a pointer and setting it to nil when all of them are false should not print the `webhooks: {}` that we were complaining about.
```suggestion
	Webhooks *Webhooks `json:"webhooks,omitempty"`
```

pkg/model/config/config_test.go
```diff
@@ -31,6 +31,43 @@ var _ = Describe("PluginConfig", func() {
...
+	resource.Webhooks = Webhooks{false, false, true}
+
+	It("should return a new Webhooks", func() {
+		Expect(resource.Group).To(BeIdenticalTo("Foo"))
+		Expect(resource.Kind).To(BeIdenticalTo("Baz"))
+		Expect(resource.Version).To(BeIdenticalTo("v1"))
+		Expect(resource.Webhooks.Defaulting).To(BeIdenticalTo(false))
+		Expect(resource.Webhooks.Validation).To(BeIdenticalTo(false))
+		Expect(resource.Webhooks.Conversion).To(BeIdenticalTo(true))
+	})
```
Adirio, line 44: This is not testing anything, you are not calling any function or methid, you just assign and check that they are assigned.

pkg/model/config/config_test.go
```diff
@@ -31,6 +31,43 @@ var _ = Describe("PluginConfig", func() {
...
+
+	It("should merge webhook when webhook is NOT equals", func() {
+		resourceOther := Data{Group: "Foo", Kind: "Baz", Version: "v1"}
+		resourceOther.Webhooks = Webhooks{false, false, false}
+		resource.merge(resourceOther)
+		Expect(resourceOther.Webhooks.Conversion).To(BeIdenticalTo(false))
```
Adirio, line 50: NIT: `resourceOther` doesn't sound like an english variable name, `otherResource` sounds better to me
NIT: `BeFalse()` is clearer than `BeIdenticalTo(false)`
```suggestion
		otherResource := Data{Group: "Foo", Kind: "Baz", Version: "v1"}
		otherResource.Webhooks = Webhooks{false, false, false}
		resource.merge(otherResource)
		Expect(otherResource.Webhooks.Conversion).To(BeFalse())
```

pkg/model/config/config_test.go
```diff
@@ -31,6 +31,43 @@ var _ = Describe("PluginConfig", func() {
...
+	})
+
+	It("should return true when has the Data is equals", func() {
+		Expect(resource.isGVKEqualTo(Data{Group: "Foo", Kind: "Baz", Version: "v1"})).To(BeIdenticalTo(true))
```
Adirio, line 54: NIT:  `BeTrue()` is clearer than `BeIdenticalTo(true)`
```suggestion
		Expect(resource.isGVKEqualTo(Data{Group: "Foo", Kind: "Baz", Version: "v1"})).To(BeTrue())
```

pkg/model/config/config_test.go
```diff
@@ -31,6 +31,43 @@ var _ = Describe("PluginConfig", func() {
...
+	})
+
+	It("should return false when Data is NOT equals", func() {
+		Expect(resource.isGVKEqualTo(Data{Group: "Foo", Kind: "Baz", Version: "v2"})).To(BeIdenticalTo(false))
```
Adirio, line 58:
```suggestion
		Expect(resource.isGVKEqualTo(Data{Group: "Foo", Kind: "Baz", Version: "v2"})).To(BeFalse())
```

pkg/model/config/config_test.go
```diff
@@ -31,6 +31,43 @@ var _ = Describe("PluginConfig", func() {
...
+	})
+
+	It("IsV2 should return true when the config is V2 or v3", func() {
```
Adirio, line 61:
```suggestion
	It("IsV2 should return true when the config is v2", func() {
```

pkg/model/config/config_test.go
```diff
@@ -31,6 +31,43 @@ var _ = Describe("PluginConfig", func() {
...
+		cfg := Config{Version: Version2}
+		Expect(cfg.IsV2()).To(BeIdenticalTo(true))
```
Adirio, line 63:
```suggestion
		Expect(cfg.IsV2()).To(BeTrue())
```

pkg/model/config/config_test.go
```diff
@@ -31,6 +31,43 @@ var _ = Describe("PluginConfig", func() {
...
+	})
+
+	It("IsV3 should return true when the config is V3", func() {
+		cfg := Config{Version: Version3Alpha}
+		Expect(cfg.IsV3()).To(BeIdenticalTo(true))
```
Adirio, line 68:
```suggestion
		Expect(cfg.IsV3()).To(BeTrue())
```

pkg/model/config/config_test.go
```diff
@@ -165,56 +202,56 @@ var _ = Describe("PluginConfig", func() {
...
-			gvk1.CRDVersion = v1beta1
-			gvk2.CRDVersion = defaultVersion
-			c.Resources = []GVK{gvk1, gvk2}
+			resource1.CRDVersion = v1beta1
+			resource2.CRDVersion = defaultVersion
+			c.Resources = []Data{resource1, resource2}
 			Expect(c.resourceAPIVersionCompatible("crd", defaultVersion)).To(BeFalse())
 		})
 
 		It("returns false for two resources containing a non-matching version (webhooks)", func() {
```
Adirio, line 251: Could you add the webhook tests for the four options above instead of just the last one only as we have right now?

pkg/model/resource/options.go
```diff
@@ -190,8 +190,8 @@ func (opts *Options) Validate() error {
 
 	// Ensure apiVersions for k8s types are empty or valid.
 	for typ, apiVersion := range map[string]string{
-		"CRD":     opts.CRDVersion,
-		"Webhook": opts.WebhookVersion,
+		"CRD":      opts.CRDVersion,
+		"Webhooks": opts.WebhookVersion,
```
Adirio, line 194: Why plural?

pkg/model/resource/resource.go
```diff
@@ -54,13 +55,14 @@ type Resource struct {
 
 	// CRDVersion holds the CustomResourceDefinition API version used for the Resource.
 	CRDVersion string `json:"crdVersion,omitempty"`
-	// WebhookVersion holds the {Validating,Mutating}WebhookConfiguration API version used for the Resource.
+
+	// WebhookVersion holds the {Validating,Mutating}WebhookConfiguration API version used for the Data.
```
Adirio, line 59:
```suggestion
	// WebhookVersion holds the {Validating,Mutating}WebhookConfiguration API version used for the Resource.
```
The automatic replacement of `Resource` to `Data` also changed this but in this case it is refering to the `Resource` type.

pkg/model/resource/resource.go
```diff
@@ -54,13 +55,14 @@ type Resource struct {
...
 	WebhookVersion string `json:"webhookVersion,omitempty"`
 }
 
-// GVK returns the group-version-kind information to check against tracked resources in the configuration file
-func (r *Resource) GVK() config.GVK {
-	return config.GVK{
+// Data returns the group-version-kind information to check against tracked resources in the configuration file
+func (r *Resource) Data() config.Data {
```
Adirio, line 64: This method should stay as `Data` even after changing the type to `ResourceData` as the `Resource` part is inplicit (`Resource` type).

pkg/plugins/golang/v3/scaffolds/webhook.go
```diff
@@ -75,19 +71,22 @@ func (s *webhookScaffolder) newUniverse() *model.Universe {
...
-	if s.conversion {
+
+	if s.webhook.Conversion {
 		fmt.Println(`Webhook server has been set up for you.
 You need to implement the conversion.Hub and conversion.Convertible interfaces for your CRD types.`)
 	}
 
-	s.config.UpdateResources(s.resource.GVK())
+	res := s.resource.Data()
+	res.Webhooks = s.webhook
```
Adirio, line 81: If you change the type to a pointer you also have to assign a pointer here. We know that at least one must be set so no need to check if all of them are false to use `nil` instead of `&s.webhook`
```suggestion
	res.Webhooks = &s.webhook
```

pkg/plugins/golang/v3/scaffolds/webhook.go
```diff
@@ -37,26 +37,22 @@ type webhookScaffolder struct {
 	boilerplate string
 	resource    *resource.Resource
 
-	// Webhook type options.
-	defaulting, validation, conversion bool
+	// Webhooks type options.
+	webhook config.Webhooks
```
Adirio, line 41: Why not plural? This is actually the struct that determines which webhooks are implemented.

Adirio (changes requested):

pkg/model/config/config.go
```diff
@@ -147,34 +147,53 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
...
 // favoring the receiver's values.
-func (r *GVK) merge(other GVK) {
+func (r *Data) merge(other Data) {
 	if r.CRDVersion == "" && other.CRDVersion != "" {
 		r.CRDVersion = other.CRDVersion
 	}
 	if r.WebhookVersion == "" && other.WebhookVersion != "" {
 		r.WebhookVersion = other.WebhookVersion
 	}
+	r.Webhooks.merge(other.Webhooks)
```
Adirio, line 188: This also needs to be updated.

pkg/model/config/config.go
camilamacedo86: isGVKEqualTo -> only checks if is the same GKV because of this we cannot change this one. we are no comparing if is the same resource data.

pkg/model/config/config.go
camilamacedo86: It was resource data before. However, you suggested it be changed for Data. But it is fine, done 👍

pkg/model/config/config.go
camilamacedo86: Good excellent catcher and suggestion 👍

pkg/model/config/config_test.go
camilamacedo86: Good cather. After the changes made it is no longer required.

pkg/plugins/golang/v3/scaffolds/webhook.go
camilamacedo86: Good nit catcher.

pkg/model/config/config.go
Adirio: Okey I see your point, but please remove the `*` as it doesn't modify the values so no need for a pointer receiver.

pkg/model/config/config.go
Adirio: Before, in https://github.com/kubernetes-sigs/kubebuilder/pull/1696#discussion_r528867742, I just meant to change to `Data` the method mentioned in https://github.com/kubernetes-sigs/kubebuilder/pull/1696#discussion_r530974252, but I guess I wasn't clear enough. I think that it is clear now. `ResourceData` for the type, `Data` for the method as the `Resource` is implied.

pkg/model/config/config.go
camilamacedo86: Done with caveats but to achieve same 👍  

```go
if other.Webhooks != nil {
	r.Webhooks.merge(*other.Webhooks)
}
```

And then, centralize the logic in the merge instead of it: 

```go
func (w *Webhooks) merge(other Webhooks) {
	if w == nil {
		w = &other
		return
	}

```

camilamacedo86:
Hi @Adirio, 

Excellent suggestion for we work with pointers. The changes requested are done and your and @estroz concerns over  `webhooks: {}` raise in https://github.com/kubernetes-sigs/kubebuilder/pull/1696#pullrequestreview-536457035 and https://github.com/kubernetes-sigs/kubebuilder/pull/1696#issuecomment-732329167 are solved.

Really thank you for your collab and your terrific review. 🥇

pkg/model/config/config.go
```diff
@@ -147,34 +147,58 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
...
 		r.CRDVersion = other.CRDVersion
 	}
 	if r.WebhookVersion == "" && other.WebhookVersion != "" {
 		r.WebhookVersion = other.WebhookVersion
 	}
+	if other.Webhooks != nil {
+		if r.Webhooks == nil {
+			r.Webhooks = other.Webhooks
+		}
+		r.Webhooks.merge(*other.Webhooks)
```
Adirio, line 192: If you assign it directly you do not need to merge it, it is already merged.
```suggestion
		} else {
			r.Webhooks.merge(other.Webhooks)
		}
```

pkg/model/config/config.go
```diff
@@ -147,34 +147,58 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
...
+	}
+}
+
+// merge compares it with another webhook by setting each webhook type individually so existing values are
+// not overwritten.
+func (w *Webhooks) merge(other Webhooks) {
```
Adirio, line 198: You have the pointer so its easier to change this signature to receive a pointer that resolving the pointer to pass it to this function and it will perform better.
```suggestion
func (w *Webhooks) merge(other *Webhooks) {
```

pkg/model/config/config_test.go
```diff
@@ -31,6 +31,34 @@ var _ = Describe("PluginConfig", func() {
...
 	}
 
+	resource := ResourceData{Group: "Foo", Kind: "Baz", Version: "v1"}
+	resource.Webhooks = &Webhooks{false, false, true}
+
+	It("should merge webhook when webhook is NOT equals", func() {
+		otherResource := ResourceData{Group: "Foo", Kind: "Baz", Version: "v1"}
+		otherResource.Webhooks = &Webhooks{false, false, false}
+		resource.merge(otherResource)
+		Expect(otherResource.Webhooks.Conversion).To(BeFalse())
```
Adirio, line 41:
```suggestion
		Expect(resource.Webhooks.Conversion).To(BeTrue())
```

pkg/model/config/config_test.go
```diff
@@ -31,6 +31,45 @@ var _ = Describe("PluginConfig", func() {
...
+		Expect(resource.Webhooks.Validation).To(BeTrue())
+		Expect(resource.Webhooks.Conversion).To(BeTrue())
+	})
+
+	It("should merge not update the conversion and validation to false when it is true already", func() {
+		otherResource := ResourceData{Group: "Foo", Kind: "Baz", Version: "v1"}
+		otherResource.Webhooks = &Webhooks{false, false, false}
+		resource.merge(otherResource)
+		Expect(resource.Webhooks.Defaulting).To(BeFalse())
+		Expect(resource.Webhooks.Validation).To(BeTrue())
```
Adirio, line 51:
```suggestion
		Expect(resource.Webhooks.Validation).To(BeFalse())
```

camilamacedo86: It needs to be true. Note that we updated this value in the test above.

Adirio: You are right. I don't know why I expected resource to be reseted. You can resolve this comment.

Adirio (approved):
As I talked offline with Camila, waiting until we reach a consensus on what to do with the project file.

Codewise, LGTM.

camilamacedo86:
/hold

we are trying to reach a consensus to move forward here.

pkg/model/config/config.go
```diff
@@ -136,9 +134,17 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
 		var currVersion string
 		switch verType {
 		case "crd":
-			currVersion = res.CRDVersion
+			if res.API != nil {
+				currVersion = res.API.CRDVersion
+			} else {
+				currVersion = ""
+			}
```
estroz, line 141:
```suggestion
			if res.API != nil {
				currVersion = res.API.CRDVersion
			}
```

pkg/model/config/config.go
```diff
@@ -136,9 +134,17 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
...
 		case "webhook":
-			currVersion = res.WebhookVersion
+			if res.Webhooks != nil {
+				currVersion = res.Webhooks.WebhookVersion
+			} else {
+				currVersion = ""
+			}
```
estroz, line 147:
```suggestion
			if res.Webhooks != nil {
				currVersion = res.Webhooks.WebhookVersion
			}
```

pkg/model/config/config.go
```diff
@@ -147,34 +153,86 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
...
-func (r GVK) isEqualTo(other GVK) bool {
+// Webhooks contains information about scaffolded webhooks
+type Webhooks struct {
+	// Types which are scaffolded
+	Defaulting bool `json:"defaulting,omitempty"`
+	Validation bool `json:"validation,omitempty"`
+	Conversion bool `json:"conversion,omitempty"`
+
+	// WebhookVersion holds the {Validating,Mutating}WebhookConfiguration API version used for the Options.
+	WebhookVersion string `json:"apiVersion,omitempty"`
```
estroz, line 189:
```suggestion
	WebhookVersion string `json:"webhookVersion,omitempty"`
```

pkg/model/config/config.go
```diff
@@ -147,34 +153,86 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
...
+		}
 	}
-	if r.WebhookVersion == "" && other.WebhookVersion != "" {
-		r.WebhookVersion = other.WebhookVersion
+}
+
+// merge compares it with another webhook by setting each webhook type individually so existing values are
+// not overwritten.
+func (w *Webhooks) merge(other *Webhooks) {
+	w.Defaulting = w.Defaulting || other.Defaulting
```
estroz, line 224:
```suggestion
	if w.WebhookVersion == "" && other.WebhookVersion != "" {
		w.WebhookVersion = other.WebhookVersion
	}
	w.Defaulting = w.Defaulting || other.Defaulting
```

pkg/model/resource/options.go
```diff
@@ -228,9 +230,15 @@ func (opts *Options) safeImport(unsafe string) string {
...
 // NewResource creates a new resource from the options
-func (opts *Options) NewResource(c *config.Config, doResource bool) *Resource {
+func (opts *Options) NewResource(c *config.Config, doResource, doController bool) *Resource {
 	res := opts.newResource()
 
+	// the following check was only add to ensure that the behaviour would not
+	// be changed for v2 projects. This if can be removed when v2 is no longer supported
+	if !c.IsV2() {
+		res.Controller = doController
+	}
```
estroz, line 240: Setting any resource fields should be encapsulated in related methods, ex
```suggestion
	res := opts.newResource(doController)
```

pkg/plugins/golang/v3/api.go
```diff
@@ -57,6 +57,7 @@ type createAPISubcommand struct {
 	pattern string
 
 	resource *resource.Options
+	api      config.API
```
estroz, line 60: Options already contains a `config.API`
```suggestion
	resource *resource.Options
```

estroz (changes requested):
There seems to be a mismatch of data between `config.ResourceData`, `resource.Resource`, and `resource.Options`. The latter two are constructed by the plugin layer, which passes a `resource.Resource` to the scaffold layer to be used and updated. Therefore all of these structs should have the same field subset relating to a resource as an element of `config.ResourceData.Resources`.

pkg/model/resource/options.go
camilamacedo86: the `opts.newResource` has not the config. It returns a new resource from opts and not with all data.

camilamacedo86:
/hold cancel

It has all suggested attrs in the format agreed as it is addressing all suggested made.

pkg/model/resource/resource.go
```diff
@@ -43,29 +44,41 @@ type Resource struct {
 	// ImportAlias is a cleaned concatenation of Group and Version.
 	ImportAlias string `json:"-"`
 
-	// Package is the go package of the Resource.
-	Package string `json:"package,omitempty"`
+	// Endpoint is the API endpoint which is used import the go package of the Resource.
+	Endpoint string `json:"endpoint,omitempty"`
```
estroz, line 48: I don't understand why this is called `endpoint` now, and is more confusing than `path` or `package`.

camilamacedo86: Because `k8s.io/api/apps/v1` is a Rest API endpoint.

estroz: 1. This value is not related to the REST endpoint in this context, it is related to the native type import path. In any case why would a `Resource` ever need to know about a REST endpoint?
1. k8s REST endpoints have a [different structure](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-groups-and-versioning). Ex. import `k8s.io/api/apps/v1` might have a REST endpoint `/apis/apps/v1`, and `k8s.io/api/rbac/v1` might have `/apis/rbac.authorization.k8s.io/v1`.

camilamacedo86: You are definitely right here. Good point. So, yes. I would agree that path is a better name either. 👍

camilamacedo86:
Close in favor of  #1869

PR #1698, opened 2020-09-26
Update flect from v0.2.1 to v0.2.2

grdryn:
This module is used for pluralization of kinds here in
kubebuilder. There's a fix in v0.2.2 for the pluralization of a kind
name in an operator that I'm working with, where it's currently
getting the incorrect pluralization.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module sigs.k8s.io/kubebuilder
 go 1.13
 
 require (
-	github.com/gobuffalo/flect v0.2.1
+	github.com/gobuffalo/flect v0.2.2
 	github.com/onsi/ginkgo v1.12.0
 	github.com/onsi/gomega v1.9.0
 	github.com/spf13/afero v1.2.2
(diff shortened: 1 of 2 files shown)


gabbifish:
Hi @grdryn! Thank you for submitting this! Before I review this, I'd love more context about the issue--mind pointing me to a GitHub issue with your bug, or demonstrating an example of what this dependency upgrade fixes? After that I can promptly review your PR :)

grdryn:
@gabbifish Hi, thanks for taking a look.

I don't have a GitHub repo yet where the issue can be seen (it's just an operator I'm working with in my spare time so far, will have a repo soon, hopefully :slightly_smiling_face: ), but here's the change from flect that fixes it: https://github.com/gobuffalo/flect/pull/46

Essentially, without this update, if you run the following (kind is the important param):
```
kubebuilder create api --group webapp --version v1 --kind Synapse
```

Then in some of the generated code, and in some filenames, kubebuilder will put the pluralized version of the kind name, and it uses the flect library to decide what that should be. Flect has a list of general rules, and then a lot of exceptional cases because, well, English has a lot of exceptions. In this particular case, flect was using the general rule "if the singular ends with `pse`, then the plural changes to `psis`", from here: https://github.com/gobuffalo/flect/blob/64ac02ec8aaa79cec10c45f736c7c8127a2810a4/plural_rules.go#L228

So while that might be the correct way to pluralize words ending with `pse` in general, "synapse" is an exception: "synapsis" is not a correct pluralization. The norm in the flect library seems to be to just create a PR when you come across an exception that isn't covered, so that's what I did. :slightly_smiling_face:

gabbifish:
@grdryn Thank you for your thorough explanation! This totally makes sense (and wow at the level of detail flect goes into!) Running tests on this and if all tests pass (they should) I'll approve :) 

/ok-to-test

grdryn:
Thanks a lot @gabbifish and @estroz! :raised_hands:

PR #1699, opened 2020-09-29
🐛 Remove unneeded caBundle value from webhook patches (V3 only)

### LINKED ISSUE #1684: reapplying CRD yaml with conversion webhook enabled causes unexpected CRD resourceVersion update
## What broke
`kubectl apply -f` of a Kubebuilder generated CRD with conversion webhook enabled causes `resourceVersion` of CRD to be updated (which triggers watches to expire, which triggers re-list + reconciliation of all CRs of that CRD type).

## What did I expect to happen?
`resourceVersion` shouldn't be changed.

The reason `resourceVersion` was changed is because of [this section](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/v3/scaffolds/internal/templates/config/crd/enablewebhook_patch.go#L55) of the enablewebhook_patch template.

What ends up happening is:

1. Kustomize generates a CRD which includes the following snippet:
```
webhookClientConfig:
      # this is "\n" used as a placeholder, otherwise it will be rejected by the apiserver for being blank,
      caBundle: Cg==
```
2. That CRD is initially applied (so `caBundle` is `Cg==`).
3. `cert-manager` sees the `cert-manager.io/inject-ca-from` annotation and updates `caBundle` to the actual ca value.
4. `kubectl apply -f` of the CRD happens again some time later (for example due to a GitOps tool like [flux](https://github.com/fluxcd/flux), or just because somebody is reapplying CRDs for whatever reason).
5. The `caBundle` is reset to `Cg==` (causing a `resourceVersion` change).
6. `cert-manager` reapplys the `caBundle` (causing another `resourceVersion` change).

## Suggested solution
If instead of having a default of `caBundle: Cg==`, Kubebuilder generated patches that omitted `caBundle` entirely, this workflow would work much better, as per [the documented behavior of apply and how it considers last-applied-configuration](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#merging-changes-to-map-fields).

The comment:
>       # this is "\n" used as a placeholder, otherwise it will be rejected by the apiserver for being blank,
>       # but we're going to set it later using the cert-manager (or potentially a patch if not using cert-manager)
doesn't seem correct anyway since as far as I can tell this field has [been optional since Kubernetes 1.13](https://github.com/kubernetes/kubernetes/issues/69590)

I've tested this manually and confirmed that if I remove `caBundle: Cg==` from the yaml and reapply it on the CRD, it is accepted and subsequent reapplies don't change the `resourceVersion`.

* go version: 1.13
* kubebuilder version: 2.3.1
* Kubernetes version: 1.16.13

/kind bug

matthchr:
Instead of having a dummy value in `caBundle`, (`caBundle: Cg==`), instead just omit that field entirely as it is not required.
Doing this means that repeated `kubectl apply` operators on a CRD work better, in accordance with [the documented behavior of apply and how it considers last-applied-configuration](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#merging-changes-to-map-fields), since they do not change the `resourceVersion` of the CRD.

This fixes #1684.

### DIFF
--- a/docs/book/src/multiversion-tutorial/testdata/project/config/crd/patches/webhook_in_cronjobs.yaml
+++ b/docs/book/src/multiversion-tutorial/testdata/project/config/crd/patches/webhook_in_cronjobs.yaml
@@ -8,9 +8,6 @@ spec:
   conversion:
     strategy: Webhook
     webhookClientConfig:
-      # this is "\n" used as a placeholder, otherwise it will be rejected by the apiserver for being blank,
-      # but we're going to set it later using the cert-manager (or potentially a patch if not using cert-manager)
-      caBundle: Cg==
       service:
         namespace: system
         name: webhook-service
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/crd/patches/webhook_in_cronjobs.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/crd/patches/webhook_in_cronjobs.yaml
@@ -8,9 +8,6 @@ spec:
   conversion:
     strategy: Webhook
     webhookClientConfig:
-      # this is "\n" used as a placeholder, otherwise it will be rejected by the apiserver for being blank,
-      # but we're going to set it later using the cert-manager (or potentially a patch if not using cert-manager)
-      caBundle: Cg==
       service:
         namespace: system
         name: webhook-service
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/webhook/manifests.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/webhook/manifests.yaml
@@ -7,7 +7,6 @@ metadata:
   name: mutating-webhook-configuration
 webhooks:
 - clientConfig:
-    caBundle: Cg==
     service:
       name: webhook-service
       namespace: system
@@ -33,7 +32,6 @@ metadata:
   name: validating-webhook-configuration
 webhooks:
 - clientConfig:
-    caBundle: Cg==
     service:
       name: webhook-service
       namespace: system
--- a/pkg/plugin/v3/scaffolds/internal/templates/config/crd/enablewebhook_patch.go
+++ b/pkg/plugin/v3/scaffolds/internal/templates/config/crd/enablewebhook_patch.go
@@ -52,9 +52,6 @@ spec:
   conversion:
     strategy: Webhook
     webhookClientConfig:
-      # this is "\n" used as a placeholder, otherwise it will be rejected by the apiserver for being blank,
-      # but we're going to set it later using the cert-manager (or potentially a patch if not using cert-manager)
-      caBundle: Cg==
       service:
         namespace: system
         name: webhook-service
--- a/testdata/project-v3-addon/config/crd/patches/webhook_in_admirals.yaml
+++ b/testdata/project-v3-addon/config/crd/patches/webhook_in_admirals.yaml
@@ -8,9 +8,6 @@ spec:
   conversion:
     strategy: Webhook
     webhookClientConfig:
-      # this is "\n" used as a placeholder, otherwise it will be rejected by the apiserver for being blank,
-      # but we're going to set it later using the cert-manager (or potentially a patch if not using cert-manager)
-      caBundle: Cg==
       service:
         namespace: system
         name: webhook-service
--- a/testdata/project-v3-addon/config/crd/patches/webhook_in_captains.yaml
+++ b/testdata/project-v3-addon/config/crd/patches/webhook_in_captains.yaml
@@ -8,9 +8,6 @@ spec:
   conversion:
     strategy: Webhook
     webhookClientConfig:
-      # this is "\n" used as a placeholder, otherwise it will be rejected by the apiserver for being blank,
-      # but we're going to set it later using the cert-manager (or potentially a patch if not using cert-manager)
-      caBundle: Cg==
       service:
         namespace: system
         name: webhook-service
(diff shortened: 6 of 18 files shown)


matthchr:
Some of the contents of `testdata` folder seems to not have been updated even after I ran `make generate`. Specifically the following files: 
`testdata/project-v2-multigroup/config/webhook/manifests.yaml`
`testdata/project-v2/config/webhook/manifests.yaml`
`testdata/project-v3-multigroup/config/webhook/manifests.yaml`
`testdata/project-v3/config/webhook/manifests.yaml`. 

These are now the only places that have `cabundle: Cg==` and it feels like they should've been updated after I ran `make generate`.

Do I need to go remove that bit manually, or is it generated by a different command?

matthchr:
@DirectXMan12 can you weigh in when you get a chance on the question I asked above about the files that still have `Cg==` in them?

docs/book/src/multiversion-tutorial/testdata/project/config/crd/patches/webhook_in_cronjobs.yaml
```diff
@@ -8,9 +8,6 @@ spec:
   conversion:
     strategy: Webhook
     webhookClientConfig:
-      # this is "\n" used as a placeholder, otherwise it will be rejected by the apiserver for being blank,
-      # but we're going to set it later using the cert-manager (or potentially a patch if not using cert-manager)
-      caBundle: Cg==
```
camilamacedo86, line 13: The change cannot be done for v2. Only for v3+. It would be a breaking change.

testdata/project-v2-addon/config/crd/patches/webhook_in_admirals.yaml
```diff
@@ -8,9 +8,6 @@ spec:
   conversion:
     strategy: Webhook
     webhookClientConfig:
-      # this is "\n" used as a placeholder, otherwise it will be rejected by the apiserver for being blank,
-      # but we're going to set it later using the cert-manager (or potentially a patch if not using cert-manager)
-      caBundle: Cg==
```
camilamacedo86, line 13: The change cannot be done for v2. Only for v3+. It would be a breaking change.

matthchr: Thanks for the reply @camilamacedo86.

Can you explain more why this is a breaking change for v2? The comment in question is wrong - this field (`caBundle`) is optional and as far as I can tell has been optional since at least 1.13 (possibly longer). The CRD will not be rejected if this field is omitted.

Is there some behavioral breakage we're concerned with here?

I'm not super familiar with the v2 vs v3 distinction in Kubebuilder so just trying to understand why one is ok and one isn't.

DirectXMan12:
hmm... @camilamacedo86 would know that better, but `make generate` should be generating everything.   If you delete those files do they get recreated?

matthchr:
Yeah, if I delete those files they _are_ in fact regenerated (and they have `Cg==` in them). I'm not sure where it's coming from though as the only place that has `Cg==` is those files.

```
~/go/src/sigs.k8s.io/kubebuilde $ grep -rEli "Cg==" *

testdata/project-v2/config/webhook/manifests.yaml
testdata/project-v2-multigroup/config/webhook/manifests.yaml
testdata/project-v3/config/webhook/manifests.yaml
testdata/project-v3-multigroup/config/webhook/manifests.yaml
```

I must be missing someplace - will look through how `make generate` is making those files.

matthchr:
I think it's being created by controller-tools [here](https://github.com/kubernetes-sigs/controller-tools/blob/9a9db70f732af8f2ed69184a6e849f25d8e5a05e/pkg/webhook/parser.go#L248)?

Looks like maybe I need to send a PR there as well? Not entirely sure how to coordinate between the two repos - maybe controller_tools needs to merge first?

DirectXMan12:
yeah, just file a PR on controller-tools.  You can either merge that first, or merge this then file another PR here to update the CT version in v3 & update the golden files.

matthchr:
I've sent a PR for controller-tools too.

testdata/project-v2-addon/config/crd/patches/webhook_in_admirals.yaml
camilamacedo86: Hi @matthchr, 

Just bugs we are addressing to v2+ ( pkg/plugin/v2/) which means that this should be done only in ` pkg/plugin/v3/`. Could you please revert the change in the files inside of ` pkg/plugin/v2/`?

camilamacedo86:
Hi @matthchr, 

The `make generate` will remove the testdata/ directory and re-generate all projects. We ought to run that when we change anything in the templates.

camilamacedo86:
/ok-to-test
/approve 

It shows ok for me 👍 
really tks for your contribution. we need more +1 OK to get it merged.

matthchr:
@camilamacedo86 - it looks like a test timed out in the v2 tests... I don't think I changed anything there though (it was all reverted).

Is this just a flaky test?

testdata/project-v2-addon/config/crd/patches/webhook_in_admirals.yaml
matthchr: Did this

matthchr:
@camilamacedo86 @DirectXMan12 - pinging on this, would be good to get it merged now that the controller-tools PR has merged.

camilamacedo86 (changes requested):
.

camilamacedo86 (approved):
/approve

It shows ok for me. 👍

camilamacedo86:
I understand that we can 

/lgtm 

this one. 
It was done in the controller-tools already too. Also, it was checked and drove by @DirectXMan12 as well.

PR #1709, opened 2020-10-05
:seedling: the v2 webooks should use the v2 resource validation (:ghost:)

camilamacedo86:
Description

Follow up/fix of the PR:  https://github.com/kubernetes-sigs/kubebuilder/pull/1679/files#diff-5dd6baa6201e0ce7342b1eafbbb472e0R87-R89
Webhooks scaffolded with v2+ should use the V2 resource validation


gabbifish (approved):
/approve

Great catch!

PR #1710, opened 2020-10-06
:sparkles: Add scaffolding for the webhook test suite (go/v3-alpha)

### LINKED ISSUE #1707: Scaffolding for webhook tests
controller-runtime/pkg/envtest has [built-in support for webhook test](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest#WebhookInstallOptions).
It would be great if kubebuilder can generate a test runner under `api/` directory because it already can scaffold webhooks.

I have a good example of such a test runner.
https://github.com/cybozu-go/coil/blob/master/v2/api/v2/suite_test.go

/kind feature

prafull01:
Description
Added the webhook suite generation while creating the webhooks using the envtest. 

Motivation:
The webhook envtest suite setup requires setting up quite a things which would be useful for users to be programmatically generated as the envtest has supported webhook testing.

Fixes #1707

### DIFF
--- a/pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
+++ b/pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
@@ -1,0 +1,225 @@
+package api
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"sigs.k8s.io/kubebuilder/pkg/model/file"
+)
+
+var _ file.Template = &WebhookSuite{}
+var _ file.Inserter = &WebhookSuite{}
+
+// WebhookSuite scaffolds the webhook_suite.go file to setup the webhook test
+type WebhookSuite struct {
+	file.TemplateMixin
+	file.MultiGroupMixin
+	file.BoilerplateMixin
+	file.ResourceMixin
+
+	// BaseDirectoryRelativePath define the Path for the base directory when it is multigroup
+	BaseDirectoryRelativePath string
+}
+
+// SetTemplateDefaults implements file.Template
+func (f *WebhookSuite) SetTemplateDefaults() error {
+	if f.Path == "" {
+		if f.MultiGroup {
+			if f.Resource.Group != "" {
+				f.Path = filepath.Join("apis", "%[group]", "%[version]", "webhook_suite_test.go")
+			} else {
+				f.Path = filepath.Join("apis", "%[version]", "webhook_suite_test.go")
+			}
+		} else {
+			f.Path = filepath.Join("api", "%[version]", "webhook_suite_test.go")
+		}
+	}
+	f.Path = f.Resource.Replacer().Replace(f.Path)
+
+	f.TemplateBody = fmt.Sprintf(webhookTestSuiteTemplate,
+		file.NewMarkerFor(f.Path, importMarker),
+		file.NewMarkerFor(f.Path, addSchemeMarker),
+		file.NewMarkerFor(f.Path, addWebhookManagerMarker),
+		"%s",
+		"%d",
+	)
+
+	// If is multigroup the path needs to be ../../.. since it has the group dir.
+	f.BaseDirectoryRelativePath = `"..", ".."`
+	if f.MultiGroup && f.Resource.Group != "" {
+		f.BaseDirectoryRelativePath = `"..", "..",".."`
+	}
+
+	return nil
+}
+
+const (
+	// TODO: admission webhook versions should be based on the input of the user. For More Info #1664
+	admissionImportAlias    = "admissionv1beta1"
+	admissionPath           = "k8s.io/api/admission/v1beta1"
+	importMarker            = "imports"
+	addWebhookManagerMarker = "webhook"
+	addSchemeMarker         = "scheme"
+)
+
+// GetMarkers implements file.Inserter
+func (f *WebhookSuite) GetMarkers() []file.Marker {
+	return []file.Marker{
+		file.NewMarkerFor(f.Path, importMarker),
+		file.NewMarkerFor(f.Path, addSchemeMarker),
+		file.NewMarkerFor(f.Path, addWebhookManagerMarker),
+	}
+}
+
+const (
+	apiImportCodeFragment = `%s "%s"
+`
+	addschemeCodeFragment = `err = %s.AddToScheme(scheme	)
+Expect(err).NotTo(HaveOccurred())
+
+`
+	addWebhookManagerCodeFragment = `err = (&%s{}).SetupWebhookWithManager(mgr)
+Expect(err).NotTo(HaveOccurred())
+
+`
+)
+
+// GetCodeFragments implements file.Inserter
+func (f *WebhookSuite) GetCodeFragments() file.CodeFragmentsMap {
+	fragments := make(file.CodeFragmentsMap, 3)
+
+	// Generate import code fragments
+	imports := make([]string, 0)
+	imports = append(imports, fmt.Sprintf(apiImportCodeFragment, admissionImportAlias, admissionPath))
+
+	// Generate add scheme code fragments
+	addScheme := make([]string, 0)
+	addScheme = append(addScheme, fmt.Sprintf(addschemeCodeFragment, admissionImportAlias))
+
+	// Generate add webhookManager code fragments
+	addWebhookManager := make([]string, 0)
+	addWebhookManager = append(addWebhookManager, fmt.Sprintf(addWebhookManagerCodeFragment, f.Resource.Kind))
+
+	// Only store code fragments in the map if the slices are non-empty
+	if len(addWebhookManager) != 0 {
+		fragments[file.NewMarkerFor(f.Path, addWebhookManagerMarker)] = addWebhookManager
+	}
+	if len(imports) != 0 {
+		fragments[file.NewMarkerFor(f.Path, importMarker)] = imports
+	}
+	if len(addScheme) != 0 {
+		fragments[file.NewMarkerFor(f.Path, addSchemeMarker)] = addScheme
+	}
+
+	return fragments
+}
+
+const (
+	webhookTestSuiteTemplate = `
+package {{ .Resource.Version }}
+
+import (
+	"path/filepath"
+	"testing"
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	%s
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+// These tests use Ginkgo (BDD-style Go testing framework). Refer to
+// http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
+
+var cfg *rest.Config
+var k8sClient client.Client
+var testEnv *envtest.Environment
+var stopCh = make(chan struct{})
+
+func TestAPIs(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecsWithDefaultAndCustomReporters(t,
+	"Webhook Suite",
+	[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = BeforeSuite(func(done Done) {
+	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
+
+	By("bootstrapping test environment")
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths: []string{filepath.Join({{ .BaseDirectoryRelativePath }}, "config", "crd", "bases")},
+		WebhookInstallOptions: envtest.WebhookInstallOptions{
+			DirectoryPaths: []string{filepath.Join({{ .BaseDirectoryRelativePath }}, "config", "webhook")},
+		},
+	}
+
+	var err error
+	cfg, err = testEnv.Start()
+	Expect(err).NotTo(HaveOccurred())
+	Expect(cfg).NotTo(BeNil())
+
+	scheme := runtime.NewScheme()
+	err = AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	%s
+	
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme})
+	Expect(err).NotTo(HaveOccurred())
+	Expect(k8sClient).NotTo(BeNil())
+
+	// start webhook server using Manager
+	webhookInstallOptions := &testEnv.WebhookInstallOptions
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+		Scheme:             scheme,
+		Host:               webhookInstallOptions.LocalServingHost,
+		Port:               webhookInstallOptions.LocalServingPort,
+		CertDir:            webhookInstallOptions.LocalServingCertDir,
+		LeaderElection:     false,
+		MetricsBindAddress: "0",
+	})
+	Expect(err).NotTo(HaveOccurred())
+
+	%s
+
+	go func() {
+		err = mgr.Start(stopCh)
+		if err != nil {
+			Expect(err).NotTo(HaveOccurred())
+		}
+	}()
+
+	// wait for the webhook server to get ready
+	dialer := &net.Dialer{Timeout: time.Second}
+	addrPort := fmt.Sprintf("%s:%s", webhookInstallOptions.LocalServingHost, webhookInstallOptions.LocalServingPort)
+	Eventually(func() error {
+		conn, err := tls.DialWithDialer(dialer, "tcp", addrPort, &tls.Config{InsecureSkipVerify: true})
+		if err != nil {
+			return err
+		}
+		conn.Close()
+		return nil
+	}).Should(Succeed())
+
+	close(done)
+}, 60)
+
+var _ = AfterSuite(func() {
+	close(stopCh)
+	By("tearing down the test environment")
+	err := testEnv.Stop()
+	Expect(err).NotTo(HaveOccurred())
+})
+`
+)
--- a/pkg/plugin/v3/scaffolds/webhook.go
+++ b/pkg/plugin/v3/scaffolds/webhook.go
@@ -88,5 +88,15 @@ You need to implement the conversion.Hub and conversion.Convertible interfaces f
 		return err
 	}
 
+	// TODO: Add test suite for conversion webhook after #1664 has been merged & conversion tests supported in envtest.
+	if s.defaulting || s.validation {
+		if err := machinery.NewScaffold().Execute(
+			s.newUniverse(),
+			&api.WebhookSuite{},
+		); err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
--- a/testdata/project-v3-multigroup/apis/crew/v1/webhook_suite_test.go
+++ b/testdata/project-v3-multigroup/apis/crew/v1/webhook_suite_test.go
@@ -1,0 +1,115 @@
+package v1
+
+import (
+	"crypto/tls"
+	"fmt"
+	"net"
+	"path/filepath"
+	"testing"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+	// +kubebuilder:scaffold:imports
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+// These tests use Ginkgo (BDD-style Go testing framework). Refer to
+// http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
+
+var cfg *rest.Config
+var k8sClient client.Client
+var testEnv *envtest.Environment
+var stopCh = make(chan struct{})
+
+func TestAPIs(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecsWithDefaultAndCustomReporters(t,
+		"Webhook Suite",
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = BeforeSuite(func(done Done) {
+	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
+
+	By("bootstrapping test environment")
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths: []string{filepath.Join("..", "..", "..", "config", "crd", "bases")},
+		WebhookInstallOptions: envtest.WebhookInstallOptions{
+			DirectoryPaths: []string{filepath.Join("..", "..", "..", "config", "webhook")},
+		},
+	}
+
+	var err error
+	cfg, err = testEnv.Start()
+	Expect(err).NotTo(HaveOccurred())
+	Expect(cfg).NotTo(BeNil())
+
+	scheme := runtime.NewScheme()
+	err = AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	err = admissionv1beta1.AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	// +kubebuilder:scaffold:scheme
+
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme})
+	Expect(err).NotTo(HaveOccurred())
+	Expect(k8sClient).NotTo(BeNil())
+
+	// start webhook server using Manager
+	webhookInstallOptions := &testEnv.WebhookInstallOptions
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+		Scheme:             scheme,
+		Host:               webhookInstallOptions.LocalServingHost,
+		Port:               webhookInstallOptions.LocalServingPort,
+		CertDir:            webhookInstallOptions.LocalServingCertDir,
+		LeaderElection:     false,
+		MetricsBindAddress: "0",
+	})
+	Expect(err).NotTo(HaveOccurred())
+
+	err = (&Captain{}).SetupWebhookWithManager(mgr)
+	Expect(err).NotTo(HaveOccurred())
+
+	// +kubebuilder:scaffold:webhook
+
+	go func() {
+		err = mgr.Start(stopCh)
+		if err != nil {
+			Expect(err).NotTo(HaveOccurred())
+		}
+	}()
+
+	// wait for the webhook server to get ready
+	dialer := &net.Dialer{Timeout: time.Second}
+	addrPort := fmt.Sprintf("%s:%d", webhookInstallOptions.LocalServingHost, webhookInstallOptions.LocalServingPort)
+	Eventually(func() error {
+		conn, err := tls.DialWithDialer(dialer, "tcp", addrPort, &tls.Config{InsecureSkipVerify: true})
+		if err != nil {
+			return err
+		}
+		conn.Close()
+		return nil
+	}).Should(Succeed())
+
+	close(done)
+}, 60)
+
+var _ = AfterSuite(func() {
+	close(stopCh)
+	By("tearing down the test environment")
+	err := testEnv.Stop()
+	Expect(err).NotTo(HaveOccurred())
+})
--- a/testdata/project-v3-multigroup/apis/ship/v1/webhook_suite_test.go
+++ b/testdata/project-v3-multigroup/apis/ship/v1/webhook_suite_test.go
@@ -1,0 +1,115 @@
+package v1
+
+import (
+	"crypto/tls"
+	"fmt"
+	"net"
+	"path/filepath"
+	"testing"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+	// +kubebuilder:scaffold:imports
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+// These tests use Ginkgo (BDD-style Go testing framework). Refer to
+// http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
+
+var cfg *rest.Config
+var k8sClient client.Client
+var testEnv *envtest.Environment
+var stopCh = make(chan struct{})
+
+func TestAPIs(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecsWithDefaultAndCustomReporters(t,
+		"Webhook Suite",
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = BeforeSuite(func(done Done) {
+	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
+
+	By("bootstrapping test environment")
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths: []string{filepath.Join("..", "..", "..", "config", "crd", "bases")},
+		WebhookInstallOptions: envtest.WebhookInstallOptions{
+			DirectoryPaths: []string{filepath.Join("..", "..", "..", "config", "webhook")},
+		},
+	}
+
+	var err error
+	cfg, err = testEnv.Start()
+	Expect(err).NotTo(HaveOccurred())
+	Expect(cfg).NotTo(BeNil())
+
+	scheme := runtime.NewScheme()
+	err = AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	err = admissionv1beta1.AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	// +kubebuilder:scaffold:scheme
+
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme})
+	Expect(err).NotTo(HaveOccurred())
+	Expect(k8sClient).NotTo(BeNil())
+
+	// start webhook server using Manager
+	webhookInstallOptions := &testEnv.WebhookInstallOptions
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+		Scheme:             scheme,
+		Host:               webhookInstallOptions.LocalServingHost,
+		Port:               webhookInstallOptions.LocalServingPort,
+		CertDir:            webhookInstallOptions.LocalServingCertDir,
+		LeaderElection:     false,
+		MetricsBindAddress: "0",
+	})
+	Expect(err).NotTo(HaveOccurred())
+
+	err = (&Destroyer{}).SetupWebhookWithManager(mgr)
+	Expect(err).NotTo(HaveOccurred())
+
+	// +kubebuilder:scaffold:webhook
+
+	go func() {
+		err = mgr.Start(stopCh)
+		if err != nil {
+			Expect(err).NotTo(HaveOccurred())
+		}
+	}()
+
+	// wait for the webhook server to get ready
+	dialer := &net.Dialer{Timeout: time.Second}
+	addrPort := fmt.Sprintf("%s:%d", webhookInstallOptions.LocalServingHost, webhookInstallOptions.LocalServingPort)
+	Eventually(func() error {
+		conn, err := tls.DialWithDialer(dialer, "tcp", addrPort, &tls.Config{InsecureSkipVerify: true})
+		if err != nil {
+			return err
+		}
+		conn.Close()
+		return nil
+	}).Should(Succeed())
+
+	close(done)
+}, 60)
+
+var _ = AfterSuite(func() {
+	close(stopCh)
+	By("tearing down the test environment")
+	err := testEnv.Stop()
+	Expect(err).NotTo(HaveOccurred())
+})
--- a/testdata/project-v3-multigroup/apis/ship/v2alpha1/webhook_suite_test.go
+++ b/testdata/project-v3-multigroup/apis/ship/v2alpha1/webhook_suite_test.go
@@ -1,0 +1,115 @@
+package v2alpha1
+
+import (
+	"crypto/tls"
+	"fmt"
+	"net"
+	"path/filepath"
+	"testing"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+	// +kubebuilder:scaffold:imports
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+// These tests use Ginkgo (BDD-style Go testing framework). Refer to
+// http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
+
+var cfg *rest.Config
+var k8sClient client.Client
+var testEnv *envtest.Environment
+var stopCh = make(chan struct{})
+
+func TestAPIs(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecsWithDefaultAndCustomReporters(t,
+		"Webhook Suite",
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = BeforeSuite(func(done Done) {
+	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
+
+	By("bootstrapping test environment")
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths: []string{filepath.Join("..", "..", "..", "config", "crd", "bases")},
+		WebhookInstallOptions: envtest.WebhookInstallOptions{
+			DirectoryPaths: []string{filepath.Join("..", "..", "..", "config", "webhook")},
+		},
+	}
+
+	var err error
+	cfg, err = testEnv.Start()
+	Expect(err).NotTo(HaveOccurred())
+	Expect(cfg).NotTo(BeNil())
+
+	scheme := runtime.NewScheme()
+	err = AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	err = admissionv1beta1.AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	// +kubebuilder:scaffold:scheme
+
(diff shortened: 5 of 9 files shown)


pkg/plugin/v3/api.go
```diff
@@ -43,6 +43,9 @@ import (
 // TODO: remove this when a better solution for using addons is implemented.
 const KbDeclarativePatternVersion = "v0.0.0-20200522144838-848d48e5b073"
 
+// AddonPlugin is the name of the plugin for addons.
+const AddonPlugin = "addon"
```
camilamacedo86, line 47: The Addon still not a plugin. We have an issue raised to change that. See: https://github.com/kubernetes-sigs/kubebuilder/issues/1543

prafull01: Okay will change the comment. Thanks :)

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
```diff
@@ -0,0 +1,194 @@
...
+
+const (
+	webhookTestSuiteTemplate = `
+package {{ .Resource.Version }}
+
+import (
+	"path/filepath"
+	"testing"
+	"fmt"
+	. "github.com/onsi/ginkgo"
```
camilamacedo86, line 95:
```suggestion

	. "github.com/onsi/ginkgo"
```

camilamacedo86: Just a nit to sort properly the imports.

pkg/plugin/v3/webhook.go
```diff
@@ -106,9 +113,21 @@ func (p *createWebhookPlugin) GetScaffolder() (scaffold.Scaffolder, error) {
...
+		plugins = append(plugins, &addon.Plugin{})
+	default:
+		return nil, fmt.Errorf("unknown pattern %q", p.pattern)
+	}
+
 	// Create the actual resource from the resource options
 	res := p.resource.NewResource(p.config, false)
-	return scaffolds.NewWebhookScaffolder(p.config, string(bp), res, p.defaulting, p.validation, p.conversion), nil
+	return scaffolds.NewWebhookScaffolder(p.config, string(bp), res, p.defaulting, p.validation, p.conversion, plugins),
+		nil
```
camilamacedo86, line 130: It did not change the testdata project-v3-addon. Do we really add the that for the addon?

prafull01: Addon is not generating the webhooks in the testdata.

camilamacedo86: WDYT about we add some mock for that so?

prafull01: Yes, we can add the webhook for the addons as well.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
prafull01: Done

pkg/plugin/v3/api.go
prafull01: Done

pkg/plugin/v3/scaffolds/webhook.go
```diff
@@ -78,13 +81,22 @@ func (s *webhookScaffolder) scaffold() error {
...
-	if err := machinery.NewScaffold().Execute(
+	if err := machinery.NewScaffold(s.plugins...).Execute(
 		s.newUniverse(),
 		&api.Webhook{Defaulting: s.defaulting, Validating: s.validation},
 		&templates.MainUpdater{WireWebhook: true},
 	); err != nil {
 		return err
 	}
 
+	if s.defaulting || s.validation {
```
camilamacedo86, line 92: why we scaffold the suite just in these both cases?

prafull01: The conversion scenario doesn't add anything in the webhook.go it creates. No default functions and markers around those are added for conversion. Hence it doesn't create any entry for conversion Kind into the validating and mutating configuration in the folder config/webhook.This the reason no need to add the suite changes for conversion webhook. You can check the conversion webhook here: `testdata/project-v3-multigroup/apis/ship/v1beta1/frigate_webhook.go`

prafull01: There is a pending PR #1664 around implementing conversion webhook properly. Even after that this suite should not go there. There is a pending PR in controller-runtime which adds support of conversion webhook in envtest. We can add the different suite for conversion webhook once this [PR is merged](https://github.com/kubernetes-sigs/controller-runtime/pull/999)

camilamacedo86: so, could you please add a todo comment for we do not forget it?
Also, could you ping this pr in the #1664 for the author knows that if it gets merged this if required to be changed as it will be described in the todo comment?

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
```diff
@@ -0,0 +1,195 @@
...
+import (
+	"path/filepath"
+	"testing"
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	admissionv1 "k8s.io/api/admission/v1"
```
camilamacedo86, line 100: The api version used to scaffold the webhooks still v1betav1. so, would not better to use https://pkg.go.dev/k8s.io/api@v0.19.2/admission/v1beta1. 

also, if we change that then, not in the scaffold but in the code would be nice have a comment with a todo for we parametrize the version according to the version used to scaffold the webhooks. More info see: https://github.com/kubernetes-sigs/kubebuilder/pull/1644#discussion_r506385099

WDYT?

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
```diff
@@ -0,0 +1,195 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+// These tests use Ginkgo (BDD-style Go testing framework). Refer to
+// http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
+
+var cfg *rest.Config
+var k8sClient client.Client
+var testEnv *envtest.Environment
+var stopCh = make(chan struct{})
```
camilamacedo86, line 116: we have not this option in the controller. wDYT about we do it as similar as possible to the controller one and then, we can address improvements in both in a follow up if it is the case? See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller_suitetest.go#L134-L136

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
```diff
@@ -0,0 +1,195 @@
+package api
```
camilamacedo86, line 1: Could we rename the file to `webhook_suitetest.go` to be similar to `kubebuilder/pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller_suitetest.go`

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
```diff
@@ -0,0 +1,195 @@
...
+	var err error
+	cfg, err = testEnv.Start()
+	Expect(err).ToNot(HaveOccurred())
+	Expect(cfg).ToNot(BeNil())
+
+	scheme := runtime.NewScheme()
+	err = AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
+	err = admissionv1.AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
```
camilamacedo86, line 146: by using the v1betav1, could we not remove that?

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
```diff
@@ -0,0 +1,195 @@
...
+	wio := &testEnv.WebhookInstallOptions
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+		Scheme:             scheme,
+		Host:               wio.LocalServingHost,
+		Port:               wio.LocalServingPort,
+		CertDir:            wio.LocalServingCertDir,
+		LeaderElection:     false,
+		MetricsBindAddress: "0",
+	})
+	Expect(err).ToNot(HaveOccurred())
```
camilamacedo86, line 162: why we need it? the same is not done in the controller_suitetest.go?

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
```diff
@@ -0,0 +1,195 @@
...
+	dialer := &net.Dialer{Timeout: time.Second}
+	addrPort := wio.LocalServingHost + ":" + fmt.Sprint(wio.LocalServingPort)
+	Eventually(func() error {
+		conn, err := tls.DialWithDialer(dialer, "tcp", addrPort, &tls.Config{InsecureSkipVerify: true})
+		if err != nil {
+			return err
+		}
+		conn.Close()
+		return nil
+	}).Should(Succeed())
```
camilamacedo86, line 183: same ^

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
```diff
@@ -0,0 +1,195 @@
...
+
+	close(done)
+}, 60)
+
+var _ = AfterSuite(func() {
+	close(stopCh)
```
camilamacedo86, line 189:
```suggestion
```
is it really required?

pkg/plugin/v3/webhook.go
camilamacedo86: I did not see a testdata update for add-on. Did you make that already?

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
prafull01: This is needed because of the following issue with controller runtime [here](https://github.com/kubernetes-sigs/controller-runtime/issues/1020)  . Webhook testing is racy when run in a simple way, sometimes webhook server doesn't comes up and manager gets started.

pkg/plugin/v3/webhook.go
prafull01: There is no webhook creation supported for addon right now. I need to make the kb create webhook command to accept --pattern addon.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
prafull01: Got it. So for now, I make it v1beta1 and add a TODO in the scaffold to change once we finalize the approach on how to handle it

camilamacedo86: I think would be nice add value to the SetTemplate maybe. WDYT?

prafull01: Got it. Yes it would be nice when we have to handle both the version. Will do the changes.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
prafull01: You need a webhook server running to run the webhook tests with envtest. To run a webhook server you have to create a manager with host, port and certDir and attach webhook with that manager. This was not needed in controller suite. We can leave this step in users hands to implement it but I though this is static enough to be part of scaffolding. WDYT?

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
prafull01: IMHO the `controller_suitetest.go` file name should be changed to `controller_suite_test.go`. The reason behind this the _test.go files are not imported when a package is imported. When I create a file `webhook_suitetest.go` and the package is imported in controllers, this test suite is also gets imported there and go test fails... 
```
?   	sigs.k8s.io/kubebuilder/testdata/project-v3	[no test files]
?   	sigs.k8s.io/kubebuilder/testdata/project-v3/api/v1	[no test files]
panic: You may only call BeforeSuite once!

goroutine 1 [running]:
github.com/onsi/ginkgo/internal/suite.(*Suite).SetBeforeSuiteNode(0xc0001046e0, 0x15be3e0, 0x18a0fe0, 0x22f1e0d, 0x63, 0x34, 0xc0003fa160, 0xac, 0xdf8475800)
	/home/prafull/go/pkg/mod/github.com/onsi/ginkgo@v1.12.1/internal/suite/suite.go:113 +0x170
github.com/onsi/ginkgo.BeforeSuite(0x15be3e0, 0x18a0fe0, 0xc00039c990, 0x1, 0x1, 0x14)
	/home/prafull/go/pkg/mod/github.com/onsi/ginkgo@v1.12.1/ginkgo_dsl.go:484 +0xab
sigs.k8s.io/kubebuilder/testdata/project-v3/controllers.init()
	/home/prafull/go/src/github.com/prafull01/kubebuilder/testdata/project-v3/controllers/suite_test.go:52 +0x75
FAIL	sigs.k8s.io/kubebuilder/testdata/project-v3/controllers	0.012s
FAIL
make: *** [Makefile:21: test] Error 1
```

I will raise another PR to change the `controller_suitetest.go` file name.

pkg/plugin/v3/scaffolds/webhook.go
prafull01: Done

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
prafull01: Done

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
prafull01: Done

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
prafull01: This will be needed if we decide to keep the starting webhook manager as a part of suite, then this is needed to stop the manager in AfterSuite. In controller, it is not needed because we are not setting up any manager.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
camilamacedo86: Are you adding the webhooks in the controller path? The suggestion was regarded just to rename the files (boilerplate and the file that is generated) but not the path. The webhook_suitetest.go ought to be where the webhooks are scaffolded. 

So it would be to:
- Replace `webhook_suite_test.go` for `webhook_suitetest.go`
- Repalce `pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go` for `pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go`

pkg/plugin/v3/webhook.go
camilamacedo86: Catcher, we need to make the addon --partner to be a plugin as well. And then, after that, we could work to make the addon scaffold also support webhook if possible. So, it is fine 👍 Really thank you for check that.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
camilamacedo86: I agree that scaffold it is better 👍

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go
prafull01: Yes.. I am scaffolding the `webhook_suitetest.go` in `api/v1/webhook_suitetest.go` only. However, controller suite is importing the v1 package like this `crewv1 "sigs.k8s.io/kubebuilder/testdata/project-v3/api/v1"` . This import is causing above issue mentioned in my previous comment. Any test suite we define should end `_test.go` so that the package imports doesn't imports those file and functions written in those file. Or we can have separate package like `v1_test` for the testing suites so that those are not imported when original package is imported.

1. Replace webhook_suite_test.go for webhook_suitetest.go : The file should end with `_test.go`, in controllers path the name of the file is simply `suite_test.go`. We can decide the naming convention here.

2. The name of `pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suite.go`, I will replace it as per your suggestion.

pkg/plugin/v3/webhook.go
```diff
@@ -106,9 +113,21 @@ func (p *createWebhookPlugin) GetScaffolder() (scaffold.Scaffolder, error) {
...
+	plugins := make([]model.Plugin, 0)
+	switch strings.ToLower(p.pattern) {
+	case "":
+		// Default pattern
+	case AddonPlugin:
+		plugins = append(plugins, &addon.Plugin{})
+	default:
+		return nil, fmt.Errorf("unknown pattern %q", p.pattern)
+	}
+
```
camilamacedo86, line 126: Why do we need to add this code?

prafull01: I missed the train of thoughts here. Added this to have a addon plugins supported for the webhook. Just same as the api.go has. Let me remove all the out of scope code from this PR. Thanks for pointing it out.

prafull01: Removed it.

camilamacedo86 (approved):
Hi @prafull01,

Really thank you for your contribution. It shows OK for me 👍 
/approve

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
```diff
@@ -0,0 +1,223 @@
...
+	Expect(err).NotTo(HaveOccurred())
+
+	%s
+	
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme})
+	Expect(err).ToNot(HaveOccurred())
+	Expect(k8sClient).ToNot(BeNil())
+
+	// start webhook server using Manager
+	wio := &testEnv.WebhookInstallOptions
```
ellistarn, line 181: I found myself confused by the wio's everywhere. consider dropping for readability

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
```diff
@@ -0,0 +1,223 @@
...
+		MetricsBindAddress: "0",
+	})
+	Expect(err).ToNot(HaveOccurred())
+
+	%s
+
+	go func() {
+		err = mgr.Start(stopCh)
+		if err != nil {
+			panic(err)
```
ellistarn, line 197: I found myself implementing a panicIfErr utility in my controller. It might be useful to have inside controller runtime.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+	}()
+
+	// wait for the webhook server to get ready
+	dialer := &net.Dialer{Timeout: time.Second}
+	addrPort := wio.LocalServingHost + ":" + fmt.Sprint(wio.LocalServingPort)
```
ellistarn, line 203: consider using fmt.Sprintf

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
```diff
@@ -0,0 +1,223 @@
...
+	Eventually(func() error {
+		conn, err := tls.DialWithDialer(dialer, "tcp", addrPort, &tls.Config{InsecureSkipVerify: true})
```
ellistarn, line 205: InsecureSkipVerify, aren't certificates generated in local mode? This seems like a bad default that encourages security vulns. 

I use
```
	e.Manager = controllers.NewManagerOrDie(e.Config, controllerruntime.Options{
		CertDir:            e.WebhookInstallOptions.LocalServingCertDir,
		Host:               e.WebhookInstallOptions.LocalServingHost,
		Port:               e.WebhookInstallOptions.LocalServingPort,
		MetricsBindAddress: "0", // Skip the metrics server to avoid port conflicts for parallel testing
	})
```
Which gives me certs.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
```diff
@@ -0,0 +1,223 @@
...
+		CRDDirectoryPaths: []string{filepath.Join({{ .BaseDirectoryRelativePath }}, "config", "crd", "bases")},
+		WebhookInstallOptions: envtest.WebhookInstallOptions{
+			DirectoryPaths: []string{filepath.Join({{ .BaseDirectoryRelativePath }}, "config", "webhook")},
+		},
+	}
+
+	var err error
+	cfg, err = testEnv.Start()
+	Expect(err).ToNot(HaveOccurred())
+	Expect(cfg).ToNot(BeNil())
```
ellistarn, line 168: Why would this be nil? Potentially unnecessary bloat.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
```diff
@@ -0,0 +1,223 @@
...
+
+	scheme := runtime.NewScheme()
```
ellistarn, line 170: I found it was really useful to use a pattern like this in my testing
```
var (
	scheme = runtime.NewScheme()
)

func init() {
	log.PanicIfError(clientgoscheme.AddToScheme(scheme), "adding clientgo to scheme")
	log.PanicIfError(apis.AddToScheme(scheme), "adding apis to scheme")
}
```

It basically creates a global singleton scheme. I use it in both my main and suite_test

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
```diff
@@ -0,0 +1,223 @@
...
+	err = AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	%s
+	
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme})
+	Expect(err).ToNot(HaveOccurred())
+	Expect(k8sClient).ToNot(BeNil())
```
ellistarn, line 178: likewise

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	var err error
+	cfg, err = testEnv.Start()
+	Expect(err).ToNot(HaveOccurred())
+	Expect(cfg).ToNot(BeNil())
+
+	scheme := runtime.NewScheme()
+	err = AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
```
ellistarn, line 172: For single return types, consider the more brief form:
```
Expect(MyErrorableFunction()).To(Succeed())

File: pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go [hunk 21]
```diff
@@ -0,0 +1,223 @@
...
+		return nil
+	}).Should(Succeed())
+
+	close(done)
+}, 60)
+
+var _ = AfterSuite(func() {
+	close(stopCh)
+	By("tearing down the test environment")
+	err := testEnv.Stop()
```
ellistarn, line 219: Likewise.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
```diff
@@ -0,0 +1,223 @@
...
+		conn.Close()
+		return nil
+	}).Should(Succeed())
+
+	close(done)
+}, 60)
+
+var _ = AfterSuite(func() {
+	close(stopCh)
+	By("tearing down the test environment")
```
ellistarn, line 218: What do these `By` statements buy us?

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
```diff
@@ -0,0 +1,223 @@
...
+
+func TestAPIs(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecsWithDefaultAndCustomReporters(t,
+	"Webhook Suite",
+	[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = BeforeSuite(func(done Done) {
```
ellistarn, line 154: I might use the form without done given that done isn't being used.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
```diff
@@ -0,0 +1,223 @@
...
+
+	By("bootstrapping test environment")
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths: []string{filepath.Join({{ .BaseDirectoryRelativePath }}, "config", "crd", "bases")},
+		WebhookInstallOptions: envtest.WebhookInstallOptions{
+			DirectoryPaths: []string{filepath.Join({{ .BaseDirectoryRelativePath }}, "config", "webhook")},
+		},
+	}
+
+	var err error
```
ellistarn, line 165: Rather that mutating the same error every time, it's probably more idiomatic to use the form
```
foo, err := Func()
```

ellistarn:
overall a very useful example to help users get started testing webhooks.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
camilamacedo86: + for this suggestion

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
camilamacedo86: +1 for this suggestion as well. Good catcher
```suggestion
```

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
```diff
@@ -0,0 +1,223 @@
...
+		err = mgr.Start(stopCh)
+		if err != nil {
+			panic(err)
+		}
+	}()
+
+	// wait for the webhook server to get ready
+	dialer := &net.Dialer{Timeout: time.Second}
+	addrPort := wio.LocalServingHost + ":" + fmt.Sprint(wio.LocalServingPort)
+	Eventually(func() error {
```
ellistarn, line 204: I found a slightly less verbose solution
```
	Eventually(func() error {
		_, err := net.DialTimeout("tcp", fmt.Sprintf("%s:%d",
			e.WebhookInstallOptions.LocalServingHost,
			e.WebhookInstallOptions.LocalServingPort,
		), time.Second)
		return err
	}).Should(Succeed())
```

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
camilamacedo86: I prefer it as `Expect(err).NotTo(HaveOccurred())` since it shows more intuitive and follows the standard applied in the project. See the suite_test for the controllers. Also, see that this discussion was already made in other places https://github.com/operator-framework/operator-sdk/issues/3896 and the final decision reach kept the `Expect(err).NotTo(HaveOccurred())` as it is now.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
camilamacedo86: the tests uses http://onsi.github.io/ginkgo/. The `By` output the test logs and it follows the project standard. see the controllers/suite_test.go: 

https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller_suitetest.go#L168-L172

IMO we ought to keep this implementation as closely as possible of the suite_test scaffolded for the controllers.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
camilamacedo86: May, installOptions? Wdyt?

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
camilamacedo86: The cfg is nill in the failure cases. See: https://github.com/kubernetes-sigs/controller-runtime/blob/v0.6.3/pkg/envtest/server.go#L192

It is such as we have for the controllers. See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller_suitetest.go#L154-L157

Theoretically, the error will not be nil so we would not need to test the cfg. However, it might have a scenario that is not properly handled so, in POV we should not change it and keep it as it is done in the suite_test for the controller.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
camilamacedo86: Why should we `log.PanicIfError` instead of following up the project standard and use the test lib adopted(ginkgo and gomega)? IMO we should not change that and stick with the lib funcs and the same implementation done for the suite-test.go in the controllers instead of it. 

I am not sure if it would be a good approach. Should we add always the same schemas used in the main.go for all suite_tests.go? Could not it be specific for the test/controller/webhook covered?

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
camilamacedo86: I think it was a good catcher. Could we not use here the lib test func here as well?
```suggestion
			Expect(err).NotTo(HaveOccurred())
```

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
camilamacedo86: I agree with that. However, the same is applied to https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller_suitetest.go#L146. So, I think it might better be addressed in a follow up for both scenarios.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
camilamacedo86: + 1 for that 🥇 
However, IMO it needs to be addressed in https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller_suitetest.go#L154 as well. So, I think it would be better addressed in a follow up for both.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
ellistarn: Oh -- feel free to ignore the panic if error bit. You can error handle however you like. 

More crisply, I found that settling up a global singleton scheme was a clean approach for my usecase rather than reinstantiating (and forgetting) it in all my tests.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
ellistarn: Sure. I think gomega was written to work both ways.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
ellistarn: I'm not picky. I mostly found this to be cruft when kb generated my code. It's removed in my controller. If you agree its cruft, fixing in a followup SGTM.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
ellistarn: For reference, many of my comments here came from my thoughts when I was "cleaning up" the kb generated code that's setup elsewhere. Of course, everyone is entitled to their own code style and opinions. I thought I'd raise the bits I found to be awkward. I agree with consistent style. If you agree that it's awkward, a followup to address all cases sounds good.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
ellistarn: Yep :)

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
camilamacedo86: It is fine. Really thank you for your collab here as well btw :-)

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
prafull01: Thank you for review. I tried to use the `fmt.Sprintf` in my code at the first attempt, but the scaffolding doesn't really add the %s in the scaffolded code. I initially tried to add the following line:
`addrPort := fmt.Sprintf("%s:%d", wio.LocalServingHost, wio.LocalServingPort)` but these %s in the scaffolded code tried to prints the following line after scaffolding:
`addrPort := fmt.Sprintf("%!s(MISSING):%!d(MISSING)", wio.LocalServingHost, wio.LocalServingPort)`. 

I tried to find the solution so that multiline string prints %s and %d as it is, but could not find that easily. Let me check again..

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
prafull01: IMO, `webhookInstallOptions` would be the right term here, there can be other installOptions like CRDInstallOptions in envtest.

camilamacedo86: it shows nice. Let's change for  `webhookInstallOptions ` to get it merged so.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
prafull01: @camilamacedo86 Do you know how to solve the above challenge I am facing?

camilamacedo86: Did you try `%v`?

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
camilamacedo86: IHMO we should keep because it is following the test lib adopted and make clear what operation is done at this point.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
```diff
@@ -0,0 +1,223 @@
...
+		} else {
+			f.Path = filepath.Join("api", "%[version]", "webhook_suite_test.go")
+		}
+	}
+	f.Path = f.Resource.Replacer().Replace(f.Path)
+
+	f.TemplateBody = fmt.Sprintf(webhookTestSuiteTemplate,
+		file.NewMarkerFor(f.Path, importMarker),
+		file.NewMarkerFor(f.Path, addSchemeMarker),
+		file.NewMarkerFor(f.Path, addWebhookManagerMarker),
```
prafull01, line 42: @camilamacedo86 To make it work, I have to add two strings "%s", "%d" or "%v" here, so that those get replaced at the time of scaffolding... Should I do that? WDYT?

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
camilamacedo86: Closing it as solved tks.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
camilamacedo86: Closing it as solved.

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
prafull01: Same error. We have to replace template body like this:
```
f.TemplateBody = fmt.Sprintf(webhookTestSuiteTemplate,
		file.NewMarkerFor(f.Path, importMarker),
		file.NewMarkerFor(f.Path, addSchemeMarker),
		file.NewMarkerFor(f.Path, addWebhookManagerMarker),
		"%v",
		"%v",
	)
```
So that, these %v can be replaced in the multi line string we have `webhookTestSuiteTemplate`

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
prafull01: Done

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
prafull01: Done

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
prafull01: Done

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
```diff
@@ -0,0 +1,225 @@
...
+		err = mgr.Start(stopCh)
+		if err != nil {
+			Expect(err).NotTo(HaveOccurred())
+		}
+	}()
+
+	// wait for the webhook server to get ready
+	dialer := &net.Dialer{Timeout: time.Second}
+	addrPort := fmt.Sprintf("%s:%s", webhookInstallOptions.LocalServingHost, webhookInstallOptions.LocalServingPort)
+	Eventually(func() error {
```
ellistarn, line 206: Suggestion:
Consider relying on https://godoc.org/k8s.io/apimachinery/pkg/util/wait#ExponentialBackoff. In my controller, i extracted all this setup logic into a more fully featured "envtest+manager" abstraction. It would be nice to have this logic less reliant on gomega and instead rely on apimachinery. If KB were to ever make an abstraction like this (to avoid this code in every suite_test), you wouldn't want that code to rely on gomega.

camilamacedo86: Really thank you for your input. I think that the best way for it be addressed would be might via a follow-up PR. So, WDYT about after it gets merged you push a PR with your suggestion? Then, it might clarify better as well as your thoughts.

pkg/plugin/v3/webhook.go
```diff
@@ -114,7 +114,8 @@ func (p *createWebhookPlugin) GetScaffolder() (scaffold.Scaffolder, error) {
 
 	// Create the actual resource from the resource options
 	res := p.resource.NewResource(p.config, false)
-	return scaffolds.NewWebhookScaffolder(p.config, string(bp), res, p.defaulting, p.validation, p.conversion), nil
+	return scaffolds.NewWebhookScaffolder(p.config, string(bp), res, p.defaulting, p.validation, p.conversion),
+		nil
```
Adirio, line 118: What does this change have to do with this PR?

pkg/plugin/v3/api.go
```diff
@@ -43,6 +43,10 @@ import (
 // TODO: remove this when a better solution for using addons is implemented.
 const KbDeclarativePatternVersion = "v0.0.0-20200522144838-848d48e5b073"
 
+// AddonPlugin is the name of the plugin for addons.
+// TODO: the addon should be a plugin. More info: #1543
+const AddonPlugin = "addon"
+
```
Adirio, line 49: What does this change have to do with this PR?

pkg/plugin/v3/scaffolds/webhook.go
```diff
@@ -34,7 +34,6 @@ type webhookScaffolder struct {
 	config      *config.Config
 	boilerplate string
 	resource    *resource.Resource
-
```
Adirio, line 37: Why remove this blank line?

pkg/plugin/v3/scaffolds/webhook.go
```diff
@@ -86,5 +85,15 @@ You need to implement the conversion.Hub and conversion.Convertible interfaces f
 		return err
 	}
 
+	// TODO: Add test suite for conversion webhook after #1664 has been merged & conversion tests supported in envtest.
```
Adirio, line 88: Why is this a TODO? Aren't you just doing this?

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
Adirio: You are not mutating an error, you are re-assigning it. The current approach only allocates memory once while the other needs to allocate memory for every function so I don't think it is that good of an idea.

Adirio: You are not mutating an error, you are re-assigning it. The current approach only allocates memory once while the other needs to allocate memory for every function so I don't think it is that good of an idea.

pkg/plugin/v3/scaffolds/webhook.go
camilamacedo86: This todo is because of #1664. 
Just to make clear that it should be updated after that as well.

pkg/plugin/v3/api.go
camilamacedo86: I agree with @Adirio that this cleanup would be better addressed in another PR.
Could you please revert these changes and if you wish to create a pr just with the const and the comment?

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
camilamacedo86: HI @Adirio, 

The community suggestion was : 
Replace
```
var err error
cfg, err = testEnv.Start()
```

With 
  
```
cfg, err := testEnv.Start()
```

In order to keep the simplicity in the code which shows totally fine for me. 
However, the most important in my POV is to keep both suite tests for controller and webhook as closer as possible, so, in this way a new issue was tracked to address this nit in both : https://github.com/kubernetes-sigs/kubebuilder/issues/1733

So, if you do not agree with the above change could you please add your inputs to #1733?

pkg/plugin/v3/api.go
prafull01: Without this change, the golint fails that the "addon" has been used 3 time, make it a constant. Hence, I have added this change here.

camilamacedo86: Hi @prafull01, 

I think it was falling before the reviews and the changes made since it does not fail in the master/other prs.  
I think we can revert the changes made in this file. 
Also, see that the `AddonPlugin` is not used in another place so should not be exported.

pkg/plugin/v3/webhook.go
prafull01: Accidental enter, corrected it

pkg/plugin/v3/scaffolds/webhook.go
prafull01: My bad, corrected it

pkg/plugin/v3/api.go
prafull01: Yeah, reverted it. @Adirio Thanks for your review.

camilamacedo86:
After spoke with @estroz, it might be better we use `(go/v3-alpha)` instead of `(v3+ only)`. 
Could you please update the title/commit

Adirio (approved):
/lgtm

camilamacedo86:
It was rebased with the master because of this loses the @Adirio lgtm
Doing that again 👍 

/lgtm

pkg/plugin/v3/scaffolds/internal/templates/config/api/webhook_suitetest.go
ndipebot: +1 for not mutating the same error every time.

Adirio: The proposed changed is more idiomatic but doesn't change the behavior you mentioned. You are still using the same `err` variable and assigning different values to it. You are not actually mutating an error in any of the two cases. You just declare the variable the first time you use it instead of explicitly declaring the variable. So +1 for the change, but the explanation lead me to thinking the change you suggested was different.

PR #1713, opened 2020-10-07
:book: update versioning doc

camilamacedo86:
**Description**
Update the versioning doc with the final decision over plugins version

**Motivation**
When the concept of the plugin was introduced the first thought was work with X.Y.Z versions for the plugins. However, after that, we decided to not have a minor version.


camilamacedo86:
c/c @gabbifish @estroz

camilamacedo86:
Ignore. We have an open PR for that.

PR #1715, opened 2020-10-09
:sparkles: upgrade controller-runtime from v0.6.2 to v0.6.3

camilamacedo86:
**Description**

- Upgrade controller-runtime from v0.6.2 to v0.6.3
- Upgrade the makefile target test to use the v0.6.3 tag instead master to get the script since now it is released in the tag.


gabbifish (approved):
/lgtm

tested `make test` across v2 and v3 `testdata/` directories and all looks well!

PR #1716, opened 2020-10-09
 :warning: Update kubebuilder cli and plugin v3+ to use go version 1.15

rashmigottipati:
**Description**

- Update go version from 1.13 to 1.15 

**Note**
Here's the link to update the infra to run the CI jobs as well: kubernetes/test-infra#19527

### DIFF
--- a/pkg/plugin/internal/util/go_version.go
+++ b/pkg/plugin/internal/util/go_version.go
@@ -51,6 +51,8 @@ func fetchAndCheckGoVersion() error {
 	return nil
 }
 
+// checkGoVersion should only ever check if the Go version >= 1.13, since the kubebuilder binary only cares
+// that the go binary supports go modules which were stabilized in that version (i.e. in go 1.13) by default
 func checkGoVersion(verStr string) error {
 	goVerRegex := `^go?([0-9]+)\.([0-9]+)([\.0-9A-Za-z\-]+)?$`
 	m := regexp.MustCompile(goVerRegex).FindStringSubmatch(verStr)
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ os:
   - osx
 
 go:
-  - 1.13.x
+  - 1.15.x
 
 git:
   depth: 3
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,7 +10,7 @@ Please see https://git.k8s.io/community/CLA.md for more info.
 
 ## Prerequisites
 
-- [go](https://golang.org/dl/) version v1.13+.
+- [go](https://golang.org/dl/) version v1.15+.
 - [docker](https://docs.docker.com/install/) version 17.03+.
 - [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) version v1.11.3+.
 - [kustomize](https://sigs.k8s.io/kustomize/docs/INSTALL.md) v3.1.0+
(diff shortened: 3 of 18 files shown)


camilamacedo86:
Hi @rashmigottipati, 

You need to follow the instructions in https://github.com/kubernetes-sigs/kubebuilder/pull/1716#issuecomment-705920674
/ok-to-test

camilamacedo86:
Hi @DirectXMan12, 

Controller-runtime has updated o use 1.15 already. Have you any objection to this change? Just to confirm. 
For me it is OK. 

/approved

rashmigottipati:
I signed it

rashmigottipati:
@camilamacedo86 thanks for your input Camila. I've created this PR in kubernetes/test-infra updating the go version in https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/kubebuilder/kubebuilder-presubmits.yaml#L13 to 1.15. 
Here's the link: https://github.com/kubernetes/test-infra/pull/19527

pkg/plugin/internal/util/go_version.go
```diff
@@ -69,7 +69,7 @@ func checkGoVersion(verStr string) error {
 	}
 
 	if major < 1 || minor < 13 {
```
estroz, line 71:
```suggestion
	if major < 1 || minor < 15 {
```

pkg/plugin/v2/scaffolds/internal/templates/dockerfile.go
```diff
@@ -39,7 +39,7 @@ func (f *Dockerfile) SetTemplateDefaults() error {
 }
 
 const dockerfileTemplate = `# Build the manager binary
-FROM golang:1.13 as builder
```
estroz, line 42: We shouldn't be changing v2's Go version unless we account for plugin version in `checkGoVersion()`.

estroz (changes requested):
Bumping the v2 plugin to go 1.15 is a breaking change because the binary will fail for previously scaffolded projects (in `checkGoVersion()`). I recommend only bumping in v3.

pkg/plugin/internal/util/go_version.go
```diff
@@ -69,7 +69,7 @@ func checkGoVersion(verStr string) error {
 	}
```
estroz, line 69: Go version checks should be refactored into versioned plugin packages such that the binary can accept multiple go versions.

camilamacedo86: good catcher.

camilamacedo86:
Hi @estroz, 

> Bumping the v2 plugin to go 1.15 is a breaking change because the binary will fail for previously scaffolded projects (in checkGoVersion()). I recommend only bumping in v3.

I agree that is a breaking change. However, in this case, I think it is fine and we need to move forward with both otherwise it will be hard to keep maintained and used. See that: 

- It is very strange to ask for the user use 1.13 and if would like to scaffold --plugins=v3 then uninstall/install another go version. 
- Also, keep 2 diffs go versions will make hard to us collab work with
- the tests are NOT implemented to accommodate 2 go versions for example, how we can use teh go 1.13 to gen the project-v2 and go 1.15 for the project-v3?

So, IHMO we cannot have go 1.13 and go 1.15. 
c/c @rashmigottipati

pkg/plugin/v2/scaffolds/internal/templates/dockerfile.go
rashmigottipati: Reverted v2's Go version to 1.13.

estroz:
@rashmigottipati can you add a comment above `checkGoVersion()` as to why it wasn't updated? (because the kubebuilder binary only cares that the go binary supports modules, which were on by default in go 1.13)

rashmigottipati:
thanks @estroz, just added a comment as per our discussion.

estroz (approved):
/lgtm

PR #1720, opened 2020-10-13
:book: add a note that the homebrew recipe for kubebuilder is broken

the-gigi:

### DIFF
--- a/docs/book/getting_started/installation_and_setup.md
+++ b/docs/book/getting_started/installation_and_setup.md
@@ -25,6 +25,8 @@ mv kubebuilder_${version}_darwin_${arch} kubebuilder && sudo mv kubebuilder /usr
 export PATH=$PATH:/usr/local/kubebuilder/bin
 ```
 
+Don't use homebrew to install kubebuilder. The [homebrew recipe is broken](https://github.com/Homebrew/homebrew-core/issues/44175).
+
 {% sample lang="linux" %}
 ```bash
 version=1.0.8 # latest stable version


gabbifish:
I'll defer to @camilamacedo86 on this, but I think the homebrew installation actually works as desired--it installs only the Kubebuilder binary. The kube-apiserver, etcd, and kubectl binaries are requirements for using Kubebuilder's test suite. Perhaps we can reiterate that those binaries are common Kubernetes binaries that users are expected to already have if they're doing any Kubebuilder work. Moreover, installing those extra binaries could negatively interfere with preexisting binary versions users might already have on their system (e.g. someone installing Kubebuilder probably doesn't expect their local etcd binary to be updated as part of installing Kubebuilder!)

Camila, what do you think?

docs/book/getting_started/installation_and_setup.md
```diff
@@ -25,6 +25,8 @@ mv kubebuilder_${version}_darwin_${arch} kubebuilder && sudo mv kubebuilder /usr
 export PATH=$PATH:/usr/local/kubebuilder/bin
 ```
 
+Don't use homebrew to install kubebuilder. The [homebrew recipe is broken](https://github.com/Homebrew/homebrew-core/issues/44175).
```
camilamacedo86, line 28: WDYT about we fix the formula instead of adding this note?

gabbifish: I agree, in any homebrew issue we should fix the formula rather than the documentation. I still think the homebrew formula works as expected here, though?

camilamacedo86:
Hi @the-gigi, 

I hope that is fine we close this one since shows that we reach a consensus over the approach.
would be very nice to get your help to fix the formula. 

However, please feel free to ping us if you do not agree ok.

PR #1725, opened 2020-10-15
:bug: (v3-alpha) Makefile: fix source error, improve output text

estroz:
The `test` rule uses the `source` command which is a `bash` util, but the default shell is `/bin/sh` which may not have this function. Explicitly invoking commands with `bash` in these recipes guarantees that this command, and others, are available.

Additionally this target prints a bunch of commands unnecessarily. Using a make condition stops this from happening.

/kind bug

### DIFF
--- a/pkg/plugin/v3/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugin/v3/scaffolds/internal/templates/makefile.go
@@ -57,6 +57,10 @@ func (f *Makefile) SetTemplateDefaults() error {
 
 //nolint:lll
 const makefileTemplate = `
+# Set the shell used to bash for better error handling.
+SHELL = /bin/bash
+.SHELLFLAGS = -e -o pipefail -c
+
 # Image URL to use all building/pushing image targets
 IMG ?= {{ .Image }}
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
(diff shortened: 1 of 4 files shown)


pkg/plugin/v3/scaffolds/internal/templates/makefile.go
```diff
@@ -71,12 +75,20 @@ endif
...
+ifeq (,$(wildcard $(ENVTEST_ASSETS_DIR)/setup-envtest.sh))
+	mkdir -p $(ENVTEST_ASSETS_DIR)
+	curl -sSLo $(ENVTEST_ASSETS_DIR)/setup-envtest.sh $(ENVTEST_SCRIPT_URL)
+endif
+	{ \
+	source $(ENVTEST_ASSETS_DIR)/setup-envtest.sh && \
+	fetch_envtest_tools $(ENVTEST_ASSETS_DIR) && \
+	setup_envtest_env $(PWD)/$(ENVTEST_ASSETS_DIR) && \
+	go test ./... -coverprofile cover.out ; \
+	}
```
camilamacedo86, line 91: Hi @estroz, 

since the way that we would do the target was more a @joelanford's idea. I think would be nice ping him here to see what he thinks about that.

estroz:
This would be better fixed by https://github.com/kubernetes-sigs/controller-runtime/pull/1258. If/when that PR gets merged I will updated this one with the improved script usage.

/hold

testdata/project-v3-multigroup/Makefile
```diff
@@ -13,12 +17,20 @@ endif
...
+ifeq (,$(wildcard $(ENVTEST_ASSETS_DIR)/setup-envtest.sh))
+	mkdir -p $(ENVTEST_ASSETS_DIR)
+	curl -sSLo $(ENVTEST_ASSETS_DIR)/setup-envtest.sh $(ENVTEST_SCRIPT_URL)
+endif
+	{ \
+	source $(ENVTEST_ASSETS_DIR)/setup-envtest.sh && \
+	fetch_envtest_tools $(ENVTEST_ASSETS_DIR) && \
+	setup_envtest_env $(PWD)/$(ENVTEST_ASSETS_DIR) && \
+	go test ./... -coverprofile cover.out ; \
+	}
```
camilamacedo86, line 33: My main concern with this current solution are:
 
- Users are unable to run the tests to troubleshooting via delve or an IDE since the test will ONLY work if we execute them via the source.   
- The target is complex and not too much unstable which reduces it maintainability
- Its dependency with controller-runtime which make harder we keep it maintained. See that controller-runtime is only a framework and who install/use the bins is ONLY the kubebuilder. So, IMO its implementation should be done in Kubebuilder side. Note that if we want to change something on the steps required to get the binaries we need to bump a new version of the controller-runtime and wait for the next release whcih for example do not allow we do it for v2+ since controller-runtime has breaking changes in the upper version.  

IMO: I prefer the approach done in the PR https://github.com/kubernetes-sigs/kubebuilder/pull/1711 which solves all problems described above. 

c/c @estroz @joelanford @DirectXMan12

DirectXMan12: responded on #1711

Issue #1719, opened 2020-10-09
Once you move main.go scaffolding can't find it to add webhook configs


christopherhein:
When trying to scaffold a new webhook if the `main.go` has moved it will fail to create anything even just the webhook file and ignoring updating of the `main.go`.

```
→ kubebuilder create webhook --defaulting --kind NestedControlPlane --group controlplane --version v1alpha3
Writing scaffold for you to edit...
apis/controlplane.cluster.x-k8s.io/v1alpha3/nestedcontrolplane_webhook.go
2020/10/09 11:35:40 failed to create webhook: failed to open main.go: open main.go: no such file or directory
```

I'm able to do the `create` command for new APIs and it will still succeed when the `main.go` is moved even though it errors. I'd expect that I can at least create the `{resource}_webhook.go` file at least.

What versions of software are you using?  Specifically, the following are often useful:

```
→ go version
go version go1.15.2 darwin/amd64
```

```
→ kubebuilder version
Version: version.Version{KubeBuilderVersion:"2.3.1", KubernetesVendor:"1.16.4", GitCommit:"8b53abeb4280186e494b726edf8f54ca7aa64a49", BuildDate:"2020-03-26T16:42:00Z", GoOs:"unknown", GoArch:"unknown"}
```

```
sigs.k8s.io/controller-runtime v0.5.11
```

```
sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0
```

```
→ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T21:51:49Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-30T20:19:45Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

/kind bug

camilamacedo86:
Hi @christopherhein,

The tool will scaffold data in the main.go so if you change the layout by moving the main.go it will no longer work. See that the `create api` also face the same; 

**Example**

- Steps to mock

```go
[...]

PR #1726, opened 2020-10-16
:bug: Add validation for the main.go file present in root directory

prafull01:
Description of the Change:
Added validation to check if main.go is present in the root directory or not.
Added the change for v2 and v3 as this is a bug fix.

Motivation of the change:
If user moves the main.go file and then tries to create any resources like webhook, the operation is done half way, the webhook gets created but kb errors out when trying to write to main file. This should be early detected and the operation should not be allowed at all.

Fixes: #1719

### DIFF
--- a/pkg/plugin/v3/api.go
+++ b/pkg/plugin/v3/api.go
@@ -43,6 +43,9 @@ import (
 // TODO: remove this when a better solution for using addons is implemented.
 const KbDeclarativePatternVersion = "v0.0.0-20200522144838-848d48e5b073"
 
+// DefaultMainPath is default file path of main.go
+const DefaultMainPath = "main.go"
+
 type createAPIPlugin struct {
 	config *config.Config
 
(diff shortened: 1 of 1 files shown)


camilamacedo86 (approved):
/approve
/ok-to-test

Great work 🥇

prafull01:
/cc @gabbifish 
Could you please review??

pkg/plugin/v2/api.go
```diff
@@ -131,6 +134,11 @@ func (p *createAPIPlugin) Validate() error {
 		return err
 	}
 
+	// check if main.go is present in the root directory
+	if _, err := os.Stat(DefaultMainPath); os.IsNotExist(err) {
```
christopherhein, line 138: Will doing this break the current workflow that allows `kubebuilder create api` to create all necessary files even though it returns an error?

camilamacedo86: It would be a bug fix but I agree with you. let's apply it only in v3+ plugin. @prafull01 could you please do this change?

/hold 
until add only to v2

christopherhein: Thanks for validating my suspicions.

prafull01: Done.. Thanks @christopherhein for pointing it out.

christopherhein (approved):
/lgtm

PR #1728, opened 2020-10-19
:sparkles: Move the webhookcainjection_patch.yaml file creation from init project to webhook creation

### LINKED ISSUE #1727: The manifest config/default/webhookcainjection_patch.yaml ought to be scaffold in the webhook plugin
#### What did you do?

Init a project with kubebuider init

#### What did you expect to see?

Only manifest which are part of the init domain of responsibility be scaffolded. 

#### What did you see instead? Under which circumstances?

The file `config/default/webhookcainjection_patch.yaml` be scaffold as well.  

#### Describe the problem you would like to resolve

- I am an operator dev would like to see specific manifests which are required/useful for the webhooks only/if my project has webhooks  
- I am a consumer of the kubebuilder plugins would like that each plugin only scaffold manifests which are related to its domain so that, I can decide what plugins I'd like to consume or not. Example: I would like to consume the init and create api one, however, the webhook plugins are not important for me, so that I would NOT like that the init plugin do scaffolds that are only useful for the webhooks.

#### Describe the solution you'd like.

The boilerplate for this manifest is [here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/v3/scaffolds/internal/templates/config/kdefault/enablecainection_patch.go). I'd like that it would be scaffold when the command `kb create webhook` is executed instead of it. We can check if exist and if not then, do the scaffold.  

/kind bug

prafull01:
Description of the Change:
webhookcainjection_patch.yaml file is created at the init of the project, I have changed that to be created at the time of webhook creation.

Motivation:
I am a consumer of the kubebuilder plugins would like that each plugin only scaffold manifests which are related to its domain so that, I can decide what plugins I'd like to consume or not.

Fixes: #1727


pkg/plugin/v3/scaffolds/internal/templates/config/webhook/enablecainection_patch.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package kdefault
+package webhook
```
estroz, line 17: This file will still be scaffolded in `config/default`, so should remain in the `config/kdefault` package.

pkg/plugin/v2/scaffolds/internal/templates/webhook/enablecainection_patch.go
```diff
@@ -37,7 +37,8 @@ func (f *InjectCAPatch) SetTemplateDefaults() error {
 
 	f.TemplateBody = injectCAPatchTemplate
 
-	f.IfExistsAction = file.Error
+	// If file webhookcainjection_patch.yaml exist, skip its creation
```
estroz, line 40: Only the v3 plugin should be changed, so revert changes in `pkg/plugin/v2`.

estroz (commented):
If we're only making these changes to support the outcome of #1644's discussion about flags, we should hold off on reviewing this PR until that discussion is finalized. If we decide to go with `create webhook --webhook-version` over `init --webhook-version`, then this PR can go ahead for review.

/hold

pkg/plugin/v2/scaffolds/internal/templates/webhook/enablecainection_patch.go
camilamacedo86: IMO would be fine do it for v2+ since it shows more a bug. However, I would be ok with change only for v3+

estroz: If I understand correctly, scaffolding this patch during `init` is intentional because it is directly referenced in `config/default/kustomization.yaml`. @DirectXMan12 is that correct?

pkg/plugin/v3/scaffolds/internal/templates/config/webhook/enablecainection_patch.go
prafull01: Done..

pkg/plugin/v2/scaffolds/internal/templates/webhook/enablecainection_patch.go
camilamacedo86: Hi @estroz, 

I do not think that it was intentional. It may be hurting concepts such as encapsulation, single responsibility principle and almost for sure cohesion that may cause unexpected sides effects such difficulty to read, maintain/improve and extend, to mention a few. Also, IMO:

As @pwittrock spoke well another day, it is very hard for users to know and maintain so many config files, in this way, allow the init create/delete/update webhooks manifests not make harder for us(contributors) but also for users since it means more unnecessary config files for their scenarios. So, the use case that it hurts here would be `I am an operator dev would like to see specific manifests which are required/useful for the webhooks only/if my project has webhooks`

Also, note that when we introduce the plugin design we start to allow the plugins to be consumed for the other projects. Then, it means that as a developer I can decide to consume only the init and the create the plugin as well which means that my projects would NOT support webhooks then, would not make sense have webhooks flags or scaffolded manifests in the init command/plugin. It shows hurt the domains of responsibility as I described in https://github.com/kubernetes-sigs/kubebuilder/pull/1644.

camilamacedo86:
I think we can : 
/hold cancel 

based on the discussion made in the bug triage meeting.

pkg/plugin/v3/scaffolds/webhook.go
```diff
@@ -19,6 +19,8 @@ package scaffolds
 import (
 	"fmt"
 
+	"sigs.k8s.io/kubebuilder/pkg/plugin/v3/scaffolds/internal/templates/config/kdefault"
```
camilamacedo86, line 22: Nit: Sort the import. It needs to be at the end.

pkg/plugin/v3/scaffolds/internal/templates/config/kdefault/enablecainection_patch.go
```diff
@@ -37,7 +37,8 @@ func (f *InjectCAPatch) SetTemplateDefaults() error {
 
 	f.TemplateBody = injectCAPatchTemplate
 
-	f.IfExistsAction = file.Error
+	// If file webhookcainjection_patch.yaml exist, skip its creation
+	f.IfExistsAction = file.Skip
```
camilamacedo86, line 41: I liked that 👍

pkg/plugin/v3/scaffolds/init.go
```diff
@@ -126,7 +126,6 @@ func (s *initScaffolder) scaffold() error {
 		&webhook.Kustomization{},
 		&webhook.KustomizeConfigWebhook{},
 		&webhook.Service{},
```
camilamacedo86, line 128: IMO, I think we need also move these ones to the webhook plugin as well since they are specific to the webhook plugin and are not useful to projects which as not in the webhooks. I mean, it shows to be the same scenario.

pkg/plugin/v3/scaffolds/webhook.go
prafull01: Done

pkg/plugin/v3/scaffolds/init.go
prafull01: Yes, I also thinks the same..  Should I move all the webhook related scaffolding in this PR itself? Or should we wait for discussion to be finalized on #1644 ?

camilamacedo86: I understood that we agreed to move forward with into the bug triage. However, let's see if @estroz would like to hold on that until we finalized on #1644.

estroz (approved):
/lgtm

pkg/plugin/v2/scaffolds/internal/templates/webhook/enablecainection_patch.go
estroz: To clarify what I mean: the choice to scaffold this file in `init` may have been sub-optimal, but it was a conscious (intentional) one. Regardless the right move is to move this to `create webhook`.

camilamacedo86:
Since @estroz /lgtm this one and just hold until the author updates its title with :sparkles: shows that all is done to get it merged.
We will still require a follow up for https://github.com/kubernetes-sigs/kubebuilder/pull/1728/files#r508778432. 
However, it shows fine we move forward with so

/hold cancel

estroz:
Ah didn’t see the title was updated.

/hold cancel

Issue #1717, opened 2020-10-09
Controllers in a multigroup project belong to the 'controllers' package


miguelsorianod:
Hi,

I'm using operator-sdk v1.0.1, which, if I've understood correctly makes use of Kubebuilder for some of its commands.

I'm working on migrating an operator-sdk based project to operator-sdk 1.x version.

I've created the new project with the `multigroup: true` setting in the `PROJECT` file. Once that is done I generate the controller with `operator-sdk create api ...`, which after looking at the code I think is a command that belongs to kubebuilder, and I can see the directory structure is like this:
```
controllers/apps/apicast_controller.go
```
However, in the Go file of the controller I can see it has the following package definition:
```
package controllers
```

Notice how the package is `controllers` instead of `apps`. Is that expected/intentional? Might that be a bug? It seems odd.

I would have expected the package to be named `apps` in that directory.

With the current implementation I understand all controllers between different groups are sharing the same package name right now.

If you think this should be changed I'd be interested in contributing a change for this.

Software versions being used:

* Go version: 1.13.7
* kubebuilder version: I'm using the one that comes with `operator-sdk v1.0.1`, which I think uses the Go v2 plugin currently
* controller-runtime v0.6.2
* Kubernetes version: v1.18

Thank you.

/kind bug

camilamacedo86:
Hi @miguelsorianod, 

thank you for raise it. Could you please let us know if you will be working on to solve this scenario? if yes, could you please assign the issue to your self? Also, since it does not cause issues wdyt about we classify it as an RFE?

miguelsorianod:
Hi @camilamacedo86,

I'll work on this yes.
No issue with classifying this as an RFE :+1:

miguelsorianod:
Hi @camilamacedo86,

I've submitted a PR https://github.com/kubernetes-sigs/kubebuilder/pull/1729 with the changes.

Let me know what you think about them.

Thank you.

camilamacedo86:
I would classify that as RFE instead of a bug. Otherwise, I am totally all in for that 👍 
Really thank you for your contribution.

prafull01:
The PR with this RFE has been merged to master. We can close this one.
/close

PR #1729, opened 2020-10-19
⚠ Generate controllers with API Group name as package names in multi-group projects (only Go plugin v3)

miguelsorianod:
### Description

Use GroupPackageName (the API Group cleaned to be used as
the package name) as the package name when generating
controllers and controller suite test files with
non-empty API groups in MultiGroup projects.

When generating controller with an empty group name in a
MultiGroup project, or when generating a controller in a non
MultiGroup project keep the current behavior, which is
using the `controllers` package name for both the
controllers and controllers suite test files.

This applies only to Go v3 plugin as I understand this is a breaking change and breaking changes cannot be applied in v2. Can breaking changes be applied to v3 plugin code without the need of creating a new Go v4 plugin version? Is v3 still not considered "released"? By reading the versioning doc (https://github.com/kubernetes-sigs/kubebuilder/blob/master/VERSIONING.md#understanding-the-versions) it seems as if we'd need Go v4 plugin.

Let me know if there's something missing (doc? other stuff that I'm not aware of?)

### Motivation

With the behavior before this PR all controllers in a multigroup project defined their package name as "controllers" even when they were inside their own directories. In Go it's often common having the same directory name as the package name (when possible syntactically).

See issue https://github.com/kubernetes-sigs/kubebuilder/issues/1717 for more details.

### Related Issues

https://github.com/kubernetes-sigs/kubebuilder/issues/1717  Controllers in a multigroup project belong to the 'controllers' package

### DIFF
--- a/pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller.go
+++ b/pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller.go
@@ -55,7 +55,11 @@ func (f *Controller) SetTemplateDefaults() error {
 //nolint:lll
 const controllerTemplate = `{{ .Boilerplate }}
 
+{{if and .MultiGroup .Resource.Group }}
+package {{ .Resource.GroupPackageName }}
+{{else}}
 package controllers
+{{end}}
 
 import (
 	"context"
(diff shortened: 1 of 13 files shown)


camilamacedo86:
Hi @miguelsorianod, 

It is great 👍 Could you please follow up the instructions on the comment https://github.com/kubernetes-sigs/kubebuilder/pull/1729#issuecomment-712373033?

/ok-to-test

camilamacedo86:
Hi @miguelsorianod, 

For it gets merged you need to follow up the instructions on the comment #1729 (comment) and squash the commit as well.

estroz (approved):
/lgtm

estroz:
For squashing commits

/hold

miguelsorianod:
Hi,

I've squashed the commits and it seems the CLA check now passes.

PR #1731, opened 2020-10-20
🐛 use rbac.authorization.k8s.io/v1 for the auth proxy client

joelanford:
This PR updates the `metrics-reader` ClusterRole scaffolded by the v3 Go plugin to use apiVersion `rbac.authorization.k8s.io/v1` instead of the deprecated (since k8s 1.17) `rbac.authorization.k8s.io/v1beta1`.

All other template scaffolds are using `rbac.authorization.k8s.io/v1`. It seems that it was an oversight that caused this file to continue using `rbac.authorization.k8s.io/v1beta1`.

### DIFF
--- a/pkg/plugin/v2/scaffolds/internal/templates/metricsauth/clientclusterrole.go
+++ b/pkg/plugin/v2/scaffolds/internal/templates/metricsauth/clientclusterrole.go
@@ -40,7 +40,7 @@ func (f *ClientClusterRole) SetTemplateDefaults() error {
 	return nil
 }
 
-const clientClusterRoleTemplate = `apiVersion: rbac.authorization.k8s.io/v1beta1
+const clientClusterRoleTemplate = `apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: metrics-reader
--- a/pkg/plugin/v3/scaffolds/internal/templates/config/rbac/client_cluster_role.go
+++ b/pkg/plugin/v3/scaffolds/internal/templates/config/rbac/client_cluster_role.go
@@ -40,7 +40,7 @@ func (f *ClientClusterRole) SetTemplateDefaults() error {
 	return nil
 }
 
-const clientClusterRoleTemplate = `apiVersion: rbac.authorization.k8s.io/v1beta1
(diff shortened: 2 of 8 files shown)


joelanford:
Question: should this be backported to the v2 plugin? In v2, this also appears to be the only ClusterRole using `rbac.authorization.k8s.io/v1beta1`

camilamacedo86 (approved):
Hi @joelanford, 

It shows ok for me 👍 

/approve

> Question: should this be backported to the v2 plugin? In v2, this also appears to be the only ClusterRole using rbac.authorization.k8s.io/v1beta1

I do not think so. It would change the current behaviour of v2+, so by following the current approach we ough to push to only v3+
Could you please add in the title (only v3+), in POV it makes it easier for us to identify in the release notes that the change was applied only to v3+.

joelanford:
@camilamacedo86 Since v2 is already scaffolding other RBAC resources with `rbac.authorization.k8s.io/v1` (in fact ALL other RBAC resources use `v1`), to me this is a non-breaking backwards-compatible bug fix, and since this file is scaffolded during `init`, it will not have any effect on existing projects.

IMO, we should include this change for v2 as well.

camilamacedo86:
Hi @joelanford, 

> @camilamacedo86 Since v2 is already scaffolding other RBAC resources with rbac.authorization.k8s.io/v1 (in fact ALL other RBAC resources use v1), to me this is a non-breaking backwards-compatible bug fix, and since this file is scaffolded during init, it will not have any effect on existing projects.

> IMO, we should include this change for v2 as well.

I understand that ANY bug fixes can be addressed to v2+. However, we have been very restrictive and in many scenarios, besides the change be only a bug fix was decided to be applied just to v3+ because changes the behaviour.  

However, I checked [here](https://github.com/kubernetes-sigs/kubebuilder/search?q=rbac.authorization.k8s.io%2Fv1) that the other scaffolds are using v1 api already. So, I agree with you that would be ok in this case we push the change to v2 as well.

christopherhein:
I’m good with this change, it’s easy enough and “non-breaking” so much as the functionality still works no matter what apiversion as long as your control plane supports it.

joelanford:
It sounds like there's consensus to include this in the v2 plugin, so I'll update this PR to include this change there as well.

Issue #1655, opened 2020-08-28
Document kustomize version to be used with Kubebuilder 


rajasaur:
The CRDs generated by kustomize for Status Subresource handling seems to be dropped off in Kustomize 3.8.0 and beyond. After adding the `// +kubebuilder:subresource:status` marker which generates the 
``` 
subresources:
    status: {}
```
section in the config/crd/bases folder, running a `kustomize build config/crd` doesnt seem to update the final CRD with the status subresource. This causes any updates on the Status resource to fail with a. 404 because the Status object never got created in the first place.

This is the output of Kustomize 3.8.0 before.
```
emerald:d5doperator raja$ ./kustomize build config/crd | more  
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  annotations:
    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
    controller-gen.kubebuilder.io/version: v0.3.0
  name: pipelines.operations.starsystems.in
spec:
  conversion:
    strategy: Webhook
    webhookClientConfig:
      caBundle: Cg==
      service:
        name: webhook-service
        namespace: system
        path: /convert
  group: operations.starsystems.in
  names:
    kind: Pipeline
    listKind: PipelineList
    plural: pipelines
    singular: pipeline
  preserveUnknownFields: false
  scope: Namespaced
  subresources: {}
``` 

This is the output from Kustomize 3.7.0 :
```
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  annotations:
    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
    controller-gen.kubebuilder.io/version: v0.3.0
  creationTimestamp: null
  name: pipelines.operations.starsystems.in
spec:
  conversion:
    strategy: Webhook
    webhookClientConfig:
      caBundle: Cg==
      service:
        name: webhook-service
        namespace: system
        path: /convert
  group: operations.starsystems.in
  names:
    kind: Pipeline
    listKind: PipelineList
    plural: pipelines
    singular: pipeline
  preserveUnknownFields: false
  scope: Namespaced
  subresources:
    status: {}
```

Pls notice the output of the subresources in both the cases.

[...]

PR #1735, opened 2020-10-21
:book: add note about officially versions supported

camilamacedo86:
**Description**
Add a note to make clear the versions which are officially supported for the prerequisites and dependencies since are the versions used to test and whcih are scaffolded by default. 

**Motivation**
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/1655

**Preview**
https://deploy-preview-1735--kubebuilder.netlify.app/quick-start.html

### DIFF
--- a/docs/book/src/quick-start.md
+++ b/docs/book/src/quick-start.md
@@ -12,8 +12,18 @@ This Quick Start guide will cover:
 - [go](https://golang.org/dl/) version v1.15+.
 - [docker](https://docs.docker.com/install/) version 17.03+.
 - [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) version v1.11.3+.
-- [kustomize](https://sigs.k8s.io/kustomize/docs/INSTALL.md) v3.1.0+
 - Access to a Kubernetes v1.11.3+ cluster.
+
+<aside class="note">
+<h1>Versions and Supportability</h1>
+
+Projects created by Kubebuilder contain a Makefile that will install tools at versions defined at creation time. Those tools are:
+- [kustomize](https://kubernetes-sigs.github.io/kustomize/)
+- [controller-gen](https://github.com/kubernetes-sigs/controller-tools)
+
+The versions which are defined in the `Makefile` and `go.mod` files are the versions tested and therefore is recommend to use the specified versions.
+ 
+</aside>
 
 ## Installation
 


docs/book/src/quick-start.md
```diff
@@ -12,9 +12,15 @@ This Quick Start guide will cover:
 - [go](https://golang.org/dl/) version v1.15+.
 - [docker](https://docs.docker.com/install/) version 17.03+.
 - [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) version v1.11.3+.
-- [kustomize](https://sigs.k8s.io/kustomize/docs/INSTALL.md) v3.1.0+
 - Access to a Kubernetes v1.11.3+ cluster.
 
+<aside class="note">
+<h1>Versions</h1>
+
+Kubebuilder will create a project for you where the officially versions supported are defined on its Makefile and go.mod file. For example,  see that it will install a specific version for Kustomize which is a prerequisite for the tool.
```
estroz, line 20: Can you make a list of what these binaries are?

camilamacedo86: would not it be too verbose and may outdated easily? How do you think that it should be? Could you please give an example?

estroz: We should be listing project dependencies, and updating that list when we make changes. We don't add/remove tools often so I don't see this list getting outdated quickly.
```suggestion
Projects created by kubebuilder contain a Makefile that will install tools at versions defined at creation time. Those tools are:
- [kustomize](https://kubernetes-sigs.github.io/kustomize/)
- [controller-gen](https://github.com/kubernetes-sigs/controller-tools)
```

camilamacedo86: Done: 

Screen Shot 2020-10-22 at 16 08 28

Thanks for your feedback.

docs/book/src/quick-start.md
```diff
@@ -12,9 +12,18 @@ This Quick Start guide will cover:
...
 - Access to a Kubernetes v1.11.3+ cluster.
 
+<aside class="note">
+<h1>Versions and Supportability</h1>
+
+Projects created by Kubebuilder contain a Makefile that will install tools at versions defined at creation time. Those tools are:
+- [kustomize](https://kubernetes-sigs.github.io/kustomize/)
+- [controller-gen](https://github.com/kubernetes-sigs/controller-tools)
+
+Note that, the officially supported versions are that ones defined in the Makefile and go.mod files which are generated by tool.
```
estroz, line 24: No `go.mod` entry is created for any tool since they're installed in temp modules outside of a project tree
```suggestion
We test with and therefore recommend using the defined versions.
```

camilamacedo86: The pronoun `we` should be avoided. See: https://kubernetes.io/docs/contribute/style/style-guide/#avoid-using-we

Also, I'd like to make clear that the versions on go.mod and in the Makefile are the supported ones. I mean, users are able to customize it as they wish, however, we are unable to ensure the behaviour with other versions. I tried to address your suggestion and keep the idea/motivation. Let's see. Please, feel free to make a new suggestion if you wish as well.

estroz (approved):
/lgtm

PR #1738, opened 2020-10-22
:sparkles: not allow to init the project in a directory that is not cleaned (v3+ only)

camilamacedo86:
**Description**
Do not allow scaffold/init a project in a directory when it has other files already.
The only files that shows make sense be allowed are .gitignore and go.mod

**Motivation**
https://github.com/operator-framework/operator-sdk/issues/4067

### DIFF
--- a/pkg/plugin/v3/init.go
+++ b/pkg/plugin/v3/init.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v3
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -111,6 +112,11 @@ func (p *initPlugin) Validate() error {
 		}
 	}
 
+	// Check if the current directory has not files or directories which does not allow to init the project
+	if err := checkDir(); err != nil {
+		return err
+	}
+
 	// Check if the project name is a valid k8s namespace (DNS 1123 label).
 	if p.config.ProjectName == "" {
 		dir, err := os.Getwd()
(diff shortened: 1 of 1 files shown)


pkg/plugin/v3/init.go
```diff
@@ -167,3 +173,24 @@ func (p *initPlugin) PostScaffold() error {
...
+func checkDir() error {
+	files, err := ioutil.ReadDir(".")
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
+		if file.Name() != "go.mod" && file.Name() != ".gitignore" {
+			return fmt.Errorf("this directory has files already. " +
+				"Only the go.mod and .gitigore are allowed before the init.")
```
prafull01, line 192: nit: spelling
```suggestion
				"Only the go.mod and .gitignore are allowed before the init.")
```

pkg/plugin/v3/init.go
```diff
@@ -167,3 +173,26 @@ func (p *initPlugin) PostScaffold() error {
...
+// Otherwise, it might face issues to do the scaffold. The go.mod is allowed because user might run
+// go mod init before use the plugin it for not be required inform
+// the go module via the repo --flag.
+func checkDir() error {
+	err := filepath.Walk(".",
+		func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if info.Name() != "go.mod" && !strings.HasPrefix(info.Name(), ".") {
```
prafull01, line 189: The condition here should be OR
```suggestion
			if info.Name() != "go.mod" || !strings.HasPrefix(info.Name(), ".") {
```

PR #1741, opened 2020-10-24
WIP: :seedling: upgrade the cert-manager version used to get bugfix and keep the latest version before the breaking changes

camilamacedo86:
**Description**
-upgrade the cert-manager version used in the tests from 0.11.0 to 0.16.1. See: https://github.com/jetstack/cert-manager/releases/tag/v0.16.1

**Motivation**
It has a bug fix for the Kind. See that until we use the api version v1betav1 for webhooks we also need to do the tests with the cert-manager < 1.0.0. So, it will keep the latest version before the breaking changes

### DIFF
--- a/test/e2e/utils/test_context.go
+++ b/test/e2e/utils/test_context.go
@@ -27,7 +27,7 @@ import (
 	. "github.com/onsi/ginkgo" //nolint:golint
 )
 
-const certmanagerVersion = "v0.11.0"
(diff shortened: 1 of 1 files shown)


camilamacedo86:
If we upper to this version then, it is not working with `pull-kubebuilder-e2e-k8s-1-14-1`. So, I am closing this one.

Issue #1730, opened 2020-10-19
Add support to run in local mode using gcp, azure, etc auth providers.


camilamacedo86:
**What do you want to happen?** 

Add the following in the main.go: 

https://github.com/operator-framework/operator-sdk/blob/4054b5f044bbb87e452428f5df6a5be747c552b8/cmd/operator-sdk/main.go#L18-L20

If we confirm that users are unable to run projects locally (make run) in some clusters such as Azure, GCP, OIDC, etc.

**Is there currently another issue associated with this?** 
Motivated by discussion over https://github.com/operator-framework/operator-sdk/issues/4059 in the SDK bug triage meeting

**Notes**
- Add the scaffold changes only for v3+ (kubebuilder/pkg/plugin/v3/)
- Then, do not forget to run `make generate` to update the testdata dir. 
- Also, ensure that the change was scaffolded in the files (testdata dir)
- Note that we have already `_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"`: https://github.com/kubernetes-sigs/kubebuilder/blob/b2ce2e00a935578beb25d7efa20e7e5b3f6d108c/testdata/project-v2-addon/main.go#L26. So, it can be replaced for this suggestion.

/kind feature

camilamacedo86:
The user in operator-framework/operator-sdk#4059 confirmed that the project is working in the GKE when we run locally. 
So, before we do it we need to be sure that users are unable to run the projects in the cluster such as Azure, GCP, OIDC, etc. We need to be check if it is really required first at all.

PR #1742, opened 2020-10-25
:sparkles: Add all kubernetes auth plugins for Azure, OIDC and gcp (only v3+)

prafull01:
Description of Change:
Added import path for all the kubernetes auth plugins in main.go

Motivation of Change:
Currently, main.go file imports auth plugins for the gcp only, instead of import the auth plugins for all gcp, azure and OIDC.

Fixes: #1730


camilamacedo86 (approved):
It shows great. Really thank you for the contribution 🥇  

Since it is a small task that was discussed in the bug triage. I think has no reason for:

/lgtm
/approve 

This one 👍

PR #1745, opened 2020-10-27
👻 Enhance regex patterns

Adirio:
# Descrition
- Used non-capturing groups when the info is not relevant
- Simplified some expressions

# Motivation
While non-capturing groups should be a bit more performant that capturing groups, the improvement will be negligible. The motivations is mainly improving the clarity and readability of the regex expressions by simplifying some patters and not having unused capture groups.

/kind cleanup

### DIFF
--- a/pkg/internal/validation/dns.go
+++ b/pkg/internal/validation/dns.go
@@ -26,9 +26,9 @@ import (
 // "k8s.io/apimachinery" is needed, re-consider whether to add the dependency.
 
 const (
-	dns1123LabelFmt     string = "[a-z0-9]([-a-z0-9]*[a-z0-9])?"
+	dns1123LabelFmt     string = "[a-z0-9](?:[-a-z0-9]*[a-z0-9])?"
 	dns1123SubdomainFmt string = dns1123LabelFmt + "(\\." + dns1123LabelFmt + ")*"
-	dns1035LabelFmt     string = "[a-z]([-a-z0-9]*[a-z0-9])?"
+	dns1035LabelFmt     string = "[a-z](?:[-a-z0-9]*[a-z0-9])?"
 )
 
 type dnsValidationConfig struct {
--- a/pkg/model/resource/options_test.go
+++ b/pkg/model/resource/options_test.go
@@ -47,27 +47,27 @@ var _ = Describe("Resource Options", func() {
 			options := &Options{Group: "crew", Version: "1", Kind: "FirstMate"}
 			Expect(options.Validate()).NotTo(Succeed())
 			Expect(options.Validate().Error()).To(ContainSubstring(
-				`version must match ^v\d+(alpha\d+|beta\d+)?$ (was 1)`))
+				`version must match ^v\d+(?:alpha\d+|beta\d+)?$ (was 1)`))
 
 			options = &Options{Group: "crew", Version: "1beta1", Kind: "FirstMate"}
 			Expect(options.Validate()).NotTo(Succeed())
 			Expect(options.Validate().Error()).To(ContainSubstring(
-				`version must match ^v\d+(alpha\d+|beta\d+)?$ (was 1beta1)`))
+				`version must match ^v\d+(?:alpha\d+|beta\d+)?$ (was 1beta1)`))
 
 			options = &Options{Group: "crew", Version: "a1beta1", Kind: "FirstMate"}
 			Expect(options.Validate()).NotTo(Succeed())
 			Expect(options.Validate().Error()).To(ContainSubstring(
-				`version must match ^v\d+(alpha\d+|beta\d+)?$ (was a1beta1)`))
+				`version must match ^v\d+(?:alpha\d+|beta\d+)?$ (was a1beta1)`))
 
 			options = &Options{Group: "crew", Version: "v1beta", Kind: "FirstMate"}
 			Expect(options.Validate()).NotTo(Succeed())
 			Expect(options.Validate().Error()).To(ContainSubstring(
-				`version must match ^v\d+(alpha\d+|beta\d+)?$ (was v1beta)`))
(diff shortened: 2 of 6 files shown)


Adirio:
I used 👻 as this is a minor change that I don't think is worth to include in changelogs, but the verifier is not understanding it despite being mentioned in CONTRIBUTING.md. Guess I could change it to ✨ if needed.

pkg/internal/validation/dns.go
```diff
@@ -26,9 +26,9 @@ import (
 // "k8s.io/apimachinery" is needed, re-consider whether to add the dependency.
 
 const (
-	dns1123LabelFmt     string = "[a-z0-9]([-a-z0-9]*[a-z0-9])?"
+	dns1123LabelFmt     string = "[a-z0-9](?:[-a-z0-9]*[a-z0-9])?"
 	dns1123SubdomainFmt string = dns1123LabelFmt + "(\\." + dns1123LabelFmt + ")*"
-	dns1035LabelFmt     string = "[a-z]([-a-z0-9]*[a-z0-9])?"
+	dns1035LabelFmt     string = "[a-z](?:[-a-z0-9]*[a-z0-9])?"
```
camilamacedo86, line 31: @Adirio, 

For we not add ore one dependency to the project we decided to copy and paste the implementation to validate. 
We should not change this file unless it was changed in the `k8s.io/apimachinery` 

See: https://github.com/kubernetes-sigs/kubebuilder/blob/175952919092cd7304e5e8aeddb9cce37b6a614b/pkg/internal/validation/dns.go#L24-L26

camilamacedo86: Could you please revert the changes here?

pkg/plugin/plugin.go
```diff
@@ -92,7 +92,7 @@ func ParseVersion(version string) (v Version, err error) {
 	}
 
 	// Parse stage suffix, if any.
-	v.Stage = strings.TrimPrefix(submatches[3], "-")
```
camilamacedo86, line 95: Why not trim the spaces?

Adirio: We are timming the dash (`-`), not the spaces, and by modifying the regex as I did the dash is not part of the capture group now, so no need to trim it.

pkg/plugin/plugin.go
```diff
@@ -75,15 +75,15 @@ func ParseVersion(version string) (v Version, err error) {
 		return v, errors.New("plugin version is empty")
 	}
 
-	// A valid version string will have 4 submatches, each of which may be empty: the full string, "v",
-	// the integer, and the stage suffix. Invalid version strings do not have 4 submatches.
+	// A valid version string will have 3 submatches, each of which may be empty: the full string,
+	// the integer, and the stage suffix. Invalid version strings do not have 3 submatches.
 	submatches := verRe.FindStringSubmatch(version)
-	if len(submatches) != 4 {
+	if len(submatches) != 3 {
```
camilamacedo86, line 81: Why the changes here would make it easier to understand?

camilamacedo86 (changes requested):
IMO the change in https://github.com/kubernetes-sigs/kubebuilder/pull/1745/files#r512653755 needs to be revert. I am not sure why the changes made make the code more understandable or not. I think it also needs to be checked by @estroz.

pkg/plugin/plugin.go
Adirio: One of the submatches was `v`. We are not using `v` for anything so getting a submatch for it is inneficient. I just removed that submatch, and therefore I had to subtract 1 to the total length and the indexes.

pkg/internal/validation/dns.go
Adirio: I would agree with you if the code was copy-pasted, but it has been already modified. You can [check it](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go). So Convering that capture group into a non-capture group tells the future us reading the code that we are not using that capture group to take a substring out of the regex.

camilamacedo86: Could we keep the sample implementation of https://github.com/kubernetes/apimachinery/blob/v0.18.6/pkg/util/validation/validation.go ? (copy and paste)

camilamacedo86: Catcher. I would prefer if possible we keep a copy and paste of https://github.com/kubernetes/apimachinery/blob/v0.18.6/pkg/util/validation/validation.go whcih would allow us to move to use the dependency if we wish without breaking changes. However, I will let it up to you and @ezric

camilamacedo86 (commented):
HI @Adirio, 

I am ok with the changes, however, I think it needs the @estroz ok as well if possible. 
My concern would only : https://github.com/kubernetes-sigs/kubebuilder/pull/1745#discussion_r512703138

Also, note that it is no passing in the CI. could you please fix that?

camilamacedo86 (dismissed):

pkg/model/resource/options_test.go
```diff
@@ -47,27 +47,27 @@ var _ = Describe("Resource Options", func() {
 			options := &Options{Group: "crew", Version: "1", Kind: "FirstMate"}
 			Expect(options.Validate()).NotTo(Succeed())
 			Expect(options.Validate().Error()).To(ContainSubstring(
-				`version must match ^v\d+(alpha\d+|beta\d+)?$ (was 1)`))
+				`version must match ^v\d+(?:alpha\d+|beta\d+)?$ (was 1)`))
```
camilamacedo86, line 50: IMO : it should be a const

Adirio: I also thought of that to prevent having it hardcoded a doxen times the thing is that the string is not exported from its package. I think the better approach would be to create an expecific error instead a generic one and test if this yields this kind of error. But I just wanted to do the less changes to get it to pass Travis for now.

estroz (approved):
/lgtm

PR #1748, opened 2020-10-28
⚠️ [Rename] Disambiguate plugin by calling subcommand what the plugin getters returns instead of plugin

Adirio:
# Description
Rename plugin getters return objects to `Subcommand`.

# Motivation
Currently, both the structs with the getters and those returned by them are called "plugin" (`Plugin -> Plugin`), leading to missunderstandings. In order to clarify which is which, one of them should be renamed. There are two possibilities:

1. Rename the struct with the getters to something like `PluginFactory`: `PluginFactory -> Plugin`
2. Rename the returned type to something like `Subcommand`: `Plugin -> Subcommand`

The word "plugin" is already extensively used in the code (comments, variable names, ...) to refer to the left term, so I figured that it would be easier to go for the second option.

When deciding which name to use, I saw how the base class for the right term structs was already `GenericSubcommand` and it basically represents that, with methods to bind flags and execute the command, so I opted for `Subcommand`.

### DIFF
--- a/pkg/cli/cmd_helpers.go
+++ b/pkg/cli/cmd_helpers.go
@@ -45,14 +45,15 @@ func errCmdFunc(err error) func(*cobra.Command, []string) error {
 	}
 }
 
-// runECmdFunc returns a cobra RunE function that runs gsub and saves the
-// config, which may have been modified by gsub.
+// runECmdFunc returns a cobra RunE function that runs subcommand and saves the
+// config, which may have been modified by subcommand.
 func runECmdFunc(
 	c *config.Config,
-	gsub plugin.GenericSubcommand, // nolint:interfacer
-	msg string) func(*cobra.Command, []string) error {
+	subcommand plugin.Subcommand, // nolint:interfacer
+	msg string,
+) func(*cobra.Command, []string) error {
 	return func(*cobra.Command, []string) error {
-		if err := gsub.Run(); err != nil {
+		if err := subcommand.Run(); err != nil {
 			return fmt.Errorf("%s: %v", msg, err)
 		}
 		return c.Save()
--- a/pkg/plugin/interfaces.go
+++ b/pkg/plugin/interfaces.go
@@ -22,18 +22,18 @@ import (
 	"sigs.k8s.io/kubebuilder/pkg/model/config"
 )
 
-// Base is an interface that defines the common base for all plugins
-type Base interface {
+// Plugin is an interface that defines the common base for all plugins
+type Plugin interface {
 	// Name returns a DNS1123 label string identifying the plugin uniquely. This name should be fully-qualified,
 	// i.e. have a short prefix describing the plugin type (like a language) followed by a domain.
 	// For example, Kubebuilder's main plugin would return "go.kubebuilder.io".
 	Name() string
 	// Version returns the plugin's version.
 	//
-	// Note: this version is different from config version.
+	// NOTE: this version is different from config version.
 	Version() Version
-	// SupportedProjectVersions lists all project configuration versions this
-	// plugin supports, ex. []string{"2", "3"}. The returned slice cannot be empty.
+	// SupportedProjectVersions lists all project configuration versions this plugin supports, ex. []string{"2", "3"}.
+	// The returned slice cannot be empty.
 	SupportedProjectVersions() []string
 }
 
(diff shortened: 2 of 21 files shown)


Adirio:
@estroz We talked about this yesterday, what do you think of this approach?

camilamacedo86 (changes requested):
Hi @Adirio, 

The [plugins](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-1.md) design whcih was introduced following the concept of a Plug-in Architecture.  Then, it means that operations can be extending whcih allows other tools to use 1..N plugins as also they have its own plugins.  

Then, the next step that we have been discussed is the `plugin phase 2` which has the purpose to allow us to aggregate and/or customize the plugins. For example, we might have a plugin to it the project that will do a scaffold with a layout A and also another one which will create a project with a layout B and both can be used.  

See [here](https://github.com/operator-framework/operator-sdk/blob/master/internal/cmd/operator-sdk/cli/cli.go#L58-L68) how the SDK CLI call all plugins currently.  Note that in the PROJET file we write all plugins that were used to generate the files. (currently, we have just: go, ansible., helm).
 
For a further understand and follow up this WIP see; https://github.com/kubernetes-sigs/kubebuilder/issues/1378. Note that an EP will be raised based on the discussions made on this issue as well. 

In this way, I'd like to recommend you first raise an EP and/or at least an issue where you will describe: what is the problem with the current design that you would like to solve? What are the [user stories](https://en.wikipedia.org/wiki/User_story) that you would like to address  E.g `As a <role> I can <capability>, so that <receive benefit>`? Describe the technical details with examples of usage as well whcih clarifies your motivations and ideas?  

By looking at the PR shows that you are changing the plugins for each one be describe as subcommand then, it shows that goes against the purpose and architecture adopted for it.

Adirio:
Hi @camilamacedo86,

This PR follows the same design you mention, it is just a rebranding or name change.

To refer to these names I'm going to use the `X -> Y` notation, where the term in the left (`X`) is the struct that has the `GetXxxxxPlugin` methods and the term in the right (`Y`) is the struct returned by these methods.

Right now, both are called `Plugin` (`Plugin -> Plugin`). I was talking with @estroz and having both with the same name leads to missunderstandings, so we thought that one of them should be renamed (just changing the name, the implementation and design behind it stays the same).

There are two options:
1. Changing the left term: `SomethingElse -> Plugin`
2. Changing the right term: `Plugin -> SomethingElse`

The word "plugin" is already extensively used in the code (comments, variable names, ...) to refer to the left term, so I figured that it would be easier to go for the second option (`Plugin -> SomethingElse`).

When deciding which name to use, I saw how the base class for the right term structs was already `GenericSubcommand` and it basically represents that, with methods to bind flags and execute the command, so I opted for `Subcommand` (`Plugin -> Subcommand`).

Is the motivation of these change (just renaming) clearer now?

estroz (commented):
These name changes make sense to me on a high level. Haven't thoroughly gone through this PR yet.

camilamacedo86:
Hi @Adirio, 

So, by reading your comment in https://github.com/kubernetes-sigs/kubebuilder/pull/1748#issuecomment-717964428 I understand that you are doing a "cleanup" with some renames then:
- Since you rename the plugin names which are exported it is breaking change for who is consuming the plugin could you please add update the title 
- Could you please update the first comment to clarifies this PR? Note that, `Add the subcommand concept to represent the actual operations returned by plugins.` lead to belive it introduces a new design to the cli when you are doing renames mainly only. I mean, what means `Add the subcommand concept`? May something such as; 

>**Description**
> // Add a summary of what was done here specifically. E.g
> - Rename .. to ... 
> - add validation for ... 
>**Motivation**
>// Why this change is required? What are the problems that are solved with it? User story usually is helpful as well to get the motivation.

Adirio:
@camilamacedo86 Title, description and motivation updated.

Adirio:
Ping @camilamacedo86 and @estroz

pkg/plugin/interfaces.go
```diff
@@ -43,77 +43,84 @@ type Deprecated interface {
...
 	Examples string
 }
 
-// InitPluginGetter is an interface that defines gets an Init plugin
-type InitPluginGetter interface {
-	Base
-	// GetInitPlugin returns the underlying Init interface.
-	GetInitPlugin() Init
+// InitPlugin is an interface for plugins that provide an `init` subcommand
+type InitPlugin interface {
```
estroz, line 73: To avoid stuttering with the package name, how about
```suggestion
type Init interface {
```
Ditto elsewhere

pkg/cli/cmd_helpers.go
```diff
@@ -45,14 +45,15 @@ func errCmdFunc(err error) func(*cobra.Command, []string) error {
 	}
 }
 
-// runECmdFunc returns a cobra RunE function that runs gsub and saves the
-// config, which may have been modified by gsub.
+// runECmdFunc returns a cobra RunE function that runs subcommand and saves the
+// config, which may have been modified by subcommand.
 func runECmdFunc(
```
estroz, line 50: Although unrelated, this function should be
```suggestion
func runEditCmdFunc(
```

estroz (changes requested):
Looks good, just one nit about naming.

camilamacedo86 (approved):
It shows great for me now either 👍 
just address the @estroz nit https://github.com/kubernetes-sigs/kubebuilder/pull/1748#pullrequestreview-523601955 for we are able to get it merged

/approve

PR #1751, opened 2020-10-29
🌱Follow the enum pattern for the plugin version stage

Adirio:
# Description
Turn plugin version stage into a enum instead of a string delegating parse, validation, compare from the Version type to the new type. The API for the rest of the packages remains the same. The tests have been extended to cover the `Stage` type.

# Motivation
Plugin version stage can be either stable, alpha or beta, which fitsnicely into the enum pattern. This also allows to move the stage related methods (such as how to compare one stage to another, or how to parse to and from strings) to be implemented in the stage type directly instead of having it in the version type.

/kind cleanup


camilamacedo86:
Hi @Adirio, 
 
Really thank you for the contribution 🥇 Could you please rebase with the master? it has a conflict. 
The changes here shows ok for me. I think would be nice to get the @estroz input in this way, since he was the main author of the previous code. /assign @estroz

Adirio:
Rebased and added the two remaining tests for completion.

pkg/plugin/version_test.go
```diff
@@ -0,0 +1,302 @@
...
+limitations under the License.
+*/
+
+package plugin
+
+import (
+	"sort"
+	"testing"
+
+	g "github.com/onsi/ginkgo" // An alias is required because Context is defined elsewhere in this package.
```
camilamacedo86, line 23: we do not use an alias for this import across the project. I'd prefer we keep all whiteout alias in this scenario.

camilamacedo86 (approved):
Just a nit. Otherwise, the changes show fine for me here. 
/approve

However, please check if @estroz cannot help to review it as well before it get merged.

pkg/plugin/version_test.go
Adirio: In `pkg/plugin/interfaces.go` we are defining a `Context` type. Therefore we can not use a `_` as the alias because `ginkgo.Context` would conflict with it.

P.S.: I did not add this change, it was part of the previous code, I just moved it.

estroz (approved):
/lgtm

Issue #1747, opened 2020-10-28
Validate the kubebuilder edit command for the project layout


prafull01:
Scenario1: When I am trying to run edit command without any inputs with no api created, throws following error
```
prafull@EMPID18004:~/go/src/github.com/prafull01/kubebuilder/proj$ ../bin/kubebuilder edit
Error: failed to edit project with version "3-alpha": can't find "COPY apis/ apis/"
Usage:
  kubebuilder edit [flags]

Examples:
# Enable the multigroup layout
        kubebuilder edit --multigroup

        # Disable the multigroup layout
        kubebuilder edit --multigroup=false
	

Flags:
  -h, --help         help for edit
      --multigroup   enable or disable multigroup layout

2020/10/28 11:07:49 failed to edit project with version "3-alpha": can't find "COPY apis/ apis/"
```
Scenario2: The same command I ran after creating one api and moving the scaffolding to multi group. The kb edit command without any inputs moved the project scaffolding to single group.
```
prafull@EMPID18004:~/go/src/github.com/prafull01/kubebuilder/proj$ ./bin/kubebuilder create api --group praf --kind Lakers --version v1
Create Resource [y/n]
y
Create Controller [y/n]
y
Writing scaffold for you to edit...
api/v1/lakers_types.go
controllers/lakers_controller.go
Running make:
$ make
/home/prafull/go/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
go build -o bin/manager main.go
prafull@EMPID18004:~/go/src/github.com/prafull01/kubebuilder/proj$ ../bin/kubebuilder edit --multigroup=true

prafull@EMPID18004:~/go/src/github.com/prafull01/kubebuilder/proj$ cat PROJECT 
domain: my.domain
layout: go.kubebuilder.io/v3-alpha
multigroup: true
projectName: proj
repo: sigs.k8s.io/kubebuilder
resources:
- group: praf
  kind: Lakers
  version: v1
version: 3-alpha

prafull@EMPID18004:~/go/src/github.com/prafull01/kubebuilder/proj$ ../bin/kubebuilder edit 

prafull@EMPID18004:~/go/src/github.com/prafull01/kubebuilder/proj$ cat PROJECT 
domain: my.domain
layout: go.kubebuilder.io/v3-alpha
projectName: proj
repo: sigs.k8s.io/kubebuilder
resources:
- group: praf
  kind: Lakers
[...]

Issue #1640, opened 2020-08-17
would be nice the edit multi-group:true fixes the layout of the project and move the pre-existing api's


camilamacedo86:
Currently, when we use the command edit --multigroup=true it will update the Dockerfile and the PROJECT file and then the new API's will be scaffolded in a diff layout where the group is use in the path. 

However, if the project has API's scaffolded already users requires to move them and create the path with the group name manually. I think would be nice if the command could do all process instead of requires manually changes.

camilamacedo86:
c/c @estroz

prafull01:
@camilamacedo86 If no one is working on it, I would be happy to pick it up.

camilamacedo86:
Hi @prafull01, 

Your help is more than welcome 👍 it is great . Please, feel free to help and work on in any issue that pleases you.

camilamacedo86:
Hi @prafull01, 

I think would be easier if we could to do first: https://github.com/kubernetes-sigs/kubebuilder/issues/1689. WDYT?

camilamacedo86:
Another ticked that is related and could make easier we keep the current behaviour for v2 and new behaviour for v3 only would be the issue https://github.com/kubernetes-sigs/kubebuilder/issues/1639 to make the edit to be a plugin as well. See that we have a plugin to [create the api](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/v3/api.go), other to [init the project](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/v3/init.go) and another to scaffold the [webhooks](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/v3/webhook.go). 

So, feel free to analyse and see the best approach for we address that. However, just to share and try to help the following sequence might simplify its implementations:

- Make the init command be a plugin with v2, v3 having the same implementations (copy and paster):  https://github.com/kubernetes-sigs/kubebuilder/issues/1639
- Then, we could simplify the current implementation for the paths onlyfor v3 #1689
- after that shows that would be easier to make the edit plugin be smart enought to do the replaces and move the files.

prafull01:
Sounds perfect.

camilamacedo86:
Closing this one.

camilamacedo86:
It was deferred since we consider in the bug triage meeting hard to keep the RFE maintained because users might customize their files which would be hard to edit after the scaffold.

[...]

PR #1754, opened 2020-10-29
:bug: kubebuilder edit ignore the errors if the project layout is already on same type

prafull01:
**Description of the change**

Add validation of kb edit command if the scaffolding is already on same side either `multigroup: true` or `multigroup: false`. kb edit command fails when it tries to change the Dockerfile that it doesn't find the apis or api directory respectively.

**Motivation of the change**
Fixes: #1747

### DIFF
--- a/pkg/plugin/v2/scaffolds/edit.go
+++ b/pkg/plugin/v2/scaffolds/edit.go
@@ -42,7 +42,6 @@ func NewEditScaffolder(config *config.Config, multigroup bool) scaffold.Scaffold
 
 // Scaffold implements Scaffolder
 func (s *editScaffolder) Scaffold() error {
-	s.config.MultiGroup = s.multigroup
 	filename := "Dockerfile"
 	bs, err := ioutil.ReadFile(filename)
 	if err != nil {
@@ -56,21 +55,29 @@ func (s *editScaffolder) Scaffold() error {
 			str,
 			"COPY api/ api/",
 			`COPY apis/ apis/`)
-		if err != nil {
-			return err
-		}
 	} else {
 		str, err = ensureExistAndReplace(
 			str,
 			"COPY apis/ apis/",
 			`COPY api/ api/`)
-		if err != nil {
-			return err
-		}
 	}
-	// false positive
-	// nolint:gosec
-	return ioutil.WriteFile(filename, []byte(str), 0644)
+
+	// Ignore the error encountered, if the file is already in desired format.
+	if err != nil && s.multigroup != s.config.MultiGroup {
+		return err
+	}
+
+	s.config.MultiGroup = s.multigroup
+
+	// Check if the str is not empty, because when the file is already in desired format it will return empty string
+	// because there is nothing to replace.
+	if str != "" {
+		// false positive
+		// nolint:gosec
+		return ioutil.WriteFile(filename, []byte(str), 0644)
+	}
+
+	return nil
 }
 
 func ensureExistAndReplace(input, match, replace string) (string, error) {
--- a/pkg/plugin/v3/scaffolds/edit.go
+++ b/pkg/plugin/v3/scaffolds/edit.go
@@ -42,7 +42,6 @@ func NewEditScaffolder(config *config.Config, multigroup bool) scaffold.Scaffold
 
 // Scaffold implements Scaffolder
 func (s *editScaffolder) Scaffold() error {
-	s.config.MultiGroup = s.multigroup
 	filename := "Dockerfile"
 	bs, err := ioutil.ReadFile(filename)
 	if err != nil {
@@ -56,21 +55,30 @@ func (s *editScaffolder) Scaffold() error {
 			str,
 			"COPY api/ api/",
 			`COPY apis/ apis/`)
-		if err != nil {
-			return err
-		}
+
 	} else {
 		str, err = ensureExistAndReplace(
 			str,
 			"COPY apis/ apis/",
 			`COPY api/ api/`)
-		if err != nil {
-			return err
-		}
 	}
-	// false positive
-	// nolint:gosec
-	return ioutil.WriteFile(filename, []byte(str), 0644)
+
+	// Ignore the error encountered, if the file is already in desired format.
+	if err != nil && s.multigroup != s.config.MultiGroup {
+		return err
+	}
+
+	s.config.MultiGroup = s.multigroup
+
+	// Check if the str is not empty, because when the file is already in desired format it will return empty string
+	// because there is nothing to replace.
+	if str != "" {
+		// false positive
+		// nolint:gosec
+		return ioutil.WriteFile(filename, []byte(str), 0644)
+	}
+
+	return nil
 }
 
 func ensureExistAndReplace(input, match, replace string) (string, error) {


camilamacedo86 (changes requested):
I do not think that we should do that because:
- The plugin/command kb edit can have more than one arg besides it has only the multigroup
- We can enable/disable the support via kb edit --multigroup=[true or false]

However, we could check if the project has multi-group resources scaffolded already before allow remove the support. WDYT?

prafull01:
> I do not think that we should do that because:
> 
>     * The plugin/command kb edit can have more than one arg besides it has only the multigroup
> 
>     * We can enable/disable the support via kb edit --multigroup=[true or false]
> 
> 
> However, we could check if the project has multi-group resources scaffolded already before allow remove the support. WDYT?

The current implementation also fails if we try to edit a project which is already on the same type. Current, failure happens in the PostScaffolding of the `kb edit` command.
```
prafull@EMPID18004:~/go/src/github.com/prafull01/kubebuilder/proj$ ../bin/kubebuilder edit --multigroup=true
prafull@EMPID18004:~/go/src/github.com/prafull01/kubebuilder/proj$ ../bin/kubebuilder edit --multigroup=true
Error: failed to edit project with version "3-alpha": can't find "COPY api/ api/"
Usage:
  kubebuilder edit [flags]

Examples:
# Enable the multigroup layout
        kubebuilder edit --multigroup

        kubebuilder edit --multigroup=false
	

Flags:
  -h, --help         help for edit
      --multigroup   enable or disable multigroup layout

2020/10/30 00:13:52 failed to edit project with version "3-alpha": can't find "COPY api/ api/"
```
This change will give more clear message to the user that he is already on the same type of the project layout. 
Also, this change will become more important when we add #1640 . When we try to modify the layout of the project, we don't need to do anything is the project is already on the same layout.

camilamacedo86:
@prafull01,  I catcher the scenario 👍 
Great catcher, btw. 

However, it is a really bugfix in POV. So, wdyt about update the title with :bug; and to do the same for v2?

prafull01:
> @prafull01, I catcher the scenario +1
> Great catcher, btw.
> 
> However, it is a really bugfix in POV. So, wdyt about update the title with :bug; and to do the same for v2?

Done

camilamacedo86 (approved):
great 👍 
/approve

prafull01:
@camilamacedo86 Needs /ok-to-test command to run the tests. Thanks :)

camilamacedo86:
/approve cancel 

Hi @prafull01, 

After tests and think on that further  I think it is not the best approach for we avoid this issue. 
See that after this pr if we run `kubebuilder edit` it will return the issue added here instead of the helper. So, WDYT about we add a check in the Dockerfile update [here](https://github.com/kubernetes-sigs/kubebuilder/blob/9c02d557f01bce86ce7df48a85f0e93aa3f39661/pkg/plugin/v2/scaffolds/edit.go#L73) to ignore the issue if the projectfile==multigroup value and the error faced is `can't find "COPY`?

camilamacedo86 (changes requested):
/approve cancel 

Hi @prafull01, 

After tests and think about that more,  I believe that it is not the best approach for we avoid this issue. 
See that after this pr if we run `kubebuilder edit` it will return the issue added here instead of the helper. So, WDYT about we add a check in the Dockerfile update [here](https://github.com/kubernetes-sigs/kubebuilder/blob/9c02d557f01bce86ce7df48a85f0e93aa3f39661/pkg/plugin/v2/scaffolds/edit.go#L73) to ignore the issue if the projectfile==multigroup value and the error faced is `can't find "COPY`?

prafull01:
Hi @camilamacedo86 I understood your concern that, the fix is currently blocking the user to run `kb edit` command if the user is already on the same type of project layout. However, I have a different fix in mind.

I can remove these three lines from [here](https://github.com/kubernetes-sigs/kubebuilder/blob/9c02d557f01bce86ce7df48a85f0e93aa3f39661/pkg/plugin/v2/scaffolds/edit.go#L77-L79)
```
	if !strings.Contains(input, match) {
		return "", fmt.Errorf("can't find %q", match)
	}
```
Now if it found the match it will return the replaced string, else it will return the same string. This way I don't need to check any hacks or error types to ignore. WDYT?

camilamacedo86:
Hi @prafull01, 

> Now if it found the match it will return the replaced string, else it will return the same string. his way I don't need to check any hacks or error types to ignore. WDYT?

IMO we need to return the error in the other possible scenarios. So, my idea was we check if all is fine in the update.

// if err ! nil && value from Project file == flag then do not raise the issue.

pkg/plugin/v2/scaffolds/edit.go
```diff
@@ -42,6 +43,10 @@ func NewEditScaffolder(config *config.Config, multigroup bool) scaffold.Scaffold
 
 // Scaffold implements Scaffolder
 func (s *editScaffolder) Scaffold() error {
+	config, err := internalconfig.Load()
+	if err != nil {
+		return err
+	}
```
camilamacedo86, line 49: It is not required. See that we have the s.config already.

pkg/plugin/v2/scaffolds/edit.go
```diff
@@ -56,21 +61,26 @@ func (s *editScaffolder) Scaffold() error {
...
+		if err != nil && s.multigroup != config.MultiGroup {
 			return err
 		}
 	} else {
 		str, err = ensureExistAndReplace(
 			str,
 			"COPY apis/ apis/",
 			`COPY api/ api/`)
-		if err != nil {
+		if err != nil && s.multigroup != config.MultiGroup {
```
camilamacedo86, line 72: We can centralize the error check after the if/else

pkg/plugin/v2/scaffolds/edit.go
```diff
@@ -56,21 +61,26 @@ func (s *editScaffolder) Scaffold() error {
...
 	}
-	// false positive
-	// nolint:gosec
-	return ioutil.WriteFile(filename, []byte(str), 0644)
+
+	if s.multigroup != config.MultiGroup {
+		// false positive
+		// nolint:gosec
+		return ioutil.WriteFile(filename, []byte(str), 0644)
+	}
```
camilamacedo86, line 81: the error that we are trying to avoid occurs in the ensureExistAndReplace.

pkg/plugin/v2/scaffolds/edit.go
```diff
@@ -56,21 +61,26 @@ func (s *editScaffolder) Scaffold() error {
...
+
+	return nil
```
camilamacedo86, line 83: Following my suggestiion: 


```go
// Scaffold implements Scaffolder
func (s *editScaffolder) Scaffold() error {
	filename := "Dockerfile"
	bs, err := ioutil.ReadFile(filename)
	if err != nil {
		return err
	}
	str := string(bs)

	// update dockerfile
	if s.multigroup {
		str, err = ensureExistAndReplace(
			str,
			"COPY api/ api/",
			`COPY apis/ apis/`)
	} else {
		str, err = ensureExistAndReplace(
			str,
			"COPY apis/ apis/",
			`COPY api/ api/`)
	}
	// ignore the error faced to replace when it 
	// is because the file is already in the desired format
	if err != nil && s.config.MultiGroup != s.multigroup {
		return err
	}
	
	s.config.MultiGroup = s.multigroup
	// false positive
	// nolint:gosec
	return ioutil.WriteFile(filename, []byte(str), 0644)
}
```

pkg/plugin/v2/scaffolds/edit.go
prafull01: done

pkg/plugin/v2/scaffolds/edit.go
prafull01: Done, with add the check `if str != ""` before writing to the file.

camilamacedo86 (approved):
/approve 

Great work 👍

pkg/plugin/v2/scaffolds/edit.go
prafull01: Added comments

camilamacedo86:
Tested locally and the bug fixed work. 
Since it is a very tiny fix I think that is fine we get it merged. 

/lgtm

PR #1758, opened 2020-10-30
✨ Add copyright and license to scaffolded webhook suite test file

Adirio:
# Description
Add copyright and license to scaffolded webhook suite test file that was previously lacking it.

# Motivation
The file had the `BoilerplateMixin` needed to add the copyright and license but was not being scaffolded with it. As it is just a comment, I think we can consider this not-breaking even if previously scaffolded test suites may not be updated.


prafull01:
@Adirio Good catch. I missed it :+1:

camilamacedo86 (approved):
Good catcher 👍 

/lgtm
/approve

PR #1759, opened 2020-10-30
🌱 Unify bolierplate with a single template body

Adirio:
# Description
Boilerplate code has one different template bodys for each accepted license. Modify it so that a single template body is used and the license is extracted from a map of licenses.

# Motivation
This PR has two main goals:
1. Reduce the duplicated code by not having multiple copyright sections of the boilerplate.
1. Stop templating parts of the file during the `SetTemplateDefault` call, as later modifications to the template struct will not be reflected in the scaffolded file. For example, lets supose we want to modify the license  to "none" after the defaults have been set. As the template body was already selected including the License already in it, our change to the license would not yield any error but would not be applied.

### DIFF
--- a/pkg/plugin/v2/scaffolds/internal/templates/boilerplate.go
+++ b/pkg/plugin/v2/scaffolds/internal/templates/boilerplate.go
@@ -34,6 +34,9 @@ type Boilerplate struct {
 	// License is the License type to write
 	License string
 
+	// Licenses maps License types to their actual string
+	Licenses map[string]string
+
 	// Owner is the copyright owner - e.g. "The Kubernetes Authors"
 	Owner string
 
@@ -41,10 +44,39 @@ type Boilerplate struct {
 	Year string
 }
 
+// Validate implements file.RequiresValidation
+func (f Boilerplate) Validate() error {
+	if f.License == "" {
+		// A default license will be set later
+	} else if _, found := knownLicenses[f.License]; found {
+		// One of the know licenses
+	} else if _, found := f.Licenses[f.License]; found {
+		// A map containing the requested license was also provided
+	} else {
+		return fmt.Errorf("unknown specified license %s", f.License)
+	}
+
+	return nil
+}
+
 // SetTemplateDefaults implements input.Template
 func (f *Boilerplate) SetTemplateDefaults() error {
 	if f.Path == "" {
 		f.Path = filepath.Join("hack", "boilerplate.go.txt")
+	}
+
+	if f.License == "" {
+		f.License = "apache2"
+	}
+
+	if f.Licenses == nil {
+		f.Licenses = make(map[string]string, len(knownLicenses))
+	}
+
+	for key, value := range knownLicenses {
+		if _, hasLicense := f.Licenses[key]; !hasLicense {
+			f.Licenses[key] = value
+		}
 	}
 
 	if f.Year == "" {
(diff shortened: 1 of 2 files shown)


camilamacedo86:
It is not a new feat to users so we can use (:ghost:)

pkg/plugin/v2/scaffolds/internal/templates/boilerplate.go
```diff
@@ -57,24 +86,25 @@ func (f *Boilerplate) SetTemplateDefaults() error {
...
 {{ if .Owner -}}
 Copyright {{ .Year }} {{ .Owner }}.
 {{- else -}}
 Copyright {{ .Year }}.
 {{- end }}
+{{ index .Licenses .License }}*/`
+
+var knownLicenses = map[string]string{
+	"apache2": apache2,
+	"none":    "",
```
camilamacedo86, line 104: We have none test/check to ensure the none. Do you think that we could cover it with unit tests?

Adirio: We had tests for each file in the past but we removed them because they don't scale well.

camilamacedo86: Hi @Adirio, 

The e2e tests have CI effort but the unit tests are not. Then, I cannot see how unit tests would not be scalable and I have no knowledge about we remove the unit tests in the past.I understand that the goal is to cover the project with the unit as much as possible. We actually added coveralls in the project in order to help us to improve its coverage see `coverage/coveralls — Coverage remained the same at 70.581%`.   


However, I tested it locally and shows that all worked fine. 

```
$ kubebuilder init --license=none
Get controller runtime:
$ go get sigs.k8s.io/controller-runtime@v0.6.3
Update go.mod:
$ go mod tidy
Running make:
$ make
/Users/camilamacedo/go/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
Next: define a resource with:
$ kubebuilder create api

$ cat hack/boilerplate.go.txt
/*
Copyright 2020.
```

So, IMO we can merge this one


/approve

camilamacedo86 (approved):
/approve

Adirio:
This PR just requires a `lgtm` label to be merged.

camilamacedo86:
Hi @Adirio, 

> This PR just requires a lgtm label to be merged. 

Let's a second reviewer check it as well.

Adirio:
@estroz could you take a look at this? Pretty small clean up change that should be easy to review.

pkg/plugin/v2/scaffolds/internal/templates/boilerplate.go
```diff
@@ -34,19 +34,48 @@ type Boilerplate struct {
...
+	if f.License == "" {
+		f.License = "apache2"
+	}
+
+	if f.Licenses == nil || len(f.Licenses) == 0 {
+		f.Licenses = make(map[string]string, len(knownLicenses))
+		for key, value := range knownLicenses {
+			f.Licenses[key] = value
+		}
+	}
```
estroz, line 77: Probably not necessary, but the following changes makes it so the caller can set custom licenses but still gets other defaults:
```suggestion
	if f.Licenses == nil {
		f.Licenses = make(map[string]string, len(knownLicenses))
	}
	for key, value := range knownLicenses {
		if _, hasLicense := f.Licenses[key]; !hasLicense {
			f.Licenses[key] = value
		}
	}
```

estroz (approved):
/lgtm

PR #1636, opened 2020-08-11
*: remove project version 1 references, update migration guide to 2 -> 3

estroz:
Project version 1 shouldn't be mentioned in code, and any project version that isn't supported by the binary used in a project should cause an error to be returned.

This PR also adds a skeleton migration guide with some TODO's for future work that will likely be in before the next major release.

/kind documentation

PR #1763, opened 2020-11-01
:sparkles: remove v1 version and only shows --plugins flag when the project version is != V2

camilamacedo86:
**Description**
- remove v1 version 
- shows plugins --flag only if Project Version is != V2

**Motivation**

https://github.com/kubernetes-sigs/kubebuilder/pull/1636

Co-author @estroz

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -64,7 +64,7 @@ func readFrom(fs afero.Fs, path string) (c config.Config, err error) {
 
 	// kubebuilder v1 omitted version, so default to v1
 	if c.Version == "" {
-		c.Version = config.Version1
+		return config.Config{}, fmt.Errorf("project version key `version` is empty or does not exist in %s", path)
 	}
 
 	return
@@ -178,4 +178,9 @@ type saveError struct {
 
 func (e saveError) Error() string {
 	return fmt.Sprintf("unable to save the configuration: %v", e.err)
+}
+
+// IsVersionSupported returns true if version is a supported project version.
+func IsVersionSupported(version string) bool {
+	return version == config.Version2 || version == config.Version3Alpha
 }
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -193,8 +193,8 @@ func (c *cli) initialize() error {
 		c.configured = true
 		c.projectVersion = projectConfig.Version
 
-		if projectConfig.IsV1() {
-			return fmt.Errorf(noticeColor, "project version 1 is no longer supported.\n"+
+		if !internalconfig.IsVersionSupported(c.projectVersion) {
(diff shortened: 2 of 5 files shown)


pkg/model/config/config_test.go
```diff
@@ -37,8 +37,13 @@ var _ = Describe("Config", func() {
 			expectedConfig Config
 		)
 
+		By("Using config version empty")
+		config = Config{}
+		pluginConfig = PluginConfig{}
+		Expect(config.EncodePluginConfig(key, pluginConfig)).NotTo(Succeed())
+
 		By("Using config version 1")
```
estroz, line 45: You can remove the version 1 test cases.

internal/config/config.go
```diff
@@ -179,3 +179,8 @@ type saveError struct {
 func (e saveError) Error() string {
 	return fmt.Sprintf("unable to save the configuration: %v", e.err)
 }
+
+// IsVersionSupportedByCLI returns true for a project version that is supported by the tool
+func IsVersionSupportedByCLI(version string) bool {
```
estroz, line 184:
```suggestion
// IsVersionSupported returns true if version is a supported project version.
func IsVersionSupported(version string) bool {
```

pkg/cli/cli.go
```diff
@@ -193,8 +193,8 @@ func (c *cli) initialize() error {
 		c.configured = true
 		c.projectVersion = projectConfig.Version
 
-		if projectConfig.IsV1() {
-			return fmt.Errorf(noticeColor, "project version 1 is no longer supported.\n"+
+		if !internalconfig.IsVersionSupportedByCLI(c.projectVersion) {
```
estroz, line 196:
```suggestion
		if !internalconfig.IsVersionSupported(c.projectVersion) {
```

pkg/cli/init.go
```diff
@@ -46,7 +46,7 @@ func (c *cli) newInitCmd() *cobra.Command {
 	cmd.Flags().String(projectVersionFlag, c.defaultProjectVersion,
 		fmt.Sprintf("project version, possible values: (%s)", strings.Join(c.getAvailableProjectVersions(), ", ")))
 	// The --plugins flag can only be called to init projects v2+.
-	if c.projectVersion != config.Version1 {
+	if c.projectVersion != config.Version2 {
```
estroz, line 49: I think this is ok, since project version 2 isn't the default anymore and doesn't have the concept of plugins in its spec.

estroz (approved):
/lgtm

Adirio:
So basically `v2` is going to use the `go.kubebuilder.io` plugin but not really support plugins, and `v3` and onwards will support plugins. Right?

estroz:
@Adirio yes. Project version 2 uses a plugin internally; from a user's perspective there is no plugin being used.

PR #1764, opened 2020-11-01
:seedling: add plugins directory to be checked in the tests and coveralls

camilamacedo86:
**Description**

With the introduction of the plugin design, the directory plugins were added to the project. However, the target and scripts which runs the go unit tests and the coveralls were not updated to cover the new directories.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,8 +83,8 @@ endif
 ##@ Tests
 
 .PHONY: go-test
-go-test: ## Run the go tests ($ go test -v ./cmd/... ./pkg/...)
-	go test -race -v ./cmd/... ./pkg/...
+go-test: ## Run the unit test
+	go test -race -v ./cmd/... ./pkg/... ./plugins/...
 
 .PHONY: test
 test: ## Run the unit tests (used in the CI)
@@ -95,7 +95,7 @@ test-coverage:  ## Run coveralls
 	# remove all coverage files if exists
 	- rm -rf *.out
 	# run the go tests and gen the file coverage-all used to do the integration with coverrals.io
-	go test -race -failfast -tags=integration -coverprofile=coverage-all.out ./pkg/... ./cmd/...
+	go test -race -failfast -tags=integration -coverprofile=coverage-all.out ./cmd/... ./pkg/... ./plugins/...
 
 .PHONY: test-e2e-local
 test-e2e-local: ## It will run the script to install kind and run e2e tests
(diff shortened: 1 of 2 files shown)


internal/config/config_test.go
```diff
@@ -166,25 +167,6 @@ plugins:
 	})
 
 	Context("with invalid keys", func() {
-		It("should return a save error", func() {
```
camilamacedo86, line 169: This test added does not work at all. We did not get that because it has not been called. So, I only removed it.

internal/config/config_test.go
```diff
@@ -134,16 +135,16 @@ version: "2"`
 repo: github.com/example/project
 version: "3-alpha"
 plugins:
-  plugin-x:
```
estroz, line 137: I don't think these indentation changes are needed.

camilamacedo86: That was required to make the tests pass. Otherwise, it fails since we are checking if string == string.

internal/config/config_test.go
estroz: Originally `Config.Save()` returned an error when project version was < 3. Now it does not, so this test can be removed.

Makefile
```diff
@@ -83,8 +83,8 @@ endif
 ##@ Tests
 
 .PHONY: go-test
-go-test: ## Run the go tests ($ go test -v ./cmd/... ./pkg/...)
-	go test -race -v ./cmd/... ./pkg/...
+go-test: ## Run the unit test
+	go test -race ./cmd/... ./internal/... ./pkg/... ./plugins/...
```
estroz, line 87:
```suggestion
	go test -race -v ./cmd/... ./internal/... ./pkg/... ./plugins/...
```

internal/config/config_test.go
estroz: This case does not check string equality. It passes locally for me without these indent changes.

camilamacedo86:
All done 👍 Really thank you for your review and support @estroz

Makefile
```diff
@@ -95,7 +95,7 @@ test-coverage:  ## Run coveralls
 	# remove all coverage files if exists
 	- rm -rf *.out
 	# run the go tests and gen the file coverage-all used to do the integration with coverrals.io
-	go test -race -failfast -tags=integration -coverprofile=coverage-all.out ./pkg/... ./cmd/...
+	go test -race -failfast -tags=integration -coverprofile=coverage-all.out  ./cmd/... ./pkg/... ./plugins/...
```
estroz, line 98:
```suggestion
	go test -race -failfast -tags=integration -coverprofile=coverage-all.out ./cmd/... ./pkg/... ./plugins/...
```

estroz (approved):
/approve
/lgtm

PR #1766, opened 2020-11-02
🌱 Remove unused commnadName field

Adirio:
# Description
Removes a field that was not being used inside the `edit` command.

# Motivation
Excess code removal. In the rest of the commands (`init`, `create`, ...) this field is used for error string formating but in `edit` it is not needed.

/kind cleanup


camilamacedo86 (approved):
/approve 
/lgtm 

Since it is very small.

Issue #1767, opened 2020-11-02
Make manifests is not working when  only the controller is scaffolded by the create api


camilamacedo86:
**What broke**

After creating a project and scaffold just the controller the make manifests target will fail :

- `kubebuilder int` or `kubebuilder init --plugins=go/v3-alpha`
- $ kubebuilder create api --group ship --version v1beta1 --kind Frigate

```
Create Resource [y/n]
n
Create Controller [y/n]
y
```

What did you expect to happen?  What do you think went wrong?

The controller-gen is falling with the files scaffolded in this scenario: 

```
Users/camilamacedo/go/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
main.go:33:2: package testv2/api/v1beta1 is not in GOROOT (/usr/local/go/src/testv2/api/v1beta1)
Error: not all generators ran successfully
run `controller-gen object:headerFile=hack/boilerplate.go.txt paths=./... -w` to see all available markers, or `controller-gen object:headerFile=hack/boilerplate.go.txt paths=./... -h` for usage
make: *** [Makefile:62: generate] Error 1
```

- master commit 9c02d557f01bce86ce7df48a85f0e93aa3f39661

prafull01:
Just checked this one myself.
/assign

PR #1770, opened 2020-11-02
:bug: Fix the kubebuilder api creation when resource creation is set to false

prafull01:
**Description**
Fix the `kb create api` command when the create resource is set to false.
Added tests in the generate_testdata.sh so that such issue doesn't occur again.

*Motivation**
Fixes: #1767

### DIFF
--- a/generate_testdata.sh
+++ b/generate_testdata.sh
@@ -65,6 +65,7 @@ scaffold_test_project() {
             $kb create webhook --group crew --version v1 --kind FirstMate --conversion
             $kb create api --group crew --version v1 --kind Admiral --controller=true --resource=true --namespaced=false --make=false
             $kb create webhook --group crew --version v1 --kind Admiral --defaulting
+            $kb create api --group crew --version v1 --kind Laker --controller=true --resource=false --make=false
         elif [ $project == "project-v2-multigroup" ] || [ $project == "project-v3-multigroup" ]; then
             header_text 'Switching to multigroup layout ...'
             $kb edit --multigroup=true
@@ -81,6 +82,7 @@ scaffold_test_project() {
             $kb create api --group sea-creatures --version v1beta1 --kind Kraken --controller=true --resource=true --make=false
             $kb create api --group sea-creatures --version v1beta2 --kind Leviathan --controller=true --resource=true --make=false
             $kb create api --group foo.policy --version v1 --kind HealthCheckPolicy --controller=true --resource=true --make=false
+            $kb create api --group apps --version v1 --kind Pod --controller=true --resource=false --make=false
             if [ $project == "project-v3-multigroup" ]; then
               $kb create api --version v1 --kind Lakers --controller=true --resource=true --make=false
               $kb create webhook --version v1 --kind Lakers --defaulting --programmatic-validation
--- a/pkg/plugin/v2/scaffolds/internal/templates/controller/controller.go
+++ b/pkg/plugin/v2/scaffolds/internal/templates/controller/controller.go
@@ -26,11 +26,15 @@ import (
 var _ file.Template = &Controller{}
 
 // Controller scaffolds a Controller for a Resource
+// nolint:maligned
 type Controller struct {
 	file.TemplateMixin
 	file.MultiGroupMixin
 	file.BoilerplateMixin
 	file.ResourceMixin
+
+	// WireResource defines the api resources are generated or not.
+	WireResource bool
 }
 
 // SetTemplateDefaults implements input.Template
@@ -63,7 +67,9 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	{{ if .WireResource -}}
 	{{ .Resource.ImportAlias }} "{{ .Resource.Package }}"
+	{{- end }}
 )
 
 // {{ .Resource.Kind }}Reconciler reconciles a {{ .Resource.Kind }} object
@@ -87,7 +93,12 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(req ctrl.Request) (ctrl.Resul
 
 func (r *{{ .Resource.Kind }}Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
+		{{ if .WireResource -}}
 		For(&{{ .Resource.ImportAlias }}.{{ .Resource.Kind }}{}).
+		{{- else -}}
+		// Uncomment the following line adding a pointer to an instance of the controlled resource as an argument
+		// For().
+		{{- end }}
 		Complete(r)
 }
 `
--- a/pkg/plugin/v2/scaffolds/internal/templates/controller/controller_suitetest.go
+++ b/pkg/plugin/v2/scaffolds/internal/templates/controller/controller_suitetest.go
@@ -27,6 +27,7 @@ var _ file.Template = &SuiteTest{}
 var _ file.Inserter = &SuiteTest{}
 
 // SuiteTest scaffolds the suite_test.go file to setup the controller test
+// nolint:maligned
 type SuiteTest struct {
 	file.TemplateMixin
 	file.RepositoryMixin
@@ -36,6 +37,9 @@ type SuiteTest struct {
 
 	// CRDDirectoryRelativePath define the Path for the CRD when it is multigroup
 	CRDDirectoryRelativePath string
+
+	// WireResource defines the api resources are generated or not.
+	WireResource bool
 }
 
 // SetTemplateDefaults implements file.Template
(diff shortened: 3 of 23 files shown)


generate_testdata.sh
```diff
@@ -65,6 +65,7 @@ scaffold_test_project() {
             $kb create webhook --group crew --version v1 --kind FirstMate --conversion
             $kb create api --group crew --version v1 --kind Admiral --controller=true --resource=true --namespaced=false --make=false
             $kb create webhook --group crew --version v1 --kind Admiral --defaulting
+	    $kb create api --group crew --version v1 --kind Laker --controller=true --resource=false --make=false
```
camilamacedo86, line 68: nit space ^ alignment

pkg/plugin/v2/scaffolds/internal/templates/controller/controller.go
```diff
@@ -29,6 +29,11 @@ var _ file.Template = &Controller{}
 type Controller struct {
 	file.TemplateMixin
 	file.MultiGroupMixin
+
+	// WireResource defines the api resources are generated or not.
+	// Ordered the struct this way to save the memory by organizing struct
```
Adirio, line 34: These template structs will be one-of per execution at most, and discarded pretty fast so memory efficiency is the last thing to worry about. Place it after the 4 mixins and use a `//nolint:maligned` to avoid the linter error.

generate_testdata.sh
```diff
@@ -81,6 +82,7 @@ scaffold_test_project() {
             $kb create api --group sea-creatures --version v1beta1 --kind Kraken --controller=true --resource=true --make=false
             $kb create api --group sea-creatures --version v1beta2 --kind Leviathan --controller=true --resource=true --make=false
             $kb create api --group foo.policy --version v1 --kind HealthCheckPolicy --controller=true --resource=true --make=false
+	    $kb create api --group apps --version v1 --kind Pod --controller=true --resource=false --make=false
```
Adirio, line 85: Same as Camila said above, space alignment.

pkg/plugin/v2/scaffolds/internal/templates/controller/controller_suitetest.go
```diff
@@ -31,6 +31,11 @@ type SuiteTest struct {
 	file.TemplateMixin
 	file.RepositoryMixin
 	file.MultiGroupMixin
+
+	// WireResource defines the api resources are generated or not.
+	// Ordered the struct this way to save the memory by organizing struct
```
Adirio, line 36: Same as in the controller, code clarity way more important than memory efficiency here.

Adirio (changes requested):

generate_testdata.sh
camilamacedo86, line 68: Also see that it has make=false which means that is not calling the `make manifests` and it is not checking the change made.

Adirio (commented):
Code seems correct, but I haven't checked if the scaffolded files work properly.

I remember talking about a related topic with @DirectXMan12 some months ago. Having the `WireResource` field set to false is not enough to know if the CRD is an external one. For example, imagine the following sequence:
```
kubebuilder init
kubebuilder create api --group ship --version v1beta1 --kind Frigate --resource=true --controller=false
kubebuilder create api --group ship --version v1beta1 --kind Frigate --resource=false --controller=true
```
This case should be supported, maybe at the start we don't think we need a controller or we just want to do it in tweo steps.

camilamacedo86:
Hi  @prafull01, 

I agree with @Adirio comments. IMO we need to create a new attribute in the resources and add this info in the PROJECT file. The PROJECT file is our "database" and ought to have all info that is required for we re-create the project from the scratch. So, it means that my suggested solution would be store add the attribute controller:true/false for each GKV.

prafull01:
@camilamacedo86 IMO, we should add the new attribute to the `PROJECT` file for v3+, but we should be fixing this in v2 as well with this current approach. Currently, For v2 it will completely break in the scaffolding if user provides resources=false. 

@Adirio The case you have mentioned is supported, the only thing which will be missing is: `api import` and and For() in the controllers file, because controllers file will not be re written. For V2, user has to either manually do it later or he has to run --resources --controller as true again.. 

WDYT?

Adirio:
You meak using this as a hot-fix for `v2` and properly designing `v3` to have a better solution? I think I'm ok with that.

In that case, wherever a input is required from the user, at least a comment should be placed. For example in the `For()` call you mention early, we should probably tell the user that he needs to write that part.

camilamacedo86:
Hi @prafull01 and @Adirio, 

I removed my latest commend and I will try to do centralize all in just one. 

I tested locally the scenario added by @Adirio 

```
$kubebuilder init
$kubebuilder create api --group ship --version v1beta1 --kind Frigate --resource=false --controller=true
$kubebuilder create api --group ship --version v1beta1 --kind Frigate --resource=true --controller=false
```

And it worked with changes added here. 

IMO we could have the values of the flags info ` --resource` and `--controller` stored in the PROJECT file because then, we could easily achieve the goal, for example, to re-create the project from the scratch via the config. In my understanding, the PROJECT file ought to reflect all inputs used.

camilamacedo86 (approved):
/approve

Adirio:
P.S.: the scenario I mentioned and Camila tested does get scaffolded, but some lines are not written to the file as if we didn't have the CRD inside the project when we do.

pkg/plugin/v2/scaffolds/internal/templates/controller/controller.go
```diff
@@ -63,7 +67,9 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	{{ if .WireResource }}
```
Adirio, line 70: You are adding an extra blank line here. It may get removed later by the go formatter but it is better if we don't have it.
```suggestion
	{{ if .WireResource -}}
```

pkg/plugin/v2/scaffolds/internal/templates/controller/controller.go
```diff
@@ -63,7 +67,9 @@ import (
...
 	{{ .Resource.ImportAlias }} "{{ .Resource.Package }}"
+	{{end}}
```
Adirio, line 72: Same
```suggestion
	{{- end }}
```

pkg/plugin/v2/scaffolds/internal/templates/controller/controller.go
```diff
@@ -87,7 +93,9 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(req ctrl.Request) (ctrl.Resul
 
 func (r *{{ .Resource.Kind }}Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
+		{{ if .WireResource }}
```
Adirio, line 96:
```suggestion
		{{ if .WireResource -}}
```

pkg/plugin/v2/scaffolds/internal/templates/controller/controller.go
```diff
@@ -87,7 +93,9 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(req ctrl.Request) (ctrl.Resul
...
 		For(&{{ .Resource.ImportAlias }}.{{ .Resource.Kind }}{}).
+		{{end}}
```
Adirio, line 98:
```suggestion
		{{- else -}}
		// Uncomment the following line adding a pointer to an instance of the controlled resource as an argument
		// For().
		{{- end }}
```

Adirio (changes requested):
The four suggested changes need to be applied also to the v3 equivalent file.

pkg/plugin/v2/scaffolds/internal/templates/controller/controller.go
Adirio, line 70:
```suggestion
	{{ if .WireResource -}}
```

pkg/plugin/v2/scaffolds/internal/templates/controller/controller.go
Adirio, line 72:
```suggestion
	{{- end }}
```

pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller.go
```diff
@@ -69,7 +73,9 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	{{ if .WireResource }}
```
Adirio, line 76:
```suggestion
	{{ if .WireResource -}}
```

pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller.go
```diff
@@ -69,7 +73,9 @@ import (
...
 	{{ .Resource.ImportAlias }} "{{ .Resource.Package }}"
+	{{end}}
```
Adirio, line 78:
```suggestion
	{{- end }}
```

pkg/plugin/v3/scaffolds/internal/templates/config/controller/controller.go
```diff
@@ -104,7 +110,9 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(req ctrl.Request) (ctrl.Resul
 // SetupWithManager sets up the controller with the Manager.
 func (r *{{ .Resource.Kind }}Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
+		{{ if .WireResource -}}
 		For(&{{ .Resource.ImportAlias }}.{{ .Resource.Kind }}{}).
+		{{- end }}
```
Adirio, line 115:
```suggestion
		{{- else -}}
		// Uncomment the following line adding a pointer to an instance of the controlled resource as an argument
		// For().
		{{- end }}
```

Adirio (changes requested):
Still some changes needed

prafull01 (commented):
> Still some changes needed
My bad, apparently doing changes fastly so missed those. Fixed it. Thanks :+1:

Adirio (approved):

camilamacedo86 (commented):
/lgtm

PR #1771, opened 2020-11-02
:seedling: improve cli tests 

camilamacedo86:
**Describe**
- move the specific mock data used ONLY in the cli_test from the suite to this file
- add a new test to cover CLI with Extra commands

**Motivation**
- Maintainability
- basic coverage for CLI with extra commands scenario.

### DIFF
--- a/pkg/cli/cli_suite_test.go
+++ b/pkg/cli/cli_suite_test.go
@@ -21,87 +21,9 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-
-	"github.com/spf13/pflag"
-
-	internalconfig "sigs.k8s.io/kubebuilder/internal/config"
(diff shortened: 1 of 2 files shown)


Adirio:
Could ypu please add a motivation explaining why the moved code fits better in the new file vs the old one?

camilamacedo86:
HI @Adirio, 

> Could you please add a motivation explaining why the moved code fits better in the new file vs the old one?

- Regards `move the specific mock data used ONLY in the cli_test from the suite to this file` see that, in the suite_tests we should have only the calls and methods/const/funcs which are relevant to ALL tests. Everything that is specific to a test should be on it instead. 

- Regards `add a new test to cover CLI with Extra commands` to cover this scenario.

Adirio (approved):
/lgtm

Issue #1401, opened 2020-03-02
Build-related flags for executables built from source


Adirio:
# Description

Currently, when `kubebuilder` is built from source, `kubebuilder version` yields:
```
Version: version.Version{KubeBuilderVersion:"unknown", KubernetesVendor:"unknown", GitCommit:"$Format:%H$", BuildDate:"1970-01-01T00:00:00Z", GoOs:"unknown", GoArch:"unknown"}
```

These variables are set as part of the automated build process, but they are never stored in the release branch. they can be found in https://github.com/kubernetes-sigs/kubebuilder/blob/9552d129f7d12cc6c7072a1d323ed9f4b461d550/cmd/version/version.go#L27-L35

# Feature request

Set these flags as part of the development process instead so that, if kubebuilder is built from a release branch, even if it is from source, this fields will be set to provide useful information.

There are three types of variables:

 1. `kubeBuilderVersion`: should be set in the branch.
 2. Those that should be set in the build process:
    1. `goos`: can't be stored in the release branch as there is no one branch per target.
    2. `goarch`: same.
    3. `gitCommit`: the git commit sha1 can't be obtained before commiting the release, so it needs to be set afterwards.
    4. `buildDate`: obvious reasons.
 3. `kubernetesVendorVersion`: has no meaning when building from source.

/kind feature
/kind documentation

Adirio:
The last four variables could be set as part of the `make build` target.

mengqiy:
`kubernetesVendorVersion` indicates the version of the testing tools (kube-apiserver and kubectl) shipped in a release.
It's not useful when building from source.

Adirio:
@estroz [suggested some commands](https://github.com/kubernetes-sigs/kubebuilder/pull/1773#discussion_r520910104) to fetch this data and I would like to expand on them:

- `kubeBuilderVersion`: `git describe --tags --dirty --broken`
- `kubernetesVendorVersion`: I'm not sure if this is currently relevant.
- `goos`: `go env GOOS`
- `goarch`: `go env GOARCH`
- `gitCommit`: `git rev-parse HEAD`
- `buildDate`: `date -u +'%Y-%m-%dT%H:%M:%SZ'` <-- This is not Windows compatible :(

P.S.: I removed the `--always` flag from his suggested `git describe` command as we already include the commit, so no reason to default to the commit if a tag couldn't be found. I also added the `--broken` flag which should hopefully never happen but better safe than sorry.

PR #1773, opened 2020-11-03
⚠Provide version information as a CLI option

Adirio:
# Description
Provide a `WithVersion` CLI `Option` that allows to provide a version string instead of having to create a `version` command and provide it as an extra command. When a version is provided this way, the CLI will create a version command automatically that prints this version string.

# Changes
- ✨ New argument type for `cli.New`: `WithVersion`. Creating a command and providing it with `WithExtraCommands` should still work with the previous behavior, so this is not a breaking change, but frameworks using `kubebuilder` should migrate to the new approach.
- ⚠️ Removed `Version` and `NewCmd` from package `cmd/version`. This package is not part of the library itself, but as it is exported, it may be used by third parties.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -149,8 +149,8 @@ function build_kb {
   if [ "$INJECT_KB_VERSION" = "unknown" ]; then
     opts=""
   else
-    # TODO: what does this thing do.
-    opts=-ldflags "-X sigs.k8s.io/kubebuilder/v2/cmd/version.kubeBuilderVersion=$INJECT_KB_VERSION"
+    # Injects the version into the cmd/version.go file
+    opts=-ldflags "-X sigs.k8s.io/kubebuilder/v2/cmd.kubeBuilderVersion=$INJECT_KB_VERSION"
   fi
 
   GO111MODULE=on go build $opts -o $tmp_root/kubebuilder/bin/kubebuilder ./cmd
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -57,6 +57,9 @@ type Option func(*cli) error
 type cli struct {
 	// Root command name. Can be injected downstream.
 	commandName string
+	// CLI version string
+	version string
+
 	// Default project version. Used in CLI flag setup.
 	defaultProjectVersion string
 	// Project version to scaffold.
@@ -114,6 +117,14 @@ func (c cli) Run() error {
 func WithCommandName(name string) Option {
 	return func(c *cli) error {
 		c.commandName = name
+		return nil
+	}
+}
+
+// WithVersion is an Option that defines the version string of the cli.
+func WithVersion(version string) Option {
+	return func(c *cli) error {
+		c.version = version
 		return nil
 	}
 }
@@ -363,8 +374,15 @@ func (c cli) buildRootCmd() *cobra.Command {
 
 	// kubebuilder edit
 	rootCmd.AddCommand(c.newEditCmd())
+
 	// kubebuilder init
 	rootCmd.AddCommand(c.newInitCmd())
+
+	// kubebuilder version
+	// Only add version if a version string was provided
+	if c.version != "" {
+		rootCmd.AddCommand(c.newVersionCmd())
+	}
 
 	return rootCmd
 }
@@ -416,7 +434,6 @@ After the scaffold is written, api will run make on the project.
   make run
 `,
 			c.commandName, c.commandName),
-
 		Run: func(cmd *cobra.Command, args []string) {
 			if err := cmd.Help(); err != nil {
 				log.Fatal(err)
--- a/pkg/cli/version.go
+++ b/pkg/cli/version.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2017 The Kubernetes Authors.
+Copyright 2020 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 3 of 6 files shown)


Adirio:
@camilamacedo86 This PR goes in the same direction as #1691, moving another command out of the extra commands. I think that the API for frameworks with the new options is clearer, they only have to provide a string now instead of a full COBRA command, and it also has the added benefot of including `kubebuilder --version` and `kubebuilder -v` in addition to `kubebuilder version`. WDYT?

Adirio:
A note in coverage: this PR moves part of the remaining code from `cmd` to `pkg`. As `cmd` was not being checked by coveralls, it says that the coverage has decreased, but in relaity we just have included in the covered package part of the code that wasn't being covered by coveralls.

cmd/main.go
```diff
@@ -27,6 +27,7 @@ import (
 
 func main() {
 	c, err := cli.New(
+		cli.WithVersion(version.String()),
```
camilamacedo86, line 30: what is the use case for it? Why someone would like to run for example `mybin version` and get a result `kubeBuilderVersion` instead of the version of the bin?

Adirio: `mybin version` is contructed by a different `cli.New` call which would have the string they want.

```
c, err := cli.New(
	cli.WithCommandName("mybin"),
	cli.WithVersion("v0"),
	...
)
```

camilamacedo86: but the output will no be `kubeBuilderVersion:v0`? could you please add here its output for we check?

Adirio: `version.String()` outputs all the `kubebuilder` info, but `mybin` would not use `version.String()`

camilamacedo86 (approved):
I am ok with. 👍 
/approve 

@estroz do you see any reason for we do not move forward here? WDYT?

cmd/version/version.go
```diff
@@ -34,8 +32,8 @@ var (
 	buildDate = "1970-01-01T00:00:00Z" // build date in ISO8601 format, output of $(date -u +'%Y-%m-%dT%H:%M:%SZ')
 )
 
-// Version contains all the information related to the CLI version
-type Version struct {
+// version contains all the information related to the CLI version
+type version struct {
 	KubeBuilderVersion string `json:"kubeBuilderVersion"`
```
camilamacedo86, line 37:
```suggestion
	ToolVersion string `json:"toolVersion"`
```

camilamacedo86: Just a nit since it can be used by another cli as well.

Adirio: No, that is only for kubebuilder, and it needs to stay as is because it is filled by `ldflags`

pkg/cli/cli.go
```diff
@@ -424,7 +442,7 @@ After the scaffold is written, api will run make on the project.
   make run
 `,
 			c.commandName, c.commandName),
-
+		Version: c.version,
```
estroz, line 445: According to the docs, this is only necessary if the command doesn't define a `version` command.

pkg/cli/version.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func (c *cli) newVersionCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:     "version",
+		Short:   fmt.Sprintf("Print the %s version", c.commandName),
+		Long:    fmt.Sprintf("Print the %s version", c.commandName),
+		Example: fmt.Sprintf("%s version", c.commandName),
+		RunE: func(_ *cobra.Command, _ []string) error {
+			fmt.Print(c.version)
```
estroz, line 32: Nit: I'm not a fan of version printing without a trailing newline
```suggestion
			fmt.Println(c.version)
```

pkg/cli/cli.go
Adirio: According to [the docs](https://godoc.org/github.com/spf13/cobra#Command) this has to do with `-v` and `--version` flags, not with `version` command:
>// Version defines the version for this command. If this value is non-empty and the command does not
>// define a "version" flag, a "version" boolean flag will be added to the command and, if specified,
>// will print content of the "Version" variable. A shorthand "v" flag will also be added if the
>// command does not define one.

estroz: Ah I read that wrong. I don't think we want to expose a `--version` _and_ a `version` command though.

Adirio: I don't see any damage done by being able to do `kubebuilder version`, `kubebuilder --version` and `kubebuilder -v`. I think it actually improves UX as it doesn't have to know if it is a command or a flag. `kubebuilder help` and `kubebuilder --help` have the same behavior, right?

Adirio:
@estroz can we merge this? The only remaining comment is that you wouldn't add `kubebuilder --version` in addition to `kubebuilder version` but I don't see any harm indoing so. Do you have any reason to say so?

estroz:
@Adirio the only hesitation I have is that most k8s tools like `kubectl` only have the `version` command, and I'd prefer to adhere to that standard in `kubebuilder`. Also this reserves `-v` for verbose mode, which is also common.

Adirio:
@estroz if we finally set a `-v` flag for verbose setting `Version` won't create that flag.

I will remove the flag integration then.

common.sh
```diff
@@ -149,8 +149,8 @@ function build_kb {
   if [ "$INJECT_KB_VERSION" = "unknown" ]; then
     opts=""
   else
-    # TODO: what does this thing do.
-    opts=-ldflags "-X sigs.k8s.io/kubebuilder/v2/cmd/version.kubeBuilderVersion=$INJECT_KB_VERSION"
+    # Injects the version into the cmd/version.go file
+    opts=-ldflags "-X sigs.k8s.io/kubebuilder/v2/cmd.kubeBuilderVersion=$INJECT_KB_VERSION"
```
estroz, line 153: Wdyt about adding adding something like this to `make build`, ex.
```make
BUILD_VERSION=$(shell git descripe --dirty --tags --always)
BUILD_COMMIT=$(shell git rev-parse HEAD)
LD_FLAGS=-ldflags "-X sigs.k8s.io/kubebuilder/v2/cmd.kubeBuilderVersion=$(BUILD_VERSION) -X sigs.k8s.io/kubebuilder/v2/cmd.gitCommit=$(BUILD_COMMIT)"
build: ## Build the project locally
        go build $(LD_FLAGS) -o bin/kubebuilder ./cmd
```
perhaps in a follow-up?

estroz (approved):
/lgtm

common.sh
Adirio: I tracked this change in an Issue some time ago: #1401

PR #1775, opened 2020-11-03
✨Suggestions for create and alpha commands

Adirio:
# Description
Added suggestions for some commands so that when the CLI user writes a similar term it gets useful help information telling him the correct subcommand name

- `create`: `new`
- `alpha`: `experimental`

### DIFF
--- a/pkg/cli/alpha.go
+++ b/pkg/cli/alpha.go
@@ -24,9 +24,10 @@ import (
 
 func (c *cli) newAlphaCmd() *cobra.Command {
 	return &cobra.Command{
-		Use:   "alpha",
-		Short: "Expose commands which are in experimental or early stages of development",
-		Long:  `Command group for commands which are either experimental or in early stages of development`,
+		Use:        "alpha",
+		SuggestFor: []string{"experimental"},
+		Short:      "Expose commands which are in experimental or early stages of development",
+		Long:       `Command group for commands which are either experimental or in early stages of development`,
 		Example: fmt.Sprintf(`
 # scaffolds webhook server
 %s alpha webhook <params>`,
--- a/pkg/cli/create.go
+++ b/pkg/cli/create.go
@@ -22,8 +22,9 @@ import (
 
 func (c *cli) newCreateCmd() *cobra.Command {
 	return &cobra.Command{
-		Use:   "create",
-		Short: "Scaffold a Kubernetes API or webhook",
-		Long:  `Scaffold a Kubernetes API or webhook.`,
+		Use:        "create",
+		SuggestFor: []string{"new"},
+		Short:      "Scaffold a Kubernetes API or webhook",
+		Long:       `Scaffold a Kubernetes API or webhook.`,
 	}
 }


pkg/cli/alpha.go
```diff
@@ -24,9 +24,10 @@ import (
 
 func (c *cli) newAlphaCmd() *cobra.Command {
 	return &cobra.Command{
-		Use:   "alpha",
-		Short: "Expose commands which are in experimental or early stages of development",
-		Long:  `Command group for commands which are either experimental or in early stages of development`,
```
camilamacedo86, line 29: HI @Adirio, 

What command we would have here currently? Also, the currently kubebuilder design is to work with plugins which are extensible. So, in this case, would be expected plugins with alpha version instead of new commands. In this way, could you please provide an usecase for it? And then, IMO we only should add this if we have at least one command to be exposed as alpha, otherwise, it might not be a good user experience.

Adirio: The alpha command is a command group where experimental or new commands are added before they get added to the stable commands. `webhook` used to be one of the `alpha` commands in `v1`, but it was promoted to stable in `v2`.

However, if no subcommand is added to alpha, alpha is not added to kubebuilder:

https://github.com/kubernetes-sigs/kubebuilder/blob/a4ff1b68ee81455f8ece4aad310f813580767983/pkg/cli/cli.go#L340-L346

So don't worry, if we have no subcommand for `alpha`, the built cli won't have `alpha` as a command group.

camilamacedo86: Could you give examples of alpha commands (usecases) which justify we add it?
IMO we would like to promote the plugin usage instead of that.

Adirio: I'm not adding it, it is already there. And a plugin being an alpha version is completely different to a command of that plugin being alpha. Imagine for example that you want to include a `migrate` command that performrs the change to migrate from `v2` to `v3`. While you are developing this command and is still not completely tested you leave it under the `alpha` group. The same way `webhook` used to be under the `alpha` group in version 1. This doesn't mean that the whole v1 is alpha, just that command.

camilamacedo86: In the case of the Migrate command, I still think that it should be a new Plugin such as MigratePlugin which for example if be added to v3-alpha plugin would be implemented to migrate v2 to v3 and in v4 should be able to do that for v2/v3 as well for example. 

So, why we need to keep the alpha subcommand at all? Could we not just remove it?

camilamacedo86: In the case of the Migrate command, I still think that it should be a new Plugin such as MigratePlugin which for example if be added to v3-alpha plugin would be implemented to migrate v2 to v3 and in v4 should be able to do that for v2/v3 as well for example. So, why we need to keep the alpha subcommand at all? Could we not just remove it?

I am trying to check here if we really should keep it or only remove that to keep a better maintainability
I mean, if it is not used at all, why we need to keep it?

Adirio: That could be done in another PR, this just adds the suggestions for two commands so that when a user writes the command as `kubebuilder new api` the help message suggests to him to use `kubebuilder create api`, as `new` is a usual keyword in other cli tools-

camilamacedo86: Since it shows out of the scope of this PR, I think we can discuss this point in : https://github.com/kubernetes-sigs/kubebuilder/pull/1797

camilamacedo86 (approved):
/approve

camilamacedo86:
Too small for wait for 2 reviews. Let's get it merged. 

/lgtm

PR #1776, opened 2020-11-03
✨ Provide a cli option to enable and disable completion command

Adirio:
# Description
Provide a `WithCompletion` CLI `Option` that enables adds a compeltion generation command instead of having to create it and provide it as an extra command.

# Changes
- ✨ New argument type for `cli.New`: `WithCompletion`. Creating a command and providing it with `WithExtraCommands` should still work with the previous behavior, so this is not a breaking change, but frameworks using `kubebuilder` should migrate to the new approach.


Adirio:
@camilamacedo86 This PR goes in the same direction as #1691 and #1773, moving another command out of the extra commands. I think that the API for frameworks with the new options is clearer, they only have to provide an argument instead of a full COBRA command. WDYT?

cmd/main.go
```diff
@@ -35,9 +36,9 @@ func main() {
 			&pluginv2.Plugin{},
 		),
 		cli.WithExtraCommands(
-			newCompletionCmd(),
 			version.NewCmd(),
 		),
+		cli.WithCompletion,
```
camilamacedo86, line 41: IMO the Completion also ought to be a plugin.

Adirio: This completion command is valid for every plugin combination you use. It is not specific to `go.kubebuilder.io` plugin. There is no need to provide a plugin that makes completion because this command will work for the plugins you configure.

Adirio:
A note in coverage: this PR moves part of the remaining code from `cmd` to `pkg`. As `cmd` was not being checked by coveralls, it says that the coverage has decreased, but in relaity we just have included in the covered package part of the code that wasn't being covered by coveralls.

camilamacedo86 (approved):
It is only missing tests.
Otherwise, 
/approve

Looking for your tests for we get it merged.

pkg/cli/completion.go
```diff
@@ -0,0 +1,82 @@
...
+		*/
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			switch args[0] {
+			case "bash":
+				err = cmd.Root().GenBashCompletion(os.Stdout)
+			case "zsh":
+				err = cmd.Root().GenZshCompletion(os.Stdout)
+			// TODO (requires spf13.cobra update): uncomment when adding fish
+			//case "fish":
+			//	err = cmd.Root().GenFishCompletion(os.Stdout, true)
```
camilamacedo86, line 75:
```suggestion

```

docs/book/src/reference/completion.md
```diff
@@ -1,5 +1,5 @@
 # Enabling shell autocompletion
-The Kubebuilder completion script for Bash can be generated with the command `kubebuilder completion bash` as the Kubebuilder completion script for Zsh can be generated with the command `kubebuilder completion zsh`. 
+The Kubebuilder completion script can be generated with the command `kubebuilder completion [bash|zsh|fish|powershell]`.
```
camilamacedo86, line 2:
```suggestion
The Kubebuilder completion script can be generated with the command `kubebuilder completion [bash|zsh|powershell]`. 
```

camilamacedo86: The fix was not added.

pkg/cli/completion.go
```diff
@@ -0,0 +1,82 @@
...
+		/* TODO (requires spf13.cobra update): add to Long when adding fish
+		   `
+		   Fish:
+
+		   $ %[1]s completion fish | source
+
+		   # To load completions for each session, execute once:
+		   $ %[1]s completion fish > ~/.config/fish/completions/%[1]s.fish
+		   `
+		*/
```
camilamacedo86, line 66: Let's at least make more clear this TODO.

camilamacedo86 (changes requested):
Looking for the small nit fixes for we get it merged. 
See the comments

pkg/cli/completion.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+func (c *cli) newCompletionCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:                   "completion [bash|zsh|powershell]", // TODO (requires spf13.cobra update): add fish
```
camilamacedo86, line 28: It usage/description is not so good as before. See: 

**master**

```
 $ kubebuilder completion 
Output shell completion code for the specified shell (bash or zsh). 
The shell code must be evaluated to provide interactive completion of kubebuilder commands.
This can be done by sourcing ~/.bash_profile or ~/.bashrc.
Detailed instructions on how to do this are available at docs/book/src/reference/completion.md

Usage:
  kubebuilder completion [command]

Examples:
To load all completions run:
$ . <(kubebuilder completion)
To configure your shell to load completions for each session add to your .bashrc:
$ echo -e "\n. <(kubebuilder completion)" >> ~/.bashrc


Available Commands:
  bash        Generate bash completions
  zsh         Generate zsh completions

Flags:
  -h, --help   help for completion

Use "kubebuilder completion [command] --help" for more information about a command.
```

Now: 

```
 $ kubebuilder completion 
Error: accepts 1 arg(s), received 0
Usage:
  kubebuilder completion [bash|zsh|powershell]

Flags:
  -h, --help   help for completion

2020/11/04 07:50:14 accepts 1 arg(s), received 0
```

Could you please keep the same behaviour?

pkg/cli/completion.go
```diff
@@ -0,0 +1,82 @@
...
+
+Bash:
+
+$ source <(%[1]s completion bash)
+
+# To load completions for each session, execute once:
+Linux:
+  $ %[1]s completion bash > /etc/bash_completion.d/%[1]s
+MacOS:
+  $ %[1]s completion bash > /usr/local/etc/bash_completion.d/%[1]s
```
camilamacedo86, line 43: It should be in the usage/examples instead of it.

pkg/cli/completion.go
```diff
@@ -0,0 +1,87 @@
...
+		/* NOTE(fish): add the following description to Long
+		   `
+		   Fish:
+
+		   $ %[1]s completion fish | source
+
+		   # To load completions for each session, execute once:
+		   $ %[1]s completion fish > ~/.config/fish/completions/%[1]s.fish
+		   `
+		*/
```
camilamacedo86, line 71: IMO we should remove this comments and instead of that add : 

// todo: when the cobra version be upgraded to X add the fish option as described in : URL

pkg/cli/completion.go
Adirio: Moved to each of the usage examples

pkg/cli/completion.go
Adirio: Changed back to subcommands instead of flags for each shell so that iot provides better help message if the shell is missing

camilamacedo86 (approved):
/lgtm 
/approve 

Great work 👍

pkg/cli/completion.go
```diff
@@ -0,0 +1,108 @@
...
+		Short: "Load completions for the specified shell",
+		Long: fmt.Sprintf(`Output shell completion code for the specified shell.
+The shell code must be evaluated to provide interactive completion of %[1]s commands.
+Detailed instructions on how to do this for each shell are provided in their own commands.
+`, c.commandName),
+	}
+	cmd.AddCommand(c.newBashCmd())
+	cmd.AddCommand(c.newZshCmd())
+	// cmd.AddCommand(c.newFishCmd()) // TODO: uncomment when adding fish completion
+	cmd.AddCommand(c.newPowerShellCmd())
```
camilamacedo86, line 106: 👍 
/lgtm

Adirio:
Rebased and added the tests for this option to improve coverage.

camilamacedo86 (commented):
/lgtm

PR #1777, opened 2020-11-03
🌱Template folder cleanup

Adirio:
# Description
- Structure the template files in the same way as the scaffolded files
- Provide a more descriptive name and comment for each template
- Fix references to interfaces being implemented


Adirio:
@camilamacedo86 This is the PR I mentioned you I was going to make in order to make the templates folder have the same structure as the scaffolded one. I did it for both v2 and v3 as it is a non-breaking change. I also updates some file and struct names, provided a more descriptive comment for each template and fixed the references to the interfaces they are implemented, as `input` used to be the package where those interfaces were placed, but now it is called `file`.

camilamacedo86 (approved):
It shows great 🥇 
/approve

Adirio:
Rebased, once tests are passed again it will only need a `lgtm` label.

pkg/plugin/v3/scaffolds/internal/templates/config/prometheus/monitor.go
```diff
@@ -22,15 +22,15 @@ import (
 	"sigs.k8s.io/kubebuilder/pkg/model/file"
 )
 
-var _ file.Template = &ServiceMonitor{}
+var _ file.Template = &Monitor{}
 
-// ServiceMonitor scaffolds an issuer CR and a certificate CR
-type ServiceMonitor struct {
+// Monitor scaffolds a file that defines the prometheus service monitor
+type Monitor struct {
```
estroz, line 28: This name can probably stay as-is

pkg/plugin/v3/scaffolds/internal/templates/config/rbac/role_binding.go
```diff
@@ -22,15 +22,15 @@ import (
 	"sigs.k8s.io/kubebuilder/pkg/model/file"
 )
 
-var _ file.Template = &ManagerRoleBinding{}
+var _ file.Template = &RoleBinding{}
 
-// ManagerRoleBinding scaffolds the config/rbac/role_binding.yaml file
-type ManagerRoleBinding struct {
+// RoleBinding scaffolds a file that defines the role binding for the manager
+type RoleBinding struct {
```
estroz, line 28: The purpose of this scaffold is clearer as `ManagerRoleBinding` imo

pkg/plugin/v3/scaffolds/internal/templates/config/prometheus/monitor.go
camilamacedo86: +1 The name of the resource scaffold is ServiceMonitor so shows better we keep the same name. IHMO more intuitive.

pkg/plugin/v3/scaffolds/internal/templates/config/rbac/role_binding.go
camilamacedo86: have we any other role_binding or just this one used by the manager? Since I do not think that we have another then, I would be ok with the change in the name. It is up to you folks.

pkg/plugin/v3/scaffolds/internal/templates/config/prometheus/monitor.go
Adirio: I changed to monitor to reflect the file name, which is `monitor.yaml` IIRC.

pkg/plugin/v3/scaffolds/internal/templates/config/rbac/role_binding.go
Adirio: This is the main role binding. We have one for the auth proxy wich is called `AuthProxyRoleBinding`. Again the changed name reflects the file name, IIRC.

pkg/plugin/v3/scaffolds/internal/templates/config/prometheus/monitor.go
estroz: I guess if the scaffold name reflects the final file name I'm ok with this (and the other) change.

Adirio:
Tide said it had a merge conflict, so rebased. Still only needing `lgtm`.

PR #1782, opened 2020-11-04
🌱 Move internal packages

Adirio:
# Description
Move the following packages:
- `internal/config` -> `pkg/cli/internal/config`
- `internal/cmdutil` -> `pkg/plugin/internal/cmdutil`

Also copied `Scaffolder` from `pkg/plugin/scaffold` to `pkg/plugin/internal/cmdutil` as that interface was not designed to be exported. However, as it was exported, full removal requires to be a bit more careful and allow to remove it first from the thrid party projects using `kubebuilder`.

# Motivation
These two packages used to be required by `cmd`, and therefore had to be placed in a root-level internal package. After plugin phase 1, `cmd` no longer requires these packages, as the related code was moved to `pkg/cli` and `pkg/plugin`. Therefore, moving the mentioned packages closer to the place they are used, and, more importantly, eliminating the root `internal` package can be done now.


Adirio:
A note in coverage: this PR moves the remaining code from `internal` to `pkg`. As `internal` was not being checked by coveralls, it says that the coverage has decreased, but in reality we just have included in the covered package part of the code that wasn't being covered by coveralls.

Adirio:
It seems that by being outside of `pkg` `internal/config` tests were not being executed, so basically we had some tests that were not being evaluated.

Once I moved them, the tests are being evaluated and some of them were broken. I fixed them.

camilamacedo86:
Hi @Adirio, 

Let's get merged first this one https://github.com/kubernetes-sigs/kubebuilder/pull/1764 to ensure the tests.

camilamacedo86 (approved):
I agree with these changes. 
Good catcher 👍 

/lgtm

camilamacedo86 (commented):
/hold

pkg/plugin/v2/api.go
```diff
@@ -158,7 +157,7 @@ func (p *createAPIPlugin) Validate() error {
 	return nil
 }
 
-func (p *createAPIPlugin) GetScaffolder() (scaffold.Scaffolder, error) {
+func (p *createAPIPlugin) GetScaffolder() (cmdutil.Scaffolder, error) {
```
camilamacedo86, line 160: @Adirio, 

It might cause problems for consumers. See:  `"sigs.k8s.io/kubebuilder/pkg/plugin/scaffold"` that is not internal and you are replacing for `"sigs.k8s.io/kubebuilder/pkg/plugin/internal/cmdutil"`. Could you please revert it?

camilamacedo86: /lgtm cancel
/approve cancel

Adirio: `Scaffolder` was meant to be private, but it coulnd't be made private because it was being used from the `internal/cmdutil` package and from inside `pkg/plugin` (I was the one creating this interface). Now that we moved all into the same folder, we can make it private as it was intended.

The whole usage of this interface is private so no third-party library should be importing it.

Adirio:
The dummy `Scaffolder` interface has been left exported as it may be consumed in third-party projects. However, this interface is not being used for anything and thus should be removed from the public API in a following PR.

/cc @camilamacedo86

camilamacedo86:
Hi @Adirio, 

> The dummy Scaffolder interface has been left exported as it may be consumed in third-party projects. However, this interface is not being used for anything and thus should be removed from the public API in a following PR.

The dummy Scaffolder interface that is exported is no longer used because you added it in the cmdutil. I am ok with we remove the exported one, however, then it will be a breaking change which means that the title pr requires to be with the breaking change mojo and we need highlight its motivation in the description. 

For me we can merge this one +1, I liked the idea over removing the internal dir from the root. And we can do the breaking change in follow up. @estroz are you ok with it as well?

Adirio:
Description updated

estroz:
Test coverage change seems to be a fluke

/approve

Adirio:
Coverage was not considering the directory `internal` so when moving them inside `pkg`, which is covered, it considers it has more code with no tests, but the reality is we have the same code.

`lgtm` label required.

Adirio:
Rebased without changes, @camilamacedo86 could you add the `lgtm` label again please?

Adirio:
@droot @mengqiy @DirectXMan12 coverage is blocking this as we moved to the `pkg` package code that was previously outside of the coverage directories. Is there anyway we can skip this test?

camilamacedo86:
HI @Adirio, 

The `coverage/coveralls ` will not block your PR. It is used for we ONLY be able to follow up it and help us to keep a better coverage. In this scenario, it is totally fine we get it merged with `(-2.6%) to 71.216%`

camilamacedo86:
/lgtm cancel

You need to squash the commits for we get it merged.

Adirio:
Squashed

camilamacedo86 (approved):
/lgtm

camilamacedo86:
All tests here passed. However, it will not get merged by the bot because of the coveralls check which is blocking it.
The coveralls decreased because it was not checking all files. In this way, I will force the merge.

PR #1783, opened 2020-11-04
:book: add note about officially versions supported

camilamacedo86:
Cheerry-pick 8acdc2872d1704cd9d0ff48a1b55fdc0f7a51d1c

### DIFF
--- a/testdata/project-v2-addon/go.mod
+++ b/testdata/project-v2-addon/go.mod
@@ -9,5 +9,5 @@ require (
 	k8s.io/apimachinery v0.18.4
 	k8s.io/client-go v0.18.4
 	sigs.k8s.io/controller-runtime v0.6.0
-	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200816135617-dbfe418e405f
+	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20201008142216-045672be83e7
 )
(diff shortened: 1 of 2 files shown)


Adirio:
Changes to the book will be performed when a new version of kubebuilder is released. Is the information you want to add here accurate to the release 2.3.1 of `kubebuilder`? And if that is the case, why do we need to merge this into the documentation inmediately instead of waiting until a new release like the normal book developing cycle is designed?

estroz:
I don't think this applies to the released v2 book as @Adirio has described.

testdata/project-v2-addon/go.mod
```diff
@@ -9,5 +9,5 @@ require (
 	k8s.io/apimachinery v0.18.4
 	k8s.io/client-go v0.18.4
 	sigs.k8s.io/controller-runtime v0.6.0
-	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200816135617-dbfe418e405f
+	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20201008142216-045672be83e7
```
camilamacedo86, line 12: For master, we pinned the version now. However, for the book-v2 it still getting the latest. In this way, for we be able to update the docs until the next release we need to fix it.

camilamacedo86:
Hi @Adirio, 

Following the answers of your comments. 

> Changes to the book will be performed when a new version of kubebuilder is released. 

Yes, after the release we have the post-action task to rebase master on this branch to update all. However, we are also allowed to cherry-picks when we would like to push the doc change to production. 

> Is the information you want to add here accurate to the release 2.3.1 of kubebuilder? 

Yes, it is valid for 2.3.1 as well. 

> And if that is the case, why do we need to merge this into the documentation inmediately instead of waiting until a new release like the normal book developing cycle is designed?

Because this note avoids common issues that have been raised a lot since the latest version of kustomize has a breaking change. So, it is important to highlight what are the versions recommended/suggested/tested. E.g #1655

Adirio:
If the new version of Kustomize has breaking changes and the topic has atracted recent attention, I guess we can cherry-pick it.

/lgtm

Issue #1634, opened 2020-08-11
Negative Defaults Values Not Supported


donmstewart:
**What broke?**

Schema generation

**What did you expect to happen?  What do you think went wrong?** 

For int/int32/int64 have the ability to set negative default values using tags 

```
// Set the maximum number of threads (both workers and masters) to be passed
// to TBB on initialization.  Generally speaking,
// 'max_tbb_threads_per_rank' - "1" TBB workers will be created.  
//
// Use "-1" for no limit.
//
// +optional
// +kubebuilder:default:=-1
MaxTbbThreadsPerRank int `json:"maxTbbThreadsPerRank,omitempty"`
```

Unfortunately it errors out with the following:-

` Invalid value: "string":  in body must be of type integer: "string"`

Removing the default and the code generates as an integer.

Tried using `kubebuilder:validation:Type:=int` and with `kubebuilder:validation:Format:=int32` as per https://swagger.io/docs/specification/data-models/data-types/

but unfortunately that does not seem to work either.

What versions of software are you using?  Specifically, the following are often useful:

*** go version** 

1.14.2

* kubebuilder version 

2.3.1

and scaffolding version (check your `PROJECT` file)
*** controller-runtime version** 

v0.5.0

* controller-tools version

v0.2.5

*** Kubernetes & kubectl versions** 

1.18.6, 1.18.5

/kind bug

camilamacedo86:
I did not try to test it yet, however, see in the [kubebuilder docs](https://catalog.redhat.com/software/containers/openshift4/ose-helm-operator) that the marker shows be: 

```go
// +kubebuilder:validation:Optional 
// +kubebuilder:validation:Type="<type>"
// +kubebuilder:validation:default=<value>
```

Which is NOT equals that markers used by you: 

```
// +optional
// +kubebuilder:default:=-1
```

So, could you please check the Kubebuilder docs and try to follow up it? Also, if the problem still could you please check it with the master branch which will use an upper version of the controller-tools?

donmstewart:
Thanks for the follow up it's appreciated. I tried correcting the error you pointed out and alas still no go.

gabbifish:
Hi @donmstewart! I haven't tried to reproduce this issue yet either, but just to make sure I know what behavior you're running into, are you still seeing
```
Invalid value: "string": in body must be of type integer: "string"
```
even after you've modified your annotation to use: 
```
// +kubebuilder:default=-1
```

Can you also try 
```
[...]

PR #1760, opened 2020-10-30
:sparkles: Cleanup in the controller and webhook suites scaffolding (go/v3-alpha)

prafull01:
**Description of the change**

Improved the scaffolding for the webhook and controller suites. 
Removed the Done as we are not running the asynchronous test suites.

**Motivation of the change**

Fixes: #1733

PR #1756, opened 2020-10-29
:seedling: test: add `go/v3-alpha` e2e tests

estroz:
This PR adds e2e tests for go/v3-alpha and refactors test code such that it is easier to add other plugin tests going forward.

- test/e2e/utils: add util to remove tool binaries from path
- test/e2e/v3: refactor generation code so all plugins can be tested
- *: clean up scripts

Pulls refactoring/cleanup from #1644

PR #1787, opened 2020-11-04
:warning: upgrade controller-tool dep from 0.3.0 to 0.4.0 (go/v3-alpha only)

camilamacedo86:
**Description**
- upgrade controller-tool dep from 0.3.0 to 0.4.0 (go/v3-alpha only)

NOTE: The controller-tool no generates  CRD version to v1 by default (#468) and add admissionregisteration/v1 support for webhook (#469) which means that are breaking changes for v2+. Then, because of this, the fix can only be addressed to v3+ plugin. 

**Motivation**
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/1634

### DIFF
--- a/pkg/plugin/v3/scaffolds/init.go
+++ b/pkg/plugin/v3/scaffolds/init.go
@@ -39,7 +39,7 @@ const (
 	// ControllerRuntimeVersion is the kubernetes-sigs/controller-runtime version to be used in the project
 	ControllerRuntimeVersion = "v0.6.3"
 	// ControllerToolsVersion is the kubernetes-sigs/controller-tools version to be used in the project
-	ControllerToolsVersion = "v0.3.0"
+	ControllerToolsVersion = "v0.4.0"
 	// KustomizeVersion is the kubernetes-sigs/kustomize version to be used in the project
 	KustomizeVersion = "v3.5.4"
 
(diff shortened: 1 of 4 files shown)


testdata/project-v3-multigroup/controllers/apps/suite_test.go
```diff
@@ -47,30 +47,28 @@ func TestAPIs(t *testing.T) {
...
+	Expect(k8sClient).NotTo(BeNil())
 
-	close(done)
 }, 60)
 
 var _ = AfterSuite(func() {
 	By("tearing down the test environment")
 	err := testEnv.Stop()
-	Expect(err).ToNot(HaveOccurred())
+	Expect(err).NotTo(HaveOccurred())
```
Adirio, line 73: Does this belong to #1760?

camilamacedo86: See: https://github.com/kubernetes-sigs/kubebuilder/pull/1789

estroz:
This PR broke go/v3-alpha silently because testdata generation can only use one tool version per script invocation, i.e. all testdata is generated with controller-gen v0.3.0 right now. This blocks #1756, so we'll have to go ahead and merge its parent PR #1644 to fix go/v3-alpha.

PR #1789, opened 2020-11-04
:seedling: sync testdata

camilamacedo86:
**Description**

The testdata generated is checked only in the Travis and not in prow. In this way, a pr was merged without sync/updated the testdata. Doing it now.


Adirio (approved):

Adirio:
This PR fixes the testdata folder so that the rest stop being blocked.

/lgtm

Issue #1746, opened 2020-10-27
deps: bump to controller-runtime v0.7.0


estroz:
This issue tracks the v0.7.0 controller-runtime bump, which has breaking changes. Also bumps deps to k8s 
v1.19.

- [x] Bump kubebuilder-declarative-pattern
	- https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/127, blocking
	- #1804
- [x] Bump controller-runtime to an alpha release
	- #1804
- [ ] Bump controller-runtime to the stable release

/kind feature

camilamacedo86:
@estroz IMO we might need to wait for the stable release to be able to bump the dep.

estroz:
From https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/127#discussion_r513057434 (applies here too):
>Is there a good reason to wait? I don't expect any changes to go in aside from bugfixes at this point, although it is technically an alpha release. Unless there is a plan to release this project soon, bumping to alpha now then bumping to stable before a release makes sense to me since the alpha bump will only ever be on master. This also helps test out the controller-runtime v0.7.0 release.

Also we are currently pinning to unreleased code with [`sigs.k8s.io/kubebuilder-declarative-pattern`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/v3/api.go#L44), so I don't see why we can't do something similar for controller-runtime, but with a tagged release.

estroz:
@pwittrock @DirectXMan12 @droot how do you feel about pinning to commits/alpha|beta releases in master in general?

estroz:
I'll also mention that v0.6.3 has a logging bug in it (fixed by https://github.com/kubernetes-sigs/controller-runtime/pull/972) which breaks kubebuilder e2e tests (see [this comment](https://github.com/kubernetes-sigs/kubebuilder/pull/1644#issuecomment-716887694)). Not sure why the tests passed on #1721.

We can either bump to v0.7.0 or get a patch release v0.6.4 going, then bump here. I suggest going straight to v0.7.0.

camilamacedo86:
Hi @estroz, 

Following some comments. 

> Also we are currently pinning to unreleased code with sigs.k8s.io/kubebuilder-declarative-pattern 
This project is used only for the addon and it does not work with release tags

[...]

PR #1790, opened 2020-11-05
⚠ (go/v3-alpha) Adding support for scaffolding with a versioned ComponentConfig

### LINKED ISSUE #722: Implement ComponentConfig by default & stop using (most) flags
/kind feature
related KEPs here: https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/wgs

cc @DirectXMan12 as per contributor summit discussion

christopherhein:
This adds the ability to scaffold a `kubebuilder` v3 plugin project with a versioned config file (componentconfig) this was first implemented in Controller Runtime by allowing `.AndFrom[OrDie]()` to the `ctrl.Manager` examples can be found https://github.com/kubernetes-sigs/controller-runtime/blob/master/examples/configfile/builtin/main.go#L47-L49 with a config like https://github.com/kubernetes-sigs/controller-runtime/blob/master/examples/configfile/builtin/config.yaml I've first implemented this as a flag on `init --componentconfig` which is a boolean type. As developers use this the idea is they can implement custom configs with these standard configs embedded https://github.com/kubernetes-sigs/controller-runtime/blob/master/examples/configfile/custom/main.go#L49-L59 these files are decoded and made available making it easy to extend your controller config without relying on CLI flags.

For the `kubebuilder` implementation this scaffolds the config file and manager patch for volumes and configures `kustomize` to create a configmap for this config file.

I'm submitting this cause I'd like to get feedback before I go and refactor some of the fields to Mixins and Injection. As well this is blocked by #1746 for a stable `controller-runtime` 0.7.0 release and #1756 for v3 e2e tests.

Fixes #722

Signed-off-by: Chris Hein <me@chrishein.com>


Adirio:
Prow is not detecting it as a WIP so I will manually hold it, feel free to remove it when its done
/hold wip

pkg/plugin/v3/scaffolds/internal/templates/main.go
```diff
@@ -44,30 +44,27 @@ func (f *Main) SetTemplateDefaults() error {
...
+		options = ""
+		componentConfig = ".AndFromOrDie(ctrl.ConfigFile())"
+	}
+
 	f.TemplateBody = fmt.Sprintf(mainTemplate,
 		file.NewMarkerFor(f.Path, importMarker),
 		file.NewMarkerFor(f.Path, addSchemeMarker),
+		flags,
+		options,
+		componentConfig,
```
Adirio, line 61: You should not template in `SetTemplateDefaults`. Use `{{ if .ComponentConfig }` conditionals to template in the template body.

pkg/plugin/v3/init.go
```diff
@@ -40,8 +40,9 @@ type initPlugin struct {
...
 
 	// boilerplate options
-	license string
-	owner   string
+	license         string
+	owner           string
+	componentConfig bool
 
 	// flags
 	fetchDeps          bool
```
Adirio, line 48: Place it with the flags instead of with the boilerplate options
```suggestion
	license string
	owner   string

	// flags
	componentConfig    bool
	fetchDeps          bool
```

pkg/plugin/v3/scaffolds/init.go
```diff
@@ -101,11 +103,18 @@ func (s *initScaffolder) scaffold() error {
 		&templates.GitIgnore{},
 		&rbac.AuthProxyRole{},
 		&rbac.AuthProxyRoleBinding{},
-		&kdefault.AuthProxyPatch{},
+		&kdefault.AuthProxyPatch{
+			ComponentConfig: s.componentConfig,
+		},
```
Adirio, line 108: I feel like being a long nlist of templates, single lines for each are better unless really long.
```suggestion
		&kdefault.AuthProxyPatch{ComponentConfig: s.componentConfig},
```

pkg/plugin/v3/scaffolds/init.go
```diff
@@ -101,11 +103,18 @@ func (s *initScaffolder) scaffold() error {
...
 		&rbac.ClientClusterRole{},
-		&manager.Config{Image: imageName},
-		&templates.Main{},
+		&manager.Config{
+			Image:           imageName,
+			ComponentConfig: s.componentConfig,
+		},
+		&templates.Main{
+			ComponentConfig: s.componentConfig,
+		},
```
Adirio, line 117: Same as above
```suggestion
		&manager.Config{Image: imageName, ComponentConfig: s.componentConfig},
		&templates.Main{ComponentConfig: s.componentConfig},
```

pkg/plugin/v3/scaffolds/internal/templates/config/kdefault/auth_proxy_patch.go
```diff
@@ -36,14 +40,25 @@ func (f *AuthProxyPatch) SetTemplateDefaults() error {
 		f.Path = filepath.Join("config", "default", "manager_auth_proxy_patch.yaml")
 	}
 
-	f.TemplateBody = kustomizeAuthProxyPatchTemplate
+	args := argsFragment
+	if f.ComponentConfig {
+		args = ""
+	}
+
+	f.TemplateBody = fmt.Sprintf(kustomizeAuthProxyPatchTemplate, args)
```
Adirio, line 48: Same as in `main`, template inside the template body with `{{ if .ComponentConfig }}` instead of in `SetTemplateDefaults`.

pkg/plugin/v3/scaffolds/internal/templates/config/manager/config.go
```diff
@@ -38,11 +42,20 @@ func (f *Config) SetTemplateDefaults() error {
 		f.Path = filepath.Join("config", "manager", "manager.yaml")
 	}
 
-	f.TemplateBody = configTemplate
+	args := argsFragment
+	if f.ComponentConfig {
+		args = ""
+	}
+
+	f.TemplateBody = fmt.Sprintf(configTemplate, args)
```
Adirio, line 50: Same as in `main`, template inside the template body with `{{ if .ComponentConfig }}` instead of in `SetTemplateDefaults`.

pkg/plugin/v3/scaffolds/internal/templates/config/manager/kustomization.go
```diff
@@ -27,6 +27,8 @@ var _ file.Template = &Kustomization{}
 // Kustomization scaffolds the Kustomization file in manager folder.
 type Kustomization struct {
 	file.TemplateMixin
+
+	ComponentConfig bool
```
Adirio, line 31: Usage?

pkg/plugin/v3/scaffolds/init.go
```diff
@@ -37,7 +37,7 @@ import (
 
 const (
 	// ControllerRuntimeVersion is the kubernetes-sigs/controller-runtime version to be used in the project
-	ControllerRuntimeVersion = "v0.6.3"
+	ControllerRuntimeVersion = "v0.7.0-alpha.6"
```
Adirio, line 40: NIT: I'm not sure how do I feel with adding an alpha version here on our own library.

Adirio (changes requested):
Several comments:

pkg/plugin/v3/scaffolds/init.go
christopherhein: Agree, this is blocked on #1746 for this but so that I could test I wanted to push my changes.

pkg/plugin/v3/init.go
```diff
@@ -87,6 +88,8 @@ func (p *initPlugin) BindFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&p.license, "license", "apache2",
 		"license to use to boilerplate, may be one of 'apache2', 'none'")
 	fs.StringVar(&p.owner, "owner", "", "owner to add to the copyright")
+	fs.BoolVar(&p.componentConfig, "componentconfig", false,
+		"create a versioned componentconfig file, may be 'true' or 'false'")
```
camilamacedo86, line 92: Hi @christopherhein, 

Why not. create a new plugin for it? IMO we could have a ComponentConfigPlugin such as the INIT plugin which means that if users would like to use this option at the same way that the use the addon as well. 

Also, why it is a breaking change? Is only because of the upper version of controller-runtime?

christopherhein: I implemented this as not a breaking change other than the CR version which is going to be done in #1746, the original issue had mentioned to make componentconfig the default and I'm thinking it's the right idea to do that until we have in controller runtime automatic version conversion to at least `v1beta1`.

christopherhein: Regarding doing this as a plugin; can you point me to where I should read about for what that would entail?

Adirio: `pkg/plugin` has the interfaces for init, edit, create API and create webhook plugins, and `pkg/plugin/vX` implement those interfaces. The calls to this interfaces is done from `pkg/cli` from the file with the corresponding names.

- I think Camila means creating a new type of plugin (requiring new interfaces similar to `pkg/plugin.Init` and `pkg/plugin.InitSubcommand`) but I'm not sure if it is a bit of an over kill. I think it may be a good thing to add to the edit plugin, which right now is only used to turn on multi-group.
- We could have a `edit component-config` command that enables or disables this behavior but in this case we would only be targeting `v3` as it would require a new field in the project configuration file which would be a breaking change with `v2`. 
- We could also make the `edit component-config` scaffold the relevant files instead of enabling the feature, which would be `v2`-compatible.

christopherhein: @Adirio and @camilamacedo86 I'd considered doing it like the edit command first the difficult thing is we'd have to rewrite a lot of the init code to support "edit component-config" which isn't currently addressable with markers. 

For example removing all the flag declarations in `main.go` so that componentconfig takes precedence  https://github.com/kubernetes-sigs/kubebuilder/pull/1790/files#diff-6d104501fc7dc73bb67118a31ea122a23d55af27d443c7b96029054238978a36R113-R124  or where I have to remove all the flags from the config/manager and config/default/auth_proxy - https://github.com/kubernetes-sigs/kubebuilder/pull/1790/files#diff-d524621fbcb369b36df89df9480d29cb86e7779e7b6b2dec466f66f11742cae6L46-R61

I'd like to keep this from being a breaking v2 change as much as possible. There is a longer-term implementation that once we switch to it being the default we can make it so that you can supply a custom type and it would scaffold that for your controller see https://github.com/kubernetes-sigs/controller-runtime/blob/master/examples/configfile/custom/main.go#L49-L53 which might be a good place to use the `edit` changes.

pkg/plugin/v3/scaffolds/internal/templates/config/manager/kustomization.go
christopherhein: good catch, thanks.

pkg/plugin/v2/scaffolds/internal/templates/main.go
```diff
@@ -18,17 +18,14 @@ package templates
 
 import (
 	"fmt"
-	"hash/fnv"
 	"path/filepath"
-	"text/template"
```
camilamacedo86, line 23: Why we are changing the v2+ plugin? 
This change requires to bump the controller-runtime and it has to break changes.  So, we can just to do it for v3+.

pkg/plugin/v3/init.go
camilamacedo86: Hi @christopherhein, 

I catcher its end goal now. Would really be a common bahaviour users want to init a project with config component by default? Would not possible they want to change for this option after the project be already init? 

So, WDYT
- we add new attribute into the PROJECT file to let us know if the project has or not this option such as the multigroup one. (IMO: we need to store all inputs otherwise will be impossible one day we achieve end goals such as have a migration tool for example) 
- then, could we not notify and re-scaffold the main and auth_proxy if we kb edit configcomponent=true/false`? Following an example:

> kubebuilder edit --component-config
> Beware that this command will re-generate the files: main.go and config/default/manager_auth_proxy_patch.yaml. Would you like to continue? [y/n]

pkg/plugin/v3/scaffolds/init.go
camilamacedo86: For v3-alpha it is fine because it is an alpha stage and not the default plugin. See; https://github.com/kubernetes-sigs/kubebuilder/issues/1746

pkg/plugin/v2/scaffolds/internal/templates/main.go
christopherhein: We currently had copied these funcs into the v3 space and when I added support for the componentconfig implementation I was going to copy it to yet another but instead moved those to the `file/funcmap.go` so all templates can use this if they need https://github.com/kubernetes-sigs/kubebuilder/pull/1790/files#diff-13abe45cd80694321eb97754f6cb9aced5ccbbf51cd62b53baad70d4c4eff4e5R34-R41 I can revert this if we want the code should be the same just not needing a custom funcmap.

pkg/plugin/v3/init.go
camilamacedo86: I am trying to figure out a way to allow users to have this option as they have the multigroup support one or addon. Since it is an option that shows that add to the init might not really be useful in daily bases.  However, feel free to raise other ideas as well. 

Also, we might need to have a new testdata/ for v3 with this scenario as well.  See: https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/ and the projects are build from https://github.com/kubernetes-sigs/kubebuilder/blob/master/generate_testdata.sh.

christopherhein: That seems like a reasonable way of handling it with `edit`. Asking people before it regenerates could be nice. Are we doing that with multi-group to support moving controllers to `apis/` from `api/`?

@vincepri since you are interested in using this for `v1alpha4` CAPI does this experience raise any issues?

Adirio: `kubebuilder edit --multigroup` doesn't regenerate the projects, you are requested to do that manually. It just sets a flag in the project configuration file that will make future calls to `kubebuilder create ...` use the corresponding directory structure.

christopherhein: Hmm, so if I'm reading what you're saying @Adirio correctly it doesn't sound like this would be the right candidate for `edit`, since we'd need to make subtle changes like adding `.AndFromOrDie(ctrl.ConfigFile())` to the `ctrl.Options{}` initialization which wouldn't work to add Markers for. All that being keeping it as a flag on `init` seems to be the best path. 

Am and I understanding properly?

Adirio: Could that be solved by creating a options first, placing a marker for your function call after, and then creating the manager below that? We could probably also have better error handling than `AndFromOrDie` too.

camilamacedo86: H @christopherhein and @Adirio,

Actually, the edit multigroup does: 
- Update the dockerfile accordingly
- Update the project file as well
- All commands used the value in the Project file to generate the manifests accordingly
- And then, we have a few tasks to improve this behaviour too. See : https://github.com/kubernetes-sigs/kubebuilder/issues/1640

For multigroup scenario, is not required to re-generated the files. However, for the case of componentconfig the easy path shows do that at least at the first moment do it. We can try to improve the behaviour after as well.

pkg/plugin/v2/scaffolds/internal/templates/main.go
camilamacedo86: I think it is fine 👍

pkg/plugin/v3/scaffolds/internal/templates/config/kdefault/kustomization.go
```diff
@@ -70,11 +72,15 @@ bases:
...
-  # If you want your controller-manager to expose the /metrics
-  # endpoint w/o any authn/z, please comment the following line.
+# Protect the /metrics endpoint by putting it behind auth.
+# If you want your controller-manager to expose the /metrics
+# endpoint w/o any authn/z, please comment the following line.
 - manager_auth_proxy_patch.yaml
 
+# Mount the controller config file for loading manager configurations
+# through a ComponentConfig type
+{{ if .ComponentConfig }}#{{ end }}- manager_config_patch.yaml
```
Adirio, line 82:
```suggestion
{{ if not .ComponentConfig }}#{{ end }}- manager_config_patch.yaml
```
Right?

pkg/plugin/v3/scaffolds/internal/templates/config/kdefault/manager_auth_proxy_patch.go
```diff
@@ -63,8 +66,10 @@ spec:
         ports:
         - containerPort: 8443
           name: https
-      - name: manager
+      {{- if not .ComponentConfig }}
+			- name: manager
         args:
         - "--metrics-addr=127.0.0.1:8080"
         - "--enable-leader-election"
+			{{- end }}
```
Adirio, line 74: I would either do this:
```suggestion
      {{ if not .ComponentConfig -}}
      - name: manager
        args:
        - "--metrics-addr=127.0.0.1:8080"
        - "--enable-leader-election"
      {{- end }}
```
or:
```suggestion
{{- if not .ComponentConfig }}
      - name: manager
        args:
        - "--metrics-addr=127.0.0.1:8080"
        - "--enable-leader-election"
{{- end }}
```

pkg/plugin/v3/scaffolds/internal/templates/config/manager/config.go
```diff
@@ -72,8 +75,10 @@ spec:
       containers:
       - command:
         - /manager
+        {{- if not .ComponentConfig }}
         args:
         - --enable-leader-election
+        {{- end }}
```
Adirio, line 81: Apply the same approach as in the previous comment.

pkg/plugin/v3/scaffolds/internal/templates/main.go
```diff
@@ -231,13 +216,14 @@ func init() {
...
 func main() {
+	{{ if not .ComponentConfig }}
 	var metricsAddr string
 	var enableLeaderElection bool
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. " +
 		"Enabling this will ensure there is only one active controller manager.")
-
+	{{ end }}
```
Adirio, line 226: Same.

pkg/plugin/v3/scaffolds/internal/templates/main.go
```diff
@@ -246,13 +232,24 @@ func main() {
...
 	})
+	{{- else }}
+	options, err := ctrl.Options{Scheme: scheme}.AndFrom(ctrl.ConfigFile())
+	if err != nil {
+		setupLog.Error(err, "unable to load the config file")
+		os.Exit(1)
+	}
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), options)
+	{{- end }}
```
Adirio, line 252: Same.

Adirio (changes requested):

Adirio (commented):
There are a lot of changes in the `testdata` folder that I don't think that are expected, could you check them?

christopherhein:
> There are a lot of changes in the `testdata` folder that I don't think that are expected, could you check them?

Yea, the tests are failing cause of the 0.7.0 which I think we need to get the declarative addons package updated so all the files are generated properly.

christopherhein:
BTW Thanks @Adirio for all the reviews on this and the response time is incredibly fast 👏  at this point we're blocked on https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/127

estroz:
@christopherhein #1804 bumps controller-runtime to v0.7.0-alpha.6

christopherhein:
Thanks @estroz

pkg/plugin/v2/scaffolds/internal/templates/main.go
Adirio: In order to keep it clean, I think this change should be done in a separate PR with the 🌱 tag, and this PR should be tagged v3-only, so that it is clear that the breaking changes are not happening to v2.

@camilamacedo86  @christopherhein

Adirio:
Could we nhomogenize the names? I'm seeing controller-config, controller-manager-config, component-config, and many others. Aligning with controller-runtime would probably be the best bet. We also try to keep the struct, path of scaffolded file and template file names in sync inside `pkg/plugin/vX/scaffolds/internal/templates`, and I would also try to keep deployment and other kubernetes names also in sync in order to avoid confusion.

Additionally, it is scaffolding two almost identical files: `config.yaml` and `config/manager/config.yaml` where aside of the path, the only difference is the leader election flag. Is this intended?

pkg/plugin/v2/scaffolds/internal/templates/main.go
christopherhein: :+1:

christopherhein: I will remove this work after https://github.com/kubernetes-sigs/kubebuilder/pull/1812 is merged.

pkg/plugin/v3/scaffolds/internal/templates/config/kdefault/config_manager_patch.go
```diff
@@ -0,0 +1,61 @@
+/*
```
Adirio, line 1: Could you rename this to `manager_config_patch.go`? The struct is `ManagerConfigPatch` and the scaffolded file `manager_config_patch.yaml`.

pkg/plugin/v3/scaffolds/internal/templates/componentconfig.go
```diff
@@ -0,0 +1,55 @@
...
+package templates
+
+import (
+	"sigs.k8s.io/kubebuilder/v2/pkg/model/file"
+)
+
+var _ file.Template = &ComponentConfig{}
+
+// ComponentConfig scaffolds the ComponentConfig file in manager folder.
+type ComponentConfig struct {
```
Adirio, line 26: What about
```suggestion
type ManagerConfig struct {
```
or
```suggestion
type ControllerManagerConfig struct {
```
?
Apply the corresponding change also to this file name (`[controller_]manager_config.go`) and to the `f.Path` field  (`[controller_]manager_config.yaml`) too.

If you opt for the "controller" prefixed version, probably the patch should also have the same prefix to share the same scheme.

Adirio (changes requested):
I'm still unsure if scaffolding two identical files with the exception of the `leaderElect` field is really needed.

pkg/plugin/v3/scaffolds/internal/templates/config/kdefault/config_manager_patch.go
christopherhein: Absolutely, I was noticing that the directory where this lives had some mixed struct to filename associations. see https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/v3/scaffolds/internal/templates/config/kdefault/webhook_manager_patch.go and https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/v3/scaffolds/internal/templates/config/kdefault/enablecainection_patch.go

Potentially in another PR we should clean those up.

pkg/plugin/v3/scaffolds/internal/templates/componentconfig.go
christopherhein: Yeah, we also probably want to rename https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/v3/scaffolds/internal/templates/config/manager/config.go at some point too.

christopherhein:
> I'm still unsure if scaffolding two identical files with the exception of the `leaderElect` field is really needed.

I think I can find an elegant solution to not requiring both, it will unfortunately probably require keeping the leader election flag and overriding it. when deploying using Kustomize and defaulting off so that `make run` still works.

christopherhein:
blocked by #1812

pkg/plugin/v3/scaffolds/internal/templates/config/kdefault/config_manager_patch.go
Adirio: I did a first round cleaning up things but there are some left yes

pkg/plugin/v3/scaffolds/internal/templates/config/manager/manager_config.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"path/filepath"
+
+	"sigs.k8s.io/kubebuilder/v2/pkg/model/file"
+)
+
+var _ file.Template = &ControllerManagerConfig{}
+
+// ControllerManagerConfig scaffolds the ComponentConfig file in manager folder.
+type ControllerManagerConfig struct {
```
Adirio, line 28: I guess we can have it like this (Linter doesn't allow to call it `ManagerConfig` as it is inside the `manager` package) until we clean the whole templates.

Adirio (approved):

pkg/plugin/v3/scaffolds/internal/templates/config/manager/manager_config.go
christopherhein: Sorry about that, I pushed before renaming the file to match.

Adirio (approved):
/lgtm

WIP status removed:
/hold cancel

christopherhein:
@camilamacedo86 @estroz this should be all updated based on the feedback. If one of you could PTAL. Thanks!

Adirio:
Reapplying `lgtm`

/lgtm

testdata/project-v3/main.go
```diff
@@ -49,12 +49,12 @@ func init() {
 
 func main() {
 	var metricsAddr string
-	var enableLeaderElection bool
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+
+	var enableLeaderElection bool
```
camilamacedo86, line 54: I think we should not to ddo these changes.

testdata/project-v3/config/default/manager_config_patch.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: apps/v1
```
camilamacedo86, line 1: Shows that we are missing for the addon. Should we not to do the same for the addon projects?

testdata/project-v3-multigroup/config/default/kustomization.yaml
```diff
@@ -25,11 +25,15 @@ bases:
...
-  # endpoint w/o any authn/z, please comment the following line.
+# Protect the /metrics endpoint by putting it behind auth.
+# If you want your controller-manager to expose the /metrics
+# endpoint w/o any authn/z, please comment the following line.
 - manager_auth_proxy_patch.yaml
 
+# Mount the controller config file for loading manager configurations
+# through a ComponentConfig type
+#- manager_config_patch.yaml
+
```
camilamacedo86, line 36: WDYT about we use this config in the e2e tests to ensure this feature? See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v3/generate_test.go#L79

christopherhein: Yeah, I can get an e2e test for this as well.

pkg/plugin/v3/init.go
```diff
@@ -86,6 +87,8 @@ func (p *initSubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&p.license, "license", "apache2",
 		"license to use to boilerplate, may be one of 'apache2', 'none'")
 	fs.StringVar(&p.owner, "owner", "", "owner to add to the copyright")
+	fs.BoolVar(&p.componentConfig, "component-config", false,
+		"create a versioned ComponentConfig file, may be 'true' or 'false'")
```
camilamacedo86, line 91: If we will add a the new attribute that changes the scaffold then, IMO we need to persist it in the PROJECT file

christopherhein: :+1: sounds good

testdata/project-v3-multigroup/config/default/kustomization.yaml
christopherhein: Just to confirm we're taking about uncommenting this line but not using passing the `--component-config` flag into the e2e `init` command, right? The idea is to test if we can uncomment and use kustomize without breakages, not to test the controller booting while loading from the file, right?

pkg/plugin/v3/init.go
christopherhein: Done.

testdata/project-v3-multigroup/config/default/kustomization.yaml
christopherhein: For now I've added https://github.com/kubernetes-sigs/kubebuilder/pull/1790/files#diff-937f62ec43b0cd36407ae8e72d9f605f24f53968f898d3a4ada2ba47c4c70455R184-R186 which will handle at least testing that kustomize doesn't break when this is uncommented and that they controller still boots.

test/e2e/v3/generate_test.go
```diff
@@ -181,6 +181,9 @@ Count int `+"`"+`json:"count,omitempty"`+"`"+`
 	ExpectWithOffset(1, utils.UncommentCode(
 		filepath.Join(kbc.Dir, "config", "default", "kustomization.yaml"),
 		"#- ../prometheus", "#")).To(Succeed())
+	ExpectWithOffset(1, utils.UncommentCode(
+		filepath.Join(kbc.Dir, "config", "default", "kustomization.yaml"),
+		"- manager_config_patch.yaml", "#")).To(Succeed())
```
camilamacedo86, line 186:
```suggestion
	ExpectWithOffset(1, utils.UncommentCode(
		filepath.Join(kbc.Dir, "config", "default", "kustomization.yaml"),
		"#- manager_config_patch.yaml", "#")).To(Succeed())
```

camilamacedo86: Shows that we need to fix the scaffold. IMO it ought to be == the others.

testdata/project-v3/main.go
```diff
@@ -49,12 +49,12 @@ func init() {
```
camilamacedo86, line 49: IMO we need to create a new testdata_generate with this scenario as well such as we have for multigroup and addon. See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/generate_testdata.sh

WDYT?

testdata/project-v3/config/default/manager_config_patch.yaml
camilamacedo86: Solved

testdata/project-v3-multigroup/config/default/kustomization.yaml
camilamacedo86: solved

testdata/project-v3/main.go
christopherhein: Because this param is on the `init` should I have testdata generated for addon + config and multi-group + config and config without any other options? or should I just create a single case with config and no other options?

Adirio: IMHO you should only have 1 testdata set for this, not 3. Multigroup and addon don't affect the component-config setup, and adding more testdata folders add way more load to the test done in CI for very push. Bear in mind that for every testdata folder, we are creating it, checking that every file created and the ones in the repo are exactly the same character by character, and then performing their tests for several platforms (Linux and MacOs in GH Actions, and I think several others in Prow).

generate_testdata.sh
```diff
@@ -113,3 +114,6 @@ scaffold_test_project project-v2-addon 2
 scaffold_test_project project-v3 3-alpha go/v3-alpha
 scaffold_test_project project-v3-multigroup 3-alpha go/v3-alpha
 scaffold_test_project project-v3-addon 3-alpha go/v3-alpha
+scaffold_test_project project-v3-config 3-alpha go/v3-alpha "--component-config"
+scaffold_test_project project-v3-multigroup-config 3-alpha go/v3-alpha "--component-config"
+scaffold_test_project project-v3-addon-config 3-alpha go/v3-alpha "--component-config"
```
camilamacedo86, line 119: Could we not only add more 1 scaffold such as : `project-v3-component-config`? I think it would be enough to ensure/check it. IMO add more 3 testdata sample will lead us to have 9 samples which can start to make hard to keep it maintained. WDYT?

pkg/plugin/v3/scaffolds/internal/templates/config/manager/controller_manager_config.go
```diff
@@ -0,0 +1,57 @@
...
+
+const controllerManagerConfigTemplate = `apiVersion: controller-runtime.sigs.k8s.io/v1alpha1
+kind: ControllerManagerConfig
+metrics:
+  bindAddress: :8080
+webhook:
+  port: 9443
+leaderElection:
+  leaderElect: false
+  resourceLock: configmaps
```
estroz, line 55: The default in controller-runtime is `configmapsleases` as of v0.7.0. Since the manager doesn't currently specify this value, neither should the config unless we want to preserve prior behavior for some reason (we'd also have to revert #1809).

pkg/plugin/v3/scaffolds/internal/templates/main.go
```diff
@@ -213,13 +214,14 @@ func init() {
 }
 
 func main() {
+{{- if not .ComponentConfig }}
 	var metricsAddr string
-	var enableLeaderElection bool
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+{{ end }}
+	var enableLeaderElection bool
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
```
estroz, line 222: Since leader election is configured by the config, wouldn't its flag not be scaffolded if using a config patch?

pkg/plugin/v3/scaffolds/internal/templates/main.go
```diff
@@ -228,13 +230,26 @@ func main() {
...
 		MetricsBindAddress: metricsAddr,
 		Port:               9443,
 		LeaderElection:     enableLeaderElection,
 		LeaderElectionID:   "{{ hashFNV .Repo }}.{{ .Domain }}",
 	})
+{{- else }}
+	options, err := ctrl.Options{
+		Scheme:         scheme,
+		LeaderElection: enableLeaderElection,
+	}.AndFrom(ctrl.ConfigFile().AtPath("config/manager/controller_manager_config.yaml"))
```
estroz, line 245: Nit: set `const managerConfigPath = "config/manager/controller_manager_config.yaml"` at the top of the file and use variable here

generate_testdata.sh
estroz: Agreed
```suggestion
scaffold_test_project project-v3-config 3-alpha go/v3-alpha --component-config
```

pkg/plugin/v3/scaffolds/internal/templates/main.go
Adirio: I would say that, to keep it similar to the rest of the files, not using a `const` in this case is ok. If we want them to be `const` we should probably do it for all the templates and not only this one, and that would be a separate clean up PR.

generate_testdata.sh
christopherhein: Sounds good to me 👍

pkg/plugin/v3/scaffolds/internal/templates/config/manager/controller_manager_config.go
christopherhein: Oh awesome, happy to remove this, I was noticing permission issues before. Thanks for letting me know.

pkg/plugin/v3/scaffolds/internal/templates/main.go
christopherhein: Great question, if possible yes, the difficult thing is we'd have to patch this before the configmap is written and I'm not aware right now of a way to do this or else `make run` will fail because the config file won't load. My first pass created separate files for each local vs in-cluster but I fear that could be confusing to folks.

Any chance you can think of a good way around that?

pkg/plugin/v3/scaffolds/internal/templates/main.go
christopherhein: I actually think I should be making this a flag `--config` to follow the standard practice with component config implementers.

estroz: ^ sgtm

Adirio:
Milestone 3.0.0 or 3.1.0? IMHO 3.1.0
@camilamacedo86 @estroz

estroz:
@Adirio plugin go/v3 stabilization shouldn't block on this because it is an addition, so v3.1.0 sounds right. If it gets in before that happens, great!

pkg/plugin/v3/scaffolds/internal/templates/main.go
```diff
@@ -213,13 +214,17 @@ func init() {
...
 
 func main() {
+{{- if not .ComponentConfig }}
 	var metricsAddr string
 	var enableLeaderElection bool
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. " +
 		"Enabling this will ensure there is only one active controller manager.")
-
```
camilamacedo86, line 222: we can add here 
// nolint:lll

To fix it

christopherhein: Thanks, the flag message was actually wrong so I'm glad this caught it. Updating now.

christopherhein:
@estroz had recommended I also create a docs page which explains a little more about this and points to the controller runtime code.

christopherhein:
Doc PR to track against for built-in setup https://github.com/kubernetes-sigs/kubebuilder/issues/1841 and the way to extend the config.

pkg/plugins/golang/v3/scaffolds/internal/templates/main.go
```diff
@@ -213,13 +214,20 @@ func init() {
...
 		"Enable leader election for controller manager. " +
 		"Enabling this will ensure there is only one active controller manager.")
-
+{{- else }}
+  var configFile string
+	flag.StringVar(&configFile, "config", "controller_manager_config.yaml", 
+		"The controller will load its initial configuration from this file. " +
+		"Omit this flag to use the default configuration values. " +
+		"Command-line flags override configuration from this file.")
+{{ end }}
```
estroz, line 230: Nit: preserve previous spacing
```suggestion
{{- end }}

```

estroz (changes requested):
If the default value for `--config` is "", `CONFIG_FILE` in the Makefile isn't necessary. Otherwise LGTM!

estroz (approved):
/lgtm

testdata/project-v3-addon/.gitignore
```diff
@@ -6,6 +6,7 @@
 *.so
 *.dylib
 bin
+testbin/*
```
Adirio, line 9: Didn't see any changes in `pkg/plugins/golang/v3/scaffolds/templates/gitignore.go`.

testdata/project-v3-multigroup/.gitignore
```diff
@@ -6,6 +6,7 @@
 *.so
 *.dylib
 bin
+testbin/*
```
Adirio, line 9: Same

testdata/project-v3-config/.gitignore
```diff
@@ -0,0 +1,25 @@
+
+# Binaries for programs and plugins
+*.exe
+*.exe~
+*.dll
+*.so
+*.dylib
+bin
+testbin/*
```
Adirio, line 9: Same

testdata/project-v3/.gitignore
```diff
@@ -6,6 +6,7 @@
 *.so
 *.dylib
 bin
+testbin/*
```
Adirio, line 9: Same

testdata/project-v3-addon/.gitignore
Adirio: Self-reply: That line was added several months ago by #1626 and were never updated in the corresponding `testdata` directory. I think we should remove them from this PR and track this error separately, as it may mean that checking the `testdata` directory may be bugged.

camilamacedo86: What is missing here is to update the Kubebuiler .gitignore to ensure that the bins and go.sum will not be committed for the new testdata sample see:
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/.gitignore#L19
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/.gitignore#L37

test/e2e/v3/generate_test.go
```diff
@@ -200,6 +200,9 @@ Count int `+"`"+`json:"count,omitempty"`+"`"+`
 	ExpectWithOffset(1, utils.UncommentCode(
 		filepath.Join(kbc.Dir, "config", "default", "kustomization.yaml"),
 		"#- ../prometheus", "#")).To(Succeed())
+	ExpectWithOffset(1, utils.UncommentCode(
+		filepath.Join(kbc.Dir, "config", "default", "kustomization.yaml"),
+		"#- manager_config_patch.yaml", "#")).To(Succeed())
```
estroz, line 205: The e2e project wasn't scaffolded with `--component-config` so this patch will cause `kustomize build` conflicts.
```suggestion
```

estroz (changes requested):
PR needs a rebase and `make generate`

generate_testdata.sh
```diff
@@ -113,3 +114,4 @@ scaffold_test_project project-v2-addon 2
 scaffold_test_project project-v3 3-alpha go/v3-alpha
 scaffold_test_project project-v3-multigroup 3-alpha go/v3-alpha
 scaffold_test_project project-v3-addon 3-alpha go/v3-alpha
+scaffold_test_project project-v3-config 3-alpha go/v3-alpha "--component-config"
```
estroz, line 117: From #1851:
```suggestion
scaffold_test_project project-v3-config --project-version=3-alpha --plugins=go/v3-alpha --component-config
```

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
```diff
@@ -84,7 +90,7 @@ manager: generate fmt vet
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
 run: generate fmt vet manifests
-	go run ./main.go
+	go run ./main.go --config=${CONFIG_FILE}
```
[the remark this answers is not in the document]
estroz: I still think this should be removed, since a user can more easily hard-code their local config path here than set an environment variable every time they `make run`.

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
```diff
@@ -61,6 +62,11 @@ const makefileTemplate = `
 IMG ?= {{ .Image }}
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
+{{- if .ComponentConfig }}
+# Defines an empty config path when loading locally, this can be overwritten to
+# load a custom config like config/manager/controller_manager_config.yaml
+CONFIG_FILE ?= ""
+{{- end }}
```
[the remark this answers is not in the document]
estroz: Ditto

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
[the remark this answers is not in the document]
christopherhein: Sorry, I swear I removed this, might have been on my other machine 🤦‍♂️

estroz (approved):
/lgtm

christopherhein:
🎉 thanks @Adirio @estroz @camilamacedo86 really appreciate all the help getting this through.

luxas:
Kudos to everyone, amazing to see this land :tada: :clap: :1st_place_medal: !!

Issue #1562, opened 2020-06-15
enh: go get of kubebuilder by tag doesn't work


rbtcollins:
```shell
go get sigs.k8s.io/kubebuilder/cmd@v2.2.0
go get sigs.k8s.io/kubebuilder/cmd@v2.2.0: sigs.k8s.io/kubebuilder@v2.2.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

Alternative spellings fail in various related ways:

```shell
go get sigs.k8s.io/kubebuilder/cmd/v2.2.0
go: downloading sigs.k8s.io/kubebuilder v1.0.8
go get sigs.k8s.io/kubebuilder/cmd/v2.2.0: module sigs.k8s.io/kubebuilder@upgrade found (v1.0.8), but does not contain package sigs.k8s.io/kubebuilder/cmd/v2.2.0
```

I would expect kubebuilder to be go gettable, as that allows version-locking the kubebuilder version that is being used in go.mod/go.sum, with all the ensuing benefits that one would expect. Instead we have folk with curlsh scripts pulling down release builds versioned by hand...

Documentation on this is fairly comprehensive, but this bug comment is right at the heart of it: https://github.com/golang/go/issues/33010#issuecomment-510509689

tl;dr:
* Fix the import path- this repo is located on github, not on 
* Put v2 in the module import path in go.mod
* Adjust code imports in this repo accordingly
* https://github.com/google/go-github is an example go library with a release major version greater than 1, that has code in master

Right now go get of the .../cmd path works fine, as long as one uses a hash, not a commit tag.

/kind bug

DirectXMan12:
Yeah, so the issue was that kubebuilder wasn't intended to be a library originally -- it was just a tool.  The rules for fetching go modules don't really admit the admit the existence of tools.  For tools that aren't meant to be imported, bumping the major version without changing the import path is fine.  That being said, since we now have folks trying to import parts, we should fix this.

Re "Fix the import path": nope, `sigs.k8s.io` is correct.  Vanity imports are a thing in go.  The vanity import resolves correctly.

Issue #1791, opened 2020-11-05
Improve go modules fulfillment


Adirio:
- [x] Kubebuilder uses semantic versioning
- [x] Kubebuilder has a v2 or higher
- [x] Kubebuilder versions are listed on [pkg.go.dev](https://pkg.go.dev/mod/sigs.k8s.io/kubebuilder?tab=versions)

# Description
Although kubebuilder follows semantic versioning as required by go modules, only versions with major version 0 or 1 are actually obtainable by version with go modules. This can be seen in the link provided above.

The solution is to use a non-existent version and use the timestamp and commit hash to refer to a specific version. For example, [operator SDK](https://github.com/operator-framework/operator-sdk/blob/fd6179279e1cadf031bf857947b473e1ea414c2c/go.mod#L37) uses `sigs.k8s.io/kubebuilder v1.0.9-0.20201021204649-36124ae2e027` which refers to https://github.com/kubernetes-sigs/kubebuilder/commit/36124ae2e027c2643152f0d6b5bc7321c1417895.

In order to solve this, `go.mod` file needs to be changed to include `/v2` at the package name (`module sigs.k8s.io/kubebuilder/v2`) and internal imports need to reference to this same package.

I think it should definetely be implemented for future versions. However, I'm not sure what should be done with previous v2 versions. As previous versions are released, they are tagged and have their code compressed (it would change imports).

/kind feature

camilamacedo86:
I think it is totally fine we leave the previous versions as it is. See that kb has not been provided as lib before the plugins be offered.

pwittrock:
FYI -- for some reason this breaks pkger

camilamacedo86:
The v2 branch was updated with that already.See: https://github.com/kubernetes-sigs/kubebuilder/pull/1878
Then, for the next patch release, it will be done.

PR #1792, opened 2020-11-05
⚠️ Go module fullfillment for v2+

Adirio:
Targets #1791 for future versions

Fixes: #1562

It is not technically a breaking change but requires action from third-parties.


camilamacedo86 (commented):
Since the next release will be 3.0.0 and we cannot so longer push a patch release from master anyway then, IMO all here should be `v3`instead of `v2`. Otherwise, I am ok with the changes made. 

c/c @pwittrock @estroz @DirectXMan12 wdyt?

Adirio:
Are we 100% certain next kubebuilder version is going to be 3.0.0? Remember that I'm not talking about neither the kubebuilder base plugins nor the project configuration file version.

estroz (approved):
/approve

Adirio:
I think that with #1798 tracking the change to `v3`, this can be merged. It requires `lgtm`.

PR #1797, opened 2020-11-06
:seedling: remove alpha subcommand since it is no longer used

camilamacedo86:
**Description**
- remove alpha subcommand since it is no longer used

**Motivation** 

It is a dead/unused code. Also, after the [plugin implementation](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-1.md) the idea is that the Kubebuilder commands are extensible. In this way, any new feature and subcommand would be a plugin which makes this feature be useless for future possible cases as well.


Adirio (approved):

Adirio:
LGTM, as soon as it passes the tests I'l set the `lgtm` label

camilamacedo86:
I will force the merged because of `coverage/coveralls — Coverage decreased (-0.2%) to 71.774%`

PR #1801, opened 2020-11-09
🌱 [WIP] Proposal for Phase 2 Plugins

rashmigottipati:
Description of the change
This PR adds a design proposal for Kubebuilder Phase 2 Plugins

Motivation for the change
As the proposed scaffold plugin system is implemented, looking forward to achieving agreement on a way forward for Kubebuilder Phase 2 Plugins so that kubebuilder and operator sdk projects can align on requirements and progress on phase 2, which adds support for chaining plugins together

### DIFF
--- a/designs/plugins-phase-2.md
+++ b/designs/plugins-phase-2.md
@@ -1,0 +1,105 @@
+# Plugins Phase 2
+
+## Overview
+
+Phase 2 of the scaffold plugin system adds support for chaining plugins together. The idea for plugins phase 2 is to have support for a plugin architecture that enables plugins to be separate binaries discoverable by the Kubebuilder CLI binary via user specified plugin file paths and that they are language indepedent. Some of the phase 2 high-level requirements summarized from the feature request and discussions are: 
+
+* mdbook-like structure
+    * A Plugin receives a JSON blob plugin "universe" from upstream plugins containing file "nodes", each of which contain file metadata.
+    * The plugin can add or delete nodes (nodes are not mutable), then returns the modified universe so downstream plugins can do the same.
+* Pre-generate a project file for handling complex plugin use-cases.
+* Enhanced "layout" PROJECT file key.
+    * Either an ordered list of plugin {name}/{version}, or ordered map of {name}/{version} to plugin-specific config.
+    * Order matters for certain plugins, ex. a bazel plugin would need to run after all Go files are generated.
+
+## Evaluation of Plugin libraries 
+
+##### Go built-in plugin package vs Hashicorp plugin library vs custom plugin library
+
+Based on the evaluation of plugin libraries such as the built-in go-plugin library and Hashicorp’s plugin library, we have come to the conclusion that it is more suitable to write our own custom typed plugin library.
+
+The built-in plugin library seems to be a non-starter as it is more suitable for in-tree plugins rather than out-of-tree and it doesn’t offer cross language support. Hashicorp’s go plugin system seems more suitable than built-in go-plugin library as it enables cross language/platform support. However, it is more suited for long running plugins as opposed to short lived plugins and it could be overkill with the usage of protobuf.
+ 
+For the stated reasons, the proposal is to write our own plugin system that passes JSON blobs back and forth across stdin/stdout.
+The plugin specification will include type metadata to allow the potential for using other plugin libraries in the future if the need arises. From an implementation standpoint, we could create a new implementation of our plugin interface based on the type specified in the project file. We can do a type switch on the type of plugin the data is being passed to. For ex: if it is native go, then we pass the universe directly & if it is a binary wrapper, then we pass the serialized stream of JSON bytes to it. 
+
+For phase 2 plugin implementation, plugins need a data structure containing a set of file representations to read from and write to that can be passed between multiple plugins. Let's look at that below.
+
+#### Intermediate Representation
+
+The proposed plugin universe is: 
+
+```go
+type File struct {
+    // Path is the file to write
+    Path string `json:"path,omitempty"`
+    
+    // Contents is the generated output
+    Contents string `json:"contents,omitempty"`
+    
+    // Info is the os.FileInfo 
+    Info os.FileInfo  `json:"info,omitempty"`	
+}
+ 
+type Universe struct {
+    // Config stores the project file  
+    Config *config.Config `json:"config,omitempty"`
+
+    // Files contains the model of the files that are being scaffolded
+    Files map[string]File `json:"files,omitempty"`	
+}
+
+```
+
+#### Implementation Details
+
+The project file has a plugins field and one of the sub-fields of this field should be “type” that specifies either stdin/stdout or JSON. When kubebuilder reads plugins from the project file, we will have a new implementation of the plugin interface and it will be based on the type. For ex: if the plugin type is stdinStdout then we pass that struct which would implement the plugin interface and we would call a function and pass in the universe and that function returns the updated universe. The underlying implementation calls exec function on the binary and encodes into stdin stream and decodes from stdout stream into the structs we expect. So, Kubebuilder, as the calling side, is calling a function that sends a universe and receives an updated universe and the serialization happens on the kubebuilder side under the hood. And the plugin type is JSON that could be exec with stdin stdout, then we expect to get the universe as JSON stdin, read JSON, and serialize and write back on the stdout the JSON that’s expected by the kubebuilder. In the future, if we could like to extend the use cases and for some reason we would like to create a new plugin type (for example, the hashicorp’s go plugin) and in that case, instead of marshalling to JSON and setting up stdin stdout, we will receive the universe and marshal to the grpc proto format and send it over GRPC, and the plugin we ran should start up a grpc server on it’s own and implement the server interface and send the response back and we unpack that into the struct we expect.
+
+Boilerplate is not currently in the plugin Universe because it’s language specific and depending on the language, it could change. Some alternatives to this are using a boilerplate path and the plugin could check for its existence, or it could exist in the universe not as a special file but as a list of strings without comments or markers which could be read line by line, thereby adding it to any file. Along with the above `plugin.Universe`, all the raw flags that would be passed to the subcommand need to be passed along across stdin/stdout. And the idea here is to make the plugin generic such that the args are passed as they are without specifying the resource explicitly, which eliminates the need to write a plugin struct to support every possible subcommand like init, create api, and create webhook. 
+
+
+## Configuration
+
+### Project file plugin type & enhanced `layout`
+
+Currently, the `PROJECT` file specifies what base plugin generated the project under a `layout` key (which is the plugin semantic version) in the format: `Plugin.Name() + "/" + Plugin.Version()`.
+
+To support phase 2 plugins, we propose that we modify the "layout" PROJECT file key to either of these options as order matters for certain plugins: 
+    * Ordered list of plugin {name}/{version} or
+    * Ordered map of {name}/{version} to plugin-specific config
+
+Example `PROJECT` file:
+
+```yaml
+version: "3-alpha"
+layout: ["go/v1.0.0", "go.kubebuilder.io/v3-alpha"]
+domain: testproject.org
+repo: github.com/test-inc/testproject
+resources:
+- group: crew
+  kind: Captain
+  version: v1
+plugins:
+  - go.sdk.operatorframework.io/v2-alpha: {
+      type: JSON
+  }
+``` 
+
+## CLI
+
+Idea is to add support for a global --plugins flag that takes an ordered list of plugin names. The subcommand invoked with --plugins executes "downstream" plugins from that list that match those the kubebuilder binary knows about.
+
+Given that we want to pass some initial universe between plugins, that universe should be initialized by the plugin invoked via CLI (the "base" plugin). For example, kubebuilder init --plugins addon will invoke an Init plugin, which passes its generated state to the addon plugin, modifying scaffolded files then writing them to disk.
+
+## Prototype implementation
+
+Kubebuilder feature branch: https://github.com/kubernetes-sigs/kubebuilder/tree/feature/plugins-part-2-electric-boogaloo
+
+## Related issues and PRs
+
+* [Feature request: Plugins Phase 2](https://github.com/kubernetes-sigs/kubebuilder/issues/1378)
+
+## Comments & Questions
+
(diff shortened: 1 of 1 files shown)


designs/plugins-phase-2.md
```diff
@@ -0,0 +1,105 @@
...
+
+To support phase 2 plugins, we propose that we modify the "layout" PROJECT file key to either of these options as order matters for certain plugins: 
+    * Ordered list of plugin {name}/{version} or
+    * Ordered map of {name}/{version} to plugin-specific config
+
+Example `PROJECT` file:
+
+```yaml
+version: "3-alpha"
+layout: ["go/v1.0.0", "go.kubebuilder.io/v3-alpha"]
```
camilamacedo86, line 75: the layout means the father of the plugins used to scaffold the project. So, what would mean `"go/v1.0.0", "go.kubebuilder.io/v3-alpha"`? Could we give an example of a scenario that we would like to address and then have the project file as it would be in this scenario for a better understand?

designs/plugins-phase-2.md
```diff
@@ -0,0 +1,105 @@
...
+domain: testproject.org
+repo: github.com/test-inc/testproject
+resources:
+- group: crew
+  kind: Captain
+  version: v1
+plugins:
+  - go.sdk.operatorframework.io/v2-alpha: {
+      type: JSON
```
camilamacedo86, line 84: Since the tool has access to the plugin implementation, why we need to add the type here? Could you please add the example scenario and then describe why/when the Type is used in the process to illustrate why it is required at all?

designs/plugins-phase-2.md
```diff
@@ -0,0 +1,105 @@
...
+  }
+``` 
+
```
camilamacedo86, line 87: Could we add a real/valid scenario for we have a better understanding of the purposed idea? E.g Something such as: add the use case then address the scenario with the example. 

> I am as X, would like to N, so that I can Z. 

1. Run the command : `otherbin init --plugins=[a,b]`
2. Then, update the path path of the plugin
3. Do the scaffold 
4. The Project file would be like as:

designs/plugins-phase-2.md
```diff
@@ -0,0 +1,105 @@
...
+* mdbook-like structure
+    * A Plugin receives a JSON blob plugin "universe" from upstream plugins containing file "nodes", each of which contain file metadata.
+    * The plugin can add or delete nodes (nodes are not mutable), then returns the modified universe so downstream plugins can do the same.
+* Pre-generate a project file for handling complex plugin use-cases.
+* Enhanced "layout" PROJECT file key.
+    * Either an ordered list of plugin {name}/{version}, or ordered map of {name}/{version} to plugin-specific config.
+    * Order matters for certain plugins, ex. a bazel plugin would need to run after all Go files are generated.
+
+## Evaluation of Plugin libraries
```
camilamacedo86, line 15: Could we add here the use cases that we are trying to address with this proposal (`I am as X, would like to N, so that I can Z.`)? Also, could we add how we will test the plugins and ensure its quality?

camilamacedo86 (commented):
It is nice. I added a few questions.

designs/plugins-phase-2.md
```diff
@@ -0,0 +1,105 @@
...
+
+Kubebuilder feature branch: https://github.com/kubernetes-sigs/kubebuilder/tree/feature/plugins-part-2-electric-boogaloo
+
+## Related issues and PRs
+
+* [Feature request: Plugins Phase 2](https://github.com/kubernetes-sigs/kubebuilder/issues/1378)
+
+## Comments & Questions
+
+* How to handle help text? Are plugins allowed to define their own flags?
```
camilamacedo86, line 104: Could  you please add an example over who the consumer will consume and customize the plugins in phase 2? Or are we not covering it in this phase? For example, I am as a consumer(sdk) of kubebuilder plugins, would like to extend the init subcommand/plugin and define that the main.go will be a scaffold in a new path, so that for my tool can do the same scaffolds but in another path. 

Also, would be nice we have here what are the goals and no goals of this EP. So, if it is not a goal of this proposal we would easily know.

designs/plugins-phase-2.md
```diff
@@ -0,0 +1,105 @@
+# Plugins Phase 2
+
+## Overview
+
+Phase 2 of the scaffold plugin system adds support for chaining plugins together. The idea for plugins phase 2 is to have support for a plugin architecture that enables plugins to be separate binaries discoverable by the Kubebuilder CLI binary via user specified plugin file paths and that they are language indepedent. Some of the phase 2 high-level requirements summarized from the feature request and discussions are:
```
Adirio, line 5: Interpreted languages, i.e. Python, do not generate binaries. Are we leaving them apart intentionally? I think that interpreted languages are well suited for this kind of scripting purposes and considering them an option would be interesting.

designs/plugins-phase-2.md
```diff
@@ -0,0 +1,105 @@
+# Plugins Phase 2
+
+## Overview
+
+Phase 2 of the scaffold plugin system adds support for chaining plugins together. The idea for plugins phase 2 is to have support for a plugin architecture that enables plugins to be separate binaries discoverable by the Kubebuilder CLI binary via user specified plugin file paths and that they are language indepedent. Some of the phase 2 high-level requirements summarized from the feature request and discussions are: 
+
+* mdbook-like structure
+    * A Plugin receives a JSON blob plugin "universe" from upstream plugins containing file "nodes", each of which contain file metadata.
+    * The plugin can add or delete nodes (nodes are not mutable), then returns the modified universe so downstream plugins can do the same.
```
Adirio, line 9: I would like to discuss why nodes need to be inmutable. There are probably good reasons for you to consider them like that, I just want to know them.

designs/plugins-phase-2.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Evaluation of Plugin libraries 
+
+##### Go built-in plugin package vs Hashicorp plugin library vs custom plugin library
+
+Based on the evaluation of plugin libraries such as the built-in go-plugin library and Hashicorp’s plugin library, we have come to the conclusion that it is more suitable to write our own custom typed plugin library.
+
+The built-in plugin library seems to be a non-starter as it is more suitable for in-tree plugins rather than out-of-tree and it doesn’t offer cross language support. Hashicorp’s go plugin system seems more suitable than built-in go-plugin library as it enables cross language/platform support. However, it is more suited for long running plugins as opposed to short lived plugins and it could be overkill with the usage of protobuf.
+ 
+For the stated reasons, the proposal is to write our own plugin system that passes JSON blobs back and forth across stdin/stdout.
```
Adirio, line 23: Again, probably good reasons to choose stdin/stdout. Could we addd some justification why we chose this approach isntead of, for example, sockets?

designs/plugins-phase-2.md
```diff
@@ -0,0 +1,105 @@
...
+    // Info is the os.FileInfo 
+    Info os.FileInfo  `json:"info,omitempty"`	
+}
+ 
+type Universe struct {
+    // Config stores the project file  
+    Config *config.Config `json:"config,omitempty"`
+
+    // Files contains the model of the files that are being scaffolded
+    Files map[string]File `json:"files,omitempty"`
```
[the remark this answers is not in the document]
Adirio: I think it is talking about already templated files. Is there anyway that allows us to pass it before templating and templating after all plugins have run? I think we may be able to abstract part of the templating and scaffolding code so that plugins are easier to develop.

designs/plugins-phase-2.md
Adirio: I think this example is confusing, as both plugins are the base go plugin (one with the short form and the other one the full name) which doesn't make sense. Additionally, we have `v2` and `v3` versions of that plugin, no `v1.0.0` nor `v3-alpha`, to add up to the confussion.

designs/plugins-phase-2.md
Adirio: I think the type is meant to determine how the universe is passed. `JSON` doesn't seem to be a good name election as it is just a encoding language. I would say `in-tree` or `built-in` or something similar for the ones that we already have and `stdin/stdout` or something similar for the ones that you are proposing in this document.

designs/plugins-phase-2.md
```diff
@@ -0,0 +1,105 @@
...
+version: "3-alpha"
+layout: ["go/v1.0.0", "go.kubebuilder.io/v3-alpha"]
+domain: testproject.org
+repo: github.com/test-inc/testproject
+resources:
+- group: crew
+  kind: Captain
+  version: v1
+plugins:
+  - go.sdk.operatorframework.io/v2-alpha: {
```
Adirio, line 83: This doesn't match any of the plugins in `layout`. How should we interpret this?

designs/plugins-phase-2.md
Adirio, line 104: I would say that a `*.yaml` file that defines the characteristics of the plugin could be used. For example it could define that we should call it with `python file-of-the-plugin.py` and that uses stdin and stdout to communicate with the main executable. When the main executable discovers the plugins it would read this file and register that plugin. In this file we could also add thing like allowed flags, commands, descriptions, etc.

designs/plugins-phase-2.md
```diff
@@ -0,0 +1,105 @@
...
+Kubebuilder feature branch: https://github.com/kubernetes-sigs/kubebuilder/tree/feature/plugins-part-2-electric-boogaloo
+
+## Related issues and PRs
+
+* [Feature request: Plugins Phase 2](https://github.com/kubernetes-sigs/kubebuilder/issues/1378)
+
+## Comments & Questions
+
+* How to handle help text? Are plugins allowed to define their own flags?
+* Do we allow plugins to depend on environment variables?
```
Adirio, line 105: Cons: the same plugin call would do different things depending on the environment
Difficulties: hard to enforce a plugin not to use environmental variables when it is an external executable for you.

Adirio (commented):
Several questions, mostly to clarify myself:

Adirio:
Prow only detects WIP if they are the first word in the title, so I'm manually holding it as it won't get the `hold:work-in-progress` label automatically. Feel free to remove the hold once it is not a WIP.

/hold
/ok-to-test

designs/plugins-phase-2.md
Adirio, line 5: I feel like there are 2 steps here:
1. Chain plugins
2. Enable external/out-of-tree plugins
I will refer to Phase 1.5 to achieve the first step and Phase 2 to achieve both.

designs/plugins-phase-2.md
```diff
@@ -0,0 +1,105 @@
...
+
+Phase 2 of the scaffold plugin system adds support for chaining plugins together. The idea for plugins phase 2 is to have support for a plugin architecture that enables plugins to be separate binaries discoverable by the Kubebuilder CLI binary via user specified plugin file paths and that they are language indepedent. Some of the phase 2 high-level requirements summarized from the feature request and discussions are: 
+
+* mdbook-like structure
+    * A Plugin receives a JSON blob plugin "universe" from upstream plugins containing file "nodes", each of which contain file metadata.
+    * The plugin can add or delete nodes (nodes are not mutable), then returns the modified universe so downstream plugins can do the same.
+* Pre-generate a project file for handling complex plugin use-cases.
+* Enhanced "layout" PROJECT file key.
+    * Either an ordered list of plugin {name}/{version}, or ordered map of {name}/{version} to plugin-specific config.
+    * Order matters for certain plugins, ex. a bazel plugin would need to run after all Go files are generated.
```
Adirio, line 13: The current project file supports a comma-separated lists of plugins in the `layout` field and offers a unstructured field were plugins can store their config. While I do like the suggested approach of having an ordered map with plugin-keys to configs, that is a breaking change and would require project version 4. Project version 3 (current) is still able to support our needs so this can be implemented without having to wait for the fourth project version.

designs/plugins-phase-2.md
```diff
@@ -0,0 +1,105 @@
...
+
+The built-in plugin library seems to be a non-starter as it is more suitable for in-tree plugins rather than out-of-tree and it doesn’t offer cross language support. Hashicorp’s go plugin system seems more suitable than built-in go-plugin library as it enables cross language/platform support. However, it is more suited for long running plugins as opposed to short lived plugins and it could be overkill with the usage of protobuf.
+ 
+For the stated reasons, the proposal is to write our own plugin system that passes JSON blobs back and forth across stdin/stdout.
+The plugin specification will include type metadata to allow the potential for using other plugin libraries in the future if the need arises. From an implementation standpoint, we could create a new implementation of our plugin interface based on the type specified in the project file. We can do a type switch on the type of plugin the data is being passed to. For ex: if it is native go, then we pass the universe directly & if it is a binary wrapper, then we pass the serialized stream of JSON bytes to it.
```
Adirio, line 24: From the implementation viewpoint, instead of making the cli understand all different kind of plugins, we could code a plugin with the current interface that is in charge of doing this translation. For example, we have a JSON plugin that has the current plugin interface and creates the JSON blobs and pass them to the external plugin.

Now, imagine I'm a C++ plugin developer, I would copy this JSON plugin and develop my C++ plugin that interacts with the JSON-based exposed API.

By doing this, we would only need to support out-of-tree Go plugins, as the integration with other languages would be done by this JSON (and potentially others in a future) plugins. What do you hink about this approach? And would this approach bring the built-in plugin library back to the table?

By the way, I think that these plugins should work under demand more than sending all the data. The C++ plugin would request the `admiral_types.go` file representation and the plugin would sent it in a JSON blob. Obviously the plugin will send some general metadata first (for example, but not limitd to, the plugin config).

Adirio (commented):
Some additional considerations

camilamacedo86 (commented):
Hi @rashmigottipati, 

It is open for a while and without follow up to address the comments. Also, we have the 1.5 EP which was accepted and is implemented already and very close to be merged. See:  https://github.com/kubernetes-sigs/kubebuilder/pull/2060

This shows that the plugin phase 2.0 after that get merged and the issues in https://github.com/kubernetes-sigs/kubebuilder/issues/2016 are addressed might only be more about check the possibility of the plugins be as separate binaries that can be discovered by the Kubebuilder CLI binary via user-specified plugin file paths. 

So, could you please follow that and check if this one could be closed or updated to address what will be missing accordingly?

rashmigottipati:
Closing this as it makes more sense to create a new proposal that incorporates plugin phase 1.5 implementation and discusses design of plugins as separate binaries that can be discovered by the Kubebuilder CLI binary via user-specified plugin file paths.

PR #1802, opened 2020-11-09
✨ Add --plural flag

Adirio:
# Description
Add `--plural` flag to `kubebuilder create api` to allow irregular plural forms. Also, replace the `--resource` flag for `--plural` in `kubebuilder create webhook` for `v3-alpha` projects. This change was not done to `v2` as it would be a breaking change.

# Motivation
https://kubernetes.slack.com/archives/CAR30FCJZ/p1604861145449700

/kind feature

### DIFF
--- a/pkg/model/config/config.go
+++ b/pkg/model/config/config.go
@@ -165,6 +165,7 @@ type ResourceData struct {
 	Group   string `json:"group,omitempty"`
 	Version string `json:"version,omitempty"`
 	Kind    string `json:"kind,omitempty"`
+	Plural  string `json:"plural,omitempty"`
 
 	// API holds the API data
 	API *API `json:"api,omitempty"`
--- a/pkg/model/resource/resource.go
+++ b/pkg/model/resource/resource.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/gobuffalo/flect"
+
 	"sigs.k8s.io/kubebuilder/v2/pkg/model/config"
 )
 
@@ -61,13 +63,20 @@ type Resource struct {
 
 // Data returns the ResourceData information to check against tracked resources in the configuration file
 func (r *Resource) Data() config.ResourceData {
-	return config.ResourceData{
+	data := config.ResourceData{
 		Group:    r.Group,
 		Version:  r.Version,
 		Kind:     r.Kind,
 		API:      &r.API,
 		Webhooks: &r.Webhooks,
 	}
+
+	// Only store plural if it is irregular
+	if r.Plural != flect.Pluralize(strings.ToLower(r.Kind)) {
+		data.Plural = r.Plural
+	}
+
+	return data
 }
 
 func wrapKey(key string) string {
--- a/pkg/plugins/golang/v2/api.go
+++ b/pkg/plugins/golang/v2/api.go
@@ -111,9 +111,9 @@ func (p *createAPISubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&p.force, "force", false,
 		"attempt to create resource even if it already exists")
 	p.resource = &resource.Options{}
-	fs.StringVar(&p.resource.Kind, "kind", "", "resource Kind")
 	fs.StringVar(&p.resource.Group, "group", "", "resource Group")
 	fs.StringVar(&p.resource.Version, "version", "", "resource Version")
+	fs.StringVar(&p.resource.Kind, "kind", "", "resource Kind")
 	fs.BoolVar(&p.resource.Namespaced, "namespaced", true, "resource is namespaced")
 }
 
--- a/pkg/plugins/golang/v2/webhook.go
+++ b/pkg/plugins/golang/v2/webhook.go
@@ -67,7 +67,7 @@ func (p *createWebhookSubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&p.resource.Group, "group", "", "resource Group")
 	fs.StringVar(&p.resource.Version, "version", "", "resource Version")
 	fs.StringVar(&p.resource.Kind, "kind", "", "resource Kind")
-	fs.StringVar(&p.resource.Plural, "resource", "", "resource Resource")
+	fs.StringVar(&p.resource.Plural, "resource", "", "resource kind irregular plural form")
 
 	fs.BoolVar(&p.defaulting, "defaulting", false,
 		"if set, scaffold the defaulting webhook")
--- a/pkg/plugins/golang/v3/api.go
+++ b/pkg/plugins/golang/v3/api.go
@@ -125,9 +125,10 @@ func (p *createAPISubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&p.force, "force", false,
 		"attempt to create resource even if it already exists")
 	p.resource = &resource.Options{}
-	fs.StringVar(&p.resource.Kind, "kind", "", "resource Kind")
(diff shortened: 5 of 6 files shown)


pkg/model/config/config.go
```diff
@@ -118,13 +118,15 @@ type GVK struct {
 	Group   string `json:"group,omitempty"`
 	Version string `json:"version,omitempty"`
 	Kind    string `json:"kind,omitempty"`
+	Plural  string `json:"plural,omitempty"`
```
camilamacedo86, line 121: we need a test to ensure these changes. WDYT about add/update the testdata samples with this scenario?

Adirio: No idea about the addon scenario, but I guess it shouldn't affect as it was being passed a `Resource` which already ahs this field (it was always set with `flect.Pluralize(strings.ToLower(resource.Kind))` instead of providing the user the ability to define it).

Will add the test later.

prafull01 (changes requested):
IMHO, this actually doesn't solve the complete problem. It doesn't add the controller-gen markers which will generate the special plurals in the CRD. I copied this change locally and tried following scenario:
```
prafull@EMPID18004:~/go/src/github.com/prafull01/kubebuilder/proj$ ../bin/kubebuilder create api --group=praf --kind=Fernandez --version=v1 --plural=Fernandezes
Create Resource [y/n]
y
Create Controller [y/n]
y
api/v1/fernandez_types.go
controllers/fernandez_controller.go
Running make:
$ make
/home/prafull/go/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
```
PROJECT file:
```
prafull@EMPID18004:~/go/src/github.com/prafull01/kubebuilder/proj$ cat PROJECT 
domain: my.domain
layout: go.kubebuilder.io/v3-alpha
projectName: proj
repo: sigs.k8s.io/kubebuilder/v2
resources:
- group: praf
  kind: Fernandez
  plural: Fernandezes
  version: v1
version: 3-alpha
```
CRD file:
```
group: praf.my.domain
  names:
    kind: Fernandez
    listKind: FernandezList
    plural: fernandezs
    singular: fernandez
  scope: Namespaced
```

It should add marker ([here](https://github.com/kubernetes-sigs/controller-tools/blob/6fa696de4772fb44595d25355946254885a202df/pkg/crd/markers/crd.go#L246)) in the api/_types.go file which will actually add the plurals in the CRD as well.

camilamacedo86 (changes requested):
I am not sure if we should add a new flag. Could not it be solved by fixing the plural? Also, see: https://github.com/kubernetes-sigs/controller-tools/pull/522/files

WDYT about we raise an issue with the problematic scenarios and the use cases where the flag should be used for we discuss if we should add the flag or not?

Adirio:
I think they are 2 different topics.
1. Provide the user the possibility of choosing a custom plural form instead of auto-generating it: this PR.
1. `Plural` is not currently working: raised by #1805, discovered when testing this PR.

IMO #1805 needs to be adressed to solve the bug and then we could discuss if this PR is needed.

IMHO, we should add it at least to `v3`. Bear in mind that the default behavior (not setting the flag) will remain the same, using `flect` to compute the plural form. I coded it because it was requested in Slack (link in the PR description above).

pkg/model/resource/resource.go
```diff
@@ -55,11 +57,18 @@ type Resource struct {
...
+	gvk := config.GVK{
 		Group:   r.Group,
 		Version: r.Version,
 		Kind:    r.Kind,
 	}
+
+	// Only store plural if it is irregular
+	if r.Plural != flect.Pluralize(strings.ToLower(r.Kind)) {
+		gvk.Plural = r.Plural
+	}
```
camilamacedo86, line 69: 👍

pkg/plugin/v2/api.go
```diff
@@ -114,6 +114,7 @@ func (p *createAPISubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&p.resource.Kind, "kind", "", "resource Kind")
 	fs.StringVar(&p.resource.Group, "group", "", "resource Group")
 	fs.StringVar(&p.resource.Version, "version", "", "resource Version")
+	fs.StringVar(&p.resource.Plural, "plural", "", "resource kind irregular plural form")
```
camilamacedo86, line 117: as we spoke today shows fine we move forward with the new flag but for v3+ only

pkg/model/resource/resource.go
```diff
@@ -55,11 +57,18 @@ type Resource struct {
...
-	return config.GVK{
+	gvk := config.GVK{
 		Group:   r.Group,
 		Version: r.Version,
 		Kind:    r.Kind,
 	}
+
+	// Only store plural if it is irregular
+	if r.Plural != flect.Pluralize(strings.ToLower(r.Kind)) {
+		gvk.Plural = r.Plural
```
camilamacedo86, line 68: WDYT about we add a scenario in the testdata?

camilamacedo86 (changes requested):
after we chat in the bug triage meeting over it all shows that we all agreed that it is fine and we can address that for v3+plugin. So, for we are able to get it merged in POV we need to: 

- add a valid scenario in the testdata (generate samples)
- to do the changes only for v3+
- the mojo needs to be breaking change. The flag was changed for webhooks
- make clear in the title the changes made ( e.g add same option flag for webhooks and create api to allow customize pluralizations. 
- rebase with master

pkg/plugin/v2/api.go
Adirio: I already removed this

pkg/model/resource/resource.go
Adirio: Once #1805 gets solved, we can add further changes. I just rebased the previously existing commit and removed the `v2` related part.

Adirio:
> * add a valid scenario in the testdata (generate samples)
> * to do the changes only for v3+
> * rebase with master
Those are already done
> * the mojo needs to be breaking change. The flag was changed for webhooks
That only is true for v3 which is in alpha, do we need to consider it breaking? (previously it was changing it for v2 too, but not any more)

camilamacedo86:
HI @Adirio, 

> That only is true for v3 which is in alpha, do we need to consider it breaking? (previously it was changing it for v2 too, but not any more)

The mojo only helps us to know that something was removed and it is braking change. 
Since users can use the master branch any brake change in the code shows requires its mojo.  

Over what is missing here to get it merged, I think that we need to: 
- add testdata scenario with the flag to be able to review here that the flag will be stored as well and that the scaffolds will be done accordingly. 
- cover the code changes see: https://coveralls.io/builds/35052557/source?filename=pkg/model/resource/resource.go#L73

Adirio:
The thing is that this PR is no longer removing the `v2` flag that was removing before. I reverted the `v2` changes. In `v2` it only changes the help message but that is non-breaking. So I think ✨ should be fine.

camilamacedo86 (approved):
Would be nice add scenario to check it in the testdata before its get merged.  Otherwise, 
/approve

Adirio:
@gabbifish was assigned this topic IIRC, and there were additional changes that needed to be done to fully supporting irregular plurals. That is the reason why this is holded. The flag itself is incremental so could be added after 3.0.0 without it being breaking, but the other changes may not be non-breaking.

camilamacedo86:
Hi @Adirio, 

By checking your comment https://github.com/kubernetes-sigs/kubebuilder/pull/1802#issuecomment-745282744 I understand that it is in WIP. Could you please update its title and add in the first comment what is missing to get done?  Also, could you please consider in this todo list add a scenario to cover this change in the testdata?

PR #1803, opened 2020-11-09
 :book: proposal for new plugin to generate code ( `deploy-image.go.kubebuilder.io/v1beta1` )

camilamacedo86:
**Description**

- EP/proposal to add a new plugin to generate the code requires to address the common needs

**Motivation**

- Many operators will at the end have CRD's which will represent a deployment/pod of an image:tag. So, I am as oper user I'd like to have a feature to make this process easier. Also, see that it address common requests such as https://github.com/operator-framework/operator-sdk/pull/2158
- Reduce the learning curve since users are able to generate the basic operator following the good practices via this feature which makes more intuitive understand how the things work and know-how to improve/develop their solutions.  
- Address common questions over the basic features and how to for example; implement the tests, use envvars, RBAC permissions as watch features. 
- https://github.com/kubernetes-sigs/kubebuilder/issues/2010

### DIFF
--- a/designs/code-generate-image-plugin.md
+++ b/designs/code-generate-image-plugin.md
@@ -1,0 +1,309 @@
+---
+title: Neat-Enhancement-Idea
+authors:
+  - "@camilamacedo86"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2020-11-09
+last-updated: 2021-02-14
+status: implementable
+---
+
+# New Plugin (`deploy-image.go.kubebuilder.io/v1beta1`) to generate code
+
+## Summary
+
+This proposal defines a new plugin which allow users get the scaffold with the
+ required code to have a project that will deploy and manage an image on the cluster following the the guidelines and what have been considered as good practices.
+ 
+## Motivation
+
+The biggest part of the Kubebuilder users looking for to create a project that will at the end only deploy an image. In this way, one of the  mainly motivations of this proposal is to abstract the complexities to achieve this goal and still giving the possibility of users improve and customize their projects according to their requirements. 
+
+**Note:** This plugin will address requests that has been raised for a while and for many users in the community. Check [here](https://github.com/operator-framework/operator-sdk/pull/2158), for example, a request done in the past for the SDK project which is integrated with Kubebuidler to address the same need.
+
+### Goals
+
+- Add a new plugin to generate the code required to deploy and manage an image on the cluster
+- Promote the best practices as give example of common implementations
+- Make the process to develop  operators projects easier and more agil. 
+- Give flexibility to the users and allow them to change the code according to their needs
+- Provide examples of code implementations and of the most common features usage and reduce the learning curve
+ 
+### Non-Goals
+
+The idea of this proposal is provide a facility for the users. This plugin can be improved 
+in the future, however, this proposal just covers the basic requirements. In this way, is a non-goal
+allow extra configurations such as; scaffold the project using webhooks and the controller covered by tests. 
+
+## Proposal
+
+Add the new plugin code generate which will scaffold code implementation to deploy the image informed which would like such as; `kubebuilder create api --group=crew --version=v1 --image=myexample:0.0.1 --kind=App --plugins=deploy-image.go.kubebuilder.io/v1beta1` which will:
+
+- Add a code implementation which will do the Custom Resource reconciliation and create a Deployment resource for the `--image`;
+
+- Add an EnvVar on the manager manifest (`config/manager/manager.yaml`) which will store the image informed and shows its possibility to users:
+
+```yaml
+    ..   
+    spec:
+      containers:
+        - name: manager
+          env:
+            - name: {{ resource}}-IMAGE 
+              value: {{image:tag}}
+          image: controller:latest
+      ...
+```
+
+- Add a check into reconcile to ensure that the replicas of the deployment on cluster are equals the size defined in the CR:
+
+```go
+	// Ensure the deployment size is the same as the spec
+	size := {{ resource }}.Spec.Size
+	if *found.Spec.Replicas != size {
+		found.Spec.Replicas = &size
+		err = r.Update(ctx, found)
+		if err != nil {
+			log.Error(err, "Failed to update Deployment", "Deployment.Namespace", found.Namespace, "Deployment.Name", found.Name)
+			return ctrl.Result{}, err
+		}
+		// Spec updated - return and requeue
+		return ctrl.Result{Requeue: true}, nil
+	}
+```
+
+- Add the watch feature for the Deployment managed by the controller: 
+
+```go 
+func (r *{{ resource }}Reconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&cachev1alpha1.{{ resource }}{}).
+		Owns(&appsv1.Deployment{}).
+		Complete(r)
+}
+```
+
+- Add the RBAC permissions required for the scenario such as:
+
+```go
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
+```
+
+- A status [conditions][conditions] to allow users check that if the deployment occurred successfully or its errors
+
+- Add a [marker][markers] in the spec definition to demonstrate how to use OpenAPI schemas validation such as `+kubebuilder:validation:Minimum=1`
+
+- Add the specs on the `_types.go` to generate the CRD/CR sample with default values for `ImagePullPolicy` (`Always`), `ContainerPort` (`80`) and the `Replicas Size` (`3`) 
+
+- Add a finalizer implementation with TODO for the CR managed by the controller such as described in the SDK doc [Handle Cleanup on Deletion](https://sdk.operatorframework.io/docs/building-operators/golang/advanced-topics/#handle-cleanup-on-deletion) 
+   
+### User Stories
+
+- I am as user, would like to use a command to scaffold my common need which is deploy an image of my application, so that I do not need to know exactly how to implement it 
+
+- I am as user, would like to have a good example code base which uses the common features, so that I can easily learn its concepts and have a good start point to address my needs.  
+
+- I am as maintainer, would like to have a good example to address the common questions, so that I can easily describe how to implement the projects and/or use the common features.
+ 
+### Implementation Details/Notes/Constraints 
+
+**Example of the controller template**
+
+```go
+// +kubebuilder:rbac:groups=cache.example.com,resources={{ resource.plural }},verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=cache.example.com,resources={{ resource.plural }}/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=cache.example.com,resources={{ resource.plural }}/finalizers,verbs=update
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
+
+func (r *{{ resource }}.Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	ctx := context.Background()
+	log := r.Log.WithValues("{{ resource }}", req.NamespacedName)
+
+	// Fetch the {{ resource }} instance
+	{{ resource }} := &{{ apiimportalias }}.{{ resource }}{}
+	err := r.Get(ctx, req.NamespacedName, {{ resource }})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			log.Info("{{ resource }} resource not found. Ignoring since object must be deleted")
+			return ctrl.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		log.Error(err, "Failed to get {{ resource }}")
+		return ctrl.Result{}, err
+	}
+
+	// Check if the deployment already exists, if not create a new one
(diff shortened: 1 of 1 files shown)


pwittrock:
IIUC -- the motivation is to scaffold projects to operators, which typically run an instance of a container image.  This seems fine to me.  If it can be easily done as a plugin and we will want to support a variety of these sorts of project blueprints, then doing it as a plugin may be best.

Should we support StatefulSet and Deployments?

camilamacedo86:
Hi @pwittrock, 

> Should we support StatefulSet and Deployments?

I thought we started it by using a Deployment. However, we can improve the feature in follow-ups to allow the scaffold use statefulset or webhooks via a flag value such as `kubebuilder create code --image=myexample:0.0.1 --group=example --kind=App --version=v1 --type=[deployment|statefulset|webhook]`. WDYT?

pwittrock:
> I thought we start it by using a Deployment. However, we can improve the feature in follow-ups to support more default types of scaffolds which could be like kubebuilder create code --image=myexample:0.0.1 --group=example --kind=App --version=v1 --type=[deployment|statefulset|webhook]. WDYT?

I agree

pwittrock:
I have no objections to this proposal.

camilamacedo86:
/hold 

Until getting 2 approvals 👍 
For now, we have an LGTM from @pwittrock

estroz:
This feature looks a lot like the existing [`addon`](https://github.com/kubernetes-sigs/kubebuilder/tree/master/plugins/addon) plugin, which eventually should become a phase 2 plugin. Instead of having a separate subcommand it may be worth implementing this in the `create api` subcommand, given the similarities to `addon` and its scaffold dependencies.
```sh
kubebuilder create api --pattern=deployment-image ...
```

Adirio:
I was thinking the same as @estroz, this would only require to scaffold some content inside the `Reconcile` function of the controller and that could be made with a `CreateAPI` plugin in my opinion.

camilamacedo86:
HI @estroz, 

> This feature looks a lot like the existing addon plugin, which eventually should become a phase 2 plugin. Instead of having a separate subcommand it may be worth implementing this in the create api subcommand, given the similarities to addon and its scaffold dependencies.

IMO we should not add in the CreateAPI subcommand because:

- The generated code is not valid for other types of operators. So, Ansible/Helm in SDK, for example, which extending the CREATE API should not have the option/flags to generate the code. 
- The addon should be ported to be a plugin as well and it cannot be an option for other languages such as Ansible/Helm too. See the issue raised in the past which was requested by @DirectXMan12. : https://github.com/kubernetes-sigs/kubebuilder/issues/1543
- To keep better maintainability by not hurting concepts such as single responsibility and cohesion

Adirio:
We are not saying to implement in `CreateAPI` of `go.kubebuilder.io` plugin (the base plugin). We are saying that it can be implemented in another `CreateAPI` plugin, as with phase 2, multiple plugins can be queued once after the other. FIrst you will have `go.kubebuilder.io` scaffolding the file, then `deployment.my-domain.com` will update that file to include the code you suggest. `Addon` will be implemented the same way, probably calling the plugin `addon.kubebuilder.io`.

camilamacedo86:
HI @Adirio and @estroz, 

> We are not saying to implement in CreateAPI of go.kubebuilder.io plugin (the base plugin). We are saying that it can be implemented in another CreateAPI plugin

IMO it cannot be called `create api` since it is doing such more than that. 

> as with phase 2, multiple plugins can be queued once after the other. First you will have go.kubebuilder.io scaffolding the file, then deployment.my-domain.com will update that file to include the code you suggest. Addon will be implemented the same way, probably calling the plugin addon.kubebuilder.io.

I am totally fine with this new plugin/subcommand to be changed and/or cleanup accordingly with phase 2. I'd expected that phase 2 plugins would cover some cleanup and changes as part of its scope too. But I do not think that we need to wait for phase 2 to get it in place or try to define now how it could be after phase 2 since that still very abstract.

Adirio:
The issue I see with `kubebuilder create code` is that it will be deleting a file scaffolded previously by `kubebuilder create api`, which goes against the philosofy we have had about never deleting/removing/breaking code that the user may have written after he used a command.

camilamacedo86:
HI @Adirio, 

> The issue I see with kubebuilder create code is that it will be deleting a file scaffolded previously by kubebuilder create api, which goes against the philosofy we have had about never deleting/removing/breaking code that the user may have written after he used a command.

It is not the idea of this implementation. Note that another reason for POV we are unable to use the Create api is that the `--resource` and `controller` flags do not apply on this one as well.  The create code will never have these options since the API and controller requires to be scaffold as well.  

At the first moment, we will not allow using the command for pre-existent resources which means that by using this feature we will create the resource and controller always. Then, in the future, we could allow use it for pre-existent ones which could be one via markers to just add the code implementation to the files.

camilamacedo86:
Hi @Adirio and @estroz, 

I pushed changes on the EP based on your feedbacks, see the `open questions` section. Really thank your inputs.

camilamacedo86:
Folks, 

I will remove the hold because it has 2 oks already. 
Also, @estroz and @Adirio see that I added all your concerns and points in the open questions section of this EP. 
So, it shows ok to be reviewed and get merged. 

/hold cancel

Adirio:
I'm actually only seeing one ok (from pwittrock). I think you should wait for further apporval.

camilamacedo86:
/hold

After a chat with @Adirio I think we can push changes in this EP before get it merged.

Adirio:
Basically, we were talking that this could be probably be done as a plugin (lets say `deployment.kubebuilder.io/v1`) that wraps `go.kubebuilder.io` and performs the required modifications afterwards until phase 2 is merged. At that point the plugin would be upgraded to `deployment.kubebuilder.io/v2` which no longer wraps `go.kubebuilder.io` but uses the plugin chaining provided by phase 2 to obtains the same result.

camilamacedo86:
HI @Adirio, @estroz, 

It is updated by following your feedback and suggestions over its details in the implementation as well. 
See that the idea here is to be a plugin which will works as addon / declarative one after we merge the solution described and discussed in https://github.com/kubernetes-sigs/kubebuilder/issues/1941

designs/code-generate-image-plugin.md
```diff
@@ -0,0 +1,293 @@
...
+	ContainerPort int `json:"container-port,omitempty"`
+
+}
+```
+
+## Design Details
+
+### Test Plan
+
+To ensure this implementation a new project example should be generated in the [testdata](../testdata/) directory of the project. See the [generate_testdata.sh](../generate_testdata.sh). Also, we should use this scaffold in the [integration tests](../test/e2e/) to ensure that the data scaffolded with works on the cluster as expected.
```
Adirio, line 261:
```suggestion
To ensure this implementation a new project example should be generated in the [testdata](../testdata/) directory of the project. See the [test/testadata/generate.sh](../test/testadata/generate.sh). Also, we should use this scaffold in the [integration tests](../test/e2e/) to ensure that the data scaffolded with works on the cluster as expected.  
```

designs/code-generate-image-plugin.md
```diff
@@ -0,0 +1,293 @@
...
+- The attribute image with the value informed should be added to the resources model in the PROJECT file to let the tool know that the Resource get done with the common basic code implementation: 
+
+```yaml
+plugins:
+    generate.kubebuilder.io/v1: 
+        domain: example.io
+        group: crew
+        kind: Captain
+        version: v1
+        image: "<some-registry>/<project-name>:<tag>
```
Adirio, line 275: Needs to be a list because it may be used for several resources:
```suggestion
plugins:
    generate.kubebuilder.io/v1:
        resources:
          - domain: example.io
            group: crew
            kind: Captain
            version: v1
            image: "<some-registry>/<project-name>:<tag>
```

Adirio (changes requested):
Disclaimer: I didn't read thoroughly the implementation examples.

I agree with the goals of this EP. My two cents on the open questions at the end:

1. **Should we allow to scaffold the code for an API that is already created for the project?**
I agree that we should require both the API and the controller to be created with the same command call as both parts are required for this plugin.

2. **Should we support StatefulSet and Deployments?**
Starting with one and increasing the supported types incrementally sounds good to me.

3. **Could this feature be useful to other languages or is it just valid to Go based operators?**
This plugin would mainly modify `*_types.go` and `*_controller.go` which are obviously only relevant for go based operators. In a future, if other languaged based operators are supported (either officially or by the community) this plugin could be used as reference to create an equivalent one for their language. Therefore, it should probably be a "subdomain" of `go.kubebuilder.io`

NIT: I think we need to come up with a better name for this plugin. Generate code seems pretty imprecise in a scaffolding project. The plugin is about good practices and image deploying. Do you like `good-practices.go.kubebuilder.io` or `deploy-image.go.kubebuilder.io`?  I don't really like the first for a plugin name despite being descriptive.

camilamacedo86:
Hi @Adirio, 

Thank you for your comments and review. I adopted the name `deploy-image.go.kubebuilder.io` suggested and addressed your comments as well.

Adirio (approved):
Will let other mantainers the chance to take a look before merging, but LGTM

designs/code-generate-image-plugin.md
```diff
@@ -0,0 +1,296 @@
...
+
+## Open Questions 
+
+1. Should we allow to scaffold the code for an API that is already created for the project? 
+No, at least in the first moment to keep the simplicity.  
+
+2. Should we support StatefulSet and Deployments?
+The idea is we start it by using a Deployment. However, we can improve the feature in follow-ups to support more default types of scaffolds which could be like `kubebuilder create api --group=crew --version=v1 --image=myexample:0.0.1 --kind=App --plugins=deploy-image.go.kubebuilder.io/v1beta1 --type=[deployment|statefulset|webhook]`
+
+3. Could this feature be useful to other languages or is it just valid to Go based operators?
```
estroz, line 289: Since kubebuilder is only concerned with Go projects and the proposal discusses Go-only changes, I think the answer to this question is simply "no".

designs/code-generate-image-plugin.md
```diff
@@ -0,0 +1,296 @@
...
+To ensure this implementation a new project example should be generated in the [testdata](../testdata/) directory of the project. See the [test/testadata/generate.sh](../test/testadata/generate.sh). Also, we should use this scaffold in the [integration tests](../test/e2e/) to ensure that the data scaffolded with works on the cluster as expected.
+
+### Graduation Criteria
+
+- The new plugin will only be support `project-version=3` 
+- The attribute image with the value informed should be added to the resources model in the PROJECT file to let the tool know that the Resource get done with the common basic code implementation: 
+
+```yaml
+plugins:
+    deploy-image.go.kubebuilder.io/v1beta1:
```
estroz, line 270: This configuration implies that phase 2 is required to implement this plugin, since the 1.5 proposal doesn't discuss adding plugin configs to `plugins`. @Adirio that's correct right?

estroz (changes requested):
Overall lgtm, and I agree with the responses @Adirio has in https://github.com/kubernetes-sigs/kubebuilder/pull/1803#pullrequestreview-590251624.

A few things before I approve this proposal:
1. This proposal is dependent on at least phase 1.5, if not phase 2, plugins, so it can't be implemented or tested until the that requirement is complete.
2. Will this plugin live in the kubebuilder repo? I am ok with that, but the test matrix will start blowing up if more plugins are added. Something to be aware of. IMO I'd like to see a `kubebuilder-plugins` repo, with submodules for each plugin.

designs/code-generate-image-plugin.md
Adirio: project-version v3 already supports that field. Nothing specific was defined for plugin phase 1.5 but it doesn't prevent it either. So I would say it doesn't require phase 2 (supporting external binaries is something we still haven't properly designed AFAIK) but it definetely requires phase 1.5.

Adirio:
> 2. Will this plugin live in the kubebuilder repo? I am ok with that, but the test matrix will start blowing up if more plugins are added. Something to be aware of. IMO I'd like to see a `kubebuilder-plugins` repo, with submodules for each plugin.

Camila and me have already recognized the test matrix issue (there are already some permutations that we are not testing like plugin go/v2 with project-version v3) and adding plugins will make it even a greater issue. I already considered a separate repo (I called it `kubebuilder-community` but `kubebuilder-plugins` may be a better name.

designs/code-generate-image-plugin.md
camilamacedo86: I would like to leave the answer to clarify some questions raised. This plugin is only valid for go. However, the same idea could be used as an example.

camilamacedo86:
Hi @Adirio, @estroz, 

> Camila and me have already recognized the test matrix issue (there are already some permutations that we are not testing like plugin go/v2 with project-version v3) and adding plugins will make it even a greater issue. I already considered a separate repo (I called it kubebuilder-community but kubebuilder-plugins may be a better name)

> Will this plugin live in the kubebuilder repo? I am ok with that, but the test matrix will start blowing up if more plugins are added. Something to be aware of. IMO I'd like to see a kubebuilder-plugins repo, with submodules for each plugin.

This plugin address a need that has been requested in SDK for mone than one year. For example https://github.com/operator-framework/operator-sdk/pull/2158. Also, it addresses some discussions made with @pwittrock as well in order to make the things easier to end-users address the most common causes. So, this one would be born with "official/beta" support.   

And then, if you check the https://github.com/kubernetes-sigs/kubebuilder/issues/2016 the plugin phase 2.0 will look would be responsible to only check the possibility of the plugins be as separate binaries that can be discovered by the Kubebuilder CLI binary via user-specified plugin file paths. 

In this way, now we have not too many plugins. So, shows Premature Optimization care about. Let's think about that after we have at least 5 plugins and after the phase 2. Is it make sense? can we agree on that?

estroz (approved):
Despite the open issues, this is ok for now.

/lgtm

PR #1809, opened 2020-11-10
:bug: (go/v3-alpha) add leases.coordination.k8s.io to leader-election-role

estroz:
This PR adds `leases.coordination.k8s.io` to the leader-election-role Role under the same rule as `configmaps`. These permissions are required by default by controller-runtime's leader election, which uses "configmapsleases" as the default election type.

- go/v3-alpha: add leases.coordination.k8s.io to leader election RBAC


Adirio (approved):
/lgtm

Adirio:
Required for #1756

coderanger:
Echoing a comment from Slack so it doesn't get lost, we should probably do a follow up to this where we split this object to its own rule chunk in the role so it's not granting the non-existent perms core/leases and leases.coordination.k8s.io/configmap

PR #1813, opened 2020-11-11
:warning: (go/v3-alpha) Makefile: install kustomize and controller-gen locally

estroz:
This PR modifies the Makefile so that it installs kustomize and controller-gen to `<project root>/bin` if they do not exist. Targets will always use these binaries instead of installing global binaries, which pollutes userspace. Users can still use their preferred binary by running
```sh
mkdir -p bin
ln -sf $(which $binary) bin/$binary
```

This PR also cleans up the top-level Makefile and scripts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,6 @@
 # Makefile with some common workflow for dev, build and test
 #
 export GOPROXY?=https://proxy.golang.org/
-CONTROLLER_GEN_BIN_PATH := $(shell which controller-gen)
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -60,22 +59,18 @@ generate-testdata: ## Update/generate the testdata in $GOPATH/src/sigs.k8s.io/ku
...
 
 .PHONY: lint
-lint: golangci-lint ## Run golangci lint checks
-	@$(GOLANGCI_LINT) run
+lint: golangci-lint ## Run golangci-lint linter
+	$(GOLANGCI_LINT) run
 
-lint-fix: golangci-lint ## Run golangci lint to automatically perform fixes
-	@$(GOLANGCI_LINT) run --fix
+lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
```
Adirio, line 65:
```suggestion
.PHONY: lint-fix
lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
```

Adirio (commented):
A small suggestion

PR #1814, opened 2020-11-11
🌱 Set build variables during `make build`

### LINKED ISSUE #1401: Build-related flags for executables built from source

Adirio:
# Description
Several variables are set when releasing a kubebuilder version in order to output the correct information when calling `kubebuilder version`. But when building from the source with `make build`, this variables were not being set, and thus the `kubebuilder version` output was providing no information:
```
Version: version.Version{KubeBuilderVersion:"unknown", KubernetesVendor:"unknown", GitCommit:"$Format:%H$", BuildDate:"1970-01-01T00:00:00Z", GoOs:"unknown", GoArch:"unknown"}
```

This PR updates those variables to reflect the current working environment.

Fixes #1401

P.S.: it also cleans a bit the `Makefile`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,20 +38,25 @@ help:  ## Display this help
 
 ##@ Build
 
+LD_FLAGS=-ldflags " \
+    -X sigs.k8s.io/kubebuilder/v2/cmd.kubeBuilderVersion=$(shell git descripe --tags --dirty --broken) \
+    -X sigs.k8s.io/kubebuilder/v2/cmd.goos=$(shell go env GOOS) \
+    -X sigs.k8s.io/kubebuilder/v2/cmd.goarch=$(shell go env GOARCH) \
+    -X sigs.k8s.io/kubebuilder/v2/cmd.gitCommit=$(shell git rev-parse HEAD) \
+    -X sigs.k8s.io/kubebuilder/v2/cmd.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ') \
+    "
 .PHONY: build
 build: ## Build the project locally
-	go build -o bin/kubebuilder ./cmd
+	go build $(LD_FLAGS) -o bin/kubebuilder ./cmd
 
 .PHONY: install
-install: ## Build and install the binary with the current source code. Use it to test your changes locally.
-	make build
+install: build ## Build and install the binary with the current source code. Use it to test your changes locally.
 	cp ./bin/kubebuilder $(GOBIN)/kubebuilder
 
 ##@ Development
 
 .PHONY: generate
-generate: ## Update/generate all mock data. You should run this commands to update the mock data after your changes.
-	make generate-testdata
+generate: generate-testdata ## Update/generate all mock data. You should run this commands to update the mock data after your changes.
 	go mod tidy
 
 .PHONY: generate-testdata


Makefile
```diff
@@ -41,18 +41,22 @@ help:  ## Display this help
 
 .PHONY: build
 build: ## Build the project locally
-	go build -o bin/kubebuilder ./cmd
+	BUILD_VERSION=$(shell git descripe --tags --dirty --broken)
+	BUILD_OS=$(shell go env GOOS)
+	BUILD_ARCH=$(shell go env GOARCH)
+	BUILD_COMMIT=$(shell git rev-parse HEAD)
+	BUILD_DATE=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
+	LD_FLAGS=-ldflags "-X sigs.k8s.io/kubebuilder/v2/cmd.kubeBuilderVersion=$(BUILD_VERSION) -X sigs.k8s.io/kubebuilder/v2/cmd.goos=$(BUILD_OS) -X sigs.k8s.io/kubebuilder/v2/cmd.goarch=$(BUILD_ARCH) -X sigs.k8s.io/kubebuilder/v2/cmd.gitCommit=$(BUILD_COMMIT)-X sigs.k8s.io/kubebuilder/v2/cmd.buildDate=$(BUILD_DATE)"
```
estroz, line 49:
```suggestion
```

Makefile
```diff
@@ -67,15 +71,15 @@ lint-fix: golangci-lint ## Run golangci lint to automatically perform fixes
 	@$(GOLANGCI_LINT) run --fix
 
 golangci-lint:
```
estroz, line 73: This target is modified significantly by #1813, thoughts?

Makefile
```diff
@@ -22,9 +22,9 @@ CONTROLLER_GEN_BIN_PATH := $(shell which controller-gen)
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
-GOBIN=$(shell go env GOPATH)/bin
+	GOBIN=$(shell go env GOPATH)/bin
```
estroz, line 25: Idiomatic Makefiles define variables without indentation to denote that they are always global, so can you revert these changes?

estroz (changes requested):
Suggestions

Makefile
Adirio: It was to differentiate easier which one is saved under which conditions, but I'll revert it.

Adirio:
@estroz all suggestions applied

estroz (approved):
/lgtm

Adirio:
@estroz would you mind re-applying `lgtm` label as it had merge conflicts due to #1813 which also modified this file and had to rebase it and resolve them.

PR #1817, opened 2020-11-11
:sparkles: (go/v3-alpha) *: Replace 'docker build . -t ${IMG}' with '-t ${IMG} .'

wking:
Because folks may have `docker` aliased to `podman`, and Podman prefers options before positional arguments (containers/podman#2811):

```console
$ podman build --help | grep CONTEXT-DIRECTORY
   podman build [command options] CONTEXT-DIRECTORY | URL
```

Generated with:

```console
$ sed -i 's/ . -t \([^ ]*\)/build -t \1 ./' $(git grep -l 'docker.*build.* \. ')
```

### DIFF
--- a/pkg/plugin/v3/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugin/v3/scaffolds/internal/templates/makefile.go
@@ -121,7 +121,7 @@ generate: controller-gen
 
 # Build the docker image
 docker-build: test
-	docker build . -t ${IMG}
+	docker build -t ${IMG} .
 
 # Push the docker image
 docker-push:
--- a/testdata/project-v3-addon/Makefile
+++ b/testdata/project-v3-addon/Makefile
@@ -63,7 +63,7 @@ generate: controller-gen
 
 # Build the docker image
 docker-build: test
-	docker build . -t ${IMG}
+	docker build -t ${IMG} .
 
 # Push the docker image
 docker-push:
--- a/testdata/project-v3-multigroup/Makefile
+++ b/testdata/project-v3-multigroup/Makefile
@@ -63,7 +63,7 @@ generate: controller-gen
 
 # Build the docker image
 docker-build: test
-	docker build . -t ${IMG}
+	docker build -t ${IMG} .
 
 # Push the docker image
 docker-push:
--- a/testdata/project-v3/Makefile
+++ b/testdata/project-v3/Makefile
@@ -63,7 +63,7 @@ generate: controller-gen
 
 # Build the docker image
 docker-build: test
-	docker build . -t ${IMG}
+	docker build -t ${IMG} .
 
 # Push the docker image
 docker-push:


docs/kubebuilder_v0_v1_difference.md
```diff
@@ -13,7 +13,7 @@ Kubebuilder 1.0 adds a new flag `--project-version`, it accepts two different va
   bin/controller-manager --kubeconfig ~/.kube/config
 
   kubectl apply -f hack/sample/<resource>.yaml
-  docker build -f Dockerfile.controller . -t <image:tag>
+  docker build -f Dockerfile.controller -t <image:tag> .
```
estroz, line 16: Nit: revert this, since this is an old doc

Adirio:
Add 🌱 at the start of the title of the PR please

camilamacedo86 (commented):
Just a few nits: 
- Revert the change in the doc since this change should not affect that:
- Ensure that the commits are squashed 
- Update the title of this PR to ad the mojo (see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#pr-process).

docs/kubebuilder_v0_v1_difference.md
wking: Done with 7769cd92 -> 3b13378b

wking:
> Just a few nits...

I think I got all of these with 7769cd92 -> 3b13378b.

camilamacedo86:
Hi @wking, 

IMO it is not a bug. So, the mojo should be `✨`. Could you please update that?

estroz:
@wking this isn't really a bug so can you add `:seedling:` to your title instead?

wking:
> IMO it is not a bug. So, shows that the mojo should be ✨. Could you please update that?

This moves kubebuilder Makefile rules from "silently build images without tagging them" to "correctly tag built images" on older `podman`, and from "error out of attempted builds" to "correctly build and tag images" for newer `podman`.  That seems like a bug to me, although I agree that the lack of a Docker-CLI specification distinct from the `docker` implementation makes it unclear whether fixing broken `podman` support is a kubebuilder bug or feature.  But moved to :sparkles: with 3b13378b -> 1ef003d1.

docs/book/src/multiversion-tutorial/testdata/project/Makefile
```diff
@@ -63,7 +63,7 @@ generate: controller-gen
 
 # Build the docker image
 docker-build: test
-	docker build . -t ${IMG}
+	docker build -t ${IMG} .
```
camilamacedo86, line 66: It is not a breaking change. However, just to do a quick poll. 
Do you think that it should be addressed to v2+ or only v3-alpha plugin @estroz @Adirio @gabbifish ?

Adirio: I would say it is a breaking change (v2 projects scaffolded before this PR will have a different Makefile), and thsu, only to `v3-alpha`

camilamacedo86: IMO: `v2 projects scaffolded before this PR will have a different Makefile` does not characterize a breaking change since it will not break the projects scaffolds before this change. However, we have been very restrictive regards changes in v2 and it is not a critical bug fix for the tool which might is a good argumentation to be only v3+. 

+1.

camilamacedo86: Hi @wking, 

Could you please revert the changes in `pkg/plugin/v2` and run `make generate` and push it again. I think would be better to address it only for v3-alpha as well. Could you also add (go/v3-alpha) on the title to make it clear enough? Then, I think would be fine to get this one merged.

wking: > ...v2 projects scaffolded before this PR will have a different Makefile...

Ideally there would be a flow to easily re-scaffold projects to pull in minor stuff like this, instead of delaying all of that and batching it in major-version bumps.  Is there a ticket around that I can subscribe to?  If not, should I file one?

> Could you please revert the changes in `pkg/plugin/v2` and...

I think I've dropped the v2 stuff with 1ef003d1 -> cced240f.

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -63,7 +63,7 @@ generate: controller-gen
 
 # Build the docker image
 docker-build: test
-	docker build . -t ${IMG}
+	docker build -t ${IMG} .
```
Adirio, line 66: This belongs to `v2`, so please discard this change.

docs/book/src/multiversion-tutorial/testdata/project/Makefile
Adirio, line 66: Same, `v2`, discard.

Adirio (changes requested):

docs/book/src/multiversion-tutorial/testdata/project/Makefile
Adirio: The philosophy we have had until now is that once scaffolded, kubebuilder doesn't own the file, user can modify them (actually, they will modify files, kubebuilder provides the scheleton but you need to code parts) so once scaffolded we are not allowed to delete a file, and thus, we can't re-scaffold it later. The only exception is that some code is injected into some files with some special markers.

wking: > The philosophy we have had until now is that once scaffolded, kubebuilder doesn't own the file, user can modify them...

That offloads a lot of maintenance cost to folks maintaining the consuming project, doesn't it?  E.g. we land this v2 fix, and now every kubebuilder-scaffolded project is on their own to discover, diagnose, and manually patch their own project to pick up the fix for `podman` compatibility.  It seems like there should be a way to maintain kubebuilder input for portions of the scaffolding, and make it easy for downstream maintainers to ask "do the kubebuilder maintainers have any improvements to suggest for my project?", see a set of suggestions, and then accept or reject them as they see fit.  For one approach to this (tracking the scaffolding in a separate branch which is periodically merged into downstream code), see [here][1].  I expect there are many alternative approaches which would also allow for downstream maintainers to delegate scaffolding maintenance to the kubebuilder maintainers.

[1]: https://github.com/openshift/cincinnati-operator/pull/79#issuecomment-728355658

docs/book/src/cronjob-tutorial/testdata/project/Makefile
wking: Done with cced240f -> ccb11276.

docs/book/src/multiversion-tutorial/testdata/project/Makefile
wking: Done with cced240f -> ccb11276.

Adirio (approved):
/lgtm

Issue #1798, opened 2020-11-07
Change module to v3 (can be done only when we will cut the 3.0.0-alpha release)


Adirio:
- [ ] Update module name in `go.mod` to `sigs.k8s.io/kubebuilder/v3`
- [ ] Update the build command in `generate_testdata.sh` to `go build -o ./bin/kubebuilder sigs.k8s.io/kubebuilder/v3/cmd`
- [ ] Update imports to v3: `sigs.k8s.io/kubebuilder/v2` -> `sigs.k8s.io/kubebuilder/v3`
- [ ] Update `-ldflags` paths to v3 in `common.sh` and `build/.goreleaser.yml`

Equivalent v2 PR: #1792

Adirio:
@camilamacedo86 could you add this to the `v3.0.0` milestone?

camilamacedo86:
Besides we are unable to push a patch release from master the import of the pkg will still v2.3.1-abc-123. See; https://github.com/operator-framework/operator-sdk/blob/master/go.mod#L37

camilamacedo86:
Hi @Adirio and @estroz, 

We should get it done before release the alpha one. However, I think we can move forward here now as well.

camilamacedo86:
Done.

PR #1825, opened 2020-11-16
:seedling: update module v2/ to v3/

camilamacedo86:
**Description**

Bump v3/ module to prepare the master for the next release (3.0.0)

**Motivation**
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/1798


Adirio:
@camilamacedo86 I think this should be the last thing to get merged before the v3 release, and trust me, rebasing this PR is going to be a pain. It is better to wait until the end and do a search and replace that try to keep this PR up to date. I suffered it with the v2 equivalent.

camilamacedo86:
Hi @Adirio, 

over: https://github.com/kubernetes-sigs/kubebuilder/pull/1825 IMO we should merge this one and we should not actually bump v2/ at all. See that: 

Could we really do a patch release from master? 
- we remove the v1 from master
- we changed the project file and introduced the plugin layout 

In this way, if we do a release now. Will the CLI work with the same projects that it was able to work before the changes made?

camilamacedo86:
Close this one because besides we are unable to push a patch release from master the import of the pkg will still v2.3.1-abc-123.

PR #1828, opened 2020-11-17
⚠ pkg/cli overhaul

### LINKED ISSUE #1755: CLI: it would be helpful to skip PROJECT reads for extra commands
It would be helpful to skip `PROJECT` file reads for extra commands. `kubebuilder version` will throw an error if an errant `PROJECT` file exists. I do not expect this for `version` or other extra commands.

For example,

```
$ kubebuilder version
2020/10/29 12:19:36 failed to read the configuration file: read PROJECT: is a directory
```

I expected the `version` and extra commands to ignore the `PROJECT` file/directory. 
```
$ kubebuilder version
Version: version.Version{KubeBuilderVersion:"2.3.1", KubernetesVendor:"1.16.4", GitCommit:"8b53abeb4280186e494b726edf8f54ca7aa64a49", BuildDate:"2020-03-26T16:42:00Z", GoOs:"unknown", GoArch:"unknown"}
```

To reproduce:

```
mkdir PROJECT
kubebuilder version
```

This also affects the `operator-sdk` because we use the `kubebuilder` cli. I've tracked down some of the interaction.  Starting from [`cli.New`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/cli.go#L86) which always calls [`c.initialize()`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/cli.go#L99). This fails with the following error: [`FATA[0000] failed to read config: read PROJECT: is a directory`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/cli.go#L201) because of the errant `PROJECT` file. With most commands it is useful and/or required to read the `PROJECT` file but for something like `version` or other extra commands I believe it isn't required. 

/kind bug

Adirio:
# Description
pkg/cli: mayor overhaul
pkg/plugin: add `Validate` and `ValidateKey` helpers instead of `ValidateName`. The logic of this helpers was in `pkg/cli/plugins.go` previously.

# Motivation
- Prepare for plugin phase 2: #1801
- Develop tests for smaller pieces of code to improve coverage and delimit better future errors
- Fixes #1755

# TODO
- [x] Finish tests
- [x] Make commands default to one that errors if no plugin provides it: this way, the cli can still be run for those commands that do not require any plugin like `version`.

# NOTE TO REVIEWERS
Side by side comparisson of `pkg/cli/cli.go` and `pkg/cli/cli_test.go` does **not** provide any value. It will be easier to clone this branch and review the files locally together with the new `pkg/cli/options.go` and `pkg/cli/options_test.go` files.


Adirio:
@estroz and @rashmigottipati this is the cli refactoring I told you about some days ago

pkg/cli/cli.go
```diff
@@ -32,323 +32,342 @@ import (
...
-`
+	noticeColor = "\033[1;36m%s\033[0m"
 
 	projectVersionFlag = "project-version"
-	helpFlag           = "help"
 	pluginsFlag        = "plugins"
 )
 
+// equalStringSlice checks if two string slices are equal.
+func equalStringSlice(a, b []string) bool {
```
estroz, line 42: Does order matter here?

cmd/main.go
```diff
@@ -28,11 +29,15 @@ func main() {
 	c, err := cli.New(
 		cli.WithCommandName("kubebuilder"),
 		cli.WithVersion(versionString()),
+		cli.WithDefaultProjectVersion(config.Version2),
```
camilamacedo86, line 32: IMO it needs to be 3

cmd/main.go
```diff
@@ -28,11 +29,15 @@ func main() {
...
 		cli.WithPlugins(
 			&pluginv2.Plugin{},
 			&pluginv3.Plugin{},
 		),
-		cli.WithDefaultPlugins(
+		cli.WithDefaultPlugins(config.Version2,
+			&pluginv2.Plugin{},
+		),
+		cli.WithDefaultPlugins(config.Version3Alpha,
 			&pluginv2.Plugin{},
```
camilamacedo86, line 41: how we can have 2 default plugins?

pkg/cli/cli.go
Adirio: Yes, which plugins gets executed first is important.

cmd/main.go
Adirio: Default plugin for different project versions.
`kubebuilder init --project-version 2` will use the first set and `kubebuilder init --project-version 3-alpha` will use the second set.

cmd/main.go
Adirio: That will depend if `v3.0.0-alpha1` is released before or after this is merged. For now I kept the current behavior, if `v3.0.0-alpha1` is released and it swaps the defult version to `3-alpha` before this gets merged, I will update accordingly.

pkg/plugin/interfaces.go
```diff
@@ -119,8 +139,9 @@ type EditSubcommand interface {
...
 type Full interface {
-	Init
-	CreateAPI
-	CreateWebhook
-	Edit
+	Plugin
+	InitGetter
+	CreateAPIGetter
+	CreateWebhookGetter
+	EditGetter
```
Adirio, line 146: I'm not sure what changed but this was giving me duplicate methods error as all of the wrapped interfaces were wrapping `Plugin` themselves. So I basically made interfaces that didn't wrap `Plugin` as an intermediate step.

Adirio:
@estroz @camilamacedo86 @DirectXMan12 This PR should be ready now.

Coverage is reporting a 0.4% global reduction. However, the main files invoved have been extensively tested:
- `options.go`: 100% of statements covered.
- `cli.go`: 94.6% of statements covered. The only statements that are not covered are direct file system interactions (parsing the values from the read file is covered, but reading the file is not covered as there is no file to read) (6 lines) and the lines that require to execute the command in order to be covered (6 lines).

jmrodri:
@Adirio 

With this patch applied it works:
```
$ mkdir PROJECT
$ /home/jesusr/dev/kubebuilder/bin/kubebuilder version
Version: main.version{KubeBuilderVersion:"v2.3.1-391-gf260cc8e", KubernetesVendor:"unknown", GitCommit:"f260cc8ea59e64b444c4b2cf0a40da2fb61f635e", BuildDate:"2020-11-23T20:09:54Z", GoOs:"linux", GoArch:"amd64"}
```

cmd/main.go
estroz: Actually the default project version is 3-alpha right now:
```console
$ mkdir test
$ go mod init test && go get sigs.k8s.io/kubebuilder@master && rm go.mod go.sum
$ kubebuilder init
$ cat PROJECT 
domain: my.domain
layout: go.kubebuilder.io/v2
projectName: blah
repo: sigs.k8s.io/kubebuilder/v2
version: 3-alpha
```

pkg/plugin/interfaces.go
estroz: I'm not seeing any issues even after embedding `Plugin` in `Full`.

Adirio: Strange, go version? I found this error in #1869 too when I run `go test`, I think I'm using 1.15

estroz: That's strange. Composite interfaces with duplicate methods were [added](https://medium.com/@meeusdylan/changes-to-interfaces-in-go1-14-821ab533f62) in go v1.14.

Adirio: Will revert this change

cmd/main.go
Adirio: Ok, changed

Adirio:
@camilamacedo86 @estroz this just needs a `lgtm` label to be merged

pwittrock:
Force merging this because the broken check doesn't make sense.  Lets follow up with a better fix.

PR #1830, opened 2020-11-17
✨  Allow Downstream Root Commands to add custom descriptions

anmol372:
Added 4 Options to configure command descriptions.
Downstream Projects may want to change the preset descriptions and examples to something more relevant to their projects so I am proposing the following Options to configure those by calling 
`WithRootCommandConfig(cmdCfg)`
The fields of `cmdCfg` can be set by using `RootCommandConfig struct` added in the cli pkg

### DIFF
--- a/docs/book/src/reference/cli-plugins.md
+++ b/docs/book/src/reference/cli-plugins.md
@@ -171,8 +171,11 @@ import (
 // Go plugin version unless '--plugins' or 'layout' specify
 // one of the other plugins.
 func main() {
+  rootCmdCfg := cli.RootCommandConfig{
+        CommandName: "controller-builder",
+	}
   c, err := cli.New(
-    cli.WithCommandName("controller-builder"),
+    cli.WithRootCommandConfig(rootCmdCfg),
     cli.WithDefaultProjectVersion("3"),
     cli.WithExtraCommands(newCustomCobraCmd()),
     cli.WithPlugins(
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -39,6 +39,49 @@ const (
 	pluginsFlag        = "plugins"
 
 	noPluginError = "invalid config file please verify that the version and layout fields are set and valid"
+
+	defaultCommandName      = "kubebuilder"
+	defaultShortDescription = "Development kit for building Kubernetes extensions and tools."
+	defaultLongDescription  = `Development kit for building Kubernetes extensions and tools.
+	
+Provides libraries and tools to create new projects, APIs and controllers.
+Includes tools for packaging artifacts into an installer container.
+			
+Typical project lifecycle:
+			
+- initialize a project:
+			
+  placeHolder init --domain example.com --license apache2 --owner "The Kubernetes authors"
+			
+- create one or more a new resource APIs and add your code to them:
+		
+  placeHolder create api --group <group> --version <version> --kind <Kind>
+		
+Create resource will prompt the user for if it should scaffold the Resource and / or Controller. To only
+scaffold a Controller for an existing Resource, select "n" for Resource. To only define
+the schema for a Resource without writing a Controller, select "n" for Controller.
+			
+After the scaffold is written, api will run make on the project.
+`
+	defaultExample = `
+  # Initialize your project
+  placeHolder init --domain example.com --license apache2 --owner "The Kubernetes authors"
+		  
+  # Create a frigates API with Group: ship, Version: v1beta1 and Kind: Frigate
+  placeHolder create api --group ship --version v1beta1 --kind Frigate
+
+  # Edit the API Scheme
+  nano api/v1beta1/frigate_types.go
+
+  # Edit the Controller
+  nano controllers/frigate_controller.go
+		
+  # Install CRDs into the Kubernetes cluster using kubectl apply
+  make install
+		
+  # Regenerate code and run against the Kubernetes cluster configured by ~/.kube/config
+  make run
+  `
 )
 
 // equalStringSlice checks if two string slices are equal.
@@ -65,13 +108,21 @@ type CLI interface {
 	Run() error
 }
 
+//RootCommandConfig returns the configuration for root cobra.Command
+type RootCommandConfig struct {
+	CommandName string
+	Short       string
+	Long        string
+	Example     string
+}
+
 // cli defines the command line structure and interfaces that are used to
 // scaffold kubebuilder project files.
 type cli struct { //nolint:maligned
 	/* Fields set by Option */
 
-	// Root command name. It is injected downstream to provide correct help, usage, examples and errors.
-	commandName string
+	//Root command config. It is injected downstream to provide correct help, usage, examples and errors.
+	cmdCfg RootCommandConfig
 	// CLI version string.
 	version string
 	// Default project version in case none is provided and a config file can't be found.
@@ -147,9 +198,16 @@ func New(opts ...Option) (CLI, error) {
 // newCLI creates a default cli instance and applies the provided options.
 // It is as a separate function for test purposes.
 func newCLI(opts ...Option) (*cli, error) {
+	//Get default command descriptions
+	shortDescription, longDescription, example := buildDefaultDescriptors(defaultCommandName)
 	// Default cli options.
 	c := &cli{
-		commandName:           "kubebuilder",
+		cmdCfg: RootCommandConfig{
+			CommandName: defaultCommandName,
+			Short:       shortDescription,
+			Long:        longDescription,
+			Example:     example,
+		},
 		defaultProjectVersion: internalconfig.DefaultVersion,
 		defaultPlugins:        make(map[string][]string),
 		plugins:               make(map[string]plugin.Plugin),
@@ -163,6 +221,15 @@ func newCLI(opts ...Option) (*cli, error) {
 	}
 
 	return c, nil
+}
+
+// buildDefaultDescriptors replaces placeholder in default descriptors
+func buildDefaultDescriptors(commandName string) (string, string, string) {
+	replacer := strings.NewReplacer("placeHolder", commandName)
+	short := replacer.Replace(defaultShortDescription)
+	long := replacer.Replace(defaultLongDescription)
+	example := replacer.Replace(defaultExample)
+	return short, long, example
 }
 
 // getInfoFromFlags obtains the project version and plugin keys from flags.
@@ -402,8 +469,17 @@ func (c *cli) resolve() error {
 // buildRootCmd returns a root command with a subcommand tree reflecting the
 // current project's state.
 func (c cli) buildRootCmd() *cobra.Command {
-	rootCmd := c.defaultCommand()
-
+	rootCmd := &cobra.Command{
+		Use:     c.cmdCfg.CommandName,
+		Short:   c.cmdCfg.Short,
+		Long:    c.cmdCfg.Long,
+		Example: c.cmdCfg.Example,
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := cmd.Help(); err != nil {
+				log.Fatal(err)
+			}
+		},
+	}
 	// kubebuilder completion
 	// Only add completion if requested
 	if c.completionCommand {
@@ -432,61 +508,6 @@ func (c cli) buildRootCmd() *cobra.Command {
 	}
 
 	return rootCmd
-}
-
-// defaultCommand returns the root command without its subcommands.
-func (c cli) defaultCommand() *cobra.Command {
-	return &cobra.Command{
-		Use:   c.commandName,
-		Short: "Development kit for building Kubernetes extensions and tools.",
-		Long: fmt.Sprintf(`Development kit for building Kubernetes extensions and tools.
-
-Provides libraries and tools to create new projects, APIs and controllers.
-Includes tools for packaging artifacts into an installer container.
-
-Typical project lifecycle:
-
-- initialize a project:
-
-  %[1]s init --domain example.com --license apache2 --owner "The Kubernetes authors"
-
-- create one or more a new resource APIs and add your code to them:
-
-  %[1]s create api --group <group> --version <version> --kind <Kind>
-
-Create resource will prompt the user for if it should scaffold the Resource and / or Controller. To only
-scaffold a Controller for an existing Resource, select "n" for Resource. To only define
-the schema for a Resource without writing a Controller, select "n" for Controller.
-
-After the scaffold is written, api will run make on the project.
-`,
-			c.commandName),
-		Example: fmt.Sprintf(`
-  # Initialize your project
-  %[1]s init --domain example.com --license apache2 --owner "The Kubernetes authors"
-
-  # Create a frigates API with Group: ship, Version: v1beta1 and Kind: Frigate
-  %[1]s create api --group ship --version v1beta1 --kind Frigate
-
-  # Edit the API Scheme
-  nano api/v1beta1/frigate_types.go
-
-  # Edit the Controller
-  nano controllers/frigate_controller.go
-
-  # Install CRDs into the Kubernetes cluster using kubectl apply
-  make install
-
-  # Regenerate code and run against the Kubernetes cluster configured by ~/.kube/config
-  make run
-`,
-			c.commandName),
-		Run: func(cmd *cobra.Command, args []string) {
-			if err := cmd.Help(); err != nil {
-				log.Fatal(err)
-			}
-		},
-	}
 }
 
 // Run implements CLI.Run.
--- a/pkg/cli/cli_test.go
+++ b/pkg/cli/cli_test.go
@@ -562,7 +562,6 @@ var _ = Describe("CLI", func() {
 				Expect(err).To(HaveOccurred())
 			})
 		})
-
 		When("an invalid plugin key is set", func() {
 			It("should fail", func() {
 				c = &cli{
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -26,8 +26,11 @@ import (
 )
 
 func main() {
+	rootCmdCfg := cli.RootCommandConfig{
+		CommandName: "kubebuilder",
+	}
 	c, err := cli.New(
-		cli.WithCommandName("kubebuilder"),
+		cli.WithRootCommandConfig(rootCmdCfg),
 		cli.WithVersion(versionString()),
 		cli.WithDefaultProjectVersion(config.Version3Alpha),
 		cli.WithPlugins(
--- a/pkg/cli/api.go
+++ b/pkg/cli/api.go
@@ -44,7 +44,7 @@ func (c cli) newCreateAPICmd() *cobra.Command {
 
 func (c cli) newAPIContext() plugin.Context {
 	return plugin.Context{
-		CommandName: c.commandName,
+		CommandName: c.cmdCfg.CommandName,
 		Description: `Scaffold a Kubernetes API.
 `,
 	}
--- a/pkg/cli/completion.go
+++ b/pkg/cli/completion.go
@@ -35,7 +35,7 @@ Linux:
   $ %[1]s completion bash > /etc/bash_completion.d/%[1]s
 MacOS:
   $ %[1]s completion bash > /usr/local/etc/bash_completion.d/%[1]s
-`, c.commandName),
+`, c.cmdCfg.CommandName),
 		RunE: func(cmd *cobra.Command, cmdArgs []string) error {
 			return cmd.Root().GenBashCompletion(os.Stdout)
 		},
@@ -54,7 +54,7 @@ $ echo "autoload -U compinit; compinit" >> ~/.zshrc
 $ %[1]s completion zsh > "${fpath[1]}/_%[1]s"
 
 # You will need to start a new shell for this setup to take effect.
-`, c.commandName),
+`, c.cmdCfg.CommandName),
 		RunE: func(cmd *cobra.Command, cmdArgs []string) error {
 			return cmd.Root().GenZshCompletion(os.Stdout)
 		},
@@ -98,7 +98,7 @@ func (c cli) newCompletionCmd() *cobra.Command {
 		Long: fmt.Sprintf(`Output shell completion code for the specified shell.
 The shell code must be evaluated to provide interactive completion of %[1]s commands.
 Detailed instructions on how to do this for each shell are provided in their own commands.
-`, c.commandName),
+`, c.cmdCfg.CommandName),
 	}
 	cmd.AddCommand(c.newBashCmd())
 	cmd.AddCommand(c.newZshCmd())
--- a/pkg/cli/edit.go
+++ b/pkg/cli/edit.go
@@ -44,7 +44,7 @@ func (c cli) newEditCmd() *cobra.Command {
 
 func (c cli) newEditContext() plugin.Context {
 	return plugin.Context{
-		CommandName: c.commandName,
+		CommandName: c.cmdCfg.CommandName,
 		Description: `Edit the project configuration.
 `,
 	}
--- a/pkg/cli/init.go
+++ b/pkg/cli/init.go
@@ -63,7 +63,7 @@ func (c cli) newInitCmd() *cobra.Command {
 
 func (c cli) newInitContext() plugin.Context {
 	return plugin.Context{
-		CommandName: c.commandName,
+		CommandName: c.cmdCfg.CommandName,
 		Description: `Initialize a new project.
 
 For further help about a specific project version, set --project-version.
@@ -79,7 +79,7 @@ func (c cli) getInitHelpExamples() string {
   %[1]s init --project-version=%[2]s -h
 
 `,
-			c.commandName, version)
+			c.cmdCfg.CommandName, version)
 		sb.WriteString(rendered)
 	}
 	return strings.TrimSuffix(sb.String(), "\n\n")
--- a/pkg/cli/options.go
+++ b/pkg/cli/options.go
@@ -28,10 +28,28 @@ import (
 // Option is a function that can configure the cli
 type Option func(*cli) error
 
-// WithCommandName is an Option that sets the cli's root command name.
-func WithCommandName(name string) Option {
+// WithRootCommandConfig configures a cli's root command.
+func WithRootCommandConfig(cfg RootCommandConfig) Option {
 	return func(c *cli) error {
-		c.commandName = name
+		if cfg.CommandName != "" {
+			c.cmdCfg.CommandName = cfg.CommandName
+		}
+		shortDescription, longDescription, example := buildDefaultDescriptors(c.cmdCfg.CommandName)
+		if cfg.Short != "" {
+			c.cmdCfg.Short = cfg.Short
+		} else {
+			c.cmdCfg.Short = shortDescription
+		}
+		if cfg.Long != "" {
+			c.cmdCfg.Long = cfg.Long
+		} else {
+			c.cmdCfg.Long = longDescription
+		}
+		if cfg.Example != "" {
+			c.cmdCfg.Example = cfg.Example
+		} else {
+			c.cmdCfg.Example = example
+		}
 		return nil
 	}
 }
--- a/pkg/cli/options_test.go
+++ b/pkg/cli/options_test.go
@@ -45,13 +45,46 @@ var _ = Describe("CLI options", func() {
 		np4 = newMockPlugin(pluginName, pluginVersion, "a")
 	)
 
-	Context("WithCommandName", func() {
-		It("should use provided command name", func() {
-			commandName := "other-command"
-			c, err = newCLI(WithCommandName(commandName))
+	Context("WithRootCommandConfig", func() {
+		It("should use the default command name and descriptions", func() {
+			c, err = newCLI()
+			shortDescription, longDescription, example := buildDefaultDescriptors(defaultCommandName)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(c).NotTo(BeNil())
-			Expect(c.commandName).To(Equal(commandName))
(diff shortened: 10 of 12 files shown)


pkg/cli/cli.go
```diff
@@ -193,6 +212,59 @@ func WithCompletion(c *cli) error {
...
+		- create one or more a new resource APIs and add your code to them:
+		
+		%s create api --group <group> --version <version> --kind <Kind>
+		
+		Create resource will prompt the user for if it should scaffold the Resource and / or Controller. To only
+		scaffold a Controller for an existing Resource, select "n" for Resource. To only define
+		the schema for a Resource without writing a Controller, select "n" for Controller.
+		
+		After the scaffold is written, api will run make on the project.
+		`, c.commandName, c.commandName)
```
Adirio, line 241: Don't add tabs to multi-line strings, and `%[1]s` form is clearer than having to repeat the argument to `fmt.Sprintf`:
```suggestion
		c.long = fmt.Sprintf(`Development kit for building Kubernetes extensions and tools.

Provides libraries and tools to create new projects, APIs and controllers.
Includes tools for packaging artifacts into an installer container.

Typical project lifecycle:

- Initialize a project:

%[1]s init --domain example.com --license apache2 --owner "The Kubernetes authors"

- Create one or more a new resource APIs and add your code to them:

%[1]s create api --group <group> --version <version> --kind <Kind>

Create resource will prompt the user for if it should scaffold the Resource and / or Controller.
To only scaffold a Controller for an existing Resource, select "n" for Resource. To only define
the schema for a Resource without writing a Controller, select "n" for Controller.

After the scaffold is written, api will run make on the project.
`,
			c.commandName)
```

pkg/cli/cli.go
```diff
@@ -193,6 +212,59 @@ func WithCompletion(c *cli) error {
...
+		# Edit the Controller
+		nano controllers/frigate_controller.go
+	  
+		# Install CRDs into the Kubernetes cluster using kubectl apply
+		make install
+	  
+		# Regenerate code and run against the Kubernetes cluster configured by ~/.kube/config
+		make run
+	  `,
+			c.commandName, c.commandName)
```
Adirio, line 263: Same:
```suggestion
		c.example = fmt.Sprintf(`
# Initialize your project
%[1]s init --domain example.com --license apache2 --owner "The Kubernetes authors"

# Create a frigates API with Group: ship, Version: v1beta1 and Kind: Frigate
%[1]s create api --group ship --version v1beta1 --kind Frigate

# Edit the API Scheme
nano api/v1beta1/frigate_types.go

# Edit the Controller
nano controllers/frigate_controller.go

# Install CRDs into the Kubernetes cluster using kubectl apply
make install

# Regenerate code and run against the Kubernetes cluster configured by ~/.kube/config
make run
`,
			c.commandName)
```

pkg/cli/cli.go
```diff
@@ -193,6 +212,59 @@ func WithCompletion(c *cli) error {
 	return nil
 }
 
+//set default descriptions for root command
+func (c *cli) setRootCommandDefaults() error {
```
Adirio, line 216: Can you just do this checks in `func (c cli) defaultCommand() *cobra.Command`? I don't see the need to do it that early and I feel like you loose track of those fields by setting them at the start and using them at the end.

Adirio (changes requested):
I'm not sure if a single `Option` that sets the three of them is the best approach. I feel that we want one option per field. So I would say either just one `Option` per field or one `Option` per field and a forth one that sets the three at once.

anmol372:
@Adirio, I have gone with the four `Option`'s approach and reverted the default descriptions to their pre tab formatting.

Adirio (approved):
/lgtm

You can resolve the conversations to collapse them.

Adirio:
Retesting the failing test (it itmed out and the limit has been raised to 20 from the previous 15 minutes, which used to be enough but it depends on the workload of prow from several repositories). I expect it to pass it.
/retest

About the coverage decrease, could you add tests for these new options? Check how the `WithCommandName` test is performed in `pkg/cli/cli_test.go` and create a new context for the new provided `Option`s that check equivalent fields.

https://github.com/kubernetes-sigs/kubebuilder/blob/9663fb3735c2f1fde4966884778565c8bec48152/pkg/cli/cli_test.go#L280-L292

camilamacedo86:
Hi @anmol372, 

The new methods need to be converted with unit tests. See:  coverage/coveralls — Coverage decreased (-0.7%) to 71.827% and https://coveralls.io/builds/35041416/source?filename=pkg/cli/cli.go.

estroz (changes requested):
There's probably a more generic/extensible way to do this like

```diff
type cli struct {
-	// Root command name. Can be injected downstream.
-	commandName string
-	// Root command short description, Can be injected downstream.
-	short string
-	// Root command long description, Can be injected downstream.
-	long string
-	//Root command example, Can be injected downstream.
-	example string
+   // cmdCfg configures the root cobra.Command.
+   cmdCfg RootCommandConfig

    ...
}
```

```go
type RootCommandConfig struct {
    CommandName string
    Short string
    Long string
    Example string
}

// WithRootCommandConfig configures a CLI's root command.
func WithRootCommandConfig(cfg RootCommandConfig) Option {
	return func(c *cli) error {
		c.cmdCfg = cfg
		return nil
	}
}

func (c cli) defaultCommand() *cobra.Command {
    cmd := &cobra.Command{}

	// Prefer the WithCommandName() option for backwards-compat.
	if c.commandName != "" {
		cmd.Use = c.commandName
	} else {
        cmd.Use = c.cmdCfg.CommandName
    }
    cmd.Short = c.cmdCfg.Short
	if cmd.Short == "" {
		cmd.Short = "Development kit for building ..."
    }
    cmd.Long = c.cmdCfg.Long
	if cmd.Long == "" {
		cmd.Long = fmt.Sprintf(`Development kit for building ...`)
	}
    cmd.Example = c.cmdCfg.Example
	if cmd.Example == "" {
		cmd.Example = fmt.Sprintf(`...`)
	}

    return cmd
}
```

A solve-all solution would set `type RootCommandConfig cobra.Command` and use that as the base root command, but that may expose too much of the underlying CLI code and make users think they can set certain fields that will actually be overwritten/modified by other bits of the `cli` package.

Adirio:
> A solve-all solution would set `type RootCommandConfig cobra.Command` and use that as the base root command, but that may expose too much of the underlying CLI code and make users think they can set certain fields that will actually be overwritten/modified by other bits of the `cli` package.

I think that not showing thqat we are using cobra underneath as much as we could was a way to allow to remove that dependency in a future in case we wanted to.

Related to the `RootCommandConfig` struct, I think that using the `Option` declarative pattern, adding another layer of indirection would not solve anything and just adds complexity, but thats just personal preference.

anmol372:
@estroz , @Adirio ,
I have updated the code as per review suggestions and I removed the `WithCommandName()` option completely, since the package has not been released uptill now, this seemed like a cleaner approach if we do add `RootCommandConfig` struct indirection and it would allow us to add more cobra attributes if/when required.
Please let me know if you think this is OK.

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
```diff
@@ -170,8 +170,14 @@ Example Operator SDK main.go:
...
 func main() {
+    rootCmdCfg := cli.RootCommandConfig{
+        CommandName: "operator-sdk",
+        //Short: "Short help text for operator-sdk",
+        //Long: "Longer help text for operator-sdk",
+        //Example: "Example usage of operator-sdk",
+	}
 	c, err := cli.New(
-		cli.WithCommandName("operator-sdk"),
+		cli.WithRootCommandConfig("rootCmdCfg"),
```
Adirio, line 180: It should be the struct not a string:
```suggestion
		cli.WithRootCommandConfig(rootCmdCfg),
```

pkg/cli/cli_test.go
```diff
@@ -276,18 +276,43 @@ var _ = Describe("CLI", func() {
...
+			It("should use the provided command name and default descriptions", func() {
+				cfg := RootCommandConfig{
+					CommandName: "other-command",
+				}
+				c, err = New(
+					WithRootCommandConfig(cfg),
+					WithDefaultPlugins(pluginAV1),
+					WithPlugins(allPlugins...),
+				)
+				sh, lo, ex := c.(*cli).getRootCommandDefaults()
```
Adirio, line 289:
```suggestion
```

pkg/cli/cli_test.go
```diff
@@ -276,18 +276,43 @@ var _ = Describe("CLI", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+				Expect(c).NotTo(BeNil())
+				Expect(c.(*cli).cmd.Use).To(Equal(cfg.CommandName))
+				Expect(c.(*cli).cmd.Short).To(Equal(sh))
+				Expect(c.(*cli).cmd.Long).To(Equal(lo))
+				Expect(c.(*cli).cmd.Example).To(Equal(ex))
```
Adirio, line 295:
```suggestion
				Expect(c.(*cli).cmd.Short).To(Equal(defaultShortDescription))
				Expect(c.(*cli).cmd.Long).To(Equal(defaultLongDescription))
				Expect(c.(*cli).cmd.Example).To(Equal(defaultExample))
```

pkg/cli/cli.go
```diff
@@ -389,55 +397,82 @@ func (c cli) buildRootCmd() *cobra.Command {
...
+	sh, lo, ex := c.getRootCommandDefaults()
 
-Typical project lifecycle:
+	cmd.Short = c.cmdCfg.Short
+	if cmd.Short == "" {
+		cmd.Short = sh
+	}
 
-- initialize a project:
+	cmd.Long = c.cmdCfg.Long
```
Adirio, line 413:
```suggestion
	cmd.Long = fmt.Sprintf(c.cmdCfg.Long, cmd.Use)
```

pkg/cli/cli.go
```diff
@@ -389,55 +397,82 @@ func (c cli) buildRootCmd() *cobra.Command {
...
+	if cmd.Long == "" {
+		cmd.Long = lo
+	}
 
-  %s init --domain example.com --license apache2 --owner "The Kubernetes authors"
+	cmd.Example = c.cmdCfg.Example
```
Adirio, line 418:
```suggestion
	cmd.Example = fmt.Sprintf(c.cmdCfg.Example, cmd.Use)
```

Adirio (changes requested):
There are multiple places (like each command descriptions or contexts) where you are using `c.cmdCfg.CommandName` when it could be empty. The only place where you are considering that it may be empty is when assigning it to `c.cmd.Use`. By providing a default full `c.cmdCfg` and making the option only updated the non-empty ones you ensure that it is always non-empty. The following changes are all about this except for one of them which just removed the quotes around a variable.

anmol372:
Hi @Adirio,
Good Catch, I don't know how I missed that I'm setting `c.cmd` and using `c.cmdCfg` everywhere.
Thank You, I'm updating the PR as per suggestions.

camilamacedo86 (commented):
The commits require to be squashed for we are able o get this one merged here. Could you please do it?

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
```diff
@@ -170,8 +170,14 @@ Example Operator SDK main.go:
 
 ```go
 func main() {
+    rootCmdCfg := cli.RootCommandConfig{
+        CommandName: "operator-sdk",
+        //Short: "Short help text for operator-sdk",
```
camilamacedo86, line 175: I think we should not use SDK as a comment here. So, wy about update the comments with something such as `//Short: "Short help text for an external tool which is a consumer of kubebuilder",`?

docs/book/src/reference/cli-plugins.md
```diff
@@ -171,8 +171,14 @@ import (
 // Go plugin version unless '--plugins' or 'layout' specify
 // one of the other plugins.
 func main() {
+  rootCmdCfg := cli.RootCommandConfig{
+        CommandName: "controller-builder",
+        //Short: "Short help text for controller-builder",
+        //Long: "Longer help text for controller-builder",
+        //Example: "Example usage of controller-builder",
```
camilamacedo86, line 178:
```suggestion
```

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
```diff
@@ -170,8 +170,14 @@ Example Operator SDK main.go:
...
+        //Long: "Longer help text for operator-sdk",
+        //Example: "Example usage of operator-sdk",
```
camilamacedo86, line 177: Why we need to keep these commennts?

anmol372: I kept them to show an example usage of the new WithRootCommandConfig option.

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
```diff
@@ -170,8 +170,14 @@ Example Operator SDK main.go:
 
 ```go
 func main() {
+    rootCmdCfg := cli.RootCommandConfig{
```
estroz, line 173: I’d rather the original doc not be modified. PR description and godocs should be sufficient documentation.

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
Adirio: I think the example is ok. It is in the `designs` folder, not in the `pkg` one.

designs/extensible-cli-and-scaffolding-plugins-phase-1.md
Adirio: Makes sense to me

Adirio:
@anmol372 request me to re-review this once you solve the issue.

pkg/cli/cli.go
```diff
@@ -40,6 +40,7 @@ project.
 	projectVersionFlag = "project-version"
 	helpFlag           = "help"
 	pluginsFlag        = "plugins"
+	defaultCommandName = "kubebuilder"
```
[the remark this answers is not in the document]
Adirio: You can mark as resolved.

pkg/cli/cli.go
```diff
@@ -113,10 +121,10 @@ func (c cli) Run() error {
 	return c.cmd.Execute()
 }
 
-// WithCommandName is an Option that sets the cli's root command name.
-func WithCommandName(name string) Option {
+// WithRootCommandConfig configures a CLI's root command.
+func WithRootCommandConfig(cfg RootCommandConfig) Option {
 	return func(c *cli) error {
-		c.commandName = name
+		c.cmdCfg = cfg
```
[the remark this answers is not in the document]
Adirio: You didn't apply this. Providing a partial struct will overwrite with empty strings other fields.

pkg/cli/cli.go
```diff
@@ -91,7 +100,6 @@ type cli struct {
 // New creates a new cli instance.
 func New(opts ...Option) (CLI, error) {
 	c := &cli{
```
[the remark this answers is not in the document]
Adirio: You didn't apply this to set the default values at the start.

Adirio:
@anmol372 you didnt apply any of the requested changes.

anmol372:
@Adirio I have reverted the doc changes from the proposal doc: `designs/extensible-cli-and-scaffolding-plugins-phase-1.md`, and the explanatory comments in `docs/book/src/reference/cli-plugins.md`. 
But I didn't completely revert the second file to its original doc, since it's not a proposal doc and it should reference the current protocol to generate a new cmd.

pkg/cli/cli.go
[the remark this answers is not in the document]
anmol372: Yes, that's intentional, as I am setting the default for the other fields in `defaultCommand()` if they are empty after calling the `WithRootCommandConfig(cfg)` option.

I think it will do the same work as intended by this suggestion but also not set these descriptor fields early, so that we don't lose track of them when using them at the end., as per an earlier suggestion of yours. 😅

Adirio:
I mean the changes on the `WithRootCommandConfig` that only set non-zero fields, the changes that set the default values from the start, ...

pkg/cli/cli.go
anmol372: Setting the fields early, may not give the intended results with this. If the user-supplied string doesn't have any format specifiers specified with `%[1]s`, we will get a syntax error of the kind `(MISSING)(EXTRA string=cmd.Use)`. 
Then to resolve this case we might need to include a function to parse user strings first and then pass to variadic `Sprintf` based on results of parsing.

What do you think?

anmol372:
Hi @Adirio,
I have intentionally, left the setting of descriptor fields to `defaultCommand()` **with the following logic**
1. If they are empty after calling the `WithRootCommandConfig(cfg)` Option they are set to default values, with provided command-name.
2. If the Option is not called, they are set to default, with the default command name.

**For the following reasons:** 

1. I think this will achieve the same results as intended by your suggestion, but also not set these descriptor fields early, so that we don't lose track of them when using them at the end., as per an earlier suggestion of yours.

2. Further:

> `cmd.Long = fmt.Sprintf(c.cmdCfg.Long, cmd.Use)`

Setting the field with this suggestion will restrict projects to use descriptors with at least 1 format specifier specified with `%[1]s`. If the supplied string doesn't have the format specifier this will get a syntax error of the kind` (MISSING)(EXTRA string=cmd.Use).` Then to resolve this case we might need to include a function to parse user strings first and then pass to variadic` Sprintf ` based on the results of parsing.

What do you think?

Adirio:
First one thing I forgot to tell you: removing `WithCommandName` is a breaking change. You probably don't want this to be a breaking change. Offering a way of setting the 4 in a single `Option` seems reasonable but I expect to have ways to set each of them by their own too.


> I have intentionally, left the setting of descriptor fields to `defaultCommand()` **with the following logic**
> 
> 1. If they are empty after calling the `WithRootCommandConfig(cfg)` Option they are set to default values, with the provided command-name.
> 2. If the Option is not called, they are set to default, with the default command name.

Imagine the following case:
```
c := cli.New(
	WithCommandName("other-command-name"),
	WithRootCommandConfig(RootCommandConfig{
		Short: "Descritpion for other-command-name",
		Examples: `Examples for other-command-name`,
	}),
)
```
Now your command name is blank and will default to `kubebuilder` but the descriptions and examples are set for the other command. I think it is preferably to set the defaults, then parse the options to set the final values, and then template whatever is needed.

> **For the following reasons:**
> 
> 1. I think this will achieve the same results as intended by your suggestion, but also not set these descriptor fields early, so that we don't lose track of them when using them at the end., as per an earlier suggestion of yours.

I did say that, but I changed my mind, sorry. I didn't know which implications it had by then as the PR has evolved quite a bit since then. Also I had to dive deep into the `cli` code recently, so that may have had something to do with the change. Again sorry for my previous suggestion.

> 2. Further:
> 
> > `cmd.Long = fmt.Sprintf(c.cmdCfg.Long, cmd.Use)`
> 
> Setting the field with this suggestion will restrict projects to use descriptors with at least 1 format specifier specified with `%[1]s` in their descriptor fields. If the supplied string doesn't have the format specifier this will get a syntax error of the kind` (MISSING)(EXTRA string=cmd.Use).` Then to resolve this case we might need to include a function to parse user strings first and then pass to variadic`Sprintf` based on the results of parsing.

Yes, you are right, but proper documentation should be enough. It won't be a error, more of a warning, and it would be in the side of the command developer, not the final user. The command developer is expected to know the API of the `cli` library and if it requires a template with 1 param he should provide it.

So summing up, I think that the advantage of being able to provide each of them by themselves and not overriding previously set values is a stronger point that a warning in a description in case a command developer doesn't provide a `%[1]s`.

anmol372:
> First one thing I forgot to tell you: removing `WithCommandName` is a breaking change. You probably don't want this to be a breaking change. Offering a way of setting the 4 in a single `Option` seems reasonable but I expect to have ways to set each of them by their own too.

I realize that removing `WithCommandName` would be a breaking change, but I was under the impression that the cli package hadn't been released yet.

> Imagine the following case:
> 
> ```
> c := cli.New(
> 	WithCommandName("other-command-name"),
> 	WithRootCommandConfig(RootCommandConfig{
> 		Short: "Descritpion for other-command-name",
> 		Examples: `Examples for other-command-name`,
> 	}),
> )
> ```
> Now your command name is blank and will default to `kubebuilder` but the descriptions and examples are set for the other command. I think it is preferably to set the defaults, then parse the options to set the final values, and then template whatever is needed.

This scenario is the reason I choose to remove `WithCommandName` after the PR switched to using the `RootCommandConfig` struct. Because providing multiple ways to set the same value, did not make sense to me.

> I did say that, but I changed my mind, sorry. I didn't know which implications it had by then as the PR has evolved quite a bit since then. Also I had to dive deep into the `cli` code recently, so that may have had something to do with the change. Again sorry for my previous suggestion.

That's reasonable, opinions will change over the course of a simple PR that is taking longer than expected, I guess. 😄 

> Yes, you are right, but proper documentation should be enough. It won't be a error, more of a warning, and it would be in the side of the command developer, not the final user. The command developer is expected to know the API of the `cli` library and if it requires a template with 1 param he should provide it.

While the developer is expected to know the API, it seems a tad bit unnecessary to allow them to set their own descriptions in their projects with the stipulation of using a single param.😅 

Tailoring the Options to provide each of the descriptors separately may also lead to the expectation of not setting the descriptors that are not provided by the command developer, right?

Adirio:
> I realize that removing `WithCommandName` would be a breaking change, but I was under the impression that the cli package hadn't been released yet.

While there is no `kubebuilder` release, there are already projects like `operator-SDK` which are already using the library.

> This scenario is the reason I choose to remove `WithCommandName` after the PR switched to using the `RootCommandConfig` struct. Because providing multiple ways to set the same value, did not make sense to me.

If you set the defaults at the start and make the `WithRootCommandConfig` `Option` to only set non-blank descriptors, that should be solved.

> While the developer is expected to know the API, it seems a tad bit unnecessary to allow them to set their own descriptions in their projects with the stipulation of using a single param.😅

Maybe we can switch to a `Replacer` approach similar to how paths are templated. This will allow developers to provide `%[command]s` which will be replaced by the command name but it doesn't fail if you just provide a normal string.

```
replacer := strings.NewReplacer("%[command]s", c.cmdCfg.CommandName)
short := replacer.Replace(c.cmdCfg.Short)
long := replacer.Replace(c.cmdCfg.Long)
examples := replacer.Replace(c.cmdCfg.Examples)
```

> Tailoring the Options to provide each of the descriptors separately may also lead to the expectation of not setting the descriptors that are not provided by the command developer?

Well if you want to solve this you would have to make the fields be pointers to strings, so that a `nil` pointer can be distinguised from the zero string (`""`) and only setting them when they are not `nil`.

estroz:
>While there is no kubebuilder release, there are already projects like operator-SDK which are already using the library.

Treating the main branch as the dev branch means breaking changes can happen between releases, which the operator-sdk team is fine with.

>Maybe we can switch to a Replacer approach similar to how paths are templated

I like this to a degree, but wouldn't it be easier to require a command name be set?

In general, the CLI shouldn't make any assumptions about what is or isn't set: fields in `RootCommandConfig` should be used to directly set `cobra.Command` fields if not empty, otherwise use the default values (for which we know what needs to be substituted and where).

pkg/cli/cli.go
estroz: Users are setting this value, so `c.cmdCfg.Long` should be used without modification.

pkg/cli/cli.go
estroz: Users are setting this value, so `c.cmdCfg.Example` should be used without modification.

pkg/cli/cli.go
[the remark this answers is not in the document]
estroz: This will all be configured in `cli.initialize()`.

Adirio:
> Treating the main branch as the dev branch means breaking changes can happen between releases, which the operator-sdk team is fine with.

Right, but if a breaking change can be avoided, I would go for that. Also, I prefer to offer the possibility to set just one of those fields. Giving options to the user to a certain degree is nice IMO.
 
> > Maybe we can switch to a Replacer approach similar to how paths are templated
> 
> I like this to a degree, but wouldn't it be easier to require a command name be set?
> 
> In general, the CLI shouldn't make any assumptions about what is or isn't set: fields in `RootCommandConfig` should be used to directly set `cobra.Command` fields if not empty, otherwise use the default values (for which we know what needs to be substituted and where).

The issue is not the command name, @estroz, the issue is how to set it inside the description and examples. If we go for the  `fmt.Sprintf` approach, when the user provides a decription without a single `%s` or any `%[1]s`, `fmt.Sprintf` will add a ugly warning to the string. On the other hard, using a `strings.Replacer` doesn't print any warning in any case but still gives the user the ability to set the command name, or us to set it in the default description and examples.

estroz:
>If we go for the fmt.Sprintf approach, when the user provides a decription without a single %s or any %[1]s, fmt.Sprintf will add a ugly warning to the string

If a description is provided, the description should be used verbatim, i.e. don't modify user-specified help text. The only case in which the CLI should format a string with command name is if a default string is used, because the CLI knows that formatting directives exist.

Adirio:
> > If we go for the fmt.Sprintf approach, when the user provides a decription without a single %s or any %[1]s, fmt.Sprintf will add a ugly warning to the string
> 
> If a description is provided, the description should be used verbatim, i.e. don't modify user-specified help text. The only case in which the CLI should format a string with command name is if a default string is used, because the CLI knows that formatting directives exist.

Why special-case the default description to be fmted after calling the options (we have to wait until then to know the command name) and not provide this (already developed) capability to the user provided description? I agree that most of the time the user who provided the description will know the command name, and therefore, can provide the final description. And that would also allow to provide some alternative descriptors in the future. For example, a colored one.

pkg/cli/cli.go
```diff
@@ -113,10 +168,28 @@ func (c cli) Run() error {
 	return c.cmd.Execute()
 }
 
-// WithCommandName is an Option that sets the cli's root command name.
-func WithCommandName(name string) Option {
+// WithRootCommandConfig configures a CLI's root command.
```
Adirio, line 171: `CLI` is the interface, `Option`s are defined on `cli`, not on `CLI`.
```suggestion
// WithRootCommandConfig configures a cli's root command.
```

pkg/cli/cli.go
```diff
@@ -113,10 +168,28 @@ func (c cli) Run() error {
...
+		if cfg.Long != "" {
+			c.cmdCfg.Long = cfg.Long
+		} else {
+			c.cmdCfg.Long = replacer.Replace(defaultLongDescription)
+		}
+		if cfg.Example != "" {
+			c.cmdCfg.Example = cfg.Example
+		} else {
+			c.cmdCfg.Example = replacer.Replace(defaultExample)
+		}
```
Adirio, line 192: So basically only default descriptors get replaced, user-provided descriptors are expected to know the command name. I'm OK with this, it simplifies a bit the API for cli developers as we don't have to tell them that they have the `%[command]s` placeholder but it still uses the correct command name in the default descriptors in case a different command name is provided.

NIT: we may want to stick to something different that `kubebuilder` as the placeholder (even if the commnad name is `kubebuilder`, the default descriptors will replace it by itself) as that would allow us to use the actual string "kubebuilder" in comments, e.g. to provide urls. Tests would also need to be updated.

pkg/cli/cli.go
```diff
@@ -390,50 +463,10 @@ func (c cli) buildRootCmd() *cobra.Command {
 // defaultCommand returns the root command without its subcommands.
 func (c cli) defaultCommand() *cobra.Command {
```
Adirio, line 464: Being now a direct `return`, only a couple lines long, and just being called from one place, I wonder if it makes sense to have it as a separate function or we could just create it in place. Also, `defaultCommand` does not make sense any more as the function name as previously it was providing default descriptors but now we can customize them and `c.cdm` already has the "final" descriptors.

Adirio (commented):
A couple NITs

estroz:
>Why special-case the default description to be fmted after calling the options (we have to wait until then to know the command name) and not provide this (already developed) capability to the user provided description

I'm fine not doing this. I'm really just trying to expose the simplest API possible: pass the CLI a struct containing help strings that will be used verbatim, with no changes. Increasing complexity with replacement semantics means more confusion for users of an already-complex plugin system.

estroz (approved):
/approve

Adirio:
Root command recently received a rework:
- `Short` doesn't make sense for a root command as the short description is only showed when listing subcommands, and the root command is not a subcommand by definition.
- `Examples` is being built dinamically based on the available plugins and supported project versions. This is important in order to provide useful error messages when unable to resolve plugins.

This means that, from the 3 configurable fields that this PR is adding, only `Long` makes sense to be added. And, being `Long` a single field, the multi-field in one config struct approach doesn't fit any more.

@anmol372 Do you want to update this PR just to include the `Long` description or do you want it to be done in a separate PR? In that case, do you want to take the task or should we raise an issue and marked it as `help-wanted` & `good-first-issue` so that other people develops it?

anmol372:
Hi @Adirio ,
I'll take the task and do it in the same PR.

Adirio (changes requested):
Request a review from (at least) me once you get this done.

camilamacedo86:
Hi @anmol372, 

Could you please rebase it with the master for we are able to move forward here? Also, can you close all comments that are addressed?

camilamacedo86:
Hi @anmol372, 

It is open for a while without be rebased with the master? Could you please rebase this one and let us know if you could address all suggestion and if it is done for another review round? OR is no longer it required?

PR #1831, opened 2020-11-17
✨ Add `kubebuilder alpha config-gen` command to explore generating configuration with kustomize functions

pwittrock:
:sparkles:

See the `config-gen/README.md` for instructions


config-gen/apis/v1alpha1/testdata/componentconfig/expected.yaml
```diff
@@ -0,0 +1,323 @@
+apiVersion: v1
```
camilamacedo86, line 1: could not the testdata be generated in the testdata/ dir and checked in the CI.. WDYT?

config-gen/README.md
```diff
@@ -0,0 +1,81 @@
+# Config-gen
```
camilamacedo86, line 1: Should not this be in the pkg? WDYT?

config-gen/templates/patches/cert-manager/annotation.template.yaml
```diff
@@ -0,0 +1,5 @@
+{{ if .Spec.Enabled "cert-manager" }}
```
camilamacedo86, line 1: How we will ensure that KB boilerplates and the templates here are doing the same? 
Could be possible we centralize the implementation?

config-gen/README.md
camilamacedo86, line 1: I understand that the idea here is to add an alpha feature that will allow KB gen the project based on yaml file configuration instead of using the commands. 

Then, I think that needs to be a plugin. How we will know that the project was generated using this feature? WDYT about it be a plugin as well and respect the config(Project) file too? 

Could the config/gen use the https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/golang/v3/scaffolds/internal/templates instead of its own templates? In this way, we can ensure that it is centralized and when user gen the project with this feature or by using the commands the same result would be applied.

camilamacedo86: /hold

I am holding this one for it does not get merged before we stress what would be the best approach to add this feature. 

And then, I also think that we need to change the testdata to be by plugin:

```
testdata:
     golang/
           v2/
           v3/
     config-gen/
           v1alpha1/ 
````

File: config-gen/templates/patches/cert-manager/annotation.template.yaml (hunk 3 above)
Comment (pwittrock) [reply config-gen/templates/patches/cert-manager/annotation.template.yaml:1]: I think in many cases these are similar but not the same as the KB boilerplate templates.  The templates generate Kustomize files, including kustomization.yamls and patch files.  It might be possible to refactor the boilerplate templates to be reusable from both contexts.

File: config-gen/README.md (hunk 2 above)
Comment (pwittrock) [reply config-gen/README.md:1]: I'm not sure that we gain much by making this a plugin vs just a separate command.  It doesn't depend on any special project setup, nor does it emit any scaffolding files (it writes to stdout).

Since:

1. Existing commands write kustomize files, and this command emits config to stdout so it behaves differently
2. It doesn't change the project structure in anyway -- users just manually create their own `kubebuilderproject.yaml` file to use it

I think it makes more sense as being run as a separate command rather than an extension of an existing command.  WDYT?

Paths forward:

- Make it a subcommand of kubebuilder (as shown)
  - Could make it hidden from help
- Make it a separate executable (main.go) in the kubebuilder repo and folks `go get` it manually
- Put it in a separate repo independent of the kubebuilder project (e.g. `github.com/pwittrock/kubebuilder-fn`) and folks can `go get` it from there if they want
- Do one of the above as a plugin

File: go.mod [hunk 4]
```diff
@@ -3,12 +3,22 @@ module sigs.k8s.io/kubebuilder/v3
...
+	github.com/spf13/cobra v1.1.1
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/tools v0.0.0-20200403190813-44a64ad78b9b
+	golang.org/x/tools v0.0.0-20200616195046-dc31b401abb5
+	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776 // indirect
+	k8s.io/api v0.20.2 // indirect
+	k8s.io/apiextensions-apiserver v0.20.1 // indirect
+	k8s.io/apimachinery v0.20.2
+	k8s.io/utils v0.0.0-20210111153108-fddb29f9d009 // indirect
+	sigs.k8s.io/controller-tools v0.4.1
```
camilamacedo86, line 21: I do not think that we should add as dep the `sigs.k8s.io/controller-tools v0.4.1`. Currently, it is customized per project via Makefile and go/v2 is not using the same version of go/v3. So, should not be better it be in the config-gen instead of the full tool?

go.mod
```diff
@@ -3,12 +3,22 @@ module sigs.k8s.io/kubebuilder/v3
...
+	github.com/markbates/pkger v0.17.1
 	github.com/onsi/ginkgo v1.12.0
 	github.com/onsi/gomega v1.9.0
 	github.com/spf13/afero v1.2.2
-	github.com/spf13/cobra v0.0.7
+	github.com/spf13/cobra v1.1.1
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/tools v0.0.0-20200403190813-44a64ad78b9b
+	golang.org/x/tools v0.0.0-20200616195046-dc31b401abb5
+	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776 // indirect
```
camilamacedo86, line 16: I think we need to clean up here and remove the indirect ones.

go.mod
```diff
@@ -3,12 +3,22 @@ module sigs.k8s.io/kubebuilder/v3
...
+	k8s.io/api v0.20.2 // indirect
+	k8s.io/apiextensions-apiserver v0.20.1 // indirect
+	k8s.io/apimachinery v0.20.2
+	k8s.io/utils v0.0.0-20210111153108-fddb29f9d009 // indirect
+	sigs.k8s.io/controller-tools v0.4.1
+	sigs.k8s.io/kustomize/kyaml v0.10.2
```
camilamacedo86, line 22: The go/v2 and go/v3 uses a version configured in the kustomize too. So, should not be better it be in the config-gen instead of the full tool?

config-gen/examples/advancedextension/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Setup `kubebuilder config-gen` as a plugin
+
+```
+export XDG_CONFIG_HOME=~/.config
+
+# setup the kubebuilder plugin
+export KUBEBUILDER_PLUGIN=$XDG_CONFIG_HOME/kustomize/plugin/kubebuilder.sigs.k8s.io/kubebuilderproject
+mkdir -p $KUBEBUILDER_PLUGIN
+cat > $KUBEBUILDER_PLUGIN/KubebuilderProject <<EOF
```
camilamacedo86, line 16: - It is not possible for the users cut and past the example
- It is missing ````sh` as well in the markdown
- It is not defining the pre-requirements such as have the kustomize installed locally and what version should be used.

config-gen/examples/advancedextension/README.md
```diff
@@ -0,0 +1,41 @@
...
+#!/bin/bash
```
camilamacedo86, line 17: Also, not easy here to provide the examples only doc the config that should be passed to the kubebuilder alpha config-gen?

config-gen/README.md
camilamacedo86: Really thank you. As we chat will

/hold cancel

I do not block this PR. And will try to check it further. However, would really nice get other thoughts as well.

config-gen/apis/v1alpha1/testdata/componentconfig/expected.yaml
camilamacedo86, line 1: WDYT about add in the kubebuiler/testdata a sample which would be the project-v3 but with config-gen instead? Then, would not make it easier we compare both scenarios/options to get its advantages and etc?

config-gen/examples/advancedextension/README.md
pwittrock: https://github.com/kubernetes-sigs/kubebuilder/pull/1831/files#diff-7664f14e421dc7d2b17ee6081de4b2eb91be74aa6e667a6de143830768bb676e

pwittrock:
@camilamacedo86 I added testdata for the projects

config-gen/apis/v1alpha1/testdata/componentconfig/expected.yaml
pwittrock: Done

config-gen/examples/advancedextension/README.md
pwittrock: I'll clean this doc up a bit.

config-gen/apis/v1alpha1/testdata/componentconfig/expected.yaml
pwittrock: (See second commit)

camilamacedo86: I am not sure if it was what was in my mind. I thought in: 
- New testatada project. e.g project-v3-configgen
- where  project-v3-configgen would be == project-v3 without the default config scaffold. I mean, this project would be an example of your end goal idea. Are you not looking for to replace the config/ dir with the config-gen?

testdata/project-v2/_output/config-gen.yaml
```diff
@@ -0,0 +1,392 @@
...
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: (devel)
```
camilamacedo86, line 116: go/v2 and go/v3 does not support the same controller-gen version. 
The versions are defined in the Makefile. 

In this way, I understand that we should need to be able to configure the version that should be used as well. 
OR 
Could not the config-gen only generate the manifests and let the makefile targets generate controller-gen manifests?

testdata/project-v2/_output/config-gen.yaml
```diff
@@ -0,0 +1,392 @@
...
+      subresources:
+        status: {}
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
+---
+apiVersion: apiextensions.k8s.io/v1
```
camilamacedo86: See that the apiVersion used in go/v2 is v1beta1. 
For go/v3 users can choose if they will generate it in the latest version or not.
It is tracked in the PROJECT file and the tool will do the scaffolds accordingly.

testdata/project-v2/_output/config-gen.yaml
```diff
@@ -0,0 +1,392 @@
...
+kind: CustomResourceDefinition
```
camilamacedo86, line 61: In the Makefile, the tool has targets to help the users to apply the CRD's, to apply the CRs and etc. 
Could config-gen generate the files in the same directory structure to allow it?

config-gen/apis/v1alpha1/testdata/componentconfig/expected.yaml
camilamacedo86: It is adding the same configuration for all samples. However, each sample has a diff config.
My initial idea was we have a testdata sample to check the config-gen which would be like: 
- a new sample such as testdata/project-v3-config-gen 
- this sample would have the same config of project-v3
- it would be an example of its end-goal idea where the config/ directory was replaced by config-gen

Then, we could compare project-v3 with the new one and ensure that it is covering the same features.
Also, we could use this sample to run the same integration tests and ensure that all would work well as expected. WDYT?

test/testdata/generate.sh
```diff
@@ -66,6 +66,19 @@ function scaffold_test_project {
...
+apiVersion: kubebuilder.sigs.k8s.io
+kind: KubebuilderProject
+metadata:
+  name: $project
+spec:
+  image: controller:latest
+EOF
+    mkdir _output/
+    $kb alpha config-gen kubebuilderproject.yaml > _output/config-gen.yaml
+
```
camilamacedo86, line 81: We are here adding the same config for all testdata/samples. However, each sample has a diff scenario. 
So, imo we would need to add a correspondent config for each case that shows high effort OR just a new testdata sample using config-gen which covers the same project-v3 config as suggested in https://github.com/kubernetes-sigs/kubebuilder/pull/1831#discussion_r582450045. wdyt?

testdata/project-v2/_output/config-gen.yaml
```diff
@@ -0,0 +1,392 @@
+apiVersion: v1
```
camilamacedo86, line 1: Also, to imo we need to ensure that the projects using config-gen will work well as expected on the cluster. Then, we need to run the same tests https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v3/plugin_cluster_test.go. 

IMO we need to copy and paste the go/v3 scenario and tests and generate it with config-gen. Then, we comment all tests over features/configs which is not covered yet and add todo(s) for that. In this way, we will know what still missing and we will able to ensure that it is working.

go.mod
```diff
@@ -3,12 +3,18 @@ module sigs.k8s.io/kubebuilder/v3
...
 	github.com/spf13/afero v1.2.2
-	github.com/spf13/cobra v0.0.7
+	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/tools v0.0.0-20200403190813-44a64ad78b9b
+	golang.org/x/tools v0.0.0-20190920225731-5eefd052ad72
+	k8s.io/api v0.20.2 // indirect
+	k8s.io/apimachinery v0.20.2
+	sigs.k8s.io/controller-tools v0.3.0
+	sigs.k8s.io/kustomize/kyaml v0.10.10
```
camilamacedo86, line 18: IMO: 
- config-gen does not need controller-tools. If we write the file on disk it ought to use the version defined in the Makefile which is not the same for all plugin/go versions. And then, if we want to use it without write in the disk then I think it needs to be configurable in its file instead.

go.mod
```diff
@@ -3,12 +3,18 @@ module sigs.k8s.io/kubebuilder/v3
...
 	github.com/onsi/gomega v1.9.0
 	github.com/spf13/afero v1.2.2
-	github.com/spf13/cobra v0.0.7
+	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/tools v0.0.0-20200403190813-44a64ad78b9b
+	golang.org/x/tools v0.0.0-20190920225731-5eefd052ad72
+	k8s.io/api v0.20.2 // indirect
+	k8s.io/apimachinery v0.20.2
+	sigs.k8s.io/controller-tools v0.3.0
```
camilamacedo86, line 17: IMO: 
- config-gen does not need controller-tools. If we write the file on disk it ought to use the version defined in the Makefile which is not the same for all plugin/go versions. And then, if we want to use it without write in the disk then I think it needs to be configurable in its file instead.

go.mod
```diff
@@ -3,12 +3,18 @@ module sigs.k8s.io/kubebuilder/v3
...
+	github.com/markbates/pkger v0.17.1
 	github.com/onsi/ginkgo v1.12.0
 	github.com/onsi/gomega v1.9.0
 	github.com/spf13/afero v1.2.2
-	github.com/spf13/cobra v0.0.7
+	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/tools v0.0.0-20200403190813-44a64ad78b9b
+	golang.org/x/tools v0.0.0-20190920225731-5eefd052ad72
+	k8s.io/api v0.20.2 // indirect
```
camilamacedo86, line 15:
```suggestion
```

camilamacedo86: It shows not used. Could you please remove and run go mod tidy after that to ensure?

config-gen/apis/v1alpha1/types.go
```diff
@@ -0,0 +1,192 @@
...
+import (
+	"io/ioutil"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/kustomize/kyaml/errors"
+)
+
+// KubebuilderProject implements the API for generating configuration
+type KubebuilderProject struct {
```
camilamacedo86, line 28: What about it be something more generic such as ConfigGenerateProject instead ?

config-gen/apis/v1alpha1/types.go
```diff
@@ -0,0 +1,192 @@
...
+	metav1.TypeMeta   `json:",inline" yaml:",omitempty"`
+	metav1.ObjectMeta `json:"metadata,omitempty" yaml:"metadata,omitempty"`
+
+	// Spec is the configuration spec defining what configuration should be produced.
+	Spec KubebuilderProjectSpec `json:"spec,omitempty" yaml:"spec,omitempty"`
+
+	// Status is the configuration status defined at runtime.
+	Status KubebuilderProjectStatus `json:"status,omitempty" yaml:"status,omitempty"`
```
camilamacedo86, line 36: What means the KubebuilderProjectStatus in the config-gen context? What status it will hold?

config-gen/apis/v1alpha1/types.go
```diff
@@ -0,0 +1,192 @@
...
+	// ComponentConfigFilepath is the relative path to a component config file used to configure the
+	// controller-manager
+	ComponentConfigFilepath string `json:"componentConfigFilepath,omitempty" yaml:"componentConfigFilepath,omitempty"`
+}
+
+// KubebuilderProjectStatus is runtime status for the api configuration
+type KubebuilderProjectStatus struct {
+	CertCA string
+
+	CertKey string
```
camilamacedo86, line 126: Why CertCa and CertKey are attrs for the status options and not spec?

config-gen/README.md
```diff
@@ -0,0 +1,81 @@
...
+  name: simple
+spec:
+  # container image containing the controller-manager
+  image: example/simple:latest
+
+  # kubebuilder project root directory relative path from kubebuilderproject.yaml dir
+  # (defaults to current directory)
+  projectDirectory: ../project/...
+
+  # components to enable / disable generating config for
```
irvifa, line 56: For the config that's actually commented should we still keep it as it is on the example?

config-gen/apis/v1alpha1/types.go
pwittrock: Because they shouldn't be specified by the user.

config-gen/README.md
pwittrock: I'm open to anything

config-gen/apis/v1alpha1/types.go
pwittrock: Things generated by the function at runtime that need to be passed into templates

config-gen/apis/v1alpha1/types.go
pwittrock: I don't have a strong opinion.  `KubebuilderConfigGen` ?

go.mod
pwittrock: I've had a lot of trouble with `go.mod` because the latest `ginkgo` / `gomega` break our tests, and `go mod` keeps trying to update them.  I'll try again.

config-gen/apis/v1alpha1/testdata/componentconfig/expected.yaml
pwittrock: I could add that test.  This PR doesn't propose replacing the `config` directory, nor does it have any mechanism for doing so.  Do we want to add that functionality at this time, or defer support for replacing the existing workflow until we are ready to do so?

There are config-gen specific tests in its own package, and I'd prefer we keep the config between project versions consistent to catch any differences between how config is generated differently for them.

config-gen/examples/kustomize/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+This directory contains an example of using `kubebuilder alpha config-gen` as a Kustomize transformer
+plugin.
+
+This enables using `config-gen` in traditional kustomize workflows (patch, bases, etc).
+
+## Summary
+
+`config-gen` may be used from `kustomize` as a transformer plugin.  This allows the output
+to be customized using `commonLabels`, `commonAnnotations`, `namespace`, etc.
```
estroz, line 11: I think this will be an incredibly common use case. I wonder if the setup can be streamlined (not necessary right now since config-gen is alpha).

config-gen/apis/v1alpha1/types.go
irvifa: +1 for ` KubebuilderConfigGen`

config-gen/README.md
```diff
@@ -0,0 +1,81 @@
+# Config-gen
+
+`kubebuilder alpha config-gen` is a subcommand that generates configuration for kubebuilder projects as a configuration function.
+
+Supports:
+
+- Generating CRDs and RBAC from code
+- Generating webhook certificates for development
+- Selectively enabling / disabling components such as prometheus and webhooks
+  - See [types.go](apis/v1alpha1/types.go) for a list of components
```
irvifa, line 10:
```suggestion
- See [types.go](apis/v1alpha1/types.go) for a list of components
```

config-gen/README.md
irvifa: On 2nd thought, since I believe this is auto generated, then let it be as it is then

config-gen/apis/v1alpha1/testdata/componentconfig/expected.yaml
irvifa: If I'm allowed to suggest then I think it's better to separate the PR until it's ready.

config-gen/apis/v1alpha1/types.go
camilamacedo86: I was looking for we remove Kubebuilder from the name since it can be used in any place. however, not strong opiniated on that as well. I am fine with.

config-gen/apis/v1alpha1/testdata/componentconfig/expected.yaml
camilamacedo86: Totally fine with us do that in a follow-up

config-gen/apis/v1alpha1/types.go
camilamacedo86: So, have we not any scenario where people would like to use their own CertCA and CertKey? 
And then, should it always be provided by the config gen?

config-gen/README.md
camilamacedo86: IMO, the problem is that has the potential to grow a lot.  
So, if we scaffold by default all options with the default values it might starts to not be great.
I think @pwittrock is trying to follow up the pattern where when the values are not informed it assumes the default behaviour.

go.mod
Adirio: #2039 should have fixed that

config-gen/README.md
```diff
@@ -0,0 +1,81 @@
...
+#  disableAuthProxy: true
+
+   # use component config
+#  componentConfigFilepath: ./relative/path/to/controller_manager_config.yaml
+
+  # generate certificates for webhooks
+#  developmentOptions:
+#    generateCert: true
+#    certDuration: '24h'
+```
```
camilamacedo86, line 79: @pwittrock could we add an option here to enable write on disk as an option before we merge it such as alpha?

go.mod
camilamacedo86, line 18: I am not sure if we should add that as dep of kb. Could we make it specific for config-gen instead? WDYT

config-gen/README.md
pwittrock: This we written by hand.  It is mostly to document the fields that exist.  Only name and image are required, the rest are defaulted.

config-gen/README.md
pwittrock: What is the use case where letting the shell handle this is insufficient?

go.mod
pwittrock: Are you suggesting making config-gen a separate binary?

config-gen/apis/v1alpha1/types.go
pwittrock: I think we should think carefully before exposing these as an API.  Exposing these directly seems like a poor use experience, rather instead I would imagine the user specify the path to a .pem file or something.  Or using kustomize to generate the secret from a file + annotating it, and pulling that in automatically.  There are a number of ways we may want to do this, but I don't think manually specifying these fields is one of them.

config-gen/apis/v1alpha1/types.go
pwittrock: I think having kubebuilder in the name will make it more clear what this is vs other files that may be on disk.  Since the command is part of kubebuilder (as written), I am not sure if there is much to be gained by loosely coupling the `kind` to kubebuilder.

go.mod
pwittrock: Thanks!

config-gen/examples/kustomize/README.md
pwittrock: Done

testdata/project-v2/_output/config-gen.yaml
pwittrock: I like the idea of having e2e tests.  We have unit tests right now.  Lets merge this and add the e2e's before moving it out of alpha.

test/testdata/generate.sh
pwittrock: I think that is covered in the config-gen unit tests.

config-gen/apis/v1alpha1/types.go
camilamacedo86: +1 for your thoughts. However, `KubebuilderConfigGenStatus` shows only be useful to Cert/Key features. So, I still not sure about its usage as a status. we can re-think that latter as well. Not a blocker for this PR anyway.

config-gen/README.md
camilamacedo86: - I am as an operator developer would like to have the config/ directory scaffolded by config-gen so that, I do not need to work with kustomize patches and I can change the manually. 
- I am as kb/sdk maintainer would like to replace the config/ directory by config-gen without losing the backwards compatibility. 
 
So, my idea was we have an option for that which by default would be false. However, we can do that in follow-ups as well.

go.mod
camilamacedo86: It might be a good idea if the goal is to have the config-gen such as controller-gen from the controller-tools. However, if you decede to do not move in this way. IMO we need to remove the indirect deps here and add a comment to describe what deps have been added only because of the config-gen.

test/testdata/generate.sh
camilamacedo86: My point is that I do not think that we should add this change here because:
- We do not use the config-gen in this samples
- Add the same config-gen setup for all samples does not bring value since it is not showing how would be the testdata/sample config-gen output for each specific scenario 

So, imo we need to revert it and we can improve/work on its tests in follow ups.

camilamacedo86 (changes requested):
Hi @pwittrock, 

IMO the minimal changes required here for we get this PR merged are:  
- [ ] - Add the appropriate mojo on the PR title. 
- [ ] - Cleanup the go.mod 
- [ ] - Revert the change made in the testdata_generated
- [ ] - Rebase with master and have this PR squashed.

.golangci.yml
```diff
@@ -24,7 +24,6 @@ linters:
     - ineffassign
     - interfacer
     - lll
-    - maligned
```
Adirio, line 27: Reason?

pkg/cli/cli.go
```diff
@@ -429,6 +430,13 @@ func (c *cli) addSubcommands() {
 	if c.version != "" {
 		c.cmd.AddCommand(c.newVersionCmd())
 	}
+
+	alpha := &cobra.Command{
+		Use:   "alpha",
+		Short: "Alpha kubebuilder subcommands",
+	}
+	c.cmd.AddCommand(alpha)
+	alpha.AddCommand(configgen.NewCommand())
```
Adirio, line 439: Several comments about this.
1. I would create the `alpha` subcommand in its own file similar to `create`:
```
func (cli) newAlphaCmd() *cobra.Command {
	return &cobra.Command{
		Use:        "alpha",
		SuggestFor: []string{"experimental"},
		Short:      "Alpha kubebuilder commands",
		Long:       `Alpha kubebuilder commands for unstable features.
Backwards compatibility guarantee of any kind is NOT provided.`,
	}
}
```
2. Commands here are sorted in alphabetical order.
3. I would also follow the only-add-if-subcommands approach used for create.
```
	// kubebuilder alpha
	alphaCmd := c.newAlphaCmd()
	alphaCmd.AddCommand(configgen.NewCommand())
	if alphaCmd.HasSubCommands() {
		c.cmd.AddCommand(alphaCmd)
	}
```
4. I think Cobra allows us to hide some commands from help and alpha would be a very good candidate.

test/testdata/generate.sh
Adirio: I think that while it is in alpha we should not mess with the root `testdata` folder. It already has its own `testdata` folder where we can test it.

Adirio (commented):
Some comments on how I think we should include this alpha.

Haven't reviewed the internals.

go.mod
pwittrock: What is the concern here?

I'd rather not make it a separate binary as it creates friction both for development and users consuming it.  I could comment those dep lines, but go has a tool which explains the reason for transitive deps and the comment could get stale.

doc.go
```diff
@@ -0,0 +1,21 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+//go:generate go run github.com/markbates/pkger/cmd/pkger
+
+// Package kubebuilder contains pkged files compiled into the
+// go binaries.
+package kubebuilder
```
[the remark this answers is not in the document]
pwittrock: To include the templates statically in the binary.

pkg/cli/cli.go
pwittrock: > I think Cobra allows us to hide some commands from help and alpha would be a very good candidate.

Why don't we want this to be discoverable?  Is alpha insufficient?

pwittrock: > I would create the alpha subcommand in its own file similar to create

Done

> Commands here are sorted in alphabetical order.

Done
> I would also follow the only-add-if-subcommands approach used for create.

Done.  I put this logic in the creation command itself so it is more encapsulated and reduces the complexity of the `cli.go` file.

pwittrock:
> I think that while it is in alpha we should not mess with the root testdata folder. It already has its own testdata folder where we can test it.

Done

config-gen/README.md
pwittrock: Ah that makes sense.  Lets get the command stable first, then figure out a migration path.

test/testdata/generate.sh
pwittrock: Removed testdata

config-gen/apis/v1alpha1/types.go
pwittrock: SG.  The status exists as a way of making data available to the go-templates at runtime.

.golangci.yml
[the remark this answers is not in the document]
pwittrock: Fixed.

pwittrock:
@Adirio 
> I think that while it is in alpha we should not mess with the root testdata folder. It already has its own testdata folder where we can test it.

SGTM.  Updating the root testdata folder was a request in an earlier review so I added it.  I reverted that change.

config-gen/examples/kustomize/README.md
pwittrock: Added a subcommand to install the plugin

go.mod
pwittrock: Done

pwittrock:
@camilamacedo86 

Completed the items from your checklist.  PTAL

go.mod
pwittrock: I've cleaned up the file a bit by manually deleting the indirects and the tidy seems to keep them.  How does it look?

doc.go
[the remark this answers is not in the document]
Adirio: Could this be achieved with `embed` in 1.16?
(I'm not suggesting to change it, we are not supporting 1.16 yet)

go.mod
```diff
@@ -3,12 +3,19 @@ module sigs.k8s.io/kubebuilder/v3
...
 
 require (
+	github.com/cloudflare/cfssl v1.5.0
 	github.com/gobuffalo/flect v0.2.2
+	github.com/markbates/pkger v0.17.1
 	github.com/onsi/ginkgo v1.15.0
 	github.com/onsi/gomega v1.10.5
 	github.com/spf13/afero v1.2.2
-	github.com/spf13/cobra v0.0.7
+	github.com/spf13/cobra v1.1.1
```
Adirio, line 12: Do you need to update cobra for this PR?

I definetely think that bumping cobra is somethign we need to do, but if it is not needed forn this PR we should probably do it in a separate PR.

pkg/cli/alpha.go
```diff
@@ -0,0 +1,54 @@
...
+	return alpha
+}
+
+func (c *cli) addAlphaCmd() *cobra.Command {
+	alpha := c.newAlphaCmd()
+	if len(alpha.Commands()) > 0 {
+		c.cmd.AddCommand(alpha)
+	}
+	return alpha
+}
```
Adirio, line 54: I like this approach (we could probably do the same for `create`), what about?
```suggestion
func (c *cli) addAlphaCmd() {
	alpha := c.newAlphaCmd()
	if len(alpha.Commands()) > 0 {
		c.cmd.AddCommand(alpha)
	}
}
```

go.mod
```diff
@@ -3,12 +3,19 @@ module sigs.k8s.io/kubebuilder/v3
...
 	github.com/spf13/pflag v1.0.5
 	golang.org/x/tools v0.0.0-20201224043029-2b0845dc783e
+	// pin this version of yaml.v3, changing it impacts the test output
+	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776 // indirect
```
Adirio, line 16: Which import requires this? Just curious if we could somehow prevent using two different YAML parsers as we are using `sigs.k8s.io/yaml v1.2.0` internally.

pkg/cli/cli.go
Adirio: > > I think Cobra allows us to hide some commands from help and alpha would be a very good candidate.
> 
> Why don't we want this to be discoverable? Is alpha insufficient?

No strong opinion here, they can stay as they are.

camilamacedo86:
Hi @pwittrock, 

My only concern here still:  Cleanup the go.mod (more info: https://github.com/kubernetes-sigs/kubebuilder/pull/1831#discussion_r579359175 and https://github.com/kubernetes-sigs/kubebuilder/pull/1831#discussion_r583753262 )

(IMO): I think that have a go.mod ( module ) for the config-gen would be the best approach here. I do not like too much the idea over we start to add all these dependencies to KB when they are required only for config-gen.  It can bring complexities for we keep the project maintained and supported. Also, would be better we can the config-gen isolated from the project since it could indeed be moved to controller-tools for example in the future. Also, I think that the idea over we are able to `go get config-gen` can bring advantages for we are able to use it with the language plugins too.

However, if you see that is not possible then, we need to be very careful with the deps that will be added to Kubebuilder only to support this alpha feature. 
- How as maintainers (we) know what deps are required for config-gen and for KB? We need at least split then and make comments which will clarify that. E.g `// The following deps were added and are useful only for the alpha feature config-gen`
- Then,  I do not think that we can have the scenario or we need at least avoid that https://github.com/kubernetes-sigs/kubebuilder/pull/1831#discussion_r589352101

Is it make sense?

pkg/cli/alpha/config-gen/types.go
```diff
@@ -0,0 +1,192 @@
...
+	// RBACComponent will create RBAC rules
+	RBACComponent = "rbac"
+
+	// CertManagerComponent will create the Issuer and Certificate resources
+	// for CertManager to inject the certificates
+	CertManagerComponent = "cert-manager"
+
+	// PrometheusComponent will create the ServiceMonitor resource
+	PrometheusComponent = "prometheus"
+)
```
Adirio, line 65: This seems like a good candidate for an enum.

pkg/cli/alpha/config-gen/types.go
```diff
@@ -0,0 +1,192 @@
...
+
+// Enabled returns true if the component is enabled
+func (a *KubebuilderConfigGenSpec) Enabled(component string) bool {
+	return a.Components[component]
+}
+
+// ComponentConfigEnabled returns true if component config is being used
+func (a *KubebuilderConfigGenSpec) ComponentConfigEnabled() bool {
+	return a.ComponentConfigFilepath != ""
+}
```
Adirio, line 85: Does it make sense conceptually to have `ComponentConfig` special-cased? Maybe we can have two maps (one for booleans and other for strings) to enable other future components that store a string and use a single method that checks for `true` in one or for `!= ""` in the other.

pkg/cli/alpha/config-gen/types.go
```diff
@@ -0,0 +1,192 @@
...
+
+// KubebuilderConfigGenSpec defines the desired configuration to be generated
+type KubebuilderConfigGenSpec struct {
+	// Directory is the kubebuilder directory containing the code.
+	// Defaults to the directoring containing the KubebuilderConfigGen configuration file.
```
Adirio, line 90: I'm not native but I don't think directoring is a word.

pkg/cli/alpha/config-gen/types.go
```diff
@@ -0,0 +1,192 @@
...
+	Directory string `json:"projectDirectory" yaml:"projectDirectory"`
+
+	// Name is the name of project and used to generate the component and role names
+	// Defaults to metadata.name
+	Name string `json:"projectName" yaml:"projectName"`
+
+	// Namespace is the namespace to run the project in.
+	// Defaults to spec.name-system
```
Adirio, line 98: I didn't understand `spec.name-system` at first read. Later on `Defaults` I understood that you meant `spec.name` suffixes with `"-system"`.

pkg/cli/alpha/config-gen/types.go
```diff
@@ -0,0 +1,192 @@
...
+
+	// Image is the container image to run in the controller-manager Deployment
+	Image string `json:"image" yaml:"image"`
+
+	// Components is a map of components to enable or disable.
+	// Defaults to enabling:  ["namespace", "controller-manager", "rbac", "crds"]
+	Components map[string]bool `json:"components" yaml:"components"`
+
+	// DisableAuthProxy if set to true will disable the auth proxy
+	DisableAuthProxy bool `json:"disableAuthProxy,omitempty" yaml:"disableAuthProxy,omitempty"`
```
Adirio, line 109: Again, does it make sense to have it special-cased? I understand that this is a component that requires explicit disbaling instead of enabling, but probably a map makes more sense too. Would have to be checked in the `Enabled` method.

pkg/cli/alpha/config-gen/types.go
```diff
@@ -0,0 +1,192 @@
...
+		if kp.Name == "" {
+			return errors.Errorf("must specify Kubebuilder projectName field")
+		}
+		kp.Spec.Name = kp.Name
+	}
+	if kp.Spec.Namespace == "" {
+		if kp.Namespace != "" {
+			kp.Spec.Namespace = kp.Namespace
+		}
+		kp.Spec.Namespace = kp.Spec.Name + "-system"
```
Adirio, line 160: Either you are missing an `else` clause for the second assignment or the first part is always being overriden.

pkg/cli/alpha/config-gen/types.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	if kp.Spec.Image == "" {
+		return errors.Errorf("must specify Kubebuilder image field")
+	}
+	if kp.Spec.Development.CertDuration == 0 {
+		d := time.Hour * 1
+		kp.Spec.Development.CertDuration = d
```
Adirio, line 167: Is there any reason why you are not doing this?
```suggestion
		kp.Spec.Development.CertDuration = time.Hour
```

pkg/cli/alpha/config-gen/types.go
Adirio: P.S.: Am I correct thinking that this is a map instead of an struct to allow programatic addition to fields here? Like for example extensions defining new components?

pkg/cli/alpha/config-gen/cmd.go
```diff
@@ -0,0 +1,278 @@
...
+		fmt.Fprintln(os.Stderr, err.Error())
+		return nil
+	}
+
+	// Eager check to make sure pkged templates are found.
+	err = pkger.Walk("/pkg/cli/alpha/config-gen/templates/resources", func(_ string, _ os.FileInfo, err error) error {
+		return err
+	})
+	if err != nil {
+		// this shouldn't fail it it was compiled correctly
```
Adirio, line 53:
```suggestion
		// this shouldn't fail if it was compiled correctly
```

pkg/cli/alpha/config-gen/cmd.go
```diff
@@ -0,0 +1,278 @@
...
+	if os.Getenv("KUSTOMIZE_FUNCTION") == "true" {
+		// run as part of kustomize -- read from stdin
+		c.Args = cobra.MinimumNArgs(0)
+	} else {
+		c.Args = cobra.MinimumNArgs(1)
+	}
+	c.RemoveCommand(c.Commands()...)
+	c.Use = "config-gen PROJECT_FILE [RESOURCE_PATCHES...]"
+	c.Version = `v1alpha1`
+	c.Short = `Generate configuration for a controller-runtime based projects`
```
Adirio, line 98:
```suggestion
	c.Short = `Generate configuration for controller-runtime based projects`
```

pkg/cli/alpha/config-gen/cmd.go
```diff
@@ -0,0 +1,278 @@
...
+	  "bars.example.my.domain": true
+
+	# if set, use component config for the project
+	# optional -- defaults to unset
+	componentConfigFilepath: ./path/to/componentconfig.yaml
+
+	# options for project development not intended for production
+	developmentOptions:
+	  # generate the webhook certificate and wire it as needed
+	  generateCert: true
```
Adirio, line 179: You've got mixed spaces and tabs in this example what makes it hard to read.

Adirio (changes requested):
I still have a bunch of files to review but there goes some comments. Most of them are either questions to try to understand it better or small changes.

go.mod
pwittrock: No, this is just what happened with I ran `go test ; go mod tidy`

go.mod
pwittrock: the kustomize fn framework uses this

go.mod
Adirio: Checking if it works with v0.0.7 and if `go mod tidy` doesn't bump it afterwards would confirm if this is needed or not.

go.mod
Adirio: Are the output changes considerable or breaking? Or can we just update the test and remove this indirect require?

doc.go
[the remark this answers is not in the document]
pwittrock: Yes, I believe so.  :). (Yay)

pkg/cli/alpha/config-gen/types.go
pwittrock: Do you mean with `iota` or something else?  These are values users will specify.

pkg/cli/alpha/config-gen/types.go
pwittrock: Is see this as different.  Enabling component config requires setting the file path rather than setting it simply as on.

The Components could be a struct, I'm not sure if it matters too much.  It is a collection of key-value pairs for flagging individual components as enabled or disabled.

pkg/cli/alpha/config-gen/types.go
pwittrock: This is a bit different from components, it's more about configuring components.  Either map or struct is fine.

pwittrock:
@camilamacedo86 @DirectXMan12 @Adirio 

Do we have consensus on whether or not we want this in the kb binary or if the go.mod concerns are blocking?  I've heard a number of different things in the calls and on the PR comments, and I'd like to make sure we are in agreement before moving forward.

I can move this into a separate module in a sub directory of this repo or another repo, I but want to minimize churn by figuring out what we want the end state to be before doing the work.

- Do we want this as its own go-module or as part of an existing one?
- Do we want this as its own binary or as part of an existing one?
- Do we want this as its own repo or part of an existing one?
- Should it be exposed from a kubebuilder command?  If so how?

Adirio:
NIT: the rebase required is probably changing `cli` to `CLI`

I think that the end goal should be making this a plugin on its own, but plugins can do almost anything so I have no strong opinion. I would say that an alpha command would be a nice approach. We will probably want to expose a `WithExtraAlphaCommands` soon enough too in order to allow third-party libraries to add subcommands of `alpha`.

pkg/cli/alpha/config-gen/types.go
Adirio: It can be a subtype of `string`, doesn't need to be `int` using `iota`. How does the user specify this? If it does through flags there is probably someway to provide a list of valid values to a flag.

pkg/cli/alpha/config-gen/types.go
Adirio: The point I wanted to raise was, may there be additional components that require a path instead of an "on" status? And, does it make sense to have something like the following snippet?
```go
// Enabled returns true if the component is enabled
func (a *KubebuilderConfigGenSpec) Enabled(component string) bool {
	if enabled, found := a.Components[component]; found {
		return enabled
	}
	if path, found := a.ComponentsWithPath[component]; found {
		return path != ""
	}
	return false
}
```

Adirio: It just feels a bit odd to have components enabled and disabled in different levels:
```yaml
components:
  namespace: true
  componentConfig: true
disableAuthProxy: true
componentConfigFilepath: "..."
```
and then checked with different functions.

Using a struct would make the `Enabled` function more verbose and would hard-code the available components but if those are non-issues it could be done something like:
```go
type Components struct {
	Namespace           bool   `json:"enableNamespace,omitempty"`         `yaml:"enableNamespace,omitempty"`
	ControllerManager   bool   `json:"enableControllerManager,omitempty"` `yaml:"enableControllerManager,omitempty"`
	NoAuthProxy         bool   `json:"disableAuthProxy,omitempty"`        `yaml:"disableAuthProxy,omitempty"`
	ComponentConfigPath string `json:"pathForComponentConfig,omitempty"`  `yaml:"pathForComponentConfig,omitempty"`
}

func (c Componentes) Enabled(component string) bool {
	switch component {
	case NamespaceComponent:
		return c.Namespace
	case ControllerManagerComponent:
		return c.ControllerManager
	case AuthProxyComponent:
		return !c.NoAuthProxy
	case ComponentConfigComponent:
		return c.ComponentConfigPath != ""
	default:
		return false
	}
}
```

and the resulting yaml
```yaml
components:
  enableNamespace: true
  enableComponentConfig: true
  disableAuthProxy: true
  pathForComponentConfig: "..."
```

P.S.: I'm using an slightly different nomenclature to explicitly add path and enabled to the fields but feel free to use whatever nomenclature you feel first better.

pkg/cli/alpha/config-gen/examples/kustomize/README.md
```diff
@@ -0,0 +1,39 @@
...
+```sh
+kustomize build --enable-alpha-plugins .
+```
+
+## Use `kustomize` to invoke the plugin
+
+Kustomize will invoke the `kubebuilder alpha config-gen` subcommand as a transformer plugin.
+
+```sh
+kustomize build --enable-alpha-plugins .
```
Adirio, line 36: Should these two commands be identical or copy-pasted and forgot to edit?

pkg/cli/alpha/config-gen/testdata/doc.go
```diff
@@ -0,0 +1,27 @@
...
+
+// Package testdata contains input and expected output for running config-gen.
+//
+// To add a new test create a new directory with the test name, a config.yaml with
+// the input, and an expected.yaml with the expected output.
+//
+// The project directory contains a sample project used as input.  New sample projects
+// may be added as new directories and referenced from the config.yaml.
+//
+// To update the testdata automatically modify v1alpha1_test.go by uncommenting
```
Adirio, line 25:
```suggestion
// To update the testdata automatically modify ../configgen_test.go by uncommenting
```

pkg/cli/alpha/config-gen/cert-manager-patches.go
```diff
@@ -0,0 +1,38 @@
...
+			return &framework.Selector{
+				Kinds: []string{
+					"CustomResourceDefinition",
+					"ValidatingWebhookConfiguration",
+					"MutatingWebhookConfiguration",
+				},
+			}
+		},
+	}
+}
```
Adirio, line 38:
```suggestion
import (
	"sigs.k8s.io/kustomize/kyaml/fn/framework"
)

// CertManagerPatchTemplate returns the PatchTemplate for cert-manager
func CertManagerPatchTemplate(_ *KubebuilderConfigGen) framework.PT {
	return framework.PT{
		Dir: "/pkg/cli/alpha/config-gen/templates/patches/cert-manager",
		Selector: func() *framework.Selector {
			return &framework.Selector{
				Kinds: []string{
					"CustomResourceDefinition",
					"ValidatingWebhookConfiguration",
					"MutatingWebhookConfiguration",
				},
			}
		},
	}
}
```

pkg/cli/alpha/config-gen/cert-manager-patches.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"github.com/markbates/pkger"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+)
+
+// CertManagerPatchTemplate returns the PatchTemplate for cert-manager
+func CertManagerPatchTemplate(kp *KubebuilderConfigGen) framework.PT {
+	return framework.PT{
+		Dir: pkger.Dir("/pkg/cli/alpha/config-gen/templates/patches/cert-manager"),
```
Adirio, line 27:
```suggestion
	"sigs.k8s.io/kustomize/kyaml/fn/framework"
)

// CertManagerPatchTemplate returns the PatchTemplate for cert-manager
func CertManagerPatchTemplate(_ *KubebuilderConfigGen) framework.PT {
	return framework.PT{
		Dir: "/pkg/cli/alpha/config-gen/templates/patches/cert-manager",
```

pkg/cli/alpha/config-gen/crd-patches.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"github.com/markbates/pkger"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// CRDPatchTemplate returns the PatchTemplate for crd
+func CRDPatchTemplate(kp *KubebuilderConfigGen) framework.PT {
+	return framework.PT{
+		Dir: pkger.Dir("/pkg/cli/alpha/config-gen/templates/patches/crd"),
```
Adirio, line 28: NIT: redundant type conversion:
```suggestion
	"sigs.k8s.io/kustomize/kyaml/fn/framework"
	"sigs.k8s.io/kustomize/kyaml/yaml"
)

// CRDPatchTemplate returns the PatchTemplate for crd
func CRDPatchTemplate(kp *KubebuilderConfigGen) framework.PT {
	return framework.PT{
		Dir: "/pkg/cli/alpha/config-gen/templates/patches/crd",
```

pkg/cli/alpha/config-gen/types.go
Adirio: Nevermind about the enum, `const` should be better than `var` though.

pkg/cli/alpha/config-gen/templates/patches/crd/conversion.template.yaml
```diff
@@ -0,0 +1,17 @@
+{{ if .Spec.Enabled "webhooks" }}
+spec:
+  conversion:
+    strategy: Webhook
+    webhookClientConfig:
+{{- if .Spec.Development.GenerateCert }}
+      caBundle: {{ .Status.CertCA }}
+{{- else }}
+      # this is "\n" used as a placeholder, otherwise it will be rejected by the apiserver for being blank,
+	    # but we're going to set it later using the cert-manager (or potentially a patch if not using cert-manager)
```
Adirio, line 10:
```suggestion
      # but we're going to set it later using the cert-manager (or potentially a patch if not using cert-manager)
```

pkg/cli/alpha/config-gen/templates/patches/controller-manager/99-component.template.yaml
```diff
@@ -0,0 +1,17 @@
...
+        - "--config=controller_manager_config.yaml"
+        volumeMounts:
+        - name: manager-config
+          mountPath: /controller_manager_config.yaml
+          subPath: controller_manager_config.yaml
+      volumes:
+      - name: manager-config
+        configMap:
+          name: manager-config
+{{ end }}
```
Adirio, line 17:
```suggestion
{{ end }}

```

pkg/cli/alpha/config-gen/types.go
```diff
@@ -0,0 +1,192 @@
...
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/kustomize/kyaml/errors"
+)
+
+// KubebuilderConfigGen implements the API for generating configuration
+type KubebuilderConfigGen struct {
+	metav1.TypeMeta   `json:",inline" yaml:",omitempty"`
+	metav1.ObjectMeta `json:"metadata,omitempty" yaml:"metadata,omitempty"`
```
Adirio, line 30: Is there any requirement for them to be a k8s object-like type or just a design decission?

pkg/cli/alpha/config-gen/cmd.go
```diff
@@ -0,0 +1,278 @@
...
+
+	if os.Getenv("KUSTOMIZE_FUNCTION") == "true" {
+		// run as part of kustomize -- read from stdin
+		c.Args = cobra.MinimumNArgs(0)
+	} else {
+		c.Args = cobra.MinimumNArgs(1)
+	}
+	c.RemoveCommand(c.Commands()...)
+	c.Use = "config-gen PROJECT_FILE [RESOURCE_PATCHES...]"
+	c.Version = `v1alpha1`
```
Adirio, line 97: Maybe semver?
```suggestion
	c.Version = `v0.1.0`
```

pkg/cli/alpha/config-gen/cmd.go
```diff
@@ -0,0 +1,278 @@
...
+		// these are run before the templates
+		PreProcessFilters: []kio.Filter{
+			// run controller-gen libraries to generate configuration from code
+			ControllerGenFilter{KubebuilderConfigGen: kp},
+			// inject generated certificates
+			CertFilter{KubebuilderConfigGen: kp},
+		},
+
+		// generate resources
+		TemplatesFn: framework.TemplatesFromDir(pkger.Dir("/pkg/cli/alpha/config-gen/templates/resources")),
```
Adirio, line 72: NIT: redundant type conversion:
```suggestion
		TemplatesFn: framework.TemplatesFromDir("/pkg/cli/alpha/config-gen/templates/resources"),
```

go.mod
pwittrock: They aren't breaking, but `go` doesn't seem to pick a consistent version unless this is specified and the flapping between versions creates a lot of confusion and churn.

pkg/cli/alpha/config-gen/types.go
pwittrock: I am happy to change the API however we want.

As it stands today, this API makes a distinction between a component (*what*) and feature (*how*).  Features exist to support components and don't make sense on their own.

It isn't a hard distinction -- would just generating the a namespace make sense?  IDK.  But the namespace more or less stands alone and is loosely coupled enough to the other components that I think it could be.  However just generating authproxy doesn't make sense to me -- authproxy is defined by rbac roles, services, and patches to other resources.  The same is true for componentConfig.

We could remove the concept of components altogether and just make things features.  The intent was to break the config into loosely coupled groups of resource that could be generated independently of one another, but that may not make sense.  Is there a use case for *not* creating the namespace or controller-manager.  Probably not worth implementing for alpha in any case.

Why don't I simply drop the components field and always generate all the resources?

pkg/cli/alpha/config-gen/types.go
pwittrock: A design decision, but consistent with other tools such as kustomize.

pkg/cli/alpha/config-gen/examples/kustomize/README.md
pwittrock: Good catch

pkg/cli/alpha/config-gen/cmd.go
pwittrock: Done.  Good catch

pkg/cli/alpha/config-gen/types.go
pwittrock: done

pkg/cli/alpha/config-gen/types.go
pwittrock: Updated comment

pkg/cli/alpha/config-gen/types.go
pwittrock: Done

pkg/cli/alpha/config-gen/types.go
pwittrock: We should drop these if we are going to make them fields

go.mod
pwittrock: Running either `go build ./pkg/...` or `go mod tidy` bumps this to v1.0.0 or higher.

pwittrock:
Push latest changes.

The `go.mod` file is a result of running the tests and tidy.  If I revert it back to the state before this PR, the test and tidy will put it back in this state.

pwittrock:
@Adirio @camilamacedo86 

I've worked on cleaning up the API.  The PR is broken as it stands, but I want to get the API stable before going back and re-implementing it.

- Get rid of components
- Break top level API into 3 pieces -- crds, controller-manager, webhooks
- Subdivide top components into major features
- Make everything optional and a pointer so KubebuilderConfigGen resources can be patched and merged in the future
- Make certificates a discriminated union with placeholders for certmanager and manual options

pkg/cli/alpha/config-gen/types.go
pwittrock: Kustomize plugins also require this format

go.mod
estroz: I think this is ok.

pkg/cli/alpha.go
```diff
@@ -0,0 +1,53 @@
...
+		alpha.AddCommand(alphaCommands[i])
+	}
+	return alpha
+}
+
+func (c *CLI) addAlphaCmd() {
+	alpha := c.newAlphaCmd()
+	if len(alpha.Commands()) > 0 {
+		c.cmd.AddCommand(alpha)
+	}
```
estroz, line 52: Nit: this could be slightly more efficient
```suggestion
	if len(alphaCommands) > 0 {
		c.cmd.AddCommand(c.newAlphaCmd())
	}
```

estroz (commented):
As this PR stands now, considering that it is an alpha command/implementation, LGTM. The license headers need updating to 2021 though.

pwittrock:
Sync'ed up offline with @camilamacedo86 .  Plan is to leave with the shared go.mod for now.  If this becomes a problem we can move it into a separate module, and lazily `go get` it at runtime when the command is invoked.

pkg/cli/alpha.go
pwittrock: Done

pwittrock:
PTAL

estroz (approved):
/lgtm

Adirio:
This PRs updates the copyright year of 3 files (inside `pkg/cli/internal/config`) that were not touched by it. This change is not a major issue as that package is being removed in #2083.

PR #1667, opened 2020-09-08
⚠️(v3+ only)  feat: update cert-manager crd to v1.0 in kubebuilder v3

MartinForReal:
Signed-off-by: MartinForReal <fanshangxiang@gmail.cn>
Fixed: #1666 

Migrate cert-manager related CR to v1

PR #1840, opened 2020-11-18
:warning: (go/v3-alpha) bump cert-manager CRs to v1

### LINKED ISSUE #1666: cert-manager related configuration should be migrated to cert-manager.io/v1
add support for cert-manager v1.0+ 

related: #1667 

/kind feature

estroz:
This PR bumps cert-manager CR apiVersion to v1 from v1beta2 for go/v3-alpha. TestContext has been updated to be cluster version-aware such that the correct cert-manager bundle URL is applied.

- pkg/plugin/v3/scaffolds/internal/templates/config: cert-manager CRs have apiVersion v1
- test/e2e/utils: KubectlVersion -> KubernetesVersion, which now gets created in NewTestContext. cert-manager install funcs use this version to get the correct bundle for a particular cluster version
- test/e2e/{v2,v3}: update {Install,Uninstall}CertManager() to include new 'hasv1beta2CRs' boolean arg, which depends on plugin being tested

Closes #1666 
Closes #1667

### DIFF
--- a/test/e2e/utils/test_context.go
+++ b/test/e2e/utils/test_context.go
@@ -27,11 +27,6 @@ import (
 	. "github.com/onsi/ginkgo" //nolint:golint
 )
 
-const certmanagerVersion = "v0.11.0"
-const prometheusOperatorVersion = "0.33"
-const certmanagerURL = "https://github.com/jetstack/cert-manager/releases/download/%s/cert-manager.yaml"
-const prometheusOperatorURL = "https://raw.githubusercontent.com/coreos/prometheus-operator/release-%s/bundle.yaml"
-
 // TestContext specified to run e2e tests
 type TestContext struct {
 	*CmdContext
@@ -42,8 +37,9 @@ type TestContext struct {
 	Kind       string
 	Resources  string
 	ImageName  string
-	Kubectl    *Kubectl
 	BinaryName string
+	Kubectl    *Kubectl
+	K8sVersion *KubernetesVersion
 }
 
 // NewTestContext init with a random suffix for test TestContext stuff,
@@ -54,30 +50,36 @@ func NewTestContext(binaryName string, env ...string) (*TestContext, error) {
 		return nil, err
 	}
 
-	testGroup := "bar" + testSuffix
-	path, err := filepath.Abs("e2e-" + testSuffix)
+	cc := &CmdContext{
+		Env: env,
+	}
+
+	// Use kubectl to get Kubernetes client and cluster version.
+	kubectl := &Kubectl{
+		Namespace:  fmt.Sprintf("e2e-%s-system", testSuffix),
+		CmdContext: cc,
+	}
+	k8sVersion, err := kubectl.Version()
 	if err != nil {
 		return nil, err
 	}
 
-	cc := &CmdContext{
-		Env: env,
-		Dir: path,
+	// Set CmdContext.Dir after running Kubectl.Version() because dir does not exist yet.
+	if cc.Dir, err = filepath.Abs("e2e-" + testSuffix); err != nil {
+		return nil, err
 	}
 
 	return &TestContext{
 		TestSuffix: testSuffix,
 		Domain:     "example.com" + testSuffix,
-		Group:      testGroup,
+		Group:      "bar" + testSuffix,
 		Version:    "v1alpha1",
 		Kind:       "Foo" + testSuffix,
 		Resources:  "foo" + testSuffix + "s",
 		ImageName:  "e2e-test/controller-manager:" + testSuffix,
 		CmdContext: cc,
-		Kubectl: &Kubectl{
-			Namespace:  fmt.Sprintf("e2e-%s-system", testSuffix),
-			CmdContext: cc,
-		},
+		Kubectl:    kubectl,
+		K8sVersion: &k8sVersion,
 		BinaryName: binaryName,
 	}, nil
 }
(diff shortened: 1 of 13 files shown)


estroz:
Didn't see any recent activity on #1667 so I rebased and updated that code.

/ping @MartinForReal

test/e2e/utils/test_context.go
```diff
@@ -54,30 +50,36 @@ func NewTestContext(binaryName string, env ...string) (*TestContext, error) {
...
+	// Set CmdContext.Dir after running Kubectl.Version() because dir does not exist yet.
+	if cc.Dir, err = filepath.Abs("e2e-" + testSuffix); err != nil {
+		return nil, err
 	}
 
 	return &TestContext{
 		TestSuffix: testSuffix,
 		Domain:     "example.com" + testSuffix,
-		Group:      testGroup,
+		Group:      "bar" + testSuffix,
```
camilamacedo86, line 75:
```suggestion
		Group:      testSuffix,
```
why we would add "bar"?

test/e2e/utils/test_context.go
```diff
@@ -54,30 +50,36 @@ func NewTestContext(binaryName string, env ...string) (*TestContext, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
-	cc := &CmdContext{
-		Env: env,
-		Dir: path,
+	// Set CmdContext.Dir after running Kubectl.Version() because dir does not exist yet.
+	if cc.Dir, err = filepath.Abs("e2e-" + testSuffix); err != nil {
+		return nil, err
```
camilamacedo86, line 69: why it is required as well?

test/e2e/utils/test_context.go
```diff
@@ -117,6 +140,24 @@ func (t *TestContext) InstallCertManager() error {
...
+			"warning: error when running kubectl delete during cleaning up cert manager: %v\n", err)
+	}
+	if _, err := t.Kubectl.Delete(false, "namespace", "cert-manager"); err != nil {
+		fmt.Fprintf(GinkgoWriter, "warning: error when cleaning up the cert manager namespace: %v\n", err)
+	}
+}
+
+const (
+	prometheusOperatorVersion = "0.33"
+	prometheusOperatorURL     = "https://raw.githubusercontent.com/coreos/prometheus-operator/release-%s/bundle.yaml"
```
camilamacedo86, line 158: wdyt about we keep all const on the top of this file?

test/e2e/v2/plugin_cluster_test.go
```diff
@@ -39,8 +39,9 @@ var _ = Describe("kubebuilder", func() {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(kbc.Prepare()).To(Succeed())
 
+			// Install cert-manager with v1beta2 CRs.
 			By("installing cert manager bundle")
-			Expect(kbc.InstallCertManager()).To(Succeed())
+			Expect(kbc.InstallCertManager(true)).To(Succeed())
```
camilamacedo86, line 44: It shows a little strange we have a func `InstallCertManager` which receives a boolean. WDYT about `InstallCertManager(version string)` or `InstallCertManagerV2` and `InstallCertManagerV3`

test/e2e/utils/test_context.go
```diff
@@ -54,30 +50,36 @@ func NewTestContext(binaryName string, env ...string) (*TestContext, error) {
...
+		Group:      "bar" + testSuffix,
 		Version:    "v1alpha1",
 		Kind:       "Foo" + testSuffix,
 		Resources:  "foo" + testSuffix + "s",
 		ImageName:  "e2e-test/controller-manager:" + testSuffix,
 		CmdContext: cc,
-		Kubectl: &Kubectl{
-			Namespace:  fmt.Sprintf("e2e-%s-system", testSuffix),
-			CmdContext: cc,
-		},
```
camilamacedo86, line 80: Why would we change the default ns? It will probably affect the SDK tests as well.

estroz: The namespace hasn't been changed, look at the old namespace setup.

test/e2e/utils/test_context.go
estroz: The group setup hasn't been changed, look at the old group setup.

test/e2e/utils/test_context.go
estroz: From the code:
>Set CmdContext.Dir after running Kubectl.Version() because dir does not exist yet.

test/e2e/utils/test_context.go
estroz: Usually I would do that, but these variables are only relevant to two methods in the entire file. Ditto for cert-manager variables.

test/e2e/v2/plugin_cluster_test.go
estroz: `InstallCertManager(hasv1beta2CRs bool)` is plugin-agnostic, so when a go/v4 plugin is added, the e2e test can use `InstallCertManager(false)` like go/v3 does instead of adding a `InstallCertManagerV4()` method.

test/e2e/utils/test_context.go
camilamacedo86: Catcher line 57 `testGroup := "bar" + testSuffix`

test/e2e/utils/test_context.go
camilamacedo86: IMO the const centralized on top is a good practice that should be always followed. 
But lets no block it because of this nit.

test/e2e/v2/plugin_cluster_test.go
camilamacedo86: WDYT about `InstallCertManager(version string)`? the `InstallCertManager(hasv1beta2CRs bool)` approach shows harder to get how it works.

camilamacedo86 (approved):
IMO we just need to address a nit: https://github.com/kubernetes-sigs/kubebuilder/pull/1840/files#r525623927

Otherwise, it shows great 
/approve

test/e2e/v2/plugin_cluster_test.go
estroz: What version does `version` correspond to? cert-manager? Exporting the cert-manager version is probably not great since the URL chosen depends on which version you pass. Realistically there are only two options: use an "old" bundle URL for an "old" version, or use the "new" bundle URL (depending on cluster type) for a "new" version. This is why I chose a boolean, in addition to not wanting to expose `InstallCertManager()` internals. If `version` is decided to be the best parameter, I suggest requiring a full url, like `InstallCertManager(url string)`.

camilamacedo86: It is a little confusing. Maybe `InstallCertManagerWithV1beta1`. However, I a strangle to find a way to make it more clear as well by only renaming it. So, let's move forward with it we can do a follow up to improve the understandability/maintability always.

PR #1843, opened 2020-11-18
:seedling: Makefile,build/.goreleaser.yml: fix ldflag variable setting

estroz:
This PR fixes `-ldflags -X ...` settings in `build` and `install` make targets and in goreleaser's config, which should point to the `main` package.

/cc @Adirio


Adirio:
I had no idea we could target `main` directly. No real experience with goreleaser aside from editing the path.
/lgtm

PR #1844, opened 2020-11-18
:seedling: add unit tests for the config

camilamacedo86:
**Description**

Add unit tests for the config

**Motivation**

Increase the coverage (coverage/coveralls — Coverage increased (+1.2%) to 73.719%)

### DIFF
--- a/pkg/model/config/config_test.go
+++ b/pkg/model/config/config_test.go
@@ -147,6 +147,22 @@ var _ = Describe("PluginConfig", func() {
 		Expect(config.DecodePluginConfig(key, &pluginConfig)).To(Succeed())
 		Expect(pluginConfig).To(Equal(expectedPluginConfig))
 	})
+
+	It("should Marshal and Unmarshal a plugin", func() {
+		By("Using config with extra fields as struct")
+		cfg := Config{
+			Version: Version3Alpha,
+			Plugins: PluginConfigs{
+				"plugin-x": map[string]interface{}{
+					"data-1": "plugin value 1",
+				},
+			},
+		}
+		b, err := cfg.Marshal()
+		Expect(err).NotTo(HaveOccurred())
+		Expect(string(b)).To(Equal("version: 3-alpha\nplugins:\n  plugin-x:\n    data-1: plugin value 1\n"))
+		Expect(cfg.Unmarshal(b)).To(Succeed())
+	})
 })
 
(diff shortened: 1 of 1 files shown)


pkg/model/config/config_test.go
```diff
@@ -147,6 +147,39 @@ var _ = Describe("PluginConfig", func() {
 		Expect(config.DecodePluginConfig(key, &pluginConfig)).To(Succeed())
 		Expect(pluginConfig).To(Equal(expectedPluginConfig))
 	})
+
+	It("should Marshal and umarchal a plugin", func() {
```
Adirio, line 151: Typo:
```suggestion
	It("should Marshal and Unmarshal a plugin", func() {
```

pkg/model/config/config_test.go
```diff
@@ -147,6 +147,39 @@ var _ = Describe("PluginConfig", func() {
...
+				"plugin-x": map[string]interface{}{
+					"data-1": "plugin value 1",
+				},
+			},
+		}
+		b, err := cfg.Marshal()
+		Expect(err).To(BeNil())
+		Expect(string(b)).To(Equal("version: 3-alpha\nplugins:\n  plugin-x:\n    data-1: plugin value 1\n"))
+		Expect(cfg.Unmarshal(b)).To(Succeed())
+	})
```
Adirio, line 182: Duplicate?
```suggestion
```

Adirio (changes requested):

pkg/model/config/config_test.go
```diff
@@ -147,6 +147,39 @@ var _ = Describe("PluginConfig", func() {
...
+		cfg := Config{
+			Version: Version3Alpha,
+			Plugins: PluginConfigs{
+				"plugin-x": map[string]interface{}{
+					"data-1": "plugin value 1",
+				},
+			},
+		}
+		b, err := cfg.Marshal()
+		Expect(err).To(BeNil())
```
Adirio, line 163: Usually `Expect(err).NotTo(HaveOccurred())` is used with `error` variables and `Expect(SomeFunc()).To(Succeed())` with direct function calls (they do exactly the same but they are easier to read like this). `Expect(err).To(BeNil())` would do also the same but it is less clear.
```suggestion
		Expect(err).NotTo(HaveOccurred())
```

Adirio (approved):
/lgtm

PR #1846, opened 2020-11-18
:sparkles: add make run option for webhooks (go/v3-alpha)

camilamacedo86:
**Description**

Add make run option for webhooks

**Motivation**
- Keep the consistence across the project
- Allows users scaffold webhooks without run make targets


Adirio (approved):
/lgtm

Issue #1826, opened 2020-11-16
Project file with all user input data


camilamacedo86:
**Description**

Ensure that we are storing all input data in the PROJECT file. 

**Motivation**
I am maintainer like to ensure that we have stored all input data used to scaffold the projects by the user so that I could propose a subcommand/plugin to help users migrate from a plugin version X to X+1 since would be possible scaffold the project from the scratch. Also, it might be helpful in many scenarios to tooling the scaffolds. 

Mor info: See: https://github.com/kubernetes-sigs/kubebuilder/pull/1644#discussion_r491227667

**What is missing to be stored**

- [ ] Webhooks data
- [ ] resource/controller=[true|false]  

/kind feature

camilamacedo86:
PR related : https://github.com/kubernetes-sigs/kubebuilder/pull/1696

camilamacedo86:
Hi @estroz, @joelanford, @DirectXMan12, @pwittrock  I am pushing it for v3+plugin for we have the chance to discuss it and try to achieve it for the release if we agree on that.

estroz:
>hasResource and hasController for the resources

Maybe only hasResource? If you're building a type library you would set `--controller=false --resource=true`, so we'd probably want to set `hasResource: true` or `types: true` field for its GVK. Conversely, one major reason you would have `--resource=false --controller=<true|false>` set is if you're using an external type; the config can store an external type path as something like `external: <import path>` (see #1772) instead of `controller: true`.

>from the init owner and license

These are already captured in `hack/boilerplate.go.txt` so these don't need to be tracked in the config.

camilamacedo86:
Hi @estroz, 

> These are already captured in hack/boilerplate.go.txt so these don't need to be tracked in the config. 👍 

Totally agree. Good point. 

> Maybe only hasResource?

Since we can resource = false and controller=true as resource = true and controller=false, IMO this one we need to track.

estroz:
Prior to releasing v3.0.0, the config spec for version `3-alpha` needs to be stabilized. A few changes have been added to or are proposed for `3-alpha`. To summarize:
- Added `resources[].crdVersion` and `resources[].webhookVersion` when `--crd-version` and `--webhook-version` are set, respectively (#1644)
- Add `resources[].api` and `resources[].controller` when `--resource` and `--controller` are set, respectively (#1849)
- Add `resources[].webhooks` for defaulting, validating, and conversion webhooks (#1696)

[...]

Issue #1772, opened 2020-11-02
[Discussion] V3 breaking changes


Adirio:
Now that we are working on `v3`, it may be time to go ahead and do some of this changes that we can't do on an stable version (such as `v2`). This issue is meant to be a discussion to see which of them shows enough interest by the developers/community in order to implement them.

1. Split CRD and controller scaffolding into two different subcommands. In `v1` and `v2`, `kubebuilder create api` crteates both the CRD and the controller. Do we still think this is the desired behavior? Should we have a separate subcommand for the controller and eliminate the `--resource` and `--controller` flags? Should these two subcommands replace the current one or should we have three (`create resource` for the resource only, `create controller` for the controller only, `create api` for both of them)?
2. Add external resource flag to controller creating subcommand. When this flag is passed, kubebuilder won't search for the corresponding resource and will also take it into account at the tests, and when it is not passed the controller will validate that the appropiate resource is created, either by this same call, by a previous call, or even by hand.

If you think there are other related topics we should discuss, comment and I will add them here.

/cc @droot @DirectXMan12 @mengqiy @pwittrock @camilamacedo86 @estroz @joelanford

camilamacedo86:
IMO:

**Over split the command create api** 

- Pros:  keep concepts such as cohesion and single responsibility in the commands/plugins (maintainability)
- Cons: GKV would be required in both which means that the user would need to inform them twice

So, shows for me that we might be prevailing keep it ease to maintaining instead of user experience. In this way, I am afraid that I am more inclined to vote against since I think the UX is more important here. 

**Over new flag to work with external types**

I totally agree that we need to address this scenario. WDYT about we raise a new specific issue for this RFE? Also, instead of not scaffold the files I think we ought to educate and facilitate to the users how to work with for the external type which could be something such as we have a new plugin for that since we will not create an api but we will add one preexistent into the project.

Example: `kubebuiler add api  --module=github.com/openshift/api --kind=route --version=v1`

- Run `go get -u go get -u github.com/openshift/api` // run go get -u {{module}}
- Add the schema into the main.go: 

```
import (
	...
	routev1 "github.com/openshift/api/route/v1"  // see {{kindversion}} {{module}} 
	...
)
```
-  add in the external type in the suite_test.go as well. 

```
       	err = routev1.AddToScheme(scheme.Scheme)  // see {{kindversion}}
	Expect(err).NotTo(HaveOccurred())
```

estroz:
[...]

PR #1849, opened 2020-11-18
:sparkles: (go/v3-alpha) store data info to know if the controller and/or api were scaffolded

camilamacedo86:
**Description**
- Add api and controller booleans to the config model
- Set their value when the scaffold of these types are made. 
- Ensure that if only the controller for a gkv was scaffold we will have this info stored. 

**Motivation**
Partial: https://github.com/kubernetes-sigs/kubebuilder/issues/1826

### DIFF
--- a/pkg/model/resource/resource.go
+++ b/pkg/model/resource/resource.go
@@ -49,13 +49,19 @@ type Resource struct {
 	// Domain is the Group + "." + Domain of the Resource.
 	Domain string `json:"domain,omitempty"`
 
-	// Namespaced is true if the resource is namespaced.
-	Namespaced bool `json:"namespaced,omitempty"`
-
 	// CRDVersion holds the CustomResourceDefinition API version used for the Resource.
 	CRDVersion string `json:"crdVersion,omitempty"`
 	// WebhookVersion holds the {Validating,Mutating}WebhookConfiguration API version used for the Resource.
 	WebhookVersion string `json:"webhookVersion,omitempty"`
+
+	// API holds true when the api is scaffolded
+	API bool `json:"api,omitempty"`
+
+	// Controller holds true when the controller is scaffolded
+	Controller bool `json:"controller,omitempty"`
+
+	// Namespaced is true if the resource is namespaced.
+	Namespaced bool `json:"namespaced,omitempty"`
 }
 
 // GVK returns the group-version-kind information to check against tracked resources in the configuration file
@@ -66,6 +72,8 @@ func (r *Resource) GVK() config.GVK {
 		Kind:           r.Kind,
 		CRDVersion:     r.CRDVersion,
 		WebhookVersion: r.WebhookVersion,
+		API:            r.API,
+		Controller:     r.Controller,
 	}
 }
 
--- a/pkg/plugins/golang/v3/api.go
+++ b/pkg/plugins/golang/v3/api.go
@@ -170,7 +170,8 @@ func (p *createAPISubcommand) Validate() error {
 	// In case we want to scaffold a resource API we need to do some checks
 	if p.doResource {
 		// Check that resource doesn't exist or flag force was set
-		if !p.force && p.config.HasResource(p.resource.GVK()) {
+		res := p.config.GetResource(p.resource.GVK())
+		if !p.force && (res != nil && res.API) {
 			return errors.New("API resource already exists")
 		}
 
@@ -209,7 +210,7 @@ func (p *createAPISubcommand) GetScaffolder() (cmdutil.Scaffolder, error) {
 	}
 
 	// Create the actual resource from the resource options
-	res := p.resource.NewResource(p.config, p.doResource)
+	res := p.resource.NewResource(p.config, p.doResource, p.doController)
 	return scaffolds.NewAPIScaffolder(p.config, string(bp), res, p.doResource, p.doController, plugins), nil
 }
 
--- a/testdata/project-v3-addon/PROJECT
+++ b/testdata/project-v3-addon/PROJECT
@@ -3,15 +3,21 @@ layout: go.kubebuilder.io/v3-alpha
 projectName: project-v3-addon
 repo: sigs.k8s.io/kubebuilder/testdata/project-v3-addon
 resources:
-- crdVersion: v1
+- api: true
+  controller: true
+  crdVersion: v1
   group: crew
   kind: Captain
   version: v1
-- crdVersion: v1
+- api: true
+  controller: true
+  crdVersion: v1
   group: crew
   kind: FirstMate
   version: v1
-- crdVersion: v1
+- api: true
+  controller: true
+  crdVersion: v1
   group: crew
   kind: Admiral
   version: v1
--- a/pkg/model/config/config.go
+++ b/pkg/model/config/config.go
@@ -78,17 +78,14 @@ func (c Config) IsV3() bool {
 	return c.Version == Version3Alpha
 }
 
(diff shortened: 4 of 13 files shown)


pkg/plugins/golang/v3/api.go
```diff
@@ -210,6 +211,8 @@ func (p *createAPISubcommand) GetScaffolder() (cmdutil.Scaffolder, error) {
 
 	// Create the actual resource from the resource options
 	res := p.resource.NewResource(p.config, p.doResource)
+	res.API = p.doResource
+	res.Controller = p.doController
```
Adirio, line 215: I think that it will be cleaner if you:
```suggestion
	res := p.resource.NewResource(p.config, p.doResource, p.doController)
```
And then set those two fields in resource creation. Right now you can't do that because you dont want to set them for `v2` project files for backwards compatibility. But could we leave a comment about this:
```suggestion
	// TODO: API & Controller should be set inside `Options.NewResource` once v2 is no longer supported.
	res := p.resource.NewResource(p.config, p.doResource)
	res.API = p.doResource
	res.Controller = p.doController
```

Adirio (changes requested):

pkg/plugins/golang/v3/api.go
camilamacedo86: Really thank you for your input. I liked it and i addressed your suggestion. 
See that for we do not change the current behaviour for v2 we can only check the if is v2 version.

pkg/model/resource/resource.go
```diff
@@ -24,6 +24,7 @@ import (
 )
 
 // Resource contains the information required to scaffold files for a resource.
+// nolint:maligned
```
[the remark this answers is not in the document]
camilamacedo86: The Namespaced was not at the end. 
For we do not need the nolint check we need to change the order. I did that.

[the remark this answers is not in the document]
Adirio: You are right, `namespaced` was not at the end, but it wasn't at the end from previous commits either. Some lint issues seem to manifest themselves some PR after the one that changed them. Guess we need to keep an eye open for these changes. You can reoslve this.

Adirio:
NIT: You should remove the `Partial closes: #1826`  as prow doesn't understand what partial means. Leave it like `Partial: #1826` and it won't close it.

From a code perspective this LGTM. But I would like other reviewer to say if they like the idea of adding all this info to the config file, as I'm not completely convinced.

pkg/model/resource/resource.go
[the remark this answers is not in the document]
camilamacedo86: Usually to an easier way to solve the check is by adding the bools at the end. However, the maligned check is related to memory space usage. See: https://medium.com/@sebassegros/golang-dealing-with-maligned-structs-9b77bacf4b97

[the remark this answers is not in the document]
Adirio: Yes, yes. I mean mark it as resolved.

camilamacedo86:
Hi @Adirio, 

> From a code perspective this LGTM. But I would like other reviewer to say if they like the idea of adding all this info to the config file, as I'm not completely convinced.

It is fine 👍 
Note that the goal here is we be allowed to create migration tools for example in the future from one plugin to another. 
Without this info, we are unable to re-create the project from the scratch via the PROJECT file.

Adirio:
> It is the goal. We have 2 PR;'s that needs to get merged for we are able o close #1826. So, it will probably get done manually after that.

The problem is that any of them will close the Issue because it sees closes and the issue id in the same sentence. In order to do it manually you need to remove the `closes` word. If you don't, the issue will be closed by the first PR to get merged even if the other PR doesn't get merged.

testdata/project-v3-addon/PROJECT
```diff
@@ -3,15 +3,21 @@ layout: go.kubebuilder.io/v3-alpha
 projectName: project-v3-addon
 repo: sigs.k8s.io/kubebuilder/testdata/project-v3-addon
 resources:
```
estroz, line 5: With a change like this, I wonder if its now worth renaming this field to `api` to match `create api --resource --controller`, which is effectively saying "create an api for this GVK, wiring resource types and a controller if specified":

```yaml
apis:
- resource: true
  controller: true
  gvk:
   group: crew
   version: v1
   kind: Captain
  ...
```

Relates to https://github.com/kubernetes-sigs/kubebuilder/issues/1772#issuecomment-723318555.

camilamacedo86: Hi @estroz, 

Following the concerns regards the suggestion: 

- The create api creates API and not the resource. See that we can have a controller without an api.  
```
apis:
- resource: true
```

- The api and controllers are for the GKV used so we cannot does as suggested. See that each GKV has a controller and a resource related. 

```
- resource: true
  controller: true
  gvk:
 -  group: crew
 -  version: v1
 -  kind: Captain
```
 
IMO the scope of this PR is to store the data. Then, re-design it shows out of the scope, however, I am 100% all in to discuss if we can improve it in a follow-up. 

c/c @Adirio

Adirio: I don't dislike changing the name and I think that API may fit. API = [Resource] + [Controller] + [Webhooks] (all optional).

camilamacedo86: Hi folks, 

By extending Kubernetes API we can create our own apis whcih is the CRDs. A controller is NOT an API at all. Then, because of this I do not agree with the rename of resources per apis. However, it shows not part of the scope of this PR as well. IMO a redesign or new nomenclature proposals would better fit as a new issue for we discuss that and can be applied as a follow-up.

estroz: >IMO the scope of this PR

While I agree, I also think we've been changing the config file 3-alpha spec a _lot_ recently without a concerted view of these changes, which worries me because we may be fragmenting UX and treating `PROJECT` as something the user should never look at (which they will probably have to at some point). Personally I'd like to see a draft EP for #1826 before adding these fields, one that discusses a general methodology for adding/modifying the existing `resources` field with types created by `create` subcommands.

camilamacedo86: I can work on in an EP as follow up for we are able to address the suggestion made so far. 👍

camilamacedo86: Pushed an doc for the purposed changes + to doc the suggestion made here to add the new attr gkv. See: https://github.com/kubernetes-sigs/kubebuilder/pull/1861/files

estroz: https://github.com/kubernetes-sigs/kubebuilder/issues/1826#issuecomment-732460408

estroz:
Suggested alternative on where to add these fields: https://github.com/kubernetes-sigs/kubebuilder/issues/1826#issuecomment-732460408.

testdata/project-v3-addon/PROJECT
camilamacedo86: Hi @estroz,

See the comment: https://github.com/kubernetes-sigs/kubebuilder/pull/1849#issuecomment-734230778

camilamacedo86:
> Suggested alternative on where to add these fields: #1826 (comment).
It is out of the sope of the proposal of this change.

camilamacedo86:
Closed in favor of : :sparkles: store all config info (go/v3-alpha) #1696
my initial idea was to do a PR to start to track each piece part and address it step by step. However, then a lot of feedbacks and suggestions for changes were raised. All that was done so far is in #1696

PR #1851, opened 2020-11-18
:seedling: .gitignore,generate_testdata.sh: cleanup

estroz:
- .gitignore: combine bin and testbin ignore statements
- generate_testdata.sh: remove unnecessary conditions, have scaffold_test_project accept flags from arg 2 onwards

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,27 +11,13 @@ docs/book/book/
 # Skip bazel dirs
 /bazel-*
 
-# skip bin
-bin/*
-testbin/*
-testdata/project-v3-addon/testbin/*
-testdata/project-v3-multigroup/testbin/*
-testdata/project-v3/testbin/*
-docs/book/src/multiversion-tutorial/testdata/project/bin/*
-docs/book/src/cronjob-tutorial/testdata/project/bin/*
+# skip bin dirs
+**/bin/*
+**/testbin/*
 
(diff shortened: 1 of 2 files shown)


camilamacedo86:
/lgtm 

/hold
Let's merge https://github.com/kubernetes-sigs/kubebuilder/pull/1825 first to avoid another rebase.
However, after that feel free to cancel the hold

estroz:
@camilamacedo86 since #1825 should probably be merged as one of the last commits before cutting `v3.0.0-alpha.0`, I'm not going to block this on #1825. Side note: there will likely be more conflicts with #1825 from other PRs prior to cutting the alpha tag unfortunately.

/hold cancel

camilamacedo86:
HI @estroz, 

> @camilamacedo86 since #1825 should probably be merged as one of the last commits before cutting v3.0.0-alpha.0, I'm not going to block this on #1825. Side note: there will likely be more conflicts with #1825 from other PRs prior to cutting the alpha tag unfortunately. 

Is possible to do a patch release from mater branch such as 2.X ? I do not think so. See that it has to break changes that do not allow the projects built with `2.3.1` still working on with the next release such as we remove the v1 support and etc. IMO we should not actually never bump v2/ at all.  Anyway, the next release requires to be an upper version 3.X.

PR #1853, opened 2020-11-19
:seedling: build/: revamp release process

### LINKED ISSUE #686: Release kubebuilder and k8s bundles separately

### LINKED ISSUE #1694: snapshot/release cloudbuild config is b0rked
The cloudbuild configs in build/cloudbuild.yaml & build/cloudbuild_snapshot.yaml are b0rked -- they use Go 1.12 to build, and we've long since required Go 1.13.  We should fix them up.

While we're at it, they seem to still be using go-releaser, which I think we wanted to switch off of.  We should start populating kubernetes-sigs/kubebuilder-release-tools with common releasing infrastructure, and maybe consider moving over to that where possible

cc @vincepri who I think had some ideas around this

estroz:
This PR changes how kubebuilder releases are done by removing packaged envtest binaries from the release payload; these binaries can be downloaded for most k8s versions as they are built separately on the tool-releases branch. It also cleans up old build config and script files, bumps Go version to v1.15.3, and bumps Kubernetes version to v1.19.2.

- build/.goreleaser.yml: update format to match goreleaser 0.147 spec and only build/release binaries for each platform
- build/cloudbuild*.yaml: change these to only depend on build_kubebuilder.sh in local/snapshot/prod steps
- build/build_kubebuilder.sh: clean up and add release notes generator

Closes #686
Closes #1694


DirectXMan12:
> build/cloudbuild*.yaml: change these to only depend on

I await the end of this bullet point with baited breath, suspense gripping my heart

(is it supposed to just be the next line?)

build/.goreleaser.yml
```diff
@@ -14,50 +14,57 @@
...
+      - -X main.goos={{ .Os }}
+      - -X main.goarch={{ .Arch }}
+      - -X main.gitCommit={{ .Commit }}
+      - -X main.buildDate={{ .Date }}
+      - -X main.kubernetesVendorVersion={{ .Env.KUBERNETES_VERSION }}
+    targets:
+      - linux_amd64
+      - linux_arm64
+      - linux_ppc64le
+      - darwin_amd64
```
DirectXMan12, line 45: one more here soon :-p

build/build_kubebuilder.sh
```diff
@@ -13,46 +14,64 @@
...
+  esac
 done
 
-/goreleaser release --config=build/.goreleaser.yml --rm-dist --skip-validate ${SNAPSHOT}
+# Generate real release notes.
+if [ -z "$SNAPSHOT" ]; then
+  tmp_notes="$(mktemp)"
+  trap "rm -f ${tmp_notes}" EXIT
+  install_notes
+  notes | tee "$tmp_notes"
```
DirectXMan12, line 73: FYI, there's special flags for betas and whatnot, might want to feed them through here depending on the version being tagged

DirectXMan12 (changes requested):
couple of comments inline, otherwise looks good

build/.goreleaser.yml
```diff
@@ -14,50 +14,57 @@
...
+  - format: binary
+    # Setting name_template correctly maps checksums to binary names.
+    name_template: "{{ .Binary }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}{{ if .Mips }}_{{ .Mips }}{{ end }}"
+
+# Checksum all binaries.
 checksum:
-  name_template: 'checksums.txt'
+  name_template: "checksums.txt"
+
+# Use "master" as main.kubeBuilderVersion when running a snapshot.
```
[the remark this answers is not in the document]
estroz: This was a existing setting and didn't change it. We could use the default `{{ .Tag }}-SNAPSHOT-{{.ShortCommit}}` here.

build/build_kubebuilder.sh
estroz: Yeah I was wondering about that. Since this script is run automatically by cloudbuild, how would the flags being passed be determined?

build/.goreleaser.yml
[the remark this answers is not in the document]
estroz: I'll remove this.

build/build_kubebuilder.sh
estroz: You can now set these flags in the `NOTES_FLAGS` env var.

DirectXMan12: ack.  The idea is that we'd determine the flags automatically from the tag this is being released against, but we should really reverse that and have a release notes tooling that that says "given a version, determine if we should be in alpha/beta/whatever mode", but we can do that in a follow-up

PR #1861, opened 2020-11-23
:book: add doc over proposed changes to store the data in the PROJECT…

camilamacedo86:
**Description**
- EP doc with the proposed changes to be addressed in the PROJECT file. 

**Motivation**
- https://github.com/kubernetes-sigs/kubebuilder/pull/1849/files#r528995881

### DIFF
--- a/designs/store-input-data-in-project-file.md
+++ b/designs/store-input-data-in-project-file.md
@@ -1,0 +1,94 @@
+---
+title: Project file with all user input data
+authors:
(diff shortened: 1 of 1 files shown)


camilamacedo86:
Hi @estroz, 

IMO this EP is no required since the proposed changes so far are very small and do not cover any re-design at all. 
So, I am closing this one. However, we can re-open if you think that it would be required. 

c/c @Adirio

PR #1864, opened 2020-11-30
:sparkles: Changed the default shell to bash in the doc Makefile

naveensrinivasan:
Makefile default shell is `/bin/sh` which does not implement source `source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh` which is in the `test` target within the Makefile.

Changing shell to `/bin/bash` makes it possible:

 https://stackoverflow.com/a/43566158/19407


estroz:
@naveensrinivasan this is fixed by #1725

naveensrinivasan:
Thanks, @estroz!

PR #1868, opened 2020-12-02
✨ Remove spaces from machine-readable comments

Adirio:
# Description
Machine-readable comments used for markers have spaces after the comment character (`//` or `#`), they have been removed to follow common coding practices.

Closes: #1384

# Comments to reviewers
131/144 files are from the `testdata` directory, this PR is smaller than what it looks like by the number of files changed.

### DIFF
--- a/pkg/model/file/marker.go
+++ b/pkg/model/file/marker.go
@@ -19,15 +19,14 @@ package file
 import (
 	"fmt"
 	"path/filepath"
+	"strings"
 )
 
 const prefix = "+kubebuilder:scaffold:"
 
 var commentsByExt = map[string]string{
-	// TODO(v3): machine-readable comments should not have spaces by Go convention. However,
-	//  this is a backwards incompatible change, and thus should be done for next project version.
-	".go":   "// ",
-	".yaml": "# ",
+	".go":   "//",
+	".yaml": "#",
 	".yml":  "#",
 	// When adding additional file extensions, update also the NewMarkerFor documentation and error
 }
@@ -52,6 +51,12 @@ func NewMarkerFor(path string, value string) Marker {
 // String implements Stringer
 func (m Marker) String() string {
 	return m.comment + prefix + m.value
+}
+
+// EqualsLine compares a marker with a string representation to check if they are the same marker
+func (m Marker) EqualsLine(line string) bool {
+	line = strings.TrimSpace(strings.TrimPrefix(strings.TrimSpace(line), m.comment))
+	return line == prefix+m.value
 }
 
 // CodeFragments represents a set of code fragments
--- a/pkg/plugins/internal/machinery/scaffold.go
+++ b/pkg/plugins/internal/machinery/scaffold.go
@@ -343,7 +343,7 @@ func insertStrings(content string, codeFragmentsMap file.CodeFragmentsMap) ([]by
 		line := scanner.Text()
 
 		for marker, codeFragments := range codeFragmentsMap {
-			if strings.TrimSpace(line) == strings.TrimSpace(marker.String()) {
+			if marker.EqualsLine(line) {
 				for _, codeFragment := range codeFragments {
 					_, _ = out.WriteString(codeFragment) // bytes.Buffer.WriteString always returns nil errors
 				}
--- a/testdata/project-v3/api/v1/webhook_suite_test.go
+++ b/testdata/project-v3/api/v1/webhook_suite_test.go
@@ -29,7 +29,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
-	// +kubebuilder:scaffold:imports
+	//+kubebuilder:scaffold:imports
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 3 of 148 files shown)


camilamacedo86:
Hi @Adirio, 

It is not passing in the CI. Could you please check this one for we get it merged. 
Shows that it is missing run make generate only.

Adirio:
I think it is a rebase what it is needed. Will try it.

estroz (commented):
Something like this can be done for any plugin, stable/frozen or not, since machine-readable comments with spaces are backwards compatible.

camilamacedo86:
Hi @Adirio 

What was your motivation to define that #1384 (this change) would be a braking change and that should be done only for v3? If it is backwards compatible with v2 then I agree with @estroz in https://github.com/kubernetes-sigs/kubebuilder/pull/1868#pullrequestreview-550677466. 

However, i imagine that you saw some scenario that would be affected. so, could you please clarify?

Adirio:
If we apply this to v2, projects scaffolded with v2 before applyign this change will have spaces and further update operation by kubebuilder will not be able to find the placeholders. For example the ones in `mian.go` to register the controllers in the manager.

estroz:
> further update operation by kubebuilder will not be able to find the placeholders

If that’s true, it sounds like a bug in controller-tools’ comment lexer. Both forms are valid marker comments so should be parsed in the same manner.

Adirio:
I mean the comments that are meant for kubebuilder itself, for example the `MainUpdater` inserts code fragments to register new controllers. Those fragments are inserted just before a machine-readabale comment and it does use exact match.

estroz:
What about creating a method to compare marker comments like so

```go
func (m Marker) EqualsLine(line string) bool {
	line = strings.TrimSpace(strings.TrimPrefix(m.comment))
	return line == m.value
}
```

and use it [here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/internal/machinery/scaffold.go#L346):

```diff
		line := scanner.Text()
		for marker, codeFragments := range codeFragmentsMap {
-			if strings.TrimSpace(line) == strings.TrimSpace(marker.String()) {
+			if marker.EqualsLine(line) {
				...
			}
		}
```

estroz:
The above is just a suggestion btw. Your solution also works. Feel free to unhold to merge as-is.

/lgtm
/hold

Adirio:
/cancel hold

@estroz I liked your suggestion, removed the hold as the `lgtm` label is required again

pkg/model/file/marker.go
```diff
@@ -19,15 +19,14 @@ package file
...
 const prefix = "+kubebuilder:scaffold:"
 
 var commentsByExt = map[string]string{
-	// TODO(v3): machine-readable comments should not have spaces by Go convention. However,
-	//  this is a backwards incompatible change, and thus should be done for next project version.
-	".go":   "// ",
-	".yaml": "# ",
+	".go":   "//",
+	".yaml": "#",
 	".yml":  "#",
```
camilamacedo86, line 30: Tks 👍 for addressing : https://github.com/kubernetes-sigs/kubebuilder/pull/1907

pkg/plugins/internal/machinery/scaffold.go
```diff
@@ -343,7 +343,7 @@ func insertStrings(content string, codeFragmentsMap file.CodeFragmentsMap) ([]by
 		line := scanner.Text()
 
 		for marker, codeFragments := range codeFragmentsMap {
-			if strings.TrimSpace(line) == strings.TrimSpace(marker.String()) {
+			if marker.EqualsLine(line) {
```
camilamacedo86, line 346: if we added here || strings.TrimSpace(line) == strings.TrimSpace(marker.String()) would not solve for both scenarios?

camilamacedo86:
Hi @Adirio,

it is not passing in the CI.  Maybe it is missing a rebase?

Adirio:
I don't know why but seems that one of the changes in `*_webhook.go` files seems to be breaking the webhooks. @estroz do you know if this should be considered a parsing error from the side of controller-tools?

Adirio:
The part that is failing is this one: https://github.com/kubernetes-sigs/kubebuilder/blob/d501fc35a6deed5aac41b116a0f26182d85d99f3/testdata/project-v3/api/v1/webhook_suite_test.go#L126-L136

I'm not modifying anything that should break that check.

prafull01 (commented):
I have looked into the code and the reason for this failure is following scaffolding code is not generated by kubebuilder:
```
err = (&Captain{}).SetupWebhookWithManager(mgr)
```
Attaching the screenshot of failure and missing piece of code from scaffold code.
Screenshot from 2020-12-17 01-31-38
Screenshot from 2020-12-17 01-31-48

testdata/project-v3/api/v1/webhook_suite_test.go
```diff
@@ -114,7 +114,7 @@ var _ = BeforeSuite(func() {
 	err = (&Captain{}).SetupWebhookWithManager(mgr)
 	Expect(err).NotTo(HaveOccurred())
```
Adirio, line 115: It is scaffolded here, right?

prafull01: I think you have not run make generate before pushing the change. I cloned your repo and generated a project with that, it doesn't generate that code. Also, I tried to run the `make generate` locally and it is throwing same error. Could you please try that?

prafull01 (approved):

estroz (approved):
/approve
/lgtm

Issue #1854, opened 2020-11-20
Finalizer docs use invalid finalizer name


salaxander:
https://github.com/kubernetes-sigs/kubebuilder/blob/2c3ae8fd18aa9c39326bd9f8776cc1f4db71fb0c/docs/book/src/cronjob-tutorial/testdata/finalizer_example.go#L54

Finalizers have to be qualified with a `/` in the name. The book currently uses `storage.finalizers.tutorial.kubebuilder.io` as an example finalizer. This should be changed to something like `tutorial.kubebuilder.io/finalizer`.

Happy to open a PR, just wanted to start with an issue to confirm that folks were ok with that!

gabbifish:
Hi @salaxander! Thanks for bringing this up :) I think this is an important fix for us to have, and one I'd be happy to review when you have it ready! 

Consistent with the [finalizer example in the kubernetes docs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#finalizers), where the CRD's group is used in the finalizer, perhaps we should rename the finalizer in this example to `batch.tutorial.kubebuilder.io/finalizer`?

estroz:
Closed by #1877

/close

PR #1877, opened 2020-12-03
📖 Change example finalizer name to be valid

salaxander:
* This changes the name of the example finalizer used in the book to be a valid name
* I was using the book as a template and ran into errors by modeling my finalizer off of the provided example. This should hopefully help others avoid the same mistake!
* Addresses issue #1854


estroz (approved):
/approve
/lgtm

PR #1883, opened 2020-12-06
:seedling: fix the breaking change to still allowing TestContext to be used for tests outside of the cluster 

camilamacedo86:
**Description**

After the changes made in https://github.com/kubernetes-sigs/kubebuilder/commit/ea3a94a97ad4ba0cb148e711095530286d050d54#diff-7b614da29899d7d7eb4aa3b60ed598604da518c4f2dd1e6af954c942fb1c3dc3 in many scenarios, we have been facing panic issues when it was calling the kubectl.Version(). See that TextContext has *kubectl which can be nil. 

Note that, we ought to just look for the k8sVersion when/if it is required. It is not a mandatory attrs at all we might have tests that require knowing the k8sVersion and others that not.

### DIFF
--- a/test/e2e/utils/test_context.go
+++ b/test/e2e/utils/test_context.go
@@ -59,10 +59,13 @@ func NewTestContext(binaryName string, env ...string) (*TestContext, error) {
 		Namespace:  fmt.Sprintf("e2e-%s-system", testSuffix),
 		CmdContext: cc,
 	}
-	k8sVersion, err := kubectl.Version()
-	if err != nil {
-		return nil, err
-	}
+
+	// Try to populate the k8sVersion and ignore any error that might be faced.
+	// Note that, the k8sVersion is not an mandatory info and it is
+	// not required for ANY context of tests. For example, the testContext
+	// might be used to run a test outside of the cluster where this info is not
+	// available and not required at all.
+	k8sVersion, _ := kubectl.Version()
 
 	// Set CmdContext.Dir after running Kubectl.Version() because dir does not exist yet.
 	if cc.Dir, err = filepath.Abs("e2e-" + testSuffix); err != nil {
--- a/test/e2e/v3/plugin_cluster_test.go
+++ b/test/e2e/v3/plugin_cluster_test.go
@@ -86,6 +86,13 @@ var _ = Describe("kubebuilder", func() {
 			})
 
 			It("should generate a runnable project", func() {
+				// Ensure that kbc.K8sVersion is populated.
+				if kbc.K8sVersion == nil {
+					K8sVersion, err := kbc.Kubectl.Version()
+					Expect(err).NotTo(HaveOccurred())
+					kbc.K8sVersion = &K8sVersion
+				}
+
 				// Skip if cluster version < 1.16, when v1 CRDs and webhooks did not exist.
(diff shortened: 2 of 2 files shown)


test/e2e/utils/test_context.go
```diff
@@ -50,36 +50,30 @@ func NewTestContext(binaryName string, env ...string) (*TestContext, error) {
...
 		Resources:  "foo" + testSuffix + "s",
 		ImageName:  "e2e-test/controller-manager:" + testSuffix,
 		CmdContext: cc,
-		Kubectl:    kubectl,
-		K8sVersion: &k8sVersion,
+		Kubectl: &Kubectl{
+			Namespace:  fmt.Sprintf("e2e-%s-system", testSuffix),
+			CmdContext: cc,
+		},
 		BinaryName: binaryName,
```
camilamacedo86, line 77: Just revert the change made in: https://github.com/kubernetes-sigs/kubebuilder/commit/ea3a94a97ad4ba0cb148e711095530286d050d54#diff-7b614da29899d7d7eb4aa3b60ed598604da518c4f2dd1e6af954c942fb1c3dc3 and keep the k8sVersion

estroz:
Can you describe why you wouldn’t want cluster version lookup in the constructor? This will only panic if cluster version isn’t available (no running cluster), which is a nice fail-early check. There’s a fix you can make instead of removing version lookup from the constructor.

/hold

estroz (changes requested):
Cluster version is required by other [e2e setups steps](https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/utils/test_context.go#L120), which is why the lookup exists in the constructor. `TestContext` is meant for live cluster testing, so having a running cluster requirement is reasonable.

camilamacedo86:
Hi @estroz, 

> Cluster version is required by other e2e setups steps, which is why the lookup exists in the constructor. 

it is NOT true at all. It is ONLY required on the current tests after the changes made in https://github.com/kubernetes-sigs/kubebuilder/pull/1840 to use a different version of cert-manager. Cannot we write a test that does not use cert-manager at all? Should we not probably able to remove the K8SVersion in the future if we wish when we no longer need to run the tests against k8s versions 1.16 if it pleases us? So, how it can now be mandatory for ANY test done with TestContext?  

> TestContext is meant for live cluster testing, so having a running cluster requirement is reasonable.

The change made in https://github.com/kubernetes-sigs/kubebuilder/pull/1840 defines that TestContext can ONLY be used when is possible to get the K8SVersion which I cannot agree that is reasonable. 

All tests implemented on this repository that uses the TestContext currently runs on the cluster, however, it does not mean that TestContext can ONLY be used to run a test on Cluster and that it is its ONLY purpose. And then, can we not have an e2e tests that not run on the cluster? 

 I'd assume that it is in `test/e2e/utils` only because it has not been used outside of this directory so far which might not be the case in the future. Note that, the TestCntext could indeed be used to replace the shell scripts in [generate_testdata.sh](https://github.com/kubernetes-sigs/kubebuilder/blob/master/generate_testdata.sh) by similar we do in SDK to create the testdata samples.

**Conclusion**;  I cannot see any advantage and value by adding this limitation whcih in POV is actually an issue added on https://github.com/kubernetes-sigs/kubebuilder/pull/1840 or at least a bad practice to start to make mandatory values that not are really required for ANY and ALL possible tests scenarios on the constructor to the test stack.

Issue #1651, opened 2020-08-24
Chinese version for Kubebuilder docs are available and want to be merged.


majinghe:
Hi team
Cloud Native Community in China already translated the Kubebuilder docs into Chinese, now Chinese version is available on https://cloudnative.to/kubebuilder/. Community wants to confirm is it possible to merge the Chinese docs into the Kubebuilder main repo? Thanks

camilamacedo86:
Hi @lhb008, 

I think it shows great. Unfortunately, I believe that none of its owners/admins knows Chinese and would be really hard for our keep maintained many languages.it will need to rely on the community contributions. So, IMO we need clarifies that the English docs are mainly maintained just in case it gets outdated. At least, Chinese users will be able to have the thought/light to try to check the English version if something gets wrong.

WDYT about we replace English for English (Official)? And then, have you any idea over a way of we make people beware of that as well? I hope that I could clarify my concern and I am open a better way the communicate and let them beware of that.

@DirectXMan12 @pwittrock @gabbifish any suggestion ? WDYT?

camilamacedo86:
/kind documentation
kind feature

DirectXMan12:
@camilamacedo86 I talked to some folks, I think we have official language docs reviewers for the Kubernetes project.  We should be able to have them do reviews/basic checks of the docs, and then merge.  I'll file a bug in the SIG docs repo and see about getting stuff reviewed.

@lhb008 can you open a PR with the book contents?  We might want to create a separate branch and subdomain for them

camilamacedo86:
HI @lhb008, 

I'd like to know if you would like to help us to achieve it? WDYT? Please, feel free to ping me if you need any help.

lhb008:
@camilamacedo86  Sorry to reply this so late. It is great for us community(China Cloud Native Community). I will open a PR for Chinese docs.

lhb008:
I have opened a PR https://github.com/kubernetes-sigs/kubebuilder/pull/1736.  Our community has already created a new branch `zh` to store the chinese docs. For the continuous update in the future, we can find a right way to do this, for example, if there is new feature for kubebuilder, you can tell our community, we will translate this into chinese, anyway, our community will focus on this always to help user in China to get the available chinese docs.

lhb008:
Hi  @camilamacedo86 . What do you think about to create a new branch to store chinese docs, then i can open the PR with the corresponding chinese docs branch?

[...]

PR #1888, opened 2020-12-09
:book: merge Chinese docs into the repo

majinghe:
#### Description

Merge Chinese docs(translated by China Cloud Native Community) into the repo, already discussed in the issue: https://github.com/kubernetes-sigs/kubebuilder/issues/1651.


Adirio:
@DirectXMan12 would you mind updating the `book-v2-zh` branch tp `book-v2`?

camilamacedo86:
Hi @majinghe, 

Really thank you for your collab. However, shows that it is missing a few nits to get merged as follows. 

1 - It shows not aligned with the master branch. We no longer use Travis for a while. So, I'd like to ask for you rebase this branch with master and do the required updates. 
2 - We need to check the doc changes in the preview. The netfly preview shows not working here as well. It might be solved with the master rebase. 

Hi @Adirio, 

After we get it merged we can check the branch name and etc. However, I do not think that we would need to have so far a v2 and V3 Chinese docs. We can move forward here by ensuring that we have the master docs in Chinese only for the latest version.

zhangguanzhang:
any update?

majinghe:
@camilamacedo86 @zhangguanzhang  Sorry to reply so late, will handle ASAP, really a little busy recently.

majinghe:
@camilamacedo86  @zhangguanzhang  sorry to reply so late.  After rebase to main branch, a lot of contents needs to be translated again(this will take some time, will lead to the zh branch fall behind main branch again), i think this is the key point that how to keep the synchronization between zh branch and main branch.  Do you have any good way to fix such asynchronization?

My suggestion is that  treat Our community repo : https://github.com/cloudnativeto/kubebuilder as the Chinese docs repo, publish url is https://cloudnative.to/kubebuilder/ . Our community will translate the main branch to chinese regularly(for example 3 times a year).

This is just my question, any good idea from your guys side?

camilamacedo86:
HI @majinghe, 

> ch. Do you have any good way to fix such synchronization?
Dealing with conflicts is not something nice, however, I usually use the merge tools for that which allow us to solve easier.

Adirio:
If they are going to be only a couple translations per year, I would suggest translating the docs of releases instead of master.

majinghe:
@camilamacedo86  I tried to fix the conflict and translate new content caused by rebae action. It really a lot of  workload, same to translate the whole content again. I think the same issue will be occur in the feature, it is really about the synchronization keeping for english master to  zh master. I agreed to the suggestion from @Adirio . Our Cloud Native  Community  translate the  kubebuilder to Chinese based on release instead of master.  How do you think about  this?

camilamacedo86:
Hi @majinghe, 

We are near to release 3.0.0 stable. Would you like that us re-create this branch for the Chinese docs based in the latest version you are able to work on?

My main probably to get this merged is that we are unable to indeed check the docs preview. So, I think that brokes something regarding the docs.

Adirio:
I think we need to define a bit better how do we want i18n to be handled in k9r. I added it to today's agenda as there seems not to be too many topics.

PR #1893, opened 2020-12-10
:bug: fix issue to use the tests outside of the makefile target

camilamacedo86:
**Description**
- add logic to set env var bins when the KUBEBUILDER_ASSETS is not set/used at alll which is the default behaviour for go/v3-alpha plugin. 

**Motivation**

https://github.com/kubernetes-sigs/kubebuilder/pull/1893


camilamacedo86:
Closing it a favour to https://github.com/kubernetes-sigs/controller-runtime/issues/1234 and https://github.com/kubernetes-sigs/controller-runtime/issues/1300 as discussed in the bug triage meeting.

PR #1895, opened 2020-12-10
:book: docs,cli: improve migration docs and CLI warnings

estroz:
- docs/book/src/migration: reorganize migration section so project and plugin version upgrades are in separate paths
- pkg/cli/cli.go: add upgrade directions if plugin resolution fails on an unstable plugin
- pkg/plugin/version.go: add the IsUnstable() helper


camilamacedo86 (approved):
/lgtm

PR #1899, opened 2020-12-11
:warning: change structure to store crdVersion and webhookversion (go/v3-alpha)

camilamacedo86:
**Description**
Only does the changes required to store the new attributes crdVersion and webhookVersion in the format/layout define on  https://github.com/kubernetes-sigs/kubebuilder/issues/1772 and  https://github.com/kubernetes-sigs/kubebuilder/issues/1826.

**Motivation**
- https://github.com/kubernetes-sigs/kubebuilder/issues/1772
- https://github.com/kubernetes-sigs/kubebuilder/issues/1826

### DIFF
--- a/pkg/model/config/config.go
+++ b/pkg/model/config/config.go
@@ -45,7 +45,7 @@ type Config struct {
 
 	// Resources tracks scaffolded resources in the project
 	// This info is tracked only in project with version 2
-	Resources []GVK `json:"resources,omitempty"`
+	Resources []ResourceData `json:"resources,omitempty"`
 
 	// Multigroup tracks if the project has more than one group
 	MultiGroup bool `json:"multigroup,omitempty"`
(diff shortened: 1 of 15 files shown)


estroz:
I think I'm ok with this. Going to wait for an ack from someone else.

/approve

Adirio (approved):
While the `Resource` and the `Config` models need some cleanup, this PR moves the previously existing fields to their correct location, which means that the rest of the changes that are to be done in the near future will not be breaking changes from a project configutaion file perspective, so
\lgtm

pkg/model/config/config.go
```diff
@@ -147,33 +149,71 @@ func (c Config) resourceAPIVersionCompatible(verType, version string) bool {
...
+
+// merge compares it with another webhook by setting each webhook type individually so existing values are
+// not overwritten.
+func (w *Webhooks) merge(other *Webhooks) {
+	if w.WebhookVersion == "" && other.WebhookVersion != "" {
+		w.WebhookVersion = other.WebhookVersion
+	}
+}
+
+// merge compares it with another webhook by setting each webhook type individually so existing values are
```
prafull11, line 212:
```suggestion
// merge compares it with another api by setting each api type individually so existing values are
```

prafull11 (commented):
Just a nit in the comment. Apart from that LGTM

pkg/model/config/config.go
prafull01, line 212:
```suggestion
// merge compares it with another api by setting each api type individually so existing values are
```

camilamacedo86: done

PR #1900, opened 2020-12-11
:seedling: pkg/cli: fix lack of help when --help is set

estroz:
#1828 removed [`--help` handling](https://github.com/kubernetes-sigs/kubebuilder/pull/1828/files#diff-f47a5804e09ec7e41dff0b216fbc32078ca027c60c7d9bb09192fbe2af83064aL56) so that `kubebuilder -h` will only print

```console
$ kubebuilder -h
Usage of base:
      --plugins string           plugins to run
      --project-version string   project version
pflag: help requested
```

/kind bug

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -95,6 +95,10 @@ type cli struct {
 	// A filtered set of plugins that should be used by command constructors.
 	resolvedPlugins []plugin.Plugin
 
+	// Whether some generic help should be printed, i.e. if the binary
+	// was invoked outside of a project with incorrect flags or -h|--help.
+	doHelp bool
+
 	// Root command.
 	cmd *cobra.Command
 }
@@ -162,7 +166,7 @@ func newCLI(opts ...Option) (*cli, error) {
 }
 
 // getInfoFromFlags obtains the project version and plugin keys from flags.
-func getInfoFromFlags() (string, []string) {
+func (c *cli) getInfoFromFlags() (string, []string) {
 	// Partially parse the command line arguments
 	fs := pflag.NewFlagSet("base", pflag.ExitOnError)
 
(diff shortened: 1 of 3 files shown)


pkg/cli/cli.go
```diff
@@ -171,11 +175,18 @@ func getInfoFromFlags() (string, []string) {
...
+	fs.BoolVarP(&help, "help", "h", false, "help flag")
 
 	// Parse the arguments
-	_ = fs.Parse(os.Args[1:])
+	err := fs.Parse(os.Args[1:])
+
+	// User needs *generic* help if args are incorrect or --help is set and
+	// --project-version is not set. Plugin-specific help is given if a
+	// plugin.Context is updated, which does not require this field.
+	c.doHelp = err != nil || help && !fs.Lookup(projectVersionFlag).Changed
```
Adirio, line 189: `err == nil` always due to how flags are parsed ( `plfag.ExitOnError` and `pflag.ParseErrorsWhitelist{UnknownFlags: true}`).

Adirio (commented):
@estroz  I thought that I was able to find a way to print the help message without this special case. Could you elaborate on the behavior that you consider broken?

(I know it is merged already)

pkg/cli/cli.go
estroz: While true, code should handle every error where possible in case library behavior changes.

estroz:
I think the problem was that somehow flag parsing was ending in `getInfoFromFlags()` when `--help` was set (see output in the description), and flags weren't being passed to the actual flagset in `cli`. There might be a bug somewhere else causing this, but at least now I've reverted `--help` behavior to its pre-#1828 state.

Adirio:
It says `Usage of base` and `base` is the name given to the temporal flagset used to pre-decode these two arguments. WIll investigate a bit further.

PR #1901, opened 2020-12-12
:warning: *: stabilize project version 3

estroz:
This PR stabilizes project version 3-alpha as 3.

P.S.: there are a few PRs open that change project structure significantly (#1869, #1899). We may want to forego those changes until project version 4 given the impending v3.0.0 release and stability concerns, or not. Whatever the case, this PR can block until the community feels ready to stabilize.

### DIFF
--- a/pkg/plugins/golang/v2/plugin.go
+++ b/pkg/plugins/golang/v2/plugin.go
@@ -25,7 +25,7 @@ import (
 const pluginName = "go" + plugins.DefaultNameQualifier
 
 var (
-	supportedProjectVersions = []string{config.Version2, config.Version3Alpha}
+	supportedProjectVersions = []string{config.Version2, config.Version3}
 	pluginVersion            = plugin.Version{Number: 2}
 )
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -29,13 +29,13 @@ func main() {
 	c, err := cli.New(
 		cli.WithCommandName("kubebuilder"),
 		cli.WithVersion(versionString()),
-		cli.WithDefaultProjectVersion(config.Version3Alpha),
+		cli.WithDefaultProjectVersion(config.Version3),
 		cli.WithPlugins(
 			&pluginv2.Plugin{},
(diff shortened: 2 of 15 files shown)


Adirio:
You mentioned that having `3-alpha` supported by `go/v2` would help transitioning from `go/v2` to `go/v3`, should we just leave `3-alpha` and just create a new `3` project version?

pkg/plugins/golang/v2/plugin.go
```diff
@@ -25,7 +25,7 @@ import (
 const pluginName = "go" + plugins.DefaultNameQualifier
 
 var (
-	supportedProjectVersions = []string{config.Version2, config.Version3Alpha}
+	supportedProjectVersions = []string{config.Version2, config.Version3}
```
camilamacedo86, line 28: Hi @estroz, 

I think we need still support v3-alpha. So, instead of remove `Version3Alpha ` we need to add `Version3` and keep the supportability for both. WDYT?

estroz:
>should we just leave 3-alpha and just create a new 3 project version?

This is kinda ugly. For how long would we support 3-alpha? How do we tell users to upgrade? I'd rather have a policy of getting rid of alpha/beta versions as soon as the next most stable version is available, and relying on [this error message](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/cli.go#L303-L304) to inform users of an upgrade.

@camilamacedo86 this relates to your [comment](https://github.com/kubernetes-sigs/kubebuilder/pull/1901#discussion_r559786191).

Adirio:
I feel like having go/v2 supporting config v3 will stop us from improving config v3. So I would either leave config 3-alpha as a transition step (at least for some time) or remove go/v2 support for config v3.

pkg/plugins/golang/v2/plugin.go
camilamacedo86: solved in another comment.

estroz:
Closing in favor of #1959

PR #1847, opened 2020-11-18
:bug: Fix --force option to recreate the files by kubebuilder create api

prafull01:
**Description**
Added changes to overwrite the files generated by kubebuilder create api when --force option is provided.

**Motivation**
--force command is failing on kubebuilder
Fixes: #1845

PR #1903, opened 2020-12-13
:sparkles: (go/v3) Add the --force option for the webhook

prafull01:
**Description**
Add the --force option to overwrite already generated webhook code.

**Motivation**
Currently, user do not have a way to overwrite already scaffold code other than deleting the scaffold code manually.

### DIFF
--- a/pkg/model/config/config.go
+++ b/pkg/model/config/config.go
@@ -117,6 +117,17 @@ func (c Config) HasGroup(group string) bool {
 	return false
 }
 
+// HasWebhook returns true if webhook is already present
+func (c Config) HasWebhook(resource ResourceData) bool {
+	for _, r := range c.Resources {
+		if r.isGVKEqualTo(resource) {
+			return r.Webhooks != nil
+		}
+	}
+
+	return false
+}
+
 // IsCRDVersionCompatible returns true if crdVersion can be added to the existing set of CRD versions.
 func (c Config) IsCRDVersionCompatible(crdVersion string) bool {
 	return c.resourceAPIVersionCompatible("crd", crdVersion)
--- a/testdata/project-v3/main.go
+++ b/testdata/project-v3/main.go
@@ -122,6 +122,10 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "Laker")
 		os.Exit(1)
 	}
+	if err = (&crewv1.Captain{}).SetupWebhookWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create webhook", "webhook", "Captain")
+		os.Exit(1)
+	}
 	// +kubebuilder:scaffold:builder
 
 	if err := mgr.AddHealthzCheck("health", healthz.Ping); err != nil {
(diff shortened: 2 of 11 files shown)


camilamacedo86 (approved):
/approve

It shows great for me 👍

testdata/project-v3/main.go
```diff
@@ -122,6 +122,10 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "Laker")
 		os.Exit(1)
 	}
+	if err = (&crewv1.Captain{}).SetupWebhookWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create webhook", "webhook", "Captain")
+		os.Exit(1)
+	}
```
camilamacedo86, line 128: /hold

We should not duplicate it since the file will be re-created. 
Could we fix it to get it merged?

prafull01: Hi @camilamacedo86 With current behavior we can not avoid duplicating that. We can not force overwrite the main.go as the functionality is to overwrite the entire file, not the section. Also, main.go is edited by other apis as well and hence we can not do that. 

To remove the such duplicate scaffolding we need to fix [this](https://github.com/kubernetes-sigs/kubebuilder/blob/0a807f4e9428f7544a132fc2aa5ce4baa3510119/pkg/plugins/internal/machinery/scaffold.go#L315)

camilamacedo86: would not possible if we use the --force option then not update the main.go?
Also, wdyt about we raise an issue in the repo to address this todo? It might give more visibility.

prafull01: Let me raise the issue and will try to work on it.
We can not skip the update on main.go because if user runs --force for the first time in webhook create command. It will not setup the webhook manager in the main.go.

camilamacedo86: Good catcher.  I am ok with we move forward with it and work on the duplication/todo in a follow-up.  
/hold cancel

pkg/model/config/config.go
```diff
@@ -119,6 +119,17 @@ func (c Config) HasGroup(group string) bool {
 	return false
 }
 
+// HasWebhook returns true if webhook is already present
+func (c Config) HasWebhook(kind string) bool {
+	for _, r := range c.Resources {
+		if r.Kind == kind {
+			return r.WebhookVersion != ""
```
camilamacedo86, line 126: We need to rebase it with masyter change it to `r.Webhook != nil && r.Webhook.WebhookVersion != ""`. More; https://github.com/kubernetes-sigs/kubebuilder/pull/1899

pkg/model/config/config.go
```diff
@@ -119,6 +119,17 @@ func (c Config) HasGroup(group string) bool {
 	return false
 }
 
+// HasWebhook returns true if webhook is already present
+func (c Config) HasWebhook(kind string) bool {
+	for _, r := range c.Resources {
+		if r.Kind == kind {
```
camilamacedo86, line 125: we need to compare here gkv becuase we can have the same Kind for diff apis/gkv. 
/hold

camilamacedo86 (approved):
it shows fine for me. 👍 
shows the same scenario of https://github.com/kubernetes-sigs/kubebuilder/pull/1847 discussed in the bug triage meeting. 

/lgtm
/hold cancel

PR #1909, opened 2020-12-15
:seedling: Add the unit tests in pkg/model/config/config_test.go for HasWebhook 

prafull01:
**Description**
Add the HasWebhook() unit tests to improve the coverage.

**Motivation**
Follow up of the PR #1903 where the HasWebhook() function is introduced.

### DIFF
--- a/pkg/model/config/config_test.go
+++ b/pkg/model/config/config_test.go
@@ -249,14 +249,15 @@ var _ = Describe("ResourceData Version Compatibility", func() {
 
 var _ = Describe("Config", func() {
 	var (
-		c          *Config
-		gvk1, gvk2 ResourceData
+		c                *Config
+		gvk1, gvk2, gvk3 ResourceData
 	)
 
 	BeforeEach(func() {
 		c = &Config{}
 		gvk1 = ResourceData{Group: "example", Version: "v1", Kind: "TestKind"}
 		gvk2 = ResourceData{Group: "example", Version: "v1", Kind: "TestKind2"}
+		gvk3 = ResourceData{Group: "example", Version: "v1", Kind: "TestKind", Webhooks: &Webhooks{WebhookVersion: v1beta1}}
 	})
 
 	Context("UpdateResource", func() {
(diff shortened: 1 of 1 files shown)


pkg/model/config/config_test.go
```diff
@@ -295,4 +296,17 @@ var _ = Describe("Config", func() {
...
 
 	})
+
+	Context("HasWebhook", func() {
+		It("should return true when config has a webhook with the GVK", func() {
+			c.UpdateResources(gvk3)
+			Expect(c.Resources).To(Equal([]ResourceData{gvk3}))
+			Expect(c.HasWebhook(gvk3)).To(BeTrue())
+		})
+		It("should return false when config has a webhook with the GVK", func() {
```
camilamacedo86, line 306:
```suggestion
		It("should return false when config has NOT a webhook with the GVK", func() {
```

prafull01: Done

pkg/model/config/config_test.go
```diff
@@ -295,4 +296,17 @@ var _ = Describe("Config", func() {
 		})
 
 	})
+
+	Context("HasWebhook", func() {
+		It("should return true when config has a webhook with the GVK", func() {
```
Adirio, line 301:
```suggestion
		It("should return true when config has a webhook for the GVK", func() {
```

pkg/model/config/config_test.go
```diff
@@ -295,4 +296,17 @@ var _ = Describe("Config", func() {
...
+			c.UpdateResources(gvk3)
+			Expect(c.Resources).To(Equal([]ResourceData{gvk3}))
+			Expect(c.HasWebhook(gvk3)).To(BeTrue())
+		})
+		It("should return false when config NOT has a webhook with the GVK", func() {
```
Adirio, line 306: Any of the following options:
```suggestion
		It("should return false when config does not have a webhook for the GVK", func() {
```
```suggestion
		It("should return false when config has no webhook for the GVK", func() {
```

Adirio (changes requested):
Some english related comments:

prafull01:
> Some english related comments:

@Adirio Fixed those.

camilamacedo86:
Hi @prafull01,

All fine here. Since we are only adding tests the correct mojo is: `Infra/Tests/Other: 🌱 (:seedling:)`.
I will update it for we are able to merge this one.

camilamacedo86 (approved):
/approve
/lgtm

PR #1911, opened 2020-12-17
⚠ Update models

Adirio:
# Description
- [x] `Stage` model:
    - [x] Extracted to a separate model package to be shared by plugin and project configuration versions
    - [x] Cove coverage: 100%
- [x] `Options`:
    - [x] Move `Options` to `pkg/plugins/golang`
    - [x] Code coverage: 100% 
- [x] `Resource` model:
    - [x] GVK submodel 
    - [x] Make compound fields (those that can be obtained from other fields) be methods instead of fields
    - [x] Add auxiliary methods to check if each part should be scaffolded
    - [x] Update method to allow `Config` files modify the tracked information about the resources.
    - [x] Code coverage: 100%
- [x] Project configuration:
    - [x] Split each version into each own package that share a common interface
    - [x] Code coverage: 97.11%


pkg/cli/options_test.go
```diff
@@ -40,7 +40,7 @@ var _ = Describe("CLI options", func() {
 
 		p   = newMockPlugin(pluginName, pluginVersion, projectVersion)
 		np1 = newMockPlugin("Plugin", pluginVersion, projectVersion)
-		np2 = mockPlugin{pluginName, plugin.Version{Number: -1, Stage: plugin.StableStage}, []string{projectVersion}}
```
camilamacedo86, line 43: why the plugin stage was removed?

camilamacedo86 (commented):
just one question. Otherwise, it shows fine. 
But, I think that would be nice to get @estroz's review on this one. 

^ ignore my above comment it receives changes after that.

pkg/cli/options_test.go
Adirio: Stable is the zero-value for stage so it is implicitly the same and that way we avoid an extra import

Adirio:
> just one question. Otherwise it shows fine.

It is still a draft, I plan to add more updates including `Resource` and `Config`, so that this can replace #1869, but having the changes better structured

pkg/model/resource/resource.go
```diff
@@ -25,49 +25,136 @@ import (
...
+
+// HasConversion returns true if the resource has an associated conversion webhook.
+func (r Resource) HasConversion() bool {
+	return r.Webhooks != nil && r.Webhooks.Conversion
+}
+
+// GVK returns the GVK identifier of a resource.
+func (r Resource) GVK() GVK {
+	return GVK{
+		Group:   r.QualifiedGroup(),
```
camilamacedo86, line 111:
```suggestion
		Group:   r.QualifiedGroup(),
```


I do not agree with this one at all. See: https://github.com/kubernetes-sigs/kubebuilder/pull/1869/#discussion_r541467615

/hold

pkg/model/resource/utils.go
```diff
@@ -0,0 +1,51 @@
...
+
+	// Remove dashes and dots
+	safe = strings.Replace(safe, "-", "", -1)
+	safe = strings.Replace(safe, ".", "", -1)
+
+	return safe
+}
+
+// LocalPath returns the default path
+func LocalPath(repo, group, version string, multiGroup bool) string {
```
camilamacedo86, line 38:
```suggestion
func DefaultPath(repo, group, version string, multiGroup bool) string {
```
`// LocalPath returns the default path` then, IMO it should call **DefaultPath**

pkg/model/resource/utils.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+		return path.Join(repo, "apis", version)
+	}
+	return path.Join(repo, "api", version)
+}
+
+// RegularPlural returns a default plural form when none was specified
+func RegularPlural(singular string) string {
+	return flect.Pluralize(strings.ToLower(singular))
+}
```
camilamacedo86, line 51: I am not sure how valuable is this func since it is only `flect.Pluralize(strings.ToLower(singular))` andd then, by used directly makes more clear what it is doing.

pkg/plugins/golang/options.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// Namespaced is true if the resource should be namespaced.
+	Namespaced bool
+
+	// Flags that define which parts should be scaffolded
+	DoAPI        bool
+	DoController bool
+	DoDefaulting bool
+	DoValidation bool
+	DoConversion bool
```
camilamacedo86, line 97: IMO these attrs brings confusion. They have not the name of the flags used in the ux and they are not the model ones. Then, you are here creating a third representation for values that came from the flags and are stored in the model. 

We should ONLY create new attrs/options for what we need to manipulate and we do not have a direct representation on the model. Otherwise, we bring a lot of unnecessary complexities.

pkg/plugins/golang/options.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+// GVK returns the GVK identifier of a resource.
+func (opts Options) GVK() resource.GVK {
+	return resource.GVK{
+		Group:   opts.QualifiedGroup(),
+		Version: opts.Version,
+		Kind:    opts.Kind,
+	}
+}
```
camilamacedo86, line 182: We have the resouce.GVK and options.GKV. Both are a duplicate code and re returning the same data. It is very confuse. 
IMO Options should use the resource objects instead of duplicate the code. 

And then, I do not agree at all with `Group:   opts.QualifiedGroup(),` as described in https://github.com/kubernetes-sigs/kubebuilder/pull/1911/files#r545354820

pkg/plugins/golang/v2/options.go
```diff
@@ -0,0 +1,283 @@
...
+		return resource.GVK{
+			Group:   opts.Group,
+			Version: opts.Version,
+			Kind:    opts.Kind,
+		}
+	}
+	return resource.GVK{
+		Group:   opts.QualifiedGroup(),
+		Version: opts.Version,
+		Kind:    opts.Kind,
```
camilamacedo86, line 190: The same code has in more 2 diff points : https://github.com/kubernetes-sigs/kubebuilder/pull/1911/files#r545369583

camilamacedo86:
What is the closer goal of this pr?

pkg/model/resource/resource.go
Adirio: I already proved that `Group` and `Domain` are two kubebuilder concepts that together form the `Group` as known by k/k.

/hold cancel

pkg/model/resource/utils.go
Adirio: Making it a function avoids code duplication, and I think I don't need to prove why we should always avoid code duplication.

pkg/plugins/golang/options.go
Adirio: These fields are not represented in the model anywhere. They are only represented in the webhook scaffolder. Previously they didn't have the `Do` prefixes, I added the prefixes to keep the same pattern as in the API scaffolder.

pkg/plugins/golang/options.go
Adirio: Both `Options` and `Resources` need to be able to return `GVK` items as they will be used to interact with the config file, and some of these interactions are done before the `Resource` is constructed (e.g., the api scaffolder validate function). The GVK wil be used as the unique key that identifies a resource, for example for functions of the config file like `HasResource` or `GetResource` which will receive a `GVK` as an argument and will return a `bool` or a `Resource` respectively.

pkg/plugins/golang/v2/options.go
Adirio: Already explained in https://github.com/kubernetes-sigs/kubebuilder/pull/1911#discussion_r545401875 why both `Options` and `Resource` need to be able to return `GVK` objects.

Adirio:
> What is the closer goal of this pr?

This is still a draft, some additional changes are going to be done as explained in the first message. For exmple `GVK` will be part of the interface with the `Config` file.

pkg/config/v3alpha/config.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+	}
+
+	return resource.Resource{}, config.UnknownResource{GVK: gvk}
+}
+
+// AddResource implements config.Config
+func (c *cfg) AddResource(res resource.Resource) error {
+	// TODO: remove when these fields need to be tracked
+	res.Domain = ""
```
camilamacedo86, line 183: IMO we should NOT store the Domain or have its attr in the resource. 
The domain is == for the full config project and is NOT specific for a resource. See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/PROJECT#L1

Adirio: I completely disagree. In k8s, a resource is uniquely identified by its GVK (Group, Version and Kind) where the group has a domain. See: https://github.com/kubernetes/kubernetes/blob/bb376f161640529f600203faee4cd0de8ec16778/pkg/apis/admission/register.go#L25

To simplify the process for the user, in kubebuilder we have decided to allow a project-wide domain to be configured for every group, so that the user can write `--group group` instead of `--group group.domain.com` every time he wants to create an API or a webhook. To keep the UX in sync with the `Resource` model, `Group` has been kept to refer to the value of the `--group` flag, and `QualifiedGroup` to refer to the group as understood by k8s (i.e., including the domain). This proves that the domain is part of the resource definition.

Additionally, we are currently supporting core resources which do not use the project-wide domain. See https://github.com/kubernetes-sigs/kubebuilder/blob/26d7b42314dc8df4369df3594329c3d2401793d2/testdata/project-v3-multigroup/controllers/crew/captain_controller.go#L37 and https://github.com/kubernetes-sigs/kubebuilder/blob/26d7b42314dc8df4369df3594329c3d2401793d2/testdata/project-v3-multigroup/controllers/apps/pod_controller.go#L35 to see how the domain is needed per resource. This proves the need of  of this field in order to scaffold the files. (In the future we also plan for additional support for external resources which will also have different domains.)

P.S.: As of this PR, the config file hasn't been modified to include this field.

Adirio:
I still have to add a couple tests to increase `pkg/model/resource` coverage (currently at 52.78 according to coveralls) but it is ready for review.

Adirio:
Finished with the tests. `pkg/model/stage/` and `pkg/model/resource/`, `pkg/plugins/golang/options.go` and `pkg/plugins/golang/v2/options.go` have full coverage. `pkg/config` has 97% coverage, the uncovered lines are marshalling error handlers that can never happen so can't be tested.

pkg/plugin/interfaces.go
```diff
@@ -56,7 +56,7 @@ type Subcommand interface {
 	Run() error
 	// InjectConfig passes a config to a plugin. The plugin may modify the config.
 	// Initializing, loading, and saving the config is managed by the cli package.
-	InjectConfig(*config.Config)
+	InjectConfig(config.Config)
```
estroz, line 59: The original intent of passing a pointer to a config was to allow the plugin to modify the config before the CLI wrapper could make modifications/write the config back to disk, and more generally propagate plugin-specific changes to the caller. I'd prefer to keep that functionality.

Adirio: `config.Config` is an interface now

estroz: Whoops I did not notice that. Thanks for clarifying.

estroz (changes requested):
A few nits. Also, any completely new files should get a 2021 copyright date (you don't need to change those that were just moved around).

I really wish Go had generics so we wouldn't have to deal with a config interface, which will break if we need to change a method :face_with_head_bandage:. However an interface might be the best way to solve the current problem of sharing versioned data.

pkg/plugins/golang/v2/init.go
```diff
@@ -85,19 +92,20 @@ func (p *initSubcommand) BindFlags(fs *pflag.FlagSet) {
...
 
 	// project args
-	fs.StringVar(&p.config.Repo, "repo", "", "name to use for go module (e.g., github.com/user/repo), "+
+	fs.StringVar(&p.domain, "domain", "my.domain", "domain for groups")
+	fs.StringVar(&p.repo, "repo", "", "name to use for go module (e.g., github.com/user/repo), "+
 		"defaults to the go package of the current working directory.")
-	fs.StringVar(&p.config.Domain, "domain", "my.domain", "domain for groups")
-	if p.config.IsV3() {
-		fs.StringVar(&p.config.ProjectName, "project-name", "", "name of this project")
+	if p.config.GetVersion().Compare(cfgv3alpha.Version) >= 0 {
```
camilamacedo86, line 98: why we are adding this conditional?
Are not go/v2 storing and showing this option currently?

pkg/plugins/golang/v2/init.go
```diff
@@ -113,36 +121,46 @@ func (p *initSubcommand) Validate() error {
...
 		}
-		p.config.Repo = repoPath
+		p.repo = repoPath
 	}
 
 	return nil
 }
 
 func (p *initSubcommand) GetScaffolder() (cmdutil.Scaffolder, error) {
+	if err := p.config.SetDomain(p.domain); err != nil {
```
camilamacedo86, line 152: Why we need the SetDomain in the interface? Why not p.config.Domain = p.domain?

estroz:
Meta note: I'm starting to think we should do away with the "project version" concept and move to versioning `PROJECT` per plugin version. The information stored in a project config is closely tied to plugin data, which may drastically vary since a plugin could be anything.

pkg/plugins/golang/v2/init.go
estroz: `p.config` is now an interface type.

pkg/config/v3alpha/config.go
```diff
@@ -0,0 +1,352 @@
...
+}
+
+// Resources implements config.Config
+func (c cfg) ResourcesLength() int {
+	return len(c.Resources)
+}
+
+// HasResource implements config.Config
+func (c cfg) HasResource(gvk resource.GVK) bool {
+	gvk.Domain = "" // Version 3 alpha does not include domain per resource
```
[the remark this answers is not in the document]
Adirio: Version 3 will, but this PR doesn't change v3-alpha, I was planning to create the v3 config in a separate PR as this is already pretty big.

pkg/plugins/golang/v2/init.go
Adirio: No, as you can see in red above it was only being done for v3.

pkg/plugins/golang/v2/init.go
Adirio: Interfaces do not allow to define fields, only methods, so fields need to be abstracted with a setter and a getter.

Adirio:
> Meta note: I'm starting to think we should do away with the "project version" concept and move to versioning `PROJECT` per plugin version. The information stored in a project config is closely tied to plugin data, which may drastically vary since a plugin could be anything.

Bear in mind that right now we only have one plugin running per command but that is expected to change in phase 2, where multiple plugins will be chained. Also, remember that a init plugin doesn't need to match a create api, or edit plugins and therefore the PROJECT handled by plugins will be a hell to manage not only for developers but also for users.

pkg/cli/cli.go
```diff
@@ -191,90 +191,94 @@ func (c *cli) getInfoFromFlags() (string, []string, error) {
...
 	switch {
-	// If they are both blank, use the default
+	// If they are both empty, use the default
 	case len(flagPlugins) == 0 && len(cfgPlugins) == 0:
-		plugins = c.defaultPlugins[projectVersion]
-	// If they are equal doesn't matter which we choose
-	case equalStringSlice(flagPlugins, cfgPlugins):
-		plugins = flagPlugins
-	// If any is blank, choose the other
+		if defaults, hasDefaults := c.defaultPlugins[projectVersion]; hasDefaults {
```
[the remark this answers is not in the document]
Adirio: If there are no default plugins for a project version the plugin keys will be empty, resolving into a empty list of resolved plugins (obviously) and when binding the different plugins to their commands, they will not find any API plugin and so on so it will build a command that errors out.

I haven't changed this behavior, if we wish to change this I think we should discuss it in the EP in #1942

estroz:
>remember that a init plugin doesn't need to match a create api, or edit plugins and therefore the PROJECT handled by plugins will be a hell to manage not only for developers but also for users.

If we forced `init` to create some config then this could work. However I agree it may be more problematic than the current problem :grimacing:.

/approve

estroz:
@DirectXMan12 can you take a look at this? Also coveralls may need to be overridden (I think coverage change is a false positive in terms of functional coverage).

Adirio:
> P.S.: Coveralls computes the final score only considering those packages that have at least a `*_test.go` file. Package `pkg/plugins/go/v2` adds tests for its `Options` object, which includes the v2 plugin code in the coverage, resulting in a coverage false "decrease" according to coveralls. In reality, the coverage has increased. Master has 1364 lines covered ([link](https://coveralls.io/github/kubernetes-sigs/kubebuilder)) and this PR has 1863 ([link](https://coveralls.io/builds/36228881)).

This ⬆️ 

What we may want to do is place a suite test in every directory (that PR will get a horrible coveralls score) so that the following PRs do show the real result. (We probably want some exceptions, like template scaffolds which dont need to be tested in unit tests.)

pkg/config/interface.go
```diff
@@ -0,0 +1,112 @@
...
+	// ResourcesLength returns the number of tracked resources
+	ResourcesLength() int
+	// HasResource checks if the provided GVK is stored in the Config
+	HasResource(gvk resource.GVK) bool
+	// GetResource returns the stored resource matching the provided GVK
+	GetResource(gvk resource.GVK) (resource.Resource, error)
+	// AddResource adds the provided resource if it was not present, no-op if it was already present
+	AddResource(res resource.Resource) error
+	// UpdateResource adds the provided resource if it was not present, modifies it if it was already present
+	UpdateResource(res resource.Resource) error
```
camilamacedo86, line 86: It is missing getResources to return all resources from the config.

pkg/config/interface.go
```diff
@@ -0,0 +1,112 @@
...
+	// This method was introduced in project version 3-alpha.
+	SetComponentConfig() error
+	// ClearComponentConfig disables component config
+	// This method was introduced in project version 3-alpha.
+	ClearComponentConfig() error
+
+	/* Resources */
+
+	// ResourcesLength returns the number of tracked resources
+	ResourcesLength() int
```
camilamacedo86, line 78: I do not think that we should have ResourcesLength() in the interface. 
Instead of that, we should've GetResources or Resources which is more generic.
The Length is a go func

pkg/config/interface.go
```diff
@@ -0,0 +1,112 @@
...
+	// GetResource returns the stored resource matching the provided GVK
+	GetResource(gvk resource.GVK) (resource.Resource, error)
+	// AddResource adds the provided resource if it was not present, no-op if it was already present
+	AddResource(res resource.Resource) error
+	// UpdateResource adds the provided resource if it was not present, modifies it if it was already present
+	UpdateResource(res resource.Resource) error
+
+	// HasGroup checks if the provided group is the same as any of the tracked resources
+	HasGroup(group string) bool
+	// IsCRDVersionCompatible returns true if crdVersion can be added to the existing set of CRD versions.
```
camilamacedo86, line 90: I am unable to get the plugins of a config. 
/hold

pkg/config/interface.go
```diff
@@ -0,0 +1,112 @@
...
+	SetDomain(domain string) error
+
+	// GetRepository returns the project repository.
+	GetRepository() string
+	// SetRepository sets the project repository
+	SetRepository(repository string) error
+
+	// GetName returns the project name
+	// This method was introduced in project version 3-alpha.
+	GetName() string
```
camilamacedo86, line 44:
```suggestion
	GetProjectName() string
```


See config.getName shows the name of the config when config.GetProjectName makes clear whet exactly is it.

pkg/model/resource/resource.go
```diff
@@ -19,55 +19,136 @@ package resource
...
+
+// HasValidation returns true if the resource has an associated validation webhook.
+func (r Resource) HasValidation() bool {
+	return r.Webhooks != nil && r.Webhooks.Validation
+}
+
+// HasConversion returns true if the resource has an associated conversion webhook.
+func (r Resource) HasConversion() bool {
+	return r.Webhooks != nil && r.Webhooks.Conversion
+}
```
camilamacedo86, line 104: This Has should be part of the webhook struct and not resource.

pkg/plugins/golang/v2/api.go
```diff
@@ -126,29 +128,29 @@ func (p *createAPISubcommand) Run() error {
 }
 
 func (p *createAPISubcommand) Validate() error {
-	if err := p.resource.ValidateV2(); err != nil {
+	if err := p.options.Validate(); err != nil {
```
camilamacedo86, line 131: Note that we have validation for v2 and another for v3. Then, are we still respecting it?

pkg/plugins/golang/v2/api.go
```diff
@@ -126,29 +128,29 @@ func (p *createAPISubcommand) Run() error {
...
 	}
 	if !p.controllerFlag.Changed {
 		fmt.Println("Create Controller [y/n]")
-		p.doController = util.YesNo(reader)
+		p.options.DoController = util.YesNo(reader)
 	}
 
 	// In case we want to scaffold a resource API we need to do some checks
-	if p.doResource {
+	if p.options.DoAPI {
```
camilamacedo86, line 146: Why we need the options at all? Why not we can have  p.resource as before? 
See that you will get all options to create the resource at the same way.

pkg/plugins/golang/v2/api.go
```diff
@@ -175,9 +177,9 @@ func (p *createAPISubcommand) GetScaffolder() (cmdutil.Scaffolder, error) {
 		return nil, fmt.Errorf("unknown pattern %q", p.pattern)
 	}
 
-	// Create the actual resource from the resource options
-	res := p.resource.NewResource(p.config, p.doResource)
-	return scaffolds.NewAPIScaffolder(p.config, string(bp), res, p.doResource, p.doController, p.force, plugins), nil
+	// Create the resource from the options
+	res := p.options.NewResource(p.config)
```
camilamacedo86, line 181: we need to be able to create a new resource from the resource. 
The Options are not valid for all kind of plugins.

pkg/plugins/golang/options.go
```diff
@@ -0,0 +1,241 @@
...
+	return resource.GVK{
+		Group:   opts.Group,
+		Domain:  opts.Domain,
+		Version: opts.Version,
+		Kind:    opts.Kind,
+	}
+}
+
+// NewResource creates a new resource from the options
+func (opts Options) NewResource(c newconfig.Config) resource.Resource {
```
camilamacedo86, line 186: The only way to create a resource now and validate it is by using the Options. 
However, it is specific for Golang and we need these actions for any kind of plugin. (e.g ansible/helm will create resources and then required to validate them according to the kb rules)

In this way, could you add a new resource for the Resource struct and centralize its validation there. The options can use it, however, who is better to know if is valid or not than itself? In this way, IMO the Validate of a resource needs still in/with the resource.

pkg/plugins/golang/v2/api.go
```diff
@@ -175,9 +177,9 @@ func (p *createAPISubcommand) GetScaffolder() (cmdutil.Scaffolder, error) {
...
+	return scaffolds.NewAPIScaffolder(p.config, string(bp), res, p.force, plugins), nil
```
camilamacedo86, line 182: See my other comment. Options are not valid for all kinds. 
We need the : 

	res := p.resource.NewResource(p.config, p.doResource)

pkg/config/interface.go
Adirio: I didn't expose that on purpose, what is the usage of this function? Plugins info is only stored if plugins decide to do so, so using this function will not really return if a plugins is being used, only if that plugin stored previously a configuration object.

pkg/config/interface.go
Adirio: As slices and underlying arrays are mutable, we can't just return the field. We will need to (deep) copy the whole array. This also includes making a copy of anything that is a pointer. This is a very expensive operation just to compute the length of a slice, that is why `ResourcesLength` is being provided directly. So even if we do includea `GetResources` method, `len(cfg.GetResources())` will be a very bad approach.

pkg/config/interface.go
Adirio: As slices and underlying arrays are mutable, we can't just return the field. We will need to (deep) copy the whole array. This also includes making a copy of anything that is a pointer. This is a very expensive operation.

Which is the usage of `GetResources`?

pkg/config/interface.go
Adirio: `Config` is the project configuration, so I think that the "project" part is implicit.

Let's have a third opinion, can you PTAL @estroz

pkg/model/resource/resource.go
Adirio: You can't check if `Resource.Webhooks` is `nil` from inside `Webhooks`, it would just error out with a nil pointer dereference error.

pkg/plugins/golang/options.go
Adirio: That is intentional. `Resource` is exported, so any plugin/third-party library can create them. How we transform our flags into our resource is part of the go plugin implementation. Additionally, if you are doing a light wrap around the go plugin as SDK does, you can also use `Options` as it is exported to.

pkg/model/resource/resource.go
camilamacedo86: The solution propsoed here is: 

- resource.HasWebhooks - returns true when the resource has webhooks
- resoirce.webhooks.HasValidation - returns true when has the validation type

> resource.HasWebhooks && resoirce.webhooks.HasValidation

OR at least

- resource.HasValidationWebhook 

See that for example `resource. HasValidation` is not good because HasValidation is part of Webhooks and not resources and because as a developer I need to check the func implementation to know that is checking if it has a webhook with the Validation type. The meaning of HasValidation is not implicity at all.

pkg/plugins/golang/options.go
camilamacedo86: Did you change that? How can I create a newResource and Validate its Resource?

pkg/config/interface.go
camilamacedo86: Hi @Adirio, 

We need to be able to GET all resources from a config. If this layout does not allow it then we cannot move forward with this proposal. 

Otherwise, we cannot move forward with this proposal. The KB is used and extends for we are to create other plugins and customize them. in this way, has many scenarios that we need to GET all resources from the config.

pkg/model/resource/resource.go
Adirio: I'm down to change them to `HasXxxxxxWebhook`

pkg/config/interface.go
camilamacedo86: The motivation of the plugin implementation is to allow us to extend KB. 
I need to know all plugins of some config in many scenarios.  I can create tools based on the plugins used in the config, I can do customization and etc. 

In the above implementation, we need to check that to ensure that projects which were built before the plugins introduction will still working on as before.

pkg/plugins/golang/options.go
Adirio: The different `Options` struct are fit to a certain plugin flags, so it is not part of the resource model.

A global valid-for-all `Options` field is not a good design decission as it will be really hard to mantain. We would have to add additional fields as required by external plugins, and this changes will block the external plugins. We would have to keep conditionals for different versions and plugins all over the struct methods, which was one of the main reasons we moved to a `Config` interface with independent implementations for each version.

pkg/config/interface.go
estroz: `GetProjectName()` is fine with me.

pkg/plugins/golang/v2/init.go
```diff
@@ -113,36 +121,46 @@ func (p *initSubcommand) Validate() error {
...
+		return nil, err
+	}
+	if err := p.config.SetRepository(p.repo); err != nil {
+		return nil, err
+	}
+	if p.config.GetVersion().Compare(cfgv3alpha.Version) >= 0 {
+		if err := p.config.SetName(p.name); err != nil {
```
camilamacedo86, line 159: The Get was changed for GetProjectName. We need to change the set as well.

Adirio: I missed that method. Now its done.

pkg/config/v3alpha/config.go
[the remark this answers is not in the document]
camilamacedo86: We need to fix it to get this PR merged. Ansible/Helm plugins in SDK are failing with this approach because their boilerplates uses `.Resource.Domain` which is returning nil.

pkg/config/v3alpha/config.go
```diff
@@ -0,0 +1,349 @@
...
+	// Boolean fields
+	MultiGroup      bool `json:"multigroup,omitempty"`
+	ComponentConfig bool `json:"componentConfig,omitempty"`
+
+	// Resources
+	Resources []resource.Resource `json:"resources,omitempty"`
+
+	// Plugins
+	Plugins PluginConfigs `json:"plugins,omitempty"`
+}
```
camilamacedo86, line 52: @Adrio so far Kubebuilder exports the plugins field which allows SDK to do customizations on top. See: https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator/PROJECT#L12-L14

We need this in the interface otherwise it is a breaking change in the KB API for its consumer that does not allow the customizations be done.

pkg/config/v3alpha/config.go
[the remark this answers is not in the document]
Adirio: > We need to fix it to get this PR merged. Ansible/Helm plugins in SDK are failing with this approach because their boilerplates use `.Resource.Domain` which is returning nil.

This doesn't have to do with the config file in any sense. SDK is not setting the `Domain` field in the `Options` struct, and therefore the created `Resource` has no `Domain` set.

pkg/config/v3alpha/config.go
Adirio: The `Plugin` field is exported because all the fields need to be exported in order to be marshalled to JSON/YAML. `Plugin` fields clearly specify how it should be used in its comments, stating that its goal is to store plugin configurations and that they should be encoded with the two provided methods. The methods documentation also provide more info on their usage. This exact same interface is implemented here.

SDK is using this field in a different way, storing a key of certain functionalities that were applid on the scaffolded files. These are not plugins as understood by kubebuilder nor are storing any configuration in this field. You can't say this is a breaking change when exactly the same interface as before is being provided just because a third-party was using it in a "hackish" way. The "hackish" way that is being used by SDK can still be used with th two provided methods, by trying to decode a certain plugin key. If the plugin key was not found it will yield an error while if found it will succeed.

camilamacedo86: I could check. Thank you.

camilamacedo86 (commented):
Hi @Adirio, 

Great work 🥇 
I could make Ansible/Helm plugins in SDK work with the changes. 
For the golang I still needing to figure out some nits but all shows fine. 

/lgtm

PR #1852, opened 2020-11-19
 ⚠️ (go/v3-alpha) Align flag names with Kubernetes Components

christopherhein:
This aligns the current `v3` flags with other Kubernetes core components for leader election and for metrics address.

Fixes #1839

PR #1912, opened 2020-12-17
:bug: revert the change on the flag and came back it for enable-leader-election

camilamacedo86:
**Description**
The manager provides the flag `enable-leader-election` which was changed in v3/ for `--leader-elect`. This PR reverts this change. 

**Motivation**
- The nomenclature `enable-leader-election` shows more aligned with its nomenclature in controller-runtime 
- Change the flag name to `--leader-elect` shows not make easier its understatement to users and only will bring more work necessary to migrate their project to v3


camilamacedo86:
Ignore it has a good argumentation: https://github.com/kubernetes-sigs/kubebuilder/pull/1852

PR #1919, opened 2020-12-18
:book: update the docs for v3 as default plugin version for the init

camilamacedo86:
**Description**
Now it has no longer v3-alpha plugin and the init by default will use this version. So, we need to have the docs updated accordingly.


Adirio (approved):
/lgtm

PR #1936, opened 2021-01-12
🌱 Enhance Subcommand API

Adirio:
# Description
`Subcommand.UpdateContext` method was being used for both inserting cli related metadata to the subcommands and returning subcommand related metadata to the command constructor (cli). The `Context` struct didn't have anything to do with Go's `context` pacvkage which could also lead to harder to read code.

This PR uses the term `Metadata` instead of `Context` (which also has the advantage of not conflicting with `onsi/ginkgo`) and makes clearer differentioation between the input and output metadatas (cli->subcommand and subcommand->cli).

### DIFF
--- a/pkg/cli/api.go
+++ b/pkg/cli/api.go
@@ -26,32 +26,23 @@ import (
 )
 
 func (c cli) newCreateAPICmd() *cobra.Command {
-	ctx := c.newAPIContext()
 	cmd := &cobra.Command{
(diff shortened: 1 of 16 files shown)


Adirio:
#1980 implements this with additional cleanups, please consider merging the other PR first (which will close this one) unless any of the other changes there are blocking and we want to merge this first.

Issue #1915, opened 2020-12-17
Invalid plugins names should return the help with the options available


camilamacedo86:
the current behaviour is: 

```
 $ kubebuilder init --plugins -h
2020/12/17 12:14:53 invalid plugin name "-h": invalid plugin name "-h": [a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z0-9](?:[-a-z0-9]*[a-z0-9])?(\.[a-z0-9](?:[-a-z0-9]*[a-z0-9])?)*')]
```

```
$ kubebuilder init --plugins=go -h
2020/12/17 12:14:58 ambiguous plugin "go" for project version "3-alpha"
```

When I would expect to see always the options available `Available plugins: ("go.kubebuilder.io/v2", "go.kubebuilder.io/v3")` instead such as we have for projects:

```
    $ kubebuilder init --plugins=go/v2 -h
Initialize a new project.

For further help about a specific project version, set --project-version.

Usage:
  kubebuilder init [flags]

Examples:
  # Help for initializing a project with version "2"
  kubebuilder init --project-version="2" -h

  # Help for initializing a project with version "3-alpha"
  kubebuilder init --project-version="3-alpha" -h

Flags:
  -h, --help                     help for init
      --plugins strings          Name and optionally version of the plugin to initialize the project with. Available plugins: ("go.kubebuilder.io/v2", "go.kubebuilder.io/v3")
      --project-version string   project version, possible values: ("2", "3-alpha") (default "3-alpha")
```

adikul30:
Hi! Should v3 issues have a label of their own? Since the --plugins flag is not available in the _init_ command for v2.

camilamacedo86:
probably the PR https://github.com/kubernetes-sigs/kubebuilder/pull/1937 will solve this one too.

Adirio:
With #1937, the output to `kubebuilder init --plugins -h` is:

```
Error: invalid plugin name "-h": invalid plugin name "-h": [a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z0-9](?:[-a-z0-9]*[a-z0-9])?(\.[a-z0-9](?:[-a-z0-9]*[a-z0-9])?)*')]
[...]

Issue #1914, opened 2020-12-17
Error `invalid value for project version`  appears without let us know what values are available


camilamacedo86:
Unable to use the help to check the plugins versions available for an project version:

```
 $ kubebuilder init --project-version=v3-alpha --plugins=h
2020/12/17 12:00:07 invalid value for project version (regex used for validation is '[1-9][0-9]*(?:-(?:alpha|beta))?')
``` 

When I would expect to see: 

```
$ kubebuilder init --project-version=3
Error: no resolved plugins, please specify plugins with --project-version or/and --plugins flags
Usage:
  kubebuilder init [flags]

Examples:
  # Help for initializing a project with version "2"
  kubebuilder init --project-version="2" -h

  # Help for initializing a project with version "3-alpha"
  kubebuilder init --project-version="3-alpha" -h

Flags:
  -h, --help                     help for init
      --plugins strings          Name and optionally version of the plugin to initialize the project with. Available plugins: ("go.kubebuilder.io/v2", "go.kubebuilder.io/v3")
      --project-version string   project version, possible values: ("2", "3-alpha") (default "3-alpha")

2020/12/17 12:12:19 no resolved plugins, please specify plugins with --project-version or/and --plugins flags
```

PS.: It shows a bug introduced in the latest changes. 
c/c @estroz @Adirio

Adirio:
Project versions do not support the `v` prefix:
```
$ kubebuilder init --project-version=3-alpha --plugins=h
```

estroz:
Yeah this is a validation error. I don't see what better error kubebuilder could produce if the project version doesn't have the correct format.

/remove-kind bug

Adirio:
Closing this as it should be solved just by removing the invalid `v` character. Reopen if the error is still present afterwards.

camilamacedo86:
Hi @Adirio @estroz, 

>

estroz:
Outputting command help is fine with me.

camilamacedo86:
It is a bug because what is expected for any CLI/tool command is: 
- If an invalid arg or value be used to run the command then, the tool should return error with the help
- The help should provide the possible options available

Adirio:
[...]

Issue #1913, opened 2020-12-17
init --help should shows the plugins and project versions available 


camilamacedo86:
**What do you want to happen?** 

I'd like to see the project versions and plugins when I the command `kubebuilder init --help` is executed. Otherwise, it is hard to know the values. 

```
 $ kubebuilder init --help
Usage of base:
      --plugins string           plugins to run
      --project-version string   project version
pflag: help requested

```

camilamacedo86:
the PR https://github.com/kubernetes-sigs/kubebuilder/pull/1937 should fix that as well.

Adirio:
With #1937, the output to `kubebuilder init --help` is:

```
Initialize a new project including vendor/ directory and Go package directories.

Writes the following files:
- a boilerplate license file
- a PROJECT file with the domain and repo
- a Makefile to build the project
- a go.mod with project dependencies
- a Kustomization.yaml for customizating manifests
- a Patch file for customizing image for manager manifests
- a Patch file for enabling prometheus metrics
- a main.go to run

project will prompt the user to run 'dep ensure' after writing the project files.

Usage:
  kubebuilder init [flags]

Examples:
  # Scaffold a project using the apache2 license with "The Kubernetes authors" as owners
  kubebuilder init --project-version=2 --domain example.org --license apache2 --owner "The Kubernetes authors"


Flags:
      --component-config         create a versioned ComponentConfig file, may be 'true' or 'false'
      --domain string            domain for groups (default "my.domain")
      --fetch-deps               ensure dependencies are downloaded (default true)
  -h, --help                     help for init
      --license string           license to use to boilerplate, may be one of 'apache2', 'none' (default "apache2")
      --owner string             owner to add to the copyright
      --plugins strings          Name and optionally version of the plugin to initialize the project with. Available plugins: ("go.kubebuilder.io/v2", "go.kubebuilder.io/v3")
      --project-name string      name of this project
      --project-version string   project version, possible values: ("2", "3-alpha") (default "3-alpha")
      --repo string              name to use for go module (e.g., github.com/user/repo), defaults to the go package of the current working directory.
      --skip-go-version-check    if specified, skip checking the Go version
```

Should this be enough? Any suggested changes?

camilamacedo86:
Yes, it is. get my attention the statement `project will prompt the user to run 'dep ensure' after writing the project files.`. I do not think that we should have that any more. however, it is another bug.

Adirio:
[...]

PR #1937, opened 2021-01-12
✨ Build the command and use it to report user errors

Adirio:
Improve error reporting by printing the command usage
Fixes: #1913
Fixes: #1914
Fixes: #1915

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -18,7 +18,6 @@ package cli
 
 import (
 	"fmt"
-	"log"
 	"os"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


Adirio:
@estroz I removed the `doHelp` field again that you restored in #1900 but I think this PR behavior is correct now. Basically the `"base"` flagset was printing its usage message in case we had the `--help` or `-h` flags as that is hardcoded before the whitelist filter, so by adding a dummy `--help` and `-h` flag to this base FlagSet we prevent this behavior and we delegate into the cobra command to print its usage.

camilamacedo86 (approved):
/approve 

Great work. it shows terrific for me 👍

estroz:
@Adirio this behavior does look a lot better.

/lgtm

PR #1939, opened 2021-01-13
Build kubebuilder-tools v1.20.2

shysank:
Updates kubebuilder tools to version ~`1.20.1`~ `1.20.2`

### DIFF
--- a/build/cloudbuild_tools.yaml
+++ b/build/cloudbuild_tools.yaml
@@ -14,20 +14,20 @@
 
 steps:
 - name: "gcr.io/cloud-builders/docker"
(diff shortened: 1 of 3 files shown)


camilamacedo86:
Hi @shysank, 

Could you please check with @estroz? He works on the code that does the release a few time ago. I am not sure if this branch is used since it is diff from the master.

shysank:
@camilamacedo86 Confirmed with @estroz that this is the right branch. Also, based on @estroz's suggestion, updated to `1.20.2` instead of `1.20.1` which got released recently.

PR #1942, opened 2021-01-13
🌱 Plugin resolution enhancement proposal

Adirio:
Enhance the plugin resolution algorithm to simplify the user experience and allow plugins to be defined by command call.

Implementation: #2019

### DIFF
--- a/designs/plugin-resolution.md
+++ b/designs/plugin-resolution.md
@@ -1,0 +1,88 @@
+# Plugin and project version resolution
+
+## Old behavior
+
+The same process is followed for all commands, without distinction between `init` and the rest. The only difference is
+that, for `init`, the configuration file will not be present.
+
+1. Obtain the project version from project configuration file's `version` field, `--project-verison` flag, or default
+   project version configured during CLI creation (`cli.New(cli.WithDefaultProjectVersion(...))`).
+    * In case both a project configuration file and the flag are found, fail if they are different.
+2. Obtain the plugin keys from the project configuration file's `layout` field, `--plugins` flag, or default plugins
+   configured during CLI creation for the above project version (`cli.New(cli.WithDefaultPlugins(...))`).
+    * In case both a project configuration file and the flag are found, fail if they are different.
+3. In case any of the plugin keys is not fully qualified (full name and version), check if there is only one plugin that
+   fits that unqualified plugin key.
+4. Resolve the plugins with the list of qualified keys above.
+5. Verify the plugins support the project version.
+   
+
+### Flaws
+
+* Default plugins need to be provided per project version
+* Project version can't be resolved (e.g., when a single project version is supported by the provided plugins).
+* Command calls are not allowed to override plugin keys.
+
+## Proposal
+
+The following plugin resolution algorithm solves the flaws described above:
+
+1. [All commands but Init] Obtain the project version and plugin keys from the project configuration if available.
+     * If they are available, plugin keys will be fully qualified.
+2. [Init command only] Obtain the project version from `--project-version` flag.
+     * The flag is optional, we will try to resolve project version later in case it is not provided.
+3. Obtain the optional plugin keys from `--plugins` flag.
+     * [Init command only] If not present, use default plugins.
+     * [All commands but Init] If not present, use plugin keys from configuration file (step 1).
+4. Qualify the unqualified keys from step 3.
+     * If we know the project version, use this info to filter the available plugins so that we can be more accurate.
+5. Resolve the plugins with the list of qualified keys above.
+6. [Init command only] Resolve the project version in case it wasn't provided and there is only one project version
+   supported by all the resolved plugins.
+   
+### Example
+
+Based on the following main.go file:
+```go
+package main
+
+import (
+   "log"
+
+   "sigs.k8s.io/kubebuilder/v2/pkg/cli"
+   "sigs.k8s.io/kubebuilder/v2/pkg/model/config"
+   pluginv2 "sigs.k8s.io/kubebuilder/v2/pkg/plugins/golang/v2"
+   pluginv3 "sigs.k8s.io/kubebuilder/v2/pkg/plugins/golang/v3"
+)
+
+func main() {
+   c, err := cli.New(
+      cli.WithCommandName("kubebuilder"),
+      cli.WithVersion(versionString()),
+      cli.WithPlugins(
+         &pluginv2.Plugin{},
+         &pluginv3.Plugin{},
+      ),
+      cli.WithDefaultPlugins(&pluginv3.Plugin{}),
+      cli.WithCompletion,
+   )
+   if err != nil {
+      log.Fatal(err)
+   }
+   if err := c.Run(); err != nil {
+      log.Fatal(err)
+   }
+}
+```
+
+As the `go.kubebuilder.io/v3` only supports project version "3", the project version can be resolved and doesn't
+need to be specified in the flags.
+```
+kubebuilder init --plugins=go.kubebuilder.io/v3
+```
+
+```
+kuebebuilder create api --plugins=declarative.kubebuilder.io/v1 ...
+```
+For this command, the declarative plugin will be used instead of the base plugin,
+but following command calls won't use it unless provided again.


designs/plugin-resolution.md
```diff
@@ -0,0 +1,87 @@
...
+* Project version can't be resolved (e.g., when a single project version is supported by the provided plugins).
+* Command calls are not allowed to override plugin keys.
+
+## Proposal
+
+The following plugin resolution algorithm solves the flaws described above:
+
+1. [All commands but Init] Obtain the project version and plugin keys from the project configuration if available.
+     * If they are available, plugin keys will be fully qualified.
+2. [Init command only] Obtain the project version from `--project-version` flag.
```
Adirio, line 32: One of the advantages implementation-wise of this new approach is that `--project-version` stops being a global variable for all commands that need to be special cased. It will just be a local variable of the `Init` command.

designs/plugin-resolution.md
```diff
@@ -0,0 +1,88 @@
...
+
+The same process is followed for all commands, without distinction between `init` and the rest. The only difference is
+that, for `init`, the configuration file will not be present.
+
+1. Obtain the project version from project configuration file's `version` field, `--project-verison` flag, or default
+   project version configured during CLI creation (`cli.New(cli.WithDefaultProjectVersion(...))`).
+    * In case both a project configuration file and the flag are found, fail if they are different.
+2. Obtain the plugin keys from the project configuration file's `layout` field, `--plugins` flag, or default plugins
+   configured during CLI creation for the above project version (`cli.New(cli.WithDefaultPlugins(...))`).
+    * In case both a project configuration file and the flag are found, fail if they are different.
```
camilamacedo86, line 13: So, are you saying that if we use a plugin to scaffold the project and that was not found for the CLI it should fail? 

if I scaffold the project with SDK and I used the layout plugins:go/3,pluginA,pluginB, for example, IMO I should still able to use KB. The projects are both compatible. So, could we not only executed the plugins found and raise a warning to let the user knows that in this example the pluginA and B were not found.

designs/plugin-resolution.md
```diff
@@ -0,0 +1,88 @@
...
+   }
+}
+```
+
+As the `go.kubebuilder.io/v3` only supports project version "3", the project version can be resolved and doesn't
+need to be specified in the flags.
+```
+kubebuilder init --plugins=go.kubebuilder.io/v3
+```
+
```
camilamacedo86, line 83: What will be the behaviour if  `go.kubebuilder.io/v3` be supported by `project-version=3 and 4`?
It will mean that users will always need to inform the project-version? 

Imo, in this case, the tool should be able to always use the latest project-version supported by the plugin instead of force the user inform it. 



Then, it means that ifwould be mandatory to inform the project-version. Is that? Could we not in this case define a default version that should be used?

designs/plugin-resolution.md
Adirio: This is a description of the current behavior, not the proposed enhancement.

And what I'm saying is that, if right now, you call `kubebuilder init --plugins go/v3 ...` and then you run `kubebuilder create api --plugins X ...` where X is anything different than `go/v3`, it fails. This is how it works right now, the proposal suggests not to do this.

designs/plugin-resolution.md
Adirio: If `go/v3` supports project versions 3 and 4, and I run `kubebuilder init --plugins go/v3 ...` it will try to use the default project version defined for the CLI (currently project version 3, but it may be different at the time that `go/v3` supports project version 4).

camilamacedo86: it is fine. So, it addresses my concern. Thank you.

designs/plugin-resolution.md
camilamacedo86: Thank you for the clarification.

designs/plugin-resolution.md
```diff
@@ -0,0 +1,88 @@
...
+     * The flag is optional, we will try to resolve project version later in case it is not provided.
+3. Obtain the optional plugin keys from `--plugins` flag.
+     * [Init command only] If not present, use default plugins.
+     * [All commands but Init] If not present, use plugin keys from configuration file (step 1).
+4. Qualify the unqualified keys from step 3.
+     * If we know the project version, use this info to filter the available plugins so that we can be more accurate.
+5. Resolve the plugins with the list of qualified keys above.
+6. [Init command only] Resolve the project version in case it wasn't provided and there is only one project version
+   supported by all the resolved plugins.
+
```
camilamacedo86, line 42: Should we not cover here the subcommand plugins scenarios when the layout ( init ) has more than one plugin in this proposal?

designs/plugin-resolution.md
```diff
@@ -0,0 +1,88 @@
...
+need to be specified in the flags.
+```
+kubebuilder init --plugins=go.kubebuilder.io/v3
+```
+
+```
+kuebebuilder create api --plugins=declarative.kubebuilder.io/v1 ...
+```
+For this command, the declarative plugin will be used instead of the base plugin,
+but following command calls won't use it unless provided again.
```
camilamacedo86, line 88: Let's imagine the scenario:

I scaffolded the project with SDK with `cli.WithDefaultPlugins(&golangv3.Plugin{}, &manifestsv1.Plugin{}, &scorecardv1.Plugin{}),`. Then, using kubebuilder I will run the create-api for this project. 

So, in POV it should work and KB CLI should raise the warning that the plugins  &manifestsv1.Plugin{}, &scorecardv1.Plugin{}) were not found and because of this not used. 

Could we add this example/scenario to the proposal as well?

designs/plugin-resolution.md
```diff
@@ -0,0 +1,88 @@
+# Plugin and project version resolution
+
```
camilamacedo86, line 2: IMO we need to improve this doc a little. We might add a template for the proposals here such as; https://github.com/operator-framework/enhancements/blob/master/enhancements/template.md

designs/plugin-resolution.md
Adirio: From a plugin resolution perspective, there is no difference. We are already resolving multiples plugins. It is latter when we are forced to only have 1. This EP only talks about the resolution proposal.

designs/plugin-resolution.md
Adirio: I dont think it should work. I think it should error out saying that manifest could not be resolved.

designs/plugin-resolution.md
Adirio: I was thinking that this probably shouldnt be merged, we should treat this as an issue nad close it when solved.

Creating a template for EPs sounds fine though.

designs/plugin-resolution.md
```diff
@@ -0,0 +1,88 @@
...
+     * If they are available, plugin keys will be fully qualified.
+2. [Init command only] Obtain the project version from `--project-version` flag.
+     * The flag is optional, we will try to resolve project version later in case it is not provided.
+3. Obtain the optional plugin keys from `--plugins` flag.
+     * [Init command only] If not present, use default plugins.
+     * [All commands but Init] If not present, use plugin keys from configuration file (step 1).
+4. Qualify the unqualified keys from step 3.
+     * If we know the project version, use this info to filter the available plugins so that we can be more accurate.
+5. Resolve the plugins with the list of qualified keys above.
+6. [Init command only] Resolve the project version in case it wasn't provided and there is only one project version
```
estroz, line 40: What if two default plugins support the same project version?

estroz: Actually this doesn't seem to be a problem if default plugins will have a disjoint set of supported project versions.

designs/plugin-resolution.md
```diff
@@ -0,0 +1,88 @@
...
+* Command calls are not allowed to override plugin keys.
+
+## Proposal
+
+The following plugin resolution algorithm solves the flaws described above:
+
+1. [All commands but Init] Obtain the project version and plugin keys from the project configuration if available.
+     * If they are available, plugin keys will be fully qualified.
+2. [Init command only] Obtain the project version from `--project-version` flag.
+     * The flag is optional, we will try to resolve project version later in case it is not provided.
```
estroz, line 33: There will always be a default project version. If no default plugins support the default project version, then an error should be returned.

designs/plugin-resolution.md
Adirio: > What if two default plugins support the same project version? Which will be chosen?

This proposal takes into account that we will support plugin chains. Having multiple plugins set as default means that it is a default chain. This proposal removes the ability to provide different plugin/plugin-chains for each project version, but I guess we could try to keep this behavior.

designs/plugin-resolution.md
Adirio: The idea here is, in case they provide a project version, make decissions knowing it, but in case it wasn't provide, just make decissions as if it was unknown and try to resolve it later.

Imagine I have the following plugins resolved:
Plugin | Supported project version
---|---
go/v2 | 2, 3
go/v3 | 3
declarative/v1 | 2, 3

If the user provides a `--plugins=go/v3,declarative` flag, we would resolve to the last two plugins, and the only supported project version by both of them is `3` so we can implicitly resolve the project version.
If the user provides a `--plugins=go/v2,declarative` flag, we would resolve to the first and last plugin, and two different project versions are supported. If any of those is the default project version configured for the CLI, that will be used.

So we will have 3 ways of resolving the project version:
- Explicitly by providing `--project-version` flag or having an already present config file (`version` field).
- Implicitly by providing a set of plugins that only support a single common project version.
- Default in case all resolved plugins support the default project version defined in the CLI.

PR #1943, opened 2021-01-13
:sparkles: Make directory-not-empty error more self-explanatory

justinsb:
When I encountered the error, I had to go to the source to understand
the problem.  Made the message more user-focused.
    
Also skipped printing the allowlist, both to fit into lint line-length
limits and to avoid overwhelming the user.

### DIFF
--- a/pkg/plugins/golang/v3/init.go
+++ b/pkg/plugins/golang/v3/init.go
@@ -17,7 +17,6 @@ limitations under the License.
 package v3
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -188,7 +187,9 @@ func checkDir() error {
 				return err
 			}
 			if info.Name() != "go.mod" && !strings.HasPrefix(info.Name(), ".") {
-				return errors.New("only the go.mod and files with the prefix \"(.)\" are allowed before the init")
+				return fmt.Errorf(
+					"target directory is not empty (only go.mod and files with the prefix \".\" are allowed); found existing file %q",
+					path)
 			}
 			return nil
 		})


pkg/plugins/golang/v3/init.go
```diff
@@ -188,7 +187,7 @@ func checkDir() error {
 				return err
 			}
 			if info.Name() != "go.mod" && !strings.HasPrefix(info.Name(), ".") {
-				return errors.New("only the go.mod and files with the prefix \"(.)\" are allowed before the init")
+				return fmt.Errorf("target directory is not empty: found existing file %q", path)
```
Adirio, line 190: NIT: I think the file name should be enough
```suggestion
				return fmt.Errorf("target directory is not empty: found existing file %q", info.Name())
```

justinsb: I started with that actually and switched to path :-) The root cause here was that we were finding files like `.git/HEAD`.  info.Name() prints "HEAD", but path prints ".git/HEAD".

justinsb:
/retest  pull-kubebuilder-e2e-k8s-1-16-2

I can't see how that failure could be related, so rolling the dice again!

estroz (approved):
/approve

Adirio:
My only question is the one raised in https://github.com/kubernetes-sigs/kubebuilder/pull/1944#issuecomment-759508897
I will `lgtm` and `hold` it, If you consider that we should do it in two PRs just remove the `hold` and it will merge. If you opt into a single PR then close it and update the other PR accordingly.

/lgtm
/hold

pkg/plugins/golang/v3/init.go
camilamacedo86: Should we not still clarifying what files are allowed? See that any file with the prefix (.) is allowed. E.g. .gitignore

justinsb: My thought was that maybe the user doesn't care, but it is a good idea to try to reference the idea that some files are allowed, so I added it back!  Thanks!

estroz (approved):
/lgtm

PR #1944, opened 2021-01-13
:sparkles: Tolerate "dot" directories when checking if dir is empty

justinsb:
Many users will be running version control; we should ignore .git
directories also therefore.

### DIFF
--- a/pkg/plugins/golang/v3/init.go
+++ b/pkg/plugins/golang/v3/init.go
@@ -207,9 +207,16 @@ func checkDir() error {
 			if err != nil {
 				return err
 			}
-			if info.Name() != "go.mod" && !strings.HasPrefix(info.Name(), ".") {
+			// Allow the whole .git directory tree
+			if info.IsDir() && strings.HasPrefix(info.Name(), ".") && info.Name() != "." {
+				return filepath.SkipDir
+			}
+			// Also allow go.mod and dot-files
+			if info.Name() != "go.mod" && info.Name() != "go.sum" && !strings.HasPrefix(info.Name(), ".") {
 				return fmt.Errorf(
-					"target directory is not empty (only go.mod and files with the prefix \".\" are allowed); found existing file %q",
+					"target directory is not empty "+
+						"(only go.mod, go.sum, and files and directories with the prefix \".\" are allowed); "+
+						"found existing file %q",
 					path)
 			}
 			return nil


justinsb:
I'll rebase this to fix the lint line-length problem after #1943

Adirio:
Does #1943 justify doing this in 2 different PRs? Maybe you can use two commits if you want to split the clarity change and the functionality change.

pkg/plugins/golang/v3/init.go
```diff
@@ -187,8 +186,13 @@ func checkDir() error {
...
 			}
+			// Allow the whole .git directory tree
+			if info.Name() == ".git" {
+				return filepath.SkipDir
+			}
+			// Allow go.mod and dot-files
 			if info.Name() != "go.mod" && !strings.HasPrefix(info.Name(), ".") {
-				return errors.New("only the go.mod and files with the prefix \"(.)\" are allowed before the init")
+				return fmt.Errorf("target directory is not empty: only go.mod, files with the prefix \".\" and .git directories can exist before init (found %q)", path)
 			}
```
camilamacedo86, line 196: could you fix the lint issue? See that the line is too big.

Adirio:
> I'll rebase this to fix the lint line-length problem after #1943

@justinsb as you expected, this PR needs a rebase and don't forget about the lll linting error.

pkg/plugins/golang/v3/init.go
```diff
@@ -184,9 +184,16 @@ func checkDir() error {
 			if err != nil {
 				return err
 			}
+			// Allow the whole .git directory tree
+			if info.Name() == ".git" {
```
estroz, line 188: If git is being treated specially, why not mercurial/svn?

Adirio: I would say that any '.' prefixed directory/file, `go.mod` and `go.sum` should be allowed.

camilamacedo86: Yes, we need to allow any file/directory with the prefix (.).

daenney: I proposed this in #1972 but the feedback there was that this wasn’t desirable?

Adirio:
```suggestion
			if info.IsDir() && strings.HasPrefix(info.Name(), ".") && info.Name() != "." {
```
Check that it is a directory to avoid returning the skip dir error on non-dirs, checking that the directory starts by `.` and that it is not  exactly `.` (or we will skip everything).

pkg/plugins/golang/v3/init.go
```diff
@@ -184,9 +184,16 @@ func checkDir() error {
...
+				return filepath.SkipDir
+			}
+			// Also allow go.mod and dot-files
 			if info.Name() != "go.mod" && !strings.HasPrefix(info.Name(), ".") {
```
Adirio, line 192:
```suggestion
			if info.Name() != "go.mod" && info.Name() != "go.sum" && !strings.HasPrefix(info.Name(), ".") {
```

pkg/plugins/golang/v3/init.go
```diff
@@ -184,9 +184,16 @@ func checkDir() error {
...
 				return fmt.Errorf(
-					"target directory is not empty (only go.mod and files with the prefix \".\" are allowed); found existing file %q",
+					"target directory is not empty "+
+						"(only go.mod, files with the prefix \".\" and .git directories are allowed); "+
```
Adirio, line 195:
```suggestion
						"(only go.mod, go.sum, and files and directories with the prefix \".\" are allowed); "+
```

Adirio (changes requested):
Once the three suggested changes (one of them is in the above comment) are applied, this LGTM.

justinsb:
Thanks for all the suggestions - I incorporated them all (I believe!)

Adirio (approved):
/lgtm

Adirio:
Just a small NIT, can you edit the title? PR titles are used for changelogs and it is no longer only `.git`

justinsb:
Thanks @Adirio - good point - changed the PR title to match the commit title (which I had remembered to update!).

camilamacedo86 (approved):
/lgtm

PR #1945, opened 2021-01-13
:sparkles: Remove misleading help that mentioned dep

justinsb:
dep is no longer used by the project (we use go mod instead); remove
the help test that suggests we will suggest it.


camilamacedo86 (approved):
/lgtm

PR #1958, opened 2021-01-18
:warning: upgrade kubebuilder module from v2 to v3

camilamacedo86:
**Description**
upgrade kubebuilder module from v2 to v3

**Motivation**
https://github.com/kubernetes-sigs/kubebuilder/issues/1798


generate_testdata.sh
```diff
@@ -18,7 +18,7 @@ source common.sh
 
 build_kb() {
   header_text "Building kubebuilder"
-  go build -o ./bin/kubebuilder ./cmd
+  go build -o ./bin/kubebuilder sigs.k8s.io/kubebuilder/v3/cmd
```
estroz, line 21: I don't think this is necessary

estroz (approved):
One question. Also @Adirio this doesn't affect the v2.3.2 release correct?

/approve

camilamacedo86:
Hi @estroz, 

> One question. Also @Adirio this doesn't affect the v2.3.2 release correct?

How it would affect the  v2.3.2 release? It has its branch https://github.com/kubernetes-sigs/kubebuilder/tree/release-2 if we need to do a patch release.

generate_testdata.sh
Adirio: This is definetely not needed and the `./cmd` will be valid forever.

Adirio:
> Also @Adirio this doesn't affect the v2.3.2 release correct?

It shouldn't, its on a separate branch. We added v2 and it didnt affect v2.3.1 because it was already on a separate branch.

Adirio:
My issue with this PR is that it will break any pending PR and right now there are several. It needs to be done but I think we need to time it a bit better.

camilamacedo86:
Hi @Adirio, 

 > My issue with this PR is that it will break any pending PR and right now there are several. It needs to be done but I think we need to time it a bit better.

It is life. We should be done it already before the first nightly alpha release for v3. We need to move forward here to prepare a stable one. we cannot wait until have no longer open prs because it never will happen. So, It is time.

Adirio:
The changes done here are correct except for the mentioned one above. Haven't checked if there ia any v2 remaining though.

generate_testdata.sh
camilamacedo86: Solved.

PR #1961, opened 2021-01-18
:sparkles: bump controller-runtime to 0.8.3

varshaprasad96:
deps: Bump controller-runtime to 0.8.0

### DIFF
--- a/docs/book/src/migration/manually_migration_guide_v2_v3.md
+++ b/docs/book/src/migration/manually_migration_guide_v2_v3.md
@@ -359,9 +359,9 @@ require (
 	github.com/go-logr/logr v0.3.0
 	github.com/onsi/ginkgo v1.14.1
 	github.com/onsi/gomega v1.10.2
-	k8s.io/apimachinery v0.19.2
-	k8s.io/client-go v0.19.2
-	sigs.k8s.io/controller-runtime v0.7.0
+	k8s.io/apimachinery v0.20.2
+	k8s.io/client-go v0.20.2
+	sigs.k8s.io/controller-runtime v0.8.3
 )
 ```
 
--- a/docs/book/src/migration/migration_guide_v2tov3.md
+++ b/docs/book/src/migration/migration_guide_v2tov3.md
@@ -39,7 +39,7 @@ go mod init tutorial.kubebuilder.io/migration-project
 <aside class="note warning">
 <h1> Migrating to Kubebuilder v3 while staying on the go/v2 plugin </h1>
 
-You can use `--plugins=go/v2` if you wish to continue using "`Kubebuilder 2.x`" layout and avoid dealing with the breaking changes that will be faced because of the default upper versions which will be used now. See that the [controller-tools][controller-tools] `v0.4.1` & [controller-runtime][controller-runtime] `v0.7.0` are just used by default with the `go/v3` plugin layout. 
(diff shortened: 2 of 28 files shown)


varshaprasad96:
Blocked on #[146](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/146). Need to regenerate test data for `project-v3-addon` and `project-v3-config` plugins.

testdata/project-v3-addon/PROJECT
```diff
@@ -2,20 +2,4 @@ domain: testproject.org
...
-- api:
-    crdVersion: v1
-  group: crew
-  kind: FirstMate
-  version: v1
-- api:
-    crdVersion: v1
-  group: crew
-  kind: Admiral
-  version: v1
```
camilamacedo86, line 20: Hi @varshaprasad96, 

Shows that have something broked with the addon partner. we might to need to do changes first in sigs.k8s.io/kubebuilder-declarative-pattern.

c/c @estroz

varshaprasad96: Yes, the samples for v3-addon need to be regenerated after https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/146 is merged

Adirio:
Note: current PR title and description checks do not understand the `[WIP]` part, so they will fail anyway, but once you remove it, the emoji (✨ `:sparkles:`)  is expected without brackets, so you will also have to remove those.

P.S.: the next version of the PR title verifier will understand the [WIP] prefix.

varshaprasad96:
The PR has been updated. Waiting for v3.0.0 release.
/hold

Adirio:
@varshaprasad96 Could you remove the brackets from the title's emoji? Either ✨ or :sparkles: is fine, (and GitHub started understanding emojis on titles).

camilamacedo86:
HI @estroz, 

You spoke about we push it in 3.1.0. However, wdyt? Could it get merged now or should we wait? if it gets merged than it means that if we need to rebase master with the release-v3 before 3.0.0 it will ship as well. I have not an objection at all with 3.0.0 be using controller-runtime 0.8.0.

Adirio:
The reason that was given (I think it was @estroz but not 100% sure) was that there is no kubebuilder version that supports v0.7.0 which is the first version supporting kubernetes 1.19.x so that it would be nicer to have a v3.0.0 with v0.7.0 and then upgrade it for v3.1.0.

Adirio (approved):
/ok-to-test

camilamacedo86 (dismissed):
I am ok with it get merged. I do not think that would be a requirement we have any version of the tool using 0.7.0. 

@varshaprasad96 could you only please update the migration doc and bump the version 0.8.0 where we have 0.7.0? 

c/c @estroz

varshaprasad96:
Based on the discussion, I am reverting the migration doc changes, since the default version for v3 stable would be c-r 0.7.0

estroz:
>since the default version for v3 stable would be c-r 0.7.0

@varshaprasad96 yes but after this PR merges it will be 0.8.0 right?

Adirio:
Even if the docs are changed in master they won't get reflected in the published book until a new release that contains this merged PR is made, and then the version would be 0.8.0.

camilamacedo86:
@varshaprasad96 @estroz,

This PR bumps the controller-runtime  to 0.8.0
When it gets merged we need also update the docs
The production docs are only updated as part of the release where we push the master changes to the book branch. So, the master docs need to reflect the master state and update the doc is part of the PR. If we need to revert we will also revert the docs changes. 

Is it make sense?

varshaprasad96:
Sorry, got confused ! Updated the migration docs

camilamacedo86:
We need also have the commits squashed as well.

estroz:
@varshaprasad96 this can be rebased and bumped to controller-runtime v0.8.3

varshaprasad96:
@estroz Done. bumped to controller-runtime v0.8.3

PR #1959, opened 2021-01-18
:warning: make project version == 3 instead of 3-alpha

camilamacedo86:
**Description**
- make project version == 3 instead of 3-alpha

**Motivation**
- 3.0.0 release stable
- Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/1832

PR #1916, opened 2020-12-17
:book: upgrade migration guide for v3

camilamacedo86:
**Description**
- upgrade the migration guide with the steps required for v3 so far 

**To review**

Check in the preview link: https://deploy-preview-1916--kubebuilder.netlify.app/migration/v2vsv3.html

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/1829

PR #1968, opened 2021-01-22
✨ Add the rest of the missing fields and stabilize config v3

Adirio:
Create the v3 config implementation that includes all missing fields.

Closes: #1772
Closes: #1826
Closes: #1959

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -64,8 +64,8 @@ take care of building and publishing the artifacts.
 |   Name	|   Example	|  Description |
 |---	|---	|---	|
 |  KubeBuilder version | `v2.2.0`, `v2.3.0`, `v2.3.1` | Tagged versions of the KubeBuilder project, representing changes to the source code in this repository. See the [releases][kb-releases] page for binary releases. |
-|  Project version |  `"1"`, `"2"`, `"3-alpha"` | Project version defines the scheme of a `PROJECT` configuration file. This version is defined in a `PROJECT` file's `version`. |
-|  Plugin version | `v2`, `v3-alpha` | Represents the version of an individual plugin, as well as the corresponding scaffolding that it generates. This version is defined in a plugin key, ex. `go.kubebuilder.io/v2`. See the [design doc][cli-plugins-versioning] for more details. |
+|  Project version |  `"1"`, `"2"`, `"3"` | Project version defines the scheme of a `PROJECT` configuration file. This version is defined in a `PROJECT` file's `version`. |
+|  Plugin version | `v2`, `v3` | Represents the version of an individual plugin, as well as the corresponding scaffolding that it generates. This version is defined in a plugin key, ex. `go.kubebuilder.io/v2`. See the [design doc][cli-plugins-versioning] for more details. |
 
 ### Incrementing versions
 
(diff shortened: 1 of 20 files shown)


cmd/main.go
```diff
@@ -37,6 +38,7 @@ func main() {
 		),
 		cli.WithDefaultPlugins(cfgv2.Version, &pluginv2.Plugin{}),
 		cli.WithDefaultPlugins(cfgv3alpha.Version, &pluginv3.Plugin{}),
+		cli.WithDefaultPlugins(cfgv3.Version, &pluginv3.Plugin{}),
```
camilamacedo86, line 41: Will not we replace the cfgv3alpha for cfgv3? I think also it can be the next follow-up after this one. WDYT?

pkg/config/v2/config_test.go
```diff
@@ -78,7 +78,7 @@ var _ = Describe("cfg", func() {
 		})
 	})
 
-	Context("Name", func() {
+	Context("ProjectName", func() {
```
camilamacedo86, line 81: 👍

pkg/config/v3/config.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+// SetMultiGroup implements config.Config
+func (c *cfg) SetMultiGroup() error {
+	c.MultiGroup = true
+	return nil
+}
+
+// ClearMultiGroup implements config.Config
+func (c *cfg) ClearMultiGroup() error {
```
camilamacedo86, line 130: cannot the set receive a boolean instead of we need 2 methods for that?

pkg/config/v3/config.go
```diff
@@ -0,0 +1,349 @@
...
+func (c *cfg) SetComponentConfig() error {
+	c.ComponentConfig = true
+	return nil
+}
+
+// ClearComponentConfig implements config.Config
+func (c *cfg) ClearComponentConfig() error {
+	c.ComponentConfig = false
+	return nil
+}
```
camilamacedo86, line 150: Same scenario here.

pkg/plugins/golang/v3/plugin.go
```diff
@@ -26,7 +27,7 @@ import (
 const pluginName = "go" + plugins.DefaultNameQualifier
 
 var (
-	supportedProjectVersions = []config.Version{cfgv3alpha.Version}
+	supportedProjectVersions = []config.Version{cfgv3alpha.Version, cfgv3.Version}
```
camilamacedo86, line 30: IMO as described above. The goal of this PR is only to store the data in the PROJECT file. Then, a follow up would be responsible to make the project-version 3 stable by replacing/renaming v3-alpha per v3 only.

cmd/main.go
Adirio: It can be done either ina follow up or in this PR, but for now I left both so that we can start testing things.

camilamacedo86: IHMO, if you add both than it means more work do, +1 e2e test + generate data and etc .. as it will make harder to the reviewers. So, if you close the scope here for it does one 1 proposed idea whcih is store in the PROJECT file the attrs fields I think we can move smothier.

pkg/config/v3/config.go
Adirio: It could, it just was defined the other way round when I created the `Config` API

testdata/project-v3/PROJECT
```diff
@@ -5,23 +5,39 @@ repo: sigs.k8s.io/kubebuilder/testdata/project-v3
...
   kind: FirstMate
+  path: sigs.k8s.io/kubebuilder/testdata/project-v3/api/v1
   version: v1
   webhooks:
+    conversion: true
     webhookVersion: v1
 - api:
     crdVersion: v1
+  controller: true
+  domain: testproject.org
```
camilamacedo86, line 34: we will not hide, not stored the command when it is == the projects domain?

Adirio: Its unlucky but an empty domain is valid so we can't hide it to mean that it is the default.

docs/book/src/migration/project/v2_v3.md
```diff
@@ -68,47 +68,21 @@ fine.
...
-Currently, the plugin `v3-alpha` has NO breaking changes. However, until it is declared stable breaking changes can be made related to K8s API deprecations of `v1beta1` versions of `CustomResourceDefinition` and `ValidatingWebhookConfiguration` and to upgrade the `cert-manager`. More info:
-
-- [cert-manager related configuration should be migrated to cert-manager.io/v1 #1666
-](https://github.com/kubernetes-sigs/kubebuilder/issues/1666)
-- [Set preserveUnknownFields to false in the CRD conversion webhook patch #933](https://github.com/kubernetes-sigs/kubebuilder/issues/933)
-- [Migrate existing KB project to v1 CRDs and Webhooks with minimal user effort #1065
-](https://github.com/kubernetes-sigs/kubebuilder/issues/1065)
-
-</aside>
-
```
camilamacedo86, line 96: We can add here //TODO and cove that in the follow up

Adirio: Not sure what you mean here. I removed those two notes because now that go/v3 is the default plugin, they no longer apply.

camilamacedo86: It is in the migration doc. After we merge that I will update the PR https://github.com/kubernetes-sigs/kubebuilder/pull/1916 where we will cover it. so, for now, a //TODO shows fine..

testdata/project-v3/PROJECT
camilamacedo86: I think we just need to add a small doc over what is and how works the PROJECT file in https://book.kubebuilder.io/reference/reference.html. However, in PV it is not a blocker for this one at all and be done as a follow up as well.

camilamacedo86 (approved):
It shows great for me 🥇 
I added just a few nits.

@estroz  wdyt?

VERSIONING.md
```diff
@@ -76,32 +76,31 @@ Project versions should only be increased if a breaking change is introduced in
...
 You create a feature that adds a new marker to the file `main.go` scaffolded by `init`
 that `create api` will use to update that file. The changes introduced in your feature
 would cause errors if used with projects built with plugins `go.kubebuilder.io/v2`
 without users manually updating their projects. Thus, your changes introduce a breaking change
 to plugin `go.kubebuilder.io`, and can only be merged into plugin version `v3-alpha`.
-This plugin's package should exist already, so a PR must be made against the
-
-You must also add a migration guide to the [migrations](https://book.kubebuilder.io/migrations.html)
-section of the KubeBuilder book in your PR. It should detail the steps required
-for users to upgrade their projects from `v2` to `v3-alpha`.
```
camilamacedo86, line 104: Here I think we need to only replace v3-alpha for v3. We will have the migration steps after that as well.

Adirio: I placed that part before the example so that it doesn't look as part of the example and replaced `v2` and `v3-alpha` with `vX` and `v(X+1)-alpha`.

docs/book/src/migration/project/v2_v3.md
```diff
@@ -68,47 +68,23 @@ fine.
...
-Currently, the plugin `v3-alpha` has NO breaking changes. However, until it is declared stable breaking changes can be made related to K8s API deprecations of `v1beta1` versions of `CustomResourceDefinition` and `ValidatingWebhookConfiguration` and to upgrade the `cert-manager`. More info:
-
-- [cert-manager related configuration should be migrated to cert-manager.io/v1 #1666
-](https://github.com/kubernetes-sigs/kubebuilder/issues/1666)
-- [Set preserveUnknownFields to false in the CRD conversion webhook patch #933](https://github.com/kubernetes-sigs/kubebuilder/issues/933)
-- [Migrate existing KB project to v1 CRDs and Webhooks with minimal user effort #1065
-](https://github.com/kubernetes-sigs/kubebuilder/issues/1065)
-
-</aside>
+In order to migrate your projects to v3 some changes need to be manually done to the PROJECT file, and some other scaffolded files.
```
irvifa, line 71: Is there any example on this process?

camilamacedo86: It will be done in another PR.

irvifa (approved):

christopherhein:
Nice work, this lgtm!

camilamacedo86:
This PR has been addressing the changes in the PROJECT file which has been discussed for a while and that we reach a consensus. Also, I have been adding this PR for all week in the Slack channel to see if someone has some objection. It has 3 oks already then, it shows good enough to get merged: 

/lgtm 

PS.:See that the PR https://github.com/kubernetes-sigs/kubebuilder/pull/1869 adds the same fields and also was approved by @estroz as well in the past. The reason for the old pr not get merged was the discussions regards the model changes which now are addressed properly via the PR; https://github.com/kubernetes-sigs/kubebuilder/pull/1911.

PR #1972, opened 2021-01-24
:bug: Don't recurse into hidden directories

### LINKED ISSUE #1971: failed to initialize project with "go.kubebuilder.io/v3"
```                                                                                                                
➜  kubeact-operator git:(feat/v0.0.1) ✗ kubebuilder init                                                                                        git:(feat/v0.0.1|●53✚1…2 
Error: failed to initialize project with "go.kubebuilder.io/v3": only the go.mod and files with the prefix "(.)" are allowed before the init
Usage:
  kubebuilder init [flags]

Examples:
  # Scaffold a project using the apache2 license with "The Kubernetes authors" as owners
  kubebuilder init --project-version=2 --domain example.org --license apache2 --owner "The Kubernetes authors"


Flags:
      --component-config         create a versioned ComponentConfig file, may be 'true' or 'false'
      --domain string            domain for groups (default "my.domain")
      --fetch-deps               ensure dependencies are downloaded (default true)
      --license string           license to use to boilerplate, may be one of 'apache2', 'none' (default "apache2")
      --owner string             owner to add to the copyright
      --plugins strings          Name and optionally version of the plugin to initialize the project with. Available plugins: ("go.kubebuilder.io/v2", "go.kubebuilder.io/v3")
      --project-name string      name of this project
      --project-version string   project version, possible values: ("2", "3-alpha") (default "3-alpha")
      --repo string              name to use for go module (e.g., github.com/user/repo), defaults to the go package of the current working directory.
      --skip-go-version-check    if specified, skip checking the Go version

2021/01/24 07:14:18 failed to initialize project with "go.kubebuilder.io/v3": only the go.mod and files with the prefix "(.)" are allowed before the init
➜  kubeact-operator git:(feat/v0.0.1) ✗ ls                                                                                                                                       git:(feat/v0.0.1|●53✚1…2 
go.mod  go.sum
➜  kubeact-operator git:(feat/v0.0.1) ✗ 
```

* go version go1.15.6 linux/amd64
* Version: main.version{KubeBuilderVersion:"3.0.0-alpha.1", KubernetesVendor:"1.19.2", GitCommit:"1428bba7c162cf2b783c89fd6f07f4efbae97130", BuildDate:"2021-01-08T22:30:36Z", GoOs:"linux", GoArch:"amd64"}

/kind bug

daenney:
checkDir() currently recurses into hidden directories at the root. This
is a problem, because only the directory itself will have a . at the
root of it's path, but any files within it won't when calling info.Name.
This causes init to fail on a directory containing any hidden directory
with files within them, such as .git.

Since we use filepath.Walk(".") we can't skip a directory that just
starts with ., as that would include the current directory. So instead
we also check that Name() doesn't return just "."

Fixes #1971

### DIFF
--- a/pkg/plugins/golang/v3/init.go
+++ b/pkg/plugins/golang/v3/init.go
@@ -207,6 +207,9 @@ func checkDir() error {
 			if err != nil {
 				return err
 			}
+			if info.IsDir() && info.Name() != "." && strings.HasPrefix(info.Name(), ".") {
+				return filepath.SkipDir
+			}
 			if info.Name() != "go.mod" && !strings.HasPrefix(info.Name(), ".") {
 				return fmt.Errorf(
 					"target directory is not empty (only go.mod and files with the prefix \".\" are allowed); found existing file %q",


pkg/plugins/golang/v3/init.go
```diff
@@ -207,6 +207,9 @@ func checkDir() error {
 			if err != nil {
 				return err
 			}
+			if info.IsDir() && info.Name() != "." && strings.HasPrefix(info.Name(), ".") {
```
camilamacedo86, line 210: Would not it allow create a new project in a dit whcih has another dir inside?

camilamacedo86, line 210: So, it would allow scaffold the project when the dir starts with (.) as well. Am I right?

pkg/plugins/golang/v3/init.go
```diff
@@ -207,6 +207,9 @@ func checkDir() error {
...
+				return filepath.SkipDir
+			}
```
camilamacedo86, line 212: Will not it allow scaffold projects in a directory that has another directory as well? 
Then, note that it goes against the purpose of the check. We would like to not allow it because it means that we cannot ensure that the tool is able to do properly all required scaffolds.

daenney: I believe the check is too paranoid. it needs to not only exclude files starting with a dot, it needs to not recurse into hidden directories.

It will only allow you to scaffold into directories that have hidden directories at the root. For example, if you try to scaffold right now in a directory that you've just run a `git init` on, it'll fail. Though `.git` has a `.` prefix, the `HEAD` file in `.git/HEAD` does not, so it'll fail because `info.Name()' will return `HEAD`, not `.git/HEAD`.

Similarly if you have opened that directory in an editor already and something like a `.vscode` exists it'll fail to scaffold. None of these directories should have any impact on the ability to scaffold. If we ignore hidden files, we should ignore hidden directories too.

daenney:
Closed in favour of #1944

Issue #1973, opened 2021-01-26
Resource validations


Adirio:
# Description
Currently, the `Options` objects, which are used to store all the data input from flags, are the ones that are being validated so that they can create a new `Reosurce`. While some of these validations do belong to the `Options` objects, there are several of these validations that should belong to the `Resource` object itself.

# Proposal
Split the validations in those belonging to the `Resource` objects and those that should still be part of the `Options` objects so that `Resources` can be built from different sources and still be validated.

/kind feature

PR #1974, opened 2021-01-26
⚠ Resource validation

Adirio:
Split resource validation between Options and Resource objects so that Resources built in different ways can still be validated.

Fixes: #1973

### DIFF
--- a/pkg/model/resource/resource.go
+++ b/pkg/model/resource/resource.go
@@ -19,6 +19,8 @@ package resource
 import (
 	"fmt"
 	"strings"
+
+	"sigs.k8s.io/kubebuilder/v3/pkg/internal/validation"
 )
 
 // Resource contains the information required to scaffold files for a resource.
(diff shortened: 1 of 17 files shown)


pkg/model/resource/resource.go
```diff
@@ -19,6 +19,8 @@ package resource
 import (
 	"fmt"
 	"strings"
+
+	"sigs.k8s.io/kubebuilder/v3/pkg/internal/validation"
```
camilamacedo86, line 23: consumers cannot use the internal one. Should it be really internal?

Adirio: I didn't touch that, it was internal previously. I think the reason to make it internal is that it was borrowed code from another repo that we didn't wan't to add as a dependency.

pkg/plugins/golang/v2/api.go
```diff
@@ -44,6 +45,8 @@ type createAPISubcommand struct {
 
 	options *Options
 
+	resource resource.Resource
```
estroz, line 48: This field isn't set and passed to the scaffold constructor until `Run()` and no flags use it unless I'm mistaken, so what's the point of having a field instead of a local variable within `Run()`? The `resource.Validate()` call doesn't seem to be doing anything.

Adirio: I'm not sure what do you mean here. Do you mean that `resource.Validate()` doesn't change it? That we can have a local variable in `Run` and pass it to `Validate` and `GetScaffolder` methods? In this case `Validate` and `GetScaffolder` methods are used for all plugins through `cmdutil.Run` and not all of them have a `resource.Resource`.

estroz: I think I see what you mean. I forgot that `cmdutil.Run()` will call `Validate()` before `GetScaffolder()` and therefore `p.resource` is set up before `Validate()` is called.

pkg/plugins/golang/v2/api.go
```diff
@@ -142,15 +138,30 @@ func (p *createAPISubcommand) Validate() error {
...
+	p.resource = p.options.NewResource(p.config)
+
+	return cmdutil.Run(p)
+}
+
+func (p *createAPISubcommand) Validate() error {
+	if err := p.options.Validate(); err != nil {
+		return err
+	}
+
```
estroz, line 151: Won't `Options.Validate()` be a subset of `Resource.Validate()`?
```suggestion
```

Adirio: `Options.Validate` only does the validation directly related to the flag parsing. The idea is that third-party pugins don't have to use the `Options` objects as those are just how go plugin flag options are parsed, but still benefit from the validation that was previously being done there and now is mvoed to `Resource.Validate()`.

estroz: Sure the `Options.Validate()` method should still exist. I'm suggesting that in this particular method that calling `Options.Validate()` is redundant since `Resource.Validate()` is called afterwards, and `p.options` isn't used in `cmdutil.Run(p)`.

Adirio: `Options.Validate()` still does some validations like checking that the values for the flags are not empty. Those validations do not belong to the `Resource` itself.

Issue #1956, opened 2021-01-18
Scaffold files atomicity


Adirio:
# Definition
Atomicity or atomic operations in this context is understood as the capability of the scaffolder to either fully succeed or fully fail an operation execution, without the possibility of having partial results in case of failure.

# Background
Currently, kubebuilder commands are non-atomic, there are several errors that can happen that will make the commands scaffold certain files but not fully succeed. List of this cases (I may be missing some case):
1. If a file whose `IfFileExists` field is set to `Error` is placed in the n-th position, the files before (until the {n-1}-th) will be persisted to disk.
2. If a command uses several `NewScaffolder().Execute(...)` calls (case of v2/init, v2/api, v3/init, v3/api & v3/webhook) and one of them fails, files in previous calls to the failing one will be persisted to disk.
3. If the `Write` call fails, previous files will be persisted to disk.

Handling the 3rd scenario marked above is pretty complex, as it would require to store a full copy of the whole directory and revert to it in case that any `Write` fails, but reverting also requires to call `Write` which could also fail.

# Feature request
Make kubebuilder commands as atomic as possible.

- In case any non-`Write`-related error happens, kubebuilder commands should be atomic.
- In case a `Write`-related error happens, kuebuilder should print a warning telling the user that the execution was non-atomic, and probably also mention the files that were added/modified.

# Requirements
- [x] It is trasparent to plugins.
- [x] It is maintainable.
- [ ] It is able to run post-scaffold actions.

/kind feature

camilamacedo86:
@Adirio, what was not very clear to us is how it would be implemented. However, the idea shows nice. So, could you please add an EP with details of the implementation or push a PR if it would be the easier/straightway to move forward?

PR #1977, opened 2021-01-27
🌱 Scaffolded files atomicity

Adirio:
Make command calls atomic (either all files are scaffolded or none is).

Closes: #1956


Adirio:
In order to implement scaffolding atomicity the following requirements need to be met:
- [x] It is trasparent to plugins.
- [x] It is maintainable.
- [ ] It is able to run post-scaffold actions.

This approach (using a memory filesystem and persisting the chnages at post-run) is unable to handle post-scaffold actions such as running `make` or any other that expects the files to be already created.

Issue #1637, opened 2020-08-12
Deploying manager pod as non-root


harpratap:
I tried deploying the manager on our K8s platform that doesn't allow root by default but ran into following issues even though there's nothing really in the manager that requires root
I see the following two errors -
1. `Error: container has runAsNonRoot and image will run as root` 
2. `Error: container has runAsNonRoot and image has non-numeric user (nonroot), cannot verify user is non-root`

We use the following podsecuritypolicy to ensure that pods are not running as root -
```
allowPrivilegeEscalation: true
  allowedCapabilities:
  - NET_RAW
  - NET_BIND_SERVICE
  fsGroup:
    rule: RunAsAny
  requiredDropCapabilities:
  - CHOWN
  - DAC_OVERRIDE
  - SETPCAP
  - MKNOD
  - AUDIT_WRITE
  - FOWNER
  - FSETID
  - KILL
  - SETGID
  - SETUID
  - SYS_CHROOT
  - SETFCAP
  - SYS_MODULE
  - SYS_RAWIO
  - SYS_PACCT
  - SYS_ADMIN
  - SYS_NICE
  - SYS_RESOURCE
  - SYS_TIME
  - SYS_TTY_CONFIG
  - AUDIT_CONTROL
  - MAC_OVERRIDE
  - MAC_ADMIN
  - NET_ADMIN
  - SYSLOG
  - DAC_READ_SEARCH
  - LINUX_IMMUTABLE
  - NET_BROADCAST
  - IPC_LOCK
  - IPC_OWNER
  - SYS_PTRACE
  - SYS_BOOT
  - LEASE
  - WAKE_ALARM
  - BLOCK_SUSPEND
  runAsUser:
    rule: MustRunAsNonRoot
  seLinux:
    rule: RunAsAny
  supplementalGroups:
    rule: RunAsAny
  volumes:
  - configMap
  - emptyDir
  - projected
  - secret
  - downwardAPI
  - persistentVolumeClaim
```

I believe this is happening because of two reasons -
1. The kube-rbac-proxy uses root by default, since it is using distroless/static as it's base image - https://github.com/brancz/kube-rbac-proxy/blob/master/Dockerfile#L1 I got this from this comment - https://github.com/GoogleContainerTools/distroless/issues/235#issuecomment-422099277

[...]

PR #1978, opened 2021-01-27
:sparkles: (go/v3) change 'runAsUser: 65532'  to 'runAsNonRoot: true'

estroz:
This PR changes the manager pod template's security context `runAsUser: 65532`  to `runAsNonRoot: true`. User 65532 is already specified in a project's [Dockerfile](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/Dockerfile#L25) which the [security context uses](https://pkg.go.dev/k8s.io/api/core/v1#SecurityContext) when `runAsUser` is unset, making the security context's value redundant (and prone to drift if that value is changed in the Dockerfile) in kubebuilder's case.

### DIFF
--- a/pkg/plugins/golang/v3/scaffolds/internal/templates/config/manager/config.go
+++ b/pkg/plugins/golang/v3/scaffolds/internal/templates/config/manager/config.go
@@ -69,7 +69,7 @@ spec:
         control-plane: controller-manager
     spec:
       securityContext:
-        runAsUser: 65532
+        runAsNonRoot: true
       containers:
       - command:
         - /manager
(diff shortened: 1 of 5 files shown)


Adirio (approved):
LGTM (not adding the label too early to let others review it if needed as it is already aproved)

camilamacedo86:
IMO to allow it gets merged, we need to ensure that after that the following error will not be faced.

`Error: container has runAsNonRoot and image has non-numeric user (nonroot), cannot verify user is non-root`
 
It was the motivation for the change raised in https://github.com/kubernetes-sigs/kubebuilder/issues/1637. 

So, we would need to check it with @harpratap.

@harpratap could you please let us know if after this change the above error will be faced? Are you ok with? 

c/c @estroz @Adirio

estroz:
@camilamacedo86 both the conditions listed in #1637 no longer hold (both Dockerfiles now specify non-root user IDs) so this will work with a restrictive pod security policy. You can check this yourself by applying the one specified in the issue.

harpratap:
@camilamacedo86 @estroz The changes lgtm

PR #1980, opened 2021-01-28
🌱 Filesystem and config persistence cleanup

Adirio:
- Uses a single filesystem:
    - Injected from `cli` to the scaffolders instead of creating multiple per scaffold.
    - Injected from `cli` to the config persistence layer instead of creating one per configuration representation.
- Moves configuration creation, loading and saving operations to pre-/post-run hooks.
- Cleans up subcommand binding methods.
- Creates a new `Store` interface that abstracts the persistence layer used for `Config`.

Closes: #1936


Adirio:
This should probably be applied with plugin phase 1.5 described in #1990.

Adirio:
Definetely closing this in favor of plugin phase 1.5 as implemented in #1991.

PR #1981, opened 2021-01-29
🌱 Improve Makefile help and rename "manager" target to "build"

Adirio:
Changes:
- `Makefile`: remove 2 double white-spaces
- `pkg/plugins/golang/v2/*`: added `help` and `undeploy` targets. Changed `manager` target to `build` (kept an allias for backwards compatibility). Reoredered targets, added categories and improved some help messages.
- `pkg/plugins/golang/v3/*`: Changed `manager` target to `build`. Reoredered targets, added categories and improved some help messages.

By keeping a `manager` alias to `build` target in v2 projects, it can be considered non-breaking. The added targets (`help` and `undeploy`) won't be available on projects initialized before this, but they are not breaking changes as they are only additions.

Technically speaking, projects built with go/v3 can't `make manager` any more, they should use `make build` but `all` target is an alias to them so the most common use is `make` which still works. As no stable release for v3 has been deployed and the impact of this change will be really small I think we can consider this non-breaking.

### DIFF
--- a/pkg/plugins/golang/v2/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v2/scaffolds/internal/templates/makefile.go
@@ -67,60 +67,75 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
-all: manager
+all: build
 
-# Run tests
-test: generate fmt vet manifests
+##@ General
+
+# The help target prints out all targets with their descriptions organized
+# beneath their categories. The categories are represented by '##@' and the
+# target descriptions by '##'. The awk commands is responsible for reading the
+# entire set of makefiles included in this invocation, looking for lines of the
+# file as xyz: ## something, and then pretty-format the target and help. Then,
+# if there's a line with ##@ something, that gets pretty-printed as a category.
+# More info on the usage of ANSI control characters for terminal formatting:
+# https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_parameters
+# More info on the awk command:
+# http://linuxcommand.org/lc3_adv_awk.php
+
+help: ## Display this help.
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+##@ Development
+
+manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
+	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+
+generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
+	$(CONTROLLER_GEN) object:headerFile={{printf "%q" .BoilerplatePath}} paths="./..."
+
+fmt: ## Run go fmt against code.
+	go fmt ./...
+
+vet: ## Run go vet against code.
+	go vet ./...
+
+test: manifests generate fmt vet ## Run tests.
 	go test ./... -coverprofile cover.out
 
-# Build manager binary
-manager: generate fmt vet
+##@ Build
+
+build: generate fmt vet ## Build manager binary.
 	go build -o bin/manager main.go
 
-# Run against the configured Kubernetes cluster in ~/.kube/config
-run: generate fmt vet manifests
+# Backwards compatibility
+manager: build ## Build manager binary (alias for build target).
+
+run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
-# Install CRDs into a cluster
-install: manifests kustomize
+docker-build: test ## Build docker image with the manager.
+	docker build -t ${IMG} .
+
+docker-push: ## Push docker image with the manager.
+	docker push ${IMG}
+
+##@ Deployment
+
+install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/crd | kubectl apply -f -
 
-# Uninstall CRDs from a cluster
-uninstall: manifests kustomize
+uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/crd | kubectl delete -f -
 
-# Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-deploy: manifests kustomize
+deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
-# Generate manifests e.g. CRD, RBAC etc.
-manifests: controller-gen
-	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
+	$(KUSTOMIZE) build config/default | kubectl delete -f -
 
-# Run go fmt against code
-fmt:
-	go fmt ./...
 
-# Run go vet against code
-vet:
-	go vet ./...
-
-# Generate code
-generate: controller-gen
-	$(CONTROLLER_GEN) object:headerFile={{printf "%q" .BoilerplatePath}} paths="./..."
-
-# Build the docker image
-docker-build: test
-	docker build . -t ${IMG}
-
-# Push the docker image
-docker-push:
-	docker push ${IMG}
-
-# find or download controller-gen
-# download controller-gen if necessary
-controller-gen:
+controller-gen: ## Download controller-gen locally if necessary.
 ifeq (, $(shell which controller-gen))
 	@{ \
 	set -e ;\
@@ -135,7 +150,7 @@ else
 CONTROLLER_GEN=$(shell which controller-gen)
 endif
 
-kustomize:
+kustomize: ## Download kustomize locally if necessary.
 ifeq (, $(shell which kustomize))
 	@{ \
 	set -e ;\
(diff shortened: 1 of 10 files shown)


pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
```diff
@@ -85,23 +85,46 @@ all: manager
...
+
+fmt: ## Run go fmt against code.
+	go fmt ./...
+
+vet: ## Run go vet against code.
+	go vet ./...
+
 ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
-test: generate fmt vet manifests ## Run tests.
+test: manifests generate fmt vet ## Run tests.
```
camilamacedo86, line 106: why the order was changed?

pkg/plugins/golang/v2/scaffolds/internal/templates/makefile.go
```diff
@@ -67,60 +67,75 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
-all: manager
+all: build
 
-# Run tests
-test: generate fmt vet manifests
+##@ General
```
camilamacedo86, line 72: So far all changes in the Makefile has been addressing only for v3. The rule/agreement only addresses bug fix to v2. 

The reason for that I because it might be a breaking change. For SDK tool for example, as a consumer, it would be since it has plugins ship customs on top.

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
```diff
@@ -85,23 +85,46 @@ all: manager
...
 # http://linuxcommand.org/lc3_adv_awk.php
 
-.PHONY: help
-help:  ## Display this help.
+help: ## Display this help.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
+##@ Development
+
+manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResource definition objects.
```
camilamacedo86, line 93:
```suggestion
manifests: controller-gen ## ## Generate manifests e.g. CRD, RBAC, etc..
```
I prefer less verbose and more generifc info as it is now.

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
```diff
@@ -85,23 +85,46 @@ all: manager
...
+	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+
+generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
```
camilamacedo86, line 96:
```suggestion
generate: controller-gen ## Generate manifests e.g. CRD, RBAC, etc..
```

The description that we have currently shown address better the need. See that it is too verbose which will not fit too much well for the help output. I do not think that we should detail here the files gen.

camilamacedo86 (commented):
I liked the biggest part of the changes done here. 
However, IMO: 
- we should only address the changes for v3 as it has been done
- I'd prefer the previous description help which is more user friendly and fits better for the help output. I would be ok, with a comment on top of the target that is not outputted by the help with the goal to provide further details and describe tthe diff between generate and manifests.  

So, I am happy with we move forward here if we address these points.

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
Adirio: `manifests` and `generate` are the two `controller-gen` calls, so I wanted to keep them grouped. It doesn't have any real implications as `manifests` just generates yaml files under `config` dir which are not affected by any of the 3 targets.

pkg/plugins/golang/v2/scaffolds/internal/templates/makefile.go
Adirio: What do you mean by customs on top? They were just reorder in order to group them by categories.

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
Adirio: The description I used is to match the one provided in the documentation: https://book.kubebuilder.io/reference/controller-gen.html#generators

pkg/plugins/golang/v2/scaffolds/internal/templates/makefile.go
camilamacedo86: SDK changes the Makefile. Add things on that. 
It adds the help and changes the order and etc .. 

And then, we have been asking for this kind of the change just be done in v3 so far. Indeed when the help feature was discussed we decided that it should be done only for v3 as well. As we have a previous agreement of mainly only push bug fix to v2. So, I do not think that we ought to do that.

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
camilamacedo86: OK. But there is a doc that has the goal of described detailed what is done.
The output of the tool is just a helper to let users know why/when they should run the command.

pkg/plugins/golang/v2/scaffolds/internal/templates/makefile.go
Adirio: The agreement was {N-2} bug fixes, {N-1} bug fixes and easy to back port features and enhancements, N all features. And go/v2 is still N or {N-1} (depending if you consider go/v3 to be N as it still is not implemented in any stable release). Either way, it is not {N-2} in any way so it is not restricted to bug fixes only.

About the paart that SDK adds helps and commands, I don't think we should not provide help ourselves just because SDK provides it, and you can still add new commands to the file, just add them in the correct category (or create a new one for SDK related targets) and add help messages so that the help target prints them out too.

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
Adirio: I feel that the help message for the `generate` target being `Generate code` is the same as having no help message. Same applies for `manifests` target and `Generate manifests` help message.

P.S. it should say `CustomeResourceDefinition` not `CustomResource definition`, fixing that part.

pkg/plugins/golang/v2/scaffolds/internal/templates/makefile.go
camilamacedo86: It is a change in the scaffold and is a breaking change for SDK(consumers). It will break the "plugins" that does custom on the Makefile, the logic to generate the samples, the tests and docs. 

Note that usually, a plugin to add features to a Kubebuilder project or a consumer will add targets in the Makefile or custom them. So, changes in its scaffold have a good impact. 

Regards SDK the burn will be the same because its using go/v3 already. In this way, we need to do changes in many places anyway. Also, the following change is not required to add help for example: 

````
# Backwards compatibility
manager: build ## Build manager binary (deprecated, use build target instead).
```

So, if we are ok changing the Makefile for v2 has any reason for v2 and v3 does not do exactly the same scaffold?

camilamacedo86: Consumers/plugins do customs on the scaffold files. Usually, the Makefile will require new targets and etc.. So, changes in this file have a good chance to bring impact. 

Regarding specifically the effort to fix the breaking changes on SDK side, in this case, does not matter too much if the changes are applied only for v2 or v2 and v3. SDK has been using go/v3 already. So, it will probably affect anyway the "plugins" which does custom on the Makefile, the logic to generate the samples, the tests and docs. (for v2 and v3) 

Also, this PR does not only adds help or improve its text as described. For example,  the following change is not required to add help for example: 

```
# Backwards compatibility
manager: build ## Build manager binary (deprecated, use build target instead).
```

However, if we are ok in changing the Makefile for v2, has any reason for v2 and v3 does not do exactly the same scaffold less the test target using the controller-runtime?  It might reduce a few complexities.

Adirio: About the differences from v2/v3 there are 4:
1. v3 added the `preserveUnkownFields=false` option to controller-gen CRD generation. I think this is only needed for API and Webhook versions `v1` so that is the reason v2 doesn't have it as it is only supported in v3.
2. v3 the test target was changed to use controller-runtime's installation process.
3. v2 still has a `manager: build` target that acts as an alias so that if someone was using `make manager` they can still use it. For example they may have some automatic process that runs it and would break it if we don't keep the backwards compatibility.
4. `controller-gen` and `kustomize` installation scripts were modified for v3.

The rest of the files are now completely identical. They had more differences before.

About adding new targets, SDK could do two things:
1. Append `##@ SDK` as the end and add its targets under this new category.
2. Use the currently existing categories by finding `##@ <CATEGORY>` and add its targets below.

About the P.S.: I did addd that info in the first message, will add it to the title too.

P.S.2: while doing the diff between the makefiles of the 2 versions I found an space at the end of a description in v3, that's why I forcepushed to remove it.

estroz: My opinion: adding the `help` target to go/v2's Makefile is fine, since it doesn't actually change any documented features/code, but changing `build` to `manager` is not because existing go/v2 projects won't be able to follow the same docs -> breaking change.

Adirio: @estroz that is why I left `manager: build` alias target with a help message that says that `manager` is deprecated.

Line 111: https://github.com/kubernetes-sigs/kubebuilder/pull/1981/files#diff-289e61cdb2aae0c8de547a662a2592227f5f77d7628850ffe609e922db923606R111

estroz: It feels strange to scaffold a deprecated make target on `init`. As a user I could see being confused by usage despite the comments. How about adding `help` without aliasing `manager` as `build` for go/v2?

Adirio: The deprecation is just a note in the help message. Does it feel that strange?

camilamacedo86: why not remove the deprecated note and keep is v3? shows that technically we would not introducing a breaking change at all.

Adirio: Changed the note to remove the deprecation part and just leave it as an alias for build to keep backwards compatibility.

estroz (approved):
/approve

estroz:
@camilamacedo86 waiting for your lgtm

camilamacedo86 (approved):
/lgtm

Issue #1982, opened 2021-01-29
Stop running make after every command call


Adirio:
### Post-scaffold `make`
Currently, every `init`, `create api` or `create webhook` call `make` as the last step. There are some cases where this behavior can be disabled with a flag (`v2/create-api`, `v3/create-api` & `v3/create-webhook`) and others can't even be disabled (`v2/init`, `v3/init` and `v3/create-webhook`).

### `make` functionality
A call to make does four things:
1. Create `DeepCopy`, `DeepCopyTo`, and `DeepCopyObject` methods for your types.
2. `go fmt`
3. `go vet`
4. Build the manager

## Subcommands functionality evaluation
I will evaluate the need for those four steps for each of the commands.

### `init`
1. No type created yet, no-op.
2. Only scaffolded code, no-op*.
3. Only scaffolded code, no-op*.
4. No types nor controllers, useless manager built.

1, 2 and 3 are no-ops, and the manager built by 4 is completely useless. There is no reason to `make` after `init`.

### `create api`

1. If you created a type you need to update it, so you will need to regenerate this afterwards. If you didn't create a type, no use on regenerating the methods.
2. Only scaffolded code, no-op*.
3. Only scaffolded code, no-op*.
4. You just created a type/controller, you will need to code parts (type spec or reconcile loop), so you will need to rebuild afterwards.

1 and 4 will have to be done after editing and 2 and 3 are no-ops. The only reason I see to do this is to have the `DeepCopy`, `DeepCopyTo` and `DeepCopyObject` methods recognized by IDEs while you are writting the code, which would only require `make generate` and only in the case a new API was scaffolded.

### `create webhook`

1. You didn't create a type, no use on regenerating the methods.
2. Only scaffolded code, no-op*.
3. Only scaffolded code, no-op*.
4. You just created a webhook, you will need to code parts (interface methods), so you will need to rebuild afterwards.

1 has no use, 2 and 3 are no-ops, and 4 will have to be done after editing. There is no reason to `make` after `init`.

* If steps 2 or 3 are not no-ops, it means that we are scaffolding things with style errors which should be fixed in the plugins and having this post-scaffold `go fmt` and `go vet` will just hide those style errors.

## Conclusions
[...]

PR #1983, opened 2021-01-29
✨ Remove make calls from post-scaffold hooks

Adirio:
Remove non-needed calls to `make` at the end of different command calls.

Closes: #1982

### DIFF
--- a/pkg/plugins/golang/v2/api.go
+++ b/pkg/plugins/golang/v2/api.go
@@ -91,7 +91,7 @@ After the scaffold is written, api will run make on the project.
 }
 
 func (p *createAPISubcommand) BindFlags(fs *pflag.FlagSet) {
-	fs.BoolVar(&p.runMake, "make", true, "if true, run make after generating files")
+	fs.BoolVar(&p.runMake, "make", true, "if true, run `make generate` after generating files")
 
 	if os.Getenv("KUBEBUILDER_ENABLE_PLUGINS") != "" {
 		fs.StringVar(&p.pattern, "pattern", "",
@@ -198,8 +198,8 @@ func (p *createAPISubcommand) PostScaffold() error {
 		return fmt.Errorf("unknown pattern %q", p.pattern)
 	}
 
-	if p.runMake {
-		return util.RunCmd("Running make", "make")
+	if p.runMake { // TODO: check if API was scaffolded
+		return util.RunCmd("Running make", "make", "generate")
 	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


pkg/plugins/golang/v2/api.go
```diff
@@ -198,8 +198,8 @@ func (p *createAPISubcommand) PostScaffold() error {
 		return fmt.Errorf("unknown pattern %q", p.pattern)
 	}
 
-	if p.runMake {
-		return util.RunCmd("Running make", "make")
+	if p.runMake { // TODO: check if API was scaffolded
+		return util.RunCmd("Running make", "make", "generate")
```
camilamacedo86, line 202: Should we not run make manifests as well by default?

camilamacedo86 (commented):
IMO I do not think that we should move with this one. see: https://github.com/kubernetes-sigs/kubebuilder/issues/1982#issuecomment-770029211

Also, the changes here a breaking changes fo v2 too :-( . 

/hold

pkg/plugins/golang/v2/api.go
```diff
@@ -91,7 +91,7 @@ After the scaffold is written, api will run make on the project.
 }
 
 func (p *createAPISubcommand) BindFlags(fs *pflag.FlagSet) {
-	fs.BoolVar(&p.runMake, "make", true, "if true, run make after generating files")
+	fs.BoolVar(&p.runMake, "make", true, "if true, run `make generate` after generating files")
```
camilamacedo86, line 94: It is breaking change for v2. So, we should not tio do that.

DirectXMan12: Can you elaborate on your thoughts here?  At first glance, it doesn't actually quite seem breaking to me because it doesn't actually change the scaffolding at all, but I could be convinced otherwise.

pkg/plugins/golang/v2/api.go
Adirio: Make manifest requires the user to first edit the `*_types.go` file and rerun it, and you get no advantage of running it earlier. `make generate` advantage is minimal but I kept it.

camilamacedo86: cacher.

pkg/plugins/golang/v2/api.go
camilamacedo86: I agree that it is not changing the scaffold and its a nit. However, would be impacted if a user adds custom on the make:all to be executed in this scenario or a consumer of go/v2 via customizations (which is not the case of sdk).

However, yes agree. It is to nit scenarios to hold this one. 

/hold cancel

camilamacedo86:
Based on the clarifications #1982 it shows fine and has no reason for us not:
/lgtm
/approve

PR #1984, opened 2021-01-29
:seedling: sort plugins key for the help

camilamacedo86:
**Description**
Sorted plugins key for the help

**Motivation**
Allow SDK to be aligned with KB. Because the result is not idempotent it brokes the CI/doc generation.


christopherhein:
Nice job!
/lgtm

PR #1986, opened 2021-02-02
✨ Constructor for Resource

Adirio:
Provide a constructor for `Resource` in order to better split the model from the command flag options for each plugin so that third-party can consume this API instead of creating an `Options` object.

### DIFF
--- a/pkg/model/resource/constructor.go
+++ b/pkg/model/resource/constructor.go
@@ -1,0 +1,143 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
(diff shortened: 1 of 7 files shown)


Adirio:
/hold

Waiting until #1974 is merged so that validation can also be done in the constructor.

Adirio:
/hold cancel

#1974 was merged

camilamacedo86 (approved):
it shows fine for me too.

Adirio:
With plugin phase 1.5, it has become apparent that resources should be created by the CLI and injected into the different plugins so that they all refer to the same object and they can make their own modifications. This fact greatly reduces the enhancement that this PR offers, as the creation would not use any of the provided options and later modifications can't be done using options. Closing this PR, if plugin phase 1.5 approach is not followed, it may be reopened.

PR #1987, opened 2021-02-03
🌱 Go fmt templates

Adirio:
### Description
Verify that the scaffolded files are already properly formatted before running `go fmt`.


camilamacedo86 (approved):
/lgtm

PR #1989, opened 2021-02-03
📖 Add documentation to v2 controllers

Adirio:
### Description
Add the documentation that was added to v3 plugin's controllers also to v2 plugin as they are only comments. This makes this change non-breaking because controllers scaffolded with v2 before this change will behave exactly the same as after this, but now they will have some extra documentation comments.


camilamacedo86 (approved):
/lgtm

PR #1990, opened 2021-02-04
📖 Plugin phase 1.5 EP

Adirio:
This PR proposes a plugin phase 1.5 by submiting an EP.
Phase 1.5 should provide some of the goals defined for phase 2 in a shorter date.

You can read the EP with a nicer UX that in the chenged files [here](https://github.com/Adirio/kubebuilder/blob/plugin-phase-1-5/designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md).

### DIFF
--- a/designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md
+++ b/designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md
@@ -1,0 +1,258 @@
+# Extensible CLI and Scaffolding Plugins - Phase 1.5
+
+Continuation of [Extensible CLI and Scaffolding Plugins](./extensible-cli-and-scaffolding-plugins-phase-1.md).
+
+## Goal
+
+The goal of this phase is to achieve one of the goals proposed for Phase 2: chaining plugins.
+Phase 2 includes several other challenging goals, but being able to chain plugins will be beneficial
+for third-party developers that are using kubebuilder as a library.
+
+## Table of contents
+- [Goal](#goal)
+- [Motivation](#motivation)
+- [Proposal](#proposal)
+- [Implementation](#implementation)
+
+## Motivation
+
+There are several cases of plugins that want to maintain most of the go plugin functionality and add
+certain features on top of it, both inside and outside kubebuilder repository:
+- [Addon pattern](../plugins/addon)
+- [Operator SDK](https://github.com/operator-framework/operator-sdk/tree/master/internal/plugins/golang)
+
+This behavior fits perfectly under Phase 1.5, where plugins could be chained. However, as this feature is
+not available, the adopted temporal solution is to wrap the base go plugin and perform additional actions
+after its `Run` method has been executed. This solution faces several issues:
+
+- Wrapper plugins are unable to access the data of the wrapped plugins, as they weren't designed for this
+  purpose, and therefore, most of its internal data is non-exported. An example of this inaccessible data
+  would be the `Resource` objects created inside the `create api` and `create webhook` commands.
+- Wrapper plugins are dependent on their wrapped plugins, and therefore can't be used for other plugins.
+- Under the hood, subcommands implement a second hidden interface: `RunOptions`, which further accentuates
+  these issues.
+
+Plugin chaining solves the aforementioned problems but the current plugin API, and more specifically the
+`Subcommand` interface, does not support plugin chaining.
+
+- The `RunOptions` interface implemented under the hood is not part of the plugin API, and therefore
+  the cli is not able to run post-scaffold logic (implemented in `RunOptions.PostScaffold` method) after
+  all the plugins have scaffolded their part.
+- `Resource`-related commands can't bind flags like `--group`, `--version` or `--kind` in each plugin,
+  it must be created outside the plugins and then injected into them similar to the approach followed
+  currently for `Config` objects.
+
+## Proposal
+
+Design a Plugin API that combines the current [`Subcommand`](../pkg/plugin/interfaces.go) and
+[`RunOptions`](../pkg/plugins/internal/cmdutil/cmdutil.go) interfaces and enables plugin-chaining.
+The new `Subcommand` methods can be split in two different categories:
+- Initialization methods
+- Execution methods
+
+Additionally, some of these methods may be optional, in which case a non-implemented method will be skipped
+when it should be called and consider it succeeded. This also allows to create some methods specific for
+a certain subcommand call (e.g.: `Resource`-related methods for the `edit` subcommand are not needed).
+
+Different ordering guarantees can be considered:
+- Method order guarantee: a method for a plugin will be called after its previous methods succeeded.
+- Steps order guarantee: methods will be called when all plugins have finished the previous method.
+- Plugin order guarantee: same method for each plugin will be called in the order specified
+  by the plugin position at the plugin chain.
+
+All of the methods will offer plugin order guarantee, as they all modify/update some item so the order
+of plugins is important. Execution methods need to guarantee step order, as the items that are being modified
+in each step (config, resource, and filesystem) are also needed in the following steps. This is not true for
+initialization methods that modify items (metadata and flagset) that are only used in their own methods,
+so they only need to guarantee method order.
+
+Execution methods will be able to return an error. A specific error can be returned to specify that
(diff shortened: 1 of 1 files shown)


designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md
```diff
@@ -0,0 +1,286 @@
...
+`Subcommand` interface, does not support plugin chaining.
+
+- The `RunOptions` interface implemented under the hood is not part of the plugin API, and therefore
+  the cli is not able to run post-scaffold logic (implemented in `RunOptions.PostScaffold` method) after
+  all the plugins have scaffolded their part.
+- `Resource`-related commands can't bind flags like `--group`, `--version` or `--kind` in each plugin,
+  it must be created outside the plugins and then injected into them similar to the approach followed
+  currently for `Config` objects.
+
+<a id="proposal" />
```
DirectXMan12, line 49: FYI, you'll get autolinking from the heading titles, so you shouldn't need to add these in manually unless you don't want to have to worry about rewording the titles.

designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md
```diff
@@ -0,0 +1,286 @@
...
+  - [x] Create API
+  - [x] Create webhook
+- Order guarantees
+  - [x] Method order guarantee.
+  - [x] Step order guarantee.
+  - [x] Plugin order guarantee.
+
+#### Post-scaffold
+This method will be used to take actions after the main scaffolding is performed, e.g. cleanup.
+NOTE: a filesystem abstraction will **NOT** be passed to this method, it should interact directly.
```
DirectXMan12, line 183: is the recommendation that you not interact with the filesystem here then, or is this just because naturally this takes place after KB writes stuff to disk, or both?

DirectXMan12 (approved):
seems pretty reasonable to me.  Might want to elaborate on which each step gets each guarantee for posterity's sake, but generally looks good :-)

designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md
Adirio: As that is not markdown itself but GitHub markdown, I wasn't sure if it added them or not.

designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md
Adirio: It was because the scaffolding is expected to be done in the previous step, post-scaffold was designed to do tasks like code generation, make targets and things like that that do not require a filesystem. Even though the current design and implementation of plugin phase 1.5 do not pass a filesystem, maybe we want to leave the option to pass a filesystem in the future if the need arises.

Adirio: I updated the note saying that we may need to pass a filesystenm abstraction in a future. It is more a "We don't need it now" that a "We should not have a filesystem abstraction here".

designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md
```diff
@@ -0,0 +1,258 @@
...
+
+type RequiresConfig interface {
+	InjectConfig(config.Config) error
+}
+
+type RequiresResource interface {
+	InjectResource(*resource.Resource) error
+}
+
+type HasPreScaffold interface {
```
estroz, line 222: I don't see an immediate need for pre and post scaffolding, since these can be done within `Scaffold()`. Perhaps remove these to simplify things, and add them back if we find a valid use case for them?

estroz (approved):
Looks great, just one question/suggestion.

estroz:
Remove this hold when ready to merge.

/lgtm
/hold

designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md
Adirio: They are optional, so there is no need to use them if you want to develop a simple plugin (you can check the edit subcommands in the implementation PR), but they do offer a great flexibility. All plugins will run their pre-scaffold logic and can return an error (stoping the subcomman execution) before any of the plugins start to scaffold the files. This allows us to validate all the plugins before scaffolding. If we only had one hook (Scaffold) then every plugin will run its validation, scaffold its files and do the post-scaffolding tasks before the rest of the plugin can validate theirselves. Same applies for post-scaffolding tasks.

Let me illustrate this with a couple examples:
1. We have a plugin that allows us to define the fields a type will have in some yaml format that must be present before we run `kb create api`. The cli has the base go/v3 and this plugins configured for this project. One of the validations of this plugins is to check that the file is present and that it can be parsed. If we only have one hook, it will check this conditions after go/v3 has scaffolded its files, when it shouldn't because it failed to fulfill the second plugin requirements.
2. We have a plugin that adds some fields to the generated types. Again, the cli has the base go/v3 plugin and this one that edits its fields. go/v3 runs `make generate` at the end, but if we don't have a post-scaffold hook, it will be run before the second plugin performs the changes to the types file, and therefore be obsolete by the time the command finishes.

estroz: Thanks for the explanation, that makes sense.

designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md
```diff
@@ -0,0 +1,258 @@
...
+after its `Run` method has been executed. This solution faces several issues:
+
+- Wrapper plugins are unable to access the data of the wrapped plugins, as they weren't designed for this
+  purpose, and therefore, most of its internal data is non-exported. An example of this inaccessible data
+  would be the `Resource` objects created inside the `create api` and `create webhook` commands.
+- Wrapper plugins are dependent on their wrapped plugins, and therefore can't be used for other plugins.
+- Under the hood, subcommands implement a second hidden interface: `RunOptions`, which further accentuates
+  these issues.
+
+Plugin chaining solves the aforemention problems but the current plugin API, and more specifically the
```
jmrodri, line 35:
```suggestion
Plugin chaining solves the aforementioned problems but the current plugin API, and more specifically the
```

designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Proposal
+
+Design a Plugin API that combines the current [`Subcommand`](../pkg/plugin/interfaces.go) and
+[`RunOptions`](../pkg/plugins/internal/cmdutil/cmdutil.go) interfaces and enables plugin-chaining.
+The new `Subcommand` methods can be split in two different categories:
+- Initialization methods
+- Execution methods
+
+Additionally, some of this methods may be optional, in which case a non-implemented method will be skipped
```
jmrodri, line 53:
```suggestion
Additionally, some of these methods may be optional, in which case a non-implemented method will be skipped
```

designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md
```diff
@@ -0,0 +1,258 @@
...
+
+All of the methods will offer plugin order guarantee, as they all modify/update some item so the order
+of plugins is important. Execution methods need to guarantee step order, as the items that are being modified
+in each step (config, resource, and filesystem) are also needed in the following steps. This is not true for
+initialization methods that modify items (metadata and flagset) that are only used in their own methods,
+so they only need to guarantee method order.
+
+Execution methods will be able to return an error. A specific error can be returned to specify that
+no further methods of this plugin should be called, but that the scaffold process should be continued.
+This enables plugins to exit early, e.g., a plugin that scaffolds some files only for cluster-scoped
```
jmrodri, line 71: The exit early but continue is nice.

designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md
jmrodri: answers my question as well.

jmrodri (approved):
/lgtm

Adirio:
/hold
Unitl I resolve the raised comments

Issue #1941, opened 2021-01-13
Allow every command to overwrite project-configured plugins


Adirio:
When resolivng which plugins will handle each command, we currently have 3 sources for the project version and plugin keys: project configuration file, flags or default values.

Init commands will not have a project configuration file, so the precedence is clear, flags override default values. However, the rest of the commands require a project configuration so they have 2 sources that are at the same level, flags and project configuration, both overriding the default values. If both flags and project configuration are present, the current approach is to fail in case they are different. This makes total sense for the project version, as you can't run a create command for project version 3 on a project version , obviously. But in the case of plugins, we may want to allow them to be different.

Let's say that we create a plugin that modifies the `controller.go` file to insert inside the reconcile function some scaffold code and/or that creates a `controller_test.go` scaffold. We may want to use this plugin for some of the controllers generated in this project but not for all of them.

I would allow command calls' `--plugin` flag to temporaly override the `layout` field set in the project configuration file. And by temporaly override I mean that this command call will be called with the values of the flag but the project configuration `layout` field will not be modified, and therefore, succesive command calls will still use the plugins configured during project initialization.

camilamacedo86:
Hi @Adirio, 

Could you confirm our understanding? if we scaffold with A and then we use the flag propose here to use B then, would we able to overwrite A? How we will track what action was performed to config the project? Could you please provide more info/uses cases that might it required?

camilamacedo86:
in the EP you added the following example: 


```
kubebuilder init --plugins=go.kubebuilder.io/v3
```

```
kuebebuilder create api --plugins=go.kubebuilder.io/v3,extra.kubebuilder.io/v1 ...
```

So, I understand that it allow executing the create api of `go.kubebuilder.io/v3` and then the same subcommand of `extra.kubebuilder.io/v1`. Am I right? then, I have a few questions:

1 - Why we need to re-inform `go.kubebuilder.io/v3` since it is the project layout?
2 - Also, how you will check that all inputs required for A and B were provided before to do the execution? 
3-  What is the difference between your suggestion with phase 2 plugin? I mean, the phase 2 would allow we to call more than one plugin in the same execution. So, what is the motivation of this EP?
4- Could you please add some [user stories](https://en.wikipedia.org/wiki/User_story) to illustrate the problems that you are trying to solve with? ( template `As a <type of user>, I can <some goal> so that <some reason>`

Adirio:
[...]

PR #1991, opened 2021-02-04
⚠️  Plugin chaining

Adirio:
Plugin phase 1.5

Implements #1990
Closes: #2036


generate_testdata.sh
```diff
@@ -129,9 +126,9 @@ build_kb
 # Project version 2 uses plugin go/v2 (default).
 scaffold_test_project project-v2 --project-version=2
 scaffold_test_project project-v2-multigroup --project-version=2
-scaffold_test_project project-v2-addon --project-version=2
+scaffold_test_project project-v2-addon --project-version=3 --plugins="go/v2,declarative"
```
camilamacedo86, line 129: I do not think that we should need to inform  `go/v2` it is the layout f the project already. So, here we would inform just the declarative one.
Also, should we not keep the same name (addon)? have we any doc about it? If yes, we need a top update as well

testdata/project-v3-addon/PROJECT
```diff
@@ -1,5 +1,5 @@
 domain: testproject.org
-layout: go.kubebuilder.io/v3
+layout: go.kubebuilder.io/v3,declarative.kubebuilder.io/v1
```
camilamacedo86, line 2: The layout of the project is `go.kubebuilder.io/v3`. The `declarative.kubebuilder.io/v1` is an additional plugin used on top of that and should be tracked in the plugins attr instead: 

```
plugins:
  declarative.kubebuilder.io/v1: {}
```

camilamacedo86 (changes requested):
I am ok with that. However, has few points that I do not agree with here:

**1 - Project layout** 

It represents the layout used to scaffold the project. 
Then, I do not agree with:

```
layout: go.kubebuilder.io/v3,declarative.kubebuilder.io/v1
```

The `declarative.kubebuilder.io/v1` plugin is an extra plugin used to do a scaffold on top of the specific subcommand create api of `golang` plugin, the default project layout.

Then in POV, it should be tracked as:

```
plugins:
  declarative.kubebuilder.io/v1: {}
```

Also, I do not think that we should be able/allowed to change/overwrite the default layout. It can be messy and will not allow us to know what was the `base` used to create the project. 

So, the `layout` attr in the config would only allow we have more than one plugin if/when it was used in the init command ( which has the responsibility to create the project base). E.g if we create a project (init) with `go/v2,helm/v1` it would scaffold both bases. And then, for this scenario, we need to describe how we will solve many issues such as: what info will be described in the help? what args will be allowed? I mean, if we scaffold a base `go/v2,helm/v1` will it show the args to create an api valid for the go or helm plugin? Or both? 

**2 - We need to check and update any documentation over the addon as well.** 

c/c @DirectXMan12 @estroz 
/hold

Adirio:
> **1 - Project layout**
> 
> It represents the layout used to scaffold the project.
> Then, I do not agree with:
> 
> ```
> layout: go.kubebuilder.io/v3,declarative.kubebuilder.io/v1
> ```
> 
> The `declarative.kubebuilder.io/v1` plugin is an extra plugin used to do a scaffold on top of the specific subcommand create api of `golang` plugin, the default project layout.
> 
> Then in POV, it should be tracked as:
> 
> ```
> plugins:
>   declarative.kubebuilder.io/v1: {}
> ```
> 
> Also, I do not think that we should be able/allowed to change/overwrite the default layout. It can be messy and will not allow us to know what was the `base` used to create the project.
> 
> So, the `layout` attr in the config would only allow we have more than one plugin if/when it was used in the init command ( which has the responsibility to create the project base). E.g if we create a project (init) with `go/v2,helm/v1` it would scaffold both bases. And then, for this scenario, we need to describe how we will solve many issues such as: what info will be described in the help? what args will be allowed? I mean, if we scaffold a base `go/v2,helm/v1` will it show the args to create an api valid for the go or helm plugin? Or both?

You are mixing this with the choice of providing a different `--plugin` flag per command we talked in the triage meeting. That project is scaffolded as so:
```
kubebuilder init --project-version=3 --plugins="go/v2,declarative" --domain testproject.org --license apache2 --owner "The Kubernetes authors"
kubebuilder create api --group crew --version v1 --kind Captain --controller=true --resource=true --make=false
kubebuilder create api --group crew --version v1 --kind FirstMate --controller=true --resource=true --make=false
kubebuilder create api --group crew --version v1 --kind Admiral --controller=true --resource=true --namespaced=false --make=false
```
And therefore the layout is correct.

> **2 - We need to check and update any documentation over the addon as well.**

There is no documentation regarding that.

testdata/project-v3-addon/PROJECT
Adirio: No, `declarative/v1` is configured for the whole project because there is no way to configure a plkugin for only one command call. That is a future feature tracked in #1941.

generate_testdata.sh
Adirio: This again is mixing ideas from #1941 and makes no sense at all in the context of this PR.

testdata/project-v2-addon/PROJECT
```diff
@@ -1,13 +1,32 @@
...
+- api:
+    crdVersion: v1beta1
+  controller: true
+  domain: testproject.org
+  group: crew
   kind: Admiral
+  path: sigs.k8s.io/kubebuilder/testdata/project-v2-addon/api/v1
   version: v1
-version: "2"
+version: "3"
```
camilamacedo86, line 32: Why it changes the config? Should it really change it?

generate_testdata.sh
```diff
@@ -105,13 +105,10 @@ scaffold_test_project() {
       $kb create webhook --version v1 --kind Lakers --defaulting --programmatic-validation
     fi
   elif [[ $project =~ addon ]]; then
-    header_text 'enabling --pattern flag ...'
-    export KUBEBUILDER_ENABLE_PLUGINS=1
```
camilamacedo86, line 109: I am ok with that. However, has few points that I do not agree with here:

**1 - Project layout** 

It represents the layout used to scaffold the project. 
Then, I do not agree with:

```
layout: go.kubebuilder.io/v3,declarative.kubebuilder.io/v1
```

The `declarative.kubebuilder.io/v1` plugin is an extra plugin used to do a scaffold on top of the specific subcommand create api of `golang` plugin, the default project layout. 

Currently, we can decide what APIs will use or not the addon ($ kubebuilder create api --group crew --version v1 --kind Captain --controller=true --resource=true --pattern=addon) and it is scaffolds are not part of the init/base at all. So, I do not agree that we should call this plugin in the init. 

IMO: We ought to
- we can call the plugin in the create api  which means replace `--pattern=addon` by `--plugins=declarative.kubebuilder.io/v1` for the current command `($ kubebuilder create api --group crew --version v1 --kind Captain --controller=true --resource=true --pattern=addon)`
- then, it should be tracked as:

```
plugins:
  declarative.kubebuilder.io/v1: {}
```

Also, I do not think that we should be able/allowed to change/overwrite the default layout. It can be messy and will not allow us to know what was the `base` used to create the project before we define how we will: 
-  described the help
-  define what args will be allowed

I mean, if we scaffold a base `go/v2,helm/v1` will it show the args to create an api valid for the go or helm plugin? Or both? 

Also, the init command should allow we use only plugins which are `base` ones. I mean, we should not able to init a project with a plugin that is not responsible to scaffold the "core/base" files. 

**2 - We need to check and update any documentation over the addon as well.** 
/hold

camilamacedo86: We discussed it a lot today. I will close this comment and add a summary in the review.

camilamacedo86 (commented):
@Adirio and I discussed the plugin ecosystem a lot today. Then, a summary of all discussed subjects is tracked in https://github.com/kubernetes-sigs/kubebuilder/issues/2016 which indeed can be helpful for we have a better idea and overview the plugin ecosystem and all.

So, I am ok with we move forward since the current behaviour for the addon would be addressed via  #1941 where we would allow users to choose what API would or not be scaffolded with. 

/hold cancel

testdata/project-v2-addon/PROJECT
Adirio: Project config 2 doesn't have the layout field required for this change. It may not be needed when we allow per-command `--plugin` flags as we will be able to use `kb init --plugins="go/v2" ...` and add the declarative plugin in `kb create api` calls, not requiring the layout field in the project file any more.

pkg/plugins/golang/options.go
```diff
@@ -64,17 +54,7 @@ var (
...
 type Options struct {
-	// Group is the resource's group. Does not contain the domain.
-	Group string
-	// Domain is the resource's domain.
-	Domain string
-	// Version is the resource's version.
-	Version string
-	// Kind is the resource's kind.
-	Kind string
-
```
camilamacedo86, line 75: These changes are breaking changes for the lib. Why do we need to change the options to address this need?

pkg/plugins/golang/v3/edit.go
```diff
@@ -19,56 +19,47 @@ package v3
...
+	- Toggle between single or multi group projects.
+`
+	subcmdMeta.Examples = fmt.Sprintf(`# Enable the multigroup layout
+        %[1]s edit --multigroup
 
         # Disable the multigroup layout
-        %s edit --multigroup=false
-	`, ctx.CommandName, ctx.CommandName)
+        %[1]s edit --multigroup=false
+	`, cliMeta.CommandName)
```
camilamacedo86, line 48: It is also a breaking change.

pkg/cli/api.go
```diff
@@ -14,80 +14,58 @@ See the License for the specific language governing permissions and
...
 func (c cli) newCreateAPICmd() *cobra.Command {
-	ctx := c.newAPIContext()
 	cmd := &cobra.Command{
-		Use:     "api",
-		Short:   "Scaffold a Kubernetes API",
-		Long:    ctx.Description,
-		Example: ctx.Examples,
+		Use:   "api",
+		Short: "Scaffold a Kubernetes API",
+		Long: `Scaffold a Kubernetes API.
```
camilamacedo86, line 33: We can overwrite the description currently. Will still it possible?

Makefile
```diff
@@ -89,7 +89,7 @@ golangci-lint:
 
 .PHONY: go-test
 go-test: ## Run the unit test
```
camilamacedo86, line 91: this pr has many breaking changes; the mojo needs to be updated as the pr title for a more meaningful description. e.g add support to 1..N default plugins via the layout config

generate_testdata.sh
```diff
@@ -129,9 +126,9 @@ build_kb
 # Project version 2 uses plugin go/v2 (default).
 scaffold_test_project project-v2 --project-version=2
 scaffold_test_project project-v2-multigroup --project-version=2
-scaffold_test_project project-v2-addon --project-version=2
```
camilamacedo86, line 132: the feature is experimental in kb so I think would be fine.

pkg/plugins/internal/machinery/scaffold_test.go
```diff
@@ -466,18 +409,6 @@ var _ = Describe("Scaffold", func() {
 	})
 })
 
-var _ model.Plugin = fakePlugin{}
```
camilamacedo86, line 469: why we are removing the plugins?

pkg/plugins/golang/v3/edit.go
Adirio: Why is this a breaking change?

pkg/plugins/golang/options.go
Adirio: These flags need to be handled by the cli because every plugin that creates an api will need the GVK, thats why the flags need to be binded by the cli and not by a certain plugin or all plugins will be incompatible as they will all have these flags.

pkg/cli/api.go
Adirio: Yes, the `UpdateContext` function was renamed to `UpdateMetadata` (because context was conflicting with Ginkgo) and the signature modified a bit. Instead of having a single struct with input and output fields, now it has one with input fields (currently only the command name) and another one with output fields (currently examples and description).

pkg/plugins/internal/machinery/scaffold_test.go
Adirio: This is part of the old experimental plugin system used for addon.

pkg/plugins/golang/v3/edit.go
camilamacedo86: It affects the consumer. You are changing the exported methods in many places.

Makefile
camilamacedo86: Also, it will break a lot the integration with SDK which is fine, we just need to ensure that all will still compatible. c/c @estroz

christopherhein:
Nice work @Adirio this seems like it will really help to further the componentconfig implementation. We currently have a hacky way to extend the ComponentConfig type that requires a bunch of manual file deletions and removing the kubebuilder marker for crd generation. Currently for a custom config type in KB we have docs - https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/component-config-tutorial/config-type.md#adding-a-new-config-type that explain how to do it manually and this would help to automate that setup and allow custom configs to be easier to develop.

camilamacedo86 (changes requested):
The scenario https://github.com/operator-framework/operator-sdk/pull/4581#discussion_r582216189 needs also be solved before we get it merged.

pkg/cli/cmd_helpers.go
```diff
@@ -17,45 +17,256 @@ limitations under the License.
...
+type noResolvedPluginError struct{}
+
+// Error implements error interface.
+func (e noResolvedPluginError) Error() string {
+	return "no resolved plugin, please verify the project version and plugins specified in flags or configuration file"
+}
+
+// noAvailablePluginError is returned by subcommands that require a plugin when none of their specific type was found.
+type noAvailablePluginError struct {
+	subcommand string
```
camilamacedo86, line 43: imo we ought to be a warning to allow users that used SDK tool to scaffold the project to still able to use KB tool.

Adirio:
After rebasing when #2037 got merged, `go mod tidy` is bumping the dependencies in the declarative projects when it should not. The only difference from non-declarative projects is relation to modules is the import of `kubernetes-sigs/kubebuilder-declarative-pattern`. Any way to prevent this in combination with `go mod tidy` to bump all [those](https://github.com/kubernetes-sigs/kubebuilder/pull/1991/checks?check_run_id=2008582058) dependencies?

pkg/cli/cmd_helpers.go
Adirio: This is considered in the follow up #2019 though the `Bundle` concept. Not being able to resolve a plugin definetely needs to be an error or we will make the users get to broken states.

Adirio:
> After rebasing when #2037 got merged, `go mod tidy` is bumping the dependencies in the declarative projects when it should not. The only difference from non-declarative projects is relation to modules is the import of `kubernetes-sigs/kubebuilder-declarative-pattern`. Any way to prevent this in combination with `go mod tidy` to bump all [those](https://github.com/kubernetes-sigs/kubebuilder/pull/1991/checks?check_run_id=2008582058) dependencies?

I discovered the cause. `go` post-scaffold method was running `go mod tidy` and `declarative` post-scaffold method was running `go get kubernetes-sigs/kubebuilder-declarative-pattern@...`. As the `go mod tidy` was being run before `go get ...` it was importing the latest version of `kubebuilder-declarative-pattern` which bumped a bunch of versions. Later `go get ...` was defining the exact declarative pattern version but the damage was already done. Changing `go get ...` calls to the end of scaffold hooks instead of post-scaffold guarantees that they will all be run before `go mod tidy`.

Adirio:
/close

Closing this in favor of the chain of PRs #2057 - #2058 - #2059 that split this PR into several PRs to split concerns and make reviewer job easier.

PR #1992, opened 2021-02-04
:seedling: generalize tools Dockerfiles for arm64, ppc64le linux builds

### LINKED ISSUE #966: Publish packages for multi-arch
This issue to publish kubebuilder binaries for all the possible architectures like arm, ppc, s390x etc..

/kind feature

estroz:
- */Dockerfile: generalize for multiple platforms
- build/cloudbuild_tools.yaml: update with new Dockerfile args, prettify
- README.md: add new arches

**Note:** this PR also downgrades k8s version to v1.19.2 so arm64 and ppc64le builds occur for a v1.19 version. I will post a follow-up PR to upgrade the version back to v1.20.2 after this is merged.

Closes #966 

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ GCP Cloud Build watches this branch.  On every push, it runs the pipeline define
 
 - `_GOARCH=amd64 _GOOS=darwin`
 - `_GOARCH=amd64 _GOOS=linux`
+- `_GOARCH=arm64 _GOOS=linux`
+- `_GOARCH=ppc64le _GOOS=linux`
 
 (we may add more the in the future).
 
--- a/build/cloudbuild_tools.yaml
+++ b/build/cloudbuild_tools.yaml
@@ -12,22 +12,36 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+substitutions:
+  _KUBERNETES_VERSION: v1.19.2
 steps:
 - name: "gcr.io/cloud-builders/docker"
-  args: ["build", "-t", "gcr.io/kubebuilder/thirdparty-${_GOOS}:1.20.2", "./build/thirdparty/${_GOOS}"]
+  args: [
+    "build",
+    "--build-arg", "OS=${_GOOS}",
+    "--build-arg", "ARCH=${_GOARCH}",
+    "--build-arg", "KUBERNETES_VERSION=${_KUBERNETES_VERSION}",
+    "-t", "gcr.io/kubebuilder/thirdparty-${_GOOS}-${_GOARCH}:${_KUBERNETES_VERSION}",
+    "./build/thirdparty/${_GOOS}",
+  ]
   # darwin takes forever
   timeout: 30m
 
-- name: "gcr.io/kubebuilder/thirdparty-${_GOOS}:1.20.2"
-  args: ["cp", "/kubebuilder_${_GOOS}_${_GOARCH}.tar.gz", "/workspace/kubebuilder-1.20.2-${_GOOS}-${_GOARCH}.tar.gz"]
-  env:
-  - 'GOOS=${_GOOS}'
-  - 'GOARCH=${_GOARCH}'
-- name: 'gcr.io/cloud-builders/gsutil'
-  args: ['-h', 'Content-Type:application/gzip', 'cp', '/workspace/kubebuilder-1.20.2-${_GOOS}-${_GOARCH}.tar.gz', 'gs://kubebuilder-tools/kubebuilder-tools-1.20.2-${_GOOS}-${_GOARCH}.tar.gz']
-  env:
-  - 'GOOS=${_GOOS}'
-  - 'GOARCH=${_GOARCH}'
-images: ["gcr.io/kubebuilder/thirdparty-${_GOOS}:1.20.2"]
+- name: "gcr.io/kubebuilder/thirdparty-${_GOOS}-${_GOARCH}:${_KUBERNETES_VERSION}"
+  args: [
+    "cp",
+    "/kubebuilder_${_GOOS}_${_GOARCH}.tar.gz",
+    "/workspace/kubebuilder-tools-${_KUBERNETES_VERSION}-${_GOOS}-${_GOARCH}.tar.gz",
+  ]
+
+- name: "gcr.io/cloud-builders/gsutil"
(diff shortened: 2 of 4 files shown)


build/cloudbuild_tools.yaml
```diff
@@ -12,22 +12,36 @@
...
-  env:
-  - 'GOOS=${_GOOS}'
-  - 'GOARCH=${_GOARCH}'
-- name: 'gcr.io/cloud-builders/gsutil'
-  args: ['-h', 'Content-Type:application/gzip', 'cp', '/workspace/kubebuilder-1.20.2-${_GOOS}-${_GOARCH}.tar.gz', 'gs://kubebuilder-tools/kubebuilder-tools-1.20.2-${_GOOS}-${_GOARCH}.tar.gz']
-  env:
-  - 'GOOS=${_GOOS}'
-  - 'GOARCH=${_GOARCH}'
-images: ["gcr.io/kubebuilder/thirdparty-${_GOOS}:1.20.2"]
+- name: "gcr.io/kubebuilder/thirdparty-${_GOOS}-${_GOARCH}:${_KUBERNETES_VERSION}"
```
DirectXMan12, line 30: should we keep something around that cross-tags this to the old `thirdparty-${OS}` format too, so we don't confuse people?  (this is a question, not a suggestion -- I can't quite convince myself either way)

build/cloudbuild_tools.yaml
```diff
@@ -12,22 +12,36 @@
...
+    "cp",
+    "/kubebuilder_${_GOOS}_${_GOARCH}.tar.gz",
+    "/workspace/kubebuilder-tools-${_KUBERNETES_VERSION}-${_GOOS}-${_GOARCH}.tar.gz",
+  ]
+
+- name: "gcr.io/cloud-builders/gsutil"
+  args: [
+    "-h", "Content-Type:application/gzip",
+    "cp",
+    "/workspace/kubebuilder-tools-${_KUBERNETES_VERSION}-${_GOOS}-${_GOARCH}.tar.gz",
```
DirectXMan12, line 41: ditto here.

In either case, will need to update the `go.kubebuilder.io` shortlinks

build/thirdparty/darwin/Dockerfile
```diff
@@ -13,45 +13,57 @@
...
-WORKDIR /go/src/k8s.io/kubernetes
+# Install dependencies.
+RUN apt update && \
+    apt install unzip rsync -y && \
+    mkdir -p $DEST
 
+# kube-apiserver
+WORKDIR /kubernetes
+RUN git clone https://github.com/kubernetes/kubernetes . --depth=1 -b ${KUBERNETES_VERSION}
+ENV CGO 0
```
DirectXMan12, line 39: is this `CGO` or `CGO_ENABLED`?  I thought it was the latter, but I could be misremembering

build/thirdparty/darwin/Dockerfile
```diff
@@ -13,45 +13,57 @@
...
 # Build for linux first otherwise it won't work for darwin - :(
-ENV KUBE_BUILD_PLATFORMS linux/amd64
+ENV KUBE_BUILD_PLATFORMS=linux/amd64
```
DirectXMan12, line 41: some of these seem to have equals signs, others don't, what's up with that?

DirectXMan12 (changes requested):
idea is generally fine, couple of questions inline

build/thirdparty/darwin/Dockerfile
estroz: I thought `CGO` might’ve been used internally by the k8s Makefile so I left this as-is. Go’s env variable is `CGO_ENABLED`.

build/cloudbuild_tools.yaml
estroz: Is this image intended to be pulled/used outside of this build system? If so then I don’t think the old format should be kept around because it is now ambiguous.

build/thirdparty/darwin/Dockerfile
estroz: They should all have =‘s. Will update those that do not.

build/cloudbuild_tools.yaml
estroz: The tarball name format hasn't changed, and if it has it shouldn't have since it already includes os and arch.

build/cloudbuild_tools.yaml
Adirio: What about a docker manifest? I think it still requires experimental mode on docker in order to create them but nothing special is needed in order to get the images.

build/thirdparty/darwin/Dockerfile
```diff
@@ -13,45 +13,57 @@
...
+    apt install unzip rsync -y && \
+    mkdir -p $DEST
 
+# kube-apiserver
+WORKDIR /kubernetes
+RUN git clone https://github.com/kubernetes/kubernetes . --depth=1 -b ${KUBERNETES_VERSION}
+ENV CGO_ENABLED=0
 # Build for linux first otherwise it won't work for darwin - :(
-ENV KUBE_BUILD_PLATFORMS linux/amd64
+ENV KUBE_BUILD_PLATFORMS=linux/amd64
```
Adirio, line 41: Does this still hold true? Do we need to build for linux first? Could we add a reference so that out future selfs or other members understand why this to be done? Also, does it need to be for that hardcoded arch?

estroz: It doesn't, removed.

README.md
```diff
@@ -15,6 +15,8 @@ GCP Cloud Build watches this branch.  On every push, it runs the pipeline define
 
 - `_GOARCH=amd64 _GOOS=darwin`
 - `_GOARCH=amd64 _GOOS=linux`
+- `_GOARCH=arm64 _GOOS=linux`
+- `_GOARCH=ppc64le _GOOS=linux`
```
Adirio, line 19: Do this need to be listed somewhere else? I mean you will need to configure this list of os/arch somewhere (maybe not in code though).

estroz: Yeah I think @DirectXMan12 needs to update these vars in gcloud.

Adirio (approved):
/lgtm

PR #1995, opened 2021-02-08
🐛 Fix sorting issue with plugin versions and their supported project versions

Adirio:
When creating the plugin table with their supported project versions, the supported project versions were being store as a slice in the same order as the plugin keys slice.
When #1984 sorted the plugin keys slice, it didn't modify the other slice accordingly.


camilamacedo86 (approved):
/lgtm

Issue #1996, opened 2021-02-09
(go/v2) (config/v3) Webhook infomation missing in PROJECT


Adirio:
Using plugin `go.kubebuilder.io/v2` with `config/v3` is not saving to the project file the webhook information:

The following `PROJECT` configuration file corresponds to the same project used for `testdata/project-v2-multigroup` but using `config/v3` instead.
```yaml
domain: testproject.org
layout: go.kubebuilder.io/v2
multigroup: true
projectName: go-v2-config-v3
repo: sigs.k8s.io/kubebuilder/testdata/project-v2-multigroup
resources:
- api:
    crdVersion: v1beta1
    namespaced: true
  controller: true
  domain: testproject.org
  group: crew
  kind: Captain
  path: sigs.k8s.io/kubebuilder/testdata/project-v2-multigroup/apis/crew/v1
  version: v1
- api:
    crdVersion: v1beta1
    namespaced: true
  controller: true
  domain: testproject.org
  group: ship
  kind: Frigate
  path: sigs.k8s.io/kubebuilder/testdata/project-v2-multigroup/apis/ship/v1beta1
  version: v1beta1
- api:
    crdVersion: v1beta1
  controller: true
  domain: testproject.org
  group: ship
  kind: Destroyer
  path: sigs.k8s.io/kubebuilder/testdata/project-v2-multigroup/apis/ship/v1
  version: v1
- api:
    crdVersion: v1beta1
  controller: true
  domain: testproject.org
  group: ship
  kind: Cruiser
  path: sigs.k8s.io/kubebuilder/testdata/project-v2-multigroup/apis/ship/v2alpha1
  version: v2alpha1
- api:
    crdVersion: v1beta1
    namespaced: true
  controller: true
  domain: testproject.org
  group: sea-creatures
  kind: Kraken
  path: sigs.k8s.io/kubebuilder/testdata/project-v2-multigroup/apis/sea-creatures/v1beta1
  version: v1beta1
- api:
    crdVersion: v1beta1
    namespaced: true
  controller: true
  domain: testproject.org
  group: sea-creatures
  kind: Leviathan
  path: sigs.k8s.io/kubebuilder/testdata/project-v2-multigroup/apis/sea-creatures/v1beta2
  version: v1beta2
- api:
    crdVersion: v1beta1
    namespaced: true
  controller: true
  domain: testproject.org
  group: foo.policy
  kind: HealthCheckPolicy
[...]

PR #1998, opened 2021-02-09
🐛 Fix go/v2 with config/v3 resources in config file to store webhook information

Adirio:
Fix go/v2 with config/v3 resources in config file to store webhook information

Fixes: #1996


camilamacedo86 (approved):
/lgtm

PR #2000, opened 2021-02-09
📖Add page in book for how to watch objects referenced in the CRD.

HoustonPutman:
Currently there is no information in the Kubebuilder book on how to really use the Watches() functionality.

This PR adds a page that explains how to watch and reconcile an object based on changes to an object that it references in its CRD.

This feature is used extensively in the [Solr Operator](https://github.com/apache/lucene-solr-operator), so the ideas behind the test code have been tested.

Very open to suggestions/changes. This is my first PR for Kubebuilder, and I'm sure there are things I have missed.

### DIFF
--- a/docs/book/src/SUMMARY.md
+++ b/docs/book/src/SUMMARY.md
@@ -73,6 +73,9 @@
 
   - [Generating CRDs](./reference/generating-crd.md)
   - [Using Finalizers](./reference/using-finalizers.md)
+  - [Watching Resources](./reference/watching-resources.md)
+    - [Resources Managed by the Operator](./reference/watching-resources/operator-managed.md)
+    - [Externally Managed Resources](./reference/watching-resources/externally-managed.md)
   - [Kind cluster](reference/kind.md)
   - [What's a webhook?](reference/webhook-overview.md)
     - [Admission webhook](reference/admission-webhook.md)
--- a/docs/book/src/reference/reference.md
+++ b/docs/book/src/reference/reference.md
@@ -5,6 +5,11 @@
     Finalizers are a mechanism to
     execute any custom logic related to a resource before it gets deleted from
     Kubernetes cluster.
+  - [Watching Resources](watching-resources.md)
+    Watch resources in the Kubernetes cluster to be informed and take actions on changes.
+      - [Resources Managed by the Operator](watching-resources/operator-managed.md)
+      - [Externally Managed Resources](watching-resources/externally-managed.md)
+        Controller Runtime provides the ability to watch additional resources relevant to the controlled ones.
   - [Kind cluster](kind.md)
   - [What's a webhook?](webhook-overview.md)
     Webhooks are HTTP callbacks, there are 3
--- a/docs/book/src/reference/watching-resources.md
+++ b/docs/book/src/reference/watching-resources.md
@@ -1,0 +1,16 @@
+# Watching Resources
+
+Inside a `Reconcile()` control loop, you are looking to do a collection of operations until it has the desired state on the cluster.
+Therefore, it can be necessary to know when a resource that you care about is changed.
+In the case that there is an action (create, update, edit, delete, etc.) on a watched resource, `Reconcile()` should be called for the resources watching it.
+
+[Controller Runtime libraries](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/builder) provide many ways for resources to be managed and watched.
+This ranges from the easy and obvious use cases, such as watching the resources which were created and managed by the controller, to more unique and advanced use cases.
+
+See each subsection for explanations and examples of the different ways in which your controller can _Watch_ the resources it cares about.
+
+- [Watching Operator Managed Resources](watching-resources/operator-managed.md) -
+  These resources are created and managed by the same operator as the resource watching them.
+  This section covers both if they are managed by the same controller or separate controllers.
+- [Watching Externally Managed Resources](watching-resources/externally-managed.md) -
+  These resources could be manually created, or managed by other operators/controllers or the Kubernetes control plane.
--- a/docs/book/src/reference/watching-resources/externally-managed.md
+++ b/docs/book/src/reference/watching-resources/externally-managed.md
@@ -1,0 +1,31 @@
+# Watching Externally Managed Resources
+
+By default, Kubebuilder and the Controller Runtime libraries allow for controllers
+to easily watch the resources that they manage as well as dependent resources that are `Owned` by the controller.
+However, those are not always the only resources that need to be watched in the cluster.
+
+## User Specified Resources
+
+There are many examples of Resource Specs that allow users to reference external resources.
+- Ingresses have references to Service objects
+- Pods have references to ConfigMaps, Secrets and Volumes
+- Deployments and Services have references to Pods
+
+This same functionality can be added to CRDs and custom controllers.
+This will allow for resources to be reconciled when another resources it references is changed.
+
+As an example, we are going to create a `ConfigDeployment` resource.
+The `ConfigDeployment`'s purpose is to manage a `Deployment` whose pods are always using the latest version of a `ConfigMap`.
+While ConfigMaps are auto-updated within Pods, applications may not always be able to auto-refresh config from the file system.
+Some applications require restarts to apply configuration updates.
+- The `ConfigDeployment` CRD will hold a reference to a ConfigMap inside its Spec.
+- The `ConfigDeployment` controller will be in charge of creating a deployment with Pods that use the ConfigMap.
+These pods should be updated anytime that the referenced ConfigMap changes, therefore the ConfigDeployments will need to be reconciled on changes to the referenced ConfigMap.
+
+### Allow for linking of resources in the `Spec`
+
+{{#literatego ./testdata/external-indexed-field/api.go}}
+
+### Watch linked resources
+
+{{#literatego ./testdata/external-indexed-field/controller.go}}
--- a/docs/book/src/reference/watching-resources/testdata/external-indexed-field/controller.go
+++ b/docs/book/src/reference/watching-resources/testdata/external-indexed-field/controller.go
@@ -1,0 +1,208 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+// +kubebuilder:docs-gen:collapse=Apache License
+
+/*
+Along with the standard imports, we need additional controller-runtime and apimachinery libraries.
+All additional libraries, necessary for Watching, have the comment `Required For Watching` appended.
+*/
+package external_indexed_field
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	kapps "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/fields" // Required for Watching
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types" // Required for Watching
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder" // Required for Watching
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler" // Required for Watching
+	"sigs.k8s.io/controller-runtime/pkg/predicate" // Required for Watching
+	"sigs.k8s.io/controller-runtime/pkg/reconcile" // Required for Watching
+	"sigs.k8s.io/controller-runtime/pkg/source" // Required for Watching
+
+	appsv1 "tutorial.kubebuilder.io/project/api/v1"
+)
+
+/*
+Determine the path of the field in the ConfigDeployment CRD that we wish to use as the "object reference".
+This will be used in both the indexing and watching.
+*/
+const (
+	configMapField = ".spec.configMap"
+)
+
+/*
+*/
+
+// ConfigDeploymentReconciler reconciles a ConfigDeployment object
+type ConfigDeploymentReconciler struct {
+	client.Client
+	Log    logr.Logger
+	Scheme *runtime.Scheme
+}
+// +kubebuilder:docs-gen:collapse=Reconciler Declaration
+
+/*
+There are two additional resources that the controller needs to have access to, other than ConfigDeployments.
+- It needs to be able to fully manage Deployments, as well as check their status.
+- It also needs to be able to get, list and watch ConfigMaps.
+All 3 of these are important, and you will see usages of each below.
+*/
+
+//+kubebuilder:rbac:groups=apps.tutorial.kubebuilder.io,resources=configdeployments,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=apps.tutorial.kubebuilder.io,resources=configdeployments/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=apps.tutorial.kubebuilder.io,resources=configdeployments/finalizers,verbs=update
+//+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=apps,resources=deployments/status,verbs=get
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch
+
+/*
+`Reconcile` will be in charge of reconciling the state of ConfigDeployments.
+ConfigDeployments are used to manage Deployments whose pods are updated whenever the configMap that they use is updated.
+
+For that reason we need to add an annotation to the PodTemplate within the Deployment we create.
+This annotation will keep track of the latest version of the data within the referenced ConfigMap.
+Therefore when the version of the configMap is changed, the PodTemplate in the Deployment will change.
+This will cause a rolling upgrade of all Pods managed by the Deployment.
+
+Skip down to the `SetupWithManager` function to see how we ensure that `Reconcile` is called when the referenced `ConfigMaps` are updated.
+*/
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+func (r *ConfigDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	/* */
+	log := r.Log.WithValues("configDeployment", req.NamespacedName)
+
+	var configDeployment appsv1.ConfigDeployment
+	if err := r.Get(ctx, req.NamespacedName, &configDeployment); err != nil {
+		log.Error(err, "unable to fetch ConfigDeployment")
+		// we'll ignore not-found errors, since they can't be fixed by an immediate
+		// requeue (we'll need to wait for a new notification), and we can get them
+		// on deleted requests.
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+	// +kubebuilder:docs-gen:collapse=Begin the Reconcile
+
+	// your logic here
+
+	var configMapVersion string
+	if configDeployment.Spec.ConfigMap != "" {
+		configMapName := configDeployment.Spec.ConfigMap
+		foundConfigMap := &corev1.ConfigMap{}
+		err := r.Get(ctx, types.NamespacedName{Name: configMapName, Namespace: configDeployment.Namespace}, foundConfigMap)
+		if err != nil {
+			// If a configMap name is provided, then it must exist
+			// You will likely want to create an Event for the user to understand why their reconcile is failing.
+			return ctrl.Result{}, err
+		}
+
+		// Hash the data in some way, or just use the version of the Object
+		configMapVersion = foundConfigMap.ResourceVersion
+	}
+
+	// Logic here to add the configMapVersion as an annotation on your Deployment Pods.
+
+	return ctrl.Result{}, nil
+}
+
+/*
+Finally, we add this reconciler to the manager, so that it gets started
+when the manager is started.
+
+Since we create dependency Deployments during the reconcile, we can specify that the controller `Owns` Deployments.
+
+However the ConfigMaps that we want to watch are not owned by the ConfigDeployment object.
+Therefore we must specify a custom way of watching those objects.
+This watch logic is complex, so we have split it into a separate method.
+*/
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ConfigDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	/*
+		The `configMap` field must be indexed by the manager, so that we will be able to lookup `ConfigDeployments` by a referenced `ConfigMap` name.
+		This will allow for quickly answer the question:
+		- If ConfigMap _x_ is updated, which ConfigDeployments are affected?
+	*/
+
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &appsv1.ConfigDeployment{}, configMapField, func(rawObj client.Object) []string {
+		// Extract the ConfigMap name from the ConfigDeployment Spec, if one is provided
+		configDeployment := rawObj.(*appsv1.ConfigDeployment)
+		if configDeployment.Spec.ConfigMap == "" {
+			return nil
+		}
+		return []string{configDeployment.Spec.ConfigMap}
+	}); err != nil {
+		return err
+	}
+
+	/*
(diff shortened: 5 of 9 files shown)


docs/book/src/reference/testdata/watching/controller.go
```diff
@@ -0,0 +1,218 @@
...
+	Because we have already created an index on the configMap reference field, this mapping function is quite straight forward.
+	We first need to list out all `ConfigDeployments` that use `ConfigMap` given in the mapping function.
+	This is done by merely submitting a List request using our indexed field as the field selector.
+
+	When the list of `ConfigDeployments` that reference the `ConfigMap` is found,
+	we just need to loop through the list and create a reconcile request for each one.
+	If an error occurs fetching the list, or no `ConfigDeployments` are found, then no reconcile requests will be returned.
+	 */
+
+	return ctrlBuilder.Watches(
```
camilamacedo86, line 191: Your goal explain how to watch a resource that is not managed by the controller, am I right? then, I do not think that you need to do this implementation in predicate to list all. See : 

```go
err := c.Watch(&source.Kind{Type: obj}, &handler.EnqueueRequestForOwner{
		IsController: isConttroller,
		OwnerType:    owner,
	})
```

So, you can pass false in the `isConttroller`. Did you try that?

docs/book/src/SUMMARY.md
```diff
@@ -79,6 +79,7 @@
 
   - [Generating CRDs](./reference/generating-crd.md)
   - [Using Finalizers](./reference/using-finalizers.md)
+  - [Watching Objects](./reference/watching-objects.md)
```
camilamacedo86, line 82:
```suggestion
  - [Observing Resources](./reference/observing-objects.md)
```

camilamacedo86: I think we can use the word observe to make clear that we are speaking over observability and than, describe that the watch feature allows that.

docs/book/src/reference/reference.md
```diff
@@ -5,6 +5,9 @@
     Finalizers are a mechanism to
     execute any custom logic related to a resource before it gets deleted from
     Kubernetes cluster.
+  - [Watching Objects](watching-objects.md)
+    Controller Runtime provides the ability to watch objects relevant to
+    the objects that you own.
   - [Kind cluster](kind.md)
```
camilamacedo86, line 11: The watch feature allows you to observe relevant resources for your controller not only those ones that are onwed by the controller.

docs/book/src/reference/watching-objects.md
```diff
@@ -0,0 +1,21 @@
...
+by the controller. However, those are not always the only resources that need to be watched in the cluster.
+
+There are many examples of Specs that have references to external resources.
+- Ingresses have references to Service objects
+- Pods have references to ConfigMaps, Secrets and Volumes
+- Deployments and Services have references to Pods
+
+This same functionality can be added to CRDs and custom controllers.
+This will allow for objects to be reconciled when an object it references is changed.
+
```
camilamacedo86, line 14: I think that we would need to add here the controller-runtime docs over the watch as well.

docs/book/src/reference/testdata/watching/controller.go
```diff
@@ -0,0 +1,218 @@
...
+	*/
+	configMapField := ".spec.configMap"
+
+	/*
+	This field must be indexed by the manager, so that we will be able to lookup `ConfigDeployments` by a referenced `ConfigMap` name.
+	This will allow for quickly answer the question:
+	- If ConfigMap _x_ is updated, which ConfigDeployments are affected?
+	 */
+
+	if err := mgr.GetFieldIndexer().IndexField(context.TODO(), &appsv1.ConfigDeployment{}, configMapField, func(rawObj runtime.Object) []string {
```
camilamacedo86, line 161: Why the GetFieldIndexer().IndexField is required in this example?

docs/book/src/reference/testdata/watching/controller.go
HoustonPutman: So I think it would be useful to add a longer intro into the use case before diving into the code.

Basically the idea here is that the user has a `ConfigMap` which _they_ manage. However they want the Deployment, managed by the controller, to use this ConfigMap and for the Pods to be updated whenever the ConfigMap changes. So the controller is not controlling the ConfigMap.

This is a pretty common use case, especially for resources such as TLS certs in Secrets. You may want to auto-update pods when the TLS Certs have been renewed (i.e. the secret has been updated). Maybe a more concrete use case would be a good idea? My intention was to be as simple as possible, but I might have missed the mark there.

I'm not 100% certain, but I don't think the logic would be the same using:
```go
err := c.Watch(&source.Kind{Type: obj}, &handler.EnqueueRequestForOwner{
		IsController: false,
		OwnerType:    owner,
	})
```

The ConfigDeployment is not the owner of the ConfigMap, just a user of it. If we used this watch, then we would have to add an owner reference to a ConfigMap for every ConfigDeployment that uses it (not necessarily a 1<->1 mapping).

HoustonPutman: And a somewhat off-topic question about your second example.

Is that any different from using the following?
```go
 func (r *MemcachedReconciler) SetupWithManager(mgr ctrl.Manager) error {
	return ctrl.NewControllerManagedBy(mgr).
		For(&cachev1alpha1.Memcached{}).
		Owns(&corev1.ConfigMap{}).
		Owns(&appsv1.Deployment{}).
		Complete(r)
}
```

docs/book/src/reference/testdata/watching/controller.go
HoustonPutman: The field indexer is necessary in order to allow for easy lookup based on the "configMap" field in the spec. This makes the EnqueueRequestsFromMapFunc fast and easy. Otherwise we would need to list all ConfigDeployments from the namespace, and iterate through them finding the ones that use the ConfigMap passed through in the mapping function.

docs/book/src/reference/testdata/watching/controller.go
camilamacedo86: See [here](https://github.com/dev4devs-com/postgresql-operator/blob/master/pkg/controller/backup/controller.go#L61). I watch a resource `v1.Service` which is not managed by the controller and has another owner. 

IMO we need to the doc with examples; 
- the common basic scenario ( watch a secondary resource managed by the controller)
- how to watch a resource which is owned/managed by another controller in the project ( [here](https://github.com/dev4devs-com/postgresql-operator/blob/master/pkg/controller/backup/controller.go#L61))
- then, how to watch a resource that we do not know the owner/who manages and we want to use predicates to filter as for example by labels.

docs/book/src/SUMMARY.md
HoustonPutman: I used "Watching" because I think it's what most people would be looking for when trying to implement this logic. It's a keyword used all around Kubernetes and EtcD, as well as other distributed systems, like ZooKeeper and Consul.

Observing feels a little too passive for me, since we are keeping track of these resources in order to take actions on any changes. Observing is a little more removed, like keeping track of metrics.

Have no qualms changing "Objects" -> "Resources" though!

camilamacedo86 (commented):
I added some comments and suggestions. However., really thank you for a collab with. I think this doc is very important since many users raise the same questions covered by it. 

Thank you for your nice contribution. 🥇

HoustonPutman:
Thanks for such a quick and thorough review!

Hope to be contributing more in the future 😃

docs/book/src/reference/testdata/watching/controller.go
HoustonPutman: Agreed, I think it would be good to start off basic and through each example get more complex.

For the last example I think it would be good to show using a label filter. But then I think the example I gave would be a good extension on top of that. (Many CRDs I have seen have uses cases like this, and it is pretty confusing for new people who want to use it)

That would give us 4 examples, the first one being extremely simple. And the labels one should be pretty quick as well.

HoustonPutman:
I'm working on expanding the docs to multiple use cases of "Watch". Will push whenever that is in a ready-enough state to share.

docs/book/src/SUMMARY.md
Adirio: +1 to the "object" -> "resource" change. I'm fine with "watching".

docs/book/src/reference/testdata/watching/api.go
```diff
@@ -0,0 +1,77 @@
...
+// +kubebuilder:docs-gen:collapse=Apache License
+/* */
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+// +kubebuilder:docs-gen:collapse=Imports
+
+/*
```
Adirio, line 24: I think you first need to describe the example a bit. As an example:

> We are going to create a `ConfigDeployment` CRD that will hold a reference to a `ConfigMap` inside its Spec.

docs/book/src/reference/testdata/watching/api.go
```diff
@@ -0,0 +1,77 @@
...
+In the our type's Spec, we want to allow the user to pass in a reference to another object in the same namespace.
```
Adirio, line 25:
```suggestion
In our type's Spec, we want to allow the user to pass in a reference to another object in the same namespace.
```

Adirio, line 25:
```suggestion
In our type's Spec, we want to allow the user to pass in a reference to another object in the same namespace.
```

docs/book/src/reference/testdata/watching/api.go
```diff
@@ -0,0 +1,77 @@
...
+In this case it is a configMap.
```
Adirio, line 26: This would be described before.
```suggestion
```

docs/book/src/reference/testdata/watching/api.go
```diff
@@ -0,0 +1,77 @@
...
+It's also possible for this to be a namespaced reference, but in this example we will assume that the referenced object
+lives in the same namespace.
+
+This field does not need to be optional.
+If the field is required, the only the indexing code in the controller will need to be modified.
```
Adirio, line 31:
```suggestion
If the field is required, the indexing code in the controller will need to be modified.
```

docs/book/src/reference/testdata/watching/controller.go
```diff
@@ -0,0 +1,218 @@
...
+There are two necessary steps that will allow us to watch the `ConfigMaps` that are referenced by `ConfigDeployments`.
+Index the reference field and setup a watch.
+*/
+
+func (r *ConfigDeploymentReconciler) watchUserConfigMaps(mgr ctrl.Manager, ctrlBuilder *builder.Builder) (*builder.Builder, error) {
+	/*
+	Determine the path of the field in the ConfigDeployment CRD that we wish to use as the "object reference".
+	This will be used in both the indexing and watching.
+	*/
+	configMapField := ".spec.configMap"
```
Adirio, line 153: This should probably be a constant at the top of the file.

docs/book/src/reference/testdata/watching/controller.go
```diff
@@ -0,0 +1,218 @@
...
+
+/*
+Finally, we add this reconciler to the manager, so that it gets started
+when the manager is started.
+
+Since we create dependency Deployments during the reconcile, we can specify that the controller `Owns` Deployments.
+
+However the ConfigMaps that we want to watch are not owned by the ConfigDeployment object.
+Therefore we must specify a custom way of watching those objects.
+This watch logic is complex, so we have split it into a separate method.
```
Adirio, line 127: IMO it is a bit unclear extracting all to a function. I would:
- Keep the key as a `const` at the top of the file.
- Set the index first similar to how we do it in [the CronJob tutorial](https://book.kubebuilder.io/cronjob-tutorial/controller-implementation.html#setup), before calling `ctr.NewControllerManagedBy`.
- Call the `.Watches` here and extract to a separate function only what you have as an anonymous function below (inside `handler.ToRequestsFunc`).

Adirio (changes requested):

docs/book/src/reference/testdata/watching/controller.go
HoustonPutman: I have refactored this method, so it should now be how you have described.

Let me know if it's still unclear.

HoustonPutman:
I have refactored the docs to add space for the [3 additional use cases](https://github.com/kubernetes-sigs/kubebuilder/pull/2000#discussion_r573860455) of `Watching`:
- `Owns()` by the controller
- `For()` by a different controller within the operator
- External resources that are determined by the controller, such as resources that match a set of labels.

These don't have any content in them yet, but should be easy enough to fill into the spots given. I'll do what I can for that and let y'all know if I need help with a section or two.

The original use case should be fixed up according to the comments given.

The base `Watching Resources` page has a very basic introduction to watching resources. Probably needs to be spruced up a bit.

docs/book/src/watching-resources/externally-managed.md
```diff
@@ -0,0 +1 @@
+# Externally Managed Resources
```
Adirio, line 1: Is this file needed?

docs/book/src/watching-resources/operator-managed.md
```diff
@@ -0,0 +1 @@
+# Resources Managed by the Operator
```
Adirio, line 1: Is this file needed?

go.mod
```diff
@@ -7,8 +7,9 @@ require (
 	github.com/onsi/ginkgo v1.12.0
 	github.com/onsi/gomega v1.9.0
 	github.com/spf13/afero v1.2.2
-	github.com/spf13/cobra v0.0.7
+	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/tools v0.0.0-20200403190813-44a64ad78b9b
+	golang.org/x/tools v0.0.0-20200616195046-dc31b401abb5
+	sigs.k8s.io/controller-tools v0.4.1 // indirect
```
Adirio, line 13: You should definetely not be bumping dependency versions in a doc PR (applies to `go.sum` too).

docs/book/src/reference/watching-resources/testdata/external-indexed-field/controller.go
```diff
@@ -0,0 +1,210 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+// +kubebuilder:docs-gen:collapse=Apache License
+
+/*
+Along with the standard imports, we need additional controller-runtime and apimachinery libraries.
```
Adirio, line 18: You may want to explain briefly which ones are new or at least mark them with a comment so that the reader doesn't fail to add one of them.

docs/book/src/reference/watching-resources/testdata/external-indexed-field/controller.go
```diff
@@ -0,0 +1,210 @@
...
+		configMapVersion = foundConfigMap.ResourceVersion
+	}
+
+	// Logic here to add the configMapVersion as an annotation on your Deployment Pods.
+
+	/* */
+
+	return ctrl.Result{}, nil
+}
+// +kubebuilder:docs-gen:collapse=Finish the Reconcile
```
Adirio, line 124: I don't think hidding this part makes that much sense as it is just one LoC. Also, the function seems a bit odd without the closing curly bracket.

docs/book/src/reference/watching-resources/testdata/external-indexed-field/controller.go
```diff
@@ -0,0 +1,210 @@
...
+		For(&appsv1.ConfigDeployment{}).
+		Owns(&kapps.Deployment{}).
+		Watches(
+			&source.Kind{Type: &corev1.ConfigMap{}},
+			&handler.EnqueueRequestsFromMapFunc{
+				ToRequests: handler.ToRequestsFunc(r.findObjectsForConfigMap),
+			},
+			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}))
+
+	return ctrlBuilder.Complete(r)
```
Adirio, line 177:
```suggestion
	ctrlBuilder := ctrl.NewControllerManagedBy(mgr).
		For(&appsv1.ConfigDeployment{}).
		Owns(&kapps.Deployment{}).
		Watches(
			&source.Kind{Type: &corev1.ConfigMap{}},
			&handler.EnqueueRequestsFromMapFunc{
				ToRequests: handler.ToRequestsFunc(r.findObjectsForConfigMap),
			},
			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
		).
		Complete(r)
```

docs/book/src/reference/watching-resources/testdata/external-indexed-field/controller.go
```diff
@@ -0,0 +1,210 @@
...
+ConfigDeployments are used to manage Deployments whose pods are updated whenever the configMap that they use is updated.
+
+For that reason we need to add an annotation to the PodTemplate within the Deployment we create.
+This annotation will keep track of the latest version of the data within the referenced ConfigMap.
+Therefore when the version of the configMap is changed, the PodTemplate in the Deployment will change.
+This will cause a rolling upgrade of all Pods managed by the Deployment.
+
+Skip down to the `SetupWithManager` function to see how we ensure that `Reconcile` is called when the referenced `ConfigMaps` are updated.
+*/
+func (r *ConfigDeploymentReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
```
Adirio, line 86: You are using kubebuilder v2.3.1 or lower. Current documentation is already targeting v3.0.0 which defaults `go/v3` plugin. Could you please update the example to use this plugin? I would suggest you create a dummy project with a single API and its controller with your current version of kubebuilder, and then use the latest v3.0.0-alphaX release to create an equivalent project and compare the `*_types.go` and `*_controller.go` files. I know that some of the changes are:
- The [`Reconcile` function](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.7.0/pkg/reconcile#Func) signature has changed to also receive a `context.Context` as the first parameter. This removes the need to call `context.Background()` (which by the way you were assigning to `_` but later using `ctx`).
- RBAC for finalizers subresources was added.
- Both functions were added some comments to document them.

I am listing them from memory and without knowing exactly which version of kubebuilder you are using, so double check this please.

Adirio (changes requested):

Adirio:
Nice job @HoustonPutman, I added several comments, but it is starting to look good. Most of the comments are really easy to solve, like removing some extra files, not including `go.mod` and `go.sum` changes, and a couple style comments. The largest one would be the last. Feel free to ping me when you have updates.

docs/book/src/reference/testdata/watching/controller.go
Adirio: Perfect, you can resolve this conversation.

mweibel:
Not sure if this is the right place to put feedback, but thanks for this - I just had to implement a watch like this and struggled finding the right documentation until this PR. :tada: 

I followed along on the deploy preview (https://deploy-preview-2000--kubebuilder.netlify.app/reference/watching-resources/externally-managed.html) and that worked well. 
My case is I have a cluster-scoped custom resource referencing a `Secret` and need to watch the secret for changes.
Two questions have arisen which I haven't been able to answer fully yet:

a) When my cluster-scoped custom resource is added *before* the secret, the watch never gets triggered. What's the best way to handle this? Should I reconcile again after a while (with a max tries setting on the CR)?

b) The example is fetching a CM in the same namespace as the CR. My case is slightly different (as written above). I changed my implementation in that I removed the namespace from the List request as my CR is cluster-scoped. Is this the only change needed or would there be other things to improve on?

Would be great if these two questions could be answered briefly in the docs as well. 

Thanks again :+1:

Adirio:
@mweibel Use either Slack or another Issue to raise your support questions please. That way we don't make noise in this PR.

go.mod
HoustonPutman: Yeah, I was using ./docs/book/install_and_build.sh to test out the changes. Accidentally committed the go.mod/sum. Not intentional at all 😅 

Will un-commit these changes.

docs/book/src/watching-resources/externally-managed.md
HoustonPutman: So I split out these files in order to make space for the information [requested here](https://github.com/kubernetes-sigs/kubebuilder/pull/2000#discussion_r573860455).

I'm perfectly happy to get this working with one page of documentation, then split it up and add more at a later date, in a separate PR. Will defer to y'alls guidance here.

docs/book/src/reference/watching-resources/testdata/external-indexed-field/controller.go
HoustonPutman: Yep, using Kubebuilder v2.3.1.

I can do that for sure.

HoustonPutman (commented):
Thanks for all the help, will ping when the next review is ready. One question though, since you think those extra files aren't necessary. Do you think the content of the extra files is unnecessary, or do you just think that all of the "Watching" content should go on one page?

> By the way, I don't know if you are aware, but in the checks below this, there is one called "netlify/kubebuilder/deploy-preview — Deploy preview ready!" and if you click in "Details" you will be able to see how the documentation would look like after this PR is merged.

Very cool, did not know that!

docs/book/src/watching-resources/externally-managed.md
HoustonPutman: Scratch this, didn't notice that these were empty files that should not have existed.

Adirio: Just to make it clear. I was just refering to these empty files not to the fact that you are using 3 files for the documentation. I'm totally cool with that.

HoustonPutman:
Hey @Adirio, I've added a section for basic watching of controlled/owned objects. It's extremely basic and just used to show the functionality.

I've removed 2 of the 4 proposed sections for now. I'm not very familiar with the other two use cases, so I am probably not the best person to try to explain them 😅 . We also have content for both `operator-managed` and `external-resources`, so it will be easy to pop the others in later. Would it be possible to move forward with the two sections included in the PR currently and add an Issue to complete the other 2 next?

Obviously please review the basic operator-managed example, since that's new since you lasted viewed the PR.

Also should I squash all of the commits together before the merging is done?

Adirio:
> Also should I squash all of the commits together before the merging is done?

Yes

docs/book/src/reference/watching-resources.md
```diff
@@ -0,0 +1,10 @@
+# Watching Resources
+
+Inside a `Reconcile()` control loop, controllers will often use the state of the kubernetes cluster, and resources in it, to make decisions.
+Therefore, it can be necessary to know when a resource that you care about is changed.
```
camilamacedo86, line 4:
```suggestion
Inside a `Reconcile()` control loop, you are looking to do a collection of operations until it has the desired state on the cluster. Therefore, it can be necessary to know when a resource that you care about is changed.
```

docs/book/src/reference/watching-resources.md
```diff
@@ -0,0 +1,10 @@
...
+
+The [Controller Runtime libraries](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/builder) provide many ways for resources to be managed, and watched.
+This ranges from the easy and obvious use cases, such as watching the resources you are controlling,
+to more unique and advanced use cases.
```
camilamacedo86, line 8:
```suggestion
[Controller Runtime libraries](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/builder) provide many ways for resources to be managed and watched. This ranges from the easy and obvious use cases, such as watching the resources which were created and managed by the controller to more unique and advanced use cases.
```

docs/book/src/reference/watching-resources/externally-managed.md
```diff
@@ -0,0 +1,31 @@
+# Watching Externally Managed Resources
+
```
camilamacedo86, line 2: imo the goal to use watch is re-triger the reconcile when the k8s api raise an event such as create/update/edit/delete of an action performed against the resource observed. 

For the kubebuilder does I do not think that we should describe detailed scenario such as this one which can lead a misunderstands. We can be more direct ad object such as:

- You want to watch a resource that is managed by the controller then ...

- Add  a note about what means managed by the  controller as you explained with the owner  (see in the other docs the syntax to add notes)
- Add the example and a very low verbose explation

- You want to watch a resource that is NOT managed by the controller then ...
- You want to implement a customized watch for specific scenarios then ...

So, what I am trying to say is that we ought to not cover when each scenario is or not appropriate because it is up to the solution and requirements.

HoustonPutman: Ok, made the changes you suggested for wording in `watching-resources.md`, as well as fleshed out some information on externally managed vs operator managed.

> So, what I am trying to say is that we ought to not cover when each scenario is or not appropriate because it is up to the solution and requirements.

I read through the examples, and I'm not exactly sure where it says when its appropriate to use each. I give some examples of the Kubernetes resources using the paradigms, then I briefly explain what the example code is trying to accomplish.

I think it would be much more confusing for users to not understand what the code is trying to do before trying to read through it.

Would you feel better about it if I emphasized more that this is merely an example, and not necessarily the right design choice for this situation?

camilamacedo86:
/ok-to-test
/lgtm 
/approved

It is open for a while and all comments were addressed. 
Teh content is good and we can shape it if required into follow ups.

PR #2003, opened 2021-02-10
🐛 (config/v3) Track in the project configuration file resources without API

### LINKED ISSUE #2001: Info has not been tracked when the resource is not generated
By running `kubebuilder create api --group core --version v1 --kind Pod --resource=false --controller=true`;

```
$ kubebuilder create api --group core --version v1 --kind Pod --force
Create Resource [y/n]
n
Create Controller [y/n]
y
controllers/pod_controller.go
Running make:
$ make generate
go: creating new go.mod: module tmp
Downloading sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1
go: found sigs.k8s.io/controller-tools/cmd/controller-gen in sigs.k8s.io/controller-tools v0.4.1
/Users/camilamacedo/go/src/example/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
```

So, the above scenario results now in:

```ymal
 $ cat PROJECT 
domain: my.domain
layout: go.kubebuilder.io/v3
projectName: example
repo: example
version: "3"
```

When should be 

```
domain: my.domain
layout: go.kubebuilder.io/v3
projectName: example
repo: example
resources:
- controller: true
  group: core
  kind: Pod
  path: k8s.io/core/apps/v1 // we might want to leave empty this one until we implement an RFE to support external types.
  version: v1
version: "3"
```

### LINKED ISSUE #2002: Scaffold scheme registration for external resources when path is known
External resources (`--resource=false`) do not scaffold certain lines like the `For()` in the manager configuration. While this is desired when we don't know the path of the resource, if we do know it we should scaffold those lines, e.g. builting/core resources.

/kind feature

Adirio:
When an API is not scaffolded (`kubebuilder create api --resource=false --controller ...`) the resources are not being tracked in the config file. Tracking this in `config/v3` is doable as we can still know if the API was scaffolded by some of its fields, but as `config/v2` doesn't have this additional fields, it can't track this resources. The reason for this is that the pressence of a resource in `config/v2` means that the API was scaffolded at webhook validation (they require the API to be scaffolded) and this information won't be obtainable.

This change also detected that we were defining the `Path` field for those resources that do not have an API which was removed, and it allows to be a bit more intelligent when deploying core-resources.

Fixes #2001
Fixes #2002

### DIFF
--- a/generate_testdata.sh
+++ b/generate_testdata.sh
@@ -98,7 +98,7 @@ scaffold_test_project() {
 
     $kb create api --group foo.policy --version v1 --kind HealthCheckPolicy --controller=true --resource=true --make=false
 
-    $kb create api --group apps --version v1 --kind Pod --controller=true --resource=false --make=false
+    $kb create api --group apps --version v1 --kind Deployment --controller=true --resource=false --make=false
 
     if [ $project == "project-v3-multigroup" ]; then
       $kb create api --version v1 --kind Lakers --controller=true --resource=true --make=false
--- a/pkg/plugins/golang/v2/scaffolds/api.go
+++ b/pkg/plugins/golang/v2/scaffolds/api.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	"sigs.k8s.io/kubebuilder/v3/pkg/config"
+	cfgv2 "sigs.k8s.io/kubebuilder/v3/pkg/config/v2"
 	"sigs.k8s.io/kubebuilder/v3/pkg/model"
 	"sigs.k8s.io/kubebuilder/v3/pkg/model/resource"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v2/scaffolds/internal/templates"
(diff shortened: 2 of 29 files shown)


generate_testdata.sh
```diff
@@ -98,7 +98,7 @@ scaffold_test_project() {
 
     $kb create api --group foo.policy --version v1 --kind HealthCheckPolicy --controller=true --resource=true --make=false
 
-    $kb create api --group apps --version v1 --kind Pod --controller=true --resource=false --make=false
+    $kb create api --group apps --version v1 --kind Deployment --controller=true --resource=false --make=false
```
Adirio, line 101: `Pod` is not part of the `apps` group, it is part of the `core` group: [link](https://github.com/kubernetes/kubernetes/blob/11a05eb9aff104c2384781c740e053907a3849e6/staging/src/k8s.io/api/core/v1/types.go#L3664).

Replaced for `Deployment` which is part of the `apps` group: [link](https://github.com/kubernetes/kubernetes/blob/11a05eb9aff104c2384781c740e053907a3849e6/staging/src/k8s.io/api/apps/v1/types.go#L254).

pkg/plugins/golang/v2/scaffolds/api.go
```diff
@@ -89,11 +90,14 @@ func (s *apiScaffolder) scaffold() error {
 	doAPI := s.resource.HasAPI()
 	doController := s.resource.HasController()
 
-	if doAPI {
-
+	// config v2 should only be updated if the API is scaffolded
+	if doAPI || s.config.GetVersion().Compare(cfgv2.Version) == 1 {
 		if err := s.config.UpdateResource(s.resource); err != nil {
```
camilamacedo86, line 95: We should update the resource always not only is doAPI. 
If we create a webhook, or only a controller or an api, I mean in any scenario we need to store the info.

camilamacedo86 (commented):
It shows great. 
Could you please squash the commits for we are able to get it merged?

camilamacedo86 (approved):
/lgtm
/approve

Issue #2016, opened 2021-02-13
Plugins Ecosystem Next - Meta issue


camilamacedo86:
The propose of this issue is we try to centralize all discussed implementations over the plugin ecosystem for we have a summary definition of all goals and steps that we would like to persuade as their motivations. In this way, we can easily have a better idea and not allow one solution to invalidate another. The details can be found in each task/issue. 

- [x] - Plugin phase 1.5 : PR : https://github.com/kubernetes-sigs/kubebuilder/pull/2060

It will allow us to define more than 1 plugin in the default layout of a project. The layout defines the default/layout plugin(s) config for the project. By doing that, any action/command executed for this configuration would call and use all plugins defined as the default for the project.  

- [x] - Allow every command to overwrite project-configured plugins #1941 : Partial solved in https://github.com/kubernetes-sigs/kubebuilder/pull/2060 but we will still a follow up PR to cover this specific case.  

Allow a user to decide what execution should or not receive the additional plugins scaffolds. By using the `declarative.kubebuilder.io/v1` from above in this scenario users would able to decide what API should or not use the addon pattern instead of applying it for the full project which is the current behaviour before #1991 get merged. 

- [x] - Extract config/base plugin valid for any language/type from Golang: #2015

Allow the community to create any language/type plugin using the same base/config defined by KB in order to ensure its compatibility and/or maintainability. 

- [ ] - Plugin phase 2 https://github.com/kubernetes-sigs/kubebuilder/issues/1378

PS.: Shows that after the above implementations this latest task would be responsible to only check the possibility of the plugins be as separate binaries that can be discovered by the Kubebuilder CLI binary via user-specified plugin file paths

Adirio:
#2019 relates to the second point

daniel-garcia:
Unrelated question: are there any examples of writing a plugin? I can't seem to find anything in the kubebuilder book... maybe I'm not looking in the right place.

Adirio:
> Unrelated question: are there any examples of writing a plugin? I can't seem to find anything in the kubebuilder book... maybe I'm not looking in the right place.

Not really, but planned. This PRs haven't been merged yet so its a bit early. #2060 implements it so you can use the `pkg/plugins` folder as examples.

camilamacedo86:
Hi @daniel-garcia, 

[...]

PR #2019, opened 2021-02-16
✨ Plugin resolution

Adirio:
Improve the plugin resolution algorithm. From a conceptual POV, the `--project-version` is resolved after the `--plugins` flag, instead of the other way round.

It also adds the `Bundle` concept, which is a `Plugin` that wraps other plugins.

- Closes: #1941
- Closes: #1942
- [ ] Pre-requisite: #2059


Adirio:
Holding until #1991 is merged as it is dependent.
/hold

Adirio:
If #2048 is merged, this PR can be non-breaking.

cmd/main.go
```diff
@@ -30,13 +30,13 @@ func main() {
...
-		cli.WithDefaultProjectVersion(cfgv3.Version),
 		cli.WithPlugins(
 			&pluginv2.Plugin{},
 			&pluginv3.Plugin{},
+			&declarativev1.Plugin{},
 		),
-		cli.WithDefaultPlugins(cfgv2.Version, &pluginv2.Plugin{}),
-		cli.WithDefaultPlugins(cfgv3.Version, &pluginv3.Plugin{}),
+		cli.WithDefaultPlugins(&pluginv3.Plugin{}),
+		cli.WithDefaultProjectVersion(cfgv3.Version),
```
camilamacedo86, line 39: The scenario https://github.com/operator-framework/operator-sdk/pull/4581#discussion_r582216189 needs also be solved in #2019 before we get it merged.

Adirio: I will probably address that scenario in a follow up as it doesn't break the API, it will just add a new `PluginBundle` type that basically groups plugins together. This way, the operator-sdk will be able to create a bundle called `helm/v1` that groups `base.helm/v1`, `manifests/v1` and `scorecard/v1`. This way users will just need to call `operator-sdk init --plugin helm/v1` and kubebuilder will resolve it to the three plugins.

camilamacedo86: If we are able to have a workaround in place in operator-framework/operator-sdk#4581 until there I am totally all in with we do that as follow up. Otherwise, I think we need to have the solution before we get it all merged.

Adirio: This is already solved with `Bundle` and `NewBundle`

camilamacedo86:
This PR closes : 

- https://github.com/kubernetes-sigs/kubebuilder/issues/2043
- The issues #1941 and #1991 from https://github.com/kubernetes-sigs/kubebuilder/issues/2016
- the https://github.com/kubernetes-sigs/kubebuilder/issues/2056

To ensure that all changes proposed here also works with the consumer without breaking changes and indeed make the things easier maintained their @Adirio creates https://github.com/operator-framework/operator-sdk/pull/4581.

Since with all these changes, we can ensure that it has no breaking changes and all still imo we need to squash this pr and ask for a review on only this one and close the split/pre-requirements ones we cannot merge one without another because it would mean breaking changes that cannot be addressed for its conumer (SDK). 

c/c @estroz @DirectXMan12 could you help in the review?

camilamacedo86:
This PR does not close  #1941 . 
See my comment in : https://github.com/kubernetes-sigs/kubebuilder/pull/2060/files#r588914741

Adirio:
Closed in favor of #2060

PR #2020, opened 2021-02-16
:seedling: Update book-v3 with master 

camilamacedo86:
**Description**
Update book-v3 with master 

**Motivation**
Keep the branch updated for the next release


Adirio:
I think we should do this with direct git commits and not with a GitHub PR as this is not really reviewable and it just adds noise to the git history.

camilamacedo86:
closing it for we do as agreed ( from the release-v3 directly )

PR #2021, opened 2021-02-17
📖 Update static book examples to Kubebuilder v3.

HoustonPutman:
While working on the "Watching" documentation, I noticed that there were other examples in the book that had not yet been updated from v2 to v3.

This PR is pretty simple, and just updating the code differences with starting an empty project using v3.

For the finalizers, I added the RBAC finalizer permission as well, since that is available in v3.

The biggest difference is in `emptymain.go`, since there is a lot more generated code in v3 than in v2. I can scale the changes back for that file, but it now represents what a user would see when creating a new v3 project.


camilamacedo86 (approved):
/lgtm

Issue #2022, opened 2021-02-17
🐛 Actual error hidden behind unknown flag error


Adirio:
When the user provides some wrong flags, the error is reported with a help message for the root command.

For example: `kubebuilder init --project-version=2 --plugins=go/v3` will print:
```
Error:  no plugin could be resolved with key "go/v3" for project version "2"
Usage:
  kubebuilder [flags]

Examples:
[...]
[DATE] no plugin could be resolved with key "go/v3" for project version "2"
```

If an additional flag is provided, even if its valid for the subcommand, it won't be valid for the root command used to print the help message, so the error will change to `unknow flag: --flag` hiding te real error.

For example, adding the `--domain` flag to the previous call (which is a valid init flag): `kubebuilder init --project-version=2 --plugins=go/v3 --domain=my.domain` will print:
```
Error:  unknown flag: --domain
Usage:
  kubebuilder [flags]

Examples:
[...]
[DATE] unknown flag: --domain
```

The `--domain` flag is obviously not the problem here so the provided error message will only lead the user in the wrong direction.

/kind bug

PR #2023, opened 2021-02-17
🐛 Fix the bug where an error was being hidden by a potentially valid flag not being recognized by the root command

Adirio:
By telling the root command that it should not error out if found unknown flags we ensure that the error that made us print the root help message is surfaced instead of being hidden behind the `unknown flag: --flag` bug.

Fixes: #2022


camilamacedo86 (approved):
/lgtm

PR #2033, opened 2021-02-22
🌱 Update old TODOs that are no longer valid

Adirio:
The TODOs about v3 were already applied, removed them.
The TODO about the resource fields stored in the config file is out of date, updated it.


camilamacedo86 (approved):
/lgtm

PR #2034, opened 2021-02-22
🌱 Add tests to the DNS validation package

Adirio:
As the title says, this PR adds tests to the DNS validation package.
The test cases are the same as in upstream.

### DIFF
--- a/pkg/internal/validation/dns.go
+++ b/pkg/internal/validation/dns.go
@@ -27,7 +27,7 @@ import (
 
 const (
 	dns1123LabelFmt     string = "[a-z0-9](?:[-a-z0-9]*[a-z0-9])?"
-	dns1123SubdomainFmt string = dns1123LabelFmt + "(\\." + dns1123LabelFmt + ")*"
+	dns1123SubdomainFmt string = dns1123LabelFmt + "(?:\\." + dns1123LabelFmt + ")*"
 	dns1035LabelFmt     string = "[a-z](?:[-a-z0-9]*[a-z0-9])?"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/internal/validation/dns.go
```diff
@@ -27,7 +27,7 @@ import (
 
 const (
 	dns1123LabelFmt     string = "[a-z0-9](?:[-a-z0-9]*[a-z0-9])?"
-	dns1123SubdomainFmt string = dns1123LabelFmt + "(\\." + dns1123LabelFmt + ")*"
+	dns1123SubdomainFmt string = dns1123LabelFmt + "(?:\\." + dns1123LabelFmt + ")*"
```
camilamacedo86, line 30: It changes the validation. Could you please clarify why this change was made?

Adirio: It is changing the capture block into a non-capture block. You can see how every other group is a non-capture block: 

```
dns1123LabelFmt     string = "[a-z0-9](?:[-a-z0-9]*[a-z0-9])?"
```

This means that the regex matcher won't store that value in a variable, because we don't use it. All groups were change to non-capture groups in a recent PR but this one seems to have been missed.

pkg/internal/validation/dns.go
```diff
@@ -43,7 +43,8 @@ var dns1123LabelConfig = dnsValidationConfig{
 	format:   dns1123LabelFmt,
 	maxLen:   56, // = 63 - len("-system")
 	re:       regexp.MustCompile("^" + dns1123LabelFmt + "$"),
-	errMsg:   "a DNS-1123 label must consist of lower case alphanumeric characters or '-'",
+	errMsg:   "a DNS-1123 label must consist of lower case alphanumeric characters or '-', " +
+		"and must start and end with an alphanumeric character",
```
Adirio, line 47: The error message was imrpoved to match the other to checks.

pkg/internal/validation/dns.go
```diff
@@ -75,17 +76,17 @@ func (c dnsValidationConfig) check(value string) (errs []string) {
 	return errs
 }
 
+// IsDNS1123Label tests for a string that conforms to the definition of a label in DNS (RFC 1123).
+func IsDNS1123Label(value string) []string {
+	return dns1123LabelConfig.check(value)
+}
+
```
Adirio, line 83: This was sorted to keep the order consistent.

Issue #2036, opened 2021-02-23
✨ Export scaffolding machinery


Adirio:
The scaffolding machinery is not being exported as it is placed in an `internal` directory (from older kubebuilder versions when it was not designed to be a library. We should probably move this package outside of `internal` so that it can be used by third-party developers.

As an example, [operator SDK](https://github.com/operator-framework/operator-sdk) is [copying this folder](https://github.com/operator-framework/operator-sdk/tree/master/internal/kubebuilder).

/kind feature

camilamacedo86:
+1 for me.

camilamacedo86:
@Adirio will also check if nobody has any objections to it before move forward ( controller-tools meeting and @DirectXMan12 )

DirectXMan12:
Seems like the public part of these interfaces is fairly minimal, so it seems pretty reasonable to expose these without them being restrictive to internal refactor.

PR #2039, opened 2021-02-25
🐛 fix unit tests to work with later versions of ginkgo/gomega

joelanford:
I was re-using some of the internal packages in `pkg/plugins/internal` in a project that requires more recent versions of Ginkgo and Gomega and I noticed that the tests fail due to duplicate calls to `RunSpecs`.

We should call `RunSpecs` only once in each package. See https://github.com/onsi/ginkgo/issues/704#issuecomment-676642101 for what led me to this conclusion.

Also going ahead an bumping our deps to trigger this feedback in the future.


Adirio (approved):
I have to blame myself about this one.

Same happened in operator SDK (relevant commits: https://github.com/operator-framework/operator-sdk/pull/4259/commits/f1f5926cc994c73f6f3a3e60aca2785ceaee2557 and https://github.com/operator-framework/operator-sdk/pull/4259/commits/5a5c4fc20b5f4f1ce58edc3afe0f7e1952612ad1).

Also, as we are planning on exporting this packages (#2036), this fix is pretty well timed. I was actually planning on doing this myself.

/lgtm

PR #2040, opened 2021-02-25
:seedling: fix CI to use the correct go version 1.15

camilamacedo86:
**Description**
- fix CI to use the correct go version 1.15+ < 1.16
- clarifies it in the quick-start since 1.16 was released and has breaking changes for the projects built. with and tool


Adirio (approved):
/lgtm

Issue #2046, opened 2021-02-26
Final Config interface for v3.0.0


Adirio:
The [`Config` interface](https://github.com/kubernetes-sigs/kubebuilder/blob/d07dfcc555f10846f0bafdd0b7aa6c03b8e13dfb/pkg/config/interface.go#L23-L112) has 2 methods that seem a bit out of place.
`IsCRDVersionCompatible` and `IsWebhookVersionCompatible` encode inside a project configuration version the possibility of having only one or multiple different CRD/webhook versions. While this check is needed as we are not supporting multiple CRD/webhook versions, the project configuration itself is not th elimiting factor. A better API design would be replacing these two methods by `ListCRDVersions() []string` and `ListWebhookVersions() []string` letting the plugins be the ones that check if they are valid. This would allow a future plugin `go/v4` accept multiple CRD versions in a single project but still use project version 3.

/remove-kind feature

PR #2047, opened 2021-02-26
⚠️ Decouple plugin requirements from Config interface

Adirio:
Extract the single version per config file out of config as it is a plugin restriction, not a project configuration file restriction.

Closes: #2046

### DIFF
--- a/pkg/plugin/util/helpers.go
+++ b/pkg/plugin/util/helpers.go
@@ -1,0 +1,35 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"sigs.k8s.io/kubebuilder/v3/pkg/config"
+)
+
+// HasDifferentCRDVersion returns true if any other CRD version is tracked in the project configuration.
+func HasDifferentCRDVersion(config config.Config, crdVersion string) bool {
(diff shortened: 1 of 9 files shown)


pkg/plugin/util/helpers.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// IsOnlyWebhookVersion returns true if webhookVersion is the first or the only existing webhook version in the project.
+func IsOnlyWebhookVersion(config config.Config, webhookVersion string) bool {
+	return isOnlyAPIVersion(config.ListWebhookVersions(), webhookVersion)
+}
+
+func isOnlyAPIVersion(versions []string, version string) bool {
+	return len(versions) == 0 || (len(versions) == 1 && versions[0] == version)
+}
```
camilamacedo86, line 35: I understand that you change that for a List of webhook versions and List of CRD versions to allow the possibility of the tool works with many versions in the future. 

However, I do not think that we should have IsOnly funcs here. We would still have isWebhoohVersionCompatible  and IsCRDVersionCompatible where its responsibility would ensure that the version informed can or not be used.

pkg/plugins/golang/v3/api.go
```diff
@@ -189,7 +190,7 @@ func (p *createAPISubcommand) Validate() error {
 		}
 
 		// Check CRDVersion against all other CRDVersions in p.config for compatibility.
-		if !p.config.IsCRDVersionCompatible(p.resource.API.CRDVersion) {
+		if !pluginutil.IsOnlyCRDVersion(p.config, p.resource.API.CRDVersion) {
```
camilamacedo86, line 193: I do not think that this change is nice. I did not like the idea of have this funcs IsOnlyCRDVersion. The previous one represents a better need. We will pass the version of CRD or Webhook which we want to use and check if the project config allows or not it.

pkg/plugin/util/helpers.go
Adirio: We talked about this and `hasDifferentAPIVersion` seems like a better name (for the negated function).

camilamacedo86 (commented):
I agree with this change. Note that the config is responsible to have funcs to return its values, however, the rule to allow or not different APIs is specific to the plugin since we could, for example, support it for go/v4.

In this way, since it is a very small and concise change that does not affect the end-user behaviours and is only more a cleanup. I am happy in 

/lgtm

PR #2048, opened 2021-02-26
⚠️ Use a single set of default plugins instead of one per project version

Adirio:
This PR is the breaking change needed for #1941 and #1942 so that we #2019 no longer blocks v3.0.0 as it will no longer be a breaking change.

The exact same behavior has been maintained for all `kubebuilder init --plugins <PLUGINS> --project-version <PROJECT VERSION>` combination, including the absence of any of these flags.

- Follow up: #2019

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -26,8 +26,10 @@ import (
 
 	internalconfig "sigs.k8s.io/kubebuilder/v3/pkg/cli/internal/config"
 	"sigs.k8s.io/kubebuilder/v3/pkg/config"
+	cfgv2 "sigs.k8s.io/kubebuilder/v3/pkg/config/v2"
 	cfgv3 "sigs.k8s.io/kubebuilder/v3/pkg/config/v3"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
+	pluginv2 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v2"
 )
 
 const (
@@ -76,7 +78,7 @@ type cli struct { //nolint:maligned
 	// Default project version in case none is provided and a config file can't be found.
 	defaultProjectVersion config.Version
 	// Default plugins in case none is provided and a config file can't be found.
-	defaultPlugins map[config.Version][]string
+	defaultPlugins []string
 	// Plugins registered in the cli.
 	plugins map[string]plugin.Plugin
 	// Commands injected by options.
@@ -132,7 +134,7 @@ func newCLI(opts ...Option) (*cli, error) {
 	c := &cli{
 		commandName:           "kubebuilder",
 		defaultProjectVersion: cfgv3.Version,
-		defaultPlugins:        make(map[config.Version][]string),
+		defaultPlugins:        make([]string, 0),
 		plugins:               make(map[string]plugin.Plugin),
 	}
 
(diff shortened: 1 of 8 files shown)


test/testdata/generate.sh
```diff
@@ -113,10 +113,10 @@ function scaffold_test_project {
...
 build_kb
 
-# Project version 2 uses plugin go/v2 (default).
-scaffold_test_project project-v2 --project-version=2
-scaffold_test_project project-v2-multigroup --project-version=2
-scaffold_test_project project-v2-addon --project-version=2
+# Project version 2 uses plugin go/v2.
+scaffold_test_project project-v2 --plugins=go/v2 --project-version=2
+scaffold_test_project project-v2-multigroup --plugins=go/v2 --project-version=2
+scaffold_test_project project-v2-addon --plugins=go/v2 --project-version=2
```
camilamacedo86, line 119: We cannot change here to ensure that it still working as before.

pkg/cli/cli.go
```diff
@@ -187,6 +189,12 @@ func (c *cli) getInfoFromFlags() (string, []string, error) {
 		plugins[i] = strings.TrimSpace(key)
 	}
 
+	// For backwards compatibility reasons, defining the project version flag and not defining
+	// the plugins flag should be interpreted as if the plugins flag was set to go/v2.
+	if projectVersion == "2" && len(plugins) == 0 {
+		plugins = append(plugins, plugin.KeyFor(pluginv2.Plugin{}))
+	}
+
```
camilamacedo86, line 197: we need to update the first comment as well to clarifies that. 
I am ok with this change here since we are ensuring the `backwards compatibility` and not change the behaviour for `kb/sdk init project-version=2`. 


However, for we are able to move forward and get it merged we also need to check it with @estroz.
@estroz wdyt? Is it makes sense for you or do you see any edge scenario that would be affected as well? 

This change is required to address 1.5, however, I'd like to confirm with you that it can be changed to address the needs without effect any requirement.

test/testdata/generate.sh
Adirio: There is a test in the unit test suite that checks that the behavior is the same. I think that being explicit in interation tests (both testdata related tests and e2e test) is better than relying on defaults.

pkg/cli/cli.go
Adirio: PR description updated. It previously had a behavior that differ from the v2.3.2 release but it no longer does.

Just a note about Camila's comment. This PR applies the breaking changes that will be needed for #1941 and #1942. By doing them right now we ensure that #2019 (which solves #1941 implementing the proposal #1942) will be non-breaking, and therefore wil no longer block v3.0.0 release. (#1991 would still be breaking but at least it is one step less blocking the release.)

camilamacedo86: After re-think here, imo if a plugin supports n project versions than we should still able to use it without inform the --project-version. See: https://github.com/kubernetes-sigs/kubebuilder/pull/1942/files#r584166217

Adirio: You can still define a default project version, that has not changed. The only difference is that you no longer have multiple sets of default plugins.

Adirio:
This breaking change is not really needed for plugins phase 1.5 so there is no real point in merging it. Closing.

PR #2054, opened 2021-03-02
📖 update CRD path for multi-group api

ahalim-intel:
Added missing documentation for updating CRD paths update for envtest.

Currently there is no information in the Kubebuilder book on CRD path path update requirement for envtest to work properly when creating multi-group API.

This PR adds information that explains how to update the CRD path in controller test suit. This problem is discussed in [this issue](https://github.com/operator-framework/operator-sdk/issues/4434) on operator-sdk.

Signed-off-by: Abdul Halim <abdul.halim@intel.com>


Adirio (changes requested):
This only affects the migration docs. If an API was scaffolded with multi-group enabled it will use the correct path, but if migrating from a single-group project to a multi-group project this info is missing from the migration guide. The title should better reflect this change. Maybe something like: `📖 update multi-group migration guide about controller tests`.

Also, this change only affects `go/v3` as `go/v2` doesn't scaffold controller tests, so this should probably be added as a note that clearly specifies that this change is only needed with that plugin. Probably this fact is the reason why it was not present in the documentation before.

After those two issues are tackled, I will happyly approve and lgtm this. Thank you for your collaboration @ahalim-intel.

camilamacedo86 (approved):
Hi @Adirio, 

We will merge it on the master. Any user now that scaffolds a project will have the suite test and then, shows not mandatory describes that for projects which have not suite test scaffold does it requires that because it is implicit. 

I agree that the commit message could have a better description, however, I do not think that it should block we get this PR  merge to fix/improve the docs.

So, I will 

/lgtm 

This one. However, please feel free to push any PR against it with your suggestions and ideas for improvements as well. 

I hope that you agree that is not valid lose this great contribution because of these nits.

Issue #2056, opened 2021-03-03
✨ Export config file schafolding machinery


Adirio:
In order to scaffold the project configuration file, some machinery can be found in `pkg/cli/internal`. While the responsability of creating this file still relies on the CLI, the use of an internal package for this machinery, has made it grow in a somewhat "messy" way.

Exporting and cleaning up this package would be nice.

/kind feature

PR #2057, opened 2021-03-03
⚠️ Store abstraction for persisting Config

Adirio:
This PR exports the project configuration scaffolding machinery through a `Store` interface that has been implemented for the current YAML representation.

`PreRunE` and `PostRunE` are used to handle configuration file loading and saving.

Also, all the filesystem abstractions that were being created have been centralized to a single one provided by the CLI. This includes filesystem abstractions that were used for:
- Scaffolding project configuration file: previously could be several per command call.
- Scaffolding the command-related files: previously were several per command. 

NOTE: the addition of the `Store` interface and implementation is not a breaking change. However, the changes needed to inject the filesystem abstraction from the CLI to the different scaffolders does require to change certain interfaces.

- Closes: #2056
- Follow up: #2058


Adirio:
Plugin phase 1.5 (#1991) is built on top of these changes.

Adirio:
Closed in favor of #2083

PR #2062, opened 2021-03-04
:book: update redirects for v3.0.0

estroz:
- netlify.toml: update redirects for v3.0.0
- docs/book/src: update https://go.kubebuilder.io links to match new redirects and download instructions

### DIFF
--- a/docs/book/src/quick-start.md
+++ b/docs/book/src/quick-start.md
@@ -22,7 +22,7 @@ Projects created by Kubebuilder contain a Makefile that will install tools at ve
 - [controller-gen](https://github.com/kubernetes-sigs/controller-tools)
 
 The versions which are defined in the `Makefile` and `go.mod` files are the versions tested and therefore is recommend to use the specified versions.
- 
+
 </aside>
 
 ## Installation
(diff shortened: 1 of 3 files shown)


docs/book/src/quick-start.md
```diff
@@ -65,25 +54,25 @@ Kubebuilder provides autocompletion support for Bash and Zsh via the command `ku
...
+Read the [Go modules blogpost][go-modules-blogpost] if unfamiliar with the module system.
 
 </aside>
 
 <aside class="note">
 <h1>Go package issues</h1>
 
-Ensure that you activate the module support by running `$ export GO111MODULE=on` 
-to solve issues as `cannot find package ... (from $GOROOT)`.
+If you see an error like "cannot find package ... (from $GOROOT)", you may need to explicitly run `$ export GO111MODULE=on`.
```
Adirio, line 75: This should definetely not be needed in go1.15. Is there any case where the default behavior (auto) does not work and is not solved by the prefered `go mod init` command?

docs/book/src/quick-start.md
```diff
@@ -22,34 +22,23 @@ Projects created by Kubebuilder contain a Makefile that will install tools at ve
...
-
-If you are using a Kubebuilder plugin version less than version `v3+`, you must configure the Kubernetes binaries required for the [envtest][envtest], run: 
-
-```bash
-# move to a long-term location and put it on your path
-# (you'll need to set the KUBEBUILDER_ASSETS env var if you put it somewhere else)
-sudo mv /tmp/kubebuilder_2.3.1_${os}_${arch} /usr/local/kubebuilder
-export PATH=$PATH:/usr/local/kubebuilder/bin
+# download kubebuilder and install locally.
+curl -L -o kubebuilder https://go.kubebuilder.io/dl/3.0.0-beta.0/$(go env GOOS)/$(go env GOARCH)
```
Adirio, line 34: What about?

```suggestion
curl -L -o kubebuilder https://go.kubebuilder.io/dl/latest/$(go env GOOS)/$(go env GOARCH)
```

netlify.toml
```diff
@@ -56,41 +65,63 @@
...
+    force = true
+
+# Development branch redirect.
+[[redirects]]
+    from = "https://go.kubebuilder.io/releases/master/:os/:arch"
     to = "https://storage.googleapis.com/kubebuilder-release/kubebuilder_master_:os_:arch.tar.gz"
     status = 302
     force = true
 
+# Latest redirects.
```
Adirio, line 80: The order in this file matters, righ?
Wouldn't `https://go.kubebuilder.io/releases/latest` be captured by `https://go.kubebuilder.io/releases/:version` which is placed above? Same applies for the other 2 `latest` redirects.

I think `latest` redirect need to be placed above v1. See how previously, `https://go.kubebuilder.io/releases/latest/:os/:arch` was before `https://go.kubebuilder.io/releases/:version/:os/:arch`.

Adirio (changes requested):
We should definetely move forward with this as it has been asked in Slack recently, some comments:

docs/book/src/quick-start.md
estroz: No I think this can be removed completely.

docs/book/src/quick-start.md
estroz: Using `latest` isn't backwards compatible, since v4.0.0 will likely have the same `latest` endpoint.

netlify.toml
Adirio: Also, https://github.com/kubernetes-sigs/kubebuilder/releases/downloads/latest/kubebuilder_linux_amd64 doesn't seem to work.

docs/book/src/quick-start.md
Adirio: Makes complete sense. Leave it as it is.

estroz: I've backtracked and think latest makes the most sense (no updates needed for new releases either). This link can be updated to the last v3.y.z version prior to v4.0.0 when the time comes.

Adirio:
/lgtm
/hold

Holding until e2e tests are passed (#2063), feel free to remove the hold afterwards

Adirio:
NOTE: right now `latest` points to v2.3.2 (as v3.0.0-xxxxx have been marked as pre-releases). But once this gets into the actual book, there is going to be a v3.0.0 release which wil be where `latest` points to, so keeping `latest` is fine.

Issue #1253, opened 2019-12-10
Unable to export prometheus metrics be following the steps provided


camilamacedo86:
, I followed instructions in #1030 but I saw the following error in Prometheus for the metrics service endpoint:

"Get http://172.17.0.15:8443/metrics: net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x15\x03\x01\x00\x02\x02""

Note that Prometheus was using "http" to access the end point. So I added the following

    scheme: https
to "endpoints" of ServiceMonitor but then got this (expected) error:

"Get https://172.17.0.15:8443/metrics: x509: cannot validate certificate for 172.17.0.15 because it doesn't contain any IP SANs"

This is because the certificate used by the metrics service is not issued for the IP. I think we need to pass proper certificate info in "tlsConfig" (see https://coreos.com/operators/prometheus/docs/latest/api.html#tlsconfig). Either that or we need to enable no auth access to metrics end point.
Just to be clear, I am using Prometheus operator and not kube-prometheus but I doubt if that makes any difference. I will see if I can install the later and give it one more try but I would still expect similar errors as metrics end point is served only with "https".

Motivation: https://github.com/kubernetes-sigs/kubebuilder/pull/1245#discussion_r356119790

camilamacedo86:
Hi @draghuram,

As we spoke in the PR I am unable to reproduce your scenario. Following the steps used to check it. 

1. Install the Prometheus with Minikube. I am following the steps of https://github.com/coreos/kube-prometheus#quickstart. Following them:
- run `minikube delete && minikube start --kubernetes-version=v1.16.0 --memory=6g --bootstrapper=kubeadm --extra-config=kubelet.authentication-token-webhook=true --extra-config=kubelet.authorization-mode=Webhook --extra-config=scheduler.address=0.0.0.0 --extra-config=controller-manager.address=0.0.0.0`
- run `minikube addons disable metrics-server`
- Cloned the project locally `git clone git@github.com:coreos/kube-prometheus.git`
- Run `kubectl create -f manifests/setup`
- After  all be create run `kubectl create -f manifests/`
- Check that all is running `kubectl get all --namespace monitoring `
- Run `kubectl --namespace monitoring port-forward svc/prometheus-k8s 9090`

Now,  we can access the Prometheus here: http://localhost:9090/

2. Create a project and export the operator metrics by following the [quick start](https://book.kubebuilder.io/quick-start.html).

**Creating the project**
```
[...]

PR #1317, opened 2020-01-16
:bug: scaffold clusterrole for access the protected metrics

mengqiy:
- Scaffold a clusterrole for accessing metrics behind kube-rbac-proxy
- Add an e2e test for it
- Add documentation

PR #2065, opened 2021-03-05
🐛 default service monitor configuration to use https

johanneswuerbach:
By default the ServiceMonitor scrapes `port: https` https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v2/config/prometheus/monitor.yaml#L13, which is exposed by the auth proxy service https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v2/config/rbac/auth_proxy_service.yaml#L10-L12 and requires an https connection https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v2/config/default/manager_auth_proxy_patch.yaml#L15

Change the ServiceMonitor configuration to work with that by default as otherwise the scraping fails with a `400 Bad Request`.

Thanks to https://github.com/kubernetes-sigs/kubebuilder/issues/1253#issuecomment-597640633 for the required flags.

### DIFF
--- a/docs/book/src/multiversion-tutorial/testdata/project/config/prometheus/monitor.yaml
+++ b/docs/book/src/multiversion-tutorial/testdata/project/config/prometheus/monitor.yaml
@@ -11,6 +11,10 @@ spec:
   endpoints:
     - path: /metrics
       port: https
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
   selector:
     matchLabels:
       control-plane: controller-manager
--- a/pkg/plugins/golang/v2/scaffolds/internal/templates/config/prometheus/monitor.go
+++ b/pkg/plugins/golang/v2/scaffolds/internal/templates/config/prometheus/monitor.go
@@ -53,6 +53,10 @@ spec:
   endpoints:
     - path: /metrics
       port: https
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
   selector:
     matchLabels:
       control-plane: controller-manager
--- a/pkg/plugins/golang/v3/scaffolds/internal/templates/config/prometheus/monitor.go
+++ b/pkg/plugins/golang/v3/scaffolds/internal/templates/config/prometheus/monitor.go
@@ -53,6 +53,10 @@ spec:
   endpoints:
     - path: /metrics
       port: https
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
   selector:
     matchLabels:
       control-plane: controller-manager
--- a/testdata/project-v3/config/prometheus/monitor.yaml
+++ b/testdata/project-v3/config/prometheus/monitor.yaml
@@ -11,6 +11,10 @@ spec:
   endpoints:
     - path: /metrics
       port: https
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
   selector:
     matchLabels:
       control-plane: controller-manager
--- a/docs/book/src/component-config-tutorial/testdata/project/config/prometheus/monitor.yaml
+++ b/docs/book/src/component-config-tutorial/testdata/project/config/prometheus/monitor.yaml
@@ -11,6 +11,10 @@ spec:
   endpoints:
     - path: /metrics
       port: https
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
   selector:
     matchLabels:
       control-plane: controller-manager
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/prometheus/monitor.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/prometheus/monitor.yaml
@@ -11,6 +11,10 @@ spec:
   endpoints:
     - path: /metrics
       port: https
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
   selector:
     matchLabels:
       control-plane: controller-manager
--- a/docs/book/src/reference/metrics.md
+++ b/docs/book/src/reference/metrics.md
@@ -12,7 +12,7 @@ can be found at `config/rbac/auth_proxy_client_clusterrole.yaml`.
 You will need to grant permissions to your Prometheus server so that it can
 scrape the protected metrics. To achieve that, you can create a
 `clusterRoleBinding` to bind the `clusterRole` to the service account that your
-Prometheus server uses.
+Prometheus server uses. If you are using `kube-prometheus`, this cluster binding already exists.
 
 You can run the following kubectl command to create it. If using kubebuilder
 `<project-prefix>` is the `namePrefix` field in `config/default/kustomization.yaml`.
@@ -26,7 +26,7 @@ kubectl create clusterrolebinding metrics --clusterrole=<project-prefix>-metrics
 Follow the steps below to export the metrics using the Prometheus Operator:
 
 1. Install Prometheus and Prometheus Operator.
-We recommend using [kube-prometheus](https://github.com/coreos/kube-prometheus#installing) 
+We recommend using [kube-prometheus](https://github.com/coreos/kube-prometheus#installing)
 in production if you don't have your own monitoring system.
 If you are just experimenting, you can only install Prometheus and Prometheus Operator.
 2. Uncomment the line `- ../prometheus` in the `config/default/kustomization.yaml`.
@@ -38,21 +38,21 @@ It creates the `ServiceMonitor` resource which enables exporting the metrics.
 ```
 
 Note that, when you install your project in the cluster, it will create the
-`ServiceMonitor` to export the metrics. To check the ServiceMonitor, 
+`ServiceMonitor` to export the metrics. To check the ServiceMonitor,
 run `kubectl get ServiceMonitor -n <project>-system`. See an example:
 
 ```
-$ kubectl get ServiceMonitor -n monitor-system 
+$ kubectl get ServiceMonitor -n monitor-system
 NAME                                         AGE
 monitor-controller-manager-metrics-monitor   2m8s
 ```
 
 Also, notice that the metrics are exported by default through port `8443`. In this way,
-you are able to check the Prometheus metrics in its dashboard. To verify it, search 
-for the metrics exported from the namespace where the project is running 
-`{namespace="<project>-system"}`. See an example: 
+you are able to check the Prometheus metrics in its dashboard. To verify it, search
+for the metrics exported from the namespace where the project is running
+`{namespace="<project>-system"}`. See an example:
 
-<img width="1680" alt="Screenshot 2019-10-02 at 13 07 13" src="https://user-images.githubusercontent.com/7708031/66042888-a497da80-e515-11e9-9d77-d8a9fc1159a5.png">  
+<img width="1680" alt="Screenshot 2019-10-02 at 13 07 13" src="https://user-images.githubusercontent.com/7708031/66042888-a497da80-e515-11e9-9d77-d8a9fc1159a5.png">
 
 ## Publishing Additional Metrics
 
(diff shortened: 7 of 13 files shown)


pkg/plugins/golang/v2/scaffolds/internal/templates/config/prometheus/monitor.go
```diff
@@ -53,6 +53,10 @@ spec:
   endpoints:
     - path: /metrics
       port: https
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
```
estroz, line 59: At this point, no new features should go into go/v2
```suggestion
```

pkg/plugins/golang/v3/scaffolds/internal/templates/config/prometheus/monitor.go
```diff
@@ -53,6 +53,10 @@ spec:
   endpoints:
     - path: /metrics
       port: https
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
```
estroz, line 59: The webhook server already [sets up TLS](https://github.com/kubernetes-sigs/controller-runtime/blob/774f9d490254c861e9f5ca009b3c4b8b5d55fa83/pkg/webhook/server.go#L175). I wonder if that cert/key can be used here.

pkg/plugins/golang/v2/scaffolds/internal/templates/config/prometheus/monitor.go
johanneswuerbach: As the current configuration doesn’t work, it could also be seen as a bug fix, but happy to remove this part.

estroz: If that's true, then it can be left in I suppose.

Adirio: If it is a bugfix (haven't checked that) it should have the 🐛 (`:bug:`) emoji.

docs/book/src/multiversion-tutorial/testdata/project/config/prometheus/monitor.yaml
```diff
@@ -11,6 +11,10 @@ spec:
   endpoints:
     - path: /metrics
       port: https
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
   selector:
```
camilamacedo86, line 18: I do not think that we should move forward here. See the doc: https://book.kubebuilder.io/reference/metrics.html

These metrics are protected by kube-auth-proxy by default if using kubebuilder. Kubebuilder v2.2.0+ scaffold a clusterrole which can be found at config/rbac/auth_proxy_client_clusterrole.yaml.

```
You will need to grant permissions to your Prometheus server so that it can scrape the protected metrics. To achieve that, you can create a clusterRoleBinding to bind the clusterRole to the service account that your Prometheus server uses.

You can run the following kubectl command to create it. If using kubebuilder <project-prefix> is the namePrefix field in config/default/kustomization.yaml.

kubectl create clusterrolebinding metrics --clusterrole=<project-prefix>-metrics-reader --s
```

Did you follow that?

camilamacedo86: /hold

johanneswuerbach: Yes, but I'm not sure how your comment is related.

The proxy is [by default](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v2/config/default/manager_auth_proxy_patch.yaml#L15) exposed only via `https` so scraping using `scheme: http` [will result in a 400 error](https://github.com/golang/go/blob/520f3b72db7befab2028d9a47376267cf2d274a9/src/net/http/server.go#L1844-L1845). Annoyingly this directly implemented in golang, so the rbac proxy doesn't log anything and prometheus doesn't log the response body.

The `bearerTokenFile` is required for prometheus to actually [include a](https://github.com/prometheus-operator/prometheus-operator/blob/da324dd3d75ce7db24c6c55704a2231d1249fec6/pkg/prometheus/promcfg.go#L916) `Authorization` [header](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config).

This header is required so the proxy can actually validate whether the request has the necessary permissions (you get a [401 otherwise, while the missing binding would result in a 403](https://github.com/brancz/kube-rbac-proxy/blob/e4b31758aedb3d29d8f50d615828401a30d28c1e/pkg/proxy/proxy.go#L62))

The `insecureSkipVerify` is required as the rbac proxy uses a self-generate certificate in the current configuration https://github.com/brancz/kube-rbac-proxy/blob/e4b31758aedb3d29d8f50d615828401a30d28c1e/main.go#L261

@estroz proposed to use the webhook certificate https://github.com/kubernetes-sigs/kubebuilder/pull/2065#discussion_r588711493, but this seems out of scope of this PR.

pkg/plugins/golang/v3/scaffolds/internal/templates/config/prometheus/monitor.go
johanneswuerbach: This seems out of scope of this PR, but seems like a good idea otherwise.

pkg/plugins/golang/v2/scaffolds/internal/templates/config/prometheus/monitor.go
johanneswuerbach: I wasn't sure, but as the default configuration just doesn't work I think it makes sense.

docs/book/src/multiversion-tutorial/testdata/project/config/prometheus/monitor.yaml
camilamacedo86: I can be missing something here our misunderstanding. However, see: https://github.com/kubernetes-sigs/kubebuilder/pull/1317/files#diff-20a08dcfd7ad4ed3343d00cf32c7844ce41a2d8d5f5df5b6de56852a510c5517R46 from the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/1317 which via the rbac rules allows accessing metrics behind kube-rbac-proxy. 

Then, see that in the e2e tests that are checked: 

- https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v3/plugin_cluster_test.go#L165-L170
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v3/plugin_cluster_test.go#L272-L303

So, I am trying here understand why in the e2e test we get HTTP 200 from the metrics endpoint and you are facing 400 or 403. What is missing? What has been working our e2e tests and not for you? 

Is it because we are getting the [token](https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v3/plugin_cluster_test.go#L265-L270) to check and doing the handshake and you are not?  

Then, I understand that the default configuration has been allowing Prometheus to get the metrics. See that the problem point out by you was solved in the past with the rbac roles https://github.com/kubernetes-sigs/kubebuilder/pull/1317.

Anyway, shows that also has some difference between using Prometheus Operator instead of kube-prometheus which has been solved with the solution applied here as you point out : https://github.com/kubernetes-sigs/kubebuilder/issues/1253#issuecomment-597640633 

So, I will 

/hold cancel 

However, would be very nice to have a new issue with the steps required to reproduce the scenario for the future we are able to check that. Describing the cluster and Prometheus used, and if the rbac roles for allowing the Prometheus get the metrics were or not used, and what are the steps to check the error.

testdata/project-v3/config/prometheus/monitor.yaml
```diff
@@ -11,6 +11,10 @@ spec:
   endpoints:
     - path: /metrics
       port: https
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
```
camilamacedo86, line 17: Would not it remove the need to get the token in the e2e tests? 

- https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v3/plugin_cluster_test.go#L165-L170
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v3/plugin_cluster_test.go#L272-L303

If yes, I think we need to remove that from the e2e tests to validate your solution as well.

docs/book/src/multiversion-tutorial/testdata/project/config/prometheus/monitor.yaml
Adirio: Actually, Camila's point is pretty sound. If e2e tests are working and this is being tested, we need an issue describing the steps to get this error so that we can reproduce and check that it is an actual bug and not just an error on your side.

johanneswuerbach: Looking at the e2e tests, it seems they only test the service endpoint using curl, but the service monitor itself is not covered. The service monitor instructs how prometheus is supposed to scrape the service endpoint.

The test itself is using https (which is configured via scheme), `-k` which skips the certificate validation and injects a token into the request, which seems to be confirming the service monitor misconfiguration.

I‘m happy to put together a reproduction, but it requires prometheus so might be not something usable for e2e tests.

testdata/project-v3/config/prometheus/monitor.yaml
johanneswuerbach: No, this only instructs how prometheus should scrape the service, but doesn’t change anything about the service itself. As prometheus isn’t used in e2e tests this doesn’t really change anything as this part is not test covered yet.

docs/book/src/multiversion-tutorial/testdata/project/config/prometheus/monitor.yaml
Adirio: > The test itself is using https (which is configured via scheme), `-k` which skips the certificate validation and injects a token into the request, which seems to be confirming the service monitor misconfiguration.

That actually makes sense. I verified it and we are exactly testing it as you describe.

> I‘m happy to put together a reproduction, but it requires prometheus so might be not something usable for e2e tests.

The reproduction we were talking about, at least from my part, was how to reproduce the error you fix here. But changing the `curl` approach for a real e2e test would be pretty good to avoid making the same mistake in the future. Why do you think that requiring prometheus is not suitable for e2e?

camilamacedo86: > I‘m happy to put together a reproduction, but it requires prometheus so might be not something usable for e2e tests.

We install Prometheus in the e2e tests see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v3/plugin_cluster_test.go#L45-L46 

So, I think we need also cover this scenario to be sure that indeed this solution will fix it and we will not break that in the future. Could you help with that?

johanneswuerbach: I'm happy to see what can be done, currently I'm struggling with getting the e2e test suite running on the new M1 MacBook, so an update here might require a bit more time.

Adirio: Go 1.16? Not supported yet. There are some modules related changes (like some commands not tidying up go.mod anymore) that haven't been taking into account in kubebuilder nor any of their dependencies.

estroz: Ah that is unfortunate. If I have some time I can update e2e code after this is merged.

camilamacedo86: Just to highlight: We might be unable to reproduce the issue in the e2e as it now if it is specific only when to use Prometheus Operator instead of kube-prometheus.

camilamacedo86 (approved):
It shows OK since more than one person shows that have this problem and solves it the same way.  We might be unable to reproduce the issue in the e2e as it now if it is specific only when to use Prometheus Operator instead of kube-prometheus.

So, it is fine for me. 

More info: https://github.com/kubernetes-sigs/kubebuilder/pull/2065/files#r589002353

docs/book/src/multiversion-tutorial/testdata/project/config/prometheus/monitor.yaml
Adirio: If you create an issue describing the need of an e2e test that covers this case, I'm ok merging this.

johanneswuerbach: Opened https://github.com/kubernetes-sigs/kubebuilder/issues/2087

johanneswuerbach:
I created the related e2e test ticket https://github.com/kubernetes-sigs/kubebuilder/issues/2087, anything else missing here?

estroz (approved):
Since the metrics endpoint is protected by kube-rbac-proxy, I don't see a reason against merging this.

/lgtm

PR #2060, opened 2021-03-03
⚠️ Plugin phase 1.5 implementation

Adirio:
[Design doc](https://github.com/kubernetes-sigs/kubebuilder/blob/2b464030b6e3f52756e3a1f4c01369ebd58e6dcf/designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md)

It would solve the following issues.
- Closes: #1543
- Closes: #1941
- Closes: #1942

**NOTE**: 
To ensure that all changes proposed here also works with the consumer (SDK), operator-framework/operator-sdk#4581 was created.

Issue #2015, opened 2021-02-13
Extract config/base plugin valid for any language/type from Golang


camilamacedo86:
**What do you want to happen?**

A plugin base which would be responsible to scaffold what is common for any kind of plugin no matter the language and what be integrated with KB and use it's as a consumer. ( basically, it is the hack and config directory since the PROJECT file is scaffold by the CLI ) 

It would allow the community to create any language/type plugin using the same base/config defined by KB in order to ensure its compatibility and maintainability see https://github.com/operator-framework/operator-sdk/issues/2745.Following the requirements/problems that also could be addressed with:

- Make SDK use the real plugin system instead of having to do “hacky” wrappers around the `go.kubebuilder.io` plugin. To SDK Helm/Ansible plugins it would actually mean only implement what which is specifically for its types which means that SDK project would no longer need to copy and paste the config directory boilerplates from kubebuilder in order to do the same. 
- It could allow we init hybrid operators such as; `layout: config/v1,helm/v1,go/v3` which would have the init scaffold of all plugins. See: https://github.com/operator-framework/operator-sdk/issues/670
- Solve the concerns issues raised in the https://github.com/operator-framework/enhancements/pull/44 regarding the risks of the language plugins deviated too much from the upstream base and to keep they maintainable. 

**Is there currently another issue associated with this?**

This issue would be a follow up of #1991 and  #1941

**backwords compatibility**

In order to ensure the backwords compatibility if we call the init command by informing only a plugin that is not the base config, the CLI needs to scaffold the base + the config as it is done currently. E.g `operator-sdk init --plugins=helm/v2` will scaffold the base+helm

camilamacedo86:
Hi @Adirio, 

I tracked the issue regarding the scenario that we spoke about. 

wdyt @estroz @DirectXMan12

camilamacedo86:
The idea proposed regarding the project has a config-gen that also might able to address these requirements. See: https://github.com/kubernetes-sigs/kubebuilder/pull/1831. We probably also might able to replace the base/config with the config-gen when it is stable. c/c @pwittrock

camilamacedo86:
As we discussed today before we decide to move forward with this or not we need first have a better position over #1831:

- shows that @pwittrock like the idea of adding a sample ( project-v3-config-gen in the testdata which would be == project-v3 but using config-gen for we have better clarity of how it will work, what is covered or not already by it and etc)
[...]

PR #2070, opened 2021-03-09
:sparkles: (go/v3) create and bind to a non-default service account

estroz:
As of now, all managers are deployed under the same service account in a namespace. Although this namespace is unique to the manager initially, it is a common admin pattern to deploy multiple managers under the same namespace for security reasons. This pattern becomes insecure given the current one-service-account-for-all approach, since RBAC will conflict.

This PR adds a ServiceAccount (config/rbac/service_account.yaml) and changes the default service account name from default to controller-manager such that a user can specify which service account their manager should be created in. They may change this value by updating files referencing the metadata.name value in service_account.yaml.

pkg/plugins/golang/v3: update all scaffold referencing the default service account with references to the "controller-manager" service account, specified in service_account.yaml.

/kind feature

### DIFF
--- a/test/e2e/v3/plugin_cluster_test.go
+++ b/test/e2e/v3/plugin_cluster_test.go
@@ -69,6 +69,11 @@ var _ = Describe("kubebuilder", func() {
 			})
 
 			It("should generate a runnable project", func() {
+				// go/v3 uses a unqiue-per-project service account name,
+				// while go/v2 still uses "default".
+				tmp := kbc.Kubectl.ServiceAccount
+				kbc.Kubectl.ServiceAccount = "default"
+				defer func() { kbc.Kubectl.ServiceAccount = tmp }()
 				GenerateV2(kbc)
 				Run(kbc)
 			})
@@ -166,7 +171,7 @@ func Run(kbc *utils.TestContext) {
 	_, err = kbc.Kubectl.Command(
 		"create", "clusterrolebinding", fmt.Sprintf("metrics-%s", kbc.TestSuffix),
 		fmt.Sprintf("--clusterrole=e2e-%s-metrics-reader", kbc.TestSuffix),
-		fmt.Sprintf("--serviceaccount=%s:default", kbc.Kubectl.Namespace))
+		fmt.Sprintf("--serviceaccount=%s:%s", kbc.Kubectl.Namespace, kbc.Kubectl.ServiceAccount))
 	ExpectWithOffset(1, err).NotTo(HaveOccurred())
 
 	_ = curlMetrics(kbc)
@@ -263,7 +268,11 @@ func Run(kbc *utils.TestContext) {
 // curlMetrics curl's the /metrics endpoint, returning all logs once a 200 status is returned.
 func curlMetrics(kbc *utils.TestContext) string {
 	By("reading the metrics token")
-	b64Token, err := kbc.Kubectl.Get(true, "secrets", "-o=jsonpath={.items[0].data.token}")
+	// Filter token query by service account in case more than one exists in a namespace.
+	query := fmt.Sprintf(`{.items[?(@.metadata.annotations.kubernetes\.io/service-account\.name=="%s")].data.token}`,
+		kbc.Kubectl.ServiceAccount,
+	)
+	b64Token, err := kbc.Kubectl.Get(true, "secrets", "-o=jsonpath="+query)
 	ExpectWithOffset(2, err).NotTo(HaveOccurred())
 	token, err := base64.StdEncoding.DecodeString(strings.TrimSpace(b64Token))
 	ExpectWithOffset(2, err).NotTo(HaveOccurred())
@@ -271,10 +280,11 @@ func curlMetrics(kbc *utils.TestContext) string {
 
 	By("creating a curl pod")
 	cmdOpts := []string{
-		"run", "--generator=run-pod/v1", "curl", "--image=curlimages/curl:7.68.0", "--restart=OnFailure", "--",
+		"run", "--generator=run-pod/v1", "curl", "--image=curlimages/curl:7.68.0", "--restart=OnFailure",
+		"--serviceaccount=" + kbc.Kubectl.ServiceAccount, "--",
 		"curl", "-v", "-k", "-H", fmt.Sprintf(`Authorization: Bearer %s`, token),
-		fmt.Sprintf("https://e2e-%v-controller-manager-metrics-service.e2e-%v-system.svc:8443/metrics",
-			kbc.TestSuffix, kbc.TestSuffix),
+		fmt.Sprintf("https://e2e-%s-controller-manager-metrics-service.%s.svc:8443/metrics",
+			kbc.TestSuffix, kbc.Kubectl.Namespace),
 	}
 	_, err = kbc.Kubectl.CommandInNamespace(cmdOpts...)
 	ExpectWithOffset(2, err).NotTo(HaveOccurred())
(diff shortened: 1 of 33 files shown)


pkg/plugins/golang/v3/scaffolds/internal/templates/config/rbac/auth_proxy_role_binding.go
```diff
@@ -50,6 +50,6 @@ roleRef:
   name: proxy-role
 subjects:
 - kind: ServiceAccount
-  name: default
+  name: controller-manager
```
camilamacedo86, line 53: Why do we need to change the default value? Can we not only address the change without change the names of what is scaffold currently?

test/e2e/v3/plugin_cluster_test.go
```diff
@@ -166,7 +166,7 @@ func Run(kbc *utils.TestContext) {
 	_, err = kbc.Kubectl.Command(
 		"create", "clusterrolebinding", fmt.Sprintf("metrics-%s", kbc.TestSuffix),
 		fmt.Sprintf("--clusterrole=e2e-%s-metrics-reader", kbc.TestSuffix),
-		fmt.Sprintf("--serviceaccount=%s:default", kbc.Kubectl.Namespace))
+		fmt.Sprintf("--serviceaccount=%s:e2e-%s-controller-manager", kbc.Kubectl.Namespace, kbc.TestSuffix))
```
camilamacedo86, line 169: same here. It shows not required. I'd prefer as it is in go/v2, ansible/helm. 
It has no reason to be changed.

pkg/plugins/golang/v3/scaffolds/internal/templates/config/rbac/auth_proxy_role_binding.go
estroz: If the manager's namespace is changed to a one shared by more than one access-controlled object, but the service account remains as "default", RBAC already bound to "default" will be applied to the manager, on top of the manager's RBAC. This is insecure. A more secure default is to create a manager-specific service account so RBAC is never polluted.

test/e2e/v3/plugin_cluster_test.go
```diff
@@ -273,8 +273,8 @@ func curlMetrics(kbc *utils.TestContext) string {
 	cmdOpts := []string{
 		"run", "--generator=run-pod/v1", "curl", "--image=curlimages/curl:7.68.0", "--restart=OnFailure", "--",
 		"curl", "-v", "-k", "-H", fmt.Sprintf(`Authorization: Bearer %s`, token),
-		fmt.Sprintf("https://e2e-%v-controller-manager-metrics-service.e2e-%v-system.svc:8443/metrics",
-			kbc.TestSuffix, kbc.TestSuffix),
+		fmt.Sprintf("https://e2e-%[1]s-controller-manager-metrics-service.e2e-%[1]s-system.svc:8443/metrics",
```
camilamacedo86, line 276: This shows that it not passing in the CI because of this change. 
I think it is out of the scope and if you just revert this one all will work as expected.

estroz: This change is cosmetic. In any case, the namespace can be used directly as a subdomain of the SVC.

test/e2e/v3/plugin_cluster_test.go
```diff
@@ -166,7 +171,7 @@ func Run(kbc *utils.TestContext) {
 	_, err = kbc.Kubectl.Command(
 		"create", "clusterrolebinding", fmt.Sprintf("metrics-%s", kbc.TestSuffix),
 		fmt.Sprintf("--clusterrole=e2e-%s-metrics-reader", kbc.TestSuffix),
-		fmt.Sprintf("--serviceaccount=%s:default", kbc.Kubectl.Namespace))
+		fmt.Sprintf("--serviceaccount=%s:%s", kbc.Kubectl.Namespace, kbc.Kubectl.ServiceAccount))
```
camilamacedo86, line 174: This is the kind of PR that will require changes in the SDK side such as: 

- In the go e2e tests
- Apply the same feature for ansible/helm
- Change the e2e test for ansible and helm. 

So, instead of manually change ansible and helm regards the changes in the config then, I'd like to see if we can first:

- merge 1.5 https://github.com/kubernetes-sigs/kubebuilder/pull/2060 / sdk alignment operator-framework/operator-sdk#4581.
- then push a pr in kb to centralize the code https://github.com/kubernetes-sigs/kubebuilder/issues/2015 and bump SDK with where Ansible/Helm/GO will use all the same code and let us know to close the issue: https://github.com/operator-framework/operator-sdk/issues/4542
- For them, we get this merged. Since it will mean change the e2e test for all types on the SDK side as well. 

See that how much changes like this one we merge before that, harder will be for us to align and keep maintained the SDK side. 

c/c @Adirio @jmrodri

estroz: The addition of a service account is straightforward, improves default security, and is an often requested feature by users. The PR’s/steps you linked are unrelated and need a lot of review. Other PR’s aren’t being blocked by them so the argument that this one should doesn’t hold water.

camilamacedo86: So, are you planning to get it merged and as follow up; 
- bump kb in sdk
- then, to do the config of  helm and ansible plugin
- then, to do changes in the tests for both + go/v2 and go/v3?

estroz: Yep!

estroz: Is that true though? Why would merging this now prevent ansible/helm plugins from receiving the same changes later through phase 1.5 or some other machination mentioned above? Wouldn't they just get this feature for free then?

Adirio: It would definetely work the way you mention. I just meant that once phase 1.5 is implemented, the process will be much easier as changing them in one place will update all (after a kubebuilder bump). I think the point of this conversation was just to show how right now they need to be updated in 3 or 4 different places while with 1.5 would only require to be updated in one place. But by no means does that mean that it is blocked, just that it requires more work as it is right now.

estroz: Agreed. I guess my point that this PR adds a feature that doesn’t necessarily need to be added to other plugins _right now_ wasn’t clear. For example, neither of the operator-sdk’s own plugins support component configs yet the Go plugin does, and that disparity was acceptable at the time.

Also, a PR here shouldn’t be blocked by downstream stuff.

camilamacedo86: My comment here was we will spend a lot of effort which shows unnecessary if we agree to merge the open prs. Then, if it is not urgent(can wait until Tue) why spend all this effort at all? However, it is fine.

/lgtm 

/hold 
to let you get the chance to merge when you wish.

Adirio: If I'm not mistaken this should be merged after #2071, right? It seems to have the same changes also inside here.

By the way this also has my LGTM. Even if phase 1.5 was implemented, we would have to do this here, so that doesn't make a difference.

estroz:
@camilamacedo86 needs another `/lgtm` when you get a chance.

Adirio:
@pwittrock I think you may want to consider this for `config-gen`

pwittrock:
@Adirio 

> @pwittrock I think you may want to consider this for config-gen

Thx! #2085

PR #2071, opened 2021-03-09
:seedling: test/e2e: export KIND_CLUSTER so local tests get correct cluster name

estroz:
Local e2e tests do not receive the correct cluster name as KIND_CLUSTER is not set.

- test/e2e: export KIND_CLUSTER so local tests get correct cluster name


test/e2e/setup.sh
```diff
@@ -35,10 +37,11 @@ function create_cluster {
 #
```
Adirio, line 37: NIT: The description above this line (GitHub doesn't allow me to comment here) directly refers to setting a variable. Maybe update it so that it reflects that it needs to be EXPORTed.

Adirio (approved):

PR #2072, opened 2021-03-09
🌱 Fix the config registry public API

Adirio:
Remove the unexported `contructorFunc` alias to prevent an exported function have a parameter of an unexported type.

This type was being shown in the [public API](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v3/pkg/config#Register) when it was unexported.

### DIFF
--- a/pkg/config/registry.go
+++ b/pkg/config/registry.go
@@ -16,14 +16,12 @@ limitations under the License.
 
 package config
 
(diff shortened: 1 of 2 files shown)


camilamacedo86 (approved):
small fix/improvement does not effect or change anything
So, it shows fine 👍 
/lgtm

PR #2074, opened 2021-03-09
🌱 Move the internal machinery utils to appropiate exported packages

Adirio:
- Go version checks and go-modules-based repo discovery moved to pkg/plugins/golang
- Command execution and user prompt moved to pkg/plugin/util

Partial: #2036


camilamacedo86 (approved):
We are making public + adding tests. 
So, it will not affect other places and just bring the option for them to use it 👍

pkg/plugins/golang/go_version.go
```diff
@@ -54,24 +144,13 @@ func fetchAndCheckGoVersion() error {
...
-	if err != nil {
-		return fmt.Errorf("error parsing minor version '%s': %s", m[2], err)
+	var version goVersion
+	if err := version.parse(verStr); err != nil {
+		return err
 	}
 
-	if major < 1 || minor < 13 {
-		return fmt.Errorf("requires version >= 1.13")
+	if version.compare(go115) < 0 || version.compare(go116alpha1) >= 0 {
```
camilamacedo86, line 152: We cannot change here. If the person is using go 1.13 it still working.

Adirio: I'm pretty sure that go1.13 doesn't work. We didnt update the requirement here but I'm pretty sure it breaks.

Adirio: I will go back to 1.13 and track this in a separate PR.

camilamacedo86 (approved):
It does not change the behaviour.
Only add more test and allow it to be used as agreed in the bug triage meetings. 

/lgtm

PR #2080, opened 2021-03-10
⚠️ Inject filesystem from CLI instead of creating several per command

Adirio:
A new filesystem abstraction is being created in multiple places. This PRs creates a single filesystem abstraction in the CLI, and passes/injects it to any machinery that requires it.

Technically speaking this is a brekaing change, because it modifies the `Subcommand` interface in `pkg/plugin/interfaces.go`. This, and the corresponding signature update for the plugins, is the only breaking change in this PR. It is required to expose both the scaffolding machinery (#2036) and the config file machinery (#2056).

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -21,12 +21,14 @@ import (
 	"os"
 	"strings"
 
+	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
 	internalconfig "sigs.k8s.io/kubebuilder/v3/pkg/cli/internal/config"
 	"sigs.k8s.io/kubebuilder/v3/pkg/config"
 	cfgv3 "sigs.k8s.io/kubebuilder/v3/pkg/config/v3"
+	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
 )
 
@@ -88,6 +90,9 @@ type CLI struct { //nolint:maligned
 
 	// Root command.
 	cmd *cobra.Command
+
+	// Underlying fs
+	fs machinery.Filesystem
 }
 
 // New creates a new CLI instance.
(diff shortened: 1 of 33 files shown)


camilamacedo86 (approved):
It is a small change which I agree with 🥇 
However, as it is a breaking change would be nice to get a second thought as well. 
Could you please check if another person can review and is ok with either?

pkg/cli/cli.go
```diff
@@ -131,6 +135,7 @@ func newCLI(options ...Option) (*CLI, error) {
 		defaultProjectVersion: cfgv3.Version,
 		defaultPlugins:        make(map[config.Version][]string),
 		plugins:               make(map[string]plugin.Plugin),
+		fs:                    afero.NewOsFs(),
```
estroz, line 138: Right now there's no way for a user to set `fs`, so perhaps an Option to do so?

pkg/plugin/interfaces.go
```diff
@@ -53,7 +54,7 @@ type Subcommand interface {
 	// command line flags.
 	BindFlags(*pflag.FlagSet)
 	// Run runs the subcommand.
-	Run() error
+	Run(fs afero.Fs) error
```
estroz, line 57: I'm still not convinced `afero.Fs` should be exposed in the signature. How about
```suggestion
	Run(RunOptions) error
```
Where
```go
type RunOptions struct {
	FS afero.Fs
}
```
So fields can be easily deprecated/added without changing the interface itself?

Adirio: Being `RunOptions` and exported type, and `FS` an exported field, we would still expose `afero.Fs`, right?

pkg/cli/cli.go
Adirio: Do we really want to expose that? I think that we can internally use that field to set a `MemMapFs` or even achieve atomicity with a `CopyOnReadFs`, but what advantage would a CLI tool developer get from exposing that field?

Either way, creating an `Option` would be incremental, and therefore non-breaking, so I think we can leave that for a follow up after we discuss it.

pkg/plugin/interfaces.go
estroz: Yes, but you could deprecate that field if it was later decided that another fs abstraction should be used without causing a breaking change.

pkg/cli/cli.go
estroz: True that wouldn't be breaking. Sounds good.

pkg/plugin/interfaces.go
estroz: I suppose you could add parameters and not remove `afero.Fs` as to not make a breaking change, but that's uglier.

PR #2081, opened 2021-03-10
🐛 fix: ignore non-go files during Docker build

### LINKED ISSUE #2073: Scaffolded `.dockerignore` does not do what it claims
When running `make docker-build` for an operator scaffolded with kubebuilder v3, the entire directory is included in the Docker build context.

I expected only Go source files to be included in the Docker build context. Here is the scaffolded `.dockerignore` file:

```
# More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
# Ignore all files which are not go type
!**/*.go
!**/*.mod
!**/*.sum
```

To reproduce, build a dummy Docker image containing the entire build context:

```bash
docker image build -t build-context -f - . <<EOF
FROM busybox
COPY . /build-context
WORKDIR /build-context
CMD find .
EOF
```

Now run it:

```bash
docker container run --rm build-context
```

Or explore the images contents:

```bash
docker container run --rm -it build-context /bin/sh
```

I've traced the issue to this source code:
https://github.com/kubernetes-sigs/kubebuilder/blob/2018a4f1c8dbcff9b868b80af03c7879c707edbe/pkg/plugins/golang/v3/scaffolds/internal/templates/dockerignore.go#L41-L46

For reference, this is my Docker version:

```bash
Client: Docker Engine - Community
 Cloud integration: 1.0.7
 Version:           20.10.2
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        2291f61
 Built:             Mon Dec 28 16:12:42 2020
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.2
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       8891c58
  Built:            Mon Dec 28 16:15:28 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.3
  GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

```

/kind bug

busser:
This change ensures that the scaffolded `.dockerignore` file makes Docker ignore non-Go files.

This avoids large files like test binaries being unnecessarily included in Docker's build context.

Fixes #2073.

### DIFF
--- a/pkg/plugins/golang/v3/scaffolds/internal/templates/dockerignore.go
+++ b/pkg/plugins/golang/v3/scaffolds/internal/templates/dockerignore.go
@@ -39,8 +39,7 @@ func (f *DockerIgnore) SetTemplateDefaults() error {
 }
 
 const dockerignorefileTemplate = `# More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
-# Ignore all files which are not go type
-!**/*.go
-!**/*.mod
-!**/*.sum
+# Ignore build and test binaries.
+bin/
+testbin/
 `
--- a/docs/book/src/component-config-tutorial/testdata/project/.dockerignore
+++ b/docs/book/src/component-config-tutorial/testdata/project/.dockerignore
@@ -1,5 +1,4 @@
 # More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
-# Ignore all files which are not go type
-!**/*.go
-!**/*.mod
-!**/*.sum
+# Ignore build and test binaries.
+bin/
+testbin/
--- a/docs/book/src/cronjob-tutorial/testdata/project/.dockerignore
+++ b/docs/book/src/cronjob-tutorial/testdata/project/.dockerignore
@@ -1,5 +1,4 @@
 # More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
-# Ignore all files which are not go type
-!**/*.go
-!**/*.mod
-!**/*.sum
+# Ignore build and test binaries.
+bin/
+testbin/
--- a/docs/book/src/multiversion-tutorial/testdata/project/.dockerignore
+++ b/docs/book/src/multiversion-tutorial/testdata/project/.dockerignore
@@ -1,5 +1,4 @@
 # More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
-# Ignore all files which are not go type
-!**/*.go
-!**/*.mod
-!**/*.sum
+# Ignore build and test binaries.
+bin/
+testbin/
--- a/testdata/project-v3-addon/.dockerignore
+++ b/testdata/project-v3-addon/.dockerignore
@@ -1,5 +1,4 @@
 # More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
-# Ignore all files which are not go type
-!**/*.go
-!**/*.mod
-!**/*.sum
+# Ignore build and test binaries.
+bin/
+testbin/
--- a/testdata/project-v3-config/.dockerignore
+++ b/testdata/project-v3-config/.dockerignore
@@ -1,5 +1,4 @@
 # More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
-# Ignore all files which are not go type
-!**/*.go
-!**/*.mod
-!**/*.sum
+# Ignore build and test binaries.
+bin/
+testbin/
--- a/testdata/project-v3-multigroup/.dockerignore
+++ b/testdata/project-v3-multigroup/.dockerignore
@@ -1,5 +1,4 @@
 # More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
-# Ignore all files which are not go type
-!**/*.go
-!**/*.mod
-!**/*.sum
+# Ignore build and test binaries.
+bin/
+testbin/
--- a/testdata/project-v3/.dockerignore
+++ b/testdata/project-v3/.dockerignore
@@ -1,5 +1,4 @@
 # More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
-# Ignore all files which are not go type
-!**/*.go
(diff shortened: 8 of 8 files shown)


pkg/plugins/golang/v3/scaffolds/internal/templates/dockerignore.go
```diff
@@ -40,7 +40,10 @@ func (f *DockerIgnore) SetTemplateDefaults() error {
...
 const dockerignorefileTemplate = `# More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
 # Ignore all files which are not go type
-!**/*.go
-!**/*.mod
-!**/*.sum
+*
+!go.mod
+!go.sum
+!main.go
+!controllers/**/*.go
```
camilamacedo86, line 47: We need to check if it will work when is multigroup when is not since it means that it can be:
https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v3-multigroup/controllers
OR
https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v3/controllers

busser: Good call.

I just tested it and it does not work. Further investigation lead me to this issue: https://github.com/moby/moby/issues/30018. Basically, we can't write a whitelist-style `.dockerignore` file.

Should we opt for a different strategy, like ignoring directories that are known to contain large files that are unnecessary when building a Docker image for the operator? The `.dockerignore` file would look like this:

```
# More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
# Ignore build and test binaries.
bin/
testbin/
```

camilamacedo86: Hi @busser, 

If the previous suggestion works when is not multigroup then, see that we can:

- (option A) duplicate that to add the respective paths for multigroup by default as well
- (option B) we can change the file as we change  the dockerfile when we enable the multi-group support
- (option C) we can check a syntax that will ignore both paths.(I think it is the best solution)

camilamacedo86: see : 

> Beyond Go’s filepath.Match rules, Docker also supports a special wildcard string ** that matches any number of directories (including zero). For example, **/*.go will exclude all files that end with .go that are found in all directories, including the root of the build context.

https://docs.docker.com/engine/reference/builder/#dockerignore-file

So, according to the docker documentation `!**/*.go` is correct which means that we are generating the file right.

busser: Hi @camilamacedo86,

The pattern we currently use, `!**/*.go` should work according to the Docker docs, but in fact doesn't. My understanding is that we are facing a bug in how Docker handles `.dockerignore` logic. This issue (https://github.com/moby/moby/issues/30018) reports a three year-old bug in Docker: when a pattern starts with `!`, wildcard strings like `*` and `**` don't work for directories.

I don't believe this will ever be fixed. The Docker team's stated long-term goal is for `.dockerignore` to have the same behaviour as `.gitignore`. I found this section in the [`.gitignore` specification](https://git-scm.com/docs/gitignore#_pattern_format):

> It is not possible to re-include a file if a parent directory of that file is excluded. Git doesn’t list excluded directories for performance reasons, so any patterns on contained files have no effect, no matter where they are defined.

So I guess that we won't ever be able to exclude the entire working directory except for GO files... 😞

busser: In its current state, the `.dockerignore` file scaffolded by Kubebuilder ignores nothing and includes the entire directory in the Docker build context. Most of the files are just code and are relatively small. However, the Kubebuilder workflow adds large binaries to the `bin/` and `testbin/` directories. I would recommend ignoring these directories by default and letting the user add more paths to their `.dockerignore` based on their specific needs.

What do you think?

camilamacedo86: Would not the following configuration solve the need for multi-group and not multi-group layout projects?

```
!go.mod
!go.sum
!main.go
!controllers/**/*.go
!api/**/*.go
!controllers/*.go
!api/*.go
```

camilamacedo86: I will close this one to make it simpler other follow up.

Adirio: What about the `config` directory? We should probably add a marker so that plugins can add their own exceptions, and that way the config plugin would be the one adding that line.

busser: @camilamacedo86 I've update the pull request to ignore the build and test binaries, as decided above 👍

busser: @Adirio How would you go about doing this? I don't mind doing the implementation myself, but would need some guidance 🙂

camilamacedo86 (approved):
/ok-to-test

It seems ok for me in order to solve the issue. 

@Adirio makes the following suggestion:

> What about the config directory? We should probably add a marker so that plugins can add their own exceptions, and that way the config plugin would be the one adding that line.

That is fine but seems an RFE. So, @Adirio wdyt about we move forward here and then raise that as an issue for we are able to track and work on that too?

camilamacedo86 (approved):
/lgtm
I think it is fine. 
@Adirio please feel free to raise any issue if you see that it requires a follow up for improvements.

PR #2082, opened 2021-03-10
⚠️ Export scaffolding machinery

Adirio:
It is technically a breaking change (⚠️) because some errors that were previously placed in `pkg/model` and `pkg/model/file` have been moved to `pkg/machinery` and no longer expose `NewXxxxxxError` and `IsXxxxxxxError` functions, they expose themselves to use `errors.As` and `errors.Is`. It also removes the `NewUniverse` method and the `WithXxxxx` options are now part of the `NewScaffold` method, instead of having to pass a universe to every `Execute` call.

Closes: #2036
Blocked by: #2080


camilamacedo86 (approved):
The change which was significant and was discussed in the meeting is: https://github.com/kubernetes-sigs/kubebuilder/pull/2080. (which was approved and merged by Eric, who raised the concerns. )

This one is the only flag as breaking change because is moving things around and changing the names. However, the same feature options still in place. Also, it shows very small changes which means that shows fine we move forward here.

PS.: We also agree in the meetings/issues to export the machinery to remove this duplication from the consumer (SDK)

/lgtm

pkg/machinery/scaffold.go
```diff
@@ -41,32 +49,96 @@ var options = imports.Options{
 }
 
 // Scaffold uses templates to scaffold new files
-type Scaffold interface {
```
estroz, line 44: Why was this changed to a struct?

pkg/machinery/scaffold.go
```diff
@@ -375,16 +445,29 @@ func (s scaffold) writeFile(f *file.File) error {
...
+
+	// Create or truncate the file
+	writer, err := s.fs.OpenFile(f.Path, createOrUpdate, s.filePerm)
 	if err != nil {
-		return err
+		return CreateFileError{err}
 	}
+	defer func() {
+		if closeErr := writer.Close(); err == nil && closeErr != nil {
+			err = CloseFileError{err}
```
estroz, line 464: This overwrites `err` in case `Write()` fails. Should be a `log.Error(closeErr)` instead.

pkg/machinery/scaffold.go
Adirio: Because we were exporting a `Option` that was a function on an unexported struct.

pkg/machinery/scaffold.go
Adirio: We check that `err == nil`, so we only overwrite in case there was no error.

estroz: Right, missed that bit. What if `err != nil && closeErr != nil`?

pkg/machinery/scaffold.go
estroz: You can do `type ScaffoldOption func(*scaffold)` since the predefined options have access to the internal type.

estroz: I would prefer ^ but it's an unimportant design decision.

pkg/machinery/scaffold.go
Adirio: `err` is reported as it happened before. Same as it was before.

pkg/machinery/scaffold.go
Adirio: That means that we are sporting `ScaffoldOption` which has a unexported parameter. I avoided this when able.

estroz: That's a common practice. The set of options is fixed because `Scaffold`'s fields are unexported anyway.

Adirio: Oh, yes, I know. The result is exactly the same. But there was not added value by the interface either. And this way we don't have a unexported type in pkg.go.dev.

PR #2083, opened 2021-03-10
🌱 Export config machinery

Adirio:
I don't think that this PR has any breaking changes

Closes: #2056
Blocked by: #2080


pkg/cli/cmd_helpers.go
```diff
@@ -46,18 +47,39 @@ func errCmdFunc(err error) func(*cobra.Command, []string) error {
...
+// and injects it into the subcommand
+func preRunECmdFunc(subcmd plugin.Subcommand, cfg store.Store, msg string) func(*cobra.Command, []string) error {
 	return func(*cobra.Command, []string) error {
-		if err := subcommand.Run(fs); err != nil {
+		err := cfg.Load()
+		if os.IsNotExist(err) {
+			return fmt.Errorf("%s: unable to find configuration file, project must be initialized", msg)
+		} else if err != nil {
+			return fmt.Errorf("%s: unable to load configuration file: %w", msg, err)
+		}
```
camilamacedo86, line 59: I am only concern here. 
SDK has commands that should work when no PROJECT file is found. So, would it break this behaviour?

pkg/config/v3/config.go
```diff
@@ -47,12 +47,11 @@ type cfg struct {
 	Resources []resource.Resource `json:"resources,omitempty"`
 
 	// Plugins
-	Plugins PluginConfigs `json:"plugins,omitempty"`
+	Plugins pluginConfigs `json:"plugins,omitempty"`
```
camilamacedo86, line 50: Why we cannot export the PluginConfigs here?

pkg/cli/cmd_helpers.go
Adirio: This is only applied to `create api`, `create webhook` and `edit` subcommands. Kubebuilder also has commands that should work without a project configuration file: `init`, `version`, `completion`, ...

pkg/config/v3/config.go
Adirio: It doesn't make much sense to export it because is something expecific of v3 implementation of the config file. Previously it was exported because it was needed by the `pkg/cli/internal/config` package that has been replaced here and is no longer needed outside of this package.

pkg/config/store/interface.go
```diff
@@ -0,0 +1,38 @@
...
+*/
+
+package store
+
+import (
+	"sigs.k8s.io/kubebuilder/v3/pkg/config"
+)
+
+// Store represents a persistence backend for config.Config
+type Store interface {
```
camilamacedo86, line 24: I am not sure about this name. We ought to use names that represent the object and not the action. 
The store shows an interface to work with the configuration of the plugins. Why not a better name here such as PluginConfig?

pkg/config/store/yaml/store.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+const (
+	// DefaultPath is the default path for the configuration file
+	DefaultPath = "PROJECT"
+)
+
+// yamlStore implements store.Store using a YAML file as the storage backend
+// The key is translated into the YAML file path
+type yamlStore struct {
```
camilamacedo86, line 38: I do not think that the structure/object should have in its name the type of file used by it. 
Why need yamlStore and Store? Why the yaml stuff cannot be attributes of Store?

pkg/cli/cmd_helpers.go
camilamacedo86: Could we please comment on that in the code? It makes it easier for us to understand it in the future and keep the project maintained. I mean, could you please add a comment when/why that is called.

pkg/config/v3/config.go
camilamacedo86: So, I understand that we will be able to read/update and etc it by what you are calling here as yamlStore and store right?

pkg/config/store/interface.go
Adirio: `Store` is also a noun and is inpired on `React` and similar frameworks where a `Store` is used to abstract the persistence layer for some state (in our case `Config` is the state we are persisting). It would be the equivalent to how `Factory` abstracts the creation of new objects.

This said, if a better name is presented I would change this. However, `PluginConfig` or `PluginDataConfig` are not good names as this is the persistence layer of the project configuration, it has no direct relation with plugins (plugin configuration is also stored inside the project configuration but thats not all what it is stored there).

pkg/config/store/yaml/store.go
Adirio: `Store` is an interface so that in a future, we can use other persistence options. For example, we may want to store the project configuration in a database, or in JSON, or maybe base64 encoded so that users do not modify it, ... In this case, we would create a `jsonStore` or equivalent struct that would implement `Store` and we would allow the cli to choose between them somehow (probably a `WithStoreBackend` or similar option to `NewCLI` so that each third-party command can choose which one they use).

camilamacedo86: json/yaml/txt shows values of the attribute format of some structure such as  PluginConfigData.

camilamacedo86: I do not see it as a rfe, however, it is fine for now.

pkg/config/store/interface.go
camilamacedo86: Thank you for the explanation. I did not like too much the approach. However, it is fine. It addresses our requirements and we can also review that in the future if/when required. thank you for the explations.

camilamacedo86 (approved):
The blocked pr #2080 is merged. It is not a breaking change and it shows currently a requirement for we better address 1.5 plugin implementation and remove its duplication on SDK side. 

Also, it was discussed in the meetings and no objections were raised.  

So, I am ok with this one. Thank you for the collab 🥇 

/lgtm

PR #2084, opened 2021-03-11
:sparkles: (go/v3) implement scaffolding hub and spoke interfaces for…

varshaprasad96:
This PR is a follow a replacement of #1664.

This commit scaffolds hub and spoke interfaces for conversion webhooks.
With this change, users can either specify `--conversion` flag and implement
the interface on their own, or specify `--hub` and `--spoke` versions which
will scaffold the skeleton code for both the interfaces.

Example usage:
`kubebuilder create webhook --group crew --version v1 --kind FirstKind --conversion --hub v1 --spoke v2`

1. FirstKind_conversion.go scaffolded inside api/v1/firstmate_webhook.go having conversion.Hub interface.
2. FirstKind_conversion.go scaffolded inside api/v2/firstmate_webhook.go having method signatures to implement conversion.convertible interface.
3. Webhook registered in main.go with version v1.

The previous behavior is also preserved. If the `--conversion` flag is set, then the command registers the webhook in `main.go` and outputs a message stating:
`Webhook server has been set up for you. You need to implement the conversion.Hub and conversion.Convertible interfaces for your CRD types. You can also specify hub and spoke versions in the command to scaffold Convertible and Hub interfaces`


pkg/model/resource/resource.go
```diff
@@ -119,6 +119,47 @@ func (r Resource) HasConversionWebhook() bool {
 	return r.Webhooks != nil && r.Webhooks.Conversion
 }
 
+// HasConversionFlags checks if conversion webhook needs to be scaffolded. Either the
+// conversion flag has to be set or both hub and spoke versions should be specified.
```
Adirio, line 123: Flags should be completely decoupled from the `Resource` model.

Adirio (changes requested):
I was reviewing this PR and I don't see any reason to have a `--spoke` flag as it is always the same as the `--version`, right. Correct me if I'm mistaken but the flow should be something like:

```
kubebuilder init
kubebuilder create api --group crew --version v1 --kind FirstMate --resource
kubebuilder create api --group crew --version v2 --kind FirstMate --resource
kubebuilder create webhook --group crew --version v2 --kind FirstMate --conversion --hub v1
```
This would create the `ConvertTo` and `ConvertFrom` methods for v2, with v1 being the hub, right?
I think there is a lot of complexity for having both flags that is not needed.

varshaprasad96:
@Adirio, the reason for having 2 flags was that the hub version need not be explicitly the one mentioned in `--version` flag. I could also have `v2` to be the hub, and `v1` to be the spoke. The version specified with `--version` flag would be registered in `main.go` (ex: [here](https://github.com/kubernetes-sigs/kubebuilder/blob/e3e3658e5664e4c5e282923a55f38ca4352feacf/testdata/project-v3/main.go#L133-L135)). Another scenario could be there are 3 versions of APIs, and I would like to implement conversion between just 2, (ie) say `v1` has validating WH, `v2` and `v3` have conversion WH. 

The other way around of making `--version` flag optional can be possible, I suppose. If there is conversion WH between `v1` and `v2`, it shouldn't matter which one we specify in main.go to register. Ill check that out.

pkg/model/resource/webhooks.go
```diff
@@ -77,10 +83,21 @@ func (webhooks *Webhooks) Update(other *Webhooks) error {
...
+	}
+
 	return nil
 }
 
 // IsEmpty returns if the Webhooks' fields all contain zero-values.
 func (webhooks Webhooks) IsEmpty() bool {
-	return webhooks.WebhookVersion == "" && !webhooks.Defaulting && !webhooks.Validation && !webhooks.Conversion
+	return webhooks.WebhookVersion == "" && !webhooks.Defaulting && !webhooks.Validation && !webhooks.Conversion &&
+		(webhooks.Hub == "" || webhooks.Spoke == "")
```
estroz, line 102:
```suggestion
	return webhooks == (Webhooks{})
```

pkg/plugins/golang/options.go
```diff
@@ -158,13 +160,15 @@ func (opts Options) NewResource(c newconfig.Config) resource.Resource {
 		res.API = &resource.API{}
 	}
 
-	if opts.DoDefaulting || opts.DoValidation || opts.DoConversion {
+	if opts.DoDefaulting || opts.DoValidation || opts.DoConversion || (len(opts.Spoke) != 0 && len(opts.Hub) != 0) {
```
estroz, line 163: `Do{X}` must be true for this branch to execute, regardless of whether spoke/hub are set.
```suggestion
	if opts.DoDefaulting || opts.DoValidation || opts.DoConversion {
```

pkg/plugins/golang/v3/scaffolds/internal/templates/api/conversion.go
```diff
@@ -0,0 +1,108 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
estroz, line 2:
```suggestion
Copyright 2021 The Kubernetes Authors.
```

estroz (changes requested):
Some cosmetic nits, and I agree with @Adirio. Why not make `--hub` and `--spoke` mutually exclusive? One of the two must be specified in `--version` since you're targeting a specific GVK for one of the two.

I'd also like to see some validation for existence of the hub/spoke package, since a non-existent package will cause compilation errors.

pkg/plugins/golang/options.go
estroz: It may also be worth deprecating `--conversion`, since `--hub`/`--spoke` imply conversion.

Adirio: In this sense I would say that I would probably implicitly set the `doConversion` variable in the `Validate` subcommand method or `PreScaffold` in phase 1.5.

Adirio:
> I'd also like to see some validation for existence of the hub/spoke package, since a non-existent package will cause compilation errors.

We talked a bit offline and we were going in the same direction. A `GetVersionsForResource(resource.GVK) []resource.Resource` method added to `Config` will allow to retrieve the related resources. Also, removing completely the `--spoke` flag was mentioned: `k9r create webhook --version v2 --hub v1 ...` will create the `Hub` method for `v1` and the `ConvertTo` and `ConvertFrom` for `v2`. The `--spoke` flag accepting different versions that `--version` was a biut dangerous, you may be creating a validation webhook for v2 and in the same command a conversion webhook from v3 to v1.

pkg/model/resource/webhooks.go
```diff
@@ -33,6 +33,10 @@ type Webhooks struct {
 
 	// Conversion specifies if a conversion webhook is associated to the resource.
 	Conversion bool `json:"conversion,omitempty"`
+
+	// Spoke contains the spoke versions associated with the resource. If this is not
+	// empty, then it means that this resource is the hub
+	Spokes []string `json:"spokes,omitempty"`
```
varshaprasad96, line 39: @Adirio , we had spoken about having resource store the hub version. While implementing it, I realized that in that case, we would have to fetch the spoke resource and update it every time. Instead, we could have the resource store the list of spokes. We anyway update the resource while scaffolding, we would just have to add the new spoke version to the list of spokes which the resource holds. One drawback which I thought this implementation could have, is when we allow multiple hubs in the project (we would have to restrict a version to be a spoke to only one resource). Validation in this case may be a little round about, but I suppose that is still a static validation would still be better than updating the resource every time. Does that sound fine?

varshaprasad96:
@Adirio @estroz - Have modified the implementation, to include only the spoke flag. The version specified in the `--version` flag would be the hub. Example command would be:
`kubebuilder create webhook --group crew --version v1 --kind FirstKind --conversion --spoke v2`.

In the implementation, the resource stored in the plugin config holds the list of spokes. We validate if the hub has been scaffolded previously or not, and if already present we skip and scaffold just the spokes. Hence, the following sequence of command still hold good:
```sh
kubebuilder create webhook --group crew --version v1 --kind FirstKind --conversion --spoke v2
kubebuilder create webhook --group crew --version v1 --kind FirstKind --conversion --spoke v3
```

pkg/plugins/golang/options.go
```diff
@@ -91,7 +98,7 @@ func (opts Options) UpdateResource(res *resource.Resource, c config.Config) {
 		res.Controller = true
 	}
 
-	if opts.DoDefaulting || opts.DoValidation || opts.DoConversion {
+	if opts.DoDefaulting || opts.DoValidation || opts.DoConversion || opts.Spoke != "" {
```
varshaprasad96, line 101: Have intentionally left the `DoConversion` flag's implementation as it is to preserve the current functionality. Hence, if someone just specifies the `--conversion` flag in the command, we will generate the default message which we do now.

pkg/config/v2/config_test.go
```diff
@@ -225,6 +225,25 @@ var _ = Describe("cfg", func() {
 			Expect(c.HasGroup("other-group")).To(BeFalse())
 		})
 
+		It("ListResourceswithGK should be empty since there are no resources with specified GK", func() {
```
Adirio, line 228: Same, would you mind moving this to follow the same orderas the interface?

pkg/config/v2/config_test.go
```diff
@@ -225,6 +225,25 @@ var _ = Describe("cfg", func() {
...
+			versions := c.ListResourceswithGK(resource.GVK{
+				Group: "group",
+				Kind:  "Kind",
+			})
```
Adirio, line 232:
```suggestion
			versions := c.ListResourceswithGK(res.GVK)
```

pkg/config/v2/config_test.go
```diff
@@ -225,6 +225,25 @@ var _ = Describe("cfg", func() {
...
+
+		It("ListResourceswithGK should list all the resources with the specified GK", func() {
+			c.Gvks = append(c.Gvks, res.GVK)
+			versions := c.ListResourceswithGK(resource.GVK{
+				Group: "group",
+				Kind:  "Kind",
+			})
+			Expect(versions).NotTo(BeEmpty())
+			Expect(len(versions)).To(BeEquivalentTo(1))
+			Expect(versions[0]).To(BeEquivalentTo(res))
```
Adirio, line 244:
```suggestion
			Expect(versions).To(Equal([]resource.Resource{res}))
```

pkg/config/v3/config.go
```diff
@@ -271,6 +271,18 @@ func (c cfg) HasGroup(group string) bool {
...
 
+// ListResourceswithGK implements config.Config
+func (c cfg) ListResourceswithGK(gvk resource.GVK) []resource.Resource {
+	resources := make([]resource.Resource, 0)
+
+	for _, r := range c.Resources {
+		if r.Group == gvk.Group && r.Kind == gvk.Kind {
+			resources = append(resources, r)
+		}
+	}
```
Adirio, line 282: Creating a copy is important to prevent the caller to modify the internal list of resources without knowing it. We also need to restore the plural form in case it was omitted.
```suggestion
	for _, res := range c.Resources {
		if res.Group == gvk.Group && res.Kind == gvk.Kind {
			r := res.Copy()

			// Plural is only stored if irregular, so if it is empty recover the regular form
			if r.Plural == "" {
				r.Plural = resource.RegularPlural(r.Kind)
			}

			resources = append(resources, r)
		}
	}
```

pkg/config/v3/config.go
```diff
@@ -271,6 +271,18 @@ func (c cfg) HasGroup(group string) bool {
 	return false
 }
 
+// ListResourceswithGK implements config.Config
```
Adirio, line 274: Same, would you mind moving it to follow the interface order?

pkg/config/v2/config_test.go
```diff
@@ -225,6 +225,25 @@ var _ = Describe("cfg", func() {
...
+			})
+			Expect(versions).To(BeEmpty())
+		})
+
+		It("ListResourceswithGK should list all the resources with the specified GK", func() {
+			c.Gvks = append(c.Gvks, res.GVK)
+			versions := c.ListResourceswithGK(resource.GVK{
+				Group: "group",
+				Kind:  "Kind",
+			})
```
Adirio, line 241:
```suggestion
			versions := c.ListResourceswithGK(res.GVK)
```

pkg/config/v3/config_test.go
```diff
@@ -307,6 +307,25 @@ var _ = Describe("cfg", func() {
 			Expect(c.HasGroup("other-group")).To(BeFalse())
 		})
 
+		It("ListResourceswithGK should be empty since there are no resources with specified GK", func() {
+			versions := c.ListResourceswithGK(resource.GVK{
+				Group: "group",
+				Kind:  "Kind",
+			})
```
Adirio, line 314:
```suggestion
			versions := c.ListResourceswithGK(res.GVK)
```

pkg/config/v3/config_test.go
```diff
@@ -307,6 +307,25 @@ var _ = Describe("cfg", func() {
...
+			Expect(versions).To(BeEmpty())
+		})
+
+		It("ListResourceswithGK should list all the resources with the specified GK", func() {
+			c.Resources = append(c.Resources, res)
+			versions := c.ListResourceswithGK(resource.GVK{
+				Group: "group",
+				Kind:  "Kind",
+			})
```
Adirio, line 323:
```suggestion
			versions := c.ListResourceswithGK(res.GVK)
```

pkg/config/v3/config_test.go
```diff
@@ -307,6 +307,25 @@ var _ = Describe("cfg", func() {
...
+			Expect(versions).NotTo(BeEmpty())
+			Expect(len(versions)).To(BeEquivalentTo(1))
+			Expect(versions[0]).To(BeEquivalentTo(res))
```
Adirio, line 326:
```suggestion
			Expect(versions).To(Equal([]resource.Resource{res}))
```

pkg/config/v3/config_test.go
```diff
@@ -307,6 +307,25 @@ var _ = Describe("cfg", func() {
 			Expect(c.HasGroup("other-group")).To(BeFalse())
 		})
 
+		It("ListResourceswithGK should be empty since there are no resources with specified GK", func() {
```
Adirio, line 310: Same, would you mind moving these tests to follow the same order than the interface?

Adirio (changes requested):
Some initial comments regarding `config` modifications.

Offtopic: I think that we may want to change `GetResources() ([]resource.Resource, error)` to `ListResources() []resource.Resource` for consistency, but not in this PR.

Adirio (changes requested):
Several comments on the rest of the chnages:

camilamacedo86:
HI @varshaprasad96,

That is a nice PR. seems that it is closer to be finished. ( just address the suggestions )
so, I am /remove-lifecycle stale

Let me know when you need the next review round.

camilamacedo86:
Hi @varshaprasad96, 

WDYT about address the comments for we get this one merged?
 Please, feel free to ping me if you need another round of reviews.

camilamacedo86:
/remove-lifecycle stale

@varshaprasad96 do you think that would be possible to rebase it with the master and address the suggestions for w e get this one merged?

camilamacedo86:
Closing because it is open for a while without addressing the suggestions.
We created an issue and we linked this PR so that, if someone has time to continue working on this one will be able to move forward. Thank you for your contributions.

PR #2085, opened 2021-03-13
✨ config-gen create and bind to a non-default service account

pwittrock:
`kubebuilder alpha config-gen` support for non-default service accounts -- originally introduced in #2070

From #2070 description:

> As of now, all managers are deployed under the same service account in a namespace. Although this namespace is unique to the manager initially, it is a common admin pattern to deploy multiple managers under the same namespace for security reasons. This pattern becomes insecure given the current one-service-account-for-all approach, since RBAC will conflict.

> This PR adds a ServiceAccount (config/rbac/service_account.yaml) and changes the default service account name from default to controller-manager such that a user can specify which service account their manager should be created in. They may change this value by updating files referencing the metadata.name value in service_account.yaml.

> pkg/plugins/golang/v3: update all scaffold referencing the default service account with references to the "controller-manager" service account, specified in service_account.yaml.


pkg/cli/alpha/config-gen/cmd.go
```diff
@@ -172,6 +172,13 @@ The KubebuilderConfigGen resource has the following fields:
 		# generate prometheus ServiceMonitor resource
 	    enableServiceMonitor: true
 
+      serviceAccount:
+	    # configure the service account used for RBAC
+	    name: foo
+
+		# generate the service account resource
+        generate: true
+
```
Adirio, line 181: Spaces and tabs

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.15
 
 require (
 	github.com/cloudflare/cfssl v1.5.0 // for `kubebuilder alpha config-gen`
+	github.com/go-logr/logr v0.3.0 // indirect
```
Adirio, line 7: Try deleting these 3 imports and check if `go mod tidy` adds them.

pkg/cli/alpha/config-gen/templates/resources/service-account.template.yaml
```diff
@@ -0,0 +1,7 @@
+{{ if not .Spec.ControllerManager.ServiceAccount.NoGenerate }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Spec.ControllerManager.ServiceAccount.Name }}
+  namespace: {{ .Namespace }}
+{{ end }}
```
Adirio, line 7:
```suggestion
{{ end }}

```

pkg/cli/alpha/config-gen/testdata/serviceaccount/config.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: simple
+spec:
+  crds:
+    sourceDirectory: ../project/...
+
+  controllerManager:
+    image: example/simple:latest
+    serviceAccount:
+      name: foo
```
Adirio, line 12:
```suggestion
      name: foo

```

pkg/cli/alpha/config-gen/testdata/serviceaccountgenerate/config.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: simple
+spec:
+  crds:
+    sourceDirectory: ../project/...
+
+  controllerManager:
+    image: example/simple:latest
+    serviceAccount:
+      name: foo
+      generate: true
```
Adirio, line 13:
```suggestion
      generate: true

```

pkg/cli/alpha/config-gen/cmd.go
pwittrock: Done

go.mod
pwittrock: Running the tests does, but building the binary doesn't.  Will delete.

pwittrock:
PTAL

pkg/cli/alpha/config-gen/cmd.go
```diff
@@ -148,54 +148,68 @@ The KubebuilderConfigGen resource has the following fields:
...
+        name: foo
+
+        # generate the service account resource
+        generate: true
+
+      serviceAccount:
+	    # configure the service account used for RBAC
+	    name: foo
+
+		# generate the service account resource
```
Adirio, line 186: I found at least these 4 tabs
```suggestion
      # configure the service account used for RBAC
      name: foo

    # generate the service account resource
```

Adirio (changes requested):
IIRC, we are using a fixed name for the `ServiceAccount` and then prefixing it with a unique prefix per manager. This approach is a bit different, unsing a user-provided name and defaulting to the `KubebuilderConfigGen` name.

I don't have a strong opinion on which approach is better. @estroz?

pkg/cli/alpha/config-gen/cmd.go
```diff
@@ -148,54 +148,68 @@ The KubebuilderConfigGen resource has the following fields:
...
+        # generate prometheus ServiceMonitor resource
+        enableServiceMonitor: true
+
+      serviceAccount:
+        # configure the service account used for RBAC
+        name: foo
+
+        # generate the service account resource
+        generate: true
+
```
estroz, line 181: Duplicate?

estroz:
@Adirio correct. @pwittrock it seems some resources are generated [with their final name](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/alpha/config-gen/templates/resources/rbac.template.yaml#L4), and some [do not have a name prefix](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/alpha/config-gen/templates/resources/controller-manager.template.yaml#L4) such that they can be `kustomize edit set nameprefix`'d. Is this intentional?

pwittrock:
> @Adirio correct. @pwittrock it seems some resources are generated with their final name, and some do not have a name prefix such that they can be kustomize edit set nameprefix'd

What is the issue that keeps nameprefix from working?  This could be an issue with the transformer plugin ordering -- does it work if you use the plugin as a base for the kustomization that does the prefix?

pwittrock:
> IIRC, we are using a fixed name for the ServiceAccount and then prefixing it with a unique prefix per manager. This approach is a bit different, unsing a user-provided name and defaulting to the KubebuilderConfigGen name.

I suspect this is in part because of kustomize's inability to finely configure name transformations -- e.g. Kustomize doesn't allow a user to provide a name and then generate names from that, it only supports adding either a prefix or suffix.

estroz:
>What is the issue that keeps nameprefix from working?

@pwittrock nothing necessarily, I'm just wondering why some templates are generated with their final name and some are not such that kustomize can prefix their name. Shouldn't they all be prefixed by kustomize, or all templates have the final name, ex. [`{{ .Metadata.Name }}-controller-manager`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/alpha/config-gen/templates/resources/controller-manager.template.yaml#L4)?

>This could be an issue with the transformer plugin ordering -- does it work if you use the plugin as a base for the kustomization that does the prefix?

It seems kustomize does not apply builtin functions to resources post transformation within the same kustomize dir. In #2099 I place the transformer in `config/configgen/kustomization.yaml` then reference that kustomize dir from `config/default/kustomization.yaml` as a resource; the latter contains a name prefix and applies it successfully.

PR #2088, opened 2021-03-15
✨ Add fish completion

Adirio:
A recent PR bumped Cobra dependency to 1.1.1 which incudes support for Fish completion.
Previously, Fish completion was commented out as the Cobra version we were using didn't support it.
This PR adds support for Fish completion.


camilamacedo86 (approved):
Great 👍 

/lgtm

PR #2089, opened 2021-03-15
✨ Provide third-party devs the option to provide their own alpha subcommands

Adirio:
SDK is providing an alpha command, we are doing the same, when we try to align them, they will collide.
This PRs provides a way to set extra alpha commands for third-party developers.

### DIFF
--- a/pkg/cli/alpha.go
+++ b/pkg/cli/alpha.go
@@ -17,10 +17,15 @@ limitations under the License.
 package cli
 
 import (
+	"fmt"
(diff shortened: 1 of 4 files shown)


camilamacedo86 (approved):
After the https://github.com/kubernetes-sigs/kubebuilder/pull/1831 get merged we cannot align SDK with KB because both have alpha commands. This pr solves this problem. 

Great work 🥇 

/lgtm

PR #2091, opened 2021-03-15
📖 Remove kubebuilder hardcoded reference in alpha subcommand description

Adirio:
Alpha subcommand had a hardcoded reference to kubebuilder.
As kubebuilder may be used as a library by other CLI tools, the description should be more generic.


camilamacedo86 (approved):
/lgtm

PR #2092, opened 2021-03-15
🐛 Fix alpha extra commands parent command

Adirio:
#2089 was adding the commands to the root command instead of to the alpha subcommand.
This PRs fixes the bug and adds a test to verify it in the future.


camilamacedo86 (approved):
/lgtm

PR #2094, opened 2021-03-16
🌱 Fix coverage report on master

Adirio:
Master branch was not properly linted, and therefore the coverage report was not being uploaded.

Setting the emoji as 🌱 as it only affects the CI.


camilamacedo86 (approved):
/lgtm

Remove unused (deadcode)

PR #2095, opened 2021-03-16
🌱  Enhance GitHub actions CI

Adirio:
The GitHub actions CI workflow is run for both pull-requests (targeting the remote branch) and the merge commit push whey they re merged (targeting the branch where they were merged into). Pull-request-related results are exposed under the checks of the PR, which makes them pretty accessible. However, push-related results are only shown in the "Actions" tab and are usually not reviewed. As an example, a linting issue was introduced nearlly a month ago in master, which was passing the PR linting check as it only checks the modified code, but was not passing the push linting check. This means that coverage report was not being sent to coveralls about the master branch since the February 22nd, and we just realized today (March 16th).

This PR adds an additional debugging step that will create an issue if one of the "hidden" GitHub action CI workflows fail with the SHA of the offending commit, the link to the corresponding action execution wher some debugging information will alos be found (secrets are safely removed first).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,26 +22,6 @@ jobs:
           version: v1.37        # Always uses the latest patch version.
           only-new-issues: true # Show only new issues if it's a pull request
 
-  testdata:
-    name: Verify testdata directory
-    needs: lint
-    runs-on: ubuntu-latest
-    # Pull requests from the same repository won't trigger this checks as they were already triggered by the push
-    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
-    steps:
-      - name: Clone the code
-        uses: actions/checkout@v2
-      - name: Setup Go
-        uses: actions/setup-go@v2
-        with:
-          go-version: '1.15'
-      # This step is needed as the following one tries to remove
-      # kustomize for each test but has no permission to do so
-      - name: Remove pre-installed kustomize
-        run: sudo rm -f /usr/local/bin/kustomize
-      - name: Verify testdata directory
-        run: make check-testdata
-
   test:
     name: Test for ${{ matrix.os }}
     needs: lint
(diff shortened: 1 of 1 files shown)


Adirio:
@estroz 
/testing
/retest

.github/workflows/ci.yml
```diff
@@ -87,3 +65,44 @@ jobs:
...
+  debug:
+    name: Debug failing workflows
+    needs: # Add to this list every job in this workflow
+      - lint
+      - test
+      - coverage
+      - testdata
+    runs-on: ubuntu-latest
+    # Failed PRs are shown in their check but pushes are hard to notice so we create an issue
+    if: failure() && github.event_name == 'push'
```
estroz, line 98: This should probably only run on master.

Adirio: Reason?
That would prevent it running on `release-X` branches for example, and we can always close the issue easily

estroz: I'm thinking of all the branches existing in this repo having issues created each time the branch is broken and someone pushes, which is possibly often on non-trunk dev branches. Will the debug action recreate an issue for each push, or only once for a broken branch?

Adirio: It will recreate one per push, yes.
In relation to all those branches, I created several issues recently tracking some changes that should be done to clean up a bit the tree, like removing some branches, or moving some to `kubebuilder-release-tools` as we commented in a meeting. I don't have admin privileges to delete branches though.

Adirio:
Another example of this "hidden"/"quiet" fails occured with #2060 and we only realized when merging `master` into `release-3` for the next beta release.

.github/workflows/ci.yml
Adirio: Done.

estroz (approved):
/lgtm

PR #2096, opened 2021-03-17
✨ Enable the storage of multiple plugins as layout

Adirio:
Use a custom type to ensure backwards compatibility: older configuration files will unmarshall properly despite not being an array and will marshal to the new format.

/kind feature

### DIFF
--- a/pkg/config/interface.go
+++ b/pkg/config/interface.go
@@ -20,78 +20,78 @@ import (
 	"sigs.k8s.io/kubebuilder/v3/pkg/model/resource"
 )
 
-// Config defines the interface that project configuration types must follow
+// Config defines the interface that project configuration types must follow.
 type Config interface {
 	/* Version */
 
-	// GetVersion returns the current project version
+	// GetVersion returns the current project version.
 	GetVersion() Version
 
 	/* String fields */
 
-	// GetDomain returns the project domain
+	// GetDomain returns the project domain.
 	GetDomain() string
-	// SetDomain sets the project domain
+	// SetDomain sets the project domain.
 	SetDomain(domain string) error
 
 	// GetRepository returns the project repository.
 	GetRepository() string
-	// SetRepository sets the project repository
+	// SetRepository sets the project repository.
 	SetRepository(repository string) error
 
-	// GetProjectName returns the project name
+	// GetProjectName returns the project name.
 	// This method was introduced in project version 3.
 	GetProjectName() string
-	// SetProjectName sets the project name
+	// SetProjectName sets the project name.
 	// This method was introduced in project version 3.
 	SetProjectName(name string) error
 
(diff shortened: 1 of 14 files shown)


Adirio:
@camilamacedo86 @estroz @DirectXMan12 @jmrodri this is what we were discussing yesterday.

pkg/config/interface.go
```diff
@@ -20,78 +20,78 @@ import (
...
+	// SetProjectName sets the project name.
 	// This method was introduced in project version 3.
 	SetProjectName(name string) error
 
-	// GetLayout returns the config layout
+	// GetPluginChain returns the plugin chain.
 	// This method was introduced in project version 3.
-	GetLayout() string
-	// SetLayout sets the Config layout
+	GetPluginChain() []string
```
estroz, line 51: Can we keep this as `GetLayout() []string`? It's nice to have method names that map to YAML fields.

pkg/config/v3/config.go
```diff
@@ -29,15 +29,38 @@ import (
...
+		}
+		*ss = sl
+		return nil
+	}
+
+	var st string
+	if err := yaml.Unmarshal(b, &st); err != nil {
+		return err
+	}
+	*ss = stringSlice{st}
```
estroz, line 51: This is supposedly a comma-separated list right? Also no need to unmarshal.
```suggestion
	*ss = stringSlice(strings.Split(string(b), ","))
```

Adirio: It is not a comma separated list, plugin phase 1.5 planned on suign a comma separated list to keep it as a string, but it is a single string at the moment.

We either need to unmarshal or to handle quotes ourself, as `strings.Unquote` requires them to be there while we may not have them so unmarshalling is just the easist way to get it done.

pkg/config/interface.go
Adirio: I think that the `Config` interface should abstract us from the file, so I think that its methods should be descriptive and I don't think layout fits there. I would change it in the file if it wasn't backwards incompatible.

pkg/config/v3/config.go
estroz: Wasn't the whole point of `stringSlice` to support both a comma separated list and a YAML list?

Adirio: My point is, if there was no previous support for comma separated list, and we are going to use a YAML array after this, why do we need to support a comma separated list? Any time a command is run the config file will be rewritten to the YAML array format, the reason to support a string is backwards compatibility, but comma separated list has never been there so I don't see any reason to support it.

estroz: I see your point. String splitting guards against someone accidentally using a comma separated value, but perhaps it would be better to error out in that case.

PR #2097, opened 2021-03-17
:seedling: config-gen: command docs no longer depend on local environment

estroz:
When creating CLI docs using cobra's docs generator,
this doc is generated inconsistently depending on what
environment the command is run. This PR fixes that
problem by using a static path and saying where the
dynamic part comes from.

### DIFF
--- a/pkg/cli/alpha/config-gen/cmd.go
+++ b/pkg/cli/alpha/config-gen/cmd.go
@@ -19,6 +19,7 @@ package configgen
 import (
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


estroz:
Not sure why coverage is falling by so much.

estroz:
Making sure perms work

/hold

pkg/cli/alpha/config-gen/cmd.go
```diff
@@ -252,40 +251,48 @@ EOF
...
 			err = os.MkdirAll(dir, 0700)
 			if err != nil {
 				return err
 			}
 
-			return ioutil.WriteFile(pluginFile, []byte(`#!/bin/bash 
+			return ioutil.WriteFile(fullScriptPath, []byte(`#!/bin/bash
 KUSTOMIZE_FUNCTION=true kubebuilder alpha config-gen
-`), 0500)
+`), 0700)
```
estroz, line 281: These perms let the command overwrite this file. Otherwise rerunning the command will fail. @pwittrock are you ok with this for now?

camilamacedo86: We can ignore here the gosec here.

estroz: Removed this perm change. The command doesn't attempt to overwrite the file, just prints that it exists.

pwittrock:
/lgtm
/retest

+1 I like the checking more.

PR #2099, opened 2021-03-18
:sparkles: add new plugin `go-config-gen/v1`

estroz:
This PR adds the `go-config-gen/v1` plugin, which uses the recently merged [`alpha config-gen`](#1831) to configure a project. All Go code is untouched, so perhaps this idea is a good candidate for a config-only plugin as suggested in #2015.

This plugin has two modes:

Use `config-gen` directly with `kubebuilder alpha config-gen`, which will write the configgen and componentconfig manifests to the top level.
```sh
kubebuilder init --plugins go-config-gen
```

Use through `kustomize build --enable-alpha-plugins`, which will write a configgen, componentconfig, and kustomization.yaml to ./config/configgen, and a kustomization.yaml to ./config/default that calls ../configgen
```sh
kubebuilder init --plugins go-config-gen --with-kustomize
```

*The code in this PR is experimental and probably won't merge for awhile.*

/kind feature


estroz:
I'm now on the fence about whether to scaffold kustomization.yaml's by default and provide `--no-kustomize`, or stick with `--with-kustomize`. I imagine most users will eventually migrate to using kustomize, and the base kustomize scaffold is quite simple right now.

cmd/main.go
```diff
@@ -22,6 +22,7 @@ import (
 	"sigs.k8s.io/kubebuilder/v3/pkg/cli"
 	cfgv2 "sigs.k8s.io/kubebuilder/v3/pkg/config/v2"
 	cfgv3 "sigs.k8s.io/kubebuilder/v3/pkg/config/v3"
+	goconfiggenv1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang-config-gen/v1"
```
camilamacedo86, line 25: Let's use alpha here as well `go-config-gen/alphav1` until it gets mature enough to be v1.

pwittrock:
For the non-config-gen specific scaffolding, is it possible to simply reference another package?

estroz:
@pwittrock I should be able to move the plugin directory into a subdir of `pkg/plugins/golang/v3` to access the internal directory containing scaffolds so I can re-use some of them. That should slim down the changes here significantly.

pkg/plugins/golang-config-gen/v1/api.go
```diff
@@ -0,0 +1,248 @@
...
+	// (used only to gen api with --pattern=addon)
+	// TODO: remove this when a better solution for using addons is implemented.
+	KbDeclarativePatternVersion = "b84d99da021778217217885dd9582ed3cc879ebe"
+
+	// defaultCRDVersion is the default CRD API version to scaffold.
+	defaultCRDVersion = "v1"
+)
+
+// DefaultMainPath is default file path of main.go
+const DefaultMainPath = "main.go"
```
camilamacedo86, line 51: Some comments, otherwise relly nice we see it moving forward.

We cannot drop do the support to the plugins, so IMO we need to allow we use it opt-in opt which we will be to achieve with https://github.com/kubernetes-sigs/kubebuilder/issues/2015. So, can we:
-  Have a plugin that only scaffold what is of its domain of responsibility (I d not think that. we should triplicated the files here )
- Then, we can create a testdata for go v3 where we will scaffold the project, remove the config dir and then add the config-gen for we start to try to make it backwards compatible with the plugin until we are able to push/merge https://github.com/kubernetes-sigs/kubebuilder/issues/2015?

estroz: * Before this goes into master, I agree. I'd like to see this refactored into something like `config-gen/v1-alpha`, where `config/v3` is the default `config/` directory plugin.
* Sure

cmd/main.go
estroz: The package name doesn't really matter IMO, since the plugin has a `v1-alpha` version.

pkg/cli/alpha/config-gen/cmd.go
```diff
@@ -19,6 +19,7 @@ package configgen
 import (
```
estroz, line 19: All changes here are part of #2097 and will be removed once it merges.

cmd/main.go
camilamacedo86: I think that matter because:
- make explicitly that is in alpha for maintainers and consumers
- kb is also a lib/api and we need to make it clear that we can break everything that we wish with this one.

camilamacedo86:
@estroz the config-gen has no language we should not `kubebuilder init --plugins go-config-gen`. After 1.5 merge we will be able to `kubebuilder init --plugins=config-gen/alphav1,go/v3  instead.

cmd/main.go
```diff
@@ -34,6 +35,7 @@ func main() {
 		cli.WithPlugins(
 			&pluginv2.Plugin{},
 			&pluginv3.Plugin{},
+			&goconfiggenv1.Plugin{},
```
camilamacedo86, line 38: @estroz regarding`kubebuilder init --plugins go-config-gen`. After https://github.com/kubernetes-sigs/kubebuilder/pull/2060 get merged we will be able to use it as `kubebuilder init --plugins=config-gen/alphav1,go/v3`  instead. 

See that, we will probably need to create customize plugins to add things on top via its transformer func and use the bundle plugin which will be like `config-gen/alphav1,go/v3,go.transform/alphav1` which that can be `go-config-gen/alphav1` as suggested. I think is that the reason for you choose the name. just to be sure :-)

testdata/project-v3-config-gen/controllers/suite_test.go
```diff
@@ -0,0 +1,83 @@
...
+		"Controller Suite",
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = BeforeSuite(func() {
+	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
+
+	By("bootstrapping test environment")
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths:     []string{filepath.Join("..", "config", "crd", "bases")},
```
estroz, line 57: This still needs to be changed.

PR #2106, opened 2021-03-25
:sparkles: add the common plugin(s) to allow it to be used by consumers 

camilamacedo86:
**Description**
- :sparkles: add the common plugin(s) to allow it to be used by consumers
- organize the plugins as suggested to make clear what is common and what is language-based. 

Screen Shot 2021-03-27 at 08 51 28

- plugins/common: Here we will have ANY plugin that is common and not based in any language such as config-gen, and this one config-base. 
- plugins/golang: Here we have all plugins whcih are ONLY valid for Golang. 

**Motivation**
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/2015
To check more about its motivation see: https://deploy-preview-2115--kubebuilder.netlify.app/plugins/creating-plugins.html#language-based-plugins
PR which was done for SDK: https://github.com/operator-framework/operator-sdk/pull/4701/ in order to show that worked and its result.


pkg/plugins/golang/v3/scaffolds/webhook.go
```diff
@@ -93,6 +95,9 @@ func (s *webhookScaffolder) Scaffold() error {
 		&webhook.Kustomization{Force: s.force},
 		&webhook.KustomizeConfig{},
 		&webhook.Service{},
+		&certmanager.Certificate{},
+		&certmanager.Kustomization{},
+		&certmanager.KustomizeConfig{},
```
camilamacedo86, line 100: we should not scaffold the certmanager by default. 
It is valid only for webhooks and then, should only be scaffolded when the project has webhooks.

testdata/project-v3-config/Makefile
```diff
@@ -35,7 +35,7 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
-	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.txt" paths="./..."
```
camilamacedo86, line 38: the license is created for the base config and does require to be specific for Golang.
Then, has no need for it is called go.txt.

Adirio: The reason why the language needs to be part of the license file is because comments are different in other languages. For example, you could not use this license for Python operators.

cmd/main.go
```diff
@@ -22,22 +22,32 @@ import (
...
-	pluginv2 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v2"
-	pluginv3 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v3"
+	golangv2 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v2"
+	golangv3 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v3"
 )
 
 func main() {
+
+	// Bundle plugin which built the golang projects scaffold by Kubebuilder
+	gov3Bundle, _ := plugin.NewBundle(golangv3.Plugin{}.Name(), golangv3.Plugin{}.Version(),
```
Adirio, line 35: Plugin and bundle should have different names. I know we are doing this in SDK but we already made clear that is "hackish" and we don't want to do the same here. The bundle should probably use "go.kubebuilder.io" while the plugin can use something like "base.go.kubebuilder.io". Delarative should probably also be renamed to "declarative.go.kubebuilder.io" as is it Golang-specific.

Adirio (changes requested):

cmd/main.go
Adirio: IMO we should leave the `plugins`directory like this:
```
plugins/
  base/           --> "base.kubebuilder.io"
    v2/
    v3/
  golang/
    base/         --> "base.go.kubebuilder.io"
      v2/
      v3/
    declarative/  --> "declarative.go.kubebuilder.io"
      v1/

File: pkg/plugins/golang/v3/scaffolds/webhook.go (hunk 1 above)
Comment (Adirio) [reply pkg/plugins/golang/v3/scaffolds/webhook.go:98]: We should probably split this change to a different PR.

Comment (camilamacedo86) [reply pkg/plugins/golang/v3/scaffolds/webhook.go:98]: that is the intention. It is wip :-)

File: cmd/main.go (hunk 3 above)
Comment (camilamacedo86) [reply cmd/main.go:35]: I do not think that we need to do that. for v2. We can just provide the plugin base for v3 and leave v2 as it is and do not increase the effort to keep that maintained.

File: testdata/project-v3-config/Makefile (hunk 2 above)
Comment (camilamacedo86) [reply testdata/project-v3-config/Makefile:38]: that is true. however, we can add by default a hack here which would be used by all with # to do the comments.

Comment (Adirio) [reply testdata/project-v3-config/Makefile:38]: We could even do a plugin that only scaffolds this file and chooses the right format depending on the language used, but as you said lets leave this outside of this PRs scope. Another reason to keep it outside of this PR is that this change is breaking while the rest of the can probably be non-breaking at least for the user.

File: cmd/main.go (hunk 3 above)
Comment (Adirio) [reply cmd/main.go:35]: There are 2 bases, `"base.kubebuilder.io"` (language-agnostic) will be the one scaffolding the `config` directory and `"base.go.kubebuiolder.io"` (go-specific) would be the golang specific files (`*_types.go`, `*_controller.go`, ...) so that we free up the `"go.kubebuilder.io"` name for the bundle. Names can be changed. The point is that the bundle and the plugin should have different names.

Related to the base language-agnostic plugin, I think we should od it for both v2 and v3, I don't think it will be harder to mantain. I actually think that it will be harder to mantain if we have to solve a bug for v2 and v3 if we have to search comething in the "base.kubebuilder.io/v1" and in "base.go.kubebuilder.io/v2"  plugins than if we just have to check both "base.kubebuilder.io" plugins.

Comment (camilamacedo86) [reply cmd/main.go:35]: >There are 2 bases, "base.kubebuilder.io" (language-agnostic) will be the one scaffolding the config directory and "base.go.kubebuiolder.io" (go-specific) would be the golang specific files (*_types.go, *_controller.go, ...) so that we free up the "go.kubebuilder.io" name for the bundle. Names can be changed. The point is that the bundle and the plugin should have different names.

I do not agree with that because what is scaffold by the plugin in the config is not a BASE. It is their specific files that are generated according to the language plugin actions. 

Base == what equals no matter the language. Then, it does not necessary means only the config/ directory. it can be any thing that would be valid as a BASE for any language plugin. 

> Related to the base language-agnostic plugin, I think we should od it for both v2 and v3, I don't think it will be harder to mantain. I actually think that it will be harder to mantain if we have to solve a bug for v2 and v3 if we have to search comething in the "base.kubebuilder.io/v1" and in "base.go.kubebuilder.io/v2" plugins than if we just have to check both "base.kubebuilder.io" plugins.

Do we want people creating common bases with v2 for them to create their new plugins? E.g
I want to use kb only to generate the common base and then do the rest of the scaffolds using my implementation in java/python and etc.

I do not think so. We should push people to use the V3 layout and not the V2 plugin which only exist to keep the backwards compatibility with 2.x. However, we can always add that in the future if we have a real case that requires that. SDK will only consume the same base used for v3 too.

Comment (Adirio) [reply cmd/main.go]: Discussed offline to better understand ourselves.

File: pkg/plugins/golang/domain.go [hunk 4]
```diff
@@ -0,0 +1,22 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package golang
+
+import "sigs.k8s.io/kubebuilder/v3/pkg/plugins"
+
+// DefaultNameQualifier is the suffix appended to all kubebuilder plugin names.
```
Adirio, line 21:
```suggestion
// DefaultGoNameQualifier is the suffix appended to all go-specific kubebuilder plugin names.
```

testdata/project-v3/go.mod
```diff
@@ -10,5 +10,5 @@ require (
 	k8s.io/apimachinery v0.19.2
 	k8s.io/client-go v0.19.2
 	sigs.k8s.io/controller-runtime v0.7.2
-	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20210113160450-b84d99da0217
+	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200522144838-848d48e5b073
```
camilamacedo86, line 13: It means that a release was made. However, in the past, we fixed that by pinning a declarative commit-version. So, it might be a regression.

cmd/main.go
```diff
@@ -18,26 +18,36 @@ package main
 
 import (
 	"log"
-
```
Adirio, line 21: This blank line shoult not be removed.

cmd/main.go
```diff
@@ -18,26 +18,36 @@ package main
...
+
 	c, err := cli.New(
 		cli.WithCommandName("kubebuilder"),
 		cli.WithVersion(versionString()),
 		cli.WithPlugins(
-			&pluginv2.Plugin{},
-			&pluginv3.Plugin{},
+			&golangv2.Plugin{},
+			gov3Bundle,
+			&commonv1.Plugin{},
```
Adirio, line 46: We should probably also register the "base.go.kubebuilder.io" plugin so that when a different "common" plugin is implemented (for example including config-gen) we can provide `--plugins=common/v2,base.go/v3`:
```suggestion
			&golangv2.Plugin{},
			&golangv3.Plugin{},
			gov3Bundle,
			&commonv1.Plugin{},
```

pkg/cli/cli.go
```diff
@@ -215,7 +215,7 @@ func (c *CLI) getInfoFromConfig(projectConfig config.Config) error {
 // getInfoFromFlags obtains the project version and plugin keys from flags.
 func (c *CLI) getInfoFromFlags(hasConfigFile bool) error {
 	// Partially parse the command line arguments
-	fs := pflag.NewFlagSet("base", pflag.ContinueOnError)
+	fs := pflag.NewFlagSet("common", pflag.ContinueOnError)
```
Adirio, line 218: Revert.

cmd/main.go
camilamacedo86: Let's leave that for when we will be able to use it with the config-gen. IHMO let's try to provide only the minimal set for what we are trying to use now. We can add that after since is not a breaking change and it is very easy too.

testdata/project-v3-addon/go.mod
```diff
@@ -9,5 +9,5 @@ require (
 	k8s.io/apimachinery v0.19.2
 	k8s.io/client-go v0.19.2
 	sigs.k8s.io/controller-runtime v0.7.2
-	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20210113160450-b84d99da0217
+	sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200522144838-848d48e5b073
```
Adirio, line 12: `declarative/v1` was checking if the plugin was `go/v2` or `go/v3` and using the appropiate SHA. The check seems to be broken. You will need to fix it.

testdata/project-v3/go.mod
Adirio: No, we are still pinning but there is a check for go/v2 or go/v3 because different SHAs are pinned for each case and the check seems to be broken.

cmd/main.go
camilamacedo86: PS.: I am just adding the base here for we are able to provide an option to create the common base and then scaffold the rest with any language and not KB lib.

camilamacedo86: Could we agree on that?

cmd/main.go
camilamacedo86: Thank you for your help and inputs. I think it is fine close to this one.

testdata/project-v3-addon/go.mod
camilamacedo86: Do you know where? Because we did not change that at all.

pkg/plugins/golang/domain.go
```diff
@@ -0,0 +1,22 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package golang
+
+import "sigs.k8s.io/kubebuilder/v3/pkg/plugins"
+
+// DefaultGoNameQualifier is the suffix appended to all kubebuilder plugin names.
+const DefaultGoNameQualifier = "go" + plugins.DefaultNameQualifier
```
Adirio, line 22: `plugins.DefaultNameQualifier` is `.kubebuilder.io` and `golang.DefaultGoNameQualifier` is `go.kubebuilder.io`. We should choose if the `Default*NameQualifier` consts have the `.` or not and stick with that for all of them.

pkg/plugins/golang/domain.go
```diff
@@ -0,0 +1,22 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package golang
+
+import "sigs.k8s.io/kubebuilder/v3/pkg/plugins"
+
+// DefaultGoNameQualifier is the suffix appended to all kubebuilder plugin names.
```
Adirio, line 21: NIT:
```suggestion
// DefaultGoNameQualifier is the suffix appended to all kubebuilder plugin names for Golang operators.
```

pkg/plugins/golang/declarative/v1/plugin.go
```diff
@@ -22,10 +22,10 @@ import (
 	cfgv3 "sigs.k8s.io/kubebuilder/v3/pkg/config/v3"
 	"sigs.k8s.io/kubebuilder/v3/pkg/model/resource"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
-	"sigs.k8s.io/kubebuilder/v3/pkg/plugins"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang"
 )
 
-const pluginName = "declarative" + plugins.DefaultNameQualifier
+const pluginName = "declarative." + golang.DefaultGoNameQualifier
```
Adirio, line 28: Base plugins also require this dot unless you add it to the `golang.DefaultGoNameQualifier` const.

Adirio (changes requested):
There are also common files in the create API call, right?

camilamacedo86:
/hold 

I am pushing a PR against SDK as well. Just for we be sure that we covered all.

camilamacedo86:
/hold

i am checking that with sdk side

pkg/plugins/common/v2/plugin.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"sigs.k8s.io/kubebuilder/v3/pkg/config"
+	cfgv2 "sigs.k8s.io/kubebuilder/v3/pkg/config/v2"
+	cfgv3 "sigs.k8s.io/kubebuilder/v3/pkg/config/v3"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins"
+)
+
+const pluginName = "common." + plugins.DefaultNameQualifier
```
Adirio, line 27: `pugins.DefaultNameQualifier` already has the `.`
```suggestion
const pluginName = "common" + plugins.DefaultNameQualifier
```

pkg/plugins/common/v3/plugin.go
```diff
@@ -0,0 +1,50 @@
...
+package v3
+
+import (
+	"sigs.k8s.io/kubebuilder/v3/pkg/config"
+	cfgv3 "sigs.k8s.io/kubebuilder/v3/pkg/config/v3"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins"
+)
+
+const pluginName = "common." + plugins.DefaultNameQualifier
```
Adirio, line 26: Same

camilamacedo86:
/hold cancel

PR done.

Adirio (approved):

Adirio:
Some TODOs that may be interesting for follow-up PRs:
- [ ] Allow bundles to wrap other bundles.
- [ ] Provide bundles for the declarative pattern: allowing `--plugins=declarative/v3` instead of `--plugins=go/v3,declarative`.
- [ ] Expose bundles as any other plugin instead of creating them in `cmd/main.go`.

camilamacedo86:
/hold 

To give the opportunity for all check that out.

cmd/main.go
```diff
@@ -22,22 +22,40 @@ import (
 	"sigs.k8s.io/kubebuilder/v3/pkg/cli"
 	cfgv2 "sigs.k8s.io/kubebuilder/v3/pkg/config/v2"
 	cfgv3 "sigs.k8s.io/kubebuilder/v3/pkg/config/v3"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
+	commonv2 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/common/v2"
```
estroz, line 26: I don’t see a point in complicating the v2 plugin code by breaking it up. Just leave it as-is.

cmd/main.go
```diff
@@ -22,22 +22,40 @@ import (
...
+
+	// Bundle plugin which built the golang projects scaffold by Kubebuilder go/v2
+	gov2Bundle, _ := plugin.NewBundle(golang.DefaultGoNameQualifier, plugin.Version{Number: 2},
+		commonv2.Plugin{},
+		golangv2.Plugin{},
+	)
+
+	// Bundle plugin which built the golang projects scaffold by Kubebuilder go/v3
+	gov3Bundle, _ := plugin.NewBundle(golang.DefaultGoNameQualifier, plugin.Version{Number: 3},
+		commonv3.Plugin{},
```
estroz, line 44: `common` is ambiguous. Why not `kustomize-config`?

estroz (changes requested):

cmd/main.go
camilamacedo86: I agree. That was my first idea here too.

cmd/main.go
camilamacedo86: I'd like a more generic name to make clear that is the base for any plugin language. So, wdyt about base?

pkg/plugins/golang/base/v3/plugin.go
```diff
@@ -20,10 +20,10 @@ import (
 	"sigs.k8s.io/kubebuilder/v3/pkg/config"
 	cfgv3 "sigs.k8s.io/kubebuilder/v3/pkg/config/v3"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
-	"sigs.k8s.io/kubebuilder/v3/pkg/plugins"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang"
 )
 
-const pluginName = "go." + plugins.DefaultNameQualifier
+const pluginName = "base." + golang.DefaultGoNameQualifier
```
estroz, line 26: The `Go` bit is redundant
```suggestion
const pluginName = "base." + golang.DefaultNameQualifier
```

cmd/main.go
estroz: I could be convinced of `base.kustomize-config`, much like `base.go` is now a thing and reasonable.

cmd/main.go
Adirio: I prefer it to be the samne as v3, so that when we have to fix a bug in v2 we can find it in the same place as v3, and not the common part for v3 but the full plugin for v2.

cmd/main.go
Adirio: Camila's point is that it may be more things than kustomize configs. Like a base Makefile, or a LICENSE file. `seed`? `root`? `foundation`? I'm just spamming the first thing that goes through my head.

pkg/plugins/golang/base/v3/plugin.go
Adirio: The `Default` part could probably also be applied to `plugins.DefaultNameQualifier`

cmd/main.go
estroz: The current Makefile is intimately tied to both `kustomize` and `go`. Ideally, the `kustomize`-related and `go`-related portions would be split into separate Makefile fragments, then combined at runtime. In practice this is kinda difficult to do because of shared variables so I don't necessarily recommend it right now.

The license is only applied to Go files, so that would go in the `base.go` plugin.

These are specific cases, but help enforce my point that files not of the same type as those of the "meat" of the host plugin, denoted by name in `<name>.kubebuilder.io`, are still related to that type.

Adirio: I was not talking to the boilerplate license, but a LICENSE file. I feel like `base.kubebuilder.io` and `base.go.kubebuilder.io` are not too different. However, users will never have to use `base.go.kubebuilder.io` as it is not even being registered in the CLI, the bundle `go.kubebuilder.io` is the one being registered, so we could use `base.kubebuilder.io`.

estroz: >I was not talking to the boilerplate license, but a LICENSE file

Ah ok. In that case, the LICENSE file could just as easily be in the go base plugin, or a separate "common" plugin that doesn't contain any kustomize (or other) configuration files. Contrapositively, say I do call the kustomize config plugin "common", then create a config-gen plugin to replace kustomize configs with a config-gen config; what should I call the config-gen plugin, which also needs a LICENSE file but can't be paired with the "common" plugin?

>users will never have to use base.go.kubebuilder.io as it is not even being registered in the CLI, the bundle go.kubebuilder.io is the one being registered, so we could use base.kubebuilder.io.

True, since the "common" plugin is allowed as a first-class plugin, wouldn't it make sense to be more specific about what it contains? How about `common.kustomize`?

cmd/main.go
camilamacedo86: Hi @Adirio, 

See how it is now. I updated the first comment as well to clarify. IMO it is much simpler and understandable. 
Note that:
- The new config-base plugin does not need to have the same version of the language ones. 
- And we have no motivation to want to do that with `golang/v2`.  (That only exist to ensure the backwards compability)

cmd/main.go
camilamacedo86: We have the idea of having a `config-gen` plugin. @estroz is working on that.
So, I change the name of it to `config-base` and then, I hope that makes everybody happy. 

However, if not. I am OK with any name for it which does not have `kustomize` on that. Why? It brings the idea that is == golang plugins but using the only kustomize when is not. The idea of this plugin is to scaffold a common-base, common-configuration valid no matter the language.

pkg/plugins/golang/base/v3/plugin.go
camilamacedo86: Done and closed.

camilamacedo86:
HI @Adirio, 

> Provide bundles for the declarative pattern: allowing --plugins=declarative/v3 instead of --plugins=go/v3,declarative.

I do not think that it should be a bundle. IHMO as we spoke before the problem here is that for this scenario the sub-command should run global plugins + chain informed via the flag which means that users would only inform the declarative one to custom the default scaffold for their project. However, let's create an issue for we discuss that.

cmd/main.go
estroz: Your argument against using `kustomize` in the plugin name is not convincing. You can re-use `kustomize` in any plugin name and add more information to the name, so there really are no restrictions here.

pkg/plugins/common/config-base/v1/init.go
```diff
@@ -0,0 +1,110 @@
...
+  # Initialize a common project defining an specific project version
+  %[1]s init --plugins common/v3 --project-version 3
+`, cliMeta.CommandName)
+}
+
+func (p *initSubcommand) BindFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&p.domain, "domain", "my.domain", "domain for groups")
+	fs.StringVar(&p.name, "project-name", "", "name of this project")
+	fs.BoolVar(&p.componentConfig, "component-config", false,
+		"create a versioned ComponentConfig file, may be 'true' or 'false'")
```
Adirio, line 67: These 3 variables are only used to be set in config, so it can be done in any plugin. Which is the reasoning to move them to this plugin and not keeping them in the previous one?

pkg/plugins/golang/v3/init.go
```diff
@@ -86,21 +82,14 @@ func (p *initSubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&p.owner, "owner", "", "owner to add to the copyright")
 
 	// project args
-	fs.StringVar(&p.domain, "domain", "my.domain", "domain for groups")
 	fs.StringVar(&p.repo, "repo", "", "name to use for go module (e.g., github.com/user/repo), "+
 		"defaults to the go package of the current working directory.")
-	fs.StringVar(&p.name, "project-name", "", "name of this project")
-	fs.BoolVar(&p.componentConfig, "component-config", false,
-		"create a versioned ComponentConfig file, may be 'true' or 'false'")
+
```
Adirio, line 87: Remove blank line

pkg/plugins/golang/v3/scaffolds/api.go
```diff
@@ -19,6 +19,8 @@ package scaffolds
 import (
 	"fmt"
 
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v3/scaffolds/internal/templates/controllers"
```
Adirio, line 22: Import order

pkg/plugins/golang/v3/scaffolds/init.go
```diff
@@ -124,12 +109,8 @@ func (s *initScaffolder) Scaffold() error {
 			KustomizeVersion:         KustomizeVersion,
 			ControllerRuntimeVersion: ControllerRuntimeVersion,
 		},
+		&kdefault.Kustomization{},
```
Adirio, line 112: This was changed to be ordered by package, so can you keep this where it was previously instead of moving it to the midle of the `templates` package templates?

Adirio (changes requested):

pkg/plugins/common/config-base/v1/plugin.go
```diff
@@ -0,0 +1,50 @@
...
+	cfgv3 "sigs.k8s.io/kubebuilder/v3/pkg/config/v3"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins"
+)
+
+const pluginName = "config-base." + plugins.DefaultNameQualifier
+
+var (
+	pluginVersion            = plugin.Version{Number: 1}
+	supportedProjectVersions = []config.Version{cfgv3.Version}
```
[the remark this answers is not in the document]
camilamacedo86: Is go/v3 supporting project-version=2? If not, I do not think that we need to add this support now. 
We do not need to add backwards compatibility requirements for this one which is new. Otherwise, it means that we will need to support prject-version 2 until the v3 version of this plugin.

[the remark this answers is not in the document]
Adirio: Being it in a bundle it will already require it to be project-version 3, and this PRs doesn't really require project-version 3 itself, but I'm fine leaving it as is.

pkg/plugins/common/config-base/v1/init.go
camilamacedo86: This plugin requires that. That is common no matter the language. 
See: https://deploy-preview-2115--kubebuilder.netlify.app/plugins/creating-plugins.html#language-based-plugins

What we could make here specific would be only the componentConfig option. However, it is also common no matter the language and then, because of it, it is here.

cmd/main.go
camilamacedo86: Closing that since it was decided via the pool. Thank you folks,

pkg/plugins/common/config-base/v1/init.go
camilamacedo86: I am closing this one as addressed, however,  please feel free to ping me for we speak about this one if you see that its required.

cmd/main.go
camilamacedo86: I think that we are all on the same page now. However, if not. please feel free to re-open this one or ping me out.

cmd/main.go
```diff
@@ -22,22 +22,33 @@ import (
 	"sigs.k8s.io/kubebuilder/v3/pkg/cli"
 	cfgv2 "sigs.k8s.io/kubebuilder/v3/pkg/config/v2"
 	cfgv3 "sigs.k8s.io/kubebuilder/v3/pkg/config/v3"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
+	kustomizecommonv1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/common/kustomize-common/v1"
```
Adirio, line 26: `kustomize-common` directory name is redundant with the parent `common` directory, can we rename the dir to `kustomize`?
```suggestion
	kustomizecommonv1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/common/kustomize/v1"
```

pkg/plugins/common/kustomize-common/v1/init.go
```diff
@@ -0,0 +1,106 @@
...
+	config config.Config
+
+	// config options
+	domain          string
+	name            string
+	componentConfig bool
+}
+
+func (p *initSubcommand) UpdateMetadata(cliMeta plugin.CLIMetadata, subcmdMeta *plugin.SubcommandMetadata) {
+
```
Adirio, line 46: NIT: remove blank line
```suggestion
```

pkg/plugins/common/config-base/v1/init.go
Adirio: I feel like these options could be a better fit for an additional plugin (I think we shouldn't care about plugins being too small, so a plugin with only these options is fine IMO) but we can do that in a follow up.

The reason why several small plugins are not an issue is because we use the bundle concept so that the user doesn't have to write a long list of plugins, which would be the main drawback of multiple plugins (the UX).

Adirio (dismissed):
Some final comments

pkg/plugins/common/kustomize/v1/init.go
```diff
@@ -0,0 +1,105 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
estroz, line 2: Correct license years for all new files
```suggestion
Copyright 2021 The Kubernetes Authors.
```

pkg/plugins/golang/declarative/v1/api.go
```diff
@@ -134,10 +134,13 @@ func (p *createAPISubcommand) Scaffold(fs machinery.Filesystem) error {
...
 
 	// Ensure that we are pinning sigs.k8s.io/kubebuilder-declarative-pattern version
-	kbDeclarativePattern := kbDeclarativePatternForV2
+	// Just pin an old value for go/v2. It shows fine for now. However, we should improve/change it
+	// if we see that more rules based on the plugins version are required.
+	kbDeclarativePattern := kbDeclarativePatternForV3
 	for _, pluginKey := range p.config.GetPluginChain() {
-		if pluginKey == plugin.KeyFor(goPluginV3.Plugin{}) {
-			kbDeclarativePattern = kbDeclarativePatternForV3
+		if pluginKey == plugin.KeyFor(goPluginV2.Plugin{}) || pluginKey == "go.kubebuilder.io/v2" {
```
estroz, line 141: There is no difference between the output of `KeyFor(goPluginV2.Plugin{})` and this string

pkg/plugins/golang/v3/scaffolds/init.go
```diff
@@ -127,9 +112,5 @@ func (s *initScaffolder) Scaffold() error {
 		&templates.Dockerfile{},
 		&templates.DockerIgnore{},
 		&kdefault.Kustomization{},
```
estroz, line 114: This also needs to be moved.

estroz (changes requested):
All kustomize-related files should be in the kustomize plugin

pkg/plugins/golang/v3/scaffolds/internal/templates/config/kdefault/kustomization.go
```diff
@@ -39,7 +39,9 @@ func (f *Kustomization) SetTemplateDefaults() error {
 
 	f.TemplateBody = kustomizeTemplate
 
-	f.IfExistsAction = machinery.Error
+	// the common base does not have cert-manager or webhooks
```
estroz, line 42: It should. Any operator type can implement webhooks and use cert-manager.

docs/book/src/component-config-tutorial/testdata/project/config/certmanager/certificate.yaml
```diff
@@ -1,25 +0,0 @@
-# The following manifests contain a self-signed issuer CR and a certificate CR.
```
estroz, line 1: Why were these files deleted?

pkg/plugins/golang/declarative/v1/api.go
estroz: @camilamacedo86 the go/v2 plugin no longer has “base” in the name right?

docs/book/src/component-config-tutorial/testdata/project/config/certmanager/certificate.yaml
camilamacedo86: good catcher. It is a mistake.

Adirio: Those files were mopve to create webhook in a previous PR IIRC.

estroz (approved):
Nice work!

/lgtm

PR #2107, opened 2021-03-25
🌱 Split GH Actions jobs into separate workflows

Adirio:
### Advantages:
- Re-run are workflow-scoped, so having different workflows allows for better control on which to jobs to re-run.
- Allows finer-grained customization of triggers per workflow.

### Drawbacks:
- Unable to require one workflow to have succeeded before running other.


camilamacedo86 (approved):
/lgtm

PR #2111, opened 2021-03-25
:bug: do not generate the cert-manager directory by default. 

camilamacedo86:
**Description**
do not generate the cert-manager directory by default. 

**Motivation**
cert-manager directory should only be scaffold when a webhook is used


pkg/plugins/golang/v3/scaffolds/webhook.go
```diff
@@ -18,6 +18,7 @@ package scaffolds
 
 import (
 	"fmt"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v3/scaffolds/internal/templates/config/certmanager"
```
Adirio, line 21: This import needs to be sorted properly.

Adirio (approved):

PR #2112, opened 2021-03-26
✨ Allow bundles to be used as input to other bundles

Adirio:
A `Bundle` is also a `Plugin`, but as unbundling is not done recursively, when using a `Bundle` as the input of other `Bundle`, it is not properly unbundled.
This PRs solves this by unbundling Bundles at Bundle creation time to avoid having to use recursive unbundling.


camilamacedo86 (commented):
/lgtm

PR #2114, opened 2021-03-26
:warning: move declarative plugin to golang domain

camilamacedo86:
**Description**
- move the declarative plugin to golang domain

**Motivation**
it is an specific plugin that is only valid for Golang so it is better we clarify that in the API. We might have others plugins that can be common in the future such as config-gen


Adirio (approved):
/lgtm

Issue #2125, opened 2021-04-05
Regression: extra commands require a valid PROJECT file again


jmrodri:
https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/cli.go#L194 returns an error when it can not parse the config file. This means that extra commands will no longer work. This affects the `operator-sdk` because we have a command `alpha config-3alpha-to-3` that will convert `3alpha` PROJECT config to valid 3 config. But the error prevents this command from ever running.

Similarly, `version` also no longer works if we can't parse the PROJECT config file which is a regression.

What versions of software are you using?  Specifically, the following are often useful:

* go version
* kubebuilder version (`kubebuilder version`) and scaffolding version (check your `PROJECT` file): *v3.0.0-beta.1`
* controller-runtime version (check your `go.mod` file)
* controller-tools version
* Kubernetes & kubectl versions (just run `kubectl version` against your API server)

/kind bug

Adirio:
I would not consider this a regression as the previous case was when the config file didn't exist.

PR #2126, opened 2021-04-06
:bug: regression - extracommands require a valid PROJECT file again

jmrodri:
Ignore UnsupportedVersion error to allow extra commands to be able to actually run.

Fixes: #2125

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -29,6 +29,7 @@ import (
 	"sigs.k8s.io/kubebuilder/v3/pkg/config"
 	yamlstore "sigs.k8s.io/kubebuilder/v3/pkg/config/store/yaml"
 	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
+	"sigs.k8s.io/kubebuilder/v3/pkg/model/stage"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
 )
 
@@ -144,8 +145,21 @@ func newCLI(options ...Option) (*CLI, error) {
 func (c *CLI) buildCmd() error {
 	c.cmd = c.newRootCmd()
 
+	var uve config.UnsupportedVersionError
+
 	// Get project version and plugin keys.
-	if err := c.getInfo(); err != nil {
+	switch err := c.getInfo(); {
+	case err == nil:
+	case errors.As(err, &uve) && uve.Version.Compare(config.Version{Number: 3, Stage: stage.Alpha}) == 0:
+		// Check if the corresponding stable version exists, set c.projectVersion and break
+		stableVersion := config.Version{
+			Number: uve.Version.Number,
+		}
+		if config.IsRegistered(stableVersion) {
+			// Use the stableVersion
+			c.projectVersion = stableVersion
+		}
+	default:
 		return err
 	}
 
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -158,6 +158,9 @@ func (c *CLI) buildCmd() error {
 		if config.IsRegistered(stableVersion) {
 			// Use the stableVersion
 			c.projectVersion = stableVersion
+		} else {
+			// stable version not registered, let's bail out
+			return err
 		}
 	default:
 		return err
--- a/pkg/cli/cli_test.go
+++ b/pkg/cli/cli_test.go
@@ -32,6 +32,7 @@ import (
 	cfgv2 "sigs.k8s.io/kubebuilder/v3/pkg/config/v2"
 	cfgv3 "sigs.k8s.io/kubebuilder/v3/pkg/config/v3"
 	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
+	"sigs.k8s.io/kubebuilder/v3/pkg/model/stage"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
 	goPluginV3 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v3"
 )
@@ -89,6 +90,40 @@ var _ = Describe("CLI", func() {
 		c = &CLI{
 			fs: machinery.Filesystem{FS: afero.NewMemMapFs()},
 		}
+	})
+
+	Context("buildCmd", func() {
+		BeforeEach(func() {
+			projectFile := `domain: zeusville.com
+layout: go.kubebuilder.io/v3
+projectName: demo-zeus-operator
+repo: github.com/jmrodri/demo-zeus-operator
+resources:
+- crdVersion: v1
+  group: test
+  kind: Test
+  version: v1
+version: 3-alpha
+plugins:
+  manifests.sdk.operatorframework.io/v2: {}
+`
+			f, err := c.fs.FS.Create("PROJECT")
+			Expect(err).To(Not(HaveOccurred()))
+
+			_, err = f.WriteString(projectFile)
+			Expect(err).To(Not(HaveOccurred()))
+		})
+
+		When("reading a 3-alpha config", func() {
+			It("should succeed and set the projectVersion", func() {
+				err := c.buildCmd()
+				Expect(err).To(Not(HaveOccurred()))
+				Expect(c.projectVersion.Compare(
+					config.Version{
+						Number: 3,
+						Stage:  stage.Stable})).To(Equal(0))
+			})
+		})
 	})
 
 	// TODO: test CLI.getInfoFromConfigFile using a mock filesystem
--- a/pkg/cli/cli_test.go
+++ b/pkg/cli/cli_test.go
@@ -93,8 +93,10 @@ var _ = Describe("CLI", func() {
 	})
 
 	Context("buildCmd", func() {
+		var projectFile string
+
 		BeforeEach(func() {
-			projectFile := `domain: zeusville.com
+			projectFile = `domain: zeusville.com
 layout: go.kubebuilder.io/v3
 projectName: demo-zeus-operator
 repo: github.com/jmrodri/demo-zeus-operator
@@ -122,6 +124,17 @@ plugins:
 					config.Version{
 						Number: 3,
 						Stage:  stage.Stable})).To(Equal(0))
+			})
+			It("should fail when stable is not registered ", func() {
+				// overwrite project file with fake 4-alpha
+				f, err := c.fs.FS.OpenFile("PROJECT", os.O_WRONLY, 0)
+				Expect(err).To(Not(HaveOccurred()))
+				_, err = f.WriteString(strings.ReplaceAll(projectFile, "3-alpha", "4-alpha"))
+				Expect(err).To(Not(HaveOccurred()))
+
+				// buildCmd should return an error
+				err = c.buildCmd()
+				Expect(err).To(HaveOccurred())
 			})
 		})
 	})
--- a/pkg/config/registry.go
+++ b/pkg/config/registry.go
@@ -25,6 +25,12 @@ func Register(version Version, constructor func() Config) {
 	registry[version] = constructor
 }
 
+// IsRegistered returns true if the given version has been registered through Register
+func IsRegistered(version Version) bool {
+	_, ok := registry[version]
+	return ok
+}
+
 // New creates Config instances from the previously registered implementations through Register
 func New(version Version) (Config, error) {
 	if constructor, exists := registry[version]; exists {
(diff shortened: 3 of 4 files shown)


jmrodri:
@estroz I'm having trouble creating a test for this. I need a way to pass in an invalid PROJECT file. I started adding a `testdata/v3alpha-PROJECT` but then couldn't immediately find the place to pass that in.

pkg/cli/cli.go
```diff
@@ -145,8 +146,16 @@ func (c *CLI) buildCmd() error {
...
 
 	// Get project version and plugin keys.
-	if err := c.getInfo(); err != nil {
-		return err
+	err := c.getInfo()
+	if err != nil {
+		if le, ok := err.(store.LoadError); ok {
+			// ignore UnsupportedVersionError, I'd prefer to verify a specific type
+			// of error instead of the string, but alas LoadError consists wrapped
+			// errors.
```
Adirio, line 154: As you say here, this should be checked using a typed error and not an string. We already have a suitable typed error: https://github.com/kubernetes-sigs/kubebuilder/blob/84f357b50e0b25e64c1e5918089f6588b79b81f1/pkg/config/errors.go#L26-L28

So, IMO, changes should be done as follows:

https://github.com/kubernetes-sigs/kubebuilder/blob/84f357b50e0b25e64c1e5918089f6588b79b81f1/pkg/config/store/yaml/store.go#L93
should be
```go
return store.LoadError{Err: config.UnsupportedVersionError{Version: versioned.Version}} 
```

and the check should use `errors.Is` or `errors.As`, not a type assertion. By the way, I would also prefer a:
```go
switch err := c.getInfo(); {
case err == nil:
[...]
}
```
instead of nested if conditionals.

Adirio (changes requested):
I don't think that omitting the error is the best approach, as this will not set the CLI's project-version, thus trying to obtain it from `--project-version` (which would probably be not provided) and then trying to resolve it automatically from the list of plugins, which may not be possible, and thus still resulting in an error. I think that we should probably set CLI's project-version to 3 if 3-alpha is encountered inside the `switch` statement I mention below.

Adirio:
Just a small note that was commented during the triage meeting. As this variable is used for plugin resolution, I nice approach could be once a version is not found and seeing that the version is unstable (alpha/beta), if the corresponding stable version exists, set that so that plugin resolution can still work properly.

jmrodri:
The result of using this PR with the SDK is:
```
$ operator-sdk version
WARN[0000] Config version 3-alpha has been stabilized as 3, and 3-alpha is no longer supported. Run `operator-sdk alpha config-3alpha-to-3` to upgrade your PROJECT config file to version 3 
operator-sdk version: "scorecard-kuttl/v2.0.0-44-g2b2f989d", commit: "2b2f989d2a47b343d71e77c0963537e271a87979", kubernetes version: "v1.19.4", go version: "go1.15.8", GOOS: "linux", GOARCH: "amd64"

$ operator-sdk alpha config-3alpha-to-3 
WARN[0000] Config version 3-alpha has been stabilized as 3, and 3-alpha is no longer supported. Run `operator-sdk alpha config-3alpha-to-3` to upgrade your PROJECT config file to version 3 
Your PROJECT config file has been converted from version 3-alpha to 3. Please make sure all config data is correct.

$ operator-sdk version
operator-sdk version: "scorecard-kuttl/v2.0.0-44-g2b2f989d", commit: "2b2f989d2a47b343d71e77c0963537e271a87979", kubernetes version: "v1.19.4", go version: "go1.15.8", GOOS: "linux", GOARCH: "amd64"
```

pkg/cli/cli.go
```diff
@@ -145,7 +146,10 @@ func (c *CLI) buildCmd() error {
 	c.cmd = c.newRootCmd()
 
 	// Get project version and plugin keys.
-	if err := c.getInfo(); err != nil {
+	// Ignore 3-alpha error
+	switch err := c.getInfo(); {
+	case err == nil:
+	case !errors.Is(err, config.UnsupportedVersionError{Version: config.Version{Number: 3, Stage: stage.Alpha}}):
```
Adirio, line 152: What do you think of something a little bit more generic like this? (You need to declare the variable `unsupportedVersionErr` of type `config.UnsupportedVersionError` above the `switch`)
```suggestion
	case errors.As(err, &unsupportedVersionErr):
		if !unsupportedVersionErr.Version.IsStable() {
			// Check if the corresponding stable version exists, set c.projectVersion and break
		}
		fallthrough
	default:
```

pkg/cli/cli.go
```diff
@@ -145,7 +146,10 @@ func (c *CLI) buildCmd() error {
 	c.cmd = c.newRootCmd()
 
 	// Get project version and plugin keys.
-	if err := c.getInfo(); err != nil {
+	// Ignore 3-alpha error
+	switch err := c.getInfo(); {
```
estroz, line 150:
```suggestion
	switch err := c.getInfo(); err {
```

pkg/cli/cli.go
estroz: How about returning any `UnsupportedVersionError` to the caller so they can handle it appropriately?

Also this doesn't solve the general case of a command that can be run outside of a project.

pkg/config/store/yaml/store.go
```diff
@@ -90,7 +90,7 @@ func (s *yamlStore) LoadFrom(path string) error {
 	var cfg config.Config
 	cfg, err = config.New(versioned.Version)
 	if err != nil {
-		return store.LoadError{Err: fmt.Errorf("unable to create config for version %q: %w", versioned.Version, err)}
+		return store.LoadError{Err: config.UnsupportedVersionError{Version: versioned.Version}}
```
estroz, line 93: `config.New()` should be in charge of returning `UnsupportedVersionError` since this caller does not know why the error was emitted.

jmrodri: @estroz `config.New()` does return the UnsupportedVersionError already: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/config/registry.go#L34

jmrodri: So `err` in the original code: `fmt.Errorf("unable to create config for version %q: %w", versioned.Version, err)` is an UnsupportedVersionError . So I can totally just return the err as is.

pkg/cli/cli.go
Adirio: That was already solved if I understood you correctly. The problem is faced when a PROJECT with a unknown version is found (`3-alpha`) not when no PROJECT file is found. And returning an error makes the command fail, we need to skip for 3-alpha (or in the generic solution I propose here, for any unregistered unstable version whose stable version is registered).

pkg/cli/cli.go
Adirio: Nope, that would only allow to compare exactly against `err`, it doesn't allow `errors.As`.

pkg/cli/cli.go
jmrodri: @Adirio @estroz correct, a command that has no PROJECT file already works. That is checked explicitly.

camilamacedo86 (approved):
👍

pkg/cli/cli.go
```diff
@@ -144,8 +144,14 @@ func newCLI(options ...Option) (*CLI, error) {
...
+	var uve config.UnsupportedVersionError
+
 	// Get project version and plugin keys.
-	if err := c.getInfo(); err != nil {
+	switch err := c.getInfo(); {
+	case err == nil:
+		// do nothing if it's nil
+	case !errors.As(err, &uve):
+		// Ignore 3-alpha error
 		return err
```
estroz, line 155: @Adirio I think 3-alpha is a special case; in general we would want to treat non-existent project versions as actual errors right?
```suggestion
	case !errors.As(err, &uve) || uve.Version.String() != "3-alpha":
		// Ignore 3-alpha version error only.
		return err
```

Adirio: The condition would be like this right?
```suggestion
	case errors.As(err, &uve) && uve.Version.Compare(config.Version{Number: 3, Stage: stage.Alpha}) == 0:
		// Ignore 3-alpha error
	default:
		return err
```

jmrodri:
```
# verify PROJECT file has 3-alpha
[jesusr@transam demo-zeus-operator{master}]$ grep 3-alpha PROJECT 
version: 3-alpha

# verify version still works when it can't parse the 3-alpha config
[jesusr@transam demo-zeus-operator{master}]$ operator-sdk version
WARN[0000] Config version 3-alpha has been stabilized as 3, and 3-alpha is no longer supported. Run `operator-sdk alpha config-3alpha-to-3` to upgrade your PROJECT config file to version 3 
operator-sdk version: "scorecard-kuttl/v2.0.0-44-g2b2f989d", commit: "2b2f989d2a47b343d71e77c0963537e271a87979", kubernetes version: "v1.19.4", go version: "go1.15.8", GOOS: "linux", GOARCH: "amd64"

# verify conversion works
[jesusr@transam demo-zeus-operator{master}]$ operator-sdk alpha config-3alpha-to-3
WARN[0000] Config version 3-alpha has been stabilized as 3, and 3-alpha is no longer supported. Run `operator-sdk alpha config-3alpha-to-3` to upgrade your PROJECT config file to version 3 
Your PROJECT config file has been converted from version 3-alpha to 3. Please make sure all config data is correct.

[jesusr@transam demo-zeus-operator{master}]$ grep version: PROJECT 
  version: v1
version: "3"
```

jmrodri:
Using a completely invalid config results in an error still:
```
grep[jesusr@transam demo-zeus-operator{master}]$ grep version PROJECT                     
  version: v1
version: 4-alpha         
                  
[jesusr@transam demo-zeus-operator{master}]$ operator-sdk alpha config-3alpha-to-3
Error: unable to load the configuration: unable to create config for version "4-alpha": version 4-alpha is not supported
Usage:
  operator-sdk [flags]
...
FATA[0000] unable to load the configuration: unable to create config for version "4-alpha": version 4-alpha is not supported 
```

Adirio:
> Using a completely invalid config results in an error still:
> 
> ```
> grep[jesusr@transam demo-zeus-operator{master}]$ grep version PROJECT                     
>   version: v1
> version: 4-alpha         
>                   
> [jesusr@transam demo-zeus-operator{master}]$ operator-sdk alpha config-3alpha-to-3
> Error: unable to load the configuration: unable to create config for version "4-alpha": version 4-alpha is not supported
> Usage:
>   operator-sdk [flags]
> ...
> FATA[0000] unable to load the configuration: unable to create config for version "4-alpha": version 4-alpha is not supported 
> ```

That is because `config-3alpha-to-3` checks exactly for `3-alpha`

pkg/cli/cli.go
```diff
@@ -144,8 +145,21 @@ func newCLI(options ...Option) (*CLI, error) {
...
+	switch err := c.getInfo(); {
+	case err == nil:
+	case errors.As(err, &uve) && uve.Version.Compare(config.Version{Number: 3, Stage: stage.Alpha}) == 0:
+		// Check if the corresponding stable version exists, set c.projectVersion and break
+		stableVersion := config.Version{
+			Number: uve.Version.Number,
+		}
+		if config.IsRegistered(stableVersion) {
+			// Get project version and plugin info from defaults
+			c.getInfoFromDefaults()
```
Adirio, line 160:
```suggestion
			c.projectVersion = stableVersion
```

jmrodri: Sorry went back and forth on this. That makes more sense.

pkg/cli/cli.go
```diff
@@ -144,8 +145,21 @@ func newCLI(options ...Option) (*CLI, error) {
...
+	case err == nil:
+	case errors.As(err, &uve) && uve.Version.Compare(config.Version{Number: 3, Stage: stage.Alpha}) == 0:
+		// Check if the corresponding stable version exists, set c.projectVersion and break
+		stableVersion := config.Version{
+			Number: uve.Version.Number,
+		}
+		if config.IsRegistered(stableVersion) {
+			// Use the stableVersion
+			c.projectVersion = stableVersion
+		}
```
Adirio, line 161: I think this is my last comment. Shouldn't it return an error if the stable wasn't found?
```suggestion
		} else {
			return err
		}
```

jmrodri: @Adirio I don't mind the comments, I really found them valuable.

jmrodri: Returned error and added a test case for that condition.

Adirio (approved):
/lgtm

Issue #2128, opened 2021-04-07
`make test` fails in docs/book/src/cronjob-tutorial/testdata/project


dashpole:
From HEAD, running `make test` from docs/book/src/cronjob-tutorial/testdata/project fails with:

`SideEffects is required for creating v1 {Mutating,Validating}WebhookConfiguration`

If I fix that by adding the `sideEffects=None` tag to webhooks, it fails with:

`AdmissionReviewVersions is mandatory for v1 {Mutating,Validating}WebhookConfiguration`

If I fix that by adding the `admissionReviewVersions=v1` tag to webhooks, it fails during the e2e test with:

```
2021-04-07T07:50:46.247-0700	ERROR	controller-runtime.manager.controller.cronjob	Reconciler error	{"reconciler group": "batch.tutorial.kubebuilder.io", "reconciler kind": "CronJob", "name": "test-cronjob", "namespace": "default", "error": "field label \".metadata.controller\" not supported for batchv1.Job"}
```

And thats where I got stuck.  Any pointers would be appreciated.  I ran into the "field label not supported" issue on my own project, and don't know how to proceed.

* go version: 1.16
* kubebuilder version (`kubebuilder version`) and scaffolding version (check your `PROJECT` file): kubebuilder 2.3.2, [PROJECT file](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/project/PROJECT)
* controller-runtime version (check your `go.mod` file): [go.mod](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/project/go.mod) v0.7.0
* Kubernetes & kubectl versions (just run `kubectl version` against your API server): kubectl v1.16.4

/kind bug

cc @DirectXMan12

camilamacedo86:
Hi @dashpole, 

We need to update the projects in the book dir. 
See that we have shell scripts to re-generate them and then we can use an IDE to compare with the master branch and put back all comments/docs text on top. 

Would you like to work on that ?

dashpole:
Which script?

I'm mostly interested in figuring out how to fix the error: `"field label \".metadata.controller\" not supported for batchv1.Job"`, since i'm hitting that in my own project.

camilamacedo86:
You did not execute `make install` first. Let me know if you are able to do that after it.

dashpole:
After that, I get:
```
removing project ...
[...]

PR #2129, opened 2021-04-07
Fix missing attributes in the demo: 🐛 

dashpole:
This adds missing attributes required to run `make test` on the cronjob webhook demo.
Issue: #2128
It doesn't fully fix the issue, just the parts I know how to fix


camilamacedo86 (changes requested):
We need to use a tool to re-generate the projects via the helper's shell script that are in the dirs. See my comment in; https://github.com/kubernetes-sigs/kubebuilder/issues/2128#issuecomment-815016615

Also, feel free to reach out with you would like to do that and you need any help.

PR #2066, opened 2021-03-06
📖 Use client.New instead of manager client in tests

leovct:
**Description**
This PR improves the testing section of the documentation.
The idea is to recommend to use a strongly consistent client ```client.New``` instead of the manager's client in the case of tests.

**Motivation**
* Closes: #2018
* It is highly inspired by this discussion https://github.com/kubernetes-sigs/controller-runtime/issues/343.

Issue #2018, opened 2021-02-15
Don't recommend using manager.GetClient in tests


moolen:
Using the `manager.GetClient()` in tests is a bad idea and should not be recommended by the documentation. The manager's client is a cached `client.Reader` which is fine for a controller - but in tests you probably want a strongly consistent client.

This is the documentation i am referring to:
https://github.com/kubernetes-sigs/kubebuilder/blame/master/docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go#L115-L116

There's an older issue that explains the problem in more detail: https://github.com/kubernetes-sigs/controller-runtime/issues/343

I'd suggest updating the documentation about that. What do you think? Did anybody else run into that issue yet?

/kind bug
/kind documentation

Adirio:
Any help with the documentaiton is appreciated
/help
/good-first-issue

P.S.: the most relevant part of the linked conversation is https://github.com/kubernetes-sigs/controller-runtime/issues/343#issuecomment-469435686

leovct:
Hello @Adirio, I would like to contribute to the project by working on this issue :)

Adirio:
Sure, help is always apreciated. Feel free to submit a PR and assign me so that I can review it.

DirectXMan12:
To be clear for posterity (because the follow-up PR broke the tests): You shouldn't use the manager's client in *assertions* in tests.  You *should* use the manager's client in your reconciler.

PR #2131, opened 2021-04-07
:book: Fix advice on clients in tests, ensure that book gets tested

### LINKED ISSUE #2128: `make test` fails in docs/book/src/cronjob-tutorial/testdata/project

DirectXMan12:
As per #2128, the book's `make test` target did not run.  There were two issues at hand:

1. The book was not updated to use the `sideEffects` and `admissionReviewVersions` fields in the `+kubebuilder:webhook` marker (required for v1 webhooks)

2. #2066 incorrectly interpreted advice on test clients and switch the reconciler over to use a live client (the reconciler should always use the manager's client, and in this case has to, because it relies on indicies).

This fixes both those issues, and also adds the book's tests to `make test` so that we don't break this in the future, and makes sure the CI is actually running `make test` instead of a copy of the same actions.

Fixes #2128 
Related to #2066 
Related to #2018

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ golangci-lint:
 ##@ Tests
 
 .PHONY: test
-test: test-unit test-integration test-testdata ## Run the unit and integration tests (used in the CI)
+test: test-unit test-integration test-testdata test-book ## Run the unit and integration tests (used in the CI)
 
 .PHONY: test-unit
 test-unit: ## Run the unit tests
@@ -119,3 +119,7 @@ test-e2e-local: ## Run the end-to-end tests locally
 .PHONY: test-e2e-ci
 test-e2e-ci: ## Run the end-to-end tests (used in the CI)`
 	./test/e2e/ci.sh
+
+.PHONY: test-book
+test-book: ## Run the cronjob tutorial's unit tests to make sure we don't break it
+	cd ./docs/book/src/cronjob-tutorial/testdata/project && make test
--- a/docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/controllers/suite_test.go
@@ -115,14 +115,16 @@ var _ = BeforeSuite(func() {
 		The only difference is that the manager is started in a separate goroutine so it does not block the cleanup of envtest
 		when you’re done running your tests.
 
-		It is not recommended to use the manager client in tests because it is not strongly consistent. Indeed, the manager
-		client is designed to do the "right thing" for controllers by default which is to read from caches. The best solution
-		is to instantiate a new client using client.New for tests (as k8sClient above). It will provide a client that reads
-		directly from the API meaning that you can write tests expecting read-after-write consistency.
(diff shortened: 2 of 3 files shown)


camilamacedo86 (approved):
@DirectXMan12, 

I am can Ok with this one since seems that it is a blocker for something. 

However, we need to re-generate the book test data samples as we did in the past. We have shell scripts to help with that.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -54,7 +54,7 @@ This marker is responsible for generating a mutating webhook manifest.
 The meaning of each marker can be found [here](/reference/markers/webhook.md).
 */
```
camilamacedo86, line 56: We need to use the helpers to re-generate all. That was the best "manually" solution so far until we find a better way to do that.

camilamacedo86:
/hold

I am working now to push that by updating the full book testdata using the helpers.

camilamacedo86:
@DirectXMan12, 

It is solved here; https://github.com/kubernetes-sigs/kubebuilder/pull/2133

estroz (approved):
/lgtm

DirectXMan12:
uhhhh, prow?  This is up-to-date afaict.

DirectXMan12:
like, I just ran `git fetch upstream master:master && git rebase master` and I get "current branch is up to date"

DirectXMan12:
hmm...

PR #2143, opened 2021-04-14
:seedling: Use issue forms & discussions instead of templates

DirectXMan12:
This switches our issues to use the (beta) issue forms instead of just
templates & configures links in the new issues page to CR, CT, and
GH discussions.


estroz (approved):
Awesome! TIL about the new format.

/lgtm

DirectXMan12:
> Awesome! TIL about the new format.

It's in a kinda partial beta period right now, I believe

PR #2151, opened 2021-04-19
:bug: (go/v3) moved leases.coordination.k8s.io to its own proxy-role rule

estroz:
From https://github.com/kubernetes-sigs/kubebuilder/pull/1809#issuecomment-822715429:

>Echoing a comment from Slack so it doesn't get lost, we should probably do a follow up to this where we split this object to its own rule chunk in the role so it's not granting the non-existent perms core/leases and leases.coordination.k8s.io/configmap

/kind bug

### DIFF
--- a/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/rbac/auth_proxy_client_role.go
+++ b/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/rbac/auth_proxy_client_role.go
@@ -45,6 +45,8 @@ kind: ClusterRole
 metadata:
   name: metrics-reader
 rules:
-- nonResourceURLs: ["/metrics"]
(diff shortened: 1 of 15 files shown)


pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/rbac/auth_proxy_client_role.go
```diff
@@ -45,6 +45,8 @@ kind: ClusterRole
 metadata:
   name: metrics-reader
 rules:
-- nonResourceURLs: ["/metrics"]
-  verbs: ["get"]
+- nonResourceURLs:
```
estroz, line 48: I also normalized formatting for all role rules. Having a mix of 2 list format styles can be confusing to read.

Issue #2141, opened 2021-04-12
`create webhook`: return error when used with a core type


estroz:
When I run `kubebuilder create webhook` on a core type, ex. `v1.Pod`, the command attempts to define methods on a non-existent locally-defined type. Instead, the command should return an error explaining that core types are not supported by this command.

```console
$ go version
go version go1.15.2 linux/amd64
$ kubebuilder version
Version: main.version{KubeBuilderVersion:"v3.0.0-alpha.0-195-g37485126", KubernetesVendor:"unknown", GitCommit:"37485126f95ceb2e92a86497c31f400c8b89f957", BuildDate:"2021-04-12T18:30:56Z", GoOs:"linux", GoArch:"amd64"}
$ cat go.mod | grep controller-runtime
sigs.k8s.io/controller-runtime v0.7.2
```

/kind bug

camilamacedo86:
IHMO, we ought to support that. 
So, that means: 

- create api properly work with core/external types: RFE: https://github.com/kubernetes-sigs/kubebuilder/issues/2141

AND that means

- change here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/webhook.go#L107 for:

- If is core type then update the PROJECT file accordingly and scaffold the webhook
- if is external-type (we need a new flag for that) then, do the same. 

So, IMO we should **not** spend effort in change the message but actually in moving forward with the RFE https://github.com/kubernetes-sigs/kubebuilder/issues/2141

estroz:
This solution is more of a holdover until something like what is suggested [here](https://book.kubebuilder.io/reference/webhook-for-core-types.html) is implemented and kustomize-able resources are generated correctly for core/external types.

Alternatively the holdover can be skipped entirely if the `Handler` implementation can be completed without an EP, which I think it can.

@kopiczko if you'd like to work on either this and/or implement core/external type support in `create webhook`, feel free to assign yourself!

kopiczko:
Cool, thanks, I'll look into this and check what I can do in the following days. Possibly starting wit the error message is a good warmup for me.

kopiczko:
Hi @estroz,

I was trying to think what would be the good way to fix this and found that doesn't only apply to the core types.

In general:

```
$ kubebuilder create api --group "mygroup" --version "v1" --kind "MyKind" --controller=true --resource=false
Writing scaffold for you to edit...
controllers/mykind_controller.go
Update dependencies:
[...]

PR #2152, opened 2021-04-23
:bug: Fail when creating api webhook for a core type

kopiczko:
This change makes:

- `create api` fail when `--resource` is set and the API is not already scaffolded for a core type
- `create webhook` fail for a core type

Fixes https://github.com/kubernetes-sigs/kubebuilder/issues/2141

### DIFF
--- a/pkg/plugins/golang/options.go
+++ b/pkg/plugins/golang/options.go
@@ -106,24 +106,21 @@ func (opts Options) UpdateResource(res *resource.Resource, c config.Config) {
 	}
 
 	// domain and path may need to be changed in case we are referring to a builtin core resource:
-	//  - Check if we are scaffolding the resource now           => project resource
 	//  - Check if we already scaffolded the resource            => project resource
 	//  - Check if the resource group is a well-known core group => builtin core resource
 	//  - In any other case, default to                          => project resource
 	// TODO: need to support '--resource-pkg-path' flag for specifying resourcePath
-	if !opts.DoAPI {
-		var alreadyHasAPI bool
-		if c.GetVersion().Compare(cfgv2.Version) == 0 {
-			alreadyHasAPI = c.HasResource(res.GVK)
-		} else {
-			loadedRes, err := c.GetResource(res.GVK)
-			alreadyHasAPI = err == nil && loadedRes.HasAPI()
-		}
-		if !alreadyHasAPI {
-			if domain, found := coreGroups[res.Group]; found {
-				res.Domain = domain
-				res.Path = path.Join("k8s.io", "api", res.Group, res.Version)
-			}
+	var alreadyHasAPI bool
+	if c.GetVersion().Compare(cfgv2.Version) == 0 {
+		alreadyHasAPI = c.HasResource(res.GVK)
+	} else {
+		loadedRes, err := c.GetResource(res.GVK)
+		alreadyHasAPI = err == nil && loadedRes.HasAPI()
+	}
+	if !alreadyHasAPI {
+		if domain, found := coreGroups[res.Group]; found {
+			res.Domain = domain
+			res.Path = path.Join("k8s.io", "api", res.Group, res.Version)
 		}
 	}
 }
--- a/pkg/plugins/golang/v2/api.go
+++ b/pkg/plugins/golang/v2/api.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/spf13/pflag"
 
@@ -133,6 +134,11 @@ func (p *createAPISubcommand) InjectResource(res *resource.Resource) error {
 
 	// In case we want to scaffold a resource API we need to do some checks
 	if p.options.DoAPI {
+		// Generating the API for a core resource is forbidden.
+		if strings.HasPrefix(p.resource.Path, "k8s.io") {
+			return errors.New("scaffolding API for core resources is not supported")
+		}
+
 		// Check that resource doesn't have the API scaffolded or flag force was set
 		if res, err := p.config.GetResource(p.resource.GVK); err == nil && res.HasAPI() && !p.force {
 			return errors.New("API resource already exists")
--- a/pkg/plugins/golang/v2/webhook.go
+++ b/pkg/plugins/golang/v2/webhook.go
@@ -101,7 +101,7 @@ func (p *createWebhookSubcommand) InjectResource(res *resource.Resource) error {
 			return fmt.Errorf("%s create webhook requires a previously created API ", p.commandName)
 		}
 	} else {
-		if r, err := p.config.GetResource(p.resource.GVK); err != nil {
+		if r, err := p.config.GetResource(p.resource.GVK); err != nil || !r.HasAPI() {
 			return fmt.Errorf("%s create webhook requires a previously created API ", p.commandName)
 		} else if r.Webhooks != nil && !r.Webhooks.IsEmpty() {
 			return fmt.Errorf("webhook resource already exists")
--- a/pkg/plugins/golang/v3/api.go
+++ b/pkg/plugins/golang/v3/api.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/spf13/pflag"
 
@@ -142,6 +143,11 @@ func (p *createAPISubcommand) InjectResource(res *resource.Resource) error {
 
 	// In case we want to scaffold a resource API we need to do some checks
 	if p.options.DoAPI {
+		// Generating the API for a core resource is forbidden.
+		if strings.HasPrefix(p.resource.Path, "k8s.io") {
+			return errors.New("scaffolding API for core resources is not supported")
+		}
+
 		// Check that resource doesn't have the API scaffolded or flag force was set
 		if r, err := p.config.GetResource(p.resource.GVK); err == nil && r.HasAPI() && !p.force {
 			return errors.New("API resource already exists")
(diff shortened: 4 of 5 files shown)


pkg/plugins/golang/options.go
```diff
@@ -106,24 +106,22 @@ func (opts Options) UpdateResource(res *resource.Resource, c config.Config) {
...
+	var alreadyHasAPI bool
+	if c.GetVersion().Compare(cfgv2.Version) == 0 {
+		alreadyHasAPI = c.HasResource(res.GVK)
+	} else {
+		loadedRes, err := c.GetResource(res.GVK)
+		alreadyHasAPI = err == nil && loadedRes.HasAPI()
+	}
+	if !alreadyHasAPI {
+		if domain, found := coreGroups[res.Group]; found {
+			res.Core = true
```
kopiczko, line 122: Here's it's indicated that this is a core resource. I wonder if this won't be changed when external support is added but I don't think it's a big deal to rename/change it later.

pkg/model/resource/resource.go
```diff
@@ -40,6 +40,9 @@ type Resource struct {
 	// Controller specifies if a controller has been scaffolded.
 	Controller bool `json:"controller,omitempty"`
 
+	// Core indicates if this is a core resource.
+	Core bool `json:"core,omitempty"`
```
kopiczko, line 44: I'm not sure why there are those json tags. Is it marshalled somewhere?

estroz: Yep, in the [config](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/config/v3/config.go#L70). I think we can get the same functionality as setting a bool if we check `strings.HasPrefix(Resource.Path, modulePath)` where `modulePath` is from go.mod. This might break for submodules though.

If we do decide to go with an extra field, I vote for
```suggestion
	External bool `json:"external,omitempty"`
```
Since it covers both cases. If the path starts with `k8s.io/` then we can be pretty sure it is core.

kopiczko: Thanks, thinking of that there is no point of introducing `.Core` because we can basically check `strings.HasPrefix(res.Path, "k8s.io")`.

Checking against module is also not hard. There is `golang.findGoModulePath` function already defined. It could be exported.

kopiczko: @estroz I went with simple prefix check. That can be changed if for any reason `.External` has to be introduced, but I'd prefer to avoid changing `PROJECT` because something _may_ be needed.

pkg/plugins/golang/v2/webhook.go
```diff
@@ -90,6 +90,11 @@ func (p *createWebhookSubcommand) InjectResource(res *resource.Resource) error {
 		return err
 	}
 
+	alreadyHasAPI := p.config.HasResource(res.GVK)
+	if !alreadyHasAPI {
+		return fmt.Errorf("%s create webhook requires scaffolded API", p.commandName)
+	}
+
```
Adirio, line 97: This only works with `PROJECT` version 2, while go/v2 plugin may also use `PROJECT` version 3.

pkg/plugins/golang/v3/webhook.go
```diff
@@ -97,6 +97,15 @@ func (p *createWebhookSubcommand) InjectResource(res *resource.Resource) error {
...
 	}
 
+	var alreadyHasAPI bool
+	{
+		loadedResource, err := p.config.GetResource(p.resource.GVK)
+		alreadyHasAPI = err == nil && loadedResource.HasAPI()
+	}
+	if !alreadyHasAPI {
+		return fmt.Errorf("%s create webhook requires scaffolded API", p.commandName)
+	}
```
Adirio, line 107: go/v3 plugin only supports `PROJECT` version 3 so this is correct.

Adirio (changes requested):

pkg/plugins/golang/v2/webhook.go
```diff
@@ -101,9 +101,11 @@ func (p *createWebhookSubcommand) InjectResource(res *resource.Resource) error {
 			return fmt.Errorf("%s create webhook requires a previously created API ", p.commandName)
 		}
 	} else {
-		if r, err := p.config.GetResource(p.resource.GVK); err != nil {
+		loadedResource, err := p.config.GetResource(p.resource.GVK)
+		if err != nil || !loadedResource.HasAPI() {
```
Adirio, line 105: I think that this condition now fails for `PROJECT` version 2. `go/v2` needs to support both `PROJECT` versions 2 and 3.

pkg/plugins/golang/v3/webhook.go
```diff
@@ -103,9 +103,11 @@ func (p *createWebhookSubcommand) InjectResource(res *resource.Resource) error {
...
 	// check if resource exist to create webhook
-	if r, err := p.config.GetResource(p.resource.GVK); err != nil {
+	loadedResource, err := p.config.GetResource(p.resource.GVK)
+	if err != nil || !loadedResource.HasAPI() {
 		return fmt.Errorf("%s create webhook requires a previously created API ", p.commandName)
-	} else if r.Webhooks != nil && !r.Webhooks.IsEmpty() && !p.force {
+	}
+	if loadedResource.Webhooks != nil && !loadedResource.Webhooks.IsEmpty() && !p.force {
 		return fmt.Errorf("webhook resource already exists")
 	}
```
kopiczko, line 112: After rereading it again I found that the check was already there but it had a bug: if the `err != nil` that doesn't guarantee the API is created. We need to check if the loaded resource `HasAPI()` as well.

So it now all works like this:

- `Path` is always set for core resources. `create api` fails if the `Path` starts with `k8s.io`
- `create webhook` fails if there is no scaffolded API

@Adirio @estroz PTALA

pkg/plugins/golang/v2/webhook.go
kopiczko: It supports both. Check the code before `else`.

Adirio: Oh right that is checked before.

Why not just add the `|| !loadedResource.HasAPI()` to the first if condition?

kopiczko: No reason. I thought new conditions can be added without extra else blocks that way. I can go with your suggestison.

pkg/plugins/golang/v3/api.go
```diff
@@ -142,6 +143,11 @@ func (p *createAPISubcommand) InjectResource(res *resource.Resource) error {
 
 	// In case we want to scaffold a resource API we need to do some checks
 	if p.options.DoAPI {
+		// Generating the API for a core resource is forbidden.
+		if strings.HasPrefix(p.resource.Path, "k8s.io") {
+			return errors.New("scaffolding API for core resources is not supported")
+		}
```
camilamacedo86, line 149: WDYT about if is core-type then we allow to do the scaffold and just update the PROJECT file?

kopiczko: Do I understand correctly you'd like to ignore the `--resource` flag completely for the core types and silently set it to false?

kopiczko: @camilamacedo86 I believe as a first iteration `create api` may fail when for a core resource. If the handling of the `--resource` flag changes in the future it will be a non-breaking change anyway if it fails now.

camilamacedo86: We can create api for a core type however that will be added to the project domain and is required to change manually. So, the correct approach IMO is:

- create API: check if is a core-type and do not scaffold the resource and populate accordingly the PROJECT file
- webhook API: check if we are trying to create a webhook for a core type, if yes then call create API for that to do the same above and allow the default scaffold occurs.

estroz:
/remove-lifecycle rotten
/ok-to-test

@camilamacedo86 follow-up thoughts on this?

pkg/plugins/golang/v2/api.go
```diff
@@ -133,6 +134,11 @@ func (p *createAPISubcommand) InjectResource(res *resource.Resource) error {
 
 	// In case we want to scaffold a resource API we need to do some checks
 	if p.options.DoAPI {
+		// Generating the API for a core resource is forbidden.
+		if strings.HasPrefix(p.resource.Path, "k8s.io") {
+			return errors.New("scaffolding API for core resources is not supported")
```
camilamacedo86, line 139: Why that is forbidden?

pkg/plugins/golang/v3/api.go
camilamacedo86: I am not sure. You added `return errors.New("scaffolding API for core resources is not supported")`

pkg/plugins/golang/v2/webhook.go
```diff
@@ -101,7 +101,7 @@ func (p *createWebhookSubcommand) InjectResource(res *resource.Resource) error {
 			return fmt.Errorf("%s create webhook requires a previously created API ", p.commandName)
 		}
 	} else {
-		if r, err := p.config.GetResource(p.resource.GVK); err != nil {
+		if r, err := p.config.GetResource(p.resource.GVK); err != nil || !r.HasAPI() {
```
camilamacedo86, line 104: IHMO we need here to improve this check. 

If the error be faced here because the resource was not found we need to check if the resource is a core type and it is scaffold in the project file. If yes, then fine we should not return an error and allow scaffold the webhooks.

camilamacedo86 (changes requested):
And sorry for the delay in the response. 

This PR requires some small changes. Could you please check my comments?

pkg/model/resource/resource.go
camilamacedo86: We cannot change core to external.
1) it is a breaking change
2) Core type means types that are defined in the K8s API and External means we scaffold an API that is defined in another project. For example, I have an Operator A which defines APIs that will also be used by my Operator B.

Issue #1785, opened 2020-11-04
Update kube-rbac-proxy to v0.8.0 release 


hectorj2f:
Per brancz/kube-rbac-proxy#99 (comment) has been pushed to quay.io/brancz/kube-rbac-proxy:v0.8.0. What is the process to update the gcr.io/kubebuilder/kube-rbac-proxy? Is this hand tagged and pushed by a maintainer?

Could you push it or retag it ? Thanks.

/kind bug

camilamacedo86:
Hi @hectorj2f, 

It has been done manually. @DirectXMan12 @vincepri @droot could you please give a hand with this one?

hectorj2f:
Thanks @camilamacedo86.

hectorj2f:
@DirectXMan12 @vincepri @droot Is there any way I could help with the release process/tagging for this new kube-rbac-proxy release 🙏🏻  ?

camilamacedo86:
Hi @hectorj2f, 

In the `Kubebuilder, Controller Runtime, and Controller Tools` meeting latest week we spoke about it. The current images built and pushed manually to gcr.io/kubebuilder/kube-rbac-proxy . And then, the author of https://github.com/brancz/kube-rbac-proxy has the intention to donate the project to k8s org, however, until it be done we need to automate it from the repo  https://github.com/brancz/kube-rbac-proxy. 

Currently has a little people with the requires access to push the images, so @DirectXMan12 would like to address it via automation instead of we still manually doing this work. @Pete Savage, could you give a hand to us to achieve it?

camilamacedo86:
We need to the same that is done in https://github.com/brancz/kube-rbac-proxy/blob/master/.github/workflows/build.yml#L35-L42 to push the image to quay, however, to push the image to gcr.io/kubebuilder/kube-rbac-proxy

camilamacedo86:
Just to register. We need to use the latest version of `kube-rbac-proxy` because the previous one is not rotless. More info: https://github.com/kubernetes-sigs/kubebuilder/issues/1637

hectorj2f:
@camilamacedo86 is there any update here ?

camilamacedo86:
Hi @hectorj2f,

Unfortunately, I didn't have time to continue to see it.  Would you like to help us with this one?

hectorj2f:
@camilamacedo86 I'd check what I can do ;), to speed things and get this done asap.

camilamacedo86:
[...]

PR #2158, opened 2021-04-27
docker manifests need to be annotated

DirectXMan12:
Previously, we weren't actually annotating the images in the manifest,
so we ended up with all the images set as linux-amd64 instead of their
respective bits.

This should fix that, annotating with arch before pushing, as per the
docker docs.

CC #1785


Adirio (approved):
LGTM, but would prefer an extra pair of eyes.

Issue #2159, opened 2021-04-28
Extend files whitelist for init cmd v3


w3st3ry:
### What do you want to happen?

This issue is related to [this one from `operator-sdk`](https://github.com/operator-framework/operator-sdk/issues/4796).

The allowed/whitelisted files for init cmd (in v3) are quite restricted, I would like to extend the list by whitelisting `.md` files extension + capitalized files name (like OWNERS, SECURITY, WORKSPACE...).

A PR will follow.

camilamacedo86:
In the bug triage meeting, we discussed it and seems that would be fine. However, that could not allow has the PROJEC file already.

PR #2160, opened 2021-04-28
✨ Extend files whitelist for init cmd

w3st3ry:
Related to #2159 

Signed-off-by: Valentin Pichard <github@w3st.fr>


pkg/plugins/golang/v3/init.go
```diff
@@ -171,12 +172,23 @@ func checkDir() error {
 			if strings.HasPrefix(info.Name(), ".") {
 				return nil
 			}
+			// Allow files ending with '.md' extension
+			if strings.HasSuffix(info.Name(), ".md") && !info.IsDir() {
+				return nil
+			}
+			// Allow capitalized files (like LICENSE, CODEOWNERS..)
```
Adirio, line 179: I'm not sure if we should allow capitalized files. That prevents us from using any other capitalized file in the future as the user could have already have one and it would break something that previously worked.

pkg/plugins/golang/v3/init.go
```diff
@@ -171,12 +172,23 @@ func checkDir() error {
...
+			for index, l := range info.Name() {
+				if !unicode.IsUpper(l) || info.Name() == "PROJECT" {
```
Adirio, line 181: Any way, checking the name against `PROJECT` should not be done inside of the loop.

pkg/plugins/golang/v3/init.go
```diff
@@ -171,12 +172,23 @@ func checkDir() error {
...
+					break
+				}
+				if index == len(info.Name())-1 {
+					return nil
+				}
```
Adirio, line 186: I feel like it is more natural to have a flag set before the loop that gets changed if any non-uppercase is found and returning outside of the loop if the flag has changed than to check this condition every loop.

Adirio (dismissed):

pkg/plugins/golang/v3/init.go
w3st3ry: As a personal feedback, I get stuck with multiple kubebuilder init because of this restriction.
They're more and more automations that create repo with this kind of files naming. 
Btw I understand your point too, as you want then.

pkg/plugins/golang/v3/init.go
w3st3ry: Fixed!

pkg/plugins/golang/v3/init.go
w3st3ry: Fixed!

pkg/plugins/golang/v3/init.go
```diff
@@ -171,12 +172,25 @@ func checkDir() error {
...
+			}
+			// Allow capitalized files except PROJECT
+			isCapitalized := true
+			for _, l := range info.Name() {
+				if !unicode.IsUpper(l) {
+					isCapitalized = false
+					break
+				}
+			}
+			if isCapitalized && info.Name() != "PROJECT" {
```
camilamacedo86, line 187: Will not that allow init a project when the dir has the PROJECT file already? 
Then, it means that the tool cannot ensure that the source of true is consistent.

Adirio: I think this case would break earlier, mind to check?

pkg/plugins/golang/v3/init.go
camilamacedo86: (IMO) The problem here is that the PROJECT file is capitalized. This one we could not allow have already.

pkg/plugins/golang/v3/init.go
w3st3ry: @camilamacedo86 According to this comment on the one you put into the issue, should I only remain the `.md` files whitelist into this PR? And add some useful capitalized files into `allowedFiles` array?

You right, I haven't tested this case.
It break earlier here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/cmd_helpers.go#L225

Adirio: Yeah I was pretty sure it did.

pkg/plugins/golang/v3/init.go
camilamacedo86: seems solved

jmrodri (approved):
/lgtm

pkg/plugins/golang/v3/init.go
w3st3ry: @Adirio status still "requested changes", you need me to modify smthg before approval?

camilamacedo86 (commented):
/lgtm

PR #2166, opened 2021-04-29
:book: : Fix download redirects with functions

DirectXMan12:
This resolves an issue introduced in the redirect prep for v3 (#2062) that
broke our download redirects.

The problem is that v3 releases are `kubebuilder_${os}_${arch}`, whereas
v2 & v1 releases are `kubebuilder_${version}_${os}_${arch}.tar.gz`.

Since netlify can't handle wildcards that are part of a path component
(like `/2.:minorversion`) instead of a whole one (`:version`), we can't
select on major version directly in our redirects.

Instead, we introduce a function that handles this logic for us, and
then use netlify 200-pseudo-redirects to "mount the function on the
downloads part of the releases endpoint.

### DIFF
--- a/docs/book/functions/handle-version.js
+++ b/docs/book/functions/handle-version.js
@@ -1,0 +1,36 @@
+function notFound(info) {
+    return {
+        statusCode: 404,
+        headers: {'content-type': 'text/html'},
+        body: ("<h1>Not Found</h1>"+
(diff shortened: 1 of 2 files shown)


estroz (approved):
TIL about netlify supporting JS functions.

/approve
/lgtm

PR #2168, opened 2021-04-29
:book: Fix netlify functions configuration

DirectXMan12:
The local `netlify` dev server and documentation expect a top-level
configuration block named functions, but the remote netlify builder
expects a key under the build block instead.  Who knows why.


camilamacedo86 (commented):
/lgtm

PR #2173, opened 2021-05-02
:bug: Only enforce lower Go version

### LINKED ISSUE #2172: Cannot scaffold with Go 1.16

I can't scaffold a project if my Go version is 1.16 without using `--skip-go-version-check`. Using `--skip-go-version-check` results in a perfectly functional project.

Though I'm sure there are improvements in Go 1.16 that kubeduilder could take advantage of, given the Go 1 compatibility guarantee I'm a little puzzled as to why kubebuilder choses to enforce an upper boundary on supported Go versions.

Given Go 1.16 was released 2.5 months before the kubebuilder 3 release it also seems a bit strange the current version of Go isn't supported.

### Reproducing this issue

Have the `go` command on your path resolve to a Go version of 1.16

```
2021/05/02 16:00:09 failed to initialize project: unable to run pre-scaffold tasks of "base.go.kubebuilder.io/v3": go version 'go1.16.3' is incompatible because 'requires 1.13 <= version < 1.16'. You can skip this check using the --skip-go-version-check flag
```

### KubeBuilder (CLI) Version

3.0.0

### PROJECT version

_No response_

### Plugin versions

_No response_

### Other versions

_No response_

### Extra Labels

_No response_

daenney:
Go versions newer than 1.13 should be supported, including the current stable of 1.16. This brings the implementation inline with the method documentation, which states:

> // checkGoVersion should only ever check if the Go version >= 1.13, since the kubebuilder binary only cares
// that the go binary supports go modules which were stabilized in that version (i.e. in go 1.13) by default

Fixes #2172

### DIFF
--- a/pkg/plugins/golang/go_version.go
+++ b/pkg/plugins/golang/go_version.go
@@ -33,11 +33,6 @@ var (
 		major: 1,
 		minor: 13,
 	}
-	go116alpha1 = goVersion{
-		major:      1,
-		minor:      16,
-		prerelease: "alpha1",
-	}
 
 	goVerRegexp = regexp.MustCompile(goVerPattern)
 )
@@ -149,8 +144,8 @@ func checkGoVersion(verStr string) error {
 		return err
 	}
 
-	if version.compare(go113) < 0 || version.compare(go116alpha1) >= 0 {
-		return fmt.Errorf("requires 1.13 <= version < 1.16")
+	if version.compare(go113) < 0 {
+		return fmt.Errorf("requires Go >= 1.13")
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


daenney:
Given that the `go.mod` that gets generated sets `go 1.15`, the current lower boundary of 1.13 in this check doesn't feel correct either. It might be more reasonable to adopt a policy of "we support the supported versions of Go" and drop this version check entirely.

Adirio:
Sorry but this is not going to be accepted until several repos upstream support Go1.16. Including but not limited to repos like [kubernetes](https://github.com/kubernetes/kubernetes) or [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime).

daenney:
There is nothing broken about upstream repos with Go 1.16? That's what the Go 1 compatibility guarantee is for.

Adirio:
There are still thinks that break projects despite the guarantee. For example `go` tooling behavior around modules changed in Go 1.16.

PR #2176, opened 2021-05-06
:sparkles: add default conversion review versions

varshaprasad96:
The conversion webhook client config requires a list
of conversion review versions. This change scaffolds the
conversion webhook patch with default conversion review version
of v1.

### DIFF
--- a/pkg/cli/alpha/config-gen/cert-generation-filter.go
+++ b/pkg/cli/alpha/config-gen/cert-generation-filter.go
@@ -38,6 +38,7 @@ type CertFilter struct {
 }
 
 // Filter implements kio.Filter
+// TODO: when v1 CRDs are supported, scaffold conversion webhook versions.
 func (c CertFilter) Filter(input []*yaml.RNode) ([]*yaml.RNode, error) {
 
 	if c.Spec.Webhooks.CertificateSource.Type != "dev" {
--- a/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/crd/patches/enablewebhook_patch.go
+++ b/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/crd/patches/enablewebhook_patch.go
@@ -66,5 +66,7 @@ spec:
           namespace: system
           name: webhook-service
           path: /convert
+      conversionReviewVersions:
+      - {{ .Resource.API.CRDVersion }}
     {{- end }}
 `
--- a/testdata/project-v3-addon/config/crd/patches/webhook_in_admirals.yaml
+++ b/testdata/project-v3-addon/config/crd/patches/webhook_in_admirals.yaml
@@ -12,3 +12,5 @@ spec:
           namespace: system
           name: webhook-service
           path: /convert
+      conversionReviewVersions:
+      - v1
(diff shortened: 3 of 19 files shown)


varshaprasad96:
cc: @estroz

pkg/cli/alpha/config-gen/cert-generation-filter.go
```diff
@@ -107,6 +107,12 @@ func (c CertFilter) Filter(input []*yaml.RNode) ([]*yaml.RNode, error) {
 		if err != nil {
 			return nil, err
 		}
+		err = matches[i].PipeE(yaml.LookupCreate(
+			yaml.ScalarNode, "spec", "conversion", "webhookClientConfig", "conversionReviewVersions"),
```
estroz, line 111: This is not a scalar node, it's a list node

pkg/cli/alpha/config-gen/cert-generation-filter.go
```diff
@@ -107,6 +107,12 @@ func (c CertFilter) Filter(input []*yaml.RNode) ([]*yaml.RNode, error) {
...
+			yaml.FieldSetter{StringValue: "v1"})
```
estroz, line 112: This should be a list of strings, the only value being "v1".

pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/crd/patches/enablewebhook_patch.go
```diff
@@ -66,5 +66,6 @@ spec:
           namespace: system
           name: webhook-service
           path: /convert
+      conversionReviewVersions: v1
```
estroz, line 69: This is a list type not string. Also I think we can safely assume that CRD version matches conversion review version.
```suggestion
      conversionReviewVersions:
      - {{ .Resource.API.CRDVersion }}
```

pkg/cli/alpha/config-gen/cert-generation-filter.go
```diff
@@ -107,6 +107,13 @@ func (c CertFilter) Filter(input []*yaml.RNode) ([]*yaml.RNode, error) {
 		if err != nil {
 			return nil, err
 		}
+		// TODO: verify if this works, since `yaml.Append` requires a node to be created for clientConfig
```
varshaprasad96, line 110: have added a comment here. `SequenceNode` seems to be apt, but not sure if this works as required as there are no tests for this yet.

estroz: SequenceNode is the correct type. Does the list get added as expected when you test this locally?

pkg/cli/alpha/config-gen/cert-generation-filter.go
```diff
@@ -107,6 +107,13 @@ func (c CertFilter) Filter(input []*yaml.RNode) ([]*yaml.RNode, error) {
...
+		err = matches[i].PipeE(yaml.LookupCreate(
+			yaml.SequenceNode, "spec", "conversion", "webhookClientConfig", "conversionReviewVersions"),
+			yaml.FieldSetter{StringValue: "v1"})
+		if err != nil {
+			return nil, err
+		}
```
estroz, line 116: Thought about this a bit more. This filter seems to only handle v1beta1 CRDs, which have a default conversion review version list containing "v1beta1". Therefore you can just remove this change.

Issue #2067, opened 2021-03-08
[META] Go 1.16


Adirio:
This issue aims to tracks all the changes that need to/can be done once moving to go 1.16. This is obviously gated by dependencies.

- [ ] [cleanup] Go 1.16 sets `GO111MODULES` to `on` by default instead of `auto`. There are several remanents of this env variable that we should remove throughout the codebase.
- [ ] [enhancement] Go 1.16 includes a new `embed` package that allows to embed files into the binary using `//go:...`-style comments. This allows to split the files template from their logic offering way better support by IDEs (templates will no longer be a string, so they will be styled, auto-completed, and all the goodness of IDEs).

If there is any other feature enabled/required by go 1.16 that you will like to see in this list, please comment and I will update the list.

camilamacedo86:
We also need before that just to clarify, have the same bump in controller-runtime, declarative-kb and k8s-api before we are able to address that here.

kuritka:
The `go install` command can install go programs without affecting mod file dependencies. For example, executing `go install github.com/dummy/myapp@v0.1.2` is enough. Will you plan to update Makefile  ?
something like this :
```Makefile
# would work in GO1.16
generate: 
	## Download controller-gen locally if necessary.
	go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1
	## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
	controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
```





or you plan to support older GO versions (stay with `go get...` / [go-get-tool](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3-addon/Makefile#L92)), 
like this:
```Makefile
# without GO1.16
generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."

CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
controller-gen: ## Download controller-gen locally if necessary.
	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1)

# go-get-tool will 'go get' any package $2 and install it to $1.
PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
[...]

PR #2180, opened 2021-05-11
✨ Update go version in kubebuilder

sujil02:
Updates go version to 1.16 in kubebuilder. Reference: #2067 

Signed-off-by: Sujil02 <sujil.shah@gmail.com>


testdata/project-v2-addon/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.13 as builder
+FROM golang:1.16 as builder
```
jmrodri, line 2: This doesn't seem right. I don't think we should be changing the version for v2 because that was built using 1.13.

pkg/plugins/golang/go_version.go
```diff
@@ -33,10 +33,10 @@ var (
...
 		minor: 13,
 	}
-	go116alpha1 = goVersion{
-		major:      1,
-		minor:      16,
-		prerelease: "alpha1",
+	go116 = goVersion{
+		major: 1,
+		minor: 16,
+		patch: 5,
```
estroz, line 39: Make the variable name generic since we'll have to change this again in the future, and use 1.17 with strict `>` below.
```suggestion
	goVerMax = goVersion{
		major: 1,
		minor: 17,
```

pkg/plugins/golang/go_version.go
```diff
@@ -149,8 +149,8 @@ func checkGoVersion(verStr string) error {
 		return err
 	}
 
-	if version.compare(go113) < 0 || version.compare(go116alpha1) >= 0 {
-		return fmt.Errorf("requires 1.13 <= version < 1.16")
+	if version.compare(go113) < 0 || version.compare(go116) >= 0 {
+		return fmt.Errorf("requires 1.13 <= version < 1.16.5")
```
estroz, line 153:
```suggestion
	if version.compare(go113) < 0 || version.compare(goVerMax) > 0 {
		return fmt.Errorf("requires 1.13 <= version < 1.17")
```

pkg/plugins/golang/v2/scaffolds/internal/templates/dockerfile.go
```diff
@@ -39,7 +39,7 @@ func (f *Dockerfile) SetTemplateDefaults() error {
 }
 
 const dockerfileTemplate = `# Build the manager binary
-FROM golang:1.13 as builder
+FROM golang:1.16 as builder
```
estroz, line 42: Leave go/v2 as 1.13. This plugin is frozen.
```suggestion
FROM golang:1.13 as builder
```

pkg/plugins/golang/v2/scaffolds/internal/templates/dockerfile.go
```diff
@@ -55,7 +55,7 @@ COPY api/ api/
 COPY controllers/ controllers/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
```
estroz, line 58:
```suggestion
RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
```

varshaprasad96:
@sujil02 you may have to retitle the PR. 
You need to have one of these as the prefix of your PR title:

- Breaking change: ⚠ (`:warning:`)
- Non-breaking feature: ✨ (`:sparkles:`)
- Patch fix: 🐛 (`:bug:`)
- Docs: 📖 (`:book:`)
- Infra/Tests/Other: 🌱 (`:seedling:`)

This seems to be a `sparkles`

estroz (changes requested):

docs/book/src/multiversion-tutorial/testdata/project/Dockerfile
```diff
@@ -15,7 +15,7 @@ COPY api/ api/
 COPY controllers/ controllers/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
```
varshaprasad96, line 18: go 1.16 seems to have `GO111MODULE=on` by default. But I think it should be better to explicitly have this.

pkg/plugins/golang/go_version.go
sujil02: Fixed.

pkg/plugins/golang/v2/scaffolds/internal/templates/dockerfile.go
sujil02: Fixed.

sujil02:
I signed it

pkg/plugins/golang/go_version.go
```diff
@@ -33,10 +33,9 @@ var (
 		major: 1,
 		minor: 13,
 	}
-	go116alpha1 = goVersion{
-		major:      1,
-		minor:      16,
-		prerelease: "alpha1",
+	goVerMax = goVersion{
+		major: 1,
+		minor: 17,
```
Adirio, line 38: This would mean that any 1.17 pre-release and any 1.16.* versions would be accepted. We have been intentionally limiting new versions pretty strictly, I don't see any point made to switch this behavior to loosely accepting all the mentioned versions.

estroz: Ok so adding `alpha.0` to this go version would theoretically make comparison strictly < 1.17 right?

Restricting patch versions seems silly; why do you see this restriction as desirable behavior?

Adirio: I'm fine with the patch versions, just listed the whole implications of the change.

estroz:
```suggestion
	goVerMax = goVersion{
		major:      1,
		minor:      17,
		prerelease: "alpha1",
```

sujil02: Fixed.

PR #2186, opened 2021-05-20
:book: add default conversion review versions to kb book

varshaprasad96:
Update tutorials in book to scaffold default conversion review version.
follow up: #2176 


varshaprasad96:
Have created this in the master, and not in `book-v3` branch (not sure if that should be done after a release is made).

PR #2187, opened 2021-05-21
:book: update book-v3

estroz:
book-v3 update

/kind documentation


camilamacedo86 (commented):
Hi @estroz, 

We need to push from release-v3 into book-v3 instead of master to ensure that the changes in the docs are compatible with its latest release. 

 Could you please create a PR from release-v3? 

Indeed I think we need to do a new release 3.1.0 with the changes in master. Did we not bump an upper version of the controller-runtime after the 3.0.0 stable release?

/hold

estroz:
Is that true though? The release guide says merge from [`master` to `book-v3`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/RELEASE.md#release-process).

PR #2197, opened 2021-05-29
✨ Support argument based cross-compilation and multi-platform Dockerfile

### LINKED ISSUE #2175: Make it easy to build multi-platform images

Right now the Docker image generated by the scaffolder hardcodes `GOARCH=amd64` when building. This results in the image always containing an amd64 binary. This will probably also result in a broken image if one were to build this on a non-amd64 host. With Arm gaining momentum in the server space, it would be nice if easily building for a different architecture and building multi-platform images becomes possible with the same Dockerfile.

It's probably better if the Docker image doesn't hardcode the `GOARCH`. For folks that then want to build arm64 or s390 images, they can then use the same Dockerfile, and use [buildx](https://docs.docker.com/buildx/working-with-buildx/) to properly build multiarchitecture images.

### Extra Labels

_No response_

aayushrangwala:
Support the multi-platform cross-compilation Dockerfile in scaffolding.

- Added the --platform flag as input in the docker build command in scaffold Makefile with default value as local

- Changed Dockerfile to use BUILDPLATFORM, TARGETPLAFORM, TARGETOS and TARGETARCH in-built env var to build arg based binary

Signed-off-by: Ayush Rangwala <arangwala@vmware.com>

Motivation: Since the Kubebuilder scaffolding always generate `GOOS=linux` and `GOARCH=amd64` based Dockerfile, which results in broken image for other environments.

Fixes #2175

### DIFF
--- a/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
@@ -76,6 +76,10 @@ endif
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+# PLATFORM is the argument to the make 'build' target command.
+# It defines the target platform for which the cross-compilation build is supposed to be.
+PLATFORM ?= local
+
 all: build
 
 ##@ General
@@ -123,7 +127,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
 docker-build: test ## Build docker image with the manager.
-	docker build -t ${IMG} .
+	docker build --platform ${PLATFORM} -t ${IMG} .
 
 docker-push: ## Push docker image with the manager.
 	docker push ${IMG}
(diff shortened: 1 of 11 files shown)


camilamacedo86 (commented):
/ok-to-test

it seems fine for me. @estroz @jmrodri wdyt

estroz (changes requested):
`--platform` is a [docker buildx](https://docs.docker.com/buildx/working-with-buildx/#build-multi-platform-images) feature and will not work with vanilla `docker build` (unless a new docker release added support that I do not know about).

anthr76:
Hi @aayushrangwala will this support a single fat manifest?

camilamacedo86:
HI @aayushrangwala, 

I hope that is fine we close this PR based on the comment:  https://github.com/kubernetes-sigs/kubebuilder/pull/2197#pullrequestreview-673358884

However, please feel free to contact us and reach out if you think that should be re-opened.

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
```diff
@@ -76,6 +76,11 @@ endif
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+DOCKER_BUILDKIT=1
```
camilamacedo86, line 79: >  Buildx builds using the BuildKit engine and does not require DOCKER_BUILDKIT=1

Not required according to the docs; https://docs.docker.com/build/buildx/

Issue #1993, opened 2021-02-05
Wrong CRD plural generated


benjaminhuo:
- What broke (please include exact error messages if you can)
Wrong CRD plural generated, the following CRD is generated when running `kubebuilder create api --group openfunction.io --version v1alpha1 --kind FunctionBuild` or `make manifests`

![image](https://user-images.githubusercontent.com/18525465/107020772-3dabc500-67de-11eb-982b-351b2d8ece02.png)

- What did you expect to happen?  What do you think went wrong? 
`plural: functionbuilds` should be generated instead of  `plural: functionbuildren`

- What versions of software are you using?  Specifically, the following are often useful:

* go version 
   v1.14
* kubebuilder version (`kubebuilder version`) and scaffolding version (check your `PROJECT` file)
  v2.3.1
* controller-runtime version (check your `go.mod` file)
  v0.6.4
* controller-tools version
  controller-gen@v0.2.5

/kind bug

Adirio:
In `kubebuilder` v2.3.1 plurals can not be manually provided and use `flect.Pluralize(strings.ToLower(<KIND>))` to obtain the plural form. v3.0.0 which will be released shortly offers a `-plural` flag that allows users to specify other plural forms.

benjaminhuo:
Got it, thanks for the reply @Adirio 
Close this issue for now.

Ben

PR #1933, opened 2021-01-08
🌱 Tests cleanup

Adirio:
# Description

The following list defines all the checks and tests that we currently have, with some consideration on prerequisites and under which conditions are being executed (go version, devide OS/arch, and kubernetes version).

### Lint
- **Description:** verifies the code style.
- **Prerequisites:** none.
- **Make:** `make lint` to report the errors or `make lint-fix` to solve them.
- **CI:** yes (reporting, GH Actions).
- **Parameters:** single execution.

### Unit tests
- **Description:** unit tests defined in `*_test.go` files run by `go test ...` that test the different parts of the library isolated.
- **Prerequisites:** lint. *1
- **Make:** `make test-unit`.
- **CI:** yes (GH Actions & Prow). *2
- **Parameters:** multiple executions based on device OS/arch as right now. *3

### Coverage *4
- **Description:** obtains coverage info of unit tests and upload it to coveralls.
- **Prerequisites:** unit tests.
- **Make:** `make test-coverage` obtains the coverage info.
- **CI:** yes (GH Actions).
- **Parameters:** single execution.

### Integration tests *5
- **Description:** runs some basic kubebuilder commands to test the CLI.
- **Prerequisites:** unit tests.
- **File:** `test/integration.sh`.
- **Make:** `test-integration`.
- **CI:** yes (Prow).
- **Parameters:** single execution.

### Testdata check
- **Description:** verifies that the code under the `testdata` directory is up to date.
- **Prerequisites:** lint. *6
- **File:** `test/testdata/check.sh`.
- **Make:** `make check-testdata`.
- **CI:** yes (GH Actions).
- **Parameters:** single execution. *7

### Testdata tests
- **Description:** executes the unit tests of the scaffolded projects in the testdata directory.
- **Prerequisites:** integration tests. *8
- **File:** `test/testdata/test.sh`.
- **Make:** `make test-testdata`.
- **CI:** yes (Prow).
- **Parameters:** single execution. *7

### E2E tests
- **Description:** end-to-end tests.
- **Prerequisites:** none. *9
- **File:** `test/e2e/local.sh` or `test/e2e/ci.sh`.
- **Make:** `make test-e2e-local` or `make test-e2e-ci`.
- **CI:** yes (Prow).
- **Parameters:** multiple executions for several kubernetes versions.

## Notes:
1. Linting is not really a prerequisite for unit tests in the most strict sense, but as it is pretty fast, I don't think that it matters that much.
2. We should remove the overlap between GH Actions and Prow.
[...]

PR #2200, opened 2021-06-02
🐛 Update kubernetes-sigs/controller-tools to v0.5.0

Birdrock:
## Description
kubebuilder has been updated to use github.com/gobuffalo/flect v0.2.2; the v3/scaffolds/init.go specifies kubernetes-sigs/controller-tools v0.4.1, which has github.com/gobuffalo/flect v0.2.0 as a dependency. kubernetes-sigs/controller-tools v0.5.0 has github.com/gobuffalo/flect v0.2.2 as a dependency.

## Motivation
The two versions of github.com/gobuffalo/flect have slightly difference casing mechanics, so pluralizing doesn't exactly match. #1993 points this out, but was closed. While `kubebuilder create api` pluralizes properly, `make manifests` creates different pluralization in some cases (e.g.: Builder -> Buildren). Reconciling these two versions coordinates the casing mechanics.

Regardless of the underlying problem, it seems like a good idea to keep these dependencies synchronized. This could be considered a fix to part of #1933.

### DIFF
--- a/docs/book/src/migration/v2vsv3.md
+++ b/docs/book/src/migration/v2vsv3.md
@@ -41,7 +41,7 @@ Projects scaffolded with Kubebuilder v3 will use the `go.kubebuilder.io/v3` plug
   * A new option to create the projects using ComponentConfig is introduced. For more info see its [enhancement proposal][enhancement proposal] and the [Component config tutorial][component-config-tutorial]
   * Manager manifests now use `SecurityContext` to address security concerns. More info: [#1637][issue-1637] 
 - Misc:
-  * Support for [controller-tools][controller-tools] `v0.4.1` (for `go/v2` it is `v0.3.0` and previously it was `v0.2.5`) 
+  * Support for [controller-tools][controller-tools] `v0.5.0` (for `go/v2` it is `v0.3.0` and previously it was `v0.2.5`) 
   * Support for [controller-runtime][controller-runtime] `v0.8.3` (for `go/v2` it is `v0.6.4` and previously it was `v0.5.0`)
   * Support for [kustomize][kustomize] `v3.8.7` (for `go/v2` it is `v3.5.4` and previously it was `v3.1.0`)
   * Required Envtest binaries are automatically downloaded
--- a/docs/book/install-and-build.sh
+++ b/docs/book/install-and-build.sh
@@ -67,7 +67,7 @@ ${cmd} /tmp/mdbook.${ext}
 chmod +x /tmp/mdbook
 
 echo "grabbing the latest released controller-gen"
-go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1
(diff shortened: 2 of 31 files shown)


Birdrock:
I wasn't sure if an API diff was considered a breaking change since it is minor bump. It does change the output of `make manifests`, so that could be considered breaking.

Additionally, hopefully the CLA will sort itself out.

pwittrock:
Looks good to me.  Seems like the only changes to test data are line wrapping whitespace for the descriptions and the annotations with the controller-tools version.

docs/book/src/migration/v2vsv3.md
```diff
@@ -41,7 +41,7 @@ Projects scaffolded with Kubebuilder v3 will use the `go.kubebuilder.io/v3` plug
   * A new option to create the projects using ComponentConfig is introduced. For more info see its [enhancement proposal][enhancement proposal] and the [Component config tutorial][component-config-tutorial]
   * Manager manifests now use `SecurityContext` to address security concerns. More info: [#1637][issue-1637] 
 - Misc:
-  * Support for [controller-tools][controller-tools] `v0.4.1` (for `go/v2` it is `v0.3.0` and previously it was `v0.2.5`) 
+  * Support for [controller-tools][controller-tools] `v0.5.0` (for `go/v2` it is `v0.3.0` and previously it was `v0.2.5`)
```
varshaprasad96, line 44: is this version bump going to be back ported ? 
If not, can we directly bump controller-tolls to 0.6.0 and controller-runtime to 0.9.0. That would bring in k8s 1.21 support for projects scaffolded with v3 plugin.

Birdrock: I can definitely add bumps - controller-tools v0.6.0 had yet to be published at the time of PR. Is the preference for including it in this PR or should I start a separate one to include those bumps?

Birdrock:
I believe the white space differences may be a result of `go-yaml 2.3.0` defaulting line length to infinity recently. https://github.com/go-yaml/yaml/pull/571

docs/book/src/migration/v2vsv3.md
estroz: Two PRs is fine with me.

Birdrock: I'll create a second PR today at some point today to cover this (assuming following caveat is resolved). I did an initial pass updating controller-tools to v0.6.0 and controller-runtime to v0.9.0. The controller-runtime update seemed to cause some issues, but I haven't had a chance to investigate yet.

estroz:
Merging this myself since I cannot override a GH action.

docs/book/src/migration/v2vsv3.md
estroz: #2208 will handle this.

Birdrock: @estroz sg!

pwittrock:
@estroz its infuriating isn't it?

PR #2201, opened 2021-06-02
:book: controllerutil.AddFinalizer already contains code to avoid duplicates

dionysius:
Removed unneeded code and with that simplified it a bit. controllerutil.AddFinalizer already contains code to avoid duplicates

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
+++ b/docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
@@ -66,11 +66,9 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		// The object is not being deleted, so if it does not have our finalizer,
 		// then lets add the finalizer and update the object. This is equivalent
 		// registering our finalizer.
-		if !containsString(cronJob.GetFinalizers(), myFinalizerName) {
-			controllerutil.AddFinalizer(cronJob, myFinalizerName)
-			if err := r.Update(ctx, cronJob); err != nil {
-				return ctrl.Result{}, err
-			}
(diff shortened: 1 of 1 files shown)


docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
```diff
@@ -66,11 +66,9 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		// The object is not being deleted, so if it does not have our finalizer,
 		// then lets add the finalizer and update the object. This is equivalent
 		// registering our finalizer.
-		if !containsString(cronJob.GetFinalizers(), myFinalizerName) {
-			controllerutil.AddFinalizer(cronJob, myFinalizerName)
-			if err := r.Update(ctx, cronJob); err != nil {
-				return ctrl.Result{}, err
-			}
```
camilamacedo86, line 73: But then it will not add the finalizer always and without check if that exists or not?

dionysius: Well, https://github.com/kubernetes-sigs/controller-runtime/blob/fbf50b04fe17b0f091430d22557245d2aedabfd4/pkg/controller/controllerutil/controllerutil.go#L124 does all the heavy lifting to check it exists only once

But what I currently see is a secondary effect of the previous `if`. With my PR, it always tries to `r.Update`, while before it would only if it actually got added.

camilamacedo86: Yes, I think we should keep the example as it is. IHMO, that makes its purpose more intuitive. 
So, I hope that you do not mind but I am closing this one. Please, feel free to ping if you think that it should be re-opened.

PR #2202, opened 2021-06-04
✨ Update config-gen command

pwittrock:
✨

- Switch to embed for the config-gen command
- Update the controller-tools version


pwittrock:
@KnVerey I've update the config-gen command to use the framework changes you've made

pwittrock:
Seems to be an issue with the go 1.16 tools requiring deps to be specified.

PR #2208, opened 2021-06-09
:sparkles: bump k8s dependencies to 1.21

varshaprasad96:
Bump k8s dependencies in kubebuilder binary to 1.21


pkg/cli/alpha/config-gen/testdata/componentconfig/expected.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
     control-plane: controller-manager
   name: simple-system
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
varshaprasad96, line 8: this bumps the crds in configgen testdata to v1 from v1beta1
cc: @pwittrock

estroz:
@varshaprasad96 go/v3 projects need to use controller-runtime v0.9.0 too.

varshaprasad96:
@estroz since there is a [pr](https://github.com/kubernetes-sigs/kubebuilder/pull/2200) open on bumping controller-tools version, wasn't sure if it would be done there. Can I do that in a follow up PR?

Birdrock:
@varshaprasad96 I can update my PR (#2200) to keep up to date with versions, if that helps.

varshaprasad96:
@Birdrock sure, anything works. If you would like to have 2 separate PRs, then we can have k8s 1.21 bump separately. Since #2200 is approved, we can merge that first. i'll rebase this one.

go.mod
```diff
@@ -6,15 +6,15 @@ require (
 	github.com/cloudflare/cfssl v1.5.0 // for `kubebuilder alpha config-gen`
 	github.com/gobuffalo/flect v0.2.2
 	github.com/joelanford/go-apidiff v0.1.0
-	github.com/onsi/ginkgo v1.15.0
-	github.com/onsi/gomega v1.10.5
+	github.com/onsi/ginkgo v1.16.4
+	github.com/onsi/gomega v1.13.0
```
camilamacedo86, line 10: Do we need to update that? See : 

github.com/onsi/ginkgo
  Incompatible changes:
  - GINKGO_VERSION: value changed from "1.15.0" to "1.16.4"

That breaks the api-diff

camilamacedo86: However, in this case, I was also in favor of. We might need to check the api-diff as well. c/c @estroz

estroz: I don't think this api-diff conflict matters in practice.

estroz:
@varshaprasad96 I'd rather all k8s deps be updated in the same PR.

varshaprasad96:
yep, I'll bump the dependencies in plugin here.

varshaprasad96:
/hold
this is blocked by https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/160

varshaprasad96:
@estroz @camilamacedo86 - the PR is ready for review

test/testdata/generate.sh
```diff
@@ -57,7 +57,7 @@ function scaffold_test_project {
     if [ $project == "project-v2" ]; then
       $kb create api --plugins="go/v2,declarative" --group crew --version v1 --kind FirstMate --controller=true --resource=true --make=false
```
estroz, line 58: Is there a reason `declarative` is here too?
```suggestion
      $kb create api --plugins="go/v2" --group crew --version v1 --kind FirstMate --controller=true --resource=true --make=false
```

varshaprasad96: Sry, that was a mistake! I shouldn't have removed from `go/v3` either. Added it back and regenerated the samples.

estroz:
/override APIDiff
/approve
/lgtm

I'll likely have to force merge this since I don't think overrides will work on action runs.

PR #2217, opened 2021-06-14
📖  Additional Metrics Docs

VenkatRamaraju:
**A description of the change:**
Added some steps to the "Publishing Additional Metrics" section of the Metrics documentation.

**The motivation of the change:**
To clarify some of the steps in the existing docs on how to publish and view custom metrics.


docs/book/src/reference/metrics.md
```diff
@@ -92,5 +94,26 @@ func init() {
...
+
+	// Add metrics
+	goobers.Inc()
+	gooberFailures.Inc()
+
+	return ctrl.Result{}, nil
+}
+```
+
+Those metrics will be available for prometheus or
```
varshaprasad96, line 114: @VenkatRamaraju Looks good! Can you also add a note that if someone wants to publish metrics on Prometheus UI, they would have to configure their prometheus instance to select the service monitor based on labels.

docs/book/src/reference/metrics.md
```diff
@@ -92,5 +94,28 @@ func init() {
...
+reconcile loop.
+
+For example:
+```
+func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	...
+	...
+	...
+
+	// Add metrics
```
varshaprasad96, line 106:
```suggestion
```

Technically, the custom metrics can be incremented/modified anywhere in the operator. So I'd say lets remove this from here and just add a note that these custom metrics can be evaluated anywhere in the operator code.

Another thing which could be added to doc, would be the missing roles/role bindings which was causing rbac error. That is explained briefly in the previous paragraph, an example of the cluster role binding which is to be added can be included.

varshaprasad96 (changes requested):

docs/book/src/reference/metrics.md
```diff
@@ -92,5 +111,31 @@ func init() {
...
+	gooberFailures.Inc()
+	
+	...
+	...
+	
+	return ctrl.Result{}, nil
+}
+```
+
+**Note**: In order to publish metrics and view them on the Prometheus UI, the Prometheus instance would have to be configured to select the Service Monitor instance based on its labels.
```
camilamacedo86, line 134: The docs have a specific tag for notes. Could you please check in other places and use it?

camilamacedo86 (commented):
Just a nit otherwise . Thank you for the contribution 🥇

docs/book/src/reference/metrics.md
```diff
@@ -21,6 +21,23 @@ You can run the following kubectl command to create it. If using kubebuilder
 kubectl create clusterrolebinding metrics --clusterrole=<project-prefix>-metrics-reader --serviceaccount=<namespace>:<service-account-name>
 ```
```
varshaprasad96, line 23: This is some slight modification to Line 17
You can either run the following command, or apply the sample yaml file to create `clusterRoleBinding`. If using kubebuilder
`<project-prefix>` is the `namePrefix` field in `config/default/kustomization.yaml`.

docs/book/src/reference/metrics.md
```diff
@@ -21,6 +21,23 @@ You can run the following kubectl command to create it. If using kubebuilder
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: prometheus-k8s-role
+subjects:
+  - kind: ServiceAccount
+    name: prometheus-k8s
+    namespace: monitoring
+```
+
```
varshaprasad96, line 40:
```suggestion
You can also apply the following `ClusterRoleBinding`:
```yaml
kind: ClusterRoleBinding
metadata:
  name: prometheus-k8s-rolebinding
  namespace: <operator-namespace>
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: prometheus-k8s-role
subjects:
  - kind: ServiceAccount
    name: <prometheus-service-account>
    namespace: <prometheus-service-account-namespace>
```
```

Comment (varshaprasad96) [reply docs/book/src/reference/metrics.md:40]: (continuation)
The role referenced here should provide the necessary permissions to allow prometheus scrape metrics from operator pods.

File: docs/book/src/reference/metrics.md [hunk 6]
```diff
@@ -92,5 +111,31 @@ func init() {
...
+	...
+
+	goobers.Inc()
+	gooberFailures.Inc()
+	
+	...
+	...
+	
+	return ctrl.Result{}, nil
+}
```
varshaprasad96, line 131:
```suggestion
```

Actually this example is not necessary. It may look confusing to increment `goobers` and `gooberFailers` in the same reconcile loop.

varshaprasad96 (commented):
a few suggestions

docs/book/src/reference/metrics.md
```diff
@@ -92,5 +114,13 @@ func init() {
 ```
 
 You may then record metrics to those collectors from any part of your
-reconcile loop, and those metrics will be available for prometheus or
+reconcile loop. These metrics can be evaluated from anywhere in the operator code.
+
+**Note**: In order to publish metrics and view them on the Prometheus UI, the Prometheus instance would have to be configured to select the Service Monitor instance based on its labels.
```
camilamacedo86, line 119:
```suggestion

<aside class="note">
<h1>Enabling metrics in Prometheus UI</h1>

In order to publish metrics and view them on the Prometheus UI, the Prometheus instance would have to be configured to select the Service Monitor instance based on its labels. 

</aside>
```

camilamacedo86 (changes requested):
Just a nit in https://github.com/kubernetes-sigs/kubebuilder/pull/2217/files#r653440804
And please, rebase your commits for we are able to get it merged

varshaprasad96 (approved):
just a nit, rest looks good 
/lgtm

varshaprasad96:
The PR looks good! 
@camilamacedo86 @estroz if the changes look fine, can we merge this?

docs/book/src/reference/metrics.md
```diff
@@ -60,7 +80,7 @@ If you wish to publish additional metrics from your controllers, this
 can be easily achieved by using the global registry from
 `controller-runtime/pkg/metrics`.
 
-One way to achieve this is to declare your collectors as global variables and then register them using `init()`.
+One way to achieve this is to declare your collectors as global variables and then register them using `init()` in a custom controller class.
```
estroz, line 83:
```suggestion
One way to achieve this is to declare your collectors as global variables and then register them using `init()` in the controller's package.
```

estroz (approved):
/approve
/lgtm

Issue #1723, opened 2020-10-13
We should have an FAQ / common issues book section


DirectXMan12:
There's all sorts of weird k8s API server issues (and a few KB-specific ones :wink:) that folks can run into, and we should document them somewhere.

Off the top of my head:

- Weird issues when a v1beta1 CRD is already installed in the cluster (invisibly set `preserveUnknownFields`, refusal to chance certain fields, etc)

/kind docs

camilamacedo86:
Following a few examples to help who will work on in this task:
- controller-runtime; https://github.com/kubernetes-sigs/controller-runtime/blob/master/FAQ.md
- sdk: https://sdk.operatorframework.io/docs/faqs/faqs/

camilamacedo86:
We can explain that and link https://book.kubebuilder.io/cronjob-tutorial/gvks.html.

camilamacedo86:
We can close this one because we have a FAQ section now, see: https://github.com/kubernetes-sigs/kubebuilder/pull/3044

Regards the need:

> Weird issues when a v1beta1 CRD is already installed in the cluster (invisibly set preserveUnknownFields, refusal to chance certain fields, etc)

It seems outdated since v1beta1 CRD is no longer supported. 

So, it seems fine we close this one.

Issue #1318, opened 2020-01-16
Explain more about the domain field is in documentation


jroper:
One of the first things you have to do when getting started is to select a domain name. For a small project in a small organisation, this might be straight forward, but let's say you work for a large multinational that has 1000s of development teams, you don't just go using your companies domain name anywhere without first asking yourself "is this going to cause a conflict if we ever need to integrate with the operators from other teams in the company? Should I include my organisation unit name in the hostname?"

The most important information that's needed I think is whether the domain name needs to be unique - ie, if two different kubebuilder projects use the same domain name, could that cause a problem, in particular if they ever need to coexist with each other in the same system? The getting started guide should mention this, and the book should probably go into more detail about what the domain name is used for.

droot:
This fits use-case for a FAQ entry and `getting-started` can link to that FAQ entry.

camilamacedo86:
See that we have an issue to track a FAQ: https://github.com/kubernetes-sigs/kubebuilder/issues/1723

PR #2227, opened 2021-06-18
📖 Add FAQ documentation section and propose how to about webhook and Kubernetes client

JulienBreux:
Hi awesome community! 👋🏻 

I encountered an issue when I wanted to employ the best practice for using the Kubernetes client in a webhook.

Today, I wanted to suggest to the community to create a new FAQ section in the documentation with this interesting tips.
(We use such a section at [Minikube](https://minikube.sigs.k8s.io/docs/faq/))

What do you think?
Is it a good idea?
Is it in Kubebuilder's philosophy?

Happy to read more.
Have a very nice day.

### DIFF
--- a/docs/book/src/faq/faq.md
+++ b/docs/book/src/faq/faq.md
@@ -1,0 +1,43 @@
+# FAQ
+
+Frequently Ask Question
+
+## How to use Kubernetes client in webhook?
+
+You can declare the Kubernetes client in webhook from the controller manager easily.
+
+<details>
+  <summary>Read more</summary>
+
+To begin, you need to import the client and declare the variable:
+
+```go
+import (
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	C client.Client
+)
+```
+
+Next, it is very simple to fill the previous client variable from the controller manager.
+
+```go
+func (r *MyAwesomeCRD) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	C = mgr.GetClient()
+
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+```
+
+Finally, you can use the client in your validator for example:
+
+```go
+func (r *MyAwesomeCRD) ValidateDelete() error {
+	return C.Get(// ... //)
+}
+```
+</details>
--- a/docs/book/src/SUMMARY.md
+++ b/docs/book/src/SUMMARY.md
@@ -103,8 +103,13 @@
   - [Extending the CLI](./plugins/extending-cli.md)
   - [Creating your own plugins](./plugins/creating-plugins.md)
 
+---
+
+- [FAQ][faq]
+
 ---  
 [Appendix: The TODO Landing Page](./TODO.md)
 
 
 [plugins]: ./plugins/plugins.md
+[faq]: ./faq/faq.md


JulienBreux:
/assign @camilamacedo86

😊

JulienBreux:
Hi @camilamacedo86!

What do you think about this proposition?

Thanks in advance.

camilamacedo86 (commented):
I am ok with that. @varshaprasad96 you are who most work with webhooks then, could you please let us know wdyt? Also, do you think that it would be something beneficial to be scaffold by default? I mean, could we not do that as we do on the controller for example?

docs/book/src/faq/faq.md
```diff
@@ -0,0 +1,43 @@
...
+
+var (
+	C client.Client
+)
+```
+
+Next, it is very simple to fill the previous client variable from the controller manager.
+
+```go
+func (r *MyAwesomeCRD) SetupWebhookWithManager(mgr ctrl.Manager) error {
```
varshaprasad96, line 27: @JulienBreux this is scaffolded by kubebuilder by default, with the `create webhook` command. I don't think this needs to be done manually. 

According to the docs, 
```
`kubebuilder create webhook --group batch --version v1 --kind CronJob --defaulting --programmatic-validation`
This will scaffold the webhook functions and register your webhook with the manager in your main.go for you.
```

JulienBreux: Thanks, yes, I known.

This FAQ entry was just here to demonstrate how to use client in this generated webhook file.
Or "after" the generation.

camilamacedo86:
Hi @varshaprasad96 @JulienBreux, 

I was looking for we scaffold of the code into the webhooks and not a doc. Why not instead of doc that we do not scaffold this code by default? See that for all controllers we scaffold the client: 

https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/controllers/admiral_controller.go#L32

why we do not do the same for webhooks? What would be the con on this case?

docs/book/src/faq/faq.md
```diff
@@ -0,0 +1,43 @@
...
+
+To begin, you need to import the client and declare the variable:
+
+```go
+import (
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	C client.Client
```
camilamacedo86, line 20:
```suggestion
	client client.Client
```

camilamacedo86: I do not think that it will be used outside right?

camilamacedo86, line 20: I do not think that it should be exposed. You are using only here, right?

camilamacedo86 (changes requested):
Hi @JulienBreux, 

@varshaprasad96 and I spoke a little about that. I'd like to share that we really want a FAQ section, but we would like to address problems like https://github.com/kubernetes-sigs/kubebuilder/issues/1723, https://github.com/kubernetes-sigs/kubebuilder/issues/1318

However, if you were motivated to do this PR seems that was painful for you to know how to use the client in the webhooks. So, I think this info might help others. 

WDYT about we change the scaffold https://github.com/kubernetes-sigs/kubebuilder/blob/1831527c09c7ad54bd1782f0e01550b425ef9ae1/testdata/project-v3/api/v1/admiral_webhook.go#L44 to something like 

> // TODO(user): fill in your defaulting logic. You can use the client by `mgr.GetClient()`

docs/book/src/faq/faq.md
JulienBreux: Yes, you're right!

JulienBreux:
Hi @camilamacedo86,

Thanks a lot for your time and your reply.

I've just a question, what's the convention?

```go
var (
	ctrlClient client.Client
)
```

Instead of

```go
var (
	C client.Client
)
```

What do you think?

camilamacedo86:
Hi @JulienBreux, 

I think our goal is only to provide info right `// TODO(user): fill in your defaulting logic. You can use the client by mgr.GetClient()`. I do not think that we should impose the name of the var and I am not aware of any convention for the project regarding this.

estroz (changes requested):
>I encountered an issue when I wanted to employ the best practice for using the Kubernetes client in a webhook.

@JulienBreux is this really a best practice? Can you point me to an upstream doc/example saying so? Validation/mutation that involves cluster state is typically handled only by the controller by definition.

JulienBreux:
> > I encountered an issue when I wanted to employ the best practice for using the Kubernetes client in a webhook.
> 
> @JulienBreux is this really a best practice? Can you point me to an upstream doc/example saying so? Validation/mutation that involves cluster state is typically handled only by the controller by definition.

Not a best practice. But in my example, I've two CRDs and the second must check if the resource exists in the first.

estroz:
>Not a best practice. But in my example, I've two CRDs and the second must check if the resource exists in the first.

In that case, (1) your controller should be the code entity enforcing such a constraint, since it's the component that reconciles cluster state, and (2) your second CRD should have a `status.conditions` state that says "first resource does not exist", set and understood by the controller. Having a network-bound call in a webhook is bad practice since it adds significant latency to each webhook call. Logically, you should be able to create an internally valid object of the second CRD regardless of the existence of the first since it is internally valid.

camilamacedo86:
I hope that you do not mind if we close this PR based on the explanations done above.

PR #2236, opened 2021-06-25
✨  Ask user to perform generate target instead of automatically called by create api

### LINKED ISSUE #2188: Create api is not calling make manifests

The full description can be checked in: 
https://github.com/operator-framework/operator-sdk/issues/4928

### Reproducing this issue

Just run create api and see that the CRD is not generated. 

### KubeBuilder (CLI) Version

3.0.0

### PROJECT version

3.0.0

### Plugin versions

```yaml
v3
```

### Other versions

master

### Extra Labels

/kind regression

guhaneswaran:
Removes the functionality of generating the samples (CR) by default and puts a message to the user to take action on generating the target to get the manifests using `make generate` giving more control to the user. 

This action was done automatically by 'create api' earlier

Fixes #2188.

### DIFF
--- a/pkg/plugins/golang/v2/api.go
+++ b/pkg/plugins/golang/v2/api.go
@@ -74,6 +74,9 @@ make generate will be run.
   # Edit the Controller Test
   nano controllers/frigate/frigate_controller_test.go
 
+  # Generate the manifests
+  make manifests
+
   # Install CRDs into the Kubernetes cluster using kubectl apply
   make install
 
@@ -165,6 +168,7 @@ func (p *createAPISubcommand) PostScaffold() error {
 		if err != nil {
 			return err
 		}
+		fmt.Print("Next: implement your new API and generate the manifests (e.g. CRDs,CRs) with:\n$ make manifests \n")
 	}
 
 	return nil
--- a/pkg/plugins/golang/v3/api.go
+++ b/pkg/plugins/golang/v3/api.go
@@ -82,6 +82,9 @@ make generate will be run.
   # Edit the Controller Test
   nano controllers/frigate/frigate_controller_test.go
 
+  # Generate the manifests
+  make manifests
+
   # Install CRDs into the Kubernetes cluster using kubectl apply
   make install
 
@@ -188,6 +191,7 @@ func (p *createAPISubcommand) PostScaffold() error {
 		if err != nil {
 			return err
 		}
+		fmt.Print("Next: implement your new API and generate the manifests (e.g. CRDs,CRs) with:\n$ make manifests\n")
 	}
 
 	return nil
--- a/docs/book/src/cronjob-tutorial/running.md
+++ b/docs/book/src/cronjob-tutorial/running.md
@@ -1,5 +1,12 @@
 # Running and deploying the controller
 
+### Optional
+If opting to make any changes to the API definitions, then before proceeding,
+generate the manifests like CRs or CRDs with 
+```bash
+make manifests
+```
+
 To test out the controller, we can run it locally against the cluster.
 Before we do so, though, we'll need to install our CRDs, as per the [quick
 start](/quick-start.md).  This will automatically update the YAML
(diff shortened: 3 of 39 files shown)


pkg/plugins/golang/v3/api.go
```diff
@@ -184,10 +184,7 @@ func (p *createAPISubcommand) PostScaffold() error {
 		return err
 	}
 	if p.runMake && p.resource.HasAPI() {
-		err = util.RunCmd("Running make", "make", "generate")
-		if err != nil {
-			return err
-		}
+		fmt.Printf("Next: generate targets with:\n$ make generate\n")
```
camilamacedo86, line 187: IHMO I would still prefer gen all always that would make the life easier for users by sure.
However, if we would like to move in this way, then we need a better explanation. 
We need to output all commands that should be executed afterwords and why. 

c/c @estroz @jmrodri

camilamacedo86 (commented):
This PR would not be seeding.
We are changing the behaviour

pkg/plugins/golang/v3/api.go
guhaneswaran: @camilamacedo86 , I might have got the context wrong. Do you expect 'create api' to show the output message/log of 'make generate' to the user ?

estroz (changes requested):
This PR doesn't implement the solution described in #2188, which is to remove the CR scaffolds and write a message about running `make manifests` and creating a sample CR in `config/samples`.

One amendment I'd like to make to ^ is: create the CR as kubebuilder does now, but remove `foo: bar` from the spec. This bare CR may still be invalid, but it won't have any unexpected fields that require changing.

pkg/plugins/golang/v3/api.go
camilamacedo86: You can check the consensus reached out in the meeting: https://github.com/kubernetes-sigs/kubebuilder/issues/2188#issuecomment-852345561.

- do not gen the samples (cr) by default 
- then, add a nice msg to let the user knows that is required to gen the target to get the manifests.

So, are we not generating other files with `make generate` or will it only generate the CR? 
IHMO the most important here would output a msg with all further steps and make clear what need to be done and why after the create API. We might need to supplement the quick start as well. Could you please give a look at that and see if would be better we add a note there with this info as well? WDYT?

pkg/plugins/golang/v2/api.go
```diff
@@ -165,6 +165,7 @@ func (p *createAPISubcommand) PostScaffold() error {
 		if err != nil {
 			return err
 		}
+		fmt.Printf("TODO: Edit the API definition & reconciliation business logic\nNext: generate targets with:\n$ make manifests\n")
```
camilamacedo86, line 168: I am not sure if the text here is clear enough. WDYT about something more similar to the other places (e.g. [here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/init.go#L154)) such as:

```suggestion
		fmt.Print("Next: implement your new API and generate the manifests (e.g. CRDs,CRs) with:\n$ make manifests \n")
```

It is not passing in the CI because of the size of the line. You can test it locally by running `make lint`. 
Also, ensure that you executed `make testata-generated` before pushing the changes. 
Then, can you also rebase your commits?

guhaneswaran: Thanks for the comments. The changes have been updated as per the sample. Also CI checks done manually now.

guhaneswaran:
> @guhaneswaran can you squash your commits into 1? Then I'll lgtm/approve.

@estroz , now the commits are merged into 1. Thanks

estroz:
@guhaneswaran the e2e tests need to be updated so the CR's spec is implemented (add `foo: bar` to the CR file before applying it).

guhaneswaran:
@estroz , Thanks for your comments. Updated to include foo: bar during runtime in testdata for v2 and v3

guhaneswaran:
@estroz I think the previously added label - do-not-merge/hold is blocking the merge by not allowing lgtm label

PR #2241, opened 2021-06-29
:book: Fix `version` command

SeriyBg:
Seems like `--version` is an old flag that has been replaced with `version` command


estroz (approved):
/approve
/lgtm

PR #2243, opened 2021-06-30
📖 Plugin Phase 2 Enhancement Proposal

rashmigottipati:
Description of the change
This PR adds an enhancement proposal for Kubebuilder Phase 2 Plugins

Signed-off-by: Rashmi Gottipati <chowdary.grashmi@gmail.com>

### DIFF
--- a/designs/extensible-cli-and-scaffolding-plugins-phase-2.md
+++ b/designs/extensible-cli-and-scaffolding-plugins-phase-2.md
@@ -1,0 +1,356 @@
+# Extensible CLI and Scaffolding Plugins - Phase 2
+
+## Overview
+
+Plugin [Phase 1.5](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md) was designed and implemented to allow chaining of plugins. The purpose of Phase 2 plugins is to discover and use external plugins, also referred to as out-of-tree plugins (which can be implemented in any language). Phase 2 achieves both chaining and discovery of external plugins/source code not compiled with the `kubebuilder` CLI binary. By achieving this goal, we could (for example) externalize the optional [declarative plugin](https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/golang/declarative/v1) which means that the CLI would still be able to use it, however, its source code would no longer be required to be inside of the Kubebuilder repository.
+
+### Related issues and PRs
+
+* [Feature Request: Plugins Phase 2](https://github.com/kubernetes-sigs/kubebuilder/issues/1378)
+* [Extensible CLI and Scaffolding Plugins - Phase 1.5](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md)
+* [Phase 1.5 Implementation PR](https://github.com/kubernetes-sigs/kubebuilder/pull/2060)
+* [Plugin Resolution Enhancement Proposal](https://github.com/kubernetes-sigs/kubebuilder/pull/1942)
+
+### Prototype implementation
+
+[POC](https://github.com/rashmigottipati/POC-Phase2-Plugins) - Invoke an external python program that simulates a plugin from a go main and pass messages from `kubebuilder` to the plugin and vice-versa using `stdin/stdout/stderr`.
+
+### User Stories
+
+* As a plugin developer, I would like to be able to provide external plugins path for the CLI to perform the scaffolds, so that I could take advantage of external initiatives which are implemented using Kubebuilder as a lib and following its standards but are not shipped with its CLI binaries.
+
+* As a Kubebuilder maintainer, I would like to support external plugins not maintained by the core project.
+  * For example, once the Phase 2 plugin implementation is completed, some internal plugins can be re-implemented as external plugins removing the necessity to build those plugins in the `kubebuilder` binary.
+
+### Goals
+
+* `kubebuilder` is able to discover plugin binaries and run those plugins using the CLI.
+
+* Kubebuilder can use the external plugins as well as its own internal ones to do scaffolding.
+
+* `kubebuilder` should be able to show plugin specific information via the `--help` flag.
+
+* Support for standard streams i.e. `stdin/stdout/stderr` as the only IPC method between `kubebuilder` and plugins.
+
+* Kubebuilder library consumers can support chaining and discovery of out-of-tree plugins.
+
+### Non-Goals
+
+* Addition of new arbitrary subcommands other than the subcommands that we already support i.e `init`, `create api`, and `create webhook`.
+
+* Discovering plugin binaries that are not locally present on the machine (i.e. binary exists in a remote repository).
+
+* Providing other options (other than standard streams such as `stdin/stdout/stderr`) for inter-process communication between `kubebuilder` and external plugins.
+  * Other IPC methods may be allowed in the future, although EPs are required for those methods.
+
+### Examples
+
+* `kubebuilder create api --plugins=myexternalplugin/v1`
+  * should scaffold files using the external plugin as defined in its implementation of the `create api` method.
+
+* `kubebuilder create api --plugins=myexternalplugin/v1,myotherexternalplugin/v2`
+  * should scaffold files using the external plugin as defined in their implementation of the `create api` method (by respecting the plugin chaining order, i.e. in the order of `create api` of v1 and then `create api` of v2 as specified in the layout field in the configuration).
+
+* `kubebuilder create api --plugins=myexternalplugin/v1 --help`
+  * should display help information of the plugin which is not shipped in the binary (myexternalplugin/v1 is present outside of the `kubebuilder` binary).
+
+* `kubebuilder create api --plugins=go/v3,myexternalplugin/v2`
+  * should create files using the `go/v3` plugin, then pass those files to `myexternalplugin/v2` as defined in its implementation of the `create api` method by respecting the plugin chaining order.
+
+## Proposal
+
+### Discovery of plugin binaries
+
+The method [kustomize](https://kubectl.docs.kubernetes.io/guides/extending_kustomize/) uses to discover plugins, by following a GVK path scheme, is the most natural for this use case since plugins must have a group-like name and version.
+
+Every plugin gets its own directory constructed using the plugin name and plugin version for the executable to be placed in and `kubebuilder` will search for a plugin binary with the name of the plugin in the `${name}/${version}` directory of the plugin. This information (plugin name and plugin version) is obtained by `kubebuilder` via the value passed to the `--plugins` CLI flag. Once `kubebuilder` successfully locates the plugin, it will run the plugin using the CLI.
+
+Every plugin gets its own directory as below.
+
+On Linux:
+
+```shell
+    $XDG_CONFIG_HOME/kubebuilder/plugins/${name}/${version}
+```
+
+The default value of XDG_CONFIG_HOME is `$HOME/.config`.
+
+On OSX:
+
+```shell
+    ~/Library/Application Support/kubebuilder/plugins/${name}/${version}
+```
+
+Based on the above directory scheme, let's say that if the value passed to the `--plugins` CLI flag is `myexternalplugin/v1`:
+
+* On Linux:
+  * `kubebuilder` will search for the `myexternalplugin` binary in `$XDG_CONFIG_HOME/kubebuilder/plugins/myexternalplugin/v1`, where the base of this path in is the binary name.
+* On OSX:
+  * Kubebuilder will search for the `myexternalplugin` binary in `$HOME/Library/Application Support/kubebuilder/plugins/myexternalplugin/v1`.
+
+Note: If the name is ambiguous, then the qualified name `myexternalplugin.my.domain` would be used, so the path would be `$XDG_CONFIG_HOME/kubebuilder/plugins/my/domain/myexternalplugin/v1` on Linux and `$HOME/Library/Application Support/kubebuilder/plugins/my/domain/myexternalplugin/v1` on OSX.
+
+* Pros
+  * `kustomize` which is popular and robust tool, follows this approach in which `apiVersion` and `kind` fields are used to locate the plugin.
+
+  * This approach enforces naming constraints as the permitted character set must be directory name-compatible following naming rules for both Linux and OSX systems.
+
+  * The one-plugin-per-directory requirement eases creation of a plugin bundle for sharing.
+
+### What Plugin system should we use
+
+I propose we use our own plugin system that passes JSON blobs back and forth across `stdin/stdout/stderr` and make this the only option for now as it’s a language-agnostic medium and it is easy to work with in most languages.
+
+We came to the conclusion that a kubebuilder-specific plugin library should be written after evaluating plugin libraries such as the [built-in go-plugin library](https://golang.org/pkg/plugin/) and [Hashicorp’s plugin library](https://github.com/hashicorp/go-plugin):
+
+* The built-in plugin library seems to be more suitable for in-tree plugins rather than out-of-tree plugins and it doesn’t offer cross-language support, thereby making it a non-starter.
+* Hashicorp’s go plugin system is more suitable than the built-in go-plugin library as it enables cross language/platform support. However, it is more suited for long running plugins as opposed to short lived plugins and the usage of protobuf could be overkill as we will not be handling 10s of 1000s of deserializations.
+
+In the future, if a need arises (for example, users are hitting performance issues), we can then explore the possibility of using the Hashicorp’s go plugin library. From a design standpoint, to leave it architecturally open, I propose using a `type` field in the PROJECT file to potentially allow other plugin libraries in the future and make this a seperate field in the PROJECT file per plugin; and this field determines how the `universe` will be passed for a given plugin. However, for the sake of simplicity in initial design and not to introduce any breaking changes as Project version 3 would suffice for our needs, this option is out of scope in this proposal.
+
+### Project configuration
+
+Currently, the project configuration has two fields to store plugin specific information.
+
+* `Layout` field (of type []string) is used for plugin chain resolution on initialized projects. This will be the default if no plugins are specified for a subcommand.
+* `Plugins` field (of type map[string]interface{}) is used for option plugin configuration that stores configuration information of any plugin.
+
+* So, where should external plugins be defined in the configuration?
+
+  * I propose that the external plugin should get encoded in the project configuration as a part of the `layout` field.
+    * For example, external plugin `myexternalplugin/v2` can be specified through the `--plugins` flag for every subcommand and also be defined in the project configuration in the `layout` field for plugin resolution.
+
+Example `PROJECT` file:
+
+```yaml
+version: "3"
+domain: testproject.org
+layout: 
+- go.kubebuilder.io/v3
+- myexternalplugin/v2
+plugins:
+  myexternalplugin/v2:
+    resources:
+    - domain: testproject.org
+      group: crew
+      kind: Captain
+      version: v2-alpha
+  declarative.go.kubebuilder.io/v1:
+    resources:
+    - domain: testproject.org
+      group: crew
+      kind: FirstMate
+      version: v1
+repo: github.com/test-inc/testproject
+resources:
+- group: crew
+  kind: Captain
+  version: v1
+```
+
+### Communication between `kubebuilder` and external plugins
+
+* Why do we need communication between `kubebuilder` and external plugins?
+
+  * The in-tree plugins do not need any inter-process communication as they are the same process, and hence, direct calls are made to the respective functions (also referred as hooks) based on the supported subcommands for an in-tree plugin. As Phase 2 plugins is tackling out-of-tree or external plugins, there's a need for inter-process communication between `kubebuilder` and the external plugin as they are two separate processes/binaries. `kubebuilder` needs to communicate the subcommand that the external plugin should run, and all the arguments received in the CLI request by the user. These arguments contain flags which will have to be directly passed to all plugins in the chain. Additionally, it's important to have context of all the files that were scaffolded until that point especially if there is more than one external plugin in the chain. `kubebuilder` attaches that information in the request, along with the command and arguments. For the external plugin, it would need to communicate the subcommand it ran and the updated file contents information that the external plugin scaffolded to `kubebuilder`. The external plugin would also need to provide its help text if requested by `kubebuilder`. As discussed earlier, standard streams seems to be a desirable IPC method of communication for the use-cases that Phase 2 is trying to solve that involves discovery and chaining of external plugins.
+
+* How does `kubebuilder` communicate to external plugins?
+
+  * Standard streams have three I/O connections: standard input (`stdin`), standard output (`stdout`) and standard error (`stderr`) and they work well with chaining applications, meaning that output stream of one program can be redirected to the input stream of another.
+  * Let's say there are two external plugins in the plugin chain. Below is the sequence of how `kubebuilder` communicates to the plugins `myfirstexternalplugin/v1` and `mysecondexternalplugin/v1`.
+
+![Kubebuilder to external plugins sequence diagram](https://github.com/rashmigottipati/POC-Phase2-Plugins/blob/main/docs/externalplugins-sequence-diagram.png)
+
+* What to pass between `kubebuilder` and an external plugin?
+
+Message passing between `kubebuilder` and the external plugin will occur through a request / response mechanism. The `PluginRequest` will contain information that `kubebuilder` sends *to* the external plugin. The `PluginResponse` will contain information that `kubebuilder` receives *from* the external plugin.
+
+The following scenarios shows what `kubebuilder` will send/receive to the external plugin:
+
+* `kubebuilder` to external plugin:
+  * `kubebuilder` constructs a `PluginRequest` that contains the `Command` (such as `init`, `create api`, or `create webhook`), `Args` containing all the raw flags from the CLI request and license boilerplate without comment delimiters, and an empty `Universe` that contains the current virtual state of file contents that is not written to the disk yet. `kubebuilder` writes the `PluginRequest` through `stdin`.
+
+* External plugin to `kubebuilder`:
+  * The plugin reads the `PluginRequest` through its `stdin` and processes the request based on the `Command` that was sent. If the `Command` doesn't match what the plugin supports, it writes back an error immediately without any further processing. If the `Command` matches what the plugin supports, it constructs a `PluginResponse` containing the `Command` that was executed by the plugin, and modified `Universe` based on the new files that were scaffolded by the external plugin, `Error` and `ErrorMsg` that add any error information, and writes the `PluginResponse` back to `kubebuilder` through `stdout`.
+
+* Note: If `--help` flag is being passed from `kubebuilder` to the external plugin through `PluginRequest`, the plugin attaches its help text information in the `Help` field of the `PluginResponse`. Both `PluginRequest` and `PluginResponse` also contain `APIVersion` field to have compatible versioned schemas.
+
+* Handling plugin failures across the chain:
+
+  * If any plugin in the chain fails, the plugin reports errors back through `PluginResponse` to `kubebuilder` and plugin chain execution will be halted, as one plugin may be dependent on the success of another. All the files that were scaffolded already until that point will not be written to disk to prevent a half committed state.
+
+## Implementation Details/Notes/Constraints
+
+`PluginRequest` holds all the information `kubebuilder` receives from the CLI and the plugins that were executed before it and the `PluginRequest` will be marshaled into a JSON and sent over `stdin` to the external plugin. `PluginResponse` is what the plugin constructs with the updated universe and sent back to `kubebuilder`. The following structs would be defined on the Kubebuilder side.
+
+```go
+// PluginRequest contains all information kubebuilder received from the CLI
+// and plugins executed before it.
+type PluginRequest struct {
+  // Command contains the command to be executed by the plugin such as init, create api, etc.
+  Command       string              `json:"command"`
+
+  // APIVersion defines the versioned schema of the PluginRequest that is encoded and sent from Kubebuilder to plugin.
+  // Initially, this will be marked as alpha (v1alpha1).
+  APIVersion    string              `json:"apiVersion"`
+
+  // Args holds the plugin specific arguments that are received from the CLI which are to be passed down to the plugin.
+  Args          []string            `json:"args"`
+
+  // Universe represents the modified file contents that gets updated over a series of plugin runs
+  // across the plugin chain. Initially, it starts out as empty.
+  Universe      map[string]string   `json:"universe"`
+}
+
+// PluginResponse is returned to kubebuilder by the plugin and contains all files
+// written by the plugin following a certain command.
+type PluginResponse struct {
+  // Command holds the command that gets executed by the plugin such as init, create api, etc.
+  Command       string                   `json:"command"`
+
+  // Help contains the plugin specific help text that the plugin returns to kubebuilder when it receives
+  // `--help` flag from Kubebuilder.
+  Help          string                   `json:"help,omitempty"`
+
+  // APIVersion defines the versioned schema of the PluginResponse that will be written back to kubebuilder.
+  // Initially, this will be marked as alpha (v1alpha1).
+  APIVersion    string                   `json:"apiVersion"`
+
+  // Universe in the PluginResponse represents the updated file contents that was written by the plugin.
+  Universe      map[string]string        `json:"universe"`
+
+  // Error is a boolean type that indicates whether there were any errors due to plugin failures.
+  Error         bool                     `json:"error,omitempty"`
+
+  // ErrorMsg holds the specific error message of plugin failures.
+  ErrorMsg      string                   `json:"error_msg,omitempty"`
+}
+```
+
+The following function handles construction of the `PluginRequest` based on the information `kubebuilder` receives from the CLI and the request is marshaled into JSON. The command to run the external plugin by providing the plugin path will be invoked and `kubebuilder` will send the marshaled `PluginRequest` JSON to the plugin over `stdin`.
+
+```go
+func (p *ExternalPlugin) runExternalProgram(req PluginRequest) (res PluginResponse, err error) {
+  pluginReq, err := json.Marshal(req)
+  if err != nil {
+    return res, err
+  }
+
+  cmd := exec.Command(p.Path)
+  cmd.Dir = p.DirContext
+  cmd.Stdin = bytes.NewBuffer(pluginReq)
+  cmd.Stderr = os.Stderr
+
+  out, err := cmd.Output()
+  if err != nil {
+    fmt.Fprint(os.Stdout, string(out))
+    return res, err
+  }
+
+  if json.Unmarshal(out, &res); err != nil {
+    return res, err
+  }
+  return res, nil
+}
+```
+
+On the plugin side, the request JSON will be decoded and depending on what the `Command` in the `PluginRequest` is, the corresponding function to handle `init` or `create api` will be invoked thereby modifying the universe by writing the updated files to it. After `init` or `create api` functions execute successfully, the plugin will write back `PluginResponse` with updated universe and errors (if any) in JSON format through `stdout` to `kubebuilder`. `PluginResponse` also contains error fields `Error` and `ErrorMsg` that the plugin can utilize to add error context if any errors occur.
+`kubebuilder` receives the command output and decodes into `PluginResponse` struct. This is how message passing will occur between `kubebuilder` and the external plugin. Refer to [POC](https://github.com/rashmigottipati/POC-Phase2-Plugins) for specifics.
+
+### Simple Example
+
+```shell
+kubebuilder init --plugins=myexternalplugin/v1 --domain example.com
+```
+
+What happens when the above is invoked?
+
+![Kubebuilder to external plugins](https://github.com/rashmigottipati/POC-Phase2-Plugins/blob/main/docs/externalplugins-sequence-diagram-2.png)
+
+* `kubebuilder` discovers `myexternalplugin/v1` plugin binary and runs the plugin from the discovered path.
+
+* Send `PluginRequest` as a JSON over `stdin` to `myexternalplugin` plugin.
+
+`PluginRequest JSON`:
+
+```JSON
+{ 
+  "command":"init",
+  "args":["--domain","example.com"],
+  "universe":{}
+}
+```
+
+* `myexternalplugin` plugin parses the `PluginRequest` and based on the `Command` specified in the request i.e `init`, performs the necessary scaffolding.
+
+* `myexternalplugin` plugin constructs `PluginResponse` with modified `Universe` that contains the updated file contents and errors if any.
+
+* Plugin writes `PluginResponse` to stdout in a JSON format back to `kubebuilder`.
+
+* `kubebuilder` receives the command output containing the `PluginResponse` JSON which will be decoded into the `PluginResponse` struct.
+
+* `kubebuilder` writes the files in the universe to disk.
+
+`PluginResponse JSON`:
+
+```JSON
+{
+  "command": "init",
+  "universe": {
+    "LICENSE": "Apache 2.0 License\n",
+    "main.py": "..."
+  }
+}
+
+```
+
+## Alternatives
+
+### Plugin discovery
+
+#### User specified file paths
+
+A user will provide a list of file paths for `kubebuilder` to discover the plugins in. We will define a variable `KUBEBUILDER_PLUGINS_DIRS` that will take a list of file paths to search for the plugin name. It will also have a default value to search in, in case no file paths are provided. It will search for the plugin name that was provided to the `--plugins` flag in the CLI. `kubebuilder` will recursively search for all file paths until the plugin name is found and returns the successful match, and if it doesn’t exist, it returns an error message that the plugin is not found in the provided file paths. Also use the host system mechanism for PATH separation.
+
+* Alternatively, this could be handled in a way that [helm kustomize plugin](https://helm.sh/docs/topics/advanced/#post-rendering) discovers the plugin based on the non-existence of a separator in the path provided, in which case `kubebuilder` will search in `$PATH`, otherwise resolve any relative paths to a fully qualified path.
+
+* Pros
+  * This provides flexibility for the user to specify the file paths that the plugin would be placed in and `kubebuilder` could discover the binaries in those user specified file paths.
+
+  * No constraints on plugin binary naming or directory placements from the Kubebuilder side.
+
+  * Provides a default value for the plugin directory in case user wants to use that to drop their plugins.
+
+#### Prefixed plugin executable names in $PATH
+
+Another approach is adding plugin executables with a prefix `kubebuilder-` followed by the plugin name to the PATH variable. This will enable `kubebuilder` to traverse through the PATH looking for the plugin executables starting with the prefix `kubebuilder-` and matching by the plugin name that was provided in the CLI. Furthermore, a check should be added to verify that the match is an executable or not and return an error if it's not an executable. This approach provides a lot of flexibility in terms of plugin discovery as all the user needs to do is to add the plugin executable to the PATH and `kubebuilder` will discover it.
+
+* Pros
+  * `kubectl` and `git` follow the same approach for discovering plugins, so there’s prior art.
+
+  * There’s a lot of flexibility in just dropping plugin binaries to PATH variable and enabling the discovery without having to enforce any other constraints on the placements of the plugins.
+
+* Cons
+  * Enumerating the list of all available plugins might be a bit tough compared to having a single folder with the list of available plugins and having to enumerate those.
+
+  * These plugin binaries cannot be run in a standalone manner outside of Kubebuilder, so may not be very ideal to add them to the PATH var.
+
+## Open questions
+
+* Do we want to support the addition of new arbitrary subcommands other than the subcommands (init, create api, create webhook) that we already support?
+  * Not for the EP or initial implementation, but can revisit later.
+
+* Do we need to discover flags by calling the plugin binary or should we have users define them in the project configuration?
+  * Flags will be passed directly to the external plugins as a string. Flag parse errors will be passed back via `PluginResponse`.
+
+* What alternatives to stdin/stdout exist and why shouldn't we use them?
+  * Other alternatives exist such as named pipe and sockets, but stdin/stdout seems to be more suitable for our needs.
+
+* What happens when two plugins bind the same flag name? Will there be any conflicts?
+  * As mentioned in the implementation details section, flags are passed directly as a string to plugins and the same string will be passed to each plugin in the chain, so all plugins get the same flag set. Errors should not be returned if an unrecognized flag is parsed.
+
+* How should we handle environment variables?
+  * We would pass the entire CLI environment to the plugin to permit simple external plugin configuration without jumping through hoops.
+
+* Should the API version be a part of the plugin request spec?
+  * It would be nice to encode APIVersion for `PluginRequest` and `PluginResponse` so the initial schemas can be marked as `v1alpha1`.


designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
+# Extensible CLI and Scaffolding Plugins - Phase 2
+
+## Overview
+
+* Plugin Phase 1.5 was desgined and implemented to achieve chaining of plugins. The main goal of Phase 2 plugins is Kubebuilder discovering and using external plugins, also referred as out-of-tree plugins. So Phase 2 achieves both chaining of plugins and discovery of external plugins/source code which is not intented to be shipped with the binaries. For e.g., Kubebuilder being able to discover the helm plugin and use that without having to build the plugin with the binary. These external plugins can be implemented in any language.
```
camilamacedo86, line 5:
```suggestion

UPlugin [Phase 1.5](thttps://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md) was designed and implemented to allow chaining plugins. 

The purpose of Phase 2 plugins is to allow discovering and using external plugins, also referred to as out-of-tree plugins. So Phase 2 achieves both chainings of plugins and discovery of external plugins/source code which is not intended to be shipped with the CLI binary. By achieving this goal, we could externalize the optional [declarative](https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/golang/declarative/v1) which means that the CLI would still able to use it, however, its code source no longer would require to be inside of the Kubebuilder repository as built with the CLI release. 
```

We have not Helm plugin in KB. In this way, I would suggest we stick to what we can find here.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+### Related issues and PRs
+
+* [Feature Request: Plugins Phase 2](https://github.com/kubernetes-sigs/kubebuilder/issues/1378)
+* [Extensible CLI and Scaffolding Plugins - Phase 1.5](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md)
+* [Phase 1.5 Implementation PR](https://github.com/kubernetes-sigs/kubebuilder/pull/2060)
+* [Plugin Resolution Enhancement Proposal](https://github.com/kubernetes-sigs/kubebuilder/pull/1942)
+
+### Use Cases
+
+* As a CLI user, I would like to be able to provide external source code of my plugins and its boilerplates for the CLI to be able to use them by extending the subcommands provided by Kubebuilder, which will not have the plugin binaries be shipped by default.
```
camilamacedo86, line 16:
```suggestion
* I am an operator author, I would like to be able to provide external plugins path for the CLI to perform the scaffolds, so that I could take advantage of external initiatives which are implemented using Kubebuilder as a lib and following its standards but are not shipped with its CLI binaries. 
```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+* Operator SDK is able to extend Kubebuilder and provide the same functionality for plugins chaining and enabling out-of-tree plugin discovery.
+
+### Non-Goals
+
+* Addition of new arbitrary subcommands other than the subcommands that we already support i.e init, create api, and create webhook.
+
+* Discovering plugin binaries that are not locally present on the machine (i.e. exists in a remote repository).
+
+* Discovering flags that each individual plugin would support
+  * Flags would have to be provided by the user as a part of project configuration.
```
camilamacedo86, line 36: What does it mean?

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+* `kubebuilder create api --plugins=myexternalplugin/v1`
+  * should scaffold files using the source code of the external plugin as defined in its implementation of the create api method.
+
+* `kubebuilder create api --plugins=myexternalplugin/v1,myexternalplugin/v2`
+  * should scaffold files using the source code of the external plugin as defined in their implementation of the create api method(by respecting the plugin chaining order, i.e. in the order of create api of v1 and then create api of v2 as specified in the layout field in the configuration).
+
+* `kubebuilder create api --plugins=myexternalplugin/v1 --help`
+  * should provide help implementation of the plugin which is not shipped in the binary (source code is present outside of the KB repo).
+
+### How should Kubebuilder discover plugin binaries
```
camilamacedo86, line 51: I think that would make it easier if we try to make it closer to the template/standard adopted in OF. See: https://github.com/operator-framework/enhancements/blob/master/enhancements/template.md

Ditto so, I think we need here: 
- Proposal (This is where we get down to the nitty gritty of what the proposal actually is.)
- Implementation Details/Notes/Constraints [optional]
- Risks and Mitigations
- Design Details

Then, you will add here first all design and solution that you believe would be the best option/approach.

Then, we would have an `Alternatives` section where you would add the extra options.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+#### Handling plugin failures across the chain
+
+If any plugin in the chain fails, we will halt the chain execution and report errors as one plugin may be dependent on the success of another. We can use `stderr` for the plugin to write back warnings/errors in a structured format that Kubebuilder understands and return a process exit code 1. The successful return type with exit code 0 contains the list of files that were scaffolded and the file contents will be given back to Kubebuilder.
+
+In the case of plugin failures even if one plugin across the chain fails, all the files that were scaffolded already until that point will not be written to disk to prevent a half committed state.
+
+#### Open questions
+
+* Do we want to support the addition of new arbitrary subcommands other than the subcommands (init, create api, create webhook) that we already support?
```
camilamacedo86, line 193: Why this question would be part of the scope of plugin phase 2? Cannot we add new commands interfaces when/if we wish? I am not sure if I could follow the concern to add that here.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+* Do we need to discover flags by calling the plugin binary or should we have users define them in the project configuration?
```
camilamacedo86, line 195: The PROJECT is the config of the project scaffold and not of the CLI tool. 
I understand that we need to inform the tool/CLI where the plugins are for it to be able to use them. Then, the data scaffold in the PROJECT file should respect and work as it works for the plugins which are shipped in tool binary.

For example, if we run `create api` with the external plugin we will be able to know that by looking the PROJECT file such as we can check that by looking the plugins attr.  https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator/PROJECT#L4-L6

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+* What alternatives to stdin/stdout exist and why shouldn't we use them?
+  * Other alternatives exist such as named pipe and sockets, but stdin/stdout seems to be more suitable for our needs.
+
+* What happens when two plugins bind the same flag name? Will there be any conflicts?
```
camilamacedo86, line 200: What happens now with the changing implementation (plugin 1.5)? I'd expected the same bahaviour.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+* Should plugins depend on environment variables?
+  * The same plugin would behave differently based on the environment.
```
camilamacedo86, line 203: Not sure if I followed this one. I understand that we need a way to say: CLI(sdk/kb) look in path X to find my external plugins. The plugins would need to respect the contract/interfaces of KB lib. However, what the specific plugin will or not do internally it is up to the plugins is outside of the CLI/lib scope.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+* `kubebuilder create api --plugins=myexternalplugin/v1,myexternalplugin/v2`
+  * should scaffold files using the source code of the external plugin as defined in their implementation of the create api method(by respecting the plugin chaining order, i.e. in the order of create api of v1 and then create api of v2 as specified in the layout field in the configuration).
+
+* `kubebuilder create api --plugins=myexternalplugin/v1 --help`
+  * should provide help implementation of the plugin which is not shipped in the binary (source code is present outside of the KB repo).
+
+### How should Kubebuilder discover plugin binaries
+
+Main goal of phase 2 plugins is discovery of out of tree plugins that are external to Kubebuilder and not shipped with the CLI. To be able to run these plugins via the Kubebuilder CLI, it’s necessary for kubebuilder to discover these external plugin binaries and use them as internal ones. I will discuss a few approaches in the order of suitability and preference, with their respective pros and cons.
```
camilamacedo86, line 53: (IMHO) as part of this EP, we need a POC with the suggested approach such as it was done for plugin phase 1. See https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-1.md#prototype-implementation. We need a better definition over how it will be done and something that illustrates that is really achievable. Ps.: For plugin phase 1.5 the EP was made with its main implementation.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+Another approach I would like to mention is adding plugin executables with a prefix `kubebuilder-` followed by the plugin name to the PATH variable. This will enable KB to traverse through the PATH looking for the plugin executables starting with the prefix `kubebuilder-` and matching by the plugin name that was provided in the CLI. Furthermore, a check should be added to verify that the match is an executable or not and return an error if it's not an executable. This approach provides a lot of flexibility in terms of plugin discovery as all the user needs to do is to add the plugin executable to the PATH and Kubebuilder will discover it.
+
+* Pros
+  * `kubectl` and `git` follow the same approach for discovering plugins, so there’s prior art.
+
+  * There’s a lot of flexibility in just dropping plugin binaries to PATH variable and enabling the discovery without having to enforce any other constraints on the placements of the plugins.
+
+* Cons
+  * Enumerating the list of all available plugins might be a bit tough compared to having a single folder with the list of available plugins and having to enumerate those.
```
jmrodri, line 103: We could cache the list in a config after we discover them for performance.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+  * These plugin binaries cannot be run in a standalone manner outside of Kubebuilder, so may not be very ideal to add them to the PATH var.
```
jmrodri, line 105: Is that true? Nothing should prevent someone from making their plugin act like its own program if they so choose. The only restriction should be that the plugin respect the contract that we specify i.e. STDIN/STDOUT some format of information that goes between that. Other than that, a plugin is just another executable that could do whatever it wants.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+#### What Plugin system should we use
+
+Based on the evaluation of plugin libraries such as the [built-in go-plugin library](https://golang.org/pkg/plugin/) and [Hashicorp’s plugin library](https://github.com/hashicorp/go-plugin), it is more suitable to write our own custom plugin library as the built-in plugin library seems to be more suitable for in-tree plugins rather than out-of-tree plugins and it doesn’t offer cross language support, thereby making it a non-starter. Hashicorp’s go plugin system is more suitable than the built-in go-plugin library as it enables cross language/platform support. However, it is more suited for long running plugins as opposed to short lived plugins and the usage of protobuf could be overkill as we will not be handling 10s of 1000s of deserializations.
+
+For the stated reasons, I propose we use our own plugin system that passes JSON blobs back and forth across stdin/stdout and make stdin/stdout the only default option for now as it’s the most language agnostic way and it makes it easy to work with most languages.
```
jmrodri, line 111: I would flip these two paragraphs. I would start with:

I propose we use our own plugin system ...

We came to this conclusion after evaluating plugin libraries such as the built-ing go-plugin library and Hashicorp's plugin library ...

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+In the future, if a need arises (for e.g. users are hitting performance issues), we can then explore the possibility of using the Hashicorp’s go plugin library. From a design standpoint, to leave it architecturally open, I propose using a “type” field in the PROJECT file to potentially allow other plugin libraries in the future and make this a seperate field in the PROJECT file per plugin; and this field determines how the injector will be passed for a given plugin. However, for the sake of simplicity in initial design and not to introduce any breaking changes as Project version 3 would suffice for our needs, this option is out of scope in this proposal.
```
jmrodri, line 113: I don't think we should ever use Hashicorp's go plugin if it doesn't allow multiple languages. I'd just drop this whole paragraph.
```suggestion
```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
jmrodri: @camilamacedo86 I think the question is today we support (init, create api, create webhook), what if an external plugin wants to add the "camila" subcommand. Do we allow that to be added to the kubebuilder CLI even if it is NOT part of CLI.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+    // boilerplate is the copyright comment added at the top of scaffolded files.
+    boilerplate string
+
+    // resource contains the information of the API that is being scaffolded.
+    resource *resource.Resource
+}
+```
+
+One of the plugin hooks allows plugins to set up their own flags. Commong flags (like `--group`, `--version`, `--kind`) are defined by the plugin system itself, and the resource created with those values populated will be passed to the plugins. `create api` and `create webhook` commands would require the GVK being passed as well to the plugins. Currently, `InjectResource` is a required hook for `create api` and `create webhook` commands and the entire resource (which contains the group, version and kind) is being passed to the plugins for these commands.
```
jmrodri, line 175:
```suggestion
One of the plugin hooks allows plugins to set up their own flags. Common flags (like `--group`, `--version`, `--kind`) are defined by the plugin system itself, and the resource created with those values populated will be passed to the plugins. `create api` and `create webhook` commands would require the GVK being passed as well to the plugins. Currently, `InjectResource` is a required hook for `create api` and `create webhook` commands and the entire resource (which contains the group, version and kind) is being passed to the plugins for these commands.
```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+All the raw flags of the plugins would have to be passed down to the plugins through injector. The plugin specific flags would have to be specified in the project configuration by the user. And the current implmentation is such that, every subcommand can optionally implement the `InjectConfig` hook which passes the project configuration to the plugins and also enable the plugins to modify it if required.
```
jmrodri, line 177: The plugin specific flags would have to be specified in the project configuration by the user? I don't think we should require this.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+Along with flags, boilerplate with empty copyright or an apache license and in go-style comments is also being passed down to the plugins through injector. If other kind of comments were to be supported, this logic would have to be added (scaffold markers do support YAML and Go-style comments).
+
+Additionally, we would pass the entire CLI environment to the plugin, as anything executing in subshell environment would be able to read the environment variables even if it's being restricted in terms of what Kubebuilder passes to it.
+
+For displaying help text information of the plugins, `UpdateMetadata` hook is the one that allows to modify the description and examples for the `--help` message of plugins. Currently, there is no support for help implementaion of plugins so logic would have to be added to the `UpdateMetadata` hook to display plugin specific help.
```
jmrodri, line 183: So how will we pass that injector structure to say a plugin written in python? I would like to see this section expanded more. Let's assume we're passing data down as a JSON blob. What would that injector look like in JSON? What will I as a plugin implementer expect to receive in STDIN?

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+#### Handling plugin failures across the chain
+
+If any plugin in the chain fails, we will halt the chain execution and report errors as one plugin may be dependent on the success of another. We can use `stderr` for the plugin to write back warnings/errors in a structured format that Kubebuilder understands and return a process exit code 1. The successful return type with exit code 0 contains the list of files that were scaffolded and the file contents will be given back to Kubebuilder.
```
jmrodri, line 187: This is the first mention of STDERR I've seen in this proposal.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+In the case of plugin failures even if one plugin across the chain fails, all the files that were scaffolded already until that point will not be written to disk to prevent a half committed state.
```
jmrodri, line 189: How will this be enforced? Again, I have a plugin written in python and one written in java in the chain. How will I enforce that nothing was written unless it's part of the specification.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
jmrodri: @rashmigottipati I believe we should discover flags by calling the plugin binary.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
jmrodri: @rashmigottipati I think I saw above you mention that kubebuilder will pass the environment to the plugins, so I think the plugins should be able to accept or read environment variables.

jmrodri (commented):
A few items to clarify.

jmrodri:
I think we need to add more details on the following:
* what the potential structure looks like being passed in. I would start with assuming JSON format.
* think about the messages we can send to the plugins? i.e. get_flags or register, init, create_api, create_webhook, these will be part of the message being sent. Each of these will have their own set of parameters and return values. The plugin implementer should be able to handle some or all of these, at a minimum `get_flags` or `register`.

jmrodri:
Interestingly enough, a lot of what we're talking about here reminds me of [CORBA](https://en.wikipedia.org/wiki/Common_Object_Request_Broker_Architecture).

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
estroz: Agree with @jmrodri. Aside: this goal is not "as an operator author", it's "as a plugin developer".

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+* [Feature Request: Plugins Phase 2](https://github.com/kubernetes-sigs/kubebuilder/issues/1378)
+* [Extensible CLI and Scaffolding Plugins - Phase 1.5](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md)
+* [Phase 1.5 Implementation PR](https://github.com/kubernetes-sigs/kubebuilder/pull/2060)
+* [Plugin Resolution Enhancement Proposal](https://github.com/kubernetes-sigs/kubebuilder/pull/1942)
+
+### Use Cases
+
+* As a CLI user, I would like to be able to provide external source code of my plugins and its boilerplates for the CLI to be able to use them by extending the subcommands provided by Kubebuilder, which will not have the plugin binaries be shipped by default.
+
+* As a Kubebuilder/Operator SDK maintainer, I would like to be able to export an internal plugin and make that an external source, so that the plugin is no longer officially maintained by the project but still available to be used.
```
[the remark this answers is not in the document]
estroz: This goal seems to be covered by the one above.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+* Plugin Phase 1.5 was desgined and implemented to achieve chaining of plugins. The main goal of Phase 2 plugins is Kubebuilder discovering and using external plugins, also referred as out-of-tree plugins. So Phase 2 achieves both chaining of plugins and discovery of external plugins/source code which is not intented to be shipped with the binaries. For e.g., Kubebuilder being able to discover the helm plugin and use that without having to build the plugin with the binary. These external plugins can be implemented in any language.
+
+### Related issues and PRs
+
+* [Feature Request: Plugins Phase 2](https://github.com/kubernetes-sigs/kubebuilder/issues/1378)
+* [Extensible CLI and Scaffolding Plugins - Phase 1.5](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md)
+* [Phase 1.5 Implementation PR](https://github.com/kubernetes-sigs/kubebuilder/pull/2060)
+* [Plugin Resolution Enhancement Proposal](https://github.com/kubernetes-sigs/kubebuilder/pull/1942)
+
+### Use Cases
```
estroz, line 14:
```suggestion
### Goals
```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+* Kubebuilder is able to discover plugin binaries and run plugins using the CLI.
+
+* Kubebuilder can use the external plugins as its own internal ones to do scaffolding and be able to show plugin specific information such as via the --help flag.
+
+* Operator SDK is able to extend Kubebuilder and provide the same functionality for plugins chaining and enabling out-of-tree plugin discovery.
+
+### Non-Goals
+
+* Addition of new arbitrary subcommands other than the subcommands that we already support i.e init, create api, and create webhook.
```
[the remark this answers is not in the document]
estroz: Maybe, but not as part of this EP I'd say.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+* Discovering plugin binaries that are not locally present on the machine (i.e. exists in a remote repository).
+
+* Discovering flags that each individual plugin would support
+  * Flags would have to be provided by the user as a part of project configuration.
+
+* Providing other options (other than stdin stdout) for inter-process communication between Kubebuilder and the plugins.
```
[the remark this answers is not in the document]
estroz: The intent here is to choose stdin/stdout as the only supported IPC method in the scope of this EP. To define others is either an amendment or its own EP.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Examples
+
+* `kubebuilder create api --plugins=myexternalplugin/v1`
+  * should scaffold files using the source code of the external plugin as defined in its implementation of the create api method.
+
+* `kubebuilder create api --plugins=myexternalplugin/v1,myexternalplugin/v2`
```
estroz, line 45: This example isn't great because you should never call the same plugin twice
```suggestion
* `kubebuilder create api --plugins=myexternalplugin/v1,myotherexternalplugin/v2`
```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+* Pros
+  * `kustomize` which is popular and robust tool, follows this approach in which `apiVersion` and `kind` fields are used to locate the plugin.
+
+  * This appoach does not enforce naming of plugins but enforces the directories the plugins have to be in, which is a more common and typical approach.
+
+  * The one-plugin-per-directory requirement eases creation of a plugin bundle for sharing.
+
+* Cons
+  * This makes it mandatory for every plugin that is specified in the `layout` field to have additional configuration information such as Group, Version and Kind in the `plugins` field in the PROJECT file for it to be discoverable. Currently, the `plugins` field is more like an optional configuration storage field.
```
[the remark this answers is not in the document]
estroz: You should be able to use the value passed to plugins: `--plugins myexternalplugin/v1` has the plugin name and version, so something like `myexternalplugin/v1/myexternalplugin` would be used where the base of that path is the binary's name; if the name is ambiguous then the qualified name `myexternalplugin.my.domain` needs to be used anyway as per the plugin name resolution algo, so the path would be `my.domain/myexternalplugin/v1/myexternalplugin`.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+In the future, if a need arises (for e.g. users are hitting performance issues), we can then explore the possibility of using the Hashicorp’s go plugin library. From a design standpoint, to leave it architecturally open, I propose using a “type” field in the PROJECT file to potentially allow other plugin libraries in the future and make this a seperate field in the PROJECT file per plugin; and this field determines how the injector will be passed for a given plugin. However, for the sake of simplicity in initial design and not to introduce any breaking changes as Project version 3 would suffice for our needs, this option is out of scope in this proposal.
+
+#### Project configuration
+
+Currently, the project configuration has two fields to store plugin specific information.
+
+* `Layout` field (of type string) is used for plugin chain resolution on initialized projects. This will be the default if no plugins are specified for a subcommand.
```
estroz, line 119:
```suggestion
* `Layout` field (of type []string) is used for plugin chain resolution on initialized projects. This will be the default if no plugins are specified for a subcommand.
```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
estroz: Agreed, there's no reason to do this. The plugin binary should handle parsing flags internally.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+All the raw flags of the plugins would have to be passed down to the plugins through injector. The plugin specific flags would have to be specified in the project configuration by the user. And the current implmentation is such that, every subcommand can optionally implement the `InjectConfig` hook which passes the project configuration to the plugins and also enable the plugins to modify it if required.
+
+Along with flags, boilerplate with empty copyright or an apache license and in go-style comments is also being passed down to the plugins through injector. If other kind of comments were to be supported, this logic would have to be added (scaffold markers do support YAML and Go-style comments).
```
estroz, line 179: The plugin can just add comment delimiters itself.
```suggestion
Along with flags, license boilerplate without comment delimiters will be passed down to the plugins through injector.
```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+Additionally, we would pass the entire CLI environment to the plugin, as anything executing in subshell environment would be able to read the environment variables even if it's being restricted in terms of what Kubebuilder passes to it.
```
estroz, line 181: IIRC I said this is not true, and if I didn't I apologize. The child process' environment is completely controlled by the parent process. Regardless I'd say for simplicity we pass the environment to the child process. Of course this brings up security concerns, but you already agree to execute arbitrary code by specifying an external plugin so this is not a top priority imo.
```suggestion
Additionally, we would pass the entire CLI environment to the plugin.
```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
estroz: The idea here is to provide an interface for passing file data between plugins in the chain. If your plugin writes something outside of this interface, it won't be passed to downstream plugins which is undesirable from any perspective. There's nothing stopping a plugin from breaking this interface though. That should be noted here.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
estroz: I'd say no, for now. Can revisit later.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+  * No constraints on plugin binary naming or directory placements from the Kubebuilder side.
+
+  * Provides a default value for the plugin directory in case user wants to use that to drop their plugins.
+
+* Cons
+  * Adds some complexity as we need to have an extra flag to have the user provide those file paths and/or make PROJECT file changes to add an extra field to `plugins` field that is a comma separated list of file paths for every plugin that is defined in the `layout` field.
+
+#### Approach #2
+
+This is how [kustomize](https://kubectl.docs.kubernetes.io/guides/extending_kustomize/) discovers plugins.
```
estroz, line 71: I vote for this approach because this is a common way to configure plugins. See [setup-envtest](https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest) too.

jmrodri:
@estroz made a good point when he mentioned security regarding the environment. What's our answer for security of the plugins? For example, do we prevent or care if someone has malicious code that we could run arbitrarily? Do we need a signature for plugins that can be verified? Or do we leave that up to the user that installed the plugin that they have verified it is from a known source and they have chosen to install it?

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
camilamacedo86: That is an option to solve as well. So, you folks are thinking of a manifest that describes the command that would be called in each scenario, the flags and etc. if it is the preferable option then, I think that the others need to be an alternative section and we need here:

- examples of this file
- how will be the contract 
- how the CLI will use it and etc .. 

I mean more detailed info about its design.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
camilamacedo86: +1 @estroz opinion too. Let's try to make that simple as possible first (KISS)

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
jmrodri: @estroz what's the "interface"? Is it basically the list of files and their contents that's being passed around?

estroz: Effectively, yes (and error information).

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+#### Approach #1
+
+User specified file paths based approach: User will provide a list of file paths for Kubebuilder to discover the plugins in. We will define a variable `KUBEBUILDER_PLUGINS_DIRS` that will take a list of file paths to search for the plugin name. It will also have a default value to search in, in case no file paths are provided. It will search for the plugin name that was provided to the `--plugins` flag in the CLI. Kubebuilder will recursively search for all file paths until the plugin name is found and returns the successful match, and if it doesn’t exist, it returns an error message that the plugin is not found in the provided file paths.
+
+* Pros
+  * This provides flexibility for the user to specify the file paths that the plugin would be placed in and Kubebuilder could discover the binaries in those user specified file paths.
+
+  * No constraints on plugin binary naming or directory placements from the Kubebuilder side.
+
+  * Provides a default value for the plugin directory in case user wants to use that to drop their plugins.
```
varshaprasad96, line 64: Could this also be like, when the user provides the `PATH` with a separator, we would look at its location, if not look for the binary in systems's `$PATH` and the last resort would be to figure out the fully qualified path from pwd. Helm kustomize plugin does something like that: https://helm.sh/docs/topics/advanced/#post-rendering

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+  * The one-plugin-per-directory requirement eases creation of a plugin bundle for sharing.
+
+* Cons
+  * This makes it mandatory for every plugin that is specified in the `layout` field to have additional configuration information such as Group, Version and Kind in the `plugins` field in the PROJECT file for it to be discoverable. Currently, the `plugins` field is more like an optional configuration storage field.
+
+#### Approach #3
+
+Another approach I would like to mention is adding plugin executables with a prefix `kubebuilder-` followed by the plugin name to the PATH variable. This will enable KB to traverse through the PATH looking for the plugin executables starting with the prefix `kubebuilder-` and matching by the plugin name that was provided in the CLI. Furthermore, a check should be added to verify that the match is an executable or not and return an error if it's not an executable. This approach provides a lot of flexibility in terms of plugin discovery as all the user needs to do is to add the plugin executable to the PATH and Kubebuilder will discover it.
```
varshaprasad96, line 95: A doubt - wouldn't this be more error prone in the sense, I can have two versions of same plugin provided in kb command. Since we don't have a way to figure this out based on a file name, we would end up calling the same plugin twice. For example, in the previous approach we can error out immediately when the cli has two plugins in a command with same `group`, `kind`. Eventually the plugin execution will error anyway, but just a scenario where we could have additional validation before allowing the plugins to execute.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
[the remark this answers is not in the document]
camilamacedo86: I think we should not care about this now.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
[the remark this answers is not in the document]
camilamacedo86: +1

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
camilamacedo86: That seems the best I think too. 
I'd recommend we keep this one and add the others in an alternative section just to keep the history. (by following the templated suggested above)

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,203 @@
...
+
+This is how [kustomize](https://kubectl.docs.kubernetes.io/guides/extending_kustomize/) discovers plugins.
+The PROJECT file in Kubebuilder has `plugins` field and each plugin will have the resources: Group, Kind and Version, so every plugin defined in the PROJECT file configuration can get its own GVK directory for the executable to be placed in and Kubebuilder will search for a plugin with the name of the plugin in the GVK directory of the plugin. Once KB successfully locates the plugin, it will run the plugin using the CLI.
+
+Following `kustomize`'s way of placement, every plugin gets its own directory such as:
+
+```project
+    $XDG_CONFIG_HOME/kubebuilder/plugins
+    /${group}/${version}/LOWERCASE(${kind}) 
+```
```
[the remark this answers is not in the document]
camilamacedo86: +1 I think here is missing an example of its configuration and usage.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
rashmigottipati: I used helm plugin because this EP is about external plugins that can be written in any language. But I like the idea of using the declarative plugin example and mentioning about externalizing it with Phase 2 implementation. Fixed that.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
rashmigottipati: Will be switching to this template soon once there are more implementation details. Agree that it would be easier to read.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
rashmigottipati: My understanding was that we do need to discover flags. However, my initial idea was to use project configuration for that where users can put in supported flags for each plugin, which is probably not ideal. 

@estroz mentioned that we don't necessarily have to any flag discovery and we could just pass the whole command line and pass it directly to the plugin binary.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
rashmigottipati: Agreed. Changed it to user stories instead.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
rashmigottipati: Sounds good to me as well. I kept this one in the main section and moved other approaches to Alternatives section.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
rashmigottipati: Yes, this will be out of scope for this EP. Will revisit this later.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
rashmigottipati: Hashicorp's go plugin allows has cross language support. I prefer to leave this intact for now.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
rashmigottipati: Removed this statement altogether as flags shouldn't be a part of the configuration.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
rashmigottipati: Good point @varshaprasad96. I agree that the GVK path scheme would be better to handle these scenarios. Hence, I moved Approach 2 and 3 to alternatives section. I repurposed the first approach be a `${pluginname}/${pluginversion}` scheme instead of GVK exactly.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
rashmigottipati: I've made a note of this in the EP, in case we want to switch to this approach in the future.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,300 @@
...
+1. Send `PluginRequest` as a JSON that holds `Command` as `init`, `Args` containing `--domain` and `example.com`, and an initialized empty universe over stdin to `myexternalplugin` plugin
+
+1. `myexternalplugin` plugin parses the `PluginRequest` and based on the `Command` specified in the request i.e `init`, performs the necessary scaffolding
+
+1. `myexternalplugin` plugin constructs `PluginResponse` with modified `Universe` that contains the updated file contents and errors if any
+
+1. Plugin writes `PluginResponse` to stdout in a JSON format back to Kubebuilder
+
+1. Kubebuilder receives the command output containing the `PluginResponse` JSON which will then be decoded into the `PluginResponse` struct
+
```
estroz, line 253:
```suggestion

1. `kubebuilder` writes the files in the universe to disk.

```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,300 @@
...
+### Alternatives
```
estroz, line 254:
```suggestion
## Alternatives
```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,300 @@
...
+
+There are other alternative approaches to discover the plugin binary from Kubebuilder and below are some ideas.
+
+#### Approach #2
+
+User specified file paths based approach: User will provide a list of file paths for Kubebuilder to discover the plugins in. We will define a variable `KUBEBUILDER_PLUGINS_DIRS` that will take a list of file paths to search for the plugin name. It will also have a default value to search in, in case no file paths are provided. It will search for the plugin name that was provided to the `--plugins` flag in the CLI. Kubebuilder will recursively search for all file paths until the plugin name is found and returns the successful match, and if it doesn’t exist, it returns an error message that the plugin is not found in the provided file paths. Also use the host system mechanism for PATH separation.
```
estroz, line 260: Make plugin discovery its own alternatives section
```suggestion
### Plugin discovery

#### User specified file paths

A user will provide a list of file paths for Kubebuilder to discover the plugins in. We will define a variable `KUBEBUILDER_PLUGINS_DIRS` that will take a list of file paths to search for the plugin name. It will also have a default value to search in, in case no file paths are provided. It will search for the plugin name that was provided to the `--plugins` flag in the CLI. Kubebuilder will recursively search for all file paths until the plugin name is found and returns the successful match, and if it doesn’t exist, it returns an error message that the plugin is not found in the provided file paths. Also use the host system mechanism for PATH separation.
```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,300 @@
...
+Alternatively, this could be handled in a way that [helm kustomize plugin](https://helm.sh/docs/topics/advanced/#post-rendering) discovers the plugin based on the non-existence of a separator in the path provided, in which case Kubebuilder will search in `$PATH`, otherwise resolve any relative paths to a fully qualified path.
+
+* Pros
+  * This provides flexibility for the user to specify the file paths that the plugin would be placed in and Kubebuilder could discover the binaries in those user specified file paths.
+
+  * No constraints on plugin binary naming or directory placements from the Kubebuilder side.
+
+  * Provides a default value for the plugin directory in case user wants to use that to drop their plugins.
+
+#### Approach #3
```
estroz, line 270:
```suggestion
#### Prefixed plugin executable names in $PATH
```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,300 @@
...
+  * `kubectl` and `git` follow the same approach for discovering plugins, so there’s prior art.
+
+  * There’s a lot of flexibility in just dropping plugin binaries to PATH variable and enabling the discovery without having to enforce any other constraints on the placements of the plugins.
+
+* Cons
+  * Enumerating the list of all available plugins might be a bit tough compared to having a single folder with the list of available plugins and having to enumerate those.
+
+  * These plugin binaries cannot be run in a standalone manner outside of Kubebuilder, so may not be very ideal to add them to the PATH var.
+
+#### Open questions
```
estroz, line 284:
```suggestion
## Open questions
```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,300 @@
...
+
+* Do we want to support the addition of new arbitrary subcommands other than the subcommands (init, create api, create webhook) that we already support?
+  * Not for the EP or initial implementation, but can revisit later.
+
+* Do we need to discover flags by calling the plugin binary or should we have users define them in the project configuration?
+  * Flags will have to be discovered by calling the plugin binary.
```
estroz, line 290:
```suggestion
  * Flags will be passed directly to the external plugin as a string. Flag parse errors will be passed back via `PluginResponse`.
```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,300 @@
...
+
+* What alternatives to stdin/stdout exist and why shouldn't we use them?
+  * Other alternatives exist such as named pipe and sockets, but stdin/stdout seems to be more suitable for our needs.
+
+* What happens when two plugins bind the same flag name? Will there be any conflicts?
+
```
estroz, line 296:
```suggestion
  * Flags are passed directly as a string to plugins. The same string will be passed to each plugin in the chain, so all plugins get the same flag set. Errors should not be returned if an unrecognized flag is parsed.

```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,300 @@
...
+* Should plugins depend on environment variables?
+  * The same plugin would behave differently based on the environment.
+
+* Should the API version be a part of the plugin request spec?
```
estroz, line 300: Which API version? For `PluginRequest` and `PluginResponse`? It would be nice to encode these actually, so we can mark the initial schemas as alpha (`v1alpha1`).

estroz (changes requested):
Lots of cleanup needed.

Another issue has come to mind too: what happens when an external plugin wants to run a generator on some files it added to the universe? It cannot write those files itself since the universe wouldn't reflect the file's actual state, and it cannot run the generator if the files don't exist on-disk. I think some sort of post-plugin hook (executable) could be added to the `PluginResponse` struct so `kubebuilder` can execute it after writing the universe to disk. This hook executable would be in the same directory as the external plugin itself, but the plugin executable name would be different (specified in the `PluginResponse`). Thoughts?

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,300 @@
+# Extensible CLI and Scaffolding Plugins - Phase 2
+
+## Overview
+
+* Plugin [Phase 1.5](thttps://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-1-5.md) was designed and implemented to allow chaining of plugins.
+
+The purpose of Phase 2 plugins is to discover and use external plugins, also referred to as out-of-tree plugins (which can be implemented in any language). So Phase 2 achieves both chaining of plugins and discovery of external plugins/source code which is not intended to be shipped with the CLI binary. By achieving this goal, we could externalize the optional [declarative](https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/golang/declarative/v1) which means that the CLI would still be able to use it, however, its source code would no longer be required to be inside of the Kubebuilder repository.
```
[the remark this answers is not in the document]
jmrodri: What Eric suggested above in addition add the word plugin after declarative - "declarative plugin, which means ..."

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,300 @@
...
+  * For example, once the Phase 2 plugin implementation is completed, some internal plugins can be re-implemented as external plugins removing the necessity to build those plugins in the Kubebuilder binary.
+
+### Goals
+
+* Kubebuilder is able to discover plugin binaries and run those plugins using the CLI.
+
+* Kubebuilder can use the external plugins as its own internal ones to do scaffolding.
+
+* Kubebuilder is able to discover flags that each individual plugin would support.
+
```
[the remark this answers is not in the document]
jmrodri: @estroz how will you get the flags for the help from plugins? Today that happens when we call BindFlags on the internal ones. But let's say the java-plugin adds a flag to `init` how will that be displayed to the user?

jmrodri (commented):
@rashmigottipati quick nit for the overview. Apply @estroz comments and I will re-review again. I agree with eric's additions but it's hard to reconcile until they have been applied.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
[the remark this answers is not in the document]
estroz: Pass `--help` to the external plugin, then print what it returns. I think we'll have to add an extra field `Help string` to `PluginResponse` since help isn't an error @rashmigottipati.

rashmigottipati:
@estroz What do you think about adding a `GeneratorPath` field in the `PluginResponse` struct that specifies the path to the executable that Kubebuilder will need to execute post `Run`? Not all plugins would need to run generators on files, so if a plugin does not set this field, specifying the `omitempty` option will exempt that field from encoding if it has an empty value.

rashmigottipati:
After having an offline discussion with @estroz, we decided to implement the post plugin hooks in follow-ups.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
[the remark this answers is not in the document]
rashmigottipati: Makes sense! Added `Help` field to the `PluginResponse` struct.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,300 @@
...
+This is similar to how [kustomize](https://kubectl.docs.kubernetes.io/guides/extending_kustomize/) discovers plugins by following a GVK path scheme.
+
+Every plugin gets its own directory constructed using the plugin name and plugin version for the executable to be placed in and Kubebuilder will search for a plugin binary with the name of the plugin in the `${name}/${version}` directory of the plugin. This information (plugin name and plugin version) is obtained to Kubebuilder via the value passed to the `plugins` field through the CLI. Once KB successfully locates the plugin, it will run the plugin using the CLI.
+
+Every plugin gets its own directory such as:
+
+```project
+    $XDG_CONFIG_HOME/kubebuilder/plugins
+    /${name}/${version}
```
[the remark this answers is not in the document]
rashmigottipati: I had to specify a language as it's a fenced code block and leaving it empty was causing markdown lint errors. Switched to `shell` instead of `project` as it may fit better.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,300 @@
...
+repo: github.com/test-inc/testproject
+resources:
+- group: crew
+  kind: Captain
+  version: v1
+```
+
+### What to pass between Kubebuilder and plugins
+
+Currently, the `injector` type is defined as below:
```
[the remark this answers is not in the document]
rashmigottipati: Yes, this is the structure used for in-tree plugins right now. I left it intact as I wasn't sure if we would end up using `Injector` type as well, along with the `PluginRequest` and `PluginResponse` but I got clarity now. Removed the previous section and added more details on the request/response spec.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,326 @@
...
+  APIVersion    string                   `json:"apiVersion"`
+
+  // Universe in the PluginResponse represents the updated file contents that was written by the plugin.
+  Universe      map[string]string        `json:"universe"`
+
+  // Error is a boolean type that indicates whether there were any errors due to plugin failures.
+  Error         bool                     `json:"error,omitempty"`
+
+  // ErrorMsg holds the specific error message of plugin failures.
+  ErrorMsg      string                   `json:"error_msg,omitempty"`
```
jmrodri, line 219: Is it just one error? Would we ever want to allow a series of errors to come back?

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,326 @@
...
+}
+```
+
+These structs would be defined on the Kubebuilder side. `PluginRequest` holds all the information `kubebuilder` receives from the CLI and the request will be marshaled into a JSON and sent over `stdin`. `PluginResponse` is what the plugin constructs with the updated universe and sent back to `kubebuilder`.
```
jmrodri, line 223: Adapting this paragraph would fit nicely where I asked for an intro paragraph above.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,326 @@
...
+  }
+
+  if json.Unmarshal(out, &res); err != nil {
+    return res, err
+  }
+  return res, nil
+}
+```
+
+The above function handles construction of the `PluginRequest` based on the information `kubebuilder` receives from the CLI and the request is marshaled into JSON. The command to run the external plugin by providing the plugin path will be invoked and `kubebuilder` will send the marshaled `PluginRequest` JSON to the plugin over `stdin`.
```
jmrodri, line 250: Move this above the `runExternalPlugin` code and switch it to "The following function handles ..." It's easier to read IMO to have a slight intro before code dumps.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,326 @@
...
+
+On the plugin side, the request JSON will be decoded and depending on what the `Command` in the `PluginRequest` is, the corresponding function to handle `init` or `create api` will be invoked thereby modifying the universe by writing the updated files to it. After `init` or `create api` functions execute successfully, the plugin will write back `PluginResponse` with updated universe and errors (if any) in JSON format through `stdout` to `kubebuilder`. `PluginResponse` also contains error fields `Error` and `ErrorMsg` that the plugin can utilize to add error context if any errors occur.
```
jmrodri, line 252: Do we have a JSON schema or sample to show what it will look like to the plugins? If I'm writing a plugin in Python I would need to know what the JSON will look like so I can properly handle it on my side.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,326 @@
...
+
+1. `myexternalplugin` plugin parses the `PluginRequest` and based on the `Command` specified in the request i.e `init`, performs the necessary scaffolding.
+
+1. `myexternalplugin` plugin constructs `PluginResponse` with modified `Universe` that contains the updated file contents and errors if any.
+
+1. Plugin writes `PluginResponse` to stdout in a JSON format back to `kubebuilder`.
+
+1. `kubebuilder` receives the command output containing the `PluginResponse` JSON which will then be decoded into the `PluginResponse` struct.
+
+1. `kubebuilder` writes the files in the universe to disk.
```
jmrodri, line 275: This might be better as a sequence diagram. https://sequencediagram.org/

```
title This is a title

kubebuilder->kubebuilder:discover myexternalplugin/v1
kubebuilder->myexternalplugin/v1:PluginRequest{Command:"init", Args:"--domain example.com"}
myexternalplugin/v1->myexternalplugin/v1:Parse PluginRequest
myexternalplugin/v1->-myexternalplugin/v1:Perform scaffolding
myexternalplugin/v1->-myexternalplugin/v1:Construct PluginResponse
myexternalplugin/v1->kubebuilder:PluginResponse{...Universe:map["foo.go"]...
kubebuilder->filesystem:writes files
```

Something like the above generates the following image:
image

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,326 @@
...
+
+* Cons
+  * Enumerating the list of all available plugins might be a bit tough compared to having a single folder with the list of available plugins and having to enumerate those.
+
+  * These plugin binaries cannot be run in a standalone manner outside of Kubebuilder, so may not be very ideal to add them to the PATH var.
+
+## Open questions
+
+* Do we want to support the addition of new arbitrary subcommands other than the subcommands (init, create api, create webhook) that we already support?
+  * Not for the EP or initial implementation, but can revisit later.
```
jmrodri, line 311: We should not code something that would prevent this though. It should be kept in the back of our mind.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,326 @@
...
+
+* Do we need to discover flags by calling the plugin binary or should we have users define them in the project configuration?
+  * Flags will be passed directly to the external plugin as a string. Flag parse errors will be passed back via `PluginResponse`.
+
+* What alternatives to stdin/stdout exist and why shouldn't we use them?
+  * Other alternatives exist such as named pipe and sockets, but stdin/stdout seems to be more suitable for our needs.
+
+* What happens when two plugins bind the same flag name? Will there be any conflicts?
+  * Flags are passed directly as a string to plugins. The same string will be passed to each plugin in the chain, so all plugins get the same flag set. Errors should not be returned if an unrecognized flag is parsed.
```
jmrodri, line 320: Are we documenting that to plugin writers?

jmrodri (commented):
A couple things and suggestions.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,326 @@
...
+
+Currently, the project configuration has two fields to store plugin specific information.
+
+* `Layout` field (of type []string) is used for plugin chain resolution on initialized projects. This will be the default if no plugins are specified for a subcommand.
+* `Plugins` field (of type map[string]interface{}) is used for option plugin configuration that stores configuration information of any plugin.
+
+* So, where should external plugins be defined in the configuration?
+
+  * I propose that the external plugin should get encoded in the project configuration as a part of the `layout` field.
+    * For e.g., external plugin `myexternalplugin/v2` can be specified through the `--plugins` flag for every subcommand and also be defined in the project configuration in the `layout` field for plugin resolution.
```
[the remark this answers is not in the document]
rashmigottipati: Yes, the external plugin can either be specified in the CLI with the `--plugins` field or the user can manually add it to the `layout` field in the PROJECT config. 

https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/PROJECT#L3 - Currently, in-tree plugins are specified that way too.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,326 @@
...
+  version: v1
+```
+
+### Communication between `kubebuilder` and external plugins
+
+* Why do we need communication between `kubebuilder` and external plugins?
+
+  * The in-tree plugins do not need any inter-process communication as they are the same process, and hence, direct calls are made to the respective functions (also referred as hooks) based on the supported subcommand for an in-tree plugin. As Phase 2 plugins is tackling out-of-tree or external plugins, there's a need for inter-process communication between `kubebuilder` and the external plugin binaries, and as discussed earlier, standard streams seems to be a desirable IPC method of communication for the use-cases that Phase 2 is trying to solve that involves discovery and chaining of external plugins.
+  
+  * Standard streams have three I/O connections: standard input (`stdin`), standard output (`stdout`) and standard error (`stderr`) and they work well with chaining applications, meaning that output stream of one program can be redirected to the input stream of another.
```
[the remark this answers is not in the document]
rashmigottipati: Moved it to "how does kubebuilder communicate to plugins" section.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
rashmigottipati: That's a great idea! Added a sequence diagram to describe the flow of events.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
rashmigottipati: I added example JSON for both `PluginRequest and `PluginResponse` in the Implementation details section. Let me know if that would suffice.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
rashmigottipati: I mentioned this in the Implementation Details. Is there some other place to add this that would be more meaningful?

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,326 @@
...
+  // APIVersion defines the versioned schema of the PluginRequest that is encoded and sent from Kubebuilder to plugin.
+  // Initially, this will be marked as alpha (v1alpha1).
+  APIVersion    string              `json:"apiVersion"`
+
+  // Args holds the plugin specific arguments that are received from the CLI which are to be passed down to the plugin.
+  Args          []string            `json:"args"`
+
+  // Universe represents the modified file contents that gets updated over a series of plugin runs
+  // across the plugin chain. Initially, it starts out as empty.
+  Universe      map[string]string   `json:"universe"`
```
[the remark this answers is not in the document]
rashmigottipati: Added the sample JSON in the implementation details section. 

Yes, a plugin would want to write binary files in the `PluginResponse` struct, basically the post-plugin hooks that kubebuilder will run after it writes all files to disk. But this is out of scope for this EP and will be done as a follow-up implementation.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,326 @@
...
+  // Help contains the plugin specific help text that the plugin returns to kubebuilder when it receives
+  // `--help` flag from Kubebuilder.
+  Help          string                   `json:"help,omitempty"`
+
+  // APIVersion defines the versioned schema of the PluginResponse that will be written back to kubebuilder.
+  // Initially, this will be marked as alpha (v1alpha1).
+  APIVersion    string                   `json:"apiVersion"`
+
+  // Universe in the PluginResponse represents the updated file contents that was written by the plugin.
+  Universe      map[string]string        `json:"universe"`
```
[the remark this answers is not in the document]
rashmigottipati: Same answer as above.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
rashmigottipati: No objection to making this `errList []error`. `errInfo map[string][]error` might be better where key is the plugin name. @estroz what do you think?

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
rashmigottipati: I guess the sequence diagrams aren't rendering properly when I try to view this file in browser. It works fine locally though in my editor. I'll probably generate these separately and attach them as images. Will fix this.

rashmigottipati: This is what I see locally.

sequence-diagram

rashmigottipati: Fixed this issue. They are rendering fine now.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
jmrodri: I just missed it. Thanks.

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
jmrodri: @rashmigottipati yes, that is good enough for me.

jmrodri (approved):
/lgtm

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
```diff
@@ -0,0 +1,356 @@
...
+
+### Prototype implementation
+
+[POC](https://github.com/rashmigottipati/POC-Phase2-Plugins) - Invoke an external python program that simulates a plugin from a go main and pass messages from `kubebuilder` to the plugin and vice-versa using `stdin/stdout/stderr`.
+
+### User Stories
+
+* As a plugin developer, I would like to be able to provide external plugins path for the CLI to perform the scaffolds, so that I could take advantage of external initiatives which are implemented using Kubebuilder as a lib and following its standards but are not shipped with its CLI binaries.
+
+* As a Kubebuilder/Operator SDK maintainer, I would like to support external plugins not maintained by the core project.
```
estroz, line 22:
```suggestion
* As a Kubebuilder maintainer, I would like to support external plugins not maintained by the core project.
```

designs/extensible-cli-and-scaffolding-plugins-phase-2.md
estroz: Fine with me!

estroz (commented):
One small fix, otherwise lgtm! Will label after you push the change.

estroz (approved):
/lgtm
/approve

PR #2249, opened 2021-07-02
📖 Make markers more easily copyable

### LINKED ISSUE #2242: Marker values are not easily copyable

Looking at docs at https://book.kubebuilder.io/reference/markers/crd-validation.html it's difficult to copy the full marker contents, including the `// +` prefix, because of how the CSS is laid out, using a `::before` directive:

https://github.com/kubernetes-sigs/kubebuilder/blob/52a8c1272ed928c332dd714a70fdfd5d841591e3/docs/book/theme/css/markers.css#L18-L21

The HTML generated by markerdocs seems to be:

```html
<dt class="literal name">kubebuilder:validation:Optional</dt>
```

The `name` CSS class is augmented with the `// +` using the `::before` directive above, but this makes the "before" content un-copyable. 😢 

If instead it was:

```
<dt class="literal name">// +kubebuilder:validation:Optional</dt>
```

(without the CSS above)

...then the text would be more easily copyable.

If this was an explicit decision, that's fine, but if there's any appetite for fixing this I'd be willing to give it a try. 🤞 

### Extra Labels

/kind documentation

imjasonh:
Fixes #2242

See https://github.com/kubernetes-sigs/kubebuilder/issues/2242#issuecomment-870854192

cc @camilamacedo86

### DIFF
--- a/docs/book/theme/css/markers.css
+++ b/docs/book/theme/css/markers.css
@@ -15,13 +15,8 @@
     margin-bottom: 0.25em;
 }
 
-/* the marker name */
-.marker > dt.name::before {
-    content: '// +';
-}
 .marker > dt.name {
     font-weight: bold;
-    order: 0; /* hack around the ::before's positioning to get it after the line */
 }
 
 /* the target blob */
--- a/docs/book/utils/markerdocs/main.go
+++ b/docs/book/utils/markerdocs/main.go
@@ -58,7 +58,7 @@ func markerTemplate(marker *MarkerDoc) toHTML {
 
 	// the marker name
 	term := dt(classes{"literal", "name"},
-		Text(marker.Name))
+		Text("// +"+marker.Name))
 
 	// the args summary (displayed in summary mode)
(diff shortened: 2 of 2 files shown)


imjasonh:
/retest 

These look like unrelated flakes:

```
ERROR: failed to create cluster: failed to generate kubeadm config content: failed to get kubernetes version from node: failed to get file: command "docker exec --privileged kind-worker3 cat /kind/version" failed with error: exit status 1
```

```
• Failure [253.900 seconds]
kubebuilder
/home/prow/go/src/sigs.k8s.io/kubebuilder/test/e2e/v2/plugin_cluster_test.go:33
  with project version 2 scaffolding
  /home/prow/go/src/sigs.k8s.io/kubebuilder/test/e2e/v2/plugin_cluster_test.go:34
    should generate a runnable project [It]
skipped 91 lines unfold_more
[Fail] kubebuilder with project version 2 scaffolding [It] should generate a runnable project 
```

Issue #2123, opened 2021-04-02
Add target to deploy a namespace-scoped operator.


fjammes:
Add a Makefile target (i.e. `make ns-deploy`) and related `kustomize` configuration files to deploy a namespace-scoped operator. This target would automate the feature documented here: https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#changing-the-permissions-to-namespaced.

I have implemented a prototype for it and I can open a PR if needed.

It does not require a particular kubernetes version.

/kind feature

camilamacedo86:
We think that it is a common need so please feel free to push a PR with your idea to solve that. However, we are not sure if it would be better addressed via a makefile target. Let's check your proposal and go forward from there. WDYT?

Adirio:
Small note: the most that can be solved with kustomize patches the better.

fjammes:
Hi @camilamacedo86, here is my proposal "https://github.com/kubernetes-sigs/kubebuilder/pull/2254"
The current idea is to run command below in the operator repository:
```
# generate <OPERATOR_REPO_DIR>/manifests/operator-ns-scoped.yaml
make yaml-ns-scoped
# And then set namespace and deploy manually
sed 's/<NAMESPACE>/my-namespace' manifests/operator-ns-scoped.yaml | kubectl apply -f -
```
I agree this is not perfect but this first step works fine on my side.

camilamacedo86:
See: https://github.com/kubernetes-sigs/kubebuilder/pull/2254#pullrequestreview-769487971

camilamacedo86:
@TAM360 I think you are not working anymore on this one so I am removing the assign
However, if you wish to do it please feel free to re-sign to yourself.

PR #2254, opened 2021-07-06
Add proposal for namespace-scoped deployment

fjammes:
Proposal for operator namespace-scoped deployment. related to #2123

### DIFF
--- a/testdata/project-v2-addon/Makefile
+++ b/testdata/project-v2-addon/Makefile
@@ -4,6 +4,8 @@ IMG ?= controller:latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
+ROOT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
+
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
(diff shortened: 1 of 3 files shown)


camilamacedo86 (changes requested):
Hi @fjammes, 

Thank you for your contribution 🥇 . 

This PR is only changing the testdata. 
Note that the changes need to be performed on the plugins and then you will use the `make generate` to update the testdata. 

Also, the idea here to change the scope for operators be build as namespace scope instead of cluster one seems very nice. 

However, these are not the only changes that would be required to get done. Maybe look the doc https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/ can help you out check it. 

By the latest, we would need to check the better option to address this need. 
It can be a plugin (https://book.kubebuilder.io/plugins/plugins.html) that changes the default scaffold for namespace scope or an option for the kb edit command as we have for multigroup. See; https://book.kubebuilder.io/migration/multi-group.html

I would vote in this RFE be a new plugin that can be used to change the scope of the project. However, it would also need to change the RBAC and the files scaffold. See that we would also need to change the markers of the files already scaffolded as you point out:  https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#changing-the-permissions-to-namespaced. Am I right? 

And then, see that the PR is missing the correct mojo and description as well. More info: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md

Address this need bring some complexities.

PR #2255, opened 2021-07-06
✨ fix(scaffolds): better defaults

longkai:
- explicitly specify port protocol where missing
- larger resource limit for controller manager

With the original default manifest, the server-side-apply would fail since the required `protocol` field is missing.

This is actually a k8s [bug](https://github.com/kubernetes-sigs/structured-merge-diff/issues/130) and should not be omitted before 1.20.

The default controller manager resource limit is too low for downloading small amount (~30kb) of http resource. It results a OOMKilled Pod status and no logs could be found.

Sometimes you got the CrashLoopBackOff status but no logs found, which took me sometime to figure it out.

We can watch the status transition via `kubectl get po -w`:

```sh
bootstrap-controller-manager-8f88c7b5b-gplpg   2/2     Running            4          3m59s
bootstrap-controller-manager-8f88c7b5b-gplpg   1/2     OOMKilled          4          4m5s
bootstrap-controller-manager-8f88c7b5b-gplpg   1/2     CrashLoopBackOff   4          4m10s
bootstrap-controller-manager-8f88c7b5b-gplpg   1/2     Running            5          5m27s
bootstrap-controller-manager-8f88c7b5b-gplpg   2/2     Running            5          5m39s
bootstrap-controller-manager-8f88c7b5b-gplpg   1/2     OOMKilled          5          6m4s
bootstrap-controller-manager-8f88c7b5b-gplpg   1/2     CrashLoopBackOff   5          6m9s
```

I suppose this default improves the user experience especially for beginners.

### DIFF
--- a/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/manager/config.go
+++ b/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/manager/config.go
@@ -95,8 +95,8 @@ spec:
           periodSeconds: 10
         resources:
           limits:
-            cpu: 100m
-            memory: 30Mi
+            cpu: 200m
+            memory: 100Mi
           requests:
             cpu: 100m
             memory: 20Mi
--- a/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/kdefault/manager_auth_proxy_patch.go
+++ b/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/kdefault/manager_auth_proxy_patch.go
@@ -63,6 +63,7 @@ spec:
         - "--v=10"
         ports:
         - containerPort: 8443
+          protocol: TCP
           name: https
 {{- if not .ComponentConfig }}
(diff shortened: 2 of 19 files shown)


pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/manager/config.go
```diff
@@ -95,8 +95,8 @@ spec:
           periodSeconds: 10
         resources:
           limits:
-            cpu: 100m
-            memory: 30Mi
+            cpu: 200m
+            memory: 200Mi
```
estroz, line 99: 200Mi seems like a high limit. Thoughts on 100?
```suggestion
            memory: 100Mi
```

longkai: Seems reasonable and I have forced push the changes.

estroz (approved):
/lgtm
/approve

PR #2282, opened 2021-08-04
📖 improved error log message for creating a controller

hellt:
Is it better to distinct the creation phase of a controller with the running one?
Starting the controller happens later on with its own step, so it seems more appropriate to use the word `create` instead of `start`

### DIFF
--- a/pkg/plugins/golang/v3/scaffolds/internal/templates/main.go
+++ b/pkg/plugins/golang/v3/scaffolds/internal/templates/main.go
@@ -260,7 +260,7 @@ func main() {
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), options)
 {{- end }}
 	if err != nil {
-		setupLog.Error(err, "unable to start manager")
+		setupLog.Error(err, "unable to create manager")
 		os.Exit(1)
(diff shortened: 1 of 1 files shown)


hellt:
signed CLA

camilamacedo86 (changes requested):
Hi, @hellt, 

Thank you for your contribution 🥇.

 I'd like to share that this PR would not pass the tests. **Why ?**
You change the scaffolds and did not re-generate the test data samples.   
Before pushing the changes you need to run `make generate` . More info: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md

Regards the change here, the manager is initialized on the cluster and I do not really see an advantage on this change. What is the benefit that kb users would have with it?

Then, I hope that you do not mind we close this PR as not accepted. 
Also, feel free to ping us to re-open and attend the meetings if you think that it needs to be re-open. 

Also, note that the repo has a lot of issues to work on and I'd recommend you looking for: https://github.com/kubernetes-sigs/kubebuilder/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22

PR #1056, opened 2019-09-30
Document marker syntax for maps

marun:
**Description of the change:**

Add documentation for the map syntax added for marker arguments by https://github.com/kubernetes-sigs/controller-tools/pull/329

PR #2288, opened 2021-08-10
:book: Use an existing marker for the struct value example.

porridge:
The `+kubebuilder:validation:Default` marker does not appear in the output of `controller-gen crd -www` so it was probably just an invented example.

Let's use an actual existing marker here to prevent the non-existing one from being cargo-culted into other projects any more.

### DIFF
--- a/docs/book/src/reference/markers.md
+++ b/docs/book/src/reference/markers.md
@@ -89,5 +89,5 @@ each key and value is separated by a colon (`:`), and each key-value
 pair is separated by a comma:
 
 ```go
-// +kubebuilder:validation:Default={magic: {numero: 42, stringified: forty-two}}
+// +kubebuilder:default={magic: {numero: 42, stringified: forty-two}}
 ```


porridge:
cc @marun to confirm my theory since he seems to have added this back in https://github.com/kubernetes-sigs/kubebuilder/pull/1056
cc @joelanford who helped establish this theory

camilamacedo86 (approved):
It was changed here: https://github.com/kubernetes-sigs/controller-tools/commit/4a548a715cff258881e56a84a85a722eecc94835

And I can find the change in the version v0.6.1 which is used by the master

https://github.com/kubernetes-sigs/controller-tools/blob/v0.6.1/pkg/crd/markers/validation.go#L84-L85

See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/scaffolds/init.go#L35

/lgtm

PR #2313, opened 2021-08-24
⚠️ feat(makefile): add configurable binaries

georgettica:
currently docker and kubectl are hardcoded (as they are industry standards)

but as a user I want to consume the makefile without modifying the upstream version.
adding flags as these allows me to just modify the binaries and not the code inside

Breaks the `docker-build` and `docker-push`


pkg/plugins/golang/v2/scaffolds/internal/templates/makefile.go
```diff
@@ -60,6 +60,11 @@ IMG ?= {{ .Image }}
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
+# run container engine commands (such as building and pushing)
+OCI_ENGINE ?= docker
+# run commands against a kubernetes cluster
+KUBECTL ?= kubectl
+
```
camilamacedo86, line 67: You are here to change the go/v2 plugin. 
This kind of change should be addressed to go/v3 which is the latest version.

pkg/plugins/golang/v2/scaffolds/internal/templates/makefile.go
```diff
@@ -60,6 +60,11 @@ IMG ?= {{ .Image }}
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
+# run container engine commands (such as building and pushing)
+OCI_ENGINE ?= docker
```
camilamacedo86, line 64:
```suggestion
CONTAINER-TOOL ?= docker
```

pkg/plugins/golang/v2/scaffolds/internal/templates/makefile.go
```diff
@@ -60,6 +60,11 @@ IMG ?= {{ .Image }}
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
+# run container engine commands (such as building and pushing)
```
camilamacedo86, line 63:
```suggestion
# used to run container engine commands (such as building and pushing the manager image)
```

pkg/plugins/golang/v2/scaffolds/internal/templates/makefile.go
```diff
@@ -60,6 +60,11 @@ IMG ?= {{ .Image }}
...
+OCI_ENGINE ?= docker
+# run commands against a kubernetes cluster
+KUBECTL ?= kubectl
```
camilamacedo86, line 66:
```suggestion
CLUSTER_CLIENT_TOOL ?= kubectl
```

pkg/plugins/golang/v2/scaffolds/internal/templates/makefile.go
```diff
@@ -60,6 +60,11 @@ IMG ?= {{ .Image }}
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
+# run container engine commands (such as building and pushing)
+OCI_ENGINE ?= docker
+# run commands against a kubernetes cluster
```
camilamacedo86, line 65:
```suggestion
# used to run commands against a Kubernetes cluster
```

pkg/plugins/golang/v2/scaffolds/internal/templates/makefile.go
```diff
@@ -113,26 +118,26 @@ manager: build ## Build manager binary (alias for build target).
...
+	$(KUSTOMIZE) build config/crd | $(KUBECTL) delete -f -
 
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | kubectl apply -f -
+	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
 
-undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/default | kubectl delete -f -
+undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
```
camilamacedo86, line 139: Why are you calling kustomize here?

camilamacedo86 (changes requested):
Hi, @georgettica,

Please, check my comments on this PR. Note that:
- It would not be a breaking change so the mojo seems not accurate at all.
- IHMO would be better addressed only for go/v3 and not go/v3. The go/v2 is the old layout and it is kept just to keep the backwards compatibility. In this way, we ought to address it more bug fixes only. 
- Could you also add it into the title go/v3 to make clear it in the release notes.
- Also, could you please move these changes for go/v3 plugin?

pkg/plugins/golang/v2/scaffolds/internal/templates/makefile.go
georgettica: Because undeploy uses kustomize?
So it's a dependency

georgettica:
I'll look I to this Sunday :)

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
```diff
@@ -128,13 +133,18 @@ build: generate fmt vet ## Build manager binary.
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
-.PHONY: docker-build
-docker-build: test ## Build docker image with the manager.
-	docker build -t ${IMG} .
+.PHONY: container-build
+container-build: test ## Build container image with the manager.
```
camilamacedo86, line 137: We cannot address breaking changes in the plugins.
To do that we would need to for e.g.:
- Create another plugin
- Update the docs 


So, what about we have the options by vars which helps users customize and change as they wish but does not change the names of the target? I would be ok with that.

camilamacedo86 (changes requested):
Really thank you for your contribution. 
Only a few nits. Otherwise, shows fine 🥇 
- https://github.com/kubernetes-sigs/kubebuilder/pull/2313/files#r768250125

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
georgettica: I can live with that. But could you ping me when a new plugin is planning on being released so I can add the new targets?

camilamacedo86: you can create an issue and we can link to : https://github.com/kubernetes-sigs/kubebuilder/milestone/26
for now, we do have not enough requirements that would justify that.

georgettica: as this is stale for a while, I might need a quick session on how you envision this and I'll do as you requested :)

and maybe I can answer my stuff with a question:
1. would you want me to break the change to be without the new targets for now, and change them in the new release
2. create an issue and add the targets in the new plugin

camilamacedo86: /hold 

I do not think that we should change the name of these targets now
It can be very disruptive and that would mean changing all docs and etc as well.

camilamacedo86: Also, with this PR would came all changes in the docs

camilamacedo86: I will create a new comment

camilamacedo86, line 137: /hold

I do not think that we should change the name of these targets now
It can be very disruptive and that would mean changing all docs and etc as well.
So, I would suggest you raise that in the KB + C+R meeting and then we can discuss it or try to make a pool in the channel.
Also, with this PR would come all changes in the docs

camilamacedo86 (changes requested):
My concerns with the proposed changes are mainly:

- It changes the docker-<option> targets which will be very disruptive for users
- That would also mean change all docs
- Also, if we say that people can use any container tool we would ensure that we are able to support them. So, that would mean doc clarifications about what is or is not supported and maybe some additional tests to ensure that all work with another container-tool 

Thinking that users are able to customize their makefile and use the tool that they see fit and they ensure that will work as well I am not sure if the changes and disruption can be justified with the value that these changes bring. 

If you would like to move forward with this one, I'd recommend you raise it in the channel OR in the meetings so that we can discuss this with the community.

georgettica:
/close
This is fairly complex to find the right time to merge this.

I lost context and maybe someone else would have more luck

PR #2316, opened 2021-08-28
:seedling: fix lint issues on master

camilamacedo86:
**Description**
Fix lint issues that began to raise on master.

Close: https://github.com/kubernetes-sigs/kubebuilder/issues/2307
Close: https://github.com/kubernetes-sigs/kubebuilder/issues/2301
Close: https://github.com/kubernetes-sigs/kubebuilder/issues/2294

### DIFF
--- a/pkg/machinery/scaffold.go
+++ b/pkg/machinery/scaffold.go
@@ -370,10 +370,7 @@ func filterExistingValues(content string, codeFragmentsMap CodeFragmentsMap) err
 			}
 		}
 	}
-	if err := scanner.Err(); err != nil {
-		return err
-	}
-	return nil
+	return scanner.Err()
 }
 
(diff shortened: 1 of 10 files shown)


justinsb:
Thanks for fixing @camilamacedo86 ... I was trying to verify locally by running `make lint`, but the linter doesn't report any errors for me.  Any hints as to what I might be doing wrong?

pkg/machinery/scaffold.go
```diff
@@ -370,10 +370,7 @@ func filterExistingValues(content string, codeFragmentsMap CodeFragmentsMap) err
 			}
 		}
 	}
-	if err := scanner.Err(); err != nil {
```
estroz, line 373: FWIW I do not like this linter because it leads to more diff lines when changing downstream execution within a function body in later changes. However this seems to be a source of several issues so I do not want to block this from merging.

estroz (approved):
/lgtm

PR #2318, opened 2021-08-29
:sparkles: Must include Dockerfile in declarative plugin

justinsb:
We need to copy the channels directory in the Dockerfile, particularly now that we're running with non-root permissions and it is no longer a one-liner.
    
Add Dockerfile generation.

### DIFF
--- a/pkg/plugins/golang/declarative/v1/init.go
+++ b/pkg/plugins/golang/declarative/v1/init.go
@@ -1,0 +1,53 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1
+
+import (
+	"fmt"
+
+	"sigs.k8s.io/kubebuilder/v3/pkg/config"
+	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/declarative/v1/internal/templates"
+)
+
+var _ plugin.InitSubcommand = &initSubcommand{}
+
+type initSubcommand struct {
+	config config.Config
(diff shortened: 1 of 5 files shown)


justinsb:
APIDiff looks to be a false-failure.

justinsb:
@camilamacedo86 should I try to work around the APIDIff incompatible changes on the Plugin?  I don't think this API is user-facing.

camilamacedo86:
Hi @justinsb, 

Could you please rebase it with the master and run make generate for we are able to check?
I think that the problem here is that this PR is not synced. 

Also, feel free to ping me in the slack when it passes it the tests for we are able to move forward.

camilamacedo86 (changes requested):
Hi @justinsb,

The contribution is great 🥇 Just a few nits. 

See that, the declarative plugin is used to scaffold apis. See: https://book.kubebuilder.io/plugins/plugins.html#plugins

Example:

```
kubebuider create api [options] --plugins=go/v3,declarative/v1
```

So, if it requires changes into the Dockerfile we need to ensure that these changes will be performed on the create API subcommand as well. 

Also, we need to add the scaffolds inside of the scaffold dir. 

pkg/plugins/golang/declarative/v1/init.go
```diff
@@ -0,0 +1,53 @@
...
+	config config.Config
+}
+
+func (p *initSubcommand) InjectConfig(c config.Config) error {
+	p.config = c
+
+	return nil
+}
+
+func (p *initSubcommand) Scaffold(fs machinery.Filesystem) error {
```
camilamacedo86, line 40: The scaffold would need to be in inside of the scaffold dir

pkg/plugins/golang/declarative/v1/plugin.go
```diff
@@ -49,6 +50,9 @@ func (Plugin) Version() plugin.Version { return pluginVersion }
 // SupportedProjectVersions returns an array with all project versions supported by the plugin
 func (Plugin) SupportedProjectVersions() []config.Version { return supportedProjectVersions }
 
+// GetInitSubcommand will return the subcommand which is responsible for initializing and common scaffolding
+func (p Plugin) GetInitSubcommand() plugin.InitSubcommand { return &p.initSubcommand }
+
 // GetCreateAPISubcommand will return the subcommand which is responsible for scaffolding apis
 func (p Plugin) GetCreateAPISubcommand() plugin.CreateAPISubcommand { return &p.createAPISubcommand }
```
camilamacedo86, line 57: The declarative plugin is called via `create api --plugin' so its implementation shows not accurate. 
Should we not be implementing it in the create API subcommand?

camilamacedo86:
We can close this one since we could address this changes in : https://github.com/kubernetes-sigs/kubebuilder/pull/2507
c/c @justinsb please feel free to check and let us know if it is missing something

PR #2329, opened 2021-09-07
:sparkles: Create klogr plugin that uses kubernetes logging (klog)

justinsb:
Create klogr plugin that uses kubernetes logging (klog)
    
This is often requested / recommended, because client-go only supports klog.  Without enabling klog, we cannot get diagnostic output from client-go.


pkg/plugins/golang/klogr/v1/plugin.go
```diff
@@ -0,0 +1,50 @@
...
+	"sigs.k8s.io/kubebuilder/v3/pkg/config"
+	cfgv3 "sigs.k8s.io/kubebuilder/v3/pkg/config/v3"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang"
+)
+
+const pluginName = "klogr." + golang.DefaultNameQualifier
+
+var (
+	pluginVersion            = plugin.Version{Number: 2}
```
justinsb, line 29: I wasn't sure whether this should reflect the klog version number...

camilamacedo86: No. That is the plugin version. See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/VERSIONING.md

pkg/plugins/golang/klogr/v1/init.go
```diff
@@ -0,0 +1,195 @@
...
+	"github.com/spf13/afero"
+	"sigs.k8s.io/kubebuilder/v3/pkg/config"
+	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
+)
+
+var _ plugin.InitSubcommand = &initSubcommand{}
+
+type initSubcommand struct {
+	config config.Config
```
camilamacedo86, line 39: You are implementing here the INIT interface which means that the changes applied here would only be called when the project is init ( kb init command )

In this case, the logs are scaffold when we create the API then, which means that this implementation would not work. More info: https://book.kubebuilder.io/plugins/plugins.html

Also, the plugins added to the project starts to be supported by it and we need to discuss if we would like or not to have this plugin by default. 

It is an important to highlight as well that the first version of the plugin should be alpha since it would not be mature enough.  See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/VERSIONING.md

camilamacedo86 (changes requested):
Hi @justinsb, 

Really thank you for your contribution 🥇  
I added some nits. Could you please check it out?
c/c @jmrodri @rashmigottipati

pkg/plugins/golang/klogr/v1/init.go
camilamacedo86: @justinsb is that the expected behaviour? would we only want this change when we run the init command with this plugin or should it work with creating API as well?

cmd/main.go
```diff
@@ -26,6 +26,7 @@ import (
 	kustomizecommonv1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/common/kustomize/v1"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang"
 	declarativev1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/declarative/v1"
+	klogrv1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/klogr/v1"
```
camilamacedo86, line 29: Could we start the plugin with an alpha version?

pkg/plugins/golang/klogr/v1/init.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (p *initSubcommand) InjectConfig(c config.Config) error {
+	p.config = c
+
+	return nil
+}
+
+func (p *initSubcommand) Scaffold(fs machinery.Filesystem) error {
```
camilamacedo86, line 48: we need to follow the same structure that we have for the others plugins.
The scaffolds go inside of the scaffold dir. Could you please check it out?

camilamacedo86:
Closing because is open for a while without interactions.
However, an issue was raised to keep track the PR and see if we have other contributors who would like to help we move forward here.

Issue #2331, opened 2021-09-09
Configuring EnvTest documentation suggets non-working binary installation directory.


Birdrock:
### What broke? What's expected?

On [Configuring EnvTest](https://book.kubebuilder.io/reference/envtest.html#installation), the documentation for installation specifies creating the directory `/opt/kubebuilder/testbin` and extracting the envtest binaries to that directory. It then suggests that the `test` target can be changed to (omitting prerequisites/dependencies) `go test ./... -coverprofile cover.out`. When the `test` target is run, the following error occurs:
```
 Unexpected error:
      <*fmt.wrapError | 0xc00048f500>: {
          msg: "failed to start the controlplane. retried 5 times: fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory",
          err: {
              Op: "fork/exec",
              Path: "/usr/local/kubebuilder/bin/etcd",
              Err: 0x2,
          },
      }
      failed to start the controlplane. retried 5 times: fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory
  occurred
```
The instructions place the binaries in a location that is not the expected default for `controller-runtime`. For a working installation, without providing a `KUBEBUILDER_ASSETS` environment variable, the target directory for the extraction should be `/usr/local/kubebuilder`.

### KubeBuilder (CLI) Version

3.1.0

### PROJECT version

3

### Plugin versions

```yaml
layout:
- go.kubebuilder.io/v3
```

### Other versions

Go version: go version go1.16.4 linux/amd64

controller-runtime: sigs.k8s.io/controller-runtime v0.8.3

controller-gen: sigs.k8s.io/controller-tools/cmd/controller-gen@v0.6.1

kubectl version:
```
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.1", GitCommit:"632ed300f2c34f6d6d15ca4cef3d3c7073412212", GitTreeState:"clean", BuildDate:"2021-08-19T15:45:37Z", GoVersion:"go1.16.7", Compiler:"gc", Platform:"linux/amd64"}
```

Not interacting with a Kubernetes cluster for this.

### Extra Labels

/kind documentation

camilamacedo86:
The ENV TEST was changed. So, we need to update the docs to clarifies that as add the new instructions.
See https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest#where-does-it-put-all-those-binaries.

[...]

Issue #2330, opened 2021-09-08
Broken link in the envtest reference documentation


timflannagan:
### What broke? What's expected?

The markdown link listed in the [Installation](https://book.kubebuilder.io/reference/envtest.html#installation) subheader is broken. It looks like that hack script for setting up envtest has since moved to the [root tools/ directory](https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest).

### Reproducing this issue

Follow the "downloads" link referenced in https://book.kubebuilder.io/reference/envtest.html#installation.

### KubeBuilder (CLI) Version

v3

Birdrock:
@timflannagan To add some more information to this, the `hack/setup-envtest.sh` script was removed in https://github.com/kubernetes-sigs/controller-runtime/commit/782807d8d4fb287e86f6c91862529574ac42cc0a. The replacement functionality is fetched with go modules.

The text certainly needs to be updated. I can take a swing at it if you are looking for someone to update it.

rashmigottipati:
@Birdrock Please feel free to update the documentation. 
Once the fix merges into master, we need to cherry-pick that into book-v3 branch.

Birdrock:
@rashmigottipati I opened PR #2332, but forgot to cross-link them. Will link it on the PR, as well.

Birdrock:
Referenced in discussion with #2331 regarding course forward if entire section should be rewritten for `setup-envtest` tool.

camilamacedo86:
The ENV TEST was changed. So, we need to update the docs to clarifies that as add the new instructions.
See https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest#where-does-it-put-all-those-binaries.

I task to track this need was tracked: https://github.com/kubernetes-sigs/kubebuilder/issues/2389

Closing that as duplicated.

PR #2332, opened 2021-09-09
:book: Update Configuring EnvTest binary directory docs

Birdrock:
The documentation for [Configuring EnvTest](https://book.kubebuilder.io/reference/envtest.html#installation) results in a non-working configuration unless a `KUBEBUILDER_ASSETS` environment variable is provided. This seems contrary to the documentation indication that the `test` target can be changed to the provided snippet.

This should address #2331.

### DIFF
--- a/docs/book/src/reference/envtest.md
+++ b/docs/book/src/reference/envtest.md
@@ -22,8 +22,8 @@ curl -sSLo envtest-bins.tar.gz "https://storage.googleapis.com/kubebuilder-tools
 Then install them:
 
 ```sh
-mkdir /opt/kubebuilder/testbin
-tar -C /opt/kubebuilder/testbin --strip-components=1 -zvxf envtest-bins.tar.gz
+mkdir /usr/local/kubebuilder
+tar -C /usr/local/kubebuilder --strip-components=1 -zvxf envtest-bins.tar.gz
 ```
 
 Once these binaries are installed, you can either change the `test` target to:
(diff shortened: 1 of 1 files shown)


docs/book/src/reference/envtest.md
```diff
@@ -36,7 +36,7 @@ test: manifests generate fmt vet
 Or configure the existing target to skip the download and point to a [custom location](#environment-variables):
 
 ```sh
-make test SKIP_FETCH_TOOLS=1 KUBEBUILDER_ASSETS=/opt/kubebuilder/testbin
+make test SKIP_FETCH_TOOLS=1 KUBEBUILDER_ASSETS=/usr/local/kubebuilder
```
varshaprasad96, line 39: I think the whole section on configuring the envtest binaries is outdated, since we now use the `setup-envtest` [tool](https://pkg.go.dev/sigs.k8s.io/controller-runtime/tools/setup-envtest). We should probably rewrite the entire section instead of making this modification to the path? Since with the recent changes we do not have a separate envtest script, and the location of the binaries can be specified with a separate flag passed to `setup-envtest` tool.

Birdrock: @varshaprasad96 @rashmigottipati The docs are a bit outdated. There is a related issue I'm working on, as well #2330.

`kubebuilder@v3.1.0` uses `controller-runtime@v0.8.3` which isn't yet updated to the `setup-envtest` tool. Should the docs reflect the most recent release or the most recent commit to the default branch? I'm happy to open PRs against specific branches to align them with the state of the world at that time, but I don't know that multiple versions of the `kubebuilder` book would be helpful.

Regardless, updating the docs to reflect the new tooling would be incorrect with the most recent release. If the book is deployed from the most recent release, then a change to a release branch in tandem with a change to the main branch would be appropriate.

camilamacedo86: Hi @Birdrock, 

Did you test run the make test command locally with your changes? if yes, could you add the result of the test in the description of this PR? That helps us to review it. 

Also, I agree with @varshaprasad96 we need to update this doc to add the new way to set up the env tests. The master branch should reflect the state of the master code and the changes in the doc will be in prod when the code source be releassed. 

WDYT? Could you work on that and help us with it?

PR #2338, opened 2021-09-15
✨Plugin phase 2 implementation

rashmigottipati:
* a description of the change
Adds implementation to discover and run external plugins in a chained manner. 

* the motivation for the change
[Phase 2 Design Proposal](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-2.md)


pkg/cli/options.go
```diff
@@ -139,3 +150,150 @@ func WithCompletion() Option {
...
+func parseExternalPluginArgs() (extPlugins []string, args []string) {
+	var externalPlugins string
+	args = make([]string, len(os.Args)-1)
+	copy(args, os.Args[1:])
+
+	// parse os.Args and pass everything to plugin except program name,
+	// --plugins flag and its value.
+	for i, item := range args {
+		if item == "--plugins" {
+			externalPlugins = args[i+1]
```
estroz, line 164: You can also write multiple plugins like `kubebuilder init --plugins foo/v1 --plugins bar/v2`. To avoid parsing edge cases like this, copy what's going on in [this method](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/cli.go#L235-L271) to extract all plugins from args.

pkg/cli/options.go
```diff
@@ -139,3 +150,150 @@ func WithCompletion() Option {
...
+			args = append(args[:i], args[i+2:]...)
+		}
+	}
+
+	extPlugins = strings.Split(externalPlugins, ".")
```
estroz, line 169: Why split by `.`?

pkg/cli/options.go
```diff
@@ -139,3 +150,150 @@ func WithCompletion() Option {
...
+
+	return extPlugins, args
+}
+
+// setPluginsRootBasedonHost detects the host system and sets the plugins root based on the host.
+func setPluginsRootBasedonHost() (pluginsRoot string) {
```
estroz, line 175: Nit: naming, and just return an error.
```suggestion
func getPluginsRoot() (pluginsRoot string, err error) {
```

pkg/cli/options.go
```diff
@@ -139,3 +150,150 @@ func WithCompletion() Option {
...
+			continue
+		}
+
+		subinfos, err := ioutil.ReadDir(filepath.Join(pluginsRoot, info.Name()))
+		if err != nil {
+			return nil, err
+		}
+
+		for _, subinfo := range subinfos {
+			pluginInfos, err := ioutil.ReadDir(filepath.Join(pluginsRoot, info.Name(), subinfo.Name()))
```
estroz, line 237: Need a `subinfo.IsDir()` check here too.

pkg/cli/options.go
```diff
@@ -139,3 +150,150 @@ func WithCompletion() Option {
...
+				}
+			}
+		}
+
+	}
+	return ps, nil
+}
+
+// isPluginExectuable checks if a plugin is an executable based on the bitmask and returns true or false.
+func isPluginExectuable(mode os.FileMode) bool {
```
estroz, line 284: I wouldn't try to check for whether a file is executable or not. Let `ExternalPlugin.Run()` handle this error, since it will be more informative.

pkg/plugins/external/api.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (p *createAPISubcommand) InjectResource(*resource.Resource) error {
+	// Do nothing since resource flags are passed to the external plugin directly.
+	return nil
+}
+
+func (p *createAPISubcommand) Scaffold(fs machinery.Filesystem) error {
+	req := external.PluginRequest{
+		APIVersion: defaultAPIVersion,
```
estroz, line 54: You can explicitly set command and args so the external plugin can just return the desired help messages. It is totally possible for an external plugin to not define args. Also just pass args unadulterated for parsing purposes.
```suggestion
		APIVersion: defaultAPIVersion,
		Command: 	"create api",
		Args:		p.Args,
```

pkg/plugins/external/api.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return err
+	}
+
+	cmd := exec.Command(p.Path)
+	cmd.Stdin = bytes.NewBuffer(reqBytes)
+	cmd.Stderr = os.Stderr
+	out, err := cmd.Output()
+	if err != nil {
+		log.Fatal(err)
```
estroz, line 82: Don't exit in library code.
```suggestion
```

pkg/plugins/external/api.go
```diff
@@ -0,0 +1,115 @@
...
+	if res.Error {
+		log.Fatal(res.ErrorMsgs)
+	}
+
+	output, err := json.MarshalIndent(res, "", "  ")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	fmt.Fprint(os.Stdout, "Universe:\n", string(output))
```
estroz, line 101: Is this debug code?
```suggestion
```

pkg/plugins/external/api.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	res := external.PluginResponse{}
+	if err := json.Unmarshal(out, &res); err != nil {
+		return err
+	}
+
+	// Error if the plugin failed.
+	if res.Error {
+		log.Fatal(res.ErrorMsgs)
```
estroz, line 93: Don't exit in library code.
```suggestion
		return fmt.Errorf(strings.Join(res.ErrorMsgs, "\n"))
```

pkg/plugins/external/api.go
```diff
@@ -0,0 +1,115 @@
...
+
+	output, err := json.MarshalIndent(res, "", "  ")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	fmt.Fprint(os.Stdout, "Universe:\n", string(output))
+
+	for filename, data := range res.Universe {
+		scaffoldPath := path.Dir(p.Path)
```
estroz, line 104: Isn't `p.Path` the plugin's path and not project path?

estroz (dismissed):

pkg/cli/options.go
jmrodri: @estroz why? Seems like we want to ensure that the file we deem is the "executable for the plugin" is correct when we read it. In my head since we are already scanning files to see which one needs to be associated with the plugin that we can deem whether it will even work down the line. Otherwise, it will get down to the Run step and fail. 

Maybe I don't know what the error will look like yet. I just feel like it would be better to fail early than later.

pkg/cli/options.go
```diff
@@ -139,3 +150,150 @@ func WithCompletion() Option {
...
+
+					ps = append(ps, ep)
+
+				}
+			}
+		}
+
+	}
+	return ps, nil
+}
```
jmrodri, line 281: Some of the var names confused me. I read the entire method, I get the gist of what you are doing only because I conceptually know what you are trying to write. But the variables seemed weird to me. Specifically `infos`, `info`, `subinfos`, `subinfo`. 

Basically if `pluginsRoot` is `$XDG_CONFIG_HOME/kubebuilder/plugins/` then `infos, err := ioutil.ReadDir(pluginsRoot)` will return the list of plugin directories. So I think I'd use `pluginInfos` or `plugins` I don't care that they are `FileInfo` I care more what they represent. 

Then this line `subinfos, err := ioutil.ReadDir(filepath.Join(pluginsRoot, info.Name()))` will basically be returning the versions. So I'd say instead of `subinfos` I'd use `versions`. 

Then for `pluginInfos` I'd use `pluginFiles` or `pluginFileInfo`. This will tell me I'm at the level where the plugin should be defined.

I think this would make it easier to come back and fix bugs. Let's say there's a problem loading files from the plugins. Well you know it will be in the `pluginFiles` section. Or there's a problem reading the versions layer, you'd know it would be in the `versions` section.

pkg/cli/options.go
```diff
@@ -139,3 +150,150 @@ func WithCompletion() Option {
...
+
+// isPluginExectuable checks if a plugin is an executable based on the bitmask and returns true or false.
+func isPluginExectuable(mode os.FileMode) bool {
+	return mode&0111 == 0111
+}
+
+// getUserHomeDir returns $XDG_CONFIG_HOME if set, otherwise $HOME.
+func getUserHomeDir() (string, error) {
```
jmrodri, line 289: nit: I think I would've just used `getHomeDir()`. No need to change this just adding here for posterity.

pkg/plugin/external/types.go
```diff
@@ -40,14 +40,14 @@ type PluginRequest struct {
...
-	APIVersion string `json:"apiVersion"`
+	APIVersion string `json:"apiVersion,omitempty"`
 
 	// Command holds the command that gets executed by the plugin such as init, create api, etc.
 	Command string `json:"command"`
 
 	// Help contains the plugin specific help text that the plugin returns to Kubebuilder when it receives
 	// `--help` flag from Kubebuilder.
-	Help string `json:"help"`
+	Help string `json:"help,omitempty"`
```
jmrodri, line 50: I assume these are optional, and the change was intentional?

pkg/plugins/external/api.go
```diff
@@ -0,0 +1,115 @@
...
+func (p *createAPISubcommand) Scaffold(fs machinery.Filesystem) error {
+	req := external.PluginRequest{
+		APIVersion: defaultAPIVersion,
+	}
+
+	req.Universe = map[string]string{}
+
+	if req.Command == "" {
+		if len(p.Args) == 0 {
+			return fmt.Errorf("internal error: invalid length of arguments for %q subcommand", defaultCreateAPISubcommand)
```
jmrodri, line 61: Seems weird to have `internal error:` in the message. 
```suggestion
			return fmt.Errorf("invalid length of arguments for %q subcommand", defaultCreateAPISubcommand)
```

pkg/plugins/external/api.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+		req.Command = fmt.Sprintf("%s %s", p.Args[0], p.Args[1])
+		if req.Command != defaultCreateAPISubcommand {
+			return fmt.Errorf("internal error: subcommand %s in args doesn't match what's expected: %s", req.Command, defaultCreateAPISubcommand)
```
jmrodri, line 65:
```suggestion
			return fmt.Errorf("subcommand %s in args doesn't match what's expected: %s", req.Command, defaultCreateAPISubcommand)
```

pkg/plugins/external/edit.go
```diff
@@ -0,0 +1,108 @@
...
+func (p *editSubcommand) Scaffold(fs machinery.Filesystem) error {
+	req := external.PluginRequest{
+		APIVersion: defaultAPIVersion,
+	}
+
+	req.Universe = map[string]string{}
+
+	if req.Command == "" {
+		if len(p.Args) == 0 {
+			return fmt.Errorf("internal error: invalid length of arguments for %q subcommand", defaultEditSubcommand)
```
jmrodri, line 54:
```suggestion
			return fmt.Errorf("invalid length of arguments for %q subcommand", defaultEditSubcommand)
```

pkg/plugins/external/edit.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		if p.Args[0] != defaultEditSubcommand {
+			return fmt.Errorf("internal error: subcommand %s in args doesn't match what's expected: %s", p.Args[0], defaultEditSubcommand)
```
jmrodri, line 57:
```suggestion
			return fmt.Errorf("subcommand %s in args doesn't match what's expected: %s", p.Args[0], defaultEditSubcommand)
```

pkg/plugins/external/init.go
```diff
@@ -0,0 +1,108 @@
...
+func (p *initSubcommand) Scaffold(fs machinery.Filesystem) error {
+	req := external.PluginRequest{
+		APIVersion: defaultAPIVersion,
+	}
+
+	req.Universe = map[string]string{}
+
+	if req.Command == "" {
+		if len(p.Args) == 0 {
+			return fmt.Errorf("internal error: invalid length of arguments for %q subcommand", defaultInitSubcommand)
```
jmrodri, line 54:
```suggestion
			return fmt.Errorf("invalid length of arguments for %q subcommand", defaultInitSubcommand)
```

pkg/plugins/external/webhook.go
```diff
@@ -0,0 +1,114 @@
...
+func (p *createWebhookSubcommand) Scaffold(fs machinery.Filesystem) error {
+	req := external.PluginRequest{
+		APIVersion: defaultAPIVersion,
+	}
+
+	req.Universe = map[string]string{}
+
+	if req.Command == "" {
+		if len(p.Args) == 0 {
+			return fmt.Errorf("internal error: invalid length of arguments for %q subcommand", defaultCreateWebhookSubcommand)
```
jmrodri, line 60:
```suggestion
			return fmt.Errorf("invalid length of arguments for %q subcommand", defaultCreateWebhookSubcommand)
```

pkg/plugins/external/webhook.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+		req.Command = fmt.Sprintf("%s %s", p.Args[0], p.Args[1])
+		if req.Command != defaultCreateWebhookSubcommand {
+			return fmt.Errorf("internal error: subcommand %s in args doesn't match what's expected: %s", req.Command, defaultCreateWebhookSubcommand)
```
jmrodri, line 64:
```suggestion
			return fmt.Errorf("subcommand %s in args doesn't match what's expected: %s", req.Command, defaultCreateWebhookSubcommand)
```

jmrodri (changes requested):
First round pass is done. I'll take another look after you've address Eric's and my comments.

pkg/cli/options.go
rashmigottipati: We are actually setting the pluginsRoot based on the host. But I am good with changing it to `getPluginsRoot()`.

pkg/plugins/external/api.go
rashmigottipati: Yup, intended for debug purposes. Removed it.

pkg/plugins/external/api.go
rashmigottipati: You're right. I wasn't sure where the files should be scaffolded. Changed it to project path by using current working directory.

pkg/plugin/external/types.go
rashmigottipati: Yes, help text is optional so added `omitempty` option.

LICENSE
```diff
@@ -1,4 +1,4 @@
-                                 Apache License
+				Apache License
```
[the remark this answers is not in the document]
rashmigottipati: Checked out the original LICENSE file from master, so it won't show up in the diff now.

pkg/cli/options.go
rashmigottipati: Agreed, it looks better this way and easier for debugging as well. Fixed it as per your recommendation.

pkg/plugins/external/api.go
rashmigottipati: Yea, that makes sense. Fixed it to explicitly set command and args, and passing unadulterated args to the plugin now.

pkg/cli/options.go
rashmigottipati: That was kind of my hunch too. To fail early during discovery itself and not even add it to the list of plugins. 

Will check if `ExternalPlugin.Run()` handles this check.

pkg/cli/options.go
rashmigottipati: Initially intended to trim the string before the dot and compare it to the plugin executable found, but didn't end up needing `extPlugins`. I am returning it in this function but didn't end up using this [here](https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#diff-1851a71300b8a592080399ec5bc32857e772b473fbc69b5fcec115870207d2f0R266).

pkg/cli/options.go
estroz: If discovery skips a plugin file because it's not executable, then the CLI will report that the plugin is not found, which is confusing imo. Instead, just treat the plugin as found so the CLI can match passed `--plugins` values, then fail during execution with a more informative error.

pkg/plugins/external/api.go
```diff
@@ -0,0 +1,97 @@
...
+	if res.Error {
+		return fmt.Errorf(strings.Join(res.ErrorMsgs, "\n"))
+	}
+
+	currentDir, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("error getting current directory: %v", err)
+	}
+	for filename, data := range res.Universe {
+		f, err := fs.FS.Create(filepath.Join(currentDir, filename))
```
estroz, line 87: Don't forget to close these files.

pkg/cli/options.go
```diff
@@ -139,3 +150,146 @@ func WithCompletion() Option {
...
+			return "", fmt.Errorf("error: %v", err)
+		}
+
+		pluginsRoot = filepath.Join(userHomeDir, ".config", "kubebuilder", "plugins")
+
+	default:
+		// freebsd, openbsd, windows...
+		return "", fmt.Errorf("Host not supported: %v", host)
+	}
+
```
estroz, line 187: Nit: slight cleanup
```suggestion
	switch host := runtime.GOOS; host {
	case "darwin":
		logrus.Debugf("Detected host is macOS.")
		pluginsRoot = filepath.Join("Library", "ApplicationSupport", "kubebuilder", "plugins")
	case "linux":
		logrus.Debugf("Detected host is Linux.")
		pluginsRoot = filepath.Join(".config", "kubebuilder", "plugins")
	default:
		// freebsd, openbsd, windows...
		return "", fmt.Errorf("Host not supported: %v", host)
	}

	userHomeDir, err := getHomeDir()
	if err != nil {
		return "", fmt.Errorf("error retrieving home dir: %v", err)
	}
	pluginsRoot = filepath.Join(userHomeDir, pluginsRoot)
```

pkg/cli/options.go
rashmigottipati: Didn't end up using the parsed plugins list so cleaned up that code. Let me know if I still have to follow the same approach from the above link to parse the plugin keys from args. As we are passing unadulterated args to the plugin, figured there's no need to filter out `plugins` flag from args.

pkg/plugins/external/api.go
rashmigottipati: Done.

pkg/cli/options.go
rashmigottipati: Just tested on a random file that matches the same name as the plugin name passed in the CLI but that's not an executable. Either way, it would fail, but have to figure out what error messages are more informative. 

Without the check seems more informative as it shows the entire plugin path. Any thoughts @estroz @jmrodri ? 

With the check:
▶ ../bin/kubebuilder init --plugins random
2021/09/24 13:30:48 External plugin random found in path is not an executable

Without the check: 
 ▶ ../bin/kubebuilder init --plugins random
INFO[0000] Adding external plugin: random
Error: failed to initialize project: unable to scaffold with "random/v1": fork/exec /Users/rashmigottipati/Library/ApplicationSupport/kubebuilder/plugins/random/v1/random: permission denied
Usage:
  kubebuilder init [flags]

Examples:
  # Help for initializing a project with version "2"
  kubebuilder init --project-version="2" -h

  # Help for initializing a project with version "3"
  kubebuilder init --project-version="3" -h

Flags:

Global Flags:

2021/09/24 13:31:07 failed to initialize project: unable to scaffold with "random/v1": fork/exec /Users/rashmigottipati/Library/ApplicationSupport/kubebuilder/plugins/random/v1/random: permission denied

estroz: I think the first one is better actually. The check needs to be improved though. Comments below.

pkg/cli/options.go
```diff
@@ -139,3 +150,139 @@ func WithCompletion() Option {
...
+			}
+		}
+
+	}
+	return ps, nil
+}
+
+// isPluginExectuable checks if a plugin is an executable based on the bitmask and returns true or false.
+func isPluginExectuable(mode os.FileMode) bool {
+	return mode&0111 == 0111
```
estroz, line 274:
```suggestion
	return mode&0111 != 0
```

estroz (commented):
This PR is pretty close to done. I'd like to see a bunch of unit tests and 1 e2e test using the POC code prior to merging.

@camilamacedo86 I'll be on PTO next week so if you could review/approve/lgtm this PR that'd be awesome!

pkg/cli/options.go
rashmigottipati: Sounds good.

pkg/cli/options.go
rashmigottipati: Fixed this.

rashmigottipati:
@estroz thanks a lot for the review. Will work on adding some unit tests and 1 e2e test and update the PR when done.

pkg/cli/options.go
```diff
@@ -139,3 +151,152 @@ func WithCompletion() Option {
...
+
+	return args
+}
+
+// getPluginsRoot detects the host system and gets the plugins root based on the host.
+func getPluginsRoot() (pluginsRoot string, err error) {
+	switch host := runtime.GOOS; host {
+	case "darwin":
+		logrus.Debugf("Detected host is macOS.")
+		pluginsRoot = filepath.Join("Library", "ApplicationSupport", "kubebuilder", "plugins")
```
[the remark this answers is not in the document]
jmrodri: @camilamacedo86 this was called out in the EP https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-2.md#discovery-of-plugin-binaries. We're explicitly telling you this is where they need to be installed.

pkg/cli/options.go
```diff
@@ -139,3 +151,152 @@ func WithCompletion() Option {
...
+	return plugins, nil
+}
+
+// isPluginExectuable checks if a plugin is an executable based on the bitmask and returns true or false.
+func isPluginExectuable(mode fs.FileMode) bool {
+	return mode&0111 != 0
+}
+
+// getHomeDir returns $XDG_CONFIG_HOME if set, otherwise $HOME.
+func getHomeDir() (string, error) {
```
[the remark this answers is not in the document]
jmrodri: @camilamacedo86 https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-2.md#discovery-of-plugin-binaries

cmd/main.go
```diff
@@ -48,6 +54,7 @@ func main() {
 			&kustomizecommonv1.Plugin{},
 			&declarativev1.Plugin{},
 		),
+		cli.WithPlugins(externalPlugins...),
```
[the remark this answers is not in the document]
jmrodri: @camilamacedo86 no, externalPlugins ARE Plugins. Basically all external plugins can be Plugins, but not all Plugins can be external plugins. So externalPlugins can be treated as if they were Plugins from the eyes of kubebuilder.

pkg/cli/options_test.go
```diff
@@ -17,16 +17,279 @@ limitations under the License.
...
 	"sigs.k8s.io/kubebuilder/v3/pkg/config"
+	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
 	"sigs.k8s.io/kubebuilder/v3/pkg/model/stage"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
 )
 
+// func TestExternalPluginsDiscovery(t *testing.T) {
+// 	RegisterFailHandler(Fail)
+// 	RunSpecs(t, "TestExternalPluginsDiscovery")
+// }
```
camilamacedo86, line 40: @rashmigottipati why it is commented? Are this test been called?

estroz: Suite setup for cli is already handled [here](https://github.com/kubernetes-sigs/kubebuilder/blob/e8f2ae4537b3bb996018c1f7b3ef18db33885118/pkg/cli/suite_test.go#L29-L32) so this can be removed.

pkg/plugins/external/helpers.go
```diff
@@ -0,0 +1,118 @@
...
+func (o *osWdGetter) GetCurrentDir() (string, error) {
+	currentDir, err := os.Getwd()
+	if err != nil {
+		return "", fmt.Errorf("error getting current directory: %v", err)
+	}
+
+	return currentDir, nil
+}
+
+func getPluginResponse(fs machinery.Filesystem, req external.PluginRequest, path string) error {
```
estroz, line 71: Nit: better naming
```suggestion
func handlePluginResponse(fs machinery.Filesystem, req external.PluginRequest, path string) error {
```

pkg/cli/options_test.go
rashmigottipati: Yes exactly, that's the reason I commented it out. Will remove it. Was using it mainly for running the tests locally. 
Also, defining this here was causing a Gingko error too (it's a known issue for gingko/gomega) - will try to find the link for it.

pkg/plugins/external/helpers.go
```diff
@@ -0,0 +1,118 @@
...
+
+	reqBytes, err := json.Marshal(req)
+	if err != nil {
+		return err
+	}
+
+	out, err := outputGetter.GetExecOutput(reqBytes, path)
+	if err != nil {
+		return err
+	}
```
estroz, line 82: I would run all of this in the Scaffold() method of each plugin so you don't have to make a global `exec` mock.

pkg/plugins/external/helpers.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// Error if the plugin failed.
+	if res.Error {
+		return fmt.Errorf(strings.Join(res.ErrorMsgs, "\n"))
+	}
+
+	currentDir, err := currentDirGetter.GetCurrentDir()
+	if err != nil {
+		return fmt.Errorf("error getting current directory: %v", err)
+	}
```
estroz, line 97: Just pass the current wd as an argument to this function.

rashmigottipati: @estroz do you mean that I could just parameterize `GetCurrentDir` by passing current wd as an argument, and in the unit test I could just mock that value, instead of creating an [interface](https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#diff-d40306bd214f2bcd22890c26bd1d40a49b0dfba323b99814c8fcf93d85543d0dR56) for it?

pkg/plugins/external/helpers.go
rashmigottipati: This is currently being run in Scaffold() method itself. [getPluginResponse](https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#diff-992e7a26b8776b12932c030601857ed56f364b1ecd7dc0fac9dcfd8c0cb903c6R39) in `Scaffold()` calls the GetExecOutput() method.

How can I avoid creating the global `exec` mock?

cmd/main.go
```diff
@@ -19,6 +19,7 @@ package main
 import (
 	"log"
 
+	"github.com/sirupsen/logrus"
```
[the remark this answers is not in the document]
rashmigottipati: Agree with this suggestion, will modify accordingly.

pkg/plugins/external/helpers.go
estroz: Yep!

pkg/plugins/external/helpers.go
estroz:
```go
func (p *createAPISubcommand) Scaffold(fs machinery.Filesystem) error {
	req := external.PluginRequest{
		APIVersion: defaultAPIVersion,
		Command:    "create api",
		Args:       p.Args,
	}

	req.Universe = map[string]string{}
	reqBytes, err := json.Marshal(req)
	if err != nil {
		return err
	}

	out, err := external.GetExecOutput(reqBytes, p.Path)
	if err != nil {
		return err
	}

	wd, err := os.Getwd()
	if err != nil {
		return err
	}

	return handlePluginResponse(fs, out, wd)
```

You could also refactor the duplicated setup code into its own function.

rashmigottipati: Thanks for the snippet @estroz. I tried this refactor and it works fine but in the unit test, I am unable to mock the `os.Getwd()` call as it's part of Scaffold now. I can mock `handlePluginResponse` by passing in test value for wd. But in the unit tests, I'm directly calling `Scaffold()` method using mock interfaces for `GetExecOutput` and `GetCurrentDir`. Any other way to have a mock value for `os.Getwd()` call?

camilamacedo86 (commented):
Hi @rashmigottipati, 

Following what still missing be addresseed from the review: 

- a) Could you please supplement the comment to clarify what is expected get done in the todo here? What is the acceptance criteria to remove this todo? ; https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#r737212666
- b) Remove the commented code in: https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#r771596382
- c) About changing the API to allow the plugin version (APIVersion) and the plugin Help to be empty, see: https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#r737215035 and https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#r737215379. Can we revert that and ensure that any plugin will still have the version and help? 
- d) Should we not use the API version of the plugin? (https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#r737217533) Why we are seeing it fixed? Can we change it?
- e) Address Nit: https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#r771598265
- f) Address the request in: https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#r771603022

However, for we completed Plugin phase 2 still missing:

- a) docs; we need to update the [Plugin Section](https://book.kubebuilder.io/plugins/plugins.html)  
- b) we need to provide an example external plugin for users (it could be under the testdata dir) and could be a plugin that just scaffolds a test file or output hello world
- c) we need to add an e2e test calling the sample plugin and checking its output. 

Btw, how can I test your PR to ensure that it works? 
How can we verify it? 

Could you please address these ones? Also, feel free to reach out if you need help.

pkg/plugin/external/types.go
```diff
@@ -40,14 +40,14 @@ type PluginRequest struct {
 type PluginResponse struct {
 	// APIVersion defines the versioned schema of the PluginResponse that is back sent back to Kubebuilder.
 	// Initially, this will be marked as alpha (v1alpha1)
-	APIVersion string `json:"apiVersion"`
+	APIVersion string `json:"apiVersion,omitempty"`
```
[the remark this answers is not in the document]
rashmigottipati: Reverted this.

cmd/main.go
```diff
@@ -19,9 +19,12 @@ package main
 import (
 	"log"
```
camilamacedo86, line 20: Following what still missing be addresseed from the review: 

- a) Could you please supplement the comment to clarify what is expected get done in the todo here? What is the acceptance criteria to remove this todo? ; https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#r737212666
- b) Remove the commented code in: https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#r771596382
- c) About changing the API to allow the plugin version (APIVersion) and the plugin Help to be empty, see: https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#r737215035 and https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#r737215379. Can we revert that and ensure that any plugin will still have the version and help? 
- d) Should we not use the API version of the plugin? (https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#r737217533) Why we are seeing it fixed? Can we change it?
- e) Address Nit: https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#r771598265
- f) Address the request in: https://github.com/kubernetes-sigs/kubebuilder/pull/2338/files#r771603022

However, for we completed Plugin phase 2 still missing:

- a) docs; we need to update the [Plugin Section](https://book.kubebuilder.io/plugins/plugins.html)  
- b) we need to provide an example external plugin for users (it could be under the testdata dir) and could be a plugin that just scaffolds a test file or output hello world
- c) we need to add an e2e test calling the sample plugin and check its output. 

Btw, how can I test your PR to ensure that it works? 
How can we verify it? 

Could you please address these ones? Also, feel free to reach out if you need help.

PR #2341, opened 2021-09-17
✨ Add flag for ignore-not-found to make undeploy

perdasilva:
### Description

This PR adds the possibility to pass --ignore-not-found to underlying `kubectl delete` command  that gets executed by `make undeploy`. With this change, you can do `make undeploy ignore-not-found=true`. This way, if you have any errors in your kustomize your `make undeploy` or `make uninstall` won't error out when it stumbles on the missing resource.

### Motivation

This is in response to the following issue from the operator-sdk project: https://github.com/operator-framework/operator-sdk/issues/5195. Once incorporated downstream, it should address the same issue, at least for go scaffolds.

### Notes

I'm not sure this is the best approach, or name for the environment variable. Happy to get some feedback there and make adjustments as necessary ^^


testdata/project-v2-addon/Makefile
```diff
@@ -75,9 +75,11 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
-undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/default | kubectl delete -f -
 
+KB_UNDEPLOY_IGNORE_NOT_FOUND ?= 0
+undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
+	$(KUSTOMIZE) build config/default | kubectl delete --ignore-not-found=$(KB_UNDEPLOY_IGNORE_NOT_FOUND) -f -
```
camilamacedo86, line 81: What do you think about instead create an ENV VAR here we do something like:
```
action:
    echo argument is $(argument)
You would then call it make action argument=something
```
Also, would be nice add the info that args can be informed at the end in the comment `## Undeploy controller from the K8s cluster specified in ~/.kube/config.`. Why? That is the description of the target when we run make --help. 

Also, it would allow people to pass any arg instead of an specific one and would remove the need to use and doc env vars.

camilamacedo86: /ok-to-test

perdasilva: I'm still not sure how to deal with defaults in this case, though. What do you think about something like:

```Makefile
ifndef ignore-not-found
  ignore-not-found = no
endif

undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
    $(KUSTOMIZE) build config/default | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
```

perdasilva: @camilamacedo86 I've updated the code with my suggestion and also spread it to the _uninstall_ target. Seemed to make sense.

camilamacedo86 (approved):
/approved

perdasilva:
@camilamacedo86 @joelanford could I get a /lgtm plz =)

camilamacedo86:
@perdasilva 
Could you please rebase with master and run make generate ? 
It will allow us to merge this one.

perdasilva:
> @perdasilva Could you please rebase with master and run make generate ? It will allow us to merge this one.

@camilamacedo86 done ^^

PR #2344, opened 2021-09-20
🐛 Fix Certificate template to have isCA: true

### LINKED ISSUE #2342: Webhooks - Certificate spec requires isCA: true

When setting up mutating or validating webhooks, kubebuilder generates invalid Certificate yaml.

According to cert-manager docs for self-signed certificates, you must set `isCA: true` on the Certificate spec. However, the examples and code generation in kubebuilder are missing this field. https://cert-manager.io/docs/configuration/selfsigned/

I would consider this a high-impact bug, since we had a mutating webhook in our cluster suddenly fail after a self-signed Certificate renewal, with the error "x509: invalid signature: parent certificate cannot sign this kind of certificate". This affected our ability to create pods in the cluster. Updating the Certificate spec to `isCA: true` resolved the incident.

### Reproducing this issue

_No response_

### KubeBuilder (CLI) Version

3.1.0

### PROJECT version

_No response_

### Plugin versions

_No response_

### Other versions

_No response_

### Extra Labels

_No response_

armsnyder:
~Self-signed certificates must have `isCA: true`, or else the kube-apiserver may encounter TLS handshake errors connecting to the webhook server.~

EDIT: I've made this PR a draft while I investigate this further. Input/advice is appreciated.

This change adds the field to the plugin templates.

Fixes #2342 

- [x] I did a global find for `issuerRef` in order to make sure I added the `isCA` field everywhere in the book as well.
- [x] I verified that the `isCA` field exists on the v1alpha2 Certificate version used by old plugin versions.

### DIFF
--- a/docs/book/src/component-config-tutorial/testdata/project/config/certmanager/certificate.yaml
+++ b/docs/book/src/component-config-tutorial/testdata/project/config/certmanager/certificate.yaml
@@ -19,6 +19,7 @@ spec:
   dnsNames:
   - $(SERVICE_NAME).$(SERVICE_NAMESPACE).svc
   - $(SERVICE_NAME).$(SERVICE_NAMESPACE).svc.cluster.local
+  isCA: true
   issuerRef:
     kind: Issuer
     name: selfsigned-issuer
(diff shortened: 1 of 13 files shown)


armsnyder:
Not needed

PR #2350, opened 2021-09-26
✨Include base.go.kubebuilder.io/v3 plugin in CLI

### LINKED ISSUE #2349: Include base.go.kubebuilder.io plugin

On my team we bundle our operator config using Helm instead of Kustomize. It would be useful if we could use kubebuilder for golang code generation only.

However, the `go.kubebuilder.io` plugin in the `kustomize` CLI extends the `kustomize.common.kubebuilder.io` plugin, so it generates config that we do not need.

Can we get the `base.go.kubebuilder.io` plugin into the `kustomize` CLI please?

### Extra Labels

_No response_

armsnyder:
Registers the existing `base.go.kubebuilder.io/v3` plugin so that it is usable in the `kubebuilder` CLI, for users who do not use Kustomize.

Updated supported plugin table:

```
                        Plugin keys | Supported project versions
------------------------------------+----------------------------
          base.go.kubebuilder.io/v3 |                          3
   declarative.go.kubebuilder.io/v1 |                       2, 3
               go.kubebuilder.io/v2 |                       2, 3
               go.kubebuilder.io/v3 |                          3
 kustomize.common.kubebuilder.io/v1 |                          3
```

Closes #2349 

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -43,6 +43,7 @@ func main() {
 		cli.WithVersion(versionString()),
 		cli.WithPlugins(
 			golangv2.Plugin{},
+			golangv3.Plugin{},
 			gov3Bundle,
 			&kustomizecommonv1.Plugin{},
 			&declarativev1.Plugin{},


rashmigottipati:
/ok-to-test

Seems fine to me. 
@camilamacedo86 @estroz what do you think?

camilamacedo86 (approved):
So, the change here would allow scaffold only the base for go/v3 instead of the full plugin. 
I am ok with that. 

However, for you to use KB to create your own plugins I'd suggest you use it as lib and not the CLI. See: https://github.com/kubernetes-sigs/kubebuilder/issues/2349#issuecomment-932522228

PR #2354, opened 2021-09-27
🐛 Fix wrong plugin key on some OS

armsnyder:
Fixes a bug where if a CLI is compiled on an OS that does not use "/" as a path separator, the `cli.WithPlugins` CLI option would generate plugin keys that cannot be parsed.

Other helpers assume "/" is the plugin key separator, for example:

https://github.com/kubernetes-sigs/kubebuilder/blob/a89ce41d8da6f0e86f681fba9475c1875b14e5a3/pkg/plugin/helpers.go#L35-L41

### DIFF
--- a/pkg/plugin/helpers.go
+++ b/pkg/plugin/helpers.go
@@ -18,7 +18,6 @@ package plugin
 
 import (
 	"fmt"
-	"path"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


armsnyder:
I'm dumb, path.Join uses slashes. I was thinking filepath.

Issue #2337, opened 2021-09-15
Tutorial Section 1.2 Missing Colon


marshmallory:
https://book.kubebuilder.io/cronjob-tutorial/empty-main.html

`    mgr, err = ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
        Scheme:                 scheme,
        Namespace:              namespace,
        MetricsBindAddress:     metricsAddr,
        Port:                   9443,
        HealthProbeBindAddress: probeAddr,
        LeaderElection:         enableLeaderElection,
        LeaderElectionID:       "80807133.tutorial.kubebuilder.io",
    })`

The colon after err is missing.

This is my first time working in this space. Happy to submit a PR if the documentation is managed.

stevemcquaid:
#2358 & #2359 fixed this
/close

PR #2358, opened 2021-10-01
:book: Fix Tutorial Section 1.2 Missing Colon

stevemcquaid:
This is a simple fix to add the color after `err` so the short assignment operator (`:=`) is used. 
 
Fixes: #2337


camilamacedo86 (approved):
/lgtm

PR #2359, opened 2021-10-01
:book: Fix Tutorial Section 1.2 Missing Colon (2/2)

stevemcquaid:
This is a simple fix to add another color after err so the short assignment operator (:=) is used.

Fixes: #2337

In the context of the emptymain.go the colon was intentionally left out, but in the context of the tutorial the color was necessary as a replacement for the cluster-scoped and namespaced-scoped manager instantiations


camilamacedo86 (approved):
/approved

PR #2366, opened 2021-10-03
:sparkles: ensure that the whole code has the same nomenclature TODO(user)

camilamacedo86:
**Description**
ensure that the whole code has the same nomenclature TODO(user)

**Motivation**
Make it easier for users to grep/looking for the values.


estroz (approved):
/lgtm

PR #2365, opened 2021-10-03
:sparkles: Add todo(user) to make clear the need to optimize the values of the manager resources according to the needs

camilamacedo86:
**Description**
Add todo(user) to make clear the need to optimize the values of the manager resources according to the needs

Related to: https://github.com/kubernetes-sigs/kubebuilder/issues/2347

Issue #2347, opened 2021-09-23
Resource requests and limits


fgiloux:
### What do you want to happen?

Kubebuilder uses these default values for generating controller deployments:

```
        resources:
          limits:
            cpu: 200m
            memory: 100Mi
          requests:
            cpu: 100m
            memory: 20Mi
```
They have been improved with: PR #2255

Although these values can be changed by operator authors I would like to discuss and potentially revisit the defaults as:
- not every author would be aware of the need to adapt these values to the specifics of their operator
- these values may be seen as guidance when the operator author has not done any load testing to come to values that are more suitable for its operator
- Although I think that PR #2255 was an improvement I am not sure that these values are the best defaults

My current thinking is that
- 100m for cpu request seems high. From personal experience, well behaving busy operators rarely need 100m. I would rather see something around ~10m as a default. When node capacity has been fully requested it prevents additional pods requesting CPU to be scheduled on the same node, which may lead to under-utilization of the cluster resources.
- For capacity planning there is little reason to set a CPU limit (it is compressible and extra cycles on a node are allocated proportionally to the request value). That said I would be fine with keeping it at 200m to help with mitigating risks of overloading the API, although there are other mechanisms for the purpose (maxInflightRequest, API Priority and Fairness).
- On the other hand memory request and limit seem still a bit low to me. I see many operators commonly using ~100m. Unlike CPU memory is not easily compressible. I would tend to increase these values to request: 80m and limit 120m

I am missing the background on how the existing values became the defaults and I would appreciate if others would share their experience and view on this.

One other aspect is that configuring resource requests and limits is not mentioned in the kubebuilder documentation if I have not missed it. It may be beneficial to draw operator author's attention to this subject.

If this makes sense to you I would be happy to create a pull request for new defaults and documentation.

### Extra Labels

/kind documentation

fgiloux:
Any feedback?

camilamacedo86:
Hi @fgiloux,

Following my IHMO on that:
> not every author would be aware of the need to adapt these values to the specifics of their operator

It seems that should be addressed via doc. 
I'd recommend we add it into: https://sdk.operatorframework.io/docs/best-practices/common-recommendation/ to make clear that one of the good practices would be calibrate this values according to the operator needs. 

Also, I think we can scaffold in the config/manager.yml file a comment with the link for k8s doc and the explain its usage and the importance to calibrate the values something as: 

```yaml
[...]

PR #2368, opened 2021-10-04
:sparkles: increase resources limits to be less restrictive and match the values with k8s doc examples

camilamacedo86:
**Description**
 increase resources limits to be less restrictive and match the values with k8s doc examples https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/

NOTE the PR will allow making clear that this value requires to be customized: https://github.com/kubernetes-sigs/kubebuilder/pull/2365

**Motivation**
Avoid users face problems with restrictive values when they are trying to skill up and create their projects. 
Matches with the docs references/examples since the values scaffold by default are just added as an example and good practice. Note that have them already reduces the risks and optimizations should be up to the authors and their specific cases/scenarios.


fgiloux:
As written in https://github.com/kubernetes-sigs/kubebuilder/issues/2347 the CPU request should be significantly lower. There is no benefit in having such a high value for well-behaving operators and it comes with a cost in terms of resource utilization. 

Also  you have inverted the request and limit values in all the files.

fgiloux:
@camilamacedo86 Thanks. It looks good to me now.

estroz:
@fgiloux thanks a ton for your analysis here and in #2347. It's great to get feedback from a place of experience.

/lgtm

PR #2302, opened 2021-08-18
📖  Update the Kubebuilder book tutorials for v3

hickeyma:
This PR updates the Kubebuilder book tutorials to align with the current version, v3. This fix will help users navigate the book with success.

Fixes #2273

PR #2379, opened 2021-10-13
:book: Update tutorial to use cancel context to stop manager

erikgb:
This PR introduces a cancellable context in the tutorial controller test suite. This context is used to start the manager in the test suite startup, and context is cancelled before initiating the testEnv shutdown in `AfterSuite`. This seems to fix https://github.com/kubernetes-sigs/controller-runtime/issues/1571, and I think this issue should be closed - as not a problem (in controller-runtime).

The PR also reverts the workaround for this issue introduced in https://github.com/kubernetes-sigs/kubebuilder/pull/2302.

It seems like this pattern is already in use in the webhook test suite, so I also removed some uneeded non-nil error ifs in those files.

Note to reviewer: I am uncertain if more needs to be updated.... It seems like the skaffolding of the controller test suite could include more, but I didn't want to extend it as part of this fix.


estroz:
>Maybe we should postpone the K8s upgrade

Sounds good

PR #2417, opened 2021-11-10
Update cert-manager installation url since it changed

rocrisp:
The cert-manager install url is broken so we need to update with a working one. (Camila send me a new url)
- 📖 (:book:): documentation


rocrisp:
I'm going to close this and redo the title of the PR.

PR #2429, opened 2021-11-29
:sparkles: Add resource requests and limits to kube-rbac-proxy

### LINKED ISSUE #2428: Add resources requests and limits for kube-rbac-proxy.

Currently, the kube-rbac-proxy container does not have any resources amount set. 
We ought to add at least the resource request for CPU and Memory for its container. The change to add the amount of the resource would be done in this [bollerplate]Ohttps://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/kdefault/manager_auth_proxy_patch.go#L56). However, we also need to check what values we could define by default here. 

**Why?**

- If a LimitRange is activated in a namespace for computing resources like CPU and memory, users must specify requests or limits for those values. Otherwise, the system may reject Pod creation.
- If a resourQuota be configured and the pods do not have at least requests or limits for those values the system may reject Pod creation.

And then, the above configuration is commonly adopted and recommended in multi-tenancy and or production clusters. Also, setting values to the container would protect something that goes wrong with them which would make consume all available resources and/or affect other containers running on the cluster.   

More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/

### Extra Labels

_No response_

fgiloux:
Signed-off-by: Frederic Giloux <fgiloux@redhat.com>

**Description**

This change introduces default resource requests and limits for the kube-rbac-proxy container scaffolded by KubeBuilder. It also provides a hint to users that this should get amended to match the specific requirements.

**Motivation**

This is similar to what is already done for the controller-manager.

- Provides information to the scheduler for better pod placement
- Prevents the pod of getting evicted or the container of getting OOM killed as first on the line when there is memory pressure on the node but the pod/container does not consume more than what it is supposed to
- Allows the pod to play well with Quotas and LimitRanges without relying on LimitRange defaults

 Fixes #2428

### DIFF
--- a/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/kdefault/manager_auth_proxy_patch.go
+++ b/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/kdefault/manager_auth_proxy_patch.go
@@ -65,6 +65,13 @@ spec:
         - containerPort: 8443
           protocol: TCP
           name: https
+        resources:
+          limits:
+            cpu: 500m
+            memory: 128Mi
+          requests:
+            cpu: 5m
+            memory: 64Mi
 {{- if not .ComponentConfig }}
       - name: manager
         args:
--- a/testdata/project-v3-addon/config/default/manager_auth_proxy_patch.yaml
+++ b/testdata/project-v3-addon/config/default/manager_auth_proxy_patch.yaml
@@ -20,6 +20,13 @@ spec:
         - containerPort: 8443
           protocol: TCP
           name: https
+        resources:
+          limits:
+            cpu: 500m
+            memory: 128Mi
+          requests:
+            cpu: 5m
+            memory: 64Mi
       - name: manager
         args:
         - "--health-probe-bind-address=:8081"
--- a/testdata/project-v3-config/config/default/manager_auth_proxy_patch.yaml
+++ b/testdata/project-v3-config/config/default/manager_auth_proxy_patch.yaml
@@ -20,3 +20,10 @@ spec:
         - containerPort: 8443
           protocol: TCP
           name: https
+        resources:
+          limits:
+            cpu: 500m
+            memory: 128Mi
+          requests:
+            cpu: 5m
+            memory: 64Mi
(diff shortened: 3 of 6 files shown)


fgiloux:
@camilamacedo86 when you have some time can you have a look at this PR?

camilamacedo86 (approved):

camilamacedo86:
/hold

@fgiloux we might want not to go forward with this one. Why?

`CPU requests and limits are associated with Containers, but it is useful to think of a Pod as having a CPU request and limit. The CPU request for a Pod is the sum of the CPU requests for all the Containers in the Pod. Likewise, the CPU limit for a Pod is the sum of the CPU limits for all the Containers in the Pod.`

In this way, seems like that we have it set only for the operator is enough.

fgiloux:
@camilamacedo86 It is true for scheduling that the aggregated value at the Pod level is taken in consideration. On the Node it depends whether the --cgroups-per-qos flag has been set for the Kubelet. It is the case with OpenShift I am not sure for other distributions. This will determine whether a cgroup level (for aggregated accounting) gets created for Pods or not.

That said it is not neutral in terms of resource consumption to add a process, the kube-rbac-proxy in this case. I would agree that the CPU consumption may well be under the approximation level of the controller-manager. It is not the case for memory. At the end it is not that we would be requesting "double" it is that we would account for the overhead that the kube-rbac-proxy brings.

Last point, you motivated this enhancement first by the rejection of Pods when QoS and LimitRanges are configured without default values. LimitRanges can be defined for Pods but also for Containers. If a constraint is set at the Container level without default values the Pod may still get rejected if it does not  provide it.

fgiloux:
> If a constraint is set at the Container level without default values the Pod may still get rejected if it does not provide it.

I need to correct my last sentence. This scenario is not possible: If a limitRange is created with a constraint at the Container level Defaults are always created (automatically)

fgiloux:
I signed it

fgiloux:
I signed it

fgiloux:
@camilamacedo86 just missing lgtm now. Let me know if this is good to go or anything needs to be changed

varshaprasad96 (approved):
/lgtm

PR #2430, opened 2021-12-02
don't generate markers if resource is not created

kunalp-gohire:
Don't generate markers if resource is not created

If we create a kind only with controller and without resource,then in <kind>_controller.go, markers for rbac generates. As there are no resources created for kind then there is no need of rbacs.

### DIFF
--- a/pkg/plugins/golang/v3/scaffolds/internal/templates/controllers/controller.go
+++ b/pkg/plugins/golang/v3/scaffolds/internal/templates/controllers/controller.go
@@ -83,9 +83,11 @@ type {{ .Resource.Kind }}Reconciler struct {
 	Scheme *runtime.Scheme
 }
 
+{{ if not (isEmptyStr .Resource.Path) -}}
 //+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }},verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }}/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }}/finalizers,verbs=update
+{{- end }}
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.


camilamacedo86 (changes requested):
@kunalp-gohire 

Are you saying here that the scenario is:

- Create a project kb init
- Then, use create API command to create an API but choose the option to create only the  controller and not the resource
- The expected result would not check the markers and are you checking then. Am I right?

Question: Why the result would not add the markers? The markers will be required for generating the RBAC for the Kind informed. Otherwise, when you run make manifests the config/rbac will not be updated with the correct permissions to allow your Operator updated/watch/edit/remove the primary resource that you will reconcile with this controller.

Could you please clarify why do you think that the markers should not be set? 

c/c @rashmigottipati @varshaprasad96

kunalp-gohire:
> 
@camilamacedo86 
Yes, this pr is totally based on the scenario that you have mentioned.


As markers are required to generate roles/permissions to operate a particular resource or kind. If we are not creating any resource and just created a controller to reconcile or watch another resource so we manually add markers for that resource so that after making manifests config/rbac should update with correct permissions. As in this scenario, we haven't created a resource for the controller but still, markers for that resource have been created and after applying `make manifests` the permissions ( redundant permissions) are generated.

The markers should generate only if we are creating both controller and resource/crd for it. 
  
So I’m checking if a resource is created for the controller then only creates markers in _controller.go.

kunalp-gohire:
> @kunalp-gohire
> 
> Thank you for your contribution 🥇
> 
> Are you saying here that the scenario is:
> 
> * Create a project kb init
> * Then, use create API command to create an API but choose the option to create only the  controller and not the resource
> * The expected result would not check the markers and are you checking then. Am I right?
> 
> Question: Why the result would not add the markers? The markers will be required for generating the RBAC for the Kind informed. Otherwise, when you run make manifests the config/rbac will not be updated with the correct permissions to allow your Operator updated/watch/edit/remove the primary resource that you will reconcile with this controller.
> 
> Could you please clarify why do you think that the markers should not be set?
> 
> c/c @rashmigottipati @varshaprasad96

@camilamacedo86
Yes, this pr is totally based on the scenario that you have mentioned.

As markers are required to generate roles/permissions to operate a particular resource or kind. If we are not creating any resource and just created a controller to reconcile or watch another resource so we manually add markers for that resource so that after making manifests config/rbac should update with correct permissions. As in this scenario, we haven't created a resource for the controller but still, markers for that resource have been created and after applying make manifests the permissions ( redundant permissions) are generated.

The markers should generate only if we are creating both controller and resource/crd for it.

So I’m checking if a resource is created for the controller then only creates markers in _controller.go.

pkg/plugins/golang/v3/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -83,9 +83,11 @@ type {{ .Resource.Kind }}Reconciler struct {
 	Scheme *runtime.Scheme
```
camilamacedo86, line 83: **General guidance about the PRs**

a) Ensure that you run make generate to update the samples
b) Use the mojo in the title of your PR.
See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#pr-process (in this case would be `:bug:`)
c) You need to assign CLA: https://github.com/kubernetes-sigs/kubebuilder/pull/2430#issuecomment-984301237 at the first time.

pkg/plugins/golang/v3/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -83,9 +83,11 @@ type {{ .Resource.Kind }}Reconciler struct {
...
 }
 
+{{ if not (isEmptyStr .Resource.Path) -}}
 //+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }},verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }}/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }}/finalizers,verbs=update
+{{- end }}
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
```
camilamacedo86, line 93: Hi @kunalp-gohire, 

Thank you for the explanation and collaboration. 🥇 


The testdata/ dir samples are generated based on the source code (pr changes). In this case, see that we already both scenarios for we verify your change:

- (Reconcile an external type Laker) https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh#L72

- (Reconcile a core type Deployments) https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh#L96

_Note that you must run `make generate` to re-generate the samples._ 

**Then, why we would not require the permissions to have reconcile the external type (Laker)?**
*external type: It is for e.g an API/CRD/Kind defined in Project A but used the Project B

Following are some comments inline:

> As markers are required to generate roles/permissions to operate a particular resource or kind. If we are not creating any resource and just created a controller to reconcile or watch another resource so we manually add markers for that resource so that after making manifests config/rbac should update with correct permissions. As in this scenario, we haven't created a resource for the controller but still, markers for that resource have been created and

Are you:
a) Creating an API/Kind and a controller for this kind
b) Then, create another controller for the same kind defined above

Is that the scenario? 

>  after applying make manifests the permissions ( redundant permissions) are generated.

Could not it be related to : https://github.com/kubernetes-sigs/controller-tools/pull/620

camilamacedo86:
Hi @kunalp-gohire, 

I am closing this one based in the comment : https://github.com/kubernetes-sigs/kubebuilder/pull/2430/files#r768220494

I hope that you do not mind. However, if you still thinking that it is required to get done here please feel free to re-open this one and ping me out.

pkg/plugins/golang/v3/scaffolds/internal/templates/controllers/controller.go
kunalp-gohire: @camilamacedo86  There could be 2 scenarios:
1] Create a ‘xyz’ controller to reconcile or watch the ‘abc’ resource which is created in the same project.  

2] Create a ‘xyz’ controller to reconcile or watch the ‘abc’ resource (custom/ standard) 
which is created in another project.

In both cases the markers will be generated for resource ‘xyz’ but there is no resource named ‘xyz’ defined in the project. 


In test data you are creating a controller / api named Laker  without resource  `(create api --group crew --version v1 --kind Laker --controller=true --resource=false --make=false   ) `to reconcile resource Laker which is defined in another project.  In this case markers will be generated for resource Laker as here controller name is same as external type resource Laker.
 But one can create an api/controller ‘sample’ to reconcile external type resources ‘Laker’, ‘Ship’, ect.   
`(create api --group crew --version v1 --kind Sample --controller=true --resource=false --make=false   ) `In this case markers for a sample resource will be created which is not created or defined in the project. And makers for resources Laker, Ship and other resources which we intend to use in a sample controller need to be added manually.


In my opinion markers for api/controller for which resource is not created shouldn’t be generated. `(--resource=false --make=false   )` As there might be a possibility that the controller name and external resource name for which we are creating the controller might not be the same.

Issue #2434, opened 2021-12-02
kube-rbac-proxy scaffolding should not have debug logging enabled by default


geezyx:
https://github.com/kubernetes-sigs/kubebuilder/blob/4ef2eb6b64bac803d181eeec3f1b84f2935b475a/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/kdefault/manager_auth_proxy_patch.go#L63

We noticed that the kube-rbac-proxy logging was really verbose and upon inspection, `--v=10` is actually logging the bearer token of the service account that kube-rbac-proxy is using.

rashmigottipati:
@geezyx Thanks for your contribution! We discussed this in triage meeting and we wanted to make a suggestion on the PR to leave the line as commented out instead of removing it altogether. @camilamacedo86 will make a specific comment on the PR with the suggestion.

geezyx:
@rashmigottipati thank you, I'm responding to the PR feedback now.

Just curious - in the future, is it useful to have both an issue and a PR or would y'all prefer just a PR? I wasn't sure the additional issue was redundant.

Appreciate y'all maintaining a great project!

camilamacedo86:
Hi @geezyx, 

The issue usually tracks the need. Not necessary who tracks the need to work in the solution. 
Also, we try to do the backlog triage where the issues are discussed. 
 
However, if you wish to contribute with something and just send the PR it is fine.
If we see that the info in the description is not enough to understand your motivations and help in the review, or if we see that is a topic that requires more discussion which would be nice have the issue we will let you know.

camilamacedo86:
It is done.

PR #2435, opened 2021-12-02
✨ (kustomize/v1) reduce the debug log level for the sidecar container kube-rbac-proxy from 10 to 0

geezyx:
This change removes the debug logging flag from the kube-rbac-proxy patch that is generated by default, fixing #2434.

As mentioned in the issue, debug logging outputs bearer tokens of the service account that kube-rbac-proxy is using, which can be used to authenticate and perform any authorized actions that your manager is configured for.

I suppose it's okay to assume that folks using kubebuilder should be auditing the generated configs and catching this type of thing, but for folks (like me) that didn't catch it, defaulting to normal logging levels can prevent some headache.

### DIFF
--- a/testdata/project-v3-addon/config/default/manager_auth_proxy_patch.yaml
+++ b/testdata/project-v3-addon/config/default/manager_auth_proxy_patch.yaml
@@ -15,7 +15,7 @@ spec:
         - "--secure-listen-address=0.0.0.0:8443"
         - "--upstream=http://127.0.0.1:8080/"
         - "--logtostderr=true"
-        - "--v=10"
+        - "--v=0"
         ports:
         - containerPort: 8443
           protocol: TCP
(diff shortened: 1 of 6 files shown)


testdata/project-v3-addon/config/default/manager_auth_proxy_patch.yaml
```diff
@@ -15,7 +15,6 @@ spec:
         - "--secure-listen-address=0.0.0.0:8443"
         - "--upstream=http://127.0.0.1:8080/"
         - "--logtostderr=true"
-        - "--v=10"
```
camilamacedo86, line 18: WDYT about we change to 0, then we would have the same result but would still easier for users know how to debug that if/when required?

geezyx: I like that! I went back and forth between removing and setting it. I'll update to use 0 so that it's more clear that the flag is available if needed. Thank you.

geezyx: Done, apologies for the force pushes I had the wrong git config setup 🤦

camilamacedo86 (changes requested):
That is great @geezyx.

See that we use the title and mojos to generate the release note. In this way, could you please fix that? `:sparkles: (kustomize/v1) reduce the debug log level for the sidecar container kube-rbac-proxy from 10 to 0`
 
More info: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#pr-process

Also, before we get the PR merged we need to rebase the commits. Could you please do that?

camilamacedo86 (approved):
/lgtm
/approve

Issue #2389, opened 2021-10-17
EnvTest setup docs need to be updated


camilamacedo86:
### What do you want to happen?

The ENV TEST was changed. So, we need to update the docs to clarifies that as add the new instructions.
See: https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest#where-does-it-put-all-those-binaries

Then, would be very nice instead of removing the current instructions make it clear when it loses its supports. 
 
Also, note that we probably have related steps in the migration guide from v2 to v3 which seems that needs to be revisited. 

### Extra Labels

/kind documentation

camilamacedo86:
c/c @varshaprasad96

afritzler:
Running `make test` on the kubebuilder project itself on my Apple M1 (darwin/arm64) results in this: 

```
Running kubebuilder integration tests
Building kubebuilder
Installing e2e tools with setup-envtest
unable to fetch checksum for requested version: unable fetch metadata for kubebuilder-tools-1.23.1-darwin-arm64.tar.gz -- got status "404 Not Found" from GCS
make: *** [test-integration] Error 2
```

Is there any progress on the darwin/arm64 for the kubebuilder-tools? I couldn't find any issue on that. I could of course locally hack it and force envtest to use amd64 instead but that is a rather nasty solution.

camilamacedo86:
We have now for 1.23.5 and the latest release see: https://storage.googleapis.com/kubebuilder-tools
 However, note that we cannot fully support arm 64 for mac yet. More info can be found on the release page for 3.4.0. 
So, I am closing this one.

rumstead:
Can I suggest that we download the envtest tools to [a relative dir](https://github.com/kubernetes-sigs/controller-runtime/blob/bcde6f084dd15ef93bee2a2e4668184e7472eddd/tools/setup-envtest/main.go#L50)? It's clear and obvious where the binaries are downloaded and would get automatically cleaned up if run in a CI pipeline. 

```Makefile
.PHONY: test
test: manifests generate fmt vet envtest ## Run tests.
	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./... -coverprofile cover.out
```

Happy to contribute if there is agreement.

camilamacedo86:
Hi @rumstead,

[...]

PR #2498, opened 2022-01-28
✨ (go/v3) - Update kustomize from v3.8.7 to v4.4.1 and use replacements instead of vars

afritzler:
Since the usage of `vars` in kustomize will be [deprecated soon](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/vars/) it would make sense to use `replacements` instead as advised in the warning:

> WARNING: There are plans to deprecate vars. For existing users of vars, we recommend migration to replacements as early as possible. There is a guide for convering vars to replacements at the bottom of this page under “convert vars to replacements”. For new users, we recommend never using vars, and starting with replacements to avoid migration in the future.

Another issue we were facing was `vars`-name collisions when using multiple `kubebuilder` based controllers in one root kustomization as `vars` are being propagated to the top. 

To test the correct behaviour you can run

```
diff <(kustomize build ../MY_OLD_DEPLOYMENT/config/default/) <(kustomize build config/default/)
```

to make sure that the expected result matches the one from the original `var` based one. (of course make sure that the webhook configuration is commented out)

### DIFF
--- a/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/kdefault/kustomization.go
+++ b/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/kdefault/kustomization.go
@@ -89,33 +89,95 @@ patchesStrategicMerge:
 # 'CERTMANAGER' needs to be enabled to use ca injection
 #- webhookcainjection_patch.yaml
 
-# the following config is for teaching kustomize how to do var substitution
-vars:
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER' prefix.
-#- name: CERTIFICATE_NAMESPACE # namespace of the certificate CR
-#  objref:
-#    kind: Certificate
-#    group: cert-manager.io
-#    version: v1
-#    name: serving-cert # this name should match the one in certificate.yaml
-#  fieldref:
-#    fieldpath: metadata.namespace
-#- name: CERTIFICATE_NAME
-#  objref:
-#    kind: Certificate
-#    group: cert-manager.io
-#    version: v1
-#    name: serving-cert # this name should match the one in certificate.yaml
-#- name: SERVICE_NAMESPACE # namespace of the service
-#  objref:
-#    kind: Service
-#    version: v1
-#    name: webhook-service
-#  fieldref:
-#    fieldpath: metadata.namespace
-#- name: SERVICE_NAME
-#  objref:
-#    kind: Service
-#    version: v1
-#    name: webhook-service
+# Uncomment the following replacements to add the cert-manager CA injection annotations
+#replacements:
+#  - source: # Add cert-manager annotation to ValidatingWebhookConfiguration, MutatingWebhookConfiguration and CRDs
+#      kind: Certificate
+#      group: cert-manager.io
+#      version: v1
+#      name: serving-cert # this name should match the one in certificate.yaml
+#      fieldPath: .metadata.namespace # namespace of the certificate CR
+#    targets:
+#      - select:
+#          kind: ValidatingWebhookConfiguration
+#        fieldPaths:
+#          - .metadata.annotations.[cert-manager.io/inject-ca-from]
+#        options:
+#          delimiter: '/'
+#          index: 0
+#      - select:
+#          kind: MutatingWebhookConfiguration
+#        fieldPaths:
+#          - .metadata.annotations.[cert-manager.io/inject-ca-from]
+#        options:
+#          delimiter: '/'
+#          index: 0
+#      - select:
+#          kind: CustomResourceDefinition
+#        fieldPaths:
+#          - .metadata.annotations.[cert-manager.io/inject-ca-from]
+#        options:
+#          delimiter: '/'
+#          index: 0
+#  - source:
+#      kind: Certificate
+#      group: cert-manager.io
+#      version: v1
+#      name: serving-cert # this name should match the one in certificate.yaml
+#      fieldPath: .metadata.name
+#    targets:
+#      - select:
+#          kind: ValidatingWebhookConfiguration
+#        fieldPaths:
+#          - .metadata.annotations.[cert-manager.io/inject-ca-from]
+#        options:
+#          delimiter: '/'
+#          index: 1
+#      - select:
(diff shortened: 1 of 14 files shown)


pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/kdefault/kustomization.go
```diff
@@ -89,33 +89,94 @@ patchesStrategicMerge:
...
-#    version: v1
-#    name: webhook-service
-#  fieldref:
-#    fieldpath: metadata.namespace
-#- name: SERVICE_NAME
-#  objref:
-#    kind: Service
-#    version: v1
-#    name: webhook-service
+#replacements:
```
camilamacedo86, line 93: It shows great. But it is missing the comments which explain what should be uncommented and when.
Could you pleaas add the comments?

afritzler: Thanks! Will do!

afritzler:
I guess it also makes sense to remove the `var` in `annotations` in the following files:

crd/patches/enablecainjection_patch.go
kdefault/enablecainection_patch.go

as those are not really needed anymore.

afritzler:
Also how about all the test data and the documentation (book)? Those CERTIFICATE_NAMESPACE, CERTIFICATE_NAME, SERVICE_NAMESPACE and SERVICE_NAME are used all over the place and are not really needed when using the `replacements` approach.

camilamacedo86:
> Also how about all the test data and the documentation (book)? 

To update the docs the best approach is:
a) Run the scripts:
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/generate_cronjob.sh
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/component-config-tutorial/testdata/generate_componentconfig.sh
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/multiversion-tutorial/testdata/generate_multiversion.sh

b) Compare the local changes with master and then add back all that is missing on top ( some files, codes and comments/docs ) 

Also, would be nice to get this process documented in: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md

WDYT about doing that as follow up?
Then, after you do this process contribute with the above doc? 
Could you give a help with that?

camilamacedo86:
Hi @afritzler, 

Also, see that you need to update/change the e2e tests as well: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v3/generate_test.go#L105-L129

afritzler:
@camilamacedo86 thanks a lot for the feedback! I will fix the e2s tests.

I will also squash the commits once everything looks legit and the tests are passing.

Regarding docs: agreed, probably a good idea to have it as a followup PR.

afritzler:
@camilamacedo86 I now stepped into the darwin trap again https://github.com/kubernetes-sigs/kubebuilder/issues/2389#issuecomment-1025262088

testdata/project-v3-addon/Makefile
```diff
@@ -108,7 +108,7 @@ controller-gen: ## Download controller-gen locally if necessary.
 KUSTOMIZE = $(shell pwd)/bin/kustomize
 .PHONY: kustomize
 kustomize: ## Download kustomize locally if necessary.
-	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v3@v3.8.7)
+	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4@v4.4.1)
```
camilamacedo86, line 111: We need to inform you that we are updating the kustomize version in the title of this PR so, we are able to properly generate the release notes.

camilamacedo86 (dismissed):
Great contribution 🥇 

The appdiff ci checker is failing in this case because of the kustomize update. 
It is fine. 

The changes here shows fine for me.
/approved

It would be nice to get a second reviewer as well

coderanger:
For the record this will make the default skeleton incompatible with `kubectl -k` in Kubernetes 1.20 and 1.21. Because of how Kubectl versioning works this isn't a full blocker (even if it was, standalone Kustomize would still be an option), but it should be added to the release notes that this requires Kustomize >= 4.1.3 or Kubectl >= 1.22.0

camilamacedo86:
Hi @coderanger, 

Thank you for your help:
> For the record this will make the default skeleton incompatible with kubectl -k in Kubernetes 1.20 and 1.21.
https://github.com/kubernetes-sigs/kubebuilder/pull/2498/#issuecomment-1026698929

Hi @afritzler, 

Based on the above comment, we cannot move with this one. We need support for the latest 3 k8s versions at least. 
Also, in this scenario, we might need to discuss if these changes would not be better addressed in go/v4 plugin instead.

I think we need:
- Raise an issue to track this need
- Add to the go/v4 milestone
- Link this PR with the solution 
- Also, we need to improve the e2e tests these changes should make the e2e tests fail on the previous k8s versions since are not compatible with them. (this one we can do now)

I did the above actions see: https://github.com/kubernetes-sigs/kubebuilder/issues/2503

 In this way, are you ok with we closing this PR?

camilamacedo86:
/approved cancel

Since it was identified that would mean no longer support k8s < 1.22

afritzler:
@camilamacedo86 as replacements have been introduced in Kustomize v4.1.3 and given your support strategy I guess it is valid to sit this one out. Thanks a lot for your supporting me on that one!

Maybe we can revisit this one once k8s v1.25 hits the road. :-)

coderanger:
@afritzler If you want to release it before then, you can make your own Kubebuilder layout plugin that extends the core one maybe? :)

afritzler:
@coderanger I thought about it - but then you have the maintenance effort. We fixed it now via documentation™️ for us 😄

camilamacedo86:
HI @afritzler, 

Thank you a lot for your contribution. 🥇 
It will not be lost, asap we are able to do this update we can use the changes made in this PR. 

 I am closing this one for now.

PR #2507, opened 2022-02-06
 ✨ : Generate Dockerfile from declarative plugin

viveksyngh:
We need to copy the channels directory in the Dockerfile, particularly
now that we're running with non-root permissions and it is no longer a
one-liner.

Signed-off-by: Vivek Singh <svivekkumar@vmware.com>


pkg/plugins/golang/declarative/v1/internal/templates/dockerfile.go
```diff
@@ -0,0 +1,76 @@
...
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/static:nonroot
+WORKDIR /
+COPY --from=builder /workspace/manager .
+COPY --from=builder /channels /channels
+
+USER 65532:65532
+
+ENTRYPOINT ["/manager"]
+`
```
camilamacedo86, line 76: Why should we start to scaffold the dockerfile for the declarative one?
Should not these changes be applied for all plugins? 
Why the changes in the dockerfile are valid only for the declarative plugin? 

It shows that your suggestions should be done in: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/scaffolds/internal/templates/dockerfile.go instead

pkg/plugins/golang/declarative/v1/api.go
```diff
@@ -97,27 +92,11 @@ func (p *createAPISubcommand) InjectResource(res *resource.Resource) error {
...
-		machinery.WithBoilerplate(boilerplate),
-		machinery.WithResource(p.resource),
-	)
-
-	if err := scaffold.Execute(
-		&templates.Types{},
-		&templates.Controller{},
-		&templates.Channel{ManifestVersion: exampleManifestVersion},
-		&templates.Manifest{ManifestVersion: exampleManifestVersion},
-	); err != nil {
```
camilamacedo86, line 119: This change shows that will broke the plugin and make it stop the controller and types scaffolds. 
Could you please see: https://github.com/kubernetes-sigs/kubebuilder/pull/2507/files#r800561564

camilamacedo86 (changes requested):

A few questions were added could you please check?

Also, note that we need to run `make generate` to generate the samples in the testata dir with the changes performed for each PR.

pkg/plugins/golang/declarative/v1/internal/templates/dockerfile.go
viveksyngh: This PR is to address the changes required in the original PR https://github.com/kubernetes-sigs/kubebuilder/pull/2318 which adds the `init` command. Should we add the `init` command subcommand for the declarative plugin? Also, should this change be added to every plugin ?

pkg/plugins/golang/declarative/v1/api.go
camilamacedo86: c/c @viveksyngh

pkg/plugins/golang/declarative/v1/api.go
```diff
@@ -19,25 +19,20 @@ package v1
...
-
-	"github.com/spf13/afero"
 
 	"sigs.k8s.io/kubebuilder/v3/pkg/config"
 	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
 	"sigs.k8s.io/kubebuilder/v3/pkg/model/resource"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugin/util"
-	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/declarative/v1/internal/templates"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/declarative/v1/scaffolds"
```
camilamacedo86, line 28: The scaffolds need to be inside of `scaffolds/internal/templates` for we follow up the same structure of https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/golang/v3/scaffolds

Also, see that we do not want to export the templates. So, could you please do this small change @viveksyngh ?

pkg/plugins/golang/declarative/v1/api.go
camilamacedo86: solved,

camilamacedo86 (commented):
Hi @viveksyngh that is great. 
It shows for me that we have just a nit https://github.com/kubernetes-sigs/kubebuilder/pull/2507/files#r811398584. Otherwise it shows great for me.

Let's try to ping @justinsb as well.

pkg/plugins/golang/declarative/v1/api.go
camilamacedo86: Sorted out.

camilamacedo86:
I can see that the methods complained in the API diff are present in https://github.com/kubernetes-sigs/kubebuilder/blob/af24fad525316e055c9ebae5beeb01e688b29890/pkg/plugins/golang/declarative/v1/api.go#L73-L77

So

/lgtm
/approved

camilamacedo86 (approved):
/approved

camilamacedo86:
Forcing this one since it. shows an issue in the API check: https://github.com/kubernetes-sigs/kubebuilder/pull/2507#issuecomment-1047751217

PR #2522, opened 2022-02-28
:sparkles: Upgrade ginkgo to v2

haoxins:

### DIFF
--- a/docs/book/src/reference/envtest.md
+++ b/docs/book/src/reference/envtest.md
@@ -94,7 +94,7 @@ You can use environment variables and/or flags to specify the `kubectl`,`api-ser
 See that the `test` makefile target will ensure that all is properly setup when you are using it. However, if you would like to run the tests without use the Makefile targets, for example via an IDE, then you can set the environment variables directly in the code of your `suite_test.go`:
 
 ```go
-var _ = BeforeSuite(func(done Done) {
+var _ = BeforeSuite(func() {
 	Expect(os.Setenv("TEST_ASSET_KUBE_APISERVER", "../testbin/bin/kube-apiserver")).To(Succeed())
 	Expect(os.Setenv("TEST_ASSET_ETCD", "../testbin/bin/etcd")).To(Succeed())
 	Expect(os.Setenv("TEST_ASSET_KUBECTL", "../testbin/bin/kubectl")).To(Succeed())
@@ -104,9 +104,7 @@ var _ = BeforeSuite(func(done Done) {
 
 	_, err := testenv.Start()
 	Expect(err).NotTo(HaveOccurred())
-
-	close(done)
-}, 60)
+})
 
 var _ = AfterSuite(func() {
 	Expect(testenv.Stop()).To(Succeed())
--- a/testdata/project-v3-config/controllers/suite_test.go
+++ b/testdata/project-v3-config/controllers/suite_test.go
@@ -20,13 +20,12 @@ import (
 	"path/filepath"
 	"testing"
 
-	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
-	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
@@ -44,9 +43,7 @@ var testEnv *envtest.Environment
 func TestAPIs(t *testing.T) {
 	RegisterFailHandler(Fail)
 
-	RunSpecsWithDefaultAndCustomReporters(t,
-		"Controller Suite",
-		[]Reporter{printer.NewlineReporter{}})
+	RunSpecs(t, "Controller Suite")
 }
 
 var _ = BeforeSuite(func() {
@@ -71,7 +68,7 @@ var _ = BeforeSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 	Expect(k8sClient).NotTo(BeNil())
 
-}, 60)
+})
 
 var _ = AfterSuite(func() {
 	By("tearing down the test environment")
--- a/docs/book/src/cronjob-tutorial/testdata/project/api/v1/webhook_suite_test.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/api/v1/webhook_suite_test.go
@@ -25,7 +25,7 @@ import (
 	"testing"
 	"time"
 
-	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
@@ -35,7 +35,6 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
-	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
@@ -52,9 +51,7 @@ var cancel context.CancelFunc
 func TestAPIs(t *testing.T) {
 	RegisterFailHandler(Fail)
 
-	RunSpecsWithDefaultAndCustomReporters(t,
-		"Webhook Suite",
-		[]Reporter{printer.NewlineReporter{}})
+	RunSpecs(t, "Webhook Suite")
 }
 
 var _ = BeforeSuite(func() {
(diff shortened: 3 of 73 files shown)


pkg/plugins/golang/v2/scaffolds/internal/templates/controllers/controller_suitetest.go
```diff
@@ -126,7 +126,7 @@ package controllers
 import (
 	"path/filepath"
 	"testing"
-	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/v2"
```
camilamacedo86, line 129: why this change should be done for the go/v2 plugin? 
Also, is the `"github.com/onsi/ginkgo/v2"` really compatible with the dependence versions used by go/v2 and when we scaffold the removed apis v1beta1 for CRDs and webhooks?

haoxins: > when we scaffold the removed apis v1beta1 for CRDs and webhooks?

I'm not understand why this change is associated with `v1beta1 for CRDs and webhooks`? 😂

haoxins: reverted~

docs/book/src/reference/envtest.md
```diff
@@ -104,9 +104,7 @@ var _ = BeforeSuite(func(done Done) {
 
 	_, err := testenv.Start()
 	Expect(err).NotTo(HaveOccurred())
-
-	close(done)
-}, 60)
```
camilamacedo86, line 109: Why the done was removed here?

testdata/project-v3-config/controllers/suite_test.go
```diff
@@ -20,13 +20,12 @@ import (
...
 	"testing"
 
-	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
-	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
```
camilamacedo86, line 29: why the printer was removed?

docs/book/src/reference/envtest.md
haoxins: https://onsi.github.io/ginkgo/MIGRATING_TO_V2#removed-async-testing

testdata/project-v3-config/controllers/suite_test.go
```diff
@@ -44,9 +43,7 @@ var testEnv *envtest.Environment
 func TestAPIs(t *testing.T) {
 	RegisterFailHandler(Fail)
 
-	RunSpecsWithDefaultAndCustomReporters(t,
-		"Controller Suite",
-		[]Reporter{printer.NewlineReporter{}})
```
camilamacedo86, line 49: Why we are changing the default scaffold more than just the import?

haoxins: https://onsi.github.io/ginkgo/MIGRATING_TO_V2#migration-strategy-2

testdata/project-v3-v1beta1/go.mod
```diff
@@ -3,8 +3,8 @@ module sigs.k8s.io/kubebuilder/testdata/project-v3-v1beta1
 go 1.17
 
 require (
-	github.com/onsi/ginkgo v1.16.4
-	github.com/onsi/gomega v1.13.0
+	github.com/onsi/ginkgo/v2 v2.1.3
+	github.com/onsi/gomega v1.17.0
```
camilamacedo86, line 7: This version is probably incompatible. We need to revert the dep if/when the v1beta1 be used see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/commons.go#L38

camilamacedo86 (changes requested):
Some nit changes, mainly
- Ensure that we are updating the dep but does not change the default scaffold implementation
- Ensure that the update is valid only for the go/v3 plugin and we will do the downgrade if/when v1beta1 CRD be a scaffold for backwards compatibilities.  

Also, see:
- The title/commit is used to create the release notes. then, this change affects the scaffold so we need to use the emoji `:sparkles:`. More info: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#pr-process
- WDYT about : (go/v3) - Upgrade dep from github.com/onsi/ginkgo to github.com/onsi/ginkgo/v2

testdata/project-v3-config/controllers/suite_test.go
camilamacedo86: So, how could we achieve the same with the v2? Are we able to provide it via the scaffolds? How will it impact the author's projects? 

Also, it shows a breaking change for authors' projects so the emoji needs to be changed to `warning` :

haoxins:
@camilamacedo86 I just reverted the change to scaffold so that I can finish the `plugin/v3` change tomorrow.
I prefer to do that in another PR, but fine if you think they should be done as a whole.

haoxins:
> Some nit changes, mainly
> 
> * Ensure that we are updating the dep but does not change the default scaffold implementation
> * Ensure that the update is valid only for the go/v3 plugin and we will do the downgrade if/when v1beta1 CRD be a scaffold for backwards compatibilities.
> 
> Also, see:
> 
> * The title/commit is used to create the release notes. then, this change affects the scaffold so we need to use the emoji `:sparkles:`. More info: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#pr-process
> * WDYT about : (go/v3) - Upgrade dep from github.com/onsi/ginkgo to github.com/onsi/ginkgo/v2

Updated~

testdata/project-v3-v1beta1/go.mod
joelanford: @camilamacedo86 I might be totally missing something, but what does the ginkgo version have to do with the CRD version? I would expect ginkgo/v2 to work with projects that use v1beta1 CRDs. Is it just that the other dependencies we're required to use with v1beta1 CRDs use ginkgo/v1, so we're concerned that there would be issues if the kubebuilder project used v2, but the project's dependencies used v1?

The one important thing to cover is this:
- Regardless of the CRD version (or anything else), if a project was originally scaffolded with ginkgo v1, we should continue scaffolding new APIs and webhooks with ginkgo v1.

I'm not sure I saw anything for that in my review, but if we have that covered, disregard.

testdata/project-v3-config/controllers/suite_test.go
camilamacedo86: @haoxins could you please address this one ?

docs/book/src/reference/envtest.md
camilamacedo86: Shows that the migration for this one is: 

```
It("...", func() {
	done := make(chan interface{})
	go func() {
		// user test code to run asynchronously
		close(done) //signifies the code is done
	}()
	Eventually(done, timeout).Should(BeClosed())
})
```

testdata/project-v3-v1beta1/go.mod
camilamacedo86: When we scaffold CRD v1beta1 we need to ensure that we use compatible deps with controller-gen v0.6.2 and controller-runtime v0.9.2 since these are the latest versions where we can use v1beta1 with. So, my point is that:
- We need to check if the ginkgo v2 is compatible with or not, if not we need to try to downgrade ([here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/commons.go#L38))
- However, we have changes in the scaffolds as well, so that makes it very complicated if the ginkgo v2 be incompatible. :-( 
- We will also need to ensure that. the changes in the scaffolds are not a breaking change for projects which were already built with old KB versions and their previous versions. 

Since we need to change the scaffold because of the deprecations it seems like a breaking change that we would only be able to address in the go/v4 plugin. After checking the migration steps I am thinking that we should track it to address with go/v4, see: https://github.com/kubernetes-sigs/kubebuilder/milestone/26

@joelanford wdyt? 

/hold 

(it shows a breaking change and cannot be addressed on go/v3)

camilamacedo86 (changes requested):
We checked in the review that changes in the scaffolds are required and that the previous code will not work with ginkgo v2 so it is a breaking change and then we cannot move forward in go/v3. We will need a go/v4 plugin to add this one. I am thinking that we should track it to address with go/v4, see: https://github.com/kubernetes-sigs/kubebuilder/milestone/26. 

See:
- https://onsi.github.io/ginkgo/MIGRATING_TO_V2#generating-custom-reports-when-a-test-suite-completes
- https://onsi.github.io/ginkgo/MIGRATING_TO_V2#removed-async-testing

@joelanford wdyt?

/hold

(it shows a breaking change and cannot be addressed on go/v3)

docs/book/src/reference/envtest.md
haoxins: yeah, if `async` is important for us, we can do that.
Also, custom reports~

haoxins:
> 

1. 👍 If `async` and `custom reports` are required for us, I can implement these changes.
2. I just created an issue for this PR to track the upgrade for `go/v4` milestone and other discussions, see https://github.com/kubernetes-sigs/kubebuilder/issues/2532.

camilamacedo86:
Hi @haoxins, 

I am closing this one since we can plan that for go/v4 and we could check that is a breaking change and I hope that you do not mind. Thank you it will be helpful for we achieve it with go/v4 as well.

PR #2523, opened 2022-03-01
📖 Fix link for ControllerManagerConfigurationSpec

shuheiktgw:
It seems the link to ControllerManagerConfigurationSpec does not work and we need to use `[]` instead of `()` to navigate to the correct page. Thanks!

https://book.kubebuilder.io/component-config-tutorial/define-config.html


camilamacedo86 (approved):
/lgtm
/approved 

PR #2525, opened 2022-03-05
:book: docs: keep the proper noun Kubebuilder consistent

daniel-hutao:
Signed-off-by: Daniel Hu <tao.hu@merico.dev>

:book: documentation

The name of this project is `Kubebuilder`, and `Kubebuilder` is strictly used in most places in the documents, but there are still many places where `KubeBuilder` is mixed, which I think is not accurate enough.


camilamacedo86 (commented):
@pwittrock wdyt? Is that ok or should be the opposite?

camilamacedo86:
/lgtm
/approved
/ok-to-test
imo that make sense

Issue #2509, opened 2022-02-11
[ci]- Add e2e test against 1.23 k8s version


camilamacedo86:
### What do you want to happen?

See that we test the changes against specific k8s versions clusters: 

<img width="1040" alt="Screenshot 2022-02-11 at 13 59 28" src="https://user-images.githubusercontent.com/7708031/153635104-9928df04-5723-4f19-8777-8d1087fe4ed3.png">

This task is for us to add the checks against 1.23

camilamacedo86:
HI @ryantking would you like to check this one out? 
WDYT?

sunny-b:
@camilamacedo86 I'd be happy to take this on.

sunny-b:
@camilamacedo86 Just to clarify, in order to add the 1-23 e2e test, we need to add it to the `test-infra` repo in [this file](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/kubebuilder/kubebuilder-presubmits.yaml#L22-L51)? And update the `KIND_K8S_VERSION` to 1.23.1?

camilamacedo86:
Hi @sunny-b,

That is the file/repo. I think we can check what is the latest 1.23 version available and use it. I think that would be 1.23.3.

sunny-b:
Looks like it's [1.23.4](https://kubernetes.io/releases/). I'll use that.

sunny-b:
@camilamacedo86 FYI I have an open PR with test-infra: https://github.com/kubernetes/test-infra/pull/25594

camilamacedo86:
Done: 
<img width="937" alt="Screenshot 2022-03-15 at 20 32 24" src="https://user-images.githubusercontent.com/7708031/158467240-ecf0c6e9-6d5d-468b-bf20-d2a295872dd1.png">

PR #2535, opened 2022-03-11
:seedling: Other: update latest k8s version in local tests to 1.23.3

sunny-b:
I'm currently working on adding a PR e2e test for K8S 1.23.3. Since the CI pipeline will be using that version, it makes sense to use that version for the local e2e tests as well.

Addresses #2509


camilamacedo86 (approved):
I think that is fine as well

/lgtm
/approved

Issue #2501, opened 2022-02-01
Move config-gen to a alpha plugin design


camilamacedo86:
### What do you want to happen?

Check the possibility to move : https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/cli/alpha/config-gen to an alpha plugin under https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins

### Extra Labels

/kind cleanup

Kavinjsir:
Hey @camilamacedo86 I'd like to take on this. Would you provide me a reivew or some context for the issue so that I can follow up? Thx!

camilamacedo86:
See that we have a PR that was not finished but init this work: https://github.com/kubernetes-sigs/kubebuilder/pull/2099. You might want to check it to continue this work. It was close to being finished I think.

Kavinjsir:
> See that we have a PR that was not finished but init this work: #2099. You might want to check it to continue this work. It was close to being finished I think.

Hey @camilamacedo86 , thx for informing!
Just to clarify:
1. Plugin design in [#2099](https://github.com/kubernetes-sigs/kubebuilder/pull/2099) provides 3 ways:
    i) `kubebuilder init --plugins go-config-gen` to scaffold `kubebuilder alpha config-gen` in Makefile for deploy
    ii) `kubebuilder init --plugins go-config-gen --with-kustomize` to scaffold kustomize cmds as an alternative
    iii) `kubebuilder init --plugins config-gen.go` to scaffold a complete project with `kubebuilder alpha config-gen` in Makefile
Despite the plugin names, I think option `iii` fits the default needs as scaffolding without kustomize file. Option `ii` provides the way for using kustomize. And option `i` is a standalone one that only scaffolds out the configuration file.
I think they are good to apply. Can I just go on with all of them?

2. By saying moving [cli/alpha/config-gen](https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/cli/alpha/config-gen) to [pkg/plugins](https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins), do we want to move all the code of `cli/alpha/config-gen` to `pkg/plugins`? Maybe similar as the structure of [pkg/plugins/golang](https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/golang)?

3. Do we ultimately want to remove command `kubebuilder alpha config-gen` and use such plugin instead?

[...]

PR #2540, opened 2022-03-14
:sparkles: New plugin for config-gen as alpha.

Kavinjsir:
In respect to [#issue2501](https://github.com/kubernetes-sigs/kubebuilder/issues/2501), moving the package of `config-gen`  from `pkg/cli/alpha` towards `pkg/plugins`.


pkg/cli/alpha.go
```diff
@@ -21,7 +21,7 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
-	configgen "sigs.k8s.io/kubebuilder/v3/pkg/cli/alpha/config-gen"
+	configgen "sigs.k8s.io/kubebuilder/v3/pkg/plugins/config-gen"
 )
```
camilamacedo86, line 25: the idea of https://github.com/kubernetes-sigs/kubebuilder/issues/2501, would be we change it to follow up the [Plugin architecture](https://book.kubebuilder.io/plugins/plugins.html). 

In this way, its acceptance criteria/what we need to do would be like; 

- scaffold the project following the config-gen layout when we run `kubebuilder init --plugins=config-gen/alpha1`
- required to figure out what is the best behaviour, but check what should be done if we run `kubebuilder create API when the scaffold done was using config-gen
- config-gen alpha command migrated to a alpha plugin
- Its sample scaffolded under testdata (see that we do the commands calls to build the sample in https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh)
- It can be a follow-up - but we also need to figure out its docs. We might need a new section under https://book.kubebuilder.io/plugins/plugins.html to begin to document all plugins which are implemented in this repo.

Kavinjsir: Hey, @camilamacedo86 ,
Glad to have your response. Thx for providing me with contexts in details!
If I'm understanding correct, we are expecting to migrate `kb alpha config-gen ...` to `kb init --plugins=config-gen/alpha1`. That seems like a breaking change where it is likely to bring users more convenience to config their controller manager:
I assume when executing `kb create api`, a new folder of an empty `kubebuilderconfiggen.yaml` can be created.
Users can configure their controllers over there.
And so they don't have to manually apply the 'output yaml file'. Instead, `make deploy` should load `kubebuilderconfiggen.yaml` automatically and deploy the controller corresponding to users' configuration.
Is this what we are expecting for?

Kavinjsir: Hi, camila, some follow ups:
1. Considering about the migration destination, would you think `pkg/plugins/config-gen/alpha` be a good place for the code base in respect to an `alpha plugin`? (This seems similar to how current plugins directories are structured..)
2. For the layout generated by `kubebuilder init --plugins=config-gen/alpha1`, do we expect a folder containing `kubebuildconfiggen.yaml` under `config/`? I'm not sure if I'm understanding correct.. Well, the case is we are currently having `config/manager` that composes the yaml file for deploying controller. Maybe it is confusing to have the config-gen configs in parallel with `config/manager` in the same folder?
3. How do we expect to use this plugin once the cmd migrated? On the previous thread above, I was thinking if we want to integrate it into `make deploy`. And...when I pick a view on [the origin talking about `configgen`](https://www.youtube.com/watch?v=mb-aAsZ0-ws&list=PL69nYSiGNLP0SOaFeA9f3dwdCNECEKkX3&index=33&ab_channel=SollyRoss), I realized that it seems an early stage for `config-gen`, maybe it is not fully prepared to integrate it into the main workflow? So I am wondering if it is enough to provide a command to generate yaml file and let users to decide how to use that. If so, is it good to add additional scripts in `Makefile` to generate customized yaml when `init` with the plugin?

camilamacedo86: Hi @Kavinjsir, 

For 1.: For me, that shows fine. +1

For 2 and 3.

Currently, it uses the cmd command. Instead of that, we will need to ensure that when this plugin is used in the init we do not scaffold the kustomize one by default and we scaffold all that is required for the project work. 

See that we have a PR that was not finished but init this work: https://github.com/kubernetes-sigs/kubebuilder/pull/2099. You might continue this work.

cmd/main.go
```diff
@@ -24,6 +24,7 @@ import (
 	cfgv3 "sigs.k8s.io/kubebuilder/v3/pkg/config/v3"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
 	kustomizecommonv1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/common/kustomize/v1"
+	configgenv1alpha "sigs.k8s.io/kubebuilder/v3/pkg/plugins/config-gen/v1alpha"
```
camilamacedo86, line 27: Hi @Kavinjsir, 

The icon in the title needs to be ✨  (we are adding a new plugin), and I would recommend something like
> `:sparkles`: New plugin for config-gen as alpha. 

Also, could you ensure that all commits of this PR are squashed

pkg/cli/alpha/config-gen/README.md
```diff
@@ -7,7 +7,7 @@ Supports:
 - Generating CRDs and RBAC from code
 - Generating webhook certificates for development
 - Selectively enabling / disabling components such as prometheus and webhooks
-  - See [types.go](apis/v1alpha1/types.go) for a list of components
+  - See [types.go](types.go) for a list of components
```
camilamacedo86, line 10: Regards its docs I think we need to move in the direction : https://github.com/kubernetes-sigs/kubebuilder/issues/2542

cmd/main.go
Kavinjsir: Sounds good!

pkg/cli/alpha/config-gen/README.md
Kavinjsir: Yeah, that sounds good.
I can make the docs once the implementation is good.
See someone is working on the thread, if there is a successful one give out ahead of this pr, I can try attach the docs based on that.
Otherwise we may have a follow up?
Would you think this work?

pkg/plugins/config-gen/v1alpha/scaffolds/internal/templates/makefile.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	f.TemplateBody = makefileTemplate
+
+	if f.Image == "" {
+		f.Image = "controller:latest"
+	}
+
+	// FIXME: How can we customize makefile without overwrite it since it is language based?
+	f.IfExistsAction = machinery.OverwriteFile
```
Kavinjsir, line 55: Need some help here:
This `config-gen` plugin will generate a makefile that might overwrite the one from `go/v3`.  So there are two concerns:
1. Most part of the template from both are duplicate. (The main difference is just `make deploy` and `make install`). 
2. It potentially leads contradiction if another plugin is used to generate makefile.

test/testdata/generate.sh
```diff
@@ -70,6 +70,12 @@ function scaffold_test_project {
     fi
 
     $kb create api --group crew --version v1 --kind Laker --controller=true --resource=false --make=false
+  elif [ $project == "project-v3-config-gen" ]; then
+    header_text 'Creating APIs ...'
+    $kb create api --group crew --version v1 --kind Captain --controller=true --resource=true --make=false
+    $kb create webhook --group crew --version v1 --kind Captain --defaulting --programmatic-validation
+    $kb create api --group crew --version v1 --kind Admiral --plural=admirales --controller=true --resource=true --namespaced=false --make=false
+    $kb create webhook --group crew --version v1 --kind Admiral --plural=admirales --defaulting
```
Kavinjsir, line 78: Just to clarify, all the data under `testdata/project-v3-config-gen` is output by local `make generate`based on the update here. I am not sure if this is the correct way to generate test cases.

They pass the `e2e test` on my local laptop, but somehow failed in CI. It seems an issue of go version, not sure how I can fix that.

pkg/plugins/config-gen/v1alpha/scaffolds/internal/templates/makefile.go
Kavinjsir: I suppose we may expect this plugin to be a more general one.(So to make it available in sdk)

I will get some references on the plugins from sdk to see how we can adjust this.

test/testdata/generate.sh
camilamacedo86: The ci issue was not related. That is fixed now. 
Are we using the plugin go/v3 with? I think that we are not right
So, we might need to change the name of the dir to clarify that is the config-gen. 
If I am not wrong it has many samples right? if yes, should we create a dir like config-gen/ then add all samples inside? I will also need to look further on this one to have a better idea.

camilamacedo86:
@pwittrock could you give a hand on this one?

pkg/plugins/config-gen/v1alpha/scaffolds/internal/templates/makefile.go
camilamacedo86: The Makefile is part of the language plugin.
See that we have the kustomize and the base.go/v3 which are a composition to generate the = go/v3 

IMO:
How it will works? We will need to call base.go.<version > + config-gen such as we do now for go/v3: https://github.com/kubernetes-sigs/kubebuilder/blob/master/cmd/main.go#L35-L39

So, why does config-gen needs to scaffold the Makefile? The makefile is scaffolded by the Golang one because it has specific targets which are related to the language ( see that the kustomize plugin is used with other languages such as in SDK with Helm/Ansible so that their specific language plugins do not have go commands on the targets ) 


More info: See the plugin doc PR : https://github.com/kubernetes-sigs/kubebuilder/pull/2569

testdata/project-v3-config-gen/api/v1/admiral_types.go
```diff
@@ -0,0 +1,65 @@
...
+
+// AdmiralList contains a list of Admiral
+type AdmiralList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Admiral `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&Admiral{}, &AdmiralList{})
```
camilamacedo86, line 64: Why config-gen needs to scaffold the files that are done by the go.base.v3 language? 
See: https://github.com/kubernetes-sigs/kubebuilder/pull/2582/files

I'd expected, run : 

```
kubebuilder init --plugins=config-gen/v1-alpha,base.go.kubebuilder.io/v3 --domain example.org --repo example.org/guestbook-operator 
```

Then:
- The init command of config-gen would scaffold only the files which are required and specific for this plugin
- After that, in the chain the init subcommand of the base go v3 plugin will be called and scaffold all language files ( apis, controllers, makefile ) 

In this way we do not hurt concepts such as single responsibility, each plugin does only what is its responsibility.

pkg/plugins/config-gen/v1alpha/scaffolds/internal/templates/makefile.go
camilamacedo86: Please see: https://github.com/kubernetes-sigs/kubebuilder/pull/2540/files#r841070549

test/testdata/generate.sh
camilamacedo86: IMO here we would add a sample that runs: 
`kubebuilder init --plugins=config-gen/v1-alpha,base.go.kubebuilder.io/v3 --domain example.org --repo example.org/guestbook-operator 

Such as we are doing in the PR where we propose the new kustomize/v2-alpha see: https://github.com/kubernetes-sigs/kubebuilder/pull/2583
`

pkg/plugins/config-gen/v1alpha/plugin.go
```diff
@@ -0,0 +1,53 @@
...
+	cfgv3 "sigs.k8s.io/kubebuilder/v3/pkg/config/v3"
+	"sigs.k8s.io/kubebuilder/v3/pkg/model/stage"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins"
+)
+
+const pluginName = "config-gen." + plugins.DefaultNameQualifier
+
+var (
+	supportedProjectVersions = []config.Version{cfgv3.Version}
```
camilamacedo86, line 30: why the new plugin would begin with version 3 ?
That would be 1

pkg/plugins/config-gen/v1alpha/scaffolds/init.go
```diff
@@ -0,0 +1,83 @@
...
+
+	builders := []machinery.Builder{
+		&configGen,
+		&cmConfig,
+		&templates.Makefile{
+			BoilerplatePath:          s.boilerplatePath,
+			ControllerToolsVersion:   ControllerToolsVersion,
+			KustomizeVersion:         KustomizeVersion,
+			ControllerRuntimeVersion: ControllerRuntimeVersion,
+		},
```
camilamacedo86, line 75: That shows not required for config-gen

pkg/plugins/config-gen/v1alpha/scaffolds/init.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+const (
+	// ControllerRuntimeVersion is the kubernetes-sigs/controller-runtime version to be used in the project
+	ControllerRuntimeVersion = "v0.11.0"
+	// ControllerToolsVersion is the kubernetes-sigs/controller-tools version to be used in the project
+	ControllerToolsVersion = "v0.8.0"
+	// KustomizeVersion is the kubernetes-sigs/kustomize version to be used in the project
+	KustomizeVersion = "v4.0.5"
+)
```
camilamacedo86, line 36: That seems not required ether

pkg/plugins/config-gen/v1alpha/scaffolds/internal/templates/makefile.go
```diff
@@ -0,0 +1,192 @@
...
+
+	return nil
+}
+
+//nolint:lll
+const makefileTemplate = `
```
camilamacedo86, line 61: What should be diff for config-gen in the Makefile?
Have we anything that is specific for config-gen that we need to change?
Have we anything that we should remove ?

camilamacedo86: c/c @pwittrock

testdata/project-v3-config-gen/go.mod
```diff
@@ -0,0 +1,74 @@
+module sigs.k8s.io/kubebuilder/testdata/project-v3-config-gen
+
+go 1.17
+
+require (
```
camilamacedo86, line 5: In the go.mod of Kubebuilder, we see that we have deps that are used by config-gen only.
How/When these deps are used?
Should we need to change the go.mod of the scaffold?
Are we doing things with the Kubebuilder CLI to make config-gen work?
How that would end up on the plugin subcommand? 

c/c @pwittrock

pkg/plugins/config-gen/v1alpha/scaffolds/internal/templates/makefile.go
Kavinjsir: Hey @camilamacedo86 , thx for the review!

So this is the main part that confuses me of how such `alpha config-gen` get migrated to the plugin and worked by the integration with language specific plugins.

Because, once it becomes a plugin, the function of `config-gen` can't be actively triggered.
Especially when we'd like to use it as:
```
kubebuilder init --plugins=config-gen/v1-alpha,base.go.kubebuilder.io/v3 --domain example.org --repo example.org/guestbook-operator 
```

I understand that we don't expect the plugin to "update" makefile, as that is defined as "language specific".
And, it seems that most of the makefile templates(go/v3, go/v2, helm) have in common with [the following lines](https://github.com/operator-framework/operator-sdk/blob/master/internal/plugins/helm/v1/scaffolds/internal/templates/makefile.go#:~:text=docker%20push%20%24%7BIMG%7D-,%23%23%40%20Deployment,endif,-.PHONY%3A%20helm%2Doperator):
```shell
.PHONY: install
install: kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
	$(KUSTOMIZE) build config/crd | kubectl apply -f -
...
.PHONY: deploy
deploy: kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
...
```
So, if I'm understanding correctly, the language specific plugin:
1) assumes there is a `config/` folder
2) then defines `Makefile`
3) and determine how the CRD and its controller get installed.

In general, the `config/` is generated by `kustomize plugin`. (Though the language specific plugin does not care who create that folder.)

Consequently, I have this basic question:
> Once the alpha plugin get triggered by `kb init` and generates:
> - kubebuilderconfiggen.yaml
> - controller_manager_config.yaml
> 
> How can users use these files to apply their configured controllers to the cluster?

Kavinjsir: The temporary approach in the this commit is: 
Add this Makefile-Scaffolding template to the alpha plugin. Then the ultimate `Makefile` will have a difference in its `make deploy`:
```shell
deploy: ## Deploy controller to the K8s cluster specified in ~/.kube/config.
	kubebuilder alpha config-gen ./kubebuilderconfiggen.yaml | kubectl apply -f -
```
That's how `config-gen` get used.

Kavinjsir: This might not be a good one considering on the usage as:
```shell
kubebuilder init --plugins=config-gen/v1-alpha,<language-specific-plugin> ...
```

That makes the alpha plugin an alternative to the kustomize plugin.

IIUIC, the language plugins expect the `config/` folder, regardless of who create that.

So maybe the alpha plugin should generate the `config/`.

Then the language specific plugins won't necessary do/care additional things.

What do you think?

camilamacedo86: Hi @Kavinjsir, 

Thank you for working on this one, Following the comments inline.

> I understand that we don't expect the plugin to "update" makefile, as that is defined as "language specific".
And, it seems that most of the makefile templates(go/v3, go/v2, helm) have in common with [the following lines](https://github.com/operator-framework/operator-sdk/blob/master/internal/plugins/helm/v1/scaffolds/internal/templates/makefile.go#:~:text=docker%20push%20%24%7BIMG%7D-,%23%23%40%20Deployment,endif,-.PHONY%3A%20helm%2Doperator):

.The Makefile is scaffold by the language plugin. 
Why it has specific targets which are by language (e.g controller-gen only work with Go, go fmt, and etc..)

camilamacedo86: > This might not be a good one considering on the usage as:

> kubebuilder init --plugins=config-gen/v1-alpha,<language-specific-plugin> ...
> That makes the alpha plugin an alternative to the kustomize plugin.

It is the goal. The purpose of config-gen would be to replace kustomize
That means users could choose if prefer the config-gen or kustomize approach.

camilamacedo86:
@Kavinjsir, 

I spoke with its author and he is ok with us deleting or moving config-gen from the project (looking for maintainability since this option does not seems that won atraction from the users). We will need to discuss the possibilities in the community meetings.

However, I understand that with your help we came to the conclusion that we cannot easily move this implementation to a plugin since we need to generate/execute the config-gen to do the actions. 

In this way, I think we can close this PR and if we need we can re-open it. 
Feel free to re-open it if you see that it fits.

PR #2541, opened 2022-03-15
📖 docs: fix abbreviation style in quickstart

### LINKED ISSUE #2539: Suggesting a small improvement to docs quickstart section

I started reading the kubebuilder book and noticed that this paragraph on the `quickstart` section could be improved

```
If you pressed y for Create Resource [y/n] then you created an (CR)Custom Resource for your (CRD)Custom Resource Definition in your samples (make sure to edit them first if you’ve changed the API definition):
```

1. I'm wondering why it's (CR)Custom Resource instead of Custom Ressource (CR), same for CRD
2. CR and CRD have already been mentioned further up in this document without an explanation, I believe that abreviations should be explained on first mention.
3. I think it should be a Custom Resource not an, but I am not a native speaker so I might be wrong here

If you agree that this should be changed I would be very happy to make a PR for that. 
PS: I wasn't sure if this is a feature request or a bug report, but the template for bug reports seemed overwhelming for such a small change, so I went with a feature request instead.

### Extra Labels

/kind documentation

mariuskimmina:
I want to make a small improvement to the quickstart docs.

The abbreviations `CR` and `CRD` should be written out the first time they are mentioned, while atm they are written out in the middle of the document. I also think that the current writing of (CR)Custom Ressource seemed odd. Wiriting it out first followed by the abbreviation in parentheses seems to be more common and is also described as the correct way [here](https://ell.stackexchange.com/questions/18197/which-goes-in-parentheses-an-abbreviation-or-a-full-form).

Closes #2539.

### DIFF
--- a/docs/book/src/quick-start.md
+++ b/docs/book/src/quick-start.md
@@ -93,7 +93,7 @@ and the `controllers/guestbook_controller.go` where the reconciliation business 
 logic. For more info see [Designing an API](/cronjob-tutorial/api-design.md) and [What's in
 a Controller](cronjob-tutorial/controller-overview.md).
 
-If you are editing the API definitions, generate the manifests such as CRs or CRDs using
+If you are editing the API definitions, generate the manifests such as Custom Ressources (CRs) or Custom Ressource Defintions (CRDs) using
 ```bash
 make manifests
 ```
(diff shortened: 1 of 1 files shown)


camilamacedo86 (approved):
/lgtm
/approved

thank you for your contribution 🥇

Issue #2543, opened 2022-03-17
"kubebuilder create api" does not work with Go 1.18


rbrownwsws:
### What broke? What's expected?

I have just tried to create a new kubebuilder project using Go 1.18

Calling `kubebuilder create api` in a new kubebuilder project fails with an error because "controller-gen" cannot be found.

<details>

<summary>"kubebuilder create api" output</summary>

```
$ kubebuilder create api --group example.co.uk --kind Issuer --version v1alpha1
Create Resource [y/n]
y
Create Controller [y/n]
y
Writing kustomize manifests for you to edit...
Writing scaffold for you to edit...
api/v1alpha1/issuer_types.go
controllers/issuer_controller.go
Update dependencies:
$ go mod tidy
Running make:
$ make generate
go: creating new go.mod: module tmp
Downloading sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0
go: downloading sigs.k8s.io/controller-tools v0.8.0
go: downloading github.com/spf13/cobra v1.2.1
go: downloading github.com/fatih/color v1.12.0
go: downloading golang.org/x/tools v0.1.6-0.20210820212750-d4cc65f0b2ff
go: downloading github.com/gobuffalo/flect v0.2.3
go: downloading github.com/mattn/go-colorable v0.1.8
go: downloading github.com/mattn/go-isatty v0.0.12
go: downloading sigs.k8s.io/structured-merge-diff/v4 v4.1.2
go: downloading github.com/google/go-cmp v0.5.6
go: downloading github.com/inconshreveable/mousetrap v1.0.0
go: downloading golang.org/x/sys v0.0.0-20210831042530-f4d43177bf5e
go: downloading golang.org/x/mod v0.4.2
go: added github.com/fatih/color v1.12.0
go: added github.com/go-logr/logr v1.2.0
go: added github.com/gobuffalo/flect v0.2.3
go: added github.com/gogo/protobuf v1.3.2
go: added github.com/google/go-cmp v0.5.6
go: added github.com/google/gofuzz v1.1.0
go: added github.com/inconshreveable/mousetrap v1.0.0
go: added github.com/json-iterator/go v1.1.12
go: added github.com/mattn/go-colorable v0.1.8
go: added github.com/mattn/go-isatty v0.0.12
go: added github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd
go: added github.com/modern-go/reflect2 v1.0.2
[...]

PR #2548, opened 2022-03-18
:book: update prerequisites to clarifies go versions which are supported

camilamacedo86:
**Description**
📖 Update go version information in the QuickStart Prerequisites

**Motivation** 
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/2543

### DIFF
--- a/docs/book/src/quick-start.md
+++ b/docs/book/src/quick-start.md
@@ -9,9 +9,9 @@ This Quick Start guide will cover:
 
 ## Prerequisites
 
-- [go](https://golang.org/dl/) version v1.15+ (kubebuilder v3.0 < v3.1).
-- [go](https://golang.org/dl/) version v1.16+ (kubebuilder v3.1 < v3.3).
-- [go](https://golang.org/dl/) version v1.17+ (kubebuilder v3.3+).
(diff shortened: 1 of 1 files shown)


coderanger:
-1 on making this so specific. We can certainly recommend versions that are tested with but I don't think this reflects the reality of what versions are supported.

camilamacedo86:
Hi @coderanger,

We can only support these specific versions. 
We do not test all ensure it with all versions and indeed unfortunately we cannot make the project does work in many scenarios when users do not use the supported Golang versions since has breaking changes. E.g. Go 1.18 will not work with the latest release at all since go get was deprecated and no longer available in 1.18. 

these are the versions that we can support, however, users still able to use any other version as please them.

AlmogBaku:
I'm with @camilamacedo86 here. Declaring the versions which are *supported* means that this is what the team has identified as potentially broken. You are always welcome to challenge it, but this is not what we recommend.

/lgtm

AlmogBaku (approved):

camilamacedo86:
By thinking deeper here.
Why we are adding the min/ max go version for each bin released?
In the doc we try to keep the state/info only about the latest release/master

So, that makes no sense. We actually should only have now GO >= 17 and <=18
We are bumping GO 1.18 so we might solve this scenario. 

c/c @AlmogBaku

PR #2546, opened 2022-03-17
🐛 For go/v3, ensure that the plugin can only be used with its go supported version >= 1.17 and < 1.18 

Ankit152:
Hey @camilamacedo86 🙂 
I created a new PR as the previous one just got very messy. 
I am have changed the `plugins/golang/v3/init.go` min and max versions of supported go versions.

[KB is not failing for go versions not supported](https://github.com/kubernetes-sigs/kubebuilder/issues/2530)

Issue #2531, opened 2022-03-08
For go/v4, remove the option to create v1beta1 CRD and Webhooks


camilamacedo86:
### What do you want to happen?

For go/v4, remove the option to create v1beta1 CRD and Webhooks.
These APIs are no longer supported since k8s 1.22 and with versions upper than controller-gen v0.6.2 and controller-runtime v0.9.2. As a workaround to allow users to move forward and not break the backwards compatibility we have been trying to downgrade the versions, see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/commons.go#L38

Places that need to be cleaned up for go/v4: 

**For webhooks**
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/webhook.go#L72-L73
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/webhook.go#L131-L135

**For CRDs**
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/api.go#L109-L110
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/api.go#L190-L197

NikhilSharmaWe:
@camilamacedo86 Could you please respond to my [previous](https://github.com/kubernetes-sigs/kubebuilder/issues/2531#issuecomment-1066653961) comment.

NikhilSharmaWe:
@camilamacedo86 Could you please respond to my https://github.com/kubernetes-sigs/kubebuilder/issues/2531#issuecomment-1066653961 comment.

camilamacedo86:
Hi @NikhilSharmaWe,

All issues under the milestone  https://github.com/kubernetes-sigs/kubebuilder/milestone/26 are breaking changes and can only get done when we do a go/v4 plugin. 

So, if we want to do that we need to: 
- Make a copy of https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/golang/v3 to https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/golang/v4
- Then, rename the plugin. info to go/v4 + ensure that it will be alpha1 version 
- We also need to began to scaffold samples for the go/v4 as we do for go/v3 under the testdata, see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh

[...]

PR #2552, opened 2022-03-19
✨(go/v4-alpha): add new golang plugin which uses the default golang base v3 and the new alpha kustomze/v2-alpha

### LINKED ISSUE #2547: Create go/v4 plugin with alpha1 level to address changes which are breaking ones for go/v3

The goal of this task is to:
- Create go/v4 plugin with alpha1 level to address changes which are breaking ones for go/v3

Also, you can check the policy: https://github.com/kubernetes-sigs/kubebuilder/blob/master/VERSIONING.md#introducing-changes-to-plugins.

**What is required to:** 

- Make a copy of https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/golang/v3 to https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/golang/v4
- see that we define the level / version of the plugin in: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/plugin.go#L29-L31
- Then, rename the plugin. info to go/v4 + ensure that it will be alpha1 version 
- We also need to began to scaffold samples for the go/v4 as we do for go/v3 under the testdata, see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh

Screenshot 2022-03-18 at 15 45 03

For go/v4 we do not need the sample v1beta1 since we will remove its support. See: https://github.com/kubernetes-sigs/kubebuilder/issues/2531

- By latest, we also will need the e2e tests for go/v4 see: https://github.com/kubernetes-sigs/kubebuilder/tree/master/test/e2e/v3

### Extra Labels

_No response_

NikhilSharmaWe:
### Motivation
- Allow users to begin to use the new kustomze eaiser by just calling kubebuilder init --plugin=go/v4-alpha
- Add initial version for go/v4-alpha so that in follow-ups we can also create a go base v4 alpha plugin and begin to address changes like remove the logic to allow create v2beta1 CRDs and webhook which was removed with k8s 1.22
- Also, seems time for we in a follow up deprecate the go/v2 plugin which was created to ensure backward compatibility with the legacy layout of Kubebuilder < 3.0.0 and no longer can be updated

Closes: #2547


NikhilSharmaWe:
@camilamacedo86 
- Till now I have only added `pkg/plugins/golang/v4/` and changed plugin version.
https://github.com/kubernetes-sigs/kubebuilder/blob/fb7c9f303282ab6972a3c812d30a37bf024ba15e/pkg/plugins/golang/v4/plugin.go#L28-L31
Do we need to add something like `[]config.Version{config.Version{Number : 4, Stage: stage.Alpha}}` in `supportedProjectVersions` ? 
- Also about adding `pkg/config/v4`. Changes in the logic will be needed in `config.go` and `config_test.go` files for the purpose as I saw the code inside those files. Could you explain about the approach we should take.

camilamacedo86:
Hi @NikhilSharmaWe,

> Do we need to add something like []config.Version{config.Version{Number : 4, Stage: stage.Alpha}} in supportedProjectVersions ?

Yes, we need to define its level as alpha: 
https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/model/stage/stage.go#L35-L36

Also, this PR is missing the testdata samples and the integration tests (e2e)


> Also about adding pkg/config/v4. Changes in the logic will be needed in config.go and config_test.go files for the purpose as I saw the code inside those files. Could you explain about the approach we should take.

The config defines the Project config which is the structure in https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/PROJECT. The go/v4 plugin can still be using the project 3 config for now. 
See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/VERSIONING.md#versioning

c/c @Ankit152

NikhilSharmaWe:
@camilamacedo86 
Could you please confirm
- **Under kubebuilder/testdata**
 Add files for v4 similar to the files for v3 except [project-v3-v1beta1](https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v3-v1beta1).
- **Under kubebuilder/test/e2e/v4**
Copy paste the files under kubebuilder/test/e2e/v3.

NikhilSharmaWe:
@camilamacedo86 Made changes according to [this](https://github.com/kubernetes-sigs/kubebuilder/pull/2552#issuecomment-1073170689) comment.

camilamacedo86:
@NikhilSharmaWe 

Not passing in the lint: https://github.com/kubernetes-sigs/kubebuilder/pull/2546#issuecomment-1070983260
To fix run make lint-fix in this case

c/c @Ankit152

docs/book/src/reference/watching-resources/testdata/external-indexed-field/api.go
```diff
@@ -19,6 +19,7 @@ package external_indexed_field
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
```
camilamacedo86, line 22: Ideally, we should not touch on these files because they are not part of the scope.
However, I see that these changes are more related to format so that is fine

pkg/plugins/golang/v4/plugin.go
```diff
@@ -0,0 +1,64 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v4
+
+import (
+	"sigs.k8s.io/kubebuilder/v3/pkg/config"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
```
camilamacedo86, line 21: run. make lint-fix to fix this one.

test/e2e/v4/generate_test.go
```diff
@@ -0,0 +1,231 @@
...
+#    name: webhook-service`, "#")).To(Succeed())
+}
+
+// GenerateV3 implements a go/v3(-alpha) plugin project defined by a TestContext.
+func GenerateV3(kbc *utils.TestContext, crdAndWebhookVersion string) {
+	var err error
+
+	By("initializing a project")
+	err = kbc.Init(
+		"--plugins", "go/v3",
```
camilamacedo86, line 138:
```suggestion
		"--plugins", "go/v4",
```

test/e2e/v4/generate_test.go
```diff
@@ -0,0 +1,231 @@
...
+#    fieldpath: metadata.namespace
+#- name: SERVICE_NAME
+#  objref:
+#    kind: Service
+#    version: v1
+#    name: webhook-service`, "#")).To(Succeed())
+}
+
+// GenerateV3 implements a go/v3(-alpha) plugin project defined by a TestContext.
+func GenerateV3(kbc *utils.TestContext, crdAndWebhookVersion string) {
```
camilamacedo86, line 133:
```suggestion
func GenerateV4(kbc *utils.TestContext, crdAndWebhookVersion string) {
```

test/e2e/v4/generate_test.go
```diff
@@ -0,0 +1,231 @@
...
+	var err error
```
camilamacedo86, line 134: See that we should not copy all we need to add the test to generate and run the go/v4 instead

testdata/project-v4-addon/PROJECT
```diff
@@ -0,0 +1,49 @@
+domain: testproject.org
+layout:
+- go.kubebuilder.io/v3
```
camilamacedo86, line 3: These samples were not built with the new alpha plugin

camilamacedo86 (changes requested):
.

NikhilSharmaWe:
Could you please review the changes made in 
- [test/e2e/v4/generate_test.go](https://github.com/kubernetes-sigs/kubebuilder/pull/2552/files/c9c52efb24960d2d86ce3a311f92b075f01d9f58#diff-a132e584478a89842ab0f0e505efd07b930759da970d10af6229f6944bd72992)
- https://github.com/kubernetes-sigs/kubebuilder/blob/c469a16b2f2345d5972d386163e92e22d2e07e7a/test/e2e/v4/plugin_cluster_test.go
- testdata/project-v4-addon/PROJECT

Ankit152:
@NikhilSharmaWe I was going through the code and seems like you have not made any changes to the `test/testdata/generate.sh`. Kindly add the project commands there. I think that's needed to make the `go/v4` work.

Let me know your thought on this. 🙂 

c/c @camilamacedo86

NikhilSharmaWe:
@camilamacedo86 

**Regarding changes in [test/testdata/generate.sh](https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh) for scaffolding test projects**
- Should I add conditions similar to which are made for 
`project-v3`, `project-v3-multigroup`, `project-v3-addon`, `project-v3-config`
for 
`project-v4`, `project-v4-multigroup,` `project-v4-addon`, `project-v4-config`.
- Could you please respond to [this](https://github.com/kubernetes-sigs/kubebuilder/pull/2552#issuecomment-1073179136) comment.
- Also what others changes we need before we work on https://github.com/kubernetes-sigs/kubebuilder/issues/2531.

camilamacedo86:
Hi @NikhilSharmaWe, 

> Regarding changes in [test/testdata/generate.sh](https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh) for scaffolding test projects
Should I add conditions similar to which are made for
project-v3, project-v3-multigroup, project-v3-addon, project-v3-config
for
project-v4, project-v4-multigroup, project-v4-addon, project-v4-config. 

Yes, but all should be built with the go/v4 plugin so that is copied and past and then tweak. 

> Could you please respond to https://github.com/kubernetes-sigs/kubebuilder/pull/2552#issuecomment-1073179136 comment.

Done, already, see that I added comments on this PR. 
You need to change all to use the new go/v4 plugin

> Also what others changes we need before we work on https://github.com/kubernetes-sigs/kubebuilder/issues/2531.

When we have the go/v4 plugin in alpha (this PR merged) then is possible to work on this another task in a follow up

docs/book/src/reference/watching-resources/testdata/owned-resource/controller.go
```diff
@@ -134,4 +135,4 @@ func (r *SimpleDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&appsv1.SimpleDeployment{}).
 		Owns(&kapps.Deployment{}).
 		Complete(r)
-}
```
camilamacedo86, line 137: The changes n the docs will make it hard we do the review and are not part of the scope 
Can you revert, please?

test/e2e/v4/generate_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+	//nolint:golint
+	//nolint:revive
+	. "github.com/onsi/gomega"
+
+	"sigs.k8s.io/kubebuilder/v3/test/e2e/utils"
+)
+
+// GenerateV2 implements a go/v2 plugin project defined by a TestContext.
+func GenerateV2(kbc *utils.TestContext) {
```
camilamacedo86, line 38: Why we are generating a go/v2 here?

NikhilSharmaWe: So we have to 
**Regarding test/e2e/**
- Remove `test/e2e/v4` directory.
- Add a function `GenerateV4` in `test/e2e/v3/generate_test.go` similar to `GenerateV3` but with plugin : go/v4 and version : 4.
- Add Context similar to https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v3/plugin_cluster_test.go#L98-L129
but with `Root : "plugin go.kubebuilder.io/v4"` in `test/e2e/v3/plugin_cluster_test.go` and use func `GenerateV4` instead in its logic.

Could you inform I am understanding correct here.

NikhilSharmaWe:
@camilamacedo86 
So we have to 
**Regarding test/e2e/**
- Remove `test/e2e/v4` directory.
- Add a function `GenerateV4` in `test/e2e/v3/generate_test.go` similar to `GenerateV3` but with plugin : go/v4 and version : 4.
- Add Context similar to https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v3/plugin_cluster_test.go#L98-L129
but with `Root : "plugin go.kubebuilder.io/v4"` in `test/e2e/v3/plugin_cluster_test.go` and use func `GenerateV4` instead in its logic.

Could you inform am I understanding correct here based on your previous comment ?

NikhilSharmaWe:
@camilamacedo86 Could you please respond to my DMs on slack.

cmd/main.go
```diff
@@ -45,11 +53,14 @@ func main() {
...
 			golangv3.Plugin{},
 			gov3Bundle,
+			golangv4.Plugin{},
+			gov4Bundle,
 			&kustomizecommonv1.Plugin{},
 			&declarativev1.Plugin{},
 		),
 		cli.WithDefaultPlugins(cfgv2.Version, golangv2.Plugin{}),
 		cli.WithDefaultPlugins(cfgv3.Version, gov3Bundle),
+		cli.WithDefaultPlugins(config.Version{Number: 4, Stage: stage.Alpha}, gov4Bundle),
```
camilamacedo86, line 63:
```suggestion
		cli.WithDefaultPlugins(cfgv4.Version, gov4Bundle),
```

camilamacedo86: @NikhilSharmaWe ˆ

NikhilSharmaWe:
At this https://github.com/NikhilSharmaWe/kubebuilder/blob/newPlugin/test/e2e/v2/plugin_cluster_test.go#L89 statement test is failing. Also logs shows that cause of the error is different from before `panic: runtime error: invalid memory address or nil pointer dereference` , https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kubebuilder/2552/pull-kubebuilder-test/1505995450979717120.

pkg/plugins/golang/v4/plugin.go
```diff
@@ -0,0 +1,64 @@
...
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang"
+)
+
+const pluginName = "base." + golang.DefaultNameQualifier
+
+var (
+	pluginVersion            = plugin.Version{Number: 4}
+	supportedProjectVersions = []config.Version{{Number: 4, Stage: stage.Alpha}}
+)
```
camilamacedo86, line 31: It shows the reason for the issue. 
The plugin version is : 4 alpha
The Project Version ( Project file config ) : would be the same of go/v3 plugin -> Plugin Config version 3

test/e2e/v3/generate_test.go
```diff
@@ -229,3 +229,101 @@ Count int `+"`"+`json:"count,omitempty"`+"`"+`
...
 	}
 }
+
+// GenerateV4 implements a go/v4(-alpha) plugin project defined by a TestContext.
+func GenerateV4(kbc *utils.TestContext, crdAndWebhookVersion string) {
+	var err error
+
+	By("initializing a project")
+	err = kbc.Init(
+		"--plugins", "go/v4",
```
camilamacedo86, line 239: Then,. when we properly ensure that the go/v4 ( bundle plugin in the cli.go ) be == alpha we will need to change the key. IT will use alpha as well.

AlmogBaku:
since the title will be used to generate release notes I suggest writing something more modest like ":sparkels: Initial base-code for go/v4 alpha plugin"

NikhilSharmaWe:
@AlmogBaku @camilamacedo86 
Error says : `Error: failed to initialize project: unable to run pre-scaffold tasks of \"base.go.kubebuilder.io/v3\": go version 'go1.18' is incompatible because 'plugin requires go1.17 <= version < go1.18'`, but I am using go1.17 version. So, what could be the reason behind these logs?

pkg/plugins/golang/v4/init.go
```diff
@@ -0,0 +1,209 @@
...
+	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin/util"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v4/scaffolds"
+)
+
+// Variables and function to check Go version requirements.
+var (
+	goVerMin = golang.MustParse("go1.16")
```
AlmogBaku, line 38: Min version should be go1.18 (see #2561)

AlmogBaku:
Following up on our Slack chat:

1. The e2e tests are using the `kubekins-e2e` image
2. The image configures to use go 1.18 here https://github.com/kubernetes/test-infra/blob/55ba9f6da3dab1e4df7698e553a7e68661ee0839/images/kubekins-e2e/variants.yaml#L25

What I don't understand, is why do you get this error at all... because you configured the `goVerMax` to `go2.0alpha1`...

pkg/plugins/golang/v4/init.go
AlmogBaku, line 38: Min version should be go1.18 (see #2561)

AlmogBaku (changes requested):

pkg/plugins/golang/v3/init.go
```diff
@@ -36,7 +36,7 @@ import (
 // Variables and function to check Go version requirements.
 var (
 	goVerMin = golang.MustParse("go1.17")
-	goVerMax = golang.MustParse("go1.18")
+	goVerMax = golang.MustParse("go2.0alpha1")
```
NikhilSharmaWe, line 39: Should I revert this change, since the e2e tests were failing because of the issue in the tests repo.

NikhilSharmaWe: Reverted

pkg/plugins/golang/v4/init.go
camilamacedo86: This plugin uses go 1.17.
Then, if this PR goes first we can update it in your PR or in a follow since this one is alpha

NikhilSharmaWe:
@camilamacedo86 Do we also need to update the controller-runtime and k8s dependencies for go/v4 ? Are these the changes because of which the PR is at hold right now.

camilamacedo86:
Hi @NikhilSharmaWe,

We might need to re-think/re-discuss this one, see :
 
- the kustomize change actually is in the kustomize plugin : https://github.com/kubernetes-sigs/kubebuilder/pull/2583
- Then, the only changes that we have so far for go/v4-alpha would be: 

a) use the new version of test libs and change its scaffolds
b) remove the crd-version and webhook version options
c) make the kustomize/v2 the default plugin used by it

So, how can we add the go/v4-alpha and do not be overwhelmed to keep all plugins maintained? 
- Could we just do 1 sample scaffold until it is stable (project-v4-alpha only)? I think so 
- Could we re-use the v3 code for this one where it is equal and do not increase the number of duplicate files? If yes, what would be the pros/cons? 

Therefore, regards this one, I think we will need to:

a) get the kustomize pr merged first
b) change the plugin bundle to use it
c) change the samples generate sh script to gen only 1 project/sample
d) add the doc over this plugin

As follow up:

a) use the new version of test libs and change its scaffolds
b) remove the crd-version and webhook version options

Regarding your question > Do we also need to update the controller-runtime and k8s dependencies for go/v4 ? 
That is the kind of burn that we need to think about before moving forward. 


So, let's see if we can move with the kustomize first (I would suggest we hold a little this one).

NikhilSharmaWe:
@camilamacedo86 what are the progress with the kustomize? and on what more changes we are looking for before we can merge this one?
Please inform me if I can help anyway.

camilamacedo86:
HI @NikhilSharmaWe, 

Until we be able to discuss and move forward with or not:

the kustomize change actually is in the kustomize plugin : https://github.com/kubernetes-sigs/kubebuilder/pull/2583 

Then, I do not think that we can discuss this one. ( So, the pre-requirement for we can back to lock here is to check the above PR merged OR closed ) Then, that means we will have the final required decisions to know what will be the best approach to move forward.

test/e2e/v3/generate_test.go
```diff
@@ -398,3 +398,101 @@ Count int `+"`"+`json:"count,omitempty"`+"`"+`
...
 
 }
+
+// GenerateV4 implements a go/v4(-alpha) plugin project defined by a TestContext.
+func GenerateV4(kbc *utils.TestContext, crdAndWebhookVersion string) {
+	var err error
+
+	By("initializing a project")
+	err = kbc.Init(
+		"--plugins", "go/v4-alpha",
```
camilamacedo86, line 408: The v4 tests myd be in test/e2e/v4 instead 
Could you please ensure that we have a new dir with them?

camilamacedo86 (approved):

IHMO we need to move forward here and add the go/v4-alpha to make it easier for users to begin to experiment kustomize v4 with and we begin to support Apple Silicon.

Also, it seems time to deprecate go/v2 since we no longer can update it and it was created only to ensure the backwards compatibility,

cmd/main.go
```diff
@@ -41,6 +42,11 @@ func main() {
 		kustomizecommonv1.Plugin{},
 		golangv3.Plugin{},
 	)
+	// / Bundle plugin which built the golang projects scaffold by Kubebuilder go/v3 with kustomize alpha-v2
```
camilamacedo86, line 45:
```suggestion
	// Bundle plugin which built the golang projects scaffold by Kubebuilder go/v3 with kustomize alpha-v2
```
nit ^

everettraven (approved):
/lgtm

AlmogBaku (approved):

camilamacedo86:
/hold cancel

we are good to move with this one :-)

PR #2561, opened 2022-03-25
✨ upgrade golang version to 1.18

### LINKED ISSUE #2559: Support go 1.18

Make Kubebuilder supports Golang 1.18 

**Following the main places that need to be changed to achieve this goal:**

- Bump go.mod scaffold for go/v3 scaffolds 
- Also, ensure that we bump the go version in the GitHub Actions and CI. 
- Verify any other places that we should do this update and ensure that all will be properly done. 
- Ensure that you make generate before pushing the PR with the changes. 

**Following what was identified that would also be required to be done to achieve this goal**

- [x] Update for go/v3 controller-runtime release with k8s 1.24 on Kubebuilder : https://github.com/kubernetes-sigs/kubebuilder/issues/2681
- [x] Update for go/v3 controller-tools with a release using k8s 1.24 
- [x] Update K8s version to 1.24
- [x] Update prow tests to run the test with go 1.18: PR: https://github.com/kubernetes/test-infra/pull/25775
- [x] Add prow test for k8s 1.24
- [x] To build the Kubebuilder-tools for k8s 1.24

**We should NOT**
- update go/v2 since we cannot upgrade the deps used on it, and they will not be compatible with go 1.18. 
- we should not bump the go 1.18 on the Kubebuilder go.mod for now because it is used as lib: (To allow people to consume it and we should adopt the same approach from C+R and controller-tools. More info:https://kubernetes.slack.com/archives/C02MRBMN00Z/p1652178727902549?thread_ts=1652177106.531319&cid=C02MRBMN00Z)

### Extra Labels

_No response_

AlmogBaku:
**Description:**
This PR introduces Golang 1.18 support for Kubebuilder.
It makes Kubebuilder use go1.18 for both compiling the CLI and the scaffolded code. 

For more information about this, see issue #2559

**Motivation:**
Closes #2559


testdata/project-v3/go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kubebuilder/testdata/project-v3
 
-go 1.17
+go 1.18
```
camilamacedo86, line 4: Why bump golang would be a breaking change?

AlmogBaku:
This PR to the `test-infra` should upgrade prow to use golang 1.18 for our tests:
https://github.com/kubernetes/test-infra/pull/25775

.github/workflows/apidiff.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v2
```
camilamacedo86, line 20: Thank you for the contribution, but see that it still missing be updated in some place : https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kubebuilder/2561/pull-kubebuilder-test/1507421392423882752

.github/workflows/apidiff.yml
```diff
@@ -19,7 +19,7 @@ jobs:
...
         with:
-          go-version: '1.17'
+          go-version: '1.18'
       - name: Execute go-apidiff
         uses: joelanford/go-apidiff@v0.2.0
         with:
```
camilamacedo86, line 25: Question: should we define it as a breaking change?
If yes, why? What breaks for those who scaffold the project with the previous versions and now will use the next KB CLI version with this bump? Should not be it an `:sparkle:`?

.github/workflows/apidiff.yml
camilamacedo86: See: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/kubebuilder/kubebuilder-presubmits.yaml#L13

.github/workflows/apidiff.yml
AlmogBaku: TBH, I wasn't sure also how to mark it:

- It's marked as breaking changes because it requires a go 1.18 for new projects. Also, there's a high possibility we'll want to use Go 1.18 features such as generics.
- Other than that, old projects will be able to continue to use it.

WDYT?

camilamacedo86: I do not think that it is a breaking change ( indeed we cannot introduce breaking changes )
Why?
Breaking change will be only considered when we do a change that breaks something for those who scaffold a project with a KB version previously and then by getting the latest KB with the change will face some issue that is not the case. See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/VERSIONING.md

Also, we have no generics in the scaffolds. Ditto so, when/if we add any implementation that uses it by default would be a breaking change and could only be introduced in a new /future golang plugin version such as go/v4.

pkg/plugins/golang/v3/init.go
```diff
@@ -35,8 +35,8 @@ import (
 
 // Variables and function to check Go version requirements.
 var (
-	goVerMin = golang.MustParse("go1.17")
-	goVerMax = golang.MustParse("go1.18")
+	goVerMin = golang.MustParse("go1.18")
```
camilamacedo86, line 38: Could we not change the min?
We officially will support go 1.18 as always.
However, have we any reason to believe that the CLI will not provide the scaffolds currently with 1.17?
No. So, I do not think that we need to force the go 1.18 in this case

camilamacedo86 (commented):
Just a few nits for we are able to get this one merged

a) Could you please update the PR title for `:sparkles: For go/v3 only: upgrade golang version to 1.18 (adds supports to go 1.18)`
c) Could you please add in the description that this one closes: https://github.com/kubernetes-sigs/kubebuilder/pull/2571/files and also ensure that we do the updates for the golang plugin v2?

pkg/plugins/golang/v3/init.go
AlmogBaku: ok

.github/workflows/apidiff.yml
AlmogBaku: ok.  I changed the title.

I'm not sure why do you want to change the Github action go version...

.github/workflows/apidiff.yml
AlmogBaku: I haven't ;-)
See this PR
https://github.com/kubernetes/test-infra/pull/25775

.github/workflows/apidiff.yml
camilamacedo86: To ensure that we test the project with the go version used by the project

camilamacedo86 (changes requested):
Please, see that we need to do a little bit more here to get this one merged
Could you please try to address the commands/suggestions?

IF you have doubts feel free to ping me in the slack.

docs/book/src/quick-start.md
```diff
@@ -12,6 +12,7 @@ This Quick Start guide will cover:
 - [go](https://golang.org/dl/) version v1.15+ (kubebuilder v3.0 < v3.1).
 - [go](https://golang.org/dl/) version v1.16+ (kubebuilder v3.1 < v3.3).
 - [go](https://golang.org/dl/) version v1.17+ (kubebuilder v3.3+).
+- [go](https://golang.org/dl/) version v1.18+ (kubebuilder from master).
```
camilamacedo86, line 15: The master doc will become to be prod so we need to keep it == prod
Because this makes no sense we have versions per bin
We do not track these changes. 

IMO: It should be ONLY 
> go](https://golang.org/dl/) version v1.18+

pkg/cli/alpha/config-gen/controller-gen-filter.go
```diff
@@ -64,7 +64,10 @@ func (cgr ControllerGenFilter) Filter(input []*yaml.RNode) ([]*yaml.RNode, error
 
 	// set the directory
 	b := bufferedGenerator{}
-	rt, _ := gens.ForRoots(cgr.Spec.CRDs.SourceDirectory) // ignore the spurious error
```
camilamacedo86, line 67: it was ignoring the errors
Why do we need to return the error to bump 1.18?

pkg/cli/alpha/config-gen/testdata/componentconfig/expected.yaml
```diff
@@ -246,9 +101,9 @@ metadata:
     control-plane: controller-manager
 spec:
   ports:
-    - name: https
-      port: 8443
-      targetPort: https
+  - name: https
+    port: 8443
+    targetPort: https
```
camilamacedo86, line 106: Why do we need to do these changes to bump 1.18?

pkg/cli/alpha/config-gen/testdata/default/expected.yaml
```diff
@@ -270,28 +125,28 @@ spec:
...
+        name: manager
+        resources:
+          limits:
+            cpu: 100m
+            memory: 30Mi
+          requests:
+            cpu: 100m
+            memory: 20Mi
+      - name: kube-rbac-proxy
+        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0
```
camilamacedo86, line 143: Not part of this one
But we need in a follow up to upgrade the kube-proxy for 0.11.0

pkg/cli/alpha/config-gen/controller-gen-filter.go
camilamacedo86: Found : https://github.com/kubernetes-sigs/kubebuilder/pull/2581#pullrequestreview-928178949 👍

camilamacedo86: Could you add that it also solves : Closes; https://github.com/kubernetes-sigs/kubebuilder/issues/2567
Or indeed would be great if we could merge the fix and the bump in diff PRs

pkg/cli/alpha/config-gen/testdata/default/expected.yaml
AlmogBaku: It was changed when I ran the tests. Without it, the teats don't work.

pkg/cli/alpha/config-gen/testdata/componentconfig/expected.yaml
AlmogBaku: It was changed when I ran the tests. Without it, the teats don't work.

pkg/cli/alpha/config-gen/testdata/default/expected.yaml
camilamacedo86: That is fine. This is == the scaffold produced in the config/ dir 
The version of the image is old we just need another task to update the config-gen and solve its tech-debts

docs/book/src/quick-start.md
AlmogBaku: So, instead of listing the different versions, just specify 1.18?

AlmogBaku:
@camilamacedo86, FYI - following our back-and-forth here:

1. I rebased from master
2. Regenerated `config-gen` test data to support go 1.18
3. Updated the `docs/book/src/quick-start.md` to include *only* 1.18+ (as discussed)
4. Updated linter (better go 1.18 support, we'll have to track the updates there though)
5. make generate, go mod tidy, golangci-lint,  rebased to one commit, tested, etc.

This is finally met all of the requirements, fully tested, and ready to be merged :)

AlmogBaku:
Hmm... the APIDiff job is failing although in my environment it works like charm.
@joelanford can you maybe give us a hint here?

go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kubebuilder/v3
 
-go 1.17
+go 1.18
```
joelanford, line 3: Will this require all importers of `kubebuilder/v3` to also upgrade to 1.18?

If so, do we want to force that migration so soon after 1.18 was released?

I wonder if we can make the scaffolding support 1.18 without bumping the kubebuilder module itself to 1.18?

AlmogBaku: 1. Who's importing `kubebuilder/v3`? Also, they don't have to import `master` version - they can pin the version to the previous tag.
2. Not merging this will block users that build the binary locally (i.e. brew users)

camilamacedo86:
@AlmogBaku what shows be missing here: 

a) We need to discuss if we should or not keep the go/v2 plugin also updated to work with go 1.18 ( what is the pros and cons) should we move with the changes like https://github.com/kubernetes-sigs/kubebuilder/pull/2571 yes or not? 

If yes: so this PR is missing up change all scaffolds for go/v2 as well. 
if not: we need to clarify in the title description that the change affects the plugins X, Y, Z , also Kubebuilder and etc. 

Note that we use the title to generate the release notes. 

b) We also need to discuss should we move forward with the Kubebuilder upgrade or not? What are the problem and consequences? What is the best at this moment

Regards the issue in the job `pull-kubebuilder-test` . That seems to happen because it is not rebased with master and has not the changes made by you in the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/2586 Could you please check it?

pkg/plugins/golang/v2/scaffolds/internal/templates/gomod.go
```diff
@@ -50,5 +50,6 @@ go 1.13
 
 require (
 	sigs.k8s.io/controller-runtime {{ .ControllerRuntimeVersion }}
+	golang.org/x/sys v0.0.0-20220330033206-e17cdc41300f // indirect
```
camilamacedo86, line 53: Why are you adding it here?
IMO it shows that we need to upgrade all plugins to use go 1.18

testdata/project-v3/.gitignore
```diff
@@ -6,6 +6,7 @@
 *.so
 *.dylib
 bin
+testbin/*
```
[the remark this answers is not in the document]
camilamacedo86: Also, I do not see the boilerplate updated with :-(

docs/book/src/cronjob-tutorial/testdata/project/go.mod
```diff
@@ -1,6 +1,6 @@
 module tutorial.kubebuilder.io/project
 
-go 1.17
+go 1.18
```
camilamacedo86, line 3: The docs are not update accordingly
These are projects generated with the tool.
Could we revert all changes under docs/book and then update it in a follow-up?
See that we need to regenerate the samples used in the docs with scripts and ensure that it is properly updated.

pkg/plugins/golang/v2/scaffolds/internal/templates/gomod.go
AlmogBaku: that's probably the result of `go mod tidy`

Makefile
```diff
@@ -82,7 +82,7 @@ GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
 golangci-lint:
 	@[ -f $(GOLANGCI_LINT) ] || { \
 	set -e ;\
-	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell dirname $(GOLANGCI_LINT)) v1.41.1 ;\
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell dirname $(GOLANGCI_LINT)) v1.45.2 ;\
```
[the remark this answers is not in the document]
AlmogBaku: Because the new version support 1.18 better.

I can split it - your call.

testdata/project-v3/.gitignore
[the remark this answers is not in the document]
AlmogBaku: Isn't that autogenerated? I have no idea how it got there.

I tried to delete the directory and redo `make generate` and got the same results

docs/book/src/cronjob-tutorial/testdata/project/go.mod
AlmogBaku: I'm not sure I understood...

pkg/plugins/golang/v2/scaffolds/internal/templates/gomod.go
AlmogBaku: Because there's a bug with the version that the `go mod tidy` imports(an older one due to old dependencies), which this version solves.

Here, I manually override the old dependency with a newer one

AlmogBaku: I'll add a comment

camilamacedo86: We cannot solve manually. See: https://github.com/kubernetes-sigs/kubebuilder/runs/5868585695?check_suite_focus=true#step:6:929

We generate all samples for each PR change and we ensure that the result of pr/repo is == what is done by the CLI. Otherwise, we never would have that properly updated. The point here is that by doing this change any project scaffold with the go/v2 plugin will have this dep added as well.  So, IMO we need to upgrade the go/v2 plugin too.

docs/book/src/cronjob-tutorial/testdata/project/go.mod
AlmogBaku: @camilamacedo86 now I understand...
OMG, this is a nightmare 😖

Are these dependencies so important? also, can't we split it into a different PR?
I suggest I'll just update the deps and keep the rest the same (otherwise, it's just crazy without the knowledge of what's happening here)

go.mod
joelanford: 1. Anyone who has written a plugin. I know of at least projects in the operator-framework that import it. Not sure if there are others. If it was _just_ the operator-framework projects, I'd concede this isn't that big of a deal.
2. You're saying someone running Go 1.18 cannot build a Go module that declares 1.17 compatibility? That doesn't seem right since Go 1 has a back-compat guarantee. Or is it something else?

pkg/cli/alpha/config-gen/testdata/enableprometheus/expected.yaml
```diff
@@ -9,7 +9,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: (unknown)
+    controller-gen.kubebuilder.io/version: ""
```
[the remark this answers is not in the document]
joelanford: We should dig into this a little bit. IIRC there's some code in controller-gen that knows how to auto-detect the version of the binary that's running. This may have changed in 1.18 with the changes that came in with auto-embedding version information as part of `go build`.

pkg/plugins/golang/v2/scaffolds/internal/templates/gomod.go
AlmogBaku: I'm not sure I understand... I manually do that in the scaffolding - that's means I'm making sure that `go mod tidy` won't use a lower version.

go.mod
AlmogBaku: a. When I run Kubebuilder that was built using 1.17 from my environment(that has 1.18) it throws some errors
b. When installing Kubebuilder using brew, it's locally compiles the binaries using go 1.18 - which leads to a failure

AlmogBaku:
Hey all,
I'm trying to understand where this PR stand, and why it's blocked:

Please comment here on your notes, and let's take a call what are the next action items.
@joelanford @camilamacedo86

docs/book/src/cronjob-tutorial/testdata/project/go.mod
camilamacedo86: @AlmogBaku, 

My suggestion here is:
- Revert all changes done in the directory: [docs/](https://github.com/kubernetes-sigs/kubebuilder/pull/2561/files/6f233b6a3edab85f9486e2c18dc24176a10d6eb7#diff-93830d0e4cc14bb4d52a3cce90f1ade789b7293dab0f75abaabb7afb7de4ebf3)
- We do a PR to just updated the docs with all master changes done so far

Makefile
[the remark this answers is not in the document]
camilamacedo86: That would be great. Shows that the current golanglint version has some issues and is raising errors when should not. Could we try to create a PR to update it to the latest available version? WDYT?

testdata/project-v3/.gitignore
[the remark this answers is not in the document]
camilamacedo86: See:
- When we run make install -> we create a binary using the current code source 
- Then, when we run make generate -> we use the bin to execute an script which scaffolds all samples/examples under the testdata dir using the current version code

Then, when we push a pr we verify if the testdata dir is properly updated (by doing the same and using git diff)
So, if we go there and change what is inside of the testdata dir, the samples the tests will fail be cause the samples are NOT == what is the tool result.

pkg/plugins/golang/v2/scaffolds/internal/templates/gomod.go
camilamacedo86: See:
- When we run make install -> we create a binary using the current code source 
- Then, when we run make generate -> we use the bin to execute an script which scaffolds all samples/examples under the testdata dir using the current version code

Then, when we push a pr we verify if the testdata dir is properly updated (by doing the same and using git diff)
So, if we go there and change what is inside of the testdata dir, the samples the tests will fail because the samples are NOT == what is the tool result.

docs/book/src/cronjob-tutorial/testdata/project/go.mod
AlmogBaku: done

Makefile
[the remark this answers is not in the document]
AlmogBaku: done

pkg/cli/alpha/config-gen/testdata/enableprometheus/expected.yaml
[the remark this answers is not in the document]
AlmogBaku: Hi @joelanford 
I'm not sure I'm following here - how's that related to this "(unknown)," and why is it important to understand how it gets serialized?

pkg/plugins/golang/v2/scaffolds/internal/templates/gomod.go
AlmogBaku: I think there's a misunderstanding here - what I meant is that I'm *specifically* generating it in the scaffolding template, and not letting `go mod tidy` to import this dependency.

AlmogBaku: this file, is a `*.go` file...

AlmogBaku:
Hi all,
Is there anything left before merging this?
@camilamacedo86 @joelanford @AvineshTripathi

pkg/cli/alpha/config-gen/testdata/componentconfig/expected.yaml
```diff
@@ -58,7 +58,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: (unknown)
+    controller-gen.kubebuilder.io/version: ""
```
camilamacedo86, line 61: we need to understand this change

pkg/cli/alpha/config-gen/testdata/enableprometheus/expected.yaml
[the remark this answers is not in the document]
camilamacedo86: +1 for @joelanford comments. 
It seems valid we deeger a little to understand it to be sure that it is right and we are not breaking anything.

camilamacedo86 (commented):
@AlmogBaku, 

The issue was updated to clarify all that was discussed and what we will need to do to achieve this goal. See: https://github.com/kubernetes-sigs/kubebuilder/pull/2561

Thank you for working on that and helping us to identify the needs.

testdata/project-v2/go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/go-logr/logr v0.1.0
 	github.com/onsi/ginkgo v1.12.1
 	github.com/onsi/gomega v1.10.1
+	golang.org/x/sys v0.0.0-20220330033206-e17cdc41300f // indirect; we enforce this version to supports go 1.18+
```
camilamacedo86, line 9: We need to check this one.
Should we not update go/v2 to also use go 1.18 or not?

pkg/cli/alpha/config-gen/testdata/enableprometheus/expected.yaml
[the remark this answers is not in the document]
AlmogBaku: see https://github.com/kubernetes-sigs/controller-tools/pull/662

go.mod
AlmogBaku: I've tested it with `go.mod` specifying version 1.17 and it seems to work.
That being said, I don't think it makes sense to generate files for go 1.18 but not to require 1.18 for Kubebuilder:
That basically means we will not bump the version to go 1.18, but only "support" 1.18 - which will prevent us from using 1.18 features in the future.

If an upstream user wants to use go 1.17 but is not willing to upgrade, he can always specify a different version.

testdata/project-v2/go.mod
AlmogBaku: TBH, I don't see any reason not to upgrade go/v2 as well. This is the only change we introducing to go/v2 other than specifying the version

go.mod
camilamacedo86: @AlmogBaku, 
The concern here is to know for now who imports the Kubebuilder as LIB. Do we know what are the impacts that they can face? ( e.g SDK uses Kubebuilder as a lib, will SDK still be able to use it after the changes or to still importing KB will be required also upgrade the SDK project? ) 

Kubebuilder CLI users run the binary so for them only matters the go version in the scaffolds produced.

AlmogBaku: by defining the `go.mod` to 1.18 we'll require upstream users to use 1.18, and on the other hand- we'll benefit from 1.18 features.

Users (such as the Operator-SDK) that haven't upgraded to 1.18 will have to either upgrade or pin the version to v3.3.0.

AlmogBaku: As a side note- IMO, not upgrading go will make this PR **totally redundant**.

To solve compatibility issues, we just need to upgrade a few dependencies, and that's it. But then we're not upgrading anything.

joelanford: I think there's some detail lacking in this thread.

> a. When I run Kubebuilder that was built using 1.17 from my environment(that has 1.18) it throws some errors
> b. When installing Kubebuilder using brew, it's locally compiles the binaries using go 1.18 - which leads to a failure

What errors, what failure?

> By defining the go.mod to 1.18 we'll require upstream consumers to use 1.18, and on the other hand- we'll benefit from 1.18 features.

Are we introducing changes in the Kubebuilder CLI codebase that require Go 1.18? If so, can we call those out? If not, it seems better to wait until we need those 1.18 features before we force library consumers to upgrade their go version.

I'm not saying we definitely shouldn't bump the kubebuilder go.mod go version to 1.18, I'm just asking for the specific motivation for that. Kubebuilder's go dependencies can be confusing because we're talking about at least 2 different things:
1. The version of Go used to compile Kubebuilder
2. The version of Go that the code _scaffolded_ by Kubebuilder uses.

As far as I know, these are completely disjoint from each other (after all, I could write a python program that scaffolds Go code the exact same way that Kubebuilder does). Is there something we've done that couples these versions somehow?

I think the motivation for 2 is extremely obvious (we want users of the Kubebuilder CLI to be able to build projects with the latest and greatest Go version). But for 1, it seems like we're saying we _need_ to bump the Kubebuilder go.mod in order to support scaffolding 1.18 code. That's where I'm not following.

Other than my question about whether we're ready to force kubebuilder importers to use 1.18, I don't have any other qualms with bumping the kubebuilder go.mod.

I'd just like to see an explanation for why we're bumping the kubebuilder go.mod. If it's reasonable or pragmatic, great let's do it! I just think its important to have a complete understanding and consensus for why we we're doing it.

AlmogBaku: The reason is to prepare the ground for future contributions that require go 1.18.

Not doing so now will just postpone the decision.
Imagine a contributor that wants to add a simple fix using Generics, because we haven't taken the call yet - we'll have to ask him to rewrite the solution without Generics.

joelanford: I guess I'd imagine that that PR would be their generics change, plus the one liner in the go.mod file. Is it more than that?

joelanford: I could see us saying: Oh, we need to bump to 1.18 for that. Let's block your PR on this other 1.18 bump PR. But at least then there's a specific motivation to bump and we can debate the merits of requiring 1.18 vs staying at 1.17.

We could even debate your point that _right now_ if we separated this PR out into a) bump KB to 1.18 and b) bump the scaffolding to 1.18.

camilamacedo86: +1 for @joelanford comments
I do not think that using generics in the Kubebuilder API is something that we will do soon at all.
Anyway, we also identified that we ought to bump the k8s api with and update the CLI/API to use go 1.18 seems that we also need to update the controller tools with go 1.18 since config-gen is using this project as a lib.

testdata/project-v2/go.mod
camilamacedo86: As raised and checked by @varshaprasad96 we should not do that.
Why? it might introduce issues to the users since the go version will not be compatible with the deps used.

camilamacedo86:
@AlmogBaku, 

Again, thank you for your terrific work 🥇  and help us identify what we will need to do to accomplish this goal. 
However, we checked that we will need to:

a) split the changes between CLI and go/v3 ( scaffold ) 
b) also we will need to move forward with other updates and changes along with

In this way, wdyt about we close this one and then work on the new PRs?

PR #2565, opened 2022-03-26
✨ Update kustomize version for Macbook M1 support

hajowieland:
Currently we set `KustomizeVersion` to 3.8.7 where there are no kustomize arm64 assets available for Macbook M1 machines.

Starting with kustomize **4.2.0** there are darwin_arm64 packages available. We should update the kustomize version so developers with M1 machines can download the kustomize binary with `install_kustomize.sh` in `Makefile`.

### DIFF
--- a/pkg/plugins/golang/v3/scaffolds/init.go
+++ b/pkg/plugins/golang/v3/scaffolds/init.go
@@ -34,7 +34,7 @@ const (
 	// ControllerToolsVersion is the kubernetes-sigs/controller-tools version to be used in the project
 	ControllerToolsVersion = "v0.8.0"
 	// KustomizeVersion is the kubernetes-sigs/kustomize version to be used in the project
-	KustomizeVersion = "v3.8.7"
+	KustomizeVersion = "v4.2.0"
 
 	imageName = "controller:latest"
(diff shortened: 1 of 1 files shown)


camilamacedo86:
Hi @hajowieland, 

We did not make this updated because of: https://github.com/kubernetes-sigs/kubebuilder/pull/2498#issuecomment-1026698929. Also, see:  https://github.com/kubernetes-sigs/kubebuilder/pull/2498#issuecomment-1026772984

camilamacedo86:
HI @hajowieland, 

This PR would not pass because would be required to re-gen all samples ( see the contribution guide, before pushing a pr you need to run make generate ) 

However, we need to discuss if we should bump the kustomize in v3 and risk or try to ensure in another way that we do not introduce a breaking change within the Golang v3 plugin or if it should only be shipped on the Go v4 and we just support this arch there. Please, follow up on the issue:  https://github.com/kubernetes-sigs/kubebuilder/issues/2566

Let's try to centralize there the convey of our discussions and etc. 
So, that I hope that you do not mind we close this PR
However, please feel free to re-open it if you see that fits.

PR #2568, opened 2022-03-28
:book: Add metrics references.

Kavinjsir:
**Description:**
Add a description table for default metrics given by controller-runtime.

**Motivation:**
Current section of metrics mainly introduces how to expose metrics and how to create additional ones.
It may be helpful to give users the references of the default available metrics given by controller-runtime.
This may provide more insights on how to create Grafana dashboards to observe controller status and create rules/alerts based on the metrics to report abnormal.


docs/book/src/reference/metrics.md
```diff
@@ -73,6 +76,29 @@ for the metrics exported from the namespace where the project is running
...
+| controller_runtime_active_workers             | Gauge     | Number of currently used workers per controller.                                                                                                                                                                                     |
+| controller_runtime_webhook_latency_seconds    | Histogram | Histogram of the latency of processing admission requests.                                                                                                                                                                           |
+| controller_runtime_webhook_requests_total     | Counter   | Total number of admission requests by HTTP status code.                                                                                                                                                                              |
+| controller_runtime_webhook_requests_in_flight | Gauge     | Current number of admission requests being served.                                                                                                                                                                                   |
```
camilamacedo86, line 100: these 3 are defined in: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/internal/metrics/metrics.go#L28-L60

docs/book/src/reference/metrics.md
```diff
@@ -73,6 +76,29 @@ for the metrics exported from the namespace where the project is running
...
+| controller_runtime_reconcile_errors_total     | Counter   | Total number of reconciliation errors per controller.                                                                                                                                                                                |
+| controller_runtime_reconcile_time_seconds     | Histogram | Length of time per reconciliation per controller.                                                                                                                                                                                    |
+| controller_runtime_max_concurrent_reconciles  | Gauge     | Maximum number of concurrent reconciles per controller.                                                                                                                                                                              |
+| controller_runtime_active_workers             | Gauge     | Number of currently used workers per controller.                                                                                                                                                                                     |
```
camilamacedo86, line 97: These 4 ones are defined in (https://github.com/kubernetes-sigs/controller-runtime/blob/1e4d87c9f9e15e4a58bb81909dd787f30ede7693/pkg/internal/controller/metrics/metrics.go#L26-L63)

docs/book/src/reference/metrics.md
```diff
@@ -73,6 +76,29 @@ for the metrics exported from the namespace where the project is running
 
 <img width="1680" alt="Screenshot 2019-10-02 at 13 07 13" src="https://user-images.githubusercontent.com/7708031/66042888-a497da80-e515-11e9-9d77-d8a9fc1159a5.png">
 
+## Default Metrics References
+
+By default, [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime) provides controller-relative metrics in different perspectives:
+
+| Metrics name                                  | Type      | Description                                                                                                                                                                                                                          |
```
camilamacedo86, line 83: Are we not also expecting by default: https://github.com/2uasimojo/controller-runtime/blob/master/pkg/metrics/client_go_adapter.go?

docs/book/src/reference/metrics.md
```diff
@@ -73,6 +76,29 @@ for the metrics exported from the namespace where the project is running
...
+| workqueue_work_duration_seconds               | Histogram | How long in seconds processing an item from workqueue takes.                                                                                                                                                                         |
+| workqueue_unfinished_work_seconds             | Gauge     | How many seconds of work has been done that is in progress and hasn't been observed by work_duration. Large values indicate stuck threads. One can deduce the number of stuck threads by observing the rate at which this increases. |
+| workqueue_longest_running_processor_seconds   | Gauge     | How many seconds has the longest running processor for workqueue been running.                                                                                                                                                       |
+| workqueue_retries_total                       | Counter   | Total number of retries handled by workqueue.                                                                                                                                                                                        |
```
camilamacedo86, line 91: https://github.com/kubernetes-sigs/controller-runtime/blob/b704f447ea7c8f7059c6665143a4aa1f6da28328/pkg/metrics/workqueue.go#L41-L88

docs/book/src/reference/metrics.md
```diff
@@ -73,6 +76,29 @@ for the metrics exported from the namespace where the project is running
...
+| rest_client_requests_total                    | Counter   | Number of HTTP requests, partitioned by status code, method, and host.                                                                                                                                                               |
```
camilamacedo86, line 92: @Kavinjsir this one I could not find.
From where do you check it out?

docs/book/src/reference/metrics.md
```diff
@@ -73,6 +76,29 @@ for the metrics exported from the namespace where the project is running
 
 <img width="1680" alt="Screenshot 2019-10-02 at 13 07 13" src="https://user-images.githubusercontent.com/7708031/66042888-a497da80-e515-11e9-9d77-d8a9fc1159a5.png">
 
+## Default Metrics References
+
+By default, [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime) provides controller-relative metrics in different perspectives:
```
camilamacedo86, line 81: Do you think that a text like "Following the metrics which are exported and provided by controller-runtime by default:"

WDYT?

camilamacedo86 (commented):
/ok-to-test
Great work 🥇 
Just a few nits. Otherwise, that shows terrific for me.
See that I added a question.

docs/book/src/reference/metrics.md
```diff
@@ -73,6 +76,29 @@ for the metrics exported from the namespace where the project is running
 
 <img width="1680" alt="Screenshot 2019-10-02 at 13 07 13" src="https://user-images.githubusercontent.com/7708031/66042888-a497da80-e515-11e9-9d77-d8a9fc1159a5.png">
 
+## Default Metrics References
```
camilamacedo86, line 79:
```suggestion
## Exported Metrics by default
```

camilamacedo86, line 79:
```suggestion
## Default Exported Metrics References
```

varshaprasad96, line 79: Would it be better if we could add these default metrics in the reference section of the book (https://book.kubebuilder.io/reference/reference.html) instead of elaborating in the main page?

docs/book/src/reference/metrics.md
```diff
@@ -73,6 +76,29 @@ for the metrics exported from the namespace where the project is running
...
+
+| Metrics name                                  | Type      | Description                                                                                                                                                                                                                          |
+| --------------------------------------------- | :-------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| workqueue_depth                               | Gauge     | Current depth of workqueue.                                                                                                                                                                                                          |
+| workqueue_adds_total                          | Counter   | Total number of adds handled by workqueue.                                                                                                                                                                                           |
```
varshaprasad96, line 86: Are these client-go metrics under the hood (https://pkg.go.dev/k8s.io/client-go/util/workqueue#MetricsProvider) or controller-runtime specific reconciler utilizes it? If its client-go related, would be helpful to drop a link so that its easier to know where they are used.

docs/book/src/reference/metrics.md
Kavinjsir: Hey, @camilamacedo86 .
I think it's here:
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/metrics/client_go_adapter.go#L79

docs/book/src/reference/metrics.md
Kavinjsir: Hey, @varshaprasad96 , thx for reviewing!
The workqueue related metrics listed here is refer to https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/metrics/workqueue.go
More specifically, [this code block](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/metrics/workqueue.go#L47) is where I got `workqueue_adds_total`.
So it seems like controller-runtime specific?

docs/book/src/reference/metrics.md
Kavinjsir: Yeah, exactly.

docs/book/src/reference/metrics.md
Kavinjsir: 👍🏻

docs/book/src/reference/metrics.md
Kavinjsir: This sounds good!

docs/book/src/reference/metrics.md
Kavinjsir: 👍🏻

Kavinjsir: Sounds good to make it appear in the reference section! Just have some updates to make an individual page `5.11.1 Metrics Reference` for the table and get it indexed in the reference. Would you like this kind of layout?

docs/book/src/reference/reference.md
```diff
@@ -32,5 +32,8 @@
   - [Artifacts](artifacts.md)
   - [Writing controller tests](writing-tests.md)
   - [Metrics](metrics.md)
+
+      - [Metrics Reference](metrics-reference.md)
```
camilamacedo86, line 36:
```suggestion
      - [Reference](metrics-reference.md)
```
It is inside of the Metrics so we do not need to use twice the name.

docs/book/src/reference/metrics-reference.md
```diff
@@ -0,0 +1,22 @@
...
+| controller_runtime_active_workers             | Gauge     | Number of currently used workers per controller.                                                                                                                                                                                     |
+| controller_runtime_webhook_latency_seconds    | Histogram | Histogram of the latency of processing admission requests.                                                                                                                                                                           |
+| controller_runtime_webhook_requests_total     | Counter   | Total number of admission requests by HTTP status code.                                                                                                                                                                              |
+| controller_runtime_webhook_requests_in_flight | Gauge     | Current number of admission requests being served.                                                                                                                                                                                   |
```
camilamacedo86, line 22: WDYT about we point out the source code for the metrics as well? 
Also, please ensure that we use the tag from the source code so that we will keep the history and not master in the link.

docs/book/src/reference/reference.md
Kavinjsir: Gotcha 👍🏽

docs/book/src/reference/metrics-reference.md
Kavinjsir: I think it a good idea!
Was hesitating to go for it in consideration of the possible code line changes in the future.
But the tag is a nice way to resolve this 👍🏽 
..So just updated with url prefix: https://github.com/kubernetes-sigs/controller-runtime/blob/v0.11.0/...

camilamacedo86 (approved):
Great work 🥇 
I think the changes here are great enough to fly
But we can always push follow-ups and etc if we need to fix/improve that afterwards. 

Thank you for the contribution.

PR #2571, opened 2022-03-29
:sparkles:  (go/v2) : update golang version from 1.13 to 1.17

camilamacedo86:
**Description**
Upgrade the go version for the Golang plugin go/v2 from 1.13 to 1.17

**Motivation**
Ensure that we have all plugins supporting the same golang versions
Solve tech-debts
We have been updating it for go/v3 but not for go/v2

### DIFF
--- a/docs/book/src/migration/v2vsv3.md
+++ b/docs/book/src/migration/v2vsv3.md
@@ -45,7 +45,7 @@ Projects scaffolded with Kubebuilder v3 will use the `go.kubebuilder.io/v3` plug
   * Support for [controller-runtime][controller-runtime] `v0.8.3` (for `go/v2` it is `v0.6.4` and previously it was `v0.5.0`)
   * Support for [kustomize][kustomize] `v3.8.7` (for `go/v2` it is `v3.5.4` and previously it was `v3.1.0`)
   * Required Envtest binaries are automatically downloaded
-  * The minimum Go version is now `1.15` (previously it was `1.13).
+  * The minimum Go version is now `1.17` (previously it was `1.13).
 
 <aside class="note warning">
 <h1>Project customizations</h1>
--- a/testdata/project-v2-addon/go.mod
+++ b/testdata/project-v2-addon/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kubebuilder/testdata/project-v2-addon
 
-go 1.13
+go 1.17
 
 require (
 	github.com/go-logr/logr v0.1.0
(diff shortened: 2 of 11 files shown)


Kavinjsir:
Does it necessary to let `init` expect the same version?
https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v2/init.go#L39

docs/book/src/migration/v2vsv3.md
```diff
@@ -45,7 +45,7 @@ Projects scaffolded with Kubebuilder v3 will use the `go.kubebuilder.io/v3` plug
   * Support for [controller-runtime][controller-runtime] `v0.8.3` (for `go/v2` it is `v0.6.4` and previously it was `v0.5.0`)
   * Support for [kustomize][kustomize] `v3.8.7` (for `go/v2` it is `v3.5.4` and previously it was `v3.1.0`)
   * Required Envtest binaries are automatically downloaded
-  * The minimum Go version is now `1.15` (previously it was `1.13).
+  * The minimum Go version is now `1.17` (previously it was `1.13).
```
everettraven, line 48: Nit: Should probably change the "previously it was 1.13" --> "previously it was 1.15"

```suggestion
  * The minimum Go version is now `1.17` (previously it was `1.15`).
```

everettraven (commented):
I took a look and it seems pretty straightforward. I only had one nit.

Since I am new to the Kubebuilder community and am still learning I will leave explicit approvals to others.

docs/book/src/migration/v2vsv3.md
erikgb:
```suggestion
  * The minimum Go version is now `1.17`.
```
What is the motivation for stating what the _previous_ minimum Go version requirements?

camilamacedo86:
/hold

Do not merge this one.
I'd like to discuss with all if we should or not update the go/v2 plugin as well.

testdata/project-v2-addon/go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kubebuilder/testdata/project-v2-addon
 
-go 1.13
+go 1.17
```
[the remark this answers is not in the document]
camilamacedo86: @varshaprasad96 that was what I was looking for to know.
If we had any reason to not update the go/v2 or if we only forgot this one.

Really thank you to bring that. 
So, clearly we need to bump the go version with the k8s deps as well to avoid issues. 

In this way, I am closing this one as deferred.

camilamacedo86:
Based on the comment made by @varshaprasad96 and in the analyse of the problems faced in the past we have been matching the go versions with the k8s, controller-runtime deps to avoid issues. So that, shows that we should NOT update the go/v2. More info: https://github.com/kubernetes-sigs/kubebuilder/pull/2571#discussion_r850805369

Closing this one as deferred.

PR #2573, opened 2022-03-30
🐛 Assign config to global cfg

mrkm4ntr:
The global cfg is always nil because assigned cfg is local scoped.
v2 is already correct.
https://github.com/kubernetes-sigs/kubebuilder/blob/c0a0bb6dc05e04332cb7b460bef5f155b53770cc/pkg/plugins/golang/v2/scaffolds/internal/templates/controllers/controller_suitetest.go#L164-L165


camilamacedo86 (commented):
/ok-to-test

Could you please change the PR commit and title for:

> :bug: For go/v3 plugin: assign config to global config

That is used to generate the release notes and we need to make clear that the change is only for go/v3 plugin.  

Also, not sure if I understand the problem. You are saying that the error occurs because we create the cfg locally instead of set the value from EnvTest in the global var right?
Could we try to clarify in the code via a comment that the cfg is defined globally and also where?

camilamacedo86 (changes requested):
Just some nits for we get this one merged. 
Thank you for the contribution

mrkm4ntr:
> Also, not sure if I understand the problem. You are saying that the error occurs because we create the cfg locally instead of set the value from EnvTest in the global var right?

Indeed.

> Could we try to clarify in the code via a comment that the cfg is defined globally and also where?

Sorry, I'm not sure. The commend is required? Also for v2?

testdata/project-v3-v1beta1/controllers/suite_test.go
```diff
@@ -58,7 +58,8 @@ var _ = BeforeSuite(func() {
 		ErrorIfCRDPathMissing: true,
 	}
 
-	cfg, err := testEnv.Start()
+	var err error
+	cfg, err = testEnv.Start()
 	Expect(err).NotTo(HaveOccurred())
```
camilamacedo86, line 63: @mrkm4ntr, 

Will you be able to add a comment to this one?

camilamacedo86 (approved):
/lgtm

PR #2575, opened 2022-03-30
✨ (Kubebuilder CLI) upgrade k8s deps from 1.23.0 to 1.23.5

### LINKED ISSUE #2502: Bump k8s 1.23.3 to get bug fixes

Bump k8s 1.23.3 to get bug fixes

### Extra Labels

_No response_

NikhilSharmaWe:
Fixes #2502


camilamacedo86:
@NikhilSharmaWe, 

Just 2 nits. 

a) Since it is just a bump and when we see that the k8s dep was from MAJOR.MINOR.z to z+ we know that it is only to get bug fixes. Then, in this way the title here would be something like:

 > :sparkles: (Kubebuilder CLI) upgrade k8s deps from 1.23.9 to 1.23.5 

b) We also need to bump it for the plugins, tests and release
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/commons.go 
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/common.sh#L28
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/build/.goreleaser.yml#L47  
- and run make generate with the changes

c) Would be great we also update the k8s version used in the ci test (e2e) done with prow, see: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/kubebuilder/kubebuilder-presubmits.yaml#L40

Could you help us with that?

go.mod
```diff
@@ -13,8 +13,8 @@ require (
 	github.com/spf13/cobra v1.2.1
 	github.com/spf13/pflag v1.0.5
 	golang.org/x/tools v0.1.6-0.20210820212750-d4cc65f0b2ff
-	k8s.io/apimachinery v0.23.0 // for `kubebuilder alpha config-gen`
-	sigs.k8s.io/controller-runtime v0.11.0
+	k8s.io/apimachinery v0.23.5 // for `kubebuilder alpha config-gen`
+	sigs.k8s.io/controller-runtime v0.11.2
```
NikhilSharmaWe, line 17: @camilamacedo86 

> I think the projects /scaffold will only change the dep when we bump controller runtime.
> So, we need to bump the controller-runtime as a follow up to close the issue.

This should do this job right ? It was in this first commit too.

pkg/plugins/golang/v3/commons.go
```diff
@@ -78,7 +78,7 @@ manifests: controller-gen`
 		}
 
 		if err := util.ReplaceInFile("Makefile",
-			"ENVTEST_K8S_VERSION = 1.23",
+			"ENVTEST_K8S_VERSION = 1.23.5",
```
camilamacedo86, line 81: @NikhilSharmaWe sorry, this one we will only update when we update the plugin scaffolds 
This implementation is because we have a feature to still use v1beta1 CRD APIs and it is removed from the latest versions so that if someone uses the feature we need to replace the scaffold to try to downgrade it all.

NikhilSharmaWe: ok

NikhilSharmaWe:
https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/common.sh#L48
@camilamacedo86 Updated k8s version as suggested to 1.23.5, but some tests are failing, what could be the reason.
- I think we also need to update `KIND_K8S_VERSION` in this https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/common.sh#L48 statement.

pkg/plugins/golang/v3/commons.go
```diff
@@ -78,7 +78,7 @@ manifests: controller-gen`
 		}
 
 		if err := util.ReplaceInFile("Makefile",
-			"ENVTEST_K8S_VERSION = 1.23",
+			"ENVTEST_K8S_VERSION = 1.23.3",
```
camilamacedo86, line 81: this change will also change the plugin scaffolds so we need to do in the follow up,
See that it is failing in the test data generate check: https://github.com/kubernetes-sigs/kubebuilder/runs/5760100478?check_suite_focus=true#step:6:895

NikhilSharmaWe: Sorry, by mistake.

test/common.sh
```diff
@@ -25,7 +25,7 @@ function convert_to_tools_ver {
   # Tests in 1.20 and 1.21 with their counterpart version's apiserver.
   "1.20"|"1.21") echo "1.19.2";;
   "1.22") echo "1.22.1";;
-  "1.23") echo "1.23.3";;
+  "1.23") echo "1.23.5";;
```
camilamacedo86, line 28: @NikhilSharmaWe 
Sorry, this one will need to keep in a follow-up. That was my mistake . Could we revert?
See that we are facing an issue because we are trying to get kubebuilder-tools-1.23.5-darwin-amd64.tar.gz

NikhilSharmaWe: Yes, I was thinking that too

NikhilSharmaWe: Done

camilamacedo86 (approved):
That shows great 🥇 

/approved

rashmigottipati (approved):
/lgtm

rashmigottipati:
@NikhilSharmaWe when you submit a PR to update `KIND_K8S_VERSION`, you need to update the following:
[line 28](https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/common.sh#L28) and [line 48](https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/common.sh#L48)

Issue #2567, opened 2022-03-27
Unit tests are not working locally (go 1.17.5)


camilamacedo86:
### What broke? What's expected?

The make test is not working locally
That fails when we run `go test -race ./pkg/... `

### Reproducing this issue

It was reported on the channel that someone was not able to run the tests.
the tests are called in the CI and works see: https://github.com/kubernetes-sigs/kubebuilder/runs/5605172452?check_suite_focus=true#step:5:1

However, I tried to run with;
- Mac os (darwin_amd64)
- Go 1.17.5

And also faced issues: 

```
--- FAIL: TestNewCommand (0.03s)
    --- FAIL: TestNewCommand/testdata/componentconfig (0.02s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4e258ea]

goroutine 23 [running]:
testing.tRunner.func1.2({0x4f3a160, 0x59b1970})
	/usr/local/Cellar/go/1.17.5/libexec/src/testing/testing.go:1209 +0x36c
testing.tRunner.func1()
	/usr/local/Cellar/go/1.17.5/libexec/src/testing/testing.go:1212 +0x3b6
panic({0x4f3a160, 0x59b1970})
	/usr/local/Cellar/go/1.17.5/libexec/src/runtime/panic.go:1047 +0x266
sigs.k8s.io/kubebuilder/v3/pkg/cli/alpha/config-gen.ControllerGenFilter.Filter({0x0}, {0x0, 0x0, 0x4071325})
	/Users/camilamacedo86/go/src/sigs.k8s.io/kubebuilder/pkg/cli/alpha/config-gen/controller-gen-filter.go:68 +0x58a
sigs.k8s.io/kustomize/kyaml/fn/framework.(*TemplateProcessor).doPreProcess(0xc00016cfc0, {0x0, 0x0, 0x0})
	/Users/camilamacedo86/go/pkg/mod/sigs.k8s.io/kustomize/kyaml@v0.10.21/fn/framework/processors.go:280 +0x17b
sigs.k8s.io/kustomize/kyaml/kio.FilterFunc.Filter(0xc000427ae0, {0x0, 0x0, 0x0})
[...]

PR #2581, opened 2022-03-30
:seedling: fix run unit test locally

camilamacedo86:
**Description**
:seedling: fix run unit test and coverage test by removing the race flag

**Why?**
The tests implemented for the config-gen alpha command are not working locally with this option:

```
--- FAIL: TestNewCommand (0.02s)
    --- FAIL: TestNewCommand/testdata/componentconfig (0.02s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4e2e04a]
```
E.g the above was using (mac os/amd64) besides working on the ci.

We might need to look for how to come back to the -race option in the future. (I'd suggest seeing if we can migrate alpha-gen to the plugin first)
This pr unblocks contributors that want to run the tests locally.

Closes; https://github.com/kubernetes-sigs/kubebuilder/issues/2567

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,12 +103,12 @@ test: test-unit test-integration test-testdata test-book test-license ## Run the
 .PHONY: test-unit
 TEST_PKGS := ./pkg/... ./test/e2e/utils/...
 test-unit: ## Run the unit tests
-	go test -race $(TEST_PKGS)
+	go test $(TEST_PKGS)
 
 .PHONY: test-coverage
 test-coverage: ## Run unit tests creating the output to report coverage
 	- rm -rf *.out  # Remove all coverage files if exists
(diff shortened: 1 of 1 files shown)


AlmogBaku:
After investigating further, I found another issue that causing this:
1. `/pkg/cli/alpha/config-gen/testdata/project` is missing a go module. Creating a module there solves the problem
2. `/pkg/cli/alpha/config-gen/controller-gen-filer.go:L67` ignoring a critical issue

Since this is especially breaking since go 1.18, I'll add this fix as part of my PR

AlmogBaku (changes requested):
After investigating further, I found another issue that causing this:
1. `/pkg/cli/alpha/config-gen/testdata/project` is missing a go module. Creating a module there solves the problem
2. `/pkg/cli/alpha/config-gen/controller-gen-filer.go:L67` ignoring a critical issue

Since this is especially breaking since go 1.18, I'll add this fix as part of my PR

camilamacedo86:
/hold 

I'd like to

Issue #2566, opened 2022-03-27
Kustomize v3 does not provide bin for mac (arm64)


camilamacedo86:
### What broke? What's expected?

In order to support go 1.18 (https://github.com/kubernetes-sigs/kubebuilder/pull/2486) we change the Makefile:

```
KUSTOMIZE = $(shell pwd)/bin/kustomize
kustomize: ## Download kustomize locally if necessary.
	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v3@v3.8.7)
```

To use the script:

```
KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
.PHONY: kustomize
kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
$(KUSTOMIZE):
	curl -s $(KUSTOMIZE_INSTALL_SCRIPT) | bash -s -- $(subst v,,$(KUSTOMIZE_VERSION)) $(LOCALBIN)
```

However, the v3 versions do not provide darwin arm64 binaries:

See https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv3.10.0. (latest)
But it is provided for v4: https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv4.5.3

Then, we need to:
a) verify if we have not another option to install kustomize v3
OR
b) revisit the option to update to kustomize v4 (it has breaking changes, see:  https://github.com/kubernetes-sigs/kubebuilder/pull/2498#issuecomment-1026698929. Also, see https://github.com/kubernetes-sigs/kubebuilder/pull/2498#issuecomment-1026772984)

### KubeBuilder (CLI) Version

master

### PROJECT version

3

### Plugin versions

```yaml
Default version 3
```

### Extra Labels

/kind regression

camilamacedo86:
@ryantking we need to check this one.
It blocks the release.

SkYNewZ:
Hi. I cannot use the kustomize alpha plugin.

```
❯ kubebuilder init --plugins=base.go/v3,kustomize/v4-alpha
Error: no plugin could be resolved with key "kustomize/v4-alpha" (plugin version is unstable, there may be an upgrade available: https://kubebuilder.io/migration/plugin/plugins.html)
❯ kubebuilder version
[...]

Issue #2503, opened 2022-02-01
Update kustomize from v3.8.7 to v4.4.1 and use replacements instead of vars


camilamacedo86:
### What do you want to happen?

Address the changes made in: https://github.com/kubernetes-sigs/kubebuilder/pull/2498/
Make clear that go/v4 only supports Kubectl >= 1.22.0
Make go/v4 as the default plugin
Deprecate go/v2 plugin

PR #2583, opened 2022-03-31
:sparkles: add kustomize/v2-alpha plugin to support kustomize versions >= v4 (pinned version 4.5.3)

camilamacedo86:
## Description
:sparkles: add kustomize/v2-alpha plugin to support kustomize versions >= v4

## Motivation

- Provides support for the new mac arch (arm64 Mac M1) (`kubebuilder init --plugins=kustomize/v2-alpha,base.go.kubebuilder.io/v3`)
- Allow users began to experiment and take the advantage of kustomize v4 without introducing breaking changes for who require to stick with v3
- Fix the single responsibility of each plugin. The Kustomize version supported by the scaffolds and used in the Makefile must be defined in the kustomize plugin and not in the golang ones 

**Notes**
- You can check the samples: project-v3-with-kustomize-v2
- You can check that it has been tested on the cluster: new test added in e2e 
- For further understanding check the documentation added with this PR

## Why can we not upgrade the kustomize stable v1 plugin to use the kustomize v4 version instead of creating a new alpha plugin?

We cannot introduce breaking changes for those who have a project scaffolded previously using kustomize/v1. See that kustomize must bump a major version v4 because it introduces breaking change and does NOT ensure backwards compatibility with the previous versions. 

So let's think about the scenario and imagine that:

- A user scaffold a project using our current kustomize/v1 then, the plugin generates the files using the syntax that works with kustomize v3 and not necessarily with v4. 
- Now we go there and start to update the (config) the kutomize files to use all advantages of v4 and no longer v3. Therefore, we do our release Kubebuilder CLI 3.Z with the changes. 

The user should NOT change their previous scaffolds to be able to still work with KB 3.Z and the plugin already used to scaffold their project kustomize/v1. See the policy: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/plugins/plugins-versioning.md:

> Changes made to plugins only require a plugin version increase if and only if a change is made to a plugin that breaks projects scaffolded with the previous plugin version. Once a plugin version vX is stabilized (it doesn't have an "alpha" or "beta" suffix), a new plugin package should be created containing a new plugin with version v(X+1)-alpha. Typically this is done by (semantically) cp -r pkg/plugins/golang/vX pkg/plugins/golang/v(X+1) then updating version numbers and paths. All further breaking changes to the plugin should be made in this package; the vX plugin would then be frozen to breaking changes.

However, now when they call the command to create api to add its manifests on the project using the go/v3 and kustomize/v1 this subcommand will begin to scaffold the files using the new syntax (valid for kustomize dep v4 and not v3 necessarily) since the project are not implemented to work with kustomize v4 which indeed the user will not have installed locally since its Makefile target should still be downloading kustomize v3 its result will be:

- Old apis scaffolded with kustomize syntax valid for v3
- New apis and webhooks scaffold with syntax valid for v4  
- Makefile target downloading any previous version used for kustomize v3 and not v4

That files scaffold with v4 today might work with v3 (not officially supported or guaranteed and etc). However, it can change a lot on the long journey since we can indeed add new features, also we do not know how other projects importing kustomize are using or taking advantage of it. Today it might work but can we ensure that will be the case tomorrow?`

**Therefore, that shows for me a breaking change since can break the behaviour of any plugin language which has been using kustomize stable plugin v1 with kustomize v3 syntax as its binary.** 

See the policy example:

> Example: You create a feature that adds a new marker to the file main.go scaffolded by init that create api will use to update that file. The changes introduced in your feature would cause errors if used with projects built with plugins go.kubebuilder.io/v2 without users manually updating their projects. Thus, your changes introduce a breaking change to plugin go.kubebuilder.io, and can only be merged into plugin version v3-alpha. This plugin's package should exist already.

> You must also add a migration guide to the [migrations](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/migrations.md) section of the Kubebuilder book in your PR. It should detail the steps required for users to upgrade their projects from vX to v(X+1)-alpha.

That shows our case here, we will need to let the users know what they need to do in their scaffolded project using kustomize/v1 with the syntax that works with the kustomize binary version v3 to then be able to work with v4 when this new kustomize/v2-alpha plugin becomes to be kustomize/v2 stable)

### Do we not bump go versions? Why it is different?

So we bump the go versions, which means any new project scaffold using a new version of the plugin bumped with go 1.18 will have the go mod using go 1.18 with its deps. However, users that scaffold using go 1.15, for example, will NOT be affected by that, they do not need to change anything to be able to work with KB 3.Z and its new version of the go/v3 plugin, for example, they still able to run `create api` and `create webhook` for the project scaffolded with any previous 3.Z version and using any previous go version supported. 

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/2566
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/2503

**Motivations**

- Support Apple Silicon (darwin/amd64)
- Begin to be able to scaffold and use the new kustomize v4 to take advantage of its features for the default scaffold ( it has more than a year )
- Begin to provide kustomize v4 with Kubebuilder ( See: https://github.com/kubernetes-sigs/kustomize/issues/4612#issuecomment-1117981050 / v4 is more than a year old now, and kustomize do not currently have a long-term support policy for v3 )

Co -author: @afritzler


docs/book/src/plugins/kustomize-v2-alpha.md
```diff
@@ -0,0 +1,111 @@
+# Kustomize v2-alpha Plugin
+
```
camilamacedo86, line 2: See here the documentation to better understand it.
Note that after we have this plugin, tools such as SDK will also be able to allow its users try out kustomize v4 and its languages plugins Helm/Ansible using the new mac arch, doing things like:

```
operator-sdk init --plugins=kustomize/v2-alpha,base.operatorframework.helm.sdk/v1
``` 
Then, we can provide all support and advantages to the users without introducing breaking changes.
So that, when the plugins become mature and we no longer need to support k8s versions < 1.22 then we might want to have new languages plugins that adopt them by default. 

c/c @jmrodri

pkg/plugins/common/kustomize/v1/plugin.go
```diff
@@ -23,6 +23,9 @@ import (
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugins"
 )
 
+// KustomizeVersion is the kubernetes-sigs/kustomize version to be used in the project
+const KustomizeVersion = "v3.8.7"
+
```
camilamacedo86, line 28: The kustomize version supported by the Kustomize plugin must leave in the Kustomize plugin. Otherwise, we are hurting concepts like single responsibility and cohesion.

pkg/plugins/golang/v3/scaffolds/init.go
```diff
@@ -97,6 +101,20 @@ func (s *initScaffolder) Scaffold() error {
...
+	kustomizev2 := kustomizecommonv2.Plugin{}
+	pluginKeyForKustomizeV2 := plugin.KeyFor(kustomizev2)
+
+	for _, pluginKey := range s.config.GetPluginChain() {
+		if pluginKey == pluginKeyForKustomizeV2 {
+			kustomizeVersion = kustomizecommonv2.KustomizeVersion
+			break
+		}
+	}
+
```
camilamacedo86, line 117: Here is where we check if the kustomize/v2-alpha plugin was used in the chain to properly define what version of the kustomize should be added to the Makefile which is generated by the plugin language.

The language plugin is not responsible for the kustomize and it needs to ensure that will pin in the Makefile target to install the bin locally the version which was used by the kustomize plugin used with.

test/e2e/v3/generate_test.go
```diff
@@ -229,3 +229,164 @@ Count int `+"`"+`json:"count,omitempty"`+"`"+`
 		_ = pluginutil.RunCmd("Update dependencies", "go", "mod", "tidy")
 	}
 }
+
+// GenerateV3 implements a go/v3(-alpha) plugin project defined by a TestContext.
+func GenerateV3WithKustomizeV2(kbc *utils.TestContext, crdAndWebhookVersion string) {
```
camilamacedo86, line 234: Here we have the test to check the new plugin scaffold working on the cluster with go/v3 one.

test/testdata/generate.sh
```diff
@@ -131,3 +131,4 @@ scaffold_test_project project-v3-multigroup
 scaffold_test_project project-v3-addon --plugins="go/v3,declarative"
 scaffold_test_project project-v3-config --component-config
 scaffold_test_project project-v3-v1beta1
+scaffold_test_project project-v3-with-kustomize-v2 --plugins="kustomize/v2-alpha,base.go.kubebuilder.io/v3"
```
camilamacedo86, line 134: This is the line where we scaffold the sample under the testdata

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go
```diff
@@ -0,0 +1,183 @@
...
+#      name: serving-cert # this name should match the one in certificate.yaml
+#      fieldPath: .metadata.namespace # namespace of the certificate CR
+#    targets:
+#      - select:
+#          kind: ValidatingWebhookConfiguration
+#        fieldPaths:
+#          - .metadata.annotations.[cert-manager.io/inject-ca-from]
+#        options:
+#          delimiter: '/'
+#          index: 0
```
afritzler, line 108: Starting with kustomize 4.5.3 you need to add `create: true` to the `options` block:

```suggestion
#        options:
#          delimiter: '/'
#          index: 0
#          create: true
```

This change is backward compatible and needs to be added to every `options` block.

afritzler (changes requested):

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go
```diff
@@ -0,0 +1,183 @@
...
+#      - select:
+#          kind: CustomResourceDefinition
+#        fieldPaths:
+#          - .metadata.annotations.[cert-manager.io/inject-ca-from]
+#        options:
+#          delimiter: '/'
+#          index: 0
```
afritzler, line 122:
```suggestion
#        options:
#          delimiter: '/'
#          index: 0
#          create: true
```

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go
```diff
@@ -0,0 +1,183 @@
...
+#      name: serving-cert # this name should match the one in certificate.yaml
+#      fieldPath: .metadata.name
+#    targets:
+#      - select:
+#          kind: ValidatingWebhookConfiguration
+#        fieldPaths:
+#          - .metadata.annotations.[cert-manager.io/inject-ca-from]
+#        options:
+#          delimiter: '/'
+#          index: 1
```
afritzler, line 136:
```suggestion
#        options:
#          delimiter: '/'
#          index: 0
#          create: true
```

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go
```diff
@@ -0,0 +1,183 @@
...
+#      - select:
+#          kind: MutatingWebhookConfiguration
+#        fieldPaths:
+#          - .metadata.annotations.[cert-manager.io/inject-ca-from]
+#        options:
+#          delimiter: '/'
+#          index: 1
```
afritzler, line 143:
```suggestion
#        options:
#          delimiter: '/'
#          index: 0
#          create: true
```

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go
```diff
@@ -0,0 +1,183 @@
...
+#      - select:
+#          kind: CustomResourceDefinition
+#        fieldPaths:
+#          - .metadata.annotations.[cert-manager.io/inject-ca-from]
+#        options:
+#          delimiter: '/'
+#          index: 1
```
afritzler, line 150:
```suggestion
#        options:
#          delimiter: '/'
#          index: 0
#          create: true
```

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go
camilamacedo86: done

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go
camilamacedo86: done

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go
```diff
@@ -0,0 +1,183 @@
...
+#      - select:
+#          kind: Certificate
+#          group: cert-manager.io
+#          version: v1
+#        fieldPaths:
+#          - .spec.dnsNames.0
+#          - .spec.dnsNames.1
+#        options:
+#          delimiter: '.'
+#          index: 0
```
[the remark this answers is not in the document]
camilamacedo86: done

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go
```diff
@@ -0,0 +1,183 @@
...
+#      - select:
+#          kind: Certificate
+#          group: cert-manager.io
+#          version: v1
+#        fieldPaths:
+#          - .spec.dnsNames.0
+#          - .spec.dnsNames.1
+#        options:
+#          delimiter: '.'
+#          index: 1
```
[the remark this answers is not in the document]
camilamacedo86: done

camilamacedo86:
/hold 

Not merged. 
It is under discussion.

camilamacedo86:
/hold

Until get oks,

jmrodri:
@camilamacedo86 okay I re-read the kustomize 4.0.0 release notes and I do believe we need a kustomize/v2 plugin.

camilamacedo86:
I am removing the 
/hold cancel 

So, since we had +1 check/reviewer he also agrees with
 https://github.com/kubernetes-sigs/kubebuilder/pull/2583#issuecomment-1095206369

rashmigottipati:
cc @AlmogBaku

aemperador:
@camilamacedo86 @afritzler @rashmigottipati what is the projected timeframe for release of this PR after merge? we have been waiting for this update to support ppc64le as well to update kustomize to v4 in an external app

everettraven (approved):
Everything looks good to me. Just one minor docs typo.

pkg/plugins/golang/v3/scaffolds/init.go
```diff
@@ -34,13 +38,17 @@ const (
 	// ControllerToolsVersion is the kubernetes-sigs/controller-tools version to be used in the project
 	ControllerToolsVersion = "v0.8.0"
 	// KustomizeVersion is the kubernetes-sigs/kustomize version to be used in the project
+	// @Deprecated. This information outgh to came from kustomize plugin
```
rashmigottipati, line 41: typo fix: 

```suggestion
	// @Deprecated. This information ought to come from kustomize plugin
```

pkg/plugins/common/kustomize/v2/plugin.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"sigs.k8s.io/kubebuilder/v3/pkg/config"
+	cfgv3 "sigs.k8s.io/kubebuilder/v3/pkg/config/v3"
+	"sigs.k8s.io/kubebuilder/v3/pkg/model/stage"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins"
+)
+
+// KustomizeVersion is the kubernetes-sigs/kustomize version to be used in the project
+const KustomizeVersion = "v4.5.3"
```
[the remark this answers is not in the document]
camilamacedo86: Hi @rashmigottipati thank you for looking on that: 
I added a comment on the plugins/go/v2 code that clarifies why we cannot update there:

```
// @Deprecate: KustomizeVersion came from the Kustomize plugin since go/v3
	// Go/v2 plugin exist only to ensure the backwards compatibility with the old scaffold
	// produced with kubebuilder >= 2.0.0 < 3.0.0. It does not really take advantage of the plugin system
	// (see that the PROJECT file does not have the plugins layout and uses the old/legacy
	// version/schema.) This plugin will be deprecated with go/v4-alpha and can be removed
	// when go/v4 stable be publish.
```

Also, see: https://github.com/kubernetes-sigs/kubebuilder/pull/2583/files#r839541438

The kustomize version should be in the kustomize plugin from go/v3 we sorted it out.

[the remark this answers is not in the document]
rashmigottipati: Thank you for the explanation. That makes sense.

rashmigottipati (approved):
Changes look good to me. 

/lgtm

PR #2585, opened 2022-04-01
✨ Made changes so that setup-envtest module version matches with the controller-runtime version instead of latest

### LINKED ISSUE #2480: Install setup-envtest module version which matching with the controller-runtime version instead of latest

**Motivation**

The concern/motivation to raise this one is that in Kubebuilder, we are: ([here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/Makefile#L113-L116)) 

```sh
ENVTEST = $(shell pwd)/bin/setup-envtest
.PHONY: envtest
envtest: ## Download envtest-setup locally if necessary.
	$(call go-get-tool,$(ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest@latest)
```

Then, a KB user scaffolds the project today. The scaffold is done with controller-runtime version X.
Afterwards, we are still producing new releases and versions. Thus, this target can begin to download a version of this module that is no longer compatible. 

**Proposed solution**

- Scaffold the package tools with the module import:

```golang
//go:build tools
// +build tools

/*
YOUR_COPYRIGHT
*/

// This package imports things required by build scripts, to force `go mod` to see them as dependencies
package tools

import (
	_ "sigs.k8s.io/controller-runtime/tools/setup-envtest"
)
```

- Change the makefile target to go install the setup-envtest mod version which matches with the controller-runtime version in the go.mod, e.g.:

```sh
ENVTEST = $(shell pwd)/bin/setup-envtest
.PHONY: envtest
envtest: ## Download envtest-setup locally if necessary.
	$(call go-get-tool,$(ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest)
```

Then, the command `go install sigs.k8s.io/controller-runtime/tools/setup-envtest` ought to install its version which  
corresponding with the C+R version used into the project. More info: https://github.com/kubernetes-sigs/controller-runtime/issues/1670

### Extra Labels

_No response_

NikhilSharmaWe:
Fixes #2480


NikhilSharmaWe:
@camilamacedo86 Could you please review the changes and explain what is wrong in my approach here.
- Added imported `sigs.k8s.io/controller-runtime/tools/setup-envtest` in tools/tools.go.
- In [pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go](https://github.com/kubernetes-sigs/kubebuilder/pull/2585/files#diff-9dbfeec577eb6e8aa170557eddee8431becbf2468883dfa1cab788e9b4411b4e) Removed @latest from 
```
$(ENVTEST): $(LOCALBIN)
	GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
```
- Ran `go mod vendor `and `make generate`.

vendor/github.com/cloudflare/cfssl/api/api.go
```diff
@@ -0,0 +1,231 @@
...
+
+// SendResponseWithMessage builds a response from the result and the
+// provided message, sets the JSON header, and writes to the
+// http.ResponseWriter.
+func SendResponseWithMessage(w http.ResponseWriter, result interface{}, message string, code int) error {
+	response := NewSuccessResponseWithMessage(result, message, code)
+	w.Header().Set("Content-Type", "application/json")
+	enc := json.NewEncoder(w)
+	err := enc.Encode(response)
+	return err
```
camilamacedo86, line 230: We do not vendoring in Kubebuilder. 
So, can you please remove the whole vendor/ dir
also, would be great we have a new pr adding it to be ignored in the .gitignore file.

camilamacedo86 (changes requested):
Remove vendoring
Then, we need to discuss if we want or not solve this problem in this way.
Why ? 
@joelanford raise a good point in another PR, if we add the tools with the import we might begin to encourage authors to use an approach that is not desirable.

camilamacedo86:
Hi @NikhilSharmaWe,

I added the info in : https://github.com/kubernetes-sigs/kubebuilder/issues/2480#issuecomment-1085671692

 Regards : https://github.com/kubernetes-sigs/kubebuilder/pull/2585#issuecomment-1085550995

The go mod vendor will be vendoring all deps and we do not want to commit that.
We do not want to work with vendoring in Kubebuilder. 

Vendoring is required usually when you want to ensure that you have the source code of all third-deps. Maybe the link: https://gist.github.com/datagrok/8577287 can help you understand it better.

PR #2587, opened 2022-04-01
✨Made changes so that setup-envtest module version matches with the controller-runtime version instead of latest

### LINKED ISSUE #2480: Install setup-envtest module version which matching with the controller-runtime version instead of latest

NikhilSharmaWe:
Fixes #2480


camilamacedo86:
We need to discuss the approach, see: https://github.com/kubernetes-sigs/kubebuilder/issues/2480#issuecomment-1085671692

/hold

PR #2596, opened 2022-04-03
✨(kustomize/v1, go/v3) : Add new LeaderElectionReleaseOnCancel as true option commented to let users know when they can use it

### LINKED ISSUE #2593: Set `LeaderElectionReleaseOnCancel` to the scaffolded main.go

Kubebuilder is generating `main.go` scripts that are getting closed immediately when the `Manager` is done.
Given that, we can add `LeaderElectionReleaseOnCancel: true` to the manager (with an explanation comment) to improve the performance of re-election.

### Extra Labels

_No response_

NikhilSharmaWe:
Fixes #2593

### DIFF
--- a/pkg/plugins/golang/v3/scaffolds/internal/templates/main.go
+++ b/pkg/plugins/golang/v3/scaffolds/internal/templates/main.go
@@ -245,6 +245,17 @@ func main() {
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
 		LeaderElectionID:       "{{ hashFNV .Repo }}.{{ .Domain }}",
+		// LeaderElectionReleaseOnCancel defines if the leader should step down voluntarily
+		// when the Manager ends. This requires the binary to immediately end when the
+		// Manager is stopped, otherwise, this setting is unsafe. Setting this significantly
+		// speeds up voluntary leader transitions as the new leader don't have to wait
+		// LeaseDuration time first.
+		//
+		// In the default scaffold provided, the program ends immediately after 
+		// the manager stops, so would be fine to enable this option. However, 
+		// if you are doing or is intended to do any operation such as perform cleanups 
+		// after the manager stops then its usage might be unsafe.
+		// LeaderElectionReleaseOnCancel: true,
 	})
 {{- else }}
 	var err error
--- a/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/manager/controller_manager_config.go
+++ b/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/manager/controller_manager_config.go
@@ -55,4 +55,14 @@ webhook:
 leaderElection:
   leaderElect: true
   resourceName: {{ hashFNV .Repo }}.{{ .Domain }}
+#   leaderElectionReleaseOnCancel defines if the leader should step down volume 
+#   when the Manager ends. This requires the binary to immediately end when the
+#   Manager is stopped, otherwise, this setting is unsafe. Setting this significantly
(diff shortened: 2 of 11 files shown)


NikhilSharmaWe:
Error logs says `vet: ./main.go:64:3: unknown field LeaderElectionReleaseOnCancel in struct literal`. Unable to understand the cause of it. The function [`NewManager`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/main.go#L68) uses type `Options` as one of the parameters, since Options is actually `[manager.Options](https://github.com/kubernetes-sigs/controller-runtime/blob/master/alias.go#L49)` and since it has [`LeaderElectionReleaseOnCancel`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L176) type in it why this error is being displayed.

AlmogBaku:
This feature wasn't supported in the `controller-runtime:v0.6.4` which is the version that the go/v2 plugin uses - therefore, you should only patch go/v3

pkg/plugins/golang/v3/scaffolds/internal/templates/main.go
```diff
@@ -238,13 +238,16 @@ func main() {
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
 {{ if not .ComponentConfig }}
+	// Setting LeaderElectionReleaseOnCancel to true prevents main.go 
+	// scripts to close when Manager is done and improves the performance of re-election.
```
NikhilSharmaWe, line 242: @AlmogBaku All tests are passing now. Could you please review this explanation comment added before moving forward.

camilamacedo86 (changes requested):
I think we need to discuss if this option should or NOT be added by default?
Also, what is the motivation for that?
What issue are you facing that you want to add it in the default scaffold.

See : https://github.com/kubernetes-sigs/kubebuilder/issues/2593#issuecomment-1087236601

/hold

NikhilSharmaWe:
@camilamacedo86 and @AlmogBaku Do we also need to add the comment here also https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3-config/main.go#L66? How can we check that it is possible via the ComponentConfig as well (since @camilamacedo86 mentioned that this is needed for this case)?

camilamacedo86:
Hi @NikhilSharmaWe, 

I think we also need to update the template used by the ComponentConfig option so that we will check its results in the sample. Why? Ensure that all have the same options, and follow the same standard making it easier for us to keep the project maintained.

NikhilSharmaWe:
@camilamacedo86 Are talking about adding the comment for this case as well?
https://github.com/kubernetes-sigs/kubebuilder/blob/43525de6224aa6381acf2733411ca383f46c26ce/pkg/plugins/golang/v3/scaffolds/internal/templates/main.go#L262

camilamacedo86:
Hi @NikhilSharmaWe, 

The change shows to be here: https://github.com/kubernetes-sigs/kubebuilder/blob/019c1854f67e3bd61af5aa834a1573e6934b2e8b/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/manager/controller_manager_config.go#L47-L57

We need to pass the option under the leader election options
That should be also commented ( since it is an yaml file to comment you will use the prefix # )
Also, it should have the same comment explanation.

camilamacedo86 (approved):
Great work 🥇 

/hold cancel

everettraven (approved):
/lgtm

camilamacedo86 (approved):
/lgtm

Issue #2594, opened 2022-04-03
Remove the `tools` package in favor of go 1.18 go install


AlmogBaku:
## Objective
Compiling the dependent binaries using a special package `tools` can be replaced with `go install` as we are doing in the generated Makefile.

## Solution:
1. Remove the `tools` dir
2. Replace the installation of APIDiff in the Makefile with `GOBIN=$(shell dirname $(GO_APIDIFF)) go install github.com/joelanford/go-apidiff@latest`

camilamacedo86:
Hi @AlmogBaku, 

Let's create an PR to do only this change
In its description, the **Motivation** can be add support to 1.18

AlmogBaku:
Since this issue depends on https://github.com/joelanford/go-apidiff/issues/15 I suggest holding that

/hold

camilamacedo86:
Hi @AlmogBaku, 

The apidiff on Kubebuilder is only used in the Ci - Github Action
 We can still using go 1.17 for its ci if the appdiff does not work with https://github.com/kubernetes-sigs/kubebuilder/pull/2561/files#diff-dd565f122600dd57c8c3c546dd80893fe3e96c09f41ef16886f908d6d0ce7b50R21 

Also, the new tools added to the project here: https://github.com/kubernetes-sigs/kubebuilder/commit/9e94fe0f67db3ba893a8288546d37b693adb0eff  shows a mistake. 

Why we need to import apidiff on the project at all?
It is not used by the CLI so that should not be imported in the go.mod at all.

AlmogBaku:
Right now, the `make apidiff` does work because it uses an updated version of `golang.org/x/tools` from our go module.

Until https://github.com/joelanford/go-apidiff/issues/15 will be fixed, removing it will prevent developers from running `make apidiff` in local environments with go 1.18.

joelanford:
@AlmogBaku, I just released [go-apidiff v0.4.0](https://github.com/joelanford/go-apidiff/releases/tag/v0.4.0). Hopefully this solves the issue!

camilamacedo86:
go forward @mjlshen . thx

PR #2601, opened 2022-04-05
:seedling: remove go-apidiff  from kubebuilder CLI deps

camilamacedo86:
**Description**
Remove the dep github.com/joelanford/go-apidiff from the required deps for the Kubebuilder CLI/api

**Motivation**
The github.com/joelanford/go-apidiff is only used for a CI check.
It helps us to verify if we are not introducing breaking changes in the project API.
The dep was added for we are able to get it and build it locally to do the test in dev env. (See the makefile target API diff)
So, it should not be added as dep of the project since the project itself does not requires it. 

**Local Test**
Screenshot 2022-04-05 at 01 28 21

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/2594


camilamacedo86:
c/c @AlmogBaku

AlmogBaku:
TBH, I'm in favor of removing the tools completely.
Especially since https://github.com/joelanford/go-apidiff/pull/16 is fixing the issue, and we would be able to use that via `go install`

camilamacedo86:
HI @AlmogBaku, 

> TBH, I'm in favor of removing the tools completely.
> Especially since https://github.com/joelanford/go-apidiff/pull/16 is fixing the issue, and we would be able to use that via go install

The PR: https://github.com/joelanford/go-apidiff/pull/16/files changes the dep of goapidiff.
So we would still need to build it locally to have the target right?
If yes, that does not solve the issue over we have as dep for the Kubebuilder CLI/API a module that is **not** used by it at all.

AlmogBaku:
1. No, we won't need to compile it locally.
After https://github.com/joelanford/go-apidiff/pull/16 will be merged and released, we'll be able to just replace the Makefile target to use `go install` and could get rid of the `tools` package

2. The CI doesn't use it anyway. The CI uses a GitHub Action which includes the compiled binary (and not the `tools` package)

camilamacedo86:
Hi @AlmogBaku, 
Regards https://github.com/kubernetes-sigs/kubebuilder/pull/2601#issuecomment-1089101545 
So that is great and make sense. Closing this one in favor of.

PR #2608, opened 2022-04-10
🌱 [ci]: Skipping jobs which are not required when the changes are in only docs

### LINKED ISSUE #2591: [ci] - Stop to run all tests when changes are only in the docs

Currently, if we change a doc we still running all test
The goal of this task is to skip when possible the tests when the change is performed only in the MD files. 

### Extra Labels

_No response_

NikhilSharmaWe:
### Description
There is no need of running all jobs when the changes are only in docs. This PR skips the github actions which are not required in those cases.

**Examples :** 
PR making changes in go files runs all tests: https://github.com/NikhilSharmaWe/kubebuilder/pull/31
PR making changes in md files skip jobs which are not required: https://github.com/NikhilSharmaWe/kubebuilder/pull/30

### Motivation
Fixes Part of: #2591

### DIFF
--- a/.github/workflows/apidiff.yml
+++ b/.github/workflows/apidiff.yml
@@ -6,11 +6,29 @@ on:
   pull_request:
 
 jobs:
+  check_docs_only:
+    name: check_docs_only
+    runs-on: ubuntu-18.04
+    outputs:
+      skip: ${{ steps.check_docs_only.outputs.skip }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - id: check_docs_only
+        # Since PR's are squashed prior to merging to the branch checked out (default branch),
+        # HEAD^ will resolve to the previous point in history.
+        run: |
+          REF="HEAD^"
+          [[ -z "${{ github.base_ref }}" ]] || REF=$(git show-ref ${{ github.base_ref }} | head -1 | cut -d' ' -f2)
+          echo "::set-output name=skip::$(test/check-docs-only.sh $REF)"
+
   go-apidiff:
     name: Verify API differences
     runs-on: ubuntu-latest
+    needs: check_docs_only
     # Pull requests from different repository only trigger this checks
-    if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository
+    if: (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository) && needs.check_docs_only.outputs.skip != 'true'
     steps:
       - name: Clone the code
         uses: actions/checkout@v2
@@ -19,7 +37,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v2
         with:
-          go-version: '1.17'
+          go-version: "1.17"
       - name: Execute go-apidiff
         uses: joelanford/go-apidiff@v0.2.0
         with:
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -5,8 +5,23 @@ on:
   push:
   pull_request:
 
-
 jobs:
+  check_docs_only:
+    name: check_docs_only
+    runs-on: ubuntu-18.04
+    outputs:
+      skip: ${{ steps.check_docs_only.outputs.skip }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - id: check_docs_only
+        # Since PR's are squashed prior to merging to the branch checked out (default branch),
+        # HEAD^ will resolve to the previous point in history.
+        run: |
+          REF="HEAD^"
+          [[ -z "${{ github.base_ref }}" ]] || REF=$(git show-ref ${{ github.base_ref }} | head -1 | cut -d' ' -f2)
+          echo "::set-output name=skip::$(test/check-docs-only.sh $REF)"
 
   test:
     name: ${{ matrix.os }}
@@ -14,17 +29,18 @@ jobs:
     strategy:
       matrix:
         os:
-        - ubuntu-latest
-        - macos-latest
+          - ubuntu-latest
+          - macos-latest
+    needs: check_docs_only
     # Pull requests from the same repository won't trigger this checks as they were already triggered by the push
-    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+    if: (github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository) && needs.check_docs_only.outputs.skip != 'true'
     steps:
       - name: Clone the code
         uses: actions/checkout@v2
       - name: Setup Go
         uses: actions/setup-go@v2
         with:
-          go-version: '1.17'
+          go-version: "1.17"
       # This step is needed as the following one tries to remove
       # kustomize for each test but has no permission to do so
       - name: Remove pre-installed kustomize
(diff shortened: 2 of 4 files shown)


NikhilSharmaWe:
@joelanford Could you please review the changes made.

NikhilSharmaWe:
@camilamacedo86 I think there is some issue here because tests are running in both cases (not e2e).
PR changing md file https://github.com/NikhilSharmaWe/kubebuilder/pull/8
PR changing go file https://github.com/NikhilSharmaWe/kubebuilder/pull/7

Could you please explain what is the issue here with the changes. Or there is some other isssue with repo settings.

.github/workflows/unit-tests.yml
```diff
@@ -43,18 +59,17 @@ jobs:
 
   coverage:
     name: Code coverage
-    needs:
-      - test
     runs-on: ubuntu-latest
+    needs: check_docs_only
```
NikhilSharmaWe, line 63: @camilamacedo86 Now the PR is working, I was missing this statement. Just wanted to ask should we remove statements 46-47 where we declared needs to `test`, since it was not used anywhere so I removed it, is it reasonable?

.github/workflows/unit-tests.yml
```diff
@@ -43,18 +59,17 @@ jobs:
 
   coverage:
     name: Code coverage
-    needs:
-      - test
```
camilamacedo86, line 47: Why we remove this one? Could we revert?

test/check-docs-only.sh
```diff
@@ -0,0 +1,44 @@
...
+  "(changelog/)"
+  "(OWNERS)"
+  "(OWNERS_ALIASES)"
+  "(LICENSE)"
+)
+
+if ! git diff --name-only $1 | grep -qvE "$(IFS="|"; echo "${DOC_PATTERNS[*]}")"; then
+  echo "true"
+  exit 0
+fi
```
camilamacedo86, line 44: Nit could you please give space at the end of the file?

.github/workflows/unit-tests.yml
NikhilSharmaWe: I thought we need `needs: check_docs_only` in the file and since `needs: - test` was not used, I thought we could remove it.

camilamacedo86 (commented):
All sorted out.

test/check-docs-only.sh
```diff
@@ -0,0 +1,44 @@
...
+# Patterns to ignore.
+declare -a DOC_PATTERNS
+DOC_PATTERNS=(
+  "(\.md)"
+  "(\.MD)"
+  "(\.png)"
+  "(\.pdf)"
+  "(netlify\.toml)"
+  "(website/)"
+  "(changelog/)"
```
camilamacedo86, line 35: we do not have changelog/ dir so that seems that we should not have this line

test/check-docs-only.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# Patterns to ignore.
+declare -a DOC_PATTERNS
+DOC_PATTERNS=(
+  "(\.md)"
+  "(\.MD)"
+  "(\.png)"
+  "(\.pdf)"
+  "(netlify\.toml)"
+  "(website/)"
```
camilamacedo86, line 34: we do not have website/ dir as well. I shows that the dir correct for Kubebuilder in this case is the doc/ dir

test/check-docs-only.sh
```diff
@@ -0,0 +1,44 @@
...
+: ${1?"the first argument must be set to a commit-ish reference"}
+
+# Patterns to ignore.
+declare -a DOC_PATTERNS
+DOC_PATTERNS=(
+  "(\.md)"
+  "(\.MD)"
+  "(\.png)"
+  "(\.pdf)"
+  "(netlify\.toml)"
```
camilamacedo86, line 33: Have we this file?

.github/workflows/apidiff.yml
```diff
@@ -6,11 +6,29 @@ on:
...
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - id: check_docs_only
+        # Since PR's are squashed prior to merging to the branch checked out (default branch),
+        # HEAD^ will resolve to the previous point in history.
+        run: |
+          REF="HEAD^"
+          [[ -z "${{ github.base_ref }}" ]] || REF=$(git show-ref ${{ github.base_ref }} | head -1 | cut -d' ' -f2)
+          echo "::set-output name=skip::$(test/check-docs-only.sh $REF)"
```
camilamacedo86, line 24: do we need to duplicate the code OR could it be centralized in one test only? Did we check that?

.github/workflows/apidiff.yml
```diff
@@ -6,11 +6,29 @@ on:
   pull_request:
```
camilamacedo86, line 7: WDYT about we update the doc: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#where-the-ci-tests-are-configured bu adding a new section such as "Skipped jobs" and explaining what are the jobs which are or not skipped when we identify that the only docs changed?

test/check-docs-only.sh
NikhilSharmaWe: Yes

test/check-docs-only.sh
```diff
@@ -0,0 +1,42 @@
...
+  "(OWNERS)"
+  "(OWNERS_ALIASES)"
+  "(LICENSE)"
```
camilamacedo86, line 36: Should we not ignore here any change in the doc/* directory?

@NikhilSharmaWe

NikhilSharmaWe:
@camilamacedo86 Is it fine now?

.github/workflows/apidiff.yml
camilamacedo86: @NikhilSharmaWe could we not:
1 - have a job just to check the skip with this code added here
2- make the others GitHub action only check its results

Why?
Then we have this logic centralized and not duplicated across all GitHub actions.

NikhilSharmaWe:
If the approach used to prevent the duplication of the job `check_docs_only` is correct:

After the merge we can change we can change this statement https://github.com/kubernetes-sigs/kubebuilder/blob/647be49c440b42501d9eda63919e4062a98a7bf9/.github/workflows/apidiff.yml#L10
to `kubernetes-sigs/kubebuilder/.github/workflows/check_docs_only.yml@master`. Or should we do this before merge itself, please inform.

.github/workflows/apidiff.yml
```diff
@@ -6,11 +6,15 @@ on:
...
 
 jobs:
+  check_docs_only:
+    uses: NikhilSharmaWe/kubebuilder/.github/workflows/check_docs_only.yml@checkDocsOnly
```
camilamacedo86, line 10: We cannot merge it with `NikhilSharmaWe/kubebuilder`
Also, it shows a path valid only in your local env/machine

camilamacedo86 (changes requested):
We need to fix the path

NikhilSharmaWe:
@camilamacedo86 I used `NikhilSharmaWe/kubebuilder/.github/workflows/check_docs_only.yml@checkDocsOnly` because it will show an error if I used kubernetes-sigs master (since there is no workflow check_docs_only.yml present right now). 

Now I have changed it to `kubernetes-sigs/kubebuilder/.github/workflows/check_docs_only.yml@master`, it is showing an error but I think once we merge it, it will work fine.

Please inform me If I am unreasonable here.

.github/workflows/apidiff.yml
```diff
@@ -6,11 +6,15 @@ on:
   pull_request:
 
 jobs:
+  check_docs_only:
+    uses: kubernetes-sigs/kubebuilder/.github/workflows/check_docs_only.yml@master
```
camilamacedo86, line 10: we cannot move in this way. 
That would mean that it would not work in other branches. 

@NikhilSharmaWe 
/hold

.github/workflows/apidiff.yml
camilamacedo86: We discussed that it could be an improvement in a follow up as well.

.github/workflows/apidiff.yml
```diff
@@ -19,7 +36,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v2
         with:
-          go-version: '1.17'
+          go-version: "1.17"
```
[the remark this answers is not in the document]
camilamacedo86: sorted

camilamacedo86 (approved):
The change on this PR will result in skipping the following jobs when only docs files were changed.
Screenshot 2022-04-22 at 09 02 22

The main follow-ups that I can see here show to be we begin to skip the e2e tests done in the prow when only doc files are changed. So that, if the change is only on the docs we do not need to run all e2e tests unnecessarily.

everettraven (approved):
/lgtm

rashmigottipati (approved):

PR #2619, opened 2022-04-14
🐛 For go/v3 plugin: Avoid to add duplicated code fragments

mrkm4ntr:
Every time we create webhook for existing project, the following duplicated code are injected.
```go
err = admissionv1beta1.AddToScheme(scheme)
```

### DIFF
--- a/pkg/plugins/golang/v3/scaffolds/internal/templates/api/webhook_suitetest.go
+++ b/pkg/plugins/golang/v3/scaffolds/internal/templates/api/webhook_suitetest.go
@@ -57,6 +57,7 @@ func (f *WebhookSuite) SetTemplateDefaults() error {
 
 	f.TemplateBody = fmt.Sprintf(webhookTestSuiteTemplate,
 		machinery.NewMarkerFor(f.Path, importMarker),
+		admissionImportAlias,
 		machinery.NewMarkerFor(f.Path, addSchemeMarker),
 		machinery.NewMarkerFor(f.Path, addWebhookManagerMarker),
 		"%s",
@@ -113,7 +114,6 @@ func (f *WebhookSuite) GetCodeFragments() machinery.CodeFragmentsMap {
 
 	// Generate add scheme code fragments
 	addScheme := make([]string, 0)
-	addScheme = append(addScheme, fmt.Sprintf(addschemeCodeFragment, admissionImportAlias))
 
 	// Generate add webhookManager code fragments
 	addWebhookManager := make([]string, 0)
@@ -196,6 +196,9 @@ var _ = BeforeSuite(func() {
 
 	scheme := runtime.NewScheme()
 	err = AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	err = %s.AddToScheme(scheme)
 	Expect(err).NotTo(HaveOccurred())
 
 	%s
--- a/testdata/project-v3/api/v1/webhook_suite_test.go
+++ b/testdata/project-v3/api/v1/webhook_suite_test.go
@@ -84,12 +84,6 @@ var _ = BeforeSuite(func() {
 	err = admissionv1beta1.AddToScheme(scheme)
 	Expect(err).NotTo(HaveOccurred())
 
-	err = admissionv1beta1.AddToScheme(scheme)
-	Expect(err).NotTo(HaveOccurred())
-
-	err = admissionv1beta1.AddToScheme(scheme)
-	Expect(err).NotTo(HaveOccurred())
-
 	//+kubebuilder:scaffold:scheme
 
 	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme})
--- a/testdata/project-v3-config/api/v1/webhook_suite_test.go
+++ b/testdata/project-v3-config/api/v1/webhook_suite_test.go
@@ -84,9 +84,6 @@ var _ = BeforeSuite(func() {
 	err = admissionv1beta1.AddToScheme(scheme)
 	Expect(err).NotTo(HaveOccurred())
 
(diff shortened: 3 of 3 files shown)


testdata/project-v3/api/v1/webhook_suite_test.go
```diff
@@ -81,17 +81,9 @@ var _ = BeforeSuite(func() {
...
+	err = admissionv1.AddToScheme(scheme)
 	Expect(err).NotTo(HaveOccurred())
 
-	err = admissionv1beta1.AddToScheme(scheme)
-	Expect(err).NotTo(HaveOccurred())
-
-	err = admissionv1beta1.AddToScheme(scheme)
-	Expect(err).NotTo(HaveOccurred())
-
-	//+kubebuilder:scaffold:scheme
```
camilamacedo86, line 93: We are removing `//+kubebuilder:scaffold:scheme` the tool will not able to inject the code here.
Could you please let us know why it is required?

testdata/project-v3/api/v1/webhook_suite_test.go
```diff
@@ -28,7 +28,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
-	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+	admissionv1 "k8s.io/api/admission/v1"
```
camilamacedo86, line 31: Is your goal here to begin to use "k8s.io/api/admission/v1" instead of "k8s.io/api/admission/v1beta1"?

pkg/plugins/golang/v3/scaffolds/internal/templates/api/webhook_suitetest.go
```diff
@@ -198,7 +189,8 @@ var _ = BeforeSuite(func() {
 	err = AddToScheme(scheme)
 	Expect(err).NotTo(HaveOccurred())
 
-	%s
+	err = admission%s.AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
```
camilamacedo86, line 193: The changes here do not seem required for that.
We have the marker to do it. See that it was removed here https://github.com/kubernetes-sigs/kubebuilder/pull/2619/files#r850232507

camilamacedo86 (commented):

Following some questions:
a) Is your goal here to begin to use "k8s.io/api/admission/v1" instead of "k8s.io/api/admission/v1beta1"?

b) Why it is a bug? Could you please clarifies? what are the problems faced because of this? Could please raise an issue for that? Then, we can properly check it. WDYT? 

Additionally, see that:  when we change any scaffold we need to use the command `make generate` to update the samples under the testdata dir.

mrkm4ntr:
>  a) Is your goal here to begin to use "k8s.io/api/admission/v1" instead of "k8s.io/api/admission/v1beta1"?

Sorry, my main goal is deduplicate following code. This is added again and again per every kubebuilder create webhook.
```go
err = admissionv1beta1.AddToScheme(scheme)
```
I will change the PR title and commit message.

I also addressed this TODO message, but I will split this PR from this.
```
// TODO: admission webhook versions should be based on the input of the user. For More Info #1664
```

testdata/project-v3/api/v1/webhook_suite_test.go
mrkm4ntr: I think we don't need marker here because no adding scheme is needed to be injected after first scakffolding. If it is needed for future, I will restore this.

mrkm4ntr:
> Additionally, see that: when we change any scaffold we need to use the command make generate to update the samples under the testdata dir.

Of course, I already executed `make generate`.

testdata/project-v3/api/v1/webhook_suite_test.go
camilamacedo86: It is a breaking change and cannot be addressed.
It would make the feature stop working for those who already scaffold the projects using go/v3 and want to create an api. So, could you please revet this one and ensure that we will still be using a marker with the changes proposed? 

For further information check the plugin versioning policy: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/plugins/plugins-versioning.md

camilamacedo86: /hold

testdata/project-v3/api/v1/webhook_suite_test.go
camilamacedo86: Could you please clarify that in the description?
We also need to check when the k8s.io/api/admission/v1 was introduced, deprecated and will no longer be served. Could you please check the k8s docs and add on the first comment?

Additionally, what happens if I have a project scaffold with the previous versions using  v1beta1 and then I upgrade the Kubebuilder for a future release with these changes and began to scaffold using v1 for the new api/webhooks create? Will I face any issue in having apis/webhooks using v1beta1 and the others ( new scaffold using v1? Should we address any recommendations for users to migrate/update their projects? What are the steps?

mrkm4ntr: As I described [here](https://github.com/kubernetes-sigs/kubebuilder/pull/2619#issuecomment-1098989462), I changed this PR is not for this change.

testdata/project-v3/api/v1/webhook_suite_test.go
mrkm4ntr: I restored the marker.

mrkm4ntr:
@camilamacedo86 Thank you for your comments. I changed the goal of this PR. PTAL.

camilamacedo86 (approved):
Thank you for the contribution and for work on the review of this PR 🥇 
It makes sense for me 👍 

/approved
/ok-to-test

PR #2516, opened 2022-02-20
:sparkles: [build] Add support for darwin/arm64 arch

dntosas:
### Description
More and more users are now using M1 chips for development so publishing
kubebuilder-tools in darwin/arm64 arch will enable them to use the
project natively on their machines without needing to apply custom ways
of solving missing dependencies.

### Motivation
Have seen many issues on different projects that utilize `kubebuilder` regarding missing
binaries in `darwin/arm64` arch so I thought to push this. Example Project: `operator-sdk`, `cluster-api`, etc.

### Fixes
- https://github.com/kubernetes-sigs/kubebuilder/issues/2505
- https://github.com/kubernetes-sigs/controller-runtime/issues/1657
- https://github.com/kubernetes-sigs/controller-runtime/issues/1809

Issue #1932, opened 2021-01-08
Support for arm macOS


codeblooded:
Now that docker and go have some support on arm macOS, do we plan to support development using kubebuilder on Apple Silicon?

----

Notes, add the following to the raised ticket to clarify its status:
To support Apple Silicon arm64, we need to:

- [x] : Begin to produce the Kubebuilder bin for this format **( done since the 3.4.0 release )**

However, we still need to:

- [x] :Produce the kubebuilder-tools for this format see: (https://storage.googleapis.com/kubebuilder-tools). For we are able to address this need, the issue: https://github.com/kubernetes-sigs/kubebuilder/issues/2664 was created.  (we need to have etcd binary for this architecture, see: https://github.com/operator-framework/operator-sdk/issues/5654)
- [ ] :Have kustomize v3 bin in this architecture (see: https://github.com/kubernetes-sigs/kustomize/issues/4612) 
AND/OR (no longer required with the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/2758)
- [x] : be able to use kustomize v4. See the proposal PR with kustomize v4:  https://github.com/kubernetes-sigs/kubebuilder/pull/2583  **( will be available from the next release 3.5)**

PS.: We also might need to build the images with the arch/so, see: https://github.com/kubernetes-sigs/kubebuilder/issues/2695

/kind feature

camilamacedo86:
Hi @codeblooded, 

I understand that what you are looking for is to kubebuilder provide kubebuilder_darwin_arm64 binaries as well. Am I right? It is not provided currently(https://github.com/kubernetes-sigs/kubebuilder/releases).

codeblooded:
@camilamacedo86 Yes, I assume this will be blocked by the release of [Go 1.16](https://tip.golang.org/doc/go1.16) which adds support for darwin arm64. I just wanted to create the bug and see if it is a target that kubebuilder intends to support. 😄

Adirio:
Just a comment, to include it in the automatic release process, we would also probably need to include it in the CI and I don't think this arch is supported by neither Prow nor GH Actions (except hosting our own runners).

codeblooded:
That's true. I'm curious to see what they do for kubernetes/kubectl#988 which still needs triage. 

I'm sure actions will eventually support it, but I haven't seen a bug for it. I know some other CNCF projects are looking at or already using MacStadium. They now have racks of m1 minis in their data centers.

[...]

PR #2632, opened 2022-04-23
fix testenv docs

shintard:
## Description:
I was trying kubebuilder the other day in an Apple Sillicon m1 mac arm64 environment and found that it did not work when run as documented.
From the issues and PR it looks like these issues have been resolved.
I think the documentation needs to be meticulous so as not to discourage first-time users from doing the tutorial.
This is my first commit to kubebuilder, so please let me know if there is anything I can do to help.
Also, if there is anything else in the documentation that you would like to see changed, I will help.

tied [issue](https://github.com/kubernetes-sigs/kubebuilder/issues/2389)

## Motivation
https://github.com/kubernetes-sigs/controller-runtime/issues/1657
https://github.com/kubernetes-sigs/kubebuilder/pull/2516

### DIFF
--- a/docs/book/src/cronjob-tutorial/running.md
+++ b/docs/book/src/cronjob-tutorial/running.md
@@ -72,6 +72,8 @@ make docker-build docker-push IMG=<some-registry>/<project-name>:tag
 make deploy IMG=<some-registry>/<project-name>:tag
 ```
 
+If you are using an Apple Sillicon M1 Mac, you need to follow the instructions [here](/reference/envtest.md#installation).
+
 <aside class="note">
 <h1>registry permission</h1>
 
--- a/docs/book/src/reference/envtest.md
+++ b/docs/book/src/reference/envtest.md
@@ -19,6 +19,14 @@ export K8S_VERSION=1.21.2
 curl -sSLo envtest-bins.tar.gz "https://go.kubebuilder.io/test-tools/${K8S_VERSION}/$(go env GOOS)/$(go env GOARCH)"
 ```
 
+However, in the case of M1 Mac using AppleSillicon, the arm64 binary is not yet available, so the test in the Makefile must be changed.
+```makefile
+test: manifests generate fmt vet envtest ## Run tests.
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) --arch=amd64 use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
+```
+
+This way, you can use intel binaries. downloaded intel binaries are executed via Rosseta2.
+
 Then install them:
 
 ```sh


docs/book/src/reference/envtest.md
```diff
@@ -19,6 +19,14 @@ export K8S_VERSION=1.21.2
 curl -sSLo envtest-bins.tar.gz "https://go.kubebuilder.io/test-tools/${K8S_VERSION}/$(go env GOOS)/$(go env GOARCH)"
 ```
 
+However, in the case of M1 Mac using AppleSillicon, the arm64 binary is not yet available, so the test in the Makefile must be changed.
+```makefile
+test: manifests generate fmt vet envtest ## Run tests.
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) --arch=amd64 use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
```
camilamacedo86, line 25: is the amd64 working in the AppleSillicon?

docs/book/src/cronjob-tutorial/running.md
```diff
@@ -72,6 +72,8 @@ make docker-build docker-push IMG=<some-registry>/<project-name>:tag
 make deploy IMG=<some-registry>/<project-name>:tag
 ```
 
+If you are using an Apple Sillicon M1 Mac, you need to follow the instructions [here](/reference/envtest.md#installation).
```
camilamacedo86, line 75: the project does not support arm64 yet.
It is not the only place that would need to receive some tweaks to work with probably.

docs/book/src/reference/envtest.md
shintard: As you say, there is no kubebuilder-tools supporting darwin/arm64, so using this flag will allow M1Mac to run binaries for intel via Rosseta. (Temporary solution)

docs/book/src/cronjob-tutorial/running.md
shintard: I think it's a very good idea to post this description in the FAQ section!
Can we maintain the documentation that way?

docs/book/src/reference/envtest.md
camilamacedo86: We have them now :-) 
With the latest release : https://storage.googleapis.com/kubebuilder-tools

See for 1.23.5

docs/book/src/cronjob-tutorial/running.md
camilamacedo86: We have the wish to create the FAQ for too long and we have other things to add. 
I think that it is also a nice one to go there.

docs/book/src/reference/envtest.md
camilamacedo86: Hi @shintard, 

I checked all that is missing for we support the Apple Silicon and I updated the comment: https://github.com/kubernetes-sigs/kubebuilder/issues/1932#issuecomment-1112564792

camilamacedo86 (changes requested):

The goal of this PR is to let users know how to use Kubebuilder with Apple Silicon which is not supported. We have for too long the desire to have a FAQ section, so if we would move forward with IMO that would fit in well under this section. 

Also, see here what we need to do support Apple Silicon: https://github.com/kubernetes-sigs/kubebuilder/issues/1932#issuecomment-1112564792

We are unable to build the kubebuilder-tools for darwin/arm64 because we do have not this architecture for etcd. See the issue raised there: https://github.com/etcd-io/etcd/issues/14001

Also, we have a problem with kustomize. We cannot upgrade the default layout to use kustomize v4 because from v3 to v4 it is a major bump with breaking changes. We will need to have a new kustomize plugin for that see: https://github.com/kubernetes-sigs/kubebuilder/pull/2583

camilamacedo86:
Closing this one because asap we have the etcd bin https://github.com/etcd-io/etcd/issues/14001#issuecomment-1149611951 and from Kubebuilder 3.5.* release we will be able to have all working on for Apple Silicon

Issue #2634, opened 2022-04-23
Plugins do not honor the `path` defined in the PROJECT file


AlmogBaku:
### What broke? What's expected?

When I configure a different path for my resource other than the default `./api` path, the plugins do not honor it, and expect the default path.

I found that it impacting the following:
1. go/v3: Creating a webhook for an existing API
3. Kustomize: Creating a new webhook

### Reproducing this issue

1. change https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/PROJECT#L27 this line to 

### KubeBuilder (CLI) Version

master

### PROJECT version

3

### Plugin versions

```yaml
- go.kubebuilder.io/v3
```

camilamacedo86:
You are looking here for the plugins to be able to work with different paths on the scaffolds and it is not supported. 
To provide a solution that allows users to customize the default layout path we would need to have a doc proposal for it and see how they can be defined and configure the paths for each scaffold dir. What we would do when the change/config is provided for a project that was scaffolded already and etc. 

So, this requirement is a duplication of: https://github.com/kubernetes-sigs/kubebuilder/issues/932.

camilamacedo86:
Could you please describe the steps to reproduce this issue? 
What do you mean with:
> When I configure a different path for my resource other than the default ./api path

We have only 2 possible paths the default layout and the multi-group layout. 
So, what are the circumstances where the path is not the default one in this case?

camilamacedo86:
Hi @AlmogBaku, 

Thank you for providing the info and the steps:

> clone https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v3/
> move the api package to a different location (i.e., pkg/api)
> update it accordingly in the PROJECT file
> try to generate a new webhook

This operation is **not** supported so we do not have a bug here. 
Kubebuilder does not support customizations of the paths and the info tracked in the PROJECT has not the purpose to allow this customization. 

Since you are looking here for the plugins to be able to work with different paths it shows a duplication of: https://github.com/kubernetes-sigs/kubebuilder/issues/932. After phase 2 users would be able to have their own plugins used with Kubebuilder that does the scaffolds in the paths that please them.

[...]

PR #2635, opened 2022-04-23
:bug: [go/v3, kustomize] Honor PROJECT defined path for resources

AlmogBaku:
This PR fixes a bug with go/v3 that doesn't honor a different path for the API that is configured in the `PROJECT` file.

## Motivation
Issue #2634


pkg/plugins/golang/v3/scaffolds/internal/templates/api/group.go
```diff
@@ -30,12 +31,19 @@ type Group struct {
...
 	machinery.BoilerplateMixin
 	machinery.ResourceMixin
+	machinery.RepositoryMixin
 }
 
 // SetTemplateDefaults implements file.Template
 func (f *Group) SetTemplateDefaults() error {
 	if f.Path == "" {
-		if f.MultiGroup {
+		if f.Resource.Path != "" {
```
camilamacedo86, line 40: Why the path for a project that has no multi-group layout enabled is not the default one?
Could you share what are the commands/steps required to reproduce the scenario which motivates you to raise this PR?

VERSIONING.md
```diff
@@ -9,7 +9,7 @@ a "CLI project".
 
 [guidelines]: https://sigs.k8s.io/kubebuilder-release-tools/VERSIONING.md
 
-## Compability
+## Compatibility
 
 Note that we generally do not support older release branches, except in
 extreme circumstances.
```
camilamacedo86, line 15: It seems not to be related to this PR.

docs/book/src/multiversion-tutorial/deployment.md
```diff
@@ -119,3 +119,5 @@ resource.version.group` is for everything else.
 [steps for troubleshooting](/TODO.md)
 
 [ref-multiver]: /reference/generating-crd.md#multiple-versions "Generating CRDs: Multiple Versions"
+
+[crd-version-pref]: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-priority "Versions in CustomResourceDefinitions"
```
camilamacedo86, line 123: It seems not to be related to this PR.

camilamacedo86 (changes requested):
Hi @AlmogBaku, 

Follwing some points. 

1) the pr has changes in files that does not seem related to its goal
2) Why are you adding [kustomize] in the title when we have no changes in the kustomize plugin here? 
3) Also, instead of [plugins] could we use (plugins): description?
4) Could you please share with us the commands/steps to reproduce the scenario in the issue? 

Thank you a lot for your commitment and work on improve kb.

docs/book/src/multiversion-tutorial/deployment.md
AlmogBaku: Oops. rebased ;-)

VERSIONING.md
AlmogBaku: Oops. rebased ;-)

pkg/plugins/golang/v3/scaffolds/internal/templates/api/group.go
AlmogBaku: I'm looking here if the path was configured in the configuration file, otherwise I revert to the default values (=not multi group)

AlmogBaku:
Steps to reproduce:
1. clone `https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v3/`
2. move the `api` package to a different location (i.e., `pkg/api`)
3. update it accordingly in the `PROJECT` file
4. try to generate a new webhook

camilamacedo86:
Hi @AlmogBaku, 

Thank you for provide the info and the steps:

> clone https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v3/
> move the api package to a different location (i.e., pkg/api)
> update it accordingly in the PROJECT file
> try to generate a new webhook

This operation is not supported so we do not have a bug here. 
Kubebuilder does not support customizations of the paths and the info tracked in the PROJECT has not the purpose to allow this customization. 

Since you are looking here for the plugins to be able to work with different paths it shows a duplication of: https://github.com/kubernetes-sigs/kubebuilder/issues/932. After phase 2 users would be able to have their own plugins used with Kubebuilder that does the scaffolds in the paths that please them.

However, if you are looking for a way to make the available plugins be able to work with customizations of the paths then that cannot be addressed ONLY for the apis and should work in general with all that is a scaffold for the plugins. In this way, this RFE would be much more complex than the changes suggested here and we would need a design doc proposal to discuss the approach. 

In this way, I am adding this info to the issue and closing this PR as deferred. 
I hope that you do not mind it and if you would like to discuss further this scenario then. we can do that via the issue.

Issue #2623, opened 2022-04-18
go version not supprot 1.18, but docs does not say it.


twilight0620:
### What broke? What's expected?

https://book.kubebuilder.io/quick-start.html#create-a-project:
go version v1.17+ (kubebuilder v3.3+)
it says go version 1.17+， but go 1.18 is not support yet, it should be explained in docs.

### KubeBuilder (CLI) Version

3.3.0

camilamacedo86:
Hi @twilight0620, 

Would you like to collab with this one and update the docs? 
Could you check if it is updated on master?

twilight0620:
ok i will add it to docs.
/assign

PR #2636, opened 2022-04-24
add info in docs: go version v1.18+ has not been supported yet

twilight0620:
go version not supprot 1.18, but docs does not say it. 
when i tried to use go v1.18.1 and kubebuilder v3.30， it has errors.
so need to add it in docs so that user can know it.

Fixes: https://github.com/kubernetes-sigs/kubebuilder/issues/2623

- 📖 (:book:): documentation


twilight0620:
/retitle  Docs: add info in docs: go version v1.18+ has not been supported yet

Issue #2639, opened 2022-04-25
InsertCode should return error when target is not found


asmacdo:
### What broke? What's expected?

https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/util/util.go#L74

If the target is not found, `strings.Index(string(contents), target)` returns -1 rather than an error, and kubebuilder proceeds to insert the code prior to the last character.

### KubeBuilder (CLI) Version

3.3.0

### PROJECT version

NA

### Plugin versions

```yaml
NA
```

camilamacedo86:
That makes sense to me. Would you like to push the fix? 
WDYT?

NikhilSharmaWe:
I would like to work on this issue. 
@camilamacedo86 Could you please inform me if I can.

NikhilSharmaWe:
I would like to work on this.
/assign

PR #2641, opened 2022-04-26
🐛(Kubebuilder lib) : Function InsertCode should return error when target string is not found in content

NikhilSharmaWe:
**Description**
Function [`InsertCode`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/util/util.go#L67-L79) should return error when `strings.Index(string(contents), target)` == -1

**Motivation**
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/2639

### DIFF
--- a/pkg/plugin/util/util.go
+++ b/pkg/plugin/util/util.go
@@ -72,6 +72,9 @@ func InsertCode(filename, target, code string) error {
 		return err
 	}
 	idx := strings.Index(string(contents), target)
+	if idx == -1 {
+		return fmt.Errorf("string %s not found in %s", target, string(contents))
+	}
 	out := string(contents[:idx+len(target)]) + code + string(contents[idx+len(target):])
 	// false positive
 	// nolint:gosec
--- a/pkg/plugin/util/testdata/exampleFile.txt
+++ b/pkg/plugin/util/testdata/exampleFile.txt
@@ -1,0 +1,1 @@
+exampleTarget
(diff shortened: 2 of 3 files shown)


camilamacedo86 (approved):
/approved

pkg/plugin/util/util.go
```diff
@@ -72,6 +72,9 @@ func InsertCode(filename, target, code string) error {
 		return err
 	}
 	idx := strings.Index(string(contents), target)
+	if idx == -1 {
+		return fmt.Errorf("string %s not found in %s", target, string(contents))
+	}
```
camilamacedo86, line 77: Could we add a test to cover this one?

asmacdo (approved):
/approve

+1 a test would be nice

asmacdo:
@NikhilSharmaWe lmk if you'd like help on the test, if you dont have the bandwidth I'm happy to add one to this PR. Looks like we do need the test since coveralls is failing

NikhilSharmaWe:
@asmacdo Could you please take a look on this commit in which I have tried to add test for the changes made in func `InsertCode`. It is showing error in unit tests.
https://github.com/NikhilSharmaWe/kubebuilder/commit/1d691bbb40e230b240610c880e8b7fe2ce6a2b7f

pkg/plugin/util/util.go
NikhilSharmaWe: https://github.com/NikhilSharmaWe/kubebuilder/commit/1d691bbb40e230b240610c880e8b7fe2ce6a2b7f

camilamacedo86 (approved):
/lgtm

PR #2650, opened 2022-04-28
:book: update book samples

camilamacedo86:
**Description**
Update book samples


rashmigottipati (approved):
/approve
/lgtm

One question: how are we generating these book samples?

PR #2653, opened 2022-04-29
🌱 Added condition in test.sh and test_e2e.sh so that prow tests are skipped when the changes are only in docs

NikhilSharmaWe:
### Description
- This PR is made for skipping prow tests by adding conditions in [test.sh](https://github.com/kubernetes-sigs/kubebuilder/blob/master/test.sh) and [test_e2e.sh](https://github.com/kubernetes-sigs/kubebuilder/blob/master/test_e2e.sh), so that they are not run when the changes are made only in docs.
- This is a follow-up PR for https://github.com/kubernetes-sigs/kubebuilder/pull/2608, which is made for skipping ci jobs. 

### Motivation
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/2591

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -16,4 +16,9 @@
 
 # prow calls this file currently, but we can just use `make test` to test
 # the set of things we want.
-make test
+CHECK_DOCS_ONLY=$("test/check-docs-only.sh")
+if [ -z "$CHECK_DOCS_ONLY" ]; then
+    make test
+    exit 0
+fi
+echo "WARNING: The tests were skipped because only changes on the docs were faced"
--- a/test_e2e.sh
+++ b/test_e2e.sh
@@ -13,5 +13,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-./test/e2e/ci.sh
+CHECK_DOCS_ONLY=$("test/check-docs-only.sh")
+if [ -z "$CHECK_DOCS_ONLY" ]; then
+    ./test/e2e/ci.sh
+    exit 0
+fi
+echo "WARNING: The tests were skipped because only changes on the docs were faced"


test.sh
```diff
@@ -16,4 +16,9 @@
 
 # prow calls this file currently, but we can just use `make test` to test
 # the set of things we want.
-make test
+CHECK_DOCS_ONLY=$("test/check_docs_only.sh")
```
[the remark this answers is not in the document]
camilamacedo86: How do you test this one since the PR https://github.com/kubernetes-sigs/kubebuilder/pull/2608, which has the script is not merged yet? Could you please add on the first comment the steps used to do the tests and its results?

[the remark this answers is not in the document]
camilamacedo86: Also, how it can work and passing in the tests if we do not have the script yet?

NikhilSharmaWe:
@camilamacedo86 Should I add `check_docs_only.sh` in PR too.

- And I want to ask how can I make prow tests run whenever a PR is made in forked kubebuilder repo, to test that `test.sh` and `test_e2e.sh` are executed or not when the changes are only in docs?

test_e2e.sh
```diff
@@ -13,5 +13,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+CHECK_DOCS_ONLY=$("test/check_docs_only.sh")
+if [ -z "$CHECK_DOCS_ONLY" ]; then
+    ./test/e2e/ci.sh
+    exit 0
+fi
```
camilamacedo86, line 21: Also, we need to log ( echo ) something like WARNING: The tests were skipped because only changes on the docs were faced.

test_e2e.sh
```diff
@@ -13,5 +13,9 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-./test/e2e/ci.sh
+CHECK_DOCS_ONLY=$("test/check_docs_only.sh")
+if [ -z "$CHECK_DOCS_ONLY" ]; then
+    ./test/e2e/ci.sh
+    exit 0
+fi
+echo "WARNING: The tests were skipped because only changes on the docs were faced"
```
camilamacedo86, line 21: I think it is missing else here otherwise will output always. Right?

camilamacedo86: Could you create 2 draft PRs against Kubebuilder for us to check if it is working ?
1 Pr should have a change that will trigger the tests and in the other, a change that will not trigger the tests.

NikhilSharmaWe:
@camilamacedo86 I do not think that the `echo "WARNING: The tests were skipped because only changes on the docs were faced"` will be executed each time, since `exit 0` is present in if condition.

And I am a bit confused that how can we check this PR by making a draft PR in kubebuilder, since https://github.com/kubernetes-sigs/kubebuilder/pull/2608 and this PR are not merged yet. Could you please explain.

test.sh
```diff
@@ -16,4 +16,9 @@
 
 # prow calls this file currently, but we can just use `make test` to test
 # the set of things we want.
-make test
+CHECK_DOCS_ONLY=$("test/check-docs-only.sh")
+if [ -z "$CHECK_DOCS_ONLY" ]; then
+    make test
+    exit 0
+fi
```
camilamacedo86, line 23: The motivation for this task is: 
- If we just change the docs, we do not need to run all e2e tests for example
- So that we do not need wait long to check the results or spend an unnecessary resource to do that 

However, see that with your test with the PR: 

a) I could not find the WARNING in for example : https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kubebuilder/2688/pull-kubebuilder-e2e-k8s-1-23-3/1525693478816714752/build-log.txt. So are we really skipping the tests? 
b) we still checking that the kind and everything were start-ups and it still consuming a time to be executed; 

Screenshot 2022-05-18 at 10 39 49

So, we did not achieve the goals/motivations so far. Also, by looking at https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/kubebuilder/kubebuilder-presubmits.yaml#L5 I begin to wonder:

If we skip here would we still be on the prow for starting a cluster? If yes, then how much do we actually win doing the change in this way? We might realise that unless we are able to skip there we do not really add value in adding this one here.

camilamacedo86 (changes requested):
we are unable to check in the tests done so far that the tests were skipped 
Also, we need to ensure that this solution would attend its motivations
More info: https://github.com/kubernetes-sigs/kubebuilder/pull/2653/files#r875691453

test.sh
camilamacedo86: See: https://stackoverflow.com/questions/23303549/what-are-commit-ish-and-tree-ish-in-git
One way to move forward here, I think is:
- create a Makefile target that will check the docs 
- the check its result here to use in the e2e 
So that we also let the check available to contributors. 

c/c @NikhilSharmaWe

camilamacedo86:
Closing this one with: https://github.com/kubernetes-sigs/kubebuilder/issues/2591#issuecomment-1138916860

Issue #2645, opened 2022-04-28
make-test doesn't return full coverage after webhook boilerplate code is generated


Sher-Chowdhury:
### What broke? What's expected?

It looks like I lose 100% unit test coverage right after I generate the webhook boilerplate code:

```
$ make test
/Users/sherchowdhury/operators/mysql-operator/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
/Users/sherchowdhury/operators/mysql-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
KUBEBUILDER_ASSETS="/Users/sherchowdhury/Library/Application Support/io.kubebuilder.envtest/k8s/1.23.3-darwin-amd64" go test ./... -coverprofile cover.out
?       github.com/Sher-Chowdhury/mysql-operator        [no test files]
?       github.com/Sher-Chowdhury/mysql-operator/api/v1 [no test files]
ok      github.com/Sher-Chowdhury/mysql-operator/controllers    6.707s  coverage: 100.0% of statements


$ kubebuilder create webhook --group wordpress --version v1 --kind Mysql --defaulting --programmatic-validation 
Writing kustomize manifests for you to edit...
Writing scaffold for you to edit...
api/v1/mysql_webhook.go
Update dependencies:
$ go mod tidy
Running make:
$ make generate
/Users/sherchowdhury/operators/mysql-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
Next: implement your new Webhook and generate the manifests with:
$ make manifests


$ make test                                                                                                     
/Users/sherchowdhury/operators/mysql-operator/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
/Users/sherchowdhury/operators/mysql-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
KUBEBUILDER_ASSETS="/Users/sherchowdhury/Library/Application Support/io.kubebuilder.envtest/k8s/1.23.3-darwin-amd64" go test ./... -coverprofile cover.out
?       github.com/Sher-Chowdhury/mysql-operator        [no test files]
ok      github.com/Sher-Chowdhury/mysql-operator/api/v1 5.663s  coverage: 3.5% of statements
[...]

PR #2668, opened 2022-05-01
🐛 Ignore file zz_generated.deepcopy.go in default scaffold for v3 test coverage

NikhilSharmaWe:
### Description
File [`zz_generated.deepcopy.go`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/api/v1/zz_generated.deepcopy.go) is autogenerated. We should ignore this file from the test coverage for solving the issue. Read [this](https://github.com/kubernetes-sigs/kubebuilder/issues/2645#issue-1218751619) and [follow-up](https://github.com/kubernetes-sigs/kubebuilder/issues/2645#issuecomment-1113963723) comment for more info.

### Motivation
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/2645

### DIFF
--- a/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
@@ -118,6 +118,14 @@ vet: ## Run go vet against code.
 test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
 
+.PHONY: test-coverage
+test-coverage: ## Run unit tests creating the output to report coverage
+	- rm -rf *.out  # Remove all coverage files if exists
+	go test -race -failfast -tags=integration -coverprofile=cover.out.tmp
+	cat cover.out.tmp | grep -v "zz_generated.deepcopy.go" > cover.out
+	go tool cover -func cover.out
+	rm cover.out.tmp
+	
 ##@ Build
 
 .PHONY: build
--- a/testdata/project-v3-addon/Makefile
+++ b/testdata/project-v3-addon/Makefile
@@ -59,6 +59,14 @@ vet: ## Run go vet against code.
 test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
 
+.PHONY: test-coverage
+test-coverage: ## Run unit tests creating the output to report coverage
+	- rm -rf *.out  # Remove all coverage files if exists
+	go test -race -failfast -tags=integration -coverprofile=cover.out.tmp
+	cat cover.out.tmp | grep -v "zz_generated.deepcopy.go" > cover.out
+	go tool cover -func cover.out
+	rm cover.out.tmp
+	
 ##@ Build
 
 .PHONY: build
--- a/testdata/project-v3-config/Makefile
+++ b/testdata/project-v3-config/Makefile
@@ -59,6 +59,14 @@ vet: ## Run go vet against code.
 test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
 
+.PHONY: test-coverage
+test-coverage: ## Run unit tests creating the output to report coverage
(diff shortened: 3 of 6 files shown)


pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
```diff
@@ -118,6 +118,13 @@ vet: ## Run go vet against code.
 test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
 
+.PHONY: test-coverage
+test-coverage: ## Run unit tests creating the output to report coverage
+	- rm -rf *.out  # Remove all coverage files if exists
+	go test -race -failfast -tags=integration -coverprofile=cover.out.tmp
+	cat cover.out.tmp | grep -v "zz_generated.deepcopy.go" > cover.out
+	tool cover -func cover.out
+
```
NikhilSharmaWe, line 127: @camilamacedo86 Could you please take a look. 
Is this approach used reasonable?

NikhilSharmaWe, line 127: @camilamacedo86 Could you please take a look.
Does removing the file from the cover file solved the issue here?
Is this approach used reasonable? Please inform.

Sher-Chowdhury: I have a few thoughts on this:

- why is the `rm` prefixed with "- "? I've never seen that before. Is that some kind of makefile syntax? If so, then what does it do?
- I think the the `tool cover` line needs to be prefixed with `go `
- I think another line can be added at the end to delete the tmp file, i.e. `rm cover.out.tmp`
- In general, not sure if this is the right approach. It feels more like a workaround, i.e. it still won't fix the output displayed in the `make test` output

Sher-Chowdhury (changes requested):

NikhilSharmaWe:
@Sher-Chowdhury @camilamacedo86 
When I change the comments on top of [zz_generated.deepcopy.go](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/api/v1/zz_generated.deepcopy.go) to:
```
//go:build (!ignore_autogenerated && !skip)
// +build !ignore_autogenerated
// +build !skip
```

and also add -tags=skip in go test command in the makefile.

But when I run `make generate`, the changes made in the comments get lost (it changes to what it was before), why is it happening

Sher-Chowdhury:
@NikhilSharmaWe I think that's because `make test` also runs `make generate` as one of the initial steps before running the actual tests - 

https://github.com/kubernetes-sigs/kubebuilder/blob/cb03a0cddc8ae511b69c32f362da2122c1aef3dc/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go#L118

And `make generate` ends up recreating the [zz_generated.deepcopy.go](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/api/v1/zz_generated.deepcopy.go) file from scratch again. 

So I think to test out your changes, you can just delete the "generate" string from your makefile.

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
```diff
@@ -118,6 +118,14 @@ vet: ## Run go vet against code.
 test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
 
+.PHONY: test-coverage
```
camilamacedo86, line 121: The coverage today is done here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/Makefile#L58-L60

So, if we create a new target only for that we need to update the make test target.
I'd prefer we change the make test target for ignoring the file

Also, I do not think that we can address this need by doing the cat/grep/rm.
We need to look for how we properly ignore files with go test for the option -coverprofile cover.out instead.
What are the available options provided by go test to do it?
Did we check that?

/hold

camilamacedo86 (changes requested):
Thank you, @NikhilSharmaWe  for your contribution 🥇 
I made a few comments. I think it still needing some changes before we get this one merged

camilamacedo86:
@NikhilSharmaWe @Sher-Chowdhury I put my POV in : https://github.com/kubernetes-sigs/kubebuilder/issues/2645#issuecomment-1126670250

So, I would suggest we close this one as defered. WDYT?
Is that make sense? Could we agree?

camilamacedo86:
Closing this one based on the comments and discussions made on the issue.
However, please feel free to reach out if you do not agree with it.

PR #2671, opened 2022-05-03
:book: Dynamic Bundle Design Proposal

everettraven:
## Description of the change
Add a design for more a more dynamic bundle design

## Motivation for the change
Make it easier to define default functionality by implementing a more dynamic style of `Bundle`


designs/dynamic-bundle.md
```diff
@@ -0,0 +1,267 @@
...
+```go
+gov3Bundle, _ := plugin.NewBundle(golang.DefaultNameQualifier, plugin.Version{Number: 3},
+	kustomizecommonv1.Plugin{},
+	golangv3.Plugin{},
+)
+```
+
+In the above example a plugin `Bundle` is created that ensures that the `kustomize/v1` plugin is run before the `go/v3` plugin.
+ 
+With the addition of Phase 2 Plugins on the way, users using the `--plugins` flag to utilize external plugins will not get the benefits of using some default functionality that previous Kubebuilder users using the default `go/v3` plugin gained via the `kustomize/v1` plugin.
```
camilamacedo86, line 18: Could you please clarify what will no longer be a benefit with phase 2?
Note that with phase 2 all still working in the same way, phase 2 only allows us to use Kubebuilder CLI to run Plugins which are implemented and built outside of its repository/binary.

So that, a Plugin by composition ( using bundle ) can still be created on Kubebuilder or outside of it. 
In Kubebuilder, we use the Bundle API only to scaffold kustomize with the golang base. However, Kubebuilder is a LIB and if you check how it is used in SDK, for example, we use the bundle API to create a plugin with a bigger chain of composition so that we are able to scaffold many helpers and customizations with, see: https://github.com/operator-framework/operator-sdk/blob/master/internal/cmd/operator-sdk/cli/cli.go#L79-L108

In this way, we can re-use the same helpers for example to integrate the projects with OLM or to add the Scorecard features for all languages base.

designs/dynamic-bundle.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Proposal
+The proposed solution is to create a new form of `Bundle` that is more dynamic and can be used to create a default bundle that contains sensible default plugins that can be used with all plugins. This could be overriden by a command line flag.
+
+## User Stories
+- As a Kubebuilder maintainer I would like to have Kubebuilder implement sane default plugins that always run before and/or after plugins, specifying this once and it applying to all uses
```
camilamacedo86, line 24: could you please suplement the use cases to follow the form: I am as ... would like to .... so that ....

designs/dynamic-bundle.md
```diff
@@ -0,0 +1,267 @@
...
+
+- As a Kubebuilder user I would like to be able to override the default Kubebuilder plugin chain, scaffolding only exactly what I feel I need
+
+## Goals
+- `kubebuilder` is able to define sane defaults for plugins that should always be run before and/or after user specified plugins
+
+- A `Bundle` no longer has to be created for each plugin to implement sane defaults
+
+## Examples
+Context: Kubebuilder defines that the `kustomize/v1` plugin should always run before any other plugin. Kubebuilder also defines the default plugin to be run if a user does not specify plugins via the `--plugins` flag is `go/v3`
```
camilamacedo86, line 36: We have the config-gen alpha command which is a solution that replaces kustomize
When it was implemented to the project the idea was users could choose the best scaffold for their needs
So that, they could have a new Golang plugin that uses kustomize or config-gen.

If we move with this idea would that mean remove config-gen and accepted that the only way to configure the project is with kustomize. Am I right? is that something that we really want? Why we would say that a specific plugin is required for any based action done with the CLI? What are pros and cons here? How that can affect others? What is the plan to deprecate and remove config-gen (if that is the case)?

designs/dynamic-bundle.md
```diff
@@ -0,0 +1,267 @@
...
+
+- `kubebuilder init`
+	- Should scaffold using the `kustomize/v1` plugin and the `go/v3` plugin
+- `kubebuilder init --plugins other.plugin/v3`
+	- Should scaffold using the `kustomize/v1` plugin and the `other.plugin/v3` plugin
+- `kubebuilder init --plugins other.plugin/v3 --override-default-plugin-chain`
+	- Should scaffold using only the `other.plugin/v3` plugin
```
camilamacedo86, line 43: In this way, we are here proposing that kubebuilder init should only scaffold the kustomize and then we would no longer have the default plugins used by the CLI?

See that we define the default plugin that should be used by Kubebuilder/SDK or any project that uses Kubebuilder CLI Lib so that when a user runs kubebuilder init, that means scaffold the default plugin is. Today it goes/v3. 

What are the advantages for kubebuilder users?
How that would make the UX and usability better?

designs/dynamic-bundle.md
```diff
@@ -0,0 +1,267 @@
...
+
+Proposed implementation:
+
+File pkg/plugin/bundle.go:
+```go
+type dynamicBundle struct {
+	bundle
+
+	beforePlugins []Plugin
+	afterPlugins  []Plugin
```
camilamacedo86, line 77: What mean beforePlugins and afterPlugins?

designs/dynamic-bundle.md
```diff
@@ -0,0 +1,267 @@
...
+```
+
+### Struct
+A new struct would be created that implements the `DynamicBundle` interface.
+
+Proposed implementation:
+
+File pkg/plugin/bundle.go:
+```go
+type dynamicBundle struct {
```
camilamacedo86, line 73: What is the motivation for the name dynamic?
When/where/why this solution is dynamic?

designs/dynamic-bundle.md
```diff
@@ -0,0 +1,267 @@
...
+
+Proposed implementation:
+
+Defining the `DynamicBundle`:
+```go
+v3Dynamic, _ := plugin.NewDynamicBundle(
+	"dynamic."+plugins.DefaultNameQualifier,
+	plugin.Version{Number: 3},
+	[]plugin.Plugin{
+		kustomizecommonv1.Plugin{},
```
camilamacedo86, line 205: What is the main motivation with DynamicBundle?
What are the advantages of using it?
What would allow us to achieve what is not possible be done with Bundle API?

camilamacedo86: Is the goal allow override the default behaviour done with kubebuilder init?
If yes, what is the difference of:

1. Create an external plugin for i.e with a bundle of kustomize/v1 and another golang v3 base fake as you did
2. Then, when we have the external plugins ( phase 2 ) call kubebuilder init --external-plugins=mynewexternalone 

That would not scaffold the plugin implemented by you outside?
Is not the Bundle ApI enough for your re-use kustomize and any other plugin of your choice to build new external plugins by composition and re-use the implementations done in Kubebuilder or any other project that follows its interfaces?

camilamacedo86 (commented):
Btw, I loved the gifs. :-) 
Some questions were added for us better understand your idea.

designs/dynamic-bundle.md
everettraven: That's a good point, and I think I may have had a slight misunderstanding that when using the `--plugins` flag that it would override a bundle but after digging in with the PoC and you pointing this out it makes sense that this functionality would still be available. 

I think the advantage of this dynamic bundle is rather that you can define the defaults that should always be run once rather than having to define multiple different bundles repeating mostly the same implementation over and over again.

I will work on updating the doc to reflect this instead.

Just for some more details on where I think the dynamic bundle is useful, consider within the Operator SDK the same lines you sent (https://github.com/operator-framework/operator-sdk/blob/master/internal/cmd/operator-sdk/cli/cli.go#L79-L108) could be simplified using the dynamic bundle to one statement like:
```go
dynamic, _ := plugin.NewDynamicBundle(
	"dynamic."+plugins.DefaultNameQualifier,
	plugin.Version{Number: 3},
	[]plugin.Plugin{
		kustomizecommonv1.Plugin{}, // here we are saying that kustomizecommonv1.Plugin should always be run before any other plugin
	},
	[]plugin.Plugin{
		golangv3.Plugin{}, // here is just a default value, but the idea is that this value can be changed if a user passed the --plugins flag
	},
	[]plugin.Plugin{ // here we are saying that the manifestsv2 and scorecardv2 plugins should always be run after all other plugins
	        manifestsv2.Plugin{},
		scorecardv2.Plugin{},
	})
```

Let me know if that helps a bit with the understanding. If not I can try to explain it a bit further.

designs/dynamic-bundle.md
everettraven: Will work on making those changes

designs/dynamic-bundle.md
rashmigottipati, line 18: In my understanding, `--plugins` field would ensure the ordering of plugins by running the plugins in order, irrespective of whether we are including internal plugins or external plugins in the plugin chain. So we will be able to ensure that the `kustomize/v1` plugin is run before the `go/v3` plugin, just like how it's run in the `Bundle` concept.

Can you clarify about what is the default functionality and why we are not able to reap those benefits when trying to run external plugins?

designs/dynamic-bundle.md
```diff
@@ -0,0 +1,267 @@
...
+
+## User Stories
+- As a Kubebuilder maintainer I would like to have Kubebuilder implement sane default plugins that always run before and/or after plugins, specifying this once and it applying to all uses
+
+- As a Kubebuilder user I expect Kubebuilder to have default functionality to make my operator development easier
+
+- As a Kubebuilder user I would like to be able to override the default Kubebuilder plugin chain, scaffolding only exactly what I feel I need
+
+## Goals
+- `kubebuilder` is able to define sane defaults for plugins that should always be run before and/or after user specified plugins
```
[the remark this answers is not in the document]
rashmigottipati: I'd like to know some examples of what these sane defaults are going to look like. Are these default applicable for plugins or kubebuilder itself? 

If it's for plugins, each plugin might be independent of the other plugins, how can we gather the initial default values used for the plugins? Do we need an API call for that to gather plugin specific info first?

designs/dynamic-bundle.md
everettraven: The idea behind the name dynamic is that there can be defaults that are defined to run before and after all plugins and the user can "inject" plugins into the chain. The idea is that there is one bundle definition for all plugins with a specific supported project  version rather than having to repeat the definition for each new language plugin.

designs/dynamic-bundle.md
camilamacedo86: @everettraven thanks a lot for taking the time to answer my silly questions.
I am still trying to understand the advantage here. For example, using SDK as a use case. Why/what do we win exactly by changing the default implementation in this direction?

designs/dynamic-bundle.md
camilamacedo86: Thank you. @everettraven it seems that all pros are already addressed with the current implementation. So, I still trying to figure out what we can do with the proposal that we cannot do today?

designs/dynamic-bundle.md
```diff
@@ -0,0 +1,267 @@
...
+	golangv3.Plugin{},
+)
+```
+
+In the above example a plugin `Bundle` is created that ensures that the `kustomize/v1` plugin is run before the `go/v3` plugin.
+ 
+With the addition of Phase 2 Plugins on the way, users using the `--plugins` flag to utilize external plugins will not get the benefits of using some default functionality that previous Kubebuilder users using the default `go/v3` plugin gained via the `kustomize/v1` plugin.
+
+## Proposal
+The proposed solution is to create a new form of `Bundle` that is more dynamic and can be used to create a default bundle that contains sensible default plugins that can be used with all plugins. This could be overriden by a command line flag.
```
rashmigottipati, line 21: Would like to understand more about the use-cases of the `DynamicBundle`.

Currently, in Kubebuilder, there are `PreScaffold` and `PostScaffold` hooks which can be used by every subcommand (init, create api, create webhook). Pre scaffold logic allows to ensure the correct go version, validating that main.go exists, so on. Similarly, post scaffold logic can execute a set of tasks after the main scaffolding is done, such as updating dependencies.  I'm wondering if these can be customized further to set the sane defaults during `PostScaffold` logic, without having to introduce new default plugins.

designs/dynamic-bundle.md
everettraven: > What is the main motivation with DynamicBundle?
> What are the advantages of using it?

I think it is probably best to point to some of my previous answers for these questions:
- Main motivation: https://github.com/kubernetes-sigs/kubebuilder/pull/2671#discussion_r865863397
- Advantages: https://github.com/kubernetes-sigs/kubebuilder/pull/2671#discussion_r865880908 and https://github.com/kubernetes-sigs/kubebuilder/pull/2671#discussion_r865896424



> Is the goal allow override the default behaviour done with kubebuilder init? If yes, what is the difference of:
> 
>     1. Create an external plugin for i.e with a bundle of kustomize/v1 and another golang v3 base fake as you did
> 
>     2. Then, when we have the external plugins ( phase 2 ) call kubebuilder init --external-plugins=mynewexternalone
> 
> 
> That would not scaffold the plugin implemented by you outside? Is not the Bundle ApI enough for your re-use kustomize and any other plugin of your choice to build new external plugins by composition and re-use the implementations done in Kubebuilder or any other project that follows its interfaces?

I think that you would be able to do it this way but that is causing more work for plugin authors to appease users when this functionality could just be done by Kubebuilder by default. I think the core of this idea is that there is the ability to define some default plugins that are run before and after all other plugins defined by a user (or a default one used). This allows for fewer `DynamicBundle` definitions that allow flexibility in the language plugins rather than defining a new `Bundle` for each and every new language plugin.

designs/dynamic-bundle.md
everettraven: This was also brought up by @camilamacedo86 previously and I answered here: https://github.com/kubernetes-sigs/kubebuilder/pull/2671#discussion_r865863397

If there are more questions after looking at that answer let me know!

designs/dynamic-bundle.md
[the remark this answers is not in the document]
everettraven: Ideally they would apply to all plugins that have the same supported project versions

designs/dynamic-bundle.md
camilamacedo86: Hi @everettraven:

Following some coments inline: 

> This is rather proposing that a dynamic bundle be implemented so that there can be default plugins that run no matter what plugins a user passes in. These examples are just meant to show what it would be like if Kubebuilder implemented this functionality.

We have the default plugin option already and it has been used
Also, we can overwrite the default plugin by --plugins informed via the flag

> In these examples, the context is that Kubebuilder wants to make sure the kustomize plugin is always run before any other plugin.

Why would we force run kustomize in all cases?
See that API is valid for any plugin/CLI and kustomize does not fit in any scenario. 
As shared before, we indeed have a proposal in place to config the projects using the config-gen approach instead of kustomize. Also, we have people that do not use kustomize at all, see: https://github.com/kubernetes-sigs/kubebuilder/pull/2350

So, if the main goal of this proposal is to force kustomize ( as it seems to be ) I do not think that we can go along with it.

designs/dynamic-bundle.md
everettraven: I think the biggest use case is to make it so that we don't have to create a new `Bundle` every time we create a new language plugin.

The sane defaults are more so default plugins that should be run. Hopefully this comment should clear it up a bit more but let me know if you need more details: https://github.com/kubernetes-sigs/kubebuilder/pull/2671#discussion_r865863397

designs/dynamic-bundle.md
everettraven: To me the advantage is that you can define one bundle implementation that will apply to all plugins that support the same project versions.

To further expand on my last comment, the example code snippet I used would make it so that you define the one dynamic bundle and the kustomize, manifests, and scorecard plugin would always be run. This means that the go/v3, helm, ansible, and hybrid plugins would all work the same as they do now in their SDK bundle definitions but with only one bundle definition rather than the existing 4 definitions.

It also makes it so that when new plugins are added there is no extra configuration work to be done.

designs/dynamic-bundle.md
```diff
@@ -0,0 +1,267 @@
+Dynamic Bundle
```
camilamacedo86, line 1: Hi @everettraven,
 
It seems that the primary goal of this proposal would be to force the kustmize plugin to be executed by the CLI always. Then, that does not seem to go along with the project vision:

a) We do not want to force the kustomize usage. See that we indeed have an approach that could replace the kustomize and it desired we be able to have an option to opt-in/opt-out by it, which is the config-gen: https://github.com/kubernetes-sigs/kubebuilder/issues/2501

b) Not all users/consumers adopt the kustomize, see: https://github.com/kubernetes-sigs/kubebuilder/pull/2671#discussion_r865915871

c) Not all plugins will use kustomize. 

d) We also can have many versions of the kustomize plugin so that if we try to force a specific plugin to be always executed, no matter what, we would hurt the CLI API and other use cases. See: https://github.com/kubernetes-sigs/kubebuilder/pull/2583 (we are looking for to add a new version of the kustomize plugin)

Additionally, currently, Kubebuilder CLI API allows already:

- Define default plugins ( which should be the default option when the subcommand tool init is called)
- Override the default option by passing one or more plugins via the flag --plugins

Then, when we have the plugin phase 2:

- The API consumer(s) will be able to create plugins via composition to re-use kustomize or not by taking advantage of the Bundle API 
- Users will be able to override the default behaviour to inform the external plugins.

Because of both reasons, I think we cannot accept this one. In this way, my vote here would be -1.
However, please let me know if I misunderstood its goals.  

And again, thank you for proposing the idea and suggestion.

c/c @rashmigottipati @jmrodri

designs/dynamic-bundle.md
everettraven: > The API allows define the default plugins for each project version indeed, see: https://github.com/operator-framework/operator-sdk/blob/master/internal/cmd/operator-sdk/cli/cli.go#L122-L124

My understanding is that those plugins are ones to be run if a user does not use the `--plugins` flag. 

The proposal here is to implement a bundle that no matter what plugins a user passes in with the `--plugins` flag certain plugins are still run unless the user specifically wants to override that functionality. I think of it more as a way to have only one bundle definition that will apply for multiple plugins rather than a new bundle definition for each plugin to implement the same defaults.

designs/dynamic-bundle.md
camilamacedo86: @everettraven , so what you are trying to do I think would fit in the CLI API, see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/cli.go

That would help by new methods that allow us to inform preScaffoldDefaultPlugins and posScaffoldDefaultPlugins for the CLI so that each CLI could have its pre or post scaffold ones by default an option.

designs/dynamic-bundle.md
everettraven: > It seems that the primary goal of this proposal would be to force the kustmize plugin to be executed by the CLI always.

This is not the goal of this proposal. The goal is to create a new Bundle format that allows for one definition that can apply to multiple plugins.

Some more information:
1. Kubebuilder does not have to force the kustomize usage with this option. The default plugins can just be left empty.
2. **If** Kubebuilder did force kustomize usage it would still be possible to override with another flag.

> Additionally, currently, Kubebuilder CLI API allows already:
> 
>     * Define default plugins ( which should be the default option when the subcommand tool init is called)
> 
>     * Override the default option by passing one or more plugins via the flag --plugins
> 
> 
> Then, when we have the plugin phase 2:
> 
>     * The API consumer(s) will be able to create plugins via composition to re-use kustomize or not by taking advantage of the Bundle API
> 
>     * Users will be able to override the default behaviour to inform the external plugins.

My understanding is that the default plugins are just ones that are run unless a user specifies otherwise with the `--plugins` command. 

The idea here is that the `DynamicBundle` allows for one definition that applies to multiple plugins instead of having to create a new `Bundle` definition for each new plugin. In the case of Operator SDK, the  4 `Bundle` definitions for the Go, Ansible, Helm, and Hybrid plugins could be condensed into one `DynamicBundle` definition.

I may not understand enough about the way that it will work with Phase 2 Plugins so if the Phase 2 Plugins implementation would make this obsolete or unnecessary then this can be ignored since it wouldn't be much of an improvement :)

designs/dynamic-bundle.md
camilamacedo86: c/c @rashmigottipati ^

designs/dynamic-bundle.md
camilamacedo86: Hi @everettraven,
 
> If Kubebuilder did force kustomize usage it would still be possible to override with another flag.

So, we would force and change the default behaviour of overwriting by using the --plugins flag. We would need a new flag to do what --plugins flag does today. Then, how that would work when SDK for i.e scaffold the java plugin? We would need to say for people use override flag instead of the plugins flag used?

c/c @rashmigottipati @jmrodri

everettraven: That is correct. In the PoC I made this new override flag to be called `--override-default-plugin-chain`

everettraven: See here: https://github.com/kubernetes-sigs/kubebuilder/blob/ec38164ea09419efe22b3d5f418ef9db6b3791ef/designs/dynamic-bundle.md?plain=1#L42-L43

designs/dynamic-bundle.md
everettraven: I have updated this section in an attempt to better reflect the goals of this proposal

everettraven: I have updated this section to reflect the actual intentions of this proposal

designs/dynamic-bundle.md
camilamacedo86: We are here trying to spec default pre and post scaffolds for the CLI.
So, it shows me that it would fit under the CLI API domain of responsibility instead of a Bundle API.

designs/dynamic-bundle.md
camilamacedo86: So, using SDK as an example would mean:

- User would no longer be able to have Golang projects scaffolds with operator-sdk init (it would only scaffold what the pre/pos default plugins defined). Then, they would need to `operator-sdk init --plugin=base.go/v3` 
- Users would no longer be able to scaffold `operator-sdk init --plugin=java/vx`. They would need to do  `operator-sdk init --override-default-plugin-chain=java/vx` 

IHMO: it does not seem intuitive + would be a breaking change for SDK and Kuebuilder. We would need to have SDK 2.0.0 and Kubebuilder 4.0.0). Also, defining default behaviour for the CLI itself does not seem to be part of a Bundle API domain responsibility and seems to fit more under the CLI API.

everettraven: > So, using SDK as an example would mean:
> 
>     * User would no longer be able to have Golang projects scaffolds with operator-sdk init (it would only scaffold what the pre/pos default plugins defined). Then, they would need to `operator-sdk init --plugin=base.go/v3`
> 

If the `DynamicBundle` is configured to default to scaffolding Golang projects in SDK then running `operator-sdk init` would still scaffold with the Go plugin.

>
>     * Users would no longer be able to scaffold `operator-sdk init --plugin=java/vx`. They would need to do  `operator-sdk init --override-default-plugin-chain=java/vx`
>

My understanding is that users would still be able to scaffold with the java plugin but it would just result in unused files. In this they would likely want to run something like `operator-sdk init --plugins java/v1 --override-default-plugin-chain`

everettraven: Just to break it down even further, the thought behind how the `DynamicBundle` is structured is this:
(This implements the Bundle interface as well but for this example I wanted to fully expand it to try and provide a further explanation)

```go
type dynamicBundle struct {
	bundle
	name    string
	version Version
	// these would be the plugins that a user wants to use. Can also set a default value for this.
	plugins []Plugin

	// these would be the plugins that are set to run before any other plugins.
	beforePlugins []Plugin

	// these would be the plugins that are set to run after all other plugins.
	afterPlugins  []Plugin
}
```

The `DynamicBundle` implements the `Plugin` interface and can be used the exact same way as a regular `Bundle`. For reference, in the PoC the `Plugins()` function looks like this:
```go
func (db dynamicBundle) Plugins() []Plugin {
	return append(db.beforePlugins, append(db.plugins, db.afterPlugins...)...)
}
```

This returns a list of plugins just like a regular `Bundle`. In order to fully implement the idea behind this there would need to be changes in the CLI to modify the `DynamicBundle`'s `plugin` field based on the user input. 

I guess in reality this means that the new `DynamicBundle` would not in itself make any changes to the CLI, but would require changes to the CLI for it to work the way intended.

All in all, the `DynamicBundle` by itself does not introduce any new functionality except breaking the plugin chain into a before chain, user defined chain, and an after chain.

The CLI API changes would be what truly makes the `DynamicBundle` able to function the way this proposal intends by following this kind of logic:
```mermaid
flowchart TD
  A[DynamicBundle is used] --> B{override-default-plugin-chain specified?};
  B -- Yes --> C[Only use plugins specified by user];
  B -- No --> D{plugins specified by user?};
  D -- Yes --> E[Replace DynamicBundle plugins field with the plugins specified by the user];
  D -- No --> F[Use Defaults];
  E --> G[scaffold];
  F --> G;
  C --> G;
```

Now I'd like to break this down by each potential scenario and show what each field of a `DynamicBundle` might look like at every stage of the flow.

Each scenario will start with a `DynamicBundle` in this state:
```
DynamicBundle {
   beforePlugins: [ kustomize/v1 ],
   plugins: [ go/v3 ],
   afterPlugins: [ manifests/v2 ]
}
```

---

Scenario 1 -- `DynamicBundle` is used and a user runs `operator-sdk init`:

**Flow stage**: *override-default-plugin-chain specified?*

Because the `--override-default-plugin-chain` is not specified this evaluates to "No".

The `DynamicBundle` state remains:
```
DynamicBundle {
   beforePlugins: [ kustomize/v1 ],
   plugins: [ go/v3 ],
   afterPlugins: [ manifests/v2 ]
}
```

**Flow stage**: *plugins specified by user?*

Because the `--plugins` flag is not used this evaluates to "No".

The `DynamicBundle` state remains:
```
DynamicBundle {
   beforePlugins: [ kustomize/v1 ],
   plugins: [ go/v3 ],
   afterPlugins: [ manifests/v2 ]
} 
```

**Flow Stage**: *use defaults*

The `DynamicBundle` state remains:
```
DynamicBundle {
   beforePlugins: [ kustomize/v1 ],
   plugins: [ go/v3 ],
   afterPlugins: [ manifests/v2 ]
} 
```

**Flow Stage**: *scaffold*

At this stage the `DynamicBundle` state is:
```
DynamicBundle {
   beforePlugins: [ kustomize/v1 ],
   plugins: [ go/v3 ],
   afterPlugins: [ manifests/v2 ]
}
```
This results in a project being scaffolded with `kustomize/v1` followed by `go/v3` followed by `manifests/v2`

---

Scenario 2 -- `DynamicBundle` is used and a user runs `operator-sdk init --plugins other/v1`:

**Flow stage**: *override-default-plugin-chain specified?*

Because the `--override-default-plugin-chain` is not specified this evaluates to "No".

The `DynamicBundle` state remains:
```
DynamicBundle {
   beforePlugins: [ kustomize/v1 ],
   plugins: [ go/v3 ],
   afterPlugins: [ manifests/v2 ]
}
```

**Flow stage**: *plugins specified by user?*

Because the `--plugins` flag is used this evaluates to "Yes".
The `DynamicBundle` state remains:
```
DynamicBundle {
   beforePlugins: [ kustomize/v1 ],
   plugins: [ go/v3 ],
   afterPlugins: [ manifests/v2 ]
}
```

**Flow Stage**: *Replace DynamicBundle plugins field with the plugins specified by the user*

The user has specified `--plugins other/v1` so `DynamicBundle` state becomes:
```
DynamicBundle {
   beforePlugins: [ kustomize/v1 ],
   plugins: [ other/v1 ],
   afterPlugins: [ manifests/v2 ]
}
```

**Flow Stage**: *scaffold*

At this stage the `DynamicBundle` state is:
```
DynamicBundle {
   beforePlugins: [ kustomize/v1 ],
   plugins: [ other/v1 ],
   afterPlugins: [ manifests/v2 ]
}
```
This results in a project being scaffolded with `kustomize/v1` followed by `other/v1` followed by `manifests/v2`

---

Scenario 3 -- `DynamicBundle` is used and a user runs `operator-sdk init --plugins other/v1 --override-default-plugin-chain`:

**Flow stage**: *override-default-plugin-chain specified?*

Because the `--override-default-plugin-chain` is specified this evaluates to "Yes".

The `DynamicBundle` state remains:
```
DynamicBundle {
   beforePlugins: [ kustomize/v1 ],
   plugins: [ go/v3 ],
   afterPlugins: [ manifests/v2 ]
}
```

**Flow stage**: *Only use plugins specified by user*

The user has specified `--plugins other/v1` so `DynamicBundle` state becomes:
```
DynamicBundle {
   beforePlugins: [],
   plugins: [ other/v1 ],
   afterPlugins: []
}
```

**Flow Stage**: *scaffold*

At this stage the `DynamicBundle` state is:
```
DynamicBundle {
   beforePlugins: [],
   plugins: [ other/v1 ],
   afterPlugins: []
}
```
This results in a project being scaffolded with `other/v1`


---

From my point of view this makes this proposal a little bit of a CLI API change and a new kind of Bundle API.

I hope that this helps clarify the intentions behind this proposal. Again, if this isn't something that seems fitting or useful to Kubebuilder that is fine - I just wanted to make sure the intentions behind the change is as clear as possible.

camilamacedo86: @everettraven, 

a) Users will be able to still generating the Helm/Ansible/Go and Java plugins offered by SDK using exactly the same commands and flags and have the same behaviours OR they will need to change it to begin to use override-default-plugin-chain instead?
b) Why is the override-default-plugin-chain is required? 
c) Why the proposal for a new API to create Plugins via Composition changes the default behaviour of the CLI  ? Why the DynamicBundle behaviour goes beyond its domain the responsibility by changing the current behaviour defined for the plugins flag on the CLI [i.e](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/cli.go#L259)
d) So, how does that would work if we want Bundles and DynimicBundles for the same CLI? 
e) How it will work with the subcommands create-api/create-webhook? Today we can `kb create-api --plugins=declarative/v1` which will create an API and scaffold the controller using the declarative plugin? 


**Regards the motivation:**
The motivation here is only to try to avoid a few "duplication" lines [here](https://github.com/operator-framework/operator-sdk/blob/master/internal/cmd/operator-sdk/cli/cli.go#L79-L108). Note that the composition is similar to these plugins but they could not necessarily be. We can add specific plugin helpers, for example only for ansible or only for the helm to scaffold nice features that are valid only for these languages in the future. So it is not a "code duplicated".

everettraven: > a) Users will be able to still generating the Helm/Ansible/Go and Java plugins offered by SDK using exactly the same commands and flags and have the same behaviours OR they will need to change it to begin to use override-default-plugin-chain instead?

So there is a bit of a conditional with this. **IF** SDK decided to use the `DynamicBundle` the same functionality would work for Helm/Ansible/Go plugins. The Java plugin would be where users would have to pass the new `--override-default-plugin-chain` command.

Just to reiterate, this change does not have to be implemented by either Kubebuilder or SDK. They can keep their current implementation and nothing would change for the users.

Only change that would appear would be the new `--override-default-plugin-chain` flag. We could even potentially put that flag behind a conditional statement in the CLI logic to obfuscate it if a `DynamicBundle` is not being used.

> b) Why the new `override-default-plugin-chain` flag is required? Why do we need a new flag to do what is done via the plugins flag currently?

The new `--override-default-plugin-chain` is because the idea of the `DynamicBundle` is that there is the ability define the separate layers of the bundle. These layers being:
  - Before
  - Injected
  - After
These layers are processed in the order of Before -> Injected -> After

My thoughts were that if a `DynamicBundle` were being used and a user specified the `--plugins` flag that then those plugins would be used in the Injected layer. This would mean that in order to prevent the `DynamicBundle` logic of running in the order of Before > Injected > After and only running plugins in the Injected layer users would have to override that logic.

Now that I am thinking more about it, we could also perform this logic in an alternative way that I *think* would have no impact to the current functionality. This alternative way would instead rely on a different flag called something like `--inject-plugins`, where a user can define plugins that they would want to *inject* into the `DynamicBundle` flow. This would keep  the logic that if the `--plugins` flag is used then completely override and just use the exact plugin chain the user has specified.

What do you think of this potential alternative? (I actually think I prefer this alternative)


> c) So, how would that work if we want Bundles and DynimicBundles for the same CLI?

It *should* work the same because a Bundle just acts as a plugin. I don't immediately see why there would be any major conflicts 


> d) How it will work with the subcommands create-api/create-webhook? Today we can `kb create-api --plugins=declarative/v1`, which will create an API and scaffold the controller using the declarative plugin?

The same flag change would likely need to be made for the other subcommands. However if we went with the alternative solution proposed with my answer to your question C, I think it would still function exactly the same.

> e) Would Kubebuilders users still be able to scaffold only the Golang base plugin with kubebuilder init --plugin=base.go/v3 after this implementation OR they would need to use the new flag instead?

The originally proposed solution would require the use of the new `--override-default-plugin-chain` command.

The alternative solution I proposed in response to your question C, this would function exactly the same.

everettraven: @camilamacedo86 I just want to highlight a potential alternative solution that I proposed in response to one of your questions in my above response because I think I may actually prefer it because I don't *think* it would change the existing default functionality when using the `--plugins` flag, and would instead just introduce one new flag.



> Now that I am thinking more about it, we could also perform this logic in an alternative way that I _think_ would have no impact to the current functionality. This alternative way would instead rely on a different flag called something like `--inject-plugins`, where a user can define plugins that they would want to _inject_ into the `DynamicBundle` flow. This would keep the logic that if the `--plugins` flag is used then completely override and just use the exact plugin chain the user has specified.
> 
> What do you think of this potential alternative? (I actually think I prefer this alternative)

designs/dynamic-bundle.md
everettraven: After taking some time to play around with the CLI suggestion I think this definitely makes more sense. You are right that this is the best way to go about allowing this kind of functionality. I am unnecessarily adding complexity here by adding a new Bundle API on top of a CLI API change when all that is necessary is a CLI change.

Thanks for mentioning this and bringing it up @camilamacedo86 !

If this is some functionality that you think may be nice to have in the CLI API (even if Kubebuilder or SDK wouldn't be able to use it) I'd be happy to write up a new design proposal (or modify this one) and PoC for it.

everettraven:
Closing as it was determined that this does not add value to Kubebuilder.

Issue #2667, opened 2022-04-30
Error to build release artefacts automatically via Google Cloud


camilamacedo86:
### What broke? What's expected?

When we push a new tag we should build the release.
However, the following error has been faced:

```
Step #1: go: creating new go.mod: module tmp
Step #1: go: downloading sigs.k8s.io/kubebuilder-release-tools/notes v0.0.0-20201001193718-4777888c377a
Step #1: go: downloading sigs.k8s.io/kubebuilder-release-tools v0.1.1
Step #1: go: downloading github.com/blang/semver/v4 v4.0.0
Step #1: go: added github.com/blang/semver/v4 v4.0.0
Step #1: go: added sigs.k8s.io/kubebuilder-release-tools/notes v0.0.0-20201001193718-4777888c377a
Step #1: /workspace
Step #1: build/build_kubebuilder.sh: line 93: notes: command not found
Finished Step #1
ERROR
ERROR: build step 1 "goreleaser/goreleaser:v1.8.3" failed: step exited with non-zero status: 127
```

### Reproducing this issue

Build a release based on tag

### KubeBuilder (CLI) Version

master

PR #2673, opened 2022-05-06
:seedling: Ignore release notes in the release to fix it

camilamacedo86:
**Description**
Ignore release notes in the release to fix it

**Motivation**
See: https://github.com/kubernetes-sigs/kubebuilder/issues/2667

Currently, the releases have been not generated in the Google Cloud build because the script tries to install the go module to generate the release notes and then use it. But it cannot work. See; https://github.com/GoogleCloudPlatform/cloud-builders/blob/master/go/README.md#note-1-workspace-and-go
 
Then, the idea here just stops trying to generate the release notes and update the release page automatically because it can be done easily manually. However, building the assets is more complicated. 

So, let's keep the release building the assets and adding only them to the release page so that we can add the changelog manually for now.

### DIFF
--- a/build/build_kubebuilder.sh
+++ b/build/build_kubebuilder.sh
@@ -90,7 +90,10 @@ if [ -z "$SNAPSHOT" ]; then
       [[ "${TAG_NAME}" == "v"*"-beta."* ]] && NOTES_FLAGS+=" -r beta"
       [[ "${TAG_NAME}" == "v"*"-rc."* ]] && NOTES_FLAGS+=" -r rc"
   fi
-  notes $NOTES_FLAGS | tee "$tmp_notes"
+  # TODO(cmacedo): figure out how to download the release notes and let it available in the cloud build
+  # Currently it does not work: https://github.com/kubernetes-sigs/kubebuilder/issues/2667
+  # notes $NOTES_FLAGS | tee "$tmp_notes"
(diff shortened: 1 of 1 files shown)


build/build_kubebuilder.sh
```diff
@@ -90,7 +90,10 @@ if [ -z "$SNAPSHOT" ]; then
       [[ "${TAG_NAME}" == "v"*"-beta."* ]] && NOTES_FLAGS+=" -r beta"
       [[ "${TAG_NAME}" == "v"*"-rc."* ]] && NOTES_FLAGS+=" -r rc"
   fi
-  notes $NOTES_FLAGS | tee "$tmp_notes"
+  # TODO(cmacedo): figure out how to download the release notes and let it available in the cloud build
+  # Currently it is not work: https://github.com/kubernetes-sigs/kubebuilder/issues/2667
```
erikgb, line 94: Nit:

```suggestion
  # Currently it does not work: https://github.com/kubernetes-sigs/kubebuilder/issues/2667
```

erikgb:
/lgtm
/hold

One minor wording suggestion, but ok to merge as it is.

rashmigottipati (approved):

rashmigottipati (approved):
/lgtm

PR #2678, opened 2022-05-09
:bug: (go/v2) fix issue introduced by removing the GO111MODULE=on from Dockerfile

camilamacedo86:
**Description**

The ` GO111MODULE=on ` was removed from go/v2 with the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/2507
However, projects scaffolds with go/v2 are using golang 1.13 which requires this option


erikgb:
/lgtm

Do we have a plan to drop support for "old stuff" 😉

PR #2679, opened 2022-05-11
⚠️ (go/v3 and config-gen alpha command) Update controller-runtime dep from v0.11.2 to v0.12.1, controller-tools from v0.8.0 to v0.9.0 and k8s deps from 1.23 to 1.24.1

NikhilSharmaWe:
- The new release is bumped with k8s 1.24 and go 1.18. 
- Users that might do the upgrade will need to explicitly set configmapsleases when configuring leader election, more info: https://github.com/kubernetes-sigs/controller-runtime/pull/1773 if you are (migrating from v2 to v3). 
- Check out https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.12.0 for more info about the new release.

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/2681


camilamacedo86 (changes requested):
See the comment: https://github.com/kubernetes-sigs/kubebuilder/pull/2679/files#r871211532

NikhilSharmaWe:
@camilamacedo86 
I think unit tests are failing due to the versions of `k8s.io/cli-runtime@v0.23.0` and `k8s.io/kubectl@v0.22.2`, do we also need to upgrade their versions to the latest tag: `v0.24.0`.

go.mod
```diff
@@ -7,14 +7,14 @@ require (
 	github.com/gobuffalo/flect v0.2.3
 	github.com/joelanford/go-apidiff v0.1.0
 	github.com/onsi/ginkgo v1.16.5
```
camilamacedo86, line 9: Following are some things that we need to do: 

a) We need to clarify that it is breaking change
We also will need to make it very clear when we release the version with the info: https://github.com/kubernetes-sigs/controller-runtime/pull/1773 (because it impacts the scaffolds)

So, I'd like for us to update the title with

⚠️ (go/v3 and config-gen alpha command) Update controller-runtime dep from v0.11.2 to v0.12.0. ( you should explicitly set configmapsleases when configuring leader election, more info: https://github.com/kubernetes-sigs/controller-runtime/pull/1773 )

So that we will have this info for the release notes, it will help to communicate and remember that we need to highlight it to users.

b) Also, we need to update the migration guide from v2 to v3 to have this step so that users that might do the upgrade after we release KB will have the right info: https://book.kubebuilder.io/migration/v2vsv3.html

(it can be a follow-up, could you please track an issue "upgrade controller runtime to the latest ... " so can track there all details and follow-ups?)

c) This PR is missing we upgrade the scaffold of go/v3 plugin: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/scaffolds/init.go#L33

d) we will also need to run `make generate` and update the downgrade of deps in: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/commons.go#L89-L151

camilamacedo86: It is solved

go.mod
```diff
@@ -62,11 +61,11 @@ require (
...
 	github.com/zmap/zlint/v2 v2.2.1 // indirect
-	golang.org/x/crypto v0.0.0-20210817164053-32db794688a5 // indirect
+	golang.org/x/crypto v0.0.0-20220214200702-86341886e292 // indirect
 	golang.org/x/exp v0.0.0-20200224162631-6cc2880d07d6 // indirect
-	golang.org/x/mod v0.4.2 // indirect
-	golang.org/x/net v0.0.0-20211209124913-491a49abca63 // indirect
-	golang.org/x/sys v0.0.0-20211029165221-6e7872819dc8 // indirect
+	golang.org/x/mod v0.6.0-dev.0.20220106191415-9b9b3d81d5e3 // indirect
+	golang.org/x/net v0.0.0-20220127200216-cd36cc0744dd // indirect
+	golang.org/x/sys v0.0.0-20220209214540-3681064d5158 // indirect
```
camilamacedo86, line 68: Because of the breaking change on controller-runtime, I think we need to do here:

- add  set LeaderElectionResourceLock to "configmapsleases"  so that we do not change the behaviour for the default scaffolds
- ensure that it has a comment on top explaining that default behaviour change in 0.12 and has a link for the users to check the steps in: https://github.com/kubernetes-sigs/controller-runtime/pull/1903/files or we copy them.
- also, we need a new issue in the repo tracked under the go/v4 milestone to remove this option for go/v4 and we will need to doc the required changes in the migration guide from go/v3 to go/v4 

c/c @rashmigottipati @ryantking @varshaprasad96

camilamacedo86:
This PR is blocked by upgrading the declarative project: https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/217. 

Also, in order to prevent authors migrate their project and facing problems with the breaking change that affects what scaffold by default is, (IMO) we need to: https://github.com/kubernetes-sigs/kubebuilder/pull/2679#discussion_r877821944

everettraven:
> Unit tests are failing Is it due to the versions of `k8s.io/cli-runtime@v0.23.0` and `k8s.io/kubectl@v0.22.2`.
> 
> All error logs are like:
> 
> ```
> Error: cannot use oapi (type *"github.com/google/gnostic/openapiv2".Document) as type *"github.com/googleapis/gnostic/openapiv2".Document in argument to supportsDryRun
> ```

@NikhilSharmaWe these errors seem to be related to the kubebuilder-declarative-pattern dependency, so since your PR to bump it to support k8s 1.24 was merged, I think this line needs to be modified:
https://github.com/kubernetes-sigs/kubebuilder/blob/c7d57aa42786368328461e1128e7b89a6450eb7a/pkg/plugins/golang/declarative/v1/api.go#L35
to be:
```go
kbDeclarativePatternForV3 = "fe5be9431eae158f86f9de23000a9a2ec06745fc"
```

everettraven:
@NikhilSharmaWe I think I may have found the problem that is causing the tests to fail with the proto file issue. 

### Problem
The problem seems to be because the go.mod dependencies are being populated with indirect dependencies that map to the k8s packages with a version of `v0.24.0`. When the dependencies for the v1beta1 support are modified, the indirect dependencies aren't properly updated without removing all the indirect dependencies before running `go mod tidy`.

The exact indirect dependency that is causing the issue is `github.com/google/gnostic v0.5.7-v3refs // indirect` because client-go v0.24.0 and controller-runtime v0.12.1 have this dependency causing the scaffolded project to have it as an indirect dependency. client-go v0.21.2 and controller-runtime v0.9.2 use `github.com/googleapis/gnostic v0.5.5 // indirect` which is a different package path entirely.

### Solution
The solution I found is to modify the `pkg/plugins/golang/v3/commons.go` file to remove all the indirect dependencies before running `go mod tidy`. This logic looks like:
```go
// Due to some indirect dependency changes with a bump in k8s packages from 0.23.x --> 0.24.x we need to
// clear out all indirect dependencies before we run `go mod tidy` so that the above changes get resolved correctly
if err := util.ReplaceRegexInFile("go.mod", `(require \(\n(\t.* \/\/ indirect\n)+\))`, ""); err != nil {
	log.Warnf("unable to update the go.mod indirect dependencies: %s", err)
}
```

I will also leave a PR review comment around the area in the file it should be added.

pkg/plugins/golang/v3/commons.go
```diff
@@ -145,7 +145,7 @@ manifests: controller-gen`
 		}
 
 		if err := util.ReplaceInFile("go.mod",
-			"k8s.io/klog/v2 v2.30.0",
+			"k8s.io/klog/v2 v2.60.1",
 			"k8s.io/klog/v2 v2.9.0"); err != nil {
 			log.Warnf("unable to update the go.mod with k8s.io/klog/v2 v2.9.0: %s", err)
 		}
```
everettraven, line 151: After this line we should add:
```go
// Due to some indirect dependency changes with a bump in k8s packages from 0.23.x --> 0.24.x we need to
// clear out all indirect dependencies before we run `go mod tidy` so that the above changes get resolved correctly
if err := util.ReplaceRegexInFile("go.mod", `(require \(\n(\t.* \/\/ indirect\n)+\))`, ""); err != nil {
	log.Warnf("unable to update the go.mod indirect dependencies: %s", err)
}
```

The reasoning for this change can be seen here: https://github.com/kubernetes-sigs/kubebuilder/pull/2679#issuecomment-1146263855

go.mod
```diff
@@ -7,14 +7,14 @@ require (
...
-	github.com/spf13/cobra v1.2.1
+	github.com/spf13/cobra v1.4.0
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/tools v0.1.6-0.20210820212750-d4cc65f0b2ff
-	k8s.io/apimachinery v0.23.5 // for `kubebuilder alpha config-gen`
-	sigs.k8s.io/controller-runtime v0.11.2
+	golang.org/x/tools v0.1.10-0.20220218145154-897bd77cd717
+	k8s.io/apimachinery v0.24.0 // for `kubebuilder alpha config-gen`
+	sigs.k8s.io/controller-runtime v0.12.1
 	sigs.k8s.io/controller-tools v0.8.0 // for `kubebuilder alpha config-gen`
```
everettraven, line 18: controller-tools should be upgraded to `v0.9.0` since it doesn't support k8s 1.24 until that release. More info on the v0.9.0 release can be found here: https://github.com/kubernetes-sigs/controller-tools/releases/tag/v0.9.0

```suggestion
	sigs.k8s.io/controller-tools v0.9.0 // for `kubebuilder alpha config-gen`
```

everettraven (changes requested):

everettraven:
@NikhilSharmaWe it looks like the new unit test failures are related to an update made in controller-tools v0.9.0 that no longer writes the `status` field in the CRD. The PR that made this change is here: https://github.com/kubernetes-sigs/controller-tools/pull/630

Due to this we need to update the expected files to match the new format. To do this you need to:
1.  uncomment this line:
https://github.com/kubernetes-sigs/kubebuilder/blob/019c1854f67e3bd61af5aa834a1573e6934b2e8b/pkg/cli/alpha/config-gen/configgen_test.go#L30
2. In the `pkg/cli/alpha/config-gen` directory, run `go test`
3. comment the changed line again

After that the unit tests should pass again.

everettraven (approved):
Looks good to me, great job @NikhilSharmaWe !

everettraven:
/cc @joelanford

Could you take a look at this so we can try and get this in ASAP? Thanks!

jmrodri:
What does this error mean?
```
sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v3/scaffolds
  Incompatible changes:
  - ControllerRuntimeVersion: value changed from "v0.11.2" to "v0.12.1"
```

varshaprasad96:
> What does this error mean?
> 
> ```
> sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v3/scaffolds
>   Incompatible changes:
>   - ControllerRuntimeVersion: value changed from "v0.11.2" to "v0.12.1"
> ```

That seems to be because of an APIDiff check - modifying an exported field. If someone with right permissions could override that check, we can merge it.

everettraven:
> What does this error mean?
> 
> ```
> sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v3/scaffolds
>   Incompatible changes:
>   - ControllerRuntimeVersion: value changed from "v0.11.2" to "v0.12.1"
> ```

@jmrodri from what I can tell it seems to just be stating that the bump to v0.12.1 is a breaking change from v0.11.2 which we already knew. As @varshaprasad96 mentioned, I think this can be safely overridden.

jmrodri (approved):
/lgtm

jmrodri:
Looks pretty straight forward to me.

go.mod
camilamacedo86: @NikhilSharmaWe, 
We need also add this info in the title and description
I am doing tha

pkg/plugins/golang/v3/commons.go
```diff
@@ -145,11 +145,17 @@ manifests: controller-gen`
...
 
+		// Due to some indirect dependency changes with a bump in k8s packages from 0.23.x --> 0.24.x we need to
+		// clear out all indirect dependencies before we run `go mod tidy` so that the above changes get resolved correctly
+		if err := util.ReplaceRegexInFile("go.mod", `(require \(\n(\t.* \/\/ indirect\n)+\))`, ""); err != nil {
+			log.Warnf("unable to update the go.mod indirect dependencies: %s", err)
+		}
```
camilamacedo86, line 157: Not sure if we should do that for all. 
It might have other deps.
Ideally, we should be very precise here and only update the deps that are scaffold by default.
Could you please check this one?

everettraven: This was a change that I recommended. I think it is fine if we want to change this to only include changing the dependency that was causing an issue, however I will explain my reasoning for the recommendation below.

*Reasoning for recommended change*
- The indirect dependencies in the `go.mod` seem to be getting populated with the indirect dependencies that correlate to the base scaffolding (e.g. k8s 1.24 dependencies) and are not properly updated when the direct dependency versions are downgraded. When I tested this locally, running `go mod tidy` did not update the indirect dependencies when downgrading direct dependencies.
- As a result of the above, the indirect dependencies that are being used may be incompatible with the versions that the direct dependencies are using. I think we should be ensuring that the indirect dependencies that are being used are the same as those that are used when manually creating a new `go.mod` that contains the same direct dependencies. The best way I found to do this with Kubebuilder's scaffolding is to remove all the indirect dependencies and then run `go mod tidy` to repopulate them. This also helps to ensure that all the indirect dependencies will be compatible with the expectations of the downgraded direct dependencies.

@camilamacedo86 WDYT? If this should be considered out of the scope of this PR and this is something we should still change, I am happy to create an issue to track this and just have this PR fix the dependencies it needs to fix.

PR #2686, opened 2022-05-14
:seedling: move estroz and adirio for emeritus (no longer directly involved)

camilamacedo86:
**Description**
move estroz and adirio for emeritus (no longer directly involved)

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -5,12 +5,10 @@ aliases:
   # tasks on the repo, or otherwise approve any PRs.
   kubebuilder-admins:
     - pwittrock
-    - estroz
     - camilamacedo86
 
   # non-admin folks who can approve any PRs in the repo
   kubebuilder-approvers:
-    - adirio
 
   # folks who can review and LGTM any PRs in the repo (doesn't include
   # approvers & admins -- those count too via the OWNERS file)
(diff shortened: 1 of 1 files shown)


camilamacedo86:
@joelanford @jmrodri @rashmigottipati 

Both are no longer directly involved with the projects should we not move them for emeritus?

OWNERS_ALIASES
```diff
@@ -5,12 +5,10 @@ aliases:
   # tasks on the repo, or otherwise approve any PRs.
   kubebuilder-admins:
     - pwittrock
-    - estroz
     - camilamacedo86
 
   # non-admin folks who can approve any PRs in the repo
   kubebuilder-approvers:
-    - adirio
```
erikgb, line 13: Should this be an empty list to be valid? Or is it ok like this?

erikgb (commented):
Maybe a little more redundancy is needed? The list of people with elevated permissions is short now...

OWNERS_ALIASES
FillZpp: It seems ok to have empty list.

camilamacedo86: We need more contributions @erikgb and @FillZpp by sure
I hope that up to the end of this year we will be able to grow this list
and you both are very welcome as well ;-)

PR #2693, opened 2022-05-18
:sparkles: external plugins(plugin phase 2): add support to pass CLI flags

### LINKED ISSUE #2689: Phase 2 Plugins should be able to parse flags passed in by a user
## Problem
The current implementation of Phase 2 Plugins does not provide logic to parse flags that a user passes to Kubebuilder to use with an external plugin.

For example, running `kubebuilder init --plugins myexternalplugin/v1 --domain example.com` you will get an output similar to:
```
INFO[0000] Adding external plugin: myexternalplugin     
Error: unknown flag: --domain
Usage:
  kubebuilder init [flags]

Examples:
  # Help for initializing a project with version "2"
  kubebuilder init --project-version="2" -h

  # Help for initializing a project with version "3"
  kubebuilder init --project-version="3" -h

Flags:

Global Flags:

2022/05/16 09:38:53 unknown flag: --domain
```

This output occurs even if the external plugin defines this flag because the external plugin does not do any logic to bind flags. We can see in this file that a `BindFlags` function is missing: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/external/init.go

## Proposed Solution
I propose that a solution is implemented similar to this:
```mermaid
flowchart TD
    A[BindFlags] --> B[Get Flags from External Plugin];
    B ----> C{Error?};
    C -- Yes --> D[Parse all flags and provide general message to consult documentation];
    C -- No --> E[Parse only flags specified by External Plugin response];
    D ----> F[Call External Plugin with flags];
    E ----> F;
```
This would allow for external plugin authors to define a command that could be used to get the flags that are allowed, making it so that Kubebuilder can provide an experience to its users that is similar to using the built in plugins. If an external plugin author does not implement this command Kubebuilder can default to allowing all flags to be parsed when using an external plugin and provide a general message to consult the external plugin documentation for more information regarding the flags.

### Extra Labels

_No response_

everettraven:
## Description of Change
This change implements the `BindFlags()` method for the external plugin `init`, `create api`, `create webhook`, and `edit` subcommands.

It has 2 methods in which flags can be bound:
1. An external plugin defines a `flags` subcommand that accepts boolean flags to represent which subcommand to return flags to Kubebuilder for. (ex: `flags --init`)
2. All flags passed by a user are bound when an external plugin is used. This method also adds a default usage message to the flags stating that Kubebuilder was unable to verify the flags passed by a user and suggests that the user consult the external plugin documentation for more information.

The logic flow that is taken is:
```mermaid
flowchart TD
    A[BindFlags] --> B[Request Flags from External Plugin];
    B ----> C{Error?};
    C -- Yes --> D[Parse all flags and provide general message to consult documentation];
    C -- No --> E[Parse only flags specified by External Plugin response];
    D ----> F[Call External Plugin with flags];
    E ----> F;
```

## Motivation for the change
With the current Phase 2 Plugins implementation, the user passed CLI flags are not properly bound. This results in an error stating that a flag passed by a user is an unknown flag even if the external plugin defines the flag.

For example:
If `myexternalplugin/v1` defines the `--captain` flag and a user runs `kubebuilder init --plugins myexternalplugin/v1 --captain kube` the output would be:
```
INFO[0000] Adding external plugin: myexternalplugin     
Error: unknown flag: --captain
Usage:
  kubebuilder init [flags]

Examples:
  # Help for initializing a project with version "2"
  kubebuilder init --project-version="2" -h

  # Help for initializing a project with version "3"
  kubebuilder init --project-version="3" -h

Flags:

Global Flags:

2022/05/16 09:38:53 unknown flag: --captain
```

For details on the motivation see issue #2689 

## Additional Context
resolves #2689

### DIFF
--- a/pkg/cli/options.go
+++ b/pkg/cli/options.go
@@ -155,8 +155,18 @@ func WithCompletion() Option {
 
 // parseExternalPluginArgs returns the program arguments.
 func parseExternalPluginArgs() (args []string) {
-	args = make([]string, len(os.Args)-1)
-	copy(args, os.Args[1:])
+	// Loop through os.Args and only get flags and their values that should be passed to the plugins
+	// this also removes the --plugins flag and its values from the list passed to the external plugin
+	for i := range os.Args {
+		if strings.Contains(os.Args[i], "--") && !strings.Contains(os.Args[i], "--plugins") {
+			args = append(args, os.Args[i])
+
+			// Don't go out of bounds and don't append the next value if it is a flag
+			if i+1 < len(os.Args) && !strings.Contains(os.Args[i+1], "--") {
+				args = append(args, os.Args[i+1])
+			}
+		}
+	}
 
 	return args
 }
--- a/pkg/plugin/external/types.go
+++ b/pkg/plugin/external/types.go
@@ -57,4 +57,29 @@ type PluginResponse struct {
 
 	// ErrorMsgs contains the specific error messages of the plugin failures.
 	ErrorMsgs []string `json:"errorMsgs,omitempty"`
+
+	// Flags contains the plugin specific flags that the plugin returns to Kubebuilder when it receives
+	// a request for a list of supported flags from Kubebuilder
+	Flags []Flag `json:"flags,omitempty"`
+}
+
+// Flag is meant to represent a CLI flag that is used by Kubebuilder to define flags that are parsed
+// for use with an external plugin
+type Flag struct {
+	// Name is the name that should be used when creating the flag.
+	// i.e a name of "domain" would become the CLI flag "--domain"
+	Name string
+
+	// Type is the type of flag that should be created. The types that
+	// Kubebuilder supports are: string, bool, int, and float.
+	// any value other than the supported will be defaulted to be a string
+	Type string
+
+	// Default is the default value that should be used for a flag.
+	// Kubebuilder will attempt to convert this value to the defined
+	// type for this flag.
+	Default string
+
+	// Usage is a description of the flag and when/why/what it is used for.
+	Usage string
 }
--- a/pkg/plugins/external/api.go
+++ b/pkg/plugins/external/api.go
@@ -17,6 +17,8 @@ limitations under the License.
 package external
 
 import (
+	"github.com/spf13/pflag"
+
 	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
 	"sigs.k8s.io/kubebuilder/v3/pkg/model/resource"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
@@ -37,6 +39,10 @@ type createAPISubcommand struct {
 func (p *createAPISubcommand) InjectResource(*resource.Resource) error {
 	// Do nothing since resource flags are passed to the external plugin directly.
 	return nil
+}
+
+func (p *createAPISubcommand) BindFlags(fs *pflag.FlagSet) {
+	bindExternalPluginFlags(fs, "api", p.Path, p.Args)
 }
 
 func (p *createAPISubcommand) Scaffold(fs machinery.Filesystem) error {
--- a/pkg/plugins/external/edit.go
+++ b/pkg/plugins/external/edit.go
@@ -17,6 +17,8 @@ limitations under the License.
 package external
 
 import (
+	"github.com/spf13/pflag"
+
 	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugin/external"
@@ -27,6 +29,10 @@ var _ plugin.EditSubcommand = &editSubcommand{}
 type editSubcommand struct {
 	Path string
 	Args []string
+}
+
+func (p *editSubcommand) BindFlags(fs *pflag.FlagSet) {
+	bindExternalPluginFlags(fs, "edit", p.Path, p.Args)
 }
 
 func (p *editSubcommand) Scaffold(fs machinery.Filesystem) error {
--- a/pkg/plugins/external/external_test.go
+++ b/pkg/plugins/external/external_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package external
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -25,7 +26,10 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/spf13/afero"
+	"github.com/spf13/pflag"
+
 	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin/external"
 )
 
 func TestExternalPlugin(t *testing.T) {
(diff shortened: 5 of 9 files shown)


pkg/plugins/external/external_test.go
```diff
@@ -248,4 +263,257 @@ var _ = Describe("Run external plugin using Scaffold", func() {
...
+		)
+		BeforeEach(func() {
+			outputGetter = &mockInValidOutputGetter{}
+			currentDirGetter = &mockValidOsWdGetter{}
+
+			pluginFileName = "myexternalplugin.sh"
+			args = []string{"--captain", "black-beard", "--sail"}
+			flagset = pflag.NewFlagSet("test", pflag.ContinueOnError)
+			usage = "Kubebuilder could not validate this flag with the external plugin. Consult the external plugin documentation for more information."
+
```
camilamacedo86, line 364: You cna test the lint locally with make lint
also some silly issues can be fixed with make lint-fix 

however, in this case the error is the size of the line which is too big so it will not be fixed with the above option

pkg/plugins/external/init.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package external
 
 import (
+	"github.com/spf13/pflag"
 	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
```
camilamacedo86, line 21:
```suggestion

	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
```

nit space to sort the imports

pkg/plugins/external/webhook.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package external
 
 import (
+	"github.com/spf13/pflag"
 	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
```
camilamacedo86, line 21: same

pkg/cli/options.go
```diff
@@ -155,8 +155,18 @@ func WithCompletion() Option {
 
 // parseExternalPluginArgs returns the program arguments.
 func parseExternalPluginArgs() (args []string) {
-	args = make([]string, len(os.Args)-1)
```
camilamacedo86, line 158: I would like to see this scenarios tested with e2e. Example:

By having the sample plugin here and calling them as we do in the others e2e tests.
So that we can ensure the expected results, which would make it simple, we review the changes.

could we try to work on that ? WDYT @rashmigottipati @everettraven ?

rashmigottipati: @camilamacedo86 I'll work on a separate PR to add e2e test with the python external plugin I've built. But I wouldn't make e2e tests as a part of this PR, as the goal of this PR is to enable CLI flags for external plugins.

everettraven: I'd be happy to go through and add as a comment to this thread the reasoning for this change with proof when running it locally.

pkg/plugins/external/api.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package external
 
 import (
+	"github.com/spf13/pflag"
 	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
```
camilamacedo86, line 21: space nit

pkg/plugins/external/webhook.go
```diff
@@ -35,6 +36,24 @@ func (p *createWebhookSubcommand) InjectResource(*resource.Resource) error {
...
+	// Get a list of flags for the init subcommand of the external plugin
+	// If it returns an error, parse all flags passed by the user and let
+	// the external plugin return an unknown flag error.
+	flags, err := getExternalPluginFlags(req, p.Path)
+	if err != nil {
+		bindAllFlags(fs, p.Args)
+	} else {
+		bindSpecificFlags(fs, flags)
+	}
+}
```
camilamacedo86, line 55: The code changes here are equal for all places
Could we find a way of not duplicate/

pkg/plugins/external/external_test.go
everettraven: All lint issues should be fixed in https://github.com/kubernetes-sigs/kubebuilder/pull/2693/commits/9648e34decca255094c2a4b1934d92ae014caa78

pkg/plugins/external/init.go
everettraven: Should be done for all imports as of commit https://github.com/kubernetes-sigs/kubebuilder/pull/2693/commits/9648e34decca255094c2a4b1934d92ae014caa78

pkg/plugins/external/webhook.go
everettraven: Should be resolved in https://github.com/kubernetes-sigs/kubebuilder/pull/2693/commits/88a68a587f85126c4b5b68bec28a7bfa881456ad

pkg/cli/options.go
```diff
@@ -155,8 +155,18 @@ func WithCompletion() Option {
```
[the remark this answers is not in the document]
everettraven: Updated the PR title

pkg/plugins/external/edit.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package external
 
 import (
+	"github.com/spf13/pflag"
+
```
camilamacedo86, line 21: @everettraven,

We need to squash the commits for Kubebuilder as well. 
Could you please squash them?

pkg/cli/options.go
everettraven: When using the original method (adding a print statement to print the args) with the changes that pass the CLI flags to an external plugin, running `kubebuilder init --plugins myexternalplugin/v1 --domain example.com` results in this error output:
```
[init --plugins myexternalplugin/v1 --domain example.com]
INFO[0000] Adding external plugin: myexternalplugin     
usage: myexternalplugin.py [-h] [--domain DOMAIN] [--license LICENSE]
                           [--boolean | --no-boolean]
myexternalplugin.py: error: unrecognized arguments: init --plugins myexternalplugin/v1
Error: failed to initialize project: unable to scaffold with "myexternalplugin/v1": exit status 2
Usage:
  kubebuilder init [flags]

Examples:
  # Help for initializing a project with version "2"
  kubebuilder init --project-version="2" -h

  # Help for initializing a project with version "3"
  kubebuilder init --project-version="3" -h

Flags:
      --boolean                  flag to test boolean logic
      --domain string            Project domain (default "my.domain")
      --float float              flag to test float logic (default 123.45)
      --integer int              flag to test integer logic (default 123)
      --license string           Project license (default "apache2")

Global Flags:

2022/06/01 09:18:05 failed to initialize project: unable to scaffold with "myexternalplugin/v1": exit status 2

```

The important line to note in that error message is:
> myexternalplugin.py: error: unrecognized arguments: init --plugins myexternalplugin/v1

This tells us that we are parsing all arguments after the program name itself, including subcommands and flags that are likely only to be used with Kubebuilder (`--plugins` flag specifically). We can see this more so from the print statement that I added for testing here:
> [init --plugins myexternalplugin/v1 --domain example.com]

The modified function ensures that we only parse flags and the corresponding flag values. It also ensures that we do not parse the `--plugins` flag since it is likely only used with Kubebuilder.

With the new changes the output of the same command is:
```
[--domain example.com]
INFO[0000] Adding external plugin: myexternalplugin  
```

There is no error since the proper flags are passed to the external plugin.

@camilamacedo86 @rashmigottipati I hope this helps demonstrate the reason for the change!

pkg/plugin/external/types.go
```diff
@@ -57,4 +57,29 @@ type PluginResponse struct {
 
 	// ErrorMsgs contains the specific error messages of the plugin failures.
 	ErrorMsgs []string `json:"errorMsgs,omitempty"`
+
+	// Flags contains the plugin specific flags that the plugin returns to Kubebuilder when it receives
+	// a request for a list of supported flags from Kubebuilder
+	Flags []Flag `json:"flags,omitempty"`
```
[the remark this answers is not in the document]
everettraven: I took a look at doing this and it seems like the way to determine the type of the flag is an interface. I think that structure would make it difficult for external plugin authors to be able to specify the type of the flag, and the new `Flag` struct introduced below makes it a bit easier IMO.

camilamacedo86 (approved):
I am OK with the changes but we need @rashmigottipati OK here since she is who wrote the plugin phase 2 and is responsible for this piece.

But we really need to add the samples/tests with the samples and doc for phase 2 otherwise it is like impossible to keep it maintained and let people know how to use it.

PR #2698, opened 2022-05-24
:sparkles: (kustomize/v1) add validation to let users know when the architecture used to run the plugin is not supported

camilamacedo86:
**Description**
(kustomize/v1) add validation to let users know when the architecture used to run the plugin is not supported

**Motivation**
We have seen many users get stuck. So, the idea here is clarifying for contributors, KB lib consumers and end-users what are the current support architectures by kustomize/v1. Example: https://github.com/operator-framework/operator-sdk/issues/5785

### DIFF
--- a/docs/book/src/plugins/kustomize-v1.md
+++ b/docs/book/src/plugins/kustomize-v1.md
@@ -4,6 +4,19 @@ The kustomize plugin allows you to scaffold all kustomize manifests used to work
 By using the kustomize plugin, you can create your own language plugins and ensure that you will have the same configurations 
 and features provided by it. 
 
+<aside class="note">
+<h1>Supportability</h1>
+
+This plugin uses [kubernetes-sigs/kustomize](https://github.com/kubernetes-sigs/kustomize) v3 and the architectures supported are: 
+- linux/amd64
+- linux/arm64
+- darwin/amd64
+
+You might want to consider using [kustomize/v2-alpha](./kustomize-v2-alpha.md) if you are looking to scaffold projects in
+other architecture environments. (i.e. if you are looking to scaffold projects with Apple Silicon/M1 (`darwin/arm64`) this plugin 
+will not work, more info: [kubernetes-sigs/kustomize#4612](https://github.com/kubernetes-sigs/kustomize/issues/4612)).
+</aside> 
+
 Note that projects such as [Operator-sdk][sdk] consume the Kubebuilder project as a lib and provide options to work with other languages
 like Ansible and Helm. The kustomize plugin allows them to easily keep a maintained configuration and ensure that all languages have
 the same configuration. It is also helpful if you are looking to provide nice plugins which will perform changes on top of
--- a/pkg/plugins/common/kustomize/v1/init.go
+++ b/pkg/plugins/common/kustomize/v1/init.go
@@ -20,8 +20,10 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/pflag"
 
 	"sigs.k8s.io/kubebuilder/v3/pkg/config"
@@ -33,6 +35,11 @@ import (
 
 var _ plugin.InitSubcommand = &initSubcommand{}
 
+// Verify if the local environment is supported by this plugin
+var supportedArchs = []string{"linux/amd64",
+	"linux/arm64",
+	"darwin/amd64"}
+
 type initSubcommand struct {
 	config config.Config
 
@@ -43,10 +50,17 @@ type initSubcommand struct {
 }
 
 func (p *initSubcommand) UpdateMetadata(cliMeta plugin.CLIMetadata, subcmdMeta *plugin.SubcommandMetadata) {
-	subcmdMeta.Description = `Initialize a common project including the following files:
+	subcmdMeta.Description = fmt.Sprintf(`Initialize a common project including the following files:
   - a "PROJECT" file that stores project configuration
   - several YAML files for project deployment under the "config" directory
-`
+
+  NOTE: The kustomize/v1 plugin used to do this scaffold uses the v3 release (%s).
+Therefore, darwin/arm64 is not supported since Kustomize does not provide v3
+binaries for this architecture. The currently supported architectures are %q. 
+More info: https://github.com/kubernetes-sigs/kustomize/issues/4612.
+
+`, KustomizeVersion, supportedArchs)
+
 	subcmdMeta.Examples = fmt.Sprintf(`  # Initialize a common project with your domain and name in copyright
   %[1]s init --plugins common/v3 --domain example.org
 
@@ -92,6 +106,37 @@ func (p *initSubcommand) InjectConfig(c config.Config) error {
 	}
 
 	return nil
+}
+
+func (p *initSubcommand) PreScaffold(machinery.Filesystem) error {
+	arch := runtime.GOARCH
+	// It probably will never return x86_64. However, we are here checking the support for the binaries
+	// So that, x86_64 means getting the Linux/amd64 binary. Then, we just keep this line to ensure
+	// that it complies with the same code implementation that we have in the targets. In case someone
+	// call the command inform the GOARCH=x86_64 then, we will properly handle the scenario
+	// since it will work successfully and will instal the Linux/amd64 binary via the Makefile target.
+	arch = strings.Replace(arch, "x86_64", "amd64", -1)
+	localPlatform := fmt.Sprintf("%s/%s", strings.TrimSpace(runtime.GOOS), strings.TrimSpace(arch))
+
+	if !hasSupportFor(localPlatform) {
+		log.Warnf("the platform of this environment (%s) is not suppported by kustomize v3 (%s) which is "+
+			"used in this scaffold. You will be unable to download a binary for the kustomize version supported "+
+			"and used by this plugin. The currently supported platforms are: %q",
+			localPlatform,
+			KustomizeVersion,
+			supportedArchs)
+	}
+
+	return nil
+}
+
+func hasSupportFor(localPlatform string) bool {
+	for _, value := range supportedArchs {
+		if value == localPlatform {
+			return true
+		}
+	}
+	return false
 }
 
 func (p *initSubcommand) Scaffold(fs machinery.Filesystem) error {


pkg/plugins/common/kustomize/v1/init.go
```diff
@@ -94,6 +107,50 @@ func (p *initSubcommand) InjectConfig(c config.Config) error {
...
+	archFound := strings.ToLower(string(arch))
+	archFound = strings.Replace(archFound, "x86_64", "amd64", -1)
+	if err != nil {
+		log.Warnf("unable to check the architecture to verify if kustomize/v1 can be supported on "+
+			"this environment: %s", err)
+		return nil
+	}
+	localPlatform := fmt.Sprintf("%s/%s", strings.TrimSpace(osFound),
+		strings.TrimSpace(archFound))
+
```
everettraven, line 133: It might be easiest to do something like:
```go
localPlatform := fmt.Sprintf("%s/%s", strings.TrimSpace(runtime.GOOS), strings.TrimSpace(runtime.GOARCH))
```
This way you can get rid of lines 112 - 130.

WDYT?

pkg/plugins/common/kustomize/v1/init.go
```diff
@@ -94,6 +107,50 @@ func (p *initSubcommand) InjectConfig(c config.Config) error {
...
+	for _, value := range supportedArchs {
+		if value == localPlatform {
+			found = true
+			break
+		}
+	}
+
+	if !found {
+		log.Warnf("the architecture of this environment (%s) is not suppported by kustomize v3 (%s) which is "+
+			"used in this scaffold. You will be unable to download a binary for this kustomize version supported "+
```
everettraven, line 144:
```suggestion
			"used in this scaffold. You will be unable to download a binary for the kustomize version supported "+
```

pkg/plugins/common/kustomize/v1/init.go
```diff
@@ -94,6 +107,50 @@ func (p *initSubcommand) InjectConfig(c config.Config) error {
...
+			"and used by this plugin. The current supported architectures are: %q",
```
everettraven, line 145:
```suggestion
			"and used by this plugin. The currently supported architectures are: %q",
```

everettraven (changes requested):
First round. I like the idea of improving the error message to be more descriptive!

pkg/plugins/common/kustomize/v1/init.go
```diff
@@ -94,6 +107,50 @@ func (p *initSubcommand) InjectConfig(c config.Config) error {
...
+	// Check what is the ARCH
+	cmd = exec.Command("uname", "-m")
+	arch, err := cmd.CombinedOutput()
+	archFound := strings.ToLower(string(arch))
+	archFound = strings.Replace(archFound, "x86_64", "amd64", -1)
+	if err != nil {
+		log.Warnf("unable to check the architecture to verify if kustomize/v1 can be supported on "+
+			"this environment: %s", err)
+		return nil
+	}
```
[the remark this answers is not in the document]
camilamacedo86: +1

pkg/plugins/common/kustomize/v1/init.go
camilamacedo86: The replacement for the goarch is still required but yep. We follow up on your suggestion here as well.
Great !

camilamacedo86:
@everettraven thank you for the excellent review!

docs/book/src/plugins/kustomize-v1.md
```diff
@@ -4,6 +4,19 @@ The kustomize plugin allows you to scaffold all kustomize manifests used to work
...
+<aside class="note">
+<h1>Supportability</h1>
+
+This plugin uses [kubernetes-sigs/Kustomize](https://github.com/kubernetes-sigs/kustomize) v3 and the architectures supported are: 
+- linux/amd64
+- linux/arm64
+- darwin/amd64
+
+You might want to consider using [kustomize/v2-alpha](./kustomize-v2-alpha.md) if you are looking to scaffold projects in
+others architecture environments. (i.e. if you are looking to scaffold projects with Apple Silicon/M1 (`darwin/arm64`) this plugin
```
everettraven, line 16: Nit:
```suggestion
other architecture environments. (i.e. if you are looking to scaffold projects with Apple Silicon/M1 (`darwin/arm64`) this plugin 
```

pkg/plugins/common/kustomize/v1/init.go
```diff
@@ -43,10 +50,17 @@ type initSubcommand struct {
 }
 
 func (p *initSubcommand) UpdateMetadata(cliMeta plugin.CLIMetadata, subcmdMeta *plugin.SubcommandMetadata) {
-	subcmdMeta.Description = `Initialize a common project including the following files:
+	subcmdMeta.Description = fmt.Sprintf(`Initialize a common project including the following files:
   - a "PROJECT" file that stores project configuration
   - several YAML files for project deployment under the "config" directory
-`
+
+  NOTE: The kustomize/v1 plugin used to do this scaffold use its v3 release (%s).
```
everettraven, line 57: Nit:
```suggestion
  NOTE: The kustomize/v1 plugin used to do this scaffold uses the v3 release (%s).
```

pkg/plugins/common/kustomize/v1/init.go
```diff
@@ -43,10 +50,17 @@ type initSubcommand struct {
...
+Therefore, darwin/arm64 is not supported since the project does not provide
```
everettraven, line 58: Just to be a bit more specific:
```suggestion
Therefore, darwin/arm64 is not supported since Kustomize does not provide v3
```

pkg/plugins/common/kustomize/v1/init.go
```diff
@@ -94,6 +108,30 @@ func (p *initSubcommand) InjectConfig(c config.Config) error {
...
+	found := false
+	for _, value := range supportedArchs {
+		if value == localPlatform {
+			found = true
+			break
+		}
+	}
+
+	if !found {
+		log.Warnf("the architecture of this environment (%s) is not suppported by kustomize v3 (%s) which is "+
```
everettraven, line 124: Might be nicer to say platform rather than architecture here since we are mentioning which platforms are supported (i.e. linux/amd64 rather than just amd64)
```suggestion
		log.Warnf("the platform of this environment (%s) is not suppported by kustomize v3 (%s) which is "+
```

pkg/plugins/common/kustomize/v1/init.go
```diff
@@ -94,6 +108,30 @@ func (p *initSubcommand) InjectConfig(c config.Config) error {
...
+			"used in this scaffold. You will be unable to download a binary for the kustomize version supported "+
+			"and used by this plugin. The currently supported architectures are: %q",
```
everettraven, line 126: Same
```suggestion
			"and used by this plugin. The currently supported platforms are: %q",
```

pkg/plugins/common/kustomize/v1/init.go
```diff
@@ -94,6 +108,30 @@ func (p *initSubcommand) InjectConfig(c config.Config) error {
 	return nil
 }
 
+func (p *initSubcommand) PreScaffold(machinery.Filesystem) error {
+	arch := runtime.GOARCH
+	arch = strings.Replace(arch, "x86_64", "amd64", -1)
```
everettraven, line 113: Is this replace necessary? I don't think `x86_64` is a valid value for `GOARCH` so `runtime.GOARCH` should default a `x86_64` system to a value of `amd64`.

For reference, if I run this Go code:
```go
package main

import (
	"fmt"
	"runtime"
)

func main() {
	fmt.Println("PLATFORM:", runtime.GOOS+"/"+runtime.GOARCH)
}
```
with the following command:
```
go run main.go
```
I get this output (on a x86_64 arch):
```
PLATFORM: linux/amd64
```

However if I run it with:
```
GOARCH=x86_64 go run main.go
```
I get the following output:
```
cmd/go: unsupported GOOS/GOARCH pair linux/x86_64
```

This makes me think that we should not support the case where GOARCH is set to be `x86_64` because then a user is explicitly setting their Go environment to be invalid by setting GOARCH to an invalid value.

WDYT?

everettraven (changes requested):
A couple minor nits and a question regarding the `x86_64` replaces

pkg/plugins/common/kustomize/v1/init.go
everettraven: Another thing to note, when running `go tool dist list` the output does not contain any `x86_64` and the output is:
```
aix/ppc64
android/386
android/amd64
android/arm
android/arm64
darwin/amd64
darwin/arm64
dragonfly/amd64
freebsd/386
freebsd/amd64
freebsd/arm
freebsd/arm64
illumos/amd64
ios/amd64
ios/arm64
js/wasm
linux/386
linux/amd64
linux/arm
linux/arm64
linux/mips
linux/mips64
linux/mips64le
linux/mipsle
linux/ppc64
linux/ppc64le
linux/riscv64
linux/s390x
netbsd/386
netbsd/amd64
netbsd/arm
netbsd/arm64
openbsd/386
openbsd/amd64
openbsd/arm
openbsd/arm64
openbsd/mips64
plan9/386
plan9/amd64
plan9/arm
solaris/amd64
windows/386
windows/amd64
windows/arm
windows/arm64
```

camilamacedo86: the support here is regarding the binaries (have or not the bin to download)
If you check our makefile target, we are replacing linux/x86_64 by linux/amd64 because this platform is built with this name, so I'd prefer to keep the consistence.

camilamacedo86, line 113:
```suggestion
        // It probably will never return x86_64. However, we are here checking the support for the binaries
        // So that, x86_64 means getting the Linux/amd64 binary. Then, we just keep this line to ensure
        // that it complies with the same code implementation that we have in the targets. In case someone
        // call the command inform the GOARCH=x86_64 then, we will properly handle the scenario
        // since it will work successfully and will instal the Linux/amd64 binary via the Makefile target.
	arch = strings.Replace(arch, "x86_64", "amd64", -1)
```

everettraven (approved):
/lgtm

pkg/plugins/common/kustomize/v1/init.go
everettraven: @camilamacedo86 this looks good, feel free to mark it as resolved

docs/book/src/plugins/kustomize-v1.md
```diff
@@ -4,6 +4,19 @@ The kustomize plugin allows you to scaffold all kustomize manifests used to work
 By using the kustomize plugin, you can create your own language plugins and ensure that you will have the same configurations 
 and features provided by it. 
 
+<aside class="note">
+<h1>Supportability</h1>
+
+This plugin uses [kubernetes-sigs/Kustomize](https://github.com/kubernetes-sigs/kustomize) v3 and the architectures supported are:
```
FillZpp, line 10: kubernetes-sigs/kustomize

docs/book/src/plugins/kustomize-v1.md
```diff
@@ -4,6 +4,19 @@ The kustomize plugin allows you to scaffold all kustomize manifests used to work
...
+<h1>Supportability</h1>
+
+This plugin uses [kubernetes-sigs/Kustomize](https://github.com/kubernetes-sigs/kustomize) v3 and the architectures supported are: 
+- linux/amd64
+- linux/arm64
+- darwin/amd64
+
+You might want to consider using [kustomize/v2-alpha](./kustomize-v2-alpha.md) if you are looking to scaffold projects in
+other architecture environments. (i.e. if you are looking to scaffold projects with Apple Silicon/M1 (`darwin/arm64`) this plugin 
+will not work, more info: [kubernetes-sigs/kustomize/issues/4612](https://github.com/kubernetes-sigs/kustomize/issues/4612)).
```
FillZpp, line 17: kubernetes-sigs/kustomize#4612 may be better?

pkg/plugins/common/kustomize/v1/init.go
```diff
@@ -94,6 +108,35 @@ func (p *initSubcommand) InjectConfig(c config.Config) error {
...
+	// So that, x86_64 means getting the Linux/amd64 binary. Then, we just keep this line to ensure
+	// that it complies with the same code implementation that we have in the targets. In case someone
+	// call the command inform the GOARCH=x86_64 then, we will properly handle the scenario
+	// since it will work successfully and will instal the Linux/amd64 binary via the Makefile target.
+	arch = strings.Replace(arch, "x86_64", "amd64", -1)
+	localPlatform := fmt.Sprintf("%s/%s", strings.TrimSpace(runtime.GOOS), strings.TrimSpace(arch))
+	found := false
+	for _, value := range supportedArchs {
+		if value == localPlatform {
+			found = true
```
FillZpp, line 123: How about just return here, so we don't have to use `found`.

pkg/plugins/common/kustomize/v1/init.go
```diff
@@ -94,6 +108,35 @@ func (p *initSubcommand) InjectConfig(c config.Config) error {
...
+	if !found {
+		log.Warnf("the platform of this environment (%s) is not suppported by kustomize v3 (%s) which is "+
+			"used in this scaffold. You will be unable to download a binary for the kustomize version supported "+
+			"and used by this plugin. The currently supported platforms are: %q",
+			localPlatform,
+			KustomizeVersion,
+			supportedArchs)
+	}
+
+	return nil
```
FillZpp, line 137: It seems the error return of this function will always be nil?

FillZpp (commented):
Just some nits :)

pkg/plugins/common/kustomize/v1/init.go
camilamacedo86: Yes, because we will never return an error here so far.
See that it is an implementation of the interface:

```
// PreScaffold executes tasks before the main scaffolding.
	PreScaffold(machinery.Filesystem) error
```

camilamacedo86:
Hi @[FillZpp](https://github.com/FillZpp), 

Thank you for the review.
All addressed :-)

PR #2700, opened 2022-05-25
⚠️ (go/v3): applying restrictive SCC for all containers scaffolded by the tool (k8s versions < 1.19 does not work with)

camilamacedo86:
**Description**
(go/v3): applying restrictive SCC for all containers scaffolded by the tool

**Note**
If applied on clusters < 1.19 the error `unknown field "seccompProfile" in io.k8s.api.core.v1.PodSecurityContext` will be faced. Users still able to remove the seccompProfile if they need to provide patch releases for old versions

**Motivation**
Following good practices and ensuring containers are restrictive. (We want to ensure that all scaffold by default is restrictive independently if the cluster-admin enable or not for example  SeccompDefault)

> Heavily restricted policy, following current Pod hardening best practices."

More info: https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted

**Nice to know:**  from 1.22 was introduced for the SeccompDefault feature in kubelet, which allows cluster admins also set it for all pods, see: https://kubernetes.io/docs/tutorials/security/seccomp/#enable-the-use-of-runtimedefault-as-the-default-seccomp-profile-for-all-workloads

### DIFF
--- a/testdata/project-v3-addon/config/manager/manager.yaml
+++ b/testdata/project-v3-addon/config/manager/manager.yaml
@@ -26,6 +26,8 @@ spec:
     spec:
       securityContext:
         runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
       containers:
       - command:
         - /manager
@@ -35,6 +37,9 @@ spec:
         name: manager
         securityContext:
           allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - ALL
         livenessProbe:
           httpGet:
             path: /healthz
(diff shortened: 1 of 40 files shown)


testdata/project-v3-addon/config/manager/manager.yaml
```diff
@@ -26,6 +26,8 @@ spec:
     spec:
       securityContext:
         runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
```
camilamacedo86, line 30: /hold

We need to check if it will works well from 1.19

FillZpp: Yeah, it works for 1.19+. The `SeccompDefault` featureGate provided in 1.22+ is a global gate, which enables all Pods to use `seccompProfile= RuntimeDefault`, so we don't have to care about it here.

camilamacedo86: For the default scaffold, we want to ensure that all scaffold by the tool is restrictive (we do not want really on the cluster-admin to follow the good practice)

However, is nice to know that the cluster-admin is also able to enable that for all Pods. It means no harm to ensure as well in case the admin does not follow it. 

Thank you a lot.

Issue #2664, opened 2022-04-30
Build kubebuilder-tools for darwin/arm64 


camilamacedo86:
### What do you want to happen?

**Description**

The goal of this task is to begin to generate the kubebuilder-tools with darwin/arm64 in order to support Apple Silicon. The artefacts must be produced and be available in: https://storage.googleapis.com/kubebuilder-tools

Kubebuilder projects require artefacts which are used to do the tests with ENV TEST (when we call `make test` target). The build is made from the branch [tools-releases](https://github.com/kubernetes-sigs/kubebuilder/tree/tools-releases) and the trigger will call the `build/cloudbuild_tools.yaml` passing 
as argument the architecture and the SO that should be used, e.g:

<img width="553" alt="Screenshot 2022-04-30 at 10 15 41" src="https://user-images.githubusercontent.com/7708031/166099666-ae9cd2df-73fe-47f6-a987-464f63df9a19.png">

For further information see the [README](https://github.com/kubernetes-sigs/kubebuilder/blob/tools-releases/README.md).

**What was done**

- The trigger on the Google Cloud to generate the artefact is created.  

**What do you still needing to do**

The required etcd binary does not exist for this format:

```
Step #0:  ---> Running in 161150d5d6d4
Step #0: Archive:  etcd-v3.5.1-darwin-arm64.zip
Step #0:   End-of-central-directory signature not found.  Either this file is not
Step #0:   a zipfile, or it constitutes one disk of a multi-part archive.  In the
Step #0:   latter case the central directory and zipfile comment will be found on
Step #0:   the last disk(s) of this archive.
Step #0: unzip:  cannot find zipfile directory in one of etcd-v3.5.1-darwin-arm64.zip or
Step #0:         etcd-v3.5.1-darwin-arm64.zip.zip, and cannot find etcd-v3.5.1-darwin-arm64.zip.ZIP, period.
Step #0: The command '/bin/sh -c curl -sLO https://github.com/coreos/etcd/releases/download/${ETCD_VERSION}/${ETCD_BASE_NAME}.zip &&     unzip -o ${ETCD_BASE_NAME}.zip &&     cp ${ETCD_BASE_NAME}/etcd $DEST' returned a non-zero code: 9
Finished Step #0
ERROR
ERROR: build step 0 "gcr.io/cloud-builders/docker" failed: step exited with non-zero status: 9
Step #0: 
```

The following options to address this requirement are:
[...]

PR #2706, opened 2022-05-31
:sparkles: Update Makefile for go/v3 to support darwin/arm64

everettraven:
## Description of Change
- Add Bash scripts for building `kustomize` and `etcd` locally
- Add a Makefile target to build `etcd` locally with `make etcd`
- Modify the `kustomize` Makefile target to build from source locally if running on `darwin/arm64` (`Darwin` as result from `uname -s` and `aarch64` as result from `uname -m`)

## Motivation for Change
There is seemingly more and more issues occurring for users that are using the M1 Macs to build operators with both Kubebuilder and Operator SDK. This change should provide users an easier way to remediate the issues by providing a new Makefile target to build `etcd` from source and modifying the existing Makefile target for installing `kustomize` to build from source if running on `darwin/arm64`.

##  Additional Context
I do not have an M1 Mac and therefore cannot test this change myself to verify it resolves issues for users on the new M1 Macs. I was however able to test that the scripts added work as expected.

### DIFF
--- a/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
@@ -179,7 +179,13 @@ KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/k
 .PHONY: kustomize
 kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
 $(KUSTOMIZE): $(LOCALBIN)
+ifeq "$(uname -s)" "Darwin"
+ifeq "$(uname -m)" "aarch64"
+	curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kubebuilder/master/scripts/build-kustomize.sh" | bash -s -- $(KUSTOMIZE_VERSION) $(LOCALBIN)
+endif
+else
 	curl -s $(KUSTOMIZE_INSTALL_SCRIPT) | bash -s -- $(subst v,,$(KUSTOMIZE_VERSION)) $(LOCALBIN)
+endif
 
 .PHONY: controller-gen
 controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary.
@@ -190,4 +196,10 @@ $(CONTROLLER_GEN): $(LOCALBIN)
 envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(LOCALBIN)
 	GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+
+
+ETCD_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/kubebuilder/master/scripts/"
+.PHONY: etcd
+etcd: ## build etcd locally
+	curl -s $(ETCD_SCRIPT) | bash -s
 `
--- a/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
@@ -181,7 +181,7 @@ kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
 $(KUSTOMIZE): $(LOCALBIN)
 ifeq "$(uname -s)" "Darwin"
 ifeq "$(uname -m)" "aarch64"
-	curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kubebuilder/master/scripts/build-kustomize.sh" | bash -s -- $(KUSTOMIZE_VERSION) $(LOCALBIN)
+	curl -s "https://raw.githubusercontent.com/everettraven/kubebuilder/feat/temp-arm-support/scripts/build-kustomize.sh" | bash -s -- $(KUSTOMIZE_VERSION) $(LOCALBIN)
 endif
 else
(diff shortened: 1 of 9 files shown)


everettraven:
/cc @camilamacedo86 @ryantking

Do either of you have an M1 Mac and are able to test this and see if it works? If not do you know anyone who may be able to?

Thanks!

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
```diff
@@ -179,15 +179,35 @@ KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/k
 .PHONY: kustomize
 kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
 $(KUSTOMIZE): $(LOCALBIN)
+ifeq "$(uname -s)" "Darwin"
```
camilamacedo86, line 182: It is hard to keep maintained and not sustainable ( Spaghetti Code ) 
Also, we cannot do that. We cannot add a script from your repo in the default scaffold. 

**About etcd bin**

The etcd bin is used for ENV TEST and is shipped via the kubebuilder-tools, see: 
- https://github.com/kubernetes-sigs/kubebuilder/tree/tools-releases 
- https://storage.googleapis.com/kubebuilder-tools

Also, we are close to having that bin from the project:
- https://github.com/etcd-io/etcd/issues/14001#issuecomment-1146800339

As soon we have the binary there we can trigger the job in the cloud to build the kubebuilder-tools for darwin/arm64. To better understand it check the issue: https://github.com/kubernetes-sigs/kubebuilder/issues/2664

**About kustomize bin**

You can follow up: https://github.com/kubernetes-sigs/kustomize/issues/4612
Users will be able to use it with kustomize/v2 and go/v4-alpha, see: https://github.com/kubernetes-sigs/kubebuilder/pull/2552

I am not sure if we should assume the decomposability to build kustomize v2 for darwin/arm64 to make it work with go/v3 stable. That will add this kind of complexity to the Makefile and also we would need to properly build it in the cloud as we do with the kubebuilder-tools instead. It seems out of the scope from Kubebuilder itself.

However, we could doc the script for example to allow users to build it by themselves and doc as an option. But the script also needs to leave in the repo or some official place and not in our fork. 

To know more about what is required to support apple silicon and what was done or not so far see: https://github.com/kubernetes-sigs/kubebuilder/issues/1932

everettraven: Sounds good. I think at least adding some docs for a workaround would be good so that users can begin using Kubebuilder/Operator SDK to develop operators on apple silicon. 

Just as an aside, the goal of this PR is to add those scripts to Kubebuilder as well so that the scripts that are run in the makefile are from the official Kubebuilder repo, I just left it pointing to my fork so we could test that it actually works since Kubebuilder doesn't have the scripts in place yet and would modify it to point to Kubebuilder prior to merging.

If this isn't a direction that we want to explore that is fine and I can close this PR, just thought I would give it a shot at trying to provide a solution that would allow development with go/v3 on Apple Silicon. Thanks for the feedback!

camilamacedo86: > Sounds good. I think at least adding some docs for a workaround would be good so that users can begin using Kubebuilder/Operator SDK to develop operators on apple silicon.

- For the Kubebuilder side we can begin to support Apple Silicon from go/v4-alpha and with kustomize/v2-alpha
- For SDK we need to discuss how we will do it there? Could we accept the breaking change for ansible/helm/hybrid instead of creating a new plugin or not for example?

> Just as an aside, the goal of this PR is to add those scripts to Kubebuilder as well so that the scripts that are run in the makefile are from the official Kubebuilder repo, I just left it pointing to my fork so we could test that it actually works since Kubebuilder doesn't have the scripts in place yet and would modify it to point to Kubebuilder prior to merging.

I think we have some options to address these ones, follow some ideas/suggestion:

**Regards the bins used in kubebuilder-tools/env test:**

a) We could Improve the kubebuilder-tools scripts for if the bin does not exist we build them for the arch informed so that it can work in the cloud and can be used by the authors. Also, we can improve the doc: https://book.kubebuilder.io/reference/artifacts.html and let the users know about that and that they can use it to build for any platform 

**Regards kustomize**

- For the next release, we will need to clarify its support with go/v4-alpha in the quickstart probably, I am looking if we can merge https://github.com/kubernetes-sigs/kubebuilder/pull/2552 so that we can work on this
- Also, we could add a new section FAQ as we have in SDK docs ( see that we have many issues track to do that and add the info there ) then, we can have a directory under it called script and add the script that will let users know that they can use to build kustomize v2 for apple silicon or for any other platform that they need. 

**Regards SDK**
After we add the changes here in SDK we can use/ link and redirect users to not need to duplicate the content and be required to keep maintained two places

WDYT? Is that make sense?
Could you shape your scripts so we provide them to the users via the FAQ? Would you like to contribute with the kubebuilder-tools scripts to make them be able to build the project when the bin is not found?

everettraven:
Closing as per discussion in this  review comment thread: https://github.com/kubernetes-sigs/kubebuilder/pull/2706#discussion_r891847369

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
camilamacedo86: Sure, if you wish we can track issues with that as well. 
and by sure .. feel free to ping if you need any help

PR #2714, opened 2022-06-07
:sparkles: add support to go 1.18 and for (go/v3), update envtest version to use k8s 1.24

camilamacedo86:
**Description**
- add support to go 1.18
- update envtest version for k8s 1.24
- update goreleaser, ci and tests to ensure that works with 1.18.

PS.: Note that we are here only bumping the scaffolds and not the Kubebuilder tool since we do not want for now to impact who consumes the tool. But we are also updating the tests and the versions used to build the tool to ensure that it can work with 1.18

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/2559
Closes: https://github.com/kubernetes-sigs/kubebuilder/pull/2561

### DIFF
--- a/testdata/project-v3-v1beta1/Makefile
+++ b/testdata/project-v3-v1beta1/Makefile
@@ -2,7 +2,7 @@
 # Image URL to use all building/pushing image targets
 IMG ?= controller:latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.24.1
+ENVTEST_K8S_VERSION = 1.21
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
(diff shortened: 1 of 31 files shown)


testdata/project-v3-v1beta1/Makefile
```diff
@@ -2,7 +2,7 @@
 # Image URL to use all building/pushing image targets
 IMG ?= controller:latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.24.1
+ENVTEST_K8S_VERSION = 1.21
```
dims, line 5: is this going backwards? 1.24.1 -> 1.21?

camilamacedo86: Yes. A project that is scaffold with v1beta1 cannot work with k8s >= 1.22 so that cannot work with 1.24.1.
This sample is done using the crdversions v1beta1 and when this option is used we downgrade the scaffold to warn the users.

camilamacedo86: Note that the sample testdata/project-v3-v1beta1 is built using CRD v1beta1 and the optional flag crd-version=v1beta1

dims: gotcha!

PR #2717, opened 2022-06-10
✨New deploy-image/v1-alpha plugin

NikhilSharmaWe:
### Description
- This is the initial version of the **deploy-image** plugin. See https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/code-generate-image-plugin.md for getting more info about the future goals for this plugin.
- Next implementations to accomplish the goal of the [EP](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/code-generate-image-plugin.md) will be done in follow-ups.


pkg/plugins/golang/deploy-image/v1beta1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,218 @@
...
+				},
+				Spec: corev1.PodSpec{
+					SecurityContext: &corev1.PodSecurityContext{
+						RunAsNonRoot: &[]bool{true}[0],
+						SeccompProfile: &corev1.SeccompProfile{
+							Type: corev1.SeccompProfileTypeRuntimeDefault,
+						},
+					},
+					Containers: []corev1.Container{{
+						Image:   {{ .Image }},
```
NikhilSharmaWe, line 180: @camila This statement is giving an error:

> failed to create API: unable to scaffold with "deploy-image.go.kubebuilder.io/v1-beta": error updating scaffold: template: *controllers.Controller:119: function "Image" not defined

 when I run: 
`kubebuilder create api --plugins=deploy-image/v1-beta --kind Guestbook --version v1 --group webapp --image mongo`

What could be the reason because `{{ .Image }}` added [here](https://github.com/kubernetes-sigs/kubebuilder/pull/2717/files#diff-5c108238fc057fcff25517973431e59f0970f322b85b0748ccce74b22206d7cbR99) in the comments does not give the error. Does the error is related to the **field** Image instead of the value used `{{ .Image }}`?

NikhilSharmaWe, line 180: @camila This statement is giving an error:

> failed to create API: unable to scaffold with "deploy-image.go.kubebuilder.io/v1-beta": error updating scaffold: template: *controllers.Controller:119: function "Image" not defined

 when I run: 
`kubebuilder create api --plugins=deploy-image/v1-beta --kind Guestbook --version v1 --group webapp --image mongo`

The error is probably due to the fields in `[]corev1.Container` (Image, Name, ImagePullPolicy, etc.). Could you please help me in understanding what am I missing here.

pkg/machinery/scaffold.go
```diff
@@ -116,6 +116,13 @@ func WithResource(resource *resource.Resource) ScaffoldOption {
 	}
 }
 
+// WithResource provides the resource to the Scaffold
+func WithImage(image string) ScaffoldOption {
+	return func(s *Scaffold) {
+		s.injector.image = image
+	}
+}
```
camilamacedo86, line 124: `Very nice 👍

pkg/plugins/golang/deploy-image/v1beta1/api.go
```diff
@@ -0,0 +1,112 @@
...
+	config   config.Config
+	resource *resource.Resource
+	image    string
+}
+
+func (p *createAPISubcommand) UpdateMetadata(cliMeta plugin.CLIMetadata, subcmdMeta *plugin.SubcommandMetadata) {
+	subcmdMeta.Description = `Scaffold the code implementation to deploy and manage your Operand which is represented
+by the API (GVK) informed. Note that you must provide an image.
+
+<TODO: here we will add the help of the plugin>
```
camilamacedo86, line 43: Could we improve the help or remove the TODO for now

pkg/plugins/golang/deploy-image/v1beta1/api.go
```diff
@@ -0,0 +1,112 @@
...
+`
+	subcmdMeta.Examples = fmt.Sprintf(`  # Create a frigates API with Group: ship, Version: v1beta1 and Kind: Frigate
+  %[1]s create api --group ship --version v1beta1 --kind Frigate
+
+   <TODO here we will put the valid examples over how to call the plugin and its next steps>
```
camilamacedo86, line 48: Here we need to add the image flag in the example

pkg/plugins/golang/deploy-image/v1beta1/api.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (p *createAPISubcommand) InjectConfig(c config.Config) error {
+	p.config = c
+
+	return nil
+}
+
+func (p *createAPISubcommand) BindFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&p.image, "image", "", "inform the Operand image. <todo better text here>")
```
camilamacedo86, line 68:
```suggestion
	fs.StringVar(&p.image, "image", "", "inform the Operand image. The controller will be scaffolded with an example code to deploy and manage this image.")
```

pkg/plugins/golang/deploy-image/v1beta1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,229 @@
...
+// {{ .Resource.Kind }}Reconciler reconciles a {{ .Resource.Kind }} object
+type {{ .Resource.Kind }}Reconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+}
+
+//+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }},verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }}/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }}/finalizers,verbs=update
+
```
camilamacedo86, line 93:
```suggestion
//+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
//+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
```

We need to add the above markers otherwise we will not have the required permissions (RBAC) to run the deployment

pkg/plugins/golang/deploy-image/v1beta1/scaffolds/api.go
```diff
@@ -0,0 +1,78 @@
...
+	boilerplate, err := afero.ReadFile(s.fs.FS, filepath.Join("hack", "boilerplate.go.txt"))
+	if err != nil {
+		return fmt.Errorf("error updating scaffold: unable to load boilerplate: %w", err)
+	}
+
+	// Initialize the machinery.Scaffold that will write the files to disk
+	scaffold := machinery.NewScaffold(s.fs,
+		machinery.WithConfig(s.config),
+		machinery.WithBoilerplate(string(boilerplate)),
+		machinery.WithResource(&s.resource),
```
camilamacedo86, line 68: Should we not add here WithImage to pass the image?

pkg/plugins/golang/deploy-image/v1beta1/scaffolds/api.go
```diff
@@ -0,0 +1,78 @@
...
+	)
+
+	err = scaffold.Execute(
+		&controllers.Controller{Image: s.image}, // now we are passing the image for the template
```
camilamacedo86, line 72: then we remove from here

pkg/plugins/golang/deploy-image/v1beta1/scaffolds/api.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return fmt.Errorf("error updating scaffold: unable to load boilerplate: %w", err)
+	}
+
+	// Initialize the machinery.Scaffold that will write the files to disk
+	scaffold := machinery.NewScaffold(s.fs,
+		machinery.WithConfig(s.config),
+		machinery.WithBoilerplate(string(boilerplate)),
+		machinery.WithResource(&s.resource),
+	)
```
camilamacedo86, line 69: Also we need to pass the ControllerRuntimeVersion as well (compare with go/v3 plugin and ensure that we have the same code)

cmd/main.go
```diff
@@ -60,6 +61,7 @@ func main() {
 			&kustomizecommonv1.Plugin{},
 			&kustomizecommonv2.Plugin{},
 			&declarativev1.Plugin{},
+			&deployimagev1beta1.Plugin{},
 		),
```
camilamacedo86, line 65: We need to create a new testdata sample to check the plugin. (it is here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh)
The commands to run the sample are:

```
kubebuilder init 
$kb create api --group crew --version v1 --kind Captain --controller=true --resource=true --make=false --plugins=go/v3,deploy-config/v1-alpha1
```
Then, we need a new dir for that which will be: `project-v3-with-deploy-image`

pkg/plugins/golang/deploy-image/v1beta1/plugin.go
```diff
@@ -0,0 +1,58 @@
...
+	"sigs.k8s.io/kubebuilder/v3/pkg/model/resource"
+	"sigs.k8s.io/kubebuilder/v3/pkg/model/stage"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang"
+)
+
+const pluginName = "deploy-image." + golang.DefaultNameQualifier
+
+var (
+	pluginVersion            = plugin.Version{Number: 1, Stage: stage.Beta}
```
camilamacedo86, line 32: Let's use alpha which is the first expected version instead of beta

pkg/plugins/golang/deploy-image/v1beta1/plugin.go
```diff
@@ -0,0 +1,58 @@
...
+	supportedProjectVersions = []config.Version{cfgv2.Version, cfgv3.Version}
```
camilamacedo86, line 33:
```suggestion
	supportedProjectVersions = []config.Version{cfgv3.Version}
```

The new plugin does not work with the legacy layout only with the new one. 
We do not need to support the legacy

pkg/plugins/golang/deploy-image/v1alpha1/api.go
```diff
@@ -0,0 +1,110 @@
...
+var _ plugin.CreateAPISubcommand = &createAPISubcommand{}
+
+type createAPISubcommand struct {
+	config   config.Config
+	resource *resource.Resource
+	image    string
+}
+
+func (p *createAPISubcommand) UpdateMetadata(cliMeta plugin.CLIMetadata, subcmdMeta *plugin.SubcommandMetadata) {
+	subcmdMeta.Description = `Scaffold the code implementation to deploy and manage your Operand which is represented
```
camilamacedo86, line 40:
```suggestion
        //nolint: lll
	subcmdMeta.Description = `Scaffold the code implementation to deploy and manage your Operand which is represented
```

The error here occurs because the size of the line we can ignore

test/testdata/generate.sh
```diff
@@ -69,6 +69,10 @@ function scaffold_test_project {
       $kb create webhook --group crew --version v1 --kind Admiral --defaulting
     fi
 
+    if [ $project == "project-v3-with-deploy-image" ]; then
+      $kb create api --group crew --version v1 --kind Captain --controller=true --resource=true --make=false
```
camilamacedo86, line 73:
```suggestion
      $kb create api --group crew --version v1 --kind Captain --controller=true --resource=true
      $make manifests
```
Can we check with? 
In this case, we probably want to run make manifests to see the rules and etc.

pkg/plugins/golang/deploy-image/v1alpha1/api.go
```diff
@@ -0,0 +1,110 @@
...
+by the API (GVK) informed. Note that you must provide an image.
+`
+	subcmdMeta.Examples = fmt.Sprintf(`  # Create a frigates API with Group: ship, Version: v1beta1 Kind: Frigate and Image: mongo
```
camilamacedo86, line 43:
```suggestion
	subcmdMeta.Examples = fmt.Sprintf(`  # Create a frigates API with Group: ship, Version: v1beta1 Kind: Frigate to represent the Image: example.com/frigate:v0.0.1 and the controller with a code to deploy and manage this Operand. 
```

pkg/plugins/golang/deploy-image/v1alpha1/api.go
```diff
@@ -0,0 +1,110 @@
...
+  %[1]s create api --group ship --version v1beta1 --kind Frigate --image mongo
```
camilamacedo86, line 44:
```suggestion
  %[1]s create api --group ship --version v1beta1 --kind Frigate --image example.com/frigate:v0.0.1
```

pkg/plugins/golang/deploy-image/v1alpha1/api.go
```diff
@@ -0,0 +1,110 @@
...
+
+   <TODO here we will put the valid examples over how to call the plugin and its next steps>
```
camilamacedo86, line 46:
```suggestion
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return fmt.Errorf("error updating scaffold: unable to load boilerplate: %w", err)
+	}
+
+	// Initialize the machinery.Scaffold that will write the files to disk
+	scaffold := machinery.NewScaffold(s.fs,
+		machinery.WithConfig(s.config),
+		machinery.WithBoilerplate(string(boilerplate)),
+		machinery.WithResource(&s.resource),
+		machinery.WithImage(s.image),
```
camilamacedo86, line 70:
```suggestion
		machinery.WithImage(s.image),
		machinery.WithControllerRuntime(scaffolds.ControllerRuntimeVersion)
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
```diff
@@ -0,0 +1,80 @@
...
+	)
+
+	err = scaffold.Execute(
+		&controllers.Controller{ControllerRuntimeVersion: scaffolds.ControllerRuntimeVersion, Image: s.image},
```
camilamacedo86, line 74: We are using the WithImage above so by creating the method WithControllerRuntime we do not need to pass the values directly to the constructor here.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,231 @@
+/*
```
camilamacedo86, line 1: We will also need to scaffold the api file such as we do in the go/v3: https://github.com/kubernetes-sigs/kubebuilder/blob/master. So that we can add the spec size there too. 

We need to add the dir api at the same level as the controller
and copy the types.go for it.

See that we will remove the bool force https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/scaffolds/internal/templates/api/types.go#L36
and we will have only the option: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/scaffolds/internal/templates/api/types.go#L58

So that we can Overwrite the api file with this plugin (not that we always call the api from the go/v3 and then this implementation so the api file from go will be scaffolded and our goal here is to replace/overwrite that by this plugin implementation)

Then, note that we will change the content of the boilerplate here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/scaffolds/internal/templates/api/types.go#L82-L83

So that we can scaffold the size: https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator/api/v1alpha1/memcached_types.go#L33

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,231 @@
...
+	{{ .Resource.ImportAlias }} "{{ .Resource.Path }}"
+	{{- end }}
+)
+
+// {{ .Resource.Kind }}Reconciler reconciles a {{ .Resource.Kind }} object
+type {{ .Resource.Kind }}Reconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+}
+
```
camilamacedo86, line 89:
```suggestion
// The following markers are used to generate the rules permissions on config/rbac using controller-gen
// when the command `make manifests` is executed. To know more about markers see: https://book.kubebuilder.io/reference/markers.html
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,231 @@
...
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the {{ .Resource.Kind }} object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@{{ .ControllerRuntimeVersion }}/pkg/reconcile
+func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// TODO(user): your logic here
```
camilamacedo86, line 106:
```suggestion
```

We are adding the logic so we do not need the TODO anymore

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,231 @@
...
+    // Here you will add the code to create the deployment 
+    // And use the image {{ .Image }} on it
```
camilamacedo86, line 108:
```suggestion
    // The following code will create a deployment for the image {{ .Image }}
    // We are also checking here if the quantity of instances (size) on the cluster is the same
    // quantity expected by watching the changes raised via the events on the cluster and 
    // checking if the size always equals the number informed in the size spec defined in the CR
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,231 @@
...
+	log := ctrllog.FromContext(ctx)
+
+	// Fetch the {{ .Resource.Kind }} instance
```
camilamacedo86, line 111:
```suggestion
	// Fetch the {{ .Resource.Kind }} instance
	// The purpose is check if the Custom Resource for the Kind {{ .Resource.Kind }}
	// is applied on the cluster if not we return nill to stop the reconciliation
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,231 @@
...
+			return ctrl.Result{}, err
+		}
+		// Deployment created successfully - return and requeue
+		return ctrl.Result{Requeue: true}, nil
+	} else if err != nil {
+		log.Error(err, "Failed to get Deployment")
+		return ctrl.Result{}, err
+	}
+
+	// Ensure the deployment size is the same as the spec
```
camilamacedo86, line 146:
```suggestion
	// Ensure the deployment size is the same as the spec
	// The following code will ensure that the quantity of {{ .Kind }} Pod/instances are always as defined
	// in the Custom Resource via the spec size.
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,231 @@
...
+									"ALL",
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	// Set {{ .Resource.Kind }} instance as the owner and controller
```
camilamacedo86, line 209:
```suggestion
	// Set {{ .Resource.Kind }} instance as the owner and controller
	// You should use the method ctrl.SetControllerReference for all resources
	// which are created by your controller so that when the Custom Resource be deleted
	// all resources owned by it (child) will also be deleted.
	// To know more about it see: https://kubernetes.io/docs/tasks/administer-cluster/use-cascading-deletion/
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,231 @@
...
+// SetupWithManager sets up the controller with the Manager.
+func (r *{{ .Resource.Kind }}Reconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		{{ if not (isEmptyStr .Resource.Path) -}}
+		For(&{{ .Resource.ImportAlias }}.{{ .Resource.Kind }}{}).
+		{{- else -}}
+		// Uncomment the following line adding a pointer to an instance of the controlled resource as an argument
+		// For().
+		{{- end }}
+		Complete(r)
```
camilamacedo86, line 229:
```suggestion
                 Owns(&appsv1.Deployment{}). 
		Complete(r)
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,231 @@
...
+	// Ensure the deployment size is the same as the spec
+	size := {{ .Resource.Kind }}.Spec.Size
+	if *found.Spec.Replicas != size {
+		found.Spec.Replicas = &size
+		err = r.Update(ctx, found)
+		if err != nil {
+			log.Error(err, "Failed to update Deployment", "Deployment.Namespace", found.Namespace, "Deployment.Name", found.Name)
+			return ctrl.Result{}, err
+		}
+		// Spec updated - return and requeue
```
camilamacedo86, line 155:
```suggestion
		// Since it fails we want to re-queue the reconciliation
		// The reconciliation will only stop when we be able to ensure 
		// the desired state on the cluster
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil && errors.IsNotFound(err) {
+		// Define a new deployment
+		dep := r.deploymentFor{{ .Resource.Kind }}({{ .Resource.Kind }})
+		log.Info("Creating a new Deployment", "Deployment.Namespace", dep.Namespace, "Deployment.Name", dep.Name)
+		err = r.Create(ctx, dep)
+		if err != nil {
+			log.Error(err, "Failed to create new Deployment", "Deployment.Namespace", dep.Namespace, "Deployment.Name", dep.Name)
+			return ctrl.Result{}, err
+		}
+		// Deployment created successfully - return and requeue
```
camilamacedo86, line 139:
```suggestion
		// Deployment created successfully 
		// We will requeue the reconciliation so that we can ensure the state
		// and move forward for the next operations
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,231 @@
...
+		return ctrl.Result{Requeue: true}, nil
+	} else if err != nil {
+		log.Error(err, "Failed to get Deployment")
+		return ctrl.Result{}, err
```
camilamacedo86, line 143:
```suggestion
                 // Let's return the error for the reconciliation be re-trigged again 
		return ctrl.Result{}, err
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,231 @@
...
+	return dep
+}
+
+// labelsFor{{ .Resource.Kind }} returns the labels for selecting the resources
+// belonging to the given  {{ .Resource.Kind }} CR name.
+func labelsFor{{ .Resource.Kind }}(name string) map[string]string {
+	return map[string]string{"type": " {{ .Resource.Kind }}", " {{ .Resource.Kind }}_cr": name}
+}
+
+// SetupWithManager sets up the controller with the Manager.
```
camilamacedo86, line 220:
```suggestion
// SetupWithManager sets up the controller with the Manager.
// The following code specifies how the controller is built to watch a CR 
// and other resources that are owned and managed by that controller.
// In this way, the reconciliation can be re-trigged when the CR and/or the Deployment
// be created/edit/delete. Note that is possible also create filters and customize
// what events and the circumstances when the reconciliation should be re-queue
// by using predicates for event filtering. More info: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/handler#EventHandler
```

camilamacedo86: If we have time we might want to add one predicate to show how it works

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,231 @@
...
+//+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }}/finalizers,verbs=update
+//+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the {{ .Resource.Kind }} object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
```
camilamacedo86, line 101:
```suggestion

// Note: It is essential for the controller’s reconciliation loop to be idempotent. By following the Operator 
// pattern(https://kubernetes.io/docs/concepts/extend-kubernetes/operator/) you will create
// Controllers(https://kubernetes.io/docs/concepts/architecture/controller/) which provide a reconcile function
// responsible for synchronizing resources until the desired state is reached on the cluster. Breaking this
// recommendation goes against the design principles of Controller-runtime(https://github.com/kubernetes-sigs/controller-runtime) 
// and may lead to unforeseen consequences such as resources becoming stuck and requiring manual intervention.
```

pkg/plugin/subcommand.go
```diff
@@ -92,3 +92,7 @@ type CreateWebhookSubcommand interface {
 type EditSubcommand interface {
 	Subcommand
 }
+
+type CreateSubcommand interface {
+	Subcommand
+}
```
camilamacedo86, line 98: Please, revert this change. We have the interface: 

```go
// CreateAPISubcommand is an interface that represents a `create api` subcommand
type CreateAPISubcommand interface {
	Subcommand
	RequiresResource
}
```

So that we do not want another interface with another name to achieve the same result.
Your plugin must implement CreateAPISubcommand and not  CreateSubcommand 
 
It is a blocker for us to merge this one.

/hold

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
```diff
@@ -0,0 +1,82 @@
...
+		machinery.WithConfig(s.config),
+		machinery.WithBoilerplate(string(boilerplate)),
+		machinery.WithResource(&s.resource),
+		machinery.WithImage(s.image),
+		machinery.WithControllerRuntime(scaffolds.ControllerRuntimeVersion),
+	)
+
+	err = scaffold.Execute(
+		&controllers.Controller{ControllerRuntimeVersion: s.controllerRuntimeVerion, Image: s.image},
+	)
```
camilamacedo86, line 77:
```suggestion
	err = scaffold.Execute(	&controllers.Controller{}, )
```

We are setting the values above WIthImage and WithControllerRuntimeVersion so we do not need to pass here.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/api/types.go
```diff
@@ -0,0 +1,120 @@
...
+// {{ .Resource.Kind }}Spec defines the desired state of {{ .Resource.Kind }}
+type {{ .Resource.Kind }}Spec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// Size defines the number of {{ .Resource.Kind }} instances
+	Size int32 ` + "`" + `json:"size,omitempty"` + "`" + `
+	
+	// Foo is an example field of {{ .Resource.Kind }}. Edit {{ lower .Resource.Kind }}_types.go to remove/update
+	Foo string ` + "`" + `json:"foo,omitempty"` + "`" + `
```
camilamacedo86, line 80:
```suggestion
```

We can remove the foo from here. That is just an example and in our case, we are adding the Size already
Ideally, we will add more specs here, we can add the specs of the values that we hardcode in the deployment in a follow-up with an explanation over.

test/testdata/generate.sh
```diff
@@ -45,7 +45,7 @@ function scaffold_test_project {
   header_text "Initializing project ..."
   $kb init $init_flags --domain testproject.org --license apache2 --owner "The Kubernetes authors"
 
-  if [ $project == "project-v2" ] || [ $project == "project-v3" ] || [ $project == "project-v3-config" ] || [ $project == "project-v3-with-kustomize-v2" ]; then
+  if [ $project == "project-v2" ] || [ $project == "project-v3" ] || [ $project == "project-v3-config" ] || [ $project == "project-v3-with-kustomize-v2" ] || [ $project == "project-v3-with-deploy-image"]; then
```
camilamacedo86, line 48:
```suggestion
  if [ $project == "project-v2" ] || [ $project == "project-v3" ] || [ $project == "project-v3-config" ] || [ $project == "project-v3-with-kustomize-v2" ] ; then
```

we do not want to scaffold these apis for the new testdata sample
Let's have a sample that scaffold only the Memcached Kind

test/testdata/generate.sh
```diff
@@ -132,3 +137,4 @@ scaffold_test_project project-v3-addon --plugins="go/v3,declarative"
 scaffold_test_project project-v3-config --component-config
 scaffold_test_project project-v3-v1beta1
 scaffold_test_project project-v3-with-kustomize-v2 --plugins="kustomize/v2-alpha,base.go.kubebuilder.io/v3"
+scaffold_test_project project-v3-with-deploy-image --plugins="go/v3,deploy-image/v1-alpha"
```
camilamacedo86, line 140: It is not working because we have not the implementation for the init on this plugin 
So let's call the `--plugins="go/v3,deploy-image/v1-alpha"` when we create api above and remove it from here.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,254 @@
...
+		// We will requeue the reconciliation so that we can ensure the state
+		// and move forward for the next operations
+		return ctrl.Result{Requeue: true}, nil
+	} else if err != nil {
+		log.Error(err, "Failed to get Deployment")
+		// Let's return the error for the reconciliation be re-trigged again 
+		return ctrl.Result{}, err
+	}
+
+	// The API is defining that the {{ .Resource.Kind }} type, have a {{ Resource.Kind }}Spec.Size field to set the quantity of {{ Resource.Kind }} instances (CRs) to be deployed.
```
NikhilSharmaWe, line 154: @camilamacedo86 By running create api command this statement is giving error:
```
failed to create API: unable to scaffold with "deploy-image.go.kubebuilder.io/v1-alpha": error updating scaffold: template: *controllers.Controller:93: function "Resource" not defined
```

camilamacedo86: That means we have errors in the boilerplate.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
camilamacedo86: addressed

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
camilamacedo86: done

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
camilamacedo86: done

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,251 @@
...
+				MatchLabels: ls,
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: ls,
+				},
+				Spec: corev1.PodSpec{
+					SecurityContext: &corev1.PodSecurityContext{
+						RunAsNonRoot: &[]bool{true}[0],
+						SeccompProfile: &corev1.SeccompProfile{
```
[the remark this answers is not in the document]
camilamacedo86: done

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
camilamacedo86: done

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
camilamacedo86: done

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/config/samples/crd_sample.go
```diff
@@ -0,0 +1,50 @@
...
+	return nil
+}
+
+const crdSampleTemplate = `apiVersion: {{ .Resource.QualifiedGroup }}/{{ .Resource.Version }}
+kind: {{ .Resource.Kind }}
+metadata:
+  name: {{ lower .Resource.Kind }}-sample
+spec:
+	size:
+`
```
camilamacedo86, line 50:
```suggestion
const crdSampleTemplate = `apiVersion: {{ .Resource.QualifiedGroup }}/{{ .Resource.Version }}
kind: {{ .Resource.Kind }}
metadata:
  name: {{ lower .Resource.Kind }}-sample
spec:
	# TODO(user): edit the following value to ensure the number 
	# of Pods/Instances your Operand must have on cluster        
	size: 3
`
```

test/e2e/v3/generate_test.go
```diff
@@ -244,16 +259,31 @@ func GenerateV3WithKustomizeV2(kbc *utils.TestContext, crdAndWebhookVersion stri
...
+	} else {
+		err = kbc.CreateAPI(
+			"--group", kbc.Group,
+			"--version", kbc.Version,
+			"--kind", kbc.Kind,
+			"--namespaced",
+			"--resource",
+			"--controller",
+			"--make=false",
+			"--crd-version", crdAndWebhookVersion,
```
camilamacedo86, line 284: we do not use this Option for the new plugin 
the crd-version is deprecated and we just need ensure it with the default scaffold
```suggestion
```

test/e2e/v3/plugin_cluster_test.go
```diff
@@ -160,7 +160,64 @@ var _ = Describe("kubebuilder", func() {
...
+				// Skip if cluster version >= 1.22 because pre v1 CRDs and webhooks no longer exist.
+				if srvVer := kbc.K8sVersion.ServerVersion; srvVer.GetMajorInt() <= 1 && srvVer.GetMinorInt() < 19 ||
+					srvVer.GetMajorInt() <= 1 && srvVer.GetMinorInt() >= 22 {
+					Skip(fmt.Sprintf("cluster version %s does not support project defaults "+
+						"and securityContext.seccompProfile", srvVer.GitVersion))
+				}
+
+				GenerateV3(kbc, "v1beta1", true)
 				Run(kbc, sat)
 			})
```
camilamacedo86, line 222: We do not need this one with the new plugin

test/e2e/v3/generate_test.go
```diff
@@ -143,16 +143,30 @@ func GenerateV3(kbc *utils.TestContext, crdAndWebhookVersion string) {
...
-		"--controller",
-		"--make=false",
-		"--crd-version", crdAndWebhookVersion,
-	)
+	if isWithDeployImagePlugin == true {
+		err = kbc.CreateAPI(
+			"--plugins", "go/v3,deploy-image/v1-alpha",
+			"--group", "example.com",
+			"--version", "v1alpha1",
+			"--kind", "Memcached",
```
camilamacedo86, line 151:
```suggestion
			"--group", kbc.Group,
			"--version", kbc.Version,
			"--kind", kbc.kind,
```

test/e2e/v3/generate_test.go
```diff
@@ -244,16 +258,30 @@ func GenerateV3WithKustomizeV2(kbc *utils.TestContext, crdAndWebhookVersion stri
...
```
camilamacedo86, line 266: same here
also, use only the deploy-image/v1-alpha instead of go/v3

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,256 @@
...
+						// If you are looking for to produce solutions to be supported
+						// on lower versions you must remove this option.
+						SeccompProfile: &corev1.SeccompProfile{
+							Type: corev1.SeccompProfileTypeRuntimeDefault,
+						},
+					},
+					Containers: []corev1.Container{
+						Image: "{{ .Image }}",
+						Name:    "{{ .Resource.Kind }}",
+                        ImagePullPolicy: {{ .Resource.Kind }}.Spec.ContainerImagePullPolicy,
```
camilamacedo86, line 205: We have not the spec for ContainerImagePullPolicy so that will only work if we scaffold it in the API spec like we do for the size and add to the samples

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,256 @@
...
+						Command: []string{"{{ .Resource.Kind }}"},
```
camilamacedo86, line 206: That is invalid, can we ensure that any image scaffold will have this command to init a container?
No, so we need to remove the command from the default scaffold and we could fill that only when/if someone informs the command via a new option flag to the plugin `--image-command`, if informed we will use it to scaffold the deployment to create a Deployment/Pod with the Operand image that has this command.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -0,0 +1,256 @@
...
+						Ports: []corev1.ContainerPort{
+							ContainerPort: {{ .Resource.Kind }}.Spec.ContainerPort,
```
camilamacedo86, line 208: Same here. We have not the spec `ContainerPort ` scaffold by default and we cannot use it
So, the Ports we can remove OR allow inform via optional flags

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
NikhilSharmaWe: @camilamacedo86 I have added ContainerImagePullPolicy and ContainerPort in the API spec.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
```diff
@@ -0,0 +1,181 @@
...
+			return fmt.Errorf("error scaffolding container port in the controller: %v", err)
+		}
+	}
+
+	if err := scaffold.Execute(
+		&samples.CRDSample{Command: s.command, Port: s.port},
+	); err != nil {
+		return fmt.Errorf("error updating config/samples: %v", err)
+	}
+
```
camilamacedo86, line 153: We need to check the main.go:

```go vet ./...
# sigs.k8s.io/kubebuilder/testdata/project-v3-with-deploy-image
vet: ./main.go:88:13: undeclared name: examplecomv1alpha1
```
What is missing to update?
How the create API without the plugin works here?

camilamacedo86: See that the e2e fails at the same point : https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kubebuilder/2717/pull-kubebuilder-e2e-k8s-1-19-16/1538508442782666752#1:build-log.txt%3A873

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,194 @@
...
+// 	tokenRequestRawString = `{"apiVersion": "authentication.k8s.io/v1", "kind": "TokenRequest"}`
+// )
+
+// tokenRequest is a trimmed down version of the authentication.k8s.io/v1/TokenRequest Type
+// that we want to use for extracting the token.
+// type tokenRequest struct {
+// 	Status struct {
+// 		Token string `json:"token"`
+// 	} `json:"status"`
+// }
```
camilamacedo86, line 48: remove the code which is commented

testdata/project-v3-with-deploy-image/PROJECT
```diff
@@ -0,0 +1,28 @@
+domain: testproject.org
+layout:
+- go.kubebuilder.io/v3
+plugins:
+  deploy-image.go.kubebuilder.io/v1-alpha:
+    containerCommand: memcached -m=64 modern -v
+    containerPort: "11211"
+    image: memcached:1.6.15-alpine
```
camilamacedo86, line 8: I found a nit issue here.
The data : 

```
containerCommand: memcached -m=64 modern -v
    containerPort: "11211"
    image: memcached:1.6.15-alpine
```

Probably need to be under the resources.
What will be its results if we create 2 apis with diff images and etc?
We would need to have both resources with the options used for each scenario.

Could you please do a test and let us know?
I am ok with we solving it in a follow-up.

testdata/project-v3-with-deploy-image/config/samples/example.com_v1alpha1_memcached.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: memcached-sample
+spec:
+  # TODO(user): edit the following value to ensure the number 
+  # of Pods/Instances your Operand must have on cluster        
+  size: 1
+
+  # TODO(user): edit the following value to ensure the the container has the right command
+  # and parameters to be initialized 
+  containerCommand: 'memcached -m=64 modern -v'
```
camilamacedo86, line 12: We need to fix this one in the sample testdata and e2e tests to be  'memcached -m=64 -o modern -v'

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
```diff
@@ -0,0 +1,227 @@
...
+	golangV3Scaffolder := golangv3scaffolds.NewAPIScaffolder(s.config,
+		s.resource, true)
+	golangV3Scaffolder.InjectFS(s.fs)
+	return golangV3Scaffolder.Scaffold()
+}
+
+const containerTemplate = `Containers: []corev1.Container{{
+						Image: "%s",
+						Name:  "%s",
+						ImagePullPolicy: corev1.PullAlways,
```
camilamacedo86, line 205:
```suggestion
						ImagePullPolicy: corev1.PullIfNotPresent,
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,306 @@
...
+	_, err = kbc.Kubectl.Command("label", "--overwrite", "ns", kbc.Kubectl.Namespace,
+		"pod-security.kubernetes.io/audit=restricted",
+		"pod-security.kubernetes.io/enforce-version=v1.24",
+		"pod-security.kubernetes.io/enforce=restricted")
+	Expect(err).To(Not(HaveOccurred()))
+
+	By("building the controller image")
+	err = kbc.Make("docker-build", "IMG="+kbc.ImageName)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
```
camilamacedo86, line 184:
```suggestion
	By("loading the controller docker image into the kind cluster")
	err = kbc.LoadImageToKindCluster()
	ExpectWithOffset(1, err).NotTo(HaveOccurred())
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,306 @@
...
+	By("validating that manager Pod/container(s) are restricted")
+	ExpectWithOffset(1, outputMake).NotTo(ContainSubstring("Warning: would violate PodSecurity"))
+
+	By("loading the controller docker image into the kind cluster")
+	err = kbc.LoadImageToKindCluster()
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+	By("pulling image")
+	err = kbc.LoadImageToKindClusterWithName(imageCR)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
```
camilamacedo86, line 199:
```suggestion
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+	// NOTE: If you want to run the test against a GKE cluster, you will need to grant yourself permission.
+	// Otherwise, you may see "... is forbidden: attempt to grant extra privileges"
+	// $ kubectl create clusterrolebinding myname-cluster-admin-binding \
+	// --clusterrole=cluster-admin --user=myname@mycompany.com
+	// https://cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control
+	By("deploying the controller-manager")
+	err = kbc.Make("deploy", "IMG="+kbc.ImageName)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
```
camilamacedo86, line 208:
```suggestion
```

test/e2e/setup.sh
```diff
@@ -56,6 +56,7 @@ function delete_cluster {
 function test_cluster {
   local flags="$@"
```
camilamacedo86, line 58:
```suggestion

  docker pull memcached:1.4.36-alpine
  kind load docker-image --name $KIND_CLUSTER memcached:1.4.36-alpine
  
  docker pull busybox:1.28
  kind load docker-image --name $KIND_CLUSTER busybox:1.28
```

test/e2e/utils/test_context.go
```diff
@@ -274,6 +290,23 @@ func (t *TestContext) LoadImageToKindCluster() error {
 	return err
 }
 
+// LoadImageToKindClusterWithName loads a local docker image with the name informed to the kind cluster
+func (tc TestContext) LoadImageToKindClusterWithName(image string) error {
+	cmd := exec.Command("docker", "pull", image)
+	_, err := tc.Run(cmd)
+	if err != nil {
+		return err
```
camilamacedo86, line 298:
```suggestion
		warnError(err error)
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+	defer func() {
+		out, err := kbc.Kubectl.CommandInNamespace("describe", "all")
+		ExpectWithOffset(1, err).NotTo(HaveOccurred())
+		fmt.Fprintln(GinkgoWriter, out)
+	}()
+	EventuallyWithOffset(1, verifyControllerUp, time.Minute, time.Second).Should(Succeed())
+	By("pulling image")
+	_ = kbc.LoadImageToKindClusterWithName(imageCR)
+
```
camilamacedo86, line 232:
```suggestion
	err = kbc.LoadImageToKindClusterWithName(imageCR)
	Expect(err).To(Not(HaveOccurred()))
```

test/e2e/utils/test_context.go
```diff
@@ -274,6 +290,23 @@ func (t *TestContext) LoadImageToKindCluster() error {
 	return err
 }
 
+// LoadImageToKindClusterWithName loads a local docker image with the name informed to the kind cluster
+func (tc TestContext) LoadImageToKindClusterWithName(image string) error {
+	cmd := exec.Command("docker", "pull", image)
+	_, err := tc.Run(cmd)
+	if err != nil {
+		warnError(err error)
```
camilamacedo86, line 298:
```suggestion
		warnError(err)
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,296 @@
...
+	var podsOutput string
+	getPods := func() error {
+		podsOutput, err = kbc.Kubectl.Get(true, "pods", "-o",
+			"jsonpath={range .items[*]}{.metadata.name},{.status.phase} {end}")
+		if err == nil && strings.TrimSpace(podsOutput) == "" {
+			err = errors.New("empty pod output, continue")
+		}
+
+		return err
+	}
```
camilamacedo86, line 270:
```suggestion
	}
	// TODO: Change the test for a GitHub action so that we can do it faster
```

test/e2e/setup.sh
```diff
@@ -55,7 +55,13 @@ function delete_cluster {
 
 function test_cluster {
   local flags="$@"
-
+  # Pull and load required images for the deploy-image test
+  docker pull memcached:1.4.36-alpine
+  kind load docker-image --name $KIND_CLUSTER memcached:1.4.36-alpine
+  
+  docker pull busybox:1.28
+  kind load docker-image --name $KIND_CLUSTER busybox:1.28
```
camilamacedo86, line 63:
```suggestion
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,297 @@
...
+			// Skip if cluster version < 1.16, when v1 CRDs and webhooks did not exist.
+			// Skip if cluster version < 1.19, because securityContext.seccompProfile only works from 1.19
+			// Otherwise, unknown field "seccompProfile" in io.k8s.api.core.v1.PodSecurityContext will be faced
+			if srvVer := kbc.K8sVersion.ServerVersion; srvVer.GetMajorInt() <= 1 && srvVer.GetMinorInt() < 19 {
+				Skip(fmt.Sprintf("cluster version %s does not support "+
+					"securityContext.seccompProfile", srvVer.GitVersion))
+			}
+
+			var err error
+
```
camilamacedo86, line 122:
```suggestion

```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,297 @@
...
+			By("initializing a project with go/v3")
+			err = kbc.Init(
+				"--plugins", "go/v3",
+				"--project-version", "3",
+				"--domain", kbc.Domain,
+				"--fetch-deps=false",
+			)
+			ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
```
camilamacedo86, line 131:
```suggestion
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,297 @@
...
+			By("uncomment kustomization.yaml to enable prometheus")
+			ExpectWithOffset(1, util.UncommentCode(
+				filepath.Join(kbc.Dir, "config", "default", "kustomization.yaml"),
+				"#- ../prometheus", "#")).To(Succeed())
+
+			By("uncomment kustomize files to ensure that pods are restricted")
+			uncommentPodStandards(kbc)
```
camilamacedo86, line 150:
```suggestion
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+			Run(kbc, "busybox:1.28")
+		})
```
camilamacedo86, line 153:
```suggestion
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,255 @@
...
+	getPods := func() error {
+		podsOutput, err = kbc.Kubectl.Get(true, "pods", "-o",
+			"jsonpath={range .items[*]}{.metadata.name},{.status.phase} {end}")
+		if err == nil && strings.TrimSpace(podsOutput) == "" {
+			err = errors.New("empty pod output, continue")
+		}
+
+		return err
+	}
+	// TODO: Change the test for a GitHub action so that we can do it faster
```
camilamacedo86, line 229:
```suggestion
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,254 @@
...
+	crdViewerRole := filepath.Join("config", "rbac", fmt.Sprintf("%s_viewer_role.yaml", strings.ToLower(kbc.Kind)))
+	EventuallyWithOffset(1, func() error {
+		_, err = kbc.Kubectl.Apply(true, "-f", crdViewerRole)
+		return err
+	}, time.Minute, time.Second).Should(Succeed())
+
+	By("validating that pod(s) status.phase=Running")
+	var podsOutput string
+	getPods := func() error {
+		podsOutput, err = kbc.Kubectl.Get(true, "pods", "-o",
```
camilamacedo86, line 221:
```suggestion
		podsOutput, err = kbc.Kubectl.Get(true, "pods", "-l", "type=memcached", "-o", 
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,254 @@
...
+			)
+			ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+			By("creating API definition with deploy-image/v1-alpha plugin")
+			err = kbc.CreateAPI(
+				"--group", kbc.Group,
+				"--version", kbc.Version,
+				"--kind", kbc.Kind,
+				"--plugins", "deploy-image/v1-alpha",
+				"--image", "memcached:1.4.36-alpine",
```
camilamacedo86, line 93:
```suggestion
				"--image", "nginx:1.14.2",
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,254 @@
...
+				"--image-container-port", "11211",
```
camilamacedo86, line 94:
```suggestion
				"--image-container-port", "80",
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,254 @@
...
+				"--image-container-command", "memcached,-o,modern,-v",
```
camilamacedo86, line 95:
```suggestion
```

test/e2e/setup.sh
```diff
@@ -55,7 +55,11 @@ function delete_cluster {
 
 function test_cluster {
   local flags="$@"
-
+  # Pull and load required images for the deploy-image test
+  docker pull memcached:1.4.36-alpine
+  kind load docker-image --name $KIND_CLUSTER memcached:1.4.36-alpine
```
camilamacedo86, line 60:
```suggestion
  docker pull nginx:1.14.2
  kind load docker-image --name $KIND_CLUSTER nginx:1.14.2
```

test/e2e/setup.sh
```diff
@@ -55,7 +55,11 @@ function delete_cluster {
 
 function test_cluster {
   local flags="$@"
-
+  # Pull and load required images for the deploy-image test
+  docker pull nginx:1.14.2
+  kind load docker-image --name $KIND_CLUSTER nginx:1.14.2
+
```
camilamacedo86, line 61:
```suggestion
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -0,0 +1,302 @@
...
+	}, time.Minute, time.Second).Should(Succeed())
+
+	By("applying the CRD Viewer Role")
+	crdViewerRole := filepath.Join("config", "rbac", fmt.Sprintf("%s_viewer_role.yaml", strings.ToLower(kbc.Kind)))
+	EventuallyWithOffset(1, func() error {
+		_, err = kbc.Kubectl.Apply(true, "-f", crdViewerRole)
+		return err
+	}, time.Minute, time.Second).Should(Succeed())
+
+	By("validating that pod(s) status.phase=Running")
```
camilamacedo86, line 266: @NikhilSharmaWe since it is working locally:

Screenshot 2022-06-27 at 17 59 34

Let's comment it here and we will check in a follow up why does not work on prow

c/c @varshaprasad96

varshaprasad96 (approved):
/lgtm

PR #2723, opened 2022-06-13
:seedling: fix e2e testing for k8s 1.24

everettraven:
# Description
Manually create the ServiceAccount secret token for e2e tests as k8s 1.24 does not generate it automatically.

# Motivation
Fix e2e tests running on k8s 1.24 from failing.


camilamacedo86 (dismissed):

test/e2e/v3/plugin_cluster_test.go
```diff
@@ -214,6 +214,16 @@ func Run(kbc *utils.TestContext) {
...
+	// As of Kubernetes 1.24 a ServiceAccount no longer has a ServiceAccount token
+	// secret autogenerated. We have to create it manually here
+	By("Creating the ServiceAccount token")
+	secretFile, err := createServiceAccountToken(kbc.Kubectl.ServiceAccount, kbc.Dir)
+	Expect(err).NotTo(HaveOccurred())
+	Eventually(func() error {
+		_, err = kbc.Kubectl.Apply(true, "-f", secretFile)
+		return err
+	}, time.Minute, time.Second).Should(Succeed())
+
```
camilamacedo86, line 226: We will need to call that on the tests for we not use it here: https://github.com/kubernetes-sigs/kubebuilder/blob/af78fcc8deee45913809303709b902a81fc0bfbe/test/e2e/v3/plugin_cluster_test.go#L87-L95

Also, we need to do in the createServiceAccountToken and as a follow up we need to check if we should not be scaffolding it.

camilamacedo86 (commented):
/lgtm cancel

test/e2e/v3/plugin_cluster_test.go
```diff
@@ -116,7 +119,13 @@ var _ = Describe("kubebuilder", func() {
 				}
 
 				GenerateV3(kbc, "v1")
-				Run(kbc)
+
+				// only if running on Kubernetes >= 1.24 do we need to generate the ServiceAccount token Secret
```
liggitt, line 123: Auto-generated tokens should not be scraped for use outside pods on any version, they are only for use inside pods.

I'd recommend using TokenRequest to get a token (this works on all currently supported Kubernetes versions) or consistently creating a dedicated token secret, but not scraping auto-generated tokens on some versions

camilamacedo86: @liggitt that is only for the tests it is not used by the end-users and only because we curl the pod to check if we are able to export the metrics in the endpoint.

liggitt: still, I'd avoid scraping auto-generated tokens everywhere... you never know what code will get copied into more important locations

everettraven: The token also does still get used in a pod that runs the `curl` command.

I agree that it may be a good idea to see if there is a better approach to doing this as a follow up (ideally could track discussion in an issue), but I think the scope of this PR is specifically to minimally impact the current way the tests behave while still allowing them to work on K8s 1.24

everettraven: Sure thing. Will update it now.

camilamacedo86: Yes, we can fix the CI now and create an issue to fix that. We have more changes that we need to there anyway.

everettraven: Added TODO comments in commit https://github.com/kubernetes-sigs/kubebuilder/pull/2723/commits/dfa14858054ddcb077205474cbbc6ed42a8376c5

Resolving this, but feel free to unresolve if there are further changes needed. Thanks!

camilamacedo86 (approved):
We will need to address some improvements in the tests
Also, we need in a follow-up to change the code as suggested:

> I'd recommend using TokenRequest to get a token (this works on all currently supported Kubernetes versions) or consistently creating a dedicated token secret, but not scraping auto-generated tokens on some versions

@everettraven could you please help us out by ensuring that we have an issue tracked for that? 

However, just to unblock other PR's contributions I think we are ok to move forward with the alternative solution.

/lgtm

PR #2730, opened 2022-06-15
🌱 change the Kind for just one node (default config)

camilamacedo86:
**Description**
We do not need to create 3 nodes to do the tests so that this PR changes the config for the default one

**Motivation**
We are not doing any tests which require many nodes
So the idea is to keep it lightly as possible


everettraven (approved):
/lgtm

PR #2733, opened 2022-06-16
:seedling: GIT-2731: enable TokenRequest for ServiceAccount token generation

### LINKED ISSUE #2731: testing: improve token usage in e2e tests

The current method of scraping auto-generated tokens for a ServiceAccount is not best practice. This process has also caused the need for changes based on different versions of Kubernetes (i.e. Kubernetes 1.24 no longer auto-generates tokens for ServiceAccounts, which resulted in having to modify test behavior to create a token for this ServiceAccount).

It was pointed out in this [review comment](https://github.com/kubernetes-sigs/kubebuilder/pull/2723#discussion_r898208391) that we should try to utilize the TokenRequest API instead of scraping the generated tokens. The TokenRequest API should work in all versions of Kubernetes, making it easier to bump Kubernetes versions without having to resolve testing issues due to bad habits being used in the test cases.

### Extra Labels

_No response_

harshanarayana:
Closes #2731 

This addresses https://github.com/kubernetes-sigs/kubebuilder/pull/2723#discussion_r898208391

Move to using an explicit Token generated for the SA by invoking the `kubectl create token` command.

### DIFF
--- a/test/e2e/setup.sh
+++ b/test/e2e/setup.sh
@@ -27,10 +27,16 @@ install_kind
 #   export KIND_CLUSTER=<kind cluster name>
 #   create_cluster <k8s version>
 function create_cluster {
+  KIND_VERSION=$1
   : ${KIND_CLUSTER:?"KIND_CLUSTER must be set"}
   : ${1:?"k8s version must be set as arg 1"}
   if ! kind get clusters | grep -q $KIND_CLUSTER ; then
-    kind create cluster -v 4 --name $KIND_CLUSTER --retain --wait=1m --config $(dirname "$0")/kind-config.yaml --image=kindest/node:$1
+    version_prefix="${KIND_VERSION%.*}"
+    kind_config=$(dirname "$0")/kind-config.yaml
+    if test -f $(dirname "$0")/kind-config-${version_prefix}.yaml; then
+      kind_config=$(dirname "$0")/kind-config-${version_prefix}.yaml
+    fi
+    kind create cluster -v 4 --name $KIND_CLUSTER --retain --wait=1m --config ${kind_config} --image=kindest/node:$1
   fi
 }
 
--- a/test/e2e/v3/plugin_cluster_test.go
+++ b/test/e2e/v3/plugin_cluster_test.go
@@ -17,7 +17,7 @@ limitations under the License.
 package v3
 
 import (
-	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -38,11 +38,22 @@ import (
 	"sigs.k8s.io/kubebuilder/v3/test/e2e/utils"
 )
 
+const (
+	tokenRequestRawString = `{"apiVersion": "authentication.k8s.io/v1", "kind": "TokenRequest"}`
+)
+
+// tokenRequest is a trimmed down version of the authentication.k8s.io/v1/TokenRequest Type
+// that we want to use for extracting the token.
+type tokenRequest struct {
+	Status struct {
+		Token string `json:"token"`
+	} `json:"status"`
+}
+
 var _ = Describe("kubebuilder", func() {
 	Context("project version 3", func() {
 		var (
 			kbc *utils.TestContext
-			sat bool
 		)
 
 		BeforeEach(func() {
@@ -53,8 +64,6 @@ var _ = Describe("kubebuilder", func() {
 
 			By("installing the Prometheus operator")
 			Expect(kbc.InstallPrometheusOperManager()).To(Succeed())
-
-			sat = false
 		})
 
 		AfterEach(func() {
@@ -94,7 +103,7 @@ var _ = Describe("kubebuilder", func() {
 				kbc.Kubectl.ServiceAccount = "default"
 				defer func() { kbc.Kubectl.ServiceAccount = tmp }()
 				GenerateV2(kbc)
-				Run(kbc, sat)
+				Run(kbc)
 			})
 		})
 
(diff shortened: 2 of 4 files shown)


test/e2e/v3/plugin_cluster_test.go
```diff
@@ -398,43 +366,20 @@ func curlMetrics(kbc *utils.TestContext) string {
...
-	secretFile, err := createServiceAccountToken(kbc.Kubectl.ServiceAccount, kbc.Dir)
-	Expect(err).NotTo(HaveOccurred())
+	var out string
+	var err error
 	Eventually(func() error {
-		_, err = kbc.Kubectl.Apply(true, "-f", secretFile)
+		out, err = kbc.Kubectl.Command(
+			"create", "token", kbc.Kubectl.ServiceAccount,
+			"--audience",
+			fmt.Sprintf("https://e2e-%s-controller-manager-metrics-service.%s.svc:8443", kbc.TestSuffix, kbc.Kubectl.Namespace))
```
harshanarayana, line 381: @everettraven  @camilamacedo86 Do you want these items to be more specific or more free in terms of what is in the Token ?

test/e2e/v3/plugin_cluster_test.go
```diff
@@ -17,7 +17,6 @@ limitations under the License.
 package v3
```
camilamacedo86, line 17: We use the emojis to create the changelog.
This change is only in the tests and does not affect the end-users
So that is not a bug we need change it for 🌱 , see the contribution guide.

test/e2e/v3/plugin_cluster_test.go
camilamacedo86: Hi @harshanarayana I unable the tests with this change they are not working
It is fine do the change here directly
However, we test against 1.18 so if you change it for a syntax that does not work on 1.18 we need to have the logic implemented to do that only when is > than x. ( as it is now only for example scaffolding the token when the test is done against 1.24 only )

Btw, thx for your help on this one. 🥇

harshanarayana: Thanks @camilamacedo86  Let me check and update it. I am yet to get this tested locally as well.

harshanarayana: I've addressed the required changes. For now keeping the usage of `kubectl create token` for `v1.24+` and using the template to render a new secret that will provide the token required to perform the `curl` operation for Metrics fetch. 

> Auto-generated tokens should not be scraped for use outside pods on any version

A bit of the code has been refactored to address what @liggitt suggested.  So, all versions now use a secret of type `kubernetes.io/service-account-token` to get the token required to do the operations. PTAL. Hope this addresses the request

test/e2e/v3/plugin_cluster_test.go
```diff
@@ -398,43 +380,49 @@ func curlMetrics(kbc *utils.TestContext) string {
...
+// ServiceAccountToken performs a kubectl create token to request a new Token
+// for the service account in question and use that token for performing the
+// curl command calls as required. Please keep in mind that the output of
+// kubectl create token is not base64 encoded. It is the JWT token itself
+func ServiceAccountToken(kbc *utils.TestContext) (out string, err error) {
 	By("Creating the ServiceAccount token")
-	secretFile, err := createServiceAccountToken(kbc.Kubectl.ServiceAccount, kbc.Dir)
-	Expect(err).NotTo(HaveOccurred())
-	Eventually(func() error {
+	if srvVer := kbc.K8sVersion.ServerVersion; srvVer.GetMajorInt() == 1 && srvVer.GetMinorInt() >= 24 {
```
everettraven, line 389: Is there a reason the TokenRequest only works on k8s 1.24+?

I was under the impression that this should work on all the Kubernetes versions we are using because the feature was in beta status as of k8s 1.12 which is earlier than any of the versions that we are testing. I think it could potentially be related to it being disabled within the KinD node we are using but I would be curious what the earliest version we could start using the TokenRequest API in is.

I would be interested to see what happens when slowly backing the minor version down from 24 to as low as we can get it.

harshanarayana: The reason I put this check in is using the `kubectl create token` command. That is present only in 1.24.x+.

I am trying out a change to use the `TokenRequest` API call for all of them and avoid this check. Will push the changes in a while once locally tested.

harshanarayana:
## NOTE

had to make a hacky  change to the setup.sh to make it switch between the right kind config file based on versions. kind config for 1.18 and 1.19 are the same file. `kind-config-v1.19.yaml` is just a symlink to `kind-config-v1.18.yaml`

These two files introduce new Argument to the API Server to make sure the `TokenRequest` will work.

test/e2e/v3/plugin_cluster_test.go
harshanarayana: Marking this as resolved. PTAL at the changes that has addressed your suggested changes. @camilamacedo86 . Thanks

test/e2e/setup.sh
```diff
@@ -27,10 +27,16 @@ install_kind
...
   : ${KIND_CLUSTER:?"KIND_CLUSTER must be set"}
   : ${1:?"k8s version must be set as arg 1"}
   if ! kind get clusters | grep -q $KIND_CLUSTER ; then
-    kind create cluster -v 4 --name $KIND_CLUSTER --retain --wait=1m --config $(dirname "$0")/kind-config.yaml --image=kindest/node:$1
+    version_prefix="${KIND_VERSION%.*}"
+    kind_config=$(dirname "$0")/kind-config.yaml
+    if test -f $(dirname "$0")/kind-config-${version_prefix}.yaml; then
+      kind_config=$(dirname "$0")/kind-config-${version_prefix}.yaml
+    fi
+    kind create cluster -v 4 --name $KIND_CLUSTER --retain --wait=1m --config ${kind_config} --image=kindest/node:$1
```
camilamacedo86, line 39: Why did you make this changes?

camilamacedo86 (approved):
That shows great 🥇

everettraven (approved):
Looks great! Thanks so much for the contribution @harshanarayana!

PR #2741, opened 2022-06-16
:seedling: clean up samples by removing deprecated options

camilamacedo86:
**Description**

- remove samples with add-on and declarative for project-v2
- remove samples for v1beta1 CRDs and WebHooks 

**Motivation**
- We are adding new plugins and options so that we will have more samples
- The project-v2 samples uses go/v2 which can no longer be updated and exist only for backwards compatibility with the legacy layout. Projects with it cannot work on Kubernetes 1.22+ so that seems we can clean up here since we also do not encourage people to use them.  
- Also, we do not encourage users still using v1beta1 CRDs and this option is deprecated


test/testdata/generate.sh
```diff
@@ -123,12 +119,9 @@ build_kb
 
 # Project version 2 uses plugin go/v2 (default).
 scaffold_test_project project-v2 --project-version=2
```
everettraven, line 121: Should we remove this one as well?

test/testdata/generate.sh
```diff
@@ -123,12 +119,9 @@ build_kb
...
-scaffold_test_project project-v2-multigroup --project-version=2
-scaffold_test_project project-v2-addon --project-version=3 --plugins="go/v2,declarative"
 # Project version 3 (default) uses plugin go/v3 (default).
 scaffold_test_project project-v3
 scaffold_test_project project-v3-multigroup
 scaffold_test_project project-v3-addon --plugins="go/v3,declarative"
 scaffold_test_project project-v3-config --component-config
-scaffold_test_project project-v3-v1beta1
```
everettraven, line 133: I agree with removing the v2 projects since they are being deprecated, but are we also deprecating the v1beta1 support for the go/v3 plugin? If not then I think we should keep this one in.

test/testdata/test_legacy.sh
```diff
@@ -34,5 +34,4 @@ build_kb
 tools_k8s_version="1.19.2"
 fetch_tools
 test_project project-v2
```
everettraven, line 36: Should we remove this one as well?

test/testdata/generate.sh
rashmigottipati: I agree with @everettraven.

test/testdata/generate.sh
camilamacedo86: I do not think so.
Why? we need to ensure in the reviews that changes are not made in this scaffold. 
By keeping it, we can easily check that.

test/testdata/generate.sh
camilamacedo86: Yes, we do not want anybody using v1neta1. It does not work since K8s >= 1.22 and the flag is deprecated across all places

test/testdata/test_legacy.sh
camilamacedo86: Nops, since we will keep the sample

test/testdata/generate.sh
everettraven: Makes sense

test/testdata/generate.sh
everettraven: Gotcha, then this is good.

everettraven (approved):
LGTM

rashmigottipati (approved):
/lgtm

PR #2749, opened 2022-06-18
🌱 : ensure that testdata samples will be re-generated from the scratch

camilamacedo86:
## Description
ensure that testdata samples will be re-generated from the scratch

## Motivation
Ensure that we are able to properly verify the changes on the review
Do not keep files that are no longer generated


camilamacedo86:
Moving forward in order to not block contributions.

PR #2750, opened 2022-06-18
:sparkles: (kustomize/v2-alpha) fixes in the scaffold to follow the changes in the MAJOR release of kustomize

camilamacedo86:
## Description

- Replace every instance of $(SOMETHING) with any arbitrary placeholder value.
- Replace: `bases:` with `resources:`
- Replace:

```#commonLabels:
#  someName: someValue
```

With:

```
#labels:
#- includeSelectors: true
#  pairs:
#    someName: someValue
```

### DIFF
--- a/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/certmanager/certificate.go
+++ b/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/certmanager/certificate.go
@@ -60,10 +60,10 @@ metadata:
   name: serving-cert  # this name should match the one appeared in kustomizeconfig.yaml
   namespace: system
 spec:
-  # $(SERVICE_NAME) and $(SERVICE_NAMESPACE) will be substituted by kustomize
+  # SERVICE_NAME and SERVICE_NAMESPACE will be substituted by kustomize
   dnsNames:
-  - $(SERVICE_NAME).$(SERVICE_NAMESPACE).svc
-  - $(SERVICE_NAME).$(SERVICE_NAMESPACE).svc.cluster.local
+  - SERVICE_NAME.SERVICE_NAMESPACE.svc
+  - SERVICE_NAME.SERVICE_NAMESPACE.svc.cluster.local
   issuerRef:
     kind: Issuer
     name: selfsigned-issuer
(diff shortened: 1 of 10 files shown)


pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/certmanager/certificate.go
```diff
@@ -60,10 +60,10 @@ metadata:
   name: serving-cert  # this name should match the one appeared in kustomizeconfig.yaml
   namespace: system
 spec:
-  # $(SERVICE_NAME) and $(SERVICE_NAMESPACE) will be substituted by kustomize
+  # SERVICE_NAME_PLACEHOLDER and SERVICE_NAMESPACE_PLACEHOLDER will be substituted by kustomize
```
erikgb, line 63: IMO the suffix `_PLACEHOLDER` is not needed. The all-caps style is easily recognized as var-syntax style IMO.

But I wonder if this change really work? Have you tested the resulting scaffolding @camilamacedo86? If you are trying to migrate from using vars to replacements, I think there is a bit more to be done, ref. https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/vars/#more-complex-migration-example

camilamacedo86: Hi @erikgb this changes are for the new alpha plugin.
The changes from vars to replacements are done here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go#L92-L190
In this follow-up pr, I am just trying to adjust the rest of the places. 

Why did I change the var by the placeholders? The doc says:

`Replace every instance of $(SOME_SECRET_NAME) with any arbitrary placeholder value.`

However, I am still trying to figure out if we have not other places as well that we need to change.

erikgb: Ok, I understand that the migration to replacements has started. Good! But I would just stick with the existing names `SERVICE_NAME` \ `SERVICE_NAMESPACE` and drop the `_PLACEHOLDER` postfix. It still looks like a placeholder IMO.

camilamacedo86: If we run kustomize edit fix --vars to help you move forward with the new changes for 4.x they suggest PLACEHOLDER word to make clear for users that it is just a PLACEHOLDER so because of that I kept as suggested.

everettraven (approved):
Looks good to me!

camilamacedo86:
Moving forward since the @erikgb suggestions are addressed and we have also another LGTM.

PR #2753, opened 2022-06-18
:seedling: ensure that project-v3 and project-v3-with-kustomize does the same scaffold

camilamacedo86:
## Description

Update the samples to ensure that project-v3 and project-v3-with-kustomize does the same scaffolds so that we can easily compare the differences between the projects.


testdata/project-v3-with-kustomize-v2/PROJECT
```diff
@@ -37,6 +37,7 @@ resources:
   group: crew
   kind: Admiral
   path: sigs.k8s.io/kubebuilder/testdata/project-v3-with-kustomize-v2/api/v1
+  plural: admirales
```
erikgb, line 40:
```suggestion
  plural: admirals
```

I would standardize for the correct English term here, so `admirals` instead of `admirales`. Or is there a reason for doing the opposite when streamlining this?

camilamacedo86: We have an option command to change the plurals which are required in some very specific cases
The goal of this PR is we only have the samples:
- project-v3
AND
- project-v3-with-kustomize-v2-alpha plugin

Doing exactly the same scaffold. So that we can:
a) Better evaluate the need or not to have the kustomize-v2 alpha plugin
b) If still required in a follow-up (after the next release) we will replace it with the go/v4-alpha plugin
v) So that people can easily compare and check what changed from go/v3 to go/v4-alpha so far.

camilamacedo86: @erikgb

erikgb: I understand the need to streamline the scaffold, but what I don't understand is why you cannot standardize on `admirals` instead of `admirales`.

camilamacedo86: That is a sample, we scaffold the project v3 using the option to ensure that we will not break the feature. 
The goal of this PR is only to ensure that project v3 sample is == project v3 with new kustomize.
We are not change nothing in the scaffold that is only "I want the sample testdata generate with the new plugin done using the same options used to build the project v3 scaffold so that we can diff and check what was changed from one to another"

If you would like to know why we have a flag `--plural:` that allows we change the plurals, see : https://kubernetes.slack.com/archives/CAR30FCJZ/p1604861145449700

c/c @erikgb

testdata/project-v3-with-kustomize-v2/config/crd/bases/crew.testproject.org_admirales.yaml
```diff
@@ -5,20 +5,20 @@ metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.9.0
   creationTimestamp: null
-  name: admirals.crew.testproject.org
+  name: admirales.crew.testproject.org
```
FillZpp, line 8: Doing this in sample is fine. But still I have to say the plural change may lead to a high risk in the real world.

For example, the old CRD `admirals.crew.testproject.org` was already applied into K8s cluster. Then we generate a new `admirales.crew.testproject.org` CRD, how should we update it?
1. If we just apply it into cluster, there will be two CRDs with different name and plural, but the same group, version and kind.
2. If we remove the old CRD and then apply the new one, all existing CRs of this CRD will be removed.

So, I'm just afraid will this be a bad sample for users to manage their CRDs? Or we have to warn here that ppl should not change the plural unless they know what they are doing.

camilamacedo86: The purpose of the PR is only to ensure that in both tests we are doing exactly the same check

Why do we scaffold it in the testdata sample?

Because we need to ensure that changes in the code will not break the feature. Only because of that. 
The samples under testdata are used only to check the layouts and they have no implementation or are used as examples.

FillZpp: Emm.. I'm ok with this since they are used only to check the layouts and have no implementation or are used as examples.

LGTM, let's see if @erikgb has any ideas. (and it has conflicts should be resolved)

erikgb: I'm ok with this for now, after the clarifications from @camilamacedo86.

camilamacedo86: Thank you folks.

PR #2754, opened 2022-06-18
:book: update info about new kustomize plugin alpha

camilamacedo86:
##Description
:book: update info about new kustomize plugin alpha

##Motivation
Provide info about how to convert vars to replacements


camilamacedo86:
Just a nit in the doc, moving forward to cause less effort on the reviews.

PR #2755, opened 2022-06-18
:sparkles: (kustomize/v2-alpha) bump kustomize version from v4.5.3 to v4.5.5

camilamacedo86:
## Description
:sparkles: (kustomize/v2-alpha) bump kustomize version from v4.5.3 to v4.5.5

## Motivation
We provide it with the latest one available


erikgb:
LGTM, but there is a failing test. Maybe there is another place that needs to be updated as well?

camilamacedo86:
@erikgb Thank you

> LGTM, but there is a failing test. Maybe there is another place that needs to be updated as well?

The goal here is to only bump the alpha one. 
Moving forward with this one.

PR #2756, opened 2022-06-18
📖 Fix broken kustomize install link

oscr:
Hi, 

I noticed that the kustomize install link in the contributing doc was broken, so this commit replaces it with a working link.


camilamacedo86 (approved):
/lgtm

PR #2757, opened 2022-06-19
:book: quick start updates

camilamacedo86:
## Description

Remove the suggestion to call kustomize/v2-alpha,base.go.kubebuilder.io/v3 directly.

## Motivation

We probably will create a bundle go/v4-alpha and we do not want to support afterwards the workaround 
or provide migration steps from it.


camilamacedo86:
moving forward with this one because it is a nit and we need help/effort to review the prs that are most important/critical ones.

PR #2787, opened 2022-07-01
:bug: (API) remove KustomizeVersion const from golang plugin ( the version is defined in the kustomize plugins which is responsable for )

camilamacedo86:
## Description

The plugin responsable for define the Kustomize Version is the Kustomize one. 

If you have consumer Kubebuilder as a lib ensure that you properly get the version from the Kustomize Plugin instead of Golang base language. Note that to create helpers and/or plugin versions that are able to deal with more than on version you can, example:

```go
        kustomizeVersion = kustomizecommonv1.KustomizeVersion
	kustomizev2 := kustomizecommonv2alpha.Plugin{}
	pluginKeyForKustomizeV2 := plugin.KeyFor(kustomizev2)

	for _, pluginKey := range s.config.GetPluginChain() {
		if pluginKey == pluginKeyForKustomizeV2 {
			kustomizeVersion = kustomizecommonv2alpha.KustomizeVersion
			break
		}
	}

        ....

        &templates.Makefile{
	      ....
	      KustomizeVersion:         kustomizeVersion,
	      ....
      },

```

### DIFF
--- a/pkg/plugins/golang/v3/scaffolds/init.go
+++ b/pkg/plugins/golang/v3/scaffolds/init.go
@@ -37,16 +37,6 @@ const (
 	ControllerRuntimeVersion = "v0.12.1"
 	// ControllerToolsVersion is the kubernetes-sigs/controller-tools version to be used in the project
 	ControllerToolsVersion = "v0.9.0"
-	// KustomizeVersion is the kubernetes-sigs/kustomize version to be used in the project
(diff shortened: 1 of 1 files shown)


camilamacedo86:
Note in this case is expected the PR fails in the `APIDiff / Verify API differences (pull_request)`.
We will skip this one to move forward. This check is done only to help us not introduce breaking changes.

everettraven (approved):
LGTM

Issue #2765, opened 2022-06-22
Meta - New deploy-image/v1-alpha plugin follow ups 


camilamacedo86:
### What do you want to happen?

Following some tracks, we are able to check and address them as a follow-up of https://github.com/kubernetes-sigs/kubebuilder/pull/2717

- [x] - Add valid status conditions based on the fact that the deployment was successful. ([doc](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)) and [example to update the status in the controller](https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v3/memcached-operator/controllers/memcached_controller.go#L112-L135) - Done in  #2816
- [x] Add markers validations, i.e: 
For the size:

```go
        // The following markers will use OpenAPI v3 schema to validate the value
	// More info: https://book.kubebuilder.io/reference/markers/crd-validation.html
	// +kubebuilder:validation:Minimum=1
	// +kubebuilder:validation:Maximum=3
	// +kubebuilder:validation:ExclusiveMaximum=false

```

Also, would be great if we can add a check for the Port when it is informed. 

- [x] - Improve controller-test to check if the deployment/pod was made 
- [x] - Be able to do checks with images in the e2e and get the PodSatus ( that might mean moving the e2e checks to GitHub action instead of using the prow ) - PR: https://github.com/kubernetes-sigs/kubebuilder/pull/2795
- [x] - For e2e: Ensure that the deployment scaffold on the controller to deploy the Operand is checked 
- [x] - Project file has to track the options used under the plugin and not under the resources. Should we not track under the resources? What happens if we create 2 apis with the plugin? Will we have the image, command and port used for both calls tracked? 
- [x] - [RunAsUser](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go#L146) needs to be a flag arg ( we should not add it by default  )
- [x] - Scaffold controller tests see [here](https://book.kubebuilder.io/cronjob-tutorial/writing-tests.html). We will test if the deployment can be done successfully. (Done : https://github.com/kubernetes-sigs/kubebuilder/pull/2792)
- [x] - Check the project with --namespaced=false flag to create apis. Erro faced when it was used: 
[...]

PR #2792, opened 2022-07-03
✨deploy-image/v1-alpha: add tests for controller using ENV TEST

NikhilSharmaWe:
### Description
Add integration tests for the controller scaffolded by the `deploy-image/v1-alpha` plugin.

### Motivation

### DIFF
--- a/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
+++ b/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
@@ -111,6 +111,12 @@ func (s *apiScaffolder) Scaffold() error {
 		return fmt.Errorf("error updating config/samples: %v", err)
 	}
 
+	if err := scaffold.Execute(
+		&controllers.ControllerTest{},
+	); err != nil {
+		return fmt.Errorf("error creating controllers/**_controller_test.go: %v", err)
+	}
+
 	return nil
 }
 
@@ -140,9 +146,13 @@ func (s *apiScaffolder) scafffoldControllerWithImage(scaffold *machinery.Scaffol
 		// users can change the values
 		var res string
 		for _, value := range strings.Split(s.command, ",") {
-			res += fmt.Sprintf("\"%s\",", strings.TrimSpace(value))
+			res += fmt.Sprintf(" \"%s\",", strings.TrimSpace(value))
 		}
+		// remove the latest ,
 		res = res[:len(res)-1]
+		// remove the first space to not fail in the go fmt ./...
+		res = strings.TrimLeft(res, " ")
+
 		err := util.InsertCode(controllerPath, `SecurityContext: &corev1.SecurityContext{
 							RunAsNonRoot:             &[]bool{true}[0],
 							AllowPrivilegeEscalation: &[]bool{false}[0],
@@ -208,8 +218,8 @@ func (s *apiScaffolder) scaffoldCreateAPIFromGolang() error {
 }
 
 const containerTemplate = `Containers: []corev1.Container{{
-						Image: "%s",
-						Name:  "%s",
+						Image:           "%s",
+						Name:            "%s",
 						ImagePullPolicy: corev1.PullIfNotPresent,
 						// Ensure restrictive context for the container
 						// More info: https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted
@@ -225,10 +235,10 @@ const containerTemplate = `Containers: []corev1.Container{{
 					}}`
 
 const runAsUserTemplate = `
-							RunAsUser: &[]int64{%s}[0],`
+							RunAsUser:                &[]int64{%s}[0],`
 
 const commandTemplate = `
-						Command:         []string{%s},`
+						Command: []string{%s},`
 
 const portTemplate = `
 						Ports: []corev1.ContainerPort{{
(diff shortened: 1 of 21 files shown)


pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller-test.go
```diff
@@ -0,0 +1,128 @@
...
+
+// ControllerTest scaffolds the file that defines tests for the controller for a CRD or a builtin resource
+// nolint:maligned
+type ControllerTest struct {
+	machinery.TemplateMixin
+	machinery.MultiGroupMixin
+	machinery.BoilerplateMixin
+	machinery.ResourceMixin
+
+	Image string
```
camilamacedo86, line 36: @NikhilSharmaWe nit we do not use the Image so we can remove from here and does not pass to the template

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
```diff
@@ -111,6 +111,12 @@ func (s *apiScaffolder) Scaffold() error {
 		return fmt.Errorf("error updating config/samples: %v", err)
 	}
 
+	if err := scaffold.Execute(
+		&controllers.ControllerTest{Image: s.image},
```
camilamacedo86, line 115:
```suggestion
		&controllers.ControllerTest{},
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
```diff
@@ -111,6 +111,12 @@ func (s *apiScaffolder) Scaffold() error {
...
+	); err != nil {
+		return fmt.Errorf("error updating controllers/**_controller_test.go: %v", err)
```
camilamacedo86, line 117:
```suggestion
		return fmt.Errorf("error creating controllers/**_controller_test.go: %v", err)
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
```diff
@@ -111,6 +111,12 @@ func (s *apiScaffolder) Scaffold() error {
 		return fmt.Errorf("error updating config/samples: %v", err)
 	}
```
camilamacedo86, line 112: @NikhilSharmaWe see that is failing because now it will try to use/install the kubebuilder tools to run the make test on the e2e. We need to change the e2e test:

Replace make docker-build IMG=<image> with 
` docker build -t <image value> . `

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -189,7 +189,7 @@ func Run(kbc *utils.TestContext, imageCR string) {
 	ExpectWithOffset(1, err).NotTo(HaveOccurred())
 
 	By("building the controller image")
-	err = kbc.Make("docker-build", "IMG="+kbc.ImageName)
+	err = kbc.Make("docker-build", "-t", "IMG="+kbc.ImageName)
```
NikhilSharmaWe, line 192: @camilamacedo86 Added -t argument in docker build.

camilamacedo86: No, we are calling the targer `make docker-build`, see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/Makefile#L72-L74

The target calls another makefile target test and the test will need the bin which will not work on prow. Why? it will try to install things and that is not allowed in the same way.

The goal here is only to build the image.
So, that to solve the issue the change is instead of we call the make target test we just execute the docker command: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/Makefile#L74


We n

camilamacedo86, line 192:
```suggestion
	cmd := exec.Command("docker", "build","-t","IMG="+kbc.ImageName)
	_, err = kbc.Run(cmd)
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -189,15 +189,16 @@ func Run(kbc *utils.TestContext, imageCR string) {
 	ExpectWithOffset(1, err).NotTo(HaveOccurred())
 
 	By("building the controller image")
-	err = kbc.Make("docker-build", "IMG="+kbc.ImageName)
+	cmd := exec.Command("docker", "build", "-t", "IMG"+kbc.ImageName)
```
camilamacedo86, line 192:
```suggestion
	cmd := exec.Command("docker", "build", "-t", "IMG="+kbc.ImageName)
```

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -189,15 +189,16 @@ func Run(kbc *utils.TestContext, imageCR string) {
 	ExpectWithOffset(1, err).NotTo(HaveOccurred())
 
 	By("building the controller image")
-	err = kbc.Make("docker-build", "IMG="+kbc.ImageName)
+	cmd := exec.Command("docker", "build", "-t", "IMG="+kbc.ImageName)
```
camilamacedo86, line 192:
```suggestion
	cmd := exec.Command("docker", "build", "-t", "IMG="+kbc.ImageName, ".")
```

camilamacedo86: @NikhilSharmaWe missing the Dockerfile info

camilamacedo86 (approved):
@NikhilSharmaWe,

Great work 🥇 

As a follow up I will raise an issue for the controller runtime using the examples here so we know why the request.Namespace is empty which does not allow us to create the Deployment and check it on the tests.

Before we had the fake client do the tests, which is still valid but would be nice if we are able to do that using only ENV TEST since has a desire to deprecate the fake client. See an example:  https://github.com/dev4devs-com/postgresql-operator/blob/master/pkg/controller/database/controller_test.go

/lgtm
/approved

PR #2793, opened 2022-07-03
✨deploy-image/v1-alpha: add finalizers for the Custom Resource created

NikhilSharmaWe:
### Description
Add finalizers for the Custom Resource created for deleting the deployment under before deleting the CR.

### Motivation

### DIFF
--- a/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
+++ b/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
@@ -105,6 +105,10 @@ func (s *apiScaffolder) Scaffold() error {
 		return fmt.Errorf("error updating controller: %v", err)
 	}
 
+	if err := s.updateMainEventRecorder(); err != nil {
+		return fmt.Errorf("error updating main.go: %v", err)
+	}
+
 	if err := scaffold.Execute(
 		&samples.CRDSample{Port: s.port},
 	); err != nil {
@@ -117,6 +121,18 @@ func (s *apiScaffolder) Scaffold() error {
 		return fmt.Errorf("error creating controllers/**_controller_test.go: %v", err)
 	}
 
+	return nil
+}
+
+// TODO: replace this implementation by creating its own MainUpdater
+// which will have its own controller template which set the recorder
+func (s *apiScaffolder) updateMainEventRecorder() error {
+	defaultMainPath := "main.go"
+	err := util.InsertCode(defaultMainPath, `Scheme: mgr.GetScheme(),`,
+		fmt.Sprintf(recorderTemplate, strings.ToLower(s.resource.Kind)))
+	if err != nil {
+		return fmt.Errorf("error scaffolding event recorder while creating the controller in main.go: %v", err)
+	}
 	return nil
 }
 
@@ -245,3 +261,7 @@ const portTemplate = `
 							ContainerPort: %s.Spec.ContainerPort,
 							Name:          "%s",
 						}},`
+
+const recorderTemplate = `
+		Recorder: mgr.GetEventRecorderFor("%s-controller"),
+	`
--- a/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
+++ b/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
@@ -66,27 +66,33 @@ package {{ if and .MultiGroup .Resource.Group }}{{ .Resource.PackageName }}{{ el
 import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
 	"context"
 	"time"
+	"fmt"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	
 	{{ if not (isEmptyStr .Resource.Path) -}}
 	{{ .Resource.ImportAlias }} "{{ .Resource.Path }}"
 	{{- end }}
 )
 
+const {{ lower .Resource.Kind }}Finalizer = "{{ .Resource.Group }}.{{ .Resource.Domain }}/finalizer"
+
 // {{ .Resource.Kind }}Reconciler reconciles a {{ .Resource.Kind }} object
 type {{ .Resource.Kind }}Reconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
+	Recorder record.EventRecorder
 }
 // The following markers are used to generate the rules permissions on config/rbac using controller-gen
 // when the command <make manifests> is executed. 
@@ -95,6 +101,7 @@ type {{ .Resource.Kind }}Reconciler struct {
 //+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }},verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }}/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }}/finalizers,verbs=update
+//+kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
 //+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
 
@@ -119,7 +126,7 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
 	{{ lower .Resource.Kind }} := &{{ .Resource.ImportAlias }}.{{ .Resource.Kind }}{}
 	err := r.Get(ctx, req.NamespacedName, {{ lower .Resource.Kind }})
 	if err != nil {
-		if errors.IsNotFound(err) {
+		if apierrors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
@@ -131,10 +138,52 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
 		return ctrl.Result{}, err
 	}
 
+	// Let's add a finalizer. Then, we can define some operations which should
+	// occurs before the custom resource to be deleted.
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/
+	// NOTE: You should not use finalizer to delete the resources that are
+	// created in this reconciliation and have the ownerRef set by ctrl.SetControllerReference
+	// because these will get deleted via k8s api
+	if !controllerutil.ContainsFinalizer({{ lower .Resource.Kind }}, {{ lower .Resource.Kind }}Finalizer) {
+		log.Info("Adding Finalizer for {{ .Resource.Kind }}")
+		controllerutil.AddFinalizer({{ lower .Resource.Kind }}, {{ lower .Resource.Kind }}Finalizer)
+		err = r.Update(ctx, {{ lower .Resource.Kind }})
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	// Check if the {{ .Resource.Kind }} instance is marked to be deleted, which is
+	// indicated by the deletion timestamp being set.
+	is{{ .Resource.Kind }}MarkedToBeDeleted := {{ lower .Resource.Kind }}.GetDeletionTimestamp() != nil
+	if is{{ .Resource.Kind }}MarkedToBeDeleted {
+		if controllerutil.ContainsFinalizer({{ lower .Resource.Kind }}, {{ lower .Resource.Kind }}Finalizer) {
+			// Run finalization logic for memcachedFinalizer. If the
+			// finalization logic fails, don't remove the finalizer so
+			// that we can retry during the next reconciliation.
+			log.Info("Performing Finalizer Operations for {{ .Resource.Kind }} before delete CR")
+			r.doFinalizerOperationsFor{{ .Resource.Kind }}({{ lower .Resource.Kind }})
+
+			// Remove memcachedFinalizer. Once all finalizers have been
+			// removed, the object will be deleted.
+			if ok:= controllerutil.RemoveFinalizer({{ lower .Resource.Kind }}, {{ lower .Resource.Kind }}Finalizer); !ok{
+				if err != nil {
+					log.Error(err, "Failed to remove finalizer for {{ .Resource.Kind }}")
+					return ctrl.Result{}, err
+				}
+			}
+			err := r.Update(ctx, {{ lower .Resource.Kind }})
+			if err != nil {
+				log.Error(err, "Failed to remove finalizer for {{ .Resource.Kind }}")
+			}
+		}
+		return ctrl.Result{}, nil
+	}
+
 	// Check if the deployment already exists, if not create a new one
 	found := &appsv1.Deployment{}
 	err = r.Get(ctx, types.NamespacedName{Name: {{ lower .Resource.Kind }}.Name, Namespace: {{ lower .Resource.Kind }}.Namespace}, found)
-	if err != nil && errors.IsNotFound(err) {
+	if err != nil && apierrors.IsNotFound(err) {
 		// Define a new deployment
 		dep := r.deploymentFor{{ .Resource.Kind }}({{ lower .Resource.Kind }})
 		log.Info("Creating a new Deployment", "Deployment.Namespace", dep.Namespace, "Deployment.Name", dep.Name)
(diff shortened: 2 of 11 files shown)


pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -168,6 +172,36 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
 		// the desired state on the cluster
 		return ctrl.Result{Requeue: true}, nil
 	}
+
+	// Check if the {{ .Resource.Kind }} instance is marked to be deleted, which is
```
camilamacedo86, line 176: We need to fix the nits about space and go fmt

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -220,6 +254,19 @@ func labelsFor{{ .Resource.Kind }}(name string) map[string]string {
 	return map[string]string{"type": "{{ lower .Resource.Kind }}", "{{ lower .Resource.Kind }}_cr": name}
 }
 
+func (r *{{ .Resource.Kind }}Reconciler) finalize{{ .Resource.Kind }}Finalizer(ctx context.Context, log logr.Logger, m *{{ .Resource.ImportAlias }}.{{ .Resource.Kind }}) error {
```
camilamacedo86, line 257: teh deployment will be deleted we need to do an extra operation like raise en event

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -77,12 +78,15 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	
 	{{ if not (isEmptyStr .Resource.Path) -}}
 	{{ .Resource.ImportAlias }} "{{ .Resource.Path }}"
 	{{- end }}
 )
 
+const {{ lower .Resource.Kind }}Finalizer = "{{ .Resource.Group }}.{{ .Resource.Domain }}/finalizer"
+
```
camilamacedo86, line 89: We will need also add maker to give the permissions for the finalizer status

testdata/project-v3-with-deploy-image/controllers/busybox_controller.go
```diff
@@ -187,6 +221,19 @@ func labelsForBusybox(name string) map[string]string {
 	return map[string]string{"type": "busybox", "busybox_cr": name}
 }
 
+func (r *BusyboxReconciler) finalizeBusyboxFinalizer(ctx context.Context, log logr.Logger, m *examplecomv1alpha1.Busybox) error {
```
camilamacedo86, line 224: After we replace this code to create an event before remove the CR we also will need to add in the e2e test a check to ensure that the event was raised

testdata/project-v3-with-deploy-image/controllers/busybox_controller.go
```diff
@@ -120,6 +127,35 @@ func (r *BusyboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	// Check if the Busybox instance is marked to be deleted, which is
+	// indicated by the deletion timestamp being set.
+	isBusyboxMarkedToBeDeleted := busybox.GetDeletionTimestamp() != nil
+	if isBusyboxMarkedToBeDeleted {
+		if controllerutil.ContainsFinalizer(busybox, busyboxFinalizer) {
+			// Run finalization logic for busyboxFinalizer. If the
+			// finalization logic fails, don't remove the finalizer so
+			// that we can retry during the next reconciliation.
+			r.recorder.Event(busybox, "Normal", "Deleting", fmt.Sprintf("Custom Resource Definition %s/%s is being deleted", busybox.Namespace, busybox.Name))
+			r.recorder.Event(found, "Normal", "Deleting", fmt.Sprintf("Deployment %s/%s is being deleted", busybox.Namespace, busybox.Name))
```
NikhilSharmaWe, line 140: @camilamacedo86 Could you please take a look at these changes for creating an event.
What are we expecting this finalizer to do? Do we expect to log the information we record in the event when we delete the crd?

But by these changes when I delete the crd created the implementation of kubectl delete crd ** stops at just: `customresourcedefinition.apiextensions.k8s.io "memcacheds.example.com.my.domain" deleted`

Could you please explain what is wrong with the approach.

camilamacedo86: We need to raise the event and check if it was done without errors. 
If err != nil then we can update the finalizer to unlock it and let the CR be deleted.

testdata/project-v3-with-deploy-image/config/rbac/role.yaml
```diff
@@ -51,6 +51,13 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
```
camilamacedo86, line 54: @NikhilSharmaWe 

I added the review in : https://github.com/NikhilSharmaWe/kubebuilder/pull/41/files

See that the RBAC needs to be fixed also is better if we use Warning in the example so that we can easily check in the tests as well.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -95,6 +102,7 @@ type {{ .Resource.Kind }}Reconciler struct {
 //+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }},verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }}/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }}/finalizers,verbs=update
+//+kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources=events,verbs=create;patch
```
camilamacedo86, line 105:
```suggestion
//+kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
```
to fix the RBAC

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -264,6 +264,23 @@ func Run(kbc *utils.TestContext, imageCR string) {
...
 	}, time.Minute, time.Second).Should(Succeed())
 
+	//Testing the finalizer
+	EventuallyWithOffset(1, func() error {
+		_, err = kbc.Kubectl.Delete(true, "-f", sampleFilePath)
+		return err
+	}, time.Minute, time.Second).Should(Succeed())
+
+	EventuallyWithOffset(1, func() error {
+		events, err := kbc.Kubectl.Get(true, "events", "--field-selector", "type", "Warning",
```
camilamacedo86, line 274:
```suggestion
		events, err := kbc.Kubectl.Get(true, "events", "--field-selector=type=Warning",
```
TO fix the command:
Screenshot 2022-07-05 at 23 03 18

testdata/project-v3-with-deploy-image/controllers/memcached_controller.go
```diff
@@ -83,6 +92,43 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			// Run finalization logic for memcachedFinalizer. If the
+			// finalization logic fails, don't remove the finalizer so
+			// that we can retry during the next reconciliation.
+			if err := r.finalizeMemcached(log, memcached); err != nil {
+				return ctrl.Result{}, err
+			}
+
+			// Remove memcachedFinalizer. Once all finalizers have been
+			// removed, the object will be deleted.
+			controllerutil.RemoveFinalizer(memcached, memcachedFinalizer)
```
camilamacedo86, line 123: @NikhilSharmaWe we need to update here to log the error:

```
if ok:= controllerutil.RemoveFinalizer(busybox, busyboxFinalizer); !ok{
				if err != nil {
					log.Error("Unable to remove the finalizer", "CR.Namespace", cr.Namespace, "CR.Name", cr.Name)
					return ctrl.Result{}, err
				}
			}
```

testdata/project-v3-with-deploy-image/controllers/memcached_controller.go
```diff
@@ -24,20 +24,28 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	"context"
+	"fmt"
 	"time"
 
+	"github.com/go-logr/logr"
```
camilamacedo86, line 30:
```suggestion
```

@NikhilSharmaWe we need to use the log from the controller and not this one.

testdata/project-v3-with-deploy-image/go.mod
```diff
@@ -3,6 +3,7 @@ module sigs.k8s.io/kubebuilder/testdata/project-v3-with-deploy-image
 go 1.18
 
 require (
+	github.com/go-logr/logr v1.2.0
```
camilamacedo86, line 6: We should not import this one

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -168,9 +213,26 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
...
+	// resources that are not owned by this CR, like a PVC.
+	// The following implementation will raise an event
+	r.recorder.Event(cr, "Warning", "Deleting",
+		fmt.Sprintf("Custom Resource %s is being deleted from the namespace %s",
+		cr.Name,
+		cr.Namespace))
+
+	log.Info("Successfully finalized {{ lower .Resource.Kind }}")
+	return nil
+}
```
camilamacedo86, line 234:
```suggestion
func (r *{{ .Resource.Kind }}Reconciler) finalize{{ .Resource.Kind }}(cr *{{ .Resource.ImportAlias }}.{{ .Resource.Kind }}) {
	// TODO(user): Add the cleanup steps that the operator
	// needs to do before the CR can be deleted. Examples
	// of finalizers include performing backups and deleting
	// resources that are not owned by this CR, like a PVC.
	// The following implementation will raise an event
	r.recorder.Event(cr, "Warning", "Deleting",
		fmt.Sprintf("Custom Resource %s is being deleted from the namespace %s",
		cr.Name,
		cr.Namespace))
}
```
Nothing here can return an error so we need to clean up it.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -131,6 +139,43 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
...
+
+	// Check if the {{ .Resource.Kind }} instance is marked to be deleted, which is
+	// indicated by the deletion timestamp being set.
+	is{{ .Resource.Kind }}MarkedToBeDeleted := {{ lower .Resource.Kind }}.GetDeletionTimestamp() != nil
+	if is{{ .Resource.Kind }}MarkedToBeDeleted {
+		if controllerutil.ContainsFinalizer({{ lower .Resource.Kind }}, {{ lower .Resource.Kind }}Finalizer) {
+			// Run finalization logic for memcachedFinalizer. If the
+			// finalization logic fails, don't remove the finalizer so
+			// that we can retry during the next reconciliation.
+			if err := r.finalize{{ .Resource.Kind }}(log, {{ lower .Resource.Kind }}); err != nil {
```
camilamacedo86, line 164:
```suggestion
			r.doFinalizerOperationsFor{{ .Resource.Kind }}({{ lower .Resource.Kind }})
```

camilamacedo86: Let;s try have a name that better clarifies what it does

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -131,6 +139,43 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
...
+				return ctrl.Result{}, err
+			}
```
camilamacedo86, line 166:
```suggestion
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -72,21 +72,28 @@ import (
 
 	"context"
 	"time"
+	"fmt"
+	"github.com/go-logr/logr"
```
camilamacedo86, line 76:
```suggestion
```

We should use the log obtained from the context (provided by the controller-runtime only)

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -131,6 +139,43 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
...
 	}
 
+	// Let's add a finalizer. Then, we can define some operations which should
+	// occurs before the custom resource to be deleted.
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/
+	// NOTE: You should not use finalizer to delete the resources that are
+	// created in this reconciliation and have the ownerRef set by ctrl.SetControllerReference
+	// because these will get deleted via k8s api
+	if !controllerutil.ContainsFinalizer({{ lower .Resource.Kind }}, {{ lower .Resource.Kind }}Finalizer) {
+		controllerutil.AddFinalizer({{ lower .Resource.Kind }}, {{ lower .Resource.Kind }}Finalizer)
```
camilamacedo86, line 149:
```suggestion
                 log.Info("Adding Finalizer for {{ Resource.Kind }}")
		controllerutil.AddFinalizer({{ lower .Resource.Kind }}, {{ lower .Resource.Kind }}Finalizer)
```
Let's add some logs so that we can check what it is doing and we are able to troubleshooting

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -66,27 +66,35 @@ package {{ if and .MultiGroup .Resource.Group }}{{ .Resource.PackageName }}{{ el
 import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	"errors"
```
camilamacedo86, line 73: can we use only apierrors "k8s.io/apimachinery/pkg/api/errors"? Why do we need both deps?

testdata/project-v3-with-deploy-image/controllers/memcached_controller.go
NikhilSharmaWe: Yes, I was using that log only but I was using this for specifying the parameters of the finalize function but since now that parameter is not needed(due some changes in the finalize function), I am removing it.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -66,27 +66,33 @@ package {{ if and .MultiGroup .Resource.Group }}{{ .Resource.PackageName }}{{ el
 import (
```
camilamacedo86, line 66: @NikhilSharmaWe,

 
The new occurs because the recorder is empty and needs to be passed when we create the controller and pass it in the manager ( main.go ):

```
if err = (&controllers.FoorxobReconciler{
   Client:   mgr.GetClient(),
   Scheme:   mgr.GetScheme(),
   Recorder: mgr.GetEventRecorderFor("mykind-controller"),
}).SetupWithManager(mgr); err != nil {
   setupLog.Error(err, "unable to create controller", "controller", "Foorxob")
   os.Exit(1)
}
//+kubebuilder:scaffold:builder
```


Also, in the controller, it needs to be changed to be Upper case: Recorder record.EventRecorder
Then, when we call the event use Eventf instead of Event  since we want to format the message

See that this PR does not have the e2e check as well. It is missing can you come back that?

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
```diff
@@ -120,6 +124,17 @@ func (s *apiScaffolder) Scaffold() error {
...
+func (s *apiScaffolder) updateMainEventRecorder() error {
+	defaultMainPath := "main.go"
+	err := util.InsertCode(defaultMainPath, `Scheme: mgr.GetScheme(),
+	`,
+		fmt.Sprintf(recorderTemplate, strings.ToLower(s.resource.Kind)))
+	if err != nil {
+		return fmt.Errorf("error scaffolding event recorder while creating the controller in main.go: %v", err)
+	}
+	return nil
+}
```
camilamacedo86, line 136: @NikhilSharmaWe,
 
Instead of use replace the best approach here would we create a MainUpdater like:

https://github.com/NikhilSharmaWe/kubebuilder/blob/f604a6cb2c4a3ba5425471ace96125630c9a50f3/pkg/plugins/golang/v3/scaffolds/api.go#L106-L110

Then, in this updater we will have another template for the controller in the main: https://github.com/NikhilSharmaWe/kubebuilder/blob/f604a6cb2c4a3ba5425471ace96125630c9a50f3/pkg/plugins/golang/v3/scaffolds/internal/templates/main.go#L100-L106

Could you please add a TODO on top of the method at least such as 

// TODO: replace this implementation by creating its own MainUpdater which will have its own controller template which set the recorder. 

Also, could you please change the target so we ensure that you ONLY set the event.Recorder on the Controller code of the specific api instead of adding the code in all of them?

camilamacedo86 (approved):
@NikhilSharmaWe,

The fix: https://github.com/kubernetes-sigs/kubebuilder/pull/2793/files#r915313169 is not a blocker for the PR. So, let's move forward and improve it in a follow up

/lgtm
/approved

PR #2795, opened 2022-07-04
🌱 (deploy-image/v1-alpha) - e2e test checking if the pods are running

camilamacedo86:
## Description
e2e test checking if the pods are running


test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -120,7 +120,7 @@ var _ = Describe("kubebuilder", func() {
 			By("uncomment kustomize files to ensure that pods are restricted")
 			uncommentPodStandards(kbc)
 
-			Run(kbc, "memcached:1.4.36-alpine")
```
NikhilSharmaWe, line 123: @camilamacedo86 Could you please explain why the e2e tests were failing when we were checking that the pod which contains the `memcached` substring in its crd name is running?

camilamacedo86: We will not have a memcached substring on it because the KVG in the tests are random values and not Memcached, see: https://github.com/kubernetes-sigs/kubebuilder/blob/a864fbf848a5e36062e7b50be6b920119ca87ff8/test/e2e/deployimage/plugin_cluster_test.go#L102-L104

PR #2796, opened 2022-07-04
feat: use runtime package to describe goos & goarch.

vpc123:

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"runtime"
 )
 
 // var needs to be used instead of const as ldflags is used to fill this
@@ -25,8 +26,8 @@ import (
 var (
 	kubeBuilderVersion      = "unknown"
 	kubernetesVendorVersion = "unknown"
-	goos                    = "unknown"
-	goarch                  = "unknown"
+	goos                    = runtime.GOOS
+	goarch                  = runtime.GOARCH
 	gitCommit               = "$Format:%H$" // sha1 from git, output of $(git rev-parse HEAD)
 
 	buildDate = "1970-01-01T00:00:00Z" // build date in ISO8601 format, output of $(date -u +'%Y-%m-%dT%H:%M:%SZ')


everettraven:
Hi @vpc123! Thanks for the contribution!

While I like the idea of the change, I think there is a specific reason we set the `goos` and `goarch` global variables to be `"unknown"`. In the `Makefile` when building/installing Kubebuilder we set the ldflags to:
https://github.com/kubernetes-sigs/kubebuilder/blob/8b0cff163a7557af6a4059cf5de62e760336bbac/Makefile#L48-L54

This allows us to ensure that a binary is built for a specific environment and that the platform the application is built for does not change at runtime. (i.e a binary built for `darwin/arm64` should always show `darwin/arm64` in the version string. If the same binary is ran on a `linux/arm64` platform it should still show the `darwin/arm64` platform in the version string)

@camilamacedo86 what do you think?

vpc123:
> Hi @vpc123! Thanks for the contribution!
> 
> While I like the idea of the change, I think there is a specific reason we set the `goos` and `goarch` global variables to be `"unknown"`. In the `Makefile` when building/installing Kubebuilder we set the ldflags to:
> 
> https://github.com/kubernetes-sigs/kubebuilder/blob/8b0cff163a7557af6a4059cf5de62e760336bbac/Makefile#L48-L54
> 
> This allows us to ensure that a binary is built for a specific environment and that the platform the application is built for does not change at runtime. (i.e a binary built for `darwin/arm64` should always show `darwin/arm64` in the version string. If the same binary is ran on a `linux/arm64` platform it should still show the `darwin/arm64` platform in the version string)
> 
> @camilamacedo86 what do you think?

Thank you for your answer. I understand the reason for this. 
I will have a deeper understanding of the purpose of the project，
I just think it may be an optimization direction, maybe because I usually use server to develop code, thank you

cmd/version.go
```diff
@@ -18,15 +18,16 @@ package main
...
 
 // var needs to be used instead of const as ldflags is used to fill this
 // information in the release process
 var (
 	kubeBuilderVersion      = "unknown"
 	kubernetesVendorVersion = "unknown"
-	goos                    = "unknown"
-	goarch                  = "unknown"
+	goos                    = runtime.GOOS
+	goarch                  = runtime.GOARCH
```
camilamacedo86, line 30: The info will be filled, see:

> $ kubebuilder version
Version: main.version{KubeBuilderVersion:"v2.0.0-1506-g7e83b313-dirty", KubernetesVendor:"unknown", GitCommit:"7e83b31354bab9d3d875f36790d0e3f33ca32c8e", BuildDate:"2022-07-04T19:01:54Z", GoOs:"darwin", GoArch:"amd64"}

So, could you please let us know more about the motivation for this one?

camilamacedo86:
Hi @vpc123 

The @everettraven's explanation in https://github.com/kubernetes-sigs/kubebuilder/pull/2796#issuecomment-1175131473 is right. See the comment on top of the code:

// var needs to be used instead of const as ldflags is used to fill this
// information in the release process

So, i think we can close this one as deferred and I hope that you do not mind @vpc123.
However, if you have any motivation like ( when I do X I would expect Y and because of this I face Z ) to raise this pr please raise an issue for we know more about and see how we can help with it.

Also, thank you for the contribution and please feel free to help out with the project.
Your collaboration is more than welcome.

PR #2797, opened 2022-07-05
:sparkles: (grafana/v1-alpha) : add a new dashboard to show CPU and Memory usage for Pods

Kavinjsir:
### Description of the Change
Add function of Grafana plugin to scaffold new dashboard manifest of cpu&memory. Changes include:
- Add scaffolder to generate manifest
- Update e2e test cases to cover the feature
- Documentation of the new feature

### Motivation
Enhance the function of Grafana plugin to observe controller in multiple perspectives.

### Sample Dashboard based on the Manifest:
Screen Shot 2022-07-04 at 10 43 06 PM

### DIFF
--- a/docs/book/src/plugins/grafana-v1-alpha.md
+++ b/docs/book/src/plugins/grafana-v1-alpha.md
@@ -77,8 +77,20 @@ See an example of how to use the plugin in your project:
 - Description:
   - Per-second rate of total reconciliation as measured over the last 5 minutes
   - Per-second rate of reconciliation errors as measured over the last 5 minutes
+- Sample: <img width="1430" src="https://user-images.githubusercontent.com/18136486/176122555-f3493658-6c99-4ad6-a9b7-63d85620d370.png">
 
-<img width="1430" alt="Screen Shot 2022-06-28 at 3 43 10 AM" src="https://user-images.githubusercontent.com/18136486/176122555-f3493658-6c99-4ad6-a9b7-63d85620d370.png">
+#### CPU & Memory Usage
+
+- Metrics:
+  - process_cpu_seconds_total
+  - process_resident_memory_bytes
+- Query:
+  - rate(process_cpu_seconds_total{job="$job", namespace="$namespace", pod="$pod"}[5m]) \* 100
+  - process_resident_memory_bytes{job="$job", namespace="$namespace", pod="$pod"}
+- Description:
+  - Per-second rate of CPU usage as measured over the last 5 minutes
+  - Allocated Memory for the running controller
+- Sample: <img width="1381" src="https://user-images.githubusercontent.com/18136486/177239808-7d94b17d-692c-4166-8875-6d9332e05bcb.png">
 
 ## Subcommands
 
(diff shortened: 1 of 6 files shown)


docs/book/src/plugins/grafana-v1-alpha.md
```diff
@@ -77,8 +77,20 @@ See an example of how to use the plugin in your project:
...
 
-<img width="1430" alt="Screen Shot 2022-06-28 at 3 43 10 AM" src="https://user-images.githubusercontent.com/18136486/176122555-f3493658-6c99-4ad6-a9b7-63d85620d370.png">
+#### Controller Resources Usage (CPU, Memory)
+
+- Metrics:
+  - process_cpu_seconds_total
+  - process_resident_memory_bytes
+- Query:
+  - rate(process_cpu_seconds_total{job="$job", namespace="$namespace", pod="$pod"}[5m]) \* 100
+  - process_resident_memory_bytes{job="$job", namespace="$namespace", pod="$pod"}
```
camilamacedo86, line 89: but if we are looking for any pod that is not the manager one right?
WDYT about we looking for the label : https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/config/manager/manager.yaml#L5

Kavinjsir: > but if we are looking for any pod that is not the manager one right?

- Since this `process_resident_memory_bytes` metrics [roots from controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.12.2/pkg/internal/controller/metrics/metrics.go#L74), I think it is for controller-manager.
- What else pods that the `cpu/memory` panels may cover?

> WDYT about we looking for the label?

- Are you suggesting that we compose the query to select cpu/memory metrics based on `K8s labels`?
    Guess it may be hard, because the prometheus metrics by default may not contain the info of `K8s labels` right?
    For example, this `process_resident_memory_bytes`'s own labels do not cover `k8s labels` for the controller:
   Screen Shot 2022-07-05 at 12 58 25 PM

Kavinjsir: Am I missing the point that we are also expecting to monitor the status of the pods that are **launched by the operator/manager**?
(For example: the pod of a job that is scheduled by a customized cron-job operator?)

camilamacedo86: @Kavinjsir , 

But we can use the container name in this way. What do you think about we filter by container=manager?

varshaprasad96: @camilamacedo86 @Kavinjsir  The K8s labels are configurable and can be changed by the user, so that would come under custom metrics. If `container=manager` label is something which kb scaffolding [adds](https://github.com/kubernetes-sigs/kubebuilder/blob/c329411c85ce9df7ae20156713bca8b9f01fd9cb/testdata/project-v3/config/manager/manager.yaml#L5) it, then its pretty easy to change it and error out since we hardcode it. What do you think about using labels but letting the user to pass them? If not by maybe parsing the deployment manifest where manager is defined.

camilamacedo86 (approved):
@Kavinjsir,

The title is used to generate the changelog. So, I am changing it for:

:sparkles: (grafana/v1-alpha) : add new dashboard to show CPU and Memory usage for Pods

varshaprasad96 (approved):
/lgtm

PR #2810, opened 2022-07-07
:seedling: (phase 2 plugins): add simple external plugin to testdata

everettraven:
## Description of the change
Add a new project to the `testdata/` directory that is a very simple Phase 2 Plugin written in Go.

The implementation also includes comments that could be helpful for those attempting to write their own Phase 2 Plugins.

## Motivation for the change
Create a sample Phase 2 Plugin that can be used to help with e2e testing the Phase 2 Plugins implementation.


testdata/phase-2-plugins/simple-external-plugin/main.go
```diff
@@ -0,0 +1,328 @@
+/*
```
camilamacedo86, line 1: @everettraven that is great. 
However, we cannot add under testdata
Why?
All under this directory is removed and generated programmatically, see:
https://github.com/kubernetes-sigs/kubebuilder/blob/master/Makefile#L69-L72

**So, where to add the sample?**
What about we have dir for we use the sample on the docs as tutorial for how to write your own plugin?
That would be like: https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/component-config-tutorial

And we could add here:
https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/plugins

Therefore we can use this code in a tutorial on the docs/ WDYT?

everettraven: @camilamacedo86 that makes sense! I can work on moving it!

testdata/phase-2-plugins/simple-external-plugin/main.go
```diff
@@ -0,0 +1,328 @@
...
+		return fmt.Errorf("encountered error marshaling output: %w | OUTPUT: %s", err, output)
+	}
+
+	fmt.Printf("%s", output)
+
+	return nil
+}
+
+// initCmd handles all the logic for the `init` subcommand of this sample external plugin
+func initCmd(pr *external.PluginRequest) external.PluginResponse {
```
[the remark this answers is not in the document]
everettraven: @camilamacedo86  Sure thing!

everettraven:
@camilamacedo86 @rashmigottipati Would you mind reviewing this again when you have some time? Thanks!

docs/book/src/simple-external-plugin-tutorial/api.go
```diff
@@ -0,0 +1,46 @@
+/*
```
camilamacedo86, line 1: Could we:
- Add the plugin inside of a directory called tesdata? Why? In Golang it is a good practice to know that is a mock/sample. So that would be `docs/book/src/simple-external-plugin-tutorial/testdata/` dir instead
- Could we have a better layout for this sample , i.e.:

docs/book/src/simple-external-plugin-tutorial/testdata/
|--- pkg
         | ---> we add here api.go/ init.go, helpers.go and etc
- then we have under main.go, go.mod, go.sum

Also, have we not templates in this plugin? 
If so, could we have them instead of templates dir much more similar to how we structure our plugin? Look the pkg/plugins/golang/v3

docs/book/src/simple-external-plugin-tutorial/api.go
```diff
@@ -0,0 +1,46 @@
...
+Copyright 2022 The Kubernetes Authors.
+
```
camilamacedo86, line 3: Could we add here a Makefile with a target to build the plugin in the .bin/ dir ( we need ignore this dir as well )
Then, could we add an e2e test that just creates a project using the plugin and test it out against the Kind? 
That would be here: `kubebuilder/test/e2e/external-plugin/`

everettraven: I feel like this would be best in a follow up. I was hoping to keep the scope of this PR smaller (i.e just adding the sample) and add the e2e testing components in a later PR (I think @rashmigottipati mentioned that she had already started on some e2e testing work).

camilamacedo86: That is fine.

everettraven:
Just wanted to post an update: 
I haven't had time to work on this recently due to other priority work on other projects. I am still planning to address all the review comments when I have some more time. Thanks!

docs/book/src/simple-external-plugin-tutorial/api.go
everettraven: @camilamacedo86 I updated the structure of the sample external plugin to be more similar to the existing Kubebuilder plugins in commit https://github.com/kubernetes-sigs/kubebuilder/pull/2810/commits/e9e5c6b7f92df0b8a5956f596bb4ea2cb38da39b

Let me know what you think!

docs/book/src/simple-external-plugin-tutorial/api.go
everettraven: @camilamacedo86 I added a Makefile and a target named `build` that when run will build the sample external plugin and output the binary to `./bin/sampleexternalplugin` in commit https://github.com/kubernetes-sigs/kubebuilder/pull/2810/commits/ee04eee08658465719e07af06f30d100b7484bb1

If we still want to do a simple e2e test to ensure that the phase 2 plugins are working properly with this sample plugin I can create a simple test to ensure that the proper files are created and contain the right contents after running scaffold commands with the sample plugin. Unfortunately this sample doesn't scaffold anything that can actually be run on a cluster and just produces text files.

docs/book/src/simple-external-plugin-tutorial/testdata/sampleexternalplugin/v1/Makefile
```diff
@@ -0,0 +1,4 @@
+SHELL = /bin/bash
+
+build:
+	go build -o ./bin/sampleexternalplugin
```
camilamacedo86, line 4: nit just an space

docs/book/src/simple-external-plugin-tutorial/testdata/sampleexternalplugin/v1/scaffolds/internal/templates/webhook/webhookFile.go
```diff
@@ -0,0 +1,55 @@
...
+
+	if webhookFile.hooked {
+		webhookFile.Contents += WebhookFileHookedMessage
+	}
+
+	return webhookFile
+}
+
+const WebhookFileDefaultMessage = "A simple text file created with the `create webhook` subcommand"
+const WebhookFileHookedMessage = "\nHOOKED!"
```
camilamacedo86, line 55: In this example we are scaffolding a new file when we call the webhook command with, Am I right?
So, wdyt about a sample that can be called in the init or create api to be something more simple?

Note that to scaffold the webhook we need to have a resource created first. 
Also, the sample could for example just add a file when we create the project or replace/add content when we call create api for example.

camilamacedo86 (approved):
Hi @everettraven,

Since it is an initial version of the testdata sample as we spoke I think we can move forward here to make eaiser the process to add the sample, docs, and e2e tests instead of trying to do it all in the same PR. Therefore, how you agree to work on the following as when possible shows all fine. 

So, 

/lgtm
/approved

PR #2816, opened 2022-07-09
✨deploy-image/v1alpha: add code for updating controller status when the deployment is created successfully

NikhilSharmaWe:
### Description
This PR works on updating controller status when the deployment is created successfully.
### Motivation


pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -221,6 +222,23 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	var statusMessage string
+	successStatusMessage := "Deployment %s/%s have correct number of AvailableReplicas which is %s"
+	failStatusMessage := "Deployment %s/%s have incorrect number of AvailableReplicas, have %s but should be %s"
+	if found.Status.AvailableReplicas != size {
+		statusMessage = fmt.Sprintf(failStatusMessage, {{ lower .Resource.Kind }}.Name, {{ lower .Resource.Kind }}.Namespace, found.Status.AvailableReplicas, size)
```
FillZpp, line 229: `%s/%s` should be `{{ lower .Resource.Kind }}.Namespace, {{ lower .Resource.Kind }}.Name`.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -221,6 +222,23 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
...
+	} else {
+		statusMessage = fmt.Sprintf(successStatusMessage, {{ lower .Resource.Kind }}.Name, {{ lower .Resource.Kind }}.Namespace, size)
```
FillZpp, line 231: Same as above.

testdata/project-v3-with-deploy-image/controllers/busybox_controller.go
```diff
@@ -172,6 +173,23 @@ func (r *BusyboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	var statusMessage string
+	successStatusMessage := "Deployment %s/%s have correct number of AvailableReplicas which is %s"
+	failStatusMessage := "Deployment %s/%s have incorrect number of AvailableReplicas, have %s but should be %s"
+	if found.Status.AvailableReplicas != size {
+		statusMessage = fmt.Sprintf(failStatusMessage, busybox.Name, busybox.Namespace, found.Status.AvailableReplicas, size)
```
FillZpp, line 180: Same as above.

testdata/project-v3-with-deploy-image/controllers/busybox_controller.go
```diff
@@ -172,6 +173,23 @@ func (r *BusyboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	} else {
+		statusMessage = fmt.Sprintf(successStatusMessage, busybox.Name, busybox.Namespace, size)
```
FillZpp, line 182: Same as above.

testdata/project-v3-with-deploy-image/controllers/memcached_controller.go
```diff
@@ -172,6 +173,23 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	var statusMessage string
+	successStatusMessage := "Deployment %s/%s have correct number of AvailableReplicas which is %s"
+	failStatusMessage := "Deployment %s/%s have incorrect number of AvailableReplicas, have %s but should be %s"
+	if found.Status.AvailableReplicas != size {
+		statusMessage = fmt.Sprintf(failStatusMessage, memcached.Name, memcached.Namespace, found.Status.AvailableReplicas, size)
```
FillZpp, line 180: Same as above.

testdata/project-v3-with-deploy-image/controllers/memcached_controller.go
```diff
@@ -172,6 +173,23 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	} else {
+		statusMessage = fmt.Sprintf(successStatusMessage, memcached.Name, memcached.Namespace, size)
```
FillZpp, line 182: Same as above.

testdata/project-v4-with-deploy-image/controllers/busybox_controller.go
```diff
@@ -172,6 +173,23 @@ func (r *BusyboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	var statusMessage string
+	successStatusMessage := "Deployment %s/%s have correct number of AvailableReplicas which is %s"
+	failStatusMessage := "Deployment %s/%s have incorrect number of AvailableReplicas, have %s but should be %s"
+	if found.Status.AvailableReplicas != size {
+		statusMessage = fmt.Sprintf(failStatusMessage, busybox.Name, busybox.Namespace, found.Status.AvailableReplicas, size)
```
FillZpp, line 180: Same as above.

testdata/project-v4-with-deploy-image/controllers/busybox_controller.go
```diff
@@ -172,6 +173,23 @@ func (r *BusyboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	} else {
+		statusMessage = fmt.Sprintf(successStatusMessage, busybox.Name, busybox.Namespace, size)
```
FillZpp, line 182: Same as above.

testdata/project-v4-with-deploy-image/controllers/memcached_controller.go
```diff
@@ -172,6 +173,23 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	var statusMessage string
+	successStatusMessage := "Deployment %s/%s have correct number of AvailableReplicas which is %s"
+	failStatusMessage := "Deployment %s/%s have incorrect number of AvailableReplicas, have %s but should be %s"
+	if found.Status.AvailableReplicas != size {
+		statusMessage = fmt.Sprintf(failStatusMessage, memcached.Name, memcached.Namespace, found.Status.AvailableReplicas, size)
```
FillZpp, line 180: Same as above.

testdata/project-v4-with-deploy-image/controllers/memcached_controller.go
```diff
@@ -172,6 +173,23 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	} else {
+		statusMessage = fmt.Sprintf(successStatusMessage, memcached.Name, memcached.Namespace, size)
```
FillZpp, line 182: Same as above.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -221,6 +222,23 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
...
+	}
```
jmrodri, line 232: Can we initialize `statusMessage` to the successful case by default? that way the if block only needs to see if it is a failure case. 

```golang
successStatusMessage := "Deployment %s/%s have correct number of AvailableReplicas which is %s"
failStatusMessage := "Deployment %s/%s have incorrect number of AvailableReplicas, have %s but should be %s"
statusMessage := fmt.Sprintf(successStatusMessage, {{ lower .Resource.Kind }}.Name, {{ lower .Resource.Kind }}.Namespace, size)

if found.Status.AvailableReplicas != size {
	statusMessage = fmt.Sprintf(failStatusMessage, {{ lower .Resource.Kind }}.Name, {{ lower .Resource.Kind }}.Namespace, found.Status.AvailableReplicas, size)
}
```

I mean it's probably no big deal if we don't but it does remove the extra else. Basically assume it is going to work and if not, handle the failure.

jmrodri (commented):
I'm not going to block the PR on my review, but I think setting the status to success is better.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/api/types.go
```diff
@@ -89,6 +89,8 @@ type {{ .Resource.Kind }}Spec struct {
 type {{ .Resource.Kind }}Status struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
+
+	DeploymentStatus string ` + "`" + `json:"deploymentStatus,omitempty"` + "`" + `
```
camilamacedo86, line 93: To follow the good practices and shows how to do it we should use conditional status therefore here we would have: (see the doc: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)

```suggestion
	// Represents the observations of a {{ .Resource.Kind }}'s current state.
	// .{{ .Resource.Kind }}.status.conditions.type are: "Available", "Progressing", and "Degraded"
	// .{{ .Resource.Kind }}.status.conditions.status are one of True, False, Unknown.
	// .{{ .Resource.Kind }}.status.conditions.reason the value should be a CamelCase string and producers of specific
	// condition types may define expected values and meanings for this field, and whether the values
	// are considered a guaranteed API.
	// .{{ .Resource.Kind }}.status.conditions.Message is a human readable message indicating details about the transition.
	// For further information see: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
```

Also, we can remove the comment since we are scaffolding the code.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -221,6 +222,23 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
...
+	if !reflect.DeepEqual(statusMessage, {{ lower .Resource.Kind }}.Status.DeploymentStatus){
+		{{ lower .Resource.Kind }}.Status.DeploymentStatus = statusMessage
+		err := r.Status().Update(ctx, {{ lower .Resource.Kind }})
+		if err != nil {
+			log.Error(err, "Failed to update {{ lower .Resource.Kind }} status")
+			return ctrl.Result{}, err
+		}
+	}
+
```
camilamacedo86, line 241: Hi @NikhilSharmaWe,  

About the status that would be like something: https://github.com/kubernetes-sigs/kubebuilder/compare/master...camilamacedo86:kubebuilder:test-br?expand=1

See that the convention defines how we should fille the conditional one and we need to keep a track of the transaction. You can check the description of its field/spec in the code implementation here: https://github.com/kubernetes-sigs/kubebuilder/pull/2816/files#r919781621 (were we are commenting what should be added to each case scenario)

- See the example (https://kpt.dev/reference/schema/crd-status-convention/) it might help you get a better idea.
- Also, ensure that you check the documentation: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
camilamacedo86: We need to change this one. It is not using status Conditions and follows the recommendations.
If we want to add status for this transition then it would be like 
status.conditions.type = Available
status.conditions.status = True
status.conditions.reason = "Resizing"
.status.conditions.Message = "Resizing available replicas of the Deployment %s owned by the custom resource %s from %d to %d "

Therefore if we fails to update the resize then;

status.conditions.type = Available
status.conditions.status = False
status.conditions.reason = "Resizing"
.status.conditions.Message = "Fail to resizing available replicas of the Deployment %s owned by the custom resource %s from %d to %d : (%s) " -> Pass the error faced

Look at : https://github.com/kubernetes-sigs/kubebuilder/compare/master...camilamacedo86:kubebuilder:test-br?expand=1

Also, in the example; https://kpt.dev/reference/schema/crd-status-convention/ ( it will help you understand better its idea )

testdata/project-v3-with-deploy-image/api/v1alpha1/memcached_types.go
```diff
@@ -67,6 +76,24 @@ type MemcachedList struct {
 	Items           []Memcached `json:"items"`
 }
```
camilamacedo86, line 78: Nit. Could we add a comment describing what this method does?

testdata/project-v3-with-deploy-image/controllers/busybox_controller.go
```diff
@@ -164,6 +201,17 @@ func (r *BusyboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		err = r.Update(ctx, found)
 		if err != nil {
 			log.Error(err, "Failed to update Deployment", "Deployment.Namespace", found.Namespace, "Deployment.Name", found.Name)
+
+			// The following implementation will update the status
+			busybox.SetConditionBusybox(metav1.Condition{Type: typeDegradedBusybox,
+				Status: "True", Reason: "Resizing",
```
camilamacedo86, line 207: Could we use the const defined in the lib `metav1.ConditionTrue`?
same in the other places

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -263,6 +266,24 @@ func Run(kbc *utils.TestContext) {
 	}
 	EventuallyWithOffset(1, getMemcachedPodStatus, time.Minute, time.Second).Should(Succeed())
 
+	By("validating that the status of the custom resource created is updated or not")
+	var status string
+	var customResourceConditions metav1.Condition
+	//nolint:lll
+	getStatus := func() error {
+		status, err = kbc.Kubectl.Get(true, strings.ToLower(kbc.Kind),
```
camilamacedo86, line 274: Can we do as above ?

```
status, err := kbc.Kubectl.Get(true, "pods", "-l",
			fmt.Sprintf("app.kubernetes.io/name=%s", strings.ToLower(kbc.Kind)),
			"-o", "jsonpath={.items[*].status}",
		)
		ExpectWithOffset(2, err).NotTo(HaveOccurred())
```

Then, if has no error we can unmarshal the result and check.
Also, what is returned by {`.status.conditions}`?

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -263,6 +266,24 @@ func Run(kbc *utils.TestContext) {
...
+		if err == nil && strings.TrimSpace(status) == "" {
+			return errors.New("empty status, continue")
+		} else if err := goyaml.Unmarshal([]byte(status), &customResourceConditions); err != nil {
+			return errors.New("unable to Unmarshal the status")
+		} else if customResourceConditions.Type != "typeAvailable" || customResourceConditions.Status != metav1.ConditionTrue {
+			return errors.New(`conditions type should be "typeAvailable" and status should be true`)
+		}
+		return nil
+	}
+	Eventually(getStatus, time.Minute, time.Second).Should(Succeed())
```
camilamacedo86, line 285: After we unmarshal we can get the result and do the check outside 
We should ensure that you have typeAvailable == true

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -139,6 +147,13 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
 		return ctrl.Result{}, err
 	}
 
+	meta.SetStatusCondition(&{{ lower .Resource.Kind }}.Status.Conditions, metav1.Condition{Type: typeProgressing{{ .Resource.Kind }}, Status: metav1.ConditionTrue, Reason: "Reconciling", Message: "Starting reconciliation"})
+	err = r.Status().Update(ctx, {{ lower .Resource.Kind }})
+	if err != nil {
```
camilamacedo86, line 152:
```suggestion
	if err = r.Status().Update(ctx, {{ lower .Resource.Kind }}); err != nil {
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -159,21 +174,33 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
...
 			// that we can retry during the next reconciliation.
 			log.Info("Performing Finalizer Operations for {{ .Resource.Kind }} before delete CR")
+
+			// The following implementation will update the status
+			meta.SetStatusCondition(&{{ lower .Resource.Kind }}.Status.Conditions, metav1.Condition{Type: typeDegraded{{ .Resource.Kind }},
+				Status: metav1.ConditionTrue, Reason: "Finalizing",
+				Message: fmt.Sprintf("Performing finalizer operations for the custom resource: %s ", {{ lower .Resource.Kind }}.Name)})
+
+			err := r.Status().Update(ctx, {{ lower .Resource.Kind }})
+			if err != nil {
```
camilamacedo86, line 188:
```suggestion
			if err := r.Status().Update(ctx, {{ lower .Resource.Kind }}); err != nil {
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
camilamacedo86: closing because I think it changed. But I will ping you for a final review when it be done.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
camilamacedo86: The implementation change and @NikhilSharmaWe are working on the tests. We will need your help on the review when he finishes by sure :-) @jmrodri

camilamacedo86 (dismissed):

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -215,6 +251,17 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
 		if err = r.Update(ctx, found); err != nil {
 			log.Error(err, "Failed to update Deployment", 
 				"Deployment.Namespace", found.Namespace, "Deployment.Name", found.Name)
+
+			// The following implementation will update the status
+			meta.SetStatusCondition(&{{ lower .Resource.Kind }}.Status.Conditions, metav1.Condition{Type: typeDegraded{{ .Resource.Kind }},
+				Status: metav1.ConditionTrue, Reason: "Resizing",
```
camilamacedo86, line 257:
```suggestion
				Status: metav1.ConditionFalse, Reason: "Resizing",
```

We could not  resising

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -88,6 +89,13 @@ import (
 
 const {{ lower .Resource.Kind }}Finalizer = "{{ .Resource.Group }}.{{ .Resource.Domain }}/finalizer"
 
+// Definitions to manage status conditions
+const (
+	typeAvailable{{ .Resource.Kind }} = "Available"
+	typeProgressing{{ .Resource.Kind }} = "Progressing"
+	typeDegraded{{ .Resource.Kind }} = "Degraded"
+)
```
camilamacedo86, line 97: @NikhilSharmaWe : 

```suggestion
const (
	// typeAvailable{{ .Resource.Kind }} represents the status of the Deployment reconciliation
	typeAvailable{{ .Resource.Kind }} = "Available"
	// typeDegraded{{ .Resource.Kind }} represents the status used when the custom resource is deleted
	// and the finalizer operations must to occurs. 
	typeDegraded{{ .Resource.Kind }} = "Degraded"
)
```

We can have only 2 contionalTypes. 

See: 
> Conditions provide a standard mechanism for higher-level status reporting from a controller. They are an extension mechanism which allows tools and other controllers to collect summary information about resources without needing to understand resource-specific status details. Conditions should complement more detailed information about the observed status of an object written by a controller, rather than replace it. For example, the "Available" condition of a Deployment can be determined by examining readyReplicas, replicas, and other properties of the Deployment. However, the "Available" condition allows other components to avoid duplicating the availability logic in the Deployment controller.

Available -. Status True would mean The Deployment is available and with the expected replicas
Available -  Status False is when it has an error and it is not available yet or is not with the specific size replicase
Available - Without status is when we begin the reconciliation and we do have not the Available donation so far
Degraded - we should add with no status when we start the finalizer and we should update with status true when we finish the finalise 

More info: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -138,6 +146,12 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
 		return ctrl.Result{}, err
 	}
 
+	meta.SetStatusCondition(&{{ lower .Resource.Kind }}.Status.Conditions, metav1.Condition{Type: typeProgressing{{ .Resource.Kind }}, Status: metav1.ConditionTrue, Reason: "Reconciling", Message: "Starting reconciliation"})
```
camilamacedo86, line 149: @NikhilSharmaWe : 

```suggestion
	meta.SetStatusCondition(&{{ lower .Resource.Kind }}.Status.Conditions, metav1.Condition{Type: typeAvailable{{ .Resource.Kind }},, Reason: "Reconciling", Message: "Starting reconciliation"})
```

See from the API convention : 

> Controllers should apply their conditions to a resource the first time they visit the resource, even if the status is Unknown. This allows other components in the system to know that the condition exists and the controller is making progress on reconciling that resource.

> Not all controllers will observe the previous advice about reporting "Unknown" or "False" values. For known conditions, the absence of a condition status should be interpreted the same as Unknown, and typically indicates that reconciliation has not yet finished (or that the resource state may not yet be observable).

Also, before to add this status we need first check to have we status condition Available already? If not then we add.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -160,6 +174,17 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
 	if is{{ .Resource.Kind }}MarkedToBeDeleted {
 		if controllerutil.ContainsFinalizer({{ lower .Resource.Kind }}, {{ lower .Resource.Kind }}Finalizer) {
 			log.Info("Performing Finalizer Operations for {{ .Resource.Kind }} before delete CR")
+
+			// The following implementation will update the status
+			meta.SetStatusCondition(&{{ lower .Resource.Kind }}.Status.Conditions, metav1.Condition{Type: typeDegraded{{ .Resource.Kind }},
+				Status: metav1.ConditionTrue, Reason: "Finalizing",
```
camilamacedo86, line 180:
```suggestion
				, Reason: "Finalizing", Message: "Performing finalizer operations",
```

@NikhilSharmaWe same reasons here, since we are beginning to perform the finalizer operations we need to add it without the status. Therefore, before we set we need to ensure that we have NOT the status Downgrade added to the CR already. It means, we did not tried to do the finalizer operations and that fails.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -160,6 +174,17 @@ func (r *{{ .Resource.Kind }}Reconciler) Reconcile(ctx context.Context, req ctrl
...
+				Message: fmt.Sprintf("Performing finalizer operations for the custom resource: %s ", {{ lower .Resource.Kind }}.Name)})
+
+			if err := r.Status().Update(ctx, {{ lower .Resource.Kind }}); err != nil {
+				log.Error(err, "Failed to update {{ .Resource.Kind }} status")
+				return ctrl.Result{}, err
+			}
+
 			r.doFinalizerOperationsFor{{ .Resource.Kind }}({{ lower .Resource.Kind }})
```
camilamacedo86, line 189:
```suggestion
			// TODO(user): If you add operations to the doFinalizerOperationsFor{{ .Resource.Kind }} method then you need to ensure that all worked fine before deleting and updating the Downgrade status
```

camilamacedo86: @NikhilSharmaWe we need after performing the doFrinalizer operations update teh Downgrade status for True, with a message like "Finalize operations for customer resource (%s name) were successfully accomplished"

testdata/project-v3-with-deploy-image/controllers/busybox_controller.go
```diff
@@ -167,6 +203,17 @@ func (r *BusyboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		if err = r.Update(ctx, found); err != nil {
 			log.Error(err, "Failed to update Deployment",
 				"Deployment.Namespace", found.Namespace, "Deployment.Name", found.Name)
+
+			// The following implementation will update the status
+			meta.SetStatusCondition(&busybox.Status.Conditions, metav1.Condition{Type: typeDegradedBusybox,
```
camilamacedo86, line 208:
```suggestion
			meta.SetStatusCondition(&busybox.Status.Conditions, metav1.Condition{Type: typeAvailableBusybox,
```

camilamacedo86: @NikhilSharmaWe here we need to use the Available one

testdata/project-v3-with-deploy-image/controllers/busybox_controller.go
```diff
@@ -167,6 +203,17 @@ func (r *BusyboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+				Status: metav1.ConditionTrue, Reason: "Resizing",
```
camilamacedo86, line 209:
```suggestion
				Status: metav1.ConditionFalse, Reason: "Resizing",
```

testdata/project-v3-with-deploy-image/controllers/busybox_controller.go
```diff
@@ -176,6 +223,16 @@ func (r *BusyboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	// The following implementation will update the status
+	meta.SetStatusCondition(&busybox.Status.Conditions, metav1.Condition{Type: typeAvailableBusybox,
+		Status: metav1.ConditionTrue, Reason: "Reconciling",
+		Message: fmt.Sprintf("Deployment for custom resource (%s) created succefully", busybox.Name)})
```
camilamacedo86, line 229:
```suggestion
		Message: fmt.Sprintf("Deployment for custom resource (%s) with %d replicas created successfully ", busybox.Name)}) 
```

@NikhilSharmaWe let's add the number of replicas here, The size as well

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
camilamacedo86: @NikhilSharmaWe the setStatus implementation already check if that exist or not so we just need to use the set directly without the need to check, see: https://github.com/kubernetes/apimachinery/blob/v0.24.3/pkg/api/meta/conditions.go#L30

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
camilamacedo86: @NikhilSharmaWe the setStatus implementation already check if that exist or not so we just need to use the set directly without the need to check, see: https://github.com/kubernetes/apimachinery/blob/v0.24.3/pkg/api/meta/conditions.go#L30

camilamacedo86 (approved):
@NikhilSharmaWe Terrific work 🥇 

That is great for me !
/approved

testdata/project-v3-with-deploy-image/controllers/busybox_controller.go
```diff
@@ -112,8 +127,29 @@ func (r *BusyboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	if isBusyboxMarkedToBeDeleted {
 		if controllerutil.ContainsFinalizer(busybox, busyboxFinalizer) {
 			log.Info("Performing Finalizer Operations for Busybox before delete CR")
+
+			// The following implementation will update the status
+			meta.SetStatusCondition(&busybox.Status.Conditions, metav1.Condition{Type: typeDegradedBusybox,
```
camilamacedo86, line 132: /hold 

we need to fetch all the Memcached before to do the update 
therefore, the best approach is ensure that we have a method that does what we do here [https://github.com/kubernetes-sigs/kubebuilder/blob/149d3a2acd7a76ead0a54afd6ef4b5[…]/project-v3-with-deploy-image/controllers/busybox_controller.go](https://github.com/kubernetes-sigs/kubebuilder/blob/149d3a2acd7a76ead0a54afd6ef4b583ea505911/testdata/project-v3-with-deploy-image/controllers/busybox_controller.go#L85-L100)

and we will call it so we do not duplicate code

c/c @NikhilSharmaWe

camilamacedo86:
/hold cancel 

It shows great ! 
Good job 🥇

camilamacedo86:
Let's move forward with this one so that we can acctually get a lot of feedbacks.
The plugin is alpha and we can change/improve things in follow-ups.

/lgtm

PR #2817, opened 2022-07-10
✨deploy-image/v1alpha: add env var for storing the image name

NikhilSharmaWe:
### Description
This PR works on scaffolding an env var for storing the image informed while creating the API by `deploy-image/v1alpha` plugin, and that env var can be used to get the image where it is needed.
### Motivation

### DIFF
--- a/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
+++ b/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
@@ -109,6 +109,14 @@ func (s *apiScaffolder) Scaffold() error {
 		return fmt.Errorf("error updating main.go: %v", err)
 	}
 
+	if err := s.updateManagerWithNewEnvKey(); err != nil {
+		return fmt.Errorf("error updating config/manager/manager.yaml with env key: %v", err)
+	}
+
+	if err := s.updateManagerWithNewEnvVar(); err != nil {
+		return fmt.Errorf("error updating config/manager/manager.yaml with env var: %v", err)
+	}
+
 	if err := scaffold.Execute(
 		&samples.CRDSample{Port: s.port},
 	); err != nil {
@@ -121,6 +129,30 @@ func (s *apiScaffolder) Scaffold() error {
 		return fmt.Errorf("error creating controllers/**_controller_test.go: %v", err)
 	}
 
+	return nil
+}
+
+func (s *apiScaffolder) updateManagerWithNewEnvKey() error {
+	managerPath := filepath.Join("config", "manager", "manager.yaml")
+	err := util.ReplaceInFile(managerPath, `env:`, `env:`)
+	if err != nil {
+		err := util.InsertCode(managerPath, `name: manager`, `
+        env:`)
+		if err != nil {
+			return fmt.Errorf("error scaffolding env key in config/manager/manager.yaml")
+		}
+	}
+
+	return nil
+}
+
+func (s *apiScaffolder) updateManagerWithNewEnvVar() error {
+	managerPath := filepath.Join("config", "manager", "manager.yaml")
+	err := util.InsertCode(managerPath, `env:`,
+		fmt.Sprintf(envVarTemplate, strings.ToUpper(s.resource.Kind), s.image))
+	if err != nil {
+		return fmt.Errorf("error scaffolding env key in config/manager/manager.yaml")
+	}
 	return nil
 }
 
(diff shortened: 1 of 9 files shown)


pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
```diff
@@ -109,6 +109,14 @@ func (s *apiScaffolder) Scaffold() error {
 		return fmt.Errorf("error updating main.go: %v", err)
 	}
 
+	if err := s.updateManagerWithNewEnvKey(); err != nil {
+		return fmt.Errorf("error updating config/manager/manager.yaml with env key: %v", err)
+	}
+
+	if err := s.updateManagerWithNewEnvVar(); err != nil {
+		return fmt.Errorf("error updating config/manager/manager.yaml with env var: %v", err)
+	}
```
camilamacedo86, line 118: it seems duplicated

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -235,9 +236,14 @@ func (r *{{ .Resource.Kind }}Reconciler) doFinalizerOperationsFor{{ .Resource.Ki
...
 // deploymentFor{{ .Resource.Kind }} returns a {{ .Resource.Kind }} Deployment object
-func (r *{{ .Resource.Kind }}Reconciler) deploymentFor{{ .Resource.Kind }}({{ lower .Resource.Kind }} *{{ .Resource.ImportAlias }}.{{ .Resource.Kind }}) *appsv1.Deployment {
+func (r *{{ .Resource.Kind }}Reconciler) deploymentFor{{ .Resource.Kind }}(ctx context.Context, {{ lower .Resource.Kind }} *{{ .Resource.ImportAlias }}.{{ .Resource.Kind }}) *appsv1.Deployment {
 	ls := labelsFor{{ .Resource.Kind }}({{ lower .Resource.Kind }}.Name)
 	replicas := {{ lower .Resource.Kind }}.Spec.Size
+	log := log.FromContext(ctx)
+	image, err := imageFor{{ .Resource.Kind }}()
+	if err != nil {
+    	log.Error(err, "unable to get image for {{ .Resource.Kind }}")
+	}
```
camilamacedo86, line 246: Why do we need to get the image here?
Why it cannot be done inside of the method that creates the deployment?

Also, note that the first operation must to be check if the cr exists otherwise has no reason to do anything else

testdata/project-v4-with-deploy-image/controllers/busybox_controller.go
```diff
@@ -250,6 +256,17 @@ func labelsForBusybox(name string) map[string]string {
 	return map[string]string{"type": "busybox", "busybox_cr": name}
 }
 
+// imageForBusybox returns the image for the resources
+// belonging to the given Busybox CR name.
+func imageForBusybox() (string, error) {
```
camilamacedo86, line 261: Wdyt about we explain that it will be gathering the image from the ENV VAR which exists on the manager.go

camilamacedo86 (commented):
Great work 🥇 
Just a few nits

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
NikhilSharmaWe: No, @camilamacedo86 these are different. The former one is to scaffold the `env` key only and the second one is to scaffold the `name` and `value` of the `env` var. I did them in two different functions because we only want `env:` to be scaffolded one but the `name` and `value` can be scaffolded multiple times if multiple APIs are scaffolded.

Could you please give your thoughts.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
NikhilSharmaWe: Just confirming. So, we should add another argument in the `deploymentFor` function for the `image` and execute this code in the `Reconcile` function before we execute the `deploymentFor` function.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
NikhilSharmaWe: Ok, I am taking a look at it to make a better approach to dealing with it.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
camilamacedo86: Oh it is in the deployment method. That was my mistake, for some reason I thought that it was in the reconcile.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
camilamacedo86: Thank you for the explanation. I see now, it makes sense. Good work

camilamacedo86 (approved):
Great work 🥇 

/lgtm
/approved

PR #2835, opened 2022-07-23
:bug: deploy-image/v1alpha: fix small issues

NikhilSharmaWe:
### Description
This PR fixes some issues which are:

a) The scenario where we should only add the Recorder to be passed to the controller when the API/controller was scaffolded with the plugin
b) The scenario where we are scaffolding an extra }} in the controller by mistake in the logs 
c) Fix log typo


camilamacedo86 (approved):
Great work 🥇 

/lgtm

PR #2848, opened 2022-07-30
:seedling: fix readme

camilamacedo86:
## Description

Remove info from readme that is not valid to the latest release


camilamacedo86:
Moving forward since it is a doc fix only.

Issue #2856, opened 2022-08-03
Sample test code breaks when a second test is added


mogsie:
### What broke? What's expected?

The generated test suite contains the following code (taken from the test data) abbreviated slightly:

```go
    Context("Memcached controller test", func() {
        namespace := &corev1.Namespace{
            ObjectMeta: metav1.ObjectMeta{
                Name:      MemcachedName,
                Namespace: MemcachedName,
            },
        }

       BeforeEach(func() {
            By("Creating the Namespace to perform the tests")
            err := k8sClient.Create(ctx, namespace)
            Expect(err).To(Not(HaveOccurred()))
            ...
        })

        AfterEach(func() {
            By("Deleting the Namespace to perform the tests")
            _ = k8sClient.Delete(ctx, namespace)
            ...
        })

```
Then a single It() which is a test of some sort.

Addina second (empty) It() the above code will break, because the `namespace` object contains metadata not suitable for creating a namespace.

Adding a second It() yields the following error (`make test`)

```
    Unexpected error:
        <*errors.StatusError | 0xc0004abae0>: {
            ErrStatus: {
                TypeMeta: {Kind: "", APIVersion: ""},
                ListMeta: {
                    SelfLink: "",
                    ResourceVersion: "",
                    Continue: "",
                    RemainingItemCount: nil,
                },
                Status: "Failure",
                Message: "resourceVersion should not be set on objects to be created",
                Reason: "",
                Details: nil,
                Code: 500,
            },
        }
        resourceVersion should not be set on objects to be created
    occurred
```

Note that *fixing this* by resetting the Namespace object, only yields another problem, possibly in kubebuilder, that the old namespace isn't really being deleted.

### Reproducing this issue

1. Check out master
2. make the following change to testdata/project-v4-with-deploy-image/controllers/memcached_controller_test.go:
```patch
@@ -112,5 +112,9 @@ var _ = Describe("Memcached controller", func() {
                                return k8sClient.Get(ctx, typeNamespaceName, found)
                        }, time.Minute, time.Second).Should(Succeed())
                })
+
+               It("is funny", func() {
+                       By("doing nothing")
+               })
        })
 })
```
3. make test

Expected result: The new test passes

Actual results: The test suite fails to create the namespace

### KubeBuilder (CLI) Version

master

mogsie:
[...]

PR #2857, opened 2022-08-03
:book: Adds section on envtest limitations

mogsie:
Envtest doesn't support deletion of namespaces, and this can trip people up from time to time.  Added a description of why and how to deal with test isolation.

I added this after a member suggested I wrote a FAQ entry.  I felt this made more sense to place directly in the book where envtest is being discussed.

I have not added the code sample yet, since I wanted to get some feedback on if this was too much detail.

### DIFF
--- a/docs/book/src/reference/envtest.md
+++ b/docs/book/src/reference/envtest.md
@@ -152,5 +152,36 @@ expectedOwnerReference := v1.OwnerReference{
 Expect(deployment.ObjectMeta.OwnerReferences).To(ContainElement(expectedOwnerReference))
 ```
 
+
+<aside class="warning">
+
+<h2>Namesapce usage limitation</h2>
+
+EnvTest does not support namespace deletion. Deleting a namespace will seem to succeed, but the namespace will just be put in a Terminating state, and never actually be reclaimed. Trying to recreate the namespace will fail. This will cause your reconciler to continue reconciling any objects left behind, unless they are deleted.
+
+To overcome this limitation you can create a new namespace for each test. Even so, when one test completes (e.g. in "namespace-1") and another test starts (e.g. in "namespace-2"), the controller will still be reconciling any active objects from "namespace-1". This can be avoided by ensuring that all tests clean up after themselves as part of the test teardown.  If teardown of a namespace is difficult, it may be possible to wire the reconciler in such a way that it ignores reconcile requests that come from namespaces other than the one being tested:
+
+```go
+type MyCoolReconciler struct {
+	client.Client
+	...
+	Namespace     string  // restrict namespaces to reconcile
+}
+
+func (r *MyCoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = r.Log.WithValues("myreconciler", req.NamespacedName)
+
+	// Ignore requests for other namespaces, if specified
+	if r.Namespace != "" && req.Namespace != r.Namespace {
+		return ctrl.Result{}, nil
+	}
+```
+
+Whenever your tests create a new namespace, it can modify the value of reconciler.Namespace. The reconciler will effectively ignore the previous namespace.
+
+For further information see the issue raised in the controller-runtime [controller-runtime/issues/880](https://github.com/kubernetes-sigs/controller-runtime/issues/880) to add this support.
+
+</aside>
+
 [envtest]:https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest
 [setup-envtest]:https://pkg.go.dev/sigs.k8s.io/controller-runtime/tools/setup-envtest


docs/book/src/reference/envtest.md
```diff
@@ -152,5 +152,13 @@ expectedOwnerReference := v1.OwnerReference{
...
 ```
 
+Another limitation worth knowing about is that envtest does not support deletion of namespaces. Deleting a namespace will seem to succeed, but the namespace will just be put in Terminating state, and never actually be reclaimed. Trying to recreate the namespace will fail. For this reason, it is not advisable to try to create and delete namespaces as a mechanism to isolate tests from each other. To overcome this limitation, create a new namespace for each test.
+
+Even so, when one test completes (in "namespace-1") and another test starts (in "namespace-2"), the controller will still be reconciling any active objects from "namespace-1". This can be avoided by ensuring that all tests clean up after themselves as part of the test teardown.  If teardown of a namespace is difficult, it may be possible to wire the controller in such a way that it ignores reconcile requests that come from namespaces other than the one being tested.
+
+```go
+
+```
+
```
camilamacedo86, line 162:
```suggestion

<aside class="warning">

<h2>Namesapce usage limitation</h2>

Env Test does not support namespace deletion. Deleting a namespace will seem to succeed, but the namespace will just be put in a Terminating state, and never actually be reclaimed. Trying to recreate the namespace will fail. 

To overcome this limitation you can: 

Create a new namespace for each test. Even so, when one test completes (into i.e. "namespace-1") and another test starts (into i.e. "namespace-2"), the controller will still be reconciling any active objects from "namespace-1". This can be avoided by ensuring that all tests clean up after themselves as part of the test teardown.  If teardown of a namespace is difficult, it may be possible to wire the controller in such a way that it ignores reconcile requests that come from namespaces other than the one being tested.

</aside>
```

camilamacedo86: @mogsie, 

Thank you for the contribution 🥇. Just a few nits, see the above suggestion:

a) Could you please add in the description, closes: https://github.com/kubernetes-sigs/kubebuilder/issues/2856 (we need to look for it in C+R)
b) See that is good we add the issue in C+R to add this support
c) Also, it seems fit better in a note also we need to put after the code

docs/book/src/reference/envtest.md
```diff
@@ -152,5 +152,13 @@ expectedOwnerReference := v1.OwnerReference{
 Expect(deployment.ObjectMeta.OwnerReferences).To(ContainElement(expectedOwnerReference))
 ```
 
+Another limitation worth knowing about is that envtest does not support deletion of namespaces. Deleting a namespace will seem to succeed, but the namespace will just be put in Terminating state, and never actually be reclaimed. Trying to recreate the namespace will fail. For this reason, it is not advisable to try to create and delete namespaces as a mechanism to isolate tests from each other. To overcome this limitation, create a new namespace for each test.
+
+Even so, when one test completes (in "namespace-1") and another test starts (in "namespace-2"), the controller will still be reconciling any active objects from "namespace-1". This can be avoided by ensuring that all tests clean up after themselves as part of the test teardown.  If teardown of a namespace is difficult, it may be possible to wire the controller in such a way that it ignores reconcile requests that come from namespaces other than the one being tested.
```
camilamacedo86, line 157:
```suggestion

<aside class="warning">

<h2>Namesapce usage limitation</h2>

Env Test does not support namespace deletion. Deleting a namespace will seem to succeed, but the namespace will just be put in a Terminating state, and never actually be reclaimed. Trying to recreate the namespace will fail. 

To overcome this limitation you can: 

Create a new namespace for each test. Even so, when one test completes (into i.e. "namespace-1") and another test starts (into i.e. "namespace-2"), the controller will still be reconciling any active objects from "namespace-1". This can be avoided by ensuring that all tests clean up after themselves as part of the test teardown.  If teardown of a namespace is difficult, it may be possible to wire the controller in such a way that it ignores reconcile requests that come from namespaces other than the one being tested.

For further information see the issue raised in the controller-runtime [controller-runtime/issues/880](https://github.com/kubernetes-sigs/controller-runtime/issues/880) to add this support.

</aside>
```

docs/book/src/reference/envtest.md
```diff
@@ -152,5 +152,13 @@ expectedOwnerReference := v1.OwnerReference{
...
+
+```go
+
+```
```
camilamacedo86, line 161:
```suggestion
```

Remove ^

docs/book/src/reference/envtest.md
camilamacedo86: @mogsie,

Thank you for the contribution 🥇. Just a few nits, see the above suggestion:

a) Could you please add in the description, closes: https://github.com/kubernetes-sigs/kubebuilder/issues/2856 (we need to look for it in C+R)
b) See that is good we add the issue in C+R to add this support
c) Also, it seems fit better in a note also we need to put after the code

camilamacedo86 (approved):
Great work 🥇

camilamacedo86 (approved):
/lgtm 

I think we can move forward with the note.
We can improve/shape in follow-up always.

PR #2858, opened 2022-08-04
:sparkles: Experimental practice to scaffold dashboard manifest for custom metrics

Kavinjsir:
### Description
This is the experimental practice to see how possible if we can bring dashboards for custom metrics through the Grafana Plugin.

The user can add the desired metrics in `grafana/custom-metrics/config.yaml`:
```yaml
---
customMetrics: []
  # - metric: Raw custom metric (required)
  #   type:   Metric type: counter/gauge/histogram (required)
  #   expr:   Prom_ql for the metric (optional)

```
(If you execute the plugin for the first time(either by `init` or `edit`), the above config template will be automatically generated.)

 The plugin triggered by `edit` subcommand will read the `config.yaml` and generate `grafana/custom-metrics/custom-metrics-dashboard.json`, which can be directly import in the Grafana Web UI.

#### Story 1
Users get the config entry when initialize the project or introduce the `grafana-plugin` for the first time:
```shell
# Initialize a project with the plugin enabled
kubebuilder init --plugins=grafana.kubebuilder.io/v1-alpha

# Introduce the Grafana Plugin to an existing project
kubebuilder edit --plugins=grafana.kubebuilder.io/v1-alpha
```

A nested `custom-metrics/config.yaml` will be generated under `grafana/`.

#### Story 2
Users configure `grafana/custom-metrics/config.yaml` and run `edit` subcommand to get Grafana manifest:
```shell
kubebuilder edit --plugins=grafana.kubebuilder.io/v1-alpha
```

The `grafana/custom-metrics/config.yaml` will be validated and then the manifest will be generated: `grafana/custom-metrics/custom-metrics-dashboard.json`.

### Sample

1. Specify custom metrics in config.yaml
```yaml
---
customMetrics:
  - metric: memcached_operator_reconcile_total
    type: counter
  - metric: memcached_operator_reconcile_time_seconds_bucket
    type: histogram
```
2. Run `kubebuilder edit --plugins=grafana.kubebuilder.io/v1-alpha1` to generate the manifest
3. Import the manifest into Grafana Web UI:

newpre

### DIFF
--- a/docs/book/src/plugins/grafana-v1-alpha.md
+++ b/docs/book/src/plugins/grafana-v1-alpha.md
@@ -132,6 +132,48 @@ See an example of how to use the plugin in your project:
   - Per-second rate of retries handled by workqueue
 - Sample: <img width="914" src="https://user-images.githubusercontent.com/18136486/180360101-411c81e9-d54e-4b21-bbb0-e3f94fcf48cb.png">
 
+### Visualize Custom Metrics
+
+The Grafana plugin supports scaffolding manifests for custom metrics.
+
+#### Generate Config Template
+
+When the plugin is triggered for the first time, `grafana/custom-metrics/config.yaml` is generated.
+
+```yaml
+---
+customMetrics:
+#  - metric: # Raw custom metric (required)
+#    type:   # Metric type: counter/gauge/histogram (required)
+#    expr:   # Prom_ql for the metric (optional)
+```
+
+#### Add Custom Metrics to Config
+
+You can enter multiple custom metrics in the file. For each element, you need to specify the `metric` and its `type`.
+The Grafana plugin can automatically generate `expr` for visualization.
+Alternatively, you can provide `expr` and the plugin will use the specified one directly.
+
+```yaml
+---
+customMetrics:
+  - metric: memcached_operator_reconcile_total # Raw custom metric (required)
+    type: counter # Metric type: counter/gauge/histogram (required)
+  - metric: memcached_operator_reconcile_time_seconds_bucket
+    type: histogram
+```
+
+#### Scaffold Manifest
+
+Once `config.yaml` is configured, you can run `kubebuilder edit --plugins grafana.kubebuilder.io/v1-alpha` again.
+This time, the plugin will generate `grafana/custom-metrics/custom-metrics-dashboard.json`, which can be imported to Grafana UI.
+
+#### Show case:
+
+See an example of how to visualize your custom metrics:
+
+![output2](https://user-images.githubusercontent.com/18136486/186933170-d2e0de71-e079-4d1b-906a-99a549d66ebf.gif)
+
 ## Subcommands
 
 The Grafana plugin implements the following subcommands:
--- a/pkg/plugin/util/util.go
+++ b/pkg/plugin/util/util.go
@@ -242,3 +242,14 @@ func ReplaceRegexInFile(path, match, replace string) error {
 	}
 	return nil
 }
+
+// HasFileContentWith check if given `text` can be found in file
+func HasFileContentWith(path, text string) (bool, error) {
+	// nolint:gosec
+	contents, err := ioutil.ReadFile(path)
+	if err != nil {
+		return false, err
+	}
+
+	return strings.Contains(string(contents), text), nil
+}
--- a/pkg/plugins/optional/grafana/v1alpha/scaffolds/edit.go
+++ b/pkg/plugins/optional/grafana/v1alpha/scaffolds/edit.go
@@ -18,13 +18,21 @@ package scaffolds
 
 import (
 	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
 
 	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugins"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/optional/grafana/v1alpha/scaffolds/internal/templates"
+
+	"sigs.k8s.io/yaml"
 )
 
 var _ plugins.Scaffolder = &editScaffolder{}
+
+const configFilePath = "grafana/custom-metrics/config.yaml"
 
 type editScaffolder struct {
 	// fs is the filesystem that will be used by the scaffolder
@@ -41,6 +49,100 @@ func (s *editScaffolder) InjectFS(fs machinery.Filesystem) {
 	s.fs = fs
 }
 
+func fileExist(configFilePath string) bool {
+	if _, err := os.Stat(configFilePath); os.IsNotExist(err) {
+		return false
+	}
+	return true
+}
+
+func loadConfig(configPath string) ([]templates.CustomMetricItem, error) {
+	if !fileExist(configPath) {
+		return nil, nil
+	}
+
+	// nolint:gosec
+	f, err := os.Open(configPath)
+	if err != nil {
+		return nil, fmt.Errorf("error loading plugin config: %w", err)
+	}
+
+	items, err := configReader(f)
+
+	if err := f.Close(); err != nil {
+		return nil, fmt.Errorf("could not close config.yaml: %w", err)
+	}
+
+	return items, err
+}
+
+func configReader(reader io.Reader) ([]templates.CustomMetricItem, error) {
+	yamlFile, err := ioutil.ReadAll(reader)
+	if err != nil {
+		return nil, err
+	}
+
+	config := templates.CustomMetricsConfig{}
+
+	err = yaml.Unmarshal(yamlFile, &config)
+	if err != nil {
+		return nil, err
+	}
+
+	validatedMetricItems := validateCustomMetricItems(config.CustomMetrics)
+
+	return validatedMetricItems, nil
+}
+
+func validateCustomMetricItems(rawItems []templates.CustomMetricItem) []templates.CustomMetricItem {
+	// 1. Filter items of missing `Metric` or `Type`
+	filterResult := []templates.CustomMetricItem{}
+	for _, item := range rawItems {
+		if hasFields(item) {
+			filterResult = append(filterResult, item)
+		}
+	}
+
+	// 2. Fill Expr if missing
+	validatedItems := make([]templates.CustomMetricItem, len(filterResult))
+	for i, item := range filterResult {
+		validatedItems[i] = fillMissingExpr(item)
+	}
+
+	return validatedItems
+}
+
+func hasFields(item templates.CustomMetricItem) bool {
+	// If `Expr` exists, return true
+	if item.Expr != "" {
+		return true
+	}
+
+	// If `Metric` & valid `Type` exists, return true
+	metricType := strings.ToLower(item.Type)
+	if item.Metric != "" && (metricType == "counter" || metricType == "gauge" || metricType == "histogram") {
+		return true
+	}
+
+	return false
+}
+
+// TODO: Prom_ql exprs can improved to be more pratical and applicable
+func fillMissingExpr(item templates.CustomMetricItem) templates.CustomMetricItem {
+	if item.Expr == "" {
+		switch strings.ToLower(item.Type) {
+		case "counter":
+			item.Expr = "sum(rate(" + item.Metric + `{job=\"$job\", namespace=\"$namespace\"}[5m])) by (instance, pod)`
+		case "histogram":
+			// nolint: lll
+			item.Expr = "histogram_quantile(0.90, sum by(instance, le) (rate(" + item.Metric + `{job=\"$job\", namespace=\"$namespace\"}[5m])))`
+		default: // gauge
+			item.Expr = item.Metric
+		}
+	}
+	return item
+}
+
 // Scaffold implements cmdutil.Scaffolder
 func (s *editScaffolder) Scaffold() error {
 	fmt.Println("Generating Grafana manifests to visualize controller status...")
@@ -48,8 +150,20 @@ func (s *editScaffolder) Scaffold() error {
 	// Initialize the machinery.Scaffold that will write the files to disk
 	scaffold := machinery.NewScaffold(s.fs)
 
-	return scaffold.Execute(
+	configPath := string(configFilePath)
+
+	var templatesBuilder = []machinery.Builder{
 		&templates.RuntimeManifest{},
 		&templates.ResourcesManifest{},
-	)
+		&templates.CustomMetricsConfigManifest{ConfigPath: configPath},
+	}
+
+	configItems, err := loadConfig(configPath)
+	if err == nil && len(configItems) > 0 {
+		templatesBuilder = append(templatesBuilder, &templates.CustomMetricsDashManifest{Items: configItems})
+	} else if err != nil {
+		fmt.Fprintf(os.Stderr, "Error on scaffolding manifest for custom metris:\n%v", err)
+	}
+
+	return scaffold.Execute(templatesBuilder...)
 }
--- a/pkg/plugins/optional/grafana/v1alpha/scaffolds/init.go
+++ b/pkg/plugins/optional/grafana/v1alpha/scaffolds/init.go
@@ -51,5 +51,6 @@ func (s *initScaffolder) Scaffold() error {
 	return scaffold.Execute(
 		&templates.RuntimeManifest{},
 		&templates.ResourcesManifest{},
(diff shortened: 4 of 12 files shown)


pkg/plugins/optional/grafana/v1alpha/scaffolds/internal/templates/custom.go
```diff
@@ -0,0 +1,51 @@
...
+
+	f.TemplateBody = customMetricsConfigTemplate
+
+	f.IfExistsAction = machinery.SkipFile
+
+	return nil
+}
+
+// nolint: lll
+const customMetricsConfigTemplate = `---
```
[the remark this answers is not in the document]
varshaprasad96: Why do we need to scaffold the configuration file?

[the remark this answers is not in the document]
varshaprasad96: I see that we need to scaffold the config file. Please ignore my previous comment then. If we are scaffolding, we need to make sure that when the config file is empty then we don't do anything.

pkg/plugins/optional/grafana/v1alpha/scaffolds/edit.go
```diff
@@ -17,11 +17,16 @@ limitations under the License.
...
+	"errors"
 	"fmt"
+	"io/ioutil"
+	"os"
 
 	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugins"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/optional/grafana/v1alpha/scaffolds/internal/templates"
+
+	yaml "gopkg.in/yaml.v2"
```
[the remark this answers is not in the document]
Kavinjsir: Yeah, this sounds better! Thx for letting me know!

pkg/plugins/optional/grafana/v1alpha/scaffolds/edit.go
```diff
@@ -41,15 +46,70 @@ func (s *editScaffolder) InjectFS(fs machinery.Filesystem) {
...
+
+func isConfigValid() bool {
+	// TODO: Is it good to hard code the config file path here?
+	configPath := "grafana/custom-metrics/config.yaml"
+
+	if !isFileExist(configPath) {
+		return false
+	}
+
+	yamlFile, err := ioutil.ReadFile(configPath)
```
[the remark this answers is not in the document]
Kavinjsir: 👍🏼  Have updated the code based on the reference and some previous discussion.

Kavinjsir:
Since `grafana/custom-metrics/config.yaml` is exposed to customize the panels, I'm also looking for methods to test the plugin upon a sample config.

Right now the integration test will generate test data from 0 to 1. I'm wondering **how we may insert a config template** such as:
```yaml
---
customMetrics:
  - metric: memcached_operator_reconcile_total
    type: counter
  - metric: memcached_operator_reconcile_time_seconds_bucket
    type: histogram
```
And run `kubebuilder edit ...` follow by that in the integration test?

cc: @varshaprasad96 @camilamacedo86 @rashmigottipati

pkg/plugins/optional/grafana/v1alpha/scaffolds/edit.go
```diff
@@ -18,10 +18,16 @@ package scaffolds
...
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
 
 	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugins"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/optional/grafana/v1alpha/scaffolds/internal/templates"
+
+	yaml "sigs.k8s.io/yaml"
```
varshaprasad96, line 30: Nit: We don't need an alias name here. `"sigs.k8s.io/yaml"` should be enough, since package name and alias are the same.

pkg/plugins/optional/grafana/v1alpha/scaffolds/edit.go
```diff
@@ -41,15 +47,122 @@ func (s *editScaffolder) InjectFS(fs machinery.Filesystem) {
...
+func fileExist(configFilePath string) bool {
+	if _, err := os.Stat(configFilePath); os.IsNotExist(err) {
+		return false
+	}
+	return true
+}
+
+func loadConfig() ([]templates.CustomMetricItem, error) {
+	// TODO: Is it good to hard code the config file path here?
+	configPath := "grafana/custom-metrics/config.yaml"
```
varshaprasad96, line 59: For now, let's make this constant. We can look into allowing users to specify this later. Instead of declaring it locally, let's make this a global variable:
```
const configFilePath := "grafana/custom-metrics/config.yaml"
```

pkg/plugins/optional/grafana/v1alpha/scaffolds/edit.go
```diff
@@ -41,15 +47,122 @@ func (s *editScaffolder) InjectFS(fs machinery.Filesystem) {
...
+	// If `Metric` & valid `Type` exists, return true
+	metricType := strings.ToLower(item.Type)
+	if item.Metric != "" && (metricType == "counter" || metricType == "gauge" || metricType == "histogram") {
+		return true
+	}
+
+	return false
+}
+
+func fillMissingExpr(item templates.CustomMetricItem) templates.CustomMetricItem {
```
varshaprasad96, line 130: so this func adds a default expression when user does not provide it?
If so, is it compatible with all the three kinds of prometheus metrics like gauge, counter etc.

pkg/plugins/optional/grafana/v1alpha/scaffolds/edit.go
```diff
@@ -41,15 +47,122 @@ func (s *editScaffolder) InjectFS(fs machinery.Filesystem) {
...
 
 	// Initialize the machinery.Scaffold that will write the files to disk
 	scaffold := machinery.NewScaffold(s.fs)
 
-	return scaffold.Execute(
+	var templatesBuilder = []machinery.Builder{
 		&templates.RuntimeManifest{},
 		&templates.ResourcesManifest{},
-	)
+		&templates.CustomMetricsConfigManifest{},
```
varshaprasad96, line 157: Actually better option would be to take configPath as an input from here, as we do with other variables. We can let 
```go
type CustomMetricsConfigManifest {
      ConfigPath string
}
```

cc: @camilamacedo86

varshaprasad96 (commented):
Good work @Kavinjsir ! Just a few nits.

pkg/plugin/util/util.go
```diff
@@ -242,3 +242,29 @@ func ReplaceRegexInFile(path, match, replace string) error {
...
+		return err
+	}
+
+	err = ioutil.WriteFile(path, []byte(content), info.Mode())
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
camilamacedo86, line 259: ihmo we do not need both methods, 
See that we can use instead: 
https://github.com/kubernetes-sigs/kubebuilder/blob/3ac65ec5a1f429f40f0fa81acc8a7d69a9a06bbc/pkg/plugin/util/util.go#L195-L216

camilamacedo86: c/c @varshaprasad96

pkg/plugins/optional/grafana/v1alpha/scaffolds/edit.go
```diff
@@ -41,15 +49,121 @@ func (s *editScaffolder) InjectFS(fs machinery.Filesystem) {
 	s.fs = fs
 }
 
+func fileExist(configFilePath string) bool {
```
camilamacedo86, line 52:
```suggestion
func hasFileWith(configFilePath string) bool {
```
methods that return boolean usually are prefixed with has or is

pkg/plugins/optional/grafana/v1alpha/scaffolds/edit.go
```diff
@@ -41,15 +49,121 @@ func (s *editScaffolder) InjectFS(fs machinery.Filesystem) {
...
+	// 2. Fill Expr if missing
+	validatedItems := make([]templates.CustomMetricItem, len(filterResult))
+	for i, item := range filterResult {
+		validatedItems[i] = fillMissingExpr(item)
+	}
+
+	return validatedItems
+}
+
+func fieldsExist(item templates.CustomMetricItem) bool {
```
camilamacedo86, line 115:
```suggestion
func hasFields(item templates.CustomMetricItem) bool {
```

pkg/plugins/optional/grafana/v1alpha/scaffolds/init.go
```diff
@@ -51,5 +51,6 @@ func (s *initScaffolder) Scaffold() error {
 	return scaffold.Execute(
 		&templates.RuntimeManifest{},
 		&templates.ResourcesManifest{},
+		&templates.CustomMetricsConfigManifest{ConfigPath: string(ConfigFilePath)},
```
camilamacedo86, line 54: I think we do not need to pass the path because it will be fixed always
So, we can have a const with the default path inside of the template or am I missing something?

pkg/plugins/optional/grafana/v1alpha/scaffolds/internal/templates/custom.go
```diff
@@ -0,0 +1,48 @@
...
+customMetrics:
+#  - metric: # Raw custom metric (required)
+#    type:   # Metric type: counter/gauge/histogram (required)
+#    expr:   # Prom_ql for the metric (optional)
```
camilamacedo86, line 47: What about we add examples about how to fill the file commented to show to the users?

test/e2e/grafana/e2e_suite_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+// Run e2e tests using the Ginkgo runner.
+func TestE2E(t *testing.T) {
+	RegisterFailHandler(Fail)
+	fmt.Fprintf(GinkgoWriter, "Starting kubebuilder suite\n")
```
camilamacedo86, line 30:
```suggestion
	fmt.Fprintf(GinkgoWriter, "Starting grafana plugin kubebuilder suite\n")
```

test/e2e/grafana/e2e_suite_test.go
```diff
@@ -0,0 +1,32 @@
...
+	RunSpecs(t, "Kubebuilder e2e suite")
```
camilamacedo86, line 31:
```suggestion
	RunSpecs(t, "Kubebuilder grafana plugin e2e suite")
```

test/e2e/grafana/generate_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	By("uncomment grafana/custom-metrics/config.yaml to verify config template")
+	ExpectWithOffset(1, pluginutil.UncommentCode(
+		filepath.Join(kbc.Dir, "grafana", "custom-metrics", "config.yaml"),
+		"#  - metric: # Raw custom metric (required)", "#")).To(Succeed())
+	ExpectWithOffset(1, pluginutil.UncommentCode(
+		filepath.Join(kbc.Dir, "grafana", "custom-metrics", "config.yaml"),
+		"#    type:   # Metric type: counter/gauge/histogram (required)", "#")).To(Succeed())
+
+	ExpectWithOffset(1, pluginutil.UncommentCode(
```
camilamacedo86, line 77: If we are uncommented and then replace the file we do not need to uncomment.
IMO here we can change that to use ReplaceInFile and you will pass a const which has the original content to be replaced for new content, see an example: https://github.com/kubernetes-sigs/kubebuilder/blob/20828c7e87e36c005f77e9c57ed9a84371dc65d8/test/e2e/v3/generate_test.go#L345-L359

pkg/plugin/util/util.go
```diff
@@ -242,3 +242,29 @@ func ReplaceRegexInFile(path, match, replace string) error {
...
+
+// FileContains check if given `text` can be found in file
+func FileContains(path, text string) (bool, error) {
```
camilamacedo86, line 262:
```suggestion
func HasFileContentWith(path, text string) (bool, error) {
```

docs/book/src/plugins/grafana-v1-alpha.md
```diff
@@ -132,6 +132,39 @@ See an example of how to use the plugin in your project:
...
+When the plugin is triggered for the first time, `grafana/custom-metrics/config.yaml` is generated.
+
+```yaml
+---
+customMetrics:
+#  - metric: # Raw custom metric (required)
+#    type:   # Metric type: counter/gauge/histogram (required)
+#    expr:   # Prom_ql for the metric (optional)
+```
+
```
camilamacedo86, line 150: Could we add an example here?

pkg/plugins/optional/grafana/v1alpha/scaffolds/edit.go
```diff
@@ -18,14 +18,22 @@ package scaffolds
...
 	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugins"
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/optional/grafana/v1alpha/scaffolds/internal/templates"
+
+	"sigs.k8s.io/yaml"
 )
 
 var _ plugins.Scaffolder = &editScaffolder{}
 
+const ConfigFilePath = "grafana/custom-metrics/config.yaml"
```
camilamacedo86, line 35: do we use it outside, if not then we do not need the capital prefix.
If we do that we are exporting the const to be used in another places
```suggestion
const configFilePath = "grafana/custom-metrics/config.yaml"
```

pkg/plugins/optional/grafana/v1alpha/scaffolds/init.go
varshaprasad96: @camilamacedo86 this path is fixed for now. But I instead on passing it here so that we can make it configurable later. That way, if we want to introduce a feature where the user would like to scaffold config file anywhere else in the project, we can read and accept it. Wdyt?

camilamacedo86 (approved):
Great work 🥇

Kavinjsir:
See that the two test cases remain erroring. Would you like me to fix it by additional coding?
@varshaprasad96 @camilamacedo86

test/e2e/grafana/generate_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+var _ = Describe("kubebuilder", func() {
+	Context("plugin grafana/v1-alpha", func() {
+		var (
+			kbc *utils.TestContext
+		)
+
+		BeforeEach(func() {
+			var err error
+			kbc, err = utils.NewTestContext(pluginutil.KubebuilderBinName, "GO111MODULE=on")
```
Kavinjsir, line 43: Test error may come from here. Since such `NewTestContext` would call `kubectl` that assumes an existing `kind` cluster.

test/e2e/plugin.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
camilamacedo86, line 1: @Kavinjsir we should not add a new file here. 
In this case, we need only in the Kubebuilder Makefile to call the test.

We would only need to execute this test once. However, to make simple let's call it with the others and a TODO here to sort it out after. 

That means:
a) Remove this file plugin.sh
b) Add the new test in : https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/setup.sh#L65
c) Revert the changes made in the Kubebuilder Makefile

camilamacedo86: c/c @varshaprasad96 

/hold

Kavinjsir: Sounds good to me. I've updated the code based on the steps and it looks work well.

camilamacedo86 (commented):
/hold cancel

pkg/plugins/optional/grafana/v1alpha/scaffolds/internal/templates/custom_metrics.go
```diff
@@ -0,0 +1,296 @@
...
+                "color": "green",
+                "value": null
+              },
+              {
+                "color": "red",
+                "value": 80
+              }
+            ]
+          },
+          "unit": "s"
```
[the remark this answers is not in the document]
jkremser: configurable units would be much appreciated!

PR #2859, opened 2022-08-05
✨deploy-image/v1-alpha: add a todo comment for users stating envtest limitations

NikhilSharmaWe:
### Description
Added a todo comment for users in controller_test.go file for deploy-image/v1-alpha plugin, highlighting the limitations of the envtest.
### Motivation
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/2856

### DIFF
--- a/testdata/project-v3-with-deploy-image/controllers/busybox_controller_test.go
+++ b/testdata/project-v3-with-deploy-image/controllers/busybox_controller_test.go
@@ -60,6 +60,8 @@ var _ = Describe("Busybox controller", func() {
 		})
 
 		AfterEach(func() {
+			// TODO(user): Attention if you improve this code by adding other context test you MUST
+			// be aware of the current delete namespace limitations. More info: https://book.kubebuilder.io/reference/envtest.html#testing-considerations
 			By("Deleting the Namespace to perform the tests")
 			_ = k8sClient.Delete(ctx, namespace)
 
--- a/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller-test.go
+++ b/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller-test.go
@@ -107,6 +107,8 @@ var _ = Describe("{{ .Resource.Kind }} controller", func() {
 		})
 
 		AfterEach(func() {
+			// TODO(user): Attention if you improve this code by adding other context test you MUST
+			// be aware of the current delete namespace limitations. More info: https://book.kubebuilder.io/reference/envtest.html#testing-considerations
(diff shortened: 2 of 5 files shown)


camilamacedo86 (approved):
That is great for me

@mogsie 

WDYT ?

Could you move here?

mogsie:
I'm sorry, when you say "move" do you mean move my code changes, or actually propose changes to the scaffold that _fixes_ these issues (by making it possible to use several namespaces)?

The TODO helps of course. It's important that users are aware of the limitation. :+1:

But even better would be code that supported several tests.

testdata/project-v3-with-deploy-image/controllers/busybox_controller_test.go
```diff
@@ -60,6 +60,8 @@ var _ = Describe("Busybox controller", func() {
 		})
 
 		AfterEach(func() {
+			// TODO(user): Attention if you improve this code by adding other context test you MUST
+			// be aware of the current delete namespace limitations. More info: https://book.kubebuilder.io/reference/envtest.html#testing-considerations
 			By("Deleting the Namespace to perform the tests")
 			_ = k8sClient.Delete(ctx, namespace)
```
camilamacedo86, line 67: HI @mogsie,

> I'm sorry, when you say "move" do you mean move my code changes, or actually propose changes to the scaffold that fixes these issues (by making it possible to use several namespaces)?

I could not follow this one. 

> The TODO helps of course. It's important that users are aware of the limitation. 👍
> But even better would be code that supported several tests.

Will we have in all case scenarios many tests?
OR in many cases to unit test the controller we might have only one test that just checks the reconciliation.

So, I think we can add the TODO here and ensure that it is still simple. 
However, we can also add an example in the docs WDYT? Also, we probably need to vote on the C+R issue and see if we can address this one. The ideal solution for me shows we fix it in the C+R and we just let people know about that now. WDYT? Could we agree?

mogsie: I think the TODO approach is best.  It allows the developers to make the choice:

* Either I keep within the single namespace, and clean up nicely after myself whenever I need to, and avoid problems like that
* Or I decide to make new namespaces for each new test, with the additional extra work needed.

Personally, I like having the ability to write integration tests and not have to worry about cleaning up, especially while developing, when I may not have completed the _clean-up_ portion of the controller.  In those cases, cleaning up is difficult.  But it's not always needed, so I think the TODO comment is the best approach.

camilamacedo86: That is terrific. So, let's move forward with this one since we have your : 

/lgtm

PR #2862, opened 2022-08-07
🐛 :  doc: fix multiversion tutorial sample error: metadata.annotations: Too long: must have at most 262144 bytes

### LINKED ISSUE #2556: metadata.annotations: Too long: must have at most 262144 bytes

clone this project on the machine, an go to dir docs/book/src/multiversion-tutorial/testdata/project . `make install deploy`; then this error appear
```
 make install deploy
/var/tmp/kubebuilder-master/docs/book/src/multiversion-tutorial/testdata/project/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
/var/tmp/kubebuilder-master/docs/book/src/multiversion-tutorial/testdata/project/bin/kustomize build config/crd | kubectl apply -f -
The CustomResourceDefinition "cronjobs.batch.tutorial.kubebuilder.io" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
make: *** [install] Error 1
```

### Reproducing this issue

_No response_

### KubeBuilder (CLI) Version

3.3.0

### PROJECT version

3

### Plugin versions

_No response_

### Other versions

k8s:   v1.21.9
go:    1.16.2
kustomize: v4.5.2
centos: 3.10.0-693.el7.x86_64

### Extra Labels

_No response_

laxmikantbpandhare:
This will fix #2556 

Before fixing this issue,  it is failing like below:

```
laxmikantbhaskarpandhare@lpandhar-mac project % make install
/Users/laxmikantbhaskarpandhare/go/src/github.com/kubernetes-sigs/kubebuilder/docs/book/src/multiversion-tutorial/testdata/project/bin/kustomize build config/crd | kubectl apply -f -
The CustomResourceDefinition "cronjobs.batch.tutorial.kubebuilder.io" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
make: *** [install] Error 1
```

After fix, it worked.

```
laxmikantbhaskarpandhare@lpandhar-mac project % make install
test -s /Users/laxmikantbhaskarpandhare/go/src/github.com/kubernetes-sigs/kubebuilder/docs/book/src/multiversion-tutorial/testdata/project/bin/controller-gen || GOBIN=/Users/laxmikantbhaskarpandhare/go/src/github.com/kubernetes-sigs/kubebuilder/docs/book/src/multiversion-tutorial/testdata/project/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0
/Users/laxmikantbhaskarpandhare/go/src/github.com/kubernetes-sigs/kubebuilder/docs/book/src/multiversion-tutorial/testdata/project/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
/Users/laxmikantbhaskarpandhare/go/src/github.com/kubernetes-sigs/kubebuilder/docs/book/src/multiversion-tutorial/testdata/project/bin/kustomize build config/crd | kubectl replace --force -f -
customresourcedefinition.apiextensions.k8s.io/cronjobs.batch.tutorial.kubebuilder.io replaced
```


laxmikantbpandhare:
Hi @camilamacedo86 -

I was able to recreate the issue. There are two ways through which we can fix this issue.

1. As mentioned in the issue,  modify the Makefile manifests target. change crd => crd:maxDescLen=0;
2. Change `kubectl apply to `kubectl replace --force` as mentioned [here](https://github.com/bugfest/tor-controller/pull/18#issuecomment-1179899925)

Please take a look and let me know your thoughts.

everettraven (commented):
@laxmikantbpandhare I think this looks good but I have one concern with using `kubectl replace` instead of `kubectl apply`. 

My understanding with `kubectl apply` is that it will create the resource if it does not exist or update it if it does exist. This seems like the functionality we would want. 

I could be wrong, but my understanding of `kubectl replace` is that it will only replace something that already exists. If you run `make install` or `make deploy` with this change on a fresh KinD cluster does it still work as expected?

laxmikantbpandhare:
@camilamacedo86 @everettraven - Updated PR. `crd:maxDescLen=0` - It will skip the description section while creating an annotation. This change will avoid annotations too long errors.

Please take a look once.

everettraven (commented):
I think this looks good and will fix the case in the issue, but I wonder if we should also apply this to the scaffolds? 

Is there a reason why this occurs in this specific tutorial and not on a regular basis?

Thoughts @camilamacedo86 ?

laxmikantbpandhare:
> I think this looks good and will fix the case in the issue, but I wonder if we should also apply this to the scaffolds?
> 
> Is there a reason why this occurs in this specific tutorial and not on a regular basis?
> 
> Thoughts @camilamacedo86 ?

@everettraven - this is the file that is causing this issue - https://raw.githubusercontent.com/kubernetes-sigs/kubebuilder/master/docs/book/src/multiversion-tutorial/testdata/project/config/crd/bases/batch.tutorial.kubebuilder.io_cronjobs.yaml

If you see the above file then it is a large file and has too many `description` sections. I don't think we need to scaffold it out just for only one file. WDYT?

everettraven:
>this is the file that is causing this issue - https://raw.githubusercontent.com/kubernetes-sigs/kubebuilder/master/docs/book/src/multiversion-tutorial/testdata/project/config/crd/bases/batch.tutorial.kubebuilder.io_cronjobs.yaml
> 
> If you see the above file then it is a large file and has too many `description` sections. I don't think we need to scaffold it out just for only one file. WDYT?

@laxmikantbpandhare Ah, thanks for that information! I went through the tutorial and having the extra context clears up the confusion I was having. This change looks good to me so I will go ahead and approve the PR.

everettraven (approved):

laxmikantbpandhare:
@camilamacedo86 could you please take a look.

docs/book/src/multiversion-tutorial/testdata/project/Makefile
```diff
@@ -38,9 +38,10 @@ help: ## Display this help.
 
 ##@ Development
 
+## crd:maxDescLen=0 - It will skip the description section while creating an annotation. This change will avoid annotations too long errors.
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) rbac:roleName=manager-role crd:maxDescLen=0 webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```
camilamacedo86, line 44: @laxmikantbpandhare I am OK with the workaround. Just a few nits:
a) Could we change the title so that makes it clear that it is about the docs/samples. Example:

:doc: fix multiversion tutorial sample error: metadata.annotations: Too long: must have at most 262144 bytes 

b) Could you please run `make manifests` in the /docs/book/src/multiversion-tutorial/testdata/project so that we can also merge the changes in the CRD? ( After we add this option it removes all descriptions )

c) Could we move the comment from to be on top of the line which has the `crd:maxDescLen=0 ` and use only one # since it is a comment and not a section? Therefore, wdyt about something like:

```sh
# Note that the option maxDescLen=0 was added in the default scaffold in order to sort out the issue
# Too long: must have at most 262144 bytes. By using kubectl apply to create / update resources an annotation 
# is created by K8s API to store the latest version of the resource ( kubectl.kubernetes.io/last-applied-configuration). 
# However, it has a size limit and if the CRD is too big with so many long descriptions as this one it will cause the failure. 
$(CONTROLLER_GEN) rbac:roleName=manager-role crd:maxDescLen=0 webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```

laxmikantbpandhare: @camilamacedo86 - I agree with you. I updated the PR.

camilamacedo86 (approved):
/lgtm

Issue #2600, opened 2022-04-04
Phase 2 Plugins - Meta Issue


rashmigottipati:
### What do you want to happen?

The propose of this issue is to track Phase 2 Plugins effort and centralize all related discussions here so that the goals and next steps are aligned, and also share with the community on features to expect from Phase 2 Plugins Implementation. 

- [x] [Issue 1378](https://github.com/kubernetes-sigs/kubebuilder/issues/1378)
This is the original feature request created that discusses the main requirements for Phase 2 plugins, which is to have a plugin architecture that supports:
1) Plugins as separate binaries that can be discovered by the Kubebuilder CLI binary.
2) Plugins that are language-agnostic, i.e. they can be written in any language.

- [x] [Phase 2 Design Proposal](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-2.md)
This proposal discusses how to discover and use external plugins, also referred to as out-of-tree plugins (which can be implemented in any language). 

- [x] [Phase 2 Plugins implementation](https://github.com/kubernetes-sigs/kubebuilder/pull/2338)
This PR handles discovery and chaining of external plugins. 

Another goal of this this issue is to track the remaining work that needs to be completed for Phase 2 Plugins:
- [x] documentation; we need to update [Plugin Section](https://book.kubebuilder.io/plugins/plugins.html) on how to create and run an external plugin.

- [x] Provide an example of an external plugin for users. This example could reside under the testdata dir. 

- [x] e2e tests discovering and running the sample plugin.

camilamacedo86:
For we accomplish the goal: 

a) Add samples
We have the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/2810
Also, @rashmigottipati would be great if we could add the python one as well used by you to demo. So we can demonstrate that it can work with any language

b) Add the e2e test to build the binary from the samples added and check that the kubebuilder CLI is capable of to properly using them by creating a project with the external plugin and verifying the result ( very simple we do not need to check all details. )

c) Add the docs and a tutorial to let people know how to create their own plugins that can be consumed by
d) update the plugins doc to let the users know that it is implemented, see: 
- https://book.kubebuilder.io/plugins/plugins.html
- https://book.kubebuilder.io/plugins/creating-plugins.html#future-vision-for-kubebuilder-plugins

Could we update the description of this issue as well accordingly? 

c/c @everettraven

[...]

PR #2863, opened 2022-08-10
:bug: "Application Support" whitespace for mac plugins

jakobmoellersap:
This mini-PR simply fixes the Plugins Root Directory for darwin hosts for external Plugins.

Without this change the Plugin Discovery will always look in a non-existing Library Directory, causing the Lookup to permanently fail. A suggestion here would be to introduce the "Lookup Plugins" command that was discussed previously in the Discussion around https://github.com/kubernetes-sigs/kubebuilder/issues/2600 to also make this debuggable without having to change log messages or introducing a cli debug flag

### DIFF
--- a/pkg/cli/options.go
+++ b/pkg/cli/options.go
@@ -176,7 +176,7 @@ func getPluginsRoot(host string) (pluginsRoot string, err error) {
 	switch host {
 	case "darwin":
 		logrus.Debugf("Detected host is macOS.")
-		pluginsRoot = filepath.Join("Library", "ApplicationSupport", "kubebuilder", "plugins")
+		pluginsRoot = filepath.Join("Library", "Application Support", "kubebuilder", "plugins")
 	case "linux":
 		logrus.Debugf("Detected host is Linux.")
 		pluginsRoot = filepath.Join(".config", "kubebuilder", "plugins")


camilamacedo86 (approved):
@rashmigottipati @everettraven could you give a hand on this one?

jakobmoellersap:
Checked both failing tests and they seem to be timeout related, can I just rerun them?

jakobmoellersap:
golanci-lint still failing with exit 137

Exit code 137 is triggered when a pod or a container within your Kubernetes environment exceeds the amount of memory that they’re assigned

everettraven (approved):
Looks good to me. 

Seems like the error with the golangci-lint is memory related. For some reason it is running out of memory. I wonder if there is a need to bump memory limits in the GitHub Action itself.

jakobmoellersap:
@everettraven I checked your GH Actions and I think we cannot simply increase the memory limit through a setting as can be observed in https://github.com/golangci/golangci-lint-action#how-to-use, since it is a machine-limitation in the action. The only way I know to circuvment this is to adjust the GC as can be seen in https://golangci-lint.run/usage/performance/#memory-usage to increase Garbage Collection Cycles. This reduces memory footprint at the cost of heavily increased lint runtime (may not be favorable for us)

What do you I think I can do here to make this PR go in @camilamacedo86 @everettraven ?

everettraven:
@jakobmoellersap I think modifying that `GOGC` environment variable to trigger garbage collection more frequently would suffice. Looking at the performance link you sent, it seems like setting the `GOGC` environment variable to a value of `20` would save ~1GB of memory and only increase execution time by ~11 seconds. I think that this is a relatively small tradeoff to decrease the memory consumption of the linter.

@camilamacedo86 WDYT?

camilamacedo86:
@jakobmoellersap and @everettraven,

I think we need to try to upgrade the lint version to the latest. It seems a bug.
could you open a PR for that?

everettraven:
@camilamacedo86 i can take a look at it. Seems to be failing on a couple other PRs that I have seen as well.

I'll create an issue for it and assign myself.

jakobmoellersap:
Blocked by https://github.com/kubernetes-sigs/kubebuilder/issues/2878

PR #2869, opened 2022-08-11
📖 Fix typo in headline "Namespace"

mogsie:
Replaced "Namesapce" with "Namespace"

Not much to say here except sorry for the extra work.  Somewhat important because it will be a hypertext target.


camilamacedo86 (approved):
/lgtm

/ok-to-test

PR #2876, opened 2022-08-15
🌱 migrate kubebuilder CLI tests to ginkgo/v2 (this pr does not change the scaffolds done by the tool)

jakobmoellersap:
This PR makes sure to migrate kubebuilder ~and the corresponding scaffolding~ to gingko/v2. Currently the entire test suite of kubebuilder itself is migrated, however there are 2 things missing:

- Scaffolding needs to be updated (see https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/scaffolds/internal/templates/controllers/controller_suitetest.go#L134) but its a breaking change and needs to go into v4-alpha scaffolding which does not yet exist
- gingko CLI is recommended as test execution tool (especially for parallel suites) and we would have to adjust Makefiles for that, I want to take this opportunity and ask what about our POV on this


jakobmoellersap:
I would suggest to even open up this PR normally (to upgrade kubebuilder and its test) and then a follow-up PR for the scaffolding. That would make it a bit easier to maintain / review / plan. WDYT @camilamacedo86 ?

camilamacedo86:
Hi @jakobmoellersap that is fine, 

we can also

/ok-to-test

/test-all

camilamacedo86 (approved):
Hi @jakobmoellersap, 

I understand that this PR just change the CLI tests and not the tests scaffolded by the tool.

You are right and migrating the scaffold is a breaking change, however, in this case, I think we will not have the choice because controller-runtime and k8s which are dependencies of this project introduced that, see: https://github.com/kubernetes-sigs/controller-runtime/pull/1977. Therefore, to keep the compatibility and look for maintainability it seems that the best approach is in a follow up we also change the default scaffolds and add the warning (⚠️ ). In the changelog, release we can provide the info and steps to help out the users change it in their projects easily. 

So, if you would like to work on the follow up please feel free. 

Also, thank you for your terrific work and contribution to the project 🥇 

For me it is all fine: 

/approved 

c/c @everettraven @varshaprasad96

jakobmoellersap:
I will pick up the ginkgo v2 scaffolding migration next

varshaprasad96 (approved):

camilamacedo86:
/lgtm

Receive 2 approvals

PR #2877, opened 2022-08-15
[WIP] :sparkles:: Adding ability to re-scaffold exiting projects

### LINKED ISSUE #2068: Make explicit that we would like to have the migration from v3 to v4 helped by tooling
**Description**
Make explicit that we would like to have the migration from v3 to v4 helped by tooling 

**Suggested Solutions**
- Add new plugin interface documented with this information
OR
- Add doc which describes that

rumstead:
Closes #2068 

Relevant [discussion](https://github.com/kubernetes-sigs/kubebuilder/discussions/2864)

## Description
Add tooling to help migrate existing projects to newer versions of scaffolding.

### DIFF
--- a/pkg/cli/scaffold.go
+++ b/pkg/cli/scaffold.go
@@ -19,8 +19,11 @@ package cli
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/kubebuilder/v3/pkg/config/store/yaml"
 )
 
 // NewCommand return a new scaffold command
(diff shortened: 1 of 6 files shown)


pkg/cli/scaffold.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+// NewCommand return a new scaffold command
+func (c *CLI) newScaffoldCmd() *cobra.Command {
```
rumstead, line 27: This feels like it should be in the `alpha` folder but since I believe I need methods from `CLI` I cannot use a different package.

rumstead: @camilamacedo86 looking for some guidance here. Since the `scaffold` command is going to be essionaially doing an `init` + `edit` + `create`, I need access to the`cli` package.

Issue #2627, opened 2022-04-19
RFE: Generate api sub-package, with (almost) no external dependencies.


squeed:
### What do you want to happen?

When generating api types with `kubebuilder create api`, it generates the corresponding Go types in `./api/v1`. 

However, it is unergonomic for external projects to import these types. Mostly because they need to resolve **all** the project's dependencies. This is especially awkward for projects that don't use `controller-runtime`, but it can also be annoying for projects that do, but have different versions.

The fix for this is relatively simple:

1. Add a `go.mod` in `./api`, thus making it, as far as Go is concerned, a separate package
2. Add a corresponding `replace` directive in the root `go.mod`
3. Stop using `sigs.k8s.io/controller-runtime/pkg/scheme` in the generated api types, which is a very thin convenience package.

We can also consider using go 1.18's workspace mode, but it's not actually necessary in this case.

#### What did you do?

`kubebuilder create api`

#### What did you expect to see?

A standalone importable package. Something like
```
$ ls api
v1
go.mod
$ cat go.mod
...
replace (
        github.com/sample/operator/api => ./api
)
```

Furthermore, I would like to see that `api/go.mod` has almost no external dependencies (i.e. `controller-runtime`).

#### What did you see instead? Under which circumstances?

No additional go.mod, etc

squeed:
I'm happy to take a look at this -- changing the scaffolding for new projects is simple enough, but I'm not sure what should happen for existing projects.

squeed:
We could also solve #3 by publishing `sigs.k8s.io/controller-runtime/pkg/scheme` as a separate package internal to `controller-runtime`. That way API-consuming projects could still avoid importing the entire dependency tree.

camilamacedo86:
HI @squeed, 

Thank you for reaching out. WDYT about answering the following questions so that it might help us to better understand why and what you need. 

```
#### What did you do?

<!-- A clear and concise description of the steps you took (or insert a code snippet). -->

#### What did you expect to see?

<!-- A clear and concise description of what you expected to happen (or insert a code snippet). -->

#### What did you see instead? Under which circumstances?

<!-- A clear and concise description of what ACTUALLY happened (or insert a code snippet). -->

```

 Could you please provide this info?

NikhilSharmaWe:
@camilamacedo86 
Do we need to work on this?
And if yes, is the approach described in the issue description reasonable for tackling it?

squeed:
[...]

PR #2880, opened 2022-08-16
:sparkles: implement `kubebuilder init --workspace` flag for go1.18 style go.work support

jakobmoellersap:
This PR attempts to create a backwards-compatible `go.work` support in go/v3 / go/v4-alpha: Fixes https://github.com/kubernetes-sigs/kubebuilder/issues/2627

The basic implementation uses a flag called `--workspace` that can be used in `kubebuilder init`:
`kubebuilder init --workspace`

The create API then still looks like

`kubebuilder create api --version v1alpha1 --kind TestKind --group group`, which leads to a `go.work` being created (if not present) or overwritten with the following scaffold:

```
go 1.18

use ./api/v1alpha1

//+kubebuilder:scaffold:apis

use .
```

It also changes the PostScaffolding behavior of the `create api` and `init` commands by using `go mod sync` instead of `go mod tidy` to automatically resolve all modules in the workspace. This allows us to not rely on replace statements which would be harder to scaffold and maintain. Obvious downside to this is that it supports go 1.18 and above only.

Last but not least, to fully include the changes into our scaffolding, the Dockerfile Scaffolding is introduced. This ensures proper layering and the same (if not even more optimized speed during dependency resolution as before when building your image:

```Dockerfile
# Build the manager binary
FROM golang:1.18 as builder

WORKDIR /workspace
# Copy the Go Modules manifests
COPY go.mod go.mod
COPY go.sum go.sum

# Copy all Submodules from the workspace

COPY api/v1alpha1/go.mod api/v1alpha1/go.mod
COPY api/v1alpha1/go.sum api/v1alpha1/go.sum
#+kubebuilder:scaffold:module-dependency-layers

# Copy the Go Workspace manifests
COPY go.work go.work
COPY go.work.sum go.work.sum
RUN go work sync

# cache deps before building and copying source so that we don't need to re-download as much
# and so that source changes don't invalidate our downloaded layer
RUN go mod download

# Copy the go source
COPY main.go main.go
COPY api/ api/

# Build
RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go

WORKDIR /
USER 65532:65532

ENTRYPOINT ["/manager"]
```

I want to use this PR first in a draft state to collect input and feedback before progressing it and writing tests.

All existing Makefile commands as well as the already scaffolded `go.mod` stays compatible, however `go mod tidy` will not work on the core module on its own. (since `go.work` is necessary to resolve the other modules relative to the base module). Notice that for external developers, you will be able to download the module as normal and also depend on the API modules individually without actually having to import the controller.

### DIFF
--- a/pkg/machinery/marker.go
+++ b/pkg/machinery/marker.go
@@ -25,10 +25,11 @@ import (
 const prefix = "+kubebuilder:scaffold:"
 
 var commentsByExt = map[string]string{
-	".go":   "//",
-	".work": "//",
-	".yaml": "#",
-	".yml":  "#",
+	".go":        "//",
+	".work":      "//",
+	".yaml":      "#",
+	".yml":       "#",
+	"Dockerfile": "#",
 	// When adding additional file extensions, update also the NewMarkerFor documentation and error
 }
 
@@ -39,10 +40,13 @@ type Marker struct {
 }
 
 // NewMarkerFor creates a new marker customized for the specific file
-// Supported file extensions: .go, .work, .yaml, .yml
+// Supported file extensions: .go, .work, .yaml, .yml, Dockerfile
 func NewMarkerFor(path string, value string) Marker {
-	ext := filepath.Ext(path)
-	if comment, found := commentsByExt[ext]; found {
+	extOrFilename := filepath.Ext(path)
+	if extOrFilename == "" {
+		_, extOrFilename = filepath.Split(path)
+	}
+	if comment, found := commentsByExt[extOrFilename]; found {
 		return Marker{comment, value}
 	}
 
@@ -50,7 +54,7 @@ func NewMarkerFor(path string, value string) Marker {
 	for extension := range commentsByExt {
 		extensions = append(extensions, fmt.Sprintf("%q", extension))
 	}
-	panic(fmt.Errorf("unknown file extension: '%s', expected one of: %s", ext, strings.Join(extensions, ", ")))
+	panic(fmt.Errorf("unknown file extension: '%s', expected one of: %s", extOrFilename, strings.Join(extensions, ", ")))
 }
 
 // String implements Stringer
--- a/pkg/machinery/marker.go
+++ b/pkg/machinery/marker.go
@@ -26,6 +26,8 @@ const prefix = "+kubebuilder:scaffold:"
 
 var commentsByExt = map[string]string{
 	".go":   "//",
+	".mod":  "//",
+	".work": "//",
 	".yaml": "#",
 	".yml":  "#",
 	// When adding additional file extensions, update also the NewMarkerFor documentation and error
@@ -38,7 +40,7 @@ type Marker struct {
 }
 
 // NewMarkerFor creates a new marker customized for the specific file
-// Supported file extensions: .go, .yaml, .yml
+// Supported file extensions: .go, .mod, .work, .yaml, .yml
 func NewMarkerFor(path string, value string) Marker {
 	ext := filepath.Ext(path)
 	if comment, found := commentsByExt[ext]; found {
--- a/pkg/machinery/marker.go
+++ b/pkg/machinery/marker.go
@@ -26,7 +26,6 @@ const prefix = "+kubebuilder:scaffold:"
 
 var commentsByExt = map[string]string{
 	".go":   "//",
-	".mod":  "//",
 	".work": "//",
 	".yaml": "#",
 	".yml":  "#",
@@ -40,7 +39,7 @@ type Marker struct {
 }
 
 // NewMarkerFor creates a new marker customized for the specific file
-// Supported file extensions: .go, .mod, .work, .yaml, .yml
+// Supported file extensions: .go, .work, .yaml, .yml
 func NewMarkerFor(path string, value string) Marker {
 	ext := filepath.Ext(path)
 	if comment, found := commentsByExt[ext]; found {
--- a/pkg/plugins/golang/v3/api.go
+++ b/pkg/plugins/golang/v3/api.go
@@ -188,7 +188,7 @@ func (p *createAPISubcommand) PreScaffold(fs machinery.Filesystem) error {
 }
 
 func (p *createAPISubcommand) Scaffold(fs machinery.Filesystem) error {
-	scaffolder := scaffolds.NewAPIScaffolder(p.config, *p.resource, p.force, p.useWorkspaces)
+	scaffolder := scaffolds.NewWorkspaceAPIScaffolder(p.config, *p.resource, p.force, p.useWorkspaces)
 	scaffolder.InjectFS(fs)
 	return scaffolder.Scaffold()
 }
--- a/pkg/plugins/golang/v3/api.go
+++ b/pkg/plugins/golang/v3/api.go
@@ -59,6 +59,9 @@ type createAPISubcommand struct {
 
 	// runMake indicates whether to run make or not after scaffolding APIs
 	runMake bool
+
+	// useWorkspaces indicates whether to use go.work style workspaces instead of a go module
+	useWorkspaces bool
 }
 
 func (p *createAPISubcommand) UpdateMetadata(cliMeta plugin.CLIMetadata, subcmdMeta *plugin.SubcommandMetadata) {
@@ -95,6 +98,8 @@ make generate will be run.
 
 func (p *createAPISubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&p.runMake, "make", true, "if true, run `make generate` after generating files")
+	fs.BoolVar(&p.useWorkspaces, "workspace", true, "if true, scaffolds apis with `go.work` workspace differentiation, "+
+		"creating a go.mod file for each generated api.")
 
 	fs.BoolVar(&p.force, "force", false,
 		"attempt to create resource even if it already exists")
@@ -180,13 +185,12 @@ func (p *createAPISubcommand) PreScaffold(machinery.Filesystem) error {
 }
 
 func (p *createAPISubcommand) Scaffold(fs machinery.Filesystem) error {
-	scaffolder := scaffolds.NewAPIScaffolder(p.config, *p.resource, p.force)
+	scaffolder := scaffolds.NewAPIScaffolder(p.config, *p.resource, p.force, p.useWorkspaces)
 	scaffolder.InjectFS(fs)
 	return scaffolder.Scaffold()
 }
 
 func (p *createAPISubcommand) PostScaffold() error {
-
 	// Update the makefile to allow generate Webhooks to ensure backwards compatibility
 	// todo: it should be removed for go/v4
 	// nolint:lll,gosec
@@ -196,12 +200,18 @@ func (p *createAPISubcommand) PostScaffold() error {
 		}
 	}
 
-	err := util.RunCmd("Update dependencies", "go", "mod", "tidy")
-	if err != nil {
-		return err
+	if p.useWorkspaces {
+		if err := util.RunCmd("Update workspace", "go", "work", "sync"); err != nil {
+			return err
+		}
+	} else {
+		if err := util.RunCmd("Update dependencies", "go", "mod", "tidy"); err != nil {
+			return err
+		}
 	}
+
 	if p.runMake && p.resource.HasAPI() {
-		err = util.RunCmd("Running make", "make", "generate")
+		err := util.RunCmd("Running make", "make", "generate")
 		if err != nil {
 			return err
 		}
--- a/pkg/plugins/golang/v3/api.go
+++ b/pkg/plugins/golang/v3/api.go
@@ -100,7 +100,7 @@ func (p *createAPISubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&p.runMake, "make", true, "if true, run `make generate` after generating files")
 
 	// TODO Evaluate if we put that in goPlugin
-	fs.BoolVar(&p.useWorkspaces, "workspace", true, "if true, scaffolds apis with `go.work` workspace differentiation, "+
+	fs.BoolVar(&p.useWorkspaces, "workspace", false, "if true, scaffolds apis with `go.work` workspace differentiation, "+
 		"creating a go.mod file for each generated api.")
 
 	fs.BoolVar(&p.force, "force", false,
--- a/pkg/plugins/golang/v3/api.go
+++ b/pkg/plugins/golang/v3/api.go
@@ -98,6 +98,8 @@ make generate will be run.
 
 func (p *createAPISubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&p.runMake, "make", true, "if true, run `make generate` after generating files")
+
+	// TODO Evaluate if we put that in goPlugin
 	fs.BoolVar(&p.useWorkspaces, "workspace", true, "if true, scaffolds apis with `go.work` workspace differentiation, "+
 		"creating a go.mod file for each generated api.")
 
@@ -200,6 +202,8 @@ func (p *createAPISubcommand) PostScaffold() error {
 		}
 	}
 
+	// for workspace support, we use go mod sync to make sure all modules are considered, not just the root.
+	// this automatically also triggers go mod tidy for all modules and creates go.sum files.
 	if p.useWorkspaces {
 		if err := util.RunCmd("Update workspace", "go", "work", "sync"); err != nil {
 			return err
--- a/pkg/plugins/golang/v3/api.go
+++ b/pkg/plugins/golang/v3/api.go
@@ -60,7 +60,6 @@ type createAPISubcommand struct {
 	// runMake indicates whether to run make or not after scaffolding APIs
 	runMake bool
 
-	// useWorkspaces indicates whether to use go.work style workspaces instead of a go module
 	useWorkspaces bool
 }
 
@@ -98,10 +97,6 @@ make generate will be run.
 
 func (p *createAPISubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&p.runMake, "make", true, "if true, run `make generate` after generating files")
-
-	// TODO Evaluate if we put that in goPlugin
-	fs.BoolVar(&p.useWorkspaces, "workspace", false, "if true, scaffolds apis with `go.work` workspace differentiation, "+
-		"creating a go.mod file for each generated api.")
 
 	fs.BoolVar(&p.force, "force", false,
 		"attempt to create resource even if it already exists")
@@ -177,10 +172,16 @@ func (p *createAPISubcommand) InjectResource(res *resource.Resource) error {
 	return nil
 }
 
-func (p *createAPISubcommand) PreScaffold(machinery.Filesystem) error {
+func (p *createAPISubcommand) PreScaffold(fs machinery.Filesystem) error {
 	// check if main.go is present in the root directory
 	if _, err := os.Stat(DefaultMainPath); os.IsNotExist(err) {
 		return fmt.Errorf("%s file should present in the root directory", DefaultMainPath)
+	}
+
+	// Check if we use workspaces by presence of go.work, fallback to go.mod otherwise
+	p.useWorkspaces = true
+	if _, err := fs.FS.Stat("go.work"); err != nil {
+		p.useWorkspaces = false
 	}
 
 	return nil
--- a/pkg/plugins/golang/v3/init.go
+++ b/pkg/plugins/golang/v3/init.go
@@ -125,7 +125,7 @@ func (p *initSubcommand) PreScaffold(machinery.Filesystem) error {
 }
 
 func (p *initSubcommand) Scaffold(fs machinery.Filesystem) error {
-	scaffolder := scaffolds.NewInitScaffolder(
+	scaffolder := scaffolds.NewWorkspaceInitScaffolder(
 		p.config, p.license, p.owner, p.useWorkspaces)
 	scaffolder.InjectFS(fs)
 	if err := scaffolder.Scaffold(); err != nil {
--- a/pkg/plugins/golang/v3/init.go
+++ b/pkg/plugins/golang/v3/init.go
@@ -56,6 +56,7 @@ type initSubcommand struct {
 	// flags
 	fetchDeps          bool
 	skipGoVersionCheck bool
+	useWorkspaces      bool
 }
 
 func (p *initSubcommand) UpdateMetadata(cliMeta plugin.CLIMetadata, subcmdMeta *plugin.SubcommandMetadata) {
@@ -87,6 +88,9 @@ func (p *initSubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&p.license, "license", "apache2",
 		"license to use to boilerplate, may be one of 'apache2', 'none'")
 	fs.StringVar(&p.owner, "owner", "", "owner to add to the copyright")
+
+	fs.BoolVar(&p.useWorkspaces, "workspace", false, "if true, scaffolds apis with `go.work` workspace differentiation, "+
+		"creating a go.mod file for each generated api and a go.work in the root.")
 
 	// project args
 	fs.StringVar(&p.repo, "repo", "", "name to use for go module (e.g., github.com/user/repo), "+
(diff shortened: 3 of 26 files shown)


pkg/machinery/marker.go
```diff
@@ -38,7 +40,7 @@ type Marker struct {
 }
 
 // NewMarkerFor creates a new marker customized for the specific file
-// Supported file extensions: .go, .yaml, .yml
+// Supported file extensions: .go, .mod, .work, .yaml, .yml
```
camilamacedo86, line 43: why do we need the add the .work extension ?

pkg/machinery/marker.go
```diff
@@ -26,6 +26,8 @@ const prefix = "+kubebuilder:scaffold:"
 
 var commentsByExt = map[string]string{
 	".go":   "//",
+	".mod":  "//",
```
camilamacedo86, line 29: Also, why do we need to add the mod and .work?
What markers do we need to look for and replace in these files?

pkg/plugins/golang/v3/api.go
```diff
@@ -59,6 +59,9 @@ type createAPISubcommand struct {
 
 	// runMake indicates whether to run make or not after scaffolding APIs
 	runMake bool
+
+	// useWorkspaces indicates whether to use go.work style workspaces instead of a go module
+	useWorkspaces bool
```
camilamacedo86, line 64: What is go.work?
When and why it is used?
Is that I third party?

pkg/plugins/golang/v3/api.go
```diff
@@ -59,6 +59,9 @@ type createAPISubcommand struct {
...
 }
```
camilamacedo86, line 66: Why do we need to use go.work for Fixes https://github.com/kubernetes-sigs/kubebuilder/issues/2627? 
Why the go.mod is not allowed inside if the api?
Is not the controller-tools who does not accepted that?
What error is faced and from where when we try to create a go.mod inside of the api?

pkg/plugins/golang/v3/scaffolds/internal/templates/gowork.go
```diff
@@ -0,0 +1,131 @@
+/*
```
camilamacedo86, line 1: I think the idea of fixing https://github.com/kubernetes-sigs/kubebuilder/issues/2627 would see if we can help allow them to create a go. mod in this directory but we do not need to provide a default scaffold and officially support. Just to see what changes we need to do in the tools to not block their action. 

But please, share your thoughts I might missing something here.

pkg/machinery/marker.go
jakobmoellersap: The .work extension is necessary because of the scaffolding in the `go.work` file that adds api modules as they are generated:

```go.work
go 1.18

use ./api/v1alpha1 // <<=== this is generated for each kubebuilder create api request

//+kubebuilder:scaffold:apis // this is the scaffolding we need to read

use .
```

pkg/machinery/marker.go
jakobmoellersap: .mod I think is not necessary, will double check. .work is necessary because of the .work (see my comment above). The marker we can discuss, I just had to come up with something

pkg/plugins/golang/v3/api.go
jakobmoellersap: `go work sync` and `go work init` are inbuilt into the go cli since go 1.18.
You can find the spec here: https://pkg.go.dev/cmd/go/internal/workcmd#:~:text=A%20workspace%20is%20specified%20by%20a%20go.work%20file,builds%20from%20local%20modules.%20go.work%20files%20are%20line-oriented.

Or by using `go help work` which will open the go help for this. I highly recommend it!

pkg/plugins/golang/v3/api.go
jakobmoellersap: So basically if we do not want to break compatibility with `go/v3` and offer it in this plugin I have to find a way to make the existing `create api` command breaking. The only way I can do this is by checking the flag and then deciding whether to create a `go.work` in the root and `go.mod` files in the modules. I think its not at all controller-tools related and just related to our scaffolding. The errors will start if you try to combine some modules with `go.mod` and some without, then it breaks the building process since the main build has to switch to `go mod sync`

pkg/plugins/golang/v3/scaffolds/internal/templates/gowork.go
jakobmoellersap: You can always manually create go.mod and go.work files and do the scaffold I am doing here but then the `create api` commands will fail since it will try to use `go mod tidy` in the PostProcessing. This will only work if you issue replace directives in the root `go.mod` => This is what I wanted to avoid since it makes it really hard to maintain.

pkg/machinery/marker.go
jakobmoellersap: Instead of using this we could also think about using `go work edit` through the CLI but this would include it as a dependency and we did not use this pattern before so I was unsure if we should go for that

jakobmoellersap: You were right, using `.mod` is not necessary since I do not scaffold anything here. `.work` is still necessary due to the inclusion of the use directive

jakobmoellersap:
@camilamacedo86 could i get another ok-to-test? I tried to refactor the commands to make use of the init command to setup the `go.work` file since it makes sense to include it in the initial scaffolding of a project instead of every API creation

pkg/machinery/marker.go
jakobmoellersap: On that note, I noticed the Dockerfile also needs to be optimized with workspaces, so I started to introduce a Dockerfile Marker Support as well

jakobmoellersap:
Putting this as ready to review, please provide feedback before I proceed further

jakobmoellersap:
/hold

there is one big thing open that is requiring the go.mod file contents in the root

everettraven:
@jakobmoellersap I like the ideas behind this change, and I think the work you have done looks great. I just have a few of questions:
- My understanding is that this introduces a breaking change for the `create api` subcommand for the `go/v3` plugin. Is this correct? If so, I think there is probably some hesitation in going this route since the `go/v3` scaffolding logic is already widely used by both Kubebuilder and Operator SDK users when building Go based operators.
- Does this have to be done in the scaffolding of the `create api` subcommand?
    - Have we considered potentially using the `edit` subcommand to facilitate these changes? I think if we could make the changes in the `edit` subcommand we wouldn't have to implement any breaking changes to the existing `create api` logic and rather just introduce a new non-breaking change to the `edit` subcommand.

jakobmoellersap:
@everettraven Thanks so much for taking a look!

> * My understanding is that this introduces a breaking change for the `create api` subcommand for the `go/v3` plugin. Is this correct? If so, I think there is probably some hesitation in going this route since the `go/v3` scaffolding logic is already widely used by both Kubebuilder and Operator SDK users when building Go based operators.

I tried to make the entire change backwards-compatible with v3. You should be able to still run all tests and commands like before and you will not even see a difference. It will only show up once you call `kubebuilder init --workspace`. This is because I changed my original approach from having the flag in the `create api` command to now be in `init` instead. There are two changes necessary in the Go API of the module which I tried to solve by creating a second method (for initializing the Scaffolder) with one more parameter and referencing the old methods as deprecated. Since the additional parameter is bool you can even call it with the default value of false and it will behave as the old API. This is also why all API compatibility tests are still green and all cluster tests run just fine.

> * Does this have to be done in the scaffolding of the `create api` subcommand?

No => you are right it, doesn't have to be in the `create api`. Compared to the very beginning, I changed the main scaffolding of the go.work file to be included in the `init` command. the `create api` commands will only alter their behavior when a go.work file is detected.

>   * Have we considered potentially using the `edit` subcommand to facilitate these changes? I think if we could make the changes in the `edit` subcommand we wouldn't have to implement any breaking changes to the existing `create api` logic and rather just introduce a new non-breaking change to the `edit` subcommand.

I mentioned this as well in the issue and honestly I would like to use it, but I think there are some drawbacks due to the existing kubebuilder scaffolding infrastructure:
- It is not compatible with the `scaffolding` API and we would have to use an extra scaffolding that is not done before in kubebuilder based on a command. (right now we only use `go mod tidy` and make commands and we only do it in `PreScaffold` or `PostScaffold` after the actual scaffolding already ran)
- If we use the in-code dependency we will have to include it in-code as a dependency and work with their API, namely we would need a combination of the contents of `RepositoryMixIn` and `TemplateMixIn`. I think this is an option as well! I am open for feedback here. We would not have to use a `kubebuilder` scaffolding annotation that way.

jakobmoellersap:
One more addition here that I want to put up for discussion:
I tried to make this run with purely `go.mod` and `replace` directives. However there is a huge problem with this:
You cannot reference a module that has not yet been pushed to the remote (e.g. Github). 

For example:
```go.mod
module github.com/jakobmoellersap/test.com

go 1.18

require (
        sigs.k8s.io/controller-runtime v0.12.2
        github.com/jakobmoellersap/test.com/api/v1alpha1 0.0.0 //replace with released version later on
        //+kubebuilder:scaffold:module-dependency
)  
```
together with
```go.mod
replace github.com/jakobmoellersap/test.com/api/v1alpha1 => ./api/v1alpha1
```

will break even though we replaced it. This is because of the following error:
```
go: github.com/jakobmoellersap/test.com/api/v1alpha1@v0.0.0: reading github.com/jakobmoellersap/test.com/api/v1alpha1/go.mod at revision api/v1alpha1/v0.0.0: git ls-remote -q origin in modcachepath/cache/vcs/d8b5eb86b10c53eaa4c311cd6dd4ef713ab591d0cf9c05f7709ed82b6de97ab5: exit status 128:
        remote: Repository not found.
        fatal: repository 'https://github.com/jakobmoellersap/test.com/' not found
```

everettraven:
@jakobmoellersap thanks for taking the time to respond to my previous comment, this all makes sense now. Apologies for missing a couple of the things you mentioned already describing in the issue :sweat_smile: .

I am leaning towards the idea of introducing the functionality into the `kubebuilder edit` subcommand for the `go/v3` plugin. To me this seems a little bit like wanting the ability to "edit" the default scaffolding so that Go workspaces can be utilized - kind of similar to how the `kubebuilder edit` subcommand can be used to enable multigroup functionality. I am okay with having to make some changes to the scaffolding setup of the edit command to accommodate this type of change under the hood (I think all the subcommands should be able to implement the same scaffolding logic).

One thing I will note, is that if using Go workspaces is a good idea to start setting up as a best practice for the way that operators are scaffolded, I would be okay with adjusting the `go/v4-alpha` plugin to start using Go workspaces by default (unless we are concerned about the need to use `go/v4-alpha` scaffolds with a Go version < 1.18). 

What do you think?

pkg/plugins/golang/v3/api.go
```diff
@@ -59,6 +59,8 @@ type createAPISubcommand struct {
 
 	// runMake indicates whether to run make or not after scaffolding APIs
 	runMake bool
+
+	useWorkspaces bool
```
camilamacedo86, line 63: @jakobmoellersap. 

Since the workspace option changes the behaviour and layout then, I think it should be like the multi-group one. That means if we call `kubebuilder edit --workspaces=true` we will store the value in the PROJECT file (which represents the config of the project). Therefore, when we read the config we would have this information.

See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-multigroup/PROJECT#L4
And its docs: https://book.kubebuilder.io/migration/multi-group.html

pkg/plugins/golang/v3/api.go
```diff
@@ -170,23 +172,28 @@ func (p *createAPISubcommand) InjectResource(res *resource.Resource) error {
...
 	// check if main.go is present in the root directory
 	if _, err := os.Stat(DefaultMainPath); os.IsNotExist(err) {
 		return fmt.Errorf("%s file should present in the root directory", DefaultMainPath)
 	}
 
+	// Check if we use workspaces by presence of go.work, fallback to go.mod otherwise
+	p.useWorkspaces = true
+	if _, err := fs.FS.Stat("go.work"); err != nil {
+		p.useWorkspaces = false
+	}
```
camilamacedo86, line 185: By having this info tracked in the PROJECT file ( in both cases if that be a plugin or an option to enable like the multigroup one ) we can only check the config to change the behaviour accordingly.

pkg/plugins/golang/v3/api.go
```diff
@@ -170,23 +172,28 @@ func (p *createAPISubcommand) InjectResource(res *resource.Resource) error {
...
+
 	return nil
 }
 
 func (p *createAPISubcommand) Scaffold(fs machinery.Filesystem) error {
-	scaffolder := scaffolds.NewAPIScaffolder(p.config, *p.resource, p.force)
+	scaffolder := scaffolds.NewWorkspaceAPIScaffolder(p.config, *p.resource, p.force, p.useWorkspaces)
```
camilamacedo86, line 191: I do not think that we should change `NewAPIScaffolder ` , the info about the layout should be in the config so that we need to need to change anything here,

pkg/plugins/golang/v3/api.go
```diff
@@ -196,12 +203,20 @@ func (p *createAPISubcommand) PostScaffold() error {
 		}
 	}
 
-	err := util.RunCmd("Update dependencies", "go", "mod", "tidy")
-	if err != nil {
-		return err
+	// for workspace support, we use go mod sync to make sure all modules are considered, not just the root.
+	// this automatically also triggers go mod tidy for all modules and creates go.sum files.
+	if p.useWorkspaces {
```
camilamacedo86, line 208: same here

pkg/plugins/golang/v3/init.go
```diff
@@ -88,6 +89,9 @@ func (p *initSubcommand) BindFlags(fs *pflag.FlagSet) {
 		"license to use to boilerplate, may be one of 'apache2', 'none'")
 	fs.StringVar(&p.owner, "owner", "", "owner to add to the copyright")
 
+	fs.BoolVar(&p.useWorkspaces, "workspace", false, "if true, scaffolds apis with `go.work` workspace differentiation, "+
```
camilamacedo86, line 92: If we move forward with flags then I think we ought to do it as multi-group option

pkg/plugins/golang/v3/init.go
```diff
@@ -121,10 +125,10 @@ func (p *initSubcommand) PreScaffold(machinery.Filesystem) error {
 }
 
 func (p *initSubcommand) Scaffold(fs machinery.Filesystem) error {
-	scaffolder := scaffolds.NewInitScaffolder(p.config, p.license, p.owner)
+	scaffolder := scaffolds.NewWorkspaceInitScaffolder(
```
camilamacedo86, line 128: I do not think that we should change the method to initScaffolder, any required config ought to be in the PROJECT file.

pkg/plugins/golang/v3/scaffolds/api.go
```diff
@@ -44,9 +44,13 @@ type apiScaffolder struct {
 
 	// force indicates whether to scaffold controller files even if it exists or not
 	force bool
+
+	// useWorkspaces indicates whether to use go.work style workspaces instead of a go module
+	useWorkspaces bool
```
camilamacedo86, line 49: Why we would need to inform it every time that we create an API?
I think it would be like multi-group if we change the layout to accept that then, we have the info in the PROJECT file and it would be valid for all apis scaffolded.

pkg/plugins/golang/v3/scaffolds/api.go
```diff
@@ -55,6 +59,19 @@ func NewAPIScaffolder(config config.Config, res resource.Resource, force bool) p
...
+func NewWorkspaceAPIScaffolder(
+	config config.Config, res resource.Resource, force bool, useWorkspaces bool,
+) plugins.Scaffolder {
+	return &apiScaffolder{
+		config:        config,
+		resource:      res,
+		force:         force,
+		useWorkspaces: useWorkspaces,
+	}
+}
```
camilamacedo86, line 73: Same here.

camilamacedo86 (changes requested):
Some changes shows required.

pkg/machinery/marker.go
```diff
@@ -25,9 +25,11 @@ import (
...
 
 var commentsByExt = map[string]string{
-	".go":   "//",
-	".yaml": "#",
-	".yml":  "#",
+	".go":        "//",
+	".work":      "//",
+	".yaml":      "#",
+	".yml":       "#",
+	"Dockerfile": "#",
```
camilamacedo86, line 32: Why we need to add the Dockerfile here?

jakobmoellersap: As you can see in the initial comment of the PR, the Dockerfile has to be adjusted to stay optimized, otherwise the go mod download command will only download dependencies of the main module, not of API modules as well. If we can live with that, we do not have to touch the Dockerfile, but I would still say that we should optimize it here.

jakobmoellersap:
Thanks for the insight and review guys, I will adopt the PR to make use of the `kubebuilder edit` command and the `PROJECT` configuration file.

jakobmoellersap:
After spending more time into research, I think workspaces also can have major disadvantages. The best practice way of go is still to use `one module per repository`. That is to say that I think the pure best practice here would be to 

- create separate repositories and check them into vcs separately
- maintain two modules but have to undergo a lot of hands-on effort as you will need to check one module into VCS before checking in the other (otherwise the problem I described above becomes present)

I used the following resources as references:
https://github.com/pachyderm/pachyderm/pull/3870
https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository
https://github.com/golang/go/issues/27056
https://go.googlesource.com/proposal/+/master/design/45713-workspace.md#rationale-the-file
https://go.dev/doc/modules/managing-source
https://go.dev/doc/modules/developing
https://go.dev/doc/modules/version-numbers

I also looked way deeper into the go docs now on recommendations of multi-modules, and to be honest, what we are trying here with the PR is probably not an easy to achieve solution in light of the go ecosystem:

From the `golang wiki`:

> Two example scenarios where it can make sense to have more than one go.mod in a repository:
>
> if you have usage examples where the examples themselves have a complex set of dependencies (e.g., perhaps you have a small package but include an example of using your package with kubernetes). In that case, it can make sense for your repository to have an example or _example directory with its own go.mod, such as shown [here](https://pkg.go.dev/github.com/loov/hrtime).
>
> if you have a repository with a complex set of dependencies, but you have a client API with a smaller set of dependencies. In some cases, it might make sense to have an api or clientapi or similar directory with its own go.mod, or to separate out that clientapi into its own repository.
>
> However, for both of those cases, if you are considering creating a multi-module repository for performance or download size for a large set of indirect dependencies, you are strongly encouraged to first try with a GOPROXY, which will be enabled by default in Go 1.13. Using a GOPROXY mostly equals any performance benefits or dependency download size benefits that might otherwise come from creating a multi-module repository.

What I can safely say by now is that `go.work` files are not meant to be checked into VCS, but rather replace the local `replace` directives one has to issue if working with multiple modules locally. This also means we're not talking about a pure best practice solution anymore. Instead the recommendation is to use a completely different repository.

I will also not be able to make the main go.mod file work independently without a go.mod file containing a proper replace directive. But we cannot add the replace directive without first checking in the API Module 

TLDR: I have a problem now since I think we cannot solve this issue without interacting with the VCS first in the edit command to first create the API Module before initializing the second module and then issuing a `replace` statement.

pkg/machinery/marker.go
camilamacedo86: HI @jakobmoellersap, 

We can customize the Dockerfile, as we do for example with the declarative plugin. See that you can use the plugin utils to insert, replace code.

This implementation is for we create markers to inject code. Are we creating any marker for the dockerfile or these other extensions?

everettraven:
My understanding now is that the `go.work` workspaces are meant only for local development purposes as a substitute for the `replace` directives that one would put into their `go.mod` files when working with multiple modules. To me, this means that it doesn't quite solve the problem of wanting to be able to have the `api` sub-module be its own importable module with a smaller set of dependencies.

It is also my understanding that the best practice in its purest form is to have separate repositories - which I don't think will be reasonably feasible.

@jakobmoellersap Looking through some of the resources that you included, I feel it may be best to stray away from the Go workspaces since I don't think it quite solves the problem referenced in #2627. It also seems that it is likely our best route to go the direction of introducing the `go.mod` to the `api` sub-module to make it its own separate module. This ends up in the creation of a multi-module repository which while isn't absolute best practice, is acceptable when necessary. As mentioned in the [Go wiki's "should I have multiple modules in the same repository?" section](https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository):
>if you have a repository with a complex set of dependencies, but you have a client API with a smaller set of dependencies. In some cases, it might make sense to have an api or clientapi or similar directory with its own go.mod, or to separate out that clientapi into its own repository.

I think our use case aligns with this point and would warrant the creation of a multi-module repository. 

TLDR: My suggestion is we drop the Go workspaces approach and instead try the multi-module route that was originally suggested in #2627

@camilamacedo86 @jakobmoellersap What do you think?

jakobmoellersap:
@everettraven Agree with all of the above, except for
>I think our use case aligns with this point and would warrant the creation of a multi-module repository. If we do this at the time of scaffolding we shouldn't have to do any weird VCS manipulation.

I think the first case, while straight-forward, is a bit misleading to read:
>The first class: the package to which the module is being added to is not in version control yet (a new package). This case is straightforward: add the package and the go.mod in the same commit, tag the commit, and push.

This means that to make this work correctly we would still have to create a tag for the release of the module `something like v1.0.0-alpha1` and push it to make the go.mod work. Before the git push, `go mod tidy` will not work since the VCS cannot be targeted and the go commands will throw an error trying to resolve the module.

So the workflow for multi-module creation would be:
1. `kubebuilder init``
2. `kubebuilder edit --multi-module-project=true` => this will lead to the go mod entry being created as a dependency.
4. `go mod tidy` => fails because the module is not available in VCS (even when trying to use a `replace directive`
5. `git add -A && git commit -m "kubebuilder multi-module project initialization"`
6. `git tag api/v1alpha1 v1.0.0-alpha1`
7. `git push`
8. `go mod tidy` => now succeds, and can also be used together with replace statement

This is the same problem I encountered in https://github.com/kubernetes-sigs/kubebuilder/pull/2880#issuecomment-1228829539

What would be your suggestion to solve this @everettraven @camilamacedo86 ? If we do not interact with VCS here after creating the second module and depending on it, I think there is no way for kubebuilder to ensure project integrity (or that `go mod tidy` will run)

camilamacedo86:
Hi @jakobmoellersap, 

Regards your comment: https://github.com/kubernetes-sigs/kubebuilder/pull/2880#issuecomment-1231172728
Why that will fails?
Why we need to scaffold the files without any api scaffold?
Also, I do not think that we need to git add at all we just need to install the module inside of the api dir before running go mod tidy for the project. 

Why not:
Create a new plugin that enables it then when we call init|edit with plugins=multi-module/v1alpha we will:  

- a) update the Dockerfile and track it in the Project file (such as it is done with all plugins)
- b) in this case this plugin must be tracked under the layout such as we do with the declarative one. see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3-addon-and-grafana/PROJECT#L4
- c) check if has Resources created if not stop if yes, then scaffold the files on the api dir

Therefore, when we call the create api  (we do not need to pass the multi-module/v1alpha ) the create api subcommand implemented for this plugin will be called in the chain always then we can:

- a) the templates under creating api should be configured with `f.IfExistsAction = machinery.SkipFile` so that if they are created already we will not overwrite them. example: https://github.com/kubernetes-sigs/kubebuilder/blob/a44f85fab72ebbe5bb4f6dfbff9f12838c62a25b/pkg/plugins/golang/declarative/v1/scaffolds/internal/templates/channel.go#L44

pkg/machinery/marker.go
jakobmoellersap: this would have required a marker right now. But I think we can also use the way that the declarative plugin is using

jakobmoellersap:
@camilamacedo86 I think I misunderstood you earlier and retried the approach manually.

The main question for me now is in the replace statement. Is it acceptable to have a replace directive like

```go.mod
require github.com/jakobmoellersap/someothertest.com/api/v1alpha1 v0.0.1
replace github.com/jakobmoellersap/someothertest.com/api/v1alpha1 => ./api/v1alpha1
```
in the main `go.mod`. This will allow the resolution to work but it will also mean that there can be a discrepancy between the "real" v0.0.1 and the one replaced locally. If i do not issue the replace statement, the module will not run.

If it is, then I think we can go ahead with a separate plugin as mentioned above

everettraven:
@jakobmoellersap @camilamacedo86 I think having it be a new plugin is reasonable. I think it would be okay to have that `replace` directive added to the main `go.mod`. I do wonder if we go that route if we should leave a comment above the replace directive mentioning why it is there and when it may be appropriate to comment it out (or remove it)

jakobmoellersap:
@everettraven @camilamacedo86 I created the new flow with a new Plugin here: https://github.com/kubernetes-sigs/kubebuilder/pull/2901,

please have a look if you find time.

Side-Note: I think we can take our learnings here and actually use a go.work file to create a local development context (which we put in .gitignore). Then we do not have to use the replace-statement. We could then use the proper build without go.work in the dockerfile but would have to put a disclaimer in the plugin that this won't work until the release tags are set properly. For now I will leave that out and use a simple replace in the go.mod

jakobmoellersap:
Closing here in favor of the next PR as this will surely not be merged.

PR #2884, opened 2022-08-18
🌱  fix `fmt` lint issues on `master`

jakobmoellersap:
This Mini-PR fixes the go `fmt` complains from run https://github.com/kubernetes-sigs/kubebuilder/actions/runs/2876147334 by running format over them.

### DIFF
--- a/pkg/cli/alpha/config-gen/cmd.go
+++ b/pkg/cli/alpha/config-gen/cmd.go
@@ -17,9 +17,6 @@ limitations under the License.
 package configgen
 
 import (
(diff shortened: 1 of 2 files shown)


camilamacedo86:
/ok-to-test

@jakobmoellersap in this case since the change does not affect end-users we use 🌱 instead.
The emojis are used to help us to create the release notes 
Could you please change that.

Otherwise all shows fine

camilamacedo86 (approved):
Thank you for the contribution

PR #2895, opened 2022-08-25
[WIP] :sparkles:: Adding ability to re-scaffold existing projects

### LINKED ISSUE #2068: Make explicit that we would like to have the migration from v3 to v4 helped by tooling

rumstead:
Closes #2068 

Relevant [discussion](https://github.com/kubernetes-sigs/kubebuilder/discussions/2864)

## Description
Add tooling to help migrate existing projects to newer versions of scaffolding.


rumstead:
@camilamacedo86 / @joelanford (sorry for tagging you so much camilamcedo86 I just read the "don't tag in comments or commits"), looking for guidance here. If we want to reuse the functions that init, create, edit, etc us, we cannot put the command in the alpha package. Thoughts on how to proceed?

pkg/cli/alpha.go
```diff
@@ -47,6 +48,7 @@ Alpha subcommands are for unstable features.
 	for i := range alphaCommands {
 		alpha.AddCommand(alphaCommands[i])
 	}
+	alpha.AddCommand(c.newScaffoldCmd())
```
camilamacedo86, line 51: Any new alpha subcommand can be defined here; https://github.com/kubernetes-sigs/kubebuilder/blob/e7ee3941b34e3ab56362dfaa6f945bfb15a75f8b/pkg/cli/alpha.go#L32-L34

Could you please change it out?

pkg/cli/alpha/config-gen/cmd.go
```diff
@@ -124,7 +125,7 @@ func NewCommand() *cobra.Command {
 	c.Version = `v0.1.0`
 	c.Short = `Generate configuration for controller-runtime based projects`
 	c.Long = strings.TrimSpace(`
-config-gen programatically generates configuration for a controller-runtime based
+config-gen programmatically generates configuration for a controller-runtime based
```
camilamacedo86, line 128: Could you please push a PR only for this change. 
That seems out of the scope of what you are looking for to introduce in this PR

pkg/cli/cli.go
```diff
@@ -70,6 +70,8 @@ type CLI struct { //nolint:maligned
 	pluginKeys []string
 	// Project version to scaffold.
 	projectVersion config.Version
+	// projectName name of project
+	projectName string
```
camilamacedo86, line 74: Why we would need to pass the projectName here? 
The projectName is found in the PROJECT file.

pkg/cli/cli.go
```diff
@@ -214,11 +216,25 @@ func (c *CLI) getInfoFromConfigFile() error {
...
+	if path == "" {
+		return c.getInfoFromConfigFile()
+	}
+	cfg := yamlstore.New(c.fs)
+	if err := cfg.LoadFrom(path); err != nil {
+		return err
+	}
+
+	return c.getInfoFromConfig(cfg.Config())
+}
```
camilamacedo86, line 230: The project has interfaces for we are able to load and parse the PROJECT file to its structure already, see: 

https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/config/v3/config.go
https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/config/store/interface.go

pkg/cli/scaffold.go
```diff
@@ -0,0 +1,107 @@
...
+func (c *CLI) newScaffoldCmd() *cobra.Command {
+	var projectConfig, outputPath string
+	scaffoldCmd := &cobra.Command{
+		Use:   "scaffold",
+		Short: "Re-scaffold an existing kuberbuilder project",
+		// TODO: Better description
+		Long: `TODO`,
+		PreRunE: func(cmd *cobra.Command, _ []string) error {
+			if outputPath != "" {
+				if _, err := os.Stat(outputPath); os.IsNotExist(err) {
```
camilamacedo86, line 38: WDYT about use methods to implement it, such as validate ?

pkg/cli/scaffold.go
```diff
@@ -0,0 +1,107 @@
...
+				_ = os.Chdir(cwd)
+			}()
+			// init project with plugins
+			// call edit subcommands
+
+			// call create apis
+
+			// call create webhook
+
+			return nil
```
camilamacedo86, line 80: I think this operations must be inside of methods as well

pkg/cli/cli.go
rumstead: That variable gets populated by the project file [here](https://github.com/rumstead/kubebuilder/blob/e7ee3941b34e3ab56362dfaa6f945bfb15a75f8b/pkg/cli/scaffold.go#L56). 

Can I access the project name without exposing a new variable here?

pkg/cli/alpha.go
rumstead: If we want to reuse the functions that init, create, edit, etc us, we cannot put the command in the alpha package use `CLI`. If we do the latter, I have to add the command here. 

Thoughts on how to handle init, create, edit? The `CLI` struct has a lot of the logic already.

rumstead: If we want to reuse the functions that init, create, edit, etc us, we cannot put the command in the alpha package and instead use the `cli` package. If use the latter package, I wouldn't be able to add here given I need an instance of `CLI`. 

Thoughts on how to proceed with reusing the logic in init, create, and edit?

camilamacedo86: The code here will add as alpha any new command added to that list.
Because of that we just set it in the list. 


> Thoughts on how to proceed with reusing the logic in init, create, and edit?

We can indeed exec the kubebuilder commands as we do in the e2e tests.
That will be like `cmd = exec.Command("kubebuilder", "init")` for example

pkg/plugin/util/testdata/exampleFile.txt
```diff
@@ -1 +1 @@
-exampleTarget
\ No newline at end of file
+exampleTargetexampleCode
```
camilamacedo86, line 1: It does not seems part of the scope of this PR

pkg/cli/scaffold.go
```diff
@@ -0,0 +1,108 @@
+/*
```
camilamacedo86, line 1: Hi @rumstead,
 
We need to create under alpha a dir for the new command. 
I would suggest that to be regen
See a draft about how I thought that we could move achieve this goal: https://github.com/kubernetes-sigs/kubebuilder/pull/3007

rumstead:
@camilamacedo86 before I go and write tests I wanted to get your take on the structure, flow, and approach.

pkg/cli/alpha/scaffold.go
```diff
@@ -0,0 +1,257 @@
...
+
+	for _, w := range webhooks {
+		if err = createWebhook(w); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func createAPI(resource resource.Resource) error {
```
rumstead, line 125: Instead of having these methods return an error from `util.RunCmd` I may have them return an array of strings so I can unit test easier.

pkg/cli/alpha/scaffold.go
```diff
@@ -0,0 +1,257 @@
...
+		PreRunE: func(cmd *cobra.Command, _ []string) error {
+			return validate(projectConfig, outputPath)
+		},
+		Run: func(cmd *cobra.Command, args []string) {
+			cwd, err := os.Getwd()
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			projectPath := getProjectPath(cwd, projectConfig)
```
camilamacedo86, line 57: We need to copy the current project to an old-<project> name dir so that the users can compare both and add the code on top of the project that was re-scaffold

pkg/cli/alpha/scaffold.go
```diff
@@ -0,0 +1,257 @@
...
+		return util.RunCmd("kubebuilder edit", "kubebuilder", args...)
+	}
+	return nil
+}
+
+func kubebuilderInit(store store.Store) error {
+	var args []string
+	args = append(args, "init")
+	args = append(args, generateInitArgs(store)...)
+	return util.RunCmd("kubebuilder init", "kubebuilder", args...)
```
camilamacedo86, line 202: We need to check all plugins used to initiate the project and use them if we found them tracked in the PROJECT file.
Also, if the plugin has data tracked we need to use them to pass the flags with the respective values.

See for example the case of we use the deploy image p;lugin.
You can use the samples under the test data to check it out.

pkg/cli/alpha/scaffold.go
```diff
@@ -0,0 +1,257 @@
...
+
+func generateInitArgs(store store.Store) []string {
+	var args []string
+	// add existing plugins
+	plugins := store.Config().GetPluginChain()
+	if len(plugins) > 0 {
+		args = append(args, "--plugins")
+	}
+	for i, _ := range plugins {
+		args = append(args, plugins[i])
```
camilamacedo86, line 213: could we solve the lint issues ^ ?

pkg/cli/alpha/scaffold.go
rumstead: this would create a subdirectory with a re-scaffolded project, why would we need to touch the old project?

pkg/cli/alpha/scaffold.go
rumstead: `generateInitArgs` does loop over the plugin chain. So you are saying supply the `options` to the plugin as well?

pkg/cli/alpha/scaffold.go
camilamacedo86: As a user, we create the project and add code on top.
So that to migrate the project we need to re-scaffold all and then compare the "old project" with the new scaffold so that we can re-add all code done on top again. 

So, ihmo we need to do here: 

a) copy the current project to another dir ( i.e. cp -R * my-operator/ old-my-operator/ )
b) then, in the current dir we will remove all files rm -rf *
c) after that we can re-scaffold all so that a user can use git OR compare my-operator/ with old-my-operator/ in order to add the code again.

pkg/cli/alpha/scaffold.go
camilamacedo86: Note that when we initiate a project we can use plugins and indeed its flags. (i.e `kubebuilder init --plugins="go/v4-alpha,declarative,grafana/v1-alpha"`)

Also, when we run create API we can also use a plugin and pass values/flags (i.e.  ` kubebuilder create api --group example.com --version v1alpha1 --kind Memcached --image=memcached:1.4.36-alpine --image-container-command="memcached,-m=64,-o,modern,-v" --image-container-port="11211" --run-as-user="1001" --plugins="deploy-image/v1-alpha"`  )

So when we re-scaffold we need to check in the PROJECT file the plugins 
a) used in the layout, see: https://github.com/kubernetes-sigs/kubebuilder/blob/8bd569de6745d94249a526319f921bbb4f776bb2/testdata/project-v3-addon-and-grafana/PROJECT#L2-L5
b) used to create the apis, see: 
- i.e with declarative: plugins:
  declarative.go.kubebuilder.io/v1:
    resources:
    - domain: testproject.org
      group: crew
      kind: Captain
      version: v1
    - domain: testproject.org
      group: crew
      kind: FirstMate
      version: v1
    - domain: testproject.org
      group: crew
      kind: Admiral
      version: v1
  grafana.kubebuilder.io/v1-alpha: {}

- i.e. with deploy image: https://github.com/kubernetes-sigs/kubebuilder/blob/8bd569de6745d94249a526319f921bbb4f776bb2/testdata/project-v3-with-deploy-image/PROJECT#L4-L21

camilamacedo86: But we can try to make it as simple as possible. If the implementation is able to work with the deploy image, (see the sample https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v3-with-deploy-image generate ) then, it can work in the biggest part of the use cases.

camilamacedo86:
HI @rumstead,

We created a design proposal doc to move forward with this one. See:  https://github.com/kubernetes-sigs/kubebuilder/issues/3244

Also, we decide to apply this one for the GS0C 2023. If accepted, a student can work on this and move forward with the work you began to try to address.

So, I hope you do not mind if we close this one.  Who will be able to work on this one will still able to check this PR and see how we can use it to help out move forward. Please, feel free to reach out via slack if you need.

Sajiyah-Salat:
Hey @camilamacedo86 I think you should add gsoc label if this is reserved for gsoc participants.

PR #2898, opened 2022-08-26
🌱  deploy-image/v1-alpha e2e tests: fix the value of label app.kubernetes.io/name e2e tests use the Kind without be in lower case

NikhilSharmaWe:
### Description
In the e2e tests for `deploy-image/v1-alpha`, label `app.kubernetes.io/name` for pod and deployment for the Kind looks for value `string.ToLower(kbc.Kind)` but it should be only `kbc.Kind`. This PR is made to fix that.

See https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3-with-deploy-image/controllers/memcached_controller.go#L385 for example, we add `Memcached` not `memcached` to label `app.kubernetes.io/name`.

### DIFF
--- a/test/e2e/deployimage/plugin_cluster_test.go
+++ b/test/e2e/deployimage/plugin_cluster_test.go
@@ -253,7 +253,7 @@ func Run(kbc *utils.TestContext) {
 	By("validating that pod(s) status.phase=Running")
 	getMemcachedPodStatus := func() error {
 		status, err := kbc.Kubectl.Get(true, "pods", "-l",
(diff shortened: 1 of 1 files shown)


camilamacedo86 (commented):
@NikhilSharmaWe this one is only in the tests so does not affect the end-users.
In this case, the emoji that we use is 🌱 . So that, it will not be used in the changelog.

Otherwise, all fine.

camilamacedo86 (approved):

PR #2900, opened 2022-08-29
:book: feat(docs): update envtest docs

### LINKED ISSUE #2389: EnvTest setup docs need to be updated

rumstead:
# Description 

The controller runtime allows us to check out the binary assets to [an explicit location ](https://github.com/kubernetes-sigs/controller-runtime/blob/bcde6f084dd15ef93bee2a2e4668184e7472eddd/tools/setup-envtest/main.go#L50)[instead of specific locations on each OS](https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest#where-does-it-put-all-those-binaries). Updating the envtest docs to specify the new binary location defined in the [Makefile](https://github.com/kubernetes-sigs/kubebuilder/blob/2e93b289979c226ebd499a3e64ee6566ca5c12a2/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go#L118). 

Closes #2389

### DIFF
--- a/docs/book/src/reference/envtest.md
+++ b/docs/book/src/reference/envtest.md
@@ -1,45 +1,48 @@
 # Configuring envtest for integration tests
 
-The [`controller-runtime/pkg/envtest`][envtest] Go library helps write integration tests for your controllers by setting up and starting an instance of etcd and the Kubernetes API server, without kubelet, controller-manager or other components.
+The [`controller-runtime/pkg/envtest`][envtest] Go library helps write integration tests for your controllers by setting up and starting an instance of etcd and the 
+Kubernetes API server, without kubelet, controller-manager or other components.
 
 ## Installation
 
-The `test` make target, also called by the `docker-build` target,
-[downloads][setup-envtest] a set of envtest binaries (described above) to run tests with.
-Typically nothing needs to be done on your part,
-as the download and install script is fully automated,
-although it does require `bash` to run.
+Installing the binaries is as a simple as running `make envtest`. `envtest` will download the Kubernetes API server binaries to the `bin/` folder in your project
+by default. `make test` is the one-stop shop for downloading the binaries, setting up the test environment, and running the tests. 
 
-If you would like to download the tarball containing these binaries,
-to use in a disconnected environment for example,
-run the following (Kubernetes version 1.21.2 is an example version):
+The make targets require `bash` to run. 
 
-```sh
-export K8S_VERSION=1.21.2
-curl -sSLo envtest-bins.tar.gz "https://go.kubebuilder.io/test-tools/${K8S_VERSION}/$(go env GOOS)/$(go env GOARCH)"
+## Installation in Air Gaped/disconnected environments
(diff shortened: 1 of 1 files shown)


docs/book/src/reference/envtest.md
```diff
@@ -77,7 +77,17 @@ Logs from the test runs are prefixed with `test-env`.
...
+For projects built with plugin v3+ (see your PROJECT file's `layout` key), the `make test` command will install these binaries to the `bin/` directory and use them when running tests that use `envtest`.
+Ie,
+```shell             
+./bin/k8s/
+└── 1.24.2-darwin-amd64
+    ├── etcd
+    ├── kube-apiserver
+    └── kubectl
+
+1 directory, 3 files
```
camilamacedo86, line 89: @rumstead that is the default behaviour so I think we need to put on top of the doc and we no longer need to say with the plugin go/v3.

Therefore, we need to recheck the current info in this doc since it seems outdatate.

Could you please help us with?
WDYT? is that make sense?

rumstead: certainly can help and makes sense

docs/book/src/reference/envtest.md
```diff
@@ -77,7 +76,17 @@ Logs from the test runs are prefixed with `test-env`.
 
 Controller-runtime’s [envtest][envtest] framework requires `kubectl`, `kube-apiserver`, and `etcd` binaries be present locally to simulate the API portions of a real cluster.
 
-For projects built with plugin v3+ (see your PROJECT file's `layout` key), the `make test` command will install these binaries to the `testbin/` directory and use them when running tests that use `envtest`.
+The `make test` command will install these binaries to the `bin/` directory and use them when running tests that use `envtest`.
```
camilamacedo86, line 79: That is great, just one nit. I think we need to update this doc to order the info better. Note that before it was required to manually get the binaries always which is no longer the case.

So, wdyt:
a) we let the users know that they can run make test and it will call the target make setup-envtest that will install the required binaries under the bin directory by default
b) after we explain the default behaviour we can add a section like "how to customize the location where the envtest binaries will be installed"
c) then, inside of this subsection we describe how to change the arg and also we describe how it needs/can be set up in the test code.

WDYT?

camilamacedo86 (approved):
/ok-to-test

It shows great for me 🥇 
So, 

/approved

Let's also ask others to give a look and review this one. 

varshaprasad96 (approved):
Thanks for the PR @rumstead !

PR #2901, opened 2022-08-30
:sparkles: Setup multi-module alpha plugin

jakobmoellersap:
Adds a new Plugin for multiple modules per API instead of a typical mono-module repository.

Follow-Up from https://github.com/kubernetes-sigs/kubebuilder/pull/2880

Workflow:

```
kubebuilder init --plugins multi-module/v1-alpha --repo github.com/jakobmoellersap/someothertest.com 
```
or 
```
kubebuilder init --repo 
kubebuilder edit --plugins multi-module/v1-alpha --multimodule=true
```

```
kubebuilder create api --version v1alpha1 --kind TestKind --group group
```

New go.mod in `api/go.mod`:
```go.mod
module github.com/jakobmoellersap/someothertest.com/api

go 1.18

require (
        k8s.io/apimachinery v0.24.2
        sigs.k8s.io/controller-runtime v0.12.2
)

require (
        github.com/go-logr/logr v1.2.0 // indirect
        github.com/gogo/protobuf v1.3.2 // indirect
        github.com/google/gofuzz v1.1.0 // indirect
        github.com/json-iterator/go v1.1.12 // indirect
        github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
        github.com/modern-go/reflect2 v1.0.2 // indirect
        golang.org/x/net v0.0.0-20220127200216-cd36cc0744dd // indirect
        golang.org/x/text v0.3.7 // indirect
        gopkg.in/inf.v0 v0.9.1 // indirect
        gopkg.in/yaml.v2 v2.4.0 // indirect
        k8s.io/klog/v2 v2.60.1 // indirect
        k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9 // indirect
        sigs.k8s.io/json v0.0.0-20211208200746-9f7c6b3444d2 // indirect
        sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
)
```

Adjusted main `go.mod`:

```go.mod
module github.com/jakobmoellersap/someothertest.com

go 1.18

require (
        github.com/onsi/ginkgo v1.16.5
        github.com/onsi/gomega v1.18.1
        k8s.io/apimachinery v0.24.2
        k8s.io/client-go v0.24.2
        sigs.k8s.io/controller-runtime v0.12.2
)

require (
        github.com/nxadm/tail v1.4.8 // indirect
        gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
)

require (
        cloud.google.com/go v0.81.0 // indirect
        github.com/Azure/go-autorest v14.2.0+incompatible // indirect
        github.com/Azure/go-autorest/autorest v0.11.18 // indirect
        github.com/Azure/go-autorest/autorest/adal v0.9.13 // indirect
        github.com/Azure/go-autorest/autorest/date v0.3.0 // indirect
        github.com/Azure/go-autorest/logger v0.2.1 // indirect
        github.com/Azure/go-autorest/tracing v0.6.0 // indirect
        github.com/PuerkitoBio/purell v1.1.1 // indirect
        github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 // indirect
        github.com/beorn7/perks v1.0.1 // indirect
        github.com/cespare/xxhash/v2 v2.1.2 // indirect
        github.com/davecgh/go-spew v1.1.1 // indirect
        github.com/emicklei/go-restful v2.9.5+incompatible // indirect
        github.com/evanphx/json-patch v4.12.0+incompatible // indirect
        github.com/form3tech-oss/jwt-go v3.2.3+incompatible // indirect
        github.com/fsnotify/fsnotify v1.5.1 // indirect
        github.com/go-logr/logr v1.2.0 // indirect
        github.com/go-logr/zapr v1.2.0 // indirect
        github.com/go-openapi/jsonpointer v0.19.5 // indirect
        github.com/go-openapi/jsonreference v0.19.5 // indirect
        github.com/go-openapi/swag v0.19.14 // indirect
        github.com/gogo/protobuf v1.3.2 // indirect
        github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
        github.com/golang/protobuf v1.5.2 // indirect
        github.com/google/gnostic v0.5.7-v3refs // indirect
        github.com/google/go-cmp v0.5.5 // indirect
        github.com/google/gofuzz v1.1.0 // indirect
        github.com/google/uuid v1.1.2 // indirect
        github.com/imdario/mergo v0.3.12 // indirect
        github.com/jakobmoellersap/someothertest.com/api/v1alpha1 v0.0.0-v1alpha1
        github.com/josharian/intern v1.0.0 // indirect
        github.com/json-iterator/go v1.1.12 // indirect
        github.com/mailru/easyjson v0.7.6 // indirect
        github.com/matttproud/golang_protobuf_extensions v1.0.2-0.20181231171920-c182affec369 // indirect
        github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
        github.com/modern-go/reflect2 v1.0.2 // indirect
        github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
        github.com/pkg/errors v0.9.1 // indirect
        github.com/prometheus/client_golang v1.12.1 // indirect
        github.com/prometheus/client_model v0.2.0 // indirect
        github.com/prometheus/common v0.32.1 // indirect
        github.com/prometheus/procfs v0.7.3 // indirect
        github.com/spf13/pflag v1.0.5 // indirect
        go.uber.org/atomic v1.7.0 // indirect
        go.uber.org/multierr v1.6.0 // indirect
        go.uber.org/zap v1.19.1 // indirect
        golang.org/x/crypto v0.0.0-20220214200702-86341886e292 // indirect
        golang.org/x/net v0.0.0-20220127200216-cd36cc0744dd // indirect
        golang.org/x/oauth2 v0.0.0-20211104180415-d3ed0bb246c8 // indirect
        golang.org/x/sys v0.0.0-20220209214540-3681064d5158 // indirect
        golang.org/x/term v0.0.0-20210927222741-03fcf44c2211 // indirect
        golang.org/x/text v0.3.7 // indirect
        golang.org/x/time v0.0.0-20220210224613-90d013bbcef8 // indirect
        gomodules.xyz/jsonpatch/v2 v2.2.0 // indirect
        google.golang.org/appengine v1.6.7 // indirect
        google.golang.org/protobuf v1.27.1 // indirect
        gopkg.in/inf.v0 v0.9.1 // indirect
        gopkg.in/yaml.v2 v2.4.0 // indirect
        gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
        k8s.io/api v0.24.2 // indirect
        k8s.io/apiextensions-apiserver v0.24.2 // indirect
        k8s.io/component-base v0.24.2 // indirect
        k8s.io/klog/v2 v2.60.1 // indirect
        k8s.io/kube-openapi v0.0.0-20220328201542-3ee0da9b0b42 // indirect
        k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9 // indirect
        sigs.k8s.io/json v0.0.0-20211208200746-9f7c6b3444d2 // indirect
        sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
        sigs.k8s.io/yaml v1.3.0 // indirect
)

replace github.com/jakobmoellersap/someothertest.com/api=> ./api
```

Resulting Dockerfile:
```Dockerfile
# Build the manager binary
FROM golang:1.18 as builder

WORKDIR /workspace
# Copy the Go Modules manifests
COPY go.mod go.mod
COPY go.sum go.sum
# Copy the Go Sub-Module manifests
COPY api/v1alpha1/go.mod api/go.mod
COPY api/v1alpha1/go.sum api/go.sum
# cache deps before building and copying source so that we don't need to re-download as much
# and so that source changes don't invalidate our downloaded layer
RUN go mod download

# Copy the go source
COPY main.go main.go
COPY api/ api/

# Build
RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go

WORKDIR /
# copy channels
COPY --from=builder /channels /channels

USER 65532:65532

ENTRYPOINT ["/manager"]
```

Open Todos:

- [x] Clarify replace directive in main module => could be done via go.work instead
- [ ] remove apimachinery / controller-runtime dependencies if possible
- [ ] remove / move webhook dependencies if possible
- [x] Figure out why main go.mod is so weirdly formatted => this comes from the official `go edit` commands and I did not expect that :D however I think it is cleaner than setting up our own scaffolding for these files.
- [x] Verify with multi-group (should work out of the box) => it works
- [ ] Maybe create a go.work with .gitignore support for the replace statement instead of a replace in go.mod as well as GOWORK=off for CI-based builds to ensure build integrity even though there is a go.work file
- [X] Write Tests => e2e written
- [x] Write Docs


jakobmoellersap:
@camilamacedo86 would need an ok-to-test for this one to get started.

everettraven:
@jakobmoellersap I will try to get around to looking at this sometime today and leaving a review. Thanks for all the great work on this!

varshaprasad96:
@jakobmoellersap Thanks for the PR! I have a few questions related to this:
Having multiple workspaces is a great feature by Go. But I am wondering the complications of having multiple modules inside an operator project. If we would like to APIs outside the project, why not import it the entire project. 

Alternatives to creating separate go mod files would be:
1. To import the entire project as a module and use this project as a workspace.
2. To remove the APIs from here, and make them external in a separate project instead of creating multiple submodules.

Having multiple modules adds value when each sub module is expected to have a different lifecycle. The evolution of types defined in `api/` is very tightly related to changes happening in other parts of the project. We would not like to encourage isolated modification of API types. If there is a use case where a single API is being used in multiple different operator projects (in the sense separate go repositories), a better approach would be make the APIs external. That way the versioning cycle of API and the respective changes in it could be tracked separately.

jakobmoellersap:
@varshaprasad96 Thanks for taking the time to dive deeper here!

First of all: I share your sentiment and went through pretty much the same thought process while working on issue #2627, however I tend to now support multi-modules for a few reasons.

I want to give some more context on to why multi-module support is beneficial for kubebuilder:

First comes the question "why do we need separate go modules at all?":

The Separation of API and main module is especially important when scaling kubebuilder based projects with in-code type dependencies in other go projects. Specifically for projects that are enterprise-grade, a lot of dependencies usually means headaches in regards to triaging security vulnerabilities as well as keeping compatibility between the introduced modules (even though these projects can be multiple minors behind upstream with their imported packages), so reducing the transitive dependencies is paramount even when integrating with other projects / types. This significantly reduces downstream effort in managing `replace` or `exclude` directives necessary to deal with conflicting dependencies (e.g. due to a malformed upgrade or a conflict through MVS)

Even though it is a special use case (which we would acknowledge by creating an additional plugin instead of natively integrating it into the go plugin), it is even recognized by the go documentation as a valid scenario. From https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository:

> if you have a repository with a complex set of dependencies, but you have a client API with a smaller set of dependencies. In some cases, it might make sense to have an api or clientapi or similar directory with its own go.mod, or to separate out that clientapi into its own repository.

The caveat of when not to consider multiple modules that is mentioned afterwards mainly targets performance reasons (why it is not useful to exclude transitive dependencies purely to save on build resolution), but does not take into account the aforementioned reasons on why a clean dependency tree is beneficial.

To the second question: "Why not use an external Repository with the API Types in them?":

1. We would not have to rely on https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/using_an_external_type.md#edit-the-api-files for integrating a separate repository just for separating out the go.mod file that is holding the API. Instead it would work natively and we can offer a tightly coupled, clean development lifecycle while keeping the dependencies small when importing the API. On top of this we actually support development of the two projects closely together.
2. Maintaining two repositories is arguably more effort than maintaining one repository with multiple modules as long as (and thats important) the tooling in a mono-repo effortlessly supports the multiple modules and there is no argument to be made to introduce the modules across multiple repositories in your own domain. For example, one can easily issue `replace` directives with `go.work` or `go.mod` to work on both projects simultaneously without having to manage multiple repositories in VCs

To your argument that the go.mod separation only makes sense for separate lifecycles: 

> We would not like to encourage isolated modification of API types

I would argue that even though kubebuilder has tight integration of API and controllers, its goal in actuality is to make sure that APIs can be iterated and incremented on separately from the controller implementation. Not because you want to isolate the API modification, but because you want to isolate the controller modification. Mainly this is because the Kubernetes Type Versioning is meant to become stable (apart from frequent iteration in the early alpha stages), while controller can stay dynamic in what it implements against the API. This of course means that in practice, kubebuilder projects tend to iterate much less on the API Types than on the controller implementation once their API reaches a certain stability. This is mainly because the API Types serve as a contract with the API Server through the CRD as well as to in-code consumers as a dependency. 

At the same time, you still want to release the API types with incrementing versions of your code in the controller (similar to how kubernetes deals with apimachinery, see https://github.com/kubernetes/apimachinery#compatibility), which tends to be much easier in Mono-Repo setups than in multiple repositories. This is because that way, downstream consumers can easily upgrade the release version and know that the compatibility is still achieved as long as the module versions of both modules match.

Even if you think all of the above is still not applying or relevant to you (which for smaller operators it is most likely the case), you do not even have to change the kubebuilder commands and it will still work as before due to the plugin nature.

I hope I gave some context here on why I think this can be useful!

pkg/plugins/golang/multi-module/v1alpha1/api.go
```diff
@@ -0,0 +1,151 @@
...
+
+If information about whether the resource and controller should be scaffolded
+was not explicitly provided, it will prompt the user if they should be.
+
+After the scaffold is written, the dependencies will be updated and
+make generate will be run.
+
+Warning: This will also create multiple go.mod files. If you are not careful, you can break your dependency chain.
+The multi-module extension will create replace directives for local development, 
+which you might want to drop after creating your first stable API.
```
camilamacedo86, line 51: @jakobmoellersap you can create only once right?
We can check and only create if not exist

pkg/plugins/golang/multi-module/v1alpha1/plugin.go
```diff
@@ -0,0 +1,67 @@
...
+	pluginVersion            = plugin.Version{Number: 1, Stage: stage.Alpha}
+	supportedProjectVersions = []config.Version{cfgv3.Version}
+	pluginKey                = plugin.KeyFor(Plugin{})
+)
+
+var _ plugin.Init = Plugin{}
+
+// Plugin implements the plugin.Full interface
+type Plugin struct {
+	initSubcommand
```
camilamacedo86, line 39: I think we need to implement the edit as well so that users are able to edit a pre-existe project to enable it.

pkg/plugins/golang/multi-module/v1alpha1/plugin.go
```diff
@@ -0,0 +1,67 @@
...
+type pluginConfig struct {
+	Resources []resourceData `json:"resources,omitempty"`
+}
+
+type resourceData struct {
+	Group   string `json:"group,omitempty"`
+	Domain  string `json:"domain,omitempty"`
+	Version string `json:"version"`
+	Kind    string `json:"kind"`
+}
```
camilamacedo86, line 67: Can I have 1 api using the option of multi modules and others not? 
If not, then I think we do not need to track the resources, if the project was scaffolded with the plugin then all apis will use it.

cmd/main.go
```diff
@@ -32,13 +32,13 @@ import (
 	"sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang"
 	declarativev1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/declarative/v1"
 	deployimagev1alpha1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/deploy-image/v1alpha1"
+	multimodulev1alpha1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/multi-module/v1alpha1"
 	golangv2 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v2"
```
camilamacedo86, line 36: could we add a sample test for this one under the testdata sample?
Also, could we add an e2e test to verify this plugin?

pkg/plugins/golang/multi-module/v1alpha1/api.go
jakobmoellersap: Is that not desired? I would imagine if you create APIs you do not want to switch between multi-module or not. My idea would have been that it would only need to be setup once in the init. The go.mod is using create if not exist with `f.IfExistsAction = machinery.SkipFile`

pkg/plugins/golang/multi-module/v1alpha1/plugin.go
jakobmoellersap: I think this should be out of scope since its mostly confusing and I cannot see real value-add here, I will drop the tracking.

pkg/plugins/golang/multi-module/v1alpha1/api.go
camilamacedo86: So, are you saying that is possible to have a project that is multi-module and not export all apis? Should not all apis be replaced in this case?

jakobmoellersap: Ehm actually I tried to say the opposite :D sorry for my bad wording here. This is also why im gonna remove the resource from the project config, I think it should be an `all-or-nothing` for the APIs

test/testdata/generate.sh
```diff
@@ -132,14 +132,14 @@ scaffold_test_project project-v2 --project-version=2
...
-scaffold_test_project project-v3-addon-and-grafana --plugins="go/v3,declarative,grafana/v1-alpha"
+scaffold_test_project project-v3-addon-and-grafana --plugins="go/v3,declarative,multi-module,grafana/v1-alpha"
 scaffold_test_project project-v3-config --component-config
 scaffold_test_project project-v3-with-deploy-image
 
 # [Next version, alpha] Project version v4-alpha
 scaffold_test_project project-v4 --plugins="go/v4-alpha"
 scaffold_test_project project-v4-multigroup --plugins="go/v4-alpha"
-scaffold_test_project project-v4-addon-and-grafana --plugins="go/v4-alpha,declarative,grafana/v1-alpha"
+scaffold_test_project project-v4-addon-and-grafana --plugins="go/v4-alpha,declarative,multi-module,grafana/v1-alpha"
```
camilamacedo86, line 142: @jakobmoellersap I think we need a new dirs/lines here 

scaffold_test_project project-v4-multi-module --plugins="go/v4-alpha,multi-module"

So that we can check the result of this one only

Also, we will need that for go/v3 and you need to change the above lines to generated the same apis and etc for this scenario

You can run make generate to get it all done.

cmd/main.go
jakobmoellersap: Well turns out 14k lines is quite a lot :D

pkg/plugins/golang/multi-module/v1alpha1/plugin.go
jakobmoellersap: also needed to implement webhook as it could be that the go.mod was outdated due to the webhook introducing more dependencies into the api folder. Btw I think this should really be something to think about in future plugins. I do not think that we should put the Webhooks into the API folder, rather something like a machinery or into the controller.

test/testdata/generate.sh
```diff
@@ -102,6 +102,34 @@ function scaffold_test_project {
       $kb create api --version v1 --kind Lakers --controller=true --resource=true --make=false
       $kb create webhook --version v1 --kind Lakers --defaulting --programmatic-validation
     fi
+  elif [[ $project =~ multimodule ]]; then
+    header_text 'Switching to multimodule & multigroup layout ...'
+    $kb edit --multimodule=true --multigroup=true
+
+    header_text 'Creating APIs ...'
```
jakobmoellersap, line 109: Probably we dont need all of these cases, but I tried to test all kinds of scenarios. Please let me know how much you actually want me to test here.

pkg/plugins/golang/multi-module/v1alpha1/api.go
camilamacedo86: Hi @jakobmoellersap, 

Cool. So, if it is all all-or-nothing then we should not need to explicitly call the plugin every time that we run create api. We should only:

a) Call the plugin for init or edit so we enable that
b) Therefore, after that when we call create api it will be tracked in the plugin layout (such as we have the declarative one) which means that when we run create api the CLI will automatically  call all create api subcommand implemented for all plugin in the layout in a chain

Also, that means we do not need to track the API resources for this plugin since it has no specific behaviour for each API. That is valid for the whole project after you use that.

By last in case we use the edit command, we might want to try to change the default scaffold with what was done before to fix all. However, if an error is faced would be great if we raise that in the logs and let users know what they need to do manually. 

Is that make sense?

test/testdata/generate.sh
camilamacedo86: We do not need to add the new if:
We just need to add the [[ project-v3-multimodule ]] and for [[ project-v4-multimodule ]]in the line: https://github.com/kubernetes-sigs/kubebuilder/blob/b9d0e8c764d8bfb2a18c4e25210b19ee186759d5/test/testdata/generate.sh#L48 (so that we will have the same scaffold done for the other mocks but using this layout )

Then, at the bottom of this file, you will add the lines to create both projects passing this new plugin as it is done. 

Also, we need to add the mock to the tests made with them, here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/test.sh#L33-L45

(it will call make targets to verify the scafofld)

test/e2e/multimodule/e2e_suite_test.go
```diff
@@ -0,0 +1,32 @@
+/*
```
camilamacedo86, line 1: @jakobmoellersap, 

For the new plugin test be called we need to add it in : https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/setup.sh#L65

test/e2e/multimodule/e2e_suite_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+// Run e2e tests using the Ginkgo runner.
+func TestE2E(t *testing.T) {
+	RegisterFailHandler(Fail)
+	fmt.Fprintf(GinkgoWriter, "Starting kubebuilder suite\n")
```
camilamacedo86, line 30:
```suggestion
	fmt.Fprintf(GinkgoWriter, "Starting multimodule v1alpha plugin kubebuilder suite\n")
```

test/e2e/multimodule/e2e_suite_test.go
```diff
@@ -0,0 +1,32 @@
...
+	RunSpecs(t, "Kubebuilder e2e suite")
```
camilamacedo86, line 31:
```suggestion
	RunSpecs(t, "Kubebuilder e2e suite for multimodule v1alpha plugin")
```

testdata/project-v3-multimodule/PROJECT
```diff
@@ -0,0 +1,109 @@
+domain: testproject.org
+layout:
+- go.kubebuilder.io/v3
+- multi-module.go.kubebuilder.io/v1-alpha
+multigroup: true
+plugins:
+  multi-module.go.kubebuilder.io/v1-alpha:
+    ApiGoModCreated: true
```
camilamacedo86, line 8: We do not need to track the resources for this one IHMO
If we enable that then it should be valid for the whole project

pkg/plugins/golang/multi-module/v1alpha1/commons.go
```diff
@@ -0,0 +1,63 @@
...
+package v1alpha1
+
+import (
+	"errors"
+	"path/filepath"
+
+	"sigs.k8s.io/kubebuilder/v3/pkg/config"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin/util"
+)
+
```
camilamacedo86, line 26: Should this method be used outside of the plugin? I do not think so, so we should not export them. 
Also, do we need the commons here or could we leave the implementation where it is used?

pkg/plugins/golang/multi-module/v1alpha1/dockerfile.go
```diff
@@ -0,0 +1,77 @@
...
+		"\nCOPY %s %s"+
+		"\nCOPY %s %s",
+		goModPath, goModPath, goSumPath, goSumPath)
+
+	return removeCodeIfExist(dockerfile, toRemove)
+}
+
+// insertCodeIfDoesNotExist insert code if it does not already exists
+func insertCodeIfDoesNotExist(filename, target, code string) error {
+	if err := util.InsertCode(filename, target, code); err != nil && err != util.ErrContentNotFound {
```
camilamacedo86, line 65: Why not call it in the plugin methods? Do we really a new method `insertCodeIfDoesNotExist ` to only call ` util.InsertCode` for example?

pkg/plugins/golang/multi-module/v1alpha1/gomod.go
```diff
@@ -0,0 +1,92 @@
...
+	"sigs.k8s.io/kubebuilder/v3/pkg/config"
+	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
+	"sigs.k8s.io/kubebuilder/v3/pkg/plugin/util"
+	v3scaffolds "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v3/scaffolds"
+)
+
+const (
+	ControllerRuntime        = "sigs.k8s.io/controller-runtime"
+	ControllerRuntimeVersion = v3scaffolds.ControllerRuntimeVersion
+	GoVersion                = "1.18"
```
camilamacedo86, line 18: Why this plugin needs the C+R version and the Go one?
Can not we try to do something here that will make it eaiser if we keep it maintained? 

Why not use the same versions defined in the go plugin, see that for example the kustomize version is defined in the kustomize plugin ( it is part of its own domain of responsibility ) but the go base plugin needs to scaffold the makefile accordingly and use it: 

https://github.com/kubernetes-sigs/kubebuilder/blob/1502ede9731d67c591c8a9d3620d6a525ea49953/pkg/plugins/golang/v3/scaffolds/init.go#L104-L110

Can we not to do something similar here?

pkg/plugins/golang/multi-module/v1alpha1/api.go
```diff
@@ -0,0 +1,105 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
camilamacedo86, line 2: Could you please check the layout of the golang v3 plugin so we follow it here the same approach?

In the root, we would have only the subcommand call (V)
then, we need a dir called scaffolds where we have implementation for them
also, the boilerplates should be protected via an internal dir `[internal](https://github.com/kubernetes-sigs/kubebuilder/tree/1502ede9731d67c591c8a9d3620d6a525ea49953/pkg/plugins/golang/v3/scaffolds/internal)/templates/` so that nobody will import them.

jakobmoellersap: I deliberatly removed the scaffold since it cannot be used together with `go edit` commands safely. Right now I do not use any scaffold

jakobmoellersap: I think the way that the go.mod is created in the scaffold is not correct since it used a go template. Instead it should probably use either `go mod init` or the package "golang.org/x/mod/modfile".

camilamacedo86: Hi @jakobmoellersap ,

> I deliberatly removed the scaffold since it cannot be used together with go edit commands safely. Right now I do not use any scaffold

What do you mean with?
How create a directory scaffold and have inside of it the "controller (from MVC)" implementation on it can be an issue? 

> I think the way that the go.mod is created in the scaffold is not correct since it used a go template. Instead it should probably use either go mod init or the package "golang.org/x/mod/modfile".

We use the template because we want scaffold / template what should be in the go mod.
You can create a go.mod manually for any Golang application. You do not need to use go mod init at all. That is just a CLI helper.

jakobmoellersap: I agree that we could use a scaffolding here and in fact i did this in a previous commit... However I took inspiration from the comment from @everettraven to use the `go edit` commands to also extend them to `go inti` etc. Nevertheless, to stay consistent, I will adjust.

camilamacedo86: @jakobmoellersap you can use go edit that is fine. 
The suggestion is only about how/where you implement the structure of the plugin to ensure that consistency as you say.

pkg/plugins/golang/multi-module/v1alpha1/plugin.go
camilamacedo86: > also needed to implement webhook as it could be that the go.mod was outdated due to the webhook introducing more dependencies into the api folder.

I think that makes sense. How this approach would work with webhooks?


> Btw I think this should really be something to think about in future plugins. I do not think that we should put the Webhooks into the API folder, rather something like a machinery or into the controller.

The webhooks are related to the apis ( their purpose is for example convert an api ) therefore they do not fit in the controller.

pkg/plugins/golang/multi-module/v1alpha1/api.go
jakobmoellersap: For case B i spotted a difference in setting the layout with `init` vs `create api`. In create API, the plugin is not tracked in "layout" but only in "plugins". This means that you would have to continously call the plugin for every create call.

pkg/plugins/golang/multi-module/v1alpha1/plugin.go
jakobmoellersap: Currently the API Modules just incorporate more dependencies than if they were purely containing the api. If we want to deviate from that I would have to override the entire api scaffolding, probably not something we want

jakobmoellersap:
Two weird test errors now:
https://github.com/kubernetes-sigs/kubebuilder/runs/8160339818?check_suite_focus=true#step:587
This is a license check in v4alpha with deploy image that I did not touch

Lint is complaining about having no files to analyze, no idea why :O

Anyone can help me here?

Other than that I think I would put it ready for review and start writing docs.

docs/book/src/SUMMARY.md
```diff
@@ -39,7 +39,7 @@
   - [Hubs, spokes, and other wheel metaphors](./multiversion-tutorial/conversion-concepts.md)
   - [Implementing conversion](./multiversion-tutorial/conversion.md)
 
-      - [and setting up the webhooks](./multiversion-tutorial/webhooks.md)
+    - [and setting up the webhooks](./multiversion-tutorial/webhooks.md)
```
camilamacedo86, line 42: We need revert this changes since it change the numbers/layout of the docs for no related items

pkg/plugins/golang/multi-module/v1alpha1/api.go
camilamacedo86: But why we would call the plugin (informing it the --plugins create API)
Why we would need to track the resources inside of the plugin in this case ( that is only valid for the cases like deploy-image or declarative one where we can have 1 API using the plugin and another not )

camilamacedo86: I think we sorted out closing this one to make eaiser the review

pkg/plugins/golang/multi-module/v1alpha1/plugin.go
camilamacedo86: So, do you mean that if we want to use this plugin we cannot work with webhooks? 
is that? If so, we could clarify that in the docs and in have a subcommand createWebhook which would be called automatically and if so, we print a msg telling the users that and what needs to be done.

testdata/project-v3-multimodule/PROJECT
```diff
@@ -0,0 +1,52 @@
+domain: testproject.org
+layout:
+- go.kubebuilder.io/v3
+- multi-module.go.kubebuilder.io/v1-alpha
+plugins:
+  multi-module.go.kubebuilder.io/v1-alpha:
+    ApiGoModCreated: true
```
camilamacedo86, line 7:
```suggestion
    enable: true
```
Could we try to find a better name

camilamacedo86 (commented):
Hi @jakobmoellersap 

> Even though it is a special use case (which we would acknowledge by creating an additional plugin instead of natively integrating it into the go plugin), it is even recognized by the go documentation as a valid scenario. From https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository:

That is not the reason for the implementation being in a plugin acctually.
The reason we have a plugin in this case and not set in the go plugin is that not everybody would use that. And optional features, something that does not address the common cases should not be in the default layout.  

> I would argue that even though kubebuilder has tight integration of API and controllers

See: 
> Having many Kinds (such as CRDs) which are all managed by the same controller usually goes against the design proposed by [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime). Furthermore this might hurt concepts such as encapsulation, the Single Responsibility Principle, and Cohesion. Damaging these concepts may cause unexpected side effects, and increase the difficulty of extending, reusing, or maintaining the operator.

From : https://sdk.operatorframework.io/docs/best-practices/common-recommendation/

> its goal in actuality is to make sure that APIs can be iterated and incremented on separately from the controller implementation. Not because you want to isolate the API modification, but because you want to isolate the controller modification

If you have a Project that defines the APIs. Then, in another project you use these APIs as external types then you can have a controller for that as well.

However, I understand that the motivation for this plugin is: (Please let me know if I am wrong here)

I have an upstream/community version of my project and a downstream version of the same project. The downstream project (enterprise one) will use the same APIs because mainly do the same but also provide features on top. So, you would like to be able to import only the APIs defined in the upstream project (or the other side around) to make it simple to keep both projects maintained. 

In this case, it is not like project A owned and project B use them. Both projects are the owners of the APIs and you are looking for centralized code implementation. 

@jakobmoellersap 

You did incredible work here. 
I think the @varshaprasad96 thoughts are good too. 
So, what we need to do is discuss the solution and since it is a significant new feature I will just add a hold to allow everybody to review and add their thoughts on it. 

WDYT about presenting/demostrate the idea in the community meeting? 
That would also give a chance for everybody to raise their questions and you be able to clarify your points and motivations. 

c/c @varshaprasad96 

/hold

Just to allow us to have reviews and discuss it properly.

docs/book/src/SUMMARY.md
```diff
@@ -50,20 +50,22 @@
 
   - [Using a custom type](./component-config-tutorial/custom-type.md)
 
-      - [Adding a new Config Type](./component-config-tutorial/config-type.md)
-      - [Updating main](./component-config-tutorial/updating-main.md)
-      - [Defining your Custom Config](./component-config-tutorial/define-custom-config.md)
+    - [Adding a new Config Type](./component-config-tutorial/config-type.md)
+    - [Updating main](./component-config-tutorial/updating-main.md)
+    - [Defining your Custom Config](./component-config-tutorial/define-custom-config.md)
+
```
camilamacedo86, line 56: Same here we cannot remove the spaces that would mean these docs would no longer be subsections.

cmd/main.go
camilamacedo86, line 36: Question:

> kubebuilder init --repo 
kubebuilder edit --plugins multi-module/v1-alpha --multimodule=true
What happens if we do not pass the flag?
What happens if we set false instead?

docs/book/src/plugins/multi-module-v1-alpha.md
```diff
@@ -0,0 +1,77 @@
+# Multi Module Plugin (multi-module/v1-alpha)
+
+The multi-module plugin allows users to modify the existing scaffolding setup of kubebuilder to enable [Monorepo][monorepo] support. It abstracts the complexities to achieve this goal by easing the most difficult part of getting into Monorepos, namely the setup and the scaffolding required.
```
camilamacedo86, line 3: Would make sense if we call the plugin such as monorepo instead of multi-module?

docs/book/src/plugins/multi-module-v1-alpha.md
```diff
@@ -0,0 +1,77 @@
...
+See the "project-v3-multimodule" directory under the [testdata][testdata] directory of the Kubebuilder project to check an example of a scaffolding created using this plugin.
+
+</aside>
+
+## When to use it ?
+
+- If you want to reduce impact of transitive dependencies on your API being included in other projects
+- If you are looking to separately manage the lifecycle of your API release process from your controller release process.
+- If you are looking to modularize your codebase without splitting your code between multiple repositories.
+
```
camilamacedo86, line 24: What about we create a note here to clarify that scenarios like if you have Operator A and you want to use the APIs owned by it in Operator B you should consider using the external type instead. 

Also, would make sense if we describe that it would be recommended to use case scenarios where you acctually have more than one project to do the same. So both owners the API and provide a practical use case scenario. 

Example use case scenarios where would be recommended its usage:

If you have a community version of a project and an enterprise version of the same project. The enterprise version will use the same APIs mainly because does the same but also provide features on top. So, you would like to be able to import only the APIs defined in the upstream project (or the other side around) to make it simple to keep both projects maintained. 

In this case, it is not like project A owned and project B use them. Both projects are the owners of the APIs and you are looking for centralized code implementation. 

So that we do not encourage users to use it when they should not.

pkg/plugins/golang/multi-module/v1alpha1/plugin.go
camilamacedo86: Closing this one to try to sort it out.

docs/book/src/plugins/multi-module-v1-alpha.md
jakobmoellersap: Since the go documentation uses the terminology `multi-module` I think this is more fitting: https://go.dev/doc/tutorial/workspaces I can adjust the readme on this occasion

cmd/main.go
jakobmoellersap: if we do not pass the flag, it will count as false and it will disable it. This is due to
```
func (p *editSubcommand) BindFlags(fs *pflag.FlagSet) {
	fs.BoolVar(&p.multimodule, "multimodule", false, "enable or disable multimodule layout")
}
```
we could also default it to true, then we wouldnt need to pass the flag to enable it.

I think the flag is necessary to remove the multi-module support again when using edit. That way people can revert their changes. If we do not scope this in (e.g. by assuming people will revert by doing a git revert instead), then we can make the edit command easier, and I do not need to track the flag here.

pkg/plugins/golang/multi-module/v1alpha1/api.go
```diff
@@ -0,0 +1,114 @@
...
+
+	resource *resource.Resource
+
+	pluginConfig
+
+	// runMake indicates whether to run make or not after scaffolding APIs
+	runMake bool
+
+	apiPath string
+}
```
jmrodri, line 44: nit: is there a reason for all the whitespace between attributes?

jmrodri (commented):
Still in progress, but wanted to at least get my initial comments out there.

docs/book/src/plugins/multi-module-v1-alpha.md
```diff
@@ -0,0 +1,78 @@
...
+## How to use it ?
+
+After you create a new project with `kubebuilder init --plugins multi-module` you can create modularized APIs using this plugin. Ensure that you have followed the [quick start](https://book.kubebuilder.io/quick-start.html) before trying to use it.
```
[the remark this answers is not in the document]
camilamacedo86: I did not understood your question @jmrodri.
That is a plugin that can be used with go/v3 or go/v4-alpha howevever

And to use you can:

kubebuilder init --plugins go/v3,multi-module.
AND
kubebuilder init --plugins go/v4-alpha/multi-module <enable>=true (@jakobmoellersap is that right? we need to proper clarify the option as well)

OR
kubebuilder init 
kubebuilder edit --plugins multi-module --enable=true

OR 

kubebuilder init  --plugins go/v4
kubebuilder edit --plugins multi-module --enable=true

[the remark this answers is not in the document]
everettraven: I think that first sentence may be a tad misleading for some. 
It says:
> After you create a new project with `kubebuilder init --plugins multi-module`

This may convey the idea that the ONLY plugin you need to use is `multi-module` when using `kubebuilder init` but that is not the case. We may want to update this part to include the explicit usage information that @camilamacedo86 mentioned.

docs/book/src/plugins/multi-module-v1-alpha.md
```diff
@@ -0,0 +1,78 @@
...
+Therefore, before releasing your first version of your operator, make sure to [familiarize yourself with mono-repo/multi-module releases][multi-module-repositories] with multiple `go.mod` files in different subdirectories.
+
+Assuming a single API was created, the release process could look like this:
+
+```sh
+git commit
+git tag v1.0.0 # this is your main module release
+git tag api/v1.0.0 # this is your api release
+go mod edit -require github.com/my-repo@v1.0.0 # now we depend on the api module in the main module
+go mod edit -dropreplace github.com/my-repo/api # this will drop the replace directive for local development, meaning the sources from the VCS will be used instead of the ones in your monorepo checked out locally.
```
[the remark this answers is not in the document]
everettraven: I think it is possible you could have a make target that is run with something like `make release <tag>` to make it a bit easier. Those that use this plugin and that make target would still have to be aware of the sharp edges though and ensure that their CI/CD properly creates releases for both modules.

docs/book/src/plugins/multi-module-v1-alpha.md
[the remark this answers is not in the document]
camilamacedo86: Also, one question should I use or not the flag with the init as well right?

jakMoeller:
Hey guys, I have thought a lot about what was said in the issues and in the PRs and at this given point in time, I think this setup has too many disadvantage to be considered a contribution to kubebuilder:

1. It introduces complexity in module management that is referenced as not feasible/necessary in most cases by the official go documentation
2. It introduces additional effort in managing and maintaining releases, so complicated in fact that we had questions if there will be new makefile commands
3. There is always the possibility to extract your APIs into a new repository and arguably also have more control over the release process in a project spanning multiple repos relying on the same API types.
4. The plugin is written with the go cli, and introduces inconsistencies in the ordering of `go.mod` files involuntarily and we do not want to rely on the https://pkg.go.dev/golang.org/x/mod/modfile package for our adjustments since it would break our scaffolding convention
5. The plugin is significant maintenance overhead when trying to make it compatible with other plugins (e.g. deploy-image, multi-group and declarative) since it has a lot of deep changes to the module ecosystem
6. It requires at least one replace statement in the beginning either through `go.work` which is at least 2 more files plus an environment variable for build environments without GO_WORK or through `go.mod`s `replace`, which has to be manually dropped and added for every release.

All in all I think we should instead provide more dedicated guidance on how to create a repository with shared types across multiple kubebuilder projects. 
Of course it feels a bit sad to undermine my own PR here, but I think this is what would be best for the kubebuilder ecosystem

This will lead to:

1. Only having to maintaing the guide (much less effort)
2. Not having to introduce a new plugin
3. Still supporting my original planned workflow

@camilamacedo86 @jmrodri @everettraven please let me know what you think.

camilamacedo86:
HI @jakMoeller, 

Thank you for your comment and POV in https://github.com/kubernetes-sigs/kubebuilder/pull/2901#issuecomment-1245000591. That is great and we need usually try out to see if is or not the best approach. 

We have many tasks open to create a FAQ section in the docs. I think this topic as you said could be covered and fit very well in a FAQ section. Would you like to create a PR with since now you are very knowledgeable in this area? 

See, it would be like we have in SDK https://sdk.operatorframework.io/docs/faqs/
And we could add a new section (number 7) such after the plugins, see: 

Screenshot 2022-09-13 at 13 10 46

camilamacedo86:
If you prefer @jakobmoellersap, if you see that is too verbose for a FAQ section as  @everettraven highlighted that might be the case here https://github.com/kubernetes-sigs/kubebuilder/pull/2901#issuecomment-1245423714 then, you can create a topic under the Reference, see; https://book.kubebuilder.io/reference/reference.html

jakobmoellersap:
Understood. I will close this PR here and reopen a Documentation PR for discouraging mono-repos and a guide on how to make multiple repositories work. Thanks for all the valuable feedback on the way!

PR #2904, opened 2022-09-03
🐛 fix: use user/group from Dockerfile also in manager deployment's `securityContext` instead of running as root

AndiDog:
This matches the top-level security context which has `runAsNonRoot: true`, and therefore avoids errors like `container has runAsNonRoot and image will run as root`. I ran into those when starting a cluster-api (CAPI) infrastructure provider following the [implementer's guide](https://cluster-api.sigs.k8s.io/developer/providers/implementers-guide/overview.html) (which uses kubebuilder).

### DIFF
--- a/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/manager/config.go
+++ b/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/manager/config.go
@@ -93,6 +93,10 @@ spec:
           capabilities:
             drop:
               - "ALL"
+
+          # Match user/group specified in Dockerfile, do not run as root
+          runAsUser: 65532
+          runAsGroup: 65532
         livenessProbe:
           httpGet:
             path: /healthz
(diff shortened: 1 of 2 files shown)


camilamacedo86 (changes requested):
HI @AndiDog,

Thank you for your contribution. But the error `container has runAsNonRoot and image will run as root …`  is ONLY faced when/if the image does not define the UserId which is not our case. , note that the container image build for the manager has the userId and group defined: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/Dockerfile#L25

Therefore, If you are facing this issue, is because you probably customize the default scaffold and your Dockerfile is not defining the userId. 

TL'DR: **From Dockefile best practices** 

Screenshot 2022-07-12 at 00 10 10
From: https://docs.docker.com/engine/reference/builder/#user

/hold

camilamacedo86:
I am closing this one as defered.
However, if you need help please feel free to raise an issue and we will do our best to help you out.

AndiDog:
Indeed kubebuilder generates a Dockerfile with `USER 65532:65532` when doing `kubebuilder init --domain [...]`. And Kubernetes uses that correctly. I got distracted by some k8s docs which said otherwise, but that turned out unrelated. The problem seems to rather be in the cluster-api Tiltfile workflow which [does not have the `USER` instruction](https://github.com/kubernetes-sigs/cluster-api/search?q=dockerfile_contents) yet. The Dockerfile from kubebuilder is _fine_. I will take this issue to the cluster-api project instead 👍

PR #2911, opened 2022-09-06
:sparkles: add support for go1.19, update envtest version to use k8s 1.25 

oscr:
Description
* Update to go1.19
* Update envtest version for k8s 1.25
* Update goreleaser: [v1.11.2](https://github.com/goreleaser/goreleaser/releases/tag/v1.11.2)

Based on #2714


testdata/project-v3-config/go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kubebuilder/testdata/project-v3-config
 
-go 1.18
+go 1.19
 
 require (
```
oscr, line 5: @camilamacedo86 A side question. Kubebuilder seems to have updated to ginkov2, but the testdata is still `github.com/onsi/ginkgo v1.16.5`. Should this be updated? And if so where?

atoato88:
Is it needed to use modules for k8s v1.25 in `go.mod` files on this PR?
e.g:
Updated [`go.mod`](https://github.com/kubernetes-sigs/kubebuilder/pull/2911/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6) file in this PR remains to have `k8s.io/apimachinery` as `v0.24.2`.
Sould we use as `v0.25.0` like following:
```
k8s.io/apimachinery v0.25.0
```
If so, it needs to update multiple `go.mod` files in same manner.

Or, it aims to just update k8s version for envtest only?

pkg/plugins/golang/v3/init.go
```diff
@@ -35,7 +35,7 @@ import (
 
 // Variables and function to check Go version requirements.
 var (
-	goVerMin = golang.MustParse("go1.17.9")
+	goVerMin = golang.MustParse("go1.18")
```
camilamacedo86, line 38: @oscr I think the problem here is that we need to put the value with 0. So, it needs to be 1.18.0

PR #2471, opened 2022-01-06
:warning: (go/v3) upgrade controller-runtime from 0.10.0 to 0.11.0, k8s from 1.22 to 1.23 and controller-gen from v0.0.7 to v0.8.0

ankitm123:
**Description**
- upgrade controller-runtime from 0.10.0 to 0.11.0
- k8s from 1.22 to 1.23 
- controller-gen from v0.0.7 to v0.8.0 

also, envtest setup from 1.22 to 1.23

**NOTE**: This pr changes the scaffolds produced by KB CLI for go/v3 plugin. The go/v2 plugin cannot be upgraded and has been kept as a legacy. 

Signed-off-by: ankitm123 <ankitmohapatra123@gmail.com>

fixes #2448

PR #2920, opened 2022-09-09
✨ (go/v3, go/v4-alpha and declarative/v1): Update k8s deps by upgrading controller-runtime from v0.12.2 to 0.13.0 and kubebuilder-declarative-pattern from fe5be9431eae158f86f9de23000a9a2ec06745fc to e0605f0e1a40f97293cb3773f57de695c8bc76af

oscr:
Updates: 
* controller-runtime
* kubebuilder-declarative-pattern

There was no update for controller-tools, however I am watching it for releases.

I hope I correctly updated declarative-pattern

cc @camilamacedo86

### DIFF
--- a/pkg/plugins/golang/v3/commons.go
+++ b/pkg/plugins/golang/v3/commons.go
@@ -83,7 +83,13 @@ manifests: controller-gen`
 			log.Warnf("unable to update the Makefile with %s: %s", "ENVTEST_K8S_VERSION = 1.21", err)
 		}
 
-		// latest version of controller-runtime where v1beta1 is supported
+		// DO NOT UPDATE THIS VERSION
+		// Note that this implementation will update the go.mod to downgrade the versions for those that are
+		// compatible v1beta1 CRD/Webhooks k8s core APIs if/when a user tries to create an API with
+		// create api [options] crd-version=v1beta1. The flag/feature is deprecated. however, to ensure that backwards
+		// compatible we must introduce this logic. Also, note that when we bump the k8s dependencies we need to
+		// ensure that the following replacements will be done accordingly to downgrade the versions.
+		// The next version of the Golang base plugin (go/v4-alpha) no longer provide this feature.
 		const controllerRuntimeVersionForVBeta1 = "v0.9.2"
 
 		if err := util.ReplaceInFile("go.mod",
--- a/pkg/plugins/golang/v3/scaffolds/init.go
+++ b/pkg/plugins/golang/v3/scaffolds/init.go
@@ -34,7 +34,7 @@ import (
 
 const (
 	// ControllerRuntimeVersion is the kubernetes-sigs/controller-runtime version to be used in the project
-	ControllerRuntimeVersion = "v0.12.2"
+	ControllerRuntimeVersion = "v0.13.0"
 	// ControllerToolsVersion is the kubernetes-sigs/controller-tools version to be used in the project
 	ControllerToolsVersion = "v0.9.2"
 
--- a/pkg/plugins/golang/declarative/v1/api.go
+++ b/pkg/plugins/golang/declarative/v1/api.go
@@ -32,7 +32,7 @@ import (
 const (
 	// kbDeclarativePattern is the sigs.k8s.io/kubebuilder-declarative-pattern version
 	kbDeclarativePatternForV2 = "v0.0.0-20200522144838-848d48e5b073"
-	kbDeclarativePatternForV3 = "fe5be9431eae158f86f9de23000a9a2ec06745fc"
+	kbDeclarativePatternForV3 = "e0605f0e1a40f97293cb3773f57de695c8bc76af"
 )
 
 var _ plugin.CreateAPISubcommand = &createAPISubcommand{}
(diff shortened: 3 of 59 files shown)


oscr:
Question: it seems like updating the dependencies fail with go1.17. I tried it locally and it seems to be the same result. Can we bump it?

everettraven:
@oscr It looks like we may need to update the Legacy unit tests action to use Go 1.19.0 instead of 1.17.3

laxmikantbpandhare:
> @oscr It looks like we may need to update the Legacy unit tests action to use Go 1.19.0 instead of 1.17.3

@everettraven Looks like @oscr created another PR for the same #2911

oscr:
@laxmikantbpandhare @everettraven Sorry I asked to break down #2911 but I was slow to close it.

oscr:
> @oscr It looks like we may need to update the Legacy unit tests action to use Go 1.19.0 instead of 1.17.3

No problem! I will update it immediately

go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kubebuilder/v3
 
-go 1.17
+go 1.18
```
camilamacedo86, line 3: @oscr if we bump the Kubebuilder version from 0.17 that would go to 1.19 not 1.18
Kubebuilder is also used as LIB so that can impact those that consume the project.
Therefore, the bump go version would make sense to be done in a PR where we do only this change.
Note that we use the title and emojis to generate the changelogs.

oscr: @camilamacedo86 Since we're starting to use go 1.19 should I upgrade everywhere?

oscr: @camilamacedo86 Thank you for the advice. I will have it done in another pr.

camilamacedo86: a) Do we need to change the go version to bump the declarative one? If so, can we bump the go version first in another PR?

b) This pr is not updating the controller-runtime see that it is done here: 
https://github.com/kubernetes-sigs/kubebuilder/blob/c7d38ce80602d4f2915a17f6f681a27901698de6/pkg/plugins/golang/v3/scaffolds/init.go#L35-L42. ()

**We can bump only the declarative one and do the other bumps in another PR that is totally fine.**
If you want to do that, just revert the golang changes and I am happy to lgtm/approve this one to get merged 🚀  

Also, you can check an example of a PR where we bump all deps, see: https://github.com/kubernetes-sigs/kubebuilder/pull/2471

Again, thank you for all your collab 🥇

pkg/plugins/golang/v3/commons.go
```diff
@@ -84,7 +84,7 @@ manifests: controller-gen`
 		}
 
 		// latest version of controller-runtime where v1beta1 is supported
-		const controllerRuntimeVersionForVBeta1 = "v0.9.2"
+		const controllerRuntimeVersionForVBeta1 = "v0.13.0"
```
camilamacedo86, line 87: @oscr this one we cannot update.
The kubebuilder CLI has an option to create the projects using crd/webhooks API v1beta1 which is no longer provided since k8s 1.22. This logic will update the go.mod file to use the deps that works with v1beta1 if/when someone calls the command to create api [options] crd-version=v1beta1

Therefore, could you please revert it and also add a comment on the top to explain it so that others in the future does not try to update this one?

Example (something like):

````
// DO NOT UPDATE THIS VERSION
// Note that this implementation will update the go.mod to downgrade the versions for those that are
// compatible v1beta1 CRD/Webhooks k8s core APIs if/when an user tries to create a API with
// create api [options] crd-version=v1beta1. The flag/feature is deprecated. however, to ensure that backwards 
// compatible we must introduce this logic. Also, note that when we bump the k8s dependencies we need to
// ensure that the following replacements will be done accordingly to downgrade the versions. 
// The next version of the Golang base plugin (go/v4-alpha) no longer provide this feature.

Comment (camilamacedo86) [reply pkg/plugins/golang/v3/commons.go:87]: /hold

File: pkg/plugins/golang/v3/scaffolds/init.go [hunk 3]
```diff
@@ -34,7 +34,7 @@ import (
 
 const (
 	// ControllerRuntimeVersion is the kubernetes-sigs/controller-runtime version to be used in the project
-	ControllerRuntimeVersion = "v0.12.2"
+	ControllerRuntimeVersion = "v0.13.0"
 	// ControllerToolsVersion is the kubernetes-sigs/controller-tools version to be used in the project
 	ControllerToolsVersion = "v0.9.2"
```
camilamacedo86, line 39: @oscr could we also update the controller tools here to the latest version which uses k8s 1.25?
We need to ask for a new release, see: https://github.com/kubernetes-sigs/controller-tools
I am adding it to the channel now.

pkg/plugins/golang/v3/commons.go
camilamacedo86: Btw, we have been removing the testdata and samples with this option because it is deprecated. 
So, that it did not fail in the tests but we need to ensure that we have a testdata that checks it.
The ci should fail with this change. I am adding an issue so we are able to ensure that.

camilamacedo86 (approved):
/hold cancel

/approved
/lgtm

@oscr thank you for the contribution 🥇 
PS.: I hope that you do not mind but I just update the title for we are able to generate the release notes automatically with the info (see that I added the name of the plugins changed/affected by)

oscr:
@camilamacedo86 No, I don't mind at all. Thank you for updating the title to better reflect the content.

Issue #2532, opened 2022-03-09
Upgrade ginkgo to v2 for go/v4 plugin


haoxins:
### What do you want to happen?

I just submitted a PR for upgrading ginkgo to `v2`, see https://github.com/kubernetes-sigs/kubebuilder/pull/2522
After some discussions we found it's a broken change for `go/v3`.
So I created this issue for tracking the upgrade and intend to be added to [v4 milestone](https://github.com/kubernetes-sigs/kubebuilder/milestone/26).

camilamacedo86:
We checked in the review that changes in the scaffolds are required and that the previous code will not work with ginkgo v2 so it is a breaking change and then we cannot move forward in go/v3. We will need a go/v4 plugin to add this one.

See that the scaffolds require to be changed with the following migrations. 

https://onsi.github.io/ginkgo/MIGRATING_TO_V2#generating-custom-reports-when-a-test-suite-completes
https://onsi.github.io/ginkgo/MIGRATING_TO_V2#removed-async-testing
Also, you can check the policy: https://github.com/kubernetes-sigs/kubebuilder/blob/master/VERSIONING.md#introducing-changes-to-plugins

camilamacedo86:
We update the CLI: https://github.com/kubernetes-sigs/kubebuilder/pull/2876
However, we need to update the scaffolds.
Also, since C+R and k8s made the update and their a dep we probably need to ensure the same for all scaffolds indeed go/v3

camilamacedo86:
Hi @jakobmoellersap,

Are you able to help us update the scaffolds by doing the same changes in the scaffolds so we are able to do a new release with this change?

jakMoeller:
Yup, was out the last few days. Will submit a draft today.

camilamacedo86:
Done.

PR #2939, opened 2022-09-13
:warning: Migrate scaffolding of v3 / v4-alpha to ginkgo v2

jakobmoellersap:
This PR is updating the v3 and v4-alpha plugins to use ginkgo v2. It purposefully does not introduce ginkgo cli as most options can be added onto `go test` instead. It is a follow up for https://github.com/kubernetes-sigs/kubebuilder/issues/2532

Currently it does not use the NewLineReporter from the old scaffolding, but it can be manually added back (it is not recommended to do so by Ginkgo but possible with a custom hook into the rpeorting lifecycle). If this is wanted we can add docs on how to do that.

Currently the scaffolding does not automatically do this and people can manually add it.

### DIFF
--- a/docs/book/src/migration/manually_migration_guide_v2_v3.md
+++ b/docs/book/src/migration/manually_migration_guide_v2_v3.md
@@ -356,8 +356,8 @@ module example
 go 1.18
 
 require (
-    github.com/onsi/ginkgo v1.16.5
-    github.com/onsi/gomega v1.18.1
+    github.com/onsi/ginkgo/v2 v2.1.4
+    github.com/onsi/gomega v1.19.0
     k8s.io/api v0.24.0
     k8s.io/apimachinery v0.24.0
     k8s.io/client-go v0.24.0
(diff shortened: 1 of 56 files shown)


jakobmoellersap:
@camilamacedo86 

In the upstream `controller-runtime` PR, there is also a junit-aggregation logic in https://github.com/kubernetes-sigs/controller-runtime/pull/1977/files#diff-ab3c2c731b5083def78d27378838304f494f0ae5dec7bcf2339e1da30a9bf6d0R28 . we could achieve the same with a script that aggregates all `junit-report.xml` files generated with the `-ginkgo.junit-report=junit-report.xml` flag into a folder called `artifacts`. I was not sure if we want to include a script for this into the scaffolding for this purpose:

In `Makefile` we would need to modify the `test` target:

```Makefile
.PHONY: test
	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./... -coverprofile cover.out -ginkgo.junit-report=junit-report.xml
	ARTIFACTS="./artifacts" hack/aggregate-junit-reports.sh
```

Then in `hack/aggregate-junit-reports.sh` we would create a small script for the aggregation:
```sh
#!/usr/bin/env bash

set -e

if [[ -n ${ARTIFACTS:-} ]]; then
  mkdir -p "${ARTIFACTS}"
fi

find . -name '*junit-report.xml' -print0 | while IFS= read -r -d '' file
do
  ((count++))
  new_file=${file#./}
  new_file=${new_file%/junit-report.xml}
  new_file=${new_file//"/"/"-"}
  mv "$file" "$ARTIFACTS/junit_${new_file}.xml"
  echo "aggregated $file to $ARTIFACTS/junit_${new_file}.xml"
done
```

camilamacedo86 (approved):
The changes show fine. 
Great work 🥇 

If that has an issue it should also fail in the test assuming that it will pass
/approved

docs/book/src/migration/manually_migration_guide_v2_v3.md
```diff
@@ -522,6 +522,78 @@ func (r *<MyKind>Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
     log := r.Log.WithValues("cronjob", req.NamespacedName)
 ```
 
+#### Update your controller adn webhook test suite
```
everettraven, line 525: Nit:
```suggestion
#### Update your controller and webhook test suite
```

everettraven (changes requested):
Just a minor typo in the docs. Other than that looks great! Once the typo is fixed I'll approve and LGTM.

everettraven (approved):
/lgtm

Issue #2782, opened 2022-06-30
Meta - ComponentConfig scaffolds should not be done by default 


camilamacedo86:
### What broke? What's expected?

Users should have the scaffold of https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/component-config-tutorial/testdata/project/config/manager/controller_manager_config.yaml ONLY/when they create the project following the tutorial: https://book.kubebuilder.io/component-config-tutorial/api-changes.html

That means, the scaffold should only be done when we run `kubebuilder init --domain tutorial.kubebuilder.io --component-config`. However, it has been added to the default scaffolds when we run `kubebuilder init` and it has brought confusion since users do not understand how properly use it and that their code in the main.go also needs to be customized accordingly. 

### Reproducing this issue

**Examples:**
- https://github.com/kubernetes-sigs/kubebuilder/pull/2778
- https://github.com/operator-framework/operator-sdk/issues/5884

### KubeBuilder (CLI) Version

3.5.0

### Plugin versions

Therefore, these options would fit in a plugin and should not be a flag. It is only a flag because it was implemented in the past (Kubebuilder 2.X) before the plugin design was introduced. 

- [ ] **Proposed changes (Bug FIX)**
a) Ensure that the component config scaffolds are not done by default and are only done when the flag is used
b) Review the documentation and ensure that it proper clarifies how it should be used and work

- [ ] **Proposed changes in the long term (RFE)**
a) Deprecated the flag
b) Move ComponentConfig behaviour to a plugin that can be used in the init/edit subcommands.

camilamacedo86:
Re-open this one it still missing we create a plugin for this feature and deprecate the flag. 
Also, we should update the docs in the pr where it will be addressed.

laxmikantbpandhare:
@camilamacedo86 yes, working on plug-in part for this command.

laxmikantbpandhare:
@camilamacedo86 - Raised PR #2942 for one of the fixes.

Working on the below part and will raise the PR soon.

`Move ComponentConfig behavior to a plugin that can be used in the init/edit subcommands.`

laxmikantbpandhare:
@camilamacedo86 raised PR for `init/edit` subcommands: https://github.com/kubernetes-sigs/kubebuilder/pull/3022

rustycl0ck:
- Just mentioning here for information: https://github.com/kubernetes-sigs/controller-runtime/issues/895
[...]

PR #2942, opened 2022-09-14
✨ deprecate flag component-config

laxmikantbpandhare:
this will fix one of the requirements from #2782 

This fix will mark the flag `component-config` as deprecated and throw the warning to the user whenever user tries to use it.

### DIFF
--- a/pkg/plugins/common/kustomize/v1/create.go
+++ b/pkg/plugins/common/kustomize/v1/create.go
@@ -35,7 +35,10 @@ type createSubcommand struct {
 	force bool
 }
 
-func (p *createSubcommand) BindFlags(fs *pflag.FlagSet) { p.flagSet = fs }
+func (p *createSubcommand) BindFlags(fs *pflag.FlagSet) {
+	p.flagSet = fs
+	_ = fs.MarkDeprecated("component-config", "component-config flag will be deprecated in upcoming releases")
+}
 
 func (p *createSubcommand) InjectConfig(c config.Config) error {
 	p.config = c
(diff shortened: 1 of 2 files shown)


camilamacedo86:
HI @laxmikantbpandhare,

Before deprecating IMHO we need to have the solution which we will deprecate it for. 
So, first, we need to create the plugin then we can go there and say we are deprecating this flag in favour of using the plugin. We also will need to update the tutorial to let users know that.

camilamacedo86 (commented):
/hold 
Until we have the other solution in place

laxmikantbpandhare:
> HI @laxmikantbpandhare,
> 
> Before deprecating IMHO we need to have the solution which we will deprecate it for. So, first, we need to create the plugin then we can go there and say we are deprecating this flag in favour of using the plugin. We also will need to update the tutorial to let users know that.

@camilamacedo86 - Yes, I am working on it and let it be on `hold` until we have a plugin for it.

laxmikantbpandhare:
> > HI @laxmikantbpandhare,
> > Before deprecating IMHO we need to have the solution which we will deprecate it for. So, first, we need to create the plugin then we can go there and say we are deprecating this flag in favor of using the plugin. We also will need to update the tutorial to let users know that.
> 
> @camilamacedo86 - Yes, I am working on it and let it be on `hold` until we have a plugin for it.

@camilamacedo86 - PR is ready for review - https://github.com/kubernetes-sigs/kubebuilder/pull/2944

laxmikantbpandhare:
these got closed automatically as head of the my repo got deleted. I will raise PR once again.

laxmikantbpandhare:
these got closed automatically as head of the my repo got deleted. I will raise PR once again.

PR #2949, opened 2022-09-17
🌱 grammar: replace all occurrences of "the the" with "the"

oscr:
A minor cleanup to remove the occurrences of "the the" and replace it with "the".

EDIT: I need to use more words to describe this change to avoid a check from failing. But I think it's pretty clear description :)


camilamacedo86 (approved):
/lgtm
/approved

PR #2954, opened 2022-09-20
:seedling: update kubebuilder core dependencies for k8s 1.25

everettraven:
## Description of the change
- Updates the `go.mod` and `go.sum` to bump the k8s dependencies to support 1.25

## Motivation for the change
The scaffolds were updated to have dependencies bumped to k8s 1.25 but the core Kubebuilder go.mod dependencies were not. This PR fixes that.

## Open Questions
- [controller-tools](https://github.com/kubernetes-sigs/controller-tools) does not currently have a release that has the k8s1.25 bumps but in commit https://github.com/kubernetes-sigs/controller-tools/commit/afb4be356dde836b9a0c5a9540b328979aa46496 it was bumped. Should we wait for a release or base off a commit?


varshaprasad96 (commented):
/lgtm
Just a TODO: we need to bump controller-tools too, if it gets released soon. 
We could wait for a day, if it gets delayed please feel free to merge it.

everettraven:
@varshaprasad96 would it be worth it to consider bumping controller-tools to the latest master commit in the meantime?

camilamacedo86:
HI @everettraven,

Controller-tools is released and we can do the bump in a follow-up. So, I think it is OK to go.

/lgtm

PR #2957, opened 2022-09-21
🌱 add dependency bot updates

camilamacedo86:
## Description

add dependency bot updates

## Motivation

Help us check what is outdated

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,49 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    # Workflow files stored in the
+    # default location of `.github/workflows`
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: ":seedling:"
+    labels:
+      - "ok-to-test"
+
+  # Maintain dependencies for go
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "ok-to-test"
+
+  # Maintain dependencies for dockerfile in the branches
+  - package-ecosystem: docker
+    directory: "/build/thirdparty/darwin"
+    target-branch: "tools-releases"
+    schedule:
+      interval: daily
+  - package-ecosystem: docker
+    directory: "/build/thirdparty/linux"
+    target-branch: "tools-releases"
+    schedule:
+      interval: "weekly"
+
+  # Maintain dependencies for dockerfile scaffold in the projects
+  - package-ecosystem: docker
+    directory: "testdata/project-v3"
+    schedule:
+      interval: daily
+  - package-ecosystem: docker
+    directory: "testdata/project-v4"
+    schedule:
+      interval: "weekly"


.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,43 @@
...
+      interval: daily
+  - package-ecosystem: docker
+    directory: "/build/thirdparty/linux"
+    target-branch: "tools-releases"
+    schedule:
+      interval: daily
+
+  # Maintain dependencies for dockerfile scaffold in the projects
+  - package-ecosystem: docker
+    directory: "/project-v3"
```
oscr, line 37: I think this should be `/testdata/project-v3`?

"You define the directory relative to the root of the repository for all ecosystems except GitHub Actions."
 
Reference: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directory

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,43 @@
...
+    schedule:
+      interval: daily
+  - package-ecosystem: docker
+    directory: "/project-v4"
```
oscr, line 41: Same as above

oscr (commented):
Great initiative! I think this will be really helpful in the future and save lots of work.

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,43 @@
+# To get started with Dependabot version updates, you'll need to specify which
```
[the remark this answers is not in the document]
everettraven: @oscr is correct that it should be under `.github/dependabot.yaml`

Official Dependabot docs from GitHub: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#enabling-dependabot-version-updates

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,43 @@
...
+version: 2
+updates:
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    # Workflow files stored in the
+    # default location of `.github/workflows`
+    directory: "/"
+    schedule:
+      interval: "daily"
```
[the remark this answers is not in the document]
everettraven: I agree with @oscr I think `weekly` would be better

.github/workflows/dependabot.yml
everettraven: +1

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,43 @@
...
+
+version: 2
+updates:
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    # Workflow files stored in the
+    # default location of `.github/workflows`
+    directory: "/"
+    schedule:
```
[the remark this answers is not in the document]
everettraven: +1

.github/workflows/dependabot.yml
everettraven: +1

everettraven (changes requested):
I agree with the changes @oscr mentioned in their review comments

Kavinjsir (commented):
See that we've introduced a couple of labelling before merging a PR, what about introducing [label](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#labels) ?

One use case maybe `ok-to-test`:
```yaml
- package-ecosystem: "github-actions"
  ...
  labels:
  - "ok-to-test"
```
(In case if it feels necessary to run e2e tests before merging..)

camilamacedo86:
All changes are done succefully.

.github/dependabot.yml
```diff
@@ -0,0 +1,49 @@
...
+    labels:
+      - "ok-to-test"
+
+  # Maintain dependencies for go
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "ok-to-test"
```
oscr, line 27: Will this work for go dependencies used by the testdata? When I bumped ginkgo, `make test` failed until I ran `make generate`?

camilamacedo86: The label only allows the tests be executed. Just that.
Also, note that this action is just to open PR's with updates in the github actions. So, it has no relation with make generate at all. The make generate re-generate  the samples projects scaffold under testdata to ensure that the scaffolds matches with the latest changes (it only fails/or is required when you do prs to change the scaffolds)

everettraven (approved):
LGTM (avoiding `lgtm` command to allow other reviewers to approve prior to merging)

.github/dependabot.yml
oscr: Ok. My thought was if dependabot for example bumps ginkgo (~`go get -u github.com/onsi/ginkgo/v2`) and makes a pr, then GH actions will automatically run `make test` and it would fail. But this isn't a big worry for me and I could be wrong here.

PR #2981, opened 2022-09-29
:seedling: remove WORKSPACE file commited by mistake

camilamacedo86:
## Description

remove WORKSPACE file commited by mistake

## Motivation

We should not commit these configs since their specific for some kind of IDE.


camilamacedo86:
Moving forward with this one since it is too small to ask for others spending effort in review.

Issue #2940, opened 2022-09-13
Add support for windows 


camilamacedo86:
## Description

We would like to add support for Windows. However, to achieve this goal would be required to ensure that tests/ci run against this platform, change the release to build the binary for windows and indeed ensure that controller-tools used by Kubebuilder could also run in this platform. On top of that, we would have other challenges such as the Makefile works for windows as the script to generate the testdata samples as well run in this env.   

## History

This issue aggregated the whole tracked/identified issues so far. See that each issue linked here has comments and discussions about the topic and can be used as history. 

- Release #300
- Tests #1409
- Scaffolded Makefile #1408
- Kubebuilder Makefile #1407
- Document make dependency #1406
- Windows compatibility #1405
- Testdata script (called golden in the past) Golden script for Windows #1126

## Goal

Allow Kubebuilder be used as CLI and provide support for windows SO.

manojmohandas14:
Hi 
I am able to clone the repo in a Windows machine and did 

go build -o bin/kubebuilder ./cmd

I then added the generated kubebuilder to Path variable. kubebuilder works. 

I am able to initialize & create api.

IPreferWater:
To use kubebuilder on windows without WSL, I used a docker container with volume bounded to my local drive

I did a repo with the example => [kubebuilder-docker](https://github.com/IPreferWater/kubebuilder-docker)

camilamacedo86:
HI @Piwriw and @IPreferWater 

Currently we do not support windows.
See the description to check all that must be done for we are able to do so.
If you would like to contribute and collaborate with this initiative please feel free to do so. 
Your help is welcome !!!

We just do not have available effort to work on this.

vitorfloriano:
@camilamacedo86 

I have thought about tackling this issue for some time (since I myself am on Windows as well), but reached to the conclusion that adding native support for Windows may not be worth the development effort and maintenance overhead, given that:
- Kubernetes environments are mostly Linux-based;
- The project has limited maintainer bandwidth (none of the current maintainers are on Windows, so the feature would rely on Community efforts, which have been close to none in all these years);
- Full Windows support would require significant ongoing maintenance (PowerShell scripts, Windows CI, platform-specific testing).

Not to shun away Windows devs (heck, I'm one of them!) but actually offer an alternative path.

[...]

PR #2984, opened 2022-09-30
feat: Add windows support to tools-releases

codablock:
This PR adds support for Windows builds in tools-releases, so that these can be used via setup-envtest. If this is merged, it will require enabling it in the GCP Cloud Build (I have no experience with that part).

I'm currently using the branch behind this PR in https://github.com/kluctl/kubebuilder-tools-releases-windows, which simply clones my kubebuilder fork, checks out the tools-releases-windows branch and manually runs the Docker build for the Windows release. I do this to trick setup-envtest into accepting the downloaded binaries when invoking it with `-i`, which will then later allow me to simply drop the manual downloading. You can find all this here: https://github.com/kluctl/kluctl/blob/main/Makefile#L60.

When this PR gets merged and GCP Cloud Build is updated as well, I will then be able to remove the manual downloading and simply rely on setup-envtest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ GCP Cloud Build watches this branch.  On every push, it runs the pipeline define
 - `_GOARCH=arm64 _GOOS=linux`
 - `_GOARCH=arm64 _GOOS=darwin`
 - `_GOARCH=ppc64le _GOOS=linux`
+- `_GOARCH=amd64 _GOOS=windows`
 
 (we may add more the in the future).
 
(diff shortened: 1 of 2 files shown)


camilamacedo86:
/ok-to-test

I am ok with us adding this one. But is important to highlight that Kubebuilder itself does not support windows besides that be desired, see: https://github.com/kubernetes-sigs/kubebuilder/issues/2940

camilamacedo86:
@jmrodri and @varshaprasad96 wdyt?

camilamacedo86 (approved):

jmrodri (approved):
/lgtm

varshaprasad96:
Agreed with @camilamacedo86, it would be nice to explicitly mention in docs that Kubebuilder has not been tested against Windows and its upto the developer to do so. 
We could add this to docs in a follow up too. Else it looks good! 

Thanks for the contribution @codablock!
/lgtm

varshaprasad96 (approved):

camilamacedo86:
Hi @codablock, 

Yes, we will need to enable it into the Google Cloud, see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/RELEASE.md#to-build-the-kubebuilder-tools-artifacts-required-to-use-env-test

I change my email and we need help here from others to re-enable our access there. So, asap we will add it. You can follow up: https://github.com/kubernetes-sigs/kubebuilder/issues/3005

PR #2985, opened 2022-10-01
⚠️  go/v4-alpha: change the layout to follow Golang Standards

camilamacedo86:
## Description 

- The directory `apis` was renamed to `api` to follow the standard
- The `controller(s)` directory has been moved under a new directory called `internal` and renamed to singular as well `controller`
- The `main.go` previously scaffolded in the root directory has been moved under a new directory  called `cmd`

Therefore, you can check the changes in the layout results into:

```sh
...
├── cmd
│ └── main.go
├── internal
│ └── controller
└── api
```

## Motivation

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/932


camilamacedo86:
/hold

To allow everybody to give their opinion.

pkg/plugins/golang/declarative/v1/init.go
```diff
@@ -41,36 +41,46 @@ func (p *initSubcommand) InjectConfig(c config.Config) error {
...
-	}
-	return nil
+	return updateDockerfile()
 }
 
 // updateDockerfile will add channels staging required for declarative plugin
 func updateDockerfile() error {
 	fmt.Println("updating Dockerfile to add channels/ directory in the image")
-	managerFile := filepath.Join("Dockerfile")
+	dokerfile := filepath.Join("Dockerfile")
```
Kavinjsir, line 50: nits: maybe `dockerfile`?

pkg/plugins/golang/v4/init.go
```diff
@@ -0,0 +1,209 @@
...
+
+	subcmdMeta.Description = `Initialize a new project including the following files:
+  - a "go.mod" with project dependencies
+  - a "PROJECT" file that stores project configuration
+  - a "Makefile" with several useful make targets for the project
+  - several YAML files for project deployment under the "config" directory
+  - a "main.go" file that creates the manager that will run the project controllers
+`
+	subcmdMeta.Examples = fmt.Sprintf(`  # Initialize a new project with your domain and name in copyright
+  %[1]s init --plugins go/v3 --domain example.org --owner "Your name"
```
Kavinjsir, line 72: nits: maybe `--plugins go/v4`?

pkg/plugins/golang/v4/scaffolds/init.go
```diff
@@ -0,0 +1,136 @@
...
+		machinery.WithConfig(s.config),
+		machinery.WithBoilerplate(string(boilerplate)),
+	)
+
+	// If the KustomizeV2 was used to do the scaffold then
+	// we need to ensure that we use its supported Kustomize Version
+	// in order to support it
+	kustomizeVersion = kustomizecommonv1.KustomizeVersion
+	kustomizev2 := kustomizecommonv2alpha.Plugin{}
+	gov4alpha := "go.kubebuilder.io/v4-alpha"
```
Kavinjsir, line 109: See that the directory for the plugin is `plugins/golang/v4`, are we going to keep the plugin key as `.../v4-alpha` OR `v4` directly?

Kavinjsir (commented):
Looks exciting to have this big move 👍🏼 
Just some small nits and concerns.

pkg/plugins/golang/v4/scaffolds/init.go
camilamacedo86: For now, I think we should keep the key/version (go/v4-alpha).
At some point, we will stabilize the plugin which will change the key for go/v4.
Until there, we still able to introduce breaking changes.

camilamacedo86:
/hold 

We should not merge until we are able to discuss.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
```diff
@@ -104,25 +113,42 @@ func (s *apiScaffolder) Scaffold() error {
...
+			IsLegacyLayout:           false,
+		}
+		if err := scaffold.Execute(
+			controller,
+		); err != nil {
+			return fmt.Errorf("error scaffolding controller: %v", err)
+		}
+		if err := s.updateControllerCode(*controller); err != nil {
+			return fmt.Errorf("error updating controller: %v", err)
+		}
```
everettraven, line 143: Maybe we can prevent some code duplication with something like:
```go
controller := &controllers.Controller{
    ControllerRuntimeVersion: golangv3scaffolds.ControllerRuntimeVersion,
    IsLegacyLayout:           true,
}

if !isLegacyLayout {
    controller.ControllerRuntimeVersion = golangv4scaffolds.ControllerRuntimeVersion
    controller.IsLegacyLayout = false
}

if err := scaffold.Execute(controller); err != nil {
      return fmt.Errorf("error scaffolding controller: %v", err)
}

if err := s.updateControllerCode(*controller); err != nil {
	return fmt.Errorf("error updating controller: %v", err)
}
```

pkg/plugins/golang/declarative/v1/scaffolds/api.go
```diff
@@ -70,8 +71,15 @@ func (s *apiScaffolder) Scaffold() error {
 		machinery.WithResource(&s.resource),
 	)
 
+	isLegacyLayout := false
+	for _, pluginKey := range s.config.GetPluginChain() {
+		if strings.Contains(pluginKey, "go.kubebuilder.io/v3") {
+			isLegacyLayout = true
+		}
+	}
```
erikgb, line 79: This code is duplicated. Would it be possible to avoid it by adding a `LegacyLayout()` func on the `Config` struct?

pkg/plugins/golang/declarative/v1/init.go
```diff
@@ -51,21 +51,28 @@ func (p *initSubcommand) Scaffold(_ machinery.Filesystem) error {
...
+	isLegacyLayout, err := util.HasFragment(dokerfile, "COPY controllers/ controllers/")
 	if err != nil {
 		return err
 	}
 
-	err = insertCodeIfDoesNotExist(managerFile,
-		"COPY --from=builder /workspace/manager .",
-		"\n# copy channels\nCOPY --from=builder /channels /channels\n")
-	if err != nil {
-		return err
```
erikgb, line 68: Was the delete of this code intentional?

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
```diff
@@ -266,32 +292,51 @@ func (s *apiScaffolder) updateControllerCode(controller controllers.Controller)
...
+		kustomizeV2Scaffolder := kustomizev2scaffolds.NewAPIScaffolder(
+			s.config,
+			s.resource,
+			true,
+		)
+		kustomizeV2Scaffolder.InjectFS(s.fs)
+
+		if err := kustomizeV2Scaffolder.Scaffold(); err != nil {
+			return fmt.Errorf("error scaffolding kustomize files for the APIs: %v", err)
+		}
```
everettraven, line 320: minor nit for reducing code duplication since both `NewAPIScaffolder()` functions return the same type:
```go
var kustomizeScaffolder plugins.Scaffolder

if isLegacyLayout {
  kustomizeScaffolder = kustomizev1scaffolds.NewAPIScaffolder(
	s.config,
	s.resource,
	true,
  )
} else {
  kustomizeScaffolder = kustomizev2scaffolds.NewAPIScaffolder(
	s.config,
	s.resource,
	true,
  )
}

kustomizeScaffolder.InjectFS(s.fs)

if err := kustomizeScaffolder.Scaffold(); err != nil {
	return fmt.Errorf("error scaffolding kustomize files for the APIs: %v", err)
}
```

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/config/samples/crd_sample.go
```diff
@@ -36,6 +37,7 @@ func (f *CRDSample) SetTemplateDefaults() error {
 		f.Path = filepath.Join("config", "samples", "%[group]_%[version]_%[kind].yaml")
 	}
 	f.Path = f.Resource.Replacer().Replace(f.Path)
+	fmt.Println(f.Path)
```
everettraven, line 40: Debugging?

docs/book/src/plugins/go-v4-plugin.md
```diff
@@ -24,6 +24,8 @@ under the [testdata][testdata] directory on the root directory of the Kubebuilde
 - If you are looking to have your project update with the latest version available
 - if you are not targeting k8s versions < `1.16` and `1.20` if you are using webhooks
 - If you are looking to work on with scaffolds which are compatible with k8s `1.25+`
+- If you are looking for the new layout where to maintain good Go hygiene the "api(s)" and "controller(s)" are scaffold
```
erikgb, line 27: nit: Maybe refer to "Standard Go Project Layout" instead of "good Go hygiene"? I am pretty sure the current layout is not "bad Go hygiene". 😜

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -35,18 +35,28 @@ type Controller struct {
...
-			f.Path = filepath.Join("controllers", "%[kind]_controller.go")
+			if f.IsLegacyLayout {
+				f.Path = filepath.Join("controllers", "%[kind]_controller.go")
+			} else {
+				f.Path = filepath.Join("pkg", "controllers", "%[kind]_controller.go")
+			}
 		}
 	}
 	f.Path = f.Resource.Replacer().Replace(f.Path)
+	fmt.Println(f.Path)
```
everettraven, line 59: Debugging?

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
```diff
@@ -104,25 +113,42 @@ func (s *apiScaffolder) Scaffold() error {
...
-	if err := scaffold.Execute(
-		controller,
-	); err != nil {
-		return fmt.Errorf("error scaffolding controller: %v", err)
-	}
+	if isLegacyLayout {
+		controller := &controllers.Controller{
+			ControllerRuntimeVersion: golangv3scaffolds.ControllerRuntimeVersion,
+			IsLegacyLayout:           true,
+		}
```
erikgb, line 120: Maybe initialize the `Controller` struct with the isLegacy variable - to avoid duplicating the following code?

pkg/model/resource/utils.go
```diff
@@ -50,6 +50,17 @@ func safeImport(unsafe string) string {
 
 // APIPackagePath returns the default path
 func APIPackagePath(repo, group, version string, multiGroup bool) string {
```
everettraven, line 52: Would it have been an easier transition to add the `isLegacyLayout` boolean to this function signature and handle it that way instead of creating separate functions?

everettraven (changes requested):
First pass. This is a large PR so I didn't get through all the files. I will take a look through any changes and the rest of the PR at the beginning of next week.

erikgb (commented):
I did an initial scan and added a few comments. More adoption of "Standard Go Project Layout" is something I would like to see, but IMO the `controllers` package should be in `internal`, instead of `pkg`, as suggested by @shaneutt in https://github.com/kubernetes-sigs/kubebuilder/issues/932#issuecomment-1255279059.

erikgb:
After sleeping on it, I wonder if using some kind of strategy pattern could make it possible to support multiple layouts. As it seems like users are preferring different layouts, or is this just a false assumption? That could also avoid some complex conditionals in the code - that I personally find hard to read, which are also duplicated to some extend. Also, the legacy/non-legacy layout abstraction is not extendable (add more variants) or sustainable (what to do when the new becomes legacy?). WDYT?

camilamacedo86:
Hi @erikgb,

Thank you for your input. Following the answers inline.

> I did an initial scan and added a few comments. More adoption of "Standard Go Project Layout" is something I would like to see, but IMO the controllers package should be internal, instead of pkg, as suggested by @shaneutt in https://github.com/kubernetes-sigs/kubebuilder/issues/932#issuecomment-1255279059.

We need to have access to the controllers via the main.go .
The main.go should be under cmd and cannot be placed under pkg.
It is not possible to be done. However, if I missing something could you please for example change manually the layout of a sample https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v4 and share with us how do you think that it should be done? 

See: https://github.com/kubernetes-sigs/kubebuilder/issues/932#issuecomment-1264403429

> After sleeping on it, I wonder if using some kind of strategy pattern could make it possible to support multiple layouts. As it seems like users are preferring different layouts, or is this just a false assumption? That could also avoid some complex conditionals in the code - that I personally find hard to read, which are also duplicated to some extend. Also, the legacy/non-legacy layout abstraction is not extendable (add more variants) or sustainable (what to do when the new becomes legacy?). WDYT?

The if conditions If isLagacyLayout should be removed when we are able to remove the go/v3 plugin and these implementations are only required for the tool to ensure backward compatibility until we are able to deprecate and remove the go/v3 plugin (such as done in the past)

It is not really easy for the tool to work with many layouts that bring many complexities. (See: https://github.com/kubernetes-sigs/kubebuilder/issues/932#issuecomment-524984343 and https://github.com/kubernetes-sigs/kubebuilder/issues/932#issuecomment-1264651286). However, users are able to create their own plugins with their own layouts. (see: https://github.com/kubernetes-sigs/kubebuilder/issues/932#issuecomment-1268459400 )

We are here discussing here change the default layout only. However, if anyone would like to try to propose a solution for the tool/plugins works with multiple layouts then it is great. (So far no achievable solution was proposed). On top of that, it is out of our scope here and (ihmo) it would better fit in a design doc (https://github.com/kubernetes-sigs/kubebuilder/tree/master/designs) so we are able to discuss it properly.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/config/samples/crd_sample.go
camilamacedo86: No. That is to ensure the default standard of printing the path when we do the scaffold as done in the other actions.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller-test.go
```diff
@@ -33,19 +33,29 @@ type ControllerTest struct {
...
-			f.Path = filepath.Join("controllers", "%[kind]_controller_test.go")
+			if f.IsLegacyLayout {
+				f.Path = filepath.Join("controllers", "%[kind]_controller_test.go")
+			} else {
+				f.Path = filepath.Join("pkg", "controllers", "%[kind]_controller_test.go")
+			}
 		}
 	}
 	f.Path = f.Resource.Replacer().Replace(f.Path)
+	fmt.Println(f.Path)
```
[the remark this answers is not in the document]
camilamacedo86: No. That is to ensure the default standard of printing the path when we do the scaffold as done in the other actions. See; https://github.com/kubernetes-sigs/kubebuilder/search?q=fmt.Println%28f.Path%29

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/controllers/controller.go
camilamacedo86: No. That is to ensure the default standard of printing the path when we do the scaffold as done in the other actions. See; https://github.com/kubernetes-sigs/kubebuilder/search?q=fmt.Println%28f.Path%29

pkg/plugins/golang/declarative/v1/scaffolds/internal/templates/types.go
```diff
@@ -33,18 +33,31 @@ type Types struct {
...
+			if f.IsLegacyLayout {
+				f.Path = filepath.Join("api", "%[version]", "%[kind]_types.go")
+			} else {
+				f.Path = filepath.Join("pkg", "api", "%[version]", "%[kind]_types.go")
+			}
+
 		}
 	}
 	f.Path = f.Resource.Replacer().Replace(f.Path)
+	fmt.Println(f.Path)
```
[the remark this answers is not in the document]
camilamacedo86: No. That is to ensure the default standard of printing the path when we do the scaffold as done in the other actions. See; https://github.com/kubernetes-sigs/kubebuilder/search?q=fmt.Println%28f.Path%29

pkg/plugins/golang/declarative/v1/scaffolds/internal/templates/controller.go
```diff
@@ -43,6 +44,7 @@ func (f *Controller) SetTemplateDefaults() error {
 		}
 	}
 	f.Path = f.Resource.Replacer().Replace(f.Path)
+	fmt.Println(f.Path)
```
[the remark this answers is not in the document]
camilamacedo86: Yes, it is. That is to ensure the default standard to output the path when we do the scaffold as done in the other actions. See; https://github.com/kubernetes-sigs/kubebuilder/search?q=fmt.Println%28f.Path%29

pkg/plugins/golang/declarative/v1/scaffolds/api.go
camilamacedo86: The Config structure is what we have in the Project file. 
This code checks that the project was scaffolded with the previous plugin versions.
Therefore, we use the info to know if is or not a legacylayout (projects scaffold within versions go/v2 (will be removed in the next+1 release) or go/v3 - v1 is no longer supported by the tool/removed ).

pkg/plugins/golang/declarative/v1/init.go
camilamacedo86: it is under the following condition now.

pkg/plugins/golang/declarative/v1/scaffolds/internal/templates/controller.go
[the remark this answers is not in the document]
camilamacedo86: yep we should print.

erikgb:
> We need to have access to the controllers via the main.go . The main.go should be under cmd and cannot be placed under pkg. It is not possible to be done. However, if I missing something could you please for example change manually the layout of a sample https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v4 and share with us how do you think that it should be done?

@camilamacedo86 Good idea! This should show what I suggest as the new default layout (minimal changes): https://github.com/kubernetes-sigs/kubebuilder/compare/master...erikgb:new-project-structure?expand=1

pkg/plugins/golang/declarative/v1/init.go
camilamacedo86: Because it was wrong. The file update is the Dockerfile and not manager file. 
Could to be in another PR.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
camilamacedo86: We will set the defaults into `func (f *Controller) SetTemplateDefaults() error {` so, we need to have the value there.

camilamacedo86: Note that the "duplication" code here is required because we are initializing different Controller{} implementations one is from v3 and another is from v4. So it is not duplicated actually.

pkg/model/resource/utils.go
camilamacedo86: I though better keep both to after we are able to remove the legacy one instead

camilamacedo86:
Hi @erikgb,

> We need to have access to the controllers via the main.go . The main.go should be under cmd and cannot be placed under pkg. It is not possible to be done. However, if I missing something could you please for example change manually the layout of a sample https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v4 and share with us how do you think that it should be done?
> 
> @camilamacedo86 Good idea! This should show what I suggest as the new default layout (minimal changes): https://github.com/kubernetes-sigs/kubebuilder/compare/master...erikgb:kubebuilder:new-project-structure

Did you try to build? That will not work out:

Screenshot 2022-11-25 at 15 09 31

 $ make build

```cmd/main.go:22:2: use of internal package sigs.k8s.io/kubebuilder/testdata/project-v4/pkg/internal/controllers not allowed```

**Comment by erikgb** [comment]
> Did you try to build? That will not work out:

Now I am not sure what's going on. 😉 Am I doing something wrong? Of course I built it! Tests also work, BTW...
I've also tried this layout on one of our own operators.

````
D:~/projects/github/kubebuilder/testdata/project-v4 (new-project-structure) $ make
test -s /home/erikbo/projects/github/kubebuilder/testdata/project-v4/bin/controller-gen && /home/erikbo/projects/github/kubebuilder/testdata/project-v4/bin/controller-gen --version | grep -q v0.10.0 || \
GOBIN=/home/erikbo/projects/github/kubebuilder/testdata/project-v4/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.10.0
/home/erikbo/projects/github/kubebuilder/testdata/project-v4/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
/home/erikbo/projects/github/kubebuilder/testdata/project-v4/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
go build -o bin/manager cmd/main.go
D:~/projects/github/kubebuilder/testdata/project-v4 (new-project-structure) $

**Comment by camilamacedo86** [comment]
HI @erikgb, 

Thank you for the help. Regards: https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1327665944:

> Now I am not sure what's going on. 😉 Am I doing something wrong? Of course I built it! Tests also work, BTW...
I've also tried this layout on one of our own operators.

I asked for you to provide an example because I thought that I was missing something. But see the following explanation as well. How can we be able to access the /pkg/internal/controllers from /cmd/main ? I might be wrong but it does not seem possible. 

From: https://stackoverflow.com/questions/59342373/use-of-internal-package-not-allowed

> [Internal packages](https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU/edit) (packages that are inside a folder that has an internal folder in their path) can only be imported from packages rooted at the parent of the internal folder.
> 
> E.g. a package pkg/foo/internal/bar can be imported by the package pkg/foo/internal/baz and also from pkg/foo/baz, but cannot be imported by the package pkg nor can it be imported by pkg/bar. This is by design. This is so big, complex packages can be broken into smaller packages without having to expose internals.
> 
> You have to treat internal packages as "private" or non-existent from the "outside".

File: pkg/plugins/golang/declarative/v1/scaffolds/internal/templates/types.go (hunk 14 above)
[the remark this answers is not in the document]
Comment (camilamacedo86) [reply pkg/plugins/golang/declarative/v1/scaffolds/internal/templates/types.go:60]: Yes. it is.

File: testdata/project-v4-declarative-v1/Dockerfile [hunk 16]
```diff
@@ -32,9 +32,6 @@ RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o ma
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/manager .
-# copy channels
-COPY --from=builder /channels /channels
```
varshaprasad96, line 36: Is this change related? Where would channels/ come under new layout? If its just `pkg/channels`, then shouldn't we copy that too or is there any other change in declarative config's layout?

docs/book/src/plugins/go-v4-plugin.md
```diff
@@ -24,6 +24,8 @@ under the [testdata][testdata] directory on the root directory of the Kubebuilde
 - If you are looking to have your project update with the latest version available
 - if you are not targeting k8s versions < `1.16` and `1.20` if you are using webhooks
 - If you are looking to work on with scaffolds which are compatible with k8s `1.25+`
+- If you are looking for the new layout following the [Standard Go Project Layout][standard-go-project] where the "api(s)" and "controller(s)" are scaffold
```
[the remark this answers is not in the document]
camilamacedo86: See : https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1327616257 ( that is not possible )

testdata/project-v3-declarative-v1/Dockerfile
```diff
@@ -32,9 +32,6 @@ RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o ma
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/manager .
-# copy channels
-COPY --from=builder /channels /channels
```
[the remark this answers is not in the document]
camilamacedo86: That is an error I need to fix.

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -43,8 +43,16 @@ layout:
...
-- Create `config/samples/kustomization.yaml` with all CR samples specified. (see for example `testdata/project-v4/config/samples/kustomization.yaml`)
-- Replace the import `admissionv1beta1 "k8s.io/api/admission/v1beta1"` with `admissionv1 "k8s.io/api/admission/v1"` in the webhook test files
+- Create `config/samples/kustomization.yaml` with all Custom Resources samples specified into `config/samples`. _(see for example `testdata/project-v4/config/samples/kustomization.yaml`)_
+- :warning: Replace the import `admissionv1beta1 "k8s.io/api/admission/v1beta1"` with `admissionv1 "k8s.io/api/admission/v1"` in the webhook test files
+- Now, let's perform the layout changes. The `api(s)` and `controller(s)` directories have been moved under a new directory called `pkg`. Therefore, you will need to:
+  - Create a directory called `pkg`
+  - Move the `api` or `apis` (if you are using [multi-group][multi-group] layout) under the `pkg`
+  - Move the `controller` or `controllers` (if you are using [multi-group][multi-group] layout) under the `pkg`
+  - Update the `main.go` imports to look for the new path of your controllers under the `pkg` directory
```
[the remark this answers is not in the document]
camilamacedo86: Good point I need to re-organize this one.

docs/book/src/plugins/go-v4-plugin.md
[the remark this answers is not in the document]
erikgb: I think I understand now. It is not possible trying to put internal under pkg, but that is not what the standard layout suggests. The internal directory should be at the project level. And I see a lot of projects doing that to avoid exporting everything.

The pkg directory actually is more debatable. See https://eli.thegreenplace.net/2019/simple-go-project-layout-with-modules/ as an example.

But I do not have a strong opinion on this. The current layout works for me also. Bottom line, a controller project should in general not be imported by other projects.

erikgb:
Now I finally understand why there has been a lot of discussions in the PR! I think we need to agree on the goal here! I've created an overview of some alternative layouts, but feel free to add more:

````
├── alt-1
│   ├── api
│   ├── controllers
│   └── main.go
├── alt-2
│   ├── cmd
│   │   └── main.go
│   ├── internal
│   │   └── controllers
│   └── pkg
│       └── api
├── alt-3
│   ├── api
│   ├── cmd
│   │   └── main.go
│   └── internal
│       └── controllers
└── alt-4
    └── pkg
        ├── api
        ├── cmd
        │   └── main.go
        └── internal
            └── controllers
````

1) is the current layout, and 2) is closest to following the Go Standard project layout AFAIK. IMO 3) could also be considered to be within the standard, though the api directory should normally contain API specs in e.g. OpenAPI formats. But since the Go api types act as the master for the APIs, I think I could make sense. WDYT?

**Comment by camilamacedo86** [comment]
Hi @erikgb,

Thank you for all help and inputs 
Regards: https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1328026732 

That is how today:
> ├── alt-1
> │   ├── api
> │   ├── controllers
> │   └── main.go

The following one I can go along with it if that works with multigroup layouts. 
We need to ensure that is also possible to access internal/group/controllers/ from the cmd/main.go

├── alt-2
│   ├── cmd
│   │   └── main.go
│   ├── internal
│   │   └── controllers
│   └── pkg
│       └── apis

The following does not follow the standards (ihmo): 

├── alt-3
│   ├── api
│   ├── cmd
│   │   └── main.go
│   └── internal
│       └── controllers

The following one adds the cmd into the pkg which seems not right either. 

└── alt-4
    └── pkg
        ├── api
        ├── cmd
        │   └── main.go
        └── internal
            └── controllers

Then, came the layout change here that was not added to this list:

├── alt-5
│   ├── cmd
│   │   └── main.go
│   └── pkg
│       └── apis
│   │   └── controllers

So, I think the conclusion here would be: 
a) Adopt alt-2 if possible to work with multi-groups
OR
b) alt-5

File: pkg/plugins/golang/v4/init.go (hunk 2 above)
Comment (camilamacedo86) [reply pkg/plugins/golang/v4/init.go:72]: done

File: pkg/plugins/golang/declarative/v1/init.go [hunk 20]
```diff
@@ -51,21 +51,28 @@ func (p *initSubcommand) Scaffold(_ machinery.Filesystem) error {
...
+			return err
+		}
+	} else {
+		// nolint:lll
+		err := insertCodeIfDoesNotExist(dokerfile,
+			"COPY pkg/controllers/ pkg/controllers/",
+			"\n# https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/docs/addon/walkthrough/README.md#adding-a-manifest\n# Stage channels and make readable\nCOPY channels/ /channels/\nRUN chmod -R a+rx /channels/")
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
camilamacedo86: done

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
camilamacedo86: done

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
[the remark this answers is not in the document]
camilamacedo86: Sort it out. Thx

pkg/plugins/golang/declarative/v1/scaffolds/internal/templates/types.go
```diff
@@ -33,18 +33,31 @@ type Types struct {
...
+
+	IsLegacyLayout bool
 }
 
 // SetTemplateDefaults implements file.Template
 func (f *Types) SetTemplateDefaults() error {
 	if f.Path == "" {
 		if f.MultiGroup {
-			f.Path = filepath.Join("apis", "%[group]", "%[version]", "%[kind]_types.go")
+			if f.IsLegacyLayout {
```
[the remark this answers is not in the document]
camilamacedo86: Hi @varshaprasad96 I understand your suggestion here.
The if/else here is also because we need create an diff path for multigroup
I go along that it has a hall for improvements but them the refac will be bigger.

Could we agree in keep it as it is today already ( indeed before these changes ) and see if we can improve in a follow-up?

everettraven (changes requested):
Second round. Some smaller nits, but blocking specifically for the docs nits. I think it is important that we keep the docs as accurate as possible with this change since it is a pretty significant shift.

Other than these nits it looks great! Great work @camilamacedo86 !

pkg/plugins/golang/declarative/v1/scaffolds/internal/templates/types.go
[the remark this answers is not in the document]
varshaprasad96: Sure, we can do in a follow up. Just wanted to make sure that we make the conditionals as modular as possible, so that its easier to refactor when we remove support to the previous format ever in future.

testdata/project-v4-with-deploy-image/config/crd/patches/cainjection_in_busyboxes.yaml
```diff
@@ -3,5 +3,5 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
+    cert-manager.io/inject-ca-from: CERTIFICATE_NAMESPACE/CERTIFICATE_NAME
```
varshaprasad96, line 6: Is this change relevant?

camilamacedo86: I am not sure why it is not in the master branch. 
The v4 is using v2-alpha and then it should be in this away.

pkg/plugin/util/testdata/exampleFile.txt
```diff
@@ -1 +1 @@
-exampleTargetexampleCodeexampleCode
\ No newline at end of file
+exampleTargetexampleCodeexampleCodeexampleCode
```
camilamacedo86, line 1:
```suggestion
exampleTargetexampleCodeexampleCode
```

It has been updated via the script. 
It is something that we could fix/improve
This file just need to have any text for the test

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,164 @@ layout:
...
+CRDDirectoryPaths:     []string{filepath.Join("..", "config", "crd", "bases")},
+```
+
+With:
+
+```
+CRDDirectoryPaths:     []string{filepath.Join("..", "..", "config", "crd", "bases")},
+```
+
+Note that if your project has multiple groups (`multigroup:true`) then the above updated should result into `"..", "..", "..",` instead of `"..",".."`
```
umangachapagain, line 145:
```suggestion
Note that if your project has multiple groups (`multigroup:true`) then the above update should result into `"..", "..", "..",` instead of `"..",".."`
```

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,164 @@ layout:
...
+
+#### Now, let's update the PATHs in the PROJECT file accordingly
+
+The PROJECT tracks the paths of all APIs used into your project ensure that they will
+now be pointing out for `pkg/api/...` as the following example:
```
umangachapagain, line 150:
```suggestion
The PROJECT tracks the paths of all APIs used in your project. Ensure that they now point to `pkg/api/...` as the following example:
```

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,164 @@ layout:
...
+
+Before update:
+```
```
umangachapagain, line 153: Need a new line before code block for consistency?

docs/book/src/migration/v3vsv4.md
```diff
@@ -20,6 +20,8 @@ The details of all changes (breaking or otherwise) can be found in:
 - no longer scaffold webhook test files with `"k8s.io/api/admission/v1beta1"` the k8s API which is no longer served since k8s `1.25`. By default
   webhooks test files are scaffolding using `"k8s.io/api/admission/v1"` which is support from k8s `1.20`
 - no longer provide backwards compatible support with k8s versions < `1.16`
+- change the layout to accommodate the community request to follow the [Standard Go Project Layout][standard-go-project] 
+by moving the api(s)/controller(s) under a new directory called `pkg` and the `main.go` under a new directory named `cmd`
```
umangachapagain, line 24:
```suggestion
by moving the api(s) under a new directory called `pkg`, controller(s) under a new directory called `internal` and the `main.go` under a new directory named `cmd`
```

pkg/plugins/golang/deploy-image/v1alpha1/api.go
```diff
@@ -184,9 +184,20 @@ func (p *createAPISubcommand) PreScaffold(machinery.Filesystem) error {
...
+		if strings.Contains(pluginKey, "go.kubebuilder.io/v3") {
+			isGoV3 = true
+		}
+	}
+
+	defaultMainPath := "cmd/main.go"
+	if isGoV3 {
+		defaultMainPath = "main.go"
+	}
 	// check if main.go is present in the root directory
```
umangachapagain, line 198:
```suggestion
	// check if main.go is present in the cmd/ directory
```

pkg/plugins/golang/deploy-image/v1alpha1/api.go
```diff
@@ -184,9 +184,20 @@ func (p *createAPISubcommand) PreScaffold(machinery.Filesystem) error {
...
-	if _, err := os.Stat(DefaultMainPath); os.IsNotExist(err) {
-		return fmt.Errorf("%s file should present in the root directory", DefaultMainPath)
+	if _, err := os.Stat(defaultMainPath); os.IsNotExist(err) {
+		return fmt.Errorf("%s file should present in the root directory", defaultMainPath)
```
umangachapagain, line 200:
```suggestion
		return fmt.Errorf("%s file should be present in the cmd/ directory", defaultMainPath)
```

umangachapagain (changes requested):

pkg/plugins/golang/deploy-image/v1alpha1/api.go
camilamacedo86: For now, can be both.

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,163 @@ layout:
...
+
+Note that if your project has multiple groups (`multigroup:true`) then the above update should result into `"..", "..", "..",` instead of `"..",".."`
+
+#### Now, let's update the PATHs in the PROJECT file accordingly
+
+The PROJECT tracks the paths of all APIs used in your project. Ensure that they now point to `pkg/api/...` as the following example:
+
+Before update:
+```
```
camilamacedo86, line 152:
```suggestion

```
```

File: docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md [hunk 31]
```diff
@@ -35,34 +41,163 @@ layout:
...
+used by the project such as [controller-runtime][controller-runtime] and [controller-tools][controller-tools].
 
 </aside>
 
+### If you have webhooks:
+
+Replace the import `admissionv1beta1 "k8s.io/api/admission/v1beta1"` with `admissionv1 "k8s.io/api/admission/v1"` in the webhook test files
+
+### Makefile updates
+Update the Makefile with the changes which can be found in the samples under testdata for the release tag used. (see for example `testdata/project-v4/Makefile`)
```
camilamacedo86, line 194:
```suggestion

Update the Makefile with the changes which can be found in the samples under testdata for the release tag used. (see for example `testdata/project-v4/Makefile`)
```

pkg/plugins/golang/deploy-image/v1alpha1/api.go
```diff
@@ -184,9 +184,20 @@ func (p *createAPISubcommand) PreScaffold(machinery.Filesystem) error {
...
+		}
+	}
+
+	defaultMainPath := "cmd/main.go"
+	if isGoV3 {
+		defaultMainPath = "main.go"
+	}
+	// check if main.go is present in the cmd/ directory
+	if _, err := os.Stat(defaultMainPath); os.IsNotExist(err) {
+		return fmt.Errorf("%s file should be present in:", defaultMainPath)
```
camilamacedo86, line 200:
```suggestion
		return fmt.Errorf("main.go file should be present in %s", defaultMainPath)
```

testdata/project-v4-with-deploy-image/config/crd/patches/cainjection_in_busyboxes.yaml
camilamacedo86: See we should not have the `$(` (i.e https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v4-config/config/crd) but we have a bug in the master for the deploy image plugin (https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3-with-deploy-image/config/crd/patches/cainjection_in_memcacheds.yaml#L6) which are probably fixing here: https://github.com/kubernetes-sigs/kubebuilder/pull/2985/files#diff-deedc95e5fb2c50078118e8e46f65c026e90b6d7670b7b46ee5f7f2b285814f4R288-R306

camilamacedo86: @varshaprasad96 we have a bug in master, see: https://github.com/kubernetes-sigs/kubebuilder/pull/3104/files

camilamacedo86:
/hold cancel

It seems good enough to get merged now.
Thank you for all helps and reviews

pkg/model/resource/utils.go
```diff
@@ -50,6 +50,17 @@ func safeImport(unsafe string) string {
...
+	if multiGroup {
+		if group != "" {
+			return path.Join(repo, "pkg", "apis", group, version)
+		}
+		return path.Join(repo, "pkg", "apis", version)
+	}
+	return path.Join(repo, "pkg", "apis", version)
+}
+
+// APIPackagePathLegacy returns the default path
```
umangachapagain, line 62:
```suggestion
// APIPackagePathLegacy returns the default legacy path
```

pkg/plugins/golang/declarative/v1/scaffolds/internal/templates/types.go
```diff
@@ -33,18 +33,30 @@ type Types struct {
...
+				f.Path = filepath.Join("apis", "%[group]", "%[version]", "%[kind]_types.go")
+			} else {
+				f.Path = filepath.Join("api", "%[version]", "%[kind]_types.go")
+			}
 		} else {
-			f.Path = filepath.Join("api", "%[version]", "%[kind]_types.go")
+			if f.MultiGroup {
+				f.Path = filepath.Join("pkg", "apis", "%[group]", "%[version]", "%[kind]_types.go")
+			} else {
+				f.Path = filepath.Join("pkg", "apis", "%[version]", "%[kind]_types.go")
```
umangachapagain, line 53: will it be `/pkg/apis` and not `/pkg/api` even if it is not MultiGroup?

camilamacedo86: yes, that sort outs the old bug. That should never be api and apis at the same way that controllers is only controllers in both scenarios. 

See that make no sense we have api in the singular (the dir can have 1 to many independently ) and it brings a lot of complexities which was reduced in this PR for the new layout. It is something described in the migration doc as well.

camilamacedo86: Closing this one as sorted out @umangachapagain,
But feel free to re-open or ping me if you need.

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,163 @@ layout:
...
+├── internal
+│ └── controllers
+└── pkg
+    └── apis
+```
+
+##### In this way, you will need to:
+
+- Create a new directory `cmd` and move the `main.go` under it.
+- If your project does not support multi-group the APIs are scaffold under a directory called `api`. Rename this diretory to `apis`
```
varshaprasad96, line 68: nit: ```suggestion
- If your project does not support multi-group the APIs are scaffolded under a directory called `api`. Rename this diretory to `apis`
```

File: docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md [hunk 36]
```diff
@@ -35,34 +41,163 @@ layout:
...
+- Move the `apis` directory under the `pkg`
+- Move the `controllers` directory under the `internal`
```
varshaprasad96, line 70: Nit: 
```suggestion
- Move the `apis` directory under `pkg/`
- Move the `controllers` directory under `internal/` 
```

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,163 @@ layout:
...
+    └── apis
+```
+
+##### In this way, you will need to:
+
+- Create a new directory `cmd` and move the `main.go` under it.
+- If your project does not support multi-group the APIs are scaffold under a directory called `api`. Rename this diretory to `apis`
+- Move the `apis` directory under the `pkg`
+- Move the `controllers` directory under the `internal` 
+- Now ensure that the imports will be updated accordingly by:
```
varshaprasad96, line 71: Nit: ```suggestion
- Now ensure that the imports will need to be updated accordingly by:
```

File: docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md [hunk 38]
```diff
@@ -35,34 +41,163 @@ layout:
...
+  - Update the `main.go` imports to look for the new path of your controllers under the `pkg` directory
```
varshaprasad96, line 72:
```suggestion
  - Updating the `main.go` imports to look for the new path of your controllers under the `pkg` directory
```

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,163 @@ layout:
...
+  - Update all controllers imports to look for the new path of your apis under the `pkg` directory
```
varshaprasad96, line 73:
```suggestion
  - Updating all imports in controllers/ to look for the new path of your apis under the `pkg` directory
```

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,163 @@ layout:
...
+```
+  group: crew
+  kind: Captain
+  path: sigs.k8s.io/kubebuilder/testdata/project-v4/apis/crew/v1
+```
+
+### Update kustomize manifests with the changes made so far
+
+- Update the manifest under `config/` directory with all changes performed in the default scaffold done with `go/v4-alpha` plugin. (see for example `testdata/project-v4/config/`) to get all changes in the
+  default scaffolds to be applied on your project
```
varshaprasad96, line 168:
```suggestion
- Update the manifest under `config/` directory with all changes performed in the default scaffold done with `go/v4-alpha` plugin (see the example `testdata/project-v4/config/`) to get all changes in the
  default scaffolds to be applied on your project
```

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,163 @@ layout:
...
+- Create `config/samples/kustomization.yaml` with all Custom Resources samples specified into `config/samples`. _(see for example `testdata/project-v4/config/samples/kustomization.yaml`)_
```
varshaprasad96, line 169:
```suggestion
- Create `config/samples/kustomization.yaml` with all Custom Resources samples specified into `config/samples`. _(see the example `testdata/project-v4/config/samples/kustomization.yaml`)_
```

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,163 @@ layout:
...
 
 <aside class="warning">
 <h1>`config/` directory with changes into the scaffold files</h1>
 
 Note that under the `config/` directory you will find scaffolding changes since using
 `go/v4-alpha` you will ensure that you are no longer using Kustomize v3x.
 
-You can mainly compare the `config/` directory from the samples scaffolded under the `testdata`directory by 
+You can mainly compare the `config/` directory from the samples scaffolded under the `testdata`directory by
```
varshaprasad96, line 177: Nit:
```suggestion
You can mainly compare the `config/` directory from the samples scaffoldeded under the `testdata`directory by
```

varshaprasad96 (commented):
The PR looks great! Just a few nits on documentation.
Thanks @camilamacedo86 for working on this!
Its /lgtm and /approve from my end. I'm just putting on hold if anyone else likes to review (so that ci does not automatically merge it). @camilamacedo86 please feel free to remove the hold and merge it :))
/hold

joelanford:
I'm personally not a huge fan of moving `api`/`apis` under `pkg`.

IMO the entire point of the APIs is that they are meant to be used by external users. As such, burying the APIs inside a `pkg` directory makes them less discoverable, and it potentially mixes them with other library code an operator author might decide needs to exist.

Personally, I'd rather see something like this:
```
├── apis
│   ├── v1
│   └── v1alpha1
├── cmd
│   └── controller
│       └── main.go
├── internal
│   └── controllers
└── pkg
    ├── client
    ├── image
    ├── log
    └── util
```

NOTE: two tweaks I'm suggesting are:
- leave `apis` at the root for discoverability
- put `main.go` in `cmd/controller` (this makes it easier to add other binaries alongside `cmd/controller`)

erikgb:
I must say that I largely agree with the last comment from @joelanford (https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341033815). IMO we should also use the singular word (`api`) on the api directory, regardless of one/many APIs. Having so many "nobs" just adds a lot of complexity in the code and for users.

camilamacedo86:
HI @shaneutt, @A-Hilaly, @jakobmoellersap, @moserke, 

Could you please let us know wdyt about https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341033815? Is that make sense? Are you OK with we keep the apis in the root instead of create the pkg?

mhrivnak:
Some operator/controller projects have found it necessary to make the `api` package a separate go module. That enables other projects (such as another operator) to import the structs without bringing in all of the dependencies of the controller, which can often cause a lot of problems if the two projects don't happen to align on their dependencies.

I suggest not moving the `api` package, so that this pattern can continue to be possible. I think it's even worth considering as the default, to make the controller's APIs more usable.

camilamacedo86:
Hi @mhrivnak,

https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341074263

Could you please clarifies your comment? What do you mean with `I suggest not moving the api package, so that this pattern can continue to be possible`? 

It seems for me that you are worried about change the layout can broke projects. If so, the layout changes in for the next alpha plugin version so it will not break pre-existent projects. In the future, users will need at some point migrate their project to the go/v4-alpha and then follow the migration guide. So, if that is your concern (impacted in the existent projects) then. that does not seems a point valid in the scope of the changes.

mhrivnak:
> Could you please clarifies your comment? What do you mean with I suggest not moving the api package, so that this pattern can continue to be possible?

I think the `api` package should stay at the root of the repo, so that projects will continue to have the option to make it an independent go module.

camilamacedo86:
Hi @joelanford, @erikgb,

I understand that the initial thought was keep the apis and controllers under root for discoverability as described in https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341033815. However, users prefer that their project follows the Golang Standards https://github.com/kubernetes-sigs/kubebuilder/issues/932 (20 votes). So following some questions:

- Therefore, have the apis under the pkg really makes harder for the users discovery them out?  
- Also, would really make sense have the apis in the root and the pkg instead of all together? 
- Why/when we would like to import the project and not the APIs? 

My 2 cents here is that if we do not add the apis under pkg in the practical terms what will happens is:
- a) People will change their scaffold to follow Golang common Standards
- b) Then will be harder for they update their project with the changes made and etc
- c) They will raise another issue that will be voted a lot to move the apis under the pkg/

So, I still more inclined to go along with  https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341135881. But I would like to get your answers and further thoughts if possible as well.

pmalek:
Dropping my 2 cents as well into the mix: I don't have a hard opinion on putting `apis` under `pkg/` or not but I do see a benefit of doing so my making it "standard" as per the #932 and also thanks to separation of directories in a way that:

- we have `cmd/` for app's entrypoints (`main.go`, flags and such)
- `pkg/` for all the code that is exportable and usable from outside
- `internal/` where folks can put their controllers code and anything else that they don't want to share outside (make part of the API contract)

Hence my slight 👍 on putting `apis` into `pkg/`.

mhrivnak:
> However, users prefer that their project follows the Golang Standards https://github.com/kubernetes-sigs/kubebuilder/issues/932 (20 votes).

It's good to be clear that [there's not a standard, just conventions](https://github.com/golang-standards/project-layout#overview).

"It's not an official standard defined by the core Go dev team; however, it is a set of common historical and emerging project layout patterns in the Go ecosystem. Some of these patterns are more popular than others."

That pattern also describes many root-level directories, including [an `/api` directory](https://github.com/golang-standards/project-layout#api).

shaneutt:
> I think the `api` package should stay at the root of the repo, so that projects will continue to have the option to make it an independent go module.

I think it may be important to note that being under `pkg/` does not _preclude_ the notion of doing this.

> a) People will change their scaffold to follow Golang common Standards
> b) Then will be harder for they update their project with the changes made and etc
> c) They will raise another issue that will be voted a lot to move the apis under the pkg/

If we make the kubebuilder standard different than the go standard, then I think these are true: we'll just continue to repeat the cycle by not having aligned standards.

pmalek:
> > I think it may be important to note that being under pkg/ does not preclude the notion of doing this.
> 
> I might be behind on the times! Is it possible now to have a go module nested within another go module? I'm not sure how that would work. At a minimum it seems like it would be confusing, but I'm interested to know if and how people are doing this.

You might want to look at a prominent example in the wild: https://github.com/open-telemetry/opentelemetry-collector-contrib with all the processors, exporters and receivers having their own respective modules e.g. https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourceprocessor (+ the naming of the tags for nested modules).

RealHarshThakur:
I feel like having `apis` under `pkg/` might get a bit weird if it ever got its own [IDL](https://github.com/kubernetes/enhancements/pull/2310)(and I do hope it happens someday to make collobarate with other language ecosystems). In the gRPC/proto world, I've seen usally the IDL in a different folder than the server stub and client.

joelanford:
From the linked Go conventions doc about `/pkg`:
> This is a common layout pattern, but it's not universally accepted and some in the Go community don't recommend it.
>
> It's ok not to use it if your app project is really small and where an extra level of nesting doesn't add much value (unless you really want to :-)). Think about it when it's getting big enough and your root directory gets pretty busy (especially if you have a lot of non-Go app components).

So even the official conventions docs hesitate to really get behind it as a convention.

My take is that kubebuilder is scaffolding out fairly small projects. Would `kubebuilder init` or `kubebuilder create api` scaffold anything in `/pkg` other than `/pkg/api`? To me, this falls under what the doc says is "an extra level of nesting that doesn't add value", and IMO it slightly devalues the project because it hides the _one_ thing that will be importable from other projects.

I could _maybe_ see an argument for nesting APIs under `pkg` if kubebuilder also scaffolded other library code intended for external consumption.

And maybe it seems trivial, but the extra segment of the import path means consumers of the API, IDEs, etc. have to jump through one more directory (or poke around a little more on pkg.go.dev) before they find what they're looking for.

I'm not strongly against this (I see examples of both styles in upstream kubernetes), but my preference is for the APIs to live at the root.

Maybe as a compromise, we could make some of these directories relocatable. Default to `/api`, but add some flags to `kubebuilder init` that allow `/api` to be renamed to `/pkg/apis` or `/joes/weird/path/to/myapis`, and then store that choice in the PROJECT file so that future runs of `kubebuilder create api` would know where to do its thing?

camilamacedo86:
Hi @joelanford, 

Thank you a lot for all input. 

> Regards the comment https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341952427 I do not see how we would we could make it configurable and maintainable. However, I think that would be great  if possible but would be better it be properly discussed via a design/PE doc. In this away, I think the goal for now is only define what is our default layout according what has been requested by the community.

So, I will try to get all feedbacks and see if we can convey and seems for me that keep the api in the root direct wons here:

**Based on the comments**
- https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341033815
- https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341952427
- https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341610812
- https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341177421
- https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341048965

*And based on the answer of: If we add the apis/ under root would we still following the Golang Standard?** 

That pattern also describes many root-level directories, including [an /api directory](https://github.com/golang-standards/project-layout#api).

PS: IMO also makes more sense since import the apis from on operator project to another seems double as a good approach in the common cases scenarios either. 

In this I will be working on this one to address the agreement. 
:tada: Thank you a lot for all contributions.

camilamacedo86:
I think we could achieve and convey here: 

**Based on the comments**
- https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341033815
- https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341952427
- https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341610812
- https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341177421
- https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1341048965

*And based on the answer of: If we add the apis/ under root would we still following the Golang Standard?** 

That pattern also describes many root-level directories, including [an /api directory](https://github.com/golang-standards/project-layout#api).

> Obs.:IMO also makes more sense have the api in the root when I thinking about that import the apis from on operator project to another seems double as a good approach in the common cases scenarios either. 

In this I will be working on this one to address the agreement. 
:tada: Thank you a lot for all contributions.

jakobmoellersap:
Pretty late to the party, but I still want to summarise a few aspects I would like to Bring into consideration from my experience with adopting Kubebuilder on org-level:

1. In enterprise development, abiding by any standard is better than having no standard. In kubebuilder projects, it can be argument that having an "api" folder has become standard and it would be a huge breaking change in terms of layout and imports. => I would support keeping /api.
2. At the same time, it will be hard to argument that all exported packages should be consumable from a user of an operator. I think there are still cases where the usage of /internal is reasonable. Simply because there will never be a case where the control-loop should be consumed as a library. => I propose using internal for things that should not be leaked outside of the module, e.g. the control loop. 
3. Even though `pkg` is not included by default, I think it is really easy to write custom library code on top of what Kubebuilder offers and then consumer that from `internal`. At the same time, it is easy to write unexpected code which can only be used from within `pkg`. => By default, there is no need in the scaffolding to have exported `pkg` style libraries and imo we do not need to overcomplicate the setup. Instead we can leave it up to the users wether they want to maintain that library.

All of the three above would only mean one suggested change for me: moving the control loop into `internal` so it cannot be consumed from outside the module.

camilamacedo86:
Hi @jakobmoellersap,

Thank you for your input : https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1357671546

The convey made above  https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1342281605 is aligned with your POV. 

> If there is a design doc pending somewhere, would you mind sharing that process in the community?

No. It does not have a design doc because we are not adding a new feature or introducing a new design. We have a new golang plugin, go/v4, in alpha, and the discussion here is only about the layout changes for the new plugin which has been requested for a long period https://github.com/kubernetes-sigs/kubebuilder/issues/932. You can check the discussions in the issue and in this PR. I updated the issue as well to reflect them.

erikgb:
/hold
/lgtm

Nice work, @camilamacedo86! 👏 The suggested new structure summarized in the PR description is what I prefer. Putting the PR on hold for now, to allow for additional reviews.

docs/book/src/plugins/go-v4-plugin.md
```diff
@@ -24,6 +24,8 @@ under the [testdata][testdata] directory on the root directory of the Kubebuilde
 - If you are looking to have your project update with the latest version available
 - if you are not targeting k8s versions < `1.16` and `1.20` if you are using webhooks
 - If you are looking to work on with scaffolds which are compatible with k8s `1.25+`
+- If you are looking for the new layout following the [Standard Go Project Layout][standard-go-project] where 
+the "api(s)" are scaffold under the `pkg/` directory, "controller(s)" under `internal`, and the `main.go` under `cmd`
```
everettraven, line 28: I think this line is no longer accurate:
```suggestion
the "api(s)" are scaffold under the `api` directory, "controller(s)" under `internal`, and the `main.go` under `cmd`
```

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,151 @@ layout:
 
 ```
 
-### Steps to migrate
+#### Changes the layout
```
everettraven, line 44: Nit:
```suggestion
#### Changes to the layout
```

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,151 @@ layout:
...
-- Update the Makefile with the changes which can be found in the samples under testdata for the release tag used. (see for example `testdata/project-v4/Makefile`)
-- Update the `go.mod` with the changes which can be found in the samples under `testdata` for the release tag used. (see for example `testdata/project-v4/go.mod`). Then, run
-`go mod tidy` to ensure that you get the latest dependencies and your Golang code has no breaking changes.
-- Update the manifest under `config/` directory with all changes performed in the default scaffold done with `go/v4-alpha` plugin. (see for example `testdata/project-v4/config/`) to get all changes in the 
-default scaffolds to be applied on your project
-- Create `config/samples/kustomization.yaml` with all CR samples specified. (see for example `testdata/project-v4/config/samples/kustomization.yaml`)
-- Replace the import `admissionv1beta1 "k8s.io/api/admission/v1beta1"` with `admissionv1 "k8s.io/api/admission/v1"` in the webhook test files
+##### Now, let's perform the layout changes to comply within:
```
everettraven, line 46: Nit:
```suggestion
##### New layout:
```

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,151 @@ layout:
...
+```sh
+...
+├── cmd
+│ └── main.go
+├── internal
+│ └── controller
+└── api
+```
+
+##### In this way, you will need to:
```
everettraven, line 63: Nit:
```suggestion
##### Migrating to the new layout:
```

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,151 @@ layout:
...
+
+- Create a new directory `cmd` and move the `main.go` under it.
+- If your project support multi-group the APIs are scaffold under a directory called `apis`. Rename this directory to `api`
+- Move the `controllers` directory under the `internal` and rename it for `controller` 
+- Now ensure that the imports will be updated accordingly by:
+  - Update the `main.go` imports to look for the new path of your controllers under the `pkg` directory
+  - Update all controllers imports to look for the new path of your apis under the `pkg` directory
```
everettraven, line 70: I think this line here is outdated since the `api` directory will no longer be placed under the `pkg` directory.
```suggestion

```

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,151 @@ layout:
...
+
+```
+CRDDirectoryPaths:     []string{filepath.Join("..", "..", "config", "crd", "bases")},
+```
+
+Note that if your project has multiple groups (`multigroup:true`) then the above update should result into `"..", "..", "..",` instead of `"..",".."`
+
+#### Now, let's update the PATHs in the PROJECT file accordingly
+
+The PROJECT tracks the paths of all APIs used in your project. Ensure that they now point to `pkg/api/...` as the following example:
```
everettraven, line 137:
```suggestion
The PROJECT tracks the paths of all APIs used in your project. Ensure that they now point to `api/...` as the following example:
```

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,151 @@ layout:
...
+
+Before update:
+  group: crew
+  kind: Captain
+  path: sigs.k8s.io/kubebuilder/testdata/project-v4/api/crew/v1
```
everettraven, line 142:
```suggestion
  path: sigs.k8s.io/kubebuilder/testdata/project-v4/apis/crew/v1
```

docs/book/src/migration/manually_migration_guide_gov3_to_gov4.md
```diff
@@ -35,34 +41,151 @@ layout:
...
+```
+
+After Update:
+
+```
+  group: crew
+  kind: Captain
+  path: sigs.k8s.io/kubebuilder/testdata/project-v4/apis/crew/v1
```
everettraven, line 150:
```suggestion
  path: sigs.k8s.io/kubebuilder/testdata/project-v4/api/crew/v1
```

docs/book/src/migration/v3vsv4.md
```diff
@@ -20,6 +20,8 @@ The details of all changes (breaking or otherwise) can be found in:
 - no longer scaffold webhook test files with `"k8s.io/api/admission/v1beta1"` the k8s API which is no longer served since k8s `1.25`. By default
   webhooks test files are scaffolding using `"k8s.io/api/admission/v1"` which is support from k8s `1.20`
 - no longer provide backwards compatible support with k8s versions < `1.16`
+- change the layout to accommodate the community request to follow the [Standard Go Project Layout][standard-go-project] 
+by moving the api(s) under a new directory called `pkg`, controller(s) under a new directory called `internal` and the `main.go` under a new directory named `cmd`
```
everettraven, line 24: I think this line is no longer accurate:
```suggestion
by moving the api(s) under a new directory called `api`, controller(s) under a new directory called `internal` and the `main.go` under a new directory named `cmd`
```

everettraven (commented):
Just some minor docs nits

camilamacedo86:
Hi @erikgb, 

I am removing the hold because this one will need to be forced. (we API diff fails for the changes in the alpha go plugin). So, it is safe. Also, we have a lot of OKs and approvals already, and we did what was agreed here, but let's wait a little to see if others come up. 

/hold  cancel

wI2L:
I'd like to add one more thing regarding the move of the `main.go` file to the `cmd/` directory. While I don't have a strong opinion about the name of the file itself, I think it is idiomatic to put the "main" Go file in a sub-directory of `cmd/` which is named after the binary it should produce: `cmd/foo/main.go`.

Whether this would default to the name of the project, or something generic like `operator` is open to discussion. Personally, I would prefer that it follows the name of the project.

camilamacedo86:
Hi @wI2L,

> I'd like to add one more thing regarding the move of the main.go file to the cmd/ directory. While I don't have a strong opinion about the name of the file itself, I think it is idiomatic to put the "main" Go file in a sub-directory of cmd/ which is named after the binary it should produce: cmd/foo/main.go.
> Whether this would default to the name of the project, or something generic like operator is open to discussion. Personally, I would prefer that it follows the name of the project.

thank you for your input but that is out of the scope ihmo.
It was never considered in the proposal discussion.

If you would like that could you please raise an issue and also add why do think that should be made?
What is your motivation to look for that? what docs/references would describe it as good practice?
Why would we need to create another subdirectory by default inside of the cmd? 

Thank you a lot for your attention and input.

camilamacedo86:
I am moving forward with this one since:

a) It changes the default layout only for go/v4-alpha as conveyed in the discussion: https://github.com/kubernetes-sigs/kubebuilder/pull/2985#issuecomment-1342281605 (@joelanford suggestion)

b) It has the lgtm from 
- @[umangachapagain](https://github.com/umangachapagain)
umangachapagain
- @[Kavinjsir](https://github.com/Kavinjsir)
- @erikgb 
- @everettraven 
- @Jakob Möller
- @varshaprasad96 

Therefore, it seems OK to move forward, but we can address any change/improvement in a follow if required.

Issue #2874, opened 2022-08-15
Document helpers and options to write plugins


camilamacedo86:
### What do you want to happen?

Would be nice if we could add to docs the following info to help those that are looking to build their own plugins: 

If you are looking to create an external plugin that will add code on top of what is scaffold by default with the create API see that we do that [in deploy-image here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/deploy-image/v1alpha1/api.go) and you have many ways to achieve this goal:

- a) You can overwrite the scaffold done with the go/v3 (for example) by using this option: https://github.com/kubernetes-sigs/kubebuilder/blob/df1ed6ccf19df40bd929157a91eaae6a9215bfc6/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/internal/templates/api/types.go#L58

- b) You can use the plugin utils to inject and/or replace code, see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go#L170-L266

- c) Note that you can call create api --plugins=go/v3,yourplugin/v1 and just ensure that the create api subcommand will overwrite or you can import the code as we decide to do in the deploy image to allow re-use it and make eaiser for users just call create api --plugins=deploy.image/v1alpha1, see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go#L77-L98

Another option also is to use the Bundle Plugin which will build a new plugin done by composition that does the call in chain as we can do by --plugins=a/v1,b/v2 and, see: https://book.kubebuilder.io/plugins/creating-plugins.html

```go
  mylanguagev1Bundle, _ := plugin.NewBundle(language.DefaultNameQualifier, plugin.Version{Number: 1},
        kustomizecommonv1.Plugin{}, // extend the common base from Kuebebuilder
        mylanguagev1.Plugin{}, // your plugin language which will do the scaffolds for the specific language on top of the common base
    )
```

Motivated by: https://github.com/kubernetes-sigs/kubebuilder/issues/2871

### Extra Labels

/kind documentation

camilamacedo86:
Hi @Kavinjsir, 

This one is to doc how to use the helper and options. You did some docs already, so you might want to look if we can improve the docs for we finish this one.

Kavinjsir:
[...]

PR #3008, opened 2022-10-07
📖  docs: supplement plugin creation

Kavinjsir:
### Description

Add a section in `create-plugin` to provide references for users who create the plugin to scaffold on top of what `create api` brings.

### Motivation

Resolve: https://github.com/kubernetes-sigs/kubebuilder/issues/2874

### DIFF
--- a/docs/book/src/plugins/creating-plugins.md
+++ b/docs/book/src/plugins/creating-plugins.md
@@ -12,6 +12,40 @@
 
 You can extend the Kubebuilder API to create your own plugins. If [extending the CLI][extending-cli], your plugin will be implemented in your project and registered to the CLI as has been done by the [SDK][sdk] project. See its [CLI code][sdk-cli-pkg] as an example.
 
+## When it is useful?
+
+- If you are looking to create plugins which support and work with another language.
+- If you would like to create helpers and integrations on top of the scaffolds done by the plugins provided by Kubebuiler.
+- If you would like to have customized layouts according to your needs.
+
+## How the plugins can be used?
+
+Kubebuilder provides a set of plugins to scaffold the projects, to help you extend and re-use its implementation to provide additional features.
+For further information see [Available Plugins][available-plugins].
+
+Therefore, if you have a need you might want to propose a solution by adding a new plugin
+which would be shipped with Kubebuilder by default.
+
+However, you might also want to have your own tool to address your specific scenarios and by taking advantage of what is provided by Kubebuilder as a library.
+That way, you can focus on addressing your needs and keep your solutions easier to maintain.
+
+Note that by using Kubebuilder as a library, you can import its plugins and then create your own plugins that do customizations on top.
+For instance, `Operator-SDK` does with the plugins [manifest][operator-sdk-manifest] and [scorecard][operator-sdk-scorecard] to add its features.
+Also see [here][operator-sdk-plugin-ref].
+
+Another option implemented with the [Extensible CLI and Scaffolding Plugins - Phase 2][plugins-phase2-design-doc] is
+to extend Kibebuilder as a LIB to create only a specific plugin that can be called and used with
+Kubebuilder as well.
+
+<aside class="note">
+<H1> Plugins proposal docs</H1>
+
+You can check the proposal documentation for better understanding its motivations. See the [Extensible CLI and Scaffolding Plugins: phase 1][plugins-phase1-design-doc],
+the [Extensible CLI and Scaffolding Plugins: phase 1.5][plugins-phase1-design-doc-1.5] and the [Extensible CLI and Scaffolding Plugins - Phase 2][plugins-phase2-design-doc]
+design docs. Also, you can check the [Plugins section][plugins-section].
+
+</aside>
+
 ## Language-based Plugins
 
 Kubebuilder offers the Golang-based operator plugins, which will help its CLI tool users create projects following the [Operator Pattern][operator-pattern].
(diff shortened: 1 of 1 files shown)


docs/book/src/plugins/creating-plugins.md
```diff
@@ -64,3 +64,46 @@ Feel free to check the implementation under:
 
 - deploy-image: <https://github.com/kubernetes-sigs/kubebuilder/tree/v3.7.0/pkg/plugins/golang/deploy-image/v1alpha1>
 - grafana: <https://github.com/kubernetes-sigs/kubebuilder/tree/v3.7.0/pkg/plugins/optional/grafana/v1alpha>
+
+## Plugin scaffolding
+
+Your plugin may have the use case to add code on top of what is scaffolded by default with `create api`.
```
camilamacedo86, line 70: It is not only about create api. The utils provided are valid for any scaffold/action and call made in a chain.

docs/book/src/plugins/creating-plugins.md
```diff
@@ -64,3 +64,46 @@ Feel free to check the implementation under:
 
 - deploy-image: <https://github.com/kubernetes-sigs/kubebuilder/tree/v3.7.0/pkg/plugins/golang/deploy-image/v1alpha1>
 - grafana: <https://github.com/kubernetes-sigs/kubebuilder/tree/v3.7.0/pkg/plugins/optional/grafana/v1alpha>
+
```
camilamacedo86, line 67: WDYT about we add something like follow after the Overview section?

---
## When it is useful

- If you are looking for to create plugins which supports and work with another language
- If you would like to create helpers and integrations on top of the scaffolds done by the plugins provided by Kubebuilder.
- If you would like to have custmozed layouts according to your needs

## How the plugins can be used

Kubebuilder provides a set of plugins to scaffold the projects, to help you extend
and re-use its implementation as to provide additional features. For further information
see [Available Plugins](available-plugins.md).

Therefore, if you have a need you might want to propose a solution by adding a new plugin
which would be shipped with Kubebuilder by default. 

However, you might also want to have your own tool to address your specific scenarios and 
by taking advantage of what is provided by Kubebuilder as LIB alreadu for you are able to be
focus to address your needs and keep your solutions eaiser maintained. 

Note that you can use Kubebuilder as lib to import its plugins and then create your own
plugins that does customzations on top such as [Operator-SDK] does with the plugins [manifest](https://github.com/operator-framework/operator-sdk/tree/v1.23.0/internal/plugins/manifests/v2) and
[scorecard](https://github.com/operator-framework/operator-sdk/tree/v1.23.0/internal/plugins/scorecard/v2) for example to add its features. See [here](https://github.com/operator-framework/operator-sdk/blob/v1.23.0/internal/cmd/operator-sdk/cli/cli.go#L78-L160)

Another option implemented with the [Extensible CLI and Scaffolding Plugins - Phase 2][plugins-phase2-design-doc] is
to extend Kibebuilder as a LIB to cretae only a specific plugin that can be called and used with
Kubebuilder as well.

<aside class="note">
<H1> Plugins proposal docs</H1>

You can check the proposal documentation for better understand its motivations see the [Extensible CLI and Scaffolding Plugins: phase 1][plugins-phase1-design-doc],
the [Extensible CLI and Scaffolding Plugins: phase 1.5][plugins-phase1-design-doc-1.5] and the [Extensible CLI and Scaffolding Plugins - Phase 2][plugins-phase2-design-doc]
design docs. Also, you can check the [Plugins section][plugins-section].

</aside>

docs/book/src/plugins/creating-plugins.md
```diff
@@ -64,3 +64,46 @@ Feel free to check the implementation under:
...
+## Plugin scaffolding
```
camilamacedo86, line 68: WDYT about we add a section to explain how the plugins work in a chain?
I mean, that by creating a composition we would call the init first of plugin A, then plugin B and etc.

docs/book/src/plugins/creating-plugins.md
```diff
@@ -64,3 +64,46 @@ Feel free to check the implementation under:
...
+
+Your plugin may have the use case to add code on top of what is scaffolded by default with `create api`.
+This is common as you may expect your plugin to:
```
camilamacedo86, line 71: WDYT about we have here something like "Helpers in the API to do the scaffolds" and then we describe the options with something like:



### Bollerplates

You can check in the code example that the plugins will have boilerplates to generate
the files with code. Therefore, you can see that Kubebuider API provides some utils which
can help you to achieve your goals.

// TODO: Add an example of template

### You can overwrite a file

You might want for example to overwrite a scaffold done by using the option:

```
```

Example:

Let's imagine that you would like to have a helper plugin that would be called in a chain with go/v4-alpha to add customizations on top/ Therefore after we generate the code calling the subcommand to init from go/v4 alpha we would like to overwrite the Makefile to change this scaffold via our plugin. In this way, we would implement the Bollerplate for our Makefile and then use this option to ensure that it would be overwritten.

docs/book/src/plugins/creating-plugins.md
```diff
@@ -64,3 +64,46 @@ Feel free to check the implementation under:
...
+- Create API
+- Update controller manager logic
+- Generate corresponding manifests
+
+Following approaches may help:
+
+- If your plugin is expected to overwrite existing files with its own scaffold, you can set the `machinary.IfExistsAction` value to be `OverwriteFile`. Thus, when the plugin is called to scaffold, the target file will be overwritten if existing. See [example of deploy-image][example-of-deploy-image-1].
+- If your plugin is expected to modify part of the existing files with its scaffold, you may use functions provided by [sigs.k8s.io/kubebuilder/v3/pkg/plugin/util][kb-util]. See [example of deploy-image][example-of-deploy-image-2].
+
+#### Simplify the calling of multiple plugins
```
camilamacedo86, line 82: That is not for simplicity. Give a look at "Composite Pattern"
The idea is be able to re-use the components which provides pros like better maintainability.

camilamacedo86 (commented):
Hi @Kavinjsir,

I am ok with the changes made. Are you still working on this one OR is it all done for us to review and get it merged?

Kavinjsir:
> Details

Hi @camilamacedo86 , thx for reviewing!

I'm sorry to be late and do not make the update quite match to our basic idea.

If the current acceptable, I'm also good to get it merge for now and create a following PR as enhancement.

camilamacedo86 (approved):
I am ok with. :-) 

Let's see if we can get help from others as well. 
So, I will just approve.

varshaprasad96 (approved):
Nice docs! Thanks @Kavinjsir!
/approve
/lgtm

PR #2826, opened 2022-07-13
🐛  (kustomize/v1 and kustomize/v2-alpha) : ComponentConfig scaffolds should not be done by default

laxmikantbpandhare:
## A description of the change

This PR will verify whether the `--component-config` flag is passed or not. If not then it will not scaffold out the `controller_manager_config.yaml` file.

## The motivation for the change
it fixes #2782  

## Tests done to ensure the changes: 

Init command ( without the --component-config flag)

From this PR:  
```yaml
laxmikantbhaskarpandhare@lpandhar-mac test % kubebuilder init --domain my.domain --repo my.domain/guestbook

Writing kustomize manifests for you to edit...
false
Writing scaffold for you to edit...
Get controller runtime:
$ go get sigs.k8s.io/controller-runtime@v0.12.2
Update dependencies:
$ go mod tidy
Next: define a resource with:
$ kubebuilder create api
laxmikantbhaskarpandhare@lpandhar-mac test % tree
.
├── Dockerfile
├── Makefile
├── PROJECT
├── README.md
├── config
│   ├── default
│   │   ├── kustomization.yaml
│   │   ├── manager_auth_proxy_patch.yaml
│   │   └── manager_config_patch.yaml
│   ├── manager
│   │   ├── kustomization.yaml
│   │   └── manager.yaml
│   ├── prometheus
│   │   ├── kustomization.yaml
│   │   └── monitor.yaml
│   └── rbac
│       ├── auth_proxy_client_clusterrole.yaml
│       ├── auth_proxy_role.yaml
│       ├── auth_proxy_role_binding.yaml
│       ├── auth_proxy_service.yaml
│       ├── kustomization.yaml
│       ├── leader_election_role.yaml
│       ├── leader_election_role_binding.yaml
│       ├── role_binding.yaml
│       └── service_account.yaml
├── go.mod
├── go.sum
├── hack
│   └── boilerplate.go.txt
└── main.go

6 directories, 24 files
```

Init command ( with the changes and passed the --component-config flag)

```yaml
laxmikantbhaskarpandhare@lpandhar-mac testData % kubebuilder init --domain my.domain --repo my.domain/guestbook --component-config

Writing kustomize manifests for you to edit...
true
Writing scaffold for you to edit...
Get controller runtime:
$ go get sigs.k8s.io/controller-runtime@v0.12.2
Update dependencies:
$ go mod tidy
Next: define a resource with:
$ kubebuilder create api
[...]

PR #3013, opened 2022-10-10
🐛 : leader-elect arg scaffolded should be scaffolded when component config option is not used

laxmikantbpandhare:
There are 2 places where we must revert the sign which we missed as part of old PR https://github.com/kubernetes-sigs/kubebuilder/pull/2826

```
{{- if not .ComponentConfig }}
      - name: manager
        args:
        - "--health-probe-bind-address=:8081"
        - "--metrics-bind-address=127.0.0.1:8080"
        - "--leader-elect"
{{- end }}
```

it should be now modified to

```
{{- if .ComponentConfig }}
      - name: manager
        args:
        - "--health-probe-bind-address=:8081"
        - "--metrics-bind-address=127.0.0.1:8080"
        - "--leader-elect"
{{- end }}
```

As of now, we are scaffolding the controller_manager_config.yaml file only when the flag --component-config is passed.

### DIFF
--- a/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/kdefault/manager_auth_proxy_patch.go
+++ b/pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/kdefault/manager_auth_proxy_patch.go
@@ -93,7 +93,7 @@ spec:
           requests:
             cpu: 5m
             memory: 64Mi
-{{- if .ComponentConfig }}
+{{- if not .ComponentConfig }}
       - name: manager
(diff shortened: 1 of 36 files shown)


everettraven (approved):
/lgtm

varshaprasad96 (approved):
/lgtm

camilamacedo86:
/lgtm

@laxmikantbpandhare I hope that you do not mind, I just updated the title for we have a better changelog

jmrodri (approved):
/lgtm

PR #3022, opened 2022-10-17
✨ : added init and edit sub command plugin for component manager config file scaffolding

laxmikantbpandhare:
Add init and edit sub command plugin for component manager config file scaffolding.

Current output for the command:

```
laxmikantbhaskarpandhare@lpandhar-mac v1alpha1 % kubebuilder init --plugins config.kubebuilder.io/v1-alpha
Generating config/manager/controller_manager_config.yaml file...
laxmikantbhaskarpandhare@lpandhar-mac v1alpha1 % tree
.
├── PROJECT
└── config
    └── manager
        └── controller_manager_config.yaml

2 directories, 2 files
```

Also, edit sub command is working

```
laxmikantbhaskarpandhare@lpandhar-mac test-scaffold-init % tree                                             
.
└── PROJECT

0 directories, 1 file
laxmikantbhaskarpandhare@lpandhar-mac test-scaffold-init % kubebuilder edit --plugins config.kubebuilder.io/v1-alpha
Generating Grafana manifests to visualize controller status...
laxmikantbhaskarpandhare@lpandhar-mac test-scaffold-init % tree
.
├── PROJECT
└── config
    └── manager
        └── controller_manager_config.yaml

2 directories, 2 files
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -34,6 +34,7 @@ import (
 	deployimagev1alpha1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/deploy-image/v1alpha1"
 	golangv2 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v2"
 	golangv3 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/v3"
+	configv1alpha1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/optional/config/v1alpha"
 	grafanav1alpha1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/optional/grafana/v1alpha"
 )
 
@@ -71,6 +72,7 @@ func main() {
 			&declarativev1.Plugin{},
 			&deployimagev1alpha1.Plugin{},
 			&grafanav1alpha1.Plugin{},
+			&configv1alpha1.Plugin{},
 		),
 		cli.WithPlugins(externalPlugins...),
 		cli.WithDefaultPlugins(cfgv2.Version, golangv2.Plugin{}),
--- a/pkg/plugins/optional/config/v1alpha/constants.go
+++ b/pkg/plugins/optional/config/v1alpha/constants.go
@@ -1,0 +1,22 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha
+
+// nolint: lll
+const MetadataDescription = `This command will add Controller Manager Config file to the project:
+	('config/manager/controller_manager_config.yaml')
+`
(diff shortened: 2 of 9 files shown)


pkg/plugins/optional/config/v1alpha/scaffolds/internal/templates/config.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	f.TemplateBody = controllerManagerConfigTemplate
+
+	f.IfExistsAction = machinery.Error
+
+	return nil
+}
+
+const controllerManagerConfigTemplate = `apiVersion: controller-runtime.sigs.k8s.io/v1alpha1
```
camilamacedo86, line 48: We have more files changed when the component Config flag is used.
We need to ensure that all changes made when the flag is used will be performed when we use the plugin to init a project OR to edit a project that exist already.

pkg/plugins/optional/config/v1alpha/constants.go
```diff
@@ -0,0 +1,22 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha
+
+// nolint: lll
+const MetadataDescription = `This command will add Controller Manager Config file to the project:
+	('config/manager/controller_manager_config.yaml')
```
camilamacedo86, line 21: Why do we need this file?
Could we follow the same structure used/adopt in other plugins and keep the const where it will be used?

cmd/main.go
```diff
@@ -34,6 +34,7 @@ import (
 	deployimagev1alpha1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/golang/deploy-image/v1alpha1"
```
camilamacedo86, line 34: We need to ensure the changes using e2e tests.
Also, we need a sample under the testdata that will use this plugin such as we have for the other ones. 

Could you please try to address these ones?

camilamacedo86 (changes requested):
Thank you for your collaboration 🥇 
Following some needs for we are able to move with this PR

a) We need to ensure that when we use the plugin ALL files will be created and updated and we will have the same result of call the init with the flag
b) We need to ensure that we are testing the new plugin in the e2e tests
c) Also, we need to have a sample under the testdata to check out its scaffolds
d) We need to have the documentation (component Config tutorial) updated which will tell to users use the plugin instead of the flag
e) We also need to have the plugin defined under the section plugin


Could you please try to address these points?

laxmikantbpandhare:
> Thank you for your collaboration 🥇 Following some needs for we are able to move with this PR
> 
> a) We need to ensure that when we use the plugin ALL files will be created and updated and we will have the same result of call the init with the flag b) We need to ensure that we are testing the new plugin in the e2e tests c) Also, we need to have a sample under the testdata to check out its scaffolds d) We need to have the documentation (component Config tutorial) updated which will tell to users use the plugin instead of the flag e) We also need to have the plugin defined under the section plugin
> 
> Could you please try to address these points?

@camilamacedo86 - I will address all above queries.

laxmikantbpandhare:
> Thank you for your collaboration 🥇 Following some needs for we are able to move with this PR
> 
> a) We need to ensure that when we use the plugin ALL files will be created and updated and we will have the same result of call the init with the flag b) We need to ensure that we are testing the new plugin in the e2e tests c) Also, we need to have a sample under the testdata to check out its scaffolds d) We need to have the documentation (component Config tutorial) updated which will tell to users use the plugin instead of the flag e) We also need to have the plugin defined under the section plugin
> 
> Could you please try to address these points?

@camilamacedo86 - I agree with the above points. I will take a look into it.

pkg/plugins/optional/config/v1alpha/scaffolds/internal/templates/config.go
laxmikantbpandhare: @camilamacedo86 - Yes, right. In this PR, I just scaffolded one file. There are a couple of files which has the `--leader-elect` flag enabled according to the flag. I will update those too in the PR.

laxmikantbpandhare:
I started again on this PR.

PR #3023, opened 2022-10-17
✨ deprecate flag component-config

laxmikantbpandhare:
this will fix one of the requirements from https://github.com/kubernetes-sigs/kubebuilder/issues/2782

This fix will mark the flag component-config as deprecated and throw a warning to the user whenever a user tries to use it.

### DIFF
--- a/pkg/plugins/common/kustomize/v1/create.go
+++ b/pkg/plugins/common/kustomize/v1/create.go
@@ -35,7 +35,10 @@ type createSubcommand struct {
 	force bool
 }
 
-func (p *createSubcommand) BindFlags(fs *pflag.FlagSet) { p.flagSet = fs }
+func (p *createSubcommand) BindFlags(fs *pflag.FlagSet) {
+	p.flagSet = fs
+	_ = fs.MarkDeprecated("component-config", "component-config flag will be deprecated in upcoming releases")
+}
 
 func (p *createSubcommand) InjectConfig(c config.Config) error {
 	p.config = c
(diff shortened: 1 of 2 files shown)


laxmikantbpandhare:
This PR will be on hold as its replacement plugin is not merged yet https://github.com/kubernetes-sigs/kubebuilder/pull/3022

Please refer to this comment for more information: https://github.com/kubernetes-sigs/kubebuilder/pull/2942#issuecomment-1246342150

/hold

pkg/plugins/common/kustomize/v1/create.go
```diff
@@ -35,7 +35,10 @@ type createSubcommand struct {
 	force bool
 }
 
-func (p *createSubcommand) BindFlags(fs *pflag.FlagSet) { p.flagSet = fs }
+func (p *createSubcommand) BindFlags(fs *pflag.FlagSet) {
+	p.flagSet = fs
+	_ = fs.MarkDeprecated("component-config", "component-config flag will be deprecated in upcoming releases")
```
camilamacedo86, line 40: But we are deprecating when we use `MarkDeprecated`.
That will result in for example this option no longer be doc in sdk docs when we run make generate to generate the docs.

camilamacedo86: /hold

camilamacedo86:
@laxmikantbpandhare,

Could you please rebase this one with master?
We should move forward with this change since the ComponentConfig was deprecated, see:  https://github.com/kubernetes-sigs/kubebuilder/pull/3442

Also, could you please ensure that in the message we clarify the reason as described in the docs into the above PR?

PR #3034, opened 2022-10-22
🌱  golangci-lint: add recommended revive checks to linter-settings but d…

oscr:
* Updates golangci-lint version from 1.49.0 -> 1.50.1
* Adds [all recommended revive checks](https://github.com/mgechev/revive#recommended-configuration) to the config file. However to keep this pr simpler some new checks where disabled. But at least it's explicit now there are recommended checks that aren't enabled. I would gladly look into each of the disabled checks in follow up prs. 
* Fixes the finding:
   ```
   test/e2e/utils/test_context.go:273:1: receiver-naming: receiver name tc should be consistent with previous receiver name t for TestContext (revive)

   func (tc TestContext) LoadImageToKindClusterWithName(image string) error {
        cluster := "kind"
        if v, ok := os.LookupEnv("KIND_CLUSTER"); ok {
                cluster = v
        }
        kindOptions := []string{"load", "docker-image", "--name", cluster, image}
        cmd := exec.Command("kind", kindOptions...)
        _, err := tc.Run(cmd)
        return err
   }
   ```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,3 +1,7 @@
+run:
+  deadline: 5m
+  allow-parallel-runners: true
+
 issues:
   # don't skip warning about doc comments
   # don't exclude the default set of lint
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -13,8 +17,42 @@ linters-settings:
     enable=fieldalignment: true
   revive:
     rules:
-    - name: if-return
-      disabled: true
+      # The following rules are recommended https://github.com/mgechev/revive#recommended-configuration
```
camilamacedo86, line 20: I think we merged the revive by mistake. 
Why should we have revive and golangci?

Also, note that we are only using golangci and we do not test the projects with revive.
In this way, ihmo we need to remove:

```
linters-settings:
  govet:
    enable=fieldalignment: true
  revive:
    rules:
    - name: if-return
      disabled: true
```

oscr: To be honest I just noticed revive was enabled but not with the recommended configuration and therefor I made a pr to work towards the [recommended configuration](https://github.com/mgechev/revive#recommended-configuration). I believe that if you remove that section you will get [default configuration](https://github.com/mgechev/revive#default-configuration) instead.

As for the question of should there be / how much linting should a project have. I think that's a question you have more experience to answer. IMO there is a value for some linting. The question is always where do we get most value out of the least amount of linting overhead.

camilamacedo86: HI @oscr,

> I believe that if you remove that section you will get [default configuration](https://github.com/mgechev/revive#default-configuration) instead.

I would we get that?
We are not downloading or testing using revive. We just use golangci.
Did I missed something?

oscr: @camilamacedo86 golangci-lint is not a linter in itself, it's a linter runner. So you can configure golangci to run fewer or more or less linters. Then to make everything more complicated some linters (such as revive) are `metalinters` and they can be configured with even more checks. 

https://golangci-lint.run/usage/linters/

.golangci.yml
```diff
@@ -13,8 +17,42 @@ linters-settings:
...
+      - name: package-comments
+        disabled: true  # TODO: Investigate if it should be enabled. Disabled for now due to many findings.
+      - name: range
+      - name: receiver-naming
+      - name: time-naming
+      - name: unexported-return
+      - name: indent-error-flow
+      - name: errorf
+      - name: empty-block
+        disabled: true  # TODO: Investigate if it should be enabled. Disabled for now due to many findings.
```
camilamacedo86, line 45: If that is the recommendation we might want in a follow up fix the issues and then enable right?

camilamacedo86 (approved):
I am OK with the changes made.
Would be nice to fix the lint issues found in the other checks in follow-ups 🥇 

/ok-to-test

.golangci.yml
oscr: Yes, I think that would be good to do. You might want them. You might not. But I disabled the checks that found many / non trivial findings to keep the complexity and size of this pr smaller.

Issue #3002, opened 2022-10-05
Remove go/v2 plugin option


camilamacedo86:
### What do you want to happen?

Remove the implementation of go/v2 plugin and the tests and samples which still used by it.
We should also remove the legacy Project layout v2. 

Note that we plan to do that when we bump k8s 1.26 ONLY.

camilamacedo86:
Closing in favor of: https://github.com/kubernetes-sigs/kubebuilder/issues/3603

PR #3035, opened 2022-10-23
:warning: remove deprecated go/v2 which provides legacy layout - KB CLI 2x < 3.0.0

camilamacedo86:
## Description

Remove legacy layout kept to ensure backwards compatibility with KB CLI 2x.
Remove the old layout for the Project file used by it - Project config version 2.

## Motivation

We are unable to keep this plugin maintained and it does not work with the plugin ecosystem and/or the latest versions. More info: https://github.com/kubernetes-sigs/kubebuilder/discussions/3003

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/3002

/hold 
Until we are able to move forward. So, far it is planned for when we bump k8s 1.26


pkg/plugin/util/testdata/exampleFile.txt
```diff
@@ -1 +1 @@
-exampleTarget
\ No newline at end of file
+exampleTargetexampleCodeexampleCode
```
everettraven, line 1: I think we need to make sure this change doesn't get committed right? I *think* there is a test that fails if this gets committed as is

everettraven (approved):
Overall LGTM. Only one question (probably non-blocking):

camilamacedo86:
/hold 

Until we decide to move forward within

camilamacedo86:
We can close this one in favor to do that within a 4.x release.
Also, we can use this as example over what should be done.

Issue #3031, opened 2022-10-22
Update tutorial to no longer use  batch/v1beta1 API version of CronJob which is no longer be served in v1.25


camilamacedo86:
### What do you want to happen?

Update tutorial to no longer use  batch/v1beta1 API version of CronJob which is no longer be served in v1.25

### Extra Labels

/kind cleanup

krishnaindani:
hey, whuich version does this need to be updated with? Could I work on this?

camilamacedo86:
Hi @krishnaindani, 

That is great. See https://kubernetes.io/docs/reference/using-api/deprecation-guide/#cronjob-v125 we need replace batch/v1beta1 with batch/v1 in the tutorial samples: https://github.com/kubernetes-sigs/kubebuilder/search?q=batch%2Fv1beta1

That means, replace:

```
batchv1beta1 "k8s.io/api/batch/v1beta1"
```

With:


 ```
batchv1 "k8s.io/api/batch/v1"
```

krishnaindani:
Thanks, Could you assign this to me? Okay, I will start working on this.

krishnaindani:
Have created a pr here, could you please review this and let me know if any changes needs to be done?

PR #3038, opened 2022-10-24
📖 update cronjob to use batchv1 api instead of v1beta1

krishnaindani:
Issue: https://github.com/kubernetes-sigs/kubebuilder/issues/3031

Update tutorial to use the batchv1 API version for cronjob.


krishnaindani:
Have signed the cla

PR #3039, opened 2022-10-24
📖 Fix grammar in comment

seemiller:
Change `these is` to `these are` in a comment in cronjob_controller.go.

The verb `is` is singular, whereas `are` is plural. The comment is there to document iterating over multiple failed jobs. So, it would be appropriate to use a plural verb.

Here's to comment for context:
>	// NB: deleting these is "best effort" -- if we fail on a particular one,
>	// we won't requeue just to finish the deleting.


oscr:
@seemiller Thank you for your pr! Unfortunately you need to make your pr description a little longer in order for it to pass the PR desc check. 

/ok-to-test

camilamacedo86 (approved):
/lgtm

/ok-to-test

PR #3040, opened 2022-10-24
🌱 Enable revive unused-parameter check and fix findings

oscr:
Enables the revive unused-parameter check and fixes all the findings. This is one of the recommended checks to have enabled but wasn't enabled in #3034 due to many findings.


camilamacedo86 (approved):
/lgtm

PR #3043, opened 2022-10-25
🌱 Update github workflow to use same golangci-lint minor version as Makefile

oscr:
This pr ensures the GitHub workflow uses the same minor version of golangci-lint as installed by the Makefile as can be seen here: https://github.com/kubernetes-sigs/kubebuilder/blob/523907a7e70a7b3d02c2df91382f6b1ee2ca4c54/Makefile#L86


camilamacedo86 (approved):
/lgtm

Issue #3028, opened 2022-10-18
Default v3 scaffold manager pod got permission denied error


astraw99:
### What broke? What's expected?

Based on the quick start doc, did the `make manifests, make install`, then
`make docker-build docker-push IMG=xxx/xxx:tag, make deploy IMG=xxx/xxx:tag`, then
in the cluster the controller-pod got error:
```
1.6656687258729894e+09  ERROR   controller-runtime.client.config        unable to get kubeconfig        {"error": "open /var/run/secrets/kubernetes.io/serviceaccount/token: permission denied"}
sigs.k8s.io/controller-runtime/pkg/client/config.GetConfigOrDie
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/client/config/config.go:153
main.main
        /workspace/main.go:68
runtime.main
        /usr/local/go/src/runtime/proc.go:250
```

### Reproducing this issue

Based on the quick start doc, did the `make manifests, make install`, then
`make docker-build docker-push IMG=xxx/xxx:tag, make deploy IMG=xxx/xxx:tag`, then
in the cluster will reproduce this issue.

My cluster K8s version is:
```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.0", GitCommit:"cb303e613a121a29364f75cc67d3d580833a7479", GitTreeState:"clean", BuildDate:"2021-04-08T16:31:21Z", GoVersion:"go1.16.1", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.0", GitCommit:"cb303e613a121a29364f75cc67d3d580833a7479", GitTreeState:"clean", BuildDate:"2021-04-08T16:25:06Z", GoVersion:"go1.16.1", Compiler:"gc", Platform:"linux/amd64"}
```
Kubebuilder version is:
```
Version: main.version{KubeBuilderVersion:"3.7.0", KubernetesVendor:"1.24.1", GitCommit:"3bfc84ec8767fa760d1771ce7a0cb05a9a8f6286", BuildDate:"2022-09-20T17:21:57Z", GoOs:"darwin", GoArch:"amd64"}
```
With these context, will replicate the error.

### KubeBuilder (CLI) Version

[...]

Issue #3016, opened 2022-10-13
Got errors on kubebuilder v3 quick start


astraw99:
### What broke? What's expected?

Using the latest kubebuilder cli, quick start a new project based on the [official document](https://book.kubebuilder.io/quick-start.html), got these errors:
1. Enable the leader-election, then `make run`, got error:
```
go fmt ./...
go vet ./...
go run ./main.go
1.665666448465308e+09   ERROR   setup   unable to start manager {"error": "unable to find leader election namespace: not running in-cluster, please specify LeaderElectionNamespace"}
main.main
...
exit status 1
make: *** [run] Error 1
```

2. After `make deploy IMG=xxx/xxx:tag`, the controller-pod got error:
```                                    
1.6656687258729894e+09  ERROR   controller-runtime.client.config        unable to get kubeconfig        {"error": "open /var/run/secrets/kubernetes.io/serviceaccount/token: permission denied"}
sigs.k8s.io/controller-runtime/pkg/client/config.GetConfigOrDie
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/client/config/config.go:153
main.main
        /workspace/main.go:68
runtime.main
        /usr/local/go/src/runtime/proc.go:250
```

3. As the doc said `Important: Run "make" to regenerate code after modifying this file`, but `make` did not generate latest CRD fields.

### Reproducing this issue

Just quick start a new project based on the [official document](https://book.kubebuilder.io/quick-start.html), then will reproduce this issue.

### KubeBuilder (CLI) Version

Version: main.version{KubeBuilderVersion:"3.7.0", KubernetesVendor:"1.24.1", GitCommit:"3bfc84ec8767fa760d1771ce7a0cb05a9a8f6286", BuildDate:"2022-09-20T17:21:57Z", GoOs:"darwin", GoArch:"amd64"}

laxmikantbpandhare:
`make run` is working for me. But I haven't updated any code for the leader election.

```
laxmikantbhaskarpandhare@lpandhar-mac kb % make run
test -s /Users/laxmikantbhaskarpandhare/comparison-memcached-operator/remove/kb/bin/controller-gen || GOBIN=/Users/laxmikantbhaskarpandhare/comparison-memcached-operator/remove/kb/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.10.0
[...]

PR #3033, opened 2022-10-22
:sparkles: Use `klog` as kubebuilder default logger

astraw99:
Fixes #3036 
Fixes https://github.com/kubernetes-sigs/controller-runtime/issues/2024

To make the log output consistent with K8s core components (kube-apiserver, kube-controller-manager, etc), and
more human readable.

Test results:
Before this PR, the kubebuilder generated controller-manager log is:
```
1.6656692709539523e+09  INFO    controller-runtime.metrics      Metrics server is starting to listen    {"addr": "127.0.0.1:8080"}
1.665669270954326e+09   INFO    setup   starting manager
1.6656692709546785e+09  INFO    Starting server {"path": "/metrics", "kind": "metrics", "addr": "127.0.0.1:8080"}
1.6656692709546802e+09  INFO    Starting server {"kind": "health probe", "addr": "[::]:8081"}
```

After this PR, the kubebuilder generated controller-manager log is:
```
I1016 23:41:42.700531   61722 listener.go:44] "controller-runtime/metrics: Metrics server is starting to listen" addr=":8080"
I1016 23:41:42.701073   61722 main.go:115] "setup: starting manager"
I1016 23:41:42.701251   61722 internal.go:362] "Starting server" kind="health probe" addr="[::]:8081"
I1016 23:41:42.701255   61722 internal.go:362] "Starting server" path="/metrics" kind="metrics" addr="[::]:8080"
```

PR #3044, opened 2022-10-25
:book: Add kubebuilder FAQ section

### LINKED ISSUE #1318: Explain more about the domain field is in documentation

astraw99:
Fixes #1318 #1723 #3016 #3028

Add a FAQ section in the KB docs, and include some of the representational reported issues.
Will update it continuously, to help more users.

Also add `git-repo` and `git-edit` icons on the top-right page.
Tested with `mdbook serve` locally, works fine.

### DIFF
--- a/docs/book/src/faq.md
+++ b/docs/book/src/faq.md
@@ -1,0 +1,100 @@
+
+# FAQ
+
+## How does the value informed via the domain flag (i.e. `kubebuilder init --domain example.com`) when we init a project?
+
+After creating a project, usually you will want to extend the Kubernetes APIs and define new APIs which will be owned by your project. Therefore, the domain value is tracked in the [PROJECT][project-file-def] file which defines the config of your project and will be used as a domain to create the endpoints of your API(s). Please, ensure that you understand the [Groups and Versions and Kinds, oh my!][gvk].
+
+The domain is for the group suffix, to explicitly show the resource group category.
+For example, if set `--domain=example.com`:
+```
+kubebuilder init --domain example.com --repo xxx --plugins=go/v4-alpha
+kubebuilder create api --group mygroup --version v1beta1 --kind Mykind
+```
+Then the result resource group will be `mygroup.example.com`.
+
+> If domain field not set, the default value is `my.domain`.
+
+## I'd like to customize my project to use [klog][klog] instead of the [zap][zap] provided by controller-runtime. How to use `klog` or other loggers as the project logger?
+
+In the `main.go` you can replace:
+```go
+    opts := zap.Options{
+    Development: true,
+    }
+    opts.BindFlags(flag.CommandLine)
+    flag.Parse()
+    
+    ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+```
+with:
+```go
+    flag.Parse()
+	ctrl.SetLogger(klog.NewKlogr())
+```
+
+## After `make run`, I see errors like "unable to find leader election namespace: not running in-cluster..."
+
+You can enable the leader election. However, if you are testing the project locally using the `make run`
+target which will run the manager outside of the cluster then, you might also need to set the
+namespace the leader election resource will be created, as follows:
+```go
+mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme:                  scheme,
+		MetricsBindAddress:      metricsAddr,
+		Port:                    9443,
+		HealthProbeBindAddress:  probeAddr,
+		LeaderElection:          enableLeaderElection,
+		LeaderElectionID:        "14be1926.testproject.org",
+		LeaderElectionNamespace: "<project-name>-system",
+```
+
+If you are running the project on the cluster with `make deploy` target
+then, you might not want to add this option. So, you might want to customize this behaviour using
+environment variables to only add this option for development purposes, such as:
+
+```go
+    leaderElectionNS := ""
+	if os.Getenv("ENABLE_LEADER_ELECATION_NAMESPACE") != "false" {
+		leaderElectionNS = "<project-name>-system"
+	}
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme:                  scheme,
+		MetricsBindAddress:      metricsAddr,
+		Port:                    9443,
+		HealthProbeBindAddress:  probeAddr,
+		LeaderElection:          enableLeaderElection,
+		LeaderElectionNamespace: leaderElectionNS,
+		LeaderElectionID:        "14be1926.testproject.org",
+		...
+```
+
+## I am facing the error "open /var/run/secrets/kubernetes.io/serviceaccount/token: permission denied" when I deploy my project against Kubernetes old versions. How to sort it out?
+
+If you are facing the error:
+```
+1.6656687258729894e+09  ERROR   controller-runtime.client.config        unable to get kubeconfig        {"error": "open /var/run/secrets/kubernetes.io/serviceaccount/token: permission denied"}
+sigs.k8s.io/controller-runtime/pkg/client/config.GetConfigOrDie
+        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/client/config/config.go:153
+main.main
+        /workspace/main.go:68
+runtime.main
+        /usr/local/go/src/runtime/proc.go:250
+```
+when you are running the project against a Kubernetes old version (maybe <= 1.21) , it might be caused by the [issue][permission-issue] , the reason is the mounted token file set to `0600`, see [solution][permission-PR] here. Then, the workaround is:
+
+Add `fsGroup` in the manager.yaml
+```yaml
+securityContext:
+        runAsNonRoot: true
+        fsGroup: 65532 # add this fsGroup to make the token file readable
+```
+However, note that this problem is fixed and will not occur if you deploy the project in high versions (maybe >= 1.22).
+
+[gvk]: ./cronjob-tutorial/gvks.md
+[project-file-def]: ./reference/project-config.md
+[klog]: https://github.com/kubernetes/klog
+[zap]: https://github.com/uber-go/zap
+[permission-issue]: https://github.com/kubernetes/kubernetes/issues/82573
+[permission-PR]: https://github.com/kubernetes/kubernetes/pull/89193
--- a/docs/book/src/SUMMARY.md
+++ b/docs/book/src/SUMMARY.md
@@ -6,6 +6,8 @@
 
 [Architecture](./architecture.md)
 
+[FAQ](./faq.md)
+
 ---
 
 - [Tutorial: Building CronJob](cronjob-tutorial/cronjob-tutorial.md)


.gitignore
```diff
@@ -3,6 +3,7 @@
 WORKSPACE
 # don't check in the build output of the book
 docs/book/book/
+docs/book/src/docs
```
camilamacedo86, line 6: Why are we adding it to the gitignore? Why do we need to change the .gitignore to add a new doc section?

docs/book/book.toml
```diff
@@ -8,6 +8,8 @@ title = "The Kubebuilder Book"
 google-analytics = "UA-119864590-1"
 curly-quotes = true
 additional-css = ["theme/css/markers.css", "theme/css/custom.css"]
+git-repository-url = "https://github.com/kubernetes-sigs/kubebuilder"
+edit-url-template = "https://github.com/kubernetes-sigs/kubebuilder/edit/master/docs/book/{path}"
```
camilamacedo86, line 12: Why do we need to change the .book.tool to add a new doc section?
Please, check that we just add the file and we create it into the Summary.

Please, look at how the other docs/sections are done.

docs/book/src/faq.md
```diff
@@ -0,0 +1,88 @@
+
+# Kubebuilder FAQ
```
camilamacedo86, line 2: WDYT is about to be the only FAQ? Is it not a little prolixly add Kubebuilde here?

docs/book/src/faq.md
```diff
@@ -0,0 +1,88 @@
...
+
+## How to undestand the domain field?
```
camilamacedo86, line 4:
```suggestion
## How to the value informed via the domain flag when we init a project is used?
```

docs/book/book.toml
astraw99: To add git-repo and git-edit icons on the top-right page, to let other users to go to KB project quickly, and go to the edit page easily if want to modify the doc.

docs/book/src/faq.md
```diff
@@ -0,0 +1,88 @@
...
+First of all, we should get familiar with the GVK terminology, check [here](./cronjob-tutorial/gvks.md).
```
camilamacedo86, line 5: WDYT about something like:  

```suggestion
After creating a project, usually you will want to extend the Kubernetes APIs and define new APIs which will be owned by your project. Therefore, the domain value is tracked in the [PROJECT][project-file-def] file which defines the config of your project and will be used as a domain to create the endpoints of your API(s). Please, ensure that you understand the [Groups and Versions and Kinds, oh my!][gvk].
```

Note that for maintainability we keep the links at the bottom and we use an alias. You can check this standard applied in the other docs. 

So, let's add to the bottom:

``` 
[gkv]: ./cronjob-tutorial/gvks.md
[project-file-def]: https://book.kubebuilder.io/reference/project-config.html
```

docs/book/src/faq.md
```diff
@@ -0,0 +1,88 @@
...
+For example, if set `--domain=example.com`:
+```
+kubebuilder init --domain example.com --repo xxx --plugins=go/v4-alpha
+kubebuilder create api --group mygroup --version v1beta1 --kind Mykind
+```
+Then the result resource group will be `mygroup.example.com`.
+
+If domain field not set, the default value is `my.domain`.
+
+## How to validate API or set default value?
```
camilamacedo86, line 17: That is not the only way to do that.
We can use the OpenAPI schema 
And to set default values we can provide CR as an example for the users as we can also set them via the controller's reconciliation. (if/when they are not provided). Have we had any issues with this one? If so, what is the issue? Could we discuss it separated so that we can move with this PR with the other things? COuld we have a follow up PR to add this one if you see that it still fit?

docs/book/src/faq.md
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (r *Mykind) ValidateDelete() error {
+	// TODO(user): fill in your validation logic upon object deletion.
+	return nil
+}
+```
+So it is easy to add API default and validate logic.
+
+## How to use `klog` or other loggers as the project logger?
```
camilamacedo86, line 47:
```suggestion
## I'd like to customize my project to use [klog][klog-lib] instead of the [zap][zap-lib] lib provided by controller-runtime. How to use `klog` or other loggers as the project logger?`

File: docs/book/src/faq.md [hunk 8]
```diff
@@ -0,0 +1,88 @@
...
+We can set the project logger by:
```
camilamacedo86, line 48: See that we follow the same docs guidelines applied for k8s docs. Therefore, the proteome here seems that should be you and not we. Also, we need to describe where the change needs to be done. So, wdyt:

```suggestion
Into the `main.go` you can replace:

<add code>

with:

<add code example>
```

camilamacedo86: Also, I think we need here explain the options.

docs/book/src/faq.md
```diff
@@ -0,0 +1,88 @@
...
+```go
+	//ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+	ctrl.SetLogger(klog.NewKlogr())
+```
+
+## Testing leader-election outside of the cluster?
+If you are testing the project locally using the `make run` target which will run the manager outside of the cluster, then you might also need to set the namespace, as follows:
```
camilamacedo86, line 55: We need to give a space between the question and the answer.

camilamacedo86, line 55:
```suggestion

You can enable the leader election. However, if you are testing the project locally using the `make run` 
target which will run the manager outside of the cluster then, you might also need to set the 
namespace the leader election resource will be created, as follows:
```

docs/book/src/faq.md
```diff
@@ -0,0 +1,88 @@
...
+```go
+mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme:                 scheme,
+		MetricsBindAddress:     metricsAddr,
+		Port:                   9443,
+		HealthProbeBindAddress: probeAddr,
+		LeaderElection:         enableLeaderElection,
+		LeaderElectionID:       "14be1926.testproject.org",
+		LeaderElectionNamespace: "your-ns", // set this field to `make run` locally
```
camilamacedo86, line 64:
```suggestion
		LeaderElectionNamespace: "<project-name>-system",
```

docs/book/src/faq.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
```
camilamacedo86, line 66: We need also to make clear that is not might be required when we are running on the cluster

``` 
If you are running the project on the cluster with `make deploy` target
then, you might not want to add this option. So, you might want to customize this behaviour using
environment variables to only add this option for development purposes, such as:

TODO, add example with:

      > if os.Getenv("ENABLE_LEADER_ELECATION_NAMESPACE") != "false" {
      >}

```

docs/book/src/faq.md
```diff
@@ -0,0 +1,88 @@
...
+So it is easy to add API default and validate logic.
+
+## How to use `klog` or other loggers as the project logger?
+We can set the project logger by:
+```go
+	//ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+	ctrl.SetLogger(klog.NewKlogr())
+```
+
+## Testing leader-election outside of the cluster?
```
camilamacedo86, line 54: We need to add here the error faced . See how we add the titles in the FAQ of SDK docs to have a better example: https://sdk.operatorframework.io/docs/faqs/#after-deploying-my-operator-i-see-errors-like-failed-to-watch-external-type

docs/book/src/faq.md
```diff
@@ -0,0 +1,88 @@
...
+1.6656687258729894e+09  ERROR   controller-runtime.client.config        unable to get kubeconfig        {"error": "open /var/run/secrets/kubernetes.io/serviceaccount/token: permission denied"}
+sigs.k8s.io/controller-runtime/pkg/client/config.GetConfigOrDie
+        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/client/config/config.go:153
+main.main
+        /workspace/main.go:68
+runtime.main
+        /usr/local/go/src/runtime/proc.go:250
+```
+
+Reason: see [SA token permission related issues and PRs](https://github.com/kubernetes-sigs/kubebuilder/issues/3028#issuecomment-1288114462)
```
camilamacedo86, line 81: It does not occur from k8s version X, right? (ihmo)We need to better clarify it here, i.e.: 

If you are facing the error `{"error": "open /var/run/secrets/kubernetes.io/serviceaccount/token: permission denied"}` when you are running the project against a Kubernetes cluster < X it might be caused by the issue (add links) so, the reason is ..., see ... here. Then, the workaround is: 


However, not that this problem is fixed and will not occurs if you deploy the project in versions >= X.

docs/book/src/faq.md
```diff
@@ -0,0 +1,88 @@
...
+		Scheme:                 scheme,
+		MetricsBindAddress:     metricsAddr,
+		Port:                   9443,
+		HealthProbeBindAddress: probeAddr,
+		LeaderElection:         enableLeaderElection,
+		LeaderElectionID:       "14be1926.testproject.org",
+		LeaderElectionNamespace: "your-ns", // set this field to `make run` locally
+```
+
+## Got controller-pod SA permission denied error?
```
camilamacedo86, line 67: (ihmo) we need something here like

I am facing the error `{"error": "open /var/run/secrets/kubernetes.io/serviceaccount/token: permission denied"}` when I deploy my project against Kubernetes versions => X. How to sort it out?

docs/book/theme/index.hbs
```diff
@@ -139,6 +139,11 @@
                             <i id="git-repository-button" class="fa {{git_repository_icon}}"></i>
                         </a>
                         {{/if}}
+                        {{#if git_repository_edit_url}}
+                            <a href="{{git_repository_edit_url}}" title="Suggest an edit" aria-label="Suggest an edit">
+                                <i id="git-edit-button" class="fa fa-edit"></i>
```
camilamacedo86, line 144: Is this change part of the scope of this PR?
Could we create a PR for add this option? 

Ideally, we should have 1 pr to address each need so that we can revert if we need it, it has less back and forward and make easier the review process.

docs/book/src/faq.md
camilamacedo86: Also, this one we might address in the scaffolds via https://github.com/kubernetes-sigs/kubebuilder/pull/3033 so we would not add here.

astraw99:
@camilamacedo86 Updated, PTAL thanks.

docs/book/src/faq.md
camilamacedo86: Hi @astraw99,

That is not the only way to do so. I think if we would like to cover this topic we need to give further info and explain how to achieve it with all option. Also, we have a webhooks section in the tutorial guides and etc. We can add the info about how to use "--programmatic-validation" option to validate the API there as an example. 

The FAQ section would be for things like "I am facing an issue X how to solve" (when making no sense we change the code OR it is no longer faced in new versions).

camilamacedo86 (commented):
Hi @astraw99,

Great work 🥇 

The only one that I do not go along with is `## How to validate API or set default value?`. So, if you agree to remove this one it has my LGTM at least.

docs/book/src/faq.md
```diff
@@ -0,0 +1,131 @@
...
+		MetricsBindAddress:      metricsAddr,
+		Port:                    9443,
+		HealthProbeBindAddress:  probeAddr,
+		LeaderElection:          enableLeaderElection,
+		LeaderElectionNamespace: leaderElectionNS,
+		LeaderElectionID:        "14be1926.testproject.org",
+		...
+```
+
+## I am facing the error "open /var/run/secrets/kubernetes.io/serviceaccount/token: permission denied" when I deploy my project against Kubernetes versions => X. How to sort it out?
```
camilamacedo86, line 104: Is not this error ONLY faced with K8s versions < 1.21? 

So that we need to replace the ` => X.`  with <= 1.21.

docs/book/src/faq.md
```diff
@@ -0,0 +1,131 @@
...
+```
+1.6656687258729894e+09  ERROR   controller-runtime.client.config        unable to get kubeconfig        {"error": "open /var/run/secrets/kubernetes.io/serviceaccount/token: permission denied"}
+sigs.k8s.io/controller-runtime/pkg/client/config.GetConfigOrDie
+        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/client/config/config.go:153
+main.main
+        /workspace/main.go:68
+runtime.main
+        /usr/local/go/src/runtime/proc.go:250
+```
+when you are running the project against a Kubernetes cluster < X, it might be caused by the [issue][permission-issue] , the reason is the mounted token file set to `0600`, see [solution][permission-PR] here. Then, the workaround is:
```
camilamacedo86, line 116: @astraw99 here the same.

astraw99:
@camilamacedo86 Updated, PTAL thanks.

docs/book/src/faq.md
```diff
@@ -0,0 +1,100 @@
...
+		Port:                    9443,
+		HealthProbeBindAddress:  probeAddr,
+		LeaderElection:          enableLeaderElection,
+		LeaderElectionNamespace: leaderElectionNS,
+		LeaderElectionID:        "14be1926.testproject.org",
+		...
+```
+
+## I am facing the error "open /var/run/secrets/kubernetes.io/serviceaccount/token: permission denied" when I deploy my project against Kubernetes versions <= 1.21. How to sort it out?
+
```
camilamacedo86, line 74: See : https://github.com/kubernetes-sigs/kubebuilder/issues/3028#issuecomment-1293393346

- We cannot add the version if we are not 100% sure of what version it was solved (instead we can use old versions)
- We cannot add the fix PR if we are not 100% sure that it was the right one.

astraw99: Updated, PTAL thanks.

camilamacedo86 (commented):
This one is only to add a new FAQ section on the docs with some questions/answers.
It seems fine to be get merged without a second review.

Anyway we can always do follow up PRs to improve this one.

/lgtm

astraw99:
@camilamacedo86 With this PR merged, but it did not show in the website https://book.kubebuilder.io/introduction.html
PTAL thanks.

camilamacedo86:
Hi @astraw99,

The docs changes are made in the master branch and will only be in prod when we do the next release. You can check then out in the master branch. For further info see the contribution guide here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#how-to-contribute-to-docs

Issue #2286, opened 2021-08-06
Custom resource metrics


robbie-demuth:
### What do you want to happen?

Prior to the Operator Framework's Operator SDK re-architecting itself as a Kubebuilder plugin, it contained libraries (since deleted in operator-framework/operator-sdk#3249) for generating and serving custom resource metrics (implemented using libraries from kube-state-metrics) and scaffolded operators to invoke said libraries on startup. As a result, operators built using the Operator SDK not only exposed the standard controller-runtime metrics, but also simple metrics about the existence of custom resources managed by the controller (i.e. `<KIND>_info{<KIND>="<NAME>", namespace="<NAMESPACE>"} 1`).

It'd be nice to provide these metrics for free via Kubebuilder's scaffolding or at least provide an easy way for operator developers to opt-in.

If possible, it'd also be nice to provide an easy, pluggable way for operator developers to declaratively instrument their operators with custom (adjective) custom resource metrics without having to do so in `Reconcile`. For example, if a custom resource has a simple string field in its status subresource, it'd be nice to provide an API that allows operator developers to provide a metric name suffix, label name, and function for deriving label values from custom resources so that the controller generates and serves these metrics separately from `Reconcile` - preventing developers from having to instrument `Reconcile` wherever they update the status subresource's string field. 

I wasn't sure whether to open this issue against kubebuilder or controller-runtime. I'm more than happy to move it to controller-runtime if desired. I also considered opening this issue against https://github.com/operator-framework/operator-lib since it's maintained as part of the Operator Framework.

Thanks!

robbie-demuth:
RE my kube-state-metrics comment at the end, it looks like there have been some feature requests over there. The older of these two was closed because the maintainers felt that such metrics were the responsibility of operators. The other is still open and it looks like implementation is underway

https://github.com/kubernetes/kube-state-metrics/issues/457
https://github.com/kubernetes/kube-state-metrics/issues/1210

rashmigottipati:
@robbie-demuth Sounds interesting! Could you please describe some use cases on how you would use these standard controller runtime and metrics about existence of CRs?

robbie-demuth:
The simple "info" metric without additional labels could be used to count custom resources by namespace, cluster, etc. Making additional labels or metrics pluggable, however, really opens up the door to anything

camilamacedo86:
Hi @robbie-demuth, 

Thank you to raise this issue.

[...]

PR #3076, opened 2022-11-16
✨ (WIP) feat: initialize kube-state-metrics plugin

Kavinjsir:
### 2022-11-17 Updates
Hold, we may consider [the monitoring plugin](https://github.com/kubernetes-sigs/kubebuilder/issues/3078) for integrating this feature, OR [phase 2 approach](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-2.md) as making this an external plugin.

# Motivation
This is a starting PR to resolve https://github.com/kubernetes-sigs/kubebuilder/issues/2286

# Description
Introduce a new optional plugin `kube-state-metrics`.
This metrics aims to provide configuration file loaded by KSM to populate state metrics of custom resources.
The state metrics should be generated based on user-defined subresource `status`.

This PR initializes the plugin to provide the very basic `_info` metric of the CRs.
A common usage:
_"count custom resources by namespace, cluster, etc"_

## Usage
The plugin is temporary designed to be triggered by `edit` subcommand:
```shell
# Execute the command after APIs are generated
kubebuilder edit --plugins=kube-state-metrics.kubebuilder.io/v1-alpha
```
Once executed, the plugin will look for the apis recorded in `PROJECT`. Then, it will generate a config yaml at:
`kube-state-metrics/cr-config.yaml`

The file can be loaded by `kube-state-metrics` through `--custom-resource-state-config-file /path/to/kube-state-metrics/cr-config.yaml`.
The content refers to https://github.com/kubernetes/kube-state-metrics/blob/v2.6.0/docs/customresourcestate-metrics.md

## TODO:
- [x] initialize the plugin
- [ ] documentation
- [x] testdata
- [ ] e2e tests

## Example
Below is a practice over[ memcached-operator](https://github.com/operator-framework/operator-sdk/tree/v1.25.2/testdata/go/v3/memcached-operator):

case-1

case-2

### DIFF
--- a/pkg/plugins/optional/kube-state-metrics/v1alpha/scaffolds/internal/templates/config.go
+++ b/pkg/plugins/optional/kube-state-metrics/v1alpha/scaffolds/internal/templates/config.go
@@ -1,0 +1,93 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package templates
+
+import (
+	"bytes"
+	"fmt"
+	"path/filepath"
+	"text/template"
+
+	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
+	"sigs.k8s.io/kubebuilder/v3/pkg/model/resource"
+)
+
+var _ machinery.Template = &ConfigManifest{}
+
+// Kustomization scaffolds a file that defines the kustomization scheme for the prometheus folder
+type ConfigManifest struct {
+	machinery.TemplateMixin
+
+	// Custom Resource to be instrumented
+	GVKs []resource.GVK
+}
+
+// SetTemplateDefaults implements file.Template
+func (f *ConfigManifest) SetTemplateDefaults() error {
+	if f.Path == "" {
+		f.Path = filepath.Join("kube-state-metrics", "cr-config.yaml")
+	}
+
+	defaultTemplate, err := f.createTemplate()
+	if err != nil {
+		return err
+	}
+
+	f.TemplateBody = defaultTemplate
+
+	f.IfExistsAction = machinery.OverwriteFile
+
+	return nil
+}
+
+func (f *ConfigManifest) createTemplate() (string, error) {
+	t := template.Must(template.New("customResourcesConfig").Parse(customResourcesConfigTemplate))
+
+	outputTmpl := &bytes.Buffer{}
+	if err := t.Execute(outputTmpl, f.GVKs); err != nil {
+		return "", fmt.Errorf("error when generating kube-state-metrics config: %w", err)
+	}
+
+	return outputTmpl.String(), nil
+}
+
+// nolint: lll
+const customResourcesConfigTemplate = `---
+kind: CustomResourceStateMetrics
+spec:
+  resources:{{ range $i, $e := . }}
+    - groupVersionKind:
+        group: "{{.Group}}.{{.Domain}}"
+        kind: "{{.Kind}}"
+        version: "{{.Version}}"
+      labelsFromPath:
+        name:
+          - metadata
+          - name
+        namespace:
+          - metadata
+          - namespace
+        uid:
+          - metadata
+          - uid
+      metrics:
+        - name: "info"
+          each:
+            type: Info
+            info:
+              labelsFromPath:
+                version:
+                  - spec
+                  - version{{ end }}
+`
--- a/test/testdata/generate.sh
+++ b/test/testdata/generate.sh
@@ -118,6 +118,11 @@ function scaffold_test_project {
       $kb edit --plugins=grafana.kubebuilder.io/v1-alpha
   fi
 
+  if [[ $project == "project-v3" || $project == "project-v4" || $project =~ multigroup ]]; then
+     header_text 'Editing project with kube-state-metrics plugin ...'
+     $kb edit --plugins=kube-state-metrics.kubebuilder.io/v1-alpha
+  fi
+
   make generate manifests
   rm -f go.sum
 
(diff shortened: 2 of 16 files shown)


pkg/plugins/optional/kube-state-metrics/v1alpha/scaffolds/internal/templates/config.go
```diff
@@ -0,0 +1,93 @@
...
+	return outputTmpl.String(), nil
+}
+
+// nolint: lll
+const customResourcesConfigTemplate = `---
+kind: CustomResourceStateMetrics
+spec:
+  resources:{{ range $i, $e := . }}
+    - groupVersionKind:
+        group: "{{.Group}}.{{.Domain}}"
```
varshaprasad96, line 71: Do we have to scaffold config file for all GVKs in the project, or can we allow users to select the GVK they would like to?

pkg/plugins/optional/kube-state-metrics/v1alpha/scaffolds/internal/templates/config.go
```diff
@@ -0,0 +1,93 @@
...
+        namespace:
+          - metadata
+          - namespace
+        uid:
+          - metadata
+          - uid
+      metrics:
+        - name: "info"
+          each:
+            type: Info
```
varshaprasad96, line 87: Do we allow this to be user-configurable? I'm fine with the first phase being templated scaffolding, but we should probably think more about how to make this customizable in case we want to?

varshaprasad96:
@Kavinjsir adding some thoughts for the questions above:

> Whether it is necessary to install kube-state-metrics
There are a bunch of opinions for this, and we may have to look into various factors before coming to a decision. Should the KSM necessarily run on cluster. For example, in case of Prometheus we can allow users to run an off-cluster instance and just scrape the metrics through a service-monitor. Is a configuration similar to that possible. If so, we could have a plugin generate kustomize files, that scaffold out the necessary manifests (KSM deployment, service role-binding etc) like we do for Prometheus.

Does KSM expose some generic metrics (https://github.com/kubernetes/kube-state-metrics/tree/master/docs#default-resources) that are relevant for the user to know? For example, controller-runtime or even client-go by default exposes certain metrics which are useful from an operator authors perspective. Are there any specific metrics out of the lot, which we can create wrappers around and utilize for our benefits?

> Does it necessary to work with prometheus-operator/kube-prometheus

IMO though kube-prometheus is used widely we should not assume our configurations or scaffoldings to be only based on the same. There are many production use cases wherein the entire monitoring stack can be off-cluster. Also, is there any generic workaround to identify any prometheus on-cluster instance and have an option to allow users to auto-load the configuration if they like to do so?

test/testdata/generate.sh
```diff
@@ -118,6 +118,11 @@ function scaffold_test_project {
       $kb edit --plugins=grafana.kubebuilder.io/v1-alpha
   fi
 
+  if [[ $project == "project-v3" || $project == "project-v4" || $project =~ multigroup ]]; then
```
camilamacedo86, line 121: WDYT about: 
change this sample `project-v3-addon-and-grafana` name for project-v3-declarative-v1
then, we have a new one as project-v3-metrics were we will add grafana and this one.

Same for v4.

pkg/plugins/optional/kube-state-metrics/v1alpha/scaffolds/internal/templates/config.go
Kavinjsir: Yeah, I think it a good idea to give config entry for users to decide which ones to be intrumented. 👍🏼

pkg/plugins/optional/kube-state-metrics/v1alpha/scaffolds/internal/templates/config.go
Kavinjsir: Yeah, I guess the target is to make the metrics array 'configurable'.

The `_info` here is like a sample and practices.
It might can become a default one if considered common. (Well, the content/labels might need improvement)

Ac

test/testdata/generate.sh
Kavinjsir: Yeah, I think maybe we can deal with this in a separate PR? Since this one maybe hold.

Kavinjsir:
@camilamacedo86 @varshaprasad96 As discussed in the meeting, it seems like `monitoring plugin` can be a good idea to integrate both `kube-state-metrics` and `grafana` plugin, I am going to hold this PR for now.

test/testdata/generate.sh
Kavinjsir: @camilamacedo86 Here is the PR to refact `testdata`: https://github.com/kubernetes-sigs/kubebuilder/pull/3086

Kavinjsir:
> 

Hi @varshaprasad96 glad to have your ideas with considerations on different perspectives! 👍🏼 
While not have much exploration on kube-state-metrics, I feel like it is generally providing metrics based on k8s apis' status.
A common case can be the `conditions` array in objects' status field.
Regardless of the customized `subresource status`, it instruments the native apis (deploment, pod, job, ...).

The reason why I was wondering if necessary to provide approaches deploying ksm is because our feature is based on that...

And about what you've mentioned of the monitoring stacks running outside cluster, which pretty makes sense, and so I can understand maybe it is not necessary for kubebuilder to deal with kube-prometheus...

Actually, I'm interested in `the monitoring plugin` we've been discussed in the weekly. It feels like, we may handle monitoring stacks over there..

pkg/plugins/optional/kube-state-metrics/v1alpha/scaffolds/internal/templates/config.go
```diff
@@ -0,0 +1,93 @@
...
+        group: "{{.Group}}.{{.Domain}}"
+        kind: "{{.Kind}}"
+        version: "{{.Version}}"
+      labelsFromPath:
+        name:
+          - metadata
+          - name
+        namespace:
+          - metadata
+          - namespace
```
umangachapagain, line 80: Won't this cause issues with cluster-scoped resources?

Kavinjsir:
@camilamacedo86 Since the origin issue is closed by the author where kube-state-metrics [provides native support for custom resources](https://github.com/kubernetes/kube-state-metrics/blob/v2.10.1/docs/customresourcestate-metrics.md), I guess we may close this PR at this moment?

Kavinjsir:
Close this PR as the origin issue was closed by the author. Feel free to open new issue or leave comment here for relative discussions.

PR #3077, opened 2022-11-16
:sparkles: add version check for makefile template tools

coriaedu:
**Description of the change:**
Test if the installed versions of `kustomize` and `controller-gen` match the ones provided in the Makefile corresponding variables `KUSTOMIZE_VERSION` and `CONTROLLER_TOOLS_VERSION` to determine whether to re-install or not.

**Motivation for the change:**
Prevent running `controller-gen` and `kustomize` commands with a version that does not match the one specified in the corresponding variables. (details of scenario in #3068)

* Fixes #3068

PR #3079, opened 2022-11-16
:bug: fix unexpected end of file error in makefiles

### LINKED ISSUE #3068: check for tool version in Makefile before installing
#### Describe the problem you need a feature to resolve.

The **Makefile** that is scaffolded for golang has two targets that will download [`controller-gen`](https://github.com/kubernetes-sigs/kubebuilder/blob/6f1f8c43cfbf260c971037ff039cca86a0980006/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go#L205) and [`kustomize`](https://github.com/kubernetes-sigs/kubebuilder/blob/6f1f8c43cfbf260c971037ff039cca86a0980006/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go#L200) tools, with the versions corresponding to `CONTROLLER_TOOLS_VERSION` and `KUSTOMIZE_VERSION` vars.

The recipes there are currently checking for file existence before installing the tool, but they don't confirm the version.

In my scenario, where I am upgrading the OSDK, I found this issue with the current Makefile:

* checkout clean branch with OSDK 1.24.0
* Run `make controller-gen` -> This will download controller-gen v0.9.2 at `$(LOCALBIN)`
* upgrade OSDK to 1.25.1
    * This includes updating the `CONTROLLER_TOOLS_VERSION` to `v0.10.0`
* Run any command that uses `controller-gen` thinking that `v0.10.0` will be used, but it is still running `0.9.2`

The same thing applies when I switch branches that have different OSDK versions

This applies to `kustomize` also.

#### Describe the solution you'd like.

I'd like the `controller-gen` and `kustomize` targets to check for version so we ensure that it matches the corresponding VERSION var. 

### Extra Labels

_No response_

coriaedu:
**Description of the change:**
Add backslash (\) character to the `kustomize` recipe to fix `unexpected end of file` error

**Motivation for the change:**
Fix the error that caused these tests to fail: https://github.com/kubernetes-sigs/kubebuilder/pull/3077#issuecomment-1317613311

* Fixes #3068

### DIFF
--- a/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
@@ -199,9 +199,9 @@ KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/k
 .PHONY: kustomize
 kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary. If wrong version is installed, it will be removed before downloading.
 $(KUSTOMIZE): $(LOCALBIN)
-	@if test -x $(LOCALBIN)/kustomize && ! $(LOCALBIN)/kustomize version | grep -q $(KUSTOMIZE_VERSION); then
-		echo "$(LOCALBIN)/kustomize version is not expected $(KUSTOMIZE_VERSION). Removing it before installing.";
-		rm -rf $(LOCALBIN)/kustomize;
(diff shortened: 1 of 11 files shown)


coriaedu:
@everettraven here's the fix that continues the work of #3077

everettraven (approved):
/lgtm

camilamacedo86 (approved):

camilamacedo86:
How the issue passed in the tests/ci? We need to check it out,
@coriaedu could you please raise an issue for that?

PR #3102, opened 2022-11-26
🐛  (kustomize/v2-alpha): use marker to track samples and do not create tracks for core types

camilamacedo86:
## Description

Instead of overwriting the file, we can use a marker so we will just append new values instead. 

## Motivations
- allow manual changes
- fix: does not track values for core types (such as deployment)
- fix layout of the scaffold

### DIFF
--- a/pkg/plugins/common/kustomize/v2-alpha/scaffolds/api.go
+++ b/pkg/plugins/common/kustomize/v2-alpha/scaffolds/api.go
@@ -18,8 +18,6 @@ package scaffolds
 
 import (
 	"fmt"
-	"strings"
-
 	"sigs.k8s.io/kubebuilder/v3/pkg/config"
 	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
 	"sigs.k8s.io/kubebuilder/v3/pkg/model/resource"
@@ -69,23 +67,10 @@ func (s *apiScaffolder) Scaffold() error {
 		machinery.WithResource(&s.resource),
 	)
 
-	rs, err := s.config.GetResources()
-	if err != nil {
-		return err
-	}
-
-	crdManifests := []string{}
-	for _, r := range rs {
-		crdManifests = append(crdManifests, s.generateManifestsPath(r))
-	}
-
-	crdManifests = append(crdManifests, s.generateManifestsPath(s.resource))
-
 	// Keep track of these values before the update
 	if s.resource.HasAPI() {
 		if err := scaffold.Execute(
 			&samples.CRDSample{Force: s.force},
-			&samples.Kustomization{CRDManifests: crdManifests},
 			&rbac.CRDEditorRole{},
 			&rbac.CRDViewerRole{},
 			&patches.EnableWebhookPatch{},
@@ -95,12 +80,14 @@ func (s *apiScaffolder) Scaffold() error {
 		); err != nil {
 			return fmt.Errorf("error scaffolding kustomize API manifests: %v", err)
 		}
+
+		// If the gvk is non-empty
+		if s.resource.Group != "" || s.resource.Version != "" || s.resource.Kind != "" {
+			if err := scaffold.Execute(&samples.Kustomization{}); err != nil {
+				return fmt.Errorf("error scaffolding manifests: %v", err)
+			}
+		}
 	}
 
 	return nil
-}
-
-func (s *apiScaffolder) generateManifestsPath(r resource.Resource) string {
-	// nolint: lll
-	return strings.ToLower(r.GVK.Group) + "_" + strings.ToLower(r.GVK.Version) + "_" + strings.ToLower(r.GVK.Kind) + ".yaml"
 }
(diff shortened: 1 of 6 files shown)


pkg/plugins/common/kustomize/v2-alpha/scaffolds/api.go
```diff
@@ -95,6 +82,13 @@ func (s *apiScaffolder) Scaffold() error {
 		); err != nil {
 			return fmt.Errorf("error scaffolding kustomize API manifests: %v", err)
 		}
+
+		// If the gvk is non-empty
+		if s.resource.Group != "" || s.resource.Version != "" || s.resource.Kind != "" {
```
Kavinjsir, line 87: nits: Just wondering in which case this `gvk` might be empty so that this sample kustomization should be prevented?

pkg/plugins/common/kustomize/v2-alpha/scaffolds/api.go
```diff
@@ -95,6 +82,13 @@ func (s *apiScaffolder) Scaffold() error {
...
+			if err := scaffold.Execute(&samples.Kustomization{}); err != nil {
+				return fmt.Errorf("error scaffolding manifests: %v", err)
+			}
+		}
```
Kavinjsir, line 91: nits: we may remove the method `generateManifestsPath` as it is unused.

pkg/plugins/common/kustomize/v2-alpha/scaffolds/internal/templates/config/samples/kustomization.go
```diff
@@ -17,54 +17,60 @@ limitations under the License.
...
+	machinery.ResourceMixin
 }
 
-// SetTemplateDefaults implements file.Template
+// SetTemplateDefaults implements machinery.Template
 func (f *Kustomization) SetTemplateDefaults() error {
 	if f.Path == "" {
 		f.Path = filepath.Join("config", "samples", "kustomization.yaml")
 	}
+	f.Path = f.Resource.Replacer().Replace(f.Path)
```
Kavinjsir, line 41: I'm wondering why we'd like to updated `f.Path` here? Will it be values other than `config/samples/kustomization.yaml` by calling `replacer`?

pkg/plugins/common/kustomize/v2-alpha/scaffolds/internal/templates/config/samples/kustomization.go
```diff
@@ -17,54 +17,60 @@ limitations under the License.
...
+// makeCRFileName returns a Custom Resource example file name in the same format
+// as kubebuilder's CreateAPI plugin for a gvk.
+func (f Kustomization) makeCRFileName() string {
+	return f.Resource.Replacer().Replace("%[group]_%[version]_%[kind].yaml")
+}
 
+// GetCodeFragments implements file.Inserter
+func (f *Kustomization) GetCodeFragments() machinery.CodeFragmentsMap {
+	return machinery.CodeFragmentsMap{
+		machinery.NewMarkerFor(f.Path, samplesMarker): []string{fmt.Sprintf(samplesCodeFragment, f.makeCRFileName())},
```
Kavinjsir, line 69: nits: will it be good/necessary if we declare a new method like:
```go
func (f *Kustomization) sampleMaker() machinery.Marker {
	return machinery.NewMarkerFor(f.Path, samplesMarker)
}
```
So that we may keep consistency to use sample marker in various places in the file?

Kavinjsir (changes requested):
Generally looks good! Just some small nits and questions.

It feels more safe and efficient to use marker with insert than the previous approach! 👍🏼

pkg/plugins/common/kustomize/v2-alpha/scaffolds/api.go
camilamacedo86: It is only to prevent somewhere the validation was not made before.
For the kustomize I do not think that we are validating. The check is done in the golang.
So, if be called with another plugin and etc.

pkg/plugins/common/kustomize/v2-alpha/scaffolds/internal/templates/config/samples/kustomization.go
camilamacedo86: I this case I think we only use here. right? But we could improve.

pkg/plugins/common/kustomize/v2-alpha/scaffolds/api.go
camilamacedo86: good catcher :-)

pkg/plugins/common/kustomize/v2-alpha/scaffolds/internal/templates/config/samples/kustomization.go
camilamacedo86: Good catcher, we do not need this here.

everettraven (approved):
Changes look good to me!

/lgtm

Issue #3110, opened 2022-12-04
Change sufix testdata and fix samples


camilamacedo86:
### What do you want to happen?

In the latest meeting, we agreed that we would replace the name of `project-v4-with-metrics` with `project-v4-with-grafana`. The change needs to be done at: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh

Also, see that the declarative sample should not call the grafana/v1-alpha plugin. We can fix this issue this nit as well.

jason1028kr:
Made #3111 to address the specified nits. But, let me know if I've missed something. Thanks!

camilamacedo86:
Hi @jason1028kr,

That is great! thank you for your contribution.

It is just missing one detail. Could you please also update the name in the tests: https://github.com/kubernetes-sigs/kubebuilder/blob/d887e25e0a62eae1fd5882267bc3a53535198c6d/test/testdata/test.sh#L47

ShivamTyagi12345:
/ assign

PR #3111, opened 2022-12-05
🌱 Change testdata suffix and regenerate samples

jason1028kr:
### Description
For #3110  Make corresponding changes in the `generate.sh` and fix samples.

PR #3118, opened 2022-12-06
fix test name

ShivamTyagi12345:
Signed-off-by: ShivamTyagi12345 <mergesuccessful@gmail.com>

Fixes a small test name
- #3110

### DIFF
--- a/test/testdata/test.sh
+++ b/test/testdata/test.sh
@@ -44,4 +44,4 @@ test_project project-v4-multigroup
 test_project project-v4-declarative-v1
 test_project project-v4-config
 test_project project-v4-with-deploy-image
-test_project project-v4-with-metrics
+test_project project-v4-with-grafana


test/testdata/test.sh
```diff
@@ -44,4 +44,4 @@ test_project project-v4-multigroup
 test_project project-v4-declarative-v1
 test_project project-v4-config
 test_project project-v4-with-deploy-image
-test_project project-v4-with-metrics
+test_project project-v4-with-grafana
```
camilamacedo86, line 47: This one needs to be done with : https://github.com/kubernetes-sigs/kubebuilder/pull/3111

We change the name of the samples.
So, we need to change it also in the tests so we can adequately test the projects in the CI. 
PS. the same is valid for v3 one.

camilamacedo86:
cloosing this one in favor of : https://github.com/kubernetes-sigs/kubebuilder/pull/3111

Issue #2786, opened 2022-07-01
Scaffold a comment in the PROJECT file to let users know that they should not update it


camilamacedo86:
### What do you want to happen?

I'd like to see in the PROJECT file something like:

```yaml
# Code generated by tool. DO NOT EDIT.
# This file is used to track the info used to scaffold your project
# and allow the plugins properly work.
```

So that, we could make clear that users should not change it. Motivation: https://github.com/kubernetes-sigs/kubebuilder/issues/2634

PS.: Note that if we go into the store.go and try to add the comment when the file is created the following error will be faced when we try to parse / unmarshal it:

```
Error: unable to load the configuration: unable to determine config version: error converting YAML to JSON: yaml: line 18: mapping values are not allowed in this context
```

In this way, we need to see if new versions of the libs to work with YAML used can deal with comments or check if we can provide an alternative solution.

camilamacedo86:
From the KB meeting: Maybe would be better something like "# Code generated by tool. DO NOT EDIT IF YOU DO NOT KNOW WHAT YOU ARE DOING" OR AT YOUR RISK.

jason1028kr:
Hi, @camilamacedo86 and @laxmikantbpandhare.
I've tried adding simple prepend on yaml pkg's `Save()` function for this. But I can't seem to observe the same `unable to load the configuration` error, so I've made a pr for now. Let me know if you've planned a different approach.

camilamacedo86:
Hi @jason1028kr,

We can close this one as done. Thank you for your contribution !!!

Issue #3088, opened 2022-11-20
Improve PROJECT file documentation


danaradg:
### What do you want to happen?

The kubebuilder book has a page for the `PROJECT` file that documents its structure, but doesn't provide an overview of its purpose and effect.
Following [this](https://groups.google.com/g/kubebuilder/c/5c4_Xc3Fjx0/) discussion it was suggested to incorporate the answer into the documentation.

### Extra Labels

/kind documentation

varshaprasad96:
@danaradg Thanks for raising the issue. Would you open to submitting a PR to fix this issue?

danaradg:
I would love to contribute, but will probably need some time to ramp up and better understand the kubebuilder arch and plugins in general. If you don't mind waiting, I don't mind putting the effort. Would also love reading material recommendations.

Sajiyah-Salat:
I have gone through the merged pr and google group conversation. is @danaradg working on this issue.

danaradg:
Honestly, haven't gotten around to it. If you want to take over, feel free.

Sajiyah-Salat:
Great. Thank you.
/assign

ashutosh887:
Please let me work on this issue @camilamacedo86 @varshaprasad96

camilamacedo86:
HI @Sajiyah-Salat,
 
Yes, this issue still valid. We need to improve the documentation with what was shared in the email thread.

Sajiyah-Salat:
Sure. I will work on that shortly.

Sajiyah-Salat:
can you please guide me to the file where you would like to seethe proejct information.

varshaprasad96:
@Sajiyah-Salat you could add more details on what PROJECT config file is in here: https://book.kubebuilder.io/reference/project-config.html

Sajiyah-Salat:
Hello @ashutosh887 you have worked well on this pr. congrats on this. I think the work is completed now. Should we close this pr? wdyt? @camilamacedo86

ashutosh887:
Yes the PR is merged as well @Sajiyah-Salat @camilamacedo86

Sajiyah-Salat:
I know that.

ashutosh887:
Let's close this then 😄🙌🏻

PR #3137, opened 2022-12-26
✨ warning comment on PROJECT file

jason1028kr:
For #2786
Prepends comment on a call to yaml.Save() function.


pkg/config/store/yaml/store.go
```diff
@@ -31,6 +31,12 @@ import (
 const (
 	// DefaultPath is the default path for the configuration file
 	DefaultPath = "PROJECT"
+
+	// Comment for 'PROJECT' config file
+	commentStr = `# Code generated by tool. DO NOT EDIT.
+# This file is used to track the info used to scaffold your project
+# and allow the plugins properly work.
```
camilamacedo86, line 38: That is great. Could we just add? 
```suggestion
# and allow the plugins properly work. 
# More info: https://book.kubebuilder.io/reference/project-config.html
```

So that user have more info. also, we would like to improve this doc to have a better explanation, see: https://github.com/kubernetes-sigs/kubebuilder/issues/3088

camilamacedo86 (approved):
Just a nit: https://github.com/kubernetes-sigs/kubebuilder/pull/3137/files#r1059552492
Thank you a lot for collab with. 

Otherwise 
/approved

pkg/config/store/yaml/store.go
jason1028kr: sounds good! I've added the link to the comment

jason1028kr:
@camilamacedo86 Thanks! I've added the link to the comment

PR #3143, opened 2023-01-03
📖 missing shell argument in autocompletion snippet

GabriFedi97:
The documentation's snippet provided to enable the autocompletion, uses the command `kubebuilder completion zsh|bash|powershell|fish` but the argument is missing. Looks like it should be a ready to go, copy & paste, snippet to be added to the `.bash_profile` but it does not work as it is, due to the missing argument.
Since the example relies on the shell `bash`, according to the `kubebuilder completion` subcommand documentation, the missing argument should be fed with `bash`.


everettraven (approved):
@GabriFedi97 thanks for the contribution! These changes look good to me!

/lgtm

camilamacedo86 (approved):

camilamacedo86:
@GabriFedi97,

Could you please rebase this one with master branch for we are able to get it merged.

PR #3161, opened 2023-01-16
:book: remove the workaround steps from the release doc

camilamacedo86:
## Description

remove the workaround steps from the release doc.
these steps are from the legacy method, which is no longer used


camilamacedo86:
Hi @Kavinjsir 

>Quick question:
> Is that because we are using Github Action where we won't face the case to release locally any more?

Before we used Google Cloud to do the builds. So, this steps were to help out when we faced an issue with the old process.
Now, we are using github actions so that is no longer accurate.

Issue #3177, opened 2023-01-23
directory prefix . does not contain modules listed in go.work or their selected dependencies


KrystianMarek:
### What broke? What's expected?

`kubebuilder create api` failed

### Reproducing this issue

```
❯ kubebuilder create api --group webapp --version v1 --kind Guestbook
Create Resource [y/n]
y
Create Controller [y/n]
y
Writing kustomize manifests for you to edit...
Writing scaffold for you to edit...
api/v1/guestbook_types.go
api/v1/groupversion_info.go
internal/controller/suite_test.go
internal/controller/guestbook_controller.go
Update dependencies:
$ go mod tidy
Running make:
$ make generate
mkdir -p /Users/marek/Development/go_playground/13_capi/guestbook/bin
test -s /Users/marek/Development/go_playground/13_capi/guestbook/bin/controller-gen && /Users/marek/Development/go_playground/13_capi/guestbook/bin/controller-gen --version | grep -q v0.11.1 || \
	GOBIN=/Users/marek/Development/go_playground/13_capi/guestbook/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.11.1
/Users/marek/Development/go_playground/13_capi/guestbook/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
-: pattern ./...: directory prefix . does not contain modules listed in go.work or their selected dependencies
Error: not all generators ran successfully
run `controller-gen object:headerFile=hack/boilerplate.go.txt paths=./... -w` to see all available markers, or `controller-gen object:headerFile=hack/boilerplate.go.txt paths=./... -h` for usage
make: *** [generate] Error 1
Error: failed to create API: unable to run post-scaffold tasks of "base.go.kubebuilder.io/v4-alpha": exit status 2
Usage:
  kubebuilder create api [flags]

Examples:
  # Create a frigates API with Group: ship, Version: v1beta1 and Kind: Frigate
  kubebuilder create api --group ship --version v1beta1 --kind Frigate

  # Edit the API Scheme
```

### KubeBuilder (CLI) Version

Version: main.version{KubeBuilderVersion:"3.9.0", KubernetesVendor:"1.26.0", GitCommit:"26f605e889b2215120f73ea42b081efac99f5162", BuildDate:"2023-01-16T17:21:30Z", GoOs:"darwin", GoArch:"arm64"}

### PROJECT version

version: "3"

### Plugin versions

```yaml
[...]

PR #3178, opened 2023-01-23
:bug: Fixes directory prefix issue in golang/v4 plugin

KrystianMarek:
I think it solves https://github.com/kubernetes-sigs/kubebuilder/issues/3177


pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -100,23 +100,23 @@ help: ## Display this help.
```
camilamacedo86, line 100: If you change the bollerplate than you need to run make generate to ensure that you will update all samples with under the testdata.

Also, is the problem that you are trying to solve here faced with the default scaffold or only if you do some specific manually change on it?

KrystianMarek:
The problem was elsewhere. My mistake, I had a `go.work` in the parent folder.

Issue #2510, opened 2022-02-13
[DX/tests] - Update doc tutorials programatically for any PR change


camilamacedo86:
### Description 

Currently, we have the following samples which are used in the docs: 

- Component Config: https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/component-config-tutorial/testdata
- CronJob Tutorial: https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/cronjob-tutorial/testdata
- Multiversion-tutorial: https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/multiversion-tutorial/testdata

You can see that the docs are using the samples and the info provided on them are comments in the source code generated. 
However, it has been very thought we keep updated with the changes performed in the scaffolds. Currently, we need to:

- Run the helpers scripts under it samples dir ([generate_cronjob.sh](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/generate_cronjob.sh), [generate_multiversion.sh](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/multiversion-tutorial/testdata/generate_multiversion.sh), [generate_componentconfig.sh](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/component-config-tutorial/testdata/generate_componentconfig.sh) )
- Run make all
- Then, compare the local branch with the master branch and then re-add the missing files and all code generated for the samples as the comments back
- Run make all again to ensure and compare again to see if we are not missing anything

### Goal

The goal of this task is we programmatically ensure that all samples will be updated for any pull request sent to the project. It. will help us keep the docs and update and improve the DX for contributors. Also, it can help us check the changes in the review process which also help us to easily ensure them. 

Note that we have a shell script [generate.sh](https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh) which is called via the makefile target [make generate] (https://github.com/kubernetes-sigs/kubebuilder/blob/master/Makefile#L65-L67) to re-gen all samples under the [testdata](https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata) dir.

[...]

PR #3191, opened 2023-01-31
✨ Update component-config tutorial automatically

Eileen-Yu:
## Description:
1. Automatically update the scaffolded project of the component-config tutorial.
2. Introduce a new CI pipeline for docs

## Motivation:
Attempt to discuss the implementation and try to solve the first part of https://github.com/kubernetes-sigs/kubebuilder/issues/2510


Kavinjsir (changes requested):
Nice job @Eileen-Yu 👍🏼 !
I did a first-round review and left some comments.
Will take another look later.
Thx for participating in contribution!

Makefile
```diff
@@ -63,14 +63,18 @@ install: build ## Build and install the binary with the current source code. Use
...
 	go mod tidy
 
 .PHONY: generate-testdata
 generate-testdata: ## Update/generate the testdata in $GOPATH/src/sigs.k8s.io/kubebuilder
 	rm -rf testdata/
 	./test/testdata/generate.sh
 
+.PHONY: generate-docs
+generate-docs: ## Update/generate the docs in $GOPATH/src/sigs.k8s.io/kubebuilder
+	./docs.sh
```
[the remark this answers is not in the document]
Eileen-Yu: Sounds good 👍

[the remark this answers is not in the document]
camilamacedo86: Can we not call the golang implementation directly?

docs.sh
```diff
@@ -0,0 +1,26 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+source "test/common.sh"
+
+build_kb
```
camilamacedo86, line 19: we do not need to do those calls.
Whe we can add those via the Makefile right?

docs/book/src/component-config-tutorial/generate_docs.go
```diff
@@ -0,0 +1,189 @@
...
+	pluginutil "sigs.k8s.io/kubebuilder/v3/pkg/plugin/util"
+	"sigs.k8s.io/kubebuilder/v3/test/e2e/utils"
+)
+
+type Sample struct {
+	ctx *utils.TestContext
+}
+
+func NewSample(binaryPath, samplePath string) Sample {
+	log.Infof("Generating the sample context...")
```
camilamacedo86, line 33: can we add in the text the name of the sample that we are generating so that if we face problems, we will know from where the error came. WDYT?

Eileen-Yu: Sounds good 👍

Makefile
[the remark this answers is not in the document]
Eileen-Yu: Thx for reviewing @camilamacedo86 !

Does that mean we may have some alternative other than 'go run...' ? Wondering if there's any reference?

docs/docs.sh
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+go run docs/book/src/update-docs/generate_docs.go
+
+cd docs/book/src/component-config-tutorial/testdata/project
+
+make build
```
Kavinjsir, line 21: nits: what about merging the two lines:

```shell
(cd docs/book/src/component-config-tutorial/testdata/project; make build)
```

(Assuming in the future we may go to multiple tutorial directories and run the command..)

camilamacedo86: We do not need the make build to do this one. 
The build will be made in the ci already before we call make generate
Locally the contributor should work with make generate so, IMO we should keep the same behaviour.

Makefile
```diff
@@ -64,13 +64,19 @@ install: build ## Build and install the binary with the current source code. Use
...
 
 .PHONY: generate-testdata
 generate-testdata: ## Update/generate the testdata in $GOPATH/src/sigs.k8s.io/kubebuilder
 	rm -rf testdata/
 	./test/testdata/generate.sh
 
+.PHONY: generate-docs
+generate-docs: ## Update/generate the docs in $GOPATH/src/sigs.k8s.io/kubebuilder
+	source "test/common.sh" && build_kb
+	./docs/docs.sh
```
camilamacedo86, line 78:
```suggestion
	go run docs/book/src/update-docs/generate_docs.go
```

docs/book/src/component-config-tutorial/generate_docs.go
```diff
@@ -0,0 +1,196 @@
+/*
```
camilamacedo86, line 1: Could we move this to docs/book/src/hack/internal/generate_component_config.go?
Because then:
- we have the script under the hack, which makes it clear that is a script/helper
- we do not allow anybody to import this code implementation.

docs/book/src/component-config-tutorial/generate_docs.go
```diff
@@ -0,0 +1,196 @@
...
+Copyright 2023 The Kubernetes Authors.
+
```
camilamacedo86, line 3: Could you please remove the shell script we use as a helper today since we are now moving for the code generated?

docs/book/src/component-config-tutorial/generate_docs.go
```diff
@@ -0,0 +1,196 @@
...
+  resourceName: 80807133.tutorial.kubebuilder.io
+clusterName: example-test`), 0600,
+	)
+	CheckError("fixing controller_manager_config", err)
+
+	// 2. add componentconfig
+	err = pluginutil.InsertCode(
+		filepath.Join(sp.ctx.Dir, "PROJECT"),
+		"More info: https://book.kubebuilder.io/reference/project-config.html",
+		"\ncomponentConfig: true",
```
camilamacedo86, line 120: We should not need to add this one. 
ComponentConfig should be true when we scaffold with `kubebuilder init --domain tutorial.kubebuilder.io --component-config`. If we are not making it true then we have a bug, and we can sort it out in a follow-up.

docs/book/src/component-config-tutorial/generate_docs.go
```diff
@@ -0,0 +1,196 @@
...
+	)
+	CheckError("fixing PROJECT", err)
+
+	// 3. fix projectconfig_types.go
+	err = pluginutil.InsertCode(
+		filepath.Join(sp.ctx.Dir, "api/v2/projectconfig_types.go"),
+		"metav1 \"k8s.io/apimachinery/pkg/apis/meta/v1\"",
+		"\n  cfg \"sigs.k8s.io/controller-runtime/pkg/config/v1alpha1\"",
```
camilamacedo86, line 128: Are you using GolangIDE? 
If so, it has a very helpful string plugin.

By using this, we can easily create const for the strings.
If you be able to replace it with consts such as https://github.com/operator-framework/operator-sdk/blob/master/hack/generate/samples/internal/go/memcached-with-customization/memcached_with_customization.go#L1286-L1299 that would be great !!! 

**But it is not a blocker at all for we get this one merged**

docs/book/src/component-config-tutorial/generate_docs.go
```diff
@@ -0,0 +1,196 @@
...
+		"\n        args:\n        - \"--config=controller_manager_config.yaml\"\n        volumeMounts:\n        - name: manager-config\n          mountPath: /controller_manager_config.yaml\n          subPath: controller_manager_config.yaml\n      volumes:\n      - name: manager-config\n        configMap:\n          name: manager-config",
+	)
+	CheckError("fixing manager_config_patch", err)
+
+	// 6. fix main
+	err = pluginutil.InsertCode(
+		filepath.Join(sp.ctx.Dir, "main.go"),
+		"ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))",
+		//nolint:lll
+		"\n\n  var err error\n  ctrlConfig := configv2.ProjectConfig{}\n  options := ctrl.Options{Scheme: scheme}\n  if configFile != \"\" {\n    options, err = options.AndFrom(ctrl.ConfigFile().AtPath(configFile).OfKind(&ctrlConfig))\n    if err != nil {\n      setupLog.Error(err, \"unable to load the config file\")\n      os.Exit(1)\n    }\n  }",
```
camilamacedo86, line 171: It also should not be required because we are using the flag --component-config and that should do the scaffold accordingly. Can you try  --component-config=true just to see if the bug is we do have not a default true value? 
If you see that it is not working, then could you please raise an issue for us to fix as a follow? Up add a TODO here linking the issue so we are able to remove all after fixing it.

docs/book/src/component-config-tutorial/testdata/project/config/manager/controller_manager_config.yaml
```diff
@@ -17,4 +17,4 @@ webhook:
 leaderElection:
   leaderElect: true
   resourceName: 80807133.tutorial.kubebuilder.io
-clusterName: example-test
+clusterName: example-test
```
camilamacedo86, line 20: it is missing an space at the end

camilamacedo86:
/ok-to-test

That is a terrific contribution 🥇 
Just a few nits and them it will be great to get merged.

docs/book/src/component-config-tutorial/generate_docs.go
Eileen-Yu: Make sense 👍

docs/book/src/component-config-tutorial/generate_docs.go
Eileen-Yu: Thx for the suggestion!

I tried to follow the idea and converted it to string literal. Let me know if it works or not

(Sadly my pc is poor to drive GolangIDE so I'm using terminal😭)

docs/book/src/component-config-tutorial/generate_docs.go
Eileen-Yu: Same as above. Fixed.

docs.sh
Eileen-Yu: Hi @camilamacedo86 I have a question here:

I can understand that if we run `make generate`, the `build_kb` will be loaded and kubebuilder would be built.

What I'm thinking is, if it's necessary to have a separate cmd, that can individually generate docs based on the code?

What we have now is `make generate-docs`, that does not include `build_kb`.

docs/docs.sh
Eileen-Yu: Do we have the step to execute codegen for those tutorials in the CI?

I failed to find the place. Do you have the reference?

camilamacedo86: We call in the CI make check-testdata, see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/.github/workflows/testdata.yml#L27

Then, see that this target calls ./test/testdata/check.sh: https://github.com/kubernetes-sigs/kubebuilder/blob/master/Makefile#L120-L122

See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/check.sh#L17-L39

So, my thought, in the long run, is we have a makefile target that calls that run the go implementation to generate docs/ Also, add this one to make generate. And then, ensure that we are testing/checking for all PR changes if the docs were also generated/is updated accordingly. 

Is that make sense? Such as we do in SDK to verify the samples as well.

camilamacedo86, line 21: @Eileen-Yu , why are we calling the make build in the shell?
We can call that via golang see, i.e. https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v4/plugin_cluster_test.go#L118

So, we can make("build") as well.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,35 @@
...
+        uses: actions/setup-go@v3
+        with:
+          go-version: "1.19"
+      - name: Remove pre-installed kustomize
+        # This step is needed as the following one tries to remove
+        # kustomize for each test but has no permission to do so
+        run: sudo rm -f /usr/local/bin/kustomize
+      - name: Verify docs
+        run: |
+          make generate-docs
```
Kavinjsir, line 27: See that CI has reported [an error for missing `kubebuilder`](https://github.com/kubernetes-sigs/kubebuilder/actions/runs/4081288605/jobs/7034545030#step:5:34). That is possibly due to the execution of `make generate-docs` without `build_kb`.

I think we might have two approaches:
1. In the new ci yaml file, add the setup to load and run `build_kb`, and then execute `make generate-docs`.
2. Drop this new ci file, and add `generate-docs` to `make generate`.

WDYT? @Eileen-Yu @camilamacedo86 

Well, I like the idea to have a separate pipeline focusing on docs validation.

camilamacedo86: I think we should not use `build_kb` and we should do == make generate and check testdata

Kavinjsir: Hi @camilamacedo86
It makes sense to integrate the docs generation into `check-the-testdata`, that can make the implementation a lot easier and straightforward.
Based on that, I'm concerned about some following points:
1. If CI fails at `check the testdata`, will it become hard to tell at the first glance which part is wrong: either `testdata` OR `docs`? 
2. Would it be slow if we run all the generations in one pipeline versus having multiple pipelines for each generation?
3. Is the priority of keeping `testdata` consistent higher than `docs`?

camilamacedo86: It is OK to be another CI test. I am just saying that the implementation is very similar == to the target to generate the testdata samples and check it out.

Eileen-Yu: Hi @camilamacedo86 I just added CI for the docs check.
Could you pls help with review?

@Kavinjsir It sounds good to have multiple pipelines. How would you like to deal with this in some future PR?

camilamacedo86 (approved):
That is a terrific work 🥇 

@Kavinjsir I am moving forward, but if you see the hall for improvements/changes, please track those issues, and we can try to address them in a fallow up.

/lgtm

Issue #3198, opened 2023-02-03
Bump kustomize v5 into kustomize/v2-alpha before it get stable 


camilamacedo86:
### What do you want to happen?

We need to bump https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv5.0.0 to kustomize-v2 alpha plugin. You will need to do the bump and run make generate to ensure that all samples are updated with,

Also, if you face breaking changes then we need to change the scaffolds for this version.

We should also update the documents for kustomize/v2-alpha plugin, and go/v4-alpha as the migration steps from gov/v3 to go/v4-alpha to say that those are using kustomize v5 and not v4.

camilamacedo86:
Done.

PR #3199, opened 2023-02-03
⚠️ (kustomize/v2-alpha, go/v4-alpha): Upgrade kustomize from 4.5.7 to 5.0.0

Eileen-Yu:
## Description:
1. Bump kustomize from 4.5.7 to 5.0.0 for kustomize v2-alpha plugin
2. Update related samples
3. Update related docs

## Motivation:
Aims to solve https://github.com/kubernetes-sigs/kubebuilder/issues/3198


camilamacedo86 (commented):
Hi @Eileen-Yu,

Thank you !!! That is great 👍 

I just updated the title because we use it to generate the changelog
See that since it is a major bump has breaking changes and we will need to highlight them out.

Issue #3119, opened 2022-12-06
Ensure the quality of the projects scaffolded within kubuilder


camilamacedo86:
## What is the requirement?

USE CASE: I am as a developer using Kubebuilde,  I'd like that Kubebuilder to provide in the scaffold all that could be required and are common practices to ensure the quality of my project so that I do not need to implement these common needs, and I can leverage them in KB. (to speed up and skill up)

The projects scaffolded with Kubebuilder do not provide common needs:

- Does not use golangci-lint https://github.com/kubernetes-sigs/kubebuilder/issues/1887 (We could have target for that)
- Does not provide e2e tests (Kubebuilder as lib could be used to address this need). Example implementation that could be scaffolded (we need to remove the utils and ensure that we provide them via Kubebuilder as lib instead): https://github.com/operator-framework/operator-sdk/tree/master/testdata/go/v4-alpha/memcached-operator/test
- Does not provide github actions calling the e2e and unit tests 

## Proposed solution(s):

**Option A (Addressing this needs via a plugin to allow users opt-in or not by)**

Create a new plugin (qa/v1-alpha) that would:
- scaffold the golangci-lint config file ( same config used by KB ) and a new target 
- scaffold the e2e tests such as https://github.com/operator-framework/operator-sdk/tree/master/testdata/go/v4-alpha/memcached-operator/test 
- ensure that Kubebuilder can be used as lib ( exporting the same utils used in KB tests ) to allow authors perform the e2e tests (we should not scaffold the lib features as utils as in the above example)
- scaffold GitHub actions to call unit tests ( done within envtest and which are scaffolded by default ) and the e2e tests added above 
- scaffold dependabot to ensure that projects are properly updated regards their dependencies (i.e https://github.com/kubernetes-sigs/kubebuilder/blob/master/.github/dependabot.yml)

**Option B (Addressing this needs in the default scaffolds)**

In this case, the default go scaffolds would be incorporated to address the above by default.

Kavinjsir:
I think it is a very good idea to support quality assurance. 👍🏼 
I vote for the individual plugin focusing on this feature.

Some thoughts below:
1. Considering on the fact that `golangci-lint` and `e2e tests` can be language specific, what about we create under `pkg/plugins/golang/qa/v1-alpha`?
2. One point not sure to me is the position that `kubebuilder lib` gets scaffolded. Does it depend on the existing layout of the project? (e.g: `./lib` vs `./pkg/lib`)
[...]

PR #3216, opened 2023-02-10
✨ Check lint of scaffolded projects in CI

Eileen-Yu:
## Description:
Add golangci-lint check for the projects scaffolds within kubebuilder in CI.

## Motivation:
1. Ensure the project quality
2. Reduce maintenance trifles
3. Aims to solve https://github.com/kubernetes-sigs/kubebuilder/issues/3119


.github/workflows/lint.yml
```diff
@@ -36,11 +36,13 @@ jobs:
           go-version: 1.19
```
camilamacedo86, line 36: @Eileen-Yu,

I was looking this one and I think we have a better way to do that.

See that in our CI we use the lint https://github.com/kubernetes-sigs/kubebuilder/blob/master/.github/workflows/lint.yml#L40 for the Kubebuilder code base.

The github action is nice because is visual and create comments in the pr code to let us know what is wrong.
Then, I checked and it has an option where we can pass the directories 

See:

```
# Optional: working directory, useful for monorepos
          # working-directory: somedir
``` 

From : https://github.com/golangci/golangci-lint-action

So, I think we could:
a) Copy and pate the .github/workflows/lint.yml to .github/workflows/lint_testdata.yml
b) We probably will need to create many jobs one for each testdata project
c) We could have the make lint-testdata and make linte-fix-testdata which will be targets that will do the same that we do for Kubebuilder but for the testdatda projects with a shell script as you did . So that as a dev, we can run them locally to test things out.   

WDYT?

Makefile
```diff
@@ -87,6 +87,10 @@ lint: golangci-lint yamllint ## Run golangci-lint linter & yamllint
 lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
 	$(GOLANGCI_LINT) run --fix
 
+.PHONY: lint-test
+lint-test: golangci-lint ## Run golangci-lint for scaffolded project
+	./lint-testdata.sh
```
camilamacedo86, line 92:
```suggestion
.PHONY: lint-testada
lint-testada: golangci-lint ## Run golangci-lint for the lscaffolded projects under testdata directory
	./lint-testada
```
Could we call something like lint-testdata or lint-samples to be more intuitive

lint-testdata.sh
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+root_dir=$(pwd)
+
+GOLANGCI_LINT="${root_dir}/bin/golangci-lint"
+go env -w GOFLAGS=-mod=mod
```
camilamacedo86, line 20: why do need this line?

lint-testdata.sh
```diff
@@ -0,0 +1,25 @@
...
+
+for project in testdata/*; do
+  (cd $project && pwd && $GOLANGCI_LINT run)
+done
```
camilamacedo86, line 24: what about we have switch here where we can call both: 
$GOLANGCI_LINT run and $GOLANGCI_LINT run --fix

.github/workflows/lint.yml
```diff
@@ -36,11 +36,13 @@ jobs:
...
       - name: Clone the code
         uses: actions/checkout@v3
-      - name: Run linter
+      - name: Run linter for codebase
         uses: golangci/golangci-lint-action@v3
         with:
           version: v1.50
           only-new-issues: true # Show only new issues if it's a pull request
+      - name: Run linter for scaffolded projects
+        run: make lint-test
```
camilamacedo86, line 45: It seems not working because we have no failures and I do not think that the projects that will scaffold would pass successfully in the check.

.github/workflows/lint.yml
Eileen-Yu: Sounds good 👍

I tried to apply reusable workflow to drop duplications. 

I set the reference as the master branch, which supposed to be able to work when this PR gets merged. Would this be good?

lint-testdata.sh
Eileen-Yu: I dropped this as it is not necessary any more.
Thx for pointing out!

lint-testdata.sh
Eileen-Yu: This sounds better. I tried to implement in the latest update. 

Would that look good to you?

.github/workflows/lint-testdata.yml
```diff
@@ -0,0 +1,85 @@
+name: Testdata Lint Check
+
+# Trigger the workflow on pull requests and direct pushes to any branch
+on:
+  push:
+  pull_request:
+
+jobs:
+  check-lint-for-project-v2:
+    uses: kubernetes-sigs/kubebuilder/.github/workflows/lint-testdata.yml@master
```
Kavinjsir, line 10: I guess we will be using `lint-testdata-template.yml` here?

Kavinjsir (changes requested):

.github/workflows/lint-testdata.yml
```diff
@@ -0,0 +1,85 @@
...
+    with:
+      name: project-v2
+      working-directory: testdata/project-v2
+
+  check-lint-for-project-v3:
```
camilamacedo86, line 15: I think we can test only for go/v4 and no longer care about v3.
We are close to make the go/v4 stable and make the default scaffold which would mean
deprecate v3.

camilamacedo86 (commented):
Hi @Eileen-Yu,

The implementation shows fine. Just a few nits : https://github.com/kubernetes-sigs/kubebuilder/pull/3216/files#r1105460858 and https://github.com/kubernetes-sigs/kubebuilder/pull/3216/files#r1105340810

However, when you run locally the check for the testdata is that working?
Are you seen issues? I do not think that the projects would pass on that.

.github/workflows/lint-testdata.yml
Eileen-Yu: Thx for catching! 👍

Eileen-Yu:
> However, when you run locally the check for the testdata is that working? Are you seen issues? I do not think that the projects would pass on that.

Yes unfortunately there are lots of lint error :(
Would pick some time to try to fix.

camilamacedo86:
Hi @Eileen-Yu,

> Yes unfortunately there are lots of lint error :(

So, this CI check should fail right :-) 
We need to fix that. I would expected we check the comments with the lint issues on it. 
OR maybe it will only occurs after we  get it merged for a next follow up PR? 
That might be the reason for we not seem them. 

> Because these projects are ONLY scaffolded by kubebuilder commands.
> We use them to validate the code consistency.
> If we want to enforce lint to them, then we will have to modify the content of them right?

The goal is to force otherwise we think that all is OK when it is not. 
And yes, we will need to fix the issues. 
However, we can:
- a) Just check and fix go/v4 ones (go/v3 will be deprecated soon)
- b) See the errors here first and them open another PRs to sort them out.

.github/workflows/lint-testdata-template.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          go-version: 1.19
+      - name: Clone the code
+        uses: actions/checkout@v3
+      - name: Run linter for ${{ inputs.name }}
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: v1.50
+          working-directory: ${{ inputs.working-directory }}
+          only-new-issues: true # Show only new issues if it's a pull request
```
camilamacedo86, line 31:
```suggestion
```

camilamacedo86: I think is because of it we are not checking the issue right?

camilamacedo86: However, I am OK with either we move in this way and then create a task for we sort out all issues as follow up.
What do you think ?

.github/workflows/lint-testdata.yml
```diff
@@ -0,0 +1,43 @@
+name: Testdata Lint Check
+
+# Trigger the workflow on pull requests and direct pushes to any branch
+on:
+  push:
+  pull_request:
+
+jobs:
+  check-lint-for-project-v4:
+    uses: kubernetes-sigs/kubebuilder/.github/workflows/lint-testdata-template.yml@master
```
everettraven, line 10: So I think using a template is a really neat approach, but not sure it is necessary. Since we already have a target being added to the Makefile that *should* automatically capture any new testdata directories I think it would be easier to maintain an action workflow that simply calls that make target. 

That being said, if there is agreement from other maintainers on taking this approach I am fine with it as well.

ashutosh887:
@Eileen-Yu let me know if you need any help resolving the Issue. Let's resolve this together if you want 😊✨

Makefile
```diff
@@ -87,6 +87,14 @@ lint: golangci-lint yamllint ## Run golangci-lint linter & yamllint
 lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
 	$(GOLANGCI_LINT) run --fix
 
+.PHONY: lint-testdata
+lint-testdata: golangci-lint ## Run golangci-lint for the lscaffolded projects under testdata directory
+	./lint-testdata.sh
+
+.PHONY: lint-testdata-fix
+lint-testdata-fix: golangci-lint ## Run golangci-lint for the lscaffolded projects under testdata directory
+	./lint-testdata.sh fix
```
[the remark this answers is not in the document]
camilamacedo86: I do not think so, we should not run fix-lint in the CI
It is ONLY a helper for us in the develop env and does not fix all issues automatically.
It can only fix issues like "go fmt" and extra imports.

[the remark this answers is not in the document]
everettraven: >It is a helper for dev env and not be called in the CI.
We should not try to fix issues in the CI at all.
Also, it does not fix all issues automatically.
It can only fix issues like "go fmt" and extra imports.
>
>I mean:
>
  >  In the CI we just check
  >  Locally, in dev env, we can use make lint-fix to help out.
>
>It is for we have the same make lint-fix that we have for the KB code to the samples either.

Yeah, but all the projects under the `testdata/` directory are automatically generated. Using a lint fixing command will fix the problem but the problem will come back when the testdata is regenerated if it isn't fixed in the actual scaffold logic.

Sajiyah-Salat:
@camilamacedo86  whats the reason that this pr is still just open without any activity.

camilamacedo86:
I have been thinking on this one.
We could either consider create a new plugin like QA plugin which would:
- Add the lint target 
- Add GitHub actions 
- Add tests and etc

I mean, todo scaffolds on top to ensure the quality of the projects following the common practices adopt by the community. 

WDYT 

@Kavinjsir @Eileen-Yu ?

Kavinjsir:
> I have been thinking on this one. We could either consider create a new plugin like QA plugin which would:
> 
> * Add the lint target
> * Add GitHub actions
> * Add tests and etc
> 
> I mean, todo scaffolds on top to ensure the quality of the projects following the common practices adopt by the community.
> 
> WDYT
> 
> @Kavinjsir @Eileen-Yu ?

@camilamacedo86 I think introducing a new plugin to help users with QA is a good idea.

From my view of the previous context, it seems challenging for the Kubebuilder repository `itself` to deal with lint check on the 'users code'. (The scaffolded code is never complete.)

For your suggestions on the proposed QA plugin, I think:
1. It is good to introduce `golangci-lint` for code quality since it looks being used widely.
2. I'm not sure if it is necessary to provide `github action` template also, since there are also other common CI/CD tools/platforms.
3. I'm not sure quite follow the idea to `add test`, are you suggesting making scaffold to code lines for unit tests OR e2e tests?

varshaprasad96:
This was brought up in the community meeting, and one of the suggestion could be to scaffold out as a `Makefile` command (as in here: https://github.com/operator-framework/helm-operator-plugins/blob/2bd2a0ab53eed258af11beee1cf8c2537a3995dc/Makefile#L64) instead of introducing a new command or plugin for this. This would make it easier for the plugin author too, to just run `make lint` instead of using a new command altogether (having a separate command for checking link might be an overkill imo).

Issue #3214, opened 2023-02-09
PluginRequest.Universe doesn't get populated before calling an external plugin


em-r:
### What broke? What's expected?

Before sending a request to an external plugin through STDIN, Kubebuilder doesn't populate [`PluginRequest.Universe`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/external/types.go#L37) with the content of the files that were modified by plugins earlier in the chain. 

For instance, if we try out the example from [Phase 2 POC](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-2.md):
`kubebuilder create api --plugins=go/v3,myexternalplugin/v2`
Files modified by `go/v3` won't be available in `Universe` field in `PluginRequest` sent to `myexternalplugin/v2`.

### Reproducing this issue

1. Set up a basic external plugin `myexternalplugin/v1` that inspects `PluginRequest.Universe`. ([example](https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/simple-external-plugin-tutorial/testdata/sampleexternalplugin/v1))
2. Initialize a new project using 2 plugins (including the one from 1.): `kubebuilder init --plugins=go/v3,myexternalplugin/v1`
3. Notice how the files modified by `go/v3` won't be available in `PluginRequest.Universe` in `myexternalplugin`.

### KubeBuilder (CLI) Version

3.9.0

### Other versions

Go version: 1.19.4 linux/amd64

em-r:
@camilamacedo86 @everettraven - I've opened this issue as discussed in the community meeting, thanks!

em-r:
based on [helpers.go#L106](https://github.com/em-r/kubebuilder/blob/master/pkg/plugins/external/helpers.go#L106), Kubebuilder will instantiate an empty before sending `PluginRequest` to an external plugin, I believe this is one of the reasons why this happens.

camilamacedo86:
Hi @em-r, 

Your help is more than welcome. Please, feel free to raise a PR against to fix this one if you wish.

em-r:
[...]

PR #3223, opened 2023-02-12
🐛 (Kubebuilder External Plugins API): fix populate PluginRequest.Universe which is empty

em-r:
This PR aims to fix an issue that is causing the `PluginRequest.Universe` not to be populated before Kubebuilder serializes `PluginRequest` and sends it to external plugins through STDIN.

To construct `PluginRequest.Universe`, the fix will scan the root directory of the project, read every file, and then add an entry to the `PluginRequest.Universe` map, where the key will be the relative path to the file, and the value the content of the file.

fixes: #3214

### DIFF
--- a/pkg/plugins/external/external_test.go
+++ b/pkg/plugins/external/external_test.go
@@ -710,15 +710,22 @@ var _ = Describe("Run external plugin using Scaffold", func() {
 
 			// create files in Filesystem
 			for _, file := range files {
-				fs.FS.MkdirAll(file.path, 0o700)
-				f, _ := fs.FS.Create(filepath.Join(file.path, file.name))
-				f.Write([]byte(file.content))
-				f.Close()
+				err := fs.FS.MkdirAll(file.path, 0o700)
+				Expect(err).ToNot(HaveOccurred())
+
+				f, err := fs.FS.Create(filepath.Join(file.path, file.name))
+				Expect(err).ToNot(HaveOccurred())
+
+				_, err = f.Write([]byte(file.content))
+				Expect(err).ToNot(HaveOccurred())
+
+				err = f.Close()
+				Expect(err).ToNot(HaveOccurred())
 			}
 
 			universe, err := getUniverseMap(fs)
 
-			Expect(err).To(BeNil())
+			Expect(err).ToNot(HaveOccurred())
 			Expect(len(universe)).To(Equal(len(files)))
 
 			for _, file := range files {
--- a/pkg/plugins/external/external_test.go
+++ b/pkg/plugins/external/external_test.go
@@ -679,6 +679,54 @@ var _ = Describe("Run external plugin using Scaffold", func() {
 			checkMetadata()
 		})
 	})
+
+	Context("Helper functions for Sending request to external plugin and parsing response", func() {
+		It("getUniverseMap should return path to content mapping of all files in Filesystem", func() {
+			fs := machinery.Filesystem{
+				FS: afero.NewMemMapFs(),
+			}
+
+			files := []struct {
+				path    string
+				name    string
+				content string
+			}{
+				{
+					path:    "./",
+					name:    "file",
+					content: "level 0 file",
+				},
+				{
+					path:    "dir/",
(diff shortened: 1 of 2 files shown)


pkg/plugins/external/helpers.go
```diff
@@ -102,8 +105,52 @@ func makePluginRequest(req external.PluginRequest, path string) (*external.Plugi
 	return &res, nil
 }
 
+// getUniverseMap is a helper function that is used to read the current directory to build
```
camilamacedo86, line 108: Would be possible to add a unit test for this implementation?

em-r: yes sure, I'll add a unit test for this, thank you for the feedback!

pkg/plugins/external/helpers.go
```diff
@@ -102,8 +105,52 @@ func makePluginRequest(req external.PluginRequest, path string) (*external.Plugi
...
+// the universe map.
+// It will return a map[string]string where the keys are relative paths to files in the directory
+// and values are the contents, or an error if an issue occured while reading one of the files.
+func getUniverseMap(fs machinery.Filesystem) (map[string]string, error) {
```
camilamacedo86, line 112: Question: Why have the path for the files is helpful? Should we not pass here the PROJECT config instead?

em-r: @camilamacedo86 - the way I see this is that when passing the path of files, we preserve the universe's file system structure, this way, the external plugin will have access to all scaffolded files the same way it would if we were to scan the root directory of the project from the external plugin.
on other hand, if there were multiple scaffolded files in the universe with the same name but in different directories, using the path of the files as keys will prevent overriding the file content in the universe map since they will have different keys.

Example: 
plugin chain = plugin-1/v1 -> plugin-2/v1
plugin-1 scaffolds `dir-1/doc.go` and `dir-2/doc.go`
by using the file path as map key, we make sure that both entries for `dir-1/doc.go` and `dir-2/doc.go` will be available in the universe map, because the keys are unique, as opposed to if we use something like file name as the key, we will end up overriding one of the entries.
so in `plugin-2/v1`, we can for example access and make changes to any of the scaffolds the same way we would access them from file system: i.e. `pr.Universe["dir-1/doc.go"]`

pkg/plugins/external/helpers.go
em-r: @camilamacedo86 I've added a unit test in [fd9cfa](https://github.com/kubernetes-sigs/kubebuilder/pull/3223/commits/fd9cfa14aecf13de66bf8000af21435c2e461282), let me know if it needs to be adjusted or to add more test cases. 
thank you!

pkg/plugins/external/helpers.go
camilamacedo86: Thank you for the explanation.

camilamacedo86:
Hi @em-r,

That shows great for me ! 

@everettraven @rashmigottipati wdyt? Could you please give a hand in the review? 
Are you ok with? Can we have your LGTM here?

pkg/plugins/external/external_test.go
```diff
@@ -679,6 +679,54 @@ var _ = Describe("Run external plugin using Scaffold", func() {
...
+				{
+					path:    "dir/subdir",
+					name:    "file",
+					content: "level 2 file",
+				},
+			}
+
+			// create files in Filesystem
+			for _, file := range files {
+				fs.FS.MkdirAll(file.path, 0o700)
```
camilamacedo86, line 713: @em-r,

We will need to sort out the lint issues to move forward.
Could you please address them out?

em-r: Hi @camilamacedo86, the linting issues are now fixed. thank you!

camilamacedo86 (approved):
That seems great to me !!

/approved

camilamacedo86:
/lgtm

@rashmigottipati @everettraven we are moving forward with this one, but please check if it requires changes we can do a follow-up pr to address them.

Issue #3152, opened 2023-01-10
Installation Docs is outdated. 


Sajiyah-Salat:
### What do you want to happen?

https://book-v1.book.kubebuilder.io/getting_started/installation_and_setup.html 
I am new to kubebuilder and I dont get any particular brief guide about kubebuilder install. The latest stable version that is given in docs is 1.x and from github repo we can see that it 3.x . I would like to work on this issue. I want some help from contributors for installing kubebuilder. Thank you.

### Extra Labels

/kind documentation, /kind cleanup

camilamacedo86:
Hi @Sajiyah-Salat,

The link `https://book-v1.book.kubebuilder.io/getting_started/installation_and_setup.html` regards the kubebuilder v1 (very old version). Where did you find that? 

You can check in the readme and contributing guide that the link that you should use is: https://book.kubebuilder.io/

Sajiyah-Salat:
I just searched kubebuilder install Ubuntu and got this as my first link.

Sajiyah-Salat:
I think because this is old SEO ranks it first and I think many beginners would be confused can we add a link of new page on old page.

Sajiyah-Salat:
hey camilla, can I work on this?

camilamacedo86:
Hi @Sajiyah-Salat,

Feel free to assign any to yourself any issue and push a pull request with the change.
Your help is more than welcome !!! Feel free to ping me if you need help with reviews.

Sajiyah-Salat:
I would like to know where the old page live in kubebuilder repo.

camilamacedo86:
It is described here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#how-to-contribute-to-docs

You need to do a PR against the old book branches.

camilamacedo86:
I am closing this one because we can check that it is properly done:

<img width="1257" alt="Screenshot 2023-02-14 at 06 45 19" src="https://user-images.githubusercontent.com/7708031/218659899-3e60f600-d351-4cea-826c-04059e42efbd.png">

Sajiyah-Salat:
It was not there at first. Who added it?

camilamacedo86:
Hi @Sajiyah-Salat,

> It was not there at first. Who added it?

You can look for the change in the code, use the blame and check when/who added it. 
However, it seems added prior you got this task.

It is a good idea prior try to solve check if the problem exist yet. 
[...]

PR #3225, opened 2023-02-13
Add new book link to legacy docs

Sajiyah-Salat:
Issue: #3152 
When I was a beginner and searched for kubebuilder installation guide. I got v1 page and in that process I wasted two days which ended in frustration and then I took help from kubebuilder slack channel. Any beginner starting their kubernetes operator journey should not face this problem. So I added v3 link in legacy docs.
- 📖 (:book:): documentation


Sajiyah-Salat:
Hello @camilamacedo86 please review this pr and let me know if I had to add the same link in all legacy docs. Thank you.

camilamacedo86:
In this case we should to do as we do for v2 docs see: https://book-v2.book.kubebuilder.io/

camilamacedo86:
I am closing this one because we can check that it is properly done:

Screenshot 2023-02-14 at 06 45 19

PR #3231, opened 2023-02-17
🌱   Update cronjob-tutorial automatically

Eileen-Yu:
## Description:
Automatically update the cronjob tutorial.

## Motivation:
1. Keep tutorial align with the latest kb.
2. Solve the second part of https://github.com/kubernetes-sigs/kubebuilder/issues/2510


docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller.go
```diff
@@ -133,19 +132,14 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 		<aside class="note">
-
 		<h1>What is this index about?</h1>
-
 		<p>The reconciler fetches all jobs owned by the cronjob for the status. As our number of cronjobs increases,
 		looking these up can become quite slow as we have to filter through all of them. For a more efficient lookup,
 		these jobs will be indexed locally on the controller's name. A jobOwnerKey field is added to the
 		cached job objects. This key references the owning controller and functions as the index. Later in this
 		document we will configure the manager to actually index this field.</p>
-
```
camilamacedo86, line 146: We are removing the spaces.
I think that it happens because we are not using a const to inject all those contents

I would suggest we have conts like for example: 

const doc_fragment_x = `

<p>The reconciler fetches all jobs owned by the cronjob for the status. As our number of cronjobs increases,
		looking these up can become quite slow as we have to filter through all of them. For a more efficient lookup,
		these jobs will be indexed locally on the controller's name. A jobOwnerKey field is added to the
		cached job objects. This key references the owning controller and functions as the index. Later in this
		document we will configure the manager to actually index this field.</p>

`

hack/docs/internal/cronjob-tutorial/controller-implementation.go
```diff
@@ -0,0 +1,528 @@
+package cronjob
+
+
+const ControllerIntro = `
```
camilamacedo86, line 4: Do we need to export this ones? `ControllerIntro`

Could we have like `controllerCommentFragment `

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_types.go
```diff
@@ -29,7 +31,6 @@ import (
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
-
```
camilamacedo86, line 32: @Eileen-Yu we are missing the spaces

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
```diff
@@ -102,8 +101,8 @@ var _ = Describe("CronJob controller", func() {
 				After creating this CronJob, let's check that the CronJob's Spec fields match what we passed in.
 				Note that, because the k8s apiserver may not have finished creating a CronJob after our `Create()` call from earlier, we will use Gomega’s Eventually() testing function instead of Expect() to give the apiserver an opportunity to finish creating our CronJob.
 
-				`Eventually()` will repeatedly run the function provided as an argument every interval seconds until
-				(a) the function’s output matches what’s expected in the subsequent `Should()` call, or
+    `Eventually()` will repeatedly run the function provided as an argument every interval seconds until
```
camilamacedo86, line 104: Space in the const for we to have the same result.

docs/book/src/cronjob-tutorial/testdata/project/main.go
```diff
@@ -106,7 +107,6 @@ func main() {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
-
```
camilamacedo86, line 109: We are missing spaces

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go
camilamacedo86: It sometimes shows to have the same space but has not.
What I do is copy and past from another line that is working and adding on front to ensure that has the same.

hack/docs/internal/cronjob-tutorial/controller-implementation.go
```diff
@@ -0,0 +1,528 @@
...
+// +kubebuilder:docs-gen:collapse=Apache License
+
```
camilamacedo86, line 6: Nit, could we have an standard for the files name?
Can we use _ instead of - in all of them?
Also, could we not have the sufix test since it is a reserved in golang for the tests ?

hack/docs/internal/cronjob-tutorial/api-design.go
```diff
@@ -0,0 +1,129 @@
+package cronjob
```
camilamacedo86, line 1: @Eileen-Yu could we have here something like:

`hack/docs/internal/cronjob-tutorial/constants` and then a file name for each part like `controller.go`, `controller_test_code`, `main.go` and etc?

camilamacedo86, line 1: Is the package right?

hack/docs/internal/cronjob-tutorial/controller-implementation.go
Eileen-Yu: Thx for pointing out 👍

docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller.go
Eileen-Yu: It's my bad. I checked and found it is because I missed spaces between lines in a quite long const.
Now returned all spaces.

hack/docs/internal/cronjob-tutorial/controller-implementation.go
Eileen-Yu: Do you recommend I put those fragments in `generate_cronjob.go`?

Currently I split tutorial contents(mainly words) of each book page and put them in different files. Wish this can help easier maintenance in the future.

If we want to keep them in the same file, I can also move them back.

camilamacedo86 (approved):
That is amazing !!!

/lgtm

camilamacedo86:
/hold

It is failing in the testdata.

Kavinjsir:
It looks like the CI pipeline failed by missing `kubebuilder`:
https://github.com/kubernetes-sigs/kubebuilder/actions/runs/4401403491/jobs/7707582979#step:7:18

One quick fix might be modifying the path to be `/tmp/kubebuilder/bin/kubebuilder`, which is referred by:
https://github.com/kubernetes-sigs/kubebuilder/blob/0ba283012a55aa9a71991b3f8d918e9560b814c1/hack/docs/generate_samples.go#L37

However, I personally don't recommend doing so, as this would make `check.sh` highly rely on a previous make target `check-testdata` in the pipeline. It works in the CI, however, it would make an individual `make check-docs` fail  locally.

I think we should also have `kubebuilder` prepare for `make check-docs` only. As a contributor, I think it is necessary to run this target and verify the consistency between docs and features before creating a PR.

@camilamacedo86 @Eileen-Yu How would you think of this?

Eileen-Yu:
> However, I personally don't recommend doing so, as this would make `check.sh` highly rely on a previous make target `check-testdata` in the pipeline. It works in the CI, however, it would make an individual `make check-docs` fail locally.

> I think we should also have `kubebuilder` prepare for `make check-docs` only. As a contributor, I think it is necessary to run this target and verify the consistency between docs and features before creating a PR.

Yes, I passed `make check-docs` in local dev while it failed in the pipeline.
 
So I agree that we may replace "/tmp/kubebuilder/bin/kubebuilder"  with `util.KubebuilderBinName`, and build kb first in the pipeline.
Hope this helps keep a consistent binaryName, and make it possible for user to get the same check result locally as in the pipeline.

camilamacedo86 (approved):
/hold cancel

/ok-to-test

/approved

/lgtm

camilamacedo86:
Hi @Eileen-Yu,

It still with some error see the job `pull-kubebuilder-test`:

```
/home/prow/go/src/sigs.k8s.io/kubebuilder/docs/book/src/cronjob-tutorial/testdata/project/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
cmd/main.go:36:2: found packages controller (cronjob_controller.go) and controllers (cronjob_controller_test.go) in /home/prow/go/src/sigs.k8s.io/kubebuilder/docs/book/src/cronjob-tutorial/testdata/project/internal/controller
Error: not all generators ran successfully
```

hack/docs/check.sh
```diff
@@ -14,6 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+source "$(dirname "$0")/../../test/common.sh"
```
camilamacedo86, line 17: Why do we need to add this one here?

hack/docs/check.sh
```diff
@@ -14,6 +14,9 @@
...
+
+build_kb
```
camilamacedo86, line 19: Also, why are we adding it here?

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -54,7 +54,7 @@ This marker is responsible for generating a mutating webhook manifest.
 The meaning of each marker can be found [here](/reference/markers/webhook.md).
 */
 
-//+kubebuilder:webhook:path=/mutate-batch-tutorial-kubebuilder-io-v1-cronjob,mutating=true,failurePolicy=fail,sideEffects=None,groups=batch.tutorial.kubebuilder.io,resources=cronjobs,verbs=create;update,versions=v1,name=mcronjob.kb.io,admissionReviewVersions=v1
+//+kubebuilder:webhook:path=/mutate-batch-tutorial-kubebuilder-io-v1-cronjob,mutating=true,failurePolicy=fail,groups=batch.tutorial.kubebuilder.io,resources=cronjobs,verbs=create;update,versions=v1,name=mcronjob.kb.io,sideEffects=None,admissionReviewVersions=v1
```
camilamacedo86, line 57: @Eileen-Yu,
 
This change is the root cause of the issue.
Can we ensure that it is == before?

hack/docs/check.sh
Eileen-Yu: Hi @camilamacedo86 

It seems we need to prepare a kb for the pipeline based on the above discussion: https://github.com/kubernetes-sigs/kubebuilder/pull/3231#issuecomment-1465510380

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
Eileen-Yu: Have updated the code to fix the pipeline.

Issue #3117, opened 2022-12-06
Make go/v4-alpha and kustomize/v2-alpha to be stable and the default CLI scaffold


camilamacedo86:
### What do you want to happen?

The following is what needs to get done:

- [x] : check if we can move forward and get this one merged: https://github.com/kubernetes-sigs/kubebuilder/pull/2985
- [x] : check if we can close as deferred: https://github.com/kubernetes-sigs/kubebuilder/issues/615 (convey added on the issue)
- [x] : release as alpha. Then, ask for the community to check it out (if significant issues are found, we need to release it as beta)
- [x] : make the plugins stable and update the docs to no more extended reference it as alpha https://github.com/kubernetes-sigs/kubebuilder/pull/3237 and https://github.com/kubernetes-sigs/kubebuilder/pull/3234
- [x] : make the plugins stable and make them the default CLI scaffold https://github.com/kubernetes-sigs/kubebuilder/pull/3278
- [x] : update readme and quick start since now we will support darwin/arm64 by default, and we do not need to say for people use go/v4-alpha anymore
- [x] : deprecated go/v3 and kustomize/v1 - https://github.com/kubernetes-sigs/kubebuilder/pull/3235 and https://github.com/kubernetes-sigs/kubebuilder/pull/3238
- [x] : deprecated packages to let consumers informed https://github.com/kubernetes-sigs/kubebuilder/pull/3284
- [x] : update tutorials to use go/v4 https://github.com/kubernetes-sigs/kubebuilder/pull/3280, https://github.com/kubernetes-sigs/kubebuilder/pull/3281, https://github.com/kubernetes-sigs/kubebuilder/pull/3282

joejulian:
There are operators that have been built using upstream controllers. Moving these controllers to internal breaks downstream projects.

PR #3234, opened 2023-02-19
:warning: (kustomize/v2) Stabilize plugin. kustomize/v2-alpha was replaced with kustomize/v2

camilamacedo86:
## Description

Stabilize plugin. 
kustomize/v2-alpha was replaced by kustomize/v2

Motivation: https://github.com/kubernetes-sigs/kubebuilder/issues/3117


camilamacedo86:
@everettraven and @varshaprasad96 I will move forward with this one because it is blocking the others. However, we can do follow ups if we check that something is missing.

camilamacedo86:
@varshaprasad96 and @everettraven I am moving this one because it is a blocker to address the comments: https://github.com/kubernetes-sigs/kubebuilder/pull/3237#discussion_r1124865375 

Also, if we need to do changes please let me know we can do in a follow up.

PR #3246, opened 2023-02-21
[infra] Add cloudbuild

varshaprasad96:

### DIFF
--- a/cloudbuild/cloudbuild_kube-rbac-proxy.yaml
+++ b/cloudbuild/cloudbuild_kube-rbac-proxy.yaml
@@ -1,0 +1,24 @@
+#  Copyright 2021 The Kubernetes Authors.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


cloudbuild/cloudbuild_kube-rbac-proxy.yaml
```diff
@@ -0,0 +1,24 @@
...
+substitutions:
+  # This is the kube-rbac-proxy version, source image tags for which must exist remotely.
+  _KUBE_RBAC_PROXY_VERSION: v0.13.1
+steps:
+- name: "gcr.io/cloud-builders/docker"
+  env:
+  - "KUBE_RBAC_PROXY_VERSION=${_KUBE_RBAC_PROXY_VERSION}"
+  entrypoint: "/usr/bin/env"
+  args: ["bash", "-c", "./build/build.sh"]
+images: ["gcr.io/kubebuilder/kube-rbac-proxy:${_KUBE_RBAC_PROXY_VERSION}"]
```
camilamacedo86, line 24: The registry in this case should not be the same.

camilamacedo86:
@varshaprasad96,

We moved forward by duplicating the directories and have one for each scenario so that we do not broke the current state.
See: https://github.com/kubernetes-sigs/kubebuilder/pull/3362

However, we still need to check if that is working there and what would be the next steps.

So, I am closing this one but if you see that we still needing to do address anything else and etc, please feel free to re-open or open a follow up.

Issue #3224, opened 2023-02-13
External plugins discovery fails when XDG_CONFIG_HOME is set


em-r:
### What broke? What's expected?

On Ubuntu, when specifying an external plugin to any subcommand (e.g. `init`) through the `--plugins` flag, the subcommand fails if `XDG_CONFIG_HOME` is set to its default value `$HOME/.config`, with the following error:

`2023/02/11 15:35:25 no plugin could be resolved with key "myexternalplugin/v1"`

The subcommands eventually run successfully after unsetting `XDG_CONFIG_HOME`.

### Reproducing this issue

1. Set XDG_CONFIG_HOME variable in the shell: `export XDG_CONFIG_HOME=$HOME/.config`
2. Create a folder for an external plugin: `mkdir -p $XDG_CONFIG_HOME/kubebuilder/plugins/myexternalplugin/v1`
3. Copy the plugin binary to the directory created in 2).
4. Initialize a new project: `kubebuilder init --plugins=myexternalplugin/v1`
5. Notice how the execution will exit with an error, and `myexternalplugin` won't appear in the available plugins table.

### KubeBuilder (CLI) Version

3.9.0

### Other versions

Go version: 1.19.4 linux/amd64
OS version: Ubuntu 22.04.1 LTS

em-r:
Looking at the code, I believe this issue occurs because Kubebuilder concatenates the outcome of [`getHomeDir`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/options.go#L291) (which maybe $XDG_CONFIG_HOME) to `.config/kubebuilder` (see  [options.go#L191](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/options.go#L191))

So, when `XDG_CONFIG_HOME` is set, Kubebuilder will try to find external plugins in the directory `$XDG_CONFIG_HOME/.config/kubebuilder`.
For example, if `XDG_CONFIG_HOME` is set to its default value (`$HOME`), then Kubebuilder will lookup external plugins in `$HOME/.config/.config/kubebuilder`.

em-r:
I'll be happy to send a PR with a fix if this gets accepted.

camilamacedo86:
Hi @em-r,

All issues that you find if you either be able to contribute with the fix that is amazing !!! please feel free to move forward  !!!

PR #3247, opened 2023-02-23
🐛 (API - External Plugins):  Fix external plugin discovery on Linux

em-r:
This PR aims to fix a bug that's causing external plugin discovery to fail in Linux systems when `XDG_CONFIG_HOME` environment variable is set.

fixes: #3224

### DIFF
--- a/pkg/cli/options.go
+++ b/pkg/cli/options.go
@@ -171,26 +171,44 @@ func parseExternalPluginArgs() (args []string) {
 	return args
 }
 
+// isHostSupported checks whether the host system is supported or not.
+func isHostSupported(host string) bool {
+	for _, platform := range supportedPlatforms {
+		if host == platform {
+			return true
+		}
+	}
+	return false
+}
+
 // getPluginsRoot detects the host system and gets the plugins root based on the host.
 func getPluginsRoot(host string) (pluginsRoot string, err error) {
+	if !isHostSupported(host) {
+		// freebsd, openbsd, windows...
+		return "", fmt.Errorf("host not supported: %v", host)
+	}
+
+	pluginsRelativePath := filepath.Join("kubebuilder", "plugins")
+	if xdgHome := os.Getenv("XDG_CONFIG_HOME"); xdgHome != "" {
+		return filepath.Join(xdgHome, pluginsRelativePath), nil
+	}
+
 	switch host {
 	case "darwin":
 		logrus.Debugf("Detected host is macOS.")
-		pluginsRoot = filepath.Join("Library", "Application Support", "kubebuilder", "plugins")
+		pluginsRoot = filepath.Join("Library", "Application Support", pluginsRelativePath)
 	case "linux":
 		logrus.Debugf("Detected host is Linux.")
-		pluginsRoot = filepath.Join(".config", "kubebuilder", "plugins")
-	default:
-		// freebsd, openbsd, windows...
-		return "", fmt.Errorf("Host not supported: %v", host)
+		pluginsRoot = filepath.Join(".config", pluginsRelativePath)
 	}
-	userHomeDir, err := getHomeDir()
+
+	userHomeDir, err := os.UserHomeDir()
 	if err != nil {
 		return "", fmt.Errorf("error retrieving home dir: %v", err)
 	}
 	pluginsRoot = filepath.Join(userHomeDir, pluginsRoot)
 
-	return pluginsRoot, nil
+	return
 }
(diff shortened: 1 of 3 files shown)


pkg/cli/options.go
```diff
@@ -171,26 +171,44 @@ func parseExternalPluginArgs() (args []string) {
...
 	}
-	userHomeDir, err := getHomeDir()
+
+	userHomeDir, err := os.UserHomeDir()
 	if err != nil {
 		return "", fmt.Errorf("error retrieving home dir: %v", err)
 	}
 	pluginsRoot = filepath.Join(userHomeDir, pluginsRoot)
 
-	return pluginsRoot, nil
```
camilamacedo86, line 193: This method expects a return `(pluginsRoot string, err error) ` I think it is missing return it

camilamacedo86 (commented):
/ok-to-test

pkg/cli/options.go
```diff
@@ -171,26 +171,44 @@ func parseExternalPluginArgs() (args []string) {
 	return args
```
camilamacedo86, line 171: @evansheng @rashmigottipati could you please give a hand to review this one?

pkg/cli/options.go
em-r: the reason I have replaced `return pluginsRoot, nil` with just `return` is because the function uses named returns so I think specifying the variables in the return is redundant, but I can add it back (maybe because of a style convention?) just let me know.
thank you!

em-r:
I can see the coverage check failed, I'll add more tests to fix this as well.

em-r:
coverage drop has been fixed. I'm seeing a failure in k8s-1-24-7 build, however, based on its logs I don't think it's related to the changes in this PR, it seems like some sort of missing file issue in the machine where the build ran. will keep looking into it.
cc @camilamacedo86

everettraven (approved):
Thanks for this contribution @em-r , these changes look great! 

/lgtm

everettraven:
The test failure seems like some weird kustomize errors. I'm not sure off the top of my head what the problem here is

em-r:
thanks @everettraven for the review!
I'll try to look into the build errors over the weekend, and hopefully, figure out a way to resolve them. I'll keep you posted.

em-r:
thanks @camilamacedo86 for re-running the failing build.
seems to be passing now @everettraven

rashmigottipati (approved):

pkg/cli/options.go
```diff
@@ -171,26 +171,44 @@ func parseExternalPluginArgs() (args []string) {
...
+	for _, platform := range supportedPlatforms {
+		if host == platform {
+			return true
+		}
+	}
+	return false
+}
+
 // getPluginsRoot detects the host system and gets the plugins root based on the host.
 func getPluginsRoot(host string) (pluginsRoot string, err error) {
```
camilamacedo86, line 185: /hold
```suggestion
func getPluginsRoot(host string) {
```

Can we first remove the return since it is not used? 
https://github.com/kubernetes-sigs/kubebuilder/pull/3247/files#r1126146318

em-r: @camilamacedo86 this function is assigned to [`retrievePluginsRoot`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/options.go#L39), and then called by [`DiscoverExternalPlugins`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/options.go#L198). the output of the function is indeed used then to lookup external plugins in `pluginsRoot`.
to elaborate as to why I've changed `return pluginsRoot, nil` to just `return`: it's because the function uses named returns, so while using `return pluginsRoot, nil` as a return expression would work, specifying the variables would be redundant, so essentially, just using a "void" `return` will have the same outcome (because of the named return in the signature).
but either way, I can revert it back to the old style as it would work fine as well. what do you think?
thank you!

camilamacedo86 (approved):
/hold cancel

/lgtm

pkg/cli/options.go
camilamacedo86: I think that is fine as it is . We can re-check in a follow up either.
Thank you a lot for your help to address those ones.

rashmigottipati (approved):
/lgtm

pkg/cli/options.go
em-r: thank you @camilamacedo86, always happy to help!

Issue #3249, opened 2023-02-23
`make deploy` fails inside cronjob tutorial  


kallyaleksiev:
### What broke? What's expected?

Currently, inside the code for the cronjob tutorial (in `docs/book/src/cronjob-tutorial/testdata/project`) if you try to deploy it to your cluster using `make deploy` you get an error:

```
lstat kubebuilder/docs/book/src/cronjob-tutorial/testdata/project/config/manager/controller_manager_config.yaml: no such file or directory"

error: no objects passed to apply
make: *** [deploy] Error 1
```

because the file `controller_manager_config.yaml` referenced in the `kustomization.yaml` does not exist there 

this file does not seem to be needed and the entire `configMapGenerator` https://github.com/kubernetes-sigs/kubebuilder/blob/73e25fadb93daa1e240b4dda4df7c9016ef4d3ce/docs/book/src/multiversion-tutorial/testdata/project/config/manager/kustomization.yaml#L7 can be deleted, at which point `make deploy` works

### Reproducing this issue

run `make deploy` inside `docs/book/src/cronjob-tutorial/testdata/project`

### KubeBuilder (CLI) Version

3.9.0

### Extra Labels

/kind documentation

xiao-jay:
i agree @kallyaleksiev,i think should remove the code block below because the code does not seem to be needed and the entire configMapGenerator.I also create a project to test [cronjob tutorial](https://kubebuilder.io/cronjob-tutorial/controller-implementation.html),found not generate `configMapGenerator:`


[kubebuilder/docs/book/src/multiversion-tutorial/testdata/project/config/manager/kustomization.yaml](https://github.com/kubernetes-sigs/kubebuilder/blob/73e25fadb93daa1e240b4dda4df7c9016ef4d3ce/docs/book/src/multiversion-tutorial/testdata/project/config/manager/kustomization.yaml#L7)
```
generatorOptions:
  disableNameSuffixHash: true

configMapGenerator:
- name: manager-config
  files:
  - controller_manager_config.yaml
```


### test cronjob tutorial way
```
kubebuilder init --domain tutorial.kubebuilder.io --repo tutorial.kubebuilder.io/project
kubebuilder create api --group batch --version v1 --kind CronJob 
```

code in kubebuilder-cronjob/config/manager/kustomization.yaml
```
resources:
- manager.yaml
```


```
➜  manager kubebuilder version
[...]

PR #3256, opened 2023-03-04
📖 : delete cronjob tutorial Incorrect kustomization.yaml

xiao-jay:
#3249 
Delete cronjob tutorial Incorrect code in kustomization.yaml, that will lead to `make deploy` error

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/manager/kustomization.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/manager/kustomization.yaml
@@ -1,10 +1,3 @@
 resources:
 - manager.yaml
 
-generatorOptions:
-  disableNameSuffixHash: true
-
-configMapGenerator:
-- name: manager-config
-  files:
-  - controller_manager_config.yaml


camilamacedo86 (dismissed):
The right approach would be update the project.
It seems outdated. 

However, that is fine we can move forward with this only change.
Also, see that I am changing the title. The right emoji for this one is 📖

camilamacedo86 (approved):
/lgtm

everettraven (approved):
Looks reasonable to me.

/lgtm

everettraven:
These APIDiff check failures seem weird. I'm not sure why changing the docs would cause a failure in the diff

everettraven:
> These APIDiff check failures seem weird. I'm not sure why changing the docs would cause a failure in the diff

Actually just noticed these are all related to the kustomize/v2 plugin. This PR was created before the kustomize/v2 plugin was stabilized so it is in a weird state where running against the PR it is still kustomize/v2-alpha vs the main branch being kustomize/v2.

I think to fix this the PR just needs to be closed and then reopened

camilamacedo86:
Hi @everettraven,

https://github.com/kubernetes-sigs/kubebuilder/pull/3256#issuecomment-1458266775
It seems a bug with the go-apidiff.  I will force this one and we keep the eyes on it. Therefore, if we see more issues we need to further investigate.

PR #3270, opened 2023-03-08
:bug: [fix: go/v3, go/v4] Remove stale imports

varshaprasad96:
This PR removes the stale import: "pkg/envtest/printer" that is not needed after migrating to ginko/v2.

Controller-runtime v0.14.0+ has migrated to use ginko/v2 and thereby removed "printer" package. The imports were still being scaffolded though not used in the codebase.

This was causing the error mentioned [here](https://kubernetes.slack.com/archives/CAR30FCJZ/p1678204565778959) by one of the community users.


camilamacedo86 (approved):

em-r (approved):

rashmigottipati (commented):
/lgtm

PR #3275, opened 2023-03-11
🌱 : stop to generate all samples for go/v3

camilamacedo86:
## Description

The go/v3 has been deprecated. 
Therefore, we no longer need to generate the samples with all addons options for go/v3 and we can keep only project-v3 as a base so we are able to review and compare the scaffolds. 

## Motivation
Make easier to keep the project maintained


test/testdata/generate.sh
```diff
@@ -129,15 +129,10 @@ build_kb
...
-scaffold_test_project project-v3-with-grafana --plugins="go/v3"
-
-# [Next version, alpha] Project version v4-alpha
+# [Deprecated] - Project version 3 (default) uses plugin go/v3 (default).
+scaffold_test_project project-v3
+
+# [Currently, default CLI plugin] - [Next version, alpha] Project version v4-alpha
 scaffold_test_project project-v4 --plugins="go/v4"
 scaffold_test_project project-v4-multigroup --plugins="go/v4"
 scaffold_test_project project-v4-declarative-v1 --plugins="go/v4,declarative"
```
everettraven, line 138: Should we replicate the `-with-deploy-image` and `-with-grafana` samples for go/v4?

everettraven (approved):
Looks good to me, just one question:

test/testdata/generate.sh
everettraven: Forgot to expand the file to see that it already does :facepalm: - ignore this

PR #3276, opened 2023-03-11
:warning: 🐛 (API) fix deprecate interface to allow bundle plugin has deprecation message

camilamacedo86:
## Description

This PR enforces the deprecated interface to be implemented so that Bundle Plugins can also be deprecated.
If you are using Kubebuilder as lib, then:
- a) Add an empty string for all bundles plugins which are not deprecated:

Example:

```go
	// Bundle plugin which built the golang projects scaffold by Kubebuilder go/v4 with kustomize alpha-v2
	gov4Bundle, _ := plugin.NewBundle(golang.DefaultNameQualifier, plugin.Version{Number: 4},
		"", // Here you must pass an empty string. If you inform an string value here then that means this plugin is deprecated and this string will be used to create the notice message
		kustomizecommonv2alpha.Plugin{},
		golangv4.Plugin{},
	)
``` 

- b) Ensure that you implement the Deprecate interface in all plugins

```go
func (p Plugin) DeprecationWarning() string {
	return "" // It should be empty. if you inform a value then that means this plugin is deprecated. 
}

```

## Motivation 

Fix the interface to allow Bundles to be deprecated. example:

```go
	const deprecateMessageGoV3Bundle = "This version is deprecated." +
		"The `go/v3` cannot scaffold projects using kustomize versions v4x+" +
		" and cannot fully support Kubernetes 1.25+." +
		"It is recommended to upgrade your project to the latest versions available (go/v4)." +
		"Please, check the migration guide to learn how to upgrade your project"

	// Bundle plugin which built the golang projects scaffold by Kubebuilder go/v3
	gov3Bundle, _ := plugin.NewBundle(golang.DefaultNameQualifier, plugin.Version{Number: 3},
		deprecateMessageGoV3Bundle,
		kustomizecommonv1.Plugin{},
		golangv3.Plugin{},
	)
```
Screenshot 2023-03-11 at 11 38 16

**IMPORTANT** See that consumers can use it to know when they will perform the deprecation to define their own message for their own Bundle Plugins.

### DIFF
--- a/pkg/plugin/bundle.go
+++ b/pkg/plugin/bundle.go
@@ -28,11 +28,12 @@ type bundle struct {
 	plugins []Plugin
 
 	supportedProjectVersions []config.Version
+	deprecateWarning         string
 }
 
 // NewBundle creates a new Bundle with the provided name and version, and that wraps the provided plugins.
 // The list of supported project versions is computed from the provided plugins.
-func NewBundle(name string, version Version, plugins ...Plugin) (Bundle, error) {
+func NewBundle(name string, version Version, deprecateWarning string, plugins ...Plugin) (Bundle, error) {
 	supportedProjectVersions := CommonSupportedProjectVersions(plugins...)
 	if len(supportedProjectVersions) == 0 {
 		return nil, fmt.Errorf("in order to bundle plugins, they must all support at least one common project version")
@@ -55,6 +56,7 @@ func NewBundle(name string, version Version, plugins ...Plugin) (Bundle, error) 
 		version:                  version,
 		plugins:                  allPlugins,
 		supportedProjectVersions: supportedProjectVersions,
+		deprecateWarning:         deprecateWarning,
 	}, nil
 }
 
@@ -76,4 +78,9 @@ func (b bundle) SupportedProjectVersions() []config.Version {
 // Plugins implements Bundle
 func (b bundle) Plugins() []Plugin {
 	return b.plugins
+}
+
+// Plugins implements Bundle
+func (b bundle) DeprecationWarning() string {
+	return b.deprecateWarning
 }
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -39,13 +39,22 @@ import (
 
 func main() {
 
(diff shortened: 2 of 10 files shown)


pkg/plugin/bundle.go
```diff
@@ -28,11 +28,12 @@ type bundle struct {
 	plugins []Plugin
 
 	supportedProjectVersions []config.Version
+	deprecateWarning         string
 }
 
 // NewBundle creates a new Bundle with the provided name and version, and that wraps the provided plugins.
 // The list of supported project versions is computed from the provided plugins.
-func NewBundle(name string, version Version, plugins ...Plugin) (Bundle, error) {
+func NewBundle(name string, version Version, deprecateWarning string, plugins ...Plugin) (Bundle, error) {
```
everettraven, line 36: I wonder if converting this to use the options pattern would be better than adding another function parameter? I think that would prevent this function signature getting really long with future changes.

I think this would look something like:
```go
func NewBundle(opts ...BundleOptions) (Bundle, error)
```
and new bundles could be created like:
```go
bundle, err := NewBundle(
  WithName("name"),
  WithVersion(plugin.Version{Number: 4}),
  WithPlugins(plugins ...Plugin),
  WithDeprecationMessage("Deprecated!"),
)
```

Thoughts?

everettraven (approved):
Overall, this looks good to me - just have one question:

pkg/plugin/bundle.go
camilamacedo86: I liked the idea a lot. 
What about we do that in a follow-up?

varshaprasad96 (approved):

pkg/plugin/bundle.go
camilamacedo86: @everettraven here we just fixing the method (which breaks the API)
I totally agree with the approach suggested by you and for we are able to do that I tracked the need in: https://github.com/kubernetes-sigs/kubebuilder/issues/3307

camilamacedo86:
I am moving this one since it has both approvals.

PR #3277, opened 2023-03-11
🐛  fix: deprecate message should be in go/v3

camilamacedo86:
## Description

The message was added to the wrong plugin version
This PR only fix and ensure that the message is the right one.


camilamacedo86:
I am moving forward with this one because it is only a small / nit fix.

PR #3279, opened 2023-03-11
🌱 : fix sample project-v4-grafana

camilamacedo86:
## Description

Fix grafana sample v4 to use go/v4 plugin


camilamacedo86:
I am moving with this one because it is only to fix the sample.

Issue #3307, opened 2023-03-20
Deprecate the current method API to build bundles and create new version to work with Options


camilamacedo86:
### What do you want to happen?

Currently, we have the following: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/bundle.go#L33-L59
Therefore, this task is for us to deprecate the func used currently in favour of something like:
```go
func NewBundle(opts ...BundleOptions) (Bundle, error)
```
and new bundles could be created like:
```go
bundle, err := NewBundle(
  WithName("name"),
  WithVersion(plugin.Version{Number: 4}),
  WithPlugins(plugins ...Plugin),
  WithDeprecationMessage("Deprecated!"),
)
```

requester @everettraven see: https://github.com/kubernetes-sigs/kubebuilder/pull/3276#discussion_r1134064332

NikhilSharmaWe:
I would like to work on it.

PR #3312, opened 2023-03-23
✨Change `NewBundle` method to build bundles through working with options

NikhilSharmaWe:
### Description
This PR changes the [NewBundle](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/bundle.go#L33-L59) method to work with options. The pattern will like:
```
func NewBundle(opts ...BundleOptions) (Bundle, error)
```
and new bundles will be created like:
```
bundle, err := NewBundle(
  WithName("name"),
  WithVersion(plugin.Version{Number: 4}),
  WithPlugins(plugins ...Plugin),
  WithDeprecationMessage("Deprecated!"),
)
```

### Motivation
Fixes: #3307

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -46,17 +46,16 @@ func main() {
 		"Please, check the migration guide to learn how to upgrade your project"
 
 	// Bundle plugin which built the golang projects scaffold by Kubebuilder go/v3
-	gov3Bundle, _ := plugin.NewBundle(golang.DefaultNameQualifier, plugin.Version{Number: 3},
-		deprecateMessageGoV3Bundle,
-		kustomizecommonv1.Plugin{},
-		golangv3.Plugin{},
+	gov3Bundle, _ := plugin.NewBundleWithOptions(plugin.WithName(golang.DefaultNameQualifier),
+		plugin.WithVersion(plugin.Version{Number: 3}),
+		plugin.WithDeprecationMessage(deprecateMessageGoV3Bundle),
+		plugin.WithPlugins(kustomizecommonv1.Plugin{}, golangv3.Plugin{}),
 	)
 
 	// Bundle plugin which built the golang projects scaffold by Kubebuilder go/v4 with kustomize alpha-v2
-	gov4Bundle, _ := plugin.NewBundle(golang.DefaultNameQualifier, plugin.Version{Number: 4},
-		"",
-		kustomizecommonv2alpha.Plugin{},
-		golangv4.Plugin{},
+	gov4Bundle, _ := plugin.NewBundleWithOptions(plugin.WithName(golang.DefaultNameQualifier),
+		plugin.WithVersion(plugin.Version{Number: 4}),
+		plugin.WithPlugins(kustomizecommonv2alpha.Plugin{}, golangv4.Plugin{}),
 	)
 
 	fs := machinery.Filesystem{
--- a/pkg/plugin/bundle.go
+++ b/pkg/plugin/bundle.go
@@ -31,6 +31,33 @@ type bundle struct {
 	deprecateWarning         string
 }
 
+type BundleOption func(*bundle)
+
+func WithName(name string) BundleOption {
+	return func(opts *bundle) {
+		opts.name = name
+	}
+}
+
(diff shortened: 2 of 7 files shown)


cmd/main.go
```diff
@@ -46,17 +46,17 @@ func main() {
...
 	)
 
 	// Bundle plugin which built the golang projects scaffold by Kubebuilder go/v4 with kustomize alpha-v2
-	gov4Bundle, _ := plugin.NewBundle(golang.DefaultNameQualifier, plugin.Version{Number: 4},
-		"",
-		kustomizecommonv2alpha.Plugin{},
-		golangv4.Plugin{},
+	gov4Bundle, _ := plugin.NewBundle(plugin.WithName(golang.DefaultNameQualifier),
+		plugin.WithVersion(plugin.Version{Number: 4}),
+		plugin.WithDeprecationMessage(""),
```
em-r, line 58: I think this can be skipped since no deprecation message needs to show up when using go/v4

pkg/plugin/bundle.go
```diff
@@ -31,19 +31,52 @@ type bundle struct {
...
-func NewBundle(name string, version Version, deprecateWarning string, plugins ...Plugin) (Bundle, error) {
-	supportedProjectVersions := CommonSupportedProjectVersions(plugins...)
+func NewBundle(opts ...BundleOption) (Bundle, error) {
+	bundleOpts := bundle{}
+
+	for _, opts := range opts {
+		opts(&bundleOpts)
+	}
+
+	supportedProjectVersions := CommonSupportedProjectVersions(bundleOpts.plugins...)
```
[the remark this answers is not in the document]
NikhilSharmaWe: Yes, that seems reasonable. But in NewBundle func we also need to return an error when [`len(supportedProjectVersions) == 0`](https://github.com/NikhilSharmaWe/kubebuilder/blob/upgradeBundle/pkg/plugin/bundle.go#L71). 

Let's see what others think about this.

camilamacedo86 (commented):
@everettraven  could you please give a hand on the review of this one?

camilamacedo86 (commented):
Hi @NikhilSharmaWe, 

I loved the implementation. However, it is a breaking change in the API could we keep the current func implemented and deprecated it and then just add the new one using the Options? So that we do not affect those that are consuming Kubebuilder as a LIB?

pkg/plugin/bundle_test.go
```diff
@@ -67,7 +67,7 @@ var _ = Describe("Bundle", func() {
 				{p1, p2, p3},
 				{p1, p3, p4},
 			} {
-				b, err := NewBundle(name, version, "", plugins...)
```
camilamacedo86, line 70: Can we live it as it is and add a new test for a new function using the Options?

pkg/plugin/bundle.go
```diff
@@ -31,19 +31,52 @@ type bundle struct {
...
+		opts.deprecateWarning = msg
+	}
+
+}
+
 // NewBundle creates a new Bundle with the provided name and version, and that wraps the provided plugins.
 // The list of supported project versions is computed from the provided plugins.
-func NewBundle(name string, version Version, deprecateWarning string, plugins ...Plugin) (Bundle, error) {
-	supportedProjectVersions := CommonSupportedProjectVersions(plugins...)
+func NewBundle(opts ...BundleOption) (Bundle, error) {
```
camilamacedo86, line 63: Could we keep here the old func as it is and add a new func using the Options so that we do not introduce a breaking change? See that the PR is fafling in the API diff check.

NikhilSharmaWe: Should revert the changes in the docs too, right?

pkg/plugin/bundle.go
```diff
@@ -31,6 +31,33 @@ type bundle struct {
 	deprecateWarning         string
 }
```
camilamacedo86, line 33: The changes in Kubebuilder CLI can still.
I mean, we can use the method here .
We just need to have the old one available for those that import and use Kubebuilder as a LIB.

camilamacedo86 (approved):
/lgtm
/approved

PR #3344, opened 2023-04-21
:book: fix migration guide after go/v4 stabilization and links

camilamacedo86:
## Description

After the release by checking the migration guide, we see that the links are broken and that the note/warning about go/v4-alpha not being stable is still present. This PR just address those nits.


camilamacedo86:
Since it is only nits, the effort to wait for additional reviews is not valuable.
I am moving forward to fix the prod docs asap since it is an important doc done for the latest release.

PR #3346, opened 2023-04-21
:book: fix kustomize-v2 index link

camilamacedo86:
## Description

Fix the kustomize v2 links on the left side.
Update info accordingly after plugin updates


camilamacedo86 (commented):
We are moving forward to fix the prod docs. Not a valuable effort to ask for a double review on this one since it is only nits.
We can also supplement or change it in follow-ups

PR #3347, opened 2023-04-21
:book: Update release-3.10 branch with doc fixes addressed on master

camilamacedo86:
## Description

We did a release 3.10.0, and we could check that the docs have some issues regards the latest changes.
This PR address the fixes done against the master branch so that we can fix it in the book-v3.


camilamacedo86:
Moving forward, since it is only docs fixes then, we can fix the prod docs asap.

PR #3349, opened 2023-04-21
:book: Update book-v3 with latest docs fixes and release-3.10 branch with conflicts solved

camilamacedo86:
## Description

We addressed some docs fixes into master, which was synced with release-3.10
So that we can merge release-3.10 to book-v3 to update the prod docs and solve the conflicts. 

See that is the release-3.10 branch with the solved conflicts with book-v3 :

Screenshot 2023-04-21 at 10 07 00


camilamacedo86:
The errors with the CLA are a bug since we can check that those commits were merged previously against a master branch and successfully passed in the CLA check. I raised a ticket tracking the bug https://jira.linuxfoundation.org/plugins/servlet/desk/portal/4/SUPPORT-17860 so that we can move forward and not be blocked within.

camilamacedo86:
Moving forward, we will be able to fix the book-v3/prod docs asap since the docs fixed are those most required to be checked for the changes addressed in the latest release, 3.10.0. Therefore, it is good we have it fixed ASAP not to affect the end-users.

PR #3351, opened 2023-04-21
🌱added tests for exec.go

### LINKED ISSUE #3211: Increase the test coverage with unit tests

By looking at the coveralls reports we can see that we have some implementations which could be better covered. We do not cover 100% of the code but would be great increase the coverage and ensure that we have no red and orange files/dirs, see: https://coveralls.io/github/kubernetes-sigs/kubebuilder?branch=master

Example: 

We could add more tests for cmd_helpers.go, completiton.go and version.go:

Screenshot 2023-02-08 at 14 17 13

### Extra Labels

_No response_

khareyash05:
added unit tests for `pkg/plugin/exec.go`

Fixes #3211

### DIFF
--- a/pkg/plugin/util/exec_test.go
+++ b/pkg/plugin/util/exec_test.go
@@ -2,49 +2,37 @@ package util
 
 import (
 	"bytes"
+	"os"
 	"os/exec"
 	"strings"
-	"testing"
-	. "github.com/onsi/ginkgo/v2"
-	. "github.com/onsi/gomega"
+
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
 )
 
-func TestRunCmd(t *testing.T) {
-	testCases := []struct {
-		name        string
-		msg         string
-		cmd         string
-		args        []string
-		expectedErr error
-	}{
-		{
-			name:        "Success",
-			msg:         "Running command",
-			cmd:         "echo",
-			args:        []string{"hello"},
-			expectedErr: nil,
-		},
-		{
-			name:        "Error",
-			msg:         "Running command",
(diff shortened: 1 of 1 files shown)


camilamacedo86 (commented):
/ok-to-test

You MUST assign to the CLA otherwise we cannot get this one merged.
Also, see that this change does not affected end-users therefore, the right emoji is `🌱`. Could you please change the title?

pkg/plugin/util/exec_test.go
```diff
@@ -0,0 +1,48 @@
+package util
+
+import (
+	"bytes"
+	"os/exec"
+	"strings"
+	"testing"
+)
+
+func TestRunCmd(t *testing.T) {
```
camilamacedo86, line 10: See that we use the following libs to do the tests. Could you please follow the same standard to do this one?

> 	. "github.com/onsi/ginkgo/v2"
> 	. "github.com/onsi/gomega"

Example:
https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/util/helpers_test.go

khareyash05: Sure will do that

khareyash05:
@camilamacedo86 Added imports to the file

khareyash05:
@camilamacedo86 Could you please help me with the failing test. I am new to this repo and getting no idea from the build info

camilamacedo86:
Hi @khareyash05 ,

>@camilamacedo86 Added imports to the file

Thank you for the contribution. See that it is not only about add the imports but indeed use the libs and do the tests in the same format, approach that we have been doing in the project. Could you please check the other tests implemented and do it at the same away?

PR #3353, opened 2023-04-22
:bug: (go/v4) fix makefile target to build images ensuring its single responsibility

camilamacedo86:
## Description

fix makefile target to build images ensuring its single responsibility

## Motivation

When we call make build we want only to build the image and run the tests is not a pre-requirement. It can make take too long to build the image.

### DIFF
--- a/docs/book/src/component-config-tutorial/testdata/project/Makefile
+++ b/docs/book/src/component-config-tutorial/testdata/project/Makefile
@@ -78,7 +78,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # (i.e. docker build --platform linux/arm64 ). However, you must enable docker buildKit for it.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
-docker-build: test ## Build docker image with the manager.
+docker-build: ## Build docker image with the manager.
 	$(CONTAINER_TOOL) build -t ${IMG} .
 
 .PHONY: docker-push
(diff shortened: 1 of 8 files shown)


docs/book/src/component-config-tutorial/testdata/project/Makefile
```diff
@@ -72,7 +72,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # (i.e. docker build --platform linux/arm64 ). However, you must enable docker buildKit for it.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
-docker-build: test ## Build docker image with the manager.
+docker-build: ## Build docker image with the manager.
```
varshaprasad96, line 75: @camilamacedo86 Was there any specific reason to remove the `test` target apart from it taking long time. I'm still inclined towards us keeping it, as if the test fails the user would know the issue before proceeding to building the image. Imo I don't see any harm in having `test` target before building image.

camilamacedo86: Yes. Following scenarios where does not respect single responsibility one this brings bad side effect: 

If you use a tool to create a project and want to either release the project or use it for development purposes, you may encounter a situation where you need to wait for all tests to be complete before building the image. This can be frustrating if you have many tests and only want to build the image.

In addition, when you perform Continuous Integration (CI) to test your project, you may need to run all tests again before building the image if you have an end-to-end (e2e) test. This process can significantly increase the time it takes for your CI job to complete and force you to repeat all tests when you only need to build an image.

Why do I need to run the tests before building the image?

camilamacedo86:
@varshaprasad96 thank you for the review, can we merge this one now?

Kavinjsir (approved):
I'm cool with either side. To me, it feels like this is something that depends on the user side.
(And I personally prefer making the content of a target more focused on itself)

varshaprasad96 (approved):
/lgtm
/approve

PR #3354, opened 2023-04-22
🌱 cleanup tests after go/v3 deprecation

camilamacedo86:
## Description

This PR ensures that go/v3 will be tested with golang 1.19 and not upper.

Go/v3 is deprecated so we will keep testing it, however, in order to reduce the time of the CI, allow us to move forward and adopt Golang 1.20 only for go/v4 and make it easier we keep the project maintained we are also removing some tests on prow using go/v3. Ideally, we should no longer change the go/v3 plugin and new features and fixes would be addressed only within go/v4.


varshaprasad96 (approved):

camilamacedo86:
Hi @varshaprasad96 for we get this one merged we need an LGTM.
Since you approved either (thx for the review I will see it to let the bot merge this one )

PR #3355, opened 2023-04-22
:sparkles: (API and go/v4): Upgrade go version to 1.20

camilamacedo86:
## Description 

Upgrade Kubebuilder API to use go 1.20 as go/v4 (default scaffold) to use go/v4.


varshaprasad96 (approved):
/lgtm

@camilamacedo86 Do we need to do this for go/v3 also (its deprecated, but do we need to make this change to support latest Go version?)

varshaprasad96:
> /lgtm
> 
> @camilamacedo86 Do we need to do this for go/v3 also (its deprecated, but do we need to make this change to support latest Go version?)

Got my answer in https://github.com/kubernetes-sigs/kubebuilder/pull/3354. Please ignore the question :)

PR #3359, opened 2023-04-22
:bug: (go/v4) fix scaffold remove bintest dir from gitignore and dockerfile since this directory is not scaffolded

camilamacedo86:
## Description 

Remove  remove bintest dir from gitignore and dockerfile since this directory is not scaffolded


varshaprasad96 (approved):
/lgtm

Issue #3230, opened 2023-02-16
Migrate away from google.com gcp project kubebuilder


camilamacedo86:
## Description

Today we use GCP and have a project, kubebuilder. We need to move this infrastructure as described in this task: https://github.com/kubernetes/k8s.io/issues/2647. Currently, the GCP infra is used to build some [artefacts](https://book.kubebuilder.io/reference/artifacts.html). 

**kube-rbac-proxy images**

These images are used to build a sidecar for the manager, see:

https://github.com/kubernetes-sigs/kubebuilder/blob/3044376bff38c796b1ea7f4ade862e0a621b74d7/testdata/project-v2/config/default/manager_auth_proxy_patch.yaml#L11-L13

To know how they are built, you can check it [here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/RELEASE.md#to-build-the-kube-rbac-proxy-images).

**kubebuilder-tools**

The kubebuilder tools ship the required binaries to test the projects using [EnvTest](https://book.kubebuilder.io/reference/envtest.html). We scaffold a target to download them in the projects as well:

https://github.com/kubernetes-sigs/kubebuilder/blob/3044376bff38c796b1ea7f4ade862e0a621b74d7/testdata/project-v4/Makefile#L57-L59

To know how they are built, you can check it [here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/RELEASE.md#to-build-the-kubebuilder-tools-artifacts-required-to-use-env-test)

## Goal

This task aims to migrate the infrastructure used to build and provide them from the new location.

## Motivation

To know more about see: https://kubernetes.io/blog/2020/05/27/an-introduction-to-the-k8s-infrastructure-working-group/

## Impact

The projects built so far will still need to work on the images scaffold
It is a breaking change and can indeed break the workloads running on the cluster (in the case of the proxy image). 

So, we will need to work on providing those from the new places, see if we can also make a copy from all artefacts built and provide so for to the new place, and communicate asap.

IMPORTANT: We should not stop building in the old/current infrastructure until we stop using it. Therefore, that means we need to build in both for a time. I would also suggest that you build these artefacts in the new infrastructure. 

## What do we need to do

- [x] - Create the new infrastructure - https://github.com/kubernetes/k8s.io/pull/4586
- [x] - Add the jobs to start to build and produce the artefacts in the new infra : https://github.com/kubernetes/test-infra/pull/29351/files
[...]

PR #3362, opened 2023-04-22
:sparkles: add new build for new k8s registry infra-prow

camilamacedo86:
## Description

This PR moves the current implementation to build-legacy and adds a new directory to build the images for the new k8s registry within its infrastructure. More info: https://github.com/kubernetes-sigs/kubebuilder/issues/3230

Closes: https://github.com/kubernetes-sigs/kubebuilder/pull/3246/files

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ its upstream quay.io source.
 GCP Cloud Build watches this branch.  On every push, it runs the pipeline
 defined in [build/cloudbuild_kube-rbac-proxy.yaml][cloudbuild-file], which
 grabs the source images from `quay.io/brancz/kube-rbac-proxy` and tags them as
-`gcr.io/kubebuilder/kube-rbac-proxy`, with a tag for each arch as well as
+`gcr.io/k8s-staging-kubebuilder/kube-rbac-proxy`, with a tag for each arch as well as
 a single manifest bundle of:
 
 - amd64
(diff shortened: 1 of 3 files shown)


build/cloudbuild_kube-rbac-proxy.yaml
```diff
@@ -16,9 +16,9 @@ substitutions:
...
   _KUBE_RBAC_PROXY_VERSION: v0.14.1
 steps:
-- name: "gcr.io/cloud-builders/docker"
+- name: 'gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20211118-2f2d816b90'
   env:
   - "KUBE_RBAC_PROXY_VERSION=${_KUBE_RBAC_PROXY_VERSION}"
   entrypoint: "/usr/bin/env"
   args: ["bash", "-c", "./build/build.sh"]
-images: ["gcr.io/kubebuilder/kube-rbac-proxy:${_KUBE_RBAC_PROXY_VERSION}"]
+images: ["gcr.io/k8s-staging-kubebuilder/kube-rbac-proxy:${_KUBE_RBAC_PROXY_VERSION}"]
```
camilamacedo86, line 24: Hi @rikatz,

See that I just changed the step name and image to be built.
Is that right? We need to:
- a) build gcr.io/k8s-staging-kubebuilder/kube-rbac-proxy that will be triggered by:
- b) add the image built in https://github.com/kubernetes/k8s.io/blob/main/k8s.gcr.io/images/k8s-staging-kubebuilder/images.yaml so that it will be available via the registry, and we can download it with `docker pull k8s.io/kubebuilder/kube-rbac-proxy:<TAG>`

c/c @varshaprasad96

camilamacedo86: We use the image to re-tag it see `gcr.io/kubebuilder/kube-rbac-proxy` the above one. So, I don't know how we would use it anymore. Maybe I am missing something. 

See that we download quay.io/brancz/kube-rbac-proxy from https://github.com/brancz/kube-rbac-proxy and then rebuild it with a new tag so that we can store that.

ameukam: The `images` field is not required but can be useful to explicitly declare the name of the image(s) you want to build.

PR #3363, opened 2023-04-23
:bug: (go/v4, deployimage): fix the name controllers test function

camilamacedo86:
## Description

 fix the name controllers test function which has been scaffolded as` func TestAPIs(t *testing.T) {` for the controllers.


varshaprasad96 (approved):
/lgtm

PR #3364, opened 2023-04-23
✨ (deployImage) improve the controller tests scaffolded

camilamacedo86:
## Description 

This PR removes from the controller test implemented by the Deploy Image plugin.
Improves the before and after each test by applying the CR on it to ensure better practices since, for each reconciliation test, we ought to apply the resource. So, also remove the CR applied after the tests are executed.


camilamacedo86:
@varshaprasad96 thank you for the review, can we merge this one now?

Kavinjsir (approved):
lgtm

camilamacedo86:
We have an approval / lgtm so I think that is fine for we are able to move forward.

PR #3365, opened 2023-04-23
:bug:(go/v4, deployImage): fix scaffold to allow run the tests directly. Currently, is only possible run via the makefile targets

camilamacedo86:
## Description

Fix the scaffold to allow running the tests directly. Currently, it will fail because we install the required binaries at the bin directory inside of the project but that is not the default path configured in controller-runtime.

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/api/webhook_suitetest.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/api/webhook_suitetest.go
@@ -36,6 +36,10 @@ type WebhookSuite struct { //nolint:maligned
 	// todo: currently is not possible to know if an API was or not scaffolded. We can fix it when #1826 be addressed
 	WireResource bool
 
+	// K8SVersion define the k8s version used to do the scaffold
+	// so that is possible retrieve the binaries
+	K8SVersion string
+
 	// BaseDirectoryRelativePath define the Path for the base directory when it is multigroup
 	BaseDirectoryRelativePath string
 }
@@ -133,16 +137,20 @@ package {{ .Resource.Version }}
 
 import (
 	"context"
+	"crypto/tls"
+	"fmt"
+	"net"
 	"path/filepath"
 	"testing"
-	"fmt"
+	"time"
+	"runtime"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	%s
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
-	"k8s.io/apimachinery/pkg/runtime"
+	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
@@ -174,6 +182,15 @@ var _ = BeforeSuite(func() {
 	testEnv = &envtest.Environment{
 		CRDDirectoryPaths:     []string{filepath.Join({{ .BaseDirectoryRelativePath }}, "config", "crd", "bases")},
 		ErrorIfCRDPathMissing: {{ .WireResource }},
+
+		// The BinaryAssetsDirectory is only required if you want to run the tests directly
+		// without call the makefile target test. If not informed it will look for the
+		// default path defined in controller-runtime which is /usr/local/kubebuilder/.
+		// Note that you must have the required binaries setup under the bin directory to perform
+		// the tests directly. When we run make test it will be setup and used automatically.
+		BinaryAssetsDirectory: filepath.Join({{ .BaseDirectoryRelativePath }}, "bin", "k8s",
+			fmt.Sprintf("{{ .K8SVersion }}-%%s-%%s", runtime.GOOS, runtime.GOARCH)),
+
 		WebhookInstallOptions: envtest.WebhookInstallOptions{
 			Paths: []string{filepath.Join({{ .BaseDirectoryRelativePath }}, "config", "webhook")},
 		},
@@ -185,7 +202,7 @@ var _ = BeforeSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 	Expect(cfg).NotTo(BeNil())
 
-	scheme := runtime.NewScheme()
+	scheme := apimachineryruntime.NewScheme()
 	err = AddToScheme(scheme)
 	Expect(err).NotTo(HaveOccurred())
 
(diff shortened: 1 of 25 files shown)


camilamacedo86:
pull-kubebuilder-e2e-k8s-1-26-0

pratikmota:
One small observation:
BinaryAssetsDirectory: filepath.Join("..", "..", "..", "bin", "k8s",fmt.Sprintf("**1.26.1**-%s-%s", runtime.GOOS, runtime.GOARCH)),
- For all test suites, Here static string is added for k8 version - 1.26.1 , May be we can make it dynamic ? 
- We can take current version from K8SVersion = "1.26.1"
- So in future if we change K8 version then doese not require to change in all places of BinaryAssetsDirectory ?

pkg/plugins/golang/v4/scaffolds/internal/templates/controllers/controller_suitetest.go
```diff
@@ -34,6 +34,10 @@ type SuiteTest struct {
 	machinery.BoilerplateMixin
 	machinery.ResourceMixin
 
+	// K8SVersion define the k8s version used to do the scaffold
+	// so that is possible retrieve the binaries
+	K8SVersion string
+
```
varshaprasad96, line 40: @camilamacedo86 I like this change. But have this question on adding this option. Having k8s version here would not necessarily mean that my project is compatible with that particular version of K8s. It would just mean that we are scaffolding our test suite such that binaries for that version are downloaded. Do we need docs on this, or can we rename this to be `K8sTestVersion`?

camilamacedo86: The binaries downloaded are within the version defined on the Makefile. 
Please, feel free to provide any wording suggestion that you see that would fit better and make things more understandable that I will accepted :-)

camilamacedo86: I updated to have the same name used in the Makefile. (ENVTEST_K8S_VERSION)

camilamacedo86:
Hi @pratikmota,

> 
> One small observation:
> BinaryAssetsDirectory: filepath.Join("..", "..", "..", "bin", "k8s",fmt.Sprintf("1.26.1-%s-%s", runtime.GOOS, runtime.GOARCH)),
> 
> For all test suites, Here static string is added for k8 version - 1.26.1 , May be we can make it dynamic ?
> We can take current version from K8SVersion = "1.26.1"
> So in future if we change K8 version then doese not require to change in all places of BinaryAssetsDirectory ? 


How would we be able to change it dynamic? What is your suggestion to do that? 
Note that we scaffold the binary version that should be used in the makefile. So that in the same way that if users change the k8s version used and want to use another envtest bin, they must update the Makefile, they would need to update the version here as well.

pratikmota:
> Hi @pratikmota,
> 
> > One small observation:
> > BinaryAssetsDirectory: filepath.Join("..", "..", "..", "bin", "k8s",fmt.Sprintf("1.26.1-%s-%s", runtime.GOOS, runtime.GOARCH)),
> > For all test suites, Here static string is added for k8 version - 1.26.1 , May be we can make it dynamic ?
> > We can take current version from K8SVersion = "1.26.1"
> > So in future if we change K8 version then doese not require to change in all places of BinaryAssetsDirectory ?
> 
> How would we be able to change it dynamic? What is your suggestion to do that? Note that we scaffold the binary version that should be used in the makefile. So that in the same way that if users change the k8s version used and want to use another envtest bin, they must update the Makefile, they would need to update the version here as well.

Okay. I just saw similar changes without hardcode version- "BinaryAssetsDirectory: filepath.Join({{ .BaseDirectoryRelativePath }}, "bin", "k8s",fmt.Sprintf("{{ **.K8SVersion** }}-%%s-%%s", runtime.GOOS, runtime.GOARCH)),"
This might be helpful?  so will not require to change here version manually in all files.

pkg/plugins/golang/v4/scaffolds/internal/templates/api/webhook_suitetest.go
```diff
@@ -174,6 +182,15 @@ var _ = BeforeSuite(func() {
...
 		ErrorIfCRDPathMissing: {{ .WireResource }},
+
+		// The BinaryAssetsDirectory is only required if you want to run the tests directly
+		// without call the makefile target test. If not informed it will look for the
+		// default path defined in controller-runtime which is /usr/local/kubebuilder/.
+		// Note that you must have the required binaries setup under the bin directory to perform
+		// the tests directly. When we run make test it will be setup and used automatically.
+		BinaryAssetsDirectory: filepath.Join({{ .BaseDirectoryRelativePath }}, "bin", "k8s",
+			fmt.Sprintf("{{ .K8SVersion }}-%%s-%%s", runtime.GOOS, runtime.GOARCH)),
+
```
camilamacedo86, line 193: @pratikmota see ^

camilamacedo86:
Hi @pratikmota,

> @camilamacedo86 Okay. I just saw similar changes without hardcode version- "BinaryAssetsDirectory: filepath.Join({{ .BaseDirectoryRelativePath }}, "bin", "k8s",fmt.Sprintf("{{ .K8SVersion }}-%%s-%%s", runtime.GOOS, runtime.GOARCH)),"
This might be helpful? so will not require to change here version manually in all files.

It is in this way in the boilerplate/templates see https://github.com/kubernetes-sigs/kubebuilder/pull/3365/files#r1186883942. However, the tool does the scaffolds of the code within the values that exist in the K8SVersion in the specific Kubebuilder. Then, after that, if users want to change their projects they need to update manually as in any other place that we scaffold versions.

pkg/plugins/golang/v4/scaffolds/internal/templates/controllers/controller_suitetest.go
```diff
@@ -165,6 +171,14 @@ var _ = BeforeSuite(func() {
 	testEnv = &envtest.Environment{
 		CRDDirectoryPaths:     []string{filepath.Join({{ .CRDDirectoryRelativePath }}, "config", "crd", "bases")},
 		ErrorIfCRDPathMissing: {{ .Resource.HasAPI }},
+
+		// The BinaryAssetsDirectory is only required if you want to run the tests directly
+		// without call the makefile target test. If not informed it will look for the
```
varshaprasad96, line 176:
```suggestion
		// without calling the makefile target test. If not informed it will look for the
```

pkg/plugins/golang/v4/scaffolds/internal/templates/controllers/controller_suitetest.go
```diff
@@ -165,6 +171,14 @@ var _ = BeforeSuite(func() {
...
+		// default path defined in controller-runtime which is /usr/local/kubebuilder/.
+		// Note that you must have the required binaries setup under the bin directory to perform
+		// the tests directly. When we run make test it will be setup and used automatically.
```
varshaprasad96, line 179: Nit ```suggestion
		// the tests directly. When we run make test it will be setup using "setup-envtest" and used automatically.
```

varshaprasad96 (approved):
Just 2 nits in the comments. Rest looks good to me! Thanks @camilamacedo86 
/lgtm
/approve

PR #3371, opened 2023-04-27
:sparkles: (go/v4): Allows container tool to be configured in Makefile 

redhatrises:
- This allows building with other container tools like podman. Keeps docker as the default.

### DIFF
--- a/docs/book/src/component-config-tutorial/testdata/project/Makefile
+++ b/docs/book/src/component-config-tutorial/testdata/project/Makefile
@@ -11,6 +11,12 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# CONTAINER_TOOL defines the container tool to be used for building images.
+# Be aware that the target commands are only tested with Docker which is
+# scaffolded by default. However, you might want to replace it to use other
+# tools. (i.e. podman)
+CONTAINER_TOOL ?= docker
+
 # Setting SHELL to bash allows bash commands to be executed by recipes.
 # Options are set to exit when a recipe line exits non-zero or a piped command fails.
 SHELL = /usr/bin/env bash -o pipefail
@@ -73,11 +79,11 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
 docker-build: test ## Build docker image with the manager.
-	docker build -t ${IMG} .
+	$(CONTAINER_TOOL) build -t ${IMG} .
 
 .PHONY: docker-push
 docker-push: ## Push docker image with the manager.
-	docker push ${IMG}
+	$(CONTAINER_TOOL) push ${IMG}
 
 # PLATFORMS defines the target platforms for  the manager image be build to provide support to multiple
 # architectures. (i.e. make docker-buildx IMG=myregistry/mypoperator:0.0.1). To use this option you need to:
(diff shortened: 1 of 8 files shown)


pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
```diff
@@ -71,6 +71,10 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# CONTAINER_TOOL defines the container tool to be used for building images.
```
camilamacedo86, line 74: All makefile targets does not work with ANY other container tool
For example  `docker-buildx: test ## Build and push docker image for the manager for cross-platform support` will really work within any container tool or podman?

So, if you would like to create the env var here it is fine but I do not think that we can say that it can be used with podman . Also, we would need to highlight that it is only tested/support with docker by default. We do not test other tools within. 

Is that make sense?

Therefore, if we provide the further clarifications I am OK within.

redhatrises: Hello @camilamacedo86 

Iirc podman does have a buildx subcommand, but I agree as well that not all container tools will support all Makefile targets. Where would you prefer to add verbiage on this? Where the $CONTAINER_TOOL variable is set?

camilamacedo86: Hi @redhatrises, 

Thank you for the contribution and for share that. 
Please, feel free to wording it as you see fit.  I think anything like follows would be great. 

> # CONTAINER_TOOL defines the container tool to be used for building images. Note that the target commands are only tested with Docker which is scaffolded by default. However, you might want to replace it to use other tools. (i.e. podman)

Also, could you please only apply the change in the go/v4 plugin?
The plugin v3 is deprecated so ideally we should only perform changes on it if is some critical bug. So we avoid introduce regressions either motivate users to upgrade their projects.     

Again, thank you a lot for your contribution 🥇 


/ok-to-test

varshaprasad96: Thoughts on taking in a build-cmd directly? Something like [this](https://sdk.operatorframework.io/docs/cli/operator-sdk_pkgman-to-bundle/#options). If the user provides an entire cmd through an envVar we run it, or we default to docker commands. This will enable customizations on path of where binary can be located too. 
cc: @camilamacedo86

docs/book/src/component-config-tutorial/testdata/project/Makefile
```diff
@@ -11,6 +11,10 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# CONTAINER_TOOL defines the container tool to be used for building images.
+# You can use docker or podman
+CONTAINER_TOOL ?= docker
```
mhrivnak, line 16: I've seen this pattern be successful and useful in other projects. It uses podman if it's installed, otherwise defaults to docker. That seems to match people's usage fairly well, because typically the only reason someone appears to have both installed is if they use podman and have some sort of alias of docker to podman (see `dnf info podman-docker` for example).

```
CONTAINER_TOOL=$( command -v podman &> /dev/null && echo "podman" || echo "docker")
```

pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go
redhatrises: > Hi @redhatrises,
> 
> Thank you for the contribution and for share that. Please, feel free to wording it as you see fit. I think anything like follows would be great.
> 
> > CONTAINER_TOOL defines the container tool to be used for building images. Be aware that the target commands are only tested with Docker which is scaffolded by default. However, you might want to replace it to use other tools. (i.e. podman)
> 
> Also, could you please only apply the change in the go/v4 plugin? The plugin v3 is deprecated so ideally we should only perform changes on it if is some critical bug. So we avoid introduce regressions either motivate users to upgrade their projects.
> 
> Again, thank you a lot for your contribution 🥇
> 
> /ok-to-test

@camilamacedo86 I added your wording, because it works well. I also removed updates to the v3 plugin.

camilamacedo86 (approved):
 
/lgtm
/approved

Issue #3373, opened 2023-04-29
Commands should not fail when/if hack/license is not present


camilamacedo86:
### What broke? What's expected?

- Users should be able to use the tool without setting the license in all files. 

### Reproducing this issue

- Create a new project
- Remove the hack/boilerplate.go.txt
- Try to create an API

Then , you would will see the following error: 
```
2023/04/29 12:50:07 failed to create API: unable to scaffold with "base.go.kubebuilder.io/v4": error scaffolding API/controller: unable to load boilerplate: open hack/boilerplate.go.txt: no such file or directory

````
**Aceptence Criteria**

- If hack/boilerplate.go.txt is not present, then we should not fail as described above

The fix is here: https://github.com/kubernetes-sigs/kubebuilder/blob/7aa151a9ed9200a2c73e279506474f56d10e2135/pkg/plugins/golang/v4/scaffolds/api.go#L67-L70 (If cannot find the file then, we pass an empty string like “”)

### KubeBuilder (CLI) Version

3.10.0

### PROJECT version

3

ashutosh887:
Let me work on this 
/assign

camilamacedo86:
This bug should only be fixed for go/v4 plugin

yyy1000:
When I check #3377, I found this issue. I'm glad to help with this together. :)
/assign

yyy1000:
I think we can close this since #3378 is merged?

PR #3378, opened 2023-05-01
🐛 (go/v4): Scaffold when no boilerplate file

yyy1000:
### Description
This pull request addresses issue #3373, which prevented kubebuilder from creating an API when no boilerplate file was present.
I have tested the changes locally with the following steps:
1. Rebuilt the Kubebuilder binary with updated code
2. Used Kubebuilder to initialize a new project
3. Deleted the 'boilerplate' file
4. Modified the 'generate' target in the Makefile as follows:
```makefile
.PHONY: generate
	$(CONTROLLER_GEN) object paths="./..."
```
5. Executed the 'create api' command.

Additionally, I tried to make some changes that enable the 'create api' command to modify the 'Makefile' when there's no boilerplate file and let the user skip step 4 altogether, making it possible to successfully run 'post-scaffold'(make generate). It seems not a little change, but I'm glad to do that if it is worth doing. :)

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/api.go
+++ b/pkg/plugins/golang/v4/scaffolds/api.go
@@ -17,6 +17,7 @@ limitations under the License.
 package scaffolds
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/spf13/afero"
@@ -67,7 +68,11 @@ func (s *apiScaffolder) Scaffold() error {
 	// Load the boilerplate
 	boilerplate, err := afero.ReadFile(s.fs.FS, hack.DefaultBoilerplatePath)
 	if err != nil {
-		boilerplate = []byte("")
+		if errors.Is(err, afero.ErrFileNotFound) {
+			boilerplate = []byte("")
+		} else {
+			return fmt.Errorf("error scaffolding API/controller: unable to load boilerplate: %w", err)
+		}
 	}
 
 	// Initialize the machinery.Scaffold that will write the files to disk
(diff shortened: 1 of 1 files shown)


pkg/plugins/golang/v4/scaffolds/api.go
```diff
@@ -67,7 +67,7 @@ func (s *apiScaffolder) Scaffold() error {
 	// Load the boilerplate
 	boilerplate, err := afero.ReadFile(s.fs.FS, hack.DefaultBoilerplatePath)
 	if err != nil {
```
varshaprasad96, line 69: could we specifyically check for fileNotFound error and skip with `errors.Is(err, afero.ErrFileNotFound)`. In case its something else (like permission issue), we can return the error.

camilamacedo86: +1 in in the above suggestion. 

We need to only use an empty string when we know the file does not exist.

camilamacedo86 (commented):
/ok-to-test

pkg/plugins/golang/v4/scaffolds/api.go
yyy1000: Thank you! It's very helpful. 😀 I made changes on this.

camilamacedo86 (approved):

camilamacedo86:
//test pull-kubebuilder-e2e-k8s-1-24-7

camilamacedo86:
Also, 

/lgtm 

Since address the Varsha suggestions either.

PR #3379, opened 2023-05-01
✨use bundle options struct for creating new bundles

NikhilSharmaWe:
### Description
This PR uses bundle options struct instead of funcs for creating new bundles through [`NewBundleWithOptions`](https://github.com/NikhilSharmaWe/kubebuilder/blob/upgradeBundle/pkg/plugin/bundle.go#L106) function.

### Motivation
Fixes: #3307

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -37,8 +37,8 @@ import (
 	grafanav1alpha1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/optional/grafana/v1alpha"
 )
 
+// nolint:lll
 func main() {
-
 	const deprecateMessageGoV3Bundle = "This version is deprecated." +
 		"The `go/v3` cannot scaffold projects using kustomize versions v4x+" +
 		" and cannot fully support Kubernetes 1.25+." +
@@ -46,17 +46,12 @@ func main() {
 		"Please, check the migration guide to learn how to upgrade your project"
 
 	// Bundle plugin which built the golang projects scaffold by Kubebuilder go/v3
-	gov3Bundle, _ := plugin.NewBundleWithOptions(plugin.WithName(golang.DefaultNameQualifier),
-		plugin.WithVersion(plugin.Version{Number: 3}),
-		plugin.WithDeprecationMessage(deprecateMessageGoV3Bundle),
-		plugin.WithPlugins(kustomizecommonv1.Plugin{}, golangv3.Plugin{}),
-	)
+	gov3 := plugin.NewBundlePluginBuilder()
+	gov3Bundle, _ := gov3.WithName(golang.DefaultNameQualifier).WithVersion(plugin.Version{Number: 3}).WithDeprecationMessage(deprecateMessageGoV3Bundle).WithPlugins(kustomizecommonv1.Plugin{}, golangv3.Plugin{}).Build()
 
 	// Bundle plugin which built the golang projects scaffold by Kubebuilder go/v4 with kustomize alpha-v2
-	gov4Bundle, _ := plugin.NewBundleWithOptions(plugin.WithName(golang.DefaultNameQualifier),
-		plugin.WithVersion(plugin.Version{Number: 4}),
-		plugin.WithPlugins(kustomizecommonv2alpha.Plugin{}, golangv4.Plugin{}),
-	)
+	gov4 := plugin.NewBundlePluginBuilder()
+	gov4Bundle, _ := gov4.WithName(golang.DefaultNameQualifier).WithVersion(plugin.Version{Number: 4}).WithPlugins(kustomizecommonv2alpha.Plugin{}, golangv4.Plugin{}).Build()
 
 	fs := machinery.Filesystem{
 		FS: afero.NewOsFs(),
--- a/pkg/plugin/bundle.go
+++ b/pkg/plugin/bundle.go
@@ -31,31 +31,35 @@ type bundle struct {
 	deprecateWarning         string
 }
 
-type BundleOption func(*bundle)
-
-func WithName(name string) BundleOption {
-	return func(opts *bundle) {
-		opts.name = name
-	}
+type BundlePluginBuilder struct {
+	name             string
+	version          Version
+	plugins          []Plugin
+	deprecateWarning string
 }
 
-func WithVersion(version Version) BundleOption {
-	return func(opts *bundle) {
-		opts.version = version
-	}
+func NewBundlePluginBuilder() BundlePluginBuilder {
+	return BundlePluginBuilder{}
 }
 
-func WithPlugins(plugins ...Plugin) BundleOption {
-	return func(opts *bundle) {
-		opts.plugins = plugins
-	}
+func (bp *BundlePluginBuilder) WithName(name string) *BundlePluginBuilder {
+	bp.name = name
+	return bp
 }
 
-func WithDeprecationMessage(msg string) BundleOption {
-	return func(opts *bundle) {
-		opts.deprecateWarning = msg
-	}
+func (bp *BundlePluginBuilder) WithVersion(version Version) *BundlePluginBuilder {
+	bp.version = version
+	return bp
+}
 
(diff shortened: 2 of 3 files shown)


cmd/main.go
```diff
@@ -37,26 +37,21 @@ import (
 	grafanav1alpha1 "sigs.k8s.io/kubebuilder/v3/pkg/plugins/optional/grafana/v1alpha"
 )
 
+// nolint:lll
```
varshaprasad96, line 40: Any specific reason for adding this?

pkg/plugin/bundle.go
```diff
@@ -31,31 +31,38 @@ type bundle struct {
...
-		opts.name = name
+type BundlePlugin struct {
+	Options struct {
+		name                     string
+		version                  Version
+		plugins                  []Plugin
+		supportedProjectVersions []config.Version
+		deprecateWarning         string
 	}
 }
```
varshaprasad96, line 42: Wondering if we need the nested struct. Wdyt of making it this?
```suggestion
type BundlePluginOptions struct {
		name                     string
		version                  Version
		plugins                  []Plugin
		supportedProjectVersions []config.Version
		deprecateWarning         string
}
```

pkg/plugin/bundle.go
```diff
@@ -31,31 +31,38 @@ type bundle struct {
...
 
-func WithVersion(version Version) BundleOption {
-	return func(opts *bundle) {
-		opts.version = version
-	}
+func NewBundlePlugin() BundlePlugin {
+	return BundlePlugin{}
```
varshaprasad96, line 45: So if `BundlePlugin` is exported, then there is no need of calling `NewBundlePlugin` func. We can make the BundlePlugin not to be exported (bundlePlugin) and have this func.

pkg/plugin/bundle.go
```diff
@@ -96,23 +103,17 @@ func NewBundle(name string, version Version, deprecateWarning string, plugins ..
...
 		return nil, fmt.Errorf("in order to bundle plugins, they must all support at least one common project version")
 	}
 
 	// Plugins may be bundles themselves, so unbundle here
 	// NOTE(Adirio): unbundling here ensures that Bundle.Plugin always returns a flat list of Plugins instead of also
 	//               including Bundles, and therefore we don't have to use a recursive algorithm when resolving.
-	allPlugins := make([]Plugin, 0, len(bundleOpts.plugins))
-	for _, plugin := range bundleOpts.plugins {
+	allPlugins := make([]Plugin, 0, len(bp.Options.plugins))
+	for _, plugin := range bp.Options.plugins {
```
varshaprasad96, line 116: Was there a specific reason to make a nested struct?

NikhilSharmaWe: @varshaprasad96 Sorry about that. Changed it.

Actually, according to me the changes I made in this https://github.com/kubernetes-sigs/kubebuilder/pull/3312 for https://github.com/kubernetes-sigs/kubebuilder/issues/3307 were reasonable, since it implemented the functional arguments for the `NewBundleWithOptions` func for creating Bundles to prevent the long function signature for the [NewBundle](https://github.com/kubernetes-sigs/kubebuilder/blob/0ceec7702d6fa6f229e414dc3d9af861a75d6a59/pkg/plugin/bundle.go#L77) func.

But @camilamacedo86 has suggested that using struct for bundle options is more reasonable. 

But with the calling the function for bundle creation becomes very long like https://github.com/kubernetes-sigs/kubebuilder/blob/0ceec7702d6fa6f229e414dc3d9af861a75d6a59/cmd/main.go#L50 Which also makes the lint test fails to have to skip them, therefore https://github.com/kubernetes-sigs/kubebuilder/pull/3379#discussion_r1181602409.

Please inform me if my changes are reasonable and why this approach is better than https://github.com/kubernetes-sigs/kubebuilder/pull/3312. Maybe I am missing here something.

cmd/main.go
```diff
@@ -37,26 +37,21 @@ import (
...
 		"Please, check the migration guide to learn how to upgrade your project"
 
 	// Bundle plugin which built the golang projects scaffold by Kubebuilder go/v3
-	gov3Bundle, _ := plugin.NewBundleWithOptions(plugin.WithName(golang.DefaultNameQualifier),
-		plugin.WithVersion(plugin.Version{Number: 3}),
-		plugin.WithDeprecationMessage(deprecateMessageGoV3Bundle),
-		plugin.WithPlugins(kustomizecommonv1.Plugin{}, golangv3.Plugin{}),
-	)
+	bpgov3 := plugin.NewBundlePluginOptions()
+	gov3Bundle, _ := bpgov3.WithName(golang.DefaultNameQualifier).WithVersion(plugin.Version{Number: 3}).WithDeprecationMessage(deprecateMessageGoV3Bundle).WithPlugins(kustomizecommonv1.Plugin{}, golangv3.Plugin{}).NewBundleWithOptions()
```
camilamacedo86, line 50:
```suggestion
	gov3Bundle, _ := bpgov3.WithName(golang.DefaultNameQualifier).WithVersion(plugin.Version{Number: 3}).WithDeprecationMessage(deprecateMessageGoV3Bundle).WithPlugins(kustomizecommonv1.Plugin{}, golangv3.Plugin{}).Build()
```

cmd/main.go
```diff
@@ -37,26 +37,21 @@ import (
...
 		"It is recommended to upgrade your project to the latest versions available (go/v4)." +
 		"Please, check the migration guide to learn how to upgrade your project"
 
 	// Bundle plugin which built the golang projects scaffold by Kubebuilder go/v3
-	gov3Bundle, _ := plugin.NewBundleWithOptions(plugin.WithName(golang.DefaultNameQualifier),
-		plugin.WithVersion(plugin.Version{Number: 3}),
-		plugin.WithDeprecationMessage(deprecateMessageGoV3Bundle),
-		plugin.WithPlugins(kustomizecommonv1.Plugin{}, golangv3.Plugin{}),
-	)
+	bpgov3 := plugin.NewBundlePluginOptions()
```
camilamacedo86, line 49:
```suggestion
gov3 := plugin.NewBundlePluginBuilder()
```

pkg/plugin/bundle.go
```diff
@@ -31,31 +31,35 @@ type bundle struct {
 	deprecateWarning         string
 }
 
-type BundleOption func(*bundle)
+type BundlePluginOptions struct {
```
camilamacedo86, line 34:
```suggestion
type BundlePluginBuilder struct {
```

pkg/plugin/bundle.go
```diff
@@ -31,31 +31,35 @@ type bundle struct {
...
+	version          Version
+	plugins          []Plugin
+	deprecateWarning string
+}
 
-func WithName(name string) BundleOption {
-	return func(opts *bundle) {
-		opts.name = name
-	}
+func NewBundlePluginOptions() BundlePluginOptions {
```
camilamacedo86, line 41:
```suggestion
func NewBundlePluginBuilder() BundlePluginOptions {
```

pkg/plugin/bundle.go
```diff
@@ -96,23 +100,17 @@ func NewBundle(name string, version Version, deprecateWarning string, plugins ..
...
 // The list of supported project versions is computed from the provided plugins in options.
-func NewBundleWithOptions(opts ...BundleOption) (Bundle, error) {
-	bundleOpts := bundle{}
-
-	for _, opts := range opts {
-		opts(&bundleOpts)
-	}
-
-	supportedProjectVersions := CommonSupportedProjectVersions(bundleOpts.plugins...)
+func (bp *BundlePluginOptions) NewBundleWithOptions() (Bundle, error) {
```
camilamacedo86, line 103:
```suggestion
func (bp *BundlePluginOptions) Build() (Bundle, error) {
```

NikhilSharmaWe:
No need for this. 
The current implementation is reasonable and probably better.

PR #3387, opened 2023-05-04
✨Add to the PROJECT file the CLI tag version used to do the scaffold

### LINKED ISSUE #3229: Add to the PROJECT file the CLI tag version used to do the scaffold

The idea of this task is for we start to track in the PROJECT file what was the Kubebuilder version used to do the scaffold so that for proposed features like https://github.com/kubernetes-sigs/kubebuilder/pull/3221 or for support purpose for example this info can be helpful in the long term. 

### Extra Labels

_No response_

yyy1000:
### Description
This involves add a new field in PROJECT file, which is the CLI tag of Kubebuilder.
Will fix #3229

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,11 +46,11 @@ help: ## Display this help
 ##@ Build
 
 LD_FLAGS=-ldflags " \
-    -X main.kubeBuilderVersion=$(shell git describe --tags --dirty --broken) \
-    -X main.goos=$(shell go env GOOS) \
-    -X main.goarch=$(shell go env GOARCH) \
-    -X main.gitCommit=$(shell git rev-parse HEAD) \
-    -X main.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ') \
+    -X sigs.k8s.io/kubebuilder/v3/pkg/version.kubeBuilderVersion=$(shell git describe --tags --dirty --broken) \
+    -X sigs.k8s.io/kubebuilder/v3/pkg/version.goos=$(shell go env GOOS) \
+    -X sigs.k8s.io/kubebuilder/v3/pkg/version.goarch=$(shell go env GOARCH) \
+    -X sigs.k8s.io/kubebuilder/v3/pkg/version.gitCommit=$(shell git rev-parse HEAD) \
+    -X sigs.k8s.io/kubebuilder/v3/pkg/version.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ') \
     "
 .PHONY: build
 build: ## Build the project locally
--- a/docs/book/src/component-config-tutorial/testdata/project/PROJECT
+++ b/docs/book/src/component-config-tutorial/testdata/project/PROJECT
@@ -4,6 +4,7 @@
 # More info: https://book.kubebuilder.io/reference/project-config.html
 componentConfig: true
 domain: tutorial.kubebuilder.io
+kubebuilderVersion: unknown
 layout:
 - go.kubebuilder.io/v4
 projectName: project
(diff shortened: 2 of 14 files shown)


yyy1000:
So far I don't know how to handle the 'APIDiff' CI problem. :(
And also, Testdata verification fails though I run 'make generate'.🤔

Makefile
```diff
@@ -46,11 +46,11 @@ help: ## Display this help
...
-    -X main.kubeBuilderVersion=$(shell git describe --tags --dirty --broken) \
-    -X main.goos=$(shell go env GOOS) \
-    -X main.goarch=$(shell go env GOARCH) \
-    -X main.gitCommit=$(shell git rev-parse HEAD) \
-    -X main.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ') \
+    -X sigs.k8s.io/kubebuilder/v3/pkg/version.kubeBuilderVersion=$(shell git describe --tags --dirty --broken) \
+    -X sigs.k8s.io/kubebuilder/v3/pkg/version.goos=$(shell go env GOOS) \
+    -X sigs.k8s.io/kubebuilder/v3/pkg/version.goarch=$(shell go env GOARCH) \
+    -X sigs.k8s.io/kubebuilder/v3/pkg/version.gitCommit=$(shell git rev-parse HEAD) \
+    -X sigs.k8s.io/kubebuilder/v3/pkg/version.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ') \
```
camilamacedo86, line 53: can you please revert this changes that has no relation with the scope of this PR?

docs/book/src/component-config-tutorial/testdata/project/PROJECT
```diff
@@ -4,6 +4,7 @@
 # More info: https://book.kubebuilder.io/reference/project-config.html
 componentConfig: true
 domain: tutorial.kubebuilder.io
+kubebuilderVersion: unknown
 layout:
 - go.kubebuilder.io/v4
```
camilamacedo86, line 9: I think if we want to track this one we will need to have the CLI version used per plugin and scaffold.
See that we can init a project within the version X and then afterwords scaffold an API with N+2

Makefile
yyy1000: well, I use this changes to finish this PR... because I move the 'version.go' to other places.
If I don't move, I can't import that value from 'main' package.

docs/book/src/component-config-tutorial/testdata/project/PROJECT
yyy1000: Ah, got it.
So it would be applied to go/v2 and go/v3 also?

pkg/config/v2/config.go
```diff
@@ -93,6 +93,19 @@ func (c *cfg) SetProjectName(string) error {
 	}
 }
 
+// GetCLIVersion implements config.Config
+func (c cfg) GetCLIVersion() string {
```
camilamacedo86, line 97: We should not change the config/v2 it is deprecated and the changes should be addressed into the config/v3

Makefile
camilamacedo86: Hun, it does not seem to make sense. This target is used to build the binary, and making changes here does not allow you to import or not a code to be used. We should not perform those changes to address this one.

docs/book/src/component-config-tutorial/testdata/project/PROJECT
camilamacedo86: It should ONLY be applied to the Project Config v3.

camilamacedo86:
Great, I am closing this one based on the comment https://github.com/kubernetes-sigs/kubebuilder/issues/3229#issuecomment-1537293699, but we can look into it in the future.

PR #3391, opened 2023-05-06
🐛: (go/v4) Scaffold no header when lincense is type none

### LINKED ISSUE #3377: lincense type none is scaffolding a header

If we create a new project and use the flag option --license=none instead of not adding any license, it still scaffolds the boilerplate with header content `Copyright 2023 The {{ .owner }} authors.`. Therefore, this issue would be to fix this function by:

**I, as the user, would like to know to have any header added to my files when I use the option --license=none** 

In this case, the solution here is:

- Ensure that when this option is used, we will not scaffold hack/boilerplate.txt files
- Ensure when this option is used in the Makefile, we will scaffold the generate target within `object` instead of `object:headerFile="hack/boilerplate.go.txt"`: 

```
.PHONY: generate
	$(CONTROLLER_GEN) object:headerFile="" paths="./..."
````
**I am a user. I would like to have the current behaviour done when I select none for a new optional flag with a name that make s it clearer (`copyright`)**

Following the places where the change needs to be made. See:
- https://github.com/kubernetes-sigs/kubebuilder/blob/7aa151a9ed9200a2c73e279506474f56d10e2135/pkg/plugins/golang/v4/scaffolds/internal/templates/hack/boilerplate.go#L100-L111
- https://github.com/kubernetes-sigs/kubebuilder/blob/7aa151a9ed9200a2c73e279506474f56d10e2135/pkg/plugins/golang/v4/init.go#L87-L89

See that this one is a follow-up of : https://github.com/kubernetes-sigs/kubebuilder/issues/3373

### Reproducing this issue

_No response_

### KubeBuilder (CLI) Version

3.10.0

### PROJECT version

_No response_

### Plugin versions

```yaml
go/v4
```

### Other versions

_No response_

### Extra Labels

_No response_

### LINKED ISSUE #3390: 'make generate' can't work without kubebuilder binary

If we don't set the kubebuilder binary in our system PATH, the 'make generate' command will fail.
Error message:

> ERRO[0000] error Initializing the project: kubebuilder init --domain tutorial.kubebuilder.io --repo tutorial.kubebuilder.io/project --license apache2 --owner The Kubernetes authors --plugins=go/v4 --component-config failed with error: (exec: "kubebuilder": executable file not found in $PATH) 

And even we set the path, the result is not as we expect. Cause it would not use the lastest binary we compile when 'make generate'.

I think it's because the lastest PR: 
https://github.com/kubernetes-sigs/kubebuilder/pull/3231/files#diff-14a46a8334eccfba24cd4de8c126e3fc9ecf705935586be96d3b665ff403c010L37:~:text=binaryPath%20%3A%3D%20%22/tmp/kubebuilder/bin/kubebuilder%22

### Reproducing this issue

1. Fork the repo
2. Run 'make generate'

### KubeBuilder (CLI) Version

3.10

### PROJECT version

_No response_

### Plugin versions

_No response_

### Other versions

_No response_

### Extra Labels

_No response_

yyy1000:
**Description**
Fix #3377
Now it would not scaffold any header if we use 'none' in license option.

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/init.go
+++ b/pkg/plugins/golang/v4/scaffolds/init.go
@@ -81,25 +81,32 @@ func (s *initScaffolder) Scaffold() error {
 		machinery.WithConfig(s.config),
 	)
 
-	bpFile := &hack.Boilerplate{
-		License: s.license,
-		Owner:   s.owner,
-	}
-	bpFile.Path = s.boilerplatePath
-	if err := scaffold.Execute(bpFile); err != nil {
-		return err
-	}
+	if s.license != "none" {
+		bpFile := &hack.Boilerplate{
+			License: s.license,
+			Owner:   s.owner,
+		}
+		bpFile.Path = s.boilerplatePath
+		if err := scaffold.Execute(bpFile); err != nil {
+			return err
+		}
 
-	boilerplate, err := afero.ReadFile(s.fs.FS, s.boilerplatePath)
-	if err != nil {
-		return err
+		boilerplate, err := afero.ReadFile(s.fs.FS, s.boilerplatePath)
+		if err != nil {
+			return err
+		}
+		// Initialize the machinery.Scaffold that will write the files to disk
+		scaffold = machinery.NewScaffold(s.fs,
+			machinery.WithConfig(s.config),
+			machinery.WithBoilerplate(string(boilerplate)),
+		)
+	} else {
+		s.boilerplatePath = ""
+		// Initialize the machinery.Scaffold without boilerplate
+		scaffold = machinery.NewScaffold(s.fs,
+			machinery.WithConfig(s.config),
+		)
 	}
-
-	// Initialize the machinery.Scaffold that will write the files to disk
-	scaffold = machinery.NewScaffold(s.fs,
-		machinery.WithConfig(s.config),
-		machinery.WithBoilerplate(string(boilerplate)),
-	)
 
 	// If the KustomizeV2 was used to do the scaffold then
 	// we need to ensure that we use its supported Kustomize Version
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/hack/boilerplate.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/hack/boilerplate.go
@@ -106,8 +106,8 @@ Copyright {{ .Year }}.
 {{ index .Licenses .License }}*/`
 
 var knownLicenses = map[string]string{
-	"apache2": apache2,
-	"none":    "",
+	"apache2":   apache2,
+	"copyright": "",
 }
 
 const apache2 = `
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -110,7 +110,11 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
+	{{ if .BoilerplatePath -}}
 	$(CONTROLLER_GEN) object:headerFile={{printf "%q" .BoilerplatePath}} paths="./..."
+	{{- else -}}
+	$(CONTROLLER_GEN) object paths="./..."
+	{{- end }}
 
 .PHONY: fmt
 fmt: ## Run go fmt against code.


hack/docs/generate_samples.go
```diff
@@ -51,7 +50,7 @@ func UpdateComponentConfigTutorial() {
 }
 
 func UpdateCronjobTutorial() {
-	binaryPath := util.KubebuilderBinName
+	binaryPath := "/tmp/kubebuilder/bin/kubebuilder"
```
camilamacedo86, line 53: Could we revert this one and still use the const?

hack/docs/generate_samples.go
```diff
@@ -36,7 +35,7 @@ func main() {
 }
 
 func UpdateComponentConfigTutorial() {
-	binaryPath := util.KubebuilderBinName
+	binaryPath := "/tmp/kubebuilder/bin/kubebuilder"
```
camilamacedo86, line 38: Same here?

camilamacedo86 (commented):
Just please rebase with the master, for we have only your specific change and squash the commits. The change shows great. So we can get this one merged. 

/ok-to-test

hack/docs/generate_samples.go
yyy1000: I think use the const 'util.KubebuilderBinName' couldn't work. This will not use the kubebuilder binary complied by the command. 🤔 See #3390

camilamacedo86: It has been working so why it would stop to work within the changes of this PR?

yyy1000: well, I think it has been working for only two days. The last commit changing it is : https://github.com/kubernetes-sigs/kubebuilder/commit/d0a1806c1d25c70777004271a4b773b490de3bcd

camilamacedo86: This code is called when we run `make generate`.
The `make generate` is called when we do the CI checks and would NOT get merged if does not work. 

**See this code successfully executed to verify your PR: https://github.com/kubernetes-sigs/kubebuilder/actions/runs/4902477788/jobs/8754346010**

I can ensure that works either and you can properly very that by just running without these changes `make generate` locally within the master branch. 

Could you please revert these changes that have no relation to the scope of the PR?

camilamacedo86 (changes requested):
Hi @yyy1000,

Thank you for your contribution. 🥇 Just a few nits and then it seems great to fly:
 
- Remove changes in generate_samples.go, which are out of the scope of this PR and unnecessary and should not be made. We should **not** replace the const values that centralize the Kubebuilder binary with strings within the path.
- Squash the commits

yyy1000:
> Hi @yyy1000,
> 
> Thank you for your contribution. 🥇 Just a few nits and then it seems great to fly:
> 
> * Remove changes in generate_samples.go, which are out of the scope of this PR and unnecessary and should not be made. We should **not** replace the const values that centralize the Kubebuilder binary with strings within the path.
> * Squash the commits

Fixed it already. Thanks!

camilamacedo86 (approved):
It shows fine to me. 
I will approved and let see if we win a second review

yyy1000:
> 

Sorry for that. I used to think that only applied to 'license' flag.
Do you mean we can add a new optional flag called 'copyright' and make the new optional flag has the same behavior like 'license'?

camilamacedo86 (commented):
Hi @yyy1000,

Try to create a new project and use --license=none. You will see that it will create the header within the copyright. 

So, the idea is:
- a) none should be NONE. That means no license at all
- b) we can add a new optional value for the flag license to allow a user inform `--license=copyright` to have the current behaviour. 

Following the places where the change needs to be made. See:
- https://github.com/kubernetes-sigs/kubebuilder/blob/7aa151a9ed9200a2c73e279506474f56d10e2135/pkg/plugins/golang/v4/scaffolds/internal/templates/hack/boilerplate.go#L100-L111
- https://github.com/kubernetes-sigs/kubebuilder/blob/7aa151a9ed9200a2c73e279506474f56d10e2135/pkg/plugins/golang/v4/init.go#L87-L89
 
Please, let me know if that make sense and feel free to reach out if you have questions.

yyy1000:
> Hi @yyy1000,
> 
> Try to create a new project and use --license=none. You will see that it will create the header within the copyright.
> 
> So, the idea is:
> 
> * a) none should be NONE. That means no license at all
> * b) we can add a new optional value for the flag license to allow a user inform `--license=copyright` to have the current behaviour.
> 
> Following the places where the change needs to be made. See:
> 
> * https://github.com/kubernetes-sigs/kubebuilder/blob/7aa151a9ed9200a2c73e279506474f56d10e2135/pkg/plugins/golang/v4/scaffolds/internal/templates/hack/boilerplate.go#L100-L111
> * https://github.com/kubernetes-sigs/kubebuilder/blob/7aa151a9ed9200a2c73e279506474f56d10e2135/pkg/plugins/golang/v4/init.go#L87-L89
> 
> Please, let me know if that make sense and feel free to reach out if you have questions.

Sorry for missing that :) Thanks for your help, I have fixed it.

pkg/plugins/golang/v4/scaffolds/internal/templates/hack/boilerplate.go
```diff
@@ -106,8 +106,8 @@ Copyright {{ .Year }}.
 {{ index .Licenses .License }}*/`
 
 var knownLicenses = map[string]string{
-	"apache2": apache2,
-	"none":    "",
```
camilamacedo86, line 110: @yyy1000, we still need to have the option none 
Also, to keep the same behaviour that we have currently and not change the current default behaviour, we need to ensure that when the license flag is not informed the default value will be apache2.

pkg/plugins/golang/v4/scaffolds/internal/templates/hack/boilerplate.go
```diff
@@ -106,8 +106,8 @@ Copyright {{ .Year }}.
...
+	"apache2":   apache2,
+	"copyright": "",
```
camilamacedo86, line 110:
```suggestion
	"copyright": copyright, 
```

We need to ensure the type, like the above for apache2. Am I right?

pkg/plugins/golang/v4/scaffolds/internal/templates/hack/boilerplate.go
yyy1000: Well, I think it's OK to delete the option 'none'. If we use the option 'none', it will not execute `scaffold.Execute(bpFile)`. So `knownLicenses` will not be used.

pkg/plugins/golang/v4/scaffolds/internal/templates/hack/boilerplate.go
yyy1000: I think that's right. We can define the const value 'copyright' a empty string. Just like `const copyright = ""`?

camilamacedo86, line 110: We cannot have any copyright withing the value of none.
I think we might need to check if is possible to pass here: boilerplateTemplate

yyy1000: Actually I think the 'const copyright' == “”（an empty string) is enough. the ` {{ index .Licenses .License }}*/` ` will just be an empty string.

yyy1000:
As I tested it, it works OK now. Can we continue to merge it or should I add some tests in the PR first?

pkg/plugins/golang/v4/scaffolds/internal/templates/hack/boilerplate.go
camilamacedo86: So, how none can be valid and not exist in knownLicenses[f.License]?

pkg/plugins/golang/v4/scaffolds/init.go
```diff
@@ -81,26 +81,33 @@ func (s *initScaffolder) Scaffold() error {
...
-	if err := scaffold.Execute(bpFile); err != nil {
-		return err
-	}
+	if s.license != "none" {
+		bpFile := &hack.Boilerplate{
+			License: s.license,
+			Owner:   s.owner,
+		}
+		bpFile.Path = s.boilerplatePath
+		if err := scaffold.Execute(bpFile); err != nil {
```
yyy1000, line 90: well, that's a good question.  'none' is valid even if not exist in knownLicenses[f.License]. 
Let me explain it for you.
If we pass '--license=none', then it will skip this(line 90).
In `scaffold.Execute(bpFile)`, it will call `SetTemplateDefaults(`) and `Validate()`.  So if we set 'none', it will never use `knownLicenses[f.License]`. Does it make sense? @camilamacedo86

camilamacedo86 (approved):
It seems OK. Thank you for your contribution 🥇 

/approve

Issue #3186, opened 2023-01-26
Externalize declarative plugin 


camilamacedo86:
### What do you want to happen?

Kubebuilder introduces the plugin phase 2 implementation. One way we have a great example of how the community can leverage Kubebuilder to create your plugins would be by achieving the goal of moving the declarative plugin to its repository. See: https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/issues/293

This task is to track the need to work toward this goal.

aimuz:
At present, there seem to be many shortcomings in external plug-ins. For example, I want to implement a Layout plug-in. I refer to the implementation of go in the code, and cannot be well ported to external plug-ins. For now, I found that it relies on InjectConfig, but external plug-ins cannot provide

camilamacedo86:
Hi @aimuz,

To address this need @em-r began to work on it. See: https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/issues/293. The question raised on  https://github.com/kubernetes-sigs/kubebuilder/issues/3186#issuecomment-1407894107 seems specific to the API and its capabilities and would fit better in a new issue. Therefore, I created this i one : https://github.com/kubernetes-sigs/kubebuilder/issues/3245 (let's discuss it there)

em-r:
@camilamacedo86 I think this is a bit similar to our discussion in https://github.com/kubernetes-sigs/kubebuilder/issues/3214#issuecomment-1432284801
if it is worth it, I can look into adding support for injecting config into external plugins as part of https://github.com/kubernetes-sigs/kubebuilder/issues/3245

camilamacedo86:
Hi @em-r,

> if it is worth it, I can look into adding support for injecting config into external plugins as part of https://github.com/kubernetes-sigs/kubebuilder/issues/3245
I think we could improve to parse and inject it as well. Yes, please feel free your help is more than welcome !!!

camilamacedo86:
We are deprecating the plugin in kubebuilder. 
Also, I think we should keep the whole conversation about the new plugin externalized into the declarative project: https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/issues/293

PR #3395, opened 2023-05-08
:warning: deprecate declarative plugin

camilamacedo86:
## Description:

This PR deprecates the declarative plugin and adds the following notice:

Screenshot 2023-05-08 at 09 40 27

Also, we are:
- Updating the docs to make clear that it is deprecated
- Removing the sample under test data to stop to block us from moving forward with upgrades, such as now that we want to bump controller runtime alpha 0.15 and it will fail in : https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kubebuilder/3394/pull-kubebuilder-test/1655479526613848064

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/3186
Related to : https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/issues/293


varshaprasad96:
@camilamacedo86 I'm fine moving ahead with the PR. Just want to confirm if Kubebuilder declarative pattern maintainers are comfortable with us doing this, or would like to first externalize the plugin and then deprecate (followed by remove) it. cc: @justinsb

camilamacedo86:
Hi @varshaprasad96, 

> @camilamacedo86 I'm fine moving ahead with the PR. Just want to confirm if Kubebuilder declarative pattern maintainers are comfortable with us doing this, or would like to first externalize the plugin and then deprecate (followed by remove) it. cc: @justinsb

I wholeheartedly concur, and I'm also eager to hear their thoughts. However, let me shed some light on my main reasons for pushing this:

- a) We need to continue making changes in the declarative project until we can deprecate and remove these tests and samples, as done in the PR. Without doing so, we run the risk of being blocked from updating k8s and controller runtime. (as we are now)
- b) It's important that we deprecate this as soon as possible. We have a number of deprecations ongoing, and this is the last one needed to clean up Kubebuilder and make its maintenance easier. I believe we should aim to complete all deprecations as soon as possible. This way, we can stop supporting those deprecated features and plan to release Kubebuilder 4.x in line with the k8s 1.28 or 1.29 updates at the latest.

Thank you a lot for all your understanding and input!!!

Kavinjsir (approved):
Nits:
we may also remove:
1. https://github.com/kubernetes-sigs/kubebuilder/blob/7711b8f84067ff04fe4dc5dc0213c76bb8e9d14a/docs/book/src/SUMMARY.md?plain=1#L122
2. https://github.com/kubernetes-sigs/kubebuilder/blob/7711b8f84067ff04fe4dc5dc0213c76bb8e9d14a/docs/book/src/reference/project-config.md?plain=1#L87
3. https://github.com/kubernetes-sigs/kubebuilder/blob/7711b8f84067ff04fe4dc5dc0213c76bb8e9d14a/docs/book/src/plugins/to-be-extended.md?plain=1#L24
4. https://github.com/kubernetes-sigs/kubebuilder/blob/7711b8f84067ff04fe4dc5dc0213c76bb8e9d14a/docs/book/src/plugins/extending-cli.md?plain=1#L61

everettraven (approved):
/lgtm

Kavinjsir (approved):
lgtm 👍🏼

Issue #3393, opened 2023-05-08
Update Webhook For Core Types for CR 0.15


mythi:
### What broke? What's expected?

It looks to me that [the book docs](https://book.kubebuilder.io/reference/webhook-for-core-types.html) is going to get broken in CR v0.15 series.  I got mine fixed by following [this commit](https://github.com/kubernetes-sigs/controller-runtime/commit/927c09e54313d6476b5780b7f029f381d777c149).

Submitting this issue to track the necessary doc updates.

### KubeBuilder (CLI) Version

N/A

### Other versions

`sigs.k8s.io/controller-runtime v0.15.0-alpha.0`

### Extra Labels

/kind documentation

camilamacedo86:
Hi @mythi, 

We start to working in a PR to bump the alpha version and address the breaking changes; see: https://github.com/kubernetes-sigs/kubebuilder/pull/3394

Since you could sort it out accordingly, could you give us a hand and help us to fix this doc? 
WDYT? Can you push a PR within the docs fixes against the master so that it would be fixed and updated in the next release?

camilamacedo86:
Done to be shipped with Kubebuilder 3.11

PR #3399, opened 2023-05-09
:book: update webhooks for core types to match controller-runtime v0.15

mythi:
Fixes: #3393

### DIFF
--- a/docs/book/src/reference/webhook-for-core-types.md
+++ b/docs/book/src/reference/webhook-for-core-types.md
@@ -7,47 +7,57 @@ There is an [example](https://github.com/kubernetes-sigs/controller-runtime/tree
 in controller-runtime.
 
 It is suggested to use kubebuilder to initialize a project, and then you can
-follow the steps below to add admission webhooks for core types.
+follow the steps below to add admission webhooks for core types. The example shows
+how to set up a mutating webhook following the controller-runtime's webhook builder.
 
-## Implement Your Handler
+## Implement Your Webhook
 
-You need to have your handler implements the
-[admission.Handler](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/webhook/admission?tab=doc#Handler)
+You need to have your webhook to implement the
+[admission.CustomDefaulter](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/webhook/admission?tab=doc#CustomDefaulter)
 interface.
 
 ```go
-type podAnnotator struct {
-	Client  client.Client
-	decoder *admission.Decoder
+type podAnnotator struct {}
+
+func (a *podAnnotator) setupWebhookWithManager(mgr ctrl.Manager) error {
+       return ctrl.NewWebhookManagedBy(mgr).
+               For(&corev1.Pod{}).
+               WithDefaulter(a).
+               Complete()
 }
 
-func (a *podAnnotator) Handle(ctx context.Context, req admission.Request) admission.Response {
-	pod := &corev1.Pod{}
-	err := a.decoder.Decode(req, pod)
-	if err != nil {
-		return admission.Errored(http.StatusBadRequest, err)
-	}
+func (a *podAnnotator) Default(ctx context.Context, obj runtime.Object) error {
+    log := logf.FromContext(ctx)
+    pod, ok := obj.(*corev1.Pod)
+    if !ok {
+        return fmt.Errorf("expected a Pod but got a %T", obj)
+    }
 
 	// mutate the fields in pod
+    if pod.Annotations == nil {
+        pod.Annotations = map[string]string{}
+    }
+    pod.Annotations["example-mutating-admission-webhook"] = "foo"
+    log.Info("Annotated pod")
 
-	marshaledPod, err := json.Marshal(pod)
-	if err != nil {
-		return admission.Errored(http.StatusInternalServerError, err)
-	}
-	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+    return nil
 }
 ```
 
 **Note**: in order to have controller-gen generate the webhook configuration for
 you, you need to add markers. For example,
-`// +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io`
+`// +kubebuilder:webhook:path=/mutate--v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io`
+(for webhooks the `path` is of format `/mutate-<group>-<version>-<kind>`. Since this documentation uses `Pod` from the core API group, the group needs to be an empty string).
 
 ## Update main.go
 
 Now you need to register your handler in the webhook server.
 
 ```go
-mgr.GetWebhookServer().Register("/mutate-v1-pod", &webhook.Admission{Handler: &podAnnotator{Client: mgr.GetClient()}})
+    if err := (&podAnnotator{}).setupWebhookWithManager(mgr); err != nil {
+        entryLog.Error(err, "unable to create webhook", "webhook", "Pod")
+        os.Exit(1)
+    }
 ```
 
 You need to ensure the path here match the path in the marker.


camilamacedo86 (commented):
@mythi,

I am not sure if I understand this changes.
See that the Handler still existing for 0.15.0: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/webhook/admission#Handler

The only breaking change that I saw is that now we have `admission.Warnings` in the method signature, see: 

https://github.com/kubernetes-sigs/kubebuilder/pull/3444/files#diff-c157b03fcfe4b2116022e44c782e84f45db09e5e1488e2061c9007e37e634c48R119-R124
 

This change should be in placed in the docs as well for master since after we do the changes we must to run make generate and then the docs are updated.

So, do we still needing this PR?

mythi:
@camilamacedo86 I believe it's still needed. `pkg/inject` is also removed so the example makes the webhook to crash. This moves the book to use the [builder approach following the CR change](https://github.com/kubernetes-sigs/controller-runtime/commit/927c09e54313d6476b5780b7f029f381d777c149#diff-fa148e22912d4a8390270dd209fd4c81fad957622efe2d0dad8a5455d0589201) and [my fixes](https://github.com/kubernetes-sigs/controller-runtime/commit/696c28ca72bc4dc08942fe16c7a2e0dd2c38f341)

mythi:
@camilamacedo86 anything I can/need to do here?

mythi:
> So, can you let us know when you create a project, then api and a webhook what is wrong?

@camilamacedo86 hmm but this PR is not about the scaffolding process but this page https://book.kubebuilder.io/reference/webhook-for-core-types.html which is wrong.

camilamacedo86:
Hi @mythi,

> @camilamacedo86 hmm but this PR is not about the scaffolding process but this page https://book.kubebuilder.io/reference/webhook-for-core-types.html which is wrong.

It seems make sense. When it no longer be a draft could you please either ping in the slack so that we won a help from the community in the review?

camilamacedo86:
@erikgb 

I think you are very skilled in this area, could you please give a hand in the review of this PR?

docs/book/src/reference/webhook-for-core-types.md
```diff
@@ -9,57 +9,49 @@ in controller-runtime.
 It is suggested to use kubebuilder to initialize a project, and then you can
 follow the steps below to add admission webhooks for core types.
 
-## Implement Your Handler
+## Implement Your Webhook
```
erikgb, line 12: This example has now changed from a low-level generic webhook to the next-level mutating webhook (`CustomDefaulter`). I think that is good, but I wonder if we could make that clearer when introducing the example? 🤔 Or maybe not - the example seems to still be around adding annotations to pods.

docs/book/src/reference/webhook-for-core-types.md
```diff
@@ -9,57 +9,49 @@ in controller-runtime.
...
+    log := logf.FromContext(ctx)
 
-	marshaledPod, err := json.Marshal(pod)
-	if err != nil {
-		return admission.Errored(http.StatusInternalServerError, err)
-	}
-	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
-}
-```
+    d, ok := obj.(*corev1.Pod)
```
erikgb, line 26:
```suggestion
    pod, ok := obj.(*corev1.Pod)
```

To make it clearer that you now got a pod struct that can be modified by the user. Maybe also add one line adding a custom annotation - to elaborate more around the example (`podAnnotator`)?

docs/book/src/reference/webhook-for-core-types.md
```diff
@@ -9,57 +9,49 @@ in controller-runtime.
...
-func (a *podAnnotator) InjectDecoder(d *admission.Decoder) error {
-	a.decoder = d
 	return nil
 }
 ```
 
 **Note**: in order to have controller-gen generate the webhook configuration for
 you, you need to add markers. For example,
-`// +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io`
+`// +kubebuilder:webhook:path=/mutate--v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io`
```
erikgb, line 41:
```suggestion
`// +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io`
```

I think this change is non-intentional, or?

docs/book/src/reference/webhook-for-core-types.md
```diff
@@ -9,57 +9,49 @@ in controller-runtime.
...
 
 ## Update main.go
 
 Now you need to register your handler in the webhook server.
 
 ```go
-mgr.GetWebhookServer().Register("/mutate-v1-pod", &webhook.Admission{Handler: &podAnnotator{Client: mgr.GetClient()}})
+    if err := builder.WebhookManagedBy(mgr).
```
erikgb, line 48: Maybe we should add a func to `podAnnotator` to set it up with `Manager` - as we do for controllers?

`func (r *podAnnotator) SetupWithManager(mgr ctrl.Manager) error`

IMO the code will be easier to read. An advanced user can always in-line it...

erikgb (commented):
Thanks @mythi! This looks good! A couple of suggestions/nits, but none of them should block this PR from being merged.

docs/book/src/reference/webhook-for-core-types.md
mythi: It's intentional: with 'core' Group, the group name is empty so the `/mutate-<group>-<version>-<kind>` path becomes `/mutate--v1-pod`. See https://github.com/kubernetes-sigs/controller-runtime/pull/2314

docs/book/src/reference/webhook-for-core-types.md
mythi: updated

docs/book/src/reference/webhook-for-core-types.md
mythi: AFAIU, the same is covered by the new snippet. That is used with:

```
mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{})
...
if err := builder.WebhookManagedBy(mgr).
...
```

docs/book/src/reference/webhook-for-core-types.md
mythi: I clarified this a bit in the intro section.

mythi:
@erikgb @camilamacedo86 the review comments are addressed, and this is ready for another review

docs/book/src/reference/webhook-for-core-types.md
erikgb: I see. I would prefer a note/comment around this in the docs (here). It might not just be me that finds this strange and assume it's non-intentional.

mythi: @erikgb thanks, I added a note in parenthesis about the expected format.

docs/book/src/reference/webhook-for-core-types.md
erikgb: Ok, but if using kubebuilder to create the webhook, I think it will generate a `SetupWithManager` func.

docs/book/src/reference/webhook-for-core-types.md
```diff
@@ -7,59 +7,54 @@ There is an [example](https://github.com/kubernetes-sigs/controller-runtime/tree
...
-	a.decoder = d
 	return nil
 }
 ```
 
 **Note**: in order to have controller-gen generate the webhook configuration for
 you, you need to add markers. For example,
-`// +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io`
+`// +kubebuilder:webhook:path=/mutate--v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io` 
+(for core types the `path` is of format `/mutate--<version>-<kind>`).
```
everettraven, line 44: Nit: I would _personally_ prefer if we mention the whole path format and mention that the group for core types is an empty string. Maybe something like:
```suggestion
(for webhooks the `path` is of format `/mutate-<group>-<version>-<kind>`. Since this documentation focuses on the core types the group is an empty string).
```

everettraven (commented):
Overall the changes look good, just had a couple comments:

docs/book/src/reference/webhook-for-core-types.md
mythi: +1, I'll make that change, thanks

docs/book/src/reference/webhook-for-core-types.md
mythi: ack, will look into it

camilamacedo86 (changes requested):
Requested changes:
- Ensure that the changes in the doc are aligned with the default scaffold done by KB

From @erikgb and @everettraven :

> @erikgb is correct. For example, see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/api/v1/captain_webhook.go
> 
> Even though they do the same thing, I agree that for the Kubebuilder documentation we should ensure we are accurately reflecting what is scaffolded by default by Kubebuilder.

docs/book/src/reference/webhook-for-core-types.md
mythi: this should be covered now. apologies for the silence!

docs/book/src/reference/webhook-for-core-types.md
```diff
@@ -7,47 +7,58 @@ There is an [example](https://github.com/kubernetes-sigs/controller-runtime/tree
...
+follow the steps below to add admission webhooks for core types. The example shows
+how to set up a mutating webhook following the controller-runtime's webhook builder.
 
-## Implement Your Handler
+## Implement Your Webhook
 
-You need to have your handler implements the
-[admission.Handler](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/webhook/admission?tab=doc#Handler)
+You need to have your webhook to implement the
+[admission.CustomDefaulter](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/webhook/admission?tab=doc#CustomDefaulter)
```
camilamacedo86, line 16: @mythi 

All shows fine less this link. We are not implementing the CustomDefaulter right?
That seems that would be a follow up of this one.

We are here just addressing the breaking changes for the old way to do that.

mythi: This PR follows changes made in https://github.com/kubernetes-sigs/controller-runtime/commit/927c09e54313d6476b5780b7f029f381d777c149. That flow expects a `CustomDefaulter` is implemented.

camilamacedo86: Hi @mythi 

```
 WithDefaulter(a).
```

You are passing the Resource Structure itself and not the an structure created for the customer. See that we cannot change the docs without change the scaffold to use the CustomDefault and etc

We have a PR to change the scaffold: https://github.com/kubernetes-sigs/kubebuilder/pull/3723 but it has the same issue. See the example of implementation in controller runtime: https://github.com/kubernetes-sigs/controller-runtime/blob/4000e996a202917ad7d40f02ed8a2079a9ce25e9/pkg/builder/webhook.go#L70-L74

So, WDYT about we broke it in phases

1 - Now, just the update the docs to reflect the current scaffold with the previous breaking changes / deprecations addressed 

Then, when we be scaffolding the Custom we can upgrade the docs to use it.

mythi: > You are passing the Resource Structure itself and not the an structure created for the custom. See that we cannot change the docs without change the scaffold to use the CustomDefault and etc

I'm not sure what this means and that I'd agree with the comment. This doc explicitly says "Given that kubebuilder doesn’t support webhook scaffolding for core types, you have to use the library from controller-runtime to handle it." so I don't see what's the dependency to the rest of the kubebuilder functionality.

This PR merely updates to the controller-runtime example to implement `admission.CustomDefaulter `interface for `podAnnotator` type.

docs/book/src/reference/webhook-for-core-types.md
```diff
@@ -7,47 +7,58 @@ There is an [example](https://github.com/kubernetes-sigs/controller-runtime/tree
...
-	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+	return nil
 }
 ```
 
 **Note**: in order to have controller-gen generate the webhook configuration for
 you, you need to add markers. For example,
-`// +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io`
+`// +kubebuilder:webhook:path=/mutate--v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io`
+(for webhooks the `path` is of format `/mutate-<group>-<version>-<kind>`. Since this documentation focuses on the core types the group is an empty string).
```
camilamacedo86, line 51: The group is not empty for all cases, for example `deployments` is from the group `apps`  example: https://github.com/kubernetes-sigs/kubebuilder/blob/32e0fdc918b9a21485d7b042833402a62d0480f4/docs/kubebuilder_annotation.md?plain=1#L72

You might want to provide the helper to know the group which is in: https://github.com/kubernetes-sigs/kubebuilder/blob/32e0fdc918b9a21485d7b042833402a62d0480f4/pkg/plugins/golang/options.go#L29-L52

docs/book/src/reference/webhook-for-core-types.md
camilamacedo86: We are not using ` admission.CustomDefaulter ` yet in any place but OK. I think you are right we can change this doc and change the scaffolds in another PR.

docs/book/src/reference/webhook-for-core-types.md
```diff
@@ -7,47 +7,58 @@ There is an [example](https://github.com/kubernetes-sigs/controller-runtime/tree
...
+    pod.Annotations["example-mutating-admission-webhook"] = "foo"
+
+    log.Info("Annotated pod")
 
-	marshaledPod, err := json.Marshal(pod)
-	if err != nil {
-		return admission.Errored(http.StatusInternalServerError, err)
-	}
-	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+	return nil
```
camilamacedo86, line 44: before it could return an error because it was patching the value 
so, if an error were faced it will be returned now I am not sure if we we are patching the value.
Did you tested it out?

mythi: > Did you tested it out?

The code is same as in https://github.com/kubernetes-sigs/controller-runtime/blob/76d3d0826fa9dca267c70c68c706f6de40084043/examples/builtins/mutatingwebhook.go so I'm assuming it's tested by controller-runtime. FWIW, my own webhook follows the same flow also.

docs/book/src/reference/webhook-for-core-types.md
mythi: > The group is not empty for all cases, for example `deployments` is from the group `apps` example:

The group is empty for `core` and this is explicitly called out in the sentence as suggested by @erikgb.

docs/book/src/reference/webhook-for-core-types.md
camilamacedo86: It is not the same
Before it was use the client to perform the changes on the cluster side and return its result (nil OR error)
Now, it adds the value to the pod, do nothing and return nil always

docs/book/src/reference/webhook-for-core-types.md
camilamacedo86: That is not true, for deploments the group should be apps 
Such as we use in other markers Example: ` // +rbac:groups=apps,resources=deployments,verbs=get;list;watch;delete `

docs/book/src/reference/webhook-for-core-types.md
mythi: > It is not the same

I'm pretty sure it's the same. I just copied [the `Default()` func](https://github.com/kubernetes-sigs/controller-runtime/blob/76d3d0826fa9dca267c70c68c706f6de40084043/examples/builtins/mutatingwebhook.go) and compared with what is in this file and there are no functional changes.

docs/book/src/reference/webhook-for-core-types.md
mythi: I think our disconnect is with "core types" vs types belonging to `core` API group. With the latter, `<group>` is left empty.

mythi: Apologies for the confusion, I'll reword the text a bit.

docs/book/src/reference/webhook-for-core-types.md
camilamacedo86: But the example that we have here does

```
marshaledPod, err := json.Marshal(pod)
	if err != nil {
		return admission.Errored(http.StatusInternalServerError, err)
	}
	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
```

It would not add too much value just have the sample controller-runtime right?

camilamacedo86: Hi @mythi 

Sure, and we appreciate it a lot.
What I am trying to say is that we should update the doc to keep the same example that we have before but now addressing the controller-runtime changes. That would provide more value than change the current example to copy and paste an example from there.

mythi: > That would provide more value than change the current example to copy and paste an example from there.

What I tried to say was that the current example is also a copy and paste and that the diff here is how controller-runtime addressed their changes in the example. I kinda understand what you're asking me to change but the counterargument is why would we want to diverge from the example given elsewhere (especially since we used to be aligned).

drewwells:
Does this example work? I'm using this with controller-runtime 0.17.0 and my webhook is not being called. The magic around the http handler isn't very straightforward. It would be nice to be able to just setup a handler so we know it's working.

mythi:
> The example code works nicely, but the docs are misleading (to me). They don't align with the code example.

Just to clarify: when you say "example", you mean the controller-runtime example the book points to? This PR is exactly about getting those aligned.

camilamacedo86:
HI @mythi 

> Just to clarify: when you say "example", you mean the controller-runtime example the book points to? This PR is exactly about getting those aligned.

No, I said the exmaple here.
How, we should change this doc?

- Just ensure that the current example provided in the docs (same kubebuilder example) will be changed to address breakchanges
- Ensure that the example provide can properly by creating a project and checking it out

We should not:
- Copy and paste the examples from controller-runtime (it does not bring value for the users, if we want to use their example we can just link it. Make no sense we duplicate things)
- Change the example in away that it does not work with the default kubebuilder layout or not fits well on it.

mythi:
> We should not:

I don't think the current example meets any of these bullets either. However, if that is desired, we could close this and someone can pick up re-writing this page according to the new ask.

camilamacedo86:
Hi @mythi 

See: https://github.com/kubernetes-sigs/kubebuilder/pull/4061
I tried to address those changes.

mythi:
> Hi @mythi
> 
> See: #4061 I tried to address those changes.

Yeah I saw that. I think it has problems with the markers and perhaps mention when to use one over the other. Other than that looks fine

camilamacedo86:
HI @mythi 

> Yeah I saw that. I think it has problems with the markers and perhaps mention when to use one over the other. Other than that looks fine

Not sure if I can follow up.
But please, feel free to review the PR and add your inputs there.
Your collab is very welcome

PR #3406, opened 2023-05-11
🌱 Update the Owner Alias accordingly

camilamacedo86:
## Description

As requested, update the Owner alias accordingly so they know who they should contact when/if required. 

Sadly @jmrodri and @pwittrock have no longer been directly involved, but they are still part of our community and within the permissions granted under emeritus.


OWNERS_ALIASES
```diff
@@ -23,8 +21,10 @@ aliases:
...
   kubebuilder-emeritus-approvers:
+  - adirio
   - directxman12
   - droot
-  - mengqiy
   - estroz
-  - adirio
+  - jmrodri
+  - mengqiy
+  - pwittrock
```
jmrodri, line 30: Awe this makes me sad :( but it makes sense. +1

jmrodri (approved):
/lgtm

OWNERS_ALIASES
camilamacedo86: It makes me too sad. Came back?

PR #3419, opened 2023-05-18
🌱 Add e2e tests for sample external plugin

Eileen-Yu:
## Description:
Add e2e tests for sample external plugin

## Motivation:
Validate the functionality and the behavior of the sample external plugin.

### DIFF
--- a/test/e2e/externalplugin/e2e_suite_test.go
+++ b/test/e2e/externalplugin/e2e_suite_test.go
@@ -1,0 +1,32 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package externalplugin
+
+import (
+	"fmt"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// Run e2e tests using the Ginkgo runner.
+func TestE2E(t *testing.T) {
+	RegisterFailHandler(Fail)
+	fmt.Fprintf(GinkgoWriter, "Starting sample external plugin kubebuilder suite\n")
+	RunSpecs(t, "Kubebuilder grafana plugin e2e suite")
+}
--- a/test/e2e/externalplugin/e2e_suite_test.go
+++ b/test/e2e/externalplugin/e2e_suite_test.go
@@ -28,5 +28,5 @@ import (
 func TestE2E(t *testing.T) {
 	RegisterFailHandler(Fail)
 	fmt.Fprintf(GinkgoWriter, "Starting sample external plugin kubebuilder suite\n")
-	RunSpecs(t, "Kubebuilder grafana plugin e2e suite")
+	RunSpecs(t, "Kubebuilder sample external plugin e2e suite")
 }
--- a/test/e2e/externalplugin/generate_test.go
+++ b/test/e2e/externalplugin/generate_test.go
@@ -1,0 +1,111 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package externalplugin
+
+import (
+	"path/filepath"
+
+	pluginutil "sigs.k8s.io/kubebuilder/v3/pkg/plugin/util"
+
+	//nolint:golint
+	//nolint:revive
+	. "github.com/onsi/ginkgo/v2"
+
+	//nolint:golint
+	//nolint:revive
+	. "github.com/onsi/gomega"
+
+	//nolint:golint
+	//nolint:revive
+	//nolint:golint
+	//nolint:revive
+	"sigs.k8s.io/kubebuilder/v3/test/e2e/utils"
+)
+
+var _ = Describe("kubebuilder", func() {
+	Context("plugin sampleexternalplugin/v1", func() {
+		var (
+			kbc *utils.TestContext
+		)
+
+		BeforeEach(func() {
+			var err error
+			kbc, err = utils.NewTestContext(pluginutil.KubebuilderBinName, "GO111MODULE=on")
+			Expect(err).NotTo(HaveOccurred(), "Prepare NewTestContext should return no error.")
+			Expect(kbc.Prepare()).To(Succeed())
+		})
+
+		AfterEach(func() {
+			kbc.Destroy()
+		})
+
+		It("should generate a runnable project with sample external plugin", func() {
+			GenerateProject(kbc)
+		})
+
+	})
+})
+
+// GenerateProject implements a sampleexternalplugin/v1 external plugin project defined by a TestContext.
+func GenerateProject(kbc *utils.TestContext) {
+	var err error
+
+	By("initializing a project")
+	err = kbc.Init(
+		"--plugins", "sampleexternalplugin/v1",
+		"--domain", "sample.domain.com",
+	)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+	var initFileContentsTmpl = "A simple text file created with the `init` subcommand\nDOMAIN: sample.domain.com"
+	initFileContainsExpr, err := pluginutil.HasFileContentWith(
+		filepath.Join(kbc.Dir, "initFile.txt"), initFileContentsTmpl)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Check initFile.txt should return no error.")
+	ExpectWithOffset(1, initFileContainsExpr).To(BeTrue(), "The init file does not contain the expected expression.")
+
+	By("creating API definition")
+	err = kbc.CreateAPI(
+		"--plugins", "sampleexternalplugin/v1",
+		"--number=2",
+		"--group", kbc.Group,
+		"--version", kbc.Version,
+		"--kind", kbc.Kind,
+	)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+	var apiFileContentsTmpl = "A simple text file created with the `create api` subcommand\nNUMBER: 2"
+	apiFileContainsExpr, err := pluginutil.HasFileContentWith(
+		filepath.Join(kbc.Dir, "apiFile.txt"), apiFileContentsTmpl)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Check apiFile.txt should return no error.")
+	ExpectWithOffset(1, apiFileContainsExpr).To(BeTrue(), "The api file does not contain the expected expression.")
+
+	By("scaffolding webhook")
+	err = kbc.CreateWebhook(
+		"--plugins", "sampleexternalplugin/v1",
+		"--hooked",
+		"--group", kbc.Group,
+		"--version", kbc.Version,
+		"--kind", kbc.Kind,
+	)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+	var webhookFileContentsTmpl = "A simple text file created with the `create webhook` subcommand\nHOOKED!"
+	webhookFileContainsExpr, err := pluginutil.HasFileContentWith(
+		filepath.Join(kbc.Dir, "webhookFile.txt"), webhookFileContentsTmpl)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Check webhookFile.txt should return no error.")
+	ExpectWithOffset(1, webhookFileContainsExpr).To(BeTrue(), "The webhook file does not contain the expected expression.")
+}
--- a/test/e2e/ci.sh
+++ b/test/e2e/ci.sh
@@ -17,6 +17,8 @@
 source "$(dirname "$0")/../common.sh"
 source "$(dirname "$0")/setup.sh"
 
+build_sample_external_plugin
+
 export KIND_CLUSTER="kind"
 create_cluster ${KIND_K8S_VERSION}
 trap delete_cluster EXIT
--- a/test/e2e/local.sh
+++ b/test/e2e/local.sh
@@ -17,6 +17,8 @@
 source "$(dirname "$0")/../common.sh"
 source "$(dirname "$0")/setup.sh"
 
+build_sample_external_plugin
+
 export KIND_CLUSTER="local-kubebuilder-e2e"
 create_cluster ${KIND_K8S_VERSION}
 if [ -z "${SKIP_KIND_CLEANUP:-}" ]; then
--- a/test/e2e/setup.sh
+++ b/test/e2e/setup.sh
@@ -67,3 +67,33 @@ function test_cluster {
   go test $(dirname "$0")/v4 $flags -timeout 30m
   go test $(dirname "$0")/externalplugin $flags -timeout 30m
 }
+
+function build_sample_external_plugin {
+  # TODO: Dynamically set exteranl plugin destination based on OS platform
+  # EXTERNAL_PLUGIN_DESTINATION_PREFIX="${HOME}/Library/Application Support/kubebuilder/plugins"
+  # For Linux:
+  XDG_CONFIG_HOME="${HOME}/.config"
+  EXTERNAL_PLUGIN_DESTINATION_PREFIX="$XDG_CONFIG_HOME/kubebuilder/plugins"
+
+  PLUGIN_NAME="sampleexternalplugin"
+  PLUGIN_VERSION="v1"
+  EXTERNAL_PLUGIN_DESTINATION="${EXTERNAL_PLUGIN_DESTINATION_PREFIX}/${PLUGIN_NAME}/${PLUGIN_VERSION}"
+  EXTERNAL_PLUGIN_PATH="${EXTERNAL_PLUGIN_DESTINATION}/${PLUGIN_NAME}"
+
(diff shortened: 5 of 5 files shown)


test/e2e/externalplugin/generate_test.go
```diff
@@ -0,0 +1,114 @@
...
+	err = kbc.Init(
+		"--plugins", "sampleexternalplugin/v1",
+		"--domain", "sample.domain.com",
+	)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+	initFileContainsExpr, err := pluginutil.HasFileContentWith(
+		filepath.Join(kbc.Dir, "initFile.txt"),
+		`A simple text file created with the`+" `"+`init`+"`"+` subcommand
+DOMAIN: sample.domain.com`)
```
everettraven, line 77: Nit: It might be nice to make this a predefined string variable that is used or using standard double quotes for the string. 

The predefined string variable could look something like:
```go
var initFileContentsTmpl = `A simple text file created with the %s subcommand
DOMAIN: sample.domain.com`
var initFileContents = fmt.Sprintf(initFileContentsTmpl, "`init`")
```
This makes it a bit easier to read and understand what the expected contents should be without having the weird string concatenation sequences. I'm not entirely sure the start and end backticks are allowed in a double quoted string in Go but if not I imagine they are escapable characters by doing something like "\\`"

The other valid approach here would probably be to do something along the lines of:
```suggestion
		"A simple text file created with the `init` subcommand\nDOMAIN: sample.domain.com")
```
OR if the double backticks cause a problem:
```suggestion
"A simple text file created with the \`init\` subcommand\nDOMAIN: sample.domain.com")
```

test/e2e/externalplugin/generate_test.go
```diff
@@ -0,0 +1,114 @@
...
+		"--group", kbc.Group,
+		"--version", kbc.Version,
+		"--kind", kbc.Kind,
+	)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+	apiFileContainsExpr, err := pluginutil.HasFileContentWith(
+		filepath.Join(kbc.Dir, "apiFile.txt"),
+		`A simple text file created with the`+" `"+`create api`+"`"+` subcommand
+NUMBER: 2`)
```
everettraven, line 94: Ditto on the formatting of the string here. The main goal is to make it as easy as possible to understand what the expected value is without having to do too much mental string manipulation.

test/e2e/externalplugin/generate_test.go
```diff
@@ -0,0 +1,114 @@
...
+		"--group", kbc.Group,
+		"--version", kbc.Version,
+		"--kind", kbc.Kind,
+	)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+	webhookFileContainsExpr, err := pluginutil.HasFileContentWith(
+		filepath.Join(kbc.Dir, "webhookFile.txt"),
+		`A simple text file created with the`+" `"+`create webhook`+"`"+` subcommand
+HOOKED!`)
```
everettraven, line 111: Ditto

test/e2e/externalplugin/e2e_suite_test.go
```diff
@@ -0,0 +1,32 @@
+/*
+Copyright 2023 The Kubernetes Authors.
```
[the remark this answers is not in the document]
everettraven: As @camilamacedo86 points out, we need to make sure these tests are run as part of our e2e tests. To elaborate a bit further, these tests are run in Prow and we have a particular shell script that is run to configure the environment and execute the tests.

In order to configure these tests properly I think you need to:
1. Add a new `go test ...` command to this shell script function:
https://github.com/kubernetes-sigs/kubebuilder/blob/a6fca1782a7a31b13b78e6eb03168e9d0c56a020/test/e2e/setup.sh#L56-L68
2. Add a new setup function for properly building and installing the sample external plugin in https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/setup.sh
3. Use the new setup function created in 2 in:
    - https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/local.sh
    - https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/ci.sh

everettraven (changes requested):
Thanks @Eileen-Yu for putting this together! For the most part it looks good - great work!
 
 There are just a few nits and a couple more things that I think need to be done to ensure we are including these tests in CI:

test/e2e/externalplugin/generate_test.go
```diff
@@ -0,0 +1,114 @@
...
+		"--version", kbc.Version,
+		"--kind", kbc.Kind,
+	)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+	apiFileContainsExpr, err := pluginutil.HasFileContentWith(
+		filepath.Join(kbc.Dir, "apiFile.txt"),
+		`A simple text file created with the`+" `"+`create api`+"`"+` subcommand
+NUMBER: 2`)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Check apiFile.txt should return no error.")
```
Kavinjsir, line 95: I like the practice here to add description sentence to support debug when error happens.
@camilamacedo86 @everettraven Maybe we can consider applying such to other e2e test specs?

test/e2e/externalplugin/generate_test.go
```diff
@@ -0,0 +1,114 @@
...
+var _ = Describe("kubebuilder", func() {
+	Context("plugin sampleexternalplugin/v1", func() {
+		var (
+			kbc *utils.TestContext
+		)
+
+		BeforeEach(func() {
+			var err error
+			kbc, err = utils.NewTestContext(pluginutil.KubebuilderBinName, "GO111MODULE=on")
+			Expect(err).NotTo(HaveOccurred())
```
Kavinjsir, line 48: Nit: Can we also have a description sentence as the second parameter also?

Kavinjsir (changes requested):

test/e2e/externalplugin/e2e_suite_test.go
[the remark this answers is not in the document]
camilamacedo86: Thank you @everettraven.
That is what I want to share.

test/e2e/externalplugin/generate_test.go
Eileen-Yu: Sounds good! Updated🚀

Makefile
```diff
@@ -140,12 +140,12 @@ test-legacy: ## Run the legacy tests (go/v2) of the testdata directory
...
 
 .PHONY: test-e2e-local
-test-e2e-local: ## Run the end-to-end tests locally
+test-e2e-local: build-external-plugin ## Run the end-to-end tests locally
 	## To keep the same kind cluster between test runs, use `SKIP_KIND_CLEANUP=1 make test-e2e-local`
 	./test/e2e/local.sh
 
 .PHONY: test-e2e-ci
-test-e2e-ci: ## Run the end-to-end tests (used in the CI)`
+test-e2e-ci: build-external-plugin ## Run the end-to-end tests (used in the CI)`
```
camilamacedo86, line 148: Revert this change you added it in the prow already

test/e2e/build-sample-external-plugin.sh
```diff
@@ -0,0 +1,46 @@
...
+
+SOURCE_DIR="./docs/book/src/simple-external-plugin-tutorial/testdata/sampleexternalplugin/v1"
+cd $SOURCE_DIR && go build -o $PLUGIN_NAME && mv $PLUGIN_NAME "$EXTERNAL_PLUGIN_PATH"
+
+kubebuilder init --plugins="${PLUGIN_NAME}/${PLUGIN_VERSION}" --help | grep --quiet "$PLUGIN_NAME"
+
+if [ $? -eq 0 ]; then
+    echo "External plugin is built at: $EXTERNAL_PLUGIN_DESTINATION"
+else
+    echo "FAIL to build external plugin"
```
rashmigottipati, line 44: It would be nice to echo the exact error message as well, with which `init` command has failed.

test/e2e/build-sample-external-plugin.sh
```diff
@@ -0,0 +1,46 @@
...
+    rm "$EXTERNAL_PLUGIN_PATH" 
+  fi
+else
+    mkdir -p "$EXTERNAL_PLUGIN_DESTINATION"
+fi
+
+SOURCE_DIR="./docs/book/src/simple-external-plugin-tutorial/testdata/sampleexternalplugin/v1"
+cd $SOURCE_DIR && go build -o $PLUGIN_NAME && mv $PLUGIN_NAME "$EXTERNAL_PLUGIN_PATH"
+
+kubebuilder init --plugins="${PLUGIN_NAME}/${PLUGIN_VERSION}" --help | grep --quiet "$PLUGIN_NAME"
```
everettraven, line 39: Nice touch, but this introduces a dependency on the `kubebuilder` binary being in your PATH when you run this script. Ideally the build script will not use something in the PATH but rather a location we can control during the build. There are a couple options here:
1. We remove this line and don't test that kubebuilder can initialize with this plugin as part of building the plugin
2. We build a local version of the kubebuilder binary from source to use for testing this as part of this script

test/e2e/externalplugin/generate_test.go
```diff
@@ -0,0 +1,111 @@
...
+		})
+
+	})
+})
+
+// GenerateProject implements a sampleexternalplugin/v1 external plugin project defined by a TestContext.
+func GenerateProject(kbc *utils.TestContext) {
+	var err error
+
+	By("initializing a project")
```
camilamacedo86, line 67: See that before init you need to ensure that the plugin was built. 
It is not better build the plugin here instead of create the sh?
Could we not move the code to here and to do that in go so that is easier to keep maintained as well? 

See that you can run it locally either if you move all to here by just trigging the tests from your IDE

Makefile
everettraven: I think this is still needed. IIUC, just the execution of the go tests were added in the Prow configuration

test/e2e/build-sample-external-plugin.sh
everettraven: +1

test/e2e/build-sample-external-plugin.sh
everettraven: Looks like we already have a shell function for building kubebuilder here: https://github.com/kubernetes-sigs/kubebuilder/blob/804fed4839718539a6622bd04e910ec1b86f6be4/test/common.sh#L104

test/e2e/build-sample-external-plugin.sh
camilamacedo86: IMO, ideally, we need to keep all in Go.
Why? It makes it easier for us to keep the project maintained.
- See that using an IDE, we can trigger the tests directly, which means the sh would not be executed.
- Also, it makes it more intuitive since all the required steps to do the tests are placed together.
- By leasing, keeping it in Go with the test help us to troubleshot since we can add breakpoints to verify each step

test/e2e/build-sample-external-plugin.sh
```diff
@@ -0,0 +1,46 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+# TODO: Dynamically set exteranl plugin destination based on OS platform
+EXTERNAL_PLUGIN_DESTINATION_PREFIX="${HOME}/Library/Application Support/kubebuilder/plugins"
```
everettraven, line 20: This is making me wonder if it would be worth implementing the functionality to configure where kubebuilder should search for external plugins sooner rather than later. I'm not a huge fan of having e2e tests making modifications to a HOME directory.

test/e2e/externalplugin/generate_test.go
everettraven: I don't disagree that it might be easier to maintain, but I think there is a lot of precedence to build stuff in a shell script for the e2e tests. IMO we shouldn't try to reinvent the wheel by adding this to the go code and follow the existing pattern for building and configuring e2e tests with shell scripts.

test/e2e/build-sample-external-plugin.sh
everettraven, line 1: I think the shell script approach is good, however I'm not sure this needs its own shell script. Taking a look at the overall `test/` directory, there is a [common.sh](https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/common.sh) file that contains a bunch of shell functions to build and configure things for e2e tests and then are used like:
https://github.com/kubernetes-sigs/kubebuilder/blob/804fed4839718539a6622bd04e910ec1b86f6be4/test/e2e/setup.sh#L17-L19

My recommendation would be that we try to follow the existing pattern as much as possible

everettraven: For clarification, my recommendation is that we:
- Make this shell script a new function in the common.sh file
- call this function in the `test/e2e/setup.sh` script similar to the snippet I shared above

test/e2e/externalplugin/generate_test.go
camilamacedo86: Hi @everettraven,

> but I think there is a lot of precedence to build stuff in a shell script for the e2e tests.
We do not have any shell that does things which are required for the e2e tests.
You can trigger locally the e2e suite and you will see that you just need to have any k8s up and running to so.

We have shell for generate the samples and a few of them to setup the prow, which does not seems to few in the same scenario.

If @Eileen-Yu make it working using the shell (besides I do not like that because will broke the possibility to run the e2e tests in dev as described above) I am OK with we get it merged and change after in a follow up. However, see that it is not working and seems exactly because those steps are not performed on the e2e test. So, when the init command is called it fails, see:

```
     s: "kubebuilder init --plugins sampleexternalplugin/v1 --domain sample.domain.com failed with error: (exit status 1) Error: no plugin could be resolved with key \"sampleexternalplugin/v1\"\nUsage:\n  kubebuilder [flags]\n\nExamples:\nThe first step is to initialize your project:\n    kubebuilder init [--plugins=<PLUGIN KEYS> [--project-version=<PROJECT VERSION>]]\n\n<PLUGIN KEYS> is a comma-separated list of plugin keys from the following table\nand <PROJECT VERSION> a supported project version for these plugins.\n\n                             Plugin keys | Supported project versions\n-----------------------------------------+----------------------------\n               base.go.kubebuilder.io/v3 |                          3\n               base.go.kubebuilder.io/v4 |                          3\n        declarative.go.kubebuilder.io/v1 |                       2, 3\n deploy-image.go.kubebuilder.io/v1-alpha |                          3\n                    go.kubebuilder.io/v2 |                       2, 3\n                    go.kubebuilder.io/v3 |                          3\n                    go.kubebuilder.io/v4 |                          3\n         grafana.kubebuilder.io/v1-alpha |                          3\n      kustomize.common.kubebuilder.io/v1 |                          3\n      kustomize.common.kubebuilder.io/v2 |                          3\n\nFor more specific help for the init command of a certain plugins and project version\nconfiguration please run:\n    kubebuilder init --help --plugins=<PLUGIN KEYS> [--project-version=<PROJECT VERSION>]\n\nDefault plugin keys: \"go.kubebuilder.io/v4\"\nDefault project version: \"3\"\n\n\nFlags:\n  -h, --help                     help for kubebuilder\n      --plugins strings          plugin keys to be used for this subcommand execution\n      --project-version string   project version (default \"3\")\n\n2023/05/24 13:03:58 no plugin could be resolved with key \"sampleexternalplugin/v1\"\n",
      }
  occurred
```

Makefile
camilamacedo86: The prow setup is done in this one: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/setup.sh#L56-L67

We do not call any of our tests in this way, so why we would need to have this new target and call in the test-e2e-ci?

test/e2e/build-sample-external-plugin.sh
Eileen-Yu: I agree it is a good practice  to follow the current pattern, which lets Ginkgo do all e2e related tasks, instead of having an additional script. By doing this we can keep consistency and benefit from Ginkgo's feature.

The current script may be a reference to put in the docs to provide users info about how to install their external plugin.

test/e2e/build-sample-external-plugin.sh
Eileen-Yu: I completely agree with your assessment. 

Modifying the HOME directory during e2e testing is indeed not an ideal scenario, and it may lead to unforeseen issues or conflicts.

I think allowing users to configure the path for searching external plugins would significantly increase the flexibility and control over the system. 

Wondering if it's possible to read a configuration file or an environment variable provided by the user, to specify the desired path for the plugin search? Hope in this way we could avoid making assumptions about the user's directory structure or enforcing unnecessary constraints.

test/e2e/externalplugin/generate_test.go
Kavinjsir: Hey @camilamacedo86 , I'm trying to follow your idea:

Since `test/e2e/utils` provides context to bring rich prerequisites, including `k8s cluster`, `kubectl`, `kubebuilder`.
That is how e2e tests get performed for internal plugins.

So, it is not necessary/consistent to deal with prerequisites on a different pattern, such as running a new script.

Is that something you mentioned of and external plugin should NOT break?
> We currently do not utilize any shell scripting to accomplish the tasks required for our end-to-end (e2e) tests.

---

Here is my opinion:
We have a well-structured directory under `test/e2e`:
1. those single files such as `local.sh`, `setup.sh` are used to provide the general environment for e2e testing
2. plugin related stuff are put into separate files such as `v4`, `deployimage`

To maintain that way, when we create a new e2e test suite for external-plugin, it is not necessary/recommended to touch those public files.

Then, if building and configuring external-plugin is needed, we can follow what we did to other plugins:
1. At the very beginning in the ginkgo context, use go code to build and configure external-plugin
2. At the end of the ginkgo context, delete the temporary asset for external-plugin

test/e2e/build-sample-external-plugin.sh
Kavinjsir: One approach maybe:
Introduce an env var `KB_PLUGIN_PATH` when executing the kb command to tell the path of the external plugin.
```
KB_PLUGIN_PATH="/opt/app/kubebuilder/plugin" \
  kubebuilder create api --plugins=myexternalplugin/v1
```
If such env does not exist, fall back to the default (currently used) path.

camilamacedo86, line 20: Will this path exist in the prow ? @everettraven @rashmigottipati ?

Kavinjsir (approved):
lgtm

nits: better to set the external plugin path flexibly.

everettraven (approved):
This looks great! Awesome work @Eileen-Yu !

/lgtm

camilamacedo86 (approved):
Great work 🥇 
We can improve it in follow ups.

That is a really amazing job

Issue #3426, opened 2023-05-22
[Docs/book] running webhook (config/default/kustomization.yaml) not working


manavrajvanshi:
### What broke? What's expected?

I was following the kubebuilder book to add a validation webhook to my operator project (scaffolded using operator-sdk). 
The docs at [kubebuilder book](https://book.kubebuilder.io/cronjob-tutorial/running-webhook.html) says that "You need to enable the webhook and cert manager configuration through kustomize. config/default/kustomization.yaml should now look like the following:"
```# Adds namespace to all resources.
namespace: project-system

# Value of this field is prepended to the
# names of all resources, e.g. a deployment named
# "wordpress" becomes "alices-wordpress".
# Note that it should also match with the prefix (text before '-') of the namespace
# field above.
namePrefix: project-

# Labels to add to all resources and selectors.
#labels:
#- includeSelectors: true
#  pairs:
#    someName: someValue

resources:
- ../crd
- ../rbac
- ../manager
# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
# crd/kustomization.yaml
- ../webhook
# [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'. 'WEBHOOK' components are required.
- ../certmanager
# [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
- ../prometheus

patchesStrategicMerge:
# Protect the /metrics endpoint by putting it behind auth.
# If you want your controller-manager to expose the /metrics
# endpoint w/o any authn/z, please comment the following line.
- manager_auth_proxy_patch.yaml



# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
# crd/kustomization.yaml
#- manager_webhook_patch.yaml

# [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'.
# Uncomment 'CERTMANAGER' sections in crd/kustomization.yaml to enable the CA injection in the admission webhooks.
# 'CERTMANAGER' needs to be enabled to use ca injection
- webhookcainjection_patch.yaml

# [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER' prefix.
# Uncomment the following replacements to add the cert-manager CA injection annotations
replacements:
[...]

PR #3427, opened 2023-05-22
📖 Fixed manager_webhook_patch not uncommented

lauchokyip:
Fixes https://github.com/kubernetes-sigs/kubebuilder/issues/3426

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
@@ -36,7 +36,7 @@ patchesStrategicMerge:
 
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
 # crd/kustomization.yaml
-#- manager_webhook_patch.yaml
+- manager_webhook_patch.yaml
 
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'.
 # Uncomment 'CERTMANAGER' sections in crd/kustomization.yaml to enable the CA injection in the admission webhooks.
--- a/hack/docs/internal/cronjob-tutorial/generate_cronjob.go
+++ b/hack/docs/internal/cronjob-tutorial/generate_cronjob.go
@@ -571,6 +571,11 @@ func updateKustomization(sp *Sample) {
 
 	err = pluginutil.UncommentCode(
 		filepath.Join(sp.ctx.Dir, "config/default/kustomization.yaml"),
+		`#- manager_webhook_patch.yaml`, `#`)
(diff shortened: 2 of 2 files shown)


docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
```diff
@@ -36,7 +36,7 @@ patchesStrategicMerge:
 
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
 # crd/kustomization.yaml
-#- manager_webhook_patch.yaml
+- manager_webhook_patch.yaml
```
camilamacedo86, line 39: That is a very good catcher but we need to fix it in the hack/ files which generate the sample mocks
Otherwise, it will not pass in the CI since this change will be overwriting when we run `make generate` and it calls the make generate-docs to automatically update the samples under docs with the latest changes

lauchokyip: Done

camilamacedo86 (approved):
/ok-to-test

docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
manavrajvanshi, line 39: Since you've already made the changes in the file `hack/docs/internal/cronjob-tutorial/generate_cronjob.go`, do we really need the change here in this file? Asking because this will by overwritten by `make generate` (and hence make generate-docs) as mentioned by @camilamacedo86 .

manavrajvanshi: As I understand, `make generate` will overwrite the file `docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml`, right? 

If so, then do we explicitly need to make this change here? Else, this can be skipped. Thanks!

lauchokyip: Please take a look at https://github.com/kubernetes-sigs/kubebuilder/actions/runs/5059859019/jobs/9085111581?pr=3427 for the `Verify docs` session

manavrajvanshi (approved):

camilamacedo86 (approved):
/lgtm

manavrajvanshi:
@camilamacedo86 , I see the [issue](https://github.com/kubernetes-sigs/kubebuilder/issues/3426) I reported has been closed since this PR has been merged. Just curious to know how long does it take for the changes to reflect on the Kubebuilder book?
Asking because I still see the comment `#- manager_webhook_patch.yaml`.

PR #3430, opened 2023-05-23
✨ (go/v4): Allows kubectl to be configured in Makefile

redhatrises:
- This allows multiple different kubectl installs to be used.


camilamacedo86 (approved):
/lgtm
/approved

I have no objections with that at all.
We just need to keep in mind that if users change the config then, it might fails because it is not something that we can test. However, the same is valid for any customization.

redhatrises:
> /lgtm /approved
> 
> I have no objections with that at all. We just need to keep in mind that if users change the config then, it might fails because it is not something that we can test. However, the same is valid for any customization.

@camilamacedo86 Thanks! Agreed. Would you mind re-reviewing? I forgot to update documentation. Thanks!

Issue #3402, opened 2023-05-10
Kubernetes and Kubebuilder use different mechanisms to generate plurals


mrueg:
### What broke? What's expected?

Kubernetes uses a custom mechanism to guess and generate plurals for CRDs in https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/restmapper.go#L126

Kubebuilder uses https://github.com/gobuffalo/flect which has its own custom set of rules. https://github.com/kubernetes-sigs/kubebuilder/blob/6c11f05e39d25767a953878255ffd6f7d24822c5/pkg/model/resource/utils.go#L72

Although one could argue that the method kubernetes uses is broken and not ideal, it would be great if kubebuilder could apply the same default mechanism to generate plurals so it is easier for kubernetes clients to resolve a resourceplural from a GroupVersionKind mapping.

### KubeBuilder (CLI) Version

3.10.0

camilamacedo86:
It has a proposed solution in : https://github.com/kubernetes-sigs/kubebuilder/pull/3408
It seems very reasonable ask.

lauchokyip:
There are still some discussions happening on this thread https://github.com/kubernetes/kube-state-metrics/pull/2065, should we wait until the consensus is made ?

rexagod:
Boiling down that thread, the overall ask is to have a de-facto way of pluralizing that helps keep the resource names consistent throughout the ecosystem. I believe using the same logic as apimachinery should allow us to achieve that.

Kavinjsir:
Hey @mrueg @rexagod , would you provide some cases/code snippets that come with such an attempt to make the plural consistent? I'm trying to understand how the consistent plural system can help *kubernetes clients to resolve a resourceplural from a GroupVersionKind mapping.*

rexagod:
We've had to [extract](https://github.com/kubernetes/kube-state-metrics/pull/1851/files#diff-916e6863e1245c673b4e5965c98dc27bafbd72650fdb38ce65ea73ee6304e027R58) and [match](https://github.com/kubernetes/kube-state-metrics/pull/1851/files#diff-916e6863e1245c673b4e5965c98dc27bafbd72650fdb38ce65ea73ee6304e027R133) plurals to avoid instances where there may be [discrepancies](https://github.com/open-policy-agent/gatekeeper/issues/2751) between the expected and actual plurals.

rexagod:
[...]

PR #3433, opened 2023-05-27
🐛 Synchonized plural generation with api-machinery

lauchokyip:
Fixes https://github.com/kubernetes-sigs/kubebuilder/issues/3402


camilamacedo86 (commented):
/ok-to-test

camilamacedo86:
@lauchokyip 

Could you please squash the commits ?

@varshaprasad96 @everettraven wdyt about this one?
Could you please give a hand on the review?

A-Hilaly (commented):
+1 for trying to keep apimachinery and kubebuilder synced.

Quick question: this will probably break existing projects that already benefited from `gobuffalo/flect` pluralize mechanism. Maybe we should allow users to pass a set of "override" plural names (via flag, env/var...?), so at least they can work around this change?

lauchokyip:
/hold 

Definitely feel like a breaking change

camilamacedo86 (commented):
Hi @lauchokyip,

Could you please squash the commits?
Also, see that it is not passing in the lint you can run locally make lint locally to check it out. 

**By last, we need to provide a guidance for those that:**
- Scaffold the project with an previous version, create an api and then get a new kubebuilder version to create another API

It seems for me that would not work at all and this one is a breaking change. For other hand it seems a valid / reasonable request. I am just trying to think if it could be addressed for v4 or if we should only perform this change for an future version like go/v5.


c/c @everettraven @varshaprasad96 @Kavinjsir  wdyt about this one?

pkg/model/resource/utils.go
```diff
@@ -67,7 +65,29 @@ func APIPackagePathLegacy(repo, group, version string, multiGroup bool) string {
 	return path.Join(repo, "api", version)
 }
 
+// unpluralizedSuffixes is a list of resource suffixes that are the same plural and singular
+// This is only is only necessary because some bits of code are lazy and don't actually use the RESTMapper like they should.
+// TODO eliminate this so that different callers can correctly map to resources.  This probably means updating all
+// callers to use the RESTMapper they mean.
```
Kavinjsir, line 71: I think we can just leave a reference as well, which could be enough.
(This should also resolve the lint error as well)

image

Kavinjsir (commented):
Thx for your contribution @lauchokyip ! Just give a first-round review and left a nit.

@camilamacedo86 
I'm Okay to make it consistent with the upstream kb plural system, though I personally find it hacky.  Basically, the upstream plural system looks far from 'perfect':
1. The k8s plural implementation itself is kind of broken, like `helmswomans` and `fishs`. Also, the implementation looks tricky, where it maintains a hardcoded `unpluralizedSuffixes`.
2. Like as you mentioned, if users are migrating their operator projects to the new kubebuilder with such an updated plural system. It may probably bring breaking change. (Though, it might not be a common scenario.)

Also in the future, we may probably change it once again, as such plural system might not be the 'ultimate' solution. (And then users might meet the similar scenario one more time.)

pkg/model/resource/utils.go
```diff
@@ -67,7 +65,29 @@ func APIPackagePathLegacy(repo, group, version string, multiGroup bool) string {
...
+var unpluralizedSuffixes = []string{
+	"endpoints",
+}
+
```
everettraven, line 75: Do we actually need this? `Endpoints` is a core Kubernetes resource and we likely aren't going to have to differentiate this in kubebuilder.

pkg/model/resource/utils.go
```diff
@@ -67,7 +65,29 @@ func APIPackagePathLegacy(repo, group, version string, multiGroup bool) string {
...
 // RegularPlural returns a default plural form when none was specified
+// Reference: https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/restmapper.go#L126
 func RegularPlural(singular string) string {
-	return flect.Pluralize(strings.ToLower(singular))
+	singularName := strings.ToLower(singular)
+	for _, skip := range unpluralizedSuffixes {
+		if strings.HasSuffix(singularName, skip) {
+			return singular
+		}
+	}
```
everettraven, line 84: We can probably remove this as per my previous comment

everettraven (changes requested):
Thanks for the contribution @lauchokyip ! I have a couple suggested changes:

pkg/model/resource/utils.go
```diff
@@ -67,7 +65,29 @@ func APIPackagePathLegacy(repo, group, version string, multiGroup bool) string {
...
+
+	switch string(singularName[len(singularName)-1]) {
+	case "s":
+		return singularName + "es"
+	case "y":
+		return strings.TrimSuffix(singularName, "y") + "ies"
```
varshaprasad96, line 90: There is a piece of code in k8s/code-generator, that deals with this and some of the edge cases explicitly: https://github.com/kubernetes/gengo/blob/ab3349d207d4/namer/plural_namer.go#L36. May not be an immediate action item, but it would be preferable to have a TODO added here, to be able to resuse that piece of code. (It currently is very opinionated to spit out plurals for core k8s resources).

lauchokyip:
> Hi @lauchokyip,
> 
> Could you please squash the commits? Also, see that it is not passing in the lint you can run locally make lint locally to check it out.
> 
> **By last, we need to provide a guidance for those that:**
> 
> * Scaffold the project with an previous version, create an api and then get a new kubebuilder version to create another API
>   Could you please add the steps in the description of this PR? What guidance/steps should we provide in this case?
> 
> It seems for me that would not work at all and this one is a breaking change. For other hand it seems a valid / reasonable request. I am just trying to think if it could be addressed for v4 or if we should only perform this change for an future version like go/v5.
> 
> c/c @everettraven @varshaprasad96 @Kavinjsir wdyt about this one?

Hi @camilamacedo86 , do you providing the migration documents? Any guidance on how I could do that?  I don't know the code well enough to answer this question - if the user generate a CRD named Helmswoman and the plural is set to `Helmswomen`, and if the user were to generate another CRD for different version `v2`, kind `Helmswoman`, would the plural be set to `Helmswomans` ? If the plural were to set to `Helmswomans`, will that break anything besides the generated file name?

everettraven (approved):
Code looks good to me, thanks @lauchokyip !

camilamacedo86:
Hi @everettraven @varshaprasad96 @Kavinjsir @lauchokyip,

We have encountered several deprecations, and it's essential that we deprecate the declarative plugin as soon as possible. You can find the details in this pull request: https://github.com/kubernetes-sigs/kubebuilder/pull/3395

To ensure a cleaner and more maintainable codebase as we continue to grow, we plan to remove all deprecations at the beginning of next year, leading to a major bump for Kubebuilder 4.0.0.

However, since this change will have a significant impact on projects that are already scaffolded, I suggest we consider the following:

- Let's carefully evaluate whether this change is necessary, taking into account the comments shared by @Kavinjsir in this pull request: https://github.com/kubernetes-sigs/kubebuilder/pull/3433#pullrequestreview-1453231505.
- In the pull request, it would be helpful to document the manual steps required to update existing projects once this change is merged. @lauchokyip, could you please include these steps in the PR description? We need to provide guidance on how to modify projects built prior to this change, ensuring they won't be affected when a new version of Kubebuilder is released. Please specify the necessary file modifications and steps to be taken. (It is essential because it will help us properly document it in the release and migration guides)
- @lauchokyip, please ensure that the new code is only used for go/v4 and doesn't impact go/v3, which is already deprecated. By doing so, when we eventually bump to 4.x, the condition will keep the code as the default. 
- Then, I think we might want to consider whether to keep this PR for future go/v5 changes. Once @lauchokyip implements the changes for impact only go/v4 as described above, we can modify the PR for go/v5 easily and merge it at the appropriate time.

Please share your thoughts on these suggestions.

lauchokyip:
> @lauchokyip We are interested in resolving this issue. Could you please rebase this so that we revisit it again. Thanks for taking the time for working on this.

Let me get back to this in a bit

lauchokyip:
HI @varshaprasad96 , `make generate` are failing, can you please check it? I even `git clone` a brand new repo and it still failed
```
TODO: update tutorial
ERRO[0009] error fixing cronjob_webhook.go: unable to find the content to be replaced 
exit status 1
make: *** [Makefile:76: generate-docs] Error 1
```

Kavinjsir:
> HI @varshaprasad96 , `make generate` are failing, can you please check it? I even `git clone` a brand new repo and it still failed
> 
> ```
> TODO: update tutorial
> ERRO[0009] error fixing cronjob_webhook.go: unable to find the content to be replaced 
> exit status 1
> make: *** [Makefile:76: generate-docs] Error 1
> ```
> 
> Does `make generate` work for you?

@lauchokyip Would you rebase the branch based off the latest master branch? There might be an issue for `make generate` previously. The current master branch should have it fixed. Could you try and see if it works? Feel free to let me know if further questions rasied.

camilamacedo86:
Closed based on:
- Huge breaking change https://github.com/kubernetes-sigs/kubebuilder/pull/3433#issuecomment-1585538375 without a good justification. Why the change is not proposed to apimachinery?
- https://github.com/kubernetes-sigs/kubebuilder/issues/3402#issuecomment-2121490772
- https://github.com/kubernetes-sigs/kubebuilder/issues/3402#issuecomment-2122171734

PR #3437, opened 2023-06-01
✨ Make external plugin path configurable

Eileen-Yu:
## Description
1. Introduce the usage of env variable `KB_PLUGINS_PATH` to define a customized path to store external plugins.
1. Add corresponding unit tests

## Sample

```
~/programming/temp/my-external-plugins  
❯ tree .
.
└── sampleexternalplugin
   └── v1
      └── sampleexternalplugin

~/programming/temp/my-external-plugins                                                                                                                                                
❯ KB_PLUGIN_PATH="/Users/eileen/programming/temp/my-external-plugins" \
  kubebuilder init --plugins=sampleexternalplugin/v1
```

## Motivation
Make the external plugin path configurable so that users can easily manage their external plugins.

### DIFF
--- a/pkg/cli/options.go
+++ b/pkg/cli/options.go
@@ -181,13 +181,29 @@ func isHostSupported(host string) bool {
 	return false
 }
 
-// getPluginsRoot detects the host system and gets the plugins root based on the host.
+// getPluginsRoot gets the plugin root path.
 func getPluginsRoot(host string) (pluginsRoot string, err error) {
 	if !isHostSupported(host) {
 		// freebsd, openbsd, windows...
 		return "", fmt.Errorf("host not supported: %v", host)
 	}
 
+	// if user provides specific path, return
+	if pluginsPath := os.Getenv("EXTERNAL_PLUGINS_PATH"); pluginsPath != "" {
+		// verify if the path actually exists
+		if _, err := os.Stat(pluginsPath); err != nil {
+			if os.IsNotExist(err) {
+				// the path does not exist
+				return "", fmt.Errorf("the specified path %s does not exist", pluginsPath)
+			}
+			// some other error
+			return "", fmt.Errorf("error checking the path: %v", err)
+		}
+		// the path exists
+		return pluginsPath, nil
+	}
+
+	// if no specific path, detects the host system and gets the plugins root based on the host.
 	pluginsRelativePath := filepath.Join("kubebuilder", "plugins")
 	if xdgHome := os.Getenv("XDG_CONFIG_HOME"); xdgHome != "" {
 		return filepath.Join(xdgHome, pluginsRelativePath), nil
@@ -206,9 +222,8 @@ func getPluginsRoot(host string) (pluginsRoot string, err error) {
 	if err != nil {
 		return "", fmt.Errorf("error retrieving home dir: %v", err)
 	}
-	pluginsRoot = filepath.Join(userHomeDir, pluginsRoot)
 
-	return
+	return filepath.Join(userHomeDir, pluginsRoot), nil
 }
 
 // DiscoverExternalPlugins discovers the external plugins in the plugins root directory
--- a/pkg/cli/options_test.go
+++ b/pkg/cli/options_test.go
@@ -18,6 +18,7 @@ package cli
 
 import (
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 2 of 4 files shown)


pkg/cli/options_test.go
```diff
@@ -34,6 +35,65 @@ import (
...
 
 var _ = Describe("Discover external plugins", func() {
+	Context("validate plugins root path", func() {
+		var (
+			homePath   string = os.Getenv("HOME")
+			customPath string = "/usr/local/myplugins"
+			xdghome    string = os.Getenv("XDG_CONFIG_HOME")
+		)
+
+		It("based on the host os", func() {
```
everettraven, line 45: Something to consider here is how maintainable and readable these tests are. Often times reading through test cases can help someone get an understanding of the general flow of what is being tested. 

In this particular case, when I look at this from the perspective of someone who has no idea what the logic the test is for this takes me a bit longer to read through the code and understand what is happening. 

My recommendation is to look into using the various Gingko functions to break these tests into much smaller tests. Something like:
```go
...
When("using the $HOME environment variable", func{
...
  It("should return the correct path for the darwin OS", func{...})
  It("should return the correct path for the linux OS", func{...})
...
})
...
```
Following this approach we can make the individual tests significantly smaller and easier to follow, thus making them much easier to maintain.

everettraven (changes requested):
Overall really great work! My biggest request is to take another pass at the tests implementation and try to break them down into individual smaller tests

pkg/cli/options.go
```diff
@@ -181,13 +181,18 @@ func isHostSupported(host string) bool {
...
+// getPluginsRoot gets the plugin root path.
 func getPluginsRoot(host string) (pluginsRoot string, err error) {
 	if !isHostSupported(host) {
 		// freebsd, openbsd, windows...
 		return "", fmt.Errorf("host not supported: %v", host)
 	}
 
+	// if user provides specific path, return
+	if pluginsPath := os.Getenv("KB_PLUGINS_PATH"); pluginsPath != "" {
+		return pluginsPath, nil
```
rashmigottipati, line 193: Might be worth adding a check to validate that the path actually exists unless we are already doing that in the function that calls `getPluginsRoot()`.

pkg/cli/options_test.go
```diff
@@ -34,6 +35,65 @@ import (
...
+				Expect(plgPath).To(Equal(fmt.Sprintf("%s/.config/kubebuilder/plugins", homePath)))
+			}
+
+			plgPath, err = getPluginsRoot("random")
+			Expect(plgPath).To(Equal(""))
+			Expect(err).ToNot(BeNil())
+			Expect(err.Error()).To(ContainSubstring("host not supported"))
+		})
+
+		It("based on the user given path", func() {
```
rashmigottipati, line 73: Nit: make the It descriptions more like full sentences. See the examples in https://onsi.github.io/ginkgo/

camilamacedo86 (commented):
/ok-to-test

camilamacedo86:
HI @Eileen-Yu,

Can we cover a little more with the tests:

> coverage/coveralls — Coverage decreased (-0.1%) to 83.135%

It decrease here: https://coveralls.io/builds/60550613/source?filename=pkg%2Fcli%2Foptions.go

pkg/cli/options.go
```diff
@@ -181,13 +181,29 @@ func isHostSupported(host string) bool {
...
+		if _, err := os.Stat(pluginsPath); err == nil {
+			// the path exists
+			return pluginsPath, nil
+		} else if os.IsNotExist(err) {
+			// the path does not exist
+			return "", fmt.Errorf("the specified path %s does not exist", pluginsPath)
+		} else {
+			// some other error
+			return "", fmt.Errorf("error checking the path: %v", err)
+		}
```
everettraven, line 203: Nit: Overall I think this looks fine, but I personally prefer a format more like:
```suggestion
		if _, err := os.Stat(pluginsPath); err != nil {
		    if os.IsNotExist(err) {
			// the path does not exist
			return "", fmt.Errorf("the specified path %s does not exist", pluginsPath)
		    }
		    // some other error
		    return "", fmt.Errorf("error checking the path: %v", err)
		} 
		// the path exists
		return pluginsPath, nil
```
No need to change this if you don't want to, just figured I would share my 2 cents as I think this condenses it into fewer lines of code that is a bit easier to read

pkg/cli/options_test.go
```diff
@@ -34,6 +35,86 @@ import (
...
+			It("should return the correct path for the darwin OS", func() {
+				plgPath, err := getPluginsRoot("darwin")
+				Expect(err).To(BeNil())
+				if xdghome != "" {
+					Expect(plgPath).To(Equal(fmt.Sprintf("%s/kubebuilder/plugins", xdghome)))
+					// nolint:lll,gosec
+					AddReportEntry(fmt.Sprintf("Detect XDG_CONFIG_HOME: %s, generated plugin path: %s", xdghome, plgPath), ReportEntryVisibilityAlways)
+				} else {
+					Expect(plgPath).To(Equal(fmt.Sprintf("%s/Library/Application Support/kubebuilder/plugins", homePath)))
+				}
```
everettraven, line 60: IMO these should be tests that are separated even further. Ideally we don't have to do this `if ... else` check and the xdghome check is in a separate test case. 

Something we could do to verify this is working appropriately for when $XDGHOME isn't set is to get the original value, unset it before each test, reset the original value after each test. Luckily Ginkgo v2 has a handy `GinkgoT().Setenv(...)` function we can use to do this for us.

everettraven (commented):
Overall looks good and is a step in the right direction. What I would like to see is for the tests for the $XDGHOME and $HOME environment variables to be distinct tests (so no if ... else block to perform special test logic)

camilamacedo86 (commented):
Great work 🥇 

Just a nit: 
Otherwise, /lgtm

Is KB_PLUGINS_PATH the best name for the env var?
What about `EXTERNAL_PLUGINS_PATH` only?
I mean without kubebuilder since other projects can export Kubebuilder and use it as lib to create their own tools. 

c/c @everettraven @rashmigottipati

pkg/cli/options_test.go
```diff
@@ -34,6 +35,167 @@ import (
...
+
+		When("XDG_CONFIG_HOME is not set and using the $HOME environment variable", func() {
+			BeforeEach(func() {
+				err := os.Unsetenv("KB_PLUGINS_PATH")
+				Expect(err).To(BeNil())
+
+				err = os.Unsetenv("XDG_CONFIG_HOME")
+				Expect(err).To(BeNil())
+
+				xdghome = os.Getenv("XDG_CONFIG_HOME")
```
everettraven, line 66: Looks like we can probably remove this line - I don't see anywhere in this context where it is used.

pkg/cli/options_test.go
```diff
@@ -34,6 +35,167 @@ import (
 )
 
 var _ = Describe("Discover external plugins", func() {
+	Context("with valid plugins root path", func() {
+		var (
+			homePath           string = os.Getenv("HOME")
+			customPath         string = "/tmp/myplugins"
+			xdghome            string = os.Getenv("XDG_CONFIG_HOME")
+			originalPluginPath string = os.Getenv("KB_PLUGINS_PATH")
```
everettraven, line 43: Nit: I'd prefer if we did all the `os.Getenv()` calls in the `BeforeEach` for the appropriate context they are used. This will make sure that we are constantly fetching the most up to date value before a test is run.

pkg/cli/options_test.go
```diff
@@ -34,6 +35,167 @@ import (
...
+		AfterEach(func() {
+			if originalPluginPath != "" {
+				// restore the original value
+				err := os.Setenv("KB_PLUGINS_PATH", originalPluginPath)
+				Expect(err).To(BeNil())
+			} else {
+				// unset if it was originally unset
+				err := os.Unsetenv("KB_PLUGINS_PATH")
+				Expect(err).To(BeNil())
+			}
```
everettraven, line 55: We should do this logic for _all_ the environment variables we modify during testing. It looks like the only one we are probably missing here is doing this logic for the `XDG_CONFIG_HOME` variable

pkg/cli/options_test.go
```diff
@@ -34,6 +35,167 @@ import (
...
+				err = os.Setenv("XDG_CONFIG_HOME", fmt.Sprintf("%s/.config", homePath))
+				Expect(err).To(BeNil())
+
+				xdghome = os.Getenv("XDG_CONFIG_HOME")
+			})
+
+			AfterEach(func() {
+				err := os.Unsetenv("XDG_CONFIG_HOME")
+				Expect(err).To(BeNil())
+			})
```
everettraven, line 103: With this before/after each set, what happens if I set my `XDG_CONFIG_HOME` environment variable and then run this test in the same session?

IIUC this would overwrite my current value with the test value and then completely unset it, leaving my `XDG_CONFIG_HOME` variable different from what I set it to before running the tests.

everettraven (commented):
Overall looks great - I just have some more comments on the tests

pkg/cli/options_test.go
Eileen-Yu: Thanks for catching!

pkg/cli/options_test.go
Eileen-Yu: Oh, I see.
I tried to update the code following your suggestions. It seems much clearer than before!

camilamacedo86 (approved):
Terrific work 🥇 

/approved

PR #3374, opened 2023-04-30
✨ (kustomize/v2) Remove deprecated syntax usage by replacing patchesStrategicMerge with patches

Sajiyah-Salat:
Fixes #3372 

 See that the default scaffolds:

Warning:
'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run
"kustomize edit fix' to update
your Kustomization automatically.

Therefore I replaced patchesStrategicMerge by patches in the kustomize/v2 templates code.

PR #3456, opened 2023-06-14
:bug: Fix kustomize syntax during conversion to patches

### LINKED ISSUE #3425: Invalid Kustomize generation

This PR (https://github.com/kubernetes-sigs/kubebuilder/pull/3374) breaks the kustomization files generated. Had to revert back to `patchesStrategicMerge`

To reproduce:
1)
```bash
mkdir testing
cd testing
kubebuilder init --domain tutorial.kubebuilder.io --repo tutorial.kubebuilder.io/project
kubebuilder create api --group batch --version v1 --kind CronJob
kubebuilder create webhook --group batch --version v1 --kind CronJob  --conversion
make manifests
```
2) Go to `/config/crd/kustomization.yaml` and uncomment 
```yaml
patches:
# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix.
# patches here are for enabling the conversion webhook for each CRD
- patches/webhook_in_cronjobs.yaml
#+kubebuilder:scaffold:crdkustomizewebhookpatch

# [CERTMANAGER] To enable cert-manager, uncomment all the sections with [CERTMANAGER] prefix.
# patches here are for enabling the CA injection for each CRD
- patches/cainjection_in_cronjobs.yaml
#+kubebuilder:scaffold:crdkustomizecainjectionpatch
```
3) Run `kustomize` against it
```
kustomize build config/crd
Error: invalid Kustomization: json: cannot unmarshal string into Go struct field Kustomization.patches of type types.Patc
```

### Reproducing this issue

_No response_

### KubeBuilder (CLI) Version

Version: main.version{KubeBuilderVersion:"v3.10.0-57-g959952de-dirty", KubernetesVendor:"unknown", GitCommit:"959952decffecb9a35bb843e706849586883360c", BuildDate:"2023-05-16T02:27:25Z", GoOs:"linux", GoArch:"amd64"}

### PROJECT version

_No response_

### Plugin versions

_No response_

### Other versions

_No response_

### Extra Labels

_No response_

mjlshen:
This commit fixes the syntax error introduced in #3374 where pathcesStrategicMerge was replaced with patches, which now requires that every patch additionally have a "path" key when multiple patches are specified in a file.

Fixes #3425

### DIFF
--- a/docs/book/src/component-config-tutorial/testdata/project/config/crd/kustomization.yaml
+++ b/docs/book/src/component-config-tutorial/testdata/project/config/crd/kustomization.yaml
@@ -8,12 +8,12 @@ resources:
 patches:
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix.
 # patches here are for enabling the conversion webhook for each CRD
-#- patches/webhook_in_projectconfigs.yaml
+#- path: patches/webhook_in_projectconfigs.yaml
 #+kubebuilder:scaffold:crdkustomizewebhookpatch
 
 # [CERTMANAGER] To enable cert-manager, uncomment all the sections with [CERTMANAGER] prefix.
 # patches here are for enabling the CA injection for each CRD
-#- patches/cainjection_in_projectconfigs.yaml
+#- path: patches/cainjection_in_projectconfigs.yaml
 #+kubebuilder:scaffold:crdkustomizecainjectionpatch
 
 # the following config is for teaching kustomize how to do kustomization for CRDs.
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/crd/kustomization.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/crd/kustomization.yaml
@@ -8,12 +8,12 @@ resources:
 patches:
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix.
 # patches here are for enabling the conversion webhook for each CRD
-- patches/webhook_in_cronjobs.yaml
+- path: patches/webhook_in_cronjobs.yaml
 #+kubebuilder:scaffold:crdkustomizewebhookpatch
 
 # [CERTMANAGER] To enable cert-manager, uncomment all the sections with [CERTMANAGER] prefix.
 # patches here are for enabling the CA injection for each CRD
-- patches/cainjection_in_cronjobs.yaml
+- path: patches/cainjection_in_cronjobs.yaml
 #+kubebuilder:scaffold:crdkustomizecainjectionpatch
 
 # the following config is for teaching kustomize how to do kustomization for CRDs.
--- a/hack/docs/internal/cronjob-tutorial/generate_cronjob.go
+++ b/hack/docs/internal/cronjob-tutorial/generate_cronjob.go
@@ -581,12 +581,12 @@ func updateKustomization(sp *Sample) {
 	// uncomment crd/kustomization
 	err = pluginutil.UncommentCode(
 		filepath.Join(sp.ctx.Dir, "config/crd/kustomization.yaml"),
-		`#- patches/webhook_in_cronjobs.yaml`, `#`)
+		`#- path: patches/webhook_in_cronjobs.yaml`, `#`)
 	CheckError("fixing crd/kustomization", err)
 
 	err = pluginutil.UncommentCode(
 		filepath.Join(sp.ctx.Dir, "config/crd/kustomization.yaml"),
-		`#- patches/cainjection_in_cronjobs.yaml`, `#`)
+		`#- path: patches/cainjection_in_cronjobs.yaml`, `#`)
 	CheckError("fixing crd/kustomization", err)
 }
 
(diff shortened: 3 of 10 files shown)


docs/book/src/cronjob-tutorial/testdata/project/go.sum
mjlshen, line 1: I didn't intend to make these changes, but they appeared after I ran `make generate`

lauchokyip (approved):
/lgtm

docs/book/src/cronjob-tutorial/testdata/project/go.sum
camilamacedo86: Yes, probably when we bumped some dependency using the automated PRs

camilamacedo86 (approved):
It is great thank you for add this fix.

Would be amazing if we could get it in the e2e tests. 
See that we generate a project with webhooks and uncommenting all options here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v4/generate_test.go we have an e2e test using webhooks.

I think it is missing we add here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v4/plugin_cluster_test.go#L106-L108 a line to call make manifests and make generate and see if it does not finish with error.

WDYT? If you be enable to add in this PR could you please give a hand for we have the tests in place in a follow up? Or for we have tracked an issue to do so at least? 

/Otherwise

All shows fine for me.

/approved
/ok-to-test

mjlshen:
A test for this definitely makes sense, I'll get the test in this PR as well, thanks for the feedback and pointers!

docs/book/src/cronjob-tutorial/testdata/project/go.sum
mjlshen: I was using Go 1.20 and thought I rebased, but I must have done something wrong - should be fixed now.

mjlshen:
I believe I have also added a relevant test to guard against bugs like these in the future!

docs/book/src/cronjob-tutorial/testdata/project/go.sum
mjlshen: /hold

Actually I found that this change was correct - CI was failing and I think we should first fix CI in #3459 then I can rebase this PR on top of that one after it merges.

rbroggi:
Nice, I also faced this issue yesterday!

mjlshen:
The book updates should be in this PR too (the changes underneath `docs/`), unless you are referring to something else - but the process is described here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#how-to-contribute-to-docs

camilamacedo86:
HI @mjlshen,

Could you please rebase this one with the master branch to get the CI fixes?
Also, please ensure that we have only 1 commit for we are able to move forward with this one.

By last would be amazing if we could add the e2e tests in this one

docs/book/src/cronjob-tutorial/testdata/project/go.sum
mjlshen: /unhold

test/e2e/v4/plugin_cluster_test.go
```diff
@@ -107,6 +107,14 @@ func Run(kbc *utils.TestContext) {
 	err = kbc.Tidy()
 	ExpectWithOffset(1, err).NotTo(HaveOccurred())
 
+	By("run make manifests")
+	err = kbc.Make("manifests")
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+	By("run make generate")
+	err = kbc.Make("generate")
```
mjlshen, line 115: @camilamacedo86 Thanks for taking care of CI - I've rebased onto master to pickup those fixes, though I _thought_ that I already took care of the e2e tests with this change. Am I missing something?

camilamacedo86: I added the tests: https://github.com/kubernetes-sigs/kubebuilder/pull/3468 and I could not face the issue.
So, I am not sure what is the problem with the default scaffold that we have. 

Could you please raise an PR with only the new test which should be failing since this PR is not merged for we ensure the fix?

camilamacedo86: Hi @mjlshen 

See that the PR is merged and the targets are not failing.
So, when the error is faced? Only when we run `kustomize build config/crd` direct?

Can you please clarify when/how to reproduce the issue?

camilamacedo86 (changes requested):
I am unable to reproduce the issue scenario, see: https://github.com/kubernetes-sigs/kubebuilder/issues/3425#issuecomment-1598846375
Also, by looking at the kustomize docs the current code implementation shows right: https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/patches/#patch-using-inline-strategic-merge

By last, since I am unable to reproduce, we test this config in the e2e tests where we ensure that all works fine and by looking in the issue seems that the problem is that not all values that should be uncommented were not then, I do not think that we should move with this one.

So, /hold

Let's discuss this scenario in the issue.
Thank you a lot for your collaboration.

camilamacedo86 (commented):
Hi mjlshen, 

Can we add here the e2e test changes as well so that we can ensure that the same error is no longer faced with the changes performed?

mjlshen:
I've moved the e2e test fixes into this PR - thank you a lot for your collaboration as well, working on this PR has been a great experience!

test/e2e/v4/plugin_cluster_test.go
mjlshen: I ended up reproducing this in #3472 and pulling those changes into this PR

camilamacedo86 (approved):
/hold cancel

It has a test either to show that it fixes the issue

/approved
/lgtm 

PR #3461, opened 2023-06-19
🐛 Fix issue found to re-generate the scaffolds with the new alpha command and adding e2e tests

yyy1000:
### Description:
Add e2e tests for kubebuilder alpha generate command.

### Motivation:
Validate the functionality and the behavior of the new alpha command that has been merged.


camilamacedo86:
@yyy1000,

PLease squash the commits and rebase it with master so that we can move forward with this one

test/e2e/alphagenerate/e2e_suite_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// Run e2e tests using the Ginkgo runner.
+func TestE2E(t *testing.T) {
+	RegisterFailHandler(Fail)
+	fmt.Fprintf(GinkgoWriter, "Starting kubebuilder alpha generate suite\n")
```
camilamacedo86, line 30:
```suggestion
	fmt.Fprintf(GinkgoWriter, "Starting kubebuilder suite test for the alpha command generate\n")
```

test/e2e/alphagenerate/generate_test.go
```diff
@@ -0,0 +1,88 @@
...
+			kbc, err = utils.NewTestContext(pluginutil.KubebuilderBinName, "GO111MODULE=on")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(kbc.Prepare()).To(Succeed())
+		})
+
+		AfterEach(func() {
+			kbc.Destroy()
+		})
+
+		It("should generate a runnable project with alpha generate", func() {
```
camilamacedo86, line 52:
```suggestion
		It("should regenerate the project with success", func() {
```

We are not testing that the result is runnable so I think we need to change the test description

yyy1000:
I seems that the e2e tests in CI indicate other error not related to the change in this PR.

test/e2e/alphagenerate/generate_test.go
```diff
@@ -0,0 +1,88 @@
...
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+	By("regenerating the project")
+	err = kbc.Regenerate(
+		"--input-dir", kbc.Dir,
+		"--output-dir", filepath.Join(kbc.Dir, "testdir"),
+	)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+	fileContainsExpr, err := pluginutil.HasFileContentWith(
```
camilamacedo86, line 78:
```suggestion
        By("checking if the project file was generated with the expected content")
	fileContainsExpr, err := pluginutil.HasFileContentWith(
```

test/e2e/alphagenerate/generate_test.go
```diff
@@ -0,0 +1,88 @@
...
+		filepath.Join(kbc.Dir, "testdir", "PROJECT"),
+		`domain: my.domain
+		layout:
+		- go.kubebuilder.io/v4
+		projectName: testdir
+		repo: my.domain/hello
+		version: "3"`)
```
camilamacedo86, line 85: I think it is easier to keep maintained if we create a const could we use a const instead?
See the other places for example.
It might be failing due missing spaces and etc.

test/e2e/alphagenerate/generate_test.go
camilamacedo86: Let's add a description so we have further info when fails to know where it failed

test/e2e/alphagenerate/generate_test.go
```diff
@@ -0,0 +1,88 @@
...
+	)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+	fileContainsExpr, err := pluginutil.HasFileContentWith(
+		filepath.Join(kbc.Dir, "testdir", "PROJECT"),
+		`domain: my.domain
+		layout:
+		- go.kubebuilder.io/v4
+		projectName: testdir
+		repo: my.domain/hello
```
camilamacedo86, line 84: The repo, domain will not be this one since when we use the testContext and generate the project we create a domain and etc for the test. Therefore, see that 

- The domain will be kbc.Domain so you can use format string to pass the const with %s and replace the domain with the expected value
- I would suggest you remove the `repo: my.domain/hello`

camilamacedo86 (commented):
I added so comments but you are doing great !!!

test/e2e/alphagenerate/generate_test.go
yyy1000: That's a good idea.

test/e2e/alphagenerate/generate_test.go
yyy1000: Sorry for this problem.  I set domain to this because I tested in local, and whatever the old PROJECT file is. The domain of the new PROJECT file is always 'my.domain'. Maybe there's some problem with my implementation of 'alpha generate'.

camilamacedo86 (approved):
I just change the title but it seems great for me 🥇 
Good work,

Thank you

pkg/rescaffold/migrate.go
```diff
@@ -100,8 +100,24 @@ func getOutputPath(currentWorkingDirectory, outputPath string) (string, error) {
...
+	var args []string
+	plugins := store.Config().GetPluginChain()
+	if len(plugins) > 0 {
+		args = append(args, "--plugins")
+		args = append(args, plugins...)
+	}
+	domain := store.Config().GetDomain()
+	if domain != "" {
+		args = append(args, "--domain")
+		args = append(args, domain)
```
varshaprasad96, line 120: Similarly, does this work?
```suggestion
		args = append(args, "--domain", domain)
```

test/e2e/alphagenerate/generate_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"path/filepath"
+
+	pluginutil "sigs.k8s.io/kubebuilder/v3/pkg/plugin/util"
+
+	//nolint:golint
+	//nolint:revive
+	. "github.com/onsi/ginkgo/v2"
+
+	//nolint:golint
+	//nolint:revive
```
varshaprasad96, line 30: Do we need to disable the linters here? Looks like we don't in other places and it works (https://github.com/kubernetes-sigs/kubebuilder/blob/47e6cab305336e1a7a5d577dc3c1500a445a70d1/pkg/cli/version_test.go#L20C1-L21)?

varshaprasad96 (changes requested):
Great work @yyy1000. Just a few nit questions.

pkg/rescaffold/migrate.go
```diff
@@ -100,8 +100,24 @@ func getOutputPath(currentWorkingDirectory, outputPath string) (string, error) {
...
+	args = append(args, getInitArgs(store)...)
 	return util.RunCmd("kubebuilder init", "kubebuilder", args...)
 }
+
+func getInitArgs(store store.Store) []string {
+	var args []string
+	plugins := store.Config().GetPluginChain()
+	if len(plugins) > 0 {
+		args = append(args, "--plugins")
+		args = append(args, plugins...)
```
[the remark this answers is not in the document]
yyy1000: No, I tried this and it will not compile: 
```
too many arguments in call to append
        have ([]string, string, []string)
        want ([]string, ...string)
make: *** [Makefile:57: build] Error 1
```

pkg/rescaffold/migrate.go
yyy1000: This works! I think it's because when using `append` method, you either append multiple single element or a slice of elements. But can't mix them. :)

test/e2e/alphagenerate/generate_test.go
yyy1000: Ah, I just copy the code from other e2e tests code. I removed them, ran 'make lint' command and it was OK. I think I can also remove other unnecessary linters in another PR. :)

camilamacedo86 (commented):
/lgtm

PR #3468, opened 2023-06-20
🌱  Adding checks to e2e tests to verify make manifest and generate command

camilamacedo86:
## Description

Adding checks to e2e tests to verify make manifest and generate command


camilamacedo86:
Moving forward since it is juts a need to add a couple of checks in the e2e tests.

PR #3472, opened 2023-06-21
🐛 Uncomment config/crd/kustomization.yaml in e2e test

mjlshen:
This PR should cause a failing e2e run in CI and represents a reproducer of #3425 and should not be merged directly. If it looks good, these changes will be bundled into #3456


mjlshen:
/hold

cc @camilamacedo86 this should reproduce the problem reported by the issue in CI

camilamacedo86:
Great so, now we can check if with the changes in the other PR this issue is solved !!!

PR #3474, opened 2023-06-21
:bug: fix(completion): Fix completion in deprecated projects

### LINKED ISSUE #3473: Deprecation Notice breaks (atleast zsh) completion

When generating the completion script or accessing auto complete functionality while inside a directory with a deprecated project the deprecation notice is generated on standard out and therefore interfers with the normal evaluation. Since the deprecation notice is not valid script syntax the completion script cannot be evaluated and even if loaded correctly, the deprecation notice will intermingle with the correct suggestions in deprecated paths

```
source <(kubebuilder completion zsh)
# resulting error: /proc/self/fd/15:1: bad pattern: ^[[1

kubebuilder <tab>
# completion                                                                 
# create                                                                         
# edit                                                                           
# help                                                                           
# init                                                                           
# version                                                                        
# ^[[1;36m[Deprecation Notice] This version is deprecated.The `go/v3` cannot scaffold projects using kustomize versions v4x+ and cannot fully support Kubernetes 1.25+.It is recommended to upgrade your project to the latest versions available (go/v4).Please, check the migration guide to learn how to upgrade your project
```

### Reproducing this issue

```zsh
#Clone a project that will trigger a deprecation warning:
git clone https://github.com/epam/edp-keycloak-operator/tree/1a935cec5558ee7bccaf8b4bc475ca364576e183
# CD into the project so that kubebuilder recognizes a deprecated project
cd edp-keycloak-operator
# Try to generate and source the completion scripts
source <(kubebuilder completion zsh)
# resulting error: /proc/self/fd/15:1: bad pattern: ^[[1

# Proof that the deprecation notice is generated in the subshell
kubebuilder completion zsh | head -n 5 | less -R

# Proof that this does not happen outside of deprecated projects
source <(cd && ; kubebuilder completion zsh)

# If loaded correctly, attempt completion inside deprecated project
kubebuilder <tab>
# completion                                                                 
# create                                                                         
# edit                                                                           
# help                                                                           
# init                                                                           
# version                                                                        
# ^[[1;36m[Deprecation Notice] This version is deprecated.The `go/v3` cannot scaffold projects using kustomize versions v4x+ and cannot fully support Kubernetes 1.25+.It is recommended to upgrade your project to the latest versions available (go/v4).Please, check the migration guide to learn how to upgrade your project
```

The error occurs if the completion script is sourced at shell-init and the shell is launched in a deprecated project (for example when opening a terminal from an IDE while working on such a project. The above mitigation (cd / &&) can be used to always make the completion initialize cleanly (assuming the user didn't check out a deprecated project into root), this wholly unexpected and not a pattern that users ever need to follow with similar projects, where `source <($BIN completion $SHELL)` works cleanly

### KubeBuilder (CLI) Version

Version: main.version{KubeBuilderVersion:"3.10.0", KubernetesVendor:"1.26.1", GitCommit:"0fa57405d4a892efceec3c5a902f634277e30732", BuildDate:"2023-04-15T08:10:35Z", GoOs:"linux", GoArch:"amd64"}

### PROJECT version

3

### Plugin versions

```yaml
layout:
- go.kubebuilder.io/v3
```

### Other versions

```zsh
zsh --version
zsh 5.9 (x86_64-pc-linux-gnu)
```

### Extra Labels

_No response_

BronzeDeer:
Fixes #3473 

Currently, when initializing in a project with deprecated plugins the kubebuilder cli will immediately output a deprecation notice to stdout before even parsing the command. While this is acceptable and even desirable for interactive usage of the cli, it will "randomly" (i.e. dependent on $PWD and not on the command) break automated usage of the kubebuilder output. This, again is fine if kubebuilder makes no guarantees of stable machine readable output, but this does not hold for the completion sub commands which are only ever parsed by machines: The output of "kubebuilder completion <shell>" must always only output the exact script generated by cobra and the cobra internal commands __complete and __completeNoDesc must only only return an exact list of completion options.

In deprecated projects this will break shell completion if the shell is started in the directory and if the completion script is correctly initialized, all completion options will be polluted by the massive deprecation notice.

This commit instead changes the deprecation notice to output to stderr (while maintaining the return code of the actual command). This fixes the completion usecase which only reads in stdout, and should not break existing integrations, unless there exist scripts which use the length of stderr to check for errors rather than the return code of the call.

### Example broken completion output when deprecation notice is put out on stdout (ZSH)

```
source <(kubebuilder completion zsh)
# resulting error: /proc/self/fd/15:1: bad pattern: ^[[1

kubebuilder <tab>
# completion                                                                 
# create                                                                         
# edit                                                                           
# help                                                                           
# init                                                                           
# version                                                                        
# ^[[1;36m[Deprecation Notice] This version is deprecated.The `go/v3` cannot scaffold projects using kustomize versions v4x+ and cannot fully support Kubernetes 1.25+.It is recommended to upgrade your project to the latest versions available (go/v4).Please, check the migration guide to learn how to upgrade your project
```

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -444,7 +444,7 @@ func (c *CLI) addExtraCommands() error {
 func (c CLI) printDeprecationWarnings() {
 	for _, p := range c.resolvedPlugins {
 		if p != nil && p.(plugin.Deprecated) != nil && len(p.(plugin.Deprecated).DeprecationWarning()) > 0 {
-			fmt.Printf(noticeColor, fmt.Sprintf(deprecationFmt, p.(plugin.Deprecated).DeprecationWarning()))
+			fmt.Fprintf(os.Stderr, noticeColor, fmt.Sprintf(deprecationFmt, p.(plugin.Deprecated).DeprecationWarning()))
 		}
 	}
 }
--- a/pkg/cli/cli_test.go
+++ b/pkg/cli/cli_test.go
@@ -592,13 +592,13 @@ plugins:
 				)
 				deprecatedPlugin := newMockDeprecatedPlugin("deprecated", "v1", deprecationWarning, projectVersion)
 
-				// Overwrite stdout to read the output and reset it afterwards
+				// Overwrite stderr to read the deprecation output and reset it afterwards
 				r, w, _ := os.Pipe()
-				temp := os.Stdout
+				temp := os.Stderr
 				defer func() {
-					os.Stdout = temp
+					os.Stderr = temp
 				}()
-				os.Stdout = w
+				os.Stderr = w
 
 				c, err = New(
 					WithPlugins(deprecatedPlugin),


BronzeDeer:
Replaced commit with signed version

BronzeDeer:
Removed github style automatic reference to accompanying issue #3473 from commit message

camilamacedo86:
/ok-to-test

@BronzeDeer,

Could you please just add in the description of the PR a print with the result of the change since it is hard to ensure with tests? So, that we have it tracked and we can easily review this one.

BronzeDeer:
@camilamacedo86 

I have copied the output error from the linked issue into the PR description.

As for tests, I could add a test that runs the three completion commands, captures stdout and Asserts that the deprecation notice is not in the captured output. This however only really tests the broken completion indirectly and is more of a regression test. If you consider such a test valuable, I'll add it

Kavinjsir (approved):
LGTM!
I think it is a good approach and the update to the test func here looks enough to me.
Thx @BronzeDeer !

varshaprasad96 (approved):
Thanks for the contribution @BronzeDeer!
/lgtm
/approve

PR #3475, opened 2023-06-21
📖 Fix command in Quick Start guide

robinschneider:
Use `apply -k` because folder contains a kustomization and command will fail otherwise

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/README.md
+++ b/docs/book/src/cronjob-tutorial/testdata/project/README.md
@@ -12,7 +12,7 @@ You’ll need a Kubernetes cluster to run against. You can use [KIND](https://si
 1. Install Instances of Custom Resources:
 
 ```sh
-kubectl apply -f config/samples/
+kubectl apply -k config/samples/
 ```
 
 2. Build and push your image to the location specified by `IMG`:


docs/book/src/cronjob-tutorial/testdata/project/README.md
```diff
@@ -12,7 +12,7 @@ You’ll need a Kubernetes cluster to run against. You can use [KIND](https://si
 1. Install Instances of Custom Resources:
 
 ```sh
-kubectl apply -f config/samples/
+kubectl apply -k config/samples/
```
camilamacedo86, line 15: It actually should be make install. See the quick start. We should stick with the recommendations and guidances provided there. 
```suggestion
make install
```

camilamacedo86 (changes requested):
Thank you a lot for your help and contribution 🥇 .

Just a nit: Please, see my comment: https://github.com/kubernetes-sigs/kubebuilder/pull/3475/files#r1240008011

Also, the correct emoji is 🐛 in this case.

docs/book/src/cronjob-tutorial/testdata/project/README.md
camilamacedo86: @Sajiyah-Salat,

When you finish the task that you are working on .. move forward with this PR in the right way might be a good issue for you so that you start to learn and check how the bollerplates work.

Sajiyah-Salat: @camilamacedo86 Sure. I would do that. If possible need your reviews on the pr I am working on to move forward.

Sajiyah-Salat:
https://github.com/kubernetes-sigs/kubebuilder/blob/b020c3483d3877b50099324f25f360de46290c1f/pkg/plugins/golang/v4/scaffolds/internal/templates/readme.go#L48

Should this line be changed as make install.

PR #3487, opened 2023-07-05
✨  Add multigroup flag in alpha generate subcommand

yyy1000:
### Description:

The alpha generate command has initial function that re-scaffold the project.
Now we can add the 'multigroup' flag to support multi APIs.

### DIFF
--- a/pkg/rescaffold/migrate.go
+++ b/pkg/rescaffold/migrate.go
@@ -36,20 +36,24 @@ const DefaultOutputDir = "output-dir"
 func (opts *MigrateOptions) Rescaffold() error {
 	config := yaml.New(machinery.Filesystem{FS: afero.NewOsFs()})
 	if err := config.LoadFrom(opts.InputDir); err != nil {
-		log.Fatal(err)
+		log.Fatalf("Failed to load PROJECT file %v", err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/rescaffold/migrate.go
```diff
@@ -51,6 +51,10 @@ func (opts *MigrateOptions) Rescaffold() error {
 	if err := kubebuilderInit(config); err != nil {
 		log.Fatal(err)
 	}
+	// call edit subcommands to enable or disable multigroup layout
+	if err := kubebuilderEdit(config); err != nil {
+		log.Fatal(err)
```
varshaprasad96, line 56: Nit: Can we also add an error message here so that users are aware when it errored, like:
```
log.Fatalf("Failed to run edit subcommand %v", err)
```

varshaprasad96 (commented):
Just a nit! Rest looks good to me!
/lgtm
/approve

pkg/rescaffold/migrate.go
yyy1000: Sounds great, it can also applied to the Init command.

camilamacedo86 (commented):
/lgtm

PR #3489, opened 2023-07-07
✨ Add API  fields in alpha generate subcommand

yyy1000:
### Description:

We didn't inspect the resources fields in the previous version.
Now we can add support for APIs.
It will check the GVK and also other flags.
(I think the e2e test case can cover more, such as create api without resource or without controller or both, but here I don't know how to check the fields in an efficient way so I skip them now.)


Kavinjsir (approved):
Looks good to me 👍🏼

camilamacedo86 (approved):
/lgtm

PR #3490, opened 2023-07-08
📖 Updates extending kubebuilder docs to use NewBundleWithOptions

sklarsa:
The extending kubebuilder docs page uses a deprecated function call, which leads to an example that doesn't compile.  I updated the call to use https://github.com/kubernetes-sigs/kubebuilder/blob/141bb6c151c272b03cb4255b2c9415da075002ac/pkg/plugin/bundle.go#L99 instead of https://github.com/kubernetes-sigs/kubebuilder/blob/141bb6c151c272b03cb4255b2c9415da075002ac/pkg/plugin/bundle.go#L70


camilamacedo86 (approved):
/lgtm
/approved

PR #3503, opened 2023-07-21
✨ Add Grafana Plugin migration in alpha generate subcommand -- Step 1

yyy1000:
### Description:

Support Grafana Plugin migration in the alpha generate subcommand. 

This is the first step, which will only call 'kubebuilder' command.

### DIFF
--- a/pkg/rescaffold/migrate.go
+++ b/pkg/rescaffold/migrate.go
@@ -14,12 +14,17 @@ limitations under the License.
 package rescaffold
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"os"
 	"os/exec"
+	"strings"
+
+	"golang.org/x/exp/slices"
 
 	"github.com/spf13/afero"
+	"sigs.k8s.io/kubebuilder/v3/pkg/config"
 	"sigs.k8s.io/kubebuilder/v3/pkg/config/store"
 	"sigs.k8s.io/kubebuilder/v3/pkg/config/store/yaml"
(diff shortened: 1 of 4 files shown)


pkg/rescaffold/migrate.go
```diff
@@ -142,12 +152,30 @@ func kubebuilderCreate(store store.Store) error {
...
+	// If the plugin is already in the plugin chain, we don't need call 'edit' method
+	// Because the plugin is already migrated in the previous step
+	plugins := store.Config().GetPluginChain()
+	if slices.Contains(plugins, grafanaPluginKey) {
+		return nil
+	}
+	// If the plugin is not in the plugin chain, we need to call 'edit' method to add the plugin
+	var grafanaPlugin struct{}
+	err := store.Config().DecodePluginConfig(grafanaPluginKey, grafanaPlugin)
+	if err == nil {
```
Kavinjsir, line 165: Nits: I think you can drop `else` here for better readability, for instance:

```go
	if errors.As(err, &config.PluginKeyNotFoundError{}) {
		return nil
	}

	if err != nil {
		return fmt.Errorf("Failed to Decode Grafana Plugin: %s. %v", grafanaPluginKey, err)
	}

	return migrateGrafanaPlugin()
```

Kavinjsir (approved):

pkg/rescaffold/migrate.go
yyy1000: Good idea!

camilamacedo86 (approved):
/lgtm

PR #3506, opened 2023-07-24
:seedling: Replacing destroy and improve info logs for hack docs script generation

camilamacedo86:
## Description

- Replace Destroy func uses in the e2e to remove the image with code implementation to remove the doc samples directory and files in order to re-generate it
- Improve the logs by replacing `fixing suite_test.go` with meaningful info logs to help to troubleshooting when errors are faced.

### DIFF
--- a/hack/docs/internal/component-config-tutorial/generate_component_config.go
+++ b/hack/docs/internal/component-config-tutorial/generate_component_config.go
@@ -56,8 +56,10 @@ func newSampleContext(binaryPath string, samplePath string, env ...string) utils
 
 // Prepare the Context for the sample project
 func (sp *Sample) Prepare() {
-	log.Infof("destroying directory for sample project")
-	sp.ctx.Destroy()
+	log.Infof("destroying directory to create component config tutorial doc sample")
+	if err := os.RemoveAll(sp.ctx.Dir); err != nil {
+		log.Warning("unable to clean up the directory to generate reate component config tutorial doc sample ", err)
+	}
 
 	log.Infof("refreshing tools and creating directory...")
 	err := sp.ctx.Prepare()
--- a/hack/docs/internal/cronjob-tutorial/generate_cronjob.go
+++ b/hack/docs/internal/cronjob-tutorial/generate_cronjob.go
@@ -56,8 +56,10 @@ func newSampleContext(binaryPath string, samplePath string, env ...string) utils
 
 // Prepare the Context for the sample project
 func (sp *Sample) Prepare() {
-	log.Infof("destroying directory for sample project")
-	sp.ctx.Destroy()
+	log.Infof("destroying directory to create CronjobTutorial")
+	if err := os.RemoveAll(sp.ctx.Dir); err != nil {
+		log.Warning("unable to clean up the directory to create CronjobTutoria", err)
+	}
 
 	log.Infof("refreshing tools and creating directory...")
 	err := sp.ctx.Prepare()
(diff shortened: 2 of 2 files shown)


hack/docs/internal/component-config-tutorial/generate_component_config.go
```diff
@@ -56,8 +56,10 @@ func newSampleContext(binaryPath string, samplePath string, env ...string) utils
 
 // Prepare the Context for the sample project
 func (sp *Sample) Prepare() {
-	log.Infof("destroying directory for sample project")
-	sp.ctx.Destroy()
+	log.Infof("destroying directory to create component config tutorial doc sample")
+	if err := os.RemoveAll(sp.ctx.Dir); err != nil {
+		log.Warning("unable to clean up the directory to generate reate component config tutorial doc sample ", err)
+	}
```
everettraven, line 62: I think it is okay to keep the `sp.ctx.Destroy()` usage, however I think the implementation in https://github.com/kubernetes-sigs/kubebuilder/blob/d60aae7388aba080eaab85efd9462fb59c26ec25/test/e2e/utils/test_context.go#L241-L251 should be updated to not run the `docker rmi -f ...` command if the image name is an empty string. This allows us to keep consistency with how all samples are generated while improving the `TestContext` function (for any sample and any case we shouldn't knowingly run a command that might fail and we aren't properly checking for a potential failure case). I'm thinking something like:
```go
// Destroy is for cleaning up the docker images for testing
func (t *TestContext) Destroy() {
        //nolint:gosec
        if t.ImageName != "" { // <-- This if is the difference
                cmd := exec.Command("docker", "rmi", "-f", t.ImageName)
                if _, err := t.Run(cmd); err != nil {
                     warnError(err)
                }
        }
        
        if err := os.RemoveAll(t.Dir); err != nil {
                warnError(err)
        }
}
```

hack/docs/internal/cronjob-tutorial/generate_cronjob.go
```diff
@@ -56,8 +56,10 @@ func newSampleContext(binaryPath string, samplePath string, env ...string) utils
 
 // Prepare the Context for the sample project
 func (sp *Sample) Prepare() {
-	log.Infof("destroying directory for sample project")
-	sp.ctx.Destroy()
+	log.Infof("destroying directory to create CronjobTutorial")
+	if err := os.RemoveAll(sp.ctx.Dir); err != nil {
+		log.Warning("unable to clean up the directory to create CronjobTutoria", err)
+	}
```
everettraven, line 62: Same as my previous comment

hack/docs/internal/component-config-tutorial/generate_component_config.go
camilamacedo86: I agree with the suggestion !!! + 1000

@Sajiyah-Salat can you get the content of this PR and move forward with the suggestion made by @everettraven,

hack/docs/internal/cronjob-tutorial/generate_cronjob.go
```diff
@@ -507,7 +509,7 @@ var testEnv *envtest.Environment
 	/*
 		Then, we start the envtest cluster.
 	*/`)
-	CheckError("fixing suite_test.go", err)
+	CheckError("updating suite_test.go to add text to show where envtest cluster start", err)
```
Kavinjsir, line 512: @camilamacedo86 I think it is good to add details of what is the intention when an error happens.

One thing I'm concerning is:
Logs generated by `CheckErrors` can be very long. Because the `err` from `pluginutil.InsertCode` usually presents the file content, see:
https://github.com/kubernetes-sigs/kubebuilder/blob/v3.11.1/pkg/plugin/util/util.go#L75

It is helpful since we can view to the raw content to trace the error.
Well, I'm wondering if we can make the logs `formatted`, such as adding indention before printing the raw file content?

Sajiyah-Salat: sure, we can add the indentation to seperate the file content and actual error.

hack/docs/internal/component-config-tutorial/generate_component_config.go
Sajiyah-Salat: I got it. let me make a change.

camilamacedo86:
closing it was addressed in other PRS

Issue #3435, opened 2023-05-29
Drop unnecessary steps and better logging when generating docs


Kavinjsir:
### 1. Unnecessary `docker` commands
Currently executing `make generate-docs` will trigged `docker rmi -f`, which might not be necessary:
<img width="552" alt="Screen Shot 2023-05-29 at 12 03 44 PM" src="https://github.com/kubernetes-sigs/kubebuilder/assets/18136486/087f93a5-883e-4d45-9f43-a9ee5ddc68dd">

### 2. Hard to debugging when error happens during docs generation
Another case is: when error happens during docs generation by `make generate-docs`, it is hard to locate the error position from the display log info:
<img width="898" alt="Screen Shot 2023-05-29 at 12 06 27 PM" src="https://github.com/kubernetes-sigs/kubebuilder/assets/18136486/99ab5bec-ea1e-4d25-b847-9276e73d01d6">

## Suggestion

1. Remove the docker command and other unnecessary steps if possible for `make generate-docs`
2. Display rich log info when generating docs.

### One Approach for suggestion 2
For the UpdateTutorial method, we can change the comment lines to `log.Info`:
<img width="407" alt="Screen Shot 2023-05-29 at 12 09 24 PM" src="https://github.com/kubernetes-sigs/kubebuilder/assets/18136486/0e7efcdb-f69b-4902-a1bd-452810901852">

So that when running `make generate-docs`, it support trace the process:
<img width="804" alt="Screen Shot 2023-05-29 at 12 10 03 PM" src="https://github.com/kubernetes-sigs/kubebuilder/assets/18136486/4eaf9cdf-5635-4189-98b6-4c1089070b70">

### Extra Labels

/kind documentation, /kind cleanup

ashutosh887:
@Kavinjsir 
Should I work on this one?

camilamacedo86:
+1, IMO we need to:

- Improve the logs as suggested
- Ensure that all operations has a log for errors with an unique text so that when it fails we know where and the error itself.

I am adding a good first issue on this one.

ashutosh887:
Sure let me work on this @camilamacedo86 

Thanks

camilamacedo86:
What is required to be done in this task:

## 1) What is the main go of this task:

The main go of this task is ensure that for each operation done in the hack implementation to generate the doc samples we have a check to verify the errors with an information that help us to know what fails so that we can fix it. For example, see:

[...]

PR #3508, opened 2023-07-25
:seedling:  Update destroy function 

Sajiyah-Salat:
## Description
Updated destroy function to run docker only when image name is present. This removes the log error while generating docs.

Partial address: https://github.com/kubernetes-sigs/kubebuilder/issues/3435

### DIFF
--- a/test/e2e/utils/test_context.go
+++ b/test/e2e/utils/test_context.go
@@ -232,9 +232,12 @@ func (t *TestContext) Tidy() error {
 // Destroy is for cleaning up the docker images for testing
 func (t *TestContext) Destroy() {
 	//nolint:gosec
-	cmd := exec.Command("docker", "rmi", "-f", t.ImageName)
-	if _, err := t.Run(cmd); err != nil {
-		warnError(err)
+	// if image name is not present or not provided skip execution of docker command
+	if t.ImageName != "" {
+		cmd := exec.Command("docker", "rmi", "-f", t.ImageName)
+		if _, err := t.Run(cmd); err != nil {
+			warnError(err)
+		}
 	}
 	if err := os.RemoveAll(t.Dir); err != nil {
 		warnError(err)


test/e2e/utils/test_context.go
```diff
@@ -232,9 +232,12 @@ func (t *TestContext) Tidy() error {
...
 	//nolint:gosec
-	cmd := exec.Command("docker", "rmi", "-f", t.ImageName)
-	if _, err := t.Run(cmd); err != nil {
-		warnError(err)
+	// if image name is not present or not provided skip execution of docker command
+	if t.ImageName != "" {
+		cmd := exec.Command("docker", "rmi", "-f", t.ImageName)
+		if _, err := t.Run(cmd); err != nil {
+			warnError(err)
+		}
```
varshaprasad96, line 240: Looks good! Seems like when `t.ImageName` is set by hardcoding the value in some of the places while creating a test context, like this: https://github.com/kubernetes-sigs/kubebuilder/blob/a5b246f1ff5b6444ef75bb82c73ae557f47b2438/test/e2e/utils/test_context.go#L91.

So let's also add a warning when `t.ImageName` is empty and we don't execute the command. Something like this:

```suggestion
	if t.ImageName != "" {
		cmd := exec.Command("docker", "rmi", "-f", t.ImageName)
		if _, err := t.Run(cmd); err != nil {
			warnError(err)
		}
	} else {
	           warnErr(fmt.Errorf("ImageName not set, skipping cleaning up of docker image"))
	}	
```

Wdyt? 
cc: @camilamacedo86 ?

Sajiyah-Salat: if there is no image we are already giving a warning error then why adding else.

test/e2e/utils/test_context.go
```diff
@@ -232,9 +232,12 @@ func (t *TestContext) Tidy() error {
 // Destroy is for cleaning up the docker images for testing
 func (t *TestContext) Destroy() {
 	//nolint:gosec
-	cmd := exec.Command("docker", "rmi", "-f", t.ImageName)
-	if _, err := t.Run(cmd); err != nil {
-		warnError(err)
+	// if image name is not present or not provided skip execution of docker command
+	if t.ImageName != "" {
```
camilamacedo86, line 236: We could either use trim here to avoid the edge case " " string.Trim(t.ImageName) and then you check the len to see if is < 0. 

However, it is a nit I am fine to move forward with as it is .

camilamacedo86 (approved):
/lgtm

/approved

Since followed up the @varshaprasad96 suggestions.

test/e2e/utils/test_context.go
```diff
@@ -232,9 +232,12 @@ func (t *TestContext) Tidy() error {
...
+		cmd := exec.Command("docker", "rmi", "-f", t.ImageName)
+		if _, err := t.Run(cmd); err != nil {
+			warnError(err)
```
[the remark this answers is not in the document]
Sajiyah-Salat: > @varshaprasad96 Isnt this the line where we are warning about error. I am not sure that I understood the codeblock but it looks like if any error occur we are giving the warning but its not specified for not passing the image name.

I did this comment immediately after your reply. But there was a pending label and I thought you havent read it yet. It was really confusing and now I got it.

PR #3509, opened 2023-07-25
✨ Add Grafana Plugin migration in alpha generate subcommand -- Step 2

yyy1000:
### Description:

Support Grafana Plugin migration in the alpha generate subcommand. 

This is the second step, which will copy the config.yaml in metrics.

### DIFF
--- a/pkg/rescaffold/migrate.go
+++ b/pkg/rescaffold/migrate.go
@@ -21,8 +21,6 @@ import (
 	"os/exec"
 	"strings"
 
-	"golang.org/x/exp/slices"
-
 	"github.com/spf13/afero"
 	"sigs.k8s.io/kubebuilder/v3/pkg/config"
 	"sigs.k8s.io/kubebuilder/v3/pkg/config/store"
@@ -42,7 +40,7 @@ const grafanaPluginKey = "grafana.kubebuilder.io/v1-alpha"
 
 func (opts *MigrateOptions) Rescaffold() error {
 	config := yaml.New(machinery.Filesystem{FS: afero.NewOsFs()})
-	if err := config.LoadFrom(opts.InputDir); err != nil {
+	if err := config.LoadFrom(fmt.Sprintf("%s/%s", opts.InputDir, yaml.DefaultPath)); err != nil {
 		log.Fatalf("Failed to load PROJECT file %v", err)
 	}
 	// create output directory
@@ -67,7 +65,7 @@ func (opts *MigrateOptions) Rescaffold() error {
 		log.Fatalf("Failed to run create API subcommand %v", err)
 	}
 	// plugin specific migration
-	if err := kubebuilderGrafanaPlugin(config); err != nil {
+	if err := migrateGrafanaPlugin(config, opts.InputDir, opts.OutputDir); err != nil {
 		log.Fatalf("Failed to run plugin migration %v", err)
 	}
 	return nil
@@ -102,7 +100,7 @@ func getInputPath(currentWorkingDirectory string, inputPath string) (string, err
 	if _, err := os.Stat(projectPath); os.IsNotExist(err) {
 		return "", fmt.Errorf("PROJECT path: %s does not exist. %v", projectPath, err)
 	}
-	return projectPath, nil
+	return inputPath, nil
 }
 
 func getOutputPath(currentWorkingDirectory, outputPath string) (string, error) {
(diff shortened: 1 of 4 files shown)


test/e2e/alphagenerate/generate_test.go
```diff
@@ -191,4 +204,11 @@ func ReGenerateProject(kbc *utils.TestContext) {
 		filepath.Join(kbc.Dir, "testdir2", "PROJECT"), grafanaPlugin)
 	ExpectWithOffset(1, err).NotTo(HaveOccurred())
 	ExpectWithOffset(1, fileContainsExpr).To(BeTrue())
+
+	By("checking if the generated grafana config file has the same content as the old one")
+	cmp := equalfile.New(nil, equalfile.Options{}) // compare using single mode
+	equal, err := cmp.CompareFile(filepath.Join(kbc.Dir, "testdir2", "grafana/custom-metrics/config.yaml"),
+		filepath.Join(kbc.Dir, "grafana/custom-metrics/config.yaml"))
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+	ExpectWithOffset(1, equal).To(BeTrue())
```
varshaprasad96, line 213: Also, with this one. We may not need `ExpectWithOffset`.

varshaprasad96 (changes requested):

pkg/rescaffold/migrate.go
```diff
@@ -152,24 +152,27 @@ func kubebuilderCreate(store store.Store) error {
...
+		var grafanaPlugin struct{}
+		err := store.Config().DecodePluginConfig(grafanaPluginKey, grafanaPlugin)
+		// If the grafana plugin is not found, we don't need to migrate
+		if errors.As(err, &config.PluginKeyNotFoundError{}) {
+			return nil
+		}
+		if err != nil {
+			return fmt.Errorf("Failed to Decode Grafana Plugin: %s. %v", grafanaPluginKey, err)
+		}
+		err = kubebuilderGrafanaEdit()
```
Kavinjsir, line 170: Question:
Why do we need to run `kubebuilderGrafanaEdit` in such circumstance where:
> the **PluginChain** does not contain `grafana-plugin` but the `grafana-plugin` can loaded from the **config**?

yyy1000: First, in my understanding, GetPluginChain() method will return the 'layout' field in the PROJECT file, but not plugins field. Correct me if wrong. @camilamacedo86 @varshaprasad96 
In this case, if we don't run `kubebuilder init --plugins grafana.kubebuilder.io/v1-alpha` , but run 'edit' when the project has been inited. The PluginChain will not contain the grafana plugin('layout' field), but it will appears in 'plugins' field. So we need to call 'edit' and 'copy file'.
But if we init the project using two plugins, (or only the grafana plugin itself). The layout field will contain 'grafana plugin', which could be retrieved by 'GetPluginChain()' methond. Since we will also init the **new project** with all plugins. We don't need to call 'edit', but need to 'copy file'. 
Does it make sense?

test/e2e/alphagenerate/generate_test.go
```diff
@@ -191,4 +204,11 @@ func ReGenerateProject(kbc *utils.TestContext) {
 		filepath.Join(kbc.Dir, "testdir2", "PROJECT"), grafanaPlugin)
 	ExpectWithOffset(1, err).NotTo(HaveOccurred())
 	ExpectWithOffset(1, fileContainsExpr).To(BeTrue())
+
+	By("checking if the generated grafana config file has the same content as the old one")
+	cmp := equalfile.New(nil, equalfile.Options{}) // compare using single mode
+	equal, err := cmp.CompareFile(filepath.Join(kbc.Dir, "testdir2", "grafana/custom-metrics/config.yaml"),
+		filepath.Join(kbc.Dir, "grafana/custom-metrics/config.yaml"))
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
yyy1000: Good idea.  Acutally I'm still a little confused when should use this one. I think 'ExpectWithOffset' could provide the message about the 'call stack', so it will be used when in a function(maybe a layer of functions). And if not the case, we should use 'Expect'. Is it right?

pkg/rescaffold/migrate.go
```diff
@@ -152,24 +152,27 @@ func kubebuilderCreate(store store.Store) error {
...
+		// If the plugin is not in the plugin chain, we need to call 'edit' method to add the plugin
+		var grafanaPlugin struct{}
+		err := store.Config().DecodePluginConfig(grafanaPluginKey, grafanaPlugin)
+		// If the grafana plugin is not found, we don't need to migrate
+		if errors.As(err, &config.PluginKeyNotFoundError{}) {
+			return nil
+		}
+		if err != nil {
+			return fmt.Errorf("Failed to Decode Grafana Plugin: %s. %v", grafanaPluginKey, err)
+		}
```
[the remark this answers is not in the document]
yyy1000: Oh, I just search the code repo and copy it. :) https://github.com/kubernetes-sigs/kubebuilder/blob/a5b246f1ff5b6444ef75bb82c73ae557f47b2438/pkg/config/v3/config_test.go#L420-L429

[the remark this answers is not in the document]
yyy1000: Ah, after testing it. I think it should be `errors.As` here. Changing it to `Is` will fail. :)
Though I tried to use '&config.PluginKeyNotFoundError{Key: grafanaPluginKey}' when using `Is`, it still fails.

pkg/rescaffold/migrate.go
camilamacedo86: If you find the Grafana plugin in the PROJECT file then you must to:
- When you run kb init use the plugin
- After copy the file which has custom metrics config option for it in the new directory where the project was scaffolded
- Then, run kb edit with the plugin so that if custom metrics were setup for it then, you will also re-generate them.

Kavinjsir: I'm thinking if the logic can be something like:
1. If the origin PROJECT file does not contain the Grafana plugin, then exit
2. If the origin PROJECT file contains the Grafana plugin, which means the origin project uses it:
    2.1 Run `edit ...plugin ...grafana` to generate the basic grafana dir
    2.2 Copy the config.yaml to the new scaffold
    2.3 Run `edit ...plugin ...grafana` again to generate additional files.

I am imagining if we can get free from the 'complicated' logic...

yyy1000: Cool! That's much clearer. I didn't understand 'edit' command before.

varshaprasad96: +1 Agreed with @Kavinjsir's logic. If the plugin doesn't exist in the origin PROJECT file, then we need to run the edit command. We can exit immediately.

Kavinjsir (approved):
LGTM

camilamacedo86 (approved):
All suggestions are in place!!! 

Great work 🥇 

So, that seems good to fly

varshaprasad96 (approved):
/lgtm

PR #3514, opened 2023-08-01
:seedling: Image name edge cases covered

Sajiyah-Salat:
## Description:
In generating docs, if there is no image present we can skip running docker command and this was covered in https://github.com/kubernetes-sigs/kubebuilder/pull/3508
but if the image name is  of  single character or in anyother edge case the code would crash. 
In this pr I have covered the Edge cases. 
Following up on: https://github.com/kubernetes-sigs/kubebuilder/pull/3508#discussion_r1279543133

### DIFF
--- a/test/e2e/utils/test_context.go
+++ b/test/e2e/utils/test_context.go
@@ -243,10 +243,16 @@ func (t *TestContext) Destroy() {
 	//nolint:gosec
 	// if image name is not present or not provided skip execution of docker command
 	if t.ImageName != "" {
-		cmd := exec.Command("docker", "rmi", "-f", t.ImageName)
-		if _, err := t.Run(cmd); err != nil {
-			warnError(err)
+		// Check white space from image name
+		if len(strings.TrimSpace(t.ImageName)) == 0 {
+			fmt.Println("Image not set, skip cleaning up of docker image")
+		} else {
+			cmd := exec.Command("docker", "rmi", "-f", t.ImageName)
+			if _, err := t.Run(cmd); err != nil {
+				warnError(err)
+			}
 		}
+
 	}
 	if err := os.RemoveAll(t.Dir); err != nil {
 		warnError(err)


test/e2e/utils/test_context.go
```diff
@@ -243,10 +243,18 @@ func (t *TestContext) Destroy() {
 	//nolint:gosec
 	// if image name is not present or not provided skip execution of docker command
 	if t.ImageName != "" {
-		cmd := exec.Command("docker", "rmi", "-f", t.ImageName)
-		if _, err := t.Run(cmd); err != nil {
-			warnError(err)
+		// Remove white space from image name
+		trimmedImageName := strings.Trim(t.ImageName, "")
+		// Check trimmedimage name is empty
+		if len(trimmedImageName) < 0 {
```
camilamacedo86, line 249:
```suggestion
		if len(strings.Trim(t.ImageName, "")) < 0 {
```

You can do in the same line and you do not need spent a memory space to store the data in a variable

Otherwise it is great !!!

test/e2e/utils/test_context.go
```diff
@@ -243,10 +243,18 @@ func (t *TestContext) Destroy() {
...
+			fmt.Println("Image not set, skip cleaning up of docker image")
```
camilamacedo86, line 250: Also, see the above lint error 
Run make lint to test your code locally and see that you need to fix the lint issues as well prior we get merged 
Some issues will be fixed by running make lint-fix .. if not, they you can search by the error and see what you need to do to fix them.

Feel free to ping if you need to.

camilamacedo86 (commented):
Can you please squash the commits for we are able to move forward? 
@Sajiyah-Salat ?

test/e2e/utils/test_context.go
```diff
@@ -243,10 +243,16 @@ func (t *TestContext) Destroy() {
 	//nolint:gosec
 	// if image name is not present or not provided skip execution of docker command
 	if t.ImageName != "" {
-		cmd := exec.Command("docker", "rmi", "-f", t.ImageName)
-		if _, err := t.Run(cmd); err != nil {
-			warnError(err)
+		// Check white space from image name
+		if len(strings.Trim(t.ImageName, "")) == 0 {
```
varshaprasad96, line 247: This should either be:
```go
strings.Trim(t.ImageName, " ")
```
or

```go
strings.TrimSpace(t.ImageName)
```

First option has an empty space in the second argument of the trim, which means we are intending to remove all whitespaces. The second one is a helper provided by golang itself: https://pkg.go.dev/strings#TrimSpace

varshaprasad96 (changes requested):
An important correction! Rest looks good!
Good work! @Sajiyah-Salat

camilamacedo86 (approved):
/lgtm

/approved

/hold cancel

PR #3520, opened 2023-08-03
🐛: sync kb binary from source code for docs

Kavinjsir:
## Description

This PR aims to ensure that the `Kubebuilder binary` used for docs generation is built from the source code so that the docs can be updated corresponding to the latest code change.

- Create a new `hack/docs/generate.sh` that builds the kb executable from source code and run the go code to re-generate docs samples
- Modify the target `generate-docs` to call `hack/docs/generate.sh`
- Modify the target `check-docs` to drop unnecessary import and call `hack/docs/generate.sh`
- Change the go code to use kb binary `/tmp/kubebuilder/bin/kubebuilder` which is generated by `build_kb` (instead of the default kb from user's os env).

## Motivation:

Partially resolve https://github.com/kubernetes-sigs/kubebuilder/issues/3435
See comments:
1. https://github.com/kubernetes-sigs/kubebuilder/issues/3435#issuecomment-1646613790
2. https://github.com/kubernetes-sigs/kubebuilder/issues/3435#issuecomment-1654203778

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ generate-testdata: ## Update/generate the testdata in $GOPATH/src/sigs.k8s.io/ku
 
 .PHONY: generate-docs
 generate-docs: ## Update/generate the docs in $GOPATH/src/sigs.k8s.io/kubebuilder
-	go run hack/docs/generate_samples.go
+	./hack/docs/generate.sh
 
 .PHONY: check-docs
 check-docs: ## Run the script to ensure that the docs are updated
--- a/hack/docs/check.sh
+++ b/hack/docs/check.sh
@@ -14,10 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-source "$(dirname "$0")/../../test/common.sh"
-
-build_kb
-
 check_directory="$(dirname "$0")/../../docs/book/src/"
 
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -63,18 +63,14 @@ install: build ## Build and install the binary with the current source code. Use
 ##@ Development
 
 .PHONY: generate
-generate: generate-testdata generate-docs ## Update/generate all mock data. You should run this commands to update the mock data after your changes.
+generate: generate-testdata check-docs ## Update/generate all mock data. You should run this commands to update the mock data after your changes.
```
camilamacedo86, line 66: The target generate has the purpose to generate only and not check
The check is used in the CI only for we ensure that the docs were properly generate.

Kavinjsir: Sounds good!
I just follow your idea and make the corresponding updates:
- Create a new `hack/docs/generate.sh` that builds the `kb executable` from source code and run the go code to re-generate docs samples
- Modify the target `generate-docs` to call `hack/docs/generate.sh`
- Modify the target `check-docs` to drop unnecessary import and call `hack/docs/generate.sh`

Let me know how you'd like this or I'm understanding wrong :)

camilamacedo86 (approved):
Great work.
I am moving with this one because the issue faced in the preview has no relation and it is sorted in another PR.

Thank you.

Issue #3432, opened 2023-05-25
Add devcontainers scaffolding for project generation


kannon92:
### What do you want to happen?

Devcontainers/GitHub workspaces are becoming more popular ways of developing.  Has there been any consideration on how to run a kube builder generates project as a devcontainer?

### Extra Labels

/kind documentation

camilamacedo86:
Hi @kannon92,

Could you please give an example of what you are looking for?
Are you trying to create e2e tests where they would be executed in a devcontainer?

kannon92:
Yea, basically I want to develop and test my operator with a GitHub workspace/devcontainers.  Ideally run unit/integration.

So I was going to see if I could figure out how to make my kubebuilder project run in a devcontainer but I was running into some issues because I’m not as familiar with this as I should.  

But I was thinking that if there was any consideration on the roadmap for a template for generating a devcontainer for new kubebuilder projects.

Does that make sense?

camilamacedo86:
Hi @kannon92,

It is an interesting request. To test the project you will need to:
- Have k8s installed (it can be kind for example)
- Then you will build the image, load the image, run make install and `make deploy IMG=<imagename>` 

You can check the e2e tests done in : https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v4-alpha/monitoring/memcached-operator/test/e2e/e2e_test.go#L114-L133 to know better those steps.

If you would like to push a PR with the solution in place I think that would be acceptable. WDYT?

Sajiyah-Salat:
This I got from github docs as I was trying to understand the devcontainer or githubcode spaces. Acording to docs they are simple codespaces and different repository have different configuration where they add tools and things required for the the development of the specific project. Looks like we need to try to setup development configuration in codespaces of github. So it can be easy to develop things directly from github and no need of pulling the code to local.

Sajiyah-Salat:
I am interested in this pr but I would like to understand first to get to result.

camilamacedo86:
Hi @kannon92,

I remember that I did a test in the past using devcontainers to check the project running in the github workspace. I used this configuration, see: https://github.com/camilamacedo86/kubebuilder/blob/workspace/.devcontainer/devcontainer.json

See that  used minikube instead. However, if you be able to use kind that is better because it is possible to build and load the image without need to push it to a registry. I am not sure if with minikube it is possible to do the same. 

Can you please check this one and see if that works for you? 
[...]

PR #3525, opened 2023-08-07
📖 add devpod to quickstart as alternative

FabianKramm:
## Description

This PR adds [DevPod](https://github.com/loft-sh/devpod) as an alternative method to get started with Kubebuilder to the Quickstart. DevPod is a tool that uses a [devcontainer.json](https://containers.dev/) (like Github codespaces) to spin up a development container. The difference to Github Codespaces is that DevPod is Open-Source, client-only and works on any backend, such as local docker, kubernetes or any public cloud provider.

## Motivation

We saw interest for a `devcontainer.json` integration in issue #3432 and @csantanapr (who is organising the kubebuilder book club) and I decided to create a [DevPod kubebuilder template](https://github.com/loft-sh/devpod-kubebuilder-template). This template uses a `devcontainer.json` to simplify the onboarding experience for new users that are interested in trying out DevPod by creating a development container with all needed tooling preinstalled. 

This has worked pretty well so far and we think it would be a great addition to the kubebuilder book itself. Obviously, I'm also happy to change anything if needed!

### DIFF
--- a/docs/book/src/quick-start.md
+++ b/docs/book/src/quick-start.md
@@ -15,6 +15,15 @@ This Quick Start guide will cover:
 - Access to a Kubernetes v1.11.3+ cluster.
 
 <aside class="note">
+<h1>Alternative: Use DevPod to work within a development container</h1>
+
+[![Open in DevPod!](https://devpod.sh/assets/open-in-devpod.svg)](https://devpod.sh/open#https://github.com/loft-sh/devpod-kubebuilder-template)
+
+You can use [DevPod](https://devpod.sh) and the [Kubebuilder template](https://github.com/loft-sh/devpod-kubebuilder-template) to automate the installation process by starting a development container with the needed tooling preinstalled.
+
+</aside>
+
+<aside class="note">
 <h1>Versions and Supportability</h1>
 
 Projects created by Kubebuilder contain a Makefile that will install tools at versions defined at creation time. Those tools are:


docs/book/src/quick-start.md
```diff
@@ -14,6 +14,15 @@ This Quick Start guide will cover:
 - [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) version v1.11.3+.
 - Access to a Kubernetes v1.11.3+ cluster.
 
+<aside class="note">
+<h1>Alternative: Use DevPod to work within a development container</h1>
+
+[![Open in DevPod!](https://devpod.sh/assets/open-in-devpod.svg)](https://devpod.sh/open#https://github.com/loft-sh/devpod-kubebuilder-template)
+
+You can use [DevPod](https://devpod.sh) and the [Kubebuilder template](https://github.com/loft-sh/devpod-kubebuilder-template) to automate the installation process by starting a development container with the needed tooling preinstalled.
```
csantanapr, line 22: I think is important to mention three more things: 
- that the devpod template creates a Kubernetes cluster using [kind](https://sigs.k8s.io/kind)
- kubebuilder is already installed
- edit the `.devcontainer/Dockerfile` to update the versions of the prerequisite tools (ie golang, kubectl, kind, kubebuilder)

csantanapr (approved):
minor nits

csantanapr:
Thank you for the PR @FabianKramm I think devpod is an easy way to get a development environment to use kubebuilder

docs/book/src/quick-start.md
camilamacedo86: Hi @csantanapr and @FabianKramm,

How to use devContainers within should not IHMO be in the quick start at all.
It is an optional/helper information. So, I think it would fit better in the FAQ section. See: https://book.kubebuilder.io/faq

 It can be How to use deploy and test your solution using GitHub workspaces? 
 
Also, we need to add a small low verbose info here to describe when/who and why do that? (i.e If you are looking for .... then you might want to ....) 

Also, do we need to use this third party? Can we not describe how to do it without this DevPod?
Can we not provide an example of the .devcontainer config file as described in: https://github.com/camilamacedo86/kubebuilder/blob/workspace/.devcontainer/devcontainer.json

camilamacedo86: Note that, It is nice to know that this third-party can help out. However, IMO here the goal would to be to share a config without the usage of ANY third-part at all. See that we just need a config with KIND to achieve the goal or minikube and load the image locally: https://minikube.sigs.k8s.io/docs/handbook/pushing/#7-loading-directly-to-in-cluster-container-runtime

**We avoid as much as possible to relay in ANY third-party solution to provide guidance for our users.**

FabianKramm: @camilamacedo86 thanks for the feedback! Wouldn't the use case for this more development than testing and deployment? Sure I can move it to the FAQ, but I guess it would be more buried there.

DevPod is just an alternative to Github Codespaces, not an additional tool you need, so the config I created at ttps://github.com/loft-sh/devpod-kubebuilder-template/blob/main/.devcontainer/devcontainer.json also perfectly works with regular Github Codespaces (also solves the problem you mentioned with kind). DevPod just allows you to run this everywhere you want (including Kubernetes) and not just Github hosted servers. Its also officially mentioned on the devcontainer side at https://containers.dev/supporting#devpod

Kavinjsir:
I suspect the intention if ever we want to introduce a third-party tools/framework/library or whatever in the documents, especially on the quick-start page.

In my own opinion, when users start getting familiar with Kubebuilder from the official website, we should help them focus on the core content instead of introducing rich 3rd-party tools/framework/library.

No matter how good a third-party entity can be, whether it is good or suitable really depends on user themselves and they do have their own ability to choose what they want to use. The Kubebuilder community should avoid doing propagation for certain 3rd-party entity that may distract users concentration.

FabianKramm:
@Kavinjsir @camilamacedo86 apologies, sounded like a good idea to me and @csantanapr at first, but I totally agree with you, I'll close the PR, sorry for the inconvenience

Issue #3396, opened 2023-05-08
Plugins API for external plugins should provide option to allow check what is the layout and plugins used in a chain


camilamacedo86:
### What broke? What's expected?

Example scenario: 

The problem is that the declarative plugin needs to know if the project is being scaffolded with go/v4 or go/v3 so that it can perform correct scaffolding around Dockerfile in the init subcommand. and unfortunately, when the init subcommand is running, the PROJECT file won't exist at that point so extracting the layout from it won't be an option, which leads me to think that we need to either "inject" the config (the way we do with internal plugins) through the PluginRequest input. 

### Reproducing this issue

- Implement an external plugin
- Call it with kubuilder init plugins=myexternal/v1 
- Then, we would need to ensure that from the code implementation of myexternal/v1  init sub-command would be possible to have access to Project Config (v3 only), which is used to track the project info and scaffold the PROJECT file.

An alternative solution would be to ensure that we generate the PROJECT file before we call the second plugin in a chain so that would allow the external plugins read the file and parse its information.

### KubeBuilder (CLI) Version

3.10.0

camilamacedo86:
Requester (who got this limitation): @em-r 

@rashmigottipati @everettraven I think that would be a great fit for the GSC0 2023 program, either. 
@Eileen-Yu would be great if we could fix this one it seems essential in the medium and long term for the external plugins be succeed.

camilamacedo86:
HI @Eileen-Yu,

Could you please rebase the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/3526?

Also, could you please let us know if has any reason for we do not move with this one? Is that in a final state, could we get this one merged? WDYT? 

Thank you a lot for your collaboration.

Eileen-Yu:
Hi @camilamacedo86 

I've rebased the PR.
The current PR pass the config as a string, which has been verified can be successfully received and written to the PROJECT file. So, I think this PR can be a simple version.
(In other words, if passing string is acceptable, I'd assume this pr is ready to review and merge.)

Eileen-Yu:
> Hi @camilamacedo86
> 
> I've rebased the PR. The current PR pass the config as a string, which has been verified can be successfully received and written to the PROJECT file. So, I think this PR can be a simple version. (In other words, if passing string is acceptable, I'd assume this pr is ready to review and merge.)

Based on some previous discussion, passing the config using the `Cfg` struct is proposed.
[...]

PR #3526, opened 2023-08-07
✨  pass config between KB and external plugin

Eileen-Yu:
### Description:
Todo:  
- [x] Add new specs `Config` in both the `PluginRequest` and `PluginResponse`
- [x] Create a local type definition in the sampleexternalplugin path for easier testing
- [x] Inject the config in `PluginRequest` and pass it to the external plugin
- [x] Update the config on the sample external plugin side, send the updated config back via response
- [x] Enable KB to update the config based on the modified the config passed back from `PluginResponse` 
- [x] Add e2e tests
- Remove the `replace` once the PR gets merged (next PR)

Sample PROJECT updated by go sample external plugin:
 

### Motivation:
Aims to fix the issue https://github.com/kubernetes-sigs/kubebuilder/issues/3396


pkg/plugins/external/init.go
```diff
@@ -17,18 +17,22 @@ limitations under the License.
 package external
 
 import (
+	"fmt"
 	"github.com/spf13/pflag"
```
camilamacedo86, line 21: run `make lint-fix` it will fix the lint issue faced here.

pkg/plugins/external/init.go
```diff
@@ -40,16 +44,29 @@ func (p *initSubcommand) BindFlags(fs *pflag.FlagSet) {
 }
 
 func (p *initSubcommand) Scaffold(fs machinery.Filesystem) error {
+	configBytes, err := yaml.Marshal(p.config)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Scaffolding with config:", string(configBytes))
```
camilamacedo86, line 52: Should we print or you add this one just for debug purposes?

pkg/plugins/external/init.go
```diff
@@ -40,16 +44,29 @@ func (p *initSubcommand) BindFlags(fs *pflag.FlagSet) {
...
 		return err
 	}
 
 	return nil
 }
+
+func (p *initSubcommand) InjectConfig(c config.Config) error {
+	p.config = c
+	return nil
+}
```
camilamacedo86, line 72: However, to ensure that works could we add some tests to this one?

In the e2e test that we have now, could we not validate the config is filled and with the data that is expected?
Also, could we try to pass a value unto the config and see if that will be persisted in the PROJECT file?

See here how we update the config with the data to have it in tracked in the PROJECT file: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/deploy-image/v1alpha1/api.go#L221-L247

Note that this code is reponsable for we track: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-with-deploy-image/PROJECT#L8-L25

pkg/plugins/external/init.go
Eileen-Yu: Yes it's just for debug purpose, would delete it.

Sorry I should leave a comment to avoid confusion :)

docs/book/src/simple-external-plugin-tutorial/testdata/sampleexternalplugin/v1/cmd/flags.go
```diff
@@ -16,10 +16,11 @@ limitations under the License.
 package cmd
 
 import (
+	"v1/external"
 	"v1/scaffolds"
 
 	"github.com/spf13/pflag"
-	"sigs.k8s.io/kubebuilder/v3/pkg/plugin/external"
+	// "sigs.k8s.io/kubebuilder/v3/pkg/plugin/external"
```
camilamacedo86, line 23: Could you please remove the line that is commented?

docs/book/src/simple-external-plugin-tutorial/testdata/sampleexternalplugin/v1/cmd/metadata.go
```diff
@@ -16,10 +16,11 @@ limitations under the License.
 package cmd
 
 import (
+	"v1/external"
 	"v1/scaffolds"
 
 	"github.com/spf13/pflag"
-	"sigs.k8s.io/kubebuilder/v3/pkg/plugin/external"
+	// "sigs.k8s.io/kubebuilder/v3/pkg/plugin/external"
```
camilamacedo86, line 23: same here

docs/book/src/simple-external-plugin-tutorial/testdata/sampleexternalplugin/v1/external/types.go
```diff
@@ -0,0 +1,105 @@
+/*
```
camilamacedo86, line 1: I think we might need here the @everettraven thoughts.
However, why we are adding this file?
Is not it implemented in kubebuilder itself?
Could we not import kubebuilder as lib?

everettraven: I believe these types _are_ already implemented in Kubebuilder as a library under `pkg/plugin/external`. I'm _assuming_ the reasoning behind needing to copy them over was that @Eileen-Yu was unable to use the changes made in the library without there being a commit to update the go.mod to point to. That being said, I _think_ we should be able to use the go.mod `replace` directive to point to the local version of Kubebuilder with something like:
```
replace (
    sigs.k8s.io/kubebuilder/v3 => {relative path to project root}
)
```

A `replace` directive like that would allow the project to use the changes made directly in the project without having to do stuff like this where we are copying code over.

pkg/plugin/external/types.go
```diff
@@ -35,6 +35,9 @@ type PluginRequest struct {
 	// Universe represents the modified file contents that gets updated over a series of plugin runs
 	// across the plugin chain. Initially, it starts out as empty.
 	Universe map[string]string `json:"universe"`
+
+	// Config stores the project configuration file.
+	Config string `json:"config"`
```
everettraven, line 40: I'm curious, why a string type? Do we have any other types we can use here to make it more explicit what the allowed content is?

Eileen-Yu: @camilamacedo86 Fix me if I'm missing something or understanding wrong:

1. I'm wondering if  a literal string of the config content can help make it more general when parsing the data structure from different programs.

2. I'm also wondering how we'd like to define a data structure for config? 
    Shall we directly import the definition from [v3/config](https://github.com/kubernetes-sigs/kubebuilder/blob/v3.11.1/pkg/config/v3/config.go#L55C1-L75C1)? 
    One question here is what exactly content/message a general external plugin will use from the config? 
    How much do we allow the external plugin to modify the config?

everettraven: The biggest concern I have with a string literal of JSON is how do we perform validation on it? How do we ensure that the plugin isn't attempting to use configuration fields that don't exist or they have typos?

docs/book/src/simple-external-plugin-tutorial/testdata/sampleexternalplugin/v1/external/types.go
everettraven: >replace seems a solution for local dev env.

I would say that this is true for this particular use case but there are other uses that are valid for it being pinning versions, using forks, etc. That being said, I don't think it is "wrong" to use the replace directive in the "local dev env" context here specifically because this is a sample that will likely never exist outside of this repository. If it makes it easier to make changes to this sample to have this replace directive then it is worth it IMO.

>If this PR contains the lines of replace, will it be good to be merged?

I haven't done a full review of this PR so I can't say it will be good to be merged but I think it would definitely be closer in my book. I'm not really a fan of doing the copying unless we _have_ to

pkg/plugin/external/types.go
Kavinjsir: I'm wondering if we can have a number of clear criteria to help move on? More specifically, I'm curious if we can answer the following questions before implementation:

- What exactly data structure do we want to transfer for config? (Does it have to be a go struct? If yes, then what is the relationship between such with private `cfg` in config v2/v3?)
- Is it necessary to define new methods for `Config` interface so as to help generate the data that might be transferred with the ep?
- How does it affect PluginRequest and PluginResponse?
- What are the boundaries for an ep to touch with config? (Is the config read-only? OR what exactly fields an ep can and cannot touch with?)
- Is it necessary to ALWAYS pass the config to an EP? (Do we need to introduce a mechanism to make it optional?)

Kavinjsir (changes requested):
Suggestion: can we hold the implementation until we come up with clear definitions / criteria.

camilamacedo86 (approved):
It seems great for me 🥇 
However, would to be nice get the @everettraven review LGTM as well

Kavinjsir (approved):

pkg/plugin/external/types.go
camilamacedo86: @Eileen-Yu 

Here, I think, should be config.Config

camilamacedo86 (changes requested):
@Eileen-Yu 

This one I think is very close in this one
It seems for me that we here :

a) Change:

```
// Config stores the project configuration file.
	Config string `json:"config"`
```

To use config.Config instead of string.

b) Ideally add a test.go for /pkg/plugin/external/types.go . See the other unit tests as example: https://github.com/kubernetes-sigs/kubebuilder/blob/924daa742ac59c9728572396658aaae3199b770f/pkg/config/store/errors_test.go

c) Here instead of re-create the implementation: https://github.com/kubernetes-sigs/kubebuilder/pull/3526/files#diff-e6ed63afd007e4f9eacc0376b867bef072ed4b39eb92cf3e9d683f6d87ffabcf we would use the struct that we have already so that we do not need to keep both places maintained. We can create the interface using the current struct

Eileen-Yu:
Tried to update it with the suggestions.. Is this what we're looking for?

Guess I meet with the similar blocker when I did this last time: the project version would somehow be `0` (but by default it should be 3, not quite sure what caused this) and `PROECT` content failed to update with new fields... but these issues wouldn't happen when `req/res.Config` is set to string.

camilamacedo86:
Hi @Eileen-Yu 

Regards : https://github.com/kubernetes-sigs/kubebuilder/pull/3526#issuecomment-1940426602

My bet here is that when you are trying to run the commands from the external plugin, you are creating the Config struct instead of loading it from the Project file. More info: https://github.com/kubernetes-sigs/kubebuilder/pull/3526#discussion_r1487395021

pkg/plugins/external/init.go
```diff
@@ -43,13 +46,28 @@ func (p *initSubcommand) BindFlags(fs *pflag.FlagSet) {
...
 func (p *initSubcommand) Scaffold(fs machinery.Filesystem) error {
-	cfg := p.config.(*v3cfg.Cfg)
+	fileName := "PROJECT"
+
+	if _, err := os.Stat(fileName); os.IsNotExist(err) {
+		file, err := os.Create(fileName)
+		if err != nil {
+			return err
+		}
+		defer file.Close()
```
camilamacedo86, line 56: @Eileen-Yu here we need to sort out the lint issue
We need to check the error

camilamacedo86 (commented):
@Eileen-Yu 

Can you please:
a) squash the commits for we have just 1 here
b) fix the lint issue
c) rebase with master and push so that we can check it with the final state

Please, let us know if you are facing any issue that you need help with

Eileen-Yu:
Hi @camilamacedo86 , thx for the help!

>  a) squash the commits for we have just 1 here
b) fix the lint issue
c) rebase with master and push so that we can check it with the final state

I did all of the above lists.

And yes, I did see the unit tests still failed because of this version error:
```
 msg: "unable to create config for version \"0\": version 0 is not supported",
              err: <config.UnsupportedVersionError>{
                  Version: {Number: 0, Stage: 0},
```

I tried to fix this with this line: https://github.com/kubernetes-sigs/kubebuilder/pull/3526/files#diff-992e7a26b8776b12932c030601857ed56f364b1ecd7dc0fac9dcfd8c0cb903c6R66 but didn't get the luck 😔

pkg/plugins/external/init.go
```diff
@@ -40,16 +46,51 @@ func (p *initSubcommand) BindFlags(fs *pflag.FlagSet) {
 }
 
 func (p *initSubcommand) Scaffold(fs machinery.Filesystem) error {
+	fileName := "PROJECT"
+
+	if _, err := os.Stat(fileName); os.IsNotExist(err) {
```
camilamacedo86, line 51: We should not create the file
We should fail in this case.

camilamacedo86:
Hi @Eileen-Yu 

thank you a lot 
I have not this weekend for check it out
But up the following one I will run it locally and see if I can fix the nit
https://github.com/kubernetes-sigs/kubebuilder/pull/3526#issuecomment-1945344498

camilamacedo86:
Closing in favor https://github.com/kubernetes-sigs/kubebuilder/pull/4060/files

I hope that you do not mind

camilamacedo86:
Hi @Eileen-Yu 

I think you are not working on this one anymore
So, I am closing but feel free to re-open if you wish to rebase and work on it.

PR #3532, opened 2023-08-11
📖  fix quickstart documentation to only apply sample CR

typeid:
Following the quickstart guide results in the following error when applying the CRs:

```terminal
$ kubectl apply -f config/samples/
guestbook.webapp.my.domain/guestbook-sample created
error: error validating "config/samples/kustomization.yaml": error validating data: [apiVersion not set, kind not set]; if you choose to ignore these errors, turn validation off with --validate=false
```
Even though this is not an issue as the sample CR does get applied, it leads to a bit of confusion due to the error message.

As per [this comment](https://github.com/operator-framework/operator-sdk/issues/3225#issuecomment-643336975), our documentation should be updated to just apply the CR.

### DIFF
--- a/docs/book/src/quick-start.md
+++ b/docs/book/src/quick-start.md
@@ -180,7 +180,7 @@ If you pressed `y` for Create Resource [y/n] then you created a CR for your CRD 
 API definition):
 
 ```bash
-kubectl apply -k config/samples/
+kubectl apply -f config/samples/webapp_v1_guestbook.yaml
 ```
 
 ## Run It On the Cluster


docs/book/src/quick-start.md
```diff
@@ -180,7 +180,7 @@ If you pressed `y` for Create Resource [y/n] then you created a CR for your CRD
 API definition):
 
 ```bash
-kubectl apply -k config/samples/
+kubectl apply -f config/samples/webapp_v1_guestbook.yaml
```
camilamacedo86, line 183: Could you please let us know more about you motivation for this change?

typeid: Ah, I just noticed I was too quick to open a PR and did not see that this is already fixed here (`-k` as seen above vs  `-f` in the book). My bad!

Does the book only get updated on releases?

PR #3533, opened 2023-08-12
📖 : documentation: corrected install CRD command in quick start

mohammadVatandoost:
The command in quick-start page is not working, I send this PR to fix it.
`make deploy 
`
is the correct command.

### DIFF
--- a/docs/book/src/quick-start.md
+++ b/docs/book/src/quick-start.md
@@ -180,7 +180,7 @@ If you pressed `y` for Create Resource [y/n] then you created a CR for your CRD 
 API definition):
 
 ```bash
-kubectl apply -k config/samples/
+make deploy
 ```
 
 ## Run It On the Cluster


docs/book/src/quick-start.md
```diff
@@ -180,7 +180,7 @@ If you pressed `y` for Create Resource [y/n] then you created a CR for your CRD
 API definition):
 
 ```bash
-kubectl apply -k config/samples/
+make deploy
```
camilamacedo86, line 183: that is not the same command.
make deploy will deploy the Operator itself when what is done here is apply the samples

camilamacedo86 (commented):
I am closing this one as defered.
The commands in the quick start are accurate and does not requires changes.

If you see any issue with then, please feel free to raise an issue in the repo. 
So that we can analyse.

Issue #3540, opened 2023-08-17
Minor grammar errors in the Makefile template


VadimEisenberg:
### What broke? What's expected?

https://github.com/kubernetes-sigs/kubebuilder/blob/68abac1dfb5550f7159dec4ce600cd9b6db925bb/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go#L92:

_awk commands_ -> awk command

https://github.com/kubernetes-sigs/kubebuilder/blob/68abac1dfb5550f7159dec4ce600cd9b6db925bb/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go#L157

_To properly provided solutions that supports more than one platform you should use this option_ -> For properly provided solutions that support...

### KubeBuilder (CLI) Version

3.10.0

### PROJECT version

3

### Plugin versions

```yaml
go.kubebuilder.io/v4
```

Issue #3534, opened 2023-08-14
doc: quotes are displayed as &#34; in description of kubebuilder markers on CRD Validation page


tenstad:
### What broke? What's expected?

See expanded description of `+kubebuilder:default` on [CRD Validation](https://book.kubebuilder.io/reference/markers/crd-validation.html?highlight=%22#crd-validation), which appears to come from [controller-tools/pkg/crd/markers/validation.go](https://github.com/kubernetes-sigs/controller-tools/blob/498444bd78eb90dbdf228dfd9ad333d68fe354e7/pkg/crd/markers/validation.go#L218)

The struct should be `{policy: "delete"}`, not `{policy: &#34;delete&#34;}`.

camilamacedo86:
Hi @tenstad,

Thank you for raising it. Yes, the descriptions came from controller-tools project. Therefore, the content:

![Screenshot 2023-08-14 at 15 02 51](https://github.com/kubernetes-sigs/kubebuilder/assets/7708031/36cac036-1ad7-4b22-94c5-0eb7b0bb9ef5)

IN: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/reference/markers/crd-validation.md?plain=1

Then, the implementation which is responsible for generate this documentation can be found in: https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/utils/markerdocs

Also, see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/markerdocs.sh

So, in order to address the issue described here is required to change its implementation in order to properly handle `"`. 

I will mark as help wanted and it might also be a good first issue. However, is required be able to analyse the code to know how properly change it out. Either is possible to verify the change in the doc preview which is available for each PR done against master, see;

![Screenshot 2023-08-14 at 15 10 22](https://github.com/kubernetes-sigs/kubebuilder/assets/7708031/1f5cdde5-110b-4726-b9c7-54741a5ebf4e)

camilamacedo86:
It seems fixed so I am closing this one.
Feel free to re-open if you see a need for it.

PR #3541, opened 2023-08-17
📖  fix escaped html in markdown code segments

typeid:
**Issue #3534**
- we are using `markerdocs` as [mdBook preprocessor](https://rust-lang.github.io/mdBook/for_developers/preprocessors.html)  to generate most of our `References` section (e.g. [CRD Validation](https://book.kubebuilder.io/reference/markers/crd-validation))
- this preprocessor writes HTML but the text fields also contain markdown (e.g. [this line](https://github.com/kubernetes-sigs/controller-tools/blob/498444bd78eb90dbdf228dfd9ad333d68fe354e7/pkg/crd/markers/validation.go#L220))
- we call `html.EscapeString()` on the output of the preprocessor (see [here](https://github.com/kubernetes-sigs/kubebuilder/blob/68abac1dfb5550f7159dec4ce600cd9b6db925bb/docs/book/utils/markerdocs/html.go#L44))
- the preprocessed output is then parsed as markdown by mdBook

Basically, mdBook handles the content as raw strings (code block) thus quotation marks and other HTML escaped symbols are not escaped

1. Raw text:
```
 object: `{policy: "delete"}`
```

2. Preprocessor wraps it with HTML in escaped format

<pre>
object: `{policy: &#34;delete&#34;}`
</pre>

3. mdBook picks up on the markdown backticks and further wraps it in the according HTML for the markdown format without interpreting the escaped characters:
```
<code class="hljs">{policy: &#34;delete&#34;}</code>
```

This PR aims to fix this issue by not HTML escaping the markdown code segments contained in the text, leaving the behavior up to mdBook.

Fixes https://github.com/kubernetes-sigs/kubebuilder/issues/3540

### DIFF
--- a/docs/book/utils/markerdocs/html.go
+++ b/docs/book/utils/markerdocs/html.go
@@ -39,10 +39,29 @@ type toHTML interface {
 // Text is a chunk of text in an HTML doc.
 type Text string
 
-// WriteHTML writes the string as HTML to the given Writer
+// WriteHTML writes the string as HTML to the given Writer while accounting for mdBook's handling
+// of backticks. Given mdBook's behavior of treating backticked content as raw text, this function
+// ensures proper rendering by preventing unnecessary HTML escaping within code snippets. Chunks
+// outside of backticks are HTML-escaped for security, while chunks inside backticks remain as raw
+// text, preserving mdBook's intended rendering of code content.
 func (t Text) WriteHTML(w io.Writer) error {
-	_, err := io.WriteString(w, html.EscapeString(string(t)))
-	return err
+	textChunks := strings.Split(string(t), "`")
+
+	for i, chunk := range textChunks {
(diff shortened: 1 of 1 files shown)


typeid:
`./install-and-build.sh` works locally. 
I saw that there was a fix on the `netlify/kubebuilder/deploy-preview`, and the deploy preview works on another PR. So I'm wondering if this error could be sporadic. Could we re-trigger the netlify job?

typeid:
@camilamacedo86 Ah indeed, forgot to rebase. Done and the deploy-preview works now :)

camilamacedo86 (approved):
Thank you for the amazing contribution 🥇 

It is possible to check that it is fixed now:

Screenshot 2023-08-18 at 07 02 17

/approved

/lgtm

/ok-to-test

PR #3552, opened 2023-08-24
🌱 Bump to alpine 3.18

sbueringer:
Signed-off-by: Stefan Büringer buringerst@vmware.com

I think alpine 3.18 is the same as 3.18.3: https://hub.docker.com/_/alpine/tags

So just bumping it to 3.18 and the other Dockerfiles are already on 3.18


varshaprasad96 (approved):
/lgtm
/approve

PR #3555, opened 2023-08-25
⚠️ : (kustomize/v2, go/v4): Stop to scaffold CA patch injection since it is unnecessary

### LINKED ISSUE #3538: Unnecessary patch for CA injection

The [CA injection patch](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/crd/patches/enablecainjection_patch.go) is not necessary after the switch to [replacements](https://github.com/kubernetes-sigs/kubebuilder/blob/68abac1dfb5550f7159dec4ce600cd9b6db925bb/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go#L97) instead of vars. To be clear, it is not breaking anything either, it just doesn't add anything on top of what the replacements already does. The reason is that the replacements have [create: true](https://github.com/kubernetes-sigs/kubebuilder/blob/68abac1dfb5550f7159dec4ce600cd9b6db925bb/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go#L112), which means they will create the annotation if it doesn't exist.

Since the replacements adds the annotations, it also means that ALL CRDs will get the CA injected instead of just those that have the patch (when uncommenting only some of them, if you have multiple). Is this intentional? I guess it will not cause problems but wanted to double check.

### Reproducing this issue

Simply following the quick start: https://book.kubebuilder.io/quick-start

```bash
mkdir -p ~/projects/guestbook
cd ~/projects/guestbook
kubebuilder init --domain my.domain --repo my.domain/guestbook
kubebuilder create api --group webapp --version v1 --kind Guestbook --controller --resource
make manifests
```

After this you will see that `config/crd/patches` contains `cainjection_in_guestbooks.yaml`, but we need to add a webhook to make it useful.

Create a webhook:

```bash
kubebuilder create webhook --group webapp --version v1 --kind Guestbook --defaulting --programmatic-validation --conversion
make manifests
```

Now uncomment the `[WEBHOOK]` and `[CERTMANAGER]` in `config/default/kustomization.yaml`. The comments in the file explains that one should uncomment the same sections in `config/crd/kustomization.yaml`. However, commenting/uncommenting the following lines does not affect the outcome:

```yaml
# In default/kustomization.yaml
# [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'.
# Uncomment 'CERTMANAGER' sections in crd/kustomization.yaml to enable the CA injection in the admission webhooks.
# 'CERTMANAGER' needs to be enabled to use ca injection
# - webhookcainjection_patch.yaml

# In crd/kustomization.yaml
# [CERTMANAGER] To enable cert-manager, uncomment all the sections with [CERTMANAGER] prefix.
# patches here are for enabling the CA injection for each CRD
# - path: patches/cainjection_in_guestbooks.yaml
#+kubebuilder:scaffold:crdkustomizecainjectionpatch
```

We can check like this:

```bash
kustomize build config/default | grep "cert-manager.io/inject-ca-from"
```

Try running the command with the lines commented/uncommented and see that there is no difference in the output.

The output is as follows:

```bash
# With all uncommented:
    cert-manager.io/inject-ca-from: guestbook-system/guestbook-serving-cert
    cert-manager.io/inject-ca-from: guestbook-system/guestbook-serving-cert
    cert-manager.io/inject-ca-from: guestbook-system/guestbook-serving-cert

# With the two mentioned lines commented:
    cert-manager.io/inject-ca-from: guestbook-system/guestbook-serving-cert
    cert-manager.io/inject-ca-from: guestbook-system/guestbook-serving-cert
    cert-manager.io/inject-ca-from: guestbook-system/guestbook-serving-cert

# With the two mentioned lines uncommented, and the replacements commented:
    cert-manager.io/inject-ca-from: CERTIFICATE_NAMESPACE/CERTIFICATE_NAME
    cert-manager.io/inject-ca-from: CERTIFICATE_NAMESPACE/CERTIFICATE_NAME
    cert-manager.io/inject-ca-from: CERTIFICATE_NAMESPACE/CERTIFICATE_NAME
```

From this it should be clear that the replacements in `config/default/kustomization.yaml` is enough to inject the CA and the patches are not needed.

### KubeBuilder (CLI) Version

Version: main.version{KubeBuilderVersion:"3.11.1", KubernetesVendor:"1.27.1", GitCommit:"1dc8ed95f7cc55fef3151f749d3d541bec3423c9", BuildDate:"2023-07-03T13:10:56Z", GoOs:"linux", GoArch:"amd64"}

### PROJECT version

3

### Plugin versions

```yaml
layout:
- go.kubebuilder.io/v4
```

### Other versions

_No response_

### Extra Labels

_No response_

lentzi90:
This patch is no longer needed since it is injected by the [replacements](https://github.com/kubernetes-sigs/kubebuilder/blob/68abac1dfb5550f7159dec4ce600cd9b6db925bb/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go#L97) anyway.

Fixes #3538


camilamacedo86 (commented):
Following some notes and recommendations. 

Please, note that when you do a change you must run `make generate` to generate the samples.

Also, please be aware that after your change it must pass in the e2e tests so that we can ensure that all still working fine as expected.

By last, the emoji is wrong because it is a change that affects the users. In this case it seems that the best one would to be `:bug:` or `:warning:`.

Also, can you update the title for a clear explanation about the change since it is used to generate the release notes? Example: `:bug: (kustomize/v2, go/v4): Stop to scaffold CA patch injection since it is unnecessary`

lentzi90:
Hmm I'm having some issue running `make generate` even on `master`. Will have to dig a bit...

What I'm seeing is this:

```console
❯ git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean

❯ make generate
rm -rf testdata/
./test/testdata/generate.sh
Building kubebuilder
~/workspace/kubebuilder/testdata/project-v2 ~/workspace/kubebuilder
rm: cannot remove '/usr/local/bin/kustomize': Permission denied
make: *** [Makefile:72: generate-testdata] Error 1
```

camilamacedo86:
Hi @lentzi90,

It seems that the script is unable to cleanup/remove all from the dir testdata before moveforward.

Run `sudo rm-rf *` inside of the testdata dir. 
Also, please ensure that you have your branch rebase with master changes as well.

lentzi90:
Interesting... I had to do 
```bash
sudo rm -rf testdata
sudo rm -rf docs/book/src/component-config-tutorial/testdata/project/bin
sudo rm -rf docs/book/src/multiversion-tutorial/testdata/project/bin
```
to get a successful `make generate`. I'm curious if/why no one else have these issue :thinking:

lentzi90:
Ok so it is complaining that the API changed. I can understand that, but I'm not sure how to solve it. Should we keep the API but make it a no-op?

lentzi90:
Hmm I got a clean e2e test locally before pushing this :thinking: 
It seems extremely unlikely that it would be a flake but I'm going to check one more time
/test pull-kubebuilder-e2e-k8s-1-27-1

lentzi90:
Alright so it somehow was a flake. Weird. Then the only remaining question from my side is this:

> Ok so it is complaining that the API changed. I can understand that, but I'm not sure how to solve it. Should we keep the API but make it a no-op?

camilamacedo86:
> Alright so it somehow was a flake. Weird. Then the only remaining question from my side is this:
> 
> Ok so it is complaining that the API changed. I can understand that, but I'm not sure how to solve it. Should we keep the API but make it a no-op?

Kubebuilder is a LIB as well. So,  it is saying that the change broke the kubebuilder API because the option no longer exist. 

> sigs.k8s.io/kubebuilder/v3/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/crd/patches
>   Incompatible changes:
>   - EnableCAInjectionPatch: removed

That is fine.

camilamacedo86 (commented):
It seems make sense for me and the changes shows fine.
It does not seems a breaking change for those who already scaffold a project with the old versions. 

So I am /lgtm 

But I also would like to get feedback from others from the community

Let's see if others can help either. 
Thx

camilamacedo86 (approved):

lowang-bh:
> ```
>  remove '/usr/local/bin/kustomize': Permission denied
> ```
I also met this error. I try to add write permission to this dir and then works.  You can have a try.

https://github.com/kubernetes-sigs/kubebuilder/pull/3579

camilamacedo86 (changes requested):
/approved cancel

We will need to check it further

camilamacedo86 (changes requested):
@lentzi90,

Thank you for raising it. 

Upon reviewing your PR (https://github.com/kubernetes-sigs/kubebuilder/pull/3555), it seems that the root of the issue might lie in the replacements. We must consider genuine scenarios where users might want to decide which resource should have the cert-manager injected.

Rather than removing the injections altogether, it might be more appropriate to address the replacements. Our goal should be to ensure that the certmanager is only injected into the uncommented files.

**Would you like to help us to sorting it out by changing the replacement to ensure that we will ONLY inject the cert-manager  in the CRDs and/or webhooks which are uncommented?** 

Please, see the comment for further information: https://github.com/kubernetes-sigs/kubebuilder/issues/3538#issuecomment-1705068767

lentzi90:
Thanks for the explanation! I commented on the issue also. I will give it some thought, although I don't have an immediate solution to this. If I come up with something I will try to implement it and push here

Issue #3558, opened 2023-08-25
Multigroup Support in Deploy-Image Plugin


yyy1000:
### What broke? What's expected?

Kubebuilder has supported multigroup option.
But in deploy-image plugin, it can't work when enable multigroup=true.

### Reproducing this issue

Run the following commands in a directory.
- kubebuilder init --domain my.domain --repo my.domain/guestbook
- kubebuilder edit --multigroup=true
- kubebuilder create api --group example.com --version v1alpha1 --kind Memcached --image=memcached:1.6.15-alpine --image-container-command="memcached,-m=64,modern,-v" --image-container-port="11211" --run-as-user="1001" --plugins="deploy-image/v1-alpha"

There will be an error like:
```
failed to create API: unable to scaffold with "deploy-image.go.kubebuilder.io/v1-alpha": error updating main.go: error scaffolding event recorder in cmd/main.go: string if err = (&controller.MemcachedReconciler{
                Client: mgr.GetClient(),
                Scheme: mgr.GetScheme(), not found in /*
Copyright 2023.
```

The reason is: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go#L196 
It will only use '&controllers', but when multigroup is enabled, go will import another separate name..

### KubeBuilder (CLI) Version

3.11

PR #3560, opened 2023-08-26
:bug: (deployimage/v1-beta1): fix scaffold with multigroup and and add optional plugins to multigroup sample to validate changes

camilamacedo86:
## Description
Add the fix for: https://github.com/kubernetes-sigs/kubebuilder/issues/3558
Also, add call to grafana and deployimage into the multigroup sample so that we can either test this options with multigroup layout

Closes:  https://github.com/kubernetes-sigs/kubebuilder/issues/3558


camilamacedo86:
Hi @yyy1000,

Here is the fix for the issue: https://github.com/kubernetes-sigs/kubebuilder/issues/3558

@Kavinjsir @varshaprasad96 

I am moving forward with this one for we do not block the generate alpha command project.

yyy1000 (commented):
LGTM!

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
```diff
@@ -189,28 +189,15 @@ func (s *apiScaffolder) scaffoldCreateAPIFromPlugins(isLegacyLayout bool) error
 // which will have its own controller template which set the recorder so that we can use it
 // in the reconciliation to create an event inside for the finalizer
 func (s *apiScaffolder) updateMainByAddingEventRecorder(isGoV3 bool, defaultMainPath string) error {
```
Kavinjsir, line 191: Wondering if we can drop the parameter `isGoV3 bool` if it is no longer used?

Kavinjsir (changes requested):

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
camilamacedo86: Yep good point.
I changed that

camilamacedo86:
@Kavinjsir 

Could we move forward now?
Have we your LGTM as well?

Kavinjsir (approved):
LGTM!

PR #3563, opened 2023-08-26
:bug: change deprecated warning to yellow

camilamacedo86:
## Description

Currently the deprecation notice is blue. However, the standard and common practice is to use yellow for warnings and deprecations. Changing that accordingly. See:

Screenshot 2023-08-26 at 06 26 30


Kavinjsir (approved):
lgtm!

PR #3564, opened 2023-08-26
🌱 ensure that we call go tidy, make manifests and all  when we generate…

camilamacedo86:
## Description

When we generate the docs we need to:
a) Run make manifests to ensure that we are calling controller-gen to generate the resources
b) Call make all to ensure that we call the main targets
c) Call go mod tidy to fix the go deps


Kavinjsir (approved):
That's awesome! Looks good to me! Thx @camilamacedo86

PR #3565, opened 2023-08-26
🌱 : (deployimage): fix e2e tests which are using go/v3 which is dep…

camilamacedo86:
## Description

The e2e tests for DeployImage plugin are scaffolding projects with go/v3 when it should be using go/v4

## Motivation

Ensure that the e2e tests test the plugin using the layout which is not deprecated.


camilamacedo86:
Moving forward here to unblock other PRs
It is causing intermittent issues due the old kustomize version not be always available.

Kavinjsir (approved):

PR #3571, opened 2023-08-27
✨ : Standardization of Logs using Library logrus ( replace of fmt.Print )

### LINKED ISSUE #3557: Standardization of Logging using logrus

We aim to ensure consistent usage of the logrus library ("github.com/sirupsen/logrus") throughout our codebase. There are instances where other methods like fmt.Println() have been used for logging, which need to be addressed to maintain consistency and improve the user experience (UX).

Reference PR for context: [kubebuilder#3556](https://github.com/kubernetes-sigs/kubebuilder/pull/3556).

**Objectives:**

- Identify and replace non-standard logging mechanisms (e.g., fmt.Println()) with appropriate logrus methods.
- Ensure that logs use the correct severity level (log.INFO, log.WARN, etc.) that aligns with the context of the message.

**Recommendation:**

To facilitate easier code review and integration, it's advised to break down the implementation into multiple smaller PRs rather than one large PR.

### Extra Labels

/kind cleanup

lowang-bh:
Fixes #3557

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -17,8 +17,6 @@ limitations under the License.
 package main
 
 import (
-	"log"
-
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/afero"
 	"sigs.k8s.io/kubebuilder/v3/pkg/cli"
@@ -88,9 +86,9 @@ func main() {
 		cli.WithCompletion(),
 	)
 	if err != nil {
-		log.Fatal(err)
+		logrus.Fatal(err)
 	}
 	if err := c.Run(); err != nil {
-		log.Fatal(err)
+		logrus.Fatal(err)
 	}
 }
--- a/hack/docs/generate_samples.go
+++ b/hack/docs/generate_samples.go
@@ -17,7 +17,7 @@ limitations under the License.
 package main
 
 import (
-	"fmt"
+	log "github.com/sirupsen/logrus"
 
 	componentconfig "sigs.k8s.io/kubebuilder/v3/hack/docs/internal/component-config-tutorial"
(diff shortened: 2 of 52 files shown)


camilamacedo86 (approved):
Hi @lowang-bh,

It seems that a point still missing which is: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/util/exec.go#L31

However, it seems also fine to do that in a follow up.
PS.: Note that I updated the title because this change affects user so we cannot use the 🌱 and it should be in the release notes.

Tested locally:

Screenshot 2023-08-30 at 07 11 42

/lgtm

camilamacedo86 (changes requested):
See that it is not passing in two tests.

Could you please rebase it with the master and ensure that you run `make generate` to properly updated the samples under the testdata so that we can move forward here?

lowang-bh:
> It seems that a point still missing which is: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/util/exec.go#L31

Hi, I have updated it.

lowang-bh:
`make generate` finished successfully.

image

lowang-bh:
> Do we also want to update here:
> 
> 1. https://github.com/kubernetes-sigs/kubebuilder/blob/v3.11.1/pkg/cli/version.go#L32
> 2. https://github.com/kubernetes-sigs/kubebuilder/blob/v3.11.1/pkg/cli/cmd_helpers.go#L207
> 3. https://github.com/kubernetes-sigs/kubebuilder/blob/v3.11.1/pkg/plugin/util/stdin.go#L37

some print is used to interact with terminal user, so I didn't replace those.

PR #3579, opened 2023-09-03
🌱  (make generate) : fix error of 'Permission denied' about binary file

lowang-bh:
sometimes, the permission of files in testdata dir is 555 without write permission.
if no write permission to those binary file, error occurs 'etcd: Permission denied'

image


camilamacedo86 (approved):
/lgtm
/ok-to-test
/approved

camilamacedo86 (commented):

In this case we use `🌱 ` since it does not impact endusers
See that the title of the PRs are used to generate the release notes as the emojis
So, I updated it and I hope that you do not mind.

lowang-bh:
> I updated it and I hope that you do not mind.

Thanks.

PR #3584, opened 2023-09-04
🌱 runtime error on darwin arm64 machine: killed

lowang-bh:
I meet this issue on apple m2  macbook. I don't know why. After run`make install` and then run new installed kb, it will still be killed, until I remove the old one and reinstall it.
```
[1]    59938 killed     $GOPATH/bin/kubebuilder version
```
image

After rm it and re-install
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,7 @@ build: ## Build the project locally
 
 .PHONY: install
 install: build ## Build and install the binary with the current source code. Use it to test your changes locally.
+	rm -f $(GOBIN)/kubebuilder
 	cp ./bin/kubebuilder $(GOBIN)/kubebuilder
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -58,6 +58,7 @@ build: ## Build the project locally
 
 .PHONY: install
 install: build ## Build and install the binary with the current source code. Use it to test your changes locally.
+	rm -f $(GOBIN)/kubebuilder
```
camilamacedo86, line 61: Why not rm -rf ?

lowang-bh: It is just a file, not dir.

camilamacedo86 (approved):
/lgtm

/approved

/ok-to-test

PR #3594, opened 2023-09-06
📖 add tip for release notes

camilamacedo86:
## Description

Add tip for release notes


camilamacedo86 (commented):
This one is just to add a small tip for the release doc.
So, I do not think is valuble ask for someone wast time in review here. 
We can change in a follow up if required.

PR #3601, opened 2023-09-06
:sparkles: add github action to run make test

camilamacedo86:
## Description

Add to the scaffold github action to run the tests of the project
So that users are able to check it out


camilamacedo86:
/hold

Let's discuss if we should add in the default scaffold or create a new plugin

camilamacedo86:
We decided to not move forward due maintainability reasons.

PR #3616, opened 2023-09-12
🌱 fix and improve samples as add further coverage

camilamacedo86:
## Description

- Fix sample generation. project-v4 should not scaffold the apis with deploy image
- Add new sample to check project v4 with deploy image plugin and multigroup layout
- Add CI to run make test and lint for project-v4 samples using deploy image which are the layouts more complex
- Fix code source by removing unused and duplicate code from generate.sh


test/testdata/generate.sh
```diff
@@ -98,32 +98,23 @@ function scaffold_test_project {
...
 
-    if [ $project == "project-v3-multigroup" ] || [ $project == "project-v4-multigroup" ]; then
+    if [ $project == "project-v4-multigroup" ] || [ $project == "project-v4-multigroup-with-deploy-image" ] ; then
       $kb create api --version v1 --kind Lakers --controller=true --resource=true --make=false
       $kb create webhook --version v1 --kind Lakers --defaulting --programmatic-validation
     fi
-
-    # Call ALL optional plugins within multigroup layout to ensure that they can work within
-    header_text 'Creating Memcached API with deploy-image plugin ...'
-    $kb create api --group example.com --version v1alpha1 --kind Memcached --image=memcached:1.4.36-alpine --image-container-command="memcached,-m=64,-o,modern,-v" --image-container-port="11211" --run-as-user="1001" --plugins="deploy-image/v1-alpha" --make=false
```
varshaprasad96, line 108: @camilamacedo86 why remove this? Even if its v4, we still need to test if we scaffold out 2 APIs. Or am I missing the code for it somewhere out of the  diff files?

camilamacedo86: See that this code was duplicated and it was not called only for project-v4-withdeploy-image sample
Therefore, this code was scaffolding the APIs using deploy image for the project-v4 sample when should not.

We should only to do this scaffold for the samples which specific uses the deploy-image.

varshaprasad96 (approved):
/lgtm
/approve

PR #3617, opened 2023-09-12
🐛  Disabling timestamp for logrus

### LINKED ISSUE #3608: The logs has been output with [0000] so we should disable the timestamp

The logs has been output such as:

INFO[0000] Writing kustomize manifests for you to edit... 
INFO[0000] Writing scaffold for you to edit...          
INFO[0000] api/v1alpha1/memcached_webhook.go            
INFO[0000] api/v1alpha1/webhook_suite_test.go           
INFO[0000] Update dependencies:

When we would like to have 

INFO Writing kustomize manifests for you to edit... 
INFO Writing scaffold for you to edit...          
INFO api/v1alpha1/memcached_webhook.go            
INFO api/v1alpha1/webhook_suite_test.go           
INFO Update dependencies:

To achieve this goal we can disable the Timestamp, such as:

```go
logger := logrus.New()
	// Disable timestamps on the default TextFormatter
	logger.SetFormatter(&logrus.TextFormatter{
		DisableTimestamp: true,
	})
```

However, we must to looking for to check if we could not do this change centralised so that we would not need to do the same code across the whole project. 

### Reproducing this issue

_No response_

### KubeBuilder (CLI) Version

master

### PROJECT version

_No response_

### Plugin versions

_No response_

### Other versions

_No response_

### Extra Labels

_No response_

suhail34:
Created a Log funciton in utils.go file that has field disabletimestamp:true and used it all over the project
It fixes #3608


lowang-bh:
I think the log format can be initial, without import util in every package.

pkg/plugin/util/util.go
```diff
@@ -271,3 +273,11 @@ func HasFileContentWith(path, text string) (bool, error) {
 
 	return strings.Contains(string(contents), text), nil
 }
+
+func Log() *logrus.Logger {
```
camilamacedo86, line 277: What do you think about creating a cli/util/util.go file and relocating the log utility there? My rationale is that the plugin utils is designed for utilities common to the plugins. In contrast, the log utility is utilized throughout the entire CLI. Structuring it this way might provide a clearer distinction and organization.

suhail34: Okay i will do the suggested changes 
Thank you

camilamacedo86:
Closing this one since the PR https://github.com/kubernetes-sigs/kubebuilder/pull/3619 solves it.
however, feel free to re-open if you see a need to do so.

Issue #3185, opened 2023-01-26
Provide propose solution to add a helper for kubebuilder users distribute their solutions


camilamacedo86:
### What do you want to happen?

Currently, kubebuilder as a tool needs to provide its users with a solution to distribute the project made with it to their end-users. To know more about its motivations, see https://github.com/kubernetes-sigs/kubebuilder/discussions/3081. We have users creating helm charts from the project built to be able to achieve this goal. 

Therefore, the idea of this task would be to propose a design solution to address this requirement.
See some examples : https://github.com/kubernetes-sigs/kubebuilder/tree/master/designs

PR #3626, opened 2023-09-17
✨  (go/v4): Add method to allow users distribute the project

camilamacedo86:
## Description

- Add target to build the installer under dist/install.yaml
- Also, add e2e tests to ensure that installer will be working well as expected
- By last, add instruction into the README generate to make clear how to achieve this goal

## Proposed Solution

Just generate the YAML with all that should be installed (same kustomize build of make install and make deploy) so that the project can be distributed as usually it is done by many solutions like Operator Prometheus for example, where users can run: kubectl apply -f dist/install.yaml and maintainers can easily project the raw of the file. 

## Motivation
- https://github.com/kubernetes-sigs/kubebuilder/discussions/3074
- https://github.com/kubernetes-sigs/kubebuilder/discussions/3081

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/3185

IMPORTANT: For we are able to check out the changes of this PR we must to have here the fix addressed in the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/3627 . So this one is blocked until we are able to merge the fix

/hold

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,8 @@ lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
 
 .PHONY: yamllint
 yamllint:
-	@docker run --rm $$(tty -s && echo "-it" || echo) -v $(PWD):/data cytopia/yamllint:latest testdata -d '{extends: relaxed, rules: {line-length: {max: 120}}, ignore: "/testdata/project-v2/\n/testdata/project-v3/"}'
+	@files=$$(find testdata -name '*.yaml' ! -path 'testdata/*/dist/install.yaml'); \
+    	docker run --rm $$(tty -s && echo "-it" || echo) -v $(PWD):/data cytopia/yamllint:latest $$files -d "{extends: relaxed, rules: {line-length: {max: 120}}}" --no-warnings
 
 GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
 golangci-lint:
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -187,6 +187,16 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 	- $(CONTAINER_TOOL) buildx rm project-v3-builder
 	rm Dockerfile.cross
 
+.PHONY: build-installer
+build-installer: manifests generate kustomize ## Generate a consolidated YAML with CRDs and deployment.
+	mkdir -p dist
+	@if [ -d "config/crd" ]; then \
+		$(KUSTOMIZE) build config/crd > dist/install.yaml; \
+	fi
+	echo "---" >> dist/install.yaml  # Add a document separator before appending
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	$(KUSTOMIZE) build config/default >> dist/install.yaml
+
 ##@ Deployment
 
 ifndef ignore-not-found
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/readme.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/readme.go
@@ -51,7 +51,11 @@ func (f *Readme) SetTemplateDefaults() error {
 		codeFence("kubectl apply -k config/samples/"),
 		codeFence("kubectl delete -k config/samples/"),
 		codeFence("make uninstall"),
-		codeFence("make undeploy"))
+		codeFence("make undeploy"),
+		codeFence("make build-installer IMG=<some-registry>/{{ .ProjectName }}:tag"),
+		codeFence("kubectl apply -f https://raw.githubusercontent.com/<org>/{{ .ProjectName }}/"+
+			"<tag or branch>/dist/install.yaml"),
+	)
 
 	return nil
 }
@@ -108,6 +112,25 @@ You can apply the samples (examples) from the config/sample:
 %s
 
 **UnDeploy the controller from the cluster:**
+
+%s
+
+## Project Distribution
+
+Following are the steps to build the installer and distribute this project to users.
+
+1. Build the installer for the image built and published in the registry:
+
+%s
+
+NOTE: The makefile target mentioned above generates an 'install.yaml'
+file in the dist directory. This file contains all the resources built
+with Kustomize, which are necessary to install this project without
+its dependencies.
+
+2. Using the installer
+
+Users can just run kubectl apply -f <URL for YAML BUNDLE> to install the project, i.e.:
 
 %s
 
--- a/docs/book/src/component-config-tutorial/testdata/project/Makefile
+++ b/docs/book/src/component-config-tutorial/testdata/project/Makefile
@@ -123,6 +123,16 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 	- $(CONTAINER_TOOL) buildx rm project-v3-builder
 	rm Dockerfile.cross
 
+.PHONY: build-installer
+build-installer: manifests generate kustomize ## Generate a consolidated YAML with CRDs and deployment.
+	mkdir -p dist
+	@if [ -d "config/crd" ]; then \
+		$(KUSTOMIZE) build config/crd > dist/install.yaml; \
+	fi
+	echo "---" >> dist/install.yaml  # Add a document separator before appending
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	$(KUSTOMIZE) build config/default >> dist/install.yaml
+
 ##@ Deployment
 
 ifndef ignore-not-found
--- a/docs/book/src/component-config-tutorial/testdata/project/README.md
+++ b/docs/book/src/component-config-tutorial/testdata/project/README.md
@@ -66,6 +66,29 @@ make uninstall
 make undeploy
 ```
 
+## Project Distribution
+
+Following are the steps to build the installer and distribute this project to users.
+
+1. Build the installer for the image built and published in the registry:
+
+```sh
+make build-installer IMG=<some-registry>/project:tag
+```
+
+NOTE: The makefile target mentioned above generates an 'install.yaml'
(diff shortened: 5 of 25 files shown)


Makefile
```diff
@@ -91,7 +91,8 @@ lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
 
 .PHONY: yamllint
 yamllint:
-	@docker run --rm $$(tty -s && echo "-it" || echo) -v $(PWD):/data cytopia/yamllint:latest testdata -d "{extends: relaxed, rules: {line-length: {max: 120}}}" --no-warnings
+	@files=$$(find testdata -name '*.yaml' ! -path 'testdata/*/dist/install.yaml'); \
+	docker run --rm $$(tty -s && echo "-it" || echo) -v $(PWD):/data cytopia/yamllint:latest $$files -d "{extends: relaxed, rules: {line-length: {max: 120}}}" --no-warnings
```
camilamacedo86, line 96: ## Required because the lint fail for false scenarios with duplication of key "kind" in mapping for example

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go
```diff
@@ -61,7 +61,7 @@ namePrefix: {{ .ProjectName }}-
 #    someName: someValue
 
 resources:
-- ../crd
+#- ../crd
```
camilamacedo86, line 64: It is not under the scope of this PR
The fix done in https://github.com/kubernetes-sigs/kubebuilder/pull/3626 is here otherwise that fails with the grafana sample

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/rbac/role.go
```diff
@@ -0,0 +1,61 @@
+/*
```
camilamacedo86, line 1: It is not under the scope of this PR
The fix done in https://github.com/kubernetes-sigs/kubebuilder/pull/3626 is here otherwise that fails with the grafana sample

pkg/plugins/common/kustomize/v2/scaffolds/init.go
```diff
@@ -69,6 +69,9 @@ func (s *initScaffolder) Scaffold() error {
 		&rbac.AuthProxyService{},
 		&rbac.AuthProxyClientRole{},
 		&rbac.RoleBinding{},
+		// We need to create a Role because if the project
+		// has not CRD define the controller-gen will not generate this file
+		&rbac.Role{},
```
camilamacedo86, line 74: It is not under the scope of this PR
The fix done in https://github.com/kubernetes-sigs/kubebuilder/pull/3626 is here otherwise that fails with the grafana sample

camilamacedo86:
HI @dashanji, @Kavinjsir , @Jay-Madden, @LCaparelli, @vbehar, @varshaprasad96 @everettraven 

This one seems good to fly now. 
Feel free to check.

Also, see that we are testing the installer with the e2e tests.

camilamacedo86:
Removing the
/hold cancel

pkg/plugins/golang/v4/scaffolds/internal/templates/readme.go
```diff
@@ -111,6 +116,25 @@ You can apply the samples (examples) from the config/sample:
...
+
+Following the steps for you build the installer and to distribute this project for users.
+
+### Build the installer
+To build the installer for the image built and published in the registry:
+
+make build-installer IMG=<some-registry>/{{ .ProjectName }}:tag
+
+NOTE: The above makefile target will generate the install.yaml in the directory dist contains all resources
+build with Kustomize to install this project.
```
Kavinjsir, line 129:
```suggestion
NOTE: The makefile target mentioned above generates an `install.yaml` file in the dist directory. This file contains all the resources built with Kustomize, which are necessary to install this project.
```

pkg/plugins/golang/v4/scaffolds/internal/templates/readme.go
```diff
@@ -111,6 +116,25 @@ You can apply the samples (examples) from the config/sample:
 
 %s
 
+## Distributing your project
+
+Following the steps for you build the installer and to distribute this project for users.
```
Kavinjsir, line 121:
```suggestion
Following are the steps to build the installer and distribute this project to users.
```

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -187,6 +187,16 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
...
+.PHONY: build-installer
+build-installer: manifests generate kustomize ## Generate a consolidated YAML with CRDs and deployment.
+	mkdir -p dist
+	@if [ -d "config/crd" ]; then \
+		$(KUSTOMIZE) build config/crd > dist/install.yaml; \
+	fi
+	echo "---" >> dist/install.yaml  # Add a document separator before appending
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	$(KUSTOMIZE) build config/default >> dist/install.yaml
+
```
Kavinjsir, line 199: One open question here:

I'm wondering about the pros (or cons if necessary) of having a single file `install.yaml` that includes both the CRD manifests and corresponding k8s resource manifests to be applied together (`kubectl apply -f dist/install.yaml`).

For instance, when applying CRDs, one use case is: after installing CRDs, it might be helpful to run:
```shell
kubectl wait \
	--for condition=Established \
	--all CustomResourceDefinition \
	--namespace=monitoring
```

which gives the space for:
1. k8s API server to process and update the CRDs' schema
2. Resolve conflicts with existing resources
3. Check errors if existing in CRDs' definition

If that is NOT quite common OR if it is generally good and safe enough to apply CRDs with their controllers, I am also good with having such a single file.

pkg/plugins/golang/v4/scaffolds/internal/templates/readme.go
```diff
@@ -111,6 +116,25 @@ You can apply the samples (examples) from the config/sample:
...
+make build-installer IMG=<some-registry>/{{ .ProjectName }}:tag
+
+NOTE: The above makefile target will generate the install.yaml in the directory dist contains all resources
+build with Kustomize to install this project.
+
+### To install via the installer
+
+You must just run kubectl apply -f dist/install.yaml.
+Therefore, users can install this project if they have access to the repository by running:
+
```
Kavinjsir, line 135: Nits:


To install this project, simply run the following command:

```sh
kubectl apply -f dist/install.yaml
```

Additionally, users with access to the repository can install this project by executing:

```sh
kubectl apply -f https://raw.githubusercontent.com/<org>/project/<tag or branch>/dist/install.yaml
```

Kavinjsir (commented):
Looks good to me in general!
This is a great feature.
I just left some nit and one open question.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
camilamacedo86: We are testing it out with the e2e and are not facing any issues.
You can check that it is a common practice, for example; 

- The Prometheus Operator make available the bundle.yaml which has all: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/getting-started.md (you can see either that is how we install the Prometheus in the e2e tests)
- Cert-Manager: Also does the same. See that we can install by applying an YAML with all: https://cert-manager.io/docs/installation/kubectl/#steps ( Also, that is how we get it installed as well )

Regarding your concerns, all of them are done when we apply. For example, if the schema has an issue, it will fail when we try to apply and before the resource is persisted on the cluster side.

Therefore, I do not think that should be a concern.

pkg/plugins/golang/v4/scaffolds/internal/templates/readme.go
camilamacedo86: We are already adding above the make targets about how to install.
The goal here is just to show how the project can be distributed without the user needing to git clone the solution. 
I will see how we can improve the info to make it clear.

Kavinjsir: Sounds good!

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
Kavinjsir: Yeah, that makes sense to me now. The reason why I am concerning that is about how we can make migration idempotent, but it seems better to be a different topic.
Also double checked some other operators, which shows the common usage of installing by one manitest file.

camilamacedo86 (commented):
@Kavinjsir 

I just remembered that I need to push the changes.
Now we have them with your suggestions !!!

I appreciate a lot your help in review.
Please let me know what you think now.

Kavinjsir (approved):
Looks good to me!

dashanji:
@camilamacedo86
Thanks for the great feature. I think it's a good point to build a bundled YAML for distributing the kubebuilder operator.
I just wonder what's the difference between `kubectl apply -f https://raw.githubusercontent.com/<org>/project/<tag or branch>/dist/install.yaml` and `kubectl apply -k https://github/<org>/project/<tag or branch>/config/default`?
The gain I can think of is that the common kubernetes users don't need to study the kustomize grammar and a bundled YAML is very straightforward.

varshaprasad96 (approved):
Thanks Camila! This is a great addition!
/lgtm
/approve

varshaprasad96:
Looks like unit tests a failing. Would have to dig deeper, probably import issues, or version bumps in ginko?

camilamacedo86:
@varshaprasad96 

The unit tests are failing for unrelated reason, it was fixed here: https://github.com/kubernetes-sigs/kubebuilder/pull/3743 (coverage test bumped with go 1.19 instead of 1.20)
I closed and re-opened the PR to see if it will get master changes. If not, I just rebased, and we are done.

Really thank you for your help and support 🥇

PR #3627, opened 2023-09-17
🐛 (kustomize/v2, go/v4): fix role and kustomize build when no CRDs are added to the project

camilamacedo86:
## Description

A project might have no CRDs generate their for:
- We should scaffold the project with the config/default/kustomization.yaml with crd commented when we run init
- We should generate the basic Role since controller-gen is only able to do that when markers are found in the controllers
- We should uncomment the crd config from config/default/kustomization.yaml when api is generated.

### DIFF
--- a/docs/book/src/component-config-tutorial/testdata/project/config/rbac/role.yaml
+++ b/docs/book/src/component-config-tutorial/testdata/project/config/rbac/role.yaml
@@ -1,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app.kubernetes.io/name: clusterrole
+    app.kubernetes.io/instance: manager-role
+    app.kubernetes.io/component: rbac
+    app.kubernetes.io/created-by: project
+    app.kubernetes.io/part-of: project
+    app.kubernetes.io/managed-by: kustomize
+  name: manager-role
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "list", "watch"]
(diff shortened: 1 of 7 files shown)


docs/book/src/component-config-tutorial/testdata/project/config/rbac/role.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
varshaprasad96, line 2: @camilamacedo86 I'm confused on the need of cluster role separately before the crd and roles are created by controller-gen. Do we have an issue for this, to get more context? Currently, KB doesn't support fetching core types and not generating CR. If so, why do we need to create manager's RBAC preemptively?

camilamacedo86: Hi @varshaprasad96,

Thanks for your review and your help on this one

- controller-gen does indeed target config/rbac/role.yaml and not clusterroles (that must be scaffiold by kubebuilder), but I do not think that has a relation with the changes of this PR.
I will try better clarify: 

- The crux is: that not all projects will require CRDs. So, enabling CRDs by default in the scaffold might not be appropriate. Note that a project can exist without the need for its own APIs at all.
- This PR modifies the initial state of config/default/kustomization.yaml to comment out CRDs. [See the change](https://github.com/kubernetes-sigs/kubebuilder/pull/3627/files#diff-20b19ff8cca874cadba14de74e485c68d13fd42df3eb0e9dcb42decd4b442619).
If an API is created later, the CRD section will be uncommented automatically.

Hope this makes the PR's intention clear. Let me know if you have further questions!

varshaprasad96: Thanks for the detailed reply, @camilamacedo86! This makes sense and answers my question.

varshaprasad96 (approved):
/lgtm
/approve

PR #3629, opened 2023-09-17
:bug:  (kustomize/v2,go/v4): fix uncomment kustomize files config accordinly when webhooks are generate

camilamacedo86:
## Description

We should NOT let the webhook configuration be commented on when a webhook is created. Therefore, this PR ensures that when we run kubebuilder to create a webhook we will uncomment the required places under the config so that webhooks are enabled by default.

### DIFF
--- a/testdata/project-v4-multigroup/config/crd/kustomization.yaml
+++ b/testdata/project-v4-multigroup/config/crd/kustomization.yaml
@@ -19,12 +19,12 @@ resources:
 patches:
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix.
 # patches here are for enabling the conversion webhook for each CRD
-#- path: patches/webhook_in_crew_captains.yaml
-#- path: patches/webhook_in_ship_frigates.yaml
-#- path: patches/webhook_in_ship_destroyers.yaml
-#- path: patches/webhook_in_ship_cruisers.yaml
-#- path: patches/webhook_in_sea-creatures_krakens.yaml
-#- path: patches/webhook_in_sea-creatures_leviathans.yaml
+- path: patches/webhook_in_crew_captains.yaml
+- path: patches/webhook_in_ship_frigates.yaml
+- path: patches/webhook_in_ship_destroyers.yaml
+- path: patches/webhook_in_ship_cruisers.yaml
+- path: patches/webhook_in_sea-creatures_krakens.yaml
+- path: patches/webhook_in_sea-creatures_leviathans.yaml
 #- path: patches/webhook_in_foo.policy_healthcheckpolicies.yaml
 #- path: patches/webhook_in_foo_bars.yaml
 #- path: patches/webhook_in_fiz_bars.yaml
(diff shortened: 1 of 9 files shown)


testdata/project-v4-multigroup/config/crd/kustomization.yaml
```diff
@@ -19,12 +19,12 @@ resources:
...
-#- path: patches/webhook_in_ship_cruisers.yaml
-#- path: patches/webhook_in_sea-creatures_krakens.yaml
-#- path: patches/webhook_in_sea-creatures_leviathans.yaml
+- path: patches/webhook_in_crew_captains.yaml
+- path: patches/webhook_in_ship_frigates.yaml
+- path: patches/webhook_in_ship_destroyers.yaml
+- path: patches/webhook_in_ship_cruisers.yaml
+- path: patches/webhook_in_sea-creatures_krakens.yaml
+- path: patches/webhook_in_sea-creatures_leviathans.yaml
 #- path: patches/webhook_in_foo.policy_healthcheckpolicies.yaml
```
camilamacedo86, line 28: TODO:
we need to check why when is multigroup those are not uncommenting as well

Sajiyah-Salat:
I understood the issue. Can you please let me know how can I test this changes and know more about it. Also as you have said the issue is out of scope of this pr. So could you let me know what we can do in follow up pr.

testdata/project-v4-multigroup/config/crd/kustomization.yaml
camilamacedo86: Hi @Sajiyah-Salat,

> Also, as you have said, the issue is out of the scope of this PR. Could you let me know what we can address in a follow-up PR?
> Regards: [PR Discussion](https://github.com/kubernetes-sigs/kubebuilder/pull/3629#discussion_r1328065993)

In the sample project which is using a multi-group layout, we can see that some patches were not uncommented.
Why does this happen? 
Because those patches should not exist in the first place.
They have been created when we run `create api`, but should only be created in specific scenarios, those when we run `webhook create`. The change of this PR is in the webhook subcommands, therefore, it is not reflected in this case scenario.

However, since the bug found is not part of the context of this PR, I would like to refer to the issue created. I think there is a better place to discuss this. https://github.com/kubernetes-sigs/kubebuilder/issues/3630

camilamacedo86:
Hi @Sajiyah-Salat,

> I understood the issue. Can you please let me know how I can test these changes and learn more about them? 

The goal of this PR is to uncomment the kustomize files to enable webhooks when webhooks are scaffolded (`kubectl webhook create`). The e2e tests are testing the changes, you can check the code source and the changes made here to see that. Also, you can create a project, then create a webhook with the source code of this PR and ensure that the respective info commented will be uncommented at the end. 

> Also as you have said the issue is out of scope of this pr. So could you let me know what we can do in follow up pr.
See; https://github.com/kubernetes-sigs/kubebuilder/pull/3629#discussion_r1335078577

PR #3640, opened 2023-09-26
✨ (go/v4) Makefile go and container build args

redhatrises:
Add $GOFLAGS and $CONTAINER_BUILD_ARGS variables to make it easier to pass build settings when using `make`.

### DIFF
--- a/docs/book/src/component-config-tutorial/testdata/project/Makefile
+++ b/docs/book/src/component-config-tutorial/testdata/project/Makefile
@@ -20,6 +20,9 @@ GOFLAGS ?=
 # tools. (i.e. podman)
 CONTAINER_TOOL ?= docker
 
+# CONTAINER_BUILD_ARGS defines additional build arguments to pass to the $CONTAINER_TOOL during build.
+CONTAINER_BUILD_ARGS ?=
+
 # Setting SHELL to bash allows bash commands to be executed by recipes.
 # Options are set to exit when a recipe line exits non-zero or a piped command fails.
 SHELL = /usr/bin/env bash -o pipefail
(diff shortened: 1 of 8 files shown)


docs/book/src/component-config-tutorial/testdata/project/Makefile
```diff
@@ -11,12 +11,18 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# GOFLAGS defines additional golang build flags like -tags and -ldflags
+GOFLAGS ?=
```
camilamacedo86, line 15: Would we have any reason to use go flags by default? 
If so, what are the reasons? What are the flags?

docs/book/src/component-config-tutorial/testdata/project/Makefile
```diff
@@ -11,12 +11,18 @@ else
...
+
 # CONTAINER_TOOL defines the container tool to be used for building images.
 # Be aware that the target commands are only tested with Docker which is
 # scaffolded by default. However, you might want to replace it to use other
 # tools. (i.e. podman)
 CONTAINER_TOOL ?= docker
 
+# CONTAINER_BUILD_ARGS defines additional build arguments to pass to the $CONTAINER_TOOL during build.
+CONTAINER_BUILD_ARGS ?=
```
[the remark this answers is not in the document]
redhatrises: Kinda some of the same reasons as above.

docs/book/src/component-config-tutorial/testdata/project/Makefile
redhatrises: One of the most common use cases is with versioning e.g. `--ldflags="-X 'github.com/mycorp/mycorp-operator/version.Version=$(VERSION)'"`. There are other projects that pass along `-mod=vendor` or some other go compilation flags or passing tags during build to enable/disable feature builds. With the version example, maybe it also makes sense to scaffold it by default, but it might not.

camilamacedo86: That is fine. 
but why we should incorporate those in the default scaffold?
will those be required for any use case?

camilamacedo86:
Hi @ redhatrises

We really appreciated your contribution 🥇 
But due the fact of the default scaffolds does not need this option it seems hard justify this addition

I hope that you do not mind but I think we can close this one. 
However, if you see that fits please feel free to re-open this one.

PR #3659, opened 2023-10-14
Delete docs/book/src/TODO.md

pierreozoux:


docs/book/src/TODO.md
```diff
@@ -1,8 +0,0 @@
-# TODO
```
camilamacedo86, line 1: What is your motivation to remove this one?

pierreozoux: this file is a plceholder that is not used yet, just that :)

camilamacedo86: It is used when we face 404, see: https://github.com/kubernetes-sigs/kubebuilder/blob/ce24b89a23fb8a51be78b115c15fd0f08975b46a/docs/book/src/404.md?plain=1

camilamacedo86: As other places see; https://github.com/search?q=repo%3Akubernetes-sigs%2Fkubebuilder%20TODO.md&type=code

PR #3670, opened 2023-10-24
✨ Add e2e tests scaffolded in default layout

### LINKED ISSUE #3605: Add e2e tests scaffolded in the default layout

The goal of this task is scaffold in the default layout e2e tests such as: https://github.com/operator-framework/operator-sdk/tree/master/testdata/go/v4-alpha/memcached-operator/test

The implementation should be:
- a) For go/v4 code implementation: we need to cover the Operator itself 
- b) For deploy-image plugin we need increment the tests to add more checks regards the deployment scaffolded by the plugin 

Also, we must ensure that we check the e2e tests with the ci, so we not allow get PR merges that broken then. To do so we can run an ci which will call the tests scaffold for project-v4 and project-v4-with-deploy-image under the testdata. 

**Motivation** Users does not know how to implement e2e tests in the projects. Also, provide to them it done since it is possible. The e2e tests can be configured to run against kind since we need to load the image. We also should add a TODO(user) when we load the image in the kind to explain to users that they need to change that if they do not want to run the tests using kind. 

Example motivation: https://kubernetes.slack.com/archives/CAW0GV7A5/p1694137817036729

### Extra Labels

_No response_

Lauquik:
## Description

adding testing scaffold for the default layout of our project

Test perfoms the follwoing tasks
- Building the Operator Image:
- Loading the Operator Image to Kind:
- Deploying the Controller-Manager:
- Validating Controller-Manager Pod:
## Fixes
#3605

### DIFF
--- a/.github/workflows/lint-sample.yml
+++ b/.github/workflows/lint-sample.yml
@@ -51,4 +51,3 @@ jobs:
           working-directory: testdata/project-v4-multigroup-with-deploy-image
           args: --config .golangci.yml ./...
           skip-cache: true  # first lint action will handle
-
--- a/.github/workflows/test-sample-go.yml
+++ b/.github/workflows/test-sample-go.yml
@@ -1,0 +1,51 @@
+name: project-v4-sample
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  test:
+    name: Run on Ubuntu
+    runs-on: ubuntu-latest
+    env:
+      KIND_K8S_VERSION: v1.28.0
+      tools_k8s_version: 1.28.0
+      kind_version: 0.15.0
+    steps:
+      - name: Clone the code
+        uses: actions/checkout@v4
+
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '~1.20'
+
+      - name: Install Kind
+        run: go install sigs.k8s.io/kind@v$kind_version
+  
+
+      - name: Install setup-envtest
+        run: go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+
+      - name: Install e2e tools with setup-envtest
+        run: setup-envtest use $tools_k8s_version
+
+      - name: Create kind cluster
+        run: kind create cluster
+        
+      - name: Prepare the environment
+        run: |
+          KUSTOMIZATION_FILE_PATH="testdata/project-v4/config/default/kustomization.yaml"
+          sed -i '25s/^#//' $KUSTOMIZATION_FILE_PATH
+          sed -i '27s/^#//' $KUSTOMIZATION_FILE_PATH
+          sed -i '42s/^#//' $KUSTOMIZATION_FILE_PATH
+          sed -i '46,143s/^#//' $KUSTOMIZATION_FILE_PATH
+      
+      - name: Test
+        run: |
+          cd testdata/project-v4
+          go get -u ./...
+          go mod tidy
+          make test-e2e
+          
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/test/utils/utils.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/test/utils/utils.go
@@ -1,0 +1,166 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package utils
+
+import (
+	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
+)
+
+var _ machinery.Template = &Utils{}
+
+type Utils struct {
+	machinery.TemplateMixin
+	machinery.BoilerplateMixin
+}
+
+func (f *Utils) SetTemplateDefaults() error {
+	if f.Path == "" {
+		f.Path = "test/utils/utils.go"
+	}
+
+	f.TemplateBody = utilsTemplate
+
+	return nil
+}
+
+var utilsTemplate = `{{ .Boilerplate }}
+
+package utils
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2" //nolint:golint,revive
+)
+
+const (
+	prometheusOperatorVersion = "v0.68.0"
+	prometheusOperatorURL     = "https://github.com/prometheus-operator/prometheus-operator/" +
+		"releases/download/%s/bundle.yaml"
+
+	certmanagerVersion = "v1.5.3"
+	certmanagerURLTmpl = "https://github.com/jetstack/cert-manager/releases/download/%s/cert-manager.yaml"
+)
+
+func warnError(err error) {
+	fmt.Fprintf(GinkgoWriter, "warning: %v\n", err)
+}
+
+// InstallPrometheusOperator installs the prometheus Operator to be used to export the enabled metrics.
+func InstallPrometheusOperator() error {
+	url := fmt.Sprintf(prometheusOperatorURL, prometheusOperatorVersion)
+	cmd := exec.Command("kubectl", "create", "-f", url)
+	_, err := Run(cmd)
+	return err
+}
+
+// Run executes the provided command within this context
+func Run(cmd *exec.Cmd) ([]byte, error) {
+	dir, _ := GetProjectDir()
+	cmd.Dir = dir
+
+	if err := os.Chdir(cmd.Dir); err != nil {
+		fmt.Fprintf(GinkgoWriter, "chdir dir: %s\n", err)
+	}
+
+	cmd.Env = append(os.Environ(), "GO111MODULE=on")
+	command := strings.Join(cmd.Args, " ")
+	fmt.Fprintf(GinkgoWriter, "running: %s\n", command)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return output, fmt.Errorf("%s failed with error: (%v) %s", command, err, string(output))
+	}
+
+	return output, nil
+}
+
+// UninstallPrometheusOperator uninstalls the prometheus
+func UninstallPrometheusOperator() {
+	url := fmt.Sprintf(prometheusOperatorURL, prometheusOperatorVersion)
+	cmd := exec.Command("kubectl", "delete", "-f", url)
+	if _, err := Run(cmd); err != nil {
+		warnError(err)
+	}
+}
+
+// UninstallCertManager uninstalls the cert manager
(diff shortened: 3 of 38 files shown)


pkg/plugins/golang/v4/scaffolds/internal/templates/test/utils/utils.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// Run executes the provided command within this context
+func Run(cmd *exec.Cmd) ([]byte, error) {
+	dir, _ := GetProjectDir()
+	cmd.Dir = dir
+	fmt.Fprintf(GinkgoWriter, "running dir: %s\n", cmd.Dir)
+
+	// To allow make commands be executed from the project directory which is subdir on SDK repo
+	// TODO:(user) You might not need the following code
```
camilamacedo86, line 62: It should be removed the end project does not need it at all.
We just have each in the operator-sdk project to be able to run the tests due how that is configured there.

test/e2e/ci.sh
```diff
@@ -24,3 +24,13 @@ create_cluster ${KIND_K8S_VERSION}
...
+
+KUSTOMIZATION_FILE_PATH="$(dirname "$0")/../../testdata/project-v4/config/default/kustomization.yaml"
+
+sed -i '25s/^#//' $KUSTOMIZATION_FILE_PATH
+sed -i '27s/^#//' $KUSTOMIZATION_FILE_PATH
+sed -i '42s/^#//' $KUSTOMIZATION_FILE_PATH
+sed -i '46,143s/^#//' $KUSTOMIZATION_FILE_PATH
+
+cd $(dirname "$0")/../../testdata/project-v4
+make test-e2e
```
camilamacedo86, line 36: This one is called to do the prow tests
I think we need a new configuration to perform this specific tests
Ideally, use GitHub Actions instead. 
So that would result in a new GitHub action that will setup the kind and run the tests scaffold for the sample

test/testdata/test.sh
```diff
@@ -23,6 +23,7 @@ function test_project {
 
   header_text "Performing tests in dir $1"
   pushd "$(dirname "$0")/../../testdata/$1"
+  rm -rf test/
```
camilamacedo86, line 26: This check ensures that the testdata samples are updated with the latest changes.
Why are we removing tests/ dir from here?

testdata/project-v4-multigroup-with-deploy-image/api/crew/v1/webhook_suite_test.go
```diff
@@ -73,7 +73,7 @@ var _ = BeforeSuite(func() {
 		// Note that you must have the required binaries setup under the bin directory to perform
 		// the tests directly. When we run make test it will be setup and used automatically.
 		BinaryAssetsDirectory: filepath.Join("..", "..", "..", "bin", "k8s",
-			fmt.Sprintf("1.28.3-%s-%s", runtime.GOOS, runtime.GOARCH)),
+			fmt.Sprintf("1.28.0-%s-%s", runtime.GOOS, runtime.GOARCH)),
```
camilamacedo86, line 76: You need rebase your PR with master changes.

camilamacedo86 (changes requested):
Great work, but I think we still require some adjusts
- Rebase with master 
- Do not change the current e2e tests to try run the sample tests. The e2e tests executed run the e2e tests scaffolds for kubebuilder itself. We need a new GitHub action that will run the sample tests and be configured with kind
- Ensure that the final scaffold has only what is required for end users.

test/testdata/test.sh
Lauquik: if the test dir is not removed 
then scaffolded e2e tests will run for all the sample which we don't require right? we only want to test one sapmle(project-v4)

camilamacedo86: The testdata check if all samples are == they should be with the latest changes.
We are changing the default scaffold so all v4 samples should have the tests now.

We do not need to execute and run the e2e tests against all projects
If we see that does not take too long that is fine would be great. But if we see that take too long then we just test 1/2 samples. But that is related to the new GitHub action and not the change test data check.

is that make sense?

test/e2e/sample-e2e.sh
```diff
@@ -0,0 +1,35 @@
...
+KUSTOMIZATION_FILE_PATH="$(dirname "$0")/../../testdata/project-v4/config/default/kustomization.yaml"
+
+sed -i '25s/^#//' $KUSTOMIZATION_FILE_PATH
+sed -i '27s/^#//' $KUSTOMIZATION_FILE_PATH
+sed -i '42s/^#//' $KUSTOMIZATION_FILE_PATH
+sed -i '46,143s/^#//' $KUSTOMIZATION_FILE_PATH
+
+cd $(dirname "$0")/../../testdata/project-v4
+go get -u ./...
+make test-e2e
```
camilamacedo86, line 35: Nit: add a space at the end of the file.

test/e2e/ci.sh
```diff
@@ -23,4 +23,4 @@ export KIND_CLUSTER="kind"
 create_cluster ${KIND_K8S_VERSION}
 trap delete_cluster EXIT
 
-test_cluster
+test_cluster
```
camilamacedo86, line 26: Nit: add a space at the end of the file.

.github/workflows/test-sample-go.yml
```diff
@@ -0,0 +1,31 @@
...
+
+# Trigger the workflow on pull requests and direct pushes to any branch
+on:
+  push:
+  pull_request:
+
+jobs:
+  test:
+    name: ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
```
camilamacedo86, line 11: We do not need a matrix.os here
We just do the test using ubuntu linux for this one.

Makefile
```diff
@@ -152,10 +152,15 @@ test-e2e-ci: ## Run the end-to-end tests (used in the CI)`
...
+	cd ./docs/book/src/multiversion-tutorial/testdata/project && rm -rf test && make test
 
 .PHONY: test-license
 test-license:  ## Run the license check
 	./test/check-license.sh
+
+.PHONY: test-e2e-sample
+test-e2e-sample:
+	./test/e2e/sample-e2e.sh
+
```
camilamacedo86, line 166: Nit extra space

test/e2e/sample-e2e.sh
```diff
@@ -0,0 +1,35 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+source "$(dirname "$0")/../common.sh"
+source "$(dirname "$0")/setup.sh"
+
+build_sample_external_plugin
+
+export KIND_CLUSTER="kind"
```
camilamacedo86, line 22: I do not think that will work
Instead of create a new sample-e2e.sh
You can use github actions to install the kind
Then you go into the sample dir and run go mod tidy and make, and by last, the command make test-e2e to run the tests against kind. 

is that make sense?

Lauquik:
@camilamacedo86 i have created a whole new workflow for sample project-v4 and removed the sample-e2e.sh script, is it okay? or require any changes, i'm open to receiving any suggestion and guidance

camilamacedo86:
Hi @Lauquik,

> @camilamacedo86 i have created a whole new workflow for sample project-v4 and removed the sample-e2e.sh script, is it okay? or require any changes, i'm open to receiving any suggestion and guidance

- The samples should continue to be generated as they currently are.
- The existing checks should remain operational in their current form.

That said, to ensure the integrity of our new e2e tests, we need a dedicated GitHub Action. While it's not necessary to test every single sample (and if doing so simplifies the process, that's acceptable), a streamlined approach might be:

1. Set up a new GitHub Action.
2. Have this action install kind.
3. Run the e2e tests that have been scaffolded for example for project-v4.

.github/workflows/test-sample-go.yml
```diff
@@ -0,0 +1,50 @@
...
+          sed -i '27s/^#//' $KUSTOMIZATION_FILE_PATH
+          sed -i '42s/^#//' $KUSTOMIZATION_FILE_PATH
+          sed -i '46,143s/^#//' $KUSTOMIZATION_FILE_PATH
+      
+      - name: Test
+        run: |
+          cd testdata/project-v4
+          go get -u ./...
+          go mod tidy
+          make test-e2e
```
camilamacedo86, line 50: Nit: missing space

Makefile
```diff
@@ -152,9 +152,9 @@ test-e2e-ci: ## Run the end-to-end tests (used in the CI)`
 
 .PHONY: test-book
 test-book: ## Run the cronjob tutorial's unit tests to make sure we don't break it
-	cd ./docs/book/src/cronjob-tutorial/testdata/project && make test
-	cd ./docs/book/src/component-config-tutorial/testdata/project && make test
-	cd ./docs/book/src/multiversion-tutorial/testdata/project && make test
+	cd ./docs/book/src/cronjob-tutorial/testdata/project && rm -rf test && make test
+	cd ./docs/book/src/component-config-tutorial/testdata/project && rm -rf test && make test
+	cd ./docs/book/src/multiversion-tutorial/testdata/project && rm -rf test && make test
```
camilamacedo86, line 157: Why did you remove the tests from here?
Is the make test calling the new e2e tests?
If so, the best solution is not call the e2e tests in the make test

.github/workflows/lint-sample.yml
```diff
@@ -21,9 +21,9 @@ jobs:
       - name: Remove pre-installed kustomize
         run: sudo rm -f /usr/local/bin/kustomize
       - name: Run make test for project-v4-with-deploy-image
-        run: cd testdata/project-v4-with-deploy-image && go mod tidy && make all && make test
+        run: cd testdata/project-v4-with-deploy-image && rm -rf test && go mod tidy && make all && make test
```
camilamacedo86, line 24: same here, we should not remove the tests from the lint

camilamacedo86 (commented):
Hi @Lauquik 

Very terrific great work 🥇 
I think we have just a couple of nits to sort out, and then it is good to fly.

test/testdata/test.sh
camilamacedo86: Why the e2e tests will run for all samples?
We are calling make test, right?

Also, see that it is not a problem to run the e2e tests for all samples (unless it would take too long)
The idea to run for one for then is to simplify the implementation.

Makefile
Lauquik: yeah, whenever make test is called (https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/Makefile#L65 )
at the end of the line there is go test ./... which will run the go test in every directory including the e2e dir where test suite is present

Lauquik:
@camilamacedo86 i think you are right removing the test dir is not the best approach, i will think of some other solution

Lauquik:
@camilamacedo86  I have modified the scaffolded Makefile to exclude the e2e test when running 'make test,' so there's no need to remove the test directories explicitly

camilamacedo86 (approved):
Great work 🥇 
It is great for me so I am approving this one

/approved

Let's give some time to see if others also want to do some review.
If nobody came with a second review or objections we can lgtm this one.

Kavinjsir (approved):
Great job @Lauquik 👍🏼  The PR Looks good to me!

camilamacedo86 (approved):
/lgtm

It seems good to fly
we can always push improvements and etc in a follow up

PR #3679, opened 2023-11-03
✨ Bump k8s, etcd and golang to generate new tools/envtest versions

camilamacedo86:
**Description**

- Bump k8s version from v28.0 to v28.3
- Bump windows with Golang from 2.19 to 2.20
- Update ETCD from v3.5.9 to v3.5.10


camilamacedo86 (commented):
Moving forward since we are just bumping the versions and those are required for the latest release.

PR #3683, opened 2023-11-06
:book: README.md: Fix first subheading under the Kubebuilder heading

renormalize:
* Heading `Kubebuilder` starts off with explaining Kubebuilder is a framework
* Subheading which explains Kubebuilder can also be used as a library was previously `Kubebuilder is also a framework`, now changed to `Kubebuilder is also a library` making it more coherent.


camilamacedo86 (approved):
I am OK with the proposed change

/lgtm
/ok-to-test

Issue #3697, opened 2023-11-20
Add support for configuring `boilerplatePath` during API scaffolding on plugin `base.go.kubebuilder.io/v4`


ardikabs:
### What do you want to happen?

*Description*:

Currently, I've configured my operator to adhere to an opinionated directory structure preference, causing me to execute `kubebuilder` with a non-default plugin, which is `base.go.kubebuilder.io/v4`. However, `kubebuilder` seems to define a hardcoded boilerplate path attribute in `hack/boilerplate.go.txt` during API scaffolding, as can be seen below at line 70:
https://github.com/kubernetes-sigs/kubebuilder/blob/26ee6711f7cd126d3490f1dfb8e2e648ddc60fac/pkg/plugins/golang/v4/scaffolds/api.go#L66-L77

*Feature Request*:

Could it be configured to accommodate an opinionated directory structure when attempting to satisfy certain preferences? say i expect the boilerplate refer to `tools/hack/boilerplate.go.txt`, as compared to `controller-gen` tool that could refer to a custom `headerFile` like below:
```bash
controller-gen object:headerFile="/path/to/boilerplate.go.txt"`
```

*Plugin*: `base.go.kubebuilder.io/v4`

camilamacedo86:
Hi @ardikabs,

Thank you for bringing up this issue. I agree that the ability to customize all paths, not just specific ones, would be a valuable addition to Kubebuilder. A config file allowing users to define paths for controllers, hack, etc., would indeed enhance the tool's flexibility.

However, introducing such customization for only one aspect without addressing the others would lead to inconsistency in behavior. We aim for a uniform and predictable user experience. Therefore, implementing this feature across the board is crucial, albeit it introduces significant complexities.

Given these considerations, the best approach would be to develop a detailed proposal. This document would outline the implementation strategy, consider potential problems, and propose solutions. You can find a template for proposals here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/template.md, and some examples at: https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs.

As a temporary measure for the current issue, one possibility could be to issue a warning instead of an error when the boilerplate is not found, though this might be perceived as a bug by some users. However, it might be an acceptable interim solution.

We're definitely open to contributions that would help us move in this direction. 
Would you be interested in working on this feature?

ardikabs:
Hi @camilamacedo86 , seems I missed the notification, sorry for the late response.

[...]

PR #3716, opened 2023-12-21
✨ feat(go/v4): add support for custom boilerplate path

ardikabs:
> Related to #3697 

## Description

Add support to a custom boilerplate path for the `base.go.kubebuilder.io/v4` plugin.

The following are changes involved in this feature request:
1. add `--boilerplatePath` flag to `kubebuilder --plugin "base.go.kubebuilder.io/v4" init`
2. add `--boilerplatePath` flag to `kubebuilder --plugin "base.go.kubebuilder.io/v4" create api`
3. add `--boilerplatePath` flag to `kubebuilder --plugin "base.go.kubebuilder.io/v4" create webhook`

## Expectation

After applying this feature, you can customize the boilerplate file path to any path other than the default (`hack/boilerplate.go.txt`), like the following examples:

* `Init`
```bash
kubebuilder init --plugins base.go.kubebuilder.io/v4 --license apache2 --owner example.com --boilerplatePath tools/hack/boilerplate.go.txt
```

* `Create API`
```bash
kubebuilder create api --controller=false --group example --kind CustomPlugin --version v1alpha1 --boilerplatePath tools/hack/boilerplate.go.txt
```

* `Create Webhook`
```bash
kubebuilder create webhook --group example --kind CustomPlugin --version v1alpha1 --defaulting --programmatic-validation --boilerplatePath tools/hack/boilerplate.go.txt
```

### DIFF
--- a/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
+++ b/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
@@ -321,7 +321,7 @@ func (s *apiScaffolder) scaffoldCreateAPIFromGolang(isLegacyLayout bool) error {
 		return golangV3Scaffolder.Scaffold()
 	}
 	golangV4Scaffolder := golangv4scaffolds.NewAPIScaffolder(s.config,
-		s.resource, true)
+		s.resource, golangv4scaffolds.DefaultBoilerplatePath, true)
 	golangV4Scaffolder.InjectFS(s.fs)
 	return golangV4Scaffolder.Scaffold()
 }
--- a/pkg/plugins/golang/v4/api.go
+++ b/pkg/plugins/golang/v4/api.go
@@ -37,10 +37,10 @@ import (
 const (
 	// defaultCRDVersion is the default CRD API version to scaffold.
 	defaultCRDVersion = "v1"
-)
 
-// DefaultMainPath is default file path of main.go
-const DefaultMainPath = "cmd/main.go"
+	// DefaultMainPath is default file path of main.go
+	DefaultMainPath = "cmd/main.go"
+)
 
 var _ plugin.CreateAPISubcommand = &createAPISubcommand{}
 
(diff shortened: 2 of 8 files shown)


ardikabs:
As I am new to the kubebuilder codebase, please be advised on where I should put a test for this feature. Thanks!

camilamacedo86 (changes requested):
The change cannot be made within the API command. Doing so would mean that we would be able to define a different value for each API.

If we allow customization, it must be done via the kubebuilder init and kubebuilder edit commands. Additionally, the path must be stored in the PROJECT file.

Could you please take a look at it and see if you have any concerns?

varshaprasad96:
+1 to what @camilamacedo86 mentioned. Allowing boilerplate configured differently during these commands would lead to confusion of:
(1) Single project having multiple licenses.
(2) Certain commands like `create webhook` also modify existing files - causing conflict on boilerplate usage.
Though (1) is still possible, its a scenario where user can manually change the boilerplate based on their needs as it seems to be an edge case.

camilamacedo86:
> As I am new to the kubebuilder codebase, please be advised on where I should put a test for this feature. Thanks!

See the e2e tests: https://github.com/kubernetes-sigs/kubebuilder/tree/master/test/e2e

To create the tests that would be new dir with a test like the others just to check the layout after the change.
Following the steps that I see here: 

- create a project kb init --header-file-path=<path>
- then, ensure that the hack/dir is not created and instead the new path
- after that, ensure that the makefile is updated accordingly
- Call generated and ensured that the header of the files had the header of the path informed. 
Lastly, it is properly configured in the ProjectConfig via a new spec.

camilamacedo86:
Hi @ardikabs 

I am closing this one since, as we spoke if we would like to add this feature, we need to add it for the init|edit subcommand and keep the data stored in the PROJECT file so that it can be used across since it is valid for the whole project. We cannot have one specific config for each API.

Therefore, I hope that you do not mind if we close this one, and please feel free to raise a new PR with the proposed solution or reopen this one.

PR #3718, opened 2024-01-02
:sparkles: (go/v4) : Ensure versioning of tools and binaries installed via Makefile

### LINKED ISSUE #3715: Standardize installation of golang tools with better version support

Hi,

i have small problem with default makefile where linter and envtest don't have logic for check of specific version:

https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go#L248
https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go#L139

it creates problem with upgrading of tools, where some developers still use old version because they didn't manually delete previous version.
Another additional problem linked with it is using of latest version for envtest. many developers use different version, if they join on project later.

We have small extension for makefile designed primary for shared cache in pipeline which can rename binary and our binaries names contain version too which solve problem with upgrading of tools.

```makefile
## Tool Binaries
KUSTOMIZE ?= $(LOCALBIN)/kustomize-$(KUSTOMIZE_VERSION)
ENVTEST ?= $(LOCALBIN)/setup-envtest-$(ENVTEST_VERSION)
GOLANGCI_LINT ?= $(LOCALBIN)/golangci-lint-$(GOLANGCI_LINT_VERSION)
KUSTOMIZE_VERSION ?= v5.3.0
ENVTEST_VERSION ?= release-0.15
GOLANGCI_LINT_VERSION ?= v1.55.2
....
.PHONY: envtest
envtest: $(LOCALBIN) ## Download envtest-setup locally if necessary.
	$(call go-install-tool,$(ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest,${ENVTEST_VERSION})
...
# go-install-tool will 'go install' any package with custom target and name of binary, if it doesn't exist
# $1 - target path with name of binary (ideally with version)
# $2 - package url which can be installed
# $3 - specific version of package
define go-install-tool
@[ -f $(1) ] || { \
set -e; \
version=@$(3) ;\
url=$(2) ;\
if echo $$version | grep 'v[2-9][0-9]*' -q; then \
        echo $$url | grep '/v[2-9][0-9]*/' -q || version="/$$(printf $$version | grep -o 'v[2-9][0-9]*')$$version" ;\
fi ;\
echo "Downloading $${url}$${version}" ;\
GOBIN=$(GOBIN) go install $${url}$${version} ;\
binary=$$(echo "$$url" | rev | cut -d'/' -f 1 | rev) ;\
mv "$(GOBIN)/$${binary}" $(1) ;\
}
endef
```

Local bin folder:

```bash
$ ls bin
controller-gen-v0.9.0  golangci-lint-v1.53.3  kustomize-v3.8.7  setup-envtest-release-0.15
```

What do you think about this improvement?

### Extra Labels

_No response_

lukas016:
Fixes #3715 

PR solves primary issue with synchronization of tools version between developers where older copies of repo can use older versions of binaries and it isn't updated by makefile.

PR unified way how tools are installed and adding better support for versioning of binary based on binary names.
It add new function into makefile and modified installation targets for tools.
It contains more specific version of tool setup-envtest, which was set to release-0.16 branch because this tools doesn't support proper versioning. It isn't ideally solution but still better as latest.

output:
```bash
ls bin 
controller-gen-v0.13.0     golangci-lint-v1.54.2      kustomize-v5.2.1           setup-envtest-release-0.16

make generate manifests undeploy test lint
.../example/bin/controller-gen-v0.13.0 object:headerFile="hack/boilerplate.go.txt" paths="./..."
.../example/bin/controller-gen-v0.13.0 rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
.../example/bin/kustomize-v5.2.1 build config/default | kubectl delete --ignore-not-found=false -f -
...
KUBEBUILDER_ASSETS=".../example/bin/k8s/1.28.3-darwin-arm64" go test $(go list ./... | grep -v /e2e) -coverprofile cover.out
...
.../example/bin/golangci-lint-v1.54.2 run
```

### DIFF
--- a/docs/book/src/component-config-tutorial/testdata/project/Makefile
+++ b/docs/book/src/component-config-tutorial/testdata/project/Makefile
@@ -68,14 +68,6 @@ test: manifests generate fmt vet envtest ## Run tests.
 .PHONY: test-e2e  # Run the e2e tests against a Kind k8s instance that is spun up.
 test-e2e:
 	go test ./test/e2e/ -v -ginkgo.v
-	
-GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
-GOLANGCI_LINT_VERSION ?= v1.54.2
-golangci-lint:
-	@[ -f $(GOLANGCI_LINT) ] || { \
-	set -e ;\
-	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell dirname $(GOLANGCI_LINT)) $(GOLANGCI_LINT_VERSION) ;\
-	}
 
 .PHONY: lint
 lint: golangci-lint ## Run golangci-lint linter & yamllint
@@ -143,10 +135,10 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
 
 .PHONY: undeploy
-undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
+undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
 	$(KUSTOMIZE) build config/default | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
 
-##@ Build Dependencies
+##@ Dependencies
 
 ## Location to install dependencies to
 LOCALBIN ?= $(shell pwd)/bin
@@ -155,30 +147,47 @@ $(LOCALBIN):
 
 ## Tool Binaries
 KUBECTL ?= kubectl
-KUSTOMIZE ?= $(LOCALBIN)/kustomize
-CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
-ENVTEST ?= $(LOCALBIN)/setup-envtest
+KUSTOMIZE ?= $(LOCALBIN)/kustomize-$(KUSTOMIZE_VERSION)
+CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen-$(CONTROLLER_TOOLS_VERSION)
+ENVTEST ?= $(LOCALBIN)/setup-envtest-$(ENVTEST_VERSION)
+GOLANGCI_LINT = $(LOCALBIN)/golangci-lint-$(GOLANGCI_LINT_VERSION)
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.2.1
 CONTROLLER_TOOLS_VERSION ?= v0.13.0
+ENVTEST_VERSION ?= release-0.16
+GOLANGCI_LINT_VERSION ?= v1.54.2
 
 .PHONY: kustomize
-kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary. If wrong version is installed, it will be removed before downloading.
+kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
 $(KUSTOMIZE): $(LOCALBIN)
-	@if test -x $(LOCALBIN)/kustomize && ! $(LOCALBIN)/kustomize version | grep -q $(KUSTOMIZE_VERSION); then \
-		echo "$(LOCALBIN)/kustomize version is not expected $(KUSTOMIZE_VERSION). Removing it before installing."; \
-		rm -rf $(LOCALBIN)/kustomize; \
-	fi
-	test -s $(LOCALBIN)/kustomize || GOBIN=$(LOCALBIN) GO111MODULE=on go install sigs.k8s.io/kustomize/kustomize/v5@$(KUSTOMIZE_VERSION)
+	$(call go-install-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v5,$(KUSTOMIZE_VERSION))
 
 .PHONY: controller-gen
-controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
+controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary.
 $(CONTROLLER_GEN): $(LOCALBIN)
-	test -s $(LOCALBIN)/controller-gen && $(LOCALBIN)/controller-gen --version | grep -q $(CONTROLLER_TOOLS_VERSION) || \
-	GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_TOOLS_VERSION)
+	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen,$(CONTROLLER_TOOLS_VERSION))
 
 .PHONY: envtest
-envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
+envtest: $(ENVTEST) ## Download setup-envtest locally if necessary.
 $(ENVTEST): $(LOCALBIN)
-	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+	$(call go-install-tool,$(ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest,$(ENVTEST_VERSION))
+
+.PHONY: golangci-lint
+golangci-lint: $(GOLANGCI_LINT) ## Download golangci-lint locally if necessary.
+$(GOLANGCI_LINT): $(LOCALBIN)
+	$(call go-install-tool,$(GOLANGCI_LINT),github.com/golangci/golangci-lint/cmd/golangci-lint,${GOLANGCI_LINT_VERSION})
+
+# go-install-tool will 'go install' any package with custom target and name of binary, if it doesn't exist
+# $1 - target path with name of binary (ideally with version)
+# $2 - package url which can be installed
+# $3 - specific version of package
+define go-install-tool
+@[ -f $(1) ] || { \
+set -e; \
+package=$(2)@$(3) ;\
+echo "Downloading $${package}" ;\
+GOBIN=$(LOCALBIN) go install $${package} ;\
+mv "$$(echo "$(1)" | sed "s/-$(3)$$//")" $(1) ;\
+}
+endef
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -17,6 +17,8 @@ limitations under the License.
 package templates
 
 import (
+	"strings"
+
 	"sigs.k8s.io/kubebuilder/v3/pkg/machinery"
 )
 
@@ -38,6 +40,8 @@ type Makefile struct {
 	KustomizeVersion string
 	// ControllerRuntimeVersion version to be used to download the envtest setup script
 	ControllerRuntimeVersion string
+	// ControllerRuntimeReleaseBranchName store the name of the branch to be used to install setup-envtest
+	ControllerRuntimeReleaseBranchName string
 }
 
 // SetTemplateDefaults implements file.Template
@@ -54,7 +58,28 @@ func (f *Makefile) SetTemplateDefaults() error {
 		f.Image = "controller:latest"
 	}
 
+	if f.ControllerRuntimeReleaseBranchName == "" {
+		f.ControllerRuntimeReleaseBranchName = getControllerRuntimeReleaseBranch(f.ControllerRuntimeVersion)
+	}
+
 	return nil
+}
+
+func getControllerRuntimeReleaseBranch(version string) string {
+	// Remove the "v" prefix
+	version = strings.TrimPrefix(version, "v")
+
+	// Split the version string into its components
+	parts := strings.SplitN(version, ".", 3)
+
+	// Check if we have at least two components (major and minor)
+	if len(parts) < 2 {
+		// return latest working version as invalid
+		return "latest"
+	}
+
+	// Return the formatted version
+	return "release-" + parts[0] + "." + parts[1]
 }
 
 //nolint:lll
@@ -132,14 +157,6 @@ test: manifests generate fmt vet envtest ## Run tests.
 .PHONY: test-e2e  # Run the e2e tests against a Kind k8s instance that is spun up.
 test-e2e:
 	go test ./test/e2e/ -v -ginkgo.v
-	
-GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
-GOLANGCI_LINT_VERSION ?= v1.54.2
-golangci-lint:
-	@[ -f $(GOLANGCI_LINT) ] || { \
-	set -e ;\
-	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell dirname $(GOLANGCI_LINT)) $(GOLANGCI_LINT_VERSION) ;\
-	}
 
 .PHONY: lint
 lint: golangci-lint ## Run golangci-lint linter & yamllint
@@ -207,10 +224,10 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
 
 .PHONY: undeploy
-undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
+undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
 	$(KUSTOMIZE) build config/default | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
 
-##@ Build Dependencies
+##@ Dependencies
 
 ## Location to install dependencies to
 LOCALBIN ?= $(shell pwd)/bin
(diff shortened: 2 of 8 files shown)


pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -132,14 +132,12 @@ test: manifests generate fmt vet envtest ## Run tests.
 .PHONY: test-e2e  # Run the e2e tests against a Kind k8s instance that is spun up.
 test-e2e:
 	go test ./test/e2e/ -v -ginkgo.v
-	
-GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
+
+GOLANGCI_LINT = $(LOCALBIN)/golangci-lint-$(GOLANGCI_LINT_VERSION)
```
mateusoliveira43, line 136: As a suggestion, what about 
- moving `GOLANGCI_LINT` to Tool Binaries section
- moving `GOLANGCI_LINT_VERSION` to Tool Versions section 
- and moving `golangci-lint` command to Build Dependencies section

This would let things more standardized :smile:

lukas016: i like this idea, i thought about it too. but linter isn't dependency for build target so i had problem with name of section 😅 
if someone else decide it, i will do that.

mateusoliveira43: good point
Maybe renaming **Build Dependencies** to just **Dependencies** to avoid this semantic problem

lukas016: fixed

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -219,31 +211,51 @@ $(LOCALBIN):
...
+
+.PHONY: golangci-lint
+golangci-lint: $(LOCALBIN)
+	$(call go-install-tool,$(GOLANGCI_LINT),github.com/golangci/golangci-lint/cmd/golangci-lint,${GOLANGCI_LINT_VERSION})
+
+# go-install-tool will 'go install' any package with custom target and name of binary, if it doesn't exist
+# $1 - target path with name of binary (ideally with version)
+# $2 - package url which can be installed
+# $3 - specific version of package
+define go-install-tool
```
mateusoliveira43, line 247: would simplify here (adding `/v5` back to kustomize url)
```make
define go-install-tool
@[ -f $(1) ] || { \
set -e; \
package=$(2)@$(3) ;\
echo "Downloading $${package}" ;\
GOBIN=$(LOCALBIN) go install $${package} ;\
mv "$$(echo "$(1)" | sed "s/-$(3)$$//")" $(1) ;\
}
endef
```

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -219,31 +211,51 @@ $(LOCALBIN):
...
-controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
+controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary.
 $(CONTROLLER_GEN): $(LOCALBIN)
-	test -s $(LOCALBIN)/controller-gen && $(LOCALBIN)/controller-gen --version | grep -q $(CONTROLLER_TOOLS_VERSION) || \
-	GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_TOOLS_VERSION)
+	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen,$(CONTROLLER_TOOLS_VERSION))
 
 .PHONY: envtest
-envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
-$(ENVTEST): $(LOCALBIN)
```
mateusoliveira43, line 247: For standardization, would add back `$(ENVTEST): $(LOCALBIN)` line here (and also add to golangci)

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -219,31 +211,51 @@ $(LOCALBIN):
...
-	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+envtest: $(LOCALBIN) ## Download envtest-setup locally if necessary.
+	$(call go-install-tool,$(ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest,$(ENVTEST_VERSION))
+
+.PHONY: golangci-lint
+golangci-lint: $(LOCALBIN)
```
mateusoliveira43, line 240: For standardization, would add doc comment here 
`## Download golangci-lint locally if necessary.`

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -219,31 +211,51 @@ $(LOCALBIN):
...
 .PHONY: controller-gen
-controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
+controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary.
 $(CONTROLLER_GEN): $(LOCALBIN)
-	test -s $(LOCALBIN)/controller-gen && $(LOCALBIN)/controller-gen --version | grep -q $(CONTROLLER_TOOLS_VERSION) || \
-	GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_TOOLS_VERSION)
+	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen,$(CONTROLLER_TOOLS_VERSION))
 
 .PHONY: envtest
-envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
```
mateusoliveira43, line 246: Should it be setup-envtest instead of envtest-setup?

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
lukas016: i would prefer keep go-install-tool general. Some tools can use more specific path behind version. So you will have different format for VERSION what can be confusing for user.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
lukas016: name of package is setup-envtest

lukas016: it is probably copy&paste error because older versions has envtest-setup too. I will fixed it.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
lukas016: here is example for ginkgo and helm:

```
CONTROLLER_GEN_VERSION:=v0.10.0
KUSTOMIZE_VERSION:=v4.5.7
GOLANGCI_LINT_VERSION:=v1.51.2
HELMIFY_VERSION:=v0.4.1
GINKGO_VERSION:=v2.9.0
# isn't possible use tag
ENVTEST_VERSION:=release-0.14
HELM_VERSION:=v3.11.1

.PHONY: ginkgo-install
ginkgo-install: ## Download ginkgo bin locally if necessary.
	$(call go-install-tool,$(GINKGO_BIN),github.com/onsi/ginkgo/v2/ginkgo,${GINKGO_VERSION})

HELM_BIN = $(GOBIN)/helm-${HELM_VERSION}
.PHONY: helm-install
helm-install:
	$(call go-install-tool,$(HELM_BIN),helm.sh/helm/v3/cmd/helm,$(HELM_VERSION))
```

lukas016: @mateusoliveira43 do you agree with current solution?

mateusoliveira43: I am little worried with the `if` about version, I think it is little hard to understand it (that is for kustomize case, right? what does `echo "$url" | rev | cut -d'/' -f 1 | rev` do? my suggestion to use `$$(echo "$(1)" | sed "s/-$(3)$//` would remove the suffix `-version` from path installation, so I think it would be still general use case by the project tools)

But we can wait for the opinion of maintainers (and I really liked the idea of centralizing the tools installation in a single function :smile:)

lukas016: yes, i agree it is little confusing. Name of binary is last part of url, this is reason why i need keep url and version separately and why i modified version in if statement. First argument is name of target binary but i need source name for mv command and i cannot trust user input.

mateusoliveira43: Since this is will be used by developers, not users of the products created using kubebuilder, I would not be so strict here (even my suggested implementation has flaws: if tool name does not have version, it fails with the error `mv: 'path' and 'path' are the same file`), focusing more in readability then validation. 

For a complex set of validations in the function, I would suggest using a programming language different from shell for that (in this case, Go), as it easier to write, read and maintain the code. But if developers follow the conventions of the Makefile (for example, to update version of a tool, just change the `_VERSION` variable instead of directly changing the tool name/path, or to add new tool, follow how the other tools are installed), they should have a nice experience with it :smile: 

Or, for a more strict way of doing (but still making a simple logic), create new binary name variables (for example, `KUSTOMIZE_NAME=kustomize`) that each path would use (for example, `KUSTOMIZE ?= $(LOCALBIN)/$(KUSTOMIZE_NAME)-$(KUSTOMIZE_VERSION)`) and the function would be like this
```Make
# go-install-tool will 'go install' any specific version of a package, if it doesn't exist
# $1 - binary name of the package
# $2 - package url which can be installed
# $3 - specific version of package

define go-install-tool
@[ -f "$(LOCALBIN)/$(1)-$(3)"] || { \
set -e; \
package=$(2)@$(3) ;\
echo "Downloading $${package}" ;\
GOBIN=$(LOCALBIN) go install $${package} ;\
mv "$(LOCALBIN)/$(1)" "$(LOCALBIN)/$(1)-$(3)" ;\
}
endef
```

lukas016: i think, we started add more complexity. i would like hear another opinion.
we can use this simple solution but we will add more lines of makefile as we will delete from shell script.

i have little worry from hide connection between KUSTOMIZE_NAME and binary file, but we can solve it with comment but it isn't very user friendly for me (developers are users for me too).  But i agree, i will need to add comment into shell.

+ We can look on this as temporary solution (maybe for 1 year or long 😄 ) because maybe golang will allow define custom name for binary: https://github.com/golang/go/issues/44469

lukas016: i look on your first suggestion again and it looks fine. Sorry i was probably little sleepy yesterday. I will test it tomorrow.

mateusoliveira43: yesss, if `go install` adds that feature, that simplifies a lot 

Yes, better to hear opinions from others!

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
lukas016: fixed

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
lukas016: fixed

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
lukas016: fixed, implemented according to your first suggestion

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
lukas016: @mateusoliveira43 pls could you close threads where are you ok with solution. thx

mateusoliveira43: Unfortunately  I do not have ability to resolve threads, I believe because I am not a maintainer of the project

But I think the PR is ready to merge :smile:

mateusoliveira43 (approved):
LGTM :white_check_mark:

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
lukas016: weird, i can resolve them (threads == conversation)

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -219,31 +211,48 @@ $(LOCALBIN):
...
 $(KUSTOMIZE): $(LOCALBIN)
-	@if test -x $(LOCALBIN)/kustomize && ! $(LOCALBIN)/kustomize version | grep -q $(KUSTOMIZE_VERSION); then \
-		echo "$(LOCALBIN)/kustomize version is not expected $(KUSTOMIZE_VERSION). Removing it before installing."; \
-		rm -rf $(LOCALBIN)/kustomize; \
-	fi
-	test -s $(LOCALBIN)/kustomize || GOBIN=$(LOCALBIN) GO111MODULE=on go install sigs.k8s.io/kustomize/kustomize/v5@$(KUSTOMIZE_VERSION)
+	$(call go-install-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v5,$(KUSTOMIZE_VERSION))
 
 .PHONY: controller-gen
-controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
```
camilamacedo86, line 240: Should we not keep ` If wrong version is installed, it will be overwritten.`? Is not it what this change does?

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -219,31 +211,48 @@ $(LOCALBIN):
...
-ENVTEST ?= $(LOCALBIN)/setup-envtest
+KUSTOMIZE ?= $(LOCALBIN)/kustomize-$(KUSTOMIZE_VERSION)
+CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen-$(CONTROLLER_TOOLS_VERSION)
+ENVTEST ?= $(LOCALBIN)/setup-envtest-$(ENVTEST_VERSION)
+GOLANGCI_LINT = $(LOCALBIN)/golangci-lint-$(GOLANGCI_LINT_VERSION)
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= {{ .KustomizeVersion }}
 CONTROLLER_TOOLS_VERSION ?= {{ .ControllerToolsVersion }}
+ENVTEST_VERSION ?= release-0.16 # branch name
```
camilamacedo86, line 222: What about we try to use here `ControllerRuntimeVersion` and remove the value after the first dot?
See that when we bump the controllerRuntime version we bump it here: https://github.com/varshaprasad96/kubebuilder/blob/e349099a9667e717e9be36d551c460c9a9ddfbe4/pkg/plugins/golang/v4/scaffolds/init.go#L32-L40

Otherwise, we might forget to bump the Makefile as well.

camilamacedo86 (commented):

Could you please:
a) Run `make generate` then the changes made will result in the samples under testdata and docs
b) Squash the commits. we can only merge when we have 1 commit in the PR

/ok-to-test

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
mateusoliveira43: No, the new function used to donwload all tools binaries basically appends the version to binary name

example: I installed kustomize version 1.2.3, it will be stored in `bin/` folder as `kustomize-1.2.3`. Later, I update kustomize version to 2.3.4, it will add `kustomize-2.3.4` to `bin/` folder, but will not delete `kustomize-1.2.3` from there (but Makefile will use the right binary)

we loose in clean up on `bin/` folder, but it makes things more simple to manage (for example, some binaries have `version` comamnd, some have `--version` flag, and some do not have any)

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
mateusoliveira43: that would be nice! basically use tag instead of branch (did not try if works :grimacing:)

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
lukas016: i will delete last part of comment

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
lukas016: if i remember correctly, there was problem with tags. i will check it.

lukas016: go install sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.16.3                                 
go: sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.16.3: module sigs.k8s.io/controller-runtime@v0.16.3 found, but does not contain package sigs.k8s.io/controller-runtime/tools/setup-envtest

Problem is controller-runtime repo. It is multimodules repository and they don't use proper versioning for submodules.

We can technically generate branch name from version because they use the same format or we can translate tag to commit hash.

lukas016: i created issue for them: https://github.com/kubernetes-sigs/controller-runtime/issues/2646

fix isn't so hard, if they want to keep the same version for controller-runtime and setup-envtest. They just have to create one more specific tag.

lukas016: @camilamacedo86 i think that discussion about setup-envtest will take long time and i don't like long-life PR. I suggest create separate issue for it and i can prepare functionality for generate branch name from ControllerRuntimeVersion as workaround.

Do you agree with my suggestion?

camilamacedo86:
Hi @varshaprasad96 

Could you please give a help here?
Are you OK with this one as well?
WDYT

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
camilamacedo86: Hi @lukas016 

I think we can achive that in a very simple one. Follow the code/draft: 

Add the func: 

```go
func getControllerRuntimeReleaseBranch(version string) string {
	// Remove the "v" prefix
	version = strings.TrimPrefix(version, "v")

	// Split the version string into its components
	parts := strings.Split(version, ".")

	// Check if we have at least two components (major and minor)
	if len(parts) < 2 {
		return "release-" + version // Return the whole version if it's not in 'major.minor.patch' format
	}

	// Return the formatted version
	return "release-" + parts[0]
}
```

- Then, we add the variable after `// ControllerRuntimeVersion version to be used to download the envtest setup script
	ControllerRuntimeVersion string`

```go
	// ControllerRuntimeReleaseBranchName store the name of the branch to be used to setup ENV TEST
	ControllerRuntimeReleaseBranchName string
```

- Populate the new variable, call after:

```
if f.Image == "" {
		f.Image = "controller:latest"
	}
```

```go
f len(f.ControllerRuntimeReleaseBranchName) == 0 {
		f.ControllerRuntimeReleaseBranchName = getControllerRuntimeReleaseBranch(f.ControllerRuntimeVersion)
	}

```

- And by last, just use it in the template {{ .ControllerRuntimeReleaseBranchName }}

WDYT?

lukas016: it looks fine as workaround.

varshaprasad96 (commented):
Thanks for the contribution! The changes look good to me. One thing to add is that the underlying KB binaries which are installed by setup-envtest would still remain until they are cleaned up. This is something we need not handle here, but just mentioning because a full cleanup would still not occur as mentioned in the issue. 
Thinking about it, an alternative to handle it is to do a `setup-envtest cleanup` if we identify a mismatched version from `use` command. But we can handle that in follow up.

@lukas016 we still need to run `make generate` to update the testdata.

lukas016:
> Thanks for the contribution! The changes look good to me. One thing to add is that the underlying KB binaries which are installed by setup-envtest would still remain until they are cleaned up. This is something we need not handle here, but just mentioning because a full cleanup would still not occur as mentioned in the issue. Thinking about it, an alternative to handle it is to do a `setup-envtest cleanup` if we identify a mismatched version from `use` command. But we can handle that in follow up.
> 
> @lukas016 we still need to run `make generate` to update the testdata.

I will implement suggestions today and i will call make generate too.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
camilamacedo86: That would make simple we keep it maintained
Also, please squash the commits so we can get this one merged. 

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
lukas016: fixed

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
lukas016: fixed

lukas016:
i fixed all conversions, called `make generate` and squashed commits.

testdata/project-v2/api/v1/admiral_types.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2023 The Kubernetes authors.
+Copyright 2024 The Kubernetes authors.
```
camilamacedo86, line 2: @lukas016 could you pleas rebase with master ?
It is merged already

camilamacedo86 (commented):
Could you please rebase with master?

It has changes that are merged already.

Thank you

testdata/project-v2/api/v1/admiral_types.go
lukas016: hi, sorry i did rebase with my current master. It should be fixed now

docs/book/src/component-config-tutorial/testdata/project/Makefile
```diff
@@ -155,30 +147,47 @@ $(LOCALBIN):
...
+GOLANGCI_LINT = $(LOCALBIN)/golangci-lint-$(GOLANGCI_LINT_VERSION)
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.2.1
 CONTROLLER_TOOLS_VERSION ?= v0.13.0
+ENVTEST_VERSION ?= release-0.16
+GOLANGCI_LINT_VERSION ?= v1.54.2
 
 .PHONY: kustomize
-kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary. If wrong version is installed, it will be removed before downloading.
```
camilamacedo86, line 167: Could we not still removing the old version if installed?

lukas016: how do you mean that? Currently if you will update version of tool, it will create new binary with different name and old is still available. You have to delete it manually or delete whole bin folder.

camilamacedo86: make sense, :- You cna ignore

camilamacedo86 (commented):
I did not test manually but for me shows fine

/lgtm

docs/book/src/component-config-tutorial/testdata/project/Makefile
```diff
@@ -155,30 +147,47 @@ $(LOCALBIN):
...
+kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
 $(KUSTOMIZE): $(LOCALBIN)
-	@if test -x $(LOCALBIN)/kustomize && ! $(LOCALBIN)/kustomize version | grep -q $(KUSTOMIZE_VERSION); then \
-		echo "$(LOCALBIN)/kustomize version is not expected $(KUSTOMIZE_VERSION). Removing it before installing."; \
-		rm -rf $(LOCALBIN)/kustomize; \
-	fi
-	test -s $(LOCALBIN)/kustomize || GOBIN=$(LOCALBIN) GO111MODULE=on go install sigs.k8s.io/kustomize/kustomize/v5@$(KUSTOMIZE_VERSION)
+	$(call go-install-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v5,$(KUSTOMIZE_VERSION))
```
camilamacedo86, line 164: @varshaprasad96 @everettraven It seems to make a lot of sense for Kubebuilder since we do not support any other language. However, I just want to come to your attention because when the bump of a new release is done in Operator SDK for Helm/Ansible plugins, it will be required. Just take this line and replace it to use the script instead since going installed locally is not a pre-requisite.

camilamacedo86 (approved):
Based in the comment https://github.com/kubernetes-sigs/kubebuilder/pull/3718#pullrequestreview-1824087250 either 

it seems ready to go, so I am approving this one.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -54,9 +58,30 @@ func (f *Makefile) SetTemplateDefaults() error {
 		f.Image = "controller:latest"
 	}
 
+	if f.ControllerRuntimeReleaseBranchName == "" {
```
camilamacedo86, line 61: @mateusoliveira43 @varshaprasad96 

This part we will need to came back to use latest
we cannot relay on the branch. 
We need to relay on the tag

````
Downloading sigs.k8s.io/controller-runtime/tools/setup-envtest@release-0.17
go: sigs.k8s.io/controller-runtime/tools/setup-envtest@release-0.17: sigs.k8s.io/controller-runtime/tools/setup-envtest@release-0.17: invalid version: unknown revision release-0.17
````

mateusoliveira43: yeah, checked and the last release is 0.17, but there is no branch to it. Probably need some change in controller-runtime to allow use of non-latest with setup-envtest

Issue #3721, opened 2024-01-08
Avoid usage of newly deprecated webhook.Validator and webhook.Defaulter interfaces


nathanperkins:
### What do you want to happen?

In [this controller-runtime issue](https://github.com/kubernetes-sigs/controller-runtime/issues/2596), it was decided to deprecate the webhook.Validator and webhook.Defaulter interfaces, due to issues they cause with project upgrades by requiring controller-runtime import in API packages. Those functions are planned to be removed later.

We should update the Kubebuilder generated code and the docs to use admission.CustomValidator and admission.CustomDefaulter instead.

### Jan 13th edit:

We also need to implement this interface outside of the API packages so that the API packages are not coupled with controller-runtime versions (see the issue above for context). The docs should be updated to warn the reader about the dangers of implementing this on the API types.

nathanperkins:
@alvaroaleman thank you for moving this! I thought I made it in kubebuilder but I must have mixed my tabs up.

camilamacedo86:
Thank you for raising this one. 
It is very helpful. 

It would be amazing if we could track issues to fix those breaking changes and desecrations in controller-runtime always. 
Thank you for moving it @alvaroaleman !!!

camilamacedo86:
**What is required to do here:**

1  - Replace `Defaulter` with  `CustomDefaulter` - https://github.com/search?q=repo%3Akubernetes-sigs%2Fkubebuilder%20%20Defaulter&type=code
2 - `Validator` with  `CustomValidator` - https://github.com/search?q=repo%3Akubernetes-sigs%2Fkubebuilder++Validator&type=code

The change must be done for the docs/examples and default scaffold for the go/v4 plugin only.
We also need to ensure that the example/tutorial works well with the new example and perform all required caveats if necessary to make it work with the interface.

jonas-jonas:
Hi, would love to provide a PR for this, if it's not yet taken. Looking at your comment @camilamacedo86 and the code base, it seems like a pretty straight forward change.

However, there haven't been any changes to the generated files since last year, meaning all generated files will be touched (2023 -> 2024). Would be happy to provide a separate PR just updating the timestamps if that better suits the workflow here. Let me know what you prefer.

camilamacedo86:
Hi @jonas-jonas,

> Would be happy to provide a separate PR just updating the timestamps

[...]

PR #3722, opened 2024-01-11
🌱: Update timestamps in generated files

jonas-jonas:
Updates the timestamps in the generated files, as discussed in https://github.com/kubernetes-sigs/kubebuilder/issues/3721#issuecomment-1886766617.

PR #3723, opened 2024-01-11
⚠️ : (go/v4): Replace usage of deprecated webhook.Validator and webhook.Defaulter interfaces

jonas-jonas:
Replaces the usage of the deprecated `webhook.Validator` and `webhook.Defaulter` interfaces with their counterparts `admission.CustomValidator` and `admission.CustomDefaulter`.

Closes https://github.com/kubernetes-sigs/kubebuilder/issues/3721


docs/book/src/component-config-tutorial/testdata/project/README.md
```diff
@@ -75,7 +75,7 @@ More information can be found via the [Kubebuilder Documentation](https://book.k
 
 ## License
 
-Copyright 2023 The Kubernetes authors.
+Copyright 2024 The Kubernetes authors.
```
camilamacedo86, line 78: @jonas-jonas 

Really thank you for that.
Could you please push a PR only with this change?
Then, I can approve it asap. 

We should have one 1 solution/issue per PR 
and only one commit either.

jonas-jonas: Already done, here: https://github.com/kubernetes-sigs/kubebuilder/pull/3722

camilamacedo86: Ops I saw: https://github.com/kubernetes-sigs/kubebuilder/pull/3722

approved/lgtm that one

camilamacedo86 (commented):
Hi @jonas-jonas,

> As per https://github.com/kubernetes-sigs/kubebuilder/issues/3721#issuecomment-1883126875, this change only updates the usages in the golang/v4 part and the docs/ files (which were generated this way). Let me know, if the other files should be changed as well.
@camilamacedo86 I updated this branch to be re-based on master (with the new timestamps). Please review :)

It will be solved by running `make generate`. Make generate will update the docs and samples under testdata. See that the command generate all and in case of the docs we inject / add the code on top after with the code implemented in: https://github.com/kubernetes-sigs/kubebuilder/blob/master/hack/docs/generate_samples.go (example: https://github.com/kubernetes-sigs/kubebuilder/blob/master/hack/docs/internal/cronjob-tutorial/webhook_implementation.go#L50-L51). You need to change the const/scripts as well. 

I noticed that your recent pull request is not passing the CI checks. Here's a breakdown of the issues and some guidance on how to address them:

1. **APIDiff / Verify API differences**: This check tends to fail when there are potential breaking changes, which sometimes are inevitable, especially with updates in the controller-runtime. While we aim to avoid such changes, they are acceptable under these circumstances, so you can disregard this failure for now.

2. **Testdata Verification**: This is failing because `make generate` hasn't been run. To resolve this, please:
    a) Rebase your branch with the master branch.
    b) Ensure that you update the scripts in hack/docs accordingly (example: https://github.com/kubernetes-sigs/kubebuilder/blob/master/hack/docs/internal/cronjob-tutorial/webhook_implementation.go#L50-L51)
    c) Execute `make generate` to update the test data.
    d) Squash your commits. Remember, we maintain a one commit per PR policy.
   
3. **E2E Tests (`pull-kubebuilder-e2e-k8s-*`)**: These are end-to-end tests failing on your PR. For debugging:
    a) Analyze the logs provided here: [E2E Test Logs](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kubebuilder/3723/pull-kubebuilder-e2e-k8s-1-28-0/1745908850650779648/build-log.txt).
    b) Run the tests locally for a more in-depth investigation. You can use Golang IDEA to run these in debug mode, which might be helpful. Here's the test suite you'll be looking at: [E2E Test Suite](https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v4/e2e_suite_test.go).
    c) It's important to understand that these test failures indicate that something in your changes might have broken existing functionality, so careful analysis and debugging are crucial here.

Please take the time to address these issues and update your pull request accordingly. If you have any questions or need further assistance, feel free to reach out!

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -59,16 +61,16 @@ The meaning of each marker can be found [here](/reference/markers/webhook.md).
...
 /*
-We use the `webhook.Defaulter` interface to set defaults to our CRD.
+We use the `admission.CustomDefaulter` interface to set defaults to our CRD.
 A webhook will automatically be served that calls this defaulting.
 
 The `Default` method is expected to mutate the receiver, setting the defaults.
 */
 
-var _ webhook.Defaulter = &CronJob{}
+var _ admission.CustomDefaulter = &CronJob{}
```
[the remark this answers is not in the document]
nathanperkins: @camilamacedo86 suggests we do this as part of a different issue, which is fine for me :)

docs/book/src/cronjob-tutorial/webhook-implementation.md
```diff
@@ -1,8 +1,8 @@
 # Implementing defaulting/validating webhooks
 
 If you want to implement [admission webhooks](../reference/admission-webhook.md)
-for your CRD, the only thing you need to do is to implement the `Defaulter`
-and (or) the `Validator` interface.
+for your CRD, the only thing you need to do is to implement the `CustomDefaulter`
+and (or) the `CustomValidator` interface.
```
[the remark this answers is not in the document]
nathanperkins: @camilamacedo86 suggests we do this as part of a different issue, which is fine for me :)

jonas-jonas:
@camilamacedo86 thanks for the detailed write-up. I started debugging the e2e failure, and will provide the fix in the coming days. I think I got an idea, on what's going on.

jonas-jonas:
I figured out why the e2e tests are failing: the `admission.CustomDefaulter/CustomValidator` interfaces work by passing in the `runtime.Object` as a parameter, instead of using the struct itself (like `webhook.Defaulter`). This, however, means that there is no type safety inside the `Default`/`ValidateX` methods anymore. To fix this, Kubebuilder would need to add a type check and cast inside the generated code to make it possible to access the fields of the struct again. 

E.g. generate the following:

```golang
func (r *CronJob) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
	cronjoblog.Info("validate create", "name", r.Name)
	res := obj.(*CronJob) // <--- THIS IS NEW

	// TODO(user): fill in your validation logic upon object creation.
	return nil, nil
} 
```

The issue with this is, that by default the generated code doesn't compile anymore, since there is an unused variable `res` inside the function. 

To fix this, I could:
1. Add a `_ = res` to make the compiler happy (pretty ugly, IMO)
2. add `res` to the log statement, that is also generated just to "use" the variable inside the method (also not great)
3. Remove the variables in all the test cases that check whether the generated code compiles and don't use the variable (pretty cumbersome)
4. Provide a Kubebuilder wrapper for the `admission.CustomDefaulter` and `admission.CustomValidator` that provides the argument with the type of the current struct

IMO 4) would be the cleanest option, as it gives implementors type safe and compilable code that does not contain (unchecked) type casts. But I am not sure if this is even feasible in the current setup and what would be needed for this.

(And I guess, one other option would be to change the interface in the controller-runtime. There is probably a solution using generics that could work here. But, this would likely be a breaking change over there.)

Let me know if you see any other options here.

camilamacedo86 (commented):
/ok-to-test

pkg/plugins/golang/v4/scaffolds/internal/templates/api/webhook.go
```diff
@@ -112,13 +116,15 @@ func (r *{{ .Resource.Kind }}) SetupWebhookWithManager(mgr ctrl.Manager) error {
 	defaultingWebhookTemplate = `
 //+kubebuilder:webhook:{{ if ne .Resource.Webhooks.WebhookVersion "v1" }}webhookVersions={{"{"}}{{ .Resource.Webhooks.WebhookVersion }}{{"}"}},{{ end }}path=/mutate-{{ .QualifiedGroupWithDash }}-{{ .Resource.Version }}-{{ lower .Resource.Kind }},mutating=true,failurePolicy=fail,sideEffects=None,groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }},verbs=create;update,versions={{ .Resource.Version }},name=m{{ lower .Resource.Kind }}.kb.io,admissionReviewVersions={{ .AdmissionReviewVersions }}
 
-var _ webhook.Defaulter = &{{ .Resource.Kind }}{}
+var _ admission.CustomDefaulter = &{{ .Resource.Kind }}{}
 
-// Default implements webhook.Defaulter so a webhook will be registered for the type
-func (r *{{ .Resource.Kind }}) Default() {
+// Default implements admission.CustomDefaulter so a webhook will be registered for the type
```
camilamacedo86, line 121: Are we calling `admission.CustomDefaulter` for default webhooks? 
Or did I misunderstood? Should not it only be used for admission ones?

camilamacedo86 (commented):
@varshaprasad96 and @nathanperkins 

Could you please, if you have the time, lend a hand with this review? I believe we need to ensure that the changes align accordingly with the types, including defaulting, converting, and so on. Thank you a lot.

camilamacedo86:
Hi @jonas-jonas 

Regards https://github.com/kubernetes-sigs/kubebuilder/pull/3723#issuecomment-1897130391

I am not sure if I could follow up but if we need to scaffold `res := obj.(*CronJob) // <--- THIS IS NEW` we can do. See that we have the Kind in `{{ .Resource.Kind }}`. 

If that does not answer the problem, can you please point out the code that is not working out and how do you need to change it in your POV so that I might help you within.

pkg/plugin/util/util.go
```diff
@@ -172,7 +172,17 @@ func ImplementWebhooks(filename string) error {
 	str, err = EnsureExistAndReplace(
 		str,
 		"// TODO(user): fill in your validation logic upon object update.",
-		`if r.Spec.Count < 0 {
+		`if oldRes.Spec.Count != 0 && newRes.Spec.Count < 0 {
```
camilamacedo86, line 175: Is this change right?

jonas-jonas: Yes, this is needed to make the compiler happy. `oldRes` is an unused variable otherwise.

camilamacedo86: What this code does is looking for `str,` and replace it with the content.
So, it does not seems right. Where `oldRes` is defined?

camilamacedo86:
HI @jonas-jonas 

> https://github.com/kubernetes-sigs/kubebuilder/pull/3723#issuecomment-1898419140

I would need to find some time to test it locally.

pkg/plugin/util/util.go
camilamacedo86: @jonas-jonas 

Could you please replace it to keep the same previous logic `if res.Spec.Count < 0 {`
So, that if has some issue it will make easier we analyse it

pkg/plugin/util/util.go
```diff
@@ -172,7 +172,17 @@ func ImplementWebhooks(filename string) error {
...
+		return nil, errors.New(".spec.count must >= 0")
+	}`)
+	if err != nil {
+		return err
+	}
+
+	str, err = EnsureExistAndReplace(
+		str,
+		"// TODO(user): fill in your validation logic upon object deletion.",
+		`if res.Spec.Count < 0 {
```
camilamacedo86, line 185: Here, we have extra lines.
Can we try to keep the change as clean as possible to make the review easier by keeping only the required changes?

pkg/plugin/util/util.go
jonas-jonas: Reverted. But please run `make generate` and see for yourself. The code can't be compiled like this.

pkg/plugins/golang/v4/scaffolds/internal/templates/api/webhook.go
```diff
@@ -85,13 +85,11 @@ package {{ .Resource.Version }}
...
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
-	{{- if .Resource.HasValidationWebhook }}
+	{{- if or .Resource.HasValidationWebhook .Resource.HasDefaultingWebhook }}
+	"context"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 	{{- end }}
-	{{- if or .Resource.HasValidationWebhook .Resource.HasDefaultingWebhook }}
-	"sigs.k8s.io/controller-runtime/pkg/webhook"
```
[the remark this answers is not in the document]
jonas-jonas: Not sure what you mean. The new interfaces are in the `"sigs.k8s.io/controller-runtime/pkg/webhook/admission"` package which is already imported.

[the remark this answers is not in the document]
camilamacedo86: But we have more than one kind/type of webhook.
See that before, we only used `"sigs.k8s.io/controller-runtime/pkg/webhook/admission"` when we had ValidationWebhook; now, it seems that you are using it for all cases.

testdata/project-v4-multigroup-with-deploy-image/api/v1/lakers_webhook.go
```diff
@@ -31,45 +32,48 @@ var lakerslog = logf.Log.WithName("lakers-resource")
...
 
 	// TODO(user): fill in your defaulting logic.
+	return nil
 }
 
 // TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
 //+kubebuilder:webhook:path=/validate-testproject-org-v1-lakers,mutating=false,failurePolicy=fail,sideEffects=None,groups=testproject.org,resources=lakers,verbs=create;update,versions=v1,name=vlakers.kb.io,admissionReviewVersions=v1
 
-var _ webhook.Validator = &Lakers{}
+var _ admission.CustomValidator = &Lakers{}
```
camilamacedo86, line 57: Should we really use Lakers{} Or should we create a new struct for the validator?

c/c @nathanperkins ^

pkg/plugin/util/testdata/exampleFile.txt
```diff
@@ -1 +1 @@
-exampleTargetexampleCodeexampleCodeexampleCodeexampleCodeexampleCodeexampleCodeexampleCode
\ No newline at end of file
+exampleTargetexampleCodeexampleCodeexampleCodeexampleCodeexampleCodeexampleCodeexampleCodeexampleCodeexampleCodeexampleCodeexampleCodeexampleCodeexampleCode
```
camilamacedo86, line 1: @jonas-jonas 

You need rebase it with the master
And then, revert the changes here
Keep it sync with make generated,

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -46,6 +46,8 @@ Then, we set up the webhook with the manager.
 func (r *CronJob) SetupWebhookWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(r).
+		WithValidator(r).
+		WithDefaulter(r).
```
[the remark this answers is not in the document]
jonas-jonas: So, before this change, with the deprecated webhook interface `For` was enough to enable both the validation and defaulting logic, if `r` implemented either (there are runtime checks [here](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/builder/webhook.go#L201-L203), if I am not mistaken). 
For custom validator and custom defaulter, you need to specifically `WithValidator` or `WithDefaulter` to enable them.

A test for just the CustomDefaulter is [here](https://github.com/kubernetes-sigs/kubebuilder/pull/3723/files#diff-da5765d58e09cb0d9da8b09d04dde3e3762263e553a8b40e65babc36062af21fR35) and for CustomValidator [here](https://github.com/kubernetes-sigs/kubebuilder/pull/3723/files#diff-d253c7d3da23ab04bf8b977cba0892853e326975afc2938086c8fb9e8130d62bR35).

pkg/plugins/golang/v4/scaffolds/internal/templates/api/webhook.go
jonas-jonas: `webhook.CustomDefaulter` is an alias for `admission.CustomDefaulter`. I missed that when I first looked at the required changes, but have now reverted it to use `webhook.CustomDefaulter`. I hope this answers your question.

pkg/plugin/util/testdata/exampleFile.txt
jonas-jonas: Done.

pkg/plugin/util/util.go
jonas-jonas: Reverted.

jonas-jonas:
@camilamacedo86 sorry for the delays on my side. I've revisited all open conversations and pushed fixes for the remaining issues.

The e2e tests are failing, because the new `CustomDefaulter` and `CustomValidator` interfaces no longer provide a type-safe way for accessing the resources:

New signature:

```golang
func (r *CronJob) Default(ctx context.Context, obj runtime.Object) error {
```
([where `obj` is a deep copy of the `r` instance with updated values set.](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/webhook/admission/defaulter_custom.go#L74-L80))

vs. 

```golang
func (r *CronJob) Default() {
```
([where `r` can be used to access everything that's needed.](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/webhook/admission/defaulter.go#L70-L77))

Please advise on how to move forward with this.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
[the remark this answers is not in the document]
camilamacedo86: But seems that we are saying here.
The implementation is For the struct `r`
and with the validator using the same struc.
Should we not have a new type, struct for the validation?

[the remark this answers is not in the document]
camilamacedo86: See: https://github.com/kubernetes-sigs/controller-runtime/blob/4000e996a202917ad7d40f02ed8a2079a9ce25e9/pkg/builder/webhook.go#L70-L74

I understand that you pass an struct representing the Validator and not the object that will be validated.

camilamacedo86:
Hi @jonas-jonas 

Regards : https://github.com/kubernetes-sigs/kubebuilder/pull/3723#issuecomment-1925685343
By looking at the examples in the controller-runtime it seems for me that we should have

> func (r *CronJobCustomValidator) Default(ctx context.Context, obj runtime.Object) error {

See: https://github.com/kubernetes-sigs/controller-runtime/blob/4000e996a202917ad7d40f02ed8a2079a9ce25e9/pkg/builder/webhook_test.go#L869-L930

Also, we can log the obj, and we should add some implementations like the one we have in the examples. 
We can scaffold ALL that would be required for any case scenario in the default one. 

To do this task, it is required to scaffold a project with Kubebuilder, create an API and a webhook, and then try to do a basic implementation and see if that works. So that we can ensure the right way to do so. Did you make that? 
Furthermore, It would be great to ensure we also cover it with e2e tests.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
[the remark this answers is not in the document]
camilamacedo86: We need the validator struct 
We should not be passing :

```
WithValidator(r).
WithDefaulter(r).
```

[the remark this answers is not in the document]
sbueringer: Agree with Camilla. CustomValidator and CustomDefaulter should be implemented on a separate struct.

Otherwise there would have been no point to introducing these if we still implement them on the API type.

(the goal is to have CustomValidator/CustomDefaulter outside of the api package to reduce the dependencies to controller-runtime in the API package)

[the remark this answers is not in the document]
camilamacedo86: Thank you to confirm my understanding about the requirements for the changes

camilamacedo86:
Closing in favor https://github.com/kubernetes-sigs/kubebuilder/pull/4060/files

I hope that you do not mind

Issue #3740, opened 2024-01-25
Support for IBM Z (s390x architecture)


ShivangGoswami:
### What do you want to happen?

kubebuilder project is available for IBM power(ppc64le) but not on IBM Z(s390x).operators are available on both platform and there is a need for kubebuilder project on s390x as well

camilamacedo86:
All latest is available for it now, and in the next release we should have the Kubebuilder CLI itself. (https://github.com/kubernetes-sigs/kubebuilder/issues/2298#issuecomment-1912461019)

So, we can close this one. 

Thank you for raising it and for your help.

Issue #2298, opened 2021-08-12
Adding support for s390x architecture


rposts:
### What do you want to happen?

Currently `Kubebuilder` releases binaries for several platforms including `arm64` and `ppc64le`.

I am exploring the possibility of making binary available on `s390x` big-endian architecture.

I am able to successfully build and run `test-unit` on `s390x` architecture with following changes to `Makefile`:
```diff
index 2d69e6f4..23575c1a 100644
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,11 @@ test: test-unit test-integration test-testdata test-book ## Run the unit and int

 .PHONY: test-unit
 test-unit: ## Run the unit tests
+ifeq (s390x,$(shell uname -p))
+       go test -v ./pkg/...
+else
        go test -race -v ./pkg/...
+endif

 .PHONY: test-coverage
 test-coverage: ## Run unit tests creating the output to report coverage
```
However, `make test` fails because it downloads incompatible [`kubebuilder-tools`](https://storage.googleapis.com/kubebuilder-tools) from `googleapis` storage location.

I suspect in order to add `s390x` support we will require the following two steps at a minimum:
- Make available `s390x` compatible `kubebuilder-tools` on https://storage.googleapis.com/kubebuilder-tools
- Modify `Makefile` to remove `race` option for `s390x` architecture

I am looking for direction on how to make this possible so that `s390x` arch `Kubebuilder` can be obtained directly from the [`Releases` ](https://github.com/kubernetes-sigs/kubebuilder/releases) page.

Thanks.

rashmigottipati:
assign @camilamacedo86

camilamacedo86:
/remove-lifecycle stale

I think we need to revisit this one.

tdaleibm:
@camilamacedo86 I'm not sure I fully understand the process but I see you are doing a similiar for arm64 in bug #2664 , are you able to trigger on the Google Cloud to generate the artefact for s390x ? I was able to workaround `make test` by sideloading with setup-envtest, All three binaries (kubectl, kube-apiserver, etcd)  already have s390x builds available. What more is required to get this on https://storage.googleapis.com/kubebuilder-tools?

tdaleibm:
Sorry, I meant the individual packages usually contained in kubebuildre-tools ( so kubectl, kube-apiserver, and etcd ). I'm not sure how to packaging of those into https://storage.googleapis.com/kubebuilder-tools is being done for other architectures.

[...]

PR #3741, opened 2024-01-25
✨ Start to build CLI binary for linux_s390x as target

ShivangGoswami:
Issue: https://github.com/kubernetes-sigs/kubebuilder/issues/3740

this PR adds support for s390x through go release so that the required binaries are generated when these assets are generated via GitHub-actions

### DIFF
--- a/build/.goreleaser.yml
+++ b/build/.goreleaser.yml
@@ -41,6 +41,7 @@ builds:
       - linux_amd64
       - linux_arm64
       - linux_ppc64le
+      - linux_s390x
       - darwin_amd64
       - darwin_arm64
     env:


build/.goreleaser.yml
```diff
@@ -41,6 +41,7 @@ builds:
       - linux_amd64
       - linux_arm64
       - linux_ppc64le
+      - linux_s390x
```
camilamacedo86, line 44: We cannot support the release of the tool for `linux_s390x` if our envtest binaries does not run in this ENV.
See that we build them here: https://github.com/kubernetes-sigs/kubebuilder/tree/tools-releases and we would need to setup a GCP Build target to begin to support this one.

Also, will ALL images build and ship work within ? We either need to check it?

camilamacedo86:
Hi @ShivangGoswami

This PR might generate the CLI bin for this arch. 
However, we need to generate the invest binaries for it.
More info: https://github.com/kubernetes-sigs/kubebuilder/issues/2298#issuecomment-1242994510

I will try to give it a look at it prior to the next release.
Thank you for raising this one.

camilamacedo86 (commented):
/ok-to-test

camilamacedo86 (approved):
Since we are building the auth-proxy image and, in the past, it was decided in the Community Meetings that we could start to make the binaries for it, why not move forward with this one?

However, we still need to build the ENV TEST binaries to allow the tool to work correctly in this environment.

PR #3742, opened 2024-01-25
:bug: Align CRD version validation with apiextensions

### LINKED ISSUE #3739: versionPattern should be more flexible to allow internal versions

The [versionPattern](https://github.com/kubernetes-sigs/kubebuilder/blob/165af70e87848677a039cd0cd87cc67c545d983c/pkg/model/resource/gvk.go#L28) enforced by kubebuilder requires all CRD versions to start with a `v`, however this makes usage of an internal version confusing (e.g. creating a `v0`) where an `internal` or `__internal` version would communicate intent more clearly; `__internal` is even used within the [core Kubernetes codebase](https://github.com/kubernetes/apimachinery/blob/f14778da5523847e4c07346e3161a4b4f6c9186e/pkg/runtime/interfaces.go#L30).

What I'd like to see is the `versionPattern` relaxed to make an allowance for internal versions, something like so:

Before: 

```go
const (
	versionPattern = "^v\\d+(?:alpha\\d+|beta\\d+)?$"
)
```

After:

```go
const (
	versionPattern = "^(?:__internal|v\\d+(?:alpha\\d+|beta\\d+)?)$"
)

```

Doing so would be additive and not break any of the existing uses.

### Reproducing this issue

_No response_

### KubeBuilder (CLI) Version

v3.12.0

### PROJECT version

_No response_

### Plugin versions

_No response_

### Other versions

_No response_

### Extra Labels

_No response_

robzienert:
Fixes #3739 where the existing CRD version name validation is more restrictive than Kubernetes itself. This aligns the validation with [apiextensions CRD version name validation](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go#L388).


pkg/model/resource/gvk.go
```diff
@@ -60,8 +53,9 @@ func (gvk GVK) Validate() error {
 	if gvk.Version == "" {
 		return fmt.Errorf(versionRequired)
 	}
-	if !versionRegex.MatchString(gvk.Version) {
-		return fmt.Errorf("Version must match %s (was %s)", versionPattern, gvk.Version)
+	if err := validation.IsDNS1035Label(gvk.Version); err != nil {
```
camilamacedo86, line 56: Could you confirm if that is the right validation done by k8s?
@erikgb @varshaprasad96, could you lend a hand with this review?
That would be appreciate a lot

REALLY THANK YOU

robzienert:
Was that `/retest` comment supposed to comment anything afterwards? The failed tests all look transient.

camilamacedo86:
Hi @robzienert 

Really thank you for your contribution 🥇 

For us to get this merged, we need to validate/ensure (add here a link for the source code, i.e.) that it is the right validation for versions. I will need to take some time to check, but if you are able to do this check and just add in the description the info, then I am OK to approve this one.

robzienert:
Is the link in the initial description not what you're looking for? Happy to dig around if not, but want to verify.

camilamacedo86 (approved):
/lgtm
/approved

PR #3749, opened 2024-01-27
:sparkles: (go/v4): upgrade to Support k8s 1.29

camilamacedo86:
**Description**
- Upgrade from k8s 1.28 to 1.29
- Upgrade controller-runtime from 0.16.13 to 0.17.0
- Upgrade controller-tools from 0.13.0 to 0.14.0


Kavinjsir (approved):
Looks good to me 👍🏼 

See that in the release of controller-runtime v0.17.0, [admission.Validator and admission.Defaulter are deprecated](https://github.com/kubernetes-sigs/controller-runtime/pull/2639).

I think we may need to update the scaffolding of enabling webhook, as new interfaces and corresponding built-in functions are provided by controller-runtime, see:
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.17.0/examples/builtins/main.go#L77

camilamacedo86:
I am moving forward with this one.
See that the API diff will fail since we are bumping controller-runtime.

Issue #3748, opened 2024-01-27
`exampleFile.txt` get expanded for every unit-test


Kavinjsir:
### What broke? What's expected?

Currently running `make test-unit` will result in the change of `pkg/plugin/util/testdata/exampleFile.txt`.

It is an unexpected change because it results in an ever-growing file.
<img width="958" alt="Screenshot 2024-01-26 at 11 46 24 PM" src="https://github.com/kubernetes-sigs/kubebuilder/assets/18136486/6c1f0c18-9663-4dda-8ea3-8a436460c6d4">

### Reproducing this issue

1. Clone the repo, go to the root dir of the repo.
2. Run `make test-unit`
3. Run `git diff`

### KubeBuilder (CLI) Version

Version: main.version{KubeBuilderVersion:"v3.11.0-67-g500d475b", KubernetesVendor:"unknown", GitCommit:"500d475b460b79eb4a98af10fd934399fbe42ae5", BuildDate:"2023-08-26T05:11:35Z", GoOs:"darwin", GoArch:"arm64"}

### Extra Labels

/kind cleanup

Kavinjsir:
Looks like performing the util unit test case modifies the exampleFile.txt, see:
https://github.com/kubernetes-sigs/kubebuilder/blob/v3.13.0/pkg/plugin/util/util_test.go#L32-L37

One way may be:
Add an [AfterAll](https://pkg.go.dev/github.com/onsi/ginkgo/v2#AfterAll) closure to roll back any updates during the test suite regardless of pass or failure.

PR #3757, opened 2024-01-29
🌱 drop unnecessary file change for unit test

Eileen-Yu:
### Motivation:
Resolve https://github.com/kubernetes-sigs/kubebuilder/issues/3748

### Description:
Drop the latest append `exampleCode` which is only for unit test, so as to avoid checkout that file each time.


camilamacedo86 (approved):
/lgtm

PR #3726, opened 2024-01-15
📖 Small fixes for getting started overview guide

nemethloci:
Fixes #3725

Small fixes for getting started guide described in the ticket

PR #3758, opened 2024-01-29
📖 Clean up getting started docs

stmcginnis:
While reading through the Getting Started page, I noticed there were several places where things have either changed since it was originally written or incorrect names or references were copy/pasted from other docs. This updates the Getting Started page to accurately reflect what is generated when following the instructions.

There were also several typos and grammatical errors noticed while reading through the doc. Where possible, this was corrected in the Getting Started doc as well as anywhere else in the documentation where this same information was copied.


pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/manager/controller_manager_config.go
```diff
@@ -64,14 +64,14 @@ webhook:
...
 # Manager is stopped, otherwise, this setting is unsafe. Setting this significantly
-# speeds up voluntary leader transitions as the new leader don't have to wait
-# LeaseDuration time first.
+# speeds up voluntary leader transitions as the new leader doesn't have to wait
+# the LeaseDuration time first.
 # In the default scaffold provided, the program ends immediately after
-# the manager stops, so would be fine to enable this option. However,
-# if you are doing or is intended to do any operation such as perform cleanups
+# the manager stops, so it would be fine to enable this option. However,
+# if you are doing, or are intending to do, any operation such as perform cleanups
```
camilamacedo86, line 74: kustomize/v1 is deprecated and is no longer used unless you scaffold a project using go/v3 explicity
It will be removed asap after the next release
So, can you please revert those changes.

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/manager/controller_manager_config.go
```diff
@@ -64,14 +64,14 @@ webhook:
...
+  # leaderElectionReleaseOnCancel defines if the leader should step down volume
+  # when the Manager ends. This requires the binary to immediately end when the
+  # Manager is stopped, otherwise, this setting is unsafe. Setting this significantly
+  # speeds up voluntary leader transitions as the new leader doesn't have to wait
+  # the LeaseDuration time first.
+  # In the default scaffold provided, the program ends immediately after
+  # the manager stops, so it would be fine to enable this option. However,
+  # if you are doing, or are intending to do, any operation such as perform cleanups
+  # after the manager stops then its usage might be unsafe.
+  # leaderElectionReleaseOnCancel: true
```
camilamacedo86, line 76: My 2 cents is that the added space here might cause problems for other places which are replacing this content.
Can you please let us know why you are adding the space?

docs/book/src/getting-started.md
```diff
@@ -206,9 +206,9 @@ return ctrl.Result{RequeueAfter: nextRun.Sub(r.Now())}, nil
 
 #### In the context of our example
 
-When a Custom Resource is applied to the cluster, there's a designated controller to manage the Memcached Kind. You can check its reconciliation implemented:
+When a Custom Resource is applied to the cluster, there's a designated controller to manage the Memcached Kind. You can how its reconciliation is implemented:
 
-From `testdata/project-v4-with-deploy-image/internal/controller/memcached_controller.go`:
+From `internal/controller/memcached_controller.go`:
```
camilamacedo86, line 211: ^ it is the testdata dir of the project where we add the examples

docs/book/src/getting-started.md
```diff
@@ -33,7 +33,7 @@ kubebuilder init --domain=example.com
 Next, we'll create a new API responsible for deploying and managing our Memcached solution. In this instance, we will utilize the [Deploy Image Plugin][deploy-image] to get a comprehensive code implementation for our solution.
 
 ```
-kubebuilder create api --group example.com --version v1alpha1 --kind Memcached --image=memcached:1.4.36-alpine --image-container-command="memcached,-m=64,-o,modern,-v" --image-container-port="11211" --run-as-user="1001" --plugins="deploy-image/v1-alpha" --make=false
+kubebuilder create api --group cache --version v1alpha1 --kind Memcached --image=memcached:1.4.36-alpine --image-container-command="memcached,-m=64,-o,modern,-v" --image-container-port="11211" --run-as-user="1001" --plugins="deploy-image/v1-alpha" --make=false
```
camilamacedo86, line 36: I am ok with change it is we change the example/code generate under the samples testdata, see: https://github.com/kubernetes-sigs/kubebuilder/pull/3726/files#r1452333362

camilamacedo86, line 36: The example under test data is not built with the group cache. Is it?

docs/book/src/getting-started.md
stmcginnis: Yes, but as mentioned above, that content is different than these instructions result in. The locally generated content from following the commands above are under the updated path. If we want to refer to the git repo for additional examples I think that could work. But as this part is, it can be confusing referring to a path that is not part of the local content.

pkg/plugins/common/kustomize/v1/scaffolds/internal/templates/config/manager/controller_manager_config.go
stmcginnis: Sure thing!

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/manager/controller_manager_config.go
stmcginnis: My motivation was to have this properly aligned so the YAML is valid, even though it is all commented out. Basically, to have `leaderElectionReleaseOnCancel` be under the `leaderElection` section, rather than appearing to be a peer.

I did not find issues with the changes format, but if there are concerns that the risk could outweigh the benefit, I can shift that back.

docs/book/src/getting-started.md
camilamacedo86: I see; we should actually generate the sample under the docs ( with the same instructions ) as we do for the tutorials and just use the code directly from the sample.

So, that when we run `make generate` we will be able to keep this one update. 

See: 
- Here we generate the sample: https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/cronjob-tutorial/testdata/project
- Then, you can check that in the doc we are ref, the code directly like: `{{#literatego ./testdata/emptycontroller.go}}` in https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/controller-overview.md
- You can either check the make generate target and see that we build the samples using https://github.com/kubernetes-sigs/kubebuilder/tree/master/hack/docs

Otherwise, it is hard to keep it maintained.

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/manager/controller_manager_config.go
camilamacedo86: I think it is OK we move forward with this one because it is a comment
The problem with changes like that is more problematic when they are in the optional things since other tools might be looking for the text and replacing it.

camilamacedo86:
Hi @stmcginnis,

Really, thank you for all your collaboration 🥇 
Just a few nits.
- In this PR we are addressing more than one thing. For example, the changes in the customize/v2 have no relation with getting started docs. Could you please create a PR for this only?
- Regarding the required fixes in the Getting Started doc, I think we need to properly sort it out by having the sample under docs so that it will be properly updated. See: https://github.com/kubernetes-sigs/kubebuilder/pull/3758#discussion_r1470366123 if you be able to do that it would be great

docs/book/src/getting-started.md
camilamacedo86: I am OK with the change for it for cache.

camilamacedo86: I see. I am OK with the change.
Thank you for looking at this one !!!
Your help is very appreciated.

pkg/plugins/golang/v3/scaffolds/internal/templates/main.go
```diff
@@ -272,12 +272,12 @@ func main() {
...
-		// speeds up voluntary leader transitions as the new leader don't have to wait
-		// LeaseDuration time first.
+		// speeds up voluntary leader transitions as the new leader doesn't have to wait
+		// the LeaseDuration time first.
 		//
 		// In the default scaffold provided, the program ends immediately after 
-		// the manager stops, so would be fine to enable this option. However, 
-		// if you are doing or is intended to do any operation such as perform cleanups 
+		// the manager stops, so it would be fine to enable this option. However,
+		// if you are doing, or are intending to do, any operation such as perform cleanups
```
camilamacedo86, line 280: This plugin is also deprecated so we should not perform changes on it unless we need to address a critical CVE

docs/book/src/getting-started.md
varshaprasad96: +1 to as @camilamacedo86 said. If we could directly refer the paths in the doc, that would make it easier for us to maintain it. Huge + for the idea though. We would have to probably discuss more on the how the testate would be structured to refer multi-version/multi-group scenarios for the same plugins as we have today.

stmcginnis:
Thanks for the feedback - I will get an updated PR shortly.

One thing about the reference path though. There seems to be a disconnect in what this doc is doing, so I think we need to clarify that. As it is, the getting started guide is a tutorial/guide walking the reader through running commands and working with the generated code based on that. So I think it would be confusing to mix this "real world" walkthrough of generating the code with linking off to a static reference in the repo of a similar-but-not-quite-the-same example of what they have locally.

In other words, I think we need to keep the tutorial separate from and test data reference we could point them at.

We may want to think about changing what this guide is if we just want to refer to examples in the repo. As it is now, there are other issues with the getting started doc. The main one being that the beginning of the doc states what the user will go through and end up with at the end, but then the content of the doc only goes a short way towards those stated goals and stops. So there are definitely bigger issues that need to be decided here.

I see two main choices at this point:

1. Finish implementing the getting started guide to actually walk the user through all the declared goals, keeping with the locally generated content that the instructions generates for them; or
2. Change the getting started guide to be a description of what they can find as static content already in the repo, only explaining what local commands they could have run to get similar results on their own.

As it is now, it's a bit incomplete and disjointed.

camilamacedo86:
Hi @stmcginnis,

> There seems to be a disconnect in what this doc is doing, so I think we need to clarify that

That is all point. The tutorial describes how to create the project using the Deploy Image plugin
The examples should be exactly the code that is generated in the project. 
Because of that IHMO, the right way to sort it out is by doing as we do in all other places: https://github.com/kubernetes-sigs/kubebuilder/pull/3758#discussion_r1470366123

docs/book/src/getting-started.md
stmcginnis: I looked in to using `literatego` and making this more integrated, but there were some issues with that approach. The biggest challenge is that pulls in the entire source file, when in most cases here we only want to highlight a small portion of it.

I hope I've reached a happy medium. I do now generate `testdata` for the getting started docs, and I try to directly pull in relevant content where possible. For the other instances, I just updated the snippet of code the docs were highlighting, but also added a link to the file in the repo so the user has a canonical reference for anything being covered. Then if they do diverge at some point, it's at least possible to follow the link to see the current full example that should match what they get locally by running the commands.

Let me know if there are any better ideas or other ways to improve that.

stmcginnis:
Not sure about the verification test failure. Passing locally for me, so I'm not sure what is different in my environment. Will try to investigate further.

camilamacedo86:
Hi @stmcginnis,

GREAT WORK !!!

 > Not sure about the verification test failure. Passing locally for me, so I'm not sure what is different in my environment. Will try to investigate further.
 
It is because it is not rebased with the master branch
Just rebase your branch with master, re-run, make generate and then squash the commits, for we have just 1, and it is great to fly (we can get it merged. It will pass in all tests) !!!!

camilamacedo86 (approved):
/lgtm

/approved

camilamacedo86:
Hi @stmcginnis 

It is a very nice one.

PR #3754, opened 2024-01-27
:sparkles: (go/v4): Upgrade CertManager and Prometheus versions used in the e2e tests

camilamacedo86:
**Description**

Upgrade CertManager and Prometheus versions used in the e2e tests

PR #3759, opened 2024-01-30
:bug: Revert ":sparkles: (go/v4): Upgrade CertManager and Prometheus versions used in the e2e tests"

camilamacedo86:
Reverts kubernetes-sigs/kubebuilder#3754

We have a test (project-v4-sample) which should be testing it.
However, the test passed, and we could check that the Prometheus Operator installer has an issue and cannot be applied directly. However, it seems that from k8s 1.29, we no longer see the error related to the CRD being too long. 

So the project-v4-sample is passing but it will not work in k8s versions which are lower the 1.29


camilamacedo86:
I am reverting this change so that we can work in the release.

PR #3766, opened 2024-01-31
:bug: (go/v4) Fix makefile target build-installer to discarding the previous content

lukas016:
**Description**:
current implementation of makefile target build-installer cannot work properly without crd. Root cause of problem is missing discard of content, if crd isn't used.
Result of problem is appending new content behind previous into dist/install.yaml and file content is growing and contains multiple versions.

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -192,10 +192,11 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 .PHONY: build-installer
 build-installer: manifests generate kustomize ## Generate a consolidated YAML with CRDs and deployment.
 	mkdir -p dist
+	echo "---" > dist/install.yaml # Clean previous content
 	@if [ -d "config/crd" ]; then \
 		$(KUSTOMIZE) build config/crd > dist/install.yaml; \
+		echo "---" >> dist/install.yaml; \
 	fi
-	echo "---" >> dist/install.yaml  # Add a document separator before appending
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default >> dist/install.yaml
 
--- a/docs/book/src/component-config-tutorial/testdata/project/Makefile
+++ b/docs/book/src/component-config-tutorial/testdata/project/Makefile
@@ -118,10 +118,11 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 .PHONY: build-installer
 build-installer: manifests generate kustomize ## Generate a consolidated YAML with CRDs and deployment.
 	mkdir -p dist
+	echo "---" > dist/install.yaml # Clean previous content
 	@if [ -d "config/crd" ]; then \
 		$(KUSTOMIZE) build config/crd > dist/install.yaml; \
+		echo "---" >> dist/install.yaml; \
 	fi
-	echo "---" >> dist/install.yaml  # Add a document separator before appending
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default >> dist/install.yaml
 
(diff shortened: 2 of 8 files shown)


pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -192,10 +192,11 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 .PHONY: build-installer
 build-installer: manifests generate kustomize ## Generate a consolidated YAML with CRDs and deployment.
 	mkdir -p dist
+	echo "---" > dist/install.yaml # Clean previous content
 	@if [ -d "config/crd" ]; then \
 		$(KUSTOMIZE) build config/crd > dist/install.yaml; \
+		echo "---" >> dist/install.yaml  # Add a document separator before appending
 	fi
-	echo "---" >> dist/install.yaml  # Add a document separator before appending
```
camilamacedo86, line 198: @lukas016 

/ok-to-test

PS.: if it changes the final result, the CI will fail because it is missing generating the test data `make generate`. It is also tested via e2e tests, so passing on that is all fine.

camilamacedo86: It is either failing in teh ew tests

```
     mkdir -p dist
      echo "---" > dist/install.yaml # Clean previous content
      bash: -c: line 4: syntax error: unexpected end of file
      make: *** [Makefile:122: build-installer] Error 2
      
      {
          s: "make build-installer IMG=e2e-test/controller-manager:nfca failed with error: (exit status 2) /home/prow/go/src/sigs.k8s.io/kubebuilder/test/e2e/v4/e2e-nfca/bin/controller-gen-v0.14.0 rbac:roleName=manager-role crd webhook paths=\"./...\" output:crd:artifacts:config=config/crd/bases\n/home/prow/go/src/sigs.k8s.io/kubebuilder/test/e2e/v4/e2e-nfca/bin/controller-gen-v0.14.0 object:headerFile=\"hack/boilerplate.go.txt\" paths=\"./...\"\nDownloading sigs.k8s.io/kustomize/kustomize/v5@v5.3.0\nmkdir -p dist\necho \"---\" > dist/install.yaml # Clean previous content\nbash: -c: line 4: syntax error: unexpected end of file\nmake: *** [Makefile:122: build-installer] Error 2\n",
```

Can you please let me know the scenario that is not working for you?
Can you describe the steps to create a POC and reproduce the issue?
It would be great if we ensure that we also test it out in the tests.

lukas016: Hi, i am working on fix.

Problematic scenario is:
1. Remove config/crd
2. Call make build-installer 2 or more times
3. Check dist/install.yaml

camilamacedo86: Note that when you scaffold any API the tool in uncomment  `config/default/kustomization.yaml`
You probably would solve it by commenting again on the

```
Resources:
- ../crd
```

In `config/default/kustomization.yaml`

So, that is not a really issue because
You break the scaffold if you manually remove the config/crd directory.
Are you able to face any issues without deleting things scaffold manually?

Nevertheless, if you want to change it to address it, it is OK, assuming it is still working as it should for the other scenarios.

lukas016: do we need if statement in makefile in this case? We can completely remove if statement and we won't need this fix.

lukas016: i think this is better or why crd was generated independent?

```
build-installer: manifests generate kustomize ## Generate a consolidated YAML with CRDs and deployment.
	mkdir -p dist
	$(KUSTOMIZE) build config/default > dist/install.yaml
```

camilamacedo86: The above will not work:

- a) Because you can create a project without CRD ( the problem is when you create the API the scaffold change, so if you delete the config/crd manually you also need to fix the scaffold manually accordingly )
- b) My 2 cents is that will fail because will be missing `echo "---" >> dist/install.yaml  # Add a document separator before appending`
-

camilamacedo86 (approved):
/lgtm
/approved

@lukas016 

It is passing now, So all fine.
If you see any scenario that does not work please feel free to propose changes.

Issue #3784, opened 2024-02-15
Can't create an API of Kind with name DNS but creating an API of Kind with name ENS works fine


rajha-korithrien:
### What broke? What's expected?

I tried to create an API with a Kind name of `DNS` and it fails with:

```
../kubebuilder create api --group cache --version v1alpha1 --kind DNS --resource --controller
Error: failed to create API: unable to inject the resource to "base.go.kubebuilder.io/v4": invalid Plural: []string{"a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or '123-abc', regex used for validation is '[a-z](?:[-a-z0-9]*[a-z0-9])?')"}
```

I expected it to work the same as when I create an API with a Kind name of `ENS` which works correctly:
```
../kubebuilder create api --group cache --version v1alpha1 --kind ENS --resource --controller
INFO Writing kustomize manifests for you to edit...
INFO Writing scaffold for you to edit...
INFO api/v1alpha1/ens_types.go
INFO api/v1alpha1/groupversion_info.go
INFO internal/controller/suite_test.go
INFO internal/controller/ens_controller.go
INFO internal/controller/ens_controller_test.go
INFO Update dependencies:
$ go mod tidy
INFO Running make:
$ make generate
mkdir -p /Users/rajha/Projects/ICASA/PEACE/SDN/K8s/kubebuilder-test/guestbook/bin
Downloading sigs.k8s.io/controller-tools/cmd/controller-gen@v0.14.0
go: downloading sigs.k8s.io/controller-tools v0.14.0
go: downloading github.com/fatih/color v1.16.0
go: downloading github.com/spf13/cobra v1.8.0
go: downloading golang.org/x/mod v0.14.0
go: downloading github.com/go-logr/logr v1.3.0
go: downloading github.com/mattn/go-isatty v0.0.20
go: downloading golang.org/x/sys v0.15.0
/Users/test/K8s/kubebuilder-test/guestbook/bin/controller-gen-v0.14.0 object:headerFile="hack/boilerplate.go.txt" paths="./..."
Next: implement your new API and generate the manifests (e.g. CRDs,CRs) with:
$ make manifests
```

### Reproducing this issue

Follow the "getting started" instructions: https://book.kubebuilder.io/quick-start.html

Get kubebuilder and setup a new project directory to work in:
```
[...]

PR #3787, opened 2024-02-16
🐛 Revert ":bug: Align CRD version validation with apiextensions"

camilamacedo86:
Reverts kubernetes-sigs/kubebuilder#3742

Closes:  https://github.com/kubernetes-sigs/kubebuilder/issues/3784


camilamacedo86:
We need revert this one. 

Using IsDNS1035Label to validate a Kubernetes CRD Kind name is not accurate because this validation is intended for DNS labels, which have stricter requirements, such as needing to start with a lowercase letter and consisting only of lowercase alphanumeric characters and '-', which does not align with the conventions for Kind names in Kubernetes. Kind names typically follow CamelCase and can start with uppercase letters, which would not be permitted by IsDNS1035Label. Therefore, it would not allow names like "DNS" for a Kind due to these constraints.

camilamacedo86 (commented):
/lgtm


Regression

PR #3799, opened 2024-02-27
📖 Add the `Admission webhooks` link in `docs/book/src/reference/admisson-webhook.md`

renormalize:
The link to the Kubernetes docs for Admission webhooks was not present in the reference docs.
Adding for the sake of correctness.

### DIFF
--- a/docs/book/src/reference/admission-webhook.md
+++ b/docs/book/src/reference/admission-webhook.md
@@ -1,6 +1,6 @@
 # Admission Webhooks
 
-[Admission webhooks][k8s-doc-admission-webhooks] are HTTP callbacks that receive admission requests, process
+[Admission webhooks](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#what-are-admission-webhooks) are HTTP callbacks that receive admission requests, process
 them and return admission responses.
 
 Kubernetes provides the following types of admission webhooks:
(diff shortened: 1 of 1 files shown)


docs/book/src/reference/admission-webhook.md
```diff
@@ -99,5 +99,3 @@ spec:
 While certain edge scenarios might allow a mutating webhook to seamlessly modify the status, treading this path isn't a 
 universally acclaimed or recommended strategy. Entrusting the controller logic with status updates remains the 
 most advocated approach.
-
-[k8s-doc-admission-webhooks]: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#what-are-admission-webhooks
```
camilamacedo86, line 103: We should use an alias as it is now so that it makes it easier us keep things maintained (allows re-usage, easily identify where we need to update etc)

So, can you please share with us what is the motivation for this PR?
is the link wrong?

renormalize: The rendered page on the website (with the alias) does not show up as an embedded hyperlink to the docs, and shows up as `[Admission webhooks][k8s-doc-admission-webhooks]`. The link just shows up as plain text at the end of the rendered page (what I take to be the alias)

Since the rendered page did not link to the page on hover/click, I took it to be an error in the rendering, or a mistake.

If this is expected behaviour, I'm happy to close the PR, and apologize for the unnecessary PR!

camilamacedo86 (approved):
/lgtm
/approved

/ok-to-test

PR #3810, opened 2024-03-06
🌱: Add prefix to bot PR title to respect community rule

Kavinjsir:
#### Description

Add a condition to skip PR Verifier if the PR is created by dependabot

#### Motivation

Current PRs created by dependabot fail the CI/CD workflows as the title of the PR does not match the community's criteria.
It seems dependabot does NOT provide an entry to configure the title, so I am thinking if we can just skip the PR verifier in such scenarios.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,6 +22,8 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    commit-message:
+      prefix: ":seedling:"
     labels:
       - "ok-to-test"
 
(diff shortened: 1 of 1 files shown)


mateusoliveira43 (changes requested):
Can not be this achieved by using this option https://github.com/migtools/oadp-non-admin/blob/master/.github/dependabot.yml#L10 ? i kubebuilder case, setting it to 🌱

Kavinjsir:
> Can not be this achieved by using this option https://github.com/migtools/oadp-non-admin/blob/master/.github/dependabot.yml#L10 ? in kubebuilder case, setting it to 🌱

Great suggestion! Just applied that.

mateusoliveira43 (approved):
would just change PR title
LGTM

camilamacedo86 (approved):
That is great 🥇 

Thank you a lot

/approved
/lgtm
/ok-to-test

PR #3816, opened 2024-03-08
🐛 makefile: remove invalid comment 

mateusoliveira43:
## Why the changes were made

Fix command documentation.

## How the changes were made

When `make lint` command was added (https://github.com/kubernetes-sigs/kubebuilder/commit/46ea676fc75e7a8844bbf237cf8e1c4b66b24dd4#diff-e5d6d4f9d67174f1f3764674eb786688353c77cef8514e24d1323e55e5a4a7e8) I believe its documentation was copied from project Makefile (https://github.com/kubernetes-sigs/kubebuilder/commit/feb0c5463bb51d07dc9027a139a0c0a0cfc2253f#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52) which does run `yamllint`, but not command generated by kubebuilder. So, removed that part from command doc.

After applying changes, ran `make generate`. 

Confirm the documentation is correct now.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/Makefile
+++ b/docs/book/src/cronjob-tutorial/testdata/project/Makefile
@@ -69,7 +69,7 @@ test-e2e:
 	go test ./test/e2e/ -v -ginkgo.v
 
 .PHONY: lint
-lint: golangci-lint ## Run golangci-lint linter & yamllint
+lint: golangci-lint ## Run golangci-lint linter
 	$(GOLANGCI_LINT) run
 
 .PHONY: lint-fix
--- a/docs/book/src/getting-started/testdata/project/Makefile
+++ b/docs/book/src/getting-started/testdata/project/Makefile
@@ -69,7 +69,7 @@ test-e2e:
 	go test ./test/e2e/ -v -ginkgo.v
 
 .PHONY: lint
-lint: golangci-lint ## Run golangci-lint linter & yamllint
+lint: golangci-lint ## Run golangci-lint linter
 	$(GOLANGCI_LINT) run
 
 .PHONY: lint-fix
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -146,7 +146,7 @@ test-e2e:
 	go test ./test/e2e/ -v -ginkgo.v
 
 .PHONY: lint
-lint: golangci-lint ## Run golangci-lint linter & yamllint
+lint: golangci-lint ## Run golangci-lint linter
 	$(GOLANGCI_LINT) run
 
 .PHONY: lint-fix
--- a/testdata/project-v4-multigroup-with-deploy-image/Makefile
+++ b/testdata/project-v4-multigroup-with-deploy-image/Makefile
@@ -69,7 +69,7 @@ test-e2e:
 	go test ./test/e2e/ -v -ginkgo.v
 
 .PHONY: lint
-lint: golangci-lint ## Run golangci-lint linter & yamllint
+lint: golangci-lint ## Run golangci-lint linter
 	$(GOLANGCI_LINT) run
 
 .PHONY: lint-fix
--- a/testdata/project-v4-multigroup/Makefile
+++ b/testdata/project-v4-multigroup/Makefile
@@ -69,7 +69,7 @@ test-e2e:
 	go test ./test/e2e/ -v -ginkgo.v
 
 .PHONY: lint
-lint: golangci-lint ## Run golangci-lint linter & yamllint
+lint: golangci-lint ## Run golangci-lint linter
 	$(GOLANGCI_LINT) run
 
 .PHONY: lint-fix
--- a/testdata/project-v4-with-deploy-image/Makefile
+++ b/testdata/project-v4-with-deploy-image/Makefile
@@ -69,7 +69,7 @@ test-e2e:
 	go test ./test/e2e/ -v -ginkgo.v
 
 .PHONY: lint
-lint: golangci-lint ## Run golangci-lint linter & yamllint
+lint: golangci-lint ## Run golangci-lint linter
 	$(GOLANGCI_LINT) run
 
 .PHONY: lint-fix
--- a/testdata/project-v4-with-grafana/Makefile
+++ b/testdata/project-v4-with-grafana/Makefile
@@ -69,7 +69,7 @@ test-e2e:
 	go test ./test/e2e/ -v -ginkgo.v
 
 .PHONY: lint
-lint: golangci-lint ## Run golangci-lint linter & yamllint
+lint: golangci-lint ## Run golangci-lint linter
 	$(GOLANGCI_LINT) run
 
 .PHONY: lint-fix
--- a/testdata/project-v4/Makefile
+++ b/testdata/project-v4/Makefile
@@ -69,7 +69,7 @@ test-e2e:
 	go test ./test/e2e/ -v -ginkgo.v
 
 .PHONY: lint
-lint: golangci-lint ## Run golangci-lint linter & yamllint
+lint: golangci-lint ## Run golangci-lint linter
 	$(GOLANGCI_LINT) run
 
 .PHONY: lint-fix


lukas016 (changes requested):

lukas016:
@mateusoliveira43 your PR contains unformatted source code. Are you sure your suggestion for delete fmt job is correct?

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_types.go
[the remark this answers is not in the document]
mateusoliveira43: @lukas016 here is the explanation about unformatted source code

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -52,16 +52,8 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
-.PHONY: fmt
-fmt: ## Run go fmt against code.
-	go fmt ./...
```
lukas016, line 57: are you sure with that? I think it was good practice call formatting for every build/test/run. Lint is optional and fix too. So you cannot forget call this targets before pushing changes otherwise you can create many changes in commits.

I personally suggest keep fmt and don't have dependency to third party software.
it can help avoid unnecessary changes in PR.

Vet can be deleted.

lukas016: Additional complication can be linter config. Linter config is manage by user and he can very easy disable fmt linter. This hardcoded way in makefile looks better for me.

mateusoliveira43: > I personally suggest keep fmt and don't have dependency to third party software.
it can help avoid unnecessary changes in PR.

Agreed, will add back!

I was thinking about directly calling `go fmt` inside `make generate`, and still remove `fmt` command from Makefile (without changing `make generate` current behavior). This way, kubebuilder CI would still prevent unnecessary changes in PRs, and users would have more control over generated code. 

What you think?

lukas016: i am not sure. you will create dependency between formatting and generating what are independent process (they depend on the code).

Is it okay, if fmt fail and generate won't run but independently it can run?
What can be weird is make target  name. It is generate so i expect it will create something but fmt just modify current state.

I am not sure and i think we are doing overengineering. Did you heard some complain about current solution?
I personally don't see big benefits in vet because it is linter and it is part of golangci-lint so it can be deleted.
But other stuff i kept as it is because i don't see big benefits and it can create just new problems.

I think opinion from someelse can be helpful.

mateusoliveira43: My only complain is that  `fmt` command does not have a CI counterpart, so I created one in the past (called it `fmt-isuptodate`) in my project Makefile. After understanding `golangci-lint`, I removed `fmt` and `fmt-isuptodate` and started using `lint-fix` and `lint` respectively, in the project Makefile.

> What can be weird is make target name. It is generate so i expect it will create something but fmt just modify current state. 

hmmm, one other view we can have then, is updating generate command to generate the code already formatted? Looking in the generated Makefile command, the two code generation commands, `generate` and `manifests` do not run `fmt` (but I think Go is smart enough to ignore format in generated files by controller-gen, so this can be bad example :grimacing: )

lukas016: CI counterpart. CI can be extremely different and it depends from environment and CI/CD software. I am using gitlab and i have own ci which just call make fmt in projects, so i never had problem with ci/cd. It exists a lot of CI/CD solutions and i don't think it is good idea force some ci/cd solution into this project, which is technically template for another projects. I think our main target has to be easy usage for local development.

mateusoliveira43: alright! I think today I will not have time, but tomorrow I will commit changes based on our conversation :grinning:

mateusoliveira43: @lukas016 just to keep you posted: I found some problems in scripts that generate docs testdata projects. I am working on fixing them. Once I finished (this week, but not sure which day), I will open a PR for that and update this one to just remove `vet` command (reverting removal of `fmt` command)

lukas016 (approved):
LGTM

Note for thinking:
if we want to push programmers to write better code, i suggest call linter as dependency for build target.

docs/book/src/component-config-tutorial/testdata/project/Makefile
```diff
@@ -56,12 +56,8 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
 fmt: ## Run go fmt against code.
 	go fmt ./...
 
-.PHONY: vet
-vet: ## Run go vet against code.
-	go vet ./...
```
camilamacedo86, line 61: The lint will validate the vet as well.
But I am not sure if we want to remove this one.
Users can still only call vet if they wish using go 

So, I would say that I think is better we keep it as it is

camilamacedo86 (changes requested):
The lint will validate the vet as well.
But I am not sure if we want to remove this one.
Users can still only call vet if they wish using go 

Either we do not call the lint to generate the manifests and etc.

So, I would say that I think is better we keep it as it is

docs/book/src/component-config-tutorial/testdata/project/Makefile
```diff
@@ -56,12 +56,8 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
...
-
```
camilamacedo86, line 62: The lint will validate the vet as well.
But I am not sure if we want to remove this one.
Users can still only call vet if they wish using go 

Either we do not call the lint to generate the manifests and etc.

So, I would say that I think is better we keep it as it is

mateusoliveira43: For me is a duplication and can be misleading for developers

Example: first time on a project, I run `make help` and see there is `fmt`, `vet` and `lint` commands. Do I need to run all of them?

Since kubebuilder already generates everything needed for developers to use `golangci-lint` instead of `fmt` and `vet`, I would enforce developers to use it

But we can discuss in a issue for example, will undo vet removal

camilamacedo86: The lint is a good practice and has its purpose.
however, this dev targets using go features are used  for the other targets 
Note that some people might do not want the linters or those then they are able to just remove from their project

However, kubebuilder has the goal to help developer and scaffold the default common/adhered used options. 
So, IMO we should keep it as it is.

mateusoliveira43: yeah, I though about just removing on my project, but I am afraid that can break how I update kubebuilder. I did not try out `kubebuilder alpha generate` yet, but the way I do today is as described here https://github.com/migtools/oadp-non-admin?tab=readme-ov-file#architecture

so I try to not change what came in Makefile, just add more commands in the end of the file

made the changes

camilamacedo86: if you just remove go vet, go fmt and keep the manifests and etc it probably will work.

camilamacedo86 (changes requested):
I understand the intent behind streamlining our development tools by leveraging `golangci-lint` in our Makefile. However, I recommend not removing `go vet` and `go fmt` for the following reasons:

- **Complementary Checks**: While `golangci-lint` is a powerful tool that lints our project against a wide range of criteria, `go vet` and `go fmt` serve as foundational checks for code correctness and formatting. These checks play a crucial role in early problem detection and code quality maintenance.

- **Developer Workflow**: Retaining `go vet` and `go fmt` allows developers to quickly address basic code quality issues independently of the more comprehensive linting process. This facilitates a smoother development workflow, enabling focused and efficient corrections.

- **Code Quality and Consistency**: The combination of these tools ensures a high standard of code quality and consistency across the project. Each tool has its strengths, and together, they cover a broader spectrum of potential issues.

In summary, while `golangci-lint` enhances our code quality checks, `go vet` and `go fmt` provide essential, foundational checks that should not be overlooked. Their removal could lead to gaps in our code quality assurance process. Therefore, I advocate for keeping both, as they complement each other and collectively strengthen our development practices.

mateusoliveira43:
@camilamacedo86 can you take a look?

camilamacedo86 (approved):
/lgtm

@mateusoliveira43 

Thank you a lot !!

I just change the title because we use it in the release notes
In this case, we use 🐛 because it change the project scaffold for the users

PR #3834, opened 2024-03-30
🌱 Fix makefile target make generate

camilamacedo86:
Small fix for do not finish with error


camilamacedo86:
to small make no sense ask for someone waste time on reviews here

PR #3837, opened 2024-03-31
:bug: fix resource path generation for resources without a specified group

camilamacedo86:
Fix the path when a group is not informed to create an API


pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/crd/patches/enablecainjection_patch.go
```diff
@@ -34,7 +34,7 @@ type EnableCAInjectionPatch struct {
 // SetTemplateDefaults implements file.Template
 func (f *EnableCAInjectionPatch) SetTemplateDefaults() error {
 	if f.Path == "" {
```
Kavinjsir, line 36: nit: would it be more readable to trim nested `if-else` statement like:
```suggestion
	if f.Path == "" {
		pathTemplate := "config/crd/patches/cainjection_in_%[plural].yaml"
		if f.MultiGroup && f.Resource.Group != "" {
			pathTemplate = "config/crd/patches/cainjection_in_%[group]_%[plural].yaml"
		}
		f.Path = filepath.Join(pathTemplate)
	}
```

Kavinjsir (approved):
/lgtm

Issue #3761, opened 2024-01-30
webhook patches has not been generated properly for projects with multi-group layout


camilamacedo86:
### What broke? What's expected?

I would expected be able to generate the dist/installer.yaml for projects which has multi-group layout
See that we needed to ignore it in the sample:
https://github.com/kubernetes-sigs/kubebuilder/blob/5278143918eab240cb2a8f424da1ca1f06dcedff/test/testdata/generate.sh#L118-L126

### Reproducing this issue

Just update the script to generate the samples under test data https://github.com/kubernetes-sigs/kubebuilder/blob/5278143918eab240cb2a8f424da1ca1f06dcedff/test/testdata/generate.sh#L118-L126 in order to allow run `make build-installer` for all projects which go/v4 and see that it fails as described in the comment for the multigroup layout. 

### KubeBuilder (CLI) Version

master

PR #3838, opened 2024-03-31
:bug: Fix the CRD kustomization path logic to ensure webhook patches are generated exclusively for resources that are configured with webhooks

camilamacedo86:
This PR addresses a bug in the generation process of the dist/installer for the testdata sample with multi-group layouts. Previously, the process incorrectly generated webhook patches for all resources, leading to failures in installer creation. With this update, webhook patches are now correctly generated only for resources that have webhooks configured, resolving the issue and ensuring successful generation of the installer.

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/3761

### DIFF
--- a/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/crd/kustomization.go
+++ b/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/crd/kustomization.go
@@ -88,9 +88,11 @@ func (f *Kustomization) GetCodeFragments() machinery.CodeFragmentsMap {
 	if f.MultiGroup && f.Resource.Group != "" {
 		suffix = f.Resource.Group + "_" + f.Resource.Plural
 	}
-	// Generate resource code fragments
-	webhookPatch := make([]string, 0)
-	webhookPatch = append(webhookPatch, fmt.Sprintf(webhookPatchCodeFragment, suffix))
+
+	if !f.Resource.Webhooks.IsEmpty() {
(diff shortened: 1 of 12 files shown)


pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/crd/kustomization.go
```diff
@@ -88,9 +88,13 @@ func (f *Kustomization) GetCodeFragments() machinery.CodeFragmentsMap {
...
-	// Generate resource code fragments
-	webhookPatch := make([]string, 0)
-	webhookPatch = append(webhookPatch, fmt.Sprintf(webhookPatchCodeFragment, suffix))
+
+	if !f.Resource.Webhooks.IsEmpty() {
+		webhookPatch := make([]string, 0)
+		// Change the line to uncomment the patch line
+		webhookPatch = append(webhookPatch, fmt.Sprintf(webhookPatchCodeFragment, suffix))
+		fragments[machinery.NewMarkerFor(f.Path, webhookPatchMarker)] = webhookPatch
+	}
```
Kavinjsir, line 97: nit: does it make sense if we simplify the code lines like:
```suggestion
	if !f.Resource.Webhooks.IsEmpty() {
		webhookPatch := fmt.Sprintf(webhookPatchCodeFragment, suffix)
		fragments[machinery.NewMarkerFor(f.Path, webhookPatchMarker)] = []string{webhookPatch}
	}
```

Kavinjsir (approved):
/lgtm

PR #3842, opened 2024-04-02
⚠️ Change the Registry from gcr.io to registry.k8s.io

camilamacedo86:
Change the Registry to registry.k8s.io
Reason: https://aws.amazon.com/blogs/containers/changes-to-the-kubernetes-container-image-registry/

### DIFF
--- a/build-legacy/build.sh
+++ b/build-legacy/build.sh
@@ -22,7 +22,7 @@
 set -eu
 
 SOURCE_IMAGE_TAG="quay.io/brancz/kube-rbac-proxy:${KUBE_RBAC_PROXY_VERSION}"
-TARGET_IMAGE_TAG="gcr.io/kubebuilder/kube-rbac-proxy:${KUBE_RBAC_PROXY_VERSION}"
+TARGET_IMAGE_TAG="registry.k8s.io/kubebuilder/kube-rbac-proxy:${KUBE_RBAC_PROXY_VERSION}"
 
 # Each arch to pull an image for.
 declare ARCHES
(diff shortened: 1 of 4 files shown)


build-legacy/build.sh
```diff
@@ -22,7 +22,7 @@
 set -eu
 
 SOURCE_IMAGE_TAG="quay.io/brancz/kube-rbac-proxy:${KUBE_RBAC_PROXY_VERSION}"
-TARGET_IMAGE_TAG="gcr.io/kubebuilder/kube-rbac-proxy:${KUBE_RBAC_PROXY_VERSION}"
+TARGET_IMAGE_TAG="registry.k8s.io/kubebuilder/kube-rbac-proxy:${KUBE_RBAC_PROXY_VERSION}"
```
upodroid, line 25: You need to publish the image to `gcr.io/k8s-staging-kubebuilder/kube-rbac-proxy` first.

Btw, this is changing later for all staging project.

upodroid (changes requested):

build-legacy/build.sh
camilamacedo86: Can I not push from GCP to `registry.k8s.io/kubebuilder/kube-rbac-proxy` directly so that we can ensure that all images will be there ASAP? That is to build in GCP and just push

The `gcr.io/k8s-staging-kubebuilder/kube-rbac-proxy` <- I understand that will not work until kube-rbac-proxy be under kubernetes-sig anyway

upodroid: It will work. You can push to the staging image registry.

upodroid: > What I am trying to do here is to change the job to see if we can re-tag/re-build all in GCP and push to the k8s registry so that we can have all images that today is in : https://console.cloud.google.com/gcr/images/kubebuilder/GLOBAL/kube-rbac-proxy there? OR at least the latest 3 releases/versions.

I can manually copy and promote those as discussed.

camilamacedo86: That would be amazing !!!!
Thank you a lot !!!

camilamacedo86:
@

PR #3846, opened 2024-04-05
✨  Upgrade golangci from 1.54 to 1.57

camilamacedo86:
* Upgrade golangci from 1.54 to 1.57
* Fix lint issues after upgrade by:

a) Ensure that in all places we are ignoring the exception equally

```go
         // nolint:revive
	. "github.com/onsi/ginkgo/v2"
	// nolint:revive
	. "github.com/onsi/gomega"
```

b) Ensure that we use `_` for args not explicit called in the funcs. Example

```go
PreRunE: func(_ *cobra.Command, _ []string) error {
```

### DIFF
--- a/docs/book/src/component-config-tutorial/testdata/project/test/e2e/e2e_suite_test.go
+++ b/docs/book/src/component-config-tutorial/testdata/project/test/e2e/e2e_suite_test.go
@@ -20,7 +20,9 @@ import (
 	"fmt"
 	"testing"
 
+	// nolint:revive
 	. "github.com/onsi/ginkgo/v2"
+	// nolint:revive
 	. "github.com/onsi/gomega"
 )
 
--- a/.github/workflows/lint-sample.yml
+++ b/.github/workflows/lint-sample.yml
@@ -40,14 +40,14 @@ jobs:
       - name: Run linter
         uses: golangci/golangci-lint-action@v4
         with:
-          version: v1.54
+          version: v1.57
           working-directory: testdata/project-v4-with-deploy-image
           args: --config .golangci.yml ./...
           skip-cache: true  # first lint action will handle
       - name: Run linter
         uses: golangci/golangci-lint-action@v4
         with:
-          version: v1.54
+          version: v1.57
           working-directory: testdata/project-v4-multigroup-with-deploy-image
           args: --config .golangci.yml ./...
           skip-cache: true  # first lint action will handle
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Run linter
         uses: golangci/golangci-lint-action@v4
         with:
-          version: v1.54
+          version: v1.57
 
   yamllint:
     runs-on: ubuntu-latest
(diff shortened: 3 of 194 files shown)


Kavinjsir (approved):
/lgtm

docs/book/src/component-config-tutorial/testdata/project/test/e2e/e2e_suite_test.go
```diff
@@ -20,7 +20,9 @@ import (
 	"fmt"
 	"testing"
 
+	// nolint:revive
```
stmcginnis, line 23: Rather than having all of these `nolint:revive` lines, I think we are safe to assume dot imports will be used for intentional reasons. I think it would be cleaner to just remove the `dot-imports` rule from line 28 in `.golangci.yml`.

Kavinjsir: Agreed. One way maybe:
```yaml
[rule.dot-imports]
  arguments = [{ allowedPackages = ["github.com/onsi/ginkgo/v2","github.com/onsi/gomega"] }]
```

camilamacedo86: We should not use dot imports
Just the case of ginkgo and gomega which are an exception.
So, we still want the check 

If a config like 

```
[rule.dot-imports]
  arguments = [{ allowedPackages = ["github.com/onsi/ginkgo/v2","github.com/onsi/gomega"] }]
```
Works then, It is totally fine.

camilamacedo86 (commented):
Hi @stmcginnis 

> Rather than having all of these nolint:revive lines, I think we are safe to assume dot imports will be used for intentional reasons. I think it would be cleaner to just remove the dot-imports rule from line 28 in .golangci.yml.

We should still not using alias with `.`.  However, yes we have one exception which we have been using // nolint to ignore those which are only. Then, if we do not check it probably we will end up using it without a good reason too. 

```
	// nolint:revive
	. "github.com/onsi/ginkgo/v2"
	// nolint:revive
	. "github.com/onsi/gomega"
```

The lint fix of this PR just ensure that we have those applied equally in all places
Note that seems that some files we were not checking and because of this they were without the `//nolint` check.

However, to ignore this case I totally agree that would be nice if we could find a way to configure it as @Kavinjsir suggested. He also open an issue: https://github.com/kubernetes-sigs/kubebuilder/issues/3847 (which I understand would have this goal)

Therefore, IMO it seems fine we try to replace those in a follow up with the suggestion if we be able to check that can works. +1 💯 

The other rule that we need to fix here is for we use `_` instead of args name when those are not used which is also a valid rule see: (i.e. https://github.com/kubernetes-sigs/kubebuilder/pull/3846/files#diff-e5f2449f45b05bf525a27e22a412864f1d479ac4548d5ffe3c90da55fdfee3f6R34-R39)

In this way, since:
-  @Kavinjsir /lgtm this one
- It only do the bump and ensure that we are ignoring equally in all place the `ginkgo ` and `gomega ` exception regards the dots imports
- Do the fix of the `_` for args not used in the missing places
- And we have a issue to see if we can improve in a follow up
- By last it is too silly small/nit one

I think it is safe we get this one merged and I am going forward with this one. 
I hope you both has no objections. 

**But please feel free to contribute within for we improve this in a follow up as you see that it would fit.** 

**Thank you a lot for your collaboration @Kavinjsir and @stmcginnis 🎉**

PR #3848, opened 2024-04-06
📖 Fix Generate Sample for Getting Started Tutorial

camilamacedo86:
- We should not generate the sample using the flag `"--component-config"`
- We should not use either `go/v4` so that tomorrow if we have go/v5 and etc as default we do not need to update this code. we should always generate with the default plugin


camilamacedo86:
/override pull-kubebuilder-e2e-k8s-1-29-2

Reason:https://kubernetes.slack.com/archives/CCK68P2Q2/p1712378775102839

camilamacedo86 (commented):
That is too small / simple so has no reason for I make someone waste time doing this review. Moving forward.

PR #3853, opened 2024-04-07
✨  Add Network Policy

camilamacedo86:
Now, users will be able to use a helper using network policies to protected the metrics endpoint. If webhooks are scaffold then, a Network Policy to allow the traffic within is also added to the project. 

**Motivations**

Partially Close: https://github.com/kubernetes-sigs/kubebuilder/issues/3871

### DIFF
--- a/docs/book/src/reference/metrics.md
+++ b/docs/book/src/reference/metrics.md
@@ -188,11 +188,19 @@ enhance the controller-runtime and address these considerations.
 </aside>
 
 
-### By using Network Policy
+### By using Network Policy (You can optionally enable)
 
 NetworkPolicy acts as a basic firewall for pods within a Kubernetes cluster, controlling traffic
-flow at the IP address or port level. However, it doesn't handle authentication (authn), authorization (authz),
-or encryption directly like [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) solution.
+flow at the IP address or port level. However, it doesn't handle `authn/authz`.
+
+Uncomment the following line in the `config/default/kustomization.yaml`:
+
+```
+# [NETWORK POLICY] Protect the /metrics endpoint and Webhook Server with NetworkPolicy.
+# Only Pod(s) running a namespace labeled with 'metrics: enabled' will be able to gather the metrics.
+# Only CR(s) which uses webhooks and applied on namespaces labeled 'webhooks: enabled' will be able to work properly.
+#- ../network-policy
+```
 
 ### By exposing the metrics endpoint using HTTPS and CertManager
 
--- a/.github/workflows/test-sample-go.yml
+++ b/.github/workflows/test-sample-go.yml
@@ -24,7 +24,7 @@ jobs:
         run: |
           KUSTOMIZATION_FILE_PATH="testdata/project-v4/config/default/kustomization.yaml"
           sed -i '25s/^#//' $KUSTOMIZATION_FILE_PATH
-          sed -i '46s/^#//' $KUSTOMIZATION_FILE_PATH
+          sed -i '51s/^#//' $KUSTOMIZATION_FILE_PATH
 
       - name: Test
         run: |
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
@@ -27,6 +27,11 @@ resources:
 - ../prometheus
 # [METRICS] Expose the controller manager metrics service.
 - metrics_service.yaml
+# [NETWORK POLICY] Protect the /metrics endpoint and Webhook Server with NetworkPolicy.
+# Only Pod(s) running a namespace labeled with 'metrics: enabled' will be able to gather the metrics.
+# Only CR(s) which requires webhooks and are applied on namespaces labeled with 'webhooks: enabled' will
+# be able to communicate with the Webhook Server.
+#- ../network-policy
 
 # Uncomment the patches line if you enable Metrics, and/or are using webhooks and cert-manager
 patches:
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/network-policy/allow-metrics-traffic.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/network-policy/allow-metrics-traffic.yaml
@@ -1,0 +1,26 @@
+# This NetworkPolicy allows ingress traffic
+# with Pods running on namespaces labeled with 'metrics: enabled'. Only Pods on those
+# namespaces are able to gathering data from the metrics endpoint.
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  labels:
+    app.kubernetes.io/name: project
+    app.kubernetes.io/managed-by: kustomize
+  name: allow-metrics-traffic
+  namespace: system
+spec:
+  podSelector:
+    matchLabels:
+      control-plane: controller-manager
+  policyTypes:
+    - Ingress
+  ingress:
+    # This allows ingress traffic from any namespace with the label metrics: enabled
+    - from:
+      - namespaceSelector:
+          matchLabels:
+            metrics: enabled  # Only from namespaces with this label
+      ports:
+        - port: 8443
+          protocol: TCP
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/network-policy/allow-webhook-traffic.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/network-policy/allow-webhook-traffic.yaml
@@ -1,0 +1,26 @@
+# This NetworkPolicy allows ingress traffic to your webhook server running
+# as part of the controller-manager from specific namespaces and pods. CR(s) which uses webhooks
+# will only work when applied in namespaces labeled with 'webhook: enabled'
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  labels:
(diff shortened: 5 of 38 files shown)


docs/book/src/reference/metrics.md
```diff
@@ -3,13 +3,66 @@
...
+## For the future
+
+By using the [Plugins API provided by Kubebuilder](./../plugins/plugins.md),
+it might be possible to create an [external plugin](./../plugins/creating-plugins.md)
+to properly integrate [`kube-rbac-proxy`](https://github.com/brancz/kube-rbac-proxy)
+with Kubebuilder and provide a helper to allow users to opt-in as please them. Furthermore, when
+it to be part of kubernetes ecosystem the project will be able to either
+maintain its image using the registry.k8s.io.
+
+</aside>
```
camilamacedo86, line 63: @mahamed ^ Update the notice

docs/book/src/reference/metrics.md
```diff
@@ -3,13 +3,66 @@
...
+
+Projects initialized with Kubebuilder versions prior to `v3.15.0`
+utilize [`kube-rbac-proxy`](https://github.com/brancz/kube-rbac-proxy)
+to protect the metrics endpoint.
+
+## If you are using OR wish to continue using `kube-rbac-proxy`:
+
+Ensure that you use an image provided by the [`kube-rbac-proxy`](https://github.com/brancz/kube-rbac-proxy).
+See its [repository (`quay.io/brancz/kube-rbac-proxy`)](https://quay.io/repository/brancz/kube-rbac-proxy?tab=tags&tag=latest)
+**Images provided under `gcr.io/kubebuilder/` will be unavailable according to Google
```
camilamacedo86, line 35:
```suggestion
**Images provided under `gcr.io/kubebuilder/` (Container registry) will be unavailable according to Google
```

varshaprasad96 (commented):
Did a first pass, and the NetworkPolicy additions look good. Just a few nits in the documentation.

docs/book/src/component-config-tutorial/testdata/project/config/prometheus/monitor.yaml
```diff
@@ -11,11 +11,8 @@ metadata:
 spec:
   endpoints:
     - path: /metrics
-      port: https
-      scheme: https
-      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
-      tlsConfig:
-        insecureSkipVerify: true
+      port: http # Ensure this is the name of the port that exposes HTTP metrics
```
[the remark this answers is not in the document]
camilamacedo86: Before was not secure as well. 
We were using `insecureSkipVerify: true`

If we we use only `NetworkPolicy` than it is HTTP
for we are able to use `HTTPs` we need also use cert-manager

However, aiming allow users opt-in and opt-out as make it simple for who is just getting started, as described in the proposal IHMO the second phase here would be a follow up where we add as Optional the usage of cert-manager. So, if the user uncomment the kustomize section to use it with cert-manager we will patch the ServiceMonitor to be like the example added in the proposal.

See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/discontinue_usage_of_kube_rbac_proxy.md#phase-2-add-cert-manager-as-optional-option-to-be-used-with-metrics

camilamacedo86:
Hi @varshaprasad96 

Thank you a lot for your time and review. 
All your suggestions are addressed.

docs/book/src/component-config-tutorial/testdata/project/config/default/manager_metrics_patch.yaml
```diff
@@ -0,0 +1,18 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+  namespace: system
+spec:
+  template:
+    metadata:
+      labels:
+        allow: allow-metrics
```
erikgb, line 11: I don't think this extra label is required. We should instead use existing labels addressing the manager pods. This comment explains how I think this should be configured overall: https://github.com/prometheus-operator/kube-prometheus/discussions/1907#discussioncomment-3896712. The network policy gives 0 security without the namespace selector selecting namespaces to access ingress from, which I think we should be extremely careful to not mislead our users by noting this as a "replacement". In my opinion, removing the opt-in kube-rbac-proxy and adding an opt-in for a network policy are two different things.

camilamacedo86: Hi @erikgb,

Thank you for your feedback. Following the comments inline

Regards

> `The network policy gives 0 security without the namespace selector selecting namespaces to access ingress from, which I think we should be extremely careful to not mislead our users by noting this as a "replacement".`

**IMPORTANT: We are NOT saying that is a replacement at all.**
Please, see that is clarified with WARNING class in the docs and Proposal.
However, it is very important and we MUST to have the phase 1 scaffold in place. Otherwise we end up without no protection at all, and either without NO possibility to warn.

We cannot do a PR with ALL phases by once. it is too much changes to address then, because of this we agreed in the proposal to address it per phases. 

See that use ONLY network policies does not ensure a good protection at all. and we are clarifying it in the docs + proposal. I update the title of this PR if it is what brings the concern to you.

Regards the other raised points I will check each of them out and I will came back to you.

camilamacedo86: Hi @erikgb 

Again, really thank you for all your feedback.
I tried to reply all above and see that from mainly the 4/5 items suggested by you we are addressing 3 :-)

See that I added what is missing get done here in the description.

docs/book/src/cronjob-tutorial/testdata/project/config/policy/policy.yaml
```diff
@@ -0,0 +1,28 @@
...
+    - Ingress
+  ingress:
+    - from:
+      - podSelector:
+          matchLabels:
+            # Pods which will collect the metrics must have this label
+            role: allow-metrics
+      - namespaceSelector:
+          matchLabels:
+            # Only allow namespaces labeled with 'role: allow-metrics' to access metrics
```
camilamacedo86, line 24: @erikgb ˆ

docs/book/src/cronjob-tutorial/testdata/project/config/policy/allow-metrics-traffic.yaml
```diff
@@ -0,0 +1,29 @@
...
+    app.kubernetes.io/name: project
+    app.kubernetes.io/managed-by: kustomize
+  name: allow-metrics-traffic
+  namespace: system
+spec:
+  podSelector:
+    matchLabels:
+      control-plane: controller-manager
+  policyTypes:
+    - Egress
```
erikgb, line 17: Why are we addressing egress traffic in these network policies at all? I don't think egress traffic is required in general from operators, except that it must be able to reach the API server. Access to the API server should be addressed more generally if using network policies in a cluster.

As a cluster-admin I would not be happy if a workload is granted any egress traffic if not required.

docs/book/src/reference/metrics.md
```diff
@@ -70,7 +70,7 @@ operational metrics. This exposure can lead to security vulnerabilities
 where an attacker could gain insights into the system's operation
 and exploit weaknesses.
 
-### By using authn/authz (Enabled by default)
+### (Enabled by default) By using authn/authz
```
erikgb, line 73: I think the enabled/disabled postfix style reads better.

docs/book/src/reference/metrics.md
```diff
@@ -189,11 +189,19 @@ enhance the controller-runtime and address these considerations.
...
-or encryption directly like [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) solution.
+flow at the IP address or port level. However, it doesn't handle `authn/authz`.
+
+Uncomment the following line in the `config/default/kustomization.yaml`:
+
+```
+# [NETWORK POLICY] Protect the /metrics endpoint and Webhook Server with NetworkPolicy.
+# Only Pod(s) running a namespace labeled with 'metrics: enabled' will be able to gather the metrics.
+# Only CR(s) which uses webhooks and applied on namespaces labeled 'webhooks: enabled' will be able to work properly.
+#- ../policy
```
erikgb, line 203:
```suggestion
#- ../network-policy
```

Or are we trying to "catch-all" eventual policies here?

camilamacedo86: I think the rename is a good feedback !
Done

camilamacedo86:
Hi @erikgb 

All suggestions were addressed.
Really thank you for the amazing review !!!!

docs/book/src/component-config-tutorial/testdata/project/config/prometheus/monitor.yaml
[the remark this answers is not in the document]
camilamacedo86: That is changed as well !!

camilamacedo86:
@varshaprasad96 @joelanford 

I think this one is great to fly !!!
Please, also feel free to check out

varshaprasad96:
Took a second pass, and the changes look good. I'll just try these out by end of this week and update here! Thanks Camila!

camilamacedo86:
Hi @varshaprasad96 

> Took a second pass, and the changes look good. I'll just try these out by end of this week and update here! Thanks Camila!

That is great !
Thank you a lot.

See that we are testing it out in the e2e tests
If we be missing any scenario than I think would be nice we add in the e2e test so that no required manual tests is needed.
Let me know if you see any scenario that we are not checking out automatically

varshaprasad96 (approved):
/lgtm
/approve

PR #3855, opened 2024-04-08
📖 docs: client and its methods should be private to the controller

grosser:
**problem**: 
The controller example suggests to make the `Client` public in the controller,
which leads to:
- hard to understand calls like `r.Patch` (where does `Patch` come from -> scroll through the code)
- method overload issues if Patch/Get/etc are needed as methods on the controller
- the methods being publicly available on the controller for other parts of the codebase to use, which they should not be

**solution A**: make client private (ideal but abandoned)
this leads to:
- code like `FooReconciler{Client: client, ...}` breaking and these reconcilers needing `NewFooReconciler` methods ... which is technically correct but a lot of change
- old code breaking when new code is added since `r.Client` does not work

**solution B**: make Client not be mixed in (this PR)
this leads to new code being compatible with old code and less change in general
but it's not really private, but at least more obvious than before

### DIFF
--- a/designs/code-generate-image-plugin.md
+++ b/designs/code-generate-image-plugin.md
@@ -56,7 +56,7 @@ Add the new plugin code generate which will scaffold code implementation to depl
 	size := {{ resource }}.Spec.Size
 	if *found.Spec.Replicas != size {
 		found.Spec.Replicas = &size
-		err = r.Update(ctx, found)
+		err = r.Client.Update(ctx, found)
 		if err != nil {
 			log.Error(err, "Failed to update Deployment", "Deployment.Namespace", found.Namespace, "Deployment.Name", found.Name)
 			return ctrl.Result{}, err
@@ -115,7 +115,7 @@ func (r *{{ resource }}.Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 
 	// Fetch the {{ resource }} instance
 	{{ resource }} := &{{ apiimportalias }}.{{ resource }}{}
-	err := r.Get(ctx, req.NamespacedName, {{ resource }})
+	err := r.Client.Get(ctx, req.NamespacedName, {{ resource }})
 	if err != nil {
 		if errors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
@@ -131,12 +131,12 @@ func (r *{{ resource }}.Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 
 	// Check if the deployment already exists, if not create a new one
 	found := &appsv1.Deployment{}
-	err = r.Get(ctx, types.NamespacedName{Name: {{ resource }}.Name, Namespace: {{ resource }}.Namespace}, found)
+	err = r.Client.Get(ctx, types.NamespacedName{Name: {{ resource }}.Name, Namespace: {{ resource }}.Namespace}, found)
 	if err != nil && errors.IsNotFound(err) {
 		// Define a new deployment
 		dep := r.deploymentFor{{ resource }}({{ resource }})
 		log.Info("Creating a new Deployment", "Deployment.Namespace", dep.Namespace, "Deployment.Name", dep.Name)
-		err = r.Create(ctx, dep)
+		err = r.Client.Create(ctx, dep)
 		if err != nil {
 			log.Error(err, "Failed to create new Deployment", "Deployment.Namespace", dep.Namespace, "Deployment.Name", dep.Name)
 			return ctrl.Result{}, err
@@ -152,7 +152,7 @@ func (r *{{ resource }}.Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 	size := {{ resource }}.Spec.Size
 	if *found.Spec.Replicas != size {
 		found.Spec.Replicas = &size
-		err = r.Update(ctx, found)
+		err = r.Client.Update(ctx, found)
 		if err != nil {
 			log.Error(err, "Failed to update Deployment", "Deployment.Namespace", found.Namespace, "Deployment.Name", found.Name)
 			return ctrl.Result{}, err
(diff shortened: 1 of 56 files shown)


docs/book/src/cronjob-tutorial/testdata/emptycontroller.go
```diff
@@ -41,7 +41,7 @@ objects, so these are added out of the box.
 
 // CronJobReconciler reconciles a CronJob object
 type CronJobReconciler struct {
-	client.Client
+	client client.Client
```
camilamacedo86, line 44: If you run :

https://github.com/kubernetes-sigs/kubebuilder/blob/32e0fdc918b9a21485d7b042833402a62d0480f4/Makefile#L82-L85

It will be lost.

We need to the change in the hack/docs so that we can easily keep the samples from the docs updated.

grosser:
had to redo this since there were lots of places, ended up with to replacements:

```
Reconciler struct \{(\s+)client.Client
->
Reconciler struct {$1Client client.Client
```
and
```
r.(Update|Create|Status|Get|Patch|List|Delete)\(
->
r.Client.$1(
```

designs/simplified-scaffolding.md
```diff
@@ -468,7 +468,7 @@ import (
 )
 
 type MyKindReconciler struct {
-	client.Client
+	Client client.Client
```
camilamacedo86, line 471: The design docs should not be changed
Can you please revert

designs/code-generate-image-plugin.md
```diff
@@ -56,7 +56,7 @@ Add the new plugin code generate which will scaffold code implementation to depl
 	size := {{ resource }}.Spec.Size
 	if *found.Spec.Replicas != size {
 		found.Spec.Replicas = &size
-		err = r.Update(ctx, found)
```
camilamacedo86, line 59: The design docs should not be changed
Can you please revert

pkg/plugins/golang/declarative/v1/scaffolds/internal/templates/controller.go
```diff
@@ -96,7 +96,7 @@ var _ reconcile.Reconciler = &{{ .Resource.Kind }}Reconciler{}
 
 // {{ .Resource.Kind }}Reconciler reconciles a {{ .Resource.Kind }} object
 type {{ .Resource.Kind }}Reconciler struct {
-	client.Client
+	Client client.Client
```
camilamacedo86, line 99: we should not change this plugin deprecated

pkg/plugins/golang/v2/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -80,7 +80,7 @@ import (
 
 // {{ .Resource.Kind }}Reconciler reconciles a {{ .Resource.Kind }} object
 type {{ .Resource.Kind }}Reconciler struct {
-	client.Client
+	Client client.Client
```
camilamacedo86, line 83: we should not change this plugin deprecated

pkg/plugins/golang/v3/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -80,7 +80,7 @@ import (
 
 // {{ .Resource.Kind }}Reconciler reconciles a {{ .Resource.Kind }} object
 type {{ .Resource.Kind }}Reconciler struct {
-	client.Client
+	Client client.Client
```
camilamacedo86, line 83: we should not change this plugin deprecated

camilamacedo86 (commented):
Hi @grosser 

Thank you for the contribution, before you still working on that could you please raise an issue and add example of code that explain your motivation?

Why do you want do this change.

It indeed is not a doc change but we would be change the scaffold and we need to understand the scenario, what you did ? why you did ? what did you expected and what you faced instead with examples if possible so that we can check if that is the best approach or not.

designs/simplified-scaffolding.md
grosser: the point of this PR is to say: "the design is bad and should be changed"
if we can't agree on that then the PR makes no sense :)

... and if we can't agree on that I'd love the know why ("leave it like it is" or is there some gotcha I'm missing ?)

camilamacedo86: This doc is a design solution implemented already.
So, the change should not be done in this specific documentation.

pkg/plugins/golang/declarative/v1/scaffolds/internal/templates/controller.go
camilamacedo86: We can only apply changes for kustomize/v2 and go/v4

pkg/plugins/golang/v2/scaffolds/internal/templates/controllers/controller.go
camilamacedo86: We can only apply changes for kustomize/v2 and go/v4

pkg/plugins/golang/v3/scaffolds/internal/templates/controllers/controller.go
camilamacedo86: We can only apply changes for kustomize/v2 and go/v4

camilamacedo86: Indeed it brokes the v3 one: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kubebuilder/3855/pull-kubebuilder-test/1777774950044667904

grosser:
https://github.com/kubernetes-sigs/kubebuilder/issues/3859

camilamacedo86:
Closing this one: https://github.com/kubernetes-sigs/kubebuilder/issues/3859#issuecomment-2067577496

Issue #3790, opened 2024-02-19
Invalid leader election ID is generated


yaa110:
### What broke? What's expected?

generated `LeaderElectionID` ends with a period (`.`) which raises the following error:

```log
E0219 20:35:10.387636       1 leaderelection.go:336] error initially creating leader election record: Lease.coordination.k8s.io "be3caa60." is invalid: metadata.name: Invalid value: "be3caa60.": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

### KubeBuilder (CLI) Version

Version: main.version{KubeBuilderVersion:"3.13.0", KubernetesVendor:"1.27.1", GitCommit:"c8a7cc58eeb56586c019cf8845dad37286d077ff", BuildDate:"2023-11-02T20:43:44Z", GoOs:"linux", GoArch:"amd64"}

### PROJECT version

3

### Plugin versions

```yaml
- go.kubebuilder.io/v4
```

### Other versions

go version go1.22.0 linux/amd64
kubectl Client Version: v1.29.2

camilamacedo86:
Hi @yaa110,

See that the value is generated here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/main.go#L282

`LeaderElectionID:       "{{ hashFNV .Repo }}.{{ .Domain }}",`

So, can you please share with us what values do you use to to create the project?
How we can reproduce it so that it can help us to fix it?

camilamacedo86:
This one seems to be a good first issue.
What we need to do here:
- a) First, we must see if we can reproduce the described issue. To do so, create a new project and ensure that the LeaderElection is enabled in the main.go. If not, then we can close this one
- b) IF we are able to reproduce then, we need to change the implementation https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/main.go#L282 to ensure that we are generating a random valid value.

Bharadwajshivam28:
Hey @camilamacedo86 I can take a look on this issue

Bharadwajshivam28:
I need some steps to get started as I am new contributor

[...]

PR #3876, opened 2024-04-27
:bug: fix Invalid leader election ID generated when domain is empty

camilamacedo86:
If the domain is not informed then we cannot use it to generate the ID for the leader election

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/3790


camilamacedo86 (commented):
To small moving forward

Issue #3811, opened 2024-03-06
Remove path webhook marker 


mneverov:
### What do you want to happen?

Kubebuilder works as expected when I initialize a project and generate webhooks with:

```sh
kubebuilder init
kubebuilder create api --group foogroup --version v1beta1 --kind FooKind
kubebuilder create webhook --group foogroup --version v1beta1 --kind FooKind --defaulting --programmatic-validation
```

My assumption is that generated code is intended to be modified including the webhook [markers](https://book.kubebuilder.io/reference/markers/webhook). This is not true regarding `path` marker. It MUST follow the pattern described in the documentation as the controller-runtime registers webhooks on [certain paths](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/builder/webhook.go#L245).

The `path` marker modification leads to a cryptic error:

```
Internal error occurred: failed calling webhook \"validate-foo\": failed to call webhook: the server could not find the requested resource
```

Suggestions:
- remove the `path` marker, but still generate valid `webhook/manifests.yaml`
- or add a comment in autogenerated code with the explanation that if the webhook is registered via the following code the path MUST not be modified
```
return ctrl.NewWebhookManagedBy(mgr).
  For(typ).
  Complete()
```

Something like:
```
// TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
// NOTE that path must follow the pattern described in kubebuilder documentation.
//+kubebuilder:webhook:path=/validate-foogroup-my-domain-v1beta1-fookind,mutating=false,failurePolicy=fail,sideEffects=None,groups=foogroup.my.domain,resources=fookinds,verbs=create;update,versions=v1beta1,name=vfookind.kb.io,admissionReviewVersions=v1
```

camilamacedo86:
Hi @mneverov 

Thank you for raising it 

It seems that you change it for an invalid path.
Could you please provide further information for we are able to understand your scenario:

1 - Add the original result of 

```
kubebuilder init
kubebuilder create api --group foogroup --version v1beta1 --kind FooKind
kubebuilder create webhook --group foogroup --version v1beta1 --kind FooKind --defaulting --programmatic-validation
```

2 - Then, let us know how did you change it to face the error
(so that we can reproduce it)

3 - Then, explain to us why did you make the change? What was the motivation for you want to change the path? 

[...]

PR #3877, opened 2024-04-27
:bug: Add comment to clarify that webhooks path in the marker values should not be changed

camilamacedo86:
Closes: - https://github.com/kubernetes-sigs/kubebuilder/issues/3811


camilamacedo86 (commented):
To small moving forward

PR #3922, opened 2024-05-17
⚠️ remove declarative plugin

camilamacedo86:
More info: https://github.com/kubernetes-sigs/kubebuilder/discussions/3598


camilamacedo86:
Moving forward with this one.
it is communicated, agreed and etc for too long.

Issue #3934, opened 2024-05-20
kustomize: strategicMergePatch clobbers patched arrays


joelanford:
### What broke? What's expected?

When I enabled the `manager_metrics_patch.yaml` in `config/default/kustomization.yaml`, it wiped out the arguments that I had setup in the `config/manager/manager.yaml`.

### Reproducing this issue

Initialize project and build manifests with and without metrics patch commented:
```
kubebuilder init --plugins go/v4 --domain example.org --owner "Your name" --repo example.org/test-project
make kustomize && bin/kustomize-v5.3.0 build config/default > default.yaml
sed -i 's/#- path: manager_metrics_patch.yaml/- path: manager_metrics_patch.yaml/' config/default/kustomization.yaml
make kustomize && bin/kustomize-v5.3.0 build config/default > metrics-patch.yaml
```

Diff the manifests:
```
diff default.yaml metrics-patch.yaml
151,153c151
<         - --leader-elect
<         - --health-probe-bind-address=:8081
<         - --metrics-bind-address=0
---
>         - --metrics-bind-address=0.0.0.0:8080
```

### KubeBuilder (CLI) Version

v3.15.0

### PROJECT version

v3

### Plugin versions

```yaml
go.kubebuilder.io/v4
```

camilamacedo86:
Hi @joelanford,

For now I think the best is we ensure that it is as it was before: https://github.com/kubernetes-sigs/kubebuilder/pull/3937
 I tracked an issue for we analyse the suggestion to use Pod instead: https://github.com/kubernetes-sigs/kubebuilder/issues/3936

PR #3937, opened 2024-05-20
🐛 fix: preserve existing flags when applying metrics patch

camilamacedo86:
Ensure that enabling the manager_metrics_patch.yaml in config/default/kustomization.yaml does not overwrite existing arguments in config/manager/manager.yaml. The patch now appends the --metrics-bind-address argument without replacing other arguments.

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/3934

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/cmd/main.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/cmd/main.go
@@ -76,7 +76,8 @@ func main() {
 	var probeAddr string
 	var secureMetrics bool
 	var enableHTTP2 bool
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metric endpoint binds to. "+
+		"Use the port :8080. If not set, it will be '0 in order to disable the metrics server")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
--- a/docs/book/src/getting-started/testdata/project/config/default/manager_metrics_patch.yaml
+++ b/docs/book/src/getting-started/testdata/project/config/default/manager_metrics_patch.yaml
@@ -1,13 +1,4 @@
 # This patch adds the args to allow exposing the metrics endpoint securely
-apiVersion: apps/v1
-kind: Deployment
-metadata:
-  name: controller-manager
-  namespace: system
-spec:
-  template:
-    spec:
-      containers:
-      - name: manager
-        args:
-        - "--metrics-bind-address=0.0.0.0:8080"
+- op: add
+  path: /spec/template/spec/containers/0/args/0
+  value: --metrics-bind-address=:8080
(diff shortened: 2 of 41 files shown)


joelanford:
This solves it for the default set of flags that we scaffold into `config/manager/manager.yaml`, but if someone wants to add a new flag they still have to add it to both places. That's why I'm wondering if we could use [json 6902 patch](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/patches/#patch-using-path-json6902) style so that the patch file _inserts_ an argument into the list provided by the base.

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/enable_matrics_patch.go
joelanford, line 1: By the way, typo in file name:

`s/enable_matrics_patch.go/enable_metrics_patch.go/`

camilamacedo86: I saw this one as well. I was avoiding have issues to cherry-pick but lets fix it.
The correct way is keep the go file name as the name of the file that is scaffolded.
Done 👍

camilamacedo86:
Hi @joelanford 

Regards: https://github.com/kubernetes-sigs/kubebuilder/pull/3937#issuecomment-2121474809
I changed it. We can do as you suggested but still we need alert about the position since it is based in positions. 
But I think it might be better as you suggested.

docs/book/src/getting-started/testdata/project/config/default/manager_metrics_patch.yaml
```diff
@@ -1,13 +1,3 @@
-# This patch adds the args to allow exposing the metrics endpoint securely
```
camilamacedo86, line 1: Revert comment

test/testdata/generate.sh
```diff
@@ -43,7 +43,12 @@ function scaffold_test_project {
   go mod init sigs.k8s.io/kubebuilder/testdata/$project  # our repo autodetection will traverse up to the kb module if we don't do this
 
   header_text "Initializing project ..."
-  $kb init $init_flags --domain testproject.org --license apache2 --owner "The Kubernetes authors"
+  if [[ $project == project-v4-with-grafana ]]; then
+      header_text 'Editing project with Grafana plugin ...'
+      $kb init $init_flags --domain testproject.org --license apache2 --owner "The Kubernetes authors"
+  else
+      $kb init $init_flags --domain testproject.org --license apache2 --owner "The Kubernetes authors"
+  fi
```
camilamacedo86, line 51: remove not related

docs/book/src/cronjob-tutorial/testdata/project/cmd/main.go
```diff
@@ -76,7 +76,8 @@ func main() {
 	var probeAddr string
 	var secureMetrics bool
 	var enableHTTP2 bool
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":0", "The address the metric endpoint binds to. "+
```
joelanford, line 79: I think `":0"` means "bind to any open port", whereas `"0"` means "don't start the metrics server", right?

```suggestion
	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metric endpoint binds to. "+
```

camilamacedo86: That works as well. But we can change no problem at all.

PR #3943, opened 2024-05-21
🌱 cleanup testdata generate.sh

camilamacedo86:
We are just removing what is no longer used from the  generate.sh


camilamacedo86:
We are just removing what is no longer used from the  generate.sh
If it pass in the CI is good enough because we ensure that all is generated as before.
So, I will not ask for nobody waste time on this review. We have big fishes to fry.

PR #3904, opened 2024-05-12
:bug: (CLI, deploy-image/v1alpha1, go/v4) Ensure consistent spacing in marker annotations

camilamacedo86:
This commit addresses the inconsistency in marker annotations within the kubebuilder project. Previously, kubebuilder markers did not have a space between `//` and `+marker`, unlike those in controller-tools and other related tools see; 
- controller-tools: https://github.com/search?q=repo%3Akubernetes-sigs%2Fcontroller-tools+%22%2F%2F+%2B%22&type=code. 
- controller-runtime: https://github.com/search?q=repo%3Akubernetes-sigs%2Fcontroller-runtime++%22%2F%2F+%2B%22&type=code

**This inconsistency was causing confusion for end users.**

To resolve this, all kubebuilder markers are now updated to include a space, ensuring uniformity across the project. After merging this commit, you can run `make all` to regenerate the markers. If any issues arise, use find/replace to change `//+` to `// +`.

Motivates by: https://github.com/kubernetes-sigs/kubebuilder/discussions/3801

c/c @mateusoliveira43

PR #3960, opened 2024-05-27
:sparkles: Add revive golangci linter with comment-spacings rule

mateusoliveira43:
## Why changes were made

Follow up PR as discussed here https://github.com/kubernetes-sigs/kubebuilder/pull/3904#issuecomment-2119929172

## How changes were made

Added rule to check comment spaces in revive linter, in golangci-lint config files; fixed comments spaces; and then run `make generate`.

## How to test changes made

Check that all Go files now enforce comment spaces.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -61,6 +61,7 @@ linters-settings:
       #
       - name: bool-literal-in-expr
       - name: constant-logical-expr
+      - name: comment-spacings
 
 linters:
   disable-all: true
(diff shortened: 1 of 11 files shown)


camilamacedo86 (commented):
Hi @mateusoliveira43 

See that I change the title.
this change is user facing
and we are adding a new linter in the projects scaffold with the tool 
So, we need to clarify that.

camilamacedo86 (approved):
/lgtm

Issue #3657, opened 2023-10-11
Improve scaffolding of ServiceMonitor


fgiloux:
### What do you want to happen?

Currently the scaffolding of [ServiceMonitor](https://github.com/kubernetes-sigs/kubebuilder/blob/6fcae44a7078777bbb644514f94f35fea9bb5209/testdata/project-v4/config/prometheus/monitor.yaml#L20-L22) makes use of Prometheus in-cluster credentials for collecting metrics. In some Prometheus installation it is not allowed, i.e. the setting is ignored for security reason:
A problem with the approach is that Prometheus token is exposed to the user creating the ServiceMonitor. With this token the user would be able to query any metrics endpoint Prometheus has access to.

Another issue is that insecureSkipVerify is used, which deactivates an important mechanism to check that the server reached is what it pretends to be.

My use case is really about:
- increasing the security level of what is offered by default by Kubebuilder
- making the default configuration works with a larger estate

It is not related to a particular Kubernetes version. 
After a quick search I am not aware of another issue associated with this.

<details>

I am proposing to create an additional service account and token secret, used for scrapping the operator metrics. This can then get referenced in the ServiceMonitor resource.
In a similar way the TLS certificates generated to secure the operator endpoint can be referenced in the ServiceMonitor to guarantee the authenticity of the server to Prometheus.

~~~
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  labels:
    control-plane: controller-manager
    app.kubernetes.io/name: servicemonitor
    app.kubernetes.io/instance: controller-manager-metrics-monitor
    app.kubernetes.io/component: metrics
    app.kubernetes.io/created-by: project-v4
    app.kubernetes.io/part-of: project-v4
    app.kubernetes.io/managed-by: kustomize
  name: controller-manager-metrics-monitor
  namespace: system
spec:
  endpoints:
    - path: /metrics
      port: https
      scheme: https
      bearerTokenSecret:
        key: token
        name: prometheus-token
      tlsConfig:
        ca:
          secret:
            key: ca.crt
            name: serving-cert
            # SERVICE_NAME and SERVICE_NAMESPACE will be substituted by kustomize
        serverName: SERVICE_NAME.SERVICE_NAMESPACE.svc
  selector:
    matchLabels:
      control-plane: controller-manager
~~~

</details>

camilamacedo86:
Hi @fgiloux,

[...]

PR #3968, opened 2024-06-03
✨ Re-introduce authn/authz for metrics endpoint

joelanford:


docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
```diff
@@ -25,17 +25,11 @@ resources:
 - ../certmanager
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
 - ../prometheus
-# [METRICS] To enable the controller manager metrics service, uncomment the following line.
-#- metrics_service.yaml
+# [METRICS] Expose the controller manager metrics service.
+- metrics_service.yaml
```
camilamacedo86, line 29: Hi @joelanford 

IHMO it should not be enable by default at all.
WHY?

- Not everybody uses those metrics and care about.
- Users should be aware of the risks and take the explicit decision of the should or not expose the metrics
- Note that, we probably will see a lot of people impacted by https://github.com/kubernetes-sigs/kubebuilder/discussions/3907 only because it was enabled by default, but they did not were aware of or using it at all.

docs/book/src/cronjob-tutorial/testdata/project/config/prometheus/monitor.yaml
```diff
@@ -11,8 +11,11 @@ metadata:
 spec:
   endpoints:
     - path: /metrics
-      port: http # Ensure this is the name of the port that exposes HTTP metrics
-      scheme: http
+      port: https # Ensure this is the name of the port that exposes HTTP metrics
```
camilamacedo86, line 14: If I do not want to use the c+r implementation and instead I want to use security policies + cert-manager, I am be able to use it with this default scaffold?

docs/book/src/cronjob-tutorial/testdata/project/config/prometheus/monitor.yaml
```diff
@@ -11,8 +11,11 @@ metadata:
...
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
```
camilamacedo86, line 18: Should we really scaffold it by default?
How users will be aware of it? 

We had it before but I think it is wrong, we have indeed an issue open for that to ensure that the metrics were properly protect using cert-manager, see: https://github.com/kubernetes-sigs/kubebuilder/issues/3657

docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
joelanford: Yeah, I waffled on this. In 3.14.x and earlier, we had the following:
- The Kube RBAC proxy patch was enabled by default, which meant that the default scaffold would run a secure KRP proxy to a plaintext and localhost metrics server running in the controller process.
- The Role and Role binding that gave KRP permissions were enabled by default.
- The metrics service was enabled by default.

Because of that, I landed on the side of the fence of "turn it all on by default" to try to keep the behavior as close to the same as possible.

But I see the argument (especially in kubebuilder/v4) of turning it all off by default.

docs/book/src/cronjob-tutorial/testdata/project/config/prometheus/monitor.yaml
joelanford: Similar to my other comment, I was trying to just replicate what we had before with kube-rbac-proxy, because that seemed to be the main urgency of these changes.

Do you think we need to solve these other problems at the same time? Or do you think they could be covered separately in follow-ups?

docs/book/src/getting-started/testdata/project/cmd/main.go
```diff
@@ -98,9 +96,10 @@ func main() {
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme: scheme,
 		Metrics: metricsserver.Options{
-			BindAddress:   metricsAddr,
-			SecureServing: secureMetrics,
-			TLSOpts:       tlsOpts,
+			BindAddress:    metricsAddr,
+			FilterProvider: filters.WithAuthenticationAndAuthorization,
```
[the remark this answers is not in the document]
joelanford: IMO the two ideal states are:
- Metrics listening on localhost only
- Metrics listening via TLS with a valid cert chain and this (or a similar) filter.

Do you agree with that? If so, I think that means that the cert-manager overlay would actually be the thing that controls the flags that are passed into the controller container to enable the TLS listen address and the authn/authz filters.

docs/book/src/getting-started/testdata/project/cmd/main.go
```diff
@@ -55,16 +56,13 @@ func main() {
...
 	var enableHTTP2 bool
-	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metric endpoint binds to. "+
-		"Use the port :8080. If not set, it will be 0 in order to disable the metrics server")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8443", "The address the metric endpoint binds to. "+
+		"Use 0 in order to disable the metrics server")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	flag.BoolVar(&secureMetrics, "metrics-secure", false,
```
camilamacedo86, line 66: Should we not keep this one and change the default for true?
So, that we allow people set false if they wish to do so.

docs/book/src/getting-started/testdata/project/cmd/main.go
camilamacedo86, line 100: Could we add a comment explaining what it does?
Add the options to pass the certs commented and says that for production it is recommended use real certs instead  and/or use cert-manager?

docs/book/src/getting-started/testdata/project/config/default/kustomization.yaml
```diff
@@ -25,17 +25,11 @@ resources:
...
+- metrics_service.yaml
 
 # Uncomment the patches line if you enable Metrics, and/or are using webhooks and cert-manager
 #patches:
-# [METRICS] The following patch will enable the metrics endpoint. Ensure that you also protect this endpoint.
-# More info: https://book.kubebuilder.io/reference/metrics
-# If you want to expose the metric endpoint of your controller-manager uncomment the following line.
-#- path: manager_metrics_patch.yaml
-#  target:
-#    kind: Deployment
```
camilamacedo86, line 38: Could we not keep this one so that we allow people disable if they wish?
We can let enable with `:8443`

docs/book/src/getting-started/testdata/project/cmd/main.go
```diff
@@ -55,16 +56,13 @@ func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
 	var probeAddr string
-	var secureMetrics bool
 	var enableHTTP2 bool
-	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metric endpoint binds to. "+
-		"Use the port :8080. If not set, it will be 0 in order to disable the metrics server")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8443", "The address the metric endpoint binds to. "+
+		"Use 0 in order to disable the metrics server")
```
camilamacedo86, line 61:
```suggestion
flag.StringVar(&metricsAddr, "metrics-bind-address", "0",  
	"The address the metrics endpoint binds to. " +
	"Use :8443 for HTTPS with metrics-secure set to true, " +
	"Use :8080 for HTTP with metrics-secure set to false, " +
	"or leave as 0 to disable the metrics service.")
```

How about we keep the metrics service  as 0/disabled as default set? We can set up the kustomize scaffold to enable it, but leave it uncommented so that we have the metrics enabled but we give a fair chance for users easily disable/opt out  if they wish to do so.

test/e2e/utils/test_context.go
```diff
@@ -166,23 +166,6 @@ func (t *TestContext) UninstallPrometheusOperManager() {
...
-	}
-	cmd := exec.Command(kustomizePath, "build", dir)
-	output, err := t.Run(cmd)
-	if err != nil {
-		warnError(err)
-	}
-	if _, err := t.Kubectl.WithInput(string(output)).Command("delete", "-f", "-"); err != nil {
-		warnError(err)
-	}
-}
```
camilamacedo86, line 184: 👍 good catcher

test/e2e/v4/generate_test.go
```diff
@@ -63,52 +63,7 @@ func GenerateV4(kbc *utils.TestContext) {
...
-		certManagerTarget, "#")).To(Succeed())
-
-	if kbc.IsRestricted {
-		By("uncomment kustomize files to ensure that pods are restricted")
-		uncommentPodStandards(kbc)
-	}
-}
-
-// GenerateV4WithoutMetrics implements a go/v4 plugin project defined by a TestContext.
-func GenerateV4WithoutMetrics(kbc *utils.TestContext) {
```
camilamacedo86, line 83: I think would be great we keep a test to ensure that the metrics are not exposed and we have an option to opt-out from it that works.

docs/book/src/getting-started/testdata/project/cmd/main.go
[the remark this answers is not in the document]
camilamacedo86: Yes, I totally agree. 
I think the next step would be add an option to patch the cert-manager config as well

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go
```diff
@@ -70,17 +70,11 @@ resources:
 #- ../certmanager
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
 #- ../prometheus
-# [METRICS] To enable the controller manager metrics service, uncomment the following line.
-#- metrics_service.yaml
+# [METRICS] Expose the controller manager metrics service.
```
Kavinjsir, line 73: Nit:
We can probably remind users of role binding for scraping, such as:

```yaml
# To scrape metrics e.g. via Prometheus the client needs the ClusterRole `metrics-reader`
```

Kavinjsir (approved):
Looks good to me!

camilamacedo86:
Hi @joelanford 

AS we spoke I am closing this one in favor of https://github.com/kubernetes-sigs/kubebuilder/pull/4003
which has your changes rebased with the master and some improvements on top.

Again, thank you a lot for your contribution and amazing help.!!!

PR #3989, opened 2024-06-19
✨ Update Makefile and Enhance `test-e2e`

hiddenmarten:
This PR includes the following changes:

1. **Makefile Updates:**
   - Introduce `kind` and `yq` as tools to the Makefile for better management and automation
   - Create `kind-*` targets that helpful during development process

2. **`test-e2e` Rework:**
   - Enhance the `test-e2e` target by extending its functionality. Instead of just building and uploading the image to the cluster, it now includes a call to the readiness endpoint of the operator-controller as part of the tests.

---

This update improves our testing process by ensuring that the operator-controller is fully ready and operational after deployment.

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -140,9 +140,9 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	@echo "Check for kubernetes version $(K8S_VERSION_TRIMMED_V) in $(ENVTEST)"
-	@$(ENVTEST) list | grep -q $(K8S_VERSION_TRIMMED_V)
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(K8S_VERSION_TRIMMED_V) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(K8S_VERSION_TRIMMED_V) --bin-dir $(LOCALBIN) -p path)" && \
+		test $? && \
+		go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
 
 # Utilize Kind or modify the e2e tests to load the image locally, enabling compatibility with other vendors.
 .PHONY: test-e2e  # Run the e2e tests against a Kind k8s instance that is spun up.
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -221,21 +221,21 @@ install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~
 
 .PHONY: uninstall
 uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
-	$(KUSTOMIZE) build config/crd | $(KUBECTL) delete -n $(NAMESPACE) --ignore-not-found=$(ignore-not-found) -f -
+	$(KUSTOMIZE) build config/crd | $(KUBECTL) delete --namespace $(NAMESPACE) --ignore-not-found=$(ignore-not-found) -f -
 
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | $(KUBECTL) -n $(NAMESPACE) apply -f -
+	$(KUSTOMIZE) build config/default | $(KUBECTL) --namespace $(NAMESPACE) apply -f -
 	$(KUBECTL) wait deployment.apps/{{ .ProjectName }}-controller-manager --for condition=Available --namespace $(NAMESPACE) --timeout 5m
 
 .PHONY: undeploy
 undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
-	$(KUSTOMIZE) build config/default | $(KUBECTL) delete -n $(NAMESPACE) --ignore-not-found=$(ignore-not-found) -f -
+	$(KUSTOMIZE) build config/default | $(KUBECTL) delete --namespace $(NAMESPACE) --ignore-not-found=$(ignore-not-found) -f -
 
 .PHONY: redeploy
 redeploy: deploy ## Redeploy controller with new docker image.
-	$(KUBECTL) rollout restart -n $(NAMESPACE) deploy/{{ .ProjectName }}-controller-manager
+	$(KUBECTL) rollout restart --namespace $(NAMESPACE) deploy/{{ .ProjectName }}-controller-manager
 
 .PHONY: kind-load
 kind-load: docker-build kind ## Build and upload docker image to the local Kind cluster.
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -73,10 +73,11 @@ func (f *Makefile) SetTemplateDefaults() error {
 //nolint:lll
 const makefileTemplate = `# Image URL to use all building/pushing image targets
 IMG ?= {{ .Image }}
-# K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-# It also refers to the version of Kubernetes that is deployed by Kind.
-K8S_VERSION = v1.30.0
-K8S_VERSION_TRIMMED_V = $(subst v,,$(K8S_VERSION))
+# ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
+ENVTEST_K8S_VERSION = 1.30.0
+
+# KIND_K8S_VERSION refers to the version of Kubernetes that is deployed by Kind.
+KIND_K8S_VERSION = v1.30.0
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
@@ -140,7 +141,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(K8S_VERSION_TRIMMED_V) --bin-dir $(LOCALBIN) -p path)" && \
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" && \
 		test $? && \
 		go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
 
@@ -205,7 +206,7 @@ build-installer: manifests generate kustomize ## Generate a consolidated YAML wi
 
 KIND_CLUSTER_NAME ?= {{ .ProjectName }}-kind
 NAMESPACE ?= {{ .ProjectName }}-system
-KIND_IMAGE ?= kindest/node:$(K8S_VERSION)
+KIND_IMAGE ?= kindest/node:$(KIND_K8S_VERSION)
 
 PROMETHEUS_OPERATOR_VERSION = v0.74.0
 CERT_MANAGER_VERSION = v1.15.0
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -73,8 +73,10 @@ func (f *Makefile) SetTemplateDefaults() error {
 //nolint:lll
 const makefileTemplate = `# Image URL to use all building/pushing image targets
 IMG ?= {{ .Image }}
-# ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.30.0
+# K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
+# Also it refers to the version of Kubernetes that Kind set up.
+K8S_VERSION = v1.30.0
+K8S_VERSION_TRIMMED_V = $(subst v,,$(K8S_VERSION))
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
@@ -138,7 +140,9 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
+	@echo "Check for kubernetes version $(K8S_VERSION_TRIMMED_V) in $(ENVTEST)"
+	@$(ENVTEST) list | grep -q $(K8S_VERSION_TRIMMED_V)
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(K8S_VERSION_TRIMMED_V) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
 
 # Utilize Kind or modify the e2e tests to load the image locally, enabling compatibility with other vendors.
 .PHONY: test-e2e  # Run the e2e tests against a Kind k8s instance that is spun up.
@@ -199,6 +203,12 @@ build-installer: manifests generate kustomize ## Generate a consolidated YAML wi
 
 ##@ Deployment
 
+KIND_CLUSTER_NAME ?= {{ .ProjectName }}-kind
+NAMESPACE ?= {{ .ProjectName }}-system
+
+PROMETHEUS_OPERATOR_VERSION = v0.74.0
+CERT_MANAGER_VERSION = v1.15.0
+
 ifndef ignore-not-found
   ignore-not-found = false
 endif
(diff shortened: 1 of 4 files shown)


pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -138,7 +140,9 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
+	@echo "Check for kubernetes version $(K8S_VERSION_TRIMMED_V) in $(ENVTEST)"
```
camilamacedo86, line 143: It is not true. We do not check for a k8s version in the ENVT TEST
We check for the ENV TEST version which is of the release of the tag X.
We tag ENV TEST according to the k8s version used for its binaries.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -199,6 +203,12 @@ build-installer: manifests generate kustomize ## Generate a consolidated YAML wi
 
 ##@ Deployment
 
+KIND_CLUSTER_NAME ?= {{ .ProjectName }}-kind
+NAMESPACE ?= {{ .ProjectName }}-system
```
camilamacedo86, line 207: Not everybody will use Kind so we cannot set it on the default scaffold.
Kind is usually used for development purposes but the solutions can run in many vendors such as: GKE, EKS, AKS, Openshift and etc,

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -73,8 +73,10 @@ func (f *Makefile) SetTemplateDefaults() error {
 //nolint:lll
 const makefileTemplate = `# Image URL to use all building/pushing image targets
 IMG ?= {{ .Image }}
-# ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.30.0
+# K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
+# Also it refers to the version of Kubernetes that Kind set up.
+K8S_VERSION = v1.30.0
+K8S_VERSION_TRIMMED_V = $(subst v,,$(K8S_VERSION))
```
camilamacedo86, line 79: We tag the ENV TEST version which is compatible, I do not think that we should add the k8s version here. It is not used in any place.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -209,16 +219,50 @@ install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~
...
-	$(KUSTOMIZE) build config/default | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
+	$(KUSTOMIZE) build config/default | $(KUBECTL) delete -n $(NAMESPACE) --ignore-not-found=$(ignore-not-found) -f -
+
+.PHONY: redeploy
+redeploy: deploy ## Redeploy controller with new docker image.
+	$(KUBECTL) rollout restart -n $(NAMESPACE) deploy/{{ .ProjectName }}-controller-manager
+
+.PHONY: kind-load
+kind-load: docker-build kind ## Build and upload docker image to the local Kind cluster.
+	$(KIND) load docker-image ${IMG} --name $(KIND_CLUSTER_NAME)
```
camilamacedo86, line 240: I do not think we should add targets which are specific to any vendor in the default scaffolds.
Users are able to do that if they wish, and many projects are distributed without a specific vendor.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -232,13 +276,17 @@ KUBECTL ?= kubectl
 KUSTOMIZE ?= $(LOCALBIN)/kustomize-$(KUSTOMIZE_VERSION)
 CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen-$(CONTROLLER_TOOLS_VERSION)
 ENVTEST ?= $(LOCALBIN)/setup-envtest-$(ENVTEST_VERSION)
-GOLANGCI_LINT = $(LOCALBIN)/golangci-lint-$(GOLANGCI_LINT_VERSION)
+GOLANGCI_LINT ?= $(LOCALBIN)/golangci-lint-$(GOLANGCI_LINT_VERSION)
+KIND ?= $(LOCALBIN)/kind-$(KIND_VERSION)
+YQ = $(LOCALBIN)/yq-$(YQ_VERSION)
```
camilamacedo86, line 281: What is the motivation for this change?
What the problem that you are trying to sort out here that justify add more one dependency with YQ?

pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/test.go
```diff
@@ -44,106 +44,101 @@ var TestTemplate = `{{ .Boilerplate }}
 package e2e
 
 import (
+	"context"
 	"fmt"
+	"github.com/go-resty/resty/v2"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
camilamacedo86, line 52: Nit: The imports must to be sorted out accordingly here.

pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/test.go
```diff
@@ -44,106 +44,101 @@ var TestTemplate = `{{ .Boilerplate }}
...
 const namespace = "{{ .ProjectName }}-system"
+const deploymentName = "{{ .ProjectName }}-controller-manager"
 
 var _ = Describe("controller", Ordered, func() {
 	BeforeAll(func() {
-		By("installing prometheus operator")
-		Expect(utils.InstallPrometheusOperator()).To(Succeed())
+		var err error
+		By("prepare kind environment", func() {
+			cmd := exec.Command("make", "kind-prepare")
```
camilamacedo86, line 68: Same we cannot add anything that specific for a vendor.

camilamacedo86 (changes requested):
Thank you for your suggestion.
However, we cannot add code that is valid ONLY for Kind and/or vendor specific in the default scaffold.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
hiddenmarten: I probably missed something but for me, it looks like this `echo` describes action on the very next row.

I'll try to explain what is happening here:
1) `@$(ENVTEST) list` returns a list of available Kubernetes binaries in there which looks kinda like that:
```
└──╼ $ bin/setup-envtest-release-0.18 list
(installed)  v1.30.0  linux/amd64
(available)  v1.30.0  linux/amd64
(available)  v1.29.3  linux/amd64
(available)  v1.29.1  linux/amd64
...
(available)  v1.14.1  linux/amd64
(available)  v1.13.1  linux/amd64
(available)  v1.12.3  linux/amd64
```
2) This list of available Kubernetes binaries versions goes through the pipe `|` to the next command.

3) The command `grep -q $(K8S_VERSION_TRIMMED_V)` checks if the Kubernetes version specified in the variable `$(K8S_VERSION_TRIMMED_V)` exists in the list of available versions. If the specified version is found, the command returns 0, and the test begins with this exact version using `$(ENVTEST) use $(K8S_VERSION_TRIMMED_V)`. If the specified version isn't found grep returns 1, and tests do not start.

Honestly, I can consider this behavior as "Check for Kubernetes version in envtest".

Note: During my experience developing operators, I found this message and check very helpful. It is convenient to see what was executed just before the makefile target returned an exit code of 1. This step is crucial for catching issues early, especially when using tools like `renovate` or `dependabot`.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
hiddenmarten: Looking a bit forward in my PR it is [used](https://github.com/hiddenmarten/kubebuilder/blob/4fc202a92e34bf5dc511dc593ca165854fb8623a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go#L242-L250) in creating kind cluster.

I also attempted to explain this behavior in [a comment above](https://github.com/hiddenmarten/kubebuilder/blob/4fc202a92e34bf5dc511dc593ca165854fb8623a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go#L77), but I may not have been entirely clear.

I would greatly appreciate your assistance in making the comment more understandable. Thank you!

pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/test.go
hiddenmarten: Sure, on it.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
hiddenmarten: Let's discuss vendor(kind)-)-related issue in a certain thread, to not spread this discussion over all threads, I'd suggest the [first one](https://github.com/kubernetes-sigs/kubebuilder/pull/3989#discussion_r1649181013).

pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/test.go
hiddenmarten: Let's discuss vendor(kind)-)-related issue in a certain thread, to not spread this discussion over all threads, I'd suggest the [first one](https://github.com/kubernetes-sigs/kubebuilder/pull/3989#discussion_r1649181013).

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
hiddenmarten: I need YQ to parse all containers in the developer environment, which are in JSONL format, to identify the specific containers we need to monitor. This will help ensure that the developer is working with the exact Kubernetes version specified in the Makefile.

In short words, this is exactly what happening in [kind-create](https://github.com/hiddenmarten/kubebuilder/blob/a26c3926859b21335f2bf1028bdde58a6f7f03ec/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go#L243-L251) target.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
hiddenmarten: I understand your concerns, and I agree that this is generally how things should work.

I would like to point out that kind is already included in the default scaffold, as seen [here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/test.go#L97-L99) and [here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/test/utils/utils.go#L131-L141).
Additionally, it seems that there is no direct way to upload a freshly built image to a vanilla Kubernetes cluster. This step appears essential for running end-to-end tests on various Kubernetes vendors, such as GKE, EKS, AKS, OpenShift, etc. While kind have a very simple way to do it.

Therefore, my question is: Should we remove kind entirely, even from the current scaffold, or should we continue to enhance it, considering that kind is open-source and [part of Kubernetes-sigs](https://github.com/kubernetes-sigs/kind/)?

hiddenmarten: Also, I would like to have kind maintainers to participate in this discussion if you don't mind.

Cc. @aojea @BenTheElder

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
varshaprasad96: `setup-envtest use` already does it here. It looks at the location where Kubebuilder assets are stored locally, if an available version is found it uses the downloaded assets. If not, it downloads the assets for the specified k8s version. (https://github.com/kubernetes-sigs/controller-runtime/blob/8290d13680ed6066d9789c0ed59ff604387e20da/tools/setup-envtest/main.go#L191).

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
varshaprasad96: Kind has its own benefits, and its use is definitely beneficial in operator development. 
Wearing the operator developer hat, I see the benefit of using helpers for kind in projects - like https://github.com/operator-framework/operator-controller/blob/ed4e40eaca385c8bd4d1d7e8329c9f9c76013d95/Makefile#L46-L47, where for this operator we explicitly mention to other users that we use "kind" for testing purposes and here are the steps and helpers for replicating it locally. 

That being said, on the other hand for a generic scaffolding tool like KB, the use of kind here is restricted only for e2es, and not for development process. It's upto the "operator author" to use whichever flavour of k8s cluster they want based on their application needs. 

Another alternative which I can suggest here, is instead of adding these targets in Makefile, one could create just another plugin for their own use-case where modifications to Makefile and the operator project are made that tailor the needs of their target consumer's development environment.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
hiddenmarten: I'm afraid it's not always true that running `make test` will work as expected if the Kubernetes version isn't available in `setup-envtest`. For instance, when attempting to run `make test` with an unavailable Kubernetes version, such as `1.30.1`, it proceeds but shows issues. Here's an example:

```
└──╼ $ make test
/home/hiddenmarten/Git/github/seaweedfs/seaweedfs-operator/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
/home/hiddenmarten/Git/github/seaweedfs/seaweedfs-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
unable to fetch hash for requested version: unable to find archive for 1.30.1 (linux,amd64)
KUBEBUILDER_ASSETS="" go test $(go list ./... | grep -v /e2e) -coverprofile cover.out
        github.com/seaweedfs/seaweedfs-operator/api/v1          coverage: 0.0% of statements
        github.com/seaweedfs/seaweedfs-operator/cmd             coverage: 0.0% of statements
?       github.com/seaweedfs/seaweedfs-operator/internal/controller/label       [no test files]
        github.com/seaweedfs/seaweedfs-operator/internal/controller/swadmin             coverage: 0.0% of statements
        github.com/seaweedfs/seaweedfs-operator/test/utils              coverage: 0.0% of statements
ok      github.com/seaweedfs/seaweedfs-operator/internal/controller     0.015s  coverage: 0.1% of statements

```

As a solution, we can either use the current changes or apply the new changes I've prepared:

```
.PHONY: test
	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" && \
		test $? && \
		go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
```

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
hiddenmarten: I agree, let's take that our world isn't ideal and make two variables instead of one :)

pkg/plugins/golang/v4/scaffolds/internal/templates/test/utils/utils.go
```diff
@@ -91,76 +81,99 @@ func Run(cmd *exec.Cmd) ([]byte, error) {
...
+	wd, err := os.Getwd()
+	if err != nil {
+		return wd, err
+	}
+	wd = strings.Replace(wd, "/test/e2e", "", -1)
+	return wd, nil
+}
+
+// GetConfig retrieves the Kubernetes configuration file.
+func GetConfig() (*rest.Config, error) {
```
mateusoliveira43, line 95: Maybe we can use some already existing function for this job? I use this in one project https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/config#GetConfigOrDie

hiddenmarten: Thank you for your suggestion, I've applied it. :)

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
mateusoliveira43: I saw some changes in E2E files and really like them, for using Go instead of shell. Since my project uses `oc` instead of `kubectl`, I could not run E2E as it was

For my use case, one thing I would like to see change in E2E is that it does not run any shell or make command or install another tool. What we do in our project is to force developers (and CI tool) to configure a cluster (kubernetes, kind, or another) prior to running E2E. Then, E2E only run tests for the controllers and CRDs

Changing E2E to not install or configure a test cluster or use binaries like kubectl, would be a more general solution?

hiddenmarten: I understand, and based on this, I have [made changes](https://github.com/kubernetes-sigs/kubebuilder/pull/3989/commits/a241ea851baba3edab8f4e67f9349116b278acb8) to separate the logic of preparing the Kubernetes environment from creating a kind cluster. This adjustment simplifies the process of reworking e2e tests to use any other Kubernetes distribution. Now, it only requires substituting two steps in the BeforeAll section, making it easier than before.

However, I'm unclear on how the kind-related targets in the Makefile impact the preferences of the "operator author" in the scope of flavors of k8s. Could you please clarify this?

hiddenmarten: @mateusoliveira43 

> Changing E2E to not install or configure a test cluster or use binaries like kubectl, would be a more general solution?

I see this as potentially being a feature flag for using kind in autotests, enabled by default. Unfortunately, E2E tests assume the presence of a Kubernetes cluster, and I haven't found any solution more suitable than kind for this purpose.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
camilamacedo86: What you are checking in the coverage and not the tests,
I think for this specific scenario it would be better create an specific issue if you see that is required so that we can analyse and check what problems the current behaviour cause and how better we can improve it.

Above it shows that the error was faced see:
> unable to fetch hash for requested version: unable to find archive for 1.30.1 (linux,amd64)

But the result of the coverage was checked.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
camilamacedo86: Hi @hiddenmarten,

> I would like to point out that kind is already included in the default scaffold, as seen [here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/test.go#L97-L99) and [here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/test/utils/utils.go#L131-L141).

The current scaffolding for end-to-end (e2e) tests in the project primarily serves as examples of how to execute these tests, typically leveraging GitHub Actions and Kind. **It includes ONLY a single line** for loading images because we could NOT find a generic approach for the case.However, ideally if you have a suggestion that can be generic then that YES, it would be really desirable. Also, we can improve it to clarify by adding a comment to authors.

Regarding your suggestion to add multiple Kind-specific targets to the project's Makefile, it's important to consider the project's stance on vendor specificity. As it stands, the project aims to avoid dependencies that tie it to any particular vendor, which includes Kind. Therefore, any enhancements should ideally offer a more generic approach that aligns with this philosophy.

camilamacedo86: >However, I'm unclear on how the kind-related targets in the Makefile impact the preferences of the "operator author" in the scope of flavors of k8s. Could you please clarify this?

Add vendor specific Makefile targets or code does not align with the goal of the project at all.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
camilamacedo86: Hi @hiddenmarten 

I understand that you needed `YQ` here. 
However, we avoid as max as possible add dependencies to the project.
Unless we discussed and define that the dependency is really required to address a need that brings a lot of value for the users we do not add it. 

The proposed solution does not a huge value for the users, so has no reason for we add +1 dependency here.

Issue #3871, opened 2024-04-23
[META - Phases]: Discontinue Kube RBAC Proxy in Default Kubebuilder Scaffolding


camilamacedo86:
### What do you want to happen?

Address the phases and changes discussed in the proposal: https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/discontinue_usage_of_kube_rbac_proxy.md

- [x] - Phase 0: Remove the RBAC Proxy: PR: https://github.com/kubernetes-sigs/kubebuilder/pull/3899
- [x] - Phase 1: Transition to NetworkPolicies: PR: https://github.com/kubernetes-sigs/kubebuilder/pull/3853
- [x] - Phase 2: Add Cert-Manager as Optional option to be used with metrics **(help wanted)** PS.: See that we should address the request made in: https://github.com/kubernetes-sigs/kubebuilder/issues/3657 within this one. Done at: https://github.com/kubernetes-sigs/kubebuilder/pull/4243
- [x] - Phase 3: Add Controller-Runtime feature. Details can be found in: https://kubernetes.slack.com/archives/CAR30FCJZ/p1716290359950339?thread_ts=1716289540.582639&cid=CAR30FCJZ  -> DONE: PR: https://github.com/kubernetes-sigs/kubebuilder/pull/4003
- [ ] - Phase 4: When kube-rbac-proxy be accepted under the umbrella (Blocked by: https://github.com/brancz/kube-rbac-proxy/issues/238 and the project be part of the umbrella. When it happens we can reach out its maintainers and see if we can support them out to create the external plugin)

fgiloux:
Hi Camilla. Nice to see progress on this. If I am understanding things right you are considering two different ways of doing RBAC:
- the new controller-runtime feature #3657, which has merged
- kube-rbac proxy
You also have two orthogonal subjects:
 - TLS certificates for the metrics endpoint, where you propose to optionally scaffold resources for cert-manager
 - scaffolding an SA, and referencing it in the ServiceMonitors, that can be leveraged by Prometheus for scrapping the metrics
 This sounds good to me.

fgiloux:
For phase 3: Maybe you mean [issue 2781](https://github.com/kubernetes-sigs/controller-runtime/issues/2781) as blocker?

camilamacedo86:
Hi @fgiloux,

> For phase 3: Maybe you mean https://github.com/kubernetes-sigs/controller-runtime/issues/2781 as blocker?

[...]

PR #4003, opened 2024-07-04
✨ Add protection to metrics endpoint using authn/authz via controller-runtime feature

camilamacedo86:
This PR re-introduce authn/authz protection for the endpoint but without use the kube-rbac-proxy project and image.
Address item 3 mapped in : https://github.com/kubernetes-sigs/kubebuilder/issues/3871

Author: Joe Lanford <joe.lanford@gmail.com>
Co-Author: Camila Macedo


pkg/plugins/golang/v4/scaffolds/internal/templates/main.go
```diff
@@ -262,10 +266,25 @@ func main() {
...
+			// production environments as self-signed certificates do not offer the same level of trust and security 
+			// as certificates issued by a trusted Certificate Authority (CA). The primary risk is potentially allowing 
+			// unauthorized access to sensitive metrics data. Consider replacing with CertDir, CertName, and KeyName 
+			// to provide certificates, ensuring the server communicates using trusted and secure certificates.			
+			TLSOpts: tlsOpts, 
+			// FilterProvider is used to protect the metrics endpoint with authn/authz.
+			// These configurations ensure that only authorized users and service accounts
+			// can access the metrics endpoint. The RBAC are configured in 'config/rbac/kustomization.yaml'. More info:
+			// https://pkg.go.dev/sigs.k8s.io/controller-runtime@{{ .ControllerRuntimeVersion }}/pkg/metrics/filters#WithAuthenticationAndAuthorization
+			FilterProvider: filters.WithAuthenticationAndAuthorization,
```
Kavinjsir, line 287: Nits:
Similar to the concern in the previous PR, would it be necessary/good to remind user of assigning the cluster role to the clients with the permission to scrape the metrics?
See the [ref](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.4/pkg/metrics/filters#WithAuthenticationAndAuthorization):
> To scrape metrics e.g. via Prometheus the client needs a ClusterRole with the following rule: * nonResourceURLs: "/metrics", verbs: get

Kavinjsir (approved):
lgtm 👍🏼

varshaprasad96 (approved):
I did a quick check by checking out the branch and running the example locally. It does work from my end. 

Thank you for the contribution!

/lgtm
/approve

pkg/plugins/golang/v4/scaffolds/internal/templates/main.go
camilamacedo86: I think that the problem is that depends on how the Prometheus is configured.
We might could say something like Prometheus requires access to the metrics endpoint.
But we are either linking the metrics doc so users can check that, there we are clarifying it right.

Anyway, lets to do one thing: 
Since it is a nit, lets move forward with as it is, however, if you see any good way to improve the comments/docs etc, please feel free to push a PR with your idea and we can shape things around.

docs/book/src/cronjob-tutorial/testdata/project/cmd/main.go
```diff
@@ -116,10 +117,25 @@ func main() {
...
+			// production environments as self-signed certificates do not offer the same level of trust and security
+			// as certificates issued by a trusted Certificate Authority (CA). The primary risk is potentially allowing
+			// unauthorized access to sensitive metrics data. Consider replacing with CertDir, CertName, and KeyName
+			// to provide certificates, ensuring the server communicates using trusted and secure certificates.
+			TLSOpts: tlsOpts,
+			// FilterProvider is used to protect the metrics endpoint with authn/authz.
+			// These configurations ensure that only authorized users and service accounts
+			// can access the metrics endpoint. The RBAC are configured in 'config/rbac/kustomization.yaml'. More info:
+			// https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.4/pkg/metrics/filters#WithAuthenticationAndAuthorization
+			FilterProvider: filters.WithAuthenticationAndAuthorization,
```
sbueringer, line 138: I think this should only be set if secureMetrics is true.

By always enabling this folks have to always provide a ServiceAccount token. But these tokens should never be send via HTTP

camilamacedo86: we can address this one good catcher.

Issue #4009, opened 2024-07-10
update webhook marker docs


davidxia:
### What broke? What's expected?

The [webhook markers doc](https://book.kubebuilder.io/reference/markers/webhook) doesn't have `url` even though recent versions of kubebuilder supports it. Does this page need to be refreshed or regenerated from the Golang source in some way?

### Reproducing this issue

Latest kubebuilder release 4.1.0 [uses controller-tools 0.15.0](https://github.com/kubernetes-sigs/kubebuilder/blob/release-4.1/pkg/plugins/golang/v4/scaffolds/init.go#L40) which [supports the `path` marker](https://github.com/kubernetes-sigs/controller-tools/blob/release-0.15/pkg/webhook/parser.go#L135). But the webhook marker doc page doesn't mention it.

### KubeBuilder (CLI) Version

4.1.0

### PROJECT version

3

### Plugin versions

```yaml
go.kubebuilder.io/v4
```

### Other versions

go 1.22.2

### Extra Labels

/kind documentation

camilamacedo86:
Hi @davidxia 

Thank you for raising this one. Just to share:

- The page is defined in : https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/reference/markers/webhook.md?plain=1#L3-L7
- See that we have `{{#markerdocs Webhook}}` <- which generates the markers
- Here is the implementation for the Markerdocs: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/utils/markerdocs/main.go
- Here is the script which calls this one: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/markerdocs.sh
- Then, you can check that is called in https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/book.toml#L17

If someone have a time to look on that please feel free.

camilamacedo86:
Acctualy, I think that I found already.
This PR should sort it out: https://github.com/kubernetes-sigs/kubebuilder/pull/4010
We will need to update the book-v3 to make the result appears in prod.

PR #4010, opened 2024-07-10
📖 Upgrade controller-gen used to generated the docs and remove redirects for artefact images 

camilamacedo86:
This PR update the deps used to generate the data in the docs.
It should update the markers to address https://github.com/kubernetes-sigs/kubebuilder/issues/4009

Furthermore, for we are able to move forward here we need to remove the redirect for `https://go.kubebuilder.io/images` and `gcr.io/kubebuilder/thirdparty` which are deprecated and/or no longer used.

We are either adding a note in the in the Artefacts page to clarify that ENV TEST binaries are no longer produced by Kubebuilder and its build and promotion was moved to controller-gen. 

Close: https://github.com/kubernetes-sigs/kubebuilder/issues/4009

### DIFF
--- a/docs/book/install-and-build.sh
+++ b/docs/book/install-and-build.sh
@@ -71,7 +71,7 @@ chmod +x /tmp/mdbook
 
 echo "grabbing the latest released controller-gen"
 go version
-go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.12.1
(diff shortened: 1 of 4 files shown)


davidxia (approved):
thanks! rubber-stamp approval as a user

sbueringer:
Nice!

Thank you!

lgtm

PR #4013, opened 2024-07-11
✨ Makefile: Create Version-Agnostic Symlinks for Specified Versioned Binaries

schrej:
While the new approach (well, new since I last used kubebuilder) finally does proper upgrades of tools, which is very nice, it makes it much harder to use the versioned tools in scripts that aren't called by the makefile, e.g. Tilt. (https://tilt.dev/)

This MR introduces symlinks to the latest version of the tools, which are updated whenever a new version is downloaded.

The Makefile also uses those symlinks to access the tools, which makes the `go-install-tool` function a bit simpler, getting rid of the sed replace.

## Details

This change does not remove the explicit versioning, and the versioned binaries will still exists in the bin folder. There will just be an additional symlink to the version that was installed last (the symlink is only created during installation, which could be changed to support downgrades to already installed versions).
image

### Use cases:

It is helpful if you write scripts that use the binaries downloaded by the makefile. You can just use e.g. `bin/kustomize` and there is no need to adapt it to the version currently set in the makefile, and also no need to execute the script with the makefile to pass in the binary to use.
The only relevant one probably is kustomize, but it should be the same behaviour for all of them.

Another use-case can be to point your IDE to your `bin/golangci-lint` binary to always use the version specified by the project.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/Makefile
+++ b/docs/book/src/cronjob-tutorial/testdata/project/Makefile
@@ -152,10 +152,10 @@ $(LOCALBIN):
 
 ## Tool Binaries
 KUBECTL ?= kubectl
-KUSTOMIZE ?= $(LOCALBIN)/kustomize-$(KUSTOMIZE_VERSION)
-CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen-$(CONTROLLER_TOOLS_VERSION)
-ENVTEST ?= $(LOCALBIN)/setup-envtest-$(ENVTEST_VERSION)
-GOLANGCI_LINT = $(LOCALBIN)/golangci-lint-$(GOLANGCI_LINT_VERSION)
+KUSTOMIZE ?= $(LOCALBIN)/kustomize
+CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
+ENVTEST ?= $(LOCALBIN)/setup-envtest
+GOLANGCI_LINT = $(LOCALBIN)/golangci-lint
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.4.1
@@ -181,18 +181,20 @@ $(ENVTEST): $(LOCALBIN)
 .PHONY: golangci-lint
 golangci-lint: $(GOLANGCI_LINT) ## Download golangci-lint locally if necessary.
 $(GOLANGCI_LINT): $(LOCALBIN)
-	$(call go-install-tool,$(GOLANGCI_LINT),github.com/golangci/golangci-lint/cmd/golangci-lint,${GOLANGCI_LINT_VERSION})
+	$(call go-install-tool,$(GOLANGCI_LINT),github.com/golangci/golangci-lint/cmd/golangci-lint,$(GOLANGCI_LINT_VERSION))
 
 # go-install-tool will 'go install' any package with custom target and name of binary, if it doesn't exist
-# $1 - target path with name of binary (ideally with version)
+# $1 - target path with name of binary
 # $2 - package url which can be installed
 # $3 - specific version of package
 define go-install-tool
-@[ -f $(1) ] || { \
+@[ -f "$(1)-$(3)" ] || { \
 set -e; \
 package=$(2)@$(3) ;\
 echo "Downloading $${package}" ;\
+rm -f $(1) || true ;\
 GOBIN=$(LOCALBIN) go install $${package} ;\
-mv "$$(echo "$(1)" | sed "s/-$(3)$$//")" $(1) ;\
-}
+mv $(1) $(1)-$(3) ;\
+} ;\
+ln -sf $(1)-$(3) $(1)
 endef
--- a/docs/book/src/getting-started/testdata/project/Makefile
+++ b/docs/book/src/getting-started/testdata/project/Makefile
@@ -152,10 +152,10 @@ $(LOCALBIN):
 
 ## Tool Binaries
 KUBECTL ?= kubectl
-KUSTOMIZE ?= $(LOCALBIN)/kustomize-$(KUSTOMIZE_VERSION)
-CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen-$(CONTROLLER_TOOLS_VERSION)
-ENVTEST ?= $(LOCALBIN)/setup-envtest-$(ENVTEST_VERSION)
-GOLANGCI_LINT = $(LOCALBIN)/golangci-lint-$(GOLANGCI_LINT_VERSION)
+KUSTOMIZE ?= $(LOCALBIN)/kustomize
+CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
+ENVTEST ?= $(LOCALBIN)/setup-envtest
+GOLANGCI_LINT = $(LOCALBIN)/golangci-lint
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.4.1
@@ -181,18 +181,20 @@ $(ENVTEST): $(LOCALBIN)
 .PHONY: golangci-lint
 golangci-lint: $(GOLANGCI_LINT) ## Download golangci-lint locally if necessary.
 $(GOLANGCI_LINT): $(LOCALBIN)
-	$(call go-install-tool,$(GOLANGCI_LINT),github.com/golangci/golangci-lint/cmd/golangci-lint,${GOLANGCI_LINT_VERSION})
+	$(call go-install-tool,$(GOLANGCI_LINT),github.com/golangci/golangci-lint/cmd/golangci-lint,$(GOLANGCI_LINT_VERSION))
 
 # go-install-tool will 'go install' any package with custom target and name of binary, if it doesn't exist
-# $1 - target path with name of binary (ideally with version)
+# $1 - target path with name of binary
 # $2 - package url which can be installed
 # $3 - specific version of package
 define go-install-tool
-@[ -f $(1) ] || { \
+@[ -f "$(1)-$(3)" ] || { \
 set -e; \
 package=$(2)@$(3) ;\
 echo "Downloading $${package}" ;\
+rm -f $(1) || true ;\
 GOBIN=$(LOCALBIN) go install $${package} ;\
-mv "$$(echo "$(1)" | sed "s/-$(3)$$//")" $(1) ;\
-}
+mv $(1) $(1)-$(3) ;\
+} ;\
+ln -sf $(1)-$(3) $(1)
 endef
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -229,10 +229,10 @@ $(LOCALBIN):
 
 ## Tool Binaries
 KUBECTL ?= kubectl
-KUSTOMIZE ?= $(LOCALBIN)/kustomize-$(KUSTOMIZE_VERSION)
-CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen-$(CONTROLLER_TOOLS_VERSION)
-ENVTEST ?= $(LOCALBIN)/setup-envtest-$(ENVTEST_VERSION)
-GOLANGCI_LINT = $(LOCALBIN)/golangci-lint-$(GOLANGCI_LINT_VERSION)
+KUSTOMIZE ?= $(LOCALBIN)/kustomize
+CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
+ENVTEST ?= $(LOCALBIN)/setup-envtest
+GOLANGCI_LINT = $(LOCALBIN)/golangci-lint
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= {{ .KustomizeVersion }}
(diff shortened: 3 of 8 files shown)


pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -258,19 +258,20 @@ $(ENVTEST): $(LOCALBIN)
...
 # $2 - package url which can be installed
 # $3 - specific version of package
 define go-install-tool
-@[ -f $(1) ] || { \
+@[ -f "$(1)-$(3)" ] || { \
 set -e; \
 package=$(2)@$(3) ;\
 echo "Downloading $${package}" ;\
 GOBIN=$(LOCALBIN) go install $${package} ;\
-mv "$$(echo "$(1)" | sed "s/-$(3)$$//")" $(1) ;\
```
schrej, line 273: In theory we could also add a `rm $(1)-*` here to clean up old versions. Some risk of deleting other binaries if they have the same prefix though. E.g. deleting `gen-something-*` and having another `gen-something-else` binary.

camilamacedo86 (dismissed):
Sorry, I will try to understand better your motivation here.

docs/book/src/getting-started/testdata/project/Makefile
```diff
@@ -152,10 +152,10 @@ $(LOCALBIN):
 
 ## Tool Binaries
 KUBECTL ?= kubectl
-KUSTOMIZE ?= $(LOCALBIN)/kustomize-$(KUSTOMIZE_VERSION)
-CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen-$(CONTROLLER_TOOLS_VERSION)
-ENVTEST ?= $(LOCALBIN)/setup-envtest-$(ENVTEST_VERSION)
-GOLANGCI_LINT = $(LOCALBIN)/golangci-lint-$(GOLANGCI_LINT_VERSION)
```
camilamacedo86, line 158: How will ensure that the specific version is the version used?
Note that under the bin we can have many versions.
How we would keep the same behaviour with your proposal?

schrej: Whenever a _new_ version is downloaded the e.g. `$(LOCALBIN)/kustomize` symlink will be changed to point to `$(LOCALBIN)/kustomize-$(KUSTOMIZE_VERSION)` instead. This doesn't allow to do downgrades at the moment though. Initially I had it always recreate the symlink, which would also allow to do downgrades.
Alternatively we could delete the old version as described in my other comment, which would also trigger a symlink update on every change.

camilamacedo86 (changes requested):
Hi @schrej 

Thank you for your contribution and for proposing this change to the Kubebuilder Makefile.

I understand that your motivation is to make it easier to use versioned tools in scripts that are not called by the Makefile, such as Tilt. (which does not seems a common usage/adoption) Introducing symlinks to the latest version of the tools and simplifying the go-install-tool function by removing the sed replace indeed has its benefits.

**However, I have some concerns:**

- **Explicit Versioning:** The current Makefile explicitly defines the tool versions, which is beneficial for transparency and version control. By checking the bin directory, we can easily see which versions are downloaded. **This explicitness aligns with Kubebuilder's goal of making things simple and clear.**

- **Customization vs. Defaults:** While your approach might be advantageous for certain use cases, it's essential to consider the broader community. Kubebuilder aims to **provide simple and explicit default configurations as much as possible**. Users with specific needs can always customize their projects, but changes to the default scaffolding should benefit the majority.

- **Community Impact:** We need to balance making improvements with maintaining simplicity and explicitness for the entire community. **Not every change will satisfy everyone**, so it's crucial to keep the broader user base in mind when making such decisions. Keep things explicity and simple help those with special needs (which seems your case) customize the things easily. 

Threfore, I appreciate your efforts to improve our codebase However, given the goals and the need to cater to the wider community, we should carefully consider such changes. IHMO that only make harder for users understand and symlinks came with many caveats, so its hard for me see any appetize for this proposal. 

**Again, thank you very much for your help. New ideas and suggestions are always welcome.** 🎉 

**So, my question is: Do you think this change would make things better for the majority of users? Would it make things simpler to use and understand, or would it have the opposite effect?**

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
camilamacedo86: I do not think that we should remove anything without an explicit request, OK, or demand from a user.
It is not a good practice/ux delete staff without the user knowledge.

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -152,10 +152,10 @@ $(LOCALBIN):
...
 ## Tool Binaries
 KUBECTL ?= kubectl
-KUSTOMIZE ?= $(LOCALBIN)/kustomize-$(KUSTOMIZE_VERSION)
-CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen-$(CONTROLLER_TOOLS_VERSION)
-ENVTEST ?= $(LOCALBIN)/setup-envtest-$(ENVTEST_VERSION)
-GOLANGCI_LINT = $(LOCALBIN)/golangci-lint-$(GOLANGCI_LINT_VERSION)
+KUSTOMIZE ?= $(LOCALBIN)/kustomize
+CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
+ENVTEST ?= $(LOCALBIN)/setup-envtest
+GOLANGCI_LINT = $(LOCALBIN)/golangci-lint
```
camilamacedo86, line 158: While this change might be helpful for the way you are using Tilt, it's not clear how many other users would benefit from this adjustment. Could you provide more context or examples of how this change would improve the experience for the majority of users?

Also, what is Tilt? Who uses it? When it is used?

docs/book/src/getting-started/testdata/project/Makefile
camilamacedo86: So, how can this be helpful for a developer? How can this be better than explicitly setting the version and checking things out? If something breaks, you can simply revert and validate. This way, it is clear what version is being used.

schrej: Maybe there is a slight misunderstanding here: I like the explicit versioning very much, and its much improved as the current version will actually update when you change the version without requiring to remove the old binaries first.
This change does not remove the explicit versioning, and the versioned binaries will still exists in the bin folder. There will just be an additional symlink to the version that was installed last (the symlink is only created during installation, which could be changed to support downgrades to already installed versions).
image

It is helpful if you write scripts that use the binaries downloaded by the makefile. You can just use e.g. `bin/kustomize` and there is no need to adapt it to the version currently set in the makefile, and also no need to execute the script with the makefile to pass in the binary to use.
The only relevant one probably is kustomize, but it should be the same behaviour for all of them.

docs/book/src/cronjob-tutorial/testdata/project/Makefile
schrej: Tilt is a tool that helps to run the development cycle of an application in a kubernetes cluster. It's _very_ useful when developing operators/controllers since it makes it very easy to run your app in a cluster for testing. Especially webhooks can be debugged a lot easier using it. https://tilt.dev

A common use-case with tilt is to run kustomize to build the manifests from the config folder and then use them for the tilt deployment.

cluster-api is a very heavy user of tilt for example. They actually do the same thing, but with an external script that's called from the Makefile: https://github.com/kubernetes-sigs/cluster-api/blob/main/scripts/go_install.sh#L45

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -152,10 +152,10 @@ $(LOCALBIN):
```
camilamacedo86, line 152: The right emoji is `:sparkles:` since it change the scaffold provided for the users
🌱  <- Usually we use for recipes or changes that are not user face

More info: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md

So, I hope that you do not mind but I changed the title emoji

schrej:
The last two pushes should fix upgrades failing due to the binary file already existing (or rather the symlink to it) and it will also ensure the correct binary is used when downgrading by replacing the symlink every time.

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -181,18 +181,20 @@ $(ENVTEST): $(LOCALBIN)
...
 # $3 - specific version of package
 define go-install-tool
-@[ -f $(1) ] || { \
+@[ -f "$(1)-$(3)" ] || { \
 set -e; \
 package=$(2)@$(3) ;\
 echo "Downloading $${package}" ;\
+rm -f $(1) || true ;\
 GOBIN=$(LOCALBIN) go install $${package} ;\
-mv "$$(echo "$(1)" | sed "s/-$(3)$$//")" $(1) ;\
```
camilamacedo86, line 196: I like the idea of remove `sed` since it does not work always at the same way in linux and mac

docs/book/src/getting-started/testdata/project/Makefile
```diff
@@ -181,18 +181,20 @@ $(ENVTEST): $(LOCALBIN)
...
+# $1 - target path with name of binary
 # $2 - package url which can be installed
 # $3 - specific version of package
 define go-install-tool
-@[ -f $(1) ] || { \
+@[ -f "$(1)-$(3)" ] || { \
 set -e; \
 package=$(2)@$(3) ;\
 echo "Downloading $${package}" ;\
+rm -f $(1) || true ;\
```
camilamacedo86, line 195: What are we removing here? The symbolic link or binary?
If is the symbolic link 👍

docs/book/src/getting-started/testdata/project/Makefile
```diff
@@ -152,10 +152,10 @@ $(LOCALBIN):
...
 ## Tool Binaries
 KUBECTL ?= kubectl
-KUSTOMIZE ?= $(LOCALBIN)/kustomize-$(KUSTOMIZE_VERSION)
-CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen-$(CONTROLLER_TOOLS_VERSION)
-ENVTEST ?= $(LOCALBIN)/setup-envtest-$(ENVTEST_VERSION)
-GOLANGCI_LINT = $(LOCALBIN)/golangci-lint-$(GOLANGCI_LINT_VERSION)
+KUSTOMIZE ?= $(LOCALBIN)/kustomize
+CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
+ENVTEST ?= $(LOCALBIN)/setup-envtest
+GOLANGCI_LINT = $(LOCALBIN)/golangci-lint
```
camilamacedo86, line 158: What happens if I bump 3 diff versions?
Will we keep the old 3 versions or we deleted/rewrite then?

docs/book/src/getting-started/testdata/project/Makefile
schrej: The symbolic link, otherwise the `go install` will fail.

camilamacedo86 (commented):
/ok-to-test

docs/book/src/cronjob-tutorial/testdata/project/Makefile
schrej: This could also be done without the symlink stuff, in case this doesn't make it in.

docs/book/src/getting-started/testdata/project/Makefile
camilamacedo86: But what is the result?
How we would have 3 bin/kustomize with the version in the name?

schrej: like this, maybe I misunderstood the question?
image

camilamacedo86 (commented):
Hi @schrej,

Thank you for your time and collab on this one 🥇  

**Just one question here:** https://github.com/kubernetes-sigs/kubebuilder/pull/4013/files#r1676137395

Otherwise, after the explanations I think I am OK with the changes.

docs/book/src/getting-started/testdata/project/Makefile
camilamacedo86: Yes, but the kustomize-v5.4.0 or   kustomize-v5.4.1 are the binaries downloaded once right?
And kustomize is the symlink. Am I right?

docs/book/src/cronjob-tutorial/testdata/project/Makefile
camilamacedo86: If you have an alternative approach that avoids using symlinks and you believe it would be more suitable, please feel free to suggest it. The use of symlinks might not be very clear for beginner users and could cause some confusion.

docs/book/src/getting-started/testdata/project/Makefile
schrej: `kustomize` is the symlink (as IntelliJ shows with that little arrow).
The symlink is always recreated, regardless whether the version already exists or not, so it will always point to the version in the makefile. (this was changed with the last push)

camilamacedo86: Thank you for confirm my understand.

camilamacedo86 (approved):
@schrej 

I am OK with the changes proposed.
/lgtm
/approved

If you have a better approach or suggestion feel free to raise a follow up
It seems good to fly 🎉

Issue #1778, opened 2020-11-03
Status not set correctly on Cron Job tests example


EnriqueL8:
I am trying to test a resource by creating a sub resource with a set status, as done [here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller_test.go#L170 ). Yet when I query the job as such"
```
createdJob := &batchv1.Job{}
ns := types.NamespacedName{Name: JobName, Namespace: CronjobNamespace}
err := k8sClient.Get(ctx, ns, createdJob)
fmt.Printf("%+v\n", createdJob)
```
I get no status field:
```
&Job{ObjectMeta:{      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []},Spec:JobSpec{Parallelism:nil,Completions:nil,ActiveDeadlineSeconds:nil,Selector:nil,ManualSelector:nil,Template:{{      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} {[] [] [] []  <nil> <nil>  map[]   <nil>  false false false <nil> nil []   nil  [] []  <nil> nil [] <nil> <nil> <nil> map[] []}},BackoffLimit:nil,TTLSecondsAfterFinished:nil,},Status:JobStatus{Conditions:[]JobCondition{},StartTime:<nil>,CompletionTime:<nil>,Active:0,Succeeded:0,Failed:0,},}
```

I expected the Status field to be populated with `Active: 2`. Can I not retrieve the status using `k8sClient.Get()`? Or do I need to add a sleep between the creation and fetch?

What versions of software are you using?  Specifically, the following are often useful:

* go version 1.15.2
* kubebuilder version (`kubebuilder version`) and scaffolding version (check your `PROJECT` file)
* controller-runtime version up to date with master
* controller-tools version up to date with master
* Kubernetes & kubectl versions 1.19 client

/kind bug

camilamacedo86:
Hi @EnriqueL8,
 
Are you using the project in the docs `https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata` and it is not working as expected or it is about your specific test? If it is about your specific test, are you using the client to update the object after the change? I'd like to recommend you get the object, perform the change and then use the client to update it. Then, you can fetch the object again in order to check if the value is as expected. 

[...]

PR #4016, opened 2024-07-13
📖 Update test examples not to set Status on create

hsadoyan:
Change the example tests not to attempt to set Status on Create. The kubernetes api will ignore the status field on creation, leaving it blank.  A new user trying to follow the book might be surprised to find their tests failing because of an empty status.

Instead the example will update the Status immediately after creation. 

Per conversation in the [slack](https://kubernetes.slack.com/archives/CAR30FCJZ/p1720657748942639) 

Related to #1778. Updates the book documentation to match the issue resolution.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller_test.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller_test.go
@@ -163,9 +163,6 @@ var _ = Describe("CronJob controller", func() {
 						},
 					},
 				},
-				Status: batchv1.JobStatus{
-					Active: 2,
-				},
 			}
 
 			// Note that your CronJob’s GroupVersionKind is required to set up this owner reference.
@@ -175,6 +172,13 @@ var _ = Describe("CronJob controller", func() {
 			controllerRef := metav1.NewControllerRef(createdCronjob, gvk)
 			testJob.SetOwnerReferences([]metav1.OwnerReference{*controllerRef})
 			Expect(k8sClient.Create(ctx, testJob)).Should(Succeed())
+			// Note that you can not manage the status values while creating the resource. 
+			// The status field is managed separately to reflect the current state of the resource. 
+			// Therefore, it should be updated using a PATCH or PUT operation after the resource has been created.
+			// Additionally, it is recommended to use StatusConditions to manage the status. For further information see: 
+			// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
+			testJob.Status.Active = 2
+			Expect(k8sClient.Status().Update(ctx, testJob)).Should(Succeed())
 			/*
 				Adding this Job to our test CronJob should trigger our controller’s reconciler logic.
 				After that, we can write a test that evaluates whether our controller eventually updates our CronJob’s Status field as expected!
--- a/hack/docs/internal/cronjob-tutorial/writing_tests_controller.go
+++ b/hack/docs/internal/cronjob-tutorial/writing_tests_controller.go
@@ -182,9 +182,6 @@ var _ = Describe("CronJob controller", func() {
 						},
 					},
 				},
-				Status: batchv1.JobStatus{
-					Active: 2,
-				},
 			}
 
 			// Note that your CronJob’s GroupVersionKind is required to set up this owner reference.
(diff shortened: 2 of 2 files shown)


hack/docs/internal/cronjob-tutorial/writing_tests_controller.go
```diff
@@ -194,6 +191,9 @@ var _ = Describe("CronJob controller", func() {
 			controllerRef := metav1.NewControllerRef(createdCronjob, gvk)
 			testJob.SetOwnerReferences([]metav1.OwnerReference{*controllerRef})
 			Expect(k8sClient.Create(ctx, testJob)).Should(Succeed())
+			// Note that you can't set the status while creating the resource. It needs to be updated after
```
camilamacedo86, line 194: Do we need the comment? 
I think we can leave it without the comment right

camilamacedo86: IHMO just remove the comment and then we can get this one merged

hack/docs/internal/cronjob-tutorial/writing_tests_controller.go
```diff
@@ -194,6 +191,9 @@ var _ = Describe("CronJob controller", func() {
...
+			testJob.Status.Active = 2
```
camilamacedo86, line 195: Also. we can do that in a follow up. It does not need to be in the PR we need to change this example in order to promote the good practices. Following the k8s API convention we should use status conditions, see this example: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-with-deploy-image/api/v1alpha1/busybox_types.go#L41-L52

Then, see here in the controller: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-with-deploy-image/internal/controller/busybox_controller.go#L44-L50

https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-with-deploy-image/internal/controller/busybox_controller.go#L101-L119

So, we could update the cronjob tutorial and this example to work with StatusConditions instead.

camilamacedo86: Just for you know :-)

hack/docs/internal/cronjob-tutorial/writing_tests_controller.go
hsadoyan: I wouldn't have this comment in production code, but figured it might be worth it for a tutorial example. I can remove it though

hack/docs/internal/cronjob-tutorial/writing_tests_controller.go
hsadoyan: Good point. That should probably be a followup though since it'll require a bigger refactor

hack/docs/internal/cronjob-tutorial/writing_tests_controller.go
camilamacedo86: I think if we add a comment we need to refine that with something like 

```
// Note that you should not manage the status values while creating the resource. 
// The status field is managed separately to reflect the current state of the resource. 
// Therefore, it should be updated using a PATCH or PUT operation after the resource has been created.
// Additionally, it is recommended to use StatusConditions to manage the status. For further information see: 
// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
```

hsadoyan: Pushed the new comment!

camilamacedo86 (approved):
/lgtm
/ok-to-test

docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller_test.go
```diff
@@ -29,8 +29,8 @@ import (
 	"reflect"
 	"time"
 
-    . "github.com/onsi/ginkgo/v2"
-    . "github.com/onsi/gomega"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
```
camilamacedo86, line 33: hi @hsadoyan 

See: https://github.com/kubernetes-sigs/kubebuilder/actions/runs/9996175604/job/27630173536?pr=4016
It is failing in the testdata
So, you need to:
- rebase your branch with master
- then, run make generate locally. This target will update all places and docs with the changes.

hsadoyan:
> Require changes ( rebase and run `make generate` )

Done. Should be ready to merge now

camilamacedo86 (approved):
/lgtm

PR #4026, opened 2024-07-19
📖 clarify in note that metrics are protected by default via controller-runtime feature

camilamacedo86:
Just address a nit in the docs to make clear that we are protecting the metrics endpoint from 4.1.0 by default via controller-runtime feature


camilamacedo86:
Moving forward, just a small nit in the docs for better clarify the clanges and current state.

PR #4034, opened 2024-07-21
🐛 Fix golang documentation of the method LoadImageToKindClusterWithName  scaffold in the e2e tests

fengshunli:


camilamacedo86 (approved):
/lgtm
/approved

@fengshunli 

See that we use the emojis to generate the release notes
Also, the title so I hope that you do no mind but I tried tweak it accordingly.

PR #4037, opened 2024-07-22
Remove redundant comments

fengshunli:

### DIFF
--- a/docs/book/src/multiversion-tutorial/testdata/project/cmd/main.go
+++ b/docs/book/src/multiversion-tutorial/testdata/project/cmd/main.go
@@ -64,8 +64,6 @@ func init() {
 /*
  */
 func main() {
-	/*
-	 */
 	var metricsAddr string
 	var enableLeaderElection bool
 	var probeAddr string


docs/book/src/multiversion-tutorial/testdata/project/cmd/main.go
```diff
@@ -64,8 +64,6 @@ func init() {
 /*
  */
 func main() {
-	/*
-	 */
```
camilamacedo86, line 68: Are you sure that is redundant and is not been used to render the code properly in the docs?
Did you check how this part is used in the docs?

camilamacedo86 (changes requested):
Hi @fengshunli 

Thank you a lot for your contribution 

**However, this code is not redundant.** Follow the explanation and I hope that it helps you understand how it works and is used.

See that the code source in the tutorial comes from the samples. For example: [CronJob Tutorial](https://book.kubebuilder.io/cronjob-tutorial/empty-main).

In this case, the sample that you are changing and this line of code are used in: [Multiversion Tutorial - Webhooks](https://book.kubebuilder.io/multiversion-tutorial/webhooks).

Please review the preview: [Multiversion Tutorial - Webhooks Preview](https://deploy-preview-4037--kubebuilder.netlify.app/multiversion-tutorial/webhooks) and check that your change **will break and affect the docs**.

Screenshot 2024-07-23 at 06 49 47

You can compare with the production version here: [Production - Multiversion Tutorial - Webhooks](https://book.kubebuilder.io/multiversion-tutorial/webhooks)

Screenshot 2024-07-23 at 06 51 08

So, we are closing this one as deferred. I hope you do not mind.

PR #4035, opened 2024-07-21
📖 Add crd-scope and operator-scope documentation

camilamacedo86:
This PR add a required info to let users know what is API and Manager scope as how to manage them.

Closes; https://github.com/kubernetes-sigs/kubebuilder/issues/3623

PR #4048, opened 2024-07-27
Add default parameter config

fengshunli:

### DIFF
--- a/pkg/rescaffold/migrate.go
+++ b/pkg/rescaffold/migrate.go
@@ -23,6 +23,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"github.com/spf13/afero"
+
 	"sigs.k8s.io/kubebuilder/v4/pkg/config"
 	"sigs.k8s.io/kubebuilder/v4/pkg/config/store"
 	"sigs.k8s.io/kubebuilder/v4/pkg/config/store/yaml"
@@ -285,18 +286,10 @@ func getAPIResourceFlags(resource resource.Resource) []string {
 		args = append(args, "--resource=false")
 	} else {
 		args = append(args, "--resource")
-		if resource.API.Namespaced {
-			args = append(args, "--namespaced")
-		} else {
-			args = append(args, "--namespaced=false")
-		}
+		args = append(args, fmt.Sprintf("--namespaced=%t", resource.API.Namespaced))
 	}
 
-	if resource.Controller {
-		args = append(args, "--controller")
-	} else {
-		args = append(args, "--controller=false")
-	}
+	args = append(args, fmt.Sprintf("--controller=%t", resource.Controller))
 	return args
 }
 


pkg/rescaffold/migrate.go
```diff
@@ -285,18 +286,10 @@ func getAPIResourceFlags(resource resource.Resource) []string {
 		args = append(args, "--resource=false")
 	} else {
 		args = append(args, "--resource")
-		if resource.API.Namespaced {
-			args = append(args, "--namespaced")
-		} else {
-			args = append(args, "--namespaced=false")
-		}
+		args = append(args, fmt.Sprintf("--namespaced=%t", resource.API.Namespaced))
```
camilamacedo86, line 289: Hi @fengshunli,

We appreciate your proposal to change the code.
However, we cannot move forward with this change. The Namespaced field in the specs is defined with the `omitempty` tag, see: 

https://github.com/fengshunli/kubebuilder/blame/136f6ce91fa090bcc988d40e37712bdcda8cd423/pkg/model/resource/api.go#L28-L29

This means that the field can be empty (i.e., it might not be present in the JSON representation if its value is false or not set). **As a result, the proposed change does not handle the case where resource.API.Namespaced is not present, which would be interpreted as false.**

Instead, we need to explicitly check for the presence and value of resource.API.Namespaced to ensure that the argument is correctly set. **As it is currently**

This ensures that the argument is set appropriately based on whether `resource.API.Namespaced` is true, false, or not present.

pkg/rescaffold/migrate.go
```diff
@@ -285,18 +286,10 @@ func getAPIResourceFlags(resource resource.Resource) []string {
...
 	}
 
-	if resource.Controller {
-		args = append(args, "--controller")
-	} else {
-		args = append(args, "--controller=false")
-	}
+	args = append(args, fmt.Sprintf("--controller=%t", resource.Controller))
```
camilamacedo86, line 292: The same reasoning applies to the Controller field in the Resource struct. Since the Controller field is defined with the omitempty tag, it can be omitted from the JSON representation if its value is false or not set. Therefore, using fmt.Sprintf("--controller=%t", resource.Controller) would not handle the case where resource.Controller is not present.

camilamacedo86 (changes requested):
Hi @fengshunli 

But what you are proposing here would break the current code and introduce a bug. I tried to explained why, see the comments:
- https://github.com/kubernetes-sigs/kubebuilder/pull/4048/files#r1693890024
- https://github.com/kubernetes-sigs/kubebuilder/pull/4048/files#r1693891013

I tried to fully described how it works with exampled executed by me to show it for you and I put the comment in: https://github.com/kubernetes-sigs/kubebuilder/pull/4035#issuecomment-2253896949 ( other PR that you made comments )

So, I am closing this one as deferred.

Thank you for your time and motivation to contribute with the project 🥇

PR #4054, opened 2024-08-05
📖 docs: fix typo

gabriele-wolfox:
📖 Fix typo.
"Implementeble" should be "Implementable".


designs/helm-chart-autogenerate-plugin.md
mateusoliveira43, line 1: I think PR description/title does not match modification

This adds a new design 

Typo fix would be changing typo in this file, no?
https://github.com/search?q=repo%3Akubernetes-sigs%2Fkubebuilder%20Implementeble&type=code

gabriele-wolfox: You are @mateusoliveira43 right, the typo is inside the template.

PR #4058, opened 2024-08-08
:book: Update docs support

camilamacedo86:
Refine information about compatibility and supportability of dependencies


camilamacedo86 (commented):
I am moving forward here.
No objections so far and it is more a nit following up Controller Runtime and Controller-Tools docs clarifications.

PR #4060, opened 2024-08-09
⚠️ : (go/v4): Replace usage of deprecated webhook.Validator and webhook.Defaulter interfaces

camilamacedo86:
## Description

**The goal of this PR is only address the deprecations which are mandatory needs to do otherwise, the scaffold will no longer work with new C+R versions.**

The users which request this change they would like to see the webhook implementation decouple of the APIs, however, to address this we need careful consider the pros and cons and it will be a breaking change in the Golang layout. Therefore, we probably would need a new Go version v5 and it cannot be addressed in go/v4. Those questions should be discussed in : https://github.com/kubernetes-sigs/kubebuilder/issues/4062

c/c @troy0820 ^ 
## Motivation

Controller-Runtime change the interfaces and will remove the Default and Validation ones, so this PR ensure that kubebuilder scaffold will still working and addressing the deprecations. https://github.com/kubernetes-sigs/controller-runtime/pull/2877

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/3721

c/c @nathanperkins @troy0820 @jonas-jonas

Follow up of PR: https://github.com/kubernetes-sigs/kubebuilder/pull/3723/files started by @jonas-jonas


docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -65,26 +70,42 @@ A webhook will automatically be served that calls this defaulting.
...
-// Default implements webhook.Defaulter so a webhook will be registered for the type
-func (r *CronJob) Default() {
-	cronjoblog.Info("default", "name", r.Name)
+var _ webhook.CustomDefaulter = &CronJobCustomDefaulter{}
 
-	if r.Spec.ConcurrencyPolicy == "" {
-		r.Spec.ConcurrencyPolicy = AllowConcurrent
+// Default implements webhook.CustomDefaulter so a webhook will be registered for the type
+func (d *CronJobCustomDefaulter) Default(ctx context.Context, obj runtime.Object) error {
+	cronjoblog.Info("CustomDefaulter for Admiral")
```
troy0820, line 79:
```suggestion
	cronjoblog.Info("CustomDefaulter for CronJob")
```
Is this supposed to be CronJob?  The validator below it shows CronJob

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -65,26 +70,42 @@ A webhook will automatically be served that calls this defaulting.
...
+	req, err := admission.RequestFromContext(ctx)
+	if err != nil {
+		return fmt.Errorf("expected admission.Request in ctx: %w", err)
 	}
-	if r.Spec.Suspend == nil {
-		r.Spec.Suspend = new(bool)
+	if req.Kind.Kind != "CronJob" {
+		return fmt.Errorf("expected Kind Admiral got %q", req.Kind.Kind)
```
troy0820, line 85:
```suggestion
		return fmt.Errorf("expected Kind CronJob got %q", req.Kind.Kind)
```

troy0820: Same question here.

camilamacedo86: Typo issue when I were doing

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -65,26 +70,42 @@ A webhook will automatically be served that calls this defaulting.
...
+type CronJobCustomDefaulter struct{}
 
-// Default implements webhook.Defaulter so a webhook will be registered for the type
-func (r *CronJob) Default() {
-	cronjoblog.Info("default", "name", r.Name)
+var _ webhook.CustomDefaulter = &CronJobCustomDefaulter{}
 
-	if r.Spec.ConcurrencyPolicy == "" {
-		r.Spec.ConcurrencyPolicy = AllowConcurrent
+// Default implements webhook.CustomDefaulter so a webhook will be registered for the type
```
troy0820, line 77: nit:

```suggestion
// Default implements webhook.CustomDefaulter so a webhook will be registered for the type CronJob 
```

Because the type was inferred when dealing with registering the webhook when we did this on the type, the type is now only known when registering the webhook with the manager.  

Maybe adding this on the comment will explain that relationship.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/zz_generated.deepcopy.go
```diff
@@ -52,6 +52,36 @@ func (in *CronJob) DeepCopyObject() runtime.Object {
...
+// DeepCopy is an autogenerated deepcopy function, copying the receiver, creating a new CronJobCustomValidator.
+func (in *CronJobCustomValidator) DeepCopy() *CronJobCustomValidator {
+	if in == nil {
+		return nil
+	}
+	out := new(CronJobCustomValidator)
+	in.DeepCopyInto(out)
+	return out
+}
+
```
troy0820, line 84: I'm not sure if we need to deepcopy the webhook custom default/validator.  This looks like it would still be on the api layer which we want to avoid so we can deal with the types away from having controller-runtime as a dependency.

troy0820:
I believe we want to scaffold the webhooks away from the types so we don't have controller-runtime as part of the api/v1.  

I assume we want to move this to maybe  `internal` to where we are now moving the controllers since they both depend on controller-runtime away from the api of the types.  

I understand that when scaffolding the api types we still depend on controller-runtime but most projects have moved away from that dependency itself by implementing this https://github.com/kubernetes-sigs/cluster-api/pull/9045.  This may be out of scope for this PR but would moving them to a different directory that isn't the api be a viable solution?

docs/book/src/cronjob-tutorial/testdata/project/api/v1/zz_generated.deepcopy.go
camilamacedo86: This file is from the sample used in the docs
This implementation is auto-generated by controller-gen.

camilamacedo86:
Hi @troy0820,

Thank you for your input. Following the comments inline regards https://github.com/kubernetes-sigs/kubebuilder/pull/4060#issuecomment-2278765772

> I believe we want to scaffold the webhooks away from the types so we don't have controller-runtime as part of the api/v1.

So far, we have not this intention.
The goal of this PR is `no longer use the deprecated methods of controller runtime and replace them for what is recommended by its project`

> I assume we want to move this to maybe internal to where we are now moving the controllers since they both depend on controller-runtime away from the api of the types.

We cannot change the layout or address breaking changes go/v4
Any change in the layout/breaking changes would requires a new plugin version (go/v5) and needs to be discussed and analysed carreful. We do our best to avoid ANY breaking change for our users.

However, in this case, we relay in controller-runtime and they will remove the implementation. Therefore, we have no option and that are the only case scenarios where breaking changes (without a new version of the plugin and a migration guide) is allowed since it is out of our control. 

> I understand that when scaffolding the api types we still depend on controller-runtime but most projects have moved away from that dependency itself by implementing this https://github.com/kubernetes-sigs/cluster-api/pull/9045. This may be out of scope for this PR but would moving them to a different directory that isn't the api be a viable solution?

If the community / or you want to propose that Kubebuilder no longer uses controller-runtime then, I would suggest you open an issue listing all reasons to o that. It is out of the scope of this PR. It would be a change that would affect completed Kubebuilder core and direction and needs a design proposal as well. Controller-runtime abstracts the whole complexity so I am not sure why we should no longer adopt it but I am interesting to check a list of motivations if you see that would be better for the community.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/zz_generated.deepcopy.go
camilamacedo86: I am closing this one since `types away from having controller-runtime as a dependency.` that is not the goal of this PR

camilamacedo86:
Hi @troy0820,

Thank you for your input. Following the comments inline regards https://github.com/kubernetes-sigs/kubebuilder/pull/4060#issuecomment-2278765772 

> I believe we want to scaffold the webhooks away from the types so we don't have controller-runtime as part of the api/v1.
> I assume we want to move this to maybe internal to where we are now moving the controllers since they both depend on controller-runtime away from the api of the types.

The goal of this PR is specifically to stop using the deprecated methods in controller-runtime we cannot introduce breaking changes or change the layout with further considerations. I tried to clarify in the description. Regards your request to change the layout, see: https://github.com/kubernetes-sigs/kubebuilder/issues/4062 

> I understand that when scaffolding the api types we still depend on controller-runtime but most projects have moved away from that dependency itself by implementing this https://github.com/kubernetes-sigs/cluster-api/pull/9045. This may be out of scope for this PR but would moving them to a different directory that isn't the api be a viable solution?

If the community or anyone is considering proposing that Kubebuilder should no longer use controller-runtime fully or partially, I would suggest opening an issue where all the reasons for such a change can be listed. This topic is beyond the scope of this PR, as **it would imply a significant shift in Kubebuilder's core architecture and direction**. **So we must diccuss it and such a change would require a detailed design proposal.** (https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/template.md)

Controller-runtime plays a crucial role in abstracting the complexity of controller development in Kubernetes. Therefore, it’s important to carefully evaluate the benefits and drawbacks of moving away from it. However, I’m open to reviewing a list of motivations if you believe that an alternative approach would better serve the community.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -46,6 +49,8 @@ Then, we set up the webhook with the manager.
 func (r *CronJob) SetupWebhookWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(r).
+		WithValidator(&CronJobCustomValidator{}).
```
sbueringer, line 52: I think there is no need why two separate types are needed for validator/defaulter (but also no need to just have one)

I personally would put them on the same type

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -65,26 +70,42 @@ A webhook will automatically be served that calls this defaulting.
...
+	req, err := admission.RequestFromContext(ctx)
+	if err != nil {
+		return fmt.Errorf("expected admission.Request in ctx: %w", err)
 	}
-	if r.Spec.Suspend == nil {
-		r.Spec.Suspend = new(bool)
+	if req.Kind.Kind != "CronJob" {
+		return fmt.Errorf("expected Kind CronJob got %q", req.Kind.Kind)
+	}
+	castedObj, ok := obj.(*CronJob)
```
sbueringer, line 87: I would call this cronJob

docs/book/src/cronjob-tutorial/testdata/project/api/v1/zz_generated.deepcopy.go
```diff
@@ -52,6 +52,36 @@ func (in *CronJob) DeepCopyObject() runtime.Object {
 	return nil
 }
 
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
```
sbueringer, line 55: There is no need to generate deepcopy funcs for these types (I would also imagine that to be very hard / impossible in some cases (e.g. if folks add fields like a Client)

testdata/project-v4/api/v1/admiral_webhook.go
```diff
@@ -29,18 +34,35 @@ var admirallog = logf.Log.WithName("admiral-resource")
...
-func (r *Admiral) Default() {
-	admirallog.Info("default", "name", r.Name)
+// Default implements webhook.CustomDefaulter so a webhook will be registered for the type Admiral
+func (d *AdmiralCustomDefaulter) Default(ctx context.Context, obj runtime.Object) error {
+	admirallog.Info("CustomDefaulter for Admiral")
+	req, err := admission.RequestFromContext(ctx)
+	if err != nil {
+		return fmt.Errorf("expected admission.Request in ctx: %w", err)
+	}
+	if req.Kind.Kind != "Admiral" {
```
sbueringer, line 56: This kind check seems pretty redundant. Shouldn't it be enough if the type case in l.59 works?

docs/book/src/cronjob-tutorial/testdata/project/api/v1/zz_generated.deepcopy.go
camilamacedo86: That is generated automatically via controller-gen.
We have no control about that.

testdata/project-v4/api/v1/admiral_webhook.go
camilamacedo86: I think we can remove this one 👍

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
camilamacedo86: has scenarios where this types will be properly filled?
I mean, have specs? What is the purpose of the type?

sbueringer: Not sure I understand the question. These type are only there to hold the webhook implementation. They won't have a "spec"

docs/book/src/cronjob-tutorial/testdata/project/api/v1/zz_generated.deepcopy.go
sbueringer: This should be possible with these markers: "+kubebuilder:object:generate": https://book.kubebuilder.io/reference/markers/object.html?highlight=kubebuilder%3Aobject%3Agenerate#objectdeepcopy

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
sbueringer: I'm fine either way, just sharing how we implemented it

camilamacedo86: Thank you. I will give a look !!

camilamacedo86: After think about this one further, I've reach out the idea that we need to include the resource `Kind` directly within the `CustomValidator` and `CustomDefaulter` structs by default. This is because these structs will frequently interact with the resource's fields during validation and defaulting processes. (as we are updating now the cronjob sample)

In addition to including the `Kind`, we need to maintain both structures (`CustomValidator` and `CustomDefaulter`). Each serves distinct purposes that are often necessary in practice. For example:

- **CustomDefaulter:** May contain fields that provide default values not only for the resource itself but also by referencing other resources or objects that carry default settings. Example:  We could include a `DefaultConcurrencyPolicy` that automatically applies a default value if the user doesn't specify one. Additionally, references to other resources, such as a global configuration object, could supply environment-specific defaults.
- **CustomValidator:** Often requires fields that help enforce policies or ensure that the resource meets specific criteria before it is allowed to be created, updated, or deleted. Example; Fields like `MaxConcurrentJobs` or `AllowedNamespaces` could be used to enforce cluster policies, ensuring that the resource operates within predefined limits or is only deployed in approved environments.

camilamacedo86: Done !

docs/book/src/cronjob-tutorial/testdata/project/api/v1/zz_generated.deepcopy.go
camilamacedo86: I forgot of this !!!
Good point.

camilamacedo86:
Hi @sbueringer and @troy0820 

All changes here are addressed!!
Please feel free to check it out.

I think it is good to fly now 🚀 
_PS.: We will not change the layout on this one. We do not want to force people change the layout of their project without further considerations. This part of the effort we do in a follow up were we can focus on it._

varshaprasad96 (approved):
Thanks @camilamacedo86! Just one nit, but it shouldn't be important, because tried running the PR as is and it works well. 

/approve
/lgtm

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -59,32 +69,61 @@ The meaning of each marker can be found [here](/reference/markers/webhook.md).
...
+	cronjoblog.Info("CustomDefaulter for CronJob")
+	cronjob, ok := obj.(*CronJob)
+	if !ok {
+		return fmt.Errorf("expected an CronJob object but got %T", obj)
+	}
+	cronjoblog.Info("default", "name", cronjob.GetName())
+
+	// Assign the CronJob object to the CronJobCustomDefaulter struct field
+	// so that it can be used in the defaulting logic
+	d.cronjob = cronjob
```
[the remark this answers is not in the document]
camilamacedo86: That is a good point !! 
I will change

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -59,31 +69,63 @@ The meaning of each marker can be found [here](/reference/markers/webhook.md).
...
 
-var _ webhook.Defaulter = &CronJob{}
+// +kubebuilder:object:generate=false
+// CronJobCustomDefaulter struct is responsible for setting default values on the custom resource of the
+// Kind CronJob when those are created or updated.
+//
+// NOTE: The +kubebuilder:object:generate=false marker prevents controller-gen from generating DeepCopy methods,
+// as it is used only for temporary operations and does not need to be deeply copied.
+type CronJobCustomDefaulter struct {
+	cronjob *CronJob
```
sbueringer, line 85: This is not concurrency safe. What if the webhook is called for two CronJobs at the same time?

I would just drop the field and pass the cronJob into the setDefault func

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -118,40 +160,80 @@ validate anything on deletion.
...
 
-var _ webhook.Validator = &CronJob{}
+// +kubebuilder:object:generate=false
+// CronJobCustomValidator struct is responsible for validating the CronJob resource
+// when it is created, updated, or deleted.
+//
+// NOTE: The +kubebuilder:object:generate=false marker prevents controller-gen from generating DeepCopy methods,
+// as this struct is used only for temporary operations and does not need to be deeply copied.
+type CronJobCustomValidator struct {
+	cronjob *CronJob
```
sbueringer, line 170: Same here, not concurrency safe

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -59,31 +69,63 @@ The meaning of each marker can be found [here](/reference/markers/webhook.md).
...
-func (r *CronJob) Default() {
-	cronjoblog.Info("default", "name", r.Name)
+// Default implements webhook.CustomDefaulter so a webhook will be registered for the Kind CronJob
+func (d *CronJobCustomDefaulter) Default(ctx context.Context, obj runtime.Object) error {
+	cronjoblog.Info("CustomDefaulter for CronJob")
+	cronjob, ok := obj.(*CronJob)
+	if !ok {
+		return fmt.Errorf("expected an CronJob object but got %T", obj)
+	}
+	cronjoblog.Info("default", "name", cronjob.GetName())
```
sbueringer, line 103: To be honest. This log would be very very noisy (I personally wouldn't add it)

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -118,40 +160,80 @@ validate anything on deletion.
...
+var _ webhook.CustomValidator = &CronJobCustomValidator{}
+
+// ValidateCreate implements webhook.CustomValidator so a webhook will be registered for the type CronJob
+func (v *CronJobCustomValidator) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	cronjoblog.Info("Creation Validation for CronJob")
+	cronjob, ok := obj.(*CronJob)
+	if !ok {
+		return nil, fmt.Errorf("expected a CronJob object but got %T", obj)
+	}
+	cronjoblog.Info("default", "name", cronjob.GetName())
```
sbueringer, line 184: Not "default" here, but same as the log above. This will produce a huge amount of noise in the controller logs (also the log in l.179)

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
camilamacedo86: That is a very good catcher!!! 
I will change.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
camilamacedo86: very good point and cacher !!!
Thank you

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
camilamacedo86: Good point about the default
we should get the request namespace 
But I think we need to keep the logs. I will give a look with love thx

sbueringer: Up to you. But I think in a regular use cases defaulting and validating webhooks are called a lot. Logging this with log level 0 will lead to a lot of noise

sbueringer:
Lgtm, only reviewed cronjob-tutorial and testdata/project-v4/api/v1 & testdata/project-v4-multigrou/api though

camilamacedo86 (commented):
Moving forward it seems good enough to get merged
We have lgtm's 
We can address full improvements and fixes in follow up always as well.

Thank you a lot for your help @sbueringer @varshaprasad96 !!!
We could not move forward here without your contribution and amazing inputs. 🥇

Issue #1241, opened 2019-12-05
document when to use the high-level Defaulter/Validator interface vs the lower-level handler interfaces


DirectXMan12:
Namely, if you need to do anything with clients, need to configure things differently, etc, you prob want the lower-level interfaces.

/kind documentation
/help

DirectXMan12:
cc @mengqiy

PR #4061, opened 2024-08-09
📖  update and supplement webhook documentation with example to work with core types

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/pull/3399
also, closes: https://github.com/kubernetes-sigs/kubebuilder/issues/1241

### DIFF
--- a/docs/book/src/reference/webhook-for-core-types.md
+++ b/docs/book/src/reference/webhook-for-core-types.md
@@ -9,45 +9,85 @@ in controller-runtime.
 It is suggested to use kubebuilder to initialize a project, and then you can
 follow the steps below to add admission webhooks for core types.
 
-## Implement Your Handler
+## Implementing Your Handler Using `Handle`
 
-You need to have your handler implements the
-[admission.Handler](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/webhook/admission?tab=doc#Handler)
-interface.
+Your handler must implement the [admission.Handler](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/webhook/admission#Handler) interface. This function is responsible for both mutating and validating the incoming resource.
+
+### Update your webhook:
+
+**Example**
 
 ```go
+package v1
+
+import (
+    "context"
+    "encoding/json"
+    "net/http"
+    "sigs.k8s.io/controller-runtime/pkg/client"
+    "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+    corev1 "k8s.io/api/core/v1"
+)
+
+// **Note**: in order to have controller-gen generate the webhook configuration for you, you need to add markers. For example:
+
+// +kubebuilder:webhook:path=/mutate--v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io
+
 type podAnnotator struct {
-	Client  client.Client
-	decoder *admission.Decoder
+    Client  client.Client
+    decoder *admission.Decoder
 }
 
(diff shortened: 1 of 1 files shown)


docs/book/src/reference/webhook-for-core-types.md
```diff
@@ -58,13 +88,175 @@ If you need a client and/or decoder, just pass them in at struct construction ti
...
+// SetupWebhookWithManager sets up the webhook with the manager for both the defaulter and validator
+func (r *corev1.Pod) SetupWebhookWithManager(mgr ctrl.Manager) error {
+  return ctrl.NewWebhookManagedBy(mgr).
+    For(r).
+    WithValidator(&PodCustomValidator{}).
+    WithDefaulter(&PodCustomDefaulter{}).
+    Complete()
+}
+
+// +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io,admissionReviewVersions=v1
```
mythi, line 134: IIRC I had to add something like https://github.com/mythi/kubebuilder/blob/2db6ee6f869b69ba46501b734e50c41a438bc5d6/docs/book/src/reference/webhook-for-core-types.md?plain=1#L47-L50 to get it working.

docs/book/src/reference/webhook-for-core-types.md
```diff
@@ -9,45 +9,75 @@ in controller-runtime.
...
+    "encoding/json"
+    "net/http"
+    "sigs.k8s.io/controller-runtime/pkg/client"
+    "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+    corev1 "k8s.io/api/core/v1"
+)
+
+// **Note**: in order to have controller-gen generate the webhook configuration for you, you need to add markers. For example:
+
+// +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io
```
camilamacedo86, line 47:
```suggestion
// +kubebuilder:webhook:path=/mutate--v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io
```

docs/book/src/reference/webhook-for-core-types.md
camilamacedo86: The same marker is in the line 47 
What exactly is missing could you please let me know?

mythi: I had this clarification added: "(for webhooks the `path` is of format `/mutate-<group>-<version>-<kind>`. Since this documentation uses `Pod` from the core API group, the group needs to be an empty string)"

the marker is: `/mutate--v1-pod`

camilamacedo86: done good catcher

mythi (approved):
this looks good and is very informative, thanks

camilamacedo86:
/hold

We need to get merged: https://github.com/kubernetes-sigs/kubebuilder/pull/4060
So that we can have the convey and then have the examples here updated accordingly.

camilamacedo86 (commented):
We are for a good time trying to fix this doc :-) 
It has a LGTM
And the examples with CustomInterfaces are respecting the comments and the changes addressed in #4060.

So, it shows good enough to fly 🚀  since it is better than before. Therefore, moving forward with this one.

However, further improvements are welcome always.
If you see that need changes please, feel free to raise an PR with your suggestions.

mythi:
> So, it shows good enough to fly 🚀 since it is better than before. Therefore, moving forward with this one.

+1, thanks!

PR #4065, opened 2024-08-11
⚠️  (cleanup/fix): ImplementWebhooks method should only be used by the e2e tests and should be under its package

camilamacedo86:
This code implementation is valid only for the e2e tests and is not an Plugin Utils that can be used for those who consumes Kuebebuilder as a lib and should in this way be under tests/e2e/


camilamacedo86:
Just a nit, make no sense make people spend time on reviews.
we have big fishes to fry. Moving forward

PR #4066, opened 2024-08-11
🌱 (cleanup): cleanup tests for plugin/util functions

camilamacedo86:
Have a suite test for the plugin utils and remove the need to have the testdata dir scaffolded


camilamacedo86:
Just a nit, make no sense make people spend time on reviews.
we have big fishes to fry. Moving forward

PR #4069, opened 2024-08-12
📖 Upgrade the muilt-version sample to the latest

camilamacedo86:
Just update the sample used in the Multi-version tutorial with the latest changes


camilamacedo86:
Just update the sample, make no sense make people spend time on reviews.
we have bigger fishes to fry. Moving forward :-)

Issue #4074, opened 2024-08-13
https://book.kubebuilder.io/cronjob-tutorial/cronjob-tutorial


remotejob:
make run

go run ./cmd/main.go

`vet: api/v1/zz_generated.deepcopy.go:126:10: cannot use c (variable of type *CronJobSpec) as "k8s.io/apimachinery/pkg/runtime".Object value in return statement: *CronJobSpec does not implement "k8s.io/apimachinery/pkg/runtime".Object (missing method GetObjectKind)`

camilamacedo86:
Hi @remotejob 

The only issue that I found was `Too long: must have at most 262144 bytes`

```
$ make install
/Users/camiladeomacedo/go/src/sigs.k8s.io/kubebuilder/docs/book/src/cronjob-tutorial/testdata/project/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
Downloading sigs.k8s.io/kustomize/kustomize/v5@v5.4.2
/Users/camiladeomacedo/go/src/sigs.k8s.io/kubebuilder/docs/book/src/cronjob-tutorial/testdata/project/bin/kustomize build config/crd | kubectl apply -f -
The CustomResourceDefinition "cronjobs.batch.tutorial.kubebuilder.io" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
make: *** [install] Error 1
```

To address this one we should be scaffolding the Makefile with `crd:maxDescLen=0`
We need change the hack/docs to add this replace (it seems a regression), as:

```
.PHONY: manifests
manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
	# Note that the option maxDescLen=0 was added in the default scaffold in order to sort out the issue
	# Too long: must have at most 262144 bytes. By using kubectl apply to create / update resources an annotation
	# is created by K8s API to store the latest version of the resource ( kubectl.kubernetes.io/last-applied-configuration).
	# However, it has a size limit and if the CRD is too big with so many long descriptions as this one it will cause the failure.
	$(CONTROLLER_GEN) rbac:roleName=manager-role crd:maxDescLen=0 webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```

However, after change the `make manifest` target the make run worked fine:


```
$ export ENABLE_WEBHOOKS=false
$ make run
[...]

PR #4075, opened 2024-08-13
📖  docs:samples:cronjob: change makefile to generate the CRD without description to allow apply CRDs

camilamacedo86:
If the CRD are too big we will face ` Too long: must have at most 262144 bytes` when we run make install
More info: https://book.kubebuilder.io/faq#the-error-too-long-must-have-at-most-262144-bytes-is-faced-when-i-run-make-install-to-apply-the-crd-manifests-how-to-solve-it-why-this-error-is-faced

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/4074


camilamacedo86:
To small, moving forward.

PR #4081, opened 2024-08-17
📖 : Upgrade controller-gen used to generate the docs from 0.15.0 to 0.16.1

camilamacedo86:
Upgrade controller-gen used to generate the docs from 0.15.0 to 0.16.1


camilamacedo86:
Just update the docs for new controller-gen version
It has no reason for we be blocked were make someone waste time on this review

PR #4083, opened 2024-08-17
📖 (cleanup and enhance) improve and make clear info about setup envtest and its binaries and that old location is deprecated

camilamacedo86:
This PR only update the Artifact docs in order to:
- Make clear the new location and that users should no longer use https://storage.googleapis.com/kubebuilder-tools
- Remove info from the deprecated location and keep it synced with the latest changes

More info: https://github.com/kubernetes-sigs/kubebuilder/discussions/4082


camilamacedo86:
it is too small and we need it for the release
Therefore, make no sense wait for reviews.
We can always change it if required afterwords.

Cheers,

PR #4092, opened 2024-08-23
:book: update the next step from quick start accordingly

camilamacedo86:
Now that we have the Getting Started guide, which provides a solid overview, the most appropriate next step seems to be following this guide. Lastly, we are removing the note from the DeployImage section as it doesn't seem to fit well on this page.


camilamacedo86:
To small and simple make no sense we make someone waste time on this review.
So, I am moving forward

PR #4095, opened 2024-08-24
🌱 (ci) - Fix the job to check the PR title and no longer use deprecated github action

camilamacedo86:
The kubebuilder-release-tools project is no longer really maintained. The image required for checks is deprecated due 
the infrastructure used to build and promote it be deprecated. Given the minimal nature of the check, it's unnecessary to maintain a GitHub Action for this purpose.

Also, it broke 1 day ago. Note that we need to pass permissions via GitHub Token, which is not very safe. This is another reason why we need to proceed with these changes.

**When it fails:**

Screenshot 2024-08-24 at 11 21 48

**When the check pass:**

Screenshot 2024-08-24 at 11 22 03

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -1,18 +1,22 @@
-name: PR Verifier
+name: "PR Title Verifier"
 
 on:
-  pull_request_target:
-    types: [opened, edited, reopened]
-
+  pull_request:
+    types: [opened, edited, synchronize, reopened]
 
 jobs:
-
   verify:
-    name: Verify PR contents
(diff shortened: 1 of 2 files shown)


camilamacedo86:
Hi @sbueringer 

We need to deprecate it in the kubebuilder tools repo 
and communicate. I will try to move it in the follow up weeks
However, here's how we can achieve the same result without the action, which is actually safer, cleaner, and faster.

I will move with this one to fix our CI
but if you have suggestions please feel free to share and we can try to push in a follow up

PR #4093, opened 2024-08-23
:book: Minor Documentation Improvements: Add Links, Fix Typos, and Clarify Instructions

camilamacedo86:
This pull request includes several small but important improvements to the documentation, aimed at enhancing clarity and providing additional resources:

- Added a link to the CronJob Tutorial in the "Next Steps" section of the Getting Started guide.
- Fixed minor issues with documentation headings and improved consistency in the Quick Start guide.
- Clarified the need to push images in the Running guide with an added note.
- Recommended using Kind for testing in the Getting Started guide.
- Corrected a typo in the link to the Kind guide.
- Added a brief note in the "Next Steps" section about the Deploy Image plugin to offer further guidance.

PR #4096, opened 2024-08-24
:book: remove time estimates in Next Steps of QuickStarter

camilamacedo86:
Small nit to avoid specifying time estimates, as they can be relative and may lead to misunderstandings.
Follow up of: https://github.com/kubernetes-sigs/kubebuilder/pull/4093


camilamacedo86:
Moving forward it is a nit and fix my own mistake (Sorry).
I will not make people waste time on this review

PR #4099, opened 2024-08-24
🌱 move pr title checker to be under test with ther others checks

camilamacedo86:
Nit, follow up just to keep the check with the others


camilamacedo86:
to small to ask for reviews
we need help with many prs, so I am moving forward here.

PR #4100, opened 2024-08-25
:bug: (deploy-image/v1-alpha1): Fix typo issue by replacing typeNamespaceName with typeNamespacedName

camilamacedo86:
NIT: Just fix the name/term in the controller-test scaffolds which are done by the plugin.


camilamacedo86 (commented):

PR #4101, opened 2024-08-25
:bug: (deploy-image/v1-alpha1): Fix sample by ensuring the right labels

camilamacedo86:
The config/sample generated when we are using the plugin was missing the labels

````yaml
  labels:
    app.kubernetes.io/name: {{ .ProjectName }}
```


camilamacedo86 (commented):

PR #4102, opened 2024-08-25
:sparkles: (deploy-image/v1-alpha1): Add comments to clarify resource watching and reconciliation logic

camilamacedo86:
Added detailed comments in the SetupManager scaffold within the controller to explain how the resource watching and reconciliation process works. These comments clarify how the controller watches for changes in Custom Resources and Deployments, ensuring that the cluster maintains the desired state.


camilamacedo86:

PR #4104, opened 2024-08-25
:book: remove notes about crd-version and webhook-version flags which are no longer supported and provided 

camilamacedo86:
Those flgas were added to allow users still using the versions v1beta1 which is no longer supported and prvided since k8s 1.22 and were deprecated a long time ago and cannot be used with go/v4. Those are removed since is no longer possible to support those older version with the dependencies versions used currently. More info: https://github.com/kubernetes-sigs/kubebuilder/discussions/3622


camilamacedo86 (commented):

PR #4105, opened 2024-08-25
:book: fix link for reference doc to manager and crd scopes

camilamacedo86:
Just fix the link it is broken


camilamacedo86:

camilamacedo86:
To small moving forward.
We really need help on the review of more significant PRs

Issue #4089, opened 2024-08-21
Enhance E2E Test Scaffold with Optional Prometheus and Cert-Manager Installation


camilamacedo86:
### What do you want to happen?

Projects are scaffolded with e2e tests, which by default, should be executed in a specific environment. The common practice is to spin up an environment, run the tests, and then tear down the environment to ensure the tests are executed in isolation.

However, some users might prefer to run these tests in an environment where Prometheus and Cert-Manager are already installed, or in cases where their project does not require these tools. Users should still be able to customize their scaffold as needed. (Motivation: https://github.com/kubernetes-sigs/kubebuilder/issues/3971)

To facilitate this, we can add a logic that uses an environment variable to skip the installation of these tools, such as INSTALL_PROMETHEUS_E2E_TEST=false.

The changes should be made in the following section of the code:

https://github.com/kubernetes-sigs/kubebuilder/blob/94d8fce49227f87f7a4e79f9c5d121e1e942b5bf/testdata/project-v4/test/e2e/e2e_test.go#L32-L43

Additionally, we have a practice of adding `TODO(user): comments`. We must ensure that this code includes comments explaining how the tests should be used and describing the ability to disable the installation of Prometheus and Cert-Manager in the e2e tests.

TAM360:
@camilamacedo86 I would also like to work on this as it is linked with the other task too.

camilamacedo86:
> @camilamacedo86 so by default the steps to install, and delete Prometheus, and Cert Manager would be commented in a new project, and if anyone who wants to use it would have un-comment the relevant lines first?

**Honest, I do not know how much value comment it will bring or if it will bring more pain**
Because either install the solution in a cluster where it should not be done will cause problems.

Note that if someone scaffold webhooks then need install cert-manager
Also, we would need to uncomment either our tests: https://github.com/kubernetes-sigs/kubebuilder/blob/master/.github/workflows/test-sample-go.yml

In this way, wdyt about we try to do first: https://github.com/kubernetes-sigs/kubebuilder/issues/4088
Then, we revisit this one again to see if still fit and if it would make sense.

camilamacedo86:
Hi @TAM360 

Furthermore, users need to customize the implementation according to their needs.
But, lets think here what we can do to help out.

PR #4106, opened 2024-08-26
:sparkles: (go/v4) e2e tests: improve e2e tests and make test-e2e target

camilamacedo86:
- Ensure that kind is installed and running before run the tests. Otherwise it will fail since we need to load the Manager(Operator) image
- Add logic to skip the installation of CertManager and/or Prometheus via envvars.
- Ensure that the promethues and certmanager are installed in the suite test instead of beafore each test
- Ensure that the image is build and load in the suite instead instead for each test
- Add more comments to clarify the purpose of the tests
- Add TODO(user) to clarify that is expected action for the users to suplement and/or customize their e2e tests according to their needs

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/4089
c/c  @DerekTBrown, @majewsky  @TAM360 @defo89

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/Makefile
+++ b/docs/book/src/cronjob-tutorial/testdata/project/Makefile
@@ -67,9 +67,21 @@ vet: ## Run go vet against code.
 test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
 
-# Utilize Kind or modify the e2e tests to load the image locally, enabling compatibility with other vendors.
-.PHONY: test-e2e  # Run the e2e tests against a Kind k8s instance that is spun up.
-test-e2e:
+# TODO(user): To use a different vendor for e2e tests, modify the setup under 'tests/e2e'.
+# The default setup assumes Kind is pre-installed and builds/loads the Manager Docker image locally.
+# Prometheus and CertManager are installed by default; skip with:
+# - PROMETHEUS_INSTALL_SKIP=true
+# - CERT_MANAGER_INSTALL_SKIP=true
+.PHONY: test-e2e
+test-e2e: ## Run the e2e tests. Expected an isolated environment using Kind.
+	@command -v kind >/dev/null 2>&1 || { \
+		echo "Kind is not installed. Please install Kind manually."; \
+		exit 1; \
+	}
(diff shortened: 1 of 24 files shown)


camilamacedo86:
Hi @DerekTBrown, @majewsky

The changes on this PR should be enough to sort out the scenarios that you raised.
See that we need to have kind up and running for the e2e-tests
Therefore, if anyone run the integration tests without customize them against an env which is not using kind then the test will fail.

c/c @TAM360

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -67,9 +67,21 @@ vet: ## Run go vet against code.
...
+.PHONY: test-e2e
+test-e2e: ## Run the e2e tests. Expected an isolated environment using Kind.
+	@command -v kind >/dev/null 2>&1 || { \
+		echo "Kind is not installed. Please install Kind manually."; \
+		exit 1; \
+	}
+	@kind get clusters | grep -q 'kind' || { \
+		echo "No Kind cluster is running. Please start a Kind cluster before running the e2e tests."; \
+		exit 1; \
+	}
```
camilamacedo86, line 84: Hi @DerekTBrown, @majewsky

The changes on this PR should be enough to sort out the scenarios that you raised.
See that we need to have kind up and running for the e2e-tests
Therefore, if anyone run the integration tests without customize them against an env which is not using kind then the test will fail.

c/c @TAM360

TAM360: @camilamacedo86 I think the documentation should also include the information about newly introduced environment variables, and the purpose behind them.

camilamacedo86: It is documented in the makefile target (see this file)
Also, in the e2e tests scaffolds golang docs.

TAM360: @camilamacedo86 wouldn't it be better idea to include it in the docs as well?

camilamacedo86: > @camilamacedo86 wouldn't it be better idea to include it in the docs as well?

It is not the goal of this PR
but we can add that under reference in a follow up.
I am totally OK with. Could you please raise an issue to track the need?

TAM360: @camilamacedo86  I'll create a new issue, once this PR merges.

camilamacedo86:
Moving forward since we have the LGTM from who asked for it and it was discussed in the slack
Also, because the changes here block we try to fix the e2e tests for deploy image sample

We either can push new changes and further improvements in follow ups.

PR #4113, opened 2024-08-29
🌱  ci- cleanup lint samples

camilamacedo86:
The lint sample should only run the lint checks for the samples
    Either, make no sense we run for all of them. We can ensure all plugins by just running project-v4 and project-v4-with-deploy-image which has further code scafold. By last, see that the `make tests` are called in the prow already.


camilamacedo86:

PR #4114, opened 2024-08-29
🌱 ci: fix e2e tests for samples

camilamacedo86:
The current test is not properly testing the e2e of the samples. Either we have an issue the e2e tests scaffolds for deploy-image sample but we will need to fix it in a follow up PR because more changes in the scaffold will be required.


camilamacedo86:
I am moving forward here because we need to ensure the CI
So that we can be confident in merge others changes such as : https://github.com/kubernetes-sigs/kubebuilder/pull/4106

PR #4115, opened 2024-08-29
🌱 ci: no longer call github actions for changes which are only made in the docs files (*/*.md)

camilamacedo86:
Small nit to avoid we spend time and costs running unnecessary tests


camilamacedo86:

PR #4117, opened 2024-08-29
✨ (go/v4):Added checks in the e2e test scaffolds to determine if Prometheus and/or CertManager are already installed on the cluster and avoid re-installation.

Adembc:
This PR adds pre-installation checks for **CertManager** and **Prometheus Operator** in the e2e tests. These checks ensure that these components are not reinstalled if they are already present, thereby avoiding conflicts and preventing their removal after the tests are completed.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_suite_test.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_suite_test.go
@@ -36,6 +36,10 @@ var (
 	// re-installation and conflicts.
 	skipPrometheusInstall  = os.Getenv("PROMETHEUS_INSTALL_SKIP") == "true"
 	skipCertManagerInstall = os.Getenv("CERT_MANAGER_INSTALL_SKIP") == "true"
+	// isPrometheusOperatorAlreadyInstalled will be set true when prometheus CRDs be found on the cluster
+	isPrometheusOperatorAlreadyInstalled = false
+	// isCertManagerAlreadyInstalled will be set true when CertManager CRDs be found on the cluster
+	isCertManagerAlreadyInstalled = false
 
 	// projectImage is the name of the image which will be build and loaded
 	// with the code source changes to be tested.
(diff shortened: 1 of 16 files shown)


docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
```diff
@@ -75,6 +75,18 @@ func UninstallPrometheusOperator() {
...
+// installed by verifying the availability of a key deployment.
+func IsPrometheusOperatorInstalled() bool {
+	deployment := "prometheus-operator"
+	cmd := exec.Command("kubectl", "wait", "deployment.apps/"+deployment,
+		"--for", "condition=Available",
+		"--timeout", "10s",
+	)
+	_, err := Run(cmd)
+	return err == nil
+}
```
camilamacedo86, line 88: We need to check for the CRDs. If the cluster has Prometheus or CertManager we do not know the name of the deployment, the namespace where it is installed and etc. 

So, I think we need something like

```
func IsPrometheusCRDsInstalled() bool {
	// Execute the kubectl command to get all CRDs that contain "prometheus"
	cmd := exec.Command("kubectl", "get", "crds")
	output, err := Run(cmd)
	if err != nil {
		fmt.Printf("Failed to list CRDs: %v\n", err)
		return false
	}

	// Check if any of the CRDs contain "prometheus"
	crdList := GetNonEmptyLines(string(output))
	for _, crd := range crdList {
		if strings.Contains(crd, "prometheus") {
			fmt.Println("Found a Prometheus-related CRD.")
			return true
		}
	}

	fmt.Println("No Prometheus-related CRDs found.")
	return false
}
```

docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
```diff
@@ -103,6 +115,30 @@ func InstallCertManager() error {
...
+			"--timeout", "10s",
+		)
+		_, err = Run(cmd)
+		if err != nil {
+			return false
+		}
+	}
+
+	return true
+}
```
camilamacedo86, line 140: Same here 

```go
// IsCertManagerCRDsInstalled checks if any Cert Manager CRDs are installed
// by verifying the existence of key CRDs related to Cert Manager.
func IsCertManagerCRDsInstalled() bool {
	// List of common Cert Manager CRDs
	certManagerCRDs := []string{
		"certificates.cert-manager.io",
		"issuers.cert-manager.io",
		"clusterissuers.cert-manager.io",
		"certificaterequests.cert-manager.io",
		"orders.acme.cert-manager.io",
		"challenges.acme.cert-manager.io",
	}

	// Execute the kubectl command to get all CRDs
	cmd := exec.Command("kubectl", "get", "crds")
	output, err := Run(cmd)
	if err != nil {
		fmt.Printf("Failed to list CRDs: %v\n", err)
		return false
	}

	// Check if any of the Cert Manager CRDs are present
	crdList := GetNonEmptyLines(string(output))
	for _, crd := range certManagerCRDs {
		for _, line := range crdList {
			if strings.Contains(line, crd) {
				fmt.Printf("Found Cert Manager CRD: %s\n", crd)
				return true
			}
		}
	}

	fmt.Println("No Cert Manager-related CRDs found.")
	return false
}
```

camilamacedo86 (changes requested):
Hi @Adembc 

Thank you a lot for the contribution.
Just a few nits.
- a) Because it changes the scaffold cannot be 🌱 . I hope that you do not mind I will update the title since we use it for the release notes just to show 
- b) If our goal is not re-install Prometheus and/or CertManager in a cluster where it exist already, we have no idea where the Prometheus and/or CertManager were installed. Either the common problem is re-apply the CRDs. So, we cannot verify by checking the deployment and status because we do not know its name and where it will be applied. We would need to check of the CRDs exist already on the cluster.  See: https://github.com/kubernetes-sigs/kubebuilder/pull/4117/files#r1737991875

Otherwise, all great. I think by doing this small change and ensuring that it pass in the CI we can move forward with this one. 


Well done 🥇 
And great work !!!

Thank you a lot for your contribution

docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
camilamacedo86: WDYT?

Adembc: Good point. Checking for CRDs would be more reliable since the deployment names and namespaces can vary.

docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_suite_test.go
```diff
@@ -34,8 +34,10 @@ var (
 	// - CERT_MANAGER_INSTALL_SKIP=true: Skips CertManager installation during test setup.
 	// These variables are useful if Prometheus or CertManager is already installed, avoiding
 	// re-installation and conflicts.
-	skipPrometheusInstall  = os.Getenv("PROMETHEUS_INSTALL_SKIP") == "true"
-	skipCertManagerInstall = os.Getenv("CERT_MANAGER_INSTALL_SKIP") == "true"
+	skipPrometheusInstall                = os.Getenv("PROMETHEUS_INSTALL_SKIP") == "true"
+	skipCertManagerInstall               = os.Getenv("CERT_MANAGER_INSTALL_SKIP") == "true"
+	isPrometheusOperatorAlreadyInstalled = false
+	isCertManagerAlreadyInstalled        = false
```
camilamacedo86, line 40:
```suggestion
        // isPrometheusOperatorAlreadyInstalled will be set true when prometheus CRDs be found on the cluster
	isPrometheusOperatorAlreadyInstalled = false
	// isCertManagerAlreadyInstalled will be set true when certmanager CRDs be found on the cluster
	isCertManagerAlreadyInstalled        = false
```

docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_suite_test.go
```diff
@@ -64,24 +66,36 @@ var _ = BeforeSuite(func() {
 	err = utils.LoadImageToKindClusterWithName(projectImage)
 	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Failed to load the manager(Operator) image into Kind")
 
-	// Setup Prometheus and CertManager before the suite if not skipped
+	// Setup Prometheus and CertManager before the suite if not skipped and if not already installed
 	if !skipPrometheusInstall {
-		_, _ = fmt.Fprintf(GinkgoWriter, "Installing Prometheus Operator...\n")
-		Expect(utils.InstallPrometheusOperator()).To(Succeed(), "Failed to install Prometheus Operator")
+		_, _ = fmt.Fprintf(GinkgoWriter, "Checking for existing installation of Prometheus Operator...\n")
+		isPrometheusOperatorAlreadyInstalled = utils.IsPrometheusOperatorInstalled()
```
camilamacedo86, line 72:
```suggestion
                By("checking if prometheus is installed already") 
		isPrometheusOperatorAlreadyInstalled = utils.IsPrometheusOperatorInstalled()
```

camilamacedo86: I think we need add a comment to explain why we are doing those checks
Also, we need to use By("") to be able to properly check it in the tests logs

camilamacedo86 (approved):
Great !
If it pass in the CI I am OK with all changes

/lgtm

PR #4118, opened 2024-08-30
🌱 ci: fix GitHub action to test make test-e2e for deploy-image

camilamacedo86:
For the DeployImage sample, we are only using validating webhooks. 
This commit updates the scaffold to reflect this by customizing 
the configuration accordingly.

Additionally, there is a known bug in the tool where it scaffolds 
uncommented data for MutatingWebhooks. As a workaround, this commit 
comments out the unnecessary MutatingWebhook configurations to avoid 
issues during the test execution.


camilamacedo86:

Also, anyone can propose further changes in follow ups
All help is welcome !
Please, feel free.

Issue #4120, opened 2024-08-30
E2e tests scaffolded by default


camilamacedo86:
### What do you want to happen?

We have basic checks which are scaffolded as e2e tests for the samples.
We can improve the default scaffold to help more the users 
We might want to check our e2e tests and ensure that has nothing else that is a common need for all case scenarios that we would not like to add to the e2e tests scaffolds in the projects

See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v4/plugin_cluster_test.go

camilamacedo86:
See that for one we have: https://github.com/kubernetes-sigs/kubebuilder/pull/4131

PR #4121, opened 2024-08-31
:sparkles: (go/v4) Add scaffold for e2e webhook checks

camilamacedo86:
A new scaffold marker is introduced to automatically adds end-to-end tests for webhooks in the test/e2e/test.go
    
The scaffolded tests include:
- Validation that cert-manager has successfully provisioned the secret, when/if an webhook is scaffold for the project.
- Verification that the CA bundle is injected into the mutating webhooks, if a mutating webhook is scaffolded.
- Verification that the CA bundle is injected into the validating webhooks, if a validating webhook is scaffolded.

PS.: See that a new maker `+kubebuilder:scaffold:e2e-webhooks-checks` is introduced to allow we add to the e2e tests the scaffolds only when users use kb create webhook accordinly. 

Partially closes: https://github.com/kubernetes-sigs/kubebuilder/issues/4120

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_test.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_test.go
@@ -184,6 +184,46 @@ var _ = Describe("Manager", Ordered, func() {
 			))
 		})
 
+		It("should provisioned cert-manager", func() {
+			By("validating that cert-manager has the certificate Secret")
+			verifyCertManager := func(g Gomega) {
+				cmd := exec.Command("kubectl", "get", "secrets", "webhook-server-cert", "-n", namespace)
+				_, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred())
+			}
+			Eventually(verifyCertManager).Should(Succeed())
+		})
+
+		It("should have CA injection for mutating webhooks", func() {
+			By("checking CA injection for mutating webhooks")
+			verifyCAInjection := func(g Gomega) {
+				cmd := exec.Command("kubectl", "get",
+					"mutatingwebhookconfigurations.admissionregistration.k8s.io",
+					"project-mutating-webhook-configuration",
+					"-o", "go-template={{ range .webhooks }}{{ .clientConfig.caBundle }}{{ end }}")
+				mwhOutput, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(len(mwhOutput)).To(BeNumerically(">", 10))
+			}
+			Eventually(verifyCAInjection).Should(Succeed())
+		})
+
+		It("should have CA injection for validating webhooks", func() {
+			By("checking CA injection for validating webhooks")
+			verifyCAInjection := func(g Gomega) {
+				cmd := exec.Command("kubectl", "get",
+					"validatingwebhookconfigurations.admissionregistration.k8s.io",
+					"project-validating-webhook-configuration",
+					"-o", "go-template={{ range .webhooks }}{{ .clientConfig.caBundle }}{{ end }}")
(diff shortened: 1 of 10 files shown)


camilamacedo86:
c/c @Adembc

docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_test.go
```diff
@@ -108,6 +108,47 @@ var _ = Describe("controller", Ordered, func() {
...
+
+		It("should have CA injection for mutating webhooks", func() {
+			By("checking CA injection for mutating webhooks")
+			verifyCAInjection := func() error {
+				cmd := exec.Command("kubectl", "get",
+					"mutatingwebhookconfigurations.admissionregistration.k8s.io",
+					"project-mutating-webhook-configuration",
+					"-o", "go-template={{ range .webhooks }}{{ .clientConfig.caBundle }}{{ end }}")
+				mwhOutput, err := utils.Run(cmd)
+				ExpectWithOffset(2, err).NotTo(HaveOccurred())
```
mogsie, line 128: You can't use Expects inside an Eventually function without sending in the `g Gomega` to the function. The `g.Expect()` will be retried, but a naked `Expect()` will cause the test to fail, even if it _eventually_ succeeded.

docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_test.go
```diff
@@ -108,6 +108,47 @@ var _ = Describe("controller", Ordered, func() {
...
+		})
+
+		It("should have CA injection for mutating webhooks", func() {
+			By("checking CA injection for mutating webhooks")
+			verifyCAInjection := func() error {
+				cmd := exec.Command("kubectl", "get",
+					"mutatingwebhookconfigurations.admissionregistration.k8s.io",
+					"project-mutating-webhook-configuration",
+					"-o", "go-template={{ range .webhooks }}{{ .clientConfig.caBundle }}{{ end }}")
+				mwhOutput, err := utils.Run(cmd)
```
mogsie, line 127: These three lines can be replaced with one:

```
g.Expect(utils.Run(cmd)).To(HaveLen(BeNumerically(">", 10)))
```

When Expect gets a second error type as a parameter, it ensures that the error is nil before checking the assertions on the actual value.

docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_test.go
```diff
@@ -108,6 +108,47 @@ var _ = Describe("controller", Ordered, func() {
 			EventuallyWithOffset(1, verifyControllerUp, time.Minute, time.Second).Should(Succeed())
 		})
 
+		It("should provisioned cert-manager", func() {
+			By("validating that cert-manager has the certificate Secret")
+			EventuallyWithOffset(1, func() error {
```
mogsie, line 113: Here too, it is (IMHO) easier to read when you send in the `g Gomega` and do regular Gomega assertions rather than typical go _error checking_.  Something like this, perhaps:

```
EventuallyWithOffset(1, func(g Gomega) {
    cmd := ...  
    g.Expect(utils.Run(cmd)).Error().ToNot(HaveOccurred())
})
```

camilamacedo86:
HI @mogsie 

Would you like to help out on this one by getting this PR and changing the asserts as you did with : https://github.com/kubernetes-sigs/kubebuilder/pull/4131?

camilamacedo86:
We have an LGTM
And all suggestion done by @mogsie are either addressed
So, I am moving forward here (we have big fishes to fry and it will help us to validate other changes)

However, if you @mogsie  or any other person have any further suggestion to improvements and etc please, feel free to raise an PR with them. All help is welcome !!!

PR #4122, opened 2024-08-31
🐛 (go/v4): fix test e2e by ensuring that make manifests and generate are executed as part of the tests

camilamacedo86:
We are running make manifests and make generate in the github action when those should be called by the e2e tests since we can trigger those locally, for example, from IDE to troubleshooting


camilamacedo86:
c/c @Adembc

camilamacedo86:

Also, anyone can propose further changes in follow ups 
All help is welcome !
Please, feel free.

Issue #4119, opened 2024-08-30
Error When Scaffolding Only Validation Webhooks in Kubebuilder


camilamacedo86:
### What broke? What's expected?

When attempting to scaffold only `ValidatingWebhooks` in a new `kubebuilder` project, I encountered an issue where `MutatingWebhookConfiguration` references still cause errors, despite ensuring that all related configuration for `MutatingWebhookConfiguration` is commented out or removed.

I expected be able to configure the default scaffold ONLY with validation webhooks and without defaulting/multation. The scaffold should help and allow users be able to configure only one of them and work well in this case. We should ensure that with e2e tests as well. 

- Create only Validation webhooks and enable cert-manager only for it
- Create only Defaulting webhooks and enable cert-manager only for it
- In both cases we should be able to deploy and run the projects 

### Reproducing this issue

1. **Initialize a new Kubebuilder project:**

   ```bash
   kubebuilder init --domain testproject.org --repo sigs.k8s.io/test
   ```

2. **Create a new API:**

```bash
kubebuilder create api --group ship --version v1beta1 --kind Frigate
```

3. **create the webhook**

```bash
kubebuilder create webhook --group ship --version v1beta1 --kind Frigate --programmatic-validation
```

4. **Uncomment relevant sections in config/default/kustomization.yaml to enable cert-manager**

NOTE: Indeed if I only uncomment all that is required only for ValidationConfig I still facing the issue

5. **Generate the manifests and kustomize**

```
make generate
make manifests
make build-installer
```

Error

```
Error: no resource matches strategic merge patch "MutatingWebhookConfiguration.v1.admissionregistration.k8s.io/mutating-webhook-configuration.[noNs]": no matches for Id MutatingWebhookConfiguration.v1.admissionregistration.k8s.io/mutating-webhook-configuration.[noNs]; failed to find unique target for patch MutatingWebhookConfiguration.v1.admissionregistration.k8s.io/mutating-webhook-configuration.[noNs]
make: *** [build-installer] Error 1
```

### KubeBuilder (CLI) Version

master

### PROJECT version

4.2.0

### Plugin versions

```yaml
go/v4
kustomize/v2
```

jamshidi799:
Hi. may I pick this issue?

PR #4123, opened 2024-08-31
🐛 (go/v4,kustomize/v2): Fix problems by simplify scaffold and removing webhookcainjection_patch. Clarifying replacements for cert-manager

camilamacedo86:
- Removed config/default/webhookcainjection_patch.yaml to streamline the scaffold.
- Clarified replacements blocks in kustomization.yaml for easier understanding. Each block is now labeled with instructions for uncommenting based on specific webhook scenarios (ValidatingWebhook, DefaultingWebhook, ConvertingWebhook).

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/3538
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/4119


testdata/project-v4-multigroup-with-deploy-image/config/default/kustomization.yaml
```diff
@@ -45,107 +45,133 @@ patches:
...
+#     kind: Certificate
+#     group: cert-manager.io
+#     version: v1
+#     name: serving-cert # This name should match the one in certificate.yaml
+#     fieldPath: .metadata.namespace # Namespace of the certificate CR
+#   targets:
+#     - select:
+#         kind: ValidatingWebhookConfiguration
+#       fieldPaths:
+#         - .metadata.annotations.[cert-manager.io/inject-ca-from]
```
varshaprasad96, line 61: @camilamacedo86 one question in this - replacements are executed sequentially. If we enable all defaulting, mutating and conversion - wouldn't the annotation be overwritten given the field path is the same?

camilamacedo86: In out e2e tests -> we enable all of them see: https://github.com/kubernetes-sigs/kubebuilder/pull/4123/files#diff-78236af4c39bb75f281d4e8f6a6cca6c48917dd482c4b5b51360fe53c3dec199 ( we uncomment all and it woks )

 Also, the path is not the same right. One will replace the ValidatingWebhookConfiguration another the MutatingWebhookConfiguration, and etc. 
 
However, if you see any scenario where it could be an issue please, let us know so we can add an e2e tests to ensure that all is fine or not.

camilamacedo86: /hold

We are working to improve the coverage /tests so that we can ensure this one better

camilamacedo86: /hold cancel

Now, see that we are testing with e2e all samples under testdata and the tutorial ones
Also, for go/v4 we added a test to validate webhooks when they are applied in a ns that is not the same where the operator and its server is running 

So, it is very well covered too.

camilamacedo86:
c/c @lentzi90

camilamacedo86:
Hi @erikgb 

You might be a good person to help us with the review of this one.
WDYT?

varshaprasad96 (approved):
/lgtm
/approve

Thanks @camilamacedo86 !

Kavinjsir (approved):
Looks good to me!

Nit:
Could we probably remove the code also at:
- https://github.com/kubernetes-sigs/kubebuilder/blob/v4.2.0/docs/book/src/cronjob-tutorial/cert-manager.md?plain=1#L27-L29
- https://github.com/kubernetes-sigs/kubebuilder/blob/v4.2.0/docs/book/src/multiversion-tutorial/deployment.md?plain=1#L15-L16

PR #4125, opened 2024-09-01
🐛 fix: ensure that make test-e2e call the same targets of make test

camilamacedo86:
Ensure consistency across targets by adding calls to manifests, generate, fmt, and vet before running tests.


camilamacedo86:

PR #4126, opened 2024-09-02
🌱 cleanup: simplify and consolidate testdata samples

camilamacedo86:
The testdata samples are designed to showcase various options for users and enable comprehensive testing of tool commands. By aggregating the Grafana and deploy-image plugins into the multigroup layout, we streamline test coverage and reduce redundancy. This change ensures all options are tested within the multigroup layout while keeping project-v4 with and without the optional plugins, to help us ensure the references for users and keep our CI tests.

Now, instead of having a separate sample for each optional plugin, we have:

- project-v4
- project-v4-multigroup-with-plugins
- project-v4-with-plugins

This consolidation reduces the number of samples and tests, making the codebase easier to maintain without sacrificing test coverage or usability for users.


testdata/project-v4/PROJECT
```diff
@@ -45,9 +45,4 @@ resources:
   webhooks:
     defaulting: true
     webhookVersion: v1
-- controller: true
-  domain: testproject.org
-  group: crew
-  kind: Laker
-  version: v1
```
camilamacedo86, line 52: We are removing the Laker from here because we already test the creation of controllers without resource in the multi-group. Also, if we keep it here it will not allow we have a very comprehensive test e2e tests for this sample since it is an edge case and we users need to customize the controller for not see error likes

```
2024-09-02T00:37:09Z	ERROR	setup	unable to create controller	{"controller": "Laker", "error": "one of For() or Named() must be called"}
main.main
	/workspace/cmd/main.go:193
runtime.main
	/usr/local/go/src/runtime/proc.go:271
```

We cannot do that manually just for our e2e tests on the samples

camilamacedo86:
I am moving forward with this one.
I checked with @varshaprasad96 and she is also OK with this change.

PR #4129, opened 2024-09-03
🌱 fix testdata/project-v4 samples by removing call to generate Lakers controller without resource

camilamacedo86:
We are removing the Laker from here because we already test the creation of controllers without resource in the multi-group. Also, if we keep it here it will not allow we have a very comprehensive test e2e tests for this sample since it is an edge case and we users need to customize the controller for not see error likes

```
ERROR	setup	unable to create controller	{"controller": "Laker", "error": "one of For() or Named() must be called"}
main.main
	/workspace/cmd/main.go:193
runtime.main
	/usr/local/go/src/runtime/proc.go:271
```

We cannot do that manually just for our e2e tests on the samples


camilamacedo86:
That is nit that does not impacted enduser.
It will only allow us to move forward with the e2e tests inicitative and better ensure our scaffolds
Since we have a lot of PRs to review I will not make nobody waste time to checking this small one.
Just moving forward.

Issue #3454, opened 2023-06-13
Documentation request: e2e test with webhook


vsoch:
Hi Kubebuilder developers! I am looking for an example of an e2e test that also includes adding the webhook. We have some logic that is set in the webhook, and would like it to be there for the e2e test. An example adding to the testEnv might look like this:

https://github.com/kubernetes-sigs/jobset/blob/cfd9d7ce33da7599cdbd2590884481a99403afb8/test/integration/webhook/suite_test.go#L63-L70

And maybe that is close, but I wasn't able to get something similar working in the e2e test (and maybe I just need a subset, or a way to edit the cfg object generated?) It would be great if the docs here could provide an example like this. Thank you!

camilamacedo86:
I think this one is interesting. Thank you for raising it.
It seems that would be very nice add an example in the docs over how to implement e2e test for the webhooks.

That would mean we improve the doc samples used so that we can do it in the tutorial.
- See that the code source used in the docs came from the samples, such as: https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/cronjob-tutorial/testdata
- Also, see that when we run make generate we updated the samples used and we have the code implementation to so under the hack: https://github.com/kubernetes-sigs/kubebuilder/tree/master/hack/docs

vsoch:
@camilamacedo86 I don't think I can help here, but I just wanted to say I think you are a great maintainer! Your responsiveness on GitHub and in the slack (and interaction between the two) deserves a :clap:. Thank you for doing what you do! <3

ashutosh887:
I can work on this @vsoch if @camilamacedo86 agrees 

Thanks
/assign

camilamacedo86:
In this task we need to:
- Check if we can improve the file scaffold for suite_test under api for webhooks
- So, we need to check the options that we use for controllers ( under internal/controllers ) see if has something that we do there that we can either add to the suite test for webhooks.

Sajiyah-Salat:
Hello @ashutosh887 Are you working on this till now? It would be great for me to work on this and solve this issue.

ashutosh887:
Yes I'm on it @Sajiyah-Salat

Sajiyah-Salat:
Great please give me link of the pr or if you haven't started that yet

drewwells:
[...]

Issue #4062, opened 2024-08-10
Decouple  webhooks from APIs


camilamacedo86:
### What do you want to happen?

This issue aims to centralize information regarding the requests to change the kubebuilder project layouts to decouple the APIs from webhooks. What we need to do here is:

- Understand what would be the best approach and recommendations from Controller-Runtime maintainer to where/how scaffolds the projects after: 
Closes https://github.com/kubernetes-sigs/controller-runtime/issues/2596

For more context, see the discussion and concerns raised in the following issues:

- [Controller-runtime issue #2596](https://github.com/kubernetes-sigs/controller-runtime/issues/2596): Highlights the need to decouple API packages from controller-runtime versions.
- [Kubebuilder issue #3215](https://github.com/kubernetes-sigs/kubebuilder/issues/3215): The original request where users expressed their concern about the inclusion of controller-runtime code in public API packages. Key comment: [Issue #3215 Comment](https://github.com/kubernetes-sigs/kubebuilder/issues/3215#issuecomment-1446723519)

**NOTE**

We need avoid breaking changes and we cannot impact our end users. Therefore, we would need to consider creating a `go/v5-alpha` plugin. Unless we decide to move forward based in some rationality exception case, as a follow up of https://github.com/kubernetes-sigs/kubebuilder/pull/4060 and with easy options for users, OR a solution that only applied for new projects and does not impact pre-existent ones. **(we need deeply consider the impact for the existent projects)**

camilamacedo86:
c/c @troy0820 @nathanperkins

camilamacedo86:
Hi @troy0820, @nathanperkins, @malt3,

If you'd like to contribute to this effort (since it seems you were the ones who initially requested this change), I suggest starting by creating a PR that updates the go/v4 plugin as a follow-up to https://github.com/kubernetes-sigs/kubebuilder/pull/4060. We need to get that PR merged first to streamline this process.

Once those changes are made and pass the CI/tests, it will be easier to move the updates to a new plugin version if needed. This approach will also allow us to discuss the proposed changes more effectively.

Thank you for your attention and collaboration!

camilamacedo86:
After think about this one, I think it might be better:

- **1 - Create a dir for webhooks**: This will better organize the codebase.
- **2 - Add logic to avoid breaking changes**: We’ll prevent the need for a new Golang plugin version since no significant changes justify it. Maintaining both versions is always a burden for us and is inevitable at some point:

[...]

PR #4130, opened 2024-09-03
✨ (go/v4): improve the webhook tests by adding examples. Also, improve cronjob tutorial to clarify its usage and validate the changes

camilamacedo86:
In this PR we are improving the webhook tests by adding further info and examples for the users. Either we are implementing them further for the cronjob tutorial as an example and to help us to validate and spot issues on related areas when/if we need to change them

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/3454

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook_test.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook_test.go
@@ -18,29 +18,146 @@ package v1
 
 import (
 	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	// TODO (user): Add any additional imports if needed
 )
 
 var _ = Describe("CronJob Webhook", func() {
+	var (
+		obj       *CronJob
+		oldObj    *CronJob
+		validator CronJobCustomValidator
+	)
+
+	BeforeEach(func() {
+		obj = &CronJob{
+			Spec: CronJobSpec{
+				Schedule:                   "*/5 * * * *",
+				ConcurrencyPolicy:          AllowConcurrent,
+				SuccessfulJobsHistoryLimit: new(int32),
+				FailedJobsHistoryLimit:     new(int32),
+			},
+		}
+		*obj.Spec.SuccessfulJobsHistoryLimit = 3
+		*obj.Spec.FailedJobsHistoryLimit = 1
+
+		oldObj = &CronJob{
+			Spec: CronJobSpec{
+				Schedule:                   "*/5 * * * *",
+				ConcurrencyPolicy:          AllowConcurrent,
+				SuccessfulJobsHistoryLimit: new(int32),
+				FailedJobsHistoryLimit:     new(int32),
+			},
+		}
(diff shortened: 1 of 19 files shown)


camilamacedo86:
@vsoch , @ashutosh887, @Sajiyah-Salat, @drewwells

vsoch:
Assuming all the CI passes, this looks very helpful! Thanks @camilamacedo86

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook_test.go
```diff
@@ -18,29 +18,146 @@ package v1
...
+			Expect(*obj.Spec.FailedJobsHistoryLimit).To(Equal(int32(2)), "Expected FailedJobsHistoryLimit to retain its set value")
 		})
 	})
 
-	Context("When creating CronJob under Validating Webhook", func() {
-		It("Should deny if a required field is empty", func() {
+	Context("When creating or updating CronJob under Validating Webhook", func() {
+		It("Should deny creation if the name is too long", func() {
+			obj.ObjectMeta.Name = "this-name-is-way-too-long-and-should-fail-validation-because-it-is-way-too-long"
+			warnings, err := validator.ValidateCreate(ctx, obj)
```
mogsie, line 107: I think you can replace this with a single Error matcher:

```
Expect(validator.ValidateCreate(ctx, obj)).Error().To(MatchError(ContainSubstring("must be no more than 52 characters"))),
	"Expected name validation to fail for a too-long name")
```

Gomega's error validator both matches the error, _and_ that any other responses (`warnings`) are zero-valued.

camilamacedo86: Hi @mogsie 

Thank you a lot for your help on those tests !!!
That is amazing !

Since it is just a nit wdyt about we get this one merged and then you push a follow up just to improve this part? 
Would you like to contribute with the above enhancement?

mogsie: Sure. I can follow up on this.  Should I make a separate issue for it too?

camilamacedo86: Feel free tor raise the issue, but if you will work on that asap. just push the PR is fine.

camilamacedo86:
I am moving forward with this because:

- a) The requestor has already confirmed that it is very helpful.
- b) After the review, the only feedback we received was a minor nit/enhancement suggestion regarding how to perform the check, which, as agreed, will be addressed in a follow-up.
- c) We want to implement tests for the webhooks in the sample within the documentation. This will allow us to add further tests and scenarios, which are necessary since we need to decouple the webhooks from the API to fully comply with the latest changes in controller-runtime. You can find more details here: https://github.com/kubernetes-sigs/kubebuilder/issues/4062 and here: https://github.com/kubernetes-sigs/controller-runtime/issues/2596. These tests will help us validate the changes.

PR #4136, opened 2024-09-08
🌱 cleanup: Replace "-m=64" with "--memory-limit=64"

mogsie:
The memcached image fails to run when "-m=64" is passed to it.

We could either drop it, or replace it with "--memory-limit=64" which works.

When it runs, I get the error message 

``` sh
$ docker run --rm -it memcached:1.6.26-alpine3.19 memcached -m=64 -o modern -v
Cannot set item size limit higher than 1/2 of memory max.
```


camilamacedo86 (approved):
If pass in the tests that is great

/lgtm
/approved

Issue #4135, opened 2024-09-07
Some Gomega assertions are long-winded


mogsie:
### What broke? What's expected?

Kubebuilder creates/scaffolds a lot of test code. This test code uses Ginkgo and Gomega. Some of the tests as written don't take advantage of Gomega's testing framework. Since kubebuilder is likely to be one of those projects that introduce developers to Gomega, I think it is useful if the scaffolded code provides the best possible code samples.

## Avoiding Expect(err)

For example, from test/e2e/v4/plugin_cluster_test.go

``` go
		outputGet, err := kbc.Kubectl.Get(
			false,
			"pods",
			"-n", "kube-system",
			"-l", "k8s-app=calico-node",
			"-o", "jsonpath={.items[*].status.phase}",
		)
		Expect(err).NotTo(HaveOccurred(), "Failed to get Calico pods")
		Expect(outputGet).To(ContainSubstring("Running"), "All Calico pods should be in Running state")
```

adds two variables to the scope, and then checks both responses.  the two assertions could be replaced with an inline call:

``` go
		Expect(kbc.Kubectl.Get(
			false,
			"pods",
			"-n", "kube-system",
			"-l", "k8s-app=calico-node",
			"-o", "jsonpath={.items[*].status.phase}",
		).To(ContainSubstring("Running"), "All Calico pods should be in Running state")
```

This ensures that the error is nil, in addition to checking the output.

## Errors

Similarly, when the output is not checked, the code generally has Expect(err).  This can also be in-lined:

``` go
			_, err = kbc.Kubectl.Command("label", "namespaces", kbc.Kubectl.Namespace,
				"metrics=enabled")
			ExpectWithOffset(2, err).NotTo(HaveOccurred())
```

can be shortened to

``` go
			Expect(kbc.Kubectl.Command("label", "namespaces", kbc.Kubectl.Namespace,
				"metrics=enabled").Error().NotTo(HaveOccurred())
```

The `Error()` call ensures that the last return value (the error) is nil, and ignores the output.

## Eventually

The use of Eventually() often uses a mix of assertion styles because the function passed to Eventually is not passed a Gomega instance:

``` go
		EventuallyWithOffset(1, func() error {
			_, err := kbc.Kubectl.Get(
				true,
				"secrets", "webhook-server-cert")
			return err
		}, time.Minute, time.Second).Should(Succeed())
```

could be simplified to

``` go
		EventuallyWithOffset(1, func(g Gomega) {
[...]

PR #4137, opened 2024-09-08
🌱 cleanup plugin_cluster-test.go

mogsie:
Part of #4135

This commits cleans up "plugin_cluster_test.go" to be less verbose and use better use of Gomega's capabilities.

Notably, variables are used only when the output of some command is used later on in the test.  If we only assert things about the command output, it is generally in-lined.

Eventually functions are converted from `() error` to `(g Gomega)`, and Expects are converted to g.Expect inside the functions.  All offsets have been removed, since this isn't "helper" code, we want to know which line the test failed on.

Finally, the test for verifying available status has been fixed so that it checks the correct pod.

### DIFF
--- a/test/e2e/deployimage/generate_test.go
+++ b/test/e2e/deployimage/generate_test.go
@@ -47,6 +47,7 @@ func creatingAPI(kbc *utils.TestContext) {
 		"--kind", kbc.Kind,
 		"--plugins", "deploy-image/v1-alpha",
 		"--image", "busybox:1.36.1",
+		"--run-as-user", "1001",
 		"--make=false",
 		"--manifests=false",
 	)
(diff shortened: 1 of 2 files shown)


camilamacedo86 (approved):
/lgtm
/approved

camilamacedo86:
HI @mogsie 

This one broke the tests so we need to fix the changes.

camilamacedo86:
HI @mogsie

This one broke the tests so we need to fix the changes.

mogsie:
@camilamacedo86 the test that is failing is a _real failure_.  I think the previous test was masking the problem (it was not expecting correctly, and the tests passed even though they were in error).

The error is that docker / kubernetes does not support the "run as root" default of kubebuilder.  I have added the "--run-as-root" flag to the PR. I had to add this to get my tests to pass, I originally thought it was an artifact of my local kind environment, and not the tests being incorrect.

camilamacedo86 (approved):
/lgtm
/approved

Issue #3878, opened 2024-04-27
Implement e2e controller tests for the multi-version sample used in the docs 


camilamacedo86:
## Description

See that we have the tutorial: https://book.kubebuilder.io/multiversion-tutorial/tutorial
All code used in this tutorial came from the sample in: https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/multiversion-tutorial/testdata/project

This sample is tested via the unit-test when we push a PR with changes.
Then, note that currently the EnvTest tests to test the reconciliation is commented:

https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/multiversion-tutorial/testdata/project/internal/controller/cronjob_controller_test.go#L87-L102

If we uncomment this line we will check in the CI/or when we run `make tests` for this project the following error:

```
"controllerGroup": "batch.tutorial.kubebuilder.io", "controllerKind": "CronJob", "CronJob": {"name":"test-resource","namespace":"default"}, "namespace": "default", "name": "test-resource", "reconcileID": "0504dde2-9a69-4126-829d-cd143fc79f8f", "active jobs": 0, "successful jobs": 0, "failed jobs": 0}
  2024-04-26T07:47:09Z	ERROR	unable to list child Jobs	{"error": "field label \".metadata.controller\" not supported for Job"}
  tutorial.kubebuilder.io/project/internal/controller.(*CronJobReconciler).Reconcile
  	/Users/runner/work/kubebuilder/kubebuilder/docs/book/src/multiversion-tutorial/testdata/project/internal/controller/cronjob_controller.go:132
  tutorial.kubebuilder.io/project/internal/controller.glob..func1.1.3
  	/Users/runner/work/kubebuilder/kubebuilder/docs/book/src/multiversion-tutorial/testdata/project/internal/controller/cronjob_controller_test.go:94
  github.com/onsi/ginkgo/v2/internal.extractBodyFunction.func3
  	/Users/runner/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.14.0/internal/node.go:463
  github.com/onsi/ginkgo/v2/internal.(*Suite).runNode.func3
  	/Users/runner/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.14.0/internal/suite.go:889
```

```
[...]

PR #4139, opened 2024-09-08
:book: update multi-version with the latest changes and address fixes

camilamacedo86:
In this PR we : 
- regenerate the sample using the helper script to ensure that the latest changes are applied
- we compared the multi-version sample with the cronjob and we ensured that they have the same base
- we validate the rendered of the docs to ensure that it is done properly. See that by comparing with the book-v2 we could fix a couple of markers to properly render the storage code:

Screenshot 2024-09-08 at 17 38 59

- fix the issue: https://github.com/kubernetes-sigs/kubebuilder/issues/3878 by ensuring that the tests implemented for cronjob tutorial which is our base is applied here. 
- fix the webhook v2 implementation to use CustomInterfaces and comply with the changes introduced in: https://github.com/kubernetes-sigs/kubebuilder/pull/4060


camilamacedo86:
I will move forward with this one because it is a blocker to allow we move with other initiatives and it mainly does the update of the sample and address required fixes and aligments.
 
However, if anyone see that it requires changes and improvements please feel free to raise a PR against the code for the mulitversion tutorial sample.

PR #4141, opened 2024-09-08
✨ (go/v4): cleanup: in e2e/test.go, make Gomega assertions more concise

mogsie:
Some of the calls to Gomega's assertions are unneccessarily long; declaring variables, and checking both errors and values.  All of this can be done by Gomega itself.  This is part of #4135 

`Expect(func()).To(ContainSubstring("X"))` will 
* check that any err returned from the function is nil
* check that the first return value matches "X"

`Expect(func()).Error().To(HaveOccurred())` will 
* check that any err returned from the function is non-nil
* check that the first return value _is_ nil or zero.


camilamacedo86 (approved):
/lgtm
/approved
/ok-to-test

PR #4142, opened 2024-09-08
📖 cleanup: improves cronjob_controller_test.go gomega assertions

mogsie:
Replaced some bespoke Eventually() signatures (e.g. func() bool and func() (int, error)) with the standard func(g Gomega), allowing gomega assertions to happen inside the Eventually().  I reworded the section just before the Eventually, to explain what it does, in the context of the new signature.

Generally you _expect_ something _to_ happen
and _eventually_, something _should_ happen.

This is a part of #4135.


camilamacedo86 (approved):
/lgtm
/approved
/ok-to-test

Issue #4144, opened 2024-09-08
 Duplicate Webhook Names Generated When Scaffolding Multiple Versions of the Same API


camilamacedo86:
### What broke? What's expected?

When scaffolding multiple versions of the same API with webhooks (e.g., v1 and v2), the generated `MutatingWebhookConfiguration` and `ValidatingWebhookConfiguration` files create duplicate webhook names. Specifically, both versions of the resource (v1 and v2) generate webhooks with the same name (e.g., `mcronjob.kb.io` for mutating webhooks and `vcronjob.kb.io` for validating webhooks).

This leads to a conflict when trying to install these webhooks onto the control plane, resulting in the following error:

```
unable to install webhooks onto control plane: MutatingWebhookConfiguration.admissionregistration.k8s.io "mutating-webhook-configuration" is invalid: webhooks[1].name: Duplicate value: "mcronjob.kb.io"
```

Kubebuilder should generate unique webhook names for each version of the API. For example:
- `mcronjob-v1.kb.io` for v1 mutating webhook
- `mcronjob-v2.kb.io` for v2 mutating webhook

This ensures that both versions of the resource can have their webhooks installed without conflicts.

### Reproducing this issue

1. Scaffold an API with webhooks for version v1:
    ```bash
    kubebuilder create api --group batch --version v1 --kind CronJob --resource=true --controller=true
    kubebuilder create webhook --group batch --version v1 --kind CronJob --defaulting --programmatic-validation
    ```

2. Scaffold the same API for version v2:
    ```bash
    kubebuilder create api --group batch --version v2 --kind CronJob --resource=true --controller=true
    kubebuilder create webhook --group batch --version v2 --kind CronJob --defaulting --programmatic-validation
    ```

3. Run `make manifests` or `controller-gen` to generate the webhook configurations.

4. You will notice that both v1 and v2 have webhooks with the same name (`mcronjob.kb.io` for mutating and `vcronjob.kb.io` for validating).

5. Attempting to install these webhooks (e.g., using `kubectl apply` or running tests) will result in the error: `Duplicate value: "mcronjob.kb.io"`.

### KubeBuilder (CLI) Version

4.2.0

PR #4145, opened 2024-09-08
🐛 (go/v4) resolve duplicate webhook name issue in multi-version API

camilamacedo86:
Adjust webhook names for mutating and validating webhooks to avoid conflicts when creating multiple versions of the same API. 

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/4144

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
@@ -75,6 +75,8 @@ A webhook will automatically be served that calls this defaulting.
 The `Default` method is expected to mutate the receiver, setting the defaults.
 */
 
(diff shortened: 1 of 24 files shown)


docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook.go
```diff
@@ -75,6 +75,8 @@ A webhook will automatically be served that calls this defaulting.
 The `Default` method is expected to mutate the receiver, setting the defaults.
 */
 
+// +kubebuilder:webhook:path=/mutate-batch-tutorial-kubebuilder-io-v1-cronjob,mutating=true,failurePolicy=fail,sideEffects=None,groups=batch.tutorial.kubebuilder.io,resources=cronjobs,verbs=create;update,versions=v1,name=mcronjob-v1.kb.io,admissionReviewVersions=v1
+
```
mogsie, line 79: A little newline bug here?  Shouldn't the +kubebuilder directives be togeher?

camilamacedo86: One marker is for the Mutating Webhook, and the other is for the Validating Webhook. Each webhook is scaffolded with its respective code—one for mutation and the other for validation. So, it's not a bug.

mogsie (approved):

PR #4151, opened 2024-09-09
🐛 Enhances the webhook test template.

### LINKED ISSUE #4148: Enhance webhook_test template

We need to address this in the scaffold. See: [webhook_test_template.go](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/api/webhook_test_template.go).

More information can be found in the comments raised in [this PR](https://github.com/kubernetes-sigs/kubebuilder/pull/4138).

### Odd Indentation:

```
// It("Should apply defaults when a required field is empty", func() {
		//     By("simulating a scenario where defaults should be applied")
		// 	   obj.SomeFieldWithDefault = ""
```

###  Improvements to gomega usage:
- Replace :  Expect(err).NotTo(HaveOccurred())
- With: Expect(obj.Default(ctx)).To(Succeed())

- Replace:  warnings, err := obj.ValidateCreate(ctx)
- With:  Expect(obj.ValidateCreate(ctx)).Error().To(HaveOccurred())

See that we will need to either change the contents that we find/replace to generate the docs samples under the hack/docs.

### KubeBuilder (CLI) Version

master

### PROJECT version

_No response_

### Plugin versions

_No response_

### Other versions

_No response_

### Extra Labels

_No response_

mogsie:
The webhook scaffold had some oddly indented regions, resulting in the generated source code having a mix of tabs and spaces, and was generally looking quite ugly.

The gomega asserions are too chatty; errors, and return values are checked after first being assigned as variables in scope.  This is not needed, and Gomega checks all of this for us:

* Expect(someFunc()).Error().To(HaveOccurred()) checks both the error, but also that the other return values are zero or nil.
* Expect(someFunc()).To(ContainSubstring("X")) checks both the return value _and_ verifies that the error is nil.

Fixes #4148.  Part of #4135.

- Fixes odd indentation, now it is in line with the other scaffolds
- Simplifies gomega assertions, to use fewer statements
- Updated the cronjob tutorial's tests too

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook_test.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_webhook_test.go
@@ -104,32 +104,28 @@ var _ = Describe("CronJob Webhook", func() {
 	Context("When creating or updating CronJob under Validating Webhook", func() {
 		It("Should deny creation if the name is too long", func() {
 			obj.ObjectMeta.Name = "this-name-is-way-too-long-and-should-fail-validation-because-it-is-way-too-long"
-			warnings, err := validator.ValidateCreate(ctx, obj)
-			Expect(err).To(HaveOccurred(), "Expected name validation to fail for a too-long name")
-			Expect(warnings).To(BeNil())
-			Expect(err.Error()).To(ContainSubstring("must be no more than 52 characters"))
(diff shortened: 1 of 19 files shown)


camilamacedo86 (approved):
Thank you 🥇 
It is great!

/lgtm
/approved
/ok-to-test

PR #4154, opened 2024-09-10
🌱 ci:  add e2e tests for multigroup testdata layout and rename sample

camilamacedo86:
We need to rename the name of the sample because of:
    
```shell
The Service "project-v4-multigroup-with-plugins-controller-manager-metrics-service" is invalid: metadata.name: Invalid value: "project-v4-multigroup-with-plugins-controller-manager-metrics-service": must be no more than 63 characters
```
    
Therefore, in this we are:
- Creating jobs to run the e2e tests for all testdata samples in parallel
- Renaming the sample project-v4-multigroup-with-plugins to project-v4-multigroup to generate manifests names that can be accepted in k8s

PS.: See that it help us for example, found the issue: https://github.com/kubernetes-sigs/kubebuilder/pull/4162


camilamacedo86:
I am moving forward because it is a very small nit that blocks other initiatives.
Passing in the e2e tests is more than enough for we ensure this one.
So, I will not make nobody waste time on reviews. But if you have suggestions then, please feel free to raise a follow up PR.

camilamacedo86:
This one only add the tests so that we can ensure the default layout and avoid issues like https://github.com/kubernetes-sigs/kubebuilder/issues/4161

So, it only need to pass in the e2e tests to be checked, if anyone has a hall for improvements here please feel free to raise a follow up PR but I will not make nobody waste time on reviews of this one and I am moving forward.

Either it will support follow up initiatives for we ensure that all is fine.

PR #4155, opened 2024-09-10
📖 fix marker for webhook server in the tutorial samples and position of explanations

camilamacedo86:
Apply the fix introduced in PR #4122. The default markers in the code have been replaced with the values defined in the hack/docs, but this change was missed in the tutorial samples.


camilamacedo86:
Small fix for docs, but that blocks further initiatives. 

I will not make nobody waste time in reviews.
However, if you would like to change anything here and improve it please feel free and raise a PR with your suggestions
Your help is welcome !!

Moving forward.

PR #4158, opened 2024-09-11
:bug: (go/v4): Refactor e2e-tests for clearer error handling and readable logs

camilamacedo86:
Refactors all test cases that use g.Expect(utils.Run(cmd)) to improve both the readability of logs and the clarity of the code.

Changes:

- Replaced occurrences of g.Expect(utils.Run(cmd)) with:

```go
output, err := utils.Run(cmd)
g.Expect(err).NotTo(HaveOccurred())
g.Expect(string(output)).To(<condition>)
```

OR

```go
_, err := utils.Run(cmd)
g.Expect(err).NotTo(HaveOccurred())
```

**Motivation**

- **Human-readable logs:** Output is now converted to a string, providing more understandable logs by displaying actual kubectl command results instead of raw byte arrays.
- **Improved code clarity:** The previous `g.Expect(utils.Run(cmd))` usage did not make it clear that the function returns both output and error. By explicitly handling the output and err variables, the code is more transparent and easier to maintain.

**Example of problematic scenario**

Otherwise, we are unable to check the output when errors are faced. For example

Before the changes:
Example: https://github.com/kubernetes-sigs/kubebuilder/actions/runs/10807585375/job/29978696165

```sh
  [FAILED] Timed out after 120.001s.
  The function passed to Eventually failed at /home/runner/work/kubebuilder/kubebuilder/testdata/project-v4-multigroup/test/e2e/e2e_test.go:145 with:
  Metrics endpoint is not ready
  Expected
[...]

PR #4159, opened 2024-09-11
🐛 (go/v4): Refactor e2e-tests for readable logs

mogsie:
My recent PR #4141 introduced a problem with error reporting.  Previously, error resporting would contain the output of the command that it was matching against (e.g. a substring).  After #4141, a byte array was being printed out, making it unreadable.  This PR fixes that, making the error output readable again.

utils.go has a function to execute a command, and it returns the output if the command is successful. It returns a byte array, the same as the underlying exec.Cmd() provides. This, being a test heper function would make sense to return a string instead, so that the ergonomics of using Gomega assertions on strings work better by default.

Removed the now unneccesary cast from []byte to string.

Shamelessly stole the title from #4158.

Closes #4158
Part of #4135

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
@@ -41,7 +41,7 @@ func warnError(err error) {
 }
 
 // Run executes the provided command within this context
-func Run(cmd *exec.Cmd) ([]byte, error) {
+func Run(cmd *exec.Cmd) (string, error) {
 	dir, _ := GetProjectDir()
 	cmd.Dir = dir
 
@@ -54,10 +54,10 @@ func Run(cmd *exec.Cmd) ([]byte, error) {
 	_, _ = fmt.Fprintf(GinkgoWriter, "running: %s\n", command)
 	output, err := cmd.CombinedOutput()
 	if err != nil {
-		return output, fmt.Errorf("%s failed with error: (%v) %s", command, err, string(output))
+		return "", fmt.Errorf("%s failed with error: (%v) %s", command, err, string(output))
 	}
 
-	return output, nil
+	return string(output), nil
 }
 
 // InstallPrometheusOperator installs the prometheus Operator to be used to export the enabled metrics.
@@ -92,7 +92,7 @@ func IsPrometheusCRDsInstalled() bool {
 	if err != nil {
 		return false
 	}
-	crdList := GetNonEmptyLines(string(output))
+	crdList := GetNonEmptyLines(output)
 	for _, crd := range prometheusCRDs {
 		for _, line := range crdList {
 			if strings.Contains(line, crd) {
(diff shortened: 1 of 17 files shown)


docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
```diff
@@ -54,10 +54,10 @@ func Run(cmd *exec.Cmd) ([]byte, error) {
 	_, _ = fmt.Fprintf(GinkgoWriter, "running: %s\n", command)
 	output, err := cmd.CombinedOutput()
 	if err != nil {
-		return output, fmt.Errorf("%s failed with error: (%v) %s", command, err, string(output))
+		return "", fmt.Errorf("%s failed with error: (%v) %s", command, err, string(output))
```
camilamacedo86, line 57: Then, we are no longer return the output when we we should here.
We need return the output.

docs/book/src/getting-started/testdata/project/test/utils/utils.go
```diff
@@ -54,10 +54,10 @@ func Run(cmd *exec.Cmd) ([]byte, error) {
 	_, _ = fmt.Fprintf(GinkgoWriter, "running: %s\n", command)
 	output, err := cmd.CombinedOutput()
 	if err != nil {
-		return output, fmt.Errorf("%s failed with error: (%v) %s", command, err, string(output))
+		return "", fmt.Errorf("%s failed with error: (%v) %s", command, err, string(output))
```
camilamacedo86, line 57: we should return the output always.

docs/book/src/multiversion-tutorial/testdata/project/test/e2e/e2e_test.go
```diff
@@ -110,7 +110,7 @@ var _ = Describe("Manager", Ordered, func() {
 					"pods", controllerPodName, "-o", "jsonpath={.status.phase}",
 					"-n", namespace,
 				)
-				g.Expect(utils.Run(cmd)).To(BeEquivalentTo("Running"), "Incorrect controller-manager pod status")
+				g.Expect(utils.Run(cmd)).To(Equal("Running"), "Incorrect controller-manager pod status")
```
camilamacedo86, line 113: Just for curiosity,
what is the diff between `BeEquivalentTo ` and `Equal` do you know?

docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
mogsie: The output is in the error message, though.  Generally the return values are zero-valued, _or_ the error is nil.  This is what Gomega checks, and so I replaced it with "" to adhere to the go idioms.

docs/book/src/multiversion-tutorial/testdata/project/test/e2e/e2e_test.go
mogsie: Equal checks for equality while BeEquivalentTo does some type conversion, like between string and []byte.  Unfortunately, it still prints out the byte array representation.   Equals doesn't allow a byte array to be compared to a string.

Issue #4161, opened 2024-09-12
Multigroup Layout Fails with Name Conflicts Across Different Groups/Versions


camilamacedo86:
### What broke? What's expected?

 In the multigroup layout, when multiple kinds have the same name across different groups or versions, Kubebuilder fails to register the controllers due to name conflicts. For example, if two kinds with the same name exist in different groups or versions, the following errors occur:

```shell
ERROR setup unable to create controller {"controller": "Bar", "error": "controller with name bar already exists. Controller names must be unique to avoid multiple controllers reporting to the same metric"}
```

### Reproducing this issue

- Create a multigroup project with resources that share the same kind name but belong to different API groups or versions.
- Attempt to run the project.
- Observe the errors due to conflicting controller names.

### KubeBuilder (CLI) Version

4.2.0

PR #4162, opened 2024-09-12
:bug: fix: ensure unique controller names to fix name conflicts in multigroup setup

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/4161


camilamacedo86:
It is a small fix so, I will not make nobody waste time on review. 
Also, it blocks https://github.com/kubernetes-sigs/kubebuilder/pull/4154 where we can validate that the fix works

camilamacedo86:
It is a small bug fix which blocks we validate the multigroup layout where the bug was found https://github.com/kubernetes-sigs/kubebuilder/pull/4154 so that we can prevent things like that. It also blocks other initiatives since we want ensure that all possible scenarios are well tested before decouple the webhooks from apis and simplify/fix the kustomize scaffolds for cert-manager: https://github.com/kubernetes-sigs/kubebuilder/pull/4123

So, I will not make nobody waste time on this one. we have more big fishes to fry.
Moving forward since the changes are indeed either validate via e2e tests already.

PR #4165, opened 2024-09-14
📖 Simplify event recording example

rkosegi:
There is no need for `fmt.Sprintf` since `EventRecorder` already has `Eventf` method that accepts formatting string


camilamacedo86 (approved):

/lgtm
/approved
/ok-to-test

Issue #4149, opened 2024-09-09
Add E2E Tests for Applying CR and Webhooks in Non-Default Namespaces


camilamacedo86:
### What do you want to happen?

We need to add e2e tests to validate that Custom Resources (CR) and webhooks work correctly when applied in namespaces other than the default ones where the webhook server and manager are deployed. This will help ensure that everything continues to function as expected in different namespace configurations.

**Proposed Approach:**
- We supplement the tests for deploy-image: https://github.com/kubernetes-sigs/kubebuilder/tree/master/test/e2e/deployimage
- So that we keep it simple since the go/v4 tests are ensuring protection with network policies and a lot of other scenarios already, see: https://github.com/kubernetes-sigs/kubebuilder/tree/master/test/e2e/v4
- So that we can validate the metrics exposure and ensure how the deploy image is working with webhooks, in addition to confirming the primary goal: to ensure that reconciliation and webhook operations continue to function correctly when applied in a different namespace.

camilamacedo86:
Why to do that?
It might help us ensure better changes like: https://github.com/kubernetes-sigs/kubebuilder/pull/4123

PR #4167, opened 2024-09-14
🌱 e2e-tests: fix/cleanup and add test to validate webhooks outside of the manager namespace

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/4149


camilamacedo86:
I will move with this one because it is just a silly/small fix/cleanup and addition
I will not make nobody waste time on the reviews and that will help us to better ensure https://github.com/kubernetes-sigs/kubebuilder/pull/4123

PR #4169, opened 2024-09-15
:book: remove from getting started the link for watching resources since this doc acctually document predicates

camilamacedo86:
The documentation is actually trying to explain predicates.
Therefore, should not be linked in the Getting Started tutorial.


camilamacedo86:
Just a nit moving forward.

Issue #4109, opened 2024-08-27
Review Watching Resources Docs


camilamacedo86:
### What broke? What's expected?

We need to ensure that the doc is using nice and common terms and is follow the standards
The motivation for this one raise from: https://kubernetes.slack.com/archives/CAR30FCJZ/p1724708630663269

We need to review the whole section and mainly https://book.kubebuilder.io/reference/watching-resources/externally-managed

### KubeBuilder (CLI) Version

master

PR #4170, opened 2024-09-15
📖 Revamp "Watching Resources" documentation for accuracy and clarifty

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/4109


camilamacedo86:
Hi @mogsie 

Thank you a lot for raise your concern.
It is all fixed now, feel free to give a look and your help on review will be well appreciated.

docs/book/src/reference/watching-resources/secondary-resources-owned-by-other-controllers.md
```diff
@@ -0,0 +1,78 @@
+# Watching Secondary Resources Owned by Other Controllers or Projects
+
+In some cases, a controller may need to watch and react to changes in resources
+that are not `Owned` by it, but are created and managed by another controller.
+
```
mogsie, line 5: Here, I think it would be useful to list common use cases:

* when the primary resource *references* something
* when the primary object is named after some other object (?)

mogsie (changes requested):

docs/book/src/reference/watching-resources/secondary-resources-owned-by-other-controllers.md
```diff
@@ -0,0 +1,78 @@
...
+    return ctrl.NewControllerManagedBy(mgr).
+        For(&examplecomv1alpha1.BackupBusybox{}).  // Watch the primary resource (BackupBusybox)
+        Watches(
+            &source.Kind{Type: &examplecomv1alpha1.Busybox{}},  // Watch the Busybox CR
+            handler.EnqueueRequestsFromMapFunc(func(obj client.Object) []reconcile.Request {
+                // Trigger reconciliation for the BackupBusybox in the same namespace
+                return []reconcile.Request{
+                    {
+                        NamespacedName: types.NamespacedName{
+                            Name:      "backupbusybox",  // Reconcile the associated BackupBusybox resource
```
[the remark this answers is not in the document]
camilamacedo86: In the default scaffold now we add the name. Why?
If we have 2 versions of the same CR such as we have in the tutorial and a controller for each one of them then it will not work out. Also, if we are using the multigroup layout and then, we have the same Kind name and its controllers but for diff groups the same issue will occurs. 

See: https://github.com/kubernetes-sigs/kubebuilder/pull/4162

docs/book/src/reference/watching-resources.md
```diff
@@ -1,16 +1,92 @@
...
+
+Additionally, if the **Primary Resource** is deleted, Kubernetes' garbage collection
+mechanism ensures that all associated **Secondary Resources** are automatically deleted in a cascading manner.
+
+<aside class="note">
+<h1> Watching External Types (API/CRD Defined by Another Operator) </h1>
+
+In Kubebuilder, resources can be categorized into two types:
+- **Core Types**: API types defined by Kubernetes itself, such as `Pods`, `Services`, and `Deployments`.
+- **External Types**: API types defined in other projects, such as CRDs managed by another operator.
```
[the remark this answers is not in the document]
camilamacedo86: Primary and Secondary are the technical terms adopt across the community  so we need to stick with them
Then Core and External types are the kubebuilder terms to explain the k8s kinds which are registered by default and the third-part/external ones. 

I am not sure if I followed all your comments but I will re-read the text here and see.

docs/book/src/reference/watching-resources/secondary-resources-owned-by-other-controllers.md
camilamacedo86: We kind of explain that in another place.
We have the Primary Resource , everything that is required to that works on the cluster would be a Secondary resource .. but I will give a look and see if I can improve with your suggestions thx.

docs/book/src/reference/watching-resources/secondary-resources-owned-by-other-controllers.md
```diff
@@ -0,0 +1,78 @@
+# Watching Secondary Resources Owned by Other Controllers or Projects
+
+In some cases, a controller may need to watch and react to changes in resources
+that are not `Owned` by it, but are created and managed by another controller.
+
+The examples below demonstrate how a controller can watch and reconcile resources that it does not directly own.
+This applies to both Kubernetes core API resources (like `ConfigMaps`, `Services`, or `Deployments`) and
+custom resources defined by other projects (such as CRDs from other operators).
```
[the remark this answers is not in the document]
camilamacedo86: done

docs/book/src/reference/watching-resources.md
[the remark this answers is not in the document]
camilamacedo86: Revamp all so I am closing this one but please feel free to check if you see that it requires further improvements,

docs/book/src/reference/watching-resources.md
```diff
@@ -1,16 +1,177 @@
...
+
+Whether a resource is defined within your project or comes from an external project, the concept of **Primary**
+and **Secondary Resources** remains the same:
+- The **Primary Resource** is the resource the controller is primarily responsible for managing.
+- **Secondary Resources** are those that are required to ensure the primary resource works as desired.
+
+Therefore, regardless of whether the resource was defined by your project or by another project,
+your controller can watch, reconcile, and manage changes to these resources as needed.
+
+## Why watch the secondary resources matters?
```
mogsie, line 103: There is a grammatical error here.

```suggestion
## Why does watching the secondary resources matter?
```

```suggestion
## Why watching the secondary resources matters
```

```suggestion
## Why watch the secondary resources?
```

docs/book/src/reference/watching-resources.md
```diff
@@ -1,16 +1,177 @@
...
+Failing to track these resources can lead to inconsistencies in your
+controller's behavior and the overall cluster state.
+
+Secondary resources may not be directly managed by your controller,
+but changes to these resources can still significantly
+impact the primary resource and your controller's functionality.
+Here are the key reasons why it's important to watch them:
+
+- **Ensuring Consistency**:
+    - Secondary resources (e.g., child objects or external dependencies) may diverge from their expected state.
```
mogsie, line 116:
```suggestion
    - Secondary resources (e.g., child objects or external dependencies) may diverge from their last seen state.
```

... or "last reconciled state" or "last known state" — the state is not expected

docs/book/src/reference/watching-resources.md
```diff
@@ -1,16 +1,177 @@
...
+    For instance, a secondary resource may be modified or deleted, causing the system to fall out of sync.
+    - Watching secondary resources ensures that any changes are detected immediately, allowing the controller to
+    reconcile and restore the desired state.
```
mogsie, line 119:
```suggestion
    reconcile and restore consistency.
```

docs/book/src/SUMMARY.md
```diff
@@ -76,8 +76,9 @@
   - [Good Practices](./reference/good-practices.md)
   - [Raising Events](./reference/raising-events.md)
   - [Watching Resources](./reference/watching-resources.md)
-    - [Resources Managed by the Operator](./reference/watching-resources/operator-managed.md)
-    - [Externally Managed Resources](./reference/watching-resources/externally-managed.md)
+    - [Owned Resources](./reference/watching-resources/secondary-owned-resources.md)
+    - [Not Owned Resources](./reference/watching-resources/secondary-resources-not-owned)
```
mogsie, line 80: Should this be .md? it isn't rendering correctly in the preview.
```suggestion
    - [Not Owned Resources](./reference/watching-resources/secondary-resources-not-owned.md)
```

docs/book/src/reference/watching-resources/secondary-resources-not-owned.md
mogsie, line 1: At the bottom of this file, I would insert a sentence about how it's common to use this to watch referenced objects, and how EnqueueRequestsFromMapFunc works.

e.g. something like:

A common pattern is to use this when a primary resource contains *links* to secondary resources.  for example, a `BusyBox` might reference a "script" `ConfigMap`.  If the `ConfigMap` changes, the `BusyBox` instance should be reconciled.

The `EnqueueRequestsFromMapFunc` is called whenever a potentially linked Secondary resource (here, the `ConfigMap`) is changed somehow.  The `EnqueueRequestsFromMapFunc` needs to find all **Primary** resources (`BusyBox`) that *reference* it.  The function passed in to `EnqueueRequestsFromMapFunc` typically works like this:

* Let's say a Secondary object (e.g. a `ConfigMap`) has been modified.
* The function should return all of the Primary resources (`BusyBox`) that *link to that `ConfigMap`*.
* To do this it lists all Primary (`BusyBox`) resources in the same namespace (for security purposes, references should not cross namespaces).
* For each Primary object, it checks in their spec for the reference to this `ConfigMap` object.
* If the primary object does link to the given secondary object, the primary object's name is returned.

The result is then that whenever an object links to another object, and the linked object is modified, then the objects that _link_ to it will be reconciled.

mogsie (approved):
This is great!  Very clear explanation of the different terms and it's all much more consistent.  A few minor nits, and one bigger suggestions, but all in all awesome! :sparkles:

docs/book/src/reference/watching-resources.md
camilamacedo86: the righ technical common term is desired state. I will change for that

docs/book/src/reference/watching-resources.md
camilamacedo86: The  the desired state is the common. technical term so if you do not mind I will keep this one.

docs/book/src/reference/watching-resources/secondary-resources-not-owned.md
camilamacedo86: Hun, acctually how it works is 
- If the resource is OWNED / CREATED by the controller then it should be marked with `controllerutil.SetControllerReference` and we should watch with `Owned`
- If the resource is NOT Owned then we need to use `EnqueueRequestsFromMapFunc ` to watch

mogsie: This whole section is about your second bullet point.  The ConfigMap in the above example is a **referenced** (not owned) resource (i.e. it is found somewhere in the primary object's spec).  I am just suggesting that this (common?) use case is mentioned.

camilamacedo86: In the document "How to watch resources not owned", we're using another CRD as an example. However, we also mention that a Core type can be used instead.

Why use another CRD? This is because one of the most common questions we receive is: "Can I use CRDs from one project in another?". By using a CRD in the example, we address this frequent concern directly. Additionally, the example of watching owned resources already covers Core types, so it makes sense to diversify the examples.

I'll go ahead and merge this. If anyone (or you) has suggestions for improvements, feel free to open a PR with your ideas. It sometimes is easier then try to explain the idea. I hope this approach works better. 

thank you so much for your review.

Issue #1999, opened 2021-02-09
Handle external API types explicitly in `create api`


estroz:
I want to create a controller for an external type, ex. `core/v1.Pod` or `route.openshift.io/v1.Route`, and have kubebuilder be aware that this type is external (for commands I run after `create api`).

Currently external types are supported, but are treated as custom types, which can cause problems with scaffolding controller files. Additionally the edge case of the "core" group, which is "" in actual GVKs, isn't handled well.

From #1772:

Regarding creating a new command: The most common use case by far (anecdotal evidence only) is creating both a controller and resource. I do sometimes see one or the other (or neither) being created to support external types, so having the choice of either or neither mode is still useful. Having a command for each mode confounds the currently straightforward workflow. Therefore I vote to leave the `create api` command as-is, except invert flag defaults to `true`.

Regarding the extra flag on `create api`: Having an extra flag makes sense, since we need to know _where_ the external types are coming from as they aren't in `api/<version>` or `apis/<group>/<version>`. I like @camilamacedo86's idea of supplying a module/import path, perhaps with a more descriptive flag like `--external-import`. `kubebuilder create api` shouldn't be doing any `go get` stuff though; that'll be done by calling make targets. The input would also have to describe the exact import path, and `--group` would have to be the full API group:
```sh
kubebuiler create api \
	--external-import=github.com/openshift/api/route/v1@v0.18.2 \
	--group=route.openshift.io \
	--kind=Route \
	--version=v1
```
We might also want to add the resource to `resources` in the config with an `external: <boolean>` qualifier.

Related: #1975 (loosely), #1772 

/kind feature

camilamacedo86:
I'd like to make a few considerations over this one: 

> Handle external API types explicitly in `create api`

This issue ought to be we support external types across the project.  So, in POV the command `create webhook` would also accept:  

```
kubebuiler create webhook \
	--external-import=github.com/openshift/api/route/v1@v0.18.2 \
	--group=route \
        --domain= openshift.io \ 
	--kind=Route \
	--version=v1
```

Where the webhook would be created and the PROJECT file will have the resource info without api and a controller. 

> Currently external types are supported, but are treated as custom types,

Kubebuilder only supports core types and the APIs scaffolded in the project by default unless you manually change the files you will be unable to work with external-types. I mean, it is not really supported by the tool since users are able to do that only if they customize the scaffolds manually via workarrounds. 

> ```
kubebuiler create api \
[...]

PR #4150, opened 2024-09-09
:warning:  (go/v4) decouple webhooks from APIs (Move Webhooks from `api/<version>` or `api/<group>/<version>` to `internal/webhook/<version>` or `internal/webhook/<group>/<version>`)

camilamacedo86:
This PR decouples the webhooks from the API, aligning with the recent breaking changes introduced in controller-runtime to ensure that kubebuilder still compatbile with its next release.  Webhooks are now scaffolded under `internal/webhook` to comply with the latest standards.

**Context:**

Controller-runtime deprecated and removed the webhook methods in favor of CustomInterfaces (see [controller-runtime#2641](https://github.com/kubernetes-sigs/controller-runtime/issues/2641)). The motivation for this change is outlined in [controller-runtime#2596](https://github.com/kubernetes-sigs/controller-runtime/issues/2596).

See that the current master branch already reflects these changes, using the CustomInterfaces: [kubebuilder#4060](https://github.com/kubernetes-sigs/kubebuilder/pull/4060).

**Changes:**

- Webhooks are now scaffolded in `internal/webhook/<version>` or `internal/webhook/<group>/<version>`.
- However, to ensure backwards compatibility, a new `--legacy` flag is introduced. Running `kubebuilder create webhook [options] --legacy` will scaffold webhooks in the legacy location for projects that need to retain the old structure. However, users will still to address the breaking changes in the source code by replacing the old methods by the new CustomInterfaces.  

Example:

**(single group / default layout)**
```sh
$ tree
.
...
├── api
│   └── v1
│       ├── admiral_types.go
│       ├── captain_types.go
│       ├── firstmate_types.go
│       ├── groupversion_info.go
│       └── zz_generated.deepcopy.go
...
├── internal
│   ├── controller
...
│   └── webhook
│       └── v1
│           ├── admiral_webhook.go
│           ├── admiral_webhook_test.go
│           ├── firstmate_webhook.go
│           ├── firstmate_webhook_test.go
│           └── webhook_suite_test.go
...

```

**Notes:**

**If you do not want to change the layout (move your webhooks to internal/webhook then,**  ) then use the flag `--legacy` and ensure that you upgrade the source code to use the CustomInterfaces as the following examples. However, be aware that the `legacy` path is deprecated and will no longer used in future versions of the Golang Plugins (i.e. go/v5):
[...]

PR #4171, opened 2024-09-15
✨ Add support to scaffold controllers for External Types

camilamacedo86:
Introduces the option to allow users scaffold controllers for external types by running:

Example

```    
kubebuilder create api --group certmanager --version v1 --kind Certificate --controller=true --resource=false --make=false --external-api-path=github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1 --external-api-domain=cert-manager.io
```

Partial Close: https://github.com/kubernetes-sigs/kubebuilder/issues/1999
We need to either add the support for Core and External types for webhooks which is blocked by: https://github.com/kubernetes-sigs/kubebuilder/pull/4150


camilamacedo86 (commented):
Small bug fix I will not make nobody waste time on this review
when we have a lot of open prs which really needs help to be reviwed.

Moving forward.

camilamacedo86:
I need to check why it does not work with multigroup yet

camilamacedo86:
Hi @fischor, @sundaram2021,

See the changes here.

We need to add the support for controllers first, ensure the docs, change in the PROJECT config and re-scaffold helper first for then we be able to work on with those sample supports for webhooks in a follow up.

So, please feel free to help us in the review of this one for we are able to looking for in the next step.

camilamacedo86:
Hi

@hardikdr, @kopiczko, @janik-io, @oliviermichaelis, @ctrought, @armsnyder, @samuelvl, @allenhaozi, @wangjiusi

I know that you were looking for that.
The follow up here is add the same to the webhooks.
But we need to move forward step by step and we have another PR that blocks us to 
do the changes in the webhooks at the moment.

So, if you be able to help us with reviews that will be great.

pkg/plugins/golang/v4/scaffolds/internal/templates/main.go
```diff
@@ -121,12 +127,54 @@ const (
 			os.Exit(1)
 		}
 	}
+`
+
+	checkAPIKindExistsCodeFragment = `checkAPIKindExists("%s")
+`
+
+	// nolint:lll
+	checkAPIKindExistsMethodCodeFragment = `// checkAPIKindExists checks if a specific kind is available on the cluster across all groups and versions.
```
fischor, line 136: I would not make use of this method.

Exiting is not good practice I would argue, as this will likely just cause the Pod to be recreated and then crashed again (ChrashLoop).

I think the reconciler will log errors automatically when it should watch a non-existing resource. Having these logs should be sufficient and feels like the way to go here:
https://github.com/kubernetes-sigs/controller-runtime/blob/659afe993be9b2b5c67be7d3e1d20397a91061ce/pkg/internal/source/kind.go#L68-L72

docs/book/src/reference/using_an_external_resource.md
```diff
@@ -0,0 +1,101 @@
...
+This scaffolds a controller for the Core type `corev1.Pod` but skips creating new resource
+definitions since the type is already defined in the Kubernetes API.
+
+
+<aside class="note">
+<h1>Scheme Registry</h1>
+
+The CLI will not automatically register or add schemes for Core Types because these types are already included by default in Kubernetes. Therefore, you don't need to manually register the scheme for core types, as they are inherently supported.
+
+For External Types, the schemas will be added during scaffolding when you create a controller for an external type. The import path provided during scaffolding will be used to register the external type's schema in your project’s `main.go` file and in the `suite_test.go` file.
```
fischor, line 88: typo: schemas -> schemes (2x)

docs/book/src/reference/markers/scaffold.md
```diff
@@ -95,17 +95,19 @@ properly registered with the manager, so that the controller can reconcile the r
...
+| `+kubebuilder:scaffold:crdkustomizewebhookpatch` | `config/crd`                 | Marks where CRD webhook patches are added.                                                                                        |
+| `+kubebuilder:scaffold:crdkustomizecainjectionpatch` | `config/crd`                 | Marks where CA injection patches are added for the webhook.                                                                       |
+| `+kubebuilder:scaffold:manifestskustomizesamples` | `config/samples`             | Marks where Kustomize sample manifests are injected.                                                                              |
+| `+kubebuilder:scaffold:e2e-webhooks-checks` | `test/e2e`                   | Adds e2e checks for webhooks depending on the types of webhooks scaffolded.                                                       |
+| `+kubebuilder:scaffold:check-external-api` | `main.go`                    | If an controller be scaffolded for an External Type then, adds a new check for the API to ensure that the api and version exists. |
```
fischor, line 109: Typos: --> If a controller is scaffolded 
api --> API (maybe just use GVK??)

docs/book/src/reference/markers/scaffold.md
```diff
@@ -95,17 +95,19 @@ properly registered with the manager, so that the controller can reconcile the r
...
+| `+kubebuilder:scaffold:add-method-check-external-api` | `main.go`                    | If an controller be scaffolded for an External Type then, adds the method to do verify if the API/version exist on the cluster.   |
```
fischor, line 110: Same typo here

fischor (commented):
Hi @camilamacedo86,

here my review and thoughts:

**1. An additionally `--domain` flag is needed**

Currently always the domain as defined in the PROJECT.
This causes issues wherever this Domain is used in conjuction with the external resource, e.g. the `kubebuilder:rbac` markers that are scaffolded in the controller Go files.
To resolve this I think its necessary to add an additional `--domain` flag that needs to be set whenever the `--external-api-path` flag is set.
I thought about defaulting this  `--domain` flag to the domain part of the `--external-api-path` for easier use of the CLI but that almost never works out these "domains" will likely differ.

**2. Suggestion: the `--external-api-path` flag should be defined alongside the other resource flags and not as part of the `createApiSubcommand` options** 

With that, all subcommands and plugins that work with resources could make use of the flags for external resource.
Otherwise, if we add to the  `createApiSubcommand` we would need to add the same flag (+ potentially the `--domain` flag) to the  `createWebhookSubcommand` again. 
Plugin authors would need to do the same.

This would be the place to add it I think:
https://github.com/camilamacedo86/kubebuilder/blob/87a480182385aa3ba3f92f3b0aa2cbc8e8251f50/pkg/cli/resource.go#L39-L47

But since for this PR the goal is to "Add support to scaffold controllers for External Types" having it on the `createApiSubcommand` is sufficient. We maybe can then move it in a later PR if we decide to go that route.

Have you thought about alternative names for this flag? I would probably just name it `path` to align with the `resource.path` value in the PROJECT file and to have a short name.

**3. Thought: The distinction between external and core types could be removed**

It's not strictly necessary anymore, as we have ways to specify that the API path as `k8s.io/...`.
Pros having it:
- its automatically detected if its a core type and then the user does not need to specify the `--external-api-path` flag
Cons having it:
- Its hard to maintain also, since new groups can be added
- there is now an exception that needs to be handled and documented

pkg/plugins/golang/v4/scaffolds/internal/templates/main.go
camilamacedo86: Using the Kubernetes Discovery API in an Operator for external APIs ensures the external CRD is installed before the controller starts, preventing runtime errors. It dynamically checks for the required resources, allowing the operator to fail gracefully if CRDs are missing.

This approach improves the operator's robustness by ensuring external dependencies are met, which aligns with Kubebuilder’s goal of helping developers build reliable, high-quality solutions while reducing the learning curve. Some references on using third-party APIs can be found [here](https://developers.redhat.com/blog/2020/02/04/how-to-use-third-party-apis-in-operator-sdk-projects#step_4__use_the_api_in_the_controllers).

Therefore, I would strongly advocate for keeping this implementation. Otherwise, admins might have a very hard time debugging the solution or understanding why it’s not functioning correctly when it should fail during deployment due to unmet prerequisites. Either we lost a great opportunity to help users know how they can do things.

camilamacedo86:
Hi @fischor,

Thank you for your time and review! Regarding your comment in [#4171](https://github.com/kubernetes-sigs/kubebuilder/pull/4171#pullrequestreview-2320867077), I’d like to address your points and explain why we won’t be following the suggestions as outlined.

> 1. An additional `--domain` flag is needed.

The domain is derived from the `PROJECT` file and refers to the project’s domain. When scaffolding a controller for an API/CRD **not** defined in the project, this API does not belong to the project's domain, nor is it created within the project. Thus, the `--domain` flag would be irrelevant and potentially misleading in this case, as it would not be used in any part of the scaffold. 

Just to make it very clear,  **not by using the option of EXTERNAL API via this flag, it means the API is  not defined in the project and is not part of the domain of the project at all **not**. This is the same scenario as when working with core types.

> 2. Suggestion: The `--external-api-path` flag should be defined alongside other resource flags, not as part of the `createApiSubcommand` options.

Each plugin implements the `Plugin` interface, which provides the respective subcommands. Each subcommand has its own specific flags. The `--external-api-path` flag is **only** valid for the `create API` command of the specific `go/v4` plugin at this moment. **It is not valid for any other plugin or subcommand.**

Looking ahead, we plan to add a similar flag for webhooks, and the same principle will apply: the flag will only be added to the webhook subcommand of the `go/v4` plugin. We don't use global flags for these specific cases. In fact, this concept does not exist in the architecture design of the project at all and seems. If you're interested and have the time, you might want to help us with a full refactor of the plugins section. By diving into that, things might become clearer and provide a better understanding of how everything fits together, see: https://github.com/kubernetes-sigs/kubebuilder/pull/4185

Thanks again for the review, and I hope this clarifies the reasoning!

fischor:
Regarding `--domain` flag:

Here the template references `.Resource.QualifiedGroup` which uses the resources Domain which as of now always comes from the PROJECT file:

https://github.com/camilamacedo86/kubebuilder/blob/9517ac29bfbca0cd5c1e1d495a61b1bbad38541e/pkg/plugins/golang/v4/scaffolds/internal/templates/controllers/controller.go#L88-L90

```golang
// +kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }},verbs=get;list;watch;create;update;patch;delete
// +kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }}/status,verbs=get;update;patch
// +kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }}/finalizers,verbs=update
```

which results in the PROJECT domain (here `testproject.org`) being used even for external resources:

https://github.com/camilamacedo86/kubebuilder/blob/9517ac29bfbca0cd5c1e1d495a61b1bbad38541e/testdata/project-v4-multigroup/internal/controller/certmanager/certificate_controller.go#L35-L37

```go
// +kubebuilder:rbac:groups=certmanager.testproject.org,resources=certificates,verbs=get;list;watch;create;update;patch;delete
// +kubebuilder:rbac:groups=certmanager.testproject.org,resources=certificates/status,verbs=get;update;patch
// +kubebuilder:rbac:groups=certmanager.testproject.org,resources=certificates/finalizers,verbs=update
```

This is an issue because controller-gen will then generate the roles for resources that do not exist (because the groups value is not correct). 
Thats why I am thinking to fully describe a external resource alongside `group`, `version`, `kind` and `external-api-path` a `domain` is needed.

---

Regarding this:
> we don't use global flags for these specific cases. In fact, this concept does not exist in the architecture design of the project at all.

The resource flags `--group`, `--version`, `--kind` flags are actived by the CLI create commands whenever there is a plugin that implements the `RequiresResource` interface. The resource is then injected to the plugin. Doesn't make sense to me that now every plugin that wants to work with external resource needs to add explicit flags and related handling code itself.

testdata/project-v4-multigroup/internal/controller/certmanager/certificate_controller.go
```diff
@@ -0,0 +1,62 @@
...
+
+// CertificateReconciler reconciles a Certificate object
+type CertificateReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+}
+
+// +kubebuilder:rbac:groups=certmanager.testproject.org,resources=certificates,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=certmanager.testproject.org,resources=certificates/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=certmanager.testproject.org,resources=certificates/finalizers,verbs=update
```
camilamacedo86, line 37: @fischor 


Thank you for your comment https://github.com/kubernetes-sigs/kubebuilder/pull/4171#issuecomment-2367159862
I will give a look over the rbac marker for the controller-gen that is a good point.

PS.: It would be helpful you create comments based on the code and not like you are doing since it is impossible follow a tread

pkg/plugins/golang/v4/api.go
```diff
@@ -108,6 +108,9 @@ func (p *createAPISubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&p.options.DoController, "controller", true,
 		"if set, generate the controller without prompting the user")
 	p.controllerFlag = fs.Lookup("controller")
+
+	fs.StringVar(&p.options.ExternalAPIPath, "external-api-path", "",
+		"Specify the Go package path for external APIs to scaffold controllers for resources not defined in this project.")
```
camilamacedo86, line 113: > The resource flags --group, --version, --kind flags are actived by the CLI create commands whenever there is a plugin that implements the RequiresResource interface. The resource is then injected to the plugin. Doesn't make sense to me that now every plugin that wants to work with external resource needs to add explicit flags and related handling code itself.

You are speaking here about the helper:

```go
func bindResourceFlags(fs *pflag.FlagSet) *resourceOptions {
	options := &resourceOptions{}

	fs.StringVar(&options.Group, "group", "", "resource Group")
	fs.StringVar(&options.Version, "version", "", "resource Version")
	fs.StringVar(&options.Kind, "kind", "", "resource Kind")

	return options
}
```

However, not all plugins will that scaffold base on GKV can or will necessarily support External Types.
For example, we do not add this support for the Deploy Image plugin (https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go). That would change how its scaffolds works and might not make sense. 

Therefore, we cannot add the helper because not necessary any implementation of subCommands to create API/webhook will or should support external types.

testdata/project-v4-multigroup/cmd/main.go
```diff
@@ -307,3 +326,39 @@ func main() {
...
+				setupLog.Info("API kind is installed on the cluster", "kind", kind)
+				return
+			}
+		}
+	}
+
+	// Log an error if the kind is not found
+	setupLog.Error(nil, fmt.Sprintf("API kind is not available on the cluster (kind: %s)", kind))
+	os.Exit(1)
+}
```
camilamacedo86, line 364: Hi @fischor 

You made a comment that I lost regards you are not so happy with this part.
IHMO we should add the validation.
However, it need to be improved we should check the GKV and not only the Kind 
I am not happy with how it is done yet
So, we might can either let this part for an follow up where we discuss only the new markers and the inject code.

pkg/plugins/golang/v4/scaffolds/internal/templates/main.go
camilamacedo86: Hi @fischor 

IHMO we should add the validation.
**However, it need to be improved we should check the GKV and not only the Kind** 
Therefore, I am not so happy with how it is done yet
So, we might can either let this part for an follow up where we discuss only the new markers and the inject code.

testdata/project-v4-multigroup/internal/controller/certmanager/certificate_controller.go
camilamacedo86: Done 👍 
Thank you for the catcher 🎉

pkg/plugins/golang/v4/api.go
camilamacedo86: Anyway, my point here, after we add for webhooks then we can revisit it if we think that is valid.

camilamacedo86:
Hi @fischor,

See that we have a flag for the domain now and the new makers to scaffold the checks we can add/discuss in a follow so that we can shape those better and make easier the review.  (We can describe the behaviour without to demo why that is required/recommended and etc)

Thank you for your time on reviews.
Fell free to check it now.

pkg/plugins/golang/v4/api.go
fischor: Yes, makes sense.

docs/book/src/reference/using_an_external_resource.md
```diff
@@ -0,0 +1,101 @@
...
+In some cases, your project may need to work with resources that aren't defined by your own APIs. These external resources fall into two main categories:
+
+- **Core Types**: API types defined by Kubernetes itself, such as `Pods`, `Services`, and `Deployments`.
+- **External Types**: API types defined in other projects, such as CRDs defined by another solution.
+
+## Managing External Types
+
+### Creating a Controller for External Types
+
+To create a controller for an external type without scaffolding a resource, use the `create api` command with the `--resource=false` option and specify the path to the external API type using the `--external-api-path` option. This generates a controller for types defined outside your project, such as CRDs managed by other operators.
```
fischor, line 13: Maybe add description for `--external-api-domain` here

fischor (approved):
LGTM. Just some minors left in the docs.

pkg/plugins/golang/v4/api.go
```diff
@@ -131,6 +137,25 @@ func (p *createAPISubcommand) InjectResource(res *resource.Resource) error {
...
+	// Ensure that if any external API flag is set, both must be provided.
+	if len(p.options.ExternalAPIPath) != 0 || len(p.options.ExternalAPIDomain) != 0 {
+		if len(p.options.ExternalAPIPath) == 0 || len(p.options.ExternalAPIDomain) == 0 {
+			return errors.New("Both '--external-api-path' and '--external-api-domain' must be " +
+				"specified together when referencing an external API.")
+		}
+	}
+
+	// add check to ensure that if resource is true externapi and exytermal api domain cannot be set
+	// if one of those is filled then both should be informed
```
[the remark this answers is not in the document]
camilamacedo86: good catcher I will remove !

camilamacedo86:
Hi @fischor 

Thank you a lot for your time and help with the review. 🥇 
The latest NITs are not either addressed !

camilamacedo86:
c/c @lingdie

varshaprasad96 (approved):
/lgtm
/approve

/hold - in case anyone else would like to review. @camilamacedo86 please feel free to remove the hold when you see fit for merge.

PR #4172, opened 2024-09-15
:book: Update README.md - Clarify that monthly Kubebuilder sync meetings are scheduled but most syncing happens offline via Slack

camilamacedo86:
…ngs are scheduled but most syncing happens offline via Slack

We haven't been holding the meetings due to a lack of topics on the agenda, and we've been syncing more frequently offline. This message is just to clarify how we're currently working.


camilamacedo86:
Since it is just a nit I will move forward here.
We have big PRs that we really need help on review.

Issue #1513, opened 2020-05-10
Kubebuilder validation markers are documented multiple times


tamalsaha:
Kubebuilder book defines the same marker multiple times. https://book.kubebuilder.io/reference/markers/crd-validation.html

Is that by mistake? If yes, can you please fix this? It is very confusing to a new user of these markers.

/kind bug

camilamacedo86:
Hi @tamalsaha, 

Could you please give an example of what marker is defined duplicated? See that, for example, the marker `kubebuilder:validation:MaxItems` is not equals `kubebuilder:validation:MaxLength` and the idea here is to allow users to check all options.

hasbro17:
Pretty much the entire block of markers is duplicated after `//+optional` from what I can see.
Might be something with the marker generation script.

camilamacedo86:
I see what you mean 👍 . Note that first, all options are (on field) then it has (on type). 
However, I agree that it is something that requires attention.

gabbifish:
These duplicated markers are actually not duplicated, but correspond to markers used for types vs fields. We can amend the documentation to deduplicate the appearance of markers and specify that a marker can be used on types and fields.

Adirio:
IMO, we should make clear that markers are grouped by targets (field, type, ...), and therefore if a marker is valid for different targets it will appear in both sections, instead of removing them (or deduplicating them). Maybe even explicitly splitting them into targets using second-level titles.

PR #4174, opened 2024-09-15
📖 add note to clarify markers in the doc

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/1513


camilamacedo86 (commented):
It is very small nit for the docs.
Improvements and changes can be done via a follow up and everybody is welcome to just push a PR with suggestions

So, I am moving forward with this one.

PR #4176, opened 2024-09-16
✨ Webhooks for external APIs

fischor:
This is a WIP branch. But looking for feedback already.

With this PR scaffolding for core resource types will be added. Addresses #2141.

This will currently only work with core resources, not external resources per se, since a flag to describe the external Go module is missing in this PR as of now.
Currently the changes in this PR rely of this map of groups here to get the external path set if its a core resource: https://github.com/kubernetes-sigs/kubebuilder/blob/662f78fadaef9caa4302f64d71ccba8c019b8140/pkg/plugins/golang/options.go#L118

Adding a flag like `--module` or `resource-pkg-path` would make this work with general external resources.

This flag could be added as part of the Golang plugin flag set or as part of the global kubebuilder flag set.
I am unsure if adding this flag to the global kubebuilder flag set might break some plugins that never assumed they will ever have to deal with external plugins. 
So adding this flag to the global kubebuilder flag set would need further discussion I think.

I also had to change the `SetupWithManager` method that is as of now a receiver method on the resource struct type since one can not define methods on external struct types in Go.
At the moment it looks like this:

```
// SetupWebhookWithManager will setup the manager to manage the webhooks.
func SetupWebhookForJobWithManager(mgr ctrl.Manager) error {
		For(&batchv1.Job{}).
		WithValidator(&JobCustomValidator{}).
		WithDefaulter(&JobCustomDefaulter{}).
		Complete()
}
```

--- 

It seems webhooks are going to be moved to internal/webhook anyway:  #4150
Then the new schema is similiar.

A `--external-api-path` flag is added also in #4171 as part of the Golang plugin flag set.

Feedback welcome.
I would espacially need to know:
- what the plan with the flag for external APIs is (#4171)
- should I rebase my PR on #4150 (probably yes)?

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/api/webhook.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/api/webhook.go
@@ -40,6 +40,10 @@ type Webhook struct { // nolint:maligned
 	// Define value for AdmissionReviewVersions marker
 	AdmissionReviewVersions string
 
+	// If the webhook is for an external resource. External resources need to be
+	// imported.
+	ExternalAPI bool
+
 	Force bool
 }
 
@@ -97,6 +101,9 @@ import (
 	{{- if .Resource.HasValidationWebhook }}
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 	{{- end }}
+	{{- if .ExternalAPI }}
+	{{ .Resource.ImportAlias }} "{{ .Resource.Path }}"
+	{{- end }}
 )
 
 // nolint:unused
@@ -104,9 +111,17 @@ import (
 var {{ lower .Resource.Kind }}log = logf.Log.WithName("{{ lower .Resource.Kind }}-resource")
 
 // SetupWebhookWithManager will setup the manager to manage the webhooks.
+{{- if .ExternalAPI }}
+func SetupWebhookFor{{ .Resource.Kind }}WithManager(mgr ctrl.Manager) error {
+{{- else }}
 func (r *{{ .Resource.Kind }}) SetupWebhookWithManager(mgr ctrl.Manager) error {
+{{- end }}
 	return ctrl.NewWebhookManagedBy(mgr).
+		{{- if .ExternalAPI }}
+		For(&{{ .Resource.ImportAlias }}.{{ .Resource.Kind }}{}).
+		{{- else }}
 		For(r).
+		{{- end }}
 		{{- if .Resource.HasValidationWebhook }}
 		WithValidator(&{{ .Resource.Kind }}CustomValidator{}).
 		{{- end }}
@@ -137,7 +152,11 @@ var _ webhook.CustomDefaulter = &{{ .Resource.Kind }}CustomDefaulter{}
 
 // Default implements webhook.CustomDefaulter so a webhook will be registered for the Kind {{ .Resource.Kind }}.
 func (d *{{ .Resource.Kind }}CustomDefaulter) Default(ctx context.Context, obj runtime.Object) error {
+	{{- if .ExternalAPI }}
+	{{ lower .Resource.Kind }}, ok := obj.(*{{ .Resource.ImportAlias }}.{{ .Resource.Kind }})
+	{{- else }}
 	{{ lower .Resource.Kind }}, ok := obj.(*{{ .Resource.Kind }})
+	{{- end }}
 	if !ok {
 		return fmt.Errorf("expected an {{ .Resource.Kind }} object but got %T", obj)
 	}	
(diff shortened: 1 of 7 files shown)


fischor:
After looking at #4150, I am not sure if this PR adds any more value, since in combination with #4171 the main aspects of having an import for the API struct type and having a way to determine if a resource is external or not are already solved.

camilamacedo86:
/hold

Lets merge this one first: https://github.com/kubernetes-sigs/kubebuilder/pull/4150
Then, the changes here will be simplified

We will just need to add the flag and https://github.com/kubernetes-sigs/kubebuilder/pull/4171/files
Also, we need test to ensure that all work fine after that.

pkg/plugins/golang/v4/scaffolds/internal/templates/api/webhook.go
```diff
@@ -97,16 +101,27 @@ import (
...
+	{{- end }}
 )
 
 // nolint:unused
 // log is for logging in this package.
 var {{ lower .Resource.Kind }}log = logf.Log.WithName("{{ lower .Resource.Kind }}-resource")
 
 // SetupWebhookWithManager will setup the manager to manage the webhooks.
+{{- if .ExternalAPI }}
+func SetupWebhookFor{{ .Resource.Kind }}WithManager(mgr ctrl.Manager) error {
```
camilamacedo86, line 115: Those changes will not be required after : https://github.com/kubernetes-sigs/kubebuilder/pull/4150

pkg/plugins/golang/v4/scaffolds/internal/templates/api/webhook.go
```diff
@@ -97,16 +101,27 @@ import (
...
+{{- else }}
 func (r *{{ .Resource.Kind }}) SetupWebhookWithManager(mgr ctrl.Manager) error {
+{{- end }}
 	return ctrl.NewWebhookManagedBy(mgr).
+		{{- if .ExternalAPI }}
+		For(&{{ .Resource.ImportAlias }}.{{ .Resource.Kind }}{}).
```
camilamacedo86, line 121: Here should be if the Resource.Path is not empty

camilamacedo86 (changes requested):
Thank you a lot for looking on that.
But the changes here will be a little diff.

Add the support for Core and External types is blocked by:

- https://github.com/kubernetes-sigs/kubebuilder/pull/4150
- https://github.com/kubernetes-sigs/kubebuilder/pull/4171

After we get this merged then we can do that in more precisely way and add tests to ensure it. 

Please

/hold

sundaram2021 (commented):
Review Feedback:
Modular Flag Addition: The flag for external resources seems essential, but clarity is needed on whether it should be added globally or to the Golang plugin. Care should be taken to ensure backward compatibility.

Webhook Setup Changes: The SetupWebhookForJobWithManager function modification looks appropriate for external types, but detailed testing and validation are required to ensure seamless integration with external APIs.

Dependency on Other PRs: The PR (#4150) decoupling webhooks from APIs needs to be merged first, as the current PR's changes depend on it.

Next Steps:
Rebase on #4150 and #4171 once they are merged.
Add tests to validate external API support fully.
Collaborate on flag placement (global vs. plugin).

camilamacedo86:
Hi @fischor, @sundaram2021,

Thank you for looking into this.

We do plan to add support for this feature. However, I would suggest not spending time on it right now, as the approach we'll take will be slightly different from how it's done in this context. It will likely be addressed as a follow-up to the ongoing PR: https://github.com/kubernetes-sigs/kubebuilder/pull/4176#pullrequestreview-2306538293. I want to ensure that no time or effort is wasted, as there are many other priorities in the project!

If you'd like to help us move forward, we'd really appreciate your assistance with reviewing this PR: https://github.com/kubernetes-sigs/kubebuilder/pull/4150. This will help us reach our goal more efficiently.

Thanks! 🎉

pkg/plugins/golang/v4/webhook.go
```diff
@@ -96,12 +96,17 @@ func (p *createWebhookSubcommand) InjectResource(res *resource.Resource) error {
 			" --programmatic-validation and --conversion to be true", p.commandName)
 	}
 
-	// check if resource exist to create webhook
-	if r, err := p.config.GetResource(p.resource.GVK); err != nil {
-		return fmt.Errorf("%s create webhook requires a previously created API ", p.commandName)
-	} else if r.Webhooks != nil && !r.Webhooks.IsEmpty() && !p.force {
+	if p.force {
+		return nil
+	}
```
camilamacedo86, line 101: It is not the same of what was implemented before.

camilamacedo86:
Hi @fischor 

I think the changes here would not work
Would you like to cherry-pick the PRS https://github.com/kubernetes-sigs/kubebuilder/pull/4150 https://github.com/kubernetes-sigs/kubebuilder/pull/4171 and then try to achieve this goal accordingly.

See that we should add the call to generate the webhook to test it out in https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh

fischor:
@camilamacedo86 Yes, that is the plan. I am waiting for these two PRs to get merged before continuing.

camilamacedo86:
Hi @fischor 

In order to see if we can move forward with I started the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/4203
Please, your help is very appreciated. Feel free to get the PR and change it as you see fit

PR #4180, opened 2024-09-22
⚠️ cleanup/refactor: Alpha Generate command

camilamacedo86:
- Refactored the code implementation to improve error handling, encapsulate logic, and streamline execution flow.
- Moved the code implementation from `pkg` to the CLI since it pertains directly to a command.
- Additionally, moved the implementation to `internal` to prevent exposing the `Generate` code and avoid unintentional extensions on this alpha feature.

**Motivation**
Ensure better maintenance ability
Not allow unintentional extensions on this alpha feature.

### DIFF
--- a/pkg/cli/alpha/command.go
+++ b/pkg/cli/alpha/command.go
@@ -15,14 +15,13 @@ package alpha
 
 import (
 	log "github.com/sirupsen/logrus"
-
 	"github.com/spf13/cobra"
-	"sigs.k8s.io/kubebuilder/v4/pkg/rescaffold"
+	"sigs.k8s.io/kubebuilder/v4/pkg/cli/alpha/internal"
 )
 
 // NewScaffoldCommand return a new scaffold command
 func NewScaffoldCommand() *cobra.Command {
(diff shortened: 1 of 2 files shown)


camilamacedo86 (commented):
@varshaprasad96 

We have now very compreensive e2e tests see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/alphagenerate/generate_test.go

So, if pass on the tests the changes are OK.

But I would like your input / review here if you have time.
Just to ensure that you agree with the changes and mainly about we are moving the code implementation. The pkg does not seems a good place for we add the code implementation for commands like that.  

I mean, we are not really want to provide it as pkg/api method right
we probably either should use internal dir here to not export.

varshaprasad96:
Hey @camilamacedo86 - Really sorry for the delay in reply. The refactor looks good, especially moving the code to internal dir. Have no doubts on this! Thank you!

/lgtm
/approve

Kavinjsir (approved):
Looks good to me! 👍🏼

PR #4182, opened 2024-09-22
🐛 cleanup: remove func APIPackagePathLegacy which is no longer used since release 4.0.0

camilamacedo86:
The APIPackagePathLegacy function was useful in older scaffolds to ensure backward compatibility. However, with the release of version 4.0.0, all deprecated features have been removed. As a result, this function should no longer be present


camilamacedo86 (commented):
it is a small nit
So, I am moving forward here 
I will not make nobody waste time in this cleanup

PR #4184, opened 2024-09-22
📖 (doc): remove Makefile Helpers

camilamacedo86:
This section is outdated and does not seems that we need to keep it. If any requirement raise about how to do something via Makefile we can add in the FAQ section instead


mjlshen (approved):
/lgtm

PR #4186, opened 2024-09-22
📖 (doc) remove from menu Appendix: The TODO Landing Page

camilamacedo86:
It does not bring value for the docs. Therefore, we are removing it


camilamacedo86 (commented):
we have many open PRs where we need help to review
it is a very small nit so I am moving forward for we keep our effort on the PRs that we need help within

PR #4187, opened 2024-09-22
📖 (doc): remove section in the platform documentation which says that kube-rbac-proxy is an image provided by default

camilamacedo86:
The usage of kube-rbac-proxy was discontinue and the image is no longer build or promoted by the project. Therefore, this section needs to be removed to avoid misleadings. More info: https://github.com/kubernetes-sigs/kubebuilder/discussions/3907

### DIFF
--- a/docs/book/src/reference/platform.md
+++ b/docs/book/src/reference/platform.md
@@ -218,20 +218,7 @@ end up labeled with ` kubernetes.io/os=linux`
 
 </aside>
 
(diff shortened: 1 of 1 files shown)


camilamacedo86 (commented):
we have many open PRs where we need help to review
this one is quite solid we are just removing in outdated info
So, I am moving forward for not make nobody waste time here.

PR #4181, opened 2024-09-22
🌱cleanup/refactor: Implement and refactor e2e tests for 'alpha generate' command

camilamacedo86:
- Added comprehensive end-to-end tests for the 'generate' command, ensuring proper validation of the 'PROJECT' file after project initialization and regeneration.
- Verified correct handling of multigroup layouts, Grafana, and DeployImage plugins.
- Refactored test structure to align with established patterns from other tests, improving maintainability and consistency.
- Increased test coverage to support future growth and cover more scenarios.

PR #4190, opened 2024-09-23
🐛 Fix HasResource pkg method to properly stop when a match is found

camilamacedo86:
- Updated the HasResource method to correctly stop the iteration when a matching GVK is found.
- Replaced the immediate return inside the loop with a boolean flag  and a break statement to ensure the function exits early when a resource is found.
- This resolves the issue where the method was not stopping as expected after finding a matching resource

PS.: This issue was found when I was implementing the e2e tests for generate alpha command: https://github.com/kubernetes-sigs/kubebuilder/pull/4181

### DIFF
--- a/pkg/config/v3/config.go
+++ b/pkg/config/v3/config.go
@@ -160,13 +160,15 @@ func (c Cfg) ResourcesLength() int {
 
 // HasResource implements config.Config
 func (c Cfg) HasResource(gvk resource.GVK) bool {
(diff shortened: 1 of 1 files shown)


camilamacedo86 (commented):
I am moving forward since it is small fix and is blocking other changes
We have a lot of PRs which we really need help on reviews.
So, I will not make nobody waste the time here.

PR #4191, opened 2024-09-23
⚠️ 🐛fix: remove duplicate HasFragment method, replaced by HasFileContentWith

camilamacedo86:
The HasFragment method was mistakenly added, leading to code duplication. The existing HasFileContentWith method already handles the same functionality, so HasFragment has been removed to avoid redundancy.


camilamacedo86 (commented):
Passing on the test result in be safe to move forward with this small fix.
We have many PRs open which we really need help on review so I will not make anybody waste time on this one. Moving forward.

PR #4197, opened 2024-09-28
✨ (deploy-image/v1alpha): Improve tests scaffolded for the controllers

mogsie:
Motivation: #4135 

* To(Not()) → NotTo()  (It seems that NotTo is used a lot elsewhere in the repository)
* Use default eventually timeouts, so we don't need to send them everywhere
* Eventually(func() error) → Eventually(func(g Gomega)) And then use g.Expect inside those Eventually() functions
* Use Gomega assertions for the condition check, which is more robust

### DIFF
--- a/docs/book/src/getting-started/testdata/project/internal/controller/memcached_controller_test.go
+++ b/docs/book/src/getting-started/testdata/project/internal/controller/memcached_controller_test.go
@@ -18,8 +18,6 @@ package controller
 
 import (
 	"context"
-	"fmt"
-	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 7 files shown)


mogsie:
I have the possibility of rewriting that last eventually as follows:

```
Eventually(func(g Gomega) {
	conditions := []metav1.Condition{}
	g.Expect(busybox.Status.Conditions).To(ContainElement(
		HaveField("Type", Equal(typeAvailableBusybox)), &conditions))
	g.Expect(conditions).To(HaveLen(1), "Multiple conditions of type "+typeAvailableBusybox)
	g.Expect(conditions[0].Status).To(Equal(metav1.ConditionTrue))
	g.Expect(conditions[0].Reason).To(Equal("Reconciling"))
}).Should(Succeed())
```

This would improve the robustness of the test, as the test wouldn't be checking the non-computer friendly _message_ but only the fields that constitute the API: Type, Status, Reason.  It would also be testing for duplicates in the status.

The trick is that `g.Expect().To(ContainElement(..., &conditions))` actually fills out the `&conditions` with the found elements, i.e. the one(s) that have the right Type.  There should only be one in the status.conditions.  This way it only looks for the condition that has the right `Type` and then later we assert that that type has the correct status and reason.

The older test would ignore multiple duplicate conditions (i.e. conditions of the same type), since it only would assert that _one_ status matched.

I will add this too.

docs/book/src/getting-started/testdata/project/internal/controller/memcached_controller_test.go
```diff
@@ -83,30 +82,21 @@ var _ = Describe("Memcached Controller", func() {
...
+			Eventually(func(g Gomega) {
 				found := &appsv1.Deployment{}
-				return k8sClient.Get(ctx, typeNamespacedName, found)
-			}, time.Minute, time.Second).Should(Succeed())
+				g.Expect(k8sClient.Get(ctx, typeNamespacedName, found)).To(Succeed())
+			}).Should(Succeed())
 
 			By("Checking the latest Status Condition added to the Memcached instance")
-			Eventually(func() error {
-				if memcached.Status.Conditions != nil &&
```
mogsie, line 93: I just checked that this test is broken by design; the test verifies that the status of the *created object*  in line 61.  And it does not return an error if the conditions are missing.  I'm hoping the test passes now that I have rewritten it to verify that the _server_ copy has correct status.

mogsie: I have tried to get this test to work, but to me it seems that the reconciler is not actively running, it only runs reconcile once, and it probably needs a few more reconciles for it to become "healthy".

I think I would need to run through the reconcile function again:

```
_, err := busyboxReconciler.Reconcile(ctx, reconcile.Request{
	NamespacedName: typeNamespacedName,
})
```

It's clear that the Conditions were never actually checked.  What should I do here?  Run the reconcile manually a few more times?

camilamacedo86 (approved):
/lgtm

docs/book/src/getting-started/testdata/project/internal/controller/memcached_controller_test.go
```diff
@@ -83,31 +81,25 @@ var _ = Describe("Memcached Controller", func() {
 			})
 			Expect(err).NotTo(HaveOccurred())
 			By("Checking if Deployment was successfully created in the reconciliation")
-			Eventually(func() error {
+			Eventually(func(g Gomega) {
```
mogsie, line 84: I can also remove this Eventually, since after reconciling, we actually _expect_ the deployment to have been created.  Should i do that too @camilamacedo86?

varshaprasad96 (commented):
/lgtm

varshaprasad96 (approved):

PR #4211, opened 2024-10-09
📖  small fixes for the doc Using External Resources

camilamacedo86:
Fix layout and ensure that it has only documented info


camilamacedo86:
We are here only removing the note which is not required
and fixing the layout of the doc.

So, I will move forward here and do not make waste time with this review since we have a lot of open PRs which we really need help on the review

PR #4217, opened 2024-10-14
🐛 fix support for external types by allowing the domain be empty, and properly generate the sample for cert-manager.

camilamacedo86:
- Allowing the domain be empty since some scenarios this value might not be required
- Use `io` as the domain to fix the scaffold sample for cert-manager and update the example in the documentation.


pkg/plugins/golang/v4/scaffolds/internal/templates/controllers/controller.go
```diff
@@ -85,9 +85,9 @@ type {{ .Resource.Kind }}Reconciler struct {
 	Scheme *runtime.Scheme
 }
 
-// +kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }},verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }}/status,verbs=get;update;patch
-// +kubebuilder:rbac:groups={{ .Resource.QualifiedGroup }},resources={{ .Resource.Plural }}/finalizers,verbs=update
+// +kubebuilder:rbac:groups={{ if .Resource.External }}{{ .Resource.Domain }}{{ else }}{{ .Resource.QualifiedGroup }}{{ end }},resources={{ .Resource.Plural }},verbs=get;list;watch;create;update;patch;delete
```
fischor, line 88: I think its fine as it was before using the QualifiedGroup here.

The thing with certmanager (and potentially other external APIs) that they do not necessarily follow the `{domain}/{group}` schema that is used by kubebuilder. The full qualified group name for certmanager is simple `cert-manager.io` without a "sub-group". 

I think when scaffoling a controller for this you should probably have a empty `--group`. Currently the group is set via `--group certmanager`. Would that solve the issues you face at the moment?

camilamacedo86: The QualifiedGroup group seems to make sense for Kubebuilder APIs. 
We should not really allow the group be empty since it will result in a scaffold that will not work if more than one API be scaffolded.

camilamacedo86: Hi @fischor,

Thank you so much! That’s exactly the help I was looking for. We discussed the scenarios to ensure that it will work well.

Your comment made me think: actually, CertManager does not define domains for their APIs, just the group. So, we must accommodate both scenarios—allowing the API to be scaffolded with only the group or with both the group and the domain.

As properly documented, we will ensure that both scenarios are tested with appropriate testdata generation.
The changes were applied on this PR.

camilamacedo86: Hi @fischor,

Thank you so much! That’s exactly the help I was looking for. We discussed the scenarios to ensure that it will work well.

Your comment made me think: actually, CertManager does not define domains for their APIs, just the group. So, we must accommodate both scenarios—allowing the API to be scaffolded with only the group or with both the group and the domain.   In this way, see that we are fixing here how we scaffold and allowing the scaffold be done with an empty domain.

camilamacedo86:
Hi @fischor,

Thank you so much! That’s exactly the help I was looking for. We discussed the scenarios to ensure that it will work well.

Your comment made me think: actually, CertManager does not define domains for their APIs, just the group. So, we must accommodate both scenarios—allowing the API to be scaffolded with only the group or with both the group and the domain.   In this way, see that we are fixing here how we scaffold and allowing the scaffold be done with an empty domain.

docs/book/src/reference/using_an_external_resource.md
```diff
@@ -28,22 +28,22 @@ kubebuilder create api --group <theirgroup> --version <theirversion> --kind <the
 For example, if you're managing Certificates from Cert Manager:
 
 ```shell
-kubebuilder create api --group certmanager --version v1 --kind Certificate --controller=true --resource=false --external-api-path=github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1 --external-api-domain=cert-manager.io
+kubebuilder create api --group cert-manager.io --version v1 --kind Certificate --controller=true --resource=false --external-api-path=github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1
```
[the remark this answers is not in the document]
camilamacedo86: > For me "cert-manager.io" feels like a domain without a group.

Not all projects use the term "domain" explicitly, but **every API must have a group** as part of its organization within the Kubernetes ecosystem. The group provides the namespace for APIs and helps avoid conflicts with other resources. In the case of CertManager, the group is `cert-manager.io`.

### Example YAML Explanation:
```yaml
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: example-certificate
  namespace: default
spec:
  dnsNames:
  - example.com
  secretName: example-cert-secret
  issuerRef:
    name: example-issuer
    kind: Issuer
    group: cert-manager.io <-- SEE HERE IT IS THE GROUP
  duration: 90d
  renewBefore: 30d
```

- **apiVersion**: `cert-manager.io/v1` shows that `cert-manager.io` is the API group, and `v1` is the version of this group.
- **kind**: `Certificate` defines the resource type.
- **group**: In the `issuerRef`, the group `cert-manager.io` **is explicitly defined again**. This ensures that when Kubernetes processes the resource, it knows which group (`cert-manager.io`) the `Issuer` or `ClusterIssuer` belongs to.

In summary, while `cert-manager.io` it is technically the **group** that organizes and names the APIs for CertManager.

> But of course, we need a group since that is used e.g. as folder name. Can you explain why is the group mandatory? Where is it used except as folder name?

Kubernetes enforces the use of an API group to identify and organize its resources, making it essential for maintaining a structured API design. Therefore, we should not scaffold resources without group.

Kubebuilder utilizes the API group in several places, including code generation, CRDs, and RBAC markers, all of which are critical for scaffolding Kubernetes controllers. If the group is left empty, the scaffolding process generates incomplete or invalid code, leading to resource registration errors.

Kubebuilder provides two project layouts:
- **Single group**: where all APIs belong to one group.
- **Multigroup**: where multiple groups are supported under one project.

Without proper directory structure, scaffolding multiple APIs becomes problematic, as resources won't be correctly organized. This is why defining an API group is crucial for generating and organizing APIs effectively.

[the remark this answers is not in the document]
camilamacedo86: My point here is more about how we **recommend** using the feature and if we should or not make the domain mandatory at this point. We could suggest that "io" is treated as the domain and "cert-manager" as the group to provide a complete scaffold structure. At the same time, we can allow the group to remain empty for those who prefer to tweak the configuration manually. There's no need to block this flexibility at the moment but it might generate invalid scaffolds. 

Kubebuilder utilizes the API group in several places, including code generation, and RBAC markers, all of which are critical for scaffolding Kubernetes controllers. If the group is left empty, the scaffolding process generated might be incomplete or invalid. Also, we provides two project layouts:
- **Single group**: where all APIs belong to one group.
- **Multigroup**: where multiple groups are supported under one project.

Without proper directory structure, scaffolding might becomes problematic, as resources won't be correctly organized accordingly.

**Why would we recommend keeping the group empty instead of the domain?**

pkg/plugins/golang/v4/webhook.go
```diff
@@ -108,6 +108,14 @@ func (p *createWebhookSubcommand) InjectResource(res *resource.Resource) error {
...
 	}
 
+	// Ensure that if any external API flag is set, both must be provided.
+	if len(p.options.ExternalAPIPath) != 0 || len(p.options.ExternalAPIDomain) != 0 {
+		if len(p.options.ExternalAPIPath) == 0 || len(p.options.ExternalAPIDomain) == 0 {
+			return errors.New("Both '--external-api-path' and '--external-api-domain' must be " +
+				"specified together when referencing an external API.")
+		}
+	}
+
```
camilamacedo86, line 118: @fischor 

We can keep the check for now and make mandatory the usage of both and just fix the samples and docs for the cert-manager example.

pkg/plugins/golang/v4/api.go
```diff
@@ -146,14 +146,6 @@ func (p *createAPISubcommand) InjectResource(res *resource.Resource) error {
...
 	}
 
-	// Ensure that if any external API flag is set, both must be provided.
-	if len(p.options.ExternalAPIPath) != 0 || len(p.options.ExternalAPIDomain) != 0 {
-		if len(p.options.ExternalAPIPath) == 0 || len(p.options.ExternalAPIDomain) == 0 {
-			return errors.New("Both '--external-api-path' and '--external-api-domain' must be " +
-				"specified together when referencing an external API.")
-		}
-	}
-
```
camilamacedo86, line 156: Hi @fischor,

In my opinion, groups should not be empty when dealing with non-core types in scaffolds. The concepts should stay aligned with Kubernetes itself. We cannot have Kubebuilder diverging from core Kubernetes principles. However, in order to not block others scenarios we should not make mandatory users provide the Domain.

What are your thoughts?

varshaprasad96 (approved):
/lgtm
/approve

pkg/plugins/golang/v4/api.go
fischor: I am refering to the "group" (`--group`) as used in kubebuilder, not the Kubernetes apiGroup.
The apiGroup will not be empty if you do `--domain cert-manager.io --group ""`, it will be `cert-manager.io` then.

My thoughts are that is does not make sense to have --domain and --group for external APIs, since these are kubebuilder specific entities and not all external APIs use that scheme. Maybe just --api-group is better to specify the (full) apiGroup for external APIs.

PR #4230, opened 2024-10-25
🌱 Fix incorrect image reference for DeployImage plugin in test data

camilamacedo86:
We generate a sample API in the testdata directory using the DeployImage plugin. However, the flag previously contained incorrect information regarding the image that should be used. This fix ensures the correct image is referenced.


camilamacedo86:
It is too small has no reason to make someone waste time on the review.
It just need to pass in the CI.
So, I am moving forward within .

PR #4233, opened 2024-10-29
🌱 Fix simplify external plugin tests

camilamacedo86:
- Bump sigs.k8s.io/kubebuilder/v4 from 4.2.0 to 4.3.0 in /docs/book/src/simple-external-plugin-tutorial/testdata/sampleexternalplugin/v1.
- Resolve compatibility by moving tests to GitHub Actions to use source code directly.
- Remove external plugin e2e tests, as they do not run against a cluster, reducing complexity without testing across all k8s versions.

Closes: https://github.com/kubernetes-sigs/kubebuilder/pull/4232


camilamacedo86:
I will not make nobody waste time with this one since it does not affected end users and we are just moving the tests for GitHub actions which make more sense for this case scenario and let us know to do the tests with the latest changes.

PR #4234, opened 2024-10-29
✨ Upgrade controller-runtime from v0.19.0 to v0.19.1

camilamacedo86:
✨ Upgrade controller-runtime from v0.19.0 to v0.19.1


camilamacedo86:
It is just a small patch bump
The APIDiff in this case fails as a false alarm. 
moving forward

PR #4235, opened 2024-10-29
✨ Upgrade kustomize version from v5.4.3 to v5.5.0

camilamacedo86:
✨ Upgrade kustomize version from v5.4.3 to v5.5.0


camilamacedo86:
It is just a small patch bump
The APIDiff in this case fails as a false alarm. 
moving forward

PR #4236, opened 2024-10-29
✨ Upgrade golangci-lint from v1.59 to v1.61

camilamacedo86:
✨ Upgrade golangci-lint from v1.59 to v1.61


camilamacedo86:
It is just a small patch bump
passing in the CI is more than enough

Issue #4200, opened 2024-10-01
Alpha Generate Command Explicitly Calls Kubebuilder Binary


camilamacedo86:
### What do you want to happen?

The command at [this location](https://github.com/kubernetes-sigs/kubebuilder/blob/33e24da9a304755ed0c67e12100fd2b42734fb9b/pkg/cli/alpha/internal/generate.go#L104) explicitly invokes the Kubebuilder binary.

However, other projects may use Kubebuilder as a library and implement their own binaries. In such cases, the current behavior restricts the command to work only with Kubebuilder's binary.

**Proposal:**

The command should be designed to work with custom binaries implemented by projects that use Kubebuilder as a library.

**Alternative:**

If we decide not to support this use case, we should explicitly make clear that the alpha command is not intended to be imported or reused by other projects. Ideally not allow re-usage.

TAM360:
> In such cases, the current behavior restricts the command to work only with Kubebuilder's binary.
> Proposal:
> The command should be designed to work with custom binaries implemented by projects that use Kubebuilder as a library.

@camilamacedo86 question: Kubebuilder is used mainly for writing custom k8s operators? Why would someone use a custom version of Kubebuilder while invoking the `alpha` commands?

camilamacedo86:
Hi @TAM360,

If you check, Kubebuilder can also be used as a library. Therefore, in the code implementation of the plugins, we use "a command" instead of relying on a fixed `kubebuilder` binary. For example, Operator-SDK is a project that utilizes Kubebuilder for scaffolding. See; https://github.com/search?q=repo%3Akubernetes-sigs%2Fkubebuilder%20CommandName&type=code

For Operator-SDK to also use the re-generate feature and provide it for its users, we cannot rely on a fixed `kubebuilder` binary or search for a binary with a fixed name. This approach aligns with how we handle commands without tying them to a specific binary.

Let me know if this makes sense or if you have any questions!

TAM360:
> If we decide not to support this use case, we should explicitly make clear that the alpha command is not intended to be imported or reused by other projects. Ideally not allow re-usage.

I think `alpha` command should be supported at library, just like others. @camilamacedo86 what would be the reason for not supporting this type of use case?

camilamacedo86:
Hi @TAM360 

> I think alpha command should be supported at library, just like others

If possible yes.

TAM360:
@camilamacedo86 let me work on this!

[...]

PR #4247, opened 2024-10-30
🐛 clarify usage of alpha generate command as internal-only

camilamacedo86:
Added comments to `NewScaffoldCommand` to clarify that the `kubebuilder alpha generate` command is designed for Kubebuilder’s internal use only.

The updated go documentation explains that importing this command into other projects is not supported, as it relies on Kubebuilder-specific configurations, key mappings, and plugins, which may not be applicable to other setups.

Projects that use Kubebuilder as a library and require similar functionality should implement a custom solution for compatibility with their specific configurations and plugins.

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/4200

### DIFF
--- a/pkg/cli/alpha/command.go
+++ b/pkg/cli/alpha/command.go
@@ -19,7 +19,21 @@ import (
 	"sigs.k8s.io/kubebuilder/v4/pkg/cli/alpha/internal"
 )
 
-// NewScaffoldCommand return a new scaffold command
+// NewScaffoldCommand returns a new scaffold command, providing the `kubebuilder alpha generate`
+// feature to re-scaffold projects and assist users with updates.
+//
+// IMPORTANT: This command is intended solely for Kubebuilder's use, as it is designed to work
+// specifically within Kubebuilder's project configuration, key mappings, and plugin initialization.
+// Its implementation includes fixed values and logic tailored to Kubebuilder’s unique setup, which may
+// not apply to other projects. Consequently, importing and using this command directly in other projects
+// will likely result in unexpected behavior, as external projects may have different supported plugin
+// structures, configurations, and requirements.
+//
+// For other projects using Kubebuilder as a library, replicating similar functionality would require
+// a custom implementation to ensure compatibility with the specific configurations and plugins of that project.
+//
+// Technically, implementing functions that allow re-scaffolding with the exact plugins and project-specific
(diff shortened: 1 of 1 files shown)


camilamacedo86:
Hi @acornett21, JFY.

TAM360:
@camilamacedo86 shouldn't this also be mentioned in the kubebuilder's official docs? Devs rely on the official website 1st before, exploring the repo.

camilamacedo86:
Hi @TAM360 

> @camilamacedo86 shouldn't this also be mentioned in the kubebuilder's official docs? Devs rely on the official website 1st before, exploring the repo.

It is only relevant for those who maintain projects which uses Kubebuilder as Lib. Example maintainers of Operator-SDK
The only doc section relevant in this case is the plugins
Also, the golang docs which will be updated when we do a new release with those changes.

camilamacedo86:
Hi @acornett21 

It seems that we are importing the alpha generated in SDK, see: https://github.com/operator-framework/operator-sdk/blob/819984d4c1a51c8ff2ef6c23944554148ace0752/website/content/en/docs/cli/operator-sdk_alpha_generate.md?plain=1#L4

If we at some point bump KB there we might want to give a look on that and ensure that we are not using this command there. It cannot work with SDK projects.

camilamacedo86:
I am moving forward with this one since here we are only add a comment to clarify the intent purpose of the command.
If anyone would like to propose improvements please feel free to push a PR with your suggestions.

acornett21:
> It seems that we are importing the alpha generated in SDK, see: https://github.com/operator-framework/operator-sdk/blob/819984d4c1a51c8ff2ef6c23944554148ace0752/website/content/en/docs/cli/operator-sdk_alpha_generate.md?plain=1#L4
> 
> If we at some point bump KB there we might want to give a look on that and ensure that we are not using this command there. It cannot work with SDK projects.

It looks like `NewScaffoldCommand()` isn't used anywhere in `operator-sdk` and the `alpha-generate` cmd is just using the below

https://github.com/operator-framework/operator-sdk/blob/819984d4c1a51c8ff2ef6c23944554148ace0752/internal/cmd/operator-sdk/alpha/config3alphato3/convert_config_3-alpha_to_3.go#L28

maybe that this point this cmd should just be removed. Thanks for the callout.

Issue #3712, opened 2023-12-18
Analysis and Compliance Strategy for New Cyber Resilience Act


camilamacedo86:
### Issue Description:

We need to conduct a thorough analysis of the new [Cyber Resilience Act](https://www.linuxfoundation.org/blog/understanding-the-cyber-resilience-act) to understand its implications for the Kubebuilder project, particularly in terms of our release process, tooling, and dependencies.
We probably need to start to generate the SBOOMs. Also, note that today we use GCP to perform the builds and we have a desire to use only GitHub Actions to do so and no longer need to use GCP. 

### Areas of Focus:

1. **Current Release Process:**
   - Currently, our releases are managed via GCP and build triggers.
   - We utilize `goreleaser` for automating releases, triggered by pushing a new tag.
   - Relevant documentation: [Release Process](https://github.com/kubernetes-sigs/kubebuilder/blob/master/RELEASE.md) and [Goreleaser Configuration](https://github.com/kubernetes-sigs/kubebuilder/blob/master/build/.goreleaser.yml).

2. **Kubebuilder Tools:**
   - We have been developing `kubebuilder-tools`, a binary to assist users with `envtest`.
   - For more details, refer to the [tools-releases branch](https://github.com/kubernetes-sigs/kubebuilder/tree/tools-releases) and [this issue comment](https://github.com/kubernetes-sigs/controller-runtime/issues/2583#issuecomment-1824371209).

3. **Kube-RBAC-Proxy:**
   - Currently, we are handling the retagging of `kube-rbac-proxy`.
   - Considering delegating this responsibility to the `kubernetes-sig` and re-evaluating its use in default scaffolding due to recent breaking changes.

### Objectives:

1. **Create a Proposal Design Document:**
   - Utilizing the [Design Doc Template](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/template.md), we need to document how the new legislation affects our project and outline necessary compliance measures.
   - This document should serve as the foundation for understanding our obligations and planning our response.

2. **Develop Compliance Strategies:**
   - Based on the analysis, we need to propose solutions and modifications to our processes and tools to ensure compliance with the new legislation.

### Objectives:

1. **Create a Proposal Design Document:**
   - Utilizing the [Design Doc Template](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/template.md), we need to document how the new legislation affects our project and outline necessary compliance measures.
[...]

PR #4248, opened 2024-10-30
🌱 Add SBOM generation for Cyber Resilience Act compliance

camilamacedo86:
- Integrated SBOM generation in GoReleaser using Syft to produce CycloneDX SBOMs.
- Updated GitHub Actions workflow to install Syft, enabling automated SBOM creation on release.
- This enhancement is part of ongoing efforts to align with the EU Cyber Resilience Act, ensuring transparency and security in our software supply chain.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,6 @@ jobs:
   goreleaser:
     runs-on: ubuntu-latest
 
-
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -25,6 +24,10 @@ jobs:
           go-version: '~1.22'
       - name: Clean dist directory
         run: rm -rf dist || true
+      - name: Install Syft to generate SBOMs
+        run: |
+          curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b $HOME/bin
```
grzesuav, line 29: is it worth to consider caching the binary?

grzesuav (approved):

.github/workflows/release.yml
camilamacedo86: I do not think so.
We just run this action when we do the release (not so many times per year)

PR #4249, opened 2024-10-30
🌱 Update Dependabot configuration

camilamacedo86:
- Remove configurations for deprecated branches no longer in use
- Change schedule to perform checks daily for up-to-date dependency management
- Ensure dependencies in sample projects are properly included in checks

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,7 +11,7 @@ updates:
     # default location of `.github/workflows`
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "daily"
     commit-message:
       prefix: ":seedling:"
     labels:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -21,48 +21,56 @@ updates:
...
     schedule:
-      interval: daily
+      interval: "daily"
     commit-message:
       prefix: ":seedling:"
+    labels:
+      - "ok-to-test"
+
+  # Maintain dependencies for dockerfile scaffold in the projects
   - package-ecosystem: docker
```
mateusoliveira43, line 61: this is the only entry for docker. Should every folder (`/`, `/testdata/project-v4`, `/testdata/project-v4-with-plugins`, `/testdata/project-v4-multigroup` and `docs/book/src/simple-external-plugin-tutorial/testdata/sampleexternalplugin/v1`) have dependabot checks for dockerfiles?

camilamacedo86: Good point 👍

camilamacedo86: Hi @mateusoliveira43 

The docs/book/src/simple-external-plugin-tutorial/testdata/sampleexternalplugin/v1 has no docker image
And the goal is be informed about some required change
The docker file is the sample for all samples so by just checking project-v4 we will be able to be notified if any change in the Dockerfile scaffold by Kubebuilder be required.

mateusoliveira43: understood

mateusoliveira43 (approved):

PR #4251, opened 2024-10-30
:sparkles:  (go/v4): Enable `seccompProfile.type: RuntimeDefault` by default in scaffolded projects

camilamacedo86:
- Removed the comment about `seccompProfile` as `type: RuntimeDefault` is now set by default.
- Updated to reflect that Kubebuilder only tests and supports the latest three Kubernetes releases.
- Given that we’re currently using Kubernetes 1.31, there’s no longer a reason to keep this line commented, ensuring projects adhere to `restricted` Pod Security Standards by default.
- Either OCP 4.10 is in EOL already: https://endoflife.date/red-hat-openshift

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/manager/manager.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/manager/manager.yaml
@@ -49,14 +49,12 @@ spec:
       #             values:
       #               - linux
       securityContext:
+        # Projects are configured by default to adhere to the "restricted" Pod Security Standards.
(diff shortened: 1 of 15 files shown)


camilamacedo86:
/hold 

Just to not merge before we do a patch release; https://github.com/kubernetes-sigs/kubebuilder/issues/4284

camilamacedo86 (commented):
/hold cancel

We can get it merged now.
Since it is to small I will not make nobody waste time on reviews here.
We have big fishes to fry.

PR #4252, opened 2024-10-30
🌱 Add GitHub workflow to scan project with CodeQL

camilamacedo86:
This workflow introduces CodeQL to automatically scan the project for security vulnerabilities, supporting efforts to maintain a secure codebase. It also helps us comply with standards like the Cybersecurity Certification Act (CCA) by ensuring continuous security monitoring.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,58 @@
+name: "CodeQL Advanced"
+
+on:
+  # We are checking both `master` and `book-v4` branches:
+  # - `master` represents the latest development work.
+  # - `book-v4` is the latest stable release branch, which contains the latest published code,
+  #   ensuring that any issues in production are identified and addressed promptly.
+  push:
+    branches: ["master", "book-v4"]
+  pull_request:
+    branches: ["master", "book-v4"]
+  schedule:
+    - cron: '30 20 * * 1'  # Runs every Monday at 8:30 PM
+
+jobs:
+  analyze:
+    name: Analyze Go
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Build project-v4 sample project
+        run: |
+          cd testdata/project-v4
+          go mod tidy
+          make all
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v3
+        with:
+          languages: ${{ matrix.language }}
```
grzesuav, line 57: can't go be hardcoded here ?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,72 @@
...
+          build-mode: ${{ matrix.build-mode }}
+
+      - if: matrix.language == 'go' && matrix.build-mode == 'manual'
```
grzesuav, line 60: not sure when this condition is true (`manual` build mode)

.github/workflows/codeql.yml
camilamacedo86: Thx good point, I think we can simplify it !

.github/workflows/codeql.yml
camilamacedo86: Thx good point, I think we can simplify it !

camilamacedo86:
Hi @grzesuav 

Really thank you for your review.
All addressed it is very simplified now.

So, since it is just a check that does not impact the end users and we addressed all I am moving forward
make no sense I make nobody waste so much time on reviews.

However, if you see any possible improvement here as anybody else please feel free to push a PR with your suggestions.

Issue #2589, opened 2022-04-02
scaffolds hub and spoke interfaces for conversion webhooks.


camilamacedo86:
### What do you want to happen?

scaffolds hub and spoke interfaces for conversion webhooks.
With this change, users can either specify --conversion flag and implement
the interface on their own, or specify --hub and --spoke versions which
will scaffold the skeleton code for both the interfaces.

Follow up/finish the work which started : https://github.com/kubernetes-sigs/kubebuilder/pull/2084

NikhilSharmaWe:
@camilamacedo86 What do we need to do here? 

- Do we need to open a new PR with the changes done in https://github.com/kubernetes-sigs/kubebuilder/pull/2084, working on the suggested changes in the PR, and rebasing it?

NikhilSharmaWe:
@camilamacedo86 Could you please take a look at my previous comment.

camilamacedo86:
Sure. Please, feel free to work on this one.

camilamacedo86:
Yes, the goal of this task is to finalize the implementation done in the https://github.com/kubernetes-sigs/kubebuilder/pull/2084. So, you can get the changes made and address the comments. 

IN this way, this one seems a good fit for the good first issue.

momom-i:
@camilamacedo86 Is it already done? If not, I'd like to work on this!

camilamacedo86:
Hi @momom-i.

It is not. Please, feel free to work on this one. 
Your collaboration is very welcome.

Thx.

ashutosh887:
I don't think it's still resolved
I would like to work on this now

/assign

momom-i:
@ashutosh887 I’m working on this now. 
/assign

ashutosh887:
Sure @momom-i 
Please let me know if you need any help

ashutosh887:
Let me work on this @camilamacedo86 if @momom-i is unable to resolve

lauchokyip:
For anyone who is interested in working on this PR, IMHO, I do not think it's a `good-first-issue`. It requires good amount of knowledge for Go, cli and Kubernetes Webhook. You would also need to read the design proposals to come up with the best practice. **Simply copy and paste the OP's PR will not work**.

Estimated time from understanding to coding to open an PR and to get it merged would require at least a month of effort

camilamacedo86:
It was good first issue when the effort was follow up the work done.
However, now it is no longer the case.
[...]

Issue #4146, opened 2024-09-09
Unable to Add Additional Webhooks Without Forcing Re-scaffold


camilamacedo86:
### What broke? What's expected?

When trying to add an additional webhook (e.g., conversion webhook) to an API where other webhooks (like defaulting or validation) are already scaffolded, Kubebuilder throws an error stating that the webhook files already exist. Currently, the only workaround is to use the --force option, which leads to the entire webhook being re-scaffolded, overwriting any custom implementation added by the user. This behavior results in lost code and makes it difficult to incrementally add webhooks.

Kubebuilder should allow adding additional webhooks without forcing the user to use --force and without overwriting existing implementation. It should:

- Detect existing webhooks.
- Ensure appropriate markers are present.
- Scaffold only the necessary webhook logic (e.g., for conversion, validation, or defaulting) specific to the new request, preserving existing code.

Such as we do in the main.go and for the test-e2e as well. See the WebhookTestUpdater for test-e2e. We need ensure that webhooks follow the same: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/test.go#L58-L103
 

### Reproducing this issue

- 1. Scaffold an API with defaulting and validation webhooks:

```
kubebuilder create api --group batch --version v1 --kind CronJob --resource=true --controller=true
kubebuilder create webhook --group batch --version v1 --kind CronJob --defaulting --programmatic-validation
```

- 2. Try adding a conversion webhook for the same API and kind: 

```
kubebuilder create webhook --group batch --version v1 --kind CronJob --conversion
```

- 3. You’ll encounter an error indicating that the webhook files already exist.

- 4. The only option to proceed is to use --force, which will remove the current webhook implementation:

```
kubebuilder create webhook --group batch --version v1 --kind CronJob --conversion --force
```

- 5. The --force option re-scaffolds the entire webhook, leading to the loss of any custom code previously added to the webhook.

**Expected Behavior/Request** Kubebuilder should scaffold only the specific webhook (e.g., conversion) requested without overwriting the existing webhook code. It should intelligently detect existing webhooks and add the necessary logic for the new type while keeping the current implementation intact.

### KubeBuilder (CLI) Version

4.2.0

fischor:
/assign
Would like to start once #4150 is merged

camilamacedo86:
We cannot skip.
[...]

PR #4254, opened 2024-10-30
:sparkles: (go/v4):  Add Hub and Spoke for conversion webhooks

camilamacedo86:
This PR introduces support for scaffolding a conversion webhook with a hub-and-spoke model in Kubebuilder Go plugin (v4). The feature allows defining one hub version with one or many spoke versions for seamless conversion between CustomResourceDefinition (CRD) API versions.

### How It Will Work:

Users can:
```
kubebuilder create webhook --group example.com --version v1 --kind MyKind --conversion --spoke v2
```

OR 

```
kubebuilder create webhook --group example.com --version v1 --kind MyKind --conversion --spoke v2,v3
```

Closes; https://github.com/kubernetes-sigs/kubebuilder/issues/2589


pkg/plugins/golang/v4/webhook.go
```diff
@@ -48,6 +48,11 @@ type createWebhookSubcommand struct {
 	// Deprecated - TODO: remove it for go/v5
 	// isLegacyPath indicates that the resource should be created in the legacy path under the api
 	isLegacyPath bool
+
+	// Target resource GVK for conversion webhook
+	toGroup   string
+	toVersion string
+	toKind    string
```
camilamacedo86, line 55: Change here 
The flag would be better as spoke 
So it is more clear.

camilamacedo86:
/hold 

Just to not merge before we do a patch release; https://github.com/kubernetes-sigs/kubebuilder/issues/4284

camilamacedo86:
It is important because we have been adding e2e tests to ensure all features and options, including scaffolds and examples, as much as possible. The latest piece missing is the webhook conversion, which is not fully implemented yet but will be once we merge the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/4254

pkg/model/resource/webhooks.go
```diff
@@ -33,6 +33,8 @@ type Webhooks struct {
 
 	// Conversion specifies if a conversion webhook is associated to the resource.
 	Conversion bool `json:"conversion,omitempty"`
+
+	Spoke string `json:"spoke,omitempty"`
```
varshaprasad96, line 37: `Spoke` can be multiple, so this should probably be `[]string`. Or do we want the user to execute this command multiple times for each spoke version? One issue would be that we need to ensure that `hub` is not scaffolded again as it may interfere with the already generated code.

pkg/plugins/golang/v4/scaffolds/internal/templates/webhooks/webhook.go
```diff
@@ -81,7 +81,7 @@ func (f *Webhook) SetTemplateDefaults() error {
 	if f.Force {
 		f.IfExistsAction = machinery.OverwriteFile
 	} else {
-		f.IfExistsAction = machinery.Error
+		f.IfExistsAction = machinery.SkipFile
```
varshaprasad96, line 84: Is this to enable executing conversion command multiple times? Since this template covers both validation and defaulting - there could be issues with just skipping file generation. Do we want to add print statements?

pkg/model/resource/webhooks.go
```diff
@@ -77,6 +79,9 @@ func (webhooks *Webhooks) Update(other *Webhooks) error {
 	// Update conversion.
 	webhooks.Conversion = webhooks.Conversion || other.Conversion
 
+	// Update Spoke.
+	webhooks.Spoke = other.Spoke
+
```
varshaprasad96, line 84: Since we are supporting only one `hub` - is there a way we can validate, where hub is not being changed. Or the spoke version was not previously declared as a `hub`. Or would this error occur while trying to scaffold the markers?

pkg/cli/alpha/internal/generate.go
```diff
@@ -354,15 +354,15 @@ func createWebhook(resource resource.Resource) error {
...
-		args = append(args, "--conversion")
-	}
 	if resource.HasValidationWebhook() {
 		args = append(args, "--programmatic-validation")
 	}
 	if resource.HasDefaultingWebhook() {
 		args = append(args, "--defaulting")
 	}
+	if resource.HasConversionWebhook() {
+		args = append(args, "--conversion", "--spoke", resource.Webhooks.Spoke)
```
varshaprasad96, line 364: Probably add validation that `--conversion` needs to be set, to have `--spoke` version? Wondering if we need `--conversion` in that case? Can it be automatically considered true if user specifies `--spoke`?

docs/book/src/multiversion-tutorial/conversion.md
```diff
@@ -1,8 +1,16 @@
...
 
 With our model for conversion in place, it's time to actually implement
-the conversion functions.  We'll put them in a file called
-`cronjob_conversion.go` next to our `cronjob_types.go` file, to avoid
+the conversion functions.  We'll create a conversion webhook
+for our CronJob API version `v1` (Hub) to Spoke our CronJob API version
+`v2` see:
+
+```go
+kubebuilder create webhook --group batch --version v1 --kind CronJob --conversion--spoke v2
```
varshaprasad96, line 9:
```suggestion
kubebuilder create webhook --group batch --version v1 --kind CronJob --conversion --spoke v2
```

pkg/cli/alpha/internal/generate.go
varshaprasad96: Actually, rethinking about this - iiuc it is:
1. if `conversion` is specified - the existing functionality remains.
2. if `spoke` is specified - then `Hub` and `Spoke` interfaces are scaffolded.

In that case would it make sense to validate that `--conversion` is set if `--spoke` is specified.

pkg/plugins/golang/v4/scaffolds/internal/templates/api/spoke.go
```diff
@@ -0,0 +1,93 @@
...
+	"log"
+
+    "sigs.k8s.io/controller-runtime/pkg/conversion"
+    {{ .Resource.ImportAlias }} "{{ .Resource.Path }}"
+)
+
+// ConvertTo converts this {{ .Resource.Kind }} to the Hub version ({{ .Resource.Version }}).
+func (src *{{ .Resource.Kind }}) ConvertTo(dstRaw conversion.Hub) error {
+    dst := dstRaw.(*{{ .Resource.ImportAlias }}.{{ .Resource.Kind }})
+	log.Printf("Converting from %T to %T", dst.APIVersion, src.APIVersion)
```
varshaprasad96, line 77:
```suggestion
	log.Printf("Converting from %T to %T", src.APIVersion, dst.APIVersion)
```

IIUC it's the other way around - `ConverTo` converts to the hub version to store it in the API server. 
Eg: https://github.com/kubeflow/kubeflow/blob/ec82fbf58b79cff529d948b96e44ffd06bdfe679/components/notebook-controller/api/v1alpha1/notebook_conversion.go#L24

varshaprasad96 (commented):
Thanks @camilamacedo86! Just a few questions on the `hub` and `spoke` specifications.

pkg/plugins/golang/v4/scaffolds/internal/templates/webhooks/webhook.go
camilamacedo86: You are right here.
We have a terrible UX now to work with webhooks since is not possible create them without force as a follow up.
we need to fix that as well:  https://github.com/kubernetes-sigs/kubebuilder/issues/4146

I will think a solution here for we not be blocked by: https://github.com/kubernetes-sigs/kubebuilder/issues/4146
until we have this one.

pkg/cli/alpha/internal/generate.go
camilamacedo86: If we do not inform the spoke that will fails because we will need to have that right.
Without the spoke we cannot implement it

pkg/model/resource/webhooks.go
camilamacedo86: I changed for []string
but maybe the right thing here would be have a List of GKV

docs/book/src/multiversion-tutorial/testdata/project/PROJECT
```diff
@@ -18,7 +18,11 @@ resources:
   path: tutorial.kubebuilder.io/project/api/v1
   version: v1
   webhooks:
+    conversion: true
     defaulting: true
+    spoke:
+    - v2
+    - v2
```
camilamacedo86, line 25: @varshaprasad96 

Would it make sense to have Many spokes?
If I want to convert my V1 to V2
Why would I want to convert my V1 to V2 and v3 at the same time?

pkg/model/resource/webhooks.go
```diff
@@ -77,10 +101,36 @@ func (webhooks *Webhooks) Update(other *Webhooks) error {
 	// Update conversion.
 	webhooks.Conversion = webhooks.Conversion || other.Conversion
 
+	// Update Spoke (merge without duplicates)
+	if len(other.Spoke) > 0 {
```
varshaprasad96, line 105: Is there a way to find if the provided spoke version is not one of hub? Given we are not storing the hub version in `Webhooks` struct? Or are we catching that error before?

varshaprasad96 (approved):
Thanks @camilamacedo86! Tested this out, and this looks good!
/lgtm
/approve

pkg/model/resource/webhooks.go
varshaprasad96: Sorry, just tested this out. Looks like we are catching it. Pls ignore.

PR #4273, opened 2024-10-30
:seeding: Update dependabot.yml to just raise PRs against project-v4 sample

camilamacedo86:
The go.mod check for sample configuration helps us identify potential required bumps. However, note that go.mod is often updated by controller-runtime, so we can't manually change certain dependencies.

Additionally, we only need to check one sample project—there’s no need to duplicate efforts by opening the same PR across all samples.

Follow up of : https://github.com/kubernetes-sigs/kubebuilder/pull/4249

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -37,26 +37,6 @@ updates:
     labels:
       - "ok-to-test"
 
-  # Maintain dependencies for go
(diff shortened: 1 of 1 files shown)


camilamacedo86:
Moving forward to fix the duplication and so many open PRs

PR #4275, opened 2024-10-31
🐛 fix: ensure --make=false option is available for webhook creation for consistency

camilamacedo86:
Standardize resource generation options by adding the `--make=false` flag to `kubebuilder create webhook`, aligning it with the existing API creation process. This option is essential in scenarios such as project scaffolding, where running `make` commands immediately is deferred to allow modifications to generated files before execution.


camilamacedo86:
It is a very small silly fix so I will not make nobody waste time on reviews of this one.
moving forward

PR #4277, opened 2024-10-31
🌱 Update codeql.yml to run the workflow once per week only

camilamacedo86:
Follow up of : https://github.com/kubernetes-sigs/kubebuilder/pull/4252

We do not want to run for each Pull Request.
Execute it once per week is more than enough


camilamacedo86:
Moving forward within since it will affect our productive and it is just an small fix that does not impact end users,
We can review and change any time

PR #4281, opened 2024-11-01
🌱 Add bundle installer for sample projects in docs

camilamacedo86:
Introduces a bundle installer for sample projects, with config/default/kustomization.yaml settings uncommented to ensure complete bundle generation. This enhances review visibility by showing the final state of patches and Kustomize configurations directly in the generated bundle.

Motivation: Makes reviews more efficient by displaying final configurations directly in the bundle, with only relevant settings uncommented in config/default/kustomization.yaml for each sample.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
@@ -110,37 +110,37 @@ replacements:
          index: 1
          create: true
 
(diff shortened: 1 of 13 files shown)


camilamacedo86:
This change primarily benefits maintainers. It doesn’t introduce visible changes or impact end-users but is a minor enhancement to streamline our internal review process. Given our current workload and the number of pending PRs, I’ll move forward without additional review to save time. This update allows us to focus on other critical requirements while making reviews more efficient.

PR #4283, opened 2024-11-02
🐛 (kustomize/v2,go/v4): Move cert-manager uncomment block to top of kustomization.yaml for better visibility in new projects

camilamacedo86:
- Relocated the cert-manager setup block (kind: Service, version: v1, name: webhook-service) to improve user experience since this one must be uncommented always and was hard to identify as it was organized previously.
- Change only affects new projects.


camilamacedo86:
We are not really change anything here
Just the position for two reasons:
- Make easier for users be aware that the first block must be uncomment always
- Make easier for us uncomment to ensure our e2e tests.

I am moving forward due be too small and it is blocking other PRs which we really need help on reviews.

PR #4286, opened 2024-11-04
:bug: (go/v4) Fix issue with scaffolding multiple webhooks for the same resource

camilamacedo86:
Attempting to create multiple webhooks for the same resource with different configurations (e.g., `--defaulting` and `--programmatic-validation`) resulted in an error indicating that the webhook already exists.
    
This update allows multiple webhook types to be scaffolded for a single resource without conflict. Users can scaffold different webhooks for the same GKV afterwards without needing to do so all at once, as expected.

Either we are adding tests to cover the webhook undergo/v4 conversion type.
The tests are scaffolding the conversion webhook after the defaulting and validation were created already.
Therefore, we are validating the fix done in this PR. 

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/4146

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/webhooks/webhook.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/webhooks/webhook.go
@@ -81,7 +81,7 @@ func (f *Webhook) SetTemplateDefaults() error {
 	if f.Force {
 		f.IfExistsAction = machinery.OverwriteFile
 	} else {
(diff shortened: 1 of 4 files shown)


camilamacedo86 (commented):
It is a very small bug so I will move forward 
I will not make nobody waste time on the reviews here when we have a lot of open PRs that we really need a hand on the reviews.

Anyway, if someone has any suggestion for improvement and etc please fell free to send a PR against. Your help is very welcome and appreciated !

PR #4288, opened 2024-11-04
:warning: container friendly commands

mateusoliveira43:
## Why the changes were made

Replace `docker` with `container` from command names that do not necessarily need Docker to be used, but should work with any container tool (Podman, Buildah, etc).

> **Note:** I use these commands with Podman, so I validated them with this tool, but not with any other container tool.

## How the changes were made

Checked places that mentions `docker` with the command 
```sh
grep -Inr "docker" ./pkg/plugins/golang/v4/scaffolds/internal/templates/
```
and changed them to a container friendly wording.

> **Note:** Still have to run `make generate` and change `docker` appearances/references in documentation. Waiting to see if community wants this change before doing more changes.

Run new `container-build` and `container-push` commands, and confirm they work as `docker-build` and `docker-push` used to work.

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -178,12 +178,12 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # If you wish to build the manager image targeting other platforms you can use the --platform flag.
 # (i.e. docker build --platform linux/arm64). However, you must enable docker buildKit for it.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
-.PHONY: docker-build
-docker-build: ## Build docker image with the manager.
+.PHONY: container-build
+container-build: ## Build container image with the manager.
 	$(CONTAINER_TOOL) build -t ${IMG} .
 
-.PHONY: docker-push
-docker-push: ## Push docker image with the manager.
+.PHONY: container-push
+container-push: ## Push container image with the manager.
 	$(CONTAINER_TOOL) push ${IMG}
 
 # PLATFORMS defines the target platforms for the manager image be built to provide support to multiple
@@ -194,7 +194,7 @@ docker-push: ## Push docker image with the manager.
 # To adequately provide solutions that are compatible with multiple platforms, you should consider using this option.
 PLATFORMS ?= linux/arm64,linux/amd64,linux/s390x,linux/ppc64le
 .PHONY: docker-buildx
-docker-buildx: ## Build and push docker image for the manager for cross-platform support
+docker-buildx: ## Build and push container image for the manager for cross-platform support
 	# copy existing Dockerfile and insert --platform=${BUILDPLATFORM} into Dockerfile.cross, and preserve the original Dockerfile
 	sed -e '1 s/\(^FROM\)/FROM --platform=\$$\{BUILDPLATFORM\}/; t' -e ' 1,// s//FROM --platform=\$$\{BUILDPLATFORM\}/' Dockerfile > Dockerfile.cross
 	- $(CONTAINER_TOOL) buildx create --name {{ .ProjectName }}-builder
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/readme.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/readme.go
@@ -45,7 +45,7 @@ func (f *Readme) SetTemplateDefaults() error {
 		"*/", "", 1)
 
 	f.TemplateBody = fmt.Sprintf(readmeFileTemplate,
-		codeFence("make docker-build docker-push IMG=<some-registry>/{{ .ProjectName }}:tag"),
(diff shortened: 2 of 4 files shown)


pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -194,7 +194,7 @@ docker-push: ## Push docker image with the manager.
 # To adequately provide solutions that are compatible with multiple platforms, you should consider using this option.
 PLATFORMS ?= linux/arm64,linux/amd64,linux/s390x,linux/ppc64le
 .PHONY: docker-buildx
-docker-buildx: ## Build and push docker image for the manager for cross-platform support
+docker-buildx: ## Build and push container image for the manager for cross-platform support
 	# copy existing Dockerfile and insert --platform=${BUILDPLATFORM} into Dockerfile.cross, and preserve the original Dockerfile
 	sed -e '1 s/\(^FROM\)/FROM --platform=\$$\{BUILDPLATFORM\}/; t' -e ' 1,// s//FROM --platform=\$$\{BUILDPLATFORM\}/' Dockerfile > Dockerfile.cross
 	- $(CONTAINER_TOOL) buildx create --name {{ .ProjectName }}-builder
```
mateusoliveira43, line 200: TODO change `$(CONTAINER_TOOL)` here in this command to `docker`

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -178,12 +178,12 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # If you wish to build the manager image targeting other platforms you can use the --platform flag.
 # (i.e. docker build --platform linux/arm64). However, you must enable docker buildKit for it.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
-.PHONY: docker-build
-docker-build: ## Build docker image with the manager.
```
camilamacedo86, line 182: This change could significantly impact our users, as these targets are likely used in CI pipelines and other automated processes. We need to carefully weigh the pros and cons here.

pkg/plugins/golang/v4/scaffolds/internal/templates/readme.go
```diff
@@ -71,9 +71,9 @@ const readmeFileTemplate = `# {{ .ProjectName }}
 
 ### Prerequisites
 - go version v1.22.0+
-- docker version 17.03+.
-- kubectl version v1.11.3+.
-- Access to a Kubernetes v1.11.3+ cluster.
+- a container tool (Docker, Podman, Buildah, etc)
```
camilamacedo86, line 74: The issue is that we can't claim full support for all container tools since our testing is limited to Docker. To support additional tools, we would need to test across each of them as well, which would be a considerable burden for us as maintainers. Don't you think?

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
mateusoliveira43: yeah, waiting to see if community wants the change before changing the other places

pkg/plugins/golang/v4/scaffolds/internal/templates/readme.go
mateusoliveira43: So, my though here was where the requirement appears, which is the **Getting Started** section of the scaffolded README. In this section, we just need to build and push a container image to run the steps described. So should be ok for not test all options here (I tested with Podman, but not with Buildah, for example)

we also have a comment about not testing all options today
https://github.com/kubernetes-sigs/kubebuilder/blob/5f8342e45517e84a8cc8dd4eff825fe340477111/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go#L86-L90

But for other parts of the project, we can still say docker is still required, and no need for changing how is tested (for example, if a section about multi-arch release is added to scaffolded README, `docker-buildx` command would be necessary and then, docker would be necessary requirement for that section)

what you think?

camilamacedo86: Yes, we added a comment to make clear that it **Might** work with other tools. However, it is **not** tested, officially supported.

camilamacedo86: For example, if we allow the tests run here it will broke all our CI as well.

mateusoliveira43: agree with waiting for a major update to merge this one

camilamacedo86: So, could you please track an issue with the request, and we can create a go/v5 milestone to add this one there?
When we have enough request changes, we can create a new version and do this one.

Also, could you please link this PR there as a reference and close it?

mateusoliveira43: will do

mateusoliveira43: created https://github.com/kubernetes-sigs/kubebuilder/issues/4293

PR #4308, opened 2024-11-09
:bug: revert the changes done in the PR #4286

camilamacedo86:
We are reverting the changes done in the PR #4286. We are just keeping the e2e tests improvements.
We will not able to solve the issue https://github.com/kubernetes-sigs/kubebuilder/issues/4146 without add new markers.

We cannot skip the action to write an webhook since it will result in incomplete inplementation in the <kind>_webhook.go as in the <kind>_webhook_test.go


camilamacedo86:
I am just reverting the silly fix for https://github.com/kubernetes-sigs/kubebuilder/pull/4286 which end up not been so silly ;-)

PR #4243, opened 2024-10-29
:sparkles: Adds a patch to configure ServiceMonitor with to ensure TLS verification using cert-manager certificates

camilamacedo86:
Adds a patch to configure ServiceMonitor with  to ensure TLS verification using cert-manager certificates. Updates documentation and corrects misaligned comments.

Motivation: https://github.com/kubernetes-sigs/kubebuilder/issues/3657

Closes (Phase-2): https://github.com/kubernetes-sigs/kubebuilder/issues/3871

PR #4312, opened 2024-11-10
✨ Follow-up to PR #4243: Limit permissions to access the metrics-server-cert secret

camilamacedo86:
This change ensures that only the necessary permissions are granted for accessing the metrics-server-cert secret. Note that setting defaultMode: 420 is not required.


camilamacedo86 (commented):
It is a small fix as follow up.
Pass in the tests is more than enough to ensure this one
So, moving forward.

PR #4314, opened 2024-11-10
(WIP - TODO Address change agreed on review)✨ (helm/v1alpha): add samples manifests

camilamacedo86:
Add samples to the helm chart so that consumers and developers can decide if should provide good samples configurations and enable them for the chart be installed within

**motivation**

The inclusion of sample CRs in Helm charts is motivated by the need to improve user experience and ease of use. By providing default configurations and ready-to-use examples, we reduce the setup time for users, guide them in customizing their CRs, and ensure the Operator is functional right after installation. This approach not only simplifies onboarding but also allows users to quickly verify the Operator’s deployment, making the setup process more intuitive and productive.

### DIFF
--- a/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
+++ b/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
@@ -236,6 +236,7 @@ func (s *initScaffolder) copyConfigFiles() error {
 	}{
 		{"config/rbac", "dist/chart/templates/rbac", "rbac"},
 		{"config/crd/bases", "dist/chart/templates/crd", "crd"},
+		{"config/samples", "dist/chart/templates/samples", "samples"},
 		{"config/network-policy", "dist/chart/templates/network-policy", "networkPolicy"},
 	}
 
--- a/pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/values.go
+++ b/pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/values.go
@@ -120,6 +120,10 @@ crd:
   # (Certificates, Issuers, ...) due to garbage collection.
   keep: true
 
+# [SAMPLES]: To apply the sample(s) manifests set true
+samples:
+  enable: false
+
 # [METRICS]: Set to true to generate manifests for exporting metrics.
 # To disable metrics export set false, and ensure that the
 # ControllerManager argument "--metrics-bind-address=:8443" is removed.
--- a/testdata/project-v4-with-plugins/dist/chart/templates/samples/example.com_v1_wordpress.yaml
+++ b/testdata/project-v4-with-plugins/dist/chart/templates/samples/example.com_v1_wordpress.yaml
@@ -1,0 +1,10 @@
+{{- if .Values.samples.enable }}
+apiVersion: example.com.testproject.org/v1
+kind: Wordpress
+metadata:
+  labels:
+    {{- include "chart.labels" . | nindent 4 }}
+  name: wordpress-sample
+spec:
+  # TODO(user): Add fields here
+{{- end -}}
--- a/testdata/project-v4-with-plugins/dist/chart/templates/samples/example.com_v1alpha1_busybox.yaml
+++ b/testdata/project-v4-with-plugins/dist/chart/templates/samples/example.com_v1alpha1_busybox.yaml
@@ -1,0 +1,12 @@
+{{- if .Values.samples.enable }}
+apiVersion: example.com.testproject.org/v1alpha1
+kind: Busybox
+metadata:
+  labels:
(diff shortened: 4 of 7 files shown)


camilamacedo86:
Hi,  @[monteiro-renato. @mhkarimi1383, @algo7

Could you please help on this review?

mhkarimi1383:
Hi
I think this is not needed specially in production charts

Using helm chart tests is better than this one I think 
a simple test than brings up the Operator, then apply manifest and check for events/status

Using that you can test both helm chart and the operator

camilamacedo86:
Hi @mhkarimi1383 

Hi mhkarimi1383,

Thank you for your input! However, I believe providing samples is beneficial, especially with this feature that encapsulates an Operator into a Helm chart.

Adding sample CRs is considered good practice since it allows authors to include default configurations and helpful examples for customization. Many projects include CR samples during installation so users don’t have to create configurations from scratch. These samples not only simplify onboarding but also serve as a guide for users who may want to customize their CRs.

Without the CRD samples, the Operator won’t perform any actions, so including them ensures that the installation is immediately functional and users have a reference for testing and customization.

So, I am totally in favor of we move forward here.
Anyway, users can always change and customize their own thing
They can create a make target that removes the samples when for example they release or sync the chart again.
However, IMO we should encourage authors to provide the samples.

mhkarimi1383:
@camilamacedo86 

Nice but I think the better location for examples are in chart NOTES.txt

damsien:
I think that we must not force the developer to have the `samples` section in its `values.yaml`. Even if we can disable it, if the developer does not provide any samples in the chart (because it is well documented or for any reasons), it is a little bit confusing to have a section that do not make any action.

BUT, I think that it is a very good idea to implement this feature for the reasons you said.
> Many projects include CR samples during installation so users don’t have to create configurations from scratch.
> However, IMO we should encourage authors to provide the samples.

So, my proposal is to keep this feature **and** add a section in the documentation (that does not exists yet from what I know but maybe in the future) that explains how to remove this section from the scaffold so it will not be added each time we execute `kubebuilder edit --plugins=helm/v1-alpha`.

camilamacedo86:
Hi @damsien,

I think you might have brought a good idea.
We might have flags for not adding staff on the chart.
In this case, users could see flags for generating a chart without the following, for example.

- samples
- prometheus
- networkPolicy
- certmanager
- webhook 

We can raise an issue after this one gets merged and work on it as a follow-up. WDYT?

damsien:
Yes we can raise an issue to implement the `ignore-` flags!

And concerning this sentence:
> it will not be added each time we execute `kubebuilder edit --plugins=helm/v1-alpha`.
It is in the case where the developer wants to remove the `samples` section from the values but it is added each time the command is executed since the `ignore-samples` flag does not exist yet

camilamacedo86:
Hi @damsien 

> It is in the case where the developer wants to remove the samples section from the values but it is added each time the command is executed since the ignore-samples flag does not exist yet

I see. But it is an alpha version and add the flags is not a lot of work
So, we can do that as a follow-up and probably before the next release when it will be shipped.

damsien:
Yes sure

mhkarimi1383:
> Hi @mhkarimi1383 
> 
> > Nice but I think the better location for examples are in chart NOTES.txt
> 
> Samples are CR. 
> I do see it fitting in the Notes.
> They can have a default config. 
> For example, I want a MyDatabase Operator. 
> Install the Operator will not do anything if I do not create an instance (apply the CR) 
> The default configuration for MyDatabse Operator would be set in the samples. 
> That was the idea. I hope that clarifies. 
> 

I think as a user who installs the operator chart, NOTES will show me after installing the operator, So I will notice samples and any recommendations and so on
Also we could have another chart for CRs (like CloudNativePG), especially useful when we are able to have some CRs related together (also like CloudNativePG:))

camilamacedo86:
Hi @mhkarimi1383,

> I think as a user who installs the operator chart, NOTES will show me after installing the operator, So I will notice samples and any recommendations and so on
Also we could have another chart for CRs (like CloudNativePG), especially useful when we are able to have some CRs related together (also like CloudNativePG:))

For it be possible you need move with this PR
Because if we do not users will NOT have the CRs to add the info in the NOTES :-)

mhkarimi1383:
As users I mean the person who installs the operator to use, not the creator of operator

If I got you correct

camilamacedo86:
Ji @mhkarimi1383 

I see your point.
It might be a good idea.
Just remove the option for the values but sync the samples in the chart !!!
I will change it in this direction :-)

camilamacedo86:
Hi @mhkarimi1383 

The problem is that the samples have data that the cluster admin might want to configure
So, we probably need to study a solution here to see how we want to deal with
I will open an issue so we are able to discuss and track the conclusions.

For now, I am closing this PR as not accepted.

PR #4321, opened 2024-11-11
📖 (docs): removed ununsed variable and nil check

vtrenton:
Hello!

I ran though the memcached example and found a few linting errors to be cleaned up in example.

1. typeDegradedMemcached is defined but completely unused...
2. having if  memcached.Status.Conditions == nil || len(memcached.Status.Conditions) == 0 is a duplicate check. only one of the 2 checks is needed. len() is safe to call on nil slices in go since it returns 0.

my vim-go also moved the "time" import. I can move that back though if it needs to be!

let me know what you think and hope to see you at Kubecon NA!

### DIFF
--- a/docs/book/src/getting-started/testdata/project/internal/controller/memcached_controller.go
+++ b/docs/book/src/getting-started/testdata/project/internal/controller/memcached_controller.go
@@ -19,13 +19,14 @@ package controller
 import (
 	"context"
 	"fmt"
+	"time"
+
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


camilamacedo86 (changes requested):
Hi @vtrenton 

But note that the sample is generated automatically

The code implementation that call the binary and inject the code on top is here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/hack/docs/internal/getting-started/generate_getting_started.go

Then, it is required update their and run in this case `make generate-docs` so that you have all updated.

vtrenton:
Thanks for the reply!

I will close this then and create a new PR updating that file instead!

Thank you so much for all your help with my PRs :)

PR #4323, opened 2024-11-13
📖 (Docs): Cleaned up some linting errors on getting-started example

vtrenton:
Just a minor PR to fix up some nitpicky things in the examples. Making PRs as I go along reading the book :)

Please see -> 
https://github.com/kubernetes-sigs/kubebuilder/pull/4321
For more context.

These are the following changes made:
1. typeDegradedMemcached is completely unused - so there really is no point defining it.
2. if memcached.Status.Conditions == nil || len(memcached.Status.Conditions) == 0 - is repetative. checking the len is enough.

I made changes to the `generate_getting_started.go` and ran `make generate-docs` in the root of kubebuilder.

Thanks for all the hard work that goes into this!!


docs/book/src/getting-started/testdata/project/internal/controller/memcached_controller.go
```diff
@@ -90,7 +88,7 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 
 	// Let's just set the status as Unknown when no status is available
-	if memcached.Status.Conditions == nil || len(memcached.Status.Conditions) == 0 {
+	if len(memcached.Status.Conditions) == 0 {
```
camilamacedo86, line 91: Hi @vtrenton 

We need to keep this one. The `len(memcached.Status.Conditions) == 0` would work if Conditions **is always initialized as an empty slice ([]metav1.Condition{}) when empty**. However, if Conditions happens to be nil, this check won't catch it and could lead to a nil pointer error if future code assumes Conditions is always non-nil.

Because of that we need to keep this one

docs/book/src/getting-started/testdata/project/internal/controller/memcached_controller.go
```diff
@@ -39,8 +39,6 @@ import (
 const (
 	// typeAvailableMemcached represents the status of the Deployment reconciliation
 	typeAvailableMemcached = "Available"
-	// typeDegradedMemcached represents the status used when the custom resource is deleted and the finalizer operations are yet to occur.
-	typeDegradedMemcached = "Degraded"
```
camilamacedo86, line 43: 👍 That is a good catcher we are not using it !

camilamacedo86 (commented):
Hi @vtrenton,

Thank you for your contribution! 🥇

Could you please revert the change at [this line](https://github.com/kubernetes-sigs/kubebuilder/pull/4323/files#r1839666834)? We need to keep the nil check to handle cases where it hasn’t been initialized yet.

Also, please make sure to squash all changes into a single commit, as we can’t merge with multiple commits.


Thanks a lot for your help!

vtrenton:
lmaoooo sorry I messed this up trying to consolidate commits :facepalm: 
I'll just open a new one - I got a few other docs fixes I want to add too.

docs/book/src/getting-started/testdata/project/internal/controller/memcached_controller.go
vtrenton: sorry I didnt see this until now!

Issue #4327, opened 2024-11-15
(Makefile) remove-spaces command does not work on GNU system


damsien:
### What broke? What's expected?

This issue is related to the `Makefile` of Kubebuilder and **not** to the generated `Makefile` used for the operator.

The command `make remove-spaces` executed on a GNU system raises an error.
https://github.com/kubernetes-sigs/kubebuilder/blob/e9beffb5926f8fcf1abf97ada7ae3a3c675d0591/Makefile#L71-L74

That is due to a difference between the `sed` of the GNU system and the OS X system (this command works well on MacOS). It is well explained [here](https://unix.stackexchange.com/a/131940).

My proposal to solve this is to add a check of the system.
```
.PHONY: remove-spaces
remove-spaces:
	@echo "Removing trailing spaces"
        @SED_CMD="sed -i ''" && [ "$$(uname)" != "Darwin" ] && SED_CMD="sed -i"; \
	find . -type f -name "*.md" -exec $$SED_CMD 's/[[:space:]]*$$//' {} + || true
```

### KubeBuilder (CLI) Version

4.3.1

camilamacedo86:
that is working in the CI (which is ubuntu)
but I am fine with any change that works for mac and linux
feel free to open a pr with your suggestion

damsien:
From what I saw in the `.github/workflows`, there is no call to this `make remove-spaces`. This command is only called once by the `make generate` command.
https://github.com/kubernetes-sigs/kubebuilder/blob/20a99551e2e7c7681949e0f08bbc3594b6249ef4/Makefile#L66-L69

Correct me if I am wrong but the `make generate` command is never executed during the CI process.

PS: the "Trailing spaces" CI job only check for trailing spaces without trying to remove them (`make test-spaces`).
https://github.com/kubernetes-sigs/kubebuilder/blob/20a99551e2e7c7681949e0f08bbc3594b6249ef4/.github/workflows/spaces.yml#L11-L21

damsien:
Should we keep this issue open? I wanted to show you the error. In the following screenshot, the first command is executed without the "fix" I made. And the second command is executed with the "fix".

![image](https://github.com/user-attachments/assets/fa6dc39b-62ba-4649-a466-cf5b3fc10691)

The fix I made is wrong. I can try to find a friend that have a mac to actually test the command or wait for someone that want to take the issue.

camilamacedo86:
HI @damsien 

[...]

PR #4330, opened 2024-11-15
🌱 : Fix `make remove-spaces` for GNU system

damsien:
This PR is motivated by this issue: #4327.

It is a fix about the `make remove-spaces` command that was only working on OS X and not GNU based system. It is because the behavior of the `sed -i` command is different between these two systems (see [this stack exchange discussion](https://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed/131940#131940)).

- On OS X: `sed -i ''`
- On GNU: `sed -i`

So I added a check by using `uname` to use the `sed -i` in the correct way depending if we are on GNU based system or on OS X based system.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,8 @@ generate: generate-testdata generate-docs ## Update/generate all mock data. You 
 .PHONY: remove-spaces
 remove-spaces:
 	@echo "Removing trailing spaces"
-	@find . -type f -name "*.md" -exec sed -i '' 's/[[:space:]]*$$//' {} + || true
+	@SED_CMD="sed -i ''" && [ "$$(uname)" != "Darwin" ] && SED_CMD="sed -i"; \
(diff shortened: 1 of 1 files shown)


camilamacedo86:
@damsien 

See that your solution broke things

Screenshot 2024-11-16 at 09 52 35

So we will revert. 
And next one I will ensure locally either before get merged

damsien:
@camilamacedo86 I am sorry I didn't had the opportunity to test on a mac (I'm using arch linux). Hope that it didn't take you too long to revert

camilamacedo86:
Hi @damsien 

> @camilamacedo86 I am sorry I didn't had the opportunity to test on a mac (I'm using arch linux). Hope that it didn't take you too long to revert

Not at all. It is all fine
It is my fault; I should test it out locally before approving
Just ping you for yu check so that you might want try out fix it again

damsien:
Thanks! I'll create another PR when I have time to try to fix it again!

PR #4346, opened 2024-11-20
📖 Fix: broken url in external-plugins.md

monteiro-renato:
Relevant section > https://book.kubebuilder.io/plugins/extending/external-plugins#how-to-write-an-external-plugin

### DIFF
--- a/docs/book/src/plugins/extending/external-plugins.md
+++ b/docs/book/src/plugins/extending/external-plugins.md
@@ -148,4 +148,4 @@ kubebuilder create api --plugins go/v4,sampleplugin/v1
 - A [sample external plugin written in Python](https://github.com/rashmigottipati/POC-Phase2-Plugins)
 - A [sample external plugin written in JavaScript](https://github.com/Eileen-Yu/kb-js-plugin)
 
-[code-plugin-external]: ./../../../../../pkg/plugin/external/types.go
+[code-plugin-external]: https://github.com/kubernetes-sigs/kubebuilder/blob/book-v4/pkg/plugin/external/types.go


docs/book/src/plugins/extending/external-plugins.md
```diff
@@ -148,4 +148,4 @@ kubebuilder create api --plugins go/v4,sampleplugin/v1
 - A [sample external plugin written in Python](https://github.com/rashmigottipati/POC-Phase2-Plugins)
 - A [sample external plugin written in JavaScript](https://github.com/Eileen-Yu/kb-js-plugin)
 
-[code-plugin-external]: ./../../../../../pkg/plugin/external/types.go
+[code-plugin-external]: https://github.com/kubernetes-sigs/kubebuilder/blob/book-v4/pkg/plugin/external/types.go
```
camilamacedo86, line 151: Could we change to use the right relative paths?

monteiro-renato: Hmm, I think the issue here is that we are not pointing to a .md file.
But I can take a look in the mdBook docs to see if it's possible.

camilamacedo86: I see that makes sense 
we have a way to add code 
I would need to give a look, but we are probably not using the core right here
See in other places that we can inject the code on the docs via annotations

camilamacedo86 (approved):
/ok-to-test
/lgtm
/approved

Lets fix for now, after we might find ways to improve it
Thank you 🥇

monteiro-renato:
Hey 👋,
Sorry I haven't looked at this yet. It's in my TODOs but I'm prioritizing helm plugin work first. I might have time to look at this ones in the weekend.

camilamacedo86:
I will move forward because it is just doc.
I hope that does not become a mass because I am doing the squash via GitHub

PR #4351, opened 2024-11-20
🐛 fix: (helm/alpha-v1): Fix rendered value for app.kubernetes.io/name

monteiro-renato:
bilde

The fix for the extra line after "**_labels:_**" is being fixed in this PR: https://github.com/kubernetes-sigs/kubebuilder/pull/4350


monteiro-renato:
With both changes:
bilde

PR #4357, opened 2024-11-21
🐛 fix: (helm/v1alpha1) add missing name prefix and namespace to leader-election-role and leader-election-rolebinding

monteiro-renato:
I'm not sure this is the best approach but since there are already some hardcoded values in there I suppose it's not a terrible idea either.

The [Role](https://github.com/monteiro-renato/kubebuilder/blob/d0765c4eff32891f634183a2ab6487e8693b4c54/testdata/project-v4-with-plugins/config/rbac/leader_election_role.yaml) and [RoleBinding](https://github.com/monteiro-renato/kubebuilder/blob/d0765c4eff32891f634183a2ab6487e8693b4c54/testdata/project-v4-with-plugins/config/rbac/leader_election_role_binding.yaml) do not have a namespace associated with them when the project is `init`ed. I didn't actually try to deploy this but I believe this would actually fail to apply as is (without the namepace on the Role and RoleBinding - [src](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole)).

I added the expected value that a later [instruction](https://github.com/monteiro-renato/kubebuilder/blob/d0765c4eff32891f634183a2ab6487e8693b4c54/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go#L333-L334) is expecting.

There are a few other resources that do not have the prefix either. I can open other PRs to address those individually.

### DIFF
--- a/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
+++ b/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
@@ -292,6 +292,20 @@ func copyFileWithHelmLogic(srcFile, destFile, subDir, projectName string) error 
 		contentStr = strings.Replace(contentStr,
 			"name: metrics-reader",
 			fmt.Sprintf("name: %s-metrics-reader", projectName), 1)
+		contentStr = strings.Replace(contentStr,
+			"name: leader-election-role",
+			fmt.Sprintf("name: %s-leader-election-role", projectName), -1)
+		contentStr = strings.Replace(contentStr,
+			"name: leader-election-rolebinding",
+			fmt.Sprintf("name: %s-leader-election-rolebinding", projectName), 1)
+
+		// The generated files do not include the namespace
+		if strings.Contains(contentStr, "leader-election-rolebinding") ||
+			strings.Contains(contentStr, "leader-election-role") {
+			namespace := `
+  namespace: {{ .Release.Namespace }}`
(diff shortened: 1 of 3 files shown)


camilamacedo86 (changes requested):
Unfortunately, we cannot move forward with this one.
See that we need to keep the exact names of those files under the config/RBAC

https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v4/config/rbac

The name does not use the project name by default,
If users change locally the name of the file, then that will be changed.

monteiro-renato:
Hey 👋,
Apologies if I'm opening too many PRs and it's hard to keep track of what's going on.

I basically noticed these small inconsistencies when I compared the generated manifests via Kustomize and Helm side by side .
Check the install.yaml of [this one](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/dist/install.yaml#L575-L590) as an example.

camilamacedo86 (approved):
Hi @monteiro-renato 

Really thank you to catcher this one 🥇 
We can look in the dist/install.yaml to validate how kustomize
generate the staff: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-with-plugins/dist/install.yaml

You are right we should change the name 
I hope has no other places.
I think for now is OK we change in the code as you did
We might find better ways in the future

Again thank you a lot for all hard work 🥇 

/lgtm
/approve
/ok-to-test

PR #4371, opened 2024-11-22
🐛 fix: (helm/v1alpha1): env config should not be dependent on DeployImages

monteiro-renato:


camilamacedo86 (approved):
Make sense 👍 

/lgtm
/approve
/ok-to-test

Thank you 🥇

monteiro-renato:
Updated the commit message now.

monteiro-renato:
I'm still relatively new to the codebase but shouldn't it be the deployimage plugin that should be responsible for checking that the project is using the helm plugin as the dist method and therefore update the values.yaml file with the needed env vars (and other relevant config)?

I need to play around with how plugin chaining works as my previous assumption seems to be wrong. I see that the helm plugin implementation is assuming that it will (potentially) be run after the deployimage.

camilamacedo86:
Hi @monteiro-renato 

My idea was: If someone use the plugin, then the code must be there otherwise not
But you are right. If someone customize it I think has no reason to accept that.

PR #4373, opened 2024-11-22
✨ feat: (helm/v1alpha1): Allow extra pod labels to be configured

monteiro-renato:
The use case I have in mind is for example [Azure WI configuration](https://azure.github.io/azure-workload-identity/docs/topics/service-account-labels-and-annotations.html#pod).

### DIFF
--- a/pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/chart-templates/manager/manager.go
+++ b/pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/chart-templates/manager/manager.go
@@ -75,6 +75,11 @@ spec:
       labels:
         {{ "{{- include \"chart.labels\" . | nindent 8 }}" }}
         control-plane: controller-manager
+        {{ "{{- if and .Values.controllerManager.pod .Values.controllerManager.pod.labels }}" }}
+        {{ "{{- range $key, $value := .Values.controllerManager.pod.labels }}" }}
+        {{ "{{ $key }}" }}: {{ "{{ $value }}" }}
+        {{ "{{- end }}" }}
+        {{ "{{- end }}" }}
     spec:
       containers:
         - name: manager
--- a/testdata/project-v4-with-plugins/dist/chart/templates/manager/manager.yaml
+++ b/testdata/project-v4-with-plugins/dist/chart/templates/manager/manager.yaml
@@ -19,6 +19,11 @@ spec:
       labels:
         {{- include "chart.labels" . | nindent 8 }}
         control-plane: controller-manager
+        {{- if and .Values.controllerManager.pod .Values.controllerManager.pod.labels }}
+        {{- range $key, $value := .Values.controllerManager.pod.labels }}
+        {{ $key }}: {{ $value }}
+        {{- end }}
+        {{- end }}
     spec:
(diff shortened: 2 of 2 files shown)


pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/chart-templates/manager/manager.go
```diff
@@ -75,6 +75,11 @@ spec:
       labels:
         {{ "{{- include \"chart.labels\" . | nindent 8 }}" }}
         control-plane: controller-manager
+        {{ "{{- if and .Values.controllerManager.pod .Values.controllerManager.pod.labels }}" }}
+        {{ "{{- range $key, $value := .Values.controllerManager.pod.labels }}" }}
+        {{ "{{ $key }}" }}: {{ "{{ $value }}" }}
+        {{ "{{- end }}" }}
+        {{ "{{- end }}" }}
```
camilamacedo86, line 82: See: https://github.com/kubernetes-sigs/kubebuilder/blob/d40e7f2282a9607de48a01b2fbc696ca4529448d/testdata/project-v4/config/default/kustomization.yaml#L11-L15

I think we need to replicate the behaviour
We need to check this file
and ensure that we do in the chart the same what we would to do with kustomize

monteiro-renato: I think the use-case is a bit different.
This PR is an attempt to have something similar to:
```
# This patch adds an Azure Workload Identity required label to the pod
- op: add
  path: /spec/template/metadata/labels
  value:
    azure.workload.identity/use: "true"
```
------------------------------------------------------------------------------------------------
Since you mentioned that file however, I did notice what is likely an unintended behavior while testing kubebuilder with kustomize. If you run:
```
cd /tmp && \
mkdir kbtestdir && \
cd kbtestdir && \
git init && \
kubebuilder init --domain tutorial.kubebuilder.io --repo tutorial.kubebuilder.io/project && \
make build-installer && \
git add . && \
kubebuilder create api --group batch --version v1 --kind CronJob --controller --resource  && \
git add . && \
make build-installer
```
and then check the changes, you will notice that the ClusterRole in the **_config/rbac/role.yaml_** file had it's labels removed. The generated CRD in the config/crd/bases/ is also generated without labels. For this use case I think it makes sense to use kustomize to make sure even the generated resources get the labels added.

camilamacedo86: But above the option to add labels is commented.
So, it would not add new labels to the result of build-installer.
I am not sure if I follow, sorry.

But if you see that is a bug/or you would expected another behaviour
Can you please open an issue with the steps, what you have seen and what you expected?
So, we can fix that as well.

camilamacedo86 (commented):
Hi @monteiro-renato 

Lets move forward with this one.
It is an alpha version anyway we can break things if we need after
But we need to have an issue to see how we work with 

https://github.com/kubernetes-sigs/kubebuilder/pull/4373/files#r1853825165

camilamacedo86 (approved):
/lgtm

PR #4377, opened 2024-11-22
✨ (helm/v1alpha1): add GitHub action to make easier validate the chart generate

camilamacedo86:
Add action will help us and users validate the chart
See that it works: https://github.com/camilamacedo86/test-project/actions/runs/11971426907/job/33376226272


monteiro-renato:
Hi @camilamacedo86,
Should we add a step before "**_Install Helm chart for project_**" with a step that prints the final manifest?
i.e. something along the lines of `helm template --namespace=qualquer dist/chart`.

That's basically what I was doing locally to detect problems with the chart by comparing the generated helm manifest with the kustomize version.

PR #4379, opened 2024-11-22
fix: ensure that github actions are not overwritten

camilamacedo86:
Users should be able to edit those files as please them. If we do not skip then, they might can lose their customizations.

We should not allow overwritten in this one


camilamacedo86:
It is a silly one.
Moving forward.

PR #3632, opened 2023-09-24
 📖 Proposal : New Plugin to allow project distribution via helm charts

dashanji:
As titled. 

This document is currently in RFC status, and corrections and additions are welcomed.

PR #4380, opened 2024-11-22
✨ feat: (helm/v1alpha1): Allow ServiceAccount annotations to be configurable

monteiro-renato:
The use case I have in mind is for example [Azure WI configuration](https://azure.github.io/azure-workload-identity/docs/topics/service-account-labels-and-annotations.html#service-account).

Sooner or later I think this way of massaging the yaml files is going to become burdensome.

### DIFF
--- a/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
+++ b/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
@@ -292,6 +292,19 @@ func copyFileWithHelmLogic(srcFile, destFile, subDir, projectName string) error 
 		contentStr = strings.Replace(contentStr,
 			"name: metrics-reader",
 			fmt.Sprintf("name: %s-metrics-reader", projectName), 1)
+		if strings.Contains(contentStr, "-controller-manager") &&
+			strings.Contains(contentStr, "kind: ServiceAccount") &&
+			!strings.Contains(contentStr, "RoleBinding") {
+			// The generated Service Account does not have the annotations field so we must add it.
+			contentStr = strings.Replace(contentStr,
+				"metadata:", `metadata:
+  {{- if and .Values.controllerManager.serviceAccount .Values.controllerManager.serviceAccount.annotations }}
+  annotations:
+    {{- range $key, $value := .Values.controllerManager.serviceAccount.annotations }}
+    {{ $key }}: {{ $value }}
+    {{- end }}
+  {{- end }}`, 1)
+		}
 		contentStr = strings.Replace(contentStr,
 			"name: leader-election-role",
 			fmt.Sprintf("name: %s-leader-election-role", projectName), -1)
--- a/testdata/project-v4-with-plugins/dist/chart/templates/rbac/service_account.yaml
+++ b/testdata/project-v4-with-plugins/dist/chart/templates/rbac/service_account.yaml
@@ -4,6 +4,12 @@ kind: ServiceAccount
 metadata:
   labels:
     {{- include "chart.labels" . | nindent 4 }}
+  {{- if and .Values.controllerManager.serviceAccount .Values.controllerManager.serviceAccount.annotations }}
+  annotations:
+    {{- range $key, $value := .Values.controllerManager.serviceAccount.annotations }}
+    {{ $key }}: {{ $value }}
+    {{- end }}
+  {{- end }}
   name: project-v4-with-plugins-controller-manager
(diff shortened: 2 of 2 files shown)


pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
```diff
@@ -292,6 +292,19 @@ func copyFileWithHelmLogic(srcFile, destFile, subDir, projectName string) error
...
+			!strings.Contains(contentStr, "RoleBinding") {
+			// The generated Service Account does not have the annotations field so we must add it.
+			contentStr = strings.Replace(contentStr,
+				"metadata:", `metadata:
+  {{- if and .Values.controllerManager.serviceAccount .Values.controllerManager.serviceAccount.annotations }}
+  annotations:
+    {{- range $key, $value := .Values.controllerManager.serviceAccount.annotations }}
+    {{ $key }}: {{ $value }}
+    {{- end }}
+  {{- end }}`, 1)
```
camilamacedo86, line 306: I fear we give too many options to custom things that do not exist in the project itself. The chart is only a way to package the staff. It should work the same way.


So, I will need to think about it a little

monteiro-renato: [This](https://github.com/kubernetes-sigs/kubebuilder/pull/3632/files#diff-dab08fd5a430f53bb558e2a9882c69a4d882d68bd6cff6160ed70ceb530b9644R540-R545), I guess?
I don't think I fully understand what is meant by **_"The proposed solution may not be capable of generating a fully valid and deployable Helm Chart for all use cases supported by Kubebuilder."_**. Is there a good resource for me to read on the supported Kubebuilder use cases?

For my team's use case we need to simplify Azure resources management and therefore need to talk to Azure APIs. To talk to the Azure APIs we need valid credentials. The best practice right now is to use [Workload Identity](https://learn.microsoft.com/en-us/entra/workload-id/workload-identities-overview#key-scenarios). Configuring a k8s workload to use Workload Identity requires a few annotations and labels in some resources (such as a [label in the pod spec](https://azure.github.io/azure-workload-identity/docs/topics/service-account-labels-and-annotations.html#pod) and an [annotation in the k8s service account](https://azure.github.io/azure-workload-identity/docs/topics/service-account-labels-and-annotations.html#annotations-1)).

I was able to do these configurations by using Kustomize patches, however, with our current setup, it's easier to deploy via Helm than it is to deploy k8s manifests so I wanted to test how I could achieve a similar goal with Helm.

monteiro-renato: @camilamacedo86, I don't think that's ideal.

[From the docs:](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) "**_You can use Kubernetes annotations to attach arbitrary non-identifying metadata to objects. Clients such as tools and libraries can retrieve this metadata._**"
In the example use case I mentioned above, [Azure](https://azure.github.io/azure-workload-identity/docs/topics/service-account-labels-and-annotations.html#service-account) uses the annotations in the Service Account to: "**_... configure the behavior when exchanging the service account token for an AAD access token:_**" The annotations would therefore only make sense in the context of a Service Account. 

Just like particular labels only make sense in the context of particular objects: ref my other PR > https://github.com/kubernetes-sigs/kubebuilder/pull/4373

camilamacedo86: Yes, we can add annotations in the resources, but it is not ONLY valid for ServiceAccount.
Did you check if it has other places that we need to do something ? 
That is my point.

camilamacedo86 (approved):
Hi @monteiro-renato 

Lets move forward with this one. But we might need to see if has no other places that we need to do the same.

pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
monteiro-renato: I see, I thought you were talking about applying the same annotation to all resources.

I believe it's best to add those features if/when they are needed (i.e. if someone has the need for it, they can open a PR to add that feature just like I'm doing here).

P.S: I plan to join the Slack channel eventually, which will make discussing these things a bit easier.

monteiro-renato:
Removed the extra spaces in the file to fix the golint error now.

camilamacedo86 (approved):
/lgtm

PR #4381, opened 2024-11-22
:seedling: feat: Add container check CI job

mateusoliveira43:
## Why the changes were made

Check that scaffolded container commands (`make docker-build`, `make docker-push` and `make docker-buildx`) always work.

The job checks that by building and pushing container images to [ttl.sh registry](https://ttl.sh/).

The job also checks that scaffolded Dockerfile follows linting rules from [hadolint](https://github.com/hadolint/hadolint).

Used [`lint-sample.yml`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/.github/workflows/lint-sample.yml) as base for new job.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -1,0 +1,39 @@
+name: Container Check
+
+on:
+  push:
+    paths-ignore:
+      - '**/*.md'
+  pull_request:
+    paths-ignore:
+      - '**/*.md'
+
+jobs:
+  container-check:
+    runs-on: ubuntu-latest
+    if: (github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository)
+    steps:
+      - name: Clone the code
+        uses: actions/checkout@v4
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '~1.22'
+
+      - name: Prepare project-v4
+        working-directory: testdata/project-v4
+        run: go mod tidy
+
+      - name: Build and push container image
+        working-directory: testdata/project-v4
+        run: IMG=ttl.sh/kubebuilder-container-check-$(git rev-parse --short HEAD):5m make docker-build docker-push
+
+      - name: Build and push container image with buildx
+        working-directory: testdata/project-v4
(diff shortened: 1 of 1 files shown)


.github/workflows/container.yml
```diff
@@ -0,0 +1,39 @@
+name: Container Check
+
+on:
+  push:
+    paths-ignore:
+      - '**/*.md'
+  pull_request:
+    paths-ignore:
+      - '**/*.md'
```
camilamacedo86, line 9: I think we should only call it if
- testdata/project-v4/Makefile
- testdata/project-v4/Dockerfile
- OR this file changes

mateusoliveira43: Agreed, will make change

.github/workflows/container.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Clone the code
+        uses: actions/checkout@v4
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '~1.22'
+
+      - name: Prepare project-v4
+        working-directory: testdata/project-v4
```
mateusoliveira43, line 25: could run a matrix here to check all 3 testdata folders (project-v4,  project-v4-multigroup  and project-v4-with-plugins). Needed or just one is enough?

.github/workflows/container.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: go mod tidy
```
camilamacedo86, line 26: The task is to use docker-buildx, check the Docker images, and validate whether the image has been generated for all platforms.

However, I believe this is related to the changes proposed in [this PR](https://github.com/kubernetes-sigs/kubebuilder/pull/4289).

WDYT about we do that in another way? 
What do you think about running the necessary commands in the PR to ensure everything is still working as expected after your changes? Additionally, you could include the results in the PR description to demonstrate that it works as intended.

mateusoliveira43: I can do that yes

But would not be nice to have this automated?

If not, all PR that involve container related changes would require one more step, right?

mateusoliveira43:
I think this PR is important because it tests scaffolded commands that were previously not tested

Issue #4378, opened 2024-11-22
(helm/v1alpha1): Prometheus integration is not working well when enabled in the Chart


camilamacedo86:
### What broke? What's expected?

We need to add in our GitHub action to test the Helm Chart: 
https://github.com/kubernetes-sigs/kubebuilder/pull/4377/files#diff-c93a79ad917924bb6a8d743a2c2b36706eeb045ed2e6b42b67b7928da9e7726aR108-R117

Then, we need to:
- We need to ensure that we replace prometheus: false with true in the values.yaml for the CI
- We need to fix the issue

```
Run helm install my-release ./dist/chart --create-namespace --namespace test-project-system
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: resource mapping not found for name: "test-project-controller-manager-metrics-monitor" namespace: "test-project-system" from "": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
ensure CRDs are installed first
Error: Process completed with exit code 1.
```

### KubeBuilder (CLI) Version

master

### Other versions

c/c @monteiro-renato

monteiro-renato:
Hey 👋 
This is "expected" if the [CRD](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/design.md#servicemonitor) hasn't been deployed to the cluster first. 
That has been mentioned in the [design doc](https://github.com/kubernetes-sigs/kubebuilder/blob/ab96f9a0ed8cb9cda2f1d5a9fccf001b697d50ac/designs/helm-chart-autogenerate-plugin.md?plain=1#L126-L133) as well.

monteiro-renato:
I can add the missing bits to the Github Action.
Simply installing the CRDs as a dependency should fix the problem.
There's an [helm for the CRDs](https://github.com/prometheus-community/helm-charts/pkgs/container/charts%2Fprometheus-operator-crds), no need to install the bundle.

monteiro-renato:
Ready for review!

PR #4383, opened 2024-11-22
🐛 fix: (helm/v1alpha1): install the prometheus-operator CRDs before installing a chart which configures a ServiceMonitor

monteiro-renato:
Fixes: https://github.com/kubernetes-sigs/kubebuilder/issues/4378

### DIFF
--- a/.github/workflows/test-helm-samples.yml
+++ b/.github/workflows/test-helm-samples.yml
@@ -3,12 +3,12 @@ name: Helm Testdata Sample
 on:
   push:
     paths:
-      - 'testdata/project-v4-with-plugins/**'
-      - '.github/workflows/test-helm-samples.yml'
+      - "testdata/project-v4-with-plugins/**"
+      - ".github/workflows/test-helm-samples.yml"
   pull_request:
     paths:
-      - 'testdata/project-v4-with-plugins/**'
-      - '.github/workflows/test-helm-samples.yml'
+      - "testdata/project-v4-with-plugins/**"
+      - ".github/workflows/test-helm-samples.yml"
 
 jobs:
   helm-test-project-v4-with-plugins:
@@ -23,7 +23,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: '~1.22'
+          go-version: "~1.22"
 
       - name: Install the latest version of kind
         run: |
@@ -55,6 +55,12 @@ jobs:
         run: |
           helm lint testdata/project-v4-with-plugins/dist/chart
 
+      - name: Install Prometheus Operator CRDs
+        run: |
+          helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+          helm repo update
+          helm install prometheus-crds prometheus-community/prometheus-operator-crds
+
       - name: Install cert-manager via Helm
         run: |
           helm repo add jetstack https://charts.jetstack.io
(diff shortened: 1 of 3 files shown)


.github/workflows/test-helm-samples.yml
```diff
@@ -69,8 +75,12 @@ jobs:
...
-          helm install my-release testdata/project-v4-with-plugins/dist/chart --create-namespace --namespace project-v4-with-plugins-system
+          helm install my-release testdata/project-v4-with-plugins/dist/chart --create-namespace --namespace project-v4-with-plugins-system --set prometheus.enable=true
 
       - name: Check Helm release status
         run: |
           helm status my-release --namespace project-v4-with-plugins-system
+
+      - name: Check Presence of ServiceMonitor
+        run: |
+          kubectl wait --namespace project-v4-with-plugins-system --for=jsonpath='{.kind}'=ServiceMonitor servicemonitor/project-v4-with-plugins-controller-manager-metrics-monitor
```
monteiro-renato, line 86: Fails with exit 1 code if the resource does not exist.

camilamacedo86: I think we have a bug.
For we are able to check it out we need enable prometheus in the HelmChart value.

By default it is

```
prometheus:
  enable: false
```

we need to use shell to change it in the GitHUb action to set true
Something like example: https://github.com/kubernetes-sigs/kubebuilder/blob/master/.github/workflows/test-e2e-samples.yml#L43-L44

Then the error will be faced
We might either to change : 
pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/github/test_chart.go

OR the chart

monteiro-renato: I'm passing the value to the helm install command in line 78 :)
Check the --set flag of the [helm install](https://helm.sh/docs/helm/helm_install/)

Kavinjsir: nit: how about adding an explicit timeout flag for `kubectl wait`:
```
--timeout=60s
```

Also, `--for=''` could be another choice to consider?

Kavinjsir (approved):
Looks good to me!

camilamacedo86 (commented):
Hi @monteiro-renato 

We also need to change the github action taht we are scaffolding for the projects, see; https://github.com/kubernetes-sigs/kubebuilder/pull/4377/files#diff-c93a79ad917924bb6a8d743a2c2b36706eeb045ed2e6b42b67b7928da9e7726aR108-R117

.github/workflows/test-helm-samples.yml
monteiro-renato: This is a bit of an hack since the kubectl wait is supposed to be used with resources that have a Status field.
The command fails right away if the resource doesn't exist so the `--timeout` flag isn't really needed here (I guess `--timeout=0` would be ok).

If `kubectl get` on a non existent resource would somehow return a non zero exit code that's what I would have used but I don't think that's possible. And between processing the output of `kubectl get` and simply using the hack above, I think the hack above is easier to understand.


The `--for=''` could be an option yea; my thought was that it could lead to confusion as it might be less clear.

camilamacedo86 (changes requested):
Just I nit all the rest shows great 🥇 
Terrific work

See: https://github.com/kubernetes-sigs/kubebuilder/pull/4383/files#r1855209081

pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/github/test_chart.go
```diff
@@ -118,9 +124,13 @@ jobs:
 
       - name: Install Helm chart for project
         run: |
-          helm install my-release ./dist/chart --create-namespace --namespace {{ .ProjectName }}-system
+          helm install my-release ./dist/chart --create-namespace --namespace {{ .ProjectName }}-system --set prometheus.enable=true
```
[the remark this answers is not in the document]
monteiro-renato: Yep updated it now 👍

testdata/project-v4-with-plugins/.github/workflows/test-chart.yml
```diff
@@ -46,6 +46,12 @@ jobs:
         run: |
           helm lint ./dist/chart
 
+      - name: Install Prometheus Operator CRDs
+        run: |
+          helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+          helm repo update
+          helm install prometheus-crds prometheus-community/prometheus-operator-crds
+
```
camilamacedo86, line 54: @monteiro-renato 
this one need to be commented and under # TODO: Uncomment if Prometheus is enabled
We should only install it if the chart enabled the option

monteiro-renato: I thought it would be a bit easier to just install the CRDs all the time as they shouldn't add too much time to the action run.

I've shifted them to be under the TODO now 👍.

camilamacedo86 (approved):
/lgtm
/approved
/ok-to-test

Issue #4376, opened 2024-11-22
Project scaffold by default has some lint errors which must be fixed


camilamacedo86:
### What broke? What's expected?

I expected to create a project and push it for the repo and see all actions passing successfully
However, the lint is broken

```
Running [/home/runner/golangci-lint-1.61.0-linux-amd64/golangci-lint run] in [/home/runner/work/test-project/test-project] ...
  Error: internal/webhook/v1alpha1/busybox_webhook.go:84:2: comment-spacings: no space between comment delimiter and comment text (revive)
  	//TODO(user): Add more fields as needed for validation
  	^
  Error: test/utils/utils.go:95:36: unnecessary conversion (unconvert)
  	crdList := GetNonEmptyLines(string(output))
  	                                  ^
  Error: test/utils/utils.go:156:36: unnecessary conversion (unconvert)
  	crdList := GetNonEmptyLines(string(output))
  	                                  ^
  Error: test/e2e/e2e_test.go:319:30: unnecessary conversion (unconvert)
  		err = json.Unmarshal([]byte(output), &token)
  		                           ^
  Error: internal/controller/busybox_controller.go:102:5: S1009: should omit nil check; len() for []k8s.io/apimachinery/pkg/apis/meta/v1.Condition is defined as zero (gosimple)
  	if busybox.Status.Conditions == nil || len(busybox.Status.Conditions) == 0 {
  	   ^
  Error: test/e2e/e2e_test.go:96:38: printf: non-constant format string in call to fmt.Fprintf (govet)
  				_, _ = fmt.Fprintf(GinkgoWriter, fmt.Sprintf("Controller logs:\n %s", controllerLogs))
  				                                 ^
  Error: test/e2e/e2e_test.go:98:38: printf: non-constant format string in call to fmt.Fprintf (govet)
  				_, _ = fmt.Fprintf(GinkgoWriter, fmt.Sprintf("Failed to get Controller logs: %s", err))
  				                                 ^
  Error: test/e2e/e2e_test.go:105:38: printf: non-constant format string in call to fmt.Fprintf (govet)
  				_, _ = fmt.Fprintf(GinkgoWriter, fmt.Sprintf("Kubernetes events:\n%s", eventsOutput))
  				                                 ^
  
  Error: issues found
  Ran golangci-lint in 1037[28](https://github.com/camilamacedo86/test-project/actions/runs/11970757483/job/33374161657#step:4:30)ms
```

[...]

PR #4384, opened 2024-11-22
🐛 : (go/v4) fixes based on golangci-lint run -v

monteiro-renato:
Fixes: https://github.com/kubernetes-sigs/kubebuilder/issues/4376

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/internal/webhook/v1/cronjob_webhook.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/internal/webhook/v1/cronjob_webhook.go
@@ -162,7 +162,7 @@ This marker is responsible for generating a validation webhook manifest.
 // NOTE: The +kubebuilder:object:generate=false marker prevents controller-gen from generating DeepCopy methods,
 // as this struct is used only for temporary operations and does not need to be deeply copied.
 type CronJobCustomValidator struct {
-	//TODO(user): Add more fields as needed for validation
+	// TODO(user): Add more fields as needed for validation
 }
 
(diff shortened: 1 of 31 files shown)


monteiro-renato:
I also tried bumping the version of golangci-lint to 1.62  but the lint action started breaking with:
```
  Error: testdata/project-v4/test/utils/utils.go:40:21: undefined: GinkgoWriter (typecheck)
  	_, _ = fmt.Fprintf(GinkgoWriter, "warning: %v\n", err)
```

camilamacedo86:
THANK you a lot 🥇 
You are been a rock star 🚀 


By the way, 🌱 we just use when it is something that does not change anything for end users

camilamacedo86 (approved):
/lgtm
/ok-to-test

monteiro-renato:
This seems to be mergeable but I think I found another bug.
The e2e tests on the test setup I created based on your commands are failing due to the container failing to start.
```
NAME                                          READY   STATUS              RESTARTS   AGE
kb-test-controller-manager-6fd7dfb975-2dqkq   0/1     ContainerCreating   0          2m7s
```
Describing the pod displays:
```
Warning  FailedMount  29s (x9 over 2m45s)  kubelet            MountVolume.SetUp failed for volume "cert" : secret "webhook-server-cert" not found
```
Looks like there's further troubleshooting to do.

monteiro-renato:
Might not actually be a bug. Might be some missing manual configs (which may or may not be documented in the book, I haven't checked).

But yea, simply running those three commands you posted in the issue and pushing the code to Github doesn't seem enough to make the e2e tests happy.

PR #4386, opened 2024-11-24
✨ (helm/v1alpha1): Add a step to render the chart in the GitHub Action

monteiro-renato:
Related to > https://github.com/kubernetes-sigs/kubebuilder/pull/4377

If you agree it's useful, I can add it to the  test_chart.go as well.

Note for example in the picture below that some resources are not being prefixed (The issue I mentioned here: https://github.com/kubernetes-sigs/kubebuilder/pull/4357)

bilde


camilamacedo86:
Yes I loved it 👍 
/lgtm
/approve
/ok-to-test

PR #4388, opened 2024-11-25
🐛 fix: (helm/v1alpha1): Add missing prefixes in various k8s resources

monteiro-renato:
Hey 👋
I've added missing prefixes to k8s resources (except for cert-manager ones)
I've added 4 commits so it's easier to review.
I will delete the [helm-install.yaml](https://github.com/monteiro-renato/kubebuilder/blob/088436bd8f1b63c793a7f81f32301679f8998e8b/testdata/project-v4-with-plugins/dist/helm-install.yaml) file and squash the commits after the review.

### DIFF
--- a/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
+++ b/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
@@ -293,6 +293,14 @@ func copyFileWithHelmLogic(srcFile, destFile, subDir, projectName string) error 
 		contentStr = strings.Replace(contentStr,
 			"name: metrics-reader",
 			fmt.Sprintf("name: %s-metrics-reader", projectName), 1)
+
+		contentStr = strings.Replace(contentStr,
+			"name: metrics-auth-role",
+			fmt.Sprintf("name: %s-metrics-auth-role", projectName), -1)
+		contentStr = strings.Replace(contentStr,
+			"name: metrics-auth-rolebinding",
+			fmt.Sprintf("name: %s-metrics-auth-rolebinding", projectName), 1)
+
 		if strings.Contains(contentStr, "-controller-manager") &&
 			strings.Contains(contentStr, "kind: ServiceAccount") &&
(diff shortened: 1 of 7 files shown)


testdata/project-v4-with-plugins/dist/helm-install.yaml
```diff
@@ -0,0 +1,1161 @@
+---
```
camilamacedo86, line 1: I understand that you added this to make it easier to review the changes, and it seems like an interesting approach.

However, it appears this was done manually. All the files generated under testdata are created using the shell script found at [generate.sh](https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh). We cannot manually add or modify content in this directory. Instead, the logic for generating these files needs to be incorporated into the script. Also, the action that calls it will need to have helm installed. We check if the testdata is updated with the PR changes to allow get it merged see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/.github/workflows/testdata.yml

Additionally, if we decide to include this change, we must ensure that the helm-install.yaml file is re-generated when running the [Makefile target](https://github.com/kubernetes-sigs/kubebuilder/blob/master/Makefile#L91-L94). This ensures consistency and alignment with the existing build and test automation processes.

camilamacedo86: I think this change would need to be in an PR only for that.

camilamacedo86:
Hi @monteiro-renato 

Thank you for looking on this. See my comment: https://github.com/kubernetes-sigs/kubebuilder/pull/4388#discussion_r1856393007

I think we need to split this PR, and for we get merged, we should have only one commit, ideally.

testdata/project-v4-with-plugins/dist/helm-install.yaml
monteiro-renato: Hi @camilamacedo86,
I've used `make generate` to update the files in the testdata folder, it was not done manually. Those files are being done based on the changes I've made to the .go files.

PR #4393, opened 2024-11-26
🐛 fix: (helm/v1alpha1): Fix brackets in _helpers.tpl

monteiro-renato:
I haven't tested rendering the final manifest to see how it looks like but the brackets didn't add up.

bilde


camilamacedo86:
Hi @monteiro-renato 

is that a PR that we still need, or should we close this one?

monteiro-renato:
Haven't had the time to look into this so I guess we can close yea.

PR #4394, opened 2024-11-26
:book: fix: Remove default duplication in help

mateusoliveira43:
Add default info to flags

Before this PR
```console
$ kubebuilder -h
...
configuration please run:
    kubebuilder init --help --plugins=<PLUGIN KEYS> [--project-version=<PROJECT VERSION>]

Default plugin keys: "go.kubebuilder.io/v4"
Default project version: "3"


...

Flags:
  -h, --help                     help for kubebuilder
      --plugins strings          plugin keys to be used for this subcommand execution

...
```

After this PR
```console
$ kubebuilder -h
...
configuration please run:
    kubebuilder init --help --plugins=<PLUGIN KEYS> [--project-version=<PROJECT VERSION>]

...

Flags:
  -h, --help                     help for kubebuilder
      --plugins strings          plugin keys to be used for this subcommand execution (default [go.kubebuilder.io/v4])

...
```

### DIFF
--- a/pkg/cli/root.go
+++ b/pkg/cli/root.go
@@ -44,7 +44,7 @@ func (c CLI) newRootCmd() *cobra.Command {
 	}
 
 	// Global flags for all subcommands.
-	cmd.PersistentFlags().StringSlice(pluginsFlag, nil, "plugin keys to be used for this subcommand execution")
+	cmd.PersistentFlags().StringSlice(pluginsFlag, c.defaultPlugins[c.defaultProjectVersion], "plugin keys to be used for this subcommand execution")
 
 	// Register --project-version on the root command so that it shows up in help.
 	cmd.Flags().String(projectVersionFlag, c.defaultProjectVersion.String(), "project version")
@@ -58,7 +58,7 @@ func (c CLI) newRootCmd() *cobra.Command {
 
 // rootExamples builds the examples string for the root command before resolving plugins
 func (c CLI) rootExamples() string {
-	str := fmt.Sprintf(`The first step is to initialize your project:
+	return fmt.Sprintf(`The first step is to initialize your project:
     %[1]s init [--plugins=<PLUGIN KEYS> [--project-version=<PROJECT VERSION>]]
 
 <PLUGIN KEYS> is a comma-separated list of plugin keys from the following table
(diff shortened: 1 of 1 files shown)


pkg/cli/root.go
```diff
@@ -68,21 +68,8 @@ and <PROJECT VERSION> a supported project version for these plugins.
...
-    %[1]s init --help --plugins=<PLUGIN KEYS> [--project-version=<PROJECT VERSION>]
-`,
+    %[1]s init --help --plugins=<PLUGIN KEYS> [--project-version=<PROJECT VERSION>]`,
 		c.commandName, c.getPluginTable())
-
-	if len(c.defaultPlugins) != 0 {
-		if defaultPlugins, found := c.defaultPlugins[c.defaultProjectVersion]; found {
-			str += fmt.Sprintf("\nDefault plugin keys: %q\n", strings.Join(defaultPlugins, ","))
-		}
-	}
```
camilamacedo86, line 79: Why would you like to remove it?

That output provides the default plugin used by the tool, which is particularly helpful when dealing with multiple plugin versions. Now, we have only go/v4 but before the cut of 4.0.0 we have go/v2, go/v3 and go.v4.

So it is helpful. Additionally, keep in mind that Kubebuilder is also used as a library by other tooling, which may aim to achieve similar functionality with additional features or external plugins layered on top. Therefore, it's important to consider these broader scenarios as well.

pkg/cli/root.go
```diff
@@ -68,21 +68,8 @@ and <PROJECT VERSION> a supported project version for these plugins.
...
-
-	if c.defaultProjectVersion.Validate() == nil {
-		str += fmt.Sprintf("Default project version: %q\n", c.defaultProjectVersion)
-	}
```
camilamacedo86, line 83: Why would we remove it?

camilamacedo86 (changes requested):
Hi @mateusoliveira43 

But in this case, the text:

```
Default plugin keys: "go.kubebuilder.io/v4"
Default project version: "3"
```

It is not duplicated info.
- We can have 1 to Many plugins
- We can have 1 to Many Project Config versions supported

So, the information is important to let us know what the default scaffold is when we do not inform it using the—-plugin flag.

Therefore, we can not accept this one. 
Could we agree to close this one?

pkg/cli/root.go
mateusoliveira43: This info is already present in the description of the flag
```go
```
the second parameter here is the default value of the flag
https://github.com/kubernetes-sigs/kubebuilder/blob/cbc6e383c342f1337ab37ee4aa0755957a01f9c7/pkg/cli/root.go#L49-L50
which is used to generate the flag description

pkg/cli/root.go
mateusoliveira43: with this PR, we will be using the default value parameter here
https://github.com/kubernetes-sigs/kubebuilder/blob/cbc6e383c342f1337ab37ee4aa0755957a01f9c7/pkg/cli/root.go#L47
so info will not need to be printed by another function

It will look like this when calling help
```
--plugins strings          plugin keys to be used for this subcommand execution (default [go.kubebuilder.io/v4])
```

camilamacedo86: I think we should keep

```
Default plugin keys: "go.kubebuilder.io/v4"
Default project version: "3"
```

That makes clear what are the default plugin and project version for the tool
It does not make too much because currently, here we do have not too many versions

camilamacedo86:
Hi @mateusoliveira43 

Thank you for collaborating on this one.
But due to the comment https://github.com/kubernetes-sigs/kubebuilder/pull/4394#pullrequestreview-2464544470, I think we can convey not to move forward.

So, I hope that you do not mind I am closing this one. 
However, if you do not agree please feel free to re-open.

Issue #4291, opened 2024-11-05
Refactor `CronJob` and `MultiVersion` Sample Code to Align with `DeployImage` Plugin Scaffold Structure


camilamacedo86:
### What do you want to happen?

This issue proposes refactoring the [CronJob](https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/cronjob-tutorial) and [MultiVersion](https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/multiversion-tutorial) samples to follow the structure and best practices introduced by the [`DeployImage` plugin](https://book.kubebuilder.io/plugins/available/deploy-image-plugin-v1-alpha), which includes scaffolding to manage resources defined via command-line inputs, such as images.

### Objectives
The main goals for this refactoring include ensure:

1. Align the structure of the `CronJob` and `MultiVersion` examples with the scaffolded layout used by the `DeployImage` plugin.
2. Apply best practices, such as fetching the resource before updates, to prevent reconciliation conflicts due to intermediate changes.
3. Integrate status conditions to better track resource states. (see that we have indeed a task for that: https://github.com/kubernetes-sigs/kubebuilder/issues/4019)
4. Ensure that test implementations mirror the `DeployImage` scaffolded layout, with comprehensive tests for the controller logic.(see that we have indeed a task for that: https://github.com/kubernetes-sigs/kubebuilder/issues/4140)
5. Retain the original logic and functionality of the `CronJob` and `MultiVersion` examples.
6. Preserve the existing markers to ensure that documentation snippets are generated and rendered consistently in the docs.

### Reference Materials
To align with `DeployImage` scaffolding, please review the following examples:
- **Controller Implementation:** [busybox_controller.go](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-with-plugins/internal/controller/busybox_controller.go)
- **Controller Test:** [busybox_controller_test.go](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-with-plugins/internal/controller/busybox_controller_test.go#L139-L146)

### Additional Notes
All sample updates should be implemented to ensure they match the `DeployImage` plugin’s scaffold structure and best practices, while still fulfilling the unique requirements of each sample (i.e., `CronJob` and `MultiVersion` functionality).

Related tasks:

- https://github.com/kubernetes-sigs/kubebuilder/issues/4019
[...]

PR #4397, opened 2024-11-26
📖 refactor: finalizer in CronJob example 

lorenzofelletti:
## Description
Refactor `CronJob` sample code to follow the structure and best practices introduced by the [`DeployImage` plugin](https://book.kubebuilder.io/plugins/available/deploy-image-plugin-v1-alpha).

## Motivation
Follow best practices, and provide an example that works without incurring in errors like trying to do a second update without re-fetching the resource.

## References
- #4291.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
+++ b/docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
@@ -67,7 +67,12 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		// then lets add the finalizer and update the object. This is equivalent
 		// to registering our finalizer.
 		if !controllerutil.ContainsFinalizer(cronJob, myFinalizerName) {
-			controllerutil.AddFinalizer(cronJob, myFinalizerName)
+			log.Info("Adding Finalizer for CronJob")
+			if ok := controllerutil.AddFinalizer(cronJob, myFinalizerName); !ok {
+				log.Error(err, "Failed to add finalizer into the custom resource")
+				return ctrl.Result{Requeue: true}, nil
+			}
+
 			if err := r.Update(ctx, cronJob); err != nil {
 				return ctrl.Result{}, err
 			}
--- a/docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
+++ b/docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
@@ -76,6 +76,14 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			if err := r.Update(ctx, cronJob); err != nil {
 				return ctrl.Result{}, err
 			}
+
+			// we re-fetch after having updated the CronJob, so that we have the latest
+			// state of the resource, and avoid the error "the object has been modified,
+			// please apply your changes to the latest version and try again".
+			if err := r.Get(ctx, req.NamespacedName, cronJob); err != nil {
+				log.Error(err, "unable to fetch CronJob")
+				return ctrl.Result{}, client.IgnoreNotFound(err)
+			}
 		}
 	} else {
 		// The object is being deleted
--- a/docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller.go
@@ -291,7 +291,7 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		The status subresource ignores changes to spec, so it's less likely to conflict
 		with any other updates, and can have separate permissions.
 	*/
-	if err := r.Status().Update(ctx, &cronJob); err != nil {
+	if err := r.Status().Update(ctx, cronJob); err != nil {
 		log.Error(err, "unable to update CronJob status")
 		return ctrl.Result{}, err
 	}
@@ -432,7 +432,7 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 	// figure out the next times that we need to create
 	// jobs at (or anything we missed).
-	missedRun, nextRun, err := getNextSchedule(&cronJob, r.Now())
+	missedRun, nextRun, err := getNextSchedule(cronJob, r.Now())
 	if err != nil {
 		log.Error(err, "unable to figure out CronJob schedule")
 		// we don't really care about requeuing until we get an update that
@@ -536,7 +536,7 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// +kubebuilder:docs-gen:collapse=constructJobForCronJob
 
 	// actually make the job...
-	job, err := constructJobForCronJob(&cronJob, missedRun)
+	job, err := constructJobForCronJob(cronJob, missedRun)
 	if err != nil {
 		log.Error(err, "unable to construct job from template")
 		// don't bother requeuing until we get a change to the spec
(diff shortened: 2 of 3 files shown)


docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
```diff
@@ -67,10 +67,23 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		// then lets add the finalizer and update the object. This is equivalent
 		// to registering our finalizer.
 		if !controllerutil.ContainsFinalizer(cronJob, myFinalizerName) {
-			controllerutil.AddFinalizer(cronJob, myFinalizerName)
+			log.Info("Adding Finalizer for CronJob")
+			if ok := controllerutil.AddFinalizer(cronJob, myFinalizerName); !ok {
+				log.Error(err, "Failed to add finalizer into the custom resource")
+				return ctrl.Result{Requeue: true}, nil
```
camilamacedo86, line 73: we could just return the error.

docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
```diff
@@ -67,10 +67,23 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 				return ctrl.Result{}, err
 			}
+
+			// we re-fetch after having updated the CronJob, so that we have the latest
+			// state of the resource, and avoid the error "the object has been modified,
+			// please apply your changes to the latest version and try again".
+			if err := r.Get(ctx, req.NamespacedName, cronJob); err != nil {
+				log.Error(err, "unable to fetch CronJob")
+				return ctrl.Result{}, client.IgnoreNotFound(err)
+			}
```
camilamacedo86, line 86: We need to change the code under hack/docs so that when we run `make generate-docs` we re-scaffold all and just add the right code on top to ensure that all is always updated.

camilamacedo86: Indeed it will result in multi-version and crojob tutorial changed when you run `make generate-docs`

camilamacedo86 (changes requested):
Really thank you to looking on that, see: https://github.com/kubernetes-sigs/kubebuilder/pull/4397/files#r1859746824

docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
mateusoliveira43, line 73: would not be semantically better to error out here? For example
```go
return ctrl.Result{}, fmt.Errorf("message")
```

docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
```diff
@@ -67,10 +67,23 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+
 			if err := r.Update(ctx, cronJob); err != nil {
 				return ctrl.Result{}, err
 			}
+
+			// we re-fetch after having updated the CronJob, so that we have the latest
+			// state of the resource, and avoid the error "the object has been modified,
+			// please apply your changes to the latest version and try again".
+			if err := r.Get(ctx, req.NamespacedName, cronJob); err != nil {
+				log.Error(err, "unable to fetch CronJob")
```
mateusoliveira43, line 84: would be better to say `re-fetch` here?

docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
```diff
@@ -67,10 +67,23 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+				return ctrl.Result{}, client.IgnoreNotFound(err)
```
mateusoliveira43, line 85: would not be better to always error out here?

camilamacedo86: Yes you are right. We need to change it in the deploy image too.

docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
mateusoliveira43: sure, will try to make changes there

docs/book/src/cronjob-tutorial/testdata/finalizer_example.go
lorenzofelletti: may I ask why? If in the meantime the resource was deleted shouldn't that be ignored?

camilamacedo86: On top we should have : https://github.com/kubernetes-sigs/kubebuilder/blob/cbc6e383c342f1337ab37ee4aa0755957a01f9c7/testdata/project-v4-with-plugins/internal/controller/busybox_controller.go#L84C2-L99C3

If the resource is deleted, then the reconciliation stop at this point.

What we need to do is fix the tutorials to have the structure of https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-with-plugins/internal/controller/busybox_controller.go

The changes here seems for me either incomplete if we address only the finalizer

lorenzofelletti:
Since there's a lot of rework to do on this one, I'll convert it to a draft for the moment

docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller.go
```diff
@@ -111,15 +121,41 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		}
+
+		// Re-fetch the CronJob after updating the status so that we have
+		// the latest state of the resource on the cluster and avoid raising
+		// an error should we try to update it again in the following operations
+		if err := r.Get(ctx, req.NamespacedName, cronJob); err != nil {
+			log.Error(err, "Failed to re-fetch CronJob")
+			return ctrl.Result{}, err
+		}
+	}
```
camilamacedo86, line 155: @lorenzofelletti Yep, that is the idea!!!
Thank you a lot

lorenzofelletti:
don't have the permissions to re-open it haha

docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller.go
```diff
@@ -500,7 +536,7 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// +kubebuilder:docs-gen:collapse=constructJobForCronJob
 
 	// actually make the job...
-	job, err := constructJobForCronJob(&cronJob, missedRun)
+	job, err := constructJobForCronJob(cronJob, missedRun)
```
camilamacedo86, line 539: Hi @lorenzofelletti

In a way that I would do that is: (see that this task is not so trivial, you need attention to details and use a lot the IDE to compare)

- Do the changes in the testdata samples directly as you are doing
- It is possible compare the code implementation of the book samples with the samples under [testdata/p](https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v4-with-plugins) see the code for Memcached and busybox which uses the deploy image plugin and is generated by: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh

Then, you can push a PR against your on fork.
See that the samples are tested in the CI here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/.github/workflows/test-e2e-book.yml


If this workflow passes, then you can use the code changes.
Note that the test data check will fail because we generate the code of the book samples by running the commands and injecting the code on top. The whole code implementation for it is here: https://github.com/kubernetes-sigs/kubebuilder/tree/master/hack/docs/internal


When we call make generate-docs, the scripts are called, the projects generated, and the changes made afterwards; it ensures that the docs are always updated with the latest changes. 

But we can do it step by step.

lorenzofelletti: Hi @camilamacedo86!

Thanks for taking the time to write these instructions. I'll follow them and come back when I have something to share.

PR #4399, opened 2024-11-29
✨ (deployimage/v1alpha1): Improve error handling and pointer usage for value setting in controller

mateusoliveira43:
Fix https://github.com/kubernetes-sigs/kubebuilder/pull/4397#discussion_r1861344680

### DIFF
--- a/testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
+++ b/testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -123,8 +124,9 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	if !controllerutil.ContainsFinalizer(memcached, memcachedFinalizer) {
 		log.Info("Adding Finalizer for Memcached")
 		if ok := controllerutil.AddFinalizer(memcached, memcachedFinalizer); !ok {
-			log.Error(err, "Failed to add finalizer into the custom resource")
-			return ctrl.Result{Requeue: true}, nil
+			err = fmt.Errorf("finalizer for Memcached was not added")
+			log.Error(err, "Failed to add finalizer for Memcached")
+			return ctrl.Result{}, err
 		}
 
 		if err = r.Update(ctx, memcached); err != nil {
@@ -178,8 +180,9 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 			log.Info("Removing Finalizer for Memcached after successfully perform the operations")
 			if ok := controllerutil.RemoveFinalizer(memcached, memcachedFinalizer); !ok {
+				err = fmt.Errorf("finalizer for Memcached was not removed")
 				log.Error(err, "Failed to remove finalizer for Memcached")
-				return ctrl.Result{Requeue: true}, nil
+				return ctrl.Result{}, err
 			}
 
 			if err := r.Update(ctx, memcached); err != nil {
@@ -226,7 +229,7 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{RequeueAfter: time.Minute}, nil
 	} else if err != nil {
 		log.Error(err, "Failed to get Deployment")
-		// Let's return the error for the reconciliation be re-trigged again
+		// Let's return the error for the reconciliation be re-triggered again
 		return ctrl.Result{}, err
 	}
 
@@ -358,7 +361,7 @@ func (r *MemcachedReconciler) deploymentForMemcached(
 					//	 },
 					// },
 					SecurityContext: &corev1.PodSecurityContext{
-						RunAsNonRoot: &[]bool{true}[0],
+						RunAsNonRoot: ptr.To(true),
 						// IMPORTANT: seccomProfile was introduced with Kubernetes 1.19
 						// If you are looking for to produce solutions to be supported
 						// on lower versions you must remove this option.
@@ -373,9 +376,9 @@ func (r *MemcachedReconciler) deploymentForMemcached(
 						// Ensure restrictive context for the container
 						// More info: https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted
 						SecurityContext: &corev1.SecurityContext{
-							RunAsNonRoot:             &[]bool{true}[0],
-							RunAsUser:                &[]int64{1001}[0],
-							AllowPrivilegeEscalation: &[]bool{false}[0],
+							RunAsNonRoot:             ptr.To(true),
+							RunAsUser:                ptr.To(int64(1001)),
+							AllowPrivilegeEscalation: ptr.To(false),
 							Capabilities: &corev1.Capabilities{
 								Drop: []corev1.Capability{
 									"ALL",
@@ -409,7 +412,8 @@ func labelsForMemcached() map[string]string {
 	if err == nil {
 		imageTag = strings.Split(image, ":")[1]
 	}
-	return map[string]string{"app.kubernetes.io/name": "project-v4-with-plugins",
+	return map[string]string{
+		"app.kubernetes.io/name":       "project-v4-with-plugins",
 		"app.kubernetes.io/version":    imageTag,
 		"app.kubernetes.io/managed-by": "MemcachedController",
 	}
--- a/docs/book/src/getting-started/testdata/project/go.mod
+++ b/docs/book/src/getting-started/testdata/project/go.mod
@@ -8,6 +8,7 @@ require (
 	k8s.io/api v0.31.0
 	k8s.io/apimachinery v0.31.0
 	k8s.io/client-go v0.31.0
+	k8s.io/utils v0.0.0-20240711033017-18e509b52bc8
 	sigs.k8s.io/controller-runtime v0.19.1
 )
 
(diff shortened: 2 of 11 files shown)


testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
mateusoliveira43, line 1: @camilamacedo86 can you take a look? After we agree on the refactor points, I will apply changes to the template and generate code

testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
```diff
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/ptr"
```
camilamacedo86, line 34: Why we need a new dep?
I thought if we were here very surgically just change the place that would be better to create a New Error and return it instead of using requeue true

testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
```diff
@@ -117,37 +118,22 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 	}
 
-	// Let's add a finalizer. Then, we can define some operations which should
-	// occur before the custom resource is deleted.
-	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers
-	if !controllerutil.ContainsFinalizer(memcached, memcachedFinalizer) {
```
camilamacedo86, line 123: @mateusoliveira43 Why remove the func from controller-runtime used to add the finalizer we should keep it as it is

testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
```diff
@@ -117,37 +118,22 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
-		log.Info("Adding Finalizer for Memcached")
-		if ok := controllerutil.AddFinalizer(memcached, memcachedFinalizer); !ok {
-			log.Error(err, "Failed to add finalizer into the custom resource")
-			return ctrl.Result{Requeue: true}, nil
```
camilamacedo86, line 127: The only change to enhance would be here right?
Just create the error and return in the result to re-queue

testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
```diff
@@ -117,37 +118,22 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 	// indicated by the deletion timestamp being set.
-	isMemcachedMarkedToBeDeleted := memcached.GetDeletionTimestamp() != nil
-	if isMemcachedMarkedToBeDeleted {
+	if !memcached.GetDeletionTimestamp().IsZero() {
 		if controllerutil.ContainsFinalizer(memcached, memcachedFinalizer) {
 			log.Info("Performing Finalizer Operations for Memcached before delete CR")
 
 			// Let's add here a status "Downgrade" to reflect that this resource began its process to be terminated.
-			meta.SetStatusCondition(&memcached.Status.Conditions, metav1.Condition{Type: typeDegradedMemcached,
+			changed := meta.SetStatusCondition(&memcached.Status.Conditions, metav1.Condition{Type: typeDegradedMemcached,
```
camilamacedo86, line 128: Why add this changed? 
We set the status and we update it directly as in other places

testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
```diff
@@ -167,19 +154,21 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 				return ctrl.Result{}, err
 			}
 
-			meta.SetStatusCondition(&memcached.Status.Conditions, metav1.Condition{Type: typeDegradedMemcached,
+			changed = meta.SetStatusCondition(&memcached.Status.Conditions, metav1.Condition{Type: typeDegradedMemcached,
```
camilamacedo86, line 157: The changed is not required necessary
We are updating the status and we are calling update always that it needs to be updated

testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
```diff
@@ -190,23 +179,47 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		}
+
+		// TODO GET call?
+	}
+
 	// Check if the deployment already exists, if not create a new one
 	found := &appsv1.Deployment{}
 	err = r.Get(ctx, types.NamespacedName{Name: memcached.Name, Namespace: memcached.Namespace}, found)
-	if err != nil && apierrors.IsNotFound(err) {
+	if err != nil {
```
camilamacedo86, line 203: Why do this change?

testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
```diff
@@ -167,19 +154,21 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+					return ctrl.Result{}, err
+				}
+				// TODO GET call?
 			}
 
 			log.Info("Removing Finalizer for Memcached after successfully perform the operations")
 			if ok := controllerutil.RemoveFinalizer(memcached, memcachedFinalizer); !ok {
 				log.Error(err, "Failed to remove finalizer for Memcached")
-				return ctrl.Result{Requeue: true}, nil
+				return ctrl.Result{}, fmt.Errorf("Failed to remove finalizer for Memcached")
```
camilamacedo86, line 171: Oh, now I think it is clear why we are using requeue(true)
If we change, we will have two texts, which always must be equal with the same info.
So, can we here err = error.NewError(message), log it and use it in the return?

testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
```diff
@@ -117,37 +118,22 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
-
-		if err = r.Update(ctx, memcached); err != nil {
-			log.Error(err, "Failed to update custom resource to add finalizer")
-			return ctrl.Result{}, err
-		}
-	}
-
 	// Check if the Memcached instance is marked to be deleted, which is
 	// indicated by the deletion timestamp being set.
-	isMemcachedMarkedToBeDeleted := memcached.GetDeletionTimestamp() != nil
```
camilamacedo86, line 138: Hi @mateusoliveira43 

IMO: We need to be very surgical about the changes. 
The 

```
ptr.To(true),
RunAsUser:                ptr.To(int64(1001)),
AllowPrivilegeEscalation: ptr.To(false),
```

is a good enhancement

And we might can change the place of the finalizer to create an error and log it as return it
However, I do not see reasons for the others changes.

testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
```diff
@@ -241,23 +250,15 @@ func (r *MemcachedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			log.Error(err, "Failed to update Deployment",
 				"Deployment.Namespace", found.Namespace, "Deployment.Name", found.Name)
 
-			// Re-fetch the memcached Custom Resource before updating the status
-			// so that we have the latest state of the resource on the cluster and we will avoid
-			// raising the error "the object has been modified, please apply
-			// your changes to the latest version and try again" which would re-trigger the reconciliation
-			if err := r.Get(ctx, req.NamespacedName, memcached); err != nil {
-				log.Error(err, "Failed to re-fetch memcached")
-				return ctrl.Result{}, err
```
camilamacedo86, line 250: Before change we need fetch so why we are removing it

testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
mateusoliveira43: this is before logic to delete resource  https://github.com/kubernetes-sigs/kubebuilder/blob/cbc6e383c342f1337ab37ee4aa0755957a01f9c7/testdata/project-v4-with-plugins/internal/controller/memcached_controller.go#L139

I was concerned if this fails https://github.com/kubernetes-sigs/kubebuilder/blob/cbc6e383c342f1337ab37ee4aa0755957a01f9c7/testdata/project-v4-with-plugins/internal/controller/memcached_controller.go#L185

we would add finalizer again and try to delete again. I think this is corner case, so should be ok to let as is

testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
mateusoliveira43: I think it easier to read

it was
```go
	if err != nil && apierrors.IsNotFound(err) {
		// create a new deployment
	} else if err != nil {
		// error
	}
```

my suggestion is
```go
	if err != nil {
		if !apierrors.IsNotFound(err) {
			// error
		}
		// create a new deployment
	}
```

testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
mateusoliveira43: I was thinking on user interactions scenarios 

example: user creates resource -> controller reconciles it -> user changes resource label -> controller reconciles it

even though spec was not changed, without a custom predicate, every update triggers reconcile, right? 

since it was just metadata change, reconcile should do nothing with object

testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
mateusoliveira43: created error to avoid text repetition

testdata/project-v4-with-plugins/internal/controller/memcached_controller.go
mateusoliveira43: update PR

camilamacedo86 (commented):
Hi @mateusoliveira43,

Everything looks great! 🎉

We just need to run make generate to ensure that all samples are updated accordingly.

Also, since this change impacts end users and improves their experience, the appropriate emoji would be :sparkles:. I believe this is more of an enhancement rather than a bug fix.

To help us generate accurate release notes when we make a release, I’ve updated the title accordingly. I hope you don’t mind. 😊

PR #4401, opened 2024-12-01
✨ feat: Automate ENVTEST version from go.mod

camilamacedo86:
Remove the need to manually bump release branches in the Makefile by deriving the ENVTEST version directly from Kubernetes versions specified in go.mod.


docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -189,6 +190,14 @@ controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessar
 $(CONTROLLER_GEN): $(LOCALBIN)
 	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen,$(CONTROLLER_TOOLS_VERSION))
 
+.PHONY: setup-envtest
+setup-envtest: envtest ## Download the binaries required for ENVTEST in the local bin directory.
```
mateusoliveira43, line 194: with this new command, we still need `KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)"` in `make test` command?

camilamacedo86: Yes, because we need to export the values to the ENVVAR and if those are not downloaded yet, they should be.

PR #4404, opened 2024-12-02
🐛 fix: (helm/v1alpha1): Use the ServiceAccount name defined in the values.yaml

wjiec:
The ServiceAccount and RoleBinding currently generated in `dist/chart/rbac` do not reference the `controllerManager.serviceAccountName` defined in values.yaml.


wjiec:
I got my email wrong, I'll create a new one!

Issue #4317, opened 2024-11-10
Implement logic to add checks for conversion webhooks under tests/e2e using kubebuilder:scaffold:e2e-webhooks-checks


camilamacedo86:
### What do you want to happen?

Under the tests/e2e folder have the checks to validate CA injection for conversion webhooks inject via the kubebuilder marker `kubebuilder:scaffold: e2e-webhooks-checks`, similarly to how we check CA injection for mutating and validating webhooks: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/test/e2e/e2e_test.go#L247-L273

See an example of how this check can be done at: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/test/e2e/e2e_test.go#L247-L273.

Also, for reference, see the code used to inject the tests using the marker: https://github.com/kubernetes-sigs/kubebuilder/blob/40db2b8c82d6846df0d4307f130a166ca2a8ce31/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/test.go#L76-L103

PR #4407, opened 2024-12-02
:sparkles: (go/v4): add support to inject conversion webhook for the marker +kubebuilder:scaffold:e2e-webhooks-checks 

mateusoliveira43:
Fix https://github.com/kubernetes-sigs/kubebuilder/issues/4317

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/test.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/test.go
@@ -99,6 +99,18 @@ func (f *WebhookTestUpdater) GetCodeFragments() machinery.CodeFragmentsMap {
 		)
 	}
 
+	if f.Resource.HasConversionWebhook() {
+		conversionWebhookCode := fmt.Sprintf(
+			conversionWebhookChecksFragment,
+			f.Resource.Kind,
+			f.Resource.Plural+"."+f.Resource.Group+"."+f.Resource.Domain,
+		)
+		codeFragments[machinery.NewMarkerFor(f.GetPath(), webhookChecksMarker)] = append(
+			codeFragments[machinery.NewMarkerFor(f.GetPath(), webhookChecksMarker)],
+			conversionWebhookCode,
+		)
+	}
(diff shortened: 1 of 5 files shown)


testdata/project-v4/test/e2e/e2e_test.go
mateusoliveira43, line 1: will create a minikube cluster to test if this is right

mateusoliveira43: had trouble testing, because I am without docker in my current machine. Let me see if CI passes

pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/test.go
```diff
@@ -147,6 +155,22 @@ const validatingWebhookChecksFragment = `It("should have CA injection for valida
 
 `
 
+const conversionWebhookChecksFragment = `It("should have CA injection for conversion webhooks", func() {
```
camilamacedo86, line 158: That is great 🥇 
However, since we can have many APIs and in this way many e2e tests to check the injection of each case scenario, should not be nice add in the It something like `should have C! injection for the {{ .Resource.Kind }} conversion webhook `

camilamacedo86 (commented):
/ok-to-test

camilamacedo86 (commented):
Hi @mateusoliveira43 

It is terrific just a nit. 
Here `It("should have CA injection for conversion webhooks"` I think we should add the kind.

So, ensuring the nit + passing on the tests I think it is great to fly 🚀

pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/test.go
mateusoliveira43: will add this info

mateusoliveira43: done

camilamacedo86 (approved):
/lgtm
/approve
/ok-to-test

PR #4420, opened 2024-12-10
🌱Build the kubebuilder binary before and use it to regenerate the helm chart

monteiro-renato:
I believe this way is better? Specially now that the helm plugin is not yet available in any released version of kubebuilder?
For me, it also makes testing easier as I can make changes to the code and see it reflected in the generated chart without having to remember to `make install` beforehand.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,9 +89,9 @@ generate-docs: ## Update/generate the docs
 	./hack/docs/generate.sh
 
 .PHONY: generate-charts
-generate-charts: ## Re-generate the helm chart testdata only
+generate-charts: build ## Re-generate the helm chart testdata only
 	rm -rf testdata/project-v4-with-plugins/dist/chart
-	(cd testdata/project-v4-with-plugins && kubebuilder edit --plugins=helm/v1-alpha)
+	(cd testdata/project-v4-with-plugins && ../../bin/kubebuilder edit --plugins=helm/v1-alpha)
 
 .PHONY: check-docs
 check-docs: ## Run the script to ensure that the docs are updated


Makefile
```diff
@@ -89,9 +89,9 @@ generate-docs: ## Update/generate the docs
 	./hack/docs/generate.sh
 
 .PHONY: generate-charts
-generate-charts: ## Re-generate the helm chart testdata only
+generate-charts: build ## Re-generate the helm chart testdata only
```
camilamacedo86, line 92: So we need to call here make install instead right?

monteiro-renato: Personally, I never actually use `make install`, I find it too intrusive.
I prefer to use the binary in the bin folder.

camilamacedo86: Ok. So, based on your POV, we might need to leave it as it is.
We recommend using Make Install in all places. 

However, it has contributors do not like it, they will prefer as you to build
So, leaving it as it is allows people to have a choice

WDYT?

monteiro-renato: In my opinion, it makes sense to use the kubebuilder version built from main since this is regenerating data in the ./testdata directory.
The `make generate` command, and more specifically, the `make generate-testdata` command, calls the script in https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh#L123C1-L123C9 which is also building kubebuilder (https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/common.sh#L108-L113) and putting it in a tmp directory instead of relying on the locally installed version.

But feel free to close the PR if you thing it's best to leave it as is.

camilamacedo86: Your argumentation is fair enough 👍

camilamacedo86 (approved):
/ok-to-test
/approve
/lgtm

Issue #4424, opened 2024-12-13
 Improve Ginkgo/Gomega Test Practices in Scaffolded and e2e Tests


camilamacedo86:
### What do you want to happen?
Kubebuilder creates/scaffolds a lot of test code. This test code uses Ginkgo and Gomega. Some of the tests as written don't take full advantage of Gomega's testing framework. Since Kubebuilder is likely to introduce many developers to Gomega, it is important that the scaffolded code provides the best possible examples.

We should see that the tests scaffolded for end-users have this issue fixed already: [testdata/project-v4/test](https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v4/test).

What we need to do here is ensure that **ALL our e2e tests** are also following the same patterns as the end-user scaffolds. Therefore, the necessary changes must be implemented in: [test/e2e](https://github.com/kubernetes-sigs/kubebuilder/tree/master/test/e2e).

### Context

The context for this issue and similar discussions can be found here: [kubernetes-sigs/kubebuilder#4135](https://github.com/kubernetes-sigs/kubebuilder/issues/4135).

### Tasks

1. Review and compare the test scaffolds in `testdata/project-v4/test` to ensure best practices in Ginkgo/Gomega are being followed.
2. Update all e2e tests in `test/e2e` to align with these patterns and provide the best examples for developers.
3. Validate that the tests maintain readability while leveraging Gomega's framework effectively.

### Extra Labels

/kind cleanup

mogsie:
In #4135, I originally made a suggestion of replacing the following type of tests:

```
result, err := someFunc(param, param)
Expect(err).NotTo(HaveOccurred())
Expect(result).To(ContainSubstring("blah"))
```

with the more concise (but functionally equivalent)

```
Expect(someFunc(param, param)).To(ContainSubstring("blah"))
```

The one line above
* checks that err is nil
* and that the resullt of the function call matches the matcher.

Should we reconsider this?  The one line is (to me) more maintainable, although it relies Gomega implicitly testing that `err == nil` of the last return value.

From #4135:

> ... this has been decided against, as it hides the number of return values, which makes it harder to learn the API being tested. The desire is therefore to keep the three-line version, instead of in-lining it.

[...]

PR #4426, opened 2024-12-14
🌱improve Ginkgo/Gomega test style

Sijoma:
Updates a couple of e2e tests to use Gomega style mentioned in https://github.com/kubernetes-sigs/kubebuilder/issues/4135 . 

I'm not sure if I 100% understood the issue so before I try to rewrite all of them I thought it would be better to create a smaller PR to get feedback earlier. 

relates #4424

### DIFF
--- a/test/e2e/v4/plugin_cluster_test.go
+++ b/test/e2e/v4/plugin_cluster_test.go
@@ -112,43 +112,43 @@ func Run(kbc *utils.TestContext, hasWebhook, isToUseInstaller, isToUseHelmChart,
 
 	By("creating manager namespace")
 	err = kbc.CreateManagerNamespace()
-	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+	Expect(err).NotTo(HaveOccurred())
 
 	By("labeling the namespace to enforce the restricted security policy")
 	err = kbc.LabelNamespacesToEnforceRestricted()
-	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+	Expect(err).NotTo(HaveOccurred())
 
 	By("updating the go.mod")
 	err = kbc.Tidy()
-	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+	Expect(err).NotTo(HaveOccurred())
 
 	By("run make all")
 	err = kbc.Make("all")
-	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+	Expect(err).NotTo(HaveOccurred())
 
 	By("building the controller image")
 	err = kbc.Make("docker-build", "IMG="+kbc.ImageName)
-	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+	Expect(err).NotTo(HaveOccurred())
 
 	By("loading the controller docker image into the kind cluster")
 	err = kbc.LoadImageToKindCluster()
-	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+	Expect(err).NotTo(HaveOccurred())
 
 	if !isToUseInstaller && !isToUseHelmChart {
 		By("deploying the controller-manager")
 		cmd := exec.Command("make", "deploy", "IMG="+kbc.ImageName)
 		_, err = kbc.Run(cmd)
-		ExpectWithOffset(1, err).NotTo(HaveOccurred())
+		Expect(err).NotTo(HaveOccurred())
 	}
 
 	if isToUseInstaller && !isToUseHelmChart {
 		By("building the installer")
 		err = kbc.Make("build-installer", "IMG="+kbc.ImageName)
-		ExpectWithOffset(1, err).NotTo(HaveOccurred())
+		Expect(err).NotTo(HaveOccurred())
 
 		By("deploying the controller-manager with the installer")
 		_, err = kbc.Kubectl.Apply(true, "-f", "dist/install.yaml")
-		ExpectWithOffset(1, err).NotTo(HaveOccurred())
+		Expect(err).NotTo(HaveOccurred())
 	}
 
 	if isToUseHelmChart && !isToUseInstaller {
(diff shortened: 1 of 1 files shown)


test/e2e/v4/plugin_cluster_test.go
```diff
@@ -254,54 +254,53 @@ func Run(kbc *utils.TestContext, hasWebhook, isToUseInstaller, isToUseHelmChart,
...
 			mwhOutput, err := kbc.Kubectl.Get(
 				false,
 				"mutatingwebhookconfigurations.admissionregistration.k8s.io",
 				fmt.Sprintf("e2e-%s-mutating-webhook-configuration", kbc.TestSuffix),
 				"-o", "go-template={{ range .webhooks }}{{ .clientConfig.caBundle }}{{ end }}")
-			ExpectWithOffset(2, err).NotTo(HaveOccurred())
+			g.ExpectWithOffset(2, err).NotTo(HaveOccurred())
 			// check that ca should be long enough, because there may be a place holder "\n"
-			ExpectWithOffset(2, len(mwhOutput)).To(BeNumerically(">", 10))
+			g.ExpectWithOffset(2, len(mwhOutput)).To(BeNumerically(">", 10))
```
camilamacedo86, line 274: The idea is to follow the same approach as the test scaffolds.  
If you look there, you will see that it is quite similar to what we do here.  

For example, we should no longer use `ExpectWithOffset`; instead, we should use the functions and simplify it.

Look at https://github.com/kubernetes-sigs/kubebuilder/issues/4135  

And check the PRs done by @mogsie related to the tests.

Sijoma:
@camilamacedo86 does this go in the right direction? 

Some test looked like they could be refactored to a table driven test, the ones with the `Run` function. I've kept it like this for now.

It's a bit confusing since https://github.com/kubernetes-sigs/kubebuilder/issues/4135 still contains examples that contain `EventuallyWithOffset`. I've looked at https://github.com/kubernetes-sigs/kubebuilder/pull/4137/files and it looks a bit similar. Although `Eventually` still uses the global and not `g.Eventually` (Maybe this is actually incorrect from my side). 

Happy to get an example if some of these changes are not correct how to fix it.

test/e2e/v4/plugin_cluster_test.go
```diff
@@ -112,43 +112,43 @@ func Run(kbc *utils.TestContext, hasWebhook, isToUseInstaller, isToUseHelmChart,
 
 	By("creating manager namespace")
```
camilamacedo86, line 113: Hi @Sijoma 

Please, ensure that you have your changes released with master
Also, please squash the commits; We should not have more than 1 commit to be able to get it merged

Thank you 🥇

test/e2e/v4/plugin_cluster_test.go
```diff
@@ -185,25 +185,25 @@ func Run(kbc *utils.TestContext, hasWebhook, isToUseInstaller, isToUseHelmChart,
...
+	Expect(err).NotTo(HaveOccurred())
+	Expect(podOutput).To(ContainSubstring("leader-elect"),
 		"Expected manager pod to have --leader-elect flag")
-	ExpectWithOffset(1, podOutput).To(ContainSubstring("health-probe-bind-address"),
+	Expect(podOutput).To(ContainSubstring("health-probe-bind-address"),
 		"Expected manager pod to have --health-probe-bind-address flag")
 
 	By("validating that the Prometheus manager has provisioned the Service")
-	EventuallyWithOffset(1, func() error {
+	Eventually(func(g Gomega) {
```
mateusoliveira43, line 195: I think there is no need to change function inside Eventually. The `.Should(Succeed())` will do the assertion for us

Sijoma: I found it a bit simpler to not have return values inside these functions. It's a bit simpler to reason as there is nothing returned.  

That way we can already immediately see that the error should never occur. Whereas when we return it it could be still checked for `NotFound` etc.

test/e2e/v4/plugin_cluster_test.go
Sijoma: Squashed & Rebased it now. 

Thanks!

Sijoma:
Not really sure what went wrong here:

I see the following logs:
> ✗ Waiting ≤ 1m0s for control-plane = Ready ⏳
 • WARNING: Timed out waiting for Ready ⚠️

>  Events:
    Type     Reason       Age   From               Message
    ----     ------       ----  ----               -------
    Normal   Scheduled    5s    default-scheduler  Successfully assigned e2e-vlfy-system/e2e-vlfy-controller-manager-6c94678688-6k4dd to kind-control-plane
    Warning  FailedMount  4s    kubelet            MountVolume.SetUp failed for volume "kube-api-access-fkgst" : failed to sync configmap cache: timed out waiting for the condition

>Summarizing 6 Failures:
  [FAIL] kubebuilder plugin go/v4 [It] should generate a runnable project
  /home/prow/go/src/sigs.k8s.io/kubebuilder/test/e2e/v4/plugin_cluster_test.go:336

Sijoma:
Seeing it now:

>  [FAILED] Unexpected non-nil/non-zero argument at index 1:
  	<*utils.TestContext>: &utils.TestContext{CmdContext:(*utils.CmdContext)(0xc0004767c0), TestSuffix:"tdqg", Domain:"example.comtdqg", Group:"bartdqg", Version:"v1alpha1", Kind:"Footdqg", Resources:"footdqgs", ImageName:"e2e-test/controller-manager:tdqg", BinaryName:"kubebuilder", Kubectl:(*utils.Kubectl)(0xc00014cc90), K8sVersion:(*utils.KubernetesVersion)(0xc000136600), IsRestricted:false}
  In [It] at: /home/prow/go/src/sigs.k8s.io/kubebuilder/test/e2e/v4/plugin_cluster_test.go:336 @ 12/25/24 17:53:14.473

Sijoma:
Also realised now that `make test` isn't even running these tests 😄. Using `make test-e2e-local` now.

test/e2e/v4/plugin_cluster_test.go
mogsie: Using a func(Gomega) in an Eventually is more "scalable" in that if you want to do more things inside the function, you can keep to the Gomega semantics with assertions.

* The error messages tell you exactly what's wrong.
* The line number of the error is correct (the error points you to the line inside the func() not just the line of the Eventually() call).
* No need to switch between two different ways of checking for errors in tests.

So my vote is to keep this change.

test/e2e/v4/plugin_cluster_test.go
```diff
@@ -219,9 +219,8 @@ func Run(kbc *utils.TestContext, hasWebhook, isToUseInstaller, isToUseHelmChart,
 
 		if hasMetrics {
 			By("labeling the namespace to allow consume the metrics")
-			_, err = kbc.Kubectl.Command("label", "namespaces", kbc.Kubectl.Namespace,
-				"metrics=enabled")
-			ExpectWithOffset(2, err).NotTo(HaveOccurred())
+			Expect(kbc.Kubectl.Command("label", "namespaces", kbc.Kubectl.Namespace,
```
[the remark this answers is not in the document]
Sijoma: I started to work through it a bit - and initially changed a couple of more things (also incorrect ones). So I didn't want to move everything in one go. I then focused really on all the "...WithOffset" calls and the functions passed to Eventually(...)

Happy to add also change it in more places. I agree that it's better. Maybe just as a follow up pr? Or is there a preference to have it more in one PR?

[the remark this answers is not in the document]
mogsie: It's probably ok to do it in several PRs.  I'd keep it focused and to-the-point, so that PRs don't need to live too long and need constant rebasing etc.  The PR you have already fixes a lot of actual bugs (i.e. the Eventually() calls that use the global assertions).

I remember [a few months back, when I fixed assertion code like this](https://github.com/kubernetes-sigs/kubebuilder/pull/4197/commits/618a87b1e2e9b6b99ff6d2893fd9ed9207cd7b56#diff-efb8aa989aa88823884ccfe4e45f2dde83ee931d9d0b4cec73e5d42c105fb82aL133), I discovered that the test itself was [testing the completely wrong thing](https://github.com/mogsie/kubebuilder/blob/3352876ff7466ac2c5ffc636a9bb9bc2808b3d5c/testdata/project-v4-multigroup/internal/controller/example.com/busybox_controller_test.go#L132), and the incorrect assertions just happened to allow the test to pass.

So, when you do fix things like this, you might end up with tests that don't pass, and it might just be that the test itself was not testing what the original authors thought it was testing.

mogsie (approved):

camilamacedo86 (approved):
/lgtm
/ok-to-test
/approve 

Well DONE 🥇

PR #4430, opened 2024-12-15
🐛 (helm/v1alpha1): using a more precise pod selector in NetworkPolicy

wjiec:
PR's key content:

1. The `podSelector` in `NetworkPolicy` should use more precise labels.
2. Add default values for `controllerManager.pod.labels` in values.yaml, this helps the user to visualize what options are available (there are quite a few more of these issues in this plugin, do they need to be changed together now?)

### DIFF
--- a/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
+++ b/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
@@ -359,6 +359,17 @@ func copyFileWithHelmLogic(srcFile, destFile, subDir, projectName string) error 
 		contentStr = injectAnnotations(contentStr, hasWebhookPatch)
 	}
 
+	// Apply NetworkPolicy-specific replacements
+	if subDir == "networkPolicy" {
+		contentStr = strings.Replace(contentStr,
+			"matchLabels:",
+			`matchLabels:
+      {{- include "chart.selectorLabels" . | nindent 6 }}
+      {{- with .Values.controllerManager.pod.labels }}
+      {{- toYaml . | nindent 6 }}
+      {{- end }}`, 1)
+	}
+
 	// Remove existing labels if necessary
 	contentStr = removeLabels(contentStr)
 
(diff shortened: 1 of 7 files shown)


pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
```diff
@@ -359,6 +359,17 @@ func copyFileWithHelmLogic(srcFile, destFile, subDir, projectName string) error
 		contentStr = injectAnnotations(contentStr, hasWebhookPatch)
 	}
 
+	// Apply NetworkPolicy-specific replacements
```
camilamacedo86, line 362: @wjiec 

HI note that the goal of HelmChart is ONLY to package/distribute the solution in this format.
So, how is it done in the default scaffold? We need to ensure that both are doing things the same way, and we cannot address customizations and deviations that are valid only for the HelmCharts. 

In this way, can you please check the default scaffold and let us know:
- Do you think that must be changed? If yes, why and how?
- If not, why would this change only be applied to the HelmCharts?
- Also, what is the problem that this change solves? What is not possible to do with it as it is that will be possible to do after those changes?

wjiec: Hi @camilamacedo86 , I think this feature may not be necessary, but a more precise selector I think would be helpful. In the scenario I'm working in, we need to install multiple controllers into the same namespace for historical reasons, and the original podSelector in NetworkPolicy was too broad and would match controllers we didn't want to open.

With that in mind, I think it's possible to synchronize the default scaffold changes. Since we're currently focused on wanting to use the official solution for helm-based deployments, I'm ignoring the default scaffold, sorry 😅.

camilamacedo86: If you want propose changes then, we need to do that in the default scaffold
Ofen, it would have already end up in the helm charts
We will do that if we still need some week in the Helm.

wjiec: @camilamacedo86 Do you think we should extend this change to the default scaffolding?

wjiec: Yes, I think it's necessary. I'd be more than happy to add this to the default scaffolding :)

wjiec:
After the #4437 merge, the `podSelector` that the helm plugin copied from the default scaffolding I think is sufficiently used, so this PR is not needed, thanks.

PR #4425, opened 2024-12-13
✨ (go/v4): Add new makefile target to check and validate the linter config

mateusoliveira43:
Add check for linter configuration. 

Fix https://github.com/kubernetes-sigs/kubebuilder/issues/4416

PR #4432, opened 2024-12-17
🌱 : Ensure that golangci-lint CI version is passed for the templates

mateusoliveira43:
Noted while working on https://github.com/kubernetes-sigs/kubebuilder/pull/4425 that scaffold GitHub action and Makefile have different `golangci-lint` version

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/.github/workflows/lint.yml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/.github/workflows/lint.yml
@@ -20,4 +20,4 @@ jobs:
       - name: Run linter
         uses: golangci/golangci-lint-action@v6
         with:
-          version: v1.61
+          version: v1.61.0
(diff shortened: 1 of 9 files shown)


mateusoliveira43:
another thing that I like to add is a link to doc in `golangci-lint` config file. Example https://github.com/migtools/oadp-non-admin/blob/master/.golangci.yml#L1

docs/book/src/cronjob-tutorial/testdata/project/.github/workflows/lint.yml
```diff
@@ -20,4 +20,4 @@ jobs:
       - name: Run linter
         uses: golangci/golangci-lint-action@v6
         with:
-          version: v1.61
+          version: v1.61.0
```
mateusoliveira43, line 23: To simplify updates, we could get version from Makefile instead. What you think?

camilamacedo86 (approved):
/lgtm
/approved
/ok-to-test

PR #4435, opened 2024-12-19
:sparkles: feat: enforce restricted Pod Security Context Compliance in testing

### LINKED ISSUE #4423: Ensure Security Context Compliance in Tests and User-Scaffolded Tests for Restricted Environments
## Background
To ensure compliance with restricted environments, all Pods in our tests and test scaffolds must use a **Security Context**. This ensures that even when namespaces enforce strict security policies, the tests will pass, and the scaffolds users rely on will remain functional.

Currently:
- The `manager` is configured by default to work in such environments.
- However, in tests and test scaffolds, the `curl` Pod (used to verify metrics logs) lacks the necessary `SecurityContext` configuration.

### 1. Namespace Labeling for Security Policy Enforcement
- Update the e2e test namespace to be labeled for restricted environments. See that in our e2e tests we are just warning : https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v4/plugin_cluster_test.go#L60-L62 we need change it to enforce can also use the same in the scaffolds produced for end-users
- Modify [`testdata/project-v4/test/e2e/e2e_test.go#L51-L55`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/test/e2e/e2e_test.go#L51-L55) to scaffold the command that applies the required labels to the namespace.

### 2. Secure `curl` Pod Configuration
- Change the `curl` Pod creation command in [`testdata/project-v4/test/e2e/e2e_test.go#L209-L217`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/test/e2e/e2e_test.go#L209-L217) to include a proper `SecurityContext`. For example:

  ```go
  cmd = exec.Command("kubectl", "run", curlPod,
      "--image=curlimages/curl:7.87.0", "-n", namespace,
      "--restart=Never",
      "--overrides", `{
          "spec": {
              "containers": [{
                  "name": "curl",
                  "image": "curlimages/curl:7.87.0",
                  "command": ["sh", "-c", "sleep 3600"],
                  "securityContext": {
                      "allowPrivilegeEscalation": false,
                      "capabilities": {
                          "drop": ["ALL"]
                      },
                      "runAsNonRoot": true,
                      "runAsUser": 1000,
                      "seccompProfile": {
                          "type": "RuntimeDefault"
                      }
                  }
              }],
              "serviceAccountName": "<project>-controller-manager"
          }
      }`)

### 3. Update e2e Test Framework
- Modify e2e tests in [`test/e2e`](https://github.com/kubernetes-sigs/kubebuilder/tree/master/test/e2e) to ensure that all created Pods include a `SecurityContext` and are compatible with namespaces enforcing restricted security policies.

## Expected Outcome
- All e2e tests will pass in restricted environments with strict security contexts enforced.
- User-scaffolded tests will include the necessary configurations to operate in restricted environments without requiring manual adjustments.

## Additional Notes
- To change the scaffolds, we change the sample. In this case, we need to change: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/test.go
- Then, we run `make generate` to ensure that all samples under docs and testdata are properly updated.
- If you need to check what is Pod Security Standards see: https://kubernetes.io/docs/concepts/security/pod-security-standards/


### Extra Labels

_No response_

lunarwhite:
## Description

Closes #4423 

Main changes:

- Updated the test scaffolds to enforce the restricted security policy:
  - Labeled the manager namespace with `pod-security.kubernetes.io/enforce=restricted`.
  - Changed the curl Pod creation command to adhere the `securityContext` Pod spec fields
- Updated the tests under test/e2e:
  - Changed the utils func `LabelNamespacesToWarnAboutRestricted()`, `RemoveNamespaceLabelToWarnAboutRestricted()` to `LabelNamespacesToEnforceRestricted()`, `RemoveNamespaceLabelToEnforceRestricted()`.
  - Made adjustments to ensure compatibility with the strict security policy enforced.
- Changed to use `curlimages/curl:latest` instead of pinning a version.

### DIFF
--- a/test/e2e/deployimage/plugin_cluster_test.go
+++ b/test/e2e/deployimage/plugin_cluster_test.go
@@ -93,7 +93,8 @@ func Run(kbc *utils.TestContext) {
 
 	By("deploying the controller-manager")
 	cmd := exec.Command("make", "deploy", "IMG="+kbc.ImageName)
-	Expect(kbc.Run(cmd)).NotTo(ContainSubstring("Warning: would violate PodSecurity"))
+	out, _ := kbc.Run(cmd)
+	Expect(string(out)).NotTo(ContainSubstring("Warning: would violate PodSecurity"))
 
 	By("validating that the controller-manager pod is running as expected")
 	verifyControllerUp := func(g Gomega) {
--- a/testdata/project-v4-multigroup/test/e2e/e2e_test.go
+++ b/testdata/project-v4-multigroup/test/e2e/e2e_test.go
@@ -46,12 +46,19 @@ var _ = Describe("Manager", Ordered, func() {
 	var controllerPodName string
 
 	// Before running the tests, set up the environment by creating the namespace,
-	// installing CRDs, and deploying the controller.
+	// enforce the restricted security policy to the namespace, installing CRDs,
+	// and deploying the controller.
 	BeforeAll(func() {
 		By("creating manager namespace")
 		cmd := exec.Command("kubectl", "create", "ns", namespace)
 		_, err := utils.Run(cmd)
 		Expect(err).NotTo(HaveOccurred(), "Failed to create namespace")
+
+		By("labeling the namespace to enforce the restricted security policy")
+		cmd = exec.Command("kubectl", "label", "--overwrite", "ns", namespace,
+			"pod-security.kubernetes.io/enforce=restricted")
+		_, err = utils.Run(cmd)
+		Expect(err).NotTo(HaveOccurred(), "Failed to label namespace with restricted policy")
 
 		By("installing CRDs")
 		cmd = exec.Command("make", "install")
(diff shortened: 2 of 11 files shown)


testdata/project-v4-multigroup/test/e2e/e2e_test.go
```diff
@@ -210,9 +217,29 @@ var _ = Describe("Manager", Ordered, func() {
...
 				"--image=curlimages/curl:7.78.0",
-				"--", "/bin/sh", "-c", fmt.Sprintf(
-					"curl -v -k -H 'Authorization: Bearer %s' https://%s.%s.svc.cluster.local:8443/metrics",
-					token, metricsServiceName, namespace))
+				"--overrides",
+				fmt.Sprintf(`{
+					"spec": {
+						"containers": [{
+							"name": "curl",
+							"image": "curlimages/curl:7.87.0",
```
camilamacedo86, line 225: Instead of pinning a version, do you think we could use the latest to make it easier the code be maintained and we do not need to bump new versions

WDYT?

camilamacedo86:
In this case it is failing in the API diff because we are changing it at

> sigs.k8s.io/kubebuilder/v4/test/e2e/utils
  Incompatible changes:
  - (*TestContext).LabelNamespacesToWarnAboutRestricted: removed
  - (*TestContext).RemoveNamespaceLabelToWarnAboutRestricted: removed
  
But we do not really need to support its utils
That is for tests only, mainly us. we might want to make it internal in the future
Anyway, it seems fine. No worry about this failure.

camilamacedo86:
We need to have 

>	out, err := kbc.Run(cmd)
++	Expect(string(output)).NotTo(ContainSubstring("Warning: would violate PodSecurity"))

To get the output in the logs.

testdata/project-v4-multigroup/test/e2e/e2e_test.go
lunarwhite: Addressed. Ready for another review.

camilamacedo86 (commented):
/ok-to-test

test/e2e/deployimage/plugin_cluster_test.go
```diff
@@ -93,7 +93,8 @@ func Run(kbc *utils.TestContext) {
 
 	By("deploying the controller-manager")
 	cmd := exec.Command("make", "deploy", "IMG="+kbc.ImageName)
-	Expect(kbc.Run(cmd)).NotTo(ContainSubstring("Warning: would violate PodSecurity"))
+	out, _ := kbc.Run(cmd)
```
mogsie, line 96: This change ignores the error here. is that intentional?

Since it goes against idioms, maybe it should be commented?

camilamacedo86: Hi @mogsie 

IHMO: We need to have like

```
_, err = utils.Run(cmd)
		Expect(err).NotTo(HaveOccurred(), "Failed to label namespace with restricted policy")
```

Instead of `Expect(utils.Run(cmd)),` which does not make the result/output clear, we risk having an error in the string format outputted when needed.

mogsie: Reported it as #4450

PR #4438, opened 2024-12-24
:seedling: add better description to MarshalYAML and UnmarshalYAML methods

fengshunli:

### DIFF
--- a/pkg/config/interface.go
+++ b/pkg/config/interface.go
@@ -95,8 +95,8 @@ type Config interface {
 
 	/* Persistence */
 
-	// Marshal returns the YAML representation of the Config.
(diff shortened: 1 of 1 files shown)


pkg/config/interface.go
mateusoliveira43, line 1: A suggestion is to enable linter to avoid this in the future

The rule for checking this is currently disabled
https://github.com/kubernetes-sigs/kubebuilder/blob/49ad403e59c7dfb46089a3a6bb99604ca357260d/.golangci.yml#L37-L38

We could check this by adding this param and updating golangci-lint version 
```yaml
      - name: exported
        arguments:
          - "checkPublicInterface"
```

camilamacedo86 (approved):
/lgtm
/approve
/ok-to-test

PR #4446, opened 2024-12-27
✨ Upgrade scaffolds to use go 1.23

camilamacedo86:
Upgrade scaffolds from go version 1.22 to 1.23

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: '1.22'
+          go-version: '1.23'
 
       - name: Build and install Kubebuilder CLI
         run: make install
--- a/testdata/project-v4/go.mod
+++ b/testdata/project-v4/go.mod
@@ -1,6 +1,8 @@
 module sigs.k8s.io/kubebuilder/testdata/project-v4
 
-go 1.22.0
+go 1.23.0
+
+godebug default=go1.23
 
 require (
 	github.com/cert-manager/cert-manager v1.16.2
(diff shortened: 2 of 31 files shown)


.github/workflows/codeql.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: '1.22'
+          go-version: '1.23'
```
mateusoliveira43, line 28: would be better to use `go-version-file: go.mod`?

camilamacedo86: Would you like to push a PR with this change?
Yes, that it would be better if work out in this case

mateusoliveira43: yes, can work on that next week

mateusoliveira43 (approved):

camilamacedo86:
/skip pull-kubebuilder-e2e-k8s-1-32-0

not working yet

testdata/project-v4/go.mod
```diff
@@ -1,6 +1,8 @@
 module sigs.k8s.io/kubebuilder/testdata/project-v4
 
-go 1.22.0
+go 1.23.0
+
+godebug default=go1.23
```
ldemailly, line 5: curious what does that do when you also have go 1.23.0 ?

camilamacedo86: Why would we need to set the patch for debug?

ldemailly: I should have clarified: I am asking why you added these `godebug` lines with the default that happens to exactly match the go version in the go line; my understanding is that is a "noop" but I'd love to learn why it isn't

(I am not talking about patchlevel at all)

ldemailly: opened https://github.com/kubernetes-sigs/kubebuilder/issues/4786 - ty

PR #4449, opened 2024-12-27
:bug: (go/v4): Align webhook suite test with controller's suite test

mateusoliveira43:
Enable conversion webhooks for webhook `suite_test.go` `envtest` by using `k8s.io/client-go/kubernetes/scheme` instead of creating new scheme (this also already done in  controller `suite_test.go`)

- Updated the webhook suite test to follow the structure and conventions of the controller's suite test. 
- Switched to using `k8s.io/client-go/kubernetes/scheme` instead of `k8s.io/apimachinery/pkg/runtime`. 
- This ensures consistency across test suites and improves maintainability.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/internal/webhook/v1/webhook_suite_test.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/internal/webhook/v1/webhook_suite_test.go
@@ -29,8 +29,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
-	admissionv1 "k8s.io/api/admission/v1"
-	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -67,11 +66,7 @@ var _ = BeforeSuite(func() {
 	ctx, cancel = context.WithCancel(context.TODO())
 
 	var err error
-	scheme := apimachineryruntime.NewScheme()
-	err = batchv1.AddToScheme(scheme)
-	Expect(err).NotTo(HaveOccurred())
-
-	err = admissionv1.AddToScheme(scheme)
+	err = batchv1.AddToScheme(scheme.Scheme)
 	Expect(err).NotTo(HaveOccurred())
 
 	// +kubebuilder:scaffold:scheme
@@ -96,14 +91,14 @@ var _ = BeforeSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 	Expect(cfg).NotTo(BeNil())
 
-	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme})
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
 	Expect(err).NotTo(HaveOccurred())
 	Expect(k8sClient).NotTo(BeNil())
 
 	// start webhook server using Manager.
 	webhookInstallOptions := &testEnv.WebhookInstallOptions
 	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
-		Scheme: scheme,
+		Scheme: scheme.Scheme,
 		WebhookServer: webhook.NewServer(webhook.Options{
 			Host:    webhookInstallOptions.LocalServingHost,
 			Port:    webhookInstallOptions.LocalServingPort,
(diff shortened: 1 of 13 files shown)


camilamacedo86 (commented):
Hi @mateusoliveira43,

This one will require a safer approach and a deeper analysis to ensure that these changes are truly necessary and/or the best approach for the default scaffolds.

Currently, we have `envtests` implemented and everything is functioning correctly, as shown here:  
[cronjob_webhook_test.go](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/project/internal/webhook/v1/cronjob_webhook_test.go)

I don’t believe that webhook tests are heavily used by users at the moment, but that doesn’t seem to be the case for controllers. Additionally, we already have mocks with tests that are functioning correctly for controllers.

Also, we’ve never added the `CRDInstallOptions` option—it doesn’t seem to be required.

Could we raise an issue explaining why this change is necessary?  
Could you provide more detailed information in the description here?

It would be helpful if you could let us know:

- What you tried to do and what didn’t work as expected.  
- What challenges you faced and what you expected to see instead that led you to make each change.  
  _(For example, moving the `AddSchemes`, removing the import, and adding the optional spec for `envtest`.)_

/hold until we have those details and ensure that each change is required/and or is the best approach for the default scaffolds. 

Btw, thanks a lot for looking on that.

mateusoliveira43:
@camilamacedo86 will undo change to `CRDInstallOptions`, the default value is what I am adding

    // Scheme is used to determine if conversion webhooks should be enabled
    // for a particular CRD / object.
    //
    // Conversion webhooks are going to be enabled if an object in the scheme
    // implements Hub and Spoke conversions.
    //
    // If nil, scheme.Scheme is used.

Will add more info to PR description

docs/book/src/cronjob-tutorial/testdata/project/internal/controller/suite_test.go
```diff
@@ -92,25 +106,11 @@ var _ = BeforeSuite(func() {
 		Then, we start the envtest cluster.
 	*/
 
-	var err error
 	// cfg is defined in this file globally.
 	cfg, err = testEnv.Start()
 	Expect(err).NotTo(HaveOccurred())
 	Expect(cfg).NotTo(BeNil())
 
-	/*
```
camilamacedo86, line 101: Hi @mateusoliveira43 

I think we can reduce the scope of this one
We can have a PR to add the schemas before start the envtest

Then, we can have a PR that propose we remove apimachineryruntime and use  scheme.Scheme and explain why we want to do that

mateusoliveira43: done https://github.com/kubernetes-sigs/kubebuilder/pull/4466

will open next one after that one merges

camilamacedo86:
HI @mateusoliveira43 

Can you rebase it with the master and ensure that you have what was changed and why in the description?

mateusoliveira43:
@camilamacedo86 updated

docs/book/src/cronjob-tutorial/testdata/project/internal/webhook/v1/webhook_suite_test.go
```diff
@@ -29,8 +29,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
-	admissionv1 "k8s.io/api/admission/v1"
-	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
```
camilamacedo86, line 32: HI @mateusoliveira43 

Thanks for suggesting this! I’ve been thinking about it, and I believe we should stick to the current setup for a few reasons:

It gives users granular control over adding only the API groups they need for their test cases, keeping things lightweight and precise. Not that it is not a bug. It is done in this one by design. Users can add ANY required scheme. 
- Using `client-go/kubernetes/scheme` adds all API groups by default, which increases memory usage and makes debugging schema issues harder.
**- The current approach is more streamlined for custom resources, avoiding unnecessary complexity or conflicts.**


Overall, I think the flexibility of the existing setup is better suited for varied test scenarios. 
Let me know your thoughts!

Best,

camilamacedo86: OH, I think now I understand. 
The change proposed in https://github.com/kubernetes-sigs/kubebuilder/pull/4449 is to make it be as we have it setup for controllers now. 

Would be that @mateusoliveira43 ?

mateusoliveira43: yes, both for consistency (same setup as controllers test) and for fixing the bug with conversion webhooks

PR #4463, opened 2025-01-01
🌱 Enable the lint to check exported methods and fix issues

camilamacedo86:
Enable the lint to check exported methods and fix issues


camilamacedo86:
Verify API differences are reporting a false alarm since the exported changes are under internal packages.

PR #4465, opened 2025-01-01
:seedling: Enable govet settings and fix issue found

mateusoliveira43:
Fix linter error `nilness: impossible condition: nil != nil (govet)`

```
golangci-lint run
pkg/plugins/golang/v4/scaffolds/edit.go:60:9: nilness: impossible condition: nil != nil (govet)
	if err != nil && s.multigroup != s.config.IsMultiGroup() {
	       ^
make: *** [lint] Error 1
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -17,6 +17,11 @@ issues:
         - gosec
 
 linters-settings:
+  govet:
+    enable-all: true
+    disable:
+      - fieldalignment
+      - shadow
   revive:
     rules:
       # The following rules are recommended https://github.com/mgechev/revive#recommended-configuration
--- a/pkg/plugins/golang/v4/scaffolds/edit.go
+++ b/pkg/plugins/golang/v4/scaffolds/edit.go
@@ -56,11 +56,6 @@ func (s *editScaffolder) Scaffold() error {
 	}
 	str := string(bs)
 
-	// Ignore the error encountered, if the file is already in desired format.
-	if err != nil && s.multigroup != s.config.IsMultiGroup() {
-		return err
-	}
-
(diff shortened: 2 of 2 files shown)


.golangci.yml
```diff
@@ -17,6 +17,11 @@ issues:
         - gosec
 
 linters-settings:
+  govet:
+    enable-all: true
+    disable:
+      - fieldalignment
+      - shadow
```
camilamacedo86, line 24: Why do we need to add it?
The code is passing in the tests
What error are you facing that motivates you to add it?

pkg/plugins/golang/v4/scaffolds/edit.go
```diff
@@ -56,11 +56,6 @@ func (s *editScaffolder) Scaffold() error {
 	}
 	str := string(bs)
 
-	// Ignore the error encountered, if the file is already in desired format.
-	if err != nil && s.multigroup != s.config.IsMultiGroup() {
-		return err
-	}
```
camilamacedo86, line 62: Why are you removing this condition?

.golangci.yml
mateusoliveira43: these 3 govet rules, fieldalignment, shadow and nilness, are not verified in the code. The other rules are

this PR fixes nilness rule and enforces it

pkg/plugins/golang/v4/scaffolds/edit.go
mateusoliveira43: govet found this issue `nilness: impossible condition: nil != nil (govet)`

this line of code is never triggered, so I removed it

PR #4468, opened 2025-01-06
:seedling: fix: remove unused nolint comments

mateusoliveira43:
There are unnecessary `//nolint` comments in project code

This PR removes these unnecessary comments and enforces that no new `//nolint` unnecessary comments are added to the code

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -22,6 +22,8 @@ linters-settings:
     disable:
       - fieldalignment
       - shadow
+  nolintlint:
+    allow-unused: false
   revive:
     rules:
       # The following rules are recommended https://github.com/mgechev/revive#recommended-configuration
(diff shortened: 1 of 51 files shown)


.golangci.yml
```diff
@@ -22,6 +22,8 @@ linters-settings:
     disable:
       - fieldalignment
       - shadow
+  nolintlint:
+    allow-unused: false
```
camilamacedo86, line 26: Why do we need this one?
What it does?

mateusoliveira43: it checks for unnecessary `//nolint` comments

for example a line was to long, so we added `//nolint:lll` to it

in another moment, the line was changed, and it is now not too long, so `//nolint` comment is not necessary anymore

with this config, that line would be highlighted to remove the now not needed `//nolint` comment

Scutua (approved):

Issue #4478, opened 2025-01-09
Rescaffolding docs incorrect note about parameters


Sijoma:
### What broke? What's expected?

The documentation for [rescaffolding](https://book.kubebuilder.io/reference/rescaffold) mentions two contradictory statements:

> output-dir is the path to the directory where you want to generate the new project. Default i**s a subdirectory** in the current working directory.

> If neither input-dir nor output-dir are specified, the project will be regenerated in the current directory. 

The note also mentions:
>This approach facilitates comparison between your current local branch and the version stored upstream (e.g., GitHub main branch). This way, you can easily overlay your project’s code changes atop the new scaffold.

From the above defaults I expected the `output-dir` from the extra `note` was expecting something where I can easily diff the result with Git inside the same folders. Comparing with git is not really possible if its in a subfolder `output`dir`.

### Reproducing this issue

Run the `kubebuilder alpha generate` command and see that a `output-dir` is created.

### KubeBuilder (CLI) Version

4.3.1

### PROJECT version

3

### Other versions

go 1.23.4

### Extra Labels

/kind documentation

camilamacedo86:
Hi @Sijoma,

I think what raised confusion here: 'Default is a subdirectory in the current working directory.`.
However, please feel free to open a PR with any improvement/suggestion that you see fit for the documentation.
Then, we move from there. 

Thank you for looking into that.

Sijoma:
I tried to improve it a little bit in this draft PR https://github.com/kubernetes-sigs/kubebuilder/pull/4479 .

Maybe this helps.

PR #4479, opened 2025-01-09
 📖 improve docs of rescaffold

Sijoma:
So I was irritated by reading the documentation and also the CLI --help output.

I was surprised that the `output-dir` actually defaults to `output-dir`. I also did not understand how I could use git to diff the files with upstream if they are created in a sub directory. I would understand how this works if the files would be regenerated in place.

I've adjusted the docs & command to be more explicit about the default value `output-dir`.
Furthermore, I've adjusted the note to explain that you can diff the local files against the `output-dir` (the folder that was configured by the user).

relates #4478

### DIFF
--- a/docs/book/src/reference/rescaffold.md
+++ b/docs/book/src/reference/rescaffold.md
@@ -29,7 +29,7 @@ Currently, it supports two optional params, `input-dir` and `output-dir`.
 
 `input-dir` is the path to the existing project that you want to re-scaffold. Default is the current working directory.
 
-`output-dir` is the path to the directory where you want to generate the new project. Default is a subdirectory in the current working directory.
+`output-dir` is the path to the directory where you want to generate the new project. Default is a directory named "output-dir" in the current working directory.
 
 ```sh
 kubebuilder alpha generate --input-dir=/path/to/existing/project --output-dir=/path/to/new/project
@@ -39,7 +39,7 @@ kubebuilder alpha generate --input-dir=/path/to/existing/project --output-dir=/p
 <h1>Regarding `input-dir` and `output-dir`:</h1>
 
 If neither `input-dir` nor `output-dir` are specified, the project will be regenerated in the current directory.
-This approach facilitates comparison between your current local branch and the version stored upstream (e.g., GitHub main branch).
+This approach facilitates comparison between your code and regenerated files in `output-dir`.
 This way, you can easily overlay your project's code changes atop the new scaffold.
 
 </aside>
--- a/pkg/cli/alpha/command.go
+++ b/pkg/cli/alpha/command.go
@@ -16,6 +16,7 @@ package alpha
 import (
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+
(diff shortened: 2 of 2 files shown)


docs/book/src/reference/rescaffold.md
```diff
@@ -29,7 +29,7 @@ Currently, it supports two optional params, `input-dir` and `output-dir`.
 
 `input-dir` is the path to the existing project that you want to re-scaffold. Default is the current working directory.
 
-`output-dir` is the path to the directory where you want to generate the new project. Default is a subdirectory in the current working directory.
+`output-dir` is the path to the directory where you want to generate the new project. Default is a directory named "output-dir" in the current working directory.
```
camilamacedo86, line 32: If we not inform the `output-dir` it should be == $pwd , the current directory.
Are you facing a different behaviour?
Does it create a directory called `output-dir` and scaffold the project inside of it?

docs/book/src/reference/rescaffold.md
```diff
@@ -39,7 +39,7 @@ kubebuilder alpha generate --input-dir=/path/to/existing/project --output-dir=/p
 <h1>Regarding `input-dir` and `output-dir`:</h1>
 
 If neither `input-dir` nor `output-dir` are specified, the project will be regenerated in the current directory.
-This approach facilitates comparison between your current local branch and the version stored upstream (e.g., GitHub main branch).
+This approach facilitates comparison between your code and regenerated files in `output-dir`.
```
camilamacedo86, line 42: 👍

docs/book/src/reference/rescaffold.md
Sijoma: So that's the part that confused me.

The flag itself defaults to "" -> pkg/cli/alpha/command.go:59

The struct itself has a default then later here -> pkg/cli/alpha/internal/generate.go:46

image

It then sets it when this path is "" here as a default. -> pkg/cli/alpha/internal/generate.go:246

At least this is how I understand it and I get a `output-dir` folder.

camilamacedo86: OH we need to fix that.
It is a bug. By default it should be current dir == PWD

Sijoma: Just to get this right: So instead of appending `output-dir` to the PWD it should be just the PWD right? 

I can make a PR for this & maybe this then also removes the need to change the docs.

Sijoma: So something like this? 

image

camilamacedo86: We need to do a little more than that.
Also, we need to ensure it via the tests.
See: https://github.com/kubernetes-sigs/kubebuilder/pull/4500

Sijoma:
@camilamacedo86 with your changes I'm inclined to close this PR. WDYT?

camilamacedo86:
Hi @Sijoma 

> @camilamacedo86 with your changes I'm inclined to close this PR. WDYT?

I think that makes sense.
But if you see any room for improvement, please feel free to create a new PR.

Issue #4079, opened 2024-08-14
Make Kubebuilder `go install`-able


AndersBennedsgaard:
### What do you want to happen?

(I couldn't find any mentions about this, but feel free to close this if it is a duplicate)

I would like to add a target in my Makefile to install `kubebuilder` locally, such that all developers of our operator uses the same version of it.
https://book.kubebuilder.io/quick-start#installation doesn't explain how to install a specific version of Kubebuilder (you actually can't just change the version from `latest` in `https://go.kubebuilder.io/dl/latest/$(go env GOOS)/$(go env GOARCH)`) so I use the Github releases directly: `curl -L -o $(KUBEBUILDER) "https://github.com/kubernetes-sigs/kubebuilder/releases/download/$(KUBEBUILDER_VERSION)/kubebuilder_$(GOOS)_$(GOARCH)"`

This does work, but I would really prefer to just use `go install` since that is much simpler (doesn't require setting `GOOS` and `GOARCH`), I can make use of the `go-install-tool` helper, and it works even when you don't have `curl` and `chmod` installed such as on Windows.

By making Kubebuilder `go install`-able the installation steps for all distributions could be changed to use this, since it would be better (in my opinion) than the existing steps. It would also fix the installation steps for Windows in https://github.com/kubernetes-sigs/kubebuilder/issues/2940

camilamacedo86:
Hi @AndersBennedsgaard 

Thank you for raise this one. See that you can create a Makefile to download the binary as follows:

```
# Variables
KUBEBUILDER_VERSION ?= v4.1.1.  # Change to the desired version
KUBEBUILDER_BIN = $(HOME)/bin/kubebuilder
GOOS = $(shell go env GOOS)
GOARCH = $(shell go env GOARCH)
KUBEBUILDER_URL = https://github.com/kubernetes-sigs/kubebuilder/releases/download/$(KUBEBUILDER_VERSION)/kubebuilder_$(GOOS)_$(GOARCH).tar.gz

# Targets
.PHONY: all install-kubebuilder

all: install-kubebuilder

install-kubebuilder:
	@echo "Installing Kubebuilder version $(KUBEBUILDER_VERSION)..."
	@mkdir -p $(HOME)/bin
	@curl -L -o kubebuilder.tar.gz $(KUBEBUILDER_URL)
	@tar -xzf kubebuilder.tar.gz -C $(HOME)/bin/
	@rm kubebuilder.tar.gz
[...]

PR #4481, opened 2025-01-10
:bug: Make kubebuilder go install-able

migueleliasweb:
Closes https://github.com/kubernetes-sigs/kubebuilder/issues/4079

This PR makes `kubebuilder` go installable.

Changes:

- Renamed `main` package originally from `cmd/` to `cmd`
- Renamed `main` function previously inside `cmd/cmg.go` to `Run` - following Go Cobra's pattern
- Created new main file at the root of the project
- Renamed package on `doc.go` to `main`
- Fixed building `make build` target
- Fixed GoReleaser config

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,15 +46,15 @@ help: ## Display this help
 ##@ Build
 
 LD_FLAGS=-ldflags " \
-    -X main.kubeBuilderVersion=$(shell git describe --tags --dirty --broken) \
-    -X main.goos=$(shell go env GOOS) \
-    -X main.goarch=$(shell go env GOARCH) \
-    -X main.gitCommit=$(shell git rev-parse HEAD) \
-    -X main.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ') \
+    -X cmd.kubeBuilderVersion=$(shell git describe --tags --dirty --broken) \
+    -X cmd.goos=$(shell go env GOOS) \
+    -X cmd.goarch=$(shell go env GOARCH) \
+    -X cmd.gitCommit=$(shell git rev-parse HEAD) \
+    -X cmd.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ') \
     "
 .PHONY: build
 build: ## Build the project locally
-	go build $(LD_FLAGS) -o bin/kubebuilder ./cmd
+	go build $(LD_FLAGS) -o bin/kubebuilder
 
 .PHONY: install
 install: build ## Build and install the binary with the current source code. Use it to test your changes locally.
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main
+package cmd
 
 import (
 	"github.com/sirupsen/logrus"
@@ -36,7 +36,8 @@ func init() {
 	logrus.SetFormatter(&logrus.TextFormatter{DisableTimestamp: true})
 }
 
-func main() {
+// Run bootstraps & runs the CLI
+func Run() {
 	// Bundle plugin which built the golang projects scaffold with base.go/v4 and kustomize/v2 plugins
 	gov4Bundle, _ := plugin.NewBundleWithOptions(plugin.WithName(golang.DefaultNameQualifier),
 		plugin.WithVersion(plugin.Version{Number: 4}),
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -14,19 +14,22 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main
+package cmd
 
 import (
 	"fmt"
+	"runtime/debug"
 )
+
+const unknown = "unknown"
 
 // var needs to be used instead of const as ldflags is used to fill this
 // information in the release process
 var (
-	kubeBuilderVersion      = "unknown"
-	kubernetesVendorVersion = "unknown"
-	goos                    = "unknown"
-	goarch                  = "unknown"
+	kubeBuilderVersion      = unknown
+	kubernetesVendorVersion = unknown
+	goos                    = unknown
+	goarch                  = unknown
 	gitCommit               = "$Format:%H$" // sha1 from git, output of $(git rev-parse HEAD)
 
 	buildDate = "1970-01-01T00:00:00Z" // build date in ISO8601 format, output of $(date -u +'%Y-%m-%dT%H:%M:%SZ')
@@ -44,6 +47,12 @@ type version struct {
 
 // versionString returns the CLI version
 func versionString() string {
+	if kubeBuilderVersion == unknown {
+		if info, ok := debug.ReadBuildInfo(); ok && info.Main.Version != "" {
+			kubeBuilderVersion = info.Main.Version
+		}
+	}
+
 	return fmt.Sprintf("Version: %#v", version{
 		kubeBuilderVersion,
 		kubernetesVendorVersion,
--- a/build/.goreleaser.yml
+++ b/build/.goreleaser.yml
@@ -29,16 +29,15 @@ before:
 # Build a binary for each target in targets.
 builds:
   - id: kubebuilder
-    main: ./cmd
     binary: kubebuilder
     mod_timestamp: "{{ .CommitTimestamp }}"
     ldflags:
-      - -X main.kubeBuilderVersion={{ .Version }}
-      - -X main.goos={{ .Os }}
-      - -X main.goarch={{ .Arch }}
-      - -X main.gitCommit={{ .Commit }}
-      - -X main.buildDate={{ .Date }}
-      - -X main.kubernetesVendorVersion={{ .Env.KUBERNETES_VERSION }}
+      - -X cmd.kubeBuilderVersion={{ .Version }}
+      - -X cmd.goos={{ .Os }}
+      - -X cmd.goarch={{ .Arch }}
+      - -X cmd.gitCommit={{ .Commit }}
+      - -X cmd.buildDate={{ .Date }}
+      - -X cmd.kubernetesVendorVersion={{ .Env.KUBERNETES_VERSION }}
     targets:
       - linux_amd64
       - linux_arm64
--- a/main.go
+++ b/main.go
@@ -1,0 +1,23 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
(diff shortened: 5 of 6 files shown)


doc.go
```diff
@@ -18,4 +18,4 @@ limitations under the License.
 
 // Package kubebuilder contains pkged files compiled into the
 // go binaries.
-package kubebuilder
+package main
```
migueleliasweb, line 21: I couldn't find any references to this `go:generate` command anywhere. Also `github.com/markbates/pkger/cmd/pkger` is archived for a while. 

Maybe this can be removed?

camilamacedo86: Can we make this change?
We will need to check it further.
Because I believe that is what generates : https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4

So, can we change the name of the package?

Makefile
```diff
@@ -46,15 +46,15 @@ help: ## Display this help
...
-    -X main.kubeBuilderVersion=$(shell git describe --tags --dirty --broken) \
-    -X main.goos=$(shell go env GOOS) \
-    -X main.goarch=$(shell go env GOARCH) \
-    -X main.gitCommit=$(shell git rev-parse HEAD) \
-    -X main.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ') \
+    -X cmd.kubeBuilderVersion=$(shell git describe --tags --dirty --broken) \
+    -X cmd.goos=$(shell go env GOOS) \
+    -X cmd.goarch=$(shell go env GOARCH) \
+    -X cmd.gitCommit=$(shell git rev-parse HEAD) \
+    -X cmd.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ') \
```
camilamacedo86, line 53: The changes show fine 👍 
Thank you for that.

Let's do the next release, and then we check it out to ensure that nothing will break with.

doc.go
migueleliasweb: I'm really not sure that is generates anything right now. The package itself is not mentioned in `go.mod` and the `go generate` command fails for me.

Looking at the repo for `Pcker`:

Screenshot from 2025-01-12 22-17-51

the code seems to be supposed to be a predecessor of [go embed](https://pkg.go.dev/embed) commands that are now part of the standard library.

Note that https://github.com/gobuffalo/packr is also archived in favour of `go embed` native.

Have a look at the commit [fe73ff7b99](https://github.com/kubernetes-sigs/kubebuilder/commit/fe73ff7b99#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R9) and it seems it was initially tied to a command inside the alpha subcommand `kubebuilder alpha config-gen` - this doesn't seem to be available now and I assume was renamed later to `kubebuilder alpha generate`?

Sadly, changing the name of the package of `doc.go` to `kubebuilder` again would prevent the existance of the new `main.go` as both files would have different package names.

If you **really** want to leave this file alone, there are still two alternatives that although much less nice, would mean we don't have to touch this file: https://github.com/kubernetes-sigs/kubebuilder/issues/4079#issuecomment-2576400007.

migueleliasweb: I'm not sure if this file must be present at the root, if not, it could be moved somewhere else?

camilamacedo86: Hi @migueleliasweb 

I will need to have sometimes to check this one properly.

camilamacedo86:
Hi @migueleliasweb 

You can see that the changes made broke the e2e tests.
The `[pull-kubebuilder-e2e-k8s-1-32-0](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kubebuilder/4481/pull-kubebuilder-e2e-k8s-1-32-0/1878418648549298176)` you can ignore because it is a bug with prow and the 1.32 is not working yet. 

But all other failures would be required be solved

migueleliasweb:
I'll have a look at the failing tests tomorrow :saluting_face:

Scutua (approved):

cmd/version.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main
+package cmd
```
dmvolod, line 17: Seems to version getting form the `info.Main.Version` should be adopted here like in the controller-tools project https://github.com/kubernetes-sigs/controller-tools/blob/main/pkg/version/version.go

Otherwise, the version printed after the `go install ...` will be always `unknown`.

migueleliasweb: I'm not sure if Kubebuilder itself tracks the same version as controller runtime. That might be a change in behaviour. If one of the maintainers wants to make that change, it's fine for me.

dmvolod: I'm talking about controller-tools version, I just say about approach to get a `kubeBuilderVersion` value from the the binary `info.Main.Version` value for the kubebuilder project itself.

cmd/cmd.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main
+package cmd
```
camilamacedo86, line 17: Why do we need to change the name of the package?

doc.go
camilamacedo86: I created : https://github.com/kubernetes-sigs/kubebuilder/pull/4504

if the docs are generated all fine, then we get it removed
if after we discover that it is used in some place and has a purpose, then we can quickly revert.
Thank you for highlighting that.

cmd/version.go
camilamacedo86: @migueleliasweb can you please address this one ?

camilamacedo86 (commented):
Hi @migueleliasweb 

I think to get this one merged we just need:

a) address this comment https://github.com/kubernetes-sigs/kubebuilder/pull/4481/files#r1920699959. 

b) can you please squash the commits? we have the policy of 1 commit per PR

c) we should get this one merged https://github.com/kubernetes-sigs/kubebuilder/pull/4504 and then this one here rebased with so you will add the main.go in the root. If we need to revert for any reason that will be helpful, thank you a lot 🥇 

Than you a lot for your help here.

cmd/version.go
migueleliasweb: I'not sure I fully understand what needs to be done here. Using `debug.ReadBuildInfo()` might yield a change in behaviour compared to the current setup.

Would it be okay to address this in a separate PR? (I'm just trying to make this one specific for the `go-installable` feature for kubebuilder.

If we still want to move ahead, I'm happy to mimic the same logic as in https://github.com/kubernetes-sigs/controller-tools/blob/main/pkg/version/version.go.

migueleliasweb: Is this what we're looking for?

If the `kubeBuilderVersion` is not set, fallback to `info.Main.Version`.

```go
func versionString() string {
	if kubeBuilderVersion == "" {
		info, ok := debug.ReadBuildInfo()
		if ok || info.Main.Version != "" {
			kubeBuilderVersion = info.Main.Version
		}
	}

	return fmt.Sprintf("Version: %#v", version{
		kubeBuilderVersion,
		kubernetesVendorVersion,
		gitCommit,
		buildDate,
		goos,
		goarch,
	})
}
```

migueleliasweb: Pushed a change about this. Let me know if it looks good so I can squash everything before the merge. :saluting_face:

cmd/version.go
```diff
@@ -44,6 +45,12 @@ type version struct {
 
 // versionString returns the CLI version
 func versionString() string {
+	if kubeBuilderVersion == "" {
```
dmvolod, line 48: It seems to this code will never calls, as by default kubeBuilderVersion is always `unknown`.
You should introduce a new constant with "unknown" value, for example
```golang 
const unknown = "unknown"

var (
	kubeBuilderVersion      = unknown
	kubernetesVendorVersion = unknown
	goos                    = unknown
	goarch                  = unknown
	...
)
```
and compare here with it instead of empty string.

dmvolod: @migueleliasweb  this code should be changed also
```
const unknown = "unknown"

var (
	kubeBuilderVersion      = unknown
	kubernetesVendorVersion = unknown
	goos                    = unknown
	goarch                  = unknown
	...
)
```

migueleliasweb: Fixed changes. I was trying to find your comment but Github's UI wasn't helping it much. Found it a few minutes later.

camilamacedo86:
Hi @migueleliasweb 

Can you please squash the commits so we can get this one merged?

migueleliasweb:
> Hi @migueleliasweb
> 
> Can you please squash the commits so we can get this one merged?

Will do!

Do you think, maybe after this PR is merged, small changes to the docs should be done to mention the new installation method?

camilamacedo86:
Hi @migueleliasweb 

I am thinking about how we can test this one.
So, could you please let us know how you check that after those changes would be able to go install the project?

migueleliasweb:
> Hi @migueleliasweb
> 
> I am thinking about how we can test this one. So, could you please let us know how you check that after those changes would be able to go install the project?

Hi @camilamacedo86 . I reckon the quickest option here would be to merge and then get the hash of the commit so we can then issue a: `go install -v sigs.k8s.io/kubebuilder/v4@<hash>`. We can then quickly check if the binary is installed and named correctly.

The slightly more safe option would be to merge this onto a separate branch within `kubebuilder`. This would be enough for us to test it with: `go install -v sigs.k8s.io/kubebuilder/v4@<branch-name>`. Note that we can do this today with: `go install -v sigs.k8s.io/kubebuilder/v4/cmd@release-4.4` but the outcode is a `cmd` binary instead of `kubebuilder` :grimacing: .

Given the current setup we have in `cmd/version.go` some build info might not be set correctly.

I've made a POC in https://github.com/migueleliasweb/go-installable/ so you can see how we would get as much information as possible from the `build info` during a `go install`.

The TL;DR is that we have a few ways to run `kubebuilder`:

- With `go run` locally after cloning the repo
- With `go build -o kubebuilder` and then `./kubebuilder`
- With `go install sigs.k8s.io/kubebuilder/v4@<commit-hash-ref>` (should generate a bin named `kubebuilder`)
- With `go install sigs.k8s.io/kubebuilder/v4@<release-ref>` (should generate a bin named `kubebuilder`)

Give it a go with my POC and see how the behaviour is.

Try running:

- `go install -v github.com/migueleliasweb/go-installable@v1.0.5`
- `go install -v github.com/migueleliasweb/go-installable@d91d9ebe30f987e0b9cc2690fd99c7f92ac1807a`
- `go build -o go-installable` and then `go-installable`
- `go run main.go`

The same logic can be ported to `kubebuilder` if you think it's worth it.

camilamacedo86 (approved):
/lgtm

PR #4492, opened 2025-01-16
⚠️ Upgrade controller-runtime from v0.19.4 to v0.20.0 and k8s deps from 1.31 to 1.32

camilamacedo86:


camilamacedo86:
We just bumping here, so I will not make anybody waste time on reviews

PR #4494, opened 2025-01-16
✨ Upgrade GolangCI Lint from v1.62.2 to v1.63.4

camilamacedo86:
✨ Upgrade GolangCI Lint from v1.62.2 to v1.63.4


camilamacedo86:
I'm just bumping pass on the CI. So moving forward.

PR #4498, opened 2025-01-17
✨ e2e tests: Upgrade Prometheus Operator used from `v0.77.1` to `v0.79.2`

camilamacedo86:


camilamacedo86:
I think we can remove the Prometheus Operator from the E2E tests 
It is not really used on it and simplify the code implementation for end users 
So, I am closing this one

PR #4500, opened 2025-01-17
🌱 fix: alpha generate command. If the output-dir path is not informed than it should be the current directory

camilamacedo86:
If not informed of an input directory and an output directory, then the tool should re-generate the project in the current directory.


camilamacedo86:
c/c @Sijoma

Sijoma (approved):
Looks good to me 👍

PR #4501, opened 2025-01-17
🐛 fix: kubebuilder alpha generate command to allow re-create projects with webhooks for external-apis

camilamacedo86:
Allow re-generate projects which were previously scaffolded with an external API; a webhook was created for those. 

The issue is encountered when using the command before this Pull Request to re-generate the scaffold under test data, for example.


camilamacedo86:
This one is too small, so I will not make anybody waste time on reviews of it.

PR #4505, opened 2025-01-19
✨ (go/v4): Follow cert-manager org name changes

zchee:
The `cert-manager` project has been transferred to the `cert-manager` organization from the jetstack.

Fix the `certmanagerURLTmpl` value to follow them.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
@@ -33,7 +33,7 @@ const (
 		"releases/download/%s/bundle.yaml"
 
 	certmanagerVersion = "v1.16.3"
-	certmanagerURLTmpl = "https://github.com/jetstack/cert-manager/releases/download/%s/cert-manager.yaml"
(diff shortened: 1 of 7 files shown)


camilamacedo86 (commented):
@zchee 

Thank you a lot for this contribution 🥇 
But the docs and testdata samples are auto-generate when we run the makefile target `make generate`

So, this change must be done by:

a) Updating the template: https://github.com/kubernetes-sigs/kubebuilder/blob/8c048fff94da8cc846b7efa6a3958160eafd8294/pkg/plugins/golang/v4/scaffolds/internal/templates/test/utils/utils.go#L63

b) Then, running `make generate` to update all

Moreover, we need to update other places that use it for the helm plugin and our e2e tests. 

So, before running the command, we need to update the helm plugin template (assuming that it also changed, otherwise, if not, you do not need to do this update.): https://github.com/kubernetes-sigs/kubebuilder/blob/8c048fff94da8cc846b7efa6a3958160eafd8294/pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/github/test_chart.go#L100

Therefore, could you please:

- Update the templates as described
- Run `make generate` to ensure that all be generated accordingly
- Then, squash the commits and push for we get this one merged

Thank you a lot for get this one 🥇

camilamacedo86:
@zchee 

PR #4512, opened 2025-01-21
🌱 : Add YAML Linting to CI Workflow for Scaffolded Charts

### LINKED ISSUE #4503: Add YAML Linting for Sample Files and Scaffolded Chart Files

Currently, we need to ensure YAML files are properly linted for consistency and correctness. This includes:
- Linting the sample files present in the repository.
- Linting the files scaffolded under dist/chart for the project-v4-with-plugins configuration.

We already have a GitHub Action configured for this purpose at github/workflows/lint-sample.yml. This action should be updated or extended to cover both. Motivation: https://github.com/kubernetes-sigs/kubebuilder/pull/4490

@sebastian-philipp ^ thank you for raise this one 🥇 

### Extra Labels

_No response_

sarthaksarthak9:
fixes #4503

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,9 +111,9 @@ lint-config: golangci-lint ## Verify golangci-lint linter configuration
 
 .PHONY: yamllint
 yamllint:
-	@files=$$(find testdata -name '*.yaml' ! -path 'testdata/*/dist/*'); \
-    	docker run --rm $$(tty -s && echo "-it" || echo) -v $(PWD):/data cytopia/yamllint:latest $$files -d "{extends: relaxed, rules: {line-length: {max: 120}}}" --no-warnings
-
+	@files=$$(find testdata -name '*.yaml' ! -path 'testdata/*/dist/chart/install.yaml'); \
+	docker run --rm $$(tty -s && echo "-it" || echo) -v $(PWD):/data cytopia/yamllint:latest $$files -d "{extends: relaxed, rules: {line-length: {max: 120}}}" --no-warnings
+		
 GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
 golangci-lint:
 	@[ -f $(GOLANGCI_LINT) ] || { \


Makefile
mateusoliveira43, line 1: would be better if users uses helm, to add yaml lint command to scaffold Makefile?

Makefile
```diff
@@ -121,6 +121,11 @@ golangci-lint:
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell dirname $(GOLANGCI_LINT)) v1.63.4 ;\
 	}
 
+.PHONY: yamllint-chart
+yamllint-chart:
+	@files=$$(find testdata/project-v4-with-plugins/dist/chart -name '*.yaml'); \
+    	docker run --rm $$(tty -s && echo "-it" || echo) -v $(PWD):/data cytopia/yamllint:latest $$files -d "{extends: relaxed, rules: {line-length: {max: 120}}}" --no-warnings
```
camilamacedo86, line 127: Could we add the test without add the linter here?
Or if we add the linter that need to be valid for all yaml files not only dist/chart

Makefile
camilamacedo86: I think we cannot add the target only for helm by default. 

Also, see that we have the lint in the Makefile for the project

https://github.com/kubernetes-sigs/kubebuilder/blob/master/Makefile#L119-L122

So, maybe the ONLY change we need here is to change this target to ensure that it will also lint the helm-chart in the dist dir and will ONLY ignore the install.yaml. 

This target is called in the CI already: https://github.com/kubernetes-sigs/kubebuilder/blob/7ee23df2b97ccb209df4e05c78830266c7b8d8bb/.github/workflows/lint.yml#L31-L36

sarthaksarthak9: Do we have to make this changes in this pr ?

Kavinjsir: @sarthaksarthak9  I’m on board with @camilamacedo86. We don’t need a separate target like `yamllint-chart`; we can just update the existing `yamllint` target. Currently, it excludes everything under `testdata/*/dist/*`: 

```
find testdata -name '*.yaml' ! -path 'testdata/*/dist/*'
```

From what I'm following, we just need to adjust that exclusion so the linter picks up files in `*/dist/chart`, but still skips `install.yaml`.

.github/workflows/lint-sample.yml
```diff
@@ -42,3 +42,17 @@ jobs:
         working-directory: ${{ matrix.folder }}
         run: make lint
```
Kavinjsir, line 43: See `make lint` will include all linter execution, we don't have to define a separate step

Kavinjsir (changes requested):

.github/workflows/lint-sample.yml
sarthaksarthak9: yes

.github/workflows/lint-sample.yml
```diff
@@ -41,4 +41,3 @@ jobs:
       - name: Run linter via makefile target
         working-directory: ${{ matrix.folder }}
         run: make lint
-
```
camilamacedo86, line 44: We need to revert any change on this file
Can you please do that?

sarthaksarthak9: yes

Makefile
```diff
@@ -111,8 +111,8 @@ lint-config: golangci-lint ## Verify golangci-lint linter configuration
 
 .PHONY: yamllint
 yamllint:
-	@files=$$(find testdata -name '*.yaml' ! -path 'testdata/*/dist/*'); \
-    	docker run --rm $$(tty -s && echo "-it" || echo) -v $(PWD):/data cytopia/yamllint:latest $$files -d "{extends: relaxed, rules: {line-length: {max: 120}}}" --no-warnings
+	@files=$$(find testdata -name '*.yaml' ! -path 'testdata/*/dist/*' -o -path 'testdata/*/dist/chart/*.yaml' ! -path 'testdata/*/dist/chart/install.yaml'); \
+	docker run --rm $$(tty -s && echo "-it" || echo) -v $(PWD):/data cytopia/yamllint:latest $$files -d "{extends: relaxed, rules: {line-length: {max: 120}}}" --no-warnings
```
camilamacedo86, line 115: @sarthaksarthak9 

I think the ONLY change that we need here to acomplish the goal is:
- Instead of EXCLUDE  all under testdata/dist/* dir we need to exclude ONLY the install.yaml

That means probably replace `'testdata/*/dist/*'` with `! -path 'testdata/*/dist/install.yaml'`

Makefile
```diff
@@ -111,9 +111,9 @@ lint-config: golangci-lint ## Verify golangci-lint linter configuration
 
 .PHONY: yamllint
 yamllint:
-	@files=$$(find testdata -name '*.yaml' ! -path 'testdata/*/dist/*'); \
-    	docker run --rm $$(tty -s && echo "-it" || echo) -v $(PWD):/data cytopia/yamllint:latest $$files -d "{extends: relaxed, rules: {line-length: {max: 120}}}" --no-warnings
-
+	@files=$$(find testdata -name '*.yaml' ! -path 'testdata/*/dist/chart/install.yaml'); \
+	docker run --rm $$(tty -s && echo "-it" || echo) -v $(PWD):/data cytopia/yamllint:latest $$files -d "{extends: relaxed, rules: {line-length: {max: 120}}}" --no-warnings
```
camilamacedo86, line 115: @sarthaksarthak9 

Please, squash the commits. 
In order to keep a better git history and allow us to revert things if we need to do so
We should have ONLY 1 commit for this change

Can you please do that to allow us to move forward here?

camilamacedo86 (approved):
Thank you 🥇 
It is great work !!!

/lgtm
/approve

camilamacedo86:
/ok-to-test

@sarthaksarthak9

PS: I changed the emoji — I hope you don’t mind! We use 🌱 for changes related to tests and internal things that don’t impact end users, like this one.

Why does that matter? It’s helpful for us in generating the release notes. More info here: [PR Process](https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#pr-process)

PR #4525, opened 2025-01-25
:seedling: update dockerfiles to just copy the code

### LINKED ISSUE #4524: Dockerfile templates always cause confusion
For Kueue, JobSet, LWS and Instaslice-operator, we find that the default template for Dockerfiles does not age very well.

https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-with-plugins/Dockerfile

We notice that when developers add new folders to the repo, they forget to add them to the Dockerfile.

We have started just copying the folder (COPY . .) and using that instead.

Can we change the template to do this by default?

kannon92:
Fixes #4524 

As I have worked on a few kubebuilder based projects, I always find over time that I switch to using a similar behavior for the template.

It causes quite a bit of confusion as it is usually not a common practice when you add folders that you must remember to add them to the Dockerfile also.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/Dockerfile
+++ b/docs/book/src/cronjob-tutorial/testdata/project/Dockerfile
@@ -4,17 +4,8 @@ ARG TARGETOS
 ARG TARGETARCH
 
 WORKDIR /workspace
-# Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
-# cache deps before building and copying source so that we don't need to re-download as much
-# and so that source changes don't invalidate our downloaded layer
-RUN go mod download
-
 # Copy the go source
-COPY cmd/main.go cmd/main.go
-COPY api/ api/
-COPY internal/ internal/
+COPY . .
 
 # Build
 # the GOARCH has not a default value to allow the binary be built according to the host where the command
--- a/docs/book/src/getting-started/testdata/project/Dockerfile
+++ b/docs/book/src/getting-started/testdata/project/Dockerfile
@@ -4,17 +4,8 @@ ARG TARGETOS
 ARG TARGETARCH
 
 WORKDIR /workspace
-# Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
-# cache deps before building and copying source so that we don't need to re-download as much
-# and so that source changes don't invalidate our downloaded layer
-RUN go mod download
-
 # Copy the go source
-COPY cmd/main.go cmd/main.go
-COPY api/ api/
-COPY internal/ internal/
+COPY . .
 
 # Build
 # the GOARCH has not a default value to allow the binary be built according to the host where the command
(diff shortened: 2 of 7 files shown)


Kavinjsir:
Hey @kannon92 splitting out dependency installation steps in your Dockerfile can help Docker build cache more efficiently and usually speeds up the build process. You might also want to check out the official Docker CLI for generating a Go-specific Dockerfile—it shows exactly why copying go.mod and go.sum in separate steps is a best practice.

```dockerfile
# Download dependencies as a separate step to take advantage of Docker's caching.
# Leverage a cache mount to /go/pkg/mod/ to speed up subsequent builds.
# Leverage bind mounts to go.sum and go.mod to avoid having to copy them into
# the container.
RUN --mount=type=cache,target=/go/pkg/mod/ \
    --mount=type=bind,source=go.mod,target=go.mod \
    go mod download -x

# This is the architecture you're building for, which is passed in by the builder.
# Placing it here allows the previous steps to be cached across architectures.
ARG TARGETARCH

# Build the application.
# Leverage a cache mount to /go/pkg/mod/ to speed up subsequent builds.
# Leverage a bind mount to the current directory to avoid having to copy the
# source code into the container.
RUN --mount=type=cache,target=/go/pkg/mod/ \
    --mount=type=bind,target=. \
    CGO_ENABLED=0 GOARCH=$TARGETARCH go build -o /bin/server ./api

```

docs/book/src/cronjob-tutorial/testdata/project/Dockerfile
```diff
@@ -4,17 +4,8 @@ ARG TARGETOS
...
-COPY go.sum go.sum
-# cache deps before building and copying source so that we don't need to re-download as much
-# and so that source changes don't invalidate our downloaded layer
-RUN go mod download
-
 # Copy the go source
-COPY cmd/main.go cmd/main.go
-COPY api/ api/
-COPY internal/ internal/
+COPY . .
```
camilamacedo86, line 8: Thank you for your contribution. 🎉 

The proposed change replaces the current approach, where we only add specific paths (cmd/, api/, internal/, etc.), with `COPY . .`, which adds everything from the root of the project. Am I correct?

This means that unnecessary files and directories like hack/, test/, README, and other root-level files will now be included in the manager image, even if they are not required.

While this might seem convenient, however, it shows like goes against best practices by unnecessarily bloating the image. That is why we promote to explicitly copy only the paths we need to ensure a clean and optimized image, avoiding the inclusion of unrelated files.

Does that make sense?

/hold

camilamacedo86: Could we agree to close this one?

camilamacedo86:
Hi @kannon92

Thank you a lot for your contribution 🥇 and for looking to make kubebuilder

But as@Kavinjsir pointed out: https://github.com/kubernetes-sigs/kubebuilder/pull/4525#issuecomment-2614054267
Also, see my comment: https://github.com/kubernetes-sigs/kubebuilder/pull/4525#discussion_r1929721671

What you suggested does not seem to align with Docker's best practices: https://docs.docker.com/build/building/best-practices/#dont-install-unnecessary-packages

Therefore, I do not think that we should not move forward here.
Does that make sense ?

So, I hope that you do not mind, but I am closing this one and the PR as deferred

PR #4526, opened 2025-01-25
✨ (go/v4): add race flag for go test command at the makefile

PG2000:
In order to detect data races early it would make sense to use the race flag 
for the most common generated go commands.

In grown projects built on kubebuilder i observed that with the timing and growing codebases
some data races occured in tests and then its hard to find the causing commit or code chnage.

Maybe this change can help to find them earlier.

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/dockerfile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/dockerfile.go
@@ -61,7 +61,7 @@ COPY internal/ internal/
 # was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
 # the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -race -a -o manager cmd/main.go
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager cmd/main.go
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/dockerfile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/dockerfile.go
@@ -61,7 +61,7 @@ COPY internal/ internal/
 # was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
 # the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager cmd/main.go
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -race -a -o manager cmd/main.go
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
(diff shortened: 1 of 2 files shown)


camilamacedo86:
/ok-to-test

It is a good catcher.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -138,7 +138,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet setup-envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -race -coverprofile cover.out
```
camilamacedo86, line 142: Also, see that the changes here broke the tests

```
      #15 [builder 9/9] RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -race -a -o manager cmd/main.go
      #15 0.430 go: -race requires cgo; enable cgo by setting CGO_ENABLED=1
      #15 ERROR: process "/bin/sh -c CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -race -a -o manager cmd/main.go" did not complete successfully: exit code: 2
      ------
```

It does not seems to work.
So, can you please check out?

pkg/plugins/golang/v4/scaffolds/internal/templates/dockerfile.go
```diff
@@ -61,7 +61,7 @@ COPY internal/ internal/
 # was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
```
camilamacedo86, line 61: Also, see that the changes here broke the tests

```
      #15 [builder 9/9] RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -race -a -o manager cmd/main.go
      #15 0.430 go: -race requires cgo; enable cgo by setting CGO_ENABLED=1
      #15 ERROR: process "/bin/sh -c CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -race -a -o manager cmd/main.go" did not complete successfully: exit code: 2
      ------
```

It does not seems to work.
So, can you please check out?

PG2000: Will check this. Seems like CGO must be enabled for go build. Will try to leave this part out and start on the other problem where a data race was detected.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -155,7 +155,7 @@ test-e2e: manifests generate fmt vet ## Run the e2e tests. Expected an isolated
 		echo "No Kind cluster is running. Please start a Kind cluster before running the e2e tests."; \
 		exit 1; \
 	}
-	go test ./test/e2e/ -v -ginkgo.v
+	go test ./test/e2e/ -race -v -ginkgo.v
```
dmvolod, line 158: This slow-down all process for several times and required cgo to be enabled, and on non-Darwin systems requires an installed C compiler. Not sure, that it should be enabled by default.

Probably it can be an option or flag and customer should decide if it enabled or not.
For example, we are using it for long running CI processes running night, but not in regular pipelines for each PR.

camilamacedo86: That sounds great! Thank you a lot for checking it.
So, It seems we have reached a consensus that this should not be added to the default scaffold. 
Users who need it can  always update their own projects accordingly. WDYT? Is that something that make sense based on your investigation?

If so, could you please update the final conclusion in the issue and close both? This way, we can keep track of the reasoning behind not including it.

Thanks!

dmvolod: If we want to add some extra flexibility, we can introduce a field that will be empty by default with the ability to change it, and which we will add as arguments when calling tests and/or build.
We can add two different fields for build and tests.
But it is very dangerous to forcefully add the `-race`  flag by default.

camilamacedo86:
Hi @PG2000,

Based on the reviews, I believe we can conclude that having the -race flag defined in the default scaffold might not be the best approach. The primary concern is that it could introduce unnecessary performance overhead for users, as well as additional dependencies (such as requiring CGO and a C compiler on non-Darwin systems). Most users don’t need it for every test, and it’s more suitable for specific use cases like long-running CI tests or when debugging race conditions.

However, please note that users can always modify their own projects to enable the flag as needed.

With that said, I hope you don’t mind, but I’ll go ahead and close this as deferred.

Feel free to reopen it if you think it's necessary, and continue contributing to the project.

Thanks!

PR #4529, opened 2025-01-27
:bug: (go/v4): e2e test: Removed Prometheus dependency

### LINKED ISSUE #4527: Remove Prometheus Dependency from e2e Test Scaffolds in Kubebuilder

The e2e test scaffolds provided by Kubebuilder currently include a Prometheus dependency. Specifically, they apply a ServiceMonitor as part of the e2e tests. While this was originally added to align with the default Prometheus integration that Kubebuilder validates in its own e2e tests, it does not provide substantial value for end-users and introduces unnecessary complexity.

Issues:

**Unnecessary Dependency:**

- The dependency only checks if a ServiceMonitor is applied.
- It does not consume or validate metrics.

**Added Complexity:**

- Prometheus setup complicates the scaffolds.
- Not essential for basic e2e testing.

**Existing Validation:**

- Prometheus integration is already tested in Kubebuilder’s internal e2e tests.

**Proposal:**

- Remove Prometheus from e2e tests scaffolds: https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e
- Simplify scaffolds to focus on core functionality.

**Benefits:**

- Simplifies usage for end-users.
- Reduces unnecessary dependencies.
- Aligns scaffolds with practical, user-focused goals.

### Extra Labels

_No response_

anshuman-agarwala:
Fixes #4527

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_suite_test.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_suite_test.go
@@ -30,16 +30,14 @@ import (
 
 var (
 	// Optional Environment Variables:
-	// - PROMETHEUS_INSTALL_SKIP=true: Skips Prometheus Operator installation during test setup.
 	// - CERT_MANAGER_INSTALL_SKIP=true: Skips CertManager installation during test setup.
-	// These variables are useful if Prometheus or CertManager is already installed, avoiding
+	// These variables are useful if CertManager is already installed, avoiding
 	// re-installation and conflicts.
-	skipPrometheusInstall  = os.Getenv("PROMETHEUS_INSTALL_SKIP") == "true"
 	skipCertManagerInstall = os.Getenv("CERT_MANAGER_INSTALL_SKIP") == "true"
-	// isPrometheusOperatorAlreadyInstalled will be set true when prometheus CRDs be found on the cluster
-	isPrometheusOperatorAlreadyInstalled = false
 	// isCertManagerAlreadyInstalled will be set true when CertManager CRDs be found on the cluster
 	isCertManagerAlreadyInstalled = false
+	// isPrometheusOperatorAlreadyInstalled will be set true when prometheus CRDs be found on the cluster
+	isPrometheusOperatorAlreadyInstalled = false
 
 	// projectImage is the name of the image which will be build and loaded
 	// with the code source changes to be tested.
@@ -49,7 +47,7 @@ var (
 // TestE2E runs the end-to-end (e2e) test suite for the project. These tests execute in an isolated,
 // temporary environment to validate project changes with the the purposed to be used in CI jobs.
 // The default setup requires Kind, builds/loads the Manager Docker image locally, and installs
-// CertManager and Prometheus.
+// CertManager.
 func TestE2E(t *testing.T) {
 	RegisterFailHandler(Fail)
 	_, _ = fmt.Fprintf(GinkgoWriter, "Starting project integration test suite\n")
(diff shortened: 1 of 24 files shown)


docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_suite_test.go
```diff
@@ -30,14 +30,10 @@ import (
 
 var (
```
camilamacedo86, line 31: @anshuman-agarwala 

that is great 🎉 
We just need either remove here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/Makefile#L67

camilamacedo86: The change need to be done here: https://github.com/kubernetes-sigs/kubebuilder/blob/87b0139338a4bdbe79756d83b59368e27fdd0f59/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go#L143-L147


Also, can you please ensure that we have all squashed in the same commit so we can merge this one?

anshuman-agarwala: Done

camilamacedo86: You need change the template [kubebuilder/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go](https://github.com/kubernetes-sigs/kubebuilder/blob/87b0139338a4bdbe79756d83b59368e27fdd0f59/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go#L143-L147)

And then run `make install && make generate`
It will re-generate all samples and docs with this change

camilamacedo86 (commented):
/ok-to-test

camilamacedo86:
Hi @anshuman-agarwala 

Thank you a lot 🎉 
In this case, we are changing the e2e tests of the projects, which are scaffolded for end users
So, the emoji 🌱 cannot be used, since it is for changes that does not impact the end users 
I update the title since we use it for our release notes I hope that you do not mind.

anshuman-agarwala:
> Hi @anshuman-agarwala
> 
> Thank you a lot 🎉 In this case, we are changing the e2e tests of the projects, which are scaffolded for end users So, the emoji 🌱 cannot be used, since it is for changes that does not impact the end users I update the title since we use it for our release notes I hope that you do not mind.

No Problems!
I see that e2e test is failing, the logs are as follows
```
running: make deploy IMG=example.com/project:v0.0.1
  [FAILED] in [BeforeAll] - /home/runner/work/kubebuilder/kubebuilder/docs/book/src/multiversion-tutorial/testdata/project/test/e2e/e2e_test.go:71
```
Could you help me with how I can fix this?

camilamacedo86:
Hi @anshuman-agarwala,

The tutorials in the documentation are built from the sample projects under `book/../testdata`, such as:  
🔗 https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/multiversion-tutorial/testdata/project  

To ensure consistency, we update these samples based on PR changes and verify them using e2e tests. The additional code and modifications on top of the scaffolded content are implemented in:  
🔗 https://github.com/kubernetes-sigs/kubebuilder/tree/master/hack/docs  

### The Issue:  
The error occurs because Prometheus integration is **enabled** in the **Kustomize** files, but Prometheus itself is **not installed**.  

### Possible Solutions:  
✅ **Option A**:  
Update [hack/docs](https://github.com/kubernetes-sigs/kubebuilder/tree/master/hack/docs) to ensure that the removed code from the default scaffold is explicitly added to the sample used in the **Multiversion Tutorial**. This means adding the necessary code under:  
🔗 https://github.com/kubernetes-sigs/kubebuilder/tree/master/hack/docs/internal/multiversion-tutorial  

With this approach, when we run `make generate-docs` (triggered by `make generate`), we will maintain the current e2e tests at:  
🔗 https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/multiversion-tutorial/testdata/project/test/e2e  

✅ **Option B**:  
Modify the code so that it no longer uncomments the **Kustomize** files to enable Prometheus by default.  

Personally, **Option A** seems preferable since it keeps this scenario as a documented example for end users.  

**WDYT?** 🚀

anshuman-agarwala:
Hi @camilamacedo86 
Option A sounds good but I looked at the other E2E book samples runs, they have failed in the same way, is it due to the same problem? In that case we may need to look at Option B, right?

camilamacedo86:
Hi @anshuman-agarwala,

> Option A sounds good, but I noticed that other E2E book sample runs have failed in the same way. Could this be due to the same underlying issue? If so, should we consider Option B instead?  

Good catch! 🥇  

## **Getting Started Sample**
  
For the `getting-started` sample, we should **not** simply uncomment the line in the `kustomization.yaml` file to enable the integration with prometheus:  
🔗 [kustomization.yaml#L26-L27](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/getting-started/testdata/project/config/default/kustomization.yaml#L26-L27)  

### **Why?**  
- This is a basic example, and we don’t need to extend it further.  

### **How?**  
- We should **delete** the following function call instead:  
🔗 [generate_getting_started.go#L52-L57](https://github.com/kubernetes-sigs/kubebuilder/blob/master/hack/docs/internal/getting-started/generate_getting_started.go#L52-L57)  

## **CronJob and MultiVersion Samples**  
The `CronJob` and `MultiVersion` samples share a similar generation process. The `MultiVersion` sample builds on the `CronJob` implementation.  See in the `hack/docs` that we re-use it.

So, to add E2E test we can: 
- Modify the CronJob generation script:  
  🔗 [generate_cronjob.go#L85-L127](https://github.com/kubernetes-sigs/kubebuilder/blob/master/hack/docs/internal/cronjob-tutorial/generate_cronjob.go#L85-L127)  
- Introduce a **new function** that injects E2E test code.  
  - Similar to this function:  
    🔗 [generate_cronjob.go#L629-L642](https://github.com/kubernetes-sigs/kubebuilder/blob/master/hack/docs/internal/cronjob-tutorial/generate_cronjob.go#L629-L642)  
  - This function should find and replace content in the `tests/e2e` path.  

### **Where?**  
We should create a **new file** to store constants for `e2e_implementation.go` under:  
🔗 [cronjob-tutorial directory](https://github.com/kubernetes-sigs/kubebuilder/tree/master/hack/docs/internal/cronjob-tutorial)  
(similar to the existing structure).  

In this way, we cover more cases in the tests and provide more examples for end users. WDYT?

camilamacedo86:
HI @anshuman-agarwala 

Great work 🥇 
It seems missing only a nit for we are able to get this one merged.
See the lint issues. You can run `make lint` locally to check them out. 
Can you please send a fix?

camilamacedo86 (approved):
/lgtm
/approve

PR #4536, opened 2025-02-02
:bug: (kustomize/v2): fix ServiceMonitor with TLS kustomize scaffolding

### LINKED ISSUE #4533: Kubebuilder scaffolding with HTTPS endpoint for metrics endpoint with custom certs doesn't work

Broken:
Metrics server with HTTPS and cert-manager certs doesn't work
1. `config/prometheus/monitor_tls_patch.yaml` path completely overwrites the ServiceMonitor `endpoints` field, resulting in omission of existing fields like `path`, `scheme` etc
2. Prometheus is not able to validate the metric server certificate because it is connecting all underlying service pods over POD IP whereas certificate SANs have metrics service names. So it’s not able to validate the host authenticity.  Its throwing error:
```
x509: cannot validate certificate for 10.20.5.28 because it doesn't contain any IP SANs
```
PFA screenshot for the same:

Image

Expected:
1. `config/prometheus/monitor_tls_patch.yaml` patch should only update the `tlsConfig` field with the required certs, leaving all existing fields intact.
2.  Prometheus should be able to validate the metrics server certs and scrape the metrics successfully. 

### Reproducing this issue

_No response_

### KubeBuilder (CLI) Version

4.4.0

### PROJECT version

3

### Plugin versions

```yaml
go.kubebuilder.io/v4
```

### Other versions

_No response_

### Extra Labels

_No response_

abhishekdwivedi3060:
Fixes #4533 

- Fixed `config/prometheus/monitor_tls_patch.yaml` fiel
- Added `serverName` field in `monitor_tls_patch.yaml` for metric server cert validation

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
@@ -75,6 +75,17 @@ replacements:
          delimiter: '.'
          index: 0
          create: true
+     - select:
+         kind: ServiceMonitor
+         group: monitoring.coreos.com
+         version: v1
+         name: controller-manager-metrics-monitor
+       fieldPaths:
+         - spec.endpoints.0.tlsConfig.serverName
+       options:
+         delimiter: '.'
+         index: 0
+         create: true
 
  - source:
      kind: Service
@@ -90,6 +101,17 @@ replacements:
        fieldPaths:
          - spec.dnsNames.0
          - spec.dnsNames.1
+       options:
+         delimiter: '.'
+         index: 1
+         create: true
+     - select:
+         kind: ServiceMonitor
+         group: monitoring.coreos.com
+         version: v1
+         name: controller-manager-metrics-monitor
+       fieldPaths:
+         - spec.endpoints.0.tlsConfig.serverName
        options:
          delimiter: '.'
          index: 1
--- a/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/prometheus/monitor_tls_patch.go
+++ b/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/prometheus/monitor_tls_patch.go
@@ -44,24 +44,21 @@ func (f *ServiceMonitorPatch) SetTemplateDefaults() error {
 
 const serviceMonitorPatchTemplate = `# Patch for Prometheus ServiceMonitor to enable secure TLS configuration
 # using certificates managed by cert-manager
-apiVersion: monitoring.coreos.com/v1
-kind: ServiceMonitor
-metadata:
-  name: controller-manager-metrics-monitor
-  namespace: system
-spec:
-  endpoints:
-    - tlsConfig:
-        insecureSkipVerify: false
-        ca:
-          secret:
-            name: metrics-server-cert
-            key: ca.crt
-        cert:
-          secret:
-            name: metrics-server-cert
-            key: tls.crt
-        keySecret:
-          name: metrics-server-cert
-          key: tls.key
+- op: replace
+  path: /spec/endpoints/0/tlsConfig
+  value:
+    # SERVICE_NAME and SERVICE_NAMESPACE will be substituted by kustomize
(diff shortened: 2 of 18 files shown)


docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
```diff
@@ -95,6 +95,42 @@ replacements:
...
+         kind: ServiceMonitor
+         group: monitoring.coreos.com
+         version: v1
+         name: controller-manager-metrics-monitor
+       fieldPaths:
+         - spec.endpoints.0.tlsConfig.serverName
+       options:
+         delimiter: '.'
+         index: 1
+         create: true
```
camilamacedo86, line 132: We have the service already: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml#L59-L97

Could we not do the same without add more blocks on the default kustomization?

abhishekdwivedi3060: Done

docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
```diff
@@ -94,6 +105,17 @@ replacements:
          delimiter: '.'
          index: 1
          create: true
+     - select:
+         kind: ServiceMonitor
+         group: monitoring.coreos.com
+         version: v1
+         name: controller-manager-metrics-monitor
```
camilamacedo86, line 112: If we will add the fixed name here
Could we not add the fixed name in the PATCH since it is always the same?

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/prometheus/monitor_tls_patch.go
```diff
@@ -44,24 +44,21 @@ func (f *ServiceMonitorPatch) SetTemplateDefaults() error {
...
-            name: metrics-server-cert
-            key: tls.crt
-        keySecret:
-          name: metrics-server-cert
-          key: tls.key
+- op: replace
+  path: /spec/endpoints/0/tlsConfig
+  value:
+    # SERVICE_NAME and SERVICE_NAMESPACE will be substituted by kustomize
+    serverName: SERVICE_NAME.SERVICE_NAMESPACE.svc
```
camilamacedo86, line 51: Did we only need to change the kustomize default file just to pass the serviceName right?
But that is fixed. It will always be the same. 
Could we not simplify the changes and the default kustomize by adding here the name

```suggestion
    serverName: controller-manager-metrics-monitor.SERVICE_NAMESPACE.svc
```

Instead?

camilamacedo86: I am trying to see if we can avoid do changes at: [config/default/kustomization.yaml](https://github.com/kubernetes-sigs/kubebuilder/pull/4536/files#diff-82cfd4563abc54cb08f9a19275ffbb11c8ffbf1410304014e722c2685327300d)

abhishekdwivedi3060: I tried to follow the existing convention followed in kubebuilder kustomize files.
For other scenarios where the service name is referred, we follow similar convention of replacements. See [certificate](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/certmanager/certificate_metrics.go#L62)
I'm not sure if this `serverName` replacement is any different. Let me know if my understanding is wrong

camilamacedo86: I am not too happy with the complexities in the kustomize but I think that is not part of the scope and its make sense. Thank you.

PR #4538, opened 2025-02-03
🌱 ci: fix testdata - issues caused by some merge conflict with go mod

camilamacedo86:


camilamacedo86:
Moving forward to fix the testdata.
I will do a follow up to fix the new check for YMAL
We need to revert the change.

PR #4540, opened 2025-02-04
:seedling: Add util which is able to merge golang source code fragments based custom rules

dmvolod:
This is small util library which as able to merge golang source code fragments based custom rules
Custom rules is a special function who decide how and when merge similar code fragment.
Not all cases are currently handled, but need to decide is it suitable approach for current task or not. 

Preview for #4146

/hold

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,7 @@ require (
 )
 
 require (
+	github.com/dave/dst v0.27.3 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/go-logr/logr v1.4.2 // indirect
 	github.com/go-task/slim-sprig/v3 v3.0.0 // indirect
--- a/pkg/plugin/util/merge/merge.go
+++ b/pkg/plugin/util/merge/merge.go
@@ -1,0 +1,150 @@
+package merge
+
+import (
+	"go/ast"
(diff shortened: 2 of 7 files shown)


go.mod
```diff
@@ -18,6 +18,7 @@ require (
 )
 
 require (
+	github.com/dave/dst v0.27.3 // indirect
```
camilamacedo86, line 21: We need to be very consciouss about the deps added. 
I do not think that we can move within

pkg/plugin/util/merge/merge.go
```diff
@@ -0,0 +1,150 @@
...
+	"reflect"
+	"strconv"
+	"strings"
+
+	"github.com/dave/dst"
+	"github.com/dave/dst/decorator"
+	"golang.org/x/tools/go/ast/astutil"
+)
+
+func InsertCode(filename, code string) error {
```
camilamacedo86, line 17: Let's first check the approach. I do not want you to spend an effort that will not be constructive.
However, as you know, in the plugin/utils, we have a helper to insert code, replace, etc.
So, we would need to use what we have already.

dmvolod: Good point
I saw this code, but looks like it's just manipulate code as text, but this is manipulation code as code. 
But I understand, its enough for most cases, probably except the current one or similar few cases.

Issue #4546, opened 2025-02-07
docker-build should use --load to accomodate newer docker build aliased to buildx


kaovilai:
### What broke? What's expected?

Currently `make docker-build` can error if docker cli has aliased docker build to docker buildx without using `--load`
```
make docker-build
docker build -t quay.io/konveyor/oadp-non-admin:latest .
[+] Building 0.4s (17/17) FINISHED                                                                                                                                                                                                         docker-container:default
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                           0.0s
 => => transferring dockerfile: 1.34kB                                                                                                                                                                                                                         0.0s
 => [internal] load metadata for docker.io/library/golang:1.22                                                                                                                                                                                                 0.3s
 => [internal] load metadata for gcr.io/distroless/static:nonroot                                                                                                                                                                                              0.3s
 => [auth] library/golang:pull token for registry-1.docker.io                                                                                                                                                                                                  0.0s
 => [internal] load .dockerignore                                                                                                                                                                                                                              0.0s
 => => transferring context: 160B                                                                                                                                                                                                                              0.0s
 => [builder 1/9] FROM docker.io/library/golang:1.22@sha256:1cf6c45ba39db9fd6db16922041d074a63c935556a05c5ccb62d181034df7f02                                                                                                                                   0.0s
 => => resolve docker.io/library/golang:1.22@sha256:1cf6c45ba39db9fd6db16922041d074a63c935556a05c5ccb62d181034df7f02                                                                                                                                           0.0s
 => [internal] load build context                                                                                                                                                                                                                              0.0s
 => => transferring context: 124.58kB                                                                                                                                                                                                                          0.0s
 => [stage-1 1/3] FROM gcr.io/distroless/static:nonroot@sha256:6ec5aa99dc335666e79dc64e4a6c8b89c33a543a1967f20d360922a80dd21f02                                                                                                                                0.0s
 => => resolve gcr.io/distroless/static:nonroot@sha256:6ec5aa99dc335666e79dc64e4a6c8b89c33a543a1967f20d360922a80dd21f02                                                                                                                                        0.0s
 => CACHED [builder 2/9] WORKDIR /workspace                                                                                                                                                                                                                    0.0s
 => CACHED [builder 3/9] COPY go.mod go.mod                                                                                                                                                                                                                    0.0s
 => CACHED [builder 4/9] COPY go.sum go.sum                                                                                                                                                                                                                    0.0s
[...]

PR #4545, opened 2025-02-07
:bug: v4 template docker-buildx do not add duplicate --platform

kaovilai:
Fixes #4544

being PR'd also into https://github.com/migtools/oadp-non-admin/pull/182

related: https://github.com/kubernetes-sigs/kubebuilder/pull/4547

PR #4547, opened 2025-02-07
Makefile: docker-build should use --load for newer docker build aliased to buildx

kaovilai:
…ed to buildx

Fixes https://github.com/kubernetes-sigs/kubebuilder/issues/4546

Same as https://github.com/migtools/oadp-non-admin/pull/180

Related: https://github.com/kubernetes-sigs/kubebuilder/pull/4545

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -183,7 +183,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
 docker-build: ## Build docker image with the manager.
-	$(CONTAINER_TOOL) build -t ${IMG} .
+	$(CONTAINER_TOOL) build --load -t ${IMG} .
 
 .PHONY: docker-push
 docker-push: ## Push docker image with the manager.


camilamacedo86:
Hi @kaovilai 👋  

Really appreciate you taking the time to improve **Kubebuilder**! 🚀  

I hope you don’t mind, but I’ll **close this PR** with the following reference:  
[Issue #4546 - Comment](https://github.com/kubernetes-sigs/kubebuilder/issues/4546#issuecomment-2643418145).  

If you **disagree** and still think we should add this change,  
please feel free to **reopen** the PR for further discussion.  

Thanks again for your contribution! 😊

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -183,7 +183,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
 docker-build: ## Build docker image with the manager.
-	$(CONTAINER_TOOL) build -t ${IMG} .
+	$(CONTAINER_TOOL) build --load -t ${IMG} .
```
camilamacedo86, line 186: It is the default docker build it is working fine.
you probably want replace the `--flag push` with `--load` in the docker buildX instead, see: https://github.com/kubernetes-sigs/kubebuilder/pull/4547/files#diff-e5d6d4f9d67174f1f3764674eb786688353c77cef8514e24d1323e55e5a4a7e8R192-R209

However, the WARNING is only faced in the buildx(above target) when it has no flags `--load` or `--push`

kaovilai: docker buildx do not require --load at all.. if you have --push, you can skip --load.

camilamacedo86: Hi @kaovilai 

See: https://github.com/kubernetes-sigs/kubebuilder/issues/4546#issuecomment-2644025655

I reproduce your scenario:

You are calling `docker buildx` without ANY of the flags.
It is required to have --load or push

And this target here is docker build ( should not use any of those )
You can either validate the behaviour with the samples under testdata to check that all generated by Kubebuilder works fine.

So, I am closing this one as deferred if you do not mind

camilamacedo86:
Closing this based on the comment : https://github.com/kubernetes-sigs/kubebuilder/issues/4546#issuecomment-2644025655.

We should not do this change. 
But thank you a lot for raising it and trying to contribute to the project.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
kaovilai: @camilamacedo86 per https://github.com/kubernetes-sigs/kubebuilder/issues/4546#issuecomment-2644142851 this can be triggered with just docker-build per my reproduce. I did not modify the default scaffold.

kaovilai: This is one of the PRs that benefits the users of the **default scaffold without modification.**

PR #4556, opened 2025-02-10
🌱 Revert (kustomize/v2): fix ServiceMonitor with TLS kustomize scaffolding

camilamacedo86:
Reverts kubernetes-sigs/kubebuilder#4536

Those changes broke the project: Example: https://github.com/kubernetes-sigs/kubebuilder/actions/runs/13250335077/job/36986568474?pr=4555

It is strange but seems that the tests were not executed and the PR merged without them

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/prometheus/monitor_tls_patch.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/prometheus/monitor_tls_patch.yaml
@@ -1,19 +1,22 @@
 # Patch for Prometheus ServiceMonitor to enable secure TLS configuration
 # using certificates managed by cert-manager
-- op: replace
-  path: /spec/endpoints/0/tlsConfig
-  value:
-    # SERVICE_NAME and SERVICE_NAMESPACE will be substituted by kustomize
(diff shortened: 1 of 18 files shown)


camilamacedo86:
Hi @ashishranjan738 

Sorry, I do not understand how it gets merged without passing in the CI.
I will need to validate the config. But those changes broke the e2e tests, see: https://github.com/kubernetes-sigs/kubebuilder/pull/4555

I could validate that by reverting all returns to work,

docs/book/src/cronjob-tutorial/testdata/project/config/prometheus/monitor_tls_patch.yaml
```diff
@@ -1,19 +1,22 @@
...
+          secret:
+            name: metrics-server-cert
+            key: ca.crt
+        cert:
+          secret:
+            name: metrics-server-cert
+            key: tls.crt
+        keySecret:
+          name: metrics-server-cert
+          key: tls.key
```
camilamacedo86, line 22: When I came back to look at this one, I saw that this file is the PATCH (monitor_tls_patch.yaml) So,here we need to use kustomize operations instead.

PR #4559, opened 2025-02-11
🐛 Fix kubebuild alpha generate when not in GOPATH

freddyesteban:
Ran to an issue when running `kubebuilder alpha generate`, and since my project was not on the GOPATH, there was no implicit call to `go mod init`.

>  If your project is initialized within GOPATH, the implicitly called go mod init will interpolate the module path for you. Otherwise --repo=<module path> must be set.

This change updates `alpha generate` command so that `--repo` is always included when running kubebuilder init to fix issue when not on GOPATH. The value for `repo` is retrieved from the PROJECT file as other kubebuilder init options.


camilamacedo86 (commented):
/lgtm
/ok-to-test

camilamacedo86 (approved):

Issue #4566, opened 2025-02-14
(helm/v1-alpha) Nameprefix is missing in editor/viewer ClusterRoles after scaffolding


mkarlheim:
### What do you want to happen?

Hello,

I'm currently in the process moving the distribution of some Kubernetes operators from Kustomize to Helm. I used the plugin helm/v1-alpha for that and it worked really well so far. 

The only thing I'm missing is the namePrefix for the ClusterRoles in the rbac subfolder.

In Kustomize I was able to define a namePrefix in "config/default/kustomization.yaml":
```yaml
# Adds namespace to all resources.
namespace: my-operator-system

# Value of this field is prepended to the
# names of all resources, e.g. a deployment named
# "wordpress" becomes "alices-wordpress".
# Note that it should also match with the prefix (text before '-') of the namespace
# field above.
namePrefix: name-prefix-

# Labels to add to all resources and selectors.
```

This lead to generated ClusterRoles with a namePrefix. This was really handy because this helps to identify the ClusterRoles of the corresponding Kubernetes operator quickly.

There is already a special treatment for some of the roles [here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go#L278): 

```go
// Apply RBAC-specific replacements
	if subDir == "rbac" {
		contentStr = strings.Replace(contentStr,
			"name: controller-manager",
			"name: {{ .Values.controllerManager.serviceAccountName }}", -1)
		contentStr = strings.Replace(contentStr,
			"name: metrics-reader",
			fmt.Sprintf("name: %s-metrics-reader", projectName), 1)

		contentStr = strings.Replace(contentStr,
			"name: metrics-auth-role",
			fmt.Sprintf("name: %s-metrics-auth-role", projectName), -1)
		contentStr = strings.Replace(contentStr,
			"name: metrics-auth-rolebinding",
			fmt.Sprintf("name: %s-metrics-auth-rolebinding", projectName), 1)

		if strings.Contains(contentStr, ".Values.controllerManager.serviceAccountName") &&
			strings.Contains(contentStr, "kind: ServiceAccount") &&
			!strings.Contains(contentStr, "RoleBinding") {
			// The generated Service Account does not have the annotations field so we must add it.
			contentStr = strings.Replace(contentStr,
				"metadata:", `metadata:
[...]

PR #4571, opened 2025-02-20
🐛 (helm/v1-alpha): Use namePrefix from config/default/kustomization.yaml as prefix for RBAC rules and project name only if this value cannot be found.

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/4566

### DIFF
--- a/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
+++ b/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
@@ -124,6 +124,29 @@ func (s *initScaffolder) Scaffold() error {
 	return nil
 }
 
+// getNamePrefix will return the value from kustomize config so that we can append
+// in the RBAC rules manifests. If we be unable to find this value we will use
+// the projectName instead.
+func (s *initScaffolder) getNamePrefix() string {
+	filePath := "config/default/kustomization.yaml"
+	content, err := os.ReadFile(filePath)
+	if err != nil {
+		log.Fatalf("failed to read config/default/kustomization.yaml: %s", err)
+	}
+
+	var defaultConfig struct {
+		NamePrefix string `yaml:"namePrefix"`
+	}
+
+	if err := yaml.Unmarshal(content, &defaultConfig); err != nil {
+		log.Warnf("failed to parse kustomization.yaml to get namePrefix: %s", err)
+		log.Warnf("using the project name as a prefix of RBAC and manifests")
+		return s.config.GetProjectName()
+	}
+
+	return strings.TrimSpace(defaultConfig.NamePrefix)
+}
+
 // getDeployImagesEnvVars will return the values to append the envvars for projects
 // which has the APIs scaffolded with DeployImage plugin
 func (s *initScaffolder) getDeployImagesEnvVars() map[string]string {
@@ -243,7 +266,7 @@ func (s *initScaffolder) copyConfigFiles() error {
 
 		for _, srcFile := range files {
 			destFile := filepath.Join(dir.DestDir, filepath.Base(srcFile))
-			err := copyFileWithHelmLogic(srcFile, destFile, dir.SubDir, s.config.GetProjectName())
+			err = copyFileWithHelmLogic(srcFile, destFile, dir.SubDir, s.getNamePrefix())
 			if err != nil {
 				return err
 			}
@@ -255,7 +278,7 @@ func (s *initScaffolder) copyConfigFiles() error {
 
 // copyFileWithHelmLogic reads the source file, modifies the content for Helm, applies patches
 // to spec.conversion if applicable, and writes it to the destination
-func copyFileWithHelmLogic(srcFile, destFile, subDir, projectName string) error {
+func copyFileWithHelmLogic(srcFile, destFile, subDir, namePrefix string) error {
 	if _, err := os.Stat(srcFile); os.IsNotExist(err) {
 		log.Printf("Source file does not exist: %s", srcFile)
 		return err
@@ -282,14 +305,14 @@ func copyFileWithHelmLogic(srcFile, destFile, subDir, projectName string) error 
 			"name: {{ .Values.controllerManager.serviceAccountName }}", -1)
 		contentStr = strings.Replace(contentStr,
 			"name: metrics-reader",
-			fmt.Sprintf("name: %s-metrics-reader", projectName), 1)
+			fmt.Sprintf("name: %smetrics-reader", namePrefix), 1)
 
 		contentStr = strings.Replace(contentStr,
 			"name: metrics-auth-role",
-			fmt.Sprintf("name: %s-metrics-auth-role", projectName), -1)
+			fmt.Sprintf("name: %smetrics-auth-role", namePrefix), -1)
 		contentStr = strings.Replace(contentStr,
 			"name: metrics-auth-rolebinding",
-			fmt.Sprintf("name: %s-metrics-auth-rolebinding", projectName), 1)
+			fmt.Sprintf("name: %smetrics-auth-rolebinding", namePrefix), 1)
 
 		if strings.Contains(contentStr, ".Values.controllerManager.serviceAccountName") &&
 			strings.Contains(contentStr, "kind: ServiceAccount") &&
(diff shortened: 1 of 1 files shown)


camilamacedo86:
HI @mkarlheim 

Could you please check this one?

pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
```diff
@@ -255,7 +278,7 @@ func (s *initScaffolder) copyConfigFiles() error {
 
 // copyFileWithHelmLogic reads the source file, modifies the content for Helm, applies patches
 // to spec.conversion if applicable, and writes it to the destination
-func copyFileWithHelmLogic(srcFile, destFile, subDir, projectName string) error {
+func copyFileWithHelmLogic(srcFile, destFile, subDir, namePrefix string) error {
```
camilamacedo86, line 281: HI @mkarlheim

Are there any other manifests that we should append this prefix and that we are missing?

mkarlheim:
> HI @mkarlheim
> 
> Could you please check this one?

Awesome, thanks! I'll try to do it shortly.

mkarlheim:
Hi @camilamacedo86, 

using the namePrefix by default and the project name as fallback is an improvement, but the core issue still exists in my opinion:
Executing `kubebuilder create api --group testgroup --version v1 --kind Test` leads to newly generated ClusterRoles named 
`name: testgroup-test-admin-role`,

pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
```diff
@@ -282,14 +305,14 @@ func copyFileWithHelmLogic(srcFile, destFile, subDir, projectName string) error
 			"name: {{ .Values.controllerManager.serviceAccountName }}", -1)
 		contentStr = strings.Replace(contentStr,
 			"name: metrics-reader",
```
mkarlheim, line 307: No ClusterRoles with name `-editor-role`, `-admin-role`, `-viewer-role` are considered here

pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
```diff
@@ -124,6 +124,29 @@ func (s *initScaffolder) Scaffold() error {
...
+	}
+
+	var defaultConfig struct {
+		NamePrefix string `yaml:"namePrefix"`
+	}
+
+	if err := yaml.Unmarshal(content, &defaultConfig); err != nil {
+		log.Warnf("failed to parse kustomization.yaml to get namePrefix: %s", err)
+		log.Warnf("using the project name as a prefix of RBAC and manifests")
+		return s.config.GetProjectName()
```
mkarlheim, line 144: the projectName name needs a trailing `-` as it got removed in the copyFileWithHelmLogic function:

```suggestion
		return fmt.Sprintf("%s-", s.config.GetProjectName()), -1)
```

mkarlheim (changes requested):

pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
mkarlheim: The `kubebuilder create api --group testgroup --version v1 --kind Test` was only the derivation of where the ClusterRoles come from and to point out, that the names of the ClusterRoles can differ depending on the names of api and apigroups.

pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
camilamacedo86: I will test it out

pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
camilamacedo86: I am not sure if I can follow it. Could you please add the steps to reproduce and what you expected and what are you seen instead?

camilamacedo86: OH. I see what do you mean, I think we cannot move forward with this PR at all, 
I also missed the boat here. 
The prefix-name will change the name inside of the manifests not the name of the files. 

/hold

camilamacedo86: Oh, I see what you mean. The view, edit, and admin rules created for the project do not use the prefix—they aren’t generated via Kustomize.

I definitely need to review this because we shouldn’t introduce something in the Helm plugin that doesn’t align with the existing behavior.

The goal of the Helm plugin is to package the solution, not to introduce deviations from what is already done with Kubebuilder/Kustomize. What is generated by Helm should match what is generated with Kustomize.

If we need to make changes, they should be applied at the project level, rather than having Helm introduce behavior that diverges from the Kustomize output.

So I am adding a hold here to properly check it as well


/hold

camilamacedo86: Thinking better the right approach here is have the prefixName in the values and then use it in the manifests like we have with the kustomize I will need also think more about.

camilamacedo86:
Closing because I think we will need to fix it another way, by changing the scaffolds and adding a prefix in the chart values instead.

camilamacedo86:
Hi @mkarlheim

Can you please verify this one now?
It seems that it addresses your need for now.

However, ultimately, we will need to change the Helm plugin implementation to generate the chart from the kustomize install, using YAML instead.

mkarlheim:
Hi @camilamacedo86, 

many thanks for your efforts! 

Unfortunately, it is still not the expected behaviour. All roles that are created with the Kustomize/v2 plugin, don't have a prefix in their generated yaml by default. 

All the roles are prefixed in the moment when the Kustomize build (make build-installer) is executed as you described in https://github.com/kubernetes-sigs/kubebuilder/issues/4833.

The function ```copyFileWithHelmLogic``` only considers the roles with fixed names like the metrics, election and leader roles, but not the crd-roles (admin, editor, viewer). The name of these roles is not fixed as it contains resource group and kind (in multgroup projects).

I was also thinking about using the Kustomize build output as the base for the Helm chart generation, as this would adress the issue with the namePrefix.

Another argument for your proposal would be:
In the ```config/rbac/kustomization.yaml``` there is a comment saying `You can comment the following lines
# if you do not want those helpers be installed with your Project.` Those types of changes wouldn't be considered as well, if someone commented out some roles (e.g.). Using the Kustomize build output as base for the Helm chart generation would also address those types of changes.

Best
Michael

mkarlheim (commented):
Hi @camilamacedo86,

many thanks for your efforts!

Unfortunately, it is still not the expected behaviour. All roles that are created with the Kustomize/v2 plugin, don't have a prefix in their generated yaml by default.

All the roles are prefixed in the moment when the Kustomize build (make build-installer) is executed as you described in https://github.com/kubernetes-sigs/kubebuilder/issues/4833.

The function copyFileWithHelmLogic only considers the roles with fixed names like the metrics, election and leader roles, but not the crd-roles (admin, editor, viewer). The name of these roles is not fixed as it contains resource group and kind (in multgroup projects).

I was also thinking about using the Kustomize build output as the base for the Helm chart generation, as this would adress the issue with the namePrefix.

Another argument for your proposal would be:
In the config/rbac/kustomization.yaml there is a comment saying You can comment the following lines if you do not want those helpers be installed with your Project. Those types of changes wouldn't be considered as well, if someone commented out some roles (e.g.). Using the Kustomize build output as base for the Helm chart generation would also address those types of changes.

Best
Michael

camilamacedo86:
Hi @mkarlheim 

Thank you a lot for your time on this review 👍 

> Unfortunately, it is still not the expected behaviour. All roles that are created with the Kustomize/v2 plugin, don't have a prefix in their generated yaml by default.

But what we are doing here is ONLY doing what is done with kutomize
We will not change the files name. 
The prefix ONLY change the prefix of the resource/RBAC itself ( metadata.name )

I think we need to modify one of the samples scenarios and add a test for it. 
So that we can automatically validate and ensure that the result with helm is the same.

> The function copyFileWithHelmLogic only considers the roles with fixed names like the metrics, election and leader roles, but not the crd-roles (admin, editor, viewer). The name of these roles is not fixed as it contains resource group and kind (in multgroup projects).

OK, we need to add a code to cover those 👍 

> In the config/rbac/kustomization.yaml there is a comment saying You can comment the following lines if you do not want those helpers be installed with your Project. Those types of changes wouldn't be considered as well.

OK. Good point for we recheck 👍 

> I was also thinking about using the Kustomize build output as the base for the Helm chart generation, as this would adress the issue with the namePrefix.

That seems the best way to move forward with the helm plugin.


/hold until check the points above

camilamacedo86:
I am closing this one as defered we need review the helm plugin

Issue #4433, opened 2024-12-17
kubebuilder alpha generate fails to rescaffold if PROJECT contains go.kubebuilder.io/v3 


lack:
### What broke? What's expected?

I'm trying to upgrade an old project to go/v4 via kubebuilder.

Here's what I tried:
```bash
kubebuilder alpha generate --plugins "go.kubebuilder.io/v4"
```

Here's what I got:
```
INFO kubebuilder init:
$ kubebuilder init --plugins go.kubebuilder.io/v3 --domain eapol.openshift.io 
Error: no plugin could be resolved with key "go.kubebuilder.io/v3"
Usage:
  kubebuilder [flags]

Examples:
The first step is to initialize your project:
    kubebuilder init [--plugins=<PLUGIN KEYS> [--project-version=<PROJECT VERSION>]]

<PLUGIN KEYS> is a comma-separated list of plugin keys from the following table
and <PROJECT VERSION> a supported project version for these plugins.

                             Plugin keys | Supported project versions
-----------------------------------------+----------------------------
               base.go.kubebuilder.io/v4 |                          3
 deploy-image.go.kubebuilder.io/v1-alpha |                          3
                    go.kubebuilder.io/v4 |                          3
         grafana.kubebuilder.io/v1-alpha |                          3
      kustomize.common.kubebuilder.io/v2 |                          3

For more specific help for the init command of a certain plugins and project version
configuration please run:
    kubebuilder init --help --plugins=<PLUGIN KEYS> [--project-version=<PROJECT VERSION>]

Default plugin keys: "go.kubebuilder.io/v4"
Default project version: "3"


Flags:
  -h, --help                     help for kubebuilder
      --plugins strings          plugin keys to be used for this subcommand execution
      --project-version string   project version (default "3")

FATA no plugin could be resolved with key "go.kubebuilder.io/v3" 
FATA Failed to command exit status 1
```

### Reproducing this issue

Originating PROJECT file contained

```yaml
layout:
- go.kubebuilder.io/v3
```

### KubeBuilder (CLI) Version

Version: main.version{KubeBuilderVersion:"4.3.1", KubernetesVendor:"1.31.0", GitCommit:"a9ee3909f7686902879bd666b92deec4718d92c9", BuildDate:"2024-11-09T12:30:22Z", GoOs:"linux", GoArch:"amd64"}  

### PROJECT version

3

### Plugin versions

```yaml
plugins:
[...]

PR #4572, opened 2025-02-21
🐛 Allow 'kubebuilder alpha generate' rescaffolds work with no longer supported/available plugins ( go/v3 and go/v2 )

sarthaksarthak9:
Fixes: #4433 

- Fixed issue where 'kubebuilder alpha generate' failed if PROJECT contained go.kubebuilder.io/v3.  
- Automatically updates outdated plugin references (v3 → v4) to support scaffolding.  
- Added a warning to notify users before modifying the PROJECT file. 

image
<br>
image
<br>
image

### DIFF
--- a/pkg/cli/alpha/internal/generate.go
+++ b/pkg/cli/alpha/internal/generate.go
@@ -256,6 +256,24 @@ func getInputPath(inputPath string) (string, error) {
 func getInitArgs(store store.Store) []string {
 	var args []string
 	plugins := store.Config().GetPluginChain()
+
+	// Define outdated plugin versions that need replacement
+	outdatedPlugins := map[string]string{
+		"go.kubebuilder.io/v3":       "go.kubebuilder.io/v4",
+		"go.kubebuilder.io/v3-alpha": "go.kubebuilder.io/v4",
+		"go.kubebuilder.io/v2":       "go.kubebuilder.io/v4",
+	}
+
+	// Replace outdated plugins and exit after the first replacement
+	for i, plugin := range plugins {
+		if newPlugin, exists := outdatedPlugins[plugin]; exists {
+			log.Warnf("We checked that your PROJECT file is configured with the layout '%s', which is no longer supported.\n"+
+				"However, we will try our best to re-generate the project using '%s'.", plugin, newPlugin)
+			plugins[i] = newPlugin
+			break
+		}
+	}
+
 	if len(plugins) > 0 {
 		args = append(args, "--plugins", strings.Join(plugins, ","))
 	}
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -147,6 +147,14 @@ func (c *CLI) buildCmd() error {
 
 	var uve config.UnsupportedVersionError
 
+	// Workaround for kubebuilder alpha generate
+	if len(os.Args) > 2 && os.Args[1] == "alpha" && os.Args[2] == "generate" {
+		err := updateProjectFileForAlphaGenerate()
+		if err != nil {
+			return fmt.Errorf("failed to update PROJECT file: %w", err)
+		}
+	}
+
 	// Get project version and plugin keys.
 	switch err := c.getInfo(); {
 	case err == nil:
--- a/pkg/cli/cmd_helpers.go
+++ b/pkg/cli/cmd_helpers.go
@@ -20,7 +20,9 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
 	"sigs.k8s.io/kubebuilder/v4/pkg/config"
@@ -335,4 +337,41 @@ func (factory *executionHooksFactory) postRunEFunc() func(*cobra.Command, []stri
 
 		return nil
 	}
+}
+
+func updateProjectFileForAlphaGenerate() error {
+	projectFilePath := "PROJECT"
+
+	content, err := os.ReadFile(projectFilePath)
+	if err != nil {
+		return fmt.Errorf("failed to read PROJECT file: %w", err)
+	}
+
+	projectStr := string(content)
+
+	// Define outdated plugin versions that need replacement
+	outdatedPlugins := []string{"go.kubebuilder.io/v3", "go.kubebuilder.io/v3-alpha", "go.kubebuilder.io/v2"}
+	updated := false
+
+	for _, oldPlugin := range outdatedPlugins {
+		if strings.Contains(projectStr, oldPlugin) {
+			log.Warnf("Detected '%s' in PROJECT file.", oldPlugin)
+			log.Warnf("Kubebuilder v4 no longer supports this. It will be replaced with 'go.kubebuilder.io/v4'.")
+
+			projectStr = strings.ReplaceAll(projectStr, oldPlugin, "go.kubebuilder.io/v4")
+			updated = true
+			break
+		}
+	}
+
+	// Only update the file if changes were made
+	if updated {
+		err = os.WriteFile(projectFilePath, []byte(projectStr), 0644)
+		if err != nil {
+			return fmt.Errorf("failed to update PROJECT file: %w", err)
+		}
+		log.Infof("PROJECT file updated successfully.")
+	}
+
+	return nil
 }
--- a/.github/workflows/test-alpha-generate.yml
+++ b/.github/workflows/test-alpha-generate.yml
@@ -1,0 +1,40 @@
+name: Test Alpha Generate
+
+on:
+  push:
+    paths:
+      - 'pkg/cli/alpha/**'
+      - '.github/workflows/test-alpha-generate.yml'
+  pull_request:
+    paths:
+      - 'pkg/cli/alpha/**'
+      - '.github/workflows/test-alpha-generate.yml'
+      
+jobs:
+  unsupported:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Install dependencies and generate binary
+        run: make install
+
+      - name: Navigate to testdata/project-v4
+        run: cd testdata/project-v4
+
+      - name: Update PROJECT file
+        run: |
+          sed -i 's#go.kubebuilder.io/v4#go.kubebuilder.io/v3#g' testdata/project-v4/PROJECT
+
+      - name: Run kubebuilder alpha generate
+        run: |
+          cd testdata/project-v4 && kubebuilder alpha generate


pkg/cli/alpha/internal/generate.go
```diff
@@ -256,6 +256,13 @@ func getInputPath(inputPath string) (string, error) {
 func getInitArgs(store store.Store) []string {
 	var args []string
 	plugins := store.Config().GetPluginChain()
+
+	for i, plugin := range plugins {
+		if plugin == "go.kubebuilder.io/v3" {
+			plugins[i] = "go.kubebuilder.io/v4"
```
camilamacedo86, line 262: can we add a break here?
WDYT?

pkg/cli/cmd_helpers.go
```diff
@@ -336,3 +338,47 @@ func (factory *executionHooksFactory) postRunEFunc() func(*cobra.Command, []stri
...
+
+	content, err := os.ReadFile(projectFilePath)
+	if err != nil {
+		return fmt.Errorf("failed to read PROJECT file: %w", err)
+	}
+
+	projectStr := string(content)
+
+	// No need to update if v3 is not found
+	if !strings.Contains(projectStr, "go.kubebuilder.io/v3") {
```
camilamacedo86, line 356: Can we ensure that we do the same if it is any of those:
- go.kubebuilder.io/v3
- go.kubebuilder.io/v3-alpha
- go.kubebuilder.io/v2

I mean, in any case we warning and inform the user that we are doing our best to re-scaffold.
Todo so we will need update the PROJECT file by replacing X with go.kubebuilder.io/v4

pkg/cli/cmd_helpers.go
```diff
@@ -336,3 +338,47 @@ func (factory *executionHooksFactory) postRunEFunc() func(*cobra.Command, []stri
...
+		return nil 
+	}
+
+	log.Warnf("Detected 'go.kubebuilder.io/v3' in PROJECT file.")
+	log.Warnf("Kubebuilder v4 no longer supports this. It will be replaced with 'go.kubebuilder.io/v4'.")
+
+	fmt.Print("Do you want to proceed? (y/N): ")
```
camilamacedo86, line 363: That is very cool 🥇 🚀 
When we scaffold the controllers/API (kubebuilder create API) we asked if we should scaffold both controller + API. So, we have some code utils for that already, Could we try to use those? See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/util/stdin.go#L28-L40

camilamacedo86: Also, to allow us to test out this option with scripts, could we create a flag so that we do not need to ask if the flag is set? --allow-best-effort OR we can use --force true  OR any other term if you have a better idea/suggestion.

camilamacedo86 (changes requested):
Really great job 🥇 🎉 
I added some comments

pkg/cli/cli.go
```diff
@@ -151,21 +151,26 @@ func (c *CLI) buildCmd() error {
...
 		}
 	default:
 		return err
 	}
 
+	// Workaround for kubebuilder alpha generate
+	if c.cmd.CalledAs() == "alpha generate" {
+		err := updateProjectFileForAlphaGenerate()
+		if err != nil {
+			return fmt.Errorf("failed to update PROJECT file: %w", err)
```
camilamacedo86, line 170: We have tests here: https://github.com/kubernetes-sigs/kubebuilder/tree/master/test/e2e/alphagenerate

We can create a new project and add 1 API
Then, update the PROJECT file to replace `go.kubebuilder.io/v4` with go.kubebuilder.io/v2 or go.kubebuilder.io/v3 and then call the command with the flag that --force OR allow upgrade the PROJECT file without the need to say Y so we can ensure that works fine. 

WDYT?

pkg/cli/alpha/internal/generate.go
sarthaksarthak9: ig we can but if there's a possibility that "go.kubebuilder.io/v3" appears multiple times in plugins, then we should not add a break, as we need to update all occurrences.

sarthaksarthak9: If there's a possibility that "go.kubebuilder.io/v3" appears multiple times in plugins, then we should not add a break, as we need to update all occurrences.

camilamacedo86: No, it can appears just once.
We can either check the layout, see: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/PROJECT#L7

sarthaksarthak9:
@camilamacedo86 I noticed that updating the plugin list in generate.go automatically updates the PROJECT file, even though I don’t explicitly see a save function. How is this happening? Could you help me confirm where the file update is triggered?. Also, is there still a need for updateProjectFileForAlphaGenerate() in this case?

pkg/cli/cmd_helpers.go
```diff
@@ -336,3 +338,47 @@ func (factory *executionHooksFactory) postRunEFunc() func(*cobra.Command, []stri
...
+	content, err := os.ReadFile(projectFilePath)
+	if err != nil {
+		return fmt.Errorf("failed to read PROJECT file: %w", err)
+	}
+
+	projectStr := string(content)
+
+	// No need to update if v3 is not found
+	if !strings.Contains(projectStr, "go.kubebuilder.io/v3") {
+		return nil
```
camilamacedo86, line 357: You need fix the lint issues.
You can run locally `make fix-lint`

pkg/cli/cmd_helpers.go
```diff
@@ -336,3 +338,54 @@ func (factory *executionHooksFactory) postRunEFunc() func(*cobra.Command, []stri
...
+			updatedContent = strings.ReplaceAll(updatedContent, oldPlugin, "go.kubebuilder.io/v4")
+			updated = true
+		}
+	}
+
+	// If no changes were made, exit early
+	if !updated {
+		fmt.Println("No changes needed in PROJECT file.")
+		return nil
+	}
```
camilamacedo86, line 381: I think we do not need it right?
We only want inform when that fails in the scenario
so we can return here directly

pkg/cli/cmd_helpers.go
```diff
@@ -336,3 +338,54 @@ func (factory *executionHooksFactory) postRunEFunc() func(*cobra.Command, []stri
...
+				fmt.Print("Do you want to proceed? (y/n): ")
+				reader := bufio.NewReader(os.Stdin)
+				if !util.YesNo(reader) {
+					fmt.Println("Operation cancelled.")
+					return nil
+				}
+			}
+
+			updatedContent = strings.ReplaceAll(updatedContent, oldPlugin, "go.kubebuilder.io/v4")
+			updated = true
```
camilamacedo86, line 373: we can add a break here or update the project file and return already.
We will have only 1 plugin/layout in this scenario

pkg/cli/cli.go
```diff
@@ -151,21 +151,28 @@ func (c *CLI) buildCmd() error {
 	switch err := c.getInfo(); {
 	case err == nil:
 	case errors.As(err, &uve) && uve.Version.Compare(config.Version{Number: 3, Stage: stage.Alpha}) == 0:
-		// Check if the corresponding stable version exists, set c.projectVersion and break
```
camilamacedo86, line 154: I think we can keep the previous comments right?

pkg/cli/cmd_helpers.go
```diff
@@ -336,3 +338,54 @@ func (factory *executionHooksFactory) postRunEFunc() func(*cobra.Command, []stri
...
+
+	// Write changes to the PROJECT file
+	err = os.WriteFile(projectFilePath, []byte(updatedContent), 0644)
+	if err != nil {
+		return fmt.Errorf("failed to write updated PROJECT file: %w", err)
+	}
+
+	fmt.Println("PROJECT file updated successfully.")
+	return nil
+}
```
camilamacedo86, line 391: Could you test it manually?
How to test:
- run make install with those changes to generate the kubebuilder bin locallu
- go to the project-v4 sample under the testdata
- update the PROEJCT file by replacing the layout go.kubebuilder.io/v4 with go.kubebuilder.io/v3 
- run the command kubebuilder alpha generate from the dir.

If that works fine, you can re-run again without update all should work without ask about update the PROJECT file and etc.

pkg/cli/cli.go
```diff
@@ -151,21 +151,26 @@ func (c *CLI) buildCmd() error {
...
-			// stable version not registered, let's bail out
 			return err
 		}
 	default:
 		return err
 	}
 
+	// Workaround for kubebuilder alpha generate
+	if c.cmd.CalledAs() == "alpha generate" {
+		err := updateProjectFileForAlphaGenerate()
```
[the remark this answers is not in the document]
sarthaksarthak9: I previously tested the changes and observed that modifying the plugin list in getInitArgs() automatically updates the PROJECT file. Any plugin version other than v4 is replaced with go.kubebuilder.io/v4.

If getInitArgs() already ensures outdated plugins are updated before execution, then updateProjectFileForAlphaGenerate() may no longer be necessary. The PROJECT file is updated naturally when kubebuilder commands run.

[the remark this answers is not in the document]
camilamacedo86: >If getInitArgs() already ensures outdated plugins are updated before execution

that would be for sure a better solution 👍 

We just warning something like "We checked that your PROJECT file is configured with the layout: which is no longer supported. However, we will try to do our best effort to re-generated the project"

And that is done 🚀

pkg/cli/alpha/internal/generate.go
```diff
@@ -256,6 +256,24 @@ func getInputPath(inputPath string) (string, error) {
...
+
+	// Replace outdated plugins and exit after the first replacement
+	for i, plugin := range plugins {
+		if newPlugin, exists := outdatedPlugins[plugin]; exists {
+			log.Warnf("We checked that your PROJECT file is configured with the layout '%s', which is no longer supported.\n"+
+				"However, we will try our best to re-generate the project using '%s'.", plugin, newPlugin)
+			plugins[i] = newPlugin
+			break
+		}
+	}
```
camilamacedo86, line 275: See that if we only change here it will not work.
Following the test done locally with this change

- I run `make install` to have the binary with this code source
- I go to  `cd testdata/project-v4` 
- I updated the PROJECT file : 
```sh
 cat PROJECT 
domain: testproject.org
layout:
- go.kubebuilder.io/v3
projectName: project-v4
```
- Then, I run the command

```sh
kubebuilder alpha generate
Error: no plugin could be resolved with key "go.kubebuilder.io/v3" for project version "3"
Usage:
  kubebuilder [flags]

Examples:
The first step is to initialize your project:
    kubebuilder init [--plugins=<PLUGIN KEYS> [--project-version=<PROJECT VERSION>]]

<PLUGIN KEYS> is a comma-separated list of plugin keys from the following table
and <PROJECT VERSION> a supported project version for these plugins.

                             Plugin keys | Supported project versions
-----------------------------------------+----------------------------
               base.go.kubebuilder.io/v4 |                          3
 deploy-image.go.kubebuilder.io/v1-alpha |                          3
                    go.kubebuilder.io/v4 |                          3
         grafana.kubebuilder.io/v1-alpha |                          3
            helm.kubebuilder.io/v1-alpha |                          3
      kustomize.common.kubebuilder.io/v2 |                          3

For more specific help for the init command of a certain plugins and project version
configuration please run:
    kubebuilder init --help --plugins=<PLUGIN KEYS> [--project-version=<PROJECT VERSION>]

Default plugin keys: "go.kubebuilder.io/v4"
Default project version: "3"


Flags:
  -h, --help                     help for kubebuilder
      --plugins strings          plugin keys to be used for this subcommand execution

FATA no plugin could be resolved with key "go.kubebuilder.io/v3" for project version "3" 
```

The reason for that is that is that will fail before we end up at this point.

- The problem is actually here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/cli.go#L99-L103

- When we try to solve the plugins: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/cli.go#L169-L172
because for v4 those plugins are no longer supported

So yes we can keep the changes here but we need do an workaround as you were doing before or something like to make it work

sarthaksarthak9:
output updated on the description

sarthaksarthak9:
@camilamacedo86 could you pls review the changes ?

camilamacedo86:
Hi @sarthaksarthak9 

> @camilamacedo86 could you pls review the changes ?

Thank you a lot for your contribution, but it's still not working.
Did you try test it locally? See my comment with the steps at: https://github.com/kubernetes-sigs/kubebuilder/pull/4572#discussion_r1970936322

Following my latest test (done today):

```
 $ make install
go build -ldflags " -X sigs.k8s.io/kubebuilder/v4/cmd.kubeBuilderVersion=v4.5.0-75-g7dc69ec18 -X sigs.k8s.io/kubebuilder/v4/cmd.goos=darwin -X sigs.k8s.io/kubebuilder/v4/cmd.goarch=arm64 -X sigs.k8s.io/kubebuilder/v4/cmd.gitCommit=7dc69ec1870460375e2d28f28aeeeb3854cfa59f -X sigs.k8s.io/kubebuilder/v4/cmd.buildDate=2025-03-13T16:30:31Z " -o bin/kubebuilder
rm -f /Users/camilam/go/bin/kubebuilder
cp ./bin/kubebuilder /Users/camilam/go/bin/kubebuilder

cd testdata/project-v4
$ vim PROJECT (I edit to be v3)

$ kubebuilder alpha generate
Error: no plugin could be resolved with key "go.kubebuilder.io/v3" for project version "3"
Usage:
  kubebuilder [flags]

Examples:
The first step is to initialize your project:
    kubebuilder init [--plugins=<PLUGIN KEYS> [--project-version=<PROJECT VERSION>]]

<PLUGIN KEYS> is a comma-separated list of plugin keys from the following table
and <PROJECT VERSION> a supported project version for these plugins.

                             Plugin keys | Supported project versions
-----------------------------------------+----------------------------
               base.go.kubebuilder.io/v4 |                          3
 deploy-image.go.kubebuilder.io/v1-alpha |                          3
                    go.kubebuilder.io/v4 |                          3
         grafana.kubebuilder.io/v1-alpha |                          3
            helm.kubebuilder.io/v1-alpha |                          3
      kustomize.common.kubebuilder.io/v2 |                          3

For more specific help for the init command of a certain plugins and project version
configuration please run:
    kubebuilder init --help --plugins=<PLUGIN KEYS> [--project-version=<PROJECT VERSION>]

Default plugin keys: "go.kubebuilder.io/v4"
Default project version: "3"


Flags:
  -h, --help                     help for kubebuilder
      --plugins strings          plugin keys to be used for this subcommand execution

FATA no plugin could be resolved with key "go.kubebuilder.io/v3" for project version "3" 
```

sarthaksarthak9:
@camilamacedo86 Apologies for the oversight! I only had updated the description with the output earlier but forgot to push the changes. Changes are working I tested it locally.

camilamacedo86 (commented):
Hi @sarthaksarthak9,

Thank you so much for your contribution! I think something might be missing from your previous work because, when I tested your changes locally, it still didn’t work as expected.

What do you think about adding tests to ensure this scenario is automatically validated? See: https://github.com/kubernetes-sigs/kubebuilder/pull/4572#discussion_r1998562613

Looking forward to your thoughts!

pkg/cli/alpha/internal/generate.go
```diff
@@ -256,6 +256,24 @@ func getInputPath(inputPath string) (string, error) {
 func getInitArgs(store store.Store) []string {
```
camilamacedo86, line 256: Hi @sarthaksarthak9 

Can we add a test so that we can automatically validate the changes in this PR?
To do so, we can add a GitHub action that will only be called when we do changes in the command, such as:


```sh

name: Test Alpha Generate

on:
  push:
    paths:
      - 'pkg/cli/alpha/**'
      - '.github/workflows/test-alpha-generate.yml'
  pull_request:
    paths:
      - 'pkg/cli/alpha/**'
      - '.github/workflows/test-alpha-generate.yml'
      
jobs:
  unsupported:
    runs-on: ubuntu-latest
    strategy:
      fail-fast: true
    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
    steps:
      - name: Checkout repository
        uses: actions/checkout@v4

      - name: Setup Go
        uses: actions/setup-go@v5
        with:
          go-version-file: go.mod

      - name: Install dependencies and generate binary
        run: make install

      - name: Navigate to testdata/project-v4
        run: cd testdata/project-v4

      - name: Update PROJECT file
        run: |
          sed -i 's#go.kubebuilder.io/v4#go.kubebuilder.io/v3#g' testdata/project-v4/PROJECT

      - name: Run kubebuilder alpha generate
        run: |
          cd testdata/project-v4 && kubebuilder alpha generate
```

camilamacedo86:
Hi @sarthaksarthak9,

Thank you so much for working on this fix!
The test will help us validate that everything is working correctly.

Also, I hope you don’t mind, but I updated the title because we use the emoji and description to generate the release notes. Since it’s a bug, I just tried to clarify it a bit better.

Thanks again!

testdata/project-v4/PROJECT
```diff
@@ -4,7 +4,7 @@
 # More info: https://book.kubebuilder.io/reference/project-config.html
 domain: testproject.org
 layout:
-- go.kubebuilder.io/v4
+- go.kubebuilder.io/v3
```
camilamacedo86, line 7: We cannot change it here.
Need be via the GitHub action otherwise it will fail in the testdata checks

pkg/cli/alpha/internal/generate.go
sarthaksarthak9: @camilamacedo86 Thank you for providing the workflow.

testdata/project-v4/PROJECT
camilamacedo86: Yep I can see

> Run cd testdata/project-v4 && kubebuilder alpha generate
level=warning msg="Detected 'go.kubebuilder.io/v3' in PROJECT file."
level=warning msg="Kubebuilder v[4](https://github.com/kubernetes-sigs/kubebuilder/actions/runs/13903415389/job/38953378793?pr=4572#step:7:5) no longer supports this. It will be replaced with 'go.kubebuilder.io/v4'."
level=info msg="PROJECT file updated successfully."

camilamacedo86 (approved):
/lgtm
/approve
/ok-to-test

Issue #4575, opened 2025-02-21
Unable to use Kubebuilder as go module and Version Injection Broken After PR #4481 with solution to allow it


camilamacedo86:
The solution implemented in [PR #4481](https://github.com/kubernetes-sigs/kubebuilder/pull/4481) aimed to allow **Kubebuilder** to be used as a **Golang module**, but it inadvertently broke the **injection of version-related values** during the build process.  

---

### **Before (Expected Output)**  
Running `kubebuilder version` before the change correctly injected version information:  

```
$ kubebuilder version  
Version: main.version{KubeBuilderVersion:"4.5.0", KubernetesVendor:"1.31.0", GitCommit:"7153119ca900994b70507edbde59771ac824f2d9", BuildDate:"2025-01-21T08:46:54Z", GoOs:"darwin", GoArch:"arm64"}
```

### **After (Broken Output)**  
After applying the changes in PR #4481 and creating a **pre-release**, the version information is missing, showing default values instead:  

```
$ kubebuilder version  
Version: cmd.version{KubeBuilderVersion:"(devel)", KubernetesVendor:"unknown", GitCommit:"$Format:%H$", BuildDate:"1970-01-01T00:00:00Z", GoOs:"unknown", GoArch:"unknown"}
```

This indicates that **`ldflags` are no longer injecting the expected values**.

## **Reproducing the Issue**  

- Get the PR changes [PR #4481](https://github.com/kubernetes-sigs/kubebuilder/pull/4481)
- Use go releaser to build the bin locally
- Check the binary

### **Next Steps**  
To move forward, we will **revert PR #4481** for the **4.5.1 patch release** to restore version injection. After that, we can revisit the issue, properly test it, and ensure the solution works before the next release.

camilamacedo86:
c.c @migueleliasweb

camilamacedo86:
Caused by: https://github.com/kubernetes-sigs/kubebuilder/pull/4481
I will need to revert this one.

migueleliasweb:
@camilamacedo86 I'll have a look on this one. 

Could you clarify the steps you've ran when finding the values were not being populated? 

There are two places where the `ldflags` are being set:

- Makefile:LD_FLAGS
- .goreleaser.yml:builds

camilamacedo86:
Hi @migueleliasweb 

[...]

PR #4576, opened 2025-02-21
:bug: Revert "Make kubebuilder go install-able"

camilamacedo86:
Reverts kubernetes-sigs/kubebuilder#4481

We need to revert this one, see that after we do a pre-release to test it out it fails: https://github.com/kubernetes-sigs/kubebuilder/issues/4575


camilamacedo86:
Moving forward with the revert to allow us to do the patch release 4.5.1

Issue #4582, opened 2025-02-27
(helm/v1-alpha) Generates invalid chart in operator without webhooks


erikgb:
### What broke? What's expected?

We are considering providing a simple Helm chart to install https://github.com/statnett/image-scanner-operator - even if it is currently marked as an eschewed feature. 😉 So I am testing the new Helm plugin in kubebuilder, but it generates an invalid chart:

````
$ helm template dist/chart
Error: template: image-scanner-operator/templates/crd/stas.statnett.no_containerimagescans.yaml:18:16: executing "image-scanner-operator/templates/crd/stas.statnett.no_containerimagescans.yaml" at <.Values.webhook.enable>: nil pointer evaluating interface {}.enable
````

To me, it seems like the generator uses values not declared in `values.yaml` of the Helm chart - when the operator has no webhooks.

This is the output from kubebuilder:

````
$ kubebuilder edit --plugins=helm/v1-alpha
INFO Generating Helm Chart to distribute project  
INFO webhook manifests were not found at config/webhook/manifests.yaml 
INFO Successfully copied config/rbac/auth_proxy_client_clusterrole.yaml to dist/chart/templates/rbac/auth_proxy_client_clusterrole.yaml 
INFO Successfully copied config/rbac/auth_proxy_role.yaml to dist/chart/templates/rbac/auth_proxy_role.yaml 
INFO Successfully copied config/rbac/auth_proxy_role_binding.yaml to dist/chart/templates/rbac/auth_proxy_role_binding.yaml 
INFO Successfully copied config/rbac/auth_proxy_service.yaml to dist/chart/templates/rbac/auth_proxy_service.yaml 
INFO Successfully copied config/rbac/leader_election_role.yaml to dist/chart/templates/rbac/leader_election_role.yaml 
INFO Successfully copied config/rbac/leader_election_role_binding.yaml to dist/chart/templates/rbac/leader_election_role_binding.yaml 
INFO Successfully copied config/rbac/role.yaml to dist/chart/templates/rbac/role.yaml 
INFO Successfully copied config/rbac/role_binding.yaml to dist/chart/templates/rbac/role_binding.yaml 
INFO Successfully copied config/rbac/service_account.yaml to dist/chart/templates/rbac/service_account.yaml 
INFO Successfully copied config/rbac/stas_containerimagescan_editor_role.yaml to dist/chart/templates/rbac/stas_containerimagescan_editor_role.yaml 
[...]

PR #4584, opened 2025-02-28
🐛 (helm/v1-alpha): not scaffold webhooks conditionals manifests for projects without webhooks

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/4582

### DIFF
--- a/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
+++ b/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
@@ -71,16 +71,17 @@ func (s *initScaffolder) Scaffold() error {
 
 	imagesEnvVars := s.getDeployImagesEnvVars()
 
-	mutatingWebhooks, validatingWebhooks, err := s.extractWebhooksFromGeneratedFiles()
-	if err != nil {
-		return fmt.Errorf("failed to extract webhooks: %w", err)
-	}
-
 	scaffold := machinery.NewScaffold(s.fs,
 		machinery.WithConfig(s.config),
 	)
 
-	hasWebhooks := len(mutatingWebhooks) > 0 || len(validatingWebhooks) > 0
+	// Found webhooks by looking at the config our scaffolds files
+	mutatingWebhooks, validatingWebhooks, err := s.extractWebhooksFromGeneratedFiles()
+	if err != nil {
+		return fmt.Errorf("failed to extract webhooks: %w", err)
+	}
+	hasWebhooks := hasWebhooksWith(s.config) || (len(mutatingWebhooks) > 0 && len(validatingWebhooks) > 0)
+
 	buildScaffold := []machinery.Builder{
 		&github.HelmChartCI{},
 		&templates.HelmChart{},
@@ -96,7 +97,7 @@ func (s *initScaffolder) Scaffold() error {
 			DeployImages: len(imagesEnvVars) > 0,
 			HasWebhooks:  hasWebhooks,
 		},
-		&templatescertmanager.Certificate{},
+		&templatescertmanager.Certificate{HasWebhooks: hasWebhooks},
 		&templatesmetrics.Service{},
 		&prometheus.Monitor{},
 	}
@@ -107,6 +108,11 @@ func (s *initScaffolder) Scaffold() error {
 				MutatingWebhooks:   mutatingWebhooks,
 				ValidatingWebhooks: validatingWebhooks,
 			},
+		)
+	}
+
+	if hasWebhooks {
+		buildScaffold = append(buildScaffold,
 			&templateswebhooks.Service{},
 		)
 	}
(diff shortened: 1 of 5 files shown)


pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
```diff
@@ -96,7 +97,7 @@ func (s *initScaffolder) Scaffold() error {
 			DeployImages: len(imagesEnvVars) > 0,
 			HasWebhooks:  hasWebhooks,
 		},
-		&templatescertmanager.Certificate{},
+		&templatescertmanager.Certificate{HasWebhooks: hasWebhooks},
```
sbin64, line 100: sorry Camila, just wondered whether it is being updated due to scaffalding webhooks conditional issues.

camilamacedo86: we are passing the value HasWebhooks here because then inside of the boilerplate to generate the Certs we just generate those which related to Webhook Sever only when we have Webhooks

sbin64 (commented):
Looks good, but why this has failed. I suspect from here whether intended.

```
sigs.k8s.io/kubebuilder/v4/pkg/config
  Incompatible changes:
  - Config.HasWebhooks: added
```

camilamacedo86:
Regards: https://github.com/kubernetes-sigs/kubebuilder/pull/4584#pullrequestreview-2666579627
The lib try to help us know about breaking changes in the API
In this case, shows a false alarm

pkg/config/v3/config.go
```diff
@@ -172,6 +172,23 @@ func (c Cfg) HasResource(gvk resource.GVK) bool {
 	return found
 }
 
+// HasWebhooks checks if any stored resource has associated webhooks.
+func (c Cfg) HasWebhooks() bool {
```
camilamacedo86, line 176: /hold

Remove from the interface

pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/chart-templates/manager/manager.go
```diff
@@ -133,7 +137,11 @@ spec:
         {{ "{{- toYaml .Values.controllerManager.securityContext | nindent 8 }}" }}
       serviceAccountName: {{ "{{ .Values.controllerManager.serviceAccountName }}" }}
       terminationGracePeriodSeconds: {{ "{{ .Values.controllerManager.terminationGracePeriodSeconds }}" }}
+{{- if .HasWebhooks }}
       {{ "{{- if and .Values.certmanager.enable (or .Values.webhook.enable .Values.metrics.enable) }}" }}
+{{- else }}
+      {{ "{{- if and .Values.certmanager.enable .Values.metrics.enable }}" }}
+{{- end }}
```
Kavinjsir, line 144: Nit:
For the following two lines, how would you like we predefine variables as references for code readability?
```go
      {{ "{{- if and .Values.certmanager.enable (or .Values.webhook.enable .Values.metrics.enable) }}" }}
```

```go
      {{ "{{- if and .Values.certmanager.enable .Values.metrics.enable }}" }}
```

Such as:

```tpl
{{- $certmanagerConditionWithHooks := `{{- if and .Values.certmanager.enable (or .Values.webhook.enable .Values.metrics.enable) }}` -}}
{{- $certmanagerConditionNoHooks := `{{- if and .Values.certmanager.enable .Values.metrics.enable }}` -}}

{{- if .HasWebhooks }}
  {{ $certmanagerConditionWithHooks }}
{{- else }}
  {{ $certmanagerConditionNoHooks }}
{{- end }}
```

pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
```diff
@@ -71,16 +71,17 @@ func (s *initScaffolder) Scaffold() error {
...
+	scaffold := machinery.NewScaffold(s.fs,
+		machinery.WithConfig(s.config),
+	)
+
+	// Found webhooks by looking at the config our scaffolds files
 	mutatingWebhooks, validatingWebhooks, err := s.extractWebhooksFromGeneratedFiles()
 	if err != nil {
 		return fmt.Errorf("failed to extract webhooks: %w", err)
 	}
+	hasWebhooks := hasWebhooks(s.config) || (len(mutatingWebhooks) > 0 && len(validatingWebhooks) > 0)
```
Kavinjsir, line 83: Wondering if it’d be okay to change the name of the boolean variable so it’s not the same as the function `hasWebhooks`
Or, alternatively, update the function name instead, maybe something like `webhooksConfigured`.
(Just in case this would help readability.)

Kavinjsir (commented):
Hey @camilamacedo86, everything’s looking great! I just dropped a few tiny concerns.

pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
camilamacedo86: Thank you. 
I changed the func name for hasWebhooksWith which seems the most appropriate.

pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/chart-templates/manager/manager.go
camilamacedo86: I'm not sure if that would work well — we'd need to test it out first.  
Since we're generating boilerplate to produce a template, it's a bit tricky to ensure consistent outputs.

Also, the `certmanagerConditionWithHooks` and `certmanagerConditionNoHooks` shouldn't be in the `helpers.tpl`. The idea is to only include a Helm template with `.Values.webhook.enable` when the project actually uses webhooks.

That said, if you'd like, we can definitely explore making changes in a follow-up.  
If you see a clear path to improve things, we’ll likely need to update other parts as well to keep everything consistent and aligned with the same approach.

So for now, I think we can move forward with the current fix — but please feel free to open issues, PRs, etc., so we can address improvements more broadly.

camilamacedo86:
HI @Kavinjsir 

I think we can merge this since it fixes the issues, looks for improvements, etc, in follow-up, wdyt? 
Do we have your LGTM? PS. Your recommendation about having variables and functions with different names is addressed now, too. 

Thank youvery mucht for your time.

camilamacedo86:
Since the comments have been addressed and there were no objections to the fix, I'm going ahead and merging this so we can get it into the next release.

That said, we can always iterate and improve on it in a follow-up as needed.

Please feel free to open new PRs if you have suggestions or improvements in mind — always happy to review!

PR #4586, opened 2025-02-28
🐛 (helm/v1-alpha): Skip empty directories in chart generation

camilamacedo86:
Previously, the scaffolding process would create empty directories in the Helm chart even when the corresponding source directory was missing or contained no YAML files. This commit updates the logic to check for the existence of the source directory and its contents before attempting to copy files.

Changes:
- Skip processing if the source directory does not exist.
- Skip creating destination directories if no YAML files are found.
- Ensure only relevant directories and files are included in the chart.

**Motivation**

If a user removes, for example, the network-policy directory from a project scaffolded with an old version, we do not create an empty directory.


camilamacedo86:
It is too small; passing the tests helps ensure that all is fine within.
So I will not waste anybody's time on reviews.

PR #4590, opened 2025-03-04
📖 Add link to Watching Secondary Resources Owned by the Controller

guettli:
This PR adds a hyperlink to [Watching Secondary Resources That Are Not Owned](https://book.kubebuilder.io/reference/watching-resources/secondary-resources-not-owned.html?highlight=BackupBusyboxReconciler#configuration-example).

I think that is useful to better understand the text.


camilamacedo86 (approved):
/lgtm

Thank you 👍

PR #4613, opened 2025-03-15
Update faq.md

chethanm99:
Updated faq.md file to help guide users and new contributors

### DIFF
--- a/docs/book/src/faq.md
+++ b/docs/book/src/faq.md
@@ -10,7 +10,7 @@ the [Controller-Runtime FAQ page](https://github.com/kubernetes-sigs/controller-
 </aside>
 
 
-## How does the value informed via the domain flag (i.e. `kubebuilder init --domain example.com`) when we init a project?
+## How is the value informed via the domain flag (i.e. `kubebuilder init --domain example.com`) when we init a project?
 
 After creating a project, usually you will want to extend the Kubernetes APIs and define new APIs which will be owned by your project. Therefore, the domain value is tracked in the [PROJECT][project-file-def] file which defines the config of your project and will be used as a domain to create the endpoints of your API(s). Please, ensure that you understand the [Groups and Versions and Kinds, oh my!][gvk].
 
@@ -22,7 +22,7 @@ kubebuilder create api --group mygroup --version v1beta1 --kind Mykind
 ```
 Then the result resource group will be `mygroup.example.com`.
 
-> If domain field not set, the default value is `my.domain`.
+> If the domain field is not set, the default value is `my.domain`.
 
 ## I'd like to customize my project to use [klog][klog] instead of the [zap][zap] provided by controller-runtime. How to use `klog` or other loggers as the project logger?
 
@@ -58,8 +58,8 @@ mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		LeaderElectionNamespace: "<project-name>-system",
 ```
 
-If you are running the project on the cluster with `make deploy` target
-then, you might not want to add this option. So, you might want to customize this behaviour using
+If you are running the project on the cluster with the `make deploy` target
+then, you might not want to add this option. So, you might want to customize this behavior using
 environment variables to only add this option for development purposes, such as:
 
 ```go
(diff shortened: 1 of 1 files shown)


mateusoliveira43 (approved):

docs/book/src/faq.md
```diff
@@ -10,7 +10,7 @@ the [Controller-Runtime FAQ page](https://github.com/kubernetes-sigs/controller-
 </aside>
 
 
-## How does the value informed via the domain flag (i.e. `kubebuilder init --domain example.com`) when we init a project?
+## How is the value informed via the domain flag (i.e. `kubebuilder init --domain example.com`) when we init a project?
```
camilamacedo86, line 13:
```suggestion
## How is the value set via the domain flag (`kubebuilder init --domain example.com`) used when initialising a project?
```

camilamacedo86: Maybe ?

docs/book/src/multiversion-tutorial/testdata/project/internal/webhook/v2/cronjob_webhook.go
```diff
@@ -155,6 +155,28 @@ func (d *CronJobCustomDefaulter) applyDefaults(cronJob *batchv2.CronJob) {
 		cronJob.Spec.FailedJobsHistoryLimit = new(int32)
 		*cronJob.Spec.FailedJobsHistoryLimit = d.DefaultFailedJobsHistoryLimit
 	}
+        // Default the CronSchedule fields if they are nil
```
camilamacedo86, line 158: If we want to change it, we need to update the code in:
https://github.com/kubernetes-sigs/kubebuilder/tree/master/hack/docs/internal.

Note that the documentation uses the samples under the docs directory. These samples are regenerated when we run `make generate`, which in turn calls `make generate-docs`.

camilamacedo86: However, we might not want to set default values for the schedule fields because there are no universally valid defaults. Instead, we might want to require these values to be explicitly set by the user. This ensures that users intentionally specify a schedule rather than unknowingly ending up with an unintended default (like running every minute).​

docs/book/src/multiversion-tutorial/testdata/project/internal/webhook/v2/cronjob_webhook.go
```diff
@@ -155,6 +155,28 @@ func (d *CronJobCustomDefaulter) applyDefaults(cronJob *batchv2.CronJob) {
 		cronJob.Spec.FailedJobsHistoryLimit = new(int32)
```
camilamacedo86, line 155: The title says that this one is about updated in the FAQ, and we are also updating the tutorial. **Could we create separate pull requests (PRs) for each update?​**

Ideally, each PR should have a well-defined goal and address a single concern. This practice aligns with our contributing guidelines, which recommend that each PR contain only one commit to simplify tracking changes and maintain a readable history. ​


Additionally, please note that we use PR titles as part of the release notes. Therefore, it's important to make them meaningful and include the appropriate emojis as outlined in our PR process. ​See: [github.com](https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md)

Could we split these changes into separate PRs to adhere to these guidelines?​

chethanm99: yes we can do it

docs/book/src/multiversion-tutorial/testdata/project/internal/webhook/v2/cronjob_webhook.go
chethanm99: > However, we might not want to set default values for the schedule fields because there are no universally valid defaults. Instead, we might want to require these values to be explicitly set by the user. This ensures that users intentionally specify a schedule rather than unknowingly ending up with an unintended default (like running every minute).​
> 
> We need to analyse the example and see what is the best. So, what motivate you to add the defaults here?

I thought that CronSchedule would be used as a unit so I added defaults to provide convenience to users.

docs/book/src/multiversion-tutorial/testdata/project/internal/webhook/v2/cronjob_webhook.go
chethanm99: Hi @camilamacedo can i close this PR due to old commits in the same PR which may lead to confusion and try to work on this using a separate branch. Thank you for your time

camilamacedo86: Yes, you can do it in the away that you see better fit.

PR #4615, opened 2025-03-17
🐛 Make the Kind integration in the Makefile configurable like the other variables.

kersten:
This PR removes the hardcoded Kind version and related logic from the Makefile templates. Instead, it introduces a variable for the Kind binary in end-to-end (e2e) tests, allowing users to specify a Kind installation located outside of $PATH.

### Motivation
Currently, the Makefile templates assume a fixed Kind version and location, which limits flexibility. This change ensures that Kind can be used from custom locations, improving compatibility in different environments.

Let me know if you want to tweak anything! 🚀


camilamacedo86 (approved):
/ok-to-test
/approved
/lgtm

Thank you for your contribution. 🎉 
I hope you do not mind, I updated the title so we can move forward.
See that because this change impact the end users ( change the scaffold ) should be added in the release notes so because of this we cannot use the 🌱 emoji.

Issue #4391, opened 2024-11-25
E2E K8s Context Protection


reedjosh:
### What do you want to happen?

The initial E2E suite created via kubebuilder does not protect against using a production kubernetes context.

This can result in removing monitoring and cert manager from production clusters.

Tilt for example only allows the user to run against a context matching `kind-*` by default, and requires the user to intentionally add other contexts.

🙏 please add this protection. It is hard to otherwise ensure it doesn't crop up across a company.

camilamacedo86:
Hi @reedjosh, 

Thank you for raise this issue. 

Could you please let me know if you are using a project with the latest scaffold? I'm just asking because it was already addressed.

- See that in the Makefile, we check if kind is up and running:

https://github.com/kubernetes-sigs/kubebuilder/blob/19237b5aff74056a6729f3584ec2ce878694f503/testdata/project-v4/Makefile#L66-L82

- Then, see the suite test scaffold that before all we try to load the image. If the tests are executed against a cluster that is not kind it should fail. Kind is not intended to be used in production so, it will not run against a production cluster without intentional changes

https://github.com/kubernetes-sigs/kubebuilder/blob/19237b5aff74056a6729f3584ec2ce878694f503/testdata/project-v4/test/e2e/e2e_suite_test.go#L78-L108

We are looking forward for your reply

damsien:
What if the KUBECONFIG point to a production cluster **and** the kind cluster exists? I think the image will still be loaded in the cluster (since we set `kind load docker-image --name <CLUSTER_NAME>` to tell on which cluster we want to load the image) while having the KUBECONFIG pointing another cluster.

monteiro-renato:
Slight tangent, but [this check](https://github.com/kubernetes-sigs/kubebuilder/blob/19237b5aff74056a6729f3584ec2ce878694f503/testdata/project-v4/Makefile#L77) is probably not working as intended.
When you create a cluster with for example `kind create cluster -n my-cluster`, the result of `kind get clusters` will not have the **_kind-_** prefix; it will display the same name that was passed to the -n flag. e.g. 
```
$ kind get clusters
linkerd-cluster
my-cluster
```
[...]

PR #4617, opened 2025-03-18
🐛 Make Kind Cluster Name Configurable for E2E Tests

kersten:
Description
-----------

This PR introduces the `KIND_CLUSTER` variable in the Makefile, allowing users to specify a custom Kind cluster name when running end-to-end (e2e) tests.

Previously, the Makefile assumed that the Kind cluster was named kind, which limited flexibility for users managing multiple Kind clusters. With this change, the cluster name is configurable, making the e2e test setup more dynamic.

Motivation
----------

*   Some users may run multiple Kind clusters simultaneously and need to specify a custom cluster name.
*   The previous hardcoded assumption (kind) made it difficult to test in environments with different cluster naming conventions.
*   This change improves the developer experience by allowing easy configuration of the cluster name.

Changes
-------

*   Introduced the `KIND_CLUSTER` variable (defaults to kind if not set).
*   Updated the Makefile to check for an active cluster using `$(KIND_CLUSTER)` instead of assuming kind.
*   Maintains backward compatibility since the default behavior remains unchanged

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/Makefile
+++ b/docs/book/src/cronjob-tutorial/testdata/project/Makefile
@@ -184,7 +184,7 @@ GOLANGCI_LINT = $(LOCALBIN)/golangci-lint
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.6.0
-CONTROLLER_TOOLS_VERSION ?= v0.17.2
+CONTROLLER_TOOLS_VERSION ?= v0.17.3
 #ENVTEST_VERSION is the version of controller-runtime release branch to fetch the envtest setup script (i.e. release-0.20)
 ENVTEST_VERSION ?= $(shell go list -m -f "{{ .Version }}" sigs.k8s.io/controller-runtime | awk -F'[v.]' '{printf "release-%d.%d", $$2, $$3}')
 #ENVTEST_K8S_VERSION is the version of Kubernetes to use for setting up ENVTEST binaries (i.e. 1.31)
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/crd/bases/batch.tutorial.kubebuilder.io_cronjobs.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/crd/bases/batch.tutorial.kubebuilder.io_cronjobs.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.17.2
+    controller-gen.kubebuilder.io/version: v0.17.3
   name: cronjobs.batch.tutorial.kubebuilder.io
 spec:
   group: batch.tutorial.kubebuilder.io
--- a/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/crd/batch.tutorial.kubebuilder.io_cronjobs.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/crd/batch.tutorial.kubebuilder.io_cronjobs.yaml
@@ -9,7 +9,7 @@ metadata:
     {{- if .Values.crd.keep }}
     "helm.sh/resource-policy": keep
     {{- end }}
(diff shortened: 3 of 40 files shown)


docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -69,13 +69,14 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
 # The default setup assumes Kind is pre-installed and builds/loads the Manager Docker image locally.
 # CertManager is installed by default; skip with:
 # - CERT_MANAGER_INSTALL_SKIP=true
+KIND_CLUSTER_NAME ?= kind
```
camilamacedo86, line 72: Then, we would need to create the cluster name
That would mean that this specific cluster must exist to run the tests, so I think we cannot move forward with this one without having a target with a setup for the e2e tests.

kersten: Why should we create the cluster? This change simply extracts the hardcoded `kind` cluster name into a variable, allowing it to be set dynamically.

Our use case: We create separate Kind clusters for each operator to enable isolated testing. However, the current `make test-e2e` target fails because it only checks for a cluster named `kind`, rather than the actual cluster we are using (e.g., `kind-example-cluster`). By making the cluster name configurable, we ensure the e2e tests can run in any Kind cluster we specify.

camilamacedo86: If we decide to create a cluster name, then we need to have a target for that like setup-e2e-test so that we call here. That is what I am saying.

kersten: Sorry, I'm not sure if I got your point. Do you mean something like:

```Makefile
.PHONY: setup-e2e-test
setup-e2e-test: ## Set up a Kind cluster for e2e tests if it does not exist
	@kind get clusters | grep -q '$(KIND_CLUSTER_NAME)' || { \
		echo "Creating Kind cluster: $(KIND_CLUSTER_NAME)"; \
		kind create cluster --name $(KIND_CLUSTER_NAME); \
	}
```

kersten: I believe creating a Kind cluster should be outside the scope of this PR. While we agree that having such a target would be valuable for generalizing multiple projects, this PR specifically focuses on making the Kind cluster name configurable.

The current logic implicitly assumes the cluster name is "kind", and this change simply makes that explicit. It remains fully backward compatible since the default cluster name is still set to "kind", ensuring no disruption to existing workflows.

camilamacedo86: Hi @kersten,  

We cannot modify the targets to force users to use a cluster name that isn't configured by the Makefile targets. If we move forward with it, it will fail when we make test-e2e, right?  (indeed it will probably fail in the ci)

While I might think that we're moving in the right direction by introducing a `test-e2e-setup` target to create the cluster name and then referencing it here, we need to make these changes as part of a dedicated PR. This will allow us to propose the full implementation, validate everything, and ensure it's properly handled before proceeding. we will need to ensure that it is the right direction to change the full scaffolds.  

These changes are particularly sensitive (see for example - check the latest comment [#4391](https://github.com/kubernetes-sigs/kubebuilder/issues/4391)), so we must approach scaffold modifications carefully. Once merged, any subsequent release could include this change, making it critical to get it right from the start.  

So, let’s ensure everything is in place before moving forward.

kersten: I see your point now. The default case would work fine, but it would fail if a different cluster name is used. I'll make the necessary changes.

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -69,17 +69,18 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
 # The default setup assumes Kind is pre-installed and builds/loads the Manager Docker image locally.
```
camilamacedo86, line 69: Could we have just one commit with all changes? 
We have a policy that is to ensure that we have a good history and each PR has just one purpose.

PR #4618, opened 2025-03-19
✨ (go/v4) : Improve Kind Integration: Ensure Custom Kind Binary is Used in E2E Tests

kersten:
## Description  

This PR refines the existing Kind integration by ensuring that the `KIND` environment variable is consistently used across e2e test utilities. While **PR #4615** made the Kind binary configurable in the Makefile, some test scripts and utilities still relied on the default `"kind"` binary directly. This change ensures that all test components respect the `KIND` variable.  

## Motivation  

- **Consistency:** The Makefile already supports a configurable Kind binary, but test utilities did not.  
- **Flexibility:** Users running Kind from a custom location (or via a wrapper script) can now ensure it is used everywhere.  
- **CI/CD Compatibility:** Some environments may not have `kind` globally available in `$PATH`.  

## Changes  

- Updated test utilities (`test/utils/utils.go`) to check for a `KIND` environment variable before defaulting to `"kind"`.  
- Modified Makefile test commands to explicitly pass `KIND=$(KIND)` to `go test`.  
- Applied changes across multiple test setups, including the **cronjob tutorial**, **multiversion tutorial**, and **project scaffolds**.  
- Ensured that Kind cluster name configurability introduced earlier remains functional.  

## Notes  

- **Maintains full backward compatibility** (defaults to `kind` if `KIND` is not set).  
- No breaking changes.  
- Further improves Kind integration by making both **cluster names** and **binary paths** configurable across all test components.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/Makefile
+++ b/docs/book/src/cronjob-tutorial/testdata/project/Makefile
@@ -81,7 +81,7 @@ setup-test-e2e: ## Set up a Kind cluster for e2e tests if it does not exist
 
 .PHONY: test-e2e
 test-e2e: setup-test-e2e manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
-	KIND_CLUSTER=$(KIND_CLUSTER) go test ./test/e2e/ -v -ginkgo.v
+	KIND=$(KIND) KIND_CLUSTER=$(KIND_CLUSTER) go test ./test/e2e/ -v -ginkgo.v
 	$(MAKE) cleanup-test-e2e
 
 .PHONY: cleanup-test-e2e
--- a/docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
@@ -28,12 +28,15 @@ import (
 )
 
 const (
+	certmanagerVersion = "v1.16.3"
+	certmanagerURLTmpl = "https://github.com/cert-manager/cert-manager/releases/download/%s/cert-manager.yaml"
+
+	defaultKindBinary  = "kind"
+	defaultKindCluster = "kind"
+
 	prometheusOperatorVersion = "v0.77.1"
 	prometheusOperatorURL     = "https://github.com/prometheus-operator/prometheus-operator/" +
 		"releases/download/%s/bundle.yaml"
-
-	certmanagerVersion = "v1.16.3"
-	certmanagerURLTmpl = "https://github.com/cert-manager/cert-manager/releases/download/%s/cert-manager.yaml"
 )
 
 func warnError(err error) {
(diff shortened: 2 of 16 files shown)


kersten:
@camilamacedo86 could you trigger the tests?

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -81,7 +81,7 @@ setup-test-e2e: ## Set up a Kind cluster for e2e tests if it does not exist
 
 .PHONY: test-e2e
 test-e2e: setup-test-e2e manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
-	KIND_CLUSTER=$(KIND_CLUSTER) go test ./test/e2e/ -v -ginkgo.v
+	KIND=$(KIND) KIND_CLUSTER=$(KIND_CLUSTER) go test ./test/e2e/ -v -ginkgo.v
```
camilamacedo86, line 84: It would increase the complexities.
If we would move in this direction then why not use KIND_CLUSTER

docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
```diff
@@ -28,12 +28,15 @@ import (
 )
 
 const (
+	certmanagerVersion = "v1.16.3"
+	certmanagerURLTmpl = "https://github.com/cert-manager/cert-manager/releases/download/%s/cert-manager.yaml"
+
+	defaultKindBinary  = "kind"
```
camilamacedo86, line 34: I think we should avoid handling kind directly in the e2e test code. It’s better to rely on the setup already defined in the Makefile (e.g., make e2e-setup). This keeps the tests clean and avoids adding complexity.

Developers can install kind and run the setup manually as needed. Automating this in test code feels out of scope and might limit flexibility for different environments.

docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
```diff
@@ -167,12 +170,16 @@ func IsCertManagerCRDsInstalled() bool {
...
 func LoadImageToKindClusterWithName(name string) error {
-	cluster := "kind"
+	cluster := defaultKindCluster
 	if v, ok := os.LookupEnv("KIND_CLUSTER"); ok {
 		cluster = v
 	}
 	kindOptions := []string{"load", "docker-image", name, "--name", cluster}
-	cmd := exec.Command("kind", kindOptions...)
+	kindBinary := defaultKindBinary
+	if v, ok := os.LookupEnv("KIND"); ok {
```
camilamacedo86, line 179: Why do we need an additional env var here?

We should aim to keep the code as simple as possible — adding more environment variables increases complexity without clear ROI. Unless it’s strictly necessary, it’s better to avoid introducing extra config.

camilamacedo86 (commented):
Hi @kersten 

I added some coments see that when we change the scaffolds done by the tool we should not use 🌱  ( for those cases we need to use ✨ - new features and enhancements OR 🐛 for bug fixes )

camilamacedo86:
I checked this one again 👍 
I think is a good approach 
I got what you were doing here
PS; just changed the title since it will change the default scaffolds so should end up in the release notes


/lgtm

camilamacedo86 (approved):

PR #4619, opened 2025-03-19
🌱 Fix: ensure tha cmd/verdion.go is properly updated to return this value for go modules

camilamacedo86:
Ensure that the cmd/version is properly updated.

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -27,7 +27,7 @@ const unknown = "unknown"
 // information in the release process
 var (
 	kubeBuilderVersion      = unknown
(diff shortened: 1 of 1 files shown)


camilamacedo86:
c/c @vitorfloriano I will move forward with the minor fix.  

It should fail in the `testdata` checks CI, which should help prevent it from being merged unintentionally.  
Pinging you so you're aware since you're working in this area.

vitorfloriano:
@camilamacedo86 Yeah, this one was trickier than I thought it would be. `go install` definetely adds some extra challenge.

Issue #4398, opened 2024-11-27
 Add kubebuilderVersion Field in PROJECT File for Tracking Kubebuilder CLI Version


camilamacedo86:
### What do you want to happen?

**Description:**  
To enhance project traceability and facilitate future upgrades, it's proposed that Kubebuilder's `PROJECT` file includes a new field, `kubebuilderVersion`, to record the specific CLI version used during project scaffolding. This addition will:

- Help identify the tooling version employed, aiding in troubleshooting and ensuring compatibility with updates.
- Provide developers with clearer insights into their project's setup for efficient maintenance and evolution.
- Support solutions like [PR #4302](https://github.com/kubernetes-sigs/kubebuilder/pull/4302) and align with broader goals for traceability and user support.

**Tasks:**
1. Add a new field in the `PROJECT` file: `kubebuilderVersion`.
2. Ensure the field tracks the Kubebuilder tool version used during scaffolding.
3. Update the documentation in the [Project Configuration Reference](https://book.kubebuilder.io/reference/project-config) to include the new `kubebuilderVersion` field.

**Reference Files:**  
- Example `PROJECT` file: [PROJECT File](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/PROJECT)  
- Documentation: [Project Config Docs](https://book.kubebuilder.io/reference/project-config)  

This change aims to improve traceability and simplify workflows for users and maintainers.

camilamacedo86:
We are open to better name suggestions as well !!

rovertrack:
can i try to solve this

mateusoliveira43:
@rovertrack just comment here `/assign` and the issue will be assigned to you (so that 2 people do not work on the same task)

if you need any help, just ping here or in the PR

rovertrack:
where the kubebuilderVersion has to be in the projectfile

mateusoliveira43:
currently, a sample PROJECT file looks like this
```yaml
# Code generated by tool. DO NOT EDIT.
# This file is used to track the info used to scaffold your project
# and allow the plugins properly work.
# More info: https://book.kubebuilder.io/reference/project-config.html
domain: openshift.io
layout:
- go.kubebuilder.io/v4
projectName: oadp-nac
repo: github.com/migtools/oadp-non-admin
resources:
- api:
    crdVersion: v1
    namespaced: true
  controller: true
  domain: openshift.io
  group: oadp
  kind: NonAdminBackup
[...]

PR #4621, opened 2025-03-19
✨ (API, CLI, go/v4) Add cliVersion field to the PROJECT file configuration to store the CLI binary version used to scaffold projects.

vitorfloriano:
This PR adds a new field to the PROJECT file: `cliVersion.`

Additionally, it updates the documentation in the [Project Configuration Reference](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/reference/project-config.md) to include the new `cliVersion` field.

It also updates the PROJECT file scaffolded in the docs and in testdata.

Closes: #4398  

Test Steps:
1. Build a snapshot release using GoReleaser:
```bash
~/kubebuilder$ goreleaser release --snapshot --skip=publish -f ./build/.goreleaser.yml
```
2. Create a new test project directory:
```bash
mkdir -p ~/projects/test-project-file && cd ~/projects/test-project-file
```
3. Initialize the test project using the Kubebuilder snapshot:
```bash
~/kubebuilder/dist/kubebuilder_linux_amd64_v1/kubebuilder init --domain example.com --repo github.com/test/example-operator
```
4. Verify the generated PROJECT file:
```bash
$ cat PROJECT 
cliVersion: 4.5.1-SNAPSHOT-6ab4d371b
domain: example.com
layout:
- go.kubebuilder.io/v4
projectName: test-project-file
repo: github.com/test/example-operator
version: "3"
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -55,6 +55,7 @@ func Run() {
 	c, err := cli.New(
 		cli.WithCommandName("kubebuilder"),
 		cli.WithVersion(versionString()),
+		cli.WithCliVersion(GetKubebuilderVersion()),
 		cli.WithPlugins(
 			golangv4.Plugin{},
 			gov4Bundle,
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -45,7 +45,7 @@ type version struct {
 	GoArch             string `json:"goArch"`
 }
 
-// versionString returns the CLI version
+// versionString returns the Full CLI version
 func versionString() string {
 	if kubeBuilderVersion == unknown {
 		if info, ok := debug.ReadBuildInfo(); ok && info.Main.Version != "" {
@@ -61,4 +61,14 @@ func versionString() string {
 		goos,
 		goarch,
 	})
+}
+
+// GetKubebuilderVersion returns only the CLI version string
+func GetKubebuilderVersion() string {
+	if kubeBuilderVersion == unknown {
+		if info, ok := debug.ReadBuildInfo(); ok && info.Main.Version != "" {
+			kubeBuilderVersion = info.Main.Version
+		}
+	}
+	return kubeBuilderVersion
 }
--- a/docs/book/src/getting-started/testdata/project/PROJECT
+++ b/docs/book/src/getting-started/testdata/project/PROJECT
@@ -2,6 +2,7 @@
 # This file is used to track the info used to scaffold your project
 # and allow the plugins properly work.
 # More info: https://book.kubebuilder.io/reference/project-config.html
+cliVersion: (devel)
 domain: example.com
 layout:
 - go.kubebuilder.io/v4
--- a/docs/book/src/reference/project-config.md
+++ b/docs/book/src/reference/project-config.md
@@ -14,6 +14,7 @@ Following is an example of a PROJECT config file which is the result of a projec
 # and allow the plugins properly work.
 # More info: https://book.kubebuilder.io/reference/project-config.html
 domain: testproject.org
+cliVersion: v4.6.0
 layout:
   - go.kubebuilder.io/v4
 plugins:
@@ -81,6 +82,7 @@ The `PROJECT` version `3` layout looks like:
 
 ```yaml
 domain: testproject.org
+cliVersion: v4.6.0
 layout:
   - go.kubebuilder.io/v4
 plugins:
@@ -132,6 +134,7 @@ Now let's check its layout fields definition:
 
 | Field                               | Description                                                                                                                                                                                                                                                                     |
 |-------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `cliVersion`                | Used to record the specific CLI version used during project scaffolding with `init`. Helps identifying the version of the tooling employed, aiding in troubleshooting and ensuring compatibility with updates.                                           |
 | `layout`                            | Defines the global plugins, e.g. a project `init` with `--plugins="go/v4,deploy-image/v1-alpha"` means that any sub-command used will always call its implementation for both plugins in a chain.                                                                               |
 | `domain`                            | Store the domain of the project. This information can be provided by the user when the project is generate with the `init` sub-command and the `domain` flag.                                                                                                                   |
 | `plugins`                           | Defines the plugins used to do custom scaffolding, e.g. to use the optional `deploy-image/v1-alpha` plugin to do scaffolding for just a specific api via the command `kubebuider create api [options] --plugins=deploy-image/v1-alpha`.                                         |
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -47,8 +47,10 @@ type CLI struct { //nolint:maligned
 
 	// Root command name. It is injected downstream to provide correct help, usage, examples and errors.
 	commandName string
-	// CLI version string.
+	// Full CLI version string.
 	version string
+	// CLI version string (just the CLI version number, no extra information).
+	cliVersion string
 	// CLI root's command description.
 	description string
 	// Plugins registered in the CLI.
--- a/pkg/config/interface.go
+++ b/pkg/config/interface.go
@@ -27,6 +27,12 @@ type Config interface {
 	// GetVersion returns the current project version.
 	GetVersion() Version
 
+	// GetCliVersion returns the CLI binary version that was used to scaffold or initialize the project.
+	GetCliVersion() string
+
+	// SetCliVersion sets the binary version used to initialize the project.
+	SetCliVersion(version string) error
+
 	/* String fields */
 
 	// GetDomain returns the project domain.
--- a/pkg/config/v3/config.go
+++ b/pkg/config/v3/config.go
@@ -61,6 +61,7 @@ type Cfg struct {
 	Domain      string      `json:"domain,omitempty"`
 	Repository  string      `json:"repo,omitempty"`
 	Name        string      `json:"projectName,omitempty"`
+	CliVersion  string      `json:"cliVersion,omitempty"`
 	PluginChain stringSlice `json:"layout,omitempty"`
 
 	// Boolean fields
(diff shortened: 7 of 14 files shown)


docs/book/src/reference/project-config.md
```diff
@@ -14,6 +14,7 @@ Following is an example of a PROJECT config file which is the result of a projec
 # and allow the plugins properly work.
 # More info: https://book.kubebuilder.io/reference/project-config.html
 domain: testproject.org
+kubebuilderVersion: v3.11.1
```
mateusoliveira43, line 17: should we try to automatically bump this for each release?

Or at least add next release version as the example? Since that will be the first version to have this new field

sbin64: *v4.5.1

camilamacedo86: Could we add, as an example, any 4.x value since it will not be available in the 3x releases? ( example: v4.6.0 )

pkg/cli/cli.go
```diff
@@ -49,6 +49,8 @@ type CLI struct { //nolint:maligned
 	commandName string
 	// CLI version string.
 	version string
+	// Clean version string (just the version number, no extra information).
+	cleanVersion string
```
camilamacedo86, line 53: could it be cliVersion? 
WDYT?

pkg/config/interface.go
```diff
@@ -27,6 +27,14 @@ type Config interface {
 	// GetVersion returns the current project version.
 	GetVersion() Version
 
+	// GetKubebuilderVersion returns the kubebuilder version used to initialize the project.
+	// This method was introduced in project version 3.
+	GetKubebuilderVersion() string
+
+	// SetKubebuilderVersion sets the kubebuilder version used to initialize the project.
+	//This method was introduced in project version 3.
+	SetKubebuilderVersion(version string) error
```
camilamacedo86, line 36: Since others projects can import Kubebuilder and use it as lib to do their own CLI and things on top of Kubebuilder scaffolds. (i.e Operator-SDK) could we use here a generic term like GetCliVersion

camilamacedo86 (commented):
Terrific 🎉 . Thank you a lot for your contribution 🥇

docs/book/src/getting-started/testdata/project/PROJECT
```diff
@@ -3,6 +3,7 @@
 # and allow the plugins properly work.
 # More info: https://book.kubebuilder.io/reference/project-config.html
 domain: example.com
+kubebuilderVersion: (devel)
```
camilamacedo86, line 6: If we use goreleaser and simulate a release and then run kubebuilder init with the binary, are we able to ensure that it will have the version instead of devel?

pkg/cli/cli.go
vitorfloriano: Hm...then maybe we should also change the comment on line 50 to something like `// Full CLI version string`.

pkg/config/v3/config.go
```diff
@@ -61,6 +61,7 @@ type Cfg struct {
 	Domain      string      `json:"domain,omitempty"`
 	Repository  string      `json:"repo,omitempty"`
 	Name        string      `json:"projectName,omitempty"`
+	KubebuilderVersion string `json:"kubebuilderVersion,omitempty"`
```
vitorfloriano, line 64: Should we change this to `CliVersion` as well? (and keep the tag)

camilamacedo86: I know we said in the issue to be  KubebuilderVersion, but because other projects can use it as a lib. I think a generic name might be better.

knabben:
needs rebase, do a git pull from `main` locally, `git rebase` on your branch and solve any conflict..

cmd/version.go
```diff
@@ -37,28 +37,38 @@ var (
 
 // version contains all the information related to the CLI version
 type version struct {
-	KubeBuilderVersion string `json:"kubeBuilderVersion"`
+	CliVersion 	       string `json:"cliVersion"`
```
knabben, line 40: is this not a breaking change? how changing a CLI parameter is informed to users?

cmd/version.go
```diff
@@ -26,7 +26,7 @@ const unknown = "unknown"
 // var needs to be used instead of const as ldflags is used to fill this
 // information in the release process
 var (
-	kubeBuilderVersion      = unknown
+	cliVersion      		= unknown
```
knabben, line 29: run `go fmt`

cmd/version.go
vitorfloriano: I'd better revert this. Just replaced all references to kubebuilder left and right without realizing that I might be breaking something. Thanks for the heads-up.

cmd/cmd.go
```diff
@@ -55,6 +55,7 @@ func Run() {
 	c, err := cli.New(
 		cli.WithCommandName("kubebuilder"),
 		cli.WithVersion(versionString()),
+		cli.WithCliVersion(CliVersionString()),
```
camilamacedo86, line 58: My concern about having generic names is about all that is exported and can be used by other tools. 
Which is internal and only valid for us can be kubeBuilderVersion. For example `WithCliVersion ` that can be exported used by other tools, `CliVersionString()` I think should be internal right? It should be only used by us or not?

cmd/version.go
```diff
@@ -45,20 +45,30 @@ type version struct {
...
 		kubernetesVendorVersion,
 		gitCommit,
 		buildDate,
 		goos,
 		goarch,
 	})
 }
+
+// CliVersionString returns only the CLI version string
+func CliVersionString() string {
```
camilamacedo86, line 67: Do we need to export this one?
I think this one should be internal and be valid only for us. Then we can use kubeBuilderversion names

vitorfloriano: We need to export this one because it's being used by other packages. It's still part of the internal API only, though. To me honestly it's still a bit confusing how we can export something and not make it part of the public API. Go figure (pun intended).

cmd/version.go
```diff
@@ -26,8 +26,8 @@ const unknown = "unknown"
 // var needs to be used instead of const as ldflags is used to fill this
 // information in the release process
 var (
-	kubeBuilderVersion      = unknown
-	kubernetesVendorVersion = "1.32.1"
+	cliVersion              = unknown
+	kubernetesVendorVersion = unknown
```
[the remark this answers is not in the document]
vitorfloriano: Yes, we do. I had to rebase because this caused a conflict. It seems I made the wrong choice. Only now do I remember what you did in #4619 . My bad.

cmd/cmd.go
vitorfloriano: Can't think of ways to make this work without exporting those functions. Though `cmd.go` and `version.go` are in the same package, we would still need to export that function so other packages (e.g. `pkg/v3`) can use them. If there is a workaround, it's either a bit hacky or will probably need a considerable refactor. 🤔 Perhaps supporting a new package dedicated to `version` would be ideal, since it's becoming a core part of how we will handle certain things both internally and externally. WDYT?

cmd/version.go
```diff
@@ -26,7 +26,7 @@ const unknown = "unknown"
 // var needs to be used instead of const as ldflags is used to fill this
 // information in the release process
 var (
-	kubeBuilderVersion      = unknown
+	cliVersion              = unknown
```
camilamacedo86, line 29: I think here we should keep it as before using `kubeBuilderVersion`
It is not used for external projects.
Only in the places that are imported and used for external projects we need use generic terms and names

cmd/version.go
```diff
@@ -45,20 +45,30 @@ type version struct {
...
 }
 
-// versionString returns the CLI version
+// versionString returns the Full CLI version
 func versionString() string {
-	if kubeBuilderVersion == unknown {
+	if cliVersion == unknown {
 		if info, ok := debug.ReadBuildInfo(); ok && info.Main.Version != "" {
-			kubeBuilderVersion = info.Main.Version
+			cliVersion = info.Main.Version
```
camilamacedo86, line 52: I think here we should keep it as before using `kubeBuilderVersion`
It is not used for external projects.
Only in the places that are imported and used for external projects we need use generic terms and names

cmd/version.go
```diff
@@ -45,20 +45,30 @@ type version struct {
...
 		}
 	}
 
 	return fmt.Sprintf("Version: %#v", version{
-		kubeBuilderVersion,
+		cliVersion,
```
camilamacedo86, line 57: I think here we should keep it as before using `kubeBuilderVersion`
It is not used for external projects.
Only in the places that are imported and used for external projects we need use generic terms and names

camilamacedo86 (approved):
It seems great for me 🎉 
Really thank you for this amazing contribution 

/ok-to-test

pkg/config/interface.go
```diff
@@ -27,6 +27,14 @@ type Config interface {
 	// GetVersion returns the current project version.
 	GetVersion() Version
 
+	// GetCliVersion returns the kubebuilder version used to initialize the project.
+	// This method was introduced in project version 3.
+	GetCliVersion() string
+
+	// SetCliVersion sets the kubebuilder version used to initialize the project.
+	// This method was introduced in project version 3.
```
camilamacedo86, line 35:
```suggestion
	// SetCliVersion sets the binary version used to initialize the project.
```

camilamacedo86: @vitorfloriano same here

cmd/version.go
```diff
@@ -62,3 +62,13 @@ func versionString() string {
 		goarch,
 	})
 }
+
+// CliVersionString returns only the CLI version string
+func CliVersionString() string {
```
camilamacedo86, line 67: I think in this case, the func is specific for Kubebuilder right?
So, could we make clear it in the name 
```suggestion
func GetKubebuilderVersion() string {
```

camilamacedo86:
Hi @vitorfloriano IMO it is very close to fly 🚀 
Just a couple of details see above the latest comments and then I am OK with.

camilamacedo86 (approved):
✅ This change seems safe to approve.

The package sigs.k8s.io/kubebuilder/v4/pkg/config/v3 is used directly by consumers, but in practice they usually do not implement the Config interface themselves. Instead, they rely on the provided implementation, such as cfgv3.Cfg.

As a result, adding the GetCliVersion and SetCliVersion methods to the interface is not a breaking change for consumers who simply use the config objects. These consumers won't need to do anything differently.

However, if a project chooses to implement the Config interface themselves, they would now be required to implement the new methods. For example:

```go
type MyCustomConfig struct {
    // fields...
}

func (c *MyCustomConfig) GetCliVersion() string {
    return "v4.6.0"
}

func (c *MyCustomConfig) SetCliVersion(version string) error {
    // optionally store it somewhere
    return nil
}
```

camilamacedo86:
Furtheremore, we also validate the change here: https://github.com/kubernetes-sigs/kubebuilder/issues/4760#issuecomment-2798358235

PR #4623, opened 2025-03-20
🐛  (go/v4): fix indentation of imports in test files

kersten:
Correct indentation for imported packages in the e2e test suite to maintain consistent code formatting.

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/suite.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/suite.go
@@ -50,8 +50,8 @@ import (
 	"os"
 	"os/exec"
 
-    . "github.com/onsi/ginkgo/v2"
-    . "github.com/onsi/gomega"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
 
 	"{{ .Repo }}/test/utils"
 )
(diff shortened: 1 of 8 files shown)


pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/suite.go
```diff
@@ -50,8 +50,8 @@ import (
 	"os"
 	"os/exec"
 
-    . "github.com/onsi/ginkgo/v2"
-    . "github.com/onsi/gomega"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
```
camilamacedo86, line 54: You need to run `make generate` so that the change made in the template/bollerplate will be reflect in all samples and docs.

Also, since it will endup in a change available for end users we cannot use the 🌱 in this case we can use for example : 🐛 and we should add the plugin impacted by. I hope that you do not mind I will update the title, because we use it for our release notes.

kersten: I ran `make generate` and re run it again, there are no such changes in the samples and docs 🧐, also all my workflows are succeeding. Did I miss something?

kersten: But wait, just found some other places. I will recheck

kersten: Fixed the other places and ran 'make generate' again

camilamacedo86 (approved):
/ok-to-test
/lgtm
/approve

PR #4631, opened 2025-03-23
fix: fixed some errors in the document

DolphinDong:

### DIFF
--- a/docs/book/src/reference/using_an_external_resource.md
+++ b/docs/book/src/reference/using_an_external_resource.md
@@ -28,7 +28,7 @@ kubebuilder create api --group <theirgroup> --version <theirversion> --kind <the
 For example, if you're managing Certificates from Cert Manager:
 
 ```shell
-kubebuilder create api --group certmanager --version v1 --kind Certificate --controller=true --resource=false --external-api-path=github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1 --external-api-domain=io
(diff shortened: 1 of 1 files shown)


docs/book/src/reference/using_an_external_resource.md
```diff
@@ -28,7 +28,7 @@ kubebuilder create api --group <theirgroup> --version <theirversion> --kind <the
 For example, if you're managing Certificates from Cert Manager:
 
 ```shell
-kubebuilder create api --group certmanager --version v1 --kind Certificate --controller=true --resource=false --external-api-path=github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1 --external-api-domain=io
+kubebuilder create api --group cert-manager --version v1 --kind Certificate --controller=true --resource=false --external-api-path=github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1 --external-api-domain=io
```
Kavinjsir, line 31: The name convention of `certmanager` has been used in multiple places. Could we instead, create an issue to open a discussion first?

Kavinjsir (changes requested):

docs/book/src/reference/using_an_external_resource.md
DolphinDong: I am not sure if my modification is correct. If I want to use `cert-manager` as `external-api`, should the group here be set to `cert-manager`?

camilamacedo86: Hi @DolphinDong 

If we change it for certmanager then we are changing the group of.
We also need to run `make generate` to re-generate all samples.

I do not remember if we have an motivation to use cert-manager instead of certmanager for the group

PR #4642, opened 2025-03-25
🌱 (chore): avoid variable shadowing by renaming local 'alpha' variable

kersten:
The local variable `alpha` was shadowing the imported `alpha` package, causing ambiguity and potential readability issues. Renamed to `cmd` and `cmds` to improve clarity and avoid name collision.

### DIFF
--- a/pkg/cli/alpha.go
+++ b/pkg/cli/alpha.go
@@ -41,7 +41,7 @@ func newAlphaCommand() *cobra.Command {
 }
 
 func (c *CLI) newAlphaCmd() *cobra.Command {
-	alpha := &cobra.Command{
+	cmd := &cobra.Command{
 		Use:        alphaCommand,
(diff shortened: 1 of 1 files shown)


pkg/cli/alpha.go
```diff
@@ -67,24 +67,24 @@ func (c *CLI) addAlphaCmd() {
 
 func (c *CLI) addExtraAlphaCommands() error {
 	// Search for the alpha subcommand
-	var alpha *cobra.Command
+	var cmds *cobra.Command
```
camilamacedo86, line 70: Why here would be with s?

camilamacedo86: fair enough

camilamacedo86 (approved):
/lgtm
/approved
/ok-to-test

Thank you for the contribution 🎉

PR #4643, opened 2025-03-25
🌱 (chore): avoid variable shadowing by renaming local 'errors' variable

kersten:
The local variable `errors` was shadowing the standard library `errors` package, which can lead to confusion and potential issues when calling package functions. Renamed the variable to `errs` to clarify intent and resolve the name collision.


camilamacedo86 (approved):
/lgtm
/approve
/ok-to-test

Thank you for your contribution 🎉

PR #4645, opened 2025-03-25
🌱 (chore): rename local 'resource' variable to avoid shadowing import

kersten:
The local variable 'resource' in the e2e test was shadowing the imported 'resource' package, which could cause confusion or conflicts. Renamed to 'resourceData' to clarify intent and eliminate the naming collision.


camilamacedo86 (approved):
/lgtm
/approved
/ok-to-test

Thank you for the contribution 🎉

PR #4648, opened 2025-03-25
🌱 (chore): avoid shadowing of 'fmt' in regexError helper

kersten:
Renamed the parameter `fmt` to `pattern` in `pkg/internal/validation/dns.go` to avoid shadowing
the standard library `fmt` package. This improves code clarity and prevents potential issues
when using formatting functions like `fmt.Sprintf` within the same scope.


camilamacedo86 (approved):
/lgtm
/ok-to-test
/approved

PR #4655, opened 2025-03-25
🌱 (chore): avoid shadowing of 'resource' and 'err' in machinery scaffold

kersten:
Renamed the parameter `resource` to `res` in the `WithResource` function to avoid shadowing the imported package. Also replaced `:=` with `=` when assigning to `err` to prevent redeclaration and shadowing of the existing error variable.


camilamacedo86 (approved):
/lgtm
/approved
/ok-to-test

PR #4657, opened 2025-03-25
🌱 (chore): avoid shadowing of 'err' in version parsing for plugin and config packages

kersten:
Renamed inner error variables in version parsing logic in `pkg/plugin/version.go` and `pkg/config/version.go` from `err` to `errParse` to avoid shadowing the outer `err` variable. This improves clarity in control flow and prevents potential confusion in error handling.


camilamacedo86 (approved):
/lgtm
/approved
/ok-to-test

PR #4660, opened 2025-03-26
🌱 (chore): avoid shadowing of 'config', 'err', and patch error in helm and deploy-image scaffolds

kersten:
Renamed the `config` parameter to `cfg` in Helm and DeployImage plugin scaffolds to avoid
shadowing the imported package. Also replaced short variable declarations with assignments
for `err` and introduced `errPatch` for clearer separation of error contexts.

Additionally, updated a call to `util.InsertCode` in the DeployImage scaffolder to ensure
a proper newline is injected before `env:`, improving the correctness and formatting
of the generated manifest.


camilamacedo86 (approved):
/lgtm
/approved
/ok-to-test

PR #4661, opened 2025-03-26
🌱 (chore): enable shadow checks via import-shadowing linter

kersten:
Now that all variable shadowing issues have been resolved in previous commits, the `shadow` linter in govet is enabled and `import-shadowing` rule from revive is activated, to ensure consistent detection of future shadowing problems.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,7 +21,6 @@ linters-settings:
     enable-all: true
     disable:
       - fieldalignment
-      - shadow
   nolintlint:
     allow-unused: false
   revive:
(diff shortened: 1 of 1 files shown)


.golangci.yml
```diff
@@ -41,6 +40,7 @@ linters-settings:
       - name: error-naming
       - name: exported
       - name: if-return
+      - name: import-shadowing
```
camilamacedo86, line 43: We probably should enable it for the scaffolds as well.

kersten: Great idea, let me check that out

camilamacedo86:
We will need to get merged all PRs first, 
Then, rebase it with master then we can move forward with this one.

camilamacedo86 (approved):
/lgtm
/approved
/ok-to-test

PR #4673, opened 2025-03-26
✨ (go/v4): Add new makefile target to create a cluster name to run the e2e tests

kersten:
change default Kind cluster names to be project-specific for e2e tests

Previously, the default Kind cluster name was "kind", which could lead to conflicts when running multiple e2e tests concurrently across different projects. This change updates the Kind cluster name to be project-specific, reducing potential conflicts and improving the isolation of e2e test environments.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/.github/workflows/test-e2e.yml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/.github/workflows/test-e2e.yml
@@ -26,9 +26,6 @@ jobs:
       - name: Verify kind installation
         run: kind version
 
-      - name: Create kind cluster
-        run: kind create cluster
-
       - name: Running Test e2e
         run: |
           go mod tidy
--- a/docs/book/src/cronjob-tutorial/testdata/project/Makefile
+++ b/docs/book/src/cronjob-tutorial/testdata/project/Makefile
@@ -69,17 +69,24 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
 # The default setup assumes Kind is pre-installed and builds/loads the Manager Docker image locally.
 # CertManager is installed by default; skip with:
 # - CERT_MANAGER_INSTALL_SKIP=true
-.PHONY: test-e2e
-test-e2e: manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
+KIND_CLUSTER ?= project-test-e2e
+
+.PHONY: setup-test-e2e
+setup-test-e2e: ## Set up a Kind cluster for e2e tests if it does not exist
 	@command -v $(KIND) >/dev/null 2>&1 || { \
 		echo "Kind is not installed. Please install Kind manually."; \
 		exit 1; \
 	}
-	@$(KIND) get clusters | grep -q 'kind' || { \
-		echo "No Kind cluster is running. Please start a Kind cluster before running the e2e tests."; \
-		exit 1; \
-	}
-	go test ./test/e2e/ -v -ginkgo.v
+	$(KIND) create cluster --name $(KIND_CLUSTER)
+
+.PHONY: test-e2e
+test-e2e: setup-test-e2e manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
+	KIND_CLUSTER=$(KIND_CLUSTER) go test ./test/e2e/ -v -ginkgo.v
+	$(MAKE) cleanup-test-e2e
+
+.PHONY: cleanup-test-e2e
+cleanup-test-e2e: ## Tear down the Kind cluster used for e2e tests
+	@$(KIND) delete cluster --name $(KIND_CLUSTER)
 
 .PHONY: lint
 lint: golangci-lint ## Run golangci-lint linter
--- a/docs/book/src/getting-started/testdata/project/.github/workflows/test-e2e.yml
+++ b/docs/book/src/getting-started/testdata/project/.github/workflows/test-e2e.yml
@@ -26,9 +26,6 @@ jobs:
       - name: Verify kind installation
         run: kind version
 
-      - name: Create kind cluster
-        run: kind create cluster
-
       - name: Running Test e2e
         run: |
           go mod tidy
--- a/docs/book/src/getting-started/testdata/project/Makefile
+++ b/docs/book/src/getting-started/testdata/project/Makefile
@@ -65,17 +65,24 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
 # The default setup assumes Kind is pre-installed and builds/loads the Manager Docker image locally.
 # CertManager is installed by default; skip with:
 # - CERT_MANAGER_INSTALL_SKIP=true
-.PHONY: test-e2e
-test-e2e: manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
+KIND_CLUSTER ?= project-test-e2e
+
+.PHONY: setup-test-e2e
+setup-test-e2e: ## Set up a Kind cluster for e2e tests if it does not exist
 	@command -v $(KIND) >/dev/null 2>&1 || { \
 		echo "Kind is not installed. Please install Kind manually."; \
 		exit 1; \
 	}
-	@$(KIND) get clusters | grep -q 'kind' || { \
-		echo "No Kind cluster is running. Please start a Kind cluster before running the e2e tests."; \
-		exit 1; \
-	}
-	go test ./test/e2e/ -v -ginkgo.v
+	$(KIND) create cluster --name $(KIND_CLUSTER)
+
+.PHONY: test-e2e
+test-e2e: setup-test-e2e manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
+	KIND_CLUSTER=$(KIND_CLUSTER) go test ./test/e2e/ -v -ginkgo.v
+	$(MAKE) cleanup-test-e2e
+
+.PHONY: cleanup-test-e2e
+cleanup-test-e2e: ## Tear down the Kind cluster used for e2e tests
+	@$(KIND) delete cluster --name $(KIND_CLUSTER)
 
 .PHONY: lint
 lint: golangci-lint ## Run golangci-lint linter
--- a/docs/book/src/multiversion-tutorial/testdata/project/.github/workflows/test-e2e.yml
+++ b/docs/book/src/multiversion-tutorial/testdata/project/.github/workflows/test-e2e.yml
@@ -26,9 +26,6 @@ jobs:
       - name: Verify kind installation
         run: kind version
 
-      - name: Create kind cluster
-        run: kind create cluster
-
       - name: Running Test e2e
         run: |
           go mod tidy
--- a/docs/book/src/multiversion-tutorial/testdata/project/Makefile
+++ b/docs/book/src/multiversion-tutorial/testdata/project/Makefile
@@ -69,17 +69,24 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
 # The default setup assumes Kind is pre-installed and builds/loads the Manager Docker image locally.
 # CertManager is installed by default; skip with:
 # - CERT_MANAGER_INSTALL_SKIP=true
-.PHONY: test-e2e
-test-e2e: manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
+KIND_CLUSTER ?= project-test-e2e
+
+.PHONY: setup-test-e2e
+setup-test-e2e: ## Set up a Kind cluster for e2e tests if it does not exist
 	@command -v $(KIND) >/dev/null 2>&1 || { \
 		echo "Kind is not installed. Please install Kind manually."; \
 		exit 1; \
 	}
-	@$(KIND) get clusters | grep -q 'kind' || { \
-		echo "No Kind cluster is running. Please start a Kind cluster before running the e2e tests."; \
-		exit 1; \
-	}
-	go test ./test/e2e/ -v -ginkgo.v
+	$(KIND) create cluster --name $(KIND_CLUSTER)
+
+.PHONY: test-e2e
+test-e2e: setup-test-e2e manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
+	KIND_CLUSTER=$(KIND_CLUSTER) go test ./test/e2e/ -v -ginkgo.v
+	$(MAKE) cleanup-test-e2e
+
+.PHONY: cleanup-test-e2e
+cleanup-test-e2e: ## Tear down the Kind cluster used for e2e tests
+	@$(KIND) delete cluster --name $(KIND_CLUSTER)
 
 .PHONY: lint
 lint: golangci-lint ## Run golangci-lint linter
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/github/test-e2e.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/github/test-e2e.go
@@ -28,6 +28,7 @@ var _ machinery.Template = &E2eTestCi{}
 type E2eTestCi struct {
 	machinery.TemplateMixin
 	machinery.BoilerplateMixin
+	machinery.ProjectNameMixin
 }
 
 // SetTemplateDefaults implements machinery.Template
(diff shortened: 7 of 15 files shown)


docs/book/src/cronjob-tutorial/testdata/project/.github/workflows/test-e2e.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         run: kind version
 
       - name: Create kind cluster
-        run: kind create cluster
+        run: kind create cluster --name project-test-e2e
```
camilamacedo86, line 30: Hey! Thanks for the PR 🙌

I don't think we should make this change directly here.

If we want to create a cluster name, I think it would make more sense to handle that inside something like `setup-test-e2e`. Then `test-e2e` can call `setup-test-e2e` as needed.

So the change here would either:
- Remove the `Create kind cluster` step, **or**
- Replace it with a call to `setup-test-e2e`

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -69,16 +69,19 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
...
 		echo "Kind is not installed. Please install Kind manually."; \
 		exit 1; \
 	}
-	@$(KIND) get clusters | grep -q 'kind' || { \
-		echo "No Kind cluster is running. Please start a Kind cluster before running the e2e tests."; \
-		exit 1; \
+	@$(KIND) get clusters | grep -q 'project-e2e-test' || { \
+		echo "Kind cluster not running, creating Kind cluster: project-e2e-test"; \
+		$(KIND) create cluster --name project-e2e-test; \
 	}
```
camilamacedo86, line 81: It should be part of the setup-test-e2e 
we would need to ensure that the kind is up and running, then create a new cluster called X to run the tests

docs/book/src/cronjob-tutorial/testdata/project/.github/workflows/test-e2e.yml
kersten: Yes, agreed - removing it seems like the better option. Other users might rely on different tools, like GitLab, so keeping cluster setup inside `setup-test-e2e` (which is already a dependency of `test-e2e`) makes things more flexible and smoother for everyone.

docs/book/src/cronjob-tutorial/testdata/project/Makefile
kersten: I renamed the target to `setup-test-e2e` and also all other occurrences of `e2e-test` -> `test-e2e`.

camilamacedo86 (commented):
Hi @kersten 

I think the change looks good! We just need to make sure to clean up the cluster after the tests run.

Also, I hope you don’t mind—I updated the title of the PR to make it more suitable for the release notes.

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -69,16 +69,19 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
...
 	@command -v $(KIND) >/dev/null 2>&1 || { \
 		echo "Kind is not installed. Please install Kind manually."; \
 		exit 1; \
 	}
-	@$(KIND) get clusters | grep -q 'kind' || { \
-		echo "No Kind cluster is running. Please start a Kind cluster before running the e2e tests."; \
-		exit 1; \
+	@$(KIND) get clusters | grep -q 'project-test-e2e' || { \
+		echo "Kind cluster not running, creating Kind cluster: project-test-e2e"; \
+		$(KIND) create cluster --name project-test-e2e --wait 5m; \
```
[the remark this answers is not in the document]
kersten: Great idea! We already have some cleanup logic in our code, I’ll add a dedicated target for it as we have. What do you think - should we clean it up automatically after each test, or just provide a target so it can be run manually? On our side, we usually clean it up manually to avoid long spin-up times during development.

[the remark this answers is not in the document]
camilamacedo86: I think we will need something like

```
.PHONY: kind-clean
kind-clean: 
	$(KIND) delete cluster --name $(KIND_CLUSTER_NAME)
```

and call it at the end of the tests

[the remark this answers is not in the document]
kersten: I've added that already. And the variable was already requested in this PR: https://github.com/kubernetes-sigs/kubebuilder/pull/4617.

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -69,17 +69,32 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
...
 # - CERT_MANAGER_INSTALL_SKIP=true
+.PHONY: setup-test-e2e
+setup-test-e2e: ## Set up a Kind cluster for e2e tests if it does not exist
+	@command -v $(KIND) >/dev/null 2>&1 || { \
+		echo "Kind is not installed. Please install Kind manually."; \
+		exit 1; \
+	}
+	@$(KIND) get clusters | grep -q 'project-test-e2e' || { \
+		echo "Kind cluster not running, creating Kind cluster: project-test-e2e"; \
+		$(KIND) create cluster --name project-test-e2e --wait 5m; \
```
camilamacedo86, line 80:
```suggestion
	$(KIND) create cluster --name $(KIND_CLUSTER_NAME)  --wait 5m; \
```

@kersten sorry I might not be very clear.
I think that
a) above we would need to have only the line to create the cluster ( not to check if it exist )
b) use an variable for we store the name of the cluster 


Then, after we run the tests

we call a target test-e2e-tweardown or cleanup something like that will delete the cluster name

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -69,17 +69,32 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
...
 # CertManager is installed by default; skip with:
 # - CERT_MANAGER_INSTALL_SKIP=true
+.PHONY: setup-test-e2e
+setup-test-e2e: ## Set up a Kind cluster for e2e tests if it does not exist
+	@command -v $(KIND) >/dev/null 2>&1 || { \
+		echo "Kind is not installed. Please install Kind manually."; \
+		exit 1; \
+	}
+	@$(KIND) get clusters | grep -q 'project-test-e2e' || { \
+		echo "Kind cluster not running, creating Kind cluster: project-test-e2e"; \
```
camilamacedo86, line 79:
```suggestion
```
Could we remove it?
It has no reason to check if the cluster exists.
If we try to create and itexistst then, the command will fail, so we can try to simplify the syntax and the code

kersten: Ok, removed the check.

docs/book/src/cronjob-tutorial/testdata/project/Makefile
kersten: For the envvar, I would rather wait for the PR #4617, or would you like me to merge the changes into this one?

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -69,17 +69,26 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
...
-test-e2e: manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
+.PHONY: setup-test-e2e
+setup-test-e2e: ## Set up a Kind cluster for e2e tests if it does not exist
 	@command -v $(KIND) >/dev/null 2>&1 || { \
 		echo "Kind is not installed. Please install Kind manually."; \
 		exit 1; \
 	}
-	@$(KIND) get clusters | grep -q 'kind' || { \
-		echo "No Kind cluster is running. Please start a Kind cluster before running the e2e tests."; \
+	$(KIND) create cluster --name project-test-e2e --wait 5m
```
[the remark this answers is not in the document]
camilamacedo86: Also, do we really need wait 5m ? 
Why would it be required?

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -69,17 +69,26 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
...
+test-e2e: setup-test-e2e manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
+	go test ./test/e2e/ -v -ginkgo.v
+	$(MAKE) teardown-test-e2e
+
+.PHONY: teardown-test-e2e
+teardown-test-e2e: ## Tear down the Kind cluster used for e2e tests
+	@command -v $(KIND) >/dev/null 2>&1 || { \
+		echo "Kind is not installed. Please install Kind manually."; \
 		exit 1; \
 	}
```
camilamacedo86, line 90:
```suggestion
```
We already checked it to create the cluster, do we need to check if the kind exists to delete the cluster?
Could we not simplify the code by making it less verbose?

camilamacedo86 (changes requested):
Hi @kersten,

I'm OK with the proposed changes — I think it's a good addition to the default scaffolds.

I just added a couple of minor suggestions:

a) Use `KIND_CLUSTER_NAME` to parameterize the cluster name used in the tests
b) Simplify the code by removing the extra check

Thanks for the contribution! 🎉

docs/book/src/cronjob-tutorial/testdata/project/Makefile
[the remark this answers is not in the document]
kersten: It doesn't wait at least 5 minutes. As soon as the control plane is ready, it proceeds immediately. We've observed that in rare cases, kind sets up so quickly that deployments to the cluster can fail, causing the test to fail as well. I'd suggest starting with a 5-minute wait. If that turns out to be too short or too long, we can make it configurable later.

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -69,17 +69,26 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
...
-	go test ./test/e2e/ -v -ginkgo.v
+	@$(KIND) delete cluster --name project-test-e2e
```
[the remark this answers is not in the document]
kersten: Hey, could you take a look at this? https://github.com/kubernetes-sigs/kubebuilder/pull/4617

I feel like we might be talking past each other - should I include that change in this PR, or do we want to merge them separately, one after the other?

docs/book/src/cronjob-tutorial/testdata/project/Makefile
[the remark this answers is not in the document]
camilamacedo86: What do you think about starting without a timeout for now, and only adding one later if we run into issues or have a clear reason for it?

For context, I’ve never needed to use a timeout in similar projects so far, so it seems unlikely we’ll need one here either. And in case something does go wrong, it might be better to fail fast and return the error immediately rather than wait up to 5 minutes.

This way, we keep things simple and avoid premature optimization, while still leaving room to adjust if needed later on.

docs/book/src/cronjob-tutorial/testdata/project/Makefile
camilamacedo86: Could we not do all here and keep a full proposal?

kersten: Sure, I will do it later. Probably tonight.

docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
```diff
@@ -167,9 +167,9 @@ func IsCertManagerCRDsInstalled() bool {
 
 // LoadImageToKindClusterWithName loads a local docker image to the kind cluster
 func LoadImageToKindClusterWithName(name string) error {
-	cluster := "kind"
-	if v, ok := os.LookupEnv("KIND_CLUSTER"); ok {
-		cluster = v
+	cluster, ok := os.LookupEnv("KIND_CLUSTER")
+	if !ok {
+		return fmt.Errorf("KIND_CLUSTER is not set")
```
camilamacedo86, line 172: Do we need this change?

kersten: It would have worked with the old code, but since we no longer have a default `kind` cluster, it’s not necessary anymore. However, if the environment variable isn’t set, the image deployment will fail. I’d feel more comfortable if we explicitly notify the user when it's missing.

camilamacedo86: I will need some time to test this one.
If I just have kind running ( any cluster name ) and run the e2e tests, will that work fine? 
If so, I am OK with if not I think we should allow it.

kersten: I get your point. I think you will start the e2e test right out of your IDE. I will have to think about that and implement it. Currently that wouldn't work out of the box. You would have to provide the ENV var in the IDE launch config. I will provide a way, where it defaults to `kind`.

camilamacedo86: > Currently that wouldn't work out of the box

I need to check it, but I think what you are saying today does not work if we have any cluster name created right?

camilamacedo86: So, I think we should revert the change in the utils for now.
If needed, we can revisit and improve or refactor it in a separate PR since it feels a bit out of scope for this one.

That way, we can move forward with the changes in the Makefile without blocking.
WDYT?

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -69,17 +69,24 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
...
-	}
-	go test ./test/e2e/ -v -ginkgo.v
+	$(KIND) create cluster --name $(KIND_CLUSTER)
+
+.PHONY: test-e2e
+test-e2e: setup-test-e2e manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
+	KIND_CLUSTER=$(KIND_CLUSTER) go test ./test/e2e/ -v -ginkgo.v
+	$(MAKE) teardown-test-e2e
+
+.PHONY: teardown-test-e2e
```
camilamacedo86, line 87: After reviewing possible naming options, it seems that cleanup-test-e2e is the most suitable choice.

It aligns well with the existing setup-test-e2e and test-e2e targets, keeping the naming consistent and intuitive. It also clearly communicates its purpose without being overly verbose or ambiguous.

Let me know if you'd prefer a different naming style, but this feels like the most balanced option for now.

camilamacedo86 (changes requested):
Hi @kersten 👋 

Thank you so much for this change, and sorry for the delay in reviewing it!

After going through everything, I believe we're definitely good to move forward — just two small nitpicks to address before merging:

1. **Revert the unnecessary change in the `scaffolds` method under `utils`**, so that for now we keep only the Makefile improvements.  
   👉 See: https://github.com/kubernetes-sigs/kubebuilder/pull/4673/files#r2030095359

2. **Consider renaming the e2e cleanup test target to `cleanup-test-e2e`** instead of `teardown-test-e2e`.  
   I think it would simplify the name and align better with the existing naming conventions.  
   👉 See: https://github.com/kubernetes-sigs/kubebuilder/pull/4673/files#r2030095626

Thanks again for the contribution — really appreciate the effort here! 🙌

docs/book/src/reference/envtest.md
```diff
@@ -315,9 +315,9 @@ func InstallCertManager() error {
 
 // LoadImageToKindClusterWithName loads a local docker image to the kind cluster
 func LoadImageToKindClusterWithName(name string) error {
-	cluster := "kind"
-	if v, ok := os.LookupEnv("KIND_CLUSTER"); ok {
-		cluster = v
+	cluster, ok := os.LookupEnv("KIND_CLUSTER")
+	if !ok {
+		return fmt.Errorf("KIND_CLUSTER is not set")
```
camilamacedo86, line 320: Hey! Just wondering — could we not keep it as before, so that if KIND_CLUSTER isn’t set, it still works with the default kind context? We could revisit this in a follow-up if needed.

WDYT? Just trying to keep it working out of the box for folks not setting that env var. 🙏

camilamacedo86 (approved):
It seems great for me 🥇 

/lgtm

PR #4675, opened 2025-03-26
⚠️ 🐛 (API) convert CLI methods to use pointer receivers

kersten:
Updated all CLI method receivers to use pointer semantics `(*CLI)` instead of value `(CLI)`, ensuring consistent behavior across command setup functions and reducing potential for unintentional value copying.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -449,7 +449,7 @@ func (c *CLI) addExtraCommands() error {
 }
 
 // printDeprecationWarnings prints the deprecation warnings of the resolved plugins.
-func (c CLI) printDeprecationWarnings() {
+func (c *CLI) printDeprecationWarnings() {
 	for _, p := range c.resolvedPlugins {
 		if p != nil && p.(plugin.Deprecated) != nil && len(p.(plugin.Deprecated).DeprecationWarning()) > 0 {
 			_, _ = fmt.Fprintf(os.Stderr, noticeColor, fmt.Sprintf(deprecationFmt, p.(plugin.Deprecated).DeprecationWarning()))
@@ -458,7 +458,7 @@ func (c CLI) printDeprecationWarnings() {
 }
 
 // metadata returns CLI's metadata.
-func (c CLI) metadata() plugin.CLIMetadata {
+func (c *CLI) metadata() plugin.CLIMetadata {
 	return plugin.CLIMetadata{
 		CommandName: c.commandName,
 	}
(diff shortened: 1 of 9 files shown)


camilamacedo86 (approved):
/lgtm
/approved
/ok-to-test

pkg/cli/cli.go
```diff
@@ -467,11 +467,11 @@ func (c CLI) metadata() plugin.CLIMetadata {
...
 //
 // If an error is found, command help and examples will be printed.
-func (c CLI) Run() error {
+func (c *CLI) Run() error {
 	return c.cmd.Execute()
 }
 
 // Command returns the underlying root command.
-func (c CLI) Command() *cobra.Command {
+func (c *CLI) Command() *cobra.Command {
```
camilamacedo86, line 475: Hi @kersten 

I think I would be possible to have a function as before without the receiver deprecated so that we do not break the API? And in the comment we suggested use  instead `(c *CLI) Command()`


/hold 


Lets see if we can move with this change without break the API

kersten: WDYT about this approach?

```go
// Command returns the underlying root command.
//
// Deprecated: Use (c *CLI).CommandPtr() instead.
func (c CLI) Command() *cobra.Command {
	return (&c).CommandPtr()
}

// CommandPtr returns the underlying root command.
func (c *CLI) CommandPtr() *cobra.Command {
	return c.cmd
}
```

This keeps the existing `Command()` method for compatibility and introduces the pointer-based version under a new name. In a future major release, we could switch `Command()` to use the pointer receiver and deprecate `CommandPtr()` to clean things up.

Let me know what you think!

camilamacedo86: The issue with this one is that it would break the API.

Deprecating it isn’t a great option either, since we’d be left with an awkward name (CommandPtr) that we’d have to support indefinitely. It feels like an odd choice, so I’m considering how much value this actually brings compared to the potential costs and impact—especially for end users who rely on Kubebuilder as a library.

I’ll need some time to properly review and evaluate this case before deciding how to proceed.

camilamacedo86 (changes requested):
Hi @kersten,

Thanks again for the contribution and for exploring alternatives! 🎉 

After further review, we've decided to defer this change to [v5](https://github.com/kubernetes-sigs/kubebuilder/issues/4762), where breaking changes can be handled more cleanly. While the workaround avoids immediate breakage, it introduces naming and maintenance concerns that we’d prefer to address in a major release. See: https://github.com/kubernetes-sigs/kubebuilder/issues/4762

I hope that you do not mind:

Closing this for now as deferred — really appreciate your efforts and understanding!

Best,

PR #4676, opened 2025-03-26
📖 Updated migration pages 

Kuzuri247:
Added index for better viewing of the steps and removed the redundant lines and whitespaces

### DIFF
--- a/docs/book/src/migration/migration_guide_gov3_to_gov4.md
+++ b/docs/book/src/migration/migration_guide_gov3_to_gov4.md
@@ -126,7 +126,7 @@ Now, let's copy the webhook definition from `api/v1/<kind>_webhook.go` from our 
 
 ## Others
 
-If there are any manual updates in `main.go` in v3, we need to port the changes to the new `main.go`. We’ll also need to ensure all of needed controller-runtime `schemes` have been registered.
+If there are any manual updates in `main.go` in v3, we need to port the changes to the new `main.go`. We'll also need to ensure all of needed controller-runtime `schemes` have been registered.
 
 If there are additional manifests added under config directory, port them as well. Please, be aware that
 the new version go/v4 uses Kustomize v5x and no longer Kustomize v4. Therefore, if added customized
--- a/docs/book/src/migration/v3vsv4.md
+++ b/docs/book/src/migration/v3vsv4.md
@@ -24,7 +24,7 @@ The details of all changes (breaking or otherwise) can be found in:
 by moving the api(s) under a new directory called `api`, controller(s) under a new directory called `internal` and the `main.go` under a new directory named `cmd`
 
 <aside class="note">
-<H1> TL;DR of the New `go/v4` Plugin </H1>
+<H1> Details of go/v4 Plugin </H1>
 
 Further details can be found in the [go/v4 plugin section][go/v4-doc]
 
--- a/docs/book/src/migration/v3-plugins.md
+++ b/docs/book/src/migration/v3-plugins.md
@@ -6,6 +6,3 @@ since `28 Apr 2021`.
 
 Therefore, you can check here how to migrate the projects built from Kubebuilder 3.x with
 the plugin go/v3 to the latest.
-
-
-


docs/book/src/migration/v3vsv4.md
```diff
@@ -24,7 +24,7 @@ The details of all changes (breaking or otherwise) can be found in:
 by moving the api(s) under a new directory called `api`, controller(s) under a new directory called `internal` and the `main.go` under a new directory named `cmd`
 
 <aside class="note">
-<H1> TL;DR of the New `go/v4` Plugin </H1>
```
Kavinjsir, line 27: I think the intention here is a section of `TLDR` to introduce what's new in `v4` compared to `v3`.
So how about keeping the header as `TL;DR` and adding a list new stuff bringing by `v4`.

docs/book/src/migration/migration_guide_gov3_to_gov4.md
```diff
@@ -152,6 +171,16 @@ Change the image name in the Makefile if needed.
 Finally, we can run `make` and `make docker-build` to ensure things are working
 fine.
 
+<aside class="note">
```
Kavinjsir, line 174: I personally feels it might not necessary considering the case where users could gain experience already when working with v3.

Kavinjsir (changes requested):

docs/book/src/migration/migration_guide_gov3_to_gov4.md
camilamacedo86: +1

docs/book/src/migration/migration_guide_gov3_to_gov4.md
```diff
@@ -1,5 +1,24 @@
...
+## Table of Contents
+
+- [Prerequisites](#prerequisites)
+- [Initial Setup](#initialize-a-gov4-project)
+- [API and Controller Migration](#migrate-apis-and-controllers)
+- [Webhook Migration](#migrate-the-webhooks)
+- [Additional Changes](#others)
+- [Verification](#verification)
+
+## Prerequisites
```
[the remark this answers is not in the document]
camilamacedo86: +1

docs/book/src/migration/v3vsv4.md
camilamacedo86: Yep, I think it is a note to clarify detailed info so that is why we use TL'DR: 
 I am not strongly opinionated either about it, but I think we can keep it as it is.

Kuzuri247: TLDR should after the context and in subheader it's being repeated that's why I after the header gives the context and after small description TLDR is being used

docs/book/src/migration/migration_guide_gov3_to_gov4.md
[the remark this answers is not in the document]
Kuzuri247: I added this for the flow if you guys don't like it I will remove this @camilamacedo86 @Kavinjsir

Kuzuri247:
Did the necessary changes @camilamacedo86 @Kavinjsir

docs/book/src/migration/migration_guide_gov3_to_gov4.md
```diff
@@ -1,5 +1,13 @@
 # Migration from go/v3 to go/v4
 
+## Table of Contents
+
+- [Initial Setup](#initialize-a-gov4-project)
+- [API and Controller Migration](#migrate-apis-and-controllers)
+- [Webhook Migration](#migrate-the-webhooks)
+- [Additional Changes](#others)
+- [Verification](#verification)
```
camilamacedo86, line 9: Does it make sense for us to follow the migration guide?
Note that the guide shows how we handle it by re-generating the project, while the other approach involves manual changes. So it feels a bit odd to include the table of contents here but not in the other one. I'm also wondering if we really need the TOC at all.

docs/book/src/migration/v3vsv4.md
camilamacedo86: I think the main goal of the note is to say, *“Here are the details,”* right? Keeping only the title `New go/v4 Plugin` doesn't quite convey that. Maybe we should replace it with something like **Details of `go/v4` Plugin** or something along those lines?

docs/book/src/migration/migration_guide_gov3_to_gov4.md
Kuzuri247: So I can add the toc in the manual one as well . I thought toc for guide could be good . What should I do.

camilamacedo86: I'm thinking the step-by-step format works better without an index, since each step builds on the previous one. For example, we can't set up the Webhook without first completing the API or initial setup, so listing everything in an index might not add much value.

Overall, I feel it's better to keep it the way it is now—simple and not too verbose.
@Kavinjsir @varshaprasad96 wdyt?

docs/book/src/migration/v3vsv4.md
Kuzuri247: Yeah that should do it

Kuzuri247:
I will replace with the title with Details of go/v4 Plugin
@Kavinjsir What about the index

docs/book/src/migration/migration_guide_gov3_to_gov4.md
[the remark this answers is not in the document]
Kavinjsir: @Kuzuri247 probably no need to have prerequisites here, you can remove it to make the section cleaner here.

docs/book/src/migration/migration_guide_gov3_to_gov4.md
Kavinjsir: @camilamacedo86 agree.

Kuzuri247:
Did the necessary changes its ready for merging

Kuzuri247:
@camilamacedo86

camilamacedo86 (approved):
/lgtm
/approve
/ok-to-test

PR #4695, opened 2025-03-27
 📖 Fix path for samples in the Getting Started tutorial

fferrann:
Fix incorrect location of the samples folder.

Currently the documentation suggests that the location for the samples is config/crd/samples, but from what I can see when I create my getting-started project it's under config/samples.

image


camilamacedo86 (approved):
/lgtm
/approved
/ok-to-test

PR #4696, opened 2025-03-27
🌱 (chore): fix improper usage of `errors.As` in scaffold tests

kersten:
This fixes a misuse of `errors.As` where a non-pointer was passed as the second argument. Go requires the second argument to be a pointer to an interface or a type implementing the `error` interface.

The test helper now correctly uses `&errType` to ensure proper error assertion and avoid false negatives or runtime issues.


camilamacedo86 (approved):
/ok-to-test
/approve
/lgtm

PR #4701, opened 2025-03-28
🌱  (chore): refactor unit tests to isolate mutable test data in CLI spec to improve maintainability 

kersten:
### Isolate mutable test data in CLI spec to avoid Ginkgo pollution

This change moves mutable shared test data such as `projectVersion` and plugin key lists into `BeforeEach` blocks to ensure test isolation within Ginkgo. 

Previously, these variables were declared and modified in the outer scope of the `Describe` blocks, which could lead to pollution across tests due to shared state.

This improves test reliability and maintainability by making each test's state self-contained.

No behavior change.


camilamacedo86 (commented):
/ok-to-test

PR #4702, opened 2025-03-28
🌱  (chore): refactor unit tests to isolate mutable state with BeforeEach for pkg/cli

kersten:
### 🌱 (chore): Move mutable test declarations into `BeforeEach` to ensure clean test state

This refactors the `options_test.go` spec to define mutable test variables like `projectVersion`, plugin mocks, and env paths inside `BeforeEach` blocks. This avoids potential side effects and improves Ginkgo test isolation.

Previously, some of these variables were declared in outer scopes and could accidentally retain or leak state between examples.

This change improves test clarity and maintainability without impacting behavior.

### DIFF
--- a/pkg/cli/resource_test.go
+++ b/pkg/cli/resource_test.go
@@ -32,6 +32,12 @@ var _ = Describe("resourceOptions", func() {
 	)
 
 	var (
+		fullGVK     resource.GVK
+		noDomainGVK resource.GVK
+		noGroupGVK  resource.GVK
+	)
+
+	BeforeEach(func() {
 		fullGVK = resource.GVK{
 			Group:   group,
 			Domain:  domain,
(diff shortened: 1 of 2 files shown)


pkg/cli/resource_test.go
```diff
@@ -68,13 +74,14 @@ var _ = Describe("resourceOptions", func() {
...
-			func(options resourceOptions) {
+			func(getOpts func() resourceOptions) {
+				options := getOpts()
+
 				Expect(options.validate()).To(Succeed())
 
 				resource := options.newResource()
 				Expect(resource.Validate()).To(Succeed())
 				Expect(resource.GVK.IsEqualTo(options.GVK)).To(BeTrue())
-				// Plural is checked in the next test
```
camilamacedo86, line 77: Can we keep the comment — or better yet, use the testing framework's syntax by writing something like By("Checking plural...")? I think that would make the intent clearer and more consistent with the rest of the test code.

PR #4711, opened 2025-03-28
🌱 (chore): prevent variable shadowing in DecodePluginConfig test to ensure accurate assertion

kersten:
### 🌱 (chore): DecodePluginConfig test variable reuse

This PR fixes potential shadowing in table-driven tests by renaming `pluginCfg` to `decodedPluginCfg`. This avoids confusion between the decoded result and the expected struct, ensuring the assertions are valid and that Gomega receives the correct references.

This improves test clarity and avoids subtle bugs due to variable reuse within closures.

### DIFF
--- a/pkg/config/v3/config_test.go
+++ b/pkg/config/v3/config_test.go
@@ -405,14 +405,12 @@ var _ = Describe("Cfg", func() {
 		})
 
 		It("DecodePluginConfig should fail for no plugin config object", func() {
-			var pluginCfg PluginConfig
 			err := c0.DecodePluginConfig(key, &pluginCfg)
 			Expect(err).To(HaveOccurred())
 			Expect(errors.As(err, &config.PluginKeyNotFoundError{})).To(BeTrue())
 		})
 
 		It("DecodePluginConfig should fail to retrieve data from a non-existent plugin", func() {
-			var pluginCfg PluginConfig
 			err := c1.DecodePluginConfig("plugin-y", &pluginCfg)
 			Expect(err).To(HaveOccurred())
 			Expect(errors.As(err, &config.PluginKeyNotFoundError{})).To(BeTrue())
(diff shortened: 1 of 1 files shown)


pkg/config/v3/config_test.go
```diff
@@ -405,23 +405,23 @@ var _ = Describe("Cfg", func() {
 		})
 
 		It("DecodePluginConfig should fail for no plugin config object", func() {
-			var pluginCfg PluginConfig
-			err := c0.DecodePluginConfig(key, &pluginCfg)
+			var decodedPluginCfg PluginConfig
+			err := c0.DecodePluginConfig(key, &decodedPluginCfg)
 			Expect(err).To(HaveOccurred())
 			Expect(errors.As(err, &config.PluginKeyNotFoundError{})).To(BeTrue())
```
nunnatsa, line 411: Please Consider using the gomega's `MatchError` matcher.

pkg/config/v3/config_test.go
```diff
@@ -405,23 +405,23 @@ var _ = Describe("Cfg", func() {
...
 		})
 
 		It("DecodePluginConfig should fail to retrieve data from a non-existent plugin", func() {
-			var pluginCfg PluginConfig
-			err := c1.DecodePluginConfig("plugin-y", &pluginCfg)
+			var decodedPluginCfg PluginConfig
+			err := c1.DecodePluginConfig("plugin-y", &decodedPluginCfg)
 			Expect(err).To(HaveOccurred())
 			Expect(errors.As(err, &config.PluginKeyNotFoundError{})).To(BeTrue())
```
nunnatsa, line 418: Same

pkg/config/v3/config_test.go
kersten: @nunnatsa You're absolutely right - this should be addressed. I'd prefer to keep it out of scope for this PR though, as it's unrelated to the main change. There are a few places in the codebase where `errors.As` is used in the tests, so it would make sense to handle all of them together in a separate PR. WDYT?

PR #4713, opened 2025-03-29
🌱 Use dynamic controller-runtime version in doc generator to make easier update the version

### LINKED ISSUE #4679: Improve hack/docs to avoid failures when upgrading controller-runtime version
Currently, all projects generated for documentation/tutorials are created automatically via commands and scripts under `hack/docs`. This automation helps keep our documentation up to date without needing manual intervention.

When we run `make generate`, it triggers `make generate-docs`, which re-generates these example projects and syncs them with the documentation.

### The Problem

We have a hardcoded version of `controller-runtime` in the following line:

https://github.com/kubernetes-sigs/kubebuilder/blob/3ae96043f3673504f18a1e776cc32ee142deaf91/hack/docs/internal/cronjob-tutorial/generate_cronjob.go#L291

This causes `make generate-docs` to fail when we bump the controller-runtime version in the scaffold logic. For example, updating this constant:

https://github.com/kubernetes-sigs/kubebuilder/blob/3ae96043f3673504f18a1e776cc32ee142deaf91/pkg/plugins/golang/v4/scaffolds/init.go#L43

From:
ControllerRuntimeVersion = "v0.20.4"

To a newer version may result in errors like:

`ERRO error fixing cronjob_controller.go: string // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.20.3/pkg/reconcile not found in /* Copyright 2025 The Kubernetes authors.
`
### What Needs to Be Done

We should improve the logic in `hack/docs` so that it's not tied to a hardcoded controller-runtime version.

We could either:
- Remove the need for the manual replacement in the first place
- OR dynamically use the version from the `ControllerRuntimeVersion` constant in the scaffolding code

This would prevent future breakages when we upgrade dependencies and keep `make generate-docs` stable and version-aware.

p-shah256:
# 🐛 Fix hardcoded controller-runtime version in docs generation

Replace hardcoded controller-runtime version in `hack/docs/internal/cronjob-tutorial/generate_cronjob.go` with a reference to the `ControllerRuntimeVersion` constant from the scaffolding package.
This prevents the docs generation from breaking when the controller-runtime version is updated in the scaffolding logic, ensuring that `make generate-docs` remains stable across version bumps.

## Changes

- Import the scaffolds/init package to access the ControllerRuntimeVersion constant
- Replace the hardcoded version string with a dynamic reference using fmt.Sprintf
This change adheres to the DRY principle by having a single source of truth for the controller-runtime version, preventing future breakages when dependencies are updated.

## Fixed Issue
Fixes the issue where make generate-docs fails when controller-runtime is updated in only one location.
Fixes #4679

I'm open to any feedback or suggestions on how to improve this approach if there's a cleaner way to handle it!

### DIFF
--- a/hack/docs/internal/cronjob-tutorial/generate_cronjob.go
+++ b/hack/docs/internal/cronjob-tutorial/generate_cronjob.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cronjob
 
 import (
+	"fmt"
 	"os/exec"
 	"path/filepath"
 
@@ -24,6 +25,7 @@ import (
 	"github.com/spf13/afero"
 	hackutils "sigs.k8s.io/kubebuilder/v4/hack/docs/utils"
 	pluginutil "sigs.k8s.io/kubebuilder/v4/pkg/plugin/util"
+	"sigs.k8s.io/kubebuilder/v4/pkg/plugins/golang/v4/scaffolds"
 	"sigs.k8s.io/kubebuilder/v4/test/e2e/utils"
 )
 
@@ -288,7 +290,8 @@ func (sp *Sample) updateController() {
 
 	err = pluginutil.InsertCode(
 		filepath.Join(sp.ctx.Dir, "internal/controller/cronjob_controller.go"),
-		`// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.20.4/pkg/reconcile`, skipGoCycloLint)
+		fmt.Sprintf(`// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@%s/pkg/reconcile`,
+			scaffolds.ControllerRuntimeVersion), skipGoCycloLint)
 	hackutils.CheckError("fixing cronjob_controller.go", err)
 
 	err = pluginutil.ReplaceInFile(


hack/docs/internal/cronjob-tutorial/generate_cronjob.go
```diff
@@ -288,7 +290,8 @@ func (sp *Sample) updateController() {
 
 	err = pluginutil.InsertCode(
 		filepath.Join(sp.ctx.Dir, "internal/controller/cronjob_controller.go"),
-		`// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.20.4/pkg/reconcile`, skipGoCycloLint)
+		fmt.Sprintf(`// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@%s/pkg/reconcile`,
```
sbin64, line 293:
```suggestion
		fmt.Sprintf(`// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@%v/pkg/reconcile`,
```

sbin64 (changes requested):
small nit.

sbin64 (approved):
Very cool. LGTM! ㊗️

hack/docs/internal/cronjob-tutorial/generate_cronjob.go
p-shah256: Sounds good to me!

hack/docs/internal/cronjob-tutorial/generate_cronjob.go
```diff
@@ -288,7 +290,8 @@ func (sp *Sample) updateController() {
 
 	err = pluginutil.InsertCode(
 		filepath.Join(sp.ctx.Dir, "internal/controller/cronjob_controller.go"),
-		`// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.20.4/pkg/reconcile`, skipGoCycloLint)
+		fmt.Sprintf(`// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@%v/pkg/reconcile`,
```
kersten, line 293:
```suggestion
		fmt.Sprintf(`// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@%s/pkg/reconcile`,
```

Since `scaffolds.ControllerRuntimeVersion` is already a string, `%s` is more appropriate than `%v` - it's clearer and more explicit :)

camilamacedo86 (dismissed):
Thank you so much for this one! 👍
Everything looks great — we're just missing one last step: squashing the commits.

We follow a policy of having one commit per PR for each need/requirement/change. This helps us maintain a clean Git history and makes things easier when we need to revert or track changes.

Could you please squash the commits so we can get this one merged? 🙏

Thanks again!

camilamacedo86:
Thank you a lot for your contribution 🎉 

/lgtm
/approve
/ok-to-test

p-shah256:
> Thank you a lot for your contribution 🎉
> 
> /lgtm /approve /ok-to-test

Thanks! 🎉

PR #4716, opened 2025-03-29
📖 Fixed grammatical errors in designs doc

Kuzuri247:
Fixed multiple grammatical errors in the markdown files inside the designs folder and also corrected the numbering error


Kuzuri247:
@Kavinjsir @camilamacedo86

designs/template.md
```diff
@@ -2,8 +2,7 @@
 |---------------|---------------|-------------|---|
 | @name | date | Implementable | - |
 
-Title of the Design/Proposal
-===================
+# Title of the Design/Proposal
```
camilamacedo86, line 5: This one is proposing a change in template ideally should be in another PR
however, in this case as it is too small I think it is fine we move forward here.

PR #4723, opened 2025-03-31
🌱 (chore): replace errors.As() with MatchError() in error assertions

kersten:
This update improves Gomega assertions by replacing `errors.As()` with `MatchError()` where the test intends to match specific error types rather than extract them.

This change simplifies the assertions and improves readability, aligning with idiomatic Gomega style.

### DIFF
--- a/pkg/config/store/yaml/store_test.go
+++ b/pkg/config/store/yaml/store_test.go
@@ -18,6 +18,7 @@ package yaml
 
 import (
 	"errors"
+	"fmt"
 	"os"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


camilamacedo86 (approved):
/lgtm
/approve
/ok-to-test

camilamacedo86 (dismissed):
/approve cancel

Those changes broke the tests.
I am OK with any change for better syntax on the tests but we need to ensure that all still working well as before.

camilamacedo86:
Hi @kersten 

This one is still failing see:

Screenshot 2025-03-31 at 14 23 07

camilamacedo86:
I'm okay with the proposed changes — let's move forward!

However, if we start to see that it introduces complexity in terms of maintenance (e.g., things start breaking because updates require constant changes to the text, etc.), we can always revisit the approach.

For now, everything looks great.
Thanks! 🎉

camilamacedo86 (approved):
I'm okay with the proposed changes — let's move forward!

However, if we start to see that it introduces complexity in terms of maintenance (e.g., things start breaking because updates require constant changes to the text, etc.), we can always revisit the approach.

For now, everything looks great.
Thanks! 🎉

/lgtm
/approved
/ok-to-test

PR #4724, opened 2025-03-31
📖 Improve comments and fix typos for cronjob tutorial

chethanm99:
Updated the files of cronjob_tutorial with minor grammatical changes, particularly api_design.go , controller_implementation.go , e2e_implementation.go , generate_cronjob.go , main_revisited.go , sample.go , writing_tests_controller.go,  writing_tests_env.g to enhance readability and guide developers and contributors.


camilamacedo86 (approved):

camilamacedo86:
/lgtm cancel
/approve cancel

Please ensure the following steps are completed before we proceed with this PR:

- The proposed grammar improvements must be reflected in the hack docs and properly updated in the tutorials using make generate-docs.
- Run make generate-docs to ensure all sample projects are regenerated with the proposed changes.

Note: We generate the projects used in the tutorials, and then layer additional code through the hack/docs scripts.

We cannot move forward until these updates are properly applied.

Also, please squash your commits — this PR should contain only one commit for the final proposal.

Thank you!

camilamacedo86:
/approve false

This one is missing run `make generate` OR `make generate-docs`

camilamacedo86 (changes requested):
Need to run ` make generate-docs` to properly update the samples samples used in the docs with the changes.

chethanm99:
Hi @camilamacedo86 

Thanks for the guidance on running `make generate-docs`. I've been working through the errors and hit a blocker.

The command fails with:
`ERRO error removing spec explanation: unable to find the content to be replaced`

This happens in `hack/docs/internal/multiversion-tutorial/generate_multiversion.go` during the `updateAPIV1` step. It's trying to replace the large comment block defined in the `cronjobSpecComment` variable within `api/v1/cronjob_types.go`.

However, the source file (`docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_types.go`) only contains the single line `// CronJobSpec defines the desired state of CronJob.` before the struct definition, not the large block the script expects.

I've verified that my commits didn't modify this source file (`docs/.../api/v1/cronjob_types.go`). It seems the script is trying to replace text that doesn't exist in the source material.

Could you please advise on how to proceed? 

Thanks!

camilamacedo86:
> ERRO error removing spec explanation: unable to find the content to be replaced

That happened because the changes broke the sample generation process.

We automatically generate the samples under the docs/ folder, then insert, replace, or update code based on predefined patterns.

It looks like you modified a text string (probably a constant) that comes from the default scaffold. Because of that, the generator can no longer find the original text to replace. It could also be due to a customization applied on top of previously generated code.

In this way, you need to find where the error is faced ( what is the replace that is not done, and why the original text is no longer found ) .. Let us know if you need a help with.

chethanm99:
Hi @camilamacedo86 ,

Thanks, I understand the process. I've done the investigation you suggested ("find where the error is faced... and why the original text is no longer found").

The specific replace that fails is in `hack/docs/internal/multiversion-tutorial/generate_multiversion.go` trying to remove the large text block defined in `cronjobSpecComment` from `api/v1/cronjob_types.go`.

The reason the text isn't found isn't due to my modifications to that file (I verified that docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_types.go was untouched by my commits), but because the source file itself only contains "CronJobSpec defines the desired state of CronJob" instead of the large block the script expects.

So, the script seems to be looking for text that doesn't exist in the source it's processing. How should I handle this discrepancy to get make generate-docs to pass for my PR?

camilamacedo86:
Hi @chethanm99,

Could you please ensure that your PR is rebased with the latest master branch?

After rebasing, please run:

```
make generate
```

The testdata check ensures that all sample projects and docs are properly generated. This validation runs for all PRs. So if the issue persists after rebasing, it's likely caused by changes in this specific PR.

Thanks a lot! I hope this helps you move forward. 🙌

PR #4725, opened 2025-03-31
🌱 fix (ci): fix lint issue: shadow: declaration of pluginCfg shadows declaration

camilamacedo86:
Fix plugin config test by resetting struct to avoid value reuse

Avoids test contamination caused by shared PluginConfig variable by explicitly resetting it before decoding in each test case. Also resolves shadowing linter errors reported by golangci-lint.


camilamacedo86:
Moving forward to unblock other prs

PR #4727, opened 2025-04-01
🌱 Copy the new ginkgolinter configs to all .golangci's

nunnatsa:
add the new ginkgolinter configuration, to all the .golangci.yml files in the project.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,6 +18,7 @@ issues:
 
 linters-settings:
   ginkgolinter:
+    forbid-focus-container: true
     forbid-spec-pollution: true
   govet:
     enable-all: true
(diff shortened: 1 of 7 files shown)


nunnatsa:
the `forbid-spec-pollution: true` configuration, introduced too many findings. Closing.

We can still set the `forbid-focus-container: true` configuration

@camilamacedo86 - WDYT?

kersten:
@nunnatsa I think we need to update all templates first. Is that something you'd be interested in helping with? I’d really appreciate the support :)

nunnatsa:
Not sure I'll have the capacity...

kersten:
I will check how complicated that one is, the templates stuff makes it a bit harder, but there should not be many places. I hope 🤞🏻

nunnatsa:
There were too many errors. We can split it to smaller PR, but honestly, I'm not a huge fan of the `forbid-spec-pollution` ginkgolinter rule. I know it's a ginkgo's best practice, and that why it's available in the linter, but it's not always practical.

PR #4733, opened 2025-04-01
🌱 (chore): convert plugin.Version receiver methods to use pointer receiver

kersten:
This PR updates all implementations of the `plugin.Plugin` interface to return `*plugin.Version` instead of `plugin.Version` by value.

### Motivation

- Prevents unnecessary copies of the version struct
- Makes pointer semantics explicit and consistent across plugins and mocks

### Changes

- Updated all `.Version()` implementations to return a pointer
- Updated `plugin.Version` method receivers (e.g., `Validate`, `Compare`, `String`) to pointer receivers
- Adjusted affected test mocks and helpers

### DIFF
--- a/pkg/cli/suite_test.go
+++ b/pkg/cli/suite_test.go
@@ -52,7 +52,7 @@ func newMockPlugin(name, version string, projVers ...config.Version) plugin.Plug
 }
 
 func (p mockPlugin) Name() string                               { return p.name }
-func (p mockPlugin) Version() plugin.Version                    { return p.version }
+func (p mockPlugin) Version() *plugin.Version                   { return &p.version }
 func (p mockPlugin) SupportedProjectVersions() []config.Version { return p.projectVersions }
 
 type mockDeprecatedPlugin struct { //nolint:maligned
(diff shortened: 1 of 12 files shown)


camilamacedo86:
The issue with this one is that it would break the API.

Also, deprecating it isn't ideal either, because we'd end up with a strange name that we’d have to support indefinitely.

So I’ll need some time to properly review and evaluate this case before moving forward.

camilamacedo86:
Hi @kersten 

Thank you for this contribution! 🎉 

After analyse this one. Unfortunately, we cannot proceed with this change at the moment. It introduces a breaking change to a public API (plugin.Version) that is widely used in projects like Operator SDK as others which might consume Kubebuidler as a lib. Example: https://github.com/search?q=repo%3Aoperator-framework%2Foperator-sdk%20plugin.Version&type=code

While switching to pointer receivers may have internal benefits, the impact on existing consumers outweighs the value this change currently brings. Accepting it now would break compatibility for many users relying on value-based usage.

The best path forward would be to open an issue to track this enhancement. We can link this PR there and plan to revisit and include it as part of the work for a future major release — likely kubebuilder v5.x, where breaking changes are expected and manageable. I did this one: https://github.com/kubernetes-sigs/kubebuilder/issues/4761 so I hope that you do not mind but I am closing this one as deferred. 

Thanks again for the effort and your understanding!

Best regards,

PR #4736, opened 2025-04-02
🌱 (chore): replace empty slice literals with idiomatic var decls

kersten:
replace empty slice literals with idiomatic var decls for []metav1.Condition, []templates.CustomMetricItem and []external.Flag


camilamacedo86 (approved):
/ok-to-test
/lgtm
/approve

PR #4757, opened 2025-04-04
✨ Removes mode go install after v2 update

Kuzuri247:
Fixes:#4677

PR #4758, opened 2025-04-04
🌱 added backup files in gitignore 

Kuzuri247:
Added golangci backup yml while migration

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ docs/book/src/docs
 /testdata/**/go.sum
 /docs/book/src/simple-external-plugin-tutorial/testdata/sampleexternalplugin/v1/bin
 /testdata/**legacy**
+
+# skip backup files while migrating 
+.golangci.bck.yml


Kuzuri247:
@camilamacedo86

.gitignore
```diff
@@ -28,3 +28,6 @@ docs/book/src/docs
 /testdata/**/go.sum
 /docs/book/src/simple-external-plugin-tutorial/testdata/sampleexternalplugin/v1/bin
 /testdata/**legacy**
+
+# skip backup files while migrating 
+.golangci.bck.yml
```
camilamacedo86, line 33: Why should we add it?
Is not this file added just once?
When was the migration made?

Kuzuri247: I manually deleted the backup files in #4757

camilamacedo86: That's what I imagined as well, but do we actually expect this file to be present in the future, or was it just needed for a one-time operation?

Kuzuri247: The backup file is just the previous golangci.yml file which is upgraded and then golangci makes it as a backup

camilamacedo86: So, I think we can agree that there's no need to add anything here, right?
Since this is a one-time operation, and after the migration we shouldn't see this file anymore.
Would you mind closing this one as deferred?

Kuzuri247: Yes

Kuzuri247:
One time change

PR #4759, opened 2025-04-04
📖 Update versions_compatibility_supportability.md to fix link in the warning note

possani:
Fix link pointing to the wrong page

### DIFF
--- a/docs/book/src/versions_compatibility_supportability.md
+++ b/docs/book/src/versions_compatibility_supportability.md
@@ -46,4 +46,4 @@ For further information on the project layout, see the doc [What's in a basic pr
 
 </aside>
 
-[basic-project-doc]: ./reference/project-config.md
+[basic-project-doc]: ./cronjob-tutorial/basic-project.md


camilamacedo86 (changes requested):

But, the link here shows correct, we have not the path used.

https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/reference/project-config.md


What is the issue that you faced?

possani:
> Thank you for your contribution
> 
> But, the link here shows correct, we have not the path used.
> 
> https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/reference/project-config.md
> 
> What is the issue that you faced?

Hi Camila, thanks for looking into it.

If you read the paragraph where this link is mentioned, it's talking about the proposed layout created with scaffold.
You can see that even the text in the link it reads `basic-project-doc`. You can also compare with other references within the book. They all point to the same place.

camilamacedo86 (approved):
Thank you for your contirbution 🥇 

I checked this one all fine. Good catcher 🎉 

/lgtm
/approve
/ok-to-test

PR #4763, opened 2025-04-07
🌱 (chore): use pointer receiver for Validate and rename shadowed var

kersten:
use pointer receiver for Validate and rename shadowed var


camilamacedo86 (approved):
/lgtm
/approve
/ok-to-test

PR #4764, opened 2025-04-07
🌱 use `env:` to set variable. Otherwise quoting issues.

guettli:
If the title contains quotes, the current GitHub Action will fail to run.


camilamacedo86 (approved):
/approve
/lgtm
/ok-to-test

PR #4771, opened 2025-04-08
🌱 fix staticcheck lint issues - ST1008 and ST1008

camilamacedo86:


dongjiang1989 (approved):
LGTM

pkg/machinery/scaffold_test.go
mateusoliveira43, line 1: also need to remove exclusion from linter config file, right?

PR #4775, opened 2025-04-10
🌱 (fix/enhance): correct GNU sed detection method for remove-spaces makefile target

kersten:
This change updates the trailing spaces removal in the Makefile. Previously, the script used the `uname` check to determine whether to use GNU or BSD sed syntax, which did not work well on macOS when GNU sed is installed.

Now the script uses a check on `sed --version` for "GNU" to decide which sed syntax to apply. This ensures that, regardless of the OS, if GNU sed is available in the PATH, the GNU sed branch is used.


camilamacedo86 (approved):
I am OK with the change.
lets move forward 

/lgtm
/ok-to-test

PR #4776, opened 2025-04-10
🌱 (chore): unify init and edit subcommands into a single command struct

kersten:
- Replaces separate init/edit subcommands with a shared `subcommand` type
- Uses dynamic scaffolder injection for each command
- Adds suite and plugin tests for Grafana plugin


pkg/plugins/optional/grafana/v1alpha/plugin_test.go
```diff
@@ -0,0 +1,116 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/spf13/afero"
+	v3 "sigs.k8s.io/kubebuilder/v4/pkg/config/v3"
+	"sigs.k8s.io/kubebuilder/v4/pkg/machinery"
+	plugin2 "sigs.k8s.io/kubebuilder/v4/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/grafana/v1alpha"
+)
+
+var _ = Describe("Grafana Plugin", func() {
```
camilamacedo86, line 29: Do we need this change to achieve the intended goal?

If not, what do you think about starting with a PR that increases the Grafana plugin test coverage first? That way, we can define a clearer scope and build on top of that with more confidence.

Let me know your thoughts! 😊

camilamacedo86:
Hi @kersten 

Really thank so much for the effort on this one, but I think we can close this one based on: https://github.com/kubernetes-sigs/kubebuilder/issues/4902

Came out that the init implementation for helm seems not really make too much since 

I am closing it, but feel free to reopen or ping out if you disagree.

PR #4777, opened 2025-04-10
🐛 (helm/alpha-v1): Add control-plane labels to metrics service and ServiceMonitor templates

sarthaksarthak9:
The Helm plugin scaffolding was missing the control-plane: controller-manager label on the metrics service, which prevented the ServiceMonitor from finding the service and collecting metrics when deployed via Helm chart.

fixes: #4591

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/metrics/metrics-service.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/metrics/metrics-service.yaml
@@ -6,6 +6,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "chart.labels" . | nindent 4 }}
+    control-plane: controller-manager
 spec:
   ports:
     - port: 8443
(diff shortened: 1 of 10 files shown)


docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/metrics/metrics-service.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "chart.labels" . | nindent 4 }}
+    control-plane: controller-manager
```
camilamacedo86, line 9: I think we need to add to the ServiceMonitor as well, see:

https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-with-plugins/config/prometheus/monitor.yaml#L3-L6C2

https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-with-plugins/dist/chart/templates/prometheus/monitor.yaml

Could you please ensure that we have the label in the same places as we have for kustomize?

camilamacedo86:
Hi @Jaybee18

could you help us by checking that these changes sort out the issue raised by you?

camilamacedo86 (approved):

camilamacedo86 (approved):
/lgtm

It seems good to fly 🚀 

Thank you so much for your contribution 🥇

Issue #4778, opened 2025-04-10
Bug: Incorrect Webhook Import Aliasing in cmd/main.go Scaffold


camilamacedo86:
### What broke? What's expected?

When generating or updating `cmd/main.go`, the webhook import aliasing is broken. It results in duplicated imports. Example:

https://github.com/kubernetes-sigs/kubebuilder/blob/40220077697563f82e4ff51003c17dabe0e588e6/testdata/project-v4/cmd/main.go#L45-L48

Expected:
- Single group: `webhook<version>`
- Multi-group: `webhook<group><version>`

Current logic fails to apply aliasing correctly.

See logic in:
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/cmd/main.go#L160-L167
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/cmd/main.go#L196-L205

Legacy logic should remain unchanged.

### Reproducing this issue

Just check the samples under testdata.
After doing the change/fix, we should run `make generate` and all should work well as expected.
We also should be able to remove from golanci the ignore: https://github.com/kubernetes-sigs/kubebuilder/blob/40220077697563f82e4ff51003c17dabe0e588e6/testdata/project-v4/.golangci.yml#L39-L41 as part of this fix.

### KubeBuilder (CLI) Version

v4.5.2

### PROJECT version

3

### Plugin versions

```yaml
go/v4
```

dongjiang1989:
Hi @camilamacedo86 
I high priority to fix it

PR #4781, opened 2025-04-11
🐛Fix: Incorrect Webhook Import Aliasing in cmd/main.go Scaffold 

dongjiang1989:
Close https://github.com/kubernetes-sigs/kubebuilder/issues/4778

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/cmd/main.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/cmd/main.go
@@ -158,10 +158,11 @@ func (f *MainUpdater) GetCodeFragments() machinery.CodeFragmentsMap {
 		imports = append(imports, fmt.Sprintf(apiImportCodeFragment, f.Resource.ImportAlias(), f.Resource.Path))
 	}
 	if f.WireWebhook && !f.IsLegacyPath {
-		importPath := fmt.Sprintf("webhook%s", f.Resource.ImportAlias())
 		if !f.MultiGroup || f.Resource.Group == "" {
+			importPath := fmt.Sprintf("webhook%s", f.Resource.Version)
 			imports = append(imports, fmt.Sprintf(webhookImportCodeFragment, importPath, f.Repo, f.Resource.Version))
 		} else {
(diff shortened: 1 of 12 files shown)


dongjiang1989:
cc @camilamacedo86 PTAL

pkg/plugins/golang/v4/scaffolds/internal/templates/cmd/main.go
```diff
@@ -195,11 +196,21 @@ func (f *MainUpdater) GetCodeFragments() machinery.CodeFragmentsMap {
 	}
 	if f.WireWebhook {
 		if f.IsLegacyPath {
```
camilamacedo86, line 198: If is Legacy, we should not change and keep it as it was before.

pkg/plugins/golang/v4/scaffolds/internal/templates/cmd/main.go
```diff
@@ -195,11 +196,21 @@ func (f *MainUpdater) GetCodeFragments() machinery.CodeFragmentsMap {
...
-			setup = append(setup, fmt.Sprintf(webhookSetupCodeFragmentLegacy,
-				f.Resource.ImportAlias(), f.Resource.Kind, f.Resource.Kind))
+			if !f.MultiGroup || f.Resource.Group == "" {
+				setup = append(setup, fmt.Sprintf(webhookSetupCodeFragmentLegacy,
+					f.Resource.Version, f.Resource.Kind, f.Resource.Kind))
+			} else {
+				setup = append(setup, fmt.Sprintf(webhookSetupCodeFragmentLegacy,
+					f.Resource.ImportAlias(), f.Resource.Kind, f.Resource.Kind))
+			}
```
camilamacedo86, line 205: We should only change the code when it is not legacy. Is there any reason why we do it?

dongjiang1989: Thanks @camilamacedo86 
I got it
Revert it done

camilamacedo86 (approved):
Hi @dongjiang1989 

It is amazing 🎉 
/lgtm
/approve

Issue #4625, opened 2025-03-20
(helm/v1-alpha) - force command does not update values.yaml when manager spec has changed


abhijith-darshan:
### What broke? What's expected?

When scaffolding the project with `--plugins=helm.kubebuilder.io/v1-alpha` any changes made to manager spec is not copied over to values.yaml when executing `kubebuilder edit --plugins=helm/v1-alpha --force`

### Reproducing this issue

create a new project 

```shell
kubebuilder init --domain my.domain --repo my.domain/guestbook --plugins=go/v4,helm.kubebuilder.io/v1-alpha
```

The scaffold generates a `dist/chart`

Next edit the spec of `manager.yaml` in `config/manager/manager.yaml`

for example increase the `replica` count to 2

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: controller-manager
  namespace: system
  labels:
    control-plane: controller-manager
    app.kubernetes.io/managed-by: kustomize
spec:
  selector:
    matchLabels:
      control-plane: controller-manager

  replicas: 2
....
```

run `kubebuilder edit --plugins=helm/v1-alpha --force` as recommended in the documentation [update helm chart wit latest changes](https://book.kubebuilder.io/plugins/available/helm-v1-alpha#use-the-edit-command-to-update-the-helm-chart-with-the-latest-changes)

```shell
$kubebuilder edit --plugins=helm/v1-alpha --force
INFO Generating Helm Chart to distribute project  
INFO webhook manifests were not found at config/webhook/manifests.yaml 
INFO Successfully copied config/rbac/leader_election_role.yaml to dist/chart/templates/rbac/leader_election_role.yaml 
INFO Successfully copied config/rbac/leader_election_role_binding.yaml to dist/chart/templates/rbac/leader_election_role_binding.yaml 
INFO Successfully copied config/rbac/metrics_auth_role.yaml to dist/chart/templates/rbac/metrics_auth_role.yaml 
INFO Successfully copied config/rbac/metrics_auth_role_binding.yaml to dist/chart/templates/rbac/metrics_auth_role_binding.yaml 
INFO Successfully copied config/rbac/metrics_reader_role.yaml to dist/chart/templates/rbac/metrics_reader_role.yaml 
INFO Successfully copied config/rbac/role.yaml to dist/chart/templates/rbac/role.yaml 
INFO Successfully copied config/rbac/role_binding.yaml to dist/chart/templates/rbac/role_binding.yaml 
[...]

PR #4782, opened 2025-04-15
✨ (helm/v1-alpha): Update Helm values.yaml with Kustomize manager spec changes when using --force flag

sarthaksarthak9:
fixes: #4625

- Improves YAML number type handling for replicas, terminationGracePeriodSeconds, and other numeric values
- Ensures proper extraction of resource limits, requests, container args, and termination grace period from manager.yaml
- Correctly populates values.yaml with the latest configuration from manager.yaml when using the --force flag

Screenshot from 2025-04-14 22-38-12<br>
Screenshot from 2025-04-14 22-39-01

### DIFF
--- a/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
+++ b/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
@@ -19,6 +19,7 @@ package scaffolds
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -52,6 +53,13 @@ type initScaffolder struct {
 	force bool
 }
 
+// Define constants for repeated strings
+const (
+	deploymentKind        = "Deployment"
+	managerContainerName  = "manager"
+	controllerManagerName = "controller-manager"
+)
+
 // NewInitHelmScaffolder returns a new Scaffolder for HelmPlugin
 func NewInitHelmScaffolder(cfg config.Config, force bool) plugins.Scaffolder {
 	return &initScaffolder{
@@ -71,6 +79,29 @@ func (s *initScaffolder) Scaffold() error {
 
 	imagesEnvVars := s.getDeployImagesEnvVars()
 
+	// Extract manager values when --force flag is used
+	var managerValues map[string]interface{}
+	var extractErr error
+	if s.force {
+		// First try to get values directly from manager.yaml
+		managerValues, extractErr = s.extractManagerValues()
+		if extractErr != nil {
+			log.Warnf("Failed to extract manager values from manager.yaml: %v", extractErr)
+
+			// If that fails, try to get values from kustomization patches
+			managerValues, extractErr = s.extractManagerValuesFromKustomization()
+			if extractErr != nil {
+				log.Warnf("Failed to extract manager values from kustomization: %v", extractErr)
+
+				// As a last resort, try to build the manifests using kustomize
+				managerValues, extractErr = s.extractManagerValuesUsingKustomize()
+				if extractErr != nil {
+					log.Warnf("Failed to extract manager values using kustomize: %v", extractErr)
+				}
+			}
+		}
+	}
+
 	scaffold := machinery.NewScaffold(s.fs,
 		machinery.WithConfig(s.config),
 	)
@@ -86,9 +117,10 @@ func (s *initScaffolder) Scaffold() error {
 		&github.HelmChartCI{},
 		&templates.HelmChart{},
 		&templates.HelmValues{
-			HasWebhooks:  hasWebhooks,
-			DeployImages: imagesEnvVars,
-			Force:        s.force,
+			HasWebhooks:   hasWebhooks,
+			DeployImages:  imagesEnvVars,
+			Force:         s.force,
+			ManagerValues: managerValues,
 		},
 		&templates.HelmIgnore{},
 		&charttemplates.HelmHelpers{},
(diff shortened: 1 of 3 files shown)


docs/book/src/cronjob-tutorial/testdata/project/dist/chart/values.yaml
```diff
@@ -7,7 +7,6 @@ controllerManager:
       tag: latest
     args:
       - "--leader-elect"
-      - "--metrics-bind-address=:8443"
```
camilamacedo86, line 10: Why this flag end up removed?

pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/values.go
```diff
@@ -54,22 +56,40 @@ func (f *HelmValues) SetTemplateDefaults() error {
...
+{{- end }}
       requests:
+{{- if and .ManagerSpecs (index .ManagerSpecs "resources.requests") }}
+{{- range $key, $value := index .ManagerSpecs "resources.requests" }}
+        {{ $key }}: {{ $value }}
+{{- end }}
+{{- else }}
         cpu: 10m
         memory: 64Mi
+{{- end }}
```
camilamacedo86, line 92: Same here

pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/values.go
```diff
@@ -97,7 +117,7 @@ controllerManager:
     runAsNonRoot: true
     seccompProfile:
       type: RuntimeDefault
-  terminationGracePeriodSeconds: 10
+  terminationGracePeriodSeconds: {{- if and .ManagerSpecs (index .ManagerSpecs "terminationGracePeriodSeconds") }} {{ index .ManagerSpecs "terminationGracePeriodSeconds" }}{{- else }} 10{{- end }}
```
camilamacedo86, line 120: sane gere as well

pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
```diff
@@ -560,3 +571,103 @@ func hasWebhooksWith(c config.Config) bool {
 
 	return false
 }
+
+// readManagerSpecs reads the manager.yaml file from config/manager/manager.yaml
+// and extracts relevant specs to be used in values.yaml generation
+func (s *initScaffolder) readManagerSpecs() error {
```
camilamacedo86, line 577: Would it be feasible to generate the manager from configuration files, instead of maintaining boilerplate code for both the manager and the Helm plugin? Could this approach work effectively, or are there potential pitfalls we should be aware of?

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/values.yaml
```diff
@@ -7,7 +7,6 @@ controllerManager:
       tag: latest
```
camilamacedo86, line 7: In this case, it’s not a bug fix, but rather a Request for Enhancement (RFE) ✨ — the goal is to support changes and customizations made to the manager so they’re automatically reflected in the Helm plugin, which currently doesn’t happen.

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/values.yaml
sarthaksarthak9: When we run `kubebuilder edit --force`, it's expected that the --metrics-bind-address=:8443 flag gets removed from values.yaml. Why? Because kubebuilder grabs the actual arguments from the `manager.yaml file` where only `--leader-elect` and `--health-probe-bind-address=:8081` are set and uses those to overwrite the defaults in values.yaml.

Since `--metrics-bind-address` isn’t in `manager.yaml`, it doesn’t make it into the updated `values.yaml` either. That’s actually by design. The `--force flag` is all about keeping the Helm chart in sync with the actual Kubernetes configuration.

But no worries the metrics endpoint will still work as long as `metrics.enable: true` is set in values.yaml.

pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/values.go
```diff
@@ -54,22 +56,40 @@ func (f *HelmValues) SetTemplateDefaults() error {
...
     resources:
       limits:
+{{- if and .ManagerSpecs (index .ManagerSpecs "resources.limits") }}
+{{- range $key, $value := index .ManagerSpecs "resources.limits" }}
+        {{ $key }}: {{ $value }}
+{{- end }}
+{{- else }}
         cpu: 500m
         memory: 128Mi
+{{- end }}
```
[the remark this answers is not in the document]
sarthaksarthak9: The condition in the `values.yaml` template is there to give users flexibility and keep their custom settings intact. When we use the `--force flag`, Kubebuilder pulls the resource limits from `manager.yaml` and stores them in something called `ManagerSpecs`.

The conditional logic in the template —
{{- if and .ManagerSpecs (index .ManagerSpecs "resources.limits") }} —
basically checks: “Hey, do we have any custom resource limits from manager.yaml? If yes, let’s use those.”

This check is important because it allows the chart to handle two situations:

1. If we’ve defined resource limits in `manager.yaml`, they’ll be carried over into `values.yaml`.

2. If we haven’t, it falls back to defaults.

Without that condition, we’d always use the defaults, and any custom resource settings we added in `manager.yaml` would get wiped out the next time we regenerate the chart using `--force`. That would defeat the whole point of using `--force`, which is meant to keep things consistent across the config files while respecting the customizations.

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/values.yaml
camilamacedo86: If we do not set the flag, then the metrics should not work since the metrics server will receive zero by default, and it will be disabled. See:

https://github.com/kubernetes-sigs/kubebuilder/blob/9932c89b631bb03413afe858c736613c66e07882/testdata/project-v4/cmd/main.go#L73-L74

`metrics.enable: true` in the values will not change that. Unless we do a manifest to patch the metrics as we do with customise. However, it would not work because Helm cannot define which manifest is applied first or after.

**That might be one of the reasons why we have a template for the manager instead of try to read the files directly from config as we do for the other case scenarios**

camilamacedo86: PR:  If the e2e tests do not fail here, we have an issue in the e2e tests.
The changes will break the metrics for HelmChart, so e2e tests should fail.

camilamacedo86: Failed as expected: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kubebuilder/4782/pull-kubebuilder-e2e-k8s-1-32-0/1912025696591417344#1:build-log.txt%3A2589

sarthaksarthak9:
@camilamacedo86 could you please help to correct the PR title? I'm not sure why it's failing the test. Thanks!

camilamacedo86:
Hi 

>
@camilamacedo86 could you please help to correct the PR title? I'm not sure why it's failing the test. Thanks!

I will need sometime to think about this one and your proposed solution.

I will fix the title, no worry about it.

pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
```diff
@@ -560,3 +592,400 @@ func hasWebhooksWith(c config.Config) bool {
...
+			continue
+		}
+
+		var docMap map[string]interface{}
+		if err := yaml.Unmarshal([]byte(doc), &docMap); err != nil {
+			continue
+		}
+
+		kind, found := docMap["kind"].(string)
+		if found && kind == deploymentKind {
```
camilamacedo86, line 632: What happens if the person another deployment on the same file?

camilamacedo86 (changes requested):
Thanks so much for tackling this! Really appreciate the effort here 🥇

I just have a concern — this approach, while helpful, might cause issues with user customizations. It tries to patch over a deeper limitation, and I worry it’s adding complexity without fully solving the problem.

The core issue is that we use a fixed template for the manager:
https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/chart-templates/manager

In other parts of the plugin, we read directly from config/ and build from there — which has worked well. If we could extract what we need from the Kustomize files instead of using static templates, I think we’d support more use cases with less risk. The solution might be to skip the template entirely and build the manager from the config content.

Totally open to chat more about it — just wanted to share this thought. Thanks again! 🙌

### ⚠️ Potential Issues

1. **Single Deployment Assumption**  
`findDeploymentInYAML` returns the first `kind: Deployment`.  
> Might extract the wrong one if multiple exist.

2. **Name-Specific Matching**  
Matches only `controller-manager` or `manager`.  
>  Custom names might be skipped or misidentified.

3. **Partial Extraction**  
Only pulls `replicas`, `resources`, `args`, probes.  
>  Missing things like `volumes`, `env`, `affinity`, etc.

4. **Silent Fallbacks**  
Falls back from file → patch → kustomize build.  
>  May generate default/incomplete `values.yaml` silently.

camilamacedo86:
Hi @sarthaksarthak9 

Thank a lot for your contribution, I think we should close this PR in favor of solve it as proposed : https://github.com/kubernetes-sigs/kubebuilder/issues/4833

WDYT? Please, see the concerns, why not move forward in : https://github.com/kubernetes-sigs/kubebuilder/pull/4782#pullrequestreview-2866223510

PR #4783, opened 2025-04-19
📖 Update getting-started.md

idemianyshyn:
The change introduces a needRequeue flag to separate the logic for requeueing reconciliation from error handling. Previously, using err != nil to trigger a requeue was confusing, as it mixed error checking with the need for the process to be re-executed.

### DIFF
--- a/docs/book/src/getting-started.md
+++ b/docs/book/src/getting-started.md
@@ -180,7 +180,7 @@ reconcile App {
   // Check the Database Deployment's replicas size
   // If deployment.replicas size doesn't match cr.size, then update it
   // Then, restart from the beginning of the reconcile. For example, by returning `reconcile.Result{Requeue: true}, nil`.
-  if err != nil {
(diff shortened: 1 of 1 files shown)


camilamacedo86 (changes requested):
Hey! 😊

Thanks so much for the PR — really appreciate the time and effort you put into this!

I took a look, and I think the intention here is thoughtful, but we already requeue when err != nil, so this change doesn’t seem to improve the accuracy or behavior of the code. It also introduces a bit of additional complexity, which might make it harder for others to follow what’s going on — especially with the extra variable that doesn’t feel strictly necessary.

Because of that, I don’t think we’re gaining enough value here to justify the added complexity, and that’s why I don’t believe we should move forward with this one.

If you don’t mind, we’ll go ahead and close this PR for now. Thanks again for your contribution — always appreciate the collaboration! 🙏

Issue #4564, opened 2025-02-14
Cronjob example - add fianlizer


matrus2:
### What do you want to happen?

I wonder why in the cronjob example there is a condition to check if cronjob contains finalizer and if so add it. See:
```
		if !controllerutil.ContainsFinalizer(cronJob, myFinalizerName) {
			controllerutil.AddFinalizer(cronJob, myFinalizerName)
```
https://github.com/kubernetes-sigs/kubebuilder/blob/93a0823eb07d3bc6d1d601471a0fb4de64380abd/docs/book/src/cronjob-tutorial/testdata/finalizer_example.go#L69

However `ContainsFinalizer` seems to be overkill as `AddFinalizer` function do the same check inside and it will not add the finalizer once it is already there:

```
func AddFinalizer(o client.Object, finalizer string) (finalizersUpdated bool) {
	f := o.GetFinalizers()
	for _, e := range f {
		if e == finalizer {
			return false
		}
	}
	o.SetFinalizers(append(f, finalizer))
	return true
}
```

I would like to remove `ContainsFinalizer` as it is not needed and update documentation. Before doing so please comment if my reasoning is fine.

camilamacedo86:
The CronJob examples should follow the same design approach as the Deploy Image plugin. That means they should be scaffolded in a similar way, with modifications to implement CronJob-specific logic instead of managing an image.

For reference, see this example:
🔗 [BusyBox Controller Example](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-with-plugins/internal/controller/busybox_controller.go)

We have issues tracking this effort.

Could you also review the Deploy Image implementation? If you think any improvements should be made there, we should apply them consistently.

Feel free to raise PRs with your suggestions—we can iterate from there! 🚀 If you spot any areas that could be improved, go for it!

Thank you for all your contributions! 🥇

PR #4789, opened 2025-04-25
🐛 (deploy-image/v1-alpha): remove superfluous addFinalizer error handling

matrus2:
fix: #4564 

This pull request refactors the logic for adding finalizers in multiple reconciler implementations to simplify the code and improve readability. The changes replace the use of `controllerutil.ContainsFinalizer` checks with a more concise pattern that directly checks the result of `controllerutil.AddFinalizer`.


testdata/project-v4-multigroup/internal/controller/example.com/busybox_controller.go
```diff
@@ -121,14 +121,8 @@ func (r *BusyboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-	if !controllerutil.ContainsFinalizer(busybox, busyboxFinalizer) {
-		log.Info("Adding Finalizer for Busybox")
-		if ok := controllerutil.AddFinalizer(busybox, busyboxFinalizer); !ok {
-			err = fmt.Errorf("finalizer for Busybox was not added")
-			log.Error(err, "Failed to add finalizer for Busybox")
-			return ctrl.Result{}, err
-		}
-
+	if added := controllerutil.AddFinalizer(busybox, busyboxFinalizer); added {
+		log.Info("Finalizer added for Busybox")
```
camilamacedo86, line 125: Thank you very much for the suggestion and for taking the time to improve the code! 🙏  

The reason for keeping the original implementation as it is, is because:

**Separation of concerns:**  
We first check if the finalizer is already present using `ContainsFinalizer`, and only attempt to add it if necessary. This makes the logic clear and avoids unnecessary operations.

**Efficient error handling:**  
By explicitly checking first, we can log and manage errors more precisely at each step, which makes debugging and monitoring easier.

**Avoid unnecessary API server writes:**  
Updating the resource only when a change is needed helps reduce API server load, avoids unnecessary reconciliation retries, and keeps the controller efficient.

Even though `AddFinalizer` internally checks, doing the explicit check first improves clarity, and bring the above benefits which seems more aligned.

For these reasons, I think it might be better to keep the original implementation as it is.

Thanks again for your contribution and for sparking this valuable discussion! 🌟

camilamacedo86: /hold

Would you agree to close this one?

matrus2: Thank you for the quick and thorough review
I agree that using ContainsFinalizer and AddFinalizer together improves readability. However, I still believe the original code block is misleading and can be improved. It appears to contain dead code within the conditional block:

```
		if ok := controllerutil.AddFinalizer({{ lower .Resource.Kind }}, {{ lower .Resource.Kind }}Finalizer); !ok {
			err = fmt.Errorf("finalizer for {{ .Resource.Kind }} was not added")
			log.Error(err, "Failed to add finalizer for {{ .Resource.Kind }}")
			return ctrl.Result{}, err
		}
```
Since the presence of the finalizer is checked before this block (by ContainsFinalizer), the condition where AddFinalizer returns false (and thus the error logging) will never be reached. Examining the implementation of AddFinalizer confirms this, as it only appends the finalizer and returns true if it wasn't already present.

My proposal aims for a more precise approach:

```
	if !controllerutil.ContainsFinalizer(memcached, memcachedFinalizer) {
		log.Info("Adding Finalizer for Memcached")
		controllerutil.AddFinalizer(memcached, memcachedFinalizer)
		if err = r.Update(ctx, memcached); err != nil {
			log.Error(err, "Failed to update custom resource to add finalizer")
			return ctrl.Result{}, err
		}
	}
```
This revised code explicitly checks for the finalizer's absence, logs the intention to add it, performs the addition, and then handles potential errors during the update operation.

matrus2: @camilamacedo86 Would you agree with this approach?

camilamacedo86: Hi @matrus2 

Thank you for looking into that.
Your observation seems to be accurate, and your proposal seems to solve it well. 👍 

Could you please update this PR with the proposal?
Also, please ensure that you squash the commits for we have a good git history.

Thank you a lot 🥇

camilamacedo86: Hi @matrus2 

Please, ping me when you think that be good to fly
Thank you

matrus2: @camilamacedo86 Changes committed, we are good. 

I see some test failed not related to changes incorporated here, needs some investigation.

camilamacedo86 (approved):
Hi @matrus2,

Thank you so much, and sorry for the delay — I didn’t see the notifications for this one earlier.

It looks good to me! 👍

/lgtm

Issue #4807, opened 2025-05-19
(helm/v1-alpha) Incorrect service name for conversion webhook


fhirscher:
### What broke? What's expected?

If you use the Helm plugin and define a conversion webhook, a non-existent service for the conversion webhook is defined in the CRD of the Helm chart.

### Reproducing this issue

Create a project with the helm/v1-alpha plugin and a multi-version API with a convention webhook.

The convention webhook in the CRD in the Helm chart looks like this:
```
...
spec:
  {{- if .Values.webhook.enable }}
  conversion:
    strategy: Webhook
    webhook:
      clientConfig:
        service:
          namespace: {{ .Release.Namespace }}
          name: webhook-service
          path: /convert
      conversionReviewVersions:
      - v1
  {{- end }}
...
```

The service name should also contain the project name:
```
...
spec:
  {{- if .Values.webhook.enable }}
  conversion:
    strategy: Webhook
    webhook:
      clientConfig:
        service:
          namespace: {{ .Release.Namespace }}
          name: project-v4-with-plugins-webhook-service
          path: /convert
      conversionReviewVersions:
      - v1
  {{- end }}
...
```

### KubeBuilder (CLI) Version

v4.5.2

### PROJECT version

3

### Plugin versions

```yaml
go.kubebuilder.io/v4
helm.kubebuilder.io/v1-alpha
```

PR #4808, opened 2025-05-19
🐛 (helm/v1-alpha): Append project name to webhook service name

fhirscher:
fixes: #4807 

The Helm plugin scaffolding did not adjust the service name for conversion webhooks in the CRDs. This prevented the conversion webhook from being reached.

### DIFF
--- a/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
+++ b/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
@@ -424,6 +424,9 @@ func copyFileWithHelmLogic(srcFile, destFile, subDir, projectName string, hasCon
   labels:
     {{- include "chart.labels" . | nindent 4 }}`, 1)
 
+	// Append project name to webhook service name
+	contentStr = strings.ReplaceAll(contentStr, "name: webhook-service", "name: "+projectName+"-webhook-service")
+
 	var wrappedContent string
 	if isMetricRBACFile(subDir, srcFile) {
 		wrappedContent = fmt.Sprintf(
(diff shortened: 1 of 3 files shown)


pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
```diff
@@ -424,6 +424,9 @@ func copyFileWithHelmLogic(srcFile, destFile, subDir, projectName string, hasCon
   labels:
     {{- include "chart.labels" . | nindent 4 }}`, 1)
 
+	// Append project name to webhook service name
+	contentStr = strings.Replace(contentStr, "name: webhook-service", "name: "+projectName+"-webhook-service", 1)
```
camilamacedo86, line 428: Hi @fhirscher 

Thank you a lot for your contribution. 

See that we are passing : 

https://github.com/kubernetes-sigs/kubebuilder/blob/bc78974892f151a014747e6429c43062e0a4a827/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go#L209

So, here we are updating the CRD with, it seems right 👍 

However, the usage of `1` will result in update just the first one found, so I think need to be `-1` instead . WDYT?

camilamacedo86:
Hi @fhirscher 

Please, my comment above and that for which we accept your contribution, you need to sign: https://github.com/kubernetes-sigs/kubebuilder/pull/4808#issuecomment-2889798232

pkg/plugins/optional/helm/v1alpha/scaffolds/init.go
fhirscher: Hi @camilamacedo86 

That's right. I can't find any unwanted effects when we use `replaceAll()`.

PR #4585, opened 2025-02-28
✨ (CLI, go/v4): Upgrade Golang version to v1.24.0

dongjiang1989:
Upgrade go version 1.23 to 1.24

PR #4818, opened 2025-05-24
🐛 (go/v4) fix: revert accidental addition of explicit docker.io registry in Dockerfile

camilamacedo86:
✅ Why?

Docker already defaults to Docker Hub (docker.io) when no registry is specified.
So both forms are functionally the same, but:

---

2. Keeps the scaffold clean and idiomatic
- Most open source projects use the short form (golang:1.24).
- It's easier to read and better for beginners.
- This follows standard Go and Docker community practices.
- Adding docker.io just adds noise to the file.

---

3. Avoids lock-in or assumptions
- Hardcoding docker.io can break things in environments with custom or mirrored registries.
- It may confuse users who want to use private registries or CI configs.
- By writing golang:1.24, we let the Docker config decide where to pull from — which is more flexible and future-proof.

---

TL;DR
golang:1.24 already pulls from Docker Hub.
Adding docker.io is not needed — and removing it keeps the Dockerfile clean, standard, and portable.
It was added by mistake in the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/4585

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/.devcontainer/devcontainer.json
+++ b/docs/book/src/cronjob-tutorial/testdata/project/.devcontainer/devcontainer.json
@@ -1,6 +1,6 @@
 {
   "name": "Kubebuilder DevContainer",
-  "image": "docker.io/golang:1.24",
+  "image": "golang:1.24",
   "features": {
     "ghcr.io/devcontainers/features/docker-in-docker:2": {},
     "ghcr.io/devcontainers/features/git:1": {}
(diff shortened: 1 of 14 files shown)


docs/book/src/cronjob-tutorial/testdata/project/.devcontainer/devcontainer.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Kubebuilder DevContainer",
-  "image": "docker.io/golang:1.24",
+  "image": "golang:1.24",
```
camilamacedo86, line 3: Hi @dongjiang1989

I am pinging you to make you aware.
It was my mistake since I did not review your PR.

PR #4821, opened 2025-05-24
🐛 refactor: inline getKubebuilderVersion into cli.WithCliVersion setup

camilamacedo86:
Avoid exporting an unnecessary helper.
This change keeps the API surface minimal and avoids exposing internal details unless truly needed.
Note: the function was newly introduced and hasn't been included in any release yet, so we can safely change its visibility without breaking anything.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -55,7 +55,7 @@ func Run() {
 	c, err := cli.New(
 		cli.WithCommandName("kubebuilder"),
 		cli.WithVersion(versionString()),
-		cli.WithCliVersion(GetKubebuilderVersion()),
+		cli.WithCliVersion(getKubebuilderVersion()),
 		cli.WithPlugins(
 			golangv4.Plugin{},
 			gov4Bundle,
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -63,8 +63,8 @@ func versionString() string {
 	})
 }
 
-// GetKubebuilderVersion returns only the CLI version string
-func GetKubebuilderVersion() string {
+// getKubebuilderVersion returns only the CLI version string
+func getKubebuilderVersion() string {
 	if kubeBuilderVersion == unknown {
 		if info, ok := debug.ReadBuildInfo(); ok && info.Main.Version != "" {
 			kubeBuilderVersion = info.Main.Version


camilamacedo86:
> [APIDiff / Verify API differences (pull_request)](https://github.com/kubernetes-sigs/kubebuilder/actions/runs/15231654155/job/42840159135?pr=4821)

Hey @vitorfloriano 👋

I’m going ahead and making this method internal since we haven’t released it yet — so no impact for users.

Just a quick heads-up: in general, once we export something, we commit to supporting it unless we introduce a major version (like Kubebuilder CLI 5.0.0 or 6.0.0). We always try to avoid breaking changes to keep things smooth for our users, and major releases usually take 2–3 years to happen.

So this is just a small cleanup to keep our public API minimal.
Let me know if you have any concerns — and thanks! 😊

vitorfloriano:
Hi @camilamacedo86 no concerns at all! I just have a question about how the API is currently used by others. Is it mainly for the plugin ecosystem, or do other Kubebuilder components also expose parts of the API? How do users benefit from these?

I’m asking because, as a newcomer, it’s a bit hard for me to understand what’s intended to be public and what isn’t. I’ve noticed some projects use the `internal/` package pattern to clearly separate public and private APIs, which helps avoid confusion.

Would love to hear your thoughts on this.

camilamacedo86:
Hi! 😊

>  I just have a question about how the API is currently used by others. Is it mainly for the plugin ecosystem, or do other Kubebuilder components also expose parts of the API? How do users benefit from these?

Kubebuilder’s APIs are mainly used by the plugin ecosystem, but they’re not exclusive to it. Any tool can build on top — for example, Operator SDK (https://github.com/operator-framework/operator-sdk) uses Kubebuilder under the hood and adds support for Helm, Ansible, and OLM integration. This reuse lets tools avoid reimplementing core features.

See: https://book.kubebuilder.io/plugins/extending

> I’m asking because, as a newcomer, it’s a bit hard for me to understand what’s intended to be public and what isn’t. I’ve noticed some projects use the internal/ package pattern to clearly separate public and private APIs, which helps avoid confusion.

Yes — anything not meant for external use is placed in `internal/` or left unexported (lowercase). This is the standard way to indicate intended usage and avoid confusion as protected the Public API. As you see for example, our templates are under internal pakages: https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/golang/v4/scaffolds/internal/templates

Let me know if you want to dive deeper we can speak via Slack

PR #4823, opened 2025-05-24
🌱 [external-plugin] fix(ci): use Make targets for CLI install and expose plugin failures

camilamacedo86:
…


camilamacedo86:
I am moving forward here just to unblock the CI. However, we must review this sample plugin. A issue was created for we track the need: https://github.com/kubernetes-sigs/kubebuilder/issues/4824

PR #4827, opened 2025-05-25
:sparkles: Marker support for Rust file extension .rs

mabulgu:
I have a new Operator SDK plugin for Rust, and I use kubebuilder markers for scaffolding. This PR adds the `.rs` extension support to the marker.

Here is the plugin repo: https://github.com/SystemCraftsman/rust-operator-plugins

I would like to suggest this to be included in the Operator SDK org later on (gonna do that via the issue that I will open there).

Also I see the message `When adding additional file extensions, update also the NewMarkerFor documentation and error`, but no idea where I can update these. If anyone can show me the way, I can also update them to include Rust file ext.


camilamacedo86 (changes requested):
Hi @mabulgu 👋

**It is VERY cool that you are creating an external plugin!!! 🎉**  
Thank you so much for building in the ecosystem and even presenting it at KCD — that's amazing!

We might even start tracking all external plugins and promoting them somewhere in the Kubebuilder docs. **I’ll create a task for that — because your work deserves to be seen! 💡**

---

### 🧭 About the PR to Add `.rs` Marker Support

However, in this case, **Kubebuilder itself does not scaffold or use `.rs` files directly**.

So for now, we **won’t add `.rs` to the core marker support**. This helps us keep the core focused only on what it uses internally.

But **don’t worry!** You can absolutely **use Kubebuilder as a library** and create your **own custom markers** to support `.rs`, or any other extension you want 🚀

---

### 📚 We should definitely create a doc for this:
> *“How to create my own markers for my own extensions not supported in Kubebuilder itself”*

Here’s an example to help you (and others!) get started right away:

---

## ✅ Example: External Plugin With Custom `.rs` Marker Support

> Using `sigs.k8s.io/kubebuilder/v4/pkg/machinery`

```go
// plugin/marker.go
package plugin

import (
	"fmt"
	"path/filepath"
	"strings"

	"sigs.k8s.io/kubebuilder/v4/pkg/machinery"
)

const myPrefix = "+myplugin:scaffold:"

var rustCommentsByExt = map[string]string{
	".rs": "//",
}

type Marker struct {
	machinery.Marker
}

func NewRustMarker(path string, value string) Marker {
	ext := filepath.Ext(path)
	if comment, ok := rustCommentsByExt[ext]; ok {
		return Marker{
			machinery.Marker{
				prefix:  markerPrefix(myPrefix),
				comment: comment,
				value:   value,
			},
		}
	}
	panic(fmt.Errorf("unsupported file extension: %s", ext))
}

func markerPrefix(prefix string) string {
	trimmed := strings.TrimSpace(prefix)
	var builder strings.Builder
	if !strings.HasPrefix(trimmed, "+") {
		builder.WriteString("+")
	}
	builder.WriteString(trimmed)
	if !strings.HasSuffix(trimmed, ":") {
		builder.WriteString(":")
	}
	return builder.String()
}
```

```go
// plugin/scaffold.go
package plugin

import (
	"strings"

	"sigs.k8s.io/kubebuilder/v4/pkg/machinery"
)

type RustMainFile struct {
	machinery.TemplateMixin
}

var _ machinery.Template = &RustMainFile{}

func (f *RustMainFile) SetTemplateDefaults() error {
	if f.Path == "" {
		f.Path = "main.rs"
	}

	f.TemplateBody = `{{ .ScaffoldMarker }}
fn main() {
    println!("Hello, operator!");
}
`

	m := NewRustMarker(f.Path, "mainFunc")
	f.TemplateBody = strings.ReplaceAll(f.TemplateBody, "{{ .ScaffoldMarker }}", m.String())

	return nil
}
```

```go
// main.go
package main

import (
	"sigs.k8s.io/kubebuilder/v4/pkg/plugin"
	"sigs.k8s.io/kubebuilder/v4/pkg/plugin/external"

	"myplugin/plugin"
)

func main() {
	external.Run(&external.Plugin{
		Name:    "my-rust.plugin",
		Version: plugin.Version{Number: 1},
		SupportedProjectVersions: []string{"3"},
		DevPhase: plugin.Alpha,

		GetScaffolder: func(c plugin.Context, args []string) (plugin.Scaffolder, error) {
			return &external.Scaffolder{
				Templates: []machinery.Template{
					&plugin.RustMainFile{},
				},
			}, nil
		},
	})
}
```

---

Again, HUGE thanks for contributing — we love seeing people build new ideas like this.  
Keep going, and let’s grow the ecosystem together 🙌 — and yes, we will absolutely help promote these external plugins! 💖

pkg/machinery/marker.go
```diff
@@ -28,6 +28,7 @@ var commentsByExt = map[string]string{
 	".go":   "//",
 	".yaml": "#",
 	".yml":  "#",
+	".rs":   "//",
```
camilamacedo86, line 31: We cannot add a file that we do not support have officially
Could the extension not be added in your code using the lib?

mabulgu: Yeah no problem, I got that:) What I meant: I will try what you suggest above, and actually the way I had implemented it was like what you suggested above. Let me revert to what I had before and compare with your suggestion and see if that works. I dont remember why I avoided that approach, maybe just not to duplicate things (thinking I can have the `rs` extension definition in kubebuilder)

mabulgu: @camilamacedo86 ok I've tried and remembered the reason as I tried to use the same approach. Here it is:

Screenshot 2025-05-25 at 22 04 02

mabulgu: So in other words, if we can make the marker struct use public fields instead of private ones, any other plugins can use their own custom markers. 

I can create a PR for this if this sounds good to you?

mabulgu: I tried out what I mentioned above here: https://github.com/mabulgu/kubebuilder/commit/4a618554cae2a33c7dcf8a76613d0395d412525a

But getting the following:

Screenshot 2025-05-26 at 00 27 50

So this means we need to extend `NewMarkerFor` as well. I will let you kow once I have a solution for this.

camilamacedo86: Yes, sure. Thank you a lot !!!
Yes, we need to allow people does that and create their own markers for their own extensions.
Please, feel free to raise the PR.

camilamacedo86:
Hi @mabulgu 

I am closing this one based on the discussion above: https://github.com/kubernetes-sigs/kubebuilder/pull/4827#discussion_r2106277161 and the conversation. 

However, please feel free to re-open if you see that it is required.

PR #4832, opened 2025-05-26
📖 (docs): Upgrade controller-tools from v0.17.4 to v0.18.0

camilamacedo86:
Upgrade the docs to be generated from the latest released version of controller-tools.


camilamacedo86:
I am moving forward.
I will not make nobody spend time with this small one 👍

PR #4842, opened 2025-06-01
🐛 (cli): Fixed an issue where `alpha generate` skipped files and directories beginning with a dot (e.g., `.github/`, `.env`, `.golangci.yaml`). These are now properly removed during cleanup, except `.git` and `PROJECT`, which are intentionally preserved.

camilamacedo86:
Alpha generate should re-scaffold the full project.
This bug does not allow users to get changes, for example, in GitHub actions.


camilamacedo86:
Fix :-) JFY: @vitorfloriano

camilamacedo86:
Moving forward with the fix since it is too small.

PR #4843, opened 2025-06-01
🐛 fix(cli): ensure 'alpha generate' runs 'make manifests generate fmt vet lint-fix' after scaffolding

camilamacedo86:
The command should ensure that all manifest and default commands are properly executed as a post-scaffold action.


camilamacedo86:
Fix :-) JFY: @vitorfloriano

camilamacedo86:
Moving forward with the fix since it is too small.

Issue #4838, opened 2025-05-27
Fix `--input-dir` in `kubebuilder alpha generate` and add test coverage


camilamacedo86:
### What broke? What's expected?

Unable to use the command with this option.

The `--input-dir` flag in `kubebuilder alpha generate` doesn't work as expected.

Even when a valid directory is passed (containing a `PROJECT` file), it fails with:

```
failed to read PROJECT file: open PROJECT: no such file or directory
```

This appears to be a path resolution bug — likely due to reliance on current working directory inside `LoadFrom`.

### Reproducing this issue

```bash
# 1. Go to the testdata project directory
cd testdata/project-v4

# 2. Create a new subdirectory to simulate a relocated project
mkdir .tmp

# 3. Move all project files (including PROJECT) into that directory
mv * .tmp/

# 4. Attempt to generate using the relocated directory as input
kubebuilder alpha generate --input-dir .tmp
```

### KubeBuilder (CLI) Version

master 

### PROJECT version

3

### Plugin versions

```yaml

```

### Other versions

We should:

- Fix the logic to read the `PROJECT` file using the full path: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/cli/alpha/command.go
- Add a test to validate this behavior: https://github.com/kubernetes-sigs/kubebuilder/tree/master/test/e2e/alphagenerate
This will ensure `--input-dir` works reliably and independently of the working directory.

PR #4844, opened 2025-06-01
:bug: (cli) fix: allow 'alpha generate' to work with unsupported plugins by patching PROJECT file in-memory

camilamacedo86:
This change enables the 'kubebuilder alpha generate' command to work with projects that use old and unsupported plugin versions (e.g., go.kubebuilder.io/v2 or v3). It does so by loading the PROJECT config file into memory and replacing deprecated plugin keys with the latest supported version (v4) before parsing.

This prevents config loading errors and preserves support for CLI flags like --help and others.

Related to: https://github.com/kubernetes-sigs/kubebuilder/issues/4433


camilamacedo86:
Hi @sarthaksarthak9  

I am pinging you here for your learning. **The fault is mine**. I should get it in the review. 
The way that we solved https://github.com/kubernetes-sigs/kubebuilder/pull/4572 broke and introduced the regressions https://github.com/kubernetes-sigs/kubebuilder/issues/4838, which also means we are missing coverage for those with tests. :-( 

Hi @anshuman-agarwala 

I think I'm on the right track to solve this one. 
But it would be nice if we could either cover the command better with tests,
If you want help, that would be great!

For example, we should have tests using all options

anshuman-agarwala:
Hi @camilamacedo86 
I'd love to help out, shall I add the coverage in the same PR or should I create a new one?

camilamacedo86:
Hi @anshuman-agarwala 

> I'd love to help out, shall I add the coverage in the same PR or should I create a new one?

That is amazing 🎉 
See that we added an test here to ensure that we are no longer broken the OPTIONS: https://github.com/kubernetes-sigs/kubebuilder/pull/4844/files#diff-a5b72bbb61f5511f73d12ebbb5d34289393500cf36225d3613ef37c2470a94d2R38-R49 ( see the GitHub Action )

But I think would be greate if we could add 1 test scenario using: `--input-dir` flag.
See our https://github.com/kubernetes-sigs/kubebuilder/tree/master/test/e2e/alphagenerate
Would you like to work on it? If so, I think we can do that in a follow-up scenario, as we added the e2e tests here to ensure the fix.

anshuman-agarwala:
> Hi @anshuman-agarwala
> 
> > I'd love to help out, shall I add the coverage in the same PR or should I create a new one?
> 
> That is amazing 🎉 See that we added an test here to ensure that we are no longer broken the OPTIONS: https://github.com/kubernetes-sigs/kubebuilder/pull/4844/files#diff-a5b72bbb61f5511f73d12ebbb5d34289393500cf36225d3613ef37c2470a94d2R38-R49 ( see the GitHub Action )
> 
> But I think would be greate if we could add 1 test scenario using: `--input-dir` flag. See our https://github.com/kubernetes-sigs/kubebuilder/tree/master/test/e2e/alphagenerate Would you like to work on it? If so, I think we can do that in a follow-up scenario, as we added the e2e tests here to ensure the fix.

Sure, I'll add the check in a new PR.

camilamacedo86:
I am moving with this one to help out in the related iniciatives

PR #4845, opened 2025-06-01
:sparkles: Add custom path option for webhooks

### LINKED ISSUE #4295: Custom path for webhooks

Hello, I'm coming here to talk about these closed issues #1436 & #1333 that were considered as a bug.

The issue is the following: when I generate the webhook file with `kubebuilder create api`, an annotation is automatically generated in this form: `//+kubebuilder:webhook:path=/validate-xxx-xxx-v1-xxx,mutating=false,...`. This annotation can make the developer think that he can actually modify the path (which is not the case as explained [here](https://book.kubebuilder.io/reference/markers/webhook)).

The previous issues where blocked mostly because the `controller-runtime` project didn't implement the custom path feature. However, I recently made a [PR](https://github.com/kubernetes-sigs/controller-runtime/pull/2998) that have been merged. This PR adds the ability for the developer to choose to use a custom path for his webhooks.

My idea here is to let the developer choose a custom path if he wants to in his `kubebuilder` project. This choice will be made **only** when calling the creation command (`kubebuilder create webhook`) with a `--custom-path` flag. By using this flag, `kubebuilder` will generate an annotation including the custom path of the developer (`//+kubebuilder:webhook:path=/my-custom-path,mutating=false,...`).

If this feature is accepted, I would like to work on this issue.

### Extra Labels

_No response_

damsien:
Closes #4295 ! This PR brings a way to configure custom webhook path for the defaulting and validating webhooks in the `kubebuilder` cli.

Two new flags are added for the `kubebuilder create webhook` command: `--defaulting-custom-path` & `--validating-custom-path`. Both can be used independently.

Usage:
```sh
kubebuilder create webhook \
    --group mygroup.io --kind MyResource --version v1alpha1 \
    --defaulting --programmatic-validation \
    --defaulting-custom-path=/my-custom-defaulting-webhook-path \
    --validating-custom-path=/my-custom-validating-webhook-path
```

When used, it will:
1. Change the `// +kubebuilder:webhook:path` go marker.
2. Add `WithValidatorCustomPath(...).` or/and `WithDefaulterCustomPath(...).` to the webhook setuper function.

Result from the example above:
```go
func SetupMyResourceWebhookWithManager(mgr ctrl.Manager) error {
	return ctrl.NewWebhookManagedBy(mgr).For(&mygroupiov1alpha1.MyResource{}).
		WithValidator(&MyResourceCustomValidator{}).
		WithDefaulter(&MyResourceCustomDefaulter{}).
		WithValidatorCustomPath("/my-custom-validating-webhook-path").
		WithDefaulterCustomPath("/my-custom-defaulting-webhook-path").
		Complete()
}

// +kubebuilder:webhook:path=/my-custom-defaulting-webhook-path,mutating=true,failurePolicy=fail,sideEffects=None,groups=mygroup.io,resources=myresources,verbs=create;update,versions=v1alpha1,name=mmyresource-v1alpha1.kb.io,admissionReviewVersions=v1

...

// +kubebuilder:webhook:path=/my-custom-validating-webhook-path,mutating=false,failurePolicy=fail,sideEffects=None,groups=mygroup.io,resources=myresources,verbs=create;update,versions=v1alpha1,name=vmyresource-v1alpha1.kb.io,admissionReviewVersions=v1
```

### DIFF
--- a/pkg/plugins/golang/options.go
+++ b/pkg/plugins/golang/options.go
@@ -71,6 +71,12 @@ type Options struct {
 	DoValidation bool
 	DoConversion bool
 
+	// ValidatingWebhookCustomPath defines the custom path that will be used by the scaffolded validating webhooks
+	ValidatingWebhookCustomPath string
+
+	// DefaultingWebhookCustomPath defines the custom path that will be used by the scaffolded defaulting webhooks
+	DefaultingWebhookCustomPath string
+
 	// Spoke versions for conversion webhook
 	Spoke []string
 }
@@ -108,6 +114,14 @@ func (opts Options) UpdateResource(res *resource.Resource, c config.Config) {
 			res.Webhooks.Conversion = true
 			res.Webhooks.Spoke = opts.Spoke
 		}
+	}
+
+	if opts.ValidatingWebhookCustomPath != "" {
+		res.Webhooks.ValidatingCustomPath = opts.ValidatingWebhookCustomPath
+	}
+
+	if opts.DefaultingWebhookCustomPath != "" {
+		res.Webhooks.DefaultingCustomPath = opts.DefaultingWebhookCustomPath
 	}
 
 	if len(opts.ExternalAPIPath) > 0 {
--- a/pkg/model/resource/webhooks.go
+++ b/pkg/model/resource/webhooks.go
@@ -35,6 +35,12 @@ type Webhooks struct {
 	Conversion bool `json:"conversion,omitempty"`
 
 	Spoke []string `json:"spoke,omitempty"`
+
+	// ValidatingCustomPath specifies a custom path for the validating webhook
+	ValidatingCustomPath string `json:"validatingCustomPath,omitempty"`
+
+	// DefaultingCustomPath specifies a custom path for the defaulting webhook
+	DefaultingCustomPath string `json:"defaultingCustomPath,omitempty"`
 }
 
 // Validate checks that the Webhooks is valid.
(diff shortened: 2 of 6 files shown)


pkg/plugins/golang/options.go
```diff
@@ -71,6 +71,12 @@ type Options struct {
 	DoValidation bool
 	DoConversion bool
 
+	// ValidatingWebhookCustomPath defines the custom path that will be used by the scaffolded validating webhooks
+	ValidatingWebhookCustomPath string
+
+	// DefaultingWebhookCustomPath defines the custom path that will be used by the scaffolded defaulting webhooks
+	DefaultingWebhookCustomPath string
```
camilamacedo86, line 78: That’s great work, thank you! 🎉
The implementation looks solid at a glance.

**However, since this introduces API changes, I’ll need a bit more time to review it carefully and fully assess the impact.**

A couple of things we should also make sure of before merging:

- This should be compatible with the alpha generate command — we’ll likely need to include logic to ensure the project is re-scaffolded correctly with the new option.

- It should also work seamlessly with the Helm charts plugin (helm.kubebuilder.io/v1-alpha) — could you confirm compatibility or add test coverage for that?

Thanks again for the contribution! I’ll follow up with a more detailed review soon.

damsien: Thank you for your quick comment!

> we’ll likely need to include logic to ensure the project is re-scaffolded correctly with the new option.

What do you mean by "the project is re-scaffolded"?

> could you confirm compatibility or add test coverage for that?

I didn't test it with the helm chart plugin. Also, since I never worked with the helm plugin, I don't really know where to start to confirm the compatibility :sweat_smile:. Could you please provide me the test file where I should write the new test?

I don't really understand where there should be an incompatibility with the helm plugin since my feature is only related to the webhook file that is created using the CLI (actually it only affects the content of the webhook file, and it does not scaffold any other file). The feature is not related at all with the delivery of the user's operator.

camilamacedo86:
This PR has been open for quite a while without further engagement to move it forward.
To keep our list of open PRs clear and focused on those still needing review, I’m closing this one.

If you’re able to continue the work, please feel free to reopen it.

PR #4848, opened 2025-06-01
🐛 (cli)(fix): postscaffold for alpha generate should return warning (follow-up: 4843)

camilamacedo86:
Follow up #4843 ( not released yet - I realised that I committed a mistake that could break automations using it )


camilamacedo86:
Move forward since it is tiny, and do not let the master with the issue since it impacts the tests with other iniciatives

PR #4860, opened 2025-06-06
:sparkles: (ux/cli) Improve `kubebuilder version` info output formatting

vitorfloriano:
Updates `versionString()` to return a string with better formatting and increased readability. This change also helps users provide version info when reporting bugs.

- Currently, the `kubebuilder version` command outputs version info the following format:
```
vitorfloriano@pc:~/go/src/github.com/vitorfloriano/kubebuilder$ ./bin/kubebuilder version
Version: cmd.version{KubeBuilderVersion:"v4.6.0", KubernetesVendor:"1.33.0", GitCommit:"1d79aa1ec8204a11ae6be06cb96bae77dd0210bf", BuildDate:"2025-06-06T19:38:06Z", GoOs:"linux", GoArch:"amd64"}
```
- This PR improves formatting to a more readable multi-line layout, similar to other CLI tools:
```
vitorfloriano@pc:~/go/src/github.com/vitorfloriano/kubebuilder$ ./bin/kubebuilder version
Kubebuilder Version: v4.6.0-dirty
Kubernetes Vendor: 1.33.0
OS: linux
Arch: amd64
Git Commit: 1d79aa1ec8204a11ae6be06cb96bae77dd0210bf
Build Date: 2025-06-06T19:38:34Z
```
This is a small but meaningful quality-of-life improvement for the user experience.

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -53,14 +53,18 @@ func versionString() string {
 		}
 	}
 
-	return fmt.Sprintf("Version: %#v", version{
+	return fmt.Sprintf(`Kubebuilder Version: %s
+Kubernetes Vendor: %s
+OS: %s
+Arch: %s
+Git Commit: %s
+Build Date: %s`,
 		kubeBuilderVersion,
 		kubernetesVendorVersion,
-		gitCommit,
-		buildDate,
 		goos,
 		goarch,
-	})
+		gitCommit,
+		buildDate)
 }
 
 // getKubebuilderVersion returns only the CLI version string


cmd/version.go
```diff
@@ -53,14 +53,18 @@ func versionString() string {
 		}
 	}
 
-	return fmt.Sprintf("Version: %#v", version{
```
mateusoliveira43: with this change, can `type version` (in line 39) be deleted?

vitorfloriano: That's a good observation. 

But yeah, looking back at it again, it seems to me that my approach was just a quick fix for a pet peeve of mine. Probably we shouldn't even put version info output in this package, since it's a feature of the `version` subcommand. 

This file (`cmd/version.go`) should just gather data as the 'source of truth' for version info, and that's it. Version output and formatting should belong to (`cli/version.go`). 🤔 

WDYT, @mateusoliveira43 ?

mateusoliveira43: hmm I think it makes sense yes :)

vitorfloriano:
Oops! I forgot I had this PR open on `master` and ended up force-pushing unrelated work. My bad! I’ll open a clean PR later with the original intent. Sorry for the noise.

PR #4868, opened 2025-06-12
🐛 (go/v4): e2e tests SA spec name is not accurate to inform the SA

camilamacedo86:


camilamacedo86:
Moving forward since it is small fix and blocker other ones.

PR #4869, opened 2025-06-12
🐛 (go/v4): ensure manager and curl pod use readOnlyRootFilesystem to comply with PSA rules

camilamacedo86:
Kubernetes 1.25+ with Pod Security Admission (PSA) in restricted mode requires containers to run with readOnlyRootFilesystem=true. This updates both the manager deployment and the test-created curl pod accordingly.


camilamacedo86:
I am moving forward with this one since it is small and is required to fix ci.

PR #4876, opened 2025-06-19
✨ feat(makefile): add setup test e2e targets

kersten:
Currently, the test-e2e target in the Kubebuilder Makefile template hardcodes a single workflow: setup the Kind cluster and immediately run the end-to-end tests. This prevents projects from injecting any preparatory steps - such as installing CRDs, operators, or other third-party dependencies - before executing the actual tests.

By introducing the SETUP\_TEST\_E2E\_TARGETS variable, users gain the flexibility to define and run one or more prerequisite targets prior to cluster setup and test execution. This change:
*   **Enhances extensibility**: Teams can now tailor the e2e pipeline to their specific needs without forking or patching the upstream Makefile.
*   **Maintains backward compatibility**: The default behavior remains unchanged when SETUP\_TEST\_E2E\_TARGETS is unset, ensuring existing projects continue to work seamlessly.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/Makefile
+++ b/docs/book/src/cronjob-tutorial/testdata/project/Makefile
@@ -71,6 +71,9 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
 # - CERT_MANAGER_INSTALL_SKIP=true
 KIND_CLUSTER ?= project-test-e2e
 
+# Optional targets to run after setting up Kind cluster and before running e2e tests
+SETUP_TEST_E2E_TARGETS ?=
+
 .PHONY: setup-test-e2e
 setup-test-e2e: ## Set up a Kind cluster for e2e tests if it does not exist
 	@command -v $(KIND) >/dev/null 2>&1 || { \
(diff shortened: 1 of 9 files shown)


docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -71,6 +71,9 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
 # - CERT_MANAGER_INSTALL_SKIP=true
 KIND_CLUSTER ?= project-test-e2e
 
+# Optional targets to run after setting up Kind cluster and before running e2e tests
+SETUP_TEST_E2E_TARGETS ?=
```
camilamacedo86, line 75: Thank you for this contribution 🥇 

But let's check the motivation:

> By introducing the SETUP_TEST_E2E_TARGETS variable, users gain the flexibility to define and run one or more prerequisite targets prior to cluster setup and test execution. This change:

Do we have any prerequisites in the default scaffold? 

> Currently, the test-e2e target in the Kubebuilder Makefile template hardcodes a single workflow: setup the Kind cluster and immediately run the end-to-end tests. This prevents projects from injecting any preparatory steps - such as installing CRDs, operators, or other third-party dependencies - before executing the actual tests.

The key question here is: Do we need to support this in the default scaffold?

I would argue no. This kind of setup is only necessary in customised use cases. If someone chooses to modify the default scaffold to accommodate these advanced workflows, they can and should customise the Makefile accordingly as well.

In other words, we shouldn’t change the default scaffold to support edge-case scenarios that are irrelevant to the standard use case. Doing so introduces unnecessary complexity and maintenance burden, with features that don’t offer value to the common path.

camilamacedo86: /hold for we discuss the need and motivations ^

camilamacedo86:
Hi @kersten 

Thank you a lot not only for this one but for the many contributions made 🎉 
Due to the reasons described in https://github.com/kubernetes-sigs/kubebuilder/pull/4876/files#r2156536612, I am closing this one as deferred. However, if you believe we should move forward and disagree, please feel free to reopen.

PR #4883, opened 2025-06-24
🐛 fix go/v4 e2e tests: skip kind node creation if it exist already

camilamacedo86:

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/Makefile
+++ b/docs/book/src/cronjob-tutorial/testdata/project/Makefile
@@ -74,10 +74,16 @@ KIND_CLUSTER ?= project-test-e2e
 .PHONY: setup-test-e2e
 setup-test-e2e: ## Set up a Kind cluster for e2e tests if it does not exist
 	@command -v $(KIND) >/dev/null 2>&1 || { \
-		echo "Kind is not installed. Please install Kind manually."; \
+		echo "Error: 'kind' is not installed. Please install Kind manually."; \
 		exit 1; \
 	}
-	$(KIND) create cluster --name $(KIND_CLUSTER)
+	@case "$$($(KIND) get clusters)" in \
+		*"$(KIND_CLUSTER)"*) \
+			echo "Kind cluster '$(KIND_CLUSTER)' already exists. Skipping creation." ;; \
+		*) \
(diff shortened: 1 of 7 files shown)


docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -74,10 +74,16 @@ KIND_CLUSTER ?= project-test-e2e
...
 setup-test-e2e: ## Set up a Kind cluster for e2e tests if it does not exist
 	@command -v $(KIND) >/dev/null 2>&1 || { \
-		echo "Kind is not installed. Please install Kind manually."; \
+		echo "Error: 'kind' is not installed. Please install Kind manually."; \
 		exit 1; \
 	}
-	$(KIND) create cluster --name $(KIND_CLUSTER)
+	@case "$$($(KIND) get clusters)" in \
+		*"$(KIND_CLUSTER)"*) \
+			echo "Kind cluster '$(KIND_CLUSTER)' already exists. Skipping creation." ;; \
```
afzal442, line 82:
```suggestion
			echo "Warning: Kind cluster '$(KIND_CLUSTER)' already exists. Skipping creation." ;; \
```

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -74,10 +74,16 @@ KIND_CLUSTER ?= project-test-e2e
...
+		*) \
+			echo "Creating Kind cluster '$(KIND_CLUSTER)'..."; \
```
afzal442, line 84:
```suggestion
			echo "Info: Creating Kind cluster '$(KIND_CLUSTER)'..."; \
```

docs/book/src/cronjob-tutorial/testdata/project/Makefile
camilamacedo86: That makes me think that we should not add `"Error: 'kind' ' since we do not use it in other places, and the need to add information or a warning seems a little excessive. It would be better if we leave it as clean as possible. So, thank you for the review. I will remove the above

afzal442: Sure! Thanks

PR #4889, opened 2025-06-26
🛠️  add ignore flags for selective scaffolding in edit command

### LINKED ISSUE #4326: (helm/v1-alpha) Add ignore flags when generating charts

Motivated by this [discussion](https://github.com/kubernetes-sigs/kubebuilder/pull/4314#issuecomment-2476995558).

Today, when we want to use the helm plugin, the `values.yaml` is scaffolled with many defaults settings that the develope may not want to have (promtheus, some samples CRs, ...).

Our proposal is to implement `ignore-` flags that could be used when generating the `values.yaml`. The flags could be the following:
- `ignore-samples`
- `ignore-prometheus`
- `ignore-networkPolicy`
- `ignore-certmanager`
- `ignore-webhook`

## How to implement it?

Here is my proposal:
1. Add the flags here:
https://github.com/kubernetes-sigs/kubebuilder/blob/e9beffb5926f8fcf1abf97ada7ae3a3c675d0591/pkg/plugins/optional/helm/v1alpha/edit.go#L31-L34

2. Add one function per flag in [this file](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go) in order to not add the settings to the scaffold if the `ignore` flag is specified

### Extra Labels

_No response_

pabhi18:
## Description:
This PR introduces support for the following `--ignore-*` flags in the `edit` command of the Helm plugin, allowing selective scaffolding of chart files:

- `--ignore-samples`  
- `--ignore-prometheus`  
- `--ignore-networkPolicy`  
- `--ignore-certmanager`  
- `--ignore-webhook`

It can be used like this:

```bash
kubebuilder edit --plugins=helm.kubebuilder.io/v1-alpha \
  --ignore-samples \
  --ignore-prometheus \
  --ignore-networkPolicy \
  --ignore-certmanager \
  --ignore-webhook
  ```
These flags only apply to the edit command. The init command still scaffolds all components by default.
If any of the scaffolded folders are deleted manually, running edit with the respective --ignore-* flag will prevent them from being regenerated.

## Motivation:
Developers may not want all default Helm chart components (like Prometheus monitors or sample CRs). This change offers a way to skip unwanted parts cleanly.

## Fixes:
Fixes #4326

### DIFF
--- a/pkg/plugins/optional/helm/v1alpha/edit.go
+++ b/pkg/plugins/optional/helm/v1alpha/edit.go
@@ -31,6 +31,12 @@ var _ plugin.EditSubcommand = &editSubcommand{}
 type editSubcommand struct {
 	config config.Config
 	force  bool
+
+	IgnoreSamples       bool
+	IgnorePrometheus    bool
+	IgnoreNetworkPolicy bool
+	IgnoreCertManager   bool
+	IgnoreWebhook       bool
 }
 
 //nolint:lll
@@ -66,6 +72,12 @@ manifests in the chart align with the latest changes.
 
 func (p *editSubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&p.force, "force", false, "if true, regenerates all the files")
+
+	fs.BoolVar(&p.IgnoreSamples, "ignore-samples", false, "ignore scaffolding of sample CRs")
+	fs.BoolVar(&p.IgnorePrometheus, "ignore-prometheus", false, "ignore scaffolding of prometheus monitor")
+	fs.BoolVar(&p.IgnoreNetworkPolicy, "ignore-networkPolicy", false, "ignore scaffolding of network policy")
+	fs.BoolVar(&p.IgnoreCertManager, "ignore-certmanager", false, "ignore scaffolding of cert-manager manifests")
+	fs.BoolVar(&p.IgnoreWebhook, "ignore-webhook", false, "ignore scaffolding of webhooks")
 }
 
 func (p *editSubcommand) InjectConfig(c config.Config) error {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pkg/plugins/optional/helm/v1alpha/edit.go [hunk 1]
```diff
@@ -31,6 +31,12 @@ var _ plugin.EditSubcommand = &editSubcommand{}
 type editSubcommand struct {
 	config config.Config
 	force  bool
+
+	IgnoreSamples       bool
```
camilamacedo86: I think we are not adding the samples so we would not need the flag

pkg/plugins/optional/helm/v1alpha/init.go
```diff
@@ -47,7 +47,15 @@ func (p *initSubcommand) InjectConfig(c config.Config) error {
...
 
 func (p *initSubcommand) Scaffold(fs machinery.Filesystem) error {
-	scaffolder := scaffolds.NewInitHelmScaffolder(p.config, false)
+	ignoreFlags := map[string]bool{
+		"ignore-samples":       false,
+		"ignore-prometheus":    false,
+		"ignore-networkPolicy": false,
+		"ignore-certmanager":   false,
+		"ignore-webhook":       false,
+	}
```
camilamacedo86, line 55: To make this work, we’d need to ensure that we have none conditions using them in the scaffold as well.
For example, see the values.ymal and the other helm templates, we have conditions that check if has webhooks and etc.. handling those isn’t as straightforward as it seems. 

I know we opened an issue about this in the past, but I’m not sure how feasible or maintainable it is long-term. We’d also need a way to validate and guarantee support across all options.

If we go ahead with this, we should: ( we cannot add the options without ensure that all works and is done as expected in each case )
- Cover all relevant scaffold points
- Add CI tests to validate configs for each scenario
- Mock Helm charts and install them during e2e to catch issues early

We could follow the same pattern used in our `e2e` tests: create mocks per case, deploy them, and verify outcomes. See `e2e/go/v4` and related GitHub Actions as reference.

pabhi18: Thank you for the clarification!
Yes, currently the flags only prevent the files from being regenerated, but the corresponding checks in values.yaml and templates still need to be handled.
I’ll start working on adding the necessary conditions for those.
If you have any suggestions or references for adding those conditions or e2e tests, it would really help me a lot!

camilamacedo86:
Fixed with [#5058](https://github.com/kubernetes-sigs/kubebuilder/pull/5058)
Closing it

PR #4890, opened 2025-06-26
✨ Make webhook and metrics ports configurable in Helm chart template

### LINKED ISSUE #4802: (helm/v1-alpha) override value for webhook port

Webhook port seems to be harcoded to `9443` in [Service](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/chart-templates/webhook/service.go#L58-L60) and [Deployment](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/chart-templates/manager/manager.go#L108) in the generated chart.

If a user changes the port in `config` and overrides it in the manager `webhook-port ` argument, chart will still have the old value.

### KubeBuilder (CLI) Version
4.5.2

### Extra Labels

_No response_

Okabe-Junya:
close #4802 

**What I changed**

- **values.yaml**: Added `webhook.port: 9443` and `metrics.port: 8443` configuration options
- **Templates**: Updated Service and Deployment templates to reference `{{ .Values.webhook.port }}` and `{{ .Values.metrics.port }}`

### DIFF
--- a/docs/book/src/getting-started/testdata/project/dist/chart/values.yaml
+++ b/docs/book/src/getting-started/testdata/project/dist/chart/values.yaml
@@ -9,7 +9,6 @@ controllerManager:
       - "--leader-elect"
       - "--metrics-bind-address=:{{ .Values.metrics.port }}"
       - "--health-probe-bind-address=:8081"
-      - "--webhook-port={{ .Values.webhook.port }}"
     resources:
       limits:
         cpu: 500m
--- a/pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/values.go
+++ b/pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/values.go
@@ -62,8 +62,8 @@ controllerManager:
     args:
       - "--leader-elect"
       - "--metrics-bind-address=:{{ "{{" }} .Values.metrics.port {{ "}}" }}"
-      - "--health-probe-bind-address=:8081"
-      - "--webhook-port={{ "{{" }} .Values.webhook.port {{ "}}" }}"
+      - "--health-probe-bind-address=:8081"{{ if .HasWebhooks }}
+      - "--webhook-port={{ "{{" }} .Values.webhook.port {{ "}}" }}"{{ end }}
     resources:
       limits:
         cpu: 500m


pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/chart-templates/manager/manager.go
```diff
@@ -105,7 +105,7 @@ spec:
 {{- if .HasWebhooks }}
           {{ "{{- if .Values.webhook.enable }}" }}
           ports:
-            - containerPort: 9443
+            - containerPort: {{ "{{ .Values.webhook.port }}" }}
```
camilamacedo86, line 108: This ports as fixed for golang in the scaffold.
We cannot change the port only for helm
That will not work if you add other port. 
Also, see: https://github.com/kubernetes-sigs/kubebuilder/issues/4802

Okabe-Junya: Thank you, ah... I see, Understood.
I'll think about it for a moment 🙇

Okabe-Junya: Sorry, I finally understand how KubeBuilder's scaffold works.
I believe it's now configurable with this diff.

- https://github.com/kubernetes-sigs/kubebuilder/pull/4890/commits/5c69ddc302d0c4fb653fa249c70a3d8ef5d0c237#diff-2a3c437dc2f2baf7c32e26a12c886f273e5df193b2c9d1dc7881a76f04922177

Could you please review it again? Thank you

Okabe-Junya:
Oh, I woke up and the e2e test failed...
I'll check it later

docs/book/src/cronjob-tutorial/testdata/project/cmd/main.go
```diff
@@ -95,6 +96,7 @@ func main() {
 	flag.StringVar(&webhookCertPath, "webhook-cert-path", "", "The directory that contains the webhook certificate.")
 	flag.StringVar(&webhookCertName, "webhook-cert-name", "tls.crt", "The name of the webhook certificate file.")
 	flag.StringVar(&webhookCertKey, "webhook-cert-key", "tls.key", "The name of the webhook key file.")
+	flag.IntVar(&webhookPort, "webhook-port", 9443, "The port that the webhook endpoint binds to.")
```
camilamacedo86, line 99: Should it also be a string, right?

Okabe-Junya:
I'm really sorry....

I completely missed the notification (fortunately, I noticed it thanks to the conflict alert)
I'll work on it asap. I’m really sorry again

camilamacedo86:
Hi @Okabe-Junya 

Thank you a lot for this contribution. 🥇 
See that we are moving forward and deprecating helm/v1alpha in favor of helm/v2alpha -> See; https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/optional/helm/v2alpha

The new version consumes the kustomize install.yaml ( final result ) to build the chart
Therefore, the configuration came from kustomize

If you are able to customise the ports via kustomize, then we should have the option here.
So, because it is open for a while and is outdated, I am closing this one.

However, please feel free to check the new version to see if your need is not addressed
If not, then we need to address it for  v2alpha. Could you please raise an issue about this? 

Also, feel free to re-open if you see that it is needed. 
Again, thank you for your contribution !!!

PR #4891, opened 2025-06-27
✨ helm chart directory can be configured

### LINKED ISSUE #4320: target helm chart directory should be configurable

I gave the very early `helm.kubebuilder.io/v1-alpha` plugin a try. 

One downside with the generated manifests is the current chosen file path for the helm charts.
All the content of the generated helm charts goes to `./dist` folder on.
(xref: [relevant code places](https://github.com/search?q=repo%3Akubernetes-sigs%2Fkubebuilder+filepath.Join%28%5C%22dist%5C%22&type=code)).

I very often use [goreleaser](https://goreleaser.com) for the release process.
And per default, all the generated build artifacts will be placed under the `./dist` folder. As these generated artifacts should not get tracked by git, we ignore the entire `./dist` folder in our `.gitignore` file.

Per documentation, it's possible in `goreleaser` to change that location (xref: [goreleaser doc](https://goreleaser.com/customization/dist/?h=dist#dist-folder)).
But as i do not prefer to change locations from an existing integrated tool, just because of integrating a new tool i would like to see to make the target folder configurable :pray: 

I can take over the implementation if wanted

### Extra Labels

_No response_

bavarianbidi:
Per default, helm-charts will get generated under dist/ directory. Via the newly introduced --directory flag it's now possible to define an alternative directory (e.g.: chart)

fixes #4320

_additional notes:_
I've added a `importas` golangci rule for the `helmv1alpha` package as the import is handled in multiple files differently.
If this is fine, i will also raise another PR to add at least two new importas-rules for

```
	"sigs.k8s.io/kubebuilder/v4/pkg/plugins/golang/deploy-image/v1alpha1"
	"sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/grafana/v1alpha"
```

as it's super hard to read the code without having proper naming for equal sounded imports :sweat_smile:


pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
```diff
@@ -28,12 +28,15 @@ var _ machinery.Template = &HelmHelpers{}
 type HelmHelpers struct {
 	machinery.TemplateMixin
 	machinery.ProjectNameMixin
+
+	// Directory is the directory where the Helm chart will be scaffolded
+	Directory string
```
bavarianbidi, line 33: I've also thought about adding a new `machinery.HelmMixin` which could be used for all helm-based structs, but decided to not do it for now as i wanted your feedback on this first.

pkg/plugins/optional/helm/v1alpha/edit.go
```diff
@@ -66,6 +68,7 @@ manifests in the chart align with the latest changes.
 
 func (p *editSubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&p.force, "force", false, "if true, regenerates all the files")
+	fs.StringVar(&p.directory, "directory", HelmDefaultTargetDirectory, "domain for groups")
```
camilamacedo86: I think we need to as we do for the alpha generate.
I mean, use the name output-dir see

https://github.com/kubernetes-sigs/kubebuilder/blob/202a630aba03b2f4869f3fcbcad3c97959050144/docs/book/src/reference/rescaffold.md?plain=1#L28-L32

Consistency is the key :-)

.golangci.yml
```diff
@@ -13,6 +13,7 @@ linters:
     - goconst
```
camilamacedo86, line 13: We will need to add a test to ensure all helm plugin options are set.
I haven't thought of the best way yet.
However, I think we might want to add  e2e tests , similar to the others  https://github.com/kubernetes-sigs/kubebuilder/tree/master/test/e2e ,to mock scenarios with helm and verify that everything is as expected.

pkg/plugins/optional/helm/v1alpha/edit.go
bavarianbidi: `output-dir` as parameter-name for the helm-plugin will became an issue for `alpha generate`.
As `output-dir` is a parameter for `alpha generate` we can't identify if a user defined `output-dir` for the `generate` or the `helm` plugin.

I currently see the following options:
* use `output-dir` as flag for the helm-plugin, and introduce an additional flag for `alpha generate` (e.g. `--helm-output-dir` where the value is getting propagated into the `--output-dir` of `kubebuilder alpha generate --plugins=helm/v1-alpha --output-dir=<value>`
* use `output-dir` as flag for the helm-plugin and use the default-value of `dist` when `kubebuilder alpha generate --plugins=helm/v1-alpha` is getting called (user can still later do an `kubebuilder edit --plugins=helm/v1-alpha --directory=<alternative dir>` if they are not happy with the default>
* introduce `helm-output-dir` as flag all over the place (`alpha generate` and `edit --plugins=helm/v1-alpha`)

with the argument of consistency, i would prefer the last one.

.golangci.yml
bavarianbidi: regarding e2e:
do you mean, my introduced changes in the e2e-tests aren't sufficient?

* https://github.com/kubernetes-sigs/kubebuilder/pull/4891/files#diff-cf2922923a38b0c2f4b62c6feb508f914d9c4e2ff6b0971d9b3e92f38166a5e9
* https://github.com/kubernetes-sigs/kubebuilder/pull/4891/files#diff-c0980e8e998bffb85cee1a67011d9c06ddecfbc5c0e48fcf6d3685f46f663004

do you want me to introduce a new e2e-test with a custom value for the helm output dir?
In the current impl. even the default-value will get persist in the `PROJECT` file.

pkg/plugins/optional/helm/v1alpha/edit.go
camilamacedo86: Hi @bavarianbidi it should not be an issue for alpha generate because it will re-scaffold what is configured in the PROJECT file.
If we add this option to helm, the input will be in the config with the path used
We need supplement alpha generate for call the new flag with the plugin value, and all will work out.

.golangci.yml
camilamacedo86: Hi @bavarianbidi 

I thought about this one and IMO: 

1 - We need to add the new flag to the alpha generate
2 - We need to add a new mock in alpha generation that includes the Helm plugin and will utilise this option. Then, at the end, we validate that the helm chart was output in the directory path provided. That should be enough to cover both scenarios in the scope of this PR

camilamacedo86: Please review how the alpha is generated and its associated tests. If you need a hand after that, please feel free to ping me in the Kubebuilder Slack channel. I am happy to help out.

bavarianbidi:
waiting for #4901

bavarianbidi:
converted to Draft (for now) and rebased on top of #4903
will continue once #4903 get merged

camilamacedo86:
Closing this one since this request was addressed with helm v2 alpha.

Issue #4760, opened 2025-04-05
Test cliVersion generation and addition before next Kubebuilder release (v4.x.x)


camilamacedo86:
We need to ensure the new cliVersion field added in [PR #4621](https://github.com/kubernetes-sigs/kubebuilder/pull/4621) is correctly populated in generated PROJECT config files and does not introduce any issues prior to the next Kubebuilder release.

Goals:

- Verify that cliVersion is correctly written to the PROJECT file during init.
- Confirm that cliVersion does not break loading of existing projects that don't have it.
- Ensure backward compatibility with v3 and other plugin versions where applicable.

We might need to:
- A pre-release/RC (if available)
- A local emulated release build (i.e., build the CLI locally and test with it)

camilamacedo86:
Hi @vitorfloriano,

We can't proceed with this change right now, as it would break our public API.

However, we are tracking this need, and once we have enough changes to justify a major release (Kubebuilder 5.0.0), we’ll make sure not to forget this one. You can see it's already being tracked under the Kubebuilder 5.x milestone.

vitorfloriano:
Hi @camilamacedo86 does that mean I should not work on this issue and choose something else to focus on for now, or does that mean that my PR may be put on hold until we hit the need for a major version release?

camilamacedo86:
HI @vitorfloriano 

What we can do here to ensure that all works fine is:
- Generete locally a new build (using goreleaser) 
- With the bin generated via go-releaser create a new project
- Then, check the PROJECT file

Could you please follow those steps and add the output to confirm that we will have the tag/version of the CLI set in the PROJECT file?

camilamacedo86:
If you want to push any unit test that is always welcome.
But to unblock us, we need to testout it with goreleaser and ensure that all is working as expected before the release. So a manual test can help out

vitorfloriano:
@camilamacedo86 I see what you mean now. Even after setting a build tag to something like `v4.5.3-rc.1` and running goreleaser to build the tool, the PROJECT file still shows `(devel)` instead of `v4.5.3-rc.1`. Will investigate.

vitorfloriano:
@camilamacedo86 So I found out I was running `goreleaser` without passing the flag `--config build/.goreleaser.yml`, which was causing it to fallback to a default behavior, hence the `(devel)` cliVersion. I ran it again passing the flag and it actually worked as expected:

Checking the tag:
```bash
[...]

PR #4516, opened 2025-01-21
:sparkles: [Improvement] Infer version information for go-install

migueleliasweb:
Now that `kubebuilder` is go-installable :tada:, this is a small improvement to fetch more information when dealing with `go install` installations.

Currently, the output from `kubebuilder version` looks like this:

```bash
$ go install sigs.k8s.io/kubebuilder/v4@7ee23df2b97ccb209df4e05c78830266c7b8d8bb
Version: cmd.version{KubeBuilderVersion:"v4.5.1-0.20250121092837-7ee23df2b97c", KubernetesVendor:"unknown", GitCommit:"$Format:%H$", BuildDate:"1970-01-01T00:00:00Z", GoOs:"unknown", GoArch:"unknown"}
```

With this PR, we will now infer correctly `GitCommit`, `BuildDate`, `GoOs` and `GoArch` for normal `GoReleaser` builds but also `go install` installations.

PR #4898, opened 2025-07-01
✨ Improve version command output: add runtime fallbacks and unit tests.

Thedarkmatter10:
### Summary
This PR improves the internal versioning logic by making it more maintainable, testable, and robust.

### ✨ What’s Changed
- Publicly exposed version helpers:

  - VersionInfo struct (was private version)

- Removed duplication (DRY):

  - Centralized resolution logic using debug.ReadBuildInfo() into a helper

  - Eliminated repeated fallback code for resolving build info

- Added runtime fallbacks:

  - Uses runtime.GOOS and runtime.GOARCH when not set via ldflags

  - Ensures gitCommit is validated and not left as $Format:%H$

- Improved output formatting:

  - versionString() provides clean, human-readable output

  - versionJSON() supports structured, machine-readable output (useful in CI)
- Renamed function Name from `getKubebuilderVersion()` to `getKubeBuilderVersion()` .

### Added unit tests (version_test.go):

 - Validate output includes expected fields

 - Confirm versionJSON() is parsable and includes required keys

 - Ensure GetVersionInfo() returns non-empty, resolved values

### 🧠 Why It Matters
- Ensures test coverage for a previously untested code path

- Helps catch missing ldflags values (e.g. empty Git commit) early

- Lays the foundation for long-term maintainability of CLI version reporting

### 📈 Impact
- Exposes version info for programmatic and CLI use

- Adds meaningful tests that reduce risk of silent version resolution failures

- Improves developer and CI experience with cleaner output and stronger guarantees


Thedarkmatter10:
@camilamacedo86 @Kavinjsir

cmd/version.go
```diff
@@ -17,58 +17,91 @@ limitations under the License.
...
+// getKubeBuilderVersion returns just the CLI version
+func getKubeBuilderVersion() string {
+	resolveBuildInfo()
 	return kubeBuilderVersion
 }
+
+// versionJSON returns version as JSON string
+func versionJSON() string {
+	v := getVersionInfo()
+	b, _ := json.MarshalIndent(v, "", "  ")
```
camilamacedo86, line 105: Will it change how we retrieve and store the data in the PROJECT file?

Thedarkmatter10: Thanks for the question! 🙌
No, this change does not affect how we retrieve or store data in the PROJECT file.

The versionJSON() and related functions strictly rely on:

- Build-time ldflags (e.g., gitCommit, buildDate)

- runtime.GOOS / GOARCH

- debug.ReadBuildInfo() (from Go module metadata)

There's no read or write interaction with the PROJECT file — this change only impacts version metadata exposed at runtime via the CLI.
@camilamacedo86

camilamacedo86: I think we use this information to add the version in https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/PROJECT#L5 

So, we might need to test out to TDB:

We need to build the bin locally with goreleaser and then scaffold a new project with.
Like done here: https://github.com/kubernetes-sigs/kubebuilder/issues/4760#issuecomment-2798358235 to ensure that we will still be outputting only the version there. 

https://github.com/kubernetes-sigs/kubebuilder/actions/runs/16007968823/job/45252541919?pr=4898

camilamacedo86: Also, see: https://github.com/kubernetes-sigs/kubebuilder/actions/runs/16007968823/job/45252541919?pr=4898. You need to rebase with the master branch.

The testdata ci job will re-generate all samples under testdata and docs.
It is failing because it is not rebased.

camilamacedo86: Indeed we need a CI test for that, so if that passes, we are OK with it, and we do not need to remember what exactly can impact it anymore.

Thedarkmatter10: Thanks for clarifying!

That makes total sense — it’s great that CI already handles the check for `PROJECT` file changes.  
To make it even more robust, I’d be happy to help write or enhance a dedicated CI test that ensures the scaffolded `PROJECT` file remains consistent.  

Let me know if that’s something you'd like me to take on  happy to help!

Thedarkmatter10: Just a quick heads-up — I’m currently working on a Windows environment, so I’m unable to run `make generate` locally.  
As a result, I can't regenerate the testdata files required to fix the failing `Testdata-verification` CI step:


Would you be able to help regenerate the testdata on your side to ensure everything stays consistent?

I really appreciate your time and support — and I’m happy to assist with anything else needed on the PR. 🙏

camilamacedo86: That would be great 
That is how we do the release: https://github.com/kubernetes-sigs/kubebuilder/blob/master/.github/workflows/release.yml

We could call in the CI a test that does the release ( not publish it )
And run the kubebuilder init with and ensure that the cliVersion is present and with the current / expected format

camilamacedo86:
Hi @Thedarkmatter10 

See that the testdata job will be fixed with: https://github.com/kubernetes-sigs/kubebuilder/pull/4905
Also, we must to have just one commit before merge out, could you please squash it?

By last since it will produce a change for end users it should end up in our release notes
The title of PRs is our release notes ,and so, we cannot use 🌱 

We need to use like ✨ and add a text that clarifies what was done such as : 
> Improve output of version command to... 

Could you please update those things? 

Thedarkmatter10:
> Hi @Thedarkmatter10
> 
> See that the testdata job will be fixed with: #4905 Also, we must to have just one commit before merge out, could you please squash it?
> 
> By last since it will produce a change for end users it should end up in our release notes The title of PRs is our release notes ,and so, we cannot use 🌱
> 
> We need to use like ✨ and add a text that clarifies what was done such as :
> 
> > Improve output of version command to...
> 
> Could you please update those things? Thank you a lot for your contribution 🥇

Hi @camilamacedo86 👋

I tried squashing the commits into one as suggested, but I ran into some issues with rebase and merge conflicts that I wasn't able to resolve cleanly. As a result, the PR still shows extra commits.

Apologies for the inconvenience caused 🙏 . I understand the importance of keeping the commit history clean and will make sure not to repeat this in future contributions.

Please let me know if you'd prefer that I close this PR and open a fresh one with a clean history, or if there's another way you'd like to proceed.
and also i renamed title of PR.

Thanks again for your guidance and patience! 😊

cmd/version.go
Thedarkmatter10: Thank you for the explanation and the reference to the release.yml workflow!

I’d be happy to help draft a CI enhancement that:

- Simulates the release by building the CLI (without publishing).

- Runs kubebuilder init with the built binary.

- Validates that the cliVersion appears in the scaffolded PROJECT file and follows the expected format.

Let me know if you'd like me to proceed — I can open a follow-up PR for that! 😊

camilamacedo86: (because of the scenario explained above and to ensure that it will not impact the alpha update command, which is WIP)

I will need to find some time to properly review these changes and let you know more about
So, it might take a while, but I will get back to you

Thedarkmatter10: Thanks a lot , i will wait for your confirmation and the next step 😀.

cmd/version.go
```diff
@@ -17,58 +17,91 @@ limitations under the License.
...
+Build Date:          %s
+Go OS/Arch:          %s/%s`,
+		v.KubeBuilderVersion,
+		v.KubernetesVendor,
+		v.GitCommit,
+		v.BuildDate,
+		v.GoOS,
+		v.GoArch,
+	)
+}
```
vitorfloriano, line 94: @camilamacedo86 @Thedarkmatter10 could we move the output formatting to `cli/version.go`? 

I think that separating the data gathering logic from the presentation logic could give us more flexibility to deal with output formatting, add flags, and leave room for other tools/commands to define how they want to display the version. The same goes for the JSON output. We could turn that into a flag in `cli/version.go`. 

This separation was discussed briefly in a [previous PR conversation](https://github.com/kubernetes-sigs/kubebuilder/pull/4860#discussion_r2135716772) (before I messed up and closed the PR).

WDYT?

cmd/version.go
camilamacedo86: See: https://github.com/kubernetes-sigs/kubebuilder/pull/4939/files
For we add the tests

Thedarkmatter10: Thanks for setting this up, Camila! 🙌
This CI workflow perfectly validates what I had in mind  ensuring that changes like the one I submitted don't unintentionally modify the cliVersion written to the PROJECT file.
Great to see this automated check in place now really appreciate you putting it together! 😊

camilamacedo86:
Hi @Thedarkmatter10 

This one is not passing all tests.
Can you please rebase with master and fix the lint issues?

Thedarkmatter10:
> Hi @Thedarkmatter10
> 
> This one is not passing all tests. Can you please rebase with master and fix the lint issues?

Hi @camilamacedo86 ,

I’ve completed the rebase successfully.

Regarding the multiple file changes that are showing up: these seem to be caused by running golangci-lint on my Windows machine, which has likely introduced differences in line endings (CRLF vs LF). This ends up marking a lot of files as changed even though there are no logical code modifications.

Since changing all these files would introduce noise in the PR, I’ve held off committing them. Would you be able to re-run golangci-lint on a Unix/Linux environment on your side to avoid the line-ending conflict? Or alternatively, let me know how you'd prefer to handle this  I’m happy to follow any suggestion.

Thanks!

camilamacedo86:
Hi @Thedarkmatter10 

Thank you for your effort in this one.
See the lint issue:

>   level=error msg="Running error: context loading failed: no go files to analyze: running `go mod tidy` may solve the problem"
  
  Could you run go mod tidy
  And make verify?

Also, this one should tirgger: https://github.com/kubernetes-sigs/kubebuilder/blob/master/.github/workflows/release-version-ci.yml . The path is wrong, we need to add cmd/ as well. Could you please update it to add the path?

By last we need to have all in the same commit, could you please squash?

Thedarkmatter10:
Hi @camilamacedo86,

I hope you're doing well.
I wanted to kindly check if there's anything else pending from my side to get this PR merged. Please let me know if any further changes are needed.

Thank you

camilamacedo86:
Hi @Thedarkmatter10,

Hope you're doing well!

Just wanted to kindly check if there’s anything else pending on my side to get this PR merged. Please let me know if any further changes are needed.

I saw you’ve resolved all CI issues and rebased—thank you! 🎉

Could you please squash the commits? We aim for 1 commit per PR = 1 problem = 1 solution. This helps keep our git history clean and avoids conflicts when reverting or cherry-picking changes.

Once that’s done, I’ll take a final look. Now that all tests are green, it should be quick!

cmd/version.go
```diff
@@ -17,58 +17,91 @@ limitations under the License.
...
 	"fmt"
+	"runtime"
 	"runtime/debug"
 )
 
 const unknown = "unknown"
 
-// var needs to be used instead of const as ldflags is used to fill this
-// information in the release process
+// These are filled via ldflags during build
```
camilamacedo86, line 28: Could we keep the previous comments?

cmd/version.go
```diff
@@ -17,58 +17,91 @@ limitations under the License.
...
 var (
 	kubeBuilderVersion      = unknown
 	kubernetesVendorVersion = "1.33.0"
 	goos                    = unknown
 	goarch                  = unknown
-	gitCommit               = "$Format:%H$" // sha1 from git, output of $(git rev-parse HEAD)
-
-	buildDate = "1970-01-01T00:00:00Z" // build date in ISO8601 format, output of $(date -u +'%Y-%m-%dT%H:%M:%SZ')
```
camilamacedo86, line 35: Could we keep the previous comments?

cmd/version.go
```diff
@@ -17,58 +17,91 @@ limitations under the License.
...
+	gitCommit               = "$Format:%H$"
+	buildDate               = "1970-01-01T00:00:00Z"
 )
 
-// version contains all the information related to the CLI version
-type version struct {
+// VersionInfo holds all CLI version-related information
+type VersionInfo struct {
```
camilamacedo86, line 39:
```suggestion
type version struct {
```

Can we avoid externalising this one?
If we externalise it (e.g., by using v instead of V), it becomes accessible outside of Kubebuilder and effectively becomes part of its public API.

We intentionally do not want to expose this structure for use by other projects. Let's keep it internal to maintain tighter control and avoid unintended usage.

camilamacedo86, line 39:
```suggestion
// version holds all CLI version-related information
type version struct {
```

Can we avoid externalising this one?
If we externalise it (e.g., by using v instead of V), it becomes accessible outside of Kubebuilder and effectively becomes part of its public API.

We intentionally do not want to expose this structure for use by other projects. Let's keep it internal to maintain tighter control and avoid unintended usage.

cmd/version_test.go
```diff
@@ -0,0 +1,65 @@
...
+package cmd
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestVersionStringIncludesExpectedFields(t *testing.T) {
```
camilamacedo86, line 26: 👍 Amazing 🚀

cmd/version.go
```diff
@@ -17,58 +17,91 @@ limitations under the License.
...
-		buildDate,
-		goos,
-		goarch,
-	})
+	if goos == unknown {
+		goos = runtime.GOOS
+	}
+	if goarch == unknown {
+		goarch = runtime.GOARCH
+	}
```
camilamacedo86, line 60: Thanks so much for the suggestion! 🙌

However, I think we should **not fallback to `runtime.GOOS` / `runtime.GOARCH`** here.

```go
if goos == unknown {
	goos = runtime.GOOS
}
if goarch == unknown {
	goarch = runtime.GOARCH
}
```

This will always reflect the **build-time platform** — not the actual target platform of the binary.  
For example, in our case, all builds run on Ubuntu (see our [GoReleaser config](https://github.com/kubernetes-sigs/kubebuilder/blob/master/build/.goreleaser.yml)), so this will return:

```
GOOS = "linux"
GOARCH = "amd64"
```

—even for macOS binaries like `darwin/amd64`. This results in incorrect output when users run:

```bash
kubebuilder version
```

We already inject the correct values using `-ldflags` at build time, and if something goes wrong with that injection, it's **better to leave the values as `"unknown"`** to surface the issue, rather than silently reporting incorrect metadata.

Could we please revert it?

cmd/version.go
camilamacedo86: I think that make sense we could do something like

```
if jsonOutput {
    printAsJSON(GetVersionInfo())
} else {
    printPretty(GetVersionInfo())
}
```
But I would accept it in a follow up PR as well
I think the goal of this one ( unless I misunderstood) would be add only  the printPretty part

camilamacedo86 (changes requested):
Really appreciate your contribution here 🥇  — and sorry for the delay in reviewing it!

This change definitely makes the output nicer (thank you for that!) 🚀, **but we have to be extra cautious because it touches parts that can affect things in more subtle ways** — especially around how the CLI is released and used as a Go module or by other libraries.

Even small changes here can impact consumers or break the release process in ways that aren’t easily caught by tests. **That’s why we need to be conservative when altering this part of the code.**

Thanks again for the effort — it’s super valuable, and we’d love to find a safe path forward if we can!
I've added a few comments — the main concerns are:

- These methods should **remain private**. They're internal to Kubebuilder, and exposing them would mean supporting them as part of our public API, which we don’t intend to do.

- The fallback logic:

  ```go
  if goos == unknown {
  	goos = runtime.GOOS
  }
  if goarch == unknown {
  	goarch = runtime.GOARCH
  }
  ```

  can introduce misleading output. Since our builds run on **Ubuntu**
  ([GoReleaser config](https://github.com/kubernetes-sigs/kubebuilder/blob/master/build/.goreleaser.yml)), this would always return:

  ```go
  linux/amd64
  ```

  —even for macOS binaries, which can confuse users and scripts relying on `kubebuilder version`.


Would you be open to adjusting the implementation to:
- Keep the fallback logic removed — rely on `ldflags` only
- Limit the scope of the PR to just the necessary improvements
- Keep internal methods unexported and avoid API surface changes

We really appreciate the effort and want to support your contribution!

cmd/version.go
camilamacedo86: Also, see : https://github.com/kubernetes-sigs/kubebuilder/pull/4516/files 
It can either be used if kubebuilder is installed with go install module
(another behaviour that we need to consider)

camilamacedo86:
Thank you for your contirbution 🥇 
I am closing this one based on the comment: https://github.com/kubernetes-sigs/kubebuilder/pull/4898#pullrequestreview-3082325538

However, if you see that it should be reopened, please feel free to do so.

PR #4901, opened 2025-07-02
🐛 fix(helm): propagate project-name into init helm

### LINKED ISSUE #4899: init for helm/v1-alpha is generating broken helm charts

According to your documentation, it should be possible to initialize a new project with just a helm chart ([xref: kubebuilder book](https://book.kubebuilder.io/plugins/available/helm-v1-alpha#basic-usage))

`kubebuilder init --plugins=helm/v1-alpha` itself runs fine, but the generated helm-chart is broken.
By simply running `helm lint dist/chart` against the generated chart directory,
the helm linter is complaining:

```
==> Linting dist/chart
[ERROR] Chart.yaml: name is required
[ERROR] templates/: validation: chart.metadata.name is required
[ERROR] : unable to load chart
        validation: chart.metadata.name is required

Error: 1 chart(s) linted, 1 chart(s) failed
```

This is, because [ProjectName](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/chart.go#L27-L31) in the `HelmChart` struct is an empty string.
But it's value is needed for templating charts.
https://github.com/search?q=repo%3Akubernetes-sigs%2Fkubebuilder+%22%7B%7B+.ProjectName+%7D%7D%22+path%3Apkg%2Fplugins%2Foptional%2Fhelm&type=code

*additional notes*:
* using `kubebuilder edit --plugins=helm/v1-alpha`  works fine if `kubebuilder init` was called before (as with that, the `PROJECT` file contains a value for `projectName`.
* as there are no e2e-tests for `kubebuilder init --plugins=helm/v1-alpha` I'm not sure if it's even implemented to work that way

### Reproducing this issue

1. get the latest kubebuilder binary
2. create a new test folder
3. run `kubebuilder init --plugins=helm/v1-alpha`
4. either look at the `name` field in generated `dist/chart/Chart.yaml` or run `helm lint dist` (if `helm` binary is installed on your system)

### KubeBuilder (CLI) Version

4.6.0

### PROJECT version

3

### Plugin versions

```yaml
layout:
- helm.kubebuilder.io/v1-alpha
```

### Other versions

_No response_

### Extra Labels

_No response_

bavarianbidi:
Without a project-name, the generated helm chart was invalid.
Therefor it's required to have the `project-name` as flag in the `init` subcommand.

If `project-name` is not defined, the project-name will get generated by the name of the current working directory.
This is already done in the `kustomization init`. Therefore I've extended the config-Interface by `GenerateProjectName` and implemented a generic func which is then used in `kustomization init` as well as in `helm init`.

Added tests for the util-package as I wanted to have a faster feedback loop during implementation and e2e tests for `--plugins=helm/v1-alpha` case.

closes #4899


pkg/config/interface.go
```diff
@@ -51,6 +51,8 @@ type Config interface {
 	// SetProjectName sets the project name.
 	// This method was introduced in project version 3.
 	SetProjectName(name string) error
+	// GenerateProjectName generates a project name.
+	GenerateProjectName(name string) error
```
camilamacedo86, line 55: Thanks for the contribution!

We don't need this change because the **project name** is already tracked in the config, and there's no need to add a new method to the interface for it.

Just to clarify:
- The project name is already being set via `SetProjectName(name string) error`
- And it's tracked in the config file. For example:
  https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/PROJECT#L9

```
layout:
- go.kubebuilder.io/v4
projectName: project-v4 <- HERE WE HAVE THE PROJECT name
```

Hope that makes sense! 🙏

test/e2e/helm/generate_test.go
```diff
@@ -0,0 +1,87 @@
...
+	})
+})
+
+// GenerateProject implements a helm/v1(-alpha) plugin project defined by a TestContext.
+func GenerateProject(kbc *utils.TestContext) {
+	var err error
+
+	By("initializing a project")
+	err = kbc.Init(
+		"--plugins", "helm.kubebuilder.io/v1-alpha",
```
camilamacedo86, line 58: Thanks for the effort here!

Just a heads-up — this isn't quite accurate.  
To properly test this, we need to **mock a full project setup**, which means running the following steps:

- `kubebuilder init`
- `kubebuilder create api`
- `kubebuilder create webhook`
- Uncomment kustomize config if needed
- **Then** call the Helm plugin (e.g., `kubebuilder edit --plugins helm`)

This sequence is necessary to simulate a real-world use case.

✅ Here's an example of the full project setup being mocked:
https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v4/generate_test.go#L32-L68

📦 And here’s how we properly test Helm support **after** that setup:
https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v4/plugin_cluster_test.go#L158-L181

Let me know if you'd like help aligning this test with the above pattern 🙌

test/e2e/helm/generate_test.go
```diff
@@ -0,0 +1,87 @@
...
+	)
+	Expect(err).NotTo(HaveOccurred(), "Failed to initialize project")
```
camilamacedo86, line 60: Thanks for raising this — I think I now understand your point about whether we should support `helm` in the `init` interface or only in `edit`. 

**We definitely need to review this.** 

Technically, you *can* run `kubebuilder init --plugins=go/v4,helm/v1-alpha1`, which includes Helm during project initialization — **but the Helm-specific content wouldn't be fully or properly populated at that stage**. 

That’s because Helm relies on the Go project being scaffolded first. 

Originally, the intention was to allow us to automatically update Helm-related files using `edit`, since the default layout is always called in the plugin chain.

 For example, if we init a project with plugins A, B, and C — when we later call `create api`, all those plugins will be executed in the chain automatically. 

**So with that in mind, it seems clearer now that having a `helm init` doesn't really make sense.** 

I think the right path is to remove the `init` implementation for Helm and keep only `edit`. Thanks again for bringing this up — I now understand the confusion you raised earlier on Slack. Would you mind opening an issue for this? And would you be interested in working on it? 🙂

pkg/config/interface.go
bavarianbidi: yes, but if you do a `kubebuilder init --plugins=helm/v1-alpha` in an empty directory there is no project name and a project-name isn't generated. And that's why the generated helm-charts are wrong.

My thought was not duplicating code we already have in `kustomize init` (https://github.com/kubernetes-sigs/kubebuilder/pull/4901/files#diff-c81dda43edfad86b44d0309ad6026ae67301b5e2e6da69cf6f3cd59d3c7b89d4L72-L85) and do the same in `helm init` as well.

but sure, i can revert that change and copy the same steps from `kustomize init` to `helm init` (beside the domain-stuff: https://github.com/kubernetes-sigs/kubebuilder/pull/4901/files#diff-c81dda43edfad86b44d0309ad6026ae67301b5e2e6da69cf6f3cd59d3c7b89d4L67-L69)

camilamacedo86 (changes requested):
Thank you a lot 🎉 
That is amazing !!!

I added some comments please feel free to reach out if you need any further clarification.

test/e2e/helm/generate_test.go
bavarianbidi: > I think the right path is to remove the init implementation for Helm and keep only edit. Thanks again for bringing this up — I now understand the confusion you raised earlier on Slack. Would you mind opening an issue for this? And would you be interested in working on it? 🙂

Thanks for that. That was exactly what i wanted to know :sweat_smile: 
I'm definitely open to implement this.

Will raise a new issue and with that we can close my previous issue #4899 afterwards

pkg/config/interface.go
bavarianbidi: just saw your comment https://github.com/kubernetes-sigs/kubebuilder/pull/4901#discussion_r2179570929

test/e2e/helm/generate_test.go
bavarianbidi: removal-request for `helm init` created #4902

bavarianbidi:
closing this on in favor of #4902 and PR #4903

Issue #4833, opened 2025-05-26
Improve Helm plugin by generating chart from install.yaml (via Kustomize build)


camilamacedo86:
### What do you want to happen?

**Description:**

Currently, Kubebuilder provides support for generating Helm charts to help users package and deploy their operators. This is a great feature and brings real value. However, the current implementation has limitations, especially when changes are made in the config/ directory using Kustomize. 

Reference to current implementation:
https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/optional/helm/v1alpha

While the current approach proves that Helm support is possible and useful, it allows us to start discussing and getting feedback from the community about how the chart should be; as add tests to validate to cover its functionalities it is not the easiest to maintain or the best in the long term, especially as the project evolves. Improvements can make this much simpler and more robust.

**Proposal:**

Update the Helm plugin to:

- Run the same logic that generates the install.yaml from Kustomize (like we already do for make build-installer) bu on the fly:

```sh
.PHONY: build-installer  
build-installer: manifests generate kustomize ## Generate a consolidated YAML with CRDs and deployment.  
	mkdir -p dist  
	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}  
	$(KUSTOMIZE) build config/default > dist/install.yaml  
```

- Use the resulting install.yaml as input to generate the Helm chart
- We should still building the chart as it is today

This would make the Helm chart:
- Reflect exactly what’s in the kustomize result from config/
- Allow more customizations via Kustomize and flexibilities 
- Reduce manual Helm template maintenance ( if something changes for Golang/Kustomize we need to update the HelmChart templates, if the change impact the for limitations not coming from the templates)
- Solve existing feature requests like: https://github.com/kubernetes-sigs/kubebuilder/issues/4625

**Alternative Proposal** 

Keep the code as it is, but try to replace the templates with logic to generate the files from config/ instead as much as possible.

nerdeveloper:
@camilamacedo86, what is remaining here, please? If you would like to drop some information, I can pick this up.

camilamacedo86:
Fixed with [#5058](https://github.com/kubernetes-sigs/kubebuilder/pull/5058)
Closing it

PR #4903, opened 2025-07-02
⚠ (helm/v1-alpha): remove `init` command from Helm plugin as it's only meaningful with `edit`

### LINKED ISSUE #4902: remove kubebuilder init --plugin=helm support

Feature is more a deprecation "wish" :-)

During the implementation of #4891 there was some confusion about `kubebuilder init --plugin=helm`.
After several discussions, Camila summarized the issue with the `init` implementation of the helm-plugin [in this PR](https://github.com/kubernetes-sigs/kubebuilder/pull/4901#discussion_r2179570929) quite good:

> Thanks for raising this — I think I now understand your point about whether we should support `helm` in the `init` interface or only in `edit`.
>
> *We definitely need to review this.*
>
> Technically, you can run `kubebuilder init --plugins=go/v4,helm/v1-alpha1`, which includes Helm during project initialization — but the Helm-specific content wouldn't be fully or properly populated at that stage.
>
> That’s because Helm relies on the Go project being scaffolded first.
>
>Originally, the intention was to allow us to automatically update Helm-related files using edit, since the default layout is always called in the plugin chain.
>
> For example, if we init a project with plugins A, B, and C — when we later call create api, all those plugins will be executed in the chain automatically.
>
> So with that in mind, it seems clearer now that having a `helm init` doesn't really make sense.
>
> I think the right path is to remove the `init` implementation for Helm and keep only edit. Thanks again for bringing this up — I now understand the confusion you raised earlier on Slack. Would you mind opening an issue for this? And would you be interested in working on it? 🙂

I'm volunteering to implement this. As this makes #4891 more easily

### Extra Labels

_No response_

bavarianbidi:
The helm-plugin is meant to extend an existing go project. For a correct rendering of the helm-chart some information (like the project name) from the go project is needed.

closes #4902


pkg/plugins/optional/helm/v1alpha/plugin.go
```diff
@@ -54,9 +52,6 @@ func (Plugin) Version() plugin.Version { return pluginVersion }
 // SupportedProjectVersions returns an array with all project versions supported by the plugin
 func (Plugin) SupportedProjectVersions() []config.Version { return supportedProjectVersions }
 
-// GetInitSubcommand will return the subcommand which is responsible for initializing and scaffolding helm manifests
-func (p Plugin) GetInitSubcommand() plugin.InitSubcommand { return &p.initSubcommand }
-
```
camilamacedo86, line 59: Since we have only edit we might want to rename: https://github.com/kubernetes-sigs/kubebuilder/blob/4e369844ee07ab4fff8b3d34d6cedd01d4ccc0af/pkg/plugins/optional/helm/v1alpha/scaffolds/init.go

Too .. we no longer need init,go 
Just edit :-)

camilamacedo86:
@bavarianbidi we can overwrite the APIDiff that is fine no worry
But we need to pass in the other tests
The link you can check locally with make lint and make lint-fix

Thank you a lot for looking on that 🎉

test/e2e/helm/e2e_suite_test.go
```diff
@@ -0,0 +1,29 @@
+/*
```
camilamacedo86: Could we have a PR only for the e2e tests?
Ideally we should have one PR for each purpose 
AND 
with only one commit

Could you please do that for we are able to move forward?

testdata/project-v4-multigroup/go.mod
```diff
@@ -3,7 +3,7 @@ module sigs.k8s.io/kubebuilder/testdata/project-v4-multigroup
 go 1.24.0
 
 require (
-	github.com/cert-manager/cert-manager v1.18.1
+	github.com/cert-manager/cert-manager v1.18.2
```
camilamacedo86: Can you please rebase with master?

test/e2e/helm/generate_test.go
```diff
@@ -0,0 +1,166 @@
...
+			fileContainsExpr, err := pluginutil.HasFileContentWith(
+				filepath.Join(kbc.Dir, "dist", "chart", "values.yaml"),
+				`webhook:
+  enable: true`)
+			Expect(err).NotTo(HaveOccurred(), "Failed to read values.yaml file")
+			Expect(fileContainsExpr).To(BeTrue(), "Failed to get enabled webhook value from values.yaml file")
+		})
+
+		// Without --force, the webhooks should not be enabled in the values.yaml file.
+		It("should extend a runnable project with helm plugin but not running with --force", func() {
```
camilamacedo86: I think this one is kind of a duplicate, right?
We want to cover all scenarios as much as possible, but we need to be careful, for we do not want to create unnecessary tests.

test/e2e/helm/generate_test.go
```diff
@@ -0,0 +1,166 @@
...
+		})
+
+		AfterEach(func() {
+			kbc.Destroy()
+		})
+
+		It("should extend a runnable project with helm plugin", func() {
+			initTheProject(kbc)
+			generateProject(kbc)
+		})
```
camilamacedo86: Could we not just have one test for now? 
And then we think more about what we actually need and how to cover

test/e2e/helm/generate_test.go
bavarianbidi: yes and no :-)
The only difference is that we are running `kubebuilder edit --plugins helm.kubebuilder.io/v1-alpha` once with `--force` and once without `--force`.

Without `--force` the `values.yaml` doesn't contain a webhook template folder and a `webhook` object in the helm `values.yaml`.

As this is an explicit decision within the code I have thought about explicitly testing that case as well to get at least a failure once that behavior will change.

test/e2e/helm/generate_test.go
bavarianbidi: >  And then we think more about what we actually need and how to cover

I currently see these steps where a helm-plugin could be integrated. 
And with that it really depends if `--force` is needed or not:

1. extending a _basic project_ (xref: [kubebuilder book basic project](https://book.kubebuilder.io/cronjob-tutorial/basic-project))
   we expect a basic helm-chart without a `webhook` directory in the helm-chart (as no webhooks got defined)
3. extending a _basic project_ by an admission webhook (xref: [kubebuilder book webhook implementation](https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation))
   we expect the same helm-chart content from `1.` plus all the things to have `(Mutating|Validating)WebhookConfiguration` work (from a helm-chart perspective)
5. extending a _basic project_ by generating helm-charts (via `kubebuilder edit --plugins helm.kubebuilder.io/v1-alpha`) and after that extending the project by an admission webhook
   and in that case it differs if  `kubebuilder edit --plugins helm.kubebuilder.io/v1-alpha` got called with/without `--force` as we then get/not get all the things to have `(Mutating|Validating)WebhookConfiguration` work (from a helm-chart perspective)

test/e2e/helm/e2e_suite_test.go
bavarianbidi: regarding a separate PR for e2e-tests:
let us please came up with an idea on https://github.com/kubernetes-sigs/kubebuilder/pull/4903#discussion_r2185687198 and https://github.com/kubernetes-sigs/kubebuilder/pull/4903#discussion_r2185686431 first :pray: 

regarding squashing: let's decided on :point_up: and than I can squash all the commits which belong to this PR.

test/e2e/helm/generate_test.go
bavarianbidi: I think it also might be helpful to have this test-cases covered once someone starts working on:
https://github.com/kubernetes-sigs/kubebuilder/issues/4833

test/e2e/helm/e2e_suite_test.go
camilamacedo86: If we have a PR to remove the init then we can get it merged asap
If we have a PR with tests added, we just need to discuss what we are testing and how, then we can also get it merged asap, once we reach a consensus

The request here is simply for us to split so that we have a PR for each purpose, allowing us to provide better Git history and release notes. 

Regarding the commit squash, yes, we have this golden rule: if we can, it would be amazing!

bavarianbidi: just raised PR #4914 for testing the existing behavior of `kubebuilder edit --plugins helm.kubebuilder.io/v1-alpha`

test/e2e/helm/generate_test.go
bavarianbidi: move this discussion into PR #4914

test/e2e/helm/e2e_suite_test.go
bavarianbidi: rebased and pushed again - PTAL :pray:

camilamacedo86 (approved):
/lgtm

PR #4907, opened 2025-07-05
WIP: (go/v4): Ensure that default API scaffold follow k8s api conventions

camilamacedo86:


docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_types.go
```diff
@@ -169,11 +169,19 @@ type CronJobStatus struct {
...
-	metav1.TypeMeta   `json:",inline"`
+	metav1.TypeMeta `json:",inline"`
+
+	// metadata is a standard object metadata
+	// +optional
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   CronJobSpec   `json:"spec,omitempty"`
-	Status CronJobStatus `json:"status,omitempty"`
+	// spec defines the desired state of CronJob
```
mateusoliveira43, line 178: to avoid confusion, I would remove comment in Spec/Status structs (in this case, line 65 and line 145) since they will be overridden by these new comments, in all `api` files

camilamacedo86: We must add the comment so that it complies with the rules.
The spec fields should have a doc which will end up in the CRD definitions.

By doing so, all our samples under testdata and docs (the tutorials and projects) will also have those. 
Sorry, I am not sure if I understand what you mean by being overwritten. Can you please explain?

pkg/plugins/golang/v4/scaffolds/internal/templates/api/types.go
```diff
@@ -73,13 +73,16 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
camilamacedo86, line 73: Important: The real change here is just this file: pkg/plugins/golang/v4/scaffolds/internal/templates/api/types.go . All the rest are just sample and tutorial updates, as Hack generates docs; it is still able to inject the code for doc samples after the changes.

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_types.go
mateusoliveira43: for example: on line 65 in this file we have
```go
// CronJobSpec defines the desired state of CronJob
type CronJobSpec struct {
```

prior to this change, that was what populated the CRD description, now it uses 
```go
// spec defines the desired state of CronJob
```

(but maybe removing go doc from line 65, can make linter complain that exported type has no comment)

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_types.go
```diff
@@ -165,20 +165,28 @@ type CronJobStatus struct {
...
+// CronJob is the Schema for the cronjobs API
 type CronJob struct {
 	/*
 	 */
-	metav1.TypeMeta   `json:",inline"`
+	metav1.TypeMeta `json:",inline"`
+
+	// metadata is a standard object metadata
+	// +optional
 	metav1.ObjectMeta `json:"metadata,omitempty"`
```
JoelSpeed:
```suggestion
	metav1.ObjectMeta `json:"metadata,omitempty,omitzero"`
```

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_types.go
```diff
@@ -165,20 +165,28 @@ type CronJobStatus struct {
...
 
-	Spec   CronJobSpec   `json:"spec,omitempty"`
-	Status CronJobStatus `json:"status,omitempty"`
+	// spec defines the desired state of CronJob
+	// +required
+	Spec CronJobSpec `json:"spec"`
+
+	// status defines the observed state of CronJob
+	// +optional
+	Status *CronJobStatus `json:"status,omitempty"`
```
JoelSpeed: Would also work
```suggestion
	Status CronJobStatus `json:"status,omitempty,omitzero"`
```

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_types.go
camilamacedo86: I think it is a better option than a pointer

docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_types.go
camilamacedo86: The type still with the same desc ( without a dot at the end only )
Then, we added desc for 

```

	Spec   CronJobSpec   `json:"spec,omitempty"`
	Status CronJobStatus `json:"status,omitempty"`
```

That was without any before

JoelSpeed (commented):
Changes to serialization LGTM

PR #4909, opened 2025-07-05
🐛 (deploy-image/v1-alpha1): Remove default max size validation

camilamacedo86:
This removes the default max validation for the size field in the deploy-image plugin.

Since the plugin allows deploying any image, setting a default maximum doesn’t seem appropriate and could be misleading.
We believe it’s better to leave this responsibility to the user.

This change aligns with flexibility and avoids setting unnecessary constraints by default.


camilamacedo86:
@erikgb 

Could you help me out with this one?
WDYT? Does it make sense?

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/spf13/afero v1.14.0
 	github.com/spf13/cobra v1.9.1
 	github.com/spf13/pflag v1.0.6
+	golang.org/x/mod v0.25.0
```
erikgb: Not sure why this is changing in this PR. Was the `go.mod` file not tidy on master?

erikgb (commented):
/lgtm
/hold
In case you want to change/investigate the unexpected change in `go.mod` file.

go.mod
camilamacedo86: Go changes, misisng update done in another PR

PR #4911, opened 2025-07-07
🌱 [hack-generate docs]: Update errors comments to make easier find what operation fail

camilamacedo86:
We should have unique error messages to make troubleshooting easier.


camilamacedo86:
Just a nit to make it easier to work with, as it has no reason to make anybody waste time on reviews.
Moving forward.

PR #4914, opened 2025-07-09
🌱 helm(e2e): test helm plugin integration

bavarianbidi:
this will test the usage and impact to an existing project by the kubebuilder edit command for the helm plugin.

This is a follow up from #4903


pkg/plugin/util/util_test.go
```diff
@@ -101,4 +101,60 @@ var _ = Describe("Cover plugin util helpers", func() {
...
+- helm.kubebuilder.io/v1-alpha
+plugins:
+  helm.kubebuilder.io/v1-alpha: {}
+repo: github.com/example/repo
+version: "3"
+`
+		)
+
+		BeforeAll(func() {
+			err := os.MkdirAll("testdata", 0o755)
```
camilamacedo86, line 124: We should not use the testdata dir at all.
We should do as in the other tests:

https://github.com/kubernetes-sigs/kubebuilder/blob/2dc541943ee5285b2bc4176338d8d64ff4e2f83c/test/e2e/deployimage/plugin_cluster_test.go#L38-L44

By using the TestContext we create a tmp dir for the tests and we ensure that we are using the kubebuilder binary build with the PR / source code changes

test/e2e/helm/generate_test.go
```diff
@@ -0,0 +1,191 @@
...
+import (
+	"path/filepath"
+
+	pluginutil "sigs.k8s.io/kubebuilder/v4/pkg/plugin/util"
+
+	. "github.com/onsi/ginkgo/v2"
+
+	. "github.com/onsi/gomega"
+
+	"sigs.k8s.io/kubebuilder/v4/test/e2e/utils"
```
camilamacedo86: we need to fix the order of the imports

pkg/plugin/util/util_test.go
```diff
@@ -101,4 +101,60 @@ var _ = Describe("Cover plugin util helpers", func() {
...
+		})
+
+		It("should return false when file does not contain the expected content", func() {
+			content := "nonExistentContent"
+			found, err := HasFileContentWith(path, content)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(found).To(BeFalse())
+		})
+	})
 })
```
camilamacedo86, line 160: Also, we should not add ANY change in the plugin/util/util_test
We do not need to create any new func here.

test/e2e/helm/generate_test.go
```diff
@@ -0,0 +1,191 @@
...
+		It("should extend a runnable project with helm plugin", func() {
+			initTheProject(kbc)
+
+			By("extend the project by adding helm plugin")
+			err := kbc.Edit(
+				"--plugins", "helm.kubebuilder.io/v1-alpha",
+			)
+			Expect(err).NotTo(HaveOccurred(), "Failed to edit the project")
+
+			ensureCommonHelmFilesContent(kbc)
```
camilamacedo86: See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/alphagenerate/generate_test.go#L234-L257

We should load the Project file and check the data as above
We do not need to create any new function in the utils plugin or test for that. 

We will validate after parsing the PROJECT file to the config ( marshall ) if we have the helm plugin

pkg/plugin/util/util_test.go
camilamacedo86, line 124: Why are we using test data here?
We should not touch there in the unit tests

test/e2e/helm/generate_test.go
```diff
@@ -0,0 +1,191 @@
...
+		})
+
+		// This test is to ensure that the helm plugin can be added to a project
+		// that has already been initialized with the go/v4 plugin.
+		// As the project is getting extended with webhooks,
+		// it is needed to run the `kubebuilder edit --plugins helm.kubebuilder.io/v1-alpha` command
+		// with ` --force` again to ensure that the webhooks are enabled in the
+		// values.yaml file.
+		It("should extend a runnable project with helm plugin and webhooks", func() {
```
camilamacedo86: Are we validating that the project is runnable?
No right, so could we try to improve the text to be more in line with what we are checking out?

pkg/plugin/util/util_test.go
bavarianbidi: it's not the `testdata` folder from the repo-root.
It's a temporary `testdata` folder which get's generated during tests.

I've just re-used this folder-name as we already create that folder in the existing tests as well:
https://github.com/kubernetes-sigs/kubebuilder/pull/4914/files/df9368ad4c692d8af48f866066895eacf2de7232#diff-f8d5c1dc31041937679a6530c7e9779770025f88333231013b03bf0018cbb98cR35

test/e2e/helm/generate_test.go
bavarianbidi: Thanks for this.
But I will create an issue for that as well, because
the `goimports` formater (in `golangci-lint` doesn't complain) about these.

Maybe `gci` (https://golangci-lint.run/usage/formatters/#gci) might fit better than `goimports` - but let's discuss this in the issue.

test/e2e/helm/generate_test.go
bavarianbidi: > We should load the Project file and check the data as above

Will use some of the code where we marshal the config file. 

For all the helm-template files I am checking for: I've decided to use the existing `pluginutil.HasFileContentWith` function. Didn't wanted to introduce a new dependency here but we could try to marshal the helm-files - it definitely feels better than doing this file-comparision-foo :sweat_smile: 

>  We do not need to create any new function in the utils plugin or test for that.

I think this might be also a dedicated issue.
What I currently see: there are a bunch of funcs which are quite generic (or at least need some small adjustments) to make them re-usable in the end2end-tests.

test/e2e/helm/generate_test.go
bavarianbidi: got your finding.
I've used the term `runnable` as we use this term in the kubebuilder-book as well as in the existing e2e-tests few times, e.g.:

> [...] Note that most of this tutorial is generated from literate Go files that
form a runnable project, and live in the book source directory

But to be true, we mostly `run` the generated projects we are generating in the e2e-tests beside one project in the `grafana-plugin`. Unfortunately I've used the grafana-plugin e2e test as a basic for my e2e-tests :see_no_evil: 

will re-phrase the description and the comments

camilamacedo86: We might need to improve it there as well -)

test/e2e/helm/generate_test.go
bavarianbidi: raised a new PR to enforce a formatted import-section: https://github.com/kubernetes-sigs/kubebuilder/pull/4927

camilamacedo86:
Hi @bavarianbidi 

I think we will need to improve those a little.
But we can either do that in follow ups
So, it might be great enough just to get started.

/lgtm
/approve

bavarianbidi:
> Hi @bavarianbidi 
> 
> I think we will need to improve those a little.
> But we can either do that in follow ups
> So, it might be great enough just to get started.
> 
> Thank you a lot for your contribution 🥇 
> 
> /lgtm
> /approve

hi @camilamacedo86 , yes in the e2e code base is definitely room for improvement.

I'm fine with taking that with me.
Will create an issue first to document my findings and you can extend with your ideas. WDYT?

PR #4916, opened 2025-07-09
📖 Update Cronjob Tutorials API spec descriptions to better adopt  k8s API convetions

camilamacedo86:
Motivated by: https://github.com/kubernetes-sigs/kubebuilder/issues/4809

Note: A partial change is required. Here we are just fixing the description of the specs so that we reduce the number of changes to be discussed and we can properly debate and find out what markers we should use for each spec

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_types.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_types.go
@@ -64,19 +64,18 @@ import (
 
 // CronJobSpec defines the desired state of CronJob
 type CronJobSpec struct {
+	// schedule in Cron format, see https://en.wikipedia.org/wiki/Cron.
 	// +kubebuilder:validation:MinLength=0
-
-	// The schedule in Cron format, see https://en.wikipedia.org/wiki/Cron.
+	// +required
 	Schedule string `json:"schedule"`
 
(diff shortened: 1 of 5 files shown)


docs/book/src/cronjob-tutorial/testdata/project/api/v1/cronjob_types.go
```diff
@@ -64,46 +64,43 @@ import (
...
 
-	// This flag tells the controller to suspend subsequent executions, it does
+	// suspend tells the controller to suspend subsequent executions, it does
 	// not apply to already started executions.  Defaults to false.
 	// +optional
 	Suspend *bool `json:"suspend,omitempty"`
 
-	// Specifies the job that will be created when executing a CronJob.
+	// jobTemplate defines the job that will be created when executing a CronJob.
 	JobTemplate batchv1.JobTemplateSpec `json:"jobTemplate"`
```
afzal442, line 92:
```suggestion
         // +required
	JobTemplate batchv1.JobTemplateSpec `json:"jobTemplate"`
```
I think we can update this.

camilamacedo86: My idea here is just change the docs so in we can discuss those in a follow up ( the markers )

afzal442 (approved):

PR #4921, opened 2025-07-11
🌱 (chore): optimize CI time and usage of resources

camilamacedo86:
- Move the tests that do not require running on the cluster to run in the job pull-kubebuilder-test just once
- Change TestContext to work well outside of the cluster (it tries to capture the Kubernetes version so we have it tracked in case we want to skip a test for a specific version)
- Update e2e tests to move grafana and alphagenerate to run on this job, since they do not require a cluster
- Improve e2e tests for alpha generate by reducing the number of mocks — we don’t need one mock per plugin; we can use a single mock with all plugins scaffolded to validate properly. This should significantly reduce the time needed to perform the tests

We can check that the tests were executed here: https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kubebuilder/4921/pull-kubebuilder-test/1943729959755321344/build-log.txt

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,6 +156,7 @@ test-coverage: ## Run unit tests creating the output to report coverage
 .PHONY: test-integration
 test-integration: ## Run the integration tests
 	./test/integration.sh
+	./test/features.sh
 
 .PHONY: check-testdata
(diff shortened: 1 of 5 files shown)


vitorfloriano:
LGTM

PR #4927, opened 2025-07-15
🌱: introduce gci as formatter in golangci-lint

bavarianbidi:
To have a common appearance of the  import section, `gci` as formatter is enforcing the following pattern of imports:

```yaml
    gci:
      sections:
        - standard
        - default
        - prefix(sigs.k8s.io/kubebuilder)
        - dot
```

xref: `gci` documentation - https://golangci-lint.run/usage/formatters/#gci

This is a follow-up task by a PR comment in https://github.com/kubernetes-sigs/kubebuilder/pull/4914

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -83,16 +83,23 @@ linters:
         path: hack/docs/*
       - linters:
           - revive
-        text: 'should have comment or be unexported'
+        text: "should have comment or be unexported"
     paths:
       - third_party$
       - builtin$
       - examples$
 formatters:
   enable:
+    - gci
     - gofmt
     - gofumpt
     - goimports
+  settings:
+    gci:
+      sections:
+        - standard
+        - default
+        - prefix(sigs.k8s.io/kubebuilder)
   exclusions:
     generated: lax
     paths:
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/afero"
+
 	"sigs.k8s.io/kubebuilder/v4/pkg/cli"
 	cfgv3 "sigs.k8s.io/kubebuilder/v4/pkg/config/v3"
 	"sigs.k8s.io/kubebuilder/v4/pkg/machinery"
--- a/hack/docs/generate_samples.go
+++ b/hack/docs/generate_samples.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	log "github.com/sirupsen/logrus"
+
 	cronjob "sigs.k8s.io/kubebuilder/v4/hack/docs/internal/cronjob-tutorial"
 	gettingstarted "sigs.k8s.io/kubebuilder/v4/hack/docs/internal/getting-started"
 	multiversion "sigs.k8s.io/kubebuilder/v4/hack/docs/internal/multiversion-tutorial"
--- a/hack/docs/internal/cronjob-tutorial/generate_cronjob.go
+++ b/hack/docs/internal/cronjob-tutorial/generate_cronjob.go
@@ -23,6 +23,7 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/afero"
+
 	hackutils "sigs.k8s.io/kubebuilder/v4/hack/docs/utils"
 	pluginutil "sigs.k8s.io/kubebuilder/v4/pkg/plugin/util"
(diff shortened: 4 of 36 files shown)


mateusoliveira43 (approved):
should we also add to scaffolded config?

bavarianbidi:
> should we also add to scaffolded config?

I like the idea, but as we already use different formatter/linter-config, I wasn't sure about when to extend the skaffolding files:

## formatters in the project scaffolding ([xref: skaffolding golangci.yaml](https://github.com/mercedes-benz/kubebuilder/blob/f5ea91f16474dc1c01899947085c10cc85a1054d/pkg/plugins/golang/v4/scaffolds/internal/templates/golangci.go#L86-L95))

```yaml
formatters:
  enable:
    - gofmt
    - goimports
```

## formatters in the kubebuilder golangci.yaml config [xref: kubebuilder golangci yaml](https://github.com/mercedes-benz/kubebuilder/blob/f5ea91f16474dc1c01899947085c10cc85a1054d/.golangci.yml#L91-L96)

```yaml
formatters:
  enable:
    - gci
    - gofmt
    - gofumpt
    - goimports
```

.golangci.yml
```diff
@@ -90,9 +90,17 @@ linters:
...
     - gofmt
     - gofumpt
     - goimports
+  settings:
+    gci:
+      sections:
+        - standard
+        - default
+        - prefix(sigs.k8s.io/kubebuilder)
+        - dot
```
camilamacedo86: Do we need the settings?
Could not we just enable gci?

camilamacedo86:
>I like the idea, but as we already use different formatter/linter-config, I wasn't sure about when to extend the skaffolding files:

It's a good idea to see if there are things we can improve upon.
I would love get your help with
But we need one PR for each new one added there since we will need to change the scaffold

.golangci.yml
bavarianbidi: if we want to have the import section structured that way, than we have to explicitly configure it.

from the docs:
```yaml
  settings:
    gci:
      # Section configuration to compare against.
      # Section names are case-insensitive and may contain parameters in ().
      # The default order of sections is `standard > default > custom > blank > dot > alias > localmodule`.
      # If `custom-order` is `true`, it follows the order of `sections` option.
      # Default: ["standard", "default"]
      sections:
        - standard # Standard section: captures all standard packages.
        - default # Default section: contains all imports that could not be matched to another section type.
        - prefix(github.com/org/project) # Custom section: groups all imports with the specified Prefix.
        - blank # Blank section: contains all blank imports. This section is not present unless explicitly enabled.
        - dot # Dot section: contains all dot imports. This section is not present unless explicitly enabled.
        - alias # Alias section: contains all alias imports. This section is not present unless explicitly enabled.
        - localmodule # Local module section: contains all local packages. This section is not present unless explicitly enabled.
```

https://golangci-lint.run/usage/formatters/#gci

camilamacedo86: I am think if we need all seetings
Because how much caveat we add more complex it get
So, we might be able to remove the dot.
If we do not add anything we have already:  # Default: ["standard", "default"]
So, maybe just have  prefix(github.com/org/project)  would bring the same result, less grouping the dot ones?

camilamacedo86: What are your thoughts?
WDYT?

bavarianbidi: I would like to keep the dot. We only use dot imports in our tests for `gomega` and `ginkgo`.

And as it's some kind of a common pattern i would like to enforce via the formatter.

camilamacedo86: I noticed that we're using `or gomega and ginkgo` in the scaffolds as well.  
I don't think it's valuable to place the burden of changing that wording on the end users in this context.

**I’d prefer to keep things consistent across the board.**
Consistent is the key always ;-) 

Also, it seems to be organised only with `gomega and ginkgo`
But if we have many imports with it. Shouldn't they be in the respective group alignment / sorted out?

bavarianbidi: I've done a quick research over other well known repositories from the kubernetes ecosystem (k/k, controller-runtime, ...).

There is no consistency on the imports.
Mostly all of them are having dotted imports for `ginkgo` and `gomega` only. But they do not enforce it in a separate section.

I personally do not like dotted imports (even for `ginko` and `gomega`), but as they're often done I like it to have it in a separate section to have it more "eye-catching" 

But in the end, it's just my gut feeling and it's not important to me.

Will drop the dotted section and push again.

bavarianbidi: removed the dot section configuration

camilamacedo86 (approved):
/lgtm

Thank you 🥇

PR #4884, opened 2025-06-25
(book)feat: add guide for distributing Kubebuilder project

rayowang:
#4873

PR #4929, opened 2025-07-16
📖: Revert   guide for building a bundle with kubebuilder-based projects for management by OLM and/or publication on OperatorHub.io

camilamacedo86:
Reverts kubernetes-sigs/kubebuilder#4884

### DIFF
--- a/docs/book/src/faq.md
+++ b/docs/book/src/faq.md
@@ -158,36 +158,6 @@ type StructName struct {
 - Users still receive error notifications from the Kubernetes API for invalid `timeField` values.
 - Developers can directly use the parsed TimeField in their code without additional parsing, reducing errors and improving efficiency.
 
-## How to build a bundle with Kubebuilder-based projects to be managed by OLM and/or published in OperatorHub.io?
-
-You’ll need to create an [OLM bundle][olm-bundle-docs] to publish your operator.
-You can use the [Operator SDK][operator-sdk] and the command `operator-sdk generate kustomize manifests` to generate the files.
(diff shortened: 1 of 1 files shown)


docs/book/src/faq.md
```diff
@@ -158,36 +158,6 @@ type StructName struct {
...
-│   ├── app.mydomain.com_myapps.yaml
-│   ├── kubebuilderdemo-controller-manager-metrics-service_v1_service.yaml
-│   ├── kubebuilderdemo-metrics-reader_rbac.authorization.k8s.io_v1_clusterrole.yaml
-│   ├── kubebuilderdemo-myapp-admin-role_rbac.authorization.k8s.io_v1_clusterrole.yaml
-│   ├── kubebuilderdemo-myapp-editor-role_rbac.authorization.k8s.io_v1_clusterrole.yaml
-│   ├── kubebuilderdemo-myapp-viewer-role_rbac.authorization.k8s.io_v1_clusterrole.yaml
-│   └── my-example-operator.clusterserviceversion.yaml
-└── metadata
-    └── annotations.yaml
-```
```
camilamacedo86, line 190: I am testing it out, and it is not working
1 - In the bundle, I end up with CSV and metrics service only
2 - the bundle should contain either the CRDs for API, rules and etc
Sorry, it is my fault for not trying out during the review.

rayowang: > I am testing it out, and it is not working 1 - In the bundle, I end up with CSV and metrics service only 2 - the bundle should contain either the CRDs for API, rules and etc Sorry, it is my fault for not trying out during the review.

The previous detailed steps included these, so wouldn’t it be better to keep the details as they are?

PR #4937, opened 2025-07-18
🐛 (CLI) fix alpha update command to version pre-require check

mayuka-c:
Found this while I was working on writing the Unit Test cases for alpha cli commands.

The alpha update command --to-version verifies if the 'v' prefix is added by user, if not prepends 'v' to the specified version. But here, the check is done on --from-version instead of --to-version which might cause 'v' to be prefixed even if user specifies it in the command line for --to-version field


camilamacedo86 (approved):
Good catcher thank you 🥇 

/lgtm
/approve
/ok-to-test

c/c @vitorfloriano

PR #4939, opened 2025-07-19
🌱 (ci) - Add test for release mock

camilamacedo86:

### DIFF
--- a/.github/workflows/release-version-ci.yml
+++ b/.github/workflows/release-version-ci.yml
@@ -1,0 +1,59 @@
+name: Test GoReleaser and CLI Version
+
+on:
+  push:
+    paths:
+      - 'pkg/cmd/**'
+      - 'build/.goreleaser.yml'
+      - '.github/workflows/release-version-ci.yml'
+  pull_request:
+    paths:
+      - 'pkg/cmd/**'
+      - 'build/.goreleaser.yml'
+      - '.github/workflows/release-version-ci.yml'
+
+jobs:
+  mock-build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Clean dist directory
+        run: rm -rf dist || true
+
+      - name: Create temporary git tag
+        run: |
+          git tag v4.5.3-rc.1
+
+      - name: Install Syft to generate SBOMs
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release-version-ci.yml
```diff
@@ -0,0 +1,59 @@
...
+          mkdir test-operator && cd test-operator
+          go mod init test-operator
+          chmod +x ../dist/kubebuilder_linux_amd64_v1/kubebuilder
+          ../dist/kubebuilder_linux_amd64_v1/kubebuilder init --domain example.com
+
+          echo "PROJECT file content:"
+          cat PROJECT
+
+          echo "Verifying cliVersion value..."
+          grep '^cliVersion: 4.5.3-rc.1$' PROJECT
```
bavarianbidi, line 59: for the `init` subcommand we already have e2e-tests. If it's just to test the generated binary via `goreleaser`, than this might be a simpler approach - WDYT?

```suggestion
      - name: check kubebuilder cliVersion
        run: |
          echo "Verifying cliVersion value..."
          ../dist/kubebuilder_linux_amd64_v1/kubebuilder init  | grep '4.5.3-rc.1'
```

bavarianbidi (commented):
nice idea.
One small comment

.github/workflows/release-version-ci.yml
camilamacedo86: I think it is nice we have the content in case it fails, so we do not need to take the steps locally to discover the reason. WDYT? Could we proceed as it is?

bavarianbidi: sure, we can proceed as it is.

The following is just a small Feedback from a _new contributor PoV_: 
I totally got your intention and the way you've implemented this, is definitely cool.
But I also love the "defacto standard" in the k8s ecosystem where mostly a `make lint && make test` works out of the box on my machine and I get early feedback - even before i push and raise a PR.

And sometimes users have to spend more time in understanding and fixing code/tests which are not covered by a local test setup (which sometimes isn't that easy for sure).

Having this scenario as `ginkgo` based e2e-test would make it more accessible, as at least the `.github` folder is something I mostly do not care about.
But this also introduces new complexity - and we have to deal with that as well.

camilamacedo86: We use Goreleaser and GitHub Actions for our release process.
This test is specifically validating Goreleaser / Release process

Why do we check the CliVersion?
Because it's only output from the Binary released.
(tag version);Not from the one build with make install.
See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/build/.goreleaser.yml#L35


That’s why this doesn’t quite fit in the standard GoLang e2e test suite—what we really want is to test the release process itself, not just the CLI in isolation.

PR #4944, opened 2025-07-20
✨ (CLI;go/v4) Upgrade golang version from 1.24.0 to 1.24.5

camilamacedo86:


camilamacedo86:
/hold

I do not know if it will block people to get it as a lib

camilamacedo86:
/hold cancel

I think we can go.

afzal442 (approved):

PR #4945, opened 2025-07-20
✨ Upgrade Golang-CI from v2.1.6 to v2.2.2

camilamacedo86:
- For Kubebuilder CLI and scaffolds

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -84,6 +84,9 @@ linters:
       - linters:
           - revive
         text: "should have comment or be unexported"
+      - linters:
+          - revive
+        text: "var-naming: avoid meaningless package names"
     paths:
       - third_party$
       - builtin$
--- a/.github/workflows/lint-sample.yml
+++ b/.github/workflows/lint-sample.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Run linter
         uses: golangci/golangci-lint-action@v8
         with:
-          version: v2.1.6
+          version: v2.2.2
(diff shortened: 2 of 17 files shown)


camilamacedo86:
Small silly bump.
I want to see if it will solve the problem to install KAL
So, I am moving forward has no reason for nobody waste time on reviews
If it passes in the CI, it is all fine.

.golangci.yml
```diff
@@ -84,6 +84,9 @@ linters:
       - linters:
           - revive
         text: "should have comment or be unexported"
+      - linters:
+          - revive
+        text: "avoid meaningless package names"
```
erikgb:
```suggestion
        text: "var-naming: avoid meaningless package names"
```
To put some more context into this exclusion. But it would be even better if the revive linter is reconfigured instead, ref. https://golangci-lint.run/usage/linters/#revive

camilamacedo86: Thank you 🥇🚀
I accept your suggestion, but I’m not sure how to reconfigure it myself. (I would need spend some time checking it out) Maybe we can add an issue in the repo, so I or someone else can check it out later and try to improve how it works.

erikgb: Please create an issue for it! Should be pretty straight forward, ref.
- https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#var-naming
- https://golangci-lint.run/usage/linters/#revive

erikgb (commented):
/lgtm

PR #4947, opened 2025-07-20
✨ (go/v4): upgrade cert-manager from v1.16.3 v1.18.2 used in the e2e tests

camilamacedo86:


camilamacedo86:
Small bump.
We have many PRs open to review that we need a help with
This one, if it passes the tests, is fine.
So, moving forward.

PR #4948, opened 2025-07-20
🐛 remove: Prometheus install/check helpers from e2e utils

camilamacedo86:
## 🧹 What changed

We removed the Prometheus helpers from the E2E test utils:

- `installPrometheus`
- `uninstallPrometheus`
- `isPrometheusInstalled`

## 🧠 Why

We don’t use Prometheus in E2E tests anymore. The tests work fine without it because:

- We don’t actually rely on Prometheus during tests
- We can check reconciliation metrics using `controller-runtime` (which gives us basic metrics by default)

This cleanup makes the code simpler and easier to maintain.
PS: we kept the example in the docs to show how to do.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
@@ -63,50 +63,6 @@ func Run(cmd *exec.Cmd) (string, error) {
 	return string(output), nil
 }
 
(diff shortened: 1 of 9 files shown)


camilamacedo86:
Just remove unused methods
Since it is a bug and we are removing what is no longer used 
I am moving forward.
We have a lot of PRs needing help to review
So, I am trying to reduce the burn

Issue #4712, opened 2025-03-28
e2e tests leave resources behind after running


mahmoudhossam:
### What broke? What's expected?

Running e2e tests leaves cert-manager related resources behind, causing failures in subsequent runs.

Namely, the resources `leases.coordination.k8s.io/cert-manager-cainjector-leader-election` and `leases.coordination.k8s.io/cert-manager-controller` in the `kube-system` namespace.

### Reproducing this issue

Run `make test-e2e` then `kubectl get -n kube-system leases.coordination.k8s.io`

### KubeBuilder (CLI) Version

```
Version: cmd.version{KubeBuilderVersion:"4.5.2", KubernetesVendor:"1.32.1", GitCommit:"7c707052daa2e8bd51f47548c02710b1f1f7a77e", BuildDate:"2025-03-26T19:24:40Z", GoOs:"darwin", GoArch:"arm64"}
```

### PROJECT version

3

### Plugin versions

```yaml
layout:
- go.kubebuilder.io/v4
```

### Other versions

Kubectl:
```
Client Version: v1.32.3
Kustomize Version: v5.5.0
Server Version: v1.32.2
```
Go 1.23.0
sigs.k8s.io/controller-runtime v0.20.0

mahmoudhossam:
> Thanks for opening this issue!
> 
> Just to better understand the context, could you please clarify a few things?
> 
>     * What E2E tests are you running?
>       
>       * Are you referring to the Kubebuilder E2E tests used to maintain the project itself?
>       * Or are you running E2E tests scaffolded in a project generated by Kubebuilder?
> 
>     * How are you executing the tests?
>       
>       * Can you please share the exact steps or commands you're using to reproduce the scenario?
> 
> 
> This information will help us investigate and assist you more effectively. Thanks again!

I'm running the E2E tests scaffolded in a project generated by Kubebuilder.

To be more specific: the problem is with [this code](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/getting-started/testdata/project/test/utils/utils.go#L110).

This way of uninstalling cert-manager doesn't take into account the resources that could be generated by cert-manager as part of its runtime.

Maybe using helm for (un)installing cert-manager will fix this?

camilamacedo86:
Hi @mahmoudhossam,

Thank you for your response!

> I'm running the E2E tests scaffolded in a project generated by Kubebuilder.

[...]

PR #4949, opened 2025-07-20
🐛 (go/v4)(fix): (e2e) delete CertManager leftover leases in kube-system (not cleaned by default)

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/4712

> **NOTE:** The E2E tests are designed to run in isolated environments intended specifically for them,  
> such as the GitHub Actions workflows scaffolded by default.

c/c @wallrj @erikgb @mahmoudhossam

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/test/utils/utils.go
@@ -70,6 +70,19 @@ func UninstallCertManager() {
 	if _, err := Run(cmd); err != nil {
 		warnError(err)
 	}
+
+	// Delete leftover leases in kube-system (not cleaned by default)
+	kubeSystemLeases := []string{
+		"cert-manager-cainjector-leader-election",
+		"cert-manager-controller",
+	}
+	for _, lease := range kubeSystemLeases {
+		cmd = exec.Command("kubectl", "delete", "lease", lease,
+			"-n", "kube-system", "--ignore-not-found", "--force", "--grace-period=0")
+		if _, err := Run(cmd); err != nil {
+			warnError(err)
+		}
+	}
 }
 
 // InstallCertManager installs the cert manager bundle.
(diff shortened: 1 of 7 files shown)


camilamacedo86:
Hi @mahmoudhossam
 
Here is the fix for the issue that you opened !!!
It would be nice to get your help to review.

mahmoudhossam:
> Hi @mahmoudhossam
> 
> Here is the fix for the issue that you opened !!! It would be nice to get your help to review.

Hi @camilamacedo86, thank you for this!

I haven't touched the relevant part of the tests so didn't have any feedback for you, sorry about that.

However, looking at the PR code I don't see any mention of the `leases.coordination.k8s.io` API group, can you point me to where this is addressed in the code?

Bear in mind that deleting the `cert-manager` namespace won't take care of that as the resources in #4712 are in the `kube-system` namespace.

camilamacedo86:
/hold 
I will need work more on this one

erikgb (commented):
I don't understand why you're rewriting the uninstall. Shouldn't the change just be to add the cleanup of leases in kube-system namespace? Reversing the installation procedure with `kubectl delete` should work except from this.

camilamacedo86:
> I don't understand why you're rewriting the uninstall. Shouldn't the change just be to add the cleanup of leases in kube-system namespace? Reversing the installation procedure with kubectl delete should work except from this.

Hi @erikgb 

My motivation was to cover all: https://cert-manager.io/docs/installation/kubectl/#uninstalling
To avoid any other possible scenario.

However, keep it as before, with only the specific caveat that it would be a less verbose, cleaner solution.
I think that is a good idea since it is either a bug fix only

erikgb:
> My motivation was to cover all: https://cert-manager.io/docs/installation/kubectl/#uninstalling To avoid any other possible scenario.

Your current approach will make this more fragile to future changes in cert-manager. So I would recommend to just do the fix for leases.

camilamacedo86:
Hi @erikgb 

Thank you for the amazing help
Done 🎉

erikgb (commented):
/lgtm

🚀

PR #4932, opened 2025-07-16
✨  (go/v4;helm/alpha-v1;kustomize/v2): Adding default image policy for the manager

adarshagrawal38:
Adding image pull policy tag to the helm template
Fixing issue: #4859

Issue #4912, opened 2025-07-07
Generated helm chart should align with other charts from ecosystem


calind:
### What do you want to happen?

Currently, the generated helm resources do not follow the conventions in the helm ecosystem, which makes them somewhat "alien" and hard to customize.

Here are a few examples:

- Resource names are not templated by `{{ include "chart.fullname" . }}`. In Helm charts, usually resource names take the form:
```yaml
metadata:
  name: {{ include "chart.fullname" . }}
  labels:
    {{- include "chart.labels" . | nindent 4 }}
```
- Resources are missing features and customization options (like `nodeSelector`, `tolerations`). A chart scaffolded with `helm create chart` provides these default values:
```yaml
# Default values for chart.
# This is a YAML-formatted file.
# Declare variables to be passed into your templates.

# This will set the replicaset count more information can be found here: https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/
replicaCount: 1

# This sets the container image more information can be found here: https://kubernetes.io/docs/concepts/containers/images/
image:
  repository: nginx
  # This sets the pull policy for images.
  pullPolicy: IfNotPresent
  # Overrides the image tag whose default is the chart appVersion.
  tag: ""

# This is for the secrets for pulling an image from a private repository more information can be found here: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
imagePullSecrets: []
# This is to override the chart name.
nameOverride: ""
fullnameOverride: ""

# This section builds out the service account more information can be found here: https://kubernetes.io/docs/concepts/security/service-accounts/
serviceAccount:
  # Specifies whether a service account should be created
  create: true
  # Automatically mount a ServiceAccount's API credentials?
  automount: true
  # Annotations to add to the service account
  annotations: {}
  # The name of the service account to use.
  # If not set and create is true, a name is generated using the fullname template
  name: ""

# This is for setting Kubernetes Annotations to a Pod.
# For more information checkout: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
podAnnotations: {}
# This is for setting Kubernetes Labels to a Pod.
# For more information checkout: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
podLabels: {}

podSecurityContext: {}
  # fsGroup: 2000

securityContext: {}
  # capabilities:
  #   drop:
  #   - ALL
  # readOnlyRootFilesystem: true
  # runAsNonRoot: true
[...]

PR #4952, opened 2025-07-21
Improve helmchart template adding fields

ronaldosaheki:
This includes most of changes also discussed in both of this PRs:
ImagePullPolicy: https://github.com/kubernetes-sigs/kubebuilder/pull/4932
Align generated helmcharts witth most common helmcharts: https://github.com/kubernetes-sigs/kubebuilder/issues/4912

Added fields, changes done in the kustomize, helmchart and values scaffold, following the same style, for kustomize the added fields were added commented out and ready to edit similar to existing fields in kustomize scaffold.

Added fields:
- Pod annotations
- Node selector configuration
- Tolerations for node taints
- Topology spread constraints for better distribution
- Image pull secrets for private registries
- Image pull policy configuration
- Extra volume mounts for additional configs
- Extra volumes for the pod

Also updated testdata and docs to pass branch github pipeline.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/manager/manager.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/manager/manager.yaml
@@ -44,12 +44,37 @@ spec:
       #             values:
       #               - amd64
       #               - arm64
-      #               - ppc64le
-      #               - s390x
       #           - key: kubernetes.io/os
       #             operator: In
       #             values:
       #               - linux
+      # TODO(user): Uncomment the following code to configure the nodeSelector expression
+      # annotations, nodeSelector, tolerations, topologySpreadConstraints.
+      # nodeSelector:
+      #   disktype: ssd
+      # Annotations for the manager pod
+      # annotations:
+      #   example-annotation: "value"
+      # Node selector for scheduling the manager pod
+      # nodeSelector:
+      #   disktype: ssd
+      # Tolerations for the manager pod
+      # tolerations:
+      #   - key: "key1"
+      #     operator: "Equal"
+      #     value: "value1"
+      #     effect: "NoSchedule"
+      # Topology spread constraints for the manager pod
+      # topologySpreadConstraints:
+      #   - maxSkew: 1
+      #     topologyKey: "topology.kubernetes.io/zone"
+      #     whenUnsatisfiable: "ScheduleAnyway"
+      #     labelSelector:
+      #       matchLabels:
+      #         app: controller-manager
+      # imagePullSecrets for pulling images from private registries
+      # imagePullSecrets:
+      #   - name: myregistrykey
       securityContext:
         # Projects are configured by default to adhere to the "restricted" Pod Security Standards.
         # This ensures that deployments meet the highest security requirements for Kubernetes.
@@ -65,6 +90,8 @@ spec:
           - --health-probe-bind-address=:8081
         image: controller:latest
         name: manager
+        # imagePullPolicy for the manager container (e.g., Always, IfNotPresent, Never)
+        # imagePullPolicy: IfNotPresent
         ports: []
         securityContext:
           readOnlyRootFilesystem: true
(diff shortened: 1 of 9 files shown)


docs/book/src/cronjob-tutorial/testdata/project/config/manager/manager.yaml
```diff
@@ -44,12 +44,37 @@ spec:
       #             values:
       #               - amd64
       #               - arm64
-      #               - ppc64le
-      #               - s390x
```
camilamacedo86, line 48: We cannot remove those

docs/book/src/cronjob-tutorial/testdata/project/config/manager/manager.yaml
```diff
@@ -65,6 +90,8 @@ spec:
           - --health-probe-bind-address=:8081
         image: controller:latest
         name: manager
+        # imagePullPolicy for the manager container (e.g., Always, IfNotPresent, Never)
+        # imagePullPolicy: IfNotPresent
```
camilamacedo86, line 94: I think we can add that as default
Could we have a PR only for the image

docs/book/src/cronjob-tutorial/testdata/project/config/manager/manager.yaml
```diff
@@ -93,7 +120,16 @@ spec:
           requests:
             cpu: 10m
             memory: 64Mi
+        # Extra volumeMounts for the manager container
+        # volumeMounts:
+        #   - name: extra-config
+        #     mountPath: /etc/extra
```
camilamacedo86, line 126: I do not know if that will work with the patches but is a good call I think we need a PR for that and e2e tests to ensure that we do not break the behaviour

docs/book/src/cronjob-tutorial/testdata/project/config/manager/manager.yaml
```diff
@@ -44,12 +44,37 @@ spec:
...
       #               - arm64
-      #               - ppc64le
-      #               - s390x
       #           - key: kubernetes.io/os
       #             operator: In
       #             values:
       #               - linux
+      # TODO(user): Uncomment the following code to configure the nodeSelector expression
+      # annotations, nodeSelector, tolerations, topologySpreadConstraints.
+      # nodeSelector:
```
camilamacedo86, line 53: Should it be really added there or via kustomize? Is there a better way to configure it?
We will need to check out and either validate the scenario with e2e tests. 
Could we have a PR for those and keep this one only for the imagePull ?

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/values.yaml
```diff
@@ -1,10 +1,29 @@
 # [MANAGER]: Manager Deployment Configurations
 controllerManager:
   replicas: 1
+  pod:
+    # imagePullSecrets for pulling images from private registries
+    imagePullSecrets: []
```
camilamacedo86, line 6: I think I liked it 👍 less the pod above since all here would be only about controller-manager
See : https://github.com/kubernetes-sigs/kubebuilder/issues/4912
The person made a valid feedback about how we are exposing the values

I think we will need to change that we might need a PR just to change what we have today on that
but I want to share it with you.

docs/book/src/multiversion-tutorial/testdata/project/dist/chart/values.yaml
```diff
@@ -1,10 +1,29 @@
...
+    extraVolumeMounts: []
+    # extraVolumes for additional volumes in the manager pod
+    extraVolumes: []
+    # annotations for the manager pod
+    annotations: {}
   container:
     image:
       repository: controller
       tag: latest
+    # imagePullPolicy for the manager container (e.g., Always, IfNotPresent, Never)
```
camilamacedo86, line 25: Comments are not required.
it is auto-explanatory
And the default as the default avlue IfNotPresent seems the great one

pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/values.go
```diff
@@ -55,10 +55,29 @@ func (f *HelmValues) SetTemplateDefaults() error {
...
+    # extraVolumes for additional volumes in the manager pod
+    extraVolumes: []
+    # annotations for the manager pod
+    annotations: {}
   container:
     image:
       repository: controller
       tag: latest
+    # imagePullPolicy for the manager container (e.g., Always, IfNotPresent, Never)
+    imagePullPolicy: IfNotPresent
```
mateusoliveira43, line 80: this new value is not used, right? should it be removed?

camilamacedo86:
We added the helm/v2, which is based on the YAML templates.
Therefore, it should be addressed within, see: https://github.com/kubernetes-sigs/kubebuilder/pull/5058

It would be amazing if you could test out this scenario within.
So, I am closing this one, but if you see that is required please feel free to re-open

Issue #4925, opened 2025-07-13
[Help Wanted] - Increase the coverage of unit tests


camilamacedo86:
We've made good progress with e2e tests for Kubebuilder. However, we have been decreasing the amount of unit tests as shown in [Coveralls](https://coveralls.io/github/kubernetes-sigs/kubebuilder?branch=master). To enforce coverage in CI, we now need to improve **unit test coverage**.

What we need to do:
- Add more unit tests
- Add a check in the CI to enforce adding unit tests

mayuka-c:
@cmallikarjunah Can you please let me know if you have already started working on this?. Also on which files you're currently working to add unit tests. I can take up other files so that we can split the work (avoiding duplication) and complete the coverage :)

camilamacedo86:
Hi @mayuka-c,

This one will likely involve quite a bit of work across multiple people.

I recommend using the channel to coordinate and split up the work as needed. It would also be helpful to open a WIP PR early—don’t worry about perfection—just enough to show the direction you're heading in.

If coordination via the channel doesn’t work out and it feels like we need something more structured, I’m happy to help set up a better plan.

Since @cmallikarjunah was the first to ping, it might make sense for them to start from the top of the packages/dirs. You could begin from the bottom up to avoid overlapping efforts.

Let me know how it goes!

mayuka-c:
> Hi [@mayuka-c](https://github.com/mayuka-c),
> 
> This one will likely involve quite a bit of work across multiple people.
> 
> I recommend using the channel to coordinate and split up the work as needed. It would also be helpful to open a WIP PR early—don’t worry about perfection—just enough to show the direction you're heading in.
> 
> If coordination via the channel doesn’t work out and it feels like we need something more structured, I’m happy to help set up a better plan.
> 
> Since [@cmallikarjunah](https://github.com/cmallikarjunah) was the first to ping, it might make sense for them to start from the top of the packages/dirs. You could begin from the bottom up to avoid overlapping efforts.
> 
> Let me know how it goes!

Sure @camilamacedo86,

Thank you for your suggestions :)

I will kind of do bottom-up approach covering util/ helper functions. Will be putting up a PR to make sure I'm heading in right direction. Thank you for your help :)

cmallikarjunah:
Hi @camilamacedo86 /@mayuka-c,

I am starting with: 
pkg/alpha/command.go 
pkg/alpha/update.go. 

Will be putting up the PR(WiP) soon.
[...]

PR #4959, opened 2025-07-23
🌱 (CLI): Add unit tests and improve coverage for plugins/golang pkg

mayuka-c:
Issue: https://github.com/kubernetes-sigs/kubebuilder/issues/4925

Continuation from my previous UT PRs :-)
- Improves the coverage by adding UTs for plugins/golang package.
- This should cover all the UTs for plugins pkg

### DIFF
--- a/pkg/plugins/golang/options_test.go
+++ b/pkg/plugins/golang/options_test.go
@@ -82,6 +82,8 @@ var _ = Describe("Options", func() {
 						} else {
 							Expect(res.Path).To(Equal(path.Join(cfg.GetRepository(), "api", gvk.Version)))
 						}
+					} else if len(options.ExternalAPIPath) > 0 {
+						Expect(res.Path).To(Equal("testPath"))
 					} else {
 						// Core-resources have a path despite not having an API/Webhook but they are not tested here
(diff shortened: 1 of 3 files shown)


pkg/plugins/golang/options_test.go
```diff
@@ -82,6 +82,8 @@ var _ = Describe("Options", func() {
 						} else {
 							Expect(res.Path).To(Equal(path.Join(cfg.GetRepository(), "api", gvk.Version)))
 						}
+					} else if len(options.ExternalAPIPath) > 0 {
```
camilamacedo86, line 85: We might want to cover those options
To know more about see: https://kubebuilder.io/reference/using_an_external_resource

Also, you can check sample values in testdata samples like: 
https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/PROJECT#L66C1-L79C14

camilamacedo86: But we can do that in follow ups :-)

mayuka-c: Sure @camilamacedo86 I will take look into this :-)

Issue #4930, opened 2025-07-16
Proposal:  Migrate from `logrus` to standard library `log/slog`


vitorfloriano:
### Summary
We propose migrating Kubebuilder from the external `github.com/sirupsen/logrus` dependency to Go's standard library `log/slog` package. This migration would eliminate external dependencies, improve performance, and align with Go's logging best practices.

### Motivation

Dependency Reduction
- Current: Kubebuilder and all generated projects inherit the logrus dependency (which is in maintainance mode)
- Proposed: Zero external logging dependencies, reducing `go.mod` complexity
- Benefit: Cleaner dependency trees for user projects

Performance Improvements

- Based on benchmarking analysis, `slog` not only runs faster, but also uses less memory, which would improve command execution in Kubebuilder. See: https://betterstack.com/community/guides/logging/best-golang-logging-libraries/#final-thoughts

Future-Proofing
- Standard Library: Guaranteed long-term compatibility and support
- Ecosystem Alignment: Go community is moving toward `slog`
- Structured Logging: First-class support for structured data

Enhanced Flexibility and Extensibility
- `slog`'s handler-based design provides superior flexibility
- `slog` enables intelligent logging behavior based on context
- `slog`'s interface-based design allows easy extension without breaking changes
- Templates can generate different logging approaches based on project needs

That could be translated into:
1. Development vs Production: Different log formats automatically
2. Plugin Ecosystem: Each plugin can define its logging requirements
3. User Customization: Generated projects can include configurable logging
4. Debugging: Rich context without performance penalty
5. Integration: Easy integration with observability tools
6. Testing: Different logging behavior in test vs production code

This flexibility makes Kubebuilder more maintainable and extensible in the long-term.

### Current Usage Analysis

Logrus is used extensively across 47 files in the codebase.

### Breaking Changes

This would be a breaking change requiring a major version bump (v5.0.0).

### Colored Output Issue (and Solution)

Since `slog` doesn't include colored output by default, we would have to implement a custom colored handler (no third-parties, ~200 LoC). This maintains the current colored CLI experience while being CI/CD friendly. This level of customization could be seen as an advantage as well.

### Benefits Summary

  | Aspect       | Current (logrus)      | Proposed (slog)     | Improvement          |
  |--------------|-----------------------|---------------------|----------------------|
  | Dependencies | External              | Standard library    | Zero external deps   |
  | Performance  | Baseline              | 37% faster          | Significant speedup  |
[...]

PR #4968, opened 2025-07-28
⚠️  Migrate from logrus to log/slog

Shubhamag12:
We want to migrate from the external github.com/sirupsen/logrus dependency to Go's standard library log/slog package. This will eliminate external dependencies and improve performance.

Fixes: https://github.com/kubernetes-sigs/kubebuilder/issues/4930

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -17,11 +17,15 @@ limitations under the License.
 package cmd
 
 import (
-	"github.com/sirupsen/logrus"
+	"log"
+	"log/slog"
+	"os"
+
 	"github.com/spf13/afero"
 
 	"sigs.k8s.io/kubebuilder/v4/pkg/cli"
 	cfgv3 "sigs.k8s.io/kubebuilder/v4/pkg/config/v3"
+	"sigs.k8s.io/kubebuilder/v4/pkg/logging"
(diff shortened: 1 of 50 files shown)


cmd/cmd.go
```diff
@@ -17,11 +17,15 @@ limitations under the License.
 package cmd
```
camilamacedo86, line 17: That is great 🥇 
Thank you very much. I will need a little time to review this properly. 
I am allowing the tests to run; we need to ensure that they are still passing all tests first.

camilamacedo86: Could you please fix the lint issues: https://github.com/kubernetes-sigs/kubebuilder/actions/runs/16567845495/job/46921850241?pr=4968

You can test it out locally with `make lint`
Also, you can fix like the cgi ( import sort ) with make lint-fix

Thank you

Shubhamag12: Thank you for your suggestions. I’ve resolved all the lint errors and pushed the updated changes.

camilamacedo86:
Hi @Shubhamag12 

That is great 🎉 I tested locally. 
The Copilot AI found some small nits
Could you please address those so that we can get this one merged?

camilamacedo86 (approved):

PR #4970, opened 2025-07-29
🐛 Docs: Install-and-build.sh script fixes and enhancements

mayuka-c:


docs/book/install-and-build.sh
```diff
@@ -38,6 +38,8 @@ if [[ ${arch} == "amd64" ]]; then
     arch="x86_64"
 elif [[ ${arch} == "x86" ]]; then
     arch="i686"
+elif [[ ${arch} == "arm64" ]]; then
+    arch="aarch64"
 fi
```
camilamacedo86, line 45: See: https://app.netlify.com/projects/kubebuilder/deploys/68883e1b2158da0008efc736

We can see that the changes broke the job.

mayuka-c: Looks like updating mdbook to latest version was causing this issue, as the latest Netlify build images does not have latest required GLIBC version. Have reverted back to `v0.4.40`. Now it has passed.

Also I had to change arch to use x86_64 as it is the only archType available for Darwin: https://github.com/rust-lang/mdBook/releases/tag/v0.4.40.

camilamacedo86 (approved):
/lgtm
/approved

Thank you 🥇 
Note: I hope that you do not mind, see that I update the title because we use the title for release notes
In this case, the change does not affect users, so they cannot be 🐛 instead of 🌱 fitbetters.

mayuka-c:
> /lgtm /approved
> 
> Thank you 🥇 Note: I hope that you do not mind, see that I update the title because we use the title for release notes In this case, the change does not affect users, so they cannot be 🐛 instead of 🌱 fit better.

Thank you so much for updating :-)

Issue #1238, opened 2019-12-04
GuestbookStatus struct redeclared in quickstart sample code


iredelmeier:
The `GuestbookStatus` struct is currently declared twice in the sample code of the [Quick Start's Press Options section](https://github.com/kubernetes-sigs/kubebuilder/blob/696e8274156173403a50d319a7af7d0feed8f2ba/docs/book/src/quick-start.md#press-options):

```
// GuestbookStatus defines the observed state of Guestbook
type GuestbookStatus struct {
	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
	// Important: Run "make" to regenerate code after modifying this file
	
	// GuestbookStatus is the set of Guestbook node specific statuses: Active or Standby
    GuestbookStatus GuestbookStatus `json:"guestbookStatus"`
    
    // Status of Guestbook Nodes
    Nodes []string `json:"nodes"`
}

type GuestbookStatus struct {

	// PodName of the active Guestbook node. 
	Active string `json:"active"`

	// PodNames of the standby Guestbook nodes.
	Standby []string `json:"standby"`
}	
```

I'm guessing that one declaration (the first?) should be renamed `Guestbook`?

/kind documentation

DirectXMan12:
Something really weird is going on there.  It looks like a merge issue.  Indentation is also off.  Can you check the history and see when that got messed up?

The root guestbook should be something like (being a bit lazy here with the typing; and tabs):

```go
type GuestBook struct {
  // metadata here too
  Spec GuestbookSpec `json:"spec"`
  Status GuestbookStatus `json:"status"`
}
```

iredelmeier:
Looks like the only relevant commit is the [initial one](https://github.com/kubernetes-sigs/kubebuilder/commit/a9860545bb0e6bb56fc4973423eca7fb36d6e08d#diff-e7f384593803dbfd8186105c2163ede5R75-R137).

In any case, it should be a pretty quick fix (famous last words 😅). Happy to submit a PR!

PR #4973, opened 2025-07-29
✨ Add feature gates support for experimental API fields

### LINKED ISSUE #849: Feature Gates
Core Kubernetes types have Feature Gates. 
They are documented here: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

This is a good practice for anyone, whether they are building a CRD as part of the K8s org, or their own API.

Sadly, CRDs don't have feature gate support built into them.

KubeBuilder could close this gap.  It could provide a feature-gate feature.  
It might work like this:
* Developer tags a field in an API as belonging to a feature gate, like `// +feature-gate the-new-hotness
* KB (or controller-runtime or controller-tools) makes your controller binary have a `--feature-gates` flag.
* all "gate names" mentioned in tags are added as valid comma-separated arguments to `--feature-gates` flag. 
* When installing the CRD, controller-tools will blank out disabled parts of the schema.  This causes pruning to prune them.
* Generated documentation lists the supported feature gate strings.

/kind feature

wazery:
✨ Add feature gates support for experimental API fields

## Description

Implements initial feature gates functionality to enable/disable experimental features in Kubebuilder controllers. Adds `+feature-gate` marker support for API fields, automatic discovery during scaffolding, and `--feature-gates` flag for runtime control.

## Motivation

Enables gradual rollout of experimental features while maintaining backward compatibility. Similar to Kubernetes core feature gates, this allows developers to safely introduce new API fields and functionality.

## Changes

- Add feature gate parsing and validation (`pkg/machinery/featuregate.go`)
- Add marker discovery from API types (`pkg/machinery/featuregate_marker.go`)
- Integrate with scaffolding to generate feature gate options
- Add `--feature-gates` flag to controller main.go
- Add comprehensive tests and documentation to the Kubebuilder book

## Future Work

CRD schema modification requires controller-tools support ([issue #1238](https://github.com/kubernetes-sigs/controller-tools/issues/1238)). When implemented, `+kubebuilder:feature-gate` markers will automatically exclude disabled fields from CRD schemas.

Fixes #849


pkg/plugins/golang/v4/scaffolds/internal/templates/cmd/main.go
```diff
@@ -267,17 +269,18 @@ func main() {
 	var metricsAddr string
 	var metricsCertPath, metricsCertName, metricsCertKey string
 	var webhookCertPath, webhookCertName, webhookCertKey string
-	var enableLeaderElection bool
+		var enableLeaderElection bool
```
camilamacedo86: Could we revert the change?

pkg/plugins/golang/v4/scaffolds/internal/templates/cmd/main.go
```diff
@@ -267,17 +269,18 @@ func main() {
...
 	var enableHTTP2 bool
+	var featureGates string
 	var tlsOpts []func(*tls.Config)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. " +
 		"Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. " +
-		"Enabling this will ensure there is only one active controller manager.")
+			"Enabling this will ensure there is only one active controller manager.")
```
camilamacedo86: Could we revert the change?

test/e2e/v4/featuregates_discovery_test.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+func TestFeatureGateDiscoveryInTestdata(t *testing.T) {
+	// Test feature gate discovery in our testdata project
+	parser := machinery.NewFeatureGateMarkerParser()
+
+	// Parse the testdata project API types
+	markers, err := parser.ParseFile("../../../testdata/project-v4/api/v1/captain_types.go")
+	assert.NoError(t, err)
+	assert.NotEmpty(t, markers, "Should discover feature gate markers")
```
camilamacedo86: We will need to change the tests here a little to adjust it to the project style.
For all we have a suite_test like: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/deployimage/e2e_suite_test.go

Then, before each test to allow us have in the TestContext the Kubebuilder bin and a temp directory as remove all after the those are executed we: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/deployimage/plugin_cluster_test.go#L37-L50

By last we mock the scenarios like: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/deployimage/generate_test.go#L38-L52

ANd we have a utils to inject, replace code such as: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/v4/generate_test.go#L254-L278

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -119,6 +119,8 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	# TODO: When controller-tools supports feature gates (issue #1238), add:
+	# $(CONTROLLER_GEN) --feature-gates=$(FEATURE_GATES) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```
camilamacedo86: We will need to think about how we test it out with the feature gates.
We probably need a e2e tests that run with manifests for alpha/beta, etc. 
That would be called in the e2e tests, for example, to deploy the solution and would patch all experimental features. WDYT does it make sense?

pkg/plugins/golang/v4/scaffolds/api_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+	scaffolder := NewAPIScaffolder(cfg, res, false)
+	require.NotNil(t, scaffolder, "Expected scaffolder to be created")
+
+	apiScaffolder, ok := scaffolder.(*apiScaffolder)
+	require.True(t, ok, "Expected apiScaffolder type")
+
+	assert.Equal(t, cfg, apiScaffolder.config)
+	assert.Equal(t, res, apiScaffolder.resource)
+	assert.False(t, apiScaffolder.force)
```
camilamacedo86: We do not use assert on the tests.
We do all tests with 

```
	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"
```

Example: See; https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/go_version_test.go

So, could we change it in favor of use gomega and ginkgo

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/spf13/afero v1.14.0
 	github.com/spf13/cobra v1.9.1
 	github.com/spf13/pflag v1.0.7
+	github.com/stretchr/testify v1.10.0
```
camilamacedo86: We are very careful with new deps.
We use ginkgo only for the tests, so I think we need to keep the project standard.
Do you mind revert it?

docs/book/src/reference/markers/feature-gates.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Implementation Status
+
+### ✅ Production Ready
+
+- Feature gate discovery and validation
+- Controller integration with `--feature-gates` flag
+- Scaffolding integration for new projects
+
+### 🔄 Future Enhancement
```
camilamacedo86: Could we remove the emojis from the docs?

pkg/plugins/golang/v4/scaffolds/api.go
```diff
@@ -123,3 +134,35 @@ func (s *apiScaffolder) Scaffold() error {
...
+	parser := machinery.NewFeatureGateMarkerParser()
+
+	// Try to parse the API directory
+	apiDir := "api"
+	if s.config.IsMultiGroup() && s.resource.Group != "" {
+		apiDir = filepath.Join("api", s.resource.Group)
+	}
+
+	// Check if the directory exists before trying to parse it
+	if _, err := s.fs.FS.Stat(apiDir); err != nil {
```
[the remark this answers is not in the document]
wazery: this is not correct

pkg/machinery/featuregate.go
```diff
@@ -0,0 +1,131 @@
...
+			name = strings.TrimSpace(kv[0])
+			value := strings.TrimSpace(kv[1])
+
+			switch strings.ToLower(value) {
+			case "true", "1", "yes", "on":
+				enabled = true
+			case "false", "0", "no", "off":
+				enabled = false
+			default:
+				return nil, fmt.Errorf("invalid feature gate value for %s: %s", name, value)
```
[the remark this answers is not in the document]
wazery: @camilamacedo86 what do you think about this, should we use only 'true' or 'false' as suggested?

testdata/project-v4/api/v1/captain_types.go
```diff
@@ -33,6 +33,11 @@ type CaptainSpec struct {
 	// foo is an example field of Captain. Edit captain_types.go to remove/update
 	// +optional
 	Foo *string `json:"foo,omitempty"`
+
+	// Bar is an experimental field that requires the "experimental-bar" feature gate to be enabled
+	// +feature-gate experimental-bar
+	// +optional
+	Bar *string `json:"bar,omitempty"`
```
camilamacedo86: All samples are auto-generated.
Whenever we make changes, we run make install and make generate, which calls this script to regenerate the samples:
https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh

Because of this, we can’t edit the samples directly. If the samples don’t match, the testdata CI check will fail.

For e2e tests, we create mock scenarios within the tests themselves. For example:
https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/alphagenerate/generate_test.go#L36-L62

test/e2e/v4/featuregates_test.go
```diff
@@ -0,0 +1,201 @@
...
+			ctx, err = utils.NewTestContext(util.KubebuilderBinName, "GO111MODULE=on")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(ctx).NotTo(BeNil())
+
+			// Set up test context
+			ctx.Domain = "example.com"
+			ctx.Group = "crew"
+			ctx.Version = "v1"
+			ctx.Kind = "Captain"
+			ctx.Resources = "captains"
```
camilamacedo86: We cannot change the testsdata and try to use it here.
Here we need to init the project like:

https://github.com/kubernetes-sigs/kubebuilder/blob/4155789203e97cba10717b219a699696ef5664e9/test/e2e/alphagenerate/generate_test.go#L55-L59

Then, create the API like: 

https://github.com/kubernetes-sigs/kubebuilder/blob/4155789203e97cba10717b219a699696ef5664e9/test/e2e/alphagenerate/generate_test.go#L104-L114

And to add the new spec we can use the utils : https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/util/util.go

You will see that we have helpers to insert, replace and uncomment code.
It would be like we do for example here: 

https://github.com/kubernetes-sigs/kubebuilder/blob/4155789203e97cba10717b219a699696ef5664e9/test/e2e/utils/webhooks.go#L44-L75

OR in the hack and many other palces like

https://github.com/kubernetes-sigs/kubebuilder/blob/4155789203e97cba10717b219a699696ef5664e9/hack/docs/internal/getting-started/generate_getting_started.go#L153-L161

test/e2e/v4/featuregates_test.go
```diff
@@ -0,0 +1,201 @@
...
+	ExperimentalField *string ` + "`" + `json:"experimentalField,omitempty"` + "`" + `
+`
+			// Insert the new field in the Spec struct
+			updatedContent := strings.Replace(string(typesContent),
+				"Bar *string `json:\"bar,omitempty\"`",
+				"Bar *string `json:\"bar,omitempty\"`\n"+newField,
+				1)
+
+			err = os.WriteFile(typesFile, []byte(updatedContent), 0644)
+			Expect(err).NotTo(HaveOccurred())
```
camilamacedo86: We have an utility for it: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugin/util/util.go#L235

test/e2e/v4/featuregates_discovery_test.go
```diff
@@ -0,0 +1,233 @@
+/*
```
camilamacedo86, line 1: Note that these tests are not currently run in any CI.
Since they don’t interact with a cluster, they could be added to:
https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/features.sh

However, if we want to actually validate that solutions with a feature gate are deployable—and not just test the scaffold like `init`—then we need a real e2e test.
In that case, we should set it here:
https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/setup.sh#L67-L68
so it runs using a kind cluster in Prow.

I’ll need more time to review this approach in detail, but here’s how I think it should work:
1. Create the project.
2. Create the controller.
3. Add a feature gate that outputs something like "experimental feature gate" in the logs.
4. Build and deploy, patching to enable the feature gate.
5. Validate that the logs contain the expected output.

This would give us proper end-to-end coverage for feature-gate behavior.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -119,6 +119,8 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	# TODO: When controller-tools supports feature gates (issue #1238), add:
```
camilamacedo86:
```suggestion
```

I do not think that we should change the default scaffolding, adding the TODO:
But we definilly would need to document

pkg/plugins/golang/v4/scaffolds/internal/templates/cmd/featuregates.go
```diff
@@ -0,0 +1,177 @@
...
+	if f.Path == "" {
+		f.Path = "internal/featuregates/featuregates.go"
+	}
+
+	f.TemplateBody = featureGatesTemplate
+
+	return nil
+}
+
+const featureGatesTemplate = `{{ .Boilerplate }}
```
camilamacedo86, line 46: One of the challenges here is the UX.
A feature gate is an advanced capability, and not all projects will require or use it.
We need to think about how to make it optional.

One idea is to use plugins. For example, if we run:
    kubebuilder edit plugins=feature-gate
then we could add the feature gate logic into the scaffolds.

We also have markers that can inject code, which might be useful here. For reference:
https://book.kubebuilder.io/reference/markers/scaffold

pkg/plugins/golang/v4/scaffolds/internal/templates/cmd/featuregates.go
```diff
@@ -0,0 +1,177 @@
...
+
+package featuregates
```
camilamacedo86, line 48: How is your idea?
Could we have a UX where we minimise the utilities in the scaffolds? 
Could we make it simple for common use cases?

pkg/plugins/golang/v4/scaffolds/api_test.go
```diff
@@ -0,0 +1,207 @@
+/*
```
camilamacedo86, line 1: I don’t think we should scaffold those for the project.
Instead, we should think about proposing feature gates in a way that is easy to use.
We also need to be careful not to overload projects with too many helpers.
We really avoid to do that, since it is hard to keep maintained and support as for the end users understand 

See how I was thinking for this work: 
 https://github.com/kubernetes-sigs/kubebuilder/issues/849#issuecomment-3084109006

camilamacedo86: All logic to generate the CRD schemas must be in controller-runtime.
If you are doing it as an experiment, it's fine. But we cannot release and change, changes in the scaffold are burn for end users and we must be very careful,

wazery:
Hi @camilamacedo86! 👋 

Thank you for the detailed feedback so much, I truly appreciate them and how you cleared many points up for my understanding.

I've addressed all the points you raised:

### **1. Made Feature Gates Optional**
- Added `--with-feature-gates` flags to both `kubebuilder init` and `kubebuilder create api`
- Infrastructure only generated when explicitly requested OR auto-detected
- Auto-detection works when existing infrastructure is found OR `+feature-gate` markers are discovered
- No scaffolding by default - keeps projects clean for common use cases

### **2. Cleaned Up Testdata Files** 
- Removed manual feature gate additions from captain_types.go
- Fixed unit tests to expect clean testdata (no feature gate markers by default)
- Testdata files remain auto-generated as intended

### **3. E2E Tests Use Fresh Projects**
- E2E tests create new projects with `kbc.Init()` and `kbc.CreateAPI()`
- Use `pluginutil.InsertCode()` to add feature gate markers programmatically
- No dependency on pre-existing testdata with manual additions

### **4. Kubernetes-Compatible Parsing (suggested by Copilot)**
- Already using strict 'true'/'false' values only (no 'yes'/'on'/'no'/'off')
- Consistent with Kubernetes feature gate patterns

### **5. Simple UX Focus**
- CLI flags are straightforward: `--with-feature-gates`
- Auto-detection reduces friction for users with existing markers
- Documentation covers common scenarios (all features disabled, all enabled, specific gates)
- No unnecessary complexity for basic use cases

## **Implementation Summary:**

**Optional Infrastructure:**
```bash
# Explicit scaffolding
kubebuilder init --with-feature-gates
kubebuilder create api --with-feature-gates

# Auto-detection (when markers found or infrastructure exists)
kubebuilder create api  # discovers +feature-gate markers automatically
```

**Clean Integration:**
- Feature gates only appear when needed
- No TODOs in scaffolded code
- Proper validation and error handling
- Ready for future controller-tools integration

All unit tests pass, e2e tests work correctly, and the implementation follows Kubebuilder's design principles. The approach prioritises simplicity while providing the flexibility needed for experimental features.

Regarding the `kubebuilder edit plugins=feature-gate` suggestion - I opted for the `--with-feature-gates` flag approach as it's simpler, more discoverable, and aligns better with your emphasis on not overloading projects with unnecessary complexity. The plugin system IMHO seems better suited for external/third-party integrations, wdyt?

camilamacedo86 (commented):
Really thank you for all

> Auto-detection works when existing infrastructure is found OR +feature-gate markers are discovered
> No scaffolding by default - keeps projects clean for common use cases

The problem is the end-users project should not have the code for the markers
The makers should be implemented in the tools and be available for all of them
In this case, because it is related to how the CRDs are generate this marker should be implemented in controller-tools instead of kubebuilder. 

I hope that makes sense. WDYT about pushing a PR with the code to controller-tools?

camilamacedo86:
Hi @wazery 

I know that you are working on in add the feature gates in : https://github.com/kubernetes-sigs/controller-tools/pull/1259/files

Please, feel free to open a PR with only the docs or we can wait have all. See: https://github.com/kubernetes-sigs/kubebuilder/pull/4973#pullrequestreview-3115593609

I am closing this one for now, I hope that you do not mind.
But please feel free to re-open if you see fit.

wazery:
Hi @camilamacedo86 

Thank you so much for checking, and apologies for the delayed response. I was considering closing this PR yesterday until the other one in the `controller-tools` is finalised, since this is still a work in progress. I didn’t want to introduce documentation now and then have to change it later. I really appreciate your kind offer and suggestions.

Many thanks again!

camilamacedo86:
We need to say thank you for looking on that.
Really thank you !!!
And we are looking for all your amazing follow up PRs 🚀

Issue #4972, opened 2025-07-29
Missed golangci-lint version update in Makefile


Shubhamag12:
### What broke? What's expected?

In https://github.com/kubernetes-sigs/kubebuilder/pull/4957 we updated golangci-lint to v2.3.0 to address issues related to var-naming rule, however, this might have been missed for the Makefile

Currently, the following line in the Makefile still installs v2.2.2:

```
golangci-lint:
	@[ -f $(GOLANGCI_LINT) ] || { \
	GOBIN=$(shell pwd)/bin go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.2.2 ;\
	}
```

As a result, running make golangci-lint installs the outdated version, leading to below error
```
ERRO [linters_context] setup revive: cannot configure rule: "var-naming": invalid third argument to the var-naming rule. Expecting extraBadPackageNames of type slice of strings, but []interface {}
```

Expected Fix:
Update the version in the Makefile to v2.3.0.

### Reproducing this issue

Running `make-lint` will reproduce this issue

### KubeBuilder (CLI) Version

4.7.1

### Plugin versions

```yaml

```

camilamacedo86:
Yes, that are the 2 places that we need to fix: https://github.com/search?q=repo%3Akubernetes-sigs%2Fkubebuilder%20v2.2.2&type=code

Would you like to push a PR?

Shubhamag12:
> Would you like to push a PR?

Yes I will create a PR for this

PR #4957, opened 2025-07-22
✨ Upgrade Golang-CI from v2.2.2 to v2.3.0

aman4433:
This PR updates golangci-lint to v2.3.0 to resolve issues with the var-naming rule in revive.

context - https://github.com/kubernetes-sigs/kubebuilder/pull/4954#issuecomment-3102383102

PR #4974, opened 2025-07-29
✨ Version bump for golangci-lint to 2.3.0 ( follow up #4957 )

Shubhamag12:
Fixes: https://github.com/kubernetes-sigs/kubebuilder/issues/4972

This fix updates the version in the Makefile and lint yaml to v2.3.0.

### DIFF
--- a/.github/workflows/lint-sample.yml
+++ b/.github/workflows/lint-sample.yml
@@ -43,7 +43,7 @@ jobs:
       - name: Run linter
         uses: golangci/golangci-lint-action@v8
         with:
-          version: v2.2.2
+          version: v2.3.0
           working-directory: ${{ matrix.folder }}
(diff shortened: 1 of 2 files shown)


aman4433:
Hey @Shubhamag12, Thanks for finding this and making the fix! This looks good to me.

camilamacedo86 (approved):
Thank you 🚀

camilamacedo86:
@Shubhamag12 

I hope that you do not mind, but I changed the title because it is used in the release notes.
I changed for sparkles and added the follow-up because we did not release the change yet
So, is not a bug until we do a release :-) 

I added the follow-up because, in this case, I will need to manually edit the release notes to ensure that the entry has both PRs. Just to facilitate.

PR #4975, opened 2025-07-29
🌱 (alpha update) add validation for equal versions

vitorfloriano:
### Description
This PR introduces a validation step in the `kubebuilder alpha update` command to prevent updates when the `--from-version` and `--to-version` flags are set to the same value.

The goal is to avoid triggering unnecessary updates when the project is already using the latest version. This is particularly useful in cron jobs or automated workflows.

This PR also includes an end-to-end test to verify the new validation logic.

### Manual testing:
When attempting to run `alpha update` on a project already scaffolded with the latest version, the user will see the following message:

```fsh
vitor@PMI14:~/go/src/github.com/vitorfloriano/test-operator$ ../kubebuilder/bin/kubebuilder alpha update
WARN No --from-branch specified, using 'main' as default
INFO Checking if is a git repository
INFO Checking if branch has uncommitted changes
INFO Project is already at version v4.7.1. No update needed
Error: failed to validate equal versions: from-version (v4.7.1) and to-version (v4.7.1) are the same. No update operation required
Usage:
  kubebuilder alpha update [flags]

Flags:
      --force                 Force the update even if conflicts occur. Conflicted files will include conflict markers, and a commit will be created automatically. Ideal for automation (e.g., cronjobs, CI).
      --from-branch string    Git branch to use as current state of the project for the update.
      --from-version string   binary release version to upgrade from. Should match the version used to init the project and bea valid release version, e.g., v4.6.0. If not set, it defaults to the version specified in the PROJECT file.
  -h, --help                  help for update
      --to-version string     binary release version to upgrade to. Should be a valid release version, e.g., v4.7.0. If not set, it defaults to the latest release version available in the project repository.

Global Flags:

FATA error executing command: failed to validate equal versions: from-version (v4.7.1) and to-version (v4.7.1) are the same. No update operation required 
```


pkg/cli/alpha/internal/update/validate.go
```diff
@@ -37,6 +37,9 @@ func (opts *Update) Validate() error {
 	if err := opts.validateSemanticVersions(); err != nil {
 		return fmt.Errorf("failed to validate the versions: %w", err)
 	}
+	if err := opts.validateEqualVersions(); err != nil {
+		return fmt.Errorf("failed to validate equal versions: %w", err)
```
camilamacedo86: Hey, thanks for the PR — it looks great! 🙌  
It's a nice but I do not think that we can return an error here.

Otehrwise, it will not play nicely with GitHub Actions, I think we shouldn't end with an error when no update is needed.

If I run `kubebuilder alpha update` and I'm already on the latest version, I’d expect it to just log something like:

> _"Your project is already using the latest scaffolded version. No upgrade available."_

Same if `--to-version` is passed and matches the current version — a log like:

> _"Your project already uses the version specified. No action taken."_

This would help avoid CI errors in no-op cases. Thanks again! 😊

camilamacedo86 (approved):
/ok-to-test
/approved
/lgtm

camilamacedo86 (approved):
/lgtm

Issue #4977, opened 2025-08-04
Test GitHub Actions scaffolded and fix them if required


camilamacedo86:
### What to do
- Scaffold a new project using Kubebuilder
- Push the project to a new GitHub repository
- Verify that all GitHub Actions run and pass

Test the following scenarios (including combinations of options):

We need to cover a range of project configurations to ensure that all scaffolded variants are CI-compliant. Refer to:
https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/testdata/generate.sh

#### Test Scenarios & Commands

These are the scenarios to test for GitHub Actions using Kubebuilder scaffolds. Each block includes the commands needed to generate the project variant.

- [ ] Scenario 1: `init` only (basic init) - The issue found was tracked in : https://github.com/kubernetes-sigs/kubebuilder/issues/5033

```
kubebuilder init --plugins="go/v4" --domain testproject.org --license apache2 --owner "The Kubernetes authors"
```

- [x] Scenario 2: `create api` (single API) ->  🟢 Passed: https://github.com/kubernetes-sigs/kubebuilder/issues/4977#issuecomment-3152495701

```
kubebuilder init --plugins="go/v4" --domain testproject.org --license apache2 --owner "The Kubernetes authors"
kubebuilder create api --group crew --version v1 --kind Captain --controller=true --resource=true
```

- [x] Scenario 3: `create webhook` (with multigroup API) 🟢 Passed: 

```
kubebuilder init --plugins="go/v4" --domain testproject.org --license apache2 --owner "The Kubernetes authors"
kubebuilder edit --multigroup=true
kubebuilder create api --group crew --version v1 --kind Captain --controller=true --resource=true
kubebuilder create webhook --group crew --version v1 --kind Captain --defaulting --programmatic-validation
```

- [x] Scenario 4: Project with 2 APIs and webhooks in different groups using `edit --multigroup=true` -> 🟢 Passed: https://github.com/kubernetes-sigs/kubebuilder/issues/4977#issuecomment-3153477898

```
kubebuilder init --plugins="go/v4" --domain testproject.org --license apache2 --owner "The Kubernetes authors"
kubebuilder edit --multigroup=true

# API 1 + webhook
kubebuilder create api --group crew --version v1 --kind Captain --controller=true --resource=true
kubebuilder create webhook --group crew --version v1 --kind Captain --defaulting

# API 2 + webhook
[...]

PR #4985, opened 2025-08-05
🐛 Enable cert-manager for ci workflow

Shubhamag12:
This PR is a part of https://github.com/kubernetes-sigs/kubebuilder/issues/4977, scenario 6.

Below Jobs were failing because of this since it was trying to look for cert-manager CRDs but they were not configured as it was disabled in CI job workflow: 
https://github.com/Shubhamag12/kubebuilder-s6/actions (Run 1)

Adding it in CI workflow
Passed Jobs: 
https://github.com/Shubhamag12/kubebuilder-s6/actions (Run 2)

### DIFF
--- a/pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/github/test_chart.go
+++ b/pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/github/test_chart.go
@@ -92,18 +92,17 @@ jobs:
         run: |
           helm lint ./dist/chart
 
-# TODO: Uncomment if cert-manager is enabled
-#      - name: Install cert-manager via Helm
-#        run: |
-#          helm repo add jetstack https://charts.jetstack.io
-#          helm repo update
-#          helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --set installCRDs=true
-#
-#      - name: Wait for cert-manager to be ready
-#        run: |
-#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager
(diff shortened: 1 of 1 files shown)


pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/github/test_chart.go
```diff
@@ -92,18 +92,17 @@ jobs:
...
+        run: |
+          helm repo add jetstack https://charts.jetstack.io
+          helm repo update
+          helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --set installCRDs=true
+
+      - name: Wait for cert-manager to be ready
+        run: |
+          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager
+          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-cainjector
+          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-webhook
```
camilamacedo86, line 105: We cannot do that because it is ONLY required when we have webhooks

Users may need it for the metrics server (We need to check but as far I remember the default approach does not requires certmanager certs, then need to enable it)

So, for now, I think we need to change the Helm Plugin to:
- If it has webhooks OR if we have cert-mangaer uncommented in the config/kustomize, then we add it, uncommented 
- If not, we add it commented

Shubhamag12: Okay, makes sense.
I will work on the suggested approach

Issue #4989, opened 2025-08-06
`alpha update` runs `make` targets when there are conflicts


vitorfloriano:
### What broke? What's expected?

Currently, when running `kubebuilder alpha update --force`, if the merge ends with conflicts, the `make` targets are still run, but all fail with errors due to the presence of conflict markers in the code:

```
level=info msg="Preparing Merge branch with name tmp-merge-06-08-25-12-44 and performing merge"
level=warning msg="Merge completed with conflicts. Conflict markers will be committed."
level=info msg="Running make manifests:\n$ make manifests"
Makefile:75: *** missing separator.  Stop.
level=warning msg="make manifests failed: error running \"make\": exit status 2"
level=info msg="Running make generate:\n$ make generate"
Makefile:75: *** missing separator.  Stop.
level=warning msg="make generate failed: error running \"make\": exit status 2"
level=info msg="Running make fmt:\n$ make fmt"
Makefile:75: *** missing separator.  Stop.
level=warning msg="make fmt failed: error running \"make\": exit status 2"
level=info msg="Running make vet:\n$ make vet"
Makefile:75: *** missing separator.  Stop.
level=warning msg="make vet failed: error running \"make\": exit status 2"
level=info msg="Running make lint-fix:\n$ make lint-fix"
Makefile:75: *** missing separator.  Stop.
level=warning msg="make lint-fix failed: error running \"make\": exit status 2"
```
This is expected when running the `make` targets after the conflicts, but can negatively affect CI results. 

It can also confuse users about the success of the operation.

A better approach could be to NOT run the `make` targets if there are conflicts and only WARN the user to do it after resolving them.

### Reproducing this issue

Run `kubebuilder alpha update --force` in projects with distant current and target versions.

### KubeBuilder (CLI) Version

v4.7.1

### PROJECT version

v4.5.2

### Plugin versions

```yaml

```

camilamacedo86:
Hi @vitorfloriano 👋

We need to run the make targets because they help when someone compares the differences.
For example, even if there’s a conflict in a controller, we can still run make generate and regenerate the files.

One improvement we could add is: if there’s a conflict in the Makefile itself, then we skip running it — since in that case we know nothing will work anyway.

PR #4990, opened 2025-08-06
🐛 (alpha update) fix make run logic in merge

vitorfloriano:
`alpha update` will now run the `make` targets silently if the merge is completed with conflicts. The user will be warned about the conflicts and instructed to run the `make` targets again after resolving them:

```
INFO Preparing Merge branch and performing merge branch_name=tmp-merge-07-08-25-01-52
INFO Running make targets to generate, format, and lint code 
INFO Finished running make targets 
INFO Staging all changes for commit 
WARN Update completed with conflicts. Conflict markers were committed. 
WARN Resolve the conflicts that were committed, then run: 
WARN      make manifests generate fmt vet lint-fix 
WARN This ensures manifests and generated files are up to date, and the project layout remains consistent. 
```
Fixes #4989

PR #4992, opened 2025-08-07
🐛 (alpha commands): alpha update command with `--force` now runs post-merge fixes in best-effort mode raising warnings instead of errors to allow properly automation with this option

camilamacedo86:
This updates the behaviour of runMakeTargets after a merge:
- If --force is set, it runs all make targets (like generate, vet, etc.)
  Even if some fail, this helps keep the project in as consistent a state as possible when conflicts are faced.
  If the conflict is faced, but we are still able to regenerate the manifest, we try to do so. 
- If --force is not set, it fails immediately on the first failed target.

We now use `sh -c "make <target> || true"` when --force is enabled,
So we can continue execution, but still detect and log failures.

### DIFF
--- a/pkg/cli/alpha/internal/update/update.go
+++ b/pkg/cli/alpha/internal/update/update.go
@@ -210,13 +210,25 @@ func cleanupBranch() error {
 	return nil
 }
 
-// runMakeTargets is a helper function to run make with the targets necessary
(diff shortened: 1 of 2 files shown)


vitorfloriano:
/lgtm

Should we close #4989 and #4990

PR #4994, opened 2025-08-07
🌱 (e2e) Make alpha update tests easier to understand and more complete

camilamacedo86:
- Add a test where the user changes the version in the Makefile to trigger a conflict
- Check that both the user version and the new version appear in the conflict block
- Make sure the update keeps user code and updates go modules correctly
- Add clearer step messages using By() to make the test flow easier to follow
- Improve check to verify if has markers conflicts


camilamacedo86:
c/c @vitorfloriano

Issue #4971, opened 2025-07-29
Design Improvements[UTs]: Restructure and re-use helper functions


cmallikarjunah:
### **Design Improvements Description**
We need to create a **shared way** across **kubebuilder/pkg/ unittests and e2e tests** to reuse functionality of **creating/using binaries and temporary directory**. That way we re-use the same code across files and reduce the overhead of modifying multiple areas from the maintenance perspective. Thus, also improving readability and structure.

Existing helper for binary creation: 
https://github.com/kubernetes-sigs/kubebuilder/blob/869ca6b65027baa71f4d5410f68f7830c1c90857/test/e2e/utils/test_context.go#L64C1-L117C2

Existing helper for cleaning up dir: https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/utils/test_context.go#L251-L270

Example Usage: `https://github.com/kubernetes-sigs/kubebuilder/blob/869ca6b65027baa71f4d5410f68f7830c1c90857/test/e2e/v4/e2e_suite_test.go#L40-L43`

### **What do we need**

1. We need something like test/context.go in our repo where the helpers can be placed. 
2. kubebuilder/pkg/ unit tests to use the contexts.
3. Identify similar adjustments and add it to test/ and make sure we re-use code as much as possible for testing.

PR #4995, opened 2025-08-07
🌱 refactor TestContext to allow use in unit tests with directory setup

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/4971

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,4 +28,6 @@ docs/book/src/docs
 /testdata/**/go.sum
 /docs/book/src/simple-external-plugin-tutorial/testdata/sampleexternalplugin/v1/bin
 /testdata/**legacy**
-/test/e2e/alphaupdate/e2e-**
+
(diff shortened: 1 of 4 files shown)


camilamacedo86:
Hi @cmallikarjunah

Can you please help with the review of this one?
See that we can properly use the libs/funcs that we already have we do not need a new design.
Also, we must ensure that each dir has a suite_test.go so that we can run the tests properly locally.

The changes here ensure that the new tests added follow the Kubebuilder style
Thank you a lot for your collaboration 🚀 

I think we can continue adding the others tests now. 
We will need to adapt here to adapt the TextContext to not fail in the cluster and when have not kubectl and etc

camilamacedo86:
I’m moving forward with improving our tests. ( it is sometimes hard we get reviews )
If anyone sees something that needs to be changed in the future, we can always open a follow-up PR — please don’t hesitate to do so.

PR #5000, opened 2025-08-09
📖 (doc): Add doc about alpha update GitHub action and address some fixes/improvements for alpha update

camilamacedo86:


docs/book/src/reference/commands/workflow_auto_update.md
```diff
@@ -0,0 +1,145 @@
...
+            --head "$PR_BRANCH"
+          then
+            EXISTING="$(gh pr list --state open --head "$PR_BRANCH" --json number --jq '.[0].number' || true)"
+            if [ -n "${EXISTING}" ]; then
+              echo "PR #${EXISTING} already exists for ${PR_BRANCH}, branch updated."
+            else
+              echo "Failed to create PR for ${PR_BRANCH} and no open PR found."
+              exit 1
+            fi
+          fi
```
camilamacedo86, line 106: @vitorfloriano ^ see that I tried to refine a little.
I have not tested yet. If you know that it requires adjustments, please !!

About the question over we live 1 or 4 commit I think we should have only 1 to make the history clean when the PR get merged (git history) as facilite revert it, pull locally and etc, that is why we are creating a new branch with the kubebuilder version and pushing it. 

If you have ideas for improvements, let me know !!!

docs/book/src/reference/commands/alpha_update.md
```diff
@@ -111,41 +121,41 @@ These conflicts will be committed in the
...
+This approach is typically used in automation workflows where conflict markers are later addressed by a human,
+or where preserving the conflicting changes is acceptable for follow-up processing.
 
-## Requirements
+## Demonstration
 
-- A valid [PROJECT][project-config] file at the root of your project
-- A clean Git working directory (no uncommitted changes)
-- Git must be installed and available
+<iframe width="560" height="315" src="https://www.youtube.com/embed/J8zonID__8k?si=WC-FXOHX0mCjph71" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" referrerpolicy="strict-origin-when-cross-origin" allowfullscreen></iframe>
```
camilamacedo86, line 154: removed requirements since that did not addressed to much value
Added demo
@vitorfloriano ^ see

docs/book/src/reference/commands/alpha_update.md
```diff
@@ -111,41 +121,41 @@ These conflicts will be committed in the
 <aside class="note warning">
 You must manually resolve these conflicts before merging into your main branch.
 
-<aside class="note warning">
```
camilamacedo86, line 114: Fix it was broken 
I verified when I did the resolution / preview

docs/book/src/reference/commands/workflow_auto_update.md
vitorfloriano: @camilamacedo86 Should we change `alpha update` as well? We could clean up the branches that were created and soft reset the merge branch and do only one commit, just like we do in the action. Or we could leave as is so the user can inspect what happened, and we can move forward with the next steps and get back to this later in a follow up.

Issue #5004, opened 2025-08-10
Proposal: Speed up E2E tests by skipping suites when changes are irrelevant


camilamacedo86:
## Problem
Right now, our E2E tests run even when a PR only changes documentation or YAML files.  
This wastes CI time and slows down feedback.

## Goal
- Skip all E2E tests if a PR changes only:
  - Documentation files (*.md, .markdown,  and yml or yaml, etc.)
  - YAML files (*.yml, *.yaml) — case-insensitive
- Run alpha E2E suites (alphagenerate, alphaupdate) only if changes are in:
  - pkg/cli/alpha/
- Let each suite decide for itself whether to run, so the logic is per-suite, atomic, and works locally and in CI.

## Proposed Solution
We add a small Go helper in test/e2e/utils to:
1. Get the list of changed files from CI env vars or from git diff.
2. Check if the changes are only docs/YAML.
3. Check if any changed file matches a list of required paths for a suite.

Each E2E suite calls this helper in a BeforeSuite and skips itself if not needed.

The following code example was generated assisted by AI to help understand how it should be.  

```go
## Helper: test/e2e/utils/changes.go
package utils

import (
	"bytes"
	"fmt"
	"os"
	"os/exec"
	"path/filepath"
	"regexp"
	"strings"
)

type Options struct {
	RepoRoot           string
	Include            []string
	IncludeIsRegex     bool
	SkipIfOnlyDocsYAML bool
	BaseEnvVar         string
	HeadEnvVar         string
	ChangedFilesEnvVar string
}

type changed struct{ files []string }

func ShouldRun(opts Options) (bool, string, error) {
	if opts.RepoRoot == "" {
		opts.RepoRoot = "."
	}
	if opts.BaseEnvVar == "" {
		opts.BaseEnvVar = "PULL_BASE_SHA"
	}
	if opts.HeadEnvVar == "" {
		opts.HeadEnvVar = "PULL_PULL_SHA"
	}
	if opts.ChangedFilesEnvVar == "" {
		opts.ChangedFilesEnvVar = "KUBEBUILDER_CHANGED_FILES"
	}

	if raw := os.Getenv(opts.ChangedFilesEnvVar); strings.TrimSpace(raw) != "" {
		return decide(parseChanged(raw), opts)
	}

	base := os.Getenv(opts.BaseEnvVar)
	head := os.Getenv(opts.HeadEnvVar)
	if head == "" {
		head = "HEAD"
	}

[...]

PR #5005, opened 2025-08-11
🌱 Skip E2E tests for documentation and YAML-only changes

Gautam3994:
**Issue:** Proposal: Speed up E2E tests by skipping suites when changes are irrelevant - [#5004](https://github.com/kubernetes-sigs/kubebuilder/issues/5004)

## Problem
Currently, all E2E tests run even if a PR only modifies documentation (`*.md`, `*.markdown`) or YAML (`*.yml`, `*.yaml`) files. This results in unnecessary CI resource usage and slower feedback cycles.

## Goal
- Skip all E2E tests if the PR contains *only* documentation or YAML changes (case-insensitive).  
- Run alpha E2E suites (like `alphagenerate` and `alphaupdate`) only if changes are within `pkg/cli/alpha/`.  
- Keep the decision logic atomic and per-suite, so each E2E suite can decide locally and consistently whether to run — both in CI and local development.

## Proposed Solution
- Introduce a Go helper utility in `test/e2e/utils` that:  
  - Retrieves the changed files list either from CI environment variables or by running `git diff`.  
  - Detects if changes are *only* docs or YAML files.  
  - Checks if changes match any required path prefixes for a given suite.  
- Modify each E2E suite to call this helper in its `BeforeSuite` hook and `Skip` itself when appropriate.

This improves CI efficiency by avoiding needless E2E runs, speeding up feedback without sacrificing test coverage.


test/e2e/alphaupdate/e2e_suite_test.go
```diff
@@ -30,3 +32,14 @@ func TestE2E(t *testing.T) {
 	_, _ = fmt.Fprintf(GinkgoWriter, "Starting kubebuilder suite test for the alpha update command\n")
 	RunSpecs(t, "Kubebuilder alpha update suite")
 }
+
+var _ = BeforeSuite(func() {
+	run, why, _ := utils.ShouldRun(utils.Options{
+		RepoRoot:           ".",
+		Includes:           []string{"pkg/cli/alpha/"},
```
camilamacedo86: If we change the tests also should make make it be executed

camilamacedo86 (changes requested):
Thank you a lot for the nice work 🥇 

We have an error here to solve see the job: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kubebuilder/5005/pull-kubebuilder-test/1955569142404747264


```
make[1]: Leaving directory '/home/prow/go/src/sigs.k8s.io/kubebuilder/docs/book/src/getting-started/testdata/project'
./test/check-license.sh
Checking for license header...
./test/check-license.sh: line 29: file: command not found
./test/check-license.sh: line 29: file: command not found
./test/check-license.sh: line 29: file: command not found
./test/check-license.sh: line 29: file: command not found
./test/check-license.sh: line 29: file: command not found
./test/check-license.sh: line 29: file: command not found
./test/check-license.sh: line 29: file: command not found
./test/check-license.sh: line 29: file: command not found
./test/check-license.sh: line 29: file: command not found
./test/check-license.sh: line 29: file: command not found
./test/check-license.sh: line 29: file: command not found
./test/check-license.sh: line 29: file: command not found
./test/check-license.sh: line 29: file: command not found
./test/check-license.sh: line 29: file: command not found
./test/check-license.sh: line 29: file: command not found
./test/check-license.sh: line 29: file: command not found
```

GautamKumarMK:
@camilamacedo86  I noticed the missing file command in the build environment. I realize the guidelines recommend one issue per PR — should I open a separate PR for this fix, or is it okay to include it here since it’s a small change?

GautamKumarMK:
Hi @camilamacedo86  just checking in — would you prefer I open a separate PR for the missing file command fix, or should I keep it in this one?

GautamKumarMK:
@camilamacedo86 just checking in please let me know if anything else is required from my end

camilamacedo86:
Hi @Gautam3994 

I had no time to review this one properly yet
But you are right we should have another PR to fix the issue with the shell
If you want to do that will be amazing 🚀 

we need ensure that this changes will work fine and we will not skip when we should test
Give me a little time to review this one as soon as possible. Thank you a lot !!!

test/e2e/utils/suite_filter.go
```diff
@@ -0,0 +1,241 @@
...
+			}
+		}
+	}
+
+	return false, "no files matched include prefixes", nil
+}
+
+func onlyDocsOrYAML(files []string) bool {
+	pattern := `(?i)(^docs/|\.md$|\.markdown$|^\.github/|` +
+		`(OWNERS|OWNERS_ALIASES|SECURITY_CONTACTS|LICENSE)(\.md)?$|\.ya?ml$)`
```
vitorfloriano, line 193: @Gautam3994 Sometimes we change workflows (which are `.yaml` files located in the `.github` dir) that have config that could possibly break the app and we would want to validate that in a full E2E test. Would this logic exclude those files?

camilamacedo86:
Hi @Gautam3994 👋

Sorry for the delay in getting back to you.
I spent some time evaluating whether this was really the best way to achieve our goal, and it turns out we can handle this more directly in Prow — which makes things clearer and avoids reinventing the wheel.

See: https://github.com/kubernetes/test-infra/pull/35801

Thank you so much for your effort and contribution here — it’s really appreciated! 🙏
However, given the above, I think it makes sense to close this PR in favor of that approach.

I’ll go ahead and close this if you don’t mind.
Again, my sincere apologies if I caused any confusion or delay along the way.

Thanks once more!

PR #5013, opened 2025-08-16
(alpha update):alpha/update: add --push option and fix non-squash his…

camilamacedo86:
…tory output branch

- Add --push flag to upload the resulting output branch to origin.
- Ensure non-squash path (history rebuild) always writes to the stable output branch name returned by getOutputBranchName().
- Clarify CLI help and examples to include --push and to reflect the default output-branch naming for non-squash runs.
- Keep behavior: if merge has conflicts and --force is not set, Update() returns early and we don’t try to rebuild/push history.


vitorfloriano:
/lgtm

This covers the core behaviors correctly and is ready to ship as is. We can work on refinements in a follow-up.

PR #5017, opened 2025-08-18
✨(alpha update): add --git-config flag with clear defaults and replacement behavior

camilamacedo86:


docs/book/src/reference/commands/alpha_update.md
```diff
@@ -106,6 +107,23 @@ Run update and push the result to origin:
...
+Keep built-in defaults (renameLimit tweaks) AND add more by repeating them explicitly
+
+```shell
+kubebuilder alpha update \
+  --git-config merge.renameLimit=999999 \
+  --git-config diff.renameLimit=999999 \
+  --git-config merge.conflictStyle=diff3 \
+  --git-config rerere.enabled=true
+```
+
```
vitorfloriano: @camilamacedo86 This section may mislead the user into thinking that those configs in the codeblock are the defaults. WDYT?

camilamacedo86: Good point I changed 
Let me know wdyt now

camilamacedo86:
c/c @vitorfloriano could you please review this one?

docs/book/src/reference/commands/alpha_update.md
```diff
@@ -134,17 +151,18 @@ make all
...
+| `--push`          | Push the output branch to the `origin` remote after the update completes.                                                                                                                                                                                                                       |
+| `--git-config`    | Repeatable. Pass per-invocation Git config as `-c key=value`. **Default** (if omitted): `-c merge.renameLimit=999999 -c diff.renameLimit=999999`. **Important**: if you pass this flag, your values **replace** the defaults. To keep them *and* add more, re-specify the defaults explicitly.  |
+| `-h, --help`      | Show help for this command.                                                                                                                                                                                                                                                                     |
```
vitorfloriano: Just to check that I understand this: 

By default we set those two configs, right? If we pass any values to this flag, we won't have those two configs anymore, unless we re-add them along with the other configs we may pass. 

Did I get it right?

camilamacedo86: I changed it a little to improve the UX
Give a look now in the docs changes and let me know wdyt?
See if you agree with for we get it merged

docs/book/src/reference/commands/alpha_update.md
```diff
@@ -132,19 +133,46 @@ make manifests generate fmt vet lint-fix
...
+- **Add more on top of defaults**
+```shell
+kubebuilder alpha update \
+  --git-config merge.conflictStyle=diff3 \
+  --git-config rerere.enabled=true
+```
+
+- **Disable defaults entirely**
+```shell
+kubebuilder alpha update --git-config disable
```
vitorfloriano, line 151: @camilamacedo86 What if we keep the default values UNLESS the user explicitly overrides them? 

I think that would make the defaults more robust, if we believe the defaults are the way to go. But if the user disagrees, they can set THAT config as they wish.

Having three/four behaviors for this flag seems a bit confusing.

camilamacedo86: > @camilamacedo86 What if we keep the default values UNLESS the user explicitly overrides them?

Yes

> I think that would make the defaults more robust, if we believe the defaults are the way to go. But if the user disagrees, they can set THAT config as they wish.

Exactly. 

> Having three/four behaviors for this flag seems a bit confusing.

- we have not inform : that has the default values
- if set: then is default values + what is passed
- if want to disable set : disable

I agree that is not simple
But it is an very advanaced feature so people will not tough unless they understand
I could not came with a better ux than that, if you have ideas please feel free to share

vitorfloriano: We could keep the defaults regardless of the user adding different configs (ex, conflictStyle=diff3 won't remove the defaults).

The defaults would only be changed if the user explicitly does so. 

That way we wouldn't need the "disable" option, and we wouldn't need to add the defaults back in just because we want other, non-related, configs as well.

And, since the configs will only be set for those specific commands, I think we can be a bit more assertive/invasive about the configs we think are the best for what we are doing in `kubebuilder alpha update`.

camilamacedo86: >We could keep the defaults regardless of the user adding different configs (ex, conflictStyle=diff3 won't remove the defaults).

that is how it is.

> The defaults would only be changed if the user explicitly does so.
That is how it is.
The user must add "disable"

So, not sure if I follow your idea
But we must provide a option to allow the person opt-out besides we want ensure those by default

vitorfloriano: > that is how it is.

I understood that:
1. There are two configs used by default
2. If the user passes any values to `--git-config`, the defaults will no longer apply, even if the values passed have nothing to do with the defaults.
3. If the user wants to pass values to `--git-config` **AND** keep the two defaults, the defaults need to be passed **again**.
4. If the user wants to disable the defaults, they can pass `disable`.

My suggestion was that:
1. There are two configs used by default
2. The defaults are kept even when the user passes other configs to `--git-config`.
3. The defaults are only overriden if the user passes different values to the same configs

Let me know if I got anything wrong.

camilamacedo86: > If the user passes any values to --git-config, the defaults will no longer apply,

We changed that, see: https://github.com/kubernetes-sigs/kubebuilder/pull/5017/files#diff-3c12a0b0aee4383e0a39d64d93c5286c1f1b1384396e8fc067d284711ced078cR136

- If --git-config is **not** set: Then we have the defaults
- if --git-config is set: Then we have the defaults + values informed 
- if I want to disable the defaults then I need :  --git-config disable and if I wish add others I append  --git-config  <other-value>

docs/book/src/reference/commands/alpha_update.md
```diff
@@ -132,19 +133,46 @@ make manifests generate fmt vet lint-fix
...
+| `--output-branch` | Name of the output branch. Default: `kubebuilder-update-from-<from-version>-to-<to-version>`.                                                                                                                                                                                                   |
+| `--push`          | Push the output branch to the `origin` remote after the update completes.                                                                                                                                                                                                                       |
+| `--git-config`    | Repeatable. Pass per-invocation Git config as `-c key=value`. **Default** (if omitted): `-c merge.renameLimit=999999 -c diff.renameLimit=999999`. **Important**: if you pass this flag, your values **replace** the defaults. To keep them *and* add more, re-specify the defaults explicitly.  |
```
camilamacedo86: Fix ^ Outdated

docs/book/src/reference/commands/alpha_update.md
```diff
@@ -132,19 +133,46 @@ make manifests generate fmt vet lint-fix
...
+| `--force`         | Continue even if merge conflicts happen. Conflicted files are committed with conflict markers (CI/cron friendly).                                                                                                                                                                               |
+| `--show-commits`  | Keep full history (do not squash). **Not compatible** with `--preserve-path`.                                                                                                                                                                                                                   |
+| `--preserve-path` | Repeatable. **Squash mode only.** After copying the merge tree to the output branch, restore these paths from the base branch (e.g., `.github/workflows`).                                                                                                                                      |
```
camilamacedo86: restore-path ^

docs/book/src/reference/commands/alpha_update.md
```diff
@@ -53,6 +53,7 @@ The command creates three temporary branches:
     - `--preserve-path`: in squash mode, restore specific files (like CI configs) from your base branch.
```
camilamacedo86: Add co-author

camilamacedo86:
Hi @vitorfloriano 

I changed the the flag info in the doc.

vitorfloriano:
/lgtm

Nice addition for advanced users.

PR #5018, opened 2025-08-18
✨ (alpha update) Add option to allow open GitHub Issues after updates

camilamacedo86:
When running alpha update, you can now pass --open-gh-issue to automatically create a pre-filled GitHub Issue.

### DIFF
--- a/pkg/cli/alpha/internal/update/update.go
+++ b/pkg/cli/alpha/internal/update/update.go
@@ -71,6 +71,12 @@ type Update struct {
 	// Push, when true, pushes the OutputBranch to the "origin" remote after the update completes.
 	Push bool
 
+	// OpenGhIssue, when true, automatically creates a GitHub issue after the update
+	// completes. The issue includes a pre-filled checklist and a compare link from
+	// the base branch (--from-branch) to the output branch. This requires the GitHub
+	// CLI (`gh`) to be installed and authenticated in the local environment.
+	OpenGhIssue bool
+
 	// Temporary branches created during the update process. These are internal to the run
 	// and are surfaced for transparency/debugging:
 	//   - AncestorBranch: clean scaffold generated from FromVersion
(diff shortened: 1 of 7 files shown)


camilamacedo86:
c/c @vitorfloriano

pkg/cli/alpha/internal/update/update.go
```diff
@@ -170,6 +217,59 @@ func (opts *Update) Update() error {
...
+	if hasConflicts {
+		body = fmt.Sprintf(issueBodyTmplWithConflicts, opts.ToVersion, createPRURL)
+	} else {
+		body = fmt.Sprintf(issueBodyTmpl, opts.ToVersion, createPRURL)
+	}
+
+	issueCmd := exec.Command("gh", "issue", "create",
+		"--title", title,
+		"--body", body,
+		"--label", "kubebuilder-update",
```
vitorfloriano: Adding a label will cause the command to fail if the label does not exist. See: https://github.com/vitorfloriano/test-operator/actions/runs/17057667598/job/48358201308#step:6:164

A workaround would be to use `gh label create` with the `--force` flag first to ensure the label is created or updated. See: https://cli.github.com/manual/gh_label_create

To be sure about that, I removed the `--label` from the code and it worked. See the action run: https://github.com/vitorfloriano/test-operator/actions/runs/17057992426/job/48359125578#step:6:165 

...and the Issue it created: https://github.com/vitorfloriano/test-operator/issues/97

camilamacedo86: Hi @vitorfloriano 

That is amazing.
Let's keep it simple I removed the label
AND 
I improved the issue descriptions

Let me know wdyt?
Could we get this one merged now?

vitorfloriano: It works now!

Workflow: https://github.com/vitorfloriano/test-operator/actions/runs/17065072263/workflow

Issue: https://github.com/vitorfloriano/test-operator/issues/98

vitorfloriano:
/lgtm

I ran some tests and the `--open-gh-issue` flag successfully creates a GitHub Issue with a working URL for creating a PR using the branch that was created during `kubebuilder alpha update`.

pkg/cli/alpha/internal/update/update.go
camilamacedo86: Cool so I think that we can add by default :-)

PR #5023, opened 2025-08-19
🐛 (revert) only changes done in the default scaffold in the pr: 4932 ( kept helm chart changes )

camilamacedo86:
Why are we reverting the scaffold changes?

- Scaffold Stability: Changes to the default scaffold can't be done lightly. They can create a lot of extra work and cause issues for end-users when they upgrade to a new version. We want to avoid causing upgrade burden.
- Complexity: Adding a large number of defaults to the scaffold would make it overly complex and difficult to maintain over time.
- A Better Approach: We need to address the Helm chart plugin's design in a more comprehensive way, as outlined in the Kubebuilder issues:

- https://github.com/kubernetes-sigs/kubebuilder/issues/4833
- https://github.com/kubernetes-sigs/kubebuilder/milestone/39
- https://github.com/kubernetes-sigs/kubebuilder/issues/4912

Thank you to the original author for their contribution. This change is being made to ensure project stability for all users.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/manager/manager.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/manager/manager.yaml
@@ -64,7 +64,6 @@ spec:
           - --leader-elect
           - --health-probe-bind-address=:8081
         image: controller:latest
-        imagePullPolicy: IfNotPresent
         name: manager
         ports: []
         securityContext:
--- a/docs/book/src/cronjob-tutorial/testdata/project/dist/install.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/dist/install.yaml
@@ -4213,7 +4213,6 @@ spec:
         command:
         - /manager
         image: controller:latest
-        imagePullPolicy: IfNotPresent
         livenessProbe:
           httpGet:
             path: /healthz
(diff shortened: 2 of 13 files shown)


docs/book/src/cronjob-tutorial/testdata/project/config/manager/manager.yaml
```diff
@@ -64,7 +64,6 @@ spec:
           - --leader-elect
           - --health-probe-bind-address=:8081
         image: controller:latest
-        imagePullPolicy: IfNotPresent
```
camilamacedo86, line 67: @adarshagrawal38 

Hey I'm going to revert the changes in the default scaffold, and here's why.

I know I suggested this earlier, but after thinking it through, I realized it's not the best option. Changes to the stable default scaffold can't be done lightly because they can cause a lot of issues for users when they upgrade.

Adding all possible values and defaults would also make things very complex and hard, and bring confusion, seems like an anti pattern.

**Ultimately, we need a better, more maintainable approach for our Helm charts plugin**. See these links for more context:

- https://github.com/kubernetes-sigs/kubebuilder/issues/4833
- https://github.com/kubernetes-sigs/kubebuilder/milestone/39
- https://github.com/kubernetes-sigs/kubebuilder/issues/4912

I haven't had time to think of a solution, but I want to avoid making a **change that creates more work for our users when they upgrade**. If you have time, please feel free to look at these requirements and think about how we could address them in a better way.

Again, thank you so much for your contribution. I really appreciate your help.

PR #5026, opened 2025-08-20
✨ (alpha update): Rename alpha update flag preserve-path to restore-path

camilamacedo86:
(nore released yet so we can change besides is alpha)


vitorfloriano:
LGTM.

`--restore-path` is more aligned with what the flag actually does.

PR #5029, opened 2025-08-21
🐛 Fix git branch creation error when `main` already exists

Deofex:
Replaces:

```bash
git checkout -b main
```

with:

```bash
git branch -M main
```

in the initialize Git Repo function of the e2e test.

This prevents a fatal error when a developer has the following Git configuration:

```ini
[init]
  defaultBranch = main
```

In such cases, the `main` branch already exists after repository initialization, and `git checkout -b main` fails with:

```
fatal: A branch named 'main' already exists.
```

### Motivation

This change ensures the command works smoothly for all developers, regardless of their Git configuration, improving the reliability of the setup process.


camilamacedo86:
The change did not inpacted the tests so I am OK with

/lgtm
/approved

Thank you for checking it out :-)

PR #5030, opened 2025-08-21
🌱 (fix): linter check for YAML should consider files with sufix .yml as .yaml

camilamacedo86:


Makefile
```diff
@@ -126,7 +126,7 @@ lint-config: golangci-lint ## Verify golangci-lint linter configuration
 
 .PHONY: yamllint
 yamllint:
-	@files=$$(find testdata -name '*.yaml' ! -path 'testdata/*/dist/*'); \
+	@files=$$(find testdata -name '*.yaml' -o -name '*.yml' ! -path 'testdata/*/dist/*'); \
```
vitorfloriano, line 129: Hi @camilamacedo86 I was testing this patch and got a lot of errors in the dist directory. Perhaps the path is not being excluded properly?

PR #5032, opened 2025-08-22
📖 (alpha update): Add AI integration to the docs

vitorfloriano:
This PR adds documentation for integrating GitHub Models AI summaries into the `kubebuilder alpha update` workflow. 

The new documentation explains the design decisions, provides a ready-to-use GitHub Actions workflow, and outlines best practices and limitations for AI integration.

### DIFF
--- a/docs/book/src/reference/ai_integration.md
+++ b/docs/book/src/reference/ai_integration.md
@@ -1,0 +1,194 @@
+# AI integration for `alpha update`
+
+> Status: Experimental, optional, opt‑in workflow. Not enabled by default.
+
+## Design decisions
+
+We experimented with different AI integrations for the `alpha update` workflow and chose a single, reliable path based on the following non‑negotiable rules:
+
+- No Personal Access Tokens (PATs), no GitHub App tokens
+- No repository setting changes beyond standard Actions permissions
+- Must work with the default `GITHUB_TOKEN`
+- Must require as fewer permissions as possible
+
+Outcome:
+
+After exploring different integrations, we found out that:
+
+- The only feasible integration is using **GitHub Models** in **GitHub Actions** with the additional permission `models: read`.
+- Assigning a Copilot Agent to issues was discarded (requires extra permissions and is not available in all orgs).
+- Auto‑resolving conflicts with models was discarded (non‑deterministic, inconsistent, and risky).
+- Using GitHub Models to produce **summaries** has proven to be reliable and helpful.
+
+This page documents that single supported recipe.
+
+---
+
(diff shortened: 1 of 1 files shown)


docs/book/src/reference/ai_integration.md
```diff
@@ -0,0 +1,194 @@
...
+      run: |
+        cat > prompt.txt <<'EOF'
+        Start by stating that this summary is AI generated and can contain errors.
+        You are a concise update summarizer for an automated Kubebuilder scaffold update.
+        Analyze the patch and write a GitHub Issue comment that explains what changed.
+        Scan the patch for Git conflict markers (<<<<<<<, =======, >>>>>>>). For each conflict:
+        Explain what caused the conflict.
+        How to properly resolve them.
+        Inform the user about the recommended next steps.
+        Constraints: Markdown only, no raw diff snippets.
```
camilamacedo86, line 120: I think has more than we might can do with.
Note:
- All that is generated via controller-gen should be consider as the valid one
- The k8s deps bumb we should consider the higher versions
- In case of conflicts we can ask for run make target commands

docs/book/src/reference/ai_integration.md
```diff
@@ -0,0 +1,194 @@
...
+        cat update.patch >> prompt.txt
+
+    - name: Generate summary with GitHub Models # Room for parameters tuning
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        set -o pipefail
+        jq -Rs '
+        {
+          model: "openai/gpt-4o",
```
camilamacedo86, line 133: That is cool so if we use this model we do not need any extra configuration permission and we will have the summary

vitorfloriano: Yeah, now that you commented, we can actually inform users that it is possible to pick a different model. GitHub Models has some different models, for different purposes, with different token limits. I'll add something about that as well.

docs/book/src/reference/ai_integration.md
vitorfloriano: There's so much room for improvement here. Prompt engineering is really a thing. That's why I added a comment about that. 

Users can go crazy in their prompts, experiment, enhance, optmize...

But I guess this one gave me an okay result, considering we are just asking for a summary of the patch.

camilamacedo86:
Hi @vitorfloriano 

We implemented the option in the AutoUpdate Plugin, so can we close this one? 
I will move forward and get this one closed but please feel free to re-open if you think that it see fit.

PR #5039, opened 2025-08-23
🐛 (alpha update): ensure that by default is used merge.conflictStyle=merge to facilitate the review

camilamacedo86:
Switch default Git config to use `merge.conflictStyle=merge`, so conflicts render as:

```diff
<<<<<<< HEAD
foo
=======
bar
>>>>>>> branch
```

Note after we added the configs this configuration was lost.


camilamacedo86:
@vitorfloriano

camilamacedo86:
@vitorfloriano, since this one is the default used by us, and how we doc the result, and it is tiny, I will get this one merged to facilitate the progress, but feel free to raise your concerns; we can always change something in a follow-up—just reducing the burn on the reviews for we be able to be focus where we must need.

PR #5041, opened 2025-08-23
📖 (docs): Reference to Book FAQ entry with solution to too long annotations error

### LINKED ISSUE #5012: (book) Add note in the CronJob tutorial to reference FAQ for common issues
### Describe the feature you want, and what the motivation is

In the current [kubebuilder book](https://book.kubebuilder.io/), the [Running and deploying the controller](https://github.com/kubernetes-sigs/kubebuilder/blob/book-v4/docs/book/src/cronjob-tutorial/running.md) section of the CronJob tutorial is a frequent point where new readers can encounter the `metadata.annotations: Too long` error.

I have recently gone through the tutorial myself and could confirm the issue is still happening. Additionally [this comment on #2556](https://github.com/kubernetes-sigs/kubebuilder/issues/2556#issuecomment-2781476018) and other issues, like #3903, seem to indicate that no permanent fix has been provided yet.

While the FAQ section already includes an entry addressing this issue, there is no mention of it until that point in the book itself.

I propose adding a short note to [this section](https://github.com/kubernetes-sigs/kubebuilder/blob/book-v4/docs/book/src/cronjob-tutorial/running.md?plain=1#L18) explicitly pointing readers to the FAQ in case they encounter any issues. This will make troubleshooting easier for readers.

The following is an example of what such a note could look like:

```
<aside class="note">

<h1>Troubleshooting issues</h1>

If you encounter any errors when running the commands above, especially related to building and applying the CustomResourceDefinitions due to too long annotations, please refer to the [FAQ section](/faq.md).

</aside>
```

### What are your use cases? Why should we do this?

* New readers following the CronJob tutorial often face this error, as reported in many GitHub issues (see related issues below).
* Although the solution is already documented in the FAQ, it is not obvious to the reader at the time they encounter the issue, leading to repeated reports and confusion.
* Directly linking to the relevant FAQ entry from the problematic section would improve the onboarding experience, reduce repetitive questions, and make the documentation more self-contained.

### Does it require a particular Kubernetes version?

No, this request is documentation-only and does not depend on any specific Kubernetes version.

### Is there currently another issue associated with this?

Related issues: #2556, #3903, #3903, #4018
 

### Extra Labels

_No response_

pessolato:
Added a note to the book's CronJob tutorial with an explicit reference to the FAQ sessions with the solution to a common `metadata.annotations: Too long` error.

Resolves #5012

### DIFF
--- a/docs/book/src/cronjob-tutorial/running.md
+++ b/docs/book/src/cronjob-tutorial/running.md
@@ -16,6 +16,15 @@ manifests using controller-tools, if needed:
 make install
 ```
 
+<aside class="note">
+
+<h1>Too long annotations error</h1>
+
+If you encounter errors when applying the CRDs, due to `metadata.annotations` exceeding the
+262144 bytes limit, please refer to the specific entry in the [FAQ section](/faq#the-error-too-long-must-have-at-most-262144-bytes-is-faced-when-i-run-make-install-to-apply-the-crd-manifests-how-to-solve-it-why-this-error-is-faced).
+
+</aside>
+
 Now that we've installed our CRDs, we can run the controller against our
 cluster.  This will use whatever credentials that we connect to the
 cluster with, so we don't need to worry about RBAC just yet.


docs/book/src/cronjob-tutorial/running.md
```diff
@@ -16,6 +16,15 @@ manifests using controller-tools, if needed:
...
 
+<aside class="note">
+
+<h1>Too long annotations error</h1>
+
+If you encounter errors when applying the CRDs, due to `metadata.annotations` exceeding the
+262144 bytes limit, please refer to the specific entry in the [FAQ section](/faq#the-error-too-long-must-have-at-most-262144-bytes-is-faced-when-i-run-make-install-to-apply-the-crd-manifests-how-to-solve-it-why-this-error-is-faced).
+
+</aside>
+
```
camilamacedo86, line 27: I checked the change here: https://deploy-preview-5041--kubebuilder.netlify.app/cronjob-tutorial/running

All fine.

Thank you for contirbute with 🎉

Issue #5020, opened 2025-08-18
Uncomment CertManager install in GitHub action scaffolded to check the solution packaged with Helm Chart


camilamacedo86:
### What do you want to happen?

**Problem**
When scaffolding a project with Helm plugin (https://book.kubebuilder.io/plugins/available/helm-v1-alpha) and enabling webhooks, the generated GitHub Actions workflow (`.github/workflows/test-chart.yaml`) fails on Helm install because cert-manager is not installed by default. Example error:

```
Run helm install my-release ./dist/chart --create-namespace --namespace kubebuilder-s6-system
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest:
[resource mapping not found for name: "serving-cert" namespace: "kubebuilder-s6-system" from "":
no matches for kind "Certificate" in version "cert-manager.io/v1"
ensure CRDs are installed first, resource mapping not found for name: "metrics-certs" namespace: "kubebuilder-s6-system" from "":
no matches for kind "Certificate" in version "cert-manager.io/v1"
ensure CRDs are installed first, resource mapping not found for name: "selfsigned-issuer" namespace: "kubebuilder-s6-system" from "":
no matches for kind "Issuer" in version "cert-manager.io/v1"
ensure CRDs are installed first]
```

Reference: https://github.com/kubernetes-sigs/kubebuilder/pull/4985/files

**Expected**
- CI should install cert-manager automatically if the chart has webhooks.
- If no webhooks, cert-manager step should remain commented to keep CI lightweight.

  - **Uncommented** if webhooks present → install cert-manager via Helm, wait for deployments.
  - **Commented** otherwise → keep behavior as today.

**Proposal**
- When we are scaffolding the Helm Chart with the plugin: https://book.kubebuilder.io/plugins/available/helm-v1-alpha check if has Webhooks if yes then we should uncomment the code:  https://github.com/kubernetes-sigs/kubebuilder/pull/4985/files
We already have utils for this:
  - Util: https://github.com/kubernetes-sigs/kubebuilder/blob/80c17fa32a4f0db1decb0b1155b4c9973440d642/pkg/plugin/util/util.go#L136-L180
  - Example usage: https://github.com/kubernetes-sigs/kubebuilder/blob/80c17fa32a4f0db1decb0b1155b4c9973440d642/pkg/plugins/common/kustomize/v2/scaffolds/webhook.go#L331-L384
[...]

PR #5043, opened 2025-08-24
🐛 (helm/v1-alpha): When scaffolding a Helm project with webhooks, the generated GitHub Actions workflow now installs and waits for cert-manager. Without webhooks, the cert-manager step remains commented.

n2h9:
This pr implements #5020 

Notes:
- I have added additional comment character `#` in `pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/github/test_chart.go` so that when we uncomment the whole block TODO comment stays commented; 
- I specified the whole uncomment block as is in pkg/plugins/optional/helm/v1alpha/edit.go similar how it is done now in example given in `pkg/plugins/common/kustomize/v2/scaffolds/webhook.go` let me know if it is better to arrange this block to not repeat the code block :blush: 

Here is an example of a project scaffolded with the set of commands provided in the issue (with kubebuilder build from changes in this pr) **with webhook**:

```sh
kubebuilder init

kubebuilder create api --group example.com --version v1 --kind App --controller=true --resource=true

kubebuilder create webhook --group example.com --version v1 --kind App --defaulting

kubebuilder edit --plugins=helm.kubebuilder.io/v1-alpha
```

https://github.com/n2h9/kubebuilder-init-cm 

github workflows (including install cert manager and install helm) are passing :white_check_mark: 
 
<details>
<summary>screenshot</summary>
image
</details>

----

Here is an example of a project scaffolded with the set of commands provided in the issue (with kubebuilder build from changes in this pr) **without webhooks**:
```sh
kubebuilder init

kubebuilder create api --group example.com --version v1 --kind App --controller=true --resource=true

kubebuilder edit --plugins=helm.kubebuilder.io/v1-alpha
```

https://github.com/n2h9/kubebuilder-init-no-cm 

github workflows (including install helm) are passing :white_check_mark: 

<details>
<summary>screenshot</summary>
image
</details>

### DIFF
--- a/.github/workflows/test-helm-samples.yml
+++ b/.github/workflows/test-helm-samples.yml
@@ -88,3 +88,70 @@ jobs:
       - name: Check Presence of ServiceMonitor
         run: |
           kubectl wait --namespace project-v4-with-plugins-system --for=jsonpath='{.kind}'=ServiceMonitor servicemonitor/project-v4-with-plugins-controller-manager-metrics-monitor
+
+  # Test scenario: 
+  # - scaffold project without creating webhooks, 
+  # - deploy helm chart without installing cert manager;
+  # - check that deployment has been deployed;
+  #
+  # Command to use to scaffold project without creating webhooks and so no need to install cert manager:
+  # - kubebuilder init
+  # - kubebuilder create api --group example.com --version v1 --kind App --controller=true --resource=true
+  # - kubebuilder edit --plugins=helm.kubebuilder.io/v1-alpha
+  test-helm-no-webhooks:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v5
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Install the latest version of kind
+        run: |
+          curl -Lo ./kind https://kind.sigs.k8s.io/dl/latest/kind-linux-amd64
+          chmod +x ./kind
+          sudo mv ./kind /usr/local/bin/kind
+
+      - name: Create kind cluster
+        run: kind create cluster
+
+      - name: Install Helm
+        run: |
+          curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
+
+      - name: Install kubebuilder binary
+        run: make install
+
+      - name: Create test directory
+        run: mkdir -p test-helm-no-webhooks
+
+      - name: Scaffold project with kubebuilder commands
+        working-directory: test-helm-no-webhooks
+        run: |
+          go mod init test-helm-no-webhooks
+          kubebuilder init
+          kubebuilder create api --group example.com --version v1 --kind App --controller=true --resource=true
+          kubebuilder edit --plugins=helm.kubebuilder.io/v1-alpha
+
+      - name: Build and load Docker image
+        working-directory: test-helm-no-webhooks
+        run: |
+          make docker-build IMG=test-helm-no-webhooks:v0.1.0
+          kind load docker-image test-helm-no-webhooks:v0.1.0
+
+      - name: Lint Helm chart
+        working-directory: test-helm-no-webhooks
+        run: helm lint ./dist/chart
+
+      - name: Deploy Helm chart without cert-manager
+        working-directory: test-helm-no-webhooks
+        run: helm install my-release ./dist/chart --create-namespace --namespace test-helm-no-webhooks-system
+
+      - name: Verify deployment is working
+        working-directory: test-helm-no-webhooks
+        run: |
+          helm status my-release --namespace test-helm-no-webhooks-system
--- a/pkg/plugins/optional/helm/v1alpha/edit.go
+++ b/pkg/plugins/optional/helm/v1alpha/edit.go
@@ -18,12 +18,16 @@ package v1alpha
 
 import (
 	"fmt"
+	log "log/slog"
+	"os"
+	"path/filepath"
 
 	"github.com/spf13/pflag"
 
 	"sigs.k8s.io/kubebuilder/v4/pkg/config"
 	"sigs.k8s.io/kubebuilder/v4/pkg/machinery"
 	"sigs.k8s.io/kubebuilder/v4/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/v4/pkg/plugin/util"
 	"sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/helm/v1alpha/scaffolds"
 )
 
@@ -84,4 +88,56 @@ func (p *editSubcommand) Scaffold(fs machinery.Filesystem) error {
 
 	// Track the resources following a declarative approach
 	return insertPluginMetaToConfig(p.config, pluginConfig{})
+}
+
+// PostScaffold automatically uncomments cert-manager installation when webhooks are present
+func (p *editSubcommand) PostScaffold() error {
+	hasWebhooks := hasWebhooksWith(p.config)
+
+	if hasWebhooks {
+		workflowFile := filepath.Join(".github", "workflows", "test-chart.yml")
+		if _, err := os.Stat(workflowFile); err != nil {
+			log.Info(
+				"Workflow file not found, unable to uncomment cert-manager installation",
+				"error", err,
+				"file", workflowFile,
+			)
+			return nil
+		}
+		//nolint:lll
+		target := `
+#      - name: Install cert-manager via Helm
+#        run: |
+#          helm repo add jetstack https://charts.jetstack.io
+#          helm repo update
+#          helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --set installCRDs=true
+#
+#      - name: Wait for cert-manager to be ready
+#        run: |
+#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager
+#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-cainjector
+#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-webhook`
+		if err := util.UncommentCode(workflowFile, target, "#"); err != nil {
+			hasUncommented, errCheck := util.HasFileContentWith(workflowFile, "- name: Install cert-manager via Helm")
+			if !hasUncommented || errCheck != nil {
+				log.Warn("Failed to uncomment cert-manager installation in workflow file", "error", err, "file", workflowFile)
+			}
+		}
+	}
+	return nil
+}
+
+func hasWebhooksWith(c config.Config) bool {
+	resources, err := c.GetResources()
+	if err != nil {
+		return false
+	}
+
+	for _, res := range resources {
+		if res.HasDefaultingWebhook() || res.HasValidationWebhook() || res.HasConversionWebhook() {
+			return true
+		}
+	}
+
+	return false
 }
--- a/docs/book/src/cronjob-tutorial/testdata/project/.github/workflows/test-chart.yml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/.github/workflows/test-chart.yml
@@ -47,17 +47,17 @@ jobs:
           helm lint ./dist/chart
 
 # TODO: Uncomment if cert-manager is enabled
-#      - name: Install cert-manager via Helm
-#        run: |
-#          helm repo add jetstack https://charts.jetstack.io
-#          helm repo update
-#          helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --set installCRDs=true
-#
-#      - name: Wait for cert-manager to be ready
-#        run: |
-#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager
-#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-cainjector
-#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-webhook
+      - name: Install cert-manager via Helm
+        run: |
+          helm repo add jetstack https://charts.jetstack.io
+          helm repo update
+          helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --set installCRDs=true
+
+      - name: Wait for cert-manager to be ready
+        run: |
+          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager
+          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-cainjector
+          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-webhook
 
 # TODO: Uncomment if Prometheus is enabled
 #      - name: Install Prometheus Operator CRDs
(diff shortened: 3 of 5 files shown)


pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/github/test_chart.go
```diff
@@ -92,7 +92,7 @@ jobs:
         run: |
           helm lint ./dist/chart
 
-# TODO: Uncomment if cert-manager is enabled
+## TODO: Uncomment if cert-manager is enabled
```
camilamacedo86: Instead to add an extra # we can just do not add the line in : https://github.com/kubernetes-sigs/kubebuilder/pull/5043/files#diff-db0ead411f36852dfe4fa2610bcd71763c150597503b572a896817f8cd526963R99

pkg/plugins/optional/helm/v1alpha/edit.go
```diff
@@ -18,12 +18,15 @@ package v1alpha
 
 import (
```
camilamacedo86, line 19: Any change in the scaffolds requires we run `make install` and `make generate` could you please do it?
Also, please ensure that we have the commits squashed , we should have 1 commit, 1 fix/rfe per PR so that we can ensure a good git history and release notes.

pkg/plugins/optional/helm/v1alpha/edit.go
```diff
@@ -18,12 +18,15 @@ package v1alpha
...
 	"fmt"
+	"os"
+	"path/filepath"
 
 	"github.com/spf13/pflag"
 
 	"sigs.k8s.io/kubebuilder/v4/pkg/config"
 	"sigs.k8s.io/kubebuilder/v4/pkg/machinery"
 	"sigs.k8s.io/kubebuilder/v4/pkg/plugin"
+	"sigs.k8s.io/kubebuilder/v4/pkg/plugin/util"
```
camilamacedo86, line 30: Also, see that we have : https://github.com/kubernetes-sigs/kubebuilder/blob/master/.github/workflows/test-helm-samples.yml

This workflow calls the Helm tests for the samples under testdata.

We should create a new job/test for us to validate this scenario
and add a comment above describing the scenario. Could you please add it there? 
So that we do not risk adding regressions to it.

camilamacedo86 (changes requested):
@n2h9 

Thank you a lot for checking this one.
I added a few comments on what is missing for us to get this one merged
But that is an amazing quick start 🥇 🚀 

Thank you for your collab. 

PS.The  PR titles are used in the release notes, so I need to add the right emoji and information about the change that matches the user POV. If you do not mind, I will change this one so that you will know how to do the next times.

n2h9:
> @n2h9
> 
> Thank you a lot for checking this one. I added a few comments on what is missing for us to get this one merged But that is an amazing quick start 🥇 🚀
> 
> Thank you for your collab.
> 
> PS.The PR titles are used in the release notes, so I need to add the right emoji and information about the change that matches the user POV. If you do not mind, I will change this one so that you will know how to do the next times.
> 
> Could you please check it out?

Hey hey, hello @camilamacedo86 :wave: !

Thank you so much for an instant review and  feedback :hugs: .

Title looks much nicer now with this cute emoji :innocent: .

I am going over the comments and will add required updates shortly :blush: .

Thank you :bowing_woman: .

pkg/plugins/optional/helm/v1alpha/edit.go
camilamacedo86: The scenario would be:

```
kubebuilder init

kubebuilder create api --group example.com --version v1 --kind App --controller=true --resource=true

kubebuilder edit --plugins=helm.kubebuilder.io/v1-alpha
```

And then we call the helm check as we do for the other one that we have.
In this way, we will know that the GH Action will always work.

n2h9: Hey hey @camilamacedo86 :wave:  :innocent: 
Sorry, I think I didn get fully how to arrange this test properly :cry: 
So would like to confirm with you first :innocent: .

So I am looking into [test-helm-sample.yaml](https://github.com/kubernetes-sigs/kubebuilder/blob/master/.github/workflows/test-helm-samples.yml)
I understand that the idea is to add a new job into the same workflow :white_check_mark: 

Also this particular job "helm-test-project-v4-with-plugins" it has functionality close to scaffolded  test_chart.yaml, f.e. in [testdata/](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-with-plugins/.github/workflows/test-chart.yml). 

The proposed scenario is 
```sh
kubebuilder init

kubebuilder create api --group example.com --version v1 --kind App --controller=true --resource=true

kubebuilder edit --plugins=helm.kubebuilder.io/v1-alpha
```

so this is a scenario without webhooks (or should we have both :thinking: ). 

And the test will be slightly different then ""helm-test-project-v4-with-plugins"", right? So instead of deploying kind and chart, we build kubebuilder, create tmp folder, scaffold project in it and then check that generated test_chart.yaml has cert manager creation section still commented out (by doing some text comparision) ?

Or do we want to execute the  content from generated .github/workflow/test_chart.yml ?

Or smth else? 

Sorry for long text, will appreciate further direction on this :innocent: :hugs: .

pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/github/test_chart.go
n2h9: Updated thank you :hugs: :white_check_mark:

pkg/plugins/optional/helm/v1alpha/edit.go
n2h9: Updated :white_check_mark:  thank you :hugs: 

Also for this I added the condition to check if the current block was not uncommented before :slightly_smiling_face: this is necessary as during `make generate` kubebuilder edit is called twice and on the second call it fails with error that nothing to uncomment. 
I added a comment with description about it, but maybe it is too many words, let me know if it is better to remove it :blush: .


> testdata/project-v4-with-plugins/.github/workflows/test-chart.yml

makes sense, I will do a squash in the end to prepare pr as a single commit, thank you :heart:

pkg/plugins/optional/helm/v1alpha/edit.go
```diff
@@ -85,3 +89,60 @@ func (p *editSubcommand) Scaffold(fs machinery.Filesystem) error {
...
+			}
+
+			// If cert-manager installation step is already uncommented, skip
+			if strings.Contains(string(content), "- name: Install cert-manager via Helm") &&
+				!strings.Contains(string(content), "#      - name: Install cert-manager via Helm") {
+				return nil
+			}
+
+			target := `
+#      - name: Install cert-manager via Helm
```
camilamacedo86, line 109: The target I think here must `#      - name: Install Prometheus Operator CRDs

Try to copy adn paste the scring under here so you have the right spaces and etc

.github/workflows/test-helm-samples.yml
```diff
@@ -88,3 +88,72 @@ jobs:
       - name: Check Presence of ServiceMonitor
```
camilamacedo86, line 88: We need to have only 1 commit, could you please squash for we are able to move forward?

.github/workflows/test-helm-samples.yml
```diff
@@ -88,3 +88,72 @@ jobs:
...
+
+      - name: Create kind cluster
+        run: kind create cluster
+
+      - name: Install Helm
+        run: |
+          curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
+
+      - name: Build kubebuilder binary
+        run: make build
```
camilamacedo86: Hre could be make install

.github/workflows/test-helm-samples.yml
```diff
@@ -88,3 +88,72 @@ jobs:
...
+      - name: Create test directory
+        run: mkdir -p test-helm-no-webhooks
+
+      - name: Scaffold project with kubebuilder commands
+        working-directory: test-helm-no-webhooks
+        run: |
+          go mod init test-helm-no-webhooks
+          ../bin/kubebuilder init
+          ../bin/kubebuilder create api --group example.com --version v1 --kind App --controller=true --resource=true
+          ../bin/kubebuilder edit --plugins=helm.kubebuilder.io/v1-alpha
```
camilamacedo86: that here just kubebuilder without path

pkg/plugins/optional/helm/v1alpha/edit.go
n2h9: I have added here the job which scaffolds the projecy without webhooks and then deploys helm chart, similar to what is done in generated test_chart.yml :white_check_mark: 

The workflow is passing on my fork: https://github.com/n2h9/fork-kubernetes-sigs-kubebuilder-debug-00/actions/runs/17191997648/job/48768859241  

<details>
<summary>screenshot</summary>
image

</details>

.github/workflows/test-helm-samples.yml
n2h9: Updated to single commit :hugs:  :white_check_mark:

pkg/plugins/optional/helm/v1alpha/edit.go
n2h9: I think it should be `#      - name: Install cert-manager via Helm` cause we are uncommenting cert-manager rather than prometheus, right :blush:  ?

pkg/plugins/optional/helm/v1alpha/edit.go
```diff
@@ -85,3 +89,63 @@ func (p *editSubcommand) Scaffold(fs machinery.Filesystem) error {
...
+#          helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --set installCRDs=true
+#
+#      - name: Wait for cert-manager to be ready
+#        run: |
+#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager
+#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-cainjector
+#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-webhook`
+
+			if err := util.UncommentCode(workflowFile, target, "#"); err != nil {
+				return fmt.Errorf("failed to uncomment cert-manager installation in workflow file: %w", err)
```
camilamacedo86: Just a small detail. If we fail here, we should WARN, not return an error, so that if someone has already changed the scaffold, it is all fine.

pkg/plugins/optional/helm/v1alpha/edit.go
```diff
@@ -85,3 +89,63 @@ func (p *editSubcommand) Scaffold(fs machinery.Filesystem) error {
...
+			content, err := os.ReadFile(workflowFile)
+			if err != nil {
+				return fmt.Errorf("failed to read workflow file %q: %w", workflowFile, err)
+			}
+
+			// If cert-manager installation step is already uncommented, skip
+			if strings.Contains(string(content), "- name: Install cert-manager via Helm") &&
+				!strings.Contains(string(content), "#      - name: Install cert-manager via Helm") {
+				return nil
+			}
```
camilamacedo86: we should not add this the uncomment method will find in the path already

pkg/plugins/optional/helm/v1alpha/edit.go
camilamacedo86: Also, before call uncommend we should check if it is not already uncommented, see the examples shared.

n2h9: makes sense, then we do no need to check if the code has been already uncommented :slightly_smiling_face:

pkg/plugins/optional/helm/v1alpha/edit.go
n2h9: right, with log.Warn we do not need this, updated :white_check_mark:  :hugs:  thank you :bowing_woman:

pkg/plugins/optional/helm/v1alpha/edit.go
n2h9: updated, thank you :white_check_mark:  :hugs:

pkg/plugins/optional/helm/v1alpha/edit.go
```diff
@@ -85,3 +89,47 @@ func (p *editSubcommand) Scaffold(fs machinery.Filesystem) error {
...
+#          helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --set installCRDs=true
+#
+#      - name: Wait for cert-manager to be ready
+#        run: |
+#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager
+#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-cainjector
+#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-webhook`
+
+			if err := util.UncommentCode(workflowFile, target, "#"); err != nil {
+				log.Warn("Failed to uncomment cert-manager installation in workflow file", "error", err, "file", workflowFile)
```
camilamacedo86, line 123: We should ONLY warn if the code is not uncomment already. 
Example: https://github.com/kubernetes-sigs/kubebuilder/blob/61455d0ecc96e27a76912d71599c506ec4414d23/pkg/plugins/common/kustomize/v2/scaffolds/api.go#L93-L100

So, I would recommend we other encapsulate the code in a const.

camilamacedo86 (approved):
/lgtm
/approved

camilamacedo86:
/lgtm cancel
/approve cancel

we need to address first https://github.com/kubernetes-sigs/kubebuilder/pull/5043#discussion_r2299030576 so that we can get it merged

pkg/plugins/optional/helm/v1alpha/edit.go
n2h9: Hey hey, @camilamacedo86 hello :wave: :innocent: ! 

I have added a check if the uncomented code is present and only log Error if the code failed to uncomment and uncommented version is not present.  I do check by the presence of the first line of the step.

I thought of encapsulate the cert manager step into a const inside test_chart.go to avoid duplication. But test_chart.go is inside internal package and I can not export constant from there into edit.go. 

Thank you :hugs:

pkg/plugins/optional/helm/v1alpha/edit.go
```diff
@@ -85,3 +90,50 @@ func (p *editSubcommand) Scaffold(fs machinery.Filesystem) error {
...
+#
+#      - name: Wait for cert-manager to be ready
+#        run: |
+#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager
+#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-cainjector
+#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-webhook`
+		if err := util.UncommentCode(workflowFile, target, "#"); err != nil {
+			hasUncommented, errCheck := pluginutil.HasFileContentWith(workflowFile, "- name: Install cert-manager via Helm")
+			if !hasUncommented || errCheck != nil {
+				log.Error("Failed to uncomment cert-manager installation in workflow file", "error", err, "file", workflowFile)
```
camilamacedo86: we canno fail :-( 
If the person remove the GH we should still scaffolding the helm charts

n2h9: I think it is not failing, it is jus logging with Error level :thinking: 
   

No worries, I have updated back to warn :innocent: :white_check_mark:

camilamacedo86 (approved):
/lgtm
/approve

camilamacedo86:
Hi @n2h9 

It is not passing in the linter checks see: https://github.com/kubernetes-sigs/kubebuilder/actions/runs/17247973408/job/48943116786?pr=5043

Can you run `make lint-fix` locally?
And then ensure that size of the lines to fix all?

n2h9:
> Hi @n2h9
> 
> It is not passing in the linter checks see: https://github.com/kubernetes-sigs/kubebuilder/actions/runs/17247973408/job/48943116786?pr=5043
> 
> Can you run `make lint-fix` locally? And then ensure that size of the lines to fix all?


Done :hugs: , I have fixed the lint issues and marked  `target` variable  with `//nolint:lll` cause it contains a part of a template and it is not possible to split it properly :cry: similar to test_chart.go [line 46](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/optional/helm/v1alpha/scaffolds/internal/templates/github/test_chart.go#L46)

thank you :bowing_woman:

PR #5044, opened 2025-08-24
🌱  (fix): Fix inconsistent logging format  ( Follow up: #4968 )

camilamacedo86:
(fix): Fix inconsistent logging format 

Follow up: https://github.com/kubernetes-sigs/kubebuilder/pull/4968

Testing alpha update and generate revealed logs were not in the right format.

Changes:
- Initialize consistent logging handler in alpha commands to fix level=info msg= format
- Replace log.Fatalf with slog.Error + os.Exit(1) for proper error handling
- Standardize log imports to use log "log/slog" across all files
- Remove redundant command info from RunCmd logging output
- Ensure all kubebuilder processes use the same colored INFO/WARN format

Before: level=warning msg="Using current working directory..."
After:  WARN Using current working directory...

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -17,7 +17,6 @@ limitations under the License.
 package cmd
 
 import (
-	"log"
 	"log/slog"
 	"os"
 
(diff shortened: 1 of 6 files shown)


camilamacedo86:
Moving forward since it is a small fix

PR #5048, opened 2025-08-24
✨ (go/v4): expanded .dockerignore to exclude git files, build artifacts , IDE junk, go cache, and generated manifests for cleaner, faster docker builds

camilamacedo86:


docs/book/src/cronjob-tutorial/testdata/project/.dockerignore
```diff
@@ -1,3 +1,11 @@
 # More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
-# Ignore build and test binaries.
-bin/
+# Ignore everything
```
aman4433: Maybe adjust the comment here from `# Ignore everything` to something like `# Ignore everything by default, re-include only needed files to` better reflect what the file is doing.

nerdeveloper:
LGTM

PR #5056, opened 2025-08-26
🐛 (go/v4) fix simple typos in Dockerfile comments

lpmi-13:
Fix a couple typos in the Dockerfile comments.

Noticed this while building a new CRD (very smooth flow by the way, really appreciate the effort in this area!)


camilamacedo86 (commented):
Could you please run `make install` and `make generate` so that we get all samples under testdata and docs updated within? Otherwise, we cannot move forward.

Also, could you please ensure that when you push it you squash the commits so that we have a nice git history and release. Thank you a lot for your contribution !!!

camilamacedo86 (approved):
/lgtm

PR #5058, opened 2025-08-27
WIP Add helm/v2-alpha and deprecated helm/v1-alpha

camilamacedo86:


camilamacedo86:
Hey,

This update will address all your requests.

I would kindly ask your help to validate the new helm/v2-alpha using your manifests and scenarios.

We’ve made a lot of changes:

All samples under testdata and docs have been updated.

We now cover much more with end-to-end tests, ensuring stronger guarantees.

Thank you so much for your collaboration in making Kubebuilder better! 🙏

Cheers,

c/c @tchinmai7 @bavarianbidi @mkarlheim @abhijith-darshan @klaudworks @pabhi18

pkg/plugins/optional/helm/v2alpha/edit.go
```diff
@@ -0,0 +1,217 @@
...
+    ├── webhook/
+    └── ...
+`, cliMeta.CommandName, plugin.KeyFor(Plugin{}))
+}
+
+func (p *editSubcommand) BindFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&p.force, "force", false, "if true, regenerates all the files")
+	fs.StringVar(&p.manifestsFile, "manifests", DefaultManifestsFile,
+		"path to the YAML file containing Kubernetes manifests from kustomize output")
+	fs.StringVar(&p.outputDir, "output", DefaultOutputDir, "output directory for the generated Helm chart")
```
bavarianbidi: shouldn't the flag called `output-dir`?
as you suggested in my initial PR here: https://github.com/kubernetes-sigs/kubebuilder/pull/4891#discussion_r2171168191

bavarianbidi:
returned from PTO this week, will do a deeper review, starting next week, if wanted.

pkg/plugins/optional/helm/v2alpha/edit.go
camilamacedo86: fixed.

mkarlheim:
@camilamacedo86 many thanks for the PR. I've tested this version and it already looks very promising to my prior issues.

I just found some things to improve:
1. The CRDs should be placed in the helm v3 suggested way: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/. Just move the templates/crd folder to the root of the chart and name it "crds" and remove the flow control `crd.enable`. 

This a special directory introduced especially for crds in helm. By doing this, helm will install these crds if not exists by default and you can skip this with `--skip-diff`

2. Missing labels in manager.yaml
The deployment is missing the `include "chart.labels"` in labels section and the `include "chart.selectorLabels"` in matchLabels section. The podTemplate is also missing `"chart.labels"`.

3. Missing labels in RBAC ClusterRoles (editor, viewer, admin);
In the previous version there was this label section in editor-, viewer-, and admin-roles:
`labels:
    {{- include "chart.labels" . | nindent 4 }}`
This has now been replaced by fixed labels:
`labels:
        app.kubernetes.io/name: xxx`
and leads to differences in `helm diff`

I think using `chart.labels` defined in _helpers.tpl would be the preferred way.

4. The following resources are missing labels section completely:
- manager-role.yaml
- metrics-auth-role.yaml
- metrics-auth-rolebinding.yaml
- metrics-reader.yaml

Besides these small findings I think the new plugin version helm/v2-alpha is a huge improvement!

Thank you

camilamacedo86:
Hi @mkarlheim 

The chart. labels were added 👍  to allow adding extra labels.
However, note that these are the labels defined in Kustomize, and we should retain them.  
The goal of Helm charts is to package the solution, so the labels should be identical.  

Regards the CRDs, see:

Although [Helm best practices](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/) recommend placing CRDs under a top-level `crds/` directory, the Kubebuilder Helm plugin intentionally places them under `templates/crd`.**  

This was a conscious design decision aafter lengthydiscussions in [the design proposal](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/helm-chart-autogenerate-plugin.md) and community threads ([Slack discussion link](https://kubernetes.slack.com/archives/CAR30FCJZ/p1752735414123429?thread_ts=1752732217.829249&cid=CAR30FCJZ)).  

The reason: CRDs in `crds/` are **not upgraded by Helm**. Keeping them under `templates/` ensures CRDs are correctly managed and updated together with releases, avoiding drift. While this means you cannot bundle custom resources directly in the chart (a current Helm limitation), it provides a more straightforward and more reliable workflow for projects generated by Kubebuilder.  See; https://github.com/kubernetes-sigs/kubebuilder/pull/3632#discussion_r1463737806

Alternatively, here we could create a chart only for the CRDs, such as a two-chart flow (API first, then operator/app), with a simple readiness/wait check. However, if we change it, it would be better in a follow-up discussion where we can address it. Let's keep the cards under the templates, which is the most adopted approach.

cmd/cmd.go
```diff
@@ -34,6 +34,7 @@ import (
 	autoupdatev1alpha1 "sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/autoupdate/v1alpha"
 	grafanav1alpha1 "sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/grafana/v1alpha"
 	helmv1alpha1 "sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/helm/v1alpha"
+	helmv2alpha1 "sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/helm/v2alpha"
```
bavarianbidi, line 37: WDYT about enforcing this import-pattern via golangci-lint by the importas section (https://golangci-lint.run/docs/linters/configuration/#importas).
With that we get a consistent import section ...

I'm fine if we do this in a dedicated PR (have it on my list and happy to raise a PR after that)

pkg/cli/alpha/internal/generate.go
```diff
@@ -34,6 +34,7 @@ import (
 	autoupdate "sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/autoupdate/v1alpha"
 	"sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/grafana/v1alpha"
 	hemlv1alpha "sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/helm/v1alpha"
+	hemlv2alpha "sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/helm/v2alpha"
```
bavarianbidi, line 37:
```suggestion
	helmv2alpha "sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/helm/v2alpha"
```

+ the line above

enforcing import-names might help here as well (see https://github.com/kubernetes-sigs/kubebuilder/pull/5058#discussion_r2336743373)

mkarlheim:
> Hi @mkarlheim
> 
> The chart. labels were added 👍 to allow adding extra labels. However, note that these are the labels defined in Kustomize, and we should retain them. The goal of Helm charts is to package the solution, so the labels should be identical.
> 
> Regards the CRDs, see:
> 
> Although [Helm best practices](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/) recommend placing CRDs under a top-level `crds/` directory, the Kubebuilder Helm plugin intentionally places them under `templates/crd`.**
> 
> This was a conscious design decision aafter lengthydiscussions in [the design proposal](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/helm-chart-autogenerate-plugin.md) and community threads ([Slack discussion link](https://kubernetes.slack.com/archives/CAR30FCJZ/p1752735414123429?thread_ts=1752732217.829249&cid=CAR30FCJZ)).
> 
> The reason: CRDs in `crds/` are **not upgraded by Helm**. Keeping them under `templates/` ensures CRDs are correctly managed and updated together with releases, avoiding drift. While this means you cannot bundle custom resources directly in the chart (a current Helm limitation), it provides a more straightforward and more reliable workflow for projects generated by Kubebuilder. See; [#3632 (comment)](https://github.com/kubernetes-sigs/kubebuilder/pull/3632#discussion_r1463737806)
> 
> Alternatively, here we could create a chart only for the CRDs, such as a two-chart flow (API first, then operator/app), with a simple readiness/wait check. However, if we change it, it would be better in a follow-up discussion where we can address it. Let's keep the cards under the templates, which is the most adopted approach.

Thanks for clarifying.

tchinmai7:
This looks great!. the new plugin's chart output is close to kustomize. The `managed-by` label is rightly pointing to `helm` now, and only the namespace resource is missing - which is ok, helm charts can be installed with the `--create-namespace` option. Thank you for working on this!

cmd/cmd.go
camilamacedo86: Yes, we can do that for kubebuilder itself, it is fine
If you would like to propose the change in a new PR that is great

pkg/cli/alpha/internal/generate.go
camilamacedo86: Since it is following as it is now, see the v1. We do not need to change it to get this merged; we can work on it in a follow-up if required.

camilamacedo86:
Hey 

@tchinmai7  @bavarianbidi  @calind  @vaidikcode  @mkarlheim  @abhijith-darshan  @klaudworks  @FishyFishPat  @pabhi18  

I would like to ask your help on reviews and I would appreciate LGTMs if you are satisfied with the proposed changes .. Really thank you for your interest on this one and all your collab !!!

It would be great if we could delivery it with the next release

abhijith-darshan:
> Hey
> 
> @tchinmai7 @bavarianbidi @calind @vaidikcode @mkarlheim @abhijith-darshan @klaudworks @FishyFishPat @pabhi18
> 
> I would like to ask your help on reviews and I would appreciate LGTMs if you are satisfied with the proposed changes .. Really thank you for your interest on this one and all your collab !!!
> 
> It would be great if we could delivery it with the next release

Hey @camilamacedo86,

Thank you so much for this amazing implementation 🚀  

Building the binary from source [camilamacedo86:helm-change-manager](https://github.com/camilamacedo86/kubebuilder/tree/helm-change-manager) and using it to generate charts -

- changes to kustomize manifest is reflected on to chart templates (solves the [original issue](https://github.com/kubernetes-sigs/kubebuilder/issues/4625) I opened) 🎉 
- tested with basic controller, admission webhooks and conversion webhook 👍 

However I see one improvement necessary -

The args for controller-manager `container` needs to be templated and available in `values.yaml` (with defaults or empty [ ] )
Otherwise it won't be possible to dynamically include args.

Currently only static values in the manifest is carried over to the `manager.yaml` in `chart/templates`

```yaml
## Helm template manager.yaml
      containers:
      - command:
        - /manager
        args:
          - --leader-elect
          - --leader-lease-duration=15s
          - --leader-renew-timeout=10s
          - --leader-retry-period=2s
        image: controller:latest
        name: manager
```

Instead it can be something like this -

```yaml
## Helm template manager.yaml
        spec:
            containers:
                - args:
                    - --metrics-bind-address=:8443
                    {{- range .Values.controllerManager.args }}
                    - {{ . }}
                    {{- end }}
                    {{- if .Values.certManager.enable }}
                    - --webhook-cert-path=/tmp/k8s-webhook-server/serving-certs
                    {{- end }}
...
```

**values.yaml**

```yaml
# Configure the controller manager deployment
controllerManager:
  replicas: 1
  args:
    - --leader-elect
    - --leader-lease-duration=15s
    - --leader-renew-timeout=10s
    - --leader-retry-period=2s
```

camilamacedo86:
HI @abhijith-darshan 

The args might be a good thought, but I think we could check this one in a follow-up, wdyt?

abhijith-darshan:
> HI @abhijith-darshan
> 
> The args might be a good thought, but I think we could check this one in a follow-up, wdyt?

Sure. Once this is in, I can contribute towards enhancements as well, thanks.

For now, LGTM 🚀

pkg/plugins/optional/helm/v1alpha/plugin.go
```diff
@@ -55,5 +55,6 @@ func (p Plugin) GetEditSubcommand() plugin.EditSubcommand { return &p.editSubcom
 
 // DeprecationWarning define the deprecation message or return empty when plugin is not deprecated
 func (p Plugin) DeprecationWarning() string {
-	return ""
+	return "helm/v1-alpha plugin is deprecated, use helm/v2-alpha instead which " +
+		"provides dynamic Helm chart generation from kustomize output"
```
pedjak, line 59: nit: could we have a single string here  instead of concatenation?

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	f.TemplateBody = helmHelpersTemplate
+
+	f.IfExistsAction = machinery.SkipFile
+
+	return nil
+}
+
+const helmHelpersTemplate = `{{` + "`" + `{{/*
```
pedjak, line 53: for the future maintenance, would it helpful to move this and similar content into standalone files and use `go:embed` directive to access it?

pkg/plugins/optional/helm/v2alpha/edit.go
```diff
@@ -0,0 +1,216 @@
...
+#        run: |
+#          helm repo add jetstack https://charts.jetstack.io
+#          helm repo update
+#          helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --set crds.enabled=true
+#
+#      - name: Wait for cert-manager to be ready
+#        run: |
+#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager
+#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-cainjector
+#          kubectl wait --namespace cert-manager --for=condition=available --timeout=300s deployment/cert-manager-webhook`
```
pedjak: I think you can get incorporate this waiting into the previous `helm install` step by doing something like:

```
helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --set crds.enabled=true --wait --timeout 300s
#
```

pkg/plugins/optional/helm/v1alpha/plugin.go
camilamacedo86: That would cause the lint check `lll` fail

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
camilamacedo86: All templates in Kubebuilder are currently done this way. I might not have fully understood the reasoning, but if we decide to change the approach, it would probably make sense to update all places consistently. That said, we’re always open to ideas—feel free to open a PR to propose the change and demonstrate what it would look like.

pkg/plugins/optional/helm/v2alpha/edit.go
camilamacedo86: It seems a good idea. 🚀

camilamacedo86:
We have enough reviews/lgtm so far 
Also, to allow contributions on top of those changes we need get this one merged
So, I am letting it move forward. 
We have up to the next release to enhance or change what might be required.

manfredlift:
> The args for controller-manager `container` needs to be templated and available in `values.yaml` (with defaults or empty [ ] )
> Otherwise it won't be possible to dynamically include args.

@camilamacedo86, just tried this out. It seems like also the values for `image` and `resources` in manager.yaml are hardcoded (and come from dist/install.yaml) and the ones in values.yaml are not used?

camilamacedo86:
@manfredlift and @abhijith-darshan 

Both scenarios reported are tracked for helm/v2

- https://github.com/kubernetes-sigs/kubebuilder/issues/5114
- https://github.com/kubernetes-sigs/kubebuilder/issues/5113

atanasenko:
@camilamacedo86 This is so much better than v1.
We do have an additional Deployment added to the stack that is basically a web ui accompanying the the operator. However, the helm v2 plugin seems to only support a single Deployment and only the last one encountered, silently ignoring all the rest.

camilamacedo86:
Hi @atanasenko 

Thank you for the report. Please create an issue with the scenario so that we can try to address it.
Could you please add the code and how we reproduce it?

Issue #4552, opened 2025-02-10
Update Demo in README with an Asciinema Recording


camilamacedo86:
### What do you want to happen?

The current demo in the README is outdated and should be replaced with a dynamic and maintainable alternative. Using [Asciinema](https://asciinema.org/), we can create a script and a Makefile target to generate an updated demo, making it easier to refresh and maintain.

Therefore,  we need to create a script to record a demo using Asciinema.

In this way, we mainly need:

A script that will
- Start recording a new session.
- Run the required commands to demonstrate Kubebuilder functionality.
- Stop the recording and save the cast file.
- Optionally upload to Asciinema.org for easy sharing.
Then:
- Add a Makefile target to automate the recording process. (i.e demo-generate)
- Replace the current demo section in README with an embedded Asciinema player. ([![asciicast](https://asciinema.org/a/<cast_id>.svg)](https://asciinema.org/a/<cast_id>))

chethanm99:
Hi @wazery are you currently working on this issue?

Kuzuri247:
@chethanm99 I am working on this issue

Kavinjsir:
@camilamacedo86 I like the idea 👍🏼 how would you like we making a github action as well to automate and track the process?

camilamacedo86:
Hi @Kavinjsir,

Good point — I totally forgot we already have a script for this. (https://github.com/kubernetes-sigs/kubebuilder/tree/v4.5.1/scripts/demo). Thank you 👍 

The idea here would be to add a Makefile target so we can quickly run it.
Also, we should ensure we're using Asciinema, and that the demo video is updated accordingly.

So, the actual change here, IMHO, would just be:
- Create a new Makefile target (maybe update-demo)
- That target should call the existing script to generate the video, convert it, and take care of all the steps needed to update the README with the latest demo/tooling version
- Ensure that all is working fine as it should be

When you mentioned adding a GitHub Action, I initially thought about including it in make generate.
However, thinking it through, I don’t think that’s the best idea — it would slow things down unnecessarily, especially since we don’t need to update the demo that often.

Let me know what you think!

camilamacedo86:
Hey @wazery @jphamgithub @Kuzuri247,

This task was originally assigned to you, but since there haven’t been any updates for a while, I’m unassigning it so others can pick it up. If you’re still actively working on it, please let us know.

Thanks!

wazery:
[...]

PR #5068, opened 2025-09-01
(:sparkles:) feat: enhance the asciinema demo creation script

wazery:
This is a PR to address the issue #4552 which enhances the current script that creates the asciinema demo recording.


scripts/demo/run.sh
```diff
@@ -13,31 +13,72 @@
...
+mkdir kubebuilder-demo-project
+cd kubebuilder-demo-project
+
 clear
 . $(dirname ${BASH_SOURCE})/util.sh
 
-desc "Initialize Go modules"
-run "go mod init demo.kubebuilder.io"
+desc "Check if Kubebuilder is installed"
+run "kubebuilder version"
```
wazery: Not really sure if we need to start with it tbh.

scripts/demo/setup-kind.sh
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env bash
```
wazery: I checked the Makefile if we have some targets that just creates the needed kind cluster for testing, but I saw it's tightly coupled in the test scripts, so thought of introducing this minimal script just for the demo.

I feel my reasoning is not good here, and it would be nicer if I just decoupled the kind cluster creation from the test scripts, yet it's outside the scope of this PR 🤔 .. just thinking outloud 😅

Makefile
```diff
@@ -108,6 +108,49 @@ generate-charts: build ## Re-generate the helm chart testdata and docs samples
 	(cd docs/book/src/cronjob-tutorial/testdata/project && ../../../../../../bin/kubebuilder edit --plugins=helm/v1-alpha)
 	(cd docs/book/src/multiversion-tutorial/testdata/project && ../../../../../../bin/kubebuilder edit --plugins=helm/v1-alpha)
 
+.PHONY: update-demo
```
camilamacedo86, line 112: Could it not be simplified?
It seems too much 

I think I would have expected it more like it is done in the operator-controller project ,for example, see: https://github.com/operator-framework/operator-controller/blob/main/Makefile#L492-L499

wazery: Thanks a lot @camilamacedo86 the comment is addressed, and the Makefile target is now slim as in the operator controller

scripts/demo/run.sh
```diff
@@ -13,31 +13,72 @@
...
-desc "Let's initialize the project"
-run "kubebuilder init --domain tutorial.kubebuilder.io"
+desc "Let's explore the generated files structure"
+run "tree api/ internal/controller/"
 clear
 
-desc "Examine scaffolded files..."
-run "tree ."
+desc "Look at the API definition - notice the validation markers"
+run "cat api/v1/guestbook_types.go"
```
camilamacedo86: Why are we changing the example? 
I think we should have the same example.

scripts/demo/generate-demo.sh
```diff
@@ -0,0 +1,150 @@
...
+        --out="$svg_file" \
+        --window \
+        --width=120 \
+        --height=30
+    
+    log "Demo updated! New file: docs/gif/kb-demo.${version}.svg"
+    return 0
+}
+
+update_readme() {
```
camilamacedo86, line 159: The script should allow us update other videos such as the example shared so that we could add other demos

scripts/demo/README.md
```diff
@@ -1,33 +1,120 @@
...
+  ```sh
+  # macOS
+  brew install asciinema
+  # Ubuntu/Debian
+  sudo apt-get install asciinema
+  ```
+
+- `svg-term`: For converting recordings to SVG
+  ```sh
+  npm install -g svg-term-cli
```
camilamacedo86: I think we should only add pre-requirements, not the commands to install those, since they change according to the SO.

scripts/demo/generate-demo.sh
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env bash
+
+# Copyright 2023 The Kubernetes Authors.
```
camilamacedo86: 2025?

scripts/demo/generate-demo.sh
wazery: addressed this, thanks for the pointer @camilamacedo86

wazery:
Here is an example recording:


kb-demo v4 7 0

scripts/demo/README.md
```diff
@@ -1,33 +1,116 @@
...
-asciinema rec
-<path-to-KB-repo>/scripts/demo/run.sh
-
-# After each step, press <Enter> to proceed to the next step
+# Start recording
+asciinema rec --title "Kubebuilder Demo" kb-demo.cast
 
-<CTRL-C> to terminate the script
-<CTRL-D> to terminate the asciinema recording
-<CTRL-C> to save the recording locally
```
camilamacedo86, line 27: I think it was used. 
Also, see that the record has errors.

scripts/demo/setup-kind.sh
```diff
@@ -0,0 +1,108 @@
...
+kubectl cluster-info --context kind-${CLUSTER_NAME}
+kubectl get nodes --context kind-${CLUSTER_NAME}
+
+echo ""
+echo "Kind cluster is ready for the Kubebuilder demo!"
+echo "Cluster name: ${CLUSTER_NAME}"
+echo "Context: kind-${CLUSTER_NAME}"
+echo ""
+echo "To delete the cluster when done:"
+echo "   kind delete cluster --name ${CLUSTER_NAME}"
```
camilamacedo86: Not sure if we need that. 
We have already a script that setup the kind that is called with out e2e-test-local
Give a look on the Makefile.

scripts/demo/README.md
```diff
@@ -1,33 +1,116 @@
...
+```
+
+## Prerequisites for Recording
+
+- `kind`: For creating local Kubernetes clusters
+- `kubectl`: For interacting with Kubernetes
+- `asciinema`: For recording terminal sessions
+- `svg-term`: For converting recordings to SVG (requires Node.js/npm)
+
+## What the Demo Shows
```
camilamacedo86: I think that is not required.
The README should be focused in how we can use the sh utils instead.
We cannot doc here all possible demos, if it grows will be hard.

scripts/demo/setup-kind.sh
camilamacedo86: Maybe, but it seems we’d be duplicating code.
Since we already handle this in other place, it might increase maintenance costs — maybe we can find a way to centralize it instead?

scripts/demo/setup-kind.sh
```diff
@@ -0,0 +1,108 @@
...
+    nodeRegistration:
+      kubeletExtraArgs:
+        node-labels: "ingress-ready=true"
+  extraPortMappings:
+  - containerPort: 80
+    hostPort: 80
+    protocol: TCP
+  - containerPort: 443
+    hostPort: 443
+    protocol: TCP
```
camilamacedo86: Do we need those PORTs?
See that the kind config that we use is very simple. 
See; https://github.com/kubernetes-sigs/kubebuilder/blob/master/test/e2e/kind-config.yaml

scripts/demo/README.md
wazery: I added them back

scripts/demo/generate-demo.sh
```diff
@@ -0,0 +1,217 @@
...
+    log "Updating README.md with new demo..."
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+        # macOS
+        sed -i '' "s|docs/gif/kb-demo\.v[^)]*\.svg|docs/gif/kb-demo.${version}.svg|g" "${PROJECT_ROOT}/README.md"
+    else
+        # Linux
+        sed -i "s|docs/gif/kb-demo\.v[^)]*\.svg|docs/gif/kb-demo.${version}.svg|g" "${PROJECT_ROOT}/README.md"
+    fi
+    
+    log "README.md updated with new demo file ✓"
```
camilamacedo86, line 177: Is the script really generec and allow us to create demos for any place/doc? 
Or is it creating only for the README

camilamacedo86:
Hi @wazery,

Hope you don’t mind, but based on the feedback in the review (see: https://github.com/kubernetes-sigs/kubebuilder/pull/5281#pullrequestreview-3858764324
), we’ll need to rethink the best way to support these demos with asciinema and redesign how we publish the videos.

Given that, we’re going to close this PR for now.

Thank you so much for all your collaboration on this one — and for your help across other PRs as well.

Issue #5088, opened 2025-09-08
ENVTEST_VERSION and ENVTEST_K8S_VERSION got incorrect values when using replace in go.mod


dmvolod:
### What broke? What's expected?

We use a replacement in go.mod for sigs.k8s.io/controller-runtime and k8s.io/api.
When running a make test in a kubebuilder generating project, the ENVTEST_VERSION and ENVTEST_K8S_VERSION values ​​are taken from the original go versions, rather than being replaced.

### Reproducing this issue

Create basic kubebuilder project. Edit go/mod and replace sigs.k8s.io/controller-runtime and k8s.io/api component versions.

The main problem of this issue, that `go list -m -f "{{ .Version }}" k8s.io/api` command in Makefile doesn't utilize replace info in this manner. Seems to command should be like this

```bash
  go list -m k8s.io/api | rev | cut -d' ' -f1 | rev | awk -F'[v.]' '{printf "1.%d", $$3}
```
Or can be optimized a bit with raw awk code.

### KubeBuilder (CLI) Version

4.6.0

### PROJECT version

3

### Plugin versions

```yaml

```

Shubhamag12:
able to reproduce this;
would like to take this up if accepted
/assign

PR #5096, opened 2025-09-12
✨  (go/v4): Allow `ENV TEST` version to be derived from `replaces`

Shubhamag12:
fixes: https://github.com/kubernetes-sigs/kubebuilder/issues/5088

this fix ensures ENVTEST uses the replaced k8s.io/api version when replace directives are present in go.mod

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/Makefile
+++ b/docs/book/src/cronjob-tutorial/testdata/project/Makefile
@@ -193,12 +193,18 @@ GOLANGCI_LINT = $(LOCALBIN)/golangci-lint
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.7.1
 CONTROLLER_TOOLS_VERSION ?= v0.19.0
-#ENVTEST_VERSION is the version of controller-runtime release branch to fetch the envtest setup script (i.e. release-0.20)
-ENVTEST_VERSION ?= $(shell go list -m -f "{{ .Version }}" sigs.k8s.io/controller-runtime | awk -F'[v.]' '{printf "release-%d.%d", $$2, $$3}')
-#ENVTEST_K8S_VERSION is the version of Kubernetes to use for setting up ENVTEST binaries (i.e. 1.31)
-ENVTEST_K8S_VERSION ?= $(shell go list -m -f "{{ .Version }}" k8s.io/api | awk -F'[v.]' '{printf "1.%d", $$3}')
-GOLANGCI_LINT_VERSION ?= v2.5.0
 
+#ENVTEST_VERSION is the version of controller-runtime release branch to fetch the envtest setup script (i.e. release-0.20)
+ENVTEST_VERSION ?= $(shell v='$(call gomodver,sigs.k8s.io/controller-runtime)'; \
+  [ -n "$$v" ] || { echo "Set ENVTEST_VERSION manually (controller-runtime replace has no tag)" >&2; exit 1; }; \
+  printf '%s\n' "$$v" | sed -E 's/^v?([0-9]+)\.([0-9]+).*/release-\1.\2/')
+
+#ENVTEST_K8S_VERSION is the version of Kubernetes to use for setting up ENVTEST binaries (i.e. 1.31)
+ENVTEST_K8S_VERSION ?= $(shell v='$(call gomodver,k8s.io/api)'; \
+  [ -n "$$v" ] || { echo "Set ENVTEST_K8S_VERSION manually (k8s.io/api replace has no tag)" >&2; exit 1; }; \
+  printf '%s\n' "$$v" | sed -E 's/^v?[0-9]+\.([0-9]+).*/1.\1/')
+
+GOLANGCI_LINT_VERSION ?= v2.5.0
 .PHONY: kustomize
 kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
 $(KUSTOMIZE): $(LOCALBIN)
@@ -241,4 +247,8 @@ GOBIN="$(LOCALBIN)" go install $${package} ;\
 mv "$(LOCALBIN)/$$(basename "$(1)")" "$(1)-$(3)" ;\
 } ;\
 ln -sf "$$(realpath "$(1)-$(3)")" "$(1)"
+endef
+
+define gomodver
+$(shell go list -m -f '{{if .Replace}}{{.Replace.Version}}{{else}}{{.Version}}{{end}}' $(1) 2>/dev/null)
 endef
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -268,12 +268,18 @@ GOLANGCI_LINT = $(LOCALBIN)/golangci-lint
 ## Tool Versions
 KUSTOMIZE_VERSION ?= {{ .KustomizeVersion }}
 CONTROLLER_TOOLS_VERSION ?= {{ .ControllerToolsVersion }}
-#ENVTEST_VERSION is the version of controller-runtime release branch to fetch the envtest setup script (i.e. release-0.20)
-ENVTEST_VERSION ?= $(shell go list -m -f "{{ "{{ .Version }}" }}" sigs.k8s.io/controller-runtime | awk -F'[v.]' '{printf "release-%d.%d", $$2, $$3}')
-#ENVTEST_K8S_VERSION is the version of Kubernetes to use for setting up ENVTEST binaries (i.e. 1.31)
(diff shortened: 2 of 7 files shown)


vitorfloriano (commented):
Thanks for the PR!

I was wondering how this would behave in a developer workflow that involves a local fork? 

For example, if someone is testing against a local clone of `k8s.io/api` with a `go.mod` directive like `replace k8s.io/api v1.28.0 => ../my-api-fork` (suppose the developer needs to test changes against an unreleased version of Kubernetes or debug an issue in an upstream dependency)

What would the `awk '{print $NF}'` command resolve to in that case? I'm not sure it would be a version string.

Perhaps @dmvolod can chime in?

Shubhamag12:
@vitorfloriano That's a good catch, local path replacements would break this approach. I will verify this case and will try to come up with more robust approach. Thanks!

dmvolod:
> Thanks for the PR!
> 
> I was wondering how this would behave in a developer workflow that involves a local fork?
> 
> For example, if someone is testing against a local clone of `k8s.io/api` with a `go.mod` directive like `replace k8s.io/api v1.28.0 => ../my-api-fork` (suppose the developer needs to test changes against an unreleased version of Kubernetes or debug an issue in an upstream dependency)
> 
> What would the `awk '{print $NF}'` command resolve to in that case? I'm not sure it would be a version string.
> 
> Perhaps @dmvolod can chime in?

Yes, @vitorfloriano this is good point.
In case of replacement with path, both current and new approaches are not working as expected.
This should be polished a bit with new inputs.

Shubhamag12:
Hi @vitorfloriano @dmvolod 

I tested the local path replacement case and my current fix breaks:

With `k8s.io/api v0.34.1 => ../api-fork`:
- The current fix outputs `1.0` (tries to parse the path as version)

I can update the fix to handle these cases:
`go list -m k8s.io/api | awk '{if($$NF ~ /^v[0-9]/) print $$NF; else print $$2}'`

This checks if the replacement starts with "v" + digit. If yes, use replacement; otherwise, fall back to original version.

one question here: Should we fall back to the original version in this case? as `go list -m` only gives the declared version

vitorfloriano:
> one question here: Should we fall back to the original version in this case? as `go list -m` only gives the declared version

Honestly, I don't know. It seems to me that a more reliable approach would be to warn the user about not being able to resolve the version and ask them to set the version as env var, but that would require a separate program. 🤔

dmvolod:
I think that if the last field when getting the version is not the version, you should throw an error if possible and tell the user that this variable should be defined manually, like
```
ENVTEST_K8S_VERSION=v1.31.1 make ...
```

Shubhamag12:
@dmvolod Thanks for your suggestion

Here is output with this change: 
```
ENVTEST_K8S_VERSION ?= $(shell go list -m k8s.io/api | \
  awk '{ \
    if ($$NF ~ /^v[0-9]/) { \
      gsub(/^v/, "", $$NF); \
      split($$NF, v, "."); \
      major=v[1]; minor=v[2]; \
      if (major == 0) major=1; \
      printf "%d.%d", major, minor \
    } else { \
      print "Error: Cannot detect version. Set ENVTEST_K8S_VERSION manually." > "/dev/stderr"; \
      exit 1 \
    } \
  }' \

```

```
1. Without any replace
->  make setup-envtest 
Downloading sigs.k8s.io/controller-runtime/tools/setup-envtest@release-0.22
Setting up envtest binaries for Kubernetes version 1.34..

2. With env replace
->  go mod edit -replace=k8s.io/api=k8s.io/api@v0.33.0 
-> make setup-envtest
Setting up envtest binaries for Kubernetes version 1.33...

3. With local fork replace
-> go mod edit -replace=k8s.io/api=../api-fork 
-> go list -m k8s.io/api
k8s.io/api v0.34.0 => ../fake-api-fork
-> make setup-envtest
Error: Cannot detect version. Set ENVTEST_K8S_VERSION manually.

4. ENVTEST_K8S_VERSION=1.31 make setup-envtest
Setting up envtest binaries for Kubernetes version 1.31...
```

Let me know if this looks good!

dmvolod:
> However, I think this approach makes the default scaffold a bit too complex. If we can find a simpler solution, that might work better — but I’m not entirely sure.
> 
> Using a `replace` feels more like a temporary workaround until something changes. In that case, it might be better to just add a comment above it with a `TODO`, as you mentioned, and adjust the ENVTEST version when needed.

I'm totally agree with you, @camilamacedo86.  Replacement with a folder a bit specific case, but regular replacement with version or repository is a common approach and original function call fix.

camilamacedo86:
Hi @dmvolod

Thank you for your reply and input. 
What I am trying to say is:

If we want to make it work properly with replace, we should make this behaviour homogeneous for all variables, not only ENVTEST_K8S_VERSION. Also, it must work on both Linux and macOS, and stay simple / low-verbose.

For example; ( I think it might work )

```
# helper: get module version, honoring `replace`
define gomodver
$(shell go list -m -f '{{if .Replace}}{{.Replace.Version}}{{else}}{{.Version}}{{end}}' $(1) 2>/dev/null)
endef

# release-X.Y from controller-runtime (errors if version is empty)
ENVTEST_VERSION ?= $(shell v='$(call gomodver,sigs.k8s.io/controller-runtime)'; \
  [ -n "$$v" ] || { echo "Set ENVTEST_VERSION manually (controller-runtime replace has no tag?)" >&2; exit 1; }; \
  printf '%s\n' "$$v" | sed -E 's/^v?([0-9]+)\.([0-9]+).*/release-\1.\2/')

# 1.<minor> from k8s.io/api (errors if version is empty)
ENVTEST_K8S_VERSION ?= $(shell v='$(call gomodver,k8s.io/api)'; \
  [ -n "$$v" ] || { echo "Set ENVTEST_K8S_VERSION manually (k8s.io/api replace has no tag?)" >&2; exit 1; }; \
  printf '%s\n' "$$v" | sed -E 's/^v?[0-9]+\.([0-9]+).*/1.\1/')

```

Shubhamag12:
@camilamacedo86 Thank you for the suggestion, this makes sense. I agree that the scaffolding here got a bit too complex. I’ll try out the suggested solution and will let you know how it goes.

Shubhamag12:
@camilamacedo86 I’ve updated the implementation using the gomodver helper as suggested. The Makefile logic is now much simpler and consistent with other variables.
Let me know if this looks okay!

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -268,12 +268,22 @@ GOLANGCI_LINT = $(LOCALBIN)/golangci-lint
 ## Tool Versions
 KUSTOMIZE_VERSION ?= {{ .KustomizeVersion }}
 CONTROLLER_TOOLS_VERSION ?= {{ .ControllerToolsVersion }}
+
```
camilamacedo86, line 271: We do the change in the template 
and then we run 
- make install -> it will generate the bin locally with the changes
- make generate -> it will re-generate all samples, docs with the changes

Could you please ensure that do it? 
The only file that we manually change is makefile.go

Shubhamag12: Ah, my bad. I ran `make generate-testdata` instead of `make generate`, which missed docs changes

Shubhamag12: I have pushed changes after running `make generate`. 
PTAL

docs/book/src/cronjob-tutorial/testdata/project/Makefile
```diff
@@ -193,12 +193,22 @@ GOLANGCI_LINT = $(LOCALBIN)/golangci-lint
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.7.1
 CONTROLLER_TOOLS_VERSION ?= v0.19.0
+
+define gomodver
+$(shell go list -m -f '{{if .Replace}}{{.Replace.Version}}{{else}}{{.Version}}{{end}}' $(1) 2>/dev/null)
+endef
```
camilamacedo86: Since it can be valid for any after, could we add it at the end of the file?
That will help users update their projects to have those changes.
Sorry, but we need to be very mindful of changes in the scaffolds and the burn to adopt them

camilamacedo86: Only that and I am more than happy to get this one merged.

Shubhamag12: Sure, that makes sense

Shubhamag12: Done, pushed suggested changes

camilamacedo86 (approved):
/lgtm
/approved

PR #5081, opened 2025-09-05
🐛 (alpha generate command): Ensure kubebuilder reuses its own binary instead of relying on $PATH

ashishkpathak-connect:
Fix Kubebuilder executable path not found for `kubebuilder alpha generate` command.

Fixes #4900

PR #5074, opened 2025-09-03
Add unit tests for cli alpha generate.go

mayuka-c:
Updates: https://github.com/kubernetes-sigs/kubebuilder/issues/4925

- Improve the coverage by adding UTs for cli alpha generate.go.
- This is the continuation of my other UT PRs to improve coverage :)

PR #5102, opened 2025-09-16
🌱 Introduce Mockable getExecutablePath for Testing in Generate

mayuka-c:
This PR refactors the generate.go file to make the getExecutablePath function mockable, enabling better testability for functions that rely on it.

The Executable func was introduced as part of this PR: https://github.com/kubernetes-sigs/kubebuilder/pull/5081 which makes the os to use currently running executable which would fail for our UTs as it would try to use testing executable. With this change, it would enable us to mock according to our testing needs.

Here's the UT PR which I'm currently working on where I had made this change and validated: https://github.com/kubernetes-sigs/kubebuilder/pull/5074.

### DIFF
--- a/pkg/cli/alpha/internal/generate.go
+++ b/pkg/cli/alpha/internal/generate.go
@@ -42,6 +42,9 @@ type Generate struct {
 	OutputDir string
 }
 
+// Define a variable to allow overriding the behavior of getExecutablePath for testing.
+var getExecutablePathFunc = getExecutablePath
+
 // Generate handles the migration and scaffolding process.
 func (opts *Generate) Generate() error {
 	projectConfig, err := common.LoadProjectConfig(opts.InputDir)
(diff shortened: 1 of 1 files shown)


camilamacedo86:
Hi @mayuka-c 

Thank you so much for this one 🙏  

Just to clarify — we don’t need to reach 100% coverage. I’ll try to describe the issue more clearly to highlight what’s actually missing.  
We only need to add unit tests where it’s reasonable; we already have a lot of e2e tests that aren’t counted by Coveralls.  

Also, we don’t need to change existing code just to increase coverage.  
It would be better to include any additional tests as part of the related PRs, so we don’t add unnecessary requirements or discover later that they weren’t needed.  

Based on that, I’ll go ahead and close this one for now.  
If you notice that additional tests are required in a future PR, we can revisit it then.  

Thanks again for your great work!

PR #5103, opened 2025-09-16
📖 Tidy up completion.md

abenn135:
Tidy up completion.md. The doc as written is confusing and has a few issues. For example, following the instruction:

`echo “/usr/local/bin/bash” > /etc/shells`

(a) has the wrong kind of double quotes, and (b) blows away all your other shells instead of simply adding `/usr/local/bin/bash` to the list. This PR cleans that doc up a bit.

### DIFF
--- a/docs/book/src/reference/completion.md
+++ b/docs/book/src/reference/completion.md
@@ -2,43 +2,59 @@
 The Kubebuilder completion script can be generated with the command `kubebuilder completion [bash|fish|powershell|zsh]`.
 Note that sourcing the completion script in your shell enables Kubebuilder autocompletion.
 
+## Bash
+
 <aside class="note">
 <h1>Prerequisites for Bash</h1>
 
 The completion Bash script depends on [bash-completion](https://github.com/scop/bash-completion), which means that you have to install this software first (you can test if you have bash-completion already installed). Also, ensure that your Bash version is 4.1+.
 
+```bash
+$ bash --version
+```
+
 </aside>
 
+- Check that bash is an available shell:
 
-- Once installed, go ahead and add the path `/usr/local/bin/bash` in the  `/etc/shells`.
+    ```bash
+    cat /etc/shells | grep '^.*/bash'
+    ```
 
-    `echo “/usr/local/bin/bash” > /etc/shells`
+- If not, add bash to `/etc/shells`. For example, if bash is at `/usr/local/bin/bash`:
 
-- Make sure to use installed shell by current user.
+    ```bash
+    echo "/usr/local/bin/bash" >> /etc/shells
+    ```
 
-    `chsh -s /usr/local/bin/bash`
+- Make sure the current user uses bash as their shell.
 
-- Add following content in /.bash_profile or ~/.bashrc
+    ```bash
+    chsh -s /usr/local/bin/bash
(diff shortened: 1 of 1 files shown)


abenn135:
(CLA request is pending with my organization)

docs/book/src/reference/completion.md
```diff
@@ -2,43 +2,65 @@
...
 The completion Bash script depends on [bash-completion](https://github.com/scop/bash-completion), which means that you have to install this software first (you can test if you have bash-completion already installed). Also, ensure that your Bash version is 4.1+.
 
+```bash
+$ bash --version
+GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)
+Copyright (C) 2020 Free Software Foundation, Inc.
+License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
+
+This is free software; you are free to change and redistribute it.
+There is NO WARRANTY, to the extent permitted by law.
```
vitorfloriano: Does this info add any value? Couldn't we just have the command and leave out the output? WDYT?

docs/book/src/reference/completion.md
```diff
@@ -2,43 +2,65 @@
...
+```
+
 </aside>
 
+- Check that bash is an available shell:
 
-- Once installed, go ahead and add the path `/usr/local/bin/bash` in the  `/etc/shells`.
+    ```bash
+    cat /etc/shells | grep '^.*/bash'
+    ```
```
vitorfloriano, line 22: NIT: I noticed the codeblocks added in this PR have a different identation than the other codeblocks already present in the file. Is there a reason for that?

docs/book/src/reference/completion.md
```diff
@@ -2,43 +2,65 @@
...
 
-<aside class="note">
-<h1>Fish</h1>
+## Fish
 
+```
 source (kubebuilder completion fish | psub)
-
-</aside>
+```
```
vitorfloriano, line 60: Could we take this opportunity and also expand on the instructions for Zsh and Fish? Just a suggestion. WDYT?

docs/book/src/reference/completion.md
abenn135: Sure, I can remove it. Done.

docs/book/src/reference/completion.md
abenn135: This indent keeps them within the containing bullet points. So instead of

  * Blah blah

`<code at top level indent, breaking bulleted list>`

  * disconnected bullet

you get

  * blah blah

    `<code contained within bullet with paragraph break>`

  * Next bullet continues bulleted list

docs/book/src/reference/completion.md
abenn135: I don't have zsh nor fish installed, so I wouldn't be able to test those commands unfortunately.

abenn135 (commented):
Please take another look :)

docs/book/src/reference/completion.md
vitorfloriano: Oh I see it now 🤔 I hadn't looked at the preview of the file. Yeah, it looks better indented that way.

PR #5104, opened 2025-09-18
📖 (docs) drop patchStrategy/protobuf tags from Conditions in Getting Started section

pedjak:
Followup of PR #4893

These tags are of no use for CRDs. They have their use
in k8s core types in order to produce proper strategic merge patches, see
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md
but are superseded by +listType and +listMapKey in CRD world.

Added `+listType`, `+listMapKey` and `+optional` markers to the example snippet


camilamacedo86 (approved):
/lgtm
/ok-to-test

PR #5110, opened 2025-09-23
🌱 chore: enforce internal package import aliases

bavarianbidi:
As we have some inconsistency on few aliased imports (typo, different alias naming) which also leads to using wrong packages in one/two code path

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -13,6 +13,7 @@ linters:
     - goconst
     - gocyclo
     - govet
+    - importas
     - ineffassign
     - lll
     - misspell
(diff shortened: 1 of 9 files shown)


.golangci.yml
```diff
@@ -34,6 +35,23 @@ linters:
...
+        - pkg: sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/helm/v1alpha
+          alias: helmv1alpha
+        - pkg: sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/helm/v2alpha
+          alias: helmv2alpha
+        - pkg: sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/grafana/v1alpha
+          alias: grafanav1alpha
+        - pkg: "sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/autoupdate/v1alpha"
+          alias: autoupdatev1alpha
+        - pkg: sigs.k8s.io/kubebuilder/v4/pkg/plugins/golang/deploy-image/v1alpha1
+          alias: deployimagev1alpha1
```
vitorfloriano, line 50: Hi @bavarianbidi, thanks for working on this! Question: would it be too much trouble to change this to `deployimagev1alpha` for consistency? Perhaps we could open another PR to address that, considering the size of the change and the scope of this one. WDYT?

bavarianbidi: hey, it's quite easy to change this.

The current pattern is: `<pluginname><packageversionname>`
and `deploy-image` has the version `v1alpha1`.
For consistency reasons i would rather keep this instead of being inconsistent with the version name.

But i'm not too strict about this :sweat_smile: 
just wanted to have consistent import names all over the place + using the packages where they are intent to get used (xref: https://github.com/kubernetes-sigs/kubebuilder/pull/5110/files#diff-809dfec4c6e918b86602610190069c0f5a049d47b48a94a2417bcf60cf0bde65L272 where we've used the grafana-plugin package for the auto-update package)

vitorfloriano: You’re right, I missed that it was `v1alpha1`. Thanks for pointing it out, and sorry for the confusion!

PR #5119, opened 2025-10-07
🐛  (helm/v2-alpha): Fixed an issue where the `ServiceMonitor` resource name did not include the default prefix applied by Kustomize. 

liam-mackie:
**Background:**
When working with long names, you will often need to use a shortened (or abbreviated) version of the controller name as the `namePrefix` in the `kustomization.yaml` file. This means a shorter name can be used as the prefix to all the files.

When using the `v2alpha/helm` plugin, this _mostly_ works. The only resource that doesn't use the name generated by Kustomize is the `ServiceMonitor`. This is problematic, as it will use the name (with the custom prefix), then template the chart name (usually the long name) in front of it. This leads to a broken installation.

**The fix:**
This PR fixes this by attempting to get the `namePrefix` from the kustomization config path. If the `namePrefix` and the `projectName` are different, we do not template the file for the service monitor. This way, customisations are honoured, but existing behaviour is maintained.

### DIFF
--- a/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater_test.go
+++ b/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater_test.go
@@ -389,6 +389,23 @@ metadata:
 			Expect(result).To(ContainSubstring("name: test-project-controller-manager"))
 			Expect(result).NotTo(ContainSubstring("{{ include"))
 		})
+		It("should preserve name for ServiceMonitor", func() {
+			serviceMonitorResource := &unstructured.Unstructured{}
+			serviceMonitorResource.SetAPIVersion("monitoring.coreos.com/v1")
+			serviceMonitorResource.SetKind("ServiceMonitor")
+			serviceMonitorResource.SetName("test-project-controller-manager-metrics-monitor")
+
+			content := `apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: test-project-controller-manager-metrics-monitor`
+
+			result := templater.ApplyHelmSubstitutions(content, serviceMonitorResource)
+
+			// Name should remain unchanged
+			Expect(result).To(ContainSubstring("name: test-project-controller-manager-metrics-monitor"))
+			Expect(result).NotTo(ContainSubstring("{{ include"))
+		})
 	})
 
 	Context("edge cases", func() {
--- a/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/parser.go
+++ b/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/parser.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"gopkg.in/yaml.v3"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -150,4 +151,27 @@ func (p *Parser) categorizeResource(obj *unstructured.Unstructured, resources *P
 	default:
 		resources.Other = append(resources.Other, obj)
 	}
+}
+
+func (pr *ParsedResources) EstimatePrefix(projectName string) string {
+	prefix := projectName
+	if pr.Deployment != nil {
+		if name := pr.Deployment.GetName(); name != "" {
+			deploymentPrefix, found := strings.CutSuffix(name, "-controller-manager")
+			if found {
+				prefix = deploymentPrefix
+			}
+		}
+	}
+	// Double check that the prefix is also the prefix for the service names
+	for _, svc := range pr.Services {
+		if name := svc.GetName(); name != "" {
+			if !strings.HasPrefix(name, prefix) {
+				// If not, fallback to just project name
+				prefix = projectName
+				break
+			}
+		}
+	}
+	return prefix
 }
--- a/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/servicemonitor.go
+++ b/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/servicemonitor.go
@@ -29,7 +29,12 @@ var _ machinery.Template = &ServiceMonitor{}
 // ServiceMonitor scaffolds a ServiceMonitor for Prometheus monitoring in the Helm chart
 type ServiceMonitor struct {
 	machinery.TemplateMixin
-	machinery.ProjectNameMixin
+
+	// Prefix
+	NamePrefix string
+
+	// ServiceName is the full name of the metrics service, derived from Kustomize
+	ServiceName string
 
 	// OutputDir specifies the output directory for the chart
 	OutputDir string
@@ -59,7 +64,7 @@ metadata:
   labels:
     {{ "{{- include \"chart.labels\" . | nindent 4 }}" }}
     control-plane: controller-manager
-  name: {{ .ProjectName }}-controller-manager-metrics-monitor
+  name: {{ .NamePrefix }}-controller-manager-metrics-monitor
   namespace: {{ "{{ .Release.Namespace }}" }}
 spec:
   endpoints:
@@ -69,7 +74,7 @@ spec:
       bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
       tlsConfig:
         {{ "{{- if .Values.certManager.enable }}" }}
-        serverName: {{ .ProjectName }}-controller-manager-metrics-service.{{ "{{ .Release.Namespace }}" }}.svc
+        serverName: {{ .ServiceName }}.{{ "{{ .Release.Namespace }}" }}.svc
         # Apply secure TLS configuration with cert-manager
         insecureSkipVerify: false
         ca:
--- a/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/prometheus/controller-manager-metrics-monitor.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/prometheus/controller-manager-metrics-monitor.yaml
@@ -6,7 +6,7 @@ metadata:
         app.kubernetes.io/managed-by: {{ .Release.Service }}
         app.kubernetes.io/name: project
         control-plane: controller-manager
-    name: {{ include "chart.name" . }}-controller-manager-metrics-monitor
+    name: project-controller-manager-metrics-monitor
     namespace: {{ .Release.Namespace }}
 spec:
     endpoints:
--- a/docs/book/src/multiversion-tutorial/testdata/project/dist/chart/templates/prometheus/controller-manager-metrics-monitor.yaml
+++ b/docs/book/src/multiversion-tutorial/testdata/project/dist/chart/templates/prometheus/controller-manager-metrics-monitor.yaml
@@ -6,7 +6,7 @@ metadata:
         app.kubernetes.io/managed-by: {{ .Release.Service }}
         app.kubernetes.io/name: project
         control-plane: controller-manager
-    name: {{ include "chart.name" . }}-controller-manager-metrics-monitor
+    name: project-controller-manager-metrics-monitor
     namespace: {{ .Release.Namespace }}
 spec:
     endpoints:
--- a/pkg/plugins/optional/helm/v2alpha/scaffolds/edit_kustomize.go
+++ b/pkg/plugins/optional/helm/v2alpha/scaffolds/edit_kustomize.go
@@ -109,6 +109,7 @@ func (s *editKustomizeScaffolder) Scaffold() error {
 			slog.Warn("failed to remove stale generic ServiceMonitor", "path", staleSM, "error", rmErr)
 		}
 	}
+	namePrefix := resources.EstimatePrefix(s.config.GetProjectName())
 	chartConverter := kustomize.NewChartConverter(resources, s.config.GetProjectName(), s.outputDir)
 	deploymentConfig := chartConverter.ExtractDeploymentConfig()
 
@@ -119,7 +120,8 @@ func (s *editKustomizeScaffolder) Scaffold() error {
 	chartFiles := []machinery.Builder{
 		&github.HelmChartCI{},                        // GitHub Actions workflow for chart testing
 		&templates.HelmChart{OutputDir: s.outputDir}, // Chart.yaml metadata
-		&templates.HelmValuesBasic{ // values.yaml with dynamic config
+		&templates.HelmValuesBasic{
+			// values.yaml with dynamic config
 			HasWebhooks:      hasWebhooks,
 			HasMetrics:       hasMetrics,
 			DeploymentConfig: deploymentConfig,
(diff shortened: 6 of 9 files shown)


liam-mackie:
@Kavinjsir / @varshaprasad96 - is there any possibility of getting this one reviewed/approved to test? I'd love to get this in!

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater_test.go
```diff
@@ -225,6 +226,35 @@ metadata:
...
+				templaterDiff := &HelmTemplater{
+					projectName: "project-with-a-long-name",
+					namePrefix:  "test-project",
+				}
+				result := templaterDiff.ApplyHelmSubstitutions(content, serviceMonitorResource)
+				// If the namePrefix and default chart name differ, should use namePrefix variable
+				Expect(result).ToNot(ContainSubstring("{{ include \"chart.name\" . }}"), "Should not use chart.name template when namePrefix and project name differ")
+			})
+		})
+
```
camilamacedo86: Could you please fix the lint issues?

pkg/plugins/optional/helm/v2alpha/edit.go
```diff
@@ -98,6 +101,8 @@ func (p *editSubcommand) BindFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&p.force, "force", false, "if true, regenerates all the files")
 	fs.StringVar(&p.manifestsFile, "manifests", DefaultManifestsFile,
 		"path to the YAML file containing Kubernetes manifests from kustomize output")
+	fs.StringVar(&p.kustomizePath, "kustomize-path", DefaultKustomizeFile,
```
camilamacedo86: I do not think that we should pass the path here, otherwise we will need a flag for each file.
In this case, we might want to get the config from PROJECT file which has the project name instead of check the kustomize file wdyt?

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/servicemonitor.go
```diff
@@ -31,6 +31,9 @@ type ServiceMonitor struct {
 	machinery.TemplateMixin
```
camilamacedo86, line 31: The best solution would be to have the ServiceMonitor from the kustomize install.yaml
Could we look at this option?

liam-mackie: When prometheus is enabled in kustomize, it'll use the ServiceMonitor from the kustomize install.yaml - this code path is only called when prometheus is **_disabled_** in kustomize. 
I _assume_ it's to allow prometheus to be enabled/disabled in the values.yaml of the helm chart.

pkg/plugins/optional/helm/v2alpha/edit.go
liam-mackie: Unfortunately, there's no reference to the entrypoint for Kustomize in the PROJECT file, and there's no way to infer the prefix someone uses in Kustomize outside of the `kustomization.yaml` itself. Kubebuilder conventions put it in `config/default/kustomization.yaml`. Originally, I left this as unconfigurable, and simply assumed that the `kustomization.yaml` would always be in this path.

The best we have to get this programatically is the Makefile, which uses it in the `build-installer` target:
```
	$(KUSTOMIZE) build config/default > dist/install.yaml
```

The `config/default` path is what we'd use, and we'd simply look for a `kustomization.yaml` in there. I'm not sure if that's any better than simply allowing the desired kustomization file to be passed through though. 

Happy to defer to your expertise here!

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater_test.go
liam-mackie: On it! 🙇

liam-mackie: Fixed in [44e16fe](https://github.com/kubernetes-sigs/kubebuilder/pull/5119/commits/44e16fe22c77f26d973824d5cfcab0d49538fe11)

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/servicemonitor.go
camilamacedo86: Yes, that is the reason for we template this one.
But I think we can look for:

```
name: {{ .NamePrefix }}-controller-manager-metrics-monitor  
serverName: {{ .NamePrefix }}-controller-manager-metrics-service  
```

Those values in the install.yaml and use here instead of use PROJECT name. 

Example: 

- The serverName is the name of the service so we can look for: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-with-plugins/dist/install.yaml#L922-L939
- And use the prefix found above for the name

liam-mackie: Ah ok, I think I understand - so take the actual values of the service name from the install.yaml rather than assuming it's the project name or prefix?

liam-mackie: Thanks for the clarification - I'll get something up ASAP for you to take a look at 🙇

liam-mackie: I've added an option in [d2a79c6](https://github.com/kubernetes-sigs/kubebuilder/pull/5119/commits/d2a79c69417acee99655daf24c0748590224c1f7). I was wondering what your thoughts on the testing strategy should be here - there wasn't any existing tests, and I'm not fully sure where to put the seam!

camilamacedo86: But then we need to remove the rest only get the data from the install.yaml
We should avoid adding flags since it will make it harder to keep things maintained

Could we keep the solution here only looking for the install.yaml to get this value?

liam-mackie: Sorry, I meant that it's an option that we can consider, rather than an actual option that the user provides! It's not a new flag, and is the only way the data will be passed in to this template!

camilamacedo86: It will be a new flag for the plugin: fs.StringVar(&p.kustomizePath, "kustomize-path", DefaultKustomizeFile,
That means specific tests for it, and also if we pass one file, we might need to pass others, and it can grow, etc. Also, users will need to pass more than one value by default to properly generate it and bring a bad UX

So, the best approach is we get the values from the install.yaml 
We should not consume any other place.

liam-mackie: Oh, I see - you mean the kustomize path. I'll get something working and run it by you, I have an idea here.

liam-mackie: [Get data from kustomize output rather than kustomization.yaml](https://github.com/kubernetes-sigs/kubebuilder/pull/5119/commits/182aa72b34745f3e210929d31c74bcad6ce0a8df) changes the way we get the prefix to instead get it from the deployment name. Is this closer to what you were hoping for?

camilamacedo86: I think we should ONLY:

1. Check the `install.yaml` (file-informed).
2. Look for the `controller-manager-metrics-service`.
     - Get its full name.
     - Use it to populate the service name.
3. For the monitor name:
    - Extract all text that comes before.
    - Concatenate it with the default pattern:  ` <prefix obtained from service name>-controller-manager-metrics-monitor`

This should make the scaffold populate the correct values automatically, without requiring extra information or flags. Only look for a specific path if one is explicitly provided and do not bring extra burn to keep things maintained.

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/chart_converter.go
```diff
@@ -37,9 +37,9 @@ type ChartConverter struct {
 }
 
 // NewChartConverter creates a new chart converter with all necessary components
-func NewChartConverter(resources *ParsedResources, projectName, outputDir string) *ChartConverter {
+func NewChartConverter(resources *ParsedResources, projectName, namePrefix, outputDir string) *ChartConverter {
```
camilamacedo86: The projectName and namePrefix here are the same.
So, why would we need both parameters? 
Did I miss something?

liam-mackie: In the tests they're the same, but in the code itself, the prefix could possibly be different. We estimate it [here](https://github.com/kubernetes-sigs/kubebuilder/pull/5119/files#diff-23e723f761f01aa5b206dc50bf40d60017673f798b0c8f7495399b66ff80684bR112) and pass it in, as the templater needs it.

liam-mackie:
> Hi @liam-mackie
> 
> Really thank you for the amazing work 🥇 I think we are very close.
> 
> I have just one comment in the latest review — see: [#5119 (files)](https://github.com/kubernetes-sigs/kubebuilder/pull/5119/files#r2434870465)
> 
> Also, note that for us to get this one merged, we need the commits squashed. Could you please ensure that you squash all commits so that we have only **one** commit in this PR? Also, could you please rebase this one with master so that it can pass in the tests?
> 
> Lastly, I’ll trigger the tests to check if any fail.

All done! Rebased and squashed - thanks for the help!

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/parser.go
```diff
@@ -151,3 +152,26 @@ func (p *Parser) categorizeResource(obj *unstructured.Unstructured, resources *P
 		resources.Other = append(resources.Other, obj)
 	}
 }
+
+func (pr *ParsedResources) EstimatePrefix(projectName string) string {
+	prefix := projectName
+	if pr.Deployment != nil {
+		if name := pr.Deployment.GetName(); name != "" {
+			deploymentPrefix, found := strings.CutSuffix(projectName, "-controller-manager")
```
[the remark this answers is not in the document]
liam-mackie: Fixed + squashed

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater_test.go
```diff
@@ -225,6 +226,43 @@ metadata:
...
+`
+			})
+
+			It("the namePrefix and project name are the same", func() {
+				result := templater.ApplyHelmSubstitutions(content, serviceMonitorResource)
+				// If the namePrefix and default chart name match, should use chart.name template
+				Expect(result).To(ContainSubstring("{{ include \"chart.name\" . }}"),
+					"Should use chart.name template when namePrefix and project name match")
+			})
+			It("the namePrefix and project name differ", func() {
```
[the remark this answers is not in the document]
liam-mackie: Fixed and squashed

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater_test.go
```diff
@@ -225,6 +226,43 @@ metadata:
...
+				// Test content must end with a newline
+				// If not, the `name` regex in the ServiceMonitor replacement will never match
+				content = `apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: test-project-controller-manager-metrics-monitor
+`
+			})
+
+			It("the namePrefix and project name are the same", func() {
```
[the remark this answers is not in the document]
liam-mackie: Fixed and squashed

camilamacedo86:
I will need some time to properly evaluate
I think we might want to only use the logic for the prefix for serviceMonitor 

/ok-to-test

liam-mackie:
> I think we might want to only use the logic for the prefix for serviceMonitor

That's a good point - we only see the chart name being used at all in the `serviceMonitor`, but everything else is neatly pulled directly from the rendered Kustomize YAML. Do you think it's a better idea to simply standardise on using the resource names from the rendered Kustomize? I'm honestly not fully sure why we have the `{{ chart.name }}` replacement, so I'm fully open to pulling it out completely as a part of this change instead, so long as it's a sane change from your end!

camilamacedo86:
Hi @liam-mackie 👋

I think the more surgical we are here, the better.  
My suggestion is to focus only on getting the data for the `ServiceMonitor` and keep everything else as it is for now.  
If needed, we can revisit and expand later.

Could we try it out this way?  
It would also make the review process easier — otherwise, we’d need to start looking into other design aspects as well.

camilamacedo86:
Hi @liam-mackie 

Please, feel free to ping me if you be able to enhance this one.
For we get merged.

liam-mackie:
> Hi @liam-mackie
> 
> Please, feel free to ping me if you be able to enhance this one. For we get merged.


I'm working on this now, actually! I'm planning to remove the API changes and follow your guidance, specifically only affecting the ServiceMonitor. Just to confirm, my plan is as follows:
* Use the rendered name for the `ServiceMonitor` from the Kustomize output if the ServiceMonitor exists in the configuration (I.E. the user has the `ServiceMonitor` in the default install.yaml)
* Guess the correct prefix (as per existing PR changes) to add to the `ServiceMonitor` template that we scaffold if the `install.yaml` does not have the ServiceMonitor included by default
This should keep the user-facing API the same

Can you please let me know if this matches your expectations? I'm happy to sync over zoom or other types of communication should you be available and want to 🙇

camilamacedo86:
Hi @liam-mackie 

Just one thing the `[APIDiff / Verify API differences (pull_request)](https://github.com/kubernetes-sigs/kubebuilder/actions/runs/18698927569/job/53328582745?pr=5119)` is not a concern here. In this case, we can overwrite. 

Let's see if we can make the fix more surgically, changing only the specific template, and then we can get this one merged. Thank you a lot !!!

liam-mackie:
> Let's see if we can make the fix more surgically, changing only the specific template, and then we can get this one merged. 

Yes - it'll just be changing these two specific templates - they're just both ways of us creating a `ServiceMonitor` depending on the output we get from Kustomize. As long as you're happy removing the `{{ chart.name }}` template from the servicemonitor name, I think this will work well. 
I should have time to finish this off tomorrow morning, AEST (in ~16 hours), so by the time you're online tomorrow, this should be ready for review. I'll ping you specifically letting you know my changes are pushed 🙇

liam-mackie:
@camilamacedo86 I've been a little more precise here and simply _removed_ `templateServiceMonitorNames` to just keep the service monitor names as they come from Kustomize. Let me know how the PR looks now :)

camilamacedo86:
Hi @liam-mackie 

That is amazing.
Just run make lint-fix 
and then we can get this one merged !!!

Great job 🎉

liam-mackie:
> That is amazing. Just run make lint-fix and then we can get this one merged !!!

Awesome! That's done now :)

camilamacedo86:
@liam-mackie 

it is missing run `make install` and `make generate-charts` to update all samples accordingly

liam-mackie:
> @liam-mackie 
> 
> 
> 
> it is missing run `make install` and `make generate-charts` to update all samples accordingly


Argh! I'll regenerate them now, sorry!

liam-mackie:
@camilamacedo86 Done + Rebased 🙇

PR #5130, opened 2025-10-13
✨  (go/v4) Upgrade golang version from go 1.24 to 1.25

camilamacedo86:


camilamacedo86:
We have a check to ensure that the tag is still valid.
I will move forward to unblock us, but anyone, please feel free to push a PR with improvements if you see fit.

Issue #5137, opened 2025-10-16
Flakes in e2e tests for metrics


camilamacedo86:
### What broke? What's expected?

We have flakes on the e2e tests with the metrics, see: https://github.com/kubernetes-sigs/kubebuilder/actions/runs/18553635948/job/52885929295?pr=5136

We need to investigate the root cause and fix it. 

### KubeBuilder (CLI) Version

master

### Plugin versions

```yaml

```

mayuka-c:
I could probably look into this
/assign

camilamacedo86:
It is fixed in the release 4.10.1

PR #5138, opened 2025-10-17
 🐛 Fix flakes in e2e tests for metrics

mayuka-c:
Fixes: https://github.com/kubernetes-sigs/kubebuilder/issues/5137

By looking at the logs, looks like during the curl metrics pod creation, calls to the webhook is failing due to connection refused. So added precheck to ensure the webhook is ready before beginning with the pod creation.

Validation
---
Ran the e2e tests multiple times here and did not see the flakes: https://github.com/mayuka-c/kubebuilder/actions/runs/18582517449

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_test.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_test.go
@@ -216,6 +216,41 @@ var _ = Describe("Manager", Ordered, func() {
 			}
 			Eventually(verifyMetricsServerStarted).Should(Succeed())
 
+			By("waiting for webhook service to be ready")
+			verifyWebhookServiceReady := func(g Gomega) {
+				const webhookServiceName = "project-webhook-service"
+
+				// Webhook service should exist since webhooks are configured
+				cmd := exec.Command("kubectl", "get", "service", webhookServiceName, "-n", namespace)
+				_, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred(), "Webhook service should exist but was not found")
+
+				// Check if webhook server is ready by verifying pod readiness
+				cmd = exec.Command("kubectl", "get", "pods", "-l", "control-plane=controller-manager",
+					"-n", namespace, "-o", "jsonpath={.items[0].status.conditions[?(@.type=='Ready')].status}")
+				output, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(output).To(Equal("True"),
+					"Controller manager pod not ready (webhook server may not be accepting connections)")
+
+				// Check if webhook service endpoints are available
+				cmd = exec.Command("kubectl", "get", "endpoints", webhookServiceName,
+					"-n", namespace, "-o", "jsonpath={.subsets[*].addresses[*].ip}")
+				output, err = utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(output).NotTo(BeEmpty(), "Webhook service endpoints are not ready")
+
+				// Test webhook connectivity by checking if webhook server port is responding
+				cmd = exec.Command("kubectl", "run", "webhook-test", "--rm", "-i", "--restart=Never",
+					"--image=curlimages/curl:latest", "--",
+					"curl", "-k", "--connect-timeout", "5",
+					"https://"+webhookServiceName+"."+namespace+".svc:443/readyz")
+				_, err = utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred(), "Webhook server not responding on port 443")
+			}
+			Eventually(verifyWebhookServiceReady, 2*time.Minute).Should(Succeed())
+			// +kubebuilder:scaffold:e2e-webhooks-readiness
+
 			By("creating the curl-metrics pod to access the metrics endpoint")
 			cmd = exec.Command("kubectl", "run", "curl-metrics", "--restart=Never",
 				"--namespace", namespace,
--- a/docs/book/src/reference/markers/scaffold.md
+++ b/docs/book/src/reference/markers/scaffold.md
@@ -107,6 +107,7 @@ properly registered with the manager, so that the controller can reconcile the r
 | `+kubebuilder:scaffold:crdkustomizecainjectioname`                           | `config/default`             | Marks where CA injection patches are added for the conversion webhooks.                                                                                                                |
 | **(No longer supported)** `+kubebuilder:scaffold:crdkustomizecainjectionpatch` | `config/crd`                 | Marks where CA injection patches are added for the webhooks. Replaced by `+kubebuilder:scaffold:crdkustomizecainjectionns` and `+kubebuilder:scaffold:crdkustomizecainjectioname`  |
 | `+kubebuilder:scaffold:manifestskustomizesamples` | `config/samples`           | Marks where Kustomize sample manifests are injected.                            |
+| `+kubebuilder:scaffold:e2e-webhooks-readiness` | `test/e2e`                   | Adds webhooks readiness checks in e2e tests before metrics collection.          |
 | `+kubebuilder:scaffold:e2e-webhooks-checks` | `test/e2e`                   | Adds e2e checks for webhooks depending on the types of webhooks scaffolded.      |
 
 <aside class="note warning">
--- a/testdata/project-v4-multigroup/test/e2e/e2e_test.go
+++ b/testdata/project-v4-multigroup/test/e2e/e2e_test.go
@@ -211,6 +211,41 @@ var _ = Describe("Manager", Ordered, func() {
 			}
 			Eventually(verifyMetricsServerStarted).Should(Succeed())
 
+			By("waiting for webhook service to be ready")
+			verifyWebhookServiceReady := func(g Gomega) {
+				const webhookServiceName = "project-v4-multigroup-webhook-service"
+
+				// Webhook service should exist since webhooks are configured
+				cmd := exec.Command("kubectl", "get", "service", webhookServiceName, "-n", namespace)
+				_, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred(), "Webhook service should exist but was not found")
+
+				// Check if webhook server is ready by verifying pod readiness
+				cmd = exec.Command("kubectl", "get", "pods", "-l", "control-plane=controller-manager",
+					"-n", namespace, "-o", "jsonpath={.items[0].status.conditions[?(@.type=='Ready')].status}")
+				output, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(output).To(Equal("True"),
+					"Controller manager pod not ready (webhook server may not be accepting connections)")
+
+				// Check if webhook service endpoints are available
+				cmd = exec.Command("kubectl", "get", "endpoints", webhookServiceName,
+					"-n", namespace, "-o", "jsonpath={.subsets[*].addresses[*].ip}")
+				output, err = utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(output).NotTo(BeEmpty(), "Webhook service endpoints are not ready")
+
+				// Test webhook connectivity by checking if webhook server port is responding
+				cmd = exec.Command("kubectl", "run", "webhook-test", "--rm", "-i", "--restart=Never",
+					"--image=curlimages/curl:latest", "--",
+					"curl", "-k", "--connect-timeout", "5",
+					"https://"+webhookServiceName+"."+namespace+".svc:443/readyz")
+				_, err = utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred(), "Webhook server not responding on port 443")
+			}
+			Eventually(verifyWebhookServiceReady, 2*time.Minute).Should(Succeed())
+			// +kubebuilder:scaffold:e2e-webhooks-readiness
+
 			By("creating the curl-metrics pod to access the metrics endpoint")
 			cmd = exec.Command("kubectl", "run", "curl-metrics", "--restart=Never",
 				"--namespace", namespace,
--- a/docs/book/src/getting-started/testdata/project/test/e2e/e2e_test.go
+++ b/docs/book/src/getting-started/testdata/project/test/e2e/e2e_test.go
@@ -211,6 +211,8 @@ var _ = Describe("Manager", Ordered, func() {
 			}
 			Eventually(verifyMetricsServerStarted).Should(Succeed())
 
+			// +kubebuilder:scaffold:e2e-webhooks-readiness
+
 			By("creating the curl-metrics pod to access the metrics endpoint")
 			cmd = exec.Command("kubectl", "run", "curl-metrics", "--restart=Never",
 				"--namespace", namespace,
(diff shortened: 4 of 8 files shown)


camilamacedo86 (changes requested):
Hi @mayuka-c 👋

That looks great! However, this change only makes sense for projects that actually have webhooks.
Not all projects do — we use markers to inject code (see [Kubebuilder markers](https://book.kubebuilder.io/reference/markers/scaffold)
).
For example:
```
+kubebuilder:scaffold:e2e-webhooks-checks
```

Then, if we really need that we will need to see if we can use the same; it seems not. Then we need a new marker. So, we need to be 1000% sure that is the best way. Ideally we should avoid create new markers and etc since we need keep the support for those, maker harder for end user customisartions, etc. 

So, if we need to check whether the webhook is serving before the metrics, we should only add this logic when webhooks are present — meaning we’ll need to use a marker for it.

Also, please make sure the scaffolded code remains as generic as possible, so it still works even if users customise their configurations.

Lastly, since this change affects the scaffolded e2e tests in projects (which impact end users), we need to be careful with the commit type.
We should use the 🐛 emoji instead of 🌱 — the seed emoji is only used for changes that don’t affect users and can be ignored in the release notes.

mayuka-c:
Hi @camilamacedo86 
Regarding below:-

> That looks great! However, this change only makes sense for projects that actually have webhooks. Not all projects do — we use markers to inject code (see [Kubebuilder markers](https://book.kubebuilder.io/reference/markers/scaffold) ). For example:
> 
> ```
> +kubebuilder:scaffold:e2e-webhooks-checks
> ```
> 
> Then, if we really need that we will need to see if we can use the same; it seems not. Then we need a new marker. So, we need to be 1000% sure that is the best way. Ideally we should avoid create new markers and etc since we need keep the support for those, maker harder for end user customisartions, etc.
> 
> So, if we need to check whether the webhook is serving before the metrics, we should only add this logic when webhooks are present — meaning we’ll need to use a marker for it.
> 
> Also, please make sure the scaffolded code remains as generic as possible, so it still works even if users customise their configurations.

If I'm not wrong, I could add a new fragment to do the webhook readiness check [here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/test.go#L82) and replace with the marker [here](https://github.com/mayuka-c/kubebuilder/blob/1c32309af76294af5d209832db9cfe469da5a2ee/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/test.go#L397)

But the only problem that I see is, it will end up running other fragment checks too which is not required.

Shall I go ahead to define a new marker? or if you have any better solution, please do suggest :)

mayuka-c:
> Lastly, since this change affects the scaffolded e2e tests in projects (which impact end users), we need to be careful with the commit type. We should use the 🐛 emoji instead of 🌱 — the seed emoji is only used for changes that don’t affect users and can be ignored in the release notes.

Sorry my bad, I will update the title with proper emoji. Thank you :)

camilamacedo86:
Hi @mayuka-c 

Let's create a new marker 
Also, if you can review the code and ensure that is generic and good enough for any project that would be great 🎉

mayuka-c:
> Hi @mayuka-c
> 
> Let's create a new marker Also, if you can review the code and ensure that is generic and good enough for any project that would be great 🎉

Sure thank you @camilamacedo86 . I will review the code and make sure it is generic.

mayuka-c:
Hi @camilamacedo86 👋
Sorry for getting back late, as I was bust with other things.

I have added a new marker `e2e-webhook-readiness` and it seems to working fine.
I see the readiness check test cases are used in projects where the webhooks are present in Projectfile and for the basic test-data project, the webhook tests are not added.

Please let me know if this is fine or if I'm missing

docs/book/src/reference/markers/scaffold.md
```diff
@@ -107,6 +107,7 @@ properly registered with the manager, so that the controller can reconcile the r
 | `+kubebuilder:scaffold:crdkustomizecainjectioname`                           | `config/default`             | Marks where CA injection patches are added for the conversion webhooks.                                                                                                                |
 | **(No longer supported)** `+kubebuilder:scaffold:crdkustomizecainjectionpatch` | `config/crd`                 | Marks where CA injection patches are added for the webhooks. Replaced by `+kubebuilder:scaffold:crdkustomizecainjectionns` and `+kubebuilder:scaffold:crdkustomizecainjectioname`  |
 | `+kubebuilder:scaffold:manifestskustomizesamples` | `config/samples`           | Marks where Kustomize sample manifests are injected.                            |
+| `+kubebuilder:scaffold:e2e-webhook-readiness` | `test/e2e`                   | Adds webhook readiness checks in e2e tests before metrics collection.          |
```
camilamacedo86:
```suggestion
| `+kubebuilder:scaffold:e2e-webhooks-readiness` | `test/e2e`                   | Adds webhook readiness checks in e2e tests before metrics collection.          |
```

Just to keep it as the other one. WDYT?

mayuka-c: Looks like a typo in the md file. Sorry my bad. I have corrected it.
Thank you :)

testdata/project-v4-multigroup/test/e2e/e2e_test.go
```diff
@@ -211,6 +211,50 @@ var _ = Describe("Manager", Ordered, func() {
...
+				// Test webhook connectivity by checking if webhook server port is responding
+				cmd = exec.Command("kubectl", "run", "webhook-test", "--rm", "-i", "--restart=Never",
+					"--image=curlimages/curl:latest", "--",
+					"curl", "-k", "--connect-timeout", "5",
+					"https://"+webhookServiceName+"."+namespace+".svc:443/readyz")
+				_, err = utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred(), "Webhook server not responding on port 443")
+			}
+			Eventually(verifyWebhookServiceReady, 2*time.Minute).Should(Succeed())
+			// +kubebuilder:scaffold:e2e-webhooks-readiness
```
camilamacedo86, line 247: I think we may can simplify with

```go
By("waiting for webhook service to be ready if webhooks are configured")
verifyWebhookServiceReady := func(g Gomega) {
	const webhookServiceName = "{{ProjectName}}-webhook-service"

	cmd := exec.Command("kubectl", "get", "service", webhookServiceName, "-n", namespace)
	_, err := utils.Run(cmd)
	if err != nil {
		// Project has no webhook service; nothing to wait on.
		return
	}

	cmd = exec.Command("kubectl", "get", "pods", "-l", "control-plane=controller-manager",
		"-n", namespace, "-o", "jsonpath={.items[0].status.conditions[?(@.type=='Ready')].status}")
	output, err := utils.Run(cmd)
	g.Expect(err).NotTo(HaveOccurred())
	g.Expect(output).To(Equal("True"),
		"Controller manager pod not ready (webhook server may not be accepting connections)")

	cmd = exec.Command("kubectl", "get", "endpoints", webhookServiceName,
		"-n", namespace, "-o", "jsonpath={.subsets[*].addresses[*].ip}")
	output, err = utils.Run(cmd)
	g.Expect(err).NotTo(HaveOccurred())
	g.Expect(output).NotTo(BeEmpty(), "Webhook service endpoints are not ready")

	cmd = exec.Command("kubectl", "get", "--raw",
		fmt.Sprintf("/api/v1/namespaces/%s/services/https:%[2]s:https/proxy/readyz", namespace, 		webhookServiceName))
	_, err = utils.Run(cmd)
	g.Expect(err).NotTo(HaveOccurred(), "Webhook server not responding on port 443")
}
Eventually(verifyWebhookServiceReady, 2*time.Minute).Should(Succeed())
```

camilamacedo86: WDYT>

mayuka-c: Thanks you, this looks Perfect :)

Regarding this I have slightly modified. Since now we have the marker which runs exclusively for webhooks, instead of ignoring the error, will be failing if service is not found.
```
	const webhookServiceName = "{{ProjectName}}-webhook-service"

	cmd := exec.Command("kubectl", "get", "service", webhookServiceName, "-n", namespace)
	_, err := utils.Run(cmd)
	if err != nil {
		// Project has no webhook service; nothing to wait on.
		return
	}
```

to

```
      const webhookServiceName = "{{.ProjectName}}-webhook-service"
      
      // Webhook service should exist since webhooks are configured
      cmd := exec.Command("kubectl", "get", "service", webhookServiceName, "-n", namespace)
      _, err := utils.Run(cmd)
      g.Expect(err).NotTo(HaveOccurred(), "Webhook service should exist but was not found")
```

mayuka-c: Regarding this
```
	cmd = exec.Command("kubectl", "get", "--raw",
		fmt.Sprintf("/api/v1/namespaces/%s/services/https:%[2]s:https/proxy/readyz", namespace, 		webhookServiceName))
	_, err = utils.Run(cmd)
	g.Expect(err).NotTo(HaveOccurred(), "Webhook server not responding on port 443")
```

Looks like it is failing after using the service proxy to check for readiness here: https://github.com/mayuka-c/kubebuilder/actions/runs/18962257859/job/54151953303.

I will see on what could be the reason for this. Please do let me know if the endpoint used is wrong. Thanks!

camilamacedo86: If we use a pod with curl to do so is fine but we need to cleanup 
That is why I thought in some other alternative option like.

mayuka-c: Unfortunately I reverted to use curl, it does remove the pod once it finishes with `--rm` flag. Please do let me know if it is fine

camilamacedo86: We need to ensure that all that we create is cleaned.
Use -rm might be a good option if the pod is only removed after the check. 
It seems it might be a good approach I will check it out. 
Thank you a lot for looking on that 🚀 

Well done for sure.

docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_test.go
```diff
@@ -216,6 +216,41 @@ var _ = Describe("Manager", Ordered, func() {
...
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(output).NotTo(BeEmpty(), "Webhook service endpoints are not ready")
+
+				// Test webhook connectivity by checking if webhook server port is responding
+				cmd = exec.Command("kubectl", "run", "webhook-test", "--rm", "-i", "--restart=Never",
+					"--image=curlimages/curl:latest", "--",
+					"curl", "-k", "--connect-timeout", "5",
+					"https://"+webhookServiceName+"."+namespace+".svc:443/readyz")
+				_, err = utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred(), "Webhook server not responding on port 443")
```
camilamacedo86, line 249: - `-n namespace` → ensures the curl pod runs and cleans up in the test namespace.
- `-ksSf` → curl fails if it gets a non-2xx HTTP response, is silent otherwise, and ignores TLS verification errors.
- `--retry 3 --retry-delay 1` → helps handle transient webhook startup delays.
- `Improved error message` → clearer context if the readiness check fails.

```
cmd = exec.Command(
	"kubectl", "run", "webhook-test",
	"--rm", "-i", "--restart=Never",
	"-n", namespace, // ensure the pod runs in the same namespace
	"--image=curlimages/curl:latest", "--",
	"curl", "-ksSf", "--connect-timeout", "5",
	"--retry", "3", "--retry-delay", "1",
	"https://"+webhookServiceName+"."+namespace+".svc:443/readyz",
)
_, err = utils.Run(cmd)
g.Expect(err).NotTo(HaveOccurred(), "Webhook server not responding or /readyz endpoint unhealthy on port 443")
```

docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_test.go
```diff
@@ -216,6 +216,41 @@ var _ = Describe("Manager", Ordered, func() {
...
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(output).To(Equal("True"),
+					"Controller manager pod not ready (webhook server may not be accepting connections)")
+
+				// Check if webhook service endpoints are available
+				cmd = exec.Command("kubectl", "get", "endpoints", webhookServiceName,
+					"-n", namespace, "-o", "jsonpath={.subsets[*].addresses[*].ip}")
+				output, err = utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(output).NotTo(BeEmpty(), "Webhook service endpoints are not ready")
```
camilamacedo86, line 241: I think we should avoid use endpoints.
It seems deprecated

```
	cmd = exec.Command(
		"kubectl", "get", "endpointslices",
		"-l", "kubernetes.io/service-name="+webhookServiceName,
		"-n", namespace,
		"-o", "jsonpath={.items[*].endpoints[?(@.conditions.ready==true)].addresses[*]}",
	)
	output, err := utils.Run(cmd)
	g.Expect(err).NotTo(HaveOccurred(), "Failed to query EndpointSlices for the webhook service")
	g.Expect(strings.TrimSpace(output)).NotTo(BeEmpty(),
		"Webhook service has no ready endpoints (EndpointSlice has zero ready addresses)")

```

docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_test.go
```diff
@@ -216,6 +216,41 @@ var _ = Describe("Manager", Ordered, func() {
...
+				_, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred(), "Webhook service should exist but was not found")
+
+				// Check if webhook server is ready by verifying pod readiness
+				cmd = exec.Command("kubectl", "get", "pods", "-l", "control-plane=controller-manager",
+					"-n", namespace, "-o", "jsonpath={.items[0].status.conditions[?(@.type=='Ready')].status}")
+				output, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(output).To(Equal("True"),
+					"Controller manager pod not ready (webhook server may not be accepting connections)")
```
camilamacedo86, line 234: We could use `--for=condition=Ready`

camilamacedo86:
Hi @mayuka-c 

I was checking it locally and I figure out that we do not need all those checks or the new marker
See: https://github.com/kubernetes-sigs/kubebuilder/pull/5204

I could simplify a lot.

mayuka-c:
Hi @camilamacedo86 👋

Sorry I got busy with during weekdays. So I was about to resolve the comments during the weekend and looks like you have raised a PR with better approach than this. Thank you ☺️.
We could close this once the other PR is merged.

mayuka-c:
Also regarding your below query
```
Then, following some questions:

Do we need to keep // +kubebuilder:scaffold:e2e-webhooks-readiness should not that be replaced by the content?
```

Looks like the markers are getting replaced with content defined in the code fragment but still the marker exists in final TestData after running `generate`. Even for the webhook check I see the marker is still present as seen [here](https://github.com/kubernetes-sigs/kubebuilder/blob/07479d7d6ff54fc9681a701cb673e67665dea1bc/testdata/project-v4-multigroup/test/e2e/e2e_test.go#L318).

Probably I will have look into this. If I'm not wrong the expected behavior is it should not be present right?

PR #5143, opened 2025-10-23
✨ (go/v4): Add allowed files about tools version management files in go init v4

robinlioret:
### Motivation

I'm using mise-en-place to manage my tools versions. When I want to create a new kubebuilder project. I start by creating a `mise.toml`  file with the versions of go and kubebuilder I want to use. Then run the `kubebuilder init ...`  command. Since .toml files are allowed, I get the error 

```
ERROR CLI run failed error=error executing command: failed to initialize project: unable to run pre-scaffold tasks of "base.go.kubebuilder.io/v4": error walking directory: target directory is not empty and contains a disallowed file "mise.toml". files with the following extensions [.go, .yaml, .mod, .sum] are not allowed to avoid conflicts with the tooling
```

I would like to be able to use mise tool stack,  so we need to allow tools management files like mise-en-place and asdf.

### Description of the change

Adding a simple check of the filename against an hardcoded list. It's the same principle as for the extension check but against the whole filename.

### Prerequisites

- [ ] make test-unit : FAILED, even on master I was not able to run the unit tests successfully
- [X] make install
- [X] make lint 
- [X] make check-testdata
- [ ] make test-e2e-local: FAILED, even on master I was not able to run the e2e tests successfully ( unlinkat /usr/bin/kustomize: permission denied, i didn't look further)
- [X] Manual test using the compiled binary (empty dir with mise.toml, then kubebuilder init, ok)

### DIFF
--- a/pkg/plugins/golang/v4/init.go
+++ b/pkg/plugins/golang/v4/init.go
@@ -168,7 +168,7 @@ func checkDir() error {
 			}
 
 			// Only care about top-level files and directories
-			if strings.Count(path, string(os.PathSeparator)) > 1 {
+			if strings.Count(path, string(os.PathSeparator)) > 0 {
 				return nil
 			}
 


pkg/plugins/golang/v4/init.go
```diff
@@ -190,6 +190,17 @@ func checkDir() error {
 			if isCapitalized && info.Name() != "PROJECT" {
 				return nil
 			}
+			allowedFiles := []string{
+				// User might use tool versions management tools to set up the environment including kubebuilder and go version
+				"mise.toml",      // mise-en-place configuration file
+				".tool-versions", // asdf configuration file
```
camilamacedo86: Thank you so much for raising this one 🎉  

Could we check why it’s being blocked instead?  
It seems we shouldn’t block these since they’re not in `disallowedExtensions`.  

If we start creating a full list like that, it could get huge —  
so it might be better to only block the specific required cases.  

What do you think? Could you please take a look? 🙂

robinlioret: You're welcome!

Actually, I think the variable `disallowedExtensions` is misnamed. It should be `allowedExtensions` since it is treated as a guard clause before returning the error.

I agree, init should block the files that are created by kubebuilder and leave the rest be. But, maybe there were a reason why this was implemented in the first place.

I'll dig into the repo to find out why andupdate this PR.

robinlioret:
I reworked the function so it only checks for kubebuilder reserved files and directories.
Since my tests are failing when I run them locally. I tested it manually.

```bash
make install
mkdir testdata/project-v4-invalid-dir && cd testdata/project-v4-invalid-dir

touch .devcontainer Makefile mise.toml .tools-version go.mod
mkdir accepted && touch accepted/Makefile

kubebuilder init --domain=dummy.io --repo=github.com/dummy-io --license=apache2  --plugins go/v4
#>  target directory contains kubebuilder reserved directory or file: ".devcontainer" (.devcontainer, .github, api, bin, cmd, config, dist, hack, internal, test, .dockerignore, .gitignore, .golangci.yml, Dockerfile, Makefile, PROJECT, README.md, grafana)

rm .devcontainer
kubebuilder init --domain=dummy.io --repo=github.com/dummy-io --license=apache2  --plugins go/v4
#> target directory contains kubebuilder reserved directory or file: "Makefile" (...)

rm Makefile
kubebuilder init --domain=dummy.io --repo=github.com/dummy-io --license=apache2  --plugins go/v4
# No error

make generate
# It works

cd ..
rm -rf project-v4-invalid-dir
```

pkg/plugins/golang/v4/init.go
camilamacedo86: Thank you a lot !!!. Sure, the reason is:

We cannot initialize or scaffold files when the target directory already contains files that the tool itself is meant to scaffold. Doing so can lead to conflicts during the command execution — and, even worse, it may overwrite existing files with different content.

To avoid these issues, the initialization process must ensure that the directory is either empty or does not contain files that overlap with those that will be generated by the scaffolding tool.

pkg/plugins/golang/v4/init.go
```diff
@@ -167,45 +166,49 @@ func checkDir() error {
...
 				return nil
 			}
-			disallowedExtensions := []string{
-				".go",
-				".yaml",
-				".mod",
-				".sum",
+
+			// Do not allow kubebuilder reserved directories or files
+			reservedNames := []string{
```
camilamacedo86, line 176: That is definitely an option.
I will need some time to think about it.

camilamacedo86: The main issue here is how to maintain this list.
Moreover: In a multi-repository project, there will typically be a .github directory — and that’s probably fine. However, the tool would need to scaffold files inside that directory, which raises the question of how to manage and keep the list consistent across repositories.

robinlioret:
I understand, I'm not experienced enough with more complex usage of Kubebuilder. This approch is the opposite of the previous one. 

May be it is possible to dynamically build the list from the templates.

camilamacedo86:
Hi @robinlioret 

I think we have not a really good way to prevent the scenario that is accurate and maintainable 
So, I think we should remove it and only check for PROJECT file. At least we prevent try to overwrite a dir that has already in project scaffolded. See; https://github.com/kubernetes-sigs/kubebuilder/pull/5154

robinlioret:
I switched to your PR. Let's close this one.

PR #5154, opened 2025-10-29
🐛 (go/v4): Simplify init directory validation to only block kubebuilder files

camilamacedo86:
Fixes #5143

## Problem
`kubebuilder init` would fail if the target directory contained common 
development tool configuration files like:
- .envrc (direnv)
- mise.toml (mise)
- .tool-versions (asdf)
- go.mod (from running `go mod init` first)

This forced users to work around the validation by moving files out,
running init, then moving them back.

## Solution
Only block initialization if key kubebuilder files already exist:
- PROJECT (kubebuilder config file)
- Makefile  
- cmd/main.go

All other files are allowed. Users get a warning if the directory isn't 
empty, but scaffolding proceeds.

## Result
Users can now initialize kubebuilder projects in directories with existing 
tooling configs without errors.

### DIFF
--- a/pkg/plugins/golang/v4/init.go
+++ b/pkg/plugins/golang/v4/init.go
@@ -22,7 +22,6 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"unicode"
 
 	"github.com/spf13/pflag"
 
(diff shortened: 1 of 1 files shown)


robinlioret:
It looks a lot more permissive that what is in place at the moment. Will kubebuilder override existing files ?
I'm thinking about custom GitHub workflow files or git ignore or any common files.

I don't thinks it'll an issue for experienced people since we just have to look a the git diff and reset the files we want to keep as is. But it may be a good idea to add this hack to the documentation, what do you think ?

camilamacedo86:
/hold 

We need think more about.
Maybe warning and make clear that the project has not a empty dir and files can be overwrite OR conflicts be faced

robinlioret:
I have no idea what amount of work it could represent hut why not adding a step before generating the files ?

Something like:

1. Queue files to generate in a list
2. Check if one of them is going to be overwritten 
3. Display a warning with a interactive choice for the user

This way, we get rid of the need for hack documentation while keeping flexibility and automated maintenance of the kubebuilder "reserved" files.

camilamacedo86:
/hold cancel 

@robinlioret  

Thank you so much for your patience! 🙏  
I've refined the validation logic to make it both **more practical** and **easier to maintain**.

The previous logic was both inverted and too strict—it didn’t reliably prevent actual conflicts and made maintenance difficult.  
By focusing on the few critical files that kubebuilder generates, we:
- Avoid **false positives**
- Keep the logic **clear and maintainable**
- Allow users more flexibility and trust them to manage their own workspace

Would love your thoughts on this updated approach! ✨

pkg/plugins/golang/v4/init.go
```diff
@@ -158,57 +157,70 @@ func (p *initSubcommand) PostScaffold() error {
...
-			}
-			// Allow capitalized files except PROJECT
-			isCapitalized := true
-			for _, l := range info.Name() {
-				if !unicode.IsUpper(l) {
-					isCapitalized = false
-					break
-				}
+			// Skip dot directories
+			if info.IsDir() && strings.HasPrefix(info.Name(), ".") {
```
robinlioret, line 202: Why remove warnings on dot directories?
As a user, I think I would like to be warn if my GitHub workflow was about to be overridden. Or am I missing something?

pkg/plugins/golang/v4/init.go
```diff
@@ -158,57 +157,70 @@ func (p *initSubcommand) PostScaffold() error {
...
-				".mod",
-				".sum",
-			}
-			// Deny files with .go or .yaml or .mod or .sum extensions
-			for _, ext := range disallowedExtensions {
-				if strings.HasSuffix(info.Name(), ext) {
-					return nil
-				}
+			// Track if any non-hidden files/directories exist
+			if !strings.HasPrefix(info.Name(), ".") {
```
robinlioret: Same here

pkg/plugins/golang/v4/init.go
camilamacedo86: Thank you! 🙏 I tried to improve it, but just a heads-up — the warning might show up quite a bit, since `.idea` files (and others) can cause false positives. We have a ton of those.  

Please let me know if you spot any important cases that we should actually handle.  

**Examples:**  
- `.github/workflows/ci.yml` → ⚠️ warning shown  
- `.git`, `.vscode`, `.idea` → ✅ skipped (no warning)  
- Only `.gitignore`, `go.mod`, `go.sum` → ✅ ignored (no warning)

robinlioret: That's what I was missing ! I forgot that the code doesn't check only for kubebuilder files !

camilamacedo86: Hi @robinlioret 

WDYT?

Could we win your LGTM now to get this one merged?

robinlioret: Well. I thinks we should keep the warnings for all the files including the dot files. I feel that skipping them may confuse users.

Imagine the following

1. The user tries to initialize in a repo with a .GitHub dir
2. They receive a set of warnings, they assumes that the list is complete.
3. They accept the warning and get a replaced .github.

If the user doesn't check their git diff, they'll miss the  change and could cause issues later on.

This makes sense because nowadays, dot files have very important roles.

camilamacedo86: .github/workflows/ci.yml seems like the only place where this actually makes
sense. For the other directories, raising a warning doesn’t seem useful and
would just create noise. If we warn on everything, we end up training people
to ignore warnings entirely, which isn’t great practice. It’s better to only
warn when it’s meaningful.

robinlioret: You're right. The best option would be to add a preparation step that list the files about to be created by kubebuilder and check against this list. Maybe such a list could be statically created at build time?

For the time being, I don't know if it is worth the effort. So the solution in this PR could be a quick win. Let's merge!

PR #5169, opened 2025-11-02
🌱 Increase unit test coverage

camilamacedo86:
Related to: https://github.com/kubernetes-sigs/kubebuilder/issues/4925


camilamacedo86:
Add tests only if pass we can move forward.

PR #5171, opened 2025-11-02
✨ Add support for custom webhook paths

camilamacedo86:
Users can now set custom paths for webhooks using --defaulting-path and --validation-path flags. Requires controller-runtime v0.21+.

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/4295


camilamacedo86:
Hey @damsien 

Could you please give a hand in this one?
Since you request it I think you will be the best one to help us with reviews :-) 
WDYT?

camilamacedo86:
c/c @sbueringer

damsien:
Hello @camilamacedo86 , thanks for reaching me!

I tested your implementation and it works well! However, I noticed that you didn't implemented an regex checker for the path. I don't know if it is supposed to not being implemented.

If we need to add the regex validation, I would like to contribute to your PR :)

camilamacedo86:
Hi @Daniem,

> Hello @camilamacedo86 , thanks for reaching me!
> I tested your implementation and it works well! However, I noticed that you didn't implemented an regex checker for the path. I don't know if it is supposed to not being implemented.
>
> If we need to add the regex validation, I would like to contribute to your PR :)


Thank you for raising and validating this — great work! 🥇 

It seems that from the Kubernetes side, any path is accepted, so I’m not entirely sure what kind of validations we could enforce.

Our convention:
`/mutate-<group>-<version>-<kind> and /validate-<group>-<version>-<kind>`
(with dots in the group replaced by dashes and all lowercase).

That said, since this is already the default behavior, suggesting it might be redundant — so I’m not completely sure it’s worth enforcing.

Given that you’ve verified everything is implemented accurately, I’m happy for us to move forward — great job! 🚀
And please feel free to open any follow-up PRs for improvements you think would be valuable. Your collaboration is very welcome! 🙌

damsien:
To be more precise, I was talking about a regex that checks if the path is a valid path (no weird characters, etc...). The regex validation is already done in controller-runtime: https://github.com/kubernetes-sigs/controller-runtime/blob/67242822fa14d36e8ef7a5f7933022334b11a976/pkg/builder/webhook.go#L374-L383.

Since controller-runtime already implement it, if the user put a wrong path in the kubebuilder command, then he gonna have an error on runtime. So it is not strictly necessary to add the regex validation on the kubebuilder side but still it can be useful :)

PR #5172, opened 2025-11-02
✨ (helm/v2-alpha): Make webhook and metrics ports configurable

camilamacedo86:
- Add configurable webhook.port (default: 9443) and metrics.port (default: 8443) in values.yaml
- Add comprehensive unit tests for port configuration with default and custom values

Closes: https://github.com/kubernetes-sigs/kubebuilder/pull/5172

### DIFF
--- a/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
+++ b/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
@@ -38,6 +38,7 @@ const (
 	kindIssuer             = "Issuer"
 	kindValidatingWebhook  = "ValidatingWebhookConfiguration"
 	kindMutatingWebhook    = "MutatingWebhookConfiguration"
+	kindDeployment         = "Deployment"
 
 	// API versions
 	apiVersionCertManager = "cert-manager.io/v1"
@@ -79,7 +80,7 @@ func (t *HelmTemplater) ApplyHelmSubstitutions(yamlContent string, resource *uns
 	yamlContent = t.substituteRBACValues(yamlContent)
 
 	// Apply deployment-specific templating
-	if resource.GetKind() == "Deployment" {
+	if resource.GetKind() == kindDeployment {
 		yamlContent = t.templateDeploymentFields(yamlContent)
 
 		// Apply conditional logic for cert-manager related fields in deployments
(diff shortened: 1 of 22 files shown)


camilamacedo86:
@defo89 the helm/v2alpha was not released yet
I think it should be part of the initial implementation as well.

So, your request made in:  https://github.com/kubernetes-sigs/kubebuilder/issues/4802 can be closed with those changes to address it with helm/v2alpha

Please, feel free to help us in the review

camilamacedo86:
Hi @liam-mackie 

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
```diff
@@ -919,3 +925,60 @@ func (t *HelmTemplater) collapseBlankLineAfterIf(yamlContent string) string {
...
+		// Check if this deployment has webhook-server ports
+		if strings.Contains(yamlContent, "webhook-server") || strings.Contains(yamlContent, "name: webhook") {
+			isWebhook = true
+		}
+	}
+
+	// Template webhook ports (9443 by default)
+	if isWebhook {
+		// Replace containerPort: 9443 (or any value) for webhook-server with template
+		if strings.Contains(yamlContent, "webhook-server") {
```
liam-mackie, line 951: nit: should we consider using a similar pattern to [the unstructured package](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/unstructured/helpers.go#L326-L332) and trying to structure the data rather than string matching?

liam-mackie (approved):
Amazing work! Single nit on whether to continue string matching or start marshalling into structures we can traverse programatically a little easier.

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
camilamacedo86: I think we could try that, but wouldn’t that apply to whole plugin?
What do you think? Should we check in a follow-up to see if we could improve this horizontally?

liam-mackie: Yeah - I think that's something we can improve anywhere in the package with some neat generic functions. Once I'm back from Kubecon I'll likely have some more time to contribute, and I'll start implementation. This PR is fine as-is and if we decide to go with a more structured approach, we can refactor the project as a whole :)

PR #5173, opened 2025-11-02
🐛 (helm/v2-alpha): ensure that certmanager option is exposed always

camilamacedo86:
It can either be used by the metrics so should be always exposed


camilamacedo86:
Hi @liam-mackie

PR #5174, opened 2025-11-02
🐛 (helm/v2alpha) Use manager instead of controllerManager to expose manager values

camilamacedo86:
controllerManager does not seems the best name for we expose the manager values
This PR simplifies in favor of manager


camilamacedo86:
Hi @liam-mackie

liam-mackie (approved):
Looks good! I've always been a little annoyed that the resources are called `controller-manager` but within the resources, the container image etc. is usually just `manager`. This is a great step int he right direction!

PR #5175, opened 2025-11-02
📖 Improve collapsible code sections and formatting of tutorials

camilamacedo86:
The following changes impact mainly the tutorials.
To review see:
- https://deploy-preview-5175--kubebuilder.netlify.app/cronjob-tutorial/cronjob-tutorial
- https://deploy-preview-5175--kubebuilder.netlify.app/multiversion-tutorial/tutorial

### Changes

**Content Cleanup:**
- Hidden Apache License and Import blocks
- Hidden closing braces in tests
- Removed duplicate collapse markers
- Fixed extra `}` appearing outside code blocks

**Label Updates:**
- "Clock" → "Clock Code Implementation"
- "Go imports" → "Imports"
- "Validate object name" → "validateCronJobName() Code Implementation"
- "old stuff" → "Remaining code from <filename>"
- "The rest of Spec" → "CronJobSpec Full Code"

**Related:** #4862

Issue #4862, opened 2025-06-10
Improve collapsible code sections and formatting on book.kubebuilder.io


danielskowronski:
I'd like to report a few issues with collapsible codeblock sections in [book.kubebuilder.io](https://book.kubebuilder.io/) that have the potential to confuse readers, especially new users.

1. **Hidden instructional text**: initially collapsed sections that appear to only contain code, actually contain instructions as well (https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation - "Validate object name")
2. **Misleading headers in code blocks**: collapsible codeblocks have headers formatted as Golang comments and use font only slightly larger than the body of codeblock, which suggests that they are actual comments in code
3. **Insufficient separation between blocks**: consecutive code blocks (collapsed by default or not collapsible) appear as single code block, which makes it difficult to follow (https://book.kubebuilder.io/cronjob-tutorial/controller-implementation - especially "5: Get the next scheduled run")
4. **Lack of highlighting for expanded sections**: there's no visual indication of the section that was expanded, which is more problematic on very long pages
5. **Verbose license notices**: having "Apache License" block does not contribute to the guide (but it's understandably part of the source file) and may confuse users because licence text is shown as instruction text, not codeblock
6. **Overly long lines and narrow layout**: HTML template makes a hard limit on text (and codeblock) width, but there's no limit to line length in code; I could easily find a line with 390 characters, but it seems like only 89 characters fit into codeblock without horizontal scrolling
7. **Lack of proper formatting of shell commands**: 
    1. none of shell commands use syntax highlighting
    2. some commands are presented as very long lines, making them hard to copy
    3. some commands are formatted like Golang comments (`// `)
8. **Internal headers missing from TOC**: if headers are originating from Golang comments, they are not included in TOC and are hard to distinguish from the main body; https://book.kubebuilder.io/cronjob-tutorial/controller-implementation 
9. **Code of function split into sections is unclear**: when function body is split into multiple parts, it's not clear that the body continues

Many of those issues stack and increase cognitive load even more. They are present since v2, but v1 has significantly different contents to compare. I was able to figure out how to read the book only after spending an hour preparing this bug report. 

I'd like to offer some improvement suggestions, but I'm not sure whether problems originate from comments in Golang code, tooling used (`litgo.sh`, mdBook) or HTML templates.

- consider disabling collapsible sections (quickest workaround, but it's already a book)
- clearly highlight codeblock boundaries - add borders and padding on bottom
[...]

PR #5183, opened 2025-11-09
🐛 fix(docs): display license as code block

vitorfloriano:
This PR fixes the case in which the license would be extracted and displayed as normal text. Now the license is rendered as a code block. 

Related to #4862 and #5175.

### Before:

image

### After:

image

### DIFF
--- a/docs/book/utils/litgo/literate.go
+++ b/docs/book/utils/litgo/literate.go
@@ -138,11 +138,6 @@ func isBlockComment(tok token.Token, lit string) bool {
 		return false
 	}
 
-	// skip if it's a license comment
-	if strings.Contains(lit, "Copyright") || strings.Contains(lit, "Licensed under") {
-		return false
-	}
-
 	return true
 }
 
@@ -189,6 +184,16 @@ func extractPairs(contents []byte, path string) ([]commentCodePair, error) {
 		var pos token.Pos
 		var lit string
 		pos, tok, lit = scan.Scan()
+
+		// skip license comment on headers
+		if tok == token.COMMENT && strings.HasPrefix(lit, "/*") {
+			lower := strings.ToLower(lit)
+			if strings.Contains(lower, "copyright") && strings.Contains(lower, "license") {
+				lastCodeBlockStart = file.Offset(pos) + len(lit)
+				continue
+			}
+		}
+
 		collapse := getCollapse(tok, lit)
 		if collapse != "" {
 			lastPair.collapse = collapse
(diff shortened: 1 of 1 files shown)


camilamacedo86:
The appache license I think we must to hidden 100%
That does not bring any value in the docs.

docs/book/utils/litgo/literate.go
```diff
@@ -138,6 +138,11 @@ func isBlockComment(tok token.Token, lit string) bool {
 		return false
 	}
 
+	// skip if it's a license comment
+	if strings.Contains(lit, "Copyright") || strings.Contains(lit, "Licensed under") {
```
camilamacedo86: We need to hidden that 100%

vitorfloriano: @camilamacedo86 I pushed a change in the logic. We first check if the code block is a license so we don't extract it, then we check if the collapsible is empty and skip it, then we put the source citation on top of the first code block that is apparent: 
image

WDYT?

docs/book/utils/litgo/literate.go
```diff
@@ -138,11 +138,6 @@ func isBlockComment(tok token.Token, lit string) bool {
 		return false
```
camilamacedo86, line 138: Could we not do as in the other pr https://github.com/kubernetes-sigs/kubebuilder/pull/5175

Image

Image

WDYT?

vitorfloriano: Wouldn't this hide some collapsibles that are meant to stay? For example, we have some collapsibles in [here](https://kubebuilder.io/cronjob-tutorial/basic-project) and in [here](https://kubebuilder.io/cronjob-tutorial/main-revisited) that we should keep...

vitorfloriano:
Honestly, @camilamacedo86 I think we should rethink the whole literate extraction. It is basically trying to replicate what mdBook already does natively. 

IMHO, the literate extraction system got overly complex, perhaps due to clever solutions (hacks) to work around past limitations in mdBook.

I'm planning on writing a design proposal to change how we extract code from the testdata samples using only mdBook's `{{#include filename:lines}}` feature and keep all the tutorial instructions in markdown files. That would make it much easier for contributors to suggest improvements to the docs, as well as to identify changes in the test samples.

mdBook newer versions have some nice features that we can't use because our custom preprocessors (literate.go and markerdocs.go) are not compatible with them. We should strive towards moving away from them and rely on mdBook's features. They are more than sufficient for our usecase and would greatly reduce the overhead to maintain the docs updated, IMO.

Due to the gargantuan size of this refactor, it could even be a project for a new mentorship! 🤩 

How does that sound to you?

docs/book/utils/litgo/literate.go
camilamacedo86: All imports and license ones bring no value at all
That is why they are collapsed.
The workaround was collapse them out to hidden but we can hidden 100%

vitorfloriano: @camilamacedo86 I double-checked #5175 and it is hiding the collapsed licenses and imports. I'm closing this PR as duplicate.

PR #5184, opened 2025-11-09
📖 : fix(docs): display copy button on codeblocks

vitorfloriano:
This PR fixes the issue in which the copy button would not be displayed in code blocks generated by literate-go. 

Related to #4862 and #5175.

### Before:

image

### After:

image

### DIFF
--- a/docs/book/theme/css/markers.css
+++ b/docs/book/theme/css/markers.css
@@ -365,11 +365,6 @@ cite.literate-source > a::before {
     color: var(--fg);
 }
 
-/* hide the annoying "copy to clipboard" buttons */
-.literate pre > .buttons {
-    display: none;
-}
-
 /* add a bit of extra padding for readability */
 .literate pre code {
     padding-top: 0.75em;


camilamacedo86:
After your review I was looking and deliberate those were skiped 
So, I am wondering if we should really display them.
Maybe we need to check if they are working well in all places
WDYT?

vitorfloriano:
> After your review I was looking and deliberate those were skiped So, I am wondering if we should really display them. Maybe we need to check if they are working well in all places WDYT?

@camilamacedo86 I think it was removed as a matter of **personal taste**. But IMO we should keep them. Pretty much every website that displays code blocks have a copy button in them, so it's basically a standard of UX.

camilamacedo86:
@vitorfloriano 

OK. So, lets get merged and if someone complain then we revisit it.
I think you made a fair point. 

/approved
/lgtm

camilamacedo86 (approved):
/lgtm
/approved

/ok-to-test

PR #5186, opened 2025-11-09
🐛 Downgrade Go to 1.24.6 and standardize version across codebase

camilamacedo86:
### Summary
This PR downgrades the Go version to **1.24.6** and ensures consistent usage across all modules and build configurations.

### Motivation
- **Security:** Go **1.24.6** includes important fixes that address a known CVE affecting earlier patch releases in the 1.24 series.  
- **Compatibility:** The latest versions of **Kubernetes APIs**, **controller-runtime**, and **controller-tools** are still built and tested against Go **1.24.x**, so maintaining this version ensures full compatibility and avoids potential build or runtime issues.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kubebuilder/v4
 
-go 1.25.0
+go 1.24.6
 
 require (
 	github.com/gobuffalo/flect v1.0.3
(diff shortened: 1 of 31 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kubebuilder/v4
 
-go 1.25.0
+go 1.24.6
```
camilamacedo86, line 3: We upgrade the version for the tooling to workaround that was properly addressed with: https://github.com/kubernetes-sigs/kubebuilder/pull/5194

In order to push a release with a go version that address CVE and still compatible with the libs adopt we should stick with 1.24.6. Let;s just upgrade to 1.25 when we bump controller-runtime using it.

PR #5204, opened 2025-11-14
🐛 (go/v4):  Fix flaky metrics e2e tests when webhooks are scaffolded

camilamacedo86:
## Problem

When the default e2e suite is run against a project that scaffolds admission
webhooks, the "metrics" test can fail with errors such as:

```
Error from server (InternalError): Internal error occurred: failed calling webhook:
Post "https://...webhook-service:443/validate--v1-pod": dial tcp ...:443:
connect: connection refused
```

The failure happens when the test launches the `curl-metrics` pod to verify the
metrics endpoint.

## Root Cause

The original test assumed that, once the controller pod reported `Running`, the
webhook server was also ready to admit new pods. In reality:

1. The controller pod reaches `Running`.
2. cert-manager is still issuing the serving certificate secret for the webhook.
3. The webhook server only starts once that secret is created and mounted.
4. The metrics e2e test creates the curl pod while the webhook server is still initialising.
5. The validating webhook rejects the pod because it cannot accept connections yet.

Checking the controller logs for "Serving metrics server" does not guarantee the webhook server is listening, so the race persisted.

## Solution

Gate the curl pod creation on two readiness checks:

1. Wait for the controller pod to report the `Ready` condition.
2. When webhooks are scaffolded, wait for the webhook service to publish EndpointSlice addresses (the Kubernetes-recommended readiness signal).

Controller readiness:

```go
By("ensuring the controller pod is ready")
verifyControllerPodReady := func(g Gomega) {
    cmd := exec.Command("kubectl", "get", "pod", controllerPodName, "-n", namespace,
        "-o", "jsonpath={.status.conditions[?(@.type=='Ready')].status}")
    output, err := utils.Run(cmd)
    g.Expect(err).NotTo(HaveOccurred())
    g.Expect(output).To(Equal("True"), "Controller pod not ready")
}
Eventually(verifyControllerPodReady, 3*time.Minute, time.Second).Should(Succeed())
```

Webhook readiness (injected only when the project wires webhooks):

```go
verifyWebhookEndpointsReady := func(g Gomega) {
    cmd := exec.Command("kubectl", "get", "endpointslices.discovery.k8s.io", "-n", namespace,
        "-l", "kubernetes.io/service-name=%s",
        "-o", "jsonpath={range .items[*]}{range .endpoints[*]}{.addresses[*]}{end}{end}")
    output, err := utils.Run(cmd)
    g.Expect(err).NotTo(HaveOccurred(), "Webhook endpoints should exist")
    g.Expect(output).ShouldNot(BeEmpty(), "Webhook endpoints not yet ready")
}
Eventually(verifyWebhookEndpointsReady, 3*time.Minute, time.Second).Should(Succeed())
```

A new scaffold marker (`+kubebuilder:scaffold:e2e-metrics-webhooks-readiness`) adds the EndpointSlice check only for webhook-enabled projects, keeping the base template generic.

## Why this works

- The `Ready` condition requires all containers, volumes (including the webhook
  cert secret), and readiness probes to succeed, so the controller and its webhook server have finished initialising.
- EndpointSlices provide the modern, recommended signal that a service has routable endpoints, avoiding deprecated `Endpoints` objects.
- The previous log-based verification that the metrics server is serving remains in place, and the curl pod still validates the metrics output once the checks pass.

## Notes on Kubernetes 1.33+

Kubernetes 1.33+ may exhibit a brief delay before the metrics endpoint becomes
available when controller-runtime's `WithAuthenticationAndAuthorization()` is
used with self-signed certificates. The readiness sequence above absorbs that
startup delay, so the curl pod is only launched after Kubernetes itself signals
that both the controller pod and webhook service are ready.

Closes: https://github.com/kubernetes-sigs/kubebuilder/pull/5138 https://github.com/kubernetes-sigs/kubebuilder/issues/5137

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_test.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_test.go
@@ -197,24 +197,38 @@ var _ = Describe("Manager", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(token).NotTo(BeEmpty())
 
-			By("waiting for the metrics endpoint to be ready")
-			verifyMetricsEndpointReady := func(g Gomega) {
-				cmd := exec.Command("kubectl", "get", "endpoints", metricsServiceName, "-n", namespace)
+			By("ensuring the controller pod is ready")
+			verifyControllerPodReady := func(g Gomega) {
+				cmd := exec.Command("kubectl", "get", "pod", controllerPodName, "-n", namespace,
+					"-o", "jsonpath={.status.conditions[?(@.type=='Ready')].status}")
 				output, err := utils.Run(cmd)
 				g.Expect(err).NotTo(HaveOccurred())
-				g.Expect(output).To(ContainSubstring("8443"), "Metrics endpoint is not ready")
+				g.Expect(output).To(Equal("True"), "Controller pod not ready")
 			}
-			Eventually(verifyMetricsEndpointReady).Should(Succeed())
+			Eventually(verifyControllerPodReady, 3*time.Minute, time.Second).Should(Succeed())
 
 			By("verifying that the controller manager is serving the metrics server")
 			verifyMetricsServerStarted := func(g Gomega) {
 				cmd := exec.Command("kubectl", "logs", controllerPodName, "-n", namespace)
 				output, err := utils.Run(cmd)
 				g.Expect(err).NotTo(HaveOccurred())
-				g.Expect(output).To(ContainSubstring("controller-runtime.metrics\tServing metrics server"),
+				g.Expect(output).To(ContainSubstring("Serving metrics server"),
 					"Metrics server not yet started")
 			}
-			Eventually(verifyMetricsServerStarted).Should(Succeed())
+			Eventually(verifyMetricsServerStarted, 3*time.Minute, time.Second).Should(Succeed())
+
+			By("waiting for the webhook service endpoints to be ready")
+			verifyWebhookEndpointsReady := func(g Gomega) {
+				cmd := exec.Command("kubectl", "get", "endpointslices.discovery.k8s.io", "-n", namespace,
+					"-l", "kubernetes.io/service-name=project-webhook-service",
+					"-o", "jsonpath={range .items[*]}{range .endpoints[*]}{.addresses[*]}{end}{end}")
+				output, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred(), "Webhook endpoints should exist")
+				g.Expect(output).ShouldNot(BeEmpty(), "Webhook endpoints not yet ready")
+			}
+			Eventually(verifyWebhookEndpointsReady, 3*time.Minute, time.Second).Should(Succeed())
+
+			// +kubebuilder:scaffold:e2e-metrics-webhooks-readiness
 
 			By("creating the curl-metrics pod to access the metrics endpoint")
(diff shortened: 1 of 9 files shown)


camilamacedo86:
Hi @mayuka-c 

Could you help review this one?

mayuka-c (approved):
/lgtm

This looks good to me since the controller pod readiness guarantees that the webhook server is ready. The approach which I had taken directly checks for webhook readiness which has additional overhead of defining and maintaining markers so that the webhook readiness checks only runs on PROJECT which has webhooks configured.

Thank you so much @camilamacedo86 :)

docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_test.go
```diff
@@ -197,24 +197,38 @@ var _ = Describe("Manager", Ordered, func() {
...
+			By("waiting for the webhook service endpoints to be ready")
+			verifyWebhookEndpointsReady := func(g Gomega) {
+				cmd := exec.Command("kubectl", "get", "endpointslices.discovery.k8s.io", "-n", namespace,
+					"-l", "kubernetes.io/service-name=project-webhook-service",
+					"-o", "jsonpath={range .items[*]}{range .endpoints[*]}{.addresses[*]}{end}{end}")
+				output, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred(), "Webhook endpoints should exist")
+				g.Expect(output).ShouldNot(BeEmpty(), "Webhook endpoints not yet ready")
+			}
+			Eventually(verifyWebhookEndpointsReady, 3*time.Minute, time.Second).Should(Succeed())
```
camilamacedo86, line 229: @mayuka-c that is the code that we must to inject.
Without it we still risking the flakes

**Why?**

Without waiting for the webhook service to publish endpoints, the curl pod can hit the validating webhook while it’s still initializing, producing the same “connection refused” failure. The EndpointSlice check is what gives us a deterministic signal that the webhook server is actually accepting traffic before we launch the curl metrics pod.

camilamacedo86:
Hi @mayuka-c 

Sorry, I changed it to do more tests and forgot to clarify.
I tried many options. The final solution is now in place. 
Please, feel free to check

docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_test.go
mayuka-c: Yes this would be required. I mean the approach which was taken [here](https://github.com/kubernetes-sigs/kubebuilder/pull/5138) (To have the marker to check for webhook readiness).

Sorry, if I caused some misunderstandings here :)

mayuka-c:
> Hi @mayuka-c
> 
> Sorry, I changed it to do more tests and forgot to clarify. I tried many options. The final solution is now in place. Please, feel free to check

Yeah this looks perfect for me. Once this is merged, I will be closing this PR [here](https://github.com/kubernetes-sigs/kubebuilder/pull/5138).
Thanks :)

docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_test.go
camilamacedo86: Not at all — you helped a lot!
You identified the issue, and your effort was absolutely not lost.
I’ve added you as a co-author of this PR to give you proper credit for your contribution. 🙌


Screenshot 2025-11-15 at 16 37 52

PR #5207, opened 2025-11-14
📖 Migrate to mdBook v0.5.0

vitorfloriano:
This PR migrates the Kubebuilder book from mdBook v0.4.40 to v0.5.0.

Since this is a breaking change, the custom preprocessors (literatego and markerdocs) had to be slightly adjusted. The mdBook parser also got stricter and some fixes were necessary. See the [mdBook v0.5 migration guide](https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#05-migration-guide).

mdBook v0.5.0 introduces fixes and a some nice features that could improve the book DX and UX. To name a few:

- Header navigation in the Table of Contents ("On this page")
- The TOC now follows and highlights the current page
- Support for definition lists in markdown (could be useful for the markers section)
- Support for admonitions in markdown (could be useful for highlighting important info, like tips, warnings, notes, etc)

Related to #4862


docs/book/book.toml
```diff
@@ -15,9 +15,3 @@ command = "./litgo.sh"
 
 [preprocessor.markerdocs]
 command = "./markerdocs.sh"
-
-[context.environment]
-  environment = { GO_VERSION = "1.23" }
-
-[context.deploy-preview.environment]
-  environment = { GO_VERSION = "1.23" }
```
camilamacedo86, line 25: We do not need to set the go version? 
if not set, will it be the latest?

vitorfloriano: mdBook v0.5.0 now throwns an error if a config field is unknown. See ["config changes"](https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#config-changes).

If you take a look at the [mdBook config file documentation](https://rust-lang.github.io/mdBook/format/configuration/general.html), you won't find any [context] config stanzas, so they are unknow to mdBook.

This is confirmed when trying to build the book with those config values:

```bash
:~/go/src/github.com/vitorfloriano/kubebuilder/docs/book$ mdbook build
ERROR Invalid configuration file
	Caused by: TOML parse error at line 19, column 2
   |
19 | [context.environment]
   |  ^^^^^^^
unknown field `context`, expected one of `book`, `build`, `rust`, `output`, `preprocessor`
```
I tried using `build` instead of `context` but got the same error:
```bash
:~/go/src/github.com/vitorfloriano/kubebuilder/docs/book$ mdbook build
ERROR Invalid configuration file
	Caused by: TOML parse error at line 19, column 8
   |
19 | [build.environment]
   |        ^^^^^^^^^^^
unknown field `environment`, expected one of `build-dir`, `create-missing`, `use-default-preprocessors`, `extra-watch-dirs`
```
So, it seems that these config values were not being used. At least not from here. 

I believe we set the Go version in the [Netlify config file](https://github.com/kubernetes-sigs/kubebuilder/blob/master/netlify.toml#L3), before calling the install script.

camilamacedo86: We have here as well: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/utils/go.mod
We might want to bump all for 1.24.6
So that when we go to 1.25.0 those will be easily found.

camilamacedo86: https://github.com/kubernetes-sigs/kubebuilder/pull/5209

vitorfloriano: @camilamacedo86 I bumped the Go version to `1.24.6` in this branch. 

Sorry for not telling you. The deployment succeeded.

vitorfloriano:
@camilamacedo86 Now that mdBook v0.5.0 is out, I think this one is ready to ship. 

I believe the deployment is failing right now due to another [Cloudflare global network issue](https://www.cloudflarestatus.com/), so I guess we need to wait and see what happens when it's fixed.

camilamacedo86:
Hi @vitorfloriano 

That seems amazing 🎉 
That only thing that is not playing nice for us is all sub-sections be render in the menu ie:


Screenshot 2025-11-18 at 15 01 40

Is possible do not have that?
WDYT?

docs/book/book.toml
vitorfloriano, line 14: @camilamacedo86 Unfortunately, it's not possible to not fold deeply nested sections like those. See [this comment from the maintainer](https://github.com/rust-lang/mdBook/pull/2822#issuecomment-3229782236).

I could, though, include more foldable sections, like this:

image

I could also make them foldable but open by default, but the deeply nested sections would still be folded:

image

This is simply a matter of config:

```suggestion
[output.html.fold]
enable = true
level = 5    # 0 would be all folded
```

camilamacedo86: Could we not `enable = false` to not have this behaviour?

vitorfloriano: > Could we not `enable = false` to not have this behaviour?

@camilamacedo86 This config defaults to `false`. That's how it currently is. But the deep nested headers will always be folded, no matter the config.

camilamacedo86: If you find any way of reduce that, I think it would be great already

camilamacedo86: That is really bad :-( 

Screenshot 2025-11-18 at 18 00 47

camilamacedo86: very confusing :-(

vitorfloriano: @camilamacedo86 I tweaked some CSS classes. See if you like it.

vitorfloriano: The deployment is not showing the changes in the last commit...🤔 Perhaps it's showing a cached version of the website?

vitorfloriano: With the latest tweaks I was able to show all sections in my localhost, but somehow the deploy preview is not updating...

image

camilamacedo86:
Hi @vitorfloriano 

I will need spend some time to understand if has something that we could change in docs itself that would fit better on this new 5.0 one. It would be nice raise an issue for the maintainer asking if they cannot create an option to disable that

vitorfloriano:
@camilamacedo86 Could you verify again? I deleted all customizations and applied them again and now the preview seems to have refreshed.

camilamacedo86:
Hi @vitorfloriano 

I am seem it with the same thing in the bar
I created: https://github.com/rust-lang/mdBook/issues/2949

vitorfloriano:
@camilamacedo86 That's weird... 🤔  just a heads up: I had to clear the cache on my browser to see the changes. 

See how the deploy looks for me now, with no folding at all:

image

camilamacedo86:
Yep. the navegation is wirdo in many scenarios it is too confuse for us. 
Example: https://github.com/rust-lang/mdBook/issues/2949

So, if that is possible I will close the issue there.

vitorfloriano:
@camilamacedo86 I have disabled the header navigation. PTAL 👀

Issue #5211, opened 2025-11-15
go install: malformed file path ".github/*.instructions.md": invalid char '*'


chappjc:
### What broke? What's expected?

Following the tagging of `v4.10.0` today, any go get / download fails complaining about `malformed file path ".github/*.instructions.md": invalid char '*'`

### Reproducing this issue

With Go 1.25.3

```
$ GOPROXY=direct GOSUMDB=off go mod download sigs.k8s.io/kubebuilder/v4@v4.10.0
go: create zip: .github/*.instructions.md: malformed file path ".github/*.instructions.md": invalid char '*'
```

I think https://github.com/kubernetes-sigs/kubebuilder/commit/f3d4ca02f1f3ae732194774fb7d75aede6c0d303 violated some golang rule

### KubeBuilder (CLI) Version

v4.10.0

### Plugin versions

```yaml

```

camilamacedo86:
Thank you for raising it .
Lets fix asap and do a path release.

PR #5213, opened 2025-11-15
🐛 Fix Go module download failure caused by invalid filename

camilamacedo86:
Fix issue where `go get sigs.k8s.io/kubebuilder/v4@v4.10.0` failed with: "malformed file path .github/*.instructions.md: invalid char '*'"

Go modules do not allow wildcard characters (*, ?, [, ], \) in filenames when creating module archives.

Changes:
- Remove .github/*.instructions.md (invalid filename with asterisk)
- Add .github/instructions/kubebuilder.instructions.md (proper location)
- Add test/check-gomod.sh to validate Go module compatibility
- Update Makefile to include test-gomod in CI test suite

This test will prevent similar issues in future releases by catching invalid filenames during development.

Fixes the issue introduced in f3d4ca0

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/5211

### DIFF
--- a/test/check-gomod.sh
+++ b/test/check-gomod.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+# Copyright 2025 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+echo "Checking for Go module compatibility (invalid filename characters)..."
+
+# Go modules don't allow certain characters in file paths when creating archives
+# Reference: https://go.dev/ref/mod#zip-files
+# Invalid characters include: *, ?, [, ], \, and others that are shell wildcards or special characters
+
+# Get all tracked files in git (this will include all files that would be part of the go module)
+allfiles=$(git ls-files)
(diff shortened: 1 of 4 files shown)


camilamacedo86:
Hi @chappjc 

Could you give a hand on the review?

camilamacedo86:
Moving forward to allow us do a PATCH release and have the issue fixed for those who want to install the latest version with go module.

chappjc:
Thanks for the quick fix.

However, I suspect kubebuilder will need a `retract v4.10.0` in its go.mod.  The issue with the invalid file name seems to have secondary effects on some deprecation checks in the go tooling.  In a different environment, a `go install` of **`v4.9.0`** caused the following:

```
go: sigs.k8s.io/kubebuilder/v4@v4.9.0: loading deprecation for sigs.k8s.io/kubebuilder/v4: 
sigs.k8s.io/kubebuilder/v4@v4.10.0: reading <a_goproxy>sigs.k8s.io/kubebuilder/v4/@v/v4.10.0.info: 404
```

only by unsetting `GOSUMDB=off` can we work around this.

I'm hoping that the issue will go away if `v4.10.0` is retracted and not checked by whatever deprecation checks are failing here.

test/check-gomod.sh
```diff
@@ -0,0 +1,53 @@
...
+allfiles=$(git ls-files)
+invalidFiles=""
+
+for file in $allfiles; do
+  # Check if the filename (not the full path) contains invalid characters
+  filename=$(basename "$file")
+  
+  # Check for wildcard characters that are invalid in Go module archives
+  # Note: We check for *, ?, [, ], and \ (backslash)
+  if echo "$filename" | grep -qE '[*?[]|\\'; then
```
chappjc, line 37: My first time seeing this code and these rules, but the golang.org/x/mod/module code has a slightly different set: 

https://cs.opensource.google/go/x/mod/+/refs/tags/v0.30.0:module/module.go;l=282

camilamacedo86: Thank you a lot.
I made it very robust now: https://github.com/kubernetes-sigs/kubebuilder/pull/5217

Issue #5208, opened 2025-11-14
Document `WithFieldValidation`


guettli:
### What do you want to happen?

Usually there is a tight 1:1 relationship between the controllercontainer image version of the controller and the corresponding CRD which the controller owns.

But sometimes things fail, and the controller is newer/older than its CRD spec in the api-server.

This can lead to hard to debug issues. For example the apiVersion was not changed, but a new optional field was added.

What about adding a hint to [client.WithFieldValidation()](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#WithFieldValidation) in kubebuilder?

Some days ago I had a very strange behavior or my controller because the CRD resource was outdated while the controller expected the new schema.

Currently, `err := crClient.Patch(...)` does not return an error if the Patch (or Update) contains unknown fields. There is a hardly noticeable log on INFO level:

> 07:50:39 INFO  "unknown field "status.newField"" log/warning_handler.go:65 

It took me some time to understand the issue: api had changed slightly (without changing the version number), and CRD and controller used different schemas.

After switching to a controller-runtime client with field validation, you get an error if the api-server provides warnings.

This means a line like `err := crClient.Patch(...)` will result in an error. This error is better because it signals that there is a version mismatch between api-server and controller.

---

But how could that be integrated?

The change to my controller looks like this: [:seedling: Use strict schema validation  syself/cluster-api-provider-hetzner](https://github.com/syself/cluster-api-provider-hetzner/commit/e55c90a0bd7f7c0d6369979a40732798a2f5a83b)

In the kubebuilder git repo there are 34 lines like this: `mgr, err := ctrl.NewManager(...)`.

Changing all these lines to use a strict field validating client makes no sense from my perspective.

We could add a note. But where would be a good place for this?

camilamacedo86:
Hi @guettli 

Could you please let us know if this one still required or not?
Do you have thoughts about it?

guettli:
> Hi @guettli 
> 
> Could you please let us know if this one still required or not?
> Do you have thoughts about it? 

Thank you for asking. 

It's not strictly required, but I think it makes sense to add it to the docs. 

Which page would suite?

I could create a PR which extends the page.

camilamacedo86:
We have the PR: https://github.com/kubernetes-sigs/kubebuilder/pull/5215

If you want to continue this one?
But we must make clear what is or not recommended as well and when.

guettli:
[...]

PR #5215, opened 2025-11-15
📖 Add doc about Strict Server-Side Field Validation

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/5208

### DIFF
--- a/docs/book/src/reference/reference.md
+++ b/docs/book/src/reference/reference.md
@@ -5,6 +5,7 @@
     Finalizers are a mechanism to
     execute any custom logic related to a resource before it gets deleted from
     Kubernetes cluster.
+  - [Strict Field Validation](strict-field-validation.md)
   - [Watching Resources](watching-resources.md)
     Watch resources in the Kubernetes cluster to be informed and take actions on changes.
       - [Watching Secondary Resources that are `Owned` ](watching-resources/secondary-owned-resources.md)
--- a/docs/book/src/reference/strict-field-validation.md
+++ b/docs/book/src/reference/strict-field-validation.md
@@ -1,0 +1,214 @@
+# Strict Field Validation
+
+## Overview
+
+By default, when your controller writes an object that contains fields not defined in the CRD schema, the API server:
+
+- Accepts the request
+- Drops the unknown fields
+- May only log a warning
+
+This can hide bugs and version skew between:
+- The controller code (Go types) and
+- The CRD schema installed in the cluster
+
+`controller-runtime` exposes [client.WithFieldValidation][client-field-validation-docs] to turn on strict server-side field validation for all client writes. When enabled, the API server returns a hard error instead of silently dropping unknown fields.
+
+CRDs should be installed before controllers. However, during upgrades this can be best effort rather than guaranteed. Deployment tools may apply resources without strict ordering, and even tools with ordering features don't guarantee the CRD update succeeds. For controllers using external CRDs from third-party operators, ordering may not be controllable at all.
+
+When controllers and CRDs get out of sync, controller writes may fail with unknown field errors, status updates may not persist, and conversion webhooks can fail if the CRD schema is outdated. Controllers may crash-loop until CRDs are updated.
+
+## What does it solve
+
+Strict validation prevents silent failures when your controller code and CRD schemas get out of sync.
+
+For example, you add a new field `status.newField` to your controller, but the CRD in the cluster hasn't been updated yet. When the controller calls `client.Status().Patch(...)`:
+
+**Without strict validation:**
+- API server drops `status.newField` silently
+- Controller sees no error
+- Field never appears on the object - confusing debugging
+
+**With strict validation:**
+- API server returns clear error
+- Controller knows CRDs need updating
+- Fails fast instead of silent data loss
+
+## When to use it
+
+Strict validation is a good fit when:
+
+- You own both the CRDs and the controllers
+- Your upgrade process applies CRDs first or ensures they update together
+- You want to fail fast when a controller writes fields not in the schema
+- You want to catch bugs in your types or conversions early
+- You use typed schemas, or explicitly mark dynamic data with `x-kubernetes-preserve-unknown-fields: true`
+
+## When NOT to use it
+
+Avoid strict validation in production when:
+
+- Controllers and CRDs upgrade independently (i.e., common in Helm)
+- You manage third-party CRDs whose schemas evolve independently
+- Your CRDs use unstructured/dynamic data without `x-kubernetes-preserve-unknown-fields`
+- You need upgrade tolerance when controller and CRD versions are temporarily mismatched
+
+## How Kubebuilder scaffold handles CRD ordering
+
+### Makefile Targets (`make install` + `make deploy`)
+
+```bash
+make install  # Installs CRDs into the cluster
+make deploy   # Deploys the controller
+```
+
+Two separate commands. CRDs are installed first and established before the controller starts. Order is guaranteed.
+
+### YAML bundle distribution (make build-installer)
+
+```bash
+kubectl apply -f dist/install.yaml
+```
+
+The bundle positions CRDs early in the file, after Namespace and before Deployment. This works for new installations.
+
+During upgrades, if the CRD already exists and the update fails or is slow, the Deployment may still update. The new controller may start before the CRD update completes.
+
+### Helm chart distribution
+
+By using the [Helm plugin](../plugins/available/helm-v2-alpha.md), you can distribute your solution as a Helm chart package. Users can install or upgrade it with:
+
+```bash
+helm install my-operator ./dist/chart
+```
+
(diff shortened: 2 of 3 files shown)


camilamacedo86:
Hi @sbueringer and @alvaroaleman 👋

Since we are documenting a feature of controller-runtime — how to use it, when to use it, and related details — could you please give a hand with the review?
Your feedback would be very much appreciated! 🙏

Also, could you please let me know what you think?

I am ping @dlipovetsky since seems to be the author of the feature as well. 
And who request this one: @guettli 

Could you please folks approve this one if you are OK with. 
PS.: I am adding a hold to allow everybody LGTM and etc and give proper time to everybody review

/hold 

Thanks a lot! 💛

docs/book/src/reference/reference.md
```diff
@@ -5,6 +5,10 @@
     Finalizers are a mechanism to
     execute any custom logic related to a resource before it gets deleted from
     Kubernetes cluster.
+  - [Strict Field Validation](strict-field-validation.md)
+    Reject unknown fields instead of silently dropping them. Useful for development
```
alvaroaleman: What does this have to do with upgrade ordering? Is this referring to CRDs being upgraded before controllers? That is always expected to happen IMHO.

Could also mention that this has no effect for server-side apply, there validation is always strict.

camilamacedo86: Hi @alvaroaleman,

Thank you for the quick reply. The above `dev` statement wasn’t entirely accurate — it was added by mistake, so please ignore it. I’ve already fixed it.

However, **please note**:
While CRDs *should* be updated first, in many scenarios this is still best-effort rather than guaranteed. This can become a concern, especially during the lifecycle and upgrade process.

I’ve updated the documentation to address this properly.
I hope this helps clarify the concerns as well.

alvaroaleman: > While CRDs should be updated first, in many scenarios this is still best-effort rather than guaranteed.

IMHO if that fails, it is much preferrable to have errors rather than the apisever silently dropping the fields

camilamacedo86: Depends on the scenario, right? It varies based on how the project is distributed, configured, and used (Helm, GitOps, DevOps setups, etc.).

**The goal of the docs is mainly to clarify:**
– what this feature is and what it solves;
– where it works well and where it may not;
– what users must test themselves — especially upgrades and lifecycle — when using Argo CD, Flux, Helm, third-party images, or different distribution models.

**This way maintainers/users can be aware that it exists** and is a valid option, **make an informed decision** according to their needs and requirements, and understand when it might be problematic or not, and what they need to ensure/test.

Also: is there anything about the feature itself or its interaction with controller-runtime that we’re not sharing, missing, or incorrect? Any key detail we should add?

camilamacedo86: > IMHO if that fails, it is much preferrable to have errors rather than the apisever silently dropping the fields

Depends on the scenario, right? It really varies based on how the project is distributed, configured, and used (Helm, GitOps, DevOps setups, etc.).

**The docs aim to clarify:**

- What this feature is and what it solves;  
- Where it works well and where it may not;  
- What users need to test themselves — especially upgrades and lifecycle — when using Argo CD, Flux, Helm, third-party CRDs etc. Any caveats/concerns they should be aware of;  
- How it plays with the Kubebuilder scaffold and which options might cause issues.

**This way users/maintainers know it exists**, see it as a valid option and **make an informed decision**.

**Also: is there anything about the feature that we’re not explaining, missing, or getting wrong? Any key detail we should add?** 

Unless I misunderstood something — otherwise this wouldn’t be a feature and would just be the default behavior, right?  Why wasn’t it added by default?  What concerns were raised?  **Any other key concerns/considerations I might not be thinking about?**

docs/book/src/reference/strict-field-validation.md
```diff
@@ -0,0 +1,214 @@
+# Strict Field Validation
```
sbueringer, line 1: Sorry was just quickly reading over it so I might have missed this. This seems to focus only on CRDs?

If strict validation is used with builtin-resources this can mean that the controller is only compatible with exactly the Kubernetes version it was compiled against, which is often not desired

camilamacedo86: That is exactly the kind of amazing help that I am looking for !!! 🙌 
It is something that we **must** clarify. 

Thank you, I will try to add it. 
You are amazing 🚀 ⭐

docs/book/src/reference/reference.md
guettli: @camilamacedo86 @alvaroaleman maybe we should make it easier for developers to use WithFieldValidation before documenting it.

I started a thread about that in [#controller-runtime](https://kubernetes.slack.com/archives/C02MRBMN00Z/p1763371935624049)

docs/book/src/reference/strict-field-validation.md
guettli: > If strict validation is used with builtin-resources this can mean that the controller is only compatible with exactly the Kubernetes version it was compiled against, which is often not desired

I heard that Server-Side Apply always uses strict field validation. How can that work? I guess that SSA does a Patch, not an Update.

I guess that strict field validation works (even with other Kubernetes versions), when you use Patch, and the schema matches for the fields of that patch.

Are there docs about that? I would like to learn more about that.

sbueringer: SSA does neither Patch nor Update it does Apply. If you send a field that doesn't exist it will fail

No idea about docs to be honest, maybe somewhere on the Kubernetes website

sbueringer: But good point, maybe it's not too bad to use it with builtin types. You shouldn't try to set fields that might not exist on the Kubernetes versions you run against anyway. And Update is a bad idea in general anyway.

docs/book/src/reference/strict-field-validation.md
```diff
@@ -0,0 +1,214 @@
...
+This can hide bugs and version skew between:
+- The controller code (Go types) and
+- The CRD schema installed in the cluster
+
+`controller-runtime` exposes [client.WithFieldValidation][client-field-validation-docs] to turn on strict server-side field validation for all client writes. When enabled, the API server returns a hard error instead of silently dropping unknown fields.
+
+CRDs should be installed before controllers. However, during upgrades this can be best effort rather than guaranteed. Deployment tools may apply resources without strict ordering, and even tools with ordering features don't guarantee the CRD update succeeds. For controllers using external CRDs from third-party operators, ordering may not be controllable at all.
+
+When controllers and CRDs get out of sync, controller writes may fail with unknown field errors, status updates may not persist, and conversion webhooks can fail if the CRD schema is outdated. Controllers may crash-loop until CRDs are updated.
+
```
guettli, line 20: @camilamacedo86 What about adding this?

```md 
<aside class="warning">
<h1>Why Update is a bad idea</h1>

`client.Update(...)` sends the full object back to the API server. In controllers, this is usually a
poor fit because it can overwrite fields owned by other actors, increase conflict frequency, and
make version skew harder to debug.

Prefer targeted writes with `client.Patch(...)` and `client.Status().Patch(...)` so the reconciler
only changes the fields it owns.

This guidance is separate from strict field validation, but they work well together: patch-based
writes reduce the blast radius, and strict validation turns unknown-field warnings into hard errors.

</aside>
```

alvaroaleman: This is bad advice. Optimistic concurrency is there for a reason and it should not be recommended to disable it by default.

guettli: @alvaroaleman I am looking for a written guideline which helps me and other people new to controller development to understand when it is better to use Update/Patch/Apply. Afaik there is no such guide. Can you please help us? What is your recommendation?

alvaroaleman: There are no written guidelines which is why a PR like this exists in the first place. The correct answer to your question is "it depends". If you are looking for a single, never completely-wrong answer, Patch with optimistic concurrency is the safest option.

guettli: @alvaroaleman @camilamacedo86 

> There are no written guidelines which is why a PR like this exists in the first place. The correct answer to your question is "it depends".

Sorry for the long delay.

I created a PR which addresses this: https://github.com/kubernetes-sigs/kubebuilder/pull/5620

I am not an expert on this topic. But I think a guide about when to use what with examples is helpful.

camilamacedo86:
After thinking more about this, I don’t believe we should add a dedicated document for it. We can’t realistically maintain a reference document for every option in the controller-runtime library.

Instead, it would be better if the documentation on the method itself included more detailed guidance—such as when it should be used, its limitations, and any relevant context. If that level of detail doesn’t currently exist, then documenting the specific option directly within controller-runtime would be the more appropriate approach.

With that in mind, I’m going to close this PR.

PR #5226, opened 2025-11-16
📖 (doc): Move section Versions Compatibility and Supportability for references

camilamacedo86:
We must to provide this info.
However, it is not first info that new users should need to check.
So, it was moved to references since seems more appropriate.


camilamacedo86:
c/c @vitorfloriano

PR #5233, opened 2025-11-17
📖 Revamp migrations guides to reflect current tooling and options

camilamacedo86:
Removes outdated information and clarifies when each approach should be used. Adds guidance on the PROJECT file and its role in automation.


camilamacedo86:
The current migration docs are very confusing and contain a lot of outdated or invalid information.

We need to clearly document **all automation options** to better help users.  
We also need to provide **clear migration steps** that allow users to migrate from **any version to the latest one**.

I am moving forward with this to fix the documentation.  
Follow-ups and improvements are very welcome.

To make this easier and allow more people to contribute, I am going ahead with these changes.

PR #5236, opened 2025-11-18
🐛 fix(cli): format version subcommand output

vitorfloriano:
This PR changes the `version` subcommand output to a human-readable table format:
- Current:
```
$ kubebuilder version
Version: cmd.version{KubeBuilderVersion:"v4.10.1", KubernetesVendor:"1.34.1", GitCommit:"$Format:%H$", BuildDate:"1970-01-01T00:00:00Z", GoOs:"unknown", GoArch:"unknown"}
```
- After:
```
$ kubebuilder version
KubeBuilder:        v4.10.1
Kubernetes:         1.34.1
Git Commit:         c7b5b17b8e1dc4f0fdaf0910464a4d44000365bf
Build Date:         2025-12-24T01:25:15Z
Go OS/Arch:         linux/amd64
```

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -102,5 +102,16 @@ func (v *Version) applyVCSMetadata(settings []debug.BuildSetting) {
 }
 
 func (v Version) PrintVersion() string {
-	return fmt.Sprintf("Version: %#v", v)
+	return fmt.Sprintf(`KubeBuilder:          %s
+Kubernetes:           %s
+Git Commit:           %s
+Build Date:           %s
+Go OS/Arch:           %s/%s`,
+		v.KubeBuilderVersion,
+		v.KubernetesVendor,
+		v.GitCommit,
+		v.BuildDate,
+		v.GoOs,
+		v.GoArch,
+	)
 }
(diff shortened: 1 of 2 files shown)


vitorfloriano:
Hi @camilamacedo86 hope you're doing well.

I opened https://github.com/helm/helm/pull/31574 to fix a similar issue and was informed by the Helm maintainers that formating the `version` subcommand output would be considered a breaking change, so I would like to know if that could be the case for Kubebuilder as well.

I'm not sure if any other programs are consuming the output from `version`.

camilamacedo86:
> Hi @camilamacedo86 hope you're doing well.
>
> I opened https://github.com/helm/helm/pull/31574 to fix a similar issue and was informed by the Helm maintainers that formating the version subcommand output would be considered a breaking change, so I would like to know if that could be the case for Kubebuilder as well.
>
> I'm not sure if any other programs are consuming the output from version.

Thanks for working on this — the new output is much nicer for humans 👍
Yes, this can still be a breaking change in practice if some users are grepping or parsing the current output (even if they shouldn’t). Right now there’s no compatibility or machine-readable option.

I think we can move forward if we add a minimal stable output format, for example:

```
kubebuilder version -o json
# or
kubebuilder version --output=json
```

and keep the pretty output as the default. This gives tooling a stable contract while improving the human UX.

camilamacedo86:
Hi @vitorfloriano 

Could you please rebase this one?
I am looking if we can do a pacth release so we might able to do add it as well.

Thx

Issue #3692, opened 2023-11-12
[Assessment and POC] -  Integrate Server-Side-Apply Support in Kubebuilder via Controller-Tools and Code-Generator Enhancements


camilamacedo86:
### Background:
- Kubebuilder currently depends on controller-tools, which lacks server-side-apply support.
- Consequently, Kubebuilder also does not support server-side-apply.
- An open PR in controller-runtime ([#536](https://github.com/kubernetes-sigs/controller-tools/pull/536)) aims to introduce this support.
- Alternative approach suggested: using [kubernetes/code-generator](https://github.com/kubernetes/code-generator) as seen in PR [#818](https://github.com/kubernetes-sigs/controller-tools/pull/818/files) in controller-tools.
- Discussions in the Kubernetes Slack channel ([link to thread](https://kubernetes.slack.com/archives/CAR30FCJZ/p1698259290391529)) highlight the need for changes in Kubebuilder's default scaffold for compatibility with [applyconfiguration-gen](https://github.com/kubernetes/code-generator/blob/master/cmd/applyconfiguration-gen/generators/openapi.go).

### Proposed Changes:
1. **Move `+groupName` from `groupversion_info.go` to `doc.go`:**
   - Replace [current code](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/api/group.go#L67-L75) with [this implementation](https://github.com/statnett/image-scanner-operator/pull/668/files#diff-c51ff259db6b212a818d769a897ebfe9b21b40e4f833576731fd9712172a07fdL12-R15).
2. **Rename `GroupVersion` to `SchemeGroupVersion` in `groupversion_info.go`.**

### Detailed steps/tasks:
- a) Analyze how `groupName` is used in controller-tools. [see here](https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/markers/doc.go#L44-L45)
- b) Assess the feasibility of moving `groupName` to `doc.go`.
- c) Ensure controller tools and related components remain functional after changes.
- d) Evaluate if changing the default scaffold could introduce breaking changes for existing projects.
- e) Confirm the necessity of changes for compatibility with `/applyconfiguration-gen`.
[...]

PR #5237, opened 2025-11-19
✨(feat): add necessary changes to support applyconfiguration gen

yyewolf:
I'm not sure why this issue has been closed: https://github.com/kubernetes-sigs/kubebuilder/issues/3692
But I haven't seen anyone working on this.

~~I marked the PR as breaking though I am not sure of the semantics, since I renamed `GroupVersion` to `SchemeGroupVersion` I feel like this is a breaking change to the generated code, even though it's not a breaking change to the CLI.~~

I'm also working on modifying the Makefile and E2E tests which is gonna be in another PR.

### DIFF
--- a/pkg/plugins/golang/v4/api.go
+++ b/pkg/plugins/golang/v4/api.go
@@ -104,7 +104,7 @@ func (p *createAPISubcommand) BindFlags(fs *pflag.FlagSet) {
 		"if set, generate the resource without prompting the user")
 	p.resourceFlag = fs.Lookup("resource")
 	fs.BoolVar(&p.options.Namespaced, "namespaced", true, "resource is namespaced")
-	fs.BoolVar(&p.options.GenerateApplyConfiguration, "generate-apply-configuration", false, "if set, generate applyconfiguration code for the resource")
+	fs.BoolVar(&p.options.GenerateApplyConfiguration, "generate-apply-configuration", true, "if set, generate applyconfiguration code for the resource")
 
 	fs.BoolVar(&p.options.DoController, "controller", true,
 		"if set, generate the controller without prompting the user")
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -123,9 +123,9 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	{{ if .BoilerplatePath -}}
-	"$(CONTROLLER_GEN)" object:headerFile={{printf "%q" .BoilerplatePath}} paths="./..."
+	"$(CONTROLLER_GEN)" applyconfiguration:headerFile="hack/boilerplate.go.txt" object:headerFile={{printf "%q" .BoilerplatePath}} paths="./..."
 	{{- else -}}
-	"$(CONTROLLER_GEN)" object paths="./..."
+	"$(CONTROLLER_GEN)" applyconfiguration object paths="./..."
 	{{- end }}
 
 .PHONY: fmt


pkg/plugins/golang/v4/scaffolds/internal/templates/api/doc.go
```diff
@@ -0,0 +1,59 @@
+/*
```
camilamacedo86, line 1: Thank you for looking into that. 🎉 

If it turns out to be a breaking change, we need to find a way to make it backwards compatible. We can’t introduce it right now.

However, if you want to go ahead and implement the changes based on how you think it should work, and then share your findings with us, we can refine it together in a follow-up.

Because of that, I wouldn’t worry about the e2e tests at this stage. Instead, focus on making the changes in a way that, when you run make generate, the testdata samples are regenerated—or at least show how a project would look with the new behavior.

pkg/plugins/golang/v4/scaffolds/internal/templates/api/group.go
```diff
@@ -64,11 +63,11 @@ import (
 )
 
 var (
-	// GroupVersion is group version used to register these objects.
-	GroupVersion = schema.GroupVersion{Group: "{{ .Resource.QualifiedGroup }}", Version: "{{ .Resource.Version }}"}
+	// SchemeGroupVersion is group version used to register these objects.
+	SchemeGroupVersion = schema.GroupVersion{Group: "{{ .Resource.QualifiedGroup }}", Version: "{{ .Resource.Version }}"}
```
camilamacedo86: Why would those changes? 
We did not change the values. Why would we need to change the exported variable name?

pkg/plugins/golang/v4/scaffolds/internal/templates/api/types.go
```diff
@@ -64,6 +64,7 @@ func (f *Types) SetTemplateDefaults() error {
 //nolint:lll
 const typesTemplate = `{{ .Boilerplate }}
 
+// +kubebuilder:ac:generate=true
```
camilamacedo86, line 68: I think the feature must be optional.
We should not promote it by default, right?
By default, people should still be working as they do today.

pkg/plugins/golang/v4/scaffolds/internal/templates/api/types.go
```diff
@@ -116,6 +117,8 @@ type {{ .Resource.Kind }}Status struct {
 // +kubebuilder:resource:scope=Cluster
 {{- else if not .Resource.IsRegularPlural }}
 // +kubebuilder:resource:path={{ .Resource.Plural }}
+{{- else }}
+// +kubebuilder:resource
```
camilamacedo86, line 125: What does this marker do? 
Have you an example of a project using the option for we give a look?
Could you share?

pkg/plugins/golang/v4/scaffolds/internal/templates/api/group.go
yyewolf: The generated code for applyconfiguration uses this variable name, it's referenced here: https://github.com/kubernetes/code-generator/blob/master/cmd/applyconfiguration-gen/generators/util.go#L129

However in order to make it backward compatible, we can easily go back and do

```suggestion
	GroupVersion = schema.GroupVersion{Group: "{{ .Resource.QualifiedGroup }}", Version: "{{ .Resource.Version }}"}
	
	SchemeGroupVersion = GroupVersion 
```

pkg/plugins/golang/v4/scaffolds/internal/templates/api/types.go
yyewolf: controller-tools uses this marker to mark the type as a resource type:

Marker: https://github.com/kubernetes-sigs/controller-tools/blob/main/pkg/applyconfiguration/gen.go#L45C1-L45C127
Help message associated: https://github.com/kubernetes-sigs/controller-tools/blob/main/pkg/applyconfiguration/gen.go#L80

pkg/plugins/golang/v4/scaffolds/internal/templates/api/types.go
yyewolf: You're right, I'll make it optional

camilamacedo86: But if you want to add it in a way that we can first see it working, then we can evaluate it.

Do you have a project that uses this configuration? I’ve never tried to use it myself, so looking at an example would be helpful and could give us some ideas on how we might add it.

yyewolf: I don't have a project that uses this configuration, however I ran the generate testdata using it: https://github.com/kubernetes-sigs/kubebuilder/commit/537f54ef5a525ebf955862280d602bd0b9fe71d7

yyewolf: I added the flag, made it true by default (for now so we can see the generated code) and also generated the code 👍

pkg/plugins/golang/v4/scaffolds/api.go
```diff
@@ -100,6 +100,7 @@ func (s *apiScaffolder) Scaffold() error {
 		if err := scaffold.Execute(
 			&api.Types{Force: s.force},
 			&api.Group{},
+			&api.Doc{},
```
camilamacedo86, line 103: We should only generate the new files if the/when the flag is informed

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -123,9 +123,9 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	{{ if .BoilerplatePath -}}
-	"$(CONTROLLER_GEN)" object:headerFile={{printf "%q" .BoilerplatePath}} paths="./..."
+	"$(CONTROLLER_GEN)" applyconfiguration:headerFile="hack/boilerplate.go.txt" object:headerFile={{printf "%q" .BoilerplatePath}} paths="./..."
 	{{- else -}}
-	"$(CONTROLLER_GEN)" object paths="./..."
+	"$(CONTROLLER_GEN)" applyconfiguration object paths="./..."
```
camilamacedo86, line 128: So, here we should not do this change as well.
We should not change the default scaffold to be applyconfiguration

That would mean all would only work with. 

So, the first question is: 

In a project, can I have both?

Apis created with applyconfiguration and others not?
OR if we use applyconfiguration all APIs will need to generate with it?

pkg/plugins/golang/v4/api.go
```diff
@@ -104,6 +104,7 @@ func (p *createAPISubcommand) BindFlags(fs *pflag.FlagSet) {
 		"if set, generate the resource without prompting the user")
 	p.resourceFlag = fs.Lookup("resource")
 	fs.BoolVar(&p.options.Namespaced, "namespaced", true, "resource is namespaced")
+	fs.BoolVar(&p.options.GenerateApplyConfiguration, "generate-apply-configuration", true, "if set, generate applyconfiguration code for the resource")
```
camilamacedo86, line 107: Can it be done per API or the whole project need to be changed to adopted this?
I mean, can I have 3 APIs that uses applyconfiguration and 3 that do not use them in the same project?
Is that possible?

testdata/project-v4-multigroup/api/crew/v1/applyconfiguration/crew/v1/captainstatus.go
```diff
@@ -0,0 +1,47 @@
+/*
```
camilamacedo86: IN the testdata we should have no changes.
We should have all projects generate as before.
See that we have test/generate.sh

So,  we would need to have a new mock using it and then yes we would only have the scaffold change for the specific API created with the flag. If we cannot have for example X APIs using it and X APIs using not then the change/flag cannot be in the API and it need to be in the edit and init commands ( such as we have multi-group option )

yyewolf:
Sorry I've been kinda busy lately, I'll try and work on this PR after the 20th when I get more time to myself, hope you understand 😄

Issue #5246, opened 2025-11-23
[helm/v2-alpha] missing tolerations, affinity and nodeSelector parameters


robinlioret:
### What do you want to happen?

The generated dist/chart/templates/manager/manager.yaml file for the Deployment of the operator is currently not exposing the tolerations, nodeSelector nor affinity options. Which could be required in certain deployment cases.

The plugin should generate something like this:

```yaml
# dist/chart/templates/manager/manager.yaml
apiVersion: apps/v1
kind: Deployment
# ...
spec:
    # ...
    template:
        # ...
        spec:
            {{- with .Values.manager.tolerations }}
            tolerations: {{ toYaml . | nindent 16 }}
            {{- end }}
            {{- with .Values.manager.nodeSelector }}
            nodeSelector: {{ toYaml . | nindent 16 }}
            {{- end }}
            {{- with .Values.manager.affinity }}
            affinity: {{ toYaml . | nindent 16 }}
            {{- end }}
            containers:
             # ...
```

And expose it in the default values file:

```yaml
# dist/chart/values.yaml

# Configure the controller manager deployment
manager:
  # ...

  # Pods tolerations
  tolerations: []

  # Pods node selection
  nodeSelector: {}

  # Pod affinity
  affinity: {}
```

Note: I'm aware that this is kind of a duplicate of #4912 but since the previous issue was closed, i'm opening a new one.

camilamacedo86:
Thanks for the proposal.

We don’t plan to expose `tolerations`, `nodeSelector`, or `affinity` as Helm
values in the generated chart.

The Helm chart generated by Kubebuilder is only meant to **package and
distribute** a project. It should not define or change what the project
supports.

Scheduling options like these are **cluster and platform policy** decisions.
Allowing them as Helm values would let end users run the controller on nodes,
OSes, or architectures that the project authors did not build or test for.
This would break the project’s support guarantees and may result in
non-working deployments.

If a project requires specific scheduling rules, those must be **defined and
owned by the project itself** (for example in the manifests or via kustomize),
not provided as configurable Helm values.

We should only have in the Helm values what make sense be changed by who consumes the project.
Maintainers should properly configure things like that via kustomize files so that when they package their solution with Helm those should endup in the HelmChart manifest but not exposed in the values.

So, I would like to close this one as not planned.
Please, let me know if you understand the reasons and wdyt.

camilamacedo86:
From @robinlioret 

[...]

PR #5247, opened 2025-11-25
✨ (helm/v2-alpha) add tolerations, nodeselector and affinity

robinlioret:
Adds nodeSelector, affinity and tolerations to the manager deployment template and values of the helm chart (v2-alpha).

Related issue: #5246

### DIFF
--- a/docs/book/src/getting-started/testdata/project/dist/chart/values.yaml
+++ b/docs/book/src/getting-started/testdata/project/dist/chart/values.yaml
@@ -40,6 +40,15 @@ manager:
         cpu: 10m
         memory: 64Mi
 
+  # Manager pod's affinity
+  affinity: {}
+
+  # Manager pod's node selector
+  nodeSelector: {}
+
+  # Manager pod's tolerations
+  tolerations: []
+
 # Essential RBAC permissions (required for controller operation)
 # These include ServiceAccount, controller permissions, leader election, and metrics access
 # Note: Essential RBAC is always enabled as it's required for the controller to function
--- a/docs/book/src/plugins/available/helm-v2-alpha.md
+++ b/docs/book/src/plugins/available/helm-v2-alpha.md
@@ -188,6 +188,35 @@ controllerManager:
       cpu: 10m
       memory: 64Mi
 
+  # Manager pod's affinity
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+          - matchExpressions:
+            - key: kubernetes.io/arch
+              operator: In
+              values:
+                - amd64
+                - arm64
+                - ppc64le
+                - s390x
+            - key: kubernetes.io/os
+              operator: In
+              values:
+                - linux
+
+  # Manager pod's node selector
+  nodeSelector:
+    kubernetes.io/os: linux
+
+  # Manager pod's tolerations
+  tolerations:
+    - key: "node.kubernetes.io/unreachable"
+      operator: "Exists"
+      effect: "NoExecute"
+      tolerationSeconds: 6000
+
 # Essential RBAC permissions (required for controller operation)
 # These include ServiceAccount, controller permissions, leader election, and metrics access
 # Note: Essential RBAC is always enabled as it's required for the controller to function
--- a/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/manager/manager.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/manager/manager.yaml
@@ -21,6 +21,15 @@ spec:
                 app.kubernetes.io/name: project
                 control-plane: controller-manager
         spec:
+            {{- with .Values.manager.tolerations }}
+            tolerations: {{ toYaml . | nindent 16 }}
+            {{- end }}
+            {{- with .Values.manager.affinity }}
+            affinity: {{ toYaml . | nindent 16 }}
+            {{- end }}
+            {{- with .Values.manager.nodeSelector }}
+            nodeSelector: {{ toYaml . | nindent 16 }}
+            {{- end }}
             containers:
                 - args:
                     {{- if .Values.metrics.enable }}
--- a/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/values.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/values.yaml
@@ -40,6 +40,15 @@ manager:
         cpu: 10m
         memory: 64Mi
 
+  # Manager pod's affinity
+  affinity: {}
+
+  # Manager pod's node selector
+  nodeSelector: {}
+
+  # Manager pod's tolerations
+  tolerations: []
+
 # Essential RBAC permissions (required for controller operation)
 # These include ServiceAccount, controller permissions, leader election, and metrics access
 # Note: Essential RBAC is always enabled as it's required for the controller to function
--- a/docs/book/src/getting-started/testdata/project/dist/chart/templates/manager/manager.yaml
+++ b/docs/book/src/getting-started/testdata/project/dist/chart/templates/manager/manager.yaml
@@ -21,6 +21,15 @@ spec:
                 app.kubernetes.io/name: project
                 control-plane: controller-manager
         spec:
+            {{- with .Values.manager.tolerations }}
+            tolerations: {{ toYaml . | nindent 16 }}
+            {{- end }}
+            {{- with .Values.manager.affinity }}
+            affinity: {{ toYaml . | nindent 16 }}
+            {{- end }}
+            {{- with .Values.manager.nodeSelector }}
+            nodeSelector: {{ toYaml . | nindent 16 }}
+            {{- end }}
             containers:
                 - args:
                     {{- if .Values.metrics.enable }}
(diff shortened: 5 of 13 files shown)


pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/manager/config.go
```diff
@@ -96,6 +96,8 @@ spec:
       #             operator: In
       #             values:
       #               - linux
+      nodeSelector: {}
+      tolerations: []
```
camilamacedo86: We need make it work without uncomment those lines

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/manager/config.go
```diff
@@ -80,7 +80,7 @@ spec:
       # according to the platforms which are supported by your solution.
```
camilamacedo86, line 80: Could you please ensure that we have only one commit?

robinlioret: Sure, should I wait for the PR to be ready for merge before squashing the commits ?

pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/manager/config.go
robinlioret: Should it work both ways ?
* If specified in the kuztomize, then use it
* If not, use a default ?

robinlioret: It is done, the template generation will have two behaviour depending on what was found in the kustomize manifest.

docs/book/src/cronjob-tutorial/testdata/project/config/manager/manager.yaml
```diff
@@ -50,6 +50,14 @@ spec:
...
       #             values:
       #               - linux
+      # TODO(user): Uncomment the following code to configure the nodeSelector.
+      # nodeSelector: 
+      #	  kubernetes.io/os: linux
+      # TODO(user): Uncomment the following code to configure the tolerations.
+      #	tolerations:
+      # - key: "key1"
+      #   operator: "Exists"
+      #   effect: "NoSchedule"
```
camilamacedo86: Do we need to do those changes?
Can we live it as before?

camilamacedo86: I think we just need to remove this change here than all seems fine

robinlioret: We can remove it, I added them to be coherent with the other optional fields.

The rest is not fine enough, I still have a weird certificate issue on the e2e helm tests.

robinlioret: Done

robinlioret:
The PR looks ready to be merged.

docs/book/src/getting-started/testdata/project/dist/chart/values.yaml
```diff
@@ -37,6 +37,15 @@ manager:
...
         memory: 64Mi
 
+  # Pod's affinity
+  affinity: {}
+
+  # Pod's node selector
+  nodeSelector: {}
+
+  # Pod's tolerations
+  tolerations: []
```
camilamacedo86, line 50: That is valid only for the mananger pod right?
Could we just add a comment like `Manager Pod's ... ` so that we can clarify it?
Also, could we add examples over how to fill those fields in the docs? 
WDYT?

robinlioret: You're right, I made the change.

camilamacedo86 (approved):
/lgtm

robinlioret:
Thanks for approving the PR. I see two failing tests though.

- Test GoReleaser and CLI Version / go-releaser-test (pull_request)

Fails while fetching https://github.com/anchore/syft/releases/latest (ERR 503)
Is it a rate limitation ? Can we restart this specific test to confirm ?

- E2E Testdata Sample / e2e-tests-project-v4 (pull_request)

Fails like nothing was listening on the metric port.
I tried to generate the test-data again. No change.

```
manager:
    Port:          9443/TCP (webhook-server)
    Host Port:     0/TCP (webhook-server)
    Args:
      --metrics-bind-address=:8443
```
Is it a bug in the installation file ? Shouldn't it expose the 8443 port as well ?

docs/book/src/plugins/available/helm-v2-alpha.md
```diff
@@ -188,6 +188,35 @@ controllerManager:
...
+              operator: In
+              values:
+                - amd64
+                - arm64
+                - ppc64le
+                - s390x
+            - key: kubernetes.io/os
+              operator: In
+              values:
+                - linux
```
camilamacedo86, line 207: This must be defined by who maintain the project and not who consume the project
If the image is not built for those then allowing who consume the solution via HelmChart add those will just end up in issues. 

We should only expose values in the HelmCharts that make sense for who consume the solution.

camilamacedo86:
Hi @robinlioret,

Thanks for the contribution.

We can’t accept this change.

The Helm chart is only used to **package and distribute** a project.
It is **not** meant to define or change what the project supports.

Things like:
- nodeSelector
- affinity
- tolerations
- supported architectures / OS

are decisions made by the **project maintainers**, based on how the
controller image is built and tested. That must be configured in the manager and should appear in the manager file of the helm chart but not exposed via values. 

If we expose these as Helm values, end users could deploy the controller
on nodes or platforms that the project does not support.
That would break the project’s support guarantees and may not work at all.

If a project needs specific scheduling rules, they must be **defined by
the project itself** (for example in the manifests or kustomize),
not configured by Helm values.

For this reason, we don’t expose these options in the generated Helm chart.

Thanks again for the proposal.

/hold

docs/book/src/plugins/available/helm-v2-alpha.md
robinlioret: I'm confused. You asked in a previous comment to put an example in the doc. Now, I should remove it ?

camilamacedo86: This is not about the example or the docs.

This configuration should **not** be exposed to end users. It is the responsibility of the **project maintainers** to define it in the config/manager (Kustomize).

Only values that users can safely change should be exposed in Helm `values.yaml`. Anything that defines core capabilities—such as supported architectures—must be fixed by the maintainer.

Supported architectures should be configured in Kustomize and propagated into the Helm charts, but **not** exposed as user-configurable values. Allowing users to change these settings can result in broken deployments when the image or project does not actually support them.

robinlioret: Where is this PR defining supported architecture ?

robinlioret:
If I understand correctly, from your point of view. Configurations like the manager pod's `affinity`, `nodeSelector` and `tolerations` should not be exposed to end-users. This is the project maintainer responsability because any inconsiderate change to these values may lead to a broken deployment.

I disagree with that statement, here's why.

Regarding the generated Helm chart. We are currently exposing the value `manager.image.repository` where any modification can easily break deployments. Following your logic, it should not be exposed either. Then, you could argue than end-users may have an internal registry for security reasons. This would be totally legit.

The same reasoning is applicable to many (if not all) values exposed.

However, regarding `toleration` and `nodeSelector`, an end-user may need to assign the operator to a specific set of nodes.

As a matter of fact, it is my case and the reason why I opened this pull request in the first place. We have a node pool dedicated to tools and core capabilities. This nodepool have a taint. Therefore, as an end user, I need to be able to configure the `nodeSelector` and the `toleration` values. We also need to customize `affinity` to guarantee high availability if `replicaCount` is higher than 2. We currently have to re-implement `affinity`, `nodeSelector` and `tolerations` every time we regenerate the chart.

`affinity`, `nodeSelector` and `tolerations` are exposed in many charts for the same reason, even in the default chart created by helm when you run `helm create test-chart`.

This way, I see no valid reason to forbid end-users to customize these values.

We should provide:
* A transparent behaviour that ignores `affinity`, `nodeSelector` and `tolerations`  if not specified (keep the behaviour we have today)
* But keep the possibility for the end-user to customize them. Like we do for image repository and other values.
* If values are defined by the operator developper in the Kustomize files then it should be use as a default value for the helm chart

camilamacedo86:
Hi @robinlioret,

Changing `manager.image.repository` is typically about mirroring the same image
(private or air-gapped registries). **It doesn’t change what the project supports.** I do not think that we can use this an as example. 

Exposing `affinity`, `nodeSelector`, and `tolerations` is different. It doesn’t
add platform support, but it allows consumers to deploy the controller outside
the authors’ validated and supported environments (including OS/arch). That
support boundary must be owned by the operator authors, not end users. **However, I can see the cases where it might be required and useful.** 

The fact that `helm create` exposes these by default isn’t a strong argument
here. That scaffold assumes the chart is authored and consumed by the same
party. Kubebuilder’s Helm chart is meant only **to package and distribute a
solution**, not to expose additional end-user knobs that can result in unsupported
deployments. 

**However, you might be right, and we may need to add it.** 

I see there are cases where placement customization is needed, but exposing
these by default in a packaging-only scaffold is something **I’m still thinking
about.** **Can you please give me a few days to think more about this one?**

robinlioret:
I'm glad we understand each other.

I agree that this give more freedom to the end-users than the author might expect.

Nonetheless, I think that when an end-user needs such configuration possibilities, they know their requirements and are aware of what they're doing. For most users, the default configuration will be enough.

Take all the time you need, this is not urgent.

camilamacedo86:
After consider the pros and cons
I think you are probably right on this one
So

/lgtm
/approve

/hold cancel

PR #5080, opened 2025-09-05
📖 add note about Windows support

vitorfloriano:
This pull request updates the project's documentation to clarify the status of Windows support and provide guidance for Windows users.

Documentation updates regarding Windows support:

* Updated the `README.md` to direct Windows users to a new `docs/windows.md` file for instructions, and clarified that contributions towards native Windows support are not planned.
* Added a new `docs/windows.md` file explaining that native Windows support is not being pursued, and providing instructions for using Kubebuilder on Windows via **WSL2** and **Docker Desktop**, with links to relevant official documentation.

Closes #2940

PR #5251, opened 2025-11-29
(cli): Add experimental Windows platform support

903174293:
Add Windows/Cygwin/MINGW64 support to generated Makefiles

Fix Kubebuilder to work correctly on Windows environments (CMD, PowerShell,
Cygwin, and Git Bash/MINGW64) by improving the generated Makefile to handle
path format differences.

Problem:
- Cygwin/MINGW64 make provides Unix-style paths (/cygdrive/d/... or /d/...)
- Windows Go tools require Windows-style paths (D:/...)
- Setting GOBIN with Unix paths caused "go install" to fail
- Binary execution failed due to path format mismatches

Solution:
- Detect environment (Windows/Cygwin/MINGW64/Linux/macOS)
- Convert paths between Unix and Windows formats as needed
- Use Windows paths for GOBIN (Go tools)
- Use Unix paths for shell execution (in Cygwin/MINGW64)
- Handle .exe extensions on Windows
- Use module paths instead of ./... for better compatibility

Changes:
- Modified: pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
- Added path conversion functions (convert-to-windows-path, convert-to-mingw-path)
- Added environment detection (UNAME_S, IS_MINGW, IS_CYGWIN)
- Added dual path variables (LOCALBIN_FOR_GO, LOCALBIN_FOR_SHELL)
- Updated tool binary paths to use correct format per platform
- Improved go-install-tool function to handle all environments

Testing:
- Tested on Windows CMD, PowerShell, Cygwin, Git Bash (MINGW64), and Linux
- All environments pass: kubebuilder init && kubebuilder create api
- Verified generated files: bin/controller-gen.exe, api/v1/zz_generated.deepcopy.go
- No breaking changes to Linux/macOS behavior

Backward Compatibility:
- No breaking changes
- Existing projects continue to work
- Graceful degradation if tools unavailable
- No API changes

Fixes: #[issue-number]

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -118,14 +118,14 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	"$(CONTROLLER_GEN)" rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	"$(CONTROLLER_GEN)" rbac:roleName=manager-role crd webhook paths="$(CONTROLLER_GEN_PATHS)" output:crd:artifacts:config=config/crd/bases
 
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	{{ if .BoilerplatePath -}}
-	"$(CONTROLLER_GEN)" object:headerFile={{printf "%q" .BoilerplatePath}} paths="./..."
+	"$(CONTROLLER_GEN)" object:headerFile={{printf "%q" .BoilerplatePath}} paths="$(CONTROLLER_GEN_PATHS)"
 	{{- else -}}
-	"$(CONTROLLER_GEN)" object paths="./..."
+	"$(CONTROLLER_GEN)" object paths="$(CONTROLLER_GEN_PATHS)"
 	{{- end }}
 
 .PHONY: fmt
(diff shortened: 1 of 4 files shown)


903174293:
在windows os上执行创建文件件，进入到文件夹下，执行如下命令即可验证：
git init
go mod init hihonor.com/kubebuilder/app
kubebuilder init --domain hihonor.com
kubebuilder create api --group kubebuilder --version v1 --kind App

vitorfloriano:
Hi @903174293 thanks for opening the PR. I'm afraid native Windows support is not currently on the roadmap for Kubebuilder. The motivation behind not pursuing Windows support can be read in [here](https://github.com/kubernetes-sigs/kubebuilder/issues/2940), [here](https://github.com/kubernetes-sigs/kubebuilder/pull/5080) and [here](https://github.com/vitorfloriano/kubebuilder/blob/master/docs/windows.md)

Nonetheless, if you still think native support is better than the WSL2 + Docker combo we recommend, or if there are any limitations/shortcomings to that, please let us know and we can further discuss this patch.

pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -253,17 +253,72 @@ undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.
 ##@ Dependencies
 
 ## Location to install dependencies to
-LOCALBIN ?= $(shell pwd)/bin
+LOCALBIN ?= $(CURDIR)/bin
+
+# WINDOWS_CYGWIN_FIX: Convert Cygwin/Git Bash paths to Windows paths for Go tools
```
camilamacedo86, line 258: Kubebuilder should not scaffold content that is specific to Windows—or to any particular OS. 
/hold

camilamacedo86:
Hi @903174293,

Thank you for your contribution and for taking the time to submit this PR.

At this time, we’re not able to accept this change. The default scaffolds **should remain platform-agnostic and must not include Windows or other OS-specific assumptions.** In addition, adding Windows support would require us to be able to test and maintain the solution within this platform, which we’re currently unable to do.

That said, @vitorfloriano has provided a good explanation and additional context here:  
https://github.com/kubernetes-sigs/kubebuilder/pull/5251#issuecomment-3602911533

For these reasons, I’m closing this PR as deferred.

Thanks again for your contribution and understanding.

PR #5256, opened 2025-12-07
✨ (helm/v2-alpha): Add support for exporting unknown resources from install YAML via extra directory

### LINKED ISSUE #5248: [helm/v2-alpha] Manual resources (such as Service) are filtered-out and not included in the final chart

I scaffolded an operator with the operator-sdk and manually added a `Service` that my operator exposes. I would like to use the new `helm/v2-alpha` plugin to also distribute my operator as a Helm chart. However, manually added resources are not included in the final chart. Shouldn't there at least be a warning about resources that are filtered out?

### Reproducing this issue

I have the following `Service` in `config/services/my_service.yaml`

```yaml
apiVersion: v1
kind: Service
metadata:
  labels:
    app.kubernetes.io/name: my-operator
  name: my-alert-service
  namespace: system
spec:
  ports:
  - name: http
    port: 8000
    protocol: TCP
    targetPort: http
  selector:
```

Some of the `kustomization.yaml` files also include that resource. When I build the installer, the final `install.yaml` also includes this service. When I then run `kubebuilder edit --plugins=helm/v2-alpha` this service will not be included in the final chart output.

### Workaround

The service is only included if it has the string `webhook` in its name, because of this filter: https://github.com/kubernetes-sigs/kubebuilder/blob/806fa7038bfc6aada94be8bda9b77f64c4c22189/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/resource_organizer.go#L164

### KubeBuilder (CLI) Version

v4.10.1

### PROJECT version

3

### Plugin versions

```yaml
go.kubebuilder.io/v4
```

### Other versions

_No response_

### Extra Labels

_No response_

UJESH2K:
Fixes #5248

---

### Problem 

The `helm/v2-alpha` plugin currently **silently drops Kubernetes resources** that are present in the generated `install.yaml` but are **not explicitly handled** by the default Helm scaffolds.

This most commonly affects:

* user-defined `Service`s
* `ConfigMap`s
* custom resources
* or any additional manifests injected into the kustomize pipeline

Although these resources exist in `install.yaml`, they never appear in the generated Helm chart, leading to **unexpected and confusing behavior** for users.

Importantly, this failure mode is **silent**: there is no warning, error, or indication that resources were discarded.

---

### Root cause analysis

The issue does **not** originate in kustomize, but in the helm/v2-alpha chart generation pipeline.

Previously:

1. Resources were parsed from `install.yaml`
2. Only a **fixed, hard-coded subset** of resource kinds was explicitly classified and consumed
3. Any resource that did not match those categories was:

   * never marked as “handled”
   * never written into the chart
   * never reported to the user

This meant the organizer was operating on a **partial view of resources**, causing silent data loss during chart generation.

---

### Why partial fixes are insufficient

Earlier approaches attempted to special-case individual resource types (for example, generic Services).

This creates long-term problems:

* Inconsistent behavior across resource kinds
* Continuous maintenance burden as new kinds appear
* Users cannot reliably predict what *will* or *won’t* be exported

For Helm chart generation, **consistency and predictability matter more than selective support**.

---

### What this PR changes (high-level)

This PR introduces a **generic, resource-agnostic mechanism** that guarantees:

> **Every resource produced by kustomize is either scaffolded or preserved — never dropped.**

Specifically:

1. **All resources from `install.yaml` are tracked**
2. Resources consumed by default Helm scaffolds behave exactly as before
3. Any remaining (unhandled) resources are written to:

   ```
   templates/extras/
   ```
4. A warning is emitted to explicitly inform users when extras are generated

This ensures:

* no silent data loss
* consistent behavior across all resource kinds
* predictable Helm output

---

#### `parser.go`

* Now tracks **all resources** parsed from `install.yaml`
* Maintains existing type-based categorization without changing behavior

#### `resource_organizer.go`

* Explicitly marks resources consumed by default scaffolds
* Collects any remaining resources into a dedicated `extras` group
* Classification is done *after* parsing completes, ensuring a complete resource set

#### `chart_converter.go`

* Writes unhandled resources under `templates/extras`
* Emits a clear warning when extras are generated, making behavior visible to users

#### `extras_resources_test.go`

* Adds unit coverage to validate that:

  * multiple unhandled resources are preserved
  * resources are not misclassified
  * nothing is dropped silently

---

### Behavioral guarantees after this change

With this PR applied:

* ✅ Every resource in `install.yaml` appears in the Helm chart
* ✅ Known resources are scaffolded exactly as before
* ✅ Unknown or user-defined resources are preserved verbatim
* ✅ Users are explicitly informed when extras are generated
* ❌ Silent resource loss is no longer possible

---

### Tests and verification

* New unit tests validate that multiple unhandled resources (e.g. Services, ConfigMaps) are correctly routed to `templates/extras`
* Existing helm/v2-alpha tests continue to pass unchanged
* Lint and CI checks pass cleanly

---

### Why this approach is safe and future-proof

* Does not special-case resource kinds
* Does not change existing scaffold behavior
* Requires no future updates as new Kubernetes resources are introduced
* Makes limitations explicit instead of implicit


pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/resource_organizer.go
```diff
@@ -150,6 +156,18 @@ func (o *ResourceOrganizer) collectMetricsResources() []*unstructured.Unstructur
...
+func (o *ResourceOrganizer) collectGenericServices() []*unstructured.Unstructured {
+	var generic []*unstructured.Unstructured
+
+	for _, service := range o.resources.Services {
+		if o.isGenericService(service) {
+			generic = append(generic, service)
+		}
+	}
+	return generic
+}
```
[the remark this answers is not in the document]
UJESH2K: I’ve added a new unit test (generic_services_test.go) that covers the generic Service behavior mentioned here.

The test verifies that a Service without "webhook" or "metrics" in its name is:

- detected as a generic Service

- included in the "services" group

- not placed under "webhook" or "metrics"

This provides test coverage for the new logic introduced in this PR.
Let me know if additional integration-level chart output tests are needed.

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/resource_organizer.go
```diff
@@ -83,6 +83,12 @@ func (o *ResourceOrganizer) OrganizeByFunction() map[string][]*unstructured.Unst
 		groups["other"] = o.resources.Other
 	}
 
+	// Generic Services - Services that are neither webhook nor metrics
+	genericServices := o.collectGenericServices()
+	if len(genericServices) > 0 {
+		groups["services"] = genericServices
+	}
```
camilamacedo86, line 121: Probably instead of allow only any service we might want to migrate any kustomize scaffolds, but I think it is the case already. So, could we please check because as it is now I think we are already creating the Helm Chart with all that is in the kustomize install.yaml if not then I think we need a more generic solution. WDYT?

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/resource_organizer.go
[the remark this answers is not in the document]
camilamacedo86: OK, that makes sense. However, if we do this for Services, we should do it for **any** kind of resource.

**Consistent behavior is key. We cannot port some resources (like Services) while silently ignoring others.**

So we either need:
- a **generic way** to port *all* resources from `install.yaml` that are not part of the default scaffold, or  
- to **not port any of them** and clearly communicate that limitation.

Partial support would be confusing and hard to maintain.

Related discussion:
https://github.com/kubernetes-sigs/kubebuilder/issues/5248#issuecomment-3649028090

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/resource_organizer.go
UJESH2K: Thanks for the clarification — I agree.
After rechecking, we’re not currently porting all resources from install.yaml into the Helm chart; only explicitly handled kinds are included, while others are silently ignored.
I agree that adding support only for generic Services creates inconsistent behavior if other resource kinds are still dropped. A generic mechanism to port any non-scaffolded resource would be cleaner than special-casing Services.
I’m happy to refactor this toward a generic solution if that aligns with the project direction. Please let me know how you’d like to proceed.

UJESH2K:
Partial support is not ideal here. I’m happy to refactor this into a generic solution that ports any non-scaffolded resource from install.yaml, instead of special-casing Services.
If that direction works for you, I’ll update the PR accordingly.

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/resource_organizer.go
camilamacedo86: The idea gave for the folk that report this one to add all others one in a generic dir like others might be fit.
I think we may can:
- After all check if has any resource in the install.yaml that was not ported to the helm chart
- Then port to a generic dir like others 
- Then warning something like "resources that were not identified were added under template/extras"

Something like might be a good approach
open to ideas and suggestions as well

UJESH2K:
I wanted to clarify the overall approach in this PR.

Previously, not all resources from the generated install.yaml were making it through the helm/v2-alpha pipeline, which meant some user-defined resources were never even considered for classification and were silently dropped.

With this change, all resources produced by kustomize are now parsed and tracked first. From there, resources that are recognized by the default helm/v2-alpha scaffolds are classified and written as usual. Any remaining resources that are not classified are preserved under templates/extras, and a warning is emitted to make this behavior visible to users.

The key change here is ensuring that nothing is lost during parsing, so classification happens on a complete set of resources rather than a partial one.

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/resource_organizer.go
```diff
@@ -38,49 +38,86 @@ func NewResourceOrganizer(resources *ParsedResources) *ResourceOrganizer {
...
 	if len(o.resources.Other) > 0 {
 		groups["other"] = o.resources.Other
+		for _, r := range o.resources.Other {
+			used[r] = true
+		}
+	}
+
+	// Extras - any resource from install.yaml not explicitly handled above
+	extras := []*unstructured.Unstructured{}
+	for _, r := range o.resources.All() { //ASSUMPTION: ParsedResources exposes All()
```
camilamacedo86: Could we remove the extra comment?

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/resource_organizer.go
```diff
@@ -90,12 +127,10 @@ func (o *ResourceOrganizer) OrganizeByFunction() map[string][]*unstructured.Unst
 func (o *ResourceOrganizer) collectRBACResources() []*unstructured.Unstructured {
 	var rbacResources []*unstructured.Unstructured
 
-	// Service account
```
camilamacedo86, line 93: Could we came back those comments?

camilamacedo86:
@UJESH2K

Thank you for looking into this 🙂
I’ve updated the PR title — I hope that’s okay.

Could you please make sure we have a mock scenario that properly validates these changes?
For example, we should include extra ConfigMaps, Services, etc., and confirm that all of them end up in the extra directory.

Also, can we make sure we only change what is required to achieve the goal?
For example, could you please revert the removal of comments?

Lastly, could you ensure the changes pass all CI checks? I noticed the lint check is currently failing.
Please also make sure the PR description matches the latest changes.

Thank you!

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/resource_organizer.go
UJESH2K: sure!!

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/resource_organizer.go
UJESH2K: yes i did add them back

camilamacedo86:
Hi @UJESH2K 

That is amazing 🚀 
For we get it merged see:

>  Error: pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/resource_organizer.go:206:1: File is not properly formatted (gci)

Can you please just run make fix-lint and ensure that the changes are all rebased in 1 commit? 
Thank you a lot

UJESH2K:
Thank you so much 🙂
I’ve squashed the changes into a single commit and confirmed that make lint passes cleanly.
Please let me know if anything else is needed.

UJESH2K:
Hi @camilamacedo86, thanks for the patience 🙂

The CI failures I was hitting were coming from the E2E Testdata Sample. After digging into it, the root cause was that not all resources from the generated install.yaml were being consistently tracked during parsing.

Because the parser was operating on a partial set of resources, some manifests were effectively dropped before classification. This caused the Helm chart output to diverge from what the E2E tests expect, leading to the failures we were seeing.

To fix this, I updated parser.go so that all resources from install.yaml are parsed and tracked first. From there:

resources handled by existing Helm scaffolds behave exactly as before

any remaining, unhandled resources are preserved under templates/extras

This ensures nothing is silently dropped and makes the Helm output consistent and predictable, which also resolves the E2E testdata issues.

I’ve squashed everything into a single commit and confirmed CI is now running cleanly. Please let me know if anything else needs adjustment.

PR #5249, opened 2025-11-27
ci: introduce modernize linter

PascalBourdier:
add linter in golangci conf and fix code

fixes are done by `golangci-lint run --fix`

PR #5258, opened 2025-12-08
✨ (go/v4): Upgrade golangci-lint to v2.7.2 and add modernize check

dongjiang1989:
1. Replacing `interface{}` with `any` alias
2. Using `strings.Builder` instead of `+=`
3. Using `reflect.TypeFor[T]()` instead of `reflect.TypeOf((*T)(nil))`
4. Using `min()` and `max()` builtin function
5. Using `b.Loop()` in benchmarks

### DIFF
--- a/.github/workflows/lint-sample.yml
+++ b/.github/workflows/lint-sample.yml
@@ -43,7 +43,7 @@ jobs:
       - name: Run linter
         uses: golangci/golangci-lint-action@v9
         with:
-          version: v2.6.0
+          version: v2.7.2
           working-directory: ${{ matrix.folder }}
       - name: Run linter via makefile target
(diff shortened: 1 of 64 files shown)


dongjiang1989:
> Hi @dongjiang1989 thanks for working on this! I'm afraid this PR is a duplicate of #5249, but I have also noticed you changed more files, so perhaps your PR may supersede that one. Can you explain what's different between yours and 5249?

Thanks @vitorfloriano
In addition to optimizing kubebuilder, we also add modernize support for kubebuilder codegen.
1. Add codegen `.golangci.yml`
2. Add codegen golangci-lint check pass

dongjiang1989:
cc @camilamacedo86 @vitorfloriano How do I fix go-apidiff? 😑

PR #5265, opened 2025-12-10
🌱  Change `sort` to `slices` package

dongjiang1989:
Replace the sort package with the slices package to simplify the code.

Like:
https://github.com/kubernetes-sigs/controller-runtime/pull/3370
https://github.com/kubernetes-sigs/controller-tools/pull/1299

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,6 +7,7 @@ linters:
     - asciicheck
     - bidichk
     - copyloopvar
+    - depguard
     - dupl
     - errcheck
     - ginkgolinter
@@ -29,6 +30,12 @@ linters:
     - wrapcheck
     - whitespace
   settings:
+    depguard:
+      rules:
+        forbid-pkg-errors:
+          deny:
+            - pkg: sort
+              desc: Should be replaced with slices package
     ginkgolinter:
       forbid-focus-container: true
       forbid-spec-pollution: true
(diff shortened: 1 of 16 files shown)


pkg/plugins/golang/v4/scaffolds/internal/templates/golangci.go
```diff
@@ -66,6 +67,12 @@ linters:
     - unparam
     - unused
   settings:
+    depguard:
+      rules:
+        forbid-pkg-errors:
+          deny:
+            - pkg: sort
+              desc: Should be replaced with slices package
```
camilamacedo86: Should we really enforce it to end users?
If so, why? 

Note we can do this change for Kubebuilder code and tutorials, without change this file for the projects scaffolded with Kubebuilder

dongjiang1989: Thanks @camilamacedo86 
Noted. This is a strict requirement.
Revert the change for projects scaffolded with Kubebuilder.

dongjiang1989:
@camilamacedo86  Please re-check it

vitorfloriano (approved):
/lgtm

dongjiang1989:
@camilamacedo86  Maybe miss approved tag

Issue #5229, opened 2025-11-16
Add  `agents.md` file to the project scaffolds


camilamacedo86:
We should add an `agents.md` file to the project scaffolds so new projects always include the latest recommended practices for using agents.

Including this file in the scaffold ensures it stays updated automatically when testdata projects are regenerated, keeping all generated projects aligned with current common patterns and improving overall consistency and maintainability.

PR #5273, opened 2025-12-16
✨ (go/v4): Add AGENTS.md to the projects

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/5229

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/AGENTS.md
+++ b/docs/book/src/cronjob-tutorial/testdata/project/AGENTS.md
@@ -1,0 +1,282 @@
+# project - AI Agent Guide
+
+## Project Structure
+
+**Single-group layout (default):**
+```
+cmd/main.go                    Manager entry (registers controllers/webhooks)
+api/<version>/*_types.go       CRD schemas (+kubebuilder markers)
+api/<version>/zz_generated.*   Auto-generated (DO NOT EDIT)
+internal/controller/*          Reconciliation logic
+internal/webhook/*             Validation/defaulting (if present)
+config/crd/bases/*             Generated CRDs (DO NOT EDIT)
+config/rbac/role.yaml          Generated RBAC (DO NOT EDIT)
+config/samples/*               Example CRs (edit these)
+Makefile                       Build/test/deploy commands
+PROJECT                        Kubebuilder metadata Auto-generated (DO NOT EDIT)
+```
+
+**Multi-group layout** (for projects with multiple API groups):
+```
+api/<group>/<version>/*_types.go       CRD schemas by group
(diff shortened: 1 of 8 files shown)


camilamacedo86:
c/c @vitorfloriano could you please give a hand on the review

vitorfloriano (commented):
Couldn't we add instructions for building, packaging and deploying the project? 

Also, I think it could be useful to add a list with the necessary toolchain for development workflow (kind, kubectl, etc...) just so the agent checks that the user has all the tools needed before trying to work around constraints.

WDYT?

camilamacedo86:
Hi @vitorfloriano, thanks for the review 🙂

I tried to address your comment in a way that stays generic enough. We need to be careful not to be too opinionated, since users can customize their environments and workflows.

About build / package / deploy instructions: I’m not sure this adds much value for AI commands. Packaging and distribution are often project-specific and usually handled by release pipelines. There are many ways to do it, so it’s hard to define something that works well for most cases.

We should also avoid adding versions or very specific steps, as those can get outdated quickly.

For the toolchain list (kind, kubectl, etc.), it can be useful, but I think we should keep it light. Users may use different tools, and we don’t want to force a single workflow.

Overall, I tried to keep things helpful but flexible.

WDYT?

vitorfloriano (approved):
@camilamacedo86 Yeah, giving a second thought, I think keeping it generic is best. The addition of AGENTS.md is already a nice change that checks the "AI" box.

Maybe we could add a comment saying that this file was generated but users can customize it. Also, we could link to [agents.md](agents.md) where users can learn more about how to write their instructions.

But it is already good as is.

camilamacedo86:
Hi @vitorfloriano 

Thank you for checking it out.

> Maybe we could add a comment saying that this file was generated but users can customize it

It is the default behvaiour for all already.
When we do not want the end users touching at all then we say that.
For example, in the PROJECT file.

docs/book/src/cronjob-tutorial/testdata/project/AGENTS.md
```diff
@@ -0,0 +1,259 @@
...
+```
+
+## API Design (`api/<version>/*_types.go`)
+
+Use `+kubebuilder` markers to control CRD generation:
+
+```go
+// On types:
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status        // RECOMMENDED
```
camilamacedo86: TODO: Remove RECOMMEND

camilamacedo86:
Hi @vitorfloriano 

I did some changes to try shape and improve it.
Let me know wdyt?
Lets see if we can get this one merged

PR #5287, opened 2025-12-22
[helm/v2-alpha] Fix cert-manager Issuer reference after namePrefix application

### LINKED ISSUE #5280: [helm/v2-alpha] Missing name substitution in the cert-manager resources

In the chart templates, we have three cert-manager resources: two Certificates and a Issuer.
The certificates are referring to the issuer using this template:
```
    issuerRef:
        kind: Issuer
        name: {{ include "chart.name" . }}-selfsigned-issuer
```
While the issuer is named using this one, no template but hardcoded project name:
```
metadata:
    labels:
        app.kubernetes.io/name: test-project
    name: test-project-selfsigned-issuer
```
This lead to broken deployment when the chart is renamed.

I renamed the chart to `chart` to prevent weird names in the Github release (like `project/project/project` which are hard to differentiate from the other artifacts.

It is expected to have a consistent naming between the certificates and the issuer

### Reproducing this issue

To reproduce the issue, create a new project:
```
mkdir test-project && cd test-project
kubebuilder init --domain=sample.sh --owner "Tomtom" --repo "github.com/sample/project"
kubebuilder create api --version v1 --kind TestRes --controller --namespaced=false --resource
kubebuilder create webhook --version v1 --kind TestRes --defaulting --programmatic-validation
kubebuilder edit --plugins=helm/v2-alpha

# Rename and regenerate the chart
sed -iE 's/name: .*/name: chart/g' dist/chart/Chart.yaml
kubebuilder edit --plugins=helm/v2-alpha

# Render the chart locally and get only the reference / naming of the issuer
cd dist/chart
helm template . | grep selfsigned-issuer
```

You should end up with this output:
```
        name: chart-selfsigned-issuer
        name: chart-selfsigned-issuer
# Source: chart/templates/cert-manager/selfsigned-issuer.yaml
    name: test-project-selfsigned-issuer
```
Where we clearly see the inconsistency between the chart-selfsigned-issuer (from the certificates) and test-project-selfsigned-issuer (from the issuer)

### KubeBuilder (CLI) Version

4.10.0

### PROJECT version

3

### Plugin versions

```yaml
layout:
- go.kubebuilder.io/v4
plugins:
  helm.kubebuilder.io/v2-alpha:
    manifests: dist/install.yaml
    output: dist
```

### Other versions

go version go1.25.3 linux/amd64

github.com/onsi/ginkgo/v2 v2.22.0
github.com/onsi/gomega v1.36.1
k8s.io/apimachinery v0.34.1
k8s.io/client-go v0.34.1
sigs.k8s.io/controller-runtime v0.22.4

### Extra Labels

_No response_

abdulrahman11a:
Description
This PR fixes the cert-manager naming inconsistency reported in issue #5280 by adding kustomize replacements to dynamically link Certificate resources to the Issuer after namePrefix is applied.

The Problem
When using the helm/v2-alpha plugin, the following issue occurs:

config/default/kustomization.yaml applies namePrefix: test-project- to all resources.

The Issuer name becomes: test-project-selfsigned-issuer.

Certificate resources still reference the hardcoded name: selfsigned-issuer.

As a result, Certificates fail to find the Issuer, causing deployment failures.

This becomes especially problematic when users rename their Helm charts, as the issuerRef name wouldn't match the actual Issuer name.

Initial Approach (Incorrect)
I initially attempted to fix this by directly modifying the Helm templates in dist/chart/templates/cert-manager/ to use {{ include "chart.name" . }} for dynamic naming.

However, after review feedback pointing out that:

"every template is built from the kustomize files"

and that introducing template duplicates goes against the v2-alpha plugin's design philosophy, I realized this approach was incorrect.

The v2-alpha plugin is designed to generate Helm charts from kustomize output, not to maintain separate template logic. This incorrect fix would have been overwritten on every chart regeneration, making it unmaintainable.

The Correct Solution
Following the v2-alpha plugin's design philosophy of generating from kustomize output, this PR adds kustomize replacements in config/default/kustomization.yaml to:

Source the Issuer's actual name (after namePrefix is applied via fieldPath: .metadata.name).

Dynamically inject this name into all Certificate resources' spec.issuerRef.name field.

Use the same replacement pattern already established for webhook CA injection.

This ensures that Certificate issuerRef.name always matches the actual Issuer name, regardless of what namePrefix is configured.

Changes Made
Modified: config/default/kustomization.yaml

Added a new replacement block in the replacements: section:

yaml:
  source:
    kind: Issuer
    group: cert-manager.io
    version: v1
    name: selfsigned-issuer
    fieldPath: .metadata.name
  targets:
    - select:
        kind: Certificate
        group: cert-manager.io
        version: v1
        name: serving-cert
      fieldPaths:
        - .spec.issuerRef.name
      options:
        create: true
    - select:
        kind: Certificate
        group: cert-manager.io
        version: v1
        name: metrics-certs
      fieldPaths:
        - .spec.issuerRef.name
      options:
        create: true

This replacement:

Targets both serving-cert (webhook) and metrics-certs (metrics) certificates.

Uses kustomize's built-in field replacement mechanism.

Follows the existing pattern used for webhook service references.

Why This Approach?

✅ Fixes the source, not the symptom: modifies kustomize config rather than generated Helm templates.
✅ Aligns with v2-alpha design: charts are built from kustomize output, so fixing kustomize is correct.
✅ Maintainable: future chart regenerations automatically include the fix.
✅ Consistent: uses the same replacement pattern already present in the file.
✅ Zero template changes: no manual modifications to dist/chart/ files.
✅ Future-proof: won’t be overwritten by subsequent chart regenerations.

Motivation

Prevent deployment failures when users customize their Helm chart names.

Ensure cert-manager resources work correctly after namePrefix transformation.

Follow Kubebuilder best practices for the v2-alpha plugin architecture.

Maintain consistency with how webhook CA injection is already handled.

Respect the v2-alpha plugin's architecture of generating from kustomize, not duplicating logic.

image

Fixes #5280

### DIFF
--- a/config/certmanager/certificate-metrics.yaml
+++ b/config/certmanager/certificate-metrics.yaml
@@ -1,0 +1,20 @@
+# The following manifests contain a self-signed issuer CR and a metrics certificate CR.
+# More document can be found at https://docs.cert-manager.io
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  labels:
+    app.kubernetes.io/name: test-project
+    app.kubernetes.io/managed-by: kustomize
+  name: metrics-certs  # this name should match the one appeared in kustomizeconfig.yaml
+  namespace: system
+spec:
+  dnsNames:
+  # SERVICE_NAME and SERVICE_NAMESPACE will be substituted by kustomize
+  # replacements in the config/default/kustomization.yaml file.
+  - SERVICE_NAME.SERVICE_NAMESPACE.svc
+  - SERVICE_NAME.SERVICE_NAMESPACE.svc.cluster.local
+  issuerRef:
+    kind: Issuer
+    name: selfsigned-issuer
+  secretName: metrics-server-cert
--- a/config/certmanager/certificate-webhook.yaml
+++ b/config/certmanager/certificate-webhook.yaml
@@ -1,0 +1,20 @@
+# The following manifests contain a self-signed issuer CR and a certificate CR.
+# More document can be found at https://docs.cert-manager.io
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  labels:
+    app.kubernetes.io/name: test-project
+    app.kubernetes.io/managed-by: kustomize
+  name: serving-cert  # this name should match the one appeared in kustomizeconfig.yaml
+  namespace: system
+spec:
+  # SERVICE_NAME and SERVICE_NAMESPACE will be substituted by kustomize
+  # replacements in the config/default/kustomization.yaml file.
+  dnsNames:
+  - SERVICE_NAME.SERVICE_NAMESPACE.svc
+  - SERVICE_NAME.SERVICE_NAMESPACE.svc.cluster.local
+  issuerRef:
+    kind: Issuer
+    name: selfsigned-issuer
+  secretName: webhook-server-cert
--- a/config/certmanager/issuer.yaml
+++ b/config/certmanager/issuer.yaml
@@ -1,0 +1,13 @@
+# The following manifest contains a self-signed issuer CR.
+# More information can be found at https://docs.cert-manager.io
+# WARNING: Targets CertManager v1.0. Check https://cert-manager.io/docs/installation/upgrading/ for breaking changes.
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  labels:
+    app.kubernetes.io/name: test-project
+    app.kubernetes.io/managed-by: kustomize
+  name: selfsigned-issuer
+  namespace: system
+spec:
+  selfSigned: {}
--- a/config/certmanager/kustomization.yaml
+++ b/config/certmanager/kustomization.yaml
@@ -1,0 +1,7 @@
+resources:
+- issuer.yaml
+- certificate-webhook.yaml
+- certificate-metrics.yaml
+
+configurations:
+- kustomizeconfig.yaml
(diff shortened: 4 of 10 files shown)


abdulrahman11a:
> Hi @abdulrahman11a,
> 
> Thanks for taking a look and for opening the PR — I really appreciate the effort here. After reviewing the changes, though, I don’t think they address the underlying issue.
> 
> Adding a `config/` directory at the root of a Kubebuilder project will not cause the CLI to scaffold those resources automatically — especially not the `dist/chart/...` ones. Kubebuilder scaffolding is driven by plugins and templates, not by the presence of directories in the project tree.
> 
> For example, cert-manager assets are generated via templates, which you can see here: https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/certmanager
> 
> More generally, Kubebuilder works through templates under the `pkg` directory. If the intention is to change behavior or demonstrate the expected output, a more appropriate approach might be to add or update an example in the existing templates, for instance: https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config
> 
> We already have certs being generated today, such as in this template: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/certmanager/certificate_metrics.go#L49-L68
> 
> I need to take a bit more time to fully understand the problem being addressed before thinking through a proper solution, but as it stands, I don’t see how the current changes would resolve it.
> 
> Thanks again for working on this, and I’m happy to continue the discussion.




Thank you so much for taking the time to review my PR and for your detailed feedback. I really appreciate your guidance and patience in explaining the architecture and the right approach. Your insights are very helpful and gave me a clearer understanding of the issue.

Thanks again! 🙏

PR #5290, opened 2025-12-26
📖 Add 2026 Roadmap and Main Goals

camilamacedo86:


camilamacedo86:
@vitorfloriano

vitorfloriano (approved):
Nice to see `alpha update` in here.

Besides the goals listed, I still think the docs generation is too complicated and hard to update. Embedding prose into code is definitely not straightforward, IMO. If we ever wanted to add a new tutorial to the book, the effort involved would just not be worth it.

I know the book has been successful in on-boarding new users throughout the years, but maybe we could move on to something a bit more elaborate, **while keeping the book.**

I was thinking that perhaps we could move the docs from using mdBook (which is excellent for book format type of content) to something like **Hugo with Docsy theme**, which is used by many projects in the ecosystem (including [Kubernetes](https://kubernetes.io/blog/2020/06/better-docs-ux-with-docsy/), Operator-Framework, OpenTelemetry, Helm), [and it's backed by the CNCF](https://www.cncf.io/blog/2023/01/19/fast-and-effective-tools-for-cncf-and-open-source-project-websites/).

Migrating to Hugo+Docsy could allow us to have more sections in the docs. We could have a Blog to feature community operators or tips and tricks that do not need their own tutorial. We could add some additional documentation for the plugin ecosystem. In other words, the Hugo/Docsy combo is awesome and would allow us to do much more than simply have the book.

That being said, migrating the docs is a huge amount of work. So for now that may be just food for thought. 

The goals listed in here seem to be much more reasonable. 

LGTM

vitorfloriano:
@camilamacedo86 Another thing that I was thinking about is how Kubebuilder could adopt [Viper](https://github.com/spf13/viper) for configuration management.

Currently, you have to pass a lot of flags to kubebuilder and its subcommands/plugins. Viper would make it easier to adopt env vars, defaults, and even a config file to handle that.

Perhaps we could do some research on how Viper could improve kubebuilder, the pros and cons. WDYT?

camilamacedo86:
Hi @vitorfloriano,

Thank you for looking into that.

Regarding the Viper idea, I think it could be raised as an issue so we can properly analyze the pros and cons.

About the docs, I understand you see this approach as complicated, but overall it’s actually easier and requires less effort than updating the documentation for every PR. We’ve done this in the past already—trust me 😉. That said, I also think those suggestions might fit better as issues.

The goal for the docs remains the same as last year (2025): updating the samples and cronjob projects so they are closer to the "deploy image". It is missing some good practices since those are very old samples. I’ll probably work on that when I have some time. This is not about changing how things work, the tools, or the setup—it’s mainly about improving the samples and code examples themselves.

camilamacedo86:
Hi @vitorfloriano 

But please, lets create issues and let's check all ideas !!
I am totally open to them !!! 
Thank you a lot !!

camilamacedo86:
I am getting this one merged but if anyone want to change it.
Please, feel free to just push a PR proposing the updates !!

Thank you 🎉

PR #5288, opened 2025-12-24
🐛 fix(kustomize/v2): Add issuerRef replacement for cert-manager resources

abdulrahman11a:
Description
This PR fixes #5280
 by introducing a Kustomize replacement in the kustomize/v2 plugin template. It ensures that Certificate resources dynamically reference the actual Issuer name after namePrefix is applied, preventing deployment failures due to mismatched names.

Problem
When using kustomize/v2 with cert-manager resources:

config/default/kustomization.yaml applies a namePrefix to all resources.

The Issuer name becomes: {project-name}-selfsigned-issuer.

Certificates still reference the hardcoded name selfsigned-issuer.

Result: Certificates fail to locate the Issuer, causing deployments to fail.

This issue is particularly critical for the helm/v2-alpha plugin, which converts Kustomize output to Helm templates, resulting in broken charts.

Solution
Added a Kustomize replacement to the kdefault/kustomization.go template that:

Sources the Issuer's actual name (.metadata.name) after namePrefix is applied.

Dynamically injects this name into the spec.issuerRef.name field of Certificate resources.

Follows the existing pattern used for webhook CA injection.

Changes Made

Modified: pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go

Added replacement block:
replacements:
- source:
    kind: Issuer
    group: cert-manager.io
    version: v1
    name: selfsigned-issuer
    fieldPath: .metadata.name
  targets:
    - select:
        kind: Certificate
        group: cert-manager.io
        version: v1
        name: serving-cert
      fieldPaths:
        - .spec.issuerRef.name
      options:
        create: true
    - select:
        kind: Certificate
        group: cert-manager.io
        version: v1
        name: metrics-certs
      fieldPaths:
        - .spec.issuerRef.name
      options:
        create: true

Testing

Build modified Kubebuilder binary:

make build

Scaffold new project:

mkdir test-project && cd test-project
$KB_BIN init --domain=test.com --repo=test.com/test
$KB_BIN create api --group test --version v1 --kind Test --resource --controller
$KB_BIN create webhook --group test --version v1 --kind Test --defaulting --programmatic-validation

Verify replacement exists:

grep -A15 "selfsigned-issuer" config/default/kustomization.yaml

Test with Helm plugin:

$KB_BIN edit --plugins=helm/v2-alpha
cd dist/chart
helm template . | grep -A3 "kind: Issuer"
helm template . | grep -A3 "issuerRef:"

[...]

PR #5294, opened 2025-12-26
✨  (helm/v2-alpha): Add nameOverride/fullnameOverride support and fix cert-manager templating

camilamacedo86:
## Problem

When users renamed their Helm chart, cert-manager resources broke because the Issuer name was hardcoded while Certificate references were templated, causing a mismatch.

**Root cause:** Resource names were hardcoded with the project name instead of being templated, and there was confusion between chart identity (for labels) and resource naming (from kustomize).

## Solution

Adds standard Helm naming support (nameOverride/fullnameOverride) 
to the Helm v2-alpha plugin and implements dynamic resource name truncation 
to ensure Kubernetes 63-character compliance while maintaining cross-resource 
reference consistency.

Key changes:

1. Add nameOverride/fullnameOverride to values.yaml following Helm best practices
2. Simplify _helpers.tpl to match official Helm chart starter templates
3. Implement dynamic truncation: maxLen = 62 - len(suffix) (replaces hardcoded 45-char limit)
4. Fix cert-manager resource templating (Issuer names, issuerRef, annotations)
5. Add generic regex-based name substitution for all resource types
6. Update app.kubernetes.io/name label to use chart.name template

This ensures all generated Helm charts:
- Support custom naming via standard Helm values
- Respect Kubernetes 63-character name limits
- Maintain cross-resource reference consistency (Service ↔ Certificate ↔ Webhook ↔ Issuer)
- Work correctly with cert-manager (no more stuck certificates or failed issuance)
- Follow Helm and Kubernetes best practices
  
## Usage

Users can now do this:

```
helm install my-operator ./chart --set fullnameOverride=custom-name
```

```
# All resources will use "custom-name" prefix:
# - custom-name-controller-manager
# - custom-name-webhook-service
# - custom-name-selfsigned-issuer
# - etc.

# Before this change, it would break because:
# - Issuer would still reference "e2e-xxxx-selfsigned-issuer"
# - Certificates would reference wrong service names
# - Labels wouldn't match
```
 
<details> <summary><b>Click here to view design decisions</b></summary> 
#### The Kubernetes 63-Character Constraint
Kubernetes enforces a strict **63-character limit** on many resource names because they must conform to **DNS label rules (RFC 1123)**. This is a hard requirement and cannot be configured away.

Our Helm charts generate operator resources with inherently long suffixes, e.g.:
- `controller-manager-metrics-service` → **36 characters**
- `validating-webhook-configuration` → **34 characters**
- `leader-election-rolebinding` → **28 characters**

When these suffixes are combined with user-chosen `release` names or `fullnameOverride` values, **simple concatenation frequently exceeds 63 characters**, causing install-time validation failures.

#### The Cross-Resource Reference Problem
Kubebuilder-generated operators have dependency chains where **resource names must match exactly across multiple manifests**.

##### 1) Webhook Service Chain (4 locations must match)

```yaml
    # Location 1: Service Definition
    apiVersion: v1
    kind: Service
    metadata:
      name: <MUST-BE-IDENTICAL-1>

    # Location 2: TLS Certificate DNS Names
    spec:
      dnsNames:
        - <MUST-BE-IDENTICAL-1>.namespace.svc

    # Location 3: ValidatingWebhookConfiguration
    webhooks:
      - clientConfig:
          service:
            name: <MUST-BE-IDENTICAL-1>

    # Location 4: CRD Conversion Webhook
    spec:
      conversion:
        webhook:
          clientConfig:
            service:
              name: <MUST-BE-IDENTICAL-1>
```

Failure modes if these don’t match:
- TLS certificate doesn’t cover actual service DNS name → **TLS handshake fails**
- Webhook configuration references non-existent service → **connection refused**
- CRD conversion webhook can’t be reached → **multi-version operations fail**

##### 2) Issuer Reference Chain (3 locations must match)

```
    # Location 1: Issuer Definition
    apiVersion: cert-manager.io/v1
    kind: Issuer
    metadata:
      name: <MUST-BE-IDENTICAL-2>

    # Location 2: Webhook Certificate
    spec:
      issuerRef:
        name: <MUST-BE-IDENTICAL-2>

    # Location 3: Metrics Certificate
    spec:
      issuerRef:
        name: <MUST-BE-IDENTICAL-2>
```

Failure modes if these don’t match:
- cert-manager error: **"Issuer not found"**
- Certificates stuck in **Pending** indefinitely
- Pods fail to start waiting for certificate volumes

##### 3) cert-manager Annotation Chain (3 locations must match)

```
    # Location 1: Certificate Resource Name
    apiVersion: cert-manager.io/v1
    kind: Certificate
    metadata:
      name: <MUST-BE-IDENTICAL-3>
      namespace: system

    # Location 2: ValidatingWebhookConfiguration Annotation
    metadata:
      annotations:
        cert-manager.io/inject-ca-from: system/<MUST-BE-IDENTICAL-3>

    # Location 3: CRD Annotation (conversion webhooks)
    metadata:
      annotations:
        cert-manager.io/inject-ca-from: system/<MUST-BE-IDENTICAL-3>
```

Failure modes if these don’t match:
- cert-manager doesn’t inject CA bundle into webhook configurations
- Kubernetes API server rejects webhook: **"x509: certificate signed by unknown authority"**
- **ALL** custom resource create/update/delete operations fail

### Implementation

```
    {{- define "chart.resourceName" -}}
    {{- $fullname := include "chart.fullname" .context -}}
    {{- $suffix := .suffix -}}
    {{- $maxLen := sub 62 (len $suffix) | int -}}
    {{- if gt (len $fullname) $maxLen -}}
    {{- printf "%s-%s" (trunc $maxLen $fullname | trimSuffix "-") $suffix | trunc 63 | trimSuffix "-" -}}
    {{- else -}}
    {{- printf "%s-%s" $fullname $suffix | trunc 63 | trimSuffix "-" -}}
    {{- end -}}
    {{- end -}}
```

### How It Ensures Consistency
- **Deterministic algorithm:** same inputs → same output
- **Suffix-aware truncation:** computes `maxLen = 62 - len(suffix)` dynamically
- **Maximizes name preservation:** short suffixes allow longer base names
- **Applied uniformly:** all templates use the same helper for a given suffix

Example:

```
    fullnameOverride: "my-company-production-operator"          # 30 chars
    suffix:          "controller-manager-metrics-service"       # 36 chars

    maxLen = 62 - 36 = 26
    base_truncated = trunc(30, 26) = "my-company-production-o"  # 26 chars

    result = "my-company-production-o-controller-manager-metrics-service"  # 63 chars ✅
```

All resources using this suffix get the **identical** name:
- Service ✅
- Certificate ✅
- Webhook Config ✅
- RoleBindings ✅

## Why `dict` Syntax is Required
Helm (Go templates) allows **only one argument** to named templates:

```
    # ❌ INVALID: Cannot pass multiple arguments
    {{ include "chart.serviceName" "webhook-service" $ }}

    # ✅ VALID: Use dict to pass multiple values
    {{ include "chart.serviceName" (dict "suffix" "webhook-service" "context" $) }}
```

The `dict` produces a map like: `{"suffix": "webhook-service", "context": $}` which is accessed inside the helper as `.suffix` and `.context`.

### ❌ Alternative 1: Simple Concatenation

```
    name: {{ include "chart.fullname" . }}-webhook-service
```

Why rejected:
- No 63-character enforcement
- Install fails when limit exceeded
- No defense against long `fullnameOverride`

Real failure example:

```
    Error: INSTALLATION FAILED: Service "...-controller-manager-metrics-service" is invalid:
    metadata.name: Invalid value: must be no more than 63 characters
```

### ❌ Alternative 2: Truncate After Concatenation

```
    name: {{ printf "%s-webhook-service" (include "chart.fullname" .) | trunc 63 | trimSuffix "-" }}
```

Why rejected:
- Different suffix lengths truncate at different points
- Base names diverge across resources → cross-resource references break

Failure scenario:

```
    fullname: "my-very-long-production-operator-instance"   # 43 chars

    webhook suffix (short):  no truncation → base preserved
    metrics suffix (long):   truncation → base cut earlier

    Result: base names differ → cert refs / issuer refs / annotations mismatch ❌
```

### ❌ Alternative 3: Hardcoded Base Name Limit

```
    {{- if gt (len $fullname) 45 -}}
    {{- printf "%s-%s" (trunc 45 $fullname) .suffix | trunc 63 -}}
```

Why rejected:
- Arbitrary limit wastes available characters
- Short suffixes could allow much longer base names
- Unnecessary truncation frustrates users

### ❌ Alternative 4: One Helper Per Suffix

```
    {{- define "chart.name.webhook-service" -}}...{{- end -}}
    {{- define "chart.name.controller-manager" -}}...{{- end -}}
    # ... 20+ more helpers
```

Why rejected:
- Massive duplication
- Still risks inconsistent truncation patterns
- Unmaintainable, error-prone, violates DRY

## Benefits of Dynamic Truncation Design
- ✅ **Kubernetes Compliance:** guarantees names ≤ 63 chars across all user inputs
- ✅ **Cross-Resource Consistency:** same suffix → identical names everywhere (Service/Cert/Webhook/Issuer)
- ✅ **User Flexibility:** supports Helm `nameOverride` / `fullnameOverride`; maximizes name preservation dynamically
- ✅ **Production Reliability:** prevents subtle failures (TLS, webhook connectivity, cert-manager injection)
- ✅ **Industry-Standard Pattern:** aligns with mature charts that rely on deterministic helper naming and `dict` when consistency is critical

## Conclusion
Dynamic resource naming with `chart.serviceName` is not optional for Kubebuilder Helm charts. It is essential infrastructure that:
- Enforces the Kubernetes **63-character** name limit (hard requirement)
- Ensures cross-resource reference consistency (Serv

</details>

Closes: #5280 #5288


camilamacedo86:
Hi @robinlioret 

Could you please help us out with the review?

robinlioret:
Hello @camilamacedo86 
Yes, I can but not right now. I'll give it a look as soon as possible.

abdulrahman11a:
@camilamacedo86 
Thanks for pointing me to your PR — I’ve gone through it carefully.

I see now that handling the issue directly in the helm/v2-alpha templater ensures the generated Helm charts reference the correct Issuer name from the start, rather than relying on Kustomize to patch things afterward. That makes the behavior consistent regardless of whether Kustomize is used.

This helped me better understand how Kubebuilder expects fixes to be applied at the plugin level that owns the generated output. I appreciate the clear example and the comprehensive tests you added — it was very helpful for learning the intended approach.

Thanks again for the guidance.

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/cert-manager/metrics-certs.yaml
```diff
@@ -4,8 +4,8 @@ kind: Certificate
 metadata:
     labels:
         app.kubernetes.io/managed-by: {{ .Release.Service }}
-        app.kubernetes.io/name: project
-    name: project-metrics-certs
+        app.kubernetes.io/name: {{ include "chart.name" . }}
+    name: {{ include "chart.name" . }}-metrics-certs
```
robinlioret: I think we need to implement the recommended Helm labels as describe in the official documentation : https://helm.sh/docs/chart_best_practices/labels/#standard-labels

Wich could be something like:

```yaml
metadata:
  labels:
    app.kubernetes.io/name: {{ include "chart.name" . }}
    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
    app.kubernetes.io/instance: {{ .Release.Name }}
    app.kubernetes.io/version: {{ .Chart.AppVersion }}
```

The template prefix should be renamed with the chart's name: for a "thing-operator" chart, we should have `{{ include "thing-operator.name" }}`.

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/rbac/manager-rolebinding.yaml
```diff
@@ -3,7 +3,7 @@ kind: ClusterRoleBinding
 metadata:
     labels:
         app.kubernetes.io/managed-by: {{ .Release.Service }}
-        app.kubernetes.io/name: project
+        app.kubernetes.io/name: {{ include "chart.name" . }}
     name: project-manager-rolebinding
```
robinlioret: Should it be templated as well ?

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/rbac/leader-election-role.yaml
```diff
@@ -3,7 +3,7 @@ kind: Role
 metadata:
     labels:
         app.kubernetes.io/managed-by: {{ .Release.Service }}
-        app.kubernetes.io/name: project
+        app.kubernetes.io/name: {{ include "chart.name" . }}
     name: project-leader-election-role
```
robinlioret: Should it be templated as well ?

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/rbac/cronjob-viewer-role.yaml
```diff
@@ -4,7 +4,7 @@ kind: ClusterRole
 metadata:
     labels:
         app.kubernetes.io/managed-by: {{ .Release.Service }}
-        app.kubernetes.io/name: project
+        app.kubernetes.io/name: {{ include "chart.name" . }}
     name: project-cronjob-viewer-role
```
robinlioret: Should it be templated as well ?

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/rbac/cronjob-admin-role.yaml
```diff
@@ -4,7 +4,7 @@ kind: ClusterRole
 metadata:
     labels:
         app.kubernetes.io/managed-by: {{ .Release.Service }}
-        app.kubernetes.io/name: project
+        app.kubernetes.io/name: {{ include "chart.name" . }}
     name: project-cronjob-admin-role
```
robinlioret: Should it be templated as well ?

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
```diff
@@ -158,18 +158,41 @@ func (t *HelmTemplater) substituteCertificateDNSNames(yamlContent string, resour
...
+) string {
 	// Replace app.kubernetes.io/managed-by: kustomize with Helm template
 	// Use regex to handle different whitespace patterns
 	managedByRegex := regexp.MustCompile(`(\s*)app\.kubernetes\.io/managed-by:\s+kustomize`)
 	yamlContent = managedByRegex.ReplaceAllString(yamlContent, "${1}app.kubernetes.io/managed-by: {{ .Release.Service }}")
 
+	// Replace hardcoded project name in app.kubernetes.io/name label with chart name template
+	// This applies to ALL resources for consistency with Helm best practices
+	hardcodedNameLabel := "app.kubernetes.io/name: " + t.projectName
+	templatedNameLabel := "app.kubernetes.io/name: {{ include \"" + chartNameTemplate + "\" . }}"
```
robinlioret: This should remain hardcoded

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/manager/manager.yaml
```diff
@@ -3,22 +3,22 @@ kind: Deployment
 metadata:
     labels:
         app.kubernetes.io/managed-by: {{ .Release.Service }}
-        app.kubernetes.io/name: project
+        app.kubernetes.io/name: {{ include "chart.name" . }}
         control-plane: controller-manager
     name: project-controller-manager
```
robinlioret: Resources' name should be templated as well: `{{ include "chart.name" . }}-controller-manager`

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/cert-manager/metrics-certs.yaml
camilamacedo86: I think your suggestion is valid and definitely worth checking. **Thank you for raising it.**

**Could you please open an issue so we can look into this properly? We should analyze:**
- which labels we should support,
**- whether we need to add something new to Kustomize,**
**- and whether Helm charts need specific labels that should not exist for Kustomize.**

For Helm specifically, this might be a case like:
helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}

**So it would be great to track this in a separate issue and discuss it there.**

In this PR, we are only fixing the `app.kubernetes.io/name` label, since it comes from Kustomize and should use the chartName. This is directly related to the current issue/fix.

But we should definitely follow up on the broader label discussion.

Thank you again 🙏

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/manager/manager.yaml
camilamacedo86: Good catcher. I will do this fix.

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/cert-manager/metrics-certs.yaml
robinlioret: I have created #5304

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/manager/manager.yaml
camilamacedo86: Hi @robinlioret 

After investigating this further, I found the root cause **was a mismatch between resource names and their references. The issue occurred because we either have the namePrefix in the kustomize, so:**

1. **Labels** used the hardcoded project name
2. **Issuer name** was hardcoded
3. **Certificate issuerRef** was already templated with `chart.name`

When you renamed the chart, the issuerRef looked for `chart-selfsigned-issuer` but the Issuer was still named `test-project-selfsigned-issuer` → **mismatch**.

## Solution

To ensure consistency and respect kustomize's `namePrefix` configuration, I've implemented:

- **Added `namePrefix` to values.yaml** - Automatically detected from kustomize output
- **All resource names** now use `{{ .Values.namePrefix }}` 
- **Labels** use `{{ include "chart.name" . }}` for chart identity

This provides clear separation:
- **namePrefix** = resource naming (Deployment, Service, Issuer, etc.)
- **chart.name** = chart identity (labels only)

I updated the description to bring more clarity.
And in a follow up we can check the other labels

camilamacedo86:
Hi @robinlioret

Could you please give a look on this one?

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/cert-manager/metrics-certs.yaml
```diff
@@ -4,15 +4,15 @@ kind: Certificate
 metadata:
     labels:
         app.kubernetes.io/managed-by: {{ .Release.Service }}
-        app.kubernetes.io/name: project
-    name: project-metrics-certs
+        app.kubernetes.io/name: {{ include "chart.name" . }}
+    name: {{ .Values.namePrefix }}-metrics-certs
```
robinlioret: The prefix should be `{{ include "chart.fullname" . }}` to allow for standard name defaulting and for customization through the fullnameOverride value. Allowing the default "chart name" + "release name" because that will be used by tools like ArgoCD.

```
{{/*
Create a default fully qualified app name.
We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
If release name contains chart name it will be used as a full name.
*/}}
{{- define "chart.fullname" -}}
{{- if .Values.fullnameOverride }}
{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
{{- else }}
{{- $name := default .Chart.Name .Values.nameOverride }}
{{- if contains $name .Release.Name }}
{{- .Release.Name | trunc 63 | trimSuffix "-" }}
{{- else }}
{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
{{- end }}
{{- end }}
{{- end }}
```

The same way, the "chart.name" template should allow override through the nameOverride value.

```
{{/*
Expand the name of the chart.
*/}}
{{- define "chart.name" -}}
{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
{{- end }}
```

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
```diff
@@ -146,30 +150,75 @@ func (t *HelmTemplater) substituteCertificateDNSNames(yamlContent string, resour
...
+// substituteCertManagerReferences applies cert-manager specific template substitutions
+func (t *HelmTemplater) substituteCertManagerReferences(
+	yamlContent string,
+	resource *unstructured.Unstructured,
+) string {
+	kind := resource.GetKind()
 
+	if kind == kindIssuer || kind == kindCertificate {
+		hardcodedIssuerRef := t.detectedPrefix + "-selfsigned-issuer"
+		templatedIssuerRef := "{{ .Values.namePrefix }}-selfsigned-issuer"
```
robinlioret: Should be templated (see [this comment](https://github.com/kubernetes-sigs/kubebuilder/pull/5294/changes#r2661323571))

```go
templatedIssuerRef := "{{ include \"chart.fullname\" . }}-selfsigned-issuer"
```

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
```diff
@@ -178,6 +227,19 @@ func (t *HelmTemplater) substituteRBACValues(yamlContent string) string {
 	return yamlContent
 }
 
+// substituteCertManagerAnnotations replaces hardcoded certificate references in annotations
+func (t *HelmTemplater) substituteCertManagerAnnotations(yamlContent string) string {
+	hardcodedServingCert := t.detectedPrefix + "-serving-cert"
+	templatedServingCert := "{{ .Values.namePrefix }}-serving-cert"
```
robinlioret: Should "{{ .Values.namePrefix }}" be a constant ?
I think it would be a way to ensure consistency across all usages.

```go
const namePrefix = "{{ include \"chart.fullname\" $ }}"

// ...

func (t *HelmTemplater) substituteCertManagerAnnotations(yamlContent string) string {
    // ...
	templatedServingCert := namePrefix + "-serving-cert"
    // ...
}
```

Note: the use of "$" instead of "." makes sure the template uses the root of the parameters and avoid undesirable scope issues in the helm loops.

camilamacedo86:
Hi @robinlioret

Thank you a lot for all your help and support here.
You are totally right.

Note that we did the required changes
See the PR desc to check the design decisions choice. 

Could you please help us by checking this one?

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
```diff
@@ -91,17 +86,20 @@ Always uses the Helm release namespace.
...
 {{` + "`" + `{{/*
-Service name with proper truncation for Kubernetes 63-character limit.
-Takes a context with .suffix for the service type (e.g., "webhook-service").
-If fullname + suffix exceeds 63 chars, truncates fullname to 45 chars.
+Resource name with proper truncation for Kubernetes 63-character limit.
+Takes a dict with .suffix (resource name suffix) and .context (template context).
+Dynamically calculates safe truncation length based on suffix to ensure total <= 63 chars.
+Generic helper that works for any resource type.
 */}}` + "`" + `}}
 {{` + "`" + `{{- define "chart.serviceName" -}}` + "`" + `}}
```
robinlioret: serviceName could be misleading because it's used everywhere (not only for the Service resources)
Maybe something like "basename" could be a better name for this template.

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
robinlioret: I would like to suggest a more readable way to generate this template. Something nlike 

```go

func (f *HelmHelpers) SetTemplateDefaults() error {
    // ...
    f.TemplateBody = makeHelperTemplate()
    // ...
}

func makeHelperTemplate() string {
	escape := func(s string) string {
		return "{{`" + s + "`}}"
	}

	return strings.Join([]string{
		escape("{{/*"),
		"Chart name based on project name.",
		"Truncated to 63 characters for Kubernetes compatibility.",
		escape("*/}}"),
		escape("{{- define \"") + "{{ .ProjectName }}" + escape(".name\" -}}"),
		escape("{{- default .Chart.name .Values.nameOverride \"") + "{{ .ProjectName }}" + escape("\" | trunc 63 | trimSuffix \"-\""),
		escape("{{- end }}"),
		"",
       // ...
    }, "\n") + "\n" // EOL new line
}
```

Maybe it should go into a separate PR. I already have some code ready for this if needed (it seems too large for a single comment though).

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
```diff
@@ -51,34 +51,29 @@ func (f *HelmHelpers) SetTemplateDefaults() error {
 }
 
 const helmHelpersTemplate = `{{` + "`" + `{{/*
-Chart name based on project name.
-Truncated to 63 characters for Kubernetes compatibility.
+Expand the name of the chart.
 */}}` + "`" + `}}
 {{` + "`" + `{{- define "chart.name" -}}` + "`" + `}}
```
robinlioret: The templates should be named from the project name for a simple reason: unicity of the name.
In the edge case where a user use this chart as a dependency (or the other way around), we could end up with multiple templates with the same names. Leading to confusing and hard to troubleshoot errors.

Could be another PR if this one is getting too big.

Per example, not "chart.name" but "my-company-operator.name"

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/prometheus/controller-manager-metrics-monitor.yaml
```diff
@@ -4,9 +4,9 @@ kind: ServiceMonitor
 metadata:
     labels:
         app.kubernetes.io/managed-by: {{ .Release.Service }}
-        app.kubernetes.io/name: project
+        app.kubernetes.io/name: {{ include "chart.name" . }}
         control-plane: controller-manager
-    name: project-controller-manager-metrics-monitor
+    name: {{ include "chart.serviceName" (dict "suffix" "controller-manager-metrics-monitor" "context" $) }}
```
robinlioret: maybe shorten this to "controller-mentrics-monitor" ?

camilamacedo86: We cannot change that.
It came from kustomize.

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
camilamacedo86: +1 fixing

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
camilamacedo86: Yep, maybe a follow up here to see if we can simplify things would be great !!!

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
camilamacedo86: It is definilly something that we should address now.
thank you a lot for raising all those points !!!
I would like that we have all those major changes before next release so that people just fix their scaffold once.

camilamacedo86:
Hi @robinlioret 

All done. 
I think we are a good place to get all merged and unblock the other required needs for the next release
Thank you a lot !!! 
For all your amazing collab

Please, let us know if you spot anything else.

PR #5297, opened 2025-12-26
📖  Update cronjob samples to fetch Cronjob resource before update

camilamacedo86:
With this change we can close https://github.com/kubernetes-sigs/kubebuilder/issues/4291.
Then, the controller now is follow all practices applied in the Deploy Image. 
We were missing only the fetch and explain why it is required.

### DIFF
--- a/docs/book/src/multiversion-tutorial/testdata/project/internal/controller/cronjob_controller.go
+++ b/docs/book/src/multiversion-tutorial/testdata/project/internal/controller/cronjob_controller.go
@@ -151,7 +151,18 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			return ctrl.Result{}, err
 		}
 
-		// Re-fetch the CronJob after updating the status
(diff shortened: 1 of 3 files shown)


docs/book/src/multiversion-tutorial/testdata/project/internal/controller/cronjob_controller.go
vitorfloriano, line 1: I noticed that these comments are the same as the ones in the cronjob tutorial, but I imagine it doesn't matter because they won't appear in the book. Right?

camilamacedo86: They’re the same — the “multiversion” flow is essentially the cronjob sample, just with a few extra steps afterwards.

So we can reuse the existing cronjob code implementation and simply add the additional steps on top.

Also, no worries either way, since those comments won’t be rendered. If we ever feel there’s a need, we can follow up later and remove the comments.

PR #5299, opened 2025-12-27
✨ (kustomize/v2): Add NamespaceTransformer to support multi-namespace…

### LINKED ISSUE #5148: namespace override while running make deploy

There is an issue when calling make deploy, the issue is that the config/default/kustomization.yaml file.
it includes:
```
# Adds namespace to all resources.
namespace: xxxx
```
when you run `make deploy` it overrides all the namespaces as expected but it will cause issues when you are using some namespace based resources like the `rbac`.

### Reproducing this issue

i have these markers in my controller
```
// +kubebuilder:rbac:groups=apps,namespace=infrastructure,resources=deployments,verbs=get;list;watch;update;patch
// +kubebuilder:rbac:groups="",namespace=users,resources=secrets,verbs=get;
```
the rbac files that get generated via `make manifests` are fine too, but when you run `make deploy` this happens:

```
/home/alireza/playground/authelia/osiris/bin/kustomize build config/default
Error: namespace transformation produces ID conflict: [{"apiVersion":"rbac.authorization.k8s.io/v1","kind":"Role","metadata":{"annotations":{"internal.config.kubernetes.io/previousKinds":"Role","internal.config.kubernetes.io/previousNames":"manager-role","internal.config.kubernetes.io/previousNamespaces":"infrastructure"},"name":"manager-role","namespace":"users"},"rules":[{"apiGroups":["apps"],"resources":["deployments"],"verbs":["get","list","patch","update","watch"]}]} {"apiVersion":"rbac.authorization.k8s.io/v1","kind":"Role","metadata":{"name":"manager-role","namespace":"users"},"rules":[{"apiGroups":[""],"resources":["secrets"],"verbs":["get"]}]}]
```
it happens because of the override and the other role for `infrastructure` namespace goes to the `users` namespace. 
some work arounds are:

1. remove the namespace override in the `config/default/kustomization.yaml` and apply namespace to everything manually or make a transformers file that transforms the namespaces for specific resources
2. change every role name to be unique (still goes to the wrong namespace)

or any other solution that im missing, i would be happy to work on a solution.

### KubeBuilder (CLI) Version

4.9.0

### PROJECT version

3

### Plugin versions

```yaml
layout:
- go.kubebuilder.io/v4
```

### Other versions

_No response_

### Extra Labels

/kind documentation

AlirezaPourchali:
## Description

This PR fixes namespace conflicts when using namespace-scoped RBAC markers in multi-namespace scenarios.

### Problem
When using RBAC markers with explicit namespaces:
```go
// +kubebuilder:rbac:groups=apps,namespace=infrastructure,resources=deployments,verbs=get
// +kubebuilder:rbac:groups="",namespace=users,resources=secrets,verbs=get
```

The `controller-gen` correctly generates separate Role resources for each namespace. However, during `make deploy`, Kustomize's global `namespace:` field in `config/default/kustomization.yaml` overrides ALL namespaces, causing both Roles to end up in the same namespace with identical names → ID conflict error.

### Solution
Replaces the hardcoded `namespace:` field with a `NamespaceTransformer` using `unsetOnly: true`. This approach:
- ✅ Preserves explicit namespaces from RBAC markers
- ✅ Adds default namespace to resources that don't specify one
- ✅ Maintains backward compatibility (users can revert by uncommenting the old `namespace:` field)
- ✅ Follows Kustomize best practices per [NamespaceTransformer docs](https://kubectl.docs.kubernetes.io/references/kustomize/builtins/#_namespacetransformer_)

### Changes
1. **Created**: `pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/namespace_transformer.go`
   - Scaffolds `namespace-transformer.yaml` with `unsetOnly: true`
   
2. **Modified**: `pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go`
   - Commented out hardcoded `namespace:` field
   - Added `transformers:` section referencing `namespace-transformer.yaml`
   - Included documentation comments explaining the change and fallback option

3. **Modified**: `pkg/plugins/common/kustomize/v2/scaffolds/init.go`
   - Registered `NamespaceTransformer` in templates slice

### ⚠️ Known Side Effect: Helm Charts

The Helm chart generation was affected by this change. Some generated Helm templates changed from:
```yaml
```
to:
```yaml  
namespace: system
```

**Root cause**: The `metrics_service.yaml` template has a hardcoded `namespace: system` field. With `unsetOnly: true`, the NamespaceTransformer preserves this existing namespace. The Helm v2-alpha plugin's `substituteNamespace()` function in `helm_templater.go` expects the format `<projectname>-system` (e.g., `project-v4-system`) but encounters just `system`, so it doesn't convert it to `{{ .Release.Namespace }}`.

**Question for maintainers**: Should the Helm v2-alpha plugin be updated to also replace the `namespace: system` pattern? I can make that change if needed, or if there's a preferred alternative solution. The v1-alpha plugin handles this via `strings.ReplaceAll(contentStr, "namespace: system", "namespace: {{ .Release.Namespace }}")` in `edit.go:442`.

### Testing
- ✅ `make generate` - All testdata and docs regenerated successfully
- ✅ `make lint-fix` - No linting issues
- ✅ `make test-unit` - All unit tests pass
- ✅ Manual verification of generated `namespace-transformer.yaml` in test projects

### Backward Compatibility
Existing projects are unaffected. New projects created with `kubebuilder init` will use the NamespaceTransformer approach. Users who prefer the old behavior can:
1. Uncomment `namespace: {{ .ProjectName }}-system` in `config/default/kustomization.yaml`
2. Remove the `transformers:` section
3. Delete `namespace-transformer.yaml`

Fixes #5148

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/default/kustomization.yaml
@@ -1,4 +1,5 @@
 # Adds namespace to all resources.
+# If using namespace-scoped RBAC, see [NAMESPACE-TRANSFORMER] below.
 namespace: project-system
 
 # Value of this field is prepended to the
@@ -238,3 +239,10 @@ replacements:
 #     fieldPath: .metadata.name
 #   targets: # Do not remove or uncomment the following scaffold marker; required to generate code for target CRD.
 # +kubebuilder:scaffold:crdkustomizecainjectionname
+
+# [NAMESPACE-TRANSFORMER] To use the NamespaceTransformer instead of the
+# "namespace:" field (e.g. for multi-namespace RBAC), comment out "namespace:"
+# above and uncomment the transformers section below.
+# See config/default/namespace_transformer.yaml for details.
+#transformers:
+#- namespace_transformer.yaml
--- a/docs/book/src/cronjob-tutorial/testdata/project/config/default/namespace_transformer.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/config/default/namespace_transformer.yaml
@@ -1,0 +1,39 @@
+# This NamespaceTransformer is an alternative to the
+# "namespace:" field in kustomization.yaml.
+# While the "namespace:" field overrides ALL namespaces
+# (potentially breaking multi-namespace setups), this transformer
+# with "unsetOnly: true" only sets namespace on resources
+# that don't already have one.
+#
+# This is useful when using namespace-scoped RBAC markers such as:
+#   //+kubebuilder:rbac:groups=apps,namespace=infrastructure,...
+# which generate Roles with explicit namespaces that should
+# be preserved.
+#
+# TIP: When using multiple namespaces, also consider using the
+# "roleName" parameter in your RBAC markers to give each Role
+# a unique name, e.g.:
+#   //+kubebuilder:rbac:namespace=infra,roleName=infra-role,...
+# This avoids ID conflicts even before switching to the
+# NamespaceTransformer.
+#
+# Usage:
+#   1. Comment out the "namespace:" field in kustomization.yaml
+#   2. Uncomment the "transformers:" section in kustomization.yaml
+#   3. Remove "namespace: system" from any resource files under
+#      config/ that should receive the default namespace from
+#      this transformer instead
+#
+# More info:
+#   https://kubectl.docs.kubernetes.io/references/kustomize/builtins/
+#   #_namespacetransformer_
+apiVersion: builtin
+kind: NamespaceTransformer
+metadata:
+  name: namespace-transformer
+  namespace: project-system
+setRoleBindingSubjects: allServiceAccounts
+unsetOnly: true
+fieldSpecs:
+  - path: metadata/namespace
+    create: true
--- a/docs/book/src/getting-started/testdata/project/config/default/kustomization.yaml
+++ b/docs/book/src/getting-started/testdata/project/config/default/kustomization.yaml
@@ -1,4 +1,5 @@
 # Adds namespace to all resources.
+# If using namespace-scoped RBAC, see [NAMESPACE-TRANSFORMER] below.
 namespace: project-system
 
 # Value of this field is prepended to the
@@ -232,3 +233,10 @@ patches:
 #     fieldPath: .metadata.name
 #   targets: # Do not remove or uncomment the following scaffold marker; required to generate code for target CRD.
 # +kubebuilder:scaffold:crdkustomizecainjectionname
+
+# [NAMESPACE-TRANSFORMER] To use the NamespaceTransformer instead of the
+# "namespace:" field (e.g. for multi-namespace RBAC), comment out "namespace:"
+# above and uncomment the transformers section below.
+# See config/default/namespace_transformer.yaml for details.
+#transformers:
+#- namespace_transformer.yaml
--- a/docs/book/src/getting-started/testdata/project/config/default/namespace_transformer.yaml
+++ b/docs/book/src/getting-started/testdata/project/config/default/namespace_transformer.yaml
@@ -1,0 +1,39 @@
+# This NamespaceTransformer is an alternative to the
+# "namespace:" field in kustomization.yaml.
+# While the "namespace:" field overrides ALL namespaces
+# (potentially breaking multi-namespace setups), this transformer
+# with "unsetOnly: true" only sets namespace on resources
+# that don't already have one.
+#
+# This is useful when using namespace-scoped RBAC markers such as:
+#   //+kubebuilder:rbac:groups=apps,namespace=infrastructure,...
+# which generate Roles with explicit namespaces that should
+# be preserved.
+#
+# TIP: When using multiple namespaces, also consider using the
+# "roleName" parameter in your RBAC markers to give each Role
+# a unique name, e.g.:
+#   //+kubebuilder:rbac:namespace=infra,roleName=infra-role,...
+# This avoids ID conflicts even before switching to the
+# NamespaceTransformer.
+#
+# Usage:
+#   1. Comment out the "namespace:" field in kustomization.yaml
+#   2. Uncomment the "transformers:" section in kustomization.yaml
+#   3. Remove "namespace: system" from any resource files under
+#      config/ that should receive the default namespace from
+#      this transformer instead
+#
+# More info:
+#   https://kubectl.docs.kubernetes.io/references/kustomize/builtins/
+#   #_namespacetransformer_
+apiVersion: builtin
+kind: NamespaceTransformer
+metadata:
+  name: namespace-transformer
+  namespace: project-system
+setRoleBindingSubjects: allServiceAccounts
+unsetOnly: true
+fieldSpecs:
+  - path: metadata/namespace
+    create: true
--- a/docs/book/src/multiversion-tutorial/testdata/project/config/default/kustomization.yaml
+++ b/docs/book/src/multiversion-tutorial/testdata/project/config/default/kustomization.yaml
@@ -1,4 +1,5 @@
 # Adds namespace to all resources.
+# If using namespace-scoped RBAC, see [NAMESPACE-TRANSFORMER] below.
 namespace: project-system
 
 # Value of this field is prepended to the
@@ -256,3 +257,10 @@ replacements:
          index: 1
          create: true
 # +kubebuilder:scaffold:crdkustomizecainjectionname
+
+# [NAMESPACE-TRANSFORMER] To use the NamespaceTransformer instead of the
+# "namespace:" field (e.g. for multi-namespace RBAC), comment out "namespace:"
+# above and uncomment the transformers section below.
+# See config/default/namespace_transformer.yaml for details.
+#transformers:
+#- namespace_transformer.yaml
--- a/docs/book/src/multiversion-tutorial/testdata/project/config/default/namespace_transformer.yaml
+++ b/docs/book/src/multiversion-tutorial/testdata/project/config/default/namespace_transformer.yaml
@@ -1,0 +1,39 @@
+# This NamespaceTransformer is an alternative to the
+# "namespace:" field in kustomization.yaml.
+# While the "namespace:" field overrides ALL namespaces
+# (potentially breaking multi-namespace setups), this transformer
+# with "unsetOnly: true" only sets namespace on resources
+# that don't already have one.
+#
+# This is useful when using namespace-scoped RBAC markers such as:
+#   //+kubebuilder:rbac:groups=apps,namespace=infrastructure,...
+# which generate Roles with explicit namespaces that should
+# be preserved.
+#
+# TIP: When using multiple namespaces, also consider using the
+# "roleName" parameter in your RBAC markers to give each Role
+# a unique name, e.g.:
+#   //+kubebuilder:rbac:namespace=infra,roleName=infra-role,...
+# This avoids ID conflicts even before switching to the
+# NamespaceTransformer.
+#
+# Usage:
+#   1. Comment out the "namespace:" field in kustomization.yaml
+#   2. Uncomment the "transformers:" section in kustomization.yaml
+#   3. Remove "namespace: system" from any resource files under
+#      config/ that should receive the default namespace from
+#      this transformer instead
+#
+# More info:
+#   https://kubectl.docs.kubernetes.io/references/kustomize/builtins/
+#   #_namespacetransformer_
+apiVersion: builtin
+kind: NamespaceTransformer
+metadata:
+  name: namespace-transformer
+  namespace: project-system
+setRoleBindingSubjects: allServiceAccounts
+unsetOnly: true
+fieldSpecs:
+  - path: metadata/namespace
+    create: true
--- a/pkg/plugins/common/kustomize/v2/scaffolds/init.go
+++ b/pkg/plugins/common/kustomize/v2/scaffolds/init.go
@@ -67,6 +67,7 @@ func (s *initScaffolder) Scaffold() error {
 	templates := []machinery.Builder{
 		&rbac.Kustomization{},
 		&kdefault.MetricsService{},
+		&kdefault.NamespaceTransformer{},
 		&rbac.MetricsAuthRole{},
 		&rbac.MetricsAuthRoleBinding{},
 		&rbac.MetricsReaderRole{},
--- a/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go
+++ b/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/kustomization.go
@@ -45,6 +45,7 @@ func (f *Kustomization) SetTemplateDefaults() error {
 }
 
 const kustomizeTemplate = `# Adds namespace to all resources.
+# If using namespace-scoped RBAC, see [NAMESPACE-TRANSFORMER] below.
 namespace: {{ .ProjectName }}-system
 
 # Value of this field is prepended to the
@@ -278,4 +279,11 @@ patches:
 #     fieldPath: .metadata.name
 #   targets: # Do not remove or uncomment the following scaffold marker; required to generate code for target CRD.
 # +kubebuilder:scaffold:crdkustomizecainjectionname
+
+# [NAMESPACE-TRANSFORMER] To use the NamespaceTransformer instead of the
+# "namespace:" field (e.g. for multi-namespace RBAC), comment out "namespace:"
+# above and uncomment the transformers section below.
+# See config/default/namespace_transformer.yaml for details.
+#transformers:
+#- namespace_transformer.yaml
 `
--- a/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/namespace_transformer.go
+++ b/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/namespace_transformer.go
@@ -1,0 +1,84 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kdefault
+
+import (
+	"path/filepath"
+
+	"sigs.k8s.io/kubebuilder/v4/pkg/machinery"
+)
+
+var _ machinery.Template = &NamespaceTransformer{}
+
+// NamespaceTransformer scaffolds a file that defines the NamespaceTransformer
+// for the default overlay folder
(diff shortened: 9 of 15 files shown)


docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/prometheus/controller-manager-metrics-monitor.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
         app.kubernetes.io/name: project
         control-plane: controller-manager
     name: project-controller-manager-metrics-monitor
-    namespace: {{ .Release.Namespace }}
+    namespace: system
```
camilamacedo86: We cannot do those changes in the Helm Charts.

camilamacedo86 (changes requested):
Hi @AlirezaPourchali 👋

Thank you for raising this and for the work on it — much appreciated.

I think the first step should be opening a PR against controller-tools.
See this comment for context:
https://github.com/kubernetes-sigs/kubebuilder/issues/5148#issuecomment-3536838419

Once we have the fix in controller-runtime, we can then evaluate what is needed in kubebuilder. Having the fix in place will allow us to properly test and make sure the solution really works.

With the controller-runtime fix, I don’t think we need all of these changes in kubebuilder.
It seems we would only need one additional file:
config/default/namespace-transformer.yaml
(as shown in my example comment).

Also, instead of adding this to the default scaffold, it might be better to document it. This is probably not needed for ~90% of users, and adding a new default file may not help most people.

Possible alternatives:
- Add documentation in the FAQ: https://book.kubebuilder.io/faq
- Or create a small reference/tutorial explaining this use case

We could include a concrete example (similar to the CronJob references), with:
- a small mock project
- e2e tests
- and Helm validation
to ensure the approach works end-to-end.

What do you think? Does this approach make sense?

Thanks again for the contribution 🙏

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/prometheus/controller-manager-metrics-monitor.yaml
AlirezaPourchali: I've noted the Helm chart issue in the PR description and will handle it separately if needed. The core Kustomize fix is independent of Helm concerns.
I can handle it via v2-alpha plugin like the v1-alpha plugin.

Thanks again for your guidance! 🙏

AlirezaPourchali:
Hi @camilamacedo86 👋

Thank you for the feedback! I wanted to clarify the situation with controller-tools after testing extensively.

## Controller-gen Already Works Correctly ✅

I've verified that **controller-gen from controller-tools is NOT the issue** - it already correctly generates namespace fields in RBAC manifests. Here's the proof:

### Test Setup

I have a controller with namespace-specific RBAC markers:

```go
// +kubebuilder:rbac:groups=apps,namespace=infrastructure,resources=deployments,verbs=get;list;watch;update;patch
// +kubebuilder:rbac:groups="",namespace=production,resources=secrets,verbs=get
// +kubebuilder:rbac:groups=coordination.k8s.io,namespace=production,resources=leases,verbs=get;list;watch;create;update;patch;delete
// +kubebuilder:rbac:groups="",namespace=production,resources=events,verbs=get;list;watch;create;update;patch;delete
```

### Test Results

**Step 1: Fresh generation with controller-gen v0.19.0**

```bash
rm config/rbac/role.yaml
make manifests  # Uses controller-gen
cat config/rbac/role.yaml
```

**Output - controller-gen DOES generate namespace fields:**

```yaml
---
kind: Role
metadata:
  name: manager-role
  namespace: infrastructure    # ← Generated by controller-gen
rules:
- apiGroups:
  - apps
  resources:
  - deployments
  verbs:
  - get
  - list
  - patch
  - update
  - watch
---
kind: Role
metadata:
  name: manager-role
  namespace: production    # ← Generated by controller-gen
rules:
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - get
# ... more rules
```

**Step 2: Testing with Kustomize's hardcoded `namespace:` field**

```yaml
# config/default/kustomization.yaml
namespace: test-override
namePrefix: osiris-
resources:
- ../rbac
```

```bash
kustomize build config/default
```

**Result:**
```
Error: namespace transformation produces ID conflict: 
[{"kind":"Role","metadata":{"name":"manager-role","namespace":"test-override"}...},
 {"kind":"Role","metadata":{"name":"manager-role","namespace":"test-override"}...}]
```

The hardcoded `namespace:` field **overrides** both `infrastructure` and `production` to `test-override`, causing both Roles to have identical names in the same namespace → ID conflict.

**Step 3: Testing with NamespaceTransformer + `unsetOnly: true`**

```yaml
# config/default/namespace-transformer.yaml
apiVersion: builtin
kind: NamespaceTransformer
metadata:
  name: namespace-transformer
  namespace: test-override
setRoleBindingSubjects: none
unsetOnly: true    # Only transform if namespace not already set
fieldSpecs:
- path: metadata/namespace
  create: true
```

```yaml
# config/default/kustomization.yaml
namePrefix: osiris-
resources:
- ../rbac
transformers:
- namespace-transformer.yaml
```

```bash
kustomize build config/default
```

**Result: SUCCESS ✅** - Explicit namespaces are preserved:

```yaml
---
kind: Role
metadata:
  name: osiris-manager-role
  namespace: infrastructure    # ← Preserved
---
kind: Role
metadata:
  name: osiris-manager-role
  namespace: production    # ← Preserved
```

## Conclusion

1. **Controller-tools works correctly** - no PR needed there
2. **The issue is in Kubebuilder's Kustomize scaffolding** - the hardcoded `namespace:` field breaks multi-namespace RBAC
3. **This PR fixes the Kubebuilder scaffolding issue** by using NamespaceTransformer with `unsetOnly: true`

### Regarding Your Suggestion

You mentioned:
> I think the first step should be opening a PR against controller-tools

After this testing, I believe controller-tools doesn't need changes. The namespace field is already in the generated YAML.

> It seems we would only need one additional file: config/default/namespace-transformer.yaml

You're right that only one file is needed per project. However, this PR modifies how Kubebuilder **scaffolds** new projects - it generates that file automatically during `kubebuilder init`.

### Documentation vs Default Scaffolding

I understand your concern about this being needed for <10% of users. I'm happy to pivot to either approach:

**Option A: Documentation only** (your preference)
- Add FAQ entry showing how to manually create `namespace-transformer.yaml`
- Include example in tutorial/reference
- Don't change default scaffolding

**Option B: Default scaffolding** (current PR)
- All new projects get `namespace-transformer.yaml` by default
- Backward compatible (users can revert to hardcoded `namespace:`)
- Follows Kustomize best practices

Which approach would you prefer? I'm happy to adjust the PR accordingly.

### About the Helm Side Effect

I've noted the Helm chart issue in the PR description and will handle it separately if needed. The core Kustomize fix is independent of Helm concerns.

Thanks again for your guidance! 🙏

AlirezaPourchali:
Hi @camilamacedo86 👋

Thank you for the feedback! I wanted to clarify the situation with controller-tools after testing extensively.

## Controller-gen Already Works Correctly ✅

I've verified that **controller-gen from controller-tools is NOT the issue** - it already correctly generates namespace fields in RBAC manifests. Here's the proof:

### Test Setup

I have a controller with namespace-specific RBAC markers:

```go
// +kubebuilder:rbac:groups=apps,namespace=infrastructure,resources=deployments,verbs=get;list;watch;update;patch
// +kubebuilder:rbac:groups="",namespace=production,resources=secrets,verbs=get
// +kubebuilder:rbac:groups=coordination.k8s.io,namespace=production,resources=leases,verbs=get;list;watch;create;update;patch;delete
// +kubebuilder:rbac:groups="",namespace=production,resources=events,verbs=get;list;watch;create;update;patch;delete
```

### Test Results

**Step 1: Fresh generation with controller-gen v0.19.0**

```bash
rm config/rbac/role.yaml
make manifests  # Uses controller-gen
cat config/rbac/role.yaml
```

**Output - controller-gen DOES generate namespace fields:**

```yaml
---
kind: Role
metadata:
  name: manager-role
  namespace: infrastructure    # ← Generated by controller-gen
rules:
- apiGroups:
  - apps
  resources:
  - deployments
  verbs:
  - get
  - list
  - patch
  - update
  - watch
---
kind: Role
metadata:
  name: manager-role
  namespace: production    # ← Generated by controller-gen
rules:
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - get
# ... more rules
```

**Step 2: Testing with Kustomize's hardcoded `namespace:` field**

```yaml
# config/default/kustomization.yaml
namespace: test-override
namePrefix: osiris-
resources:
- ../rbac
```

```bash
kustomize build config/default
```

**Result:**
```
Error: namespace transformation produces ID conflict: 
[{"kind":"Role","metadata":{"name":"manager-role","namespace":"test-override"}...},
 {"kind":"Role","metadata":{"name":"manager-role","namespace":"test-override"}...}]
```

The hardcoded `namespace:` field **overrides** both `infrastructure` and `production` to `test-override`, causing both Roles to have identical names in the same namespace → ID conflict.

**Step 3: Testing with NamespaceTransformer + `unsetOnly: true`**

```yaml
# config/default/namespace-transformer.yaml
apiVersion: builtin
kind: NamespaceTransformer
metadata:
  name: namespace-transformer
  namespace: test-override
setRoleBindingSubjects: none
unsetOnly: true    # Only transform if namespace not already set
fieldSpecs:
- path: metadata/namespace
  create: true
```

```yaml
# config/default/kustomization.yaml
namePrefix: osiris-
resources:
- ../rbac
transformers:
- namespace-transformer.yaml
```

```bash
kustomize build config/default
```

**Result: SUCCESS ✅** - Explicit namespaces are preserved:

```yaml
---
kind: Role
metadata:
  name: osiris-manager-role
  namespace: infrastructure    # ← Preserved
---
kind: Role
metadata:
  name: osiris-manager-role
  namespace: production    # ← Preserved
```

## Conclusion

1. **Controller-tools works correctly** - no PR needed there
2. **The issue is in Kubebuilder's Kustomize scaffolding** - the hardcoded `namespace:` field breaks multi-namespace RBAC
3. **This PR fixes the Kubebuilder scaffolding issue** by using NamespaceTransformer with `unsetOnly: true`

### Regarding Your Suggestion

You mentioned:
> I think the first step should be opening a PR against controller-tools

After this testing, I believe controller-tools doesn't need changes. The namespace field is already in the generated YAML.

> It seems we would only need one additional file: config/default/namespace-transformer.yaml

You're right that only one file is needed per project. However, this PR modifies how Kubebuilder **scaffolds** new projects - it generates that file automatically during `kubebuilder init`.

### Documentation vs Default Scaffolding

I understand your concern about this being needed for <10% of users. I'm happy to pivot to either approach:

**Option A: Documentation only** (your preference)
- Add FAQ entry showing how to manually create `namespace-transformer.yaml`
- Include example in tutorial/reference
- Don't change default scaffolding

**Option B: Default scaffolding** (current PR)
- All new projects get `namespace-transformer.yaml` by default
- Backward compatible (users can revert to hardcoded `namespace:`)
- Follows Kustomize best practices

Which approach would you prefer? I'm happy to adjust the PR accordingly.

### About the Helm Side Effect

I've noted the Helm chart issue in the PR description and will handle it separately if needed. The core Kustomize fix is independent of Helm concerns.

Thanks again for your guidance! 🙏

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/prometheus/controller-manager-metrics-monitor.yaml
camilamacedo86: Could you please help me understand why these changes were required?

From what I can see, we can’t change this safely — if we merge it as-is, we’ll introduce a bug. The namespace needs to come from `.Release.Namespace`, since that value is defined at install time (e.g., when running `helm install --namespace=<value>`).

AlirezaPourchali: **To be clear: this Helm side effect is a bug that needs to be fixed.** The changes in the Helm charts are unintended consequences of the NamespaceTransformer approach. The core Kustomize fix is correct, but it exposed an existing inconsistency in how we handle the metrics service namespace.

## Why This Worked Before But Breaks Now

**Before (with hardcoded `namespace:` field):**
1. `metrics_service.yaml` generated with `namespace: system`
2. Kustomize's hardcoded `namespace: project-v4-system` **forcefully overrode** ALL namespaces
3. Final output: `namespace: project-v4-system` (not `system`)
4. Helm plugin saw `project-v4-system`, matched the pattern, replaced with `{{ .Release.Namespace }}` ✅

**Now (with NamespaceTransformer + `unsetOnly: true`):**
1. `metrics_service.yaml` generated with `namespace: system`
2. NamespaceTransformer sees namespace **already set**, so **preserves it** (because `unsetOnly: true`)
3. Final output: `namespace: system` (unchanged)
4. Helm plugin sees `system`, doesn't match `project-v4-system` pattern, no replacement ❌

**The aggressive namespace override was masking an existing inconsistency.** The `metrics_service.go` template should have been using the same pattern as other resources all along, but nobody noticed because the hardcoded `namespace:` field was forcing everything to the correct value.

Now that we use `unsetOnly: true` (which correctly preserves explicit namespaces for multi-namespace RBAC), the inconsistency is exposed.

## The Root Cause

The issue is in how `metrics_service.go` template is defined. Currently:

**File: `pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/kdefault/metrics_service.go`** (line 52)

```yaml
metadata:
  name: controller-manager-metrics-service
  namespace: system    # ← Hardcoded "system"
```

The Helm v2-alpha plugin's `substituteNamespace()` function expects the pattern `<projectname>-system` (e.g., `project-v4-system`):

**File: `pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go`** (line 114)

```go
func (t *HelmTemplater) substituteNamespace(yamlContent string, resource *unstructured.Unstructured) string {
	hardcodedNamespace := t.projectName + "-system"  // ← Expects "project-v4-system"
	namespaceTemplate := "{{ .Release.Namespace }}"
	
	yamlContent = strings.ReplaceAll(yamlContent, hardcodedNamespace, namespaceTemplate)
	return yamlContent
}
```

When it encounters just `system`, it doesn't match the expected pattern, so it doesn't get replaced with `{{ .Release.Namespace }}`.

## Two Options to Fix

### Option 1: Fix metrics_service.go Template (Recommended)

Change the hardcoded `system` to use the project name pattern:

```go
const metricsServiceTemplate = `apiVersion: v1
kind: Service
metadata:
  labels:
    app.kubernetes.io/name: {{ .ProjectName }}
  name: controller-manager-metrics-service
  namespace: {{ .ProjectName }}-system    # ← Use project name pattern
spec:
  ...
```

This way:
- Kustomize with NamespaceTransformer will preserve `{{ .ProjectName }}-system` (it's a template variable, not a literal namespace)
- Helm v2-alpha plugin will match the pattern and convert to `{{ .Release.Namespace }}`

### Option 2: Update Helm v2-alpha Plugin

Add a fallback in `substituteNamespace()` to also handle plain `system`:

```go
func (t *HelmTemplater) substituteNamespace(yamlContent string, resource *unstructured.Unstructured) string {
	hardcodedNamespace := t.projectName + "-system"
	namespaceTemplate := "{{ .Release.Namespace }}"
	
	yamlContent = strings.ReplaceAll(yamlContent, hardcodedNamespace, namespaceTemplate)
	
	// Fallback: also replace plain "system" (for metrics service)
	yamlContent = strings.ReplaceAll(yamlContent, "namespace: system", "namespace: "+namespaceTemplate)
	
	return yamlContent
}
```

## My Recommendation

I think **Option 1 is cleaner** - the `metrics_service.go` template should follow the same pattern as other resources. This would make the Helm plugin's logic simpler and more consistent.

**This needs to be fixed before merging this PR.** The core NamespaceTransformer changes are sound, but we can't introduce this Helm regression.

Should I:
1. Update `metrics_service.go` to use `{{ .ProjectName }}-system` pattern and regenerate testdata? (My preference)
2. Update the Helm v2-alpha plugin with the fallback for plain `system`?
3. Revert to the hardcoded `namespace:` field approach until we can coordinate both fixes?

I'm ready to implement whichever solution you prefer. Let me know and I'll get it done! 🙏

camilamacedo86:
Hi @AlirezaPourchali,

This is really helpful—thank you so much for the thorough explanations and for sharing your tests and research. I’d like a bit of time to review everything properly, and I’ll reply back here once I’ve gone through it.

AlirezaPourchali:
Sure, it's totally fine!
Thanks for the feedback, let me know when to implement the changes.

AlirezaPourchali:
Hi @camilamacedo86, Could you checkout my changes.

docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/prometheus/controller-manager-metrics-monitor.yaml
AlirezaPourchali: fixed the namespace override bug for helm charts

docs/book/src/cronjob-tutorial/testdata/project/config/certmanager/certificate-webhook.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
     app.kubernetes.io/name: project
     app.kubernetes.io/managed-by: kustomize
   name: serving-cert  # this name should match the one appeared in kustomizeconfig.yaml
-  namespace: system
+  namespace: project-system
```
camilamacedo86: That would mean we’d need to rename the namespace in all cases in the default scaffold, and I don’t think we should do that. The namespace should be centralised in the default config. 

I think we need to look at how to make it work with as few changes as possible.

Also, we should consider whether we really want this in the default scaffold, or if it’s better to provide it as sample test data and/or documentation under References.

AlirezaPourchali: Hi @camilamacedo86 👋

Thanks for the review — you're right. I've reworked the PR completely based on your feedback.

**What changed:**

The previous approach renamed `namespace: system` across ~15 template files. This new version touches only **3 source files** with **zero changes to any existing template files**:

1. **New file:** namespace_transformer.go — scaffolds `config/default/namespace_transformer.yaml` (not active by default)
2. **Modified:** kustomization.go — added a small `[NAMESPACE-TRANSFORMER]` commented-out section at the bottom (same pattern as `[WEBHOOK]`, `[CERTMANAGER]`, `[PROMETHEUS]`) and a one-line breadcrumb on the `namespace:` field
3. **Modified:** init.go — registered the new template (1 line)

**Default behavior is 100% unchanged.** The `namespace:` field stays active and centralized.

**Why scaffold instead of docs-only:**

The `namespace_transformer.yaml` contains all the detailed instructions — what the problem is, how `roleName` helps (from [controller-tools#1334](https://github.com/kubernetes-sigs/controller-tools/pull/1334)), and step-by-step opt-in. The kustomization.yaml itself only adds 8 lines total (2 at top + 6 at bottom).

The main motivation is discoverability: when someone hits the `namespace transformation produces ID conflict` error, they open `kustomization.yaml` and the `[NAMESPACE-TRANSFORMER]` breadcrumb is right there pointing them to the solution file. No docs searching needed.

The full PR diff is **15 files, 375 additions, 0 deletions** — the additions are the new `namespace_transformer.yaml` files in testdata/docs + 8 comment lines in each `kustomization.yaml`.

Happy to adjust further!

PR #5307, opened 2025-12-30
🐛 bugfix: Prevent Unintended Cert-Manager Removal from Production Clusters

mandarjog:
## Bug Fix: Prevent Unintended Cert-Manager Removal from Production Clusters

**Problem:**
The current implementation has a critical race condition that can delete cert-manager from real clusters. When `BeforeSuite` is interrupted (Ctrl-C, timeout, etc.) before setting `isCertManagerAlreadyInstalled = true`, the variable remains `false` by default. `AfterSuite` then executes cleanup and **uninstalls cert-manager even though the test never installed it**.

This has caused production outages.

**Root Cause:**
We're using `isCertManagerAlreadyInstalled` (what existed before) as a proxy for `certManagerWasInstalledByTest` (what we actually need to know). These are fundamentally different states with different default values when interrupted.

**The Fix:**
Track what the test *actually did* (installed cert-manager), not what it *discovered* (cert-manager was already there). Set the flag **after successful installation**, not after discovery.

**Safety Improvement:**
- **Before:** Interruption → defaults to `false` → **deletes cert-manager from your cluster**
- **After:** Interruption → defaults to `false` → **skips cleanup, leaving your cluster untouched**

The worst case is now failing to clean up our own installation (safe) instead of removing infrastructure from production (catastrophic).

**Variable Rename Rationale:**
While I can keep the variable name if strongly preferred, `certManagerWasInstalledByTest` more accurately represents what we're tracking and makes the safety logic self-documenting for future maintainers. The semantic difference matters when reasoning about defaults and failure modes.

This addresses issues like: https://github.com/kubernetes-sigs/kubebuilder/issues/4391 and others.

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/suite.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/suite.go
@@ -65,7 +65,7 @@ var (
 	// These variables are useful if CertManager is already installed, avoiding
 	// re-installation and conflicts.
 	skipCertManagerInstall = os.Getenv("CERT_MANAGER_INSTALL_SKIP") == "true"
-	// certManagerWasInstalledByTest will be set true is installed during the test.
+	// certManagerWasInstalledByTest will be set true if installed during test.
 	certManagerWasInstalledByTest = false
 
 	// projectImage is the name of the image which will be build and loaded


mandarjog:
> Thank you a lot 🥇
> 
> To avoid introducing a lot of changes in the scaffold—which makes it harder for users to upgrade their projects—could we keep the same variable names and only change what’s required for the bug fix? Since it’s a boolean, the current is prefix is still appropriate as well.

Thanks for the review. 
`isCertManagerAlreadyInstalled` and `certManagerWasInstalledByTest` are saying fundamentally different things, so I am not sure how to do it..
Do you want to minimize the lines for code change? or just not change a variable name?

pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/suite.go
```diff
@@ -65,8 +65,8 @@ var (
 	// These variables are useful if CertManager is already installed, avoiding
 	// re-installation and conflicts.
 	skipCertManagerInstall = os.Getenv("CERT_MANAGER_INSTALL_SKIP") == "true"
-	// isCertManagerAlreadyInstalled will be set true when CertManager CRDs be found on the cluster
-	isCertManagerAlreadyInstalled = false
+	// certManagerWasInstalledByTest will be set true if installed during test.
+	certManagerWasInstalledByTest = false
```
camilamacedo86, line 69: Why was cert-manager installed by the test?

If cert-manager is already installed on the cluster (regardless of whether it was installed by the test or not—probably not by the test itself since the test cleans it up), then our check will mark it as present and the variable will be set to true.

So I don’t think we should change the name.

pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/suite.go
```diff
@@ -100,20 +100,20 @@ var _ = BeforeSuite(func() {
...
 	// Setup CertManager before the suite if not skipped and if not already installed
 	if !skipCertManagerInstall {
-		By("checking if cert manager is installed already")
-		isCertManagerAlreadyInstalled = utils.IsCertManagerCRDsInstalled()
-		if !isCertManagerAlreadyInstalled {
+		By("checking if cert manager is installed already") 
+		if !utils.IsCertManagerCRDsInstalled() {
 			_, _ = fmt.Fprintf(GinkgoWriter, "Installing CertManager...\n")
 			Expect(utils.InstallCertManager()).To(Succeed(), "Failed to install CertManager")
+			certManagerWasInstalledByTest = true
```
camilamacedo86, line 107: What problem does this change solve? Aren’t we doing exactly the same thing as before?

In the original code, the flag defaults to false (so we don’t skip the install).
If we detect cert-manager CRDs on the cluster (regardless of how cert-manager was installed), we set it to true and skip the install.

So functionally, the behavior seems unchanged.

camilamacedo86 (changes requested):
Hi @mandarjog,

Thanks for the reply and the collaboration. I left my comments on the PR, but I’m not sure there’s a strong benefit to merging this as-is.

From what I can tell, the main change is renaming, while the underlying logic stays the same. I’d prefer to avoid unnecessary changes in the scaffolds because it makes upgrades harder for users. ( ideally we should only change scaffolds when we must to do so )

I agree we may need to review and improve the e2e tests, but that would be better done in one focused pass rather than through partial, incremental changes.

For now, I think we should only change what we must change: bug fixes, not names/terms.

pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/suite.go
mandarjog: I have the behavior change in PR description.

The key change is when the flag is set before the pr and after the PR.
AfterSuite is executed regardless of whether beforeSuite completes or not.

when the test gets a Ctlr-C before it has a chance to set global variable to  IsCertManagerAlreadyInstalled to true, AfterSuit gets run. 
So now aftersuit thinks that IsCertManagerAlreadyInstalled=false regardless of the reality since the code never got executed.

this is how after suit ends up uninstalling certManager from a cluster.

This behavior *has caused* outages, so this PR is not academic. 

The change: now we record that *we installed certManager* after we are successful in installing it.
so we will only uninstall what we have installed.

If Ctrl-C is hit after install succeeds but before we update the variable, then the worst case scenario is we don’t clean up.
This is much safer than removing cert manager from a real cluster.

mandarjog: @camilamacedo86 if you have a better/safer way to fix the problem then we should do that. I am not insisting on this specific solution.

however as the code stands today, there is a serious bug.

mandarjog:
@camilamacedo86 i have significantly updated the PR description to explain  the change. Please check it out. Perhaps  should have opened it as an issue too :-)

pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/suite.go
camilamacedo86: Thank you for all your collab and detailed explanations.
Let me check more details because the code here is not good so we might need to change all by once. I am just trying to avoid send changes that people need constantly change their layouts.

camilamacedo86:
Hi @mandarjog,

Re: this issue — I opened a PR to address it:
https://github.com/kubernetes-sigs/kubebuilder/pull/5329

The larger diff is expected. Since we changed the scaffold, we had to run `make generate` and update all samples under `testdata` (these are also used in the docs).

Also, the tutorial docs inject Prometheus as an example, so we updated the hack/docs flow to match this approach too.

Since we may need to touch this again, I think it’s a good chance to refactor the implementation a bit. The current approach isn’t great, and we can make it easier so users only need to apply this kind of change once. I also added the manual steps in the PR description.

If you have a moment, I’d really appreciate your review. Thanks a lot for raising this! 🙏 

Would you mind if we close the original issue/PR in favor of this PR https://github.com/kubernetes-sigs/kubebuilder/pull/5329?

mandarjog:
Yes absolutely. I will close in in favor of the other PR, and will review the PR too. Thanks . 

Superseded by https://github.com/kubernetes-sigs/kubebuilder/pull/5329

PR #5327, opened 2026-01-01
🐛 (fix): AGENTS.md generated with extra \ and missing hub info

camilamacedo86:
We need to do those fixes before release


vitorfloriano (approved):
/lgtm

PR #5329, opened 2026-01-02
🐛 (go/v4): Fix E2E suite race condition that could delete cert-manager from environments inadvertently

### LINKED ISSUE #4391: E2E K8s Context Protection

camilamacedo86:
## Problem

The e2e suite could delete cert-manager from real clusters if interrupted (Ctrl-C, timeout, panic).

**Bug:** Variable `isCertManagerAlreadyInstalled` tracked discovery ("was it there?") instead of action ("did we install it?"). If interrupted before the check, it defaulted to `false`, causing cleanup to remove existing cert-manager.

**Impact:** Production outages when tests run on non-isolated clusters or get interrupted.

**Note: E2e tests should ONLY run on isolated environments, but this protects users who may accidentally run them on real clusters.**

## Fix

- Track action instead of discovery: `shouldCleanupCertManager`
- Set flag **before** installation (handles interruptions)
- Only cleanup what we installed
- Applied same pattern to Prometheus

**Safety:**
- Before: Interruption → deletes cert-manager ❌
- After: Interruption → skips cleanup ✅

## Migration (Recommended)

**Not breaking** - only affects new projects, but strongly recommended for safety.

1. Replace e2e suite file

```
curl -o test/e2e/e2e_suite_test.go \
  https://raw.githubusercontent.com/kubernetes-sigs/kubebuilder/v4.11.0/testdata/project-v4/test/e2e/e2e_suite_test.go
```

2. Update imports (adjust paths)

```
sed -i 's|sigs.k8s.io/kubebuilder/testdata/project-v4|YOUR_MODULE|g' test/e2e/e2e_suite_test.go
sed -i 's|example.com/project-v4|YOUR_IMAGE|g' test/e2e/e2e_suite_test.go
```

3. In test/e2e/e2e_test.go: replace projectImage → managerImage

## Benefits

- ✅ Prevents cert-manager deletion from existing clusters
- ✅ Handles interruptions safely
- ✅ Cleaner, more maintainable code

Fixes #4391 and related issues
Related PR: #5307 (inspiration for this fix)
Closes: https://github.com/kubernetes-sigs/kubebuilder/pull/5307

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/suite.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/suite.go
@@ -49,9 +49,9 @@ package e2e
 
 import (
 	"fmt"
-	"testing"
 	"os"
 	"os/exec"
+	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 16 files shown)


camilamacedo86:
@mandarjog 

Could you please help us in the review?

pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/suite.go
```diff
@@ -60,62 +60,68 @@ import (
...
-	// isCertManagerAlreadyInstalled will be set true when CertManager CRDs be found on the cluster
-	isCertManagerAlreadyInstalled = false
-
-	// projectImage is the name of the image which will be build and loaded
-	// with the code source changes to be tested.
-	projectImage = "example.com/{{ .ProjectName }}:v0.0.1"
+	// managerImage is the manager image to be built and loaded for testing.
+	managerImage = "example.com/{{ .ProjectName }}:v0.0.1"
+	// shouldCleanupCertManager tracks whether CertManager was installed by this suite.
+	shouldCleanupCertManager = false
```
mandarjog, line 66: +1 safe default.
 slightly different than my PR, but this also ensures better cleanup under interruption.

pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/suite.go
```diff
@@ -60,62 +60,68 @@ import (
...
+		return
+	}
+
+	By("checking if CertManager is already installed")
+	if utils.IsCertManagerCRDsInstalled() {
+		return
+	}
+
+	// Mark for cleanup before installation to handle interruptions and partial installs.
+	shouldCleanupCertManager = true
```
mandarjog, line 111: +1 partial installs will be cleaned up.
Safe default.

camilamacedo86:
Hi @mandarjog — thanks for the careful review!

**Re: kubeconfig/context**

Agreed it's a footgun, but it's **standard behavior across the ecosystem**. If someone has their default context pointing at production, *any* kubectl-based tool (helm, terraform, kubectl itself) can target prod unless explicitly pinned with `--context`.

**The difference:** Our scaffolded e2e tests actually **fail immediately on non-Kind** before any operations:

```go
By("loading the manager image on Kind")
err = utils.LoadImageToKindClusterWithName(managerImage)
ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Failed to load the manager image into Kind")
// ❌ This fails on non-Kind clusters: "kind cluster not found"
```

For example:

```bash
# If developer has production as default:
$ kubectl config current-context
production-us-east-1

# ALL these run on production:
$ helm install myapp ./chart          # Deploys to production
$ terraform apply                     # Applies to production
$ kubectl apply -f manifest.yaml      # Applies to production
$ make test-e2e                       # ❌ FAILS at image load (not on Kind)
```

You can customize your own scaffold. But is hard for us add a code that could work well for all. Anyway, 
I’ve opened a follow-up issue so we can discuss it: https://github.com/kubernetes-sigs/kubebuilder/issues/5335 we can discuss it further there. 

Thanks for the collaboration! 🙌

PR #5333, opened 2026-01-03
🐛 (autoupdate/v1-alpha): Make GitHub Models opt-in to prevent 403 errors

camilamacedo86:
Default scaffold now works without GitHub Models permissions, preventing
403 Forbidden errors for most users. AI summaries remain available as an
opt-in feature.

**Changes:**
- Remove `--use-gh-models` from default workflow scaffold
- Add `--use-gh-models` flag to `init` and `edit` commands
- Track `useGHModels` preference in PROJECT file (like helm/v2-alpha)
- Add runtime warning when `alpha update` runs without GH Models enabled
- Update `alpha generate` to preserve user's GH Models preference
- Update the documentation accordingly
- Add support to reenarate the projects with this option

### DIFF
--- a/docs/book/src/plugins/available/autoupdate-v1-alpha.md
+++ b/docs/book/src/plugins/available/autoupdate-v1-alpha.md
@@ -8,7 +8,7 @@ This automation uses the [`kubebuilder alpha update`][alpha-update-command] comm
 refresh your project scaffold, and wraps it in a GitHub Actions workflow that opens an **Issue** with a **Pull Request compare link** so you can create the PR and review it.
 
 <aside class="warning">
-<h1>Protect your branches</h1>
+<h3>Protect your branches</h3>
 
 This workflow by default **only** creates and pushes the merged files to a branch
 called `kubebuilder-update-from-<from-version>-to-<to-version>`.
@@ -20,17 +20,16 @@ changes aren't pushed or merged without proper review.
 
 ## When to Use It
 
-- When you don’t deviate too much from the default scaffold — ensure that you see the note about customization [here](https://book.kubebuilder.io/versions_compatibility_supportability#project-customizations).
+
 - When you want to reduce the burden of keeping the project updated and well-maintained.
-- When you want to guidance and help from AI to know what changes are needed to keep your project up to date
-as to solve conflicts.
+- When you want guidance and help from AI to know what changes are needed to keep your project up to date and to solve conflicts (requires `--use-gh-models` flag and GitHub Models permissions).
 
 ## How to Use It
 
 - If you want to add the `autoupdate` plugin to your project:
 
 ```shell
-kubebuilder edit --plugins="autoupdate.kubebuilder.io/v1-alpha"
+kubebuilder edit --plugins="autoupdate/v1-alpha"
 ```
 
 - If you want to create a new project with the `autoupdate` plugin:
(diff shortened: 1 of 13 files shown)


camilamacedo86:
@vitorfloriano WDYT?
Could you please help in the review?

docs/book/src/plugins/available/autoupdate-v1-alpha.md
```diff
@@ -20,17 +20,16 @@ changes aren't pushed or merged without proper review.
...
-as to solve conflicts.
+- When you want guidance and help from AI to know what changes are needed to keep your project up to date and to solve conflicts (requires `--use-gh-models` flag and GitHub Models permissions).
 
 ## How to Use It
 
 - If you want to add the `autoupdate` plugin to your project:
 
 ```shell
-kubebuilder edit --plugins="autoupdate.kubebuilder.io/v1-alpha"
+kubebuilder edit --plugins="autoupdate/v1-alpha"
```
vitorfloriano, line 32: Why this change?

docs/book/src/plugins/available/autoupdate-v1-alpha.md
```diff
@@ -39,43 +38,135 @@ kubebuilder edit --plugins="autoupdate.kubebuilder.io/v1-alpha"
...
 
-If conflicts arise, AI-generated comments call them out and provide next steps, such as:
+**Conflict help** (when needed):
 
 <img width="600" height="188" alt="Conflicts" src="https://github.com/user-attachments/assets/2142887a-730c-499a-94df-c717f09ab600" />
 
-### Workflow details
+## Troubleshooting
+
+#### If you so the error 403 Forbidden Error
```
vitorfloriano:
```suggestion
#### If you get the 403 Forbidden Error
```

vitorfloriano (commented):
I just commented a nitpick and a question, but overall looks good. AI features should always be opt-in instead of opt-out.

I took the branch for a test and _almost_ everything works. 

The only routine that did not work is when the user has already scaffolded the `auto_update.yml` and then later runs `kubebuilder edit --plugins="autoupdate/v1-alpha" --use-gh-models`. The `PROJECT` file gets updated with `useGHModels: true`, but the workflow does not update with the `models: read` permission and the `--use-gh-models` flag. Is that the expected behavior?

docs/book/src/plugins/available/autoupdate-v1-alpha.md
camilamacedo86: To keep short and simple. 
Why we need to keep the full name instead of key?

camilamacedo86:
Hi @vitorfloriano 

Thank you for looking on this one and your amazing review 🎉 
Regards :

> The only routine that did not work is when the user has already scaffolded the auto_update.yml and then later runs kubebuilder edit --plugins="autoupdate/v1-alpha" --use-gh-models. The PROJECT file gets updated with useGHModels: true, but the workflow does not update with the models: read permission and the --use-gh-models flag. Is that the expected behavior?

That was amazing catcher. No, that was wrong.
The fix is the template. 

Instead of use `f.IfExistsAction = machinery.SkipFile` we should use `f.IfExistsAction = machinery.OverwriteFile` so that the file is overwrite when we re run the command

Could you please check? 
It should be fixed now.

vitorfloriano:
> Hi @vitorfloriano
> ...
> Could you please check? It should be fixed now.

Fixed!

/lgtm

PR #5342, opened 2026-01-03
📖 (doc) Streamline quick start plugins and next steps sections

camilamacedo86:


docs/book/src/quick-start.md
```diff
@@ -272,49 +272,35 @@ make undeploy
...
 
-- Proceed with the [Getting Started Guide][getting-started], which should take no more than 30 minutes and will
-  provide a solid foundation.
-- Afterward, dive into the [CronJob Tutorial][cronjob-tutorial] to deepen your
-  understanding by developing a demo project.
-- Ensure that you understand the APIs and Groups concepts [Groups and Versions and Kinds, oh my!][gkv-doc]
-before designing your own API and project.
+- [Getting Started Guide][getting-started] (~30 min) - build a solid foundation
+- [CronJob Tutorial][cronjob-tutorial] - learn by building a demo project
+- [Groups, Versions, and Kinds][gkv-doc] - understand API design concepts
```
camilamacedo86, line 303: @vitorfloriano 

Here I am trying to just be less verbose since is a quick start
WDYT?

vitorfloriano: It's good, but I think it lost some of that sequential meaning (do this > then do that > finally, do this).

Idk if there is a logical sequence in the tutorials, like, do they build upon previous knowledge or could they be taken independently?

If the next logical step is the Getting Started Guide, we could remove the other two.

camilamacedo86: That depends on the person.

- If they want a basic foundation (and have no prior knowledge), they should start with **Getting Started**.
- If they want something deeper, they should go to **Tutorials** for more detail and advanced topics.
- If they want to understand what an **API** is (since it was mentioned above), they should check the **What is an API?** section.

The idea here is more give a direction, I think.

PR #5345, opened 2026-01-04
🐛 (CLI): Fix error unwrapping in updateFileModel for missing files

camilamacedo86:
Use errors.Is instead of os.IsNotExist to properly detect missing files with wrapped errors, fixing IgnoreFile action for inserters.


camilamacedo86:
it is a very small bug.
We are testing with new tests now.
I will not make nobody waste time on this review because we have a lot of PRs that we really need help with.
So, I am moving forward.

PR #5346, opened 2026-01-04
🌱 (test): Add missing kustomize/v2 and autoupdate plugins to coverage report

camilamacedo86:


camilamacedo86:
Just to cover all.
We have many PRs that requires review which we really need help with.
I will not make nobody waste time in this silly one
Moving forward.

PR #5347, opened 2026-01-04
🌱 Increase coverage for the CLI/API

camilamacedo86:


camilamacedo86:
Just add tests.
We have many PRs that requires review which we really need help with.
I will not make nobody waste time in this silly one
Moving forward.

PR #5348, opened 2026-01-04
🌱 (test): Fix coverage config and exclude template directories

camilamacedo86:
Excluded: Template directories (mostly static content, tested via E2E)


camilamacedo86:
Just fixing the config for coveralls show an accurate info.
Many files are just static will never have the tests + helm v1 is deprecated we should not care about.

We have many PRs that we really need help with Reviews so I will not make nobody waste time on this one.

PR #5357, opened 2026-01-06
✨ (helm/v2-alpha): Improve generated values.yaml comment conventions (single # vs ##)

camilamacedo86:
Update values.yaml generation to use ## for section headers and documentation blocks, following Helm chart best practices.

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/values.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/values.yaml
@@ -6,7 +6,8 @@
 ##
 # fullnameOverride: ""
 
-# Configure the controller manager deployment
+## Configure the controller manager deployment
+##
 manager:
   replicas: 1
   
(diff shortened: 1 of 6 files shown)


docs/book/src/cronjob-tutorial/testdata/project/dist/chart/values.yaml
```diff
@@ -6,7 +6,8 @@
 ##
 # fullnameOverride: ""
 
-# Configure the controller manager deployment
+## Configure the controller manager deployment
+##
```
vitorfloriano, line 10: Hi @camilamacedo86 could you explain the reasoning behind this change? I was trying to find something about using double hashtags in the official docs but couldn't.

For what I was able to learn this is mostly a formatting for readme-generation tools used in the industry. Right?

camilamacedo86: Thanks for asking!

This isn’t a Helm/YAML “requirement” (so you won’t find it mandated in the Helm docs), it’s mainly a usability convention for `values.yaml` files.

In `values.yaml` we have two different kinds of commented lines:

1) **Commented-out values users may want to enable**  
These use a *single* `#`, so if someone removes that `#` (manually or via an editor “uncomment selection” shortcut) the line becomes valid YAML, e.g.:

- `# fullnameOverride: ""`

2) **Pure documentation / section headers / prose**  
These should **never** become YAML keys/values. If they were `# Arguments`, and someone bulk-uncomments a block, that line would become `Arguments` (invalid YAML) and could break `helm install -f values.yaml` since values files are plain YAML.

So we use `##` for those docs/headers (and keep separator lines as `##/## ...`) so that a “remove one leading #” operation still leaves them commented (`# ...`) and the file stays valid. This is exactly what the change is doing in this PR (e.g. `# Arguments` → `## Arguments`).

This convention is also pretty common in the ecosystem (for example, the prometheus-community chart uses lots of `##` doc lines and single-`#` commented examples).

camilamacedo86: I updated the title to better clarify.

vitorfloriano: Got it. Thanks!

PR #5358, opened 2026-01-06
✨  (helm/v2-alpha): improve _helper generator readability

### LINKED ISSUE #5356: [helm/v2-alpha] need for a more readable helper generator

Currently, the chart _helper.tpl file is generated from a golang preformatted string using go template and escaped go template. It can be hard to read, and to maintain.

I would like to propose the following format:

```go
func (f *HelmHelpers) SetTemplateDefaults() error {
    // ...
    f.TemplateBody = makeHelperTemplate()
    // ...
}

func makeHelperTemplate() string {
	escape := func(s string) string {
		return "{{`" + s + "`}}"
	}

	return strings.Join([]string{
		escape("{{/*"),
		"Chart name based on project name.",
		"Truncated to 63 characters for Kubernetes compatibility.",
		escape("*/}}"),
		escape("{{- define \"") + "{{ .ProjectName }}" + escape(".name\" -}}"),
		escape("{{- default .Chart.name .Values.nameOverride \"") + "{{ .ProjectName }}" + escape("\" | trunc 63 | trimSuffix \"-\""),
		escape("{{- end }}"),
		"",
       // ...
    }, "\n") + "\n" // EOL new line
}
```

The change makes the template being generated at runtime rather than being a constant at compile time. I can't see any relevant issues doing this.

We started a discussion about this in https://github.com/kubernetes-sigs/kubebuilder/pull/5294

I'll create a PR for this after the next release

### Extra Labels

_No response_

robinlioret:
Rework the way the _helper.tpl file is generated for helm chart. It makes it more readable and probably easier to maintain.

Closes #5356

### DIFF
--- a/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
+++ b/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
@@ -17,8 +17,8 @@ limitations under the License.
 package charttemplates
 
 import (
-	"fmt"
 	"path/filepath"
+	"strings"
 
 	"sigs.k8s.io/kubebuilder/v4/pkg/machinery"
 )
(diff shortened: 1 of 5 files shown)


pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
```diff
@@ -53,91 +53,85 @@ func (f *HelmHelpers) SetTemplateDefaults() error {
...
-	// Use project name as prefix (e.g., "project-v4-with-plugins")
-	// This creates templates like "project-v4-with-plugins.name" instead of generic "chart.name"
-	// preventing collisions when chart is used as a Helm dependency
-	prefix := f.ProjectName
+	escape := func(s string) string {
+		return "{{`" + s + "`}}"
+	}
 
-	return fmt.Sprintf(helmHelpersTemplate, prefix, prefix, prefix, prefix, prefix, prefix, prefix, prefix, prefix)
+	return strings.Join([]string{
```
camilamacedo86, line 61: /lgtm cancel
/approved cancel

Could you please add it via constant as it was before? 
If you see all templates in kubebuilder uses the same machinary 
we should ensure that all works at the same way

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
```diff
@@ -53,91 +53,85 @@ func (f *HelmHelpers) SetTemplateDefaults() error {
...
+		"Only includes name and instance for consistent selection.",
+		escape("*/}}"),
+		escape("{{- define \"") + f.ProjectName + escape(".selectorLabels\" -}}"),
+		escape("app.kubernetes.io/name: {{ include \"") + f.ProjectName + escape(".name\" . }}"),
+		escape("app.kubernetes.io/instance: {{ .Release.Name }}"),
+		escape("{{- end }}"),
+	}, "\n") + "\n" // Add EOF new line
 }
-
-const helmHelpersTemplate = `{{` + "`" + `{{/*
```
camilamacedo86, line 66: Could we please keep the template as we have with the const to use the lib?

camilamacedo86:
Hi @robinlioret,

I went ahead and applied your suggested changes in https://github.com/kubernetes-sigs/kubebuilder/pull/5359, without introducing a new mechanism in the templates — I tried to keep the existing standards/conventions intact.

I also added you as a co-author so you get proper credit. Thanks a lot for your help!

Hope this helps. I didn’t want to create extra work for you having to rebase and re-apply the changes.

I’m closing this PR in favor of https://github.com/kubernetes-sigs/kubebuilder/pull/5359, but please feel free to re-open if you think it’s needed.

Issue #5354, opened 2026-01-05
[helm/v2-alpha] plugin ignores namespace-scoped Roles when generating Helm chart


AlirezaPourchali:
### What broke? What's expected?

When using `kubebuilder edit --plugins=helm/v2-alpha`, the plugin generates the Helm chart but **only converts the ClusterRole** from `config/rbac/role.yaml`. Namespace-scoped Roles (used for leader election, cross-namespace permissions, etc.) are silently ignored and not included in the generated Helm chart.

This causes **runtime RBAC permission errors** when the operator is deployed via Helm, even though the same operator works fine when deployed with `make deploy` (Kustomize).

Expected behavior: The Helm plugin should generate templates for **all** RBAC resources in `config/rbac/role.yaml`, including:
- ClusterRole → `dist/chart/templates/rbac/manager-role.yaml` ✅ (works)
- Namespace-scoped Roles → `dist/chart/templates/rbac/<namespace>-role.yaml` ❌ (missing)
- Corresponding RoleBindings for each Role ❌ (missing)

### Reproducing this issue

I have these RBAC markers in my controller:

```go
//+kubebuilder:rbac:groups=apps,namespace=infrastructure,resources=deployments,verbs=get;list;watch;update;patch
//+kubebuilder:rbac:groups=coordination.k8s.io,namespace=production,resources=leases,verbs=get;list;watch;create;update;patch;delete
//+kubebuilder:rbac:groups="",namespace=production,resources=events,verbs=create;patch;update
```

The RBAC files generated via `make manifests` are correct and include both ClusterRole and namespace-scoped Roles:

**`config/rbac/role.yaml`:**
```yaml
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: manager-role
rules:
- apiGroups: ["example.com"]
  resources: ["myresources"]
  verbs: ["get", "list", "watch", "create", "update"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: manager-role
  namespace: infrastructure
rules:
- apiGroups: ["apps"]
  resources: ["deployments"]
  verbs: ["get", "list", "patch", "update", "watch"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: manager-role
  namespace: production
rules:
- apiGroups: ["coordination.k8s.io"]
  resources: ["leases"]
[...]

PR #5360, opened 2026-01-06
🐛 (helm/v2-alpha): Fix cross-namespace RBAC file naming and namespace handling

camilamacedo86:
## What This Fixes

This PR fixes how the Helm plugin (v2-alpha) handles namespaces when converting your Kustomize project to a Helm chart.

### Problems Before This Fix

❌ **Missing RBAC Resources**  
If you used namespace-scoped Roles (like for leader election or cross-namespace permissions), they were ignored. Only ClusterRole was converted. Your Helm-deployed operator would fail with permission errors.

❌ **Broken Resource Names**  
If your namespace was `user` and you had a resource called `users`, it would break:
- Input: `users`
- Broken output: `{{ .Release.Namespace }}s` 😱

❌ **Incomplete DNS Replacement**  
Only `.svc` DNS names worked. Other patterns like `.pod`, `.svc.cluster.local`, or `.endpoints` were left with hardcoded namespaces.

❌ **File Name Collisions**  
Multiple Roles with the same name in different namespaces would overwrite each other.

❌ **Hardcoded Annotations**  
Things like `cert-manager.io/inject-ca-from: my-namespace/cert` stayed hardcoded.

---

## What This Fixes

✅ **All RBAC Resources Converted**  
Both ClusterRole AND namespace-scoped Roles/RoleBindings are now converted to Helm templates.

✅ **Safe Namespace Replacement**  
Resource names like `users`, `deployments`, `pods` are never touched. Only actual namespace fields are replaced.

✅ **All Kubernetes DNS Patterns**  
Every DNS format works: `.svc`, `.svc.cluster.local`, `.pod`, `.endpoints`, with ports, with paths, everything.

✅ **Smart Cross-Namespace Handling**  
- Your manager namespace (e.g., `myapp-system`) → `{{ .Release.Namespace }}`
- Other namespaces (e.g., `infrastructure`, `production`) → kept as-is

✅ **No More File Collisions**  
Files get unique names: `leader-election-role.yaml` vs `manager-role-infrastructure.yaml`

✅ **Complete Templating**  
Annotations, DNS names, resource references - everything that should be templated is templated.

---

## How It Works (Simple Explanation)

When you run `kubebuilder edit --plugins=helm/v2-alpha`, the plugin reads your Kustomize output and converts it to Helm templates.

### The Smart Rules

1. **If it's YOUR namespace** → Replace with `{{ .Release.Namespace }}`
2. **If it's SOMEONE ELSE'S namespace** → Keep it exactly as-is
3. **If it's a RESOURCE NAME** → Never touch it!

### Example: Your Namespace

```yaml
# Kustomize (input)
namespace: myapp-system

# Helm Chart (output)
```

When someone installs your chart:
```bash
helm install my-release ./chart --namespace production
```
It becomes: `namespace: production` ✅

### Example: External Namespace

```yaml
# Kustomize (input)
namespace: infrastructure

# Helm Chart (output)
namespace: infrastructure  # Kept as-is!
```

This is for cross-namespace permissions that must use a specific namespace.

---

### Example 1: Leader Election Role

**What you write** (RBAC marker):
```go
//+kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;list;watch;create;update;patch;delete
```

**What Kustomize generates**:
```yaml
kind: Role
metadata:
  name: myapp-leader-election-role
  namespace: myapp-system
rules:
- apiGroups: ["coordination.k8s.io"]
  resources: ["leases"]
  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```

**What the Helm plugin generates**:
- **File**: `dist/chart/templates/rbac/leader-election-role.yaml`
- **Content**:
```yaml
kind: Role
metadata:
  name: {{ include "myapp.resourceName" (dict "suffix" "leader-election-role" "context" $) }}
  namespace: {{ .Release.Namespace }}  # ← Templated! Will use whatever namespace you install to
rules:
- apiGroups: ["coordination.k8s.io"]
  resources: ["leases"]               # ← NOT templated! Still "leases", not "lease{{ .Release.Namespace }}s"
  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```

**Why this is important**: Your operator can now be installed in ANY namespace:
```bash
helm install prod ./chart --namespace production        # Works!
helm install staging ./chart --namespace staging        # Works!
helm install dev ./chart --namespace development        # Works!
```

---

### Example 2: Cross-Namespace Permissions

**What you write** (RBAC marker with explicit namespace):
```go
//+kubebuilder:rbac:groups=apps,namespace=infrastructure,resources=deployments,verbs=get;list;watch
```

**What Kustomize generates**:
```yaml
kind: Role
metadata:
  name: myapp-manager-role
  namespace: infrastructure  # Explicit namespace!
rules:
- apiGroups: ["apps"]
  resources: ["deployments"]
  verbs: ["get", "list", "watch"]
```

**What the Helm plugin generates**:
- **File**: `dist/chart/templates/rbac/manager-role-infrastructure.yaml` (note the `-infrastructure` suffix!)
- **Content**:
```yaml
kind: Role
metadata:
  name: {{ include "myapp.resourceName" (dict "suffix" "manager-role" "context" $) }}
  namespace: infrastructure  # ← NOT templated! Kept as "infrastructure" because it's not your manager namespace
rules:
- apiGroups: ["apps"]
  resources: ["deployments"]  # ← NOT templated! Still "deployments"
  verbs: ["get", "list", "watch"]
```

**Why this is important**: You can have permissions in specific namespaces:
```bash
# Your operator installs to "myapp-prod" but needs to watch deployments in "infrastructure"
helm install myapp ./chart --namespace myapp-prod

# Result:
# - Role created in "infrastructure" namespace (as specified)
# - Your operator runs in "myapp-prod" namespace (from --namespace)
```

---

### Example 3: ConfigMap with Service URLs

**What Kustomize generates**:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: app-config
  namespace: myapp-system
data:
  # Internal service (same namespace)
  database-url: "postgresql://db.myapp-system.svc:5432/mydb"
  
  # Internal service with full DNS
  redis-url: "redis://cache.myapp-system.svc.cluster.local:6379"
  
  # External service (different namespace)
  monitoring-url: "http://prometheus.monitoring-system.svc:9090"
  
  # Certificate reference
  tls-cert: "myapp-system/tls-cert"
```

**What the Helm plugin generates**:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ include "myapp.resourceName" (dict "suffix" "app-config" "context" $) }}
  namespace: {{ .Release.Namespace }}  # ← Templated
data:
  # Internal services - namespace is templated
  database-url: "postgresql://db.{{ .Release.Namespace }}.svc:5432/mydb"
  redis-url: "redis://cache.{{ .Release.Namespace }}.svc.cluster.local:6379"
  
  # External service - namespace is PRESERVED
  monitoring-url: "http://prometheus.monitoring-system.svc:9090"
  
  # Reference - templated
  tls-cert: "{{ .Release.Namespace }}/tls-cert"
```

**Why this is important**: Your service URLs automatically work in any namespace:
```bash
helm install myapp ./chart --namespace staging

# Results in:
# database-url: "postgresql://db.staging.svc:5432/mydb"
# redis-url: "redis://cache.staging.svc.cluster.local:6379"
# monitoring-url: "http://prometheus.monitoring-system.svc:9090"  # Still points to monitoring-system!
```

---

### Example 4: The Substring Bug Fix

**Scenario**: Your namespace is `user` and you have a resource type called `users`.

**Before this fix** (BROKEN ❌):
```yaml
# Input
namespace: user
rules:
- resources:
  - users
  - users/finalizers

# Broken output (blind string replacement)
rules:
- resources:
  - {{ .Release.Namespace }}s        # BROKEN! 😱
  - {{ .Release.Namespace }}s/finalizers  # BROKEN! 😱
```

**After this fix** (CORRECT ✅):
```yaml
# Input
namespace: user
rules:
- resources:
  - users
  - users/finalizers

# Correct output (smart replacement)
rules:
- resources:
  - users              # ✅ Preserved!
  - users/finalizers   # ✅ Preserved!
```

**How it works**: The code uses smart patterns that only match actual namespaces:
- `namespace: user` → Has `namespace:` field, **replace it**
- `users` → Just a word, no namespace field, **keep it**

---

### ✅ Gets Templated (Manager Namespace → `{{ .Release.Namespace }}`)

| Context | Example Input | Example Output |
|---------|---------------|----------------|
| Namespace field | `namespace: myapp-system` | `namespace: {{ .Release.Namespace }}` |
| Service DNS | `api.myapp-system.svc` | `api.{{ .Release.Namespace }}.svc` |
| Full DNS | `api.myapp-system.svc.cluster.local` | `api.{{ .Release.Namespace }}.svc.cluster.local` |
| DNS with port | `api.myapp-system.svc:8080` | `api.{{ .Release.Namespace }}.svc:8080` |
| Pod DNS | `pod.myapp-system.pod.cluster.local` | `pod.{{ .Release.Namespace }}.pod.cluster.local` |
| Annotation refs | `myapp-system/my-cert` | `{{ .Release.Namespace }}/my-cert` |
| ConfigMap refs | `configMapRef: myapp-system/config` | `configMapRef: {{ .Release.Namespace }}/config` |

### ✅ Gets Preserved (External Namespaces)

| Context | Example | Why |
|---------|---------|-----|
| Cross-namespace Role | `namespace: infrastructure` | Explicit namespace for cross-namespace permissions |
| External service | `.monitoring-system.svc` | Service in another namespace |
| External reference | `external-namespace/resource` | Reference to resource in another namespace |

### ✅ Never Touched (Resource Names)

| Context | Example | Why |
|---------|---------|-----|
| Resource types | `users`, `deployments`, `pods` | These are Kubernetes resource types, not namespaces |
| Labels | `app: frontend` | Labels are not namespaces |
| Arbitrary strings | `my-application` | Just regular strings |

---

### RBAC Files Go to `rbac/` Directory

**All** RBAC resources are organized in `dist/chart/templates/rbac/`:

```
dist/chart/templates/rbac/
├── controller-manager.yaml           (ServiceAccount)
├── manager-role.yaml                 (ClusterRole)
├── manager-rolebinding.yaml          (ClusterRoleBinding)
├── leader-election-role.yaml         (Role in manager namespace)
├── leader-election-rolebinding.yaml  (RoleBinding in manager namespace)
├── manager-role-infrastructure.yaml  (Role in "infrastructure" namespace)
├── manager-rolebinding-infrastructure.yaml (RoleBinding in "infrastructure")
└── ...
```

**Note**: They're **never** in `extras/`, even custom ServiceAccounts.

### Filename Rules

| Resource Type | Namespace | Example Filename |
|---------------|-----------|------------------|
| ClusterRole | (none) | `manager-role.yaml` |
| Role | Manager namespace | `leader-election-role.yaml` |
| Role | Cross-namespace | `manager-role-infrastructure.yaml` |
| RoleBinding | Manager namespace | `leader-election-rolebinding.yaml` |
| RoleBinding | Cross-namespace | `manager-rolebinding-infrastructure.yaml` |

The `-<namespace>` suffix prevents file collisions when you have multiple Roles with the same name in different namespaces.

---

### Before This Fix

```bash
$ kubebuilder edit --plugins=helm/v2-alpha
$ ls dist/chart/templates/rbac/
manager-role.yaml              # ← Only ClusterRole
manager-rolebinding.yaml       # ← Only ClusterRoleBinding
# WHERE IS MY LEADER ELECTION ROLE?? 😱

$ helm install myapp dist/chart/ --namespace production
# Result: Deployment fails with "forbidden: User cannot get leases" 💥
```

### After This Fix

```bash
$ kubebuilder edit --plugins=helm/v2-alpha
$ ls dist/chart/templates/rbac/
controller-manager.yaml
manager-role.yaml
manager-rolebinding.yaml
leader-election-role.yaml              # ← NOW INCLUDED! ✅
leader-election-rolebinding.yaml       # ← NOW INCLUDED! ✅
manager-role-infrastructure.yaml       # ← Cross-namespace roles too! ✅
manager-rolebinding-infrastructure.yaml

$ helm install myapp dist/chart/ --namespace production
# Result: Everything works! ✅
```

---

### Why Regex Instead of YAML Parsing?

**Short answer**: The content has Helm templates mixed with YAML, which breaks YAML parsers.

**Long answer**: By the time we need to replace namespaces, the content looks like this:

```yaml
metadata:
  name: {{ include "chart.name" . }}    # ← Helm template (not valid YAML!)
  namespace: myapp-system                # ← We need to replace this
  {{- if .Values.certManager.enable }}  # ← Helm conditional (not valid YAML!)
  annotations:
    cert-manager: enabled
  {{- end }}
```

If we try to parse this as YAML:
```go
yaml.Unmarshal(yamlContent, &data)
// Error: "could not find expected ':'"
```

So we use **smart regex patterns** that only match what we want:

1. **Namespace fields**: `namespace: myapp-system` (only if it's on its own line)
2. **DNS names**: `.myapp-system.svc` (only if there are dots on both sides)
3. **References**: `myapp-system/cert` (only if followed by `/`)

### Safety Guarantees

Each pattern is designed to prevent false matches:

| Pattern | What It Matches | What It Skips |
|---------|----------------|---------------|
| `(?m)^(\s*)namespace:\s+<value>\s*$` | `namespace: myapp-system` | `message: "namespace: myapp-system"` (in string) |
| `\b<value>/` | `myapp-system/cert` | `https://example.com/myapp-system/docs` (URL) |
| `\.<value>\.` | `.myapp-system.svc` | `users` (no dots), `app=myapp-system` (no dots on both sides) |

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/5354

### DIFF
--- a/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/chart_writer.go
+++ b/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/chart_writer.go
@@ -31,15 +31,17 @@ import (
 
 // ChartWriter handles writing Helm chart template files
 type ChartWriter struct {
-	templater *HelmTemplater
-	outputDir string
+	templater        *HelmTemplater
+	outputDir        string
+	managerNamespace string
 }
 
 // NewChartWriter creates a new chart writer
-func NewChartWriter(templater *HelmTemplater, outputDir string) *ChartWriter {
+func NewChartWriter(templater *HelmTemplater, outputDir string, managerNamespace string) *ChartWriter {
 	return &ChartWriter{
-		templater: templater,
-		outputDir: outputDir,
+		templater:        templater,
+		outputDir:        outputDir,
+		managerNamespace: managerNamespace,
 	}
 }
 
@@ -157,6 +159,17 @@ func (w *ChartWriter) generateFileName(resource *unstructured.Unstructured, inde
 			fileName = resource.GetKind()
 			if fileName == "" {
 				fileName = "resource"
+			}
+		}
+
+		// For namespace-scoped RBAC (Role, RoleBinding), append namespace suffix
+		// only for cross-namespace resources to prevent filename collisions.
+		// Resources in the manager namespace don't need a suffix since they're unique.
+		kind := resource.GetKind()
+		namespace := resource.GetNamespace()
+		if namespace != "" && (kind == "Role" || kind == "RoleBinding") {
+			if namespace != w.managerNamespace {
+				fileName = fmt.Sprintf("%s-%s", fileName, namespace)
 			}
 		}
 
--- a/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
+++ b/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
@@ -48,15 +48,17 @@ const (
 
 // HelmTemplater handles converting YAML content to Helm templates
 type HelmTemplater struct {
-	detectedPrefix string
-	chartName      string
+	detectedPrefix   string
+	chartName        string
+	managerNamespace string
 }
 
 // NewHelmTemplater creates a new Helm templater
-func NewHelmTemplater(detectedPrefix, chartName string) *HelmTemplater {
+func NewHelmTemplater(detectedPrefix, chartName, managerNamespace string) *HelmTemplater {
 	return &HelmTemplater{
-		detectedPrefix: detectedPrefix,
-		chartName:      chartName,
+		detectedPrefix:   detectedPrefix,
+		chartName:        chartName,
+		managerNamespace: managerNamespace,
 	}
 }
 
@@ -119,16 +121,58 @@ func (t *HelmTemplater) substituteProjectNames(yamlContent string, _ *unstructur
 	return yamlContent
 }
 
-// substituteNamespace replaces hardcoded namespace references with Release.Namespace
+// substituteNamespace replaces manager namespace references with {{ .Release.Namespace }}
+// while preserving cross-namespace references (e.g., infrastructure, production).
+//
+// DESIGN RATIONALE:
+// We use regex-based replacement (not YAML parsing) because the content already contains
+// Helm templates from previous substitutions, which would break YAML parsing.
+//
+// SAFETY GUARANTEES:
+// 1. Namespace fields: Only replaces `namespace: <exact-value>` (line-anchored regex)
+// 2. DNS names: Only replaces `.<namespace>.` (dots on both sides prevent substring matches)
+// 3. References: Only replaces `<namespace>/` (word boundary prevents false matches)
+//
+// TESTED SCENARIOS:
+// - All standard K8s resource types (ConfigMap, Secret, Ingress, etc.)
+// - All monitoring resources (ServiceMonitor, PodMonitor)
+// - All RBAC resources (Role, RoleBinding, with cross-namespace support)
+// - All DNS patterns (.svc, .svc.cluster.local, .pod, .endpoints)
+// - Custom CRDs with any structure
+// - Cross-namespace preservation (infrastructure, production, etc.)
+// - Substring bug prevention (namespace "user" doesn't break resource "users")
 func (t *HelmTemplater) substituteNamespace(yamlContent string, resource *unstructured.Unstructured) string {
-	hardcodedNamespace := t.detectedPrefix + "-system"
+	managerNamespace := t.managerNamespace
 	namespaceTemplate := "{{ .Release.Namespace }}"
 
-	// Replace hardcoded namespace references everywhere, including in the Namespace resource
-	// so that metadata.name becomes the Helm release namespace.
-	yamlContent = strings.ReplaceAll(yamlContent, hardcodedNamespace, namespaceTemplate)
+	// 1. NAMESPACE FIELDS: Replace `namespace: <manager-namespace>`
+	//    Pattern: Line-anchored to prevent false matches
+	//    Example: `namespace: project-system` → `namespace: {{ .Release.Namespace }}`
+	namespaceFieldPattern := regexp.MustCompile(`(?m)^(\s*)namespace:\s+` + regexp.QuoteMeta(managerNamespace) + `\s*$`)
+	yamlContent = namespaceFieldPattern.ReplaceAllString(yamlContent, "${1}namespace: "+namespaceTemplate)
 
-	// Replace service DNS name placeholders in certificates
+	// 2. RESOURCE REFERENCES: Replace `<manager-namespace>/resource-name`
+	//    Pattern: Word boundary ensures we don't match partial words
+	//    Example: `cert-manager.io/inject-ca-from: project-system/cert` → `{{ .Release.Namespace }}/cert`
+	//    Example: `configMapRef: project-system/config` → `{{ .Release.Namespace }}/config`
+	refPattern := regexp.MustCompile(`\b` + regexp.QuoteMeta(managerNamespace) + `/`)
+	yamlContent = refPattern.ReplaceAllString(yamlContent, namespaceTemplate+"/")
+
+	// 3. DNS NAMES: Replace `.<manager-namespace>.` in Kubernetes DNS patterns
+	//    Pattern: Dots on both sides ensure we only match DNS, not arbitrary strings
+	//    Handles ALL K8s DNS patterns: .svc, .svc.cluster.local, .pod, .endpoints, etc.
+	//    Example: `service.project-system.svc` → `service.{{ .Release.Namespace }}.svc`
+	//    Example: `pod.project-system.pod.cluster.local` → `pod.{{ .Release.Namespace }}.pod.cluster.local`
+	//
+	//    SAFETY: This won't match:
+	//    - Resource names: "users" (no dots around it)
+	//    - Arbitrary strings: "my-application" (no dots)
+	//    - Labels: "app=project-system" (no dots on both sides)
+	dnsPattern := regexp.MustCompile(`\.` + regexp.QuoteMeta(managerNamespace) + `\.`)
+	yamlContent = dnsPattern.ReplaceAllString(yamlContent, "."+namespaceTemplate+".")
+
+	// 4. CERTIFICATE-SPECIFIC: Additional service name templating for cert-manager
+	//    This is additive only and doesn't interfere with the above replacements
 	if resource.GetKind() == kindCertificate {
 		yamlContent = t.substituteCertificateDNSNames(yamlContent, resource)
 	}
--- a/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater_test.go
+++ b/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater_test.go
@@ -34,8 +34,9 @@ var _ = Describe("HelmTemplater", func() {
 
 	BeforeEach(func() {
 		templater = &HelmTemplater{
-			detectedPrefix: "test-project",
-			chartName:      "test-project",
+			detectedPrefix:   "test-project",
+			chartName:        "test-project",
+			managerNamespace: "test-project-system",
 		}
 	})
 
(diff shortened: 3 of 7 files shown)


camilamacedo86:
@AlirezaPourchali

Could you help us in the review?

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/chart_writer.go
```diff
@@ -160,6 +160,14 @@ func (w *ChartWriter) generateFileName(resource *unstructured.Unstructured, inde
 			}
 		}
 
+		// For namespace-scoped resources (Role, RoleBinding), append namespace to filename
+		// to avoid collisions when multiple resources have the same name in different namespaces
+		kind := resource.GetKind()
+		namespace := resource.GetNamespace()
+		if namespace != "" && (kind == "Role" || kind == "RoleBinding") {
+			fileName = fmt.Sprintf("%s-%s", fileName, namespace)
+		}
```
camilamacedo86, line 174: /hold
Requires changes

AlirezaPourchali: The filename collision fix is good.
The logic for cross-namespace RBAC filenames makes sense

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
```diff
@@ -119,16 +121,17 @@ func (t *HelmTemplater) substituteProjectNames(yamlContent string, _ *unstructur
...
+	managerNamespace := t.managerNamespace
 	namespaceTemplate := "{{ .Release.Namespace }}"
 
-	// Replace hardcoded namespace references everywhere, including in the Namespace resource
-	// so that metadata.name becomes the Helm release namespace.
-	yamlContent = strings.ReplaceAll(yamlContent, hardcodedNamespace, namespaceTemplate)
+	// Replace only the manager namespace with template variable.
+	// Cross-namespace values (infrastructure, production, etc.) are naturally preserved
+	// because they don't match the manager namespace string.
+	yamlContent = strings.ReplaceAll(yamlContent, managerNamespace, namespaceTemplate)
```
AlirezaPourchali: This string replacement is too aggressive and will break resource names that contain the namespace as a substring.

**Example bug:**
- CRD resource name: `users`  
- Manager namespace: `user`
- Result: `{{ .Release.Namespace }}s` ❌ (should stay `users`)

**Suggested fix:**
Replace only in the namespace field context:
```go
yamlContent = strings.ReplaceAll(yamlContent, "namespace: "+managerNamespace, "namespace: "+namespaceTemplate)
```
or even a better solution:
Parse YAML → Modify specific fields → Re-serialize
This way we know what we are changing.

camilamacedo86: All that is under the manager namespace is the namespace we define when we install Helm: `{{ .Release.Namespace }}`.

If we follow your suggestion as-is, we’ll end up with a namespace bug: the namespace would be set to `<project-name>-system`, because that value is coming from kustomize in this case.

So, in the same way we replace all occurrences of `<project-name>-system` with `{{ .Release.Namespace }}`, this will continue to work correctly.

We just need a caveat around this, due to the ability to customize the namespace via options in the default `config/kustomize` configuration.

However, if the ns is not == the manager's one, then we keep it.

AlirezaPourchali: I see your point, but i think you misunderstood me.

The replacement via `strings.ReplaceAll` works well on everything expect on the rbac yaml files for my example.
My CRD name is `users` and my manager deployment file which is `config/manager/manager.yaml` has the namespace field `namespace: user`. The namespace is a substring of the CRD name and when the helm chart is being created it replaces every `user` word with `{{ .Release.Namespace }}`.
This replacement will break the RBAC files including the CRD name inside it (`resources` field), it will create the following rbac template file:
```
kind: ClusterRole
metadata:
    name: {{ include "osiris.resourceName" (dict "suffix" "manager-role" "context" $) }}
rules:
    - apiGroups:
        - ""
      resources:
        - configmaps
      verbs:
        - get
        - list
        - patch
        - update
        - watch
    - apiGroups:
        - apps
      resources:
        - deployments
      verbs:
        - get
        - list
        - watch
    - apiGroups:
        - identity.me.cloud
      resources:
        - {{ .Release.Namespace }}s
      verbs:
        - create
        - delete
        - get
        - list
        - patch
        - update
        - watch
    - apiGroups:
        - identity.me.cloud
      resources:
        - {{ .Release.Namespace }}s/finalizers
      verbs:
        - update
    - apiGroups:
        - identity.me.cloud
      resources:
        - {{ .Release.Namespace }}s/status
      verbs:
        - get
        - patch
        - update

```
`{{ .Release.Namespace }}s` is the issue here.
the correct generated file would be:
```
kind: ClusterRole
metadata:
    name: {{ include "osiris.resourceName" (dict "suffix" "manager-role" "context" $) }}
rules:
    - apiGroups:
        - ""
      resources:
        - configmaps
      verbs:
        - get
        - list
        - patch
        - update
        - watch
    - apiGroups:
        - apps
      resources:
        - deployments
      verbs:
        - get
        - list
        - watch
    - apiGroups:
        - identity.me.cloud
      resources:
        - users
      verbs:
        - create
        - delete
        - get
        - list
        - patch
        - update
        - watch
    - apiGroups:
        - identity.me.cloud
      resources:
        - users/finalizers
      verbs:
        - update
    - apiGroups:
        - identity.me.cloud
      resources:
        - users/status
      verbs:
        - get
        - patch
        - update

```

AlirezaPourchali: i think treating the yaml files as a structured object would be our solution here, in that way we wouldn't blindly replace the substring inside the yaml files.

camilamacedo86: The Role that has ns == `"<project-name>-system"` in the install YAML must be applied in the same namespace where the manager is installed.

If you added a namespace via the RBAC markers, it won’t be replaced with "{{ .Release.Namespace }}", because that substitution is tied to `"<project-name>-system"`. So, unless you literally set the marker namespace to "<project-name>-system" (which wouldn’t make sense), Helm won’t swap it to "{{ .Release.Namespace }}".

Did you test this locally? 
If so, can you please share the exact steps to reproduce the issue that you have seen?

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater_test.go
```diff
@@ -751,6 +752,206 @@ metadata:
...
+			roleResource := &unstructured.Unstructured{}
+			roleResource.SetAPIVersion("rbac.authorization.k8s.io/v1")
+			roleResource.SetKind("Role")
+			roleResource.SetName("test-project-leader-election-role")
+
+			content := `apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: test-project-leader-election-role
+  namespace: production
```
camilamacedo86, line 810: @AlirezaPourchali you can check in the tests.
The ns will be the marker one.
We only replace when no ns is defined 
That means in install.yaml/kustomize we have <project-name>-system. 
In this case, any resource that in kustomize has the ns <project-name>-system should in the chart have ns  `{{ .Release.Namespace }} ` no matter the kind.

camilamacedo86: I think I understood now the scenario. 
I changed accordingly.
Thank you for that.

camilamacedo86:
Hi @AlirezaPourchali

Thank you for all your reviews.
I think I understood your scenario.
See that I addressed changes based on your comments.
The description is either updated. 

Could you please check it now?

camilamacedo86:
Hi @AlirezaPourchali,

I’m going to merge this one now because it addresses all your requests and the great findings from your review.

As an exception, I’m merging it because I opened https://github.com/kubernetes-sigs/kubebuilder/issues/5375 to track the release and ask the community to help validate these Helm changes (along with the other Helm PRs).

Please feel free to leave any additional comments — we can address them in a follow-up. If you can also test it and let us know if you still see any issues, that would help a lot.

This is currently a release blocker because we’re prioritizing getting all Helm changes in. It impacts end users, and ideally we can ask them to regenerate the Helm chart with `--force` only once.

Thanks again for the collaboration! 🙌

PR #5361, opened 2026-01-06
🐛 (helm/v2-alpha): Fix force option to overwriten all files less Chart.yaml

camilamacedo86:
Chart.yaml contains user-managed version metadata and should never be overwritten, even with --force. The --force flag now regenerates all other preserved files (values.yaml, _helpers.tpl, .helmignore, test-chart.yml) while respecting Chart.yaml.

Added integration tests to verify force flag behavior and Chart.yaml preservation.


camilamacedo86:
I am moving forward with this one to unblock the releae since this one is just a small fix for the force

Issue #5373, opened 2026-01-14
Solve flakes in the e2e tests


camilamacedo86:
We have been facing flakes like: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kubebuilder/5371/pull-kubebuilder-e2e-k8s-1-33-0/2011015766689714176

We need to improve the e2e tests in Helm to make them more robust and solve those.

PR #5376, opened 2026-01-18
🌱 (ci): make e2e tests more robust for prow to avoid flakes

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/5373


camilamacedo86:
/override pull-kubebuilder-e2e-k8s-1-35-0

It seems to reduce lets see in follow ups if now it is solved or not

PR #5377, opened 2026-01-18
📖 Recommend options of migrations

mloskot:
Since there are multiple (at least four) migration options, users may
appreciate guidance on which is the recommended one, especially for
straightforward cases.

----

Although the [How to contribute to docs](https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#how-to-contribute-to-docs) section refers to the Kubernetes docs style guide, I decided to follow [the pre-existing convention](https://github.com/kubernetes-sigs/kubebuilder/blob/66bfbc11338c8ead832f5583b5ade892dbb16134/docs/book/src/introduction.md?plain=1#L1-L2) presenting a tip note as it seems the Kubebuilder documenation does not use the [Hugo Shortcodes](https://kubernetes.io/docs/contribute/style/style-guide/#shortcodes) anywhere.

----

This PR stems from discussion [thread I opened on the Slack](https://kubernetes.slack.com/archives/CAR30FCJZ/p1768737780914779) where @camilamacedo86 kindly (and patiently) addressed my questions about the migrations guide.

### DIFF
--- a/docs/book/src/migrations.md
+++ b/docs/book/src/migrations.md
@@ -41,7 +41,11 @@ For further information on the project layout, see [What's in a basic project?][
 
 ## Migration Options
 
-### Automated Updates via GitHub Actions
+
+> [!TIP]
+> To reduce effort, we recommend enabling the [AutoUpdate Plugin][autoupdate-v1-alpha] (GitHub Actions). You can also run [alpha update](./reference/commands/alpha_update.md) locally—both use the same update logic. Use the other options mainly for older projects that do not have `cliVersion` in the `PROJECT` file as a one-time step to reach a supported version; after that, use these workflows for future updates (older versions cannot use these automation features).
+
+###  **(Recommended)** AutoUpdate/GitHub Action: Get Notified of New Kubebuilder Releases via Issues with a PR Link to Review and Upgrade
 
 The [AutoUpdate Plugin][autoupdate-v1-alpha] scaffolds an action that automatically monitors for new Kubebuilder releases and
 opens a GitHub Issue with a Pull Request compare link when updates are available. This is ideal for
@@ -65,7 +69,7 @@ kubebuilder edit --plugins="autoupdate/v1-alpha"
 
 See the [AutoUpdate Plugin documentation][autoupdate-v1-alpha] for complete details.
 
-### Using Alpha Update Locally
+### **(Recommended)** Use `alpha update` to Upgrade Without Losing Customisations (Logic Behind AutoUpdate/GitHub Action)
 
 If you prefer to run updates locally instead of relying on GitHub Actions, you can use the same logic
 as the [AutoUpdate Plugin][autoupdate-v1-alpha] directly from your command line.


mloskot:
image

so, I'm baffled about the PR Title Verifier failure that `Your PR title must start with one of the following indicators:`

docs/book/src/migrations.md
```diff
@@ -43,6 +43,9 @@ For further information on the project layout, see [What's in a basic project?][
 
 ### Automated Updates via GitHub Actions
 
+>[!Tip]
+>**Strongly recommended** migration option which is easy and helps avoid conflicts.
```
vitorfloriano: I wouldn't say it helps avoid conflicts. If the users is making a big jump between versions, there will probably be conflicts that will be committed along with the changes. The users will get a warning in the logs, tests will break and they'll need to checkout the branch, resolve the conflicts and push.

IMO, the main advantage of this plugin is that it creates an issue every time there's a new version out, and automates all the necessary steps for the upgrades. 

It's more about convenience and awareness, I think.

camilamacedo86: I think we cannot go with this word too. 
That helps be aware of, facilitate the process automate it. 
I will think in a way to improve the docs in order to clarify the questions : https://kubernetes.slack.com/archives/CAR30FCJZ/p1768737780914779 

Maybe a note with the info. 

Also, we do not have this `[!Tip]` class thing

mloskot: @vitorfloriano @camilamacedo86 I see. I have simplified the strong recommendation.

> Also, we do not have this `[!Tip]` class thing

I searched the docs for admonition styles in use and I found one here, and used it too:
https://github.com/kubernetes-sigs/kubebuilder/blob/66bfbc11338c8ead832f5583b5ade892dbb16134/docs/book/src/introduction.md?plain=1#L1-L2

camilamacedo86: Oh I see that works cool

docs/book/src/migrations.md
```diff
@@ -43,6 +43,9 @@ For further information on the project layout, see [What's in a basic project?][
```
camilamacedo86, line 43:
```
> [!TIP]
> To reduce effort, we recommend enabling the [AutoUpdate Plugin][autoupdate-v1-alpha] (GitHub Actions). You can also run `kubebuilder alpha update` locally—both use the same update logic. Use the other options mainly for older projects that do not have `cliVersion` in the `PROJECT` file as a one-time step to reach a supported version; after that, use these workflows for future updates (older versions cannot use these automation features)

```

camilamacedo86: WDYT about a more complete tip above?

docs/book/src/migrations.md
```diff
@@ -43,6 +43,9 @@ For further information on the project layout, see [What's in a basic project?][
 
 ### Automated Updates via GitHub Actions
 
+>[!Tip]
+>**Strongly recommended** migration option.
```
camilamacedo86:
```suggestion
```

docs/book/src/migrations.md
```diff
@@ -67,6 +70,9 @@ See the [AutoUpdate Plugin documentation][autoupdate-v1-alpha] for complete deta
 
 ### Using Alpha Update Locally
 
+>[!Tip]
+>Recommended migration option for local workflow.
+
```
camilamacedo86:
```suggestion
```

docs/book/src/migrations.md
```diff
@@ -43,6 +43,9 @@ For further information on the project layout, see [What's in a basic project?][
 
 ### Automated Updates via GitHub Actions
```
camilamacedo86:
```suggestion
###  **(Recommended)** AutoUpdate/GitHub Action: Get Notified of New Kubebuilder Releases via Issues with a PR Link to Review and Upgrade
```

docs/book/src/migrations.md
```diff
@@ -67,6 +70,9 @@ See the [AutoUpdate Plugin documentation][autoupdate-v1-alpha] for complete deta
 
 ### Using Alpha Update Locally
```
camilamacedo86:
```suggestion
### **(Recommended)** Use `alpha update` to Upgrade Without Losing Customisations (Logic Behind AutoUpdate/GitHub Action)
```

docs/book/src/migrations.md
mloskot: That is very good! Thanks @camilamacedo86 
Shall I commit it in GitHub (and perhaps then squash locally and force-push single commit) or are we waiting for what @vitorfloriano, others think?

camilamacedo86 (changes requested):
Hi @mloskot 


Thank you a lot to checking this one.
I did some suggestions see what do you think about?

I think that would fully address your questions for the others. WDYT?

docs/book/src/migrations.md
mloskot: LGTM

docs/book/src/migrations.md
mloskot: LGTM

docs/book/src/migrations.md
camilamacedo86: @mloskot 

I improved a little the text see;

To reduce effort, we recommend enabling the [AutoUpdate Plugin][autoupdate-v1-alpha] (GitHub Actions). You can also run `kubebuilder alpha update` locally—both use the same update logic. Use the other options mainly for older projects that do not have `cliVersion` in the `PROJECT` file as a one-time step to reach a supported version; after that, use these workflows for future updates (older versions cannot use these automation features)

camilamacedo86: Also i think we must to link alpha update docs here `kubebuilder alpha update`

mloskot: @camilamacedo86 Done both.

mloskot:
@camilamacedo86
> I did some suggestions see what do you think about?

I've committed them as they certainly capture my intention in more comprehensive way.

On Slack, my suggestion was much more simplistic, adding `recommended` tags to the headings:

- Automated Updates via GitHub Actions] (strongly recommended)
- Using Alpha Update Locally (recommended)

but then I wasn't sure if lengthy headers are a good idea, also changing URL anchors, etc. 

> I think that would fully address your questions for the others. WDYT?

Yes, it does indeed!

Thank you very much for completing my proposal!

camilamacedo86 (approved):
/lgtm

Amazing 🎉 

Thank you a lot for your help on this one

docs/book/src/migrations.md
```diff
@@ -65,7 +69,7 @@ kubebuilder edit --plugins="autoupdate/v1-alpha"
 
 See the [AutoUpdate Plugin documentation][autoupdate-v1-alpha] for complete details.
 
-### Using Alpha Update Locally
+### **(Recommended)** Use `alpha update` to Upgrade Without Losing Customisations (Logic Behind AutoUpdate/GitHub Action)
```
vitorfloriano, line 72: Nit:
```suggestion
### **(Recommended)** Use `alpha update` to Upgrade Without Losing Customizations (Logic Behind AutoUpdate/GitHub Action)
```
It's just a British to American English spelling difference.

docs/book/src/migrations.md
vitorfloriano, line 89: @camilamacedo86 A bit out of scope for this PR, but shouldn't the first version here also be `v4.6.0` instead of `v4.5.0`?

docs/book/src/migrations.md
vitorfloriano: Oopsie, too little too late! Already merged 😅

docs/book/src/migrations.md
mloskot: @camilamacedo86 by the way, in case you are not aware, the GitHub Markdown supports admonitions, see https://github.com/orgs/community/discussions/16925

docs/book/src/migrations.md
camilamacedo86: We could inform the cliVersion to alpha update
But alpha update uses alpha generate 
AND alpha generate had a bug in versions < 4.5.0 :-( that does not work well.

docs/book/src/migrations.md
camilamacedo86: I do not know if we should use American or Britsh :-) 
feel free to push a PR to do the change or not .
We probably should define that and validate the whole documents. 
Are we using Britsh or American in all something like

docs/book/src/migrations.md
vitorfloriano: > @camilamacedo86 by the way, in case you are not aware, the GitHub Markdown supports admonitions, see https://github.com/orgs/community/discussions/16925

Hi @mloskot @camilamacedo86 

Just to clarify, support for markdown admonitions was introduced to mdBook in `v0.5.0`.  I added [two of them](https://github.com/kubernetes-sigs/kubebuilder/pull/5207/changes#diff-69c89e6eb2b17c071cc4123e00198ef0affcc24c76ce4803bac93dc04a68d775R1-R5) when we migrated to that version.

I think we could improve the docs by making use of the available admonitions in other pages in the book as well. 

See: https://rust-lang.github.io/mdBook/format/markdown.html#admonitions

docs/book/src/migrations.md
mloskot: FYI, British is fine for me ;)

docs/book/src/migrations.md
mloskot: @vitorfloriano 
> I added [two of them](https://github.com/kubernetes-sigs/kubebuilder/pull/5207/changes#diff-69c89e6eb2b17c071cc4123e00198ef0affcc24c76ce4803bac93dc04a68d775R1-R5) when we migrated to that version.

Yup, that's where I learned that `[!tip]` can be used in Kubebuilder book :)

> I think we could improve the docs by making use of the available admonitions in other pages in the book as well.

A perfect 'good first issue' candidate indeed!

PR #5393, opened 2026-01-25
📖 (docs): Clarify PR title and commit message format guidelines

camilamacedo86:
Add clear distinction between PR titles (use emojis for release notes) and commit messages (use text prefixes). Include format examples and plugin scope guidelines to help contributors follow consistent conventions.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -132,23 +132,53 @@ make test              # CI aggregate (all of above + license)
 
 ## PR Submission
 
-### Title Format (MANDATORY)
-```
-:emoji: (optional/scope): User-facing description
-```
+### PR Title Format (MANDATORY)
+
+PR titles use **emojis** (appear in release notes).
+
+Format: `:emoji: [(plugin/version)]: Description`
+
+The `(plugin/version)` scope is optional; omit it for repo-wide or documentation-only changes.
 
 **Emojis:**
-- ⚠️ - Breaking change
-- ✨ - New feature
-- 🐛 - Bug fix
-- 📖 - Documentation
-- 🌱 - Infrastructure/tests/non-user-facing/refactor
+- ⚠️ (`:warning:`) - Breaking change
+- ✨ (`:sparkles:`) - New feature
+- 🐛 (`:bug:`) - Bug fix
+- 📖 (`:book:`) - Documentation
+- 🌱 (`:seedling:`) - Infrastructure/tests/refactor
 
 **Examples:**
 ```
-✨ (helm/v2-alpha): Add chart generation for cluster-scoped resources
-🐛: Fix project creation failure when GOBIN is unset
-📖: Update migration guide for Go 1.25 compatibility
+🐛 Resolve nil pointer panic in scaffold generator
+✨ (helm/v2-alpha): Add cluster-scoped resource support
+📖 (go/v4): Update deployment documentation
+✨ Update dependencies to latest versions
+```
+
+### Commit Message Format
+
+Commit messages follow the [Conventional Commits](https://www.conventionalcommits.org/) standard.
+
+Format: `<type>[optional scope]: <description>`
+
+The `[optional scope]` is typically the plugin/version (e.g., `helm/v2-alpha`, `go/v4`); omit it for repo-wide or non-plugin changes.
+
+**Types:**
+
+- **feat**: A new feature for the user or a plugin
+- **fix**: A bug fix for the user or a plugin
+- **docs**: Documentation changes only
+- **test**: Adding or updating tests
+- **refactor**: Code change that neither fixes a bug nor adds a feature
+- **chore**: Changes to build process, dependencies, or maintenance tasks
+- **breaking**: A breaking change (can be combined with other types)
+
+**Examples:**
+```
+fix: Resolve nil pointer panic in scaffold generator
+feat(helm/v2-alpha): Add cluster-scoped resource support
+docs(go/v4): Update deployment documentation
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -132,24 +132,62 @@ make test              # CI aggregate (all of above + license)
...
 **Examples:**
 ```
 ✨ (helm/v2-alpha): Add chart generation for cluster-scoped resources
-🐛: Fix project creation failure when GOBIN is unset
+🐛 (go/v4): Fix project creation failure when GOBIN is unset
 📖: Update migration guide for Go 1.25 compatibility
+🌱 (kustomize/v2): Refactor manifest generation logic
+```
+
+### Commit Message Format
```
vitorfloriano, line 158: I think we could stick to the Conventional Commits standard and use parentheses for scope only. 

For example:

```git
fix(docs): add missing link
```

We could also point contributors to the official specs.

See: https://www.conventionalcommits.org/en/v1.0.0/

camilamacedo86: Need to be convetional + (plugin/version) if is ref a specific plugin. 
I will update it. Thx

camilamacedo86: See now, wdyt?

vitorfloriano: The format is still not in the standard conventional commit. 


```diff
-(fix)(go/v4): fix issue xyz
+fix(go/v4): fix issue xyz
```
Note that only the scope is within parentheses.

camilamacedo86:
All done !!
Thank you for the review @vitorfloriano 
Let me know if you think that we can move forward now

CONTRIBUTING.md
```diff
@@ -149,34 +149,43 @@ See:
...
+📖 (go/v4): Update deployment documentation
+✨ Update dependencies to latest versions
+🌱 Add new GitHub action to test out doc samples
+```
+
+### Commit Message Format
 
-If the PR is "plugin" scoped, you may also append the responding plugin names in the prefix.
-[For instance](https://github.com/kubernetes-sigs/kubebuilder/commit/0b36d0c4021bbf52f29d5a990157466761ec180c):
+Commit messages use **text prefixes** (not emojis) in a format inspired by [Conventional Commits](https://www.conventionalcommits.org/).
```
vitorfloriano: In AGENTS.md we say that commit messages follow Conventional Commits but here we say they are "inspired by" Conventional Commits. 
```suggestion
Commit messages follow the [Conventional Commits](https://www.conventionalcommits.org/) standard.
```

CONTRIBUTING.md
```diff
@@ -149,34 +149,43 @@ See:
...
 
+Format: `(type) [(plugin/version)]: Description`
```
vitorfloriano: The Conventional Commits page supply the structure for commits. 

Perhaps we could use that as well.
```suggestion
Format: <type>[optional scope]: <description>
```

CONTRIBUTING.md
```diff
@@ -149,34 +149,43 @@ See:
...
+- 👻 (`:ghost:`) - No release note (unreleased changes only)
 
-You can also use the equivalent emoji directly, since GitHub doesn't
-render the `:xyz:` aliases in PR titles.
+**Examples:**
+```
+🐛 Resolve nil pointer panic in scaffold generator
+✨ (helm/v2-alpha): Add cluster-scoped resource support
+📖 (go/v4): Update deployment documentation
+✨ Update dependencies to latest versions
```
vitorfloriano, line 169: This is more like a chore than a feature.
```suggestion
🌱 Update dependencies to latest versions
```

AGENTS.md
```diff
@@ -132,23 +132,45 @@ make test              # CI aggregate (all of above + license)
...
-- 🌱 - Infrastructure/tests/non-user-facing/refactor
+### Commit Message Format
+
+Commit messages follow the [Conventional Commits](https://www.conventionalcommits.org/) standard.
+
+Format: `type(scope): description`
+
+The `(scope)` is optional; omit it for repo-wide or non-plugin changes.
+
+**Types:** `fix`, `feat`, `chore`, `docs`, `test`, `refactor`, `breaking`
```
vitorfloriano: Could we add that to CONTRIBUTING.md as well? We could clarify when to use each, similar to what the Angular docs do. See: https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type

Also, the Conventional Commits specs say that we should add a exclamation mark (`!`) before the colon (`:`) for breaking changes. For example: `chore(deps)!: Upgrade controller-runtime to 0.24.0`.

We MAY add a `BREAKING CHANGE:` to the footer of the commit description when commiting a breaking change, but it's optional.

CONTRIBUTING.md
camilamacedo86: Update of dependencies if is for the end users is not 🌱 .
That can be ✨ or ⚠️ ( example controller runtime upgrade )
🌱 is for changes that should not end up in the release notes ( i.e e2e tests, ci, things like )

AGENTS.md
camilamacedo86: We can go with the first one.
For the second, it feels a bit more complicated — what do you think?

Honestly, this is more of a guideline than a hard requirement. If the PR is squashed, we can adjust the PR title for the release notes before merging. I don’t think we should refuse contributions just because they didn’t follow this perfectly.

vitorfloriano: Sure. We don't have any checkers in CI so we can be more flexible.

PR #5395, opened 2026-01-25
📖 Add webhook bootstrap problem guide and warning

### LINKED ISSUE #1672: Webhook configuration to allow namespaceSelector
On the documentation about Webhook's on core types (like pods) https://book.kubebuilder.io/cronjob-tutorial/running-webhook.html t is mentioned that we should use namespaceSelector.

However if I am not wrong it's not part of the API according to this  https://book.kubebuilder.io/reference/markers/webhook.html

I am aware that we can always add it on the yaml afterwards but still it would be nice to have it in the markers as well

/kind feature

camilamacedo86:
## Problem

When creating webhooks for core Kubernetes types (Pod, Deployment, Job, etc.), the webhook can block its own controller Pod from starting, causing a deployment deadlock.

**Example:** A webhook that validates Pods will intercept its own controller Pod creation. The Pod can't start until the webhook validates it, but the webhook server isn't ready yet because it's IN that Pod. **Result: Deadlock.**

This is documented in the [CronJob tutorial](https://book.kubebuilder.io/cronjob-tutorial/running-webhook.html) as requiring manual YAML edits to add `namespaceSelector` or `objectSelector`.

Fixes #1672

## Solution

- Show warning when scaffolding core type webhooks
- Add comprehensive guide with manual steps to prevent bootstrap deadlock
- Update tutorial to reference new documentation

## Why not automate the scaffolds

Because Kustomize has limitations around patches, it’s hard for us to create a scaffold that’s both flexible and easy to automate across all use-case scenarios.

### DIFF
--- a/docs/book/src/reference/webhook-bootstrap-problem.md
+++ b/docs/book/src/reference/webhook-bootstrap-problem.md
@@ -1,0 +1,194 @@
+# Webhook Bootstrap Problem
+
+## The Problem
+
+When you create a webhook for a **core Kubernetes type** (Pod, Deployment, Job, etc.), the webhook can block its own controller Pod from starting, causing a deployment deadlock.
+
+**Example command:**
+```bash
+kubebuilder create webhook --group core --version v1 --kind Pod --programmatic-validation
+```
+
+**Example scenario:**
+
+1. You create a validating webhook for Pods
+2. You deploy your controller (which runs in a Pod)
+3. Kubernetes tries to create your controller Pod
+4. Your webhook intercepts this Pod creation
+5. The webhook server isn't ready yet (it's inside the Pod being created)
+6. The Pod creation hangs waiting for webhook validation
+7. The webhook never starts because the Pod is blocked
+
+**Result:** Deadlock. Your deployment fails.
+
+## When Does This Occur?
+
+### Core Kubernetes Types
+
+The bootstrap problem occurs when creating webhooks for built-in Kubernetes resources:
+
+- `core` group: Pod, Service, Namespace, ConfigMap, Secret
+- `apps` group: Deployment, StatefulSet, DaemonSet, ReplicaSet
+- `batch` group: Job, CronJob
+- Other built-in types
+
+**Why?** Your webhook validates the same type of resource that your controller deployment creates (typically Pods or Deployments).
+
+### Custom CRDs
+
+The bootstrap problem **does not occur** with custom resource webhooks:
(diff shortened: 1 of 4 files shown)


camilamacedo86:
@YiannisGkoufas @akirillov @ashishraman96 @migueleliasweb @j-bolivar-lt 

Could you please give a hand on the review of this one?

docs/book/src/reference/webhook-bootstrap-problem.md
```diff
@@ -0,0 +1,194 @@
...
+## The Problem
+
+When you create a webhook for a **core Kubernetes type** (Pod, Deployment, Job, etc.), the webhook can block its own controller Pod from starting, causing a deployment deadlock.
+
+**Example command:**
+```bash
+kubebuilder create webhook --group core --version v1 --kind Pod --programmatic-validation
+```
+
+**Example scenario:**
```
migueleliasweb, line 12: I've been bitten by this so many times at this point 😬 . Thanks for documenting this!

camilamacedo86: So, you are the right person to help us in the review.
WDYT? Can we get this one merged?
Do we have your LGTM @migueleliasweb ?

migueleliasweb: Definitely LGTM! 🚀 

As a future improvement, would it be possible to enable the namespace selector to be specified directly on the marker?

Something like:

```golang
// +kubebuilder:webhook:path=/foo,mutating=false,namespaces=namespace1,namespace2
```

Or something similar.

camilamacedo86: I think that need to be raised for controller-gen right? Or have we this option available there already?

migueleliasweb: Hummmm let me check. I think this has been raised already 🤔 .

migueleliasweb: https://github.com/kubernetes-sigs/controller-tools/issues/129 😞 . Yeah, maybe we need to resurrect this one 💀 .

camilamacedo86: I will move forward with the doc. 
But please if you see how we could improve open an issue for kb or controller-gen or both.
Your help and input is very welcome !!!

Issue #5330, opened 2026-01-02
Explore adding logging best-practices + logcheck to the Kubebuilder default scaffold or optional plugin for k8s conventions linter


camilamacedo86:
Kubebuilder is used to build Kubernetes-native controllers, but logging guidance can be confusing for newcomers. Kubernetes logging conventions say log messages should start with a capital letter and not end with a period, while Go error strings should not be capitalised. This can lead to inconsistent logs in scaffolded projects.

Can we explore if Kubebuilder should:
1) Document the recommended pattern clearly (log message style vs error string style),
2) Ensure the default scaffold follows it,
3) Consider enabling or recommending the `logcheck` linter.

Links:
- K8s logging message style: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md#message-style-guidelines
- Go error strings: https://go.dev/wiki/CodeReviewComments#error-strings
- logcheck linter: https://github.com/kubernetes-sigs/logtools/tree/main/logcheck

mayuka-c:
Hi @camilamacedo86 👋
I would like to this up.
I will try to explore more and come up on what might be the best.
/assign

PR #5396, opened 2026-01-26
✨ (go/v4): Add custom k8s linter for logs to promote better observability

camilamacedo86:
## What this PR does

Adds the custom linter `sigs.k8s.io/logtools/logcheck` to validate logs in Kubebuilder scaffolds.

This linter checks for common problematic logging scenarios and fails CI with actionable errors. Example failure output:
https://github.com/kubernetes-sigs/kubebuilder/actions/runs/21430987832/job/61710036984?pr=5396

## Why we’re doing this

We checked with the `logcheck` maintainers and, while adoption is relatively low, the project appears stable and maintained:
https://kubernetes.slack.com/archives/C020CCMUEAX/p1769592452395209

Given there have been no reported bugs since 2024, it seems like a good fit for improving the quality of scaffolded logging patterns in Kubebuilder.

## Links

Fixes: https://github.com/kubernetes-sigs/kubebuilder/issues/5330

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller.go
@@ -414,7 +414,7 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); client.IgnoreNotFound(err) != nil {
 				log.Error(err, "unable to delete old failed job", "job", job)
 			} else {
-				log.V(0).Info("deleted old failed job", "job", job)
+				log.V(1).Info("deleted old failed job", "job", job)
 			}
 		}
 	}
@@ -441,7 +441,7 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
 				log.Error(err, "unable to delete old successful job", "job", job)
 			} else {
-				log.V(0).Info("deleted old successful job", "job", job)
+				log.V(1).Info("deleted old successful job", "job", job)
 			}
 		}
 	}
--- a/docs/book/src/multiversion-tutorial/testdata/project/internal/controller/cronjob_controller.go
+++ b/docs/book/src/multiversion-tutorial/testdata/project/internal/controller/cronjob_controller.go
@@ -414,7 +414,7 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); client.IgnoreNotFound(err) != nil {
 				log.Error(err, "unable to delete old failed job", "job", job)
 			} else {
-				log.V(0).Info("deleted old failed job", "job", job)
+				log.V(1).Info("deleted old failed job", "job", job)
 			}
 		}
 	}
@@ -441,7 +441,7 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
 				log.Error(err, "unable to delete old successful job", "job", job)
 			} else {
-				log.V(0).Info("deleted old successful job", "job", job)
+				log.V(1).Info("deleted old successful job", "job", job)
 			}
 		}
 	}
--- a/hack/docs/internal/cronjob-tutorial/controller_implementation.go
+++ b/hack/docs/internal/cronjob-tutorial/controller_implementation.go
@@ -403,7 +403,7 @@ const controllerReconcileLogic = `log := logf.FromContext(ctx)
 			if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); client.IgnoreNotFound(err) != nil {
 				log.Error(err, "unable to delete old failed job", "job", job)
 			} else {
-				log.V(0).Info("deleted old failed job", "job", job)
+				log.V(1).Info("deleted old failed job", "job", job)
 			}
 		}
 	}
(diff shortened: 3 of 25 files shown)


docs/book/src/multiversion-tutorial/testdata/project/internal/controller/cronjob_controller.go
```diff
@@ -441,7 +441,7 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
 				log.Error(err, "unable to delete old successful job", "job", job)
 			} else {
-				log.V(0).Info("deleted old successful job", "job", job)
+				log.V(1).Info("deleted old successful job", "job", job)
```
camilamacedo86, line 444: Fix issue found in the docs samples within

docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller.go
```diff
@@ -441,7 +441,7 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
 				log.Error(err, "unable to delete old successful job", "job", job)
 			} else {
-				log.V(0).Info("deleted old successful job", "job", job)
+				log.V(1).Info("deleted old successful job", "job", job)
```
camilamacedo86, line 444: Fix issue found in the docs samples within

hack/docs/internal/cronjob-tutorial/controller_implementation.go
```diff
@@ -430,7 +430,7 @@ const controllerReconcileLogic = `log := logf.FromContext(ctx)
 			if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
 				log.Error(err, "unable to delete old successful job", "job", job)
 			} else {
-				log.V(0).Info("deleted old successful job", "job", job)
+				log.V(1).Info("deleted old successful job", "job", job)
```
camilamacedo86, line 433: Fix issue found in the docs samples within

pkg/plugins/golang/v4/scaffolds/init.go
```diff
@@ -117,10 +117,10 @@ func (s *initScaffolder) Scaffold() error {
...
 		if err != nil {
 			if errors.Is(err, afero.ErrFileNotFound) {
-				log.Warn("Unable to find boilerplate file. "+
-					"This file is used to generate the license header in the project. "+
-					"Note that controller-gen will also use this. Therefore, ensure that you "+
-					"add the license file or configure your project accordingly.",
+				log.Warn("Unable to find boilerplate file "+
+					"used to generate the license header in the project "+
+					"Note that controller-gen will also use this Therefore ensure that you "+
+					"add the license file or configure your project accordingly",
```
mloskot: This seems like a dup change from PR #5406 
See also my https://github.com/kubernetes-sigs/kubebuilder/pull/5406#discussion_r2735915593

camilamacedo86: Thank you .. it should not be part of the changes in this PR.
Out of scope.

camilamacedo86:
@vitorfloriano any reason for we not LGTM this one? 
WDYT?

vitorfloriano (approved):
I was a bit unsure at first because we use `log/slog` in Kubebuilder itself, but turns out the scaffolds use the logger from `controller-runtime,` which use `logr` and `klog`, so the linter works for the scaffolds.

\lgtm

camilamacedo86:
> I was a bit unsure at first because we use log/slog in Kubebuilder itself, but turns out the scaffolds use the logger from controller-runtime, which use logr and klog, so the linter works for the scaffolds.

We were able to reproduce the issue as well, and it shows up in our logs, so the current setup appears to be working correctly on our side.
That said, I’m going to raise a question for C+R: why aren’t we using klog here? Since Kubernetes uses klog, we often prefer to align with Kubernetes standards—so understanding the reasoning would help.

mloskot:
@camilamacedo86 
> why aren’t we using klog here? Since Kubernetes uses klog

Long ago [I asked similar question on the Slack channel #klog](https://kubernetes.slack.com/archives/CG3518SFJ/p1705777237944709) and I think you may find the answers I received there from @pohly (*) informative in the context here, quoting:

> In Kubernetes we are moving away from using "traditional" klog towards structured, contextual logging via the go-logr API.
> See https://github.com/kubernetes/component-base/tree/master/logs/example for an example that enables -v/-vmodule and JSON log output.
> (...) klog has lots of command line flags and function calls which are not used and not recommended by Kubernetes.

(*) I hope Patrick doesn't mind the mention, but I wanted to include persistent reference to source of that quote - I'm not 100% if the history of the Kubernetes Slack space persist.

camilamacedo86:
@mloskot I understand the the linter still valid right?
I will ask for its mantainers too but let me know if you see any reason for not

mloskot:
@camilamacedo86 I'd expect the linter to be still valid, but I don't know for sure as I did not dig the topic of future of klog in K8s any further.

vitorfloriano:
I believe `logr`+`klog` was preferred in Kubernetes to avoid breaking changes. It seems that by passing the logger by context (contextual logging), they could work around that constraint. That's what I understood from [KEP 3077](https://github.com/kubernetes/enhancements/tree/0463ac3359c31e054a6d10dcaac055eda8effb45/keps/sig-instrumentation/3077-contextual-logging). 

`log/slog` was not considered a viable alternative because it would represent a breaking change in the API of tools like controller-runtime, as per [this consideration in KEP 3077](https://github.com/kubernetes/enhancements/tree/0463ac3359c31e054a6d10dcaac055eda8effb45/keps/sig-instrumentation/3077-contextual-logging#use-logslog-instead-of-kloglogr):

> #### Use log/slog instead of klog+logr
> This isn't viable because slog doesn't provide a mechanism to pass a logger through a context. Therefore it would not be possible to support contextual logging in packages like client-go where adding an explicit logger parameter would be a major API break.

But that's just my reading of the whole thing.

mloskot:
@vitorfloriano  Thanks for those references. It looks like there was a plan, that had changed along the way. Here is the part that, I think, can be considered as an update to the quote I posted in https://github.com/kubernetes-sigs/kubebuilder/pull/5396#issuecomment-3817483020, in this section:

> [Clean separation of contextual logging and traditional klog logging](https://github.com/kubernetes/enhancements/tree/0463ac3359c31e054a6d10dcaac055eda8effb45/keps/sig-instrumentation/3077-contextual-logging#clean-separation-of-contextual-logging-and-traditional-klog-logging):
> 
> The initial revision of this KEP described a plan for moving all code for contextual logging into a `k8s.io/klogr` repository (...) However, that transition would have been complicated (...) Therefore the scope of the KEP was reduced from "remove dependency on klog" to "remove dependency on global logger in klog".

PR #5400, opened 2026-01-26
🐛 fix(api): move package cmd to internal

vitorfloriano:
This PR moves the `cmd` package to internal. 

The `cmd` package is not documented and should not be used by external users. 

See: https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.11.0/cmd

Also, we should note that the `cmd` package has only one function that is only imported by the main function in `main.go` in root. 

We decided not to simply move the logic in `Run()` to `main.go` to give us more flexibility when writing tests in the future.


camilamacedo86:
It seems fine.
I do not think will broke nobody that is using Kubebuilder as a lib

PR #5406, opened 2026-01-28
🐛 (go/v4): Follow Kubernetes logging conventions

camilamacedo86:
Updated all scaffolded log messages to follow [Kubernetes logging conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md#message-style-guidelines):
- Capital letter, no period, past tense for completed actions
- Added logging best practices to AGENTS.md with link to K8s conventions

Changes:
- Add section under good practices to clarify logging
- Add AI instructions for Kuebuilder and scaffolded projects
- Fix templates to generate logs accordingly with conventions

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/5330

PR #5409, opened 2026-01-29
🐛 fix(CLI): Fix logs to follow best practices

camilamacedo86:


pkg/model/resource/api.go
```diff
@@ -58,7 +58,7 @@ func (api *API) Update(other *API) error {
 		if api.CRDVersion == "" {
 			api.CRDVersion = other.CRDVersion
 		} else if api.CRDVersion != other.CRDVersion {
-			return fmt.Errorf("CRD versions do not match")
+			return fmt.Errorf("crd versions do not match")
```
vitorfloriano: Nit: Error strings should not be capitalized (unless beginning with proper nouns or acronyms, which is the case here).

See: https://go.dev/wiki/CodeReviewComments#error-strings


```suggestion
			return fmt.Errorf("CRD versions do not match")
```

pkg/model/resource/utils.go
```diff
@@ -30,7 +30,7 @@ func validateAPIVersion(version string) error {
 	case "v1":
 		return nil
 	default:
-		return fmt.Errorf("API version must be one of: v1beta1, v1")
+		return fmt.Errorf("api version must be one of: v1beta1, v1")
```
vitorfloriano:
```suggestion
		return fmt.Errorf("API version must be one of: v1beta1, v1")
```

pkg/plugins/common/kustomize/v2/scaffolds/api.go
```diff
@@ -94,7 +94,7 @@ func (s *apiScaffolder) Scaffold() error {
 		if err != nil {
 			hasCRUncommented, errCheck := pluginutil.HasFileContentWith(kustomizeFilePath, "- ../crd")
 			if !hasCRUncommented || errCheck != nil {
-				log.Error("Unable to find the target #- ../crd to uncomment in the file ",
+				log.Error("unable to find the target #- ../crd to uncomment in the file",
```
vitorfloriano, line 97: Ok, so I'm a bit confused here. 

Go conventions say we should not capitalize error messages, because they can be wrapped and yada yada. But Kubernetes conventions say we should capitalize logs. 

So, I suppose we should capitalize all the `log.Something` methods and **not** capitalize `fmt.Errorf` and similar Go stdlib methods?

mloskot: That's why I suggested in https://github.com/kubernetes-sigs/kubebuilder/pull/5406#pullrequestreview-3715827433 to include an explicit clarification that log messages != error messages

> Especially new users may confuse error strings with log messages and be wondering why Go says to not capitalise while Kubernetes says to capitalise :)

FWIW, an error message can also be logged, but then it is logged and becomes a log message including an error message ;)

camilamacedo86: For GO CLI all is lowercase
Only logs that we scaffold in the end user projects are upper cases because those are part of controllers and etc.

camilamacedo86: We covered it in the docs already, see; https://master.book.kubebuilder.io/reference/good-practices.html#understanding-go-vs-kubernetes-logging-conventions

camilamacedo86: @vitorfloriano 

See: https://github.com/search?q=repo%3Akubernetes%2Fkubernetes+ogger.Info%28%22&type=code

pkg/model/resource/resource.go
```diff
@@ -159,24 +159,24 @@ func (r Resource) Copy() Resource {
...
+		return fmt.Errorf("cannot update resource (Plural %q) with another with non-matching Plural %q",
 			r.Plural, other.Plural)
 	}
 
 	if other.Path != "" && r.Path != other.Path {
 		if r.Path == "" {
 			r.Path = other.Path
 		} else {
-			return fmt.Errorf("unable to update Resource (Path %q) with another with non-matching Path %q", r.Path, other.Path)
+			return fmt.Errorf("cannot update resource (Path %q) with another with non-matching Path %q", r.Path, other.Path)
```
vitorfloriano, line 179: "cannot" because it's not allowed or because it failed?

pkg/plugins/common/kustomize/v2/scaffolds/api.go
vitorfloriano, line 97: Ok, so in this particular log (`kusmotize/v2/scaffolds/api.go`), it should be capitalized, right?
```suggestion
				log.Error("Failed to find the target #- ../crd to uncomment in the file",
```

pkg/plugins/common/kustomize/v2/scaffolds/webhook.go
```diff
@@ -230,8 +230,8 @@ func uncommentCodeForConversionWebhooks(r resource.Resource) {
 			`configurations:
 - kustomizeconfig.yaml`)
 		if !hasWebHookUncommented || errCheck != nil {
-			log.Warn("Unable to find the target configurations with kustomizeconfig.yaml "+
-				"to uncomment in the file. ConverstionWebhooks requires this configuration "+
+			log.Warn("unable to find the target configurations with kustomizeconfig.yaml "+
```
vitorfloriano, line 233: Again, this is for the scaffolds, right?
```suggestion
			log.Warn("Failed to find the target configurations with kustomizeconfig.yaml "+
```

pkg/plugins/golang/v4/scaffolds/init.go
vitorfloriano, line 1: Ok so I'm still a bit unsure. 

There are still a couple of lower case logs that I believe should be capitalized, since they're meant for the scaffolds. I didn't mark all of them, for the sake of brevity. 

Please, correct me if I got this wrong. 

Also, there a couple of `unable` left that I believed could be reworded for clarity. Even `cannot` feels a bit ambiguous.

pkg/plugins/common/kustomize/v2/scaffolds/api.go
camilamacedo86: No, because it is a log from the CLI
Just the logs that will output things on the clusters should be capitalized
Kubebuilder  CLI (lower/Golang) -> Generates the project that is an Operator (a solution that extend k8s API/CRDs) -> The controllers , webhooks and all code inside of the generated project should be capitalized in order to follow k8s convetions since those are from the "native k8s solution" that will be deployed on the clusters

I hope that clarifies better.

pkg/plugins/common/kustomize/v2/scaffolds/webhook.go
camilamacedo86: That will raise warnings from the tooling.
Those are not code inside of the projects that runs on k8s clusters

pkg/model/resource/resource.go
camilamacedo86: yep

pkg/plugins/common/kustomize/v2/scaffolds/api.go
vitorfloriano: Oh, I see now. Thanks! 

So, only remarks left are in regards to `unable` and `cannot`. I think it would be better if we used `not allowed` and `failed`.

pkg/plugins/golang/v4/scaffolds/init.go
camilamacedo86: IN this PR we are only touching those that should not be.
Those that are not logs from controllers/webhooks that will be outputed on k8s clusters when the project itself be running. Therefore, should not follow k8s conventions only Golang one.

pkg/plugins/golang/deploy-image/v1alpha1/scaffolds/api.go
```diff
@@ -87,14 +87,14 @@ func (s *apiScaffolder) Scaffold() error {
 	boilerplate, err := afero.ReadFile(s.fs.FS, boilerplatePath)
 	if err != nil {
 		if errors.Is(err, afero.ErrFileNotFound) {
-			log.Warn("Unable to find boilerplate file. "+
+			log.Warn("unable to find boilerplate file. "+
 				"This file is used to generate the license header in the project.\n"+
-				"Note that controller-gen will also use this. Therefore, ensure that you "+
-				"add the license file or configure your project accordingly.",
+				"Note that controller-gen will also use this. Ensure that you "+
+				"add the license file or configure your project accordingly",
```
mloskot, line 93: nit: Since it's a multi-sentence message, it looks like the last one lost the full stop
i.e. `s/accordingly/accordingly./`

camilamacedo86: I think we should not have . anyway right?

mloskot: I assumed, possibly wrongly, that multi-sentence messages a logical/implicit exception to the logging conventions.

PR #5413, opened 2026-01-30
🐛 (helm/v2-alpha): wrap webhook resources with .Values.webhook.enable conditional

camilamacedo86:
Webhook Service and Webhook Configurations (ValidatingWebhookConfiguration, MutatingWebhookConfiguration) are now only deployed when .Values.webhook.enable is true. Previously, these resources were always created regardless of the webhook.enable setting.


camilamacedo86:
It is a silly bug so I am moving forward.
We have many open prs that need help with reviews this one is not required.

PR #5424, opened 2026-02-02
🐛 CLI release version): Preserve clean version tags in GoReleaser builds

camilamacedo86:
When GoReleaser builds binaries, it may create artifacts (SBOMs, etc.) during the build process, causing vcs.modified=true. This was appending -dirty to tagged releases (e.g., v4.5.3-rc.1+dirty).

Now we:
- Keep clean tag versions for proper releases (v4.5.3-rc.1)
- Use (devel) for development/pseudo-version builds when dirty
- Maintain -dirty suffix for git commits

Reason start to see issues, see: https://github.com/kubernetes-sigs/kubebuilder/actions/runs/21560712441/job/62139051157?pr=5422

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -76,6 +76,12 @@ func resolveMainVersion(main debug.Module) string {
 	return develVersion
 }
 
+// isPseudoVersion reports whether a version is a pseudo-version
+// (e.g., v0.0.0-20191109021931-daa7c04131f5 or v1.2.4-0.20191109021931-daa7c04131f5)
+func isPseudoVersion(v string) bool {
+	return strings.Contains(v, "-0.")
+}
+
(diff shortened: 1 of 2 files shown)


camilamacedo86:
@vitorfloriano Could you please help us with this one? WDYT?

camilamacedo86:
@vitorfloriano I will move forward with this one to unblock other fixes but please give a look to ensure that all is fine.

camilamacedo86:
I discovery the issue
The problem was in gitignore 
We should ignore dist from the root because Goreleaser but not from testdata.

vitorfloriano:
> The problem was in gitignore

@camilamacedo86 Yeah, I added `**/dist` instead of `/dist`. 😬 😅

camilamacedo86:
All fine :-)

PR #5422, opened 2026-02-01
🐛 (helm/v2-alpha): Fix chart generation and template rendering issues

camilamacedo86:
This commit resolves multiple issues discovered during Helm chart testing that prevented charts from rendering correctly in certain configurations:

- Fix template context in certificate DNS names: metrics certificates now use root context ($) instead of current context (.), preventing rendering failures in nested template scopes

- Fix service categorization logic: changed from substring matching to suffix matching to prevent false positives when chart names contain "webhook" or "metrics" keywords

- Fix ServiceMonitor YAML indentation: corrected tls.key field alignment to prevent YAML parsing errors

- Enable tracking of dist/chart files in testdata projects for better chart verification in version control

- Add integration tests to validate chart consistency across different naming configurations (Chart.yaml name vs project name, custom kustomize namePrefix)

These fixes ensure Helm charts render correctly regardless of project naming conventions and improve overall chart generation reliability.

Generate-by: Cursor

PR #5425, opened 2026-02-02
🌱 fix(ci): Use helm/v2-alpha plugin in test-helm-no-webhooks workflow

camilamacedo86:


camilamacedo86:
we will solve this one with https://github.com/kubernetes-sigs/kubebuilder/pull/5422

PR #5426, opened 2026-02-02
🐛 fix(alpha): respect custom project name in generate command

camilamacedo86:
## Problem

If a user creates a project with a custom name that differs from the directory name, that custom name is not used when re-generating the project via alpha generate or alpha update.

Here is visual example how I came across that issue

```bash
$ mkdir my-operator
$ cd my-operator
$ kubebuilder init --project-name zing-operator  # using Kubebuilder v4.10
$ kubebuilder create api ...                     # using Kubebuilder v4.10
...
$ kubebuilder alpha update                       # using Kubebuilder v4.11
```

and all `zing-operator` occurrences got renamed to `my-operator` after the directory name.

## Motivation

https://kubernetes.slack.com/archives/CAR30FCJZ/p1770038438744889

### DIFF
--- a/pkg/cli/alpha/internal/generate.go
+++ b/pkg/cli/alpha/internal/generate.go
@@ -436,6 +436,9 @@ func getInitArgs(s store.Store) []string {
 	if repo := s.Config().GetRepository(); repo != "" {
 		args = append(args, "--repo", repo)
 	}
+	if projectName := s.Config().GetProjectName(); projectName != "" {
+		args = append(args, "--project-name", projectName)
+	}
 	return args
 }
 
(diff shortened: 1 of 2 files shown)


mloskot:
Thank you for the fix.

I'd just refine the problem description slightly

> If an user has a custom name...

to read something like this

> If an user created project with custom name that is different from directory name...


Here is visual example how I came across that issue

```bash
$ mkdir my-operator
$ cd my-operator
$ kubebuilder init --project-name zing-operator  # using Kubebuilder v4.10
$ kubebuilder create api ...                     # using Kubebuilder v4.10
...
$ kubebuilder alpha update                       # using Kubebuilder v4.11
```

and all `zing-operator` occurrences got renamed to `my-operator` after the directory name.

mloskot:
Thank you @camilamacedo86 very much!

PR #5438, opened 2026-02-05
📖 docs: add note on limitations to CRD validation markers reference

orangecms:
Adapted from the [API design tutorial](https://book.kubebuilder.io/cronjob-tutorial/api-design.html?highlight=numbers).
I came here because I was puzzled by the [Cilium API using `int32` for `EndPort` in `PortRange`](http://github.com/cilium/cilium/tree/main/pkg/policy/api/l4.go), but internally using the expected `uint16` type, and I was hinted on Kubebuilder's limitations when asking in their Slack, and then got confused again because this was only mentioned in a tutorial, but not in the reference. So here we are!

 :rabbit:
 :hole:

### DIFF
--- a/docs/book/src/reference/markers/crd-validation.md
+++ b/docs/book/src/reference/markers/crd-validation.md
@@ -17,5 +17,17 @@ The grouping ensures clarity by showing how the same marker can be reused for di
 
 </aside>
 
+<aside class="note">
+<h1>Schema & Validation</h1>
+
+Custom resources are validated using the generated OpenAPI v3 schema and must comply with Kubernetes structural schema rules.
+Only field types and constraints that can be represented in the CRD schema are enforced by the API server.
+
+Numeric types are constrained by [Kubernetes CRD OpenAPI v3 schema compatibility](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation).
+In practice, prefer Go types that map cleanly to the supported OpenAPI formats (for example, `int32` and `int64` for integers).
+For values that require decimal-like representation, use `resource.Quantity`.
+
+Use Kubebuilder validation markers to declare additional constraints — such as minimum and maximum values, length limits, patterns, enums, and list or map rules — so they are reflected in the generated CRD and validated at runtime.
+</aside>
 
 {{#markerdocs CRD validation}}


docs/book/src/reference/markers/crd-validation.md
```diff
@@ -17,5 +17,12 @@ The grouping ensures clarity by showing how the same marker can be reused for di
 
 </aside>
 
+<aside class="note">
+<h2>Limitations</h2>
+
+Numeric values are limited for API compatibility.
+There are `int32` and `int64` for integers, and `resource.Quantity` for decimals.
+You can provide minimum and maximum values for further precision.
+</aside>
```
vitorfloriano, line 31: Hi @orangecms thank you for your contribution! It looks that combining `<aside class="note">` with `<h2>` does not go as well as combining it with `<h1>`, as per the preview:

Image

Can we try to change that to `<h1>` to see how it gets rendered?

Or, even better, we could use a markdown admonition to call attention to that. 


```suggestion
> [!IMPORTANT]
> Numeric values are limited for API compatibility. There are `int32` and `int64` for integers, and `resource.Quantity` for decimals. You can provide minimum and maximum values for further precision.
</aside>
```

orangecms: gotcha, I'll see about that second option; `<h1>` should generally only appear once in a document, see
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements#avoid_using_multiple_h1_elements_on_one_page

I would also propose to change the existing `<h1>` here, but that's for a different PR. :)

vitorfloriano: Noice! I personally do not fancy the style of the notes in the book overall. And agreed, sometimes there's too many `<h1>` in the page, which hinders accessibility and make it less semantic. 

If you have any ideas on how to improve the notes boxes in the book, feel free to open a PR. You could start here:

https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/theme/css/markers.css#L350-L368

docs/book/src/reference/markers/crd-validation.md
```diff
@@ -17,5 +17,9 @@ The grouping ensures clarity by showing how the same marker can be reused for di
 
 </aside>
 
+>[!IMPORTANT]
+> Numeric values are limited for API compatibility.
+> There are `int32` and `int64` for integers, and `resource.Quantity` for decimals.
+> You can provide minimum and maximum values for further precision.
```
camilamacedo86: Thank you for raise that.
I am not sure if I fully understand this one.

What do you mean with "Kubebuilder's limitations when asking in their Slack"?
The markers allows define the validations of the API / your own CRD.

Could you please help me out understand more?

orangecms: Yea. I have generated code based on Cilium's API specs, and noticed that I got int32 types for PortEnd. That is for TCP/UDP port numbers, which are uint16, really. So I was confused, joined their Slack and asked. And they told me that it's a limitation in Kubebuilder that uint16 is not available, but int32. And that is only stated in a tutorial, but not int the reference. See also the links I added in fhis PR's description.

orangecms: It's not relevant for the content of this PR though, just for context, so I only put that in the PR description.

camilamacedo86 (changes requested):
Hey @orangecms — I see what you’re trying to clarify here, but I don’t think this note is correct as written, and I’m going to block the PR until we adjust it.

This is **not** a Kubebuilder limitation. The “available numeric shapes” are primarily a consequence of **Kubernetes CRD / OpenAPI v3 schema constraints** (what Kubernetes accepts in the CRD schema), and Kubebuilder/controller-tools just generate that schema from types/markers.

In particular, saying “There are `int32` and `int64` for integers” is confusing in this context because CRD/OpenAPI uses `integer` (optionally with `format: int32/int64`) and `number` — and types like `uint16` aren’t a schema scalar. The correct way to express “port-like” constraints is via validation (`minimum`/`maximum`), which you already mention.

If we want to keep a note here, I think would need to be reworded with something like to attribute the constraint to Kubernetes CRD/OpenAPI schema, not Kubebuilder, e.g.:

> [!IMPORTANT]
> Numeric types and formats are constrained by [Kubernetes CRD OpenAPI v3 schema compatibility](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation).
> For additional precision or to model bounded ranges (e.g. ports), use validation such as `minimum` and `maximum`.
> Use `resource.Quantity` for values that require decimal-like representation.

orangecms:
Oh I see, thank you for digging this out!

I'll adjust it tomorrow. :)

orangecms:
Oh I see, thank you for digging this out!

I'll adjust it tomorrow. :)

docs/book/src/reference/markers/crd-validation.md
```diff
@@ -17,5 +17,10 @@ The grouping ensures clarity by showing how the same marker can be reused for di
 
 </aside>
 
+>[!IMPORTANT]
+> Numeric types are constrained by [Kubernetes CRD OpenAPI v3 schema compatibility](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation).
+> Thus you can only use `int32` and `int64` for integers.
+> Use `resource.Quantity` for values that require decimal-like representation.
+> Declare further bounds through minimum and maximum values.
```
camilamacedo86:
```suggestion
> You can use markers to validate the bounds such as for minimum and maximum values.
```

docs/book/src/reference/markers/crd-validation.md
```diff
@@ -17,5 +17,10 @@ The grouping ensures clarity by showing how the same marker can be reused for di
 
 </aside>
 
+>[!IMPORTANT]
```
camilamacedo86: Look the rendered page I think it would be better with a class Note
WDYT?

orangecms: That is where we came from, and I was suggested to go with this IMPORTANT thing instead:
https://github.com/kubernetes-sigs/kubebuilder/pull/5438#discussion_r2770633323

orangecms: I'd rather suggest changing the note above into a tip, since there shouldn't be multiple `<h1>` on a page, and there is already one for the main heading with the logo.

orangecms: Or into a !NOTE, since that exists. Just grepping real quick I see that this note-with-h1 thing is used in a lot of places though; are there guidelines to what to use when? :grimacing:

orangecms: Or did you mean changing !IMPORTANT into !NOTE?

orangecms: Or did you mean changing !IMPORTANT into !NOTE?

orangecms: Thought on the headings in aside boxes real quick:
Those should not be headings at all, as they do not start a chapter or section or otherwise structure the page, but rather put an emphasis on something. As such, `<em>` would be semantically adequate.
I'll track that in a new issue.

docs/book/src/reference/markers/crd-validation.md
```diff
@@ -17,5 +17,10 @@ The grouping ensures clarity by showing how the same marker can be reused for di
...
+> Numeric types are constrained by [Kubernetes CRD OpenAPI v3 schema compatibility](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation).
+> Thus you can only use `int32` and `int64` for integers.
+> Use `resource.Quantity` for values that require decimal-like representation.
+> You can use markers to validate the bounds such as for minimum and maximum values.
```
camilamacedo86:
```suggestion
<aside class="note">
<h1>Schema & Validation</h1>

Custom resources are validated using the generated OpenAPI v3 schema and must comply with Kubernetes structural schema rules.
Only field types and constraints that can be represented in the CRD schema are enforced by the API server.

Numeric types are constrained by [Kubernetes CRD OpenAPI v3 schema compatibility](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation).
In practice, prefer Go types that map cleanly to the supported OpenAPI formats (for example, `int32` and `int64` for integers).
For values that require decimal-like representation, use `resource.Quantity`.

Use Kubebuilder validation markers to declare additional constraints—such as minimum and maximum values, length limits, patterns, enums, and list or map rules—so they are reflected in the generated CRD and validated at runtime.
</aside>
```


I think we must to have something more generic as above
And use the Note class instead.

docs/book/src/reference/markers/crd-validation.md
camilamacedo86: Hi @orangecms I toke some time to think in the better wording.
See that I added an suggestion bellow
Let me know wdyt

camilamacedo86 (changes requested):

docs/book/src/reference/markers/crd-validation.md
camilamacedo86: @orangecms Could we do the change so that we can get this one merged?

Thank you a lot for your collab on it.

orangecms: I've taken your suggestion as-is, added spaces around the hyphens, and added you as a co-author. Thanks.

PR #5440, opened 2026-02-07
✨ (go/v4): add --multigroup flag to init command

camilamacedo86:
Users can now enable multigroup layout during project initialization instead of running a separate edit command afterward.

Before:
  kubebuilder init --domain example.org
  kubebuilder edit --multigroup

After:
  kubebuilder init --domain example.org --multigroup

The edit command still works for existing projects that need migration.


camilamacedo86:
This ensures the commands are consistent and homogeneous between init and edit. It’s a small change, and I’m moving forward with it to unblock the release.

PR #5450, opened 2026-02-08
🐛 fix(go/v4): Add more checks for webhooks in e2e tests to avoid flakes

camilamacedo86:


camilamacedo86:
Because it is a small fix and we have PRs that we really need help with the review and is hard to get help with those I am moving forward with this one. Therefore, the community can focus the effort were we most need.

PR #5455, opened 2026-02-08
🌱 docs/book: fix the install script

orangecms:
There is no (or no more) 32-bit x86 builds in the mdBook releases. See https://github.com/rust-lang/mdBook/releases/tag/v0.5.2 The aarch64 builds are now included, so use them.


vitorfloriano:
/lgtm 

Thanks!

PR #5458, opened 2026-02-08
✨ (go/v4): (Alpha) Add support to Server-Side Apply

### LINKED ISSUE #2514: Add Server Side Apply documentation to the book

Hi,

On the blogpost about Server Side Apply becoming GA, it references that when it's ready, the kubebuilder-book will be updated with examples. See https://kubernetes.io/blog/2021/08/06/server-side-apply-ga/#using-server-side-apply-in-a-controller

Is there an issue in this or in one of the related repositories that I could follow? I'm now trying to use it, and while receiving nice examples on slack, and finding all kinds of ways to implement it online, I'm still struggling a lot. A reference in the book would be really appreciated. So if there is not an issue, consider this as a feature request please ;-)

### Extra Labels

/kind documentation

camilamacedo86:
This PR adds a new flag to create api command to scaffolds APIs with controllers using Kubernetes Server-Side Apply (SSA) patterns.  How that works: See the doc added: https://deploy-preview-5458--kubebuilder.netlify.app/reference/server-side-apply.html also for the auot generated docs that we have now, see: https://deploy-preview-5458--kubebuilder.netlify.app/reference/markers/applyconfiguration.html                                        
                                                                                                  
### What is it?                                           

Server-Side Apply helps when multiple actors (your controller, users, other controllers) need to manage different fields of the same resource. Instead of overwriting the entire object, SSA tracks field ownership and only updates the fields you explicitly manage.                       
                                                            
### Usage                                                                                       
   
  ```bash                                                                                         
  # Initialize project                                      
  kubebuilder init --domain example.com
                                                                                                  
  # Create API with SSA plugin
  kubebuilder create api \                                                                        
    --group apps \                                                                                
    --version v1 \                                                                                
    --kind Application \                                                                          
    --ssa"                                                        
                                                                                                  
  # Generate apply configurations
  make generate                                                                                   
```
                                                            
## What it scaffolds                                                                               
                                                            
  1. API types with SSA markers - adds +genclient and +kubebuilder:ac:generate=true markers                                                                                  
  2. Makefile updates - adds applyconfiguration generation (first time only)
                                                                                                   
  After running make generate, you get type-safe apply configuration helpers:                     
  
```                                                          
  // Build desired state - only specify fields you want to manage                                 
  appApply := appsv1apply.Application(name, namespace).                                           
      WithSpec(appsv1apply.ApplicationSpec().                                                     
          WithReplicas(3))                                                                        
                                                                                                 
  // Apply - preserves fields managed by users/other controllers                                  
  r.Patch(ctx, app, client.Apply, client.ForceOwnership,                                          
      client.FieldOwner("my-controller"))                                                         
```
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/2514
Closes: https://github.com/kubernetes-sigs/kubebuilder/pull/5237

### CODE REVIEW

**Comment by camilamacedo86** [comment]
Hi @yyewolf and @wmuizelaar could you please give a hand on the review of this one?

File: docs/book/src/reference/server-side-apply.md
Comment (yyewolf) [inline docs/book/src/reference/server-side-apply.md:1]: Should we specify if there's an easy path to "enable" the plugin on existing APIs? And if there's not maybe simply add a note that suggests that a person willing to do that should regenerate their API and migrate manually?

Comment (camilamacedo86) [reply docs/book/src/reference/server-side-apply.md:1]: Offen in kubebuilder for we migrate something, we would need to re-scaffold. In this case, I think the best would be to re-scaffold using the plugin, or add the markers in the API and update the makefile.

File: docs/book/src/plugins/available/ssa-v1-alpha.md [hunk 1]
```diff
@@ -0,0 +1,332 @@
...
+        // Decide: retry, log, or force ownership
+    }
+}
+```
+
+## Additional resources
+
+For more details on Server-Side Apply concepts and patterns, see:
+
+- [Server-Side Apply Reference](../../reference/server-side-apply.md) - Concepts and theory
```
vitorfloriano: Broken link

docs/book/src/reference/server-side-apply.md
```diff
@@ -0,0 +1,332 @@
...
+kubebuilder create api --group apps --version v1 --kind App \
+  --plugins=ssa/v1-alpha  # OR deploy-image/v1-alpha
+```
+
+## Example in Testdata
+
+To see examples of SSA API scaffolds, check the testdata samples:
+
+**Single-group project:** [testdata/project-v4-with-plugins][testdata-single]
+- [`api/v1/application_types.go`][testdata-single-types] - API with `+genclient` marker
```
vitorfloriano: Broken link

docs/book/src/reference/server-side-apply.md
```diff
@@ -0,0 +1,332 @@
...
+- [`api/v1/groupversion_info.go`][testdata-single-gv] - Package with `+kubebuilder:ac:generate=true`
+- [`internal/controller/application_controller.go`][testdata-single-controller] - Minimal SSA controller template
```
vitorfloriano: Broken link

docs/book/src/reference/server-side-apply.md
```diff
@@ -0,0 +1,332 @@
...
+
+**Multi-group project:** [testdata/project-v4-multigroup][testdata-multi]
+- [`api/sea-creatures/v1/prawn_types.go`][testdata-multi-types] - API with `+genclient` marker
+- [`api/sea-creatures/v1/groupversion_info.go`][testdata-multi-gv] - Package with `+kubebuilder:ac:generate=true`
+- [`internal/controller/sea-creatures/prawn_controller.go`][testdata-multi-controller] - Minimal SSA controller template
```
vitorfloriano: Broken links

docs/book/src/plugins/available/ssa-v1-alpha.md
```diff
@@ -0,0 +1,332 @@
...
+
+**Multi-group project:** [testdata/project-v4-multigroup][testdata-multi]
+- [`api/sea-creatures/v1/prawn_types.go`][testdata-multi-types] - API with `+genclient` marker
+- [`api/sea-creatures/v1/groupversion_info.go`][testdata-multi-gv] - Package with `+kubebuilder:ac:generate=true`
+- [`internal/controller/sea-creatures/prawn_controller.go`][testdata-multi-controller] - Minimal SSA controller template
+
+## Memcached Operator Example with SSA
+
+For a complete example showing how to build a Memcached operator using Server-Side Apply,
+see the [Getting Started with SSA (`docs/book/src/getting-started-ssa/testdata/project`)][getting-started-ssa].
```
vitorfloriano: Broken link

vitorfloriano (commented):
Hi @camilamacedo86, on this first pass I found a few broken links in the docs.

pkg/plugins/golang/ssa/v1alpha1/plugin.go
vitorfloriano, line 1: Just a silly question: why should this plugin be placed in `pkg/plugins/golang` and not in `pkg/plugins/optional`?

camilamacedo86: That’s a good question, so I’ll explain how things are currently categorized.

Since it generates Golang code for the go/v4 layout, it falls under the Go category—similar to “Deploy Image One,” which is also under Go.

In the Optional category, we only include plugins that scaffold files which are not written in Golang.

vitorfloriano: Hmm...got it. I hadn't noticed that. 

But couldn't there be a way to structure this that would imply that this is also an optional plugin? Something like this:

```
pkg/
└── plugins/
    └── golang/
        └── optional/
            ├── deploy-image  # though I think moving deploy-image would be a breaking change
            └── server-side-apply 
```
Or:

```
pkg/
└── plugins/
    └── optional/
        ├── helm
        ├── grafana
        ├── autoupdate
        └── golang/
            ├── deploy-image  # breaking change
            └── server-side-apply
```

Digressing a little bit, another way to go about this would be to put the plugins that can be extended by other plugins all in common:

```
pkg/
└── plugins/
    ├── common/
    │   ├── golang
    │   └── kustomize
    └── optional/
        ├── helm
        ├── grafana
        ├── autoupdate
        └── golang/
            ├── deploy-image
            └── server-side-apply
```
But that would be a major breaking change. 

Anyway, just food for thought.

camilamacedo86: I could organize it in another way it is true.  
But it is as it is, we can only change when we do a 5x version if we wish to do so anyway.

docs/book/src/reference/server-side-apply.md
camilamacedo86: We can study a option to migrate manually and using AI, for example, but then it is out os scope of this one where we need to define how to introduce the feature.

docs/book/src/plugins/available/ssa-v1-alpha.md
camilamacedo86: Solved

docs/book/src/reference/server-side-apply.md
camilamacedo86: Solved

docs/book/src/reference/server-side-apply.md
camilamacedo86: Solved

docs/book/src/plugins/available/ssa-v1-alpha.md
camilamacedo86: Solved

camilamacedo86 (commented):
@vitorfloriano 

thank you for your review all addressed.

camilamacedo86:
@JoelSpeed @sbueringer 

Hey folks,

We are adding a plugin here to allow users to use SSA.
Users can opt in to use SSA per API.
Could you please help us out by checking if it has anything that is not done properly or could be addressed in another way using controller-tools and controller-runtime, regarding this one?

Thank you so much !!!

testdata/project-v4-multigroup/internal/controller/example.com/application_controller.go
```diff
@@ -0,0 +1,74 @@
...
+// +kubebuilder:rbac:groups=example.com.testproject.org,resources=applications,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=example.com.testproject.org,resources=applications/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=example.com.testproject.org,resources=applications/finalizers,verbs=update
+
+// Reconcile uses Server-Side Apply to manage resources for the Application.
+//
+// Server-Side Apply (SSA) provides declarative field ownership, allowing multiple actors
+// (users, controllers) to safely manage different fields of the same resource. This controller
+// only takes ownership of fields it explicitly declares, preserving user customizations to other fields.
+//
```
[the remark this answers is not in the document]
camilamacedo86: no

docs/book/src/plugins/available/ssa-v1-alpha.md
vitorfloriano, line 1: Rookie question: if SSA is a core feature in Kubernetes, shouldn't this be bundled with the default scaffold? 

Also, what about using a `--server-side` flag in `create api` instead of `create api --plugins ssa/v1-alpha`?

camilamacedo86: Yes we could add a flag.
My concerns to add a flag are:
- We need to update the Makefile as part of it so the plugin seems better for it
- If we add the flag we cannot change, it is more hard address breaking changes and I am not 100% yet of this is the best approach. 

If it get done: https://github.com/kubernetes-sigs/controller-tools/issues/1363 than I think have just the flag might be better indeed.

camilamacedo86: But that is a valid point I have the same doubt if it should not be a a flag

camilamacedo86:
/hold

https://github.com/kubernetes-sigs/kubebuilder/pull/5458/files#diff-58718ddb72dfe3cb3acf78882cedb1ff2053c3c62c982de8d5b73687bd1c46c1

We need to think if a flag would not be better

docs/book/src/reference/server-side-apply.md
yyewolf: Yeah that's totally fine and probably not too hard to rescaffold and manually "merge" the scaffolding

yyewolf: It does feel uniform to keep the plugin stance, it's not unnatural to have it that way

docs/book/src/plugins/available/ssa-v1-alpha.md
```diff
@@ -0,0 +1,330 @@
...
+    resourceApply := myv1apply.MyResource(req.Name, req.Namespace).
+        WithSpec(myv1apply.MyResourceSpec().
+            WithReplicas(3)).
+        WithLabels(map[string]string{
+            "managed-by": "my-controller",
+        })
+
+    // Apply - only manages the fields you specified above
+    // User's custom labels/annotations are preserved!
+    if err := r.Patch(ctx, resource, client.Apply, client.ForceOwnership,
```
yyewolf: Just for my own culture, is proning `client.ForceOwnership` a result of "a single resource should not be managed by multiple operator" logic ?

docs/book/src/reference/server-side-apply.md
camilamacedo86: /hold

We must consider propely both plugin and ssa flag

docs/book/src/plugins/available/ssa-v1-alpha.md
camilamacedo86: Updated. Good catcher !!!

docs/book/src/reference/server-side-apply.md
camilamacedo86: Changing for flag

docs/book/src/reference/server-side-apply.md
```diff
@@ -0,0 +1,373 @@
...
+
+## When to use it?
+
+Use this flag when:
+
+- **Multiple controllers manage the same resource**: Your controller manages some fields while other controllers or users manage others
+- **Users customize your CRs**: Users add their own labels, annotations, or spec fields that your controller shouldn't overwrite
+- **Partial field management**: You only want to manage specific fields and leave others alone
+- **Avoiding conflicts**: You want declarative field ownership tracking to prevent accidental overwrites
+
```
yyewolf: I'm not quite sure on how to phrase it, but should we also say "Expect other people to automate your CRs through higher level operators" ?

As in the operator that has generated SSA types would not be the one using them but merely providing them for other operators to import ?

yyewolf (approved):
That's a really nice improvement, I can't wait to try it and we (at my company) will have a lot of work to do to use this as effectively as possible 🔥

docs/book/src/reference/server-side-apply.md
```diff
@@ -0,0 +1,373 @@
...
+                "managed-by": "my-controller",
+            },
+        },
+        Spec: myv1.MyResourceSpec{
+            Replicas: 3,
+        },
+    }
+
+    // Apply using SSA - only manages the fields specified above
+    if err := r.Patch(ctx, resource, client.Apply,
```
Gentoli: Stumbled on this PR while searching for recommendations on SSA for kubebuilder controllers. looks like [client.Apply](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#Apply) is deprecated. Is this still the way to do SSA?

testdata/project-v4/Makefile
```diff
@@ -49,7 +49,7 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
-	"$(CONTROLLER_GEN)" object:headerFile="hack/boilerplate.go.txt",year=$(YEAR) paths="./..."
+	"$(CONTROLLER_GEN)" object:headerFile="hack/boilerplate.go.txt",year=$(YEAR) applyconfiguration:headerFile="hack/boilerplate.go.txt" paths="./..."
```
camilamacedo86: @JoelSpeed @alvaroaleman @sbueringer 
It would be great if we could not need to add at all here the `applyconfiguration` since it will only generate the resources when the ac marker is used and a project can have CRDs that are SSA and others that not.

The need to add this does not bring a nice UX. 
Could we not change the ssa marker / generator to work as any other marker in the controller-tools?

docs/book/src/reference/server-side-apply.md
```diff
@@ -0,0 +1,99 @@
...
+- Handle conflicts by checking `apierrors.IsConflict(err)` and requeueing
+- Only specify fields your controller manages using the builder pattern
+- Use `client.ForceOwnership` only when your controller must have final authority
+
+## Additional resources
+
+- [Kubernetes Server-Side Apply Documentation][server-side-apply]
+- [controller-gen CLI Reference](./controller-gen.md)
+
+[server-side-apply]: https://kubernetes.io/docs/reference/using-api/server-side-apply/
```
camilamacedo86, line 121: @JoelSpeed @alvaroaleman @sbueringer  

Could you please give a hand on the review of this doc?
Could you please let us know if has any info here that is not accurate or can/should be shaped?

pkg/plugins/golang/v4/scaffolds/api.go
```diff
@@ -151,3 +165,117 @@ func (s *apiScaffolder) Scaffold() error {
...
+		return
+	}
+	if hasMarker {
+		return
+	}
+
+	// Try multiple patterns to handle different Makefile formats
+	// 1. Try with boilerplate and YEAR variable (current default)
+	//nolint:lll
+	err = util.ReplaceInFile(makefilePath, makefileOldObjectGenWithBoilerplateAndYear, makefileNewObjectGenWithBoilerplateAndYear)
```
v47: This works, but the nested fallback structure is a little hard to scan. Since the intent is “try these known Makefile formats in order and succeed on the first match,” I think this could be made table-driven.

Something like:

```go

func replaceObjectGenInMakefile(makefilePath string) error {
	replacements := []struct {
		old string
		new string
	}{
		{
			old: makefileOldObjectGenWithBoilerplateAndYear,
			new: makefileNewObjectGenWithBoilerplateAndYear,
		},
		{
			old: makefileOldObjectGenWithBoilerplate,
			new: makefileNewObjectGenWithBoilerplate,
		},
		{
			old: makefileOldObjectGenNoBoilerplate,
			new: makefileNewObjectGenNoBoilerplate,
		},
	}
	for _, replacement := range replacements {
		if err := util.ReplaceInFile(makefilePath, replacement.old, replacement.new); err == nil {
			return nil
		}
	}
	return fmt.Errorf("none of the known controller-gen object generator patterns matched")
}
```
Then the caller can stay simple:
```
if err := replaceObjectGenInMakefile(makefilePath); err != nil {
	log.Warn("unable to find standard controller-gen object generator line in Makefile. "+
		"Add applyconfiguration generation manually",
		"error", err)
	return
}
```
This avoids the nested if err != nil chain and removes the need for `//nolint:lll`.
Small note, it assumes `ReplaceInFile` returning an error means the pattern did not match in this context. If it can also fail because of read or write errors, it may be worth distinguishing those cases separately.

test/e2e/internal/helpers/generate_v4.go
```diff
@@ -240,6 +240,46 @@ func GenerateV4WithCustomWebhookPath(kbc *utils.TestContext) {
...
+// GenerateV4WithSSA implements a go/v4 plugin project with Server-Side Apply enabled.
+func GenerateV4WithSSA(kbc *utils.TestContext) {
+	initingTheProject(kbc)
+
+	By("creating API with Server-Side Apply (--ssa)")
+	err := kbc.CreateAPI(
+		"--group", kbc.Group,
+		"--version", kbc.Version,
+		"--kind", kbc.Kind,
+		"--namespaced",
```
v47, line 252: Can SSA also be cluster-scoped?

v47:
Since SSA ownership is only claimed when the user actually calls the Apply with a field owner, maybe the flag description/docs should avoid saying the scaffolded controller uses SSA. It enables SSA support, but does not claim ownership until user code calls something like:

```go
r.Client.Status().Apply(ctx, desired, client.FieldOwner("foo-controller"))
```

pkg/plugins/golang/v4/scaffolds/api.go
```diff
@@ -151,3 +165,117 @@ func (s *apiScaffolder) Scaffold() error {
...
+}
+
+// updateMakefile modifies the existing controller-gen object generator line to also run
+// applyconfiguration generation. Only runs when the first SSA API is created.
+// On failure, logs a warning and does not stop scaffolding.
+func (s *apiScaffolder) updateMakefile() {
+	makefilePath := "Makefile"
+
+	// Skip if already updated
+	hasMarker, err := util.HasFileContentWith(makefilePath, makefileApplyConfigurationMarker)
```
mayuka-c: Just my thought

I was thinking would it be better than we return error from here on the reason why the updateMakefile failed and then callee [here](https://github.com/kubernetes-sigs/kubebuilder/pull/5458/changes#diff-2b45b3690775d205ba84e67e27f8fbda705efb65adc71c8c6c87ade18c7ffe80R117) will handle the error and mention explicitly with `log.Warn(err, "Failed to automatically update Makefile")`. I feel this would make it much cleaner based on the logs on what exactly the issue is.

Note: We can still not fail the scaffolding process and let it continue

pkg/plugins/golang/v4/scaffolds/api.go
```diff
@@ -151,3 +165,117 @@ func (s *apiScaffolder) Scaffold() error {
...
+		}
+	}
+	return true
+}
+
+// updateMakefile modifies the existing controller-gen object generator line to also run
+// applyconfiguration generation. Only runs when the first SSA API is created.
+// On failure, logs a warning and does not stop scaffolding.
+func (s *apiScaffolder) updateMakefile() {
+	makefilePath := "Makefile"
```
mayuka-c: Also I was wondering, do we want to add a troubeshooting/manual steps if automatic update of Makefile failed. Currently we are just logging with warning messages.

pkg/plugins/golang/v4/scaffolds/api.go
camilamacedo86: We should not return an error because that would means fail when the Makefile is customized and does not allow move forward if the end user change their Makefile.

If we be unable to update the Makefile we warn and then end users are aware of to update it manually

pkg/plugins/golang/v4/scaffolds/api.go
camilamacedo86: We warn and end users can manually update their project. 
It is too simple, has not really a reason for troubeshooting. 
What we can do is provide a better warn message that can be more helpful.

camilamacedo86: Done 🚀

docs/book/src/reference/server-side-apply.md
camilamacedo86: done

docs/book/src/reference/server-side-apply.md
camilamacedo86: fixed

test/e2e/internal/helpers/generate_v4.go
camilamacedo86: yes work with both.

v47: I think the test exists only for namespaced, but no test for non-nemaspaced.

mayuka-c (approved):
LGTM

PR #5442, opened 2026-02-07
✨ (helm/v2-alpha): Add Makefile targets for Helm Development

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/5351

PR #5468, opened 2026-02-12
✨ feat(helm/v2-alpha): Add install-helm target and upgrade to Helm 4

camilamacedo86:
Adds install-helm Makefile target to scaffolded projects and update GitHub actions to use the target

Follow up of: https://github.com/kubernetes-sigs/kubebuilder/pull/5442
We should either have a target to install helm as we install the others bins and requirements via Makefile.


camilamacedo86:
Due this one is small I am moving forward to unblock release.

Issue #5473, opened 2026-02-12
PROJECT file cannot represent multiple controllers for the same GVK (scaffolding/management ambiguity)


guilhem:
### What do you want to happen?

I’d like to be able to scaffold and track multiple controllers for the same API type (same Group/Version/Kind) in a Kubebuilder project, and have this represented in the PROJECT file in a supported way.

Concretely, I want something like:
- same GVK (e.g. apps.example.com/v1, Kind=Foo)
- two different reconcilers/controllers (e.g. FooController and FooBackupController, or “default” vs “specialized” controller filtered by predicates/labels)
- both registered in main.go and supported by re-scaffolding (kubebuilder alpha generate, kubebuilder edit, or other workflows that rely on PROJECT as source of truth)

### What is the problem?

Today the `PROJECT` file seems to assume a single controller per resource (per GVK). resources entries appear keyed by GVK, and there is no obvious supported field for distinguishing multiple controllers (e.g. controller name / controller ID) for the same GVK.

As a result:

- It’s unclear how to declare “multiple controllers for the same resource” in PROJECT without duplicating the same GVK entry.
- If you implement multiple controllers manually in code, the PROJECT file no longer fully reflects the project state, and re-scaffolding may not be able to preserve or manage these controllers.
- Controller naming must be unique at runtime (metrics/controller registration), so even if multiple controllers are added manually, the scaffolding story is unclear.

### Why this matters / use cases

Some real use cases where multiple controllers for the same GVK are useful:
- Split responsibilities (e.g. one controller manages status/conditions, another manages a subset of dependent resources)
- “Modes” of reconciliation (e.g. different behavior based on labels/annotations, each controller filtered with predicates)
Migration scenarios (run “old” and “new” controller side-by-side temporarily, each targeting a subset)
- Different queues / concurrency settings per sub-controller for the same Kind

camilamacedo86:
Thank you for raising this one. 
It seems a very interesting use case.

In this case, I think we would need to have:

- a) unique name per controller, to set in: https://github.com/kubernetes-sigs/kubebuilder/blob/f1d36f89f1a68c5fce4afb353bd70144ad816f6a/testdata/project-v4/internal/controller/admiral_controller.go#L61
- b) The ReconciliationName also would need to use the controllerName, like: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/cmd/main.go#L187-L193
[...]

PR #5474, opened 2026-02-13
:sparkles: add a controller-name for multi-controller

guilhem:
Naive implementation of #5473

This pull request introduces support for scaffolding multiple controllers for the same Group/Version/Kind (GVK) in Kubebuilder, allowing users to split reconciliation responsibilities, run different reconciliation modes, or manage migration scenarios. The main enhancement is the addition of the `--controller-name` flag to the `create api` command, along with corresponding updates to documentation, configuration, and code generation logic.

**Feature: Multiple Controllers for the Same GVK**
- Added support for the `--controller-name` flag in `create api`, enabling scaffolding of additional controllers for an existing API. Each controller gets its own file, struct, and unique registration, and the controller name is tracked in the PROJECT file. [[1]](diffhunk://#diff-c5829230aebd15fc2e06d9780edde61507b9bd146210fc4ab72861d90c74e371R163-R185) [[2]](diffhunk://#diff-0f8e7df2c2a72f7a5268b5368d4e10a3280644ebd925bb17b93d4b66b7f7305bR36-R56) [[3]](diffhunk://#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R275-R292) [[4]](diffhunk://#diff-308bcfa3b43879beda396725af2d0a7ad7098ce10543da7df9de8cde38dc7f17R72-R101) [[5]](diffhunk://#diff-308bcfa3b43879beda396725af2d0a7ad7098ce10543da7df9de8cde38dc7f17R196)
- Updated the resource model and configuration logic to allow multiple entries for the same GVK with different controller names, including proper handling in resource addition and updating. [[1]](diffhunk://#diff-5bff77782709554dc2f357ae882753d32fdf47bc6154bae87a18d20c5fb4aeb6R43-R47) [[2]](diffhunk://#diff-fa67f07438bee89278763ce38ab2be285585aced8093f51ccdabfe7a331e0ab0R240-R249) [[3]](diffhunk://#diff-fa67f07438bee89278763ce38ab2be285585aced8093f51ccdabfe7a331e0ab0R260-R270) [[4]](diffhunk://#diff-fa67f07438bee89278763ce38ab2be285585aced8093f51ccdabfe7a331e0ab0L267-R285) [[5]](diffhunk://#diff-5bff77782709554dc2f357ae882753d32fdf47bc6154bae87a18d20c5fb4aeb6R123-R151) [[6]](diffhunk://#diff-5bff77782709554dc2f357ae882753d32fdf47bc6154bae87a18d20c5fb4aeb6R227-R229)
- Improved code generation to use the controller name for file naming, struct naming (PascalCase conversion), and registration, and updated templates and replacers accordingly. [[1]](diffhunk://#diff-6998b85f7106294249797b98b5c86c0f3c68ce35021a19bef1714e5fe3631b62R36) [[2]](diffhunk://#diff-6998b85f7106294249797b98b5c86c0f3c68ce35021a19bef1714e5fe3631b62R52-R62) [[3]](diffhunk://#diff-5bff77782709554dc2f357ae882753d32fdf47bc6154bae87a18d20c5fb4aeb6R252) [[4]](diffhunk://#diff-5bff77782709554dc2f357ae882753d32fdf47bc6154bae87a18d20c5fb4aeb6R123-R151) [[5]](diffhunk://#diff-2a3c437dc2f2baf7c32e26a12c886f273e5df193b2c9d1dc7881a76f04922177R24)

**Validation and CLI Enhancements**
- Added validation for the controller name to ensure it conforms to DNS-1035 label requirements, and integrated the flag into the CLI and scaffolding logic. [[1]](diffhunk://#diff-91c570a618c9932bc98f6db8bb8da3a05ba7d3921dc0718b804bc5c4736ad79eR29) [[2]](diffhunk://#diff-91c570a618c9932bc98f6db8bb8da3a05ba7d3921dc0718b804bc5c4736ad79eR117-R120) [[3]](diffhunk://#diff-91c570a618c9932bc98f6db8bb8da3a05ba7d3921dc0718b804bc5c4736ad79eR166-R190)

**Documentation Updates**
- Expanded documentation to describe the new feature, including use cases, configuration examples, and instructions for tracking controller names in the PROJECT file. [[1]](diffhunk://#diff-c5829230aebd15fc2e06d9780edde61507b9bd146210fc4ab72861d90c74e371R163-R185) [[2]](diffhunk://#diff-0f8e7df2c2a72f7a5268b5368d4e10a3280644ebd925bb17b93d4b66b7f7305bR36-R56) [[3]](diffhunk://#diff-308bcfa3b43879beda396725af2d0a7ad7098ce10543da7df9de8cde38dc7f17R72-R101) [[4]](diffhunk://#diff-308bcfa3b43879beda396725af2d0a7ad7098ce10543da7df9de8cde38dc7f17R196) [[5]](diffhunk://#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R275-R292)

**Code Generation and Template Changes**
- Updated scaffolding logic and templates to support multiple controllers, including changes to `cmd/main.go` and controller test files. [[1]](diffhunk://#diff-2b45b3690775d205ba84e67e27f8fbda705efb65adc71c8c6c87ade18c7ffe80L111-R131) [[2]](diffhunk://#diff-2a3c437dc2f2baf7c32e26a12c886f273e5df193b2c9d1dc7881a76f04922177R24)

These changes make Kubebuilder more flexible for advanced controller patterns and improve clarity and maintainability in projects with complex reconciliation needs.

### DIFF
--- a/docs/book/src/reference/project-config.md
+++ b/docs/book/src/reference/project-config.md
@@ -41,7 +41,8 @@ resources:
   - api:
       crdVersion: v1
       namespaced: true
-    controller: true
+    controllers:
+      - name: memcached
     domain: testproject.org
     group: example.com
     kind: Memcached
@@ -53,13 +54,15 @@ resources:
   - api:
       crdVersion: v1
       namespaced: true
-    controller: true
+    controllers:
+      - name: busybox
     domain: testproject.org
     group: example.com
     kind: Busybox
     path: sigs.k8s.io/kubebuilder/testdata/project-v4-with-deploy-image/api/v1alpha1
     version: v1alpha1
-  - controller: true
+  - controllers:
+      - name: certificate
     domain: io
     external: true
     group: cert-manager
@@ -74,22 +77,16 @@ version: "3"
 
 Kubebuilder supports scaffolding multiple controllers for the same API type (same Group/Version/Kind). This is useful for splitting responsibilities across controllers, running different reconciliation modes, or migration scenarios.
 
-When using the `--controller-name` flag with `create api`, the controller name is tracked in the PROJECT file via the `controllerName` field. Each entry with a different `controllerName` is treated as a distinct resource entry:
+When using the `--controller-name` flag with `create api`, controller names are tracked in the PROJECT file via the `controllers` list under the resource:
 
 ```yaml
 resources:
   - api:
       crdVersion: v1
       namespaced: true
-    controller: true
-    controllerName: memcached-health
-    domain: example.com
-    group: cache
-    kind: Memcached
-    path: example.com/memcached-operator/api/v1alpha1
-    version: v1alpha1
-  - controller: true
-    controllerName: memcached-status
+    controllers:
+      - name: memcached-health
+      - name: memcached-status
     domain: example.com
     group: cache
     kind: Memcached
@@ -97,7 +94,7 @@ resources:
     version: v1alpha1
 ```
 
-In the example above, two controllers exist for the same `Memcached` Kind: `memcached-health` and `memcached-status`. The first entry also owns the API definition, while the second is controller-only (`--resource=false`).
+In the example above, a single `Memcached` resource tracks two controllers: `memcached-health` and `memcached-status`.
 
 ## Why do we need to store the plugins and data used?
 
@@ -147,7 +144,8 @@ resources:
   - api:
       crdVersion: v1
       namespaced: true
-    controller: true
+    controllers:
+      - name: memcached
     domain: testproject.org
     group: example.com
     kind: Memcached
@@ -159,13 +157,15 @@ resources:
   - api:
       crdVersion: v1
       namespaced: true
-    controller: true
+    controllers:
+      - name: busybox
     domain: testproject.org
     group: example.com
     kind: Busybox
     path: sigs.k8s.io/kubebuilder/testdata/project-v4-with-deploy-image/api/v1alpha1
     version: v1alpha1
-  - controller: true
+  - controllers:
+      - name: certificate
     domain: io
     external: true
     group: cert-manager
@@ -192,8 +192,8 @@ Now let's check its layout fields definition:
 | `resources.api`                     | The API scaffolded in the project via the sub-command `create api`.                                                                                                                                                                                                             |
 | `resources.api.crdVersion`          | The Kubernetes API version (`apiVersion`) used to do the scaffolding for the CRD resource.                                                                                                                                                                                      |
 | `resources.api.namespaced`          | The API RBAC permissions which can be namespaced or cluster scoped.                                                                                                                                                                                                             |
-| `resources.controller`              | Indicates whether a controller was scaffolded for the API.                                                                                                                                                                                                                      |
-| `resources.controllerName`          | **(Optional)** A custom name for the controller, provided via the `--controller-name` flag. When set, it allows multiple controllers for the same GVK. The name is used for the controller file name, the reconciler struct name, and the controller's `Named()` registration. Must be a DNS-1035 label (lowercase alphanumeric and hyphens). |
+| `resources.controller`              | **(Deprecated, compatibility only)** Legacy boolean kept for backward compatibility with older PROJECT files. New projects should rely on `resources.controllers`.                                                                                                             |
+| `resources.controllers`             | **(Optional)** A list of controllers scaffolded for the resource. Each entry contains a `name` provided via `--controller-name` (or the default lowercase kind). Names are used for controller file names, reconciler struct naming, and `Named()` registration. Must be DNS-1035 labels (lowercase alphanumeric and hyphens). |
 | `resources.domain`                  | The domain of the resource which was provided by the `--domain` flag when the project was initialized or via the flag `--external-api-domain` when it was used to scaffold controllers for an [External Type][external-type].                                                   |
 | `resources.group`                   | The GKV group of the resource which is provided by the `--group` flag when the sub-command `create api` is used.                                                                                                                                                                |
 | `resources.version`                 | The GKV version of the resource which is provided by the `--version` flag when the sub-command `create api` is used.                                                                                                                                                            |
--- a/pkg/config/v3/config.go
+++ b/pkg/config/v3/config.go
@@ -207,6 +207,7 @@ func (c Cfg) GetResource(gvk resource.GVK) (resource.Resource, error) {
 	for _, res := range c.Resources {
 		if gvk.IsEqualTo(res.GVK) {
 			r := res.Copy()
+			normalizeResourceControllers(&r)
 
 			// Plural is only stored if irregular, so if it is empty recover the regular form
 			if r.Plural == "" {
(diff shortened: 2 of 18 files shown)


docs/book/src/reference/project-config.md
```diff
@@ -69,6 +69,36 @@ resources:
...
+      crdVersion: v1
+      namespaced: true
+    controller: true
+    controllerName: memcached-health
+    domain: example.com
+    group: cache
+    kind: Memcached
+    path: example.com/memcached-operator/api/v1alpha1
+    version: v1alpha1
+  - controller: true
```
camilamacedo86: How it would be if instead of replicated GKV for each controller 
we have:
controller
name:
- nameA
- nameB
- nameC

I think that would be a more clean structure 
It would either facilitate the implementation for alpha generate.
WDYT?

docs/book/src/reference/project-config.md
```diff
@@ -163,6 +193,7 @@ Now let's check its layout fields definition:
...
 | `resources.api.namespaced`          | The API RBAC permissions which can be namespaced or cluster scoped.                                                                                                                                                                                                             |
 | `resources.controller`              | Indicates whether a controller was scaffolded for the API.                                                                                                                                                                                                                      |
+| `resources.controllerName`          | **(Optional)** A custom name for the controller, provided via the `--controller-name` flag. When set, it allows multiple controllers for the same GVK. The name is used for the controller file name, the reconciler struct name, and the controller's `Named()` registration. Must be a DNS-1035 label (lowercase alphanumeric and hyphens). |
```
camilamacedo86: Let's try check out it with a list of Namespaces

controller.names LIST
WDYT?

pkg/config/v3/config.go
```diff
@@ -264,7 +282,7 @@ func (c *Cfg) UpdateResource(res resource.Resource) error {
 	}
 
 	for i, r := range c.Resources {
-		if res.IsEqualTo(r.GVK) {
+		if res.IsEqualTo(r.GVK) && res.ControllerName == r.ControllerName {
```
camilamacedo86: I think we need a method to check in the LIST something like
resource.HasControllerName()

pkg/config/v3/config_test.go
```diff
@@ -609,6 +697,102 @@ version: "3"
...
+								Group:   "cache",
+								Version: "v1alpha1",
+								Kind:    "Memcached",
+							},
+							Controller:     true,
+							ControllerName: "memcached-health",
+							API:            &resource.API{CRDVersion: "v1", Namespaced: true},
+						},
+						{
+							GVK: resource.GVK{
```
camilamacedo86: We should not duplicate the GKV we could not have more than one on the cluster less yet in the project

pkg/plugins/golang/v4/api.go
```diff
@@ -154,12 +163,31 @@ func (p *createAPISubcommand) InjectResource(res *resource.Resource) error {
 		return errors.New("'--external-api-module' requires '--external-api-path' to be specified")
 	}
 
+	// Validate controller name if provided
+	if p.options.ControllerName != "" {
+		if !p.options.DoController {
+			return errors.New("cannot use '--controller-name' with '--controller=false'")
```
camilamacedo86, line 180: Valiations should be in the preScaffold 
We probably need to add this one inside of the method that validate GKV already

pkg/plugins/golang/v4/api.go
```diff
@@ -154,12 +163,31 @@ func (p *createAPISubcommand) InjectResource(res *resource.Resource) error {
...
+		}
+		if errs := validation.IsDNS1035Label(p.options.ControllerName); len(errs) != 0 {
+			return fmt.Errorf("invalid controller name %q: must be a DNS-1035 label "+
+				"(lowercase, alphanumeric, hyphens, starting with a letter)", p.options.ControllerName)
```
camilamacedo86: Same here.

pkg/plugins/golang/v4/api.go
```diff
@@ -154,12 +163,31 @@ func (p *createAPISubcommand) InjectResource(res *resource.Resource) error {
...
+	}
+
 	p.options.UpdateResource(p.resource, p.config)
 
 	if err := p.resource.Validate(); err != nil {
 		return fmt.Errorf("error validating resource: %w", err)
 	}
 
+	// When adding a named controller for an existing GVK, skip API re-scaffolding
+	if p.options.ControllerName != "" && p.options.DoAPI {
```
camilamacedo86: I think it is not accurate.
At the same way that we have many controllers for a core API we should be able to have many controllers for a external type too.

pkg/plugins/golang/options.go
```diff
@@ -68,6 +68,10 @@ type Options struct {
 	// Namespaced is true if the resource should be namespaced.
 	Namespaced bool
 
+	// ControllerName is an optional custom name for the controller.
+	// When set, allows scaffolding multiple controllers for the same GVK.
+	ControllerName string
```
camilamacedo86: Lets try to do a list 
But we need to properly think in how we will design the API here
I think we will need a new struct Controller and then inside of it we add name

Therefore, in the future we can add other specs for Controller if we have the need.

docs/book/src/reference/project-config.md
guilhem: That's a cleaner structure indeed :)

Is something like that ok?
```yaml
controllers:
- name: nameA
- name: nameB
- name: nameC
```

It gives the opportunity to add more fields in the future for each controllers.

guilhem: or, if we want to prevent an array:
```yaml
controllers:
  nameA: {}
  nameB: {}
  nameC: {}
```

camilamacedo86 (changes requested):
HI @guilhem 

Thank you for looking on that.
I did some comments.
Mainly IHMO:

- The API must be a design choice that allow us to grow if we need. So we will probably need to have a new struct Controller inside of Resources and that allows pass a List of names.
- The validations for GKV need to be done in the preScaffold. Those should be in the funcs that we have already today in the API
- We will need either ensure that alpha generate work with. So, if we have resources.Crontroller.names we should call create api resource false and controller true with the name for each name after create the API.

docs/book/src/reference/project-config.md
guilhem: I just fear when using the cli:
When the first name hasn't been set, we will have to set "nameA" to the first computed and nameB to the one set.
(not really a problem, just something to do)

pkg/plugins/golang/options.go
guilhem: I totally agree

docs/book/src/reference/project-config.md
camilamacedo86: > When the first name hasn't been set, we will have to set "nameA" to the first computed and nameB to the one set.

We have a method that do the "merge" of GKV. 
I think we need to have exactly the same but for the new Controller as well. 

We will need to think a lot here to ensure that we have a great API 
Because we cannot change easily those after we release the changes.

So maybe it will be acctually like

```
resource:
controller:
   - name: A
controller: 
    -Name B 
```    
The list I think should be of controllers.
ONE API/GKV can have MANY controllers ;-) 

So that tomorrow we can either

```
resource:
controller:
   - name: A
   - key: value
    - key: value
controller: 
    -Name B 
```

guilhem:
@camilamacedo86 I changed to new API for PROJECT and passed this PR as a draft to be certain it's not merged until the API is stabilized.

PR #5476, opened 2026-02-13
🐛 fix(helm/v2-alpha): Replace bitnami/kubectl:latest with busybox:1.37

camilamacedo86:
## Summary

Replace `bitnami/kubectl:latest` with `busybox:1.37` + runtime kubectl download from official Kubernetes releases (`dl.k8s.io`).

## Problem

The current Helm chart tests use `bitnami/kubectl:latest`, which has several issues:
- Third-party dependency (Bitnami) that could be deprecated or removed
- Unpredictable `:latest` tag breaks reproducibility
- Large image size (~130MB)
- Not aligned with Kubebuilder being an official Kubernetes project

## Investigation: Why Not Use registry.k8s.io Images?

We thoroughly investigated using **only** official Kubernetes registry images:

### Option 1: `registry.k8s.io/kubectl:v1.35.0`
❌ **No shell** - Image contains only the kubectl binary, cannot run test scripts that require `/bin/sh`

### Option 2: `registry.k8s.io/busybox:1.27.2`
❌ **Outdated TLS** - From 2017, uses old SSL/TLS libraries that fail to download from modern HTTPS endpoints

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/tests/test-manager-ready.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/tests/test-manager-ready.yaml
@@ -45,16 +45,6 @@ rules:
       - get
       - list
       - watch
-  {{- if .Values.certManager.enable }}
-  - apiGroups:
-      - cert-manager.io
-    resources:
-      - certificates
-    verbs:
-      - get
-      - list
-      - watch
-  {{- end }}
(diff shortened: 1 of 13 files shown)


camilamacedo86:
We cannot release without change the image for a image that we can trust in long term.

camilamacedo86:
Close we will need to revert this addition to the plugin
The complexities and risks in long term does not really justifiable value to end users

PR #5415, opened 2026-01-30
✨ (helm/v2-alpha): Add Helm chart tests and call in the GitHub action to validate the chart

camilamacedo86:
This commit adds built-in Helm tests to validate deployment health in
scaffolded Helm charts. Tests automatically verify that the manager
deployment, services, and certificates are properly configured and ready.

The test Pod uses bitnami/kubectl image and includes proper RBAC permissions
to check deployments, services, and certificates. Tests are automatically
cleaned up after completion via helm.sh/hook-delete-policy annotation.

This improves CI/CD reliability by catching deployment issues early and
provides a standardized testing approach for Helm chart validation.

PR #5478, opened 2026-02-14
⚠️  (helm/v2-alpha): remove helm test added (reverts pre-release addition to be safe)

camilamacedo86:
Remove Helm test scaffolding from helm/v2-alpha plugin add in #5415 before the v4.12.0 release.

## Background

Before finalizing the release, we re-evaluated the Helm test feature that was recently added. After further consideration, we've decided to revert it due to maintenance complexity and limited end-user value.

### 1. Image Dependency Challenges

The tests require an image with both shell and kubectl, leading to complex solutions:
- No suitable image exists in `registry.k8s.io` with modern TLS support
- `registry.k8s.io/busybox:1.27.2` is too old (2017) and cannot download from modern HTTPS
- `registry.k8s.io/kubectl` has no shell
- Requires downloading kubectl at runtime or using third-party images (see #5476)

### 2. Registry Reliability Concerns

We must be extremely careful about image dependencies. Past experience with `gcr.io/kubebuilder/kube-rbac-proxy` taught us this lesson:
- Registry became unavailable, breaking user projects
- Required migration effort from contributors
- See discussion: https://github.com/kubernetes-sigs/kubebuilder/discussions/3907

**Relying on external third-party artifacts creates long-term risks we cannot control.**

### 3. Air-Gapped Environment Issues

Downloading kubectl from `dl.k8s.io` at runtime creates problems for:
- Air-gapped environments
- Restricted networks
- Environments with strict image policies

### 4. Limited End-User Value

The tests only verify basic deployment status (pod running, deployment available), which:
- Kubernetes already validates via readiness probes
- Users can check with simple `kubectl get pods` commands
- Doesn't test actual controller functionality
- Runs AFTER installation, when problems are already visible

### 5. Maintenance Burden vs. Value

The complexity required to maintain this feature (image selection, kubectl download logic, multi-arch support, documentation) outweighs the minimal value it provides to end users.

## Going Forward

If Helm tests become valuable in the future (e.g., testing actual controller functionality), we can reconsider adding them with a better design. For now, users can verify deployments using standard kubectl commands or customize their charts with their own tests, going with something towards to: https://github.com/kubernetes-sigs/kubebuilder/pull/5476


camilamacedo86:
Hi @AlirezaPourchali and @vitorfloriano 

I am ping you both for you are aware of.
See the description, but after review all changes for the release I do not think we can move forward with this one.
I added previous problematic case scenarios.

Sorry, I was who proposed it but after re-consider it is not a good idea.
If we change our mind we can reconsider it always.
Thank you for all your collab.

Issue #5489, opened 2026-02-19
helm/v2-alpha: unergonomic env var support


porridge:
### What broke? What's expected?

Kubebuilder generates the `values.yaml` file with a bunch of env variables that are needed for the operator to work correctly. These are in the "list of key+value objects" format that the container spec expects. For example for my operator:
```
## Configure the controller manager deployment
##
manager:
[...]
  ## Environment variables
  ##
  env:
    - name: RELATED_IMAGE_MAIN
    - name: RELATED_IMAGE_SCANNER
    - name: RELATED_IMAGE_SCANNER_SLIM
    - name: RELATED_IMAGE_SCANNER_DB
    - name: RELATED_IMAGE_SCANNER_DB_SLIM
    - name: RELATED_IMAGE_COLLECTOR
    - name: RELATED_IMAGE_ROXCTL
    - name: RELATED_IMAGE_CENTRAL_DB
    - name: RELATED_IMAGE_SCANNER_V4_DB
    - name: RELATED_IMAGE_SCANNER_V4
    - name: MEMORY_LIMIT_BYTES
      valueFrom:
        resourceFieldRef:
            containerName: manager
            resource: limits.memory
```

So far, so good.

But now if a user _at install time_ wants to define an additional variable to tweak a feature, they cannot just `--set manager.env.SOMEVAR=true` (because `manager.env` is not a dict, it's a list), they need to set the whole `manager.env`, duplicating the whole list from stock `values.yaml`.

### Reproducing this issue

1. Run `kubebuilder edit --plugins=helm/v2-alpha` on a project which has non-empty `spec.template.spec.containers[0].env` in `operator/config/manager/manager.yaml`

### KubeBuilder (CLI) Version

v4.12.0

### PROJECT version

3

### Plugin versions

```yaml
layout:
- go.kubebuilder.io/v4
plugins:
  helm.kubebuilder.io/v2-alpha:
    manifests: dist/install.yaml
    output: dist
  manifests.sdk.operatorframework.io/v2: {}
  scorecard.sdk.operatorframework.io/v2: {}
```

asergeant01:
@porridge - I see what you mean, there is [a test](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater_test.go#L161) that even explicitly checks existing envs are removed - I'm not sure if this is a conscious design decision.

[...]

PR #5495, opened 2026-02-21
✨ (helm/v2-alpha): Add manager.envOverrides for CLI env overrides

camilamacedo86:
Charts define `manager.env` as a list to support Kubernetes env entries with `valueFrom` (e.g. `fieldRef`, `secretKeyRef`, etc.). The list format is required for operator defaults like `RELATED_IMAGE_*`, `MEMORY_LIMIT_BYTES`, `WATCH_NAMESPACE`, but it makes Helm CLI overrides painful because `--set` doesn’t target list items.

This PR adds `manager.envOverrides` (map) for install-time overrides/additions. The chart renders `manager.env` + `manager.envOverrides`, and when the same env var name exists in both, the override value wins.

Example:
```bash
helm install my-op ./dist/chart/ \
  --set manager.envOverrides.LOG_LEVEL=debug \
  --set manager.envOverrides.DEBUG=true
```

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/5489

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/values.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/values.yaml
@@ -25,6 +25,11 @@ manager:
   ##
   env: []
 
+  ## Env overrides (--set manager.envOverrides.VAR=value)
+  ## Same name in env above: this value takes precedence.
+  ##
+  envOverrides: {}
+
   ## Image pull secrets
   ##
   imagePullSecrets: []
--- a/docs/book/src/getting-started/testdata/project/dist/chart/values.yaml
+++ b/docs/book/src/getting-started/testdata/project/dist/chart/values.yaml
@@ -25,6 +25,11 @@ manager:
   ##
   env: []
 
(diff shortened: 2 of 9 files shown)


camilamacedo86:
@asergeant01 @porridge 

Could you please give a hand in this one?
Could you please review it?

porridge:
> @asergeant01 @porridge
> 
> Could you please give a hand in this one? Could you please review it?

Wow, thanks for the super fast turn around @camilamacedo86 ! I will be back in the office on Tuesday and should be able to take a look then.

asergeant01 (approved):
Looks good from my side

porridge:
This looks good @camilamacedo86 , can't wait to try this out 😉 

One potential issue is that some linters might complain about seeing a given var twice in `env`.
But if this is indeed a problem, it can probably be solved by changing `toYaml .Values.manager.env` to omit the entries  which are also included in `envOverrides`.

camilamacedo86:
@porridge 

If you face any problem please let us know, raise an issue for we have a better context.
Thank you for all your help !!!

Issue #5485, opened 2026-02-16
helm/v2-alpha: No support for configurable volumes/volumeMounts in generated charts


asergeant01:
### What do you want to happen?

I want to provide and load secrets to the container through volumes and volumeMounts which is a widely used approach.

This would be achieved by using the already configurable args to provide the file path to the controller.

```yaml
manager:

  args:
    - --leader-elect
    - --health-probe-bind-address=:8081
    - --super-secret-configuration-file=/etc/my-secrets
```

but also allowing the following amendments:

```yaml

# ... existing config (replicas, image, env, etc.) ...

  ## Additional volume mounts for the manager container
  ## These are in addition to webhook/metrics volumes which are handled automatically
  ##
  additionalVolumeMounts: []
    # - name: custom-volume
    #   mountPath: /etc/my-secrets
    #   readOnly: true

  ## Additional volumes for the manager pod
  ## These are in addition to webhook/metrics volumes which are handled automatically
  ##
  additionalVolumes: []
    # - name: custom-volume
    #    secret:
    #      name: super-secrets
```

camilamacedo86:
Keeping in mind that:
- Helm charts should be used to package the generated solution, not become another project that diverges from the the  scaffolding.
- Also, `values.yaml` should stay limited to options that make sense for end users and have sensible, working defaults.

**On the specific request to expose volumes/volumeMounts via values:** does it really make sense to let end users change those? In what concrete scenarios would an operator user need to override volumes?

I don’t think this is something we should expose by default for scaffolded projects. **If someone needs custom mounts, they can customize their own Helm chart to match their requirements since:**
- User chart customization is not overwritten when re-running `kubebuilder edit --plugins=helm/chart`.
- Custom changes should not be lost when the project is updated via the AutoUpdate plugin (GitHub Actions) or when running `kubebuilder alpha update`.

Given that, I suggest we close this as “deferred”, unless there’s a clear, common use case for the default projects produced by Kubebuilder where end users genuinely need to configure volumes via chart values.
What specific use cases do you have in mind?

isumitsolanki:
I want to work on this

camilamacedo86:
HI @isumitsolanki 

Thank you for looking on that. But I do not think we should move with this one. Please, see: https://github.com/kubernetes-sigs/kubebuilder/issues/5485#issuecomment-3909693851

asergeant01:
[...]

PR #5496, opened 2026-02-21
✨ (helm/v2-alpha): add extra volumes support

camilamacedo86:
Support manager.extraVolumes and manager.extraVolumeMounts: extract from kustomize (excluding webhook-certs/metrics-certs), inject into values when present, and template in manager deployment (including when volumes: []). Document in Helm v2-alpha plugin page.

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/5485

### DIFF
--- a/docs/book/src/plugins/available/helm-v2-alpha.md
+++ b/docs/book/src/plugins/available/helm-v2-alpha.md
@@ -318,6 +318,15 @@ prometheus:
   enable: false
 ```
 
+### Extra volumes
+
+The chart supports additional volumes and volume mounts for the manager (e.g. secrets, config files), alongside the built-in webhook and metrics cert volumes.
+
+- **Config volumes**: Volumes in the manager deployment (e.g. `config/manager/manager.yaml` or kustomize patches) are written into the chart template. Re-running `kubebuilder edit --plugins=helm/v2-alpha` updates the template from config; `values.yaml` is not overwritten.
+- **Values**: When the manager deployment has extra volumes (other than webhook/metrics), `values.yaml` gets `manager.extraVolumes` and `manager.extraVolumeMounts`. Use them to add more entries; the template appends them after the config volumes. Same structure as in a Pod spec; mount names must match volume names.
+
+Webhook and metrics (`webhook-certs`, `metrics-certs`) are not in `extraVolumes`. They are conditional on `certManager.enable` and `metrics.enable`, like the rest of the chart.
+
 ### Installation
 
 The first time you run the plugin, it adds convenient Helm deployment targets to your `Makefile`:
--- a/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/manager/manager.yaml
+++ b/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/manager/manager.yaml
@@ -87,6 +87,9 @@ spec:
           {}
           {{- end }}
         volumeMounts:
+          {{- if .Values.manager.extraVolumeMounts }}
+          {{- toYaml .Values.manager.extraVolumeMounts | nindent 10 }}
+          {{- end }}
         {{- if and .Values.certManager.enable .Values.metrics.enable }}
         - mountPath: /tmp/k8s-metrics-server/metrics-certs
           name: metrics-certs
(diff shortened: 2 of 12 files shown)


docs/book/src/plugins/available/helm-v2-alpha.md
```diff
@@ -318,6 +318,15 @@ prometheus:
   enable: false
 ```
 
+### Extra volumes
+
+The chart supports extra volumes and volume mounts for the manager (e.g. secrets, config files) in addition to the built-in webhook and metrics cert volumes.
+
+- **From kustomize**: If your deployment in `dist/install.yaml` includes volumes other than the default `webhook-certs` and `metrics-certs` (from the scaffolded patches), the plugin adds them to `values.yaml` as `manager.extraVolumes` and `manager.extraVolumeMounts`. The manager template merges them into the deployment.
```
asergeant01: To clarify this point, if they are defined in the kustomize output, other than metrics-certs and web hook-certs, the values are extracted and added to the values?

I would have expected them to be added into the chart template and not the values. As any further additions would require the plugin to be run with --force.

camilamacedo86: > To clarify this point, if they are defined in the kustomize output, other than metrics-certs and web hook-certs, the values are extracted and added to the values?

Yes. 

> I would have expected them to be added into the chart template and not the values. As any further additions would require the plugin to be run with --force.

Yes, that is how it is. To update the values you need to run the force. 
We cannot update the values without force otherwise it would lost the customizations.

asergeant01: I think you missed my point or I didn't articulate myself very well. 

I think if the volumes are defined in kustomize, they should be added directly to the template and NOT to the values.yaml

The values.yaml should only be for additional volumes IMO.

camilamacedo86: > I think if the volumes are defined in kustomize,

All should be defined in the kustomize.
They are the source of truth.
The Helm Chart is just a way to package the solution in this format. 
It should not deviate from the project config, and we should not add anything in the Helm charts. It only package the project for this format.

asergeant01: My example:

```yaml
# config/manager/manager.yaml (or a kustomize patch)
spec:
  template:
    spec:
      volumes:
        - name: app-secret-1
          secret:
            secretName: app-secret-1
```

```yaml

# dist/chart/values.yaml
manager:
  extraVolumes:
    - name: app-secret-1
      secret:
        secretName: app-secret-1
```

```yaml
# dist/chart/templates/deployment.yaml (snippet)
spec:
  template:
    spec:
      volumes:
        {{- if .Values.manager.extraVolumes }}
        {{- toYaml .Values.manager.extraVolumes | nindent 8 }}
        {{- end }}
```

Lets say the maintainer makes modifications to the `values.yaml` file but also adds another volume to the config....

```yaml
# config/manager/manager.yaml (or a kustomize patch)
spec:
  template:
    spec:
      volumes:
        - name: app-secret-1
          secret:
            secretName: app-secret-1
        - name: app-secret-2
          secret:
            secretName: app-secret-2
```

The only way to have that new volume in the chart is either add it manually to the `values.yaml` or re-run `kubebuilder edit --plugins=helm/v2-alpha --force` which would overwrite any manual changes.


What I am saying is, I believe the generated template should be:

```yaml
# dist/chart/templates/deployment.yaml (snippet)
spec:
  template:
    spec:
      volumes:
        - name: app-secret-1
          secret:
            secretName: app-secret-1
        - name: app-secret-2
          secret:
            secretName: app-secret-2
        {{- if .Values.manager.extraVolumes }}
        {{- toYaml .Values.manager.extraVolumes | nindent 8 }}
        {{- end }}
```

camilamacedo86: Hi @asergeant01 

Let me know if the changes in place address your request.

asergeant01: @camilamacedo86 I didn't test it, but it certainly looks like it addresses the request. Thank you

camilamacedo86:
I will move forward with this one otherwise, it will be hard for others contribute with the open RFEs for helm/v2-alpha

Issue #5504, opened 2026-02-25
helm/v2-alpha: disable cluster wide rbac resource creation


Allex1:
### What do you want to happen?

In some multi-tenant Kubernetes environments the user cannot load the ClusterRoles and ClusterRoleBindings during normal deployment. In this case we side-load them via a different process before the actual helm deployment. For this to be possible we should have a feature flag (enabled by default) that allow the use to choose whether to load these resources while deploying the helm release.
For allowing the operator to still be able to manage resources at namespace level a local rb should bind the operator serviceaccount to the clusterRole deployed outside of this chart (by a cluster admin) with then name being configurable.

PR #5505, opened 2026-02-25
(helm/v2-alpha): disable cluster wide rbac resource creation

Allex1:
Addresses https://github.com/kubernetes-sigs/kubebuilder/issues/5504

### DIFF
--- a/docs/book/src/getting-started/testdata/project/dist/chart/values.yaml
+++ b/docs/book/src/getting-started/testdata/project/dist/chart/values.yaml
@@ -72,6 +72,17 @@ manager:
   ##
   tolerations: []
 
+## RBAC configuration
+##
+rbac:
+  ## Cluster-scoped RBAC resources (ClusterRole/ClusterRoleBinding)
+  ##
+  clusterScope:
+    # Set to false to skip cluster-scoped RBAC, useful when the operator
+    # should be restricted to a single namespace or when cluster-wide
+    # permissions are managed externally.
+    enabled: true
+
 ## Helper RBAC roles for managing custom resources
 ##
 rbacHelpers:
--- a/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
+++ b/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
@@ -1491,7 +1491,12 @@ func (t *HelmTemplater) addConditionalWrappers(yamlContent string, resource *uns
 			// Metrics RBAC depends on metrics being enabled
 			return fmt.Sprintf("{{- if .Values.metrics.enable }}\n%s{{- end }}\n", yamlContent)
 		}
-		// Essential RBAC (controller-manager, leader-election, manager roles) - always enabled
+		// Cluster-scoped RBAC (ClusterRole/ClusterRoleBinding) - conditional on rbac.clusterScope
+		// This allows operators to be deployed without cluster-wide permissions when needed
+		if kind == kindClusterRole || kind == kindClusterRoleBinding {
+			return fmt.Sprintf("{{- if .Values.rbac.clusterScope.enabled }}\n%s{{- end }}\n", yamlContent)
+		}
+		// Namespace-scoped RBAC (Role/RoleBinding, ServiceAccount) - always enabled
 		// These are required for the controller to function properly
 		return yamlContent
 	case kind == kindValidatingWebhook || kind == kindMutatingWebhook:
--- a/testdata/project-v4-with-plugins/dist/chart/values.yaml
+++ b/testdata/project-v4-with-plugins/dist/chart/values.yaml
@@ -80,6 +80,17 @@ manager:
   ##
   tolerations: []
 
+## RBAC configuration
+##
+rbac:
+  ## Cluster-scoped RBAC resources (ClusterRole/ClusterRoleBinding)
+  ##
+  clusterScope:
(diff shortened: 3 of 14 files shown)


docs/book/src/getting-started/testdata/project/dist/chart/values.yaml
```diff
@@ -72,6 +72,17 @@ manager:
...
+##
+rbac:
+  ## Cluster-scoped RBAC resources (ClusterRole/ClusterRoleBinding)
+  ##
+  clusterScope:
+    # Set to false to skip cluster-scoped RBAC, useful when the operator
+    # should be restricted to a single namespace or when cluster-wide
+    # permissions are managed externally.
+    enabled: true
+
```
camilamacedo86, line 85: We have either :

```
## Helper RBAC roles for managing custom resources
##
rbacHelpers:
  # Install convenience admin/editor/viewer roles for CRDs
  enable: false
```

So, should we have rbac:
Then, insde helpers and default for example instead?

How that would work when the solution is scaffold with the namespaced flag and the RBAC are not cluster scopes? How the Helm Chart would be in this case?

testdata/project-v4-with-plugins/dist/chart/values.yaml
```diff
@@ -80,6 +80,17 @@ manager:
...
+## RBAC configuration
+##
+rbac:
+  ## Cluster-scoped RBAC resources (ClusterRole/ClusterRoleBinding)
+  ##
+  clusterScope:
+    # Set to false to skip cluster-scoped RBAC, useful when the operator
+    # should be restricted to a single namespace or when cluster-wide
+    # permissions are managed externally.
+    enabled: true
```
camilamacedo86, line 92: Here for example we have no cluster scope rules since this sample is namespaced scope. 
See; https://book.kubebuilder.io/migration/namespace-scoped.html

camilamacedo86 (changes requested):
Thanks for raising this — I think we need think through a few design questions before we decide:

a) What’s the best scaffolding approach to support both namespaced and cluster-scoped solutions?
b) We have RBAC helper, so what’s the cleanest way to model/parameterize them in values.yaml?
c) Does it actually make sense to expose this in values at all? Should this be a decision made by the consumer rather than the project authors?

docs/book/src/getting-started/testdata/project/dist/chart/values.yaml
Allex1: @camilamacedo86 I could reuse the rbacHelpers.enabled for the main clusterrole/crb if that's what you're asking. The problem right now is that you can only generate helm charts only for cluster admins.

testdata/project-v4-with-plugins/dist/chart/values.yaml
Allex1: ok, that could be useful but what I'm trying to achieve is having generated helm charts templates for [clickhouse-operator](https://github.com/ClickHouse/clickhouse-operator/pull/99) that can work for both cluster admins and name-spaced users.

Allex1:
@camilamacedo86 to answer your questions:

a) I'm fine either way: we could reuse the `rbacHelpers` for the mai cr/crb or add a new key `clusterScope`
b) we can use the `enabled` as a catchall or do you prefer creating a sub key ?
c) Not sure I understand, I think we should keep it enabled by default in order to not create breaking changes, users can then choose to not deploy the cr/crb by setting the flag to false.

docs/book/src/getting-started/testdata/project/dist/chart/values.yaml
camilamacedo86: I see I am think in a way to make it generic enough and address this need.
Could you give me 1 to 2 weeks?  I will push that and you can help us to know if that is a good approach. Your collab is very required and welcome btw.

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
```diff
@@ -1491,7 +1491,12 @@ func (t *HelmTemplater) addConditionalWrappers(yamlContent string, resource *uns
 			// Metrics RBAC depends on metrics being enabled
 			return fmt.Sprintf("{{- if .Values.metrics.enable }}\n%s{{- end }}\n", yamlContent)
 		}
-		// Essential RBAC (controller-manager, leader-election, manager roles) - always enabled
+		// Cluster-scoped RBAC (ClusterRole/ClusterRoleBinding) - conditional on rbac.clusterScope
+		// This allows operators to be deployed without cluster-wide permissions when needed
+		if kind == kindClusterRole || kind == kindClusterRoleBinding {
+			return fmt.Sprintf("{{- if .Values.rbac.clusterScope.enabled }}\n%s{{- end }}\n", yamlContent)
+		}
+		// Namespace-scoped RBAC (Role/RoleBinding, ServiceAccount) - always enabled
```
camilamacedo86, line 1499: I liked 👍

docs/book/src/getting-started/testdata/project/dist/chart/values.yaml
camilamacedo86: Hi @Allex1 

I was looking on this one and based on your amazing contribution I came with: https://github.com/kubernetes-sigs/kubebuilder/pull/5579

Could you please check out? 
Let me know wdyt? 
Also, could you please help us with reviews and to validating this one.

Allex1:
replaced by https://github.com/kubernetes-sigs/kubebuilder/pull/5579

Issue #5500, opened 2026-02-24
Kubebuilder v4.12.0 panics on startup: panic: edit flag redefined: force


sandeepkunusoth:
After upgrading to Kubebuilder v4.12.0, the binary panics immediately during CLI initialization — even when running kubebuilder version.

This appears to be caused by a duplicate --force flag registration inside the Helm v2-alpha plugin (edit subcommand).

Environment

Kubebuilder version: v4.12.0

OS: Linux (GitHub Actions Ubuntu runner)

GOOS/GOARCH: linux/amd64

Installation method:

```
curl -L -o kubebuilder "https://go.kubebuilder.io/dl/latest/$(go env GOOS)/$(go env GOARCH)"
chmod +x kubebuilder
sudo mv kubebuilder /usr/local/bin/
kubebuilder version
```

https://github.com/valkey-io/valkey-operator/actions/runs/22331029600/job/64613161970

error
```
panic: edit flag redefined: force

goroutine 1 [running]:
github.com/spf13/pflag.(*FlagSet).AddFlag(0xc000019100, 0xc0002a9540)
	/home/runner/go/pkg/mod/github.com/spf13/pflag@v1.0.10/flag.go:878 +0x5ec
github.com/spf13/pflag.(*FlagSet).VarPF(0xc000019100, {0xeb9db0, 0xc00016c790}, {0xce799c, 0x5}, {0x0, 0x0}, {0xd3e2cb, 0x22})
	/home/runner/go/pkg/mod/github.com/spf13/pflag@v1.0.10/flag.go:861 +0x105
github.com/spf13/pflag.(*FlagSet).BoolVarP(0x7f6130a26530?, 0xc0002910f8?, {0xce799c?, 0xc0dd00?}, {0x0?, 0xd63201?}, 0x30?, {0xd3e2cb?, 0xc000180b29?})
	/home/runner/go/pkg/mod/github.com/spf13/pflag@v1.0.10/bool.go:55 +0x3e
github.com/spf13/pflag.(*FlagSet).BoolVar(...)
	/home/runner/go/pkg/mod/github.com/spf13/pflag@v1.0.10/bool.go:50
sigs.k8s.io/kubebuilder/v4/pkg/plugins/optional/helm/v2alpha.(*editSubcommand).BindFlags(0xc00016c780, 0xc000019100)
[...]

PR #5506, opened 2026-02-25
🐛 fix: resolve panic when multiple plugins define the same flag (e.g. edit with go/v4 + helm)

camilamacedo86:
- Merge duplicate flags instead of binding twice; sync parsed value to all plugins in PreRunE.
- Aggregate flag help as "For plugin (key): desc AND for plugin (key): desc".
- Return a clear error when the same flag name is used with different types (e.g. bool vs string).

Example:

```
$ kubebuilder edit --plugins=go/v4,helm/v2-alpha --help
...

Flags:
      --force               For plugin (base.go.kubebuilder.io/v4): overwrite scaffolded files to apply changes (manual edits may be lost) AND for plugin (helm.kubebuilder.io/v2-alpha): if true, regenerates all the files
``` 

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/5500

### DIFF
--- a/pkg/cli/cmd_helpers.go
+++ b/pkg/cli/cmd_helpers.go
@@ -20,8 +20,10 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 
 	"sigs.k8s.io/kubebuilder/v4/pkg/config"
 	"sigs.k8s.io/kubebuilder/v4/pkg/config/store"
@@ -115,8 +117,9 @@ func collectSubcommands(
 	}}
 }
 
-// applySubcommandHooks runs the initialization hooks and configures the commands pre-run,
-// run, and post-run hooks with the appropriate execution hooks.
+// applySubcommandHooks runs the initialization hooks and wires pre-run, run, and post-run for the command.
+// Used by init, create api, create webhook, and edit. When several plugins define the same flag,
+// one flag is shown and its value is synced to all plugins after parse.
 func (c *CLI) applySubcommandHooks(
 	cmd *cobra.Command,
 	subcommands []keySubcommandTuple,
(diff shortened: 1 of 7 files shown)


pkg/cli/cmd_helpers.go
```diff
@@ -166,12 +174,74 @@ func (c *CLI) appendPluginTable(cmd *cobra.Command, filter func(plugin.Plugin) b
...
+// the flag is not added again; its Value is stored in duplicateValues for later sync and the existing
+// Usage is extended. Returns an error if the same flag name is used with a different value type.
+func mergeFlagSetInto(
+	dest *pflag.FlagSet,
+	src *pflag.FlagSet,
+	duplicateValues map[string][]pflag.Value,
+	pluginKey string,
+	firstPluginByFlag map[string]string,
+) error {
+	var err error
```
v47, line 200: Sorry, early returns will make it easier to follow. I mean something like:
```
	var err error

	src.VisitAll(func(flag *pflag.Flag) {
		if err != nil {
			return
		}

		existing := dest.Lookup(flag.Name)
		if existing == nil {
			dest.AddFlag(flag)
			added := dest.Lookup(flag.Name)
			firstPluginByFlag[flag.Name] = pluginKey
			added.Usage = "For plugin (" + pluginKey + "): " + strings.TrimSpace(flag.Usage)
			return
		}

		if existing.Value.Type() != flag.Value.Type() {
			firstKey := firstPluginByFlag[flag.Name]
			err = fmt.Errorf(
				"plugins %q and %q use the same flag name %q but expect different value types: one %s, other %s",
				firstKey, pluginKey, flag.Name, existing.Value.Type(), flag.Value.Type(),
			)
			return
		}

		duplicateValues[flag.Name] = append(duplicateValues[flag.Name], flag.Value)
		existing.Usage += " AND for plugin (" + pluginKey + "): " + strings.TrimSpace(flag.Usage)
	})

	return err
```

pkg/cli/cmd_helpers.go
```diff
@@ -166,12 +174,74 @@ func (c *CLI) appendPluginTable(cmd *cobra.Command, filter func(plugin.Plugin) b
...
+// duplicate-flag values to sync after parse.
+type initHooksResult struct {
+	options             *resourceOptions
+	duplicateFlagValues map[string][]pflag.Value
+}
+
+// mergeFlagSetInto merges flags from src into dest. If a flag name already exists,
+// the flag is not added again; its Value is stored in duplicateValues for later sync and the existing
+// Usage is extended. Returns an error if the same flag name is used with a different value type.
+func mergeFlagSetInto(
```
vitorfloriano, line 187: Couldn't we use `pflag.FlagSet.AddFlagSet` instead? See: https://pkg.go.dev/github.com/spf13/pflag#FlagSet.AddFlagSet

pkg/cli/cmd_helpers.go
```diff
@@ -166,12 +174,77 @@ func (c *CLI) appendPluginTable(cmd *cobra.Command, filter func(plugin.Plugin) b
...
+			existing.Usage = "For plugin (" + pluginKey + "): " + strings.TrimSpace(flag.Usage)
+			return
+		}
+		if existing.Value.Type() != flag.Value.Type() {
+			firstKey := firstPluginByFlag[flag.Name]
+			err = fmt.Errorf(
+				"plugins %q and %q use the same flag name %q but expect different value types: one %s, other %s",
+				firstKey, pluginKey, flag.Name, existing.Value.Type(), flag.Value.Type(),
+			)
+			return
```
vitorfloriano, line 217: ~~Is there anything we can do here to not fail the operation if this happens?~~ 

Thinking again, I think it's better to fail than trying to make it work.

camilamacedo86: If the type is not the same, then having the same name is not the same thing. We need to fail.

PR #5507, opened 2026-02-25
🌱 chore: add yamllint config for 2-space indentation and Kubernetes YAML best practices

camilamacedo86:


vitorfloriano:
Q: Why not just run the yaml linter on testdata in CI?

camilamacedo86:
> Q: Why not just run the yaml linter on testdata in CI?

We are running in on testdata in CI
We are either running it for the samples used for the docs.

That is what this PR does.
But we cannot run directly over the Helm Charts because those have `{{}}`

Therefore this PR:
- Ensure that we call the check for helm charts output and config yamls

Makefile
```diff
@@ -115,10 +115,26 @@ lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
...
+	@files=$$(find testdata -name '*.yaml' ! -path 'testdata/.helm-rendered.yaml' \( ! -path 'testdata/*/dist/*' -o -path 'testdata/*/dist/chart/Chart.yaml' -o -path 'testdata/*/dist/chart/values.yaml' \) 2>/dev/null); \
+	docker run --rm $$(tty -s && echo "-it" || echo) -v $(PWD):/data -w /data cytopia/yamllint:latest $$files -c .yamllint --no-warnings
+
+.PHONY: yamllint-helm-charts
+yamllint-helm-charts:
+	@which helm >/dev/null 2>&1 || { echo "Skipping yamllint-helm-charts (helm not in PATH)"; exit 0; }; \
+	for chart in $$(find testdata docs/book -type d -path '*/dist/chart' 2>/dev/null); do \
+	  echo "Checking Helm chart output (2-space indentation): $$chart"; \
+	  helm template release $$chart --namespace=release-system 2>/dev/null | \
+	  docker run --rm -i -v $(PWD):/data -w /data cytopia/yamllint:latest \
```
v47: Using the latest is more flexible, but can lead to sudden changes in behaviour.

Makefile
```diff
@@ -115,10 +115,26 @@ lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
...
+	  -d "{extends: relaxed, rules: {indentation: {spaces: 2, indent-sequences: whatever}}}" --no-warnings - || \
```
v47: For the Helm-rendered YAML linting: instead of embedding a separate inline ruleset via `yamllint -d ...`. I would suggest using a second config file that extends the repo’s `.yamllint`.

For example:
	•	Keep `.yamllint` as the primary policy for repo YAML
	•	Add `.yamllint-helm`:
```
extends: .yamllint
rules:
  # Rendered output is generated; relax rules that are noisy for templates.
  # (Example: line-length, trailing-spaces, etc — whatever you find too strict.)
  line-length: disable
```
Then it can be used this way:
	•	`yamllint -c .yamllint ...` for normal YAML files
	•	`yamllint -c .yamllint-helm <rendered-output>` for Helm output

This keeps the policy DRY and makes it obvious which rules differ for rendered/generated manifests.

Makefile
camilamacedo86: It is fine because it is only in our Makefile
So, does not impact end users, only us Maintainers, and if an issue occurs, we can easily fix it.

camilamacedo86: It is less of a burn to keep it up to date than to bump a version and then maintain it.

Makefile
camilamacedo86: DOne

camilamacedo86:
HI @v47 

All simplified and done !!!
Thank you for your collab 🥇

v47:
Looks good.
Thanks for reinforcing the CI.

Issue #5502, opened 2026-02-25
helm/v2-alpha: inconsistent dist/chart file permissions


porridge:
### What broke? What's expected?

Some files/directories are created with the expected 0644/0755 permissions, but others with 0700/0600.
The restrictive permissions cause problems when using some containerized workflows.

Judging by some logs, apart from the below, I think it also creates other files/directories such as `.github/` as 0700 if it's missing.

```
 27163740      0 drwxr-xr-x   1 4200660  4200660         34 Feb 25 08:06 ./operator/dist
 27163741    248 -rw-r--r--   1 4200660  4200660     253505 Feb 25 08:06 ./operator/dist/install.yaml
 27163742      0 drwxr-xr-x   1 4200660  4200660         82 Feb 25 08:06 ./operator/dist/chart
 27163743      0 drwxr-xr-x   1 4200660  4200660        104 Feb 25 08:06 ./operator/dist/chart/templates
 27163781      0 drwxr-xr-x   1 4200660  4200660        230 Feb 25 08:06 ./operator/dist/chart/templates/crd
 27163782    176 -rw-r--r--   1 4200660  4200660     178581 Feb 25 08:06 ./operator/dist/chart/templates/crd/centrals.platform.stackrox.io.yaml
 27163783    156 -rw-r--r--   1 4200660  4200660     159399 Feb 25 08:06 ./operator/dist/chart/templates/crd/securedclusters.platform.stackrox.io.yaml
 27163784     16 -rw-r--r--   1 4200660  4200660      16273 Feb 25 08:06 ./operator/dist/chart/templates/crd/securitypolicies.config.stackrox.io.yaml
 27163785      0 drwxr-xr-x   1 4200660  4200660        354 Feb 25 08:06 ./operator/dist/chart/templates/rbac
 27163786      4 -rw-r--r--   1 4200660  4200660        229 Feb 25 08:06 ./operator/dist/chart/templates/rbac/controller-manager.yaml
 27163787      4 -rw-r--r--   1 4200660  4200660        763 Feb 25 08:06 ./operator/dist/chart/templates/rbac/leader-election-role.yaml
 27163788      4 -rw-r--r--   1 4200660  4200660        883 Feb 25 08:06 ./operator/dist/chart/templates/rbac/manager-role.yaml
 27163789      4 -rw-r--r--   1 4200660  4200660        328 Feb 25 08:06 ./operator/dist/chart/templates/rbac/metrics-reader.yaml
[...]

PR #5508, opened 2026-02-26
🐛 fix(go/v4,helm/v2-alpha): use 0644/0755 for generated files

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/5502


pkg/cli/cli.go
```diff
@@ -310,7 +310,7 @@ func patchProjectFileInMemoryIfNeeded(fs afero.Fs, path string) error {
 	}
 
 	if modified != original {
-		err := afero.WriteFile(fs, path, []byte(modified), 0o755)
+		err := afero.WriteFile(fs, path, []byte(modified), 0o644)
```
v47: Below in `pkg/machinery/scaffold.go`, there are constants. Defining them in one place will create consistency and provide a single source of truth. No drifts in future.

pkg/config/store/yaml/store.go
```diff
@@ -144,7 +144,7 @@ func (s yamlStore) SaveTo(path string) error {
 	content = append([]byte(commentStr), content...)
 
 	// Write the marshalled configuration
-	err = afero.WriteFile(s.fs, path, content, 0o600)
+	err = afero.WriteFile(s.fs, path, content, 0o644)
```
v47: Below in `pkg/machinery/scaffold.go`, there are constants. Defining them in one place will create consistency and provide a single source of truth. No drifts in future.

pkg/cli/cli.go
camilamacedo86: Good catcher 👍 Done

pkg/config/store/yaml/store.go
camilamacedo86: Good catcher 👍 Done

Makefile
```diff
@@ -132,6 +132,20 @@ yamllint-helm:
...
+SAMPLE_ROOTS := testdata \
+	docs/book/src/getting-started/testdata \
+	docs/book/src/cronjob-tutorial/testdata \
+	docs/book/src/multiversion-tutorial/testdata
+
+.PHONY: check-sample-permissions
+check-sample-permissions: ## Fail if any file/dir under testdata or docs samples has wrong permissions (expect 0644/0755). bin/ and *.sh excluded.
+	@for d in $(SAMPLE_ROOTS); do \
+		test -d "$$d" || continue; \
+		bad=$$(find "$$d" -path '*/bin' -prune -o \( \( -type f ! -perm 0644 ! -name '*.sh' \) -o \( -type d ! -perm 0755 \) \) -print 2>/dev/null); \
```
vitorfloriano: What if a `*.sh` file has the wrong permissions, how would we catch it?

camilamacedo86: I think we can remove that we have no longer sh files in the testdata.

camilamacedo86: Done

PR #5513, opened 2026-02-26
use an isolated kind config for executing e2e tests in order to avoid disruption other kubernetes activities to other contexts

rotty3000:
- ✨ (:sparkles:): new non-breaking feature
This change just has kind use a separate (and ignored) kind config which it then passes to the e2e tests invocation. The result is that it does not touch the local kube config file and in so doing avoids disruption of other kubernetes activities.

### DIFF
--- a/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
+++ b/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
@@ -145,6 +145,7 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
 # CertManager is installed by default; skip with:
 # - CERT_MANAGER_INSTALL_SKIP=true
 KIND_CLUSTER ?= {{ .ProjectName }}-test-e2e
+KIND_KUBECONFIG ?= test-e2e-kubeconfig.yaml~
 
 .PHONY: setup-test-e2e
 setup-test-e2e: ## Set up a Kind cluster for e2e tests if it does not exist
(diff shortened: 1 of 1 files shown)


pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go
```diff
@@ -157,12 +158,12 @@ setup-test-e2e: ## Set up a Kind cluster for e2e tests if it does not exist
 			echo "Kind cluster '$(KIND_CLUSTER)' already exists. Skipping creation." ;; \
 		*) \
 			echo "Creating Kind cluster '$(KIND_CLUSTER)'..."; \
-			$(KIND) create cluster --name $(KIND_CLUSTER) ;; \
+			$(KIND) create cluster --name $(KIND_CLUSTER) --kubeconfig $(KIND_KUBECONFIG) ;; \
```
camilamacedo86, line 161: @rotty3000

Thanks for the PR!

Just to note: we already run the tests in an isolated KIND cluster using `$(KIND_CLUSTER)`. So the test cluster itself is already separated.

We don’t want to pass a custom kubeconfig by default. By default, `kubectl`, `helm`, `terraform` (and any other Makefile target) use the user’s default kubeconfig/context. That is the normal Kubernetes workflow and what most users expect.

If someone wants a separate kubeconfig, they can always customize their own scaffold/Makefile to do that. But making it the default in Kubebuilder scaffolding does not seems the right direction.

Thanks again!

rotty3000: I'm sorry if it wasn't clear. The goal is NOT to use the default context.

Suppose I am using K3D and have active connections to that cluster. OR I am connected to GCP's Kubernetes via my default config... now when I execute my e2e tests it will disconnect me from those clusters and I have to go back and reconnect everything. 

The idea is that e2e tests should NOT disrupt my ongoing work outside of tests.

:)

camilamacedo86: @rotty3000

Thanks for clarifying — I understand your concern.

You are absolutely right that running `kind create cluster` can switch the current context and that this may disrupt someone who is actively working with another cluster (k3d, GKE, etc.).

However, this behavior comes from how Kind integrates with the default kubeconfig and current-context. It is not something specific to Kubebuilder’s scaffolding — it is part of the normal Kubernetes tooling workflow.

In our case, the scaffolding already:
- Creates a dedicated Kind cluster name for e2e (`$(KIND_CLUSTER)`).
- Cleans it up after the tests.
- Assumes standard kubeconfig behavior like any other Kubernetes-based workflow.

Changing the default template to always use a separate kubeconfig introduces:
- A non-standard workflow compared to typical Kubernetes tooling.
- Extra files/artifacts in the project directory.
- Additional complexity in the default boilerplate for a scenario that depends on a specific local workflow.

Your use case is valid, but it is also environment-specific. Many users rely on the default kubeconfig behavior and expect the context to switch when a cluster is created.

If someone prefers not to touch their default kubeconfig, they can:
- Override the Makefile in their own project, or
**- Export `KUBECONFIG` manually before running tests, or ( I would recommend that )**
- Customize the scaffold to pass `--kubeconfig`.

For these reasons, we don’t think this should be the default behavior in Kubebuilder’s boilerplate.

Thanks again for raising the concern and for the contribution.

rotty3000: no problem.

fbozic: I would personally like to see this (or a similar form of this) in the kubebuilder. 

Here is issue that we have faced:
- We started e2e tests for our operator locally.
- Tests are running for a while.
- We have accidentally changed current context to point to actual cluster.
- Once tests have finished we have deleted cert manager in actual cluster, causing issues for other teams.

I understand that isolated kubeconfig might not be the best solution but some form of "safety net" would be nice. I can imagine this happening to someone else as well.

PR #5516, opened 2026-02-27
🐛 fix(API): Move test/e2e/helpers to test/e2e/internal

vitorfloriano:
This PR moves the `test/e2e/helpers` package to `./test/e2e/internal/`.

This package is not meant to be imported by external users but has been exported to the public API (see [here](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.12.0/test/e2e/helpers)).

It is underdocumented and is not stable, hence it belongs in internal.

> [!IMPORTANT]
> The test/e2e/helpers package has not been imported by any other modules, as per the pkg.go.dev import stats (see [here](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.12.0/test/e2e/helpers?tab=importedby)), so **this is not a breaking change for any users**.

---

The `test` subdir structure will go from this:

```
├── test
│   ├── e2e
│   │   ├── all
│   │   ├── helpers    <--- in the public API
│   │   └── utils
│   └── testdata
```

To this:

```
├── test
│   ├── e2e
│   │   ├── all
│   │   ├── internal
│   │   │   └── helpers
│   │   └── utils
│   └── testdata
```
---

We decided to not move the `test/e2e/helpers` package to `./internal/test` to keep the package next to where it is consumed. Since it is only consumed by packages in the `e2e` subdir, we can place the internal subdir in the same level as `all` and `utils`.

> [!NOTE]
> We can't move package `test/e2e/utils` to `internal` because it would be a breaking change for existing dependants (see importers [here](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.12.0/test/e2e/utils?tab=importedby)).

Related to: https://github.com/kubernetes-sigs/kubebuilder/discussions/5277 (see https://github.com/kubernetes-sigs/kubebuilder/discussions/5277#discussioncomment-15311956)


test/e2e/all/plugin_helm_test.go
```diff
@@ -28,7 +28,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	pluginutil "sigs.k8s.io/kubebuilder/v4/pkg/plugin/util"
-	"sigs.k8s.io/kubebuilder/v4/test/e2e/helpers"
+	"sigs.k8s.io/kubebuilder/v4/test/e2e/internal/helpers"
```
camilamacedo86, line 31: It will impact SDK right?
So, I am not sure if we should do it.

vitorfloriano: Operator-SDK does not import `test/e2e/helpers` anywhere, as per [this search result](https://github.com/search?q=repo%3Aoperator-framework%2Foperator-sdk+e2e%2Fhelpers&type=code).

Operator-SDK only imports `test/e2e/utils`, as per [this search result](https://github.com/search?q=repo%3Aoperator-framework%2Foperator-sdk+e2e%2Futils&type=code), that's why we are not moving it to internal.

Also, as per the [import stats](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.12.0/test/e2e/helpers?tab=importedby) of this package in `pkg.go.dev`, there are no known importers of `test/e2e/helpers`.

So I assume it's safe to move it to internal.

PR #5520, opened 2026-02-27
WIP 🐛 fix(API): Move autoupdate/deploy-image/helm packages to internal

vitorfloriano:
This PR moves the packages for the optional plugins `autoupdate`, `deploy-image` and `helm` to `./internal/plugins/optional`.

Those optional plugins are meant only for CLI users and their packages should not be part of the public API of the SDK, considering that **they are underdocumented and not stable**, hence they belong in internal.

Here's a summary of the current situation of each plugin:

| package | no known importers |
|--------|--------|
| [autoupdate/v1alpha](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.13.0/pkg/plugins/optional/autoupdate/v1alpha) | [import stats](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.13.0/pkg/plugins/optional/autoupdate/v1alpha?tab=importedby) |
| [autoupdate/v1alpha/scaffolds](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.13.0/pkg/plugins/optional/autoupdate/v1alpha/scaffolds) | [import stats](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.13.0/pkg/plugins/optional/autoupdate/v1alpha/scaffolds?tab=importedby) |
| [deploy-image/v1alpha](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.13.0/pkg/plugins/golang/deploy-image/v1alpha1) | [import stats](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.13.0/pkg/plugins/golang/deploy-image/v1alpha1?tab=importedby) |
| [deploy-imagem/v1alpha/scaffolds](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.13.0/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds) | [import stats](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.13.0/pkg/plugins/golang/deploy-image/v1alpha1/scaffolds?tab=importedby) | 
[...]

PR #5536, opened 2026-03-12
🐛 fix(API): move alphaupdate plugin to internal

vitorfloriano:
This PR moves the Alpha Update optional plugin to `./internal/plugins/optional`.

The Alpha Update optional plugin is meant for CLI users only and its packages should not be part of the public API of the SDK. Also, its packages are underdocumented and not stable.

The `autoupdate` packages do not seem to have already been imported by other modules:

| package | no known importers |
|--------|--------|
| [autoupdate/v1alpha](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.13.0/pkg/plugins/optional/autoupdate/v1alpha) | [import stats](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.13.0/pkg/plugins/optional/autoupdate/v1alpha?tab=importedby) |
| [autoupdate/v1alpha/scaffolds](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.13.0/pkg/plugins/optional/autoupdate/v1alpha/scaffolds) | [import stats](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4@v4.13.0/pkg/plugins/optional/autoupdate/v1alpha/scaffolds?tab=importedby) |

---

The project structure will go from this:

```
├── internal
│   ├── cli
│   └── logging
├── pkg
│   └── plugins
│       ├── common
│       ├── external
│       ├── golang
│       │   ├── deploy-image
│       │   └── v4
│       └── optional
│           ├── autoupdate     <--- in the public API
│           ├── grafana
│           └── helm
```

To this:

```
├── internal
│   ├── cli
│   ├── logging
│   └── plugins
│       └── optional
│           └── autoupdate     <--- not importable anymore
├── pkg
│   └── plugins
│       ├── common
│       ├── external
│       ├── golang
│       │   ├── deploy-image
│       │   └── v4
│       └── optional
│           ├── grafana
│           └── helm
```

---

This is part of the effort to improve Kubebuilder's SDK API. 

Related to: https://github.com/kubernetes-sigs/kubebuilder/discussions/5277#discussioncomment-15311956


internal/cli/alpha/internal/generate.go
```diff
@@ -25,13 +25,13 @@ import (
 	"strings"
 
 	"sigs.k8s.io/kubebuilder/v4/internal/cli/alpha/internal/common"
+	autoupdatev1alpha "sigs.k8s.io/kubebuilder/v4/internal/plugins/optional/autoupdate/v1alpha"
```
camilamacedo86, line 28: All bollerplates are under internal.
The plugins themselves are not. 

If we move this one, then we would need to change it for all.
I mean, it would need to be horizontal change.

Did you check if the others are in use and you are moving this one only because it is the only one not imported yet?

vitorfloriano: - Grafana and deploy-image have already been imported by other modules (mainly Operator-SDK and some forks), so we either keep it as is OR deprecate them, move them to internal and keep a shim until the next big version.

- Helm hasn't been imported yet(I thought Operator-SDK was importing it but turns out it is not), but there are just too many PRs touching it so I decided to hold this change.

My reasoning for this proposed change is that I feel we should tighten the API boundary as much as possible so we don't have users depending on the plugins packages.

IMO we should only expose the code necessary for building plugins (a core library) and keep the optional plugins in internal. I think that it would make it easier for us to maintain the library documentation if we keep it small, and users wouldn't get overwhelmed or confused about what is meant for them to use in their own projects and what is not.

The alternative to moving the optional plugins to internal is to improve the documentation of all the packages we expose and help users make sense of how they can use them in their own projects. **But I strongly feel we shouldn't do this.** 

The way I see it, plugins are mostly for CLI users, not for projects building their own scaffolders or plugins (though some plugin packages have been imported). It appears to me that we have exposed what should be an internal implementation and we haven't given it the docs necessary.

Anyway, it all boils down to these questions: what do we want to keep as a library? Do we want others depending in our plugins packages? Can we maintain their documentation? For example, do we really want others depending on Alpha Update as a library? What about the Grafana plugin (which has already been imported), do we have the bandwidth or expertise to maintain it as a library?

As a side note, take as an example Operator-SDK, which is much more conservative about its API and puts most of its code in internal.

We can discuss this on a call if you see fit. :)

vitorfloriano: ps: I tried to do the horizontal move in https://github.com/kubernetes-sigs/kubebuilder/pull/5520 but I closed it when I realized it was too big of a scope for a single PR.

camilamacedo86: > Grafana and deploy-image have already been imported by other modules

That is the point. Other tools can import the plugins.
That is why only bollerplates are in internal. ( we want to have freedom to change the bollerplates ) 

Other tools can import and add things on top. 
That was the motivations for the plugin.

That is why I do not think that we should change those. 

See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-1.md

vitorfloriano: I don't think we should maintain the optional plugins as part of the SDK. IMO that is not scalable. The more plugins we offer, the more we need to maintain, and if people start depending on it, the less freedom we'll have to experiment with it and break things. 

Most extensible (if not all) CLI apps decouple the plugins from the core API. 

That is actually aligned with Phase 2 of the plugin architecture:

https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-2.md

> As a Kubebuilder maintainer, I would like to support external plugins not maintained by the core project.
>
> For example, once the Phase 2 plugin implementation is completed, **some internal plugins can be re-implemented as external plugins removing the necessity to build those plugins in the kubebuilder binary**.

From what I understand, plugins are actually meant to be decoupled from the core library and we should move away from offering plugins as Go libraries and start offering them as binaries. They would ideally live in their own repository and "talk" to the main CLI through stdin/stdout. This architectural decoupling via stdin/stdout is what will allow the core to stay lean while the ecosystem grows.

If we keep offering plugins as part of the SDK, the full migration to Phase 2 will be much harder, since we'll have more users depending on plugin packages.

That's why I think it's better to keep as many plugins as possible in internal, to avoid having to support them as libraries because users depend on it.

For example, the Grafana plugin was developed by one contributor (mostly) who is now absent in the project and now other modules depend on it and we need to support it. Moving things to internal prevents this from happening with the next set of plugins.

camilamacedo86: > some internal plugins can be re-implemented as external plugins removing the necessity to build those plugins in the kubebuilder binary.

External plugins are a different topic. The concern is that we do not want Kubebuilder to become a hub that hosts many project-specific plugins.

For example, if someone wants a plugin that integrates their operator with ArgoCD, that should live in the ArgoCD project as an external plugin. We do not know how ArgoCD works and we cannot maintain that integration. If ArgoCD changes, the plugin/scaffolding would also need to change. Keeping it outside respects repository ownership and domain responsibility.

**About plugins**

Plugins that live in this repository should be usable by other tools as well. If we keep a plugin here, we must maintain it and allow other projects to import and use it. Note that the boilerplates we change are not exported. That is something that we cannot export at all. 

Moreover, see that can still introduce breaking changes until the plugin becomes stable: https://book.kubebuilder.io/plugins/plugins-versioning.html and if the plugin is stable. However, if the plugin is stable we should not introduce breaking changes at all for end users. 

For example, there is no reason why Operator SDK or other tools built on top of Kubebuilder should not be able to use the Helm chart plugin. That kind of plugin provides value to any tool using Kubebuilder.

So we should keep a consistent standard: plugins in this repo should be reusable by other tools, not internal to Kubebuilder.

In the AutoUpdate case, I agree it feels a bit unusual because it scaffolds a Kubebuilder-specific feature. Still, keeping things homogeneous across plugins is probably better than having a single internal exception.

> For example, the Grafana plugin was developed by one contributor (mostly) who is now absent in the project and now other modules depend on it and we need to support it.

Grafana plugin is alpha v1. If we think that we need to remove, for example, we can. **However, we cannot remove ANY plugin that is stable not matter if other tool is using or not.**

vitorfloriano:
> Plugins that live in this repository should be usable by other tools as well. If we keep a plugin here, we must maintain it and allow other projects to import and use it. Note that the boilerplates we change are not exported. That is something that we cannot export at all.

Noted. I'm closing this PR as not planned. 

But I still think we need to improve the documentation of the packages we expose. While the Kubebuilder Book is often praised for its quality, the GoDocs have seem to be not worked on for a while. Of the 29 packages we expose, **only two** (`cli` and `go/v4/scaffolds`) have a somewhat good documentation, but still not ideal. 

As an example, compare the [controller-runtime pkg godocs](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.23.3/pkg) (especially its subdirectories) and [kubebuilder pkg godocs](https://pkg.go.dev/sigs.k8s.io/kubebuilder/v4/pkg@v4.13.0). 

IMO, exposing packages just because the repo is a library while also not giving the packages the necessary documentation is bad practice. This negatively affects the developer experience when building with the kubebuilder library.

If we want to foster the community plugin ecosystem, then maybe we should start there.

Thanks for taking the time to discuss this, @camilamacedo86 !

PR #5539, opened 2026-03-13
✨ (go/v4): Add support for multiple controllers per GVK

camilamacedo86:
Change API to allow it via a `controllers[]` list in the PROJECT file, with a new `--controller-name` flag for scaffolding additional controllers while maintaining full backward compatibility with existing `controller: true` behavior.

Closes: https://github.com/kubernetes-sigs/kubebuilder/issues/5473
Closes: https://github.com/kubernetes-sigs/kubebuilder/pull/5474

### DIFF
--- a/internal/cli/alpha/internal/generate.go
+++ b/internal/cli/alpha/internal/generate.go
@@ -207,15 +207,23 @@ func kubebuilderCreate(s store.Store) error {
 		return fmt.Errorf("failed to get resources: %w", err)
 	}
 
-	// First, scaffold all APIs
+	// Scaffold APIs first, as controllers and webhooks depend on them
 	for _, r := range resources {
 		if err = createAPI(r); err != nil {
 			return fmt.Errorf("failed to create API for %s/%s/%s: %w", r.Group, r.Version, r.Kind, err)
 		}
 	}
 
-	// Then, scaffold all webhooks
-	// We cannot create a webhook for an API that does not exist
+	// Scaffold controllers on top of APIs
+	// Multiple controllers can be scaffolded for the same API
+	for _, r := range resources {
+		if err = createControllers(r); err != nil {
+			return fmt.Errorf("failed to create controllers for %s/%s/%s: %w", r.Group, r.Version, r.Kind, err)
+		}
+	}
+
+	// Scaffold webhooks on top of APIs
+	// Webhooks require the API to exist
 	for _, r := range resources {
 		if err = createWebhook(r); err != nil {
 			return fmt.Errorf("failed to create webhook for %s/%s/%s: %w", r.Group, r.Version, r.Kind, err)
@@ -491,9 +499,59 @@ func getDeployImageOptions(resourceData deployimagev1alpha1.ResourceData) []stri
 }
 
 // Creates an API resource.
+// Controllers are scaffolded separately to support multiple controllers per API.
 func createAPI(res resource.Resource) error {
 	args := append([]string{"create", "api"}, getGVKFlags(res)...)
 	args = append(args, getAPIResourceFlags(res)...)
+
+	// Add the external API flags if the resource is external
+	if res.IsExternal() {
+		args = append(args, "--external-api-path", res.Path)
+		args = append(args, "--external-api-domain", res.Domain)
+		// Add module if specified
+		if res.Module != "" {
+			args = append(args, "--external-api-module", res.Module)
+		}
+	}
+
+	if err := util.RunCmd("kubebuilder create api", "kubebuilder", args...); err != nil {
+		return fmt.Errorf("failed to run kubebuilder create api command: %w", err)
+	}
+
+	return nil
+}
+
+// Creates controllers for a resource.
+// This function supports multiple controllers per resource (GVK).
+func createControllers(res resource.Resource) error {
+	if res.Controllers == nil || res.Controllers.IsEmpty() {
+		if !res.Controller {
+			return nil
+		}
+		return createControllerWithName(res, "")
+	}
+
+	for _, controller := range *res.Controllers {
+		if err := createControllerWithName(res, controller.Name); err != nil {
+			return fmt.Errorf("failed to create controller %q: %w", controller.Name, err)
+		}
+	}
+
+	return nil
+}
+
+// Creates a single controller for a resource with a specific name.
+func createControllerWithName(res resource.Resource, controllerName string) error {
+	args := append([]string{"create", "api"}, getGVKFlags(res)...)
+
+	// Always set --resource=false since we're only creating the controller
+	args = append(args, "--resource=false")
+	args = append(args, "--controller=true")
+
+	// Add controller name if specified
+	if controllerName != "" {
+		args = append(args, "--controller-name", controllerName)
+	}
 
 	// Add the external API flags if the resource is external
 	if res.IsExternal() {
@@ -526,11 +584,11 @@ func getAPIResourceFlags(res resource.Resource) []string {
 			args = append(args, "--namespaced=false")
 		}
 	}
-	if res.Controller {
-		args = append(args, "--controller")
-	} else {
-		args = append(args, "--controller=false")
-	}
+
+	// Always disable controller creation in the API scaffolding step
+	// Controllers will be created separately in createControllers()
+	args = append(args, "--controller=false")
+
 	return args
 }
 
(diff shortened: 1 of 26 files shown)


camilamacedo86:
@guilhem could you please help us with the review of this one?

pkg/plugins/golang/v4/api.go
```diff
@@ -139,6 +143,20 @@ func (p *createAPISubcommand) InjectResource(res *resource.Resource) error {
 		p.options.DoController = util.YesNo(reader)
 	}
 
+	// If the resource already exists in the PROJECT file and we're NOT creating an API,
```
v47: The function is getting a bit long and complex.
It can be translated into 3 functions.
Maybe something like this can be an option:
```
func (p *createAPISubcommand) InjectResource(res *resource.Resource) error {
	p.resource = res

	reader := bufio.NewReader(os.Stdin)
	if !p.resourceFlag.Changed {
		log.Info("Create Resource [y/n]")
		p.options.DoAPI = util.YesNo(reader)
	}
	if !p.controllerFlag.Changed {
		log.Info("Create Controller [y/n]")
		p.options.DoController = util.YesNo(reader)
	}

	// If the resource already exists in the PROJECT file and we're NOT creating an API,
	// copy essential fields from the existing resource (like Path) needed for controller scaffolding.
	// Don't copy API, Controllers, or Webhooks - those will be managed by UpdateResource.
	if !p.options.DoAPI {
		if existingRes, err := p.config.GetResource(res.GVK); err == nil {
			p.resource.Path = existingRes.Path
			p.resource.Plural = existingRes.Plural
			p.resource.External = existingRes.External
			p.resource.Core = existingRes.Core
			p.resource.Module = existingRes.Module
		}
	}

	// Ensure that external API options cannot be used when creating an API in the project.
	if p.options.DoAPI &&
		(len(p.options.ExternalAPIPath) != 0 ||
			len(p.options.ExternalAPIDomain) != 0 ||
			len(p.options.ExternalAPIModule) != 0) {
		return errors.New(
			"cannot use '--external-api-path', '--external-api-domain', or '--external-api-module' " +
				"when creating an API in the project with '--resource=true'. " +
				"Use '--resource=false' when referencing an external API",
		)
	}

	// Validate that --external-api-module requires --external-api-path
	if len(p.options.ExternalAPIModule) != 0 && len(p.options.ExternalAPIPath) == 0 {
		return errors.New("'--external-api-module' requires '--external-api-path' to be specified")
	}

	p.options.UpdateResource(p.resource, p.config)

	if err := p.resource.Validate(); err != nil {
		return fmt.Errorf("error validating resource: %w", err)
	}

	if err := p.validateAPI(); err != nil {
		return err
	}

	if err := p.validateController(); err != nil {
		return err
	}

	return nil
}

func (p *createAPISubcommand) validateAPI() error {
	if !p.options.DoAPI {
		return nil
	}

	// Check that resource doesn't have the API scaffolded or flag force was set
	if r, err := p.config.GetResource(p.resource.GVK); err == nil && r.HasAPI() && !p.force {
	}

	// Check that the provided group can be added to the project
	if !p.config.IsMultiGroup() && p.config.ResourcesLength() != 0 && !p.config.HasGroup(p.resource.Group) {
		return fmt.Errorf(
			"multiple groups are not allowed by default, " +
				"to enable multi-group visit https://kubebuilder.io/migration/multi-group.html",
		)
	}

	return nil
}

func (p *createAPISubcommand) validateController() error {
	if !p.options.DoController || p.options.ControllerName == "" {
		return nil
	}

	existingRes, err := p.config.GetResource(p.resource.GVK)
	if err != nil {
		return nil
	}

	controllers := existingRes.Controllers
	if controllers != nil && controllers.HasController(p.options.ControllerName) {
		return fmt.Errorf(
			"controller with name %q already exists for resource %s/%s/%s",
			p.options.ControllerName,
			p.resource.Group,
			p.resource.Version,
			p.resource.Kind,
		)
	}

	if controllers != nil {
		newNormalized := normalizeForCollisionCheck(p.options.ControllerName)
		for _, existing := range *controllers {
			if newNormalized == normalizeForCollisionCheck(existing.Name) {
				return fmt.Errorf(
					"controller name %q conflicts with existing controller %q: both would generate the same reconciler struct name",
					p.options.ControllerName,
					existing.Name,
				)
			}
		}
	}

	// Also check legacy controller: true case
	if existingRes.Controller && p.options.ControllerName == strings.ToLower(p.resource.Kind) {
		return fmt.Errorf(
			"controller with name %q conflicts with the existing legacy controller",
			p.options.ControllerName,
		)
	}

	return nil
}
```

v47:
Should `-controller=true` without `--controller-name` when the resource already uses `controllers` explicitly rejected? Right now, the empty-name path seems reserved for legacy `controller: true`, while multi-controller support is documented as requiring explicit names.

camilamacedo86:
Hi @v47 

Thank you for your help and time to review !!
All addressed !!

Could you please give a look?
I think we are in a good shape to get this one merged now. 
WDYT?

pkg/plugins/golang/v4/api.go
camilamacedo86: DONE

internal/cli/alpha/internal/generate.go
```diff
@@ -207,14 +207,22 @@ func kubebuilderCreate(s store.Store) error {
 		return fmt.Errorf("failed to get resources: %w", err)
 	}
 
-	// First, scaffold all APIs
+	// First, scaffold all APIs (without controllers)
```
v47: Looking at the flow holistically,  I would consider reframing `First, scaffold all APIs (without controllers)`  into something like `Scaffold APIs, before controllers can be scaffolded on top of them`. Same with comments bellow. To describe the relationship APIs -> Controllers -> Webhooks. It's just a thought.

internal/cli/alpha/internal/generate.go
```diff
@@ -512,6 +520,57 @@ func createAPI(res resource.Resource) error {
 	return nil
 }
 
+// Creates controllers for a resource.
+// This function supports multiple controllers per resource (GVK).
+func createControllers(res resource.Resource) error {
```
v47, line 526: Again me being picky, just double negation doesn't feel so natural and extra nesting can be avoided:
```
func createControllers(res resource.Resource) error {
	if res.Controllers == nil || res.Controllers.IsEmpty() {
		if !res.Controller {
			return nil
		}
		return createControllerWithName(res, "")
	}

	for _, controller := range *res.Controllers {
		if err := createControllerWithName(res, controller.Name); err != nil {
			return fmt.Errorf("failed to create controller %q: %w", controller.Name, err)
		}
	}

	return nil
}
```

pkg/plugins/golang/options.go
```diff
@@ -101,7 +107,35 @@ func (opts Options) UpdateResource(res *resource.Resource, c config.Config) {
 	}
 
 	if opts.DoController {
-		res.Controller = true
+		// Use the new Controllers field if a controller name is specified
```
v47: Creating sub-functions will make the flow easier to read, something like this:
```
// updateControllers applies controller-related options to the resource.
func (opts Options) updateControllers(res *resource.Resource) {
	if opts.ControllerName == "" {
		if res.Controllers == nil || res.Controllers.IsEmpty() {
			res.Controller = true
		}
```
A warning can be added here, it can be something like: `resource already uses explicit controllers; ignoring legacy --controller=true without --controller-name.`
```
		return
	}

	if res.Controllers == nil {
		res.Controllers = &resource.Controllers{}
	}

	if res.Controller {
		defaultName := strings.ToLower(res.Kind)
		_ = res.Controllers.AddController(defaultName)
		res.Controller = false
	}

	_ = res.Controllers.AddController(opts.ControllerName)
}
```

internal/cli/alpha/internal/generate.go
camilamacedo86: DOne

camilamacedo86:
Hi @v47 

Thank you for your help for we shape and ensure this one.
Could you please give a look?
Let me know wdyt now?

camilamacedo86:
Thank you for all the help @v47 and your amazing contributions 
For we shape this one !!!

guilhem:
Thank you both @camilamacedo86 and @v47 for your work :)

sorry to have been away

PR #5540, opened 2026-03-14
🐛 fix(config/v3):  Unknown fields in the PROJECT file should be ignored for forward compatibility

camilamacedo86:


camilamacedo86:
Mvoing forward with this one.
If we have external plugins population the PROJECT file we should still able to work with at the same way if we add more fields on it.

PR #5549, opened 2026-03-25
✨ : add `--project-file` CLI flag

### LINKED ISSUE #5548: Add a flag for customizing PROJECT file location
## Feature

Add a `--project-file` flag to allow overriding the hardcoded `PROJECT` path.

## Motivation

Temporarily using modified project configuration, without dirtying the project git repo.

## My use case

Note: I specifically need this just for the helm/v2-alpha plugin, but it seems that the most straightforward way to add this is on the top level.

The description below is basically condensing of the related [slack discussion](https://kubernetes.slack.com/archives/CAR30FCJZ/p1773667718401809).

My project needs to build two flavors of the helm chart (upstream vs. downstream) where the primary difference is the project name (which prefixes most/all resource names). The whole project has a notion of branding all the way to its web UI, so we don't need to use separate git repos for the two flavors.

For the plain YAML installer and OLM-related manifests we achieve the distinction by using [parallel kustomization directories](https://github.com/stackrox/stackrox/tree/master/operator/config/flavors), one is selected at build time [via an env variable](https://github.com/stackrox/stackrox/blob/133160390bb693f9f0929e02fb05e1584285d94f/operator/Makefile#L317).

However the helm/v2-alpha plugin depends on the project name which is hardcoded in the PROJECT file, the config is [loaded here](https://github.com/kubernetes-sigs/kubebuilder/blob/66f313147145b4beb948ec8d639dbb5988014088/pkg/cli/cmd_helpers.go#L420), then [injected here](https://github.com/kubernetes-sigs/kubebuilder/blob/66f313147145b4beb948ec8d639dbb5988014088/pkg/cli/cmd_helpers.go#L453) and [used for example here by the plugin](https://github.com/kubernetes-sigs/kubebuilder/blob/66f313147145b4beb948ec8d639dbb5988014088/pkg/plugins/optional/helm/v2alpha/scaffolds/edit_kustomize.go#L129-L130). Specifically, value for the [chart name](https://github.com/kubernetes-sigs/kubebuilder/blob/66f313147145b4beb948ec8d639dbb5988014088/pkg/plugins/optional/helm/v2alpha/scaffolds/edit_kustomize.go#L130) unconditionally depends on the project name from the config. So it's not possible to customize it using kustomize.

The path to the PROJECT file in turn is the [currently hardcoded](https://github.com/kubernetes-sigs/kubebuilder/blob/66f313147145b4beb948ec8d639dbb5988014088/pkg/config/store/yaml/store.go#L33).

### Option 1

My current workaround is to temporarily `sed -e 's,^projectName: flavor1,projectName: flavor2,' PROJECT`  and restore it afterwards. Ugly, and temporarily dirties the git repo, which has ripple effects, which we need to [work around](https://github.com/stackrox/stackrox/blob/4f8098a1e75264b1f129e146afd4540e76352969/operator/Makefile#L327) using another kludge on top. 🤮 

### Option 2

Create a small, private, special-purpose binary that uses `kubebuilder` as a Go library, that allows for overriding the path. However it looks like I would need a custom version of the `preRunEFunc` helper that allows overriding the path unlike [the upstream one](https://github.com/kubernetes-sigs/kubebuilder/blob/66f313147145b4beb948ec8d639dbb5988014088/pkg/cli/cmd_helpers.go#L420) which loads it from the hardcoded path at the moment. That in turn probably also means a custom `applySubcommandHooks`, all in all duplicating a lot of existing code 😟 

### Option 4

Introduce a `--project-file` flag. Doing it on the root level, instead of only for the `helm/v2-alpha` plugin means that changes can be limited to the `pkg/cli` package.

### Option 5

Introduce a `--project-name` flag in `kubebuilder`. This would meet my specific use case, but:
- it seems the plumbing would be a bit more involved than option 4, and
- less future-proof. I mean: `--project-file` allows killing more (future) birds with one stone, without introducing additional flags in case someone needs to override yet another thing defined in `PROJECT`

### Extra Labels

_No response_

porridge:
A new flag that allows overriding the previously-hardcoded default of `PROJECT`.

Useful for building alternative flavors of the `helm/v2-alpha` charts.

Fixes: #5548

### DIFF
--- a/pkg/cli/cmd_helpers.go
+++ b/pkg/cli/cmd_helpers.go
@@ -157,6 +157,7 @@ func (c *CLI) applySubcommandHooks(
 		store:               yamlstore.New(c.fs),
 		subcommands:         subcommands,
 		errorMessage:        errorMessage,
+		projectFile:         c.projectFile,
 		projectVersion:      c.projectVersion,
 		pluginChain:         pluginChain,
 		cliVersion:          c.cliVersion,
@@ -296,6 +297,8 @@ type executionHooksFactory struct {
 	subcommands []keySubcommandTuple
 	// errorMessage is prepended to returned errors.
 	errorMessage string
+	// projectFile is the path to the project configuration file (empty means use default: PROJECT).
+	projectFile string
 	// projectVersion is the project version that will be used to create new project configurations.
 	// It is only used for initialization.
 	projectVersion config.Version
@@ -407,7 +410,13 @@ func (factory *executionHooksFactory) preRunEFunc(
 		}
 		if createConfig {
 			// Check if a project configuration is already present.
-			if err := factory.store.Load(); err == nil || !errors.Is(err, os.ErrNotExist) {
+			var err error
+			if factory.projectFile != "" {
+				err = factory.store.LoadFrom(factory.projectFile)
+			} else {
+				err = factory.store.Load()
+			}
+			if err == nil || !errors.Is(err, os.ErrNotExist) {
 				return fmt.Errorf("%s: already initialized", factory.errorMessage)
 			}
 
@@ -417,7 +426,13 @@ func (factory *executionHooksFactory) preRunEFunc(
 			}
 		} else {
 			// Load the project configuration.
-			if err := factory.store.Load(); os.IsNotExist(err) {
+			var err error
+			if factory.projectFile != "" {
+				err = factory.store.LoadFrom(factory.projectFile)
+			} else {
+				err = factory.store.Load()
+			}
+			if os.IsNotExist(err) {
 				return fmt.Errorf("%s: failed to find configuration file, project must be initialized",
 					factory.errorMessage)
 			} else if err != nil {
@@ -507,7 +522,13 @@ func (factory *executionHooksFactory) runEFunc() func(*cobra.Command, []string) 
 // and executes the post-scaffold hook.
 func (factory *executionHooksFactory) postRunEFunc() func(*cobra.Command, []string) error {
 	return func(*cobra.Command, []string) error {
-		if err := factory.store.Save(); err != nil {
+		var err error
+		if factory.projectFile != "" {
+			err = factory.store.SaveTo(factory.projectFile)
+		} else {
+			err = factory.store.Save()
+		}
+		if err != nil {
 			return fmt.Errorf("%s: failed to save configuration file: %w", factory.errorMessage, err)
 		}
 
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -40,6 +40,7 @@ const (
 
 	pluginsFlag        = "plugins"
 	projectVersionFlag = "project-version"
+	projectFileFlag    = "project-file"
 )
 
 // CLI is the command line utility that is used to scaffold kubebuilder project files.
@@ -73,6 +74,8 @@ type CLI struct {
 	pluginKeys []string
 	// Project version to scaffold.
 	projectVersion config.Version
+	// Project file path to use instead of default PROJECT file.
+	projectFile string
 
 	// A filtered set of plugins that should be used by command constructors.
 	resolvedPlugins []plugin.Plugin
@@ -211,6 +214,12 @@ func (c *CLI) getInfoFromConfigFile() error {
 	// Read the project configuration file
 	cfg := yamlstore.New(c.fs)
 
+	// Determine which project file to use.
+	projectPath := yamlstore.DefaultPath
+	if c.projectFile != "" {
+		projectPath = c.projectFile
+	}
+
 	// Workaround for https://github.com/kubernetes-sigs/kubebuilder/issues/4433
 	//
 	// This allows the `kubebuilder alpha generate` command to work with old projects
@@ -222,12 +231,12 @@ func (c *CLI) getInfoFromConfigFile() error {
 
 	if isAlphaGenerateCommand(os.Args[1:]) {
 		// Patch raw file bytes before unmarshalling
-		if err := patchProjectFileInMemoryIfNeeded(c.fs.FS, yamlstore.DefaultPath); err != nil {
+		if err := patchProjectFileInMemoryIfNeeded(c.fs.FS, projectPath); err != nil {
 			return err
 		}
 	}
 
-	if err := cfg.Load(); err != nil {
+	if err := cfg.LoadFrom(projectPath); err != nil {
 		return fmt.Errorf("error loading configuration: %w", err)
 	}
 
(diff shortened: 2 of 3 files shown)


pkg/cli/cmd_helpers.go
```diff
@@ -407,7 +410,13 @@ func (factory *executionHooksFactory) preRunEFunc(
...
 		if createConfig {
 			// Check if a project configuration is already present.
-			if err := factory.store.Load(); err == nil || !errors.Is(err, os.ErrNotExist) {
+			var err error
+			if factory.projectFile != "" {
+				err = factory.store.LoadFrom(factory.projectFile)
+			} else {
+				err = factory.store.Load()
+			}
+			if err == nil || !errors.Is(err, os.ErrNotExist) {
```
porridge, line 419: This and the following two hunks are making things somewhat verbose 😟 
The best way I could come up with to avoid this verbosity and keep things in a single line, would be to treat the empty project file path as a sentinel value meaning "use the default". I.e. `store.LoadFrom("")` would be equivalent to `store.Load()`. Please let me know if I should do that.

porridge (commented):
Please let me know if this looks sensible, and I can look into adding a test or two.

pkg/cli/cmd_helpers.go
v47: No nesting and early return is usally peferable in golang. 
This flow for example has less nesting and duplications:
```
...
var err error
if factory.projectFile != "" {
	err = factory.store.LoadFrom(factory.projectFile)
} else {
	err = factory.store.Load()
}

if createConfig {
	// Check if a project configuration is already present.
	if err == nil || !errors.Is(err, os.ErrNotExist) {
		return fmt.Errorf("%s: already initialized", factory.errorMessage)
	}

	// Initialize the project configuration.
	if err := factory.store.New(factory.projectVersion); err != nil {
		return fmt.Errorf("%s: error initializing project configuration: %w", factory.errorMessage, err)
	}

	return nil
}

if os.IsNotExist(err) {
	return fmt.Errorf(
		"%s: failed to find configuration file, project must be initialized",
		factory.errorMessage,
	)
}
if err != nil {
	return fmt.Errorf("%s: failed to load configuration file: %w", factory.errorMessage, err)
}
....
```

vitorfloriano:
> Please let me know if this looks sensible, and I can look into adding a test or two.

Hi @porridge

Although the issue described in #5548 touches a bit on the use cases, I think it would be nice if you could add documentation for this flag in this PR as well. That would make it easier to reason if the flag is sensible or not, and in which cases it could be useful, what problems it could solve, what value it could bring to users, what possibilities it could open.

There is no specific reference section in the book for the cli subcommands and flags (as far as I know), but I think it could be addressed in the FAQ.

Once @camilamacedo86 agrees on adding this flag, I think there could be a couple of input validation checks with helpful error messages to guide users. For example, what should happen when the user passes a file that doesn't exist? I gave it a try and it renamed the PROJECT file to the value that was passed, which could be a bit problematic, since other features rely on the existence of a file named PROJECT:

```
@vitorfloriano ➜ /workspaces/test $ kubebuilder init --project-file nonexistent
INFO Writing kustomize manifests for you to edit... 
INFO Writing scaffold for you to edit... 
INFO Get controller runtime 
INFO Update dependencies 
Next: define a resource with:
$ kubebuilder create api

@vitorfloriano ➜ /workspaces/test $ ls
AGENTS.md  Dockerfile  Makefile  README.md  cmd  config  go.mod  go.sum  hack  nonexistent  test
                                                                               ^^^^^^^^^^^

@vitorfloriano ➜ /workspaces/test $ cat nonexistent 
cliVersion: 4.0.0-20260325080735-02a527b2819a
domain: my.domain
layout:
- go.kubebuilder.io/v4
projectName: test
repo: testtyyy
version: "3"
```

camilamacedo86:
Hi @porridge, @vitorfloriano 

In `alpha generate` we currently have the flags `--input-dir` and `--output-dir`.

I noticed that we already define the output directory here:
https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/optional/helm/v2alpha/edit.go#L73-L74

Given that, could we rely on `--input-dir` as the source of truth for the path used to generate the project? And make it work at the same way of alpha generate? 

This would make things more flexible—if we introduce additional files in the future, they could simply be resolved relative to `--input-dir`.

WDYT?

Additionally, in `kubebuilder init` we have a `--project-name` flag. We might consider adding a similar flag here as well.

WDYT?

camilamacedo86:
Hi @porridge 

I added you as co-author of: https://github.com/kubernetes-sigs/kubebuilder/pull/5573 where I continue your effort here.
- This one is consistency with the current behaviour.
- Will allow people generate the Helm Chart in a monorepo 
- And solves your scenario.

Could you please check out?

porridge:
Thanks for looking into this @vitorfloriano 

> Although the issue described in #5548 touches a bit on the use cases, I think it would be nice if you could add documentation for this flag in this PR as well. That would make it easier to reason if the flag is sensible or not, and in which cases it could be useful, what problems it could solve, what value it could bring to users, what possibilities it could open.
> 
> There is no specific reference section in the book for the cli subcommands and flags (as far as I know), but I think it could be addressed in the FAQ.

My only _direct_ use of `kubebuilder` is the `helm/v2-alpha` plugin, so I'm not sure what other use cases might find it useful. I can only say something generic like "temporary override/customization of the PROJECT file location/content". Like if you need to run a (sub)command with slightly different configuration than in your checked-in PROJECT file, this flag lets you do it. I think it's one of those features that "if you don't know whether you need it, then it means you don't need it; but if you need it, you'll love it".

> Once @camilamacedo86 agrees on adding this flag, I think there could be a couple of input validation checks with helpful error messages to guide users. For example, what should happen when the user passes a file that doesn't exist?

I guess it depends on whether the given command wants to read or write the file (or both)?

> I gave it a try and it renamed the PROJECT file to the value that was passed, which could be a bit problematic, since other features rely on the existence of a file named PROJECT

Do you mean that:
1. there are features that hardcode the `./PROJECT` path (and would not be affected by this flag)` or
2. that all features **by default** expect the `./PROJECT` path and therefore if you supplied a different path to command A, then you might need to keep providing it to subsequent commands for them to be able to see changes made by command A?

If 1, then please kindly point me at them, they should also be changed in this PR.

If 2, then it's expected. Think of it as of setting `$GIT_CONFIG_GLOBAL` when invoking `git`. If you set it when invoking `git config --global foo.bar baz` but then **don't** set it on a subsequent `git config get foo` command, then it's probably **not** going to show you `baz`.

Thinking about this now, maybe this would be more appropriate to implement it in terms of an environment variable rather than a global flag? 🤔

PR #5556, opened 2026-03-27
🌱 infra: Add Pinact GitHub Actions workflow

### LINKED ISSUE #5552: Following Kubernetes Policies for GitHub Actions
## Context

The Kubernetes project has adopted a new [GitHub Actions Security Policy](https://github.com/kubernetes/community/blob/main/github-management/github-actions-policy.md) [1] to strengthen our supply chain security across all repositories under Kubernetes GitHub organizations.
This policy is created in response to recent security incidents, including the [Trivy GitHub Actions vulnerability](https://github.com/aquasecurity/trivy/security/advisories/GHSA-69fq-xp46-6x23) [2].

As per the new policy, all projects under Kubernetes GitHub organizations using GitHub Actions must reference actions in workflow files using full-length commit SHA hashes instead of mutable references such as `latest`, tags, branches (like `master`, `main`).
This prevents potential supply chain attacks where compromised actions could inject malicious code through force-updated references.

Immediate Action Required:
Project maintainers are requested to update their GitHub Actions workflows to comply with this policy by April 15, 2026. 
After this date, the Kubernetes project will enforce the "Require actions to be pinned to a full-length commit SHA" policy at the enterprise level, and any github action workflows using mutable references will fail to run.

Key Requirements:
- Pin all actions to 40-character commit SHA hashes (e.g., `uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1`)
- New workflows must comply before merge
- We recommend enabling Dependabot for GitHub Actions to automatically keep SHA-pinned actions up to date

If you have questions, please reach out on the #github-management [3] slack channel.

## What we need to do:

- Check if we should not either adopt the policy
- Check if we should not change the GtiHub actions for Kubebuilder itssef as for the scaffolds produced by the tool. 

### Extra Labels

_No response_

vitorfloriano:
This PR adds a workflow that uses [Pinact](https://github.com/suzuki-shunsuke/pinact) to ensure all PRs that touch actions follow the recent GitHub Actions Security Policy (see kubernetes/community#8911).

Fixes #5552

### DIFF
--- a/.github/workflows/pinact.yml
+++ b/.github/workflows/pinact.yml
@@ -1,0 +1,32 @@
+name: Pinact
+on:
+  pull_request:
+    paths: 
+     - '**/workflows/*.yml'
+jobs:
+  pinact:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+          
+      - name: Set up Go
+        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
+        with:
+          go-version: '1.26'
+
+      - name: Install pinact
+        run: |
+          go install github.com/suzuki-shunsuke/pinact/v3/cmd/pinact@v3.5.0
+
+      - name: Display changed workflows in PR
+        run: |
+          git diff --name-only origin/${{ github.base_ref }} HEAD
+          
+      - name: Run pinact on changed workflow files
+        run: |
+          git diff --name-only origin/${{ github.base_ref }} HEAD \
+          | grep -E 'workflows/.*\.ya?ml$' \
+          | xargs pinact run --diff --check --verify              


.github/workflows/pinact.yml
```diff
@@ -0,0 +1,32 @@
...
+          skip_push: "true"
+          includes: |
+            .github/workflows/*.yml
+            ./docs/book/src/cronjob-tutorial/testdata/project/.github/workflows/*.yml
+            ./docs/book/src/multiversion-tutorial/testdata/project/.github/workflows/*.yml
+            ./docs/book/src/getting-started/testdata/project/.github/workflows/*.yml
+            ./pkg/plugins/golang/v4/scaffolds/internal/templates/github/*.go
+            ./pkg/plugins/optional/autoupdate/v1alpha/scaffolds/internal/github/*.go
+            ./pkg/plugins/optional/helm/*/scaffolds/internal/templates/github/*.go
+            ./testdata/*/.github/workflows/*.yml
```
camilamacedo86: Why it does not fail? 
Should not it fail since the current source code does not pinned the SHA?

.github/workflows/pinact.yml
```diff
@@ -0,0 +1,32 @@
+name: Pinact
```
camilamacedo86, line 1: Since this PR only updates a GitHub Action in our repository and does not affect the projects generated by the tool, it should not be highlighted in the release notes outside of the full changelog. Therefore, the appropriate emoji is 🌱.

vitorfloriano: Got it. I'll change the commit title as well.

vitorfloriano: Done!

.github/workflows/pinact.yml
vitorfloriano: Yeah, it definitely should've failed. I'll look into it.

vitorfloriano:
@camilamacedo86 I pivoted the workflow to use the `pinact` binary directly instead of the pre-made `pinact-action`. 

Then I unpinned an action and changed the version comment on purpose to test and it seems `pinact` did its job. 

See: https://github.com/kubernetes-sigs/kubebuilder/actions/runs/23700018027/job/69041847860?pr=5556

PR #5558, opened 2026-03-28
🐛 fix(go/v4): disable kubectl kuberc in the e2e tests by default

sanadhis:
To make kubectl executions consistent when running e2e across differences workstations, we should disable kuberc by default by setting `KUBECTL_KUBERC=false`

### Context:
With the introduction of kuberc https://kubernetes.io/docs/reference/kubectl/kuberc/, kubectl commands can now behave non-universally.

### DIFF
--- a/docs/book/src/multiversion-tutorial/testdata/project/Makefile
+++ b/docs/book/src/multiversion-tutorial/testdata/project/Makefile
@@ -69,6 +69,8 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
 
 # TODO(user): To use a different vendor for e2e tests, modify the setup under 'tests/e2e'.
 # The default setup assumes Kind is pre-installed and builds/loads the Manager Docker image locally.
+# kubectl kuberc is disabled by default for test isolation; enable with:
+# - KUBECTL_KUBERC=true
 # CertManager is installed by default; skip with:
 # - CERT_MANAGER_INSTALL_SKIP=true
 KIND_CLUSTER ?= project-test-e2e
--- a/docs/book/src/cronjob-tutorial/testdata/project/Makefile
+++ b/docs/book/src/cronjob-tutorial/testdata/project/Makefile
@@ -69,6 +69,8 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
 
 # TODO(user): To use a different vendor for e2e tests, modify the setup under 'tests/e2e'.
 # The default setup assumes Kind is pre-installed and builds/loads the Manager Docker image locally.
+# kubectl kuberc is disabled by default for test isolation; enable with:
+# - KUBECTL_KUBERC=true
 # CertManager is installed by default; skip with:
 # - CERT_MANAGER_INSTALL_SKIP=true
 KIND_CLUSTER ?= project-test-e2e
--- a/docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_suite_test.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/test/e2e/e2e_suite_test.go
@@ -43,6 +43,8 @@ var (
 // TestE2E runs the e2e test suite to validate the solution in an isolated environment.
 // The default setup requires Kind and CertManager.
 //
+// To enable kubectl kuberc (use custom kubectl configurations), set: KUBECTL_KUBERC=true
+// By default, kuberc is disabled to ensure consistent test behavior across different environments.
 // To skip CertManager installation, set: CERT_MANAGER_INSTALL_SKIP=true
 func TestE2E(t *testing.T) {
 	RegisterFailHandler(Fail)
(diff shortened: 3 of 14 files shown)


sanadhis:
To add more context, with example use case

Let's say we perform logs checking `kubectl logs <pod>` as part of e2e test.
e2e pass successfully in CI but because of a kuberc in local workstation that enables follow as default:
```
apiVersion: kubectl.config.k8s.io/v1beta1
kind: Preference
defaults:
- command: logs
  options:
    - name: follow
      default: "true"
```

This will lead to timeout when running e2e. 
The goal of disabling this is to make sure kubectl execution in e2e to be more consistent

camilamacedo86:
Hi @sanadhis,

Thanks a lot for raising this — I took a look and I completely agree, this is something we should include 👍|

I’d suggest applying the changes as outlined below (mainly around making kubectl kuberc explicitly configurable while keeping tests isolated by default).

After that, could you please run:

```
make install
make generate
```

so that all samples under testdata and the docs are properly updated?

Once that’s done, please address the review comments so we can get your contribution merged 🚀

Thanks again for the great work!

### Code suggestions 

```suggestion
Update `pkg/plugins/golang/v4/scaffolds/internal/templates/test/e2e/suite.go`:

// The default setup requires Kind and CertManager.
//
// To skip CertManager installation, set: CERT_MANAGER_INSTALL_SKIP=true
// To enable kubectl kuberc (use custom kubectl configurations), set: KUBECTL_KUBERC=true
// By default, kuberc is disabled to ensure consistent test behavior across different environments.
func TestE2E(t *testing.T) {
    RegisterFailHandler(Fail)
    _, _ = fmt.Fprintf(GinkgoWriter, "Starting {{ .ProjectName }} e2e test suite\n")
}

...

// Disable kubectl kuberc by default for test isolation.
// This prevents local kubectl configurations from affecting test behavior.
// To enable kuberc, set: KUBECTL_KUBERC=true
if os.Getenv("KUBECTL_KUBERC") != "true" {
    By("disabling kubectl kuberc for test isolation")
    err = os.Setenv("KUBECTL_KUBERC", "false")
    ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Failed to disable kubectl kuberc")
    _, _ = fmt.Fprintf(GinkgoWriter, "kubectl kuberc disabled for consistent test behavior (override with KUBECTL_KUBERC=true)\n")
} else {
    _, _ = fmt.Fprintf(GinkgoWriter, "kubectl kuberc enabled (KUBECTL_KUBERC=true)\n")
}

Update `pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go`:

# The default setup assumes Kind is pre-installed and builds/loads the Manager Docker image locally.
# CertManager is installed by default; skip with:
# - CERT_MANAGER_INSTALL_SKIP=true
# kubectl kuberc is disabled by default for test isolation; enable with:
# - KUBECTL_KUBERC=true
KIND_CLUSTER ?= {{ .ProjectName }}-test-e2e

test-e2e: setup-test-e2e manifests generate fmt vet ## Run the e2e tests in an isolated environment. Use KUBECTL_KUBERC=true to enable custom kubectl configs.
    KIND=$(KIND) KIND_CLUSTER=$(KIND_CLUSTER) go test -tags=e2e ./test/e2e/ -v -ginkgo.v
    $(MAKE) cleanup-test-e2e
```

sanadhis:
@camilamacedo86 thank you for your suggestion. I have applied the changes and run `make install` and `make generate`

Note that I re-order the kuberc-related comments and set up before the cert manager one because this should take precedence over `setupCertManager` which beneath use `kubectl`:
```
func InstallCertManager() error {
	url := fmt.Sprintf(certmanagerURLTmpl, certmanagerVersion)
	cmd := exec.Command("kubectl", "apply", "-f", url)
```

Let me know if it does make sense. Thanks again for your support 👍

docs/book/src/multiversion-tutorial/testdata/project/Makefile
```diff
@@ -88,7 +88,7 @@ setup-test-e2e: ## Set up a Kind cluster for e2e tests if it does not exist
 	esac
 
 .PHONY: test-e2e
-test-e2e: setup-test-e2e manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
+test-e2e: setup-test-e2e manifests generate fmt vet ## Run the e2e tests in an isolated environment using Kind. Use KUBECTL_KUBERC=true to enable custom kubectl configs.
```
camilamacedo86: I am wondering we we should really change here.
WDYT about just keep as it is?
It seems that if someone want to enable ( very unlikely ) still possible to check the option/info in the code itself. WDYT?

sanadhis: makes sense. reverted this part 👍

camilamacedo86:
@sanadhis it is failing in the linter check.
Could you please run `make lint-fix`?

sanadhis:
> @sanadhis it is failing in the linter check. Could you please run `make lint-fix`?

Sorry I missed this one. Fixed and re-pushed (I amend my commit message at the same time).

PR #5561, opened 2026-03-29
📖 docs(helm/v1alpha): Add deprecation notice

vitorfloriano:
This PR adds deprecation notices to the `helm/v1alpha` and `helm/v1alpha/scaffolds` packages to warn users about importing them.

Users are directed to use `helm/v2-alpha` packages instead.

This is necessary to get warns from IDEs and linters, and [pkg.go.dev](https://pkg.go.dev/) will hide their docs by default.

Relates to: https://github.com/kubernetes-sigs/kubebuilder/discussions/5277#discussioncomment-15400518

### DIFF
--- a/pkg/plugins/optional/helm/v1alpha/scaffolds/edit.go
+++ b/pkg/plugins/optional/helm/v1alpha/scaffolds/edit.go
@@ -14,6 +14,18 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Package scaffolds implements the scaffolding engine for the Helm v1-alpha plugin.
+//
+// Deprecated: The helm/v1alpha/scaffolds package is deprecated. Use [helm/v2alpha/scaffolds] instead.
+// The new helm/v2-alpha plugin replaces the deprecated helm/v1-alpha and brings major improvements in
+// flexibility and maintainability, championing changes driven by community feedback. Chart values are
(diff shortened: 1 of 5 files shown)


pkg/plugins/optional/helm/v1alpha/scaffolds/edit.go
```diff
@@ -14,6 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Package scaffolds implements the scaffolding engine for the Helm v1-alpha plugin.
+//
+// Deprecated: The helm/v1-alpha plugin is deprecated. Use [helm/v2alpha/scaffolds] instead.
+//
+// This package is kept for compatibility and no new functionality will be added.
```
camilamacedo86: That is a good catcher 🎉 

We will remove that eventually.
this one we do not need to wait for 5x since it is v1-alpha
What we need is ensure that people migrate to v2-alpha 
Then, we can remove. 

Also, the `// Deprecated:` must be the first thing for we properly deprecate that.
Could you please check the plugin deprecation notice and use something similar here?
See: https://github.com/kubernetes-sigs/kubebuilder/releases/tag/v4.10.0

vitorfloriano: @camilamacedo86 I added the deprecation notice you pointed me to. PTAL.

Issue #5544, opened 2026-03-16
(helm/v2-alpha plugin) Allow customizing pod spec fields (e.g. `priorityClassName`)


camilamacedo86:
## Problem

The Helm chart generated by the `helm/v2-alpha` plugin exposes some pod configuration through the `values.manager.<value>` pattern. However, certain pod spec fields such as `priorityClassName` are not currently configurable.

This makes it harder for users to customize scheduling or other pod settings without modifying the generated chart templates.

## Proposal

We could consider one of the following approaches:

- Add support for specific fields such as `priorityClassName`.
- Introduce a more flexible option such as `extraPodSpec` that allows users to extend the pod spec with additional fields not explicitly exposed in the chart values (similar to patterns used in other Helm charts).
- Explore dynamically exposing options only when they exist in the kustomize manifests to avoid divergence.

The goal would be to keep the Helm chart aligned with the kustomize manifests, while still allowing users to customize pod settings when needed.

## Additional Context

Suggestion originally raised by @Liam Mackie in Slack: https://kubernetes.slack.com/archives/CAR30FCJZ/p1773631814394009

cmallikarjunah:
Hi @camilamacedo86 @liam-mackie 
I was thinking how the extra (but frequently used) pod spec could be exposed in helm charts given that it should align with kustomize manifest. This is how I understand it could work in the simplest way (Pls do correct me if my understanding is not correct. I'm just starting to deep dive into the kubebuilder code):

1. To configure fields like `priorityClassName` in a deployment, additional resource need to be created: `PriorityClass` resource. User applies customizations (like addition of any extra Pod Spec and new resources as required) via Kustomize patches in the config/ directory in the Kubebuilder scaffolded project. 
2. The user attempts to generate the helm charts using:
`kubebuilder edit --plugins=helm/v2-alpha
3. The kustomize output `dist/install.yaml` file is generated based on the project's config/ including the custom patches.
4. The kubebuilder kustomize output to helm chart converter code should now support the new fields i.e; identify these new fields and reflect in helm charts accordingly.

camilamacedo86:
I think we can map all that the Pod, Deployment can have that would make sense to expose in the helm values, and then only add to the helm values if found in the yaml. That seems a good strategy.

cmallikarjunah:
@liam-mackie Are you still working on this? If not I would like to pick it up and try it out. Thanks!

liam-mackie:
[...]

PR #5566, opened 2026-03-31
:sparkles: [WiP] helm/valpha2: Add extra pod spec

cmallikarjunah:
The PR adds support for extra pod spec in helm plugin. 
Issue: https://github.com/kubernetes-sigs/kubebuilder/issues/5544

**TODO:**
1. Add placements for extra pod spec with "kubebuilder edit --plugins-helm/v2-alpha" - Done
2. Additionally should we also add support for extra pod spec when converting chart from kustomize output if the fields are present already?
3. Check for additional requirements
4. Generate testdata/ samples
5. Add documentation

(Implementation is based on my reference to https://github.com/kubernetes-sigs/kubebuilder/pull/5496/changes and AI code assistants)

### DIFF
--- a/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
+++ b/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
@@ -558,6 +558,13 @@ func (t *HelmTemplater) substituteCertManagerAnnotations(yamlContent string) str
 func (t *HelmTemplater) templateDeploymentFields(yamlContent string) string {
 	// Template replicas from values.yaml (manager.replicas)
 	yamlContent = t.templateReplicas(yamlContent)
+	// Template strategy at Deployment spec level
+	yamlContent = t.templateBasicWithStatement(
+		yamlContent,
+		"strategy",
+		"spec",
+		".Values.manager.strategy",
+	)
 	// Template configuration fields
 	yamlContent = t.templateImageReference(yamlContent)
 	yamlContent = t.templateEnvironmentVariables(yamlContent)
@@ -587,6 +594,30 @@ func (t *HelmTemplater) templateDeploymentFields(yamlContent string) string {
 		"spec.template.spec",
 		".Values.manager.tolerations",
 	)
+
+	// Template additional pod spec fields
+	podSpecFields := []string{
+		"priorityClassName",
+		"initContainers",
+		"ephemeralContainers",
+		"hostname",
+		"dnsConfig",
+		"dnsPolicy",
+		"hostIPC",
+		"hostNetwork",
+		"restartPolicy",
+		"nodeName",
+		"topologySpreadConstraints",
+		"schedulerName",
+	}
+	for _, field := range podSpecFields {
+		yamlContent = t.templateBasicWithStatement(
+			yamlContent,
+			field,
+			"spec.template.spec",
+			".Values.manager."+field,
+		)
+	}
 
 	return yamlContent
 }
--- a/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/values_basic_test.go
+++ b/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/values_basic_test.go
@@ -172,6 +172,9 @@ var _ = Describe("HelmValuesBasic", func() {
 			Expect(content).To(ContainSubstring("affinity: {}"))
 			Expect(content).To(ContainSubstring("nodeSelector: {}"))
 			Expect(content).To(ContainSubstring("tolerations: []"))
+			Expect(content).To(ContainSubstring("# priorityClassName:"))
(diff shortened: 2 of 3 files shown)


pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/values_basic_test.go
```diff
@@ -172,6 +172,9 @@ var _ = Describe("HelmValuesBasic", func() {
 			Expect(content).To(ContainSubstring("affinity: {}"))
 			Expect(content).To(ContainSubstring("nodeSelector: {}"))
 			Expect(content).To(ContainSubstring("tolerations: []"))
+			Expect(content).To(ContainSubstring("# priorityClassName:"))
+			Expect(content).To(ContainSubstring("# initContainers:"))
+			Expect(content).To(ContainSubstring("# ephemeralContainers:"))
 		})
 	})
```
camilamacedo86, line 180: Hi @cmallikarjunah 

See wdyt: https://github.com/kubernetes-sigs/kubebuilder/pull/5577
I think we can move with the above. Could you please help us out in the review?

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/values_basic_test.go
```diff
@@ -172,6 +172,9 @@ var _ = Describe("HelmValuesBasic", func() {
 			Expect(content).To(ContainSubstring("affinity: {}"))
 			Expect(content).To(ContainSubstring("nodeSelector: {}"))
 			Expect(content).To(ContainSubstring("tolerations: []"))
+			Expect(content).To(ContainSubstring("# priorityClassName:"))
+			Expect(content).To(ContainSubstring("# initContainers:"))
```
camilamacedo86, line 176: `initContainers` change how the operator starts and runs, it either depends on the internal code implementation of the Operator. I do not think that we should expose it for those who a consumer the solution. What would be the use case scenario?

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/values_basic_test.go
```diff
@@ -172,6 +172,9 @@ var _ = Describe("HelmValuesBasic", func() {
...
+			Expect(content).To(ContainSubstring("# ephemeralContainers:"))
```
camilamacedo86, line 177: `ephemeralContainers` are meant for debugging, not for normal application runtime.

They are used to inspect running containers, not to define how the operator works. I do not think that we should expose it in the Helm values. what would be the use case?

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
```diff
@@ -588,6 +595,30 @@ func (t *HelmTemplater) templateDeploymentFields(yamlContent string) string {
...
+		"ephemeralContainers",
+		"hostname",
+		"dnsConfig",
+		"dnsPolicy",
+		"hostIPC",
+		"hostNetwork",
+		"restartPolicy",
+		"nodeName",
+		"topologySpreadConstraints",
+		"schedulerName",
```
camilamacedo86, line 611: I took some time today to go through this in detail and added a comment here:
https://github.com/kubernetes-sigs/kubebuilder/issues/5544#issuecomment-4179876301

I tried to map all the fields and, IMO, we should not expose everything listed. We should be more strict about what is part of the chart contract vs what is internal/operator-level behavior.

I do not think that we should expose:

- `initContainers` 
- `ephemeralContainers`  
- `hostname`  
- `dnsConfig`  
- `dnsPolicy`  
- `hostIPC`  
- `hostNetwork`  
- `restartPolicy`  
- `nodeName`  
- `schedulerName`  

Mainly, when would it really make sense to allow the end user or consumer of an operator packaged with Helm to change any of the following fields? Can we think through that first? If we find real and reasonable scenarios where a field should be configurable by the chart consumer, then we can add those specific fields to be added to the values when/only the kustomize have those. 

However, I agree that we should expose: 

- `priorityClassName`
- `topologySpreadConstraints`  

The Helm chart should expose only what users can safely configure without modifying the operator itself.

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/values_basic_test.go
cmallikarjunah: Oh okay, got it. 
Yes I think it doesn't make sense to expose initContainers.

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/values_basic_test.go
cmallikarjunah: Sure @camilamacedo86. Thanks for the PR and the detailed explanations, I am looking into it.

Issue #5562, opened 2026-03-30
Add explicit permissions to GitHub Actions workflow


camilamacedo86:
### What do you want to happen?

This task is for we add explicit permissions to GitHub Actions workflows for our actions and those that are expose, scaffold to end users in the CLI.

Therefore, we expected here 2 PRs. 

The goal is to address alerts such as :

<img width="1015" height="500" alt="Image" src="https://github.com/user-attachments/assets/b626e885-aaa7-44ce-be13-6c4bc983b908" />

camilamacedo86:
@vitorfloriano I am pinging you here since you were looking on similar things, you might want to work on this one.

camilamacedo86:
@vitorfloriano what we need to here seems to be : https://github.com/step-security/secure-repo?tab=readme-ov-file#before-and-after-the-fix

camilamacedo86:
Hi @vitorfloriano the alerts now are solved for Kubabuilder repo it self.
Therefore, we can apply all our learns to all GitHub actions scaffolded to end users. (CLI scaffolds) 
WDYT?

PR #5565, opened 2026-03-31
🌱 infra(CI): Add gha linter (zizmor) and fix workflows security issues

vitorfloriano:
This PR fixes security issues in GitHub Actions workflows that run in Kubebuilder CI. It also adds a specialized linter ([zizmor](https://github.com/zizmorcore/zizmor)) in a dedicated job in CI.

More info on why the fixes were necessary can be found in zizmor docs:

- [artipacked](https://docs.zizmor.sh/audits/#artipacked)
- [cache-poisoning](https://docs.zizmor.sh/audits/#cache-poisoning)
- [dependabot-cooldown](https://docs.zizmor.sh/audits/#dependabot-cooldown)
- [excessive-permissions](https://docs.zizmor.sh/audits/#excessive-permissions)
- [template-injection](https://docs.zizmor.sh/audits/#template-injection)

Addresses: #5562

PR #5570, opened 2026-04-01
🌱 infra(CI): Add zero-trust top-level permissions in workflows

vitorfloriano:
This PR sets the top-level permission in all Kubebuilder CI workflows to the most restrictive, zero-trust, `permissions: {}`. This remediates all the token permission vulnerabilities found by the scorecard workflow.

Now, by default, no jobs are allowed to do anything, unless explicitly permitted in job-level permissions. 

This appears to be the _crème de la crème_ in security practices.

> [!NOTE]
> We could have set them to `permissions: contents: read`, but that would be less secure than `permissions: {}`, considering that it would still grant some privileges that could be explored in attacks. 

Adresses: #5562

### DIFF
--- a/.github/workflows/apidiff.yml
+++ b/.github/workflows/apidiff.yml
@@ -8,6 +8,8 @@ on:
     paths-ignore:
       - '**/*.md'
 
+permissions: {}
+
 jobs:
   go-apidiff:
     permissions:
--- a/.github/workflows/docs-accessibility.yml
+++ b/.github/workflows/docs-accessibility.yml
@@ -8,8 +8,12 @@ on:
     paths:
       - '**/*.md'
 
+permissions: {}
+
 jobs:
   lint:
+    permissions:
+      contents: read
     name: "Check Docs Accessibility"
     runs-on: ubuntu-latest
     # Pull requests from the same repository won't trigger this checks as they were already triggered by the push
@@ -17,5 +21,7 @@ jobs:
     steps:
       - name: Clone the code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
       - name: Run check
         run: make test-docs-accessibility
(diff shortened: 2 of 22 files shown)


vitorfloriano:
@camilamacedo86 all the 20 token permission vulnerabitilies found by scorecard were remediated by this patch.

.github/workflows/docs-accessibility.yml
```diff
@@ -8,14 +8,20 @@ on:
...
+      contents: read
     name: "Check Docs Accessibility"
     runs-on: ubuntu-latest
     # Pull requests from the same repository won't trigger this checks as they were already triggered by the push
     if: (github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository)
     steps:
       - name: Clone the code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
```
vitorfloriano, line 25: This was necessary to pass CI. Considering it's within scope (security), I added to this PR.

.github/workflows/apidiff.yml
```diff
@@ -8,6 +8,8 @@ on:
     paths-ignore:
       - '**/*.md'
 
+permissions: {}
```
camilamacedo86, line 11: I think all that is needed is permission to read.
Right? Only the goreleaser one needs to write too.

vitorfloriano: @camilamacedo86 the best practice is to be very strict in top-level and add granular permissions in job-level.

I have already added all the necessary granular permissions to all jobs in Kubebuilder CI in the previous PR (#5565), and those permissions override the restrictive top-level permission. The linter I added also checks for this.

For example, goreleaser needs `contents: write` for its job. Since I moved this permission to job-level, it will override the top-level `permissions: {}`.

👉🏻 **This is also a best practice adopted in Kubernetes.** 

See: https://github.com/kubernetes/release/pull/4041/changes (notice how they set top-level to `permissions: {}` and add more permission in job-level as needed)

camilamacedo86: Thank you !! I was looking on that and I figure out it.
So, thank you so much to confirm and sharing as to do this nice work !!

PR #5572, opened 2026-04-02
🐛fix(helm/v2alpha): fix duplicate tolerations block in generated manager template

### LINKED ISSUE #5569: helm/v2-alpha: Tolerations added both to the default values and to the manager template

In my project created with kubebuilder 4.9.0, I added some tolerations to the manager template in `config/manager/manager.yaml`.
Running `kubebuilder edit --plugins=helm/v2-alpha` created a Chart, where the tolerations are configured in the template `dist/chart/templates/manager/manager.yaml`.

Yesterday I upgraded kubebuilder from version 4.9.0 to 4.13.1, and successfully re-scaffolded the project via `kubebuilder alpha update`.
Now the chart generation extracts these tolerations to helm values, stored in `dist/chart/values.yaml`, and also adds them to the template `dist/chart/templates/manager/manager.yaml` causing a duplication and parsing errors when the chart is installed.

I consider this wrong, since the tolerations are now added to the default values, then they shouldn't be added to the manager template too.

### Reproducing this issue

_No response_

### KubeBuilder (CLI) Version

4.13.1

### PROJECT version

3

### Plugin versions

```yaml
layout:
- go.kubebuilder.io/v4
plugins:
  helm.kubebuilder.io/v2-alpha:
    manifests: dist/install.yaml
    output: dist
```

### Other versions

go 1.26.1
sigs.k8s.io/controller-runtime v0.23.3
kubectl v1.34.6
kubernetes v1.33.8-gke.1026000

### Extra Labels

_No response_

v47:
**Description:**
When a project with hand-authored `tolerations` in `config/manager/manager.yaml` runs `kubebuilder edit --plugins=helm/v2alpha`, the generated `dist/chart/templates/manager/manager.yaml` ends up with a duplicate `tolerations` block: the raw list items from the source manifest are left behind alongside the new Helm `{{- with .Values.manager.tolerations }}` template block. This causes `helm install` / `helm upgrade` to fail with a YAML parse error.

**Root cause:**
`templateBasicWithStatement` determined the end of a YAML block with `indentLenLine <= indentLenSearch`. `tolerations` is sequence-valued — its list items (`- effect: NoSchedule`) sit at the same indentation as the parent's tolerations:` key, so the check fired on the first list item and left every `- effect: ...` line behind. `nodeSelector` and `affinity` are map-valued (children always at greater indentation) and were not affected.

**Changes:**
Fixed by updating the inner block-end scan to continue past lines at equal indentation when they start with `- ` (YAML sequence items), only stopping when it reaches a sibling key at equal indentation or a line at lesser indentation. Also adds a break after the first key match in the outer loop, tightens the match condition to require a trailing colon, and adds an idempotency guard.

Fixes #5569

### DIFF
--- a/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
+++ b/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
@@ -1120,6 +1120,10 @@ func (t *HelmTemplater) templateBasicWithStatement(
 	parentKey string,
 	valuePath string,
 ) string {
+	if strings.Contains(yamlContent, valuePath) {
+		return yamlContent
+	}
+
 	lines := strings.Split(yamlContent, "\n")
 	yamlKey := fmt.Sprintf("%s:", key)
 
@@ -1151,9 +1155,9 @@ func (t *HelmTemplater) templateBasicWithStatement(
 		}
 		_, indentLen = leadingWhitespace(lines[start])
 	} else {
-		// Find the existing block
+		// Find the existing block - stop at the first match.
 		for i := range len(lines) {
-			if !strings.HasPrefix(strings.TrimSpace(lines[i]), key) {
+			if !strings.HasPrefix(strings.TrimSpace(lines[i]), yamlKey) {
 				continue
 			}
 			start = i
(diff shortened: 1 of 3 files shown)


pkg/plugins/optional/helm/v2alpha/scaffolds/internal/kustomize/helm_templater.go
```diff
@@ -1181,6 +1120,10 @@ func (t *HelmTemplater) templateBasicWithStatement(
 	parentKey string,
 	valuePath string,
 ) string {
```
camilamacedo86, line 1122: Hi @v47 

Thank you for looking on that. The bug here seems to be:

For any value exposed by default (see: https://github.com/kubernetes-sigs/kubebuilder/issues/5544#issuecomment-4142477769 for more context): if that exist in the YAML we should ensure that it is added to the values with the manager values. We should not expose those empty.

v47: Hi @camilamacedo86,

Thanks for the context from  #5544. I understand the distinction - these fields should only appear in `values.yaml` when actually found in the source deployment, not always with empty defaults.

The current PR only fixes the template duplication bug. The `values_basic.go` change to suppress empty defaults for tolerations, affinities, and nodeSelectors fields when absent is a separate improvement. Would you prefer that to be addressed in this PR or tracked as a follow-up? Happy to extend the scope, if that's the preferred approach.

camilamacedo86: Hi @v7, 

My apprologies, I understand now. Thank you for the clarification.
regards the https://github.com/kubernetes-sigs/kubebuilder/issues/5544 that is out of scope here.
We can discuss that in the right place I am working a PR for that.

camilamacedo86 (approved):
/lgtm

Thank you so much for the nice contirbution 🎉 

/ok-to-test

PR #5466, opened 2026-02-11
✨ feat(alpha/update): add git commands to GH issue description

vitorfloriano:
The issue created by `alpha update` now contains git commands to facilitate checking out the branch generated by the `alpha update` command **when there are merge conflicts**.

PR #5584, opened 2026-04-05
WIP: chore(alpha update/AutoUpdate): Changes to ensure better usage and allow granular workflow and permissions

camilamacedo86:
Closes: https://github.com/kubernetes-sigs/kubebuilder/pull/5466

### DIFF
--- a/docs/book/src/getting-started/testdata/project/.github/workflows/auto_update.yml
+++ b/docs/book/src/getting-started/testdata/project/.github/workflows/auto_update.yml
@@ -1,9 +1,9 @@
 name: Auto Update
 
 # The 'kubebuilder alpha update' command requires write access to the repository to create a branch
-# with the update files and allow you to open a pull request using the link provided in the issue.
+# with the update files and optionally create a pull request (--open-gh-pr) and/or issue (--open-gh-issue).
 # The branch created will be named in the format kubebuilder-update-from-<from-version>-to-<to-version> by default.
-# To protect your codebase, please ensure that you have branch protection rules configured for your 
+# To protect your codebase, please ensure that you have branch protection rules configured for your
 # main branches. This will guarantee that no one can bypass a review and push directly to a branch like 'main'.
 permissions: {}
 
@@ -16,7 +16,8 @@ jobs:
   auto-update:
     permissions:
       contents: write  # Create and push the update branch
-      issues: write    # Create GitHub Issue with PR link
+      issues: write  # Create GitHub Issue notification
+      pull-requests: write  # Create Pull Request
     runs-on: ubuntu-latest
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -57,10 +58,11 @@ jobs:
       # --force: Completes the merge even if conflicts occur, leaving conflict markers.
       # --push: Automatically pushes the resulting output branch to the 'origin' remote.
       # --restore-path: Preserves specified paths (e.g., CI workflow files) when squashing.
-      # --open-gh-issue: Creates a GitHub Issue with a link for opening a PR for review.
+      # --open-gh-issue: Creates a GitHub Issue.
+      # --open-gh-pr: Creates a GitHub Pull Request directly for review.
       #
       # WARNING: This workflow does not use GitHub Models AI summary by default.
-      # To enable AI-generated summaries in GitHub issues, you need permissions to use GitHub Models.
+      # To enable AI-generated summaries, you need permissions to use GitHub Models.
       # If you have the required permissions, re-run:
       #   kubebuilder edit --plugins="autoupdate/v1-alpha" --use-gh-models
       run: |
@@ -68,4 +70,5 @@ jobs:
           --force \
           --push \
           --restore-path .github/workflows \
-          --open-gh-issue
+          --open-gh-issue \
+          --open-gh-pr
--- a/internal/cli/alpha/internal/update/update.go
+++ b/internal/cli/alpha/internal/update/update.go
@@ -25,6 +25,7 @@ import (
 	log "log/slog"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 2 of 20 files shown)


docs/book/src/getting-started/testdata/project/.github/workflows/auto_update.yml
vitorfloriano, line 1: @camilamacedo86

WDYT about moving the bulk of this workflow logic into a [Composite Action](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action) within the Kubebuilder repo?

We could store the implementation in `./actions/autoupdate/action.yml`. This would keep the scaffolded workflow much leaner and encapsulate the "setup" steps (like installing the `kubebuilder` binary and the `gh-models` extension).

The scaffolded workflow would have a step like this:

```yaml
    - name: Run Kubebuilder Auto Update
      uses: kubernetes-sigs/kubebuilder/actions/autoupdate@<SHA> # vX.Y.Z
      with:
        token: ${{ secrets.GITHUB_TOKEN }}
        force: true
        push: true
        restore-path: .github/workflows
        merge-message: "chore..."
        conflict-message: "chore..."
        # use-gh-models
        # open-gh-issue
```
### Why I think this is better:

**Cleaner Scaffold**: Users don't have to manage 20+ lines of shell scripts. It's also easier to pass options using `with:`.

**Easier Updates**: If we need to change how the update runs, we update the action here in the repo, and users get a simple Dependabot PR to bump the version.

camilamacedo86: **I think this may be the right direction**, but there is a clear tradeoff.

A composite action would make the scaffold much cleaner and centralize maintenance, which is very attractive. On the other hand, the current workflow is very explicit, which helps users understand what is happening, troubleshoot issues, and run the same steps manually if needed.

So this is mainly a tradeoff between cleaner UX and easier maintenance vs. transparency and debuggability.

| Approach | Pros | Cons |
|---|---|---|
| **Current explicit workflow** | - Easy to read step by step<br>- Easy for users to troubleshoot<br>- Easy to run manually outside CI (users can see exactly what it does)<br>- Clear permissions and flags<br>- Easy to customize per repo<br>- No extra release or infra needed (everything stays in the same repo/branch) | - More YAML in the scaffold<br>- More setup logic visible to users<br>- Harder to update across many repos<br>- Some duplicated logic |
| **Composite action** | - Much cleaner workflow<br>- Centralized logic (update in one place)<br>- Easier to roll out improvements<br>- Users just bump the version<br>- Less duplicated setup code | - More of a black box<br>- Harder to see what is happening<br>- Debugging can be harder<br>- Manual reproduction less obvious<br>- Less flexible for custom changes<br>- Requires release process (versioning, CI, maintenance)<br>- May need separate branch or repo to manage properly |

So I think we can move forward with the current approach for now, since everything is clear and lives in one place. The logic still the same. 

Then as a next step, we can evaluate moving to a composite action. will we create branch for that? we will need to decide either how to release, etc. We might want as a follow up now or later just deprecate the AutoUpdate plugin in favor of the Composite action.

vitorfloriano: > Then as a next step, we can evaluate moving to a composite action. will we create branch for that? we will need to decide either how to release, etc. We might want as a follow up now or later just deprecate the AutoUpdate plugin in favor of the Composite action.

I believe we would still need the AutoUpdate plugin to add the workflow to the scaffold. The only difference is that we would use the actions in our repo instead of scaffolding the logic in the autoupdate.yml itself. 

How we implement the action would be up to us. 

In regards to release, I think that anything in `./actions/...` would be tagged along with the repo, wouldn't it? 

Then we would only need to scaffold a workflow that has something like:

```yaml
steps:
  - name: AutoUpdate
  - uses: kubernetes-sigs/kubebuilder/actions@v4.13.0 # example version
    with:
      force:
      all the other flags here...
```
Then Dependabot could bump that version to get the latest improvements in that action.

But sure, we can leave that to another PR.

camilamacedo86: IHMO, I think you’re thinking in the right direction, and I believe that’s a good direction to pursue
Lets try to do that next, after we change the behaviour we can try to do it as next first thing to do.

internal/cli/alpha/internal/update/update.go
vitorfloriano, line 1: @camilamacedo86 I think I'm onto something. Bear with me for a minute:

> TL;DR
> I think we can improve the workflow by:
> 1. Perform the update using only two branches (instead of four).
> 2. Open the PR against the snapshot-of-main branch, not against main.

1. WHAT IF instead of creating four different branches (ancestor, current, update and merge), we only create two? We could rename them accordingly.

Currently, we create four different branches to perform the update. Something like this:

```mermaid
---
config:
  gitGraph:
  theme: base
---
gitGraph
    commit id: "HEAD"
    branch ancestor
    commit id: "alpha generate with from version"
    branch current
    commit id: "restore main to copy current state"
    checkout ancestor
    branch update
    commit id: "alpha generate with to version"
    branch merged
    merge current id: "merge the updated branch into current"
    checkout main
    merge merged id: "merge the final product into main"
```
But thinking back, we could have achieved the same by creating only two branches:

```mermaid
---
config:
  gitGraph:
  theme: base
---
gitGraph
    commit id: "HEAD"
    branch ancestor/update
    commit id: "alpha generate with from version"
    branch current
    commit id: "restore main to copy current state"
    checkout ancestor/update
    commit id: "alpha generate with to version"
    merge current id: "merge the updated branch into current"
    checkout main
    merge ancestor/update id: "merge the final product into main"
```
Much simpler, and achieves the same. A three-way merge that preserves custom code, with less cleanup to do after the operation.

Now here comes my epiphany: 

2. AND WHAT IF we open a PR against the copy of main we create, instead of main, without forcing the merge conflicts? That way, we could open PRs that lead to merge conflicts and developers would have a chance to resolve them manually or with AI assistance!

If we open a PR from the updated branch onto the snapshot branch, we'll get merge conflicts. If we try to open the PR against main, we won't get any conflicts, because main and our update branch don't share the same parent commit.

Take a look at this: 

A PR with merge conflicts: https://github.com/vitorfloriano/test-update-operator/pull/3
The same PR that had merge conflicts now resolved by AI: https://github.com/vitorfloriano/test-update-operator/pull/2

If I had open the PR against main, I would have overwritten a bunch a stuff and AI would have a lot of work to do: https://github.com/vitorfloriano/test-update-operator/pull/4

Also, I had overlooked the fact that it was possible to open PRs against other branches and they would appear in the PR tab in the repo in the same way as the others.

Anyway, I know it's a lot to take in. But give it a thought or two and let me know. It seems to me to be a good direction to explore.

camilamacedo86: I think this is definitely something worth exploring, but it’s out of scope for this PR. If we try to tackle everything at once, it becomes harder to review and increases the risk of introducing issues.

I’d suggest breaking this down into smaller steps:

1. Fix the current behavior (especially around granular permissions). We can open a separate PR or create an issue to track this.
2. Explore using a composite action.
3. Look into improving or enhancing the current logic.

I’d prioritize (1) and (2) first since they directly impact end users. Point (3) is more about internal improvements to how things work.

That said, we should definitely follow up on this—great thinking 🎉

camilamacedo86:
Closing this one.
We need to ensure that the design doc covers it:
a) We cannot by default open the PR because by default it is not allowed and often requires org permissions. That is why we did it via an issue. 
b) However, we can add an option to raise the PR automatically when this permission can be granted.

Issue #5593, opened 2026-04-09
Kubebuilder Book: confusing comment describing controller requeue


kylittle:
### What do you want to happen?

I want to preface by saying I am new to Kubebuilder and as such this issue may not really be that important.

I was going through the kubebuilder book and came across the line:

[cronjob_controller.go](https://github.com/kubernetes-sigs/kubebuilder/blob/5dead900137df5a127a822eaca1f71a75ef7ab74/docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller.go#L664-L665)
```
// actually make the job...
job, err := constructJobForCronJob(&cronJob, missedRun)
if err != nil {
	log.Error(err, "unable to construct job from template")
	// don't bother requeuing until we get a change to the spec
	return scheduledResult, nil
}
```

This code works fine but I feel like the comment could be improved for learning purposes. It seems like perhaps it might be better to say something like
```
// don't immediately requeue, wait till the next window since the spec needs to be fixed
```

Or something along those lines. If the maintainers agree that this is clearer, I would be happy to submit a PR with this change.

### Extra Labels

/kind documentation

vitorfloriano:
Hi @kylittle 

Thanks for the feedback! 

IMHO it has more to do with the original author's writing style. The inline comments have the same playful and witty tone as the rest of the tutorial. 

Perhaps that's the voice the author chose to go with, a tradeoff between being instructional vs conversational.

But yeah, a bit of explaining wouldn't hurt. 

Feel free to open a PR with your suggestion!

PR #5602, opened 2026-04-10
📖 fix(docs): changed comment wording

kylittle:
This PR just changes a single comment as mentioned in #5593

Hopefully this is more clear? I definitely get and appreciate the tone of the book (it was very helpful)!

### DIFF
--- a/docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller.go
@@ -662,7 +662,7 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	job, err := constructJobForCronJob(&cronJob, missedRun)
 	if err != nil {
 		log.Error(err, "unable to construct job from template")
-		// don't bother requeuing until we get a change to the spec
+		// don't immediately requeue, wait till the next window since the spec needs to be fixed
 		return scheduledResult, nil
 	}
 
--- a/docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller.go
+++ b/docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller.go
@@ -662,7 +662,9 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	job, err := constructJobForCronJob(&cronJob, missedRun)
 	if err != nil {
 		log.Error(err, "unable to construct job from template")
-		// don't immediately requeue, wait till the next window since the spec needs to be fixed
+		// don't retry immediately; this failure occurred while constructing the Job.
+		// we'll reconcile again at the next scheduled run, and updates to the CronJob
+		// can also trigger reconciliation sooner
 		return scheduledResult, nil
 	}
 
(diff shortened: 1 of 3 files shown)


docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller.go
```diff
@@ -662,7 +662,7 @@ func (r *CronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	job, err := constructJobForCronJob(&cronJob, missedRun)
 	if err != nil {
 		log.Error(err, "unable to construct job from template")
-		// don't bother requeuing until we get a change to the spec
+		// don't immediately requeue, wait till the next window since the spec needs to be fixed
```
camilamacedo86:
```suggestion
		// don't immediately requeue; since the failure is caused by the CronJob spec,
		// we rely on a future update to the resource to trigger reconciliation
```

WDYT?

camilamacedo86:
Hi @kylittle,

Thanks a lot for raising this!

I’ve left a suggestion—could you please take a look? Also, can you ensure the change is reflected in the hack docs here:
https://github.com/kubernetes-sigs/kubebuilder/blob/e2ba0c971e21588fc0fdd65d39ef7276fef70f9f/hack/docs/internal/cronjob-tutorial/controller_implementation.go#L654

All samples in the docs are generated via `make generate-docs`, so please make sure to apply your change and run this command. Otherwise, the update won’t persist and CI will fail.

Thanks again! Looking forward to the update so we can get this merged 🚀

camilamacedo86 (changes requested):

docs/book/src/cronjob-tutorial/testdata/project/internal/controller/cronjob_controller.go
kylittle: This looks a lot better than what I had, though I am wondering if there should be an extra line so it ends up like:

```
// don't immediately requeue; since the failure is caused by the CronJob spec,
// we rely on a future update to the resource to trigger reconciliation
// note that the job is still requeued for the next window so we don't miss our execution
```
Since unless I am mistaken I think that is what the code actually does right? Not sure I like the new sentence since the execution would probably still be missed since the spec is not going to magically fix itself without an event driven retrigger I would guess.

I am wondering if it would be better to set the status to degraded here with a reason of "InvalidSpec", then simply return a non-error non-requeue result?

camilamacedo86: The trigger will occur here because we `.For(&batchv1.CronJob{})` that means we are watching the changes in the spec. 

```
// don't immediately requeue; this error is caused by the CronJob spec,
// so another reconciliation would not help until the resource is changed.
// when the CronJob is updated, the watch configured with For(&batchv1.CronJob{})
// will trigger reconciliation again
```

kylittle: So I could be misunderstanding but it seems like we also requeue anyway since in the next line we go:

```
scheduledResult := ctrl.Result{RequeueAfter: nextRun.Sub(r.Now())} // save this so we can re-use it elsewhere
...
// Skipped a bunch here but after the comment we do:
return scheduledResult, nil
```

So I guess what I find confusing (unless I misunderstand how kubebuilder works) is that if we don't edit the spec it seems like we still requeue at the time nextRun is. So while I agree that editing the spec triggers reconciliation it also seems like we still attempt again at the next window regardless of if the spec was changed.

I am wondering if the comment should reflect this or if perhaps we should mark the Status as degraded so that the spec must be changed before it will requeue?

camilamacedo86: WDYT about?

```
// don't retry immediately; this failure occurred while constructing the Job.
// we'll reconcile again at the next scheduled run, and updates to the CronJob
// can also trigger reconciliation sooner
```

It might clarify better.

camilamacedo86 (approved):

Issue #5517, opened 2026-02-27
helm/v2-alpha: We had issues while migrating to from helm/v1-alpha1


videlov:
### What do you want to happen?

With `helm/v2-alpha` on re-generating the Helm scaffolding with `kubebuilder edit --plugins=helm/v2-alpha`
it did replace previous customisation we had on dist/chart/templates/manager/manager.yaml

we were additionally customising per values additionally:
* pod labels
* volumes and volumeMounts
* additional manager args
* env variables
* we had enable flags for manager and rbac resources
* service account name

Options we would like to have:
* implement eventually a flag that it does not recreate the manager.yaml template on scaffolding (which we would like to run continuously on pipeline runs).
* since helm generation happens based on kustomize output of make target `build-installer`, introduce a flag that can specify on which make target to use the output from. we have 2 flavours of chart, and could have multiple `config`'s for kustomize.

camilamacedo86:
Hi @videlov,

We release 4.13.0 with some bug fixes for helm/v2-alpha. 

However, some of things expose for end users I think we have open RFEs to address:

- pod labels
- volumes and volumeMounts
- we had enable flags for manager and rbac resources
- service account name

Also, I think the following RFEs are nice use case:

- implement eventually a flag that it does not recreate the manager.yaml template on scaffolding (which we would like to run continuously on pipeline runs).

Then, regards it should be supported already.

- additional manager args
- env variables

Could you please check?

Moreover, If you be able to share your solution that can either help.

By last, could you please let us know more about "since helm generation happens based on kustomize output of make target build-installer, introduce a flag that can specify on which make target to use the output from. we have 2 flavours of chart, and could have multiple config's for kustomize." What do you mean with that? Could you please add something like "today we do A" and we "have B" when we would to "have C and D"?

camilamacedo86:
Hi @videlov 

I think we could address all items that you raised.
Your help to test things out before the next release is very appreciated.

If you can, please pull the master branch locally and run: `make install` and try out the plugin using the flag --force or by removing your chart and re-generating it with this one. 

See: https://github.com/kubernetes-sigs/kubebuilder/blob/master/CONTRIBUTING.md#how-to-build-kubebuilder-locally

Thank you so much. 
Please, feel free to raise new issues and suggestions. 

I am marking this one as close.

PR #5603, opened 2026-04-10
✨ (helm/v2-alpha): Expose service account into the values

camilamacedo86:
**Motivations**
Address request in: https://github.com/kubernetes-sigs/kubebuilder/issues/5517 and in https://github.com/kubernetes-sigs/kubebuilder/pull/5579#issuecomment-4193243070 (partially)

### DIFF
--- a/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
+++ b/pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
@@ -64,7 +64,7 @@ func (f *HelmHelpers) generateHelpersTemplate() string {
 	// preventing collisions when chart is used as a Helm dependency
 	prefix := f.ProjectName
 
-	return fmt.Sprintf(helmHelpersTemplate, prefix, prefix, prefix, prefix, prefix)
+	return fmt.Sprintf(helmHelpersTemplate, prefix, prefix, prefix, prefix, prefix, prefix, prefix)
 }
 
 const helmHelpersTemplate = `{{` + "`" + `{{/*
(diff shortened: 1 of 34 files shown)


camilamacedo86:
Hi @sam-mcbr, @videlov, @v47, @porridge, @Alex1 

Could you please look for this one?
Your input is very helpful.

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
```diff
@@ -118,4 +118,17 @@ Dynamically calculates safe truncation to ensure total name length <= 63 chars.
 {{` + "`" + `{{- printf "%%s-%%s" $fullname $suffix | trunc 63 | trimSuffix "-" }}` + "`" + `}}
 {{` + "`" + `{{- end }}` + "`" + `}}
 {{` + "`" + `{{- end }}` + "`" + `}}
+
+{{` + "`" + `{{/*
```
v47, line 122: It's a potentially confusing contract:
* the docs/feature framing suggest that disabling creation means - use an external SA that will be specified
* while the helper behavior may allow: disable creation, specify nothing, silently fall back to default name.
I think, i already have seen something simular in the codebase. I think it was related to the multy controller scafolding.

camilamacedo86: I think that is simple:
- If you want use an external SA than you inform the name otherwise will use the one that is shipped with ( default behaviour ) 

We need to expose the SA as allow use external one and I do not see a better, easier way than that.

v47: Thanks, it makes sense. I just overthought the flows.
A tiny example for the external SA with explicit name case can be useful, and another sentence for the reuse the default name, but do not create it case.

v47 (approved):
Looks good.

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/templates/chart-templates/helpers_tpl.go
camilamacedo86: Before the release, let’s take some time to review all the changes and shape everything together.

Since we’ve been adding features and documentation incrementally, it’s become a bit hard to see the full picture as a whole.

So please feel free to contribute—this will help us ensure everything is in great shape before we release.

Your help is essential for us 🎉

PR #5609, opened 2026-04-12
🌱 (ci):Add kube linter to helm verify and linter checks

camilamacedo86:


Makefile
```diff
@@ -219,7 +219,12 @@ test-legacy:  ## Run the tests to validate legacy path for webhooks
 
 .PHONY: install-helm
 install-helm: ## Install the latest version of Helm locally
-	@curl -fsSL https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-4 | bash
+	@command -v helm >/dev/null 2>&1 || curl -fsSL https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-4 | bash
+
+.PHONY: install-kube-linter
+install-kube-linter: $(KUBE_LINTER) ## Install kube-linter locally if necessary
+$(KUBE_LINTER): $(LOCALBIN)
+	$(call go-install-tool,$(KUBE_LINTER),golang.stackrox.io/kube-linter/cmd/kube-linter,$(KUBE_LINTER_VERSION))
```
vitorfloriano, line 227: I'm getting this message when running this target:
```
$ make install-kube-linter 
make: Nothing to be done for 'install-kube-linter'.
```

camilamacedo86: Fixed

Makefile
```diff
@@ -228,6 +233,13 @@ helm-lint: install-helm ## Lint all Helm charts in testdata and docs
 		helm lint $$chart || exit 1; \
 	done
 
+.PHONY: kube-linter
+kube-linter: install-helm install-kube-linter ## Lint all Helm charts with kube-linter
+	@for chart in $(HELM_CHARTS); do \
+		echo "Validating $$chart with kube-linter..."; \
```
vitorfloriano, line 239: Question: isn't this going to fail in CI? I mean, if we don't address the unpinned tag error (which in our case is not an issue). 

Is there any flag to silent this audit? Or will we need to add yet another lint config file?

camilamacedo86: I think we can address that: :https://github.com/kubernetes-sigs/kubebuilder/pull/5617

camilamacedo86:
HI @vitorfloriano 

Moving forward. It is just a CI check for us ( maintainers )
If we see that does not work well in the long term we can remove

Thank you for suggest that. 🎉

PR #5620, opened 2026-04-14
 📖 Add update patch apply docs

guettli:
Add docs about when to use Update, Patch, Apply.


camilamacedo86 (changes requested):
Thanks for working on this.

I initially thought this would be about `WithFieldValidation` (https://github.com/kubernetes-sigs/kubebuilder/issues/5208), rather than PUT/PATCH/APPLY methods—but I appreciate you raising this.

That said, this section feels a bit over-described and not fully accurate.
I think this could be simplified to a short note, for example:

> Use `PUT` for replacement-style updates and `PATCH` for partial updates. In Kubernetes, server-side apply is implemented as a specialized `PATCH` operation. See: https://kubernetes.io/docs/reference/using-api/api-concepts/

I am not sure where the note would fit better.
Maybe when we support SSA (https://github.com/kubernetes-sigs/kubebuilder/pull/5458) we can add that.

camilamacedo86:
Hi @guettli 

Based on this https://github.com/kubernetes-sigs/kubebuilder/pull/5620#pullrequestreview-4115202356 I am closing this one as defered. I hope that you do not mind.

If you do not agree please feel free to re-open, Also,see that you need sign CLA to allow us get any contribution merged. Thank you so much for your help and input. 

Cheers,

PR #5621, opened 2026-04-14
🌱 (helm/v2-alpha): Refactory the code

camilamacedo86:
Refactors the Helm v2alpha plugin’s Helm chart scaffolding pipeline into distinct stages (parse → extract → template → generate), moving extraction/templating responsibilities into dedicated packages and updating generated chart/test artifacts accordingly.

**Changes:**
- Introduces a new `extractor` package for metadata/feature/value extraction and wires it into a new chart scaffolder.
- Replaces the previous writer/organizer approach with a generator + dynamic machinery templates (`templater` + `appliers` + `DynamicTemplate`).
- Updates chart templates/values and expands/adjusts integration + unit test coverage.
- Address review feedback and fixes

Generated-by: Claude


pkg/plugins/optional/helm/v2alpha/scaffolds/internal/extractor/extractor.go
v47, line 1: 🚀

pkg/plugins/optional/helm/v2alpha/internal/common/consts.go
```diff
@@ -0,0 +1,62 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package common provides shared constants for the Helm v2-alpha plugin.
+package common
```
v47, line 18: The naming here may be a bit too generic for what this package is actually doing.

`internal/common/consts.go` sounds like broadly shared plugin constants. Still, the contents look more specific to manifest/chart templating: resource kinds, API versions, YAML key markers, Helm/Kubernetes label keys, plus the default chart output dir.

Would it make sense to rename this package to reflect that scope more explicitly? For example, something along the lines of `internal/templater/constants.go`, depending on which layer is intended to own these values.

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/extractor/deployment.go
```diff
@@ -0,0 +1,593 @@
...
+
+	port, err := strconv.Atoi(portPart)
+	if err != nil || port <= 0 || port > 65535 {
+		return 0
+	}
+	return port
+}
+
+// extractContainerPorts extracts port configurations from container ports.
+func extractContainerPorts(container map[string]any, config map[string]any) {
```
v47, line 404: The containerPort parsing only handles int64 and int. For unstructured-decoded objects, numeric fields can come back in more than one concrete type depending on the decode path, so this feels narrower than it should be.

For example, something like that can be an option:
```
func extractContainerPorts(container map[string]any, config map[string]any) {
	portsField, found, err := unstructured.NestedSlice(container, "ports")
	if !found || err != nil {
		return
	}

	for _, p := range portsField {
		portMap, ok := p.(map[string]any)
		if !ok {
			continue
		}

		name, _ := portMap["name"].(string)
		containerPort, ok := toInt(portMap["containerPort"])
		if !ok {
			continue
		}

		if isWebhookPortName(name) {
			if _, exists := config["webhookPort"]; !exists {
				config["webhookPort"] = containerPort
			}
		}
	}
}

func isWebhookPortName(name string) bool {
	name = strings.ToLower(name)
	return name == "webhook-server" || name == "webhook"
}

func toInt(v any) (int, bool) {
	switch n := v.(type) {
	case int:
		return n, true
	case int32:
		return int(n), true
	case int64:
		return int(n), true
	case float64:
		return int(n), true
	default:
		return 0, false
	}
}
```

pkg/plugins/optional/helm/v2alpha/internal/common/consts.go
camilamacedo86: I understand from where you came from.
However, one of the const is the path for the default output (dist)
So if we add the consts at internal/templater/ we cannot reuse across all :-(

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/extractor/deployment.go
camilamacedo86: Done. Simplified good idea.

pkg/plugins/optional/helm/v2alpha/internal/common/consts.go
camilamacedo86: This one is the const:

```
// DefaultOutputDir is the default output directory for Helm charts.
const DefaultOutputDir = "dist"
```

pkg/plugins/optional/helm/v2alpha/scaffolds/internal/chart_scaffolder.go
```diff
@@ -0,0 +1,175 @@
...
+			OutputDir: s.config.OutputDir,
+			Force:     s.config.Force,
+		},
+	}
+
+	// Add generic ServiceMonitor only if project doesn't provide one via kustomize
+	if !extraction.Features.HasPrometheus {
+		metricsServiceName := extraction.Metadata.DetectedPrefix + "-controller-manager-metrics-service"
+		for _, svc := range resources.Services {
+			svcName := svc.GetName()
```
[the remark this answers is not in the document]
camilamacedo86: We should not remove

camilamacedo86:
Hi @v47 

I will move forward with this one (as an exception bases) to facilitate contributions on top.
I will ask help for everybody try out helm/v2-alpha all master after all changes 
Also, if you see any hall for improvements please feel free to send a PR.

Your help is very welcome. ❤️

Issue #5551, opened 2026-03-26
Add AI skills in the scaffolds


camilamacedo86:
### What do you want to happen?

We might able to scaffold AI skill to help users work with the end solution for example.

nerdeveloper:
i submitted a propoal for this which was later closed,  so do you want me to reopen it?

camilamacedo86:
> i submitted a propoal for this which was later closed, so do you want me to reopen it?

I think we should evaluate the possibility of creating a specialized DevOps, Kubebuilder skill for Operators. We might also want to explore a separate skill focused on observability for Kubernetes operators, especially for code review use cases.

Before defining the solution, we should identify the most common workflows and the target personas. From my perspective, this seems slightly different from what you proposed earlier, unless I’m mistaken.

camilamacedo86:
We might want here start for skills for Kubebuilder repo itself and then see if we want automate some process. 
Aka review, bump of dependencies things like

camilamacedo86:
The no vendoring way ( generic to work with any AI ) is by using SKILLS.MD , see; https://agentskills.io/what-are-skills

c/c @nerdeveloper (Just to share)

rayowang:
@camilamacedo86 Hi, camila,Thanks for raising this issue and sharing the context around AI agent skills!

I’ve been working on internal agent/skill-related tooling at my company recently, and I’ve spent some time putting together this initial implementation to address the issue.

I’ve submitted PR #5588 that adds a vendor-neutral SKILLS.MD at the repository level and scaffolds it into go/v4 projects, while keeping AGENTS.md for backward compatibility.

This is intended as a practical baseline for generic AI tools to understand project layout, commands, conventions, and common workflows. It’s kept minimal as a first step and can later be split into more focused, specialized skills.

Looking forward to your feedback!

vitorfloriano:
For scaffolds, I think we should package skills in a plugin. Here's why:

### What's an agent skill

At its core, a skill is a **folder** containing a SKILL.md file and relevant ancillary files. It works best when you organize the skills into separate subdirectories. Something like this:

```
.
└── .agents/
    └── skills/
        ├── kubebuilder-migration/
        │   ├── SKILL.md
        │   ├── from-v1-to-v2.md
        │   ├── from-v2-to-v3.md
        │   ├── from-v3-to-v4.md
        │   ├── migration-docs.md
        │   ├── any-relevant-info.md
        │   └── ...
        ├── crd-development/
[...]

PR #5635, opened 2026-04-16
🌱 Add maintenance skill for helm/v2

camilamacedo86:
Related: https://github.com/kubernetes-sigs/kubebuilder/issues/5551


camilamacedo86:
@vitorfloriano 

I am moving forward with this one since is valid only for us and we could define the structure.
We can shape it in follow up PRs. So that it allows we receive contributions here as well. 

Please, feel free to raise PR agains this one.
Thank you so much for all your help.

PR #5644, opened 2026-04-17
🐛 fix: Pin otel/sdk in testdata to address 2 CVEs

vitorfloriano:
This PR pins `go.opentelemetry.io/otel/sdk` to `v1.43.0` to remediate vulnerabilities found in **testdata**.

See: [CVE-2026-39883](https://github.com/advisories/GHSA-hfvc-g4fc-pqhx) and [CVE-2026-24051](https://github.com/advisories/GHSA-9h8m-3fm2-qjrq).


test/testdata/generate.sh
```diff
@@ -141,6 +141,10 @@ function scaffold_test_project {
   rm -f go.sum
   # Pin google.golang.org/grpc to the patched version (CVE: SNYK-GOLANG-GOOGLEGOLANGORGGRPC-15691172)
   go get google.golang.org/grpc@v1.79.3
+  
+  # Pin go.opentelemetry.io/otel/sdk to the patched version (CVE-2026-39883)
+  go get go.opentelemetry.io/otel/sdk@v1.43.0
```
camilamacedo86, line 146: If we remove the both ones would we take the upper version? 
From where the dep came, can we not update the dep directly?

camilamacedo86: The fix must be done in controller-runtime
By pinning here we just fix our docs, samples but we do not solve it for end users. 
We need to raise an issue there: https://github.com/kubernetes-sigs/controller-runtime/issues/3503 and fix there.

vitorfloriano:
Yeah, this needs to be fixed in upstream.

PR #5645, opened 2026-04-17
📖 chore: remove dependency pins and update

camilamacedo86:
Dependency pins create unnecessary maintenance burden. Removed pins, update the certamanger import version used in the testdata to latest and updated all dependencies with `go get -u  ./...`.


docs/book/src/cronjob-tutorial/testdata/project/go.sum
```diff
@@ -1,5 +1,5 @@
-cel.dev/expr v0.25.1 h1:1KrZg61W6TWSxuNZ37Xy49ps13NUovb66QLprthtwi4=
```
camilamacedo86, line 1: @vitorfloriano  see this one.

I do not think we should have the pins in the docs because it is easy to forget, hard to keep things maintained in long term.
Also, I am adding 📖 because it mainly just update the docs not really the end users solutions.

docs/book/src/multiversion-tutorial/testdata/project/go.sum
```diff
@@ -155,24 +155,24 @@ github.com/tidwall/sjson v1.2.5 h1:kLy8mja+1c9jlljvWTlSazM7cKDRfJuR/bOJhcY5NcY=
...
-go.opentelemetry.io/otel/metric v1.39.0/go.mod h1:jrZSWL33sD7bBxg1xjrqyDjnuzTUB0x1nBERXd7Ftcs=
-go.opentelemetry.io/otel/sdk v1.39.0 h1:nMLYcjVsvdui1B/4FRkwjzoRVsMK8uL/cj0OyhKzt18=
-go.opentelemetry.io/otel/sdk v1.39.0/go.mod h1:vDojkC4/jsTJsE+kh+LXYQlbL8CgrEcwmt1ENZszdJE=
-go.opentelemetry.io/otel/sdk/metric v1.39.0 h1:cXMVVFVgsIf2YL6QkRF4Urbr/aMInf+2WKg+sEJTtB8=
-go.opentelemetry.io/otel/sdk/metric v1.39.0/go.mod h1:xq9HEVH7qeX69/JnwEfp6fVq5wosJsY1mt4lLfYdVew=
-go.opentelemetry.io/otel/trace v1.39.0 h1:2d2vfpEDmCJ5zVYz7ijaJdOF59xLomrvj7bjt6/qCJI=
-go.opentelemetry.io/otel/trace v1.39.0/go.mod h1:88w4/PnZSazkGzz/w84VHpQafiU4EtqqlVdxWy+rNOA=
+go.opentelemetry.io/otel/metric v1.36.0 h1:MoWPKVhQvJ+eeXWHFBOPoBOi20jh6Iq2CcCREuTYufE=
+go.opentelemetry.io/otel/metric v1.36.0/go.mod h1:zC7Ks+yeyJt4xig9DEw9kuUFe5C3zLbVjV2PzT6qzbs=
+go.opentelemetry.io/otel/sdk v1.36.0 h1:b6SYIuLRs88ztox4EyrvRti80uXIFy+Sqzoh9kFULbs=
```
vitorfloriano, line 170: To address the CVE we need this to be at least v1.43.0 (patched version), right? 

See: https://github.com/advisories/GHSA-hfvc-g4fc-pqhx

camilamacedo86: We cannot address in the sample.
End users will still facing the issue.
And those pins are hard to maintain, after we forget them and the dep is not properly updated.

The solution is fix controller-runtime, https://github.com/kubernetes-sigs/controller-runtime/issues/3503 then yes, we bump the new controller-runtime version all will be fixed for us and users.

camilamacedo86: Can we move with this one?
WDYT? If you agree could you please LGTM

vitorfloriano (approved):
/lgtm

